Issue #1070, opened 2021-08-04
[BUG] `make ci` fails if there is a harvester directory under `package`


yeahdongcn:
**Describe the bug**

`make ci` fails if there is a `harvester` directory under `package`

**To Reproduce**
1. `make ci` to build the image
2. Acccidently untar package/harvester-0.2.0.tgz (`harvester` directory exists under package)
3. `make ci` again and an error occurs (cp: cannot overwrite directory ./harvester with non-directory ../bin/harvester)

**Expected behavior**

No error occurs

**Support bundle**

**Environment:**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

yeahdongcn:
Since #1071 is merged, let's close this issue.

PR #1071, opened 2021-08-04
Remove harvester directory or file if any

yeahdongcn:

**Problem:**

`make ci` fails if there is a harvester directory under `package`

**Solution:**

Before copying `harvester` binary to package, remove harvester `directory` or file under `package` if any

**Related Issue:**

https://github.com/harvester/harvester/issues/1070

**Test plan:**

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -16,6 +16,7 @@ if [ -e ${DOCKERFILE}.${ARCH} ]; then
     DOCKERFILE=${DOCKERFILE}.${ARCH}
 fi
 
+rm -rf ./harvester
 cp ../bin/harvester .
 
 docker build --build-arg VERSION=${VERSION} -f ${DOCKERFILE} -t ${IMAGE} .


scripts/package
```diff
@@ -16,6 +16,7 @@ if [ -e ${DOCKERFILE}.${ARCH} ]; then
     DOCKERFILE=${DOCKERFILE}.${ARCH}
 fi
 
+rm -rf ./harvester
```
guangbochen, line 19: this is not necessary as we are building with the container runtime, so each time the generated asset is unique with the `make` command.

yeahdongcn: If you look closely at #1070, this might be useful to ensure the local build is more robust.

guangbochen: cool, the reproducing steps make sense, thanks for the contribution.

guangbochen (approved):
LGTM

futuretea (approved):
LGTM

Issue #838, opened 2021-05-08
[BUG] Observed a panic: "invalid memory address or nil pointer dereference"


futuretea:
**Describe the bug**

```bash
❯ kubectl get po
NAME                                                   READY   STATUS             RESTARTS   AGE
cdi-apiserver-54d588847-jhtwd                          1/1     Running            1          149m
cdi-deployment-65fbdb7588-wf9c7                        1/1     Running            1          149m
cdi-operator-74dc8595f5-dbqhz                          1/1     Running            1          150m
cdi-uploadproxy-544fcccc78-96drq                       1/1     Running            1          149m
harvester-9f85f8dd6-52n9t                              0/1     Pending            0          150m
harvester-9f85f8dd6-n27jt                              0/1     Pending            0          150m
harvester-9f85f8dd6-vmpjt                              0/1     CrashLoopBackOff   6          150m
harvester-network-controller-6k4mn                     1/1     Running            1          150m
harvester-network-controller-manager-87d66b4b6-sq8wt   1/1     Running            1          150m
harvester-network-controller-manager-87d66b4b6-zbg8m   1/1     Running            1          150m
virt-api-5fbd56d6c9-4xp65                              1/1     Running            1          148m
virt-api-5fbd56d6c9-8nwpl                              1/1     Running            1          148m
virt-controller-75c7bbb964-fg2ch                       1/1     Running            1          147m
virt-controller-75c7bbb964-t7rgl                       1/1     Running            1          147m
virt-handler-kc7mv                                     1/1     Running            1          147m
virt-operator-7dd868747b-fw62v                         1/1     Running            1          150m
```
```bash
E0508 07:26:26.734594       7 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1305 [running]:
github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime.logPanic(0x22bca40, 0x4326e30)
        /go/src/github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0x92
github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /go/src/github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x82
panic(0x22bca40, 0x4326e30)
[...]

PR #1079, opened 2021-08-09
Fix random api nil pointer exception

guangbochen:
**Problem:**
Randomly Observed a panic: "invalid memory address or nil pointer dereference" of Harvester API server.

**Solution:**
Bump go mod as it is addressed via the dynamicLister  https://github.com/rancher/dynamiclistener/pull/41.

**Related Issue:**
https://github.com/harvester/harvester/issues/838

**Test plan:**
Make sure the API server doesn't return the above error


gitlawr (approved):
LGTM

Issue #755, opened 2021-04-24
[FEATURE] Enhance user experience of Harvester/Rancher UI interaction


yasker:
Currently, we just jump to the Rancher UI as a pop-out window, which left more to be desired in the UX integration.

The team is already working on a design of updated UI interaction between Harvester and the upcoming Rancher 2.6.

We're waiting for a design decision before implementing the enhancement.

yasker:
UI integration PR has been merged. All future issues can be tracked by separate tickets.

PR #1090, opened 2021-08-16
Set default Harvester ui configs and remove built-in node driver

guangbochen:
**Problem:**
Will need to remove the built-in Harvester node driver and patching the rancher dashboard setting configs.

**Solution:**
Set default Harvester UI configs and remove built-in node driver

**Related Issue:**
https://github.com/harvester/harvester/issues/755

**Test plan:**


gitlawr (commented):
Suggestions:
1. Use setting controller instead of polling
2. Update the default once instead of overriding the values every time

guangbochen:
Updated

gitlawr (approved):
LGTM

Issue #1096, opened 2021-08-17
[BUG] harvester network controller manager failed to start


guangbochen:
**Describe the bug**
harvester network controller manager failed to start with the following errors:

```
harv-rmdv [~]$ kubectl -n harvester-system get po
NAME                                                    READY   STATUS             RESTARTS   AGE
harvester-fdb4d775f-hzrgr                               1/1     Running            0          103m
harvester-fdb4d775f-nr8nd                               0/1     Pending            0          103m
harvester-fdb4d775f-q8stb                               0/1     Pending            0          103m
harvester-network-controller-manager-67f7d797db-j2lqj   1/1     Running            20         103m
harvester-network-controller-manager-67f7d797db-w9z7b   0/1     CrashLoopBackOff   20         103m
harvester-network-controller-ml5f6                      1/1     Running            0          103m
harvester-webhook-6d94fd86cb-2kwwc                      0/1     Pending            0          103m
harvester-webhook-6d94fd86cb-5mmfc                      0/1     Pending            0          103m
harvester-webhook-6d94fd86cb-qpgnb                      1/1     Running            0          103m
virt-api-57fd7db5cc-bbmj8                               1/1     Running            0          102m
virt-api-57fd7db5cc-s4wn4                               1/1     Running            0          102m
virt-controller-b49dfc74-8bl9v                          1/1     Running            0          101m
virt-controller-b49dfc74-ltj8j                          1/1     Running            0          101m
virt-handler-ptv54                                      1/1     Running            0          101m
virt-operator-78697f6ff8-5r2jw                          1/1     Running            0          103m
harv-rmdv [~]$ kubectl -n harvester-system logs -f harvester-network-controller-manager-67f7d797db-w9z7b
I0817 09:47:51.387378       1 main.go:105] Starting network controller with 2 threads.
I0817 09:47:51.387495       1 main.go:108] Starting network controller with no namespace.
W0817 09:47:51.388065       1 client_config.go:614] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0817 09:47:51.681326       1 leaderelection.go:243] attempting to acquire leader lease kube-system/harvester-network-controllers...
I0817 09:48:38.127327       1 leaderelection.go:253] successfully acquired lease kube-system/harvester-network-controllers
time="2021-08-17T09:48:55Z" level=info msg="Do not update the existing CRD NetworkAttachmentDefinition"
[...]

PR #1098, opened 2021-08-17
fix harvester-network-controller chart CRD error

guangbochen:

**Problem:**
fix harvester-network-controller chart CRD error

**Solution:**
- replacing the harvester network controller chart with the one from https://github.com/harvester/charts
- update the network controller CRDs according to https://github.com/harvester/network-controller-harvester/commit/c68dcf51f5fd09625640a4c219aff6a1869941d5

**Related Issue:**
https://github.com/harvester/harvester/issues/1096

**Test plan:**
N/A


gitlawr (approved):
LGTM

PR #1107, opened 2021-08-18
Fix builder machine type panic

futuretea:
if VirtualMachine.Spec.Template.Spec.Domain.Machine == nil will panic

**Related issued**
harvester/harvester#1108


guangbochen (approved):
LGTM

Issue #570, opened 2021-03-25
[FEATURE] Image upload


yasker:
We want to support users to upload images from the local desktop to the Harvester cluster.

yaocw2020:
It has been supported in v0.2.0.

yasker:
@yaocw2020 with the removal of MinIO, we won't able to support it anymore in v0.2.0.

vtolstov:
Why minio remived?

yasker:
@vtolstov Since MinIO is one more thing to maintain. We need to start MinIO in HA mode, and operate it. Also we need to support the component by ourselves if it's a part of stack. We once had an issue with MinIO lost service temporarily during certain circumstances, and have to rely on the upstream MinIO to fix it after a month.

With MinIO only used as an image store, and now Longhorn is able to store the information about the image (by using the backing image feature), we want to replace MinIO's image repo functionality with Longhorn itself.

vtolstov:
Nice, thank you for explaining

penghuima:
Before using the V0.1 version to create virtual machines, I was curious why users could not use local virtual machine images to create virtual machines, and there are two storage systems, MinIO, and longhorn, in the architecture of Harvester. MinIO, which is only used to store images. At that time, I wondered if longhorn could not be used to store images instead of MinIO.  I did not guess that MinIO would be banned in v0.2, which was shocked, but not so surprising.

guangbochen:
this will depends on https://github.com/longhorn/longhorn/issues/2404

ArtiePeterson:
> We want to support users to upload images from the local desktop to the Harvester cluster.

Awesome..

noahgildersleeve:
Tested in master-1a041d23-head. Validated with ISO and live image.

xeruf:
This should be adjusted: https://harvesterhci.io/import-image/

PR #1109, opened 2021-08-18
Support uploading images

gitlawr:
https://github.com/harvester/harvester/issues/570

Add upload support with Longhorn backing image enhancement. Adapt to the backing image API changes.
Add a required field `spec.sourceType` in vmimage, need to coordinate with UI and E2E tests for this API change.


pkg/data/crd.go
```diff
@@ -39,6 +39,10 @@ func createCRDs(ctx context.Context, restConfig *rest.Config) error {
 			crd.FromGV(harvesterv1.SchemeGroupVersion, "VirtualMachineRestore", harvesterv1.VirtualMachineRestore{}),
 			crd.FromGV(harvesterv1.SchemeGroupVersion, "Preference", harvesterv1.Preference{}),
 			crd.FromGV(harvesterv1.SchemeGroupVersion, "SupportBundle", harvesterv1.SupportBundle{}),
+			// The BackingImage struct is not compatible with wrangler schemas generation, pass nil as the workaround.
+			// The expected CRD will be applied by Longhorn chart.
```
guangbochen, line 42: can u please file an issue on the LH side to check if this is something they will need to make it compatible or the wrangler is better to support it?

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	return image, nil
+}
+
+func (h *vmImageHandler) initialize(image *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
+	if err := h.createBackingImage(image); err != nil && !errors.IsAlreadyExists(err) {
+		return nil, err
+	}
+	if err := h.createStorageClass(image); err != nil && !errors.IsAlreadyExists(err) {
+		return nil, err
+	}
```
guangbochen, line 76: can we add a proper error message to the failed initialization?

pkg/controller/master/image/backing_image_controller.go
```diff
@@ -0,0 +1,73 @@
...
+	backingImageCache lhv1beta1.BackingImageCache
+}
+
+func (h *backingImageHandler) OnChanged(_ string, backingImage *v1beta1.BackingImage) (*v1beta1.BackingImage, error) {
+	if backingImage == nil || backingImage.DeletionTimestamp != nil {
+		return nil, nil
+	}
+	if backingImage.Spec.SourceType != types.BackingImageDataSourceTypeUpload || backingImage.Annotations[util.AnnotationImageID] == "" {
+		return nil, nil
+	}
```
guangbochen: since the LH backing image also tracked its progress for URL data source type, please add to sync its status instead of making it an instant complete as before.

gitlawr: It is not a 1-1 mapping, LH backing image records progress for multiple disk files. The download progress varies on different nodes.

pkg/data/crd.go
gitlawr: I think this is something to be improved in wrangler

pkg/controller/master/image/backing_image_controller.go
guangbochen: understand, as long as we can use the first complete node as the default tracking progress, it will make the status look much solid when compared with the instant completion, WDUT.

pkg/controller/master/image/vm_image_controller.go
gitlawr: The controller is expected to reconcile with these errors so we don't set initialization as failed here.

pkg/controller/master/image/backing_image_controller.go
gitlawr: Updated to sync progress for download type. Changed the Uploaded condition to Imported to fit both types.

guangbochen (approved):
LGTM, @gitlawr does the UI need to change anything in regards to the last modification?

futuretea (approved):
LGTM

Issue #1114, opened 2021-08-24
Task: Switch support bundle to use the new `support-bundle-kit`


yasker:
https://github.com/rancher/support-bundle-kit

guangbochen:
related to the newOS change https://github.com/harvester/harvester-installer/pull/93, will need to make sure the new support bundle kit works with RKE2 using journals, additionally would prefer to make the supported k8s provider configurable e.g like `k8s-provider: rke2/k3s/etc.`

guangbochen:
close in favor of `master-48c97ea7-head`, the support-bundle was updated to `rancher/support-bundle-kit:v0.0.3`
```
ks get settings support-bundle-image -oyaml
apiVersion: harvesterhci.io/v1beta1
default: rancher/support-bundle-kit:v0.0.3
kind: Setting
metadata:
  creationTimestamp: "2021-09-26T04:30:37Z"
  generation: 1
  name: support-bundle-image
  resourceVersion: "5354"
  uid: ce67f380-e2f8-4015-b169-9f8c2ee7646f
status: {}
```

PR #1146, opened 2021-09-02
Fix support bundle prefix url path

guangbochen:

**Problem:**
currently only URL with `/v1/harvester` prefix-path` will be served the rancher aggregated apiserver 

**Solution:**
Fix support bundle prefix url path

**Related Issue:**
https://github.com/harvester/harvester/issues/1114

**Test plan:**


gitlawr (approved):
LGTM

Issue #1139, opened 2021-08-31
[BUG]Rancher UI on first screen when Installing Harvester using IPXE Boot


chajain:
**Describe the bug**
When trying to install Harvester using ipxe Boot, the first screen which pops up on first login is Rancher UI

**To Reproduce**
1. Install Harvester using instructions on https://github.com/harvester/ipxe-examples/tree/main/vagrant-pxe-harvester

**Expected behavior**
Harvester Ui Login page should appear

**Support bundle**

**Environment:**
 - Harvester ISO version: Master
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): IPXE Harvester Boot

**Additional context**

lucidd:
Not sure if what i'm seeing is different from this bug but i wanted to put it here just in case. I'm not using IPXE but after the first login i get to `/dashboard/home` which shows the "welcome to rancher" dashboard. After some more time (~5 minutes for me) on the same url after refreshing the message changes to "welcome to harvester" and navigating to / opens the harvester ui. I also noticed that on the terminal the harvester-installer shows `Ready` while its still serving the rancher ui.

yasker:
We shouldn't show the Rancher UI at all from Harvester.

noahgildersleeve:
I tested this and the rancher login prompt happens while harvester is still doing the post install setting up step. Once it goes to ready it will show the harvester prompt. I imagine that when run locally with less resources the step takes longer so it is easier to hit that fail condition. We should probably just disable the web UI while it's not ready. That would probably resolve the issue.

guangbochen:
that's exactly the reason why some rancher elements are showing on the single Harvester dashboard. On the console, we should already display the server as `NotReady` when the Harvester server is not running, but as long as the Rancher server is ready and users are visiting with the correct URL we can't disable the web UI.

And the server only needs to config it once at the initial stage.

yasker:
@guangbochen can we find a way to disable Rancher server UI temporarily until the Harvester setup is done? We might need to reach out to the Rancher team to see if there is a way.

guangbochen:
closed in favor of `master-48c97ea7-head`, currently only the Harvester UI and name was pop up.

Issue #1133, opened 2021-08-30
[BUG] Rancher enabled settings does not enable the Rancher button in the header


noahgildersleeve:
**Describe the bug**

When you enable rancher in the advanced settings the UI element doesn't show
**To Reproduce**
1. Navigate to advanced settings
2. Set rancher-enabled to `true`
3. Check header

**Expected behavior**

Either we need to fix the UI issue, or we need to remove the setting in advanced settings.
**Support bundle**

**Environment:**
 - Harvester ISO version: master-d25964a2-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): virtualized Harvester in hyper-v

**Additional context**
![Greenshot 2021-08-30 11 13 24](https://user-images.githubusercontent.com/83787952/131385386-73d12f0b-9843-465d-b6f6-17543aa0faab.png)

yasker:
Embedded Rancher will be for debugging purposes, at least we need to change the wording of the setting, or remove it but make it accessible through another way (e.g. hidden URL).

guangbochen:
this setting will be removed eventually.

guangbochen:
close as the rancher setting has been removed. verified in the `master-48c97ea7-head`

PR #1152, opened 2021-09-03
Add local-ui-dev mode and fix rancher ui expose registry

guangbochen:
**Problem:**
- [BUG] Rancher enabled settings does not enable the Rancher button in the header #1133 
- [BUG]Rancher UI on first screen when Installing Harvester using IPXE Boot #1139 

**Solution:**
- remove deprecated settings 
- add local UI dev mode with set `--add-local-ui=true` to enable local UI for local development
- add rancher-expose svc after patched rancher settings on the Harvester server.

**Related Issue:**
https://github.com/harvester/harvester/issues/1133
https://github.com/harvester/harvester/issues/1139

**Test plan:**


pkg/server/router.go
```diff
@@ -75,27 +74,30 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
...
-		if err != nil {
-			logrus.Fatal(err)
-		}
-		rancherHandler := &proxy.Handler{
-			Host:   host,
-			Scheme: scheme,
+	// enable local dashboard ui for dev
+	if r.options.AddLocalUI {
+		loginHandler := auth.NewLoginHandler()
+		m.Path("/v1-public/auth").Handler(loginHandler)
```
gitlawr: I thought we removed the auth part. Do we still need this?

main.go
```diff
@@ -76,6 +76,12 @@ func main() {
 			Usage:       "Specify the URL to connect to the Rancher server",
 			Destination: &options.RancherURL,
 		},
+		cli.BoolFlag{
+			Name:        "add-local-ui",
+			Usage:       "Add local dashboard UI(true, false)",
```
bk201:
```suggestion
			Usage:       "Add local dashboard UI (true, false)",
```

nit.

guangbochen:
Updated, can u guys PTAL thanks.

Additionally add fix of invalid feature gates config, previous the feature gates of Kubevirt only contains the `LiveMigration`, but `HotplugVolumes` is also required.
```helm template deploy/charts/harvester | grep "KubeVirt" --context=30```

**Review by gitlawr — APPROVED** [review]
LGTM

File: pkg/server/router.go [hunk 3]
```diff
@@ -75,27 +73,27 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
...
-			Host:   host,
-			Scheme: scheme,
+	// enable local dashboard ui for dev
+	if r.options.AddLocalUI {
+		vueUI := ui.Vue
+		m.Handle("/dashboard/", vueUI.IndexFile())
+		m.PathPrefix("/dashboard/").Handler(vueUI.IndexFileOnNotFound())
+		m.PathPrefix("/api-ui").Handler(vueUI.ServeAsset())
+
+		if r.options.AddLocalUI || r.options.RancherURL != "" {
```
bk201: Everything is good except the first condition is not needed.

bk201 (approved):

pkg/server/router.go
guangbochen: fixed, thanks

PR #1153, opened 2021-09-03
Use scale-available annotation for harvester services auto-scaling

gitlawr:
With this solution, there won't be pending pods if the number of nodes is less than 3.


guangbochen:
need to address the CI failure.

gitlawr:
Fixed. The annotation does not support 0 replicas which is used in the integration tests. Fall back to spec.replicas in this case to comply with the integration test setup.

guangbochen (approved):
LGTM

Issue #632, opened 2021-04-01
[FEATURE] Built-in HA load balancer


yasker:
We want to have a built-in HA load balancer shipped with the solution, so we can provide VIP(virtual IP address) feature of the Harvester cluster.

Additionally, it can act as a cloud provider in Kubernetes with https://github.com/harvester/harvester/issues/633 enhancement.

We can use kube-vip(https://kube-vip.io/) potentially.

larssb:
Allow me to comment on this one. I would like you to consider https://purelb.gitlab.io/docs/ ... especially as it has support for the `OSPF` routing protocol. `BGP` is, I think, outdated and well known to be insecure. PureLB evolved from, started as far as I know, as a fork of MetalLB. And it's implementation is quite nice. It uses the Bird Router for routing so a nice separation of concerns was done in regards to what PureLB do and do not do.

Please give it a look. Thank you.

yaocw2020:
https://purelb.gitlab.io/docs/how_it_works/bundledbird/

olljanat:
Afaiu this was implemented by #1155 but not closed automatically because description misses magic Word "Closes"

yasker:
@olljanat Thanks for the reminder. Before we can close an issue, our QA will need to validate the issue has been indeed fixed. So that's why we don't use `close` keyword in the PRs.

lanfon72:
Verified the feature implemented on *v0.3.0-rc1* and newer.

[bare metal v0.3.0-rc1]
![image](https://user-images.githubusercontent.com/5169694/135636816-73418b42-f52c-4786-9245-247f69003a47.png)
[qemu/KVM master-82506f55-head]
![image](https://user-images.githubusercontent.com/5169694/135637398-d3f7fda3-75e1-46f1-b06a-7d8d0321ce45.png)

Test Information
----
* Environment: **Bare mental DL360** 3 nodes, **qemu/KVM** 5 nodes
* Harvester Version: **v0.3.0-rc1**, **master-82506f55-head**
### Verify Steps:
  1. install harvester with 2 secondary nodes
  2. open dashboard with VIP
  3. shutdown or reboot any node
  4. check the dashboard still working and the node shouldn't be marked as `active`

PR #1155, opened 2021-09-04
Support VIP

yaocw2020:
Provide a VIP for the Harvester.

**Related Issue:**
https://github.com/harvester/harvester/issues/632


deploy/charts/harvester/templates/service.yaml
```diff
@@ -6,6 +6,13 @@ metadata:
   # name: {{ template "harvester.fullname" . }}
   # because we can easily confirm this resource from the corresponding namespace.
   name: harvester
+  annotations:
+{{- if and (eq .Values.service.harvester.type "LoadBalancer") (eq .Values.service.harvester.vipMode "DHCP") }}
+{{- if and (ne .Values.service.harvester.vip "") (ne .Values.service.harvester.vipHwAddr "") }}
```
gitlawr: a non-empty string is evaluated to be true. So we don't need the ne checks.

deploy/charts/harvester/values.yaml
```diff
@@ -221,7 +220,7 @@ service:
     ## Specify the port of HTTPs endpoint.
     ## defaults to "8443".
     ##
-    httpsPort: 8443
+    httpsPort: 443
```
gitlawr, line 223: Have we changed the port?

yaocw2020: Yes. The default port of HTTPS is 443.

gitlawr: Also, update the default value in the comment

deploy/charts/harvester/values.yaml
```diff
@@ -377,3 +387,8 @@ webhook:
     tag: master-head
 
   httpsPort: 9443
+
+kube-vip:
+  enabled: true
+  config:
+      vip_interface: eth0
```
bk201: Does this mean VIP is always on `eth0`?

yaocw2020: It's the default value. Once users enable the VLAN network and use eth0 as the VLAN NIC, eth0 is unavailable for the VIP interface and we will switch the VIP interface as `harvester-br0`.

gitlawr: I don't see we override it in installation. Should we set it to be the management NIC?

gitlawr (approved):
LGTM

deploy/charts/harvester/values.yaml
```diff
@@ -377,3 +387,8 @@ webhook:
     tag: master-head
 
   httpsPort: 9443
+
+kube-vip:
```
guangbochen, line 414: will need to add the image of the kube-vip, kube-vip-cloud-provider, and harvester-load-balancer respectively.

guangbochen (approved):
LGTM

bk201 (approved):
I can confirm the images are also pre-loaded in the installer side.

PR #1157, opened 2021-09-06
Adapt support-bundle-kit changes

bk201:
**Problem:**
Support-bundle-utils is renamed to support-bundle-kit

**Solution:**
Adapt the change.

Here is a generated bundle for reference: 
[supportbundle_c9eafbd9-6984-463f-b34c-af3ca8bdb085_2021-09-06T03-02-06Z.zip](https://github.com/harvester/harvester/files/7113060/supportbundle_c9eafbd9-6984-463f-b34c-af3ca8bdb085_2021-09-06T03-02-06Z.zip)

**Related Issue:**
https://github.com/harvester/harvester/issues/1114

**Test plan:**
- Click `Support` on the bottom-left corner of Harvester GUI.
- Click `Generate Support Bundle`

### DIFF
--- a/pkg/controller/master/supportbundle/manager.go
+++ b/pkg/controller/master/supportbundle/manager.go
@@ -43,6 +43,7 @@ func (m *Manager) Create(sb *harvesterv1.SupportBundle, image string) error {
 	logrus.Debugf("creating deployment %s with image %s", deployName, image)
 
 	pullPolicy := m.getImagePullPolicy()
+	namespaces := []string{sb.Namespace, "longhorn-system"}
 
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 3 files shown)


guangbochen (approved):
LGTM

pkg/controller/master/supportbundle/manager.go
```diff
@@ -43,6 +43,7 @@ func (m *Manager) Create(sb *harvesterv1.SupportBundle, image string) error {
 	logrus.Debugf("creating deployment %s with image %s", deployName, image)
 
 	pullPolicy := m.getImagePullPolicy()
+	namespaces := []string{sb.Namespace, "longhorn-system", "default"}
```
gitlawr: The default namespace is not special now. Why do we want to set it here?

bk201: Removed.
But anyway it's the [default namespaces](https://github.com/rancher/support-bundle-kit/blob/189e740444790dc3a3c7f96a49b4ae9ca2db181c/pkg/manager/cluster.go#L86) hard-coded in SBK for now.

gitlawr: Does SBKit collect resources/logs from these namespaces? If that's the case we might also need kube-system and cattle-system

bk201: Two namespaces lists will be merged. By default, SBKit collects common namespaces in Rancher.

gitlawr (approved):
LGTM

PR #1163, opened 2021-09-07
Remove bundle package in CI build and fix build-iso

gitlawr:
Remove the bundle image build which is for previous upgrade implementation.
Update build-iso step for the new installer changes


guangbochen (approved):
LGTM

bk201 (approved):

PR #1169, opened 2021-09-08
Add authed roles

gitlawr:
Make harvester settings readable to authed users
Use apply.Apply to create built-in resources so that they get updated in furture changes

### DIFF
--- a/pkg/data/add.go
+++ b/pkg/data/add.go
@@ -11,15 +11,17 @@ func Init(ctx context.Context, mgmtCtx *config.Management, options config.Option
 	if err := createCRDs(ctx, mgmtCtx.RestConfig); err != nil {
 		return err
 	}
-	if err := createPublicNamespace(mgmtCtx); err != nil {
+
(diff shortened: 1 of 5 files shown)


pkg/data/add.go
```diff
@@ -11,15 +11,17 @@ func Init(ctx context.Context, mgmtCtx *config.Management, options config.Option
...
+
+	if err := addAPIService(mgmtCtx.Apply, options.Namespace); err != nil {
 		return err
 	}
-	if err := createTemplates(mgmtCtx, publicNamespace); err != nil {
+	if err := addAuthenticatedRoles(mgmtCtx.Apply); err != nil {
 		return err
 	}
-	if err := createAPIService(mgmtCtx, options.Namespace); err != nil {
+	if err := addPublicNamespace(mgmtCtx.Apply); err != nil {
```
guangbochen: move this up since it is need of addAuthenticatedRoles

guangbochen (approved):
LGTM

Issue #1174, opened 2021-09-08
[BUG] Nodes can't join a cluster


bk201:
**Describe the bug**
Nodes can't join clusters.

**To Reproduce**
1. Start the first node and create a Harvester cluster.
2. Start another node and join the cluster.
3. The second node doesn't appear in the Kubernetes clusters. (`kubectl get nodes`)

**Expected behavior**
The second node should appear in the Kubernetes clusters. (`kubectl get nodes`)

**Support bundle**

**Environment:**
 - Harvester ISO version: master ISO Sep. 8th. (https://github.com/harvester/harvester-installer/commit/7967202c8a8c9b2f583ee9b55dba03e4c955477e)
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): VMs

**Additional context**

rancher-system-agent logs on second nodes:
```
Sep 08 10:53:07 node2 rancher-system-agent[2327]: time="2021-09-08T10:53:07Z" level=debug msg="[K8s] Processing secret custom-5ec24e11e79d-machine-plan in namespace fleet-local at generation 0 with resource version 3922"
Sep 08 10:53:12 node2 rancher-system-agent[2327]: time="2021-09-08T10:53:12Z" level=debug msg="[K8s] Processing secret custom-5ec24e11e79d-machine-plan in namespace fleet-local at generation 0 with resource version 3922"
Sep 08 10:53:17 node2 rancher-system-agent[2327]: time="2021-09-08T10:53:17Z" level=debug msg="[K8s] Processing secret custom-5ec24e11e79d-machine-plan in namespace fleet-local at generation 0 with resource version 3922"
Sep 08 10:53:22 node2 rancher-system-agent[2327]: time="2021-09-08T10:53:22Z" level=debug msg="[K8s] Processing secret custom-5ec24e11e79d-machine-plan in namespace fleet-local at generation 0 with resource version 3922"
Sep 08 10:53:27 node2 rancher-system-agent[2327]: time="2021-09-08T10:53:27Z" level=debug msg="[K8s] Processing secret custom-5ec24e11e79d-machine-plan in namespace fleet-local at generation 0 with resource version 3922"
Sep 08 10:53:32 node2 rancher-system-agent[2327]: time="2021-09-08T10:53:32Z" level=debug msg="[K8s] Processing secret custom-5ec24e11e79d-machine-plan in namespace fleet-local at generation 0 with resource version 3922"
[...]

PR #1175, opened 2021-09-08
Enable RKE2 cloud controller on promoted nodes

bk201:
**Problem:**
Fix the problem that nodes can't join a cluster.

**Solution:**
Enable RKE2 cloud controller on promoted nodes

**Related Issue:**
https://github.com/harvester/harvester/issues/1174

**Test plan:**


guangbochen (approved):
LGTM

PR #1217, opened 2021-09-14
Promote nodes by setting machine labels

bk201:
Switch to the[ preferred way](https://github.com/rancher/system-agent-installer-rke2/pull/13) to promote agent to server in Rancher
by setting the following labels on corresponding machines:

```
rke.cattle.io/control-plane-role: "true"
rke.cattle.io/etcd-role: "true"
```

This PR also fixes the VIP config is not set in promoted nodes:
```
tls-san:
  - $VIP
```


gitlawr (approved):
LGTM

Issue #1193, opened 2021-09-09
[BUG] UI websocket disconnect error


guangbochen:
**Describe the bug**

When visiting the Harvester UI via the VIP, at times the WebSocket will be disconnected.

**Expected behavior**
Websocket should not be disconnected regularly.

as a temporary workaround is to click re-enable the cert 
![WechatIMG18](https://user-images.githubusercontent.com/4569037/132688488-f77457f2-8f93-4c6a-b6c6-4e48306234d1.png)

**Support bundle**

**Environment:**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

noahgildersleeve:
Tested in master-8afedf0a-head. During testing I haven't seen any websocket disconnected errors. I will keep an eye out as testing proceeds and will re-open issue if it comes up. Closing as fixed.

Issue #1131, opened 2021-08-30
[BUG] Use the downloaded kubeconfig to connect to the cluster, display the certificate error


WuJun2016:
**Describe the bug**

Use the downloaded kubeconfig to connect to the cluster, display the certificate error

**To Reproduce**
1. Download the kubeconfig file through the header menu
2. Use this configuration to connect to the cluster

![image](https://user-images.githubusercontent.com/24985926/131312315-5f2f4775-1bff-4907-b09e-bb0d17a118f2.png)

**Expected behavior**

Can connect to the cluster normally

**Support bundle**

**Environment:**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

noahgildersleeve:
Tested in master-8afedf0a-head. fixed.

PR #1223, opened 2021-09-15
update tls

gitlawr:
https://github.com/harvester/harvester/issues/1131
https://github.com/harvester/harvester/issues/1193

Sync cacerts setting;
Add VIP to certificate SANs;
Add 80 port to the expose service so that http://vip is accessible(redirected to https://).

### DIFF
--- a/pkg/controller/master/rancher/embedded_rancher.go
+++ b/pkg/controller/master/rancher/embedded_rancher.go
@@ -1,6 +1,8 @@
 package rancher
 
 import (
+	"strings"
+
 	rancherv3api "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/sirupsen/logrus"
 
@@ -18,6 +20,10 @@ func (h *Handler) RancherSettingOnChange(key string, setting *rancherv3api.Setti
 		return nil, nil
 	}
 
+	if setting.Name == internalCACertsSetting && setting.Value != "" {
+		return nil, h.initializeTLS(setting)
+	}
+
 	for name, value := range UpdateRancherUISettings {
 		if setting.Name == name && setting.Default != value {
 			logrus.Debugf("Updating rancher dashboard setting %s, %s => %s", name, setting.Default, value)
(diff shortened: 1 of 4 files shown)


pkg/controller/master/rancher/embedded_rancher.go
```diff
@@ -18,6 +20,10 @@ func (h *Handler) RancherSettingOnChange(key string, setting *rancherv3api.Setti
 		return nil, nil
 	}
 
+	if setting.Name == "internal-cacerts" && setting.Value != "" {
```
guangbochen: let's move the `internal-cacerts` name to the constant at `register.go` for better consistency.

pkg/controller/master/rancher/embedded_rancher.go
```diff
@@ -30,3 +36,67 @@ func (h *Handler) RancherSettingOnChange(key string, setting *rancherv3api.Setti
...
+// initializeTLS writes internal-cacerts to cacerts value and adds the VIP to certificate SANs
+// cacerts is used in generated kubeconfig files
+func (h *Handler) initializeTLS(setting *rancherv3api.Setting) error {
+	cacerts, err := h.RancherSettingCache.Get("cacerts")
+	if err != nil {
+		return err
+	}
+	if cacerts.Value != "" {
+		return nil
+	}
```
guangbochen, line 49: seems duplicate with `h.setCACerts(setting.Value)`

pkg/controller/master/rancher/embedded_rancher.go
```diff
@@ -30,3 +36,67 @@ func (h *Handler) RancherSettingOnChange(key string, setting *rancherv3api.Setti
...
+	return nil
+}
+
+func (h *Handler) setCACerts(cert string) error {
+	cacerts, err := h.RancherSettingCache.Get("cacerts")
+	if err != nil {
+		return err
+	}
+	if cacerts.Value == cert {
```
guangbochen, line 66: should this be `if cacerts.Value == cert || cacerts.Value != ""`

gitlawr: The `cacerts.Value != ""` check is used to skip the update when it is already initialized. But for a function like `setCACerts(cert)`, it doesn't have to care about that.

gitlawr:
updated

guangbochen (approved):
LGTM

Issue #1019, opened 2021-07-14
[FEATURE] Harvester CCM & Guest CSI Automatic Integration with RKE2 Provisoner


yaocw2020:
**Is your feature request related to a problem? Please describe.**
Both Harvester CCM & Guest CSI driver should install and configure by default when the user spins up a k8s cluster with the Harvester RKE2 node driver.

dependes on https://github.com/rancher/rancher/issues/33611

**Describe the solution you'd like**
- [x] RKE2 Chart integration
    - [x] Add Harvester CCM chart @yaocw2020 
    - [x] Add Guest CSI driver chart @gitlawr 
 - [x] RKE2 integration @gitlawr 
    - Add Harvester as a `cloud-provider` to the RKE2 [kontainer driver metadata](https://github.com/rancher/kontainer-driver-metadata/blob/78a6b22a0d4bc52fef5d57b707e92c619efe94c3/channels-rke2.yaml#L84) and update the RKE2 [dockerfile integration](https://github.com/rancher/rke2/pull/924/files)
 - [x] Harvester API @yaocw2020 
    - Support generating `kubeconfig` file for the guest k8s cluster based on cluster name and VM namespace
- [x] UI integration: @n313893254 
    - set the default cloud provider to the `harvester`
    - call the imported cluster API to get the generated `kubeconfig` file and set it to the `cloud_provider_config`

Notes: For the external harvester cluster, the user has to copy & paste the `kubeconfig` file manually to the `Cloud Provider Config` in order to use the CCM and guest CSI features.

**Additional context**
https://github.com/harvester/cloud-provider-harvester

yaocw2020:
dependency: https://github.com/harvester/harvester/pull/1061

yasker:
Imported Harvester v0.3.0 cluster with RKE2 built-in cloud provider support has been validated in https://github.com/rancher/dashboard/issues/4208. Guest CSI driver will be validated in https://github.com/harvester/harvester/issues/580 . Close the issue as done.

PR #1224, opened 2021-09-15
Get server URL from vip configmap

yaocw2020:
Get server URL from vip configmap

issue: https://github.com/harvester/harvester/issues/1019

### DIFF
--- a/pkg/api/kubeconfig/generate_handler.go
+++ b/pkg/api/kubeconfig/generate_handler.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
 	"strings"
 	"time"
 
+	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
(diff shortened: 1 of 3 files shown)


guangbochen (approved):
LGTM

guangbochen:
@yaocw2020 can we add server URL address validation, thanks

pkg/api/kubeconfig/generate_handler.go
```diff
@@ -36,9 +41,11 @@ type GenerateHandler struct {
 	roleBindingCache  ctlrbacv1.RoleBindingCache
 	secretCache       ctlcorev1.SecretCache
 	secretClient      ctlcorev1.SecretClient
+	configMapCache    ctlcorev1.ConfigMapCache
+	namespace         string
 }
 
-func NewGenerateHandler(scaled *config.Scaled) *GenerateHandler {
+func NewGenerateHandler(scaled *config.Scaled, option *config.Options) *GenerateHandler {
```
guangbochen: don't need to pass the option as pointer, pass value is sufficient enough

gitlawr:
Why don't we use the kubernetes core API?

yaocw2020:
> Why don't we use the kubernetes core API?

Agree, I have deleted the deleteKubeconfig API.

gitlawr (approved):
LGTM

Issue #568, opened 2021-03-25
[FEATURE] Create image from volume


yasker:
It's common for the user to modify the existing VM content, then want to use the updated volume content as an image for the new VM.

We need to support create an image from a volume, for both root image and non-root image.

Depends on: https://github.com/longhorn/longhorn/issues/2403

futuretea:
If this is implemented, then a packer provider can also be implemented to automate the entire process of creating images.

WuJun2016:
Done

gitlawr:
Some UI issues to be fixed:
1. When the volume export is in progress, the image status shows `Uploading`.
2. Edit an image created from a volume, it shows `Upload Files`
![image](https://user-images.githubusercontent.com/5697937/134793965-64d3f211-a69d-4a89-a6eb-da8019a01287.png)

guangbochen:
Few Typo and enhancement:
- Export Image => Export to Image
- system namespaces are not filtered and should use the user's current selected default namespace
- swap the order of the namespace first and name, also why the margin is quite big and the height is quite different between those two fields?
![image](https://user-images.githubusercontent.com/4569037/134799547-36d74f31-d8ed-40c1-963d-f856ea6e395d.png)

gitlawr:
![image](https://user-images.githubusercontent.com/5697937/134801500-9f173338-0eb6-44a0-bd0b-6e45fd510873.png)
The error happens when exporting an image in a different namespace. @shangma

shangma:
### To Test:

1. Download the latest harvester [ISO](https://releases.rancher.com/harvester/master/harvester-amd64.iso) and boot a virtual machine from it
2. Log into Harvester
3. Go to "Images" section and create an original os image from an url, say [this one](https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-amd64-disk1.img)
4. Create a virtual machine from the image above with proper config. A good example can be found [here](https://docs.harvesterhci.io/v0.2/vm/create-vm/#cloud-config-examples).
5. Once the vm is created, log into it and make some changes. The default username for the image above is "ubuntu" and the password can be configure from step 3
6. Go to "Volumes" section and select the volume for the virtual machine above. And then click on the three dots on the far right and choose "Export Image"
![image](https://user-images.githubusercontent.com/2396545/134824988-35ba8f37-5ae2-4dc8-93c7-95e0c51ae93f.png)

[...]

PR #1226, opened 2021-09-15
create backingimage from harvester volume

shangma:
[havester issue #568](https://github.com/harvester/harvester/issues/568)

### DIFF
--- a/pkg/api/vm/types.go
+++ b/pkg/api/vm/types.go
@@ -37,3 +37,8 @@ type AddVolumeInput struct {
 type RemoveVolumeInput struct {
 	DiskName string `json:"diskName"`
 }
+
+type ExportVolumeInput struct {
+	DisplayName string `json:"displayName"`
+	Namespace   string `json:"namespace"`
+}
--- a/pkg/api/volume/formatter.go
+++ b/pkg/api/volume/formatter.go
@@ -1,0 +1,90 @@
+package volume
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/rancher/apiserver/pkg/apierror"
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/rancher/wrangler/pkg/schemas/validation"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/harvester/harvester/pkg/api/vm"
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+)
+
+const (
+	actionExport = "export"
+)
+
+func Formatter(request *types.APIRequest, resource *types.RawResource) {
+	resource.Actions = make(map[string]string, 1)
+	resource.AddAction(request, actionExport)
+}
+
+type ExportActionHandler struct {
+	images v1beta1.VirtualMachineImageClient
+}
+
+func (h ExportActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	if err := h.do(rw, req); err != nil {
+		status := http.StatusInternalServerError
+		if e, ok := err.(*apierror.APIError); ok {
+			status = e.Code.Status
+		}
+		rw.WriteHeader(status)
+		_, _ = rw.Write([]byte(err.Error()))
+		return
+	}
+	rw.WriteHeader(http.StatusOK)
+	_, _ = rw.Write([]byte("Successfully export the volume."))
+}
+
+func (h ExportActionHandler) do(rw http.ResponseWriter, r *http.Request) error {
+	vars := mux.Vars(r)
+	action := vars["action"]
+	pvcName := vars["name"]
+	imageNamespace := vars["namespace"]
+
+	switch action {
+	case actionExport:
+		var input vm.ExportVolumeInput
+		if err := json.NewDecoder(r.Body).Decode(&input); err != nil {
+			return apierror.NewAPIError(validation.InvalidBodyContent, "Failed to decode request body: %v "+err.Error())
+		}
+		if input.DisplayName == "" {
+			return apierror.NewAPIError(validation.InvalidBodyContent, "Parameter `displayName` is required")
+		}
+		if input.Namespace == "" {
+			return apierror.NewAPIError(validation.InvalidBodyContent, "Parameter `namespace` is required")
+		}
+		return h.exportVolume(r.Context(), input.Namespace, pvcName, input.DisplayName, imageNamespace)
+	default:
+		return apierror.NewAPIError(validation.InvalidAction, "Unsupported action")
+	}
+}
+
+func (h ExportActionHandler) exportVolume(ctx context.Context, pvcNamespace, pvcName, imageDisplayName, imageNamespace string) error {
+	vmImage := &harvesterv1.VirtualMachineImage{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "image-",
+			Namespace:    imageNamespace,
+		},
+		Spec: harvesterv1.VirtualMachineImageSpec{
+			DisplayName:  imageDisplayName,
+			SourceType:   harvesterv1.VirtualMachineImageSourceTypeExportVolume,
+			PVCName:      pvcName,
+			PVCNamespace: pvcNamespace,
+		},
+	}
+	_, err := h.images.Create(vmImage)
+	if err != nil {
+		logrus.Errorf("Unable to create image from volume %s", pvcName)
+		return err
+	}
+	return nil
+}
--- a/pkg/api/volume/schema.go
+++ b/pkg/api/volume/schema.go
@@ -1,0 +1,35 @@
+package volume
+
+import (
+	"net/http"
+
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/rancher/steve/pkg/schema"
+	"github.com/rancher/steve/pkg/server"
+	"github.com/rancher/wrangler/pkg/schemas"
+
+	"github.com/harvester/harvester/pkg/api/vm"
+	"github.com/harvester/harvester/pkg/config"
+)
+
+func RegisterSchema(scaled *config.Scaled, server *server.Server, options config.Options) error {
+	server.BaseSchemas.MustImportAndCustomize(vm.ExportVolumeInput{}, nil)
+	t := schema.Template{
+		ID: "persistentvolumeclaim",
+		Customize: func(s *types.APISchema) {
+			s.Formatter = Formatter
+			s.ResourceActions = map[string]schemas.Action{
+				actionExport: {
+					Input: "exportVolumeInput",
+				},
+			}
+			s.ActionHandlers = map[string]http.Handler{
+				actionExport: ExportActionHandler{
+					images: scaled.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineImage(),
+				},
+			}
+		},
+	}
+	server.SchemaFactory.AddTemplate(t)
+	return nil
+}
--- a/pkg/apis/harvesterhci.io/v1beta1/image.go
+++ b/pkg/apis/harvesterhci.io/v1beta1/image.go
@@ -12,8 +12,9 @@ var (
 )
 
 const (
-	VirtualMachineImageSourceTypeDownload = "download"
-	VirtualMachineImageSourceTypeUpload   = "upload"
+	VirtualMachineImageSourceTypeDownload     = "download"
+	VirtualMachineImageSourceTypeUpload       = "upload"
+	VirtualMachineImageSourceTypeExportVolume = "export-from-volume"
 )
 
 // +genclient
@@ -39,8 +40,14 @@ type VirtualMachineImageSpec struct {
 	DisplayName string `json:"displayName"`
 
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Enum=download;upload
+	// +kubebuilder:validation:Enum=download;upload;export-from-volume
 	SourceType string `json:"sourceType"`
+
+	// +optional
+	PVCName string `json:"pvcName"`
+
+	// +optional
+	PVCNamespace string `json:"pvcNamespace"`
 
 	// +optional
 	URL string `json:"url"`
--- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
 	"github.com/longhorn/longhorn-manager/types"
+	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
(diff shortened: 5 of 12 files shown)


pkg/api/volume/formatter.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+}
+
+func (h ExportActionHandler) exportVolume(ctx context.Context, namespace, displayName, pvcName string) error {
+	// We only permit volume source from existing PersistentVolumeClaim at this moment.
+	// KubeVirt won't check PVC existence so we validate it on our own.
+	pvc, err := h.pvcCache.Get(namespace, pvcName)
+	if err != nil {
+		return fmt.Errorf("failed to get pvc %s/%s, error: %s", pvcName, namespace, err.Error())
+	}
```
gitlawr: Move this to the validation webhook(pkg/webhook/resources/virtualmachineimage/validator.go)

pkg/api/volume/formatter.go
```diff
@@ -0,0 +1,83 @@
...
+	vmi := &harvesterv1.VirtualMachineImage{
+		Spec: harvesterv1.VirtualMachineImageSpec{
+			DisplayName: displayName,
+			SourceType:  harvesterv1.VirtualMachineImageSourceTypeExportVolume,
+			VolumeName:  pvc.Spec.VolumeName,
+		},
+	}
+	_, err = h.images.Create(vmi)
+	if err != nil {
+		logrus.Errorf("Unable to create image from volume %s", pvc.Spec.VolumeName)
```
gitlawr: return the error

pkg/api/vm/types.go
```diff
@@ -37,3 +37,7 @@ type AddVolumeInput struct {
 type RemoveVolumeInput struct {
 	DiskName string `json:"diskName"`
 }
+
+type ExportVolumeInput struct {
```
gitlawr, line 41: This is unused. Set it to be the action input in the schema and parse it in the handler. Check `backupInput` as the example

pkg/api/volume/formatter.go
```diff
@@ -0,0 +1,83 @@
...
+
+const (
+	actionExport = "export"
+)
+
+func Formatter(request *types.APIRequest, resource *types.RawResource) {
+	resource.Actions = make(map[string]string, 1)
+	if resource.APIObject.Data().String("spec", "sourceType") == harvesterv1.VirtualMachineImageSourceTypeExportVolume {
+		resource.AddAction(request, actionExport)
+	}
```
gitlawr: I think we need to add the action to volume(pvc) resources. Looks like we are adding it to vmimages here

pkg/api/volume/schema.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/rancher/steve/pkg/schema"
+	"github.com/rancher/steve/pkg/server"
+	"github.com/rancher/wrangler/pkg/schemas"
+
+	"github.com/harvester/harvester/pkg/config"
+)
+
+func RegisterSchema(scaled *config.Scaled, server *server.Server, options config.Options) error {
+	t := schema.Template{
+		ID: "harvesterhci.io.exportvolume",
```
gitlawr: this doesn't look like a valid resource ID

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -127,16 +127,26 @@ func (h *vmImageHandler) createBackingImage(image *harvesterv1.VirtualMachineIma
...
+
 	_, err := h.backingImages.Create(bi)
 	return err
 }
 
 func (h *vmImageHandler) createStorageClass(image *harvesterv1.VirtualMachineImage) error {
+	// we don't need to create storageclass if we are exporting existing volume to image
+	if image.Spec.SourceType == harvesterv1.VirtualMachineImageSourceTypeExportVolume {
+		return nil
+	}
```
gitlawr: We should create a storageclass as well. A VMImage type volume of a VM is a PVC with the storageclass of the image.

gitlawr (commented):
We also need to run `go generate` to update CRD schemas& docs with the API changes

pkg/api/volume/formatter.go
shangma: @gitlawr fixed

pkg/api/volume/formatter.go
shangma: @gitlawr  fixed

pkg/api/vm/types.go
shangma: @gitlawr updated.

pkg/api/volume/formatter.go
shangma: @gitlawr updated

pkg/api/volume/schema.go
shangma: @gitlawr updated

pkg/controller/master/image/vm_image_controller.go
shangma: @gitlawr  updated

pkg/api/volume/formatter.go
shangma: @gitlawr updated

pkg/api/vm/types.go
```diff
@@ -37,3 +37,7 @@ type AddVolumeInput struct {
...
+	DisplayName string `json:"displayName"`
```
gitlawr, line 42: We also need the namespace input so as to support cross-namespace creation. In validation webhook we need to reject the creation if users have no image creation permission in the target namespace(Probably use the SelfSubjectAccessReview resource).

pkg/api/volume/formatter.go
```diff
@@ -0,0 +1,87 @@
...
+	default:
+		return apierror.NewAPIError(validation.InvalidAction, "Unsupported action")
+	}
+}
+
+func (h ExportActionHandler) exportVolume(ctx context.Context, namespace, displayName, pvcName string) error {
+	pvc, err := h.pvcCache.Get(namespace, pvcName)
+	if err != nil {
+		return fmt.Errorf("failed to get pvc %s/%s, error: %s", pvcName, namespace, err.Error())
+	}
```
gitlawr: We still need to move the check to the validation webhook. If the pvcName is invalid we should catch the error from the validation webhook in L81

pkg/api/volume/formatter.go
```diff
@@ -0,0 +1,87 @@
...
+	vmi := &harvesterv1.VirtualMachineImage{
+		Spec: harvesterv1.VirtualMachineImageSpec{
```
gitlawr: The metadata is missing here. Does the creation succeed? I thought we need namespace and generateName here.

There's another resource called virtualMachineInstance, let's not use vmi for the variable name

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -72,6 +72,8 @@ func (v *virtualMachineImageValidator) Create(request *types.Request, newObj run
 		return werror.NewInvalidError(`url is required when image source type is "download"`, "spec.url")
 	} else if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeUpload && newImage.Spec.URL != "" {
 		return werror.NewInvalidError(`url should be empty when image source type is "upload"`, "spec.url")
+	} else if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeExportVolume && newImage.Spec.URL != "" {
```
gitlawr: Let's combine these and check by `sourceType != download && url != ""`

pkg/api/volume/formatter.go
```diff
@@ -0,0 +1,87 @@
...
+			DisplayName: displayName,
+			SourceType:  harvesterv1.VirtualMachineImageSourceTypeExportVolume,
+			VolumeName:  pvc.Spec.VolumeName,
```
gitlawr: I think we need to use PVCName for this image field. `pvc.Spec.VolumeName(PV)` is not something we expose to the users. We can fetch the PersistentVolumeName in the controller and set it to backing images.

pkg/api/vm/types.go
shangma: @gitlawr updated.

pkg/api/volume/formatter.go
shangma: @gitlawr updated

pkg/api/volume/formatter.go
shangma: @gitlawr updated.

pkg/webhook/resources/virtualmachineimage/validator.go
shangma: @gitlawr  updated.

pkg/api/volume/formatter.go
shangma: @gitlawr updated.

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -127,6 +129,16 @@ func (h *vmImageHandler) createBackingImage(image *harvesterv1.VirtualMachineIma
 		bi.Spec.SourceParameters[types.DataSourceTypeDownloadParameterURL] = image.Spec.URL
 	}
 
+	if image.Spec.SourceType == harvesterv1.VirtualMachineImageSourceTypeExportVolume {
+		pvc, err := h.pvcCache.Get(image.Namespace, image.Spec.VolumeName)
```
gitlawr: We cannot use the image namespace to get PVC. The PVC namespace may be different from the image namespace.

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -70,8 +70,12 @@ func (v *virtualMachineImageValidator) Create(request *types.Request, newObj run
 
 	if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeDownload && newImage.Spec.URL == "" {
 		return werror.NewInvalidError(`url is required when image source type is "download"`, "spec.url")
-	} else if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeUpload && newImage.Spec.URL != "" {
-		return werror.NewInvalidError(`url should be empty when image source type is "upload"`, "spec.url")
+	} else if newImage.Spec.SourceType != v1beta1.VirtualMachineImageSourceTypeDownload && newImage.Spec.URL != "" {
+		return werror.NewInvalidError(`url should be empty when image source type is not "download"`, "spec.url")
+	} else if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeExportVolume {
+		if _, err := v.pvcCache.Get(newImage.Namespace, newImage.Spec.VolumeName); err != nil {
+			return werror.NewInvalidError(`volumename should be valid when image source type is "export-from-volume"`, "spec.volumename")
```
gitlawr: Print the original err in the message. It could fail for other reasons even if the name is valid

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -70,8 +70,12 @@ func (v *virtualMachineImageValidator) Create(request *types.Request, newObj run
 
 	if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeDownload && newImage.Spec.URL == "" {
 		return werror.NewInvalidError(`url is required when image source type is "download"`, "spec.url")
-	} else if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeUpload && newImage.Spec.URL != "" {
-		return werror.NewInvalidError(`url should be empty when image source type is "upload"`, "spec.url")
+	} else if newImage.Spec.SourceType != v1beta1.VirtualMachineImageSourceTypeDownload && newImage.Spec.URL != "" {
+		return werror.NewInvalidError(`url should be empty when image source type is not "download"`, "spec.url")
+	} else if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeExportVolume {
```
gitlawr, line 80: Can we add a permission check to ensure the user has related PVC permission when it references the PVC in the image object?

pkg/apis/harvesterhci.io/v1beta1/image.go
```diff
@@ -39,9 +40,12 @@ type VirtualMachineImageSpec struct {
 	DisplayName string `json:"displayName"`
 
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Enum=download;upload
+	// +kubebuilder:validation:Enum=download;upload;export-from-volume
 	SourceType string `json:"sourceType"`
 
+	// +optional
+	VolumeName string `json:"volumeName"`
```
gitlawr: PersistentVolumeClaimName might be better. I think the namespace is also needed

pkg/webhook/resources/virtualmachineimage/validator.go
shangma: @gitlawr updated.

pkg/api/volume/formatter.go
```diff
@@ -0,0 +1,91 @@
...
+		status := http.StatusInternalServerError
+		if e, ok := err.(*apierror.APIError); ok {
+			status = e.Code.Status
+		}
+		rw.WriteHeader(status)
+		_, _ = rw.Write([]byte(err.Error()))
+		return
+	}
+	rw.WriteHeader(http.StatusOK)
+	_, _ = rw.Write([]byte("Successfully export the volume."))
```
shangma, line 44: @gitlawr  @guangbochen  I have to add this line (write sth, anything to http.ResponseWriter) to stop this "parsererror"
Screenshot from 2021-09-22 23-11-40
:

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -70,8 +75,36 @@ func (v *virtualMachineImageValidator) Create(request *types.Request, newObj run
...
```
gitlawr, line 80: We can check required fields early. Otherwise, we may return "failed to check user permission" error when pvcName is missing which is not user-friendly.

pkg/api/volume/formatter.go
```diff
@@ -0,0 +1,91 @@
...
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/rancher/apiserver/pkg/apierror"
+	"github.com/rancher/apiserver/pkg/types"
+	"github.com/rancher/wrangler/pkg/schemas/validation"
+	"github.com/sirupsen/logrus"
+
```
gitlawr: nit: remove this empty line

pkg/webhook/resources/virtualmachineimage/validator.go
shangma: @gitlawr updated

pkg/api/volume/formatter.go
shangma: @gitlawr updated

pkg/apis/harvesterhci.io/v1beta1/image.go
shangma: @gitlawr updated

pkg/webhook/resources/virtualmachineimage/validator.go
shangma: @gitlawr updated

pkg/controller/master/image/vm_image_controller.go
shangma: @gitlawr updated

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

Issue #1219, opened 2021-09-14
[BUG] Air-gapped environment won't load the dashboard


noahgildersleeve:
**Describe the bug**

When you install Harvester to an air-gapped environment it finishes the install but won't load the dashboard or the auth/setup pages
**To Reproduce**
1. Install to an air-gapped environmnet
2. Finish Install
3. Reboot
4. Wait for harvester to show as ready
5. Navigate to URL
**Expected behavior**

Dashboard should load
**Support bundle**

**Environment:**
 - Harvester ISO version: harvester-preview1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): virtualized environment with KVM

**Additional context**
![Screenshot from 2021-09-14 09-05-00](https://user-images.githubusercontent.com/83787952/133293500-7c545750-90bd-4b64-b778-9138efd9b0a6.png)

guangbochen:
Will need to add support to serve UI from embedded cluster instead of remote server URL.

noahgildersleeve:
I thought that was a supported option, it just defaulted to the remote server URL.

noahgildersleeve:
![image](https://user-images.githubusercontent.com/83787952/133296699-0167e9ef-3b44-42a9-8a60-9d7c1bbd16a2.png)

gitlawr:
To test:
For tag release, check that UI is loaded as expected in air-gapped env.

For master build, it still defaults to fetch the latest remote UI.
Run
```
kubectl patch setting ui-source --type=merge -p 'value: bundled'
```
Then verify it works in air-gapped env.

guangbochen:
verified using the `master-c6474a74-head`
1. go to the Advanced => Settings and config `ui-source` to `Bundled`
2. Refresh the page again, make sure the UI is loaded successfully and the JS/CSS assets are loaded from the local server.
![image](https://user-images.githubusercontent.com/4569037/134847777-38752ce8-1686-4ada-8565-c0bbf2cca91e.png)

PR #1230, opened 2021-09-16
Serve bundled UI in Harvester server

gitlawr:
https://github.com/harvester/harvester/issues/1219

**Problem:**
Fail to load UI in offline environments.

**Solution:**
Get dashboard UI from Harvester api-server.

**Test plan:**
For tag release, check that UI is loaded as expected in air-gapped env.

For master build, it still defaults to fetch the latest remote UI. 
Run
```
kubectl patch setting ui-source --type=merge -p 'value: bundled'
```
Then verify it works in air-gapped env.

### DIFF
--- a/pkg/controller/master/rancher/embedded_rancher.go
+++ b/pkg/controller/master/rancher/embedded_rancher.go
@@ -5,14 +5,10 @@ import (
 
 	rancherv3api "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/sirupsen/logrus"
-
-	"github.com/harvester/harvester/pkg/settings"
 )
 
 var UpdateRancherUISettings = map[string]string{
-	"ui-dashboard-index":   settings.DefaultDashboardUIURL,
-	"ui-offline-preferred": "false",
-	"ui-pl":                "Harvester",
+	"ui-pl": "Harvester",
 }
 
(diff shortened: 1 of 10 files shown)


pkg/controller/master/rancher/embedded_rancher.go
```diff
@@ -4,17 +4,8 @@ import (
...
 	rancherv3api "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
-	"github.com/sirupsen/logrus"
-
-	"github.com/harvester/harvester/pkg/settings"
 )
 
-var UpdateRancherUISettings = map[string]string{
-	"ui-dashboard-index":   settings.DefaultDashboardUIURL,
-	"ui-offline-preferred": "false",
-	"ui-pl":                "Harvester",
```
guangbochen, line 15: the `ui-pl: Harvester` is required since the UI uses this to identify whether it is a Harvester cluster before login.

guangbochen:
additionally, I think we should display the `ui-source` setting on the settings page of a single Harvester Cluster UI. cc @WuJun2016

pkg/controller/master/rancher/embedded_rancher.go
gitlawr: Updated

guangbochen (approved):
LGTM

PR #1294, opened 2021-09-26
Mark node schedulable only if it supports virtualization

### LINKED ISSUE #1290: [FEATURE] Support running Harvester on VM
Now when Harvester is based on RKE2 it is actually easiest way to create RKE2 clusters. That why it would be very useful to have support to run Harvester also on VMs which would allow configurations like:
* Master nodes on VMs + workers on bare metal
* Shared HCI platform on bare metal and separate dev and production application RKE2 clusters on VMs (Harvester in Harvester)

Afaiu only requirement for Harvester to support VM mode is that it should detect nodes which which have virtualization support, label them and use `nodeSelector` on kubevirt and kubevirt-operator charts.

olljanat:
**Problem:**
I did piloting by running Harvester on VM without enabling nested virtualization and noticed that node still had `kubevirt.io/schedulable=true` label.

**Solution:**
Added extra check that node will be marked schedulable only if `/dev/kvm` exist like it is done on debug script too:
https://github.com/harvester/harvester/blob/40dfbba994d4dce19e2909a73135369098f603a2/hack/host-check.sh#L30-L37

**Related Issue:**
Closes #1290

**Test plan:**
Disable virtualization support from VM/bare metal, install Harvester to it and verify that it have `kubevirt.io/schedulable=false` label. Then enable virtualization support and make sure that node gets `kubevirt.io/schedulable=true` label.

### DIFF
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -3,6 +3,8 @@ package node
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


olljanat:
CI needs a bit hacky trick because it is actually using software emulation but afaiu other way this works now as expected.

Also I was not able to do full test set yet because there looks to be problems with building working ISO files on current code base (because of recent changes on harvester-installer afaiu).

tlehman:
Since #1290 was closed, should this PR be closed?

Issue #1072, opened 2021-08-04
[FEATURE] Kubeconfig download button


noahgildersleeve:
**Is your feature request related to a problem? Please describe.**

We need a button in Harvester for downloading the cluster kubeconfig
**Describe the solution you'd like**

There should be a button that downloads the kubeconfig for the kubernetes cluster. **Describe alternatives you've considered**

Currently there is no way other than going into embedded rancher, or SSH into the node. We could also do it via an API call, but if that is the case there should still be a UI option.
**Additional context**

The kubeconfig download button is also currently not working in embedded rancher.  We also should probably restrict this to admins in the user roles.

noahgildersleeve:
Tested in master-8afedf0a-head. Fixed.

noahgildersleeve:
Re-opening. Misunderstood implementation.

guangbochen:
![image](https://user-images.githubusercontent.com/4569037/133730725-1508301e-1989-40b0-95f0-8f6d77c26d76.png)
u should be able to download it from the top navbar and use kubectl to access this cluster e.g `KUBECONFIG=./cluster.yaml kubectl get node`

noahgildersleeve:
Tested in master-1a041d23-head with the steps outlined above.

1. Create a harvester setup with master
2. download kubeconfig from topbar
3. Use kubeconfig to access cluster with yaml with `kubectl get node`

I got an error with the certificate data. Unable to connect to the server: x509: certificate is valid for x.x.x.x, x.x.x.y, x.x.y.y, not y.y.y.y. I took a look and the kubeconfig was different from what was on the node at `/etc/var/rancher/rke2/rke2.yaml`. In the downloaded cubeconfig there isn't anything about `certificate-authority-data`, or `client-certificate-data`

gitlawr:
@noahgildersleeve kubeconfig from the UI is generated per user, so it's different from the one on the node.

In master-1a041d23-head,
- When I access Harvester with https://VIP and download the kubeconfig, it works.
- When I access Harvester with https://node-ip:30443 and download the kubeconfig, it shows the x509 error. The kubeconfig uses the node-ip as ther cluster server endpoint but the IP is not added to certificate SANs.

@WuJun2016 We need to bring back the server-url setting on the first login. The generated kubeconfig will use the consistent server-url once it's set.

WuJun2016:
server-url is set through the backend

noahgildersleeve:
[...]

Issue #1001, opened 2021-07-05
[BUG] No filtering on Projects/Namespaces page


ldevulder:
**Describe the bug**
In `Projects/Namespaces` tab, when I try to filter by selecting `All Namespaces`, `Only System Namespaces` or something else I still have the same results. It seems that the filtering is not working at all.

**To Reproduce**
1. Go to `Projects/Namespaces`
2. Select another filter than the default one: list is still the same.

**Expected behavior**
Filter should work.

**Support bundle**
[harvester-supportbundle_e707144b-450c-4130-b3a9-3255c3159c1d_2021-07-05T10-12-41Z.zip](https://github.com/harvester/harvester/files/6763444/harvester-supportbundle_e707144b-450c-4130-b3a9-3255c3159c1d_2021-07-05T10-12-41Z.zip)

**Environment:**
 - Harvester ISO version: master-88f1d22c-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): HP ProLiant DL360 Gen9

**Additional context**
None.

WuJun2016:
@gitlawr    We should add annotation to the namespace to filter the harvester-system namespace

guangbochen:
as per https://github.com/rancher/rancher/issues/34911 discussion, we will follow the following conditions to filtering those system-related namespaces from the UI:
- We maintain a setting SystemNamespaces that has the default system namespaces
- namespaces added dynamically for the fleet for example and they are added to the system project and carry an annotation "management.cattle.io/system-namespace": "true"

noahgildersleeve:
Tested in master-f38adc70-head. 

- Showing available namespaces
![image](https://user-images.githubusercontent.com/83787952/135151481-0618d937-14ca-4a2d-b000-a86a3fed5204.png)
- List of VMs in namespaces
![image](https://user-images.githubusercontent.com/83787952/135151595-95f98a14-6e77-45dd-967b-bb42ba9aa498.png)

PR #1295, opened 2021-09-27
Init server-url and system-namespaces settings

gitlawr:
https://github.com/harvester/harvester/issues/1072
https://github.com/harvester/harvester/issues/1001

### DIFF
--- a/pkg/controller/master/rancher/embedded_rancher.go
+++ b/pkg/controller/master/rancher/embedded_rancher.go
@@ -1,10 +1,14 @@
 package rancher
 
 import (
+	"fmt"
 	"strings"
+	"time"
 
 	rancherv3api "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	ranchersettings "github.com/rancher/rancher/pkg/settings"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/controller/master/rancher/embedded_rancher.go
```diff
@@ -33,6 +44,48 @@ func (h *Handler) RancherSettingOnChange(key string, setting *rancherv3api.Setti
...
+		// The default admin and its namespace is not created, enqueue and check later
+		h.RancherSettingController.EnqueueAfter(setting.Name, 30*time.Second)
+		return nil, nil
+	}
+	defaultAdminNamespace := users[0].Name
+	harvesterDefaultSystemNamespaces := strings.Join([]string{
+		ranchersettings.SystemNamespaces.Default,
+		"cattle-dashboards",
+		"cattle-fleet-clusters-system",
+		"cattle-monitoring-system",
```
guangbochen, line 76: Do u think if we use `cattle-*` as a general filter?

gitlawr: No. The default system namespaces list in Rancher uses full name of each namespace. And we don't want to prevent cattle lovers from using namespaces with this pattern :smiley:

guangbochen (approved):
LGTM

Issue #1298, opened 2021-09-28
[BUG] can't join the worker node of x509: certificate signed by unknown authority


guangbochen:
**Describe the bug**

3rd Node can't join the cluster via the PXE boot e.g refer to https://github.com/harvester/ipxe-examples/tree/main/vagrant-pxe-harvester, when the agent node curl the master VIP, it reports `error while connecting to Kubernetes cluster: Get "https://192.168.0.131/version": dial timeout` 

```
Sep 28 01:35:44 harvester-node-2 rancher-system-agent[2502]: time="2021-09-28T01:35:44Z" level=info msg="Rancher System Agent version v0.1.0 (14a2b5d) is sta>
Sep 28 01:35:44 harvester-node-2 rancher-system-agent[2502]: time="2021-09-28T01:35:44Z" level=info msg="Using directory /var/lib/rancher/agent/work for work"
Sep 28 01:35:44 harvester-node-2 rancher-system-agent[2502]: time="2021-09-28T01:35:44Z" level=debug msg="Instantiated new image utility with imagesDir: /var>
Sep 28 01:35:44 harvester-node-2 rancher-system-agent[2502]: time="2021-09-28T01:35:44Z" level=info msg="Starting remote watch of plans"
Sep 28 01:36:14 harvester-node-2 rancher-system-agent[2502]: panic: error while connecting to Kubernetes cluster: Get "https://192.168.0.131/version": dial t>
Sep 28 01:36:14 harvester-node-2 rancher-system-agent[2502]: goroutine 8 [running]:
Sep 28 01:36:14 harvester-node-2 rancher-system-agent[2502]: github.com/rancher/system-agent/pkg/k8splan.(*watcher).start(0xc000469300, 0x1757da0, 0xc00049ed>
Sep 28 01:36:14 harvester-node-2 rancher-system-agent[2502]:         /go/src/github.com/rancher/system-agent/pkg/k8splan/watcher.go:96 +0xc74
Sep 28 01:36:14 harvester-node-2 rancher-system-agent[2502]: created by github.com/rancher/system-agent/pkg/k8splan.Watch
Sep 28 01:36:14 harvester-node-2 rancher-system-agent[2502]:         /go/src/github.com/rancher/system-agent/pkg/k8splan/watcher.go:57 +0xbe
Sep 28 01:36:14 harvester-node-2 systemd[1]: rancher-system-agent.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
[...]

PR #1303, opened 2021-09-28
sets cacerts default on cacerts syncs

gitlawr:
**Problem:**
Currently, we set the cacerts value once. When a rancher pod(with --no-cacerts) is created/restarted, the cacerts will be reset to empty.

**Solution:**
Sets cacerts value on cacerts syncs.

**Related Issue:**
https://github.com/harvester/harvester/issues/1298


guangbochen (approved):
LGTM

PR #1310, opened 2021-09-28
Enable upgrade controller

bk201:
- System Upgrade Controller is in `cattle-system` namespace
- Do not process jobs and pods that are not in upgrade namespace.


guangbochen (approved):

gitlawr (approved):
LGTM

Issue #1307, opened 2021-09-28
[BUG] VLAN Network can be deleted when a Running VM uses it


FrankYang0529:
**Describe the bug**
From admission webhook test plan https://github.com/harvester/harvester/pull/919, VLAN network should not be able to delete when there is a VM attached with it.

**To Reproduce**
1. Enable VLAN network in settings
2. Create a network with VLAN 5 and assume its name is `my-network`.
3. Create a VM on VLAN 5.
4. Delete the network `my-network`.

**Expected behavior**
Network can't be deleted.

**Support bundle**

**Environment:**
 - Harvester ISO version: master-b7b757e9-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

**Additional context**

johnliu55tw:
## Test info
- Verified version: master-c4a76706-head

## Test cases and results
- Create a VLAN network
  - [x] Create a VM using the network, then delete the network should be prohibited - PASS
  - [x] Delete the VM, then delete the VLAN network should succeed - PASS

PR #1319, opened 2021-09-29
fix(webhooks): use correct network name formats to find VMs

FrankYang0529:
**Problem:**
Network attachment definition can be deleted and it is still used by a running VM.

**Solution:**
Fix network name which webhooks use to filter VM indexer.

**Related Issue:**
https://github.com/harvester/harvester/issues/1307

**Test plan:**
* Enable VLAN network in settings
* Create a network with VLAN 5 and assume its name is `my-network`.
* Create a VM on VLAN 5. Delete network `my-network` and it should fail with `admission webhook "[validator.harvesterhci.io](http://validator.harvesterhci.io/)" denied the request: default/my-network is still used by vm(s): vm` in a modal.


weihanglo (approved):
Taking care of both possible names is great!

Tested via teleprescene and ISO. Both LGTM. 
image

bk201 (approved):
LGTM

Issue #1320, opened 2021-09-29
[BUG] Unable to generate ISO using command "make build-iso"


johnliu55tw:
**Describe the bug**
`$ make build-iso` results in error:
```
INFO[0255] 🍜  Adding files to ISO from overlay
INFO[0257] 🍹 Generate ISO /go/src/github.com/harvester/harvester-installer/dist/artifacts/harvester-v0.2.4-amd64.iso
xorriso 1.5.0 : RockRidge filesystem manipulator, libburnia project.

Drive current: -outdev 'stdio:/go/src/github.com/harvester/harvester-installer/dist/artifacts/harvester-v0.2.4-amd64.iso'
Media current: stdio file, overwriteable
Media status : is blank
Media summary: 0 sessions, 0 data blocks, 0 data,  653g free
Added to ISO image: directory '/'='/tmp/luet-geniso486379853/tempISO'
xorriso : UPDATE :     625 files added in 1 seconds
xorriso : FAILURE : Given path does not exist on disk: -boot_image system_area='/tmp/luet-geniso486379853/tempISO/boot/syslinux/isohdpfx.bin'
xorriso : FAILURE : Cannot find in ISO image: -boot_image ... bin_path='/boot/syslinux/isolinux.bin'
xorriso : UPDATE :     625 files added in 1 seconds
xorriso : aborting : -abort_on 'FAILURE' encountered 'FAILURE'
INFO[0257] exit status 5
ERRO[0258] fatal error: exit status 5
FATA[0574] exit status 1
make: *** [Makefile:11: build-iso] Error 1
```

**To Reproduce**
1. Clone `harvester` repository
2. Run `make build-iso`

**Expected behavior**
`make build-iso` should run without any error.

**Support bundle**

**Environment:**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
Running `make` from `harvester-installer` repository generates the ISO without any issue. I think it's because they use different versions of `luet-makeiso`. harvester uses `0.3.3` and harvester-installer uses `0.3.8`. The older version of `luet-makeiso` does not support changing the ISO boot file.

gitlawr:
Verified on master head

PR #1321, opened 2021-09-29
Bump luet-makeiso version to 0.3.8

johnliu55tw:
**Problem:**
Running command `make build-iso` to generate ISO results in error.

**Solution:**
Update the version of `luet-makeiso` so that it supports changing `boot_file`, `boot_catalog`, `and isohybrid_mbr`, which are required for Grub2.

**Related Issue:**
https://github.com/harvester/harvester/issues/1320

**Test plan:**
`make build-iso` should generate ISO without any issue.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -56,7 +56,7 @@ RUN curl -Lo /usr/bin/kind https://kind.sigs.k8s.io/dl/v0.11.1/kind-linux-amd64 
 
 # makeiso
 RUN wget --quiet https://github.com/mudler/luet/releases/download/0.17.0/luet-0.17.0-linux-amd64 -O /usr/bin/luet && chmod +x /usr/bin/luet
-RUN wget --quiet https://github.com/mudler/luet-makeiso/releases/download/0.3.3/luet-makeiso-0.3.3-linux-amd64 -O /usr/bin/luet-makeiso && chmod +x /usr/bin/luet-makeiso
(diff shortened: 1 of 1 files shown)


bk201 (approved):
Good catch. LGTM

weihanglo (approved):
Confirmed that with luet-makeiso 0.3.8, we can `make build-iso` successfully.

gitlawr:
Thanks! I can build successfully with the fix.
Another issue is When I create a tag(e.g. v0.3.0-dev) and `make build-iso`, it generates the following artifacts. I would expect we use the Harvester version instead:
```
$ ls dist/artifacts/
harvester-v0.2.4-amd64.iso  harvester-v0.2.4-amd64.iso.sha256  harvester-v0.2.4-initrd-amd64  harvester-v0.2.4-rootfs-amd64.squashfs  harvester-v0.2.4-vmlinuz-amd64
```

bk201:
> Thanks! I can build successfully with the fix. Another issue is When I create a tag(e.g. v0.3.0-dev) and `make build-iso`, it generates the following artifacts. I would expect we use the Harvester version instead:
> 
> ```
> $ ls dist/artifacts/
> harvester-v0.2.4-amd64.iso  harvester-v0.2.4-amd64.iso.sha256  harvester-v0.2.4-initrd-amd64  harvester-v0.2.4-rootfs-amd64.squashfs  harvester-v0.2.4-vmlinuz-amd64
> ```

if DRONE_TAG presents, that should be change to it: https://github.com/harvester/harvester-installer/blob/212059f8bc7561477e258bb3554ed7ba008f3a36/scripts/package-harvester-os#L22-L24

gitlawr (approved):
LGTM

gitlawr:
@johnliu55tw Can you help to pick this commit to v0.3 branch?

johnliu55tw:
@gitlawr Done: #1322

Issue #1218, opened 2021-09-14
[BUG] Missing http proxy settings on rke2 and rancher pod


bashofmann:
**Describe the bug**
When you run Harvester in an airgapped environment, where you only have internet connectivity through an http proxy, the proxy settings that are set during the installation process are not applied to RKE2 or the Rancher deployment.

**To Reproduce**
Steps to reproduce the behaviour:
* Install Harvester in an airgapped environment
* Provide the necessary http_proxy, https_proxy and no_proxy settings
* Check if the Rancher Deployment/Pods have these environment variables configured
* Check if the rke2-server systemd unit has these environment variables configured

**Expected behavior**
The environment variables should be there.

**Environment:**
 - Harvester ISO version: 0.3.0-preview1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): baremetal

**Additional context**
If they are missing, RKE2 can not pull/update additional container images from the internet. Also the remote UI functionality of Rancher, plus additional features where Rancher connects to the internet (KDM, chart repos, fleet git repos, ...) do not work.

For 0.3.0-preview1 this means that rancher answers with A HTTP 200 response code and a blank body, because it can not fetch the remote index.html.

From the UI side, it would be great to have an HTTP 5xx response code with a sensible error message and something in visible in the error logs.

gitlawr:
~~Depending on https://github.com/rancher/rke2/issues/1834~~

yasker:
~~@gitlawr What's the impact on missing the RKE2 proxy part? We cannot start any Pods, except for the internal packaged images? Does RKE1 work with proxy?~~

~~Also, is there any manual workaround for this problem?~~

The above questions might not be necessary anymore since it seems RKE2 team has fixed the problem. Now we need the reproduce steps to validate.

gitlawr:
Also related to https://github.com/rancher/rancher/issues/34908

guangbochen:
move this one to v1.0.0 since we need to wait for the https://github.com/rancher/rancher/issues/34908 to fully address the air-gap env support of the Rancher RKE2 server.

janeczku:
**Workaround for container image downloading via a HTTP proxy**

In the Harvester install config YAML add the following write_files stanza (see [docs](https://docs.harvesterhci.io/v0.3/install/harvester-configuration/#oswrite_files)):

```
os:
  write_files:
  - path: /etc/default/rke2-server
    content: |
      # Environment variables
[...]

Issue #1012, opened 2021-07-09
[BUG] Failed to create image when deployed in private network environment


johnliu55tw:
**Describe the bug**
I've deployed Harvester using ISO in a private network, where HTTP Proxy is required to fetch external URLs. During the installation, the HTTP Proxy was set accordingly.

I tried to create an image using an external URL (https://cloud-images.ubuntu.com/focal/20210707/focal-server-cloudimg-amd64.img), it failed and shows error message:
> Head https://cloud-images.ubuntu.com/focal/20210707/focal-server-cloudimg-amd64.img: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
![image](https://user-images.githubusercontent.com/5047724/125034010-a915fd80-e0c2-11eb-8d73-5a6469d6a6ae.png)

Then I SSHed into harvester and run command `sudo k3os config --dump` then I notice there was no `http_proxy` and `https_proxy` inside the config.

To check if it's other networking issues, I set up an Nginx HTTP server to serve the same image within the private network, then Harvester can create the image without any issue.
![image](https://user-images.githubusercontent.com/5047724/125035462-6bb26f80-e0c4-11eb-9392-b2ebb12e33c3.png)

**To Reproduce**

1. Install Harvester using ISO, and properly set up HTTP Proxy information.
2. Wait until Harvester to enter ready state.
3. Create an image using URL https://cloud-images.ubuntu.com/focal/20210707/focal-server-cloudimg-amd64.img
4. The creation **failed** with error message: `Head https://cloud-images.ubuntu.com/focal/20210707/focal-server-cloudimg-amd64.img: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`

**Expected behavior**
Create images with external URLs in a private network using HTTP Proxy should be fine.

**Support bundle**

**Environment:**
 - Harvester ISO version: harvester-v0.2.0-amd64.iso
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Baremetal

**Additional context**
I've tried updated HTTP Proxy information by adding `/var/lib/rancher/k3os/config.yaml` with following content:
```
k3os:
  environment:
    http_proxy: http://our-proxy:8080/
    https_proxy: http://our-proxy:8080/
```
Then enter the command `sudo k3os config` but this doesn't seem to work. I'm not familiar with k3os, maybe there's a way to configure HTTP Proxy manually?

gitlawr:
[...]

PR #1337, opened 2021-09-30
Add http-proxy setting

gitlawr:
**Problem:**
Unable to use HTTP proxy to access external services, including downloading images and backup to s3 services.

**Solution:**
When the setting is updated, inject HTTP proxy envvars to system pods and the backup secret

**Related Issue:**
Address https://github.com/harvester/harvester/issues/1012
Partially address https://github.com/harvester/harvester/issues/1218, the missing part is envvar propagation to RKE2

**Test plan:**
1. Set up Harvester in a private network environment that can only access the internet via an HTTP proxy
2. Configure the HTTP proxy in the Harvester settings page(minimum NO_PROXY: `localhost,127.0.0.1,0.0.0.0,10.0.0.0/8,cattle-system.svc,.svc,.cluster.local,<your-internal-network>`)
3. Verify creating images from URL and backup to s3 works.
4. Verify updating/removing the proxy configuration works.


pkg/util/constants.go
```diff
@@ -9,6 +9,9 @@ const (
 	AnnotationTimestamp            = prefix + "/timestamp"
 	AnnotationVolumeClaimTemplates = prefix + "/volumeClaimTemplates"
 	AnnotationImageID              = prefix + "/imageId"
+	AnnotationHash                 = prefix + "/hash"
 
+	BackupTargetSecretName      = "harvester-backup-target-secret"
+	CattleSystemNamespaceName   = "cattle-system"
```
guangbochen: can we please replace the other redundant `cattle-system` constant, thanks?

pkg/webhook/resources/pod/mutator.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+var matchingLabels = map[string]string{
+	"longhorn.io/component":       "backing-image-data-source",
+	"app.kubernetes.io/component": "apiserver",
+	"app":                         "rancher",
+}
```
guangbochen: what about the Harvester API server?

pkg/controller/master/setting/http_proxy.go
```diff
@@ -0,0 +1,92 @@
...
+	if setting.Value == "" && setting.Annotations[util.AnnotationHash] == "" {
+		return nil, nil
+	}
+
+	hash := sha256.New224()
+	hash.Write([]byte(setting.Value))
+	currentHash := fmt.Sprintf("%x", hash.Sum(nil))
+	if currentHash == setting.Annotations[util.AnnotationHash] {
+		return nil, nil
+	}
```
guangbochen: prefer to add a short comment on the hash usage.

pkg/controller/master/setting/http_proxy.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		return err
+	}
+	toUpdate := secret.DeepCopy()
+	if toUpdate.Data == nil {
+		toUpdate.Data = make(map[string][]byte)
+	}
+	toUpdate.Data["HTTP_PROXY"] = []byte(httpProxyConfig.HTTPProxy)
+	toUpdate.Data["HTTPS_PROXY"] = []byte(httpProxyConfig.HTTPSProxy)
+	toUpdate.Data["NO_PROXY"] = []byte(httpProxyConfig.NoProxy)
```
guangbochen: I think we should mitigate the `NO_PROXY` config of those known configs needed of the k8s service, e.g 
`localhost,127.0.0.1,cattle-system.svc,.svc,.cluster.local` and get them focus on their system configs only.

pkg/webhook/resources/pod/mutator.go
gitlawr: `"app.kubernetes.io/component": "apiserver"` is for Harvester apiserver. I will update the chart and use "harvester-apiserver" to avoid confusion

pkg/controller/master/setting/http_proxy.go
gitlawr: Sounds good

gitlawr:
Updated. Restructure some code for extensibility to other settings like CA certs.

pkg/webhook/server/server.go
```diff
@@ -130,7 +131,7 @@ func (s *AdmissionWebhookServer) listenAndServe(clients *clients.Clients, handle
 						CABundle: secret.Data[corev1.TLSCertKey],
 					},
 					Rules:                   mutationRules,
-					FailurePolicy:           &failPolicyFail,
+					FailurePolicy:           &failPolicyIgnore,
```
guangbochen, line 134: why do we need to update the failure policy to `ignore`?

gitlawr: Webhook for pods is critical, we may shoot ourselves in the foot using Fail policy. e.g., in disaster recovery scenarios.

guangbochen (approved):
LGTM

Issue #1338, opened 2021-09-30
[BUG] Cannot restore VM after changing the Backup Target


johnliu55tw:
**Describe the bug**
Restoring from backup results in error after changing the Backup Target.

**To Reproduce**
1. Set up a backup target to S3 bucket `bucket1`
2. Create a VM, then create a backup of it
3. Verify the backup is ready
4. Change the backup target to another S3 bucket: `bucket2`
5. Verify the backup still exists
6. "Restore New" from the backup
7. The new VM stuck at "Scheduling" status with error: 
    ```
    0/3 nodes are available: 3 pod has unbound immediate PersistentVolumeClaims.
    ```

**Expected behavior**
I think there are several viable behaviors:
1. Backup/Restore should work just fine after changing the Backup Target.
2. Prohibit users from changing the Backup Target if they already created several backups. 
3. Warn users that all the backups would be removed if they changed the Backup Target, and we removed all the backups.

**Support bundle**

**Environment:**
 - Harvester ISO version: v0.3.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

**Additional context**
- **Error:**
![image](https://user-images.githubusercontent.com/5047724/135433258-57bea57d-c665-4857-8016-dd22c2913b3d.png)

- Everything went back to normal after I switched back to the original Backup Target (`bucket1`).

yasker:
Change the backup target should refresh all the backups (not deleting from the backup target). Since the backup doesn't exist in the new backup target, we definitely shouldn't show them.

FrankYang0529:
I have checked that longhorn delete all backups when we change the backup target, so we should also delete VM backup too when the backup target changed.

yasker:
@FrankYang0529 Can you confirm Longhorn only deletes the Backup CR object, not the real backup in the backup target?

FrankYang0529:
> @FrankYang0529 Can you confirm Longhorn only deletes the Backup CR object, not the real backup in the backup target?

Yes, Longhorn only deletes Backup CR objects. I still can see backups in my S3 bucket.

guangbochen:
One thing I have concerned about here is whether we should remove all the existing VMbackup CRs automatically when the user modified the backup target, I prefer to leave it for the users to manage those VM backups for the following reasons:
- users may change the backup target back and forth for validation or testing purposes
- we can disable users from using those VM backups if the current backup target is not equal to the stored backup target(e.g via webhook validation)

[...]

PR #1368, opened 2021-10-07
fix(backupTarget): delete inconsistent vmbackup

FrankYang0529:
**Problem:**
Old vmbackup CR objects are listed in the Backups page when we change Backup Target.

**Solution:**
Delete vmbackup CR objects which annotations are not consistent with Backup Target.

**Related Issue:**
https://github.com/harvester/harvester/issues/1338

**Test plan:**
Follow steps in the issue https://github.com/harvester/harvester/issues/1338.

**Comments**
We may need to warn users that vmbackup can't be retrieved when they change Backup Target back. Longhorn can scan Backup Target and retrieve backup CR objects. I'm not sure how to do it in Harvester side.


guangbochen:
close in favor of https://github.com/harvester/harvester/pull/1394

Issue #1351, opened 2021-10-05
[BUG] backup-target NFS unable to mount with error exit status 32


lanfon72:
### Describe the bug
Unable to set backup-target to use NFS `nfs://10.84.44.120:/mnt/nfs_share` as mount point.
![image](https://user-images.githubusercontent.com/5169694/135975120-92b02b42-01b8-4b6f-ab27-d2808995ee57.png)

### To Reproduce
1. create an NFS server to accept imcoming mount.
2. login as admin, nagivate to advance -> settings
3. edit `backup-target` for type `NFS`, input Endpoint follow the rule: `nfs://<server>:/path/`
4. failed to mount with error exit status 32

### Expected behavior
Success mount endpoint as backup-target.

## Environment:
 - Harvester ISO version: **master-c6474a74-head**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 3nodes**

### Additional context
- NFS server's `/etc/exports`:
```
/mnt/nfs_share 10.84.0.0/16(rw,sync,no_subtree_check)
```
- tested no permission issue from NFS:
![image](https://user-images.githubusercontent.com/5169694/135973269-0385786f-a308-4179-8e81-60c28c7ad536.png)

_Advanced_ Test Case No.6

joshimoo:
Just leaving some random thoughts here: for nfs permissions issues there are things like [nfsidmap](https://man7.org/linux/man-pages/man5/nfsidmap.5.html) that can be a mismatch and root squash settings on the server, etc.

As a first step, I would try the mounts from the associated pods where you are getting the above failure response from. For the longhorn side you should try the nfs mounts from `longhorn-manager, instance-manager-e, instance-manager-r) to verify.

yasker:
Also, we might be missing something on the OS side for the NFS.

gitlawr:
Previously, we add the SYS_ADMIN capability so that we can do mount inside Harvester pods. It doesn't work when apparmor is enabled. We can use the annotation: https://kubernetes.io/docs/tutorials/clusters/apparmor/#securing-a-pod

chajain:
I tried with the latest server version master-c9b341e9-head but still see the same issue

yasker:
@chajain `c9b341e9` doesn't have it. It's in the next commit `0136b81`. See https://github.com/harvester/harvester/commits/master .

chajain:
The backup-target NFS worked on latest master-0136b815-head image

[...]

PR #1393, opened 2021-10-08
Set apparmor profile of harvester containers

gitlawr:
https://github.com/harvester/harvester/issues/1351

**Problem:**
mount command fails inside the Harvester pods.

**Solution:**
Set apparmor profile of Harvester containers to unconfined


guangbochen:
LGTM,
previously config the NFS v4 server and failed with access denied:
image


after adding the metadata annotation:
image

guangbochen (approved):

PR #1407, opened 2021-10-10
Fix go mod dependency alerts

guangbochen:
Fix go mod dependency alerts https://github.com/harvester/harvester/security


gitlawr (approved):
LGTM

Issue #1406, opened 2021-10-10
[BUG] Read-only user was able to manage API actions


guangbochen:
**Describe the bug**
Read-only user was able to manage API actions, e.g stop/restart/start VMs

**To Reproduce**
1. import the downstream Harvester to the Rancher
2. create a `read-only` user and assign it to the `default` project
3. login as the `read-only` user and found the user was able to use the API actions.

![image](https://user-images.githubusercontent.com/4569037/136691117-deb308c7-b05f-4cd7-ad36-e520d306d0d7.png)

**Expected behavior**
- Read-only users should not be able to manage resources through the API actions.

**Environment:**
 - Harvester ISO version: v0.3.0-master-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Bare-metal

**Additional context**

gitlawr:
With the fix, there's one thing that needs to be checked on the UI side @WuJun2016 :

For a read-only user, actions are not available in volume APIs, but UI still shows the export action. A js error will pop up if the action is performed.
![image](https://user-images.githubusercontent.com/5697937/137092210-86d77a51-2817-4a9d-bd78-91134ff29260.png)

![image](https://user-images.githubusercontent.com/5697937/137091860-7fc997f2-0e2a-4a97-87aa-f5029691f09a.png)

TachunLin:
@guangbochen, @WuJun2016 Verified on master-ebfbf44a-head. 
Read-only user can't manage API actions now but still remain some settings in volume and images page

### Result

After checking, Read-only user can't do API management actions on VM, Image and Volume.
While for volume and images , there's still UI settings remains like **"export to image"** in Volume and **"create virtual machine"** in images. But read-user can't actually do these actions. 

Personal suggestion is we can verify it again after all relating read-only user UI settings have hided. 
Thus move it to review pipeline.  

Display export to image to read-only user
![image](https://user-images.githubusercontent.com/29251855/138629271-3be58a86-ec4f-46ba-9ca1-9840c0403566.png)

Display create virtual machine to read-only user
![image](https://user-images.githubusercontent.com/29251855/138629687-ebec6cbf-399a-4c07-894c-97c13216a72a.png)


### Test Information
* Test Environment: Provo bare machines 3 nodes (hpd8s7, hpd8s8, hpd8s9 in RackD8)
* Harvester version: master-ebfbf44a-head

### Verify Steps
[...]

PR #1423, opened 2021-10-13
Access control for resource actions

gitlawr:
**Problem:**
Read-only users can invoke actions

**Solution:**
Add access control in formatter

**Related Issue:**
https://github.com/harvester/harvester/issues/1406


guangbochen (approved):
LGTM

Issue #988, opened 2021-07-01
[FEATURE] backup VM required metadata to the external storage server


guangbochen:
**Is your feature request related to a problem? Please describe.**
Currently, only the VM volumes will be backup to the external storage server, we should also save the VM YAML to the external storage server so that it can be redeployed to the other cluster.

**Describe the solution you'd like**
- [x] store the VM metadata to the external storage server

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

yasker:
Those can be set using the label in the Backup.

FrankYang0529:
## Test plan

### Case 1 - Sync VMBackup Metadata back.

* Setup an NFS/S3 backup target.
* Create a VM `vm`.
* Modify a file like `tmp.txt` and input `123` in it.
* Create a VM Backup `backup1`.
* Modify a file like `tmp.txt` and change content to `456`.
* Create a VM Backup `backup2`.
  * Check associated VolumeSnapshot / VolumeSnapshotContent/ LH Backup exist.
  * Check `default-backup1.cfg` and `default-backup2.cfg` are in the backup target.
  * Check `LonghornBackupName` exists in VM Backup `backup1` and `backup2`.
  * Check `LonghornBackupName` in CR `backup1` and `backup2` is same as `default-backup1.cfg` and `default-backup2.cfg` in backup target.
* Change backup target to another NFS/S3 backup target.
* Delete the VM Backup `backup1` and `backup2` and VM `vm`.
  * Check associated VolumeSnapshot / VolumeSnapshotContent / LH Backup are deleted.
  * Check `default-backup1.cfg` and `default-backup2.cfg` are still in the original backup target.
* Change backup target back.
  * Check VMBackup `backup1` and `backup2` and associated VolumeSnapshot / VolumeSnapshotContent / LH Backup are created again.
* Restore `backup1` to `vm1`
  * Check whether the file has `123` in it.
* Restore `backup2` to `vm2`
  * Check whether the file has `456` in it.

### Case 2 - Delete VM Backup should delete the remote file.

* Setup an NFS/S3 backup target.
* Create a VM and create a VM Backup `backup1`.
  * Check whether `default-backup1.cfg` is in the backup target.
* Delete the VM Backup.
  * Check whether `default-backup1.cfg` is deleted in the backup target.

### Case 3 - Sync VM Backup to a fresh cluster
* Create a cluster 1 and follow Case 1 to create VM Backup `backup1` and `backup2`.
[...]

PR #1447, opened 2021-10-18
feat(backup): upload vmbackup metadata

FrankYang0529:
**Problem:**
We didn't store VMBackup metadata, so we can't redeploy to other clusters.

**Solution:**
Save VMBackup metadata to the backup target.

**Related Issue:**
https://github.com/harvester/harvester/issues/988

**Test plan:**
Case 1 - Sync VMBackup Metadata back.
* Setup an NFS/S3 backup target.
* Create a VM `vm`.
* Modify a file like `tmp.txt` and input `123` in it.
* Create a VM Backup `backup1`.
* Modify a file like `tmp.txt` and change content to `456`.
* Create a VM Backup `backup2`.
  * Check associated VolumeSnapshot / VolumeSnapshotContent/ LH Backup exist.
  * Check `default-backup1.cfg` and `default-backup2.cfg` are in the backup target.
  * Check `LonghornBackupName` exists in VM Backup `backup1` and `backup2`.
  * Check `LonghornBackupName` in CR `backup1` and `backup2` is same as `backup1.cfg` and `backup2.cfg` in backup target.
* Change backup target to another NFS/S3 backup target.
* Delete the VM Backup `backup1` and `backup2` and VM `vm`.
  * Check associated VolumeSnapshot / VolumeSnapshotContent / LH Backup are deleted.
  * Check `default-backup1.cfg` and `default-backup2.cfg` are still in the original backup target.
* Change backup target back.
  * Check VMBackup `backup1` and `backup2` and associated VolumeSnapshot / VolumeSnapshotContent / LH Backup are created again.
* Restore `backup1` to `vm1`
  * Check whether the file has `123` in it.
* Restore `backup2` to `vm2`
  * Check whether the file has `456` in it.

Case 2 - Delete VM Backup should delete the remote file.
* Setup an NFS/S3 backup target.
* Create a VM and create a VM Backup `backup1`.
  * Check whether `default-backup1.cfg` is in the backup target.
* Delete the VM Backup.
  * Check whether `default-backup1.cfg` is deleted in the backup target.

Case 3 - Delete VirtualMachineImage and restore VM. **(Not support in this PR. Just leave the test case as a reminder)**
* Follow Case 1, but also delete `VirtualMachineImage` after change backup target.
* Create `VirtualMachineImage` again with same `BackingImage` name.
* Restore VM Backup to VM. VM should run successfully.

**Environment Setup**
We move backup target validation to webhook and update `VirtualMachineBackup` CRD. If users would like to test the PR before it merged, please remember to update `harvester-webhook` deployment and `VirtualMachineBackup` CRD.

Screen Shot 2021-11-23 at 5 32 58 PM

Screen Shot 2021-11-23 at 5 33 19 PM


FrankYang0529:
I will create another PR to check backup metadata when users change the backup target. If there is backup metadata, we will create VMBackup CR in the cluster. Please help me review the upload flow and metadata format in this PR first. Thank you.

gitlawr:
How do we restore in another cluster, is it a manual procedure?

FrankYang0529:
> How do we restore in another cluster, is it a manual procedure?

No, I will create another PR for this part. I just want someone to help to review the upload flow and metadata format first.

For restoring, I will synchronize data when users change the backup target. Thanks.

guangbochen (commented):
We need to handle the delete case as well, also would prefer to merge it all together with the restore process as that would fully address the issue https://github.com/harvester/harvester/issues/988, u may add additional commit of this to the PR, thanks.

guangbochen:
note: with this PR change https://github.com/harvester/harvester/pull/1487, we are currently storing the cloud-init config in the secret resource, will need to backup this resource as well. @FrankYang0529

FrankYang0529:
> note: with this PR change #1487, we are currently storing the cloud-init config in the secret resource, will need to backup this resource as well. @FrankYang0529

@guangbochen @gitlawr I add the synchronization part. Thank you.

pkg/controller/master/backup/backupTaget.go
```diff
@@ -143,11 +154,11 @@ func (h *TargetHandler) updateLonghornTarget(backupTarget *settings.BackupTarget
...
-		"AWS_ACCESS_KEY_ID":     target.AccessKeyID,
-		"AWS_SECRET_ACCESS_KEY": target.SecretAccessKey,
-		"AWS_ENDPOINTS":         target.Endpoint,
-		"AWS_CERT":              target.Cert,
-		"VIRTUAL_HOSTED_STYLE":  strconv.FormatBool(target.VirtualHostedStyle),
+		awsAccessKey:           target.AccessKeyID,
+		awsSecretKey:           target.SecretAccessKey,
+		awsEndpoints:           target.Endpoint,
+		awsCERT:                target.Cert,
+		"VIRTUAL_HOSTED_STYLE": strconv.FormatBool(target.VirtualHostedStyle),
```
guangbochen: It is better to move all strings to the const for better consistency.

pkg/controller/master/backup/backupTaget.go
```diff
@@ -104,6 +115,10 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
 		}
 	}
 
+	if err = h.syncVMBackup(target); err != nil {
+		return nil, err
+	}
+
```
guangbochen: We should either consider moving the backup target validation to the admission webhook or putting the `syncVMBackup` in a separate controller.
This is because syncing the `VMBackups` from the target server should not be bonded to backup target update and it may need to reconcile due to some program errors.

pkg/controller/master/backup/backupTaget.go
```diff
@@ -303,6 +314,49 @@ func (h *TargetHandler) validateS3BackupTarget(target *settings.BackupTarget) er
...
+			},
+			Spec: backupMetadata.BackupSpec,
+			Status: &harvesterv1.VirtualMachineBackupStatus{
+				SourceSpec:    backupMetadata.VMSourceSpec,
+				VolumeBackups: backupMetadata.VolumeBackups,
+				SecretBackups: backupMetadata.SecretBackups,
+			},
+		}); err != nil && !apierrors.IsAlreadyExists(err) {
+			return err
+		}
```
guangbochen: it's tedious to validate the CR existence by creating each CR manually, you should validate against the VMBackup CRs and only create it if not exist.
Also, should also restore its cloud-init secrets.

pkg/controller/master/backup/backupMetadata.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+
+	vmBackupMetadata := &VirtualMachineBackupMetadata{
+		Name:          vmBackup.Name,
+		BackupSpec:    vmBackup.Spec,
+		VMSourceSpec:  vmBackup.Status.SourceSpec,
+		VolumeBackups: vmBackup.Status.VolumeBackups,
+		SecretBackups: vmBackup.Status.SecretBackups,
+	}
```
guangbochen: you should also create the backup VM's related cloud-init secret so that it can be restored on the cluster if not found.

pkg/controller/master/backup/backupTaget.go
FrankYang0529: Updated it. Thank you.

pkg/controller/master/backup/backupTaget.go
FrankYang0529: I move backup target validation to the admission webhook and register a new controller `harvester-backup-metadata-controller`.

pkg/controller/master/backup/backupTaget.go
FrankYang0529: Updated it. Also, I create secrets if there is SecretVolumes in VM backup metadata.

pkg/controller/master/backup/backupMetadata.go
FrankYang0529: Updated it. Thank you.

deploy/charts/harvester/templates/deployment.yaml
```diff
@@ -160,6 +162,9 @@ spec:
         name: harvester-webhook
         args: []
         imagePullPolicy: {{ .Values.webhook.image.imagePullPolicy }}
+        securityContext:
+          capabilities:
+            add: ["SYS_ADMIN"]
```
weihanglo, line 167: Why escalating the permission level?

pkg/controller/master/backup/restore.go
```diff
@@ -577,7 +589,7 @@ func (h *RestoreHandler) deleteOldPVC(vmRestore *harvesterv1.VirtualMachineResto
 
 func (h *RestoreHandler) restartVM(vm *kv1.VirtualMachine) error {
 	logrus.Infof("restarting the vm %s, current state running:%v", vm.Name, *vm.Spec.Running)
-	if vm.Spec.Running == nil || *vm.Spec.Running {
+	if vm.Spec.Running == nil || !*vm.Spec.Running {
```
weihanglo, line 592: why restart a `Running = false` vm?

pkg/controller/master/backup/restore.go
```diff
@@ -79,6 +81,15 @@ func RegisterRestore(ctx context.Context, management *config.Management, opts co
 	pvcs := management.CoreFactory.Core().V1().PersistentVolumeClaim()
 	secrets := management.CoreFactory.Core().V1().Secret()
 
+	copyConfig := rest.CopyConfig(management.RestConfig)
+	copyConfig.GroupVersion = &k8sschema.GroupVersion{Group: "subresources.kubevirt.io", Version: "v1"}
```
weihanglo: NIT
```suggestion
	copyConfig.GroupVersion = &k8sschema.GroupVersion{Group: kv1.SubresourceGroupName, Version: kv1.ApiLatestVersion}
```

deploy/charts/harvester/templates/deployment.yaml
FrankYang0529: We move backup target validation to webhook and leverage [longhorn/backupstore](https://github.com/longhorn/backupstore/blob/master/nfs/nfs.go#L66-L73) to do it. For NFS validation, we need to mount a folder, so we need `SYS_ADMIN` for it.

pkg/controller/master/backup/restore.go
FrankYang0529: For restoring to existent VM, we need existent VM to be stopped first, so we need to restart it.

pkg/controller/master/backup/restore.go

weihanglo:
For case 1 (NFS), the vmbackup was synced backup perfectly, but got below events while restoring VM for that backup. The PV was missing and its related PVC was pending and unbounded indefinitely.

```
default            0s          Warning   ProvisioningFailed              persistentvolumeclaim/restore-vm0-ef260676-283f-4a94-b749-ed4b9b54ee7d-disk-0   failed to provision volume with StorageClass "longhorn-image-9c2v7": error getting handle for DataSource Type VolumeSnapshot by Name vm0-volume-vm0-disk-0-hmruo: error getting snapshot vm0-volume-vm0-disk-0-hmruo from api server: volumesnapshots.snapshot.storage.k8s.io "vm0-volume-vm0-disk-0-hmruo" not found
default            0s          Warning   FailedScheduling                pod/virt-launcher-vm0-n5vfh                                                     0/1 nodes are available: 1 pod has unbound immediate PersistentVolumeClaims.
```

OTOH, backups without the backup-target change can be restored as usual.

pkg/webhook/resources/setting/validator.go
```diff
@@ -117,6 +123,40 @@ func validateVMForceDeletionPolicy(setting *v1beta1.Setting) error {
...
+	if target.Type == settings.S3BackupType && (target.SecretAccessKey == "" || target.AccessKeyID == "") {
+		return nil
+	}
+
+	// Set OS environment variables for S3
+	if target.Type == settings.S3BackupType {
+		os.Setenv(backup.AWSAccessKey, target.AccessKeyID)
+		os.Setenv(backup.AWSSecretKey, target.SecretAccessKey)
+		os.Setenv(backup.AWSEndpoints, target.Endpoint)
+		os.Setenv(backup.AWSCERT, target.Cert)
```
guangbochen, line 151: ditto

pkg/controller/master/backup/backup_target.go
```diff
@@ -143,11 +150,11 @@ func (h *TargetHandler) updateLonghornTarget(backupTarget *settings.BackupTarget
...
 	return map[string]string{
-		"AWS_ACCESS_KEY_ID":     target.AccessKeyID,
-		"AWS_SECRET_ACCESS_KEY": target.SecretAccessKey,
-		"AWS_ENDPOINTS":         target.Endpoint,
-		"AWS_CERT":              target.Cert,
-		"VIRTUAL_HOSTED_STYLE":  strconv.FormatBool(target.VirtualHostedStyle),
+		AWSAccessKey:       target.AccessKeyID,
+		AWSSecretKey:       target.SecretAccessKey,
+		AWSEndpoints:       target.Endpoint,
+		AWSCERT:            target.Cert,
```
guangbochen, line 156: The cert should be retrieved from the [`additional-ca` setting](https://github.com/harvester/harvester/pull/1579/files#diff-bed3ddcc216127b5463aad139741d989834923dc2cf31635427e10d5b228a7f8R167) due to the new [ca-cert changes](https://github.com/harvester/harvester/blob/master/pkg/controller/master/setting/additional_ca.go).

pkg/controller/master/backup/backup.go
```diff
@@ -280,65 +332,275 @@ func (h *Handler) getBackupPVC(namespace, name string) (*corev1.PersistentVolume
...
+	return nil
+}
+
+func (h *Handler) getVolumeSnapshot(namespace, name string) (*snapshotv1.VolumeSnapshot, error) {
+	snapshot, err := h.snapshotCache.Get(namespace, name)
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	} else if err != nil {
+		return nil, err
+	}
```
guangbochen:
```GOLANG
   return nil, err
}
```

pkg/controller/master/backup/backup.go
```diff
@@ -280,65 +332,275 @@ func (h *Handler) getBackupPVC(namespace, name string) (*corev1.PersistentVolume
...
+	return volumeSnapshot, nil
+}
+
+func (h *Handler) createVolumeSnapshotContent(
+	vmBackup *harvesterv1.VirtualMachineBackup,
+	volumeBackup harvesterv1.VolumeBackup,
+	snapshotClass *snapshotv1.VolumeSnapshotClass,
+) (*snapshotv1.VolumeSnapshotContent, error) {
+	logrus.Debugf("attempting to create VolumeSnapshotContent %s", getVolumeSnapshotContentName(volumeBackup))
+	snapshotContent, err := h.snapshotContents.Get(getVolumeSnapshotContentName(volumeBackup), metav1.GetOptions{})
```
guangbochen: get via cache.

pkg/controller/master/backup/backup.go
```diff
@@ -280,65 +332,275 @@ func (h *Handler) getBackupPVC(namespace, name string) (*corev1.PersistentVolume
...
+
+	sc, err := h.snapshotClassCache.Get(settings.VolumeSnapshotClass.Get())
+	if err != nil {
+		return nil, fmt.Errorf("%s/%s VolumeSnapshot requested but no storage class, err: %s",
+			vmBackup.Namespace, volumeBackup.PersistentVolumeClaim.ObjectMeta.Name, err.Error())
+	}
+
+	volumeSnapshotSource := snapshotv1.VolumeSnapshotSource{}
+	// If LonghornBackupName exists, we should create volume snapshot from LH Backup instead of from current PVC.
+	if volumeBackup.LonghornBackupName != nil {
```
guangbochen, line 402: What's the benefit of creating a snapshot from LH backup vs PVC?

pkg/controller/master/backup/backup.go
```diff
@@ -280,65 +332,275 @@ func (h *Handler) getBackupPVC(namespace, name string) (*corev1.PersistentVolume
...
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	return snapshot, nil
+}
+
+func (h *Handler) createVolumeSnapshot(vmBackup *harvesterv1.VirtualMachineBackup, volumeBackup harvesterv1.VolumeBackup) (*snapshotv1.VolumeSnapshot, error) {
```
guangbochen, line 390: u will also need to check if the `volumeSnapshot` exist on restore otherwise the PVC may failed to create with the following case:
1. config the backup target to another one
2. delete vmbackups and its volumesnaphost is not removed due to `bound-protection` finalizer
3. re-config backup target to the original one
4. the VMbackup CR is restored but the volumesnapshot is missing(will be delete after a period of seconds)

pkg/controller/master/backup/util.go
```diff
@@ -165,3 +169,7 @@ func sanitizeVirtualMachineForRestore(restore *harvesterv1.VirtualMachineRestore
 func getCloudInitSecretRefVolumeName(vmName string, secretName string) string {
 	return fmt.Sprintf("vm-%s-%s-volumeref", vmName, secretName)
 }
+
+func getVMBackupMetadataFileName(vmBackupName string) string {
+	return fmt.Sprintf("%s.cfg", vmBackupName)
```
weihanglo: We should consider namespace as well, so `<namespace>-<name>` might be better?

pkg/controller/master/backup/backup.go
FrankYang0529: If VM Backup already has `LonghornBackupName`, it means the VM Backup has associated LH Backup. So we need to create VolumeSnapshot from LH Backup, not from PVC.

weihanglo:
- ✅ Confirmed that case 2 works for me.
- ❌ For case 1 the restored VM always gets `FailedUnschedulable` and the corresponding PV was not provisioned.

pkg/controller/master/backup/backup.go
```diff
@@ -229,38 +283,58 @@ func (h *Handler) getSecretBackupFromSecret(namespace, name string) (*harvesterv
...
+		backupCpy.Annotations = map[string]string{}
+	}
+
+	if backupCpy.Annotations[BackupTargetAnnotation] == "" && backupCpy.Annotations[BackupBucketNameAnnotation] == "" {
+		switch target.Type {
+		case settings.NFSBackupType:
+			backupCpy.Annotations[BackupTargetAnnotation] = target.Endpoint
+		case settings.S3BackupType:
+			backupCpy.Annotations[BackupBucketNameAnnotation] = target.BucketName
+			backupCpy.Annotations[BackupBucketRegionAnnotation] = target.BucketRegion
```
gitlawr: I am wondering if we should set them in vmbackup.status instead of using the annotations.

pkg/controller/master/backup/backup.go
```diff
@@ -47,90 +58,133 @@ func RegisterBackup(ctx context.Context, management *config.Management, opts con
...
+		return nil, h.setStatusError(vmBackup, err)
+	}
+
+	// reconcile backup status of volume backups, validate if those volumeSnapshots are ready to use
+	if err := h.updateConditions(vmBackup); err != nil {
+		return nil, err
+	}
+
+	// generate vm backup metadata and upload to backup target
+	if err := h.uploadVMBackupMetadata(vmBackup, target); err != nil {
```
gitlawr: It seems like we do upload whenever a vmbackup is changed

pkg/controller/master/backup/util.go
FrankYang0529: Updated it. Thank you.

pkg/controller/master/backup/backup.go
FrankYang0529: I check whether `volumeSnapshot` has `DeletionTimestamp`. If yes, we will requeue `vmBackup` again.

pkg/controller/master/backup/backup.go
FrankYang0529: Updated it. Thank you.

pkg/controller/master/backup/backup.go
FrankYang0529: Updated it. Thank you.

pkg/controller/master/backup/backup_target.go
FrankYang0529: Updated it. Thank you.

FrankYang0529:
> * ✅ Confirmed that case 2 works for me.
> * ❌ For case 1 the restored VM always gets `FailedUnschedulable` and the corresponding PV was not provisioned.

Hi @weihanglo, thanks for helping me validate the PR. I fix the issue. May you help me re-validate again? Thank you.

Originally, VolumeSnapshot and VolumeSnapshotContent can't be deleted immediately, because they have finalizers. When we re-config backup target back, finalizers will be passed and VolumeSnapshot/VolumeSnapshotContent will be totally deleted. However, the `deletionPolicy` of out storage class `longhorn` is `delete`, so LH Backup may be deleted too. This will delete LH Backup in backup target too, so we lose LH Backup data and PVC can't be created successfully.

My fix is trying to force delete VolumeSnapshot and VolumeSnapshotContent if the backup target setting is not same as backup target in VM Backup annotation.

pkg/webhook/resources/setting/validator.go
FrankYang0529: Updated it. Thank you.

guangbochen (approved):
LGTM

pkg/controller/master/backup/backup_target.go
```diff
@@ -38,9 +29,18 @@ const (
...
 	longhornBackupTargetSecretSettingName = "backup-target-credential-secret"
+
+	AWSAccessKey       = "AWS_ACCESS_KEY_ID"
+	AWSSecretKey       = "AWS_SECRET_ACCESS_KEY"
+	AWSEndpoints       = "AWS_ENDPOINTS"
+	AWSCERT            = "AWS_CERT"
+	VirtualHostedStyle = "VIRTUAL_HOSTED_STYLE"
+	HttpProxy          = "HTTP_PROXY"
+	HttpsProxy         = "HTTPS_PROXY"
+	NoProxy            = "NO_PROXY"
```
gitlawr: Where do we use these constants

pkg/controller/master/backup/restore.go
gitlawr: I don't get it. Why do we restart a VM when we want it to be stopped?

gitlawr:
>Case 3 - Delete VirtualMachineImage and restore VM. (Not support in this PR. Just leave the test case as a reminder)

Does it mean that we cannot support VM restore to another fresh Harvester cluster?

pkg/controller/master/backup/restore.go
FrankYang0529: Correct me if I am wrong. I think users may want their VM to go back to running state if they restore to an existent VM.

FrankYang0529:
> > Case 3 - Delete VirtualMachineImage and restore VM. (Not support in this PR. Just leave the test case as a reminder)
> 
> Does it mean that we cannot support VM restore to another fresh Harvester cluster?

No, for a fresh Harvester cluster, we need users to create VirtualMachineImage / BackingImage / StorageClass with the same name as the previous cluster. If they can do it, they do it, they still can restore VM. We will document this.

pkg/controller/master/backup/backup.go
FrankYang0529: Yes, if vmbackup is changed, we upload its metadata to make sure metadata is correct.

pkg/controller/master/backup/backup.go
FrankYang0529: I think both vmbackup.status and annotations work for me. I just followed the previous design from this [PR](https://github.com/harvester/harvester/commit/85a12ff72013cdd7b2ba86a1f2607bbb8c367745#diff-75556fad8ea25a4e0f7d5c45bfea2abc49909d662b944d47f1301e0843b97469R276-R286).

https://github.com/harvester/harvester/blob/b7e562d382019d15d5e4a0394b1dfb5afa6fe630/pkg/controller/master/backup/backup.go#L322-L332

pkg/controller/master/backup/backup_target.go
FrankYang0529: Sorry, I forget to delete them. Thank you.

pkg/controller/master/backup/restore.go
gitlawr: That makes sense, but we should call `start` instead of `restart` when it's not running.

FrankYang0529: Updated it. Thank you.

pkg/controller/master/backup/backup.go
gitlawr: We don't need to extend the API via annotation unless it's not under our control. Now that the code depends a lot on it, can you set it in status? cc @guangbochen

pkg/controller/master/backup/restore.go
gitlawr: On the second thought, I think the reason for restarting VM is to recreate the VM instance so that the new VM spec is applied. Is it expected to start a stopped VM instead of restarting a running VM here?

pkg/controller/master/backup/backup.go
guangbochen: +1, set it on the status is a good design to follow, we use it due to some legacy design's followed with the kubevirt, but it is not important now.

pkg/controller/master/backup/restore.go
FrankYang0529: From webhook, I think we can't restore to a running VM. We have to stop it first.

https://github.com/harvester/harvester/blob/b7e562d382019d15d5e4a0394b1dfb5afa6fe630/pkg/webhook/resources/restore/validator.go#L92-L94

pkg/controller/master/backup/backup.go
FrankYang0529: > We don't need to extend the API via annotation unless it's not under our control. Now that the code depends a lot on it, can you set it in status? cc @guangbochen


Good catch! @guangbochen, do you think we update the design in this PR? Or we put it to the next PR?

guangbochen: Can we change it in this PR? also the front-end will need to sync its change with this one cc @WuJun2016, thanks.

WuJun2016: What should be modified on the front-end ？

pkg/controller/master/backup/backup.go
gitlawr: I think we only need to upload once when vmbackup is ready. Why do we want to upload intermediate states?

FrankYang0529: Okay, I will move it after `isBackupReady`.

FrankYang0529: Updated it. Thank you.

pkg/controller/master/backup/backup.go
FrankYang0529: @guangbochen @gitlawr I move backup target data from annotations to status and run test cases 1 & 2. It works fine. Thank you.

@WuJun2016 In Backup page, front-end determines Backup Target from VM Backup annotations. Now, we move this information to VM Backup status. Thank you.

pkg/controller/master/backup/backup.go
```diff
@@ -93,44 +110,72 @@ func (h *Handler) OnBackupChange(key string, vmBackup *harvesterv1.VirtualMachin
...
+	target, err := settings.DecodeBackupTarget(settings.BackupTargetSet.Get())
+	if err != nil {
+		return nil, err
+	}
+
 	if isBackupReady(vmBackup) {
-		return nil, nil
+		// generate vm backup metadata and upload to backup target
+		return nil, h.uploadVMBackupMetadata(vmBackup, target)
 	}
```
gitlawr, line 121: We may still do upload multiple times. For example when Harvester server restarts and syncs all the vmBackups. All the ready backups will get uploaded again. Can we have a condition to ensure it is uploaded once?

FrankYang0529: How about we check whether remote data exists? If it exists and the data is same, we skip upload here.

gitlawr: That may work too. I'm not sure how much effort is needed for the check

FrankYang0529: Updated it. Thank you.

gitlawr (approved):
LGTM

Issue #571, opened 2021-03-25
[FEATURE] Create a template from VM, including root and data disks


yasker:
**Is your feature request related to a problem? Please describe.**
Sometimes the user would want to create a template that includes all the states of the VM and use the template to create copies of the VM in the future.

**Describe the solution you'd like**
Support create a VM template with data from an existing VM.

**Describe alternatives you've considered**
VM clone (#569) is another way to solve a similar problem, but it won't work once all the copies are down.

**Additional context**
Depends on #568

guangbochen:
Create a template from VM with includes its root and data disks means users can restore one or multiple new VMs from the template even if the original VM was removed.

With the above condition, I think we can leverage the existing [Create image from volume](https://github.com/harvester/harvester/issues/568) feature to persist `root` disk and `data` disks respectively. 

The only concern that I have is about the disk space availability of those backing image volumes created the VM templates, WDUT cc @shangma @yasker @shuo-wu

yasker:
Yes. We need to create new images from existing VMs to make this feature work. @shuo-wu do we have capacity check for base images?

shuo-wu:
When a Longhorn volume is launched **with a backing image**, the new backing image exported/created from it won't waste space to store zero data **only if the type is qcow2**. But there is one more potential issue besides this: If the upper layer/filesystem writes a file then removed it, the corresponding **sectors were used by the file cannot be reclaimed as free after the file removal**. Then meaningless sector data will be consumed extra space when the volume is exported to a backing image.

e.g.:
1. Upload an empty backing image with apparent size 1Gi. Notice that both the apparent size & actual size of the backing image is 1Gi after upload (I may need to check this part).
```
### before upload, in local machine
~ mkfs.ext4 bi-upload.raw
mke2fs 1.46.1 (9-Feb-2021)
Creating filesystem with 262144 4k blocks and 65536 inodes
Filesystem UUID: c2f3498b-2a71-4eb7-8ee7-02c79d032d70
Superblock backups stored on blocks:
	32768, 98304, 163840, 229376

Allocating group tables: done
Writing inode tables: done
Creating journal (8192 blocks): done
Writing superblocks and filesystem accounting information: done

~ ls -alh bi-upload.raw
-rw-r--r--  1 shuo_wu  staff   1.0G Oct 15 12:36 bi-upload.raw

~ du -h bi-upload.raw
 49M	bi-upload.raw
```
```
### after upload, in the cluster
[...]

PR #1452, opened 2021-10-20
genearate image from volume while creating template

shangma:
harvester issue #571


shangma:
@gitlawr @guangbochen  I understand that for issue #568, when we export volume to an image, we made the ui to ask for a namespace as an argument so that we can export the image to a different ns, but here one vm may have multiple pvc. So instead of asking a namespace argument for each of the pvc, I use the namespace of the vm itself. Let me know if this is appropriate. The code also assume that the pvc owned by a vm will share the same namespace. Let me know if this makes sense.

gitlawr:
Make sense. The VM and PVCs should be in the same namespace

shangma:
Add "To Test" section in the original [issue](https://github.com/harvester/harvester/issues/571#issuecomment-948086400).

pkg/apis/harvesterhci.io/v1beta1/template.go
```diff
@@ -68,7 +68,7 @@ type VirtualMachineTemplateVersionSpec struct {
 	Description string `json:"description,omitempty"`
 
 	// +optional
-	ImageID string `json:"imageId,omitempty"`
+	ImageID []string `json:"imageId,omitempty"`
```
gitlawr, line 71: Any impact on upgrade with this API change?

pkg/api/vm/handler.go
```diff
@@ -428,6 +454,7 @@ func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTem
 				Description: fmt.Sprintf("Template drived from virtual machine [%s]", vmID),
 				VM:          removeMacAddresses(vm),
 				KeyPairIDs:  keyPairIDs,
+				ImageID:     imageIDs,
```
gitlawr, line 457: When we create a VM from the template, how do we know the mapping of the images and the volumes

shangma: @gitlawr  how about we use a map with a format such as this {"disk-0": "generated-image-id-0"; "disk-1": "generated-image-id-1"} ?

pkg/apis/harvesterhci.io/v1beta1/template.go
shangma: @gitlawr  I did need to rebuild the webhook image to allow this changes. Other that, the change was functional, see screenshot [here](https://github.com/harvester/harvester/issues/571#issuecomment-948086400).

gitlawr: I mean how do we handle legacy resources that use string for ImageID in the previous version

pkg/api/vm/handler.go
gitlawr: I think we can set the volumeClaimTempaltes annotation: https://github.com/harvester/harvester/blob/d05a6cc730feccd7e242798b8b23973d709ce7e5/pkg/data/template.go#L121

guangbochen:
Close as this is resolved in another PR https://github.com/harvester/harvester/pull/3097; thanks for the contribution.

Issue #1260, opened 2021-09-22
[FEATURE] Configuring additional trusted CA certificates


bashofmann:
**Is your feature request related to a problem? Please describe.**
It should be possible to configure additional trusted CA certificates for all components that communicate outside of the cluster. In Harvester this affects for example image downloads which could be hosted by web servers that are secured with TLS certificates from a corporate CA or where a proxy or firewall intercepts TLS.

**Describe the solution you'd like**
Similarly to Rancher (https://rancher.com/docs/rancher/v2.6/en/installation/install-rancher-on-k8s/chart-options/#additional-trusted-cas) it should be possible to configure additional trusted CAs. Either with a config/helm chart option during installation, or as a global setting.
These certificates should be picked up by all components that are communicating with the outside, probably also the integrated rancher and longhorn (external backups).

yasker:
Ref #761

yasker:
From #761 , we can also consider automatically generate SSL certificates using e.g. Let's Encrypt.

alexdepalex:
@janeczku

gitlawr:
To test:
1. set up a self-signed HTTP server to serve virtual machine images.
2. set up a self-signed s3 server(e.g., https://longhorn.io/docs/1.2.0/snapshots-and-backups/backup-and-restore/set-backup-target/#set-up-a-local-testing-backupstore)
3. Configure additional trusted CA certificates in Harvester settings
4. Verify image download/VM backup works using those servers.

lanfon72:
### Verify items
- [x] image download with self-signed additional-ca (IP-based)
- [x] VM backup with self-signed additional-ca (IP-based)
    - [x] ~~blocked by #1562~~
- [ ] ~~image download with self-signed additional-ca (domain-based)~~  won't test due to it actually same as IP-based
- [ ] ~~VM backup with self-signed additional-ca (domain-based)~~  won't test due to it actually same as IP-based


Test Information
----
* Environment: **qemu/KVM** 2 nodes
* Harvester Version: **master-f5bde7fa-head**
* **ui-source** Option: **Auto**

### Verify Steps:
**Image Download**
1. Install Harvester with ipxe-example which includes https://github.com/harvester/ipxe-examples/pull/36
2. Upload any valid iso to **pxe-server**'s `/var/www/`
3. Use Browser to access `https://<pxe-server-ip>/<iso-file>` should be valid
4. Add self-signed cert to Harvester
[...]

PR #1484, opened 2021-10-26
Support configuring additional trusted CA

gitlawr:
**Related Issue:**
https://github.com/harvester/harvester/issues/1260

**Test plan:**
1. Set up self-signed-cert HTTP/S3 servers
2. Configure CA cert in Harvester settings
3. Verify image download and VM backup/restore to the self-signed-cert server works

Notes:
This is based on the work in PR https://github.com/harvester/harvester/pull/1337
With the setting available we can remove the certificate option in backup target setting in the UI.

### DIFF
--- a/pkg/controller/master/setting/additional_ca.go
+++ b/pkg/controller/master/setting/additional_ca.go
@@ -1,0 +1,63 @@
+package setting
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+func (h *Handler) syncAdditionalTrustedCAs(setting *harvesterv1.Setting) error {
+	// Add envs to the backup secret used by Longhorn backups
+	backupConfig := map[string]string{
+		"AWS_CERT": setting.Value,
+	}
+	if err := h.updateBackupSecret(backupConfig); err != nil {
+		return err
+	}
+
+	// Distribute CA secrets to required system namespaces
+	if err := h.syncAdditionalCASecrets(setting); err != nil {
+		return err
+	}
+
+	//redeploy system services. The CA certs will be injected by the mutation webhook.
+	return h.redeployDeployment(h.namespace, "harvester")
+}
+
+func (h *Handler) syncAdditionalCASecrets(setting *harvesterv1.Setting) error {
+	namespaces := []string{h.namespace, util.LonghornSystemNamespaceName}
+
+	for _, namespace := range namespaces {
+		secret, err := h.secretCache.Get(namespace, util.AdditionalCASecretName)
+		if errors.IsNotFound(err) {
+			newSecret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      util.AdditionalCASecretName,
+					Namespace: namespace,
+				},
+				Data: map[string][]byte{
+					util.AdditionalCAFileName: []byte(setting.Value),
+				},
+			}
+
+			if _, err := h.secrets.Create(newSecret); err != nil {
+				return err
+			}
+			continue
+		} else if err != nil {
+			return err
+		}
+
+		toUpdate := secret.DeepCopy()
+		toUpdate.Data = map[string][]byte{
+			util.AdditionalCAFileName: []byte(setting.Value),
+		}
+		if _, err := h.secrets.Update(toUpdate); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/pkg/controller/master/setting/register.go
+++ b/pkg/controller/master/setting/register.go
@@ -27,6 +27,7 @@ func Register(ctx context.Context, management *config.Management, options config
 	}
 
 	syncers = map[string]syncerFunc{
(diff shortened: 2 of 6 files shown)


pkg/util/proxy.go
```diff
@@ -0,0 +1,33 @@
...
+
+var builtInNoProxy = []string{
+	"localhost",
+	"127.0.0.1",
+	"0.0.0.0",
+	"10.0.0.0/8",
+	"cattle-system.svc",
+	".svc",
+	".cluster.local",
+}
```
guangbochen: longhorn has includes the `192.168.0.0/16 (internal IPs in the cluster)` as one of its default no_proxy config, IIRC this is used by some k8s providers like `kubeadm` so we don't need to include this one?

pkg/controller/master/setting/additional_ca.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+func (h *Handler) syncAdditionalTrustedCAs(setting *harvesterv1.Setting) error {
+	// Add envs to the backup secret used by Longhorn backups
+	backupConfig := map[string]string{
+		"AWS_CERT": setting.Value,
+	}
+	if err := h.updateBackupSecret(backupConfig); err != nil {
+		return err
+	}
```
guangbochen, line 19: user was already able to specify the backup target's `AWS_CERT` via `setting -> backupTarget`
image

pkg/controller/master/setting/additional_ca.go
```diff
@@ -0,0 +1,63 @@
...
+	if err := h.syncAdditionalCASecrets(setting); err != nil {
+		return err
+	}
+
+	//redeploy system services. The CA certs will be injected by the mutation webhook.
+	return h.redeployDeployment(h.namespace, "harvester")
+}
+
+func (h *Handler) syncAdditionalCASecrets(setting *harvesterv1.Setting) error {
+	namespaces := []string{h.namespace, util.LonghornSystemNamespaceName}
```
guangbochen, line 31: what about the `cattle-system` ns?

pkg/util/proxy.go
gitlawr: I think it's an example CIDR in longhorn docs, not the default. It is environment-specific and we don't have those providers in Harvester. Users can configure that anyway.

pkg/controller/master/setting/additional_ca.go
gitlawr: As in the PR description:
>With the setting available we can remove the certificate option in backup target setting in the UI.

I suggest removing this option. It would be confusing for users to have multiple configurations the CA certs.

pkg/controller/master/setting/additional_ca.go
gitlawr: I don't think adding it to rancher server is useful at the moment given that we don't expose the embedded rancher service.

pkg/controller/master/setting/additional_ca.go
guangbochen: Can users add multi-CAs to the additional CA filed, if not I would prefer to keep the current backup target CA config as the additional SSL CA may not be valid for the backup server, e.g using a custom SSL cert of the AWS but in general, users may still want to have config other additional CA of its private server.

guangbochen:
@bk201 can u PTAL of this PR and check if this has anything impact on the OS upgrade, thanks

bk201:
> @bk201 can u PTAL of this PR and check if this has anything impact on the OS upgrade, thanks

I try modifying the image tags of harvester and harvester-webhook, everything looks fine. There will be a running webhook pod to do the mutation for new harvester pods during the upgrade. The HTTP proxy env variables are set after the upgrade.

Or did I misunderstand your concern?

pkg/controller/master/setting/additional_ca.go
```diff
@@ -0,0 +1,63 @@
...
+		return err
+	}
+
+	// Distribute CA secrets to required system namespaces
+	if err := h.syncAdditionalCASecrets(setting); err != nil {
+		return err
+	}
+
+	//redeploy system services. The CA certs will be injected by the mutation webhook.
+	return h.redeployDeployment(h.namespace, "harvester")
```
bk201, line 27: @gitlawr  The mutation hook also checks pods with the label `"longhorn.io/component": "backing-image-data-source"`.
Does anything else need to be restarted here?

gitlawr: They are ephemeral like job pods

guangbochen:
> > @bk201 can u PTAL of this PR and check if this has anything impact on the OS upgrade, thanks
> 
> I try modifying the image tags of harvester and harvester-webhook, everything looks fine. There will be a running webhook pod to do the mutation for new harvester pods during the upgrade. The HTTP proxy env variables are set after the upgrade.
> 
> Or did I misunderstand your concern?

Nope, that's correct, thanks.

guangbochen:
@gitlawr can u please help to rebase the PR, thanks.

pkg/controller/master/setting/additional_ca.go
gitlawr: Users can concatenate multiple certificates

gitlawr:
Rebased

pkg/controller/master/setting/additional_ca.go
guangbochen: cc @WuJun2016 with this PR merge, we can remove the `Certificate` filed of the backup target config(optionally u can add a link to ask to add additional CA by jumping to the additional CA setting page.)

guangbochen (approved):
LGTM

Issue #1014, opened 2021-07-09
[BUG] Trying to set maintenance mode on the last available node shouldn't be allowed


ldevulder:
**Describe the bug**
On a 5 nodes cluster where 4 nodes are maintenance/cordon mode I tried to set the last mode in maintenance and so the cluster tries to migrate 2 VMs, which is not possible as there is no free worker. So the `Migrating` state for VMs stay after hours and the node as the same issue with `Entering maintenance mode` state. And, of course, the VMs that are in `Migrating` state are not available.

**To Reproduce**
1. Create some VM on a cluster with at least 3 nodes
2. Set 2 nodes in maintenance/cordon state
3. Tries to set maintenance mode on the remaining node

**Expected behavior**
We shouldn't be able to try to set the remaining node in maintenance mode, or at least a warning should be shown.

**Support bundle**
FYI: I was not able to generate a support bundle for this issue: it stucks at 0%... Is it because all nodes are in maintenance/cordon mode? Is a another issue needed for this?

**Environment:**
 - Harvester ISO version: `master-902bee60-head`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): HP ProLiant DL360 Gen9

**Additional context**
None.

ldevulder:
After a quick discussion on slack here a comment from @yasker :
```
I feel maintenance should be allowed but we should have a warning showing that migrating all the running VMs
won't be possible. Since this will happen not only for the last node, but also if the cluster is at full capacity.
We definitely shouldn't shutdowns all the VMs automatically.
```

FrankYang0529:
@guangbochen do we need to consider whether a node is a server or agent for this issue? For example, we restrict users to enable maintenance mode on the last server node if there is only one server node or on the last 2-server node if there is a cluster.

TachunLin:
Verified **behavior as expected** for 3 nodes scenario on master-b5dadc82-head 
We need further confirm how 2 nodes expected behavior should be as above comment

### Result
Within 3 nodes and 3 virtual machines testing environment.
Set maintenance from node 3 -> node 2 -> node 1 in sequence 
The final remaining node will prompt `can't enable maintenance in last available node `

![image](https://user-images.githubusercontent.com/29251855/140376151-a93b69f5-a3d6-4120-b754-c39aec4feadf.png)

![image](https://user-images.githubusercontent.com/29251855/140376294-4bf9d0a1-bcf6-4908-b286-0f79f3d9003b.png)


[...]

PR #1490, opened 2021-10-28
add node admission webhook

FrankYang0529:
**Problem:**
The last available node can enable maintenance mode.

**Solution:**
Add a node admission webhook to restrict that.

**Related Issue:**
https://github.com/harvester/harvester/issues/1014

**Test plan:**

The last available node can't be marked as maintenance mode.

* Enable maintenance mode on the agent node.
* Enable maintenance mode on the server node. Check it is a bad request and the server node can't enable maintenance mode.

The maintenance mode can be disabled.

* Follow the test case above.
* Disable maintenance mode on agent node should not get error and maintenance mode on the agent node will be removed.

The server node can enable maintenance mode if there is another available node.

* Follow the test cases above.
* Enable maintenance mode on the server node. Check it can turn into maintenance mode.

**Other information:**
It looks like frontend can't show webhook error message on the Hosts page. We may need another PR from frontend.
Kapture 2021-10-28 at 17 24 44

### DIFF
--- a/pkg/webhook/resources/node/validator.go
+++ b/pkg/webhook/resources/node/validator.go
@@ -1,0 +1,73 @@
+package node
+
+import (
+	v1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	ctlnode "github.com/harvester/harvester/pkg/controller/master/node"
+	werror "github.com/harvester/harvester/pkg/webhook/error"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+func NewValidator(nodeCache v1.NodeCache) types.Validator {
+	return &nodeValidator{
+		nodeCache: nodeCache,
+	}
+}
+
+type nodeValidator struct {
+	types.DefaultValidator
+	nodeCache v1.NodeCache
+}
+
+func (v *nodeValidator) Resource() types.Resource {
+	return types.Resource{
+		Name:       "nodes",
+		Scope:      admissionregv1.ClusterScope,
+		APIGroup:   corev1.SchemeGroupVersion.Group,
+		APIVersion: corev1.SchemeGroupVersion.Version,
+		ObjectType: &corev1.Node{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Update,
+		},
+	}
+}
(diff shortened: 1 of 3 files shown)


guangbochen:
cc @WuJun2016 will need to add showing the error webhook message properly on the host page.

pkg/webhook/resources/node/validator.go
```diff
@@ -0,0 +1,70 @@
...
+	nodeList, err := v.nodeCache.List(labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	nodeMaintainCount := 0
+	for _, node := range nodeList {
+		if _, ok := node.Annotations[ctlnode.MaintainStatusAnnotationKey]; !ok {
+			nodeMaintainCount++
+		}
```
guangbochen: both maintenance nodes and the cordoned nodes should be treated as un-schedulable.
https://kubernetes.io/docs/concepts/architecture/nodes/#condition

pkg/webhook/resources/node/validator.go
```diff
@@ -0,0 +1,70 @@
...
+
+	nodeList, err := v.nodeCache.List(labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	nodeMaintainCount := 0
+	for _, node := range nodeList {
+		if _, ok := node.Annotations[ctlnode.MaintainStatusAnnotationKey]; !ok {
+			nodeMaintainCount++
```
gitlawr: Instead of traversing all nodes, we can return once we find one available node.

pkg/webhook/resources/node/validator.go
FrankYang0529: Updated it. Thank you.

pkg/webhook/resources/node/validator.go
FrankYang0529: Updated it. Thank you.

pkg/webhook/resources/node/validator.go
```diff
@@ -0,0 +1,73 @@
...
+		if node.Name == oldNode.Name {
+			continue
+		}
+
+		// Return when we find another available node
+		if _, ok := node.Annotations[ctlnode.MaintainStatusAnnotationKey]; !ok && !node.Spec.Unschedulable {
+			return nil
+		}
+	}
+	return werror.NewBadRequest("can't enable maintenance mode or cordon on latest available node")
```
gitlawr: s/latest/the last/

FrankYang0529: Thank you! Updated it.

gitlawr (approved):
LGTM

guangbochen (approved):

Issue #1241, opened 2021-09-16
[FEATURE] Disk devices used for VM storage should be globally configurable


janeczku:
The user should be able to specify, during the installation, (e.g. as part of the create-config or join-config) the list of disk devices (e.g. by device path) that should be used for VM storage capacity across all nodes (given they share a consistent hardware configuration of course).
This should also cover the ability to exclude the root disk from the VM storage. Use case for excluding the root disk include:

-  use of a (cheaper) small capacity root disk
-  ensure isolation of IO resources between Harvester critical processes and VM workloads.

The current approach of having to configure additional disks post-install via the ui/api for each and every Harvester server separately does not work well for automated rollouts.

**Possible Implementation Details:**

The new option should be both supported in the create-config and join-config (the latter overriding the value specified in the former).
The option could either be a list of device paths (e.g. /dev/nvme0n1,/dev/nvme0n2) or a regex matcher.
When this option is specified by the user, the OS disk device configured in the install.device option is no longer added to the LH storage unless it's specifically included via the new option.

janeczku:
The preferred implementation would be a configuration option that allows the user to define a conclusive list of all disks to use for LH storage where each list item can be either a regex expression or a fully qualified device name.

`storage_devices`: `/dev/nvme0*, /dev/sdb`

johnliu55tw:
To add additional disks to Harvester right after installation, I'm thinking of utilizing Node Disk Manager (NDM):
- During the installation, create GPT with one EXT4 partition **with a certain label (e.g. `HARV_DISK_*`)** for those additional disks user specified.
- After installation and Harvester is up and running, NDM will scan the block devices. If it sees the partition has a label `HARV_DISK_*`, it automatically adds necessary spec fields (`provisioned`, `forceFormatted`, `mountPoint`, ...) and create corresponding CRD
- The disk should be added to Harvester during the reconciliation of NDM controller

@weihanglo @guangbochen please help to check if this is a feasible way to add disks automatically, thanks!

johnliu55tw:
**2021-10-26 Update**: No need to add this functionality to the Installer UI. For now we could only implement this feature for PXE/Remote config installation.

johnliu55tw:
# High-level design
## Some notes
- Seems like Longhorn always expects one default path to exist. We can change the path, but cannot disable it during the installation [REF](https://longhorn.io/docs/1.2.2/volumes-and-nodes/multidisk/#configuration)
[...]

PR #1491, opened 2021-10-28
Load settings provided by installer

johnliu55tw:
Add settings controller to load setting into Setting CR, and restart workloads if needed.

* Add immutable ConfigMap "installer-provided-settings" to store settings provided by the installer. This ConfigMap is updated by installer using HelmChartConfig value `service.installerProvidedSettings`.

* Add a new setting CR "LoadSettingsFromInstaller" to indicate whether the controller should update settings from the installer-provided values (the ConfigMap).

* Add controller reconcile function "LoadSettingsFromInstallerOnChanged" to update settings from installer-provided values. It only updates once during the Helm chart deployment by manipulating the "LoadSettingFromInstaller" CR.

* Update NDM DaemonSet when settings changed. If users change settings related to NDM, update the DaemonSet to ensure pods are reloaded. Currently only "AutoAddDiskPaths" setting is associated with NDM.

See https://github.com/harvester/harvester-installer/pull/158 for how to add settings from Installer.

**Related issues**
- #1241
- https://github.com/harvester/harvester-installer/pull/158

### DIFF
--- a/deploy/charts/harvester/templates/configmap.yaml
+++ b/deploy/charts/harvester/templates/configmap.yaml
@@ -64,3 +64,16 @@ data:
       sleep 2
     done
     `}}
+
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: installer-provided-settings
+  labels:
+{{ include "harvester.labels" . | indent 4 }}
+immutable: true
+data:
+{{- range $key, $value := .Values.service.installerProvidedSettings }}
+  {{ $key }}: {{ $value | printf "%q" -}}
+{{ end }}
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -260,6 +260,7 @@ service:
     mode: "static"
     ip: ""
     hwAddress: ""
+  installerProvidedSettings: {}
 
 ## Specify the lifecycle jobs.
 ##
--- a/pkg/controller/master/setting/controller.go
+++ b/pkg/controller/master/setting/controller.go
@@ -1,13 +1,29 @@
 package setting
 
 import (
+	appsv1 "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
+	corev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	harvctlv1beta1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	systemSetting "github.com/harvester/harvester/pkg/settings"
+)
+
+const (
+	harvesterSystemNamespace = "harvester-system"
+	installerSettingsCmName  = "installer-provided-settings"
 )
 
 // Handler updates the log level on setting changes
 type Handler struct {
+	dsClient          appsv1.DaemonSetClient
+	dsCache           appsv1.DaemonSetCache
+	cmClient          corev1.ConfigMapClient
+	cmCache           corev1.ConfigMapCache
+	settingController harvctlv1beta1.SettingController
+	settingCache      harvctlv1beta1.SettingCache
 }
 
 func (h *Handler) LogLevelOnChanged(key string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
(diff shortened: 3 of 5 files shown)


deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -64,3 +64,16 @@ data:
...
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: installer-provided-settings
+  labels:
+{{ include "harvester.labels" . | indent 4 }}
+immutable: true
+data:
+{{- range $key, $value := .Values.service.installerProvidedSettings }}
```
bk201, line 77: Is there a better alternative to the `installerProvided` term?
Like `default` or `preConfigured`.  The term `installer` is too specific.
I can load the chart and provide some default settings and the source doesn't have to be the installer.

johnliu55tw: What about `userSettings`? This indicates that these are provided by the user.

gitlawr:
We already have the settings CRD. It doesn't look right to store configurations in configmaps then update the settings CRs.
I can come up with two ways to configure Harvester settings during installation:

1. Support updating chart values in harvester config. Then we can set it in Harvester server envs. We have [the logic](https://github.com/harvester/harvester/blob/b5dadc82329fec7d631055e83d2af3a2e3fb46fb/pkg/settings/settings.go#L129) for reading settings from envs.
e.g., 
```
containers:
  apiserver:
    env:
    - name: HARVESTER_AUTO_ADD_DISK_PATHS
      value: /foo
```
2. Creates Harvester settings CRs directly.

bk201:
> 1. Creates Harvester settings CRs directly.

@gitlawr will a setting be reverted by helm if the user edits it?

rancher uses env variables and SUC/longhorn load default values from ConfigMap.

johnliu55tw:
> 1. Support updating chart values in harvester config. Then we can set it in Harvester server envs. We have [the logic](https://github.com/harvester/harvester/blob/b5dadc82329fec7d631055e83d2af3a2e3fb46fb/pkg/settings/settings.go#L129) for reading settings from envs.
> e.g.,
> 
> ```
> containers:
>   apiserver:
>     env:
>     - name: HARVESTER_AUTO_ADD_DISK_PATHS
>       value: /foo
> ```

In this case, whenever users updated the Setting CR, the controller needs to sync the setting back to the `env` of harvester's Deployment or remove the variable from `env` entirely, to prevent loading old settings if Pods restarted. I feel like the controller manipulating its own environment variables doesn't sound right...?

> 2. Creates Harvester settings CRs directly.

Then I would probably create a Job and leverage [RKE2 Auto-Deploying Manifests](https://docs.rke2.io/advanced/#auto-deploying-manifests) to run a script **only once** after the Chart deployment to overwrite Setting CR. I've done some PoC for this case and it seems to work. Adding Setting CR before controller is deployed might cause it to be reverted by the controller?

And one of the benefits of using an immutable ConfigMap is that we have a clear picture of what users have changed.

gitlawr:
Yes we should avoid manipulating the envs. If these are default values and we expect users to change them afterwards, I think we can create setting CRs. By setting the default value and leaving the value empty it should not be reverted by helm or controllers(and we can tell the change when value!=default). @johnliu55tw Can you check if that's the case?

johnliu55tw:
For now, restarting the controller would revert the setting CR back to the default value set by [settings.go](https://github.com/harvester/harvester/blob/b5dadc82329fec7d631055e83d2af3a2e3fb46fb/pkg/settings/settings.go#L12). Steps:
1. Update `log-level` setting using `kubectl` with the following resource:
```
kind: Setting
metadata:
  name: log-level
default: debug
```
2. Verify the setting `log-level`'s `default` value becomes `debug`
3. Restart the deployment "harvester"
4. Check the setting `log-level` again, it's reverted to `info`

And I've observed another issue: If we put the setting CR in `/var/lib/rancher/rke2/server/manifest/` for auto-deploying ([REF](https://docs.rke2.io/advanced/#auto-deploying-manifests)), it will be re-applied by RKE2 again after a reboot, thus overwriting the current setting.

I guess we could change controller's behavior to **not** overwriting the setting if it already exists, but the RKE2 re-applying issue is not easy to solve. cc @gitlawr

guangbochen:
don't change the default value of the setting but use the value instead. The RKE2 `manifest` directory is by design to overwrite with its default value so u need to modify the manifest file if u want to update the content. 
```
kind: Setting
metadata:
  name: log-level
default: debug
value: info
```

johnliu55tw:
> don't change the default value of the setting but use the value instead. The RKE2 `manifest` directory is by design to overwrite with its default value so u need to modify the manifest file if u want to update the content.

I think this means the controller needs to access the file on the Node using `hostPath` volume, in order to sync Setting CR back to the manifest file? In this case, I'd say an immutable ConfigMap would be a bit simpler.

johnliu55tw:
Close due to design changes. See https://github.com/harvester/harvester/pull/1515 for the new design.

Issue #1429, opened 2021-10-14
[BUG] Harvester 0.3 does not allow CPU-overprovisioning


ugar-ugar:
**Describe the bug**
 Harvester 0.3 does not allow CPU-overprovisioning

**To Reproduce**
1. Create a 8 Core Harvester 0.3 node
2. Create a 4 Core VM on this node
3. VM will not start with "insufficient CPU" as hint

**Expected behavior**
a 4 Core VM should run on a 8 core node

**Support bundle**
is attached

**Environment:**
 - Harvester ISO version: 0.30 
 - Underlying Infrastructure:
cluster with 3x  Virtual Machine with 8 cores, 16 GB RAM, 128 GB Disk
[supportbundle_0463918b-0ee5-438f-880d-0f87183552f0_2021-10-14T15-52-18Z.zip](https://github.com/harvester/harvester/files/7347576/supportbundle_0463918b-0ee5-438f-880d-0f87183552f0_2021-10-14T15-52-18Z.zip)
 Virtual Machine with 8 cores, 16 GB RAM, 128 GB Disk
.

lanfon72:
ref to #1376

yasker:
We need to support overprovisioning by default.

weihanglo:
FWIW there are some interesting configs in KubeVirt. 

For CPU overcommitment, there are two related cluster-wide config on KubeVirt CR

- `spec.configuration.cpuRequest`: set the global default CPU request [if a request is send without specifying CPU][cpurequest]. Default to 100m
-  `spec.configuration.developerConfiguration.cpuAllocationRatio`: a ratio that [controls CPU allocation amount][cpuallocratio]. 10 means allocates 1/10 of the request. Default to 10.

For memory there are both cluster-wide and instance-level config. We can set

- `spec.configuration.developerConfiguration.MemoryOvercommit` to [control cluster-wide memory resource request][memoryovercommit]. It's also a ratio, 100% means committing exactly the same amount that guest VM uses. 200% mean the guest VM would request half of it actually uses.

There is also per-instance memory overcommit setting but not for CPU at this moment. Please refer to KubeVirt official guide ["Node overcommit"](https://kubevirt.io/user-guide/operations/node_overcommit/).

[cpurequest]: https://github.com/kubevirt/kubevirt/blob/b6156e522dc1d77fefd57564c09cf20803a64c73/pkg/virt-api/webhooks/mutating-webhook/mutators/vmi-mutator.go#L339-L348
[...]

PR #1492, opened 2021-10-28
Add resource overcommit feature

weihanglo:
**Problem:**
Add overcommit setting and mutating/validating admission webhook for CPU, memory and storage

**Solution:**

Introduce a new harvester setting `overcommit-config` with three keys. Each key controls different resources overcommit percentage:

- `cpu`: controls VM's request/limit on CPU via mutating admisson webhook. Default to 200%.
- `memory`: controls VM's request/limit on memory via mutating admisson webhook. Default to 200%.
- `storage`: proxies to longhorn setting resource `storage-over-provisioning-percentage`. Default to 200%.

A mutating admission webhook  is added because KubeVirt only applies that its own overcommit feature on VM without any specified `resources.limits` or `resources.requests`. Harvester here handles the case which only specifies `resources.requests`. In that case, the mutating webhook simply sets the requests as limits and use the limits to calculate overcommited resources usage as the new requested values.

**Related Issue:**
 #1429 

**Test plan:**
1. Build the new harvester and run it in cluster
1. `kubectl apply -f` the content below
    ```yaml
    kind: Setting
    metadata:
      name: overcommit-config
    status: {}
    value: '{"cpu":300,"memory":300,"storage":300}'
    ```
1. Head to `Advanced > Settings` and change ui-index to `https://weihanglo-harv-dashboard.s3.us-east-2.amazonaws.com/dashboard/index.html`
1. Create a VM with more core than your host machine has. It should succeed.
1. Check the pod of the VM. Propably named `virt-launcher-xxxx` in default namespace. It's CPU/memory request should be lower than what you previously set during VM creation.


weihanglo:
The modified dashboard is based on this commit: https://github.com/weihanglo/dashboard/commit/b0b43f1765fb68fd8d49d73bdd3a3a01d6059a53

weihanglo:
There is a small caveat in UI that if your memory usage is not divisible by MiB. It'll show KiB or Bytes instead. And that reduces readability. Actually the column should display `resources.limits` instead so it is more like an UI issue.
image

pkg/controller/master/setting/controller.go
```diff
@@ -24,3 +52,75 @@ func (h *Handler) LogLevelOnChanged(key string, setting *harvesterv1.Setting) (*
...
+
+func (h *Handler) cpuOvercommitPercentageOnChanged(setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	percentage, err := strconv.ParseInt(setting.Value, 10, 64)
+	if percentage < 100 {
+		// You can only overcommit not undercommit
+		return setting, fmt.Errorf("Cannot undercommit. Should be greater that 100 but got %d", percentage)
+	}
+	if err != nil {
+		return setting, err
+	}
```
bk201: Should we check the error first?

weihanglo: Fixed and rebased.

pkg/settings/settings.go
```diff
@@ -29,6 +29,9 @@ var (
 	SupportBundleImage           = NewSetting("support-bundle-image", "rancher/support-bundle-kit:v0.0.3")
 	SupportBundleImagePullPolicy = NewSetting("support-bundle-image-pull-policy", "IfNotPresent")
 	DefaultStorageClass          = NewSetting("default-storage-class", "longhorn")
+	CpuOvercommitPercentage      = NewSetting("cpu-overcommit-percentage", "1000")    // KubeVirt's default is "10"x ratio. To align with others, we change it to percentage.
+	MemoryOvercommitPercentage   = NewSetting("memory-overcommit-percentage", "100")  // KubeVirt's default
+	StorageOvercommitPercentage  = NewSetting("storage-overcommit-percentage", "200") // Longhorn's default
```
guangbochen: prefer to use one setting but a JSON string config instead. e.g "overcommit-config": {"cpu-overcommit": "200", "memory-overcommit": "200", ....}

pkg/controller/master/setting/controller.go
```diff
@@ -24,3 +52,75 @@ func (h *Handler) LogLevelOnChanged(key string, setting *harvesterv1.Setting) (*
...
+
+func (h *Handler) cpuOvercommitPercentageOnChanged(setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	percentage, err := strconv.ParseInt(setting.Value, 10, 64)
+	if err != nil {
+		return setting, err
+	}
+	if percentage < 100 {
+		// You can only overcommit not undercommit
+		return setting, fmt.Errorf("Cannot undercommit. Should be greater that 100 but got %d", percentage)
+	}
```
guangbochen: resolve duplicate code with L86-93

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -0,0 +1,115 @@
...
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+			admissionregv1.Update,
+		},
+	}
+}
+
+func (m *vmMutator) Create(request *types.Request, newObj runtime.Object) (types.PatchOps, error) {
+	vm := newObj.(*kubevirtv1.VirtualMachine)
+	patchOps, err := m.patchResourceOvercommit(vm)
```
gitlawr, line 48: Why do we patch VM resources when we are using kubevirt's overcommit feature?

weihanglo: If we specify `resources.limits` on a VM, we'll lost the functionality of cluster-wide overcommit configuration of KubeVirt due to its implementation. See https://github.com/harvester/harvester/issues/1429#issuecomment-953031270

weihanglo: The reason of updating KubeVirt overcommit config here is just a simple synchronization to prevent future bugs.

gitlawr: What does `feature bugs` mean and how can we prevent by setting these values? If we are not using the configuration I propose not touching it

weihanglo: Sorry for the typo. The unused configuration part has been removed. Thanks.

pkg/settings/settings.go
```diff
@@ -29,9 +29,7 @@ var (
 	SupportBundleImage           = NewSetting("support-bundle-image", "rancher/support-bundle-kit:v0.0.3")
 	SupportBundleImagePullPolicy = NewSetting("support-bundle-image-pull-policy", "IfNotPresent")
 	DefaultStorageClass          = NewSetting("default-storage-class", "longhorn")
-	CpuOvercommitPercentage      = NewSetting("cpu-overcommit-percentage", "1000")    // KubeVirt's default is "10"x ratio. To align with others, we change it to percentage.
-	MemoryOvercommitPercentage   = NewSetting("memory-overcommit-percentage", "100")  // KubeVirt's default
-	StorageOvercommitPercentage  = NewSetting("storage-overcommit-percentage", "200") // Longhorn's default
+	OvercommitConfig             = NewSetting("overcommit-config", `{"cpu":1000,"memory":100,"storage":200}`) // KubeVirt & Longhorn's defaults
```
guangbochen: I would expect the default overcommit of both CPU and memory will be 200%, also can we do the ratio calculation in the method and keep the setting config here as a percentage only.

weihanglo: Sorry but didn't get the point of "do the ratio calculation in the method". In which method do you mention to do the calculation? 

> keep the setting config here as a percentage only.

Aren't we already use percentage here? Thanks.

weihanglo: Updated all defaults to 200% and rebased.

pkg/controller/master/setting/controller.go
```diff
@@ -24,3 +47,27 @@ func (h *Handler) LogLevelOnChanged(key string, setting *harvesterv1.Setting) (*
 	logrus.SetLevel(level)
 	return setting, nil
 }
+
+func (h *Handler) overcommitConfigOnChanged(setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	overcommit := &settings.Overcommit{}
+	if err := json.Unmarshal([]byte(setting.Value), overcommit); err != nil {
+		msg := fmt.Sprintf("Invalid JSON `%s`: %s", setting.Value, err.Error())
+		return setting, werror.NewInvalidError(msg, "Value")
```
guangbochen: use `fmt.Errorf()` probably is more starting forward.

guangbochen (commented):
Just one little fix, also can we squash some of the commits, thanks.

pkg/controller/master/setting/controller.go
weihanglo: Rebased and added `mutator_test.go` for virtualmachine. Thanks for the review!

guangbochen (approved):
LGTM

pkg/settings/settings.go
gitlawr: @guangbochen Any reason for the default values? The 200% memory default looks aggressive to me. FYI openstack use 16:1 for CPU and 1.5:1 for memory as the default: https://docs.openstack.org/arch-design/design-compute/design-compute-overcommit.html

guangbochen: I think the referenced doc is more meaningful in regards to VM's overcommit, @weihanglo can u PTAL and this will need a doc PR of the new changes, thanks.

weihanglo: Just rebased and updated to 16:1 for CPU and 1.5:1 for memory. 

The number looks reasonable to me since CPU overcommitment is not that harmful than out-of-memory. KubeVirt also defaults to a similar large ratio such as 10:1 for CPU. Thanks.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (m *vmMutator) getOvercommit() (*settings.Overcommit, error) {
+	s, err := m.setting.Get("overcommit-config")
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	if s.Value == "" {
```
bk201, line 114: The setting has a default value:
```
	OvercommitConfig             = NewSetting("overcommit-config", `{"cpu":1600,"memory":150,"storage":200}`)`
```

Do we want to use the default value if `s.Value` is empty? Sorry if I missed any discussion, resources of VM spec are not patched if I don't manually add the setting `value`.

If we want to use the default value, the storage part also needs to be passed to longhorn too.

weihanglo: We might not use defaults here if a user removes the value of that setting intentionally.

gitlawr (approved):
LGTM

Issue #983, opened 2021-06-28
[BUG] You can't migrate a VM off of a host that is unavailable


noahgildersleeve:
**Describe the bug**

You can't migrate a VM off of a host that is unavailable
**To Reproduce**
1. Create a new VM on a host in a cluster with at least 4 nodes
2. Turn off that host
3. Try to migrate the VM to a new host
**Expected behavior**

The VM should be able to be migrated
**Support bundle**

[harvester-supportbundle_c77a57eb-bc3d-4ed3-9037-6f9da0c0e2f6_2021-06-28T23-39-53Z.zip](https://github.com/harvester/harvester/files/6729711/harvester-supportbundle_c77a57eb-bc3d-4ed3-9037-6f9da0c0e2f6_2021-06-28T23-39-53Z.zip)

**Environment:**
 - Harvester ISO version: master-158a35d3-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): bare metal on DL160

**Additional context**
It gets stuck on migrating. 
![Greenshot 2021-06-28 16 46 26](https://user-images.githubusercontent.com/83787952/123716828-6a658500-d830-11eb-861d-636d92311e56.png)

guangyee:
We basically cannot do anything to the VM right now if the host where it is running suddenly goes offline.

noahgildersleeve:
@guangbochen This was marked as blocked by https://github.com/harvester/harvester/issues/982 but was moved into ready for testing? Is it not blocked anymore?

noahgildersleeve:
Tested in master-a412145f-head.
1. Turned off host that had VMs on it
2. Waited for host to show as powered off in ILO
3. Tried to migrate VM to new host
I got an error that said that the VM wasn't in ready status and a 500 error in the console.
![Greenshot 2021-11-08 11 55 34](https://user-images.githubusercontent.com/83787952/140809207-a8fddfc9-0db2-41d4-a4c2-ca75e6ad7fde.png)
![Greenshot 2021-11-08 11 55 39](https://user-images.githubusercontent.com/83787952/140809210-0997c269-629d-40bc-a5d4-71173e8359f5.png)
![Greenshot 2021-11-08 11 56 04](https://user-images.githubusercontent.com/83787952/140809212-8d351afe-f436-4254-ba56-f46a7834e7a5.png)
[...]

PR #1500, opened 2021-11-01
fix(vmi): restrict users to migrate VM when it is not ready

FrankYang0529:
**Problem:**
VM can be migrated when it is not Ready.

**Solution:**
Restrict users to migrate a VM when it is not Ready.

**Related Issue:**
https://github.com/harvester/harvester/issues/983

**Test plan:**

* Create a 2-node environment
* Turn off the agent node and wait for the VMI controller to update the Ready field of VMI to `False`.
* Try to migrate the VM to the server node. Check there is an error message like the following
Screen Shot 2021-11-01 at 4 22 16 PM

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -255,6 +255,9 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
 	if !vmi.IsRunning() {
 		return errors.New("The VM is not in running state")
 	}
+	if !isReady(vmi) {
+		return errors.New("Can't migrate the VM, the VM is not in ready status")
+	}
 	if !canMigrate(vmi) {
 		return errors.New("The VM is already in migrating state")
 	}
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -3,6 +3,7 @@ package vm
 import (
 	"github.com/rancher/apiserver/pkg/types"
 	"github.com/rancher/wrangler/pkg/data/convert"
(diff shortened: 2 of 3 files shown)


pkg/api/vm/handler.go
```diff
@@ -255,6 +255,14 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
...
 		return errors.New("The VM is not in running state")
 	}
+	for _, condition := range vmi.Status.Conditions {
+		if condition.Type == kv1.VirtualMachineInstanceReady {
+			if condition.Status != corev1.ConditionTrue {
+				return errors.New("The VM's Ready status is not True")
+			}
+			break
+		}
+	}
```
guangbochen: Can we wrap this with a method, and /s/The VM's Ready status is not True/Can't migrate the VM, the VM is not in ready status.

gitlawr (approved):
LGTM

pkg/api/vm/handler.go
```diff
@@ -255,6 +255,9 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
 	if !vmi.IsRunning() {
 		return errors.New("The VM is not in running state")
 	}
+	if !isReady(vmi) {
+		return errors.New("Can't migrate the VM, the VM is not in ready status.")
```
guangbochen: should remove `.` as the UI will append it by default. Otherwise, everything else is LGTM.

guangbochen (approved):
LGTM

Issue #982, opened 2021-06-28
[BUG] VM status doesn't update when host goes down


noahgildersleeve:
**Describe the bug**

VM status doesn't update when you take a host offline
**To Reproduce**
1. Create a VM on a host
2. Turn off the host
3. Wait for host to become unavailable
4. Check status of VM
**Expected behavior**

VM should either show that it's unavailable or migrate to a new host
**Support bundle**

[harvester-supportbundle_c77a57eb-bc3d-4ed3-9037-6f9da0c0e2f6_2021-06-28T23-07-22Z.zip](https://github.com/harvester/harvester/files/6729674/harvester-supportbundle_c77a57eb-bc3d-4ed3-9037-6f9da0c0e2f6_2021-06-28T23-07-22Z.zip)

**Environment:**
 - Harvester ISO version: master-158a35d3-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): bare metal on DL160

**Additional context**
![Greenshot 2021-06-28 16 06 27](https://user-images.githubusercontent.com/83787952/123715701-b3680a00-d82d-11eb-88b9-f3af3250b092.png)
![Greenshot 2021-06-28 16 06 35](https://user-images.githubusercontent.com/83787952/123715703-b4993700-d82d-11eb-98f8-e6ce2587686e.png)
![Greenshot 2021-06-28 16 06 41](https://user-images.githubusercontent.com/83787952/123715705-b531cd80-d82d-11eb-8fb2-a527340980db.png)

weihanglo:
<strike>Duplicate? https://github.com/harvester/harvester/issues/913</strike>

yasker:
Not sure if there is anything we can do for this release. Need to investigate first.

FrankYang0529:
For this issue and https://github.com/harvester/harvester/issues/983, Pods are stuck in the `Terminating` state when we turn off a node. This is also including Longhorn Pods. We can't gracefully remove Pods. Even if we set `--grace-period=0`, Pods still can't be removed. Also, there is no finalizer for VM Pods, so deleting finalizers doesn't work. We can only delete Pods by adding `--force`.

For a workable way, I think we should also check the `Ready` status in `VirtualMachineInstance`. If the `Ready` status is `False`, we show the `NotReady` state for the VM and hide the migration button. We can give a hint for users if they want to reschedule VMs, we can force delete Pods.

[...]

PR #1502, opened 2021-11-02
feat(node): add controller to handle node down

FrankYang0529:
**Problem:**
If a node is shut down, the VMI controller creates a new Pod only if the prior one has been deleted.  Also, the node can't report Pods' status, so VMI Pods will be stuck in `Terminating` status. We can't migrate or terminate them.

**Solution:**
We add a setting `vm-force-deletion-policy` for users to force delete VMI Pods that are on an unhealthy node.

**Related Issue:**
https://github.com/harvester/harvester/issues/982

**Test plan:**

Case 1: `vm-force-deletion-policy` has already been enabled.
* Set `vm-force-deletion-policy` as `'{"enable": true, "period": 60}'`.

Case 2: Enable `vm-force-deletion-policy` after a node is down.
* Set `vm-force-deletion-policy` as `'{"enable": true, "period": 60}'`.

**Experiment Result**
```
> kubectl get pods -w
NAME                     READY   STATUS      RESTARTS   AGE
virt-launcher-vm-tg5cj   1/1     Running     0          5m27s
virt-launcher-vm-tg5cj   1/1     Terminating   0          6m46s
virt-launcher-vm-ckplj   0/1     Pending       0          0s
virt-launcher-vm-ckplj   0/1     ContainerCreating   0          6m12s
virt-launcher-vm-ckplj   1/1     Running             0          6m13s
```

```
> kubectl get vmi -w
NAME   AGE   PHASE     IP           NODENAME           READY
vm     45m   Running   10.52.1.34   harvester-node-1   True
vm     45m   Running   10.52.1.34   harvester-node-1   False
vm     50m   Failed    10.52.1.34   harvester-node-1   False
vm     0s    Pending
vm     1s    Pending                                   False
vm     1s    Scheduling                                   False
vm     6m15s   Scheduled                 harvester-node-0   False
vm     6m15s   Running      10.52.0.135   harvester-node-0   True
```

### DIFF
--- a/pkg/controller/master/node/node_down_controller.go
+++ b/pkg/controller/master/node/node_down_controller.go
@@ -1,0 +1,159 @@
+package node
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	kv1 "kubevirt.io/client-go/api/v1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/config"
+	v1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/settings"
+)
+
+const (
+	nodeDownControllerName = "node-down-controller"
+	vmCreatorLabel         = "harvesterhci.io/creator"
+)
+
+// nodeDownHandler force deletes VMI's pod when a node is down, so VMI can be reschduled to anothor healthy node
+type nodeDownHandler struct {
+	nodes                       ctlcorev1.NodeController
+	nodeCache                   ctlcorev1.NodeCache
+	pods                        ctlcorev1.PodClient
+	virtualMachineInstanceCache v1.VirtualMachineInstanceCache
+}
+
+// NodeDownRegister registers a controller to delete VMI when node is down
+func NodeDownRegister(ctx context.Context, management *config.Management, options config.Options) error {
+	nodes := management.CoreFactory.Core().V1().Node()
+	pods := management.CoreFactory.Core().V1().Pod()
+	setting := management.HarvesterFactory.Harvesterhci().V1beta1().Setting()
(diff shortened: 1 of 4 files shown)


pkg/settings/settings.go
```diff
@@ -160,3 +169,24 @@ func InitBackupTargetToString() string {
...
 }
+
+func InitVMForceDeletionPolicy() string {
+	policy := &VMForceDeletionPolicy{
+		Enable: true,
+		Period: 5 * 60, // 5 minutes
+	}
+	policyStr, err := json.Marshal(policy)
+	if err != nil {
+		logrus.Errorf("failed to init string vmForceDeletionPolicy, error: %s", err.Error())
```
guangbochen: /s/logrus.Errorf("failed to init %s, error: %s", VMForceDeletionPolicySettingName, err.Error())

pkg/controller/master/backup/restore.go
```diff
@@ -49,7 +49,7 @@ const (
 	restoreNameAnnotation = "restore.harvesterhci.io/name"
 	lastRestoreAnnotation = "restore.harvesterhci.io/last-restore-uid"
 
-	vmCreatorLabel = "harvesterhci.io/creator"
+	VmCreatorLabel = "harvesterhci.io/creator"
```
guangbochen: \s\VmCreatorLabel\VMCreatorLabel

pkg/controller/master/node/node_down_controller.go
```diff
@@ -0,0 +1,159 @@
...
+	if node == nil || node.DeletionTimestamp != nil {
+		return node, nil
+	}
+
+	// get Ready condition
+	cond := getNodeCondition(node.Status.Conditions, corev1.NodeReady)
+	if cond == nil {
+		logrus.Warnf("can't find %s condition in Node", corev1.NodeReady)
+		return node, nil
+	}
```
guangbochen: just return the error instead since a node must contain the ready status.

pkg/settings/settings.go
FrankYang0529: Updated it. Thank you.

pkg/controller/master/backup/restore.go

pkg/controller/master/node/node_down_controller.go
FrankYang0529: Updated it. Thank you.

guangbochen (approved):
LGTM

gitlawr:
Can we add a validation webhook check to ensure this setting is in the correct format?

FrankYang0529:
> Can we add a validation webhook check to ensure this setting is in the correct format?

Added it. Thank you.

gitlawr (approved):
LGTM

abonillabeeche:
Can you share where to configure the vm-force-deletion-policy value?

Issue #1395, opened 2021-10-08
task: trigger the ISO build with the new Harvester commit


yasker:
Currently, the ISO only builds on either `harvester/harvester-installer` change or the nightly build. It should be triggered (and gated in the future) the `harvester/harvester` build as well. Otherwise, we will get the latest ISO with an older version of Harvester.

REF: https://github.com/harvester/harvester/issues/1351#issuecomment-939056608

w13915984028:
Target:
  When new commit is merged into master, the harvester ISO needs to be built and put into release.

Solution:
  In drone CI, add triggers on "push" event, so as to build and upload ISO.
 
In local test environment, the "Merge pull request" triggers build-iso, in last step, it upload the iso, please skip the error, as there is no related google_auth_key. 
 
![image](https://user-images.githubusercontent.com/31133476/140044635-26bdcd51-3ba1-40f9-b5fd-05cee30feba7.png)

guangbochen:
close in favor of https://github.com/harvester/harvester/pull/1506
ISO build of the Harvester repo is available via https://drone-publish.rancher.io/harvester/harvester/430/3/3

gitlawr:
One of the two builds failed: https://drone-publish.rancher.io/harvester/harvester/429/3/2
The reason is unclear. Let's keep an eye on it.

PR #1506, opened 2021-11-03
trigger the ISO build with the new Harvester commit

w13915984028:

**Problem:**

Currently, the ISO only builds on either harvester/harvester-installer change or the nightly build. It should be triggered (and gated in the future) the harvester/harvester build as well. Otherwise, we will get the latest ISO with an older version of Harvester.

**Solution:**

Add trigger on "push" event in the drone ci, when "merge pull request" happens, the "build-iso" will happen, and put the final iso into "releases.rancher.com/harvester/master-head".

As per discussion with @guangbochen , we do not keep history ISOs here, only put the latest one in master-head. In future, when necessary, we can also put each ISO into folders like 
"VERSION=${DRONE_BRANCH}-${DRONE_COMMIT_SHA:0:8}-head
target: releases.rancher.com/harvester/${VERSION}"

**Related Issue:**
task: trigger the ISO build with the new Harvester commit #1395 
**Test plan:**

Locally tested with Drone server and Gitea. After the merge of this pull request, the drone server will be triggered to build and upload harvester ISO, by check the drone CI, this commit will be verified.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -246,10 +246,32 @@ steps:
   when:
     instance:
       - drone-publish.rancher.io
+    event:
+      - tag
+
+- name: upload_iso_release_master
+  image: plugins/gcs
+  settings:
+    acl:
+      - allUsers:READER
+    cache_control: "public,no-cache,proxy-revalidate"
+    source: dist/artifacts
+    target: releases.rancher.com/harvester/master
+    token:
+      from_secret: google_auth_key
+  when:
+    instance:
+      - drone-publish.rancher.io
+    event:
+      - push
 
 trigger:
   event:
-  - tag
+    - tag
+    - push
+  ref:
+    - "refs/heads/master"
+    - "refs/heads/release/v*"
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -246,10 +246,32 @@ steps:
...
+      - tag
+
+- name: upload_iso_release_master
+   image: plugins/gcs
+   settings:
+     acl:
+       - allUsers:READER
+     cache_control: "public,no-cache,proxy-revalidate"
+     source: dist/artifacts
+     target: releases.rancher.com/harvester/master-head
```
guangbochen: \s\master-head\master

w13915984028: in new commit, it is named as below:
target: releases.rancher.com/harvester/master

guangbochen (approved):
LGTM, can u squash the commits, thanks

gitlawr (approved):
LGTM

Issue #1380, opened 2021-10-07
[BUG] Add Image upload notification message


yasker:
It seems I cannot get image upload to work with the latest master
<img width="453" alt="Screen Shot 2021-10-07 at 4 30 24 PM" src="https://user-images.githubusercontent.com/3344618/136475837-1b196dd2-5e8d-4c02-aa7d-d5ef6ed3f2f8.png">

yasker:
The second try works for me. It seems due to I've refreshed the page during the upload for the first time, since I didn't see any warning against it.

guangbochen:
Should notify user do not to refresh the page while uploading the file locally.

simonflood:
I'm seeing this with Harvester 0.3.0 when trying to upload ~250MB qcow2 files, either from a local directory on my laptop or a remote directory mounted via CIFS. I am NOT refreshing the Images page during upload.

simonflood:
I should note that the file size is not picked up on the Images page.

I am able to Create an image if I set up a local HTTP server using Python in the source directory and use the appropriate URL for the same files (http://<localipaddress>:8000/filename) - the file upload is almost instant with file size picked up.

noahgildersleeve:
I can repro this error, but it seems to be happening on dmg files. It's possible to upload them, are we supporting using that in the virtual drive? I'm not sure what kind of an OS level filetype list we're enforcing. 
I tested this with `master-d05a6cc7-head` and the google chrome dmg.
![image](https://user-images.githubusercontent.com/83787952/139736433-a60e3856-d335-4b2f-8b07-8935ba179c85.png)

n313893254:
We support .iso, .img, .qcow2 and .raw file types, the .dmg is the same as .img on the chrome browser view. All the images require the size should be a multiple of 512 bytes since Longhorn uses directIO by default.

How about adding an error notification on UI after the upload fail?

guangbochen:
Yep, adding a failed error message is helpful as the current `Failed to wait for xx` message is meaningless.

WolframRavenwolf:
I could reproduce this issue with the latest k3os-amd64.iso (v0.20.11-k3s2r1, 514 MB) and a remastered version of that. Tried it in Firefox and Chrome, didn't refresh the page, got the error message "Failed to wait for processing start in 60 seconds" every time.

TachunLin:
Update my test status on version master-d21b9161-head for reference.

[...]

PR #1507, opened 2021-11-03
Record error message when image upload request fails

gitlawr:
**Problem:**
When the upload request to Longhorn fails, the error is not recorded in image status.

**Solution:**
Set returned message in image status on upload request failure.

**Related Issue:**
https://github.com/harvester/harvester/issues/1380

### DIFF
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -119,7 +119,9 @@ func (h UploadActionHandler) uploadImage(rw http.ResponseWriter, req *http.Reque
 		return fmt.Errorf("failed to read response body: %w", err)
 	}
 	if uploadResp.StatusCode >= http.StatusBadRequest {
-		return fmt.Errorf("upload failed: %s", string(body))
+		// err will be recorded in image condition in the defer function
+		err = fmt.Errorf("upload failed: %s", string(body))
+		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/api/image/formatter.go
```diff
@@ -119,7 +119,8 @@ func (h UploadActionHandler) uploadImage(rw http.ResponseWriter, req *http.Reque
 		return fmt.Errorf("failed to read response body: %w", err)
 	}
 	if uploadResp.StatusCode >= http.StatusBadRequest {
-		return fmt.Errorf("upload failed: %s", string(body))
+		err = fmt.Errorf("upload failed: %s", string(body))
+		return err
```
guangbochen, line 124: why not return directly instead.

gitlawr: it will be handled in https://github.com/harvester/harvester/blob/9cde746ba1946cecb207a5b56d6e1cb300dd8aa5/pkg/api/image/formatter.go#L83-L89

guangbochen: can we add a comment to this one to avoid maloperation in the future, thanks?

gitlawr:
Updated

guangbochen (approved):
LGTM

Issue #1330, opened 2021-09-29
[FEATURE] Enhance the import experience to Rancher


yasker:
Currently user need to log into a Harvester physical node to run `kubectl` to import the Harvester cluster to a Rancher cluster.

We should allow that in the UI. Also, we need to enforce that one Harvester cluster can only be managed by one Rancher.

jnfinitym:
I also noticed the binding script in Rancher 2.6.1 seems to look for k3s.yaml instead of rke2.yaml which is created in the installation of Harvester 0.3; copying `/etc/rancher/rke2/rke2.yaml` to `/etc/rancher/k3s/k3s.yaml` allowed me to add default as admin-role and subsequently join the Rancher. Should I file this in a dedicated issue or is this just part of this one?

yasker:
@jnfinitym can you elaborate? Which binding script you're talking about? Might be another issue.

jnfinitym:
I will file a seperate issue then, so I have mroe space to describe it - I am not home right now, so I will try it this weekend, so I can add the exact error messages I got and what exactly I did to get around them; I am specifically talking about importing Harvester to a Rancher instance, but it looks to be easily resolvable, but I am thinking it may actually be a Rancher Problem and not a Harvester Problem… My workaround was on the Harvester side though

jnfinitym:
Weird, today I installed Harvester again from the same USB Stick and used the same version of Rancher deployed in the same way (in a VM on docker) and didn’t get the same issue!? - can’t reproduce it anymore, so I guess it was just a fluke.

guangbochen:
Refer from https://github.com/harvester/harvester/issues/1437

**Expected:**

that explain how to import a Harvester cluster. E.g. the specific steps to be able to run the "kubectl" command in harvester.

**Actual:**

user is shown generic instructions to run the command on "an existing Kubernetes cluster running a supported Kubernetes version"


<img width="841" alt="Screenshot 2021-10-15 at 12 56 32" src="https://user-images.githubusercontent.com/3813921/137477163-0ef47d87-7965-4c3b-9dd6-9b2624231345.png">

gitlawr:
Proposal for the import UX:
1. In Rancher UI, we show the instructions for importing Harvester using Harvester UI.
![image](https://user-images.githubusercontent.com/5697937/139803392-03c644a2-ffb9-4d29-88a4-6bc2abddcb07.png)

[...]

PR #1509, opened 2021-11-03
Support importing to Rancher via Harvester setting

gitlawr:
**Related Issue:**
https://github.com/harvester/harvester/issues/1330

### DIFF
--- a/pkg/controller/master/setting/cluster_registration_url.go
+++ b/pkg/controller/master/setting/cluster_registration_url.go
@@ -1,0 +1,67 @@
+package setting
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/rancher/wrangler/pkg/objectset"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/yaml"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+)
+
+// registerCluster imports Harvester to Rancher by applying manifests from the registration URL.
+func (h *Handler) registerCluster(setting *harvesterv1.Setting) error {
+	url := setting.Value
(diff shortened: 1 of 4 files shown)


pkg/controller/master/setting/register.go
```diff
@@ -16,16 +19,26 @@ func Register(ctx context.Context, management *config.Management, options config
...
 		deployments:     deployments,
 		deploymentCache: deployments.Cache(),
+		httpClient: http.Client{
+			Timeout: 30 * time.Second,
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true,
+				},
+			},
+		},
```
guangbochen, line 38: should we control request over proxy and CA cert if they are specified on the setting?

pkg/controller/master/setting/cluster_registration_url.go
```diff
@@ -0,0 +1,67 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/yaml"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+)
+
+// registerCluster imports Harvester to Rancher by applying manifests from the registration URL.
+func (h *Handler) registerCluster(setting *harvesterv1.Setting) error {
```
guangbochen, line 21: if the user deletes the imported harvester cluster on the Rancher side, we should not re-import the cluster again after the Harvester API server reboot.

And I was wondering should we reset this setting from the upstream Rancher or deal it on the downstream Harvester?

pkg/controller/master/setting/register.go
gitlawr: We don't need to specify explicitly and they will be applied.

pkg/controller/master/setting/cluster_registration_url.go
gitlawr: When users delete it upstream, it will not be re-imported after Harvester restarts. We have the setting hash so that it is done once.

guangbochen (approved):
LGTM

PR #1514, opened 2021-11-05
Reuse SettingsProvider between webhook and main controller

weihanglo:

**Problem:**
Per https://github.com/harvester/harvester/pull/1492#discussion_r742904953, the `setting.Default` is not used even when `setting.Value` is empty. 

**Solution:**
The cause is that mutator does not reuse existing `settings` module and `settingsProvider`. In order not to duplicate code, we extract `settingsProvider` from controller and reuse the logic between webhook and main controller

**Related Issue:**
#1429 

**Test plan:**
Same as #1492 but do not edit `overcommit-config` setting. Leave its `value` field empty.

### DIFF
--- a/pkg/settings/provider.go
+++ b/pkg/settings/provider.go
@@ -1,0 +1,147 @@
+package settings
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+)
+
+type Provider interface {
+	Get(name string) string
+	Set(name, value string) error
+	SetIfUnset(name, value string) error
+	SetAll(settings map[string]Setting) error
+}
+
+type SettingsProvider struct {
+	context       context.Context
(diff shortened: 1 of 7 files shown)


pkg/settings/provider.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+			if newSetting.Value == "" {
+				fallback[newSetting.Name] = newSetting.Default
+			} else {
+				fallback[newSetting.Name] = newSetting.Value
+			}
+			_, err := s.settings.Create(newSetting)
+			if err != nil {
+				return err
+			}
```
gitlawr, line 117: Now that we run settings controllers in the webhook server, will it be a race condition here?
Assume that :
A setting value is set in Harvester server env. Webhook server starts before Harvester server. It will create the setting with an empty value which is not expected.

weihanglo: Hmm... That is a practical concern. If we choose not go this way, in order to resolve https://github.com/harvester/harvester/pull/1492#discussion_r742904953, we might need to duplicate the "use default value" logic for webhook.

gitlawr: We already used a client to fetch the setting resource. Can we simply check the default field?

weihanglo: Yep. I think we can close this PR and check the default here.
https://github.com/harvester/harvester/blob/979b7f81b941a1f8643994641c59f686b04bc426/pkg/webhook/resources/virtualmachine/mutator.go#L106-L122

weihanglo:
Close this in favor of #1526

PR #1515, opened 2021-11-05
Add preconfig loading controller

johnliu55tw:
**Problem:**
Users want to modify some settings of Harvester during the installation process, so we need a way to pull those modifications into the `Setting` CRs while bootstrapping Harvester cluster.

Note: RKE2 has the ability to [auto-deploy manifests](https://docs.rke2.io/advanced/#auto-deploying-manifests) stored in `/var/lib/rancher/rke2/server/manifests`, but these manifests will be re-applied whenever the RKE2 system service restarted (happens when Node rebooted). If we put Setting CRs manifest directly in the directory, data stored in the cluster will be overwritten by these manifests.

**Solution:**
These "preconfigs" will be stored as manifest files in `/var/lib/rancher/rke2/server/manifests` during the installation process. These files **only contain annotations**. Fields like `value` or `default` are absent so they won't be re-applied while Node rebooted. The `PreconfigController` reads the annotation, then determines whether to overwrite `value` from the preconfig based on conditions of the CR.

**Related Issue:**

**Test plan:**
Currently, it's not easy to test this because we don't have the ability to add preconfig Setting CR to `/var/lib/rancher/rke2/server/manifests` yet. Here's the manual step to test the controller:

1. Install Harvester using ISO and wait for it to be Ready
2. SSH into the system and add a file named `settings.yaml` to `/var/lib/rancher/rke2/server/manifests`:
  ```
  ---
  kind: Setting
  metadata:
    annotations:
      harvesterhci.io/preconfigValue: DEBUG
    name: log-level
  ```
3. Restart the `harvester` Pod, either kill the Pod or restart its Deployment should work.
4. Wait until the `harvester` Pod to be Running.
5. Check the `log-level` Setting CR in the cluster and **verify its value is now `DEBUG`**.
6. Use Harvester GUI or `kubectl` to update the `log-level` Setting CR again, with value `Info`
7. Restart the Node
8. Wait until it's Ready again, and verify the value of `log-level` is still `Info`. This shows that the value **will not be overwritten by manifest files` after rebooted.

### DIFF
--- a/pkg/apis/harvesterhci.io/v1beta1/settings.go
+++ b/pkg/apis/harvesterhci.io/v1beta1/settings.go
@@ -6,7 +6,8 @@ import (
 )
 
 var (
-	SettingConfigured condition.Cond = "configured"
+	SettingConfigured      condition.Cond = "configured"
+	SettingPreconfigLoaded condition.Cond = "preconfigLoaded"
 )
 
 // +genclient
--- a/pkg/controller/master/setting/load_from_config.go
+++ b/pkg/controller/master/setting/load_from_config.go
@@ -1,0 +1,36 @@
+package setting
+
+import (
+	"github.com/sirupsen/logrus"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+type LoadingPreconfigHandler struct {
+	namespace string
+	settings  v1beta1.SettingClient
+}
+
+func (h *LoadingPreconfigHandler) settingOnChanged(_ string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	if setting == nil || setting.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	if value, ok := h.getPreconfigValue(setting); ok && !harvesterv1.SettingPreconfigLoaded.IsTrue(setting) {
+		settingCpy := setting.DeepCopy()
+		logrus.Infof("loading preconfig value for setting '%s'", setting.Name)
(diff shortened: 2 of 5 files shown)


pkg/apis/harvesterhci.io/v1beta1/settings.go
```diff
@@ -6,7 +6,8 @@ import (
 )
 
 var (
-	SettingConfigured condition.Cond = "configured"
+	SettingConfigured      condition.Cond = "configured"
+	SettingPreconfigLoaded condition.Cond = "preconfigLoaded"
```
bk201, line 10: @gitlawr @guangbochen we have two different naming styles for condition type in the code. Some start with upper case and some start with lower case. Do we want to be consistent? Conditions in Kubernetes resources seem to use upper case more.

gitlawr: We want to use a consistent naming style. For that change, we need to consider upgrades and coordination with the UI. We'd better track it in a separate issue.

guangbochen (approved):
LGTM

pkg/controller/master/setting/load_from_config.go
```diff
@@ -0,0 +1,37 @@
...
+	namespace string
+	settings  v1beta1.SettingClient
+}
+
+func (h *LoadingPreconfigHandler) settingOnChanged(_ string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	if setting == nil || setting.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	settingCpy := setting.DeepCopy()
```
gitlawr: Move this into the following if block so that we only do deep copy when the update is needed.

johnliu55tw: Moved into the if block.

gitlawr (approved):
LGTM

gitlawr (approved):
LGTM

PR #1526, opened 2021-11-09
Use default overcommit value for CPU/memory

weihanglo:
**Problem:**
Use default overcommit value for CPU/memory.

**Solution:**
Add ad-hoc logic to use default overcommit values. Favor this over #1514

**Related Issue:**
#1429

**Test plan:**


guangbochen (approved):
LGTM

gitlawr (approved):
LGTM

PR #1532, opened 2021-11-10
Add auto prov setting and controller

johnliu55tw:
**Problem:**
We need to add the host's disks as VM storage automatically if the disks' device path matches a given shell glob pattern. This pattern can be given by the remote config during the installation process, and can be modified afterward using CustomResource Setting `auto-disk-provision-paths`.

Do notice that using the remote config to set up the pattern **only works when creating a new cluster**. Installing in join mode **cannot** overwrite any system settings, even with remote config provided.

After disks are added, changing the pattern won't affect any added disks.

**Solution:**
Add a controller to detect changes on Setting "auto-disk-provision-paths" and update NDM chart value in harvester `ManagedChart` accordingly.

**Related Issue:**

**Test plan:**

> :information_source: **Note: Testing on KVM** :information_source: 
> If you are using KVM to test this feature, you might notice that disks are not added automatically. You might see something like this in Node Disk Manager's log:
> ```
> # kubectl logs -n harvester-system harvester-node-disk-manager-XXXXX
> ...
> time="2021-11-19T07:20:40Z" level=warning msg="failed to read disk uuid of /dev/sdd : exit status 2\n"
> time="2021-11-19T07:20:40Z" level=warning msg="failed to read disk uuid of /dev/sdd : exit status 2\n"
> time="2021-11-19T07:20:40Z" level=warning msg="failed to read disk uuid of /dev/sdd : exit status 2\n"
> time="2021-11-19T07:20:41Z" level=warning msg="failed to generate GUID for device sdd"
> ...
> ```
> This is because the disks created by KVM (QEMU) have no identifiable ID, so Node Disk Manager cannot recognize them. You have several options to work around this issue:
> - Manually creating GPT for the disk using `parted` or `gdisk` during the installation process (Use `Ctrl+Alt+F2` to switch to another terminal), so we could use the UUID generated by GPT.
> - Enable Node Disk Manager feature [`autoGPTGenerate`](https://github.com/harvester/charts/blob/153951f5eccc352a812afa74dd4f74b3b0420997/charts/harvester-node-disk-manager/values.yaml#L75)
> - (Not tested) Adding disk WWN. See https://libvirt.org/manpages/virsh.html#attach-disk
>
> This should **not** happen on bare-metal systems (If so, please file a bug). Since I only tested on KVM, I'm not sure how other Hypervisors would behave.

The key idea here is that all the disks on the machine (besides the one used to install Harvester) that match the Glob pattern (`/dev/sd*`) would be added automatically. For example, if you have **four** SCSI/SATA drives on the machine, one of them will be used to install Harvester and the other three should be added automatically.

1. Prepare a VM or bare-metal machine with **two** SATA/SCSI drives. One for installing Harvester and the other for VM storage.
2. Install Harvester **with a remote config** to overwrite `auto-disk-provision-paths` system setting:
    ```
    system_settings:
      auto-disk-provision-paths: /dev/sd*
    ```
3. Wait for Harvester to be ready. Go to GUI and verify **the second drive is added automatically**
    image
4. Prepare another VM or bare-metal machine with **two** SATA/SCSI drives.
5. Install Harvester onto the second machine, in **join** mode.
6. Wait for the second machine to be ready. Go to GUI and verify that for the second machine, the other disk is also added automatically.

**Dependencies:**
- https://github.com/harvester/node-disk-manager/pull/18
- https://github.com/harvester/harvester-installer/pull/169
- https://github.com/harvester/docs/pull/74

### DIFF
--- a/pkg/controller/master/setting/auto_disk_provision.go
+++ b/pkg/controller/master/setting/auto_disk_provision.go
@@ -1,0 +1,49 @@
+package setting
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+)
+
+const (
+	ManagedChartNamespace     = "fleet-local"
+	HarvesterManagedChartName = "harvester"
+	NDMName                   = "harvester-node-disk-manager"
+)
+
+func (h *Handler) syncNDMAutoProvisionPaths(setting *harvesterv1.Setting) error {
+	mChart, err := h.managedChartCache.Get(ManagedChartNamespace, HarvesterManagedChartName)
+	if err != nil {
+		return err
+	}
+	mChartCopy := mChart.DeepCopy()
+
+	NDMValues, ok := mChartCopy.Spec.Values.Data[NDMName]
+	if !ok {
+		return fmt.Errorf("NDM chart value not found in ManagedChart")
+	}
+
+	NDMValuesMap, ok := NDMValues.(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("NDM chart value is not a map[string]interface{}")
+	}
+
+	autoProvFilters := strings.Split(setting.Value, ",")
+	for i, filter := range autoProvFilters {
+		autoProvFilters[i] = strings.TrimSpace(filter)
+	}
+
+	NDMValuesMap["autoProvisionFilter"] = autoProvFilters
+	mChartCopy.Spec.Values.Data[NDMName] = NDMValuesMap
+
+	logrus.Debugf("NDM values to be updated to ManagedChart: %v", mChartCopy.Spec.Values.Data[NDMName])
+	if _, err := h.managedCharts.Update(mChartCopy); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/controller/master/setting/handler.go
+++ b/pkg/controller/master/setting/handler.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"time"
 
+	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	provisioningv1 "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
 	"github.com/rancher/wrangler/pkg/apply"
 	v1 "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
@@ -37,6 +38,8 @@ type Handler struct {
 	longhornSettingCache ctllonghornv1.SettingCache
 	configmaps           ctlcorev1.ConfigMapClient
 	configmapCache       ctlcorev1.ConfigMapCache
+	managedCharts        mgmtv3.ManagedChartClient
+	managedChartCache    mgmtv3.ManagedChartCache
 }
 
 func (h *Handler) settingOnChanged(_ string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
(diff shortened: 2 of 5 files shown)


pkg/controller/master/setting/auto_disk_provision.go
```diff
@@ -0,0 +1,52 @@
...
+					dsCopy.Spec.Template.Spec.Containers[i].Env,
+					corev1.EnvVar{Name: NDMAutoProvisionEnvName, Value: setting.Value},
+				)
+			}
+
+			break
+		}
+	}
+
+	if _, err := h.daemonSets.Update(dsCopy); err != nil {
```
gitlawr: NDM is deployed using helm but we are patching the daemonset directly. My concern is the changes will be lost during upgrades. Should we update via helm release? For example, using https://docs.rke2.io/helm/#customizing-packaged-components-with-helmchartconfig

johnliu55tw: I need to check this out, thanks!

johnliu55tw: @gitlawr unless we also update the file `/var/lib/rancher/rke2/server/manifests/harvester-config.yaml`, everything would be reverted by RKE2 after a reboot. Also, we can't directly modify the file since it's generated by cOS-toolkit's OEM file `/oem/99_custom.yaml`.

guangbochen: I found this is a common issue that we have encountered in many places and has raised it on the https://suse.slack.com/archives/C02D01ARDL5/p1636703672048800

pkg/controller/master/setting/auto_disk_provision.go
```diff
@@ -0,0 +1,51 @@
...
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+)
+
+const (
+	HarvesterManagedChartName = "harvester"
+	NDMName                   = "harvester-node-disk-manager"
+)
+
+func (h *Handler) syncNDMAutoProvisionPaths(setting *harvesterv1.Setting) error {
+	mChart, err := h.managedChartCache.Get("fleet-local", HarvesterManagedChartName)
```
guangbochen: prefer to move to const.

pkg/controller/master/setting/auto_disk_provision.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return err
+	}
+	mChartCopy := mChart.DeepCopy()
+
+	NDMValues, ok := mChartCopy.Spec.Values.Data[NDMName]
+	if !ok {
+		logrus.Error("NDM value not found in ManagedChart")
```
guangbochen: should update the error message to the configured status so it is obvious to the users.

pkg/controller/master/setting/auto_disk_provision.go
```diff
@@ -0,0 +1,51 @@
...
+		// No need to retry if Chart values for NDM not found
+		return nil
+	}
+
+	NDMValuesMap, ok := NDMValues.(map[string]interface{})
+	if !ok {
+		logrus.Error("NDM value is not a map[string]interface{}")
```
guangbochen: ditto.

pkg/controller/master/setting/auto_disk_provision.go
johnliu55tw: After thoroughly read the code I think I'll let the logic [here](https://github.com/harvester/harvester/blob/f5bde7fa462c09d0513b8f370f71f2dd974f298c/pkg/controller/master/setting/handler.go#L64-L69) handle the condition updating.

pkg/controller/master/setting/auto_disk_provision.go
johnliu55tw: After thoroughly reading the code I think I'll let the logic [here](https://github.com/harvester/harvester/blob/f5bde7fa462c09d0513b8f370f71f2dd974f298c/pkg/controller/master/setting/handler.go#L64-L69) handle the condition updating.

pkg/controller/master/setting/auto_disk_provision.go
johnliu55tw: Done

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

Issue #1536, opened 2021-11-11
[BUG] NFS backups don't finish


noahgildersleeve:
**Describe the bug**

When you start an NFS backup it doesn't complete
**To Reproduce**
1. Create new VM
2. Create a backup with an NFS backup target
**Expected behavior**

Backup should complete
**Support bundle**

[supportbundle_caa0bfe9-902d-45d6-bb88-9b74fe60e0eb_2021-11-10T21-27-20Z.zip](https://github.com/harvester/harvester/files/7517000/supportbundle_caa0bfe9-902d-45d6-bb88-9b74fe60e0eb_2021-11-10T21-27-20Z.zip)

**Environment:**
 - Harvester ISO version: master-119c2c48-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): DL160

**Additional context**
The backups show as complete in Longhorn, but they don't show as finished in the list in backups. This was done to a local NFS share created on the embedded rancher using the YAML here. 
https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/deploy/backupstores/nfs-backupstore.yaml
backup target was set as `nfs://longhorn-test-nfs-svc.default:/opt/backupstore`
![Greenshot 2021-11-10 13 16 59](https://user-images.githubusercontent.com/83787952/141220968-a3f990c6-8a1c-4f7a-8642-301c8819ba43.png)
![Greenshot 2021-11-10 13 17 06](https://user-images.githubusercontent.com/83787952/141220970-382af19b-f98b-404b-bd1b-0eb77c890241.png)

FrankYang0529:
I can reproduce the issue. The root cause is that the value in secret for `cloudInitNoCloud.secretRef` and `cloudInitNoCloud.networkDataSecretRef` must be encoded by base64. So empty string will cause an error.

```
Invalid value: \"\": status.secretBackups.data.networkdata in body must be of type byte: \"\"
```

```
apiVersion: v1
data:
  networkdata: "" <-- this will cause error
[...]

PR #1539, opened 2021-11-12
fix(backup): remove empty string in secret backup

FrankYang0529:
**Problem:**
If there is an empty string in secret backup, we will get an invalid value error.

```
Invalid value: \"\": status.secretBackups.data.networkdata in body must be of type byte: \"\"
```

**Solution:**
Remove empty string in the secret backup.

**Related Issue:**
https://github.com/harvester/harvester/issues/1536

**Test plan:**

* Create a secret like the following:
```
apiVersion: v1
kind: Secret
metadata:
  name: secret-volume
type: Opaque
data:
  userdata: "I2Nsb3VkLWNvbmZpZwpwYXNzd29yZDogcGFzc3dvcmQKY2hwYXNzd2Q6CiAgZXhwaXJlOiBmYWxzZQpzc2hwd2F1dGg6IHRydWUKcGFja2FnZV91cGRhdGU6IHRydWUKcGFja2FnZXM6CiAgLSBxZW11LWd1ZXN0LWFnZW50CnJ1bmNtZDoKICAtIC0gc3lzdGVtY3RsCiAgICAtIGVuYWJsZQogICAgLSAnLS1ub3cnCiAgICAtIHFlbXUtZ2Euc2VydmljZQpzc2hfYXV0aG9yaXplZF9rZXlzOgogIC0gPi0KICAgIHNzaC1yc2EKICAgIEFBQUFCM056YUMxeWMyRUFBQUFEQVFBQkFBQUJnUUM5Nzc4OFNvb3BPamZ2YmZQbkF2MzVHcWJrVFhENFRoWWdVZFZhT2FBV2ErTzVoSFNnZ05RTDkvK2xCQ0RsWGI3SzEvSGlmOFJiUkdURnpxU3d6RDJWNVhpcWJXcFFEZlJsVFg4T0s4SjdqTWxUYU96TVc3SUFyMjJGSFRveVlHQTAxUGtWdXkxV2UvSzV6RmIrdWVsZEt2NEZSZ0NkcVlqaGdhaklVZEYxRWRlOTF6WEFtb3BuSExRenk3SmZIZi94MEJrVnp0S1N1Y2d0aW1PdUJzODNFR1dENnBGbnh2UmNQMXBITGl4TXI0d3FQU1lSUFVzeW5nL0tKejVUOFd2NGdTL0NtS09Kc3EydjNxNGpWVTREMWhuR3l5SXovS0JqejI3ckluRUlRbVpmMXpOTkdBemlHdWc0cXdwbS9INVMwbEtlQWRyTmlOWitwY1p2MGZORUVLa292TmkyWm9DRDJzeGJ0TTZOM3d6ODhVdkRpck95a2JpdmY2U0ZnWlVVQ29JS0xpMnh5Ym84bFBUUm05emx1QUZrVjJrQzgvUDl3bkR4SzY1U0hwL1R5bFYxeXlvakowY2o5QUpyTU93czQ0czlYTUJVWUpNSjAyMEZqbXB0R3NBT1pPT3k1Q0x4OXZlMk5iNnJGcktiOFE4NkJjdU12TUsrbHcvSnFPMD0KICAgIG5vYWguZ2lsZGVyc2xlZXZlQHN1c2UuY29tCg=="
  networkdata: ""
```

* Create a VM with the above secret:
```
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  # ...
spec:
    # ...
    spec:
      # ...
      volumes:
      # ...
      - cloudInitNoCloud:
          networkDataSecretRef:
            name: secret-volume
          secretRef:
            name: secret-volume
        name: cloudinitdisk
```

* Setup NTP/S3 backup target
* Backup VM and check vmbackup can go into Ready state
* Try to restore vmbackup to new vm and check the vm can go into Running state.

### DIFF
--- a/pkg/controller/master/backup/backup.go
+++ b/pkg/controller/master/backup/backup.go
@@ -180,32 +180,53 @@ func (h *Handler) getVolumeBackups(backup *harvesterv1.VirtualMachineBackup, vm 
 
 // getSecretBackups helps to build a list of SecretBackup upon the cloud init secrets used by the backup VM
 func (h *Handler) getSecretBackups(vm *kv1.VirtualMachine) ([]harvesterv1.SecretBackup, error) {
-	var secretBackups []harvesterv1.SecretBackup
+	secretRefs := []*corev1.LocalObjectReference{}
 
 	for _, volume := range vm.Spec.Template.Spec.Volumes {
 		if volume.CloudInitNoCloud != nil && volume.CloudInitNoCloud.UserDataSecretRef != nil {
-			secret, err := h.secretCache.Get(vm.Namespace, volume.CloudInitNoCloud.UserDataSecretRef.Name)
-			if err != nil {
-				return nil, err
-			}
-			secretBackups = append(secretBackups, harvesterv1.SecretBackup{
-				Name: secret.Name,
-				Data: secret.Data,
-			})
+			secretRefs = append(secretRefs, volume.CloudInitNoCloud.UserDataSecretRef)
 		}
 		if volume.CloudInitNoCloud != nil && volume.CloudInitNoCloud.NetworkDataSecretRef != nil {
-			secret, err := h.secretCache.Get(vm.Namespace, volume.CloudInitNoCloud.NetworkDataSecretRef.Name)
-			if err != nil {
-				return nil, err
-			}
-			secretBackups = append(secretBackups, harvesterv1.SecretBackup{
-				Name: secret.Name,
-				Data: secret.Data,
-			})
+			secretRefs = append(secretRefs, volume.CloudInitNoCloud.NetworkDataSecretRef)
 		}
 	}
 
+	secretBackups := []harvesterv1.SecretBackup{}
+	secretBackupMap := map[string]bool{}
+	for _, secretRef := range secretRefs {
+		// users may put UserDataSecretRef and NetworkDataSecretRef in a same secret, so we only keep one
+		secretFullName := fmt.Sprintf("%s/%s", vm.Namespace, secretRef.Name)
+		if v, ok := secretBackupMap[secretFullName]; ok && v {
+			continue
+		}
+
+		secretBackup, err := h.getSecretBackupFromSecret(vm.Namespace, secretRef.Name)
+		if err != nil {
+			return nil, err
+		}
+		secretBackupMap[secretFullName] = true
+		secretBackups = append(secretBackups, *secretBackup)
+	}
+
 	return secretBackups, nil
+}
+
+func (h *Handler) getSecretBackupFromSecret(namespace, name string) (*harvesterv1.SecretBackup, error) {
+	secret, err := h.secretCache.Get(namespace, name)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove empty string. If there is empty string in secret, we will encounter error.
+	// ref: https://github.com/harvester/harvester/issues/1536
+	data := secret.DeepCopy().Data
+	for k, v := range secret.Data {
+		if len(v) == 0 {
+			delete(data, k)
+		}
+	}
+
+	return &harvesterv1.SecretBackup{Name: secret.Name, Data: data}, nil
 }
 
 // updateBackupStatusContent helps to store the backup source and volume contents within the VM backup status
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -18,7 +18,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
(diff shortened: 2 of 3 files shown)


pkg/controller/master/backup/backup.go
```diff
@@ -180,31 +180,57 @@ func (h *Handler) getVolumeBackups(backup *harvesterv1.VirtualMachineBackup, vm
...
+			secretBackupMap[secretFullName] = harvesterv1.SecretBackup{
 				Name: secret.Name,
-				Data: secret.Data,
-			})
+				Data: data,
+			}
 		}
 	}
 
+	secretBackups := make([]harvesterv1.SecretBackup, 0, len(secretBackupMap))
```
w13915984028: It looks the secretBackupMap is used to skip duplicated names, when yes,  we may consider to improve it:

`secretBackupMap := map[string]harvesterv1.SecretBackup{} `  -----> `map[string]int/bool`

in the loop,  
when volume.CloudInitNoCloud.UserDataSecretRef
  generate name, add name into map, add Data into  secretBackups 
 volume.CloudInitNoCloud.NetworkDataSecretRef
   generate name, check if in map: yes, continue;  no, add name into map, add Data into secretBackups 

finally, return `secretBackups `directly, no need to do following for loop  to generate it

FrankYang0529: Good catch. Updated it. Thank you.

pkg/controller/master/backup/backup.go
```diff
@@ -181,26 +181,56 @@ func (h *Handler) getVolumeBackups(backup *harvesterv1.VirtualMachineBackup, vm
...
+			// Remove empty string. If there is empty string in secret, we will encounter error.
+			// ref: https://github.com/harvester/harvester/issues/1536
+			data := secret.DeepCopy().Data
+			for k, v := range secret.Data {
+				if len(v) == 0 {
+					delete(data, k)
+				}
+			}
+
+			secretBackupMap[secretFullName] = true
```
guangbochen: seems duplicated with the above check and can we move it to a method.

pkg/controller/master/backup/backup.go
```diff
@@ -181,26 +181,56 @@ func (h *Handler) getVolumeBackups(backup *harvesterv1.VirtualMachineBackup, vm
...
 			})
 		}
 		if volume.CloudInitNoCloud != nil && volume.CloudInitNoCloud.NetworkDataSecretRef != nil {
 			secret, err := h.secretCache.Get(vm.Namespace, volume.CloudInitNoCloud.NetworkDataSecretRef.Name)
 			if err != nil {
 				return nil, err
 			}
+
+			// users may put UserDataSecretRef and NetworkDataSecretRef in a same secret, so we only keep one
+			secretFullName := fmt.Sprintf("%s/%s", secret.Namespace, secret.Name)
```
guangbochen: should move the check to the top as RefName and vm.Namespace is same as the secret one.

pkg/controller/master/backup/restore.go
```diff
@@ -335,29 +168,60 @@ func (h *RestoreHandler) VMOnChange(key string, vm *kv1.VirtualMachine) (*kv1.Vi
...
+		restoreCpy.Status.VolumeRestores = volumeRestores
+	}
+
+	if restoreCpy.Status.DeletedVolumes == nil {
+		var deletedVolumes []string
+		for _, vol := range backup.Status.VolumeBackups {
+			deletedVolumes = append(deletedVolumes, vol.PersistentVolumeClaim.ObjectMeta.Name)
+		}
+		restoreCpy.Status.DeletedVolumes = deletedVolumes
+	}
```
guangbochen: probably skip adding deleted volumes of new restored VM and retained policy is preferred.

pkg/controller/master/backup/backup.go
FrankYang0529: Refactor it. Thank you.

pkg/controller/master/backup/restore.go
FrankYang0529: Check whether is the new VM or has retained policy. Thank you.

pkg/controller/master/backup/backup.go
FrankYang0529: Updated it. Thank you.

pkg/controller/master/backup/restore.go
```diff
@@ -695,75 +551,115 @@ func (h *RestoreHandler) createRestoredPVC(vmRestore *harvesterv1.VirtualMachine
...
-	return err
+	return nil
 }
 
-func updateRestoreCondition(r *harvesterv1.VirtualMachineRestore, c harvesterv1.Condition) {
-	r.Status.Conditions = updateCondition(r.Status.Conditions, c, true)
+func (h *RestoreHandler) restartVM(vm *kv1.VirtualMachine) error {
+	logrus.Infof("restarting the vm %s, current state running:%v", vm.Name, *vm.Spec.Running)
+	var running = true
+	if vm.Spec.Running == nil || vm.Spec.Running == &running {
```
gitlawr: `*vm.Spec.Running` instead of `vm.Spec.Running == &running`

FrankYang0529: Fixed it. Thank you.

guangbochen (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1537, opened 2021-11-11
[BUG] Memory is being saved as the wrong amount in VMs


noahgildersleeve:
**Describe the bug**

When you save a VM with a raw image template the amount of memory is showing as incorrect in the config
**To Reproduce**
1. Create a VM with a raw image template
2. Leave defaults as they are (1vcpu, 2gi memory)
3. Save VM
4. Check list for memory in VM
5. Open VM and check amount of memory
**Expected behavior**

The amount of memory should be what is set on creation
**Support bundle**

[supportbundle_59f99d92-cd77-47ab-96c9-27d2042ac7c4_2021-11-11T18-20-25Z.zip](https://github.com/harvester/harvester/files/7522529/supportbundle_59f99d92-cd77-47ab-96c9-27d2042ac7c4_2021-11-11T18-20-25Z.zip)

**Environment:**
 - Harvester ISO version: master-119c2c48-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): DL160

**Additional context**
This seems to be a regression and if you try to just open and save the config again after creating it then you get an error that says that it isn't a valid amount. It's showing as 1.33gi memory. The YAML is showing a few different amounts of memory. If you edit the VM after creation and set it to 2gi memory it shows up properly after saving. I'm attaching screenshots.
![Greenshot 2021-11-11 10 20 29](https://user-images.githubusercontent.com/83787952/141350496-dece6437-683d-4686-bd77-4e797498fa72.png)
![Greenshot 2021-11-11 10 26 35](https://user-images.githubusercontent.com/83787952/141350497-df07ba9b-cdee-486e-a6c9-dbddebcc927a.png)
![Greenshot 2021-11-11 10 30 29](https://user-images.githubusercontent.com/83787952/141350501-4b3d6b3f-9f9c-4e17-a148-08ca3d10bc74.png)
![Greenshot 2021-11-11 10 32 08](https://user-images.githubusercontent.com/83787952/141350651-2b8e98ea-1176-4b08-98df-43fe816786a2.png)

guangbochen:
Memory is calculated in bytes but I think to be more user friendly can we calculate it in `Mi`, cc @weihanglo

weihanglo:
> Memory is calculated in bytes but I think to be more user friendly can we calculate it in `Mi`, cc @weihanglo

AFAIK `resources` for memory only accept integer and there will be a precision loss if we truncating memory like that. Though it is technically achievable.

[...]

Issue #1534, opened 2021-11-10
[BUG] VM memory display in long bytes value instead GB on virtual machine page


TachunLin:
**Describe the bug**

Create VM with 4GB memory, but it displays memory value `2863311530` on the virtual machine page

![image.png](https://images.zenhubusercontent.com/61519853321ea20d65443929/c08db590-d337-40c5-bdca-10cac804ff74)

**To Reproduce**
1. Create an virtual machine
2. Give 4GB memory 
3. Provide required information and create vm 
4. Check the memory value on virtual machine page 

**Expected behavior**

Should display virtual machine value in GB unit 

**Support bundle**

**Environment:**
 - Harvester ISO version: master-119c2c48-head
 - Underlying Infrastructure: 3 nodes harvester on provo bare machines (RackD8 ClusterA) 

**Additional context**
Memroy unit on VM creation page 
![image.png](https://images.zenhubusercontent.com/61519853321ea20d65443929/ff18decc-bbe5-4fdc-85d9-ca39f18609f8)

WuJun2016:
@TachunLin  I set up 4GB of memory without reproducing this issue, the long unit should be returned by the backend, I don't know under what circumstances the backend will convert the unit 。  cc @guangbochen

TachunLin:
@WuJun2016 
This issue can be reproduced with master-119c2c48-head on both my local laptop and 
provo bare machines (RackD8, ClusterA) 
Today I tested again on my local kvm, it still happen here.
![image](https://user-images.githubusercontent.com/29251855/141231209-b245a4b1-e23a-47db-bc55-9cc496230cc6.png)

I just create images, setup network, add ssh keys then create a virtual machine. 
Then the VM record shows long bit value.

lanfon72:
this should be caused by #1429

weihanglo:
Note that the value is the exact bytes after memory overcommit calculation. If we want to truncate that value in the backend we'll also affect the actual memory requested.

WuJun2016:
@TachunLin  What is the value of overcommit-config in your environment?  It seems that the default overcommit-config value does not have a unit problem, I need to reproduce it to convert to the correct unit。

TachunLin:
@WuJun2016 
The overcommit-config value keeps in default without changing it. This issue can be reproduce by using the master-119c2c48-head version with out extra configuration. 
[...]

Issue #1234, opened 2021-09-16
[BUG] VMs are crashing when writing bulk data to additional volumes


janeczku:
**Describe the bug**

Running `dd` to write data to an additional volume causes the VM to crash.

**To Reproduce**
1. Create a VM with an additional volume attached. Configure VM Memory limit to 10GB.
2. Run`dd if=/dev/zero of=/dev/sda bs=1M conv=sync` on the volume and wait for 10-20 seconds
3. The VM crashes and is automatically restarted

**Expected behavior**

A VM should not crash when writing data to a secondary disc.

**Environment:**
 - Harvester ISO version: 0.3.0-preview

**Additional context**

It appears the qemu pod is OOM killed because it exceeds the memory limit of 10GB that Harvester has configured.

```
Sep 16 08:38:09 lpedge01003 kernel: oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=06ea4e4c81e3b77d6c6d4d150016783adb9476391a3d713e4ccf0180889a4557,mems_allowed=0-1,oom_memcg=/kubepods/pod4f6371a7-c7ea-416a-837a-a4692d874756,task_memcg=/kubepods/pod4f6371a7-c7ea-416a-837a-a4692d874756/06ea4e4c81e3b77d6c6d4d150016783adb9476391a3d713e4ccf0180889a4557,task=qemu-system-x86,pid=1526,uid=107
Sep 16 08:38:09 lpedge01003 kernel: Memory cgroup out of memory: Killed process 1526 (qemu-system-x86) total-vm:11226804kB, anon-rss:10439204kB, file-rss:24704kB, shmem-rss:4kB
Sep 16 08:38:09 lpedge01003 kernel: oom_reaper: reaped process 1526 (qemu-system-x86), now anon-rss:0kB, file-rss:12kB, shmem-rss:4kB
```

janeczku:
https://github.com/kubevirt/kubevirt/issues/4560

FrankYang0529:
I can reproduce the bug on harvester v0.3.0-preview1, but not on v0.2.0. After checking resources yaml, I find that v0.3.0-preview1 has `limits` for memory, but v0.2.0 doesn't. However, even if I add `limits` on the `VirtualMachine` resource and restart the VM, v0.2.0 still works fine.

I will try to create VM on KubeVirt directly with other storage provider and check whether it's upstream issue.

alexdepalex:
[...]

PR #1543, opened 2021-11-15
Simplify resource overcommit calculation

weihanglo:
**Problem:**
Simplify the overcommit calculation logic, as well as format the unit to MiB for memory.

**Solution:**

- Calculate `resources.requests` from `resource.limits` passed in from API client with overcommit percentage.
- Move [QEMU reservation logic from the frontend](https://github.com/harvester/dashboard/blob/171f5efdde2fd37f09e67f102a6ad86cbb8ed1a5/mixins/harvester-vm/index.js#L431-L436) to the backend and set `spec.domain.memory.guest` if needed.
- `pkg/builder/vm.go` now sets `resources.limits` instead of requests.

> Note that in order to display the value set by the user, dashboard UI need to change to display `resource.limits` of VM instead.

**Related Issue:**

- #1429 
- #1534
- #1537

**Test plan:**

1. Deploy a VM with the following manifest

```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  name: vm-cirros
spec:
  running: true
  template:
    spec:
      domain:
        devices:
          disks:
          - disk:
              bus: virtio
            name: rootdisk
        resources:
          limits:
            memory: 512Mi
            cpu: 1
      volumes:
      - containerDisk:
          image: cirros:0.5.2
        name: rootdisk
```

2. Check if 
    - The overcommitted `domain.resources.requests` is 341Mi.
    - The `domain.memory.guest` is about 412MiB.
    - The virt-launcher pod gets the correct resources limits and requests inherited from VMI.

### DIFF
--- a/pkg/builder/vm.go
+++ b/pkg/builder/vm.go
@@ -51,8 +51,9 @@ func NewVMBuilder(creator string) *VMBuilder {
 		Cores: defaultVMCPUCores,
 	}
 	resources := kubevirtv1.ResourceRequirements{
-		Requests: corev1.ResourceList{
+		Limits: corev1.ResourceList{
 			corev1.ResourceMemory: resource.MustParse(defaultVMMemory),
+			corev1.ResourceCPU:    *resource.NewQuantity(defaultVMCPUCores, resource.DecimalSI),
 		},
 	}
 	template := &kubevirtv1.VirtualMachineInstanceTemplateSpec{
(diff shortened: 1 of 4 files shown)


gitlawr:
`domain.memory.guest` is the amount visible in the VM. I would expect it to be 512Mi when I set 512Mi for the VM

weihanglo:
> `domain.memory.guest` is the amount visible in the VM. I would expect it to be 512Mi when I set 512Mi for the VM

As the dashboard already has [this feature](https://github.com/harvester/dashboard/blob/171f5efdde2fd37f09e67f102a6ad86cbb8ed1a5/mixins/harvester-vm/index.js#L431-L436), the PR here is simply moving the logic from the frontend to the backend. Though I have the same feeling as you, it might be somehow confusing to users. We can then add more documentations for how Harvester calculate resources overcommit.

gitlawr:
Get it. So 412MiB is not the overcommitted value, and is used for https://github.com/harvester/harvester/issues/1234#issuecomment-925628629

gitlawr (approved):
LGTM

pkg/builder/vm.go
```diff
@@ -51,8 +51,9 @@ func NewVMBuilder(creator string) *VMBuilder {
 		Cores: defaultVMCPUCores,
 	}
 	resources := kubevirtv1.ResourceRequirements{
-		Requests: corev1.ResourceList{
+		Limits: corev1.ResourceList{
```
guangbochen, line 54: Will need to bump the harvester [docker-machine-driver](https://github.com/harvester/docker-machine-driver-harvester/blob/master/go.mod#L11) to leverage the `builder pkg file` changes. cc @futuretea

guangbochen (approved):
LGTM

weihanglo:
I've opened a WIP PR for that https://github.com/harvester/docker-machine-driver-harvester/pull/27

pkg/builder/vm.go
weihanglo: I've opend a WIP PR for that 😆 
https://github.com/harvester/docker-machine-driver-harvester/pull/27

Issue #1458, opened 2021-10-21
[FEATURE] Supprt Network IP Pools config on the Harvester UI


guangbochen:
**Is your feature request related to a problem? Please describe.**
Before Harvester v1.2.0, users has to have to config the VIP CIDR pools manually using the `kubectl` command, we should add support to allow config on the dashboard.
```
apiVersion: v1
kind: ConfigMap
name: kubevip
namespace: kube-system
data:
  cidr-default: 192.168.0.200/29
  cidr-development: 192.168.0.210/29
  cidr-finance: 192.168.0.220/29
  cidr-testing: 192.168.0.230/29
```

https://docs.harvesterhci.io/v0.3/rancher/cloud-provider/#ipam
https://github.com/kube-vip/kube-vip-cloud-provider#global-and-namespace-pools

**Proposed Solution:**
- In Harvester v1.2.0, we have added a custom cloud provider of VIP, and users will be able to manage IP pools via `Networks -> IP Pools`, and it should allow users to:
  - config IP range via CIDR or range
  - config IP pools selector by VM network, namespace names
     -  multi-cluster UI will support additional selectors with project and guest cluster name

Note: for the design proposal please refer to https://github.com/harvester/harvester/blob/master/enhancements/20220214-harvester-cloud-provider-enhancement.md

janeczku:
The current implementation of Harvester's built-in load balancer feature requires editing a ConfigMap for setup and comes with a few other important limitations, including:

- Lacks **support for central management of VIP pools** via Rancher or Harvester API/UI
- Lacks **support for managing VIPs in different VLANs** (ConfigMap is not VLAN aware)
- Lacks **support for assigning multiple /32 VIP addresses per tenant**
- Lacks support for allocation of VIPs to specific downstream clusters
- **Traffic for different VIPs is not load balanced across multiple Harvester nodes**, creating a bandwidth bottleneck and SoP.

Expected State:

We need a more flexible and operatable VIP implementation with first class integration in Rancher and Harvester UI **that addresses the above limitations**.

The VIP Pool custom resource managed by user must have following attributes:

- mapped to a specific VLAN
- mapped to a specific Harvester tenant project/namespace

Multiple VIP pools may exist for same pair of VLAN and tenants. 

When provisioning a Harvester guest cluster in Rancher then a VIP is automatically assigned from a pool matching both the Harvester namespace *and* the VLAN network configured in the cluster options.

janeczku:
[...]

PR #1544, opened 2021-11-15
Add vip pools config syncer

guangbochen:

**Problem:**
Add support of VIP CIDR pool config on the Harvester UI settings

**Solution:**
Add support of VIP CIDR pool config on the Harvester UI settings

**Related Issue:**
https://github.com/harvester/harvester/issues/1458

**Test plan:**
- Config the CIDR pools via the UI settings(per namespace or `global`)
- Check if the user was able to create LoadBalancer type of service with the `pool` option in the guest k8s cluster, also the service should get the IP assigned correctly within the IP pools config.

### DIFF
--- a/pkg/controller/master/setting/vip_pools_config.go
+++ b/pkg/controller/master/setting/vip_pools_config.go
@@ -1,0 +1,70 @@
+package setting
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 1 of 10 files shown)


pkg/controller/master/setting/vip_pools_config.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+
+	poolsData := make(map[string]string, len(pools))
+	for ns, pool := range pools {
+		k := fmt.Sprintf("cidr-%s", ns)
+		poolsData[k] = pool
+	}
+
+	vipConfigmap, err := h.configmaps.Get(util.KubeSystemNamespace, KubevipConfigmapName, metav1.GetOptions{})
```
gitlawr: Get from cache

gitlawr (approved):
LGTM

PR #1545, opened 2021-11-16
Sync http proxy setting to rke2

gitlawr:
Now we can sync the proxy settings to rke2 by updating local cluster spec.

**Related Issue:**


guangbochen (commented):
Will this trigger the RKE2 cluster to reboot in order to apply proxy config properly? also if a reboot is needed will there be a downtime of the cluster or node?

gitlawr:
Rke2 restarts. Nodes don't reboot so it does not affect running services

guangbochen (approved):
LGTM

Issue #1549, opened 2021-11-16
[BUG] Unable to login after initial password setup


guangyee:
**Describe the bug**
Admin user unable to log back in after the initial password setup. Admin user was prompted to create a password at initial login. After creating a password and signed into the UI the first time, subsequent logins all result in errors.

We suspect this is due to recent upstream Rancher change in enforcing a default password policy (i.e. minimal of 12 characters) which is not fully reflected in the Harvester UI.

**To Reproduce**
1. PXE install Harvester in a Vagrant (https://github.com/harvester/ipxe-examples/tree/main/vagrant-pxe-harvester) or baremetal environment.
2. Access the Harvester UI at "https://<VIP>/" and set the initial admin password to "Password12345". Admin user is able to successfully login at this point.
3. Logout from the UI.
4. Try to re-login and see the failure.

**Expected behavior**
Admin user should be able to login.

**Support bundle**

**Environment:**
 - Harvester ISO version: master ISO
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): This is reproducible in both vagrant (https://github.com/harvester/ipxe-examples/tree/main/vagrant-pxe-harvester) and baremetal environment.

**Additional context**
Add any other context about the problem here
![Harvester_login_error](https://user-images.githubusercontent.com/1808417/142046538-7a4317a4-d179-4812-bcc2-c7d36cea88b5.png)
.
![Harvester_set_password](https://user-images.githubusercontent.com/1808417/142046549-95f07438-bfaf-4dfd-b617-808e3f225845.png)

gitlawr:
Workaround:
```
echo "apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: userattributes.management.cattle.io
spec:
  conversion:
    strategy: None
  group: management.cattle.io
  names:
    kind: UserAttribute
    listKind: UserAttributeList
    plural: userattributes
    singular: userattribute
  scope: Cluster
  versions:
  - name: v3
    schema:
      openAPIV3Schema:
        type: object
        x-kubernetes-preserve-unknown-fields: true
    served: true
    storage: true" | kubectl apply -f -
kubectl delete userattributes.management.cattle.io --all
```

gitlawr:
The userattributes CRD created by Harvester is incompatible with the current Rancher version.
```
# rancher created
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
[...]

PR #1552, opened 2021-11-17
Remove userattributes CRD creation

gitlawr:
Address https://github.com/harvester/harvester/issues/1549

Solution:
Remove the userattributes CRD creation. Use Rancher created CRD instead.


guangbochen (approved):
LGTM

PR #1554, opened 2021-11-17
Update `resources.lmits` for pre-defined VM templates

weihanglo:
**Problem:**
Pre-defined VM templates still specify `resources.requests`.

**Solution:**
Due to the change of resource overcommit feature, every pre-defined VM
templates now sepcify `resources.limits` instead of requests. As a
result, it can compute overcommit percentage correctly.

**Related Issue:**
#1537 

**Test plan:**
As #1537 described

### DIFF
--- a/pkg/data/template.go
+++ b/pkg/data/template.go
@@ -173,8 +173,9 @@ spec:
               - name: default
                 masquerade: {}
             resources:
-              requests:
+              limits:
                 memory: 2048Mi
+                cpu: 1
           networks:
           - name: default
             pod: {}
@@ -232,8 +233,9 @@ spec:
               - name: default
                 masquerade: {}
             resources:
-              requests:
+              limits:
                 memory: 2048Mi
+                cpu: 1
           networks:
           - name: default
             pod: {}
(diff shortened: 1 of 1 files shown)


weihanglo:
cc @WuJun2016 
Could you check that the editing of memory/cpu against `virtualmachinetemplateversion` would update `resources.limits` accordingly instead of `resources.requests`? That means when a user

- edits CPU: update values of `domain.cpu.cores` and `domain.resources.limits.cpu` in the payload. 
- edits memory: update values of `domain.resources.limits.memory` in the payload.

The behavior is just the same as editing `virtualmachine`. Thanks!

gitlawr:
If users create a VM without the memory limit from API, how do we show it in the UI?

WuJun2016:
@weihanglo  @gitlawr  I think the limits property should always be present on the UI, if the user creates it via yarm but uses the requests property to save cpu and memory, the backend should use the requests value to set the limits.  cc @guangbochen

weihanglo:
> if the user creates it via yarm but uses the requests property to save cpu and memory, the backend should use the requests value to set the limits.

Agree. Good advise. Would there be a case that a user sets `requests` but doesn't want any value on `limits`? IMO that is not useful in Harvester, since we have overcommit feature and users should always calculate and arrange resources carefully.

guangbochen (approved):
LGTM, both requests and limits should be required in Harvester, especially when leveraging the Rancher's resource quota feature.

gitlawr (approved):
LGTM

Issue #1497, opened 2021-11-01
[BUG] Reboot the deleted node and it will be added to the cluster again


guangbochen:
Related to https://github.com/rancher/rancher/issues/35258

weihanglo:
Confirmed that this is also reproducible via Harvester dashboard.

guangbochen:
We should consider cleaning up the Harvester node on node delete, but I just want to make sure the users are well notified as this action will wipe all the system components like rke2 and kubelet and can't be undone.
https://docs.rke2.io/install/linux_uninstall/

weihanglo:
We encounrted a situation that we cannot rely on pod/job to execute [standard RKE2 uninstallation](https://docs.rke2.io/install/linux_uninstall/) procedure: When one deletes the node via K8s API, any further pod/job is not expected to schedule on the node.

As a result, we tried to leverage [rancher-system-agent](https://github.com/rancher/system-agent), a systemd unit, to deal with uninstallation outside K8s. However, the `Plan` structure and logic are more like implementation details which we may not want to depend on. Moreover, if the node is already shutdown, Neither can we delete node nor do do any operation on it via neither system-agent.

---

We here propose another way to resolve the issue.

1. When deleting a node via Dashboard UI, label the node as schedule-to-delete, such as `harvesterhci.io/schedule-to-delete: "true"`
2. A node controller watches that label and then does followings:
    1. Creates a job to perform the standard RKE2 uninstallation.
    2. After the uninstallation, delete the node resource as normal.
4. If the node is shutdown or not scheduable, the job won't be able to schedule. Therefore, the node will be in a `deleting` status and only after the node is booted, the job would be scheduled and executed.

There is a obvious caveat in this approach. One cannot delete prevent re-joining if they deletes the node via standard `DELETE` API, they then should label the node instead.

If aforementioned method seems not appropriate, we may just document the `node deletion` process and info users that if they want to delete a node, they should disable RKE2 systemd unit manually. This is also acceptable to me since deletion is actually a destructive operation that needs to take care of and system admin **should** read the manual.

weihanglo:
We've tested some workarounds to resolve this issue but all went stuck.

The first attempt is leverage the finalizer feature in K8s. We added a finalizer on node and it'll spawn a job to uninstall RKE2 on the node to delete. Unfortunately there is also other logic that disconnect the node from the cluster network, so the node became `unreachable` and the pod scheduler cannot schedule any pod onto it.

[...]

PR #1566, opened 2021-11-19
Remove RKE2 config file to prevent deleted node from rejoining

### LINKED ISSUE #2144: Support node hard delete method of the RKE2 node
**Environmental Info:**
RKE2 Version: v1.22.3+rke2r1

**Is your feature request related to a problem? Please describe.**
- delete one of the nodes from the cluster either via kubectl delete node or through the Rancher API
- reboot the deleted node and it will be added to the cluster again
- expecting a node hard delete methodology that it will not join the cluster again after node reboot

RKE2 related issue: https://github.com/rancher/rke2/issues/1959 may also get addressed via the node hard delete method.

Other related issues: 
https://github.com/rancher/rancher/issues/35258
https://github.com/harvester/harvester/issues/1497

weihanglo:
**Problem:**
A deleted node will re-join the cluster after rebooting it.

**Solution:**

> This is a temporary workaround. We're waiting for the upstream fix https://github.com/rancher/rke2/issues/2144.

To hard delete a node from a cluster, we need to uninstall or disable RKE from the machine permanently. However, there are two obstacles when dealing with the node deletion:

1. The node might be down when a user request a deletion, so we cannot execute the uninstallation scipt.
2. We cannot request standard `delete` kube-api before executing uninstallation script, since the node will become unreachable/unschedulable and none of any Job/Pod can schedule to perform further action on the machine.

The temporary fix here tries to **remove RKE2 config file** to prevent the node from rejoining after rebooting

The whole delete process is like followings:

1. A user deletes the node via kube-api or dashboard UI.
1. The `harvesterhci.io/delete-node-controller` finalizer prevents node from delete
1. The `delete-node-controller` spawns a job to remove RKE2 config file (/etc/rancher/rke2/config.yaml.d/50-rancher.yaml), and then take off the finalizer from the node
1. The node resource now is succeeded to delete.

**Related Issue:**

#1497

**Test plan:**
1. Create a multi-node Harvester cluster.
2. Build the harvester server image and replace the old one.
3. Delete a node via either dashboard UI or kube-api
4. Login the machine of the deleted node. Check if `etc/rancher/rke2/config.yaml.d/50-rancher.yaml` is removed.
5. Reboot the node and wait few minutes. The rebooted node must not rejoin the cluster

### DIFF
--- a/pkg/controller/master/node/delete_controller.go
+++ b/pkg/controller/master/node/delete_controller.go
@@ -1,0 +1,241 @@
+package node
+
+import (
+	"context"
+
+	ctlbatchv1 "github.com/rancher/wrangler/pkg/generated/controllers/batch/v1"
+	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/rancher/wrangler/pkg/name"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/utils/pointer"
+
+	"github.com/harvester/harvester/pkg/config"
+	v1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+)
+
+const (
+	HarvesterDeleteNodeLabelKey = "harvesterhci.io/delete-node"
+	deleteImage                 = "busybox:1.32.0"
+	deleteNodeControllerName    = "delete-node-controller"
+	deleteNodeFinalizer         = "harvesterhci.io/delete-node-controller"
+	deleteRootMountPath         = "/host"
+	rke2ConfigFilePath          = "/etc/rancher/rke2/config.yaml.d/50-rancher.yaml"
+)
+
+// deleteNodeHandler handles the finalization of `harvesterhci.io/delete-node-controller` finalizer.
+// At this moment, it does the following tasks:
+//
+// - Add finalizer `harvesterhci.io/delete-node-controller` onto nodes.
+// - Spawn job `harvester-delete-node-<node-name>` to remove RKE2 config file.
+// - Delete the complete job `harvester-delete-node-<node-name>` and remove the finalizer.
+type deleteNodeHandler struct {
+	nodes                       ctlcorev1.NodeClient
+	nodeCache                   ctlcorev1.NodeCache
+	virtualMachineInstanceCache v1.VirtualMachineInstanceCache
+	jobs                        ctlbatchv1.JobClient
+	jobCache                    ctlbatchv1.JobCache
+	namespace                   string
+}
+
+// DeleteRegister registers the node controller
+func DeleteRegister(ctx context.Context, management *config.Management, options config.Options) error {
+	nodes := management.CoreFactory.Core().V1().Node()
+	jobs := management.BatchFactory.Batch().V1().Job()
+	vmis := management.VirtFactory.Kubevirt().V1().VirtualMachineInstance()
+	deleteNodeHandler := &deleteNodeHandler{
+		nodes:                       nodes,
+		nodeCache:                   nodes.Cache(),
+		jobs:                        jobs,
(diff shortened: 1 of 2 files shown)


weihanglo:
Unfortunately, the approach "spawning a job to execute `/usr/local/bin/rke2-uninstall.sh`" failed. I guess that's because the uninstall script disables rke2-server systemd service first, so the node became `NotReady` and the job being killed immediately without completion. 

That being said, we cannot leverage a K8s `Job` to run the uninstall script. That's all I perceived so far. We need a mechanism outside K8s cluster.

gitlawr:
We don't need to run the rke2-uninstall script. We can simply remove the rke2 config file(/etc/rancher/rke2/config.yaml.d/50-rancher.yaml) so that it won't be able to rejoin after restart.

gitlawr:
I think we don't need to do it in custom API action. The controller can work as a finalizer so that the node deletion will wait until it completes.

weihanglo:
> We don't need to run the rke2-uninstall script. We can simply remove the rke2 config file(/etc/rancher/rke2/config.yaml.d/50-rancher.yaml) so that it won't be able to rejoin after restart.

Great idea. Maybe we can remove that file and then delete the node.


> I think we don't need to do it in custom API action. The controller can work as a finalizer so that the node deletion will wait until it completes.

Not really. I've tried finalizer way. Once you delete a node, it become unreachable (tainted) and cannot schedule pods no matter the progress of finalizers.

gitlawr:
What taint is it? Can we add tolerance to the cleanup pod?

weihanglo:
~~I don't have the environment at hands. I remember there are `node.kubernetes.io/network-unavailable` and `node.kubernetes.io/unreachable`. Are we able to schedule pod on a node with unreachable network?~~

It wouldn't be tainted. My fault. The finalizer way is good to go

weihanglo:
Thank @gitlawr for providing a great solution 👍🏾. This is ready for test.

pkg/controller/master/node/delete_controller.go
```diff
@@ -0,0 +1,244 @@
...
+
+	return job, nil
+}
+
+func (h *deleteNodeHandler) createDeleteNodeJob(node *corev1.Node) (*batchv1.Job, error) {
+	job := createDeleteNodeJob(h.namespace, node)
+	return h.jobs.Create(job)
+}
+
+func (h *deleteNodeHandler) removeFinalizer(node *corev1.Node) (*corev1.Node, error) {
```
gitlawr, line 126: We have the finalizer operation logic in the framework, can we use that?
https://github.com/rancher/wrangler/blob/master/pkg/generic/remove.go#L39-L56

In short, an onRemove handler adds the finalizer and removes it when it succeeds.

weihanglo: So, for reusing the finalizer logic from the framework, the controller needs to emit an error for await the job removing RKE2 file. And then the job updates some label on the node in order to  re-run `OnRemove` handler.

It seems awkward to me if we emits an error only for the sake of reusing the implicit finalizer logic. Is it a conventional use case of extending finalizer `wrangler.cattle.io/xxxx`? My first impression is that it's a basic protection for any "real error" inside `OnRemove`, but not a fake error we are going to emit.

gitlawr: Make sense. I'm good with the current approach

pkg/controller/master/node/delete_controller.go
```diff
@@ -0,0 +1,244 @@
...
+							},
+							ImagePullPolicy: corev1.PullIfNotPresent,
+							SecurityContext: &corev1.SecurityContext{
+								Privileged: pointer.BoolPtr(true),
+							},
+						},
+					},
+					HostIPC:     true,
+					HostPID:     true,
+					HostNetwork: true,
```
gitlawr: Are these configurations necessary?

weihanglo: No. Removed. Thanks for the catch!

gitlawr:
@weihanglo Can you check the CI failure?

weihanglo:
@gitlawr Thanks for the reminder. Just fixed lint errors. Should be good to go.

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

Issue #1555, opened 2021-11-17
[BUG] Backup target endpoint processing code has a bug


w13915984028:
**Describe the bug**

The current code causes panic when use "Use the default value" button

**To Reproduce**
1. Set anything in setting backup target
2. Edit again, and hit "Use the default value" (it must be usable, then can produce)
3. Harvester pod has a crash reported

**Expected behavior**

No crash

**Support bundle**

Existing code has a bug:

```
func (h *TargetHandler) validateTargetEndpoint(target *settings.BackupTarget) (*settings.BackupTarget, error) {
switch target.Type {
	case settings.NFSBackupType:
..
		return target, validateNFSBackupTarget(target.Endpoint)
	case settings.S3BackupType:
..
		return target, err
	default:
		return `nil`, fmt.Errorf("unknown type of the backup target, currently only support NFS and S3")
}

func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {

	target, err = h.validateTargetEndpoint(target)
	if err != nil {
		logrus.Errorf("invalid backup target, error: %s", err.Error())
		return h.updateBackupTargetSetting(settingCpy, target, err)
	}

func (h *TargetHandler) updateBackupTargetSetting(setting *harvesterv1.Setting, target *settings.BackupTarget, err error) (*harvesterv1.Setting, error) {
..
	// reset the s3 credentials to prevent controller reconcile and not to expose secret key
	target.SecretAccessKey = ""
```

`OnBackupTargetChange  `should not call h.`updateBackupTargetSetting`() when err != nil

`target `is nil, and target.xxx cause panic

```
/go/src/github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x86
panic(0x1f0e060, 0x36c8b70)
        /usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/harvester/harvester/pkg/controller/master/backup.(*TargetHandler).updateBackupTargetSetting(0xc000198d20, 0xc002adc180, 0x0, 0x24c5c00, 0xc000eda2f0, 0x1, 0x1, 0x7fe1b944afff)
        /go/src/github.com/harvester/harvester/pkg/controller/master/backup/backupTaget.go:114 +0x7a
[...]

Issue #1051, opened 2021-07-28
[BUG] Error message about invalid backup target doesn't go away


noahgildersleeve:
**Describe the bug**

If you get an error message about invalid backup target it doesn't go away when you remove the backup target
**To Reproduce**
1. Create new backup target with S3
2. Enter in bogus info like `7` for each entry
3. Save
4. Refresh page
5. Look for error
6. Edit config again and remove all entries
**Expected behavior**

If you remove the entries for the backup target it should clear the error
**Support bundle**

[harvester-supportbundle_0278c74f-2ced-4d9a-ae93-7776ce875960_2021-07-28T22-01-11Z.zip](https://github.com/harvester/harvester/files/6896692/harvester-supportbundle_0278c74f-2ced-4d9a-ae93-7776ce875960_2021-07-28T22-01-11Z.zip)

**Environment:**
 - Harvester ISO version: master-bc40fc79-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): baremetal with DL160

**Additional context**
![Greenshot_2021-07-28_15_00_30](https://user-images.githubusercontent.com/83787952/127402173-e46cc880-7562-4987-87af-e1a05525dfc7.jpg)

guangbochen:
will need to reset the config using the  `Use the default value` option, this will require both UI and backup modification support. cc @WuJun2016

w13915984028:
When the backup target config is failed, the error will always be there, to notify the user.
User can modify the config to a correct one, or clear/reset the error config. The latter does not work now.

We may  either add a new button beside "show/hide backup target", called "reset/clear backup target";  or in the "edit" pop up, enable "Use the default value" button (it is not usable now), to reset/clear current config.

"Use the default value" button looks to be a better option. UI is checking why it is disabled.
@guangbochen @WuJun2016 @n313893254

w13915984028:
The "Use the default value" button in the harvester UI (edit backup target) is available now, the corresponding backend source code must be added to do related processing. Coding.

DON´T use/test the button at the moment.

w13915984028:
With "Use the default value" button, and the coming commit, we can set the backup target to initial value, the error message will disappear.
Sure, it also needs a fix regarding #1555, it will be committed together.

w13915984028:
After PR of #1447, the error message will be shown only once, will not be always there.

And, when the UI encounters error, the config will not be saved.  After refresh, you will still find the previous successful config or initial config.

[...]

PR #1567, opened 2021-11-19
Fix set backup target to default and a panic

w13915984028:

**Problem:**

Error message about invalid backup target doesn't go away #1051, it is a hint about failure in setting,
There is also no method to reset the backup target to default value.

**Solution:**

1. UI enable the "Use the default value" button
2. Backend add processing of the default value, to set backup target back

**Related Issue:**
#1051
#1555

**Test plan:**

1. Add any invalid backup target, S3 or NFS are both OK,
2. The error message will always be there,
3. In case the "Use the default value" button is usable, and use it
  Before this commit, the harvester pod will crash. After this commit, it is fixed
  And, the backup target will be restored to default, the error message will be gone.

### DIFF
--- a/pkg/controller/master/backup/backupTaget.go
+++ b/pkg/controller/master/backup/backupTaget.go
@@ -91,7 +91,7 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
 	target, err = h.validateTargetEndpoint(target)
 	if err != nil {
 		logrus.Errorf("invalid backup target, error: %s", err.Error())
-		return h.updateBackupTargetSetting(settingCpy, target, err)
+		return nil, err
 	}
 
 	if err = h.updateLonghornTarget(target); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/master/backup/backupTaget.go
```diff
@@ -91,7 +91,7 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
 	target, err = h.validateTargetEndpoint(target)
 	if err != nil {
 		logrus.Errorf("invalid backup target, error: %s", err.Error())
-		return h.updateBackupTargetSetting(settingCpy, target, err)
+		return nil, err
```
guangbochen, line 96: should return the error if it is an invalid config and leverage the [setting handler](https://github.com/harvester/harvester/blob/f5bde7fa462c09d0513b8f370f71f2dd974f298c/pkg/controller/master/setting/handler.go#L64-L69) to update the configured status properly.

w13915984028:
A new PR is opened after rebase the source code:
https://github.com/harvester/harvester/pull/1576

guangbochen:
close in favor of #1576

PR #1573, opened 2021-11-22
Fix invalid vip data

guangbochen:

**Problem:**
Should update the data with `cidr-` prefix.

**Solution:**
Update configmap data with the correct one.

**Related Issue:**
https://github.com/harvester/harvester/issues/1458

**Test plan:**


gitlawr (approved):
LGTM

Issue #1562, opened 2021-11-18
[BUG]Inconsistency in API HeadBucket between S3 and Minio


w13915984028:
**Describe the bug**

After the fix of #1339 , it works with S3, but minio replacement breaks.

**To Reproduce**
1. Edit backup target to "S3", with a minio installed in cluster or independently, edit config like 
{
  "type": "s3",
  "endpoint": "http://192.168.3.31:9000",
  "accessKeyId": "",
  "secretAccessKey": "",
  "bucketName": "backupbucket",
  "bucketRegion": "us-east-1",
  "cert": "",
  "virtualHostedStyle": true
}

it will report error:
operation error S3: HeadBucket, exceeded maximum number of attempts, 3, https response error StatusCode: 0, RequestID: , HostID: , request send failed, Head "http://backupbucket.192.168.3.31:9000/": dial tcp: lookup backupbucket.192.168.3.31 on 10.53.0.10:53: no such host      

**Expected behavior**

The minio replacement of S3 should still work.

**Support bundle**

**Environment:**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):
Baremetal Server, single node harvester, master-head ISO

**Additional context**

related issues:
#1051
#1339

w13915984028:
A quick analysis, 

When "HeadBucket" is called and the "Endpoint" is not NULL, AWS sdk will try to build an URI,  here it becomes "http://backupbucket.192.168.3.31:9000/":

But the DNS does not know it. It causes the error.

Before commit of #1339, the "ListBuckets" API is called, it does not care about bucket name,  the credential is used to get all buckets. the URL is not changed.

S3 endpoint is in different format comparing with minio, need to consider how to identify them.

The worst solution is to try "HeadBucket" first, then "ListBucket", if both fail, then report fail.

w13915984028:
Planed solution is:

check the endpoint, if it is not empty and not contains "amazonaws.com", assume it is NONE s3, then call ListBuckets, 
in other cases, call HeadBucket

It is verified locally.

```
awss3EndpointKeyword = "amazonaws.com"
       if target.Endpoint != "" && !strings.Contains(target.Endpoint, awss3EndpointKeyword) {
              // s3 replacement, like minio
               output, err := client.ListBuckets(h.ctx, &s3.ListBucketsInput{})
               if err != nil {
                       return err
           } else {
             // aws s3
[...]

Issue #1339, opened 2021-09-30
[BUG] Backup-target failing for the harvester-ci-testing bucket


chajain:
**Describe the bug**
The backup-target in settings is failing when trying to use the aws account credentials provided for harvester-ci-testing bucket. It fails with error
backup-target Modified
Custom back target to store VM backups.
{
  "type": "s3",
  "endpoint": "",
  "accessKeyId": "",
  "secretAccessKey": "",
  "bucketName": "harvester-ci-testing",
  "bucketRegion": "us-east-2",
  "cert": "",
  "virtualHostedStyle": false
}
operation error S3: ListBuckets, https response error StatusCode: 403, RequestID: E0NEZH6VAXNEEWCT, HostID: lePh+/9xeaHXfS6WLYF6JHqfmk1NXfdd3BJDLZXLSBBxgqkkraaaYMOBwQF3SVXc7INtUJLY2BybQzLIj8bl+g==, api error AccessDenied: Access Denied

I performed the following steps on Longhorn UI and backup/restore seems to work there fine.

- Created aws-secret on harvester-node-0 using kubectl command

- Did settings on longhorn UI
      Backup Target - s3://harvester-ci-testing@us-east-2/
      Backup Target Credential Secret- aws-secret

- Created a backup and restored and it is successful

**To Reproduce**
Try to modify the backup-target in settings with aws credentials of harvester-ci-testing bucket. 

**Expected behavior**
backup-target should be created without any errors

**Support bundle**

**Environment:**
 - Harvester ISO version: master
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): ipxe vagrant harvester setup

**Additional context**

yasker:
It seems Harvester is performing a connection check that requires additional permission for S3 than necessary.

w13915984028:
The code 
` validateS3BackupTarget ` in `backupTaget.go`
tries to verify if the bucket is existing,
as per S3 API, HeadObject (longhorn uses) or HeadBucket is the proper API to use, it is high efficiency, and will not cause potential "AccessDenied",  harvester uses ListBucket.

HeadObject is a general API to check if a single object exists.
----
AWS has 3 APIs related here:
```
// * If you have the
// s3:ListBucket permission on the bucket, Amazon S3 returns an HTTP status code
// 404 ("no such key") error.
//
// * If you don’t have the s3:ListBucket permission,
// Amazon S3 returns an HTTP status code 403 ("access denied") error.

[...]

PR #1575, opened 2021-11-22
Fix inconsistency in API HeadBucket between S3 and Minio

w13915984028:

**Problem:**

After the fix of #1339 , it works with S3, but minio replacement breaks.
HeadBucket API is not supported by minio at the moment.

**Solution:**

Check the endpoint, if it is not empty and not contains "amazonaws.com", assume it is NONE s3, then call ListBuckets,
in other cases, call HeadBucket

**Related Issue:**
#1562 
**Test plan:**

1. start up (single node) harvester cluster
2. edit backup target, with a valid s3 bucket, save return no error
3. edit backup target, with an invalid s3 bucket, save return related error message
4. setup an minio, in cluster or independent
5. edit backup target, endpoint points to minio, with a valid s3 bucket, save return no error
6. edit backup target, endpoint points to minio, with an invalid s3 bucket, save return related error message

notice, error message are different between s3 and minio, please refer info listed in #1562

### DIFF
--- a/pkg/controller/master/backup/backupTaget.go
+++ b/pkg/controller/master/backup/backupTaget.go
@@ -38,6 +38,8 @@ const (
 
 	longhornBackupTargetSettingName       = "backup-target"
 	longhornBackupTargetSecretSettingName = "backup-target-credential-secret"
+
+	awss3EndpointKeyword = "amazonaws.com"
 )
 
 // RegisterBackupTarget register the setting controller and validate the configured backup target server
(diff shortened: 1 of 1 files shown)


guangbochen:
The validation check is moved to the [webhook validator](https://github.com/harvester/harvester/blob/83b44c164437554c4eb4da6cd9aba57f2b19e231/pkg/webhook/resources/setting/validator.go#L126-L165) of https://github.com/harvester/harvester/pull/1447, @w13915984028 can u please revalidate it again, thanks.

w13915984028:
> The validation check is moved to the [webhook validator](https://github.com/harvester/harvester/blob/83b44c164437554c4eb4da6cd9aba57f2b19e231/pkg/webhook/resources/setting/validator.go#L126-L165) of #1447, @w13915984028 can u please revalidate it again, thanks.

Validate:

It is improved, no inconsistency between s3 and minio.

PR #1576, opened 2021-11-22
Fix set backup target to default and a panic

w13915984028:

**Problem:**

Error message about invalid backup target doesn't go away #1051, it is a hint about failure in setting,
There is also no method to reset the backup target to default value.

**Solution:**

1. UI enable the "Use the default value" button (already in master)
2. Backend add processing of the default value, to set backup target back

**Related Issue:**
#1051
#1555

**Test plan:**

Add any invalid backup target, S3 or NFS are both OK,
The error message will always be there,
In case the "Use the default value" button is usable, and use it
Before this commit, the harvester pod will crash. After this commit, it is fixed
And, the backup target will be restored to default, the error message will be gone.

### DIFF
--- a/pkg/controller/master/backup/backupTaget.go
+++ b/pkg/controller/master/backup/backupTaget.go
@@ -91,7 +91,7 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
 	target, err = h.validateTargetEndpoint(target)
 	if err != nil {
 		logrus.Errorf("invalid backup target, error: %s", err.Error())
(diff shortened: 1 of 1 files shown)


guangbochen:
please rebase the PR of https://github.com/harvester/harvester/pull/1447

w13915984028:
After PR of #1447, the "Use the default value" will be refused by webhook, the error message is:

`admission webhook "validator.harvesterhci.io" denied the request: Destination URL hasn't been specified`

w13915984028:
After PR of #1447, this PR is outdated. Close.

PR #1579, opened 2021-11-23
Fix http proxy

gitlawr:
The PR resolves the following issues:
1. When the HTTP proxy setting is configured before the backup target secret is created, the proxy is not set in the backup secret.
2. Fail to import Harvester because the proxy envs are not applied in custom HTTP transport
3. When Harvester is imported, embedded Rancher needs to connect to upstream Rancher for aggregation API, therefore we need to inject proxy envs to Rancher pods as well.

**Related Issue:**


guangbochen:
will need to rebase due to https://github.com/harvester/harvester/pull/1447

guangbochen (approved):
LGTM

Issue #1524, opened 2021-11-08
[BUG] Can't create support bundle if one node is off


noahgildersleeve:
**Describe the bug**

You can't complete generating a support bundle if one node is turned off
**To Reproduce**
1. On a multi-node harvester cluster power off one node
2. Navigate to support and try to complete generating support bundle
**Expected behavior**

Support bundle creates successfully.
**Support bundle**

**Environment:**
 - Harvester ISO version: master-a412145f-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): DL160

**Additional context**
 If you wait it will eventually time out. If you turn the node back on it works
![Greenshot 2021-11-08 12 11 32](https://user-images.githubusercontent.com/83787952/140827121-40bfc020-c3b9-46bc-9119-21c9b62bec94.png)
![Greenshot 2021-11-08 12 11 38](https://user-images.githubusercontent.com/83787952/140827124-d8baa714-69c8-42d1-bd81-d0ffd5b2cd05.png)
![Greenshot 2021-11-08 12 17 44](https://user-images.githubusercontent.com/83787952/140827127-7ca87cfc-5eec-4b24-b1be-02afc98eccfd.png)

noahgildersleeve:
Tested in master-f510945f-head on 2 node virtual environment in KVM with one node down. support bundle downloaded successfullly. Closing
[supportbundle_365f6006-bfee-4eaa-a1d0-894b3d69a15a_2021-12-01T00-07-09Z.zip](https://github.com/harvester/harvester/files/7629650/supportbundle_365f6006-bfee-4eaa-a1d0-894b3d69a15a_2021-12-01T00-07-09Z.zip)
.

PR #1591, opened 2021-11-29
Bump support-bundle-kit to v0.0.4

weihanglo:
**Problem:**
Can't create support bundle if one node is off

**Solution:**
rancher/support-bundle-kit#27

**Related Issue:**
#1524
harvester/harvester-installer#176

**Test plan:**
1. Create a multi-node cluster
2. Shutoff one node
3. Should download the support bundle successfully


guangbochen (approved):
LGTM

Issue #1585, opened 2021-11-25
[BUG] Support Bundle failed to create due to timeout


guangbochen:
**Describe the bug**
Support Bundle failed to create due to timeout，but the file has generated successfully on the support bundle's pod after a period of time.

**To Reproduce**
1. Create a support bundle via the support page
2. Wait for a period of time and get the wait timeout error 
```
# SupportBundle CR
status": {
"conditions": [
{
"error": true,
"lastUpdateTime": "2021-11-25T04:45:38Z",
"message": "fail to generate supportbundle: timeout",
"status": "False",
"transitioning": false,
"type": "Initialized"
}
],
"progress": 40,
"state": "error"
}
```

**Expected behavior**
- Increase the default timeout to 10 minutes as it will take a longer time for people having a larger cluster or cluster containing a long historical data
- Make the timeout configurable via setting or should only return error when the pod contains fatal.

**Support bundle**
```
 ks logs -f supportbundle-manager-bundle-wkoos-7dc8c58455-2qd5g
+ exec tini -- /usr/bin/support-bundle-kit manager
time="2021-11-25T04:37:36Z" level=info msg="running phase init"
time="2021-11-25T04:37:36Z" level=debug msg="Create a local state store. (harvester-system/bundle-wkoos)"
time="2021-11-25T04:37:36Z" level=debug msg="Get supportbundle harvester-system/bundle-wkoos state generating"
time="2021-11-25T04:37:36Z" level=info msg="succeed to run phase init. Progress (20)."
time="2021-11-25T04:37:36Z" level=info msg="running phase cluster bundle"
time="2021-11-25T04:37:36Z" level=debug msg="generating cluster bundle..."
time="2021-11-25T04:37:36Z" level=debug msg="Get supportbundle harvester-system/bundle-wkoos"
W1125 04:37:36.081898       7 warnings.go:70] v1 ComponentStatus is deprecated in v1.19+
W1125 04:37:36.103189       7 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
W1125 04:37:37.503395       7 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W1125 04:37:37.974792       7 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
[...]

PR #1592, opened 2021-11-29
feat(setting): add support bundle timeout

FrankYang0529:
**Problem:**
Support Bundle failed to create due to timeout, but the file has generated successfully on the support bundle's pod after a period of time.

**Solution:**
Add support bundle timeout setting. If users input 0, we disable the timeout.

**Related Issue:**
https://github.com/harvester/harvester/issues/1585

**Test plan:**

Case 1 - Support bundle timeout webhook
* Set `support-bundle-timeout` setting as a negative integer string. We should get an error response.
* Set `support-bundle-timeout` setting as an invalid integer string. We should get an error response.
* Set `support-bundle-timeout` setting as an empty string. We should pass.
* Set `support-bundle-timeout` setting as a positive integer string. We should pass.

Case 2 - Support bundle controller
* Set `support-bundle-timeout` setting as "1" and generate a support bundle. We shouldn't get any timeout error if the duration is less than 1 minute.
* Set `support-bundle-timeout` setting as "0" and generate a support bundle. We shouldn't get any timeout error.

### DIFF
--- a/pkg/controller/master/supportbundle/controller.go
+++ b/pkg/controller/master/supportbundle/controller.go
@@ -2,6 +2,7 @@ package supportbundle
 
 import (
 	"fmt"
+	"strconv"
 	"time"
 
 	ctlappsv1 "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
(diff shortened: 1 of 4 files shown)


guangbochen (approved):
LGTM, @WuJun2016 will need to add the new setting of  `support-bundle-timeout` after this PR is merged, and the description will be `Support Bundle timeout config in minutes, use 0 to disable the timeout.

bk201 (approved):
LGTM

PR #1593, opened 2021-11-29
 Update backup target validator

gitlawr:
Problem:
Backup target validation in webhook uses proxy and certificate from env on initialization. Changing them afterwards has no effects.

Solution: Customize HTTP transport on backup target validation.

**Related Issue:**
https://github.com/harvester/harvester/issues/1260


guangbochen (approved):
LGTM

FrankYang0529 (approved):
LGTM

PR #1594, opened 2021-11-29
fix 1051: reset backup target to default

w13915984028:

**Problem:**

After PR of #1447, the "Use the default value" of "backup target" will be refused by webhook, the error message is:

`admission webhook "validator.harvesterhci.io" denied the request: Destination URL hasn't been specified`

It is not possible to reset backup target to the default value in the web UI.

**Solution:**

In webhook, check the backup target, if it is the default value then allow it pass.

**Related Issue:**
#1051 

**Test plan:**

1. Setup a harvester cluster (one node is also OK)
2. Set the backup target
2.1 When backup target has error (endpoint is not reachable/bucket name is wrong/bad credentials...), "save" will be refused with an error message, and after refreshing the web page, the backup target is still the last setting
2.2 When backup target is correct, it is saved. "Edit" again, click "Use the default value" button and "save", it will be success, and backup target is reset to default value.


FrankYang0529:
We may also need to update `pkg/controller/master/backup/backup_metadata.go` to skip syncing remote VM Backup with default backup target setting.

PR #1600, opened 2021-12-01
Bump harvester network controller 0.2.5

yaocw2020:


deploy/charts/harvester/values.yaml
```diff
@@ -341,13 +341,13 @@ jobs:
...
-  enabled: false
+  enabled: true
 
   image:
     repository: rancher/harvester-network-controller
-    pullPolicy: Always
+    pullPolicy: IfNotPresent
     # Overrides the image tag whose default is the chart appVersion.
-    tag: master-head
+    tag: v0.2.5
```
guangbochen: those changes are not needed as we will tag the version on the tag branch, but I think the helper image values are missing.

deploy/charts/harvester/values.yaml
```diff
@@ -349,6 +349,11 @@ harvester-network-controller:
     # Overrides the image tag whose default is the chart appVersion.
     tag: master-head
 
+harvester-network-helper:
+  image:
+    repository: rancher/harvester-network-helper
+    tag: master-head
```
guangbochen: should move it underneath the `harvester-network-controller:`

guangbochen (approved):
LGTM

PR #1601, opened 2021-12-01
Fix validation error on empty HTTP proxy

gitlawr:
https://github.com/harvester/harvester/issues/1260#issuecomment-983328139

Fix the unmarshal error in validation webhook when http proxy is empty.
Improve error message to provide the context.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -179,11 +179,13 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error 
 func (v *settingValidator) customizeTransport() error {
 	httpProxySetting, err := v.settingCache.Get(settings.HttpProxySettingName)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get HTTP proxy setting: %v", err)
 	}
 	var httpProxyConfig util.HTTPProxyConfig
-	if err := json.Unmarshal([]byte(httpProxySetting.Value), &httpProxyConfig); err != nil {
-		return err
+	if httpProxySetting.Value != "" {
+		if err := json.Unmarshal([]byte(httpProxySetting.Value), &httpProxyConfig); err != nil {
+			return fmt.Errorf("failed to parse HTTP proxy config: %v", err)
+		}
 	}
 	os.Setenv(util.HTTPProxyEnv, httpProxyConfig.HTTPProxy)
 	os.Setenv(util.HTTPSProxyEnv, httpProxyConfig.HTTPSProxy)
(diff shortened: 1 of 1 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -179,19 +179,21 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
...
 	}
 	var httpProxyConfig util.HTTPProxyConfig
-	if err := json.Unmarshal([]byte(httpProxySetting.Value), &httpProxyConfig); err != nil {
-		return err
+	if httpProxySetting.Value != "" {
+		if err := json.Unmarshal([]byte(httpProxySetting.Value), &httpProxyConfig); err != nil {
+			return fmt.Errorf("failed to parse HTTP proxy config: %v", err)
+		}
 	}
 	os.Setenv(util.HTTPProxyEnv, httpProxyConfig.HTTPProxy)
```
w13915984028, line 190: Should those setting be put right before the final successful return? when any of following error happens, the ENV is set in a middle state.

pkg/webhook/resources/setting/validator.go
```diff
@@ -179,19 +179,21 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
...
 	os.Setenv(util.HTTPSProxyEnv, httpProxyConfig.HTTPSProxy)
 	os.Setenv(util.NoProxyEnv, util.AddBuiltInNoProxy(httpProxyConfig.NoProxy))
 
 	caSetting, err := v.settingCache.Get(settings.AdditionalCASettingName)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get additional CA setting: %v", err)
 	}
 	if caSetting.Value != "" {
```
w13915984028, line 198:
```
	customTransport, ok := http.DefaultTransport.(*http.Transport)
	if ok {
	
	} 
```
	
	needs an else , or check !OK first.

pkg/webhook/resources/setting/validator.go
gitlawr: It is expected to use the configured proxy envs. We don't need to ensure it is set only when it succeeds

pkg/webhook/resources/setting/validator.go
gitlawr: Actually we don't expect !ok to happen. Use the ok check for secure type assertion

guangbochen (approved):
LGTM

FrankYang0529 (approved):
LGTM

Issue #1599, opened 2021-12-01
[BUG] Extra disk auto provision from installation may cause NDM can't find a valid longhorn node to provision


TachunLin:
**Describe the bug**

While testing disk auto provision feature for #1241  

In `BIOS` firmware + `No MBR` (Default) + `Auto disk provisioning` config, 
After the 2nd harvester node (Join mode) installation complete. 
Extra disk `sda` and `sdc` both display error message on host disk page

![image](https://user-images.githubusercontent.com/29251855/144076347-be8c4ee0-2684-46be-b7a2-417580d0f4d4.png)

* Disk usage information:
  - /dev/`sdb`: Used for harvester installation (150GB IDE)
  - /dev/`sda`: Extra disk for auto provision (60GB SATA)
  - /dev/`sdb`: Extra disk for auto provision (40GB SCSI)

```
failed to mount the device /dev/sdc1 to path /var/lib/harvester/extra-disks/2bf5824103011662c7c505aa26dc372b, error: Failed
to execute: nsenter [--mount=/host/proc/1/ns/mnt --net=/host/proc/1/ns/net mount -t ext4 -o
journal_checksum,journal_ioprio=0,barrier=1,errors=remount-ro,relatime /dev/sdc1 /var/lib/harvester/extra
disks/2bf5824103011662c7c505aa26dc372b], output , stderr, mount: /var/lib/harvester/extra
disks/2bf5824103011662c7c505aa26dc372b: /dev/sdc1 already mounted on /var/lib/harvester/extra
disks/2bf5824103011662c7c505aa26dc372b.
, error exit status 32
```

**To Reproduce**
 * **Harvester Create Node**
1. Prepare VM server and storage according to above `Disk usage information`
2. ISO install harvester
3. Send `ctrl+alt+F2` key change to console 
4. Login with rancher/rancher
5. Run `lsblk` to check all available disks
6. Run `gdisk /dev/sda or /sdb`
7. Press `W` to create GPT table
8. Create a harvester-config.yaml file add the following lines content 
  ```
  system_settings:
    auto-disk-provision-paths: /dev/sd*
  ```
9. put harvester-config on http server 
10. Send `ctrl+alt+F1` key back to iso installation
11. Input harvester config url (e.g http://192.168.1.106:8080/harvester-config.yaml) 
12. Wait for harvester installation complete
13. Access harvester and open Host page
14. Check all /dev/sda , /dev/sdb and /dev/sdc are provisioned and mounted correctly 

[...]

PR #1603, opened 2021-12-01
Bump node-disk-manager to 0.3.2

weihanglo:
**Problem:**
Disk auto-provisioning feature

**Solution:**
Bump harvester-node-disk-manager to 0.3.2

**Related Issue:**
https://github.com/harvester/harvester/issues/1599

**Test plan:**
See #1241


deploy/charts/harvester/Chart.lock
```diff
@@ -7,10 +7,10 @@ dependencies:
   version: 0.1.0
 - name: harvester-network-controller
   repository: https://charts.harvesterhci.io
-  version: 0.2.1
+  version: 0.2.4
```
guangbochen: can we please help to reset this value, it will be bumped in another PR, thanks.

weihanglo: Fixed. Thanks!

guangbochen (approved):
LGTM

Issue #1602, opened 2021-12-01
[BUG] exported image can't be deleted after vm removed


aiwantaozi:
**Describe the bug**

**To Reproduce**
1. create a VM with 2 volumes
```
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/volumeClaimTemplates: '[{"metadata":{"name":"test-vm-disk-0-9xocg","annotations":{"harvesterhci.io/imageId":"default/image-smjv2"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"volumeMode":"Block","storageClassName":"longhorn-image-smjv2"}},{"metadata":{"name":"test-vm-disk-1-pjaew"},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"volumeMode":"Block","storageClassName":"longhorn"}}]'
    kubevirt.io/latest-observed-api-version: v1
    kubevirt.io/storage-observed-api-version: v1alpha3
    network.harvesterhci.io/ips: '[]'
  labels:
    harvesterhci.io/creator: harvester
    harvesterhci.io/os: linux
  name: test-vm
  namespace: default
spec:
  running: true
  template:
    metadata:
      annotations:
        harvesterhci.io/sshNames: '[]'
      creationTimestamp: null
      labels:
        harvesterhci.io/vmName: test-vm
    spec:
      domain:
        cpu:
          cores: 1
          sockets: 1
          threads: 1
        devices:
          disks:
          - bootOrder: 1
            disk:
              bus: virtio
            name: disk-0
          - bootOrder: 2
            disk:
              bus: virtio
            name: disk-1
          - disk:
              bus: virtio
            name: cloudinitdisk
          inputs:
          - bus: usb
            name: tablet
            type: tablet
          interfaces:
          - macAddress: 52:54:00:0d:4c:79
            masquerade: {}
            model: virtio
            name: default
        machine:
          type: q35
        memory:
          guest: 924Mi
        resources:
          limits:
            cpu: "1"
            memory: 1Gi
          requests:
            cpu: 62m
            memory: 512Mi
      evictionStrategy: LiveMigrate
      hostname: test-vm
      networks:
      - name: default
        pod: {}
      volumes:
      - name: disk-0
        persistentVolumeClaim:
          claimName: test-vm-disk-0-9xocg
      - name: disk-1
[...]

PR #1610, opened 2021-12-02
Fix exported image can't be deleted after volume removed

futuretea:

**Problem:**

deletionTimestamp will be updated when vmImage is deleted, which triggers webhook of vmimage. Since PVC no longer exists, validate failed.

**Solution:**

1. if deletionTimestamp is not zero(deleting vmimage), ignore vmimage update validate
2. only check pvc exist when vm create
3. user cannot change spec.sourceType, spec.pvcNamespace, spec.pvcName

**Related Issue:**
#1602

**Test plan:**

case 1：
1. create vm "vm-1"
2. create a image "img-1" by export the volume used by vm "vm-1"
3. delete vm "vm-1"
4. delete image "img-1"
image "img-1" will be deleted

case 2：
1. create vm "vm-1"
2. create a image "img-1" by export the volume used by vm "vm-1"
3. delete vm "vm-1"
4. update image "img-1" labels
image "img-1" will be updated

### DIFF
--- a/pkg/webhook/resources/virtualmachineimage/validator.go
+++ b/pkg/webhook/resources/virtualmachineimage/validator.go
@@ -55,7 +55,14 @@ func (v *virtualMachineImageValidator) Resource() types.Resource {
 
 func (v *virtualMachineImageValidator) Create(request *types.Request, newObj runtime.Object) error {
 	newImage := newObj.(*v1beta1.VirtualMachineImage)
+	if err := v.CheckImageDisplayNameAndURL(newImage); err != nil {
+		return err
+	}
 
+	return v.CheckImagePVC(request, newImage)
+}
+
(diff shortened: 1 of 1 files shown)


guangbochen (approved):
LGTM

gitlawr:
Skipping validation on deletion may not cover all cases. Consider the following one:
1. Create a VMImage from a volume. Wait until it's imported.
2. Delete the volume.
3. Now updating the VMImage(e.g., set labels) results in the same error given that the source PVC in its spec is gone. This image is still valid and the operation should be supported.

I think we should skip the source PVC check after an image is imported.

futuretea:
Updated

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

Issue #1360, opened 2021-10-06
[BUG] VM's function won't work when host is down


lanfon72:
### Describe the bug
This issue contain several cases but point to the same thing.
When VMs running on a certain Node1, and Node1 had connection issues, those functions for VM still available, you can click them but nothing happened even no error messages.

** This is discovered when testing Negative test cases(4, 5, 14, 15) of *Backup and Restore*.

### To Reproduce
1. Install harvester with 3+ nodes
2. Create VMs on specific Node (for example, Node1)
3. reboot Node1, then Node1's state changed to `In Progress`
4. all VMs be hosted on Node1 still shows Running/Stopping/processing(backup)...etc

### Expected behavior
Update Node's state to `unavailable` or other state instead of `In Progress` when cluster can't receive its status. (This should be configure-able in Settings)
When entering the state, VMs function should be frozen too.

## Environment:
 - Harvester ISO version: **master-82506f55-head**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 3nodes**

### Additional context
[host's state]
![image](https://user-images.githubusercontent.com/5169694/136208805-3e6c0fc6-e64e-4302-91ee-c22e69d322ca.png)

[VM's status]
![image](https://user-images.githubusercontent.com/5169694/136208896-5e11e160-4be8-4f6e-9ad3-55b3160ae530.png)

[backup's status]
![image](https://user-images.githubusercontent.com/5169694/136209074-c1ca80df-cc78-41d4-8d62-fd958e73f87c.png)

lanfon72:
There also have some cascading issues as #1361's situation. (And we don't have any test cases for them yet)
so I will verify this when it is fixed and ready for testing. :wink:

yasker:
Since the UI code is frozen already, we need to release note this for v0.3.0 then punt it to v1.0.0.

lanfon72:
Ref exists issues: #982, #983 
All VM's negative test cases(1,2,3,4,5,7,8) is blocked due to this.

WuJun2016:
track issue:   https://github.com/harvester/harvester/issues/982

lanfon72:
@FrankYang0529, @WuJun2016, I will put this back to _review pipeline_ due to VM can't bring back after host is back.

- After host is back, VM will change to **Starting** state, but always hit _**FailedAttachVolume**_, and keep retrying.
[...]

PR #1625, opened 2021-12-03
fix(node-down-controller): remove creator label

FrankYang0529:
**Problem:**
node-down-controller uses labels to select pods, so it can't get correct pods to force delete.

**Solution:**
Remove the error label for the selector.

**Related Issue:**
https://github.com/harvester/harvester/issues/1360

**Test plan:**

Case 1: `vm-force-deletion-policy` has already been enabled.
* Set `vm-force-deletion-policy` as `'{"enable": true, "period": 60}'`.

Case 2: Enable `vm-force-deletion-policy` after a node is down.
* Disable vm-force-delete-policy.
* Set `vm-force-deletion-policy` as `'{"enable": true, "period": 60}'`.

### DIFF
--- a/pkg/controller/master/node/node_down_controller.go
+++ b/pkg/controller/master/node/node_down_controller.go
@@ -20,7 +20,6 @@ import (
 
 const (
 	nodeDownControllerName = "node-down-controller"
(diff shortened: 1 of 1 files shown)


guangbochen (approved):
LGTM

Issue #1272, opened 2021-09-24
[BUG] Shut down a node with maintenance mode should show red label


FrankYang0529:
**Describe the bug**
Maintenance mode label doesn't change from yellow to red when a node has been shut down.
![Screenshot from 2021-09-24 16-56-04](https://user-images.githubusercontent.com/4927361/134647555-d8a77116-880b-45f8-a3bb-fbfc1f89a436.png)

**To Reproduce**
1. Put a node in maintenance mode.
2. When a node shows Maintenance mode, then turn off the node.

**Expected behavior**
The node should be updated from yellow maintenance mode to red.

**Support bundle**

**Environment:**
 - Harvester ISO version: master-466a24f6-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

**Additional context**

TachunLin:
Verified **failed** on master-b5dadc82-head. Move issue to review stage.

### Result
Node keeps in yellow maintenance mode when node has been turned off

![image](https://user-images.githubusercontent.com/29251855/139856722-e5e7dc8e-da64-4a6f-a4e5-308a4300bbfb.png)

### Test Information
* Test Environment: 3 nodes harvester on local kvm
* Harvester version: master-b5dadc82-head

### Verify Steps

1. Open host page
2. Set a node to maintenance mode
3. Turn off host vm of the node
4. Check node status
 
### Additional Context
![image](https://user-images.githubusercontent.com/29251855/139859173-24004a98-df23-458b-8ce2-25eb4f271b16.png)

n313893254:
```
v1/nodes/qy
{
"error": true,
"lastHeartbeatTime": "2021-11-16T09:27:39Z",
"lastTransitionTime": "2021-11-16T09:29:16Z",
"lastUpdateTime": "2021-11-16T09:29:16Z",
"message": "Kubelet stopped posting node status.",
"reason": "NodeStatusUnknown",
"status": "Unknown",
"transitioning": false,
"type": "Ready"
}

=====
v1/harvester/nodes/qy
{
"error": false,
"lastHeartbeatTime": "2021-11-16T09:27:39Z",
"lastTransitionTime": "2021-11-16T09:29:16Z",
"lastUpdateTime": "2021-11-16T09:29:16Z",
"message": "Kubelet stopped posting node status.",
"reason": "NodeStatusUnknown",
"status": "Unknown",
"transitioning": false,
"type": "Ready"
}
],
```

[...]

PR #1631, opened 2021-12-06
Bump wrangler

gitlawr:
**Related Issue:**
https://github.com/harvester/harvester/issues/1272

**Problem:**
Incorrect error state when a node is not ready.

**Solution:**
This is fixed in the upstream: https://github.com/rancher/wrangler/pull/178

With this PR we still need UI changes to fix the related issue.

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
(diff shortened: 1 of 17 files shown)


guangbochen (approved):
LGTM

Issue #1022, opened 2021-07-15
Zero downtime upgrade


yasker:
The epic issue to track all upgrade-related issues.

## pre-1.0.0 dependencies
- [ ] https://github.com/longhorn/longhorn/issues/3234 (Enhancement included in `v1.2.3-rc1`)
- [x] https://github.com/harvester/harvester/issues/1521
- [x] https://github.com/rancher/rancher/issues/35464 (To eliminate double draining operation)
- [x] https://github.com/rancher-sandbox/cOS-toolkit/issues/928

yasker:
Unfortunately, we don't have enough time to finish the upgrade support in v0.3.0. Move to v1.0.0.

bk201:
For design doc and discussions, please head to https://hackmd.io/@bk201/H1rTAGaLt

bk201:
See this issue when upgrading a controller node today: https://github.com/harvester/harvester/issues/1521

bk201:
**Update**
Note I found it's not possible to test a master ISO due the the `rancher/harvester-upgrade` image for a dev build is not published.
Currently, we need a hack like https://github.com/bk201/harvester-installer/commit/040fb91f7aff0b77ecf3083f4c928acf9ff680f4 to include the image in a testing ISO.


**Test plan to upgrade from 1.0.0-rc2 to latest mater**

- Download master ISO and put it in an HTTP server or use the master ISO URL directly
- Get sha512 sum of the ISO.
- Create a version resource. This pretends there is a new upgrade. Adjust the `isoChecksum` and `isoURL` below. Keep the other fields.
```
cat > versions.yaml <<EOF
apiVersion: harvesterhci.io/v1beta1
kind: Version
metadata:
  name: 1.0.2
  namespace: harvester-system
spec:
  isoChecksum: 2765d89e5e8024347c81d36cc84b7168bf358a3a2f4da0c63164ad834a94bcbbaa13af4900aad4caa0a27550e4b4db5bf2c4726daa88664378e0c581a79ad2cb
  isoURL: http://10.10.0.1/harvester/harvester.iso
  minUpgradableVersion: 0.3.0
  releaseDate: "20211231"
  tags:
  - dev
  - test
EOF
```

```
kubectl create -f versions.yaml
```

- Refresh the GUI, there should be a new upgrade notification on the dashboard page (first page). Do an upgrade.
- How to verify
[...]

PR #1635, opened 2021-12-06
Support zero-downtime upgrade

bk201:
### Start an upgrade
When backend starts, it periodically gets available versions from a remote server and records upgradable ones with `Version` resources:
```
kind: Version
metadata:
  name: 1.0.2
  namespace: harvester-system
spec:
  isoChecksum: xxx
  isoURL: http://10.10.0.1/harvester/harvester.iso
  minUpgradableVersion: 0.3.0
  releaseDate: "20211231"
  tags:
  - dev
  - test
```

To start an upgrade, we can create an `Upgrade` resource:
```
kind: Upgrade
metadata:
  generateName: hvst-upgrade-
  namespace: harvester-system
spec:
  version: "1.0.2"
#  image: harvester-system/image-gsjrt
```

The `version` field has to be one of the `Version` resource names. The upgrade controller then uses the ISOURL record in the version resource to start an upgrade.

For air-gapped scenarios, the user can upload an ISO to the system as a new `VirtualMachineImage`, and specify the image ID in the `image` field to start an upgrade.

### Upgrade flow

The overall flow looks like the following figure:
upgrade-Overview - integrated drawio

- **Phase1**: A VM is created from a release ISO. We call it the upgrade repo VM. The VM serves ISO contents with an Nginx server.
- **Phase2**: The upgrade controller creates a SUC plan to run jobs on each node. The job preload container images from the upgrade repo VM. The jobs use image `rancher/harvester-upgrade:<current-version>`.
- **Phase3**: A job upgrade Rancher server and all Harvester ManagedCharts. The job uses image `rancher/harvester-upgrade:<new-version>`. Because all new images are already preloaded, we can safely use the new image. The benefit is we can add new fixes here.
- **Phase4**: Upgrade the underlying RKE2 cluster. When rancher upgrades RKE2 runtime, it supports[ pre-drain and post-drain hooks](https://github.com/rancher/rancher/issues/35464). We integrate the nodes upgrade into the process:
     - Pre-drain: VMs on a node are live-migrated out. Non-migratable VMs will be shut down. 
     - Drain a node
     - Post-drain: Patch the OS image and reboot.
     - After the node is back, continue to the next node.
- **Phase5**: Clean up. The repo VM is deleted after the upgrade is deleted.

Note for single-node cases, **Phase 4** works differently. Because there is only one node, thus we can't rely on the pre-drain/post-drain mechanism. The controller creates a job to do the node upgrade:
- Download needed file from repo VM.
- Shutdown All VMs.
- Upgrade RKE2.
- Patch OS and reboot.

### CRD change

A new field is added to VirtualMachineImage CRD:
```
type VirtualMachineImageSpec struct {

<...>
	// +optional
	Checksum string `json:"checksum"`
```

The field can contain an optional checksum value (in SHA-512) to be passed to Longhorn backing image service.

### More to enhance
- [ ] Images Clean-up on hosts
- [ ] Add more tests

**Related Issue:**
https://github.com/harvester/harvester/issues/1022

**Test plan (for reviewer):**

- Run a cluster with master build
- Pause `harvester-crd` ManagedChart. Otherwise, CRDs will be reverted back by Helm.
   ```
   # kubectl edit managedchart harvester-crd -n fleet-local

   spec:
      paused: true
   ```
- Delete some changed CRDs
   ```
   kubectl delete crd upgrades.harvesterhci.io
   kubectl delete crd virtualmachineimages.harvesterhci.io 
   ```
- Build a harvester image with this PR change and patch `harvester` ManagedChart to use this image. Both backend and webhook.

   e.g. (Please change the tag to fit your environment)
   ```
    $ cat patch-managedchart.json
    {
      "spec": {
        "values": {
          "replicas": 1,
          "containers": {
            "apiserver": {
              "image": {
                "imagePullPolicy": "Always",
                "repository": "rancher/harvester",
                "tag": "kiefer-dev"
              },
              "debug": true
            }
          },
          "webhook": {
            "replicas": 1,
            "image": {
              "imagePullPolicy": "Always",
              "repository": "rancher/harvester-webhook",
              "tag": "kiefer-dev"
            }
          }
        }
      }
    }

    $ cat patch-managedchart.sh
    #!/bin/bash
    kubectl patch managedcharts.management.cattle.io harvester --namespace fleet-local --patch-file patch-managedchart.json  --type merge   
   ```
- Build upgrade images and push them to somewhere for use.
   ```
   make package-upgrade
   ```

   Push the image. Note this part is tricky. During the upgrade, two versions of this image are used.
   - `harvester-upgrade:<current_version>`:  `kubectl get settings.harvesterhci.io  server-version`
   - `harvester-upgrade:<next_version>`: see next step
- Build an ISO from harvester-installer master code. Create a dump commit to make version changes. Store the ISO in a HTTP server. The ISO contains a file `harvester-release.yaml`, get the `harvester: ` version from there to use as the `<next_version>` tag in the previous step.
- Create versions CR, this pretends there are new upgrades:
   ```
   $ cat version.yaml
   kind: Version
   metadata:
     name: 1.0.2
     namespace: harvester-system
   spec:
     isoChecksum: d8f5c1fb41f8a6b7b8d2913223cc1da492279f1194b63d628046debf75eb647849b4fbf485725e6c4f4435fd70d6dfa9a9c0bf4c1e7611df56b37f0819aedd9a                               <-- the build ISO checksum
     isoURL: http://10.10.0.1/harvester/harvester.iso      <-- the build ISO
     minUpgradableVersion: 0.3.0
     releaseDate: "20211231"
     tags:
     - dev                 <--- `dev` tag is important
     - test

    $ kubectl create -f version.yaml
     ```
- Create an upgrade:
   ```
   $ cat upgrade.yaml
   kind: Upgrade
   metadata:
     generateName: hvst-upgrade-
     namespace: harvester-system
   spec:
     version: "1.0.2"

   $ kubectl create -f upgrade.yaml
   ```


vendor/sigs.k8s.io/cluster-api/api/v1alpha4/groupversion_info.go
```diff
@@ -33,4 +33,7 @@ var (
 
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
+
+	// for generate codes
+	SchemeGroupVersion = schema.GroupVersion{Group: "cluster.x-k8s.io", Version: "v1alpha4"}
```
gitlawr: Are we patching the vendor file here? This workaround looks fragile as it can easily get overwritten without awareness.

pkg/controller/master/upgrade/image_controller.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+)
+
+type vmImageHandler struct {
```
gitlawr, line 14: Can we add some description for each controller so we can easily tell the purpose of it

pkg/controller/master/upgrade/common.go
```diff
@@ -454,6 +667,16 @@ func (n *nodeBuilder) WithLabel(key, value string) *nodeBuilder {
...
 }
 
-func (n *nodeBuilder) Build() *v1.Node {
+func (n *nodeBuilder) Build() *corev1.Node {
 	return n.node
 }
+
+func upgradeReference(upgrade *harvesterv1.Upgrade) metav1.OwnerReference {
+	return metav1.OwnerReference{
+
```
gitlawr: nit: remove the empty line

pkg/controller/master/upgrade/image_controller.go
```diff
@@ -0,0 +1,47 @@
...
+
+	upgrade, err := h.upgradeCache.Get(upgradeNamespace, image.Labels[harvesterUpgradeLabel])
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return image, nil
+		}
+		return nil, err
+	}
+
+	if harvesterv1.ImageImported.GetStatus(image) == string(corev1.ConditionTrue) {
```
gitlawr: Can we use `harvesterv1.ImageImported.IsTrue(image)`?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -1,95 +1,275 @@
...
 	harvesterLatestUpgradeLabel    = "harvesterhci.io/latestUpgrade"
 	harvesterUpgradeComponentLabel = "harvesterhci.io/upgradeComponent"
-	upgradeImageRepository         = "rancher/harvester-bundle"
+	harvesterNodeLabel             = "harvesterhci.io/node"
+	upgradeImageRepository         = "rancher/harvester-upgrade"
+
+	harvesterNodePendingOSImage     = "harvesterhci.io/pendingOSImage"
+	harvesterUpgradeImageAnnotation = "harvesterhci.io/upgradeImage"
+	repoInfoAnnotation              = "harvesterhci.io/repoInfo"
+	singleNodeAnnotation            = "harvesterhci.io/singleNode"
```
gitlawr: Is it better to set them in upgrade status other than in upgrade annotations

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -1,95 +1,275 @@
...
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	kubevirtctrl "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
 	upgradectlv1 "github.com/harvester/harvester/pkg/generated/controllers/upgrade.cattle.io/v1"
 	"github.com/harvester/harvester/pkg/settings"
 )
 
+var (
+	upgradeControllerLock sync.Mutex
```
gitlawr, line 27: Can you explain the reason for the lock

pkg/controller/master/upgrade/machine_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	toUpdate := upgrade.DeepCopy()
+	setNodeUpgradeStatus(toUpdate, nodeName, nextState, "", "")
+	if _, err := h.upgradeClient.Update(toUpdate); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *machineHandler) checkPendingHookJobs(upgrade string) error {
```
gitlawr, line 110: Can you explain this check?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -1,95 +1,275 @@
...
+				return h.upgradeClient.Update(toUpdate)
+			}
+			if toUpdate.Annotations == nil {
+				toUpdate.Annotations = make(map[string]string)
+			}
+			toUpdate.Annotations[harvesterUpgradeImageAnnotation] = upgrade.Spec.Image
+			setImageReadyCondition(toUpdate, corev1.ConditionTrue, "", "")
+		}
+		harvesterv1.ImageReady.CreateUnknownIfNotExists(toUpdate)
+		// setRepoProvisionedCondition(toUpdate, corev1.ConditionTrue, "", "")
```
gitlawr: Remove unused codes

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -114,3 +294,77 @@ func (h *upgradeHandler) resetLatestUpgradeLabel(latestUpgradeName string) error
...
+	}
+
+	toUpdate := cluster.DeepCopy()
+	toUpdate.Spec.KubernetesVersion = kubernetesVersion
+
+	if toUpdate.Spec.RKEConfig == nil {
+		toUpdate.Spec.RKEConfig = &provisioningv1.RKEConfig{}
+	}
+
+	drain := true
```
gitlawr: The variable name looks confusing. It seems to be a constant

vendor/sigs.k8s.io/cluster-api/api/v1alpha4/groupversion_info.go
bk201: It'll definitely break when updating the vendor again. Any suggestion to fix this? Wrangler uses `SchemeGroupVersion` which is not defined in the vendor package.

pkg/controller/master/upgrade/upgrade_controller.go
bk201: `Status` is read and set between controllers. Not sure if the value might change when entering a controller's handler function.

pkg/controller/master/upgrade/upgrade_controller.go
bk201: Fixed in https://github.com/harvester/harvester/pull/1635/commits/c79c94c3c27d33ee40e2139fde2c76fb041090fe
I can't use a constant because I need to [pass its pointer](https://github.com/harvester/harvester/pull/1635/commits/c79c94c3c27d33ee40e2139fde2c76fb041090fe#diff-ad1688890111ebb67c1e9830de55577e7b83f1828d79f6f1f9609122cac236d3R308).

pkg/controller/master/upgrade/upgrade_controller.go
bk201: Fixed. Thanks.

pkg/controller/master/upgrade/machine_controller.go
bk201: it's a safety check to allow only one pre-drain/post-drain job to exist.

pkg/controller/master/upgrade/upgrade_controller.go
bk201: https://github.com/harvester/harvester/pull/1635/commits/2adcd6ba0ddf8041f42a4b0cd191021fb7881808
Fixed. I moved all annotation hacks into status.

pkg/controller/master/upgrade/image_controller.go
bk201: Fixed in https://github.com/harvester/harvester/pull/1635/commits/02f43bdf63e555b32abe7984a5680cef19f44bcc#diff-15276b9422a58603cacc38542d540cf52d04053a27b5e9dadd702b49d1cf9435R37-R52
I also include the status report if image importing fails.

pkg/controller/master/upgrade/image_controller.go
bk201: Added.

vendor/sigs.k8s.io/cluster-api/api/v1alpha4/groupversion_info.go
gitlawr: I can go generate, build and run without this change so I think we change drop this commit

weihanglo: @bk201, what's the problem you met? In runtime or compile time? I didn't encounter any compile error so far.

bk201: My IDE detects an error. But since that path is not called, so it builds in our case. I'll remove the commit.

pkg/controller/master/upgrade/common.go
```diff
@@ -14,14 +14,13 @@ import (
...
-	agentComponent    = "agent"
-	serverComponent   = "server"
+	nodeComponent     = "node"
 	manifestComponent = "manifest"
 
 	labelArch               = "kubernetes.io/arch"
 	labelCriticalAddonsOnly = "CriticalAddonsOnly"
 
-	defaultTTLSecondsAfterFinished = 900
+	defaultTTLSecondsAfterFinished = 1200
```
weihanglo, line 23: Is it an empirical value from your experiments?

weihanglo:
Sorry. Not that related two the PR itself. Just need to clarify the upgrade dependency chain.

IIUC, the upgrade dependency chain is: Rancher -> Harvester -> RKE2 -> OS. That is, if some changes need two or more components being upgraded simultaneously, we'll need maintain two version on downstream components, right? 

Like, if we need an RKE2 patch which also requires a new version of Harvester, the downstream Harvester need to compatible with old and new RKE2 since it upgrades first.

Moreover, do we support upgrade from any version. For example, if we release 1.0, 1.1, 2.0, can user upgrade to 2.0 from 1.0? Or user can only upgrade one version up in an upgrade without skipping any version?

bk201:
This is definitely related to this PR.
Generally, we can update to the latest version. I think things we might need to tackle are doing special fixes for certain previous versions.

> Sorry. Not that related two the PR itself. Just need to clarify the upgrade dependency chain.
> 
> IIUC, the upgrade dependency chain is: Rancher -> Harvester -> RKE2 -> OS. That is, if some changes need two or more components being upgraded simultaneously, we'll need maintain two version on downstream components, right?
> 
> Like, if we need an RKE2 patch which also requires a new version of Harvester, the downstream Harvester need to compatible with old and new RKE2 since it upgrades first.
> 
> Moreover, do we support upgrade from any version. For example, if we release 1.0, 1.1, 2.0, can user upgrade to 2.0 from 1.0? Or user can only upgrade one version up in an upgrade without skipping any version?

vendor/sigs.k8s.io/cluster-api/api/v1alpha4/groupversion_info.go
bk201: Removing the workaround fails the tests.
```
# make test
<..>
Running unit tests
# github.com/harvester/harvester/pkg/generated/clientset/versioned/typed/cluster.x-k8s.io/v1alpha4
pkg/generated/clientset/versioned/typed/cluster.x-k8s.io/v1alpha4/cluster.x-k8s.io_client.go:70:8: undefined: v1alpha4.SchemeGroupVersion
?       github.com/harvester/harvester/pkg/api  [no test files]
?       github.com/harvester/harvester/pkg/api/image    [no test files]
?       github.com/harvester/harvester/pkg/api/keypair  [no test files]
?       github.com/harvester/harvester/pkg/api/kubeconfig       [no test files]
?       github.com/harvester/harvester/pkg/api/node     [no test files]
?       github.com/harvester/harvester/pkg/api/proxy    [no test files]
?       github.com/harvester/harvester/pkg/api/supportbundle    [no test files]
FAIL    github.com/harvester/harvester/pkg/api/vm [build failed]
?       github.com/harvester/harvester/pkg/api/vmtemplate       [no test files]
?       github.com/harvester/harvester/pkg/api/volume   [no test files]
?       github.com/harvester/harvester/pkg/builder      [no test files]
?       github.com/harvester/harvester/pkg/cmd  [no test files]
?       github.com/harvester/harvester/pkg/codegen      [no test files]
?       github.com/harvester/harvester/pkg/codegen/cleanup      [no test files]
?       github.com/harvester/harvester/pkg/config       [no test files]
?       github.com/harvester/harvester/pkg/controller/admission [no test files]
```

pkg/util/fakeclients/upgrade.go
```diff
@@ -51,8 +51,8 @@ func (c UpgradeCache) List(namespace string, selector labels.Selector) ([]*harve
 		return nil, err
 	}
 	result := make([]*harvesterv1.Upgrade, 0, len(list.Items))
-	for _, upgrade := range list.Items {
-		result = append(result, &upgrade)
+	for i := range list.Items {
+		result = append(result, &list.Items[i])
```
weihanglo, line 55: Just curious why this change? Anything gets weird with `range`?

pkg/controller/master/upgrade/job_controller.go
```diff
@@ -1,26 +1,46 @@
...
-	harvesterChartname = "harvester"
+	StateUpgrading               = "Upgrading"
+	StateCreatingUpgradeImage    = "CreatingUpgradeImage"
+	StatePreparingRepo           = "PreparingRepo"
+	StateRepoPrepared            = "RepoPrepared"
+	StatePreparingNodes          = "PreparingNodes"
+	StateUpgradingSystemServices = "UpgradingSystemServices"
+	StateUpgradingNodes          = "UpgradingNodes"
+	StateSucceeded               = "Succeeded"
+	StateFailed                  = "Failed"
```
weihanglo, line 30: NIT: Prefer to call it Phase instead. They are series of event of a process, silmiar to PodPhase

vendor/sigs.k8s.io/cluster-api/api/v1alpha4/groupversion_info.go
bk201: Use the same workaround for patching nad: https://github.com/harvester/harvester/pull/1635/commits/af95fff41f23b82eec4d4d8376d95b054089e080
This is not perfect, but it should sustain future go generate.

bk201:
Add more changes
- d41a92c: Image can be in importing state, this makes GUI's life easier to support the air-gapped scenario.
- e5c7d4e: `version` is optional. We don't need a version in the air-gapped scenario.
- cffbd1e: add drone job to publish `harvester-upgrade` images.

cc @guangbochen Do we need a new ticket to push to `rancher/harvester-upgrade` ?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -1,95 +1,277 @@
...
+		logrus.Info("Starting upgrade repo VM")
 		toUpdate := upgrade.DeepCopy()
-		if _, err := h.planClient.Create(serverPlan(upgrade, disableEviction)); err != nil && !apierrors.IsAlreadyExists(err) {
-			setNodesUpgradedCondition(toUpdate, corev1.ConditionFalse, "", err.Error())
+		if err := repo.Bootstrap(); err != nil && !apierrors.IsAlreadyExists(err) {
+			setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, err.Error(), "")
 			return h.upgradeClient.Update(toUpdate)
 		}
-		initStatus(toUpdate)
+		toUpdate.Labels[upgradeStateLabel] = StatePreparingRepo
```
weihanglo, line 138: I wonder do we query Update CR by this `upgradeStateLabel`. If not, I think it might be nice to put it in `Status`. (Yes, similar to `pod.status.phase`)

package/upgrade/upgrade_manifests.sh
```diff
@@ -0,0 +1,109 @@
...
+{
+  mkdir -p $UPGRADE_TMP_DIR/harvester
+  cd $UPGRADE_TMP_DIR/harvester
+
+  cat > harvester.yaml <<EOF
+spec:
+  version: $REPO_HARVESTER_CHART_VERSION
+EOF
+  kubectl patch managedcharts.management.cattle.io harvester -n fleet-local --patch-file ./harvester.yaml --type merge
+  # TODO: Is there a way to wait for Harvester and all components are fully upgraded (and maybe operational)?
```
weihanglo, line 90: Maybe we can do the same trick as installer showing "Ready" in TUI?

package/upgrade/upgrade_manifests.sh
```diff
@@ -0,0 +1,109 @@
...
+upgrade_rancher()
+{
+  mkdir -p $UPGRADE_TMP_DIR/images
+  mkdir -p $UPGRADE_TMP_DIR/rancher
+
+  # Download rancher system agent install image from upgrade repo
+  download_image_archives_from_repo "agent" $UPGRADE_TMP_DIR/images
+
+  # Extract the Rancher chart and helm binary
+  wharfie --images-dir $UPGRADE_TMP_DIR/images rancher/system-agent-installer-rancher:$REPO_RANCHER_VERSION $UPGRADE_TMP_DIR/rancher
```
weihanglo, line 22: First time saw [wharfie]. Really convenient!

[wharfie]: https://github.com/rancher/wharfie

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -1,95 +1,277 @@
...
+		for _, job := range jobs {
+			logrus.Debugf("Deleting job %s/%s", job.Namespace, job.Name)
+			if err := h.jobClient.Delete(job.Namespace, job.Name, &metav1.DeleteOptions{}); err != nil {
+				return nil, err
+			}
+		}
+
+		logrus.Debugf("Deleting plan %s/%s", plan.Namespace, plan.Name)
+		if err := h.planClient.Delete(plan.Namespace, plan.Name, &metav1.DeleteOptions{}); err != nil {
+			return nil, err
```
weihanglo, line 250: Shouldn't plans and jobs get garbage-collected once their parent Upgrade was deleted?

pkg/controller/master/upgrade/upgrade_repo.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	image, err := r.h.vmImageCache.Get(tokens[0], tokens[1])
+	if err != nil {
+		return nil, err
+	}
+	return image, nil
+}
+
+func (r *UpgradeRepo) createVM(image *harvesterv1.VirtualMachineImage) (*kv1.VirtualMachine, error) {
```
weihanglo, line 139: Generally this function looks great! I am thinking is it possible to reuse `vmbuilder` under `pkg/builder` package? Maybe it is unwise to reuse that, I am not sure. But I know that node provisioning in Rancher uses vmbuilder.

package/upgrade/upgrade_node.sh
```diff
@@ -0,0 +1,322 @@
...
+  mount $tmp_rootfs_squashfs $tmp_rootfs_mount
+
+  bash -x $HOST_DIR/usr/sbin/cos upgrade --directory $tmp_rootfs_mount
+  umount $tmp_rootfs_mount
+  rm -rf $tmp_rootfs_squashfs
+
+  umount -R /run
+
+  # https://github.com/rancher-sandbox/cOS-toolkit/issues/928
+  rebrand_grub || true
```
weihanglo, line 258: I guess we can test the fix now.
https://github.com/rancher-sandbox/cOS-toolkit/pull/959

package/upgrade/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM registry.suse.com/suse/sle15:15.3
+
+ARG ARCH=amd64
+ENV ARCH=${ARCH}
+RUN zypper rm -y container-suseconnect && \
+    zypper ar --priority=200 http://download.opensuse.org/distribution/leap/15.3/repo/oss repo-oss && \
+    zypper --no-gpg-checks ref && \
+    zypper in -y curl e2fsprogs rsync awk zstd jq && zypper clean
+
+ENV KUBECTL_VERSION v1.20.4
```
weihanglo: Is there any reason choosing 1.20.4 over newer version?

pkg/controller/master/upgrade/common.go
```diff
@@ -71,173 +108,296 @@ func markComplete(upgrade *harvesterv1.Upgrade) {
...
+							Operator: corev1.TolerationOpEqual,
+							Effect:   corev1.TaintEffectNoSchedule,
+							Value:    "arm64",
+						},
+						{
+							Key:      labelArch,
+							Operator: corev1.TolerationOpEqual,
+							Effect:   corev1.TaintEffectNoSchedule,
+							Value:    "arm",
+						},
```
weihanglo: Found this part duplicate over and over again. Can we extract it?

pkg/controller/master/upgrade/job_controller.go
```diff
@@ -29,27 +49,141 @@ type jobHandler struct {
...
+	if !ok {
+		return job, nil
+	}
+
+	node, err := h.nodeCache.Get(nodeName)
+	if err != nil {
+		return job, nil
+	}
+
+	machineName, ok := node.Annotations["cluster.x-k8s.io/machine"]
```
weihanglo: I guess this should also extract as a const since both node and job controllers use it.

weihanglo (commented):
Looks good to me overall. Just need some time to test it manually 😆

pkg/util/fakeclients/upgrade.go
bk201:
```
	for _, upgrade := range list.Items {
		result = append(result, &upgrade)
```
The pointer to `upgrade` variable is appended to `result` in each iteration. So in the end the result contains all identical values. I often make this mistake. https://stackoverflow.com/a/48826629 has some info.

weihanglo: That's too bad with this kind of pitfall. Anyway thanks for the info!

pkg/codegen/main.go
```diff
@@ -121,3 +122,19 @@ func nadControllerInterfaceRefactor() {
...
+func capiWorkaround() {
+	absPath, _ := filepath.Abs("pkg/generated/clientset/versioned/typed/cluster.x-k8s.io/v1alpha4/cluster.x-k8s.io_client.go")
+	input, err := ioutil.ReadFile(absPath)
+	if err != nil {
+		logrus.Fatalf("failed to read the clusters.cluster.x-k8s.io client file: %v", err)
+	}
+	output := bytes.Replace(input, []byte("v1alpha4.SchemeGroupVersion"), []byte("v1alpha4.GroupVersion"), -1)
+
+	if err = ioutil.WriteFile(absPath, output, 0644); err != nil {
+		logrus.Fatalf("failed to update the network-attachment-definition file: %v", err)
```
gitlawr: Fix the log

bk201: Nice catch. Fixed. 4c9ac5e

pkg/controller/master/upgrade/common.go
bk201: Thanks. Fixed in 1b5ea72

pkg/controller/master/upgrade/job_controller.go
bk201: Thanks. Fixed in 1b5ea72

package/upgrade/Dockerfile
bk201: Fixed. Use the same version of the current RKE2.

pkg/controller/master/upgrade/upgrade_repo.go
bk201: I'll keep what it is, the image builder looks handy but I'll need to add some new codes to it and make sure existing integration tests work.

pkg/controller/master/upgrade/common.go
bk201: Considering a physical server needs several minutes to reboot, I increase the wait time to 20 minutes.
We can always fine-tune this value.

bk201:
New changes:
- bb30ebf: check new fields in case Upgrade responder doesn't contain the values.
- fde1e85: add upgrade images in chart values file. This allow harvester-installer to pull in upgrade images.

package/upgrade/upgrade_node.sh
bk201: Waiting on the new packages to be published.

package/upgrade/upgrade_node.sh
```diff
@@ -0,0 +1,322 @@
...
+  mount -o remount,rw $HOST_DIR/run/initramfs/cos-state/
+  chroot $HOST_DIR grub2-editenv /run/initramfs/cos-state/grub_oem_env set "default_menu_entry=$REPO_OS_PRETTY_NAME"
+  mount -o remount,ro $HOST_DIR/host/run/initramfs/cos-state/
+}
+
+upgrade_os() {
+  CURRENT_OS_VERSION=$(source $HOST_DIR/etc/os-release && echo $PRETTY_NAME)
+
+  if [ "$REPO_OS_PRETTY_NAME" = "$CURRENT_OS_VERSION" ]; then
+    echo "Skip upgrading OS. The version version is already \"$CURRENT_OS_VERSION\"."
```
johnliu55tw: Seems like a typo? Not sure if this is the correct word:
```suggestion
    echo "Skip upgrading OS. The Harvester version is already \"$CURRENT_OS_VERSION\"."
```

guangbochen:
> cc @guangbochen Do we need a new ticket to push to `rancher/harvester-upgrade` ?

@bk201 yes, and I saw the EIO ticket, but IIUC the `harvester-upgrade` image is not required as we have already pre-load the images, but if we want to release a hot-fix version then it will be tied to the Harvester release due to the drone CI config, is this expected?

bk201:
> IIUC

We need it because the images need to be packed in harvester-installer.

package/upgrade/upgrade_node.sh
bk201: Thanks, fixed.

bk201:
Need to rebase and adapt the CRD location changes.

bk201:
- Rebased and moved CRDs to harvester-chart
- Support upgrade `harvester-crd` chart: eb01246

deploy/charts/harvester/values.yaml
```diff
@@ -413,6 +413,11 @@ webhook:
 
   httpsPort: 9443
 
+upgrade:
+  image:
+    repository: rancher/harvester-upgrade
+    tag: master-head
```
guangbochen, line 419: I think the ISO build would fail if the image does not exist, seem we can only merge this PR after the repo is prepared.

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM, @bk201 can we squash some of the commit, thanks

pkg/controller/master/upgrade/common.go
bk201: FWIW, it takes ~12 minutes to run a post-drain job in QA's environment. 
The job upgrades RKE2 and the OS image.
4VMs on a host, CPU: E5-2680 v3 + SSD.

Issue #1637, opened 2021-12-06
[BUG] When you change the memory on a VM it doesn't change guest memory


noahgildersleeve:
**Describe the bug**

When you edit the memory on a VM it doesn't affect the actual guest memory
**To Reproduce**
1. Create a VM with 2GB of RAM
2. Start the VM
3. Check the memory available with `top`
4. Change the memory via Harvester UI to 4GB
5. Wait for VM to restart
6. Check new available memory with `top`
**Expected behavior**

The memory should now have 4GB available to the VM
**Support bundle**

[supportbundle_57ddef0a-5e9c-4a55-9f91-96f8494dab45_2021-12-06T21-33-05Z.zip](https://github.com/harvester/harvester/files/7663677/supportbundle_57ddef0a-5e9c-4a55-9f91-96f8494dab45_2021-12-06T21-33-05Z.zip)

**Environment:**
 - Harvester ISO version: master-140adb56-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): DL160 4 node

**Additional context**
This is happening on the nightly build and RC1. It seems that the issue is that the limits for the memory are being changed in the YAML, but the guest memory is not. I think this might be a result of the changes we made to allow for CPU overcommit. If you manually change the guest memory in the YAML it will update the available memory.
![Greenshot 2021-12-06 13 33 06](https://user-images.githubusercontent.com/83787952/144926828-9b9093b0-644c-4f55-8826-e00bfbd4f2be.png)
![Greenshot 2021-12-06 13 34 13](https://user-images.githubusercontent.com/83787952/144926832-390b04e8-4756-42d4-9d89-7a0562da1b5f.png)
![Greenshot 2021-12-06 13 38 46](https://user-images.githubusercontent.com/83787952/144926835-31e568dd-2914-4d7b-8b69-e91c7dbeda1d.png)

weihanglo:
I found that the PUT payload contains the old values in `domain.memory.guest` field. It might prevent KubeVirt from re-calculating a new guest memory.

Below are my changing on a VM with 1GB memory to 3GB, and the change didn't reflect on guest memory in the payload.

<img src="https://user-images.githubusercontent.com/14314532/144943936-22e95483-89a6-47a4-84f5-0fd3759f050c.png" width="600px">

@WuJun2016 Can you help check this issue on dashboard side? If it is the case, we might need to leave `domain.memory.guest` blank when update VM as well as virtualmachinetemplateversion resources.

[...]

PR #1638, opened 2021-12-07
Patch existing guest memory on all occasions

weihanglo:
**Problem:**
Guest memory does not reflect the memory change in either UI dashboard or API

**Solution:**
Should not check `memory.guest` existence if we always want to patch the guest memory.

**Related Issue:**
#1637

**Test plan:**
See #1637


FrankYang0529 (approved):
LGTM

guangbochen (approved):
LGTM

PR #1639, opened 2021-12-07
feat(webhook): check vmbackup status in backup target validation

FrankYang0529:
**Problem:**
Enhancement: Stop users to change backup target if there is VMBackup in creating or deleting progress.

**Solution:**
Check VMBackup status in backup target validation.

**Related Issue:**
https://github.com/harvester/harvester/issues/988#issuecomment-987509872

**Test plan:**

* Setup Backup Target.
* Create a VM.
* Create a VMBackup.
* Try to change Backup Target when VMBackup is not ready to use.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -17,6 +17,8 @@ import (
 	"github.com/rancher/wrangler/pkg/slice"
 	"golang.org/x/net/http/httpproxy"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
@@ -49,9 +51,10 @@ var validateSettingFuncs = map[string]validateSettingFunc{
 	settings.SSLParametersName:                validateSSLParameters,
 }
 
-func NewValidator(settingCache ctlv1beta1.SettingCache) types.Validator {
+func NewValidator(settingCache ctlv1beta1.SettingCache, vmBackupCache ctlv1beta1.VirtualMachineBackupCache) types.Validator {
 	validator := &settingValidator{
-		settingCache: settingCache,
+		settingCache:  settingCache,
+		vmBackupCache: vmBackupCache,
 	}
 	validateSettingFuncs[settings.BackupTargetSettingName] = validator.validateBackupTarget
 	return validator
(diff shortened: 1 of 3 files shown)


guangbochen (approved):
LGTM

pkg/webhook/resources/setting/validator.go
```diff
@@ -335,3 +347,12 @@ func getSystemCerts() *x509.CertPool {
...
 	return certs
 }
+
+func hasVMBackupInCreatingOrDeletingProgress(vmBackups []*v1beta1.VirtualMachineBackup) bool {
+	for _, vmBackup := range vmBackups {
+		if vmBackup.DeletionTimestamp != nil || vmBackup.Status == nil || !*vmBackup.Status.ReadyToUse {
+			return true
+		}
+	}
+	return false
```
w13915984028, line 357: Question:
when will   `vmBackup.Status == nil` ?

`if vmBackup.DeletionTimestamp != nil || vmBackup.Status == nil || !*vmBackup.Status.ReadyToUse`
or
`if vmBackup.DeletionTimestamp != nil || (vmBackup.Status != nil && !*vmBackup.Status.ReadyToUse)`

FrankYang0529: There is no `status` in VM Backup when it's just created. If a VM Backup doesn't have `status`, it means it is in creating process.
https://github.com/harvester/harvester/blob/655c251ada5808399c8b9c475abc75e437a51d09/pkg/controller/master/backup/backup.go#L133-L149

pkg/webhook/resources/setting/validator.go
```diff
@@ -162,6 +166,14 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
 		return werror.NewInvalidError("NFS backup target should not have bucket name or region ", "value")
 	}
 
+	vmBackups, err := v.vmBackupCache.List(metav1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+	if yes := hasVMBackupInCreatingOrDeletingProgress(vmBackups); yes {
```
weihanglo:
```suggestion
	if hasVMBackupInCreatingOrDeletingProgress(vmBackups) {
```

pkg/webhook/resources/setting/validator.go
```diff
@@ -162,6 +166,14 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
 		return werror.NewInvalidError("NFS backup target should not have bucket name or region ", "value")
 	}
 
+	vmBackups, err := v.vmBackupCache.List(metav1.NamespaceAll, labels.Everything())
```
weihanglo, line 169: Is it possible to query specific namespaces only?

weihanglo (approved):
Tested. Overall LGTM

image

pkg/webhook/resources/setting/validator.go
FrankYang0529: Updated it.

pkg/webhook/resources/setting/validator.go
FrankYang0529: Probably no. VMBackup depends on VM in which namespace, so it's better to check all namespace.

gitlawr (approved):
LGTM

Issue #1588, opened 2021-11-26
[BUG] VM backup cause harvester pod to crash


guangbochen:
**Describe the bug**
VM backup cause harvester pod to crash after upgrade

**To Reproduce**
1. create a VM backup with v0.3.0 cluster
1. upgrade harvester from v0.3.0 to the master-head version
1. the harvester pod keeps crash

**Additional context**
```
time="2021-11-26T02:32:10Z" level=error msg="error syncing 'test/yz-template-test-8v6jx': handler template-version-controller: virtualmachinetemplates.harvesterhci.io \"yz-template-test\" not found, requeuing"
E1126 02:32:10.964692       7 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1018 [running]:
github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime.logPanic(0x2147040, 0x3b3a920)
	/go/src/github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0x95
github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/src/github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x86
panic(0x2147040, 0x3b3a920)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/harvester/harvester/pkg/controller/master/backup.IsBackupTargetSame(...)
	/go/src/github.com/harvester/harvester/pkg/controller/master/backup/util.go:31
github.com/harvester/harvester/pkg/controller/master/backup.(*Handler).uploadVMBackupMetadata(0xc00396c900, 0xc0000ec580, 0xc0005c5200, 0x0, 0x0)
	/go/src/github.com/harvester/harvester/pkg/controller/master/backup/backup.go:564 +0x6c
github.com/harvester/harvester/pkg/controller/master/backup.(*Handler).OnBackupChange(0xc00396c900, 0xc000a82130, 0xb, 0xc0000ec580, 0xd50be1e5edfc61b6, 0x40409a, 0x403bcb)
	/go/src/github.com/harvester/harvester/pkg/controller/master/backup/backup.go:120 +0x39f
[...]

PR #1487, opened 2021-10-28
 Support cloud config secrets for templates and backup/restores

gitlawr:
Generate dedicated cloud-init secrets for each template version. Store related secrets in vm backups.

**Related Issue:**
https://github.com/harvester/harvester/issues/760

PR #1642, opened 2021-12-08
fix(backup): nil BackupTarget pointer

FrankYang0529:
**Problem:**
If we return `h.vmBackups.Update(vmBackupCpy)`, we may lose `status.BackupTarget` in the new object.

**Solution:**
Return `vmBackupCpy`.

**Related Issue:**
https://github.com/harvester/harvester/issues/1588

**Test plan:**

* Create a VM Backup with v0.3.0 cluster
* Upgrade VM Backup CRD. Add `longhornBackupName` to `volumeBackups`. Add `secretBackups` and `backupTarget` to `status`.
  * https://github.com/harvester/harvester/pull/1447/commits/cb02453bf9aefdb06313438f09376a7e50c52569#diff-91562ef2708a3a87b91f4fc1ccefc8c95954eb8ca2949f1b76e5f16cc40273ce
  * https://github.com/harvester/harvester/pull/1487/commits/4345ab61f6154439f18cfabca816e108ba7d0626#diff-91562ef2708a3a87b91f4fc1ccefc8c95954eb8ca2949f1b76e5f16cc40273ce
Screen Shot 2021-11-29 at 11 33 02 AM
Screen Shot 2021-11-29 at 11 33 17 AM

* Upgrade harvester image from v0.3.0 to the master-head version.
  * Check backup target information moved from annotation to status.
  * Check `LonghornBackupName` is in `VolumeBackups`.
  * Check `<namespace>-<name>.cfg` is in remote backup target.
  * Check information in `<namespace>-<name>.cfg` is same as VM Backup resources.
* Upgrade harvester-webhook image from v0.3.0 to the master-head version and grant permission to harvester-webhook like the following.
  * https://github.com/harvester/harvester/pull/1447/commits/2c1b9c2c062dbad5e59df48af675182da1db7a7c#diff-cbaabc6d4bb79d51b3ac588ab97ffb29c95c69ec8b1d49c00ff8544353459b0e
Screen Shot 2021-11-29 at 11 20 09 AM
* Restore a VM from VM Backup.
  *  Check VM can be started and data is same.

### DIFF
--- a/pkg/controller/master/backup/backup.go
+++ b/pkg/controller/master/backup/backup.go
@@ -565,6 +565,11 @@ func (h *Handler) deleteVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
 
 func (h *Handler) uploadVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBackup, target *settings.BackupTarget) error {
 	var err error
+	// if users don't update VMBackup CRD, we may lose backup target data.
+	if vmBackup.Status.BackupTarget == nil {
+		return fmt.Errorf("no backup target in vmbackup.status")
+	}
+
 	if target == nil {
 		if target, err = settings.DecodeBackupTarget(settings.BackupTargetSet.Get()); err != nil {
 			return err
@@ -714,8 +719,5 @@ func (h *Handler) configureBackupTargetOnStatus(vmBackup *harvesterv1.VirtualMac
 		BucketName:   vmBackup.Annotations[backupBucketNameAnnotation],
 		BucketRegion: vmBackup.Annotations[backupBucketRegionAnnotation],
 	}
-	delete(vmBackup.Annotations, backupTargetAnnotation)
-	delete(vmBackup.Annotations, backupBucketNameAnnotation)
-	delete(vmBackup.Annotations, backupBucketRegionAnnotation)
 	return h.vmBackups.Update(vmBackupCpy)
 }


pkg/controller/master/backup/backup.go
```diff
@@ -714,8 +714,9 @@ func (h *Handler) configureBackupTargetOnStatus(vmBackup *harvesterv1.VirtualMac
...
 		BucketRegion: vmBackup.Annotations[backupBucketRegionAnnotation],
 	}
-	delete(vmBackup.Annotations, backupTargetAnnotation)
-	delete(vmBackup.Annotations, backupBucketNameAnnotation)
-	delete(vmBackup.Annotations, backupBucketRegionAnnotation)
-	return h.vmBackups.Update(vmBackupCpy)
+	delete(vmBackupCpy.Annotations, backupTargetAnnotation)
+	delete(vmBackupCpy.Annotations, backupBucketNameAnnotation)
+	delete(vmBackupCpy.Annotations, backupBucketRegionAnnotation)
+	_, err := h.vmBackups.Update(vmBackupCpy)
```
guangbochen: Would prefer to update the status directly and don't need to delete the annotations, also we can skip updating when the `status.backupTarget` is not empty, WDUT?

FrankYang0529: I tried it, but it looks like we don't have `UpdateStatus` in `VirtualMachineBackupClient`.

https://github.com/harvester/harvester/blob/655c251ada5808399c8b9c475abc75e437a51d09/pkg/generated/controllers/harvesterhci.io/v1beta1/virtualmachinebackup.go#L55-L64

weihanglo: You can get a generated `UpdateStatus` by dropping the pointer and using the struct directly.

https://github.com/harvester/harvester/blob/655c251ada5808399c8b9c475abc75e437a51d09/pkg/apis/harvesterhci.io/v1beta1/backup.go#L50

weihanglo (commented):
I didn't encountered the bug when I upgraded vmbackup CRD first before upgrading harvester workload from 0.3.0 to 1.0.0-rc1. I am think that this is an upgrade/migrate timing issue, and the CRD surely should be there sooner or later. 

In this fix, the returned `vmBackupCpy` still contains an old `ResourceVersion`, this might cause subsequent update failure if someone calls kube-api with that `vmBackupCpy` (for now no one does that).

Would it be better just let it die and it should recover once CRD being applied?

w13915984028:
> I didn't encountered the bug when I upgraded vmbackup CRD first before upgrading harvester workload from 0.3.0 to 1.0.0-rc1. I am think that this is an upgrade/migrate timing issue, and the CRD surely should be there sooner or later.
> 
> In this fix, the returned `vmBackupCpy` still contains an old `ResourceVersion`, this might cause subsequent update failure if someone calls kube-api with that `vmBackupCpy` (for now no one does that).
> 
> Would it be better just let it die and it should recover once CRD being applied?

	_, err := h.vmBackups.Update(vmBackupCpy)
	return vmBackupCpy, err
	
	---
	Do you mean, even after the `h.vmBackups.Update ` ,   if someone calls kube-api with that `vmBackupCpy`, it will fail?
	
	Are there some detailed examples? 
	
If some controllers are OnChange of vmBackup, when they runs before vmBackupController, 
 they get the current value (say, without Status), 
when they run after vmBackupController, they get the updated vmBackup with Status.

But for a general rule, the "Status" is a pointer of vmBackup, any usage of it should first check if it is nil.

pkg/controller/master/backup/backup.go
w13915984028: > 

@guangbochen regarding your question `"also we can skip updating when the status.backupTarget is not empty"`

This commit check `isBackupMissingStatus`, normally when it is resource created before v1.0.0, it is true, then create a `backupCpy.Status` pointer, a `backupCpy.Status.BackupTarget` pointer,  they are considered as place holder to differentiate old and new, also did the upgrade work. Finally update self.

no matter the value of the old backup target, the update of self looks is essential.


```
func isBackupMissingStatus(backup *harvesterv1.VirtualMachineBackup) bool {
	return backup.Status == nil || backup.Status.SourceSpec == nil || backup.Status.VolumeBackups == nil || backup.Status.BackupTarget == nil
}
```

```
func (h *Handler) OnBackupChange(key string, vmBackup *harvesterv1.VirtualMachineBackup) (*harvesterv1.VirtualMachineBackup, error) {
if isBackupMissingStatus(vmBackup) {
..
		return nil, h.initBackup(vmBackup, sourceVM, target)
..
}
```
	
```
func (h *Handler) initBackup(backup *harvesterv1.VirtualMachineBackup, vm *kv1.VirtualMachine, target *settings.BackupTarget) error {
...
backupCpy.Status = &harvesterv1.VirtualMachineBackupStatus{

	backupCpy.Status.BackupTarget = &harvesterv1.BackupTarget{
		Endpoint:     target.Endpoint,
		BucketName:   target.BucketName,
		BucketRegion: target.BucketRegion,
	}

	if _, err := h.vmBackups.Update(backupCpy); err != nil {
		return err
	}
	return nil
}
```

weihanglo:
> But for a general rule, the "Status" is a pointer of vmBackup, any usage of it should first check if it is nil.

That's right! Would it be better to check the nilbility in where it is used than where it is created?

pkg/controller/master/backup/backup.go
FrankYang0529: > a `backupCpy.Status.BackupTarget` pointer, they are considered as place holder to differentiate old and new, also did the upgrade work. Finally update self.

@weihanglo , I am not sure whether we can backport from pointer to struct. If we want to use struct, I will suggest let's do it in the next version API like `v1beta2`.

FrankYang0529:
> > But for a general rule, the "Status" is a pointer of vmBackup, any usage of it should first check if it is nil.
> 
> That's right! Would it be better to check the nilbility in where it is used than where it is created?

For updating without nil pointer panic, I will check whether `vmbackup.Status.BackupTarget` is nil in `uploadVMBackupMetadata`.

weihanglo (approved):
LGTM

guangbochen (approved):
LGTM

PR #1656, opened 2021-12-09
fix 1051: reset backup target to default

w13915984028:
Problem:

After PR of #1447, the "Use the default value" of "backup target" will be refused by webhook, the error message is:

admission webhook "validator.harvesterhci.io" denied the request: Destination URL hasn't been specified

It is not possible to reset backup target to the default value in the web UI.

Solution:

In webhook, check the backup target as per specific requirements to each field,  S3, NFS, Default are different.
For S3, there is an re-update operation, needs special check to identify it is from user input or internal

In controller, add process of default backup target.

Related Issue:
#1051

Test plan:

Setup a harvester cluster (one node is also OK)

1. The initial backup target is blank (default), when you create a VM and operate the "take backup", will be refused.

2. Set the backup target
2.1 When backup target has error (endpoint is not reachable/bucket name is wrong/bad credentials...), "save" will be refused with an error message, and after refreshing the web page, the backup target is still the last setting
2.2 When backup target is correct, it is saved. 
2.2.1 "take backup" on VM will be allowed
 (another commit will block user change the backup target if a backup target is ongoing)
 
2.3 "Edit" again, click "Use the default value" button and "save", it will be success, and backup target is reset to default value.
2.3.1 "take backup" on VM will be refused

2.4 Set a correct backup target of S3/minio, then "edit", if you leave input access key / ID  blank, and change the bucket name/region/endpoint,  save will NOT be allowed,   before this commit, you can do it, it is a BUG

2.5 Set a correct backup target of S3/minio, use kubectl can check a related secret created
```
kubectl get secrets -A | grep backup
longhorn-system                          harvester-backup-target-secret                                       Opaque                                8      57s
```
change the backup target to NFS or reset to default, this secret will be deleted.  Before this commit, it is leaved there.

### DIFF
--- a/pkg/controller/master/backup/backup.go
+++ b/pkg/controller/master/backup/backup.go
@@ -119,6 +119,8 @@ func (h *Handler) OnBackupChange(key string, vmBackup *harvesterv1.VirtualMachin
 		return nil, err
 	}
 
+	logrus.Debugf("OnBackupChange: vmBackup name:%s, target:%s:%s", vmBackup.Name, target.Type, target.Endpoint)
+
 	if isBackupReady(vmBackup) {
 		// We've changed backup target information to status since v1.0.0.
 		// For backport to v0.3.0, we move backup target information from annotation to status.
@@ -537,6 +539,12 @@ func (h *Handler) deleteVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
 		}
 	}
 
+	// when backup target has been reset to default, skip following
+	if target.IsDefaultBackupTarget() {
+		logrus.Debugf("vmBackup delete:%s, backup target is default, skip", vmBackup.Name)
+		return nil
+	}
+
 	if !IsBackupTargetSame(vmBackup.Status.BackupTarget, target) {
 		return nil
 	}
@@ -579,6 +587,12 @@ func (h *Handler) uploadVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
 		}
 	}
 
+	// when current backup target is default, skip following steps
+	// if backup target is default, IsBackupTargetSame is true when vmBackup.Status.BackupTarget is also default value
+	if target.IsDefaultBackupTarget() {
+		return nil
+	}
+
 	if !IsBackupTargetSame(vmBackup.Status.BackupTarget, target) {
 		return nil
 	}
@@ -618,13 +632,14 @@ func (h *Handler) uploadVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
 
 	shouldUpload := true
 	destURL := filepath.Join(metadataFolderPath, getVMBackupMetadataFileName(vmBackup.Namespace, vmBackup.Name))
-	if exist := bsDriver.FileExists(destURL); exist {
+	if bsDriver.FileExists(destURL) {
 		if remoteVMBackupMetadata, err := loadBackupMetadataInBackupTarget(destURL, bsDriver); err != nil {
 			return err
 		} else if reflect.DeepEqual(vmBackupMetadata, remoteVMBackupMetadata) {
 			shouldUpload = false
 		}
 	}
+
 	if shouldUpload {
 		logrus.Debugf("upload vm backup metadata %s/%s to backup target %s", vmBackup.Namespace, vmBackup.Name, target.Type)
 		if err := bsDriver.Write(destURL, bytes.NewReader(j)); err != nil {
--- a/pkg/controller/master/backup/backup_metadata.go
+++ b/pkg/controller/master/backup/backup_metadata.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/longhorn/backupstore"
 	_ "github.com/longhorn/backupstore/nfs"
@@ -84,14 +85,21 @@ func (h *MetadataHandler) OnBackupTargetChange(key string, setting *harvesterv1.
 		return nil, nil
 	}
 
-	target, err := settings.DecodeBackupTarget(settings.BackupTargetSet.Get())
+	target, err := settings.DecodeBackupTarget(setting.Value)
 	if err != nil {
 		return setting, err
 	}
 
+	logrus.Debugf("backup target change, sync vm backup:%s:%s", target.Type, target.Endpoint)
+
+	// when backup target is reset to default, do not trig sync
+	if target.IsDefaultBackupTarget() {
+		return nil, nil
+	}
+
 	if err = h.syncVMBackup(target); err != nil {
-		logrus.Errorf("can't sync vm backup metadata, err: %v", err)
-		h.settings.Enqueue(setting.Name)
+		logrus.Errorf("can't sync vm backup metadata, target:%s:%s, err: %v", target.Type, target.Endpoint, err)
+		h.settings.EnqueueAfter(setting.Name, 5*time.Second)
 		return nil, nil
 	}
 
--- a/pkg/controller/master/backup/backup_target.go
+++ b/pkg/controller/master/backup/backup_target.go
@@ -10,6 +10,7 @@ import (
 
 	longhornv1 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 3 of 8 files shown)


pkg/controller/master/backup/backup.go
```diff
@@ -366,7 +368,7 @@ func (h *Handler) reconcileVolumeSnapshots(vmBackup *harvesterv1.VirtualMachineB
 		if volumeSnapshot == nil {
 			volumeSnapshot, err = h.createVolumeSnapshot(vmBackupCpy, volumeBackup)
 			if err != nil {
-				logrus.Debugf("create volumeSnapshot %s/%s error: %v", volumeSnapshot.Namespace, volumeSnapshot.Name, err)
```
w13915984028: when err, volumeSnapshot is nil, access volumeSnapshot.Namespace caused panic @FrankYang0529

pkg/controller/master/backup/backup_target.go
```diff
@@ -79,32 +80,68 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
...
+			return nil, nil
+		}
+
+		return nil, fmt.Errorf("Invalid backup target type:%s or parameter", target.Type)
+	}
 }
 
-func (h *TargetHandler) updateBackupTargetSetting(setting *harvesterv1.Setting, target *settings.BackupTarget, err error) (*harvesterv1.Setting, error) {
+func (h *TargetHandler) reUpdateBackupTargetSettingSecret(setting *harvesterv1.Setting, target *settings.BackupTarget, err error) (*harvesterv1.Setting, error) {
 	harvesterv1.SettingConfigured.SetError(setting, "", err)
```
w13915984028: `harvesterv1.SettingConfigured.SetError(setting, "", err)  `
--> when calls happen, err can only be nil.    because all errors have already been returned, no chance to here.
who/where  set the "SettingConfigured" when real error happens?
@FrankYang0529 @guangbochen

w13915984028:
FYI. 
I still have a few open questions regarding this commit, but would like to have a pull request to review the overall code changes first.

pkg/controller/master/backup/backup.go
FrankYang0529: Yes, you're right. Thank you!

pkg/controller/master/backup/backup_target.go
FrankYang0529: Yes, it looks like err can only be nil before my last update. For me, it's okay to remove err parameter here.

https://github.com/harvester/harvester/pull/1447/commits/2c1b9c2c062dbad5e59df48af675182da1db7a7c#diff-b8a98fcbcd22961df4c633ac883a54f7e28155bcbecccd9a2aa966f004d07035

w13915984028:
@FrankYang0529 @guangbochen 

The "use the default value" of backup target, change / refactor the original part, please take time to review it, thanks.
Please skip the "CodeFactor" issues, some of them are already existing, only because this commit changes some of the related file.

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -84,14 +85,22 @@ func (h *MetadataHandler) OnBackupTargetChange(key string, setting *harvesterv1.
 		return nil, nil
 	}
 
-	target, err := settings.DecodeBackupTarget(settings.BackupTargetSet.Get())
+	//target, err := settings.DecodeBackupTarget(settings.BackupTargetSet.Get())
```
FrankYang0529: Probably, we can remove this line

FrankYang0529 (approved):
LGTM. I can verify following test cases.

1. Set default backup target to S3.
2. Set S3 to NFS.
3. Set NFS to default backup target.
4. Set default backup target to NFS.

pkg/controller/master/backup/backup_metadata.go
w13915984028: > Probably, we can remove this line

removed now.

w13915984028:
The panic caused by nil pointer access, is merged via 
https://github.com/harvester/harvester/pull/1674

pkg/controller/master/backup/backup_target.go
```diff
@@ -204,6 +263,22 @@ func (h *TargetHandler) updateBackupTargetSecret(target *settings.BackupTarget)
...
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	if err := h.longhornSettings.Delete(util.LonghornSystemNamespaceName, longhornBackupTargetSecretSettingName, nil); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
```
guangbochen: why do we need to delete the LH setting?

pkg/controller/master/backup/backup_target.go
```diff
@@ -204,6 +263,22 @@ func (h *TargetHandler) updateBackupTargetSecret(target *settings.BackupTarget)
 	return h.updateLonghornBackupTargetSecretSetting(target)
 }
 
+func (h *TargetHandler) deleteBackupTargetSecret(target *settings.BackupTarget) error {
+	if err := h.secrets.Delete(util.LonghornSystemNamespaceName, util.BackupTargetSecretName, nil); err != nil {
```
guangbochen: if err != nil && !apierrors.IsNotFound(err)

pkg/webhook/resources/setting/validator.go
```diff
@@ -146,6 +147,59 @@ func validateVMForceResetPolicy(setting *v1beta1.Setting) error {
 	return nil
 }
 
+// chech if this backup target is updated again by controller to strip secret information
+func (v *settingValidator) isUpdatedS3BackupTarget(target *settings.BackupTarget) bool {
+	if target.SecretAccessKey != "" || target.AccessKeyID != "" {
```
guangbochen: return if the type is not `s3`

pkg/controller/master/backup/backup_target.go
```diff
@@ -79,32 +80,79 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
...
+
+			// it has no effect to call
+			//  harvesterv1.SettingConfigured.SetError(setting, "", fmt.Errorf("reset value to default"))
+			//  to set True/False of a config, as the operation itself will finally return nil, nil
+			// hence
+			//  it is not effective to call harvesterv1.SettingConfigured.IsTrue(targetSetting), e.g. in checkBackupTargetConfigured
+			//  to check if the backup target is set or not
+			//  as the "use the default value" action itself will be success, it "reset" the value to default, but itself is successful
+			// the explicit check of "default value" is added to related parts
+			return nil, nil
```
guangbochen: an empty string is allowed for the LH backup target, so don't delete the setting CR.
also can we simply the return comments, thanks.

pkg/controller/master/backup/backup_target.go
```diff
@@ -79,32 +80,79 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
...
+		}
+
+		// delete the may existing previous secret of S3
+		if err = h.deleteBackupTargetSecret(target); err != nil {
+			return nil, err
+		}
+
+		// same operation as S3 does, but it looks to be not as expected
+		// the framework will set the True/False according to the final return of here
+		harvesterv1.SettingConfigured.SetError(setting, "", err)
```
guangbochen: remove it since it will be updated by https://github.com/harvester/harvester/blob/master/pkg/controller/master/setting/handler.go#L83

pkg/controller/master/backup/backup_target.go
```diff
@@ -79,32 +80,79 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
...
+		}
 
-	if target.Type == settings.S3BackupType {
 		if err = h.updateBackupTargetSecret(target); err != nil {
 			return nil, err
 		}
-	}
 
-	return h.updateBackupTargetSetting(setting.DeepCopy(), target, err)
+		return h.reUpdateBackupTargetSettingSecret(setting, target, err)
```
guangbochen: I think we can just update the secrets on the validation webhook and strip the credentials through mutation webhook, and therefore the controller doesn't need to reconcile it.

pkg/webhook/resources/setting/validator.go
w13915984028: > return if the type is not `s3`

It is a local function, the caller has checked S3, then call here, it is not necessary to check s3 again.

pkg/controller/master/backup/backup_target.go
w13915984028: Longhorn may be in similar situation as us, it the backup target is reset back to default, has some hidden bugs.

The reset to empty, is different than initial empty.  When initial, check like `harvesterv1.SettingConfigured.IsTrue(targetSetting)`  returns false, but when resetting back to initial, the `harvesterv1.SettingConfigured.IsTrue(targetSetting)` returns true.

The better way is to clear it.

pkg/controller/master/backup/backup_target.go
w13915984028: This idea may be verified and committed in next PR.
The secrets are used to create secrets in api server first, finally be stripped, not sure if mutation webhook works.

pkg/controller/master/backup/backup_target.go
w13915984028: > remove it since it will be updated by https://github.com/harvester/harvester/blob/master/pkg/controller/master/setting/handler.go#L83

It is added to keep logic same as before, because in s3 path, it called this API. 
I also wondered, this explicit setting has no effect.

pkg/controller/master/backup/backup_target.go
w13915984028: updated.

pkg/webhook/resources/setting/validator.go
w13915984028: updated

pkg/controller/master/backup/backup_target.go
w13915984028: updated

pkg/controller/master/backup/backup_target.go
w13915984028: keep now, wait for further discuss.

pkg/controller/master/backup/backup_target.go
w13915984028: updated

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -84,14 +85,21 @@ func (h *MetadataHandler) OnBackupTargetChange(key string, setting *harvesterv1.
 		return nil, nil
 	}
 
-	target, err := settings.DecodeBackupTarget(settings.BackupTargetSet.Get())
+	target, err := settings.DecodeBackupTarget(setting.Value)
```
gitlawr, line 88: Will it return json unmarshal error when `setting.Value` is empty?

w13915984028: setting.Value is checked nil before calling Decode,  Decode itself does not check if/not value is nil.

```
	if setting == nil || setting.DeletionTimestamp != nil ||
		setting.Name != settings.BackupTargetSettingName || setting.Value == "" {
		return nil, nil
	}

	-target, err := settings.DecodeBackupTarget(settings.BackupTargetSet.Get())
	//target, err := settings.DecodeBackupTarget(settings.BackupTargetSet.Get())
	+target, err := settings.DecodeBackupTarget(setting.Value)
```

gitlawr (approved):
LGTM

guangbochen (approved):
LGTM, thanks.

Issue #1661, opened 2021-12-10
[BUG] rename vm-force-deletion-policy for vm-force-reset-policy


abonillabeeche:
The setting name infers a deletion of the virtual machine based on this policy but the real effect is that the virtual machine is being reset in order to start on another node. We should rename this setting to avoid confusion.

yasker:
@guangbochen can we rename? I think the name comes from a similar setting in Longhorn but since we're deleting pods, not VMs, it's better to use `reset` instead of `deletion` in the name.

FrankYang0529:
@guangbochen, we may also need a PR to rename `vm-force-deletion-policy` to `vm-force-reset-policy` on dashboard. Thank you.

guangbochen:
it should be updated automatically once the backend is changed, but need to sync the setting whitelist cc @WuJun2016, also need to change the description to `Config the force-reset action when a VM is stuck on a node that is down.`

yasker:
Confirmed that the setting name has been updated in Harvester when using UI from upstream.
<img width="1568" alt="Screen Shot 2021-12-15 at 8 33 38 AM" src="https://user-images.githubusercontent.com/3344618/146226444-93e1ef2a-3f9b-457f-bdb2-5744720ae821.png">

**However, the option doesn't show up when accessing the Harvester UI from Rancher.** @guangbochen @n313893254  is it by design?

n313893254:
My bad. As harvester UI dose not create PR for dashboard UI for now. we should replace the ui-dashboard-index setting of Rancher to 'https://releases.rancher.com/harvester-ui/dashboard/latest/index.html' to test Harvester UI from Rancher.

![image.png](https://images.zenhubusercontent.com/59db4297b0222d5de47ad632/9aa57f3d-9e7f-4b5e-96c6-6478764b1c4b)

noahgildersleeve:
Tested in master-5e8d9764-head. Verified as fixed in the harvester ui-index that hasn't been merged upstream yet
## Test Steps
1. Update ui-offline-preferred to Remote
2. Update ui-dashboard-index to https://releases.rancher.com/harvester-ui/dashboard/latest/index.html
3. Check advanced settings in Harvester UI and Rancher Harvester dashboard and verify setting is named vm-force-reset-policy
![Screenshot from 2021-12-16 13-00-13](https://user-images.githubusercontent.com/83787952/146448317-a259d86d-2020-4bed-adc2-f19ecf0d3fbb.png)
[...]

PR #1664, opened 2021-12-13
rename: vm-force-deletion-policy to vm-force-reset-policy

FrankYang0529:
**Problem:**
We don't really delete VM. We just delete the pod of VM and let VM be rescheduled.

**Solution:**
Change `vm-force-deletion-policy` to `vm-force-reset-policy`.

**Related Issue:**
https://github.com/harvester/harvester/issues/1661

**Test plan:**

Case 1: `vm-force-reset-policy` has already been enabled.
* Set `vm-force-reset-policy` as `'{"enable": true, "period": 60}'`.

Case 2: Enable `vm-force-reset-policy` after a node is down.
* Set `vm-force-reset-policy` as `'{"enable": false, "period": 60}'`.
* Set `vm-force-reset-policy` as `'{"enable": true, "period": 60}'`.

### DIFF
--- a/pkg/controller/master/node/node_down_controller.go
+++ b/pkg/controller/master/node/node_down_controller.go
@@ -44,15 +44,15 @@ func NodeDownRegister(ctx context.Context, management *config.Management, option
 	}
 
 	nodes.OnChange(ctx, nodeDownControllerName, nodeDownHandler.OnNodeChanged)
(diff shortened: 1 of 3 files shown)


gitlawr (approved):
LGTM

guangbochen (approved):
LGTM

PR #1675, opened 2021-12-14
Support bundle: collect more namespaces

bk201:
Collect resources from more namespaces and add a setting `support-bundle-namespaces` if users want to include even more.

From:
Screen Shot 2021-12-14 at 5 50 36 PM

To:
Screen Shot 2021-12-14 at 11 59 09 PM

### DIFF
--- a/pkg/controller/master/supportbundle/controller.go
+++ b/pkg/controller/master/supportbundle/controller.go
@@ -19,7 +19,6 @@ import (
 type Handler struct {
 	supportBundles          v1beta1.SupportBundleClient
 	supportBundleController v1beta1.SupportBundleController
-	settingCache            v1beta1.SettingCache
 	nodeCache               ctlcorev1.NodeCache
 	podCache                ctlcorev1.PodCache
 	deployments             ctlappsv1.DeploymentClient
--- a/pkg/controller/master/supportbundle/manager.go
+++ b/pkg/controller/master/supportbundle/manager.go
@@ -18,7 +18,6 @@ import (
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/controller/master/supportbundle/types"
 	"github.com/harvester/harvester/pkg/settings"
-	"github.com/harvester/harvester/pkg/util"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


gitlawr:
Can you clarify the reason for this change? I think many of these namespaces are irrelevant to the Harvester use case.

bk201:
> Can you clarify the reason for this change? I think many of these namespaces are irrelevant to the Harvester use case.

I follow the logic here: https://github.com/harvester/harvester/blob/1c02888d76c00f18a4cf828f4eb573b722b885ad/pkg/controller/master/rancher/embedded_rancher.go#L60-L70
How about not including the rancher settings but only these ones:
```
		"cattle-dashboards",
		"cattle-fleet-clusters-system",
		"cattle-monitoring-system",
		"fleet-local",
		"harvester-system",
		"local",
		"longhorn-system",
```

bk201:
I file this change because when I need to check ManagedChart and Bundle, I found `fleet-local` namespace is not included.

gitlawr:
Yeah, I would prefer we only collect the namespaces needed for debugging Harvester

weihanglo (commented):
The default is somehow lengthy and scary. BTW, where do we set the `system-namespaces` setting in Harvester?

guangbochen:
It was not exposed via the settings page and users can only modify it using the kubectl or API-UI, I prefer to add a new setting named `support-bundle-namespaces` that allows users to specify any additional namespaces required for the support bundle.

gitlawr:
Most are set by Rancher. We add a few Harvester-related namespaces. It's mainly used for resource filtering in the dashboard

weihanglo:
Second for the `support-bundle-namespaces` setting! Sounds flexible enough for what this PR needs originally. Do we have list-like UI for setting at this moment? Or put simply just ask user writing comma-separate string.

guangbochen:
the UI team is capable to do any customization, but for now, maybe be just using a comma-separated string is good enough due to limited time concerns(but can be improved in the next version).

bk201:
Reduce the default NS list to:
```
		"cattle-dashboards",
		"cattle-fleet-clusters-system",
		"cattle-monitoring-system",
		"fleet-local",
		"harvester-system",
		"local",
		"longhorn-system",
```

and add a new setting `support-bundle-namespaces` to let users add more.

guangbochen (approved):
LGTM, cc @WuJun2016 will need to add `support-bundle-namespaces` after this PR is merged,  add a description as `Specify resources in other namespaces to be collected by the support package`

gitlawr (approved):
LGTM

Issue #1644, opened 2021-12-08
[BUG] Harvester pod crashes after upgrading from v0.3.0 to v1.0.0-rc1 (contain vm backup before upgrade)


TachunLin:
**Describe the bug**

Prepare a 4 nodes harvester cluster with v0.3.0 with a vm backup then do manual upgrade to v1.0.0-rc1 
Can't access harvester, harvester pod crashes after the upgrade. 

**To Reproduce**
1. Download harvester v0.3.0 iso and do checksum
2. Download harvester v1.0.0 iso and do checksum
3. Use ISO Install a 4 nodes harvester cluster 
4. Create several OS images from URL
5. Create ssh key
6. Enable vlan network with `harvester-mgmt`
7. Create virtual network `vlan1` with id `1`
8. Create 2 virtual machines
- ubuntu-vm: 2 core, 4GB memory, 30GB disk
- centos-vm: 2 core, 4GB memory, 30GB disk

![image](https://user-images.githubusercontent.com/29251855/144856833-1fccfb10-b9b2-4f5d-a9e5-c0cbbb4b0aca.png)

9. Setup backup target
![image](https://user-images.githubusercontent.com/29251855/144973526-7fc71bc4-8a95-41f7-b660-5aa22a8fe7be.png)

10. Take a backup from ubuntu vm
![image](https://user-images.githubusercontent.com/29251855/144973570-4db239c1-421a-4932-9780-1ed154e27c7d.png)

**upgrade process**
Follow the manual upgrade steps to upgrade from v0.3.0 to v1.0.0-rc1
https://github.com/harvester/docs/pull/67/files

**Expected behavior**

Can manual upgrade harvester from v0.3.0 to v1.0.0-rc1 with vm backup.
Harvester pods keeps working without crash

**Support bundle**

[bundle.zip](https://github.com/harvester/harvester/files/7673834/bundle.zip)

**Environment:**
 - Harvester ISO version before upgrade: v0.3.0
 - Harvester ISO version after upgrade: v1.0.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 4 nodes harvester cluster on local kvm machines

**Harvester network information**
 - VIP: 192.168.122.194
 - node1: 192.168.122.36
 - node2: 192.168.122.85
 - node3: 192.168.122.186
 - node4: 192.168.122.97

[...]

PR #959, opened 2021-06-17
Remove harvester user

lucidd:
**Problem:**
explained in issue

**Solution:**
explained in issue

**Related Issue:**

https://github.com/harvester/harvester/issues/921

**Test plan:**

PR #1680, opened 2021-12-15
Add harvester-crd chart

FrankYang0529:

**Problem:**
Currently, helm can't upgrade CRDS.
https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations

**Solution:**
Separate CRDs into another chart `harvester-crd`.

**Related Issue:**
https://github.com/harvester/harvester/issues/1644

**Test plan:**


scripts/generate-manifest
```diff
@@ -4,7 +4,7 @@ set -e
 # The root of the harvester directory
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
 
-out_dir="${ROOT_DIR}/deploy/charts/harvester/crds"
+out_dir="${ROOT_DIR}/deploy/charts/harvester-crd/templates/crds"
```
guangbochen: we can use the templates directly since it is a `harvester-crd` chart.

deploy/charts/harvester/Chart.yaml
```diff
@@ -49,5 +49,5 @@ dependencies:
     version: 0.1.0
     repository: https://kube-vip.io/helm-charts
   - name: harvester-load-balancer
-    version: 0.1.1
+    version: 0.1.2
```
weihanglo, line 52: I saw these changes. Do we ought to run `helm dep update` under this directory for updating deps tarball?

scripts/generate-manifest
FrankYang0529: Updated it. Thank you.

deploy/charts/harvester/Chart.yaml
FrankYang0529: Added it. Thank you.

deploy/charts/harvester/crds/harvesterhci.io_users.yaml
```diff
@@ -1,71 +0,0 @@
-
```
weihanglo, line 1: It would be better if we can point out the reason of deleting this CRD.

guangbochen (approved):

weihanglo (approved):
Seems fair. Thanks for the PR.

gitlawr (approved):
LGTM

deploy/charts/harvester/crds/harvesterhci.io_users.yaml
FrankYang0529: I am not sure what is this CRD for because I have never seen this resource since I joined Harvester project. Do you have any idea about this? Thank you! @guangbochen @gitlawr

guangbochen: The `users.harvesterhci.io` CRD is a leftover that we missed to clean up, currently we are using rancher users only.

gitlawr: FYI https://github.com/harvester/harvester/pull/959

bk201 (approved):

Issue #1692, opened 2021-12-16
[BUG] After backup target is set, it is not usable 


w13915984028:
**Describe the bug**

**To Reproduce**
1. Install a new Harvester
2. Setup a correct backup target
3. Take backup action
4. It is refused

![image](https://user-images.githubusercontent.com/31133476/146414255-b328d6e4-4636-42a0-b8f0-bea050c907a7.png)

**Expected behavior**

After backup target is setting, it should be allowed to do backup

**Support bundle**

**Environment:**
 - Harvester ISO version:  Harvester master-head of 16.12.2021
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

w13915984028:
@guangbochen  @FrankYang0529 
This is an partial update to PR 1656, follow the original process.
https://github.com/harvester/harvester/pull/1656

Not as our expectation, the flag still needs to be set explicitly.

Otherwise below function  `checkBackupTargetConfigured ->  harvesterv1.SettingConfigured.IsTrue(targetSetting)`   will return error.  
Alternative is to change it, but not this time.
```
pkg/api/vm/handler.go
 func (h *vmActionHandler) checkBackupTargetConfigured() error 
 
 if err == nil && harvesterv1.SettingConfigured.IsTrue(targetSetting) {   ----> not meet, and return error
 
```

w13915984028:
BTW:  With this fix, the #1673  is verified on Harvester ISO of today.

w13915984028:
Solved in master-head version.

Issue #1673, opened 2021-12-14
[BUG] Error : can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists


futuretea:
**Describe the bug**

harvester logs:
```
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
time="2021-12-14T06:24:28Z" level=error msg="can't sync vm backup metadata, err: virtualmachinebackups.harvesterhci.io \"test-backup\" already exists"
```

**To Reproduce**
1. create a vm "test" in non-default namespace such as 'harvester-public'
2. create backup "test-backup" from vm "test"
3. restart harvester (use kubectl to delete harvester pod)

**Expected behavior**

no such error log

**Support bundle**

**Environment:**
 - Harvester ISO version:  harvester v1.0.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

w13915984028:
[...]

PR #1693, opened 2021-12-16
fix 1692 set backup target is configured flag after setting

w13915984028:

**Problem:**

Without setting the flag, the "take backup" action is refused

**Solution:**

Set the configured flag when setting backup target
Then it is allowed to "take backup"

**Related Issue:**
#1692

**Test plan:**

1. Install Harvester
2. Set backup target
3. "take backup" on VM, and is allowed
image

4. Edit backup target to default value
5. "take backup" on VM, and is refused

image

### DIFF
--- a/pkg/controller/master/backup/backup_target.go
+++ b/pkg/controller/master/backup/backup_target.go
@@ -111,6 +111,9 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
 			return nil, err
 		}
 
+		// set configured flag, as s3 does
+		harvesterv1.SettingConfigured.SetError(setting, "", nil)
+
 		return nil, nil
 
 	default:
@@ -137,6 +140,9 @@ func (h *TargetHandler) reUpdateBackupTargetSettingSecret(setting *harvesterv1.S
 	if target.Type != settings.S3BackupType {
 		return nil, nil
 	}
+
+	// set configured flag
+	harvesterv1.SettingConfigured.SetError(setting, "", nil)
 
 	// reset the s3 credentials to prevent controller reconcile and not to expose secret key
 	target.SecretAccessKey = ""


w13915984028:
@guangbochen @FrankYang0529
This is an partial update to PR 1656, follow the original process.
#1656

Not as our expectation, the flag still needs to be set explicitly.

Otherwise below function `checkBackupTargetConfigured -> harvesterv1.SettingConfigured.IsTrue(targetSetting`) will return error.
Alternative is to change it, but not this time.

```
pkg/api/vm/handler.go
 func (h *vmActionHandler) checkBackupTargetConfigured() error 
 
 if err == nil && harvesterv1.SettingConfigured.IsTrue(targetSetting) {   ----> not meet, and return error
 
```

w13915984028:
BTW:  With this fix, the https://github.com/harvester/harvester/issues/1673 is verified on Harvester ISO of today.

guangbochen (approved):
LGTM

pkg/controller/master/backup/backup_target.go
```diff
@@ -111,6 +111,9 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
 			return nil, err
 		}
 
+		// set configured flag, as s3 does
+		harvesterv1.SettingConfigured.SetError(setting, "", nil)
```
gitlawr, line 115: We need to get a deep copy of the resource, change it then use the client to update it. Otherwise, we are updating the resource in the memory cache and it's not persistent in etcd.
FYI this is how we did this before: https://github.com/harvester/harvester/blob/v0.3/pkg/controller/master/backup/backupTaget.go#L95

FrankYang0529 (approved):
LGTM

pkg/controller/master/backup/backup_target.go
guangbochen: @gitlawr I think u already did in https://github.com/harvester/harvester/blob/f5bde7fa462c09d0513b8f370f71f2dd974f298c/pkg/controller/master/setting/handler.go#L65 so it is not needed?

gitlawr: It only affects specific settings declared in https://github.com/harvester/harvester/blob/f5bde7fa462c09d0513b8f370f71f2dd974f298c/pkg/controller/master/setting/register.go#L47-L54

gitlawr (approved):

PR #1706, opened 2021-12-20
Upgrade: remove the GRUB rebranding workaround

bk201:
Remove the GRUB rebranding workaround because upstream has resolved the problem.


weihanglo:
Was testing this PR with harvester/harvester-installer#199. I can build the ISO successfully but cannot boot from it. The boot process seems stuck at squashfs.

bk201:
This will  be addressed together in https://github.com/harvester/harvester/pull/2166

Issue #1763, opened 2022-01-04
[BUG] Node promotion might be stuck occasionally


bk201:
**Describe the bug**

Sometimes a node promotion (agent ->server) might be stuck.

**To Reproduce**

Create a 3-node cluster and the promotion will start automatically. There is a chance the promotion never ends.

```
$ kubectl get nodes
NAME    STATUS                     ROLES                       AGE    VERSION
node1   Ready                      control-plane,etcd,master   14m    v1.21.7+rke2r1
node2   Ready,SchedulingDisabled   <none>                      9m3s   v1.21.7+rke2r1
node3   Ready                      <none>                      9m3s   v1.21.7+rke2r1
node4   Ready                      <none>                      9m3s   v1.21.7+rke2r1
```

harvester-promote-node2 logs:
```
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
harvester-promote-node2-l8jpr Waiting for custom machine <no value>...
```

**Expected behavior**

The promotion should succeed.

**Support bundle**

**Environment:**
 - Harvester ISO version: `v1.0.0`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

lanfon72:
Verified this bug has been fixed.

Tested on `v1.0.0` 10 times, node promotion might stuck occasionally.
Tested on `master-ebde990d-head` 10 times, node promotion successfully every times.

PR #1764, opened 2022-01-04
Node promotion: wait until custom machine label exists

bk201:
The label `cluster.x-k8s.io/machine` of a node might be set after the node promotion script is run. Wait until it's set.

**Related Issue:**
https://github.com/harvester/harvester/issues/1763

**Test plan:**
- Create a cluster that contains more than or equal to 3 nodes.
- 3 nodes should become control nodes. Role is `control-plane,etcd,master` in `kubectl get nodes` output.

Note: there is no reliable way to repro this issue, we can keep monitoring if this is fixed.


johnliu55tw (approved):
Verified that with this change, nodes are promoted as control plane as usual.

guangbochen (approved):
LGTM

Issue #1727, opened 2021-12-21
[BUG] Move support bundle image version out of the backend code


bk201:
**Describe the bug**
Support bundle kit image is coded as the default value of a setting in the backend. https://github.com/bk201/harvester/blob/d52cb934ed7bc2c1a22d693b6d56709e18b1d190/pkg/settings/settings.go#L29
This makes harvester-installer hard to automate the image pulling process. (Currently, it's a hard-coded value: https://github.com/harvester/harvester-installer/blob/3af28b4701bc5cd05f9ec3e8ac1b0111cf36de39/scripts/build-bundle#L108

We need to move the image value out of the code.
A proposal is to move to the chart values file.

FrankYang0529:
Proposal:
1. Add support bundle image value to `harvester/deploy/charts/harvester/values.yaml` and remove `support-bundle-image` in Settings.
2. Get ManagedChart `fleet-local/harvester`. If there is support bundle image value, it means users replace the default helm chart value, so we use it to create support bundle Deployment.
3. If there is not, we get Apps `harvester-system/harvester` to get the default support bundle image value.

FrankYang0529:
### Test plan

1. Check the controller can get support bundle image version from App.

* Set following yaml in `spec.chart.values` of `harvester-system/harvester` app:
```
support-bundle-kit:
  image:
    imagePullPolicy: Always
    repository: rancher/support-bundle-kit
    tag: v0.0.4
```
* Create a support bundle.
* Check image pull policy is `Always`, repository is `rancher/support-bundle-kit`, and tag is `v0.0.4` in support bundle pod.
* Check default value of `support-bundle-image` setting is `{"repository":"rancher/support-bundle-kit","tag":"v0.0.4","imagePullPolicy":"Always"}`.

2. Check the controller can merge support bundle image value in App and ManagedChart.

* Following above case, set following yaml in `spec.values` of `fleet-local/harvester` managed chart:
```
support-bundle-kit:
  image:
    imagePullPolicy: IfNotPresent
```
* Create a support bundle.
* Check image pull policy in support bundle pod is `IfNotPresent`.
* Check default value of `support-bundle-image` setting is `{"repository":"rancher/support-bundle-kit","tag":"v0.0.4","imagePullPolicy":"IfNotPresent"}`.

3. Check value from users can override the default value in `support-bundle-image` setting

[...]

PR #1771, opened 2022-01-04
Move support bundle image to chart

FrankYang0529:
**Problem:**
We set the default support bundle image value in the code. It's hard to maintain the version between harvester and harvester-installer.

**Solution:**
Move support bundle image value to harvester chart, so we can get the default support bundle image version from the chart. If users would like to override it, they can still set the value in the `support-bundle-image` setting.

**Related Issue:**
https://github.com/harvester/harvester/issues/1727

**Test plan:**

1. Check the controller can get support bundle image version from App.

* Set following yaml in `spec.chart.values` of `harvester-system/harvester` app:
```
support-bundle-kit:
  image:
    imagePullPolicy: Always
    repository: rancher/support-bundle-kit
    tag: v0.0.4
```
* Create a support bundle.
* Check image pull policy, repository, and tag in support bundle pod are the same.
* Check default value of `support-bundle-image` setting is `{"repository":"rancher/support-bundle-kit","tag":"v0.0.4","imagePullPolicy":"Always"}`.

2. Check the controller can merge support bundle image value in App and ManagedChart.

* Following above case, set following yaml in `spec.values` of `fleet-local/harvester` managed chart:
```
support-bundle-kit:
  image:
    imagePullPolicy: IfNotPresent
```
* Create a support bundle.
* Check image pull policy in support bundle pod is `IfNotPresent`.
* Check default value of `support-bundle-image` setting is `{"repository":"rancher/support-bundle-kit","tag":"v0.0.4","imagePullPolicy":"IfNotPresent"}`.

3. Check value from users can override the default value in `support-bundle-image` setting

* Set value in `support-bundle-image` setting as `{"repository":"rancher/support-bundle-kit","tag":"v0.0.5","imagePullPolicy":"IfNotPresent"}`.
* Create a support bundle.
* Check image tag is `v0.0.5` in support bundle pod.


pkg/controller/master/supportbundle/controller.go
```diff
@@ -50,6 +67,66 @@ func (h *Handler) OnSupportBundleChanged(key string, sb *harvesterv1.SupportBund
...
+		return "", corev1.PullPolicy(""), fmt.Errorf("unknown image yaml struct in %s", SupportBundleRespository)
+	}
+
+	for _, key := range []string{"repository", "tag", "imagePullPolicy"} {
+		if _, ok = imageYaml[key]; !ok {
+			return "", corev1.PullPolicy(""), fmt.Errorf("%s is not in %s.image yaml", key, SupportBundleRespository)
+		}
+	}
+	return fmt.Sprintf("%s:%s", imageYaml["repository"].(string), imageYaml["tag"].(string)), corev1.PullPolicy(imageYaml["imagePullPolicy"].(string)), nil
+}
```
weihanglo: [`mapstructure`](https://pkg.go.dev/github.com/mitchellh/mapstructure) is in harvester's dependency graph, so I guess we can just use it and it's really self-explaining. Something like this [go playground](https://go.dev/play/p/uHkrVGfhnw1)

```go
func (h *Handler) getSupportBundleImageFromYaml(supportBundleYaml map[string]interface{}) (string, corev1.PullPolicy, error) {
	// support bundle yaml struct
	var imageYaml struct {
		Image struct {
			ImagePullPolicy corev1.PullPolicy
			Repository string
			Tag string
		}
	}
	if err := mapstructure.Decode(supportBundleYaml, &imageYaml); err != nil {
		return "", corev1.PullPolicy(""), err
	}
	return fmt.Sprintf("%s:%s", imageYaml.Image.Repository, imageYaml.Image.Tag), imageYaml.Image.ImagePullPolicy, nil
}
```

weihanglo (approved):
The functionality is fine. Thanks!

pkg/controller/master/supportbundle/controller.go
FrankYang0529: Thanks for your suggestion! I move `getSupportBundleImageFromYaml` to `util.GetImageFromYaml`.

guangbochen (commented):
I would still prefer to allow users to customize the support bundle image accordingly, by default the setting value should be empty and loaded from the catalog or chart config.
- we should check and update the default value when the Harvester version is bumped
- if the setting values are set then use the values one, otherwise use the default one.
- optional: we may put both `image` and `imagePullPolicy` as one JSON string

FrankYang0529:
@guangbochen, do you think it's better to let users maintain all harvester-related images versions through managed chart? Although we don't deploy a support-bundle pod by the managed chart, we create it by harvester controller. Ideally, support-bundle is also a part of harvester images.

guangbochen:
I don't think we should allow users to change the Harvester `managedChart` values or images directly as they should only be managed by the harvester system with an OS version upgrade, however, the support bundle image is an exception as it is an independent system tool instead of Harvester system component. 
Also, saving the SB image to the setting will simply avoid parsing the YAML config each time uses it.

deploy/charts/harvester/values.yaml
```diff
@@ -438,3 +438,9 @@ kube-vip-cloud-provider:
   image:
     repository: kubevip/kube-vip-cloud-provider
     tag: 0.1
+
+support-bundle-kit:
+  image:
+    imagePullPolicy: IfNotPresent
+    repository: rancher/support-bundle-kit
+    tag: v0.0.5
```
bk201: I'd suggest changing the tag to `master-head` and we can use a formal release tag when creating a new Harvester release.

FrankYang0529: Updated it. Thank you.

bk201 (approved):
LGTM

weihanglo (approved):
LGTM

guangbochen (approved):

Issue #1702, opened 2021-12-18
[BUG] Don't allow restart/stop vm when backup is in progress


khushboo-rancher:
**Describe the bug**
The vms can be restarted or stopped when backup is in progress, which results in backup to fail.

**To Reproduce**
1. Create a vm, write data into it.
2. Take a backup.
3. while the backup is in progress, restart the vm or stop the vm.
4. System doesn't warn or stop user to restart the vm. VM gets restarted and backup fails.

**Expected behavior**
System should warn or stop user from restarting the VM when backup is in progress.

**Environment:**
 - Harvester ISO version: v1.0.0-rc2

FrankYang0529:
We need to show webhook error messages to users.

![image](https://user-images.githubusercontent.com/4927361/153380448-65989539-d002-44e8-a739-6405320d0848.png)

WuJun2016:
<img width="1674" alt="image" src="https://user-images.githubusercontent.com/24985926/156002237-79923477-6899-44d8-8a1a-e6a904753105.png">

DaiYuzeng:
@FrankYang0529  PR from UI side merged, can you please do a test?

lanfon72:
Verified this bug has been fixed.

![image](https://user-images.githubusercontent.com/5169694/156318244-f7f8e1cc-5fe9-40b4-b087-392ae67e3bf3.png)


Test Information
----
* Environment: **Bare metal DL360 3 nodes**
* Harvester Version: **master-bfd6a8a9-head**
* **ui-source** Option: **Auto**

### Verify Steps:
1. Install Harvester with any nodes
2. Login to Dashboard, Create Image for VM creation
3. Follow the _Steps to reproduce_ https://github.com/harvester/harvester/issues/1702#issue-1083933203
4. VM should NOT able to be `restart / stop` when backup is not ready
5. Error message should be displayed when user trying to `restart / stop` the VM and the backup is not ready

PR #1786, opened 2022-01-06
prevent users to stop or restart vm if there is processing backup

FrankYang0529:
**Problem:**
If users stop/restart a VM when there is a processing backup for the VM, it may fail to create the backup.

**Solution:**
* Add webhook for VM to prevent users from stopping/restarting a VM if there is a processing backup.

**Related Issue:**
https://github.com/harvester/harvester/issues/1702

**Test plan:**

Stop/Restart a VM when there is a processing backup for it:
* Create a VM.
* Create a VMBackup for it.
* Before VMBackup is done, stop/restart the VM. Verify VM can't be stopped/restarted.

### DIFF
--- a/pkg/webhook/indexeres/indexer.go
+++ b/pkg/webhook/indexeres/indexer.go
@@ -1,0 +1,22 @@
+package indexeres
+
+import (
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/webhook/clients"
+)
+
+const (
+	VMBackupBySourceUIDIndex = "harvesterhci.io/vmbackup-by-source-uid"
+)
+
+func RegisterIndexers(clients *clients.Clients) {
+	vmBackupCache := clients.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineBackup().Cache()
+	vmBackupCache.AddIndexer(VMBackupBySourceUIDIndex, vmBackupBySourceUID)
+}
+
+func vmBackupBySourceUID(obj *harvesterv1.VirtualMachineBackup) ([]string, error) {
+	if obj.Status != nil && obj.Status.SourceUID != nil {
+		return []string{string(*obj.Status.SourceUID)}, nil
+	}
+	return []string{}, nil
+}
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -12,26 +12,32 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/ref"
 	"github.com/harvester/harvester/pkg/util"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
-func NewValidator(pvcCache v1.PersistentVolumeClaimCache) types.Validator {
+func NewValidator(
+	pvcCache v1.PersistentVolumeClaimCache,
+	vmBackupCache ctlharvesterv1.VirtualMachineBackupCache,
+) types.Validator {
 	return &vmValidator{
-		pvcCache: pvcCache,
+		pvcCache:      pvcCache,
+		vmBackupCache: vmBackupCache,
 	}
 }
 
 type vmValidator struct {
 	types.DefaultValidator
-	pvcCache v1.PersistentVolumeClaimCache
+	pvcCache      v1.PersistentVolumeClaimCache
+	vmBackupCache ctlharvesterv1.VirtualMachineBackupCache
 }
 
 func (v *vmValidator) Resource() types.Resource {
 	return types.Resource{
-		Name:       "virtualmachines",
+		Names:      []string{"virtualmachines"},
 		Scope:      admissionregv1.NamespacedScope,
 		APIGroup:   kubevirtv1.SchemeGroupVersion.Group,
 		APIVersion: kubevirtv1.SchemeGroupVersion.Version,
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -17,6 +17,7 @@ import (
 	"github.com/harvester/harvester/pkg/util"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
+	webhookutil "github.com/harvester/harvester/pkg/webhook/util"
 )
 
 func NewValidator(
@@ -37,7 +38,7 @@ type vmValidator struct {
 
 func (v *vmValidator) Resource() types.Resource {
 	return types.Resource{
-		Names:      []string{"virtualmachines"},
+		Names:      []string{"virtualmachines", "virtualmachines/status"},
 		Scope:      admissionregv1.NamespacedScope,
 		APIGroup:   kubevirtv1.SchemeGroupVersion.Group,
 		APIVersion: kubevirtv1.SchemeGroupVersion.Version,
@@ -51,6 +52,9 @@ func (v *vmValidator) Resource() types.Resource {
 
 func (v *vmValidator) Create(request *types.Request, newObj runtime.Object) error {
 	vm := newObj.(*kubevirtv1.VirtualMachine)
+	if vm == nil {
+		return nil
+	}
 
 	if err := v.checkVMSpec(vm); err != nil {
 		return err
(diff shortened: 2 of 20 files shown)


pkg/webhook/resources/virtualmachineinstance/validator.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (v *vmiValidator) checkVMBackup(vmi *kubevirtv1.VirtualMachineInstance) error {
+	if vmi == nil {
+		return nil
+	}
+
+	var ownerRef *metav1.OwnerReference
+	for _, o := range vmi.OwnerReferences {
+		if o.Kind == kubevirtv1.VirtualMachineGroupVersionKind.Kind {
```
w13915984028: break loop when equal?

pkg/webhook/resources/virtualmachineinstance/validator.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	if ownerRef == nil {
+		return nil
+	}
+
+	if exist, err := util.HasInProgressingVMBackupBySourceUID(v.vmBackupCache, string(ownerRef.UID)); err != nil {
+		return werror.NewInternalError(err.Error())
+	} else if exist {
+		return werror.NewBadRequest("there is vmbackup is progressing")
+	}
```
w13915984028: vmbackup in/on processing ?

pkg/webhook/util/filter.go
```diff
@@ -0,0 +1,25 @@
...
+
+func HasInProgressingVMBackupBySourceUID(cache ctlharvesterv1.VirtualMachineBackupCache, sourceUID string) (bool, error) {
+	vmBackups, err := cache.GetByIndex(indexeres.VMBackupBySourceUIDIndex, sourceUID)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	for _, vmBackup := range vmBackups {
```
w13915984028, line 16: One reliability question:
If one VM`s backup takes long time and not finished (due to various of possible reasons), but the user is urgent to restart the VM, is it possible to delete/stop the related vm backup, then restart the VM?

The assumption is: VM stop/restart is in higher priority than VM backup, the backup needs to give way for user to end backup asap, 

at least, avoid the potential dead loop: backup is never ending (due to bug ...), stop/restart VM is also not allowed.

pkg/webhook/util/filter.go
```diff
@@ -0,0 +1,25 @@
...
+		if ctlbackup.IsBackupProgressing(vmBackup) || ctlbackup.GetVMBackupError(vmBackup) != nil {
```
w13915984028, line 17: Another point is:

Consider using the general used "Status" in VirutalMachine  instance, the "backup" action updates the state of VirtualMachine instance,   the webhook then get the info from VirtualMachine "Status" directly, knows if/not allowing the stop/restart operation.

pkg/webhook/resources/virtualmachineinstance/validator.go
FrankYang0529: Updated it. Thank you.

pkg/webhook/resources/virtualmachineinstance/validator.go
FrankYang0529: Fixed it. Thank you.

pkg/webhook/util/filter.go
FrankYang0529: @w13915984028, my thought is to let users remove error/progressing backup before stopping/restarting a VM. If VM stop/restart is in higher priority, we may need a setting for users to choose whether ignore the webhook.
@guangbochen WDYT? Thank you.

pkg/webhook/util/filter.go
FrankYang0529: Thanks for the suggestion. Since VMI is a KubeVirt resource, if we want to update it, we may need to update its annotations. In this case, we should maintain data consistency between VMBackup status and VMI annotations. I think it's better to get VMBackup information from its status.

pkg/webhook/util/filter.go
guangbochen: I think it makes sense to ask users to remove the backup first if VM restart or shutdown action is required during the backup. Also, we need to make sure not to block deleteing the backup.

pkg/webhook/indexeres/indexer.go
```diff
@@ -0,0 +1,23 @@
...
+	VMBackupBySourceUIDIndex = "harvesterhci.io/vmbackup-by-source-uid"
+)
+
+func RegisterIndexers(clients *clients.Clients) error {
+	vmBackupCache := clients.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineBackup().Cache()
+	vmBackupCache.AddIndexer(VMBackupBySourceUIDIndex, vmBackupBySourceUID)
+	return nil
+}
+
+func vmBackupBySourceUID(obj *harvesterv1.VirtualMachineBackup) ([]string, error) {
```
w13915984028, line 17: The `error ` is always nil, is it necessary to as an return param?

pkg/webhook/indexeres/indexer.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/harvester/harvester/pkg/webhook/clients"
+)
+
+const (
+	VMBackupBySourceUIDIndex = "harvesterhci.io/vmbackup-by-source-uid"
+)
+
+func RegisterIndexers(clients *clients.Clients) error {
+	vmBackupCache := clients.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineBackup().Cache()
+	vmBackupCache.AddIndexer(VMBackupBySourceUIDIndex, vmBackupBySourceUID)
```
w13915984028, line 14: could `AddIndexer ` be failed and return an error?

FrankYang0529: No, I remove the return error. Thank you.

pkg/webhook/indexeres/indexer.go
FrankYang0529: Remove it. Thank you.

FrankYang0529: Interface of the function has error param.

https://github.com/harvester/harvester/blob/c2c025a51b78788d4d6dcc8522df8077e6651de5/pkg/generated/controllers/harvesterhci.io/v1beta1/virtualmachinebackup.go#L74

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -53,11 +63,31 @@ func (v *vmValidator) Create(request *types.Request, newObj runtime.Object) erro
...
+	if oldVM == nil {
+		return nil
+	}
+
+	// Prevent users to stop/restart VM when there is VMBackup in progress.
+	// KubeVirt send stop request or change running from true to false when users stop a VM.
+	// KubeVirt send restart request (stop and start combination request) when users restart a VM.
+	// Stop reference: https://github.com/kubevirt/kubevirt/blob/c9e87c4cb6292af33ccad8faa5fb9bf269c0fbf4/pkg/virt-api/rest/subresource.go#L508-L511
+	// Restart reference: https://github.com/kubevirt/kubevirt/blob/c9e87c4cb6292af33ccad8faa5fb9bf269c0fbf4/pkg/virt-api/rest/subresource.go#L260-L263
+	if (*oldVM.Spec.Running && !*newVM.Spec.Running) ||
```
weihanglo, line 85: Out of curious. Can a start request affect the backup as well?

pkg/webhook/util/filter.go
```diff
@@ -0,0 +1,25 @@
...
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/webhook/indexeres"
+)
+
+func HasInProgressingVMBackupBySourceUID(cache ctlharvesterv1.VirtualMachineBackupCache, sourceUID string) (bool, error) {
+	vmBackups, err := cache.GetByIndex(indexeres.VMBackupBySourceUIDIndex, sourceUID)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
```
weihanglo: NIT: Can safely pass `nil` into range keyword

https://go.dev/play/p/xG9O-VqIcl7

```suggestion
	if err != nil && !apierrors.IsNotFound(err) {
```

weihanglo (approved):
Tested and overall is fine. Though there is no UI response for the error and it needs an enhancement.

image

pkg/webhook/resources/virtualmachine/validator.go
weihanglo: Tested and it won't affect any.

pkg/webhook/util/filter.go
FrankYang0529: Updated it. Thank you.

pkg/webhook/resources/virtualmachine/validator.go
FrankYang0529: For the start request, KubeVirt set running = true or add start request in `stateChangeRequest`, so we don't block it. Thank you.

https://github.com/kubevirt/kubevirt/blob/c9e87c4cb6292af33ccad8faa5fb9bf269c0fbf4/pkg/virt-api/rest/subresource.go#L439-L442

weihanglo (approved):
LGTM

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -12,26 +12,33 @@ import (
...
 	types.DefaultValidator
-	pvcCache v1.PersistentVolumeClaimCache
+	pvcCache      v1.PersistentVolumeClaimCache
+	vmBackupCache ctlharvesterv1.VirtualMachineBackupCache
 }
 
 func (v *vmValidator) Resource() types.Resource {
 	return types.Resource{
-		Name:       "virtualmachines",
+		Names:      []string{"virtualmachines", "virtualmachines/status"},
```
guangbochen, line 41: why do we need to specify the `virtualmachines/status` here?

FrankYang0529: Since `virtualmachines/status` is a subresource, if we don't add it here, we can't get the status object in the webhook server.

guangbochen (approved):
LGTM

bk201 (approved):
LGTM, works as expected. But the GUI doesn't display the error returns by the backend, is this expected?

FrankYang0529:
> LGTM, works as expected. But the GUI doesn't display the error returns by the backend, is this expected?

No, we need a PR from UI side. But we may need to merge Backend PR first, so UI can test it.

Issue #1027, opened 2021-07-19
[FEATURE] restore supports restoring to the specified namespace


WuJun2016:
**Is your feature request related to a problem? Please describe.**

Currently, restore only supports restoring to the namespace of the VM that created the backup. It is better to support restoring to any namespace that the user is able to use.
**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

### Test Plan

**Case 1: Restore VM in another namespace.**
* Create a VM `vm` in namespace `default`.
* Create a file `~/test.txt` with content `test`.
* Create a VMBackup `default-vm-backup` for it.
* Create a new namepsace `new-ns`.
* Create a VMRestore `restore-default-vm-backup-to-new-ns` in `new-ns` namespace based on the VMBackup `default-vm-backup` to create a new VM.
![Screen Shot 2022-02-17 at 5 38 31 PM](https://user-images.githubusercontent.com/4927361/154448078-ac006e07-85c9-4233-a85d-11b1f002b052.png)
* A new VM in `new-ns` namespace should be created. It should have the file `~/test.txt` with content `test`.

**Case 2: Restore VM in the original namespace.**
* Create a VM `vm` in namespace `default`.
* Create a file `~/test.txt` with content `test`.
* Create a VMBackup `default-vm-backup` for it.
* Create a VMRestore `restore-default-vm-backup-to-default-ns` in `default` namespace based on the VMBackup `default-vm-backup` to create a new VM.
* A new VM in `default` namespace should be created. It should have the file `~/test.txt` with content `test`.

**Case 3: Restore to existing VM.**
* Create a VM `vm` in namespace `default`.
* Create a file `~/test.txt` with content `test`.
* Create a VMBackup `default-vm-backup` for it.
* Update the file `~/test.txt` with content `new-test`.
* Create a VMRestore `restore-default-vm-backup` in `default` namespace based on the VMBackup `default-vm-backup` to replace original VM.
* The original VM in `default` namespace should be replaced. It should have the file `~/test.txt` with content `test`.

guangbochen:
currently, this is not supported with restoring the PVC from different namespaced volume snapshot, https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#persistentvolumeclaimspec-v1-core

FrankYang0529:
[...]

PR #530, opened 2021-03-12
Add vm restore enhancement

guangbochen:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**

Add VM backup & restore enhancement

**Solution:**
- support restore Filesystem type volumes
- add VM restore API validation
- fix s3 backup target secret is not set on the first configuration
- add config backup target annotation early
- fix VM action does not reflect on backupTarget setting configuration changes

**Related Issue:**
https://github.com/rancher/harvester/issues/385

**Test plan:**
N/A

PR #1801, opened 2022-01-12
Restore VM to specified namespace

FrankYang0529:
**Problem:**
We can only restore VM to the original namespace. We would like to restore VM to another namespace.

**Solution:**
Create VolumeSnapshot and VolumeSnapshotContent in another namespace to support the feature.

**Related Issue:**
https://github.com/harvester/harvester/issues/1027

**Test plan:**

Restore VM in another namespace.
* Create a VM `vm` in namespace `default`.
* Create a VMBackup `default-vm-backup` for it.
* Create a new namepsace `new-ns`.
* Create a VMRestore `restore-default-vm-backup-to-new-ns` in `new-ns` namespace based on the VMBackup `default-vm-backup`.
```
kind: VirtualMachineRestore
metadata:
  name: restore-default-vm-backup-to-new-ns
  namespace: new-ns
spec:
  newVM: true
  target:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: new-ns-vm
  virtualMachineBackupName: default-vm-backup
```
* A new VM in `new-ns` namespace should be created.

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -14,6 +14,7 @@ import (
 
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/rancher/wrangler/pkg/name"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -29,7 +30,10 @@ import (
 	"github.com/harvester/harvester/pkg/generated/clientset/versioned/scheme"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	ctllonghornv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta1"
+	ctlsnapshotv1 "github.com/harvester/harvester/pkg/generated/controllers/snapshot.storage.k8s.io/v1beta1"
 	"github.com/harvester/harvester/pkg/ref"
+	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
 )
 
@@ -60,15 +64,20 @@ const (
 type RestoreHandler struct {
 	context context.Context
 
-	restores          ctlharvesterv1.VirtualMachineRestoreClient
-	restoreController ctlharvesterv1.VirtualMachineRestoreController
-	backupCache       ctlharvesterv1.VirtualMachineBackupCache
-	vms               ctlkubevirtv1.VirtualMachineClient
-	vmCache           ctlkubevirtv1.VirtualMachineCache
-	pvcClient         ctlcorev1.PersistentVolumeClaimClient
-	pvcCache          ctlcorev1.PersistentVolumeClaimCache
-	secretClient      ctlcorev1.SecretClient
-	secretCache       ctlcorev1.SecretCache
+	restores             ctlharvesterv1.VirtualMachineRestoreClient
+	restoreController    ctlharvesterv1.VirtualMachineRestoreController
+	backupCache          ctlharvesterv1.VirtualMachineBackupCache
+	vms                  ctlkubevirtv1.VirtualMachineClient
+	vmCache              ctlkubevirtv1.VirtualMachineCache
+	pvcClient            ctlcorev1.PersistentVolumeClaimClient
+	pvcCache             ctlcorev1.PersistentVolumeClaimCache
+	secretClient         ctlcorev1.SecretClient
+	secretCache          ctlcorev1.SecretCache
+	snapshots            ctlsnapshotv1.VolumeSnapshotClient
+	snapshotCache        ctlsnapshotv1.VolumeSnapshotCache
+	snapshotContents     ctlsnapshotv1.VolumeSnapshotContentClient
+	snapshotContentCache ctlsnapshotv1.VolumeSnapshotContentCache
+	lhbackupCache        ctllonghornv1.BackupCache
 
 	recorder   record.EventRecorder
 	restClient *rest.RESTClient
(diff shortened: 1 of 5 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -52,20 +53,26 @@ var validateSettingFuncs = map[string]validateSettingFunc{
...
 
-func NewValidator(settingCache ctlv1beta1.SettingCache, vmBackupCache ctlv1beta1.VirtualMachineBackupCache) types.Validator {
+func NewValidator(
+	settingCache ctlv1beta1.SettingCache,
+	vmBackupCache ctlv1beta1.VirtualMachineBackupCache,
+	snapshotClassCache ctlsnapshotv1.VolumeSnapshotClassCache,
+) types.Validator {
 	validator := &settingValidator{
 		settingCache:  settingCache,
 		vmBackupCache: vmBackupCache,
```
weihanglo: Forgot adding this?
```suggestion
		vmBackupCache: vmBackupCache,
		snapshotClassCache: snapshotClassCache,
```

pkg/controller/master/backup/restore.go
```diff
@@ -148,6 +166,35 @@ func (h *RestoreHandler) RestoreOnChanged(key string, restore *harvesterv1.Virtu
 	return nil, h.updateStatus(restore, backup, vm, isVolumesReady)
 }
 
+// RestoreOnRemove delete VolumeSnapshotContent which is created by restore controller
+// Since we would like to prevent LH Backups from being removed when users delete the VM,
+// we use Retain policy in VolumeSnapshotContent.
+// We need to delete VolumeSnapshotContent by restore controller, or there will have remaining VolumeSnapshotContent in the system.
+func (h *RestoreHandler) RestoreOnRemove(key string, restore *harvesterv1.VirtualMachineRestore) (*harvesterv1.VirtualMachineRestore, error) {
+	if restore.Status == nil {
```
weihanglo: Should it check the existence of restore as well? Most of `OnRemove` callbacks in Harvester did that.
```suggestion
	if restore == nil || restore.Status == nil {
```

pkg/webhook/resources/setting/validator.go
FrankYang0529: Added it. Thank you.

pkg/controller/master/backup/restore.go

FrankYang0529:
@guangbochen, do we need to handle permission issues in the controller? For example, restrict users to restore a VM to another namespace which they don't have permission.

guangbochen:
> @guangbochen, do we need to handle permission issues in the controller? For example, restrict users to restore a VM to another namespace which they don't have permission.

We can enable the permission check on the webhook when users restore a new VM on different namespaces. 
Nit: u can use [SelfSubjectAccessReview](https://kubernetes.io/docs/reference/access-authn-authz/authorization/) to validate user with RBAC permission.

weihanglo:
@guangbochen Is there any interaction between Rancher user and K8s built-in role that we need to handle? I am uncertain that built-in RBAC permission is sufficient but it also somehow weird dealing with Rancher role in the webook.

guangbochen:
@weihanglo Rancher users also leverage the native k8s RBAC permission with aggregated clusterRole/Roles, and in the webhook, we only deal with the k8s RBAC,e.g validate if the user has read/write permission to the backup volumesnapshot by namespace.

pkg/controller/master/backup/restore.go
```diff
@@ -513,54 +560,160 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+		return volumeSnapshotContent, nil
 	}
 
-	if pvc.Annotations == nil {
-		pvc.Annotations = make(map[string]string)
+	lhBackup, err := h.lhbackupCache.Get(util.LonghornSystemNamespaceName, *volumeBackup.LonghornBackupName)
+	if err != nil {
+		return nil, err
 	}
+	snapshotHandle := fmt.Sprintf("bs://%s/%s", volumeBackup.PersistentVolumeClaim.ObjectMeta.Name, lhBackup.Name)
```
weihanglo, line 652: NIT: Could we add a reference for this magic URL?

```suggestion
	// Ref: https://longhorn.io/docs/1.2.3/snapshots-and-backups/csi-snapshot-support/restore-a-backup-via-csi/#restore-a-backup-that-has-no-associated-volumesnapshot
	snapshotHandle := fmt.Sprintf("bs://%s/%s", volumeBackup.PersistentVolumeClaim.ObjectMeta.Name, lhBackup.Name)
```

pkg/controller/master/backup/restore.go
```diff
@@ -513,54 +560,160 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+		if err != nil {
+			return err
+		}
+		dataSourceName = volumeSnapshot.Name
+	}
+
+	annotations := map[string]string{}
+	for key, value := range volumeBackup.PersistentVolumeClaim.ObjectMeta.Annotations {
+		needSkip := false
+		for _, prefix := range restoreAnnotationsToDelete {
```
weihanglo, line 591: Out of curious. What is the purpose of `restoreAnnotationsToDelete`.

weihanglo (commented):
It works and the code looks good!
Do we aim to resolve permission issue in this PR?

pkg/controller/master/backup/restore.go
FrankYang0529: Added it. Thank you.

pkg/controller/master/backup/restore.go
FrankYang0529: Not sure why we need to delete some annotations. I just try to refactor how we build annotations map structure. It comes from https://github.com/harvester/harvester/pull/530.

weihanglo (approved):
I'm fine with the current patch. We can address permission issue in follow-ups.

pkg/controller/master/backup/restore.go
```diff
@@ -148,6 +166,35 @@ func (h *RestoreHandler) RestoreOnChanged(key string, restore *harvesterv1.Virtu
...
+	}
+
+	backup, err := h.getVMBackup(restore)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, volumeBackup := range backup.Status.VolumeBackups {
+		if volumeSnapshotContent, err := h.snapshotContentCache.Get(h.constructVolumeSnapshotContentName(restore.Namespace, restore.Name, *volumeBackup.Name)); err != nil {
+			if !apierrors.IsNotFound(err) {
```
guangbochen: prefer to check with `err != nil && !apierrors.IsNotFound(err)` first.

FrankYang0529: Fixed it. Thank you.

guangbochen (approved):
LGTM

PR #1802, opened 2022-01-12
add webhook for support-bundle-image

FrankYang0529:
**Problem:**
We didn't validate support-bundle-image setting value.

**Solution:**
Add webhook for it.

**Related Issue:**
https://github.com/harvester/harvester/issues/1727

**Test plan:**
1. Empty value should pass.
2. Non-JSON value should fail.
3. Value with any blank field (repository, tag, or imagePullPolicy) should fail.
4. `'{"repository": "rancher/support-bundle-kit","tag":"v0.0.5","imagePullPolicy":"IfNotPresent"}'` should pass.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -45,6 +45,7 @@ type validateSettingFunc func(setting *v1beta1.Setting) error
 var validateSettingFuncs = map[string]validateSettingFunc{
 	settings.HttpProxySettingName:            validateHTTPProxy,
 	settings.VMForceResetPolicySettingName:   validateVMForceResetPolicy,
+	settings.SupportBundleImageName:          validateSupportBundleImage,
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Tested. LGTM 🎉

guangbochen (approved):
LGTM

w13915984028:
Question: 
when repository is invalid, e.g. tag v.1.1.1 is not existing, should the validator check the existing of it?

'{"repository": "rancher/support-bundle-kit","tag":"v0.0.5","imagePullPolicy":"IfNotPresent"}'

FrankYang0529:
> Question: when repository is invalid, e.g. tag v.1.1.1 is not existing, should the validator check the existing of it?
> 
> '{"repository": "rancher/support-bundle-kit","tag":"v0.0.5","imagePullPolicy":"IfNotPresent"}'

Not sure whether we need to validate whether the image is existent. If yes, we may need to run `crictl` to check it. @guangbochen, do you think we have to add this? Thank you

Issue #1782, opened 2022-01-06
[FEATURE] Optimize the way to configure the static VIP


yaocw2020:
**Is your feature request related to a problem? Please describe.**

Related with https://github.com/harvester/harvester/issues/1772, if the kube-vip supports allocating a static VIP address to service by setting the spec.loadBalancerIP field in the service manifest, we don't have to set it in the configmap anymore.

**Describe the solution you'd like**

Set the harvester cluster VIP in the spec.loadBalancerIP field of the service `kube-system/ingress-expose`, ranther than saving in the configmap and allocating by the kube-vip cloud provider.

**Describe alternatives you've considered**

**Additional context**

yaocw2020:
For testing:
Use the master-head to spin up the Harvester and the VIP will be normal.

TachunLin:
Verified **fixed** on master-9091d7d0-head (02/14) . Wait for #1772 verification status to close this issue.

### Result
Create a 3 nodes harvester with static VIP in the following scenarios, 
the harvester cluster can work as normal and access dashboard via VIP

1. Bare metal machine with static VIP on pxe install
2. Local kvm machine with static VIP on ipxe vagrant example

### Test Information
* Test Environment1: 3 nodes harvester on provo bare machine
* Test Environment2: 3 nodes harvester on local kvm
* Harvester version: master-9091d7d0-head (02/14)

### Verify Steps
1. Use static VIP to setup a 3 nodes harvester cluster on bare machine
2. Check harvester cluster work as normal, 
3. Check can access dashboard via the static VIP
4. Use ipxe example to setup a 3 nodes harvester cluster on local kvm machine
5. Check harvester cluster work as normal, 
6. Check can access dashboard via the static VIP

### Additional Context

yaocw2020:
If this issue is verified successfully, the issue [1772](https://github.com/harvester/harvester/issues/1772) could be closed as well because this issue is a real case of issue 1772. In other words, this optimization must not be completed if issue 1772 is still not fixed. @TachunLin

TachunLin:
Close this issue since the related issuse #1772 have been verified and close.

Issue #1772, opened 2022-01-04
[BUG] Static VIP address (spec.loadBalancerIP field) does not work


janeczku:
**Describe the bug**

Kube-VIP supports allocating a static VIP address to a service by setting the `spec.loadBalancerIP` field in the service manifest. 

Ref: https://kube-vip.io/usage/on-prem/#the-kube-vip-cloud-provider-configmap

When creating such service manifest in a Harvester guest cluster, then instead of assigning the address in `spec.loadBalancerIP` field, a random IP address from the pool is assigned.

**To Reproduce**

In a guest cluster, create a service resource with a static loadBalancerIP address:

```
apiVersion: v1
kind: Service
metadata:
  name: nginx
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
  selector:
    app: nginx
  type: LoadBalancer
  loadBalancerIP: "10.65.100.10"
```

yaocw2020:
It works well with the kube-vip of the 0.4 version. We have to upgrade the kube-vip to the latest version.

yaocw2020:
For testing
- Set up a harvester
- Create a deployment and related load balancer type service
- Specify the field `spec.loadBalancerIP` with a static IP
- Request the svc through the load balancer IP

noahgildersleeve:
Tested in Harvester `master-1a861fc1-head` and Rancher `da382a8`. with the test steps 



> For testing
> 
> * Set up a harvester
> * Create a deployment and related load balancer type service
> * Specify the field `spec.loadBalancerIP` with a static IP
> * Request the svc through the load balancer IP

I creted a deployment in Rancher with nginx, then created a load balancer service and specified the load balancer IP. It isn't routing it's still going to the endpoint that's in the YAML. The load balancer IP doesn't appear to be routing. Also the static IP I was trying to assign was in the same subnet as the one it randomly assigned.

TachunLin:
Verified **fixed** on `master-b09e3233-head`(02/14). Close this issue.

### Result
In Harvester cluster, we can create nginx service with static IP set on spec.loadBalancerIP. 
And can reach the services through the static IP  

```
harvester-one-220214:/home/rancher # kubectl describe service nginx
Name:                     nginx
Namespace:                default
Labels:                   <none>
Annotations:              <none>
Selector:                 app=nginx
Type:                     LoadBalancer
IP Family Policy:         SingleStack
IP Families:              IPv4
IP:                       10.53.137.109
[...]

Issue #1681, opened 2021-12-15
[BUG] Power on 3 harvester nodes in sequence may cause VIP connectivity lost


TachunLin:
**Describe the bug**

Power on 3 harvester nodes in sequence have possibility failed to access VIP
Need to wait for long time (in this case, wait 20~30 min) to be able to recover connection 

```
hpd8s7:/home/rancher # ping 10.84.44.181
PING 10.84.44.181 (10.84.44.181) 56(84) bytes of data.
From 10.84.44.114 icmp_seq=1 Destination Host Unreachable
From 10.84.44.114 icmp_seq=2 Destination Host Unreachable
From 10.84.44.114 icmp_seq=3 Destination Host Unreachable
```

**To Reproduce**
1. Prepare 3 nodes harvester
2. Enable virtual network with `harvester-mgmt`
3. create vlan network  `vlan 806` with id `806`
4. create several vm
5. Provision a RKE2 cluster 
6. Press 'Momentary Press` button to shutdown node 3
7. Press 'Momentary Press` button to shutdown node 2
8. Press 'Momentary Press` button to shutdown node 1
9. Wait for 20 minutes
10. Press 'Momentary Press` button to power on node 1
11. Press 'Momentary Press` button to power on node 2
12. Press 'Momentary Press` button to power on node 3
13. Wait for startup complete
14. Check node and VIP accessibility

**Expected behavior**

Should be able to access VIP 

**Support bundle**

[supportbundle_29976b55-cbd5-4e7d-a7a2-9c883c9b375c_2021-12-15T09-47-35Z.zip](https://github.com/harvester/harvester/files/7718167/supportbundle_29976b55-cbd5-4e7d-a7a2-9c883c9b375c_2021-12-15T09-47-35Z.zip)

**Environment:**
 - Harvester ISO version: master-5d6f3db0-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 nodes harvester on provo bare machines

**Server information:**
 - VIP: 10.84.44.181
 - node 1: 10.84.44.114
 - node 2: 10.84.44.115
 - node 3: 10.84.44.116

**Additional context**
Not always happen. There are two ways to solve if encounter this issue
1. Wait for a while, the kube-vip pod may recover the VIP connectivity
2. Kill kube-vip pod

yaocw2020:
Log from kube-vip
```
[...]

Issue #1521, opened 2021-11-08
[BUG] Agent node becomes NotReady when one of the controller node fails


bk201:
**Describe the bug**

The HA doesn't seem to work well. When a controller node fails, agent nodes might fail too.
VMs on the agent nodes might become Terminating.

**To Reproduce**
**NOTE** this doesn't always repro.

1. Create a 4-node cluster.
2. 3 nodes will become controller nodes.
```
# kubectl get nodes
kubectl get nodes -o wide
NAME    STATUS   ROLES                       AGE     VERSION          INTERNAL-IP   EXTERNAL-IP   OS-IMAGE               KERNEL-VERSION         CONTAINER-RUNTIME
node1   Ready    control-plane,etcd,master   7d11h   v1.21.5+rke2r1   10.10.0.11    <none>         v20211007 (04ec006)   5.3.18-59.24-default   containerd://1.4.9-k3s1
node2   Ready    <none>                      7d11h   v1.21.5+rke2r1   10.10.0.12    <none>         v20211007 (04ec006)   5.3.18-59.24-default   containerd://1.4.9-k3s1
node3   Ready    control-plane,etcd,master   7d11h   v1.21.5+rke2r1   10.10.0.13    <none>         v20211007 (04ec006)   5.3.18-59.24-default   containerd://1.4.9-k3s1
node4   Ready    control-plane,etcd,master   7d11h   v1.21.5+rke2r1   10.10.0.14    <none>         v20211007 (04ec006)   5.3.18-59.24-default   containerd://1.4.9-k3s1
```
3. Log in to the agent node, check `/var/lib/rancher/rke2/agent/etc/rke2-agent-load-balancer.json`
```
# on node2
# cat /var/lib/rancher/rke2/agent/etc/rke2-agent-load-balancer.json
{
  "ServerURL": "https://10.10.0.13:9345",
  "ServerAddresses": [
    "10.10.0.11:9345",
    "10.10.0.13:9345",
    "10.10.0.14:9345"
  ],
  "Listener": null
}
```
4. Stop or shut down the controller node whose IP address is `10.10.0.13`.
5. After the controller becomes `NotReady`, the agent node becomes `NotReady` too.
```
kubectl get nodes
NAME    STATUS     ROLES                       AGE     VERSION
node1   Ready      control-plane,etcd,master   7d11h   v1.21.5+rke2r1
node2   NotReady   <none>                      7d11h   v1.21.5+rke2r1
[...]

PR #1805, opened 2022-01-12
Bump kube-vip

yaocw2020:
Solution:
- The kube-vip could use the network interface based on the default route automatically. If the default routes changes, it will be panic and restart as a daemonset pod. Thus, we don't need a VIP controller in the harvester network controller to change the VIP interface of the kube-vip anymore. The advance is architecture decoupling to solve the problems like [1681](https://github.com/harvester/harvester/issues/1681)
- The kube-vip v0.4.1 supports static VIP configured in the field `.spec.loadBalancerIP` of the service.

Related issue: 
https://github.com/harvester/harvester/issues/1772
https://github.com/harvester/harvester/issues/1782
https://github.com/harvester/harvester/issues/1681

- Bump kube-vip v0.4.1 
- Change the way to configure static VIP


gitlawr:
How does upgrade work with this change?

yaocw2020:
> How does upgrade work with this change?

After discussion with @bk201, the upgrade component should add something to support the dependency chart parameters update in the harvester managedChart.

The related [issue](https://github.com/harvester/harvester/issues/1810) about upgrading.

deploy/charts/harvester/values.yaml
```diff
@@ -429,9 +429,9 @@ kube-vip:
   enabled: false
   image:
     repository: plndr/kube-vip
-    tag: v0.3.7
-  config:
-    vip_interface: eth0
+    tag: main
+  nodeSelector:
+    node-role.kubernetes.io/control-plane: "true"
```
guangbochen, line 434: why should we only deploy the `kubevip` pod to the control-plane node only?

bk201: Just want to add we [suggested users join nodes with VIP for high availability](https://github.com/harvester/harvester/issues/1521).
HA in RKE2 doesn't work if VIP is on an agent node.

guangbochen (approved):
LGTM, and will need QA to validate the upgrade issue tracked on the  https://github.com/harvester/harvester/issues/1810

gitlawr (approved):
LGTM

Issue #1728, opened 2021-12-21
[FEATURE] Allow Longhorn to use another disk other than the root disk during installation


janeczku:
**Is your feature request related to a problem? Please describe.**

In many hardware setups involving multiple disks per server a dedicated physical disk will be used for Harvester installation.
It should be possible to exclude that disc from being used to store VMs for several reasons:

- hard storage IO separation between harvester system and VM workloads
- root install disk may have a different performance than the disks dedicated for VM storage (e.g. SSD vs NVMe)

**Expected UX**

There is a global setting to disable Longhorn default storage path creation on the install disc.

guangbochen:
By default, `/var/lib/longhorn` on the host will be used for storing the volume data. We can avoid using the default directory by adding a new disk, then disable scheduling for `/var/lib/longhorn` of both Harvester installer and PXE installation.

johnliu55tw:
@guangbochen I'm actually thinking of a simpler way to achieve this: **Mount the additional disk (partition) to `/var/lib/longhorn` so it would be the default storage**. This has a few benefits:

- Still works when there's only one Harvester node
    - Longhorn won't decommission the default directory (`/var/lib/longhorn`) after disabling scheduling if there are not enough replicas. With only one node, it's almost not possible for Longhorn to create enough replicas. Of course, we could force Longhorn to remove the default directory once the new disk is added, but it would increase the complexity.

- No need to wait for NDM to provision the additional disk
    - NDM would require some time to provision the additional disk. Users might have to wait a while to see all data being migrated to the additional disks.

guangbochen:
@johnliu55tw Longhorn supports [default disk customization](https://longhorn.io/docs/1.2.3/advanced-resources/default-disk-and-node-config/#customizing-default-disks-for-new-nodes) that we can leverage to customize the default data path of the node. Also, I'm fine to use the default path and mount it to additional disks.

johnliu55tw:
**This design is obsoleted. See [HEP Select another disk to store VM data during installation](https://github.com/harvester/harvester/blob/master/enhancements/20220120-select-another-disk-to-store-vm-data-during-installation.md)**
# Brief Design (Obsolete)
## User Interface
- Add a new installer config `install.vm_data_disk` to specify the disk to be used as the default VM storage (Longhorn) disk.
```
install:
  vm_data_disk: /dev/sda
``` 

- Add config `install.vm_data_disk` to Installer UI to let users select one of the detected storage devices.

## System
[...]

PR #1832, opened 2022-01-18
Add the ability to apply resources during upgrade

johnliu55tw:

**Problem:**
Sometimes during the upgrade process, we need to modify some resources that are not managed by the Helm chart, such as custom resources.

For instance, we need to update the Longhorn setting `default-data-path` from the default to `/var/lib/harvester/defaultdisk` for feature https://github.com/harvester/harvester/issues/1728, so that Nodes joining the cluster would have `/var/lib/harvester/defaultdisk` as the default disk.

**Solution:**
Packaging the resources into the `harvester-upgrade` container image, then during the upgrade process, deploy these resources by `kubectl apply`.

**Related Issue:**
- https://github.com/harvester/harvester/issues/1728
- https://github.com/harvester/harvester-installer/pull/217

**Test plan:**
1. Create a 1.0.0 cluster
2. Upgrade it using `master` branch code
3. Verify Longhorn setting `default-data-disk` has been updated to `/var/lib/harvester/defaultdisk`
    ```
    # kubectl get -n longhorn-system setting.longhorn.io default-data-path
    NAME                VALUE                            AGE
    default-data-path   /var/lib/harvester/defaultdisk   62m
    ```

**Dependencies**
- Need to wait for https://github.com/harvester/harvester-installer/pull/217

### DIFF
--- a/package/upgrade/extra_manifests/longhorn-settings-default-data-path.yaml
+++ b/package/upgrade/extra_manifests/longhorn-settings-default-data-path.yaml
@@ -1,0 +1,6 @@
+apiVersion: longhorn.io/v1beta1
+kind: Setting
+metadata:
+  name: default-data-path
+  namespace: longhorn-system
+value: /var/lib/harvester/defaultdisk
--- a/package/upgrade/Dockerfile
+++ b/package/upgrade/Dockerfile
@@ -18,3 +18,4 @@ RUN curl -sfL https://github.com/kubevirt/kubevirt/releases/download/v0.45.0/vir
 COPY upgrade_node.sh /usr/local/bin/
 COPY upgrade_manifests.sh /usr/local/bin/
 COPY lib.sh /usr/local/bin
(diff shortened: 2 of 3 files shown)


package/upgrade/extra_manifests/longhorn-settings-default-data-path.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: longhorn.io/v1beta1
+kind: Setting
+metadata:
+  name: default-data-path
+  namespace: longhorn-system
+value: /var/lib/longhorn_data
```
bk201: Should this be adjusted to `/var/lib/harvester/defaultdisk` as well?

johnliu55tw: Yes we should, thanks!

guangbochen (approved):
LGTM

bk201 (approved):
LGTM. Nice add-on.

Issue #1703, opened 2021-12-18
[BUG] Backup shows status as 'progressing' even when the backup is failed


khushboo-rancher:
**Describe the bug**
Backup shows status as 'progressing' even fail the backup is failed. 

**To Reproduce**
1. Take a backup of vm.
2. While the backup is in progress, stop the vm.
3. Backup status doesn't show failed, it keep showing 'progressing'

**Expected behavior**
Backup should show the correct status.

**Environment:**
 - Harvester ISO version:  v1.0.0-rc2

WuJun2016:
The error message should be displayed in `metadata.state.Error` should not be displayed in `status.error` inside


```
apiVersion: harvesterhci.io/v1beta1
kind: VirtualMachineBackup
metadata:
  creationTimestamp: "2022-01-12T02:52:30Z"
  finalizers:
  - wrangler.cattle.io/harvester-vm-backup-controller
  generation: 7
  managedFields:
  - apiVersion: harvesterhci.io/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .: {}
          v:"wrangler.cattle.io/harvester-vm-backup-controller": {}
      f:spec:
        .: {}
        f:source:
          .: {}
          f:apiGroup: {}
          f:kind: {}
          f:name: {}
      f:status:
        .: {}
        f:backupTarget:
          .: {}
          f:bucketName: {}
          f:bucketRegion: {}
          f:endpoint: {}
        f:conditions: {}
        f:error:
          .: {}
          f:message: {}
          f:time: {}
        f:readyToUse: {}
        f:secretBackups: {}
        f:source:
          .: {}
          f:metadata:
            .: {}
            f:annotations: {}
            f:creationTimestamp: {}
            f:labels: {}
            f:name: {}
            f:namespace: {}
          f:spec:
            .: {}
            f:running: {}
            f:template:
              .: {}
              f:metadata:
                .: {}
                f:annotations: {}
                f:creationTimestamp: {}
                f:labels: {}
              f:spec:
                .: {}
                f:domain:
                  .: {}
                  f:cpu:
                    .: {}
                    f:cores: {}
                    f:sockets: {}
                    f:threads: {}
                  f:devices:
                    .: {}
                    f:disks: {}
                    f:inputs: {}
                    f:interfaces: {}
                  f:machine:
                    .: {}
                    f:type: {}
                  f:memory:
                    .: {}
                    f:guest: {}
[...]

PR #1843, opened 2022-01-19
fix error message in condition to show it in UI

FrankYang0529:
**Problem:**
Dashboard can't show error state because controller doesn't set conditions correctly.

**Solution:**
If there is error message, we set `condition.Reason` as `Error`.

**Related Issue:**
https://github.com/harvester/harvester/issues/1703

**Test plan:**
* Take a backup of vm.
* While the backup is in progress, stop the vm.
* Dashboard should show backup failed and its error message.


weihanglo (approved):
LGTM

image

guangbochen (approved):
LGTM

PR #1848, opened 2022-01-20
HEP: Select another disk to store VM data during installation

johnliu55tw:
https://github.com/harvester/harvester/issues/1728

### DIFF
--- a/enhancements/20220120-select-another-disk-to-store-vm-data-during-installation.md
+++ b/enhancements/20220120-select-another-disk-to-store-vm-data-during-installation.md
@@ -79,7 +79,7 @@ No API changes.
 ### Implementation Overview
 
 #### Fresh install
-- Change Default Data Path to `/var/lib/longhorn_data`.
+- Change Default Data Path to `/var/lib/harvester/defaultdisk`.
   It means Longhorn will use this path to set up the Default Longhorn disk other than the default
   `/var/lib/longhorn`, where the Longhorn binaries locate.
   We can modify the Longhorn chart values to deploy Longhorn with this setting.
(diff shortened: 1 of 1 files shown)


enhancements/20220120-select-another-disk-to-store-vm-data-during-installation.md
```diff
@@ -0,0 +1,190 @@
...
+  see [here](https://longhorn.io/docs/1.2.3/references/settings/#default-data-path)
+  for more information.
+
+- **VM disk**:
+  The disk users selected for storing VM data.
+
+### Implementation Overview
+
+#### Fresh install
+- Change Default Data Path to `/var/lib/longhorn_data`.
```
bk201: NDM setup mount points for new-added disk in `/var/lib/harvester/extra-disks/<block-id>`. How about we also use the path for this data disk. Or, maybe something like `/var/lib/harvester/default-disk/`.

The proposal looks good.

bk201 (approved):

enhancements/20220120-select-another-disk-to-store-vm-data-during-installation.md
johnliu55tw: Good suggestion, I think `/var/lib/harvester/default-disk` is a better choice.

The somewhat related topic here is maybe we should avoid using `-` character in the persistent state path. cOS would replaces the `/` in the path with `-` and creates the "real" persistent directory in `/usr/local/.state/<real persistent dir>`, then bind-mount this real persistent dir onto the path. For example, cOS would create a directory `/usr/local/.state/var-lib-foo-bar.bind/` for both `/var/lib/foo/bar/` and `/var/lib/foo-bar`, then bind-mount this directory onto both of them.

Generally, this should be fine as long as we don't have such conflicted persistent state paths specified in `/run/cos/cos-layout.env`, but I still think we should avoid this. Therefore, I'd say we should use path like `/var/lib/harvester/defaultdisk`

enhancements/20220120-select-another-disk-to-store-vm-data-during-installation.md
```diff
@@ -0,0 +1,190 @@
...
+It's possible to add labels, but not annotations.
+
+Luckily, the labels and annotations could be added to the Node after it has joined the cluster.
+Longhorn would follow the annotation to create default disks once it appears,
+**as long as default disks have never been created on the Node.**
+Therefore, we should be able to rely on NDM or other controllers to
+help add the labels and annotations after it joined the cluster, but it requires extra effort.
+
+### Version incompatibility
+We should prohibit users from adding v1.0.0 Nodes into the cluster with this enhancement.
```
yasker, line 176: Very good catch. We should add `require/release-note` to the original GitHub issue to call it out in the release note.

yasker (approved):
Very well written. LGTM.

johnliu55tw:
I've created the issue for the version incompatibility check: https://github.com/harvester/harvester/issues/1860

guangbochen (approved):
LGTM, for HEP docs we can skip the CI via https://docs.drone.io/pipeline/skipping/

Issue #1718, opened 2021-12-20
[BUG] Automatic disk provisioning result in unusable ghost disks on NVMe drives


janeczku:
**Steps**

1. Install Harvester 1.0.0-rc2 on hardware servers with multiple NVMe disks
2. Add all the disks to Harvester using the global setting:

```
system_settings:
  auto-disk-provision-paths: /dev/nvme*
```

**Issue**

1. All the internal nvme drives seem to be active, but there are some ghost disks in the ui, that don't show up on the system or in blockdevices.harvesterhci.io
2. Even though all nodes have identical disks, the UI shows different amount of storage for each node

**Extra Info**

```
sandboxc1n10:~ # mount|grep extra-disk
/dev/nvme0n1p1 on /var/lib/harvester/extra-disks/8b964ba6f099f9c37ad67f7e0dcfbc88 type ext4 (rw,relatime,errors=remount-ro)
/dev/nvme1n1p1 on /var/lib/harvester/extra-disks/9a2271cb4fe2209b47d01bc330294418 type ext4 (rw,relatime,errors=remount-ro)
/dev/nvme2n1p1 on /var/lib/harvester/extra-disks/12df0e8f129b4bac41cb464be4f18c61 type ext4 (rw,relatime,errors=remount-ro)
/dev/nvme3n1p1 on /var/lib/harvester/extra-disks/c0e69dd92a250119fc1acf9ee386587f type ext4 (rw,relatime,errors=remount-ro)
/dev/nvme5n1p1 on /var/lib/harvester/extra-disks/c3c3365dfe86e88b61d0fa0a6bb275fa type ext4 (rw,relatime,errors=remount-ro)
/dev/nvme4n1p1 on /var/lib/harvester/extra-disks/7f95ff328dda0610d5386f0e9b33447b type ext4 (rw,relatime,errors=remount-ro)
sandboxc1n10:~ # blkid|grep nvme
/dev/nvme3n1p1: UUID="24bd3f55-268b-4bbd-8f3e-7775e6c29609" BLOCK_SIZE="4096" TYPE="ext4" PARTLABEL="primary" PARTUUID="df093500-30e2-4537-9a76-9d924a1f429a"
/dev/nvme2n1p1: UUID="2ce9698f-2ffb-463c-8851-329f9e504b9d" BLOCK_SIZE="4096" TYPE="ext4" PARTLABEL="primary" PARTUUID="09e97962-91da-4495-af08-c1c939b0c0e6"
[...]

PR #1866, opened 2022-01-25
[WIP] HEP: Concurrency Control on Host Disk Operations

weihanglo:
[Rendered](https://github.com/weihanglo/harvester/blob/395bec9745319874adc92fa5071bba90b9e85bfb/enhancements/20220125-concurrency-control-on-host-disk-operations.md)

**Problem:**
#1718
harvester/node-disk-manager#23

### DIFF
--- a/enhancements/20220125-concurrency-control-on-host-disk-operations.md
+++ b/enhancements/20220125-concurrency-control-on-host-disk-operations.md
@@ -1,0 +1,271 @@
+# Concurrency Control on Host Disk Operations
+
+## Summary
+
+Some operations, such as formatting, might take a while for operating system to
+finish. It causes problems if the controller of host disk manager re-enters
+these criticial sections. To prevent re-entrance, additional disk management of
+Harvester needs a robust concurrency control on host disk operations.
+
+### Related Issues
+
+- Original bug report: https://github.com/harvester/harvester/issues/1718
+- Previous PR: https://github.com/harvester/node-disk-manager/pull/23
+
+## Motivation
+
+### Goals
+
+- Prevent criticial disk operations from re-entrances.
+- Control maximum count of concurrent disk operations within a time period.
+- Refactor the logic of blockdevice reconcilation to a more deterministic,
+  readable, and maintainable state.
+
+### Non-goals
+
+- Auto-recovering from disk operation failures is not in the scope. Node disk
+  manager does not know how to deal with this kind of failures (might due to 
+  hardware or OS failures). User should come in and tackle the issue manually
+  before re-triggering any futher operation.
+
+## Proposal
+
+This proposal is originally for tackling [issue #1718]. During the investigation
+of that issue, we found several potential flaws in the current implementation:
+
+1. After introduced [feature auto-provision], user is able to format and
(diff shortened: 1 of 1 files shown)


enhancements/20220125-concurrency-control-on-host-disk-operations.md
```diff
@@ -0,0 +1,271 @@
...
+  hardware or OS failures). User should come in and tackle the issue manually
+  before re-triggering any futher operation.
+
+## Proposal
+
+This proposal is originally for tackling [issue #1718]. During the investigation
+of that issue, we found several potential flaws in the current implementation:
+
+1. After introduced [feature auto-provision], user is able to format and
+   provision a bunch of raw block device at once. And that leads to several
```
yasker, line 37: I don't understand why race conditions can happen here. Is each block device a separate CR? Kubernetes controller makes sure at any given time only one CR of the same name will be handled in the same process (which I assume will be one per node since block devices are local to the node). The queue is a built-in part of Kubernetes controller pattern.

enhancements/20220125-concurrency-control-on-host-disk-operations.md
```diff
@@ -0,0 +1,271 @@
...
+This proposal is originally for tackling [issue #1718]. During the investigation
+of that issue, we found several potential flaws in the current implementation:
+
+1. After introduced [feature auto-provision], user is able to format and
+   provision a bunch of raw block device at once. And that leads to several 
+   race conditions between each format or provision operation, which might fail
+   eventually and make disks result in a wrong state.
+2. Operations like formatting a disk may take time. When `OnBlockDeviceChange`
+   has been called the first time and the disk started formatting, and later it
+   was called the second time but the former formatting has not yet finished,
```
yasker, line 42: There are two possibilities here:
1. The first format is in progress then the second one retried, which should be 100% prevented from the controller side.
2. The first format failed or succeed, then the second time the object was requeued. We only need to check if the device is formatted to know it's status. If it's not formated when we're handling the disk, then we format it. We won't format the same disk twice at the same time due to controller limits the access to the object.

One very common issue would be, the format operation itself succeed, but the update to the object failed due to update conflict, so the CR will be requeued. Even in this case, we don't need to redo the format in this case if we detect the device has already been formatted.

enhancements/20220125-concurrency-control-on-host-disk-operations.md
```diff
@@ -0,0 +1,271 @@
...
+device directly, so users previous depends on these functionalities will find
+they cannot after this enhancement. Thankfully, these changes won't affect any
+real user experience on disk provisions, since node disk manager won't persist
+mounts for extra disks between boots, and it always re-mounts all extra disk
+after reboot.
+
+### API changes
+
+- :pencil2: Add several new phases to `blockdevice.status.provisionPhase`
+    - :new: `Formatting`
```
yasker, line 97: Don't use `ing` state as indication what's happened. Check the disk status to see if it's formatted/mounted as expected instead.

yasker (changes requested):

weihanglo:
Close in favor of simpler fix https://github.com/harvester/node-disk-manager/pull/27

Issue #1864, opened 2022-01-25
[FEATURE] Support bundle for a single node (Live/Installed)


bk201:
**Is your feature request related to a problem? Please describe.**

It's hard to debug a failed installation or un-provisioned node without asking for much info from users.
e.g., https://github.com/harvester/harvester/issues/1863

**Describe the solution you'd like**

It would be nice to add a command to generate a node support bundle for a node that is in live mode or installed mode.

We can leverage or extend node collector https://github.com/rancher/support-bundle-kit/blob/master/hack/collector-harvester. Another idea is to use OpenSUSE supportutils (supportconfig).

**Describe alternatives you've considered**

**Additional context**

irishgordo:
@bk201 I've had a chance to validate this:

## The Setup:
- VM in virt-manager, 8 core, 18GiB RAM, 200GiB disk

## The ISO Used For Install Test's `md5sum`:
```
4fe096476e1205c05096f4be70cb0db8  harvester-master-amd64.iso
```

## Results
Following the steps I was able to replicate the test strategy and validate that the support config gets generated and provides an available path.
I validated that the `.txz` can be extracted with `tar xvf`. 


![Screenshot from 2022-04-27 15-33-00](https://user-images.githubusercontent.com/5370752/165647582-d487b4ec-dcf1-451d-a063-6502ebd97012.png)
![Screenshot from 2022-04-27 15-02-39](https://user-images.githubusercontent.com/5370752/165647594-b3529472-e606-4d12-8953-54ff5d86c00e.png)

This looks good to me :smile:

irishgordo:
Opened PR to update manual test doc.
Will close this ticket out in the meantime

PR #1881, opened 2022-01-28
Add single node support bundle HEP

bk201:
[Rendered](https://github.com/harvester/harvester/blob/46e5f755d497115dc2701dda2ebc07ffb7e7a92c/enhancements/20220128-single-node-support-bundle.md)

**Related Issue:**
https://github.com/harvester/harvester/issues/1864

**sample bundle**
[scc_node1_220128_0755.zip](https://github.com/harvester/harvester/files/7956831/scc_node1_220128_0755.zip)

The bundle is a tarball created by the command: `supportconfig -x NET -k -c`. And the `harvester` plugin is included (check  `plugin-harvester.txt` and `harvester` directory).

### DIFF
--- a/enhancements/20220128-single-node-support-bundle.md
+++ b/enhancements/20220128-single-node-support-bundle.md
@@ -1,0 +1,167 @@
+# Add single node support bundle feature
+
+## Summary
+
+We can add a command to collect a single-node support bundle to reduce the issue resolution time.
+
+A single-node support bundle is handy in these cases:
+- Troubleshooting a failed installation.
+- Troubleshooting a failed bootstrapped node. e.g., a node can't join a Harvester cluster.
+- When the Harvester backend or Kubernetes runtime has problems.
+
+### Related Issues
+
+- https://github.com/harvester/harvester/issues/1864
+
+## Motivation
+
+### Goals
+
+- A user can run a command on a node to generate a node support bundle.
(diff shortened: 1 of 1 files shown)


guangbochen:
+1, as this would be very helpful for future debugging support. As per the issue description, we may consider leveraging the [openSUSE Supportutils](https://en.opensuse.org/Supportutils) since the Harvester baseOS is already based on the openSUSE kernel. (May consider adding this one after v1.0.1 release upon the time buffer)

bk201:
@ibrokethecloud Pleas feel free to modify the HEP if it doesn't fit your design, thanks!

ibrokethecloud:
the PR introduces the basic functionality for the same:  https://github.com/harvester/harvester-installer/pull/272

We can also add ability to upload these objects to a NFS or Object store endpoint.

enhancements/20220128-single-node-support-bundle.md
```diff
@@ -0,0 +1,167 @@
...
+harvester/oem
+harvester/oem/99_custom.yaml
+harvester/oem/harvester.config
+harvester/console.log
+```
+
+Similar mechanisms apply to other plugins as well.
+
+Note the config files might contain sensitive information like passwords. We need to remove them when generating a support bundle.
+
```
ibrokethecloud, line 156: # Upload support config
Introduce support for NFS / Object store based upload options for support config.
This will make it easier to collate support configs at a common store, and may be useful for MaaS capability.

Issue #1650, opened 2021-12-09
[BUG] Restoring status from backup never changes to running


noahgildersleeve:
**Describe the bug**

When you restore a vm from backup it goes to restoring status, but never back to running
**To Reproduce**
1. Create new VM
2. Take backup
3. restore from backup
**Expected behavior**

After restore is done the status should go back to running
**Support bundle**

[supportbundle_754ee1c7-96c6-45f2-98a9-705f42b7a2f6_2021-12-09T01-38-27Z.zip](https://github.com/harvester/harvester/files/7681560/supportbundle_754ee1c7-96c6-45f2-98a9-705f42b7a2f6_2021-12-09T01-38-27Z.zip)

**Environment:**
 - Harvester ISO version: master-655c251a-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 node vagrant ipxe setup

**Additional context**
If you refresh the page it will update with running, but it requires a manual page load

### Test plan

Take Backup without error.

1. Create a VM.
2. Setup Backup Target.
3. Take a VMBackup for the VM.
4. Before VMBackup is ready, there should not have any error message in `VMBackup.status`.

yasker:
sounds like the websocket problem. can you check if there are error messages in the developer's console? Reenable the warning should solve it.

noahgildersleeve:
I tried it again and checked the web console and the network traffic via developer tools and didn't see anything. I also did all of this after resetting the enable warning setting on the SSL cert. A refresh still fixed the issue.

noahgildersleeve:
When you create a backup the label updates after creation with an error. I tested this on an NFS server that was created on the embedded Rancher, but it should do it on any one.
## Validation steps
1. Set a valid NFS target in settings
2. Create a new VM
3. Log into VM and create a file and put some text in it
4. run `sync` to write out memory to the disk
5. run a NFS backup
6. immediately check backups page in Harvester UI
![Screenshot from 2022-01-28 12-17-29](https://user-images.githubusercontent.com/83787952/151616869-885dec3e-4297-40db-a832-12b644a480f9.png)
[supportbundle_a321024d-5846-41f4-afc3-6ac406a12476_2022-01-28T20-17-44Z.zip](https://github.com/harvester/harvester/files/7961659/supportbundle_a321024d-5846-41f4-afc3-6ac406a12476_2022-01-28T20-17-44Z.zip)

WuJun2016:
[...]

PR #1900, opened 2022-02-09
Ignore "timestamp: nil Timestamp" error in VMBackup

FrankYang0529:
**Problem:**
VMBackup shows an error message `timestamp: nil Timestamp` from VolumeSnapshot, but it's temporary. It will be removed after `backup.snapshotCreatedAt` has a non-empty value. Also, the VMBackup is healthy and can be used to restore an existing/ new VM.

**Solution:**
Ignore the `timestamp: nil Timestamp` error message, so users don't see the weird behavior.

**Related Issue:**
https://github.com/harvester/harvester/issues/1650
https://github.com/longhorn/longhorn/issues/3392

**Test plan:**
1. Create a VM.
2. Create a VMBackup for it.
3. UI should not show Error status before VMBackup is Ready.

### DIFF
--- a/pkg/controller/master/backup/backup_status.go
+++ b/pkg/controller/master/backup/backup_status.go
@@ -33,6 +33,13 @@ func (h *Handler) updateConditions(vmBackup *harvesterv1.VirtualMachineBackup) e
 		}
 
 		if vb.Error != nil {
+			// LH CSI shows "timestamp: nil Timestamp" error before LH backup.SnapshotCreated has a value.
+			// However, this is a tempoary error, so we ignore it here.
+			// We will remove this after LH issue 3392 is resolved.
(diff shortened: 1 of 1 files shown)


guangbochen:
Thanks for the PR, I just double-checked the upstream LH issue https://github.com/longhorn/longhorn/issues/3392 is planning to back-port the fix to its `v1.2.4` release, so let's move this issue to ice-box and wait for its release, if the upstream didn't fix this one we can merge this PR to resolve it temporarily, thanks.

PhanLe1010:
Upstream Longhorn PR is created at https://github.com/longhorn/longhorn-manager/pull/1233. Harvester might not need to fix this issue

guangbochen:
should be fine to close this one of the LH dependency bumps to `v1.2.4-rc1`, thanks.

Issue #1908, opened 2022-02-11
Remove unused nfs.go file


FrankYang0529:
`pkg/controller/master/backup/nfs.go` is not referenced by any file. We've already changed to use LH [backupstore](https://github.com/longhorn/backupstore) to operate NFS/S3 backup.

lanfon72:
Does this need any tests?

FrankYang0529:
> Does this need any tests?

No, I think it's good to close it.

noahgildersleeve:
Tested in `master-45182389-head`.

## Verification steps

1. Create new VM
2. Add file1 and content
3. create backup
4. edit file1 and add file2
5. create backup
6. edit file2 create file3 and content
7. Create backup
8. Restore to backup1
9. Verify that only original file1 exists

PR #1909, opened 2022-02-11
Remove unused file nfs.go

FrankYang0529:
**Problem:**
`pkg/controller/master/backup/nfs.go` is not referenced by any file. We've already changed to use LH [backupstore](https://github.com/longhorn/backupstore) to operate NFS/S3 backup.

**Solution:**
Remove it and check whether NFS backup can work.

**Related Issue:**
https://github.com/harvester/harvester/issues/1908

**Test plan:**
1. Setup NFS backup target.
2. Create a VM and create a file with content `test-123`.
3. Run `sync` command.
4. Create a VMBackup. It should be created without any error.
5. Modify file content in the VM.
6. Restore the VMBackup to the existing/new VM.
7. The file in the existing/new VM should have content `test-123`.


guangbochen (approved):
LGTM

Issue #1868, opened 2022-01-26
[FEATURE] Switch Harvester components base image to SLE BCI


yasker:
It's maintained much better than e.g. alpine.

The scope includes all the Harvester components, e.g. cloud provider etc. 

E.g. https://github.com/rancher/rancher/blob/release/v2.6/package/Dockerfile

bk201:
Harvester image list that we might want to switch:
- [x] https://github.com/harvester/harvester
  - [x] rancher/harvester
  - [x] rancher/harvester-upgrade
  - [x] rancher/harvester-webhook
  - [x] rancher/harvester-cluster-repo
- [x] https://github.com/harvester/harvester-installer
- [x] https://github.com/harvester/network-controller-harvester
  - [x] rancher/harvester-network-controller
  - [x] rancher/harvester-network-helper
- [x] https://github.com/harvester/node-disk-manager
  - [x] rancher/harvester-node-disk-manager
- [x] https://github.com/harvester/load-balancer-harvester
  - [x] rancher/harvester-load-balancer
- [x] https://github.com/harvester/cloud-provider-harvester
- [x] https://github.com/harvester/terraform-provider-harvester
- [x] https://github.com/harvester/docker-machine-driver-harvester
- [x] https://github.com/harvester/harvester-csi-driver - will be fixed separately via https://github.com/harvester/harvester/issues/1995
- [x] [rancher/support-bundle-kit](https://github.com/rancher/support-bundle-kit)

guangbochen:
Close as a majority of the Harvester components is migrated to use the SLE BCI image, during the Harvester v1.0.1 regression test we should be able to verify if it will cause any side effects.

PR #1910, opened 2022-02-11
Migrate base image to BCI image

bk201:
Migrate base image to BCI image

**Related Issue:**
https://github.com/harvester/harvester/issues/1868

**Test plan:**

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,6 +1,10 @@
-FROM alpine
-RUN apk update && apk add -u --no-cache git curl unzip tar tini bash nfs-utils && \
-    adduser -D harvester && su -l harvester && \
+FROM registry.suse.com/bci/bci-base:15.3
+
+# nfs-client is needed by the dep https://github.com/longhorn/backupstore to check backup store availability.
+RUN zypper -n rm container-suseconnect && \
+    zypper -n install curl gzip tar nfs-client && \
+    zypper -n clean -a && rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/* && \
+    useradd -M harvester && \
     mkdir -p /var/lib/harvester/harvester && \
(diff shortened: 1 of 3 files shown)


guangbochen (approved):
LGTM

package/Dockerfile
```diff
@@ -1,6 +1,9 @@
-FROM alpine
-RUN apk update && apk add -u --no-cache git curl unzip tar tini bash nfs-utils && \
```
weihanglo, line 2: Why remove nfs-utils? Is nfs utility built-in for bci-base image?

bk201: https://github.com/harvester/harvester/pull/1909
It should be safe to remove nfs-utils since Harvester doesn't use it.

guangbochen: Actually, the `nfs-utils` is still needed as we just move the dependency library to https://github.com/longhorn/backupstore, and the nfs mount within the container will still require to use this library.

bk201: Thanks, I add the corresponding package `nfs-client`: https://github.com/harvester/harvester/compare/75cc5c6fd7b1b6ac25b0256106a24532d7bec41a..47630964c440f93cb9148a6044246bb7eb129b64

weihanglo (approved):
LGTM

It will be better if we can document why the package is needed. Anyhow, this could be done in the follow-up PR if the timeframe is not that tight.

bk201:
> LGTM
> 
> It will be better if we can document why the package is needed. Anyhow, this could be done in the follow-up PR if the timeframe is not that tight.

Comments added: https://github.com/harvester/harvester/pull/1910/files#diff-c55d1ee74f32cd4b9eeeceaec981a180f52e761acb158a5b0b0e088c7725c9caR3 Thanks.

weihanglo (approved):
LGTM

guangbochen (approved):

Issue #1894, opened 2022-02-07
[BUG] login with member role, no secret api access


WuJun2016:
**Describe the bug**

The user login with the member role, views the vm template, and the page reports an error display no secret access

![image](https://user-images.githubusercontent.com/24985926/153567327-408f7717-5288-40b3-ba7c-b093de6e8bed.png)

![image](https://user-images.githubusercontent.com/24985926/153567416-4dd3a71f-dde7-46c4-8067-e303b4ccfc64.png)

The cloud-init in the template is saved via secret, so we need to have access to the secret
**To Reproduce**
1. login with member role
2. click  Advanced => Templates
3. click view config

**Expected behavior**

view config works fine

**Support bundle**

**Environment:**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

---
# Dev update @johnliu55tw 
## Detail reproduce steps
1. Create a Harvester cluster (one node would be sufficient)
2. Follow the instructions on https://docs.harvesterhci.io/v1.0/rancher/rancher-integration/ to create a Rancher deployment and add the Harvester cluster into Rancher
3. In Rancher, create a new user `someuser` as "Standard User"
4. Go to the "Virtualization Management" page of Rancher -> Click "Manage" of the Harv. cluster -> Edit Config -> Add user `someuser` as "Member", instead of "Owner" or "Custom"
5. Logout from Rancher, and login with user `someuser`
6. Go to Virtualization Management -> Click the Harv. cluster -> Go to Advanced/Templates -> Click any template -> Click "View Config" on the menu on the top-right corner
7. **It shows nothing**

johnliu55tw:
## Root cause
Users with the "Member" role are not granted to view `secrets` resources.

## Fix
The permissions to view the Harvester resources are controlled by ClusterRole [harvesterhci.io:view](https://github.com/harvester/harvester/blob/master/deploy/charts/harvester/templates/rbac.yaml#L53-L83). We could add resource `secrets` to this ClusterRole to grant the users with Member role to view the secrets:
```
...
rules:
- apiGroups:
    - ""
  resources:
    - secrets
  verbs:
    - get
    - list
    - watch
...
```

[...]

PR #1916, opened 2022-02-15
Add permission to view secrets

johnliu55tw:

**Problem:**
Cloudconfig in VM templates is stored as a secret resource, so users with "Member" type of role must have the permissions to **view secrets** of the namespace in order to view the VM templates.

**Solution:**
The permissions to view the Harvester resources are controlled by ClusterRole [harvesterhci.io:view](https://github.com/harvester/harvester/blob/master/deploy/charts/harvester/templates/rbac.yaml#L53-L83). We could add resource `secrets` to this ClusterRole to grant the users with Member role to view the secrets:
```
...
rules:
- apiGroups:
    - ""
  resources:
    - secrets
  verbs:
    - get
    - list
    - watch
...
```

This ClusterRole will first be [aggregated](https://github.com/harvester/harvester/blob/9091d7d0de9e99ddc677e32c988d758065f87fbb/deploy/charts/harvester/templates/rbac.yaml#L60) into another ClusterRole `view`, then this `view` will be bound to a Group `system:authenticated` by a RoleBinding `harvester-public` in the namespace `harvester-public`. In other words, only secrets in `harvester-public` namespace will be revealed to the users with Member role.

**Related Issue:**
https://github.com/harvester/harvester/issues/1894

**Test plan:**
1. Create a Harvester cluster (one node would be sufficient)
2. Follow the instructions on https://docs.harvesterhci.io/v1.0/rancher/rancher-integration/ to create a Rancher deployment and add the Harvester cluster into Rancher
3. In Rancher, create a new user `someuser` as "Standard User"
4. Go to the "Virtualization Management" page of Rancher -> Click "Manage" of the Harv. cluster -> Edit Config -> Add user `someuser` as "Member", instead of "Owner" or "Custom"
5. Logout from Rancher, and login with user `someuser`
6. Go to Virtualization Management -> Click the Harv. cluster -> Go to Advanced/Templates -> Click any template
7. Verify that being a **"Member" role of user, you are able to see the detail of the VM template**.

### DIFF
--- a/deploy/charts/harvester/templates/rbac.yaml
+++ b/deploy/charts/harvester/templates/rbac.yaml
@@ -81,6 +81,14 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - list
+      - watch
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole


weihanglo (approved):
Tested and it behaves as-is. LGTM

guangbochen (approved):
LGTM
```
$ks auth can-i list secrets -n harvester-public
yes
$ks auth can-i list secrets -n default
no
```

Issue #1695, opened 2021-12-17
[BUG] VM with unschedule disks doesn't show clear warning message


yasker:
**Describe the bug**
VM with unschedule disks doesn't show clear warning message

**To Reproduce**
1. Create a VM that exceed the storage capacity, e.g. 100T
2. VM failed to start, but no error message giving to the user
3. The only available information in the event is

Reason | Resource | Date
-- | -- | --
FailedMount | Pod virt-launcher-test-nzg65Unable to attach or mount volumes: unmounted volumes=[disk-0], unattached volumes=[container-disks sockets disk-0 cloudinitdisk-ndata public ephemeral-disks hotplug-disks libvirt-runtime cloudinitdisk-udata private]: timed out waiting for the condition | 1.3 mins ago
FailedAttachVolume | Pod virt-launcher-test-nzg65AttachVolume.Attach failed for volume "pvc-db775cbb-a5a2-4479-83f6-d7d9af85bfb7" : rpc error: code = Aborted desc = volume pvc-db775cbb-a5a2-4479-83f6-d7d9af85bfb7 is not ready for workloads | 1.3 mins ago
FailedMount | Pod virt-launcher-test-nzg65Unable to attach or mount volumes: unmounted volumes=[disk-0], unattached volumes=[ephemeral-disks private libvirt-runtime cloudinitdisk-udata public container-disks hotplug-disks cloudinitdisk-ndata disk-0 sockets]: timed out waiting for the condition

**Expected behavior**
VM should show similar warning information to the user like `insufficent storage`

**Support bundle**

**Environment:**
 - Harvester ISO version: v1.0.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): any

**Additional context**

guangbochen:
There is no obvious message currently showing on the Harvester side due to complicated resource mapping, will make an enhancement on the next version, with the current version you can visit the [Longhorn UI](https://docs.harvesterhci.io/v1.0/troubleshooting/harvester/#access-embedded-longhorn) and check the detailed volume issues accordingly.

w13915984028:
reproduced:
1. volume with massive size
![image](https://user-images.githubusercontent.com/31133476/149165237-c4f91b1f-8fa7-4986-8006-d8c8c0d3454c.png)
2. vm is in "Starting"
![image](https://user-images.githubusercontent.com/31133476/149163669-89767991-d8fa-40a8-8c8a-abeeb58b49b3.png)

[...]

Issue #97, opened 2020-09-01
image list and detail page issue


DoNotSayYes:
- [x] 1. The state of the image on the list page, but unknown on the detail page
![image](https://user-images.githubusercontent.com/57614736/91790860-4fdc7280-ec44-11ea-845a-5bbef3e4a35b.png)
![image](https://user-images.githubusercontent.com/57614736/91790950-7ef2e400-ec44-11ea-9749-e5cf5b4ef570.png)

- [x] 2. Sort by the state on the list page could make the page work inappropriately, something like the menubar does not pop up by clicking the ellipsis
![image](https://user-images.githubusercontent.com/57614736/91791291-57e8e200-ec45-11ea-99eb-4fb767c3ffe8.png)

- [x] 3. Select one of the items, proceed clone as form, users are not able to add more than one label or annotation
![image](https://user-images.githubusercontent.com/57614736/91791483-d80f4780-ec45-11ea-93fc-23fa5d88ec9c.png)

- [ ] 4. Is it necessary to suppress the `create a virtual machine ` on the menu bar if the image state is not healthy?
![image](https://user-images.githubusercontent.com/57614736/91792896-38ec4f00-ec49-11ea-8f30-9fa46079e6c2.png)

- [x] 5. This functionality does work
![image](https://user-images.githubusercontent.com/57614736/91801882-21b75c80-ec5d-11ea-9fae-9920dbea7aae.png)

guangbochen:
remove the protips hint from the dialog.

DoNotSayYes:
Validated all except the 4th item

PR #1923, opened 2022-02-17
fix1695: update LH volume status to pvc and vm annotations

w13915984028:

**Problem:**

LH volumes` bad status like disk space is not enough,  is not easily found in Harvester WEBUI and kubectl.

**Solution:**

This fix tries to abstract the error status of LH volume, and update them to PVC and VM`s annotations
 With web UI`s fix (coming in another PR), the info will be displayed in PVC and VM page.
 
1. update LH volume status to PVC when change and remove
2. update PVC status to VM when change and remove
3. strip VM volume status when it is restored from vm backup

**Related Issue:**
#1695 

**Test plan:**

image

image

### DIFF
--- a/pkg/controller/master/virtualmachine/register.go
+++ b/pkg/controller/master/virtualmachine/register.go
@@ -12,6 +12,7 @@ const (
 	vmControllerUnsetOwnerOfPVCsControllerName         = "VMController.UnsetOwnerOfPVCs"
 	vmiControllerUnsetOwnerOfPVCsControllerName        = "VMIController.UnsetOwnerOfPVCs"
 	vmControllerSetDefaultManagementNetworkMac         = "VMController.SetDefaultManagementNetworkMacAddress"
+	vmControllerSetAnnotationsVolumeStatusName         = "VMController.SetVmAnnotationsVolumeStatus"
 )
 
 func Register(ctx context.Context, management *config.Management, options config.Options) error {
@@ -50,6 +51,19 @@ func Register(ctx context.Context, management *config.Management, options config
 		vmiClient: vmiClient,
 	}
 	virtualMachineInstanceClient.OnChange(ctx, vmControllerSetDefaultManagementNetworkMac, vmNetworkCtl.SetDefaultNetworkMacAddress)
+
+	// register the vm volume status controller, which tries to update LH Volume status to PVC and VM
+	var volumeClient = management.LonghornFactory.Longhorn().V1beta1().Volume()
+	var vmVolumeStatusController = &VMVolumeStatusController{
+		vmClient:  vmClient,
+		vmCache:   vmCache,
+		pvcClient: pvcClient,
+		pvcCache:  pvcCache,
+	}
+	volumeClient.OnChange(ctx, vmControllerSetAnnotationsVolumeStatusName, vmVolumeStatusController.OnLHVolumeChange)
+	volumeClient.OnRemove(ctx, vmControllerSetAnnotationsVolumeStatusName, vmVolumeStatusController.OnLHVolumeRemove)
+	pvcClient.OnChange(ctx, vmControllerSetAnnotationsVolumeStatusName, vmVolumeStatusController.OnPVCChange)
+	pvcClient.OnRemove(ctx, vmControllerSetAnnotationsVolumeStatusName, vmVolumeStatusController.OnPVCRemove)
 
 	return nil
 }
--- a/pkg/controller/master/virtualmachine/vm_volume_status_controller.go
+++ b/pkg/controller/master/virtualmachine/vm_volume_status_controller.go
@@ -1,0 +1,490 @@
+package virtualmachine
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	longhornv1 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+	corectl "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubevirtapis "kubevirt.io/api/core/v1"
+
+	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+
+	"github.com/harvester/harvester/pkg/ref"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	// only specific LH volume status is checked and updated into virtual machine
+	// all other states are categorized into  "OK"
+	volumeStatusOK                  = volumeStatus("ok")
+	volumeStatusInsufficientStorage = volumeStatus("insufficient storage")
+)
+
+type VMVolumeStatusController struct {
+	vmClient  ctlkubevirtv1.VirtualMachineClient
+	vmCache   ctlkubevirtv1.VirtualMachineCache
+	pvcClient corectl.PersistentVolumeClaimClient
+	pvcCache  corectl.PersistentVolumeClaimCache
+}
+
+type volumeStatus string
+
+type volumeStatusAnnotation struct {
+	// name of volume
+	Name string `json:"name,omitempty"`
+	// abstracted status based on longhorn (LH) volume status
+	Status string `json:"status,omitempty"`
+}
+
+type volumeStatusAnnotationSlice []volumeStatusAnnotation
+
+func (v volumeStatus) String() string {
+	return string(v)
+}
+
+// try to update LH Volume status to PVC Status
+func (h *VMVolumeStatusController) OnLHVolumeChange(key string, lhVolume *longhornv1.Volume) (*longhornv1.Volume, error) {
+	if lhVolume == nil || lhVolume.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	// when volume is not connected to pvc, skip
+	pvcName := lhVolume.Status.KubernetesStatus.PVCName
+	namespace := lhVolume.Status.KubernetesStatus.Namespace
+	if pvcName == "" || namespace == "" {
+		return nil, nil
+	}
+
+	volStatus := volumeStatusOK
+	// before LH has a better solution, we can only hard code those conditions to deduce one volume is in none-OK state
+	// new status should also be generated here
+	if lhVolume.Status.State == longhornv1.VolumeStateDetached && lhVolume.Status.Robustness == longhornv1.VolumeRobustnessUnknown {
+		if c, ok := lhVolume.Status.Conditions[longhornv1.VolumeConditionTypeScheduled]; ok {
+			if c.Status == longhornv1.ConditionStatusFalse && c.Reason == longhornv1.VolumeConditionReasonReplicaSchedulingFailure {
+				// assume insufficient storage
+				volStatus = volumeStatusInsufficientStorage
+			}
+		}
+	}
+	return nil, h.updateLHVolumeStatusToPVC(lhVolume.Name, namespace, pvcName, volStatus)
+}
+
+// try to remove volume status from PVC Status if it is there
+func (h *VMVolumeStatusController) OnLHVolumeRemove(key string, lhVolume *longhornv1.Volume) (*longhornv1.Volume, error) {
+	if lhVolume == nil {
+		return nil, nil
+	}
+
+	pvcName := lhVolume.Status.KubernetesStatus.PVCName
+	namespace := lhVolume.Status.KubernetesStatus.Namespace
+	if pvcName == "" || namespace == "" {
+		return nil, nil
+	}
+
+	return nil, h.removeLHVolumeStatusFromPVC(lhVolume.Name, namespace, pvcName)
+}
+
+// LH volume is changed, update the possible volume status to PVC
+func (h *VMVolumeStatusController) updateLHVolumeStatusToPVC(volName string, pvcNamespace string, pvcName string, status volumeStatus) error {
+	var err error
+	var pvc *corev1.PersistentVolumeClaim
+	// pvc may be deleted, but LH volume is still there, in such case, give up updating
+	if pvc, err = h.getPVC(pvcNamespace, pvcName); err != nil {
+		if !errors.IsNotFound(err) {
+			logrus.Infof("fail to get pvc:(%s/%s), give up to update LH volume:%s status:%s, error:%s", pvcNamespace, pvcName, volName, status.String(), fmt.Errorf("%w", err))
+		}
+		return nil
+	}
+
+	if status == volumeStatusOK {
+		if _, ok := pvc.Annotations[util.AnnotationVolumeStatus]; ok {
+			// remove existing value
+			if _, err := h.removePVCAnnotationsVolumeStatus(pvc); err != nil {
+				if !errors.IsConflict(err) {
+					logrus.Infof("pvc:(%s/%s), fail to remove volume:%s status, error:%s", pvcNamespace, pvcName, volName, fmt.Errorf("%w", err))
+				}
+				return err
+			}
+		}
+	} else {
+		if volStatus, ok := pvc.Annotations[util.AnnotationVolumeStatus]; !ok || volStatus != status.String() {
+			if _, err := h.updatePVCAnnotationsVolumeStatus(pvc, status.String()); err != nil {
+				if !errors.IsConflict(err) {
+					logrus.Infof("pvc:(%s/%s), fail to update volume:%s status:%s, error:%s", pvcNamespace, pvcName, volName, status.String(), fmt.Errorf("%w", err))
+				}
+				return err
+			}
+
+		}
+	}
+
+	return nil
+}
+
+// LH volume is deleted, remove the possible volue status from PVC
+func (h *VMVolumeStatusController) removeLHVolumeStatusFromPVC(volName string, pvcNamespace string, pvcName string) error {
+	var err error
+	var pvc *corev1.PersistentVolumeClaim
+	if pvc, err = h.getPVC(pvcNamespace, pvcName); err != nil {
+		if !errors.IsNotFound(err) {
+			logrus.Infof("fail to get owner pvc:(%s/%s), give up to remove status of LH volume:%s, error:%s", pvcNamespace, pvcName, volName, fmt.Errorf("%w", err))
+		}
+		return nil
+	}
+
+	if _, ok := pvc.Annotations[util.AnnotationVolumeStatus]; ok {
+		if _, err := h.removePVCAnnotationsVolumeStatus(pvc); err != nil {
+			if !errors.IsConflict(err) {
+				logrus.Infof("pvc:(%s/%s), fail to remove LH volume:%s status, error:%s", pvcNamespace, pvcName, volName, fmt.Errorf("%w", err))
+			}
+			return err
+		}
+	}
+
+	return nil
+}
+
+// as per: func (o AnnotationSchemaOwners) List(ownerGK schema.GroupKind) []string
+func (h *VMVolumeStatusController) getPVCRefOwnerVMs(pvc *corev1.PersistentVolumeClaim) ([]string, error) {
+	if pvcOwners, err := ref.GetSchemaOwnersFromAnnotation(pvc); err != nil {
+		return nil, err
+	} else if len(pvcOwners) == 0 {
+		// pvc is not attched to any owner yet
+		return nil, nil
+	} else {
+		vmGVK := kubevirtapis.VirtualMachineGroupVersionKind
+		vmGK := vmGVK.GroupKind()
+		refOwnerVms := pvcOwners.List(vmGK)
+		return refOwnerVms, nil
+	}
+}
+
+// pvc annotation is further updated to VM annotation
+func (h *VMVolumeStatusController) OnPVCChange(key string, pvc *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+	if pvc == nil || pvc.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
+	var err error
+	// process "pvc is detached vm" first, pvc may be updated after this call, following needs to return pvc
+	if pvc, err = h.removeVolumeStatusFromDetachedVM(pvc); err != nil {
+		return nil, err
+	}
+
+	var refOwnerVms []string
+	if refOwnerVms, err = h.getPVCRefOwnerVMs(pvc); err != nil {
+		return pvc, err
+	} else if len(refOwnerVms) == 0 {
+		// pvc is not attched to any owner yet
+		return pvc, nil
+	}
+
+	update := false
+	var volStatus string
+	var ok bool
+	// when pvc has !empty volume annotations, then update(add) to vm, otherwise remove from vm
+	if volStatus, ok = pvc.Annotations[util.AnnotationVolumeStatus]; ok && volStatus != "" {
+		update = true
+	}
+
+	for _, owner := range refOwnerVms {
+		namespace, vmName := ref.Parse(owner)
+		var err error
+		var vm *kubevirtapis.VirtualMachine
+		if vm, err = h.getVM(namespace, vmName); err != nil {
+			if !errors.IsNotFound(err) {
+				logrus.Infof("fail to get owner vm:(%s/%s), give up to update pvc status:%s, error:%s", namespace, vmName, volStatus, fmt.Errorf("%w", err))
+			}
+			continue
+		}
+
+		if update {
+			_, err = h.updateVolumeStatusToVM(vm, volumeStatusAnnotation{Name: pvc.Name, Status: volStatus})
+		} else {
+			_, err = h.removeVolumeStatusFromVM(vm, volumeStatusAnnotation{Name: pvc.Name})
+		}
+		if err != nil {
+			if !errors.IsConflict(err) {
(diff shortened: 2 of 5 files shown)


w13915984028:
The design consideration:

image

image

image

image

weihanglo:
As a side note: I believe that current KibeVirt supports hotplug and hotunplug disks without restart VMs, but it seems that harvesters haven't integrated that part well.

pkg/util/constants.go
```diff
@@ -12,6 +12,8 @@ const (
 	AnnotationVolumeClaimTemplates = prefix + "/volumeClaimTemplates"
 	AnnotationImageID              = prefix + "/imageId"
 	AnnotationHash                 = prefix + "/hash"
+	AnnotationVolumeStatus         = prefix + "/volume-status"
+	AnnotationDetachedVM           = prefix + "/detached-vm"
```
weihanglo: Would it possible to document the usage of these annotations **inline**? It helps a lot for future maintainers reading through the source code itself.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+	pvcClient corectl.PersistentVolumeClaimClient
+	pvcCache  corectl.PersistentVolumeClaimCache
+}
+
+type volumeStatus string
+
+type volumeStatusAnnotation struct {
+	// name of volume
+	Name string `json:"name,omitempty"`
+	// abstracted status based on long-horn (LH) volume status
```
weihanglo: NIT: long-horn -> longhorn

```suggestion
	// abstracted status based on longhorn (LH) volume status
```

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+	// before LH has a better solution, we can only hard code those conditions to deduce one volume is in none-OK state
+	// new status should also be generated here
+	if lhVolume.Status.State == longhornv1.VolumeStateDetached && lhVolume.Status.Robustness == longhornv1.VolumeRobustnessUnknown {
+		if c, ok := lhVolume.Status.Conditions[longhornv1.VolumeConditionTypeScheduled]; ok {
+			if c.Status == longhornv1.ConditionStatusFalse && c.Reason == longhornv1.VolumeConditionReasonReplicaSchedulingFailure {
+				// assume insufficient storage
+				volStatus = volumeStatusInsufficientStorage
+			}
+		}
+	}
```
weihanglo, line 74: If I understand it correctly, these lines are just a heuristic deduction for `volumeStatusInsufficientStorage`, right?

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+	// pvc may be deleted, but LH volume is still there, in such case, give up updating
+	if pvc, err = h.getPVC(pvcNamespace, pvcName); err != nil {
+		if !errors.IsNotFound(err) {
+			logrus.Infof("fail to get pvc:(%s/%s), give up to update LH volume:%s status:%s, error:%s", pvcNamespace, pvcName, volName, status.String(), fmt.Errorf("%w", err))
+		}
+		return nil
+	}
+
+	if status == volumeStatusOK {
+		if _, ok := pvc.Annotations[util.AnnotationVolumeStatus]; ok {
```
weihanglo, line 106: NIT: This `pvc.Annotations[util.AnnotationVolumeStatus]` duplicates with line 116. Can we avoid the duplication here?

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+
+	for _, owner := range refOwnerVms {
+		namespace, vmName := ref.Parse(owner)
+		var err error
+		var vm *kubevirtapis.VirtualMachine
+		if vm, err = h.getVM(namespace, vmName); err != nil {
+			if !errors.IsNotFound(err) {
+				logrus.Infof("fail to get owner vm:(%s/%s), give up to update pvc status:%s, error:%s", namespace, vmName, volStatus, fmt.Errorf("%w", err))
+			}
```
weihanglo, line 203: Should this kind of error be at log level "Warn" or "Error"?

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+		update = true
+	}
+
+	for _, owner := range refOwnerVms {
+		namespace, vmName := ref.Parse(owner)
+		var err error
+		var vm *kubevirtapis.VirtualMachine
+		if vm, err = h.getVM(namespace, vmName); err != nil {
+			if !errors.IsNotFound(err) {
+				logrus.Infof("fail to get owner vm:(%s/%s), give up to update pvc status:%s, error:%s", namespace, vmName, volStatus, fmt.Errorf("%w", err))
```
weihanglo, line 202: From my experience, formatting with `%w` in `logrus.Infof` works. (Other places need similar updates as well)

```suggestion
				logrus.Infof("fail to get owner vm:(%s/%s), give up to update pvc status:%s, error:%w", namespace, vmName, volStatus, err)
```

weihanglo (commented):
Overall it's in a good shape. Would you mind adding some tests, as other controllers do?

Edit: All test case you listed work as expected.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: Yes, as comment in L65, it is hard coded as per LH's implementation now.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: `logrus does not support %w`, it will print something like 

time="2022-02-11T10:44:51Z" level=error msg="fail to get pvc:(default/vol-600g), give up to map LH volume:pvc-7717ffe7-f610-4496-aa0a-bc5c6feb04ff status:insufficient storage, error:`%!w(*fmt.wrapError=&{fail to get pvc(default/vol-600g), persistentvolumeclaims \"vol-600g\" not found 0xc003334960})"`

weihanglo:
If I interpret it correctly, the last test case in the test plan is adding a hotplug volume. At first glance, it seems weird that removing volume need a VM restart. However, if we have the "Remove volume" (hot-unplug) button, then it should be removed without restart the VM. I feel like you don't need to handle that situation in this PR.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: The consideration is, this feature is a patch to current implementation,  tries best to update status in-between different objects. Info looks to be the proper level, it leaves useful information to trouble-shooting, and does not bring big pressure to users.

pkg/util/constants.go
w13915984028: It is simply following the existing ones in constants.go, people needs to check the implementation part to understand what those constants are doing.

I don't mind adding comments here, but it will looks unusual comparing existing code.

w13915984028:
> If I interpret it correctly, the last test case in the test plan is adding a hotplug volume. At first glance, it seems weird that removing volume need a VM restart. However, if we have the "Remove volume" (hot-unplug) button, then it should be removed without restart the VM. I feel like you don't need to handle that situation in this PR.

--
The tricky is, one volume may be in `"unusable"` real state even though it shows` "Ready" `, no matter normal `"add volume"` or  `"hotplug"`  , we should block those actions. Otherwise, the effect is unpredictable.

I filed #1931 to track it.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: Good, will change.

pkg/util/constants.go
weihanglo: IMO it's worth doing even only these two annotations containing comments. There are lots of code have no comment explaining their business logic and for me it's hard to figure them out. Still, only my own opinion. The decision is up to you 😃.

w13915984028:
> Overall it's in a good shape. Would you mind adding some tests, as other controllers do?
> 
> Edit: All test case you listed work as expected.

--
The "test" file will be designed and committed in another PR. The status is updated from LH volume, to PVC, to VM, is a bit complex. This PR makes sure those manual test cases are well covered first.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: Candidate is putting` "volStatus, ok := pvc.Annotations[util.AnnotationVolumeStatus]"`  to L104, then in case "volumeStatusOK " go may complain `volStatus defined but not used` . 
Keep unchanged will not hurt too much.

pkg/util/constants.go
w13915984028: Updated.

weihanglo (approved):
Thanks for the update. Looks great from my side

weihanglo:
Oh unfortunately the code hit the CodeFactor CI check 🤯 
You might need to extract items from the so-called "complex code" to other files.
That's my bad suggesting you to add doc comment on those items 🙇🏾

w13915984028:
> Oh unfortunately the code hit the CodeFactor CI check 🤯 You might need to extract items from the so-called "complex code" to other files. That's my bad suggesting you to add doc comment on those items 🙇🏾

--
The `CodeFactor `issue here is interesting.
I did a test:
<details>
It claims complex code of `pkg/controller/master/backup/restore.go`
image

There are 3 lines added in `restore.go`
image

image


I tried to remove those 3 lines from `restore.go`

`CodeFactor `does not complain anymore, all other 4 files are fine.
image

image
</details>

Conclusion: `CodeFactor ` here complains the existing code in  `pkg/controller/master/backup/restore.go`, please skip it.

w13915984028:
As tested and explained in https://github.com/harvester/harvester/pull/1923#issuecomment-1048564904

Please skip the `CodeFactor ` failing in this PR, it is complaining the existing code in `pkg/controller/master/backup/restore.go`, saying some parts are complex codes, but it does not mention which function.

@weihanglo @FrankYang0529 @guangbochen

FrankYang0529 (approved):
Build and test on my cluster. It can work. LGTM

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+	return h.removePVCAnnotationsDetachedVM(pvc)
+}
+
+func (h *VMVolumeStatusController) getPVC(namespace string, name string) (*corev1.PersistentVolumeClaim, error) {
+	if pvc, err := h.pvcCache.Get(namespace, name); err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, err
+		} else {
+			return h.pvcClient.Get(namespace, name, metav1.GetOptions{})
+		}
```
guangbochen, line 296: we should return the error and reconcile it directly.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+		return pvc, nil
+	}
+}
+
+func (h *VMVolumeStatusController) getVM(namespace string, name string) (*kubevirtapis.VirtualMachine, error) {
+	if vm, err := h.vmCache.Get(namespace, name); err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, err
+		} else {
+			return h.vmClient.Get(namespace, name, metav1.GetOptions{})
```
guangbochen, line 307: ditto

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+
+	return nil, nil
+}
+
+// after pvc is detached/unmounted from VM, remove its volume status from VM
+func (h *VMVolumeStatusController) removeVolumeStatusFromDetachedVM(pvc *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+	var namespace, vmName string
+	var ok bool
+	if namespace, vmName, ok = h.decodeAnnotationDetachedVMString(pvc); !ok {
```
guangbochen: prefer to aggregate those lines into one using `:=` assignment for better project code consistency.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+		return pvc, nil
+	}
+
+	logrus.Debugf("remove pvc:%s status from previous owner vm:(%s/%s)", pvc.Name, namespace, vmName)
+
+	var err error
+	var vm *kubevirtapis.VirtualMachine
+	if vm, err = h.getVM(namespace, vmName); err != nil {
```
guangbochen: ditto

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+}
+
+// update/add volume status to vm
+func (h *VMVolumeStatusController) updateVolumeStatusToVM(vm *kubevirtapis.VirtualMachine, status volumeStatusAnnotation) (*kubevirtapis.VirtualMachine, error) {
+	var volumeStatusAnno volumeStatusAnnotationSlice
+	var err error
+	if volumeStatusAnno, err = getVMVolumeStatusAnnotationSlice(vm); err != nil {
+		return nil, err
+	}
```
guangbochen: ditto.

pkg/controller/master/virtualmachine/register.go
```diff
@@ -51,5 +52,18 @@ func Register(ctx context.Context, management *config.Management, options config
...
+	// register the vm volume status controller, which tries to update LH Volume status to PVC and VM
+	var volumeClient = management.LonghornFactory.Longhorn().V1beta1().Volume()
+	var vmVolumeStatusController = &VMVolumeStatusController{
+		vmClient:  vmClient,
+		vmCache:   vmCache,
+		pvcClient: pvcClient,
+		pvcCache:  pvcCache,
+	}
+	volumeClient.OnChange(ctx, vmControllerSetAnnotationsVolumeStatusName, vmVolumeStatusController.OnLHVolumeChange)
+	volumeClient.OnRemove(ctx, vmControllerSetAnnotationsVolumeStatusName, vmVolumeStatusController.OnLHVolumeRemove)
```
guangbochen, line 64: Watching LH `Volume` resources on the Harvester side is probably a bit over resource-consuming as all of the volume CRs will be synced automatically every few seconds from the LH side.
I found the PVC status will only contain a `phase: Pending` when it is an `out-of-range` issue (this is 100% accurate according to https://github.com/harvester/harvester/issues/1695#issuecomment-1011181595, but I think the actual status can only be checked by the LH as a CSI driver, as a temporary workaround this might be acceptable on the Harvester side):
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    harvesterhci.io/imageId: default/image-rs882
    harvesterhci.io/owned-by: '[{"schema":"kubevirt.io.virtualmachine","refs":["default/big-disks"]}]'
    volume.beta.kubernetes.io/storage-provisioner: driver.longhorn.io
  creationTimestamp: "2022-03-05T08:12:08Z"
  finalizers:
  - kubernetes.io/pvc-protection
  name: big-disks-disk-0-lx8qo
  namespace: default
  resourceVersion: "22567268"
  uid: a23c2497-cec4-4574-8e3c-48de972f4247
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: "9223372036854775807"
  storageClassName: longhorn-image-rs882
  volumeMode: Block
status:
  phase: Pending
```

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: Because namespace, vmName will be used later  in the function

There are 3 ways to do here:

A. try best to use `:=` assignment and condition check in same line, like
if namespace, vmName, ok := h.decodeAnnotationDetachedVMString(pvc); !ok {
}
all following code will be nested in the {} as long as they need to use namespace or vmName, 
it is quite easy to write code which has several nested {}, for complex functionality, the multi nested {} will not be better than flat structure.

B. separate variable assignment and condition check
namespace, vmName, ok := h.decodeAnnotationDetachedVMString(pvc)
if (!ok) {}

C. use `var ` to declare variable, then let assignment and condition check be in one line, as here

B,C are roughly same, they give the benefit of reduced level of nested {}
 and both are seen in Harvester project.
 
 @guangbochen are you talking about B or A ?

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: refer comment in line L97: // pvc may be deleted, but LH volume is still there, in such case, give up updating

when we can not retrieve PVC, wo do not return error, just give up updating status, return nil, otherwise, it may block LH volume from being deleted (#L131) or updated here.

pkg/controller/master/virtualmachine/register.go
w13915984028: True. A better solution rely on LH. Here we can only try to cover some special and important cases.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: same reason as getPVC, the reverse updating will only try once and not return error when can not find upstream object, because the upstream object may have already bee deleted.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
guangbochen: Prefer to use `B` for better consistency, thanks.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
guangbochen: You don't need to get from the `pvcClient` again since the cache already contains the latest data with the k8s ListWatch mechanism. Return `nil, nil` if is errors.IsNotFound, otherwise defaults to return `pvc, nil`, and in line #97 u should return an error instead of swallowing the other potential errors.

w13915984028:
Remark: CodeFactor is happy again with all 5 changed files in https://github.com/harvester/harvester/commit/1980296fd9c353b59ad23cc6dcbece490aef78ba.
image

Here, it complained `pkg/controller/master/backup/restore.go`
https://github.com/harvester/harvester/pull/1923#issuecomment-1048564904


But those 2 commits have no different change in restore.go.

No idea why CodeFactor does so.

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: updated in https://github.com/harvester/harvester/commit/67f62d3557323ef2ec0893c2112b33303523a02c

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: updated in https://github.com/harvester/harvester/commit/67f62d3557323ef2ec0893c2112b33303523a02c

pkg/controller/master/virtualmachine/vm_volume_status_controller.go
w13915984028: You don't need to get from the pvcClient again since the cache already contains the latest data with the k8s ListWatch mechanism
"""

A POD normally watches a lot of data, when this POD starts up, those watched data will come in random order. I did see that when LH volume update came, but the related PVC data is not in cache yet. And LH will only update change (volume goes into unusable state) once. Additional call to h.pvcClient.Get will solve such cases. On the other hand, the "remove" will truly lead to errors.IsNotFound as that object has already been deleted.  There are more "update" than "remove", in "update", the "local cache get" will not miss for most time. It is a trade-off to do so.

This whole fix is a workaround to current implementation, it is more an "observer" other than a "controller", that's why I try not to return error as normal controller, because this may lead to side-effect to current features.  There is no strict data consistency insurance, both upstream and downstream are not knowing this "observer".

L97 is also following this, leave a warning message for trouble-shooting, but still return nil.

pkg/controller/master/virtualmachine/register.go
w13915984028: @guangbochen :
**Conclusion: PVC/PV is not holding on "Pending", "Pending" is indeed the init/default status, will move to "Bound" quickly.**


The status `Pending` only exists in any PVC for a very short time, it will be updated to `Bound` quickly, no matter its size, it seems as long as LH creates the volume object, it will update PVC to `Bound`.

**The WebUI is also checking the "Bound" status, then display as "Ready"** , that is why it also changes quickly.

```
k get pvc v-2g -o JSON
    "status": {
        "phase": "Pending"  -- at beginning
    }
    
        "status": {
        "accessModes": [
            "ReadWriteMany"
        ],
        "capacity": {
            "storage": "2Gi"
        },
        "phase": "Bound"  -- move to here
    }

k get pvc vol-300g -o JSON
    "status": {
        "accessModes": [
            "ReadWriteMany"
        ],
        "capacity": {
            "storage": "300Gi"
        },
        "phase": "Bound"
    }
```

   
image


vol-300g, vol-2t are out of scope
image


k8s source code 

```
	// VolumePending means the Volume is pending and cannot be attached to the node yet.
	VolumePending VolumePhase = "Pending"
	// VolumeBound means the Volume is bound and can be attach to the node.
	VolumeBound VolumePhase = "Bound"

type PersistentVolumePhase string

const (
	// used for PersistentVolumes that are not available
	VolumePending PersistentVolumePhase = "Pending"
	// used for PersistentVolumes that are not yet bound
	// Available volumes are held by the binder and matched to PersistentVolumeClaims
	VolumeAvailable PersistentVolumePhase = "Available"
	// used for PersistentVolumes that are bound
	VolumeBound PersistentVolumePhase = "Bound"
```

vendor/k8s.io/kubernetes/pkg/apis/core/v1/defaults.go
```
func SetDefaults_PersistentVolume(obj *v1.PersistentVolume) {
	if obj.Status.Phase == "" {
		obj.Status.Phase = v1.VolumePending
	}
	if obj.Spec.PersistentVolumeReclaimPolicy == "" {
		obj.Spec.PersistentVolumeReclaimPolicy = v1.PersistentVolumeReclaimRetain
	}
	if obj.Spec.VolumeMode == nil {
		obj.Spec.VolumeMode = new(v1.PersistentVolumeMode)
		*obj.Spec.VolumeMode = v1.PersistentVolumeFilesystem
	}
}
func SetDefaults_PersistentVolumeClaim(obj *v1.PersistentVolumeClaim) {
	if obj.Status.Phase == "" {
		obj.Status.Phase = v1.ClaimPending
	}
}
```
PV, PVC status are same in "Bound"
image

guangbochen: Are u using the `master-head` Harvester version with embedded LH `v1.2.4-rc1`, this did not happen to me as creating an `out-of-range` disk will get it stuck on the pending status and its related volumes will not be created.
<details>
e.g 

```
$ ks get pvc
test-disk-0-6vxjb                        Pending                                                                        longhorn-image-kvsfj   29m
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    harvesterhci.io/imageId: default/image-kvsfj
    harvesterhci.io/owned-by: '[{"schema":"kubevirt.io.virtualmachine","refs":["default/test"]}]'
    volume.beta.kubernetes.io/storage-provisioner: driver.longhorn.io
  creationTimestamp: "2022-03-09T02:23:46Z"
  finalizers:
  - kubernetes.io/pvc-protection
  name: test-disk-0-6vxjb
  namespace: default
  resourceVersion: "4085568"
  uid: 70ee3e51-092a-4ae0-ac6e-f0b66c777b12
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: "9223372036854775807"
  storageClassName: longhorn-image-kvsfj
  volumeMode: Block
status:
  phase: Pending
```

</details>

w13915984028: longhornio/longhorn-manager:v1.2.4-rc1

The ISO is built per https://github.com/harvester/harvester-installer/pull/240, 
image

image

kl longhorn-manager-qnr6q -o JSON  | grep image
            ..
                "image": `"longhornio/longhorn-manager:v1.2.4-rc1",`
                "imagePullPolicy": "IfNotPresent",
                "image": "docker.io/longhornio/longhorn-manager:v1.2.4-rc1",
                "imageID": "sha256:d5f302f87a2db67c8d0acdc4abdde37be5efb0b84cb47b0d171969c73dc72a02",

```
k get pvc
NAME       STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
..
vol-2t     Bound    pvc-b6b977d8-1f9f-443d-8f26-dfb774b75d23   2000Gi     RWX            longhorn       9h
vol-300g   Bound    pvc-95a08266-c865-4282-a8df-78028d608482   300Gi      RWX            longhorn       9h
```

w13915984028: Update:
When creating Harvester Volume, if it is image-backed (Guangbo's case), it implicitly has 2 operations: create and write,  the over-sized volume will stuck in "Ponding" status;  if it is none-image related, the over-sized goes into "Bound" quickly.

w13915984028:
update:

LH will have a solution to update status/message to PVC, Harvester will get the real status from PVC directly. The Harvester (planned in v1.0.2) implementation will re-adapt with LH (planned in v1.2.5) solution .

w13915984028:
replaced by PR: https://github.com/harvester/harvester/pull/3869

Issue #1855, opened 2022-01-21
[BUG] Additional virtual machine and images created during zero downtime upgrade should be deleted after completion 


TachunLin:
**Describe the bug**

Additional virtual machine and images created and remains in harvester-system namespace after zero downtime upgrade 

* Additional virtual machine under harvester-system namespace
![image](https://user-images.githubusercontent.com/29251855/150509581-a79e2059-0380-4e02-b041-71e28426e433.png)

* Additional images created unser harvester-system namespace
![image](https://user-images.githubusercontent.com/29251855/150510207-308228c8-b67f-49fe-995e-3a4610828f2a.png)

**To Reproduce**
Do zero downtime upgrade from v1.0 to latest master head, steps refer to #1022 
1. Prepare 3 nodes harvester with 1.0.0 GA
2. Prepare latest master head iso 
3. Upload iso file to http or file server
4. ssh to harvester controller node 
5. Create the versions.yaml 
    ```
    cat > versions.yaml <<EOF
    apiVersion: harvesterhci.io/v1beta1
    kind: Version
    metadata:
      name: 1.0.2
      namespace: harvester-system
    spec:
      isoChecksum: 2765d89e5e8024347c81d36cc84b7168bf358a3a2f4da0c63164ad834a94bcbbaa13af4900aad4caa0a27550e4b4db5bf2c4726daa88664378e0c581a79ad2cb
      isoURL: http://10.10.0.1/harvester/harvester.iso
      minUpgradableVersion: 0.3.0
      releaseDate: "20211231"
      tags:
      - dev
      - test
    EOF
    ```
6. Run kubectl create -f versions.yaml
7. Refresh harvester page
8. Click the Upgrade button
9. Click the green circle to display upgrade status page 
10. Check the status and wait until zero downtime upgrade finish

**Expected behavior**

Additional virtual machine and image created during zero downtime upgrade should be deleted from harvester virtual machine and images page. 

**Support bundle**

[supportbundle_9f470b6f-44de-4cf9-891e-9bce00060bd7_2022-01-21T10-27-47Z.zip](https://github.com/harvester/harvester/files/7912467/supportbundle_9f470b6f-44de-4cf9-891e-9bce00060bd7_2022-01-21T10-27-47Z.zip)

**Environment:**
 - Harvester ISO version: v1.0.0 GA upgrade to v1.0 master head 
[...]

PR #1945, opened 2022-03-01
Remove upgrade repo VM and image when upgrade is done

bk201:
**Problem:**

Upgrade repo VM and image are not removed after finishing an upgrade.

**Solution:**

When an upgrade is done:
- Delete the upgrade repo VM.
- Remove the upgrade repo image. This can't be deleted immediately because upgrade repo VM's PVC still has reference to the image and Harvester webhook will forbid the deleting. We use the following approaches:
  - If PVC of the upgrade repo VM exists, we add the PVC as the OwnerReference to the repo image. Once the repo VM is deleted and VM's PVC is deleted, the repo image will be garbage collected.
  - If PVC of the upgrade repo VM doesn't exist, we delete the image immediately.

**Related Issue:**
https://github.com/harvester/harvester/issues/1855

**Test plan:**

For reviewer:
- Create a cluster. Either from master or 1.0.0 ISO.
- Build the backend and replace the harvester pod.
- This part is tricky, we need to tag a current-version harvester-upgrade image:
  - Get current version:
  ```
  $ kubectl get settings server-version
  NAME             VALUE
  server-version   2ff5a0cb-dirty
  ```
  - **On each node**
  ```
  $ docker tag rancher/harvester-upgrade:master-head rancher/harvester-upgrade:2ff5a0cb-dirty
  ```

- Create a version:
```
$ cat version.yaml
kind: Version
metadata:
  name: 1.0.2
  namespace: harvester-system
spec:
  isoChecksum: xxx                              <-- the build ISO sha512 checksum
  isoURL: http://10.10.0.1/harvester/harvester.iso      <-- the build ISO
  minUpgradableVersion: 0.3.0
  releaseDate: "20211231"
  tags:
  - dev                 <--- `dev` tag is important
  - test

 $ kubectl create -f version.yaml
```

- Create an upgrade:
```
$ cat upgrade.yaml
kind: Upgrade
metadata:
  generateName: hvst-upgrade-
  namespace: harvester-system
spec:
  version: "1.0.2"

$ kubectl create -f upgrade.yaml
```

- After upgrade is done, there should be no VM and image exists in `harvester-system` namespace.

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_repo.go
+++ b/pkg/controller/master/upgrade/upgrade_repo.go
@@ -10,8 +10,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -136,8 +138,12 @@ func (r *UpgradeRepo) GetImage(imageName string) (*harvesterv1.VirtualMachineIma
 	return image, nil
 }
 
+func (r *UpgradeRepo) getVMName() string {
+	return fmt.Sprintf("%s%s", repoVMNamePrefix, r.upgrade.Name)
+}
+
 func (r *UpgradeRepo) createVM(image *harvesterv1.VirtualMachineImage) (*kubevirtv1.VirtualMachine, error) {
-	vmName := fmt.Sprintf("%s%s", repoVMNamePrefix, r.upgrade.Name)
+	vmName := r.getVMName()
 	vmRun := true
 	var bootOrder uint = 1
 	evictionStrategy := kubevirtv1.EvictionStrategyLiveMigrate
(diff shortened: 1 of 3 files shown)


pkg/controller/master/upgrade/upgrade_repo.go
```diff
@@ -303,6 +309,69 @@ func (r *UpgradeRepo) createVM(image *harvesterv1.VirtualMachineImage) (*kubevir
 	return r.h.vmClient.Create(&vm)
 }
 
+func (r *UpgradeRepo) deleteVM() error {
+	vmName := r.getVMName()
+
+	vm, err := r.h.vmCache.Get(upgradeNamespace, vmName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
```
FrankYang0529: Not sure whether I understand correctly. If we can't find a VM here, we don't delete the image directly as the solution.

bk201: You are correct, I'll fix it. Thanks.

bk201: Fixed.

FrankYang0529 (approved):
LGTM

bk201:
Conflict is resolved.

pkg/controller/master/upgrade/upgrade_repo.go
```diff
@@ -303,6 +309,74 @@ func (r *UpgradeRepo) createVM(image *harvesterv1.VirtualMachineImage) (*kubevir
...
+
+	logrus.Infof("Delete upgrade repo VM %s/%s", vm.Namespace, vm.Name)
+	return r.h.vmClient.Delete(vm.Namespace, vm.Name, &metav1.DeleteOptions{})
+}
+
+// deleteImage deletes a repo image
+// (1) If a PVC based on the image exists, it adds the PVC as the OwnerReference to the repo image.
+// Once the repo VM is deleted and VM's PVC is deleted, the repo image will be garbage collected.
+// (2) If there is no PVCs based on the repo image. We delete the image directly.
+func (r *UpgradeRepo) deleteImage(pvcName string) error {
```
guangbochen, line 337: Maybe we can keep the image as the upgrade may fail or they can delete it manually?

bk201: I'm good to keep it.
Just the original issue states the image needs to be deleted as well. @TachunLin any takes?

TachunLin: Checked the upgrade image can't be used on for normal re-upgrade user behavior. 
I would suggest to delete the image to save some space and clean up the Volumes page.

bk201: @guangbochen As David pointed out,  the kept image can't be used by users directly. With normal UX, the user still needs to trigger an upgrade from a remote ISO. Thus the image needs to be downloaded again.

guangbochen (approved):
LGTM, @johnliu55tw @weihanglo can one of u guys take a review, thanks.

weihanglo (approved):
LGTM.

Issue #1952, opened 2022-03-02
[BUG] unable to setup NFS backup-target


lanfon72:
### Describe the bug
Unable to setup **NFS** backup-target on dashboard. (error message as the snapshot)
Able to mount the backup-target in console.
![image](https://user-images.githubusercontent.com/5169694/156313697-8719dad4-8531-4466-b9bb-1d83a88f4ced.png)

### To Reproduce
1. Install Harvester with any node
2. Login to dashboard and setup backup-target with **NFS** type

### Expected behavior
setup successfully.

## Environment:
 - Harvester ISO version: **master-e214d829-head**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **qemu/KVM 3 nodes**

### Additional context
Function works on other versions:
- [x] v1.0.0
- [x] master-bfd6a8a9-head

### Support bundle
N/A

bk201:
This might be related to the [BCI image change](https://github.com/harvester/harvester/pull/1910).

johnliu55tw:
The container for running the webhook is missing the package `nfs-client`, thus the command failed.

bk201:
@johnliu55tw Does this mean we can remove the `nfs-client` package in the backend container image? 
It's a bit weird before switching to BCI base image, we seems not install NFS stuff in the webhook container image: https://github.com/harvester/harvester/commit/9c2461f610f3f984f2f1887ffe94db71303a9ee3#diff-49ccf96bb66e42627b495efc688852146c1f0aca84f508c6dc67d97e55c91198

johnliu55tw:
@bk201 The `mount` command in the old image can mount NFS. I think busybox supports that.
```
# busybox mount -t nfs -o nfsvers=4.2 -o actimeo=1 192.168.122.1:/share /mnt
# mount | grep 192.168.122.1
192.168.122.1:/share on /mnt type nfs4 (rw,relatime,vers=4.2,rsize=1048576,wsize=1048576,namlen=255,acregmin=1,acregmax=1,acdirmin=1,acdirmax=1,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=10.4.0.3,local_lock=none,addr=192.168.122.1)
```

As for whether backend container needs the `nfs-client`, I need to do some further checks.

lanfon72:
Verified this bug has been fixed.


Test Information
----
* Environment: **qemu/KVM single node**
* Harvester Version: **master-dba85940-head**
[...]

PR #1957, opened 2022-03-03
Add missing package for webhook container image

johnliu55tw:

**Problem:**

The container for running webhook is missing package "nfs-client", causing the validation of NFS backup target to fail.

**Solution:**
Add package "nfs-client".

**Related Issue:**
- https://github.com/harvester/harvester/issues/1952

**Test plan:**
1. Set up NFS backup target and make sure no error is reported.


FrankYang0529 (approved):
LGTM

bk201 (approved):
Confirm the backup store can be added and used.
Thanks for the fix.

Issue #1730, opened 2021-12-21
[BUG] Windows Server ISOs reboot system before getting to the installation screen


Yabbo:
**Describe the bug**
When trying to install a Windows server from ISO or QCOW files on both Harvester 1.0.0 and 0.3.0 the windows ISO will start to boot up and get to the windows logo with the spinning circles and then the system will reboot before any menu screens come up.

**To Reproduce**
1. boot from any windows server ISO (2021, 2019 R2, 2012) 

**Expected behavior**
Boot into the installer and allow installation of windows server

**Support bundle**
[supportbundle_15168620-813b-473f-b970-920ec49356b8_2021-12-21T17-05-17Z.zip](https://github.com/harvester/harvester/files/7756925/supportbundle_15168620-813b-473f-b970-920ec49356b8_2021-12-21T17-05-17Z.zip)

**Environment:**
 - Harvester ISO version: 1.0.0 and 0.3.0
 - Underlying Infrastructure Baremetal 2 node cluster with 10G nics 

**Additional context**
I can build and install Ubuntu and Arch based systems with no issue on this cluster

yasker:
I’ve just tried Windows 2012 R2 and it works. Not sure if it’s related but I set my root disk to 50G and give it 4c8g./

I do have issues with Windows 2019 at the moment. The VM keeps rebooting. Might be related to https://github.com/harvester/harvester/issues/276

FrankYang0529:
I can't reproduce this issue with Harvester `master-a67f96d-head` by following steps:

* Launch instance from `windows-iso-image-base-version` template.
* Give a VM name.
* Click `Volumes` tab.
  * Choose a windows image in `cdrom-disk`.
  * Volume Type will auto change to disk. Change it back to `cd-rom`.
* Click Create.
* Open in Web VNC for the VM.
* Choose `Custom: Install Windows only`.
![Screen Shot 2022-01-28 at 4 48 21 PM](https://user-images.githubusercontent.com/4927361/151515792-f2259f00-5aa4-4949-bed7-e38b71464e24.png)
* Click `Load driver`.
![Screen Shot 2022-01-28 at 4 49 56 PM](https://user-images.githubusercontent.com/4927361/151515985-dbf4af65-eb55-4668-8d70-c52bc3e1a71d.png)
* Click `Browse`.
* Choose driver under `vmdp-200205_virt` disk.
  * For WIndows 2022 & 2019, install `win 10-2019/x64/pvvx`.
[...]

PR #1965, opened 2022-03-04
Add reserve memory annotation

FrankYang0529:
**Problem:**
We hard code reserved memory in webhook. In some OS like Windows, we may need more reserved memory, but others may not. It is hard to estimate, so we choose to let users determine how much memory they want to reserve.

**Solution:**
Add an annotation `harvesterhci.io/reservedMemory`. Users can input how much memory they would like to reserve and we use `limits.memory - reservedMemory` as guest memory.

**Related Issue:**
https://github.com/harvester/harvester/issues/1730

**Test plan:**

1. Reserved memory can't be less than 0.
* Set `harvesterhci.io/reservedMemory` as `-20Mi`. Webhook will show an error message.
2. Reserved memory can't be equal or greater than `limits.memory`.
* Set `harvesterhci.io/reservedMemory` as same as `limits.memory` or greater than `limits.memory`. Webhook will show an error message.
3. Guest memory should be equal to `limits.memory - reservedMemory`
* Set `limits.memory` as `1Gi` and `harvesterhci.io/reservedMemory` as `300Mi`. We should get `memory.guest` as `724Mi`.
4. Default reserved memory is `100Mi`.
* Set `limits.memory` as `1Gi` and don't set `harvesterhci.io/reservedMemory`. We should get `memory.guest` as `924Mi`.

### DIFF
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -11,6 +11,7 @@ const (
 	AnnotationTimestamp            = prefix + "/timestamp"
 	AnnotationVolumeClaimTemplates = prefix + "/volumeClaimTemplates"
 	AnnotationImageID              = prefix + "/imageId"
+	AnnotationReservedMemory       = prefix + "/reservedMemory"
 	AnnotationHash                 = prefix + "/hash"
 
 	BackupTargetSecretName      = "harvester-backup-target-secret"
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -13,6 +13,7 @@ import (
 
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/settings"
+	"github.com/harvester/harvester/pkg/util"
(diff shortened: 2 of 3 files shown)


guangbochen (approved):
LGTM

weihanglo (approved):
Tested and LGTM 👍🏾

pkg/util/constants.go
```diff
@@ -11,6 +11,7 @@ const (
 	AnnotationTimestamp            = prefix + "/timestamp"
 	AnnotationVolumeClaimTemplates = prefix + "/volumeClaimTemplates"
 	AnnotationImageID              = prefix + "/imageId"
+	AnnotationReservedMemory       = prefix + "/reservedMemory"
```
w13915984028, line 14: As discussed with @yasker before, he preferred the annotations to be with format of `reserved-memory`

yasker: It seems the recommendation from K8s is `camel-case`, though I will leave it up to the engineering team to decide. e.g. https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

guangbochen: We can file a new issue to track those types of issues, for now, we can keep it for better consistency.

Issue #1810, opened 2022-01-13
[FEATURE] Support update the dependency charts parameters in the harvester managedcharts


yaocw2020:
**Is your feature request related to a problem? Please describe.**

During upgrading, the Harvester installer only allows updating the harvester chart version, not the dependency chart parameters. We should support it.

https://github.com/harvester/harvester/blob/ebde990de70056a975a7e6be389cbd594adf9fca/package/upgrade/upgrade_manifests.sh#L85-L91

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Chart values to update**
- [ ] Removal of kube-vip values in harvester/harvester-installer@f8a4c77
- [ ] Add defaultDataPath: "/var/lib/harvester/defaultdisk". Note this doesn't have any effect, just to be consistent.

bk201:
For a context, this issue is used to track the removal of kube-vip values in https://github.com/harvester/harvester-installer/pull/216/commits/f8a4c77faa1ec87a1c3338b42ddc047b3956016a

TachunLin:
Verified **fixed** on one click upgrade to v1.0.1-rc2. Close this issue.

### Result
1. **`kubevip.config`** did not exists
2.  Add defaultDataPath: longhorn.defaultSettings.defaultDataPath = **"/var/lib/harvester/defaultdisk"**

  `kubectl get managedcharts harvester -n fleet-local -o yaml | yq e '.spec.values' -`

  ```
  harvester-x4js2:~ # kubectl get managedcharts harvester -n fleet-local -o yaml | yq e '.spec.values' -
  containers:
    apiserver:
      hciMode: true
      image:
        imagePullPolicy: IfNotPresent
  harvester-load-balancer:
    enabled: true
  harvester-network-controller:
    enabled: true
    image:
      pullPolicy: IfNotPresent
    vipEnabled: true
  harvester-node-disk-manager:
    enabled: true
    image:
      pullPolicy: IfNotPresent
    labelFilter:
      - COS_*
      - HARV_*
  kube-vip:
    enabled: true
  kube-vip-cloud-provider:
    enabled: true
  kubevirt:
    spec:
      monitorAccount: rancher-monitoring-operator
      monitorNamespace: cattle-monitoring-system
  longhorn:
    defaultSettings:
      defaultDataPath: /var/lib/harvester/defaultdisk
      taintToleration: kubevirt.io/drain:NoSchedule
    enabled: true
  multus:
    enabled: false
[...]

PR #2007, opened 2022-03-14
Upgrade: allow adding version-specific changes to ManagedCharts

bk201:
**Problem:**

Sometimes when we upgrade from a previous version, we need to alter `values` in ManagedCharts.

**Solution:**

Provide a hook mechanism to allow changing values.

Say you need to do something to a ManagedChart if the previous version is `v1.0.0`.
You can create a file `v1.0.0.sh` in `./managed_charts/` directory to apply the fixes you need.

If the upgrade_manifests.sh script detects the system is upgraded from `v1.0.0`, it will call the script with:

```
v1.0.0.sh <chart_name> <chart_manifest>
```

- The `<chart_name>` could be `harvester`, `harvester-crd`, `rancher-monitoring`, etc.
- The `<chart_manifest>` is the manifest file that contains the current managed chart dump.

**Related Issue:**
https://github.com/harvester/harvester/issues/1810

**Test plan:**
- Upgrade from v1.0.0 to master-head.
- Dump the harvester ManagedChart, there should be no changes:
  - `spec.values.kube-vip.config` field is removed.
  - `.spec.values.longhorn.defaultSettings.defaultDataPath` should be `"/var/lib/harvester/defaultdisk"`.

### DIFF
--- a/package/upgrade/migrations/managed_charts/v1.0.0.sh
+++ b/package/upgrade/migrations/managed_charts/v1.0.0.sh
@@ -1,0 +1,24 @@
+#!/bin/bash -ex
+
+CHART_NAME=$1
+CHART_MANIFEST=$2
+
+remove_vip_config()
+{
+  # VIP interface isn't needed after kube-vip v0.4.1 (https://github.com/harvester/harvester-installer/pull/216)
+  yq e 'del(.spec.values.kube-vip.config)' $CHART_MANIFEST -i
+}
+
+
+add_longhorn_settings()
+{
+  # Add config "DataDisk" to specify the default Longhorn partition (https://github.com/harvester/harvester-installer/pull/217)
+  yq e '.spec.values.defaultDataPath.defaultSettings.defaultDataPath = "/var/lib/harvester/defaultdisk"' $CHART_MANIFEST -i
+}
+
+case $CHART_NAME in
+  harvester)
+    remove_vip_config
+    add_longhorn_settings
+    ;;
+esac
--- a/package/upgrade/migrations/managed_charts/v1.0.0.sh
+++ b/package/upgrade/migrations/managed_charts/v1.0.0.sh
@@ -13,7 +13,7 @@ remove_vip_config()
 add_longhorn_settings()
 {
   # Add config "DataDisk" to specify the default Longhorn partition (https://github.com/harvester/harvester-installer/pull/217)
-  yq e '.spec.values.defaultDataPath.defaultSettings.defaultDataPath = "/var/lib/harvester/defaultdisk"' $CHART_MANIFEST -i
+  yq e '.spec.values.longhorn.defaultSettings.defaultDataPath = "/var/lib/harvester/defaultdisk"' $CHART_MANIFEST -i
 }
 
 case $CHART_NAME in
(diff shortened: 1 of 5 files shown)


package/upgrade/version_hacks/managed_charts/v1.0.0.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -ex
+
+CHART_NAME=$1
+CHART_MANIFEST=$2
+
+remove_vip_config()
+{
+  # VIP interface isn't needed after kube-vip v0.4.1 (https://github.com/harvester/harvester-installer/pull/216)
+  yq e 'del(.spec.values.kube-vip.config)' $CHART_MANIFEST -i
```
bk201: cc @yaocw2020 , this will remove the unused `kube-vip.config`.

weihanglo:
Just wonder why not call it `version_migrations` but `version_hacks`?

bk201:
> Just wonder why not call it `version_migrations` but `version_hacks`?

That's indeed a good idea!

package/upgrade/version_hacks/README.md
```diff
@@ -0,0 +1,15 @@
+The directory contains drop-ins to fix things when upgrading from a previous version. 
+
+# `./managed_charts/`
+
+Say you need to do something to a ManagedChart if the previous version is `v1.0.0`.
+You can create a file `v1.0.0.sh` in `./managed_charts/` directory to apply the fixes you need.
+
+If the upgrade_manifests.sh script detects the system is upgraded from `v1.0.0`, it will call the script with:
```
weihanglo: I found the version here only referring to Harvester version, right? Say that other managedchart might have a version other than v1.0.0 (e.g. rancher-monitoring) and but the script 1.0.0.sh cannot be distinguished whether it's for harveter or other managedchart. 

Anyway I think it's not urgent. We can ignore this at this moment.

bk201: https://github.com/harvester/harvester/blob/e34c2404f118a288a37579899b800d1f6b9a2cc3/package/upgrade/version_hacks/managed_charts/v1.0.0.sh#L19-L20

CHART_NAME can be used to distinguish different components.

bk201:
@weihanglo I just use the word `migrations`. Just like the traditional DB schema migrations directory.

package/upgrade/migrations/managed_charts/v1.0.0.sh
```diff
@@ -0,0 +1,24 @@
...
+{
+  # VIP interface isn't needed after kube-vip v0.4.1 (https://github.com/harvester/harvester-installer/pull/216)
+  yq e 'del(.spec.values.kube-vip.config)' $CHART_MANIFEST -i
+}
+
+
+add_longhorn_settings()
+{
+  # Add config "DataDisk" to specify the default Longhorn partition (https://github.com/harvester/harvester-installer/pull/217)
+  yq e '.spec.values.defaultDataPath.defaultSettings.defaultDataPath = "/var/lib/harvester/defaultdisk"' $CHART_MANIFEST -i
```
weihanglo:
```suggestion
  yq e '.spec.values.longhorn.defaultSettings.defaultDataPath = "/var/lib/harvester/defaultdisk"' $CHART_MANIFEST -i
```

bk201: Thanks!

weihanglo (approved):
Confirm that values in managedchart would be changed. LGTM

johnliu55tw (approved):
Verified that the ManagedChart value is updated.

Issue #1861, opened 2022-01-24
[FEATURE] Upgrade after v1.0.0


TachunLin:
The epic issue to continue tracking zero downtime upgrade related issues and test plan after v1.0.0 release
This extend the development and testing status from #1022 to perform more testing in different stages to fully check the zero downtime on bumped rke2, rancher and longhorn version. 

## pre-1.0.0 dependencies
- [x] https://github.com/longhorn/longhorn/issues/3234 (Enhancement included in `v1.2.3-rc1`)
- [ ] [Move to `v1.0.2`] https://github.com/harvester/harvester/issues/1707 
  - (For `one node` zero downtime upgrade image from `URL`)
- [x]  [Fixed in Rancher, need to test more times] https://github.com/harvester/harvester/issues/1717 
  - (For `four nodes` zero downtime upgrade image from `File`)

## v1.0.0 dependencies
- [ ] https://github.com/harvester/docs/pull/131
- [x] https://github.com/harvester/harvester/issues/1852
- [x] https://github.com/harvester/harvester/issues/1855
- [x] https://github.com/harvester/harvester/issues/1810
- [x] https://github.com/harvester/harvester/pull/1832
- [ ] ~~Remove settings `vlan`: see https://github.com/harvester/harvester/issues/1739~~
- [x] https://github.com/rancher/rancher/issues/36536
- [x] https://github.com/harvester/harvester/issues/1982
- [x] [Fixed in Rancher] https://github.com/harvester/harvester/issues/1984
- [x] https://github.com/harvester/harvester/issues/1985

## Known issues
- [x] https://github.com/harvester/harvester/issues/1983
- [x] https://github.com/harvester/harvester/issues/2029
- [x] https://github.com/harvester/harvester/issues/2053
- [x] https://github.com/harvester/harvester/issues/2011
- [x] https://github.com/harvester/harvester/issues/2114
- [x] https://github.com/harvester/harvester/issues/2119

**Additional context**

Please refer to #1635 for upgrade design and test execution steps

bk201:
I drop the item:
- Remove settings vlan: see https://github.com/harvester/harvester/issues/1739

The setting `vlan` doesn't exist in the previous version, it's a misuse of the installer config. We have [the following installer PR](https://github.com/harvester/harvester-installer/pull/226) to avoid this happening in the future.

bk201:
Close this epic one, we can track upgrade issues with `area/upgrade-related` label and release.

PR #2019, opened 2022-03-16
Don't compare upgrade to image to determine an update

weihanglo:
**Problem:**
Wrong comparision between upgrade and image. 

**Solution:**
Should compare the old and the new upgrade object instead

**Related Issue:**
#1861 

**Test plan:**


bk201 (approved):
Nice finding!

guangbochen (approved):
LGTM

FrankYang0529 (approved):
LGTM

Issue #2033, opened 2022-03-18
[BUG] Kube-vip image version should be v0.4.2.


yaocw2020:
**Describe the bug**

Kube-vip image version should be v0.4.2.

**Environment:**
 - Harvester ISO version: v1.0.1 rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Baremetal with Dell PowerEdge R630

**Additional context**

TachunLin:
Current Kube-vip image version on harvester `master-9285ca26-head` (03/25) is `kube-vip:main` 
https://github.com/kube-vip/kube-vip/pkgs/container/kube-vip

```
harvester-node01-220325:~ # kubectl get pods kube-vip-lgv9q -n harvester-system -o json | grep image
                "image": "ghcr.io/kube-vip/kube-vip:main",
                "imagePullPolicy": "IfNotPresent",
                "image": "ghcr.io/kube-vip/kube-vip:main",
                "imageID": "sha256:1c402444bab2d634b4fb2dccfb8f1652b29e5c8b58daa149e80de82b16c76857",
``` 

According to previous discussion, plan to verify this issue on v1.0.1-rc2 

Test plan would cover:
1.  https://harvester.github.io/tests/manual/network/disable-and-enable-vlan-cluster-network/ 
2.  Install 3 nodes harvester cluster use DHCP VIP
3.  Install 3 nodes harvester cluster use Fixed VIP
4.  Reboot and shutdown cluster nodes scenarios

guangbochen:
We have bumped the `kube-vip` image version to `main` to include a PR https://github.com/kube-vip/kube-vip/pull/383 which will help fix a known issue https://github.com/harvester/harvester/issues/1985. If the VIP team doesn't cut a newer version, then we'll just use the specific master tag `344cb491eb84c0a8dc73e54fa043ebf4ea44bf02` instead.

TachunLin:
Verified fixed on Harvester `v1.0.1-rc2`. Closet this issue.

* Current Kube-vip image version on harvester v1.0.1-rc2 is `kube-vip:main`
```
harvester-node02-101rc2:~ # kubectl get pods kube-vip-8xrvh -n harvester-system -o json | grep image
                "image": "ghcr.io/kube-vip/kube-vip:main",
                "imagePullPolicy": "IfNotPresent",
                "image": "ghcr.io/kube-vip/kube-vip:main",
[...]

PR #2034, opened 2022-03-18
Update kube-vip image version

yaocw2020:

**Related Issue:** https://github.com/harvester/harvester/issues/2033

**Test plan:**

- Spin up Harvester with v1.0.1 rc1
- Enable VLAN with default network interface `harvester-mgmt`
- The kube-vip pod will be crash and restart. Execute `kubectl logs -p -n harvester-system <kube-vip-pod>` to see the log like that
```
time="2022-03-18T02:42:28Z" level=fatal msg="crash: default route deleted and the default interface may be invalid"
```
- Disable VLAN. The kube-vip pod will be crash and restart

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -429,7 +429,7 @@ kube-vip:
   enabled: false
   image:
     repository: ghcr.io/kube-vip/kube-vip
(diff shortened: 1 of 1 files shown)


guangbochen (approved):
LGTM

PR #2035, opened 2022-03-18
version script: unset the DIRTY variable because it might be set

bk201:
The `version` script is [sourced by harvester-installer](https://github.com/harvester/harvester-installer/blob/ededb753e24b2b345c7643da2a66b9912cbbf5c8/scripts/package-harvester-os#L11-L15) and if harvester-installer repo is dirty, harvester repo will also
be treated as dirty. Unset the DIRTY variable to avoid this.

### DIFF
--- a/scripts/version
+++ b/scripts/version
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+unset DIRTY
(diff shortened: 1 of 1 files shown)


futuretea (approved):
LGTM

guangbochen (approved):

PR #2039, opened 2022-03-18
Push image if set env PUSH

futuretea:

**Problem:**

in harvester repo, `make build-iso` would fail if the current branch is not master

**Solution:**

add replace_image_repo function to https://github.com/harvester/harvester-installer/blob/master/scripts/patch-harvester and add docker push to https://github.com/harvester/harvester/blob/master/scripts/package, https://github.com/harvester/harvester/blob/master/scripts/package-upgrade, https://github.com/harvester/harvester/blob/master/scripts/package-webhook
After that, we can use the following command to build a custom ISO using a local docker registry, such as harbor
```
export REPO=harbor.futuretea.me/rancher
export PUSH=true
make
make build-iso
```

**Related Issue:**
https://github.com/harvester/harvester/issues/2048

**Dependent PR:**
https://github.com/harvester/harvester-installer/pull/249

**Test plan:**

build docker images and iso successfully by the following commands, (replace `harbor.futuretea.me`  with your container registry URL)
```bash
docker login harbor.futuretea.me
git clone --branch local-build git@github.com:futuretea/harvester.git
cd harvester
cp ~/.docker . -r
export REPO=harbor.futuretea.me/rancher
export PUSH=true
make
make build-iso
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /harvester
 .DS_Store
 api.test
+.docker
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -66,7 +66,7 @@ COPY --from=makeiso /usr/bin/luet-makeiso /usr/bin/luet-makeiso
 # -- for dapper
 ENV DAPPER_RUN_ARGS --privileged --network host -v /run/containerd/containerd.sock:/run/containerd/containerd.sock
 ENV GO111MODULE off
-ENV DAPPER_ENV REPO TAG DRONE_TAG DRONE_BRANCH CROSS GOPROXY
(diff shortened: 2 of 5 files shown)


johnliu55tw (approved):
Nice! Now I can build an ISO using my own images :+1: 

FWIW, it's also possible to use public container registries such as Docker Hub to build the ISO:
```
cp -r ~/.docker .
export REPO=johnliu55tw
export PUSH=true
make
make build-iso
```
But do be careful that the directory `.docker` contains the credential to access the public container registry.

weihanglo (approved):
LGTM. Thanks!

Maybe this should be written into in dev documentation or somewhere.

guangbochen (approved):

Issue #2037, opened 2022-03-18
[BUG] report internal error validator when using runStrategy: Manual


tjjh89017:
**Describe the bug**

Create a VM with `runStrategy: Manual`
and remove `running: true`
It will create a stopped VM.
Press start button to start, it will report an internal error in webhook validator.

**To Reproduce**
1. create VM with `runStrategy: Manual` and remove `running: true` in YAML
2. start the VM

**Expected behavior**

It should run properly.

**Support bundle**

**Environment:**
 - Harvester ISO version: 1.0.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

**Additional context**

**Test plan**

test case 1:
1. create VM with runStrategy: Manual and remove running: true in YAML
2. start the VM
3. it should run properly without error.

test case 2:
1. create VM with runStrategy: Always and remove running: true in YAML
2. run properly and no error

test case 3:
1. create VM with runStrategy: Always and remove running: true in YAML
2. start the VM
3. start VM backup
4. during backup, press stop or restart
5. should report error "there is vmbackup in progress for vm <NAMEAPCE>/<VM NAME>, please wait for the vmbackup or remove it before stop/restart the vm"
6. stop or restart should be canceled.

lanfon72:
Verified this bug has been fixed.

- [x] `spec.runStrategy: Manual` and ~~`spec.running:true`~~
- [x] `spec.runStrategy: Always` and ~~`spec.running:true`~~
- [x] `spec.runStrategy: Always` and ~~`spec.running:true`~~ and `restart/stop` the VM when backup is processing
![image](https://user-images.githubusercontent.com/5169694/159662507-965a4d22-9a0d-4edb-ab49-ed8597dfae66.png)


Test Information
----
* Environment: **qemu/KVM 2 nodes**
* Harvester Version: **master-2a657038-head**
* **ui-source** Option: **Auto**

### Verify Steps:
- Follow **Test Plan** in https://github.com/harvester/harvester/issues/2037#issue-1173296996

PR #2050, opened 2022-03-21
webhook: vm: validator: refactor stopping check

tjjh89017:
check pointer is nil or not before deference
check status with runStrategy rather than running

**Problem:**

Could not change running to runStrategy via YAML
it will have nil pointer dereference problem.

**Solution:**

copy running status to runStrategy to determine 

**Related Issue:**
https://github.com/harvester/harvester/issues/2037

**Test plan:**

test case 1:
1. create VM with runStrategy: Manual and remove running: true in YAML
2. start the VM
3. it should run properly without error.

test case 2:
1. create VM with runStrategy: Always and remove running: true in YAML
2. run properly and no error

test case 3:
1. create VM with runStrategy: Always and remove running: true in YAML
2. start the VM
3. start VM backup
4. during backup, press stop or restart
5. should report error "there is vmbackup in progress for vm <NAMEAPCE>/<VM NAME>, please wait for the vmbackup or remove it before stop/restart the vm"
6. stop or restart should be canceled.

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -79,12 +79,7 @@ func (v *vmValidator) Update(request *types.Request, oldObj runtime.Object, newO
 	}
 
 	// Prevent users to stop/restart VM when there is VMBackup in progress.
-	// KubeVirt send stop request or change running from true to false when users stop a VM.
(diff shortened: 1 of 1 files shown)


FrankYang0529 (approved):
LGTM

guangbochen (approved):
LGTM

PR #2058, opened 2022-03-22
Add node exporter limit migration

bk201:
The node exporter limit is changed in https://github.com/harvester/harvester-installer/pull/250, this PR makes sure the values are also applied when upgrading from v1.0.0.

**Test plan**

- Perform an upgrade
- Check ManagedChart `rancher-monitoring`, values should reflect:
```
$ kubectl get managedchart rancher-monitoring -n fleet-local -ojsonpath="{.spec.values.prometheus-node-exporter}" | jq
{
  "resources": {
    "limits": {
      "cpu": "200m",
      "memory": "180Mi"           <-- previously, it's `50Mi`
    },
    "requests": {
      "cpu": "100m",
      "memory": "30Mi"
    }
  }
}
```

```
╰─$  kubectl get managedchart rancher-monitoring -n fleet-local -ojsonpath="{.spec.values.prometheus.prometheusSpec.resources}" | jq

{
  "limits": {
    "cpu": "1000m",
    "memory": "2500Mi"
  },
  "requests": {
    "cpu": "750m",
    "memory": "1750Mi"
  }
}
```

- Check `prometheus-rancher-monitoring-prometheus-0` pod and `rancher-monitoring-prometheus-node-exporter-xxx` pods, their limit should change too.

### DIFF
--- a/package/upgrade/migrations/managed_charts/v1.0.0.sh
+++ b/package/upgrade/migrations/managed_charts/v1.0.0.sh
@@ -16,9 +16,21 @@ add_longhorn_settings()
   yq e '.spec.values.longhorn.defaultSettings.defaultDataPath = "/var/lib/harvester/defaultdisk"' $CHART_MANIFEST -i
 }
 
+adjust_prometheus_resources()
+{
+  # Increate prometheus pod limit and request (https://github.com/harvester/harvester-installer/pull/240)
+  yq e '.spec.values.prometheus.prometheusSpec.resources = {"limits": {"cpu": "1000m", "memory": "2500Mi"}, "requests": {"cpu": "750m", "memory": "1750Mi"}}' $CHART_MANIFEST -i
+
+  # Increase limits.memory (https://github.com/harvester/harvester-installer/pull/250)
+  yq e '.spec.values.prometheus-node-exporter.resources = {"limits": {"cpu": "200m", "memory": "180Mi"}, "requests": {"cpu": "100m", "memory": "30Mi"}}' $CHART_MANIFEST -i
+}
+
 case $CHART_NAME in
   harvester)
     remove_vip_config
     add_longhorn_settings
+    ;;
+  rancher-monitoring)
+    adjust_prometheus_resources
     ;;
 esac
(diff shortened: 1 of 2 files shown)


package/upgrade/migrations/managed_charts/v1.0.0.sh
```diff
@@ -16,9 +16,18 @@ add_longhorn_settings()
   yq e '.spec.values.longhorn.defaultSettings.defaultDataPath = "/var/lib/harvester/defaultdisk"' $CHART_MANIFEST -i
 }
 
+adjust_node_exporter_limits()
+{
+  # Increase limits.memory (https://github.com/harvester/harvester-installer/pull/250)
+  yq e '.spec.values.prometheus-node-exporter.resources = {"limits": {"cpu": "200m", "memory": "180Mi"}, "requests": {"cpu": "100m", "memory": "30Mi"}}' $CHART_MANIFEST -i
+}
+
```
guangbochen: The Prometheus resource config is also changed per https://github.com/harvester/harvester-installer/pull/240/files, also just wondering if we can build a reusable mechanism through the upgrade controller instead of dealing it case by case, e.g compare the latest manifest YAML with the applied managedChart and patch the values accordingly?

bk201: These settings are recorded in rancherd bootstrap resources templates and applied by rancherd during bootstrap.
It's not visible to the backend controller.
A better way to prevent this kind of hotfix is to adjust those settings within charts' default values files. Since these PRs adjust the override values in the ManagedChart, we can only patch the ManagedChart spec.values to be consistent.

weihanglo: If we want to go in this way, we might want to reuse Helm's 3-way merge or `kubectl patch` merge strategy. And that makes me feel the responsibility should be on managedchart side? Not sure about that.

bk201: @guangbochen Add [Prometheus pod change](https://github.com/harvester/harvester/compare/950dde99b8ae63a5e07330fc7883c53199d8bb9e..80eed1b3549d9f0744b1e03235bb4671dec125e7) as well.

weihanglo (approved):
LGTM

johnliu55tw (approved):
LGTM

Issue #2029, opened 2022-03-17
[BUG] Upgrade: VMs fail to live-migrate to other hosts in some cases


bk201:
**Describe the bug**

The destination pod of a live-migrating VM continues erroring. The migration can never succeed.

**To Reproduce**
1. Create a v1.0.0 cluster.
2. Upgrade to a custom ISO.

I saw the upgrade-repo VM can't be live-migrated:
<img width="1871" alt="Screen Shot 2022-03-18 at 12 59 02 AM" src="https://user-images.githubusercontent.com/1691518/158855767-7e1019ab-c5cc-40bb-86fe-e6d1be159595.png">

The failure happens on the second migration of VM, the first migration succeeds.
- First migration (OK):
  - source pod image: `registry.suse.com/suse/sles/15.3/virt-launcher:0.45.0-8.4.3`
  - destination pod image: `registry.suse.com/harvester-beta/virt-launcher:0.49.0-2`
- Second migration and rest (Failure):
  - source pod image:`registry.suse.com/harvester-beta/virt-launcher:0.49.0-2`
  - destination pod image: `registry.suse.com/harvester-beta/virt-launcher:0.49.0-2` 

Destination pod fails with:
```
{"component":"virt-launcher","kind":"","level":"info","msg":"Prepared migration target pod","name":"upgrade-repo-hvst-upgrade-pxs8j","namespace":"harvester-system","pos":"server.go:171","timestamp":"2022-03-17T16:44:16.138343Z","uid":"5be1bfa7-df6a-4ef2-be38-c6b8e7d91857"}
{"component":"virt-launcher","kind":"","level":"info","msg":"Signaled target pod virt-launcher-upgrade-repo-hvst-upgrade-pxs8j-fkmcj to cleanup","name":"upgrade-repo-hvst-upgrade-pxs8j","namespace":"harvester-system","pos":"server.go:151","timestamp":"2022-03-17T16:44:16.941232Z","uid":"5be1bfa7-df6a-4ef2-be38-c6b8e7d91857"}
{"component":"virt-launcher","kind":"","level":"info","msg":"Signaled target pod virt-launcher-upgrade-repo-hvst-upgrade-pxs8j-fkmcj to cleanup","name":"upgrade-repo-hvst-upgrade-pxs8j","namespace":"harvester-system","pos":"server.go:151","timestamp":"2022-03-17T16:44:16.969113Z","uid":"5be1bfa7-df6a-4ef2-be38-c6b8e7d91857"}
[...]

PR #2064, opened 2022-03-23
[Upgrade] Various fixes

bk201:
This PR improves the upgrade node script by:

- [Upgrade: shutdown VMs if live-migration takes too long](https://github.com/harvester/harvester/commit/c26f323205b12b0363706663654a9af7bc284984)
  - Could be a workaround to https://github.com/harvester/harvester/issues/2029.
- [Upgrade: image enhancement](https://github.com/harvester/harvester/commit/db7a56406056b71a42cbc7ab19b74c53cbe480de)
  - Clean RKE2 image tarball when upgrade is done.
  - Preload RKE2 images. RKE2 images might be needed during nodes upgrade. A waiting-for-upgrade node doesn't load RKE2 images yet because its RKE2 server/agent doesn't restart. We need to preload RKE2 images beforehand.

**Related Issue:**
https://github.com/harvester/harvester/issues/2029
https://github.com/harvester/harvester/issues/1861

**Test plan:**
- Perform an upgrade
- There should be no RKE2 tarballs under the images directory.
  ```
  node1:~ # ls /var/lib/rancher/rke2/agent/images/rke2*
  ls: cannot access '/var/lib/rancher/rke2/agent/images/rke2*': No such file or directory
  ```

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -1,7 +1,6 @@
 #!/bin/bash -ex
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-HOST_DIR="${HOST_DIR:-/host}"
 UPGRADE_TMP_DIR=$HOST_DIR/usr/local/upgrade_tmp
 
 source $SCRIPT_DIR/lib.sh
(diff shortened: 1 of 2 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -116,6 +117,30 @@ wait_vms_out()
   done
 }
 
+wait_vms_out_or_shutdown()
+{
+  local vm_count
+  local max_retries=240
```
weihanglo, line 100: Just wonder why 240 times of retry?

weihanglo (approved):
LGTM

package/upgrade/upgrade_node.sh
bk201: it's around 20 minutes :) (240 *5)

FrankYang0529 (approved):
LGTM.

weihanglo (approved):
LGTM. Those images were deleted successfully.

PR #2089, opened 2022-03-28
Upgrade: Preload image again with new harvester-upgrade image

bk201:
**Problem:**

**Note**: this is a special fix to v1.0.0.

RKE2 upgrade is done node by node, after upgrading RKE2 runtime on a node, there might be some RKE2 pods schedules on the non-upgraded nodes. In air-gapped environments, pods might error because RKE2 images are not loaded yet (they are loaded only when RKE2 agent/server restarts).
Screen Shot 2022-03-18 at 11 13 22 AM

**Solution:**

The PR create a plan to preload image with the new `harvester-ugprade` image again. We can guarantee [RKE2 images preload](https://github.com/harvester/harvester/blob/3c3004ed372cc6f0ac1e0e0d8f41ad79f7237689/package/upgrade/upgrade_node.sh#L71).

In the next release, we'll always preload RKE2 images first.

**Related Issue:**
https://github.com/harvester/harvester/issues/1861
**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -134,7 +134,7 @@ wait_kubevirt()
       fi
     fi
 
-    echo "Libvirt current version: $current_target_version, target version: $version"
+    echo "KubeVirt current version: $current_target_version, target version: $version"
     sleep 5
   done
 }
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -ex
+#!/bin/bash -e
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Tested and the upgrade succeeded. LGTM

package/upgrade/upgrade_manifests.sh
```diff
@@ -124,6 +134,7 @@ wait_kubevirt()
       fi
     fi
 
+    echo "Libvirt current version: $current_target_version, target version: $version"
```
johnliu55tw: I think it's `KubeVirt`?
```suggestion
    echo "KubeVirt current version: $current_target_version, target version: $version"
```

johnliu55tw (approved):
Verified RKE2 images are preloaded during the manifest upgrade step.

package/upgrade/upgrade_manifests.sh
bk201: That's embarrassing, thanks for spotting this.

PR #2097, opened 2022-03-30
Upgrade: workaround for rancher `v2.6.4-harvester1`

bk201:
The specified minimum rancher-webhook chart (1.0.4+up0.2.5) is not in
the rancher image, we can work around by lowering the required version.

**Related issue**
https://github.com/rancher/rancher/issues/37113

PR #2100, opened 2022-03-31
Revert "Upgrade: workaround for rancher `v2.6.4-harvester1`"

bk201:
Reverts harvester/harvester#2097

This is no longer needed because Rancher resolves the known issues.

NOTE: Sorry I was not aware using Github's revert feature creates a branch in the harvester repo, will avoid this in the future.


guangbochen (approved):
LGTM

Issue #2105, opened 2022-04-01
[BUG] typo in error message


dexwood:
**Describe the bug**
I found a typo in an error message dealing in persistentvolumeclaim/validator.go

admission webhook "validator.harvesterhci.io" denied the request: resizing is only supported for detached volumes. The volueme is being used by VM default/testubuntu. Please stop the VM first.

**To Reproduce**
1. Try to edit a persistent volume claim while it is attached.

**Expected behavior**
The following error message is expected:

admission webhook "validator.harvesterhci.io" denied the request: resizing is only supported for detached volumes. The volume is being used by VM default/testubuntu. Please stop the VM first.

**Environment:**
 - Harvester ISO version: 1.0.0

slash2314:
See pull request #2106

guangbochen:
close as the PR https://github.com/harvester/harvester/pull/2106 is merged, thanks for ur contribution!

PR #2106, opened 2022-04-01
Correct typo in error message.

dexwood:
**Problem:**
See Issue typo in error message #2105
There is a typo in the follow:
admission webhook "validator.harvesterhci.io" denied the request: resizing is only supported for detached volumes. The volueme is being used by VM default/testubuntu. Please stop the VM first.

It should be 
admission webhook "validator.harvesterhci.io" denied the request: resizing is only supported for detached volumes. The volume is being used by VM default/testubuntu. Please stop the VM first.

**Solution:**
Correct the error string

**Test plan:**
Try to edit attached volume to see the correct error message.


guangbochen (approved):
LGTM and thanks!

Issue #2060, opened 2022-03-22
[FEATURE] Add VM run strategy support (RerunOnFailure by default)


tjjh89017:
**Is your feature request related to a problem? Please describe.**

Harvester will set up VM with `running: true` for now. VM will act as container's policy but not in the common hypervisor.
When shutdown VM from inside, kubevirt will start VM back to running state to keep the policy.

switch to `runStrategy: Manual` will be closer to the behavior of the common hypervisor.
related to https://github.com/harvester/harvester/issues/1882

**Describe the solution you'd like**

Switch creation in UI and template to manual
and use runStrategy to determine the status of VM itself.

**Describe alternatives you've considered**

**Additional context**

depends: (need to rebase before merge it.)
- [x] https://github.com/harvester/harvester/pull/1963
- [x] https://github.com/harvester/harvester/pull/2050
- [x] https://github.com/harvester/harvester/pull/2061
- [x] https://github.com/harvester/dashboard/pull/271
- [x] https://github.com/harvester/harvester/pull/2118
- [x] https://github.com/harvester/dashboard/pull/278
    - [x] still need dashboard template PR for "acpi enable by default"
    - [x] still need dashboard PR for "default runStrategy: Manual in UI"
    - [x] still need dashboard PR for "Start virtual machine on creation"

optional:
- [ ] update terraform driver and VMBuilder https://github.com/harvester/terraform-provider-harvester/blob/759882b0b04297d57dc4867dd7a29fe8b6cc428e/internal/provider/virtualmachine/resource_virtualmachine_constructor.go

**Test Plan**

### Test Case 1
1. create VM from UI, it should use `runStrategy: Manual`
2. VM should have `ACPI enabled true`
3. with `Start virtual machine on creation`, VM should start automatically.
    - without `Start virtual machine on creation`, VM should be off state
4. Pressing Stop button to stop VM shouldn't have any error

### Test Case 2
1. create VM from UI, it should use `runStrategy: Manual`
2. Execute `sudo poweroff` command in VM, VM should keep off state when VM is off.
    - VM should not be ON state without manually start.

### Test Case 3
1. create VM from UI, it should use `runStrategy: Manual`
2. backup it
3. restore to existing VM.
4. VM should still use `runStrategy: Manual`

### Test Case 4
1. create VM from UI, it should use `runStrategy: Manual`
2. backup it
3. restore to new VM
4. VM should still use `runStrategy: Manual`

### Test Case 5
[...]

PR #2118, opened 2022-04-06
enable ACPI by default

tjjh89017:

**Problem:**

https://github.com/harvester/harvester/issues/2060

**Solution:**

enable ACPI for VM by default

**Related Issue:**
https://github.com/harvester/harvester/issues/2060

**Test plan:**

please refer test plan in https://github.com/harvester/harvester/issues/2060

### DIFF
--- a/pkg/data/template.go
+++ b/pkg/data/template.go
@@ -156,6 +156,9 @@ spec:
         spec:
           evictionStrategy: LiveMigrate
           domain:
+            features:
+              acpi:
+                enabled: true
             cpu:
               cores: 1
             devices:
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Approved. This seems no harm, though I found that it should be enabled by default according to the doc https://kubevirt.io/api-reference/v0.49.0/definitions.html#_v1_features

bk201 (approved):
`virtualmachinetemplateversions` CRs are created once, we might also need to add this feature to the upgrade path.

guangbochen (approved):

Issue #2139, opened 2022-04-11
[BUG] Harvester "make package" fails


w13915984028:
**Describe the bug**

```
/go/src/github.com/harvester/harvester$ make package
...
IMAGE_PUSH_TAG: debug-2013-head
./scripts/package: line 11: cd: ./scripts/package: Not a directory
FATA[0011] exit status 1                                
make: *** [Makefile:11: package] Error 1
...
```

**To Reproduce**
1. Harvester source code master-head
2. make build
3. make package, the failure occurs

**Expected behavior**

`make package` should be successful

**Support bundle**

**Environment:**
 - Harvester ISO version: V1.0.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Linux, X86

**Additional context**

w13915984028:
unfortunately, directly "make package-webhook" is also failed,  only "make package-upgrade " is successful

```
/go/src/github.com/harvester/harvester$ make package-webhook 

ARCH: amd64
VERSION: cf837a6f-dirty
CHART_VERSION: 0.0.0-fix-2139-cf837a6f-dirty
APP_VERSION: fix-2139-cf837a6f-dirty
SUFFIX: -amd64
REPO: rancher
TAG: dev
IMAGE_PUSH_TAG: fix-2139-head
./scripts/package-webhook: line 11: cd: ./scripts/package: Not a directory
FATA[0010] exit status 1                                
make: *** [Makefile:11: package-webhook] Error 1


```

w13915984028:
There is conflict between `"make ci" `and `"make package/package-webhook"`

A straightforward fix helps` "make package" `but breaks` "make ci"`
![image](https://user-images.githubusercontent.com/31133476/162944590-9a69074d-b5ca-4060-8d49-56926951f797.png)


`make ci `fails
![image](https://user-images.githubusercontent.com/31133476/162945119-fef90f2f-7655-4365-afbf-1321fa6c7b79.png)

noahgildersleeve:
Tested on harvester master `05b5da2dff3a9188a9d1defca30d905e326e7083` with included repro steps. Fixed.



> **Describe the bug**
> 
> ```
> /go/src/github.com/harvester/harvester$ make package
> ...
> IMAGE_PUSH_TAG: debug-2013-head
> ./scripts/package: line 11: cd: ./scripts/package: Not a directory
> FATA[0011] exit status 1                                
> make: *** [Makefile:11: package] Error 1
> ...
[...]

PR #2141, opened 2022-04-12
fix failure of 'make package'

w13915984028:

**Problem:**

Harvester 'make package' fails due to incorrect dir

**Solution:**

Fix with correct dir

**Related Issue:**
#2139 

**Test plan:**

```
1. git clone harvester master-head source code
2. make build
3. make package
```
step 3 should be successful

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -1,14 +1,18 @@
 #!/bin/bash
 set -e
 
-source $(dirname $0)/version
+TOP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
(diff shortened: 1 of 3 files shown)


tlehman (approved):

bk201:
Nice spot. I guess `make package-upgrade` and `make package-webhook` have the same issue. Should we fix together?

Another approach is not to rely on related path: https://github.com/bk201/harvester/commit/9e4ec032be510d7cd753749871758aa33c8277f1

bk201 (approved):
LGTM!

guangbochen (approved):

Issue #1390, opened 2021-10-08
[FEATURE] Support configuring a VLAN at the management interface in installer config


bashofmann:
**Is your feature request related to a problem? Please describe.**
Currently you can't configure a VLAN directly at the harvester management interface, see https://docs.harvesterhci.io/v0.3/install/harvester-configuration/#installnetworks.

**Describe the solution you'd like**
Add more network configuration options to the installation options, so that you can add an additional tagged vlan to a bonded management interface.

A resulting network config could be

![network-interfaces](https://user-images.githubusercontent.com/243056/136570782-05686026-5ae1-420b-9000-09e1bdf3369b.png)

With such a routing config

![routes](https://user-images.githubusercontent.com/243056/136570903-0ea88895-6ec3-48a4-8f3e-0667204a0093.png)

The harvester vip should be attachable to the vlan

![vip_interface](https://user-images.githubusercontent.com/243056/136571163-cfb7d26c-5cf2-445a-8c8d-025877fee86d.png)

FrankYang0529:
## Test plan

### Case 1 - No VLAN

* Set `vlan_id` as `0` or don't set it.
```
install:
# ...
  networks:
    harvester-mgmt:
      interfaces:
      - name: ens5
      method: static
      ip: 192.168.0.30
      subnet_mask: 255.255.255.0
      gateway: 192.168.0.1
      vlan_id: 0
```
```
install:
# ...
  networks:
    harvester-mgmt:
      interfaces:
      - name: ens5
      method: static
      ip: 192.168.0.30
      subnet_mask: 255.255.255.0
      gateway: 192.168.0.1
```
* Install Harvester.
* Check `harvester-mgmt` has IP `192.168.0.30` and there is no `vlan0@harvester-mgmt`.

### Case 2 - VLAN

* Set `vlan_id` as `1`.
```
install:
# ...
  networks:
    harvester-mgmt:
      interfaces:
      - name: ens5
      method: static
      ip: 192.168.0.30
      subnet_mask: 255.255.255.0
      gateway: 192.168.0.1
      vlan_id: 1
```
* Install Harvester.
* Check `vlan1@harvester-mgmt` is existent and `vlan1@harvester-mgmt` has IP `192.168.0.30`.

FrankYang0529:
We need further PR to update network-controller-harvester. If users set VLAN on harvester-mgmt and enable vlan feature in harvester, `harvester-br0` should inherit the VLAN tag of `harvester-mgmt`.

[...]

PR #2160, opened 2022-04-18
HEP: vlan enhance

tjjh89017:

**Problem:**

IT usually configures the switch with VLAN trunk mode rather than VLAN access mode or hybrid mode in a production environment. Therefore, we should add VLAN support for harvester-mgmt and allow multiple Linux bridges in Harvester node configuration.

preview page: https://github.com/tjjh89017/harvester/blob/vlan_enhance/enhancements/20220418-vlan-enhancement.md

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/1390

**Test plan:**


tjjh89017:
create a new PR with [CI SKIP]
in https://github.com/harvester/harvester/pull/2161

PR #2161, opened 2022-04-18
HEP: vlan enhance

tjjh89017:

**Problem:**

IT usually configures the switch with VLAN trunk mode rather than VLAN access mode or hybrid mode in a production environment. Therefore, we should add VLAN support for harvester-mgmt and allow multiple Linux bridges in Harvester node configuration.

preview page: https://github.com/tjjh89017/harvester/blob/vlan_enhance/enhancements/20220418-vlan-enhancement.md

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/1390

**Test plan:**


guangbochen (commented):
Thanks for the HEP documentation, here's some personal feedback:
1. Can u please add the relevant modifications to the Network CRD YAML file?
1. Can you describe the changes involved in the upgrade first (maybe a solution will be added later)?
1. For network diagrams, prefer to change the `Not Possible` to `Not Support`

tjjh89017:
> Thanks for the HEP documentation, here's some personal feedback:
> 
> 1. Can u please add the relevant modifications to the Network CRD YAML file?

I listed `ClusterNetwork` and `NodeNetwork` that we need to modify or we need to create some new CRDs to fit the requirement.
I'm going to create some CRD examples for this enhancement.

> 2. Can you describe the changes involved in the upgrade first (maybe a solution will be added later)?

It will depend on implementation.
I will add it later if we have more discussion about details.

> 3. For network diagrams, prefer to change the `Not Possible` to `Not Support`

Updated.

Thanks a lot

tjjh89017:
update example CRD design.

tjjh89017:
close this
and track in https://github.com/harvester/harvester/pull/3066

Issue #2114, opened 2022-04-05
[BUG] Failed to upgrade harvester when node name too long cause error "Invalid value: ... must be no more than 63 characters"


TachunLin:
**Describe the bug**

When upgrade harvester from v1.0.0 to v1.0.1-rc2, if the node name too long like `harvester-101upgrade-node1`
The upgrade will fail with error message
  ```
  Job.batch "hvst-upgrade-grwbw-single-node-upgrade-harvester-101upgrade-node1" is invalid: spec.template.labels: 
  Invalid value: "hvst-upgrade-grwbw-single-node-upgrade-harvester-101upgrade-node1": must be no more than 63 characters
  ```
  ![image](https://user-images.githubusercontent.com/29251855/161706107-a54d33a0-9198-4d41-87db-8af48bfe6dee.png)

**To Reproduce**
1. Prepare v1.0.0 one node harvester
2. Set a longer node name `harvester-101upgrade-node1`
3. Create a ubuntu image
4. Create a virtual machine
5. Shutdown VM
6. Follow the steps to upgrade to v1.0.1-rc2
https://github.com/harvester/docs/blob/42598035cfe79856f7e159d4c47c4d7d29178900/docs/upgrade/automatic.md

**Expected behavior**

Should allow longer node name to prevent the no longer than 63 characters error during the upgrade to v1.0.1

**Support bundle**

[supportbundle_37fc8ce4-a657-4380-b86d-b517ee995c0e_2022-04-05T07-52-07Z.zip](https://github.com/harvester/harvester/files/8419019/supportbundle_37fc8ce4-a657-4380-b86d-b517ee995c0e_2022-04-05T07-52-07Z.zip)

**Environment:**
 - Harvester ISO version: v1.0.0 GA to v1.0.1-rc2
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 1 node harvester

**Additional context**

bk201:
https://github.com/harvester/harvester/blob/03ee62133765a03394e1a35ccd4859ad1afd1302/pkg/controller/master/upgrade/common.go#L200
Here the job name is automatically set to label `job-name` by Kubernetes. We need a way to handle the name length > 63 characters. 

A possible way to fix is follow how SUC define a plan's job names: https://github.com/rancher/system-upgrade-controller/blob/79da9f058209b9906bfd6163d58672e9ae67ea5d/pkg/upgrade/job/job.go#L103

TachunLin:
[...]

PR #2165, opened 2022-04-18
Upgrade: concatenate node jobs names safely

bk201:
**Deps**
- I'd suggest testing with https://github.com/harvester/harvester/pull/2166 to make an upgrade complete.

**Problem:**

Fail to create node job if node name is too long.
**Solution:**

Safely concatenate node job names. The idea is to add some hash in the name.

**Related Issue:**
https://github.com/harvester/harvester/issues/2114

**Test plan:**

- Build Harvester ISO, the rancher/harvester image should contain the change
- Create a Harvester v1.0.1 cluster. At least one node should contain a long enough hostname. An example is `harvester-101upgrade-node1`.
- Upgrade to the build ISO.

### DIFF
--- a/pkg/controller/master/upgrade/common.go
+++ b/pkg/controller/master/upgrade/common.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	upgradev1 "github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io/v1"
+	"github.com/rancher/wrangler/pkg/name"
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
LGTM

To other reviewers, here is [the source code of `SafeConcatName`](https://github.com/rancher/wrangler/blob/6970ad98ba7bd2755312ccfc6540a92bc9a9e316/pkg/name/name.go#L52-L66)

FrankYang0529 (approved):
LGTM. Validate it can work with https://github.com/harvester/harvester/pull/2166. Thanks.

Issue #2152, opened 2022-04-14
[BUG] bundledeployment rancher-monitoring always has status "modified"


w13915984028:
**Describe the bug**

For a successfully installed Harvester cluster, it runs well, but the bundledeployment` rancher-monitoring` always has status `"modified"`  with `"mutatingwebhookconfiguration.admissionregistration.k8s.io rancher-monitoring-admission modified"`.

```
$ kubectl get bundledeployment -A 

NAMESPACE                                NAME                                          DEPLOYED   MONITORED   STATUS
..
cluster-fleet-local-local-1a3d67d0a899   mcc-rancher-monitoring                        True       True        mutatingwebhookconfiguration.admissionregistration.k8s.io rancher-monitoring-admission modified {"webhooks":[{"admissionReviewVersions":["v1","v1beta1"],"clientConfig":{"caBundle":"LS..=","service":{"name":"rancher-monitoring-operator","namespace":"cattle-monitoring-system","path":"/admission-prometheusrules/mutate","port":443}},"failurePolicy":"Ignore","matchPolicy":"Equivalent","name":"prometheusrulemutate.monitoring.coreos.com","namespaceSelector":{},"objectSelector":{},"reinvocationPolicy":"Never","rules":[{"apiGroups":["monitoring.coreos.com"],"apiVersions":["*"],"operations":["CREATE","UPDATE"],"resources":["prometheusrules"],"scope":"*"}],"sideEffects":"None","timeoutSeconds":10}]}
cluster-fleet-local-local-1a3d67d0a899   mcc-rancher-monitoring-crd                    True       True  
```

**To Reproduce**
```
1. Install Harvester cluster (one node or multi nodes)
2. When the system UI is ready, execute "kubectl get bundledeployment -A ", get the related status.
3. Meanwhile, continuous retrying in the log of fleet-agent.
```

**Expected behavior**

System should be steady, no such warnings.

**Support bundle**

**Environment:**
 - Harvester ISO version:  v1.0.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM based.

**Additional context**
Other related issues:
#1531 
#2013

w13915984028:
Analysis:

The rancher-monitoring chart, which is used by Harvester, is at  https://github.com/rancher/charts/tree/dev-v2.6/charts/rancher-monitoring/100.1.0%2Bup19.0.3

1. Values in values.yaml
[...]

PR #2182, opened 2022-04-25
Addition of script for v1.0.1 ugprade which introduces monitoring fixes

ibrokethecloud:

**Problem:**

This PR introduces the upgrade script for harvester upgrades from v1.0.1 to handle the patching of managed monitoring charts to include the diff for managed charts to ignore the modified Validating and Mutating webhook configurations.

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/2152

This is related to harvester-installer PR: https://github.com/harvester/harvester-installer/pull/274

**Test plan:**

### DIFF
--- a/package/upgrade/migrations/managed_charts/v1.0.1.sh
+++ b/package/upgrade/migrations/managed_charts/v1.0.1.sh
@@ -1,0 +1,43 @@
+#!/bin/bash -ex
+
+CHART_NAME=$1
+CHART_MANIFEST=$2
+
+remove_vip_config()
+{
+  # VIP interface isn't needed after kube-vip v0.4.1 (https://github.com/harvester/harvester-installer/pull/216)
+  yq e 'del(.spec.values.kube-vip.config)' $CHART_MANIFEST -i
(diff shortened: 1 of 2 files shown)


bk201:
Thanks for the PR.
Because the upgrade support path is from 1.0.1 to 1.0.2, so we can assume some values are already there.
I'd say `patch_prometheus_resources` is the only migration we need for now.

johnliu55tw:
With this fix, we could change the target upgraded state of `rancher-monitoring` from `modified` to `ready`:
https://github.com/harvester/harvester/blob/05b5da2dff3a9188a9d1defca30d905e326e7083/package/upgrade/upgrade_manifests.sh#L337-L338

@ibrokethecloud could you help replace `modified` with `ready`? Thanks!

ibrokethecloud:
@johnliu55tw this is updated now.

johnliu55tw (approved):
LGTM

bk201 (approved):
LGTM.

PR #2192, opened 2022-04-25
Upgrade: fix upgrade temporary directory path

bk201:
The upgrade temporary directory is supposed to be on a node's`/usr/local/upgrade_tmp`, not in container's file system.
Although it makes no difference at this moment, let's fix the bad refactor.

The original intention was to use a directory in node's `/usr/local`, because:
- It's easier to debug. Files are not gone with the container's termination.
- It will reside in the persistent storage from our design.


weihanglo (approved):
Looks reasonable. 👍🏾

johnliu55tw (approved):
LGTM

Issue #2172, opened 2022-04-21
[BUG] VM created from windows-iso-image template can't start automatically, need the extra manual start effort  


TachunLin:
**Describe the bug**

When create a windows virtual machine from windows-iso-image template.
The virtual machine can't start automatically and stay in the `off` 

```
start virtual machine default/windows-template failed, persistentvolumeclaims "windows-template-cdrom-disk-rvk33" not found
```
![image](https://user-images.githubusercontent.com/29251855/164446279-949ffce1-c7f6-46c4-9b90-4c04a4b63d80.png)

User need to manual start the windows VM to make it running. 

**To Reproduce**
1. Create a windows iso image from file or upload via URL
  https://storage.googleapis.com/harvester/windows_server_2019_essentials.iso
2. Create a VM , select `windows-iso-image-base-version` template
3. Edit the appropriate CPU and memory
  ![image](https://user-images.githubusercontent.com/29251855/164449378-22ee09a9-6487-4893-95fe-304399035319.png)
4. Select windows iso image with type `cd-rom`
![image](https://user-images.githubusercontent.com/29251855/164449439-7d1a1895-2c7f-4af2-bc04-f4fc5feb7817.png)
5. Click the create button

**Expected behavior**

VM created from windows-iso-image template can start automatically without error 

**Support bundle**

[supportbundle_28479d0f-c787-4ec6-b759-673c63ce6c76_2022-04-21T11-38-33Z.zip](https://github.com/harvester/harvester/files/8530610/supportbundle_28479d0f-c787-4ec6-b759-673c63ce6c76_2022-04-21T11-38-33Z.zip)

**Environment:**
 - Harvester ISO version: `master-03ee6213-head` (04/18) 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 1 node harvester on local kvm machine

**Additional context**
  ![image.png](https://images.zenhubusercontent.com/61519853321ea20d65443929/bcc7793e-25bc-41f5-9966-c511a27c3084)
  ![image.png](https://images.zenhubusercontent.com/61519853321ea20d65443929/6ee9cf32-c4f9-4d62-9a57-74fde8844e9d)

WuJun2016:
It's not a problem that happens 100% of the time.  It is not necessarily related to whether or not the window template is selected to create the vm.     Need  @tjjh89017  to help troubleshoot this issue.

tjjh89017:
[...]

PR #2208, opened 2022-04-29
Add vm run strategy enhancement

guangbochen:

**Problem:**
- Add workaround of the upstream kubevirt issue https://github.com/kubevirt/kubevirt/issues/7295 (after stopping and starting the VM the original run strategy will be changed to `Always`)
- After restoring a VM the default run strategy is always default to the manual, it should use the run strategy from the backup.

**Solution:**
- store the last run strategy to the annotation and patch the run strategy from the annotation after a VM is restarted from the halted status.
- use the saved run strategy from the backup status and set the default value to `RerunOnFailure` if it is not defined.

**Related Issue:**
https://github.com/harvester/harvester/issues/2172

## **Test plan:**
For local testing, u can patch the harvester and harvester-webhook image with `guangbo/harvester:fix-run2` and `guangbo/harvester-webhook:fix-run2` respectively, thanks. (use the upstream `master-head` image if this PR is merged)

### Test Case 1 - RerunOnFailure
1. create VM and set the run strategy to `RerunOnFailure` via the `Advanced Option` tab.
2. the VM will start by default.
3. stop and restart the VM the run strategy will remain the same.
4. stop the VM inside the guestOS with `sudo shutdown` and it will NOT restart automatically. (need to wait 20-30 seconds for the status change)

### Test Case 2 - Always
1. create VM and set the run strategy to `Always` via the `Advanced Option` tab.
2. the VM will start by default.
3. stop and restart the VM the run strategy will remain the same.
4. stop the VM inside the guestOS with `sudo shutdown` and it will restart automatically.

### Test Case 3 - Manual
1. create VM and set the run strategy to `Manual` via the `Advanced Option` tab.
2. the VM will NOT start by default.
3. stop and restart the VM the run strategy will remain the same.
4. stop the VM inside the guestOS with `sudo shutdown` and it will NOT restart automatically. (need to wait 20-30 seconds for the status change)

### Test Case 4 - Backup
1.  Config the backup target. (e.g. using the [LH minio](https://longhorn.io/docs/1.2.4/snapshots-and-backups/backup-and-restore/set-backup-target/#set-up-a-local-testing-backupstore))
2. Create a VM and take a backup
3. Restore a new VM from the backup, and make sure its run strategy is reminded as same as the backup one.

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -178,8 +178,7 @@ func (vf *vmformatter) canStop(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv1.Vi
 		case kubevirtv1.RunStrategyHalted:
 			return false
 		case kubevirtv1.RunStrategyManual, kubevirtv1.RunStrategyRerunOnFailure:
-			// need to check VMI status
-			if vmi == nil || vmi.Status.Phase != kubevirtv1.Running {
+			if vmi == nil {
 				return false
 			}
 			return true
@@ -223,9 +222,14 @@ func (vf *vmformatter) canDoBackup(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv
 		return false
 	}
 
-	if vmi != nil && vmi.Status.Phase != kubevirtv1.Running {
+	if vm.DeletionTimestamp != nil || (vmi != nil && vmi.DeletionTimestamp != nil) {
 		return false
 	}
+
+	if vmi != nil && vmi.Status.Phase != kubevirtv1.Running && vmi.Status.Phase != kubevirtv1.Succeeded {
+		return false
+	}
+
 	return true
 }
 
@@ -246,9 +250,14 @@ func (vf *vmformatter) isVMStarting(vm *kubevirtv1.VirtualMachine) bool {
 }
 
 func (vf *vmformatter) canCreateTemplate(vmi *kubevirtv1.VirtualMachineInstance) bool {
-	if vmi != nil && vmi.Status.Phase != kubevirtv1.Running {
+	if vmi != nil && vmi.DeletionTimestamp != nil {
 		return false
 	}
+
+	if vmi != nil && vmi.Status.Phase != kubevirtv1.Running && vmi.Status.Phase != kubevirtv1.Succeeded {
+		return false
+	}
+
 	return true
 }
 
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -509,7 +509,10 @@ func (h *RestoreHandler) createNewVM(restore *harvesterv1.VirtualMachineRestore,
 		return nil, err
 	}
 
-	defaultRunStrategy := kubevirtv1.RunStrategyManual
+	defaultRunStrategy := kubevirtv1.RunStrategyRerunOnFailure
+	if backup.Status.SourceSpec.Spec.RunStrategy != nil {
+		defaultRunStrategy = *backup.Status.SourceSpec.Spec.RunStrategy
+	}
 
 	vm := &kubevirtv1.VirtualMachine{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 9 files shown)


guangbochen:
Also, this has been discussed with the UI team and they found it a bit tricky to add a manual start callback on the UI (because they had a hard time determining if the PVC was created or not, so in some cases, users still ran into https://github. com/harvester/harvester/issues/2172), so we decided to use `RerunOnFailure` as the default strategy, thanks.

pkg/api/vm/formatter.go
```diff
@@ -179,7 +179,7 @@ func (vf *vmformatter) canStop(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv1.Vi
 			return false
 		case kubevirtv1.RunStrategyManual, kubevirtv1.RunStrategyRerunOnFailure:
 			// need to check VMI status
-			if vmi == nil || vmi.Status.Phase != kubevirtv1.Running {
+			if vmi == nil {
```
tjjh89017, line 181: So if VMI is existing, harvester will allow to force stop VM to remove VMI?
Is it the purpose?

pkg/controller/master/backup/restore.go
```diff
@@ -755,13 +758,16 @@ func (h *RestoreHandler) deleteOldPVC(vmRestore *harvesterv1.VirtualMachineResto
...
+		return err
+	}
 
+	logrus.Infof("starting the vm %s, current state running:%v", vm.Name, runStrategy)
 	switch runStrategy {
-	case kubevirtv1.RunStrategyAlways:
+	case kubevirtv1.RunStrategyAlways, kubevirtv1.RunStrategyRerunOnFailure:
 		return nil
-	case kubevirtv1.RunStrategyManual, kubevirtv1.RunStrategyRerunOnFailure:
+	case kubevirtv1.RunStrategyManual:
```
tjjh89017, line 770: This will cause,
If we execute poweroff command inside VM, it will never have "Start button" but only have "restart button".
We could only "restart" it.
Do we expect this behavior?

If yes, this patch is good to me, just ignore this comment.

tjjh89017 (approved):
If those behaviors in comments are expected.
LGTM.

pkg/api/vm/formatter.go
guangbochen: Yes, users should be able to retry the `stop` action as long as the VM is not stopped.

pkg/controller/master/backup/restore.go
guangbochen: Actually I found users is still be able to have the `Start` button even if we shutdown the VM inside the guestOS, check the [canStart action formatter](https://github.com/harvester/harvester/blob/master/pkg/api/vm/formatter.go#L147)
image

tjjh89017: But if VM is RerunOnFailure, it will never have start button.
and RerunOnFailure VM is able to shutdown from inside.

pkg/api/vm/formatter.go
tjjh89017: Ok LGTM.

pkg/controller/master/backup/restore.go
guangbochen: hmm, wired, we can check offline if u have env that can reproduce this behavior, at least it works in my env, thanks.
image

tjjh89017: That is weird.
I will test it next Wed.
And we could merge first.
if i found this issue, i will create a new ticket for this.

guangbochen: thx, btw I also found a small issue, when the VMI is in a `Succeeded` status(shutdown inside the guestOS), we should allow users to create a backup, thanks for bringing it up.

tjjh89017: yep, if we have VMI, we should let users backup it.

FrankYang0529 (approved):
LGTM. All test cases can pass. I also test change RunStrategy by editing config. It also works. Thanks.

guangbochen:
Fixed some VM action formatter checks, can u guys please take a re-review, thanks.

tjjh89017 (approved):
LGTM

pkg/api/vm/formatter.go
```diff
@@ -179,7 +179,7 @@ func (vf *vmformatter) canStop(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv1.Vi
 			return false
 		case kubevirtv1.RunStrategyManual, kubevirtv1.RunStrategyRerunOnFailure:
 			// need to check VMI status
```
FrankYang0529: We can remove this comment if we don't check `vmi.Status` in the following logic.

FrankYang0529 (approved):
LGTM. Thanks.

pkg/api/vm/formatter.go
guangbochen: resolved, thanks.

Issue #2221, opened 2022-05-05
[BUG] VM has temporary `ErrorUnschedulable` status


FrankYang0529:
**Describe the bug**
When we create a new VM, it shows a temporary status `ErrorUnschedulable`. It's caused that we create VM before we have a PVC in the cluster.

![image](https://user-images.githubusercontent.com/4927361/166856783-4da62770-1b0a-4296-9807-2abde466c7e0.png)

**To Reproduce**
1. Create a VM.

**Expected behavior**
VM doesn't have this temporary status.

**Environment:**
 - Harvester ISO version: master-e2985453-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

FrankYang0529:
KubeVirt marks VM as `ErrorUnschedulable` when `PodScheduled` condition is `False` and its reason is `Unschedulable`.

https://github.com/kubevirt/kubevirt/blob/f8160a0b5eac2acafab74fd7916bfcaf0c39a6a3/pkg/virt-controller/watch/vm.go#L1774-L1780

FrankYang0529:
Close the issue since we can let users create storageclass with `WaitForFirstConsumer`. If they use that kind of storageclass, they don't encounter this issue anymore.

![Screenshot 2023-02-22 at 1 51 53 PM](https://user-images.githubusercontent.com/4927361/220534893-c54b3647-2832-4ce7-a576-ec41b6aa5b4b.png)

Issue #2088, opened 2022-03-28
[BUG] post-upgrade hook of longhorn-storageclass-override.yaml has hidden issue


w13915984028:
**Describe the bug**

Fleet-agent/Helm is always complaing "configmap.v1 longhorn-system/longhorn-storageclass missing", and fleet-agent will try again every 5 minutes.

`time="2022-03-27T12:02:02Z" level=warning msg="DEV bd: name:mcc-harvester MonitorBundle: has modified status, EnqueueAfter 5 minutes, status[0]:configmap.v1 longhorn-system/longhorn-storageclass missing"`

fleet-agent has lots of log like:
`time="2022-03-27T09:14:13Z" level=info msg="Helm: Upgrading mcc-harvester"`

The possible root cause:
The pose-upgrade hook modified resources, which caused Helm assumes failures when checking consistency of the whole release. (Need to be confirmed)

https://helm.sh/docs/topics/charts_hooks/  does not explains clearly if a  post-install/post-upgrade hook could modify/update a resource

![image](https://user-images.githubusercontent.com/31133476/160362110-8c2180eb-5712-452a-a82d-991960318521.png)

![image](https://user-images.githubusercontent.com/31133476/160362139-b28d539c-c267-4706-90ea-fcacd10d4120.png)

The 5 minutes loop behaviour may also influence the real system upgrade.

**To Reproduce**
1. Looks in any running Harvester, this message is there.

**Expected behavior**

When the cluster is stable, there should not be so many backgroud upgades.

**Support bundle**

**Environment:**
 - Harvester ISO version:  V1.0.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):  KVM based Harvester.

**Additional context**

https://github.com/harvester/harvester/issues/857

https://github.com/harvester/harvester/issues/1893  could also be related

https://github.com/harvester/harvester/issues/2013

w13915984028:
change log of  longhorn-storageclass

![image](https://user-images.githubusercontent.com/31133476/160389144-079acb3d-37ca-4461-90ae-606b0b7b8ecf.png)

![image](https://user-images.githubusercontent.com/31133476/160389035-205b95f2-9a73-41e4-a971-ccfe9fcad667.png)

```
$ kubectl logs -n harvester-system -l app.kubernetes.io/component=apiserver 

[...]

PR #2235, opened 2022-05-09
change volumeBindingMode to WaitForFirstConsumer

FrankYang0529:
**Problem:**
When we create a new VM, it shows a temporary status `ErrorUnschedulable` and error message `unbound immediate PersistentVolumeClaims`.

**Solution:**
Change `volumeBindingMode` in `StorageClass` to `WaitForFirstConsumer`, so we don't get this unscheduled error from pod. Also, VM's first status will change to `WaitingForVolumeBinding`.

**Related Issue:**
https://github.com/harvester/harvester/issues/2221

**Test plan:**
2. Import Image and check its `volumeBindingMode` in `StorageClass` is `WaitingForVolumeBinding`.
3. Run `kubectl get vm -w` on the terminal.
4. Create a VM with our imported image.
5. Check there is no `ErrorUnschedulable` shown on the terminal.
```
> kubectl get vm -w
NAME   AGE   STATUS   READY
vm     0s
vm     0s    WaitingForVolumeBinding   False
vm     5s    Starting                  False
vm     23s   Running                   False
vm     23s   Running                   True
```

### DIFF
--- a/deploy/charts/harvester/templates/longhorn-storageclass-override.yaml
+++ b/deploy/charts/harvester/templates/longhorn-storageclass-override.yaml
@@ -19,7 +19,7 @@ data:
     provisioner: driver.longhorn.io
     allowVolumeExpansion: true
     reclaimPolicy: "{{ .Values.longhorn.persistence.reclaimPolicy }}"
-    volumeBindingMode: Immediate
+    volumeBindingMode: WaitForFirstConsumer
     parameters:
       numberOfReplicas: "{{ .Values.longhorn.persistence.defaultClassReplicaCount }}"
       staleReplicaTimeout: "30"
--- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -148,7 +148,7 @@ func (h *vmImageHandler) createBackingImage(image *harvesterv1.VirtualMachineIma
 
 func (h *vmImageHandler) createStorageClass(image *harvesterv1.VirtualMachineImage) error {
 	recliamPolicy := corev1.PersistentVolumeReclaimDelete
-	volumeBindingMode := storagev1.VolumeBindingImmediate
+	volumeBindingMode := storagev1.VolumeBindingWaitForFirstConsumer
 	sc := &storagev1.StorageClass{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: getImageStorageClassName(image.Name),


weihanglo (approved):
Looks lovely!
Verified that it behaves as expected.

FrankYang0529:
Add `volumeBindingMode: WaitForFirstConsumer` to both `longhorn` and `longhorn-image-xxx` StorageClass:
* Create Volume without VM. Volume will be stuck in `NotReady`. It will become `Ready` after there is a VM using it.
* Create VM with both StorageClass Volumes. VM will not show any `ErrorScheduling` message.

Only add `volumeBindingMode: WaitForFirstConsumer` to `longhorn-image-xxx` StorageClass:
* Create `longhorn` Volume without VM. Volume can go into `Ready`.
* Create VM with both StorageClass Volumes. VM may show `ErrorScheduling` message.
```
kubectl get vm -w
NAME   AGE   STATUS   READY
vm     0s
vm     0s    WaitingForVolumeBinding   False
vm     0s    ErrorUnschedulable        False
vm     14s   Starting                  False
vm     33s   Running                   True
```

cc @guangbochen. `volumeBindingMode` is an immutable field, users cannot change it after the harvester is deployed. Do you think we still need to add this setting? Or we just ignore this temporary error and fix it with UI PR https://github.com/harvester/dashboard/pull/307?

weihanglo:
> ...`volumeBindingMode` is an immutable field, users cannot change it after the harvester is deployed.

Can the current upgrade path update that immutable field?

BTW I tested the feature by modifying that immutable field via `kubectl edit -n longhorn-system configmap longhorn-storageclass`

FrankYang0529:
> > ...`volumeBindingMode` is an immutable field, users cannot change it after the harvester is deployed.
> 
> Can the current upgrade path update that immutable field?
> 
> BTW I tested the feature by modifying that immutable field via `kubectl edit -n longhorn-system configmap longhorn-storageclass`

If it can be modified through `configmap`, I think it can be update during the upgrade. Thanks for your information! Probably, we can have a setting for users. If they want to create volume immediately, they can change the `lognhorn-storageclass configmap`.

FrankYang0529:
After testing, I also can update `longhorn-storageclass configmap`. After updating it to `Immediate`, the `Pending` Volume will become `Ready`.

w13915984028:
we are considering to remove deploy/charts/harvester/templates/longhorn-storageclass-override.yaml, due to https://github.com/harvester/harvester/issues/2088

w13915984028:
Unfortunately, per https://github.com/longhorn/longhorn/pull/3881/files

LH is not exposing `volumeBindingMode` to chart values, if we use HOOK to force changing it, the fleet-agent may still always complain.

@FrankYang0529 @guangbochen @ibrokethecloud

guangbochen:
We are not going to remove the default `longhorn` storage class but considering migrating to use a custom `harvester-specific` LH storage class instead.

And like @w13915984028 has mentioned, keep editing the LH's default SC probably is not recommended due to both system upgrade and LH's best practice concerns.

FrankYang0529:
@guangbochen, for #2221, do you think it's okay to delay PVC binding if there is no consumer? If not, we may still get `ErrorUnschedulable` due to `unbound immediate PersistenVolumeCliam`.

w13915984028:
@FrankYang0529 change policy may cause other problmes.
e.g.  user creates a bunch of "volume" in advance, and then create VM(s) to use those existing volume,  delay  binding  may cause burst, and capacity congestion is another concern.
I am not sure we can cover all cases quickly.

back to the issue itself, it it caused by user selecting: create VM and add (implicitly create)  volumes  together?

FrankYang0529:
> back to the issue itself, it it caused by user selecting: create VM and add (implicitly create) volumes together?

Yeah, in the harvester VM controller, we check the `harvesterhci.io/volumeClaimTemplates` annotation. If a PVC is existent, we skip it. If not, we create it.

Also, if we set `volumeBindingModel: Immediate` in StorageClass, there will be a short time that PVC is just created and it's not bound to a PV. So pod shows the error `unbound Immediate PersistentVolumeClaim` in `PodScheduled` condition.

https://github.com/harvester/harvester/blob/fdb64376010ad8993c5488b4779ced0edf88b7b2/pkg/controller/master/virtualmachine/vm_controller.go#L43-L48

w13915984028:
Seems another way is: the vm controller checks the PVCs, until PVCs meet requirements in a defined time, otherweise, report creating VM failure.

FrankYang0529:
> Seems another way is: the vm controller checks the PVCs, until PVCs meet requirements in a defined time, otherweise, report creating VM failure.

Actually, the pod is created from the KubeVirt controller. If there is no PVC, they will show `ErrorPvcNotFound` and will not create the pod.

https://github.com/kubevirt/kubevirt/issues/7697#issuecomment-1119572637

They don't check whether PVC is bound, because there is a status for it `WaitingForVolumeBinding`. However, if there is an error from the pod, VM may show the error from the pod. You can check the second case in this comment. VM shows `WaitingForVolumeBinding` first, and then `ErrorUnschedulable`.

https://github.com/harvester/harvester/pull/2235#issuecomment-1122137543

w13915984028:
We have another issue: https://github.com/harvester/harvester/issues/1695, and an requirement to LH: https://github.com/longhorn/longhorn/issues/3734

LH is in async mode, PVC created != PVC is really usable.

Now we leave those burdens to KubeVirt,  let it to bind VM with PVC with more.

On the other hand, the vm controller may try to do a lit more, check those conditions, then have a user friendly feedback. Sure, it is not a tiny task, more a HEP.

At the moment, I would rather keep this issue unchanged and write it into document.

guangbochen:
@FrankYang0529 this can be closed in favor of https://github.com/harvester/harvester/pull/2768, and the `volumeBindingMode` will be configurable via the UI in v1.1.0.

Issue #2238, opened 2022-05-09
[BUG] Some settings don't have `configured` condition


FrankYang0529:
**Describe the bug**
Even if we configure some settings successfully like `backup-target` to NFS or `release-download-url`, they don't have condition `configured` in `status`. This may make some functions broken:

* https://github.com/harvester/harvester/blob/7fcf04f8df9797020d51d72e0d2ff9aa8c84b904/pkg/api/vm/handler.go#L414-L427

![Screen Shot 2022-05-09 at 4 28 00 PM](https://user-images.githubusercontent.com/4927361/167370695-5c7af656-db2c-4de9-a8f6-769642c01e64.png)

**To Reproduce**
1. Setup a harvester clusetr.
2. Configure `backup-target` to NFS. (LH NFS ref: https://longhorn.io/docs/1.2.4/snapshots-and-backups/backup-and-restore/set-backup-target/#nfs-backupstore)
3. Check status in `backup-target`. It's empty.
4. Configure `release-download-url` to a random URL.
5. Check status in `release-download-url`. It's empty.

**Expected behavior**
They should have condition `configured` as `true` in the `status`.

**Environment:**
 - Harvester ISO version: master-head
 - Underlying Infrastructure: KVM

lanfon72:
is this a case of #2209 ?

FrankYang0529:
@lanfon72, if kinchlo's backup-target doesn't have `configured` condition, I think these two cases are the same. Let me check in issue #2209.

lanfon72:
Verified this bug has been fixed.

![image](https://user-images.githubusercontent.com/5169694/167804005-c685f4fa-1476-4d80-a5da-7a215eb2c287.png)
![image](https://user-images.githubusercontent.com/5169694/167803727-f02f0bc2-f6cd-4030-9baa-94b1c3875cd1.png)


Test Information
----
* Environment: **qemu/KVM single node**
* Harvester Version: **v1.0.2-rc2**
* **ui-source** Option: **Auto**

### Verify Steps:
- Follow steps in https://github.com/harvester/tests/issues/335#issue-1230650793

janeczku:
https://github.com/harvester/harvester/issues/2209#issuecomment-1124049483

dseralathan:
kubectl create -f backup-target.yaml

[...]

PR #2240, opened 2022-05-09
Set configured condition for all settings

FrankYang0529:
**Problem:**
In before, we only set `configured` condition for some settings which have `syncer` functions. This may break some function like backup which check whether `configured` condition in `backup-target` setting is `True`.

* https://github.com/harvester/harvester/blob/24240d896c1add07de72bfb4fb3d7bd446a6dba0/pkg/controller/master/setting/register.go#L60-L70
* https://github.com/harvester/harvester/blob/24240d896c1add07de72bfb4fb3d7bd446a6dba0/pkg/controller/master/setting/handler.go#L81-L101
* https://github.com/harvester/harvester/blob/7fcf04f8df9797020d51d72e0d2ff9aa8c84b904/pkg/api/vm/handler.go#L414-L427

**Solution:**
Set `configured` condition for all settings.

**Related Issue:**
https://github.com/harvester/harvester/issues/2238

**Test plan:**
1. Setup a harvester cluster.
2. Set `backup-target` setting as NFS. (LH NFS ref: https://longhorn.io/docs/1.2.4/snapshots-and-backups/backup-and-restore/set-backup-target/#nfs-backupstore)
3. Check `configured` condition in `backup-target` setting is `True`.
4. Set `release-download-url` setting as a random URL.
5. Check `configured` condition in `release-download-url` setting is `True`.

### DIFF
--- a/pkg/controller/master/setting/handler.go
+++ b/pkg/controller/master/setting/handler.go
@@ -78,27 +78,21 @@ func (h *Handler) settingOnChanged(_ string, setting *harvesterv1.Setting) (*har
 		return nil, nil
 	}
 
-	for key, syncer := range syncers {
-		if setting.Name != key {
-			continue
-		}
-
+	if syncer, ok := syncers[setting.Name]; ok {
 		if err := syncer(setting); err != nil {
 			if updateErr := h.setConfiguredCondition(setting.DeepCopy(), err); updateErr != nil {
 				return setting, updateErr
 			}
 			return setting, err
 		}
-
-		toUpdate := setting.DeepCopy()
-		if toUpdate.Annotations == nil {
-			toUpdate.Annotations = make(map[string]string)
-		}
-		toUpdate.Annotations[util.AnnotationHash] = currentHash
-		return setting, h.setConfiguredCondition(toUpdate, nil)
 	}
 
-	return nil, nil
+	toUpdate := setting.DeepCopy()
+	if toUpdate.Annotations == nil {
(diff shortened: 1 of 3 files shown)


weihanglo (approved):
It works. LGTM.

pkg/controller/master/setting/handler.go
```diff
@@ -89,16 +89,15 @@ func (h *Handler) settingOnChanged(_ string, setting *harvesterv1.Setting) (*har
...
 			return setting, err
 		}
-
-		toUpdate := setting.DeepCopy()
-		if toUpdate.Annotations == nil {
-			toUpdate.Annotations = make(map[string]string)
-		}
-		toUpdate.Annotations[util.AnnotationHash] = currentHash
-		return setting, h.setConfiguredCondition(toUpdate, nil)
+		break
```
guangbochen: using `break` to reconcile the setting doesn't look correct to me, prefer to add the resync method to those required settings accordingly.

FrankYang0529: I think what we do here is not reconcile the setting. One setting can only be mapped to one `syncerFunc`. If we don't have error from `syncer`, we can just leave the for loop and update the setting's annotation and condition. If there is an error, we write it in condition and return.

https://github.com/harvester/harvester/blob/24240d896c1add07de72bfb4fb3d7bd446a6dba0/pkg/controller/master/setting/handler.go#L81-L91

bk201 (approved):
LGTM

guangbochen (approved):

Issue #2252, opened 2022-05-10
[BUG] Can't provision RKE clusters in Rancher Harvester node driver


noahgildersleeve:
**Describe the bug**

You can't rpovision RKE1 or RKE2 clusters in Harvester from Rancher node driver
**To Reproduce**
1. Import Harvester into Rancher
2. Add cloud credentials
3. Create RKE cluster from node driver on Harvester
**Expected behavior**

The clusters hsould provision successfully
**Support bundle**

[supportbundle_804d6ba1-ba58-4bd5-bee0-71f85cc60b18_2022-05-10T22-11-00Z.zip](https://github.com/harvester/harvester/files/8665301/supportbundle_804d6ba1-ba58-4bd5-bee0-71f85cc60b18_2022-05-10T22-11-00Z.zip)

**Environment:**
 - Harvester ISO version: v1.0.2-rc2
 - Rancher version: v2.6.5-rc9
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 node Harvester libvirt vagrant setup and Rancher was a single node docker instance

**Additional context**
This was working on RC1 when I tested it last week. There was an Error creating machine: Error in driver during machine creation: Internal error occurred: admission webhook "virtualmachine-validator.kubevirt.io" denied the request: Running and RunStrategy are mutually exclusive` This seems like it might be related to https://github.com/harvester/harvester/pull/2208 based on the timing.
![Screenshot from 2022-05-10 15-14-48](https://user-images.githubusercontent.com/83787952/167730535-754e59cf-0dc0-4ab6-91ab-7939a288d2c9.png)
![Screenshot from 2022-05-10 15-10-28](https://user-images.githubusercontent.com/83787952/167730536-ebec664d-7174-48ad-8483-bb965d55ba1e.png)

TachunLin:
Also checked on Rancher `v2.6.5-rc10` with Harvester `v1.0.2-rc2`.
Currently we can't provision `RKE2`, `K3s` and `RKE1` guest clusters in the following kubernetes versions:

1. RKE2: `v1.23.6+rke2r2` and `v1.22.9+rke2r2`
2. K3s: `v1.23.6+k3s1` and `v1.22.9+k3s1`
3. RKE1: `v1.23.6-rancher1-1` and `v1.22.9-rancher1-1`

The RKE2 provisioning process keep looping in the flow of `Creationerror` -> `Deleting`  

* RKE2 cluster VM creation Error
[...]

PR #2253, opened 2022-05-11
Fix run strategy mutual-exclusive error on webhook

guangbochen:

**Problem:**
Running and RunStrategy are mutually exclusive of the VM spec.

**Solution:**
Do not patch the run strategy when the `spec.running` filed is specified by the users.

**Related Issue:**
https://github.com/harvester/harvester/issues/2252

**Test plan:**
- patch the webhook image via `$kubectl set image deployment/harvester-webhook harvester-webhook=guangbo/harvester-webhook:fix-wh`
- create an RKE2 cluster with the Harvester node driver, and wait for the cluster to be up and running. (previously it will encounter the `mutual-exclusive` error)
- check the VM YAML and its `spec.running` value is kept.

image


bk201 (approved):
LGTM.
I can provision an RKE2 cluster.

FrankYang0529 (approved):
LGTM

Issue #2263, opened 2022-05-11
[BUG] Unable to stop VM which in starting state


lanfon72:
### Describe the bug
**Stop** button been disabled when the VM in a long starting state.

![image](https://user-images.githubusercontent.com/5169694/167813844-014051c6-01f2-4c8a-9e59-a3d5fd517a5a.png)
![image](https://user-images.githubusercontent.com/5169694/167813871-2db1d406-f427-4e8b-a4be-adbdd230c211.png)

### To Reproduce
1. Install Harvester with any nodes
2. Create an Windows iso image for VM creation
3. Create the Windows VM by using the iso image
4. When the VM in **Starting** state, **Stop** button should able to click and work as expected

### Expected behavior
VM should be able to stop in **Starting** state

## Environment:
 - Harvester ISO version: **v1.0.2-rc1**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 3nodes **

lanfon72:
Verified **Stop** button available and work as expected.


Test Information
----
* Environment: **qemu/KVM 3 nodes**
* Harvester Version: **master-a242948b-head**
* **ui-source** Option: **Auto**

### Verify Steps:
- Follow **Steps to reproduce** in https://github.com/harvester/harvester/issues/2263#issue-1232274843

PR #2268, opened 2022-05-12
A starting VM can be stopped

FrankYang0529:
**Problem:**
A starting VM cannot be stopped. However, sometime there may be some settings error, so the VM cannot be started. Users should be able to stop a starting VM.

**Solution:**
Update `vmformatter.canStop` function.

**Related Issue:**
https://github.com/harvester/harvester/issues/2263

**Test plan:**

Case 1: RunStartegy Always / RerunOnFailure
1. Create a VM with RunStrategy `Always` or `RerunOnFailure`.
2. VM will get into starting status.
3. Users can click the stop button to stop a starting VM.
4. After VM is off, users can click the start button to start the VM again. Also, the VM should be able to get into running status.

Case 2: RunStartegy Manual / Halted
1. Create a VM with RunStrategy `Manual` or `Halted`.
2. VM will not start. Also, there is no `stop` button on UI.
3. Click the start button and the `stop` button shows on UI.
4. Users can click the stop button to stop a starting VM.
5. After VM is off, users can click the start button to start the VM again. Also, the VM should be able to get into running status.

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -168,10 +168,6 @@ func (vf *vmformatter) canRestart(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv1
 }
 
 func (vf *vmformatter) canStop(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv1.VirtualMachineInstance) bool {
-	if vm.Spec.Running != nil && !*vm.Spec.Running {
-		return false
-	}
-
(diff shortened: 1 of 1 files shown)


guangbochen (approved):
LGTM, ~~@bk201 should we considering to include this PR in v1.0.2 if the QA can help to validate this one?~~

weihanglo (approved):
Thanks for the quick fix!

Issue #2265, opened 2022-05-11
[BUG] VM's RunStrategy will be updated to `Halted` when stop the VM


lanfon72:
### Describe the bug
There have two different behaviors:
- Stop the VM by clicking **Stop** Button, RunStrategy will be updated `RerunOnFailure -> Halted`
- Stop the VM by executing `poweroff` insidie the VM, RunStrategy _**will not be update**_, but user can still click **Stop** Button to stop the VM again. (even if the VM is in `Off` state)

### To Reproduce
1. Install Harvester with any nodes
2. Create an image for VM creation
3. Create the VM `vm1`
4. Login to `vm1`, execute `sudo poweroff`
5. The VM's state should be `Off`
6. Navigate to VM's Config, **RunStrategy** in _Advanced Options_ should be `RerunOnFailure`
7. Select the VM, **Stop** Button is not disabled
8. Click **Stop** button, then navigate to VM's Config
9. **RunStrategy** in _Advanced Options_ will be updated to `Halted`

### Expected behavior
No idea, at least they should be consistent?

## Environment:
 - Harvester ISO version: **v1.0.2-rc2**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **qemu/KVM single node**

### Additional context
**Generate Template** function will be affected, if user do **Generate Template** on the VM which is `Off`, the default `RunStrategy` will be **Halted**.

related issues: #2172, #1953, #574, #1882, #2154

FrankYang0529:
VMI has many phases. If we run `sudo poweroff` in the VM, the pod of the VM will become `complete` and VMI will become `Succeed`. We should also determine the phase of VMI in `canStop`.

https://github.com/kubevirt/kubevirt/blob/a3aaf1ace40c44c9dafaf2c091851d5cdb829fd7/staging/src/kubevirt.io/api/core/v1/types.go#L648-L668

lanfon72:
@guangbochen please help to update the milestone and priority accordingly.

FrankYang0529:
@lanfon72, sending a stop request through UI to a VM with `RerunOnFailure` or `Always` RunStrategy and it becomes `Halted` is a known issue. Please check https://github.com/kubevirt/kubevirt/issues/7295.

I think it's better to change the issue title to `UI should not show the stop button when users already run "poweroff" in the VM`.

lanfon72:
[...]

PR #2276, opened 2022-05-13
UI should not show the stop button when users already run `poweroff` in the VM

FrankYang0529:
**Problem:**
Users can stop a VM through UI when we already run `poweroff` in the VM.

**Solution:**
Check `vmi.status.phase`. Only if it is `Pending`, `Scheduling`, `Scheduled`, or `Running`, we can send a stop request.

**Related Issue:**
https://github.com/harvester/harvester/issues/2265

**Test plan:**
2. Create a Linux VM.
3. Run `sudo poweroff` in it.
4. UI should not show `stop` button.


tjjh89017 (approved):
LGTM

weihanglo (approved):
Make sense. Thanks!

guangbochen (approved):

PR #2281, opened 2022-05-17
Use management node for clarity

simonflood:
**Problem:**
Server node is ambiguous

**Solution:**
Management node is used in docs


guangbochen (approved):
Thanks for the update.

Issue #2175, opened 2022-04-22
[BUG] Harvester fails to pull Rancher agent image from private registry


janeczku:
**Describe the bug**
When registering Harvester with Rancher in an air-gapped infrastructure,  Harvester fails to pull the `cattle-cluster-agent` image from the private registry due to certificate errors.

**To Reproduce**

1. Create a private PKI consisting of root CA and server certificate
2. Setup a private container registry, e.g. Harbor, using the server certificate
3. Mirror the `rancher/rancher-agent` image to the private registry
4. Configure the private registry as "system-default-registry" in the Rancher settings
5. Configure the private CA root certificate in Harvester settings -> additional-ca
6. Register Harvester cluster in Rancher

**Expected behaviour**

Additional CA certificate is added to Harvester OS trust store. Harvester successfully pulls the image.

**Actual behaviour**

Registration fails. Harvester fails to pull the `cattle-cluster-agent` image  from the private registry because the certificate CA is not trusted.

FrankYang0529:
Currently, the workaround for this issue is adding the crt file to `/etc/pki/trust/anchors` and running `update-ca-certificates` on each host.

If we want to fix this issue properly, we need to introduce `harvester-system-agent` for handling system stuff. I tried to use following YAML to mount host folder and `harvester-additional-ca` secret to a container. In this container, I try to move `additional-ca.pem` to `/host/etc/pki/truct/anchors` and run `update-ca-certificates`. Although the file appears on the host, running `update-ca-certificates` in the container cannot influence the host.

cc @guangbochen @bk201 

```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: sync-additional-ca
  namespace: harvester-system
spec:
  selector:
    matchLabels:
      name: sync-additional-ca
  template:
    metadata:
      labels:
        name: sync-additional-ca
    spec:
      tolerations:
      - key: node-role.kubernetes.io/control-plane
        operator: Exists
        effect: NoSchedule
      - key: node-role.kubernetes.io/master
        operator: Exists
        effect: NoSchedule
      containers:
      - name: sync-additional-ca
        image: alpine
        command: [ "/bin/sh", "-c", "--" ]
        args: [ "while true; do sleep 30; done;" ]
        securityContext:
          privileged: true
          capabilities:
            add: ["SYS_ADMIN"]
        resources:
          limits:
            memory: 200Mi
          requests:
            cpu: 100m
            memory: 200Mi
        volumeMounts:
[...]

PR #2285, opened 2022-05-19
add addition-ca secret and plan

FrankYang0529:
**Problem:**
When users add an additional-ca, they also expect this can be applied to the host. For example, pulling images can also use this ca.

**Solution:**
Using SUC plan to change ca on the host.

**Related Issue:**
https://github.com/harvester/harvester/issues/2175

**Test plan:**

Setup:
* Create a harvester cluster and a ubuntu server. Make sure they can reach each other.
* On the ubuntu server, install docker and run the following commands.
```
mkdir -p certs
openssl req \
  -newkey rsa:4096 -nodes -sha256 -keyout certs/domain.key \
  -addext "subjectAltName = DNS:myregistry.local" \
  -x509 -days 365 -out certs/domain.crt
sudo mkdir -p /etc/docker/certs.d/myregistry.local:5000
sudo cp certs/domain.crt /etc/docker/certs.d/myregistry.local:5000/domain.crt
docker run -d \
  -p 5000:5000 \
  --restart=always \
  --name registry \
  -v "$(pwd)"/certs:/certs \
  -v "$(pwd)"/registry:/var/lib/registry \
  -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
  -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
  registry:2
docker pull rancher/rancher-agent:v2.6.5
docker tag rancher/rancher-agent:v2.6.5 myregistry.local:5000/rancher/rancher-agent:v2.6.5
docker push myregistry.local:5000/rancher/rancher-agent:v2.6.5
```
* On each harvester node, add ubuntu IP to `/etc/hosts`.
```
# vim /etc/hosts
192.168.0.50 myregistry.local
```
* Create a rancher v2.6.5 (it's can be a docker container or VM). After it starts, update `system-default-registry` setting to `myregistry.local:5000`

Case 1: Import Harvester to Rancher
* Import harvester cluster to the rancher and doesn't have any error.
* Check image in `cattle-system/cattle-cluster-agent` deployment is `myregistry.test:5000/rancher/rancher-agent:v2.6.5`.

Case 2: Add another node to Harvester cluster
* Add another node to harvester cluster and add `192.168.0.50 myregistry.local` to `/etc/hosts`.
* Login to the new node and run the following command. It should not have error.
```
curl https://myregistry.local:5000
```

Case 3: Remove additional-ca
* Set `additional-ca` setting as default.
* Wait for new `cattle-system/apply-sync-additional-ca-on-xxx` jobs finish.
* Login to any harvester node. It should have `SSL certificate problem` with the following command.
```
curl https://myregistry.local:5000
```

### DIFF
--- a/deploy/charts/harvester/templates/cattle-system.yaml
+++ b/deploy/charts/harvester/templates/cattle-system.yaml
@@ -1,0 +1,24 @@
+# system-upgrade-controller(https://github.com/rancher/system-upgrade-controller) watches secret and node change.
+# If there is a new node, system-upgrade-controller creates a job on the node to change CA certificates.
+# If additional-ca is changed, system-upgrade-controller creates a job on each node to change CA certificates.
+apiVersion: upgrade.cattle.io/v1
+kind: Plan
+metadata:
+  name: sync-additional-ca
+  namespace: cattle-system
+spec:
+  concurrency: 1
+  nodeSelector:
+    matchLabels:
+      harvesterhci.io/managed: "true"
+  serviceAccountName: system-upgrade-controller
+  secrets:
+    - name: harvester-additional-ca
+      path: /ca
+  version: v1.0.2
+  upgrade:
+    image: {{ .Values.generalJob.image.repository }}:{{ .Values.generalJob.image.tag }}
+    command: ["/bin/sh", "-c"]
+    args:
+      - cp /ca/additional-ca.pem /host/etc/pki/trust/anchors;
+      - chroot /host update-ca-certificates
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -445,8 +445,8 @@ support-bundle-kit:
     repository: rancher/support-bundle-kit
     tag: master-head
 
-loop-device-cleaner:
+generalJob:
   image:
     imagePullPolicy: IfNotPresent
-    repository: alpine
-    tag: 3.12
+    repository: registry.suse.com/bci/bci-base
+    tag: 15.3
--- a/pkg/data/secret.go
+++ b/pkg/data/secret.go
@@ -1,0 +1,32 @@
+package data
+
+import (
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/harvester/harvester/pkg/config"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+func createSecrets(mgmt *config.Management) error {
+	secrets := mgmt.CoreFactory.Core().V1().Secret()
+
+	// Initializing the secret for Plan cattle-system/sync-additional-ca,
+	// so sync-additional-ca doesn't fail to mount harvester-additional-ca secret to jobs.
+	additionalCA := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: util.CattleSystemNamespaceName,
+			Name:      util.AdditionalCASecretName,
+		},
+		Data: map[string][]byte{
+			util.AdditionalCAFileName: []byte(""),
+		},
+	}
+	if _, err := secrets.Create(&additionalCA); err != nil && !apierrors.IsAlreadyExists(err) {
+		return errors.Wrapf(err, "Failed to create secret %s/%s", additionalCA.Namespace, additionalCA.Name)
+	}
+
+	return nil
+}
--- a/tests/integration/runtime/construct.go
+++ b/tests/integration/runtime/construct.go
@@ -18,15 +18,14 @@ func Construct(ctx context.Context, kubeConfig *restclient.Config) error {
 		return nil
 	}
 
-	// create namespace
-	err := client.CreateNamespace(kubeConfig, testHarvesterNamespace)
-	if err != nil {
-		return fmt.Errorf("failed to create target namespace, %v", err)
-	}
-
-	err = client.CreateNamespace(kubeConfig, testLonghornNamespace)
-	if err != nil {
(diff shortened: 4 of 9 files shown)


deploy/charts/harvester/templates/cattle-system.yaml
```diff
@@ -0,0 +1,24 @@
...
+  nodeSelector:
+    matchLabels:
+      harvesterhci.io/managed: "true"
+  serviceAccountName: system-upgrade-controller
+  secrets:
+    - name: harvester-additional-ca
+      path: /ca
+  version: v1.0.2
+  upgrade:
+    image: alpine:3.12
```
weihanglo: Should we continue using apline? I thought we are in the transition to BCI images or else

pkg/data/secret.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func createSecrets(mgmt *config.Management) error {
+	secrets := mgmt.CoreFactory.Core().V1().Secret()
+
+	// Initializing the secret for Plan cattle-system/sync-additional-ca,
+	// so sync-additional-ca doesn't fail to mount harvester-additional-ca secret to jobs.
+	additionalCA := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: cattleNamespace,
```
weihanglo: Could we reuse this?

```suggestion
			Namespace: util.CattleSystemNamespaceName,,
```

FrankYang0529: Updated it. Thank you.

deploy/charts/harvester/templates/cattle-system.yaml
FrankYang0529: @johnliu55tw @bk201 do you have any recommended image to run the script here? Thank you.

bk201: I'd recommend BCI image and better template it from values.yaml file.

weihanglo (commented):
`additional-ca.pem` is synced successfully for those test cases. Great job.

deploy/charts/harvester/templates/cattle-system.yaml
FrankYang0529: Updated it. Thank you.

tests/integration/runtime/construct.go
```diff
@@ -18,15 +18,14 @@ func Construct(ctx context.Context, kubeConfig *restclient.Config) error {
...
-	err = client.CreateNamespace(kubeConfig, testLonghornNamespace)
-	if err != nil {
-		return fmt.Errorf("failed to create target namespace, %v", err)
+	// create namespaces
+	var err error
+	namespaces := []string{testHarvesterNamespace, testLonghornNamespace, testCattleNamespace}
+	for _, namespace := range namespaces {
+		err = client.CreateNamespace(kubeConfig, namespace)
+		if err != nil {
+			return fmt.Errorf("failed to create target namespace, %v", err)
```
w13915984028: Add namespace's value to error meesage to identify which one is failed.

deploy/charts/harvester/templates/cattle-system.yaml
```diff
@@ -0,0 +1,24 @@
...
+  secrets:
+    - name: harvester-additional-ca
+      path: /ca
+  version: v1.0.2
+  upgrade:
+    image: {{ .Values.syncAdditionalCA.image.repository }}:{{ .Values.syncAdditionalCA.image.tag }}
+    command: ["/bin/sh", "-c"]
+    args:
+      - cp /ca/additional-ca.pem /host/etc/pki/trust/anchors;
+      - chroot /host update-ca-certificates
```
w13915984028, line 24: The initial additional-ca secrete is empty, will it also trigger here, should it be excluded ?

meanwhile, is it possible to delete additional-ca ?

deploy/charts/harvester/values.yaml
```diff
@@ -450,3 +450,8 @@ loop-device-cleaner:
     imagePullPolicy: IfNotPresent
     repository: alpine
     tag: 3.12
+
+syncAdditionalCA:
```
bk201: Thanks for the update. Do you think there is a need for a more general field name that can be re-used by jobs. For example, there is a workaround daemonset to remove stale longhorn loop devices. Just feel this kind of job can share the same image.

tests/integration/runtime/construct.go
FrankYang0529: Updated it. Thank you.

deploy/charts/harvester/templates/cattle-system.yaml
FrankYang0529: Yes, it also triggers, but it's okay. An empty additional-ca will remove the last added additional-ca, so we can pass the test case 3 `Remove additional-ca`.

deploy/charts/harvester/values.yaml
FrankYang0529: Yes, I update it. Thank you.

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):
LGTM!

weihanglo (approved):
Looks nice!

deploy/charts/harvester/templates/cattle-system.yaml
```diff
@@ -0,0 +1,24 @@
+# system-upgrade-controller watches secret and node change.
```
guangbochen:
```suggestion
# system-upgrade-controller(https://github.com/rancher/system-upgrade-controller) watches secret and node change.
```

guangbochen (commented):
LGTM, @FrankYang0529 can u please help to backport the PR to the `v1.0` branch and also need to create a backport issue for v1.1.0, thanks.

deploy/charts/harvester/templates/cattle-system.yaml
FrankYang0529: Added it. Thank you.

PR #2313, opened 2022-05-24
Use Embedded Rancher to avoid confusion

simonflood:
Use Embedded Rancher to avoid confusion with Rancher that should be installed separately to integrate with Harvester

**Problem:**

**Solution:**

**Related Issue:**

**Test plan:**


bk201 (approved):
Much clear by using the embedded terms. Thanks!

guangbochen (approved):

Issue #1919, opened 2022-02-15
[BUG] Non-exhaustive type checks causing failed schema inspection


joshmeranda:
When running harvester with binary built from compiling source at f303b27e it fails with an error:

`FATA[0000] failed to create harvester server: failed inspecting type v1beta1.VirtualMachineTemplateVersion, field spec: failed inspecting type v1beta1.VirtualMachineTemplateVersionSpec, field vm: failed inspecting type v1beta1.VirtualMachineSourceSpec, field spec: failed inspecting type v1.VirtualMachineSpec, field template: failed inspecting type v1.VirtualMachineInstanceTemplateSpec, field spec: failed inspecting type v1.VirtualMachineInstanceSpec, field domain: failed inspecting type v1.DomainSpec, field devices: failed inspecting type v1.Devices, field disks: failed inspecting type v1.Disk, field bootOrder: unknown type kind uint`

#### Expected
 - Harvester should startup normally

#### Observed
 - Harvester fails with the error above

#### To reproduce
```
>>> export KUBECONFIG="$HOME/.kube/config" NAMESPACE="harvester-system" RANCHER_EMBEDDED="true" HCI_MODE="true"
>>> go build -ldflags "-X google.golang.org/protobuf/reflect/protoregistry.conflictPolicy=warn -X github.com/harvester/harvester/vendor/google.golang.org/protobuf/reflect/protoregistry.conflictPolicy=warn"
>>> ./harvester
```

After digging into the source it looks like this is caused by type checks in [Wrangler](https://github.com/rancher/wrangler)'s reflection type checks being non-exhaustive and omitting the `Uint` go type at [reflection.go#271](https://github.com/rancher/wrangler/blob/master/pkg/schemas/reflection.go#L271) and [reflection.go#467](https://github.com/rancher/wrangler/blob/master/pkg/schemas/reflection.go#L467) which cause the `unknown type kind uint` error thrown as [reflection.go#524](https://github.com/rancher/wrangler/blob/master/pkg/schemas/reflection.go#L524).

joshmeranda:
Here is a code sample  with an example of the issue:

```go 
import (
"fmt"
"github.com/rancher/wrangler/pkg/schemas/openapi"
)

type sample struct {
	BootOrder uint
}

func main() {
	json, err := openapi.ToOpenAPIFromStruct(sample{})
	if err != nil {
		fmt.Printf("Error: %s\n", err)
	}

	_ = json

}
```

This code will output:

```
[...]

PR #2314, opened 2022-05-24
update wrangler vendor dependency

joshmeranda:

**Problem:**

Fix for related issues was merged, but the harvester dependency was never updated.

**Solution:**

Update the harvester wrangler vendor dependency

**Related Issue:**
https://github.com/harvester/harvester/issues/1919#issuecomment-1135242066


weihanglo:
Sorry I am a bit confused. Which command you ran to generate vendor deps? I don't see any change in go.mod and not sure how you got those vendor sources.

Issue #2147, opened 2022-04-13
[FEATURE] Storage Tiering


janeczku:
**Is your feature request related to a problem? Please describe.**

- Requirement to use mixed storage hardware of different performance levels to optimize cost
- Using heterogeneous storage devices on Harvester nodes leads to unpredictable performance today because there is no control over what device tier will back a specific volume/replica.

**Describe the solution you'd like**
Harvester should allow admins to define storage tiers based on at least:
- Node label selectors
- Disk label selectors

janeczku:
Requirement: It should be possible to define the number of replicas per storage tiers.

https://github.com/harvester/harvester/issues/2511

abonillabeeche:
Can we have an update on this feature request? It's in Roadmap but don't yet see PRs or a discussion around it.

futuretea:
@abonillabeeche 
design PR: https://github.com/harvester/harvester/pull/2335
backend PR: https://github.com/harvester/harvester/pull/2358
UI PR: https://github.com/harvester/dashboard/pull/415

irishgordo:
@futuretea 
I've been working to try to validate this.

Following some of the [testing steps](https://github.com/harvester/harvester/pull/2358#issue-1263010285) that have been highlighted, I've encountered this error when trying to upgrade from v1.0.3 - v1.1.0-rc1.  Following [upgrading harvester via more of an air-gapped method](https://docs.harvesterhci.io/v1.0/upgrade/automatic/#prepare-an-air-gapped-upgrade)

```
suse-workstation-team-harvester➜  tests : vm-import-controller ✘ :✭ ᐅ  kubectl --kubeconfig ~/.kube/vm_harvester_v103_on_supermicro.yaml get bundle -A
NAMESPACE     NAME                                          BUNDLEDEPLOYMENTS-READY   STATUS
fleet-local   fleet-agent-local                             1/1                       
fleet-local   local-managed-system-agent                    1/1                       
fleet-local   mcc-harvester                                 0/1                       Modified(1) [Cluster fleet-local/local]; configmap.v1 longhorn-system/longhorn-storageclass missing
fleet-local   mcc-harvester-crd                             1/1                       
fleet-local   mcc-local-managed-system-upgrade-controller   0/1                       ErrApplied(1) [Cluster fleet-local/local: another operation (install/upgrade/rollback) is in progress]
fleet-local   mcc-rancher-monitoring                        1/1                       
[...]

Issue #2027, opened 2022-03-17
[BUG] Longhorn Replica (if there is only one) should be on the same node as the VM


meysholdt:
**Describe the bug**
We've got fast disks (NVME with rw > 2GB/sec) but a slow network (1GB/sec) and we run ephemeral VMs (it's ok if we lose data).

On the initial Harvester setup, disk read/write throughput was clearly bound by the network. For three replicas, I got 300 MB/s read and 100MB/sec write. 

I assumed I can eliminate the network as the bottleneck by reducing the replicas to one. 
Unfortunately, my assumption turned out to be false because the replica and VM did not get scheduled to the same node. Thus, all disk read/write went over the network again and was therefore limited to 100MB/sec read and 100MB/sec write. 

**To Reproduce**
1. create a Harvester cluster with e.g. four nodes
2. upload a VM image
3. copy the images storage class and edit the number of replicas to be 1. E.g. via `kubectl get storageclass > foo.yaml`, edit foo.yaml, `kubectl apply -f foo.yaml`)
4. Start a VM using the new storage class. 
5. Open the Longhorn UI and the Harvester UI and see that volume-replica and VM are not on the same node.

**Expected behavior**
Volume replica and VM should be on the same node.

**Support bundle**
N/A.

**Environment:**
 - Harvester ISO version: 1.0.0
 - Underlying Infrastructure: Harvester runs inside a VM in CentOS on bare-metal Hetzner servers.

**Additional context**

meysholdt:
In case there is a workaround for this behaiour... please share! :)

flaw:
@meysholdt you can try adding `dataLocality: "best-effort"` into the storageClass definition. 

See https://longhorn.io/docs/1.2.4/high-availability/data-locality/

Note that you can also edit that setting from the Longhorn UI if that makes it easier to test.

meysholdt:
That worked. Thank you!

w13915984028:
connect with #2659

Issue #857, opened 2021-05-12
[BUG] Chart-wrong ConfigMap for longhorn StorageClass after upgrading


joy717:
**Describe the bug**
The longhorn-storageclass-override.yaml will be overridden by charts/longhorn after upgrading.

**To Reproduce**
1. install via helm chart.  
`kubectl get cm -n longhorn-system longhorn-storageclass -o yaml`. 
 will find "migratable" property.
2. upgrade via helm. 
3. `kubectl get cm -n longhorn-system longhorn-storageclass -o yaml`.  
now  "migratable" property is missing.

**Expected behavior**
longhorn-storageclass-override.yaml should work correctly after upgrading.

**Support bundle**

**Environment:**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

gitlawr:
The storageclass-override stuff is kind of a workaround due to the release time frame. It would be more straightforward if we can make the parameter configurable in the Longhorn chart. I've opened a Longhorn issue for the request: https://github.com/longhorn/longhorn/issues/2594

innobead:
> The storageclass-override stuff is kind of a workaround due to the release time frame. It would be more straightforward if we can make the parameter configurable in the Longhorn chart. I've opened a Longhorn issue for the request: [longhorn/longhorn#2594](https://github.com/longhorn/longhorn/issues/2594)

Please check https://github.com/longhorn/longhorn/issues/2594#issuecomment-841667027. I would suggest just using an individual storage class to avoid this kind of coupling/dependency.

cc @joshimoo

gitlawr:
To test:
Dev mode
1. Install Harvester by [helm chart](https://docs.harvesterhci.io/v0.2/dev/dev-mode/#option-1-install-using-helm)
2. Check the params in the longhorn storageclass: `kubectl get sc longhorn -o yaml`
3. Upgrade using helm: 
```
helm upgrade harvester harvester \
--namespace harvester-system \
--set longhorn.enabled=true \
--set a=b
```
4. Check the params in the longhorn storageclass: `kubectl get sc longhorn -o yaml`

HCI mode
1. Install harvester by ISO image
2. Check the params in the longhorn storageclass: `kubectl get sc longhorn -o yaml`
3. Upgrade helm release by `kubectl -n kube-system edit helmchart harvester`, then add `a: b` to `spec.set`
4. Wait and check the storageclass params

[...]

PR #2335, opened 2022-05-31
HEP: Support Storage Tiering [CI SKIP]

futuretea:

preview: https://github.com/futuretea/harvester/blob/storage-tiering/enhancements/20220531-storage-tiering.md

**Related Issue:**
https://github.com/harvester/harvester/issues/2147

**Test plan:**


futuretea:
@yasker @guangbochen @bk201 @Vicente-Cheng, add a new commit https://github.com/harvester/harvester/pull/2335/commits/448c3f0e63abe280dd8f35a0c34919469e80c486 for:
1. If `extraStorageClassParameters` are allowed to be modified, the backend deletion and reconstruction of the storageClass being used may cause unpredictable consequences.
2. And it is difficult to know from the storageClass information what the parameters were when the PVC was created.
This makes displaying diskSelector and nodeSelector on the UI potentially inaccurate.

3. Currently, the URL of image is also not allowed to be modified by the user on the UI
image

---
So users should not be allowed to modify the `extraStorageClassParameters` contents.
For different scheduling rules for the same image file, users need to create different images.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
+# Storage Tiering
+
+## Summary
```
guangbochen, line 3: A summary is required, which can be one line paragraph.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
++ Node tag selectors
++ Disk tag selectors
+
+### Non-goals [optional]
+
+## Proposal
+
+### User Stories
+
+Use mixed different performance levels storage hardware to
```
guangbochen:
```suggestion
Use a mix of storage hardware at different performance levels to 
```

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+
++ optimize cost
++ predictable performance
```
guangbochen:
```suggestion
+ provide predictable performance
```

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+
+### User Experience In Detail
+
+#### 1 Node Tag
+
+##### 1.1 Edit Node Tag
```
guangbochen, line 32: You can add text to the placeholder images to provide a better design insight.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+
+![image](./20220531-storage-tiering/edit-disk-tag.png)
+
+##### 2.2 Show Disk Tag
+`Hosts` > host > `Storage`
+
+![image](./20220531-storage-tiering/show-disk-tag.png)
+
+#### 3 Volume
+Since disk tag and node tag can have many combinations and the pvc's storageClass must exist before a pvc be created, we need to let users select storageClass during volume creating.
```
guangbochen: Since there can be various combinations of disk tags and node tags, and the `storageClass` of pvc must exist before the pvc is created, we need to let the user select the `storageClass` when creating a volume.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+
+Users can access [Embedded Rancher UI](https://docs.harvesterhci.io/v1.0/troubleshooting/harvester/#access-embedded-rancher) to manage storageClass, or we can add Harvester's `StorageClass` page.
```
guangbochen: I think we should add the Harvester `StorageClass` page since `Embedded Rancher UI` is just for debugging purposes.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+
+##### 3.1 Create Volume
+Select storageClass (default to longhorn).
```
guangbochen: Need to consider using the Harvester-specific storage class as the default storage class(e.g., due to https://github.com/harvester/harvester/pull/2235 concerns)

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+
++ `Volumes` > `Create`（source=New), backing image storageClass need to be filtered out.
+![image](./20220531-storage-tiering/create-volume.png)
++ `Virtual Machines` > `Create` > `Volumes` > `Add Volume`, backing image storageClass need to be filtered out.
+![image](./20220531-storage-tiering/vm-add-volume.png)
```
guangbochen: If those are empty volumes, they should not contain any backing image keywords here.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
++ `Snapshots` > snapshot > `Restore`
+  - For snapshot of an image volume, if source storageClass is not found, it can not be restored.
+  - For snapshot of other volume, only show storageClasses those have the same provisioner of source storageClass and choose source storageClassName by default. If source storageClass is not found, need show a warning message.
+  - For the above two points, save source provisioner, storageClassName,ImageID in snapshot's annotations during volume snapshot.
```
guangbochen, line 69: we should add those validation checks to both webhook and API actions if needed.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+
+##### 3.2 Show Volume
+```
+if provisioner is driver.longhorn.io
+  show diskSelector and nodeSelector 
+else
+  show storageClassName
+fi
+```
```
guangbochen: can u also add an example here of the pseudo-code, thanks?

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+
++ `Volumes` > volume
+![image](./20220531-storage-tiering/show-volume.png)
+
+#### 4 Image
+Since the storageClass of backing image is created dynamically, we only need to let users select diskSelector and nodeSelector during vm image creating.
```
guangbochen:
```suggestion
The `storageClass` of the backing image is automatically created by the Harvester controller and bound to its disk and node selectors, so we just need to ask the user to select `diskSelector` and `nodeSelector` during VM image creation.
```

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+### API changes
+
+Add a new field: `extraStorageClassParameters`, type `map[string]string`, to `harvesterhci.io.virtualmachineimages.spec`.
+
+If `extraStorageClassParameters` are allowed to be modified, the backend deletion and reconstruction of the storageClass being used may cause unpredictable consequences.
+And it is difficult to know from the storageClass information what the parameters were when the PVC was created.
+This makes displaying diskSelector and nodeSelector on the UI potentially inaccurate.
+Currently, the URL of image is also not allowed to be modified by the user on the UI
+So users should not be allowed to modify the `extraStorageClassParameters` contents.
+For different scheduling rules for the same image file, users need to create different images.
```
guangbochen:
```suggestion
The `extraStorageClassParameters` will not allow any update because deleting and rebuilding the in-use `storageClass` will lead to unpredictable consequences. In addition, this also makes the `diskSelector` and `nodeSelector` information displayed on the UI inconsistent with the previously created volumes.

If you want to configure different scheduling rules for the same image, you must create VM images with different names.
```

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+    migratable: "true"
+    numberOfReplicas: "3"
+    staleReplicaTimeout: "30"
+```
+
+## Design
+
+### Implementation Overview
+
+1. Specify diskSelector and nodeSelector in `spec.extraStorageClassParameters` when vm image creating or updating.
```
guangbochen: 1. should `updating` be omitted here? 
2. maybe creating an image with the same rules, URL and namespace should be prevented.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+2. The `vm image mutator` in `harvester-webhook` will autofill other extraStorageClassParameters according to the `image-storage-class-parameters` settings.
+3. The `vm image controller` in `harvester` will create or update a backing image storageClass according to the `spec.extraStorageClassParameters`.
+
+#### New Settings `image-default-storage-class-parameters`
```
guangbochen: might be good to include https://github.com/harvester/harvester/issues/2027 support

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+spec:
+  checksum: ""
+  displayName: ubuntu-20.04-minimal-cloudimg-amd64.img
+  extraStorageClassParameters:
+    nodeSelector: "large"
+    diskSelector: "hdd"
+    migratable: "true"
+    numberOfReplicas: "3"
+    staleReplicaTimeout: "30"
+```
```
guangbochen: are all those parameters allowed to be configured by the admin users?

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+### Related Issues
+https://github.com/harvester/harvester/issues/2147
+
+## Motivation
+
+### Goals
+Allow admins to define storage tiers based on:
+
++ Node tag selectors
++ Disk tag selectors
```
guangbochen: May need goals for both admin and regular users.

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,142 @@
...
+#### New Settings `image-default-storage-class-parameters`
+
+Default to `{"numberOfReplicas":"3","staleReplicaTimeout":"30","migratable":"true"}`
+```go
+ImageDefaultStorageClassParameters   = NewSetting("image-default-storage-class-parameters", `{"numberOfReplicas":"3","staleReplicaTimeout":"30","migratable":"true"}`)
+```
+
+#### Add vm image mutator to harvester-webhook
+
+![image](./20220531-storage-tiering/backingimage.png)
```
guangbochen: 1. let's remove the other SC for now since they are out of the scope.
2. In the architecture flow, the Harvester will list/watch the `vmimage`.

w13915984028:
@futuretea @guangbochen @bk201 

Question:
 Considering issue: #2659 , #2027  , there are frequent requirements of setting the desired replica counts.
 
should we summarize all the related information to `StorageClass`, which allows addtional param ? instead of `extraStorageClassParameters`.  

When users want something of storage different thant default, they need to create a StorageClass;  when they create image or volume, they need to select a `StorageClass`, either default one (LH, 3 replicas) or the customized one.

`StorageClass.Parameters` is a map, LH will only filter it's wanted param when creating image/volume, after that, the change to `StorageClass` won't affect existing image/volume. Harvester may utilize this flexibility, and add/filer the desired params.

```
type StorageClass struct {
	metav1.TypeMeta `json:",inline"`
	// Standard object's metadata.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
	// +optional
	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`

	// Provisioner indicates the type of the provisioner.
	Provisioner string `json:"provisioner" protobuf:"bytes,2,opt,name=provisioner"`

	// Parameters holds the parameters for the provisioner that should
	// create volumes of this storage class.
	// +optional
	Parameters map[string]string `json:"parameters,omitempty" protobuf:"bytes,3,rep,name=parameters"`
...
```

After exposing and focusing on `StorageClass`, we will not create `StorageClass` internally, which blocks user customization. 

Suppose we have `harvester-default-storageclass-1replica`, `harvester-default-storageclass-2replica`, `harvester-default-storageclass-3replica` predefined.

In Harvester installer, user may select `default replica count`, which is one of the above three, and affects `rancher-monitoring`, `alerting`, `logging` in bootstrap stage.

After that, user always select a `StorageClass` when creating image/volume.

The map based StorageClass.Parameters is also free to expand.

/// 
3. The `vm image controller` in `harvester` will create or update a backing image storageClass according to the `spec.extraStorageClassParameters`.

w13915984028:
https://github.com/harvester/harvester/issues/2088  (a fix to #857) will also disappear when `Harvester` does not use LH default `StorageClass` directly.

Harvester level of `StorageClass` may also give us a choice to map params from `Harvester` to `Longhorn` or other CSI.

futuretea:
@w13915984028 

Each image has its own storageClass
The Harvester will create the corresponding storageClass when creating the image, and these different storageClass will use different backimages.

to support https://github.com/harvester/harvester/issues/2659:
For image volumes, user only need to change `numberOfReplicas` in the settings `image-default-storage-class-parameters` to 2, 
In this way, the `spec.extraStorageClassParameters` of the newly created image will be 2, and the replicas of the volume created using these images will be 2.

For empty volumes, the user needs to create a storageClass with `parameters[numberOfReplicas]=2 `and then use it when creating volumes.

I think your question is about the empty volume part?

w13915984028:
@futuretea You are right.   At the moment, for none-image volumes, users have no chance to create `StorageClass` and use it.      Some addtional information:

#2659   #2027  are more about none-image volumes, as multi replicas means dynamic data syncing when VM is running, it adds pressure to network and storage; on the other hand, image based volume is read only, data sync happens in image creating stage.

The difference between image and volume is, when creating `image`, user selects a `StorageClass` as a templage/base (as the baking image field is different for each image, a new `StorageClass` must be created dynamically); when creating `volume`,it is a specific `SorageClass`.

For both of them, users wish to customize parameters like replica, node selector, disk selector and more.

The unified solution will be exposing `StorageClass` to users, a bunch of issues will be solved with it.

futuretea:
@w13915984028 Yes, refer to https://github.com/harvester/harvester/pull/2335/files#r945481227, we will add a storageClass page in Harvester for user.

Perhaps I didn't make it clear in my documentation, but the design includes exposing the storageClass to the user.

You mentioned allowing users to specify the number of replicas in harvester-installer. I think it's a good idea for users, though it may require further research and design. I prefer to implement this feature through another HEP and let this HEP focus on storage tiering.

w13915984028:
@futuretea Thanks for your explanation. I have following further questions:

(1) What's the exact definition of `storage tiers`.  It needs detailed describtion about how to establish such `tiers`, how to add/remove node/disk from tier, the HA or related features. How to check if an `image/volume` has replica in specific `node`/`disk` when user changing it's tags.

Are `Node tag selectors` and `Disk tag selectors` well supported by LH, what's the `failure policy` when LH can not find the sufficient `node` & `disk` to creat the required replicas ? Are there `fallback` policies, is it `must` or `best-effort` ?

The more tags user adds to construct `tiers`, the more potential issues. It eliminates the flexibility and extensibility of LH. 

(2) Another concern, is that it's better to only expose `StorageClass` to user, the `spec.extraStorageClassParameters` or `spec.StorageClassParameters` is an internal concept, it is a copy of that `StorageClass` at the moment of `Creating`, it's not allowed to change, and the change of `StorageClass` won't affect it.   That is similar to LH, which uses `StoageClass` right when creating something.

w13915984028:
@joshimoo, @PhanLe1010 
Could you please also take a look, thanks.

futuretea:
>   Another concern, is that it's better to only expose StorageClass to user, the spec.extraStorageClassParameters or spec.StorageClassParameters is an internal concept, it is a copy of that StorageClass at the moment of Creating, it's not allowed to change, and the change of StorageClass won't affect it. That is similar to LH, which uses StoageClass right when creating something.

@w13915984028 Do you mean we shouldn't expose `spec.extraStorageClassParameters` to users?
we need to allow users to define diskSelector and nodeSelector when image creating.
refer to https://github.com/harvester/harvester/pull/2335#issuecomment-1153122068, we won't support user change the `spec.extraStorageClassParameters` after image created

w13915984028:
> > Another concern, is that it's better to only expose StorageClass to user, the spec.extraStorageClassParameters or spec.StorageClassParameters is an internal concept, it is a copy of that StorageClass at the moment of Creating, it's not allowed to change, and the change of StorageClass won't affect it. That is similar to LH, which uses StoageClass right when creating something.
> 
> @w13915984028 Do you mean we shouldn't expose `spec.extraStorageClassParameters` to users? we need to allow users to define diskSelector and nodeSelector when image creating. refer to [#2335 (comment)](https://github.com/harvester/harvester/pull/2335#issuecomment-1153122068), we won't support user change the `spec.extraStorageClassParameters` after image created

When we take `diskSelector` and `nodeSelector` as parameters of `StorageClass`, then it is unified. When creating image, user selects a `StorageClass`  (as the base/template, we internally inheriate a new one with the only difference of `backing-image` field); when creating volume, user also selects a `StorageClass`.

The advantages are: users dont' need to set each parameters again and again when creating images;  the user interface will be relatively stable, extending the `StorageClass` won't affect image/volume;  the `replica` is also one param of `StorageClass`, users don't need to set a default replica in setting, it is a global value and afffect all,  but each `StorageClass` has it's dedicated replia.

futuretea:
@w13915984028 Do you mean:
1. Use field `spec.StorageClassParameters` instead of `spec.extraStorageClassParameters` in image CRD
2. remove settings `image-default-storage-class-parameters`
3. User select `StorageClass` instead of selecting `diskSelector` and `nodeSelector`
4. `UI` auto-fill `spec.StorageClassParameters` based on user selected `StorageClass`

w13915984028:
> @w13915984028 Do you mean:
> 
> 1. Use field `spec.StorageClassParameters` instead of `spec.extraStorageClassParameters` in image CRD
> 2. remove settings `image-default-storage-class-parameters`
> 3. User select `StorageClass` instead of selecting `diskSelector` and `nodeSelector`
> 4. `UI` auto-fill `spec.StorageClassParameters` based on user selected `StorageClass`

@futuretea  That's my general idea.  For point 4, it needs a further look of the code implementation, to make sure `kubectl ... ` also works,  the `StorageClass` will be expanded into params in mutator or controller instead of UI.

enhancements/20220531-storage-tiering.md
futuretea: 1. removed `updating`
2. should allow this case if name is different

enhancements/20220531-storage-tiering.md
futuretea: updated according to https://github.com/harvester/harvester/pull/2335#issuecomment-1216292744

enhancements/20220531-storage-tiering.md
futuretea: updated

enhancements/20220531-storage-tiering.md
futuretea: after updated according to https://github.com/harvester/harvester/pull/2335#issuecomment-1216292744
Yes

enhancements/20220531-storage-tiering.md
futuretea: will add validation to API actions

enhancements/20220531-storage-tiering.md
```diff
@@ -0,0 +1,131 @@
...
+
+##### 1.2 Show Node Tag
+`Hosts` > host > `Storage`
+
+![show node tag](./20220531-storage-tiering/show-node-tag.png)
+
+#### 2 Disk Tag
+
+##### 2.1 Edit Disk Tag
+`Hosts` > host > `Edit Config` > `Storage`
```
bk201, line 45: Do we need NDM's help to add some tags during the "add disk" stage?

Vicente-Cheng (commented):
Overall looks good to me.
Have the same questions as above. 
How do we tag disks? Manually or by NDM?

enhancements/20220531-storage-tiering.md
futuretea: @bk201 @Vicente-Cheng 
It's a good idea, we can make it as a new NDM enhancement. but maybe not in v1.1.0? 
cc @guangbochen

bk201 (approved):
LGTM

Vicente-Cheng (approved):
LGTM~ thanks!

guangbochen (approved):

Issue #569, opened 2021-03-25
[FEATURE] VM Clone support without backupstore


yasker:
We should be able to support clone VM along with all of the data.

~~Depends on: https://github.com/longhorn/longhorn/issues/1815~~

After investigation, it depends on https://github.com/longhorn/longhorn/issues/2534 .

yasker:
Due to the dependency on https://github.com/longhorn/longhorn/issues/2534 , move the feature to v1.1.0.

FrankYang0529:
For this feature, I think we can leverage the current design in VM. We use the annotation `harvesterhci.io/volumeClaimTemplates` to create PVC in the harvester controller. We can add a `dataSource` field to it, so users can get a new VM with cloned PVC from the target VM.

So we may need two PR to achieve this:
* A UI PR to add a `clone` button on VM. When users click it, UI will go to the `Virtual Machine: Create` page for users to modify the VM name or other settings. Also, in VM YAML, UI needs to add `dataSource` to each PVC in the `harvesterhci.io/volumeClaimTemplates` annotation.
* A webhook PR to check whether the `dataSource` is valid (`kind` must be `PersistentVolumeClaim` and source PVC is existent ).

cc @guangbochen @weihanglo

FrankYang0529:
I've tested that adding `dataSource` to `harvesterhci.io/volumeClaimTemplates` can work.

Steps:
* Create a VM.
* Login to the VM and add a file `test.txt` with content `123`.
* Create another VM, but don't click the `Create` button directly. Click `Edit as YAML`.
* Add `"dataSource":{"name":"vm-disk-0-xxx","kind":"PersistentVolumeClaim"}` to `harvesterhci.io/volumeClaimTemplates` annotation. The name of PVC is from first VM.
* Click the `Create` button.
* Login to the new VM and check whether the file is existent and the content is the same.

weihanglo:
The proposal looks great! Leveraging CSI volume-cloning is really reasonable. I wonder why @yasker thought it should depend on CSI snapshot but volume cloning.

yasker:
I vaguely remembered some discussion around the area. But @PhanLe1010 is the expert on this part and he should know better.

weihanglo:
FWIW, there are a proposal and an implementation of VM cloning support in KubeVirt:

- https://github.com/kubevirt/community/pull/159
- https://github.com/kubevirt/kubevirt/pull/7336


Though there are some unneglectable limitations described in that proposal:

[...]

PR #2338, opened 2022-05-31
Add clone vm

FrankYang0529:
**Problem:**
Move clone VM logic to backend, so we don't need to put too much business logic in frontend. Also, using [CSI volume cloning](https://longhorn.io/docs/1.2.4/snapshots-and-backups/csi-volume-clone/) to support cloning a VM with data.

**Solution:**
Add a `clone` handler in backend.

**Related Issue:**
https://github.com/harvester/harvester/issues/569

**Test plan:**

2. Create a VM `source-vm` with 3 volumes:
  * Image Volume
  * Volume
  * Container Volume
3. After VM starts, run command `echo "123" > test.txt && sync`.
4. Enable Developer tools & Features.
5. Click `View in API` for the `source-vm`.
Screen Shot 2022-06-02 at 11 39 28 AM
6. Click `clone` button.
Screen Shot 2022-06-02 at 11 40 14 AM
7. Input `target-vm` and click `Show Request` and `Send Request`.
Screen Shot 2022-06-02 at 11 41 27 AM
8. A new VM `target-vm` can run.
9. Check `target-vm` YAML. There should have `dataSource` field in `harvesterhci.io/volumeClaimTemplates` annotation. Also format of PVC name should be `<vm-name>-<volume-name>-<random-string>` and format of secret reference should be `<vm-name>-<random-string>`.
Screen Shot 2022-06-02 at 11 45 40 AM
Screen Shot 2022-06-02 at 11 46 45 AM
10. Run command `cat ~/test.txt` in the `target-vm`. The result should be `123`.

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -19,11 +19,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/rest"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	volumeapi "github.com/harvester/harvester/pkg/api/volume"
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/builder"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
 	"github.com/harvester/harvester/pkg/settings"
@@ -169,6 +171,20 @@ func (h *vmActionHandler) doAction(rw http.ResponseWriter, r *http.Request) erro
 			return apierror.NewAPIError(validation.InvalidBodyContent, "Parameter `volumeName` are required")
 		}
 		return h.removeVolume(r.Context(), namespace, name, input)
+	case cloneVM:
+		var input CloneInput
+		if err := json.NewDecoder(r.Body).Decode(&input); err != nil {
+			return apierror.NewAPIError(validation.InvalidBodyContent, "Failed to decode request body: %v "+err.Error())
+		}
+
+		if input.TargetVM == "" {
+			return apierror.NewAPIError(validation.InvalidBodyContent, "Parameter targetVm are required")
+		}
+
+		if err := h.cloneVM(name, namespace, input); err != nil {
+			return err
+		}
+		return nil
 	default:
 		return apierror.NewAPIError(validation.InvalidAction, "Unsupported action")
 	}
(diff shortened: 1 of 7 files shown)


FrankYang0529:
~For reviewers, currently, we are still waiting for bumping the LH version in Harvester. If you want to test the branch, you can use following steps to build a new ISO.~

1. ~Clone https://github.com/longhorn/longhorn.~
2. ~In longhorn folder, checkout to v1.3.0-rc2 tag.~
3. ~In longhorn folder, run `helm package chart` and a new file `longhorn-1.3.0-rc2.tgz` will be generated.~
4. ~Move `longhorn-1.3.0-rc2.tgz` to `deploy/charts/harvester/charts/` and remove old one `longhorn-1.2.4.tgz`.~
5. ~Update LH version to `1.3.0-rc2` in `deploy/charts/harvester/Chart.yaml`.~
6. ~Update `github.com/longhorn/longhorn-manager` version to `v1.3.0-rc2` in `go.mod`.~
7. ~Run following commands:~
```
go mod tidy
export REPO=<your dockerhub username>
export PUSH=true
make
make build-iso
```
8. ~A new ISO will be in `dist/artifacts` folder.~

Since we don't rely on new features in LH v1.3.0, we can just use LH v1.2.4 to test it.

pkg/api/vm/handler.go
```diff
@@ -626,6 +641,116 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+func (h *vmActionHandler) cloneVM(name string, namespace string, input CloneInput) error {
+	vm, err := h.vmCache.Get(namespace, name)
+	if err != nil {
+		return fmt.Errorf("cannot get vm %s/%s, err: %w", namespace, name, err)
+	}
+	newVM := &kubevirtv1.VirtualMachine{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        input.TargetVM,
+			Namespace:   namespace,
+			Annotations: map[string]string{},
```
weihanglo: Should we need to take care of other customized annotations here? Maybe there is an annotation controlling an important functionality, such like volume-claim-template you deal with below.

pkg/api/vm/handler.go
```diff
@@ -626,6 +641,116 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+						UID:        newVM.UID,
+					},
+				},
+			},
+			Data:       secret.Data,
+			StringData: secret.StringData,
+			Type:       secret.Type,
+		}
+		if _, err = h.secretClient.Create(&newSecret); err != nil {
+			return fmt.Errorf("cannot create secret %+v, err: %w", newSecret, err)
```
weihanglo: Maybe better not print secret contents out to the error message? Guess a secret name is sufficient?

FrankYang0529: Fixed it. Thank you.

pkg/api/vm/handler.go
FrankYang0529: I search source code about VM annotations. We use it in the following cases, so I think it's ok to set it empty here:
* `harvesterhci.io/volumeClaimTemplates`
* VMRestore: https://github.com/harvester/harvester/blob/ff26eab62ad3b1799ee5f4e713a20410c0fb15a1/pkg/controller/master/backup/restore.go#L521-L523
* Upgrade Repo: https://github.com/harvester/harvester/blob/ff26eab62ad3b1799ee5f4e713a20410c0fb15a1/pkg/controller/master/upgrade/upgrade_repo.go#L186-L190

weihanglo (approved):
LGTM

pkg/api/vm/handler.go
bk201: @FrankYang0529 For completeness, should we also copy `labels` here. I can see the GUI send:

```
labels": {
"harvesterhci.io/creator": "harvester"
},
```

bk201 (approved):
LGTM

pkg/api/vm/handler.go
```diff
@@ -626,6 +642,119 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+		return fmt.Errorf("cannot get vm %s/%s, err: %w", namespace, name, err)
+	}
+	newVM := &kubevirtv1.VirtualMachine{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        input.TargetVM,
+			Namespace:   namespace,
+			Annotations: map[string]string{},
+			Labels:      vm.Labels,
+		},
+		Spec: *vm.Spec.DeepCopy(),
```
guangbochen: should we also reset the VM's mac address?

pkg/api/vm/handler.go
```diff
@@ -626,6 +642,119 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+			Type:       secret.Type,
+		}
+		if _, err = h.secretClient.Create(&newSecret); err != nil {
+			return fmt.Errorf("cannot create a new secret from %s/%s, err: %w", namespace, oldSecretName, err)
+		}
+	}
+	return nil
+}
+
+func (h *vmActionHandler) cloneVolumes(newVM *kubevirtv1.VirtualMachine) ([]corev1.PersistentVolumeClaim, map[string]string, error) {
```
guangbochen, line 697: just wondering why don't we create the new PVC and secrets directly first?

pkg/api/vm/handler.go
FrankYang0529: Yes, I update it. Thank you.

pkg/api/vm/handler.go
FrankYang0529: For PVC, we follow the same steps to create PVC for a new VM, so we let the controller handle it. In the future, if we want to change some implementation of creating PVC, we can only update the controller part.

For secret, if we create a secret first, we have to send an update request again to add the owner reference. Currently, for a new VM, we also create VM first and then secrets.

Screen Shot 2022-06-20 at 1 53 56 PM
Screen Shot 2022-06-20 at 1 54 02 PM

guangbochen (approved):
LGTM, thanks.

Issue #2325, opened 2022-05-27
[FEATURE] Harvester control plane should spread across failure domains


janeczku:
**Is your feature request related to a problem? Please describe.**
Currently, Harvester assigns node roles based on the order in which nodes are added to the cluster. 

> The first node always defaults to be a management node of the cluster. When there are three or more nodes, the two other nodes that first joined are automatically promoted to management nodes to form a HA cluster.
Source: https://docs.harvesterhci.io/v1.0/host/host/

This approach fails to ensure that the cluster control plane is spread across infrastructure failure domains, such as racks, network fabric, server rooms etc. 

**Describe the solution you'd like**
To increase system stability, the Harvester node promotion should take into account the topology and failure domains of the clusters.

Specifically, automatic node promotion should ensure that control plane nodes are spread across the failure domains defined by node labels such as [topology.kubernetes.io/zone](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone) or other custom labels.

Additionally, there could be a feature to manually promote a specific node as management node

futuretea:
**Test plan:**

1. Install first node, the role of this node should be `Management Node`
2. Install second node, the role of this node should be `Compute Node`, the second node shouldn't be promoted to `Management Node`
3. Add label `topology.kubernetes.io/zone=zone1` to the first node
4. Install third node, the second node and third node shouldn't be promoted
5. Add label `topology.kubernetes.io/zone=zone1` to the second node, the second node and third node shouldn't be promoted 
6. Add label `topology.kubernetes.io/zone=zone3` to the third node, the second node and third node shouldn't be promoted
7. Change the value of label `topology.kubernetes.io/zone` from `zone1` to `zone2` in the second node, the second node and third node will be promoted to `Management Node` one by one

TachunLin:
Verified **fixed** on `master-aa3d95d2-head` (7/15). Close this issue.

### Result
Checked can pass the following test scenarios. 

1. Install **first** node, the role of this node should be `Management Node`
  ```
  harv-0715-node1:~ # kubectl get nodes
  NAME              STATUS   ROLES                       AGE     VERSION
  harv-0715-node1   Ready    control-plane,etcd,master   8m19s   v1.21.11+rke2r1
  ```
2. Install **second** node, the role of this node should be `Compute Node`, the second node shouldn't be promoted to `Management Node`
  ```
  harv-0715-node1:~ # kubectl get nodes
[...]

PR #2339, opened 2022-05-31
added topology zone awareness to harvester control plane

joeyloman:
**Problem:**
Currently the Harvester Control Plane is not topology/zone aware. 

**Solution:**
This PR makes the controller promotion code topology/zone aware. If the topology.kubernetes.io/zone label is set it will form a control plane based on three different zones. If there are less then three zones configured the number of control plane nodes will be limited to the number of zones, this is expected behavior. If there are no zone labels set, it will default in the oldest node behavior.

I tested this code in a running environment and also during a fresh deployment. When the os.labels.topology.kubernetes.io/zone is set in the Harvester create and join configs it will form a zone aware control plane during the deployment as well.

**Related Issue:**
[2325](https://github.com/harvester/harvester/issues/2325)

### DIFF
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -10,6 +10,7 @@ import (
 	ctlbatchv1 "github.com/rancher/wrangler/pkg/generated/controllers/batch/v1"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/name"
+	"github.com/rancher/wrangler/pkg/slice"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -246,53 +247,67 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
 // If the cluster doesn't need to be promoted, return nil
 func selectPromoteNode(nodeList []*corev1.Node) *corev1.Node {
 	var (
-		managementNumber     int
-		managementRoleNumber int
-		canPromoteNumber     int
-		promoteNode          *corev1.Node
+		managementNumber               int
+		promoteNode                    *corev1.Node
+		managementNodeTopologyZoneList []string
 	)
+
+	for _, node := range nodeList {
+		if isPromoteStatusIn(node, PromoteStatusRunning, PromoteStatusFailed, PromoteStatusUnknown) {
+			// wait until the node promotion is completed or the failed or unknown status is cleared
+			return nil
+		} else if !isManagementRole(node) && isPromoteStatusIn(node, PromoteStatusComplete) {
+			// worker promotion is complete but node is not yet labeled as a management node
+			return nil
+		} else if isManagementRole(node) {
+			managementNumber++
+
+			if topologyZone := node.Labels[corev1.LabelTopologyZone]; topologyZone != "" {
+				managementNodeTopologyZoneList = append(managementNodeTopologyZoneList, topologyZone)
+			}
+		}
+
+		// return if there are already enough management nodes
+		if managementNumber == defaultSpecManagementNumber {
+			return nil
+		}
+	}
+
+	// return if the management node count is equal to the total amount of nodes (there are no more nodes left to promote)
+	nodeCount := len(nodeList)
+	if managementNumber == nodeCount {
+		return nil
+	}
 
 	promoteNode = nil
 	for _, node := range nodeList {
-		if isManagementRole(node) {
-			managementNumber++
-			managementRoleNumber++
-		} else if hasPromoteStatus(node) {
-			managementNumber++
-		} else if isHarvesterNode(node) && isHealthyNode(node) {
-			canPromoteNumber++
-			if promoteNode == nil || node.CreationTimestamp.Before(&promoteNode.CreationTimestamp) {
-				promoteNode = node
+		if !isHealthyNode(node) {
+			// decrease the nodeCount if a node is not healthy
+			nodeCount--
+		}
+
+		// return if the nodeCount is below the defaultSpecManagementNumber
+		if nodeCount < defaultSpecManagementNumber {
+			return nil
+		}
+
+		if !isManagementRole(node) && isHarvesterNode(node) && isHealthyNode(node) {
+			if len(managementNodeTopologyZoneList) > 0 {
+				// only promote the node if the label is set to a new zone
+				if topologyZone := node.Labels[corev1.LabelTopologyZone]; topologyZone != "" &&
+					!slice.ContainsString(managementNodeTopologyZoneList, topologyZone) {
+					return node
+				}
+			} else {
+				if promoteNode == nil || node.CreationTimestamp.Before(&promoteNode.CreationTimestamp) {
+					promoteNode = node
+				}
 			}
 		}
 	}
 
-	// waiting for the other node completed
-	if managementRoleNumber != managementNumber {
-		return nil
-	}
-
-	// there is no need to promote if the spec number has been reached
-	specManagementNumber := getSpecManagementNumber(len(nodeList))
-	promoteNodeNumber := specManagementNumber - managementNumber
-	if promoteNodeNumber <= 0 {
-		return nil
-	}
-
-	// make sure have enough nodes can be promote
-	if canPromoteNumber < promoteNodeNumber {
-		return nil
-	}
-
+	// promote the oldest node
 	return promoteNode
-}
-
-// getSpecMasterNumber get spec management number by all node number
-func getSpecManagementNumber(nodeNumber int) int {
-	if nodeNumber < 3 {
-		return 1
-	}
-	return defaultSpecManagementNumber
 }
 
 // isHealthyNode determine whether it's an healthy node
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -292,13 +292,8 @@ func selectPromoteNode(nodeList []*corev1.Node) *corev1.Node {
 		}
 
 		if !isManagementRole(node) && isHarvesterNode(node) && isHealthyNode(node) {
-			if len(managementNodeTopologyZoneList) > 0 {
-				// only promote the node if the label is set to a new zone
-				if topologyZone := node.Labels[corev1.LabelTopologyZone]; topologyZone != "" &&
-					!slice.ContainsString(managementNodeTopologyZoneList, topologyZone) {
-					return node
-				}
-			} else {
+			topologyZone := node.Labels[corev1.LabelTopologyZone]
+			if len(managementNodeTopologyZoneList) > 0 && topologyZone != "" && !slice.ContainsString(managementNodeTopologyZoneList, topologyZone) || len(managementNodeTopologyZoneList) == 0 {
 				if promoteNode == nil || node.CreationTimestamp.Before(&promoteNode.CreationTimestamp) {
 					promoteNode = node
 				}
--- a/pkg/controller/master/node/promote_controller_test.go
+++ b/pkg/controller/master/node/promote_controller_test.go
@@ -131,6 +131,7 @@ var (
 	w1z1 = NewDefaultNodeBuilder().Name("w-1-z1").Zone("zone1").Harvester().Worker()
 	w2z2 = NewDefaultNodeBuilder().Name("w-2-z2").Zone("zone2").Harvester().Worker()
 	w3z3 = NewDefaultNodeBuilder().Name("w-3-z3").Zone("zone3").Harvester().Worker()
+	w4z3 = NewDefaultNodeBuilder().Name("w-4-z3").Zone("zone3").Harvester().Worker()
 )
 
 func Test_selectPromoteNode(t *testing.T) {
@@ -433,6 +434,13 @@ func Test_selectPromoteNode(t *testing.T) {
 			name: "two management in zone1 and zone2 and one worker in zone3",
 			args: args{
 				nodeList: []*corev1.Node{m1z1, m2z2, w3z3},
+			},
+			want: w3z3,
+		},
+		{
+			name: "two management in zone1 and zone2 and two worker in zone3",
+			args: args{
+				nodeList: []*corev1.Node{m1z1, m2z2, w4z3, w3z3},
 			},
 			want: w3z3,
 		},
--- a/pkg/controller/master/node/promote_controller_test.go
+++ b/pkg/controller/master/node/promote_controller_test.go
@@ -34,6 +34,11 @@ func (n *NodeBuilder) Name(name string) *NodeBuilder {
 	return n
 }
 
+func (n *NodeBuilder) Zone(name string) *NodeBuilder {
+	n.node.Labels[corev1.LabelTopologyZone] = name
+	return n
+}
+
 func (n *NodeBuilder) Harvester() *NodeBuilder {
 	n.node.Labels[HarvesterManagedNodeLabelKey] = "true"
 	return n
@@ -65,6 +70,11 @@ func (n *NodeBuilder) Failed() *NodeBuilder {
 	return n
 }
 
+func (n *NodeBuilder) Unknown() *NodeBuilder {
+	n.node.Annotations[HarvesterPromoteStatusAnnotationKey] = PromoteStatusUnknown
+	return n
+}
+
 func (n *NodeBuilder) NotReady() *NodeBuilder {
 	ready := corev1.NodeCondition{
 		Type:   corev1.NodeReady,
@@ -75,6 +85,7 @@ func (n *NodeBuilder) NotReady() *NodeBuilder {
 }
 
 var (
+	// normal nodes
 	mu1 = NewDefaultNodeBuilder().Name("m-unmanaged-1").Management()
 
 	m1 = NewDefaultNodeBuilder().Name("m-1").Harvester().Management()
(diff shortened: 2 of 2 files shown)


pkg/controller/master/node/promote_controller.go
```diff
@@ -34,6 +34,8 @@ const (
 	HarvesterPromoteNodeLabelKey        = HarvesterLabelAnnotationPrefix + "promote-node"
 	HarvesterPromoteStatusAnnotationKey = HarvesterLabelAnnotationPrefix + "promote-status"
 
+	HarvesterTopologyZoneLabel = "topology.kubernetes.io/zone"
```
futuretea: can be replaced by https://github.com/harvester/harvester/blob/ff26eab62ad3b1799ee5f4e713a20410c0fb15a1/vendor/k8s.io/api/core/v1/well_known_labels.go#L22

pkg/controller/master/node/promote_controller.go
```diff
@@ -246,12 +248,21 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
...
-		promoteNode          *corev1.Node
+		managementNumber           int
+		managementRoleNumber       int
+		canPromoteNumber           int
+		promoteNode                *corev1.Node
+		masterNodeTopologyZoneList []string
 	)
 
+	for _, node := range nodeList {
+		if node.ObjectMeta.Labels[HarvesterTopologyZoneLabel] != "" {
```
futuretea:
```go
	if topologyZone := node.Labels[corev1.LabelTopologyZonel]; topologyZone != "" && isManagementRole(node) {
		managementNodeTopologyZoneList= append(managementNodeTopologyZoneList, topologyZone)
	}
```

pkg/controller/master/node/promote_controller.go
```diff
@@ -246,12 +248,21 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
...
+			if isManagementRole(node) {
+				masterNodeTopologyZoneList = append(masterNodeTopologyZoneList, node.ObjectMeta.Labels[HarvesterTopologyZoneLabel])
```
futuretea, line 261: node.Labels

pkg/controller/master/node/promote_controller.go
```diff
@@ -261,8 +272,18 @@ func selectPromoteNode(nodeList []*corev1.Node) *corev1.Node {
 			managementNumber++
 		} else if isHarvesterNode(node) && isHealthyNode(node) {
 			canPromoteNumber++
-			if promoteNode == nil || node.CreationTimestamp.Before(&promoteNode.CreationTimestamp) {
-				promoteNode = node
+
+			if len(masterNodeTopologyZoneList) > 0 {
```
futuretea: import "github.com/rancher/wrangler/pkg/slice"

```go
	if len(managementNodeTopologyZoneList) > 0 {
		if promoteNode == nil {
			// Only promote the node if the label is set to a zone
			if topologyZone := node.Labels[corev1.LabelTopologyZonel]; topologyZone != "" && !slice.ContainsString(managementNodeTopologyZoneList, topologyZone) {
				promoteNode = node
			}
		}
	} else {
		xxxx
	}
```

pkg/controller/master/node/promote_controller.go
```diff
@@ -332,6 +353,16 @@ func isManagementRole(node *corev1.Node) bool {
 	return false
 }
 
+func isNewZone(MasterNodeTopologyZoneCacheList []string, zone string) bool {
```
futuretea: can be replaced by ContainsString in github.com/rancher/wrangler/pkg/slice

pkg/controller/master/node/promote_controller.go
```diff
@@ -246,12 +248,21 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
...
 	var (
-		managementNumber     int
-		managementRoleNumber int
-		canPromoteNumber     int
-		promoteNode          *corev1.Node
+		managementNumber           int
+		managementRoleNumber       int
+		canPromoteNumber           int
+		promoteNode                *corev1.Node
+		masterNodeTopologyZoneList []string
```
futuretea: s/masterNodeTopologyZoneList/managementNodeTopologyZoneList/

weihanglo: shouldn't it be a set instead?

https://pkg.go.dev/k8s.io/apimachinery@v0.21.5/pkg/util/sets#String

joeyloman:
@futuretea Thanks for the code review and the improvement suggestions. It should be ok now. I also retested the code successfully in a running cluster and during a redeploy.

futuretea:
@joeyloman Thanks, can you help to add some tests  to https://github.com/harvester/harvester/blob/master/pkg/controller/master/node/promote_controller_test.go ?

pkg/controller/master/node/promote_controller.go
joeyloman: @weihanglo what would be the benefit of using a set?

weihanglo: Never mind. I assume the list won't be too large. A slice is sufficient.

pkg/controller/master/node/promote_controller.go
```diff
@@ -246,12 +247,19 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
...
-		canPromoteNumber     int
-		promoteNode          *corev1.Node
+		managementNumber               int
+		managementRoleNumber           int
+		canPromoteNumber               int
+		promoteNode                    *corev1.Node
+		managementNodeTopologyZoneList []string
 	)
 
+	for _, node := range nodeList {
```
w13915984028, line 255: For performance, and also to eliminate the "complex code", this function is better to re-design:

1. following check should be done asap, because most of the time, it has already reached the state
```
	// waiting for the other node completed
	if managementRoleNumber != managementNumber {
		return nil
	}

	// there is no need to promote if the spec number has been reached
	specManagementNumber := getSpecManagementNumber(len(nodeList))
	promoteNodeNumber := specManagementNumber - managementNumber
	if promoteNodeNumber <= 0 {
		return nil
	}
```

2. consider 2 times of iteration:
(1) first round, check mgmt node role & promopt node,   by the way, get the managementNodeTopologyZoneList
then check number, if reached, or has active promotion, return;  otherwise, go next

(2) second round, find the desired node, with check of topology

this way, the complex code will also be solved.

joeyloman: @w13915984028 I totally agree on this. I rewrote the function like you suggested and it's much simpeler now. The only thing I don't know the reason of is why we are promoting the oldest node instead of the first in line?

futuretea: @joeyloman Selecting the oldest node ensures the predictability of the promote behavior of the node. The user can always know the promote order of the node, that is, the order in which they add the node to the cluster.

futuretea: ~A cluster will usually only do a node promote once, and the node promotion usually occurs  when size of the cluster is small,~ so correctness is more important than performance. The top priority should be evaluation of unit tests. Refactoring generally should not break existing unit tests. If some existing unit tests don't make sense, they should be discussed first. For the `Complex Method` fix, personally, it's ~optional~ nice to have.

w13915984028: When a function is complained as `Complex Method`, it means it's very difficult to make sure the correctness. Such method is hard to understand and maintain.

Another reason to pay attention to the performance, is that, `selectPromoteNode` is always called `OnNodeChanged` ,  there could be frequent `OnNodeChanged`, but for most of the time, there is noting related to node promotion (e.g, 3 management nodes,  100+ work nodes). The performance lose in `selectPromoteNode` deserves an optimization.

```
func (h *PromoteHandler) OnNodeChanged(key string, node *corev1.Node) (*corev1.Node, error) {
		return node, nil
	}

	nodeList, err := h.nodeCache.List(labels.Everything())
	if err != nil {
		return nil, err
	}

	promoteNode := selectPromoteNode(nodeList)
	if promoteNode == nil {
		return node, nil
	}
```

futuretea: @w13915984028 Sorry, I missed the `OnNodeChange`. Thank you for pointing that out. Improving `selectPromoteNode` performance and reducing complexity while ensuring correctness is certainly great.

futuretea: @joeyloman Not sure two iteration will solve the complexity problem completely, maybe need to split this method into some small methods too, for example:

```go
func selectPromoteNode(nodeList []*corev1.Node) *corev1.Node {
	managementNumber, managementRoleNumber, managementNodeTopologyZoneList := getManagementNodeInfo(nodeList)

	// waiting for the other node completed
	if managementRoleNumber != managementNumber {
		return nil
	}

	// there is no need to promote if the spec number has been reached
	specManagementNumber := getSpecManagementNumber(len(nodeList))
	promoteNodeNumber := specManagementNumber - managementNumber
	if promoteNodeNumber <= 0 {
		return nil
	}

	return selectPromoteNodeOne(nodeList, managementNodeTopologyZoneList, promoteNodeNumber)
}

func selectPromoteNodeOne(nodeList []*corev1.Node, managementNodeTopologyZoneList []string, promoteNodeNumber int) *corev1.Node {
	var (
		canPromoteNumber int
		promoteNode      *corev1.Node
	)

	for _, node := range nodeList {
		if isManagementRole(node) || hasPromoteStatus(node) || !isHarvesterNode(node) || !isHealthyNode(node) {
			continue
		}
		canPromoteNumber++
		if len(managementNodeTopologyZoneList) > 0 && promoteNode == nil {
			// Only promote the node if the label is set to a zone
			if topologyZone := node.Labels[corev1.LabelTopologyZone]; topologyZone != "" && !slice.ContainsString(managementNodeTopologyZoneList, topologyZone) {
				promoteNode = node
			}
		} else {
			if promoteNode == nil || node.CreationTimestamp.Before(&promoteNode.CreationTimestamp) {
				promoteNode = node
			}
		}
	}

	// make sure have enough nodes can be promoted
	if canPromoteNumber < promoteNodeNumber {
		return nil
	}

	return promoteNode
}

func getManagementNodeInfo(nodeList []*corev1.Node) (managementNumber int, managementRoleNumber int, managementNodeTopologyZoneList []string) {
	for _, node := range nodeList {
		if isManagementRole(node) {
			if topologyZone := node.Labels[corev1.LabelTopologyZone]; topologyZone != "" {
				managementNodeTopologyZoneList = append(managementNodeTopologyZoneList, topologyZone)
			}
			managementNumber++
			managementRoleNumber++
		} else if hasPromoteStatus(node) {
			managementNumber++
		}
	}
	return
}
```
cc @w13915984028

futuretea (commented):
Just suggestion: https://github.com/harvester/harvester/pull/2339#discussion_r896295297, and the unit tests need to cover the logic about the topology zone. You can use `make test` to run the unit tests locally. Thanks.

joeyloman:
@futuretea / @w13915984028: I added the unit tests to cover the topology zone logic and also fixed some test failures on the refactored method. Everything works like it should now. What do you think about this approach instead of the old method in terms of performance? It returns quicker if a node is in promotion or if there are already 3 management nodes. 

About the Complex method: can we park/move this to another PR in the future?

pkg/controller/master/node/promote_controller.go
```diff
@@ -246,53 +247,63 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
 // If the cluster doesn't need to be promoted, return nil
 func selectPromoteNode(nodeList []*corev1.Node) *corev1.Node {
 	var (
-		managementNumber     int
-		managementRoleNumber int
```
futuretea, line 250: we also need to add two cases:
```go
   wu1 = NewDefaultNodeBuilder().Name("w-unknown-1").Harvester().Unknown().Worker()
	wc1 = NewDefaultNodeBuilder().Name("w-complete-1").Harvester().Complete().Worker()

{
			name: "one management and one promote unknown worker and one worker",
			args: args{
				nodeList: []*corev1.Node{m1, wu1, w1},
			},
			want: nil,
		},
		{
			name: "one management and one promoted worker and one worker",
			args: args{
				nodeList: []*corev1.Node{m1, wc1, w1},
			},
			want: nil,
		},
```
and these cases will failed after remove variable `managementRoleNumber` and turn method `hasPromoteStatus` to unused

futuretea:
```go
func (n *NodeBuilder) Unknown() *NodeBuilder {
	n.node.Annotations[HarvesterPromoteStatusAnnotationKey] = PromoteStatusUnknown
	return n
}
```

joeyloman: Yes they are covered now in the tests (including the zone aware ones) and also in the selectPromoteNode method.

pkg/controller/master/node/promote_controller.go
```diff
@@ -246,53 +247,68 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
...
+		managementNodeTopologyZoneList []string
 	)
 
-	promoteNode = nil
 	for _, node := range nodeList {
-		if isManagementRole(node) {
-			managementNumber++
-			managementRoleNumber++
-		} else if hasPromoteStatus(node) {
+		if isPromoteStatusIn(node, PromoteStatusRunning) || isPromoteStatusIn(node, PromoteStatusFailed) ||
```
futuretea: if isPromoteStatusIn(node, PromoteStatusRunning, PromoteStatusFailed, PromoteStatusUnknown) {
}

futuretea (approved):
LGTM, @joeyloman can u help to squash commits?

joeyloman:
> LGTM, @joeyloman can u help to squash commits?

Yes done. Thanks for all the feedback and assistance 👍

guangbochen (commented):
The code changes LGTM, can we please refactor the selectPromoteNode a little bit to resolve the `Complex Method` CI error, thanks.

janeczku:
@guangbochen Could you or someone from the team provide a suggestion on how to refactor the method?

futuretea:
We can ignore the complexity check prompt for this method
- If else nesting is necessary for the logic itself
- If we split the method, there are many places where it cannot be returned immediately, which affects performance

guangbochen:
> We can ignore the complexity check prompt for this method
> 
> * If else nesting is necessary for the logic itself
> * If we split the method, there are many places where it cannot be returned immediately, which affects performance

That makes sense, thanks for the explanation.

guangbochen (approved):

pkg/controller/master/node/promote_controller.go
```diff
@@ -246,53 +247,67 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
...
 
-// getSpecMasterNumber get spec management number by all node number
-func getSpecManagementNumber(nodeNumber int) int {
-	if nodeNumber < 3 {
-		return 1
+		if !isManagementRole(node) && isHarvesterNode(node) && isHealthyNode(node) {
+			if len(managementNodeTopologyZoneList) > 0 {
+				// only promote the node if the label is set to a new zone
+				if topologyZone := node.Labels[corev1.LabelTopologyZone]; topologyZone != "" &&
+					!slice.ContainsString(managementNodeTopologyZoneList, topologyZone) {
```
w13915984028: strictly speaking, here it still can not return, need to get the smallest timestamp NODE in same new zone

janeczku: Yes, looks like this needs to be changed.

janeczku: This should work @w13915984028 ?

Before:
```
			if topologyZone := node.Labels[corev1.LabelTopologyZone]; topologyZone != "" &&
				!slice.ContainsString(managementTopologyZoneList, topologyZone) {
				return node
			}
```
After:
```
			if topologyZone := node.Labels[corev1.LabelTopologyZone]; topologyZone != "" &&
				slice.ContainsString(managementTopologyZoneList, topologyZone) {
				continue
			}
```

w13915984028:
A re-designed assistant data structure is needed:

e.g.

(1): running node, and node 9 is down
image

(2) select a new one to prompt
image

(3) the possible candidate
image


The possible data structure:

2 map, ZONE is key,  one map to store the oldes node index;   one map to store the management node count

Your will sort the map 2, to get a fewest mgmt NODE ZONE, then get a node index from map 1.

Here, ZONE B, ZONE D will be checked and select the oldest NODE.

image

janeczku:
@guangbochen @futuretea  Here is a replacement PR implementing the comments from @w13915984028  
This PR also removes the `Complex Code` error

https://github.com/harvester/harvester/pull/2457

pkg/controller/master/node/promote_controller.go
futuretea: @janeczku 
in selectPromoteNode:
```go
		if !isManagementRole(node) && isHarvesterNode(node) && isHealthyNode(node) {
			topologyZone := node.Labels[corev1.LabelTopologyZone]
			if len(managementNodeTopologyZoneList) > 0 && topologyZone != "" && !slice.ContainsString(managementNodeTopologyZoneList, topologyZone) || len(managementNodeTopologyZoneList) == 0 {
				if promoteNode == nil || node.CreationTimestamp.Before(&promoteNode.CreationTimestamp) {
					promoteNode = node
				}
			}
		}
```


add unit test:
```go
	w4z3 = NewDefaultNodeBuilder().Name("w-4-z3").Zone("zone3").Harvester().Worker()
```
```go
		{
			name: "two management in zone1 and zone2 and two worker in zone3",
			args: args{
				nodeList: []*corev1.Node{m1z1, m2z2, w4z3, w3z3},
			},
			want: w3z3,
		},
```

futuretea: added commit https://github.com/harvester/harvester/pull/2339/commits/5b39f8d31fdba5df986428041b8e9edefc7e2855 to this PR

joeyloman: LGTM

futuretea:
@guangbochen @w13915984028 added a new commit https://github.com/harvester/harvester/pull/2339/commits/5b39f8d31fdba5df986428041b8e9edefc7e2855 to fix this comment https://github.com/harvester/harvester/pull/2339#discussion_r914660181 in this PR

w13915984028:
LGTM, thanks.

w13915984028 (approved):
LGTM, thanks.

PR #2358, opened 2022-06-07
Support storage tiering

futuretea:

HEP: https://github.com/harvester/harvester/pull/2335

**Related Issue:**
https://github.com/harvester/harvester/issues/2147

**Test plan:**

Build images:
```
export REPO=<your registry repo name>
export PUSH=true
make
```
Or use my prebuild image: `futuretea/harvester:st-head` and `futuretea/harvester-webhook:st-head`

Need a Harvester cluster in v1.0.2 version and pre-create some images in the cluster for test backward compatibility.

Upgrade CRDs:
```
cd deploy/charts/harvester-crd
helm upgrade harvester-crd ./ -n harvester-system
```

Upgrade Harvester components:
```
kubectl -n harvester-system set image deployment/harvester apiserver=futuretea/harvester:st-head
kubectl -n harvester-system set image deployment/harvester-webhook harvester-webhook=futuretea/harvester-webhook:st-head

```

in one single node cluster, need a force harvester-webhook restart 
```
kubectl -n harvester-system scale deploy harvester-webhook --replicas=0
kubectl -n harvester-system scale deploy harvester-webhook --replicas=1
```

Add node/disk tags  using the Longhorn UI https://docs.harvesterhci.io/v1.0/troubleshooting/harvester/#access-embedded-longhorn

1. Node -> Select one node -> Edit Node and Disks
2. Click +New Node Tag or +New Disk Tag to add new tags.

Test cases:
1. Old version images and storageClasses and backingimages should not be changed.

2. Edit any old version image, change Description from UI

- The storageClass and backingimage of the image should not be changed.

3. Edit any old version image, add some parameters to `spec.storageClassParameters`.

- this request should be refused by harvester-webhook

4. Create a new image, do not specify `metadata.Annotations["harvesterhci.io/storageClassName"]`

- The field `spec.storageClassParameters` in the image will be mutated and the value is the same as the paramters value of the default storageClass `longhorn` 
- The filed `parameters` in the backingimage storageClass should be the same as `spec.storageClassParameters` in image（key-value "backingimage: xxx" only in the backingimage storageClass )

5. Create a new image, specify `metadata.Annotations["harvesterhci.io/storageClassName"]`
- The field `spec.storageClassParameters` in the image will be mutated and the value is the same as the paramters value of the user specified storageClass 
- The filed `parameters` in the backingimage storageClass should be the same as `spec.storageClassParameters` in image （key-value "backingimage: xxx" only in the backingimage storageClass )

6. create a volume by choose source=image, export the image volume to a new image via `View API` and specify `storageClassName` 

-  the new exported image's `spec.storageClassParameters` should be the same as the paramters value of the storageClass which we specify during the volume export action

7. restore a volume snapshot to a new volume via `View API` and specify `storageClassName` 

-  the new restore volumes `spec.storageClassName` should be the same as the value which we specify during the volume snapshot restore action


futuretea:
Change this PR to draft, need to update image validator and add volume export argument `diskSelecor` and `nodeSelector`

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -45,11 +40,8 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
 		return h.initialize(image)
 	} else if image.Spec.URL != image.Status.AppliedURL {
 		// URL is changed, recreate the storageclass and backingimage
-		if err := h.backingImages.Delete(util.LonghornSystemNamespaceName, getBackingImageName(image), &metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
-			return image, err
-		}
-		if err := h.storageClasses.Delete(getImageStorageClassName(image.Name), &metav1.DeleteOptions{}); err != nil && !errors.IsNotFound(err) {
-			return image, err
+		if err := h.deleteBackingImageAndStorageClass(image); err != nil {
+			return nil, err
```
guangbochen:
```suggestion
			return image, err
```

pkg/api/volumesnapshot/handler.go
```diff
@@ -52,18 +55,36 @@ func (h *ActionHandler) do(rw http.ResponseWriter, r *http.Request) error {
...
 	volumeSnapshot, err := h.snapshotCache.Get(snapshotNamespace, snapshotName)
 	if err != nil {
 		return err
 	}
 
+	sourceStorageClassName := volumeSnapshot.Annotations[util.AnnotationStorageClassName]
+	sourceStorageProvisioner := volumeSnapshot.Annotations[util.AnnotationStorageProvisioner]
+	sourceImageId := volumeSnapshot.Annotations[util.AnnotationImageID]
+	// default to using source storageClass
+	if storageClassName == "" {
```
guangbochen, line 74: Q: Did LH snapshot and restore support to use different storage tiers? also would a VM with multi-disks work with mixed storage?

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -143,6 +144,10 @@ func (v *virtualMachineImageValidator) Update(request *types.Request, oldObj run
 		return nil
 	}
 
+	if !reflect.DeepEqual(newImage.Spec.StorageClassParameters, oldImage.Spec.StorageClassParameters) {
+		return werror.NewInvalidError("storageClassParameters cannot be modified", "spec.storageClassParameters")
```
guangbochen:
```suggestion
		return werror.NewInvalidError("storageClassParameters of the VM Image cannot be modified", "spec.storageClassParameters")
```

pkg/util/volume.go
```diff
@@ -2,13 +2,18 @@ package util
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 const (
 	AnnStorageProvisioner     = "volume.kubernetes.io/storage-provisioner"
 	AnnBetaStorageProvisioner = "volume.beta.kubernetes.io/storage-provisioner"
 )
 
+var (
+	PersistentVolumeClaimsKind = schema.GroupVersionKind{Group: "", Version: "v1", Kind: "PersistentVolumeClaim"}
```
guangbochen, line 14: what about to use  `pvc.GroupVersionKind()` method?

pkg/webhook/resources/virtualmachineimage/mutator.go
```diff
@@ -0,0 +1,116 @@
...
+	newImage := newObj.(*harvesterv1.VirtualMachineImage)
+
+	return m.patchImageStorageClassParams(newImage)
+}
+
+func (m *virtualMachineImageMutator) patchImageStorageClassParams(newImage *harvesterv1.VirtualMachineImage) ([]string, error) {
+	var patchOps types.PatchOps
+
+	storageClassName := newImage.Annotations[util.AnnotationStorageClassName]
+	storageClass, err := m.getStorageClass(storageClassName)
```
guangbochen, line 55: storage class of backing images is a special design in Harvester and I believe its `classParams` like replicas, stale should be consistent across all images. Storage tiers should not apply to the backing images as any VM images should be utilized to create a VM in any nodes. So I would prefer to:
1. create a particular default storage class for the backing images, and it can be modified via the `kubectl` only when needed (but not exposed in the UI by default).
2. define those default values in the Harvester code.

futuretea: the case is using different diskSelectors in different images

pkg/util/volume.go
futuretea:
```go
// GroupVersionKind satisfies the ObjectKind interface for all objects that embed TypeMeta
func (obj *TypeMeta) GroupVersionKind() schema.GroupVersionKind {
	return schema.FromAPIVersionAndKind(obj.APIVersion, obj.Kind)
}
```
if obj.APIVersion == "" or obj .Kind == ""
this method will not work as expected.

Not sure why the apiVersion and Kind objects from cache get are empty sometimes

pkg/api/volumesnapshot/handler.go
futuretea: The answer to both questions is yes

pkg/controller/master/image/vm_image_controller.go
futuretea: updated

pkg/webhook/resources/virtualmachineimage/mutator.go
guangbochen: sure, then as long as we use an appropriate default SC it should be ok, thanks.

guangbochen (approved):
LGTM, thanks.

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -158,32 +150,37 @@ func (h *vmImageHandler) createBackingImage(image *harvesterv1.VirtualMachineIma
...
 
-func getBackingImageName(image *harvesterv1.VirtualMachineImage) string {
-	return fmt.Sprintf("%s-%s", image.Namespace, image.Name)
+func (h *vmImageHandler) deleteStorageClass(image *harvesterv1.VirtualMachineImage) error {
+	return h.storageClasses.Delete(util.GetImageStorageClassName(image.Name), &metav1.DeleteOptions{})
+}
+
+func (h *vmImageHandler) deleteBackingImageAndStorageClass(image *harvesterv1.VirtualMachineImage) error {
+	if err := h.deleteBackingImage(image); err != nil && !errors.IsNotFound(err) {
+		return err
```
Vicente-Cheng, line 180: could we leave some tips for debugging when deleteBackingImage error?

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -158,32 +150,37 @@ func (h *vmImageHandler) createBackingImage(image *harvesterv1.VirtualMachineIma
...
+	}
+	if err := h.deleteStorageClass(image); err != nil && !errors.IsNotFound(err) {
+		return err
```
Vicente-Cheng, line 183: ditto, for deleteStorageClass

pkg/webhook/resources/virtualmachineimage/mutator.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+func (m *virtualMachineImageMutator) Create(request *types.Request, newObj runtime.Object) (types.PatchOps, error) {
+	newImage := newObj.(*harvesterv1.VirtualMachineImage)
+
+	return m.patchImageStorageClassParams(newImage)
+}
+
+func (m *virtualMachineImageMutator) patchImageStorageClassParams(newImage *harvesterv1.VirtualMachineImage) ([]string, error) {
```
Vicente-Cheng, line 51: According to the following discussion.
Do we only allow users to patch the parameters like node/disk selector on storage class in VM Image?
If yes, could we add some restrictions on mergeStorageClassParams()?

pkg/webhook/resources/virtualmachineimage/mutator.go
```diff
@@ -0,0 +1,116 @@
...
+	var patchOps types.PatchOps
+
+	storageClassName := newImage.Annotations[util.AnnotationStorageClassName]
+	storageClass, err := m.getStorageClass(storageClassName)
+	if err != nil {
+		return patchOps, err
```
Vicente-Cheng, line 57: I thought that would return the isNotFound error. 
Could we also add some error messages here to make debugging easy?

pkg/api/volumesnapshot/handler.go
```diff
@@ -52,18 +55,36 @@ func (h *ActionHandler) do(rw http.ResponseWriter, r *http.Request) error {
...
+	}
+	sc, err := h.storageClassCache.Get(storageClassName)
+	if err != nil {
+		return apierror.NewAPIError(validation.InvalidBodyContent, fmt.Sprintf("Can not restore volume snapshot with a storageClass that does not exist: %v", err))
+	}
+	if sourceStorageProvisioner != "" && sc.Provisioner != sourceStorageProvisioner {
+		return apierror.NewAPIError(validation.InvalidBodyContent, "Can not use different provisioner for restoring volume snapshot")
+	}
+	if sourceImageId != "" && storageClassName != sourceStorageClassName {
+		return apierror.NewAPIError(validation.InvalidBodyContent, "Can not use different storageClass for restoring image volume snapshot")
```
Vicente-Cheng, line 85: I do not understand this error.
On the above comment, looks like we could restore with different storage classes?
If yes, why this checking would pop up the APIError?

futuretea: For image volumes, only support restore with same storage classes
For other volumes, support restore  with different storage classes

pkg/webhook/resources/virtualmachineimage/mutator.go
futuretea: Any error here should return
already has log message in https://github.com/harvester/harvester/blob/4a972de680447102c78d8831741a423f674448e2/pkg/webhook/types/admission.go#L107

pkg/controller/master/image/vm_image_controller.go
futuretea: wrangler framework will print the deleteBackingImage error

pkg/controller/master/image/vm_image_controller.go
futuretea: wrangler framework will print the deleteStorageClass error

pkg/webhook/resources/virtualmachineimage/mutator.go
futuretea: Updated. Only allow user patch image storageclass `nodeSelector` `diskSelector` `numberOfReplicas` `staleReplicaTimeout`
https://github.com/harvester/harvester/pull/2358/commits/218dfd23449a2b393e20ac80a40a0072f970cad6

bk201 (approved):
LGTM! I also verify backup/restore path.

pkg/api/volumesnapshot/handler.go
Vicente-Cheng: OK, I see. Thanks for explaining it.

pkg/controller/master/image/vm_image_controller.go
Vicente-Cheng: OK, got it!

pkg/webhook/resources/virtualmachineimage/mutator.go
Vicente-Cheng: Ah, did not trace on the upper level... Thanks for the tips!

pkg/webhook/resources/virtualmachineimage/mutator.go
Vicente-Cheng: Thanks for working on it!

Vicente-Cheng (approved):
LGTM, thanks for it!

Issue #2101, opened 2022-03-31
[BUG] test TestPlanHandler_OnChanged sometimes fails


bk201:
**Describe the bug**

The unit tests sometimes fail with:

```
--- FAIL: TestPlanHandler_OnChanged (0.01s)
    plan_controller_test.go:120: 
        	Error Trace:	plan_controller_test.go:120
        	Error:      	Not equal: 
        	            	expected: upgrade.output{plan:(*v1.Plan)(nil), upgrade:(*v1beta1.Upgrade)(0xc000416b60), err:error(nil)}
        	            	actual  : upgrade.output{plan:(*v1.Plan)(nil), upgrade:(*v1beta1.Upgrade)(0xc0000b5860), err:error(nil)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -68,3 +68,3 @@
        	            	      Status: (v1.ConditionStatus) (len=4) "True",
        	            	-     LastUpdateTime: (string) (len=20) "2022-03-31T07:00:51Z",
        	            	+     LastUpdateTime: (string) (len=20) "2022-03-31T07:00:52Z",
        	            	      LastTransitionTime: (string) "",
        	Test:       	TestPlanHandler_OnChanged
        	Messages:   	case "set NodesPrepared condition when prepare plan completes"
time="2022-03-31T07:00:52Z" level=info msg="Creating upgrade repo image"
time="2022-03-31T07:00:52Z" level=info msg="Creating upgrade repo image"
FAIL
coverage: 26.7% of statements
FAIL	github.com/harvester/harvester/pkg/controller/master/upgrade	0.232s
```

https://drone-publish.rancher.io/harvester/harvester/574/1/2

**Expected behavior**

**Support bundle**

**Environment:**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
Sometimes the process of test takes more than 1 second, we need to clean the plan timestamp.

guangbochen:
can we also backport this to the v1.0 branch, thanks?

chrisho:
> can we also backport this to the v1.0 branch, thanks?

ok

guangbochen:
close as this is just a CI bug and no additional test is required.

PR #2362, opened 2022-06-09
fix test TestPlanHandler_OnChanged sometimes fails

chrisho:
**Problem:**
Test failure occurs when TestPlanHandler_OnChanged is executed.

**Solution:**
Fix the problem of function `emptyConditionsTime`.

**Related Issue:**
#2101 

**Test plan:**
Execute several unit tests.

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller_test.go
+++ b/pkg/controller/master/upgrade/upgrade_controller_test.go
@@ -167,8 +167,8 @@ func fakeImageExist(imageCache ctlharvesterv1.VirtualMachineImageCache, displayN
 }
 
 func emptyConditionsTime(conditions []harvesterv1.Condition) {
-	for _, c := range conditions {
-		c.LastTransitionTime = ""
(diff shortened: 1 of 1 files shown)


bk201 (approved):
Thanks for fixing this!

pkg/controller/master/upgrade/plan_controller_test.go
```diff
@@ -117,6 +117,11 @@ func TestPlanHandler_OnChanged(t *testing.T) {
 			sanitizeStatus(&actual.plan.Status)
 		}
 
+		if tc.expected.upgrade != nil {
+			emptyConditionsTime(tc.expected.upgrade.Status.Conditions)
```
futuretea: seem has same logic in https://github.com/harvester/harvester/blob/a242948beeac83eab142bb5ee073aa53f8a1f2dc/pkg/controller/master/upgrade/plan_controller_test.go#L109

futuretea (approved):
LGTM

Issue #2332, opened 2022-05-31
[Backport v1.0] Harvester fails to pull Rancher agent image from private registry


FrankYang0529:
This is a backport issue for https://github.com/harvester/harvester/issues/2175.

TachunLin:
Verified **fixed** on the backport version `v1.0-9e056ebc-head`. Close this issue.
For details verification steps please refer to #2175 

### Result
PASS the following cases in https://github.com/harvester/tests/issues/347  

### Case 1: Import Harvester to Rancher

- [x]  **PASS** Import harvester cluster to the rancher and doesn't have any error.
![image](https://user-images.githubusercontent.com/29251855/173025751-1e74e273-c214-4a47-a0f2-3528bf26fb80.png)

- [x] **PASS** Check image in cattle-system/cattle-cluster-agent deployment is myregistry.local:5000/rancher/rancher-agent:v2.6.5.
![image](https://user-images.githubusercontent.com/29251855/173051739-0563ca28-d0d1-4681-bf3a-5216cc932d6c.png)


### Case 2: Reboot Harvester
- [x] Reboot Harvester node and check we still can query https://myregistry.local:5000
  ```
  node1-backport-v10:~ # curl https://myregistry.local:5000
  node1-backport-v10:~ #
  ```

### Case 3: Add another node to Harvester cluster
- [x] **PASS** Login to the new node and run the following command. It should not have error.
```
node2-backport-v10:~ # curl https://myregistry.local:5000
node2-backport-v10:~ #

```
### Case 4: Remove additional-ca
- [x] **PASS** Login to any harvester node. It should have SSL certificate problem with the following command.

```
node1-backport-v10:~ # curl https://myregistry.local:5000
curl: (60) SSL certificate problem: self signed certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

PR #2364, opened 2022-06-09
fix HARVESTER_INSTALLER_VERSION

FrankYang0529:
**Problem:**
Harvester v1.0 branch keeps using harvester-installer v1.0.2 to build ISO.

**Solution:**
Use v1.0 branch.

**Related Issue:**
https://github.com/harvester/harvester/issues/2332

### DIFF
--- a/scripts/build-iso
+++ b/scripts/build-iso
@@ -7,7 +7,7 @@ cd $(dirname $0)/..
 
 echo "Start building ISO image"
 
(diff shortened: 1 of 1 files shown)


bk201 (approved):
LGTM. 

We can think about how to improve branch management.
Otherwise this need to be modified again when cutting v1.0.3.

guangbochen (approved):

PR #2457, opened 2022-07-06
Added topology zone awareness to harvester control plane (refactored)

janeczku:
Replacement PR for https://github.com/harvester/harvester/pull/2339 with refactored node promotion


janeczku:
Apparently, the new model changes the expected behavior of promoting at most 1 node per availability zone (at least that's how the tests are defined).

janeczku:
Updated the logic to skip promotion until there are nodes from at least 3 zones available (or no zones are used at all).

janeczku:
This can be reviewed @futuretea @guangbochen

janeczku:
Not needed anymore

Issue #2310, opened 2022-05-24
[BUG] Cluster member and cluster owner see inconsistent statistics


WuJun2016:
**Describe the bug**

cluster owner:
![image](https://user-images.githubusercontent.com/24985926/169994337-00ed7ab8-4acf-41bd-8930-2d32347efb48.png)
![image](https://user-images.githubusercontent.com/24985926/169994374-c49e8999-5cda-4833-89de-9df498023897.png)

cluster member:
<img width="1674" alt="image" src="https://user-images.githubusercontent.com/24985926/169994497-4f3efca5-f067-496f-87a4-e55bc4f28cb3.png">
<img width="1679" alt="image" src="https://user-images.githubusercontent.com/24985926/169994541-aad53650-c713-403c-aeb1-876d22fedb6d.png">

**To Reproduce**
1. Create two users and assign cluster member and cluster owner privileges respectively
2. go to harvester, View CPU, Memory, Storage statistics

**Expected behavior**

expect to see consistent statistics

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

n313893254:
Notice that the rancher cluster shows statistics using `status.requested` of the crd `management.cattle.io.cluster`. Which API endpoint is `/v1/management.cattle.io.clusters/[CLUSTER_ID]`

But for the Harvester cluster, the `status.requested` is zero, Cloud we return the CPU and memory data as the Rancher cluster like?

![image.png](https://images.zenhubusercontent.com/59db4297b0222d5de47ad632/0f9600e4-6ffe-440c-91ff-5c624ee71790)

![image.png](https://images.zenhubusercontent.com/59db4297b0222d5de47ad632/31671732-3265-4c9d-b6c6-00965d35ed34)

TachunLin:
Verified **not fixed** on `master-8f01b558-head` with Rancher `v2.6.6`. Move this issue to implement stage.

ui-index: https://releases.rancher.com/harvester-ui/dashboard/release-harvester-v1.0/index.html
version: index.8caabe861906430877bf.js

### Result
The `cluster_member` rule can't see the correct resource usage statistics on host and dashboard page
Most of the statistics are empty. 

* Dashboard page of cluster_owner
![image](https://user-images.githubusercontent.com/29251855/179163953-bcde4607-ca0a-4a01-aed1-38c47a94b710.png)

[...]

PR #2483, opened 2022-07-12
Add Pod Resources OnChanged to rancher controller

futuretea:

**Problem:**

After Rancher imports the other clusters, in rancher cluster agent, [podresource controller](https://github.com/rancher/rancher/tree/release/v2.6/pkg/controllers/managementagent/podresources) will calculating pod request and store the result on node's annotation.

Since rancher service is exist in the Harvester cluster, rancher cluster agent will not start rancher in cluster agent
(https://github.com/rancher/rancher/blob/8ce1e0de80abe096299020e03c49ed7e159696f9/pkg/agent/rancher/rancher.go#L89)

And `multi-cluster-management-agent` is [disabled](https://github.com/harvester/harvester-installer/blob/9b8e08df839c34b649225a95d6217bc0fb86d79f/package/harvester-os/files/usr/share/rancher/rancherd/config.yaml.d/50-defaults.yaml#L6) in the Harvester embedded Rancher

So podresource controller won't run in any component

Currently, for Harvester cluster, the pod resources are calculated by the Harvester UI, but standard users do not have access to all pods, so they cannot get correct pod requests value.This calculation should be done by the backend

**Solution:**

copy Rancher [podresource controller](https://github.com/rancher/rancher/tree/release/v2.6/pkg/controllers/managementagent/podresources) logic to Harvester

**Related Issue:**
https://github.com/harvester/harvester/issues/2310

**Test plan:**

1. change harvester image to futuretea/harvester:podresources-head
```bash
kubectl -n harevster-system edit deploy harvester
```

2. check if the annotation `management.cattle.io/pod-limits` and `management.cattle.io/pod-requests` is exist in Harvester node annotation
```bash
kubectl get no -oyaml
```

### DIFF
--- a/pkg/controller/master/rancher/pod_resources.go
+++ b/pkg/controller/master/rancher/pod_resources.go
@@ -1,0 +1,132 @@
+package rancher
+
+import (
+	"encoding/json"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// Since multi-cluster-management-agent is disabled in the Harvester embedded Rancher,
+// copy and modified this file from Rancher to fix https://github.com/harvester/harvester/issues/2310
+// source file: https://github.com/rancher/rancher/tree/release/v2.6/pkg/controllers/managementagent/podresources
+
+const (
+	RequestsAnnotation = "management.cattle.io/pod-requests"
+	LimitsAnnotation   = "management.cattle.io/pod-limits"
+)
+
+func (h *Handler) PodResourcesOnChanged(key string, node *corev1.Node) (*corev1.Node, error) {
+	if node == nil {
+		return node, nil
+	}
+
+	h.nodeController.EnqueueAfter(node.Name, 15*time.Second)
+
+	pods, err := h.getNonTerminatedPods(node)
+	if err != nil {
+		return nil, err
+	}
+
+	requests, limits, err := getPodResourceAnnotations(pods)
+	if err != nil {
+		return nil, err
+	}
+
+	if node.Annotations[RequestsAnnotation] != requests ||
+		node.Annotations[LimitsAnnotation] != limits {
+		toUpdate := node.DeepCopy()
+		if toUpdate.Annotations == nil {
+			toUpdate.Annotations = map[string]string{}
+		}
+		toUpdate.Annotations[RequestsAnnotation] = requests
+		toUpdate.Annotations[LimitsAnnotation] = limits
+		return h.nodeController.Update(toUpdate)
+	}
+
+	return node, nil
+}
+
+func (h *Handler) getNonTerminatedPods(node *corev1.Node) ([]*corev1.Pod, error) {
+	var pods []*corev1.Pod
+	fromCache, err := h.podCache.List("", labels.NewSelector())
+	if err != nil {
+		return pods, err
+	}
+
+	for _, pod := range fromCache {
+		if pod.Spec.NodeName != node.Name {
+			continue
+		}
+		// kubectl uses this cache to filter out the pods
+		if pod.Status.Phase == "Succeeded" || pod.Status.Phase == "Failed" {
+			continue
+		}
+		pods = append(pods, pod)
+	}
+	return pods, nil
+}
+
+func getPodResourceAnnotations(pods []*corev1.Pod) (string, string, error) {
+	requests, limits := aggregateRequestAndLimitsForNode(pods)
+	requestsBytes, err := json.Marshal(requests)
+	if err != nil {
+		return "", "", err
+	}
+
+	limitsBytes, err := json.Marshal(limits)
+	return string(requestsBytes), string(limitsBytes), err
+}
+
+func aggregateRequestAndLimitsForNode(pods []*corev1.Pod) (map[corev1.ResourceName]resource.Quantity, map[corev1.ResourceName]resource.Quantity) {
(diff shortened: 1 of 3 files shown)


pkg/controller/master/rancher/pod_resources.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return pods, err
+	}
+
+	for _, pod := range fromCache {
+		if pod.Spec.NodeName != node.Name {
+			continue
+		}
+		// kubectl uses this cache to filter out the pods
+		if pod.Status.Phase == "Succeeded" || pod.Status.Phase == "Failed" {
```
chrisho, line 63: It should be better to use k8s constants here
if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == PodFailed

futuretea: just keep same code with rancher for future sync

chrisho (approved):
LGTM

pkg/controller/master/rancher/pod_resources.go
```diff
@@ -0,0 +1,132 @@
...
+
+	requests, limits, err := getPodResourceAnnotations(pods)
+	if err != nil {
+		return nil, err
+	}
+
+	if node.Annotations[RequestsAnnotation] != requests ||
+		node.Annotations[LimitsAnnotation] != limits {
+		toUpdate := node.DeepCopy()
+		if toUpdate.Annotations == nil {
```
w13915984028, line 42: If really worry about `toUpdate.Annotations == nil `, then it should be before L38,  as `node.Annotations[RequestsAnnotation]` may cause panic when it's nil

futuretea: read nil map won't panic

futuretea: A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic
https://go.dev/blog/maps

pkg/controller/master/rancher/pod_resources.go
```diff
@@ -0,0 +1,132 @@
...
+		toUpdate.Annotations[LimitsAnnotation] = limits
+		return h.nodeController.Update(toUpdate)
+	}
+
+	return node, nil
+}
+
+func (h *Handler) getNonTerminatedPods(node *corev1.Node) ([]*corev1.Pod, error) {
+	var pods []*corev1.Pod
+	fromCache, err := h.podCache.List("", labels.NewSelector())
```
w13915984028: Could here directly conditional `List` those pods which meet  `pod.Spec.NodeName == node.Name` ?  dramatically  reduce the return `fromCache` size.

futuretea: The List method cannot do this. We can do this by indexer

futuretea: Updated in a new commit https://github.com/harvester/harvester/pull/2483/commits/aec64208131cbbd35b8db111d48c92e2af979304

pkg/controller/master/rancher/pod_resources.go
```diff
@@ -0,0 +1,132 @@
...
+
+func addMapForInit(data1 map[corev1.ResourceName]resource.Quantity, data2 map[corev1.ResourceName]resource.Quantity) {
+	for name, quantity := range data1 {
+		value, ok := data2[name]
+		if !ok {
+			data2[name] = quantity.DeepCopy()
+			continue
+		}
+		if quantity.Cmp(value) > 0 {
+			data2[name] = quantity.DeepCopy()
```
w13915984028, line 128: Question: 
Why `addMapForInit` is different with `addMap` in processing the `quantity`?

futuretea: not sure what is the reason, this file is just a copy for https://github.com/rancher/rancher/blob/release/v2.6/pkg/controllers/managementagent/podresources/register.go

In my opinion, the reason is `Each init container must complete successfully before the next one starts.`
more info: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/

pkg/controller/master/rancher/pod_resources.go
w13915984028: A bit surprised that the `Rancher` code is simply retriving all PODs. Given 100 NODEs * 30 PODs, it will retrieve and loop 3000 PODs every 15 seconds.

```
rancher: pkg/controllers/managementagent/podresources/register.go
func (p *podResources) getNonTerminatedPods(node *corev1.Node) ([]*corev1.Pod, error) {
	var pods []*corev1.Pod
	fromCache, err := p.podLister.List("", labels.NewSelector())
..
	for _, pod := range fromCache {
		if pod.Spec.NodeName != node.Name {
			continue
		}	
```

guangbochen (approved):
LGTM

pkg/controller/master/rancher/pod_resources.go
w13915984028: It may not be fully correct here. 

As `container` state is not checked, the last `init container` with  resource requests/limits value will overwrite others value, no matter which `init container` is trully occupying resources.

Normally, when the `initContainers` finished, they will release the resources, their related values seems should be 0 then.

futuretea: The calculation here should be the minimum resources needed to make sure the pod is running/restart properly

futuretea: max init container requests + all container requests

futuretea: Totally, the result here is `request`, not `used`

w13915984028 (approved):
LGTM, thanks.

Issue #2336, opened 2022-05-31
[BUG] coredns autoscaler keeps being oom-killed


alexdepalex:
**Describe the bug**
`rke2-coredns-rke2-coredns-autoscaler` keeps running out of its memory limits and therefore gets killed.

**To Reproduce**
1. Install a Harvester cluster.
2. Wait.

**Expected behavior**
Autoscaler shouldn't get killed.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.0.1
 - Underlying Infrastructure: R740xd (2x 20 core Intel(R) Xeon(R) Gold 5218R CPU @ 2.10GHz, 768GB)
 
**Additional context**
Memory limit for the autoscaler process is set to 20Mi, but regularly spikes above that. Since 10d uptime, the autoscaler was restarted 300 times.

![coredns-autoscaler-memory-usage](https://user-images.githubusercontent.com/713019/171137907-75143d3f-20c0-4de6-8549-9d02ca887e9f.png)

alexdepalex:
Still need to add some info.

guangbochen:
Would be helpful if u can share the support bundle to harvester-support-bundle@suse.com with the correct issue ID with more detailed system & container logs, thanks.

alexdepalex:
@guangbochen Bundle has been uploaded through scc.

guangbochen:
Related to https://github.com/rancher/rke2/issues/2175, as a temp workaround, we can patch the default memory limit to 50MiB.

irishgordo:
## Testing Setup:
- ran `make` via `harvester-installer` on `commit 01d120018139c4cf58aba50132b5c10b7f1b92bb (HEAD, tag: v1.0.3-rc1, upstream/v1.0)`, manually building iso, which held the commits for PR (Bump rke2 to v1.22.12-rc3+rke2r1, and others)
- tested via QEMU/KVM, 24354MiB, 8CPU Box, 150GB Disk, with built .iso

## Results:
- Wasn't able to notice in 30min+ of running any OOM (out of memory) errors with `rke2-coredns-rke2-coredns-autoscaler-UUID` pod.
Cross-Reference:
![memory-utilization-rke2-coredns-rke2-coredns-autoscaler](https://user-images.githubusercontent.com/5370752/180505912-c265cd03-2e09-4381-a862-3fa01b8ee375.png)
- Autoscaler seem to run appropriately:
[...]

PR #2493, opened 2022-07-13
overwrite labels in promote script

FrankYang0529:
**Problem:**
Promotion jobs may fail if there is label on the machine.

**Solution:**
Add `--overwrite`.

**Related Issue:**
https://github.com/harvester/harvester/issues/2336


guangbochen (approved):

bk201 (approved):
LGTM

PR #2504, opened 2022-07-15
Make sure there are enough zones in node promotion

futuretea:

**Problem:**

2 etcd is unstable
we should only promote nodes if there has enough zones

**Solution:**

1. Add unit tests to make sure only promote nodes if there has enough zones
2. Optimize code to improve performance and make it easier to read

**Related Issue:**
https://github.com/harvester/harvester/issues/2325

**Test plan:**

1. Install first node, the role of this node should be `Management Node`
2. Install second node, the role of this node should be `Compute Node`, the second node shouldn't be promoted to `Management Node`
3. Add label `topology.kubernetes.io/zone=zone1` to the first node
4. Install third node, the second node and third node shouldn't be promoted
5. Add label `topology.kubernetes.io/zone=zone1` to the second node, the second node and third node shouldn't be promoted 
6. Add label `topology.kubernetes.io/zone=zone3` to the third node, the second node and third node shouldn't be promoted
7. Change the value of label `topology.kubernetes.io/zone` from `zone1` to `zone2` in the second node, the second node and third node will be promoted to `Management Node` one by one

### DIFF
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -10,7 +10,6 @@ import (
 	ctlbatchv1 "github.com/rancher/wrangler/pkg/generated/controllers/batch/v1"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/name"
-	"github.com/rancher/wrangler/pkg/slice"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


chrisho (approved):
LGTM

pkg/controller/master/node/promote_controller.go
```diff
@@ -247,56 +246,77 @@ func (h *PromoteHandler) setPromoteResult(job *batchv1.Job, node *corev1.Node, s
 // If the cluster doesn't need to be promoted, return nil
 func selectPromoteNode(nodeList []*corev1.Node) *corev1.Node {
 	var (
-		managementNumber               int
-		promoteNode                    *corev1.Node
-		managementNodeTopologyZoneList []string
+		promoteNode                             *corev1.Node
+		healthyHarvesterWorkers                 []*corev1.Node
+		managementOrHealthyHarvesterWorkerZones = make(map[string]bool)
+		managementZones                         = make(map[string]bool)
```
guangbochen, line 252: May prefer to set the init size to `0`.

futuretea: the init size is `0`

guangbochen (approved):
LGTM and let's also backport this to v1.0, thanks.

Martin-Weiss:
Thanks for adding this feature! 

Do we have a documentation of this feature available?
Does this allow a configuration of two masters in zone1 and one master in zone2 or is it always required to have three zones in case this feature is used?

futuretea:
> Thanks for adding this feature!
> 
> Do we have a documentation of this feature available? Does this allow a configuration of two masters in zone1 and one master in zone2 or is it always required to have three zones in case this feature is used?

- No documentation at present
- It always required to have three zones

Martin-Weiss:
> > Thanks for adding this feature!
> > Do we have a documentation of this feature available? Does this allow a configuration of two masters in zone1 and one master in zone2 or is it always required to have three zones in case this feature is used?
> 
>     * No documentation at present

Looking forward to get this, soon ;-)

> 
>     * It always required to have three zones

Ok - so something like this should work:

server1 - zone1 - master
server2 - zone1 - worker
server3 - zone2 - master
server4 - zone2 - worker
server5 - zone3 - master
server6 - zone3 - worker

So this will result in having masters i.e. on server1+server3+server5. (or more or less random selection per zone)

And in case server5 fails for some time - will server6 be changed from worker to master automatically? Or just in case server5 is removed / uninstalled completely?

futuretea:
> And in case server5 fails for some time - will server6 be changed from worker to master automatically? Or just in case server5 is removed / uninstalled completely?

in case server5 is removed / uninstalled completely

If master node fails, admin need to handle it manually.

PR #2532, opened 2022-07-19
get rke2 version without rc

FrankYang0529:
**Problem:**
We can't upgrade to RKE2 rc version.

**Solution:**
RKE2 doesn't expect users to update from a rc version to another in the same minor version and we don't do it either. We remove `-rcX` to compare the string.

**Related Issue:**
https://github.com/harvester/harvester/issues/2336

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -228,7 +228,10 @@ EOF
 }
 
 wait_rke2_upgrade() {
-  until [ "$(get_node_rke2_version)" = "$REPO_RKE2_VERSION" ]
+  # RKE2 doesn't show '-rcX' in nodeInfo, so we remove '-rcX' in $REPO_RKE2_VERSION.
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -228,7 +228,8 @@ EOF
 }
 
 wait_rke2_upgrade() {
-  until [ "$(get_node_rke2_version)" = "$REPO_RKE2_VERSION" ]
+  REPO_RKE2_VERSION_WITHOUT_RC=$(echo -n $REPO_RKE2_VERSION | sed 's/-rc[[:digit:]]*//g')
```
bk201, line 233: It would be nice to add a comment about why we need to strip `-rc*` here.

FrankYang0529: Added it. Thank you.

bk201 (approved):
LGTM, confirm works

guangbochen (approved):
LGTM, thanks.

Issue #2429, opened 2022-06-28
Hardware Passthrough Support


mircea-pavel-anton:
**Is your feature request related to a problem? Please describe.**

I have tried migrating my Proxmox workloads to Harvester and I was faced with a lack of support for configuring hardware passthrough from the GUI. Among those workloads are things such as:

- virtualizing a TrueNAS instance -> passing through an HBA
- virtualizing a pfSense instance -> passing through multiple NICs
- virtualizing a workstation -> passing through GPUs and USB controllers

While I'd have no problem configuring these from the command line if needed, i also couldn't find proper documentation on how to implement these things.

**Describe the solution you'd like**

Ideally, after configuring the devices to use the vfio drivers, a drop down should be present allowing us to pick available devices to pass through, just like proxmox does it.

**Describe alternatives you've considered**

Alternatively, providing some documentation and examples of configuring such setups by hand would be a great first step, and much quicker than the "full" solution.

**Additional context**

N/A

fanlix:
follow #1029  #1710

mircea-pavel-anton:
> follow #1029  #1710

I am. I also replied to that thread and did some preliminary testing myself on the GPU passthrough, but i didn't manage to find anything on, for example, passing through a NIC or a HBA.

The progress on this topic will be tracked in that thread, i assume?

kingnarmer:
Hello,

I need to be able to add storage LUNs presented to harvester nodes via Fiber Channel Qlogic HBA card. I can see the LUNs on OS level using "cat /proc/partitions", however I don't see them when I try to add disks. 

Appreciate help getting them added. 

Thanks

tlehman:
This feature is now present as an addon in [release 1.1.0](https://github.com/harvester/harvester/releases/tag/v1.1.0). I have a video walkthrough of how to use the feature [here](https://tobilehman.com/posts/suse-harvester-pci/).

mircea-pavel-anton:
@tlehman thanks for that! I will take a look tomorrow and set it up in my homelab.

AndreiLazarescu:
Hello,

Does it support Qlogic NetXtreame 10Gb SPF NIC?

Kind regards,
Andrei

Issue #1029, opened 2021-07-19
[FEATURE] Kubevirt GPU/vGPU Configuration


kralicky:
Kubevirt has support for provisioning GPUs and other host devices according to the [docs](https://kubevirt.io/user-guide/virtual_machines/host-devices/)

There is also an [existing Nvidia plugin](https://github.com/NVIDIA/kubevirt-gpu-device-plugin) which will discover GPUs and vGPUs on nodes so they can be added to VMs.

It would be great if the Harvester UI could display available GPUs and add the necessary configuration into VirtualMachineInstances.

I've done some experimenting with assigning GPUs to Harvester-provisioned VMs which I documented in this [gist](https://gist.github.com/kralicky/0f9994526eac7ddc1808bcbfea6a8444)

Edit: Include GPU support for VMs and K8s workloads.

rebeccazzzz:
Added a line to the description above: “Include GPU support for VMs and K8s workloads.”

ProjectInitiative:
Adding "hookscripts" that allow a script to run during different phases of the VM lifecycle would be beneficial when dealing with binding, unbinding and cleaning up the VFIO-PCI drivers.

WuJun2016:
Hi,  @tlehman  should this function include UI related changes?  If so, please help to add `require/ui` label.  thanks

tlehman:
The version of this feature we are targeting for release 1.1.0 will be based on #992 and use PCI Passthrough. The driver for your specific GPU will need to be installed in a VM, but the VM will have direct access to the underlying PCI device by using PCI passthrough.

tlehman:
Closing this issue as a duplicate of #992

tlehman:
The vGPU work is being tracked in #2764

Issue #992, opened 2021-07-02
[FEATURE] PCI-E passthrough (and GPU support)


A-vesalius:
**Is your feature request related to a problem? Please describe.**
VM cannot utilize the full network capability of baremetal hardware, especially 10gbe and above as well as DPDK with harvester.

**Describe the solution you'd like**
Leverage Multus which has the ability to use pci-e passthrough, IOMMU and VT-d, devices and SR-IOV VFs for kubernetes networks. Once these VF's aresetup on baremetal then harvestor GUI should allow the user to hand them out to VMs in an otherwise transparent way equivalent to physical nics. 

**Describe alternatives you've considered**
https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin#quick-start

https://docs.mellanox.com/pages/releaseview.action?pageId=25145021

**Additional context**

tlehman:
Just to clarify what SR-IOV gives you, I want to make a simple diagram that motivates why this is such a high priority:

When a packet arrives at the NIC, there's one core that is interrupted and handles that interrupt. Then the destination VM's core is interrupted and stops, and the hypervisor copies the packet to the VM's memory space, then the VM core continues, and so does the core that handles the NIC's interrupts.

## Without SR-IOV:

```mermaid
sequenceDiagram
Packet ->> NIC: 2 packets arrive
NIC ->> Hypervisor: copy packet 1
Hypervisor ->> VM3: copy packet 1
VM3 ->> VM3: continues
NIC ->> Hypervisor: copy packet 2
Hypervisor ->> VM6: copy packet 2
VM6 ->> VM6: continues
````

## With SR-IOV
```mermaid
sequenceDiagram
Packet ->> NIC: 2 packets arrive
NIC ->> VF3: queued
NIC ->> VF6: queued
VF3 ->> VM3: DMA to VM
VF6 ->> VM6: DMA to VM
```

(DMA is short for Direct Memory Access)

In summary, SR-IOV cuts the hypervisor out of the pathway from the NIC to the VM, dramatically increasing the throughput in your virtualized environment.

w13915984028:
connect with https://github.com/harvester/harvester/issues/2236
SRIOV may also be represented as a kind of network to reuse existing features

gusraggio:
Rancher is using this chart to configure this, https://github.com/rancher/charts/pull/1211#pullrequestreview-668492770

tlehman:
[...]

PR #2555, opened 2022-07-22
[Draft] HEPs (SR-IOV, PCI-PT, GPU) [CI SKIP]

tlehman:
**Problem:**
The general problem this branch is aiming to solve is enabling Harvester users to make full use of their devices. Specifically, network interfaces with SR-IOV features, and GPUs.

**Solution:**
The solutions are sketched out in these HEP documents. This is a draft, but I would like feedback early and often as I get these finished up.

**NOTE:**
If you want the rich text with the diagrams, you can click the rich diff button or go directly to these links:
- https://github.com/tlehman/harvester/blob/sriov-hep/enhancements/20220614-single-root-io-virtualization.md
- https://github.com/tlehman/harvester/blob/sriov-hep/enhancements/20220722-gpu-support.md
- https://github.com/tlehman/harvester/blob/sriov-hep/enhancements/20220722-pci-passthrough.md

**Related Issue:**
- https://github.com/harvester/harvester/issues/992
- https://github.com/harvester/harvester/issues/1029
- https://github.com/harvester/harvester/issues/2429

**Test plan:**
CI Disabled for this PR, since it's just static markdown documents.


enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+        VM1 --copy packets--> h([Hypervisor]);
+        VM2 --copy packets--> h([Hypervisor]);
+        h --copy packets--> PF;
+        subgraph driver
+            PF;
+        end
+    end
+```
+
+With SR-IOV enabled, the driver exposes multiple VFs, each of which can be assigned to be exclusively owned by Pods or VMs. Further, SR-IOV gives the VM direct memory access (DMA) to the PF's driver memory.
```
guangbochen:
```suggestion
With SR-IOV enabled, the driver exposes multiple VFs, each of which can be assigned to be exclusively owned by Pods or VMs. Further, SR-IOV gives the VM direct memory access (DMA) to the PF's driver memory for better network performance.
```

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Related Issue(s)
+
+ - https://github.com/harvester/harvester/issues/992
+
+## Motivation
+
+### Goals
+
+- Empower our customers to make full use of their SR-IOV hardware, speeding up their virtualized workloads
```
guangbochen, line 64:
```suggestion
- Empower Harvester users to make full use of their SR-IOV hardware, speeding up network performance of their virtualized workloads.
```

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+  - For example, [this user](https://github.com/harvester/harvester/issues/993#issue-935351354) who wrote the issue can't get 10Gbps throughput. After we ship SR-IOV, if they can get 10Gbps in their VMs, then this enhancement will be considered a success.
+
+### Non-goals [optional]
+
+- GPU Support (there is a separate [HEP](20220722-gpu-support.md) covering that)
```
guangbochen, line 69: Wondering why GPU is related to this topic since SR-IOV should be a network-related enhancement?

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+### Related Issue(s)
+
+ - https://github.com/harvester/harvester/issues/992
+
+## Motivation
+
+### Goals
+
+- Empower our customers to make full use of their SR-IOV hardware, speeding up their virtualized workloads
+  - For example, [this user](https://github.com/harvester/harvester/issues/993#issue-935351354) who wrote the issue can't get 10Gbps throughput. After we ship SR-IOV, if they can get 10Gbps in their VMs, then this enhancement will be considered a success.
```
guangbochen: Seems the issue link is not correct.

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+```json
+{
+  "sriov_enabled": false,
+  "num_vfs": 0,
+}
+```
+
+After changing `sriov-config`, in the background, Harvester unloads the driver for the sriov-capable PCI devices, and then reloads 
+the driver with `max_vfs = $sriov-config.num_vfs`. It then installs the `sriov-network-operator` and `cert-manager` to make 
+the SR-IOV network operator's admission webhooks work.
```
guangbochen, line 171: This is more related to the implementation details, but I found the Rancher's [sriov chart](https://github.com/rancher/charts/tree/dev-v2.6/charts/sriov/100.0.3%2Bup0.1.0) is literally identical with the `sriov-network-operator` except we don't need to deal with the cert-manager and sdk-operator stuff, so I would prefer to using the Rancher's chart + helm install approach.

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+
+After the installation is complete, it then creates an SriovNetworkNodePolicy CR:
+
+```yaml
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetworkNodePolicy
+metadata:
+  name: sriov-policy-202207221730
+  namespace: openshift-sriov-network-operator
```
guangbochen: need to define a proper namespace for the sriov deployment. e.g., `harvester-sriov-network-operator`

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+After creating the `SriovNetwork` object, the operator will create a corresponding `NetworkAttachmentDefinition` with the same name.
+This `NetworkAttachmentDefinition` will be visible from the Advanced Networks UI. When Alice attaches a VM to the `SriovNetwork`, 
+the operator's network resources injector will add the `resource` field to the first container of the pod. This will give the VM access to a VF.
+
+#### Story 2 in detail
+[Story 2 summary](#story-2). 
+
+TODO
+
+### API changes
```
guangbochen, line 220: The SR-IOV network operator introduces the following new CRDs, and u may need to explain a little bit more of its role and expected users, e.g.,
 - SriovNetwork (all users) - what is our expected ipam type and does it support VLAN network? 
- SriovNetworkNodePolicy (cluster-admin) - Do we consider support adding multiple `SriovNetworkNodePolicy`, since a Harvester cluster might contain different SR-IOV pci devices and it will be installed in a different region/zone of nodes. 
- SriovNetworkNodeState (operator-managed) - should we consider showing those sriov NIC and vfs info on the Harvester node page?

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+  "sriov_enabled": false,
+  "num_vfs": 0,
+}
+```
+
+After changing `sriov-config`, in the background, Harvester unloads the driver for the sriov-capable PCI devices, and then reloads 
+the driver with `max_vfs = $sriov-config.num_vfs`. It then installs the `sriov-network-operator` and `cert-manager` to make 
+the SR-IOV network operator's admission webhooks work.
+
+After the installation is complete, it then creates an SriovNetworkNodePolicy CR:
```
guangbochen, line 173: Seems the `SriovNetworkNodePolicy` custom resource should be managed by the cluster admin only, and the above `sriov-config` setting may only be responsible for installing the `sriov-operator` chart when enabled and adding some default configurations.

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+  numVfs: 8
+  nicSelector:
+    pfNames: ["eth1v#1-8"]
+  deviceType: vfio-pci
+```
+
+Harvester checks the ``SriovNetworkNodeState`` object (automatically built by the operator) to populate the `nicSelector`.
+
+The choice of `pfNames` with a range of NIC names assumes that all the machines in the cluster have the same hardware.
+This sets up a pool of VF resources that can be allocated to VM pods automatically by the `sriov-network-operator`'s CNI plugin.
```
guangbochen, line 196: We probably can't assume that all the machines in the cluster will have the same sriov pic devices, it should be configurable by the admin users. (e.g. using the SriovNetworkNodePolicy CR, please refer to the #API Changes comment, thanks.)

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+The `SriovNetwork` is used to assign VFs to VMs.
+
+To create an `SriovNetwork`, Alice goes to the Advanced menu:
+
+![advanced sriov](20220614-single-root-io-virtualization/advanced-sriov-networks.png) 
+
+After creating the `SriovNetwork` object, the operator will create a corresponding `NetworkAttachmentDefinition` with the same name.
+This `NetworkAttachmentDefinition` will be visible from the Advanced Networks UI. When Alice attaches a VM to the `SriovNetwork`, 
+the operator's network resources injector will add the `resource` field to the first container of the pod. This will give the VM access to a VF.
+
```
guangbochen, line 214: Will also need to add a VM usage example like the https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#start-an-sr-iov-vm

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+Alice now has to create an `SriovNetwork`. 
+The `SriovNetwork` is used to assign VFs to VMs.
+
+To create an `SriovNetwork`, Alice goes to the Advanced menu:
+
+![advanced sriov](20220614-single-root-io-virtualization/advanced-sriov-networks.png) 
+
+After creating the `SriovNetwork` object, the operator will create a corresponding `NetworkAttachmentDefinition` with the same name.
+This `NetworkAttachmentDefinition` will be visible from the Advanced Networks UI. When Alice attaches a VM to the `SriovNetwork`, 
+the operator's network resources injector will add the `resource` field to the first container of the pod. This will give the VM access to a VF.
```
guangbochen: this should go to the implementation details instead of user stories as it will be invisible to the users.
Additionally, we should at least add an example for both admin users and standard users (e.g., with project or namespace permissions only).

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+3: eth0v3 <BROADCAST,MULTICAST...
+4: eth0v4 <BROADCAST,MULTICAST...
+```
+
+Then four different virtual machines could each use one of the above VFs. 
+SR-IOV allows the VM to get direct memory access to the PF. 
+This way the VM can copy data directly to the device without going through the hypervisor.
+
+## Summary
+
```
guangbochen, line 22: It might be helpful to add some pre-requisite here e.g.,

## Prerequisites
- You will need to make sure your host has an SR-IOV capable NIC plugged in, and that it supports it. You may need to adjust BIOS settings to make it work. For example, make sure VT-x (hardware virtualization) and Intel VT for Directed I/O are enabled in BIOS.
- Supported Hardware - https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/doc/supported-hardware.md

enhancements/20220614-single-root-io-virtualization.md
```diff
@@ -0,0 +1,229 @@
...
+### API changes
+
+## Design
+
+### Implementation overview
+
+##### Changes to [Harvester installer](https://github.com/harvester/harvester-installer) 
+- During installation, use `lspci -s PCI_ADDRESS -v | grep SR-IOV` to check if the node has any SR-IOV capable hardware.
+   - If so, then label the node with `feature.node.kubernetes.io/network-sriov.capable=true`, this is the default `nodeSelector`, 
+     it allows the operator to find the nodes and start building out the `SriovNetworkNodeState`
```
guangbochen, line 229: 1. I don't think u would be able to know the `PCI_ADDRESS` during the installation
2. Probably check it when the sriov chart is enabled and use a [custom controller](https://github.com/harvester/harvester/tree/master/pkg/controller/master/node) to sync the node labels.  
3. On the harvester-installer u may want to add the sriov as an embedded chart like the [monitoring chart](https://github.com/harvester/harvester-installer/blob/master/pkg/config/templates/rancherd-13-monitoring.yaml).
     - https://github.com/harvester/harvester-installer/blob/d37e60721c6b9ead4f819edda1f5b973771032c2/scripts/build-bundle#L56-L58

enhancements/20220614-single-root-io-virtualization.md
tlehman: GPU is dependent on PCI-PT, and SR-IOV in guest cluster mode is also dependent on PCI-PT, so I wanted to show that even though they both depend on PCI-PT, this HEP doesn't depend on GPU

enhancements/20220614-single-root-io-virtualization.md
tlehman: Oh, accidental Ctrl-A incremented the issue number, should be 992

enhancements/20220614-single-root-io-virtualization.md
tlehman: It still has the admission webhooks, and the webhooks need TLS certs to work, how does Rancher get around that problem without something like cert-manager?

enhancements/20220614-single-root-io-virtualization.md
tlehman: ctrl-a incremented 992 to 993, will fix

enhancements/20220614-single-root-io-virtualization.md
tjjh89017: I think we should also add a limitation.
We cannot use mgmt interface to be SR-IOV pool NIC.

enhancements/20220722-pci-passthrough.md
```diff
@@ -0,0 +1,219 @@
...
+    nodes: [
+        {name: "node1", pciDevicesToPassThrough: [1]}
+    ]
+}
+```
+
+After that setting is updated, the settings controller will loop through the devices and prepare them 
+for PCI Passthrough:
+
+The preparation for passthrough consists of unbinding the driver, doing a driver override, then binding the
```
bk201, line 126: I'm not quite sure why we need a driver in the host if it's eventually unbound. By searching tutorials on Internet, most of them are just binding the device to vfio-pci.

tlehman:
Closing in favor of https://github.com/harvester/harvester/pull/3189

Issue #2559, opened 2022-07-25
[BUG] Backup unable to be restored and the VM can't be deleted


lanfon72:
### Describe the bug
VM hanging on `restoring` state and can not be removed.

### To Reproduce
1. Install Harvester with any nodes
2. Login to Dashboard then navigate to _Advanced/Settings_, setup `backup-target` with NFS or S3
3. Create Image for VM creation
4. Create VM `vm1`
5. Take backup from `vm1` as `vm1b`
6. Take backup from `vm1` as `vm1b2`
7. Click **Edit YAML** of `vm1b`, update field `status.source.spec.spec.domain.cpu.cores`, increase `1`
8. Stop VM `vm1`
9. Restore backup `vm1b2` with **Replace Existing**
10. Restore backup `vm1b` with **Replace Existing** when the VM `vm1` still in state `restoring`

### Expected behavior
VM should be restored successfully.

## Environment:
 - Harvester ISO version: **v1.0.3-rc1**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **qemu/KVM 3 nodes**

### Additional context

https://user-images.githubusercontent.com/5169694/180805112-2d6f3c5e-1f4e-486d-9441-010ecf8e8f24.mp4

lanfon72:
update to `reproduce/often` because I can delete the VM sometimes.

irishgordo:
While the `vm1` is in the 'restoring' state from replace-existing of backup `vm1b2` - then trying to immediately 'replace existing' with `vm1b` I'm noticing that there is an error of:
```
admission webhook "validator.harvesterhci.io" denied the request. Please wait for the previous VM restore on the default/vm1 to be complete first.
```
![must-wait](https://user-images.githubusercontent.com/5370752/182722180-3e2f606b-beef-4f8b-8f33-8d235587db4b.png)

But then if you wait just a few seconds it does go through 
![right-after-works](https://user-images.githubusercontent.com/5370752/182723481-c9eceaf8-e817-4c2e-b7a4-cf50c72507e3.png)

@lanfon72 was that error something you were seeing if you tried to restore while it was restoring?

I built the harvester iso by checking out harvester-installer at branch `v1.0`, then ran `make` then used that in a qemu/kvm vm with Harvester version: v1.0-60921953-head

FrankYang0529:
[...]

PR #2568, opened 2022-07-26
fix(webhook/vmrestore): check on the same target

FrankYang0529:
**Problem:**

**Solution:**
Add webhook to prevent it.

**Related Issue:**

**Test plan:**
4. Create VM `vm1`
7. Stop VM `vm1`

### Additional context
Since v1.0.3 released date is close, I create the PR for `v1.0` first. I will cherry-pick it to `master` branch.

### DIFF
--- a/pkg/webhook/resources/restore/validator.go
+++ b/pkg/webhook/resources/restore/validator.go
@@ -15,6 +15,7 @@ import (
 	"github.com/harvester/harvester/pkg/settings"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
+	"github.com/harvester/harvester/pkg/webhook/util"
 )
 
 const (
@@ -27,20 +28,23 @@ func NewValidator(
 	vms ctlkubevirtv1.VirtualMachineCache,
 	setting ctlharvesterv1.SettingCache,
 	vmBackup ctlharvesterv1.VirtualMachineBackupCache,
+	vmRestore ctlharvesterv1.VirtualMachineRestoreCache,
 ) types.Validator {
 	return &restoreValidator{
-		vms:      vms,
-		setting:  setting,
-		vmBackup: vmBackup,
+		vms:       vms,
+		setting:   setting,
+		vmBackup:  vmBackup,
+		vmRestore: vmRestore,
 	}
 }
 
 type restoreValidator struct {
 	types.DefaultValidator
 
-	vms      ctlkubevirtv1.VirtualMachineCache
-	setting  ctlharvesterv1.SettingCache
-	vmBackup ctlharvesterv1.VirtualMachineBackupCache
+	vms       ctlkubevirtv1.VirtualMachineCache
+	setting   ctlharvesterv1.SettingCache
+	vmBackup  ctlharvesterv1.VirtualMachineBackupCache
+	vmRestore ctlharvesterv1.VirtualMachineRestoreCache
 }
 
 func (v *restoreValidator) Resource() types.Resource {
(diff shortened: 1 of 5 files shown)


pkg/webhook/resources/restore/validator.go
```diff
@@ -92,6 +96,12 @@ func (v *restoreValidator) Create(request *types.Request, newObj runtime.Object)
 		return werror.NewInvalidError(fmt.Sprintf("please stop the VM %q before doing a restore", vm.Name), fieldTargetName)
 	}
 
+	if result, err := util.HasInProgressingVMRestoreOnSameTarget(v.vmRestore, vm.Namespace, vm.Name); err != nil {
+		return werror.NewInternalError(fmt.Sprintf("can't check whether there is another in progressing vmrestore, err: %+v", err))
+	} else if result {
+		return werror.NewInvalidError(fmt.Sprintf("please wait for previous vmrestore complete on the same target %s/%s", vm.Namespace, vm.Name), fieldTargetName)
```
bk201, line 102: The feature works well.
How about let's rephrase the messages more user-friendly (end user):
Screen Shot 2022-07-29 at 12 21 40 PM

`vmrestore` looks like an internal variable naming.

pkg/webhook/resources/restore/validator.go
```diff
@@ -92,6 +96,12 @@ func (v *restoreValidator) Create(request *types.Request, newObj runtime.Object)
 		return werror.NewInvalidError(fmt.Sprintf("please stop the VM %q before doing a restore", vm.Name), fieldTargetName)
 	}
 
+	if result, err := util.HasInProgressingVMRestoreOnSameTarget(v.vmRestore, vm.Namespace, vm.Name); err != nil {
+		return werror.NewInternalError(fmt.Sprintf("can't check whether there is another in progressing vmrestore, err: %+v", err))
```
guangbochen, line 100:
```suggestion
		return werror.NewInternalError(fmt.Sprintf("failed to get the VM-related restores, err: %+v", err))
```

pkg/webhook/resources/restore/validator.go
guangbochen: ditto, may change the words to `please wait for the previous VM restore on the $ns/name to be complete first.`

pkg/webhook/util/filter.go
```diff
@@ -20,3 +24,21 @@ func HasInProgressingVMBackupBySourceUID(cache ctlharvesterv1.VirtualMachineBack
...
 }
+
+func HasInProgressingVMRestoreOnSameTarget(cache ctlharvesterv1.VirtualMachineRestoreCache, targetNamespace, targetName string) (bool, error) {
+	vmRestores, err := cache.GetByIndex(indexeres.VMRestoreByTargetNamespaceAndName, fmt.Sprintf("%s-%s", targetNamespace, targetName))
+	if err != nil && !apierrors.IsNotFound(err) {
+		return false, err
+	}
+
+	for _, vmRestore := range vmRestores {
+		if vmRestore != nil && vmRestore.Status != nil {
```
Vicente-Cheng, line 35: Just curious to know what situation would make `vmRestore` nil if they are already inited?

Vicente-Cheng (commented):
overall LGTM besides the above message improvement and something in curious to know

pkg/webhook/resources/restore/validator.go
FrankYang0529: Fixed it in https://github.com/harvester/harvester/pull/2601. Thank you.

pkg/webhook/resources/restore/validator.go
FrankYang0529: Fixed it in https://github.com/harvester/harvester/pull/2601. Thank you.

pkg/webhook/util/filter.go
FrankYang0529: I don't think there is `nil` vmrestore. However, the return type for `cache.GetByIndex` is `[]*harvesterv1.VirtualMachineRestore`, so I would like to add the check to prevent panic in Golang.

FrankYang0529:
Close the issue since we would like to fix it in master-head first. Please track PR https://github.com/harvester/harvester/pull/2601.

bk201:
@FrankYang0529 We can still keep this one :) Just we prefer to fix in master and backport later.

Issue #2560, opened 2022-07-25
[BUG] VM hanging on restoring state when backup-target disconnected suddenly


lanfon72:
### Describe the bug
When restoring a VM and the backup-target suddenly disconnected, the VM will hang on `restoring` state and can not be deleted.

### To Reproduce
1. Install Harvester with any nodes
2. Login to Dashboard then navigate to _Advanced/Settings_, setup `backup-target` with NFS or S3
3. Create Image for VM creation
4. Create VM `vm1`
5. Take Backup `vm1b` from `vm1`
6. Restore the backup `vm1b` to New/Existing VM
7. When the VM still in `restoring` state, update **backup-target** settings to _Use the default value_ then setup it back.

### Expected behavior
VM might not be able to restore successfully, but at least to be deleted successfully.

## Environment:
 - Harvester ISO version: **v1.0.3-rc1**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **qemu/KVM 3 nodes**

### Additional context
N/A

lanfon72:
update to `reproduce/often` because I can delete the VM sometimes.

FrankYang0529:
Hi @lanfon72, thanks for filing the issue. I will add a webhook to prevent user update backup-target when there is any in progressing VMBackup/VMRestore. If you want to simulate network issues, probably you need to cut off network on KVM. It's cause that updating backup-target, LH will try to remove old LH Backup on system.

lanfon72:
Verified this bug has been fixed.

![image](https://user-images.githubusercontent.com/5169694/182815313-71dd4e4e-51a2-4510-a70a-180765848334.png)
![image](https://user-images.githubusercontent.com/5169694/182815277-98baa7bc-42d1-4404-be87-d60f3b6ba1fd.png)


Test Information
----
* Environment: **qemu/KVM 2 nodes**
* Harvester Version: **v1.0-fc72be9f-head**
* **ui-source** Option: **Auto**
* **ui-index** URL: https://releases.rancher.com/harvester-ui/dashboard/release-harvester-v1.0/index.html

### Verify Steps:
1. Install Harvester with any nodes
2. Navigate to _Advanced/Settings_ and update **ui-index** to `https://releases.rancher.com/harvester-ui/dashboard/release-harvester-v1.0/index.html` and force refresh to make it applied.
3. Create a Image for VM creation
4. Create VM `vm1` and add some files or execute `ping 127.0.0.1 > hello`
[...]

PR #2575, opened 2022-07-27
fix(webhook/backup-target): check in progressing vmrestore

FrankYang0529:
**Problem:**
Users can change backup target when there is in progressing vmrestore.

**Solution:**
Add a webhook to prevent it.

**Related Issue:**
https://github.com/harvester/harvester/issues/2560

**Test plan:**

4. Create VM `vm1`
5. Take Backup `vm1b` from `vm1`
6. Restore the backup `vm1b` to New/Existing VM
7. When the VM still in `restoring` state, update **backup-target** settings to _Use the default value_ then setup it back. Webhook should return an invalid message.

### Additional context
Since v1.0.3 released date is close, I create the PR for `v1.0` first. I will cherry-pick it to `master` branch.


FrankYang0529:
Close the issue since we would like to fix this in `master-head` first. Please check https://github.com/harvester/harvester/pull/2602.

PR #2576, opened 2022-07-27
Allow passing RKE2_IMAGE_REPO variable

bk201:
This makes it possible to use an alternative RKE2 URL during `make build-iso`.

Signed-off-by: Kiefer Chang <kiefer.chang@suse.com>


Vicente-Cheng (approved):
LGTM

guangbochen (approved):
LGTM

PR #2579, opened 2022-07-27
image updates: harvester-upgrade and pre-delete job image

bk201:
image update:
- `rancher/harvester-upgrade`
- Helm chart pre-delete job image: https://github.com/harvester/harvester/blob/1c06027537a7d1ad035501842e4ef3c099ef5dbf/deploy/charts/harvester/values.yaml#L325

PR #2580, opened 2022-07-27
[backport v1.0.3] image updates: harvester-upgrade and pre-delete job image

bk201:
backport of https://github.com/harvester/harvester/pull/2579

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -316,17 +316,9 @@ jobs:
         ##
         image:
 
-          ## Specify the pull policy of image.
+          ## Note we use the harvester-upgrade image. It contains kubectl.
           ##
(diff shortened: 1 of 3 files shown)


FrankYang0529 (approved):
LGTM

deploy/charts/harvester/values.yaml
```diff
@@ -322,11 +322,11 @@ jobs:
 
           ## Specify the repository of image.
           ##
-          repository: bitnami/kubectl
+          repository: rancher/harvester-upgrade
```
guangbochen: maybe we can just update the job template to use the `upgrade.image.repository` value directly.

bk201: Fixed by updating template. Thanks!

guangbochen (approved):
LGTM, thanks.

PR #2582, opened 2022-07-27
Add a script to check if image tags are up to date [CI SKIP]

bk201:
Sometimes, we forget to bump a dependency component's image tag when cutting a release.
The script checks if there are newer commits after the coded tag in values.yaml file.

Example:
Screen Shot 2022-07-27 at 8 50 19 PM

### DIFF
--- a/scripts/check-dep-image-tags
+++ b/scripts/check-dep-image-tags
@@ -1,0 +1,81 @@
+#!/bin/bash -e
+
+TOP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
+VALUES_FILE=$TOP_DIR/deploy/charts/harvester/values.yaml
+
+: ${INSTALLER_BRANCH:="v1.0"}
+
+if [ -z ${WORK_DIR+x} ]; then
+  WORK_DIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


scripts/check-dep-image-tags
```diff
@@ -0,0 +1,79 @@
...
+
+
+build_meta() {
+  # metadata format 
+  # <git_URL> <branch_name> <version>
+  # version:
+  # (1) Just a version number with quote ("")
+  # (2) A YAML key in values.yaml file
+  cat > $META_FILE <<EOF
+https://github.com/harvester/harvester-installer v1.0 "$(get_installer_tag)"
```
guangbochen: Should we set the `v1.0` as env like `HARV_BRANCH_VERSION` for better maintenance and reusability?

guangbochen (commented):
We should also add this to our confluence's release page so this script will be used when people take it as a reference, thanks.

scripts/check-dep-image-tags
bk201: Add in https://github.com/harvester/harvester/compare/8dad678195e9c9b6b0fa6df762dd228fe8fb5661..228c5a44e6b59992a1616fe4e207d9bf482f3263

guangbochen (approved):
LGTM, thanks, merge this one and we can use it for the v1.0 release test.

PR #2601, opened 2022-08-01
fix(webhook/vmrestore): check on the same target

FrankYang0529:
**Problem:**

**Solution:**
Add webhook to prevent it.

**Related Issue:**

**Test plan:**
4. Create VM `vm1`
7. Stop VM `vm1`

### DIFF
--- a/pkg/webhook/resources/restore/validator.go
+++ b/pkg/webhook/resources/restore/validator.go
@@ -68,10 +68,10 @@ func (v *restoreValidator) Create(request *types.Request, newObj runtime.Object)
 	newVM := newRestore.Spec.NewVM
 
 	if targetVM == "" {
-		return werror.NewInvalidError("target VM name is empty", fieldTargetName)
+		return werror.NewInvalidError("Target VM name is empty", fieldTargetName)
 	}
 	if backupName == "" {
-		return werror.NewInvalidError("backup name is empty", fieldVirtualMachineBackupName)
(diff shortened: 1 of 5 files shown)


pkg/webhook/resources/restore/validator.go
```diff
@@ -92,6 +96,12 @@ func (v *restoreValidator) Create(request *types.Request, newObj runtime.Object)
 		return werror.NewInvalidError(fmt.Sprintf("please stop the VM %q before doing a restore", vm.Name), fieldTargetName)
 	}
 
+	if result, err := util.HasInProgressingVMRestoreOnSameTarget(v.vmRestore, vm.Namespace, vm.Name); err != nil {
+		return werror.NewInternalError(fmt.Sprintf("failed to get the VM-related restores, err: %+v", err))
```
guangbochen: I've found that we use uppercase in some places and lowercase in others; it might be better to be consistent with using uppercase only.

FrankYang0529: Update it. Thank you.

guangbochen (approved):
LGTM, thanks.

Vicente-Cheng (approved):
As I said in #2568, just curious about that when would `vmRestore` be nil.
It looks like that is the safer checking. So we keep it.

LGTM and thanks for explaining.

bk201 (approved):
lgtm

PR #2602, opened 2022-08-01
fix(webhook/backup-target): check in progressing vmrestore

FrankYang0529:
**Problem:**
Users can change backup target when there is in progressing vmrestore.

**Solution:**
Add a webhook to prevent it.

**Related Issue:**
https://github.com/harvester/harvester/issues/2560

**Test plan:**

4. Create VM `vm1`
5. Take Backup `vm1b` from `vm1`
6. Restore the backup `vm1b` to New/Existing VM
7. When the VM still in `restoring` state, update **backup-target** settings to _Use the default value_ then setup it back. Webhook should return an invalid message.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -58,11 +58,13 @@ func NewValidator(
 	settingCache ctlv1beta1.SettingCache,
 	vmBackupCache ctlv1beta1.VirtualMachineBackupCache,
 	snapshotClassCache ctlsnapshotv1.VolumeSnapshotClassCache,
+	vmRestoreCache ctlv1beta1.VirtualMachineRestoreCache,
 ) types.Validator {
 	validator := &settingValidator{
 		settingCache:       settingCache,
 		vmBackupCache:      vmBackupCache,
 		snapshotClassCache: snapshotClassCache,
+		vmRestoreCache:     vmRestoreCache,
 	}
 	validateSettingFuncs[settings.BackupTargetSettingName] = validator.validateBackupTarget
 	validateSettingFuncs[settings.VolumeSnapshotClassSettingName] = validator.validateVolumeSnapshotClass
@@ -75,6 +77,7 @@ type settingValidator struct {
 	settingCache       ctlv1beta1.SettingCache
 	vmBackupCache      ctlv1beta1.VirtualMachineBackupCache
 	snapshotClassCache ctlsnapshotv1.VolumeSnapshotClassCache
+	vmRestoreCache     ctlv1beta1.VirtualMachineRestoreCache
 }
 
 func (v *settingValidator) Resource() types.Resource {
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -228,12 +231,20 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
 
 	vmBackups, err := v.vmBackupCache.List(metav1.NamespaceAll, labels.Everything())
 	if err != nil {
-		return werror.NewInternalError(err.Error())
+		return werror.NewInternalError(fmt.Sprintf("Can't list VMBackup, err: %+v", err.Error()))
```
guangbochen:
```suggestion
		return werror.NewInternalError(fmt.Sprintf("Can't list VM backups, err: %+v", err.Error()))
```

pkg/webhook/resources/setting/validator.go
```diff
@@ -228,12 +231,20 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
...
 	}
 	if hasVMBackupInCreatingOrDeletingProgress(vmBackups) {
 		return werror.NewBadRequest("There is VMBackup in creating or deleting progress")
 	}
 
+	vmRestores, err := v.vmRestoreCache.List(metav1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("failed to get the VM-related restores, err: %+v", err.Error()))
```
guangbochen:
```suggestion
		return werror.NewInternalError(fmt.Sprintf("Failed to get the VM restore objects, err: %+v", err.Error()))
```

pkg/webhook/resources/setting/validator.go
```diff
@@ -228,12 +231,20 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
...
+	}
+	if hasVMRestoreInCreatingOrDeletingProgress(vmRestores) {
+		return werror.NewBadRequest("There is a VM-related restore in creating or deleting progress")
```
guangbochen:
```suggestion
		return werror.NewBadRequest("Can't update the backup target, has a VM either in the process of creating or deleting")
```

pkg/webhook/resources/setting/validator.go
FrankYang0529: Updated it. Thank you.

pkg/webhook/resources/setting/validator.go
FrankYang0529: Updated it. Thank you.

pkg/webhook/resources/setting/validator.go
FrankYang0529: Updated it. Thank you.

guangbochen (approved):
LGTM, thanks.

pkg/webhook/resources/setting/validator.go
```diff
@@ -228,12 +231,20 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
...
 	if hasVMBackupInCreatingOrDeletingProgress(vmBackups) {
 		return werror.NewBadRequest("There is VMBackup in creating or deleting progress")
 	}
 
+	vmRestores, err := v.vmRestoreCache.List(metav1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("Failed to get the VM restore objects, err: %+v", err.Error()))
+	}
+	if hasVMRestoreInCreatingOrDeletingProgress(vmRestores) {
+		return werror.NewBadRequest("Can't update the backup target, has a VM either in the process of creating or deleting")
```
Vicente-Cheng: `has a VM either ...` --> `there is a VM either ...` sounds better?

FrankYang0529: Yes, sounds good to me. How about `Can't update the backup target. There is a VM either in the process of creating or deleting`

Vicente-Cheng: agree 👍🏼 , looks good!

FrankYang0529: Updated it. Thank you.

bk201 (approved):
LGTM!

Vicente-Cheng (approved):
LGTM, thanks!

PR #2608, opened 2022-08-02
[backport v1.0.3] fix(webhook/vmrestore): check on the same target

FrankYang0529:
Backport #2601 

**Problem:**

**Solution:**
Add webhook to prevent it.

**Related Issue:**

**Test plan:**
4. Create VM `vm1`
7. Stop VM `vm1`


guangbochen (approved):
LGTM

bk201 (approved):
LGTM, thanks for the backport.

PR #2614, opened 2022-08-02
chart: Bump NDM to v0.4.5

Vicente-Cheng:
bump NDM to v0.4.5 on master.

(I would backport it to v1.0 after it merged)


guangbochen (commented):
`helm dep update` will re-generate the local dependency tarball even if there is no changes, so we would prefer to skip adding those tgz files if they are not modified, e.g., the csi-snapshotter and kubevirt charts.

Vicente-Cheng:
> `helm dep update` will re-generate the local dependency tarball even if there is no changes, so we would prefer to skip adding those tgz files if they are not modified, e.g., the csi-snapshotter and kubevirt charts.

OK, thanks @guangbochen for tips on this!

I push a new version and do not change the tarballs that are already here.

please review it again, thanks :)

bk201 (approved):

guangbochen (approved):
LGTM, will also need to bump it in the v1.0 branch, thanks.

Issue #2610, opened 2022-08-02
[BUG]  A restore action that does not meet expectations


WuJun2016:
**Describe the bug**

A vm is not backed up, but the restore action appears after stopping the vm

**To Reproduce**
1. create a vm
2. stop vm
3. Check vm's Action menu，You can see the restore action button
![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/0da1d29b-a18b-4b0e-82f5-497f431bc4e0)

**Expected behavior**

vm is not backed up, there should not be a restore operation

**Support bundle**

**Environment**
 - Harvester ISO version:  v1.0.3-head   (August 2)
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

TachunLin:
Verified **fixed** on `master-9631b136-head` (08/16). Close this issue.

### Result
Once the VM haven't take the backup, there will be no `Restore` option display on the VM options under the following scenarios
  1. Stop the VM, 
    ![image](https://user-images.githubusercontent.com/29251855/185862704-fb992049-3e71-44fb-8953-079c5b4cd1a5.png)
  
  2. Restart the VM
    ![image](https://user-images.githubusercontent.com/29251855/185863654-feb22d7e-c003-477b-9431-12c3db2ae58e.png)
    
  3. Pause the VM
    ![image](https://user-images.githubusercontent.com/29251855/185865040-36c74ef2-86de-4af2-a631-270535efbf74.png)

* The `Restore` option only exists when VM already have a backup and in `Stop` mode 
![image](https://user-images.githubusercontent.com/29251855/185866956-752e6dcd-f3b2-4c25-b6a9-0270a19d1830.png)


### Test Information
* Test Environment: 1 node harvester on local kvm machine
* Harvester version: `master-9631b136-head` (08/16)

### Verify Steps
1. Create a VM 
2. Stop the VM
3. Restart the VM
4. Pause the VM
5. Unpause the VM
6. Set Backup target in settings
7. Take the backup of current VM
8. Stop the VM

PR #2623, opened 2022-08-03
fix(api): check whether vm has backups before allowing restore

FrankYang0529:
**Problem:**
VM shows restore action when there is no backup.

**Solution:**
Fix `canDoRestore` function.

**Related Issue:**
https://github.com/harvester/harvester/issues/2610

**Test plan:**

* Create a cluster.
* Create a VM `vm-test`.
* Run `echo "123" > test.txt && sync` in `vm-test`.
* Take a backup for it.
* After backup finished, run `echo "456" > test.txt && sync` in `vm-test`.
* Stop the VM `vm-test`.
* After the related VMI is gone, we can see `Restore` action in the list. Restore the backup to `vm-test`.
Screen Shot 2022-08-03 at 5 55 33 PM

* Run `cat test.txt` in `vm-test`. It should show `123`.
* Delete the VM `vm-test`.
* Create a new VM with the same name `vm-test`.
* Stop the VM `vm-test`. After the related VMI is gone, we _can't_ see `Restore` action in the list, because they don't have the same `UID`.

Screen Shot 2022-08-03 at 6 05 09 PM

### DIFF
--- a/pkg/api/register.go
+++ b/pkg/api/register.go
@@ -12,6 +12,7 @@ import (
 	"github.com/harvester/harvester/pkg/api/vmtemplate"
 	"github.com/harvester/harvester/pkg/api/volume"
 	"github.com/harvester/harvester/pkg/config"
+	"github.com/harvester/harvester/pkg/indexeres"
 )
 
 type registerSchema func(scaled *config.Scaled, server *server.Server, options config.Options) error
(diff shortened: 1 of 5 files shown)


guangbochen (approved):
LGTM

futuretea (approved):
LGTM

bk201 (approved):
LGTM!

Issue #1436, opened 2021-10-15
[FEATURE]  Allowing users to download backing images


guangbochen:
**Is your feature request related to a problem? Please describe.**
Allowing users to download the backing images from UI, so that those images can be shared accordingly.

Depends on: https://github.com/longhorn/longhorn/issues/3155

shuo-wu:
The corresponding Longhorn backend API is `/v1/backingimages/{name}/download`

And the Longhorn UI PR is: https://github.com/longhorn/longhorn-ui/pull/519 (This PR also includes one upload-API-related modification, which fits the recent backing image refactoring)

lanfon72:
Verified this feature has been implemented.

![image](https://user-images.githubusercontent.com/5169694/189675005-a7509189-f0c3-42e4-b5a4-d8c1bc1f6341.png)


Test Information
----
* Environment: **qemu/KVM 2 nodes**
* Harvester Version: **master-a579442f-head**
* **ui-source** Option: **Auto**

### Verify Steps:
1. Install Harvester with any nodes
2. Create a Image `img1`
3. Click the details of `img1`, **Download** Button should be available
4. Click **Download** button, `img1` should able to be downloaded and downloaded successfully.

PR #2639, opened 2022-08-08
Implement the feature about download backing image

Vicente-Cheng:

**Problem:**
Let user could download backing images

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/1436

**Test plan:**

### DIFF
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -39,7 +39,7 @@ func Formatter(request *types.APIRequest, resource *types.RawResource) {
 	}
 }
 
-type UploadActionHandler struct {
+type ImageHandler struct {
 	httpClient                  http.Client
 	Images                      v1beta1.VirtualMachineImageClient
 	ImageCache                  v1beta1.VirtualMachineImageCache
@@ -47,7 +47,7 @@ type UploadActionHandler struct {
 	BackingImageDataSourceCache ctllhv1beta1.BackingImageDataSourceCache
 }
 
-func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+func (h ImageHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	if err := h.do(rw, req); err != nil {
 		status := http.StatusInternalServerError
 		if e, ok := err.(*apierror.APIError); ok {
@@ -60,7 +60,7 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
 	rw.WriteHeader(http.StatusOK)
 }
 
-func (h UploadActionHandler) do(rw http.ResponseWriter, req *http.Request) error {
+func (h ImageHandler) do(rw http.ResponseWriter, req *http.Request) error {
 	vars := mux.Vars(req)
 	action := vars["action"]
 	switch action {
@@ -71,7 +71,7 @@ func (h UploadActionHandler) do(rw http.ResponseWriter, req *http.Request) error
 	}
 }
 
-func (h UploadActionHandler) uploadImage(rw http.ResponseWriter, req *http.Request) error {
+func (h ImageHandler) uploadImage(rw http.ResponseWriter, req *http.Request) error {
 	vars := mux.Vars(req)
 	namespace := vars["namespace"]
 	name := vars["name"]
@@ -127,7 +127,7 @@ func (h UploadActionHandler) uploadImage(rw http.ResponseWriter, req *http.Reque
 	return nil
 }
 
-func (h UploadActionHandler) waitForBackingImageDataSourceReady(name string) error {
+func (h ImageHandler) waitForBackingImageDataSourceReady(name string) error {
 	retry := 30
 	for i := 0; i < retry; i++ {
 		ds, err := h.BackingImageDataSources.Get(util.LonghornSystemNamespaceName, name, metav1.GetOptions{})
@@ -147,7 +147,7 @@ func (h UploadActionHandler) waitForBackingImageDataSourceReady(name string) err
 	return errors.New("timeout waiting for backing image data source to be ready")
 }
 
-func (h UploadActionHandler) updateImportedConditionOnConflict(image *apisv1beta1.VirtualMachineImage,
+func (h ImageHandler) updateImportedConditionOnConflict(image *apisv1beta1.VirtualMachineImage,
 	status, reason, message string) error {
 	retry := 3
 	for i := 0; i < retry; i++ {
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -94,7 +94,7 @@ func (h ImageHandler) uploadImage(rw http.ResponseWriter, req *http.Request) err
 		return err
 	}
 
-	uploadUrl := fmt.Sprintf("http://longhorn-backend.longhorn-system:9500/v1/backingimages/%s-%s", namespace, name)
+	uploadUrl := fmt.Sprintf("%s/backingimages/%s-%s", util.LonghornDefaultManagerURL, namespace, name)
 	uploadReq, err := http.NewRequestWithContext(req.Context(), http.MethodPost, uploadUrl, req.Body)
 	if err != nil {
 		return fmt.Errorf("failed to create the upload request: %w", err)
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -2,6 +2,7 @@ package image
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -25,7 +26,8 @@ import (
 )
 
 const (
-	actionUpload = "upload"
+	actionUpload   = "upload"
+	actionDownload = "download"
 )
 
 func Formatter(request *types.APIRequest, resource *types.RawResource) {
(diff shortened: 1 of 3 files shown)


pkg/api/image/formatter.go
```diff
@@ -60,18 +62,60 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
...
+	if err != nil {
+		return fmt.Errorf("Failed to send the upload request: %w", err)
+	}
+	defer downloadResp.Body.Close()
+
+	if downloadResp.StatusCode != http.StatusOK {
+		return fmt.Errorf("Unexpected http Status code %d.", downloadResp.StatusCode)
+	}
+
+	rw.Header().Set("Content-Disposition", "attachment; filename="+name)
```
FrankYang0529: May we set the file name to `vmimage.spec.displayName`? Currently, `name` will be something like `image-<random-string>`. It's hard for users to map to vmimage on dashboard.

pkg/api/image/schema.go
```diff
@@ -17,10 +17,19 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, options config
...
-				actionUpload: UploadActionHandler{
+				operationUpload: ImageHandler{
+					httpClient:                  http.Client{},
+					Images:                      scaled.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineImage(),
+					ImageCache:                  scaled.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineImage().Cache(),
+					BackingImageDataSources:     scaled.LonghornFactory.Longhorn().V1beta1().BackingImageDataSource(),
+					BackingImageDataSourceCache: scaled.LonghornFactory.Longhorn().V1beta1().BackingImageDataSource().Cache(),
+				},
+			}
+			s.LinkHandlers = map[string]http.Handler{
```
FrankYang0529, line 46: May we add some comments here for the difference between `ActionHandlers` and `LinkHandlers`? Thank you.

FrankYang0529 (approved):
Function works as expected. LGTM. Thanks.

pkg/api/image/formatter.go
Vicente-Cheng: sure, that makes sense to me!

pkg/api/image/schema.go
Vicente-Cheng: OK, I will add some comments for `ActionHandlers` and `LinkHandlers`!

pkg/api/image/formatter.go
```diff
@@ -60,18 +62,60 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
...
+	bkimgName := fmt.Sprintf("%s-%s", namespace, name)
+	downloadUrl := fmt.Sprintf("http://longhorn-backend.longhorn-system:9500/v1/backingimages/%s/download", bkimgName)
+	downloadReq, err := http.NewRequestWithContext(req.Context(), http.MethodGet, downloadUrl, nil)
+	if err != nil {
+		return fmt.Errorf("Failed to create the download request: %w", err)
+	}
+
+	downloadResp, err := h.httpClient.Do(downloadReq)
+	if err != nil {
+		return fmt.Errorf("Failed to send the upload request: %w", err)
```
w13915984028: download request

pkg/api/image/formatter.go
```diff
@@ -60,18 +62,60 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
...
 	}
 }
 
-func (h UploadActionHandler) uploadImage(rw http.ResponseWriter, req *http.Request) error {
+func (h ImageHandler) downloadImage(rw http.ResponseWriter, req *http.Request) error {
+	vars := mux.Vars(req)
+	namespace := vars["namespace"]
+	name := vars["name"]
+
+	bkimgName := fmt.Sprintf("%s-%s", namespace, name)
```
w13915984028, line 104: should check both namespace and name are not "" ?

pkg/api/image/formatter.go
```diff
@@ -60,18 +62,60 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
...
+	downloadUrl := fmt.Sprintf("http://longhorn-backend.longhorn-system:9500/v1/backingimages/%s/download", bkimgName)
+	downloadReq, err := http.NewRequestWithContext(req.Context(), http.MethodGet, downloadUrl, nil)
+	if err != nil {
+		return fmt.Errorf("Failed to create the download request: %w", err)
```
w13915984028: suggest to add `bkimgName` into error message for here and following

pkg/api/image/formatter.go
```diff
@@ -60,18 +62,60 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
...
-	switch action {
-	case actionUpload:
+	operation := ""
+	if _, found := vars["action"]; found {
+		operation = vars["action"]
+	} else if _, found := vars["link"]; found {
+		operation = vars["link"]
+	}
+	switch operation {
+	case operationUpload:
```
w13915984028: wondering, if mixing `action` and `operation` is better

vars["action"] represents `upload`, vars["link"] represent `download`

when keeping name `action` unchanged, the only tricky is that when vars["links"] is existing, the `action` is changed from `upoad` to `download`, but it's still an action


when reading L40 `resource.AddAction(request, operationUpload)` , it will be a bit confusing

pkg/api/image/formatter.go
Vicente-Cheng: nice catch! thanks :)

pkg/api/image/formatter.go
Vicente-Cheng: I am thinking about that both `namespace` and `name` come from API. That should not be empty or there is something error that would have happened early.

pkg/api/image/formatter.go
Vicente-Cheng: sure, I would improve the error logging.

pkg/api/image/formatter.go
Vicente-Cheng: done

pkg/api/image/schema.go
Vicente-Cheng: Already add some comments for `ActionHandlers` and `LinkHandlers`.
Please check again, thanks!

pkg/api/image/formatter.go
Vicente-Cheng: fixed!

pkg/api/image/formatter.go
Vicente-Cheng: As @FrankYang0529 suggested, I change the output file name to the user-readable image name.
It would get the Image first, so it will fail when getting the image failure.
Maybe we do not need another check?

pkg/api/image/formatter.go
Vicente-Cheng: Already change the logging, please check again.

pkg/api/image/formatter.go
Vicente-Cheng: Already rewrite here.

I prefer to use `operation` because maybe we need to implement other image operations in the future.
The `action` means the operation is`POST` method.
The `link` means the operation is`GET` method.

That, the `operation` to present what we want to do next. No matter where it comes from `action` or `link`.

Vicente-Cheng:
hi @FrankYang0529, @w13915984028 already update the new version.
Please help to check again :)

Vicente-Cheng:
Add WIP back because I am checking about RABC constraints.
You could still review the download behavior. Thanks!

pkg/api/image/formatter.go
w13915984028: Please take a look of other related files, all are following `action` naming.  It's better for `image/formatter` also in same line, even though it may not fully meet your expectation.
 
That will not confuse other engineer when they read&maintain code.

```
node/formatter.go
volume/handler.go
vm/handler.go
...
```

pkg/api/image/formatter.go
w13915984028: Please take a look of LH API, hardcode LH backend may break someday. Both upload and download.

```
vendor/github.com/longhorn/longhorn-manager/types/types.go
	DefaultAPIPort           = 9500
func GetDefaultManagerURL() string {
	return "http://longhorn-backend:" + strconv.Itoa(DefaultAPIPort) + "/v1"
}
```

pkg/api/image/formatter.go
```diff
@@ -25,7 +26,8 @@ import (
 )
 
 const (
-	actionUpload = "upload"
+	operationUpload   = "upload"
```
w13915984028: Please follow existing naming rules, explained in L72.

pkg/api/image/formatter.go
bk201: making them constant would be good enough at this stage: https://github.com/harvester/harvester/blob/538d41fe793c41360cb170f128206c768ce83735/pkg/util/constants.go#L22
We can do whole code base refactoring by using the [namespace value pass via Helm chart values](https://github.com/harvester/harvester/blob/538d41fe793c41360cb170f128206c768ce83735/deploy/charts/harvester/values.yaml#L377) in the future.

pkg/api/image/formatter.go
Vicente-Cheng: Already change back to `action`. Please check it again, thanks :)

pkg/api/image/formatter.go
Vicente-Cheng: Add into constants, and use the related type on both upload and download. 
Please check it again, thanks. :)

Vicente-Cheng:
Remove WIP again because I am checked about RABC constraints.
Now works well with the owner, member as expected.

pkg/api/image/schema.go
bk201: Perhaps we can create just one ImageHandler instance for Link and Action.

Vicente-Cheng: good suggestion! Done.

pkg/api/image/formatter.go
```diff
@@ -60,18 +62,76 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
...
+
+	return apierror.NewAPIError(validation.InvalidAction, "Unsupported method")
+}
+
+func (h ImageHandler) doGet(link string, rw http.ResponseWriter, req *http.Request) error {
+	switch link {
+	case actionDownload:
+		return h.downloadImage(rw, req)
+	default:
+		return apierror.NewAPIError(validation.InvalidAction, "Unsupported GET action")
```
w13915984028: add action string into error message will be better

pkg/api/image/formatter.go
```diff
@@ -60,18 +62,76 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
...
+	}
+}
+
+func (h ImageHandler) doPost(action string, rw http.ResponseWriter, req *http.Request) error {
 	switch action {
 	case actionUpload:
 		return h.uploadImage(rw, req)
 	default:
-		return apierror.NewAPIError(validation.InvalidAction, "Unsupported action")
+		return apierror.NewAPIError(validation.InvalidAction, "Unsupported POST action")
```
w13915984028: add action string into error message will be better

pkg/api/image/formatter.go
```diff
@@ -60,18 +62,76 @@ func (h UploadActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request
...
+	}
+}
+
+func (h ImageHandler) downloadImage(rw http.ResponseWriter, req *http.Request) error {
+	vars := mux.Vars(req)
+	namespace := vars["namespace"]
+	name := vars["name"]
+	vmImage, err := h.Images.Get(namespace, name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed to get VMImage.")
```
w13915984028: add namespace, name, `err` into error message

FrankYang0529 (approved):
LGTM. Thanks!

pkg/api/image/formatter.go
```diff
@@ -127,7 +187,7 @@ func (h UploadActionHandler) uploadImage(rw http.ResponseWriter, req *http.Reque
 	return nil
 }
 
-func (h UploadActionHandler) waitForBackingImageDataSourceReady(name string) error {
+func (h ImageHandler) waitForBackingImageDataSourceReady(name string) error {
 	retry := 30
```
w13915984028, line 191: not sure `60s` is enough for slow environment. Is it marked in Harvester Doc ?

w13915984028 (commented):
Almost done. Please check those last polishments, thanks.

pkg/api/image/formatter.go
bk201: @w13915984028 The value here is not added in this PR. Please create another issue if you feel it's not reasonable in the image upload path.

pkg/api/image/formatter.go
Vicente-Cheng: sure, that would make the error log more clarify!
updated

pkg/api/image/formatter.go
Vicente-Cheng: done

pkg/api/image/formatter.go
Vicente-Cheng: done

pkg/api/image/formatter.go
Vicente-Cheng: Thanks, @bk201, 
Agree. that 30s is the previous commits mechanism.
I thought that if we want a more timeout, we could raise another issue to discuss what value is more suitable.

Vicente-Cheng:
Hi all, please check again with the latest version.
Thanks :)

w13915984028 (approved):
LGTM, thanks.

pkg/api/image/formatter.go
w13915984028: Let's track it in https://github.com/harvester/harvester/issues/2718 .

bk201 (approved):
lgtm

guangbochen (approved):

Issue #2630, opened 2022-08-04
[FEATURE] Sync image `display_name` to image labels


futuretea:
**Is your feature request related to a problem? Please describe.**

For cli tools for Harvester, such as terraform-provider-harvester, when get image by `display_name`
need a quick get method rather than list all images

**Describe the solution you'd like**

add/update this label in vmimage controller `OnChange` 
add label selector to List method

**Describe alternatives you've considered**

**Additional context**

futuretea:
For test:
check images name duplicate validation still works as excepted:
![image](https://user-images.githubusercontent.com/15064560/187172259-3ea6aaa0-4564-4c22-8b6d-c00d8ef73f37.png)

TachunLin:
Verified **fixed** on `master-69a6b1b2-head` (08/29). Close this issue.

### Feature description
Add image displayName to image's labels thus we can use labelSelector to get images when we needed

### Result
* In image API content, label `harvesterhci.io/imageDisplayName` added to labels, and it's value should be the displayName value
  ![image](https://user-images.githubusercontent.com/29251855/187353496-39c20027-f438-43de-a212-4f38b2dfbbae.png)
* Image with the same display name in label would be denied by admission webhook "validator.harvesterhci.io"
  ![image](https://user-images.githubusercontent.com/29251855/187354352-ea2f08f3-01a1-4088-899b-d92e25433781.png)
* Image with the same display name but different url would also be denied 
  ![image](https://user-images.githubusercontent.com/29251855/187355241-845b09b5-953b-4e90-9948-ca8b025a6f5d.png)

### Test Information
* Test Environment: 1 node harvester on local kvm machine
* Harvester version: `master-69a6b1b2-head` (08/29)

### Verify Steps
1. Login harvester dashboard
2. Access the Preference page
3. Enable developer tool 
  ![image](https://user-images.githubusercontent.com/29251855/187353113-495af11e-a3e5-4f8e-b03b-174b4f0660ea.png)
4. Create an ubuntu focal image from url  https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-amd64.img
  ![image](https://user-images.githubusercontent.com/29251855/187353177-52516c6d-8e68-4ac5-8b40-4006f6460773.png)
5. View API of the created image 
  ![image](https://user-images.githubusercontent.com/29251855/187353338-1f0691f3-b19a-4382-a26f-ab5897842474.png)
6. Check can found the display name in the image API content 
[...]

PR #2671, opened 2022-08-17
Sync image display_name to image labels

futuretea:

**Problem:**

In some [case](https://github.com/harvester/terraform-provider-harvester/blob/126661ff5cf35c8992c03198009c0fde50168d2c/internal/provider/image/datasource_image.go#L37),  we need get image by image's displayName:

currently, we have to list all images and check displayName one by one

**Solution:**

If we sync image displayName to image's labels, we can use `labelSelector` to get images we needed

**Related Issue:**
https://github.com/harvester/harvester/issues/2630

**Test plan:**

replace harvester image to `futuretea/harvester:sync-display-name-head`

1. check the labels of existing images by `View In Api`, label `harvesterhci.io/imageDisplayName` will be added to labels, and it's value should be the displayName value

2. create an image by specify an existing displayName
for example, there is an image `harvester-public/ubuntu-20.04-minimal-cloudimg-amd64.img` exist in the cluster, we create a new image `harvester-public/ubuntu-20.04-minimal-cloudimg-amd64.img` again
image
the request should be denied by admission webhook "validator.harvesterhci.io"

### DIFF
--- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -53,6 +53,17 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
 		}
 		return h.initialize(image)
 	}
+
+	// sync display_name to labels in order to list by labelSelector
+	if image.Spec.DisplayName != image.Labels[util.LabelImageDisplayName] {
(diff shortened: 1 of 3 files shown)


chrisho (approved):
LGTM

guangbochen (approved):
LGTM, thanks.

Issue #2647, opened 2022-08-09
[FEATURE] Collect system logs


w13915984028:
**Is your feature request related to a problem? Please describe.**

Collect system level log.
It should be possible to collect log from given path.
Avoid duplicated logs from same path.

**Describe the solution you'd like**

Collect 
following UNITs of journal: `rke2-server.service, rke2-agent.service, rancherd.service,rancher-system-agent.service, wicked.service, iscsid.service`
`kernel message`
`/var/lib/rancher/rke2/agent/logs/kubelet.log`
`/var/log/audit/audit.log`

**Describe alternatives you've considered**

**Additional context**

EPIC:
https://github.com/harvester/harvester/issues/577

HEP DOC:
https://github.com/joshmeranda/harvester/blob/logging/enhancements/20220525-system-logging.md

w13915984028:
FOR TEST:

In v1.1.0, test `logging/event/audit` output type of: Graylog, Splunk, they are required by customers;  
webhook server is for quick validation.

Set output manually :
How to send log to graylog:
https://github.com/w13915984028/harvester-develop-summary/blob/main/integrate-harvester-logging-with-log-servers.md

Test logging, audit, event with webhook server
https://github.com/w13915984028/harvester-develop-summary/blob/main/test-log-event-audit-with-webhook-server.md


Set output in UI:
https://github.com/harvester/harvester/issues/2646
[FEATURE] Config logging in Harvester Dashboard

https://github.com/harvester/harvester/issues/2749
[FEATURE] Harvester UI support config (cluster) flow & output for kube-audit log and event log

lanfon72:
Verified this feature has been implemented.


Test Information
----
* Environment: **qemu/KVM 3 nodes**
* Harvester Version: **master-250f41e4-head**
* **ui-source** Option: **Auto**

### Verify Steps:
1. Install _Graylog_ via docker[^1]
2. Install Harvester with any nodes
3. Login to Dashboard then navigate to _Monitoring & Logging/Logging_
4. Create **Cluster Output** with following:
    - **Name**: gelf-evts
    - **Type**: `Logging/Event`
    - **Output**: GELF
    - **Target**: `<Graylog_IP>, <Graylog_Port>, <UDP>`
5. Create **Cluster Flow** with following:
    - **Name**: gelf-flow
    - **Type** of Matches: `Logging`
    - **Cluster Outputs**: `gelf-evts`
6. Create an Image for VM creation
7. Create a vm `vm1` and start it
[...]

Issue #2646, opened 2022-08-09
[FEATURE] Config logging in Harvester Dashboard


w13915984028:
**Is your feature request related to a problem? Please describe.**

The logging feature is configurable in Harvester Dashboard.

**Describe the solution you'd like**

Configure Harvester logging:
```
  fluentbit, fluentd POD memory, CPU resources.
  cluster & namespace leve flows and ouputs:
```

Please UI team investigate following Rancher feature, when reusing the YAML input feature is possilbe, that will be great.

https://rancher.com/docs/rancher/v2.5/en/logging/custom-resource-config/flows/
https://rancher.com/docs/rancher/v2.5/en/logging/custom-resource-config/outputs/

**Describe alternatives you've considered**

**Additional context**

EPIC:
https://github.com/harvester/harvester/issues/577

HEP DOC:
https://github.com/joshmeranda/harvester/blob/logging/enhancements/20220525-system-logging.md

w13915984028:
@WuJun2016 @joshmeranda @guangbochen

DaiYuzeng:
Hello @w13915984028 ! Can you please complement the checklist? Thanks! (#2749, #2643 also need to be complemented)

w13915984028:
FOR TEST:

In v1.1.0, test `logging/event/audit` output type of: Graylog, Splunk, they are required by customers;  
webhook server is for quick validation.

Set output manually :
How to send log to graylog:
https://github.com/w13915984028/harvester-develop-summary/blob/main/integrate-harvester-logging-with-log-servers.md

Test logging, audit, event with webhook server
https://github.com/w13915984028/harvester-develop-summary/blob/main/test-log-event-audit-with-webhook-server.md


Set output in UI:
https://github.com/harvester/harvester/issues/2646
[FEATURE] Config logging in Harvester Dashboard

https://github.com/harvester/harvester/issues/2749
[FEATURE] Harvester UI support config (cluster) flow & output for kube-audit log and event log

DaiYuzeng:
The PR from GUI side had been merged so I moved this issue to the `Ready For Testing` panel.
If you have any concerns, just ping me, thank you! @w13915984028 @WuJun2016

lanfon72:
Verified this feature has been implemented.

![image](https://user-images.githubusercontent.com/5169694/191697822-6bc0d7b8-2c56-42e0-805a-408c1ef19845.png)
[...]

Issue #2645, opened 2022-08-09
[FEATURE] Logging is backed by PVC


w13915984028:
**Is your feature request related to a problem? Please describe.**

Logging data should be persistant in the cluster.

**Describe the solution you'd like**

Use PVC to store the data, PVC needs a proper initial size.

**Describe alternatives you've considered**

**Additional context**

EPIC:
https://github.com/harvester/harvester/issues/577

HEP DOC:
https://github.com/joshmeranda/harvester/blob/logging/enhancements/20220525-system-logging.md

update 2022.09.16: 
This feature is moved out from `v1.1.0`.

To persistant logging, user normally have dedicated logging servers. Meanwhile, Harvester may have limited storage, the logging PVC is difficult to set a proper size.

@joshmeranda

w13915984028:
When local test and debugging, a LH issue is encountered and tracked in :

https://github.com/longhorn/longhorn/issues/4546

In single-node cluster, the workaround is to add a 1-replica storageclass for usage of `rancher-monitoring` etc.

w13915984028:
update 2022.09.16:

This feature is moved out from v1.1.0.

To persistant logging, user normally have dedicated logging servers. Meanwhile, Harvester may have limited storage, the logging PVC is difficult to set a proper size.

guangbochen:
@w13915984028 should we close this issue since currently logging is managed by the users? Users may refer to https://github.com/harvester/harvester/blob/master/enhancements/20220525-system-logging.md#viewing-logs-via-grafana-loki if they want to set up the logging system manually within the Harvester cluster.

w13915984028:
@guangbochen I close it.

Not an requirement at the moment.

Issue #2644, opened 2022-08-09
[FEATURE] Enable logging by default


w13915984028:
**Is your feature request related to a problem? Please describe.**

Harvester will support logging feature.

**Describe the solution you'd like**

Harvester enables logging by default, after installation, it is available.

Due to existing limitation, `disable logging` is not available in v1.1.0.

**Describe alternatives you've considered**

**Additional context**

EPIC:
https://github.com/harvester/harvester/issues/577

HEP DOC:
https://github.com/joshmeranda/harvester/blob/logging/enhancements/20220525-system-logging.md

**Future enhancement**
[FEATURE] Harvester supports enable/disable managedchart in day2
https://github.com/harvester/harvester/issues/2719

With it, the `logging` can be enabled/disabled freely.

w13915984028:
The resources usage of logging (https://github.com/harvester/harvester/issues/577). cc @joshmeranda @guangbochen @bk201

(tested in single-node cluster)
```
# kubectl top pod -A


NAMESPACE                   NAME                                                     CPU(cores)   MEMORY(bytes)   

cattle-fleet-local-system   fleet-agent-57497ff7dc-fmlct                             25m          552Mi           

cattle-fleet-system         fleet-controller-5746685958-dd7qx                        3m           166Mi           
cattle-fleet-system         gitjob-cc9948fd7-ng72b                                   1m           121Mi           

cattle-logging-system       rancher-logging-7b54b444c8-2djf4                         3m           118Mi           /// logging operator, one instance in cluster
cattle-logging-system       rancher-logging-root-fluentbit-q9zpz                     4m           30Mi            //// logging agent, per POD
cattle-logging-system       rancher-logging-root-fluentd-0                           5m           159Mi           //// logging server, one instance in cluster

cattle-monitoring-system    alertmanager-rancher-monitoring-alertmanager-0           1m           50Mi            // alertmanager

cattle-monitoring-system    prometheus-rancher-monitoring-prometheus-0               78m          530Mi           

cattle-monitoring-system    rancher-monitoring-grafana-d9c56d79b-jblt5               6m           162Mi
```

@joshmeranda when you have the resources usage of multi-node, please also post it, plus the grafana loki.

The `rancher-logging` creates a new namespace `cattle-logging-system`, which is also new to `Harvester`.

Enable `logging` by default will be good to business customers, the system has logging from beginning. It is also in same user experience as other Harvester features.
[...]

PR #2681, opened 2022-08-19
Logging chart

joshmeranda:
**Problem:**
Implement the logging feature as described in [577](https://github.com/harvester/harvester/issues/577).

**Solution:**
Installer the `rancher-logging` and `rancher-logging-crd` charts as `ManagedCharts` by default and route logs to local loki deployment using the logging operator from BonzaiCloud.

**Related Issue:**
https://github.com/harvester/harvester/issues/577
https://github.com/harvester/harvester/issues/2644
https://github.com/harvester/harvester/issues/2645
https://github.com/harvester/harvester/issues/2646
https://github.com/harvester/harvester/issues/2647

**Test plan:**

cc @guangbochen @bk201 @w13915984028


.drone.yml
```diff
@@ -95,7 +95,7 @@ steps:
     dockerfile: package/Dockerfile.webhook
     password:
       from_secret: docker_password
-    repo: "rancher/harvester-webhook"
+    repo: "joshmeranda/harvester-webhook"
```
bk201: @joshmeranda I guess we can keep all repos with existing values. And don't worry that the PR needs to be tested with another installer PR, reviewers can handle this.

deploy/charts/harvester-journal/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
```
bk201: This looks like a Harvester-specific chart. The convention is to create the chart in the https://github.com/harvester/charts repo and publish it. Then we can use it as a dependency chart in the `harvester/harvester` repo. You can check [other charts](https://github.com/harvester/charts/tree/master/charts) as an example. 

cc @guangbochen

.drone.yml
bk201: And similar changes in other files.

scripts/build-iso
```diff
@@ -7,9 +7,10 @@ cd $(dirname $0)/..
 
 echo "Start building ISO image"
 
-HARVESTER_INSTALLER_VERSION=master
+HARVESTER_INSTALLER_VERSION=logging
```
bk201: ditto, please remove these changes. We can keep the original values for
- build-iso
- ci
- test-integration

deploy/charts/harvester-journal/Chart.yaml
joshmeranda: created PR for harvester-charts: https://github.com/harvester/charts/pull/94/

scripts/build-iso
joshmeranda: reversed changes to those files

.drone.yml
joshmeranda: reversed changes to those files

w13915984028:
Replaced by https://github.com/harvester/harvester-installer/pull/333 , close.

Issue #2662, opened 2022-08-12
[BUG] VM label names should be consistent before and after the restore task is done


starbops:
**Describe the bug**

There's a label in the VirtualMachine resource called `harvesterhci.io/vmName`, which will propagate to the respective launcher Pod `virt-laucher-xxxxx`. Backing up a VM and restoring a new VM with the backup causes the label name to change from `harvesterhci.io/vmName` to `harvesterhci.io/vm-name`, which is inconsistent behavior.

VM before restore:
<img width="731" alt="Screen Shot 2022-08-11 at 15 16 49" src="https://user-images.githubusercontent.com/1827717/184329408-f3d9df93-e4b7-42e6-b58a-9a762dc8434f.png">

VM after restore:
<img width="958" alt="Screen Shot 2022-08-11 at 16 01 25" src="https://user-images.githubusercontent.com/1827717/184329454-0c8e297e-9e02-4136-a8c2-6165df219f30.png">

**To Reproduce**

1. Create one VM (use defaults)
2. Make sure that the label name is `harvesterhci.io/vmName` by checking its YAML
3. Wait for the VM to fully boot up then take a backup
4. After the backup task is done, delete the VM
5. Restore a new VM with the same name using the newly created backup (This is just for the sake of highlighting the inconsistency. Using a different name will still cause the problem. The name of the VM is not the point.)
6. After the restore task is completed, check its YAML for the label name, it should be changed to `harvesterhci.io/vm-name`

**Expected behavior**

The label name should be consistent before and after the backup/restore, i.e. `harvesterhci.io/vmName`.

**Support bundle**

**Environment**

- Harvester ISO version: v1.0.2
- Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Baremetal (Supermicro X10DRL-i/Intel Xeon E5-2680 v4/64GB RAM)

**Additional context**

[...]

PR #2686, opened 2022-08-23
fix(restore): make vmNameLabel consistent after restore

starbops:

**Problem:**

Restoring new VMs using created backup results in "VM name label inconsistency" issue as described in detail in #2662.

**Solution:**

Changing `vmNameLabel`'s value from `harvesterhci.io/vm-name` to `harvesterhci.io/vmName`.

**Related Issue:**

#2662 

**Test plan:**

1. Create one VM (use defaults)
1. Make sure that the label name is `harvesterhci.io/vmName` by checking its YAML
1. Wait for the VM to fully boot up then take a backup
1. After the backup task is done, delete the VM
1. Restore a new VM with the same name using the newly created backup
1. After the restore task is completed, check its YAML for the label name, it should still be `harvesterhci.io/vmName`

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -55,7 +55,7 @@ const (
 	lastRestoreAnnotation = "restore.harvesterhci.io/last-restore-uid"
 
 	vmCreatorLabel = "harvesterhci.io/creator"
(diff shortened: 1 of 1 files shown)


bk201 (approved):
LGTM!

FrankYang0529 (approved):
LGTM. Thanks.

Issue #2259, opened 2022-05-11
[FEATURE] Support multiple VLAN physical interfaces


janeczku:
**Is your feature request related to a problem? Please describe.**

Use case: VM workloads must be able to connect to VLANs that are switched on two different physical switches. Example includes a connection to an out-of-band (OOB) management network that is typically hosted on a dedicated physical switch isolated from the  TOR or leaf switch.

**Expected implementation**

Harvester allows user to define VLAN networks on more than just 1 network interface.

w13915984028:
Supporting multi-networks (different NICs are binded to different networks) is planned in:
https://github.com/harvester/harvester/issues/2236

TachunLin:
Verified **fixed** on `v1.1.0-rc3`. Close this issue.

### Result
* Can create different vlan on each cluster network
  ![image](https://user-images.githubusercontent.com/29251855/196589014-f00bc352-4b0c-45ee-9a20-c91d7ac54844.png)

* Can create VM using vlan id `91` and retrieve IP address correctly
![image](https://user-images.githubusercontent.com/29251855/196592572-5f9f3b79-9c41-4714-9c88-821ac41f923d.png)

* Can create VM using vlan id `92` and retrieve IP address correctly
![image](https://user-images.githubusercontent.com/29251855/196592853-1d092799-23ae-411c-97ee-163c88195c14.png)

* Can create VM using vlan id `1` and retrieve IP address correctly
![image](https://user-images.githubusercontent.com/29251855/196593224-bf7d2e6a-4dbf-46ae-ab10-4168e8de1014.png)

### Test Information
* Test Environment: 3 nodes Harvester bare machines
* Harvester version: `v1.1.0-rc3`

### Verify Steps
1. Create cluster network `cn1`
  ![image](https://user-images.githubusercontent.com/29251855/196580297-57541544-48f5-4492-b3e9-a3450697f490.png)

2. Create a vlanconfig `config-n1` on `cn1` which applied to node 1 only
  ![image](https://user-images.githubusercontent.com/29251855/196580491-0572c539-5828-4f2e-a0a6-59b40fcc549b.png)

3. Select an available NIC on the Uplink 
  ![image](https://user-images.githubusercontent.com/29251855/196580574-d38d59de-251c-4cf8-885d-655b76a78659.png)

4. Create a vlan, the cluster network `cn1` vlanconfig and provide valid vlan id `91`
[...]

PR #2687, opened 2022-08-23
Move the network webhook to harvester network controller

yaocw2020:
It's more prefer to move the network webhook to harvester network controller.

Related issue: https://github.com/harvester/harvester/issues/2259
Dependency PR: https://github.com/harvester/network-controller-harvester/pull/59


tjjh89017 (approved):
LGTM

guangbochen (commented):
need to remove NAD testings cases at https://github.com/harvester/harvester/blob/master/tests/integration/api/network_apis_test.go, also consider migrating those tests to the harvester-network-controller repo as this it is also important to guard the tests over there.

yaocw2020:
> need to remove NAD testings cases at https://github.com/harvester/harvester/blob/master/tests/integration/api/network_apis_test.go, also consider migrating those tests to the harvester-network-controller repo as this it is also important to guard the tests over there.

This case will pass after bumping the new harvester network controller chart.

guangbochen:
@yaocw2020 network-controller is disabled by default in the integration test https://github.com/harvester/harvester/blob/master/tests/integration/runtime/construct.go#L50, I can't recall exactly why, but maybe due to the limitation of kind cluster.

yaocw2020:
> @yaocw2020 network-controller is disabled by default in the integration test https://github.com/harvester/harvester/blob/master/tests/integration/runtime/construct.go#L50, I can't recall exactly why, but maybe due to the limitation of kind cluster.

The new version VLAN implementation maybe not be limited by the kind cluster anymore. Let me delete the test temporarily and check it after bumping the latest harvester network controller chart.

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM, thanks!

yaocw2020:
Let's consider bringing the network integration test back in rc2.

Issue #2650, opened 2022-08-10
[BUG]   cluster-registration-url cannot be set to null. 


WuJun2016:
**Describe the bug**

1. cluster-registration-url cannot be set to null. 
2. Import the harvester in the rancher cluster, and after the harvester is successfully imported, go to another rancher cluster and import the harvester cluster again, at this time the imported harvester cluster is always in the pending state

**To Reproduce**
case 1:
1. Create a harvester cluster, set cluster-registration-url to any value， save
2.  Then reset the cluster-registration-url to empty
![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/75d57070-b967-4e4f-a16f-d4f2f24d1d61)
3. It does not make sense to display an error message when the value is empty, because this is the default value

case 2:
1. Import harvester in the rancher cluster and wait for the import to succeed.
2. Import the previous harvester cluster again in another rancher cluster，  （The cluster will always be in pending and no error message will be displayed）

**Expected behavior**

case 2:  When importing harvester in multiple clusters, the last imported harvester cluster should import successfully and the previous imported harvester clusters should become unavailable

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
3

irishgordo:
Validated that while #2510 passes, the ability to set the cluster-registration-url to null is still exists as an issue in Harvester Master iso: 9631b136
and v2.6.7-rc-9 Rancher.
![Screenshot from 2022-08-17 11-25-21](https://user-images.githubusercontent.com/5370752/185214980-7a2faeaf-70ca-4610-a0d4-96be03dad17e.png)

noahgildersleeve:
@FrankYang0529 How long should it take for the cluster-registration to clear after deleting it in Harvester settings? Is it on some sort of a polling timer?

noahgildersleeve:
I tested this in `master-eb3e1f3e-head` with a 3 node ipxe setup from the vagrant ipxe-examples setup. The Rancher setup was on a single node Docker instance. I tested it with `v2.6.7` and `v2.6.8`.

## Validation Steps
### case 1

1. Create a harvester cluster and rancher v2.6.8.
2. Import the harvester cluster to rancher.
3. After successful import, reset the cluster-registration-url to empty.

### Expected Results

- You should get an error in the virtualization management section of Rancher on the connection

Failed

### case 2

[...]

PR #2690, opened 2022-08-24
feat(cluster-retistration-url): support set to default

FrankYang0529:
**Problem:**
The setting `cluster-registration-url` cannot be set to empty.

**Solution:**
Check whether the URL is empty in the controller.

**Related Issue:**
https://github.com/harvester/harvester/issues/2650

**Test plan:**
case 1:
1. Create a harvester cluster and rancher v2.6.7.
2. Import the harvester cluster to rancher.
3.  After successful import, reset the `cluster-registration-url` to empty.

case 2:
1. Create a harvester cluster and rancher v2.6.7.
2. Import the harvester cluster to rancher.
3. After successful import, create another rancher v2.6.7.
4. Import the harvester cluster to the new rancher. Import should be successful.

### DIFF
--- a/pkg/controller/master/setting/cluster_registration_url.go
+++ b/pkg/controller/master/setting/cluster_registration_url.go
@@ -9,6 +9,8 @@ import (
 	"net/http"
 
 	"github.com/rancher/wrangler/pkg/objectset"
+	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
@@ -20,6 +22,9 @@ import (
 // registerCluster imports Harvester to Rancher by applying manifests from the registration URL.
 func (h *Handler) registerCluster(setting *harvesterv1.Setting) error {
 	url := setting.Value
+	if url == "" {
+		return h.cleanupClusterAgent()
+	}
 	resp, err := h.httpClient.Get(url)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/controller/master/setting/cluster_registration_url.go
```diff
@@ -65,3 +71,16 @@ func (h *Handler) registerCluster(setting *harvesterv1.Setting) error {
...
 }
+
+func (h *Handler) cleanupClusterAgent() {
+	// cleanup rancher related resources
+	// ref: https://rancher.com/docs/rancher/v2.6/en/cluster-provisioning/registered-clusters/#registering-a-cluster
+	// ref: https://rancher.com/docs/rancher/v2.6/en/cluster-provisioning/rke-clusters/rancher-agents/
+	if _, err := h.deploymentCache.Get("cattle-system", "cattle-cluster-agent"); err == nil {
+		if err = h.deployments.Delete("cattle-system", "cattle-cluster-agent", nil); err != nil {
+			logrus.Errorf("Can't delete cattle-system/cattle-cluster-agent deployment, err: %+v", err)
+		}
```
guangbochen: should we return and do reconciliation instead?

FrankYang0529: Yeah, after thinking again, it's better to do best effort to clean up `cattle-system/cattle-cluster-agent` deployment if we can find it. Updated it. Thank you.

pkg/controller/master/setting/handler.go
```diff
@@ -104,7 +104,11 @@ func (h *Handler) setConfiguredCondition(settingCopy *harvesterv1.Setting, err e
 			return err
 		}
 	} else if err == nil {
-		harvesterv1.SettingConfigured.True(settingCopy)
+		if settingCopy.Value == "" {
+			harvesterv1.SettingConfigured.False(settingCopy)
```
bk201, line 108: @guangbochen, any takes on making the `configured` condition false if the value is empty?

guangbochen: looks fine to me since empty value can be treated as not configured.

bk201 (approved):
LGTM, thanks!

guangbochen (approved):
LGTM, thanks.

Issue #2631, opened 2022-08-04
[FEATURE] Add backup-taget connection status


lanfon72:
### The problem
When the **backup-target** disconnected, User can still restore VM and they will not aware the backup-target is not connectable.

### The solution
Update **Backups** page to display some message that the **backup-target** is not connectable.

### The Alternatives
no idea

### Additional Context
this will improve the case of #2560

fanlix:
maybe a button to test connection

FrankYang0529:
We can add a handler here https://github.com/harvester/harvester/blob/master/pkg/server/router.go.

For example, `m.Path("/v1/harvester/backup-target/{backupTargetName}/healthy").Methods("GET").Handler(backupTargetHandler)`. Currently, `backupTargetName` can only be `default`, but let's leave flexibility for further supporting multiple backup target.

In `backupTargetHandler`, we can setup [backupstore driver](https://github.com/longhorn/backupstore/blob/master/driver.go). For NFS driver, it try to run [List](https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/nfs/nfs.go#L74) in `initFunc`. We may need enhancement in S3 [initFunc](https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/s3/s3.go#L38-L87) to test the connection.

WuJun2016:
UI change:
1. The health check was successful 
<img width="1727" alt="image" src="https://user-images.githubusercontent.com/24985926/190136751-96d11688-ed35-45d7-bf95-c84239219d97.png">

2. The health check was unsuccessful 
<img width="1724" alt="image" src="https://user-images.githubusercontent.com/24985926/190136824-b81d8b8b-a310-40b4-b96f-ea334eca4eaa.png">

3.  If the backup is abnormal, an error message will be displayed when the user goes to the VM Backup page. If the backup is normal, it will not be displayed
<img width="1727" alt="image" src="https://user-images.githubusercontent.com/24985926/190137118-3c55eaba-c179-4f6e-a677-ab2f436ba5ad.png">

lanfon72:
Verified this feature has been implemented.

![image](https://user-images.githubusercontent.com/5169694/190369936-c07b0a5f-8685-4813-8108-1032caf09183.png)


Test Information
----
* Environment: **qemu/KVM 2 nodes**
* Harvester Version: **master-032742f0-head**
* **ui-source** Option: **Auto**

[...]

PR #2698, opened 2022-08-25
feat: add backup-target healthy check api

FrankYang0529:
**Problem:**
UI can't know whether the backup target connection is healthy.

**Solution:**
Add an API for it.

**Related Issue:**
https://github.com/harvester/harvester/issues/2631

**Test plan:**

Case 1: empty backup-target
* Create a new harvester cluster and a rancher server.
* Import harvester to rancher.
* Query `https://<ip>:<port>/k8s/clusters/<cluster-id>/v1/harvester/backuptarget/healthz`. The server should return a message "backup-target setting is not set" with HTTP status 400.

Case 2: happy path
* Set up a NFS server. `kubectl apply -f https://raw.githubusercontent.com/longhorn/longhorn/v1.3.0/deploy/backupstores/nfs-backupstore.yaml`
* Set backup-target to NFS type with `nfs://longhorn-test-nfs-svc.default:/opt/backupstore` endpoint.
* Query `https://<ip>:<port>/k8s/clusters/<cluster-id>/v1/harvester/backuptarget/healthz`. The server should return HTTP status 200.

Case 3: set backup-target to default
* Set backup-target as default value. The value will become `{type: "", ...}`.
* Query `https://<ip>:<port>/k8s/clusters/<cluster-id>/v1/harvester/backuptarget/healthz`. The server should return a message "backup-target setting is not set" with HTTP status 400.

Case 4: remove NFS server.
* Set backup-target to NFS type with `nfs://longhorn-test-nfs-svc.default:/opt/backupstore` endpoint.
* Remove NFS server. `kubectl delete -f https://raw.githubusercontent.com/longhorn/longhorn/v1.3.0/deploy/backupstores/nfs-backupstore.yaml`
* Query `https://<ip>:<port>/k8s/clusters/<cluster-id>/v1/harvester/backuptarget/healthz`. The server should return a message like "can't connect to backup-target ..." with HTTP status 503.

### DIFF
--- a/pkg/api/backuptarget/healthy_handler.go
+++ b/pkg/api/backuptarget/healthy_handler.go
@@ -1,0 +1,72 @@
+package backuptarget
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/longhorn/backupstore"
+	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+
+	"github.com/harvester/harvester/pkg/config"
+	"github.com/harvester/harvester/pkg/controller/master/backup"
+	"github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/settings"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+type HealthyHandler struct {
+	context      context.Context
+	settingCache v1beta1.SettingCache
+	secretCache  ctlcorev1.SecretCache
+}
+
+func NewHealthyHandler(scaled *config.Scaled) *HealthyHandler {
+	return &HealthyHandler{
+		context:      scaled.Ctx,
+		settingCache: scaled.HarvesterFactory.Harvesterhci().V1beta1().Setting().Cache(),
+		secretCache:  scaled.CoreFactory.Core().V1().Secret().Cache(),
+	}
+}
+
+func (h *HealthyHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	backupTargetSetting, err := h.settingCache.Get(settings.BackupTargetSettingName)
+	if err != nil {
+		util.ResponseError(rw, http.StatusInternalServerError, fmt.Errorf("can't get %s setting, error: %w", settings.BackupTargetSettingName, err))
+		return
+	}
+	if backupTargetSetting.Value == "" {
+		util.ResponseError(rw, http.StatusBadRequest, fmt.Errorf("%s setting is not set", settings.BackupTargetSettingName))
+		return
+	}
+
+	target, err := settings.DecodeBackupTarget(backupTargetSetting.Value)
+	if err != nil {
+		util.ResponseError(rw, http.StatusInternalServerError, fmt.Errorf("can't decode %s setting %s, error: %w", settings.BackupTargetSettingName, backupTargetSetting.Value, err))
+		return
+	}
+	if target.IsDefaultBackupTarget() {
+		util.ResponseError(rw, http.StatusBadRequest, fmt.Errorf("can't check the backup target healthy, %s setting is not set", settings.BackupTargetSettingName))
+		return
+	}
+
+	if target.Type == settings.S3BackupType {
+		secret, err := h.secretCache.Get(util.LonghornSystemNamespaceName, util.BackupTargetSecretName)
+		if err != nil {
+			util.ResponseError(rw, http.StatusInternalServerError, fmt.Errorf("can't get backup target secret: %s/%s, error: %w", util.LonghornSystemNamespaceName, util.BackupTargetSecretName, err))
+			return
+		}
+		os.Setenv(backup.AWSAccessKey, string(secret.Data[backup.AWSAccessKey]))
+		os.Setenv(backup.AWSSecretKey, string(secret.Data[backup.AWSSecretKey]))
+		os.Setenv(backup.AWSEndpoints, string(secret.Data[backup.AWSEndpoints]))
+		os.Setenv(backup.AWSCERT, string(secret.Data[backup.AWSCERT]))
+	}
+
+	_, err = backupstore.GetBackupStoreDriver(backup.ConstructEndpoint(target))
+	if err != nil {
+		util.ResponseError(rw, http.StatusServiceUnavailable, fmt.Errorf("can't connect to backup target %+v, error: %w", target, err))
+		return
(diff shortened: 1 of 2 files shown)


pkg/api/backuptarget/healthy_handler.go
```diff
@@ -0,0 +1,80 @@
...
+		secretCache:  scaled.CoreFactory.Core().V1().Secret().Cache(),
+	}
+}
+
+func (h *HealthyHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	backupTargetName := mux.Vars(r)["backupTargetName"]
+	if backupTargetName != "default" {
+		util.ResponseError(rw, http.StatusBadRequest, fmt.Errorf("currently, only support default backup target"))
+		return
+	}
```
guangbochen: do we need the backup target name since we only got one backup setting?

pkg/api/backuptarget/healthy_handler.go
```diff
@@ -0,0 +1,80 @@
...
+
+	target, err := settings.DecodeBackupTarget(backupTargetSetting.Value)
+	if err != nil {
+		util.ResponseError(rw, http.StatusInternalServerError, errors.Wrapf(err, "can't decode %s setting %s", settings.BackupTargetSettingName, backupTargetSetting.Value))
+		return
+	}
+	if target.IsDefaultBackupTarget() {
+		util.ResponseError(rw, http.StatusBadRequest, fmt.Errorf("%s setting is not set", settings.BackupTargetSettingName))
```
guangbochen:
```suggestion
		util.ResponseError(rw, http.StatusBadRequest, fmt.Errorf("can't check the backup target healthy, %s setting is not set", settings.BackupTargetSettingName))
```

pkg/api/backuptarget/healthy_handler.go
```diff
@@ -0,0 +1,80 @@
...
+		return
+	}
+
+	if target.Type == settings.S3BackupType {
+		secret, err := h.secretCache.Get(util.LonghornSystemNamespaceName, util.BackupTargetSecretName)
+		if err != nil {
+			util.ResponseError(rw, http.StatusInternalServerError, errors.Wrapf(err, "can't get %s/%s secret", util.LonghornSystemNamespaceName, util.BackupTargetSecretName))
```
guangbochen:
```suggestion
			util.ResponseError(rw, http.StatusInternalServerError, errors.Wrapf(err, "can't get backup target secret:%s/%s, error: %s", util.LonghornSystemNamespaceName, util.BackupTargetSecretName, err.Error()))
```

pkg/api/backuptarget/healthy_handler.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		os.Setenv(backup.AWSAccessKey, string(secret.Data[backup.AWSAccessKey]))
+		os.Setenv(backup.AWSSecretKey, string(secret.Data[backup.AWSSecretKey]))
+		os.Setenv(backup.AWSEndpoints, string(secret.Data[backup.AWSEndpoints]))
+		os.Setenv(backup.AWSCERT, string(secret.Data[backup.AWSCERT]))
+	}
+
+	_, err = backupstore.GetBackupStoreDriver(backup.ConstructEndpoint(target))
+	if err != nil {
+		util.ResponseError(rw, http.StatusServiceUnavailable, errors.Wrapf(err, "can't connect to backup target %+v", target))
```
guangbochen: always return the error message, don't swallow it.

guangbochen:
btw, would be nice to update the issue's PR checklist before moving it to the review pipeline, thanks.

pkg/api/backuptarget/healthy_handler.go
FrankYang0529: Removed it. I think we can add this after we support multiple backup-target. Thank you.

pkg/api/backuptarget/healthy_handler.go
FrankYang0529: Updated it. Thank you.

pkg/api/backuptarget/healthy_handler.go
FrankYang0529: Updated it. Thank you.

pkg/api/backuptarget/healthy_handler.go
FrankYang0529: After reading this thread, I change to use `fmt.Errorf("... %w", err)`. Thank you.

https://stackoverflow.com/questions/61933650/whats-the-difference-between-errors-wrapf-errors-errorf-and-fmt-errorf

pkg/server/router.go
```diff
@@ -48,6 +49,9 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
 
 	sbDownloadHandler := supportbundle.NewDownloadHandler(r.scaled, r.options.Namespace)
 	m.Path("/v1/harvester/supportbundles/{bundleName}/download").Methods("GET").Handler(sbDownloadHandler)
+
+	btHealthyHandler := backuptarget.NewHealthyHandler(r.scaled)
+	m.Path("/v1/harvester/backuptarget/healthy").Methods("GET").Handler(btHealthyHandler)
```
bk201: nit: how about just `health`? noun seems to be more compliant with REST.

FrankYang0529: How about `healthz`?
ref: https://kubernetes.io/docs/reference/using-api/health-checks/

bk201: Good idea!

FrankYang0529: Updated it. Thank you.

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM, @WuJun2016 Could you take a look if the API look good? Thanks

WuJun2016:
> LGTM, @WuJun2016 Could you take a look if the API look good? Thanks

I have no problem on my side, already discussed with yuhang.

bk201:
Thanks @WuJun2016

WuJun2016:
@bk201  @FrankYang0529   For this function, I think the frontend should not need to pay attention to the api, the frontend should get the status value from the `setting/barkup-target's` status,  right ?

FrankYang0529:
> @bk201 @FrankYang0529 For this function, I think the frontend should not need to pay attention to the api, the frontend should get the status value from the `setting/barkup-target's` status, right ?

Sorry, I think frontend needs to rely on API for this situation because backend doesn't watch on the connection between server and NFS/S3.

Issue #2361, opened 2022-06-08
[BUG] Increase resource for POD grafana


w13915984028:
**Describe the bug**

In a single-node cluster, with one guest VM, the Grafana POD's memory usage is close to limit. It is necessary to increase the limits value.
![image](https://user-images.githubusercontent.com/31133476/172632067-99001bc0-ef53-43ca-a8fe-ad8375d94320.png)

```
                "resources": {
                    "limits": {
                        "cpu": "200m",
                        "memory": "200Mi"
                    },
                    "requests": {
                        "cpu": "100m",
                        "memory": "100Mi"
                    }
                },
```

**To Reproduce**
1. Install Harvester with master-head
2. Create a few VMs
3. Check the memory usage of POD grafana from Harvester metrics.

**Expected behavior**

Has a proper memory limit for POD grafana.

**Support bundle**

**Environment**
 - Harvester ISO version:  V1.0.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM based Cluster

**Additional context**

DaiYuzeng:
@w13915984028 Hi! I realized there is the `area/ui` label. 

Users cannot modify the configuration of Grafana on GUI now, they can just modify configurations below:
![image](https://user-images.githubusercontent.com/15041915/183551382-58899795-ac48-47f2-9d71-e3e195ffbe6a.png)

Should we display the resources configurations of Grafana on GUI?

w13915984028:
@DaiYuzeng  I would like to suggest that we add similar resource configuration for Grafana POD in Harvester GUI.
This issue is in lower priority than others.

noahgildersleeve:
Tested in master-4ea12f4f-head. It looks like the limit was increased to the following
![Screenshot_20220831_130009](https://user-images.githubusercontent.com/83787952/187771896-255f42a5-867e-4aa5-9460-038816515c27.png)
This was what I saw with `describe pod rancher-monitoring-grafana-647fd577f4-qfks2`. The memory usage seemed better after creating a few VMs.
![Screenshot_20220831_125525](https://user-images.githubusercontent.com/83787952/187772092-b2459361-44c1-466a-b2e4-fb9dd7110faa.png)
Closing as fixed.

Issue #2282, opened 2022-05-17
[BUG] rancher-monitoring-grafana is too small and it keeps growing


abonillabeeche:
**Describe the bug**
The rancher-monitoring-grafana volume keeps growing but it's size should remain the same. Since the volume is low in size, it eventually grows more than it should.

**To Reproduce**
1. Turn Off Harvester servers.
2. Turn On Harvester servers.
3. Eventually the volume grows over the assigned 10Mi

**Expected behavior**
The volume should be created with a size that doesn't cause trouble should be encountered and the volume should not be growing so drastically.

**Environment:**
 - Harvester ISO version: 1.0.1 and 1.0.2
 - Underlying Infrastructure: R620

**Additional context**
The only action I do that is unusual, is that I power off/on the servers frequently to preserve power.

w13915984028:
Under investigation.
```
Option A:
 keep pvc, 
 set default size to 10Gb according to grafana's recommended default value,
 expand pvc when upgrading
 
option B:
  remove PVC for grafana, by default, it is not needed

```

Option B also patially improve #2013

TachunLin:
Verified **fixed** on `master-3bcba9d3-head` (09/13). Close this issue.

### Result
* The `rancher-monitoring-grafana` default allocated with `2Gi` and Actual usage `108 Mi` after running after 24 hours
  ![image](https://user-images.githubusercontent.com/29251855/191000121-9c3c640e-7d7f-4d1b-84f6-39745abca0ce.png)

* Turn off then turn on the specific vip harvester node machine, the The `rancher-monitoring-grafana` keep stable in `107 Mi` after turning on 60 minutes
  ![image](https://user-images.githubusercontent.com/29251855/191011632-06f03787-6bac-4bff-b964-787b414407ab.png)

* Turn off then turn on all four harvester node machines, the The `rancher-monitoring-grafana` keep stable in `107 Mi` after turning on 60 minutes
![image](https://user-images.githubusercontent.com/29251855/191023195-661275d3-1edb-4739-b585-cca23d17a3d9.png)


### Test Information
* Test Environment: 3 nodes harvester on provo bare machine
* Harvester version: `master-3bcba9d3-head` (09/13)

### Verify Steps
1. Harvester cluster running after 24 hours
2. Access Harvester Longhorn dashboard via https://<vip>/dashboard/c/local/longhorn
3. Open the Longhorn UI
4. Open the volume page
5. Check the `rancher-monitoring-grafana` size and usage 
[...]

PR #2711, opened 2022-08-29
Upgrade patch of rancher-monitoring

w13915984028:
https://github.com/harvester/harvester/issues/2361
https://github.com/harvester/harvester-installer/pull/287   fix2361 increase grafana resource limit
<del>
https://github.com/harvester/harvester/issues/2282
https://github.com/harvester/harvester-installer/pull/327  adjust grafana POD PVC size to 2GiB
</del>
Signed-off-by: Jian Wang <w13915984028@gmail.com>

**Problem:**

Resources change of rancher-monitoring grafana should be applied to old version when upgrading.

**Solution:**

Add upgrade patch for rancher-monitoring POD

**Related Issue:**
Per title

**Test plan:**

1. The POD resources chagne will be effective after upgrading.
https://github.com/harvester/harvester-installer/pull/287   fix2361 increase grafana resource limit

<del>

2. The grafana PVC size change will not be effective right after upgrading.
https://github.com/harvester/harvester-installer/pull/327  adjust grafana POD PVC size to 2GiB

This patch will make sure the `PVC size` is 2Gi after upgrading.
How to patch existing PVC, needs more investigation. 
(more scripts like disable grafana, delete PVC, re-enable grafana;  or accordging to PVC reset steps as per LH's document;  both of them are more than quick inheritage from current upgrading framework @bk201 @guangbochen 
)

</del>

The patch to PR:https://github.com/harvester/harvester-installer/pull/327 is removed.

Will update `Harvester doc` and `release-notes` about how to extend running PVC size, the limitation and workaround.
https://github.com/harvester/harvester/issues/2724

### DIFF
--- a/package/upgrade/migrations/managed_charts/v1.0.3.sh
+++ b/package/upgrade/migrations/managed_charts/v1.0.3.sh
@@ -1,0 +1,16 @@
+#!/bin/bash -ex
+
+CHART_NAME=$1
+CHART_MANIFEST=$2
+
+patch_grafana_resources()
+{
+  # Increase grafana pod limit and request (https://github.com/harvester/harvester-installer/pull/287)
+  yq e '.spec.values.grafana.resources = {"limits": {"cpu": "200m", "memory": "500Mi"}, "requests": {"cpu": "100m", "memory": "200Mi"}}' $CHART_MANIFEST -i
+}
+
+case $CHART_NAME in
+  rancher-monitoring)
+    patch_grafana_resources
+    ;;
+esac


package/upgrade/migrations/managed_charts/v1.1.0.sh
```diff
@@ -0,0 +1,19 @@
...
+CHART_NAME=$1
+CHART_MANIFEST=$2
+
+patch_grafana_resources()
+{
+  # Increase grafana pod limit and request (https://github.com/harvester/harvester-installer/pull/287)
+  yq e '.spec.values.grafana.resources = {"limits": {"cpu": "200m", "memory": "500Mi"}, "requests": {"cpu": "100m", "memory": "200Mi"}}' $CHART_MANIFEST -i
+
+  # Increase grafana PVC size to 2Gi (https://github.com/harvester/harvester-installer/pull/327)
+  yq e '.spec.values.grafana.persistence.size = "2Gi"' $CHART_MANIFEST -i
```
guangbochen, line 12: only patching the PVC size values without actually changing the real volume size is not valuable and may make it ambiguous for the user, would prefer to leave this unchanged and use both docs and release notes to provide a manual workaround solution, thanks.

guangbochen: only patching the PVC size values without actually changing the real volume size is not valuable and may make it ambiguous for the user, would prefer to leave this unchanged and use both docs and release notes to provide a manual workaround solution, thanks.

w13915984028: Sound reasonable. Extending running PVC size needs a guide in `Harvester Doc`. Thanks.

w13915984028: Update doc and release notes is tracked in: https://github.com/harvester/harvester/issues/2724

bk201 (changes requested):
The file name should be named as the previous version. (Please check https://github.com/harvester/harvester/blob/master/package/upgrade/migrations/README.md). So here it should be `v1.0.3`.

guangbochen (approved):
LGTM, thanks.

bk201:
LGTM

bk201 (approved):

Issue #1055, opened 2021-07-29
[FEATURE]Dedicated storage network


yasker:
We would like to have a dedicated storage network interface that can be specified (either as single nic or bonded NICs) to separate the storage traffic from the overlay traffic.

Depends on https://github.com/longhorn/longhorn/issues/2285 and #1048

yasker:
Set the milestone to v0.3.0 for now. If https://github.com/longhorn/longhorn/issues/2285 cannot be completed for v1.2.0, then we need to push it out.

tjjh89017:
HEP PR: https://github.com/harvester/harvester/pull/2685
PoC PR: https://github.com/harvester/harvester/pull/2706

TODO:
whereabouts charts
webhook to verify NAD and VM status

tjjh89017:
`harvester-harvester-vm-import-controller` is also using Volume.
Need to shutdown it also.
Need investigation

tjjh89017:
Testing UI: https://yuzeng-dev.oss-cn-beijing.aliyuncs.com/dashboard/storage-network-new/index.html

## Test Step

### Test1

1. Shutdown all VMs
2. setup vlan config & cluster network
3. set settings named "storage-network" with value below in Harvester settings
```json
{
    "vlan": 1,
    "clusterNetwork": "<your cluster network>",
    "range": "<IP range>"
}
```
6. wait and check Longhorn settings "storage-network" is "storagenetwork-XXXXXXXXXXX".
7. check Longhorn `instance-manager-e-XXXX` has an interface named "lhnet1"

### Test2

1. do Test1 and wait
2. set value=""
3. wait and check Longhorn "storage-network" is empty
4. check Longhoorn `instance-manager-e-XXXX` didn't has an interface name "lhnet1"

### Test 3
1. Start a VM
2. modify storage-network setting with any value.
3. webhook should deny change

lanfon72:
@tjjh89017 the link of **Testing UI** seems not work, could you update it ?

tjjh89017:
@WuJun2016 Did you remove the storage network testing UI from aliyuncs?
Thanks a lot.

WuJun2016:
@lanfon72  You can take the latest master address to test： https://releases.rancher.com/harvester-ui/dashboard/latest/index.html

lanfon72:
Verified this feature not fully implemented.
As it is a dedicated network, we should have a fallback option when the node not being configured in the `clusterNetwork`.
[...]

PR #2716, opened 2022-08-31
storage: add whereabouts cni charts for storage network

tjjh89017:

**Problem:**

need whereabouts CNI charts

**Solution:**

add whereabouts CNI charts into dependency_charts and charts

**Related Issue:**
https://github.com/harvester/harvester/issues/1055

**Test plan:**

Install new version harvester and check daemonset whereabouts is existing


bk201:
@tjjh89017 This looks like an upstream chart. Could you document the reason not to use the upstream repository?
If you need the chart maintained in Harvester, the better place is to add in [harvester/charts](https://github.com/harvester/charts) first and publish.

tjjh89017:
> This looks like an upstream chart. Could you document the reason not to use the upstream repository?

whereabouts chart in upstream is out of date. We need to maintain a new version for it.

> If you need the chart maintained in Harvester, the better place is to add in [harvester/charts](https://github.com/harvester/charts) first and publish.

ok, I will send PR to that repo

FrankYang0529:
@bk201 Looks like [harvester/charts](https://github.com/harvester/charts) is for some components we build. If it's more like upstream component, will it be better to add it like kubevirt in dependency_charts?

bk201:
@FrankYang0529 Indeed, although those two charts are created by us too. @guangbochen, do you have any suggestions?

guangbochen:
I think it is fine to keep this as a dependency chart in the `harvester/harvester` repo since we will be the only ones to use it.
Note: would be good to link the upstream-related repo and leave a note of what this chart has changed, thanks.

bk201:
Thanks, @guangbochen
@tjjh89017, let's keep it in the harvester, and if we need to test it individually, we can move it out. Sorry for the hassle.

deploy/charts/harvester/dependency_charts/whereabouts/Chart.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v2
+name: whereabouts
```
guangbochen, line 2: probably a name with `whereabouts-cni` is more preferred. Additionally, can we please add a README page for this chart to describe which repo it was referring to and why we need this dependency-chart, thanks.

deploy/charts/harvester/dependency_charts/whereabouts/values.yaml
```diff
@@ -0,0 +1,50 @@
+# Default values for whereabouts.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: ghcr.io/k8snetworkplumbingwg/whereabouts
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "v0.5.4-amd64"
```
guangbochen, line 9: need to add the image values to the https://github.com/harvester/harvester/blob/master/deploy/charts/harvester/values.yaml so it will be packaged for the air-gap env.

deploy/charts/harvester/dependency_charts/whereabouts/Chart.yaml
tjjh89017: i will consider to keep it as upstream.
So I will add readme for it.

deploy/charts/harvester/dependency_charts/whereabouts/values.yaml
tjjh89017: added.
thanks a lot

bk201 (approved):
LGTM

bk201:
@tjjh89017 The namespace is `kube-system`; should we move to `harvester-system`? cc @guangbochen

guangbochen (approved):
LGTM, thanks. Seems the chart files are identical to the upstream whereabouts manifests so probably keeping the original one is preferred for now.

FrankYang0529 (approved):
Test and it looks good to me. Thanks.

Issue #2654, opened 2022-08-11
[FEATURE] Harvester UI plugin


WuJun2016:
**Is your feature request related to a problem? Please describe.**

https://github.com/rancher/dashboard/issues/4109
 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

TachunLin:
#### Test plan 
1. Regression test of Rancher `v2.6.9-rc2` integrate with Harvester `v1.1.0-rc1`
  - Access Harvester cluster from Rancher 
  - RBAC related Harvester permission on Harvester 
 
2. Regression test of Rancher `v2.6.9-rc2` integrate with Harvester `v1.0.3`
  -  Access Harvester cluster from Rancher 
  - RBAC related Harvester permission on Harvester 

Test plan stated in https://harvester.github.io/tests/manual/harvester-rancher/

lanfon72:
Both test plan should be verified in Rancher `v2.6.9-rc2`, so this issue still pending on Rancher's docker image release

lanfon72:
Verified this feature has been implemented.

https://user-images.githubusercontent.com/5169694/192982019-616e6e23-281a-4e93-a018-f6fc3ed5840e.mp4


Test Information
----
* Environment: **baremetal DL360G9 4 nodes** and **qemu/KVM 2 nodes**
* Harvester Version: **master-7dd398f0-head** and **v1.0.3**
* **ui-source** Option: **External** and **Bundled**
* **Rancher** Version: **v2.6.9-rc2**
* **Rancher** _ui-offline-preferred_: **Dynamic**

### Verify Steps:
1. Install Rancher `v2.6.9-rc2`
2. Navigate to _Virtualization Management_, Import Existing Harvester
3. Harvester should imported successfully
4. Access Harvester by Rancher's _Virtualization Management_
5. Harvester's dashboard pages should display correctly
6. Create an image for VM then create a VM from it
7. Page's state (viewing from Rancher) should synced with Harvester Dashboard

PR #2720, opened 2022-09-01
Add ui info api

futuretea:

- pack ui plugin tar ball `https://releases.rancher.com/harvester-ui/plugin/harvester-latest.tar.gz` to harvester docker image
- new setting `ui-plugin-index` default to `https://releases.rancher.com/harvester-ui/plugin/harvester-latest/harvester-latest.umd.min.js`
- new api `/v1/harvester/ui-info` expose ui and plugin ui info
- new api `/v1/harvester/plugin-assets` serve bundled harvester ui plugin

**Related Issue:**
https://github.com/harvester/harvester/issues/2654

**Test plan:**

replace harvester image to `futuretea/harvester:ui-info-head`

1. view api `/v1/harvester/ui-info`
image

2. view api `/v1/harvester/plugin-assets/`
image

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -15,6 +15,8 @@ ENV HARVESTER_UI_PATH /usr/share/harvester/harvester
 # Please update the api-ui-version in pkg/settings/settings.go when updating the version here.
 ENV HARVESTER_API_UI_VERSION 1.1.9
 
+ENV HARVESTER_UI_PLUGIN_BUNDLED_VERSION latest
+
 ARG ARCH=amd64
 ARG VERSION=dev
 ENV HARVESTER_SERVER_VERSION ${VERSION}
@@ -32,6 +34,9 @@ RUN mkdir -p /usr/share/harvester/harvester && \
     mkdir -p /usr/share/harvester/harvester/api-ui && \
     cd /usr/share/harvester/harvester/api-ui && \
     curl -sL https://releases.rancher.com/api-ui/${HARVESTER_API_UI_VERSION}.tar.gz | tar xvzf - --strip-components=1 && \
+    mkdir -p /usr/share/harvester/harvester/v1/harvester/plugin-assets && \
+    cd /usr/share/harvester/harvester/v1/harvester/plugin-assets && \
+    curl -sL https://releases.rancher.com/harvester-ui/plugin/harvester-${HARVESTER_UI_PLUGIN_BUNDLED_VERSION}.tar.gz | tar xvzf - --strip-components=1 && \
     cd /var/lib/harvester/harvester
 
 COPY entrypoint.sh harvester /usr/bin/
--- a/pkg/api/uiinfo/ui_info_handler.go
+++ b/pkg/api/uiinfo/ui_info_handler.go
@@ -1,0 +1,38 @@
+package uiinfo
+
+import (
+	"net/http"
+	"os"
+
+	"github.com/harvester/harvester/pkg/config"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/settings"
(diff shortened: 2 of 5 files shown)


pkg/server/router.go
```diff
@@ -46,6 +47,10 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
 	kcGenerateHandler := kubeconfig.NewGenerateHandler(r.scaled, r.options)
 	m.Path("/v1/harvester/kubeconfig").Methods("POST").Handler(kcGenerateHandler)
 
+	uiInfoHandler := uiinfo.NewUIInfoHandler(r.scaled, r.options)
+	m.Path("/v1/harvester/uiinfo").Methods("GET").Handler(uiInfoHandler)
```
guangbochen: perfer to use `/s/uiinfo/ui-info/`

package/Dockerfile
```diff
@@ -32,6 +34,9 @@ RUN mkdir -p /usr/share/harvester/harvester && \
     mkdir -p /usr/share/harvester/harvester/api-ui && \
     cd /usr/share/harvester/harvester/api-ui && \
     curl -sL https://releases.rancher.com/api-ui/${HARVESTER_API_UI_VERSION}.tar.gz | tar xvzf - --strip-components=1 && \
+    mkdir -p /usr/share/harvester/harvester/v1/harvester/plugin-assets && \
+    cd /usr/share/harvester/harvester/v1/harvester/plugin-assets && \
+    curl -sL https://releases.rancher.com/harvester-ui/plugin/harvester-${HARVESTER_UI_PLUGIN_BUNDLED_VERSION}.tar.gz | tar xvzf - --strip-components=1 && \
```
guangbochen, line 39: @WuJun2016 do we need to keep the original UI assets, or only using the plugin UI would be enough?

pkg/api/uiinfo/ui_info_handler.go
```diff
@@ -0,0 +1,38 @@
...
+			uiSource = "external"
+		} else {
+			uiSource = "bundled"
+		}
+	}
+	util.ResponseOKWithBody(rw, map[string]string{
+		settings.UISourceSettingName:               uiSource,
+		settings.UIIndexSettingName:                settings.UIIndex.Get(),
+		settings.UIPluginIndexSettingName:          settings.UIPluginIndex.Get(),
+		settings.UIPluginBundledVersionSettingName: os.Getenv(settings.GetEnvKey(settings.UIPluginBundledVersionSettingName)),
```
guangbochen, line 36: wondering what this setting response is used for?

futuretea: cc @WuJun2016

pkg/server/router.go
futuretea: updated

pkg/api/uiinfo/ui_info_handler.go
WuJun2016: The front-end needs to know what version is saved inside plugin-assets.  I know that normally the version of the plugin is the same as the harvester version, so its value should be the same as the `server-version`, but I'm not sure if using the `server-version` value is always appropriate.    
So to make sure I can always get the right value, so I added a property to save the version of the plugin (this plugin version is determined by the name of the packaged file)

If you don't add this attribute, then you need to expose the `server-version` to ui-info.

WuJun2016: I tested it, the `server-version` user is able to change, which is actually unreasonable.   
if the user changes this value, the user will not be able to access the Harvester from multiple clusters

package/Dockerfile
WuJun2016: We still need the old address, which is for a single cluster UI, and the newly added address for multiple clusters

guangbochen (approved):
LGTM, thanks.

bk201 (approved):

Issue #2431, opened 2022-06-29
[FEATURE] Check version compatibility during an upgrade


bk201:
**Is your feature request related to a problem? Please describe.**

The harvester controller fetches upgradeable versions from the Harvester responder, but the user still can upgrade to an incompatible version by writing a [Version manifest](https://docs.harvesterhci.io/v1.0/upgrade/automatic/#prepare-an-air-gapped-upgrade).

**Describe the solution you'd like**

We should verify a running cluster can upgrade to a new ISO.

**Describe alternatives you've considered**

**Additional context**

TachunLin:
Verified on the latest v1.1.0 master with and without release tag. 

### Result
Verified can PASS the following scenarios

#### Test Plan 1: v1.0.2 upgrade to v1.1.0 **with** release tag 
* Upgrade from v1.0.2 upgrade to v1.1.0 with release tag `failed` on upgrade dashboard UI
  ![image](https://user-images.githubusercontent.com/29251855/193567678-7653aa7b-b93a-4cf2-98a8-37e4fcfd06f6.png)

* The upgrade pod display `Error` 
```
harvester102:/home/rancher # kubectl -n harvester-system get po -l harvesterhci.io/upgradeComponent=manifest
NAME                                       READY   STATUS   RESTARTS   AGE
hvst-upgrade-xtkwx-apply-manifests-4gw6h   0/1     Error    0          20m
hvst-upgrade-xtkwx-apply-manifests-55bll   0/1     Error    0          18m
hvst-upgrade-xtkwx-apply-manifests-f4j8r   0/1     Error    0          19m
hvst-upgrade-xtkwx-apply-manifests-f78gj   0/1     Error    0          20m
hvst-upgrade-xtkwx-apply-manifests-fwjx2   0/1     Error    0          20m
hvst-upgrade-xtkwx-apply-manifests-mjgkg   0/1     Error    0          10m
hvst-upgrade-xtkwx-apply-manifests-nsnzb   0/1     Error    0          15m

```

* The pod logs display `Current version is not supported. Abort the upgrade.` 
```
harvester102:~ # kubectl -n harvester-system logs hvst-upgrade-xtkwx-apply-manifests-fwjx2
harvester: v1.1.0
harvesterChart: 1.1.0
os: Harvester eeeb1be-dirty
kubernetes: v1.23.10+rke2r1
rancher: v2.6.4-harvester3
monitoringChart: 100.1.0+up19.0.3
[...]

PR #2730, opened 2022-09-05
feat(upgrade): block upgrades from unsupported versions

starbops:
**Problem:**

Users can upgrade from any version as long as they have the ISO image. They can host the image and create a Version resource that points the URL to the image location or even upload the image from the dashboard to directly trigger the upgrade. There is no version support matrix provided nor version checking logic implemented. The unsupported upgrade path might introduce unpredictable issues and burden the team with a heavy load.

**Solution:**

Users who upgrade from the version older than the specified minimum version required, which was retrieved from `harvester_release.yaml`, will be errored out in phase 2 right before the image preloading step of the entire upgrade flow.

The current strategy is we pick a version as the minimum required version to do the upgrade in each release, put that version string into a file, and package the info into the ISO image. While upgrading, a repo VM will be spun up and hosted `harvester-release.yaml` which contains the minimum upgradable version string. Then we put a version checking logic in the `upgrade_node.sh` script, right before the image preloading step. If the checking fails, the upgrade Job will fail.

Assumed v1.1.0 release:

- cur_ver=v1.0.0, min_upgradable_ver=v1.0.3 => check failed
- cur_ver=v1.0.3, min_upgradable_ver=v1.0.3 => check pass
- cur_ver=v1.0.4, min_upgradable_ver=v1.0.3 => check pass

**Related Issue:**

The original issue is #2431, and this PR also required PR harvester/harvester-installer#330 to work.

**Test plan:**


starbops:
Will there be any concern regarding breaking the upgrade brutally? cc @bk201

starbops:
TODOs:

1. Maintaining a full support matrix instead of just one minimum upgradable version string for further integrations (like docs and upgrade-responder)
2. Implementing the version checking logic inside the upgrade controller for double assurance

starbops:
The branch name should not be that long due to the Kubernetes label length restriction (<63 chars). Close this PR.

Issue #2741, opened 2022-09-07
[BUG] Upgrade: stuck at pre-drained state


bk201:
**Describe the bug**

We have spotted some situations instance-manager pod can't be drained out of a node:
- Longhorn is being upgraded, and there are 2 `instance-manager-r` pods on a node.
- A node is just drained and comes back; if we drain the next node, there is a high probability the next node is stuck.

**To Reproduce**
1. Upgrade to master ISO from v1.0.3. (which bumps Longhorn to 1.3.1).
2. Many users report upgrades being stuck in pre-drain problems.

**Expected behavior**

The upgrade should not be stuck.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.0.3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

bk201:
We can test this when other upgrade works are done.

lanfon72:
Tested:
- [x] single node cluster upgrade

noahgildersleeve:
Tested in 3 node ipxe-examples vagrant upgrade scenario with Harvester v1.0.3 -> v1.1.0-rc2. I did this with a Harvester v1.1.0-rc2 iso stored on the ipxe server.

It was stuck in pre-drained state for hours. 
![Screenshot_20221010_184101](https://user-images.githubusercontent.com/83787952/194978968-1c171f19-f07e-4254-9ee4-372715722c10.png)
![Screenshot_20221010_182239](https://user-images.githubusercontent.com/83787952/194978971-729bebf2-7dfc-497c-90af-0bd618b10b67.png)
![Screenshot_20221010_131925](https://user-images.githubusercontent.com/83787952/194978972-09272d6a-09c7-4268-9722-137189474058.png)



[supportbundle_63f35a4d-17ba-4f4c-bb61-9e58685a5305_2022-10-11T01-41-18Z.zip](https://github.com/harvester/harvester/files/9750298/supportbundle_63f35a4d-17ba-4f4c-bb61-9e58685a5305_2022-10-11T01-41-18Z.zip)
Here's the `version.yaml` I used. I used the steps for upgrading [here](https://docs.harvesterhci.io/v1.0/upgrade/automatic#prepare-an-air-gapped-upgrade)


```
apiVersion: harvesterhci.io/v1beta1
kind: Version
metadata:
  name: v1.1.0-rc2
  namespace: harvester-system
spec:
[...]

PR #2744, opened 2022-09-07
Upgrade: fix stuck in the pre-drained state

bk201:

**Problem:**
We have spotted some situations instance-manager pod can't be drained out of a node:

- Longhorn is being upgraded, and there are 2 instance-manager-r pods on a node.
- A node is just drained and comes back; if we drain the next node, there is a high probability the next node is stuck.

**Solution:**

- Wait Longhorn finish upgrading instance-manager-r pods
- Wait for replicas ready before draining a node

**Related Issue:**
https://github.com/harvester/harvester/issues/2741

**Test plan:**

Build and upgrade from v1.0.3.
- Test 1-node cluster upgrade. (changes in this PR not run )
- Test 2-node cluster upgrade.
- Test 3-node cluster upgrade.

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -139,6 +139,71 @@ wait_kubevirt()
   done
 }
 
+wait_longhorn_manager() {
+  echo "Waiting for longhorn-manager to be upgraded..."
+
+  lm_repo=$(kubectl get apps.catalog.cattle.io/harvester -n harvester-system -o json | jq -r .spec.chart.values.longhorn.image.longhorn.manager.repository)
+  lm_tag=$(kubectl get apps.catalog.cattle.io/harvester -n harvester-system -o json | jq -r .spec.chart.values.longhorn.image.longhorn.manager.tag)
+  lm_image="${lm_repo}:${lm_tag}"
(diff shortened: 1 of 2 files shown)


bk201:
@starbops @Vicente-Cheng Just found for 1-node and 2-node cluster, `instance-manager-r` count never become `1. I need to do more investigation.

bk201:
Updated the PR, do not wait for instance-manager-r for 1-node/2-node clusters. Degraded volumes can't be upgraded in this case. https://longhorn.io/docs/1.3.1/deploy/upgrade/auto-upgrade-engine/#2-the-behavior-of-longhorn-with-different-volume-conditions

Vicente-Cheng (approved):
LGTM, and tested it. After the upgrade, VM looks fine and the data is correct.
Thanks!

bk201:
@Vicente-Cheng @starbops I add one more case in [8138383](https://github.com/harvester/harvester/pull/2744/commits/813838318ed9ed601daf8ff4dd7627ed49280b77). it's possible for the user to configure 1x replica in the next version.

starbops (approved):
LGTM! Tested on 1/2/3/4-node upgrades.

Issue #2692, opened 2022-08-24
[FEATURE] Create a harvester-specific StorageClass for Longhorn


bk201:
Longhorn chart creates a `longhorn` StorageClass by default and we are using it with [some modification](https://github.com/harvester/harvester/blob/dc6a30894d63a07ba8b6db5433a7054f71445fde/deploy/charts/harvester/templates/longhorn-storageclass-override.yaml).
This might cause some issues and we should create another one (e.g., `longhorn-harvester`).

### ToDos
- [x] For new clusters, update Prometheus, grafana, alertmanager, logging to use the new storageclass by default.
- [x] storage class change for the alert manager in upgrade path: https://github.com/harvester/harvester/pull/2721/files#diff-bdbacb497dee5a09e5becbca17802c0d09476c955d098ebd1d20b2bb20d5207eR19

w13915984028:
@ibrokethecloud @bk201 @Vicente-Cheng 
Gaurav just talked about #2088 

A harvester specific default `storageclass` will leave longhorn default storageclass untouched. It is also used by `rancher-monitoring` etc. in bootstrap stage.


/go/src/github.com/harvester/harvester-installer/pkg/config/templates/rancherd-13-monitoring.yaml

```
...
          storageSpec:
            volumeClaimTemplate:
              spec:
                accessModes:
                - ReadWriteOnce
                resources:
                  requests:
                    storage: 50Gi
                storageClassName: longhorn  /////////////// replace here to the new one
                volumeMode: Filesystem
```

bk201:
@w13915984028 #2088 is listed under the blocking section (need zenhub to see).

guangbochen:
@Vicente-Cheng @bk201 we may need to consider to patch the original `longhorn` SC to none-default SC druing v1.1 upgrade, after adding the storage tiering support we will have a new page to display all of the storage classes, also to avoid having multiple default SC, @futuretea will help to add validation webhook to only allow one default SC per cluster, please let me know if there are any concerns about this requirement, thanks. cc @n313893254

w13915984028:
@guangbochen @bk201 @futuretea @Vicente-Cheng 

The issue https://github.com/harvester/harvester/issues/2088 has bacame a blocker with feature of logging and alert. I tried to add a `harvester-general-3-replicas` storageclass , the `rancher-monitoring` etc will use it. 

[...]

PR #2768, opened 2022-09-15
Create a harvester-specific StorageClass

Vicente-Cheng:

**Problem:**
https://github.com/harvester/harvester/issues/2692

**Solution:**
Add the harvester's own storage class called `harvester-longhorn` and make it to default
Also, keep the old default storage class 'longhorn' in case something we do not consider and upgrade flow.

**Related Issue:**

**Test plan:**

1. Check there are two storage classes `harvester-longhorn` and `longhorn`
1.1 `longhorn` should contain the parameter `migratable: true` and it should not be default SC.

2. Before upgrading, create a VM then attach a volume (That means VM root and volume use the current storage class `longhorn`). Then back up this VM.
2.1 After upgrading, try to migrate the VM to other nodes.
2.2 After upgrading, try to restore the previous backup. Then check the restore target.


deploy/charts/harvester/templates/longhorn-storageclass-override.yaml
```diff
@@ -15,7 +15,7 @@ data:
     metadata:
       name: longhorn
       annotations:
-        storageclass.kubernetes.io/is-default-class: {{ .Values.longhorn.persistence.defaultClass | quote }}
+        storageclass.kubernetes.io/is-default-class: "false"
```
w13915984028: Please check comment https://github.com/harvester/harvester/issues/2692#issuecomment-1243725721

We need to remove `longhorn-storageclass-override.yaml`, it causes issues.

I am wondering, we should keep `longhorn` untouched and as the default `storageclass`

deploy/charts/harvester/templates/harvester-storageclass.yaml
```diff
@@ -0,0 +1,19 @@
+{{ if .Values.longhorn.enabled -}}
```
w13915984028, line 1: Should we put this `storageclass` into `harvester-installer`, as there has assumed that the `storageclass` is longhorn.

And the `rancher-monitoring` in `harvester-installer`, has already claimed PVC from `longhorn`.

Change `default storageclass` may also cause break with previous version.

package/upgrade/migrations/managed_charts/v1.0.3.sh
```diff
@@ -16,7 +16,7 @@ patch_alertmanager_enable()
   yq e '.spec.values.alertmanager.config.global.resolve_timeout = "5m"' $CHART_MANIFEST -i
   yq e '.spec.values.alertmanager.alertmanagerSpec.retention = "120h"' $CHART_MANIFEST -i
   yq e '.spec.values.alertmanager.alertmanagerSpec.resources = {"limits": {"cpu": "1000m", "memory": "600Mi"}, "requests": {"cpu": "100m", "memory": "100Mi"}}' $CHART_MANIFEST -i
-  yq e '.spec.values.alertmanager.alertmanagerSpec.storage.volumeClaimTemplate.spec.storageClassName = "longhorn"' $CHART_MANIFEST -i
+  yq e '.spec.values.alertmanager.alertmanagerSpec.storage.volumeClaimTemplate.spec.storageClassName = "harvester-longhorn"' $CHART_MANIFEST -i
```
w13915984028, line 19: For upgrade, the `storageclassname` seems should keep unchanged. As `storageclassname` is only used one time when PVC is created, later change of `storageclassname` has no effect, and may confuse user.

deploy/charts/harvester/templates/harvester-storageclass.yaml
w13915984028: Another benefit:

When this `harvester-longhorn` storageclass is in `harvester-installer`, it's convenient to change and make ISO, e.g., set replica to `1` instead of `3`.

bk201: From discussion with @guangbochen, If the user needs a different replica setting. He can create a new storage class on day2 and make it as default if needed.

deploy/charts/harvester/templates/longhorn-storageclass-override.yaml
w13915984028: The overwride of LH configmap causes: https://github.com/harvester/harvester/issues/2088

this file must be removed.

deploy/charts/harvester/templates/harvester-storageclass.yaml
w13915984028: The internal PVC like monitoring, they can only be changed via installation, like https://github.com/harvester/harvester/issues/2692#issuecomment-1243725721

deploy/charts/harvester/templates/longhorn-storageclass-override.yaml
w13915984028: The value of `is-default-class` should be changed via `.Values.longhorn.persistence.defaultClass`, if really needed.

package/upgrade/migrations/managed_charts/v1.0.3.sh
guangbochen: alertmanage pvc should not exist in the previous version, so it should be fine to update in here.

deploy/charts/harvester/templates/longhorn-storageclass-override.yaml
guangbochen: For new clusters we may consider to set `longhorn` [as non-default SC](https://github.com/longhorn/charts/blob/master/charts/longhorn/values.yaml#L57), for exisitng clusters we may try to patch it manually via upgrade script? so this file can be removed completely.

deploy/charts/harvester/templates/harvester-storageclass.yaml
guangbochen: We will keep the original SC `longhorn` and adding a new one like `harvester-longhorn`(the default one), the exisitng PVC will remain unchanged.

package/upgrade/migrations/managed_charts/v1.0.3.sh
w13915984028: OK, sounds reasonable.

deploy/charts/harvester/templates/harvester-storageclass.yaml
w13915984028: I would suggest to put this new one, in `harvester-installer`, in `rancherd-10-harvester`, it is defined before `managedchart` reference it.

package/upgrade/migrations/managed_charts/v1.0.3.sh
w13915984028: Then another patch file is needed to create `harvester-longhorn` `storageclass` in upgrade. right ?

deploy/charts/harvester/templates/longhorn-storageclass-override.yaml
Vicente-Cheng: Done, I remove this file. and add the related config to `Values`.

Thanks for the reminder and please review it again

deploy/charts/harvester/templates/harvester-storageclass.yaml
bk201: It's better to stay in the helm chart here, considering the upgrade path. rancherd-10-harvester doesn't run in upgrade path.

package/upgrade/migrations/managed_charts/v1.0.3.sh
bk201: If it's in the new helm chart, it will be available.

pkg/controller/master/backup/restore.go
```diff
@@ -631,6 +639,12 @@ func (h *RestoreHandler) createRestoredPVC(
 		Kind:     volumeSnapshotKindName,
 		Name:     dataSourceName,
 	}
+	/* Because we deprecate the storage class `longhorn` from v1.1.0,
+	 * we need to create the restore volume to the new default storage class
+	 */
+	if *spec.StorageClassName == "longhorn" && defaultSC != nil {
```
w13915984028: Question: will `longhorn` be blocked by validator from v1.1.0 , does it mean, we have code to deny any further use of storageclass `longhorn` ?

Vicente-Cheng: I thought that we should block any volume created with storage `longhorn`.
The parameter `Migratable` is used for harvester volume (that does not create SM when access mode is RWX).
If we use the RWX volume that missing migratable, that would cause the VMs to do migration failure with this volume.

We should deny it from 1.1.0

guangbochen: good catch, IIRC any modification to the longhorn SC will be reverted by the longhorn controller, and the only way to patch it is through the configmap.
So from the Harvester UI, we should either consider to hide the `longhorn` SC or make it not configurable, but users should still be able to use the existing longhorn SC if wanted and the hard coded change in here proabbly is not a good option,
If we choose the hide the original `longhorn` SC from the UI, the best way for VM restores to be migrated to the new SC is to allow select differnt SC during the restore(and the longhorn SC can't be selected). cc @bk201 @futuretea @FrankYang0529

package/upgrade/migrations/managed_charts/v1.0.3.sh
w13915984028: This advantage is weight.

By contrast, the upgrade of `managedchart` is done by those scripts.

bk201: This is similar to having a new CRD or any manifest. Normally we ship it within the chart, not adding a new script to create it.
The migration script here is used to migrate chart values or more hacking things.

pkg/controller/master/backup/restore.go
Vicente-Cheng: The old `longhorn` SC should not be used (the restore operation would create a new PVC for it) after we do not override (due to missing the `Migratable` parameter). There are some problems if we use the new volume and its PVC created by non-override `longhorn` SC.

Hardcoded `longhorn` here because we only need to prevent creating any volume with the old SC `longhorn` for the above reason. The other volumes with different SC could work well.

For now, I just assign the old `longhorn` to the new default SC. Other volumes keep restoring to the same SC. 
Agree that the best way to restore is that allows users to select what they want (sure, filter out something we do not want users to know).

Vicente-Cheng: After testing, the old `longhorn` would modify the parameter `migratable` by dependency charts.
We do not need to change the restore behavior until starting to work on issue #2771.

Thanks for the opinions from @w13915984028, @guangbochen, @bk201

Vicente-Cheng:
Hi all, please check the latest changes again. Thanks :)

guangbochen (approved):
LGTM, thanks. and we just need to verify more cases during the rc1 upgrade.

w13915984028 (commented):
LGTM, thanks.

FrankYang0529 (approved):
LGTM. Thanks.

Vicente-Cheng:
link to [[FEATURE] let users choose the storage tiering to the restore target](https://github.com/harvester/harvester/issues/2771#top).
it's easier to track

bk201 (approved):
LGTM!

w13915984028 (commented):
LGTM, thanks.

Issue #2767, opened 2022-09-14
[BUG] vm cannot backup


WuJun2016:
**Describe the bug**

Install harvester with the latest harvester iso, vm cannot backup.

![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/974c48ae-06d3-4ab4-9d7a-19d84c4c10e9)

**To Reproduce**
1. Create a vm
2. backup this vm

**Expected behavior**

Can backup normally

**Support bundle**

**Environment**
 - Harvester ISO version:  master-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

FrankYang0529:
It looks like using a default value in kubebuilder is not enough. I will try to add `kubebuilder:validation:Optional` and see whether it can auto set to `backup`.

https://github.com/harvester/harvester/blob/032742f0508935c47a785f1434cb373e8826464c/deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachinebackups.yaml#L79-L87

TachunLin:
Verified **fixed** on `master-9c3f8a63-head` (09/20). Close this issue.

### Result
* Can initialize the VM backup to Amazon S3 
  ![image](https://user-images.githubusercontent.com/29251855/191224295-4e2ed661-1deb-418a-ad7a-5de44270f453.png)
* And complete the backup
![image](https://user-images.githubusercontent.com/29251855/191226026-8d292375-447f-4580-b486-10e43e126875.png)
 
* Can initialize the VM backup to NFS 
  ![image](https://user-images.githubusercontent.com/29251855/191235390-fd02a76e-eb80-49fa-a8b1-53563b730482.png)

### Test Information
* Test Environment: 1 node Harvester local kvm machine
* Harvester version: `master-9c3f8a63-head` (09/20)

### Verify Steps
1. Create a virtual machine
2. Setup the Amazon S3 backup target 
3. Create a VM backup to S3 
4. Set the NFS backup target on longhorn, refer to https://github.com/harvester/harvester/issues/2725#issuecomment-1244959046
5. Create a VM backup to NFS

PR #2773, opened 2022-09-16
fix: change backup type to optional

FrankYang0529:
**Problem:**
1. Can't create VMBackup through UI.
2. If we try to create a VMBackup without setting backup-target, it shows `spec.source.name` error, but it should be `spec.type` error.
3. If we create a VMBackup and delete the source VM, the VMBakcup can't be restored, because it tries to mount source VM Volumes.

**Solution:**
1. Change `VMBackup.spec.type` field as option.
2. Change error field name in webhook.
3. Only mount VM Volumes if it's a VM snapshot.

**Related Issue:**
https://github.com/harvester/harvester/issues/2767

**Test plan:**
Case 1: Webhook can show correct error type.
2. Create a VM `source-vm`.
3. Try to apply the following YAML, and it should show there is an error on `spec.type`.
```
kind: VirtualMachineBackup
metadata:
  name: vm-backup
  namespace: default
spec:
  source:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: source-vm
```

Case 2: We can create a VMBackup through UI.
1. Setup backup-target.
2. Click `Take Backup` on the `source-vm`.
3. A new VMBackup should be created successfully.

Case 3: VMBackup can be restored without `source-vm`
1. Delete `source-vm`.
2. Restore a VM from the case 2 VMBackup. VM should be restored successfully.

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -6683,8 +6683,7 @@
     "harvesterhci.io.v1beta1.VirtualMachineBackupSpec": {
       "type": "object",
       "required": [
-        "source",
-        "type"
+        "source"
       ],
       "properties": {
         "source": {
@@ -6692,8 +6691,7 @@
           "$ref": "#/definitions/k8s.io.v1.TypedLocalObjectReference"
         },
         "type": {
-          "type": "string",
-          "default": ""
+          "type": "string"
         }
       }
     },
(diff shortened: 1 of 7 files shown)


guangbochen (approved):
LGTM, thanks.

bk201 (approved):
Thanks for update the swagger part.

PR #2649, opened 2022-08-10
Add v1.0.3 release

guangbochen:
**Description:**
Add Harvester v1.0.3 release. Both UI and installer version bumps are identical to the `v1.0.3-rc2` releases.

PR #2777, opened 2022-09-19
feat(upgrade): block upgrades from unsupported versions

starbops:
**Problem:**

Users can upgrade from any version as long as they have the ISO image. They can host the image and create a Version resource that points the URL to the image location or even upload the image from the dashboard to directly trigger the upgrade. There is no version support matrix provided nor version checking logic implemented. The unsupported upgrade path might introduce unpredictable issues and burden the team with a heavy load.

**Solution:**

Users who upgrade from the version older than the specified minimum version required, which was retrieved from `harvester_release.yaml`, will be errored out in phase 2 right before the image preloading step of the entire upgrade flow.

The current strategy is by introducing a support matrix YAML file and picking a version as the minimum required version to do the upgrade in each release, put that version string into a file, and package the info into the ISO image. While upgrading, a repo VM will be spun up and hosted `harvester-release.yaml` which contains the minimum upgradable version string. Then we put version checking logic in the `upgrade_node.sh` & `upgrade_manifest.sh` scripts, right before the image preloading step and manifests upgrading, respectively. If the checking fails, the upgrade Job will fail.

And to further improve the waiting time for an unsupported upgrade to fail, we also do the checking logic inside the upgrade controller so that the upgrade will fail even before the upgrade Plan resource is created. The relevant code path is only effective in future upgrades after the code has been deployed on the target clusters since Harvester uses the currently running version of the controller in the earlier upgrade process (until the images have been preloaded).

Assumed v1.1.0 release:

- cur_ver=v1.0.0, min_upgradable_ver=v1.0.3 => check failed
- cur_ver=v1.0.3, min_upgradable_ver=v1.0.3 => check pass
- cur_ver=v1.0.4, min_upgradable_ver=v1.0.3 => check pass

**Related Issue:**

The original issue is #2431, and this PR also required PR harvester/harvester-installer#338 to work.

**Test plan:**

The test plan includes two parts: 

The first part is the ISO image building. This is to check whether the corresponding `minUpgradableVersion` is correctly injected into the ISO image (However, this depends on PR harvester/harvester-installer#338. Without it being merged, the test cannot continue). Here we should at least test two scenarios:

1. Build an ISO image with a release tag, say v1.1.0 (the resulting ISO image name will be like `harvester-v1.1.0-amd64.iso`)
2. Build an ISO image without a release tag (the resulting ISO image name will be like `harvester-xxxxxxx-dirty-amd64.iso`)

To check the value of `minUpgradableVersion`, execute `isoinfo -J -i harvester-xxxxxxx-amd64.iso -x /harvester-release.yaml | yq e ".minUpgradableVersion"`. It should return "v1.0.3" as a result for the first case and an empty string for the second case.

---

The second part of the test plan is focusing on the upgrade behaviors with different ISO images. There will be 3 test scenarios in total. The first test is to upgrade an unsupported version of cluster with a release ISO image. The second test is to upgrade a supported version of cluster with a release ISO image. The last test is to upgrade a cluster with a dev-built ISO image, i.e. using an ISO image built without a release tag.

Test 1: Build an ISO image **with** release tag, say v1.1.0 (provided minUpgradableVersion: v1.0.3), then upgrade a **v1.0.2** cluster

1. Prepare a v1.0.2 cluster
2. Host the ISO image on an HTTP server
3. Create a `Version` resource

```bash
# cat <<EOF | kubectl apply -f -
kind: Version
metadata:
  name: v1.1.0
  namespace: harvester-system
spec:
  isoURL: "http://192.168.122.1:8000/harvester-c76dfbd-dirty-amd64.iso"
EOF
```

4. Click the "Upgrade" button and choose "v1.1.0"
5. Upgrade should **fail** with the following logs and UI should indicate the upgrade Job has failed

```bash
# kubectl -n harvester-system get po -l harvesterhci.io/upgradeComponent=manifest
NAME                                       READY   STATUS   RESTARTS   AGE
hvst-upgrade-tc9sm-apply-manifests-6hxw7   0/1     Error    0          14m
hvst-upgrade-tc9sm-apply-manifests-82tx9   0/1     Error    0          18m
hvst-upgrade-tc9sm-apply-manifests-n2khj   0/1     Error    0          8m29s
hvst-upgrade-tc9sm-apply-manifests-vfccp   0/1     Error    0          19m
hvst-upgrade-tc9sm-apply-manifests-vhp6r   0/1     Error    0          19m
hvst-upgrade-tc9sm-apply-manifests-wdl4g   0/1     Error    0          18m
hvst-upgrade-tc9sm-apply-manifests-wk2t9   0/1     Error    0          17m
# kubectl -n harvester-system logs hvst-upgrade-tc9sm-apply-manifests-6hxw7
...
harvester: v1.1.0
harvesterChart: 1.1.0
os: Harvester c76dfbd-dirty
kubernetes: v1.22.12+rke2r1
rancher: v2.6.4-harvester3
monitoringChart: 100.1.0+up19.0.3
kubevirt: 0.54.0
minUpgradableVersion: v1.0.3
rancherDependencies:
  fleet:
    chart: 100.0.3+up0.3.9
    app: 0.3.9
  fleet-crd:
    chart: 100.0.3+up0.3.9
    app: 0.3.9
  rancher-webhook:
    chart: 1.0.4+up0.2.5
    app: 0.2.5
Current version: 1.0.2
Minimum upgradable version: 1.0.3
Current version is not supported. Abort the upgrade.
```

image

TEST 2: Build an ISO image **with** release tag, say v1.1.0 (provided minUpgradableVersion: v1.0.3), then upgrade a **v1.0.3** cluster

1. Prepare a v1.0.3 cluster
2. Host the target ISO image on an HTTP server
3. Create a `Version` resource on the cluster

```bash
# cat <<EOF | kubectl apply -f -
kind: Version
metadata:
  name: v1.1.0
  namespace: harvester-system
spec:
  isoURL: "http://192.168.122.1:8000/harvester-c76dfbd-dirty-amd64.iso"
EOF
```

4. Click the "Upgrade" button and choose "v1.1.0"
5. Upgrade should **success**

TEST 3: Build an ISO image **without** release tag then upgrade a **v1.0.2** cluster

1. Prepare a v1.0.2 cluster
2. Host the target ISO image on an HTTP server
3. Create a `Version` resource on the cluster

```bash
# cat <<EOF | kubectl apply -f -
kind: Version
metadata:
  name: v1.1.0
  namespace: harvester-system
spec:
  isoURL: "http://192.168.122.1:8000/harvester-c76dfbd-dirty-amd64.iso"
EOF
```

4. Click the "Upgrade" button and choose "v1.1.0" on the UI
5. Upgrade should **success** since upgrading to any dev builds is not in restriction for dev purposes

### DIFF
--- a/package/upgrade-matrix.yaml
+++ b/package/upgrade-matrix.yaml
@@ -1,0 +1,9 @@
+versions:
+- name: v1.1.0
+  minUpgradableVersion: v1.0.3
+- name: v1.0.3
+  minUpgradableVersion: v1.0.2
+- name: v1.0.2
+  minUpgradableVersion: v1.0.1
+- name: v1.0.1
+  minUpgradableVersion: v1.0.0
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sync"
 
+	semverv3 "github.com/Masterminds/semver/v3"
 	provisioningv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	provisioningctrl "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
@@ -155,10 +156,38 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 		}
 		toUpdate.Status.SingleNode = singleNode
 
-		repoInfo, err := repo.getInfoStr()
+		repoInfo, err := repo.getInfo()
 		if err != nil {
 			setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, err.Error(), "")
 			return h.upgradeClient.Update(toUpdate)
+		}
+		repoInfoStr, err := repoInfo.Marshall()
+		if err != nil {
+			setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, err.Error(), "")
+			return h.upgradeClient.Update(toUpdate)
+		}
+
+		logrus.Info("Check minimum upgradable version")
+		minUpgradableVersion := repoInfo.Release.MinUpgradableVersion
+		if minUpgradableVersion == "" {
+			logrus.Debug("No minimum upgradable version specified, continue the upgrading")
+		} else {
+			constraint := fmt.Sprintf(">= %s", minUpgradableVersion)
+
+			c, err := semverv3.NewConstraint(constraint)
+			if err != nil {
+				return nil, err
+			}
+			v, err := semverv3.NewVersion(upgrade.Status.PreviousVersion)
+			if err != nil {
+				return nil, err
+			}
+
+			if a := c.Check(v); !a {
+				message := fmt.Sprintf("The current version %s is less than the minimum upgradable version %s.", upgrade.Status.PreviousVersion, minUpgradableVersion)
+				setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, "Current version not supported.", message)
+				return h.upgradeClient.Update(toUpdate)
+			}
 		}
 
 		logrus.Debug("Start preparing nodes for upgrade")
@@ -168,7 +197,7 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 		}
 
 		toUpdate.Labels[upgradeStateLabel] = StatePreparingNodes
-		toUpdate.Status.RepoInfo = repoInfo
+		toUpdate.Status.RepoInfo = repoInfoStr
 		harvesterv1.NodesPrepared.CreateUnknownIfNotExists(toUpdate)
 		return h.upgradeClient.Update(toUpdate)
 	}
--- a/scripts/version
+++ b/scripts/version
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+TOP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
+PACKAGE_DIR="${TOP_DIR}/package"
+
 unset DIRTY
 if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
     DIRTY="-dirty"
@@ -12,8 +15,13 @@ GIT_TAG=$(git tag -l --contains HEAD | head -n 1)
 
 if [[ -z "$DIRTY" && -n "$GIT_TAG" ]]; then
     VERSION=$GIT_TAG
+    # Search for minimum upgradable version.
+    # If the release does not come with a corresponding minimum upgradable version, the build will error out.
+    MIN_UPGRADABLE_VERSION=$(yq -e e ".versions | map(select(.name == \"$VERSION\")) | .[0].minUpgradableVersion" "${PACKAGE_DIR}"/upgrade-matrix.yaml)
 else
     VERSION="${COMMIT}${DIRTY}"
+    # Dev builds do not come with this restriction.
+    MIN_UPGRADABLE_VERSION=""
 fi
 
 # Chart tag.
@@ -51,3 +59,4 @@ echo "SUFFIX: $SUFFIX"
 echo "REPO: $REPO"
 echo "TAG: $TAG"
 echo "IMAGE_PUSH_TAG: $IMAGE_PUSH_TAG"
+echo "MIN_UPGRADABLE_VERSION: $MIN_UPGRADABLE_VERSION"
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -19,7 +19,7 @@ RUN apt-get update -qq && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/*
 
 # install yq
-RUN GO111MODULE=on go install github.com/mikefarah/yq/v4@v4.6.0
+RUN GO111MODULE=on go install github.com/mikefarah/yq/v4@v4.27.5
 # set up helm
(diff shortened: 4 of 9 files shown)


package/upgrade-matrix.yaml
```diff
@@ -0,0 +1,9 @@
+versions:
+- name: v1.1.0
+  minUpgradableVersion: v1.0.3
+- name: v1.0.3
+  minUpgradableVersion: v1.0.2
+- name: v1.0.2
+  minUpgradableVersion: v1.0.1
+- name: v1.0.1
+  minUpgradableVersion: v1.0.0
```
bk201, line 9: I recommend moving the matrix file to the `harvester/harvester` repo. It's more natural to update the file when we need to cut a release (e.g.,: https://github.com/harvester/harvester/pull/2649)

Other parts lgtm, nice work!

bk201: My bad. I thought the repo is the installer.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -161,6 +162,32 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 			return h.upgradeClient.Update(toUpdate)
 		}
 
+		logrus.Info("Check minimum upgradable version")
+		minUpgradableVersion, err := repo.getMinUpgradableVersionStr()
```
bk201: Maybe we can refactor the code `getInfoStr()` to `getInfo()` to avoid creating a new method?

bk201:
Please squash commits as well; thanks!

scripts/version
```diff
@@ -12,8 +15,13 @@ GIT_TAG=$(git tag -l --contains HEAD | head -n 1)
 
 if [[ -z "$DIRTY" && -n "$GIT_TAG" ]]; then
     VERSION=$GIT_TAG
+    # Search for minimum upgradable version.
+    # If the release does not come with a corresponding minimum upgradable version, the build will error out.
+    MIN_UPGRADABLE_VERSION=$(yq -e e ".versions | map(select(.name == \"$VERSION\")) | .[0].minUpgradableVersion" "${PACKAGE_DIR}"/upgrade-matrix.yaml)
```
bk201, line 20: It would be nice to print out `MIN_UPGRADABLE_VERSION` at the end of the file.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -161,6 +162,32 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
...
+				return nil, err
+			}
+			v, err := semverv3.NewVersion(upgrade.Status.PreviousVersion)
+			if err != nil {
+				return nil, err
+			}
+
+			if a := c.Check(v); !a {
+				reason := fmt.Sprintf("The current version %s is less than the minimum upgradable version %s.", upgrade.Status.PreviousVersion, minUpgradableVersion)
+				setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, reason, "Current version not supported.")
```
bk201: Here is what's in the condition when there is an error:

```
  conditions:
  - lastUpdateTime: "2022-09-26T10:32:07Z"
    message: Current version not supported.
    reason: The current version v1.1.0-dev1 is less than the minimum upgradable version
      v1.0.3.
    status: "False"
    type: Completed
```

The string "The current version v1.1.0-dev1 is less than the minimum upgradable version v1.0.3." is better to stay in the `message` field. One reason is that GUI displays the message first then the reason. Currently, it's not very informative:

Screen Shot 2022-09-26 at 9 56 43 PM

We can check Nodes as an example; there are reasons like `KubeletHasSufficientMemory` and `KubeletHasNoDiskPressure`.

The existing `setUpgradeCompletedCondition` calls put the error message as `reason` (my bad); I think we need to refactor them (feel free to do it with another PR).

scripts/version
starbops: Updated it, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Swapped the reason and message. Thanks!

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Since both `getInfoStr()` and `getMinUpgradableVersionStr()` are one-off methods, maybe we can just simply extract what we need from `getInfo()` in the caller function. Is that what you meant?

starbops: Updated it, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Now the error message is like:

Screen Shot 2022-09-27 at 14 17 35

```bash
harv-2431:~ # k -n harvester-system get upgrades hvst-upgrade-rhnv6 -o jsonpath='{.status.conditions}' | jq .
[
  {
    "lastUpdateTime": "2022-09-27T06:16:15Z",
    "message": "The current version v1.1.0 is less than the minimum upgradable version v1.1.1.",
    "reason": "Current version not supported.",
    "status": "False",
    "type": "Completed"
  },
  {
    "lastUpdateTime": "2022-09-27T06:15:07Z",
    "status": "True",
    "type": "ImageReady"
  },
  {
    "lastUpdateTime": "2022-09-27T06:16:14Z",
    "status": "True",
    "type": "RepoReady"
  }
]
```

bk201:
@guangbochen @FrankYang0529  The fix is in two places:
- Upgrade controller: fail early if the user upgrades an in-compatible ISO. This only takes effect in the next version.
- upgrade_manifest.sh and upgrade_node.sh script. This serves as a guard for previous versions. Because these new scripts are shipped with the new ISO.

bk201 (approved):

FrankYang0529 (approved):
LGTM. Thanks.

guangbochen (approved):
LGTM, thanks.

Issue #2782, opened 2022-09-19
[BUG] Upgrade fails because upgrade-repo VM is not becoming healthy and gets deleted again


bashofmann:
**Describe the bug**
The Harvester cluster was installed with 1.0.2 and is imported into a Rancher 2.6.8. Now it should be updated to 1.0.3, but the update fails when starting the upgrad-repo VM.

**To Reproduce**
Start an update from 1.0.2 to 1.0.3. The update starts and the upgrade download finishes successfully after a few minutes. The upgrade-repo VM is being started and also comes up. In Harvester you can see that the VM is running, but not healthy. In Longhorn you can see that a VM disk volume was successfully created and mounted.
You can see the following events on the VM:

```
Readiness probe failed: Get "http://10.52.3.3:80/harvester-iso/harvester-release.yaml": dial tcp 10.52.3.3:80: connect: no route to host
Readiness probe failed: Get "http://10.52.3.3:80/harvester-iso/harvester-release.yaml": dial tcp 10.52.3.3:80: connect: connection refused
```

`10.52.3.3` is the IP of the upgrade-repo VM.

![Bildschirmfoto 2022-09-19 um 10 33 09](https://user-images.githubusercontent.com/243056/191052064-d2dc7841-5150-4ce7-81f2-37c2ebd53c3b.png)

Then the VM is deleted again. The upgrade fails with a `502 Host not found` error.

![Bildschirmfoto 2022-09-19 um 10 33 27](https://user-images.githubusercontent.com/243056/191052003-286b2534-861c-47f6-88af-c9c3332f9964.png)

**Expected behavior**
The update successfully runs through.

**Support bundle**
Can be provided internally.

**Environment**
 - Harvester ISO version:  1.0.2
 - Underlying Infrastructure: 
 4 nodes, each 64 cores, 756gb ram. 2TB disks. Usage of all resources very low.
 Internet access through HTTP proxy. HTTP proxy is configured within Harvester and working. The HTTP proxy does not intercept or break TLS. no_proxy is configured will all internal ip ranges and TLDs. We also added `longhorn-system` and `cattle-system` manually to rule out https://github.com/harvester/harvester/issues/2436.

bashofmann:
According to the logs in the support bundle, there are `apply-system-agent-upgrader` jobs failing with this:

```
2022-09-19T08:36:16.202270909Z + eval 'v="$HTTP_PROXY"'
[...]

PR #2808, opened 2022-09-22
fix(setting): add harvester-system as additional noProxy value

starbops:

**Problem:**

If users choose to enable and configure the http(s)Proxy but leave the noProxy field empty, they will encounter failure while upgrading Harvester. This is not an air-gapped specific problem. If users do not configure http proxy, they won't face this issue even in an air-gapped environment. The problem is when the repo VM is spun up, a corresponding launcher Pod will continue to monitor its healthiness by hitting a certain endpoint periodically, this is the so-called "readiness probe". According to the access log on the HTTP proxy server, the readiness probes are hitting `http://upgrade-repo-hvst-upgrade-xxxxx.harvester-system:80/harvester-iso/harvester-release.yaml`. Since the `harvester-system` domain is not excluded, the probes are being proxied. The HTTP proxy has no clue where it can reach that URL because it's only reachable inside the Kubernetes cluster.

**Solution:**

Simply adding `harvester-system` as another noProxy domain can solve the problem. The noProxy environment variable will propagate to the launcher Pod of the repo VM. We have to do this secretly while users configure their own http(s)Proxy and noProxy values.

This makes sure the internal traffic of Kubernetes like readiness probes of repo VM is not proxied.

**Related Issue:**

#2782 

**Test plan:**

1. Set up a Harvester node with an older version, e.g. v1.0.2, which is then able to upgrade
2. Set up an HTTP proxy, e.g. Squid
3. Configure HTTP proxy in the Setting page of Harvester UI
4. Wait until the backends are back (single node will have to wait)
5. Start the upgrade
6. The upgrade should be successful

### DIFF
--- a/pkg/util/proxy.go
+++ b/pkg/util/proxy.go
@@ -14,6 +14,7 @@ var builtInNoProxy = []string{
 	"longhorn-system",
 	"cattle-system",
 	"cattle-system.svc",
+	"harvester-system",
 	".svc",
 	".cluster.local",
 }
--- a/pkg/util/proxy_test.go
+++ b/pkg/util/proxy_test.go
@@ -15,17 +15,17 @@ func Test_AddBuiltInNoProxy(t *testing.T) {
 		{
 			name:   "single item",
 			input:  "192.168.1.0/24",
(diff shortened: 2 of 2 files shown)


bk201 (approved):
LGTM

guangbochen (approved):
LGTM, also need to update the doc, thanks.

PR #2829, opened 2022-09-26
initial implementation for addon framework

ibrokethecloud:

**Problem:**

The PR introduces an Addon CRD and controller, which wraps around the HelmChart CRD.
This allows new optional components to be packaged with harvester, while not enabling them.
This ensure that the core harvester requirements stay within reasonable bounds.

In addition, this also allows us to move off ManagedCharts and eventually HelmCharts as the means to package components.

**Solution:**

**Related Issue:**

**Test plan:**

*Enable vm-import-controller add-on*
- No addon is enabled by default, you can check with the command.
```
kubectl get addons.harvesterhci.io -A
```
- Enable the `vm-import-controller` addon:
```
kubectl edit addons.harvesterhci.io vm-import-controller -n harvester-system
```
And change `spec.enabled` to `true`.
- You should see a vm-import-controller HelmChart CR:
```
$ kubectl get helmchart -n harvester-system
NAME                   AGE
vm-import-controller   2s
```
- And a pod
```
$ kubectl get pods -n harvester-system | grep vm-import
helm-install-vm-import-controller-k5hm9                           0/1     Completed   0             61s
vm-import-controller-harvester-vm-import-controller-84749cns69l   1/1     Running     0             58s
```

*Disable vm-import-controller add-on*

- The add-on can also be disabled:
```
kubectl edit addons.harvesterhci.io vm-import-controller -n harvester-system
```
And change `spec.enabled` to `false`.

- You should see a vm-import-controller HelmChart CR is gone:
```
$  kubectl get helmchart -n harvester-system
No resources found in harvester-system namespace.
```

- And no vm-import-controller pod:
```
kubectl get pods -n harvester-system | grep vm-import
```


pkg/apis/harvesterhci.io/v1beta1/addon.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	AddonEnabled  = "AddonEnabled"
+	AddonDeployed = "AddonDeploySuccessful"
+	AddonFailed   = "AddonDeployFailed"
+)
```
guangbochen: use constant type is more preferred.

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+	helmv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
+	ctlhelmv1 "github.com/k3s-io/helm-controller/pkg/generated/controllers/helm.cattle.io/v1"
+	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/batch/v1"
+	"github.com/rancher/wrangler/pkg/relatedresource"
+	batchv1 "k8s.io/api/batch/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"strings"
+	"time"
```
guangbochen: fix import style with system utils, harvester generated, and 3rd parties.

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+	job   ctlcorev1.JobController
+}
+
+var (
+	defaultDeployDuration = &metav1.Duration{Duration: 10 * time.Minute}
+)
+
+const (
+	defaultValuesAnnotation = "harvesterhci.io/addon-defaults"
+)
```
guangbochen: can be moved to /pkg/util/constants.go

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	if job.Status.Succeeded > 0 {
+		a.Status.Status = harvesterv1.AddonDeployed
+	}
+
+	return h.addon.UpdateStatus(a)
+}
+
+// OnAddonChange will reoncile the Addon CRD's
```
guangbochen:
```suggestion
// OnAddonChange will reconcile the Addon CRDs
```

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+func (h *Handler) OnAddonChange(key string, a *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+	if a == nil || !a.DeletionTimestamp.IsZero() {
```
guangbochen: prefer to use `a.DeletionTimestamp != nil` for consistency.

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	if !a.Spec.Enabled && ok {
+		return h.removeHelmChart(a)
+	}
+	return nil, nil
+}
+
+func (h *Handler) checkHelmChartExists(a *harvesterv1.Addon) (bool, error) {
+	_, err := h.helm.Get(a.Namespace, a.Name, metav1.GetOptions{})
```
guangbochen: get from the cache.

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+	return nil
+}
+
+// MonitorChanges will trigger updates to helm chart based on chages in values
+func (h *Handler) MonitorChanges(key string, a *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+	if a == nil || !a.DeletionTimestamp.IsZero() {
+		return nil, nil
+	}
+
+	if a.Status.Status != harvesterv1.AddonDeployed && a.Status.Status != harvesterv1.AddonFailed {
```
guangbochen, line 54: check if the status is not nil first.

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+		return a, nil
+	}
+
+	hc, err := h.helm.Get(a.Namespace, a.Name, metav1.GetOptions{})
```
guangbochen: use get from the cache.

pkg/apis/harvesterhci.io/v1beta1/addon.go
guangbochen: Prefer to add constant type to it, also can we consider adding conditions so UI can render the error message properly if the deployment is failed.

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+			redeployChart = true
+		} else {
+			return a, fmt.Errorf("error fetching helm chart during monitor changes: %v", err)
+		}
+	}
+
+	// if chart is being deleted then but addon is not then redeploy chart
+	// this is possible since helm-controller adds a finalizer to uninstall the chart when HelmChart CRD is being deleted
+
+	if hc.DeletionTimestamp != nil {
```
guangbochen: if hc is not found it might be nil here

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+	return h.addon.UpdateStatus(a)
+}
+
+// MonitorAddon will track the deployment of Addon
+func (h *Handler) MonitorAddon(key string, a *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+	if a == nil || !a.DeletionTimestamp.IsZero() {
+		return nil, nil
+	}
+
+	if a.Status.Status != harvesterv1.AddonEnabled {
```
guangbochen, line 99: may also return if the Addon is deployed successfully?

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+		return a, nil
+	}
+
+	hc, err := h.helm.Get(a.Namespace, a.Name, metav1.GetOptions{})
```
guangbochen: ditto

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+	if err != nil {
+		return a, fmt.Errorf("error querying helmchart in checkHelmChartExists: %v", err)
+	}
+
+	if hc.Status.JobName == "" {
+		return a, fmt.Errorf("waiting for jobname to be populated in HelmChart status")
+	}
+
+	job, err := h.job.Get(a.Namespace, hc.Status.JobName, metav1.GetOptions{})
```
guangbochen, line 112: ditto and don't swallow error

pkg/controller/master/addon/addon.go
bk201: +1 in `addons_tests.go`, `job_controllers.go` and `suite_test.go`.

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	addonController.OnChange(ctx, "deploy-addon", h.OnAddonChange)
+	addonController.OnChange(ctx, "monitor-addon", h.MonitorAddon)
+	addonController.OnChange(ctx, "monitor-changes", h.MonitorChanges)
+	relatedresource.Watch(ctx, "watch-helmcharts", h.ReconcileHelmChartOwners, addonController, helmController)
+	return nil
+}
+
+// MonitorChanges will trigger updates to helm chart based on chages in values
```
bk201:
```suggestion
// MonitorChanges will trigger updates to helm chart based on changes in values
```

ibrokethecloud:
Related to: https://github.com/harvester/harvester/issues/2845

pkg/controller/master/addon/addon.go
ibrokethecloud: Ideally this would only be called for the first deploy. Any subsequent changes to Addon/Helm chart would be via the MonitorChanges handler

package/upgrade/lib.sh
```diff
@@ -184,3 +184,32 @@ detect_upgrade()
...
+  version=$(cat ./addons/${name}.yaml | yq e .spec.version)
+
+  cat > addon-patch.yaml <<EOF
+spec:
+  version: $version
+EOF
+
+  item_count=$(kubectl get addons.harvesterhci $name -n $namespace -o  jsonpath='{..name}')
+  if [ -z "$item_count" ]; then
+    install_chart $name $namespace
```
w13915984028: do not find definition of `install_chart`, is it `install_addon` ?

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,295 @@
...
+	// if chart is being deleted then but addon is not then redeploy chart
+	// this is possible since helm-controller adds a finalizer to uninstall the chart when HelmChart CRD is being deleted
+
+	if hc.DeletionTimestamp != nil {
+		redeployChart = true
+	}
+
+	if redeployChart {
+		a.Status.Status = ""
+		return h.addon.UpdateStatus(a)
```
w13915984028, line 66: Question: When updating status, shoudl `a` be copied out first ?

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,295 @@
...
+	vals, err := defaultValues(a)
+	if err != nil {
+		return a, fmt.Errorf("error generating default values during monitor changes: %v", err)
+	}
+	if hc.Spec.ValuesContent != vals || hc.Spec.Version != a.Spec.Version || hc.Spec.Chart != a.Spec.Chart || hc.Spec.Repo != a.Spec.Repo {
+		hc.Spec.Chart = a.Spec.Chart
+		hc.Spec.Repo = a.Spec.Repo
+		hc.Spec.Version = a.Spec.Version
+		hc.Spec.ValuesContent = vals
+		_, err = h.helm.Update(hc)
```
w13915984028, line 80: It's essential to add log here, saying helm is forcely updated.

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,295 @@
...
+			return nil
+		} else {
+			return fmt.Errorf("error looking up chart in checkAndDeleteChart: %v", err)
+		}
+	}
+
+	var addonOwned bool
+	for _, v := range hc.GetOwnerReferences() {
+		if v.Kind == a.Kind && v.APIVersion == a.APIVersion && v.UID == a.UID && v.Name == a.Name {
+			addonOwned = true
```
w13915984028, line 219: break loop when match

package/upgrade/lib.sh
```diff
@@ -184,3 +184,32 @@ detect_upgrade()
...
+
+  version=$(cat ./addons/${name}.yaml | yq e .spec.version)
+
+  cat > addon-patch.yaml <<EOF
+spec:
+  version: $version
+EOF
+
+  item_count=$(kubectl get addons.harvesterhci $name -n $namespace -o  jsonpath='{..name}')
+  if [ -z "$item_count" ]; then
```
w13915984028, line 233: Double check: 
in v1.0.3, run `kubectl get addons.harvesterhci` return error, not sure it meets expectation

image

In another similar patch, I use the `if test $UPGRADE_PREVIOUS_VERSION = 1.0.3; then` to make sure the patch only run in this dedicated version.
image

tests/integration/controllers/fake/jobs_controller.go
```diff
@@ -0,0 +1,128 @@
...
+		helm: hc,
+	}
+
+	//jc.OnChange(ctx, "fake-job-controller", h.OnJobChange)
+	hc.OnChange(ctx, "fake-helmchart-controller", h.OnHelmChartChange)
+	return nil
+}
+
+func (h *handler) OnJobChange(key string, j *batchv1.Job) (*batchv1.Job, error) {
+
```
w13915984028: guess this first blank line is ignored

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,296 @@
...
+	if redeployChart {
+		a.Status.Status = ""
+		return h.addon.UpdateStatus(a)
+	}
+
+	vals, err := defaultValues(a)
+	if err != nil {
+		return a, fmt.Errorf("error generating default values during monitor changes: %v", err)
+	}
+	if hc.Spec.ValuesContent != vals || hc.Spec.Version != a.Spec.Version || hc.Spec.Chart != a.Spec.Chart || hc.Spec.Repo != a.Spec.Repo {
```
w13915984028, line 74: Question:
If we use another `repo` to do debug, it will be reverted here, should `repo` be skipped from check ?

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,296 @@
...
+		return a, fmt.Errorf("waiting for job %s to complete", job.Name)
+	}
+
+	// check Failed before since in jobs with more than 1 container then
+	// even a single failure should count as a failure
+	if job.Status.Failed > 0 {
+		a.Status.Status = harvesterv1.AddonFailed
+	}
+
+	if job.Status.Succeeded > 0 {
```
w13915984028: should here be `else if job.Status.Succeeded > 0` ?

pkg/controller/master/addon/addon.go
```diff
@@ -0,0 +1,296 @@
...
+	hc, err := h.helm.Cache().Get(a.Namespace, a.Name)
+	if err != nil {
+		return a, fmt.Errorf("error querying helmchart in checkHelmChartExists: %v", err)
+	}
+
+	if hc.Status.JobName == "" {
+		return a, fmt.Errorf("waiting for jobname to be populated in HelmChart status")
+	}
+
+	job, err := h.job.Get(a.Namespace, hc.Status.JobName, metav1.GetOptions{})
```
w13915984028, line 112: curious, does `helmchart` object itself have status to indicate the deployment is finished or not, why we should check the internal job status ?

ibrokethecloud: helm chart doesnt store the chart status in the status object, only the job name. So figured it may make sense to reconcile job status to reflect in the Addon status for UI.

package/upgrade/lib.sh
ibrokethecloud: the CRD doesnt exist yet. The upgrade script will ship with the version packaging the addon CRD and onwards.

package/upgrade/lib.sh
ibrokethecloud: good spot.. fixed now.

package/upgrade/lib.sh
ibrokethecloud: I am not sure if we want to limit version upgrades from a specific version to another. Ideally with this, since the components are not core harvester you could skip versions if needed.

tests/integration/controllers/fake/jobs_controller.go
ibrokethecloud: I have removed the controller, its not needed any more since i moved the environment setup to use `kind` to leverage k8s object gc.

pkg/controller/master/addon/addon.go
ibrokethecloud: ideally hc is owned by the addon object. To test may be best to update the addon object to use the debug repo.

package/upgrade/lib.sh
guangbochen: Yes, we will have a minimumUpgradeVersion for each new release, so basically for v1.1.0 users can only upgrade starting from v1.0.3. But we can sleep and retry here if CRD is not found since the Harvester API server will apply the missing CRDs eventually.

pkg/controller/master/addon/addon.go
guangbochen: Yes, making a copy before updating is always preferred. cc @ibrokethecloud

pkg/controller/master/setup.go
```diff
@@ -2,6 +2,7 @@ package master
 
 import (
 	"context"
+	"github.com/harvester/harvester/pkg/controller/master/addon"
```
guangbochen, line 6: fix the format here.

tests/integration/controllers/addons_test.go
```diff
@@ -0,0 +1,679 @@
...
+			Eventually(func() error {
+				h, err := helmController.Get(a.Namespace, a.Name, metav1.GetOptions{})
+				if err != nil {
+					return err
+				}
+
+				if h.Status.JobName != "" {
+					jobName = h.Status.JobName
+					GinkgoWriter.Printf("found job name: %s\n", jobName)
+					//fmt.Printf("found job name: %s\n", h.Status.JobName)
```
guangbochen: to remove.

tests/integration/controllers/suite_test.go
```diff
@@ -0,0 +1,160 @@
...
+	harvesterStartTimeOut = 20
+)
+
+func TestAPI(t *testing.T) {
+	defer GinkgoRecover()
+
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "api suite")
+}
```
guangbochen, line 60: Note: not required for the rc2, but seems we can merge this one with the API tests, additionally should we try to deploy the official [helm-controller](https://github.com/k3s-io/helm-controller/blob/master/manifests/deploy-cluster-scoped.yaml) to the kind cluster so we don't need to mimic the job controller and doing a real integration test to detect any potential issues from it, thanks.

ibrokethecloud: yeah i agree.. initially i wanted it to be separate to avoid tainting failures from the existing api testing.

package/upgrade/upgrade_manifests.sh
```diff
@@ -450,6 +450,17 @@ pause_all_charts()
   done
 }
 
+upgrade_addons()
+{
+ if test $UPGRADE_PREVIOUS_VERSION = 1.0.3; then
```
guangbochen: use `< 1.1.0` to help cover potential future v1.0 patch releases.

guangbochen (commented):
In general, this is LGTM, we can prepare to squish the commits where applicable, thanks.

pkg/controller/master/addon/addon.go
w13915984028: Helm stores the `release` in `secrets`, could the `release` reflect the final status of the `chart` ?

package/upgrade/upgrade_manifests.sh
```diff
@@ -450,6 +450,17 @@ pause_all_charts()
   done
 }
 
+upgrade_addons()
+{
+ if test $UPGRADE_PREVIOUS_VERSION < 1.1.0; then
```
w13915984028: Here we have a mistake, please update:

`if test "$UPGRADE_PREVIOUS_VERSION" \< "v1.1.0"; then `
or
`if test "$UPGRADE_PREVIOUS_VERSION" \< v1.1.0; then `

bash needs `\<` to represent "<";  

and the version is  `v1.0.3`, not `1.0.3`

bk201: And be aware this only works for formal releases. Comparison like `test "v1.1.0-rc1" \< "v1.1.0"` is false.
I guess we can keep this one.

bk201:
@ibrokethecloud The PR needs a rebase. Thanks.

package/upgrade/upgrade_manifests.sh
ibrokethecloud: I have changed the logic slightly to make it easier to compare versions

ibrokethecloud:
@bk201 i have rebased the PR.

package/upgrade/lib.sh
```diff
@@ -210,3 +210,54 @@ detect_upgrade()
...
+
+upgrade_addon()
+{
+  local name=$1
+  local namespace=$2
+  local version=$3
+
+  version=$(cat ./addons/${name}.yaml | yq e .spec.version)
+
+  cat > addon-patch.yaml <<EOF
```
bk201, line 227: @ibrokethecloud, I see this error:
```
2022-10-04T05:51:01.376850609Z stderr F cat: ./addons/vm-import-controller.yaml: No such file or directory
2022-10-04T05:51:01.535749767Z stderr F Error from server (NotFound): addons.harvesterhci.io "vm-import-controller" not found
```

We need to add the files in https://github.com/harvester/harvester/blob/master/package/upgrade/Dockerfile

ibrokethecloud: @bk201 let me have a quick look

ibrokethecloud: @bk201 this should now be present.

package/upgrade/lib.sh
```diff
@@ -210,3 +210,54 @@ detect_upgrade()
...
   UPGRADE_PREVIOUS_VERSION=$(echo "$upgrade_obj" | yq e .status.previousVersion -)
 }
+
+upgrade_addon()
+{
+  local name=$1
+  local namespace=$2
+  local version=$3
+
+  version=$(cat ./addons/${name}.yaml | yq e .spec.version)
```
bk201: Need to use the directory in `/usr/local/share`

ibrokethecloud: fixed.

package/upgrade/lib.sh
```diff
@@ -210,3 +210,54 @@ detect_upgrade()
...
+  local version=$3
+
+  version=$(cat /usr/local/share/addons/${name}.yaml | yq e .spec.version)
+
+  cat > addon-patch.yaml <<EOF
+spec:
+  version: $version
+EOF
+
+  item_count=$(kubectl get addons.harvesterhci $name -n $namespace -o  jsonpath='{..name}')
```
bk201: Got error and the job fail (there is `-e` flag in the script)
```
Error from server (NotFound): addons.harvesterhci.io "vm-import-controller" not found
```

I can make it work with:
```
  if ! kubectl get addons.harvesterhci vm-import-controller -n harvester-system &> /dev/null; then
     # install
  else
```

ibrokethecloud: this should now be fixed @bk201

bk201 (approved):
LGTM.

package/upgrade/upgrade_manifests.sh
w13915984028: @bk201   the `v1.1.0` is  < `v1.1.0-rc1`

```
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
..
>>> "v1.1.0" < "v1.1.0-rc1"
True

```

package/upgrade/lib.sh
```diff
@@ -242,17 +242,17 @@ install_addon()
...
 wait_for_addons_crd()
 {
-  item_count=$(kubectl get customresourcedefinitions addons.harvesterhci.io -o  jsonpath='{.metadata.name}')
+  item_count=$(kubectl get customresourcedefinitions addons.harvesterhci.io -o  jsonpath='{.metadata.name}' || true)
   while [ -z "$item_count" ]
   do
     echo "wait for addons.harvesterhci.io crd to be created"
     sleep 10
-    item_count=$(kubectl get customresourcedefinitions addons.harvesterhci.io -o  jsonpath='{.metadata.name}')
+    item_count=$(kubectl get customresourcedefinitions addons.harvesterhci.io -o  jsonpath='{.metadata.name}' || true)
```
w13915984028, line 255: Is here waiting form CRD to be applied by `kubectl` ?  should it be `while [ ! -z "$item_count" ]` ?

package/upgrade/lib.sh
```diff
@@ -222,14 +222,14 @@ upgrade_addon()
   local namespace=$2
   local version=$3
 
-  version=$(cat ./addons/${name}.yaml | yq e .spec.version)
+  version=$(cat /usr/local/share/addons/${name}.yaml | yq e .spec.version)
```
w13915984028, line 225: seems   `version=$(yq e '.spec.version' /usr/local/share/addons/${name}.yaml)` is straightforward

package/upgrade/lib.sh
ibrokethecloud: i think its correct, the item_count will not be set until the harvester controller has deployed the crd

package/upgrade/upgrade_manifests.sh
ibrokethecloud: do we plan to allow upgrades between RC's?

package/upgrade/lib.sh
w13915984028: em, it's correct.

package/upgrade/upgrade_manifests.sh
bk201: No. But we can have a better way to check symvers. Does `lower_version_check` work here?

package/upgrade/lib.sh
w13915984028: maybe should loop at most certain times and then break to report error, avoid dead loop

package/upgrade/upgrade_manifests.sh
w13915984028: image

w13915984028 (approved):
LGTM, thanks.

bk201:
I've created a ticket: https://github.com/harvester/harvester/issues/2877. We need a handy tool to do version comparison.

Issue #2833, opened 2022-09-27
[FEATURE] Add more checks before upgrading


bk201:
**Is your feature request related to a problem? Please describe.**

We need to add some webhook validation checks to forbid users from performing upgrades in some cases.
This gives the user a smooth upgrade experience and ensures data reliability. 

Related issue: https://github.com/harvester/harvester/issues/2741

**Describe the solution you'd like**

Add a validation webhook and reject creating an upgrade if:
- There is any degraded volume (when node count >= 3).
- Node space is not enough. We need ~25GB of free space to preload images.

Add a validation webhook to reject deleting upgrades if
- CAPI cluster is under-provisioning. This is controlled by Rancher, and we should not start over an upgrade again.
  - `kubectl get clusters.cluster.x-k8s.io/local -n fleet-local`
  - Add an exception if there is a special annotation on upgrade to skip.

**Describe alternatives you've considered**

**Additional context**

bk201:
Note, so far, we can't find a way to detect free space in root fs. We can see if this can be done.

noahgildersleeve:
Is this being implemented in v1.1.0? If so would the testing path be for us to simulate an upgrade from v1.1.0 to another version? Mostly I'm asking if this will work in an upgrade from v1.0.3 to v1.1.0. CC @FrankYang0529

bk201:
@noahgildersleeve Yes, The feature will only appear in new v1.1.0 cluster. It doesn't exist in v1.0.3.

noahgildersleeve:
Tested the degraded volume functionality in `v1.1.0-rc2` with an upgrade to the same version `v1.1.0-rc2`. 

## Verification Steps

1. Create an upgrade version.yaml and upload the ISO as outlined [here](https://docs.harvesterhci.io/v1.0/upgrade/automatic/#prepare-an-air-gapped-upgrade)
2. Before running the upgrade navigate to the Longhorn UI at `https://<harvesterIp>/dashboard/c/local/longhorn`
![Screenshot_20221011_152637](https://user-images.githubusercontent.com/83787952/195209777-44287fc3-ba9c-4b00-bb3b-6e85cf3c040f.png)
3. Navigate to the volume tab then open one of the volumes. 
![Screenshot_20221011_152544](https://user-images.githubusercontent.com/83787952/195209854-6157273b-9305-49c1-a8bc-bfc9d2c63972.png)
[...]

PR #2836, opened 2022-09-27
Add upgrade webhook

FrankYang0529:
**Problem:**
We need to add some webhook validation checks to forbid users from performing upgrades in some cases.
This gives the user a smooth upgrade experience and ensures data reliability.

**Related Issue:**
https://github.com/harvester/harvester/issues/2833

**Test plan:**

Case 1: A degraded volume in 3-node cluster can't upgrade.
1. Create a harvester cluster (node >= 3).
2. Create a VM with a big enough volume. Make sure the robustness of the volume is `degraded`.
3. Create a `version` CR.

```
kind: Version
metadata:
  name: v1.1.0
  namespace: harvester-system
spec:
  isoURL: "http://192.168.122.1:8000/harvester-c76dfbd-dirty-amd64.iso"
EOF
```

4. Click the "Upgrade" button and choose "v1.1.0". We can't do upgrade.

Screen Shot 2022-09-28 at 3 35 24 PM

Case 2: A `provisioning` cluster can't upgrade.
1. Follow case 1. Clean up the VM and degraded volumes.
2. Click the "Upgrade" button and choose "v1.1.0".
3. Update `spec.rkeConfig.provisionGeneration` in `fleet-local/local` clusters.provisioning.cattle.io.
```
spec:
  rkeConfig:
    provisionGeneration: 1
```
4. When the phase of `fleet-local/local` clusters.cluster.x-k8s.io is `Provisioning`, delete the upgrade. The upgrade can't be deleted.

Case 3: An upgrade with the `harvesterhci.io/skipWebhook: "true"` annotation can be deleted.
1. Follow case 2.
2. Update the upgrade with annotation `harvesterhci.io/skipWebhook: "true"`.
3. If the phase of `fleet-local/local` clusters.cluster.x-k8s.io is `Provisioned`, you can update `spec.rkeConfig.provisionGeneration` in `fleet-local/local` clusters.provisioning.cattle.io to trigger it becomes `Provisioning`.
4. Delete the upgrade. This time upgrade CR can be deleted.


pkg/webhook/resources/upgrade/validator.go
```diff
@@ -65,5 +85,58 @@ func (v *upgradeValidator) Create(request *types.Request, newObj runtime.Object)
 		return werror.NewConflict(msg)
 	}
 
+	hasDegradedVolume, err := v.hasDegradedVolume()
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+
+	if hasDegradedVolume {
+		return werror.NewBadRequest("there are degraded volumes, please check all volumes are healthy")
```
guangbochen, line 100: Note: we will need to consider exposing the LH volume status in future versions; otherwise it would hard for users to tell and fix those degraded volumes.

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -65,5 +85,58 @@ func (v *upgradeValidator) Create(request *types.Request, newObj runtime.Object)
...
+		}
+	}
+
+	cluster, err := v.clusters.Get(util.FleetLocalNamespaceName, util.LocalClusterName)
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("can't find %s/%s cluster, err: %+v", util.FleetLocalNamespaceName, util.LocalClusterName, err))
+	}
+
+	if cluster.Status.Phase == string(clusterv1alpha4.ClusterPhaseProvisioning) {
+		return werror.NewBadRequest(fmt.Sprintf("cluster %s/%s status is provisioning, please wait for it's provisioned", util.FleetLocalNamespaceName, util.LocalClusterName))
```
guangbochen:
```suggestion
		return werror.NewBadRequest(fmt.Sprintf("cluster %s/%s status is provisioning, please wait for it to be provisioned", util.FleetLocalNamespaceName, util.LocalClusterName))
```

guangbochen (commented):
Those changes are straight forward and LGTM, do we plan to include this one in v1.1.0? cc @bk201

bk201:
> Those changes are straight forward and LGTM, do we plan to include this one in v1.1.0? cc @bk201

Yes. And this only takes effect when the user upgrades from v1.1.0 to its next version.

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -65,5 +85,58 @@ func (v *upgradeValidator) Create(request *types.Request, newObj runtime.Object)
 		return werror.NewConflict(msg)
 	}
 
+	hasDegradedVolume, err := v.hasDegradedVolume()
```
bk201, line 94: @FrankYang0529 Please also help add the `skipWebhookAnnotation` annotation exception here. 
I think users should have all volumes healthy first before doing an upgrade, but let's still keep a way to skip the check. Thanks!

pkg/webhook/resources/upgrade/validator.go
FrankYang0529: Updated it. Thank you.

pkg/webhook/resources/upgrade/validator.go
FrankYang0529: Updated it. Thank you.

bk201 (approved):
LGTM.
We can skip CodeFactor warning, I think that function is reasonable.

guangbochen (approved):
LGTM, thanks.

pkg/webhook/resources/upgrade/validator.go
bk201: Create an issue to track this; thanks! https://github.com/harvester/harvester/issues/2856

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -65,5 +85,64 @@ func (v *upgradeValidator) Create(request *types.Request, newObj runtime.Object)
...
+			return nil
+		}
+	}
+
+	cluster, err := v.clusters.Get(util.FleetLocalNamespaceName, util.LocalClusterName)
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("can't find %s/%s cluster, err: %+v", util.FleetLocalNamespaceName, util.LocalClusterName, err))
+	}
+
+	if cluster.Status.Phase == string(clusterv1alpha4.ClusterPhaseProvisioning) {
```
w13915984028, line 143: seems better to check `ClusterPhaseProvisioned` instead of `ClusterPhaseProvisioning`

bk201: @FrankYang0529 It's a good idea, do you think we can change it?

FrankYang0529: Yeah, I think we can change it to `if cluster.Status.Phase != string(clusterv1alpha4.ClusterPhaseProvisioned)`. I will update it in further PR.

w13915984028: @FrankYang0529 @bk201 
Again, 
the Rancher cluster status check is added in `func (v *upgradeValidator) Delete`,  should we also add it to `(v *upgradeValidator) Create` ?   the wait cluster ready is done in shell script, and the validator should check it at first, right?

bk201: It's added in https://github.com/harvester/harvester/pull/3093

PR #2848, opened 2022-09-29
Update readme [CI SKIP]

vickyhella:
**Problem:**
The new Docusaurus-based site does not have `/latest` URLs. Therefore, they need to be changed to `/v1.0`.

**Solution:**
This PR changes `/latest/` in the doc URLs to `/v1.0/`.


guangbochen (approved):
LGTM, thanks for the fix.

PR #2853, opened 2022-09-29
Add HostDevices feature gate for issue #992

tlehman:
Move the feature gate to this chart instead of dynamically adding it in the [pcidevices](https://github.com/harvester/pcidevices) controller.

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -65,7 +65,7 @@ kubevirt:
 
       ## Specify kubevirt feature gates
       developerConfiguration:
-        featureGates: ["LiveMigration", "HotplugVolumes"]
+        featureGates: ["LiveMigration", "HotplugVolumes", "HostDevices"]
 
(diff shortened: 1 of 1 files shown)


deploy/charts/harvester/values.yaml
```diff
@@ -65,7 +65,7 @@ kubevirt:
 
       ## Specify kubevirt feature gates
       developerConfiguration:
-        featureGates: ["LiveMigration", "HotplugVolumes"]
+        featureGates: ["LiveMigration", "HotplugVolumes", "HostDevices"]
```
guangbochen, line 68: do we also need to enable the GPU feature gate? as I remember it was one of the requested feature support.

tlehman: No, we can support GPUs using only PCI Passthrough. vGPU is a separate feature

ibrokethecloud (approved):
LGTM

bk201 (approved):

Issue #2867, opened 2022-10-04
[BUG] Upgrade: fail to start repo VM in single node case


bk201:
**Describe the bug**

In the single node upgrade case, there is a chance upgrade-repo VM can't start due to API not being ready.

**To Reproduce**
1. Upgrade v1.0.3 to master ISO.
2. In the upgrade-node job, we can see this failure:
```
+ local repo_vm_status
++ kubectl get virtualmachines.kubevirt.io upgrade-repo-hvst-upgrade-4q6hf -n harvester-system '-o=jsonpath={.status.printableStatus}'

Upgrade single node
+ repo_vm_status=Stopped
+ '[' Stopped '!=' Running ']'
+ virtctl start upgrade-repo-hvst-upgrade-4q6hf -n harvester-system
Error starting VirtualMachine the server is currently unable to handle the request
+ true
+ curl -sfL http://upgrade-repo-hvst-upgrade-4q6hf.harvester-system/harvester-iso/harvester-release.yaml
+ echo 'Wait for upgrade repo ready...'
+ sleep 5
```
3. The VM can be started with the `virtctl` command after a while.

**Expected behavior**

The upgrade repo VM should start to serve ISO contents.

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

lanfon72:
Verified this bug has been fixed.

![image](https://user-images.githubusercontent.com/5169694/194038219-d50e0be0-043b-4496-999e-5835e00652c6.png)


Test Information
----
* Environment: **qemu/KVM single node**
* Harvester Version: **v1.0.3** to **master-797245b6-head**
* **ui-source** Option: **Auto**

### Verify Steps:
1. Install Harvester `v1.0.3` with single node
2. Perform Upgrade to master-head via https://docs.harvesterhci.io/v1.0/upgrade/automatic#prepare-an-air-gapped-upgrade
3. Execute command as following to monitor status:
    - `kubectl get jobs -A | grep single-node-upgrade` (to check the _upgrade-node_ job applied)
    - `kubectl logs -fn harvester-system hvst-upgrade-<random_num>-single-node-upgrade-<node-name>`
4. When error message displayed after **Try to bring up the upgrade repo VM...**, it should retry after few seconds.

PR #2868, opened 2022-10-04
fix(upgrade): add retry in case repo VM start failed

starbops:
**Problem:**

In the single node upgrade case, there is a chance upgrade-repo VM can't start due to API not being ready.

**Solution:**

Add a retry logic in case the upgrade-repo VM fails to start.

**Related Issue:**

#2867 

**Test plan:**

Upgrade a single-node cluster, it should finish successfully.

### DIFF
--- a/package/upgrade/lib.sh
+++ b/package/upgrade/lib.sh
@@ -130,7 +130,11 @@ wait_repo()
   # Start upgrade repo VM in case it's shut down due to migration timeout or job failure
   repo_vm_status=$(kubectl get virtualmachines.kubevirt.io $UPGRADE_REPO_VM_NAME -n $UPGRADE_NAMESPACE -o=jsonpath='{.status.printableStatus}')
   if [ "$repo_vm_status" != "Running" ]; then
-    virtctl start $UPGRADE_REPO_VM_NAME -n $UPGRADE_NAMESPACE || true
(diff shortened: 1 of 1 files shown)


package/upgrade/lib.sh
```diff
@@ -123,15 +123,20 @@ ${CURRENT_VERSION}"
...
+}
 
+wait_repo()
+{
   # Start upgrade repo VM in case it's shut down due to migration timeout or job failure
-  repo_vm_status=$(kubectl get virtualmachines.kubevirt.io $UPGRADE_REPO_VM_NAME -n $UPGRADE_NAMESPACE -o=jsonpath='{.status.printableStatus}')
-  if [ "$repo_vm_status" != "Running" ]; then
+  until [[ "$(get_repo_vm_status)" == "Running" ]]
+  do
+    echo "Try to bring up the upgrade repo VM..."
```
w13915984028, line 136: VM may take time to startup, here as long as it is not in "Running", `virtctl start` will run every 10s.

Please double check, if vm is staring, the multi-time `virtctl start` will hurt or not.

Or only 'start' when vm is stopped or error ?

starbops: There's no harm executing `virtctl start` multiple times. Update the PR to the original version. Please check again, thank you!

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):
LGTM, thanks!

PR #2884, opened 2022-10-06
chart: bump longhorn to v1.3.2-rc2

Vicente-Cheng:
We currently could use this PR to verify some issues related to Longhorn.

Related issues:
- https://github.com/harvester/harvester/issues/2906
- https://github.com/harvester/harvester/issues/2911


guangbochen (commented):
Because this is just for validation purposes, so instead of folk the LH templates we would prefer just to replace the zip file or simply update the image version if the manifests remain unchanged from v1.3.1 to v1.3.2.-rc1.

bk201 (approved):
LGTM, thanks for bumping.

guangbochen (approved):
LGTM

Issue #2889, opened 2022-10-07
[BUG] Can't provision RKE2 cluster in Rancher


noahgildersleeve:
**Describe the bug**

When you try to provision RKE2 in Rancher 
**To Reproduce**
1. Import Harvester into Rancher
2. Create cloud creds
3. Start to provision RKE2 cluster
4. Try to complete

**Expected behavior**
RKE2 should provision
**Support bundle**
[supportbundle_ba2138e2-7e7f-4493-acbb-16d3edd63195_2022-10-06T23-57-31Z.zip](https://github.com/harvester/harvester/files/9729328/supportbundle_ba2138e2-7e7f-4493-acbb-16d3edd63195_2022-10-06T23-57-31Z.zip)

**Environment**
 - Harvester ISO version: v1.1.0-rc2
 - Rancher version: v2.6.9-rc3 on single node docker
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): vagrant ipxe-examples

**Additional context**
This seems to be working on RKE1. There is a 400 error on `https://192.168.0.50/k8s/clusters/c-m-d24lpbml/v1/harvester/kubeconfig` I'm attaching screenshots of the errors as well
![Screenshot_20221006_165443](https://user-images.githubusercontent.com/83787952/194439287-575a0a83-b2ca-4b60-97a9-cbac0cbbbba5.png)
![Screenshot_20221006_165338](https://user-images.githubusercontent.com/83787952/194439288-1986cecc-edb2-4452-b9cd-8d0c7613f0e5.png)
![Screenshot_20221006_165302](https://user-images.githubusercontent.com/83787952/194439289-2391fd6a-a45b-4c9e-9a3a-e8d2e9795dd5.png)

khushboo-rancher:
@bk201 Do you think this could be a Rancher issue?

cc: @rebeccazzzz

guangbochen:
could be related to https://github.com/rancher/dashboard/issues/7125, cc @n313893254

futuretea:
refer to https://github.com/rancher/rancher/issues/37711#issuecomment-1126413928

guangbochen:
@n313893254 can we show the kubeconfig error message more obviously on the UI, this will benefit non-developer users, thanks.

guangbochen:
[...]

PR #2897, opened 2022-10-08
Generate a service account token to make it work in 1.24

futuretea:

**Problem:**

Versions of Kubernetes before v1.22 automatically created credentials for accessing the Kubernetes API. This older mechanism was based on creating token Secrets that could then be mounted into running Pods. In more recent versions, including Kubernetes v1.25, API credentials are obtained directly by using the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API

**Solution:**

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token

**Related Issue:**
https://github.com/harvester/harvester/issues/2889

**Test plan:**

### DIFF
--- a/pkg/api/kubeconfig/generate_handler.go
+++ b/pkg/api/kubeconfig/generate_handler.go
@@ -182,12 +182,12 @@ func (h *GenerateHandler) createRoleBindingIfNotExit(clusterRoleName string, sa 
 
 // ensureSaAndSecret returns the serviceAccount and the associated secret
 func (h *GenerateHandler) ensureSaAndSecret(namespace, name string) (*corev1.ServiceAccount, *corev1.Secret, error) {
-	_, err := h.saCache.Get(namespace, name)
+	sa, err := h.saCache.Get(namespace, name)
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, nil, err
 	}
 	if apierrors.IsNotFound(err) {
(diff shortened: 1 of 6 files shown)


pkg/api/kubeconfig/generate_handler.go
```diff
@@ -197,6 +197,35 @@ func (h *GenerateHandler) ensureSaAndSecret(namespace, name string) (*corev1.Ser
...
+				OwnerReferences: []metav1.OwnerReference{
+					{
+						APIVersion: rbacv1.SchemeGroupVersion.Version,
+						Kind:       rbacv1.ServiceAccountKind,
+						Name:       sa.Name,
+						UID:        sa.UID,
+					},
+				},
+				Annotations: map[string]string{
+					"kubernetes.io/service-account.name": sa.Name,
```
guangbochen: for the key prefer to use `v1.ServiceAccountNameKey` instead and add `v1.ServiceAccountUIDKey` as well.

futuretea: updated

yaocw2020 (approved):
LGTM, thanks.

guangbochen (approved):
LGTM, and verified it works well with Rancher v2.6.9-rc3.
image

Issue #2863, opened 2022-10-03
[BUG] Multiple StorageClass can Be Set as "default"


irishgordo:
**Describe the bug**
Within Advanced -> Storage Classes - Storage Classes Index, you can set all storage classes as "default".
Also within:
```
"allowVolumeExpansion": true,
"apiVersion": "storage.k8s.io/v1",
"kind": "StorageClass",
"metadata": {
    "annotations": {
        "field.cattle.io/description": "main sc",
        "storageclass.beta.kubernetes.io/is-default-class": "true",
        "storageclass.kubernetes.io/is-default-class": "true"
    },
```
You can see that each Storage Class set as default has the: `metadata.annotations.storageclass.beta.kubernetes.io/is-default-class: "true"` & `metadata.annotations.storageclass.kubernetes.io/is-default-class: "true"` per: `https://your_harvester_vip_ip/v1/harvester/storage.k8s.io.storageclasses`.

**To Reproduce**
1. Prepare by having multiple StorageClasses created, via Advanced -> Storage Class
1. Set each Storage Class create as "default"
1. View the results within the index
1. View the results within the API resource:  https://your_harvester_vip_ip/v1/harvester/storage.k8s.io.storageclasses

**Environment**
 - Harvester ISO version: v1.1.0-rc1
 - Underlying Infrastructure: qemu/kvm

**Additional context**

![Screenshot from 2022-09-30 11-29-46](https://user-images.githubusercontent.com/5370752/193668873-53b81ade-2ca9-4e0a-81c8-e8b3db174d78.png)
![Screenshot from 2022-09-30 11-24-49](https://user-images.githubusercontent.com/5370752/193668877-fb4671c6-f440-4c34-bcc6-5417a42a395b.png)

irishgordo:
@chrisho & @WuJun2016 thanks for the fix!
This looks great!

I validated that this checks out on:
Harvester, ISO: `master-39c2bb2a-head`

I'll go ahead and close this out :smile:

PR #2900, opened 2022-10-09
Add storageclass validator and validate unique default

chrisho:
**Problem:**
Multiple StorageClass can be set as "default"

**Solution:**
Add a StorageClass validator to the harvester/webhook, if it is checked that a default sc already exists, it cannot be configured, need to reset the existing one.

**Related Issue:**
https://github.com/harvester/harvester/issues/2863

**Test plan:**
https://github.com/harvester/harvester/issues/2863

### DIFF
--- a/pkg/webhook/resources/storageclass/validator.go
+++ b/pkg/webhook/resources/storageclass/validator.go
@@ -1,0 +1,74 @@
+package storageclass
+
+import (
+	ctlstoragev1 "github.com/rancher/wrangler/pkg/generated/controllers/storage/v1"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/harvester/harvester/pkg/util"
+	werror "github.com/harvester/harvester/pkg/webhook/error"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+func NewValidator(storageClassCache ctlstoragev1.StorageClassCache) types.Validator {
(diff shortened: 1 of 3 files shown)


pkg/webhook/resources/storageclass/validator.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return validateSetUniqueDefault(newSC, scList)
+}
+
+func validateSetUniqueDefault(newSC *storagev1.StorageClass, scList []*storagev1.StorageClass) error {
+	// if new storage class haven't storageclass.kubernetes.io/is-default-class annotation or the value is false.
+	if v, ok := newSC.Annotations[util.AnnotationIsDefaultStorageClassName]; !ok || v == "false" {
+		return nil
+	}
+
```
futuretea, line 55: prefer to move v.storageClassCache.List logic to here (line 64)

and merge two validateSetUniqueDefault function

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -0,0 +1,74 @@
...
+	for _, sc := range scList {
+		if sc.Name == newSC.Name {
+			continue
+		}
+
+		// return set default error
+		// when find another have storageclass.kubernetes.io/is-default-class annotation and value is true.
+		if v, ok := sc.Annotations[util.AnnotationIsDefaultStorageClassName]; ok && v == "true" {
+			return werror.NewInvalidError("reset existing default storage class first", "")
```
guangbochen:
```suggestion
			return werror.NewInvalidError("default storage class %s already exists, please reset it first", sc.Name)
```

guangbochen (approved):
LGTM, and please squash the commits.

futuretea (approved):
LGTM

Issue #2892, opened 2022-10-07
[BUG] rancher-system-agent keeps showing error on a new node in an upgraded cluster


FrankYang0529:
**Describe the bug**
Upgrade a cluster from v1.0.3 to v1.1.0-rc2. Delete an agent node after the upgrade is done.
When we re-join a new node (with the same hostname) to the cluster, the `rancher-system-agent` keeps showing errors.
```
Oct 07 03:35:06 harvester-node-3 rancher-system-agent[2009]: W1007 03:35:06.728098    2009 reflector.go:443] pkg/
mod/github.com/rancher/client-go@v1.24.0-rancher1/tools/cache/reflector.go:168: watch of *v1.Secret ended with: a
n error on the server ("unable to decode an event from the watch stream: stream error: stream ID 5; INTERNAL_ERRO
R; received from peer") has prevented the request from succeeding
```

After restarting the `rancher-system-agent` on the new node, the error message changes.
```
Oct 07 04:05:55 harvester-node-3 rancher-system-agent[31948]: time="2022-10-07T04:05:55Z" level=info msg="Startin
g remote watch of plans"
Oct 07 04:08:05 harvester-node-3 rancher-system-agent[31948]: panic: error while connecting to Kubernetes cluster: Get "https://192.168.3.131/version": dial tcp 192.168.3.131:443: connect: connection timed out
Oct 07 04:08:05 harvester-node-3 rancher-system-agent[31948]: goroutine 38 [running]:
Oct 07 04:08:05 harvester-node-3 rancher-system-agent[31948]: github.com/rancher/system-agent/pkg/k8splan.(*watcher).start(0xc00010a200, {0x18bd5c0?, 0xc0001ee640})
Oct 07 04:08:05 harvester-node-3 rancher-system-agent[31948]:         /go/src/github.com/rancher/system-agent/pkg/k8splan/watcher.go:102 +0xa13
Oct 07 04:08:05 harvester-node-3 rancher-system-agent[31948]: created by github.com/rancher/system-agent/pkg/k8splan.Watch
Oct 07 04:08:05 harvester-node-3 rancher-system-agent[31948]:         /go/src/github.com/rancher/system-agent/pkg/k8splan/watcher.go:63 +0x155
Oct 07 04:08:05 harvester-node-3 systemd[1]: rancher-system-agent.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
[...]

Issue #2665, opened 2022-08-15
[BUG] reinstall 1st node


tgazik:
**Describe the bug**
I tried to reinstall 1st node (installed with mode: create) and the node after reinstallation didn`t joined existing cluster.

**To Reproduce**
1. install 3 node cluster 
2. turn off 1st node (installed with mode: create), remove it from cluster via GUI - hosts - delete host, wipe all disks (with fdisk remove partititions and mkfs.ext4 /dev/sda) and reinstall this node (install mode: join) to join existing cluster with rest 2 nodes with new hostname.
3. reinstalled node is not able to join existing cluster (rancher bootstrap finished successfully, but node is not joined).

**Expected behavior**
Reinstalled node should join existing cluster with rest 2 nodes.

**Support bundle**
can provide if needed

**Environment**
 - Harvester ISO version: v1.0.3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): tried on 2 different environments with different HW

**Additional context**
When i tried to reinstall 2nd or 3rd node (installation mode: join) they join the cluster successfully after reinstallation. The only problem is with the 1st node.

w13915984028:
@tgazik Please add the support bundle file, thanks.
 https://docs.harvesterhci.io/v1.0/troubleshooting/harvester/#generate-a-support-bundle
 
As a workaround, could you please try to install the first node again, with a different hostname and host IP ?  The failure may be related to reusing the same hostname and IP, but there are potential conflict remainings, we are investigating.

tgazik:
Tried to replicate with different hostname and IP, but the result is same, node didn`t joined the cluster.
i can see in journalctl -b -u rancherd:
```
2022-08-15T11:02:22Z" level=info msg="Bootstrapping Rancher (v2.6.4-harvester3/v1.22.12+rke2r1)"
"2022-08-15T11:02:23Z" level=info msg="Writing plan file to /var/lib/rancher/rancherd/plan/plan.json"
"2022-08-15T11:02:23Z" level=info msg="Applying plan with checksum "
"2022-08-15T11:02:23Z" level=info msg="No image provided, creating empty working directory /var/lib/rancher/rancherd/plan/work/>
"2022-08-15T11:02:23Z" level=info msg="Running command: /usr/bin/env [sh /var/lib/rancher/rancherd/install.sh]"
[...]

Issue #2470, opened 2022-07-08
[BUG] Removing the first node from a cluster, and rebuilding it, it will not rejoin the cluster.


jzmatrix:
**Describe the bug**
After creating a multi-node cluster (3 in this case), and removing the first node from the UI, and then physically rebuilding the node, it will not rejoin the cluster.   The remaining (2) nodes are still functional and accessible, but no new nodes are able to join the remaining cluster.

**To Reproduce**
1. On the first node, create a new cluster.
2. Config 'x' nodes, all joining the initial cluster (waiting for every node to go 'Ready' before adding a new node.
3. Remove the first node from the Harvester UI, leaving the remaining nodes intact and available
4. Wipe (or replace) the drives in the initial load and perform a clean node, requesting for it to join the cluster.  The node will show as 'NotReady' and will not join the cluster.

**Expected behavior**
The reloaded node should successfully join the existing cluster.

**Support bundle**
[01-supportbundle_0b532da3-637c-4b06-8b26-bbc7c63be1f1_2022-07-08T01-58-31Z.zip](https://github.com/harvester/harvester/files/9068656/01-supportbundle_0b532da3-637c-4b06-8b26-bbc7c63be1f1_2022-07-08T01-58-31Z.zip)
- The '01' support-bundle is a baseline, prior to removing the node from the cluster.

[02-supportbundle_0b532da3-637c-4b06-8b26-bbc7c63be1f1_2022-07-08T02-23-51Z.zip](https://github.com/harvester/harvester/files/9068658/02-supportbundle_0b532da3-637c-4b06-8b26-bbc7c63be1f1_2022-07-08T02-23-51Z.zip)
- The '02' support-bundle was generated after removing the node from the cluster.

**Environment**
 - Harvester ISO version:  1.0.2
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Whitebox machine, X64, 8C, 32G each node (3)

**Additional context**

w13915984028:
> 4 Wipe (or replace) the drives in the initial load and perform a clean node, requesting for it to join the cluster. The node will show as 'NotReady' and will not join the cluster.
 
 @jzmatrix 
 Could you please describe in details about how step 4 was done,  e.g. What `drives` were wiped, and how to `perform a clean node`. thanks.

jzmatrix:
[...]

PR #2918, opened 2022-10-12
fix(upgrade): remove rkeConfig

FrankYang0529:
**Problem:**
We kept `rkeConfig` in cluster CR after the upgrade, so when we tried to add a new management node in an upgraded cluster, it triggered pre/post drain again on each node.

**Solution:**
Remove `rkeConfig` after an upgrade.

**Related Issue:**
https://github.com/harvester/harvester/issues/2470
https://github.com/harvester/harvester/issues/2665
https://github.com/harvester/harvester/issues/2892

**Test plan:**
1. Create a 3-node v1.0.3 cluster.
3. Remove 1 node.
4. After the node is removed, provision a new node.
5. Check the node can join the cluster.

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -3,6 +3,7 @@ package upgrade
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sync"
 
 	semverv3 "github.com/Masterminds/semver/v3"
@@ -133,7 +134,7 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 
 	// clean upgrade repo VMs and images if a upgrade succeeds or fails.
 	if harvesterv1.UpgradeCompleted.IsTrue(upgrade) || harvesterv1.UpgradeCompleted.IsFalse(upgrade) {
-		return nil, repo.deleteVM()
+		return nil, h.cleanup(upgrade, harvesterv1.UpgradeCompleted.IsTrue(upgrade))
 	}
 
 	if harvesterv1.ImageReady.IsTrue(upgrade) && harvesterv1.RepoProvisioned.GetStatus(upgrade) == "" {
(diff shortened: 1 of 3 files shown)


pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -256,38 +257,63 @@ func (h *upgradeHandler) OnRemove(_ string, upgrade *harvesterv1.Upgrade) (*harv
...
+		return err
+	}
+
+	// remove rkeConfig in fleet-local/local cluster
+	cluster, err := h.clusterCache.Get("fleet-local", "local")
+	if err != nil {
+		return err
+	}
+	clusterToUpdate := cluster.DeepCopy()
+	clusterToUpdate.Spec.RKEConfig = nil
```
bk201:
```suggestion
	clusterToUpdate.Spec.RKEConfig = &provisioningv1.RKEConfig{}
```

assigning nil here seems to delete the cluster.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -256,38 +257,63 @@ func (h *upgradeHandler) OnRemove(_ string, upgrade *harvesterv1.Upgrade) (*harv
 	}
 
 	logrus.Debugf("Deleting upgrade %s", upgrade.Name)
+	return upgrade, h.cleanup(upgrade, true)
+}
+
+func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) error {
+	// delete vm and images
+	repo := NewUpgradeRepo(h.ctx, upgrade, h)
```
guangbochen, line 265: can we set repo to the handler instead.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -256,38 +257,63 @@ func (h *upgradeHandler) OnRemove(_ string, upgrade *harvesterv1.Upgrade) (*harv
...
-		}
-		jobs, err := h.jobCache.List(plan.Namespace, set.AsSelector())
-		if err != nil {
-			return nil, err
-		}
-		for _, job := range jobs {
-			logrus.Debugf("Deleting job %s/%s", job.Namespace, job.Name)
-			if err := h.jobClient.Delete(job.Namespace, job.Name, &metav1.DeleteOptions{}); err != nil {
-				return nil, err
+		if cleanJobs {
```
guangbochen, line 295: should we move if to the outside blocker?

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: Hi Guangbo, do you mean to set it as a parameter of the function or a field of `upgradeHandler` structure? Thank you.

bk201: Let's improve this in another following PR due to time constraints.

bk201 (approved):
LGTM, test and can join new nodes after upgrading.

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: https://github.com/harvester/harvester/issues/2939

starbops (approved):
LGTM, tested and rkeConfig removed.

guangbochen (approved):
LGTM, thanks.

PR #2923, opened 2022-10-12
fix(upgrade): make sure only one node pre/post-draining

starbops:
This is a workaround for #2907, to prevent multi-node draining simultaneously.

**Problem:**

In a multi-node setup, the upgrade might be stuck in the middle due to the RKE2 multi-node upgrading issue.

**Solution:**

Implemented a check before creating pre/post-draining hook Jobs. If there is any ongoing pre-drain/drain/post-drain stuff in the middle on any of the nodes, raise an error about it so that it can be retried at the next reconcile loop.

**Related Issue:**

#2907

**Test plan:**

1. Prepare a v1.0.3 Harvester cluster with >= 3 nodes
2. Upgrade the cluster to v1.1.0-rc2 with this PR patch
3. The upgrade should end up without problems


bk201 (approved):
LGTM, thanks for the speedy fix.

FrankYang0529 (approved):
LGTM. Thanks.

Issue #2893, opened 2022-10-07
[BUG] Upgrade stuck: kube-controller-manager fail to start


bk201:
**Describe the bug**

An upgrade is stuck in the pre-drained state and won't continue.

**To Reproduce**
1. Upgrade from v1.0.3 to v1.1.0-rc2
2. There is a chance to stuck in the pre-drained state.
![image](https://user-images.githubusercontent.com/1691518/194473786-6f871217-5cb8-4494-ac18-0b5e4780636b.png)

**Expected behavior**

The upgrade should finish.

**Support bundle**

[supportbundle_1525861a-48d6-4313-a67b-eb93485b6081_2022-10-07T05-24-13Z.zip](https://github.com/harvester/harvester/files/9730909/supportbundle_1525861a-48d6-4313-a67b-eb93485b6081_2022-10-07T05-24-13Z.zip)
[debug.tar.gz](https://github.com/harvester/harvester/files/9730911/debug.tar.gz)

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
- From rancher log:
```
rancher-68c449f596-qqhmj 2022/10/07 03:23:57 [INFO] [planner] rkecluster fleet-local/local: waiting: configuring bootstrap node(s) custom-df5877e771dd: waiting for probes: kube-controller-manager
rancher-68c449f596-qqhmj 2022/10/07 03:24:20 [INFO] [planner] rkecluster fleet-local/local: waiting: configuring bootstrap node(s) custom-df5877e771dd: waiting for probes: kube-controller-manager
```

- Check `custom-df5877e771dd` (node `harvester-vtgk8`). kube-controller-manager fails to start:
```
harvester-vtgk8:/var/log/containers # crictl ps -a | grep kube-controller-manager
3ed94fe307847       0a5067ab04e9a       17 hours ago        Exited              kube-controller-manager            2                   9bf00efca6127       kube-controller-manager-harvester-vtgk8
```

```
harvester-vtgk8:/var/log/containers # cat cloud-controller-manager-harvester-vtgk8_kube-system_cloud-controller-manager-b153f14e940006b6257cb8ea67c129837a649d10257379e5d6b43bdb72892ff0.log
2022-10-06T11:50:32.955155052Z stderr F goroutine 179 [chan receive]:
[...]

PR #2924, opened 2022-10-12
upgrade: cleanup the static pod yaml

Vicente-Cheng:
**Problem:**
Because `kube-controller-manager` need to authorize with `kube-apiserver`, `kube-controller-manager` will fail if starting before `kube-apiserver`.

**Solution:**
ensure the static pods bootup sequence

**Related Issue:**
#2893 

**Test plan:**


bk201 (approved):
LGTM.

starbops (approved):
LGTM, thanks!

guangbochen (approved):
good catch, LGTM thanks.

PR #2959, opened 2022-10-17
Upgrade: remove canal interface in ConfigMap

bk201:
Simply deleting HelmChartConfig doesn't seem to work in every situation.


guangbochen (approved):
LGTM, thanks.

Issue #2982, opened 2022-10-19
[BUG] enhance node scheduling when vm selects network


WuJun2016:
**Describe the bug**

When the vm is created, the node matched by the network may be different from the node scheduled by the final vm.

**To Reproduce**
1. go to `Cluster Networks / Config` page,  create a new Cluster Network (eg: test)
2. Create a new `network config` in the `test` Cluster Network. (Select a specific node)
![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/431ba9b2-56e7-48af-bf4d-6e0ba964ebd3)
3. go to `Network` page
4. to create a new network (e.g:  `test-untagged`), select `UntaggedNetwork` type and select `test` cluster network. click `Create` button
5. go to VM create page,    fill all required value,  Click `Networks` tab,   select `default/test-untagged` network,  click `Create` button
6. The VM is successfully created, but the scheduled node may not match the Network Config
![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/e7e942d2-4200-481d-a96a-8038bdc74ae0)

![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/ff9c6ad9-1cb0-4d8e-8728-2f93576d0238)

**Expected behavior**

The node that the VM actually schedules should match the node that the Network Config selects

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

lanfon72:
Verified this bug has been fixed.

Scheduling rule added automatically when select specific network
![image](https://user-images.githubusercontent.com/5169694/197729616-a6fcda2e-42ba-469f-b6c1-9c297bef1a45.png)


Test Information
----
* Environment: **baremetal DL360G9 4 nodes**
* Harvester Version: **master-402df54b-head**
* **ui-source** Option: **Auto**

### Verify Steps:
- Follow **Steps to reproduce** in https://github.com/harvester/harvester/issues/2982#issue-1414522175

PR #3000, opened 2022-10-20
Add node affinity to the VM

yaocw2020:
Add node affinity to let the VM be scheduled to nodes with available cluster network

Related issue: https://github.com/harvester/harvester/issues/2982


pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -3,31 +3,42 @@ package virtualmachine
...
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
```
futuretea: prefer to use logrus like other logs in the webhook server
use `logrus.Debugf` instead of `klog.Infof`

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -217,3 +247,191 @@ func (m *vmMutator) getOvercommit() (*settings.Overcommit, error) {
...
+	return append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/affinity", "value": %s}`, string(bytes))), nil
+}
+
+func (m *vmMutator) getNodeSelectorRequirementFromNetwork(network kubevirtv1.Network) (*v1.NodeSelectorRequirement, error) {
+	if network.Multus == nil {
+		return nil, nil
+	}
+
+	words := strings.Split(network.Multus.NetworkName, "/")
+	if len(words) != 2 {
```
futuretea:
```go
	var namespace, name string
	switch len(words) {
	case 1:
		namespace, name = vm.Namespace, words[0]
	case 2:
		namespace, name = words[0], words[1]
	default:
		return nil, fmt.Errorf("invalid network name %s", network.Multus.NetworkName)
	}
```

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -217,3 +247,191 @@ func (m *vmMutator) getOvercommit() (*settings.Overcommit, error) {
...
+	}
+
+	return &v1.NodeSelectorRequirement{
+		Key:      fmt.Sprintf("%s/%s", networkGroup, clusterNetwork),
+		Operator: v1.NodeSelectorOpIn,
+		Values:   []string{"true"},
+	}, nil
+}
+
+func isContainTargetNodeSelectorRequirement(term v1.NodeSelectorTerm, target v1.NodeSelectorRequirement) (bool, int) {
```
futuretea: prefer to return (int, bool)

pkg/webhook/resources/virtualmachine/mutator_test.go
```diff
@@ -127,7 +127,8 @@ func Test_virtualmachine_mutator(t *testing.T) {
 			}
 			err := clientset.Tracker().Add(settingCpy)
 			assert.Nil(t, err, "Mock resource should add into fake controller tracker")
-			mutator := NewMutator(fakeclients.HarvesterSettingCache(clientset.HarvesterhciV1beta1().Settings))
+			mutator := NewMutator(fakeclients.HarvesterSettingCache(clientset.HarvesterhciV1beta1().Settings),
+				fakeclients.NetworkAttachmentDefinitionCache(clientset.K8sCniCncfIoV1().NetworkAttachmentDefinitions))
```
futuretea, line 131: It's better to have test cases in `Test_virtualmachine_mutator`

FrankYang0529 (approved):
Test the case in https://github.com/harvester/harvester/issues/2982, and it can select the correct network. LGTM. Thanks

pkg/webhook/resources/virtualmachine/mutator_test.go
yaocw2020: It's not easy to add the unit test. Let's add it later via https://github.com/harvester/harvester/issues/3007.

futuretea (approved):
LGTM

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -217,3 +247,196 @@ func (m *vmMutator) getOvercommit() (*settings.Overcommit, error) {
...
+
+	if oldVm.Spec.Template.Spec.Affinity.String() != newVm.Spec.Template.Spec.Affinity.String() {
+		return true
+	}
+
+	oldVmNetworks, newVmNetworks := oldVm.Spec.Template.Spec.Networks, newVm.Spec.Template.Spec.Networks
+	if len(getMultusNetworkIncrement(oldVmNetworks, newVmNetworks)) != 0 ||
+		len(getMultusNetworkIncrement(newVmNetworks, oldVmNetworks)) != 0 {
+		return true
+	}
```
guangbochen: Those condition checks will bypass the VMs created in the previous version, and their affinity will never be updated until affinity rules are edited manually.
We can assume that when a VM has a NIC(except the cluster network),  the affinity must not be nil.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -217,3 +247,206 @@ func (m *vmMutator) getOvercommit() (*settings.Overcommit, error) {
...
+		return false
+	}
+
+	if isContainMultusNetwork(newVm.Spec.Template.Spec.Networks) {
+		return true
+	}
+
+	oldVmAffinity, newVmAffinity := oldVm.Spec.Template.Spec.Affinity, newVm.Spec.Template.Spec.Affinity
+	if (oldVmAffinity == nil) != (newVmAffinity == nil) ||
+		oldVmAffinity != nil && newVmAffinity != nil && oldVmAffinity.String() != newVmAffinity.String() {
```
guangbochen: can we consider splitting the check conditions here as the yare not intuitive, thanks?
additionally, the `.String()` method already checks the nil for u.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -217,3 +247,206 @@ func (m *vmMutator) getOvercommit() (*settings.Overcommit, error) {
...
+				nodeSelector.NodeSelectorTerms[i] = term
+				isChanged = true
+			}
+		}
+	}
+
+	return isChanged, nil
+}
+
+func isContainMultusNetwork(networks []kubevirtv1.Network) bool {
```
guangbochen, line 449: need to check both new and old VM network specs.

guangbochen (commented):
tested with the following 3 cases, only the last one failed:
1. create a VM before the webhook upgrade, and update the webhook server, the affinity is updated as expected.
2. create a new VM after upgrading the webook, creating and updating cases are working as expected.
3. create a new VM with multiple NICs, and then remove all of the NICs except the mgmt one, the affinity is not removed as expected. (Failed)

bk201:
If users specify a VM to run on a node, but the network doesn't span to that node, we'll see that the VM can't schedule because it can't find a good place. I guess a similar situation could happen when using PCI passthrough.
Screen Shot 2022-10-24 at 2 20 17 PM
It's OK to have such a result at this stage, but maybe we can improve more to forbid the creation in the future. Nice work!

bk201 (approved):

guangbochen (approved):
LGTM, thanks.

Issue #3073, opened 2022-10-31
[FEATURE] Upgrade: Do not fail pre-drain or post-drain jobs


bk201:
**Is your feature request related to a problem? Please describe.**

Failing pre-drain or post-drain jobs is not a good idea at this moment because:
- Job pods' log might be deleted. It's hard to debug and provide support.
- Cluster is in provisioning (provisioningv2), and it's hard for users to "resume" an upgrade.

**Describe the solution you'd like**

Do not fail the jobs by:
- Either set a very big BackoffLimit (not recommended because logs might be missing).
- Wrap the script with a wrapper.
- Other methods.

**Describe alternatives you've considered**

**Additional context**

starbops:
Question: when we have such a wrapper, i.e. `do_upgrade_node.sh`, is it still necessary/adequate to implement **endless** retry mechanisms inside the real upgrade scripts like bringing up the repo VM or downloading files using curl?

TachunLin:
Verified **fixed** on master-5a18d059-head (11/08). Close this issue.

### Result
When we set the first node to fail in post-drain according to the steps in https://github.com/harvester/harvester/pull/3102#issuecomment-1303769203 

Then we check the post-drain job of the first node, it will continuously prompt the log of retrying in 10 minutes. 
Which make the upgrade dashboard of node1 post-drain don't display `Reach the backoff limit` error message 

### Test Information 
* Test Environment: 4 nodes Harvester bare machines
* Harvester version: master-5a18d059-head (11/08)

### Verify Steps
1. Prepare a 4 nodes v1.0.3 Harvester cluster
2. Enable network on available NIC
3. Create two available VLANs
4. Upgrade from v1.0.3 to master-5a18d059-head
5. After the harvester upgraded in apply-manifest stage, the server-version is changed 
6. Make the first post-drain failed by running 
  ```
  node-0:~ # touch /tmp/elemental
  node-0:~ # mount --bind /tmp/elemental /usr/bin/elemental
  ```
7. Wait for upgrade proceed to first node in pre-drain and post-drain status 
8. Check the post-drain job log 
9. Check the upgrade dashboard status of the first node

Issue #1926, opened 2022-02-17
[Enhacement] Display intermediate jobs log of an upgrade


bk201:
**Is your feature request related to a problem? Please describe.**

NOTE: this issue is moved out of this [comment](https://github.com/harvester/harvester/issues/1852#issuecomment-1020682414).

There are some intermediate jobs during an upgrade, it would be nice to see them when needed.

**Describe the solution you'd like**

Display job logs in the GUI upgrade dialog.

**Describe alternatives you've considered**

Access jobs log manually via k8s way.

**Additional context**

yasker:
@bk201 Do you think it's necessary to include this feature in v1.0.1 upgrade path or we can wait for a later release?

yasker:
@bk201 is this done?

bk201:
Originally, we plan to leverage the logging feature to support this. But we are not going to have logging sinks inside Harvester by default (instead, just allow users to send logs to their data analysis system).

We'll see if there is a way to store logs, some job logs disappear after nodes reboot.

lanfon72:
Verified this enhancement not fully implemented.

After cluster upgraded, Log unable to be downloaded, dashborad keep polling upgradelog.

One of the response content:
```js
[...]

PR #3074, opened 2022-10-31
Upgrade: do not fail pre-drain/post-drain jobs

bk201:
**Problem:**

There are two reasons we don't want to fail a pre-drain or post-drain jobs:
- It's hard to debug. Some pods' logs can't retain. (This should not be an issue after https://github.com/harvester/harvester/issues/1926)
- Users can't resume an upgrade if jobs fail. Rancher is in its provisioning state. Thus I feel it's better to wait there for trouble shooting and then continue.

**Solution:**

- Wrap the original upgrade_node.sh with retry loop. I once consider using a huge backoff limit value in jobs, but this might create flooding pods and make it even harder to debug (old logs vanish).

**Related Issue:**
https://github.com/harvester/harvester/issues/3073

**Test plan:**

- There are some way to simulate a failing pre-drain or post-drain job.
  - Insert the `exit 1` statement to upgrade_node.sh.
  - make the `elemental` command fail (will fail the post-drain job).
    ```
    touch /tmp/fake
    mount --bind /tmp/fake /usr/bin/elemental
    ```

### DIFF
--- a/package/upgrade/do_upgrade_node.sh
+++ b/package/upgrade/do_upgrade_node.sh
@@ -1,0 +1,33 @@
+#!/bin/bash -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+retries=0
+
+until $SCRIPT_DIR/upgrade_node.sh $@; do
+  # Do not fail pre-drain/post-drain jobs
+  case $1 in
+  "pre-drain"|"post-drain")
+    ;;
+  *)
+    echo "[Error] Running \"upgrade_node.sh $*\""
+    exit 1
+    ;;
+  esac
+
+  # A fallback to exit if seeing some files
+  # This could be useful to modify a running job container and resume an upgrade.
+  if [ -e "/tmp/skip-retry-with-fail" ]; then
+    echo "Exit 1 manually."
+    exit 1
+  fi
+
+  if [ -e "/tmp/skip-retry-with-succeed" ]; then
+    echo "Exit 0 manually."
+    exit 0
+  fi
+
+  echo "[$(date)] Running \"upgrade_node.sh $*\" errors, will retry ($retries retries)..."
+  sleep 60
+  retries=$((retries+1))
+done
--- a/package/upgrade/Dockerfile
+++ b/package/upgrade/Dockerfile
@@ -15,6 +15,7 @@ RUN curl -sfL https://github.com/kubevirt/kubevirt/releases/download/v0.49.0/vir
     curl -sfL https://github.com/mikefarah/yq/releases/latest/download/yq_linux_${ARCH} -o /usr/bin/yq && chmod +x /usr/bin/yq && \
     curl -sfL https://github.com/rancher/wharfie/releases/latest/download/wharfie-amd64  -o /usr/bin/wharfie && chmod +x /usr/bin/wharfie
 
+COPY do_upgrade_node.sh /usr/local/bin/
 COPY upgrade_node.sh /usr/local/bin/
 COPY upgrade_manifests.sh /usr/local/bin/
 COPY lib.sh /usr/local/bin
--- a/pkg/controller/master/upgrade/common.go
+++ b/pkg/controller/master/upgrade/common.go
@@ -178,7 +178,7 @@ func preparePlan(upgrade *harvesterv1.Upgrade) *upgradev1.Plan {
 			},
 			Upgrade: &upgradev1.ContainerSpec{
 				Image:   fmt.Sprintf("%s:%s", upgradeImageRepository, imageVersion),
-				Command: []string{"upgrade_node.sh"},
+				Command: []string{"do_upgrade_node.sh"},
 				Args:    []string{"prepare"},
 				Env: []corev1.EnvVar{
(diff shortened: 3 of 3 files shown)


package/upgrade/do_upgrade_node.sh
```diff
@@ -0,0 +1,26 @@
...
+  esac
+
+  # A fallback to exit if seeing some files
+  if [ -e "/tmp/skip-retry-with-fail" ]; then
+    exit 1
+  fi
+
+  if [ -e "/tmp/skip-retry-with-succeed" ]; then
+    exit 0
+  fi
```
starbops: I'm looking for these two temp files but cannot find any. When are these two temp files created? Thanks.

bk201: They are just there in case we want to continue the jobs:)
So we can shell into the pod and touch the file.

starbops: Cool, thanks for the explanation!

package/upgrade/do_upgrade_node.sh
```diff
@@ -0,0 +1,26 @@
...
+
+until $SCRIPT_DIR/upgrade_node.sh $@; do
+  # Do not fail pre-drain/post-drain jobs
+  case $1 in
+    "pre-drain"|"post-drain")
+	;;
+    *)
+	exit 1
+	;;
+  esac
```
starbops: We also call the `do_upgrade_node.sh` in the image preloading phase, but that argument (`prepare`) is not considered here. If the preload somehow failed, the wrapper will exit with return code 1. Is the design intended?

bk201: I initially included the `prepare`. But considering it's still in the early stage and the user can start over and upgrade again, it's excluded.

starbops: That seems reasonable. Thanks.

package/upgrade/do_upgrade_node.sh
```diff
@@ -0,0 +1,26 @@
...
+
+  # A fallback to exit if seeing some files
+  if [ -e "/tmp/skip-retry-with-fail" ]; then
+    exit 1
```
w13915984028, line 22: please add echo something like `shortcurt with return error 1`

package/upgrade/do_upgrade_node.sh
```diff
@@ -0,0 +1,26 @@
...
+  fi
+
+  if [ -e "/tmp/skip-retry-with-succeed" ]; then
+    exit 0
```
w13915984028, line 27: similar

package/upgrade/do_upgrade_node.sh
```diff
@@ -0,0 +1,26 @@
...
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+until $SCRIPT_DIR/upgrade_node.sh $@; do
+  # Do not fail pre-drain/post-drain jobs
+  case $1 in
+    "pre-drain"|"post-drain")
+	;;
+    *)
+	exit 1
```
w13915984028: echo error before exit 1

package/upgrade/do_upgrade_node.sh
```diff
@@ -0,0 +1,26 @@
...
+  if [ -e "/tmp/skip-retry-with-fail" ]; then
+    exit 1
+  fi
+
+  if [ -e "/tmp/skip-retry-with-succeed" ]; then
+    exit 0
+  fi
+
+  echo "Running \"upgrade_node.sh $*\" errors, will retry..."
+  sleep 60
```
w13915984028, line 31: please add a counting and echo in each round
please also execute `date` to print out timestamp

w13915984028: Question: should there be an overall timeout ? e.g. 24 hours.  or try endlessly for better trouble-shooting.

bk201: Updated:
```
[Thu Nov  3 05:11:09 UTC 2022] Running "upgrade_node.sh pre-drain" errors, will retry (36 retries)...
```

package/upgrade/do_upgrade_node.sh
bk201: Fixed. Add "Exit 1 manually".

package/upgrade/do_upgrade_node.sh
bk201: Fixed!

package/upgrade/do_upgrade_node.sh
bk201: > Question: should there be an overall timeout ? e.g. 24 hours. or try endlessly for better trouble-shooting.

Better to be endless, there is a chance to fix things in the original `upgrade_node.sh` script and get things right in the next iteration.

w13915984028 (approved):
LGTM, thanks.

starbops (approved):
Tested on a 4-node cluster, looking good. LGTM, thanks!

Vicente-Cheng (approved):
LGTM, thanks!

FrankYang0529 (approved):
LGTM. Thanks.

Issue #3040, opened 2022-10-26
[BUG] The upgrade version check does not handle prerelease cases


starbops:
**Describe the bug**

Users reported in #3044 the upgrade from v1.1.0-rc3 to v1.1.0 will fail with the following message:

```
The current version v1.1.0-rc3 is less than the minimum upgradable version v1.0.3.
```

**To Reproduce**
1. Prepare a v1.1.0-rc3 Harvester cluster with any number of nodes
2. Kickstart the upgrade to v1.1.0
3. The upgrade should fail in the early stage complaining "The current version v1.1.0-rc3 is less than the minimum upgradable version v1.0.3."

**Expected behavior**

Though we do not support upgrades from a prerelease version, the version-checking logic and the error message should not be confusing.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.1.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

Workaround: ~~fake the current version `kubectl edit settings.harvesterhci.io server-version` and edit the value from a prerelease version to a formal release version, e.g. from `v1.1.0-rc3` to `v1.1.0`.~~

Please check the comment [below](https://github.com/harvester/harvester/issues/3040#issuecomment-1298057127) for the updated workaround.

starbops:
In the case of upgrading from prerelease versions, either we block such behaviors or let users take the risks doing that.

Regardless of either case, we need to warn the users, both on the documentation and the UI.

w13915984028:
It is discussed in https://github.com/harvester/harvester/pull/2829#discussion_r987636298,  

`v1.1.0` is lexically < `v1.1.0-rc*`

And @bk201 created an issue: https://github.com/harvester/harvester/issues/2877  [FEATURE] Upgrade path: tool to do semantic version comparison

w13915984028:
cc @guangbochen

irishgordo:
:eyes: - looking into this

irishgordo:
This can be easily reproduced on `rc-3` for instance.
```YAML
vm-v110-rc3-test:~ # curl http://192.168.1.44/version_v110_our_release_day_forced_upgrade_yaml_for_testing_only.yaml
apiVersion: harvesterhci.io/v1beta1
kind: Version
metadata:
  name: v1.1.0
  namespace: harvester-system
spec:
[...]

PR #3077, opened 2022-11-01
fix(upgrade): take prereleases into account

starbops:
If users got a prerelease version installed and want to upgrade to the formal release version, the version checking should let them pass. For example, upgrading from v1.1.0-rc3 to v1.1.0 should not be blocked by the version-checking mechanism, though such behaviors are not recommended.

**Problem:**

Users reported the upgrade from v1.1.0-rc3 to v1.1.0 will fail with the following message:

```
The current version v1.1.0-rc3 is less than the minimum upgradable version v1.0.3.
```

**Solution:**

This change contains two parts:
- the upgrade controller, will only take effect **after** the next upgrade
  - according to [the document](https://github.com/Masterminds/semver#working-with-prerelease-versions) of the library we leveraged, a suffix is essential to make the comparisons become prerelease-aware
- the upgrade script, takes effect in the next upgrade
  - the version sorting is good enough, just added some additional handling about rc versions

**Related Issue:**

#3040 
#3633

**Test plan:**

**Test case 1**

1. Prepare a v1.1.0-rc3 Harvester cluster with any number of nodes
2. Kickstart the upgrade to v1.1.0 (contain this PR)
3. The upgrade should continue til the end and not be blocked by the version checking

The following message should not show up in the condition of the upgrade CR:

```
The current version v1.1.0-rc3 is less than the minimum upgradable version v1.0.3.
```

The following line should not show up in the logs of any upgrade Jobs:

```
Current version is not supported. Abort the upgrade.
```

**Test case 2**

1. Prepare a v1.0.3-rc2 Harvester cluster with any number of nodes
2. Kickstart the upgrade to v1.1.0 (contain this PR)
3. The upgrade should be blocked by the version checking since v1.0.3-rc2 is older than v1.0.3, the required minimum upgradable version

The following line should show up in the logs of the apply-manifest Job:

```
Current version is not supported. Abort the upgrade.
```

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -221,8 +221,7 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 		}
 
 		logrus.Info("Check minimum upgradable version")
-		minUpgradableVersion := repoInfo.Release.MinUpgradableVersion
-		if err := isVersionUpgradable(toUpdate, minUpgradableVersion); err != nil {
+		if err := isVersionUpgradable(toUpdate.Status.PreviousVersion, repoInfo.Release.MinUpgradableVersion); err != nil {
 			setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, err.Error(), "")
 			return h.upgradeClient.Update(toUpdate)
 		}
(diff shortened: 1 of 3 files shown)


starbops:
Reopen this because we need to go through "upgrade to the same version" scenarios to verify some upgrade-related features/fixes.

bk201 (approved):
LGTM, thanks!

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -199,7 +199,8 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 		if minUpgradableVersion == "" {
 			logrus.Debug("No minimum upgradable version specified, continue the upgrading")
 		} else {
-			constraint := fmt.Sprintf(">= %s", minUpgradableVersion)
+			// to enable comparisons against prerelease versions
+			constraint := fmt.Sprintf(">= %s-z", minUpgradableVersion)
```
guangbochen: prefer to set the minUpgradableVersion to `1.2.3-0` instead according to https://github.com/Masterminds/semver#working-with-prerelease-versions
Also, would be helpful to wrap the comparison check into a method and add unit test for it, thanks.

starbops:
After discussing with @guangbochen and @bk201, we decide to go with a simpler plan: adding "-0" suffix to `minUpgradableVersion` in `package/upgrade-matrix.yaml` while cutting future pre-releases. This will lift the restriction to RC versions' upgrade and keep the code of the upgrade controller and the script untouched.

starbops:
A new issue #3633 occurred while testing upgrades from v1.1.0 to v1.1.2-rc2 after we applied the fix mentioned above.

Since adding a "-0" suffix to `minUpgradableVersion` in `package/upgrade-matrix.yaml` only partially addresses the original issue from the controller side, we're reopening this PR to solve the issue for both the controller and the scripts.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Updated with unit tests. Please take a look, thanks!

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -509,3 +492,27 @@ func getCachedRepoInfo(upgrade *harvesterv1.Upgrade) (*RepoInfo, error) {
 	}
 	return repoInfo, nil
 }
+
+func isVersionUpgradable(currentVersion, minUpgradableVersion string) error {
+	if minUpgradableVersion == "" {
+		logrus.Debug("No minimum upgradable version specified, continue the upgrading")
+	} else {
```
bk201:
```suggestion
	if minUpgradableVersion == "" {
		logrus.Debug("No minimum upgradable version specified, continue the upgrading")
		return nil
	}
```

Suggest returning here and move the else block out.

starbops: Updated, thanks!

Vicente-Cheng (approved):
LGTM, thanks for adding the test!

bk201 (approved):
LGTM, thanks for the fix!

bk201:
@mergifyio backport v1.1

Issue #3067, opened 2022-10-28
[EPIC] Add further pre and post checks in harvester upgrade


Martin-Weiss:
**Describe the bug**
Just ran into an issue with the upgrade from 1.0.3 to 1.1.0 with two different cases:

1. there was a "orphaned" machine object in fleet-default
2. the upgrade from harvester 1.0.2 to 1.0.3 was completed but the cluster status "local" has been stuck in "Provisioning" 

**Expected behavior**
There should be additonal pre and post checks before and after any upgrade to verify the cluster is 100% ok before and 100% ok after the upgrade so that no left over actions / pending actions are there that might cause issues in the future.

**Environment**
 - Harvester ISO version: 1.0.2/1.0.3/1.1.0

**Additional context**

We should check 
- [ ] timesync status on all nodes (timedatectl)
- [x] the status of all volumes (health ok or detached)  (kubectl get volumes -A)
- [x] for sufficient free disk space (/, /usr/local, /var/lib/longhorn, /var/lib/harvester/defaultdisk and other longhorn disks)
- [x] node health / all nodes ready
- [x] kubectl get clusters.cluster.x-k8s.io local -n fleet-local
- [x] kubectl get machines -n fleet-local
- [ ] conflicting customizations in /oem/* (i.e. registries.yaml)
- [ ] sufficient memory and CPU in the nodes 
  - especially in case we add additional workload like monitoring/logging
  - For live-migration cases, we need to check if the rest nodes can absorb VMs from a upgrading node.

For some we should print a warning and block the upgrade and for some others we should allow the upgrade even with warning.

Martin-Weiss:
Seems we have a "deprecated volumes check, already":
![grafik](https://user-images.githubusercontent.com/13362573/198676884-5f1d0a01-ebab-4de9-9948-dcbe40cdda9b.png)

But how to check this without the Longhorn UI available?

starbops:
When LH UI is unavailable, the following command can be used to check the status of all the LH volumes:

```
$ kubectl -n longhorn-system get lhv
NAME                                       STATE      ROBUSTNESS   SCHEDULED   SIZE          NODE     AGE
pvc-8f9a780e-17dd-41bc-8cc8-a3f0a3e71636   attached   healthy      True        53687091200   node-3   7d12h
pvc-eab05ecb-6fb8-4f3c-a1f7-17d60fd30a25   attached   healthy      True        10485760      node-3   7d12h
```

[...]

PR #3093, opened 2022-11-03
feat(upgrade): add checks

FrankYang0529:
**Enhancement:**
Add node and cluster checks before starting the upgrade.

**Related Issue:**
https://github.com/harvester/harvester/issues/3067
https://github.com/harvester/harvester/issues/3298
https://github.com/harvester/harvester/issues/3391

**Test plan:**

Case 1: A cluster with any non-ready node can't be upgraded.
1. Create a 2-node harvester cluster.
2. Stop the agent node.
3. Apply a new version.
```
kind: Version
metadata:
  name: master-head
  namespace: harvester-system
spec:
  isoURL: "http://192.168.0.181:8000/harvester-master-amd64.iso"
EOF
```
4. Click the upgrade on the dashboard.
5. Webhook should stop the upgrade because a node is not ready.

Case 2: A cluster with any maintenance node can't be upgraded.
2. Enable maintenance mode on any node.
3. Click the upgrade on the dashboard.
4. Webhook should stop the upgrade because a node is in maintenance mode.

Case 3: A cluster with any non-ready managed chart can't be upgraded.
2. Create another harvester-webhook image with a different name.
3. Change the harvester-webhook image in the cluster.
4. Check the harvester ManagedChart is not ready.
5. Click the upgrade on the dashboard.
6. Webhook should stop the upgrade because the harvester ManagedChart is not ready.

### DIFF
--- a/pkg/webhook/resources/upgrade/validator.go
+++ b/pkg/webhook/resources/upgrade/validator.go
@@ -5,8 +5,11 @@ import (
 	"strings"
 
 	longhornv1beta1 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+	fleetv1alpha1 "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
+	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	v1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
@@ -25,6 +28,7 @@ import (
 const (
 	upgradeStateLabel     = "harvesterhci.io/upgradeState"
 	skipWebhookAnnotation = "harvesterhci.io/skipWebhook"
+	managedChartNamespace = "fleet-local"
 )
 
 func NewValidator(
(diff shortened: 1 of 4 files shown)


starbops (commented):
Besides the NodeReady condition, do we also want to check if there are nodes in maintenance mode?

bk201:
@w13915984028 suggests we also check all ManagedChart, I guess it's a good idea and could we add it?

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -100,7 +108,20 @@ func (v *upgradeValidator) Create(request *types.Request, newObj runtime.Object)
...
 	}
 
-	return nil
+	cluster, err := v.clusters.Get(util.FleetLocalNamespaceName, util.LocalClusterName)
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("can't find %s/%s cluster, err: %+v", util.FleetLocalNamespaceName, util.LocalClusterName, err))
+	}
+
+	if cluster.Status.Phase != string(clusterv1alpha4.ClusterPhaseProvisioned) {
+		return werror.NewBadRequest(fmt.Sprintf("cluster %s/%s status is provisioning, please wait for it to be provisioned", util.FleetLocalNamespaceName, util.LocalClusterName))
```
w13915984028: better to replace `provisioning` with `string(cluster.Status.Phase)` to reflect the actual phase

Vicente-Cheng (approved):
LGTM! thanks

pkg/webhook/resources/upgrade/validator.go
FrankYang0529: Updated it. Thanks.

w13915984028 (approved):
LGTM, thanks.

BTW, please @ibrokethecloud also take a look, should enabled addon status also be checked? If some enabled addon has issue in deployment, will it also affect later upgrade waiting?

bk201 (approved):
LGTM

starbops (approved):
LGTM, thanks!

ibrokethecloud:
I am not sure we need to check addons. They are just pods and should be migrated to other healthy nodes in the cluster.

I would just the k8s deployment / daemonset reconcile take care of such issues. Only vm-import-controller will have a pvc claim, and that is only storing temporary data. In worst case end users can restart the import process.

guangbochen (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.1

Issue #3079, opened 2022-11-01
[BUG] Need retry logic around curl in detect_repo


starbops:
**Describe the bug**

We've spotted that failures might occur around `curl` commands in `detect_repo` during node upgrades due to network issues. This leads to the total failure of the node upgrade Pod because we have `set -e` in the upgrade scripts.

```
node-0:/home/rancher # kubectl -n harvester-system logs hvst-upgrade-52qzw-post-drain-mike-2vl7p
+++ dirname /usr/local/bin/upgrade_node.sh
++ cd /usr/local/bin
++ pwd
+ SCRIPT_DIR=/usr/local/bin
+ source /usr/local/bin/lib.sh
++ UPGRADE_NAMESPACE=harvester-system
++ UPGRADE_REPO_URL=http://upgrade-repo-hvst-upgrade-52qzw.harvester-system/harvester-iso
++ UPGRADE_REPO_VM_NAME=upgrade-repo-hvst-upgrade-52qzw
++ UPGRADE_REPO_RELEASE_FILE=http://upgrade-repo-hvst-upgrade-52qzw.harvester-system/harvester-iso/harvester-release.yaml
++ UPGRADE_REPO_SQUASHFS_IMAGE=http://upgrade-repo-hvst-upgrade-52qzw.harvester-system/harvester-iso/rootfs.squashfs
++ UPGRADE_REPO_BUNDLE_ROOT=http://upgrade-repo-hvst-upgrade-52qzw.harvester-system/harvester-iso/bundle
++ UPGRADE_REPO_BUNDLE_METADATA=http://upgrade-repo-hvst-upgrade-52qzw.harvester-system/harvester-iso/bundle/metadata.yaml
++ CACHED_BUNDLE_METADATA=
++ HOST_DIR=/host
+ UPGRADE_TMP_DIR=/host/usr/local/upgrade_tmp
+ mkdir -p /host/usr/local/upgrade_tmp
+ case $1 in
+ command_post_drain
+ wait_repo
++ get_repo_vm_status
++ kubectl get virtualmachines.kubevirt.io upgrade-repo-hvst-upgrade-52qzw -n harvester-system '-o=jsonpath={.status.printableStatus}'
+ [[ Running == \R\u\n\n\i\n\g ]]
+ curl -sfL http://upgrade-repo-hvst-upgrade-52qzw.harvester-system/harvester-iso/harvester-release.yaml
+ echo 'Wait for upgrade repo ready...'
[...]

PR #3094, opened 2022-11-03
fix(upgrade): add retries for curl downloads

starbops:

**Problem:**

Failures might occur around curl commands in `detect_repo` and many other places in the upgrade scripts during node upgrades due to network issues. This leads to a total failure of the node upgrade Pod because we have `set -e` in the upgrade scripts.

**Solution:**

Add retry for file downloading using curl.

**Related Issue:**

#3079 

**Test plan:**

1. Prepare a v1.0.3 Harvester cluster with any number of nodes
2. Kickstart the upgrade to v1.1.0 which contains the changes in this PR
3. Wait until the upgrade entering phase 4: node upgrade
4. Copy the output of the upgrade repo VM's Service resource with `kubectl -n harvester-system get svc upgrade-repo-hvst-upgrade-lbwg6 -o yaml > upgrade-repo-svc.yaml` for example
5. Remove the Service resource with `kubectl -n harvester-system delete svc upgrade-repo-hvst-upgrade-lbwg6` for example
6. Check the logs of the running post-drain Pod, there should be some `curl` retrying logs
7. Resume the upgrade by putting back the Service resource with `kubectl apply -f upgrade-repo-svc.yaml`
8. Check the logs of the running post-drain Pod, the retry should end, and the upgrade proceeds

### DIFF
--- a/package/upgrade/lib.sh
+++ b/package/upgrade/lib.sh
@@ -13,12 +13,12 @@ download_file()
   local url=$1
   local output=$2
 
-  local i=0
-  until curl -sfL "$url" -o "$output"
+  local i=1
+  until curl -sSfL "$url" -o "$output"
   do
-    i=$((i + 1))
-    echo "Cannot download the requested file \"$output\" from \"$url\", retrying ($i)..."
+    echo "Failed to download the requested file \"$output\" from \"$url\" with error code: $?, retrying ($i)..."
     sleep 10
+    i=$((i + 1))
   done
 }
 
--- a/package/upgrade/lib.sh
+++ b/package/upgrade/lib.sh
@@ -13,14 +13,8 @@ download_file()
   local url=$1
   local output=$2
 
-  local i=1
   echo "Downloading the file from \"$url\" to \"$output\"..."
-  until curl -sSfL "$url" -o "$output"
-  do
-    echo "Failed to download the requested file from \"$url\" to \"$output\" with error code: $?, retrying ($i)..."
-    sleep 10
-    i=$((i + 1))
-  done
+  curl -sSfL --retry 10 --retry-connrefused "$url" -o "$output"
 }
 
 detect_repo()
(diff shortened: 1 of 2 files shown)


package/upgrade/lib.sh
```diff
@@ -8,10 +8,24 @@ UPGRADE_REPO_BUNDLE_METADATA="$UPGRADE_REPO_URL/bundle/metadata.yaml"
 CACHED_BUNDLE_METADATA=""
 HOST_DIR="${HOST_DIR:-/host}"
 
+download_file()
+{
+  local url=$1
+  local output=$2
+
```
bk201, line 15: Let's add a statement like "Downloading xxx to yyy...".
Do we want to use the `--retry` flag of curl?

package/upgrade/lib.sh
```diff
@@ -8,10 +8,24 @@ UPGRADE_REPO_BUNDLE_METADATA="$UPGRADE_REPO_URL/bundle/metadata.yaml"
...
+  local i=1
+  until curl -sSfL "$url" -o "$output"
+  do
+    echo "Failed to download the requested file \"$output\" from \"$url\" with error code: $?, retrying ($i)..."
```
bk201: ~~Please help remove the incomplete file (if it exists).~~

bk201: Updated. I guess it's OK since the original one will be overwritten.

package/upgrade/lib.sh
starbops: Updated.
The flag does not support something like `-1` for endless retry, so we just go with the `until` loop.

starbops: Updated from endless retry to limited retry since we're gonna have job-level retry.

starbops:
Updated from endless retry to limited retry since we're gonna have job-level retry.

bk201 (approved):
LGTM

Vicente-Cheng (approved):
LGTM!

PR #3097, opened 2022-11-03
feat(vmtemplate): add with data

FrankYang0529:
**Enhancement:**
Add the `withData` field when users create a template from a VM, so system will automatically create new images from source VM volumes.

**Related Issue:**
https://github.com/harvester/harvester/issues/571

**Test plan:**

1. Create a harvester cluster and set up an image.
2. Create a `source-vm` with 2 volumes.
3. After `source-vm` provisioned, run `echo "123" > test.txt && sync`.
4. Enable developer feature on the dashboard.
5. Create a `vm-template` with data form the `source-vm`.
Screenshot 2022-11-03 at 2 12 14 PM
6. Before new images are ready, check `vm-template` ready condition is false.
7. After new images are ready, check `vm-template` ready condition is true.
8. Delete `source-vm`.
9. Create a `new-vm` from `vm-template`.
Screenshot 2022-11-03 at 2 14 58 PM
10. After `new-vm` provisioned, run `cat test.txt` and check the result is `123`.

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -9,19 +9,23 @@ import (
 	"strings"
 
 	"github.com/gorilla/mux"
+	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	"github.com/pkg/errors"
 	"github.com/rancher/apiserver/pkg/apierror"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	ctlstoragev1 "github.com/rancher/wrangler/pkg/generated/controllers/storage/v1"
 	wranglername "github.com/rancher/wrangler/pkg/name"
 	"github.com/rancher/wrangler/pkg/schemas/validation"
 	"github.com/rancher/wrangler/pkg/slice"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/utils/pointer"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	volumeapi "github.com/harvester/harvester/pkg/api/volume"
@@ -59,6 +63,9 @@ type vmActionHandler struct {
 	secretCache               ctlcorev1.SecretCache
 	virtSubresourceRestClient rest.Interface
 	virtRestClient            rest.Interface
+	vmImages                  ctlharvesterv1.VirtualMachineImageClient
+	vmImageCache              ctlharvesterv1.VirtualMachineImageCache
+	storageClassCache         ctlstoragev1.StorageClassCache
 }
 
 func (h vmActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
@@ -467,6 +474,30 @@ func getMigrationUID(vmi *kubevirtv1.VirtualMachineInstance) string {
 
 // createTemplate creates a template and version that are derived from the given virtual machine.
 func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTemplateInput) error {
+	vm, err := h.vmCache.Get(namespace, name)
+	if err != nil {
+		return err
+	}
+
+	keyPairIDs, err := getSSHKeysFromVMITemplateSpec(vm.Spec.Template)
+	if err != nil {
+		return err
+	}
+
+	vmtvName := fmt.Sprintf("%s-%s", input.Name, rand.String(5))
+	vmSourceSpec, err := h.sanitizeVirtualMachineForTemplateVersion(vmtvName, vm, input.WithData)
+	if err != nil {
+		return err
+	}
+
+	var pvcStorageClassMap map[string]string
+	if input.WithData {
+		pvcStorageClassMap, err = h.getPVCStorageClassMap(vm)
+		if err != nil {
+			return err
+		}
+	}
+
 	vmt, err := h.vmTemplateClient.Create(
 		&harvesterv1.VirtualMachineTemplate{
 			ObjectMeta: metav1.ObjectMeta{
@@ -481,19 +512,8 @@ func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTem
 		return err
 	}
 
-	vm, err := h.vmCache.Get(namespace, name)
-	if err != nil {
-		return err
-	}
-
-	keyPairIDs, err := getSSHKeysFromVMITemplateSpec(vm.Spec.Template)
-	if err != nil {
-		return err
-	}
-
 	vmID := fmt.Sprintf("%s/%s", vmt.Namespace, vmt.Name)
 
-	vmtvName := fmt.Sprintf("%s-%s", vmt.Name, rand.String(5))
 	vmtv, err := h.vmTemplateVersionClient.Create(
 		&harvesterv1.VirtualMachineTemplateVersion{
 			ObjectMeta: metav1.ObjectMeta{
@@ -503,12 +523,18 @@ func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTem
 			Spec: harvesterv1.VirtualMachineTemplateVersionSpec{
 				TemplateID:  vmID,
 				Description: fmt.Sprintf("Template drived from virtual machine [%s]", vmID),
-				VM:          sanitizeVirtualMachineForTemplateVersion(vmtvName, vm),
+				VM:          vmSourceSpec,
 				KeyPairIDs:  keyPairIDs,
 			},
 		})
 	if err != nil {
 		return err
+	}
+
+	if input.WithData {
+		if err := h.createVMImages(vmtv, vm, pvcStorageClassMap); err != nil {
+			return err
+		}
 	}
 
 	return h.createSecrets(vmtv, vm)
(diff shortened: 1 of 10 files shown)


pkg/controller/master/template/template_version_controller.go
```diff
@@ -88,6 +101,33 @@ func (h *templateVersionHandler) OnChanged(key string, tv *harvesterv1.VirtualMa
...
 
+func (h *templateVersionHandler) isVmImagesReady(tv *harvesterv1.VirtualMachineTemplateVersion) (bool, error) {
+	volumeClaimTemplatesStr, ok := tv.Spec.VM.ObjectMeta.Annotations[util.AnnotationVolumeClaimTemplates]
+	if !ok || volumeClaimTemplatesStr == "" {
+		return true, nil
+	}
+
+	var volumeClaimTemplates []corev1.PersistentVolumeClaim
+	if err := json.Unmarshal([]byte(volumeClaimTemplatesStr), &volumeClaimTemplates); err != nil {
+		return false, fmt.Errorf("can't unmarshal %s annotation, err: %+w", util.AnnotationVolumeClaimTemplates, err)
```
chrisho: May be can remove the "+" flag

pkg/indexeres/indexer.go
```diff
@@ -68,3 +77,25 @@ func VMBackupBySourceVMUID(obj *harvesterv1.VirtualMachineBackup) ([]string, err
...
 	return []string{obj.Spec.Source.Name}, nil
 }
+
+func VMTemplateVersionByImageId(obj *harvesterv1.VirtualMachineTemplateVersion) ([]string, error) {
+	volumeClaimTemplateStr, ok := obj.Spec.VM.ObjectMeta.Annotations[util.AnnotationVolumeClaimTemplates]
+	if !ok || volumeClaimTemplateStr == "" {
+		return []string{}, nil
+	}
+
+	imageIds := []string{}
```
chrisho: It's better to use `var imageIds []string` instead

pkg/indexeres/indexer.go
```diff
@@ -68,3 +77,25 @@ func VMBackupBySourceVMUID(obj *harvesterv1.VirtualMachineBackup) ([]string, err
...
+	var volumeClaimTemplates []corev1.PersistentVolumeClaim
+	if err := json.Unmarshal([]byte(volumeClaimTemplateStr), &volumeClaimTemplates); err != nil {
+		return []string{}, fmt.Errorf("can't unmarshal %s, err: %+w", util.AnnotationVolumeClaimTemplates, err)
```
chrisho: May be can remove the "+" flag

chrisho (commented):
There are a few suggests, thanks.

pkg/api/vm/handler.go
```diff
@@ -791,6 +845,50 @@ func replaceSecrets(templateVersionName string, vm *kubevirtv1.VirtualMachine) *
...
+		}
+
+		pvc, err := h.pvcCache.Get(vm.Namespace, volume.PersistentVolumeClaim.ClaimName)
+		if err != nil {
+			return nil, err
+		}
+
+		vmImageName := getTemplateVersionVmImageName(templateVersionName, index)
+		pvcName := getTemplateVersionPvcName(templateVersionName, index)
+		volumeCliamTemplates = append(volumeCliamTemplates, corev1.PersistentVolumeClaim{
```
ibrokethecloud: nit: typo in volumeClaimTemplates

pkg/api/vm/handler.go
```diff
@@ -562,6 +572,43 @@ func (h *vmActionHandler) copySecret(sourceName, targetName string, templateVers
...
+				PVCName:      volume.PersistentVolumeClaim.ClaimName,
+				PVCNamespace: vm.Namespace,
+			},
+		}
+
+		if _, err := h.vmImages.Create(vmImage); err != nil {
+			return err
+		}
+	}
+	return nil
```
ibrokethecloud: is the plan to only use the default storage class or should we be copying the storage class from the underlying pvc?

FrankYang0529: Do you mean `storageClassParameters` in VMImage?

pkg/controller/master/template/template_version_controller.go
FrankYang0529: Fixed it. Thanks.

pkg/indexeres/indexer.go
FrankYang0529: Updated it. Thanks.

pkg/indexeres/indexer.go
FrankYang0529: Updated it. Thanks.

pkg/api/vm/handler.go
FrankYang0529: Fixed it. Thanks.

pkg/indexeres/indexer.go
FrankYang0529: If I use `var imageIds []string`, lint will show me `consider pre-allocating (prealloc)`. However, I can't preallocate length here, because we may skip some `imageID` here.

chrisho (approved):
LGTM and generated VM template and versions, thanks.

pkg/api/vm/handler.go
ibrokethecloud: yes.. should we copy these?

FrankYang0529: This is a good question. I prefer not to copy it because we don't let users modify `storageClassParameters` from the dashboard or terraform. It's better to let the webhook mutator patch the default setting currently.

ibrokethecloud: in the current form new VM's launched will be on the default storage class

FrankYang0529: Hi @ibrokethecloud, I already fixed this part. Could you review it again? Thank you.

ibrokethecloud (approved):
LGTM

pkg/api/vm/handler.go
```diff
@@ -562,6 +575,60 @@ func (h *vmActionHandler) copySecret(sourceName, targetName string, templateVers
...
+func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMachineTemplateVersion, vm *kubevirtv1.VirtualMachine) error {
+	for index, volume := range vm.Spec.Template.Spec.Volumes {
+		if volume.PersistentVolumeClaim == nil {
+			continue
+		}
+		pvc, err := h.pvcCache.Get(vm.Namespace, volume.PersistentVolumeClaim.ClaimName)
+		if err != nil {
+			return err
+		}
+		storageClassName := "harvester-longhorn"
```
guangbochen: should get default SC manually as this is configurable by the users. And the provisioner has to be `driver.longhorn.io` for now.

pkg/controller/master/template/vm_image_controller.go
```diff
@@ -0,0 +1,36 @@
...
+
+	vmTemplateVersions, err := h.templateVersionCache.GetByIndex(indexeres.VMTemplateVersionByImageIDIndex, fmt.Sprintf("%s/%s", vmImage.Namespace, vmImage.Name))
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return vmImage, nil
+		}
+		return vmImage, err
+	}
+
+	for _, vmTemplateVersion := range vmTemplateVersions {
```
guangbochen, line 32: Should we skip when the templateVersion status is expected?

guangbochen:
Additionally, I found the second disk is not restored correctly from the template image(no data), only the root disk works well, below are my validation steps:
1. create a VM with two volumes.
2. format the disk with ext4 and mount it to a custom path
3. persist the mount via add it to `/etc/fstab`
4. write a data to both volume paths(one for root disk and another one for the additional volume) and run `sync`
5. take a template with data, and after both images and template are ready, create a new VM from it.

pkg/controller/master/template/vm_image_controller.go
FrankYang0529: Yes, we can skip it. Updated it. Thanks.

FrankYang0529:
> Additionally, I found the second disk is not restored correctly from the template image(no data), only the root disk works well

I tried your test steps and I can find existing data in the additional disk. Following are my steps to mount the additional disk and add data:

```
sudo mkfs.ext4 /dev/vdb
sudo mkdir /mnt/vdb
sudo mount /dev/vdb /mnt/vdb
sudo vim /mnt/vdb/test.txt
# add some data and save
sync
```

pkg/api/vm/handler.go
```diff
@@ -593,6 +617,81 @@ func (h *vmActionHandler) updateVMVolumeClaimTemplate(vm *kubevirtv1.VirtualMach
...
+					storageClassName = vmImage.Annotations[util.AnnotationStorageClassName]
+				}
+			}
+		}
+
+		sc, err := h.storageClassCache.Get(storageClassName)
+		if err != nil {
+			return pvcStorageClassMap, err
+		}
+		if sc.Provisioner != "driver.longhorn.io" {
```
futuretea: s/driver.longhorn.io/longhorntypes.LonghornDriverName/

pkg/api/vm/handler.go
```diff
@@ -593,6 +617,81 @@ func (h *vmActionHandler) updateVMVolumeClaimTemplate(vm *kubevirtv1.VirtualMach
 	return nil
 }
 
+func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (map[string]string, error) {
```
futuretea, line 633: Question: Why don't we just get the pvc's storageclass from `pvc.Spec.StorageClassName`

FrankYang0529: If a PVC is from a VMImage, its `pvc.Spec.StorageClassName` is a VMImage storage class like `longhorn-image-xxx`.

pkg/api/vm/handler.go
FrankYang0529: Updated it. Thanks.

pkg/api/vm/handler.go
futuretea: If this pvc is not from an image, should we get it from pvc.Spec.StorageClassName? not default storageclass

futuretea: This default storageclass setting should be deprecated

FrankYang0529: Yes, we can get `pvc.Spec.StorageClassName` directly when a PVC is not from an image. However, if a PVC is from an image, we may still need the default storageclass setting, because we can't guarantee `harvesterhci.io/storageClassName` annotation is in VMImage.

futuretea: we should get the default storageclass from the storageclass list
```go
	storageClasses, err := h.storageClassCache.List(labels.Everything())
	if err != nil {
		return nil, err
	}

	for _, storageClass := range storageClasses {
		if storageClass.Annotations[util.AnnotationIsDefaultStorageClassName] == "true" {

		}
	}
```

futuretea: If no `harvesterhci.io/storageClassName`  in source pvc's image, we can leave the new image's`harvesterhci.io/storageClassName` empty.
refer to https://github.com/harvester/harvester/blob/affd8002d97ea9a1ac96c03de577fe63cd751daa/pkg/webhook/resources/virtualmachineimage/mutator.go#L96
the webhook will use the default storageclass

FrankYang0529: Thanks for your information. I didn't know that we use `AnnotationIsDefaultStorageClassName` in VMImage. I will update it!

pkg/api/vm/handler.go
```diff
@@ -593,6 +618,97 @@ func (h *vmActionHandler) updateVMVolumeClaimTemplate(vm *kubevirtv1.VirtualMach
...
+					} else {
+						sc = nil
+					}
+				} else {
+					sc = nil
+				}
+			} else {
+				sc = nil
+			}
+		}
```
futuretea:
```suggestion
		pvcStorageClass, err := h.storageClassCache.Get(*pvc.Spec.StorageClassName)
		if err != nil {
			return pvcStorageClassMap, err
		}
		var sc *storagev1.StorageClass
		if _, ok := pvcStorageClass.Parameters["backingImage"]; ok {
			// If "backingImage" is set, the storage class is from a VM image.
			// We can't use it directly. We need to find storageClassName annotation.
			if imageID, ok := pvc.Annotations[util.AnnotationImageID]; ok {
				imageIDSplit := strings.Split(imageID, "/")
				if len(imageIDSplit) == 2 {
					vmImage, err := h.vmImageCache.Get(imageIDSplit[0], imageIDSplit[1])
					if err != nil {
						return pvcStorageClassMap, err
					}
					if storageClassName, ok := vmImage.Annotations[util.AnnotationStorageClassName]; ok {
						sc, err = h.storageClassCache.Get(storageClassName)
						if err != nil {
							return pvcStorageClassMap, err
						}
					}
				}
			}
		} else {
			sc = pvcStorageClass
		}
```

futuretea (approved):
LGTM，thanks!

guangbochen (approved):
LGTM, and verified additional data volume template is working as expected.

guangbochen:
@mergify backport v1.1

Issue #3104, opened 2022-11-04
[FEATURE] Upgrade: display waiting CAPI cluster logs


bk201:
**Is your feature request related to a problem? Please describe.**

We wait for CAPI to become provisioned in [upgrade_manifest.sh](https://github.com/harvester/harvester/blob/3362968b18dd3d79a6e681923de2316f77033792/package/upgrade/upgrade_manifests.sh#L245), but there is no log for it.
When the user's cluster is somehow in the Provisioning state before an upgrade, it's hard to tell the reason why upgrade_manifest.sh can't proceed.

For example, in https://github.com/harvester/harvester/issues/3041
All we can see is:
```
managedchart.management.cattle.io/harvester patched (no change)
managedchart.management.cattle.io/harvester-crd patched (no change)
managedchart.management.cattle.io/rancher-monitoring patched (no change)
managedchart.management.cattle.io/rancher-monitoring-crd patched (no change)
Upgrading Rancher
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   117  100   117    0     0  16714      0 --:--:-- --:--:-- --:--:-- 16714
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 57.9M  100 57.9M    0     0  29.1M      0  0:00:01  0:00:01 --:--:-- 29.1M
time="2022-10-26T07:15:56Z" level=info msg="Extract mapping / => /tmp/upgrade/rancher"
time="2022-10-26T07:15:56Z" level=info msg="Checking local image archives in /tmp/upgrade/images for index.docker.io/rancher/system-agent-installer-rancher:v2.6.9"
time="2022-10-26T07:15:57Z" level=info msg="Extracting file run.sh to /tmp/upgrade/rancher/run.sh"
time="2022-10-26T07:15:58Z" level=info msg="Extracting file rancher-2.6.9.tgz to /tmp/upgrade/rancher/rancher-2.6.9.tgz"
time="2022-10-26T07:15:58Z" level=info msg="Extracting file helm to /tmp/upgrade/rancher/helm"
Rancher values:
antiAffinity: required
bootstrapPassword: admin
features: multi-cluster-management=false,multi-cluster-management-agent=false
hostPort: 0
ingress:
  enabled: false
noDefaultAdmin: false
rancherImage: rancher/rancher
[...]

PR #3105, opened 2022-11-04
Upgrade: add logs when waiting CAPI cluster to be provisioned

bk201:
**Problem:**
No logs when waiting for CAPI cluster.

**Solution:**

Add some logs.

**Related Issue:**
https://github.com/harvester/harvester/issues/3104

**Test plan:**

Perform an upgrade and you could see this in apply-manifests job:
```
CAPI cluster fleet-local/local is provisioned (current generation: 54).
```
If the cluster is somehow under provisioning, you can see logs like:
```
Waiting for CAPI cluster fleet-local/local to be provisioned (current phase: Provisioning, current generation: 55)...
```


FrankYang0529 (approved):
LGTM. Thanks.

starbops (approved):
LGTM, thanks!

PR #3106, opened 2022-11-04
fix(upgrade): use loop to do the download retry

starbops:

**Problem:**

The behavior of curl's retry is not as expected in the pod. The reason might be the version (7.66).

Previous PR: #3094 

**Solution:**

Fall back to the traditional while loop.

**Related Issue:**

#3079 

**Test plan:**

Same as PR #3094

### DIFF
--- a/package/upgrade/lib.sh
+++ b/package/upgrade/lib.sh
@@ -14,7 +14,13 @@ download_file()
   local output=$2
 
   echo "Downloading the file from \"$url\" to \"$output\"..."
-  curl -sSfL --retry 10 --retry-connrefused "$url" -o "$output"
+  local i=0
+  while [[ "$i" -lt 100 ]]; do
+    curl -sSfL "$url" -o "$output" && break
+    echo "Failed to download the requested file from \"$url\" to \"$output\" with error code: $?, retrying ($i)..."
+    sleep 10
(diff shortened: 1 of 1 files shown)


package/upgrade/lib.sh
```diff
@@ -14,7 +14,13 @@ download_file()
   local output=$2
 
   echo "Downloading the file from \"$url\" to \"$output\"..."
-  curl -sSfL --retry 10 --retry-connrefused "$url" -o "$output"
+  local i=0
+  while [[ "$i" -lt 100 ]]; do
```
Vicente-Cheng, line 18: In previous PR, retry times is 10. Why change to 100 in this PR?

starbops: Because the retry algorithm of curl is exponential backoff by default, the accumulated time of 10 retries is about 1023 (1+2+4+8+...+512) seconds. 1000 (10*100) seconds is nearly the same.

Vicente-Cheng: OK, you want the retry interval is the same as previous PR.

Vicente-Cheng (approved):
LGTM, thanks.

bk201 (approved):
LGTM!

Issue #3064, opened 2022-10-28
[FEATURE] Collect plan secrets in support bundles


bk201:
**Is your feature request related to a problem? Please describe.**

We remove all secrets in a support bundle, but plan secrets help debug upgrades.

**Describe the solution you'd like**

Collect plan secrets in a support bundle, including the metadata value.
We can ignore the `.data` part since it contains tokens.
```
kubectl get secret -n fleet-local --field-selector type=rke.cattle.io/machine-plan -o yaml
```

**Describe alternatives you've considered**

**Additional context**

Considering SBK is a generic framework, we can include it for Harvester in https://github.com/rancher/support-bundle-kit/blob/master/hack/collector-harvester

irishgordo:
### Summary:
This looks good :+1: :smile: - I'll go ahead and close this out.
Tested on Harvester Version: master-7c1f4a24-head 3 Node Cluster & v0.0.12 Support-Bundle-Kit

### Strategy:
Built support-bundle, waited for zip from Harvester dashboard.
Then to prep, checked out support-bundle-kit at v0.0.12 as that's the latest tag.
Then kicked off the `make` hook only after spinning up Rancher Desktop.
Validated dapper build built everything correctly, visually audited `dist/artifacts` checked timestamps on built binaries like that of `support-bundle-kit-amd64`.
Spun down Rancher Desktop, since having it up would interfere with the embedded etcd server that gets spun up with `support-bundle-kit-amd64 simulator --bundle-path ~/path/to/unzipped/supportbundle/directory/to/be/used`.

Leveraging then the `admin.kubeconfig` that gets built from support bundle kit via:
`kubectl --kubeconfig ~/.sim/admin.kubeconfig get secret -n fleet-local --field-selector type=rke.cattle.io/machine-plan -o yaml`

Was visually able to see elements and data.
Was also verifying that `secrets.yaml` within the `yamls/namespaced/fleet-local/v1` folder in the unzipped support bundle yielded. only `machine-plan` as expected.
```
cat secrets.yaml | grep -nr "type:"
secrets.yaml:81:        f:type: {}
secrets.yaml:110:  type: rke.cattle.io/machine-plan
secrets.yaml:195:        f:type: {}
secrets.yaml:210:  type: rke.cattle.io/machine-plan
secrets.yaml:289:        f:type: {}
secrets.yaml:318:  type: rke.cattle.io/machine-plan
```
Also checked for data elements:
```
[...]

PR #3109, opened 2022-11-04
SupportBundle: add extra collectors

Vicente-Cheng:
related issues: #3064 
blocked by: https://github.com/rancher/support-bundle-kit/pull/47

After support-bundle-kit PR reviewed, this change need to bump the newer support-bundle-kit.
Also, I thought maybe we need UI to implement that what extra collectors for the support bundle? cc @bk201 @guangbochen

### DIFF
--- a/pkg/controller/master/supportbundle/manager.go
+++ b/pkg/controller/master/supportbundle/manager.go
@@ -115,6 +115,10 @@ func (m *Manager) Create(sb *harvesterv1.SupportBundle, image string, pullPolicy
 									Name:  "SUPPORT_BUNDLE_EXCLUDE_RESOURCES",
 									Value: m.getExcludeResources(),
 								},
+								{
+									Name:  "SUPPORT_BUNDLE_EXTRA_COLLECTORS",
+									Value: m.getExtraCollectors(),
+								},
 							},
 							Ports: []corev1.ContainerPort{
 								{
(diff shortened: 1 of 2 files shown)


guangbochen:
Added `require-ui` label to the issue https://github.com/harvester/harvester/issues/3064 and cc @WuJun2016, we can discuss about the required design changes next Monday if the backend PR is ready.

Vicente-Cheng:
> Added `require-ui` label to the issue #3064 and cc @WuJun2016, we can discuss about the required design changes next Monday if the backend PR is ready.

Sure, Thanks for it!

Vicente-Cheng:
After some discussion, I though we do not need UI change for it.
harvester should only support the specific `harvester` module to SB kit. 
If the LH need something specific resource to collect, we could just add it directly.

What do you think?  cc @bk201 @guangbochen

bk201:
> 

I guess it's a sane choice.

Vicente-Cheng:
Tested with SB-kit PR. I thought this PR would be ready to review.

FrankYang0529 (approved):
LGTM. Thanks.

bk201 (approved):
LGTM. The change works with `rancher/support-bundle-kit:master-head`.

guangbochen (approved):
LGTM, thanks.

PR #3114, opened 2022-11-07
Upgrade: adjust the always-retry loop interval to 10 minutes

bk201:
**Problem:**

We added a retry loop in https://github.com/harvester/harvester/pull/3074, but the retry interval is one minute and too aggressive. Prefer to adjust it to 10 minutes.

**Solution:**

Increase the retry interval.

**Related Issue:**
https://github.com/harvester/harvester/issues/3073

**Test plan:**


starbops (approved):
LGTM!

w13915984028 (approved):
LGTM, thanks.

Issue #3070, opened 2022-10-28
[BUG] user report: v1.0.3 to v1.1.0 upgrade stucks in post-drain in first node


w13915984028:
**Describe the bug**

![image](https://user-images.githubusercontent.com/31133476/198639754-858e99b7-013b-486b-a8fc-e5efb8a391d4.png)

Directly observed:

```
seeing the log messages:
+ virtctl start upgrade-repo-hvst-upgrade-xpf52 -n harvester-system
Try to bring up the upgrade repo VM...
Error starting VirtualMachine [virtualmachine.kubevirt.io](http://virtualmachine.kubevirt.io/) "upgrade-repo-hvst-upgrade-xpf52" not found
State is:
System service upgrade succeeded but  the first node failed.

Do not see also not the VM upgrade-repo-hvst-upgrade-xpf52 at all?
```
3-node cluster, each node:
```
harvester1:/usr/local # lscpu
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Byte Order:                      Little Endian
Address sizes:                   46 bits physical, 48 bits virtual
CPU(s):                          40
On-line CPU(s) list:             0-39
Thread(s) per core:              2
Core(s) per socket:              10
Socket(s):                       2
```

```
harvester1:/usr/local #  uptime
 12:43:59  up 21 days  6:23,  1 user,  load average: 1.10, 1.42, 1.20
```
```
k get pod
NAME                                                    READY   STATUS    RESTARTS        AGE
harvester-859d59d7c4-9nlk8                              0/1     Pending   0               145m
harvester-859d59d7c4-csrdf                              1/1     Running   0               153m
harvester-859d59d7c4-dlw2q                              1/1     Running   0               153m
harvester-load-balancer-686957bdfc-pktm6                1/1     Running   2 (21d ago)     35d
harvester-network-controller-22hnv                      1/1     Running   0               153m
harvester-network-controller-4h8xc                      1/1     Running   1 (143m ago)    153m
harvester-network-controller-cpskp                      1/1     Running   0               153m
harvester-network-controller-manager-7f56fd5d45-f2l45   1/1     Running   0               153m
harvester-network-controller-manager-7f56fd5d45-lq8vn   1/1     Running   0               145m
harvester-network-webhook-57f74f7568-lm4wr              1/1     Running   0               153m
harvester-node-disk-manager-5kl9p                       1/1     Running   0               153m
[...]

PR #3118, opened 2022-11-07
Upgrade: wrap the elemental CLI for fixing the upgrade failure

Vicente-Cheng:
related issues: #3070 

NOTE: we need to remove this trick when we bump latest elemental tools on later version

#### Test Plan
1. attach another disk (the size should be the same as your root disk)
2. use elemental command to install to the above disk
    - mount squashfs to any folder (e.g. /mnt)
    - do `elemental install --directory /mnt <the above attached disk>`
3. check about the disk label by `lsblk` command, you should there are duplicated disk label
4. do upgrade, then you should not fail anymore
5. check the file `grub_oem_env` (on the path `run/initramfs/cos-state`), it should contain the upgraded version

### DIFF
--- a/package/upgrade/Dockerfile
+++ b/package/upgrade/Dockerfile
@@ -1,3 +1,13 @@
+FROM registry.suse.com/bci/bci-base:15.3 AS temp_elemental
+
+RUN zypper rm -y container-suseconnect && \
+    zypper ar --priority=200 http://download.opensuse.org/distribution/leap/15.3/repo/oss repo-oss && \
+    zypper --no-gpg-checks ref && \
+    zypper in -y curl tar gzip && zypper clean -a
+
+RUN mkdir -p /tmp/elemental_cli && \
+    curl -sfL https://github.com/rancher/elemental-cli/releases/download/v0.0.14-harvester1/elemental-v0.0.14-harvester1-Linux-x86_64.tar.gz |tar -xz -C /tmp/elemental_cli
+
 FROM registry.suse.com/bci/bci-base:15.3
 
 ARG ARCH=amd64
@@ -14,6 +24,8 @@ RUN curl -sfL https://storage.googleapis.com/kubernetes-release/release/${KUBECT
 RUN curl -sfL https://github.com/kubevirt/kubevirt/releases/download/v0.49.0/virtctl-v0.49.0-linux-${ARCH} -o /usr/bin/virtctl && chmod +x /usr/bin/virtctl && \
     curl -sfL https://github.com/mikefarah/yq/releases/latest/download/yq_linux_${ARCH} -o /usr/bin/yq && chmod +x /usr/bin/yq && \
     curl -sfL https://github.com/rancher/wharfie/releases/latest/download/wharfie-amd64  -o /usr/bin/wharfie && chmod +x /usr/bin/wharfie
+
+COPY --from=temp_elemental /tmp/elemental_cli /usr/local/bin/elemental_cli
 
 COPY do_upgrade_node.sh /usr/local/bin/
 COPY upgrade_node.sh /usr/local/bin/
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -1,6 +1,7 @@
 #!/bin/bash -ex
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+ELEMENTAL_DIR="elemental_cli"
 
 source $SCRIPT_DIR/lib.sh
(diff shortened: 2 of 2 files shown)


package/upgrade/Dockerfile
```diff
@@ -23,3 +23,4 @@ COPY patch_99_custom.sh /usr/local/bin
 COPY extra_manifests /usr/local/share/extra_manifests
 COPY migrations /usr/local/share/migrations
 COPY addons /usr/local/share/addons
+COPY elemental_cli /usr/local/bin/elemental_cli
```
guangbochen: how do we get the elemental_cli and what's the current installed version?

Vicente-Cheng: This element-cli is copy from [v0.0.14-harvester1](https://github.com/rancher/elemental-cli/releases/tag/v0.0.14-harvester1).

That related discussions are:
https://github.com/rancher/elemental-cli/issues/348
https://github.com/rancher/elemental-cli/pull/350

That could resolve the issue that failure when elemental upgrade on duplicated disk label environment.

bk201: @Vicente-Cheng Let's not add the binary to the source but download it in the Dockerfile with RUN and curl.

package/upgrade/upgrade_node.sh
```diff
@@ -443,19 +445,22 @@ upgrade_os() {
...
   elemental_upgrade_log="${UPGRADE_TMP_DIR#"$HOST_DIR"}/elemental-upgrade-$(date +%Y%m%d%H%M%S).log"
   local ret=0
+  mount --bind $new_elemental_cli $target_elemental_cli
   chroot $HOST_DIR elemental upgrade --logfile "$elemental_upgrade_log" --directory ${tmp_rootfs_mount#"$HOST_DIR"} || ret=$?
   if [ "$ret" != 0 ]; then
     echo "elemental upgrade failed with return code: $ret"
     cat "$HOST_DIR$elemental_upgrade_log"
     exit "$ret"
   fi
+  umount $target_elemental_cli
```
bk201, line 460: need to umount in the trap function `clean_up_tmp_files` or perhaps copying the file to a temporary place.

starbops (approved):
LGTM, thanks!

package/upgrade/upgrade_node.sh
bk201: NVM, it's already handled.

bk201:
The GRUB menu entry is not patched. It still displays v1.0.3 after reboot. (If we have another partition labeled with `COS_STATE`). Looks like we need to fix the https://github.com/rancher/elemental-cli/blob/7aaac4aa24abf39d5bd2c4cef26141f593080819/pkg/action/upgrade.go#L313 path as well.

```
node3:~ # lsblk -o NAME,LABEL,PARTLABEL
NAME   LABEL           PARTLABEL
loop0  COS_ACTIVE
sda
├─sda1                 p.bios
├─sda2                 oem
├─sda3 COS_STATE       state
├─sda4                 recovery
├─sda5                 persistent
└─sda6                 part6
vda
├─vda1                 p.bios
├─vda2 COS_OEM         oem
├─vda3 COS_STATE       state
├─vda4 COS_RECOVERY    recovery
├─vda5 COS_PERSISTENT  persistent
└─vda6 HARV_LH_DEFAULT part6

node3:~ # mount /dev/sda3 /tmp/mnt/
node3:~ # cat /tmp/mnt/
grub_oem_env  lost+found/
node3:~ # find /tmp/mnt/
/tmp/mnt/
/tmp/mnt/grub_oem_env
/tmp/mnt/lost+found
node3:~ # cat /tmp/mnt/grub_oem_env
# GRUB Environment Block
# WARNING: Do not edit this file other than by grub2-editenv
default_menu_entry=Harvester c9ed2c9-dirty

node3:~ # cat /run/initramfs/cos-state/grub_oem_env
# GRUB Environment Block
# WARNING: Do not edit this file other than by grub2-editenv
default_menu_entry=Harvester v1.0.3

```

Vicente-Cheng:
update the above issues: https://github.com/rancher/elemental-cli/issues/356
fix patch is already merged and retag. I thought we could re-build and test again.

bk201:
Confirm the GRUB is rebranded with the new elemental-cli.

bk201 (approved):

guangbochen (approved):
LGTM, thanks.

starbops (approved):
Tested on a single-node environment with two disks attached. `vda` is the original one and `vdb` is cloned using `dd` so they both have the same partition layout. After upgrading from v1.0.3 to v1.1.0 with this PR (replacing harvester-upgrade image during the upgrade), I can see the GRUB menu is updated and the node is upgraded successfully.

Note: need to detach `vdb` to boot into the OS, otherwise, it'll crash.

LGTM, thanks!

Issue #2984, opened 2022-10-19
[BUG] Can't display monitoring dashboard after upgrade, alertmanager, prometheus and grafana monitoring pods Terminating


TachunLin:
**Describe the bug**

After upgrade from v1.0.3 to `v1.1.0-rc3` finished, we found the grafana monitoring dashboard display empty with error
![image](https://user-images.githubusercontent.com/29251855/196621376-d1840592-7324-46a9-ab4e-a691e05659bc.png)

Further check the following pods display `Terminating` status:
1. alertmanager-rancher-monitoring-alertmanager
2. prometheus-rancher-monitoring-prometheus-0 
3. rancher-monitoring-grafana-647fd577f4-mszdm

![image](https://user-images.githubusercontent.com/29251855/196623294-760be695-5a85-4130-a8d7-4e716168c305.png)

**To Reproduce**
1. Steps to reproduce the behavior:
2. Prepare a 4 nodes v1.0.3 Harvester cluster
3. Install several images
4. Create three VMs
5. Enable Network
6. Create vlan1 network
7. Shutdown all VMs
8. Upgrade to v1.1.0-rc3

**Expected behavior**

1. The Grafana motoring dashboard of cluster and VM metrics should display correctly after upgrade complete
2. The `alertmanager-rancher-monitoring-alertmanager`, `prometheus-rancher-monitoring-prometheus-0` and `rancher-monitoring-grafana-647fd577f4-mszdm` pods have correct  running status 

**Support bundle**

[supportbundle_d28e7921-aa9a-40e9-8e4f-55404b168291_2022-10-19T07-33-59Z.zip](https://github.com/harvester/harvester/files/9817913/supportbundle_d28e7921-aa9a-40e9-8e4f-55404b168291_2022-10-19T07-33-59Z.zip)

**Environment**
 - Harvester ISO version: `v1.1.0-rc3`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 4 nodes bare metal machines

**Additional context**
After the upgrade, the vip connection lost for several hours, then it come back to normal

bk201:
The pods are stuck with terminating. Take pod `prometheus-rancher-monitoring-prometheus-0` as an example (uid: `4f20ef1b-45be-431c-ae31-ebe6696135fc`).

We see these messages flooding the kuebelet log:
```
[...]

PR #3121, opened 2022-11-08
Upgrade: set provisionGeneration when reset RKEConfig

bk201:
To prevent another RKE2 server/agent restart on all nodes when an upgrade is done.

**Problem:**

We reset RKEConfig to default after an upgrade and this trigger RKE2 server/agent restarts.

**Solution:**

Fill in the right provisionGeneration so the rancher agent doesn't restart RKE2 server or agents.

**Related Issue:**
https://github.com/harvester/harvester/issues/2984

**Test plan:**

- Upgrade from v1.0.3 to master ISO (with the change in the PR) (at least 2 nodes, better 3).
- After the upgrade, monitor pods should not be stuck in terminating (like those in https://user-images.githubusercontent.com/29251855/196623294-760be695-5a85-4130-a8d7-4e716168c305.png)

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -273,9 +273,15 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
 		return err
 	}
 	clusterToUpdate := cluster.DeepCopy()
-	clusterToUpdate.Spec.RKEConfig = &provisioningv1.RKEConfig{}
+	provisionGeneration := clusterToUpdate.Spec.RKEConfig.ProvisionGeneration
+	clusterToUpdate.Spec.RKEConfig = &provisioningv1.RKEConfig{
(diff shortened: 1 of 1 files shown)


FrankYang0529:
References:
* In [rancher](https://github.com/rancher/rancher/blob/08bf3fedd660c884fab1780a172593d5ad59f7cf/pkg/provisioningv2/rke2/planner/config.go#L258-L271), it writes `ProvisionGeneration` as `restart_stamp`.
* In [system-agent](https://github.com/rancher/system-agent-installer-rke2/blob/63890afa7a6485f15988e5b25d636c10951406d2/package/run.sh#L37-L45), it read `restart_stamp` from the file and check with prior value. If they are different, it will restart rke2.

FrankYang0529 (approved):
Test upgrade from v1.0.3 to this PR. It works fine. Thank you.

guangbochen (approved):
LGTM and verified that the rke2 server would not restart again on all nodes after clean up, thanks.

starbops (approved):
Tested on 4-node upgrade, looking good. LGTM, thanks!

PR #3130, opened 2022-11-10
Release v1.1.1-rc1

bk201:
Version changes:
- UI: v1.1.1-rc1
- Installer: v1.1.1-rc1
- Upgrade matrix: minimum v1.0.3

Issue #3100, opened 2022-11-03
[BUG] Can not specify multiple vip pool CIRDs with comma separation


Martin-Weiss:
**Describe the bug**
Trying to specify multiple vip-pools in the default namespace comma separated: 

![grafik](https://user-images.githubusercontent.com/13362573/199679678-23ae8b29-4032-4de4-8790-04f4e7e21dc4.png)

**To Reproduce**
Try to specify two CIDRs with comma separated for the default namespace.

**Expected behavior**
Multiple CIDRs should be able to be specified.

**Environment**
 - Harvester ISO version: 1.1.0

Martin-Weiss:
We have multiple VLANs and we need LoadBalancer VIPs via CPI in different ones. Not sure how this should/will work from a conceptional point of view via the Harvester CPI..

VM: VLAN1, VLAN2, VLAN3 and in each VLAN we need different VIPs..

lanfon72:
Verified this bug has been fixed.

![image](https://user-images.githubusercontent.com/5169694/200796787-4ca57ef9-540b-4ea4-8974-f8f79f39ca95.png)


Test Information
----
* Environment: **qemu/KVM single node**
* Harvester Version: **master-1a7693fd-head**
* **ui-source** Option: **Auto**

### Verify Steps:
1. Install Harvester with any nodes
2. Login to Dashboard then navigate to _Advanced/Settings_
3. Edit `vip-pools` and add multiple CIDR which separated by comma (no space between)
4. CIDR should able to added

guangbochen:
reopen for doc, @futuretea can u please help to add the `vip-pools` setting to the https://docs.harvesterhci.io/v1.1/advanced/settings page, thanks.

guangbochen:
close as the doc PR is added via https://github.com/harvester/docs/pull/256.

PR #3131, opened 2022-11-10
Fix vip pools cidr test

futuretea:

**Problem:**

CI failed in PR https://github.com/harvester/harvester/pull/3130

https://drone-pr.rancher.io/harvester/harvester/2732

**Solution:**

Remove the  unnecessary error example (this error is followed by a separate use case)

Ensure that only one error example is included in a test case to avoid random errors resulting from random map reads in random order.

**Related Issue:**
https://github.com/harvester/harvester/issues/3100

**Test plan:**


guangbochen (approved):
LGTM, thanks.

bk201 (approved):
Thanks for the quick fix.

Issue #3152, opened 2022-11-15
[BUG] Upgrade: finished upgrades interfere with newer ones


bk201:
**Describe the bug**
This is spotted when debugging https://github.com/harvester/harvester/issues/2248#issuecomment-1313656552. A previous successful or failed upgrade can interfere with new upgrades.

**To Reproduce**
1. Create a failed upgrade. An easy way is to create an upgrade CR with a non-existent version.
    ```
    apiVersion: harvesterhci.io/v1beta1
    kind: Upgrade
    metadata:
      generateName: hvst-upgrade-
      namespace: harvester-system
    spec:
      version: "v8.8.8"
    ```
2. Create another valid upgrade. The upgrade may be stuck during the upgrading nodes phase, but somehow all nodes' Kubernetes runtime is upgraded to the newer version.

**Expected behavior**

The upgrade should complete, and nodes should upgrade in order.

**Support bundle**

[supportbundle_c7adddb8-a8ff-47b2-be67-bba2ba5b2fa5_2022-11-14T13-45-10Z.zip](https://github.com/harvester/harvester/files/10013378/supportbundle_c7adddb8-a8ff-47b2-be67-bba2ba5b2fa5_2022-11-14T13-45-10Z.zip)

**Environment**
 - Harvester ISO version: v1.0.3 -> v1.1.1-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

1. The cluster has 3 nodes:
    ```
    NAME              STATUS                     ROLES                       AGE    VERSION
    harvester-2mkvs   Ready,SchedulingDisabled   control-plane,etcd,master   6d4h   v1.24.7+rke2r1
    harvester-9bx6c   Ready                      control-plane,etcd,master   6d4h   v1.24.7+rke2r1    <--- second upgrade
    harvester-mrrzx   Ready                      control-plane,etcd,master   6d4h   v1.24.7+rke2r1    <--- upgrade first and done
    ```
2. Two upgrades:
    ```
    NAMESPACE          NAME                 AGE
    harvester-system   hvst-upgrade-642kq   28h  <--- create first but fail with user error (mismatched checksum)
    harvester-system   hvst-upgrade-xm6dq   28h  <--- create later, on-going.
    ```
[...]

PR #3157, opened 2022-11-16
fix(upgrade): check latest upgrade

FrankYang0529:
**Problem:**
A previous successful or failed upgrade can interfere with new upgrades.

**Solution:**
Check whether an upgrade has `harvesterhci.io/latestUpgrade: true` label.

**Related Issue:**
https://github.com/harvester/harvester/issues/3152

**Test plan:**

1. Setup a harvester v1.0.3 cluster.
2. Update `rkeConfig` in fleet-local/local `clusters.provisioning.cattle.io`.
```yaml
  rkeConfig:
    provisionGeneration: 0
    upgradeStrategy:
      controlPlaneConcurrency: "1"
      controlPlaneDrainOptions:
        deleteEmptyDirData: true
        enabled: true
        force: true
        ignoreDaemonSets: true
        postDrainHooks:
        - annotation: harvesterhci.io/post-hook
        preDrainHooks:
        - annotation: harvesterhci.io/pre-hook
      workerConcurrency: "1"
      workerDrainOptions:
        deleteEmptyDirData: true
        enabled: true
        force: true
        ignoreDaemonSets: true
        postDrainHooks:
        - annotation: harvesterhci.io/post-hook
        preDrainHooks:
        - annotation: harvesterhci.io/pre-hook
```
3. Run `python3 -m http.server` in your image folder.
4. Create a version. (Remember to change `isoURL`)
```
kind: Version
metadata:
  name: master-head
  namespace: harvester-system
spec:
  isoURL: "http://192.168.0.181:8000/harvester-master-amd64.iso"
EOF
```
5. Create an upgrade.
6. Before VMImage finish, stop the python3 server. Now, we have one failed upgrade in our cluster.
7. Start the python3 server again.
8. Create an upgrade again.
9. Before the upgrade gets into `UpgradingNodes` status, check `rkeConfig` in fleet-local/local `clusters.provisioning.cattle.io` is not changed.

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -89,6 +89,7 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 	repo := NewUpgradeRepo(h.ctx, upgrade, h)
 
 	if harvesterv1.UpgradeCompleted.GetStatus(upgrade) == "" {
+		logrus.Infof("Initialize upgrade %s/%s", upgrade.Namespace, upgrade.Name)
 		if err := h.resetLatestUpgradeLabel(upgrade.Name); err != nil {
 			return nil, err
 		}
@@ -130,6 +131,13 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 		}
 		harvesterv1.ImageReady.CreateUnknownIfNotExists(toUpdate)
 		return h.upgradeClient.Update(toUpdate)
+	}
+
+	logrus.Infof("handle upgrade %s/%s with labels %v", upgrade.Namespace, upgrade.Name, upgrade.Labels)
+
+	// only run further operations for latest upgrade
+	if upgrade.Labels == nil || upgrade.Labels[harvesterLatestUpgradeLabel] != "true" {
+		return upgrade, nil
 	}
 
 	// clean upgrade repo VMs and images if a upgrade succeeds or fails.


pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -132,6 +133,12 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 		return h.upgradeClient.Update(toUpdate)
 	}
 
+	// only run further operations for latest upgrade
+	if upgrade.Labels == nil || upgrade.Labels[harvesterLatestUpgradeLabel] != "true" {
+		return upgrade, nil
+	}
+
+	logrus.Infof("handle upgrade %s/%s", upgrade.Namespace, upgrade.Name)
```
starbops: Is it possible to move this condition checking before L91? If I understand it correctly, we only want to handle the upgrade CR which is the "latest", we can short-circuit others in the early stage.

FrankYang0529: We only want to handle latest upgrade CR. However, when users create a new upgrade, there is no `harvesterhci.io/latestUpgrade` on it, so we need L91 to initialize the label. Thanks.

starbops: Thanks for the explanation!

starbops (approved):
Tested on 4-node upgrade, looking good. LGTM, thanks!

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -132,6 +133,12 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 		return h.upgradeClient.Update(toUpdate)
 	}
 
+	// only run further operations for latest upgrade
+	if upgrade.Labels == nil || upgrade.Labels[harvesterLatestUpgradeLabel] != "true" {
```
guangbochen, line 139: prefer to add a log here with proper message and label info.

FrankYang0529: Added it. Thank you.

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM! Thanks!

Issue #3175, opened 2022-11-18
[BUG] Upgrade: post-drain job fails when upgrading to the same version


bk201:
**Describe the bug**

The post-drain job fails if upgrading the same ISO twice.

**To Reproduce**
1. Create a cluster with master ISO after v1.1.1-rc1.
2. Upgrade to the same ISO again.
3. post-drain job will keep retrying with:
```
hvst-upgrade-ss75g-post-drain-node1-fzl69 +++ VARIANT_ID=Harvester-20221109
hvst-upgrade-ss75g-post-drain-node1-fzl69 +++ GRUB_ENTRY_NAME='Harvester d0cb3f1-dirty'
hvst-upgrade-ss75g-post-drain-node1-fzl69 ++ echo Harvester d0cb3f1-dirty
hvst-upgrade-ss75g-post-drain-node1-fzl69 + CURRENT_OS_VERSION='Harvester d0cb3f1-dirty'
hvst-upgrade-ss75g-post-drain-node1-fzl69 + '[' 'Harvester d0cb3f1-dirty' = 'Harvester d0cb3f1-dirty' ']'
hvst-upgrade-ss75g-post-drain-node1-fzl69 + echo 'Skip upgrading OS. The OS version is already "Harvester d0cb3f1-dirty".'
hvst-upgrade-ss75g-post-drain-node1-fzl69 + return
hvst-upgrade-ss75g-post-drain-node1-fzl69 + clean_up_tmp_files
hvst-upgrade-ss75g-post-drain-node1-fzl69 + '[' -n '' ']'
hvst-upgrade-ss75g-post-drain-node1-fzl69 + '[' -n '' ']'
hvst-upgrade-ss75g-post-drain-node1-fzl69 + echo 'Clean up tmp files...'
hvst-upgrade-ss75g-post-drain-node1-fzl69 + [[ -n '' ]]
hvst-upgrade-ss75g-post-drain-node1-fzl69 + [[ -n '' ]]
hvst-upgrade-ss75g-post-drain-node1-fzl69 Clean up tmp files...
hvst-upgrade-ss75g-post-drain-node1-fzl69 [Fri Nov 18 05:28:47 UTC 2022] Running "upgrade_node.sh post-drain" errors, will retry after 10 minutes (0 retries)...
```
Looks like the trap threw the wrong exit code.

**Expected behavior**

The job should succeed.

**Support bundle**

**Environment**
 - Harvester ISO version: 
[...]

PR #3176, opened 2022-11-18
fix(upgrade): make sure trap will throw no errors

starbops:

**Problem:**

When upgrading with the same ISO, the cleanup trap in the upgrade_os will throw out an error that triggers the post-drain to retry again. This is because the line:

```
[[ -n "$NEW_OS_SQUASHFS_IMAGE_FILE" ]] && \rm -vf "$NEW_OS_SQUASHFS_IMAGE_FILE"
```

will short-circuit with a non-zero return code if the variable is not set.

**Solution:**

A trap should never exit with a non-zero return code, all cases should be handled carefully.

**Related Issue:**

#3175

**Test plan:**

1. Create a cluster with master ISO after v1.1.1-rc1
2. Upgrade to the same ISO again
3. Check if the post-drain job ended successfully

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -17,8 +17,14 @@ clean_up_tmp_files()
     umount $target_elemental_cli || echo "Umount $target_elemental_cli failed with return code: $?"
   fi
   echo "Clean up tmp files..."
-  [[ -n "$NEW_OS_SQUASHFS_IMAGE_FILE" ]] && \rm -vf "$NEW_OS_SQUASHFS_IMAGE_FILE"
-  [[ -n "$tmp_rootfs_squashfs" ]] && \rm -vf "$tmp_rootfs_squashfs"
+  if [ -n "$NEW_OS_SQUASHFS_IMAGE_FILE" ]; then
+    echo "Try to remove $NEW_OS_SQUASHFS_IMAGE_FILE..."
+    rm -vf "$NEW_OS_SQUASHFS_IMAGE_FILE"
+  fi
+  if [ -n "$tmp_rootfs_squashfs" ]; then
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -17,8 +17,14 @@ clean_up_tmp_files()
     umount $target_elemental_cli || echo "Umount $target_elemental_cli failed with return code: $?"
   fi
   echo "Clean up tmp files..."
-  [[ -n "$NEW_OS_SQUASHFS_IMAGE_FILE" ]] && \rm -vf "$NEW_OS_SQUASHFS_IMAGE_FILE"
-  [[ -n "$tmp_rootfs_squashfs" ]] && \rm -vf "$tmp_rootfs_squashfs"
+  if [ -n "$NEW_OS_SQUASHFS_IMAGE_FILE" ]; then
+    echo "Try to remove $NEW_OS_SQUASHFS_IMAGE_FILE..."
+    \rm -vf "$NEW_OS_SQUASHFS_IMAGE_FILE"
```
guangbochen: prefer to remove `\` if not required?

starbops: The leading backslash is intended to bypass any possible aliases for `rm`. If this is not desirable or needed, I can remove it, thanks.

guangbochen: let's use `rm` directly since it should be reserved by default, btw the CI is still failing, maybe rebase the PR to the upstream first to rerun it, thanks.

starbops: Updated, thanks!

guangbochen (approved):
LGTM, thanks.

futuretea (approved):
LGTM

bk201 (approved):
LGTM

Issue #3168, opened 2022-11-17
[BUG] Incomplete dependency chart installation after upgrading 1.0.3 to 1.1.0 and master-head.


tjjh89017:
**Describe the bug**

After upgrading Harvester from v1.0.3 to v1.1.0 or master-head, whereabouts CRD wasn't installed correctly.
Storage-network will not work properly without whereabouts CRD.
If user configure storage-network, Longhorn will not able to start instance-manager according incorrect whereabouts' crd.

**To Reproduce**
1. Install harvester v1.0.3 and upgrade to v1.1.0 or above.
2. Check Whereabouts IPPools is missing
3. `kubectl get crd | grep ippools` (we should see `ippools.whereabouts.cni.cncf.io` but we cannot)

**Expected behavior**

We have `ipools.whereabouts.cni.cncf.io` CRD

**Support bundle**

**Environment**
 - Harvester ISO version: v1.0.3 and upgrade to v1.1.0 or above
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

But we can see daemonset is installed correctly.
```
$ kubectl get daemonset -A | grep whereabouts
kube-system                harvester-whereabouts                         4         4         4       4            4           beta.kubernetes.io/arch=amd64                     4h53m
```

**Workaround**
Before configuring storage-network, please apply these [two CRDs](https://github.com/harvester/harvester/tree/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds)

If you already configured storage-network, please disable storage-network and apply these [two CRDs](https://github.com/harvester/harvester/tree/v1.1.0/deploy/charts/harvester/dependency_charts/whereabouts/crds).
And re-configrue storage-network again and check if the Longhorn instance managers are ready without errors.

tjjh89017:
add docs https://github.com/harvester/docs/pull/248

bk201:
https://github.com/harvester/harvester/pull/3181 is a workaround. I create a [following-up issue](https://github.com/harvester/harvester/issues/3185) to find the root cause.

TachunLin:
Verified **fixed** on `master-7c1f4a24-head` (11/22). Close this issue.

### Result
After successfully upgrade 3 nodes Harvester from v1.0.3 to latest master (11/22) 

1. Execute  `kubectl get crd ippools.whereabouts.cni.cncf.io` without errors. 
  ```
[...]

PR #3181, opened 2022-11-19
storagenetwork: apply whereabouts crds before using it.

tjjh89017:
apply whereabout CRDs in apply_extra_manifests stage. it will be executed before finishing the upgrade process.

**Problem:**

https://github.com/harvester/harvester/issues/3168

**Solution:**

apply_extra_manifests will apply whereabouts CRDs.

**Related Issue:**

**Test plan:**

1. Install harvester v1.0.3 with 3 nodes.
2. Upgrade to master-head with this PR.
3. `kubectl get crd ippools.whereabouts.cni.cncf.io` without errors. (We need to have this CRD)
4. create and setup storage-network in the harvester without errors.
5. Longhorn instance-managers are working.
6. Create and start a VM without errors.

### DIFF
--- a/package/upgrade/extra_manifests/whereabouts.cni.cncf.io_ippools.yaml
+++ b/package/upgrade/extra_manifests/whereabouts.cni.cncf.io_ippools.yaml
@@ -1,0 +1,71 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.1
+  creationTimestamp: null
+  name: ippools.whereabouts.cni.cncf.io
+spec:
+  group: whereabouts.cni.cncf.io
+  names:
+    kind: IPPool
+    listKind: IPPoolList
+    plural: ippools
+    singular: ippool
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: IPPool is the Schema for the ippools API
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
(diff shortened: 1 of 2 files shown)


package/upgrade/extra_manifests/whereabouts.cni.cncf.io_ippools.yaml
```diff
@@ -0,0 +1,71 @@
+
```
guangbochen, line 1: @bk201 wondering if you know whether removing this file in the future version will also delete the related CRDs?

bk201: Guess not. The script applies everything it found under `extra_manifests`.

guangbochen: great thanks.

bk201 (approved):
LGTM.

starbops (approved):
Tested on a 4-node cluster upgrade. Logs from the apply-manifest pod shows that the CRDs are actually applied:

```
2022-11-21T04:46:17.782445165Z stdout F Applying extra manifests
2022-11-21T04:46:18.040127478Z stderr F Warning: resource settings/default-data-path is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
2022-11-21T04:46:18.077077602Z stdout F setting.longhorn.io/default-data-path configured
2022-11-21T04:46:18.348019358Z stdout F customresourcedefinition.apiextensions.k8s.io/ippools.whereabouts.cni.cncf.io created
2022-11-21T04:46:18.835772863Z stdout F customresourcedefinition.apiextensions.k8s.io/overlappingrangeipreservations.whereabouts.cni.cncf.io created
```

After the upgrade succeeded:

```
$ kubectl get crd ippools.whereabouts.cni.cncf.io
NAME                              CREATED AT
ippools.whereabouts.cni.cncf.io   2022-11-21T04:46:18Z
```

LGTM, thanks!

guangbochen (approved):
LGTM, and verified that the upgrade from v1.0.3 will includes the related CRDs, thanks.

w13915984028:
Questions:
1. do we know the reason why the crd is not created when upgrading?
2. will this script be run in any version upgrading ? is it controlled to run only when `v1.0.3 -> v1.1.0/v1.1.1`
thanks.

guangbochen:
> 1. do we know the reason why the crd is not created when upgrading?

tracked via https://github.com/harvester/harvester/issues/3168#issuecomment-1321492753

> 2. will this script be run in any version upgrading ? is it controlled to run only when `v1.0.3 -> v1.1.0/v1.1.1`

it will be applied in any new upgraded version, we should consider removing those files after fixing the root cause.

Issue #2810, opened 2022-09-22
[BUG] resize of a disk in harvester_virtualmachine via terraform is returing success but does nothing


Martin-Weiss:
**Describe the bug**
Creation of a new VM with harvester_virtualmachine and then changing the size of a disk via terraform does not work as expected. Changing size via terraform apply succeeds but does nothing.

**To Reproduce**
Create a VM via terraform harvester_virtualmachine with two disks.
Then increase the size of the second disk.

**Expected behavior**
The size of the second disk should be increased.
In case this is not possible a failure should be reported.

**Environment**
 - Harvester ISO version: 1.0.3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): bare metal

w13915984028:
@futuretea Do you have new updates of this issue? thanks.

futuretea:
I will try to handle disk resize in vm controller.

TachunLin:
Verified **fixed** on `v1.1.2-rc1`. Close this issue.

### Result
Verified can pass the following test cases 

#### Case 1: Resizing a volume via Volumes page doesn't break the VM. 
We can resize volume correctly from virtual machine edit page 

#### Case 2: Can't resize volumes on a running VM. 
  When we resize the volume size in a running VM on edit yaml, it will indicate error like
  ```
  please stop the VM before resizing volumes.
  ```
  ![image](https://user-images.githubusercontent.com/29251855/220115507-ae3cb2e6-172e-4c84-a0bd-ba540a9f9852.png)
  
  
  ![image](https://user-images.githubusercontent.com/29251855/220095542-8781c019-680f-4793-9f6b-33b857101954.png)

#### Case 3: Can't resize volumes to a smaller size. 
When we resize a volume to smaller size on edit yaml, it will indicate error like 
  ```
  PVC requests storage can't be less than previous value.
  ```
  ![image](https://user-images.githubusercontent.com/29251855/220095755-9b00a448-0bbe-4872-a38f-2b9f593a48fa.png)
  
  ![image](https://user-images.githubusercontent.com/29251855/220095946-eadb5992-1071-4e43-af16-eb277ff93f18.png)

#### Case 4: We can resize volumes through VM annotation. 
We can successfully resize volume to larger size after vm stopped in edit yaml 
  ![image](https://user-images.githubusercontent.com/29251855/220115654-5c2cc866-63dc-484c-b8ce-fa9b8daa47e5.png)
  
[...]

PR #3191, opened 2022-11-22
feat(vm): support resize pvc

FrankYang0529:
**Feature:**
Support resizing PVC via VM.

**Related Issue:**
https://github.com/harvester/harvester/issues/2810

**Test plan:**

### Setup

2. Upload a VM image.

### Case 1: Resizing a volume via Volumes page doesn't break the VM.

1. Create a VM with a 10G disk.
2. Stop the VM.
3. In the Volumes page, resize the disk to 20G.
4. After the disk is resized, start the VM. We should not have any error message in `harvester/harvester` deployment.
5. Remove the VM and the disk.

### Case 2: Can't resize volumes on a running VM.

1. Create a VM with a 10G disk.
2. Update storage size to 20G in `harvesterhci.io/volumeClaimTemplates` annotation of the VM. We should get an error message like `please stop the VM before resizing volumes`.
3. Remove the VM and the disk.

### Case 3: Can't resize volumes to a smaller size.

1. Create a VM with a 10G disk.
2. Update storage size to 5G in `harvesterhci.io/volumeClaimTemplates` annotation of the VM. We should get an error message like `PVC requests storage can't be less than previous value`.
3. Remove the VM and the disk.

### Case 4: We can resize volumes through VM annotation.

1. Create a VM with a 10G disk.
2. Stop the VM.
3. Update storage size to 20G in `harvesterhci.io/volumeClaimTemplates` annotation of the VM. We should not get any error in this step.
4. After the disk is resized, start the VM. VM can run successfully.

### DIFF
--- a/pkg/controller/master/virtualmachine/vm_controller.go
+++ b/pkg/controller/master/virtualmachine/vm_controller.go
@@ -51,14 +51,34 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
 	if err := json.Unmarshal([]byte(volumeClaimTemplates), &pvcs); err != nil {
 		return nil, err
 	}
-	for _, pvc := range pvcs {
-		pvc.Namespace = vm.Namespace
-		if _, err := h.pvcCache.Get(vm.Namespace, pvc.Name); apierrors.IsNotFound(err) {
-			if _, err := h.pvcClient.Create(pvc); err != nil {
+
+	var (
+		pvc *corev1.PersistentVolumeClaim
+		err error
+	)
+	for _, pvcAnno := range pvcs {
+		pvcAnno.Namespace = vm.Namespace
+		if pvc, err = h.pvcCache.Get(vm.Namespace, pvcAnno.Name); apierrors.IsNotFound(err) {
+			if _, err = h.pvcClient.Create(pvcAnno); err != nil {
 				return nil, err
 			}
+			continue
 		} else if err != nil {
 			return nil, err
+		}
+
+		// Users also can resize volumes through Volumes page. In that case, we can't track the update in VM annotation.
+		// If storage request in the VM annotation is smaller than or equal to the real PVC, we skip it.
+		if pvcAnno.Spec.Resources.Requests.Storage().Cmp(*pvc.Spec.Resources.Requests.Storage()) <= 0 {
+			continue
+		}
+
+		toUpdate := pvc.DeepCopy()
+		toUpdate.Spec.Resources.Requests = pvcAnno.Spec.Resources.Requests
+		if !reflect.DeepEqual(toUpdate, pvc) {
+			if _, err = h.pvcClient.Update(toUpdate); err != nil {
+				return nil, err
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


FrankYang0529:
The CodeFactor error is from complex method `SetOwnerOfPVCs`. I don't touch the function in this PR, so please ignore it. Thanks.

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -51,15 +51,35 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
...
 				return nil, err
 			}
+			continue
 		} else if err != nil {
 			return nil, err
 		}
+
+		// Users also can resize volumes through Volumes page. In that case, we can't track the update in VM annotation.
+		// If storage request in the VM annotation is smaller than the real PVC, we skip it.
+		if pvcAnno.Spec.Resources.Requests.Storage().Cmp(*pvc.Spec.Resources.Requests.Storage()) == -1 {
```
futuretea: should be <= 0?

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -138,6 +140,56 @@ func (v *vmValidator) checkVolumeClaimTemplatesAnnotation(vm *kubevirtv1.Virtual
...
+	if err := json.Unmarshal([]byte(oldVM.Annotations[util.AnnotationVolumeClaimTemplates]), &oldPvcs); err != nil {
+		return werror.NewInvalidError(fmt.Sprintf("failed to unmarshal %s", oldVM.Annotations[util.AnnotationVolumeClaimTemplates]), fmt.Sprintf("metadata.annotations.%s", util.AnnotationVolumeClaimTemplates))
+	}
+	if err := json.Unmarshal([]byte(newVM.Annotations[util.AnnotationVolumeClaimTemplates]), &newPvcs); err != nil {
+		return werror.NewInvalidError(fmt.Sprintf("failed to unmarshal %s", newVM.Annotations[util.AnnotationVolumeClaimTemplates]), fmt.Sprintf("metadata.annotations.%s", util.AnnotationVolumeClaimTemplates))
+	}
+
+	oldPvcMap, newPvcMap := map[string]*corev1.PersistentVolumeClaim{}, map[string]*corev1.PersistentVolumeClaim{}
+	for _, pvc := range oldPvcs {
+		oldPvcMap[pvc.Name] = pvc
```
futuretea, line 158: maybe need to use Namespace/Name as the map key

FrankYang0529: A pod can't use PVC in different namespaces, so I think using PVC name is ok here.

Ref: https://discuss.kubernetes.io/t/accessing-persistent-volumes-in-pods-across-differernt-namespaces/19703

pkg/controller/master/virtualmachine/vm_controller.go
FrankYang0529: Updated it. Thank you.

futuretea (approved):
LGTM. Thanks! tested follow the `Test Plan`

bk201 (approved):

guangbochen (approved):
LGTM, thanks.

FrankYang0529:
@Mergifyio backport v1.1

Issue #3221, opened 2022-11-29
[BUG] Upgrade from v1.1.0 to v1.1.1 stuck in ErrApplied with harvester-longhorn StorageClass


guangbochen:
**Describe the bug**
Upgrade from v1.1.0 to v1.1.1 stuck in ErrApplied with harvester-longhorn StorageClass
```
bundleState: ErrApplied
      message: 'cannot patch "harvester-longhorn" with kind StorageClass: admission
        webhook "validator.harvesterhci.io" denied the request: default storage class
        %!s(MISSING) already exists, please reset it first'
```

**To Reproduce**
1. Create a new storageClass via the `Advanced` tab and set it as the default one.
2. Upgrade existing v1.1.0 cluster to v1.1.1.
3. The upgrade is stuck in `Upgrading System Service` with 50% status.
4. Check the upgrade yaml via  `$ks get upgrades.harvesterhci.io hvst-upgrade-d54pm -oyaml` and it was stuck with `Current version: 1.1.1, Current state: ErrApplied, Current generation: 9`
5. Check the Harvester management chart `$ks get managedcharts -n fleet-local harvester -oyaml | less` and it was stuck with the above StorageClass webhook denied error.

**Expected behavior**
The upgrade should be continued successfully with other configured default storage class.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.1.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Baremental

**Additional context**

guangbochen:
Workaround:
1. go to the `Advanced` > `Storage Classes` page
2. set the `harvester-longhorn` as the default storage class first.

bk201:
Prefer to test this after https://github.com/harvester/harvester-installer/pull/387 is merged. Thanks!

bk201:
Test plans:
## case 1
- Upgrade from v1.0.3 to master and check storage class:
```
$ kubectl get sc
NAME                           PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
harvester-longhorn (default)   driver.longhorn.io   Delete          Immediate           true                   37s
longhorn                       driver.longhorn.io   Delete          Immediate           true                   37s
```
A new `harvester-longhorn` storage class is created as the default storage class, and the `longhorn` storage class still exists.

## case 2
- Create a new v1.1.0 cluster.
[...]

PR #3225, opened 2022-11-30
fix storage class issue when upgrading

chrisho:

**Problem:**
#3221 

**Solution:**
Add validation sc to the upgrade_harvester function in upgrade_manifest.sh

**Related Issue:**
#3221 

**Test plan:**
#3221

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -330,6 +330,12 @@ jobs:
             cpu: 250m
             memory: 128Mi
 
+storageClass:
+  ## Specify the default Storage Class of harvester-longhorn.
+  ## Will be set "false" when upgrading for existing default Storage Class.
+  ## defaults to "true".
+  defaultStorageClass: true
+
 harvester-network-controller:
   ## Specify to install harvester network controller,
   ## defaults to "false".
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -394,6 +394,12 @@ EOF
 
   upgrade_managed_chart_from_version $UPGRADE_PREVIOUS_VERSION harvester harvester.yaml
   NEW_VERSION=$REPO_HARVESTER_CHART_VERSION yq e '.spec.version = strenv(NEW_VERSION)' harvester.yaml -i
+
+  local sc=$(kubectl get sc -o json | jq '.items[] | select(.metadata.annotations."storageclass.kubernetes.io/is-default-class" == "true" and .metadata.name != "harvester-longhorn")')
+  if [ -n "$sc" ] && [ "$UPGRADE_PREVIOUS_VERSION" != "v1.0.3" ]; then
+      yq e '.spec.values.storageClass.defaultStorageClass = false' -i harvester.yaml
+  fi
+
   kubectl apply -f ./harvester.yaml
 
   pause_managed_chart harvester "false"
--- a/deploy/charts/harvester/templates/harvester-storageclass.yaml
+++ b/deploy/charts/harvester/templates/harvester-storageclass.yaml
@@ -4,8 +4,10 @@ kind: StorageClass
 apiVersion: storage.k8s.io/v1
 metadata:
   name: harvester-longhorn
+{{- if .Values.storageClass.defaultStorageClass }}
   annotations:
     storageclass.kubernetes.io/is-default-class: "true"
+{{- end }}
 provisioner: driver.longhorn.io
 allowVolumeExpansion: true
 reclaimPolicy: "{{ .Values.longhorn.persistence.reclaimPolicy }}"


package/upgrade/upgrade_manifests.sh
```diff
@@ -405,6 +415,21 @@ EOF
   wait_kubevirt harvester-system kubevirt $REPO_KUBEVIRT_VERSION
 }
 
+check_has_harvester_longhorn_sc() {
```
Vicente-Cheng: Are the following two functions not used?

package/upgrade/upgrade_manifests.sh
```diff
@@ -394,6 +394,16 @@ EOF
 
   upgrade_managed_chart_from_version $UPGRADE_PREVIOUS_VERSION harvester harvester.yaml
   NEW_VERSION=$REPO_HARVESTER_CHART_VERSION yq e '.spec.version = strenv(NEW_VERSION)' harvester.yaml -i
+
+  if [ "$UPGRADE_PREVIOUS_VERSION" == "v1.0.3" ]; then
+    yq e '.spec.values.longhorn.createStorageClass = true' -i harvester.yaml
+  fi
+
+  local sc=$(kubectl get sc -o json | jq '.items[] | select(.metadata.annotations."storageclass.kubernetes.io/is-default-class" == "true")')
```
Vicente-Cheng: Do we only check any `is-default-class`?
Or do we also need to check the `default` SC is not `harvester-longhorn`?

package/upgrade/upgrade_manifests.sh
chrisho: The next commit will remove.

package/upgrade/upgrade_manifests.sh
chrisho: > Do we only check any `is-default-class`? Or do we also need to check the `default` SC is not `harvester-longhorn`?

please wait, I will rewrite it in the next commit.

deploy/charts/harvester/values.yaml
```diff
@@ -394,6 +394,8 @@ longhorn:
     ## Specify the concurrent automatic engine upgrade per node limit
     concurrentAutomaticEngineUpgradePerNodeLimit: 3
 
+  defaultStorageClass: true
```
bk201: Putting a value under `.longhorn` generally means we want to pass the variable to the Longhorn dependency chart. Since the variable is used in the Harvester chart, perhaps moving it to the upper level?

package/upgrade/upgrade_manifests.sh
```diff
@@ -394,6 +394,12 @@ EOF
 
   upgrade_managed_chart_from_version $UPGRADE_PREVIOUS_VERSION harvester harvester.yaml
   NEW_VERSION=$REPO_HARVESTER_CHART_VERSION yq e '.spec.version = strenv(NEW_VERSION)' harvester.yaml -i
+
+  local sc=$(kubectl get sc -o json | jq '.items[] | select(.metadata.annotations."storageclass.kubernetes.io/is-default-class" == "true" and .metadata.name != "harvester-longhorn")')
+  if [ -n "$sc" and "$UPGRADE_PREVIOUS_VERSION" != "v1.0.3" ]; then
```
bk201:
```suggestion
  if [ -n "$sc" -a "$UPGRADE_PREVIOUS_VERSION" != "v1.0.3" ]; then
```
`and` is not a bash operator.

bk201:
@chrisho CodeFactor's suggestion is better. Let's try it. Thanks!

bk201 (approved):
LGTM

Vicente-Cheng (approved):
update with the following SC (v1.1.0 environment):
```
NAME                           PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
harvester-longhorn             driver.longhorn.io   Delete          Immediate           true                   9h
longhorn                       driver.longhorn.io   Delete          Immediate           true                   9h
longhorn-harvester-iso-qvr9k   driver.longhorn.io   Delete          Immediate           true                   16s
longhorn-ubuntu2004            driver.longhorn.io   Delete          Immediate           true                   159m
upgrade-breaker (default)      driver.longhorn.io   Delete          Immediate           true                   159m
```
And the upgrade was completed.

LGTM, thanks!

guangbochen (approved):
LGTM and verified that upgrade from v1.1.0 to v1.1.1-head works, @chrisho can you please help to squash the commit, thanks.

starbops (approved):
Tested upgrade from v1.1.0 to master-head successfully. LGTM, thanks!

```
$ k get sc
NAME                           PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION
harvester-longhorn             driver.longhorn.io   Delete          Immediate           true
longhorn                       driver.longhorn.io   Delete          Immediate           true
longhorn-harvester-iso-fpd8g   driver.longhorn.io   Delete          Immediate           true
test (default)                 driver.longhorn.io   Delete          Immediate           true
```

```
$ k -n fleet-local get managedcharts harvester -o yaml | yq '.spec.values | with_entries(select(.key | test("storageClass")))'
storageClass:
  defaultStorageClass: false
```

Issue #2226, opened 2022-05-06
[FEATURE] Alternate endpoint to serve harvester IPXE artifacts


ibrokethecloud:
**Is your feature request related to a problem? Please describe.**

Currently Harvester IPXE artifacts are served via https://releases.rancher.com

After a change last year, which involved migrating the DNS hosting and SSL termination https://github.com/rancher/rancher/issues/35419 IPXE booting on public cloud providers like Equinix metal is broken.

This is likely because the cert chain is too large, and how IPXE handles the cert chain and more details can be found here: https://github.com/ipxe/ipxe/pull/116

Some additional context: https://github.com/flatcar-linux/Flatcar/issues/527

**Describe the solution you'd like**

A different artifact endpoint say `releases.harvesterhci.io` pointing to the same backend but serving a different (smaller) cert chain.

**Describe alternatives you've considered**

**Additional context**

antonym:
Can confirm... unable to boot kernels directly from [releases.rancher.com](https://releases.rancher.com) endpoint via iPXE.

hrittikhere:
Looking for updates. I can't boot via iPXE too

antonym:
Ended up mirroring the release as a GitHub release which lets me boot from iPXE.

hwinkel:
hey, it's vital to bring up a Harvester Cluster in our EQXM Locations. Can you fix this please otherwise it stucks

hwinkel:
> Ended up mirroring the release as a GitHub release which lets me boot from iPXE.

can you provide some hints how to do it?

antonym:
I have a release mirror here and iPXE seems to be fine with it downloading it from Github Releases:

https://github.com/netbootxyz/asset-mirror/releases/tag/v1.0.1-e4b9482d

The code to create the release is [here](https://github.com/netbootxyz/asset-mirror/tree/harvester), all it's doing is downloading it from releases.rancher.com/harvester and pushing it into a Github release.

hwinkel:
many thanks, I will try the Release URLs for a trial and trial only.

ibrokethecloud:
We are working to internally address the hosting aspect.

For now I have set up a sample endpoint for testing `https://harvester.ibrokethe.cloud`

A sample script leveraging the same to boot in equinix metal is available here: https://github.com/ibrokethecloud/custom_pxe/blob/master/mac.ipxe

hrittikhere:
> We are working to internally address the hosting aspect.
> 
[...]

PR #3241, opened 2022-12-04
publish kernel/initrd artifacts to GH releases

ibrokethecloud:

**Problem:**

IPXE boot using releases.rancher.com cert chain currently fails due to the length of the cert chain.

**Solution:**

We should publish the kernel and initrd artifacts to GH releases.
As tested the certchain on GH releases endpoints allows us PXE boot devices successfully.

We only need the kernel / initrd as once the kernel is booted, the rootfs can be successfully pulled from releases.rancher.com and harvester installer can also download the iso from releases.rancher.com

**Related Issue:**
https://github.com/harvester/harvester/issues/2226

**Test plan:**

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -381,6 +381,23 @@ steps:
     event:
       - push
 
+- name: upload_kernel_initrd_releases
+  image: plugins/github-release
+  settings:
+    prerelease: true
+    checksum:
+      - sha256
+    files:
+    - "dist/artifacts/harvester*initrd-amd64"
+    - "dist/artifacts/harvester*vmlinuz-amd64"
+    api_key:
+      from_secret: gh_token
+  when:
+    instance:
+      - drone-publish.rancher.io
+    event:
+      - tag
+
 trigger:
   event:
     - tag
--- a/.drone.yml
+++ b/.drone.yml
@@ -385,6 +385,7 @@ steps:
   image: plugins/github-release
   settings:
     prerelease: true
+    draft: true
     checksum:
       - sha256
     files:


futuretea (approved):
LGTM

.drone.yml
```diff
@@ -381,6 +381,23 @@ steps:
...
 
+- name: upload_kernel_initrd_releases
+  image: plugins/github-release
+  settings:
+    prerelease: true
+    checksum:
+      - sha256
+    files:
+    - "dist/artifacts/harvester*initrd-amd64"
+    - "dist/artifacts/harvester*kernel-amd64"
```
Vicente-Cheng: Seems we do not build the `harvester*kernel-amd64`?
Did you mean the `harvester*vmlinuz-amd64`?

For PXE boot, we may need vmlinuz and initrd.

ibrokethecloud: good spot.. let me fix it up

Vicente-Cheng (approved):
LGTM, thanks!

.drone.yml
```diff
@@ -381,6 +381,23 @@ steps:
     event:
       - push
 
+- name: upload_kernel_initrd_releases
+  image: plugins/github-release
+  settings:
+    prerelease: true
```
bk201, line 387: As discussed in DM, we also need the `draft` setting set to `true` here.
Usually, we tag first and publish the corresponding release if everything goes well.

ibrokethecloud: thanks for the spot. this is now fixed @bk201

guangbochen (approved):
LGTM, and will need to backport to the `v1.1` branch.

Issue #3052, opened 2022-10-27
[FEATURE] Storage Tiering support of the Harvester CSI Driver


dmytrobabeshko:
Hello,

Storage tiering was introduced in Harveser v1.1.0 recently https://github.com/harvester/harvester/issues/2147

Currently, the `harveste-csi-driver` doesn't have the option to have several storage classes which are bound to different storage classes in the harvester.

Currently, the name of the storage class used by the driver is taken from env variable from `harvester-csi-driver-pod`.
https://github.com/harvester/harvester-csi-driver/blob/0f6455e0bc7a4e53dd01ab811eab37182ba6030e/main.go#L39

Thank you in advance!

guangbochen:
this is a known limitation of the v1.1.0 version, we will consider adding support in its future releases, and transfer the issue to the harvester repo.

Vicente-Cheng:
Update the user story and controller flow:

Users want to create volumes with different purposes (which means different performance/capability requirements) on the `harvester` cluster at this guest cluster.

With v1.1.0, we could create different tiers (StorageClass) on the harvester. Now, `harvester-CSI-driver` should support creating PVC on the specific tier on the host.
For example, the user needs 2 replicas for some data, which only needs lower data reliability.
We could create a StorageClass for it like following:
```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: harvester-replica-2
provisioner: driver.harvesterhci.io
reclaimPolicy: Delete
volumeBindingMode: Immediate
parameters:
  hostTierName: replica-2
```
Then, the user could use this StorageClass `harvester-replica-2` to create PVC with only 2 replica.
If the user wants to create the default tier on the host, use the default StorageClass `harvester` or simply not specify any for it.

The controller flow:

PVC StorageClass would be set up from StorageClass parameter `hostTierName` or controller config `host-storage-class`.
The prioritize is SC parameter > controller config.
If w/o any definition, PVC would use the default StorageClass on the host cluster.

cc @bk201, @guangbochen

bk201:
@guangbochen @yasker Does this sound good to you? 
- User creates a new storage class with `paramters.hostTierName` set to an existing storage class in the host cluster.
- User then creates volumes from the new storage class.

yasker:
Sounds good to me. Also need Guangbo's input. 

[...]

PR #3318, opened 2022-12-28
HEP: add HEP for various StorageClass supported on Harvester CSI Driver [CI SKIP]

Vicente-Cheng:
related issues: https://github.com/harvester/harvester/issues/3052

### DIFF
--- a/enhancements/20221228-various-storage-class-support.md
+++ b/enhancements/20221228-various-storage-class-support.md
@@ -1,0 +1,114 @@
+# Various StorageClass support of Harvester CSI Driver
+
+## Summary
+
+Harvester supports the user-defined StorageClass after v1.1.0. Because of that, Harvester CSI Driver should also let users use the custom StorageClass mentioned above.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/3052
+
+## Motivation
+
+### Goals
+
+The user using the Rancher integration with a Harvester Cluster (called Guest Cluster) can create the volume with the different StorageClass associated with StorageClass on the Host Harvester Cluster.
+
+## Proposal
+
+We will implement the mechanism to get the StorageClass parameter from the Guest Cluster and try to create it on the Host Harvester Cluster.
+
+### User Stories
+
+#### General use cases
+
+Generally, users use the default StorageClass (`harvester`). The default StorageClass has three replicas, which means higher reliability. Users can create a volume with the default StorageClass (`harvester`) and use it without any additional configuration.
+
+#### More specific goals
+
+Sometimes users will have different purposes for storage volume. Like `data archiving` or `Storage tiering`. For the above requirements, users can change the data locality to HDD. Users can complete it by creating a new StorageClass and configuring its data locality to HDD. For performance purposes, users can reduce the number of replicas (remember, the default replica number is three). Reducing the replica number increases the volume's performance, but results in some lost data reliability. For example, users can create a new StorageClass and configure its replica number from 3 to 2 and make the storage volume with this StorageClass to get better performance. 
+
+### User Experience In Detail
+
+- Harvester CSI Driver no longer hardcodes the `longhorn` StorageClass name when creating volume from the Guest Cluster. Furthermore, The default StorageClass is associated with the `Default` StorageClass on the Host Harvester Cluster.
+
+- Users need to create StorageClass with the parameter `hostStorageClass` on the Guest Cluster. This parameter is associated with the StorageClass on the Host Harvester Cluster.
+
+### API changes
+
+No
+
+## Design
+
+### Implementation Overview
+
+PVC StorageClass will reference the `hostStorageClass`, which is from the StorageClass parameter or controller config hostStorageClass. The priority is StorageClass parameter > controller config. If undefined, the PVC would use the default StorageClass on the host cluster.
+
+We prefer to keep the default StorageClass, `harvester`, of harvester-CSI-driver. That would benefit the general cases because users do not need to change anything if they do not want extra StorageClass. If users need a new StorageClass for different purposes, they can create a new one.
+
+### StorageClass manifest Sample
+
+``` yaml
+  allowVolumeExpansion: true
+  apiVersion: storage.k8s.io/v1
+  kind: StorageClass
+  metadata:
+    name: <StorageClass Name>
+  provisioner: driver.harvesterhci.io
+  reclaimPolicy: Delete
+  volumeBindingMode: Immediate
+  parameters:
+    hostStorageClass: <StorageClass Name on the Host Harvester Cluster>
+```
+
+### Notible Variable
+
+- **hostStorageClass** - This is associated with the StorageClass Name on the Host Harvester Cluster.
+
+### Flow Chart
+
+```mermaid
+flowchart TB
+    Start[Create Volume] -->|Some operations| GetVolParameters(Get Volume Parameters)
+  GetVolParameters --> GetHostStorageClassParam{Check `hostStorageClass` on Volume Parameters }
+  GetHostStorageClassParam -->|exist| D[Using `hostStorageClass` of volume parameters] --> CreateVolume
+  GetHostStorageClassParam -->|not exist| E[Using the `hostStorageClass` from configuration] --> CreateVolume
+  CreateVolume[Create volume and patch StorageClass with `hostStorageClass`]
+```
+
+### Test plan
+
+#### Verify the Default StorageClass behavior
+
+  1. Create the volume `vol001` with the default StorageClass on the Guest Cluster
+  2. Check the `vol001` is created with the Default StorageClass on the Host Harvester Cluster.
+
+#### Verify the user-defined StorageClass behavior
+
+1. Create a new StorageClass `test` on the Host Harvester Cluster.
+2. Create a new StorageClass using the following manifest on the Guest Cluster.
+
+    ``` yaml
+    allowVolumeExpansion: true
+    apiVersion: storage.k8s.io/v1
+    kind: StorageClass
+    metadata:
+      name: harvester-SC-test
+    provisioner: driver.harvesterhci.io
+    reclaimPolicy: Delete
+    volumeBindingMode: Immediate
+    parameters:
+      hostStorageClass: test
+    ```
+
+3. Create the volume `vol002` with the above StorageClass on the Guest Cluster.
+4. Check the `vol002` is created with the `test` StorageClass on the Host Harvester Cluster.
+
+### Upgrade strategy
+
+No
+
+## Note [optional]
+
+### Rancher UI enhancement
+- The user needs to create a StorageClass from a manifest instead of the UI because the user can not select the provisioner `driver.harvesterhci.io` when creating a StorageClass through the UI.


enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,98 @@
...
+    allowVolumeExpansion: true
+    apiVersion: storage.k8s.io/v1
+    kind: StorageClass
+    metadata:
+      name: harvester-SC-test
+    provisioner: driver.harvesterhci.io
+    reclaimPolicy: Delete
+    volumeBindingMode: Immediate
+    parameters:
+      host-storage-class: test
```
bk201: I'd suggest using `hostStorageClass` to follow the convention.

Vicente-Cheng: That sounds makes sense. Updated, thanks!

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,96 @@
...
+3. Create the volume `vol002` with the above StorageClass on the Guest Cluster.
+4. Check the `vol002` is created with the `test` StorageClass on the host Harvester Cluster.
+
+### Upgrade strategy
+
+No
+
+## Note [optional]
+
+- The user needs to create StorageClass from manifest instead UI because the user could not select the provisioner `driver.harvesterhci.io` when creating StorageClass on UI.
```
guangbochen: This could be addressed via a UI enhancement.

guangbochen (approved):
Just a small nit and the rest is LGTM, thanks.

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,96 @@
...
+### Goals
+
+The user using rancher integration with Harvester Cluster (Called Guest Cluster) could create the volume with the different StorageClass associated with StorageClass on the Host Harvester Cluster.
+
+## Proposal
+
+We will implement the mechanism to get the StorageClass parameter from the Guest Cluster and try to create it on the Host Harvester Cluster.
+
+### User Stories
+
```
w13915984028, line 22: suggestion about user stories:

those 2 are more like motivations/goals, but not user stories

the stories are roughly aligned with test plans below, or precisely, test plans are derived from user stories

we may list a few stories like:

1. guest cluster uses default storageclass when nothing is changed

2. guest cluster gets its related storageclass manually, if UI has menu to query storageclass, add related content 

3. guest cluster may specify a customized storageclass name when cluster creating, as per detailed steps ...
  after cluster is created, any change to the storageclass has no effect (please confirm and list)
  if the storageclass name is wrong (not existing, or not in the namespace), is it checked in the validator or where?
  
 4. when a storageclass is used by a guest, is it allowed to delete this storageclass ? clarify it
 
 thanks.

Vicente-Cheng: Hi @w13915984028, making user stories aligned with test plans make sense to me. I will update this again.
Thanks for the suggestions!

enhancements/20221228-various-storage-class-support.md
Vicente-Cheng: Already updated the HEP and GH issue on rancher here (https://github.com/rancher/rancher/issues/40067)

enhancements/20221228-various-storage-class-support.md
Vicente-Cheng: Hi @w13915984028, I already updated the HEP again. Please help to review.

> 1. guest cluster uses default storageclass when nothing is changed

Add some information for it.

> 2. guest cluster gets its related storageclass manually, if UI has menu to query storageclass, add related content

Now guest cluster could not get the related StorageClass name on the host cluster because of the permission.

> 3. guest cluster may specify a customized storageclass name when cluster creating, as per detailed steps ...
>    after cluster is created, any change to the storageclass has no effect (please confirm and list)
>    if the storageclass name is wrong (not existing, or not in the namespace), is it checked in the validator or where?

We may prefer to keep the original StorageClass as the default StorageClass. That would benefit the general cases.

For the StorageClass name-checking with the host cluster. We cannot handle it well because of permission.

> 4. when a storageclass is used by a guest, is it allowed to delete this storageclass ? clarify it

Delete the related StorageClass would not affect the created PV/PVC. That I thought should be updated on StorageClass-related HEP.

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,114 @@
+# Various StorageClass support of Harvester CSI Driver
+
+## Summary
+
+Harvester supports the user-defined StorageClass after v1.1.0. Because of that, Harvester CSI Driver should also let users use the customer StorageClass mentioned above.
```
bk201:
```suggestion
Harvester supports the user-defined StorageClass after v1.1.0. Because of that, Harvester CSI Driver should also let users use the custom StorageClass mentioned above.
```

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):
@btat It would be nice if you can proofread the HEP. Thank you!

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/3052
+
+## Motivation
+
+### Goals
+
+The user using rancher integration with Harvester Cluster (Called Guest Cluster) could create the volume with the different StorageClass associated with StorageClass on the Host Harvester Cluster.
```
btat:
```suggestion
The user using the Rancher integration with a Harvester Cluster (called Guest Cluster) can create the volume with the different StorageClass associated with StorageClass on the Host Harvester Cluster.
```

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Proposal
+
+We will implement the mechanism to get the StorageClass parameter from the Guest Cluster and try to create it on the Host Harvester Cluster.
+
+### User Stories
+
+#### General use cases
+
+Generally, users use the default StorageClass(`harvester`). The default StorageClass has three replicas means higher reliability. Users do not do any extra configuration. Create volume with the default StorageClass(`harvester`), then use.
```
btat:
```suggestion
Generally, users use the default StorageClass (`harvester`). The default StorageClass has three replicas, which means higher reliability. Users can create a volume with the default StorageClass (`harvester`) and use it without any additional configuration.
```

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### More specific goals
+
+Sometimes users will have different purposes for storage volume. Like `data archiving` or `Storage tiering`. For the above requirements, users can change the data locality to HDD. Users can complete it by creating a new StorageClass and configuring its data locality to HDD. Also, for the `performance` purpose. Users can reduce the number of replicas (remember, the default replica number is three). Reducing the replica number from 3 to 2 would increase the volume performance, but it lost some data reliability. So, users can create a new StorageClass and config its replica number to 2 and make the storage volume with this StorageClass to get better performance.
```
btat: - Is changing the data locality to HDD considered part a change to increase performance? If not, it might be more clear if "Also, for the `performance` purpose. Users..." is changed to "For performance purposes, users...".
- Reducing the replicas from 3 to 2 is called out, is 2 the minimum or is going from 3 to 1 a possibility?
- `but it lost some data reliability` -> `but results in some lost data reliability`

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### More specific goals
+
+Sometimes users will have different purposes for storage volume. Like `data archiving` or `Storage tiering`. For the above requirements, users can change the data locality to HDD. Users can complete it by creating a new StorageClass and configuring its data locality to HDD. Also, for the `performance` purpose. Users can reduce the number of replicas (remember, the default replica number is three). Reducing the replica number from 3 to 2 would increase the volume performance, but it lost some data reliability. So, users can create a new StorageClass and config its replica number to 2 and make the storage volume with this StorageClass to get better performance. 
+
+### User Experience In Detail
+
+- Harvester CSI Driver no longer hardcodes the `longhorn` StorageClass name when creating volume from the Guest Cluster. Furthermore, The default StorageClass is associated with the `Default` StorageClass on the host Harvester Cluster.
```
btat: Not specific to this line, but if `host Harvester Cluster` and `Host Harvester Cluster` are both referring to the same thing it'd be better to use the same form throughout the doc to avoid ambiguity that they're different things.

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,114 @@
...
+
+### API changes
+
+No
+
+## Design
+
+### Implementation Overview
+
+PVC StorageClass will reference the `hostStorageClass`, which is from the StorageClass parameter or controller config hostStorageClass. The prioritize is StorageClass parameter > controller config. If w/o any definition, PVC would use the default StorageClass on the host cluster.
```
btat:
```suggestion
PVC StorageClass will reference the `hostStorageClass`, which is from the StorageClass parameter or controller config hostStorageClass. The priority is StorageClass parameter > controller config. If undefined, the PVC would use the default StorageClass on the host cluster.
```

enhancements/20221228-various-storage-class-support.md
```diff
@@ -0,0 +1,114 @@
...
+4. Check the `vol002` is created with the `test` StorageClass on the host Harvester Cluster.
+
+### Upgrade strategy
+
+No
+
+## Note [optional]
+
+### Rancher UI enhancement
+- The user needs to create StorageClass from manifest instead UI because the user could not select the provisioner `driver.harvesterhci.io` when creating StorageClass on UI.
```
btat:
```suggestion
- The user needs to create a StorageClass from a manifest instead of the UI because the user can not select the provisioner `driver.harvesterhci.io` when creating a StorageClass through the UI.
```

btat (commented):
Added a few minor suggestions.

enhancements/20221228-various-storage-class-support.md
Vicente-Cheng: > * Is changing the data locality to HDD considered part a change to increase performance? If not, it might be more clear if "Also, for the `performance` purpose. Users..." is changed to "For performance purposes, users...".

No, they are different purposes. I will change to "For performance purposes...". Thanks!

> * Reducing the replicas from 3 to 2 is called out, is 2 the minimum or is going from 3 to 1 a possibility?

I thought that reducing the replica to 1 was too aggressive. So I do not want to write it down. If we only focus on performance, 1 replica would get the better performance with write IO. WDUT?

> * `but it lost some data reliability` -> `but results in some lost data reliability`

OK!

enhancements/20221228-various-storage-class-support.md
Vicente-Cheng: OK, I will check again on this term! Thanks for the reminder!

enhancements/20221228-various-storage-class-support.md
btat: > I thought that reducing the replica to 1 was too aggressive. So I do not want to write it down. If we only focus on performance, 1 replica would get the better performance with write IO. WDUT?

Not explicitly mentioning 3 -> 1 sounds good to me. 

Not absolutely necessary, but maybe just a slight adjustment to indicate 3 -> 2 is an example and not the only choice.

E.g. `Reducing the replica number increases the volume's performance, but results in some lost data reliability. For example, users can create a new StorageClass and configure its replica number from 3 to 2 and make the storage volume with this StorageClass to get better performance.`.

btat (approved):

enhancements/20221228-various-storage-class-support.md
Vicente-Cheng: Done!  Thanks for the review helping.

Issue #3340, opened 2023-01-05
[BUG] An empty log-level value gets error


FrankYang0529:
**Describe the bug**
Update log-level value and then update it back to empty. We will get error message in status.

```
status:
  conditions:
  - lastUpdateTime: "2023-01-05T06:13:19Z"
    message: 'not a valid logrus Level: ""'
    status: "False"
    type: configured
```

**To Reproduce**
1. Update `log-level` setting to `debug`
2. Update `log-level` setting to empty.
3. We get error in status.

**Expected behavior**
The `log-level` should use default value.

**Environment**
 - Harvester ISO version: v1.1.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

TachunLin:
Verified **fixed** on `master-aa7fcafc-head` (02/07) . Close this issue.

### Result
After we set the log-level value to empty (By deleting the value in settings/log-level yaml) 
There is no error message display and can set correctly log-level to `Info` in UI

```
 apiVersion: harvesterhci.io/v1beta1                                                                                                                        │
│ default: info                                                                                                                                              │
│ kind: Setting                                                                                                                                              │
│ metadata:                                                                                                                                                  │
│   annotations:                                                                                                                                             │
│     harvesterhci.io/hash: d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f                                                                         │
│   creationTimestamp: "2023-02-08T02:46:49Z"                                                                                                                │
│   generation: 5                                                                                                                                            │
│   name: log-level                                                                                                                                          │
│   resourceVersion: "370565"                                                                                                                                │
│   uid: a67f914e-c9ad-455e-8f78-81f5453a3684                                                                                                                │
│ status:                                                                                                                                                    │
│   conditions:                                                                                                                                              │
│   - lastUpdateTime: "2023-02-08T09:15:27Z"                                                                                                                 │
│     status: "False"                                                                                                                                        │
│     type: configured
```

### Test Information
* Test Environment: 4 nodes Harvester bare machines
* Harvester version: `master-aa7fcafc-head` (02/07) 

### Verify Steps
1. Access Harvester dashboard and open settings
2. Change log-level to `Debug`
3. Login Harvester management node and change to root
4. Run k9s
5. search `settings`
6. Click `e` to on log-level
7. Delete the line of `value: debug`
  ```
  # Please edit the object below. Lines beginning with a '#' will be ignored,
[...]

PR #3341, opened 2023-01-05
fix(log-level): use default value if value is empty

FrankYang0529:
**Problem:**
Update log-level value and then update it back to empty. We will get error message in status.

**Solution:**
Use default log-level value if the value is empty.

**Related Issue:**
https://github.com/harvester/harvester/issues/3340

**Test plan:**
1. Update `log-level` setting to debug
2. Update `log-level` setting to empty.
3. There should not have error in status.


bk201 (approved):
LGTM, nice spot.

guangbochen (approved):
LGTM, thanks.

bk201:
@mergifyio backport v1.1

PR #3274, opened 2022-12-12
HEP: Upgrade Observability [CI SKIP]

starbops:
Related issue: #1926

https://github.com/starbops/harvester/blob/log-enhance-hep/enhancements/20221207-upgrade-observability.md

PR #3379, opened 2023-01-17
Enable logging with Harvester upgrades

starbops:

**Problem:**

Upgrade-related logs are hard to keep:

- The logs of the upgrade Jobs are gone almost immediately after the Jobs completed, e.g., pre-drain Jobs
- Existing services like harvester, rancher, and longhorn-manager could have crucial information which might be lost during the upgrade because Pods/containers restart

And support bundles have limitations:

- It can only snapshot a point in time of the cluster; what was lost was lost.

**Solution:**

Introducing UpgradeLog CRD and its accompanying controller collaborate with the existing upgrade procedure to act like a recording assistant.

For detailed information, please refer to the HEP #3274 

**Related Issue:**

#1926

**Test plan:**

1. Prepare a Harvester cluster in master-head version (including this PR)
2. Upgrade with the same ISO
3. Check if the corresponding UpgradeLog created
4. After the DownloadReady condition of the UpgradeLog becomes True, try to generate an archive by executing `curl -k -X POST -u "${CATTLE_ACCESS_KEY}:${CATTLE_SECRET_KEY}" https://<harvester-vip>/v1/harvester/harvesterhci.io.upgradelogs/harvester-system/<upgradelog-name>?action=generate`. Copy the response in stdout (it's the name of the to-be-generated archive)
5. Wait until the archive becomes ready (check this via `upgradelog.status.archives.<archiveName>.ready`), and download it by executing `curl -kfL -u "${CATTLE_ACCESS_KEY}:${CATTLE_SECRET_KEY}" https://<harvester-vip>/v1/harvester/harvesterhci.io.upgradelogs/harvester-system/<upgradelog-name>/download?archiveName=<archiveName> -o <archiveName>.tar.gz`. Check the content of the archive.
6. Users can generate several archives per UpgradeLog and download each of them by doing step 4 and 5 repeatedly during an upgrade
7. After the upgrade is ended (either in success or failure), users can still generate archives and download them. But the content of the archives will be the same because the collection has stopped.
9. Remove the Upgrade resource should purge all the related resources, i.e., the UpgradeLog itself, the ManagedChart (if installed), the Logging resource, the ClusterFlow/Output resources, the log downloader Deployment, the log packager Jobs, and the PVC

Example of commands used to test this PR:

```
# generate an archive
$ curl -k -X POST -u token-jdpkd:k56qxnt28h4nmjp7cpqhzzgkflpkqhbhpppzmkk5bdfmr4p5j2gsn8 \
    https://192.168.48.130/v1/harvester/harvesterhci.io.upgradelogs/harvester-system/hvst-upgrade-lz2q7-upgradelog?action=generate
"hvst-upgrade-lz2q7-upgradelog-archive-2023-03-01T07-26-15Z" # this is the output

# download the archive
$ curl -fsSL -k -u token-jdpkd:k56qxnt28h4nmjp7cpqhzzgkflpkqhbhpppzmkk5bdfmr4p5j2gsn8 \
    https://192.168.48.130/v1/harvester/harvesterhci.io.upgradelogs/harvester-system/hvst-upgrade-lz2q7-upgradelog/download?archiveName=hvst-upgrade-lz2q7-upgradelog-archive-2023-03-01T07-26-15Z \
    -o hvst-upgrade-lz2q7-upgradelog-archive-2023-03-01T07-26-15Z.tar.gz

# check tarball contents
$ tar -tf hvst-upgrade-lz2q7-upgradelog-archive-2023-03-01T07-26-15Z.tar.gz
./
./logs/
./logs/kube-system.harvester-whereabouts-h6wm4.whereabouts.20230228.log
./logs/kube-system.snapshot-controller-6cc58d858f-9zrvl.snapshot-controller.20230228.log
./logs/kube-system.snapshot-controller-6cc58d858f-8mt8m.snapshot-controller.20230228.log
./logs/kube-system.kube-controller-manager-node-0.kube-controller-manager.20230228.log
./logs/longhorn-system.instance-manager-r-299489e1.replica-manager.20230228.log
./logs/kube-system.rke2-ingress-nginx-controller-5cnzf.rke2-ingress-nginx-controller.20230228.log
./logs/kube-system.kube-apiserver-node-0.kube-apiserver.20230228.log
./logs/cattle-system.rancher-5b8d85d4bc-ksfdg.rancher.20230228.log
./logs/longhorn-system.longhorn-manager-kvnlw.longhorn-manager.20230228.log
./logs/kube-system.rke2-canal-d6lqd.calico-node.20230228.log
./logs/kube-system.etcd-node-0.etcd.20230228.log
./logs/harvester-system.harvester-bcd5f8566-99qr9.apiserver.20230228.log
./logs/cattle-system.apply-hvst-upgrade-lz2q7-prepare-on-node-0-with-4a96a36ff-nfxsx.upgrade.20230228.log
./logs/longhorn-system.instance-manager-e-2db66e15.engine-manager.20230228.log
./logs/harvester-system.hvst-upgrade-lz2q7-apply-manifests-gjbgf.apply.20230228.log
```


pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,478 @@
...
+	if harvesterv1.UpgradeLogReady.GetStatus(upgradeLog) == "" {
+		logrus.Infof("[%s] Initialize upgradeLog %s/%s", upgradeLogControllerName, upgradeLog.Namespace, upgradeLog.Name)
+
+		toUpdate := upgradeLog.DeepCopy()
+		harvesterv1.UpgradeLogReady.CreateUnknownIfNotExists(toUpdate)
+
+		logrus.Infof("[%s] Deploy logging-operator", upgradeLogControllerName)
+		harvesterv1.OperatorDeployed.CreateUnknownIfNotExists(toUpdate)
+
+		// NOTE: As of v1.1.1, the logging-operator is by default deployed (rancher-logging), so we set the condition to true directly.
```
w13915984028: In v1.2.0, the logging feature will be `addon`, it may not be enabled by default

w13915984028 (commented):
General question:

Ever consider to use `helm` to deploy the upgrade logging operator, flow and output, or use `managedchart` or `addon`, those framework will help to deploy the related parts ?  thanks.

pkg/controller/master/upgradelog/controller.go
starbops: So basically, along with the conversion of the logging feature, there are two scenarios that we need to concern of:

1. When users upgrade Harvester from v1.1.1 to v1.1.2

Users cannot benefit from the enhancement due to the nature of Harvester's upgrade design, i.e., the newly introduced controllers will be effective only after the apply-manifest phase of an upgrade. Also, the hooking point of UpgradeLog creation is at the beginning of an upgrade, and we will miss that timing. But we still want to prepare this enhancement for the next upgrade (to v1.2.0)

2. When users upgrade Harvester from v1.1.2 to v1.2.0

This is the time things become complicated. I've checked the related PRs(#3380 and harvester/harvester-installer#402), and it seems there will be a deletion of the `rancher-logging` ManagedChart during an upgrade. If we leverage the existing `rancher-logging/-crd` ManagedCharts, logs can only be collected until the deletion. The entire logging infrastructure will be removed due to the deletion of the logging-operator and its CRDs.

---

So our best shot might be "deploy another logging ManagedChart dedicated to the upgrade observability enhancement, remove it after the upgrade ended."

What do you think?

w13915984028: that is good idea.

we may utilize managedchart/helm or even addon to deploy, instead of coding the deployment in controller, then more feature changes may reside in the .yaml file

logging components use certain amount of resources, they also need to be added into preflight check when it is enabled

starbops: > we may utilize managedchart/helm or even addon to deploy, instead of coding the deployment in controller, then more feature changes may reside in the .yaml file

Agreed. But there might be CRDs or other clustered resource conflicts while installing the charts. I just experimented with this: installing logging-operator in a namespace different from `cattle-logging-system` with helm on a v1.1.1 Harvester node. It complained a PodSecurityPolicy already exists, thus aborting the installation.

```
$ helm upgrade --install logging-operator banzaicloud-stable/logging-operator --namespace=harvester-system
Release "logging-operator" does not exist. Installing it now.
W0118 16:17:03.085337    4036 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
Error: rendered manifests contain a resource that already exists. Unable to continue with install: PodSecurityPolicy "psp.logging-operator" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "logging-operator": current value is "rancher-logging"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "harvester-system": current value is "cattle-logging-system"
```

We might encounter similar issues in a v1.2.0 environment and upgrade to a future release, say v1.2.1:

- User initiates the upgrade with the logging addon enabled
- User initiates the upgrade with the logging addon disabled

So, in general, the UpgradeLog controller should be able to detect the existence of the logging-operator and its CRDs, either in the form of ManagedCharts or Addons, to decide whether to deploy the dedicated logging ManagedChart or not.

> logging components use certain amount of resources, they also need to be added into preflight check when it is enabled

Unlike the existing `rancher-logging`, we tailored the upgrade logging infrastructure with as few components as possible. It will not collect events, audit logs, and node-level information. With that said, a preflight check is still good to have.

w13915984028: @starbops  I tried and checked again, now my rough idea is:

(1) Prepare an upgrade-log dedicated chart,  it is cloned, tailered and modified from the `rancher-logging` chart, replace the namespace, cluster role and so on, to let upgrade-log to be deployed in harvester-system namespace;  but all CRDs are sharing from `rancher-logging-crd` managedchart, this crd managedchart is not converted into addon.

(2) In upgrade script/controller, deploy this chart, wait the rollout

(3) Create flow/output in parallel of chart, or even add them into the chart (values.yaml), use the framework to create reosurce, deploy

From the work flow below, those are done step by step, but flow/output object does not rely on logging-operator, the operator is a controller which watches those resources.
https://github.com/harvester/harvester/blob/81a55a8dfbcd316f4b4f8f17bc7af490135739c9/enhancements/20221207-upgrade-observability.md#implementation-overview

(4) Finally, delete the upgrade chart, which will delete all upgrade-log related resources.

the overall idea is reusing helm chart/managedchart, write chart instead of controller to do those deployments. please double check, thanks.

cc @bk201 @guangbochen

starbops:
Also add a temporary fix for #3444 by requeueing the upgrade resource if the HTTP request to the repo VM failed due to a connection refused.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -344,10 +419,11 @@ func (h *upgradeHandler) isSingleNodeCluster() (string, error) {
 
 func initStatus(upgrade *harvesterv1.Upgrade) {
 	harvesterv1.UpgradeCompleted.CreateUnknownIfNotExists(upgrade)
+	// harvesterv1.LogReady.CreateUnknownIfNotExists(upgrade)
```
bk201: Suggest removing if this is not used.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,518 @@
...
+
+	if harvesterv1.OperatorDeployed.IsTrue(upgradeLog) && harvesterv1.InfraScaffolded.GetStatus(upgradeLog) == "" {
+		logrus.Infof("[%s] Start to scaffold the logging infrastructure for upgrade procedure", upgradeLogControllerName)
+
+		toUpdate := upgradeLog.DeepCopy()
+
+		if _, err := h.pvcClient.Create(preparePvc(upgradeLog)); err != nil && !apierrors.IsAlreadyExists(err) {
+			return nil, err
+		}
+		if _, err := h.loggingClient.Create(prepareLogging(upgradeLog)); err != nil && !apierrors.IsAlreadyExists(err) {
```
bk201, line 149: Doesn't work in air-gapped env. Got `ImagePullBackOff` on these pods:
```
│ harvester-system           hvst-upgrade-hkvvl-upgradelog-infra-fluentbit-68mrp              ●  0/1          0 ErrImagePullΔ
│ harvester-system           hvst-upgrade-hkvvl-upgradelog-infra-fluentbit-dl6n6              ●  0/1          0 ImagePullBackOff
│ harvester-system           hvst-upgrade-hkvvl-upgradelog-infra-fluentbit-dnd9b              ●  0/1          0 ImagePullBackOff
│ harvester-system           hvst-upgrade-hkvvl-upgradelog-infra-fluentbit-jfpfm              ●  0/1          0 ImagePullBackOff
│ harvester-system           hvst-upgrade-hkvvl-upgradelog-infra-fluentd-0                    ●  0/2          0 ImagePullBackOff
```

The missing images are:
```
fluent/fluent-bit:1.9.3
jimmidyson/configmap-reload:v0.4.0
ghcr.io/banzaicloud/fluentd:v1.14.6-alpine-5
```

Need to figure out a way to determine what's available on the system.

pkg/controller/master/upgradelog/controller.go
starbops: Now that in the operator deploying phase, the controller will detect whether the rancher-logging Addon/ManagedChart is installed or not. If both of them are not enabled/installed, the controller will try to deploy the upgradelog-version ManagedChart. After the upgrade is done, the installed ManagedChart will be removed.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Updated, thanks!

pkg/controller/master/upgradelog/controller.go
starbops: Change to rancher mirrored ones, which are already bundled in the ISO image. Please check it again, thanks!

pkg/controller/master/upgradelog/common.go
```diff
@@ -0,0 +1,959 @@
...
+					Enabled: true,
+					Path:    "/fluentd/log/out",
+					Size:    "10485760",
+				},
+			},
+		},
+	}
+}
+
+func prepareClusterFlow(upgradeLog *harvesterv1.UpgradeLog) *loggingv1.ClusterFlow {
```
w13915984028, line 169: could we put all of those definition into one yaml file, and use `kubectl apply -f ` to create

the indepedent yaml file will make it easier to maintain and extend

e.g.

https://github.com/harvester/harvester-installer/blob/master/pkg/config/templates/rancherd-10-harvester.yaml#L69

pkg/controller/master/upgradelog/controller.go
w13915984028: Yes, rancher mirroed ones are packed in ISO

https://github.com/rancher/charts/blob/dev-v2.7/charts/rancher-logging/100.1.1%2Bup3.17.3/values.yaml#L158

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -166,10 +194,34 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
...
 		if err != nil {
-			setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, err.Error(), "")
-			return h.upgradeClient.Update(toUpdate)
+			retryCount := 0
+			if toUpdate.Annotations == nil {
+				toUpdate.Annotations = make(map[string]string)
+				toUpdate.Annotations[getRepoRetryCountAnnotation] = "0"
+			}
+
+			retryCount, err = strconv.Atoi(toUpdate.Annotations[getRepoRetryCountAnnotation])
```
FrankYang0529: If any controller set annotations, but there is no `getRepoRetryCountAnnotation`. We may have an error here. We can update like:

```suggestion
        retryCountStr, ok := toUpdate.Annotations[getRepoRetryCountAnnotation]
        if !ok {
            retryCountStr = "0"
        }
        retryCount, err = strconv.Atoi(retryCountStr)
```

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -327,6 +379,29 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
 			return err
 		}
 	}
+
+	// tear down logging infra if any
+	if upgrade.Spec.LogEnabled {
+		upgradeLogName := fmt.Sprintf("%s-upgradelog", upgrade.Name)
```
FrankYang0529: It's better to use a util function to construct `upgradeLogName`.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: This one is obsolete. The relevant commit is cherry-picked and merged with a different solution in #3486. I will rebase and push again. Thanks!

pkg/controller/master/upgradelog/common.go
bk201: @w13915984028 I'd suggest we keep the way to create these objects programmatically. These resources are rarely changed and we can save the hassle to embed or manipulate YAML files. Let's put our effort into how to work well with the logging add-on transition first.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Updated, thanks!

pkg/controller/master/upgradelog/controller.go
starbops: @w13915984028 Can we disable the default Logging resource installed by rancher-logging ManagedChart? The UpgradeLog controller will try to install its own version of Logging resource. Right now, they will both exist; therefore, there are two sets of fluentd and fluentbit installations, but only one of them is used by UpgradeLog.

starbops: After discussing with @w13915984028, we have to patch the rancher-logging chart to disable the default Logging resource. The default Logging resource will bring up a fluentd StatefulSet and a fluentbit DaemonSet that the UpgradeLog will never use. But without configuring any Flows and Outputs, the system resources they consume are neglectable. So we will leave the current implementation as is due to cost efficiency and the release time schedule.

FrankYang0529:
Most of test cases work fine. Only in case 8, if I deleted `Upgrade`, `UpgradeLog` was deleted, too.

starbops:
> Most of test cases work fine. Only in case 8, if I deleted `Upgrade`, `UpgradeLog` was deleted, too.

Hi @FrankYang0529, I have updated the test plan. The original behavior was not to delete UpgradeLog resource while the corresponding Upgrade resource was removed. The new behavior is the UpgradeLog resource will be removed, too.

pkg/server/router.go
```diff
@@ -57,6 +58,11 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
 
 	btHealthyHandler := backuptarget.NewHealthyHandler(r.scaled)
 	m.Path("/v1/harvester/backuptarget/healthz").Methods("GET").Handler(btHealthyHandler)
+
+	ulGenerateHandler := upgradelog.NewGenerateHandler(r.scaled)
+	m.Path("/v1/harvester/upgradelog/{upgradeLogName}").Methods("POST").Handler(ulGenerateHandler)
```
FrankYang0529: Since `UpgradeLog` is a harvester resource, do we want to use `RegisterSchema` for it like other resources? So we can have API requests with `action` in parameters.

```
curl -u "${CATTLE_ACCESS_KEY}:${CATTLE_SECRET_KEY}" --insecure -X POST https://<harvester-vip>/v1/harvester/upgradelog/<upgradelog-name>?action=generateArchive'
```

https://github.com/harvester/harvester/blob/d8b5b546c5db7079e650f034649cbabe1b53fd95/pkg/api/register.go#L31-L39

FrankYang0529 (approved):
Test and LGTM. Thanks.

go.mod
```diff
@@ -58,6 +58,8 @@ replace (
 
 require (
 	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/banzaicloud/logging-operator/pkg/sdk v0.8.14
+	github.com/banzaicloud/operator-tools v0.28.2
```
guangbochen: both versions are old, can we bump to newer ones?

pkg/api/upgradelog/download_handler.go
```diff
@@ -0,0 +1,118 @@
...
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/config"
+	ctlupgradelog "github.com/harvester/harvester/pkg/controller/master/upgradelog"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	upgradeLogNamespace = "harvester-system"
```
guangbochen: This is defined in multiple places, let's move all to the `util/constants.go`

pkg/api/upgradelog/download_handler.go
```diff
@@ -0,0 +1,118 @@
...
+		upgradeLogCache:  scaled.HarvesterFactory.Harvesterhci().V1beta1().UpgradeLog().Cache(),
+		httpClient: &http.Client{
+			Timeout: 30 * time.Second,
+		},
+	}
+}
+
+func (h *DownloadHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	upgradeLogName := mux.Vars(r)["upgradeLogName"]
+	archiveName := mux.Vars(r)["archiveName"]
```
guangbochen: Let's leverage the encode method once the PR is merged https://github.com/harvester/harvester/pull/3549/files#diff-e290efac4355b20d1b6858649bd29946bab12964336247c96ad1370e51e4503bR44

pkg/api/upgradelog/download_handler.go
```diff
@@ -0,0 +1,118 @@
...
+		rw.Header().Set("Content-Disposition", "attachment; filename="+fmt.Sprintf("%s%s", archiveName, archiveSuffix))
+		contentType := resp.Header.Get("Content-Type")
+		if contentType != "" {
+			rw.Header().Set("Content-Type", contentType)
+		}
+
+		if _, err := io.Copy(rw, resp.Body); err != nil {
+			util.ResponseError(rw, http.StatusInternalServerError, err)
+			return
+		}
```
guangbochen: I think we need a proper retry of this, e.g., with either a timeout limit or retry maximum

pkg/api/upgradelog/generate_handler.go
```diff
@@ -0,0 +1,99 @@
...
+		context:          scaled.Ctx,
+		jobClient:        scaled.BatchFactory.Batch().V1().Job(),
+		upgradeCache:     scaled.HarvesterFactory.Harvesterhci().V1beta1().Upgrade().Cache(),
+		upgradeLogClient: scaled.HarvesterFactory.Harvesterhci().V1beta1().UpgradeLog(),
+		upgradeLogCache:  scaled.HarvesterFactory.Harvesterhci().V1beta1().UpgradeLog().Cache(),
+	}
+}
+
+func (h *GenerateHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	upgradeLogName := mux.Vars(r)["upgradeLogName"]
```
guangbochen: ditto

pkg/api/upgradelog/generate_handler.go
```diff
@@ -0,0 +1,99 @@
...
+	return &GenerateHandler{
+		context:          scaled.Ctx,
+		jobClient:        scaled.BatchFactory.Batch().V1().Job(),
+		upgradeCache:     scaled.HarvesterFactory.Harvesterhci().V1beta1().Upgrade().Cache(),
+		upgradeLogClient: scaled.HarvesterFactory.Harvesterhci().V1beta1().UpgradeLog(),
+		upgradeLogCache:  scaled.HarvesterFactory.Harvesterhci().V1beta1().UpgradeLog().Cache(),
+	}
+}
+
+func (h *GenerateHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
```
guangbochen: ditto

pkg/api/upgradelog/download_handler.go
```diff
@@ -0,0 +1,118 @@
...
+		podCache:         scaled.CoreFactory.Core().V1().Pod().Cache(),
+		upgradeLogClient: scaled.HarvesterFactory.Harvesterhci().V1beta1().UpgradeLog(),
+		upgradeLogCache:  scaled.HarvesterFactory.Harvesterhci().V1beta1().UpgradeLog().Cache(),
+		httpClient: &http.Client{
+			Timeout: 30 * time.Second,
+		},
+	}
+}
+
+func (h *DownloadHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
```
guangbochen: can we add some comments to describe what the handler is doing, thanks.?

pkg/api/upgradelog/generate_handler.go
```diff
@@ -0,0 +1,99 @@
...
+		if _, err := h.jobClient.Create(ctlupgradelog.PrepareLogPackager(upgradeLog, imageVersion, archiveName, component)); err != nil {
+			util.ResponseError(rw, http.StatusInternalServerError, errors.Wrap(err, "fail to create log packager job"))
+			return
+		}
+		toUpdate := upgradeLog.DeepCopy()
+		ctlupgradelog.SetUpgradeLogArchive(toUpdate, archiveName, archiveSize, generatedTime, false)
+		if _, err := h.upgradeLogClient.Update(toUpdate); err != nil {
+			util.ResponseError(rw, http.StatusInternalServerError, errors.Wrap(err, "fail to update upgradelog resource"))
+			return
+		}
```
guangbochen: ditto, prefer to add retry mechanism.

pkg/codegen/main.go
```diff
@@ -159,3 +179,19 @@ func capiWorkaround() {
 		logrus.Fatalf("failed to update the clusters.cluster.x-k8s.io client file: %v", err)
 	}
 }
+
+// loggingWorkaround replaces the variable `SchemeGroupVersion` with `GroupVersion` in logging.banzaicloud.io client because
```
guangbochen, line 183: Let's track it and plan to fix it in future versions.

pkg/config/context.go
```diff
@@ -166,6 +169,13 @@ func SetupScaled(ctx context.Context, restConfig *rest.Config, opts *generic.Fac
 	scaled.CniFactory = cni
 	scaled.starters = append(scaled.starters, cni)
 
+	logging, err := loggingv1.NewFactoryFromConfigWithOptions(restConfig, opts)
+	if err != nil {
+		return nil, nil, err
+	}
+	scaled.LoggingFactory = logging
+	scaled.starters = append(scaled.starters, cni)
```
guangbochen, line 177: we can skip this if this is only needed by the master-controller.

pkg/controller/master/upgradelog/common.go
```diff
@@ -0,0 +1,1008 @@
...
+			},
+			Name:      name.SafeConcatName(upgradeLog.Name, LogArchiveComponent),
+			Namespace: upgradeLogNamespace,
+			OwnerReferences: []metav1.OwnerReference{
+				upgradeLogReference(upgradeLog),
+			},
+		},
+		Spec: corev1.PersistentVolumeClaimSpec{
+			AccessModes: []corev1.PersistentVolumeAccessMode{
+				corev1.ReadWriteOnce,
```
guangbochen, line 53: Does the PVC need to be migratable during the update?

pkg/controller/master/upgradelog/common.go
```diff
@@ -0,0 +1,1008 @@
...
+		},
+	}
+}
+
+func SetUpgradeLogArchive(upgradeLog *harvesterv1.UpgradeLog, archiveName string, archiveSize int64, generatedTime string, ready bool) {
+	if upgradeLog == nil {
+		return
+	}
+	if upgradeLog.Status.Archives == nil {
+		upgradeLog.Status.Archives = make(map[string]harvesterv1.Archive)
```
guangbochen: set with proper init size if it is known, e.g., 1

pkg/controller/master/upgradelog/common.go
```diff
@@ -0,0 +1,1008 @@
...
+		return
+	}
+	upgradeLog.Status.Archives[archiveName] = harvesterv1.Archive{
+		Size:          archiveSize,
+		GeneratedTime: generatedTime,
+		Ready:         ready,
+	}
+}
+
+func GetDownloaderPodIP(podCache ctlcorev1.PodCache, upgradeLog *harvesterv1.UpgradeLog) (string, error) {
```
guangbochen: Prefer to use the svc instead of podIP if applicable:
1. pod IP is ephemeral; it can be changed after get IP.
2. service name is always predictable and can be checked using its related endpoint.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+	packagerImageRepository             = "rancher/harvester-upgrade"
+	downloaderImageRepository           = "rancher/harvester-upgrade"
+	addonNamespace                      = "cattle-logging-system"
+	managedChartNamespace               = "fleet-local"
+	operatorNamespace                   = "cattle-logging-system"
+	rancherLoggingChart                 = "rancher-logging"
+	rancherLoggingAddonName             = "rancher-logging"
+	rancherLoggingManagedChartName      = "rancher-logging"
+
+	upgradeNamespace                 = "harvester-system"
```
guangbochen: can we aggregate some common values and use the ones from the `constant.go` if already exist?

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+	pvcClient           ctlcorev1.PersistentVolumeClaimClient
+	statefulSetClient   ctlappsv1.StatefulSetClient
+	statefulSetCache    ctlappsv1.StatefulSetCache
+	upgradeClient       ctlharvesterv1.UpgradeClient
+	upgradeCache        ctlharvesterv1.UpgradeCache
+	upgradeLogClient    ctlharvesterv1.UpgradeLogClient
+	upgradeLogCache     ctlharvesterv1.UpgradeLogCache
+}
+
+func (h *handler) OnUpgradeLogChange(_ string, upgradeLog *harvesterv1.UpgradeLog) (*harvesterv1.UpgradeLog, error) {
```
guangbochen, line 76: can we reduce the method complexity of this by dividing into different methods? thanks.

pkg/apis/harvesterhci.io/v1beta1/upgradelog.go
```diff
@@ -0,0 +1,48 @@
+package v1beta1
+
+import (
+	"github.com/rancher/wrangler/pkg/condition"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	OperatorDeployed condition.Cond = "OperatorReady"
```
guangbochen: \s\Operator\LoggingOperator

pkg/apis/harvesterhci.io/v1beta1/upgradelog.go
```diff
@@ -0,0 +1,48 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   UpgradeLogSpec   `json:"spec"`
+	Status UpgradeLogStatus `json:"status,omitempty"`
+}
+
+type UpgradeLogSpec struct {
+	// +kubebuilder:validation:Required
+	Upgrade string `json:"upgrade,omitempty"`
```
guangbochen: Is this referring to the Upgrade Name?  prefer to \s\Upgrade\UpgradeName

pkg/controller/master/upgrade/common.go
```diff
@@ -391,13 +416,14 @@ func getDefaultTolerations() []corev1.Toleration {
...
-	testUpgradeImage = "test-upgrade-image"
-	testPlanHash     = "test-hash"
+	testJobName        = "test-job"
+	testPlanName       = "test-plan"
+	testNodeName       = "test-node"
+	testUpgradeName    = "test-upgrade"
+	testUpgradeLogName = "test-upgrade-upgradelog"
+	testVersion        = "test-version"
+	testUpgradeImage   = "test-upgrade-image"
+	testPlanHash       = "test-hash"
```
guangbochen: let's move this and its methods to the test file if they are only used by the test.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -94,13 +97,35 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
...
 		initStatus(toUpdate)
 
+		if !upgrade.Spec.LogEnabled {
+			logrus.Info("Upgrade observability is administratively disabled")
+			setLogReadyCondition(toUpdate, corev1.ConditionFalse, "Disabled", "Upgrade observability is administratively disabled")
+			toUpdate.Labels[upgradeStateLabel] = StateLoggingInfraPrepared
+			return h.upgradeClient.Update(toUpdate)
+		}
+		logrus.Info("Enabling upgrade observability")
+		if upgradeLog, err := h.upgradeLogClient.Create(prepareUpgradeLog(upgrade)); err != nil && !apierrors.IsAlreadyExists(err) {
```
guangbochen: prefer to break up long-line expressions for better project consistency. e.g.,
```
upgradeLog, err := h.upgradeLogClient.Create(prepareUpgradeLog(upgrade)); 
if err != nil && !apierrors.IsAlreadyExists(err) {
  // handle the logic here
}
```

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -345,6 +370,28 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
 			return err
 		}
 	}
+
+	// tear down logging infra if any
+	if upgrade.Spec.LogEnabled && upgrade.Status.UpgradeLog != "" {
```
guangbochen: a failed upgrade will also tear down the upgradelogs, is this expected? and how do they collect the logs?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -345,6 +370,28 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
+		upgradeLogToUpdate := upgradeLog.DeepCopy()
+		harvesterv1.UpgradeEnded.SetStatus(upgradeLogToUpdate, string(corev1.ConditionTrue))
+		harvesterv1.UpgradeEnded.Reason(upgradeLogToUpdate, "")
+		harvesterv1.UpgradeEnded.Message(upgradeLogToUpdate, "")
```
guangbochen, line 387: is this a real teardown seems we are only updating the status.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+
+	if harvesterv1.OperatorDeployed.GetStatus(upgradeLog) == "" {
+		logrus.Info("Check if there are any existing logging-operator")
+
+		toUpdate := upgradeLog.DeepCopy()
+		harvesterv1.OperatorDeployed.CreateUnknownIfNotExists(toUpdate)
+
+		// Detect rancher-logging Addon
+		if addon, err := h.addonCache.Get(addonNamespace, rancherLoggingAddonName); err != nil {
```
guangbochen: ditto, let's break up the line

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+		} else {
+			if addon.Status.Status == harvesterv1.AddonEnabled {
+				setOperatorDeployedCondition(toUpdate, corev1.ConditionTrue, "Skipped", "rancher-logging Addon is enabled")
+				return h.upgradeLogClient.Update(toUpdate)
+			}
+			logrus.Info("rancher-logging Addon is not enabled")
+		}
+
+		// Detect the rancher-logging ManagedChart
+		if managedChart, err := h.managedChartCache.Get(managedChartNamespace, rancherLoggingManagedChartName); err != nil {
```
guangbochen: ditto, let's break up the line

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+			}
+			logrus.Warn("rancher-logging ManagedChart is not ready")
+			return nil, err
+		}
+
+		// If none of the above exists, install the customized rancher-logging ManagedChart
+		logrus.Info("Deploy logging-operator")
+		if _, err := h.managedChartClient.Create(prepareOperator(upgradeLog)); err != nil && !apierrors.IsAlreadyExists(err) {
+			return nil, err
+		}
```
guangbochen, line 133: we should not install the `rancher-logging` managedChart if the add-on is not enabled since from v1.2 logging should only be managed via add-ons, and for any versions before v1.2 the logging managedChart should always exist.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+
+		return h.upgradeLogClient.Update(toUpdate)
+	}
+
+	if harvesterv1.OperatorDeployed.IsTrue(upgradeLog) && harvesterv1.InfraScaffolded.GetStatus(upgradeLog) == "" {
```
guangbochen: can you please help to add comments on each condition checks, it will be easier to maintain and read, thanks.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+		fluentdAnnotation, ok := upgradeLog.Annotations[upgradeLogFluentdAnnotation]
+		if !ok {
+			return upgradeLog, nil
+		}
+
+		if isInfraReady := (fluentBitAnnotation == upgradeLogFluentBitReady) && (fluentdAnnotation == upgradeLogFluentdReady); isInfraReady {
+			logrus.Info("Logging infrastructure is ready")
+			setInfraScaffoldedCondition(toUpdate, corev1.ConditionTrue, "", "")
+			return h.upgradeLogClient.Update(toUpdate)
+		}
```
guangbochen: prefer to check with real-time data instead of using an annotation indicator because:
- its status might get changed after an annotation is set.
- it is less intuitive in design.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+		return upgradeLog, nil
+	}
+
+	if harvesterv1.InfraScaffolded.IsTrue(upgradeLog) && harvesterv1.UpgradeLogReady.IsUnknown(upgradeLog) {
+		logrus.Info("Check if the log-collecting rules are installed")
+
+		toUpdate := upgradeLog.DeepCopy()
+
+		clusterFlowAnnotation := upgradeLog.Annotations[upgradeLogClusterFlowAnnotation]
+		clusterOutputAnnotation := upgradeLog.Annotations[upgradeLogClusterOutputAnnotation]
```
guangbochen, line 189: ditto

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+
+		// handle upgradeLog resource
+		toUpdate := upgradeLog.DeepCopy()
+		if toUpdate.Annotations == nil {
+			toUpdate.Annotations = make(map[string]string)
+		}
+		toUpdate.Annotations[upgradeLogStateAnnotation] = upgradeLogStateCollecting
+		harvesterv1.UpgradeEnded.CreateUnknownIfNotExists(toUpdate)
+
+		return h.upgradeLogClient.Update(toUpdate)
```
guangbochen: it is better to update the upgrade CR on its own controller by listAndWatch the `UpgradeLog`:
- upgradeLog is owned by upgrade CR, and it is better not to manage its parent object.
- tidy up the upgradeLog controller and it should only focus on managing the logging stuff only.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,615 @@
...
+	}
+	if err := h.clusterOutputClient.Delete(upgradeLogNamespace, name.SafeConcatName(upgradeLog.Name, OutputComponent), &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if err := h.loggingClient.Delete(name.SafeConcatName(upgradeLog.Name, InfraComponent), &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if err := h.managedChartClient.Delete(managedChartNamespace, name.SafeConcatName(upgradeLog.Name, OperatorComponent), &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
```
guangbochen: should skip deleting the managedChart

pkg/server/router.go
starbops: @FrankYang0529 Thanks for the suggestion! I updated the APIs to align with other resources, please take a look.

pkg/config/context.go
starbops: IIRC, there will be errors accessing the generate/download APIs without adding this to the scaled context. Or is there a better way to address this? Thanks!

pkg/controller/master/upgradelog/common.go
starbops: Yes, it will follow the fluentd Pod. And other Pods like the log packager and downloader have PodAffinities specified to coexist with the fluentd Pod on the same node.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Only the log-collecting infra will be torn down. The log downloader will still be at your service.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: It is the upgrade controller that signals whether the upgradelog infra should be torn down or not. The actual tear down is in the upgradelog controller.

bk201 (approved):
LGTM!

starbops:
Hi @guangbochen, I updated the PR per the review comments
👍 means addressed
👀 means will enhance in future releases
Thanks for your advice!

pkg/api/upgradelog/handler.go
```diff
@@ -0,0 +1,342 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctlupgradelog "github.com/harvester/harvester/pkg/controller/master/upgradelog"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	archiveSuffix                = ".tar.gz"
```
w13915984028: shall we keep the same archive format with support-bundle, use ".zip", which is also easy to process in windows

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -94,13 +97,36 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
...
+
 		if err := h.resetLatestUpgradeLabel(upgrade.Name); err != nil {
 			return nil, err
 		}
-		logrus.Info("Creating upgrade repo image")
+
 		toUpdate := upgrade.DeepCopy()
 		initStatus(toUpdate)
 
+		if !upgrade.Spec.LogEnabled {
```
w13915984028, line 108: when the Spec.LogEnabled is false, why need to update condition here ?  seems not necessary

when other places are checking the condtion, they can check Spec.LogEnabled first

w13915984028 (commented):
A few questions:

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,625 @@
...
+	upgradeLogClusterOutputReady      = "ClusterOutputReady"
+
+	// Annotation indicating the state of UpgradeLog
+	upgradeLogStateAnnotation = "harvesterhci.io/upgradeLogState"
+	upgradeLogStateCollecting = "Collecting"
+	upgradeLogStateStopped    = "Stopped"
+
+	// Logging infra images
+	fluentBitImageRepo      = "rancher/mirrored-fluent-fluent-bit"
+	fluentBitImageTag       = "1.9.5"
```
w13915984028, line 45: not a good practice to hard code the image tag into source code

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,625 @@
...
+		// Detect rancher-logging Addon
+
+		addon, err := h.addonCache.Get(util.CattleLoggingSystemNamespaceName, util.RancherLoggingName)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				return nil, err
+			}
+			logrus.Info("rancher-logging Addon is not installed")
+		} else {
+			if addon.Status.Status == harvesterv1.AddonEnabled {
```
w13915984028: use `addon.Spec.Enabled` instead of `addon.Status.Status` ?

pkg/controller/master/upgradelog/controller.go
w13915984028: to reduce the complexity of potential battling between `managedchart`, `addon` and here, we may consider: 

only when `managedchart` or `addon` (and is enabled) is there, then enable upgradeLog, namely, the `logging-operator` should have already been there,  otherwise, do not deploy

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,625 @@
...
+		logrus.Info("Deploy logging-operator")
+		if _, err := h.managedChartClient.Create(prepareOperator(upgradeLog)); err != nil && !apierrors.IsAlreadyExists(err) {
+			return nil, err
+		}
+
+		return h.upgradeLogClient.Update(toUpdate)
+	}
+
+	// Try to scaffold the logging infrastructure by creating a customized Logging resource
+	if harvesterv1.LoggingOperatorDeployed.IsTrue(upgradeLog) && harvesterv1.InfraScaffolded.GetStatus(upgradeLog) == "" {
```
w13915984028: `scaffold` seems to only be noun, the `InfraScaffolded` looks a bit ...

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,625 @@
...
+			return nil, err
+		}
+		// The creation of the Logging resource will indirectly bring up fluent-bit DaemonSet and fluentd StatefulSet
+		if _, err := h.loggingClient.Create(prepareLogging(upgradeLog)); err != nil && !apierrors.IsAlreadyExists(err) {
+			return nil, err
+		}
+
+		harvesterv1.InfraScaffolded.CreateUnknownIfNotExists(toUpdate)
+
+		return h.upgradeLogClient.Update(toUpdate)
```
w13915984028, line 155: should `deepequal` be used to check if it is really needed to Update ?, the APIServer operation Update is much heavier than local check of duplication/unchanged.

for all `Update` in this function, the logic is similar, as OnChange itself may run many times.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -0,0 +1,625 @@
...
+	}
+
+	// Signal to proceed the original upgrade flow
+	if harvesterv1.UpgradeLogReady.IsTrue(upgradeLog) && harvesterv1.UpgradeEnded.GetStatus(upgradeLog) == "" {
+		logrus.Info("Logging infrastructure is ready, proceed the upgrade procedure")
+
+		// handle corresponding upgrade resource
+		upgradeName := upgradeLog.Spec.UpgradeName
+		upgrade, err := h.upgradeCache.Get(util.HarvesterSystemNamespaceName, upgradeName)
+		if err != nil {
```
w13915984028, line 222: If in some cases, the `Upgrade` itself has been deleted, here the none existing should be checked to terminate upgradeLog

w13915984028 (commented):
At first, thanks @starbops has done a lot of work to implement this feature from beginning.

The `pkg/controller/master/upgradelog/controller.go` has an internal state machine to control the deployment of log related mechanism, and the overall feature is embedded into `upgrade` process.

I am afraid we can test all throughly in v1.1.2.

In case there are bugs happening inbetween the upgrade, do we have quick way to recover from the middle state of it?

pkg/api/upgradelog/handler.go
starbops: Thanks for the heads up. It's better to align with the support bundle.

pkg/controller/master/upgradelog/controller.go
starbops: Checking the current state instead of the desired state is preferred. I'm not sure if there is any chance the `addon.Spec.Enabled` is `true` but the `addon.Status.Status` is `false`. If that won't happen, we can go with `addon.Spec.Enabled` without problems.

Or do you suggest, in the first place, we should trust the `rancher-logging` addon will eventually roll out by its controller and will become active in `addon.Status.Status` in case we mistakenly deploy a duplicated logging-operator?

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Again, thanks for the heads up. There are inconsistencies in the UpgradeLog checking: some places use `Upgrade.Spec.LogEnabled`, and others use `Upgrade.Status.UpgradeLog` to check if the UpgradeLog is running. This makes the existence of the `Upgrade.Status.Conditions[LogReady]` a bit useless. We should check the `LogReady` condition for judging the overall readiness of the UpgradeLog mechanism.

pkg/controller/master/upgradelog/controller.go
starbops: It seems the usage of `scaffold` as a verb is not so common. I can change it to something else like `created` or just `ready`.

pkg/controller/master/upgradelog/controller.go
starbops: Thanks for bringing up a different way of thinking. The causality of "enabling rancher-logging addon" and "enabling logging for upgrades" is a bit implicit since there's no strong binding between them. And we have a `LogEnabled` toggle in the upgrade CRD for users to choose from (it will be `true` by default). It's still better to have the deployment of the logging-operator handled by the upgradelog controller. More complied with the style of declaration.

cc @bk201 @guangbochen

pkg/controller/master/upgradelog/controller.go
starbops: Good spot, in that case, the upgradelog should be torn down immediately. Thanks!

guangbochen (approved):
LGTM, please help to squish the commits before merge, thanks.

pkg/controller/master/upgradelog/controller.go
starbops: I went through again all the `Update` calls in the `OnUpgradeLogChange` function. As you mentioned, this is designed as a state machine, and when it comes to each `Update` call, the `toUpdate` object is pretty sure variated from the one in the cache. Most of the `Update` will run only once. I added `DeepEqual` for those who are not the case. Like the one in the `OnClusterFlowChange` function. If there's still something missing, please let me know. Thanks.

pkg/controller/master/upgradelog/controller.go
w13915984028: Spec is from user input, it is the desired result, the controller is reconcilling to align with the spec, status reflect a snapshot in the timeline, it could be that when you check it, it is not ready, but soon it becomes ready

here we could blindly believe the addon controller will roll the addon out, otherwise, its addon's bug

starbops: Make sense to me. I updated it, thanks.

w13915984028 (approved):
LGTM, thanks for your great effort @starbops .

starbops:
A huge thanks to all the reviewers. There are two things that remain to be done in this PR, others could be enhanced in future releases.

- Extracting fluentd and fluent-bit's image repos/tags from the chart instead of hardcoded in the controller code
- Increasing the range of collected logs, which we learned from our customer support experiences

bk201:
Let's squash the commits before merging. Thanks.

starbops:
@bk201 Commits squashed, please check it again. Thanks!

bk201:
Got `ImagePullBackOff` on the image `rancher/mirrored-fluent-fluent-bit:1.9.3` in an air-gapped env. I'd suggest we use a hard-code tag in the code now, but switch to getting values from logging add-ons in v1.2.0. Please help file an issue to track this, thanks!

bk201 (approved):
Thanks for the great feature.

bk201:
@mergifyio backport v1.1

Issue #3032, opened 2022-10-25
[FEATURE] Filter list of migration to match VM scheduling rules


lanfon72:
### The problem
based on #2982, when VM created with node scheduling rules, The **Migrate** list should not show other nodes not in scheduling rules.

### The solution
Filter out those nodes not match rules

### The Alternatives
Mark those nodes with `(Not in scheduling rules)`

lanfon72:
Verified this feature has been implemented.

![image](https://github.com/harvester/harvester/assets/5169694/bf9abe3b-95b6-45b5-9fe1-2e9a4f8cadf3)


Test Information
----
* Environment: **qemu/KVM 3 nodes**
* Harvester Version: **master-d739c2a4-head**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with at least 2 nodes
2. Create an image for VM creation
3. Create VM `vm1` and **Node Scheduling** _on matching scheduling rules with *Priority:Required* to specific host (by using `Key=kubernetes.io/hostname, Operator=In, Value=<Host name>`
4. `vm1` should started successfully
5. Click migrate option on `vm1`, the dropdown list should be empty (with no matching options)

PR #3399, opened 2023-01-30
Migratable nodes option

chrisho:

**Problem:**
https://github.com/harvester/harvester/issues/3032#issue-1422120838

**Solution:**
Filter migratable nodes through by VM network to choose from.

**Related Issue:**
https://github.com/harvester/harvester/issues/3032

**Test plan:**
1. Create a ClusterNetwork.
2. Create a Network Config(at least two nodes) for ClusterNetwork.
3. Create a VM Network with the ClusterNetwork.
4. Create a VM with the VM Network.
5. Get and choose a migratable node from API: https://harvester/v1/harvester/kubevirt.io.virtualmachines/default/{your_vm}?action=getMigratableNodes.
```
{
    "nodes": [
        "node2"
    ]
}
```
6. (Optional) verify failure, choose a node without the ClusterNetwork and migrate, and check the response for error.
image
7. Migrate to a migratable node and wait for the VM migration to succeed.


pkg/api/vm/handler.go
```diff
@@ -41,6 +43,9 @@ const (
 	vmResource    = "virtualmachines"
 	vmiResource   = "virtualmachineinstances"
 	sshAnnotation = "harvesterhci.io/sshNames"
+
+	networkGroup      = "network.harvesterhci.io"
+	keyClusterNetwork = networkGroup + "/clusternetwork"
```
futuretea: can use https://github.com/harvester/network-controller-harvester/blob/7765e83555157622612e3a12f0df14204a3a97bf/pkg/utils/labels.go#L10 instead

pkg/api/vm/handler.go
```diff
@@ -369,6 +382,24 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
 	return err
 }
 
+func (h *vmActionHandler) isMigratableNode(targetNode string, vmi *kubevirtv1.VirtualMachineInstance) (bool, error) {
+	if targetNode == "" {
+		return true, nil
```
futuretea, line 386: if targetNode == “”， can we check if there has migratable nodes?

pkg/api/vm/handler.go
```diff
@@ -397,6 +428,85 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
...
+		Nodes []string `json:"nodes"`
+	}{
+		Nodes: nodes,
+	}
+
+	b, err := json.Marshal(resp)
+	if err != nil {
+		return err
+	}
+	_, err = rw.Write(b)
```
futuretea: should we return `Content-type` `application/json`?

pkg/api/vm/handler.go
```diff
@@ -397,6 +428,85 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
...
+		if requirement != nil {
+			nodeSelector = nodeSelector.Add(*requirement)
+		}
+	}
+	nodes, err := h.nodeCache.List(nodeSelector)
+	if err != nil {
+		return migratableNodes, err
+	}
+
+	// ignore the node where the VM is running
```
futuretea, line 462: var migratableNodes = make([]string, 0, len(nodes)-1)

pkg/api/vm/handler.go
```diff
@@ -369,6 +382,24 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
...
+	if targetNode == "" {
+		return true, nil
+	}
+
+	nodes, err := h.getMigratableNodesByVMI(vmi)
+	if err != nil {
+		return false, err
+	}
+
+	for _, node := range nodes {
```
futuretea: return slice.ContainsString(nodes, targetNode), nil

pkg/api/vm/handler.go
```diff
@@ -397,6 +428,85 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
 	return nil
 }
 
+func (h *vmActionHandler) getMigratableNodes(rw http.ResponseWriter, namespace, name string) error {
+	vmi, err := h.vmiCache.Get(namespace, name)
+	if err != nil {
+		return err
+	}
+	nodes, err := h.getMigratableNodesByVMI(vmi)
+	resp := struct {
```
futuretea: Prefer to add the Output struct to types.go

pkg/api/vm/schema.go
```diff
@@ -50,6 +50,7 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, options config
 	vmtv := scaled.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineTemplateVersion()
 	vmImages := scaled.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineImage()
 	storageClasses := scaled.StorageFactory.Storage().V1().StorageClass()
+	nadCache := scaled.CniFactory.K8s().V1().NetworkAttachmentDefinition().Cache()
```
futuretea: nads := scaled.CniFactory.K8s().V1().NetworkAttachmentDefinition()

pkg/api/vm/schema.go
```diff
@@ -77,6 +78,7 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, options config
 		backupCache:               backups.Cache(),
 		restores:                  restores,
 		settingCache:              settings.Cache(),
+		nadCache:                  nadCache,
```
futuretea: nadCache: nads.Cache(),

pkg/api/vm/handler.go
```diff
@@ -331,6 +340,13 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
 			return fmt.Errorf("Volume %s is container disk, needs to be removed before migration", volume.Name)
 		}
 	}
+
+	if ok, err := h.isMigratableNode(nodeName, vmi); !ok {
+		return errors.New("The target node is non-migratable")
```
futuretea: missing the err content

pkg/api/vm/formatter.go
```diff
@@ -16,21 +16,22 @@ import (
...
+	startVM            = "start"
+	stopVM             = "stop"
+	restartVM          = "restart"
+	softReboot         = "softreboot"
+	pauseVM            = "pause"
+	unpauseVM          = "unpause"
+	ejectCdRom         = "ejectCdRom"
+	migrate            = "migrate"
+	abortMigration     = "abortMigration"
+	getMigratableNodes = "getMigratableNodes"
```
futuretea: the http method of actions is `POST`, but the name `getMigratableNodes` seems like a GET API.
maybe `findMigratableNodes` would be better？

pkg/api/vm/handler.go
```diff
@@ -397,6 +429,82 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
 	return nil
 }
 
+func (h *vmActionHandler) findMigratableNodes(rw http.ResponseWriter, namespace, name string) error {
+	vmi, err := h.vmiCache.Get(namespace, name)
+	if err != nil {
+		return err
+	}
+	nodes, err := h.findMigratableNodesByVMI(vmi)
```
futuretea, line 439: this err missing check

futuretea (approved):
LGTM, Thanks! test plan passed

pkg/api/vm/handler.go
```diff
@@ -331,6 +340,15 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
 			return fmt.Errorf("Volume %s is container disk, needs to be removed before migration", volume.Name)
 		}
 	}
+
+	if ok, err := h.isMigratableNode(nodeName, vmi); nodeName == "" && err != nil {
```
guangbochen: the nodeName and err check should be separated; it is possible that nodeName is an empty string and err is not nil, but in that case, the error message returned would not indicate the actual problem, let's move it to the top method instead. Additionally, simply the check conditions into:

An updated pseudo code:
```
if ok, err := h.isMigratableNode(nodeName, vmi); err != nil {
	return fmt.Errorf("can't migrate the VM to the node %s: %s", nodeName, err.Error())
} else if !ok {
	return errors.New("the target node is non-migratable")
}
```

pkg/api/vm/handler.go
```diff
@@ -369,6 +384,23 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
...
 }
 
+func (h *vmActionHandler) isMigratableNode(targetNode string, vmi *kubevirtv1.VirtualMachineInstance) (bool, error) {
+	nodes, err := h.findMigratableNodesByVMI(vmi)
+	if err != nil {
+		return false, err
+	}
+
+	if len(nodes) == 0 {
+		return false, errors.New("no migration nodes")
```
guangbochen:
```suggestion
		return false, errors.New("no matching migratable nodes found")
```

pkg/api/vm/handler.go
```diff
@@ -369,6 +384,23 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
...
+	}
+
+	if targetNode == "" {
+		return true, nil
+	}
```
guangbochen: ditto, remove to the top

pkg/api/vm/handler_test.go
```diff
@@ -161,14 +163,34 @@ func TestMigrateAction(t *testing.T) {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "fake-node1",
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "fake-node2",
+			},
+		},
+	}
```
guangbochen, line 171: can we add proper testing cases to validate the `findMigratableNodes` output?

pkg/api/vm/handler.go
```diff
@@ -397,6 +429,85 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
...
+			nodeSelector = nodeSelector.Add(*requirement)
+		}
+	}
+	nodes, err := h.nodeCache.List(nodeSelector)
+	if err != nil {
+		return nil, err
+	}
+	if len(nodes) == 0 {
+		return nil, nil
+	}
```
guangbochen: prefer to define the `migratableNodes` var on the top and return it here instead.

pkg/api/vm/handler.go
```diff
@@ -397,6 +429,85 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
...
+	for _, network := range vmNetworks {
+		requirement, err := h.getNodeSelectorRequirementFromNetwork(vmi.Namespace, network)
+		if err != nil {
+			return nil, err
+		}
+		if requirement != nil {
+			nodeSelector = nodeSelector.Add(*requirement)
+		}
+	}
+	nodes, err := h.nodeCache.List(nodeSelector)
```
guangbochen: will need to filter out nodes with not-ready, drained, and under maintenance conditions.

pkg/api/vm/handler.go
```diff
@@ -397,6 +429,85 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
...
+	for _, node := range nodes {
+		if vmi.Status.NodeName == node.Name {
+			continue
+		}
+		migratableNodes = append(migratableNodes, node.Name)
+	}
+	return migratableNodes, nil
+}
+
+func (h *vmActionHandler) getNodeSelectorRequirementFromNetwork(defaultNamespace string, net kubevirtv1.Network) (*labels.Requirement, error) {
```
guangbochen: You should be able to use the existing rules from the VM/VMI; you also need to add the user's custom selector rules into account.

chrisho:
The latest commit has significant changes and needs to be reviewed again, thanks.

pkg/api/vm/formatter.go
```diff
@@ -226,6 +228,13 @@ func canAbortMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
 	return false
 }
 
+func isSpecificNodeFromVMINodeSelector(vmi *kubevirtv1.VirtualMachineInstance) bool {
```
guangbochen: \s\isSpecificNodeFromVMINodeSelector\hasNodeSelector

pkg/api/vm/handler.go
```diff
@@ -397,6 +427,93 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
...
+
+		migratableNodes = append(migratableNodes, node.Name)
+	}
+	return migratableNodes, nil
+}
+
+func isDrained(node *corev1.Node) bool {
+	if node.Spec.Unschedulable {
+		return true
+	}
```
guangbochen, line 487: you may also check if the node already contains the [harvester-drained annotation](https://github.com/harvester/harvester/blob/master/pkg/util/drainhelper/helper.go#L31) and return it directly

pkg/api/vm/handler_test.go
```diff
@@ -622,3 +644,239 @@ func (c fakeVirtualMachineInstanceMigrationClient) Watch(namespace string, opts
...
+	return result, err
+}
+
+func (c fakeNodeCache) AddIndexer(indexName string, indexer ctlcorev1.NodeIndexer) {
+	panic("implement me")
+}
+
+func (c fakeNodeCache) GetByIndex(indexName, key string) ([]*corev1.Node, error) {
+	panic("implement me")
+}
```
guangbochen: resue the fakeclient on the https://github.com/harvester/harvester/tree/master/pkg/util/fakeclients

pkg/api/vm/handler.go
```diff
@@ -320,6 +325,10 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
 	if !canMigrate(vmi) {
 		return errors.New("The VM is already in migrating state")
 	}
+	if isSpecificNodeFromVMINodeSelector(vmi) {
```
futuretea: check node selector in canMigrate.

---
for the vmi has node selector case,
UI will call findMigratableNodes and report an error when users click the `Migrate` button.
image
if we check node selector in `canMigrate`, the `Migrate` button will be hidden

pkg/api/vm/handler.go
```diff
@@ -397,6 +427,93 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
...
 }
 
+func (h *vmActionHandler) findMigratableNodes(rw http.ResponseWriter, namespace, name string) error {
+	vmi, err := h.vmiCache.Get(namespace, name)
+	if err != nil {
+		return err
+	}
+
+	if !canMigrate(vmi) {
+		return errors.New("The VM is already in migrating state")
```
futuretea: The error message does not match the function name

pkg/api/vm/types.go
```diff
@@ -42,3 +42,7 @@ type RemoveVolumeInput struct {
 type CloneInput struct {
 	TargetVM string `json:"targetVm"`
 }
+
+type GetMigratableNodesOutput struct {
```
futuretea: FindMigratableNodesOutput

futuretea (approved):
LGTM, Thanks. We need to figure out why CI is failing

guangbochen (dismissed):
LGTM, thanks.

pkg/api/vm/handler.go
```diff
@@ -397,6 +426,97 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
...
+		}
+	}
+
+	return false
+}
+
+func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.VirtualMachineInstance) (labels.Selector, error) {
+	if vmi == nil {
+		return nil, nil
+	}
```
futuretea:
```suggestion
	if vmi == nil || vmi.Spec.Affinity == nil || vmi.Spec.Affinity.NodeAffinity == nil || vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
		return labels.Everything(), nil
	}
```

pkg/api/vm/formatter.go
```diff
@@ -206,7 +208,8 @@ func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
 		vmi.Annotations[util.AnnotationMigrationUID] == "" {
 		return true
 	}
-	return false
+
+	return vmi.Spec.NodeSelector == nil
```
futuretea:
```go
func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
	if vmi == nil {
		return false
	}

	if !vmi.IsRunning() {
		return false
	}

	// The VM is already in migrating state
	if vmi.Annotations[util.AnnotationMigrationUID] != "" {
		return false
	}

	// The VM is set to run on a specific node
	if vmi.Spec.NodeSelector != nil && vmi.Spec.NodeSelector[corev1.LabelHostname] != "" {
		return false
	}

	return true
}
```

futuretea: fix CI failed. due to `vmi.Spec.NodeSelector` panic when 
```
	if canMigrate(vmi) {
	}
```

pkg/api/vm/handler.go
futuretea: if the vm is using management Network, `vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms` will panic
and we also need return `labels.Everything()` is this case. if we return `nil`, `nodeCache.List(nodeSelector)` will panic too.

futuretea (dismissed):
drone CI failed

guangbochen (approved):

futuretea (approved):

Issue #2466, opened 2022-07-07
[ENHANCEMENT] Shutdown grace period of 30s not valid for VMs that need a long time to shutdown


Martin-Weiss:
**Describe the bug**
It seems that the current shutdown wait time (before hard power of) for a guest VM is 30s.
It is not a valid approach to power of any VM during shutdown if it is not powered of after 30s.
(even some RKE2 nodes need much more time)

Every "hard" power off is a risk for data in the running machine - so we should avoid any "automated" hard power off as long and as good as we can. The "hard power off" should be the administrators decision.

**To Reproduce**
Deply a VM with a process that takes >30s to shutdown.
Shut down the VM.
Realize it is powered of before all processes could stop gracefully.

**Expected behavior**
The default should be to wait forever before the machine is "hard killed".

Beside the "shutdown graceful stop" button in the UI we also need a "forcefull power off" button in case the machine hangs or does not react on the ACPI based shutdown command.

**Environment**
 - Harvester ISO version: 1.0.2

FrankYang0529:
Hi @Martin-Weiss, from [KubeVirt Virtual Machine Graceful Shutdown document](https://github.com/kubevirt/kubevirt/blob/main/docs/graceful-shutdown.md), it looks like we can't have a "shutdown graceful stop" to wait forever before the machine is "hard killed", but we can let users modify `terminationGracePeriodSeconds` to adjust a waiting period before "hard killed". We can have a global default setting and users can give a different value when they create a new VM.

For "forcefull power off" function, we can add an API to trigger stop and then force delete VM pod.

Martin-Weiss:
Sounds valid. In general data protection should have highest priority and hard killing a machine will always be a risk for data and cause loss of grey data. 
So setting the terminationGracePeriodSeconds to something like i.e. 7 days might be a compromise while we also might need an additional button in the UI to force the power off in case this timer is waiting..
And yes - we need to make this configurable via global settings and per VM / per Namespace settings I assume.

TachunLin:
Verified **fixed** on `master-b49e264e-head` (05/22). Close this issue.

### Result
#### [PASS] Test Case 1: 
If users don't set default-termination-grace-period-seconds setting, terminationGracePeriodSeconds in a VM is 30 seconds

* The `default-vm-termination-grace-period-seconds` settings is `120` seconds
  ![image](https://github.com/harvester/harvester/assets/29251855/afbc2771-47e8-4daf-aef6-03fe8d3b3d3a)

[...]

PR #3417, opened 2023-02-07
feat(vm): add force stop

FrankYang0529:
**Feature:**
If users set `terminationGracePeriodSeconds`, we need provide a way to force stop the VM.

**Related Issue:**
https://github.com/harvester/harvester/issues/2466

**Test plan:**

Case 1: If users don't set `default-termination-grace-period-seconds` setting, `terminationGracePeriodSeconds` in a VM is 120 seconds

1. Create a VM.
2. Check whether `spec.template.spec.terminationGracePeriodSeconds` in the VM is `120`.

Case 2: `default-termination-grace-period-seconds` setting can't be negative.

1. Set value in `default-termination-grace-period-seconds` setting as `"-1"`. The webhook should return an error.

Case 3: `default-termination-grace-period-seconds` setting should be able to be parsed as an integer.

1. Set value in `default-termination-grace-period-seconds` setting as `"not an integer"`. The webhook should return an error.

Case 4: Set `default-termination-grace-period-seconds` setting and `terminationGracePeriodSeconds`in a new VM should follow the setting.

1. Set value in `default-termination-grace-period-seconds` setting as `"60"`.
2. Create a VM.
3. Check whether `spec.template.spec.terminationGracePeriodSeconds` in the VM is `60`.

Case 5: The controller should wait for `terminationGracePeriodSeconds`, before really removing the VM.

1. Follow Case 4.
2. SSH to the VM and run following commands.
```
# sudo vim /etc/systemd/system/set_gracefulshutdown.service
[Unit]
Description=Set flag for graceful shutdown
DefaultDependencies=no
RefuseManualStart=true
Before=shutdown.target

[Service]
Type=oneshot
TimeoutStartSec=50
ExecStart= /bin/bash -c 'sleep infinity'

[Install]
WantedBy=shutdown.target

# sudo systemctl daemon-reload
# sudo systemctl enable set_gracefulshutdown
```
3. On VM dashboard, stop the VM. We can see OS will wait for 170 seconds and eventually shutdown at 180 seconds:
Screenshot 2023-05-18 at 3 45 46 PM

Case 6: Users can force stop the VM if they don't wait for a graceful shutdown.

1. Follow Case 5.
2. Enable Developer Tools & Features.
3. Start the VM again.
4. Stop the VM again.
5. Click "View in API" on the VM.
6. Click "forceStop" and the controller will stop the VM immediately.

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -32,6 +32,7 @@ const (
 	addVolume           = "addVolume"
 	removeVolume        = "removeVolume"
 	cloneVM             = "clone"
+	forceStopVM         = "forceStop"
 )
 
 type vmformatter struct {
@@ -105,6 +106,10 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 
 	if vf.canCreateTemplate(vmi) {
 		resource.AddAction(request, createTemplate)
+	}
+
+	if vf.canForceStop(vm) {
+		resource.AddAction(request, forceStopVM)
 	}
 }
 
@@ -299,4 +304,15 @@ func (vf *vmformatter) getVMI(vm *kubevirtv1.VirtualMachine) *kubevirtv1.Virtual
 		return vmi
 	}
 	return nil
+}
+
+func (vf *vmformatter) canForceStop(vm *kubevirtv1.VirtualMachine) bool {
+	if vm == nil {
+		return false
+	}
+
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusStopping {
+		return false
+	}
+	return true
 }
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -70,6 +70,11 @@ func (m *vmMutator) Create(request *types.Request, newObj runtime.Object) (types
 		return nil, err
 	}
 
+	patchOps, err = m.patchTerminationGracePeriodSeconds(vm, patchOps)
+	if err != nil {
+		return nil, err
+	}
+
 	return patchOps, nil
 }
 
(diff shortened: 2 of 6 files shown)


pkg/api/vm/formatter.go
```diff
@@ -284,3 +289,14 @@ func (vf *vmformatter) getVMI(vm *kubevirtv1.VirtualMachine) *kubevirtv1.Virtual
 	}
 	return nil
 }
+
+func (vf *vmformatter) canForceStop(vm *kubevirtv1.VirtualMachine) bool {
+	if vm == nil {
+		return false
+	}
+
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusStopping {
```
bk201, line 314: Do we also have the need to force stop a running VM?

guangbochen: ~~should we check if the VM is not stopped? also force stop a VM with stopping status should be supported.~~

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -373,3 +378,27 @@ func (m *vmMutator) addNodeSelectorTerms(defaultNamespace string, nodeSelector *
 
 	return nil
 }
+
+func (m *vmMutator) patchTerminationGracePeriodSeconds(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) (types.PatchOps, error) {
```
guangbochen, line 383: seems there is no global config on the kubevirt CR?

FrankYang0529: Yeah, I don't find global config on KubeVirt CR.

futuretea (approved):
LGTM, Thanks!

bk201 (approved):
LGTM!

guangbochen (approved):

PR #3453, opened 2023-02-14
Update installer version to v1.1 branch

bk201:
Fix installer code after tag v1.1.1 is not included.


guangbochen (approved):
LGTM, thanks.

Issue #3464, opened 2023-02-14
[BUG] NFS Backup Target `Driver nfs is not supported!`


irishgordo:
**Describe the bug**
When setting up a backup target via NFS - a failure is encountered.

**To Reproduce**
1. Go to Settings -> Backup Target -> Edit Setting
1. Select NFS
1. Click Save

**Expected behavior**
NFS be able to be configured as a backup-target without issue 

**Environment**
 - Harvester ISO version: Version: master-6fd9a5bb-head ( 02/10/23 master-head release )
 - Underlying Infrastructure: qemu/kvm

**Additional context**
PUT to https://harvester-vip-ip/v1/harvester/harvesterhci.io.settings/backup-target
Request:
```
{
  "id": "backup-target",
  "type": "harvesterhci.io.setting",
  "links": {
    "remove": "https://192.168.122.57/v1/harvester/harvesterhci.io.settings/backup-target",
    "self": "https://192.168.122.57/v1/harvester/harvesterhci.io.settings/backup-target",
    "update": "https://192.168.122.57/v1/harvester/harvesterhci.io.settings/backup-target",
    "view": "https://192.168.122.57/apis/harvesterhci.io/v1beta1/settings/backup-target"
  },
  "apiVersion": "harvesterhci.io/v1beta1",
  "kind": "Setting",
  "metadata": {
    "creationTimestamp": "2023-02-10T17:29:50Z",
    "fields": [
      "backup-target",
      null
    ],
    "generation": 2,
    "managedFields": [
      {
        "apiVersion": "harvesterhci.io/v1beta1",
        "fieldsType": "FieldsV1",
        "fieldsV1": {
          "f:status": {
            ".": {},
            "f:conditions": {}
          }
        },
        "manager": "harvester",
        "operation": "Update",
        "time": "2023-02-10T17:29:51Z"
      }
    ],
    "name": "backup-target",
    "relationships": null,
    "resourceVersion": "7719",
    "state": {
      "error": false,
      "message": "Resource is current",
      "name": "active",
      "transitioning": false
    },
    "uid": "f9ee218c-9ef3-47d5-80f8-8811212b5d0d"
  },
  "status": {
    "conditions": [
      {
        "error": false,
        "lastUpdateTime": "2023-02-10T17:29:51Z",
        "status": "False",
        "transitioning": false,
        "type": "configured"
      }
    ]
  },
[...]

Issue #3462, opened 2023-02-14
[BUG] S3 Backup Target Fails `Driver s3 is not supported!` 


irishgordo:
**Describe the bug**
When setting up a backup target via S3 - a failure is encountered.
Either in an attempt to set up a backup target with Minio or AWS S3.

**Pre-reqs**
1.  Either a Minio bucket or AWS S3 bucket

**To Reproduce**
1. Go to Settings -> Backup Target -> Edit Setting
1. Select S3, fill out details for either a Minio bucket or AWS S3 Bucket
1. Click Save

**Expected behavior**
S3 be able to be configured as a backup-target without issue for AWS S3 or Minio.

**Environment**
 - Harvester ISO version:  Version: master-6fd9a5bb-head ( 02/10/23 master-head release )
 - Underlying Infrastructure: qemu/kvm

**Additional context**
(S3 Req - Response shown)
A PUT to: `https://harvester-vip-ip/v1/harvester/harvesterhci.io.settings/backup-target`
Yields the response of:
```
{
  "type": "error",
  "links": {},
  "code": "Invalid",
  "message": "admission webhook \"validator.harvesterhci.io\" denied the request: Driver s3 is not supported!",
  "status": 422
}
```
With the request body of:
```
{
  "id": "backup-target",
  "type": "harvesterhci.io.setting",
  "links": {
    "remove": "https://192.168.122.57/v1/harvester/harvesterhci.io.settings/backup-target",
    "self": "https://192.168.122.57/v1/harvester/harvesterhci.io.settings/backup-target",
    "update": "https://192.168.122.57/v1/harvester/harvesterhci.io.settings/backup-target",
    "view": "https://192.168.122.57/apis/harvesterhci.io/v1beta1/settings/backup-target"
  },
  "apiVersion": "harvesterhci.io/v1beta1",
  "kind": "Setting",
  "metadata": {
    "creationTimestamp": "2023-02-10T17:29:50Z",
    "fields": [
      "backup-target",
      null
    ],
    "generation": 2,
    "managedFields": [
      {
        "apiVersion": "harvesterhci.io/v1beta1",
        "fieldsType": "FieldsV1",
        "fieldsV1": {
          "f:status": {
            ".": {},
            "f:conditions": {}
          }
        },
        "manager": "harvester",
        "operation": "Update",
        "time": "2023-02-10T17:29:51Z"
      }
    ],
    "name": "backup-target",
    "relationships": null,
    "resourceVersion": "7719",
[...]

PR #3477, opened 2023-02-16
fix: add nfs and s3 back to vendor

FrankYang0529:
**Problem:**
NFS and S3 were deleted in vendor, so we got `Driver nfs/s3 is not supported`.

**Solution:**
Add NFS and S3 back to vendor.

**Related Issue:**
https://github.com/harvester/harvester/issues/3464
https://github.com/harvester/harvester/issues/3462

**Test plan:**

* Case 1: Test backup / restore with NFS
* Case 2: Test backup / restore with S3


guangbochen (approved):
LGTM, and thanks for the fixing.

guangbochen:
The Trivy CI will be addressed via https://github.com/harvester/harvester/pull/3479 so we can ingore it.

bk201 (approved):
Thanks for adding more comments to explain why we need the blank imports.

FrankYang0529:
@Mergifyio backport v1.1

PR #3255, opened 2022-12-08
pvc cleanup fixes

ibrokethecloud:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**

In certain cases when the VM is created and deleted immediately the ownership references may not be cleaned up from the PVC. 
The harvester validation webhook, prevents deletion of such PVCs. The only options in such a scenario are to edit the PVC to drop the ownership annotation or remove the same from longhorn console.

The problem is likely caused by a race condition in the virtualmachine controller, which has an OnChange method and OnRemove method to SetOwnerReferences and RemovedOwnerReferences from the PVC's making up the VM.

**Solution:**

The PR introduces a possible fix, by merging the OnChange and OnRemove handlers into a single OnChange handler, along with custom finalizers to ensure resource cleanup.

**Related Issue:**
https://github.com/harvester/harvester/issues/2677

**Test plan:**

To replicate the issue, I used the following VM json and shell script to trigger the VM creation and deletion.

```json
{
    "apiVersion": "kubevirt.io/v1",
    "kind": "VirtualMachine",
    "metadata": {
        "annotations": {
            "harvesterhci.io/vmRunStrategy": "RerunOnFailure",
            "harvesterhci.io/volumeClaimTemplates": "[{\"metadata\":{\"name\":\"workload-1-disk-0-8p4bi\",\"annotations\":{\"harvesterhci.io/imageId\":\"default/image-98xqv\"}},\"spec\":{\"accessModes\":[\"ReadWriteMany\"],\"resources\":{\"requests\":{\"storage\":\"10Gi\"}},\"volumeMode\":\"Block\",\"storageClassName\":\"longhorn-image-98xqv\"}}]",
            "kubevirt.io/latest-observed-api-version": "v1",
            "kubevirt.io/storage-observed-api-version": "v1alpha3",
            "network.harvesterhci.io/ips": "[]"
        },
        "labels": {
            "harvesterhci.io/creator": "harvester",
            "harvesterhci.io/os": "ubuntu"
        },
        "name": "workload-1",
        "namespace": "default"
    },
    "spec": {
        "runStrategy": "RerunOnFailure",
        "template": {
            "metadata": {
                "annotations": {
                    "harvesterhci.io/sshNames": "[]"
                },
                "creationTimestamp": null,
                "labels": {
                    "harvesterhci.io/vmName": "workload-1"
                }
            },
            "spec": {
                "domain": {
                    "cpu": {
                        "cores": 2,
                        "sockets": 1,
[...]

Issue #2677, opened 2022-08-18
[BUG] Volumes bounded on a deleted VM and unable to be removed


lanfon72:
### Describe the bug
https://user-images.githubusercontent.com/5169694/185370936-e07979ed-0079-4754-b2d0-bcb9872a5ebe.mp4

It discovered by daily e2e testing.

### To Reproduce
1. Install Harvester with any nodes
2. Create a Image for VM creation
3. Create the VM `vm1` via API endpoint: `https://<CLUSTER_VIP>/apis/kubevirt.io/v1/namespaces/default/virtualmachines` (POST method)
4. Delete the VM `vm1` and volumes simultaneously. (DELETE method)
    - VM's API endpoint: `https://<CLUSTER_VIP>/apis/kubevirt.io/v1/namespaces/default/virtualmachines/<VM_NAME>`
    - Volume's API endpoint: `https://<CLUSTER_VIP>/api/v1/namespaces/default/persistentvolumeclaims/<VOLUME_NAME>`

### Expected behavior
Volumes should able to be removed

## Environment:
 - Harvester ISO version: **v1.0.3**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 4 nodes**

### Additional context
Dashboard's operation will NOT encountered this.

The issue can be reproduced by running `test_create_vm_do_not_start` e2e tests several times.

bk201:
@lanfon72 I guess this is expected since the volume is still used.

If this blocks some QA tests, perhaps it's better to use calls like:

```
DELETE /v1/harvester/kubevirt.io.virtualmachines/default/test-vm-1?removedDisks=disk-0
```

Or add some code to wait for the volume to be unbound.

lanfon72:
@bk201 Volumes attached VM has been removed, but volumes still here and can not be deleted.

this is not always reproduced, and only happened when the VM create and delete simultaneously.

Martin-Weiss:
I have seen this also when using the terraform provider for Harvester to create and then destroy VMs..

bk201:
The owner annotation in volume is not removed. Some investigation:
```
harvester-lzxcs:~ # kubectl get pvc
NAME                                            STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS                                AGE
[...]

PR #3508, opened 2023-02-21
additional logic to remove legacy wrangler finalizer on VM

ibrokethecloud:

**Problem:**

PR https://github.com/harvester/harvester/pull/3255 changed how we apply and remove custom finalizers on VM's to ensure PVC's are cleaned up.

Since we are no longer using wrangler OnRemove handler for the vm controller, on older installations, post upgrade this will not finalizer will not be removed, which will cause VM's to be stuck during deletion.

**Solution:**

This PR introduces an additional check to remove the old Wrangler finalizer if present on the VM.

**Related Issue:**
https://github.com/harvester/harvester/issues/2677

**Test plan:**


futuretea (approved):
LGTM, thanks

bk201 (approved):
LGTM

guangbochen (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.1

Issue #3550, opened 2023-02-24
[BUG] Upgrade failed with "no route to host" error while retrieving repo info


starbops:
**Describe the bug**

This one is similar to #3444 but has an uncaught networking error: "no route to host."

<img width="762" alt="Screen Shot 2023-02-24 at 18 18 32" src="https://user-images.githubusercontent.com/1827717/221155001-f7d4141f-fb5d-435d-b892-8110076d8782.png">

The log lines from the `harvester` Pod:

```
...
time="2023-02-24T07:58:41Z" level=info msg="Starting upgrade repo VM"
time="2023-02-24T07:58:41Z" level=info msg="handle upgrade harvester-system/hvst-upgrade-vh6bs with labels map[harvesterhci.io/latestUpgrade:true harvesterhci.io/upgradeState:PreparingRepo]"
W0224 07:58:56.281689       7 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
time="2023-02-24T07:59:37Z" level=info msg="handle upgrade harvester-system/hvst-upgrade-vh6bs with labels map[harvesterhci.io/latestUpgrade:true harvesterhci.io/upgradeState:RepoPrepared]"
time="2023-02-24T07:59:44Z" level=warning msg="Repo info retrieval failed with: Get \"http://upgrade-repo-hvst-upgrade-vh6bs.harvester-system/harvester-iso/harvester-release.yaml\": dial tcp 10.53.230.35:80: connect: no route to host"
time="2023-02-24T07:59:44Z" level=info msg="handle upgrade harvester-system/hvst-upgrade-vh6bs with labels map[harvesterhci.io/latestUpgrade:true harvesterhci.io/upgradeState:Failed]"
time="2023-02-24T07:59:44Z" level=info msg="Add OwnerReference upgrade-repo-hvst-upgrade-vh6bs-disk-0 to upgrade repo image harvester-system/image-stlwr"
time="2023-02-24T07:59:44Z" level=info msg="Delete upgrade repo VM harvester-system/upgrade-repo-hvst-upgrade-vh6bs"
...
```

**To Reproduce**
1. Deploy a Harvester cluster (single node is okay) with master-head build
2. Upgrade with the same version ISO image
3. After the repo VM is up, chances are the upgrade will fail (not always reproducible)

**Expected behavior**

The upgrade should succeed.

**Support bundle**

**Environment**
 - Harvester ISO version: master
[...]

Issue #3444, opened 2023-02-10
[BUG] Can't upgrade from master version to master version


bk201:
**Describe the bug**

Upgrading to the same version fails (with the current master code).

**To Reproduce**
1. Build an ISO with the current master code (6fd9a5bb3b99a329101ec70f657a30e24906035f) and deploy a cluster.
2. Upgrade with the ISO.
3. The upgrade fails with

    ![image](https://user-images.githubusercontent.com/1691518/218121471-fb0d21b0-c7e2-41cf-ba46-e49d08978634.png)

**Expected behavior**

The upgrade should proceed and succeed.

**Support bundle**

**Environment**
 - Harvester ISO version: 6fd9a5bb3b99a329101ec70f657a30e24906035f
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

starbops:
By taking several rounds of testing, we noticed that the upgrade could only pass the first and second phases, i.e., `ImageReady` and `RepoReady`.

```
  status:
    conditions:
    - lastUpdateTime: "2023-02-13T03:45:41Z"
      reason: 'Get "http://upgrade-repo-hvst-upgrade-rxcvc.harvester-system/harvester-iso/harvester-release.yaml":
        dial tcp 10.53.200.90:80: connect: connection refused'
      status: "False"
      type: Completed
    - lastUpdateTime: "2023-02-13T03:44:44Z"
      status: "True"
      type: ImageReady
    - lastUpdateTime: "2023-02-13T03:45:41Z"
      status: "True"
      type: RepoReady
```

Soon it failed at the beginning of the `NodesPrepared` phase: the upgrade controller tried to get the information from the repo VM it spawned up but failed with the "connection refused" error, thus it marked the upgrade as failed by setting the `Completed` condition to "False".

https://github.com/harvester/harvester/blob/6fd9a5bb3b99a329101ec70f657a30e24906035f/pkg/controller/master/upgrade/upgrade_controller.go#L168-L172

https://github.com/harvester/harvester/blob/6fd9a5bb3b99a329101ec70f657a30e24906035f/pkg/controller/master/upgrade/upgrade_repo.go#L418-L429

Then the repo VM is purged by the upgrade controller immediately.

https://github.com/harvester/harvester/blob/6fd9a5bb3b99a329101ec70f657a30e24906035f/pkg/controller/master/upgrade/upgrade_controller.go#L143-L146

[...]

PR #3552, opened 2023-02-24
Mark network unreachable as retriable error

starbops:

**Problem:**

The upgrade will fail at retrieving repo info with "no route to host" error occasionally.

**Solution:**

In the previous fix, we only retry on "timed out" and "connection refused" errors. The "network unreachable" error should also be included.

**Related Issue:**

#3444
#3550

**Test plan:**

1. Deploy a Harvester cluster (single node is okay) with the fix
2. Upgrade with the same ISO image that used to install the cluster
3. The upgrade should succeed

### DIFF
--- a/pkg/util/errors.go
+++ b/pkg/util/errors.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"net/url"
 	"syscall"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/util/errors.go
```diff
@@ -36,12 +35,13 @@ func NewStillExists(qualifiedResource schema.GroupResource, name string) *apierr
...
-func IsConnectionRefusedOrTimeout(err error) bool {
-	urlErr := err.(*url.Error)
-	if urlErr.Timeout() {
+func IsRetriableNetworkError(err error) bool {
+	if errors.Is(err, syscall.ETIMEDOUT) {
 		return true
 	} else if errors.Is(err, syscall.ECONNREFUSED) {
 		return true
+	} else if errors.Is(err, syscall.EHOSTUNREACH) {
+		return true
```
w13915984028, line 50: we may consider add those erorrs also into retryiable network error
```
ENETUNREACH
ENETDOWN
ECONNRESET
```

starbops: Updated, thanks.

bk201 (approved):
LGTM!

w13915984028 (approved):
LGTM, thanks.

bk201:
@mergifyio backport v1.1

PR #3558, opened 2023-02-25
WIP HEPTAPODS [CI SKIP] [FIXES #2707]

### LINKED ISSUE #2707: [FEATURE] increase 110 pods per node limit
## problem
k8s has default limit:  110 pods per node.
My harvester control node with 20 vm already used 100 pods now.

####  pod numbers by namespaces:
* cattle  13 pods
* kube-system 19 pods
* longhorn 29 pods
* harvester-system 15 pods
* user vm only 24 pods 

#### harvester k8 config shows 110 limit
```
  allocatable:
    cpu: "40"
    memory: 395731228Ki
    pods: "110"
  capacity:
    cpu: "40"
    memory: 395731228Ki
    pods: "110"
```

## Describe the solution you'd like
* need an easy way to increase the limit number 110 -> 200
* or make 200 as the default harvester limit

#### for normal k8s cluster
* kubelet args: `--max-pods 200`
* kubelet config file: `/var/lib/rancher/rke2/agent/kubelet.kubeconfig`
* rancher kubelet config extra_args:  `max-pods: 250`

## Describe alternatives you've considered
any chance to reduce harvester/rancher/longhorn pods?

tlehman:
**Problem:**
(Increase max-pods limit) https://github.com/harvester/harvester/issues/2707

**Solution:**
https://github.com/tlehman/harvester/blob/hep-ta-pods/enhancements/20230224-max-pods.md#max-pods-configuration-in-harvester

**Related Issue:**
https://github.com/harvester/harvester/issues/2707

**Test plan:**
TODO

### DIFF
--- a/enhancements/20230224-max-pods.md
+++ b/enhancements/20230224-max-pods.md
@@ -1,0 +1,156 @@
+# Max Pods Configuration in Harvester
+
+This is the **H**arvester **E**nhancement **P**roposal for the **T**opmost **A**mount of **Pods** or HEPTAPODS proposal.
+
+## Summary
+
+It started innocently, as a request to increase the per-node pod limit from 110 to 200. Then the issue of CIDR (Classless Inter-Domain Routing) came up, and the potential problem that allowing the number of pods to increase could cause an overlap between non-Pod IPs and the Cluster CIDR-defined range of IPs.
+
+CIDR Defaults as of Harvester 1.1.1:
+
+- The default cluster CIDR: `10.52.0.0/16`
+    - Range: (10.52.0.1-10.52.255.254)
+    - Length: 65,536 = $2^{32-16} = 2^{16}$
+- The default service CIDR: `10.53.0.0/16`
+    - Range: (10.53.0.1-10.53.255.254)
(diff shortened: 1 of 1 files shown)


Martin-Weiss:
As a general question - what do you do in case the customer has a NTP server or DNS server i.e. on 10.52.123.12 "outside" of harvester?

-> as part of increasing the max number of pods on a node - I guess we need to allow customers to change the CIDRs used "in general".

tlehman:
> As a general question - what do you do in case the customer has a NTP server or DNS server i.e. on 10.52.123.12 "outside" of harvester?
> 
> -> as part of increasing the max number of pods on a node - I guess we need to allow customers to change the CIDRs used "in general".

Great question, let's assume the external IPs are unchangeable, I'll add the ability to introduce custom CIDR ranges. If so, we should also have some validation to avoid overlap in the cluster. I'll look for any prior art here.

tlehman:
Closing this proposal after a meeting. In the meeting we worked through what supporting the live change and upgrade path would require. It's not worth building this and committing the resources to testing it. It suffices to raise the default to 200 (confirmed that this would not have CIDR implications).

Issue #3404, opened 2023-02-02
[BUG] fix invalid havester-csi-ccm-versions setting


WuJun2016:
**Describe the bug**

The current default value for the `harvester-csi-ccm-versions` setting is invalid because it restricts to an invalid maximum version, and prevents users from successfully selecting the Harvester cloud provider when creating or modifying the RKE2 Kubernetes cluster.

Original issue： - https://github.com/harvester/harvester/issues/3389

Impacted version: Rancher >= v2.6.9, v2.7.0 + RKE2 with Harvester CCM >= v0.1.15 (starting from RKE Year 23 Jan. Releases)

`v1/harvester/harvesterhci.io.settings/harvester-csi-ccm-versions`   

![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/1de4e9a3-3ee6-4e0f-8574-757a69d8f2e0)

**To Reproduce**
1. import Harvester (<= v1.1.2-rc3) into a upstream Rancher (v2.6.9 or v2.7-head)
2. trying to create a new RKE2 cluster with newer RKE2 versions
3. user will not be able to select the `Harvester` cloud provider as its default `harvester-csi-ccm-versions` setting version is less than the RKE2's built-in one.

**Expected behavior**
Users should still be able to select the `Harvester` cloud provider as the latest RKE2's built-in Harvester CSI & CCM chart is compatible with Harvester v1.1 clusters.
- [ ] remove `do not check when the Harvester version is <v1.2` constrain on the upstream Rancher UI (starting from v2.7.2) @n313893254 
- [ ] need to update the [SUSE Harvester support matrix ](https://www.suse.com/suse-harvester/support-matrix/all-supported-versions/harvester-v1-1-1/)of the above changes

**Support bundle**

**Environment**
 - Harvester ISO version: Harvester >= v1.1.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

TachunLin:
#### Test plan 
Ensure the Rancher Kubernetes version can support CSI & CCM maximum versions of the following scenarios: 

| Harvester version | RKE2 with old CCM design | RKE2 with new CCM design( v0.2 - https://github.com/harvester/harvester/issues/2134) |
| -- | -- | -- |
| `>= v1.2` | Supported | Supported |
| `< v1.2` | Supported | Not Supported |

#### Test Environment
Harvester: v1.2.0 and v1.1.2
[...]

PR #3592, opened 2023-03-06
Remove max version in settings for harvester-cloud-provider and harvester-csi-provider.

masteryyh:

**Problem:**

Need to remove max version in settings.go for harvester-cloud-provider and harvester-csi-provider, which could leads to problems.

**Solution:**

Remove max version in settings.go for harvester-cloud-provider and harvester-csi-provider

**Related Issue:**
https://github.com/harvester/harvester/issues/3404


futuretea (approved):
LGTM, Thanks

Issue #3597, opened 2023-03-06
[BUG] VM state stuck in *Restoring* State but actually in *Running* State


lanfon72:
### Describe the bug
https://user-images.githubusercontent.com/5169694/223130767-87e649c0-049d-46da-b50f-8cca85f8d8ff.mp4

### To Reproduce
1. Install Harvester with any nodes
2. Login to Dashboard, Setup **backup-target** in _Advanced/Settings_
3. Create an image for VM creation
4. Create VM `vm1`
5. Backup `vm1` into `vm1b`
6. Restore `vm1b` into `vm1n` and delete `vm1b` immediately

### Expected behavior
- VM's state should reflect to its real state
- backup-target should able to change

## Environment:
 - Harvester ISO version: **v1.1-a002d078-head**
 - **ui-source** Option: **Auto**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **qemu/KVM 3 nodes**

### Additional context
- I didn't tested it on a large backed up VM
- There have a simple workaround as:
    1. Take a back up for `vm1n` into `vm1b`
    2. Delete `vm1n`

bk201:
@FrankYang0529 Please help check, thanks!

lanfon72:
Verified this bug has been fixed.
![image](https://user-images.githubusercontent.com/5169694/223704906-61b7dd08-1119-4eb9-ab2d-dea36c682110.png)


Test Information
----
* Environment: **qemu/KVM 2 nodes**
* Harvester Version: **v1.1-0fe9676b-head**
* **ui-source** Option: **Auto**

### Verify Steps:
1. Follow **Step to reproduce** in https://github.com/harvester/harvester/issues/3597#issue-1611507033

PR #3606, opened 2023-03-07
fix(vmbackup): prevent users from deleting vmbackup when there is vmrestore in progress

FrankYang0529:
**Problem:**
Users may delete VMBackup when there is VMRestore in progress. This makes VMRestore controller fail.

**Solution:**
Prevent users from doing it.

**Related Issue:**
https://github.com/harvester/harvester/issues/3597

**Test plan:**
1. Create a VM.
2. Take a backup.
3. Restote the backup to a new VM.
4. Beofer the new VM is done, delete the backup. Webhook should deny the request.

### DIFF
--- a/pkg/webhook/indexeres/indexer.go
+++ b/pkg/webhook/indexeres/indexer.go
@@ -8,8 +8,9 @@ import (
 )
 
 const (
-	VMBackupBySourceUIDIndex          = "harvesterhci.io/vmbackup-by-source-uid"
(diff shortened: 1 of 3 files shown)


guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM, the feature works well!

bk201:
@mergifyio backport v1.1

Issue #3124, opened 2022-11-08
[FEATURE] Full or partial Quota usage prevents Live Migration of VM when additional VM exceeds Quota


rebeccazzzz:
**Describe the bug**

Fully or partially utilizing the Quota set on the Project and/or Namespace prevents the Live Migration of VMs.

The example below illustrates the problem and prevents the Live Migration of a VM.
* Quota: 64GB Memory
* VMs: 3 x 16GB Memory Nodes (together 48GB)

The expectation is that Live Migration could be done, but unfortunately the below message was seen.

/// 
(combined from similar events): Error creating pod: pods "virt-launcher-awhs-int-pool1-0c06dbc6-c7wnq-7rdpv" is forbidden: exceeded quota: default-gmp6p, requested: limits.memory=17405661185, used: limits.memory=52216983555, limited: limits.memory=64Gi
 ///
 
Also, even when a user fully utilizes its quota, the expectation is that the existing VMs are allowed to live migrate to another node.

**Expected behavior**
* Actual behavior:  Live migration not possible when quota is set
* Expected behavior:  Live migration should be possible with partial or full quota utilization

chrisho:
After creating VMs, because the virt-component in the Pod requires a certain memory overhead, resulting in a higher actual quota of Pod memory than expected from the VM configuration, however, there is a possibility that the number of VMs created by the user or the number of simultaneous migrations will not reach the expectation.

Example:
The user configures the ResourceQuota of 10c and 10Gi, and the maintenance quota is 50% each, i.e. VM Available Resource and Maintenance Available Resource are 5c and 5Gi each. Although the user can create 5 VMs at this time, the 5 VMs cannot be migrated at the same time because the total memory used by the 5 VMs is more than 5Gi, so only 4 VMs can be migrated at the same time.
In general, Maintenance Resources will only be overused in accordance with general usage.

solution: https://github.com/chrisho/harvester/blob/maintenance-resource-hep/enhancements/20230228-resource-quota-enhancement.md#vm-overhead-resource

lanfon72:
Verified this feature has been implemented.


Test Information
----
* Environment: **Baremetal DL360 3 nodes**
* Harvester Version: **v1.2.0-rc1**
* **ui-source** Option: **Auto**
* Rancher Version:
    - **v2.6-a4b924d21cd5d90b28ab8009edc18d96542952ac-head**, Docker Image: **a1d1e43988d2**
[...]

PR #3623, opened 2023-03-09
HEP: Resource Quota Enahancement [CI SKIP]

chrisho:

**Problem:**
Full or partial Quota usage prevents Live Migration of VM when additional VM exceeds Quota.

**Solution:**
https://github.com/chrisho/harvester/blob/maintenance-resource-hep/enhancements/20230228-resource-quota-enhancement.md

**Related Issue:**
https://github.com/harvester/harvester/issues/3124

**Test plan:**


chrisho:
Re-design

PR #3638, opened 2023-03-13
Switch installer to v1.1 again

bk201:
Switch the installer back to v1.1 again, [we'll fix this by using release branches](https://github.com/harvester/harvester/issues/3639).


guangbochen (approved):
LGTM, thanks.

Issue #3616, opened 2023-03-08
[BUG] Harvester single node upgrade will get `another operation (install/upgrade/rollback) is in progress` error


guangbochen:
**Describe the bug**
Harvester single node upgrade will encounter `another operation (install/upgrade/rollback) is in progress` error after node reboot. Therefore, it will block the next Harvester upgrade or managedChart update. Potentially related to https://github.com/helm/helm/issues/8987#issuecomment-786149813

**To Reproduce**
1. Install a Harvester cluster with an old version, e.g., v1.1.1
2. upgrade the Harvester cluster to a newer version, e.g., v1.1.2-head.iso
3. after the upgrade is complete, the upgrade status shows success, then check the `harvester` managedChart status and it will contain an error of:
```
conditions:
  - lastUpdateTime: "2023-03-08T05:42:32Z"
    message: 'ErrApplied(1) [Cluster fleet-local/local: another operation (install/upgrade/rollback)
      is in progress]; daemonset.apps harvester-system/kube-vip [progressing] Available:
      0/1; kubevirt.kubevirt.io harvester-system/kubevirt [progressing] Deployin
```

**Expected behavior**
single node upgrade should not contain the above error.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.1.1 upgrade
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
Workaround can be referred to https://github.com/helm/helm/issues/8987#issuecomment-786149813; however, ensure the rollback version is correct.

bk201:
@w13915984028 I guess this is not completely addressed or perhaps a new regression.

w13915984028:
@bk201 looks so, this issue still happens, I will look into fleet-agent's new code when free.

@guangbochen do you have a support-bundle file ?

w13915984028:
reproduced in a single-node upgrade:

from the helm release history, the `harvester` chart is upgraded several times inbetween upgrade, the last 3 times are in almost same time, will continue check why this happens.

```
harv31:~ # kk get bundle -A
NAMESPACE     NAME                                          BUNDLEDEPLOYMENTS-READY   STATUS
fleet-local   fleet-agent-local                             1/1                       
fleet-local   local-managed-system-agent                    1/1                       
[...]

PR #3641, opened 2023-03-13
Upgrade: gracefully shutdown fleet-agent in the single node case

bk201:
**Problem:**
In a single node upgrade case, fleet agent could be terminated unexpectedly due to the upgrade of kube-api server. 
This leaves managed charts (bundles) in the pending status.

**Solution:**
Bring down fleet-agent before upgrading RKE2 and enable it again when the upgrade is done.

**Related Issue:**
https://github.com/harvester/harvester/issues/3616

**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -586,6 +586,13 @@ command_single_node_upgrade() {
   # Add logging related kube-audit policy file
   patch_logging_event_audit
 
+  # During the RKE2 upgrade, the kube-api server will be restarted and fleet-agent will restart.
+  # The fleet agent might not clear applying charts state in this case.
+  # We stop the agent and let it gracefully handle the clearn up.
+  echo "Scale down the fleet agent..."
+  kubectl scale --replicas=0 deployment/fleet-agent -n cattle-fleet-local-system
+  kubectl rollout status deployment fleet-agent -n cattle-fleet-local-system
+
   # Upgarde RKE2
   upgrade_rke2
   wait_rke2_upgrade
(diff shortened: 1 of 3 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -586,6 +586,13 @@ command_single_node_upgrade() {
   # Add logging related kube-audit policy file
   patch_logging_event_audit
 
+  # During the RKE2 upgrade, the kube-api server will be restarted and fleet-agent will restart.
+  # The fleet agent might not clear applying charts state in this case.
+  # We stop the agent and let it gracefully handle the clearn up.
+  echo "Scale down the fleet agent..."
+  kubectl scale --replicas=0 deployment/fleet-agent -n cattle-fleet-local-system
```
w13915984028, line 593:
```
  name: fleet-agent-86dc84bdb7-pvw8r
  namespace: cattle-fleet-local-system


  serviceAccountName: fleet-agent
  terminationGracePeriodSeconds: 30
```

We can either scale it down, or simply wait all `bundles` to be ready. The `scale-down` may still have issue, given the grace period is only 30s.

will test it.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -302,6 +304,24 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
 		return err
 	}
 
+	// For single node case, we scale fleet agent deployment replicas to 0 before upgrading RKE2.
+	// Restore the replica back when upgrade is done.
+	if upgrade.Status.SingleNode != "" {
+		fleetAgentDP, err := h.deploymentClient.Get("cattle-fleet-local-system", "fleet-agent", metav1.GetOptions{})
```
w13915984028, line 310: one possible drawback of using controller is  (but we do not have better choice now):

if the upgrade is deleted somehow, could the fleet-agent be left with replicas=0 ?

then we need to add this to documents, user needs to recover it manually

bk201: `cleanup` is called when an upgrade succeeds/fails or is deleted. https://github.com/harvester/harvester/blob/master/pkg/controller/master/upgrade/upgrade_controller.go#L295

Issue #2679, opened 2022-08-19
[FEATURE] Add bare-metal container and MCM support


andrewbabbittdev:
*Sorry if an existing ticket exists for this, hard to find anything similar with search*

**Is your feature request related to a problem? Please describe.**
Currently, the [rancher integration](https://docs.harvesterhci.io/v1.0/rancher/rancher-integration/) requires you to either have rancher deployed elsewhere, as a VM inside of harvester using docker (not recommended), or deployed in a cluster under harvester which you would have to manually deploy.

This is not ideal for many edge and lab oriented scenarios where you may not be running an external rancher instance. Dedicating additional resources to a cluster just for running rancher is non ideal in more resource constrained environments.

**Describe the solution you'd like**
[Rancher is already embedded in harvester](https://docs.harvesterhci.io/v1.0/troubleshooting/harvester/#access-embedded-rancher), it would be nice if cluster oriented features could be exposed to make it easy to deploy rke and k3s clusters directly through the harvester UI.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

yasker:
Depend on #2769 .

Also, the reason we don't support using embedded Rancher is it's very hard for the Harvester release to bind with the Rancher release. For Harvester, only a small subset feature of Rancher is used, e.g. managing node provisioning (RKE2) and managing upgrades process.

If we're going to support embedded Rancher, we would have to release Harvester every time when Rancher has a new release, thus resulting in the binding of the Harvester and the Rancher versions. It would be very challenging for the engineering team.

But we acknowledge there is a need to have a built-in Rancher. Currently, we're considering achieving it via #2769, like via creating one-click deployment for the Rancher HA cluster in Harvester. The embedded Rancher will still be there but it would still be reserved for debugging purposes only.

davidspek:
For the time being, you can have a look at https://github.com/rancher/quickstart/pull/218.

ibrokethecloud:
# Test plan:

The test plan is split into two parts. 

1. Install vcluster addon
2. Use vcluster deployed rancher to manage container workloads

If a user already has an external rancher cluster, managing harvester then vcluster addon is not needed, and we can directly validate container workload view. To leverage container workload view the minimal rancher version needs to be v2.7.5

## vcluster addon:
* an experimental `rancher-vcluster` addon is available at: https://github.com/harvester/experimental-addons/tree/main/rancher-vcluster
[...]

PR #3642, opened 2023-03-13
fix(embedded_rancher): trim spaces

FrankYang0529:
**Problem:**
There are remaining spaces which may make `CA_CHECKSUM` error.

**Solution:**
Trim the remaining spaces.

**Related Issue:**

**Test plan:**
2. Run `curl --insecure -sfL https://<vip>/v3/settings/cacerts | jq -r '.value | select(length > 0)'`. There should not have `\n` in the end of line.

### DIFF
--- a/pkg/controller/master/rancher/embedded_rancher.go
+++ b/pkg/controller/master/rancher/embedded_rancher.go
@@ -91,6 +91,9 @@ func (h *Handler) syncCACert(setting *rancherv3api.Setting) error {
 		}
 		cacert = internalCACerts.Value
 	}
+	// Trim spaces to avoid checksum error in rancher.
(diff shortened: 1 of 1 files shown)


bk201:
@FrankYang0529 Could we add a reference to the Rancher code which might cause the issue, thanks.

FrankYang0529:
> @FrankYang0529 Could we add a reference to the Rancher code which might cause the issue, thanks.

Updated it. Thanks.

guangbochen (approved):
LGTM, thanks.

futuretea (approved):
LGTM, thanks

bk201 (approved):

PR #3646, opened 2023-03-14
fix(upgradelog): direct copy truncated log files

starbops:

**Problem:**

It is possible that the following happened when users want to download the upgrade log archive after the upgrade has finished. The download button keeps greying out and there's no file returned.

image

And from the packager's logs we can discover the packaging has some problems

image

This is due to the incomplete log line at the end of that specific file, and that prevents the packager from processing the file using `jq` since it's not in a valid JSON format (truncated).

**Solution:**

Keep the log in its original format (JSON) and let the packaging continue when processing those kinds of files.

**Related Issue:**

#1926

**Test plan:**

1. Provision a v1.1.2-rc3 single-node cluster
2. Upgrade to master-head
3. After the upgrade ended, try to download the upgrade log archive, it should succeed

### DIFF
--- a/pkg/api/upgradelog/handler.go
+++ b/pkg/api/upgradelog/handler.go
@@ -41,10 +41,11 @@ cd /archive/logs
 
 for f in *.log
 do
-    cat $f | awk '{$1=$2=""; print $0}' | jq -r .message > $tmpdir/logs/$f
-	if [ $? -eq 4 ]; then
-	    echo "Incomplete JSON format log line, abort processing the file $f."
-	    continue
+	awk '{$1=$2=""; print $0}' $f | jq -r .message > $tmpdir/logs/$f || ret=$?
+	if [ -n "$ret" ]; then
+		echo "Failed to process the file $f with ret=$ret"
+		echo "Copy the original file to the destination"
+		cp $f $tmpdir/logs/$f
 	fi
 done
 


bk201 (approved):
lgtm

pkg/api/upgradelog/handler.go
```diff
@@ -41,10 +41,13 @@ cd /archive/logs
 
 for f in *.log
 do
-    cat $f | awk '{$1=$2=""; print $0}' | jq -r .message > $tmpdir/logs/$f
-	if [ $? -eq 4 ]; then
-	    echo "Incomplete JSON format log line, abort processing the file $f."
-	    continue
+	cat $f | awk '{$1=$2=""; print $0}' | jq -r .message > $tmpdir/logs/$f
```
w13915984028: normally, it is not necessary to `cat ` and then awk, use awk to work directly against the file 

`awk '{$1=$2=""; print $0}' $f | jq ... `

pkg/api/upgradelog/handler.go
```diff
@@ -30,7 +30,7 @@ const (
 	defaultJobBackoffLimit int32 = 5
 	logPackagingScript           = `
 #!/usr/bin/env sh
-set -ex
+set -x
```
w13915984028: may not be a good idea to remove `-e`, it will hide other potential errors

to escape and check following error, it could be:

`awk '{$1=$2=""; print $0}' $f | jq ... || ret=$?`

then check `ret`, you will know if error happens in `awk & jq`

pkg/api/upgradelog/handler.go
starbops: Updated, thanks!

pkg/api/upgradelog/handler.go
starbops: Agreed on the `-e` part. But for the `ret` part, I think it's more simple and clear to keep the original style since using `||` is a short-circuit. If the previous command is executed successfully, there will be no assignment for `ret` variable, hence causing trouble to the following `if` statement.

starbops: You're right, catching the return code after `||` is necessary since the error-prone command is the last one in the pipeline. I've updated the code, please take a look. Thanks!

Vicente-Cheng (approved):
LGTM, thanks!

w13915984028 (approved):
LGTM, thanks

Vicente-Cheng:
@mergify backport v1.1

Issue #3648, opened 2023-03-14
[BUG] 3 Node AirGap Harvester w/ AirGap Rancher v2.6.9 Fails v1.1.1 -> v1.1.2-rc3 Upgrade


irishgordo:
**Describe the bug**
3 node airgap harvester cluster w/ Rancher v2.6.9 fails upgrade 

**To Reproduce**
1. Create airgap 3 node Harvester 1 rancher node
1. Have file server on Rancher node
1. create version.yaml, process upgrade

**Expected behavior**
Node not to be stuck in `pre-drained` state

**Environment**
 - Harvester ISO version: v1.1.1 -> v1.1.2-rc3
 - Underlying Infrastructure: qemu/kvm
 
**Additional context**
![Screenshot from 2023-03-08 18-22-00](https://user-images.githubusercontent.com/5370752/225095030-5db4de56-14cc-4798-b1e6-8b4312f388e4.png)

TachunLin:
In the latest trial on 3 nodes fully airgapped upgrade from v1.1.1 to v1.1-head

The upgrade process stuck in `Pre-drained` state of the first node
![image](https://user-images.githubusercontent.com/29251855/225567775-c74dc1f1-b3c5-4cf3-b8b1-3d9619e4cd19.png)


* Attach the support bundle for the reference.
[supportbundle_cf38f094-f41b-41c6-8727-822b498a7fb1_2023-03-16T08-38-38Z.zip](https://github.com/harvester/harvester/files/10988755/supportbundle_cf38f094-f41b-41c6-8727-822b498a7fb1_2023-03-16T08-38-38Z.zip)

irishgordo:
The latest trial on 3 nodes fully airgapped no proxy with the loadout of:
Rancher-v2.6.11
Harvester-v1.1.1 -> with upgrade to -> v1.1.2-rc3
Failed in the same state

![Screenshot from 2023-03-16 09-26-16](https://user-images.githubusercontent.com/5370752/225691885-7a02b9b4-5e28-4eed-8927-d59b0d20ca12.png)
![Screenshot from 2023-03-16 09-24-47](https://user-images.githubusercontent.com/5370752/225691888-40714cf5-7b3e-4613-83fe-c7998b899eb2.png)
![Screenshot from 2023-03-15 18-02-13](https://user-images.githubusercontent.com/5370752/225691889-dc17138a-7f9f-4e41-adf8-cbf9ef56e3e3.png)
[...]

PR #3670, opened 2023-03-17
Detach volumes when there is no workload or in-progress VMBackup/VMRestore

FrankYang0529:
**Problem:**
When users created a backup on a shutdown VM, the upgrade controller cannot drain the node because the volume is attached on the node.

**Solution:**
If the VM is shutdown, detach volumes after taking a VM backup.

**Related Issue:**
- https://github.com/harvester/harvester/issues/3648
- https://github.com/harvester/harvester/issues/3681

**Test plan:**

### Setup: Mount volumes without a running workload or in-progress VMBackup/VMRestore.

1. Create the master-head Harvester.
2. Create `svm-backup` and `svm-snapshot` VMs.
3. Stop two VMs.
4. Take a backup on `svm-backup` and a snapshot on `svm-snapshot`.
5. VMs are not running, but their volumes are mounted.

### Case 1: Mounted volumes without a running workload or in-progress VMBackup/VMRestore will be detached.

1. Use this branch to build a new image and replace the image in `harvester/harvester` deployment.
2. Volumes from `svm-backup` and `svm-snapshot` VM will be detached.

### Case 2: The detaching volumes controller takes care of race conditions.

1. Follow Case 1.
2. Take a new backup for `svm-backup`.
3. Before the backup from the step 2 is finished. Take another backup for `svm-backup`.
4. All backups can be finished.
5. After backups are finished, volumes will be detached again because `svm-backup` is stopped.

### Case 3: Restore a new VM from a backup can work.

1. Follow Case 2.
2. Use any backup to restore the VM. Volumes from `svm-backup` will not be attached.
3. Restored VM has correct data and works fine.

### Case 4: VM snapshot and restore from the snapshot work fine

1. Create a new snapshot from `svm-snapshot`. Volumes from `svm-snapshot` will be attached.
2. After the snapshot is done, volumes from `svm-snapshot` will be detached.
3. Restore a new VM from the snapshot. Volumes from `svm-snapshot` will be attached again.
4. After the restored VM is done, volumes from `svm-snapshot` will be detached again.

### Case 5: Volume snapshot works fine

1. Create a new volume.
2. Take a snapshot for this volume. It will be attached.
3. After the snapshot is done, the volume will be detached.
4. Restore the snapshot to a new volume, the source volume will be attached again.
5. After the restored volume is done, the source volume will be detached again.

### DIFF
--- a/pkg/api/volumesnapshot/handler.go
+++ b/pkg/api/volumesnapshot/handler.go
@@ -8,14 +8,18 @@ import (
 
 	"github.com/gorilla/mux"
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
+	lhv1beta1 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	"github.com/rancher/apiserver/pkg/apierror"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	ctlstoragev1 "github.com/rancher/wrangler/pkg/generated/controllers/storage/v1"
 	"github.com/rancher/wrangler/pkg/schemas/validation"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	ctllonghornv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta1"
 	ctlsnapshotv1 "github.com/harvester/harvester/pkg/generated/controllers/snapshot.storage.k8s.io/v1beta1"
 	"github.com/harvester/harvester/pkg/util"
 )
@@ -23,6 +27,8 @@ import (
 type ActionHandler struct {
 	pvcs              ctlcorev1.PersistentVolumeClaimClient
 	pvcCache          ctlcorev1.PersistentVolumeClaimCache
+	volumes           ctllonghornv1.VolumeClient
+	volumeCache       ctllonghornv1.VolumeCache
 	snapshotCache     ctlsnapshotv1.VolumeSnapshotCache
 	storageClassCache ctlstoragev1.StorageClassCache
 }
@@ -124,5 +130,40 @@ func (h *ActionHandler) restore(ctx context.Context, snapshotNamespace, snapshot
 		return err
 	}
 
+	if sc.Provisioner == longhorntypes.LonghornDriverName {
+		if err = h.mountSourcePVC(volumeSnapshot); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (h *ActionHandler) mountSourcePVC(volumeSnapshot *snapshotv1.VolumeSnapshot) error {
+	if volumeSnapshot.Spec.Source.PersistentVolumeClaimName == nil {
+		return nil
+	}
+
+	pvc, err := h.pvcCache.Get(volumeSnapshot.Namespace, *volumeSnapshot.Spec.Source.PersistentVolumeClaimName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("can't find pvc %s/%s, err: %w", volumeSnapshot.Namespace, *volumeSnapshot.Spec.Source.PersistentVolumeClaimName, err)
+	}
+
+	volume, err := h.volumeCache.Get(util.LonghornSystemNamespaceName, pvc.Spec.VolumeName)
+	if err != nil {
+		return fmt.Errorf("failed to get volume %s/%s, error: %s", util.LonghornSystemNamespaceName, pvc.Spec.VolumeName, err.Error())
+	}
+
+	if volume.Status.State == lhv1beta1.VolumeStateDetached || volume.Status.State == lhv1beta1.VolumeStateDetaching {
+		volCpy := volume.DeepCopy()
+		volCpy.Spec.NodeID = volume.Status.OwnerID
+		logrus.Infof("mount detached volume %s to the node %s", volCpy.Name, volCpy.Spec.NodeID)
+		if _, err = h.volumes.Update(volCpy); err != nil {
+			return err
+		}
+	}
 	return nil
 }
--- a/pkg/controller/master/volume/controller.go
+++ b/pkg/controller/master/volume/controller.go
@@ -1,0 +1,121 @@
+package volume
+
+import (
+	"fmt"
+	"time"
+
+	lhv1beta1 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+	v1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+
+	ctllonghornv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta1"
+	ctlsnapshotv1 "github.com/harvester/harvester/pkg/generated/controllers/snapshot.storage.k8s.io/v1beta1"
+	"github.com/harvester/harvester/pkg/indexeres"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	detachVolumeEnqueueInterval = 5 * time.Second
+)
+
+type Controller struct {
+	pvcCache         v1.PersistentVolumeClaimCache
+	volumes          ctllonghornv1.VolumeClient
+	volumeController ctllonghornv1.VolumeController
+	volumeCache      ctllonghornv1.VolumeCache
+	snapshotCache    ctlsnapshotv1.VolumeSnapshotCache
+}
+
+// Detach unused volumes, so attached volumes don't block node drain.
+func (c *Controller) DetachVolumesOnChange(_ string, volume *lhv1beta1.Volume) (*lhv1beta1.Volume, error) {
+	if volume == nil || volume.DeletionTimestamp != nil || volume.Status.KubernetesStatus.PVCName == "" {
+		return volume, nil
+	}
+
+	if isVolumeDetached(volume) {
+		return volume, nil
+	}
+
+	pvc, err := c.pvcCache.Get(volume.Status.KubernetesStatus.Namespace, volume.Status.KubernetesStatus.PVCName)
+	if err != nil {
+		return volume, fmt.Errorf("can't find pvc %s/%s, err: %w", volume.Status.KubernetesStatus.Namespace, volume.Status.KubernetesStatus.PVCName, err)
+	}
+	if pvc.Status.Phase == corev1.ClaimPending {
+		c.volumeController.EnqueueAfter(volume.Namespace, volume.Name, detachVolumeEnqueueInterval)
+		return volume, nil
+	}
+
+	canDetach, watchAgain, err := c.checkDetachVolume(pvc)
+	if err != nil {
+		return volume, err
+	}
+	logrus.Debugf("pvc: %s/%s, canDetach: %t, watchAgain: %t", pvc.Namespace, pvc.Name, canDetach, watchAgain)
+	if canDetach {
+		if err := c.detachVolume(pvc); err != nil {
+			return volume, err
+		}
+	}
+	if watchAgain {
+		c.volumeController.EnqueueAfter(volume.Namespace, volume.Name, detachVolumeEnqueueInterval)
+	}
+
+	return volume, nil
+}
+
+func (c *Controller) checkDetachVolume(pvc *corev1.PersistentVolumeClaim) (canDetach, watchAgain bool, err error) {
+	// 1. check whether any pod uses the PVC
+	volume, err := c.volumeCache.Get(util.LonghornSystemNamespaceName, pvc.Spec.VolumeName)
+	if err != nil {
+		return false, false, fmt.Errorf("can't find volume %s/%s, err: %w", util.LonghornSystemNamespaceName, pvc.Spec.VolumeName, err)
+	}
+	for _, workload := range volume.Status.KubernetesStatus.WorkloadsStatus {
+		// For running workload, we don't want to watch again
+		if workload.PodStatus == string(corev1.PodRunning) || workload.PodStatus == string(corev1.PodPending) {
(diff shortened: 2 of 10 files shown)


futuretea (commented):
1. Concurrent problem
operations such as the following may not be safe
1). create vm
2). shutdown vm
3). create backup 1
4). create backup 2 before backup1 ready

2. Other scenarios can also cause this issue
1): take volume snapshot:
https://github.com/harvester/harvester/blob/c8a8494b043646e55bb0e77706cae5e1e83a0ccc/pkg/api/volume/handler.go#L298

---
Some ideas may not be right:

A volume can be dettach safely under the following conditions:
1. No pod uses this volume
2. All volumesnapshots based on this volume are in the ready to use state

We can:

Add an indexer for volume ->volumesnapshots

Add a volumesnapshot controller, when the status of a volumesnapshot becomes ready to use or is deleted. the controller checks whether the volumesnapshot's volume meets the above two conditions. If yes and the volume is attached, dettach the volume.

 Add a volume controller., when the workloadsStatus of a volume is empty (Condition 1 above) and attached, the controller checks whether the volume meets condition 2 above. If yes, dettach the volume.

---
Or can we determine if the volume can be dettach during drain?

bk201:
> Or can we determine if the volume can be dettach during drain?

We can do this before node draining and use it as a workaround.

FrankYang0529:
> A volume can be dettach safely under the following conditions:
> 
> 1. No pod uses this volume
> 2. All volumesnapshots based on this volume are in the ready to use state

This may still have some edge cases. We also need to mount volumes when we restore a VM snapshot. In this case, the PVC is not in the above two conditions. For example:

1. Create a VM.
2. Create a VM snapshot.
3. Stop the VM.
4. Restore the VM snapshot to a new VM.
5. Before the restore is finished, backup the original VM.

If we only need to unmount volumes before draining nodes, I suggest only adding this script when we do an upgrade.

w13915984028:
Question: 

to solve such issue from the root, shall we consider:

(1) check any ongoing backup/snapshot before upgrade

(2) check any ongoing upgrade before taking backup/snapshot

do not allow them run in parallel in system level or given node level

FrankYang0529:
> (1) check any ongoing backup/snapshot before upgrade
> 
> (2) check any ongoing upgrade before taking backup/snapshot

If we want to let users run some workloads or backup / restore during the upgrade, adding these webhooks may limit the cluster's ability.

pkg/controller/master/volume/register.go
```diff
@@ -0,0 +1,32 @@
...
+
+func Register(ctx context.Context, management *config.Management, options config.Options) error {
+	var (
+		pvcCache      = management.CoreFactory.Core().V1().PersistentVolumeClaim().Cache()
+		volumeClient  = management.LonghornFactory.Longhorn().V1beta1().Volume()
+		volumeCache   = volumeClient.Cache()
+		snapshotCache = management.SnapshotFactory.Snapshot().V1beta1().VolumeSnapshot().Cache()
+	)
+
+	// registers the volumeSnapshot controller
```
futuretea: volume controller

pkg/controller/master/volume/volume_controller.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, fmt.Errorf("can't find pvc %s/%s, err: %w", volume.Status.KubernetesStatus.Namespace, volume.Status.KubernetesStatus.PVCName, err)
+	}
+	if pvc.Status.Phase == corev1.ClaimPending {
+		h.volumeController.EnqueueAfter(volume.Namespace, volume.Name, 5*time.Second)
+		return volume, nil
+	}
+
+	canDetach, watchAgain, err := h.checkDetachVolume(pvc)
+	if err != nil {
+		return nil, err
```
futuretea: return volume, err

pkg/controller/master/volume/volume_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	logrus.Debugf("pvc: %s/%s, canDetach: %t, watchAgain: %t", pvc.Namespace, pvc.Name, canDetach, watchAgain)
+	if canDetach {
+		if err := h.detachVolume(pvc); err != nil {
+			return nil, err
```
futuretea: ditto

futuretea (commented):
Thanks. Great job getting this fix done so quickly. It would be nice if the PR included a volume_controller_test.go

pkg/controller/master/volume/register.go
FrankYang0529: Updated it. Thanks.

pkg/controller/master/volume/volume_controller.go
FrankYang0529: Updated it. Thanks.

pkg/controller/master/volume/volume_controller.go
FrankYang0529: Updated it. Thanks.

FrankYang0529:
> It would be nice if the PR included a volume_controller_test.go

I added some test cases for it, but they don't cover all codes because I haven't found a good way to mock `GetByIndex`. Thanks.

pkg/api/volumesnapshot/handler.go
```diff
@@ -124,5 +131,43 @@ func (h *ActionHandler) restore(ctx context.Context, snapshotNamespace, snapshot
...
+	if err != nil {
+		return fmt.Errorf("failed to get volume %s/%s, error: %s", util.LonghornSystemNamespaceName, pvc.Spec.VolumeName, err.Error())
+	}
+
+	volCpy := volume.DeepCopy()
+	if volume.Status.State == lhv1beta1.VolumeStateDetached || volume.Status.State == lhv1beta1.VolumeStateDetaching {
+		volCpy.Spec.NodeID = volume.Status.OwnerID
+	}
+
+	if !reflect.DeepEqual(volCpy, volume) {
```
guangbochen: prefer to check the `spec` or the specific field if this is the only thing that we need to compare.

pkg/controller/master/volume/controller.go
```diff
@@ -0,0 +1,119 @@
...
+
+type Controller struct {
+	pvcCache         v1.PersistentVolumeClaimCache
+	volumes          ctllonghornv1.VolumeClient
+	volumeController ctllonghornv1.VolumeController
+	volumeCache      ctllonghornv1.VolumeCache
+	snapshotCache    ctlsnapshotv1.VolumeSnapshotCache
+}
+
+func (h *Controller) DetachVolumesOnChange(_ string, volume *lhv1beta1.Volume) (*lhv1beta1.Volume, error) {
```
guangbochen: - Use `c` for controller abbreviation instead.
- Could add some additional comments explaining the controller's purpose and how it fits into the overall management.
- Simplify the logic in the `DetachVolumesOnChange`: this can be simplified by checking if the volume is attached(e.g., pvStatus is bound) before proceeding with the rest of the checks and updates.

pkg/controller/master/volume/register.go
```diff
@@ -0,0 +1,32 @@
+package volume
+
+import (
+	"context"
+
+	"github.com/harvester/harvester/pkg/config"
+)
+
+const (
+	volumeControllerDetachVolume = "VolumeController.DetachVolume"
```
guangbochen: let's try out best to keep the controller name more consistent by using `-` to join the words.

pkg/controller/master/volume/controller.go
```diff
@@ -0,0 +1,119 @@
...
+	volCpy := volume.DeepCopy()
+	if volume.Status.State == lhv1beta1.VolumeStateAttached || volume.Status.State == lhv1beta1.VolumeStateAttaching {
+		volCpy.Spec.NodeID = ""
+	}
+	if !reflect.DeepEqual(volCpy, volume) {
+		logrus.Infof("detach volume %s", volCpy.Name)
+		if _, err = h.volumes.Update(volCpy); err != nil {
+			return err
+		}
+	}
```
guangbochen: Move the update function to the above condition check bracket so we can skip the deepEuqal method.

pkg/controller/master/volume/controller.go
```diff
@@ -0,0 +1,119 @@
...
+	volumeCache      ctllonghornv1.VolumeCache
+	snapshotCache    ctlsnapshotv1.VolumeSnapshotCache
+}
+
+func (h *Controller) DetachVolumesOnChange(_ string, volume *lhv1beta1.Volume) (*lhv1beta1.Volume, error) {
+	if volume == nil || volume.DeletionTimestamp != nil {
+		return volume, nil
+	}
+
+	if volume.Status.State == lhv1beta1.VolumeStateDetached || volume.Status.State == lhv1beta1.VolumeStateDetaching {
```
guangbochen: It would be helpful to extract those condition check into separate functions with descriptive names, such as `isVolumeDetached`. This would make the code easier to read and understand.

pkg/controller/master/volume/controller.go
```diff
@@ -0,0 +1,119 @@
...
+		return volume, err
+	}
+	logrus.Debugf("pvc: %s/%s, canDetach: %t, watchAgain: %t", pvc.Namespace, pvc.Name, canDetach, watchAgain)
+	if canDetach {
+		if err := h.detachVolume(pvc); err != nil {
+			return volume, err
+		}
+	}
+	if watchAgain {
+		h.volumeController.EnqueueAfter(volume.Namespace, volume.Name, 5*time.Second)
```
guangbochen: ditto

futuretea (approved):
LGTM, Thanks! There is no problem in testing according to the test plan

pkg/controller/master/volume/controller.go
```diff
@@ -0,0 +1,119 @@
...
+	if volume.Status.State == lhv1beta1.VolumeStateDetached || volume.Status.State == lhv1beta1.VolumeStateDetaching {
+		return volume, nil
+	}
+
+	pvc, err := h.pvcCache.Get(volume.Status.KubernetesStatus.Namespace, volume.Status.KubernetesStatus.PVCName)
+	if err != nil {
+		return volume, fmt.Errorf("can't find pvc %s/%s, err: %w", volume.Status.KubernetesStatus.Namespace, volume.Status.KubernetesStatus.PVCName, err)
+	}
+	if pvc.Status.Phase == corev1.ClaimPending {
+		h.volumeController.EnqueueAfter(volume.Namespace, volume.Name, 5*time.Second)
```
guangbochen: Don't use the magic number 5 seconds here; define it with a meaningful constant name at the top.

pkg/controller/master/volume/controller.go
FrankYang0529: Added `defaultEnqueueInterval`. Thanks.

pkg/controller/master/volume/controller.go
FrankYang0529: Added `defaultEnqueueInterval`. Thanks.

pkg/controller/master/volume/controller.go
FrankYang0529: Added `isVolumeDetached` to util.go. Thanks.

pkg/controller/master/volume/controller.go
FrankYang0529: Updated it. Thanks.

pkg/controller/master/volume/register.go
FrankYang0529: Updated it to `detach-volume-controller`. Thanks.

pkg/api/volumesnapshot/handler.go
FrankYang0529: Moved the update function to the above condition check. Thanks.

pkg/controller/master/volume/controller.go
FrankYang0529: * Updated to use `c`.
* Added some comments.
* Yes, that is why we check `isVolumeDetached` and whether PVC is pending.

pkg/api/volumesnapshot/handler.go
```diff
@@ -124,5 +130,40 @@ func (h *ActionHandler) restore(ctx context.Context, snapshotNamespace, snapshot
...
+		}
+		return fmt.Errorf("can't find pvc %s/%s, err: %w", volumeSnapshot.Namespace, *volumeSnapshot.Spec.Source.PersistentVolumeClaimName, err)
+	}
+
+	volume, err := h.volumeCache.Get(util.LonghornSystemNamespaceName, pvc.Spec.VolumeName)
+	if err != nil {
+		return fmt.Errorf("failed to get volume %s/%s, error: %s", util.LonghornSystemNamespaceName, pvc.Spec.VolumeName, err.Error())
+	}
+
+	volCpy := volume.DeepCopy()
```
guangbochen: Moving this to the following bracket if it is not used outside.

pkg/controller/master/volume/controller.go
```diff
@@ -0,0 +1,121 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	ctllonghornv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta1"
+	ctlsnapshotv1 "github.com/harvester/harvester/pkg/generated/controllers/snapshot.storage.k8s.io/v1beta1"
+	"github.com/harvester/harvester/pkg/indexeres"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	defaultEnqueueInterval = 5 * time.Second
```
guangbochen: \s\defaultEnqueueInterval\detachVolumeEnqueueInterval

pkg/controller/master/volume/controller.go
```diff
@@ -0,0 +1,121 @@
...
+	return true, false, nil
+}
+
+func (c *Controller) detachVolume(pvc *corev1.PersistentVolumeClaim) error {
+	volume, err := c.volumeCache.Get(util.LonghornSystemNamespaceName, pvc.Spec.VolumeName)
+	if err != nil {
+		return fmt.Errorf("can't find volume %s/%s, err: %w", util.LonghornSystemNamespaceName, pvc.Spec.VolumeName, err)
+	}
+
+	volCpy := volume.DeepCopy()
```
guangbochen: ditto

guangbochen (approved):
just a small nit, everything else is LGTM, thanks.

pkg/api/volumesnapshot/handler.go
FrankYang0529: Moved it. Thanks.

pkg/controller/master/volume/controller.go
FrankYang0529: Moved it. Thanks.

pkg/controller/master/volume/controller.go
FrankYang0529: Updated it. Thanks.

bk201 (approved):
LGTM. Test upgrade with
- Shutdown VMs and create backup.
- Create volumes and snapshot.
The volumes are detached correctly after upgrading the harvester chart.

bk201:
@mergifyio backport v1.1

Issue #3671, opened 2023-03-17
[ENHANCEMENT] Increase Replica Replenishment Wait Interval during upgrade


bk201:
**Is your enhancement related to a problem? Please describe.**

The current Longhorn ["Replica Replenishment Wait Interval"](https://longhorn.io/docs/1.3.2/references/settings/#replica-replenishment-wait-interval) setting is just 10 minutes. If a replica is offline and online after 10 minutes, then the replica is discarded without being reused. Longhorn will find a new place to rebuild a new replica, which not only increases cluster I/O but the pre-drain phase time of the next upgrading node.

When upgrading and rebooting a node, the offline of a replica starts after the draining phase. The rest operations are:
- Upgrade RKE2
- Upgrade OS
- Reboot: modern machines need minutes to reboot, which generally means it's not easy to finish the whole process in 10 minutes.
- Node is back again.

**Describe the solution you'd like**

Increase the value to a reasonable value like 30 minutes and restore it to what the user sets after the upgrade is done.

**Describe alternatives you've considered**

Advise user to set the value before an upgrade.

**Additional context**

innobead:
["Replica Replenishment Wait Interval"](https://longhorn.io/docs/1.3.2/references/settings/#replica-replenishment-wait-interval) can be used for intended node/cluster maintenance, so increasing it seems reasonable, but because this is a global setting, it would delay other volume replica rebuilding on other running nodes.

starbops:
The plan is to increase the `replica-replenishment-wait-interval` LH setting from the default 600 to 1800 seconds at the beginning of each upgrade, and the setting will be configured back to the original value after the upgrade ends. The original value will be stored in the upgrade CR's specific annotation instead of creating a new field of the upgrade CRD.

w13915984028:
@innobead Could LH also enhances to support volume level  "[Replica Replenishment Wait Interval"](https://longhorn.io/docs/1.3.2/references/settings/#replica-replenishment-wait-interval) setting ? thus Harvester can set the upgrade related volume with specific long wait value without touch the cluster level setting back and forth. thanks.

TachunLin:
Verified **fixed** on `master-f27ede00-head` (3/23). Close this issue.

### Result
During upgrade process enter into node upgrade stage
1. We can find `longhorn.io/replica-replenishment-wait-interval` annotation set in the upgrade CR
  ```
  node1:~ # kubectl -n harvester-system get upgrades hvst-upgrade-9j6px -o yaml
[...]

PR #3679, opened 2023-03-20
Make LH reuse data

starbops:

**Problem:**

**Solution:**

The upgrade controller will extend the `replica-replenishment-wait-interval` setting to 30 mins and will revert back to the original value after the upgrade ends. By extending the value to 30 mins, the probability of reusing data of a failed replica is increased in the node upgrade scenario, thus improving the entire upgrade speed.

**Related Issue:**

#3671 

**Test plan:**

1. Provision a multi-node Harvester cluster using the ISO image containing the PR
2. Upgrade the cluster with the same ISO image
3. There should be a `longhorn.io/replica-replenishment-wait-interval` annotation set in the upgrade CR after the upgrade progress is entering NodeUpgrade
4. Check the Longhorn setting `replica-replenishment-wait-interval` is changed to 1800 (30 mins)
5. Wait for the upgrade ends
6. After the upgrade ends, check the Longhorn setting `replica-replenishment-wait-interval`, it should be reverted back to the original value (normally 600, the default value)

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strconv"
 	"sync"
 	"time"
 
@@ -25,6 +26,7 @@ import (
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	kubevirtctrl "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	ctllonghornv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta1"
 	upgradectlv1 "github.com/harvester/harvester/pkg/generated/controllers/upgrade.cattle.io/v1"
 	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
@@ -57,6 +59,10 @@ const (
 	rke2PostDrainAnnotation = "rke.cattle.io/post-drain"
 
 	upgradeComponentRepo = "repo"
+
+	replicaReplenishmentWaitIntervalSetting  = "replica-replenishment-wait-interval"
+	replicaReplenishmentAnnotation           = "harvesterhci.io/" + replicaReplenishmentWaitIntervalSetting
+	extendedReplicaReplenishmentWaitInterval = 1800
 )
 
 // upgradeHandler Creates Plan CRDs to trigger upgrades
@@ -85,6 +91,9 @@ type upgradeHandler struct {
 	clusterClient    provisioningctrl.ClusterClient
 	clusterCache     provisioningctrl.ClusterCache
 	deploymentClient ctlappsv1.DeploymentClient
+
+	lhSettingClient ctllonghornv1.SettingClient
+	lhSettingCache  ctllonghornv1.SettingCache
 }
 
 func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*harvesterv1.Upgrade, error) {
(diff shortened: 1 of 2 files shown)


pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -272,6 +281,14 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
 				return h.upgradeClient.Update(toUpdate)
 			}
 		} else {
+			// save the original value of replica-replenishment-wait-interval setting and extend it with a longer value
```
bk201, line 284: @starbops Please help add a check: If the value is already more than 30 minutes (e.g., the user set it), let's not change it.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -543,3 +567,43 @@ func isVersionUpgradable(currentVersion, minUpgradableVersion string) error {
...
+	str, ok := upgrade.Annotations[replicaReplenishmentAnnotation]
+	if !ok {
+		return fmt.Errorf("no original replica-replenishment-wait-interval value set")
+	}
+	value, err := strconv.Atoi(str)
+	if err != nil {
+		return err
+	}
+	if err := h.setReplicaReplenishmentValue(value); err != nil {
+		return err
```
bk201: The codefactor suggestion is good.

w13915984028 (commented):
seems few improvements are needed:

(1)As @bk201 said, check the value if it has already been > 30 minutes

(2)Try restore value only when label/annotation is there, and the value is different than current value, skip blank label/annotation instead of error out, skip same value

(3) Try also in the `cleanup()` branch, suppose users delete the `upgrade` ?

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Updated, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Updated, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -57,6 +59,10 @@ const (
 	rke2PostDrainAnnotation = "rke.cattle.io/post-drain"
 
 	upgradeComponentRepo = "repo"
+
+	replicaReplenishmentWaitIntervalSetting  = "replica-replenishment-wait-interval"
+	replicaReplenishmentAnnotation           = "longhorn.io/" + replicaReplenishmentWaitIntervalSetting
```
guangbochen: Prefer to use the `harvesterhci.io` prefix if this annotation will not take any effects on the LH side.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -375,6 +399,13 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
 		}
 	}
 
+	// restore Longhorn replica-replenishment-wait-interval setting (multi-node cluster only)
```
guangbochen, line 402: why only applies to the multi-node cluster?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -272,6 +281,21 @@ func (h *upgradeHandler) OnChanged(key string, upgrade *harvesterv1.Upgrade) (*h
...
+			// save the original value of replica-replenishment-wait-interval setting and extend it with a longer value
+			// skip if the value is already larger than extendedReplicaReplenishmentWaitInterval
+			replicaReplenishmentWaitIntervalValue, err := h.getReplicaReplenishmentValue()
+			if err != nil {
+				return nil, err
+			}
+			if replicaReplenishmentWaitIntervalValue < extendedReplicaReplenishmentWaitInterval {
+				if err := h.saveReplicaReplenishmentToUpgradeAnnotation(toUpdate); err != nil {
+					return nil, err
+				}
```
guangbochen, line 293: Can we store the annotation in the LH setting instead? so we don't need to update two separated objects?

pkg/controller/master/upgrade/upgrade_controller.go
bk201: There is no need for a single node cluster so far. The replica is certainly re-used.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: The upgrade CR here is going to be updated with or without this function call, so there is no additional update overhead imposed here. We're just hitchhiking the original workflow.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Updated, thanks

bk201 (approved):
LGTM, thanks!

w13915984028 (commented):
The last question is:

If upgrade fails after the LH value is changed but not reverted, and the upgrade object is deleted by user manually, will the LH value be restored automatically?

starbops:
> The last question is:
> 
> If upgrade fails after the LH value is changed but not reverted, and the upgrade object is deleted by user manually, will the LH value be restored automatically?

Yes, we have addressed this in the `cleanup()` call when upgrade CR is going to be removed:

https://github.com/harvester/harvester/pull/3679/files#diff-ad1688890111ebb67c1e9830de55577e7b83f1828d79f6f1f9609122cac236d3R402-R408

w13915984028 (approved):
LGTM, thanks.

guangbochen (approved):
LGTM, thanks.

bk201:
@mergifyio backport v1.1

Issue #2935, opened 2022-10-13
[Enhancement] Add disk space checking before upgrading


noahgildersleeve:
**Is your feature request related to a problem? Please describe.**

Add a check for cluster resource before completing an upgrade.

**Describe the solution you'd like**

Add a validation webhook and reject creating an upgrade if:

- Node space is not enough. We need ~25GB of free space to preload images.
- ~CPU/memory is not enough.~ (move to another ticket, this is used to trace disk space only)
  - We need CPU allocation for additional pods (like duplicate longhorn instance-manager-e/r).
  - New resources that are needed by new components.
  - For live-migration cases, we need to check if the rest nodes can absorb VMs from a upgrading node.

**Describe alternatives you've considered**

**Additional context**

This is breaking out some of the functionality from a previous ticket [here](https://github.com/harvester/harvester/issues/2833)

- We have seen weird behavior when longhorn cannot have enough CPU to spawn new pods.

Martin-Weiss:
Adding check is great - thanks!! But - just to be sure to be sure - please add the documentation what to do in case a check fails.. ;-)

bk201:
@noahgildersleeve I've added more resource checking.

bk201:
Consolidate into https://github.com/harvester/harvester/issues/3067 and close this one.

bk201:
Reopen this for disk space checking. Add to [epic](https://github.com/harvester/harvester/issues/2935#issuecomment-1471182660).

bk201:
One way is to use node exporter metric in Prometheus:

Query: `node_filesystem_avail_bytes{mountpoint="/usr/local"}`
```
node_filesystem_avail_bytes{container="node-exporter", device="/dev/vda5", endpoint="metrics", fstype="ext4", instance="10.10.0.11:9796", job="node-exporter", mountpoint="/usr/local", namespace="cattle-monitoring-system", pod="rancher-monitoring-prometheus-node-exporter-gz4qd", service="rancher-monitoring-prometheus-node-exporter"} | 36105703424
node_filesystem_avail_bytes{container="node-exporter", device="/dev/vda5", endpoint="metrics", fstype="ext4", instance="10.10.0.12:9796", job="node-exporter", mountpoint="/usr/local", namespace="cattle-monitoring-system", pod="rancher-monitoring-prometheus-node-exporter-p4t4f", service="rancher-monitoring-prometheus-node-exporter"} | 38456356864
[...]

PR #3700, opened 2023-03-22
Check disk space before upgrade

FrankYang0529:
**Problem:**
If any node doesn't have enough space, we can't load images during the upgrade.

**Solution:**
Add a webhook to check it

**Related Issue:**
https://github.com/harvester/harvester/issues/2935

**Test plan:**

### Setup

1. Create a new Harvester cluster. One of the nodes' disk space should be 200G.
2. If every node's `/usr/local` free space is more than 30G, you can use `dd` to write files to make sure one of the nodes is less than 30G. For example: `dd if=/dev/zero of=/usr/local/test.img bs=1G count=5`.
3. Create a test version.
```yaml
kind: Version
metadata:
  name: 1.1.2-test1
  namespace: harvester-system
spec:
  isoURL: http://192.168.0.181:8000/harvester-master-amd64.iso
  minUpgradableVersion: 1.1.1
  releaseDate: "20211231"
  tags:
  - dev
  - test
```

Screenshot 2023-03-22 at 4 55 00 PM

### Case 2: using `harvesterhci.io/minFreeDiskSpaceGB` annotation to change the minimal required disk space

1. Check the minimal free disk space of nodes.
2. Use a number lower than the value above to update the test version. Assume it's `25`.
```yaml
kind: Version
metadata:
  annotations:
    harvesterhci.io/minFreeDiskSpaceGB: "25"
  name: 1.1.2-test1
  namespace: harvester-system
spec:
  isoURL: http://192.168.0.181:8000/harvester-master-amd64.iso
  minUpgradableVersion: 1.1.1
  releaseDate: "20211231"
  tags:
  - dev
  - test
```
3. Click the upgrade again. We should not get any error messages.

### Case 3: invalid value for `harvesterhci.io/minFreeDiskSpaceGB` annotation can't create any version

1. Use invalid values like `-1`, `one G` for `harvesterhci.io/minFreeDiskSpaceGB` annotation.
2. We can't create a version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -110,6 +110,7 @@ require (
 	k8s.io/gengo v0.0.0-20211129171323-c02415ce4185
 	k8s.io/kube-openapi v0.0.0-20220803162953-67bda5d908f1
 	k8s.io/kubectl v0.24.2
+	k8s.io/kubelet v0.24.10
 	k8s.io/utils v0.0.0-20230406110748-d93618cff8a2
 	kubevirt.io/api v0.54.0
 	kubevirt.io/containerized-data-importer-api v1.47.0
--- a/pkg/webhook/resources/upgrade/validator.go
+++ b/pkg/webhook/resources/upgrade/validator.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/docker/go-units"
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	fleetv1alpha1 "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
@@ -279,7 +280,9 @@ func (v *upgradeValidator) checkDiskSpace(node *corev1.Node, minFreeDiskSpace ui
 		return werror.NewInternalError(fmt.Sprintf("can't get node %s available bytes from %s, err: %+v", node.Name, kubeletURL, err))
 	}
 	if *summary.Node.Fs.AvailableBytes < minFreeDiskSpace {
-		return werror.NewBadRequest(fmt.Sprintf("minimal free disk space for upgrade is %d, node %s only has %d", minFreeDiskSpace, node.Name, *summary.Node.Fs.AvailableBytes))
+		min := units.BytesSize(float64(minFreeDiskSpace))
+		avail := units.BytesSize(float64(*summary.Node.Fs.AvailableBytes))
+		return werror.NewBadRequest(fmt.Sprintf("Node %q has insufficient free space (%s). An upgrade requires at least %s of free space.", node.Name, avail, min))
 	}
 	return nil
 }
--- a/pkg/webhook/resources/upgrade/validator.go
+++ b/pkg/webhook/resources/upgrade/validator.go
@@ -1,7 +1,11 @@
 package upgrade
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strconv"
 	"strings"
 
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
@@ -13,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
+	kubeletv1 "k8s.io/kubelet/pkg/apis/stats/v1alpha1"
 	clusterv1alpha4 "sigs.k8s.io/cluster-api/api/v1alpha4"
 
 	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
@@ -22,13 +27,16 @@ import (
 	ctllhv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
 	"github.com/harvester/harvester/pkg/util"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
+	versionWebhook "github.com/harvester/harvester/pkg/webhook/resources/version"
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
 const (
-	upgradeStateLabel     = "harvesterhci.io/upgradeState"
-	skipWebhookAnnotation = "harvesterhci.io/skipWebhook"
-	managedChartNamespace = "fleet-local"
+	upgradeStateLabel              = "harvesterhci.io/upgradeState"
+	skipWebhookAnnotation          = "harvesterhci.io/skipWebhook"
+	rkeInternalIPAnnotation        = "rke2.io/internal-ip"
+	managedChartNamespace          = "fleet-local"
+	defaultMinFreeDiskSpace uint64 = 30 * 1024 * 1024 * 1024 // 30GB
 )
 
 func NewValidator(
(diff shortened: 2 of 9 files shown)


pkg/webhook/resources/upgrade/validator.go
```diff
@@ -197,6 +209,44 @@ func (v *upgradeValidator) checkNodes() error {
...
+
+		var summary kubeletv1.Summary
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return werror.NewInternalError(fmt.Sprintf("can't read response from %s, err: %+v", kubeletURL, err))
+		}
+		if err = json.Unmarshal(body, &summary); err != nil {
+			return werror.NewInternalError(fmt.Sprintf("can't parse json response %s, err: %+v", string(body), err))
+		}
+		if summary.Node.Fs.AvailableBytes == nil {
```
w13915984028: please double check
https://github.com/kubernetes/kubernetes/issues/68522
and also
https://github.com/elastic/beats/issues/12792

w13915984028:
meanwhile, 

we are looking from data in this path  `node_filesystem_avail_bytes{mountpoint="/usr/local"}`

does `kubelet` expose the value for this fs ?  more spaces in other mountpoint may not guarantee the upgrade usage.

pkg/webhook/resources/upgrade/validator.go
FrankYang0529: I still can find the `/stats/summary` endpoint in the latest code. We can probably use this to get disk space first and keep an eye on k8s issue. Thanks.

https://github.com/kubernetes/kubernetes/blob/d2be69ac11346d2a0fab8c3c168c4255db99c56f/pkg/kubelet/server/stats/handler.go#L141

FrankYang0529:
> does `kubelet` expose the value for this fs ?

Yes, if you use `findmnt /usr/local`, it's mounted to `/dev/vda5`. If you use `findmnt /dev/vda5`, you can find `/var/lib/kubelet` is also mounted to this device. I tried to write a 5G file under `/usr/local`, and kubelet `/stats/summary` endpoint also updates `availableBytes`.

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -26,9 +30,11 @@ import (
...
 
 const (
-	upgradeStateLabel     = "harvesterhci.io/upgradeState"
-	skipWebhookAnnotation = "harvesterhci.io/skipWebhook"
-	managedChartNamespace = "fleet-local"
+	upgradeStateLabel       = "harvesterhci.io/upgradeState"
+	skipWebhookAnnotation   = "harvesterhci.io/skipWebhook"
+	rkeInternalIPAnnotation = "rke2.io/internal-ip"
+	managedChartNamespace   = "fleet-local"
+	minFreeDiskSpace        = 30 * 1024 * 1024 * 1024 // 30GB
```
w13915984028: `30 GB` is an estimated value, it may not always be fit in different releases, or for all user environments, e.g. from v1.2.0 to v1.2.1 it needs min 35 GB; some user may also want to change it to `20GB` in v1.1.2->v1.2.0 upgrade as his NODE does not have that much of space.

more flexible solution will be:

let it be a Harvester setting, or read certain `label/annotation` from `Upgrade` to overwrite this value

the Harvester controller/webhook POD will not be upgraded before an upgrade, those hardcoded value may be a barrier, it needs a way to load those requirements dynamically.

cc @bk201

FrankYang0529: Yes, it's not good to hard code here. I think it's good to add an annotation like `harvesterhci.io/minFreeDiskSpace` when we release [version.yaml](https://releases.rancher.com/harvester/v1.1.1/version.yaml). If we can't find the annotation, we use the default value (30 GB).

pkg/webhook/resources/upgrade/validator.go
bk201: it's quite a big chunk, how about let's move the kubelet check to a function?

w13915984028 (approved):
LGTM, thanks.

go.mod
```diff
@@ -110,6 +110,7 @@ require (
 	k8s.io/gengo v0.0.0-20211129171323-c02415ce4185
 	k8s.io/kube-openapi v0.0.0-20220803162953-67bda5d908f1
 	k8s.io/kubectl v0.24.2
+	k8s.io/kubelet v0.0.0
```
guangbochen: Can we specify the required version, e.g., v0.2.4.10?

pkg/webhook/util/http_client.go
```diff
@@ -0,0 +1,35 @@
...
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/rest"
+)
+
+func GetHTTPTransportWithCertificates(config *rest.Config) (*http.Transport, error) {
+	logrus.Infof("config: %+v", config)
```
guangbochen: Maybe is better to use debug level log here?

pkg/webhook/util/http_client.go
```diff
@@ -0,0 +1,35 @@
...
+	caCert, err := ioutil.ReadFile(config.TLSClientConfig.CAFile)
```
guangbochen, line 16: Check if TLSClientConfig field is set first.

pkg/webhook/resources/version/validator.go
```diff
@@ -0,0 +1,59 @@
...
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	werror "github.com/harvester/harvester/pkg/webhook/error"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+const (
+	MinFreeDiskSpaceGBAnnotation = "harvesterhci.io/minFreeDiskSpaceGB"
```
guangbochen, line 16: Use lower case annotation keys for better project consistency.

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -197,8 +234,53 @@ func (v *upgradeValidator) checkNodes() error {
...
+
+	resp, err := v.httpClient.Do(req)
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("can't make http request to get %s, err: %+v", kubeletURL, err))
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return werror.NewInternalError(fmt.Sprintf("http response from %s is %d", kubeletURL, resp.StatusCode))
+	}
```
guangbochen: Add `node.Name` in all related error logs; otherwise, it would be hard for users to find out which node has contained the error.

go.mod
FrankYang0529: Yes, update it. Thanks.

pkg/webhook/util/http_client.go
FrankYang0529: Updated it. Thanks.

pkg/webhook/util/http_client.go
FrankYang0529: TLSClientConfig is not a pointer, so we can't check whether it's nil first.

pkg/webhook/resources/version/validator.go
FrankYang0529: Another file `pkg/webhook/resources/upgrade/validator.go` also uses this variable, so I can't use lower case here.

pkg/webhook/resources/upgrade/validator.go
FrankYang0529: Updated it. Thanks.

bk201:
Slightly adjust the error message to:
Screenshot 2023-05-25 at 4 15 28 PM

bk201 (approved):

starbops (approved):
Tested with the default minFreeDiskSpace and the customized one. LGTM, thanks!

guangbochen (approved):
LGTM, thanks.

w13915984028:
add https://github.com/harvester/harvester/issues/4020, as we encounter the error of `ctr adding image fail`, the root cause is `/usr/local` does not have enough space

bk201:
@mergifyio backport v1.1

Issue #3660, opened 2023-03-15
[ENHANCEMENT] add webhook to validate addons


w13915984028:
**Is your enhancement related to a problem? Please describe.**

<del>
Logging, monitoring are converted to addons, and the values are in a string field valuesContent, it should be at least valid JSON format.
</del>

Logging, monitoring are converted to addons, webhook is needed to check certain cases

**Describe the solution you'd like**

<del>
Use webhook to validate the valuesContent  in json format
</del>

With the wide usage of harvester addon, webook is required for following scenarios.
(1) cannot create more than one instance for one addon
(2) cannot change addon field chart

We did not find feasible solution for below 2 checks, leave them blank now.

(3) check the existing of given chart (chart + repo + version) (TBD)
(4) check the valuesContent is an valid yaml/json format (TBD)

**Describe alternatives you've considered**

**Additional context**

related issue:
https://github.com/harvester/harvester/issues/2992
UI enhancement:
https://github.com/harvester/harvester/issues/3235

Epic:
https://github.com/harvester/harvester/issues/3262

**Test plan:**

please refer `test plan` description in PR:  https://github.com/harvester/harvester/pull/3704

irishgordo:
@w13915984028  I'm noticing that it seems the validation fails on v1.2.0-rc1 as user is not suppose to be able to edit the chart?

**validation failed**

To reproduce:
- create v1.2.0-rc1 cluster
- once cluster is up and running, do not manually enable addons
- download kubeconfig
- `kubectl edit addons.harvesterhci.io -n harvester-system pcidevices-controller`
- edit spec.chart
- save

You can successfully edit spec.chart.
In the attached example, an extra "r" was added to become: `harvester-pcidevices-controllerr`
![Screenshot from 2023-06-08 11-33-22](https://github.com/harvester/harvester/assets/5370752/c6addecf-ddae-4bc2-b1ee-39e1ae573d83)

w13915984028:
@irishgordo Looks the PR is not in RC1, please validate it with master-head.

https://github.com/harvester/harvester/commits/v1.2
![image](https://github.com/harvester/harvester/assets/31133476/f67ab797-b7a0-4338-8488-13140745594e)

https://github.com/harvester/harvester/releases/tag/v1.2.0-rc1

irishgordo:
Ahh thanks @w13915984028  - I'll re-validate with master-head release :smile:

irishgordo:
[...]

PR #3704, opened 2023-03-22
add validator webhook for harvester addon

w13915984028:

**Problem:**

With the wide usage of `harvester addon`, webook is required for following scenarios.
(1) cannot create more than one instance for one addon
(2) cannot change addon field `chart`

We did not find feasible solution for below 2 checks, leave them blank now.
<del>
(3) check the existing of given chart (chart + repo + version)  (TBD)
(4) check the valuesContent is an valid yaml/json format (TBD)
</del>

**Solution:**

Add webhook to validate addon.

**Related Issue:**
https://github.com/harvester/harvester/issues/3660

**Test plan:**

(1) cannot create more than one instance for one addon
```
harv31:/home/rancher # kk create -f new_addon.yam 

Error from server (Conflict): error when creating "new_addon.yam": admission webhook "validator.harvesterhci.io" denied the request: addon with Chart "harvester-vm-import-controller" has been created, cannot create a new one
```

(2) cannot change addon field `chart`
```
kk edit addons.harvesterhci.io -n harvester-system pcidevices-controller

error: addons.harvesterhci.io "pcidevices-controller" could not be patched: admission webhook "validator.harvesterhci.io" denied the request: chart field cannot be changed.
You can run `kubectl replace -f /tmp/kubectl-edit-2576031672.yaml` to try this update again.
```

### DIFF
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -1,0 +1,79 @@
+package addon
+
+import (
+	"fmt"
+
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	werror "github.com/harvester/harvester/pkg/webhook/error"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+func NewValidator(addons ctlharvesterv1.AddonCache) types.Validator {
+	return &addonValidator{
+		addons: addons,
+	}
+}
+
+type addonValidator struct {
+	types.DefaultValidator
+
+	addons ctlharvesterv1.AddonCache
+}
+
+func (v *addonValidator) Resource() types.Resource {
+	return types.Resource{
+		Names:      []string{v1beta1.AddonResourceName},
+		Scope:      admissionregv1.NamespacedScope,
+		APIGroup:   v1beta1.SchemeGroupVersion.Group,
+		APIVersion: v1beta1.SchemeGroupVersion.Version,
+		ObjectType: &v1beta1.Addon{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+			admissionregv1.Update,
+		},
+	}
+}
+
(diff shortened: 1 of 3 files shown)


w13915984028:
cc @ibrokethecloud @WuJun2016 @DaiYuzeng 

when more checks are needed, please add here, thanks.

ibrokethecloud:
@w13915984028 LGTM. can we please add some basic tests for this and we can keep building on top of these as we add more validations.

pkg/webhook/resources/addon/validator.go
```diff
@@ -0,0 +1,95 @@
...
+		APIVersion: v1beta1.SchemeGroupVersion.Version,
+		ObjectType: &v1beta1.Addon{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+			admissionregv1.Update,
+		},
+	}
+}
+
+func (v *addonValidator) Create(request *types.Request, newObj runtime.Object) error {
```
guangbochen, line 44: Let's add a basic description of what this validation is about.

pkg/webhook/resources/addon/validator.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	if err := v.checkValuesContent(newAddon); err != nil {
+		return err
+	}
+
+	return v.checkChart(newAddon)
+}
+
+func (v *addonValidator) Update(request *types.Request, oldObj runtime.Object, newObj runtime.Object) error {
```
guangbochen, line 56: ditto

w13915984028:
> @w13915984028 LGTM. can we please add some basic tests for this and we can keep building on top of these as we add more validations.

@ibrokethecloud please take a look of the tests, thanks.

BTW: for `deleting an addon`, should user `disable` them at first ?

ibrokethecloud:
deleting the addon, should trigger gc for the helm chart, which should be same as disabling the addon.

ibrokethecloud (approved):
LGTM

pkg/webhook/resources/addon/validator.go
w13915984028: It is simply following with other validators, most of them have no comments on code.

the description of validator is PR itself

w13915984028: updated.

pkg/webhook/resources/addon/validator.go
w13915984028: updated.

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
lgtm. @w13915984028 Do we need to backport this to v1.1.3?

w13915984028:
@bk201 This commit is more an enhancement, it would be `optional` for v1.1.3, depends on the planned effort to this release. thanks.

Issue #3705, opened 2023-03-23
[BUG] Log packager skips processing all the remaining log files


starbops:
**Describe the bug**

We've spotted that if there is log truncating happening during an upgrade (higher chance in a single node upgrade), the log packager will not only skip processing the impacted file but all the files after that file in lexical order.

Observed in log packager's log:

```
start to package upgrade logs
+ echo 'start to package upgrade logs'
++ mktemp -d
+ tmpdir=/tmp/tmp.cwOGu5UzxU
+ mkdir /tmp/tmp.cwOGu5UzxU/logs
+ cd /archive/logs
+ for f in *.log
+ awk '{$1=$2=""; print $0}' harvester-system.harvester-86bfd6897c-gjqtr.apiserver.20230323.log
+ jq -r .message
parse error: Invalid numeric literal at line 21, column 0
+ ret=4
+ '[' -n 4 ']'
+ echo 'Failed to process the file harvester-system.harvester-86bfd6897c-gjqtr.apiserver.20230323.log with ret=4'
+ echo 'Copy the original file to the destination'
+ cp harvester-system.harvester-86bfd6897c-gjqtr.apiserver.20230323.log /tmp/tmp.cwOGu5UzxU/logs/harvester-system.harvester-86bfd6897c-gjqtr.apiserver.20230323.log
Failed to process the file harvester-system.harvester-86bfd6897c-gjqtr.apiserver.20230323.log with ret=4
Copy the original file to the destination
+ for f in *.log
+ jq -r .message
+ awk '{$1=$2=""; print $0}' harvester-system.hvst-upgrade-bls5n-upgradelog-infra-fluentbit-hgtqv.fluent-bit.20230323.log
+ '[' -n 4 ']'
+ echo 'Failed to process the file harvester-system.hvst-upgrade-bls5n-upgradelog-infra-fluentbit-hgtqv.fluent-bit.20230323.log with ret=4'
+ echo 'Copy the original file to the destination'
+ cp harvester-system.hvst-upgrade-bls5n-upgradelog-infra-fluentbit-hgtqv.fluent-bit.20230323.log /tmp/tmp.cwOGu5UzxU/logs/harvester-system.hvst-upgrade-bls5n-upgradelog-infra-fluentbit-hgtqv.fluent-bit.20230323.log
Failed to process the file harvester-system.hvst-upgrade-bls5n-upgradelog-infra-fluentbit-hgtqv.fluent-bit.20230323.log with ret=4
[...]

PR #3707, opened 2023-03-23
fix(upgradelog): unset log-processing return value

starbops:

**Problem:**

When there's an error happened to process the log lines, every log file after the problematic log file (lexical order) will be mistakenly skipped.

**Solution:**

Reset the return value variable at the end of each run.

**Related Issue:**

#3705

**Test plan:**

1. Provision a single-node Harvester cluster with an ISO image including the fix
2. Trigger an upgrade to the same version using the same ISO image
3. `kubectl exec` into the fluentd Pod and mess up a random log file, for example, `echo "AA AA AA" >> /archive/logs/harvester-system.harvester-86bfd6897c-gjqtr.apiserver.20230323.log`
4. Click "Download Log" on the UI
5. Check the downloaded log archive, the only file in JSON format should be the file you messed up with, other files should be normal

### DIFF
--- a/pkg/api/upgradelog/handler.go
+++ b/pkg/api/upgradelog/handler.go
@@ -46,6 +46,7 @@ do
 		echo "Failed to process the file $f with ret=$ret"
 		echo "Copy the original file to the destination"
 		cp $f $tmpdir/logs/$f
+		unset ret
 	fi
 done
 


w13915984028 (commented):
regarding https://man7.org/linux/man-pages/man1/unset.1p.html

```
       If neither -f nor -v is specified, name refers to a variable; if
       a variable by that name does not exist, it is unspecified whether
       a function by that name, if any, shall be unset.
```

the better solution will be:
```
ret=0 # declare var first, as the code is not put under a function, the `local` will not work
for...
do
  .... || ret=$?
  if ret != 0
  ...
  ret=0 # set back to 0
done
```

but for here, a simple nested shell code, the `unset` also works.

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

bk201:
@mergifyio backport v1..1

bk201:
@mergifyio backport v1.1

PR #3718, opened 2023-03-24
Cut v1.1.2-rc4

bk201:
Deps:

- harvester-network-controller: v0.3.2-rc4
- support-bundle-kit: v0.0.20
- GUI: v1.1.2-rc4
- Installer:v1.1.2-rc4


Vicente-Cheng (approved):
Tested with the upgrade path and basic operation. Also, create a downstream cluster.
LGTM, thanks!

guangbochen (approved):
LGTM and the upgrade works, also the bundles and charts are all upgraded correctly.
image

Issue #3739, opened 2023-03-30
[BUG] When we add another management network to VM and save. VM keeps in `Starting` state


TachunLin:
**Describe the bug**

When we add a new `management network` to the virtual machine which already have an external vlan on it. (Without stop VM first)
After we save it and restart VM, the VM keep in `Starting` state 
![image](https://user-images.githubusercontent.com/29251855/228782221-bc032166-71a2-4b51-80a4-1c58e58dfead.png)

And event have `FailedMount` and `FailedMapVolume` error
![image](https://user-images.githubusercontent.com/29251855/228782266-cbfcb7ae-d029-4eab-a9ff-d82de1e57b14.png)

**To Reproduce**
1. Create a cluster network 
2. Create a network config on the cluster network and set uplink NIC to **All nodes** 
3. Create a external vlan network `vlan101`
4. Create a VM and set external vlan `vlan101`
5. Wait the VM in Running state
6. Edit the VM
7. Add new network and set `management network` 
8. Click save and restart VM
![image](https://user-images.githubusercontent.com/29251855/228789007-83cb1ad5-2c35-44e3-a441-ce6c6d635282.png)

**Expected behavior**

Should be able to start VM correctly after adding the management network to VM

**Support bundle**

[supportbundle_fe5a95fb-b45e-4d62-87e6-b1cdba8bb20d_2023-03-30T08-39-45Z.zip](https://github.com/harvester/harvester/files/11109719/supportbundle_fe5a95fb-b45e-4d62-87e6-b1cdba8bb20d_2023-03-30T08-39-45Z.zip)

**Environment**
 - Harvester ISO version: `v1.1.2-rc5`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 4 nodes kvm machines

**Additional context**
This issue did not happen when we add management network on VM have `vlan1` network

**Workaround**
Stop the VM and click start again. The VM can start in `Running` state correctly

rrajendran17:
@TachunLin I tried this in v.1.3.1 (different iterations with similar steps) and do not see any issues.Possibly should have been fixed by https://github.com/harvester/harvester/pull/3741.Please let me know your views.

lanfon72:
close as unable to reproduce in `v1.5.0-dev-20250203`.


Test Information
----
* Environment: **baremetals DL360G9 4 nodes**
* Harvester Version: **v1.5.0-dev-20250203**
* **ui-source** Option: **External**

### Verify Steps
[...]

Issue #3727, opened 2023-03-28
[BUG] VM stuck in *Starting* with _FailedMapVolume_ event


lanfon72:
### Describe the bug
This could be reproduced manually, but it is discovered by e2e testing.

![image](https://user-images.githubusercontent.com/5169694/228180284-7b5d3342-0180-41cd-b726-98c43215d525.png)
![image](https://user-images.githubusercontent.com/5169694/228180341-a824c78b-faff-4135-86ce-2dfbe70636c8.png)

### To Reproduce
1. Install Harvester with any nodes
2. Create a image for VM creation
3. Create VM `vm1` and start it
4. Stop `vm1` and start it again when its **State** becomes `stop`

### Expected behavior
VM should be started successfully

## Environment:
 - Harvester ISO version: **v1.1-65768fee-head** and **v1.1.2-rc4**
 - **ui-source** Option: **Auto**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 5 nodes**

### Additional context
The bug could be resolved with restart the VM

### Support bundle
[supportbundle_26c7012a-9fe8-49ef-88a2-105bd2e675ae_2023-03-28T08-51-34Z.zip](https://github.com/harvester/harvester/files/11087335/supportbundle_26c7012a-9fe8-49ef-88a2-105bd2e675ae_2023-03-28T08-51-34Z.zip)

lanfon72:
when do `restart` will also encounter this occassionally. (tested with `v1.1.2-rc5`, KVM environment 3 nodes)

lanfon72:
Tested this could be temporarily resolved with the workaround (might need several tries):
1. Stop the VM
2. check the VM is in `Off` state
3. Start the VM

When the VM was encountered this issue when doing `restart`, you will encounter it again if you try `restart` after workaround.

to notice `softreboot` will not be affected.

lanfon72:
For VM which created with `halted` and do `start` will also encounter this bug occasionally.

lanfon72:
remove label `not-require/test-plan`, we would need to create test cases to test this scenario.

bk201:
It's pretty easy to reproduce this by restarting a VM.

lanfon72:
Add code snippet for reproducing the bug.

```python3
#!/usr/bin/env python3
import sys
from time import sleep
from itertools import cycle
from datetime import datetime, timedelta

import urllib3

from harvester_api import API

[...]

Issue #3681, opened 2023-03-20
[BUG] Upgrade 3 nodes Harvester stuck in `Pre-drained` of the first node (create volume with snapshot)


TachunLin:
**Describe the bug**

* When upgrade from v1.1.1 to v1.1.2-rc3, the upgrade stuck in the `Pre-drained` state of the first node
![image](https://user-images.githubusercontent.com/29251855/226308714-98299bdc-f672-438a-a76b-77476d2b8c6a.png)

* Prior to upgrade, we have vm backup created while vm is running, create vm snapshot and create volume with volume snapshot. 

* Some pod keep in pending state `rancher-7d68567b9d-ldvwz`, `harvester-8546cc6cf4-jln8c` and `harvester-webhook-79df879d8b-gv4z8` 
![image](https://user-images.githubusercontent.com/29251855/226308111-3ac897d1-3125-4a23-b14a-f5e73c7edcea.png)

**To Reproduce**
1.  Provision a three nodes `v1.1.1` Harvester cluster 
2. Create three images
3. Create three virtual machines separate on each node
4. Create three VM backups
5. Add some file to running VM
6. create two VM snapshots 
7. Create two new volume `opensuse-volume`, `ubuntu-volume` 
8. Create two volume snapshots
9. Perform offline upgrade to v1.1.2-rc3

**Expected behavior**

Should be able to upgrade successfully with no stuck in the `Pre-drained` node state when we have vm backup, vm and volume snapshot

**Support bundle**

[supportbundle_cf1194db-d39e-42a4-81f3-153e3d2c7a92_2023-03-20T10-05-48Z.zip](https://github.com/harvester/harvester/files/11016869/supportbundle_cf1194db-d39e-42a4-81f3-153e3d2c7a92_2023-03-20T10-05-48Z.zip)

**Environment**
 - Harvester ISO version: v1.1.1 -> v1.1.2-rc3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 nodes bare machines

bk201:
This is a similar issue with https://github.com/harvester/harvester/issues/3648 and will be addressed together in https://github.com/harvester/harvester/pull/3670.

TachunLin:
Verified **fixed** on `v1.1.2-rc4` (3/24). Close this issue.

### Result
We can upgrade successfully with no stuck in the Pre-drained node state when Harvester have vm backup, vm snapshot and volume snapshot created before upgrade


[...]

PR #3741, opened 2023-03-30
Fix volume detach

futuretea:

**Problem:**

The pod  status of the restarted VM is `Succeeded` for a moment, causing the volume controller to detach volumes attached by longhorn, resulting  volumes cannot be mounted in the NodeStageVolume stage

**Solution:**

if volume's workload.WorkloadType is `VirtualMachineInstance`, add more check

**Related Issue:**
- https://github.com/harvester/harvester/issues/3648
- https://github.com/harvester/harvester/issues/3681
- https://github.com/harvester/harvester/issues/3739
- https://github.com/harvester/harvester/issues/3727
**Test plan:**

VMs can be restarted normally

More test refer to the test plan in https://github.com/harvester/harvester/pull/3670#issue-1628642961

### DIFF
--- a/pkg/controller/master/volume/controller.go
+++ b/pkg/controller/master/volume/controller.go
@@ -8,7 +8,9 @@ import (
 	v1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	kubevirtv1 "kubevirt.io/api/core/v1"
 
+	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
 	ctllonghornv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta1"
 	ctlsnapshotv1 "github.com/harvester/harvester/pkg/generated/controllers/snapshot.storage.k8s.io/v1beta1"
 	"github.com/harvester/harvester/pkg/indexeres"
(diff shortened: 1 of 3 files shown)


FrankYang0529:
Hi Hang, thanks for the hot fix. #3727 is not always reproducible in my environment. If a pod takes a longer time to terminate, volume can still be attached for a new pod.

#3727  may happen in the container workload too. If we want to take care of it, it's better to add a new `attach-volume-controller`. This controller can watch pods. If there is a pending pod with LH PVCs, the controller tries to attach the volumes.

cc @bk201 WDYT? Thanks.

FrankYang0529:
It looks like we have some edge case like:

1. Stop a VM.
2. If the volume gets into detaching, start the VM immediately.
3. Volume can't be attached again.

https://user-images.githubusercontent.com/4927361/228999261-302235ee-5f75-494d-9c84-c12ce150498e.mov

futuretea:
Thanks, I also found that there was one scenario that I didn't consider
1. Delete the VM but not the volume
2. snapshot the volume
3. Wait for the snapshot to complete

The expected behavior was that the volume would be detached after the snapshot was completed, but it did not. The reason is that the workloadsStatus of volume cannot truly reflect the real-time relationship between volume and workload.

futuretea:
I think I can turn this PR off because there are a few scenarios that are not considered. @FrankYang0529 feel free to create a new PR. CC @bk201

bk201:
@FrankYang0529 Please see if we can base on @futuretea's work and make him a co-author.

PR #3743, opened 2023-03-31
fix: detach and attach volumes

### LINKED ISSUE #3727: [BUG] VM stuck in *Starting* with _FailedMapVolume_ event

FrankYang0529:
**Problem:**
When a previous pod is terminated and the volume is in detaching state, starting a new pod with the volume may have error, because the volume can't be attached.  

**Solution:**
Add a controller to watch pods. If there is a pending pod, the controller tries to attach volumes.

**Related Issue:**
https://github.com/harvester/harvester/issues/3727

**Test plan:**

### Case 1: hotplug volumes

1. Create a new VM.
2. Attach a hotplug volume 1.
3. Attach another hotplug volume 2.
4. Remove the hotplug volume 1. (need to restart VM)
5. Remove the hotplug volume 2. (need to restart VM)
6. All steps work fine, and the VM can be restarted.

### Case 2: Stop and restart VM

1. Create a new VM.
2. Use CLI to watch VM's volumes.
3. Stop the VM.
4. Once VM's volumes get into detaching state, start the VM again.
5. All steps work fine, and the VM can be restarted.

All test cases in https://github.com/harvester/harvester/pull/3670

**Additional context**
Thanks @futuretea's another PR https://github.com/harvester/harvester/pull/3741 to fix #3727. We learned some experience from it and knew that if the volume is in detaching state and another new pod uses it, LH may not attach the volume correctly. This is why we add `attach-volume-controller` in this PR.

### DIFF
--- a/pkg/controller/master/volume/controller.go
+++ b/pkg/controller/master/volume/controller.go
@@ -20,6 +20,8 @@ const (
 )
 
 type Controller struct {
+	podCache         v1.PodCache
+	podController    v1.PodController
 	pvcCache         v1.PersistentVolumeClaimCache
 	volumes          ctllonghornv1.VolumeClient
 	volumeController ctllonghornv1.VolumeController
@@ -65,14 +67,13 @@ func (c *Controller) DetachVolumesOnChange(_ string, volume *lhv1beta1.Volume) (
 
 func (c *Controller) checkDetachVolume(pvc *corev1.PersistentVolumeClaim) (canDetach, watchAgain bool, err error) {
 	// 1. check whether any pod uses the PVC
-	volume, err := c.volumeCache.Get(util.LonghornSystemNamespaceName, pvc.Spec.VolumeName)
+	pods, err := c.podCache.GetByIndex(indexeres.PodByPVCIndex, fmt.Sprintf("%s/%s", pvc.Namespace, pvc.Name))
 	if err != nil {
-		return false, false, fmt.Errorf("can't find volume %s/%s, err: %w", util.LonghornSystemNamespaceName, pvc.Spec.VolumeName, err)
+		return false, false, fmt.Errorf("can't get pods by index %s with pvc %s/%s, err: %w", indexeres.PodByPVCIndex, pvc.Namespace, pvc.Name, err)
 	}
-	for _, workload := range volume.Status.KubernetesStatus.WorkloadsStatus {
-		// For running workload, we don't want to watch again
-		if workload.PodStatus == string(corev1.PodRunning) || workload.PodStatus == string(corev1.PodPending) {
-			logrus.Debugf("workload %s/%s is %s, don't detach pvc: %s/%s", volume.Status.KubernetesStatus.Namespace, workload.WorkloadName, workload.PodStatus, pvc.Namespace, pvc.Name)
+	for _, pod := range pods {
+		if pod.Status.Phase == corev1.PodRunning || pod.Status.Phase == corev1.PodPending {
+			logrus.Debugf("pod %s/%s is %s, don't detach pvc: %s/%s", pod.Namespace, pod.Name, pod.Status.Phase, pvc.Namespace, pvc.Name)
 			return false, false, nil
 		}
 	}
@@ -115,6 +116,16 @@ func (c *Controller) detachVolume(pvc *corev1.PersistentVolumeClaim) error {
 		logrus.Infof("detach volume %s", volCpy.Name)
 		if _, err = c.volumes.Update(volCpy); err != nil {
 			return err
+		}
+		// A race condition is that, we detach the volume and create a new pod at the same time.
+		// In this situation, attach-volume-controller may get an attached volume, but the volume is detaching in detach-volume-controller.
+		// So we need to enqueue the pod to make sure the volume can be attached again.
+		pods, err := c.podCache.GetByIndex(indexeres.PodByPVCIndex, fmt.Sprintf("%s/%s", pvc.Namespace, pvc.Name))
+		if err != nil {
+			return err
+		}
+		for _, pod := range pods {
+			c.podController.EnqueueAfter(pod.Namespace, pod.Name, attachVolumeEnqueueInterval)
 		}
 	}
 	return nil
--- a/pkg/controller/master/volume/pod_controller.go
+++ b/pkg/controller/master/volume/pod_controller.go
@@ -1,0 +1,78 @@
+package volume
+
+import (
+	"fmt"
+	"time"
+
+	lhv1beta1 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+	longhorntypes "github.com/longhorn/longhorn-manager/types"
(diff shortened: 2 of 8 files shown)


pkg/controller/master/volume/pod_controller.go
```diff
@@ -0,0 +1,72 @@
...
+
+		pvc, err := c.pvcCache.Get(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
+		if err != nil {
+			return pod, fmt.Errorf("can't find pvc %s/%s, err: %w", pod.Namespace, volume.PersistentVolumeClaim.ClaimName, err)
+		}
+		if util.GetProvisionedPVCProvisioner(pvc) != longhorntypes.LonghornDriverName {
+			continue
+		}
+
+		volume, err := c.volumeCache.Get("longhorn-system", pvc.Spec.VolumeName)
```
futuretea: const

pkg/controller/master/volume/pod_controller.go
```diff
@@ -0,0 +1,72 @@
...
+			return pod, fmt.Errorf("can't find volume %s, err: %w", pvc.Spec.VolumeName, err)
+		}
+		if isVolumeAttached(volume) {
+			continue
+		}
+		if volume.Status.State == lhv1beta1.VolumeStateDetaching {
+			c.podController.EnqueueAfter(pod.Namespace, pod.Name, attachVolumeEnqueueInterval)
+			break
+		}
+		if volume.Status.State == lhv1beta1.VolumeStateDetached {
```
futuretea: do we need to check if volume.Spec.NodeID != volume.Status.OwnerID ?

pkg/controller/master/volume/pod_controller.go
FrankYang0529: Fixed it. Thanks.

pkg/controller/master/volume/pod_controller.go
FrankYang0529: Add a check for it. Thanks.

pkg/controller/master/volume/pod_controller.go
```diff
@@ -0,0 +1,74 @@
...
+
+		pvc, err := c.pvcCache.Get(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
+		if err != nil {
+			return pod, fmt.Errorf("can't find pvc %s/%s, err: %w", pod.Namespace, volume.PersistentVolumeClaim.ClaimName, err)
+		}
+		if util.GetProvisionedPVCProvisioner(pvc) != longhorntypes.LonghornDriverName {
+			continue
+		}
+
+		volume, err := c.volumeCache.Get(util.LonghornSystemNamespaceName, pvc.Spec.VolumeName)
```
futuretea, line 56: if pvc.Spec.VolumeName == "" {
			continue
		}

pkg/controller/master/volume/pod_controller.go
```diff
@@ -0,0 +1,78 @@
...
+		pvc, err := c.pvcCache.Get(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
+		if err != nil {
+			return pod, fmt.Errorf("can't find pvc %s/%s, err: %w", pod.Namespace, volume.PersistentVolumeClaim.ClaimName, err)
+		}
+		if util.GetProvisionedPVCProvisioner(pvc) != longhorntypes.LonghornDriverName {
+			continue
+		}
+		if pvc.Spec.VolumeName == "" {
+			c.podController.EnqueueAfter(pod.Namespace, pod.Name, attachVolumeEnqueueInterval)
+			break
```
Vicente-Cheng, line 53: Should we use `continue` as @futuretea reminder?
That could check the next volume on this pod.
Or is there any reason we need to enqueue again to wait for the next round?

futuretea: EnqueueAfter is better, will check volume after next enqueue

Vicente-Cheng: OK, thanks for the explanation!

FrankYang0529: Hi @Vicente-Cheng, thanks for the question. I don't have a strong reason here. I just think it's safer if we can have more checks.

Vicente-Cheng (approved):
LGTM, and tested with the following cases.
3 VMs environment,
- 1 VM (shutdown) with backup and snapshot
- 1 VM + 1 hotplug volume (shutdown) with backup and snapshot
- 1 VM + 1 hotplug volume (running) with backup and snapshot
The upgrade looks good, and the running VM cloud migrates/restarts normally.
Thanks!

pkg/controller/master/volume/controller.go
```diff
@@ -116,6 +117,16 @@ func (c *Controller) detachVolume(pvc *corev1.PersistentVolumeClaim) error {
...
+		// So we need to enqueue the pod to make sure the volume can be attached again.
+		pods, err := c.podCache.GetByIndex(indexeres.PodByPVCIndex, fmt.Sprintf("%s/%s", pvc.Namespace, pvc.Name))
+		if err != nil {
+			return err
+		}
+		for _, pod := range pods {
+			c.podController.EnqueueAfter(pod.Namespace, pod.Name, attachVolumeEnqueueInterval)
+		}
 	}
 	return nil
```
futuretea, line 131: not about the test case 1
```go
func (c *Controller) detachVolume(pvc *corev1.PersistentVolumeClaim) error {
	volume, err := c.volumeCache.Get(util.LonghornSystemNamespaceName, pvc.Spec.VolumeName)
	if err != nil {
		return fmt.Errorf("can't find volume %s/%s, err: %w", util.LonghornSystemNamespaceName, pvc.Spec.VolumeName, err)
	}

	if !isVolumeAttached(volume) {
		return nil
	}
	// other codes
	return nil
}
```

futuretea (commented):
test case1 failed on commit 3791fcc57de7845666f920bb375f5951070dfec6 
image

FrankYang0529:
Close the issue because we have a better solution.

https://github.com/harvester/harvester/pull/3749

PR #3749, opened 2023-04-05
fix: detach volume when the node is on pre-drained state

futuretea:

In order to avoid https://github.com/harvester/harvester/issues/3681, we need to detach volumes
To reduce the risk (https://github.com/harvester/harvester/issues/3727), detach should be only done during node upgrades

**Test manifest setup**
  - Create a vm `foo`
  - Create a vm backup `foo-backup` for vm `foo`
  - Create a vm snapshot `foo-snapshot` for vm `foo`
  - Create a vm `bar`
  - Stop the vm `bar`
  - Create a vm backup `bar-backup` for vm `bar`
  - Create a vm snapshot `test-snapshot` for vm `bar`
  - Create a volume `bar-volume`
  - Create a volume snapshot `bar-volume-snapshot`  for volume `bar`
 
**Test plan:**

**Full Upgrade Test**

1. Prepare a 3 nodes Harvester cluster (v1.1.1 version)
2. Follow the **Test manifest setup**
3. Build iso from the PR, refer to https://github.com/harvester/harvester/wiki/Build-ISO-images and https://github.com/harvester/harvester/wiki/Speed-up-Harvester-ISO-build
4. Upgrade cluster refer to https://docs.harvesterhci.io/v1.1/upgrade/automatic#prepare-an-air-gapped-upgrade


w13915984028 (commented):
The first question is:

If below code is running in controller, then it will work in v1.1.2->v1.2.0 upgrade path, how to deal with v1.1.1->v1.1.2 upgrade path?

In general, if v1.1.2 plans to solve issues existing in v1.1.1 version, the upgrade script is the place to fix them.

The `Full Upgrade Test` section https://github.com/harvester/harvester/pull/3749#issue-1655134714 may not work, because in v1.1.1->v1.1.2 upgrade path, the `Harvester` POD image is not replaced in advance.

thanks.

futuretea:
> The first question is:
> 
> If below code is running in controller, then it will work in v1.1.2->v1.2.0 upgrade path, how to deal with v1.1.1->v1.1.2 upgrade path?
> 
> In general, if v1.1.2 plans to solve issues existing in v1.1.1 version, the upgrade script is the place to fix them.
> 
> thanks.

@w13915984028 Refer to https://docs.harvesterhci.io/v1.1/upgrade/troubleshooting#diagnose-the-upgrade-flow

We will upgrade Harvester first, so this controller can deal with v1.1.1->v1.1.2 upgrade path

pkg/controller/master/node/node_volume_detach_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	// get all longhorn volume attach to this node
+	volumes, err := c.volumeCache.GetByIndex(indexeres.VolumeByNodeIndex, node.Name)
+	if err != nil {
+		return node, err
+	}
+	waitPVCs := make([]*corev1.PersistentVolumeClaim, 0, len(volumes))
+	for _, volume := range volumes {
+		// get pvc by volume name
+		pvcNamespace := volume.Status.KubernetesStatus.Namespace
+		pvcName := volume.Status.KubernetesStatus.PVCName
```
w13915984028: not 100% sure, LH may have orphan volume (no PVC),  but that kind of volume should not affect later process,  we may skip PVC non-existing error

when with volume state check at first, to here is safe, because `attached` volume must have a PVC

pkg/controller/master/node/node_volume_detach_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	logrus.Debugf("node %s is in draining, try to detach unused volumes", node.Name)
+
+	// get all longhorn volume attach to this node
+	volumes, err := c.volumeCache.GetByIndex(indexeres.VolumeByNodeIndex, node.Name)
+	if err != nil {
+		return node, err
+	}
+	waitPVCs := make([]*corev1.PersistentVolumeClaim, 0, len(volumes))
+	for _, volume := range volumes {
```
w13915984028, line 75: Need to exclude `detached` / filter `attached` volumes  at first, no need to check all volumes in following steps.

```
harv2:~ # kk get volume -A
NAMESPACE         NAME                                       STATE      ROBUSTNESS   SCHEDULED   SIZE          NODE    AGE
longhorn-system   pvc-6be2504c-dc80-4824-a8ec-8d19ec3b2e0b   attached   degraded                 5368709120    harv2   12d
longhorn-system   pvc-6c6a5088-38c1-41f0-b3f9-7b7e8668fb59   attached   degraded                 53687091200   harv2   12d
longhorn-system   pvc-c5083f9c-cb04-4a5c-8869-8a2d3bb23d86   attached   degraded                 2147483648    harv2   12d


longhorn-system   pvc-f160a980-0d0e-4a48-bfed-7099e46e3b3e   detached   unknown                  1073741824            62m
```

pkg/controller/master/node/node_volume_detach_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	return nil
+}
+
+// OnNodeChanged is called when a node is changed
+func (c *nodeVolumeDetachController) OnNodeChanged(key string, node *corev1.Node) (*corev1.Node, error) {
+	if node == nil || node.DeletionTimestamp != nil {
+		return node, nil
+	}
+
+	// skip if node is not in draining
```
w13915984028: add further check of upgrade context to make sure this is only working in upgrade ?

pkg/controller/master/node/node_volume_detach_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	if !isNodeInDraining(node) {
+		return node, nil
+	}
+	logrus.Debugf("node %s is in draining, try to detach unused volumes", node.Name)
+
+	// get all longhorn volume attach to this node
+	volumes, err := c.volumeCache.GetByIndex(indexeres.VolumeByNodeIndex, node.Name)
```
w13915984028: LH defined volume state:
```
const (
	VolumeStateCreating  = VolumeState("creating")
	VolumeStateAttached  = VolumeState("attached")
	VolumeStateDetached  = VolumeState("detached")
	VolumeStateAttaching = VolumeState("attaching")
	VolumeStateDetaching = VolumeState("detaching")
	VolumeStateDeleting  = VolumeState("deleting")
)
```

For `creating`, `attaching`, `detaching`, add to wait directly

For `detached`, `deleting`, simply skip them

For `attached`,  continue following checks

w13915984028 (commented):
questions:

pkg/controller/master/node/node_volume_detach_controller.go
futuretea: @w13915984028  Updated

w13915984028 (approved):
LGTM, thanks.

FrankYang0529 (approved):
LGTM. Thanks.

bk201 (approved):
LGTM, testing upgrade works.

Vicente-Cheng (approved):
Tested with the upgrade.
The scenario is below:
1. Create VM
2. Create VM snapshot
3. Create VM from the above snapshot
4. Shutdown the source VM
5. Backup/Snapshot the volume of the source VM
6. Do upgrade
everything looks good after the upgrade. Thanks for it!

guangbochen (approved):
LGTM and verified that the upgrade from v1.1.1 works fine for me.
image
Also, @futuretea let's delete the `Simple Fast Test` case if they are no longer used, thanks.

guangbochen:
@mergify backport v1.1

Issue #3756, opened 2023-04-10
[FEATURE] Support LH priorityClass


FrankYang0529:
**Is your feature request related to a problem? Please describe.**
LH is a critical service. We don't want that it is the first to be evicted when eviction happens, so we would like to set priority class to LH workloads.

**Describe the solution you'd like**
We must detach all volumes when updating the priority class in old clusters. We have to update the LH helm chart to support old and new values in the same template.

**Describe alternatives you've considered**
If we want to support the value in old clusters, we should introduce the state machine like storage network.

**Additional context**

bk201:
We need a doc to describe what to do in the upgrade path. Reviewing: https://github.com/harvester/harvesterhci.io/pull/37

TachunLin:
### Test Plans

#### New Cluster 

- [x] 1. When the cluster is ready, check Longhorn system-managed components have priority class set.

    <details><summary>Test Result $~~$ $\color{green}{\textsf{PASS}}$ </summary>
    
        
      node1:~ # kubectl get priorityclass system-cluster-critical
      NAME                      VALUE        GLOBAL-DEFAULT   AGE
      system-cluster-critical   2000000000   false            173m
        
          
    </details> 

- [x] 2. Verify all system-managed components' pods have the correct priority, including 


    <details><summary>Test Result $~~$ $\color{green}{\textsf{PASS}}$ </summary> 
    
        node1:~ # kubectl get pods -n longhorn-system -o custom-columns="Name":metadata.name,"Priority":.spec.priority
        Name                                                  Priority
        csi-attacher-6d46d6df-6wdfk                           2000000000
        csi-attacher-6d46d6df-d9sg6                           2000000000
        csi-attacher-6d46d6df-fnv5w                           2000000000
        csi-provisioner-789fb46d84-224k5                      2000000000
        csi-provisioner-789fb46d84-q7b2b                      2000000000
        csi-provisioner-789fb46d84-rdclf                      2000000000
        csi-resizer-7b4c499ff9-df79r                          2000000000
        csi-resizer-7b4c499ff9-dv9p9                          2000000000
        csi-resizer-7b4c499ff9-vfpmm                          2000000000
        csi-snapshotter-865d9b6f88-76nrt                      2000000000
        csi-snapshotter-865d9b6f88-rzkxj                      2000000000
        csi-snapshotter-865d9b6f88-x4x95                      2000000000
        engine-image-ei-1d169b76-5mw5m                        2000000000
[...]

PR #3757, opened 2023-04-10
feat(longhorn): add priority class

FrankYang0529:
**Problem:**
Support LH priority class.

**Solution:**
Use helm chart functions to determine whether the cluster is new or old. If it's new, we set default `system-cluster-critical`. If it's old, we don't update the value.

**Related Issue:**
https://github.com/harvester/harvester/issues/3756

**Test plan:**

### Case 1: Old cluster remains the original value

1. Create a v1.1.1 harvester cluster.
3. Priority class in `longhorn-system/longhorn-default-setting` config map is empty.

### Case 2: New cluster use `system-cluster-critical`

1. Create a new harvester cluster from this branch.
2. Priority class in `longhorn-system/longhorn-default-setting` config map is `system-cluster-critical`.


FrankYang0529:
Close the PR. I update related field in upstream. https://github.com/longhorn/longhorn/pull/5784

PR #3761, opened 2023-04-11
feat(setting): add rancher-manager-support

FrankYang0529:
**Solution:**
Add a setting to show the rancher management page.

**Related Issue:**

**Test plan:**

### Setup

2. Update `status.dynamic` in `multi-cluster-management` feature to `true`.

### Case 1: `rancher-manager-support` can't be enabled if `mcm-cluster-management` is disabled.

1. Change `rancher-manager-support` setting to `true`. We should get an error.

### Case 2: `rancher-manager-support` can be enabled/disabled if `mcm-cluster-management` is enabled.

1. Update `multi-cluster-management=true` in `cattle-system/rancher` deployment.
2. Change `rancher-manager-support` setting to `true`. We should not get any error.
3. Change `rancher-manager-support` setting to `false`. We should not get any error.

### Case 3: Only `true` and `false` are valid values

1. Change `rancher-manager-support` setting to `test`. We should get an error.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -18,6 +18,7 @@ import (
 	// S3 Ref: https://github.com/longhorn/backupstore/blob/3912081eb7c5708f0027ebbb0da4934537eb9d72/s3/s3.go#L33-L37
 	_ "github.com/longhorn/backupstore/nfs" //nolint
 	_ "github.com/longhorn/backupstore/s3"  //nolint
+	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	"github.com/rancher/wharfie/pkg/registries"
 	"github.com/rancher/wrangler/pkg/slice"
 	"github.com/sirupsen/logrus"
@@ -39,6 +40,10 @@ import (
 	tlsutil "github.com/harvester/harvester/pkg/util/tls"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+const (
+	mcmFeature = "multi-cluster-management"
 )
 
 var certs = getSystemCerts()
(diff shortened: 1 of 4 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -665,3 +674,31 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
...
+	}
+
+	value := strings.ToLower(newSetting.Value)
+	if (value != "true") && (value != "false") {
+		return werror.NewInvalidError("value should be true or false", "value")
+	}
+
+	if value == "false" {
+		return nil
+	}
```
guangbochen: the `value` variable can be defined as a boolean type which will make it more readable and also help to simplify the code. e.g.,
```
enableManager, err := strconv.ParseBool(strings.ToLower(newSetting.Value))
if err != nil {
    return werror.NewInvalidError("value should be either true or false", "value")
}

if !enableManager {
    return nil
}
```

pkg/webhook/resources/setting/validator.go
```diff
@@ -665,3 +674,31 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
...
+
+	// check if the multi-cluster-management feature is enabled
+	feature, err := v.featureCache.Get(mcmFeature)
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+
+	if (feature.Spec.Value == nil && !feature.Status.Default) ||
+		(feature.Spec.Value != nil && !*feature.Spec.Value) {
+		return werror.NewInvalidError("the multi-cluster-management feature is not enabled", "value")
```
guangbochen:
```suggestion
		return werror.NewInvalidError("cannot enable Rancher Manager support. The multi-cluster-management feature is not enabled", "value")
```

pkg/webhook/resources/setting/validator.go
FrankYang0529: Updated it. Thanks.

pkg/webhook/resources/setting/validator.go
FrankYang0529: Updated it. Thanks.

guangbochen (approved):
LGTM, thanks.

futuretea (approved):
LGTM, Thanks

bk201 (approved):
LGTM!

Issue #3773, opened 2023-04-14
[BUG] Image stuck in 0% progress but showing *Active*


lanfon72:
### Describe the bug
![image](https://user-images.githubusercontent.com/5169694/232009014-33ada664-8b3e-49ed-b207-0f3cae7a50d2.png)

### To Reproduce
1. Install Harvester with any nodes
2. Create a Image by URL

### Expected behavior
State of Image should not changed to **Active** if it is not usable.

## Environment:
 - Harvester ISO version: **v1.1.2-rc7**
 - **ui-source** Option: **Auto**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 3 nodes**

### Additional context
Discovered with do e2e testing for `integration/test_vm.py`

### Support bundle
[supportbundle_9e8eea13-86fe-4b6c-bb85-52d734bda8a7_2023-04-14T09-44-05Z.zip](https://github.com/harvester/harvester/files/11231430/supportbundle_9e8eea13-86fe-4b6c-bb85-52d734bda8a7_2023-04-14T09-44-05Z.zip)

lanfon72:
response json of the Image (via API calls):
```json
{
    "apiVersion": "harvesterhci.io/v1beta1",
    "kind": "VirtualMachineImage",
    "metadata":
    {
        "annotations":
        {
            "field.cattle.io/description": ""
        },
        "creationTimestamp": "2023-04-14T09:00:06Z",
        "finalizers":
        [
            "wrangler.cattle.io/vm-image-controller"
        ],
        "generation": 2,
        "managedFields":
        [
            {
                "apiVersion": "harvesterhci.io/v1beta1",
                "fieldsType": "FieldsV1",
                "fieldsV1":
                {
                    "f:metadata":
                    {
                        "f:finalizers":
                        {
                            ".":
                            {},
                            "v:\"wrangler.cattle.io/vm-image-controller\"": {}
                        }
                    },
                    "f:spec":
                    {
                        "f:checksum":
                        {},
                        "f:pvcName":
                        {},
                        "f:pvcNamespace":
                        {}
                    },
                    "f:status":
                    {
                        ".":
                        {},
                        "f:failed":
                        {}
                    }
                },
                "manager": "harvester",
                "operation": "Update",
                "time": "2023-04-14T09:00:06Z"
            },
            {
                "apiVersion": "harvesterhci.io/v1beta1",
                "fieldsType": "FieldsV1",
[...]

PR #3809, opened 2023-04-21
fix(image): delete and create race condition

FrankYang0529:
**Problem:**
We didn't set `OwnerReference` in BackingImage, so VMImage can be deleted when BackingImage is in deleting progress. At this time, if we create another VMImage with the same name, it may get a BackingImage with `DeletionTimestamp`. Because the status of the BackingImage is good, we don't process the VMImage. Finally, we may get a VMImage with 0% progress on the dashboard ([ref](https://github.com/harvester/harvester/issues/3773#issuecomment-1508253405)). 

**Solution:**
If the BackingImage has `DeletionTimestamp`, we enqueue the VMImage again.

**Related Issue:**
https://github.com/harvester/harvester/issues/3773

**Test plan:**

We can not manually reproduce the issue, because the time gap is too small. We can run [harvester/tests](https://github.com/harvester/tests) to check it.

```sh
tox -e py38 -- harvester_e2e_tests --html=test_result.html -s -vv -k 'test_vm.py'
```

### DIFF
--- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -26,6 +26,10 @@ import (
 	"github.com/harvester/harvester/pkg/util"
 )
 
+const (
+	checkBackingImageInterval = 1 * time.Second
+)
+
 // vmImageHandler syncs status on vm image changes, and manage a storageclass & a backingimage per vm image
 type vmImageHandler struct {
 	httpClient        http.Client
@@ -64,6 +68,9 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
 			return image, err
 		}
 		needRetry = true
+	} else if backingImage.DeletionTimestamp != nil {
+		h.imageController.EnqueueAfter(image.Namespace, image.Name, checkBackingImageInterval)
+		return image, nil
 	} else {
 		for _, status := range backingImage.Status.DiskFileStatusMap {
 			if status.State == v1beta1.BackingImageStateFailed {


WebberHuang1118:
Hi Frank,

I'm wondering about certain cases, that may still get VMImages stuck under current solution.

Assuming the following circumstance:
1. A VMImage `image-abc` is created, and a BackingImage `default-image-abc` is also created along with it (for example from the VMImage file upload operation)
2. VMImage `image-abc` is deleted, and a deletion request is issued for BackingImage `default-image-abc`
3. Since the deletion for BackingImage `default-image-abc` isn't atomic, it could go through three phases in order
        a. The `DeletionTimestamp` field is null
        b. The `DeletionTimestamp` field is set
        c. BackingImage `default-image-abc` destroyed
4. If during BackingImage `default-image-abc` is in phase (a), a new VMImage `image-abc` is created with the same name, it would pass the checking about BackingImage
5. The BackingImage `default-image-abc` complete reamining phase (b) and (c), and be deleted finally
6. Since BackingImage `default-image-abc` no longer exists, the new VMImage `image-abc` may still go stuck

Is this possible to happen ? Need your comment on it

thanks,
Webber

FrankYang0529:
> 1. A VMImage `image-abc` is created, and a BackingImage `default-image-abc` is also created along with it (for example from the VMImage file upload operation)
> 2. VMImage `image-abc` is deleted, and a deletion request is issued for BackingImage `default-image-abc`
> 3. Since the deletion for BackingImage `default-image-abc` isn't atomic, it could go through three phases in order
>    a. The `DeletionTimestamp` field is null
>    b. The `DeletionTimestamp` field is set
>    c. BackingImage `default-image-abc` destroyed
> 4. If during BackingImage `default-image-abc` is in phase (a), a new VMImage `image-abc` is created with the same name, it would pass the checking about BackingImage
> 5. The BackingImage `default-image-abc` complete reamining phase (b) and (c), and be deleted finally
> 6. Since BackingImage `default-image-abc` no longer exists, the new VMImage `image-abc` may still go stuck
> 
> Is this possible to happen ? Need your comment on it

We have a [OnRemove](https://github.com/harvester/harvester/blob/8413da412fbfa0431c92d25f6c5a68ab35ac295b/pkg/controller/master/image/vm_image_controller.go#L87) function for VMImage. It adds a finalizer to VMImage. From [k8s finalizer document](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/):
> An object remains in a terminating state while controllers take the actions defined by the finalizers. After these actions are complete, the controller removes the relevant finalizers from the target object. When the `metadata.finalizers` field is empty, Kubernetes considers the deletion complete and deletes the object.

Based on the document, we will not have a situation that the VMImage is deleted and BackingImage's `DeletionTimestamp` is null.

WebberHuang1118 (approved):
LGTM, thanks

bk201 (approved):
LGTM!

guangbochen (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.1

PR #3815, opened 2023-04-21
Upgrade path for max-pods: 200 [Fixes #2707]

### LINKED ISSUE #2707: [FEATURE] increase 110 pods per node limit

tlehman:
This PR is for the upgrade path for issue #2707 

I tested this bash function on a 2-node cluster and it had the expected effect no the /etc/rancher/rancherd/config.yaml file. It also was idempotent, as expected.

Default max-pods for new installs: https://github.com/harvester/harvester-installer/pull/485

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -494,6 +494,15 @@ spec:
 EOF
 }
 
+# issue 2707 is where this default was set, https://github.com/harvester/harvester/issues/2707 
+set_max_pods() {
+  local cfg=$HOST_DIR/etc/rancher/rancherd/config.yaml
+  # if no max-pods is set, then set it to 200
+  if [ $(grep max-pods $cfg | wc -l) -eq 0 ]; then
+	printf "kubelet-arg:\n  max-pods: \"200\"" >> $cfg
+  fi
+}
+
 upgrade_os() {
   # The trap will be only effective from this point to the end of the execution
   trap clean_up_tmp_files EXIT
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -494,6 +494,15 @@ spec:
 EOF
 }
 
+# issue 2707 is where this default was set, https://github.com/harvester/harvester/issues/2707 
+set_max_pods() {
+  local cfg=$HOST_DIR/etc/rancher/rancherd/config.yaml
```
bk201, line 499: rancherd config is only loaded in bootstrap time. We need to update RKE2 config before upgrade RKE2.

bk201:
We'll continue working on the fix in https://github.com/harvester/harvester/pull/3916, thanks @tlehman

Issue #2204, opened 2022-04-28
[FEATURE] build-iso: migrate luet-makeiso to `elemental build-iso`


bk201:
**Is your feature request related to a problem? Please describe.**

https://github.com/rancher-sandbox/cOS-toolkit/issues/1217

cOS team build the luet-makeiso feature into elemental command. We should migrate to get further support.

https://github.com/harvester/harvester-installer/blob/1c1b8c4dfb7aa11fb7e912e1f3f5ca9414259e0f/Dockerfile.dapper#L2
https://github.com/harvester/harvester-installer/blob/1c1b8c4dfb7aa11fb7e912e1f3f5ca9414259e0f/scripts/package-harvester-os#L67

https://github.com/harvester/harvester/blob/e2985453c65193b407d47d53c3c96a249e2cf474/Dockerfile.dapper#L2

bk201:
Move to v1.2.0, luet-makeiso still works and this is only for the building process.

bk201:
Try to bump elemental to build ISO, but seems can't boot the ISO due to the change that only supports UEFI.

bk201:
Add `require/release-note`. Dropping legacy support needs a heads-up in the release note.

irishgordo:
@bk201 for validation should I also check the ability to upgrade to an iso built by `harvester-installer`?

I validated the artifacts were built:
```
╭─mike at suse-workstation-team-harvester in ~/Projects/harvester-installer/dist on master✔
╰─± ls -alh artifacts 
total 6.6G
drwxr-xr-x 2 mike mike 4.0K May 23 12:43 .
drwxr-xr-x 3 mike mike 4.0K May 23 12:41 ..
-rw-r--r-- 1 mike mike  60M Jul  1  2022 harvester-ea2c3e7-dirty-initrd-amd64
-rw-r--r-- 1 mike mike 735M Jul  1  2022 harvester-ea2c3e7-dirty-rootfs-amd64.squashfs
-rw-r--r-- 1 mike mike 8.7M Jun 17  2022 harvester-ea2c3e7-dirty-vmlinuz-amd64
-rw-r--r-- 1 mike mike 5.1G May 23 12:43 harvester-master-amd64.iso
-rw-r--r-- 1 mike mike  647 May 23 12:43 harvester-master-amd64.sha512
-rw-r--r-- 1 mike mike  86M May 23 12:41 harvester-master-initrd-amd64
[...]

PR #3821, opened 2023-04-24
Added elemental base image and rsync.

masteryyh:

**Problem:**

We need to migrate from `luet-makeiso` to `elemental build-iso`.

**Solution:**

Add `elemental` base image for elemental binary and `rsync` package for elemental `build-iso` command execution.

**Related Issue:**
https://github.com/harvester/harvester/issues/2204

**Test plan:**

Boot:
1. Build an ISO from `elemental-build` branch (both `harvester` and `harvester-installer` repo);
2. Try to boot up using this ISO in UEFI environment, it should boot up without problem;
3. Try to boot up using this ISO in legacy BIOS environment too.

Upgrade:
1. Build an ISO from `elemental-build` branch (both `harvester` and `harvester-installer` repo);
2. Install a clean environment using v1.1.2;
3. Create a http server to serve the ISO file;
4. Create a version object in Harvester and let Harvester upgrade to the ISO just built;
5. Check if Harvester upgraded successfully.

**Additional Information:**
We might need to create another issue to add legacy BIOS boot capability if it's necessary.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,5 +1,5 @@
 FROM quay.io/costoolkit/releases-green:luet-toolchain-0.21.2 as luet
-FROM quay.io/costoolkit/releases-green:luet-makeiso-toolchain-0.4.0 as makeiso
+FROM quay.io/costoolkit/elemental-cli:v0.2.5 as elemental
 
 FROM golang:1.18.3-buster
 
@@ -34,6 +34,7 @@ RUN apt-get update -qq && apt-get install -y --no-install-recommends \
         ca-certificates \
         curl \
         gnupg \
+        rsync \
     && rm -rf /var/lib/apt/lists/*; \
     \
     curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - >/dev/null; \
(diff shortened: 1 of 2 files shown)


Dockerfile.dapper
```diff
@@ -1,5 +1,6 @@
 FROM quay.io/costoolkit/releases-green:luet-toolchain-0.21.2 as luet
 FROM quay.io/costoolkit/releases-green:luet-makeiso-toolchain-0.4.0 as makeiso
+FROM quay.io/costoolkit/elemental-cli:v0.2.2 as elemental
```
guangbochen: ditto, should we use the latest version instead?

guangbochen (commented):
Have we tested the upgrade from the previous version with the new ISO image?

Dockerfile.dapper
```diff
@@ -1,5 +1,6 @@
 FROM quay.io/costoolkit/releases-green:luet-toolchain-0.21.2 as luet
 FROM quay.io/costoolkit/releases-green:luet-makeiso-toolchain-0.4.0 as makeiso
```
Vicente-Cheng: If we do not need luet-makeiso, we could remove all lines related to it. (include the `COPY` line below)

starbops (approved):
Tested upgrade from v1.1.2 to this PR with legacy BIOS and UEFI.
LGTM, thanks!

Vicente-Cheng (approved):
Tested with the upgrade with legacy mode. All basic operations and downstream cluster verify look fine. LGTM, thanks!

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM, thanks for this good enhancement!

PR #3841, opened 2023-04-28
feat(lh): add priority-class

FrankYang0529:
**Solution:**
Add priority-class to LH workload.

**Related Issue:**
https://github.com/harvester/harvester/issues/3756

**Test plan:**

**Case 1: New cluster**
1. Create a cluster from this branch.
2. Check whether LH workloads have priority-class.

**Case 2: Old cluster**
1. Create a cluster with an older version like `v1.1.2`.
3. LH workloads don't have priority-class if there is any attached volume.
4. Stop workloads that use volumes.
5. All LH pods restart.
6. Recheck LH workloads. They have priority-class now.

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -420,6 +420,7 @@ longhorn:
   defaultSettings:
     ## Specify the concurrent automatic engine upgrade per node limit
     concurrentAutomaticEngineUpgradePerNodeLimit: 3
+    priorityClass: system-cluster-critical
 
 rancherEmbedded: false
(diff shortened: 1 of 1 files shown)


guangbochen (commented):
should we consider adding all harvester-managed systems with this priority class together? cc @bk201

bk201:
We can also do it but prefer to do that in a separate issue and PR.

bk201 (approved):

guangbochen (approved):
LGTM, thanks. Do we need to backport this to v1.1?

PR #3845, opened 2023-05-02
feat(upgrade): add v1.1.2 pre-hook

FrankYang0529:
**Problem:**
In old clusters, the `status.dynamic` of the `multi-cluster-management` feature is `false`, so we can't change the feature's value.

**Solution:**
Set `status.dynamic` to `true`, so we can update the value of the `multi-cluster-management` feature to `true` in the new version.

**Related Issue:**

**Test plan:**

1. Install a v1.1.2 cluster.
2. Upgrade the cluster to the version of this branch.
3. Check whether the `status.dynamic` of the `multi-cluster-management` feature is `true`.

### DIFF
--- a/package/upgrade/migrations/upgrade_manifests/v1.1.2/pre-hook.sh
+++ b/package/upgrade/migrations/upgrade_manifests/v1.1.2/pre-hook.sh
@@ -1,0 +1,4 @@
+#!/bin/bash -e
+
+echo "Set multi-cluster-management feature as dynamic"
+kubectl patch feature multi-cluster-management --type merge --patch  '{"status": {"dynamic": true}}'


ibrokethecloud:
the pre-hook runs before the rancher upgrade, as a result the changes are lost.

this can be verified easily:

* pre-upgrade:
image

* perform upgrade:
image
  the manifest is patched before the rancher helm chart upgrade, which seems to reset the changes

* post-upgrade:
image
  can observe a change in revision number but no change of status

This PR can be closed in favor of: https://github.com/harvester/harvester/pull/3894

bk201:
Fixed in https://github.com/harvester/harvester/pull/3894, thanks @FrankYang0529 !

Issue #3639, opened 2023-03-13
[ENHANCEMENT] Cutting releases on release branches


bk201:
**Is your enhancement related to a problem? Please describe.**

In the supported branches (like v1.1), we add references to other projects.
For example, we [add the installer v1.1 head](https://github.com/harvester/harvester/blob/367ae91de3d69a8db18a5fd63adbc1cbe98aea7a/scripts/build-iso#L10) in `build-iso` script, this ensures we always have ISOs build with v1.1 head installer code. When we do a release, we [change the reference to a tag](https://github.com/harvester/harvester/pull/3635/files#diff-733237cd73ce4fc4e226ab47d23adc621e81b46aefc3a04817c09a683e335246L10), but it's quite easy to forget to switch back to supported branch head later.

**Describe the solution you'd like**

Use release branches to cut releases, not touching the references in supported branches.

**Describe alternatives you've considered**

**Additional context**

Expect some modification to the `.drone.yaml` files.

guangbochen:
Close as this is added via https://github.com/harvester/harvester/tree/release-1.2 and verified by https://github.com/harvester/harvester/releases/tag/v1.2.0-rc3 release.

PR #3865, opened 2023-05-08
build-iso: switch to v1.1 head

bk201:
Still needs to do this before https://github.com/harvester/harvester/issues/3639 is implemented.


guangbochen (approved):
LGTM, thanks

Issue #1931, opened 2022-02-18
[FEATURE] When one volume is added to an running VM, check the real state of it


w13915984028:
**Is your feature request related to a problem? Please describe.**

In Harvester v1.0.0, when an over-sized volume is created independently, it still goes into "ready" state, and it can be added to one already running VM. 

![image](https://user-images.githubusercontent.com/31133476/154658321-370de1d7-4c2b-41cb-82db-d4d8745a4450.png)

**Describe the solution you'd like**

After fix of #1695,
in case volume has abnormal status annotations, prohibit it from being added to the running VM.

For none-running state VM, after adding the abnormal volume, it will cause the VM starts up fail, a warning message to user will be helpful.

Also confirm: when the VM is "taking backup", can in-using volume be resized, or deleted, or new normal volume be added.  

**Describe alternatives you've considered**

**Additional context**

#1346
#1695

w13915984028:
Longhorn is considering enhancement, before it is available, we rely on the fix of #1695 (a temporary solution for now)
https://github.com/longhorn/longhorn/issues/3626

bk201:
To check if PVC can provide more error messages in newer Longhorn.

bk201:
>in case volume has abnormal status annotations, prohibit it from being added to the running VM.

We need to implement this in the GUI and ~~webhook~~ backend action handlers.

WebberHuang1118:
#3869

TachunLin:
Verified **fixed** on `master-b49e264e-head`. Close this issue.

### Result

* When we attach a oversize volume of the default storage class to a running vm, the dashboard will prompt message
`insufficient disk` to prohibit 
  ![Screenshot from 2023-05-22 18-46-02](https://github.com/harvester/harvester/assets/29251855/26be77c8-8667-4719-91db-c90777036a91)
  
  ![image](https://github.com/harvester/harvester/assets/29251855/9bbd9bb8-5a65-4f2d-b0f7-0b97abeb79f5)


*  When we attach a oversize volume of the customized storage class having `volumeBindingMode: WaitForFirstConsumer`
![image](https://github.com/harvester/harvester/assets/29251855/826efff8-8182-4e6f-8810-6117796143d4)

* VM can attach the volume and prompt `insufficient disk` message
  ![image](https://github.com/harvester/harvester/assets/29251855/c748aa11-6c6a-43c4-9256-bc388fbef93b)
  
[...]

PR #3869, opened 2023-05-08
Check volume's real state before adding it to running VM

WebberHuang1118:
**Problem:**
Since LH volume creation and replica scheduling are asynchronous. Even if CreateVolume CSI call success, the replica scheduling still may fail. In such a circumstance, the volume is impossible to be mounted.

**Solution:**
Up-to-date Longhorn supports **[updating PV annotation according to volume schedule condition](https://github.com/longhorn/longhorn-manager/pull/1307)**, we could check if there is replica scheduling error for an LH volume, and prohibit it from being added to the running VM if need

**Related Issue:**
#1931 

**Test plan:**
1. Creating an oversize volume
2. Checking the volume has "insufficient storage" condition Screenshot_oversize_vol
3. Creating a running VM
4. Adding the oversize volume to the running VM
5. The oversize volume should be prohibited to be added Screenshot_insufficient_storage

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/gorilla/mux"
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	"github.com/pkg/errors"
 	"github.com/rancher/apiserver/pkg/apierror"
@@ -64,6 +65,7 @@ type vmActionHandler struct {
 	nadCache                  ctlcniv1.NetworkAttachmentDefinitionCache
 	nodeCache                 ctlcorev1.NodeCache
 	pvcCache                  ctlcorev1.PersistentVolumeClaimCache
+	pvCache                   ctlcorev1.PersistentVolumeCache
 	secretClient              ctlcorev1.SecretClient
 	secretCache               ctlcorev1.SecretCache
 	virtSubresourceRestClient rest.Interface
@@ -838,12 +840,58 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
 	return nil
 }
 
+// Since the LH volume creation and replica scheduling are asynchronous
+// Even the CreateVolume csi call success, the replica scheduling may still fail
+// However for the sc with VolumeBindingMode is Immediate
+// we could check annotation for corresponding error if it exists
+func (h *vmActionHandler) checkAttachable(pvc *corev1.PersistentVolumeClaim) error {
+	// Even Harvester specify default StorageClass,
+	// we check the sc existence for robustness
+	if pvc.Spec.StorageClassName == nil {
+		return fmt.Errorf("volme %v with empty StorageClass is not permitted", pvc.Name)
+	}
+
+	sc, err := h.storageClassCache.Get(*pvc.Spec.StorageClassName)
+	if err != nil {
+		return err
+	}
+
+	if sc.VolumeBindingMode == nil {
+		return fmt.Errorf("volme %v not specify VolumeBindingMode is not permitted", pvc.Name)
+	}
+
+	if *sc.VolumeBindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
+		return nil
+	}
+
+	if pvc.Status.Phase != corev1.ClaimBound {
+		return fmt.Errorf("volme %v not bound yet", pvc.Name)
+	}
+
+	pv, err := h.pvCache.Get(pvc.Spec.VolumeName)
+	if err != nil {
+		return err
+	}
+
+	//Even the volume is in degrade mode, it still could be attached
+	scheduleErrAnno := pv.Annotations[longhorntypes.PVAnnotationLonghornVolumeSchedulingError]
+	if scheduleErrAnno != "" && scheduleErrAnno != longhorn.ErrorReplicaScheduleSchedulingFailed {
+		return fmt.Errorf("volme %v with error %v", pvc.Name, scheduleErrAnno)
+	}
+
+	return nil
+}
+
 // addVolume add a hotplug volume with given volume source and disk name.
 func (h *vmActionHandler) addVolume(ctx context.Context, namespace, name string, input AddVolumeInput) error {
 	// We only permit volume source from existing PersistentVolumeClaim at this moment.
 	// KubeVirt won't check PVC existence so we validate it on our own.
 	pvc, err := h.pvcCache.Get(namespace, input.VolumeSourceName)
 	if err != nil {
+		return err
+	}
+
+	if err := h.checkAttachable(pvc); err != nil {
 		return err
 	}
 
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -18,6 +18,7 @@ import (
 	wranglername "github.com/rancher/wrangler/pkg/name"
 	"github.com/rancher/wrangler/pkg/schemas/validation"
 	"github.com/rancher/wrangler/pkg/slice"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -857,7 +858,8 @@ func (h *vmActionHandler) checkAttachable(pvc *corev1.PersistentVolumeClaim) err
 	}
 
 	if sc.VolumeBindingMode == nil {
-		return fmt.Errorf("volme %v not specify VolumeBindingMode is not permitted", pvc.Name)
+		logrus.Infof("volme %v not specify VolumeBindingMode", pvc.Name)
+		return nil
 	}
 
 	if *sc.VolumeBindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
--- a/pkg/api/vm/schema.go
+++ b/pkg/api/vm/schema.go
@@ -45,6 +45,7 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, options config
 	settings := scaled.HarvesterFactory.Harvesterhci().V1beta1().Setting()
 	nodes := scaled.CoreFactory.Core().V1().Node()
 	pvcs := scaled.CoreFactory.Core().V1().PersistentVolumeClaim()
+	pvs := scaled.CoreFactory.Core().V1().PersistentVolume()
 	secrets := scaled.CoreFactory.Core().V1().Secret()
 	vmt := scaled.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineTemplate()
 	vmtv := scaled.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineTemplateVersion()
@@ -81,6 +82,7 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, options config
 		nadCache:                  nads.Cache(),
 		nodeCache:                 nodes.Cache(),
 		pvcCache:                  pvcs.Cache(),
+		pvCache:                   pvs.Cache(),
 		secretClient:              secrets,
 		secretCache:               secrets.Cache(),
 		virtSubresourceRestClient: virtSubresourceClient,


WebberHuang1118:
#1931

FrankYang0529:
We allow users to create a customized StorageClass with `volumeBindingMode: WaitForFirstConsumer`. If a PVC has no consumer, its status is `Pending` and will not have a corresponding PV. It's better to check StorageClass first before checking PVC status and PV annotation.

Screenshot 2023-05-09 at 1 52 56 PM

WebberHuang1118:
**Solution update**
- For adding PV with StoageClass which has VolumeBindingMode "Immediate" (default)
  -  We check if there is scheduling error for an LH volume, before adding it to a running VM
- For adding PV with StoageClass which has VolumeBindingMode "WaitForFirstConsumer"
  -  Since the PV isn't provided during this phase, the handling works as fix for issue [#1695](https://github.com/harvester/harvester/issues/1695), the PV would be added to VM, even if the storage is insufficient, but users can find the corresponding info from Virtual Machine page

**Test plan update**
- Prerequisites
  - Creating a running VM
- Case1
  - Creating an oversize volume
  - Checking the volume has "insufficient storage" condition Screenshot_oversize_vol
  - Adding the oversize volume to the running VM
  - The oversize volume should be prohibited to be added Screenshot_insufficient_storage
- Case2
  - Creating a customized StorageClass with `volumeBindingMode: WaitForFirstConsumer` customize_sg
  - Creating an oversize volume with customized StorageClass
  - Adding the oversize volume to the running VM
  - The insufficient space message shows up on the VM info page Screenshot_insufficient_cutom-sg

pkg/api/vm/handler.go
```diff
@@ -717,6 +719,37 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
 	return nil
 }
 
+// Since the LH volume creation and replica scheduling are asynchronous
+// Even the CreateVolume csi call success, the replica scheduling may still fail
+// However for the sc with VolumeBindingMode is Immediate
+// we could check annotation for corresponding error if it exists
+func (h *vmActionHandler) checkForConventionSC(pvc *corev1.PersistentVolumeClaim) error {
+	sc, err := h.storageClassCache.Get(*pvc.Spec.StorageClassName)
```
w13915984028, line 855: the `pvc.Spec.StorageClassName` is allowed to be nil, better to check != nil first

pkg/api/vm/handler.go
```diff
@@ -717,6 +719,37 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
...
+	if err != nil {
+		return err
+	}
+
+	if *sc.VolumeBindingMode == storagev1.VolumeBindingWaitForFirstConsumer {
```
w13915984028, line 865: similar

w13915984028 (commented):
robustness enhancement suggestion:

pkg/api/vm/handler.go
WebberHuang1118: Thanks, @w13915984028. I added checking based on your suggestion.

pkg/api/vm/handler.go
WebberHuang1118: Thanks, @w13915984028. I added checking based on your suggestion.

pkg/api/vm/handler.go
```diff
@@ -717,6 +719,43 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
...
+	}
+
+	pv, err := h.pvCache.Get(pvc.Spec.VolumeName)
+	if err != nil {
+		return err
+	}
+
+	scheduleErrAnno := pv.Annotations[longhorntypes.PVAnnotationLonghornVolumeSchedulingError]
+	if scheduleErrAnno == longhorn.ErrorReplicaScheduleInsufficientStorage {
+		return fmt.Errorf("volme %v with insufficient storage", pvc.Name)
```
bk201: Since the annotation generally means that the PVC has an error condition, should we reject the attaching if we see any error here? More than just insufficient storage cases.

pkg/api/vm/handler.go
```diff
@@ -717,6 +719,43 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
 	return nil
 }
 
+// Since the LH volume creation and replica scheduling are asynchronous
+// Even the CreateVolume csi call success, the replica scheduling may still fail
+// However for the sc with VolumeBindingMode is Immediate
+// we could check annotation for corresponding error if it exists
+func (h *vmActionHandler) checkForConventionSC(pvc *corev1.PersistentVolumeClaim) error {
```
Vicente-Cheng: Just curious about the function naming. 
The checking mechanism includes both SC/PV statuses.
The `ConventionSC` would make people understand only SC checking here. Or did I get some misunderstanding here? Thanks!

pkg/api/vm/handler.go
```diff
@@ -717,6 +719,43 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
...
+	sc, err := h.storageClassCache.Get(*pvc.Spec.StorageClassName)
+	if err != nil {
+		return err
+	}
+
+	// Even harvester specify default StorageClass,
+	// we check the sc existence for robustness
+	if sc == nil {
```
Vicente-Cheng: When the sc is nil, most CSI driver will use the default SC.
I checked with your comment above if we assume the sc should not be empty.
We may need to return the error here?

pkg/api/vm/handler.go
WebberHuang1118: I mean check the PVC with conventional StorageClass, maybe the naming is a bit misleading. Do you have any suggestions? How about "checkPVCwithConventionSC"?

pkg/api/vm/handler.go
WebberHuang1118: If the sc is not specified during pvc creation, we would find the sc points to default sc, rather than nil. I added this check for robustness from https://github.com/harvester/harvester/pull/3869#pullrequestreview-1424150708.

Your comment is a fair point. I would change the checking here. Thanks

pkg/api/vm/handler.go
bk201: From the caller's perspective, maybe `checkAttachable` or `verifyAttachable`...

pkg/api/vm/handler.go
Vicente-Cheng: I do some tests here. Looks like we could not have nil StorageClassName, so we should check with an empty string and return an error because we want the sc to exist. Should we?

WebberHuang1118: Agree, I would update the checking per your suggestion.

pkg/api/vm/handler.go
WebberHuang1118: Add rejecting volume attaching if [Longhorn reports any error](https://github.com/harvester/harvester/blob/117e04788b98d2778612c8181cecfcf776564e16/vendor/github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2/node.go#L36), except "replica scheduling failed" which means volume is in degrade mode but still can be attached.

pkg/api/vm/handler.go
WebberHuang1118: Rename function "checkForConventionSC" to "checkAttachable" for better naming.

pkg/api/vm/handler.go
WebberHuang1118: Modify checking: return error if sc is found to be nil.

Vicente-Cheng (approved):
LGTM, thanks!

pkg/api/vm/handler.go
```diff
@@ -717,6 +719,48 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
...
+	if pvc.Spec.StorageClassName == nil {
+		return fmt.Errorf("volme %v with empty StorageClass is not permitted", pvc.Name)
+	}
+
+	sc, err := h.storageClassCache.Get(*pvc.Spec.StorageClassName)
+	if err != nil {
+		return err
+	}
+
+	if sc.VolumeBindingMode == nil {
```
w13915984028, line 860: as `sc.VolumeBindingMode` is not defined and controlled by this controller, would suggest to return nil rather than an error, log warning message instead

upstream may change, if we return error, it will block in certain cases

this feature is `try to alert user`, only return error when everything meets the condition

pkg/api/vm/handler.go
```diff
@@ -717,6 +719,48 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
...
+// Since the LH volume creation and replica scheduling are asynchronous
+// Even the CreateVolume csi call success, the replica scheduling may still fail
+// However for the sc with VolumeBindingMode is Immediate
+// we could check annotation for corresponding error if it exists
+func (h *vmActionHandler) checkAttachable(pvc *corev1.PersistentVolumeClaim) error {
+	// Even Harvester specify default StorageClass,
+	// we check the sc existence for robustness
+	if pvc.Spec.StorageClassName == nil {
+		return fmt.Errorf("volme %v with empty StorageClass is not permitted", pvc.Name)
+	}
```
w13915984028, line 853: similar with L738, prefer skip the checking and log warning, return nil

w13915984028 (commented):
please double check:

bk201 (approved):
lgtm

pkg/api/vm/handler.go
```diff
@@ -717,6 +719,48 @@ func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMac
...
+
+	pv, err := h.pvCache.Get(pvc.Spec.VolumeName)
+	if err != nil {
+		return err
+	}
+
+	//Even the volume is in degrade mode, it still could be attached
+	scheduleErrAnno := pv.Annotations[longhorntypes.PVAnnotationLonghornVolumeSchedulingError]
+	if scheduleErrAnno != "" && scheduleErrAnno != longhorn.ErrorReplicaScheduleSchedulingFailed {
+		return fmt.Errorf("volme %v with error %v", pvc.Name, scheduleErrAnno)
```
bk201, line 881: @LucasSaintarbor Could you help revise the error message? An example is on: https://user-images.githubusercontent.com/19387129/236769070-7024e4c0-55ee-4e3e-bd5d-5a9a166bba1c.png
Thanks!

pkg/api/vm/handler.go
WebberHuang1118: Since during VolumeBindingMode is unset, "Immediate" mode is used by default, https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode. "VolumeBindingMode" never points to nil. How about keeping this checking for robustness as we discussed before? Thanks

pkg/api/vm/handler.go
WebberHuang1118: Since we could not have nil StorageClassName, how about keeping this checking for robustness as in the previous discussion https://github.com/harvester/harvester/pull/3869#discussion_r1194516279 ? Thanks

pkg/api/vm/handler.go
w13915984028: The source code reveals more. 

When a CSI driver sees `VolumeBindingMode` is nil in the SC, it uses `Immediate` as the value. But it does not mean, the SC's `VolumeBindingMode` is never nil. The generated protocol buffer code, also checks !nil to do something.

vendor/k8s.io/api/storage/v1/types.go

```
type StorageClass struct {
...
	// VolumeBindingMode indicates how PersistentVolumeClaims should be
	// provisioned and bound.  When unset, VolumeBindingImmediate is used.
	// This field is only honored by servers that enable the VolumeScheduling feature.
	// +optional
	VolumeBindingMode *VolumeBindingMode `json:"volumeBindingMode,omitempty" protobuf:"bytes,7,opt,name=volumeBindingMode"`
...
```

```
vendor/k8s.io/api/storage/v1/generated.pb.go:	if m.VolumeBindingMode != nil {
..
vendor/k8s.io/api/storage/v1/generated.pb.go:	if m.VolumeBindingMode != nil {
..
```

WebberHuang1118: Update checking for StorageClass `VolumeBindingMode` based on suggestion. If this field point to nil, we just log warning message and return nil. The remaining handling works as fix for issue #1695. Thanks.

WebberHuang1118:
Update checking for StorageClass' VolumeBindingMode based on [suggestion](https://github.com/harvester/harvester/pull/3869#discussion_r1196187403). If this field point to nil, we just log a warning message and return nil. The remaining handling works as fix for issue https://github.com/harvester/harvester/issues/1695. Thanks.

w13915984028 (approved):
LGTM, thanks.

Issue #3600, opened 2023-03-06
[BUG] VMs with CD ROM Container or Container Disk Migrate When Node is in Maintenance Mode


irishgordo:
**Describe the bug**
A VM with a CD ROM Container or Container Disk live-migrates when its Host Node is put into Maintenance Mode

**Expected behavior**
That the VM would not Live-Migrate because of the constraints of:
- with 1 container disk & 1 container cd rom
- with 1 container cd rom
- with 1 container disk

**Environment**
 - Harvester ISO Version: v1.1-a72900af-head (03/01/23, v1.1-head)
 - Underlying Infrastructure: 3-Node Bare metal ThinkCentre Cluster
 
**Additional context**
![Screenshot from 2023-03-02 14-52-08](https://user-images.githubusercontent.com/5370752/223180512-64962d3c-94fe-4f99-b6d1-5b5e72328bd2.png)
![Screenshot from 2023-03-02 14-50-47](https://user-images.githubusercontent.com/5370752/223180519-9e8e821b-6372-481b-8ad2-52d31bb82b42.png)
![Screenshot from 2023-03-02 14-49-48](https://user-images.githubusercontent.com/5370752/223180525-eacd1620-d307-4c4e-8654-c1cf0d61a91d.png)

irishgordo:
Also cross-ref: https://github.com/harvester/harvester/issues/3522

w13915984028:
looks battling :)
the `maintenance mode` will forcely migrate all VMs automatically (?), and the #3522 will block user migration operation if those disks are attached to VM

cc @guangbochen @bk201 @ibrokethecloud

ibrokethecloud:
Added an additional check to query VM's with cd-rom or container disk when putting node in maintenance mode

![image](https://github.com/harvester/harvester/assets/19664014/54b1d875-f3bc-4cf8-a0cd-84f66b592748)

w13915984028:
As a further enhancement, we may consider to support eject CD automatically.

https://github.com/harvester/harvester/issues/3880

irishgordo:
This works great :smile: :+1: !
Thanks for the fix @ibrokethecloud 
Tested on Harvester Version: master-a0aed9f2-head w/ bare-metal single node dell poweredge r720

![Screenshot from 2023-05-22 16-44-55](https://github.com/harvester/harvester/assets/5370752/2340d8a4-5ffb-49c6-a709-7ca44683c096)

ibrokethecloud:
[...]

PR #3879, opened 2023-05-10
List VM's with cdrom and container disks during maintenance mode 

ibrokethecloud:

**Problem:**

Migration of VM's with cdrom or container disk is disabled when a VM specific migration is performed.

image

New maintenance mode capability introduced in 1.1.2 attempts to migrate VM's to other nodes in the cluster.  This only checks for VM's with unhealthy volumes but ignores VM's with cdrom or container disk attachments.

**Solution:**

The PR adds an additional check to the existing `listUnhealthyVM` to list VM's on node which may have a cdrom or container disk.

**Related Issue:**
https://github.com/harvester/harvester/issues/3600

**Test plan:**

To test:
* Create a VM with a cdrom or container disk.
* Once VM is running attempt to place the node VM is scheduled to in maintenance mode.
* This should error with a warning listing VM's which have a cdrom or container disk attached.
* Remove cdrom / container disk from listed VM's and try maintenance mode again.
* This should not be successful, unless the node is hosting last healthy replica for other VM's.

### DIFF
--- a/pkg/api/node/formatter.go
+++ b/pkg/api/node/formatter.go
@@ -188,12 +188,25 @@ func (h ActionHandler) retryMaintenanceModeUpdate(nodeName string, updateFunc ma
 
 func (h ActionHandler) listUnhealthyVM(rw http.ResponseWriter, node *corev1.Node) error {
 	ndc := nodedrain.ActionHelper(h.nodeCache, h.virtualMachineInstanceCache, h.longhornVolumeCache, h.longhornReplicaCache)
-	vmList, err := ndc.FindAndListVM(node)
+
+	var respObj ListUnhealthyVM
+
+	nonMigrtableVMList, err := ndc.FindAndListNonMigratableVM(node)
 	if err != nil {
 		return err
 	}
 
-	var respObj ListUnhealthyVM
+	if len(nonMigrtableVMList) > 0 {
+		respObj.Message = "Please detach CDROM or ContainerDisk from following VMs:"
+		respObj.VMs = nonMigrtableVMList
+		rw.WriteHeader(http.StatusOK)
+		return json.NewEncoder(rw).Encode(&respObj)
+	}
+
+	vmList, err := ndc.FindAndListVM(node)
+	if err != nil {
+		return err
+	}
 
 	if len(vmList) > 0 {
 		respObj.Message = "Following unhealthy VMs will be impacted by the node drain:"
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -249,19 +249,20 @@ func (ndc *ControllerHandler) FindAndListVM(node *corev1.Node) ([]string, error)
 // FindAndListNonMigratableVM is called by action handler to leverage caches to find VM's which may have a cdrom or container disk
 // attached to vmi
 func (ndc *ControllerHandler) FindAndListNonMigratableVM(node *corev1.Node) ([]string, error) {
-	labelSelector, err := labels.Parse(fmt.Sprintf("%s in (%s)", kubevirtv1.NodeNameLabel, node.Name))
-	if err != nil {
-		return nil, err
+	labelsMap := map[string]string{
+		kubevirtv1.NodeNameLabel: node.Name,
 	}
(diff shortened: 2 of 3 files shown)


pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -246,6 +246,27 @@ func (ndc *ControllerHandler) FindAndListVM(node *corev1.Node) ([]string, error)
...
+	if err != nil {
+		return nil, err
+	}
+	vmiList, err := ndc.virtualMachineInstanceCache.List("", labelSelector)
+	if err != nil {
+		return nil, fmt.Errorf("error listing VMI: %v", err)
+	}
+
+	var impactedVMI []string
+	for _, v := range vmiList {
```
futuretea: s/v/vmi/

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -246,6 +246,27 @@ func (ndc *ControllerHandler) FindAndListVM(node *corev1.Node) ([]string, error)
...
 }
 
+// FindAndListNonMigratableVM is called by action handler to leverage caches to find VM's which may have a cdrom or container disk
+// attached to vmi
+func (ndc *ControllerHandler) FindAndListNonMigratableVM(node *corev1.Node) ([]string, error) {
+	labelSelector, err := labels.Parse(fmt.Sprintf("%s in (%s)", kubevirtv1.NodeNameLabel, node.Name))
+	if err != nil {
+		return nil, err
+	}
+	vmiList, err := ndc.virtualMachineInstanceCache.List("", labelSelector)
```
futuretea: use `corev1.NamespaceAll` instead of ""

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -246,6 +246,27 @@ func (ndc *ControllerHandler) FindAndListVM(node *corev1.Node) ([]string, error)
 	return impactedVMDetails, nil
 }
 
+// FindAndListNonMigratableVM is called by action handler to leverage caches to find VM's which may have a cdrom or container disk
+// attached to vmi
+func (ndc *ControllerHandler) FindAndListNonMigratableVM(node *corev1.Node) ([]string, error) {
+	labelSelector, err := labels.Parse(fmt.Sprintf("%s in (%s)", kubevirtv1.NodeNameLabel, node.Name))
```
futuretea: use labels.SelectorFromSet instead of labels.Parse

guangbochen (approved):
LGTM, and u may need to rebase the PR as the upstream master has bumped the LH apiversion, thanks.

futuretea (approved):
LGTM, Thanks

pkg/api/node/formatter.go
```diff
@@ -188,12 +188,25 @@ func (h ActionHandler) retryMaintenanceModeUpdate(nodeName string, updateFunc ma
...
 	if err != nil {
 		return err
 	}
 
-	var respObj ListUnhealthyVM
+	if len(nonMigrtableVMList) > 0 {
+		respObj.Message = "Please detach CDROM or ContainerDisk from following VMs:"
+		respObj.VMs = nonMigrtableVMList
+		rw.WriteHeader(http.StatusOK)
+		return json.NewEncoder(rw).Encode(&respObj)
```
chrisho, line 203: prefer to use `util.ResponseOKWithBody` instead.
https://github.com/ibrokethecloud/harvester/blob/06196d2d2ea33d036e740c025b3ec8d65fbec1ec/pkg/util/http.go#L19-L23

ibrokethecloud: `util.ResponseOKWithBody` absorbs the error during the json Marshalling or http response write. Do we still prefer to use this?

chrisho: I think exposure to response would be better, @guangbochen what do you think?

guangbochen: I think we can keep the current path to expose errors directly since it will be a composite call of this method on the UI, thanks.

chrisho (approved):
LGTM, thanks.

ibrokethecloud:
@Mergifyio backport v1.1

bk201:
@Mergifyio backport v1.1

PR #3884, opened 2023-05-11
context: fix the scaled setup

Vicente-Cheng:

**Problem:**
Code level issue, we need to add the correct component after initialization.

**Solution:**
 We should add the correct component after initialization

**Related Issue:**
None

**Test plan:**
No need


guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM, good spot!

bk201:
@mergifyio backport v1.1

Issue #3133, opened 2022-11-10
[BUG] Timesync config and timesync monitoring missing


Martin-Weiss:
**Describe the bug**
Harvester requires proper timesync and at the moment the NTP server can just be specified during installation and can not changed with a proper user-interface after deployment nor do we have proper timesync status check nor warnings in the harvester UI.

**To Reproduce**
Install with a broken NTP server.
Power of configured NTP server after Harvester deployment.

**Expected behavior**
UI shows timesync status and gives warning when servers are not in sync.
UI settings allow NTP configuration to be modified.

**Environment**
 - Harvester ISO version: 1.1.0

Vicente-Cheng:
The annotation about time sync looks like the following:
annotation name: `harvesterhci.io/ntp-service`
Content:
```
ntpSyncStatus: "disable"/"unsynced"/"synced"
ntpConfigSyncStatus: "unsynced"/"synced"/"unknown"
currentNtpServers: "ntp server string"
wantedNtpServers: "ntp server string"
```
e.g.
```
harvesterhci.io/ntp-service: '{"ntpSyncStatus":"unsynced","ntpConfigSyncStatus":"synced","currentNtpServers":"not.working-2.ntp.org","wantedNtpSer
vers":"not.working-2.ntp.org"}'
```

irishgordo:
**Validation Issues**

@Vicente-Cheng  
I opened up https://github.com/harvester/harvester/issues/4099
As I'm not sure if the errors from `systemd-timesyncd` on the Harvester nodes are intended to bubble up or not?

I was able to validate that on v1.2.0-rc2 that the NTP servers can be set/modified.
![Screenshot from 2023-06-16 11-32-32](https://github.com/harvester/harvester/assets/5370752/9a445b98-3ccc-4e8d-99a8-7e7b1fa294a6)
![Screenshot from 2023-06-16 16-20-08](https://github.com/harvester/harvester/assets/5370752/f3e03a4a-d3a5-4939-9805-4d2b4fd33737)
![Screenshot from 2023-06-16 16-19-02](https://github.com/harvester/harvester/assets/5370752/21b0536f-d7d6-42e3-aef4-bd6753c37b2a)
![Screenshot from 2023-06-16 16-10-51](https://github.com/harvester/harvester/assets/5370752/6603e5e2-1f17-462e-9440-9abc1ccef6b6)
[...]

PR #3891, opened 2023-05-11
settings: add NTP settings validator and controller

Vicente-Cheng:

**Problem:**
We need an easy way to set up the NTP configuration.

**Solution:**
Add new settings for NTP configuration.
That also needs both a validator and controller.

**Related Issue:**
https://github.com/harvester/harvester/issues/3133

**Test plan:**
1. create new CRD [nodeconfigs](https://github.com/harvester/node-manager/pull/10/commits/4fc4d47ddcf4e1a2349d6d968205672619288d23#diff-5270618f5e98c90c08716eeeb09acb482045eeeaaf2021fccdcea5b24f6b5479) which from the related PR https://github.com/harvester/node-manager/pull/10
2. edit the setting `ntp-servers` by kubectl
3. add new value on it, like follows
```
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
kind: Setting
metadata:
  annotations:
    harvesterhci.io/hash: a7359b246565ecf5d02aeef13a09ee9c92c92ce0a93d22ecaae3a7bb
  creationTimestamp: "2023-05-11T09:43:31Z"
  generation: 8
  name: ntp-servers
  resourceVersion: "3145932"
  uid: 55bba6e1-93e1-40b3-9c00-dca9f0169574
status: {}
value: '{"ntpServers":["0.suse.pool.ntp.org","1.suse.pool.ntp.org"]}'
```
4. check the `nodeconfig` be created, and it should be looks like as follows
```
apiVersion: node.harvesterhci.io/v1beta1
kind: NodeConfig
metadata:
  annotations:
    AppliedConfig: '{"ntpServers":"not.working.ntp.org"}'
  creationTimestamp: "2023-06-12T08:28:35Z"
  finalizers:
  - wrangler.cattle.io/harvester-node-config-controller
  generation: 4
  name: harvester-node-0
  namespace: harvester-system
  resourceVersion: "7020623"
  uid: 5457953b-030e-4324-a5de-84740636d1ba
spec:
  ntpConfigs:
    ntpServers: 0.suse.pool.ntp.org 1.suse.pool.ntp.org
status:
  ntpConditions:
  - lastProbeTime: null
    lastTransitionTime: null
    message: NTP config is created or modified, need to wait it to be applied
    reason: NTPConfigModified
    status: "False"
    type: Modified
  - lastProbeTime: null
    lastTransitionTime: null
    message: NTP config is Applied
    reason: NTPConfigApplied
    status: "True"
    type: Applied
```

**Additional Context**
This PR also needs the node-manager controller.
PR: https://github.com/harvester/node-manager/pull/10


pkg/controller/master/setting/ntp_server.go
```diff
@@ -0,0 +1,91 @@
...
+
+const (
+	nodeConfigNS = "harvester-system"
+)
+
+func (h *Handler) syncNTPServer(setting *harvesterv1.Setting) error {
+	logrus.Infof("Processing setting %s, value: %s", setting.Name, setting.Value)
+
+	ntpServers := strings.Replace(setting.Value, ",", " ", -1)
+	if ntpServers == "" {
```
w13915984028: In which scenario, the ntpServers will be empty?

pkg/controller/master/setting/ntp_server.go
```diff
@@ -0,0 +1,91 @@
...
+			logrus.Errorf("Get node config with nodeName %s failure: %v", nodeName, err)
+			return err
+		}
+		logrus.Debugf("Nodeconfig content: %+v, notFound: %v", nodeConfig, errors.IsNotFound(err))
+		if nodeConfig == nil || errors.IsNotFound(err) {
+			nodeConfig := generateNodeConfig(nodeName, ntpServers)
+			_, err := h.nodeConfigs.Create(nodeConfig)
+			if err != nil {
+				logrus.Errorf("Create node config with nodeName %s failure: %v", nodeName, err)
+			}
```
w13915984028, line 50: Looks each NODE will have a related nodeconfig, but `ntp-server` should be a global config, each node should use the same (set) of NTP servers, do we allow config ntp server separately?

Or is it only used to reflect that status of applying in each NODE ?

w13915984028 (commented):
A few questions:
(1) The global level npt-setting will propogate downstream object nodeconfig, if user change nodeconfig and let it be different with global setting, what will happen?
(2) If only about ntp-server, the nodeconfig seems to be redundant with `harvesterv1.Setting`, need more description about this new CRD.

thanks.

w13915984028 (commented):
Another solution will be:

User set values in setting
Node-manager watchs the value, apply it and update the setting's status,  from the setting itself,  the status records the appied/waitingapply in each node.
there will have no data consistency challenge between setting and nodeconfig

pkg/controller/master/setting/ntp_server.go
ibrokethecloud: the harvester installer already allow configuration of a NTP server on a per node basis. Would this change mean that a common NTP server is rolled out to all nodes in the cluster overwriting the one specified during the install?

pkg/controller/master/setting/ntp_server.go
```diff
@@ -0,0 +1,90 @@
...
+		} else {
+			nodeConfigCpy := nodeConfig.DeepCopy()
+			if nodeConfigCpy.Spec.NTPConfig == nil {
+				nodeConfigCpy.Spec.NTPConfig = generateNTPConfig(ntpServers)
+			} else {
+				nodeConfigCpy.Spec.NTPConfig.NTPServers = ntpServers
+				nodeConfigCpy.Spec.NTPConfig.ConfigModified = true
+			}
```
ibrokethecloud: why not make `nodeConfigCpy.Spec.NTPConfig = generateNTPConfig(ntpServers)` the default behaviour?

go.mod
```diff
@@ -7,6 +7,7 @@ replace (
 	github.com/docker/distribution => github.com/docker/distribution v2.8.0+incompatible // oras dep requires a replace is set
 	github.com/docker/docker => github.com/docker/docker v20.10.9+incompatible // oras dep requires a replace is set
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.7.7
+	github.com/harvester/node-manager => github.com/Vicente-Cheng/node-manager v0.0.0-20230511091905-e09f64949031
```
ibrokethecloud: should we switch to the latest tag for `harvester/node-manager`

Vicente-Cheng: Yes, I will switch back once the node-manager PR is merged

pkg/controller/master/setting/ntp_server.go
Vicente-Cheng: I do not see any chance that the `ntpServers` will be empty.
But I am afraid something I do not consider more, so I am adding this check here.

pkg/controller/master/setting/ntp_server.go
Vicente-Cheng: >Looks each NODE will have a related nodeconfig, but ntp-server should be a global config, each node should use the same (set) of NTP servers, do we allow config ntp server separately?
>
>Or is it only used to reflect that status of applying in each NODE ?

The node settings could be created as cluster or node levels. Currently, the NTP setting is only configurable on the cluster level. Not only NTP but other system-level settings could be set up by the node-manager. We keep the capability on node-manager to let it work on the node-level.

So that is why the `ntp-servers` is now cluster level, but we will create it like node-level.

>the harvester installer already allow configuration of a NTP server on a per node basis. Would this change mean that a common NTP server is rolled out to all nodes in the cluster overwriting the one specified during the install?

Yes, that will be overwritten when we set `ntp-server` on UI.

The user will set the NTP server during the installation. After that, he may want to change the `ntp-server`, it will overwrite the previous values.

Vicente-Cheng:
> Node-manager watchs the value, apply it and update the setting's status, from the setting itself, the status records the appied/waitingapply in each node.
> there will have no data consistency challenge between setting and nodeconfig

hi @w13915984028, 
Thanks for your suggestions. I am not sure that the node-manager could change the settings even though he knows the config is modified and different from the settings.
I mean that the design of settings is for user configuration. We may not let anyone controller change it.

WDUT?

BTW, you are correct that the nodeconfig will be inconsistent if the user modifies the nodeconfig CR manually.

pkg/controller/master/setting/ntp_server.go
Vicente-Cheng: That is a good suggestion. I will rewrite this.  Thanks!

pkg/controller/master/setting/ntp_server.go
bk201: After some discussion with @Vicente-Cheng , I'd suggest we adjust the setting to this format (a JSON string):
```json
{
  "ntpServers": [
    "server1",
    "server2"
  ],
  "override": {
    "node1": {
      "ntpServers": [
        "server3",
        "server4"
      ]
    },
    "node2": {
      "ntpServers": [
        "server5"
      ]
    }
  }
}
```

The `override` filed contains a map of node-specific settings. By default, the global `ntpServers`setting is applied on each node unless there is a user-specified overriding.

pkg/webhook/resources/setting/validator.go
```diff
@@ -401,6 +408,70 @@ func (v *settingValidator) customizeTransport() error {
 	return nil
 }
 
+func validateNTPServers(setting *v1beta1.Setting) error {
```
w13915984028, line 414: seems different than harvester-installer,  and the NTP server is not trully validated if it works or not.

what's the consideration?

https://github.com/harvester/harvester-installer/blob/master/pkg/console/util.go#L101

w13915984028 (commented):
In current implementation, the NTP servers are saved to `cfg.SaveTo("/etc/systemd/timesyncd.conf")` in each NODE, there is no CRD objects storing this data.

Not fully sure about the original design, generally speacking, all NODEs in the cluster should use same set of NTP Servers, DNS Servers, it is a challenge for each each NODE to have different NTP, DNS Servers. We do not have a mechanism to sync those to the newly added NODEs in the cluster.

When it comes to this feature, should we consolidate it to a global NTP Server setting, any settings to this object, will be applied to all NODEs in the cluster.

For DNS Server, it will be same, a global setting to apply to all NODEs.

cc @guangbochen @bk201

w13915984028:
Suppose:

For `create` mode NODE, convert the DNS Servers & NTP Servers into `setting` object automatically after installation. 
For `join` mode NODE, they will inheriate the `setting` after installation.
Later, when user configures in the UI, changes are applied to all NODEs. The `reset to default` function in UI will revert the values to the initial ones.

pkg/webhook/resources/setting/validator.go
Vicente-Cheng: Yap, I checked the harvester installer before I implemented the validator. I skip this check because we already have the monitoring mechanism on node-manager. Also, if the NTP server leaves temporarily, that should not pop an error to the user. So for this scenario, I prefer to do the monitoring by `node-manager`.

bk201:
> Suppose:
> 
> 
> 
> For `create` mode NODE, convert the DNS Servers & NTP Servers into `setting` object automatically after installation. 
> 
> For `join` mode NODE, they will inheriate the `setting` after installation.
> 
> Later, when user configures in the UI, changes are applied to all NODEs. The `reset to default` function in UI will revert the values to the initial 

I'd suggest we keep the installation configuration. Specifying ntp servers for both create and join roles is simple enough. And we also make sure nodes' time have a chance to sync before they join a cluster (we have this implementation now). Previously, we saw joining failure because time was not synced between hosts (certificate time issue). 

This also doesn't break existing users' configuration files.

w13915984028:
> > Suppose:
> > For `create` mode NODE, convert the DNS Servers & NTP Servers into `setting` object automatically after installation.
> > For `join` mode NODE, they will inheriate the `setting` after installation.
> > Later, when user configures in the UI, changes are applied to all NODEs. The `reset to default` function in UI will revert the values to the initial
> 
> I'd suggest we keep the installation configuration. Specifying ntp servers for both create and join roles is simple enough. And we also make sure nodes' time have a chance to sync before they join a cluster (we have this implementation now). Previously, we saw joining failure because time was not synced between hosts (certificate time issue).

Not changing them. 

It is that, after the `installation` of first NODE, we will have `npt-server setting` auto generated by controller, from the data user configured in UI.  User can check the setting to review the NTP server later and update them.

For `join` node, it is same, installer will not change. Only after installation, controller will try to sync data with CRD objects from first step.

pkg/webhook/resources/setting/validator.go
w13915984028: If the NTP server IP is wrong, the validator should reject, rather than let `node-manager` report error later.

When validator does not report error, user may assume the configure is successful.

If the server is temporary offline, the `reject` in configuration is still acceptable. User need to wait until the server is up and then reconfigure it.

w13915984028:
Back to user-story, suppose:
```
1. User creates a multi node cluser
2. User plans to update the NTP servers
2.1 User clicks the `setting` NTP servers, he should be able to review the current servers.
2.2 User clicks `edit` to input a new set of NTP servers
2.2.1 some NTP servers IP is wrong, should be denied
2.3 User clicks `save`, the new NTP Servers are applied to all nodes
2.4 If some NODE have error when applying the config, there should be places to check, e.g. k8s event
```

pkg/controller/master/setting/ntp_server.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	nodeConfigNS = "harvester-system"
```
w13915984028: reuse following
```
util/constants.go:      HarvesterSystemNamespaceName      = "harvester-system"
```

pkg/controller/master/setting/ntp_server.go
```diff
@@ -0,0 +1,90 @@
...
+		nodeName := node.Name
+		ntpServers := util.GetNTPServers(ntpSettings)
+
+		nodeConfig, err := h.nodeConfigs.Get(nodeConfigNS, nodeName, metav1.GetOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			logrus.Errorf("Get node config with nodeName %s failure: %v", nodeName, err)
+			return err
+		}
+		logrus.Debugf("Nodeconfig content: %+v, notFound: %v", nodeConfig, errors.IsNotFound(err))
+		newNodeConfig := generateNodeConfig(nodeName, ntpServers)
```
w13915984028: move #L48 after #L49, generate config only when needed

pkg/controller/master/setting/ntp_server.go
```diff
@@ -0,0 +1,90 @@
...
+		if nodeConfig == nil || errors.IsNotFound(err) {
+			_, err := h.nodeConfigs.Create(newNodeConfig)
+			if err != nil {
+				logrus.Errorf("Create node config with nodeName %s failure: %v", nodeName, err)
```
w13915984028, line 48: return err when err != nil

pkg/controller/master/setting/ntp_server.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+		} else {
+			nodeConfigCpy := nodeConfig.DeepCopy()
+			updateNodeConfig(nodeConfigCpy, newNodeConfig)
+			if !reflect.DeepEqual(nodeConfig, nodeConfigCpy) {
+				_, err := h.nodeConfigs.Update(nodeConfigCpy)
+				if err != nil {
+					logrus.Errorf("Update node config with nodeName %s failure: %v", nodeName, err)
```
w13915984028, line 58: return err when err != nil

w13915984028 (commented):
Thanks.

Seems touching those low level settings and files bringing more complexities.

First part, will check the left bunch of codes later.

pkg/controller/master/setting/ntp_server.go
Vicente-Cheng: Updated!, Thanks for the tips.

pkg/controller/master/setting/ntp_server.go
Vicente-Cheng: Good suggestions, updated!

pkg/controller/master/setting/ntp_server.go
Vicente-Cheng: fixed

pkg/controller/master/setting/ntp_server.go
Vicente-Cheng: fixed

pkg/webhook/resources/setting/validator.go
```diff
@@ -404,6 +411,77 @@ func (v *settingValidator) customizeTransport() error {
...
```
ibrokethecloud, line 414: `validateNTPServers` needs to be registered with the settings validation handler at: https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/setting/validator.go#L106-L114

currently this is not being called and I am able to submit incorrect ntp server settings: 

```shell
time="2023-05-28T23:54:52Z" level=info msg="Processing setting ntp-servers, value: 192.168.122.1"
time="2023-05-28T23:54:52Z" level=error msg="error syncing 'ntp-servers': handler harvester-setting-controller: failed to parse NTP settings: invalid character '.' after top-level value, requeuing"
```

ibrokethecloud: please ignore, should have updated the webhook image during testing.

pkg/controller/master/setting/ntp_server.go
Vicente-Cheng: After discussion, we do not support different NTP configs on each node. Instead, we use the global NTP config, and if the nodes cross region, we could set multiple NTP servers to make it works.

Vicente-Cheng:
Just a reminder, this PR needs https://github.com/harvester/node-manager/pull/10 for `go.mod`.
I will update again after the node-manager PR is merged.

bk201 (approved):
LGTM

ibrokethecloud (approved):
lgtm. thanks.

w13915984028 (approved):
LGTM, thanks.

Vicente-Cheng:
> Just a reminder, this PR needs [harvester/node-manager#10](https://github.com/harvester/node-manager/pull/10) for `go.mod`. I will update again after the node-manager PR is merged.

`go.mod` updated

go.mod
```diff
@@ -58,27 +58,29 @@ replace (
...
 	github.com/iancoleman/strcase v0.2.0
 	github.com/k3s-io/helm-controller v0.11.7
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v0.0.0-20200331171230-d50e42f2b669
 	github.com/kubernetes-csi/external-snapshotter/client/v4 v4.2.0
 	github.com/kubernetes/dashboard v1.10.1
-	github.com/longhorn/backupstore v0.0.0-20230505043341-f871cdfdf2fc
-	github.com/longhorn/longhorn-manager v1.4.2
-	github.com/mattn/go-isatty v0.0.14
+	github.com/longhorn/backupstore v0.0.0-20230505020448-631a9fe919eb
+	github.com/longhorn/longhorn-manager v1.4.1
```
guangbochen: should we keep v1.4.2 instead?

guangbochen (approved):
Just a small `go.mod` issue, everything else is LGTM

go.mod
Vicente-Cheng: yap, change back to v1.4.2. Thanks for the reminder!

guangbochen:
@mergify backport v1.2

PR #3916, opened 2023-05-17
update kubelet args for max-pods as part of the upgrade

ibrokethecloud:

This PR is for the upgrade path for issue https://github.com/harvester/harvester/issues/2707

**Problem:**

**Solution:**

**Related Issue:**

**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -566,6 +566,8 @@ command_post_drain() {
   wait_repo
   detect_repo
 
+  # update max-pods to 200 #
+  set_max_pods
   # A post-drain signal from Rancher doesn't mean RKE2 agent/server is already patched and restarted
   # Let's wait until the RKE2 settled.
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -605,6 +615,8 @@ command_single_node_upgrade() {
 
   convert_nodenetwork_to_vlanconfig
 
+  # update max-pods to 200 #
+  set_max_pods
```
bk201: Should we move this before `upgrade_rke2` function?
Putting the call here then the change takes effect after rebooting. In theory there is no difference.

ibrokethecloud: sounds good. makes it better to organize the rke2 changes together.

Vicente-Cheng (approved):
LGTM, thanks!

bk201 (approved):
LGTM!

Issue #3912, opened 2023-05-16
[BUG] VM Clones Don't Copy Over `metadata.annotations.harvesterhci.io/reservedMemory` key or value


irishgordo:
**Describe the bug**
Clone VMs do not copy the original VM's `metadata.annotations.harvesterhci.io/reservedMemory` - the key & value are missing

**To Reproduce**
1. Go to create a vm
1. Provide a reserved memory count for the vm (good for windows vms)
1. Clone the vm
1. Check both `edit as yaml` and also just the `config -> advanced options -> show more`

**Expected behavior**
When cloning the VM, the VM will truly be a clone with a 1:1 from the original VM

**Environment**
 - Harvester ISO Version: master-c9d8825c-head 
 - Underlying Infrastructure: 3 Node HP Proliant Bare Metal Cluster
 
**Additional context**
![Screenshot from 2023-05-16 12-34-17](https://github.com/harvester/harvester/assets/5370752/08e99ce8-4f7e-4207-a2d0-287008f6701b)

WuJun2016:
This is a clone that contains the volume data.
<img width="1608" alt="image" src="https://github.com/harvester/harvester/assets/24985926/ea772dde-ce7c-40ad-8833-d79be71f176c">

lanfon72:
Verified this bug has been fixed.

![image](https://github.com/harvester/harvester/assets/5169694/5dcae6a1-99bd-4d2a-8cb0-5ef8e03178bb)


Test Information
----
* Environment: **qemu/KVM 3 nodes**
* Harvester Version: **master-d739c2a4-head**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Create an image for VM creation
3. Create VM `vm1` with **Reserved Memory** (in advanced options) added
4. Clone `vm1` into `vm2` with volume data
5. `vm2` should also have **Reserved Memory** configured and the value should be the same as `vm1`

PR #3925, opened 2023-05-17
Keep source reserved memory after vm clone

futuretea:

Keep source reserved memory after vm clone

**Related Issue:**
https://github.com/harvester/harvester/issues/3912

**Test plan:**

1. Go to create a vm
1. Provide a reserved memory count for the vm (good for windows vms)
1. Clone the vm (clone volume data mode)
1. Check both `edit as yaml` and also just the `config -> advanced options -> show more`

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -47,6 +47,12 @@ const (
 	sshAnnotation = "harvesterhci.io/sshNames"
 )
 
+var (
+	cloneVMAnnotationKeys = []string{
+		util.AnnotationReservedMemory,
+	}
+)
+
 type vmActionHandler struct {
 	namespace                 string
 	vms                       ctlkubevirtv1.VirtualMachineClient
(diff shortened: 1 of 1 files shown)


pkg/api/vm/handler.go
```diff
@@ -1244,6 +1244,9 @@ func getClonedVMYamlFromSourceVM(newVMName string, sourceVM *kubevirtv1.VirtualM
 		},
 		Spec: *sourceVM.Spec.DeepCopy(),
 	}
+	if sourceReservedMemory := sourceVM.Annotations[util.AnnotationReservedMemory]; sourceReservedMemory != "" {
+		newVM.Annotations[util.AnnotationReservedMemory] = sourceReservedMemory
+	}
```
guangbochen: let's define a constant array key and loop to add if they exist for better extendability, thanks.

futuretea: updated, add a slice var ([]string cannot be a const)

guangbochen (approved):
LGTM, thanks.

bk201 (approved):

bk201:
I guess we also need to backport.

futuretea:
@mergify backport v1.1

Issue #3747, opened 2023-04-04
[ENHANCEMENT] component upgrade of Harvester v1.2


guangbochen:
**Is your enhancement related to a problem? Please describe.**
Bump Harvester components to newer versions.

**Describe the solution you'd like**
Will need to address the following ones:
- [x] OS @Vicente-Cheng 
   - [x] https://github.com/harvester/harvester-installer/pull/490
- [x] KubeVirt v0.54.0(with hot-plugin & monitoring backport fix) @guangbochen 
  - [x] https://github.com/harvester/harvester/pull/3910
  - [x] https://github.com/harvester/harvester/pull/4351
- [x] Longhorn to v1.4.3 @guangbochen 
  - [x] https://github.com/harvester/harvester/pull/3910
  - [x] need to address https://github.com/harvester/harvester/issues/3943 via LH v1.4.3(target on 07/06)
- [x] RKE2 v1.25.9 May Release @bk201 
  - [x] https://github.com/harvester/harvester-installer/pull/491
  - [x] Upgrade path need to turn of Rancher chart PSP (seems not needed).
- [x] Rancher to v2.7.5 @bk201
- [x] verify Rancher v2.7.6 upgrade for both MCM and embedded Rancher

TBD:
- Bump Golang to v1.19 @guangbochen 

~~- [ ] Bump k8s dependency to v1.25.9 @guangbochen - k8s client not compatible after upgrade, need to bump kubevirt dep in the future version~~ 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

guangbochen:
Close as all components upgrade PRs are merged.

PR #3963, opened 2023-05-24
Disable some RKE2 charts

bk201:
**Problem:**

These charts are managed as Harvester dependency charts. Thus we don't want RKE2 to deploy them.

- rke2-snapshot-controller
- rke2-snapshot-controller-crd
- rke2-snapshot-validation-webhook

**Solution:**

Disable the following charts:
- rke2-snapshot-controller
- rke2-snapshot-controller-crd
- rke2-snapshot-validation-webhook

There are two situations RKE2 might restart and load configuration files:
- A node is promoted.
- A node is upgraded from the previous version.

Add proper configuration files in these two cases.

**Related Issue:**

https://github.com/harvester/harvester/issues/3747

**Test plan:**

### DIFF
--- a/deploy/charts/harvester/templates/configmap.yaml
+++ b/deploy/charts/harvester/templates/configmap.yaml
@@ -69,6 +69,15 @@ data:
     audit-policy-file: /etc/rancher/rke2/config.yaml.d/92-harvester-kube-audit-policy.yaml
     EOF
 
+    # Disable snapshot-controller related charts because we manage them in Harvester.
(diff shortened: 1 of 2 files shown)


futuretea (approved):
LGTM, Thanks

guangbochen (approved):
LGTM, single node upgrade works, and the rke2 snapshot controllers are not enabled after the upgrade.
image

Vicente-Cheng (approved):
Test this PR with https://github.com/harvester/harvester-installer/pull/491.
Upgraded from v1.1.2, verify with basic operations.
Also, creating the downstream cluster and PVC works well.
LGTM, thanks!

PR #3965, opened 2023-05-25
added a handler to patch apps.catalog object if it corresponds to addon

ibrokethecloud:

**Problem:**

When mcm mode is enabled in rancher the users can browse to apps catalog on the local cluster and edit / update existing harvester integrations like, monitoring, logging, and any optional addons.

These should only be upgraded as part of the harvester upgrade.

**Solution:**

PR introduces additional logic in addon controller to perform the following:
* patch apps.catalog object corresponding to addons to ensure they are reported as "managed" by rancher, which should prevent a user to update said chart from rancher marketplace
* patch predefined list of apps as "managed" for the same effect. For example, `rancher` deployment is marked managed to avoid it from being upgraded from the app store.
**Related Issue:**
https://github.com/rancher/dashboard/issues/8800
**Test plan:**

### DIFF
--- a/pkg/controller/master/addon/addon.go
+++ b/pkg/controller/master/addon/addon.go
@@ -4,10 +4,14 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"reflect"
 	"strings"
 
 	helmv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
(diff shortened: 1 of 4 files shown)


guangbochen (approved):
LGTM, thanks.

bk201 (approved):
lgtm!

Issue #1723, opened 2021-12-21
[Enhancement] Data sync is required before taking backup of a VM


khushboo-rancher:
**Describe the bug**
Data sync should happen before taking backup of a vm. Sometimes the data of backup miss content from vm because the data is still in buffer and haven't written on the disk. 

**To Reproduce**
1. Create a vm.
2. Write multiple files in a loop.
3. Take backup
4. Restore the backup, verify not all the content of the files which are written in the last are there in the restored data.

**Expected behavior**
Sync should be done before taking the backup so that all the data can be backed up.

**Environment:**
 - Harvester ISO version:  v1.0.0-rc2

futuretea:
link to https://github.com/harvester/harvester/issues/2279

rodrigo14miguel:
This might need volume snapshot #553 #2294. Backup solutions normally use snapshots for either k8s or hypervisors.

albinsun:
Can reproduce in `v1.1.2` with following steps. Will verify fix in `v1.2.0-rc2+`.

## Verification Steps
1. Create a VM.
2. Open console and create 100 txt files using loop.
4. Take backup
5. Restore backup (Replace Existing)
6. Files created in step 2 should still exist. :x:

[repro-112-render1.webm](https://github.com/harvester/harvester/assets/2773781/9e623b36-6d41-441a-840e-178570828996)

albinsun:
Tested in `v1.2.0-rc2`, verified as fixed.

## Test Environment
* Harvester
  * version: `v1.2.0-rc2`
  * profile: KVM, single node
  * **ui-source**: Auto

## Verification Steps
### Scenario: Replace Existing
1. Create a VM.
2. Open console and create 100 txt files using loop.
3. Take backup
4. Restore backup (Replace Existing)
5. Files created in step 2 should still exist. :heavy_check_mark:
### Scenario: Restore New
1. Create a VM.
2. Open console and create 100 txt files using loop.
3. Take backup
4. Restore backup (Restore New)
5. Files created in step 2 should still exist. :heavy_check_mark:

[repro-120rc2-render.webm](https://github.com/harvester/harvester/assets/2773781/db52020e-de7b-4631-aae2-778b81abe4c2)

PR #3976, opened 2023-05-25
snapshot/backup: freeze guest fs if possible

connorkuehl:

**Problem:**

Snapshots and backups are currently "crash-consistent" (i.e., the snapshots contain whatever the guest's disk contains at the time of the snapshot as if the power cable was unplugged.)

This means that the snapshot doesn't include any I/O that the guest's kernel hasn't flushed to disk, so there's a possibility that a user might take a snapshot but due to their guest's I/O patterns the snapshot might not contain exactly what the user expected it to at the time the snapshot was issued.

**Solution:**

In order to take more consistent snapshots, check to see if the QEMU guest agent is running in the guest at the time of snapshot. If it is, send a freeze-fs message to sync the filesystems prior to the snapshot to improve the consistency of snapshots/backups in Harvester.

**Related Issue:** https://github.com/harvester/harvester/issues/1723

**Test plan:**

Test case 1: ensure VM backup/snapshot controller is not regressed

1. In the VM: Write some data to a file (`echo 123 > test.txt && sync`)
2. Create a VM Snapshot
3. Expected result: VM Snapshot succeeds

Test case 2: guest agent, nontrivial amount of data, produces a consistent snapshot

1. In the VM: install and/or enable qemu-guest-agent
2. In the VM: Write a nontrivial amount of data to a file (`dd if=/dev/urandom of=test.txt bs=1M count=1000 status=progress`)
4. Create a VM Snapshot
5. In the VM: Measure `test.txt` (`sha256sum test.txt`). Keep note of this output for asserting at end of test case.
6. Restore into new VM.
7. In the new VM: Measure `test.txt` (`sha256sum test.txt`).
8. Expected result: SHAs are the same
9. Expected result: In the original VM: qemu-guest-agent reports freezing file systems (`sudo journalctl -u qemu-guest-agent | grep -q "guest-fsfreeze called" && echo ok || echo fail`) -- should print "ok"

Test case 3: no guest agent, nontrivial amount of data likely produces inconsistent snapshot

1. In the VM: disable and/or uninstall qemu-guest-agent
2. In the VM: Write a nontrivial amount of data to a file (`dd if=/dev/urandom of=test.txt bs=1M count=1000 status=progress`)
3. Create a VM Snapshot
4. In the VM: Measure `test.txt` (`sha256sum test.txt`). Keep note of this output for asserting at end of test case.
5. Restore into new VM.
6. In the new VM: Measure `test.txt` (`sha256sum test.txt`).
7. Expected result: One of: [SHAs match (unlikely), SHAs are different (likely), new VM test.txt does not exist (possible)] -- it's possible the amount of data written in the test case will need to be increased if this test case does not reproduce an inconsistent snapshot.

---

Notes: in order to best reproduce this, I have had to automate SSH'ing into the guest, running the `dd` command and then apply a VirtualMachineBackup type: snapshot with kubectl to ensure the snapshot is taken before all of the guest's pending I/O flushes to disk. It's far superior to my slow human reflexes lol

### DIFF
--- a/pkg/controller/master/backup/backup.go
+++ b/pkg/controller/master/backup/backup.go
@@ -9,10 +9,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"time"
 
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
@@ -29,12 +31,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sschema "k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/config"
+	"github.com/harvester/harvester/pkg/generated/clientset/versioned/scheme"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
 	ctllonghornv2 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
@@ -55,6 +60,8 @@ const (
 	backupTargetAnnotation       = "backup.harvesterhci.io/backup-target"
 	backupBucketNameAnnotation   = "backup.harvesterhci.io/bucket-name"
 	backupBucketRegionAnnotation = "backup.harvesterhci.io/bucket-region"
+
+	defaultFreezeDuration = 1 * time.Second
 )
 
 var vmBackupKind = harvesterv1.SchemeGroupVersion.WithKind(vmBackupKindName)
@@ -67,31 +74,44 @@ func RegisterBackup(ctx context.Context, management *config.Management, opts con
 	secrets := management.CoreFactory.Core().V1().Secret()
 	storageClasses := management.StorageFactory.Storage().V1().StorageClass()
 	vms := management.VirtFactory.Kubevirt().V1().VirtualMachine()
+	vmis := management.VirtFactory.Kubevirt().V1().VirtualMachineInstance()
 	volumes := management.LonghornFactory.Longhorn().V1beta2().Volume()
 	lhbackups := management.LonghornFactory.Longhorn().V1beta2().Backup()
 	snapshots := management.SnapshotFactory.Snapshot().V1().VolumeSnapshot()
 	snapshotContents := management.SnapshotFactory.Snapshot().V1().VolumeSnapshotContent()
 	snapshotClass := management.SnapshotFactory.Snapshot().V1().VolumeSnapshotClass()
 
+	virtSubsrcConfig := rest.CopyConfig(management.RestConfig)
+	virtSubsrcConfig.GroupVersion = &k8sschema.GroupVersion{Group: "subresources.kubevirt.io", Version: "v1"}
+	virtSubsrcConfig.APIPath = "/apis"
+	virtSubsrcConfig.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
+	virtSubresourceClient, err := rest.RESTClientFor(virtSubsrcConfig)
+	if err != nil {
+		return err
+	}
+
 	vmBackupController := &Handler{
-		vmBackups:            vmBackups,
-		vmBackupController:   vmBackups,
-		vmBackupCache:        vmBackups.Cache(),
-		pvCache:              pv.Cache(),
-		pvcCache:             pvc.Cache(),
-		secretCache:          secrets.Cache(),
-		storageClassCache:    storageClasses.Cache(),
-		vms:                  vms,
-		vmsCache:             vms.Cache(),
-		volumeCache:          volumes.Cache(),
-		volumes:              volumes,
-		lhbackupCache:        lhbackups.Cache(),
-		snapshots:            snapshots,
-		snapshotCache:        snapshots.Cache(),
-		snapshotContents:     snapshotContents,
-		snapshotContentCache: snapshotContents.Cache(),
-		snapshotClassCache:   snapshotClass.Cache(),
-		recorder:             management.NewRecorder(backupControllerName, "", ""),
+		vmBackups:                 vmBackups,
+		vmBackupController:        vmBackups,
+		vmBackupCache:             vmBackups.Cache(),
+		pvCache:                   pv.Cache(),
+		pvcCache:                  pvc.Cache(),
+		secretCache:               secrets.Cache(),
+		storageClassCache:         storageClasses.Cache(),
+		vms:                       vms,
+		vmsCache:                  vms.Cache(),
+		vmis:                      vmis,
+		vmisCache:                 vmis.Cache(),
+		volumeCache:               volumes.Cache(),
+		volumes:                   volumes,
+		lhbackupCache:             lhbackups.Cache(),
+		snapshots:                 snapshots,
+		snapshotCache:             snapshots.Cache(),
+		snapshotContents:          snapshotContents,
+		snapshotContentCache:      snapshotContents.Cache(),
+		snapshotClassCache:        snapshotClass.Cache(),
+		virtSubresourceRestClient: virtSubresourceClient,
+		recorder:                  management.NewRecorder(backupControllerName, "", ""),
 	}
 
 	vmBackups.OnChange(ctx, backupControllerName, vmBackupController.OnBackupChange)
@@ -102,24 +122,27 @@ func RegisterBackup(ctx context.Context, management *config.Management, opts con
 }
 
 type Handler struct {
-	vmBackups            ctlharvesterv1.VirtualMachineBackupClient
-	vmBackupCache        ctlharvesterv1.VirtualMachineBackupCache
-	vmBackupController   ctlharvesterv1.VirtualMachineBackupController
-	vms                  ctlkubevirtv1.VirtualMachineClient
-	vmsCache             ctlkubevirtv1.VirtualMachineCache
-	pvCache              ctlcorev1.PersistentVolumeCache
-	pvcCache             ctlcorev1.PersistentVolumeClaimCache
-	secretCache          ctlcorev1.SecretCache
-	storageClassCache    ctlstoragev1.StorageClassCache
-	volumeCache          ctllonghornv2.VolumeCache
-	volumes              ctllonghornv2.VolumeClient
-	lhbackupCache        ctllonghornv2.BackupCache
-	snapshots            ctlsnapshotv1.VolumeSnapshotClient
-	snapshotCache        ctlsnapshotv1.VolumeSnapshotCache
-	snapshotContents     ctlsnapshotv1.VolumeSnapshotContentClient
-	snapshotContentCache ctlsnapshotv1.VolumeSnapshotContentCache
-	snapshotClassCache   ctlsnapshotv1.VolumeSnapshotClassCache
-	recorder             record.EventRecorder
+	vmBackups                 ctlharvesterv1.VirtualMachineBackupClient
+	vmBackupCache             ctlharvesterv1.VirtualMachineBackupCache
+	vmBackupController        ctlharvesterv1.VirtualMachineBackupController
+	vms                       ctlkubevirtv1.VirtualMachineClient
+	vmsCache                  ctlkubevirtv1.VirtualMachineCache
+	vmis                      ctlkubevirtv1.VirtualMachineInstanceClient
+	vmisCache                 ctlkubevirtv1.VirtualMachineInstanceCache
+	pvCache                   ctlcorev1.PersistentVolumeCache
+	pvcCache                  ctlcorev1.PersistentVolumeClaimCache
+	secretCache               ctlcorev1.SecretCache
+	storageClassCache         ctlstoragev1.StorageClassCache
+	volumeCache               ctllonghornv2.VolumeCache
+	volumes                   ctllonghornv2.VolumeClient
+	lhbackupCache             ctllonghornv2.BackupCache
+	snapshots                 ctlsnapshotv1.VolumeSnapshotClient
+	snapshotCache             ctlsnapshotv1.VolumeSnapshotCache
+	snapshotContents          ctlsnapshotv1.VolumeSnapshotContentClient
+	snapshotContentCache      ctlsnapshotv1.VolumeSnapshotContentCache
+	snapshotClassCache        ctlsnapshotv1.VolumeSnapshotClassCache
+	virtSubresourceRestClient rest.Interface
+	recorder                  record.EventRecorder
 }
 
 // OnBackupChange handles vm backup object on change and reconcile vm backup status
@@ -224,6 +247,112 @@ func (h *Handler) getBackupSource(vmBackup *harvesterv1.VirtualMachineBackup) (*
 	}
 
 	return sourceVM, nil
+}
+
+func (h *Handler) getBackupSourceInstance(vmBackup *harvesterv1.VirtualMachineBackup) (*kubevirtv1.VirtualMachineInstance, error) {
+	if vmBackup.Spec.Source.Kind != kubevirtv1.VirtualMachineGroupVersionKind.Kind {
+		return nil, fmt.Errorf("unsupported source: %+v", vmBackup.Spec.Source)
+	}
+
+	sourceVMI, err := h.vmisCache.Get(vmBackup.Namespace, vmBackup.Spec.Source.Name)
+	if err != nil {
+		return nil, err
+	}
+	if sourceVMI.DeletionTimestamp != nil {
+		return nil, fmt.Errorf("vm %s/%s is being deleted", vmBackup.Namespace, vmBackup.Spec.Source.Name)
+	}
+
+	return sourceVMI, nil
+}
+
+func (h *Handler) tryFreezeFS(backup *harvesterv1.VirtualMachineBackup) error {
+	backup, err := h.withFreezeFSAnnotation(backup)
+	if err != nil {
+		return err
+	}
+
+	freezeText := backup.Annotations[util.AnnotationSnapshotFreezeFS]
+	freeze, err := strconv.ParseBool(freezeText)
+	if err != nil {
+		return fmt.Errorf("annotation %q=%q, should be \"true\" or \"false\"", util.AnnotationSnapshotFreezeFS, freezeText)
+	}
+
+	if !freeze {
+		return nil
+	}
+
+	sourceVMI, err := h.getBackupSourceInstance(backup)
+	if apierrors.IsNotFound(err) {
+		return errors.New("virtual machine must be running for qemu-guest-agent-assisted snapshot/backup")
+	}
+	if err != nil {
+		return err
+	}
+
+	err = h.freezeFS(context.Background(), sourceVMI, defaultFreezeDuration)
+	if err != nil {
+		logrus.WithError(err).Warn("qemu-guest-agent fs-freeze")
+		return errors.New("failed to freeze file system")
+	}
+
+	return nil
+}
+
+func (h *Handler) withFreezeFSAnnotation(backup *harvesterv1.VirtualMachineBackup) (*harvesterv1.VirtualMachineBackup, error) {
+	// Annotation already exists with valid bool text
+	val, ok := backup.Annotations[util.AnnotationSnapshotFreezeFS]
+	if ok && (val == "true" || val == "false") {
+		return backup, nil
+	}
+
+	sourceVMI, err := h.getBackupSourceInstance(backup)
+	if apierrors.IsNotFound(err) {
+		return h.saveFreezeFSAnnotation(backup, false)
+	}
+	if err != nil {
+		return backup, err
+	}
+
+	hasAgent := false
+
+	for _, condition := range sourceVMI.Status.Conditions {
+		if condition.Type == kubevirtv1.VirtualMachineInstanceAgentConnected {
+			hasAgent = condition.Status == corev1.ConditionTrue
+			break
+		}
+	}
+
+	return h.saveFreezeFSAnnotation(backup, hasAgent)
+}
+
+func (h *Handler) saveFreezeFSAnnotation(backup *harvesterv1.VirtualMachineBackup, value bool) (*harvesterv1.VirtualMachineBackup, error) {
+	backupCopy := backup.DeepCopy()
+
+	if backupCopy.Annotations == nil {
+		backupCopy.Annotations = make(map[string]string)
+	}
+
+	backupCopy.Annotations[util.AnnotationSnapshotFreezeFS] = strconv.FormatBool(value)
+
+	return h.vmBackups.Update(backupCopy)
+}
+
+func (h *Handler) freezeFS(ctx context.Context, vmi *kubevirtv1.VirtualMachineInstance, timeout time.Duration) error {
+	body, err := json.Marshal(kubevirtv1.FreezeUnfreezeTimeout{UnfreezeTimeout: &metav1.Duration{
+		Duration: timeout,
+	}})
+	if err != nil {
+		return err
+	}
+
+	res := h.virtSubresourceRestClient.Put().
+		Namespace(vmi.Namespace).
+		Resource("virtualmachineinstances").
+		Name(vmi.Name).
+		SubResource("freeze").
+		Body(body).
+		Do(ctx)
+	return res.Error()
 }
 
 // getVolumeBackups helps to build a list of VolumeBackup upon the volume list of backup VM
@@ -462,6 +591,11 @@ func (h *Handler) reconcileVolumeSnapshots(vmBackup *harvesterv1.VirtualMachineB
 		}
 
 		if volumeSnapshot == nil {
+			err := h.tryFreezeFS(vmBackupCpy)
+			if err != nil {
+				return err
+			}
+
 			volumeSnapshotClass := csiDriverVolumeSnapshotClassMap[volumeBackup.CSIDriverName]
 			volumeSnapshot, err = h.createVolumeSnapshot(vmBackupCpy, volumeBackup, &volumeSnapshotClass)
 			if err != nil {
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -14,6 +14,7 @@ const (
 	AnnotationReservedMemory       = prefix + "/reservedMemory"
 	AnnotationHash                 = prefix + "/hash"
 	AnnotationRunStrategy          = prefix + "/vmRunStrategy"
+	AnnotationSnapshotFreezeFS     = prefix + "/snapshotFreezeFS"
 	LabelImageDisplayName          = prefix + "/imageDisplayName"
 	LabelVMName                    = prefix + "/vmName"
 


ibrokethecloud (approved):
thanks for the PR.

LGTM.

markhillgit (approved):
looks good

pkg/controller/master/backup/backup.go
```diff
@@ -226,6 +247,65 @@ func (h *Handler) getBackupSource(vmBackup *harvesterv1.VirtualMachineBackup) (*
...
+			hasAgent = condition.Status == corev1.ConditionTrue
+			break
+		}
+	}
+
+	if !hasAgent {
+		return
+	}
+
+	if err := h.freezeFS(context.Background(), sourceVMI, defaultFreezeDuration); err != nil {
```
WebberHuang1118: May I have two questions,
1.  Can we guarantee one second is long enough for all CSI drivers (Longhorn currently) to take a snapshot? Since it may be related to complex operations, as Longhorn for example, the driver has to [create a new layer for replicas and update read indexes on all nodes ](https://longhorn.io/docs/1.4.2/concepts/#241-how-snapshots-work)
2. As stated above, if taking a snapshot requires longer than we expected, filesystem I/O could be back to work before snapshot completion, and it may bring unexpected behaviour. Should we consider freezing vm filesystem before taking a snapshot, and unfreezing it after the snapshot, like [KubeVirt's handling for snapshot](https://kubevirt.io/user-guide/operations/snapshot_restore_api/#snapshot-a-virtualmachine)? 

Thanks

w13915984028: FYI:
I run Harvester HOST on KVM VM, and take vm snapshort/restore frequently in Virt-manager, to switch between diffrent Harvester versions, for such scale of VM (10 core, 20G memory, 200G disk), even in really fast PC, it still takes a long time to finish a snapshort, inbetween that, the VM is shown as suspend/freezed.

Back to here, seems difficult to set an one-fit-all timeout parameter, we may need to calculate it per VM scale and experience.

Or even expose the timeout as a parameter to `take snapshort`, give an initial estimated value, and user can further change it.

https://bugzilla.redhat.com/show_bug.cgi?id=1989263
https://github.com/kubevirt/kubevirt/pull/6219

w13915984028 (commented):
Another question is about LH, need to double check. 

LH diable file cache by default. But as a distributed filesystem, the ongoing activity could be replica writing, file syncing, the not finishied UNIX socket data processing ....

From the backend interface of LH, there seems no `freeze/suspend` related interface.  When `freeze` happens, how will LH be notified and finish the related work ?

@connorkuehl Do you know which VM/OS/filesystem level API is called per `freeze`.

@innobead David, could you help double confirm, thanks.

```
type Backend interface {
    ReaderWriterAt
    io.Closer
    Snapshot(name string, userCreated bool, created string, labels map[string]string) error
    Expand(size int64) error
    Size() (int64, error)
    SectorSize() (int64, error)
    RemainSnapshots() (int, error)
    GetRevisionCounter() (int64, error)
    SetRevisionCounter(counter int64) error
    GetMonitorChannel() MonitorChannel
    StopMonitoring()
    IsRevisionCounterDisabled() (bool, error)
    GetLastModifyTime() (int64, error)
    GetHeadFileSize() (int64, error)
}
```

pkg/controller/master/backup/backup.go
w13915984028: image

```
$ date
Wed 31 May 2023 03:40:41 PM CEST

$ date
Wed 31 May 2023 03:41:37 PM CEST
```

Did a test, this snapshot took about one minute. The VM is shown as `paused` in virt-manager.

connorkuehl:
> Do you know which VM/OS/filesystem level API is called per freeze.

@w13915984028 The qemu-guest-agent calls[1] the `fsfreeze` ioctl[2], which defers to the filesystem-specific handler for fs freeze, if supported.

[1] https://gitlab.com/qemu-project/qemu/-/blob/master/qga/commands-linux.c#L220
[2] https://github.com/torvalds/linux/blob/48b1320a674e1ff5de2fad8606bee38f724594dc/fs/ioctl.c#L386

w13915984028:
> > Do you know which VM/OS/filesystem level API is called per freeze.
> 
> @w13915984028 The qemu-guest-agent calls[1] the `fsfreeze` ioctl[2], which defers to the filesystem-specific handler for fs freeze, if supported.
> 
> [1] https://gitlab.com/qemu-project/qemu/-/blob/master/qga/commands-linux.c#L220 [2] https://github.com/torvalds/linux/blob/48b1320a674e1ff5de2fad8606bee38f724594dc/fs/ioctl.c#L386

Got it, fs will try to freeze if it can, and LH is the low level block dev, when fs stops writing, LH/anyother will have higher chance to sync data.

Do we care about the potential failure from low level, and, will `un-freeze` (``qmp_guest_fsfreeze_thaw` -> qmp_guest_fsfreeze_do_thaw`) be called explicitly?

pkg/controller/master/backup/backup.go
connorkuehl: > Can we guarantee one second is long enough for all CSI drivers (Longhorn currently) to take a snapshot? Since it may be related to complex operations, as Longhorn for example, the driver has to [create a new layer for replicas and update read indexes on all nodes ](https://longhorn.io/docs/1.4.2/concepts/#241-how-snapshots-work)

No, I don't think so.

I think the ideal freeze duration is not the duration of the snapshot, but instead however long it takes to flush all I/O that was pending at the time the user executed the snapshot. I believe that duration is much, much smaller than the duration of a whole snapshot. Maybe 1 second is too low, but the duration of the snapshot would be the worst case, IMO. I'll comment more on this below.

> As stated above, if taking a snapshot requires longer than we expected, filesystem I/O could be back to work before snapshot completion, and it may bring unexpected behaviour. 

I think this is also true for how snapshots/backups in Harvester work today.

I don't know anything about Longhorn internals, but ideally during a snapshot, it would be redirecting new writes from the guest to some sort of overlay that can be merged down to the original disk after the snapshot completes. :thinking: If this isn't how it behaves for snapshots then indeed this complicates things.

> Should we consider freezing vm filesystem before taking a snapshot, and unfreezing it after the snapshot, like [KubeVirt's handling for snapshot](https://kubevirt.io/user-guide/operations/snapshot_restore_api/#snapshot-a-virtualmachine)?

I would prefer to freeze for exactly how long it takes to flush pending I/O to disk at time of snapshot vs freezing for the duration of the snapshot, which I suspect will take much longer than flushing the pending I/O.

I might be misunderstanding how the FS freezing mechanism works, but if the file system is frozen, then none of the guest's writes are making it to durable storage. So, for the window that the FS is frozen, the risk of dataloss is greater in case of an unexpected shutdown from the guest, which is why I'd prefer to freeze for as small of a duration as possible vs the duration of the entire snapshot.

That said, freezing the FS for exactly the amount of time it takes to flush pending I/O is a very hand-wavey thing to say, and probably the best thing we can do is make it tunable.

---

> I run Harvester HOST on KVM VM, and take vm snapshort/restore frequently in Virt-manager, to switch between diffrent Harvester versions, for such scale of VM (10 core, 20G memory, 200G disk), even in really fast PC, it still takes a long time to finish a snapshort, inbetween that, the VM is shown as suspend/freezed.

Yeah, I think in this case the "suspend/freeze" isn't referring to the FS being frozen, but instead libvirt pausing the domain to take a full system snapshot (cpu registers, RAM, disk), resulting in downtime. The alternative to this is having libvirt do an external disk snapshot where it just snapshots the disk while it redirects new writes from the live guest to a scratch file and then merges that back into the original disk once the snapshot completes. (Sorry if none of this is new info, I just want to be careful where I use the word "freeze" in this PR so I can be clear which mechanism I'm referring to).

Using the qemu-guest-agent in Harvester as shown won't pause the VM, so the VM should not experience downtime in that way.

>Back to here, seems difficult to set an one-fit-all timeout parameter, we may need to calculate it per VM scale and experience.
> Or even expose the timeout as a parameter to take snapshort, give an initial estimated value, and user can further change it.

I think exposing the parameter is probably the most user-friendly thing here. We can have a conservative default of a few seconds, but users can choose to extend this if they want, based on the size of their VM or how I/O intensive their workload is.

Is this PR blocked on that parameter?

connorkuehl:
> Got it, fs will try to freeze if it can, and LH is the low level block dev, when fs stops writing, LH/anyother will have higher chance to sync data.

Right, that's my understanding as well. The FS gets frozen -> guest kernel flushes pending I/O to disk -> no new writes and this allows LH/anyone else time to sync.

> Do we care about the potential failure from low level, and, will un-freeze (``qmp_guest_fsfreeze_thaw -> qmp_guest_fsfreeze_do_thaw) be called explicitly?

Our call to KubeVirt includes a duration[1] after which KubeVirt should automatically try to thaw the file system:

> Freeze command has additional parameter - the unfreezeTimeout - the timeout when an automatic unfreeze command is executed in the virt launcher to prevent the vmi from being stuck on freeze. It can be set to 0 to disable automatic unfreeze, but that is not recommended.

from: https://github.com/kubevirt/kubevirt/blob/main/docs/freeze.md#freezeunfreeze-subresource-api

[1] https://github.com/harvester/harvester/pull/3976/files#diff-75556fad8ea25a4e0f7d5c45bfea2abc49909d662b944d47f1301e0843b97469R293

w13915984028:
@connorkuehl thanks for you perfect reply :)

pkg/controller/master/backup/backup.go
w13915984028: No, it does not block here.

This PR tries to freeze the fs for a moment (and thaw automatically), is an indepedent enhancement.

For the potential data crash/in-consistency, we need to further analyze other places, sure, which is out of scope of this PR. thanks.

w13915984028 (approved):
LGTM, thanks.

pkg/controller/master/backup/backup.go
WebberHuang1118: IMHO,

> > Can we guarantee one second is long enough for all CSI drivers (Longhorn currently) to take a snapshot? Since it may be related to complex operations, as Longhorn for example, the driver has to [create a new layer for replicas and update read indexes on all nodes ](https://longhorn.io/docs/1.4.2/concepts/#241-how-snapshots-work)
> 
> No, I don't think so.
> 
> I think the ideal freeze duration is not the duration of the snapshot, but instead however long it takes to flush all I/O that was pending at the time the user executed the snapshot. I believe that duration is much, much smaller than the duration of a whole snapshot. Maybe 1 second is too low, but the duration of the snapshot would be the worst case, IMO. I'll comment more on this below.
> 
> > As stated above, if taking a snapshot requires longer than we expected, filesystem I/O could be back to work before snapshot completion, and it may bring unexpected behaviour.
> 
> I think this is also true for how snapshots/backups in Harvester work today.
> 
> I don't know anything about Longhorn internals, but ideally during a snapshot, it would be redirecting new writes from the guest to some sort of overlay that can be merged down to the original disk after the snapshot completes. thinking If this isn't how it behaves for snapshots then indeed this complicates things.

From my understanding, Longhorn has to create a new layer for all replicas and update the read indexes for all nodes. If during this process, the filesystem is thawed and back to process I/Os again. The snapshotted storage status could be uncertain.  And different CSI drivers may have different behaviors?

> 
> > Should we consider freezing vm filesystem before taking a snapshot, and unfreezing it after the snapshot, like [KubeVirt's handling for snapshot](https://kubevirt.io/user-guide/operations/snapshot_restore_api/#snapshot-a-virtualmachine)?
> 
> I would prefer to freeze for exactly how long it takes to flush pending I/O to disk at time of snapshot vs freezing for the duration of the snapshot, which I suspect will take much longer than flushing the pending I/O.
> 
> I might be misunderstanding how the FS freezing mechanism works, but if the file system is frozen, then none of the guest's writes are making it to durable storage. So, for the window that the FS is frozen, the risk of dataloss is greater in case of an unexpected shutdown from the guest, which is why I'd prefer to freeze for as small of a duration as possible vs the duration of the entire snapshot.

According to:
- https://github.com/kubevirt/kubevirt/blob/main/docs/freeze.md "Kubevirt provides a way to freeze/thaw the guest filesystem. That can be used to ensure the filesystem consistency when making a backup or a VM snapshot." 
- https://linux.die.net/man/8/fsfreeze "all dirty data, metadata, and log information are written to disk"

I think filesystem freeze is here to guarantee data consistency, but like what you pointed, the new coming I/Os would be halted

> 
> That said, freezing the FS for exactly the amount of time it takes to flush pending I/O is a very hand-wavey thing to say, and probably the best thing we can do is make it tunable.

Agree, making timeout tunable seems a practical solution. [The default deadline for KubeVirt vmSnapshot is 5 minutes](https://kubevirt.io/user-guide/operations/snapshot_restore_api/#snapshot-a-virtualmachine), and if the vmSnapshot has not successfully completed before the deadline, it will be marked as Failed. 
> 
> > I run Harvester HOST on KVM VM, and take vm snapshort/restore frequently in Virt-manager, to switch between diffrent Harvester versions, for such scale of VM (10 core, 20G memory, 200G disk), even in really fast PC, it still takes a long time to finish a snapshort, inbetween that, the VM is shown as suspend/freezed.
> 
> Yeah, I think in this case the "suspend/freeze" isn't referring to the FS being frozen, but instead libvirt pausing the domain to take a full system snapshot (cpu registers, RAM, disk), resulting in downtime. The alternative to this is having libvirt do an external disk snapshot where it just snapshots the disk while it redirects new writes from the live guest to a scratch file and then merges that back into the original disk once the snapshot completes. (Sorry if none of this is new info, I just want to be careful where I use the word "freeze" in this PR so I can be clear which mechanism I'm referring to).
> 
> Using the qemu-guest-agent in Harvester as shown won't pause the VM, so the VM should not experience downtime in that way.
> 
> > Back to here, seems difficult to set an one-fit-all timeout parameter, we may need to calculate it per VM scale and experience.
> > Or even expose the timeout as a parameter to take snapshort, give an initial estimated value, and user can further change it.
> 
> I think exposing the parameter is probably the most user-friendly thing here. We can have a conservative default of a few seconds, but users can choose to extend this if they want, based on the size of their VM or how I/O intensive their workload is.
> 
> Is this PR blocked on that parameter?

WebberHuang1118 (approved):
IMHO,

> > Can we guarantee one second is long enough for all CSI drivers (Longhorn currently) to take a snapshot? Since it may be related to complex operations, as Longhorn for example, the driver has to [create a new layer for replicas and update read indexes on all nodes ](https://longhorn.io/docs/1.4.2/concepts/#241-how-snapshots-work)
> 
> No, I don't think so.
> 
> I think the ideal freeze duration is not the duration of the snapshot, but instead however long it takes to flush all I/O that was pending at the time the user executed the snapshot. I believe that duration is much, much smaller than the duration of a whole snapshot. Maybe 1 second is too low, but the duration of the snapshot would be the worst case, IMO. I'll comment more on this below.
> 
> > As stated above, if taking a snapshot requires longer than we expected, filesystem I/O could be back to work before snapshot completion, and it may bring unexpected behaviour.
> 
> I think this is also true for how snapshots/backups in Harvester work today.
> 
> I don't know anything about Longhorn internals, but ideally during a snapshot, it would be redirecting new writes from the guest to some sort of overlay that can be merged down to the original disk after the snapshot completes. thinking If this isn't how it behaves for snapshots then indeed this complicates things.

From my understanding, Longhorn has to create a new layer for all replicas and update the read indexes for all nodes. If during this process, the filesystem is thawed and back to process I/Os again. The snapshotted storage status could be uncertain.  And different CSI drivers may have different behaviors?

> 
> > Should we consider freezing vm filesystem before taking a snapshot, and unfreezing it after the snapshot, like [KubeVirt's handling for snapshot](https://kubevirt.io/user-guide/operations/snapshot_restore_api/#snapshot-a-virtualmachine)?
> 
> I would prefer to freeze for exactly how long it takes to flush pending I/O to disk at time of snapshot vs freezing for the duration of the snapshot, which I suspect will take much longer than flushing the pending I/O.
> 
> I might be misunderstanding how the FS freezing mechanism works, but if the file system is frozen, then none of the guest's writes are making it to durable storage. So, for the window that the FS is frozen, the risk of dataloss is greater in case of an unexpected shutdown from the guest, which is why I'd prefer to freeze for as small of a duration as possible vs the duration of the entire snapshot.

According to:
- https://github.com/kubevirt/kubevirt/blob/main/docs/freeze.md "Kubevirt provides a way to freeze/thaw the guest filesystem. That can be used to ensure the filesystem consistency when making a backup or a VM snapshot." 
- https://linux.die.net/man/8/fsfreeze "all dirty data, metadata, and log information are written to disk"

I think filesystem freeze is here to guarantee data consistency, but it would halt newly coming I/Os and let applications experience long I/O latency. 

I had an experiment to check this:
- Staring a VM with qemu guest agent on Harvester
- Writing big file in VM (e.q. dd if=/dev/zero of=file bs=1M count=20480)
- Freeze VM filesystem with [virt-freezer](https://github.com/kubevirt/kubevirt/blob/main/docs/freeze.md#virt-freezer)
- We can find the file stop to grow, but there is no error from dd
- Unfreeze VM filesystem with [virt-freezer](https://github.com/kubevirt/kubevirt/blob/main/docs/freeze.md#virt-freezer)
- We can find the dd continued and created a file with correct size in the end

Based on the observation, it seems filesystem freeze doesn't bring data loss, but it could introduce high I/O latency. If it lasts too long, applications may treat it as an I/O timeout, but it depends on the application's implementation. So a timeout value for filesystem freeze is necessary.

> 
> That said, freezing the FS for exactly the amount of time it takes to flush pending I/O is a very hand-wavey thing to say, and probably the best thing we can do is make it tunable.

Agree, making timeout tunable seems a practical solution. For exmaple, [KubeVirt has a default deadline for vmSnapshot, if the vmSnapshot has not successfully completed before the deadline, it will be marked as Failed](https://kubevirt.io/user-guide/operations/snapshot_restore_api/#snapshot-a-virtualmachine).
> 
> > I run Harvester HOST on KVM VM, and take vm snapshort/restore frequently in Virt-manager, to switch between diffrent Harvester versions, for such scale of VM (10 core, 20G memory, 200G disk), even in really fast PC, it still takes a long time to finish a snapshort, inbetween that, the VM is shown as suspend/freezed.
> 
> Yeah, I think in this case the "suspend/freeze" isn't referring to the FS being frozen, but instead libvirt pausing the domain to take a full system snapshot (cpu registers, RAM, disk), resulting in downtime. The alternative to this is having libvirt do an external disk snapshot where it just snapshots the disk while it redirects new writes from the live guest to a scratch file and then merges that back into the original disk once the snapshot completes. (Sorry if none of this is new info, I just want to be careful where I use the word "freeze" in this PR so I can be clear which mechanism I'm referring to).
> 
> Using the qemu-guest-agent in Harvester as shown won't pause the VM, so the VM should not experience downtime in that way.
> 
> > Back to here, seems difficult to set an one-fit-all timeout parameter, we may need to calculate it per VM scale and experience.
> > Or even expose the timeout as a parameter to take snapshort, give an initial estimated value, and user can further change it.
> 
> I think exposing the parameter is probably the most user-friendly thing here. We can have a conservative default of a few seconds, but users can choose to extend this if they want, based on the size of their VM or how I/O intensive their workload is.
> 
> Is this PR blocked on that parameter?

Aside from making filesystem freeze timeout tunable, 
LGTM, thanks

pkg/controller/master/backup/backup.go
```diff
@@ -226,6 +247,65 @@ func (h *Handler) getBackupSource(vmBackup *harvesterv1.VirtualMachineBackup) (*
...
+		return nil, err
+	}
+	if sourceVMI.DeletionTimestamp != nil {
+		return nil, fmt.Errorf("vm %s/%s is being deleted", vmBackup.Namespace, vmBackup.Spec.Source.Name)
+	}
+
+	return sourceVMI, nil
+}
+
+func (h *Handler) tryFreezeFS(backup *harvesterv1.VirtualMachineBackup) {
```
guangbochen: why we are not returning the error here? it should retry the freeze before creating the volumeSnapshot object.

connorkuehl: I initially didn't add a retry so that freezing the FS is best-effort but doesn't block the snapshot from being carried out.

Should it retry only a certain amount of times or indefinitely? What kind of behavior would you prefer?

guangbochen: I believe the fs freeze should be a required operation so that the snapshotContent will always be expected and more consistent. 
The controller framework will reconcile it as long as you have returned an appropriate error, I would suggest we add the error message to the vmBackup status and clean it up after it was succeeded.

w13915984028: For this specific scenario, it is better not to return error and retry, a couple of considerations:
(1) the freeze/thaw happens before the snapshot/backup, if VM is keeping writing, this action may not help too much, but at least, no harm
(2) it is difficult to define an unified timeout param, in certain environment, if fail, it may always fail, e.g., timeout is not enough, or low level fs / driver return error due to other reasons, the retry will block this feature

(3) besides this independent enhancement, we need to further investigate & enhance the end to end backup/snapshot

image

we may log a warning / k8s event when freeze fails

connorkuehl: @guangbochen did @w13915984028's comment sway you or should I push a revision that returns an error?

guangbochen: @w13915984028 @connorkuehl Adding a proper log and event would be helpful, but for each API call or controller reconciliation, we need to make sure the business logic and result should be as idempotent as possible.
The fsfreeze will help to halt new access to the filesystem and creates a stable image on disk, so at least it will help to make sure the snapshot content is more expected than the previous random behavior, and in regards to the above concerns:
- since this is a live snapshot/backup, so the content taken by the snapshot would only be the best effort; if users want  a consistent one, shutdown the VM and taking either a snapshot or backup probably is the best option(and we can document this)。
- we need to add proper status if the fsfreeze call fails, so the user can decide whether to delete and create a new one, or wait for the controller to do the reconciliation(instead of skipping the error as the user may never check the log/event).
- TODO: you probably don't want to call fs to freeze when the VM is shut down

connorkuehl: > TODO: you probably don't want to call fs to freeze when the VM is shut down

I think this is currently the case, since the function returns early if there's no VMI.

connorkuehl: > but for each API call or controller reconciliation, we need to make sure the business logic and result should be as idempotent as possible.

Just to make sure I'm parsing this correctly, does this mean that if the backup controller confirms the qemu-guest-agent is present and it tries to freeze the FS but it errors, then subsequent reconciliation attempts should remember that the first attempt was made with fsfreeze and so it should always try to freeze the FS even if the qemu-guest-agent is no longer running?

So instead of this:

```mermaid
flowchart TD
    backupchange(OnBackupChange)
    createvolsnap(h.createVolumeSnapshot)
    hasga{Has guest agent?}
    putfrz(PUT freeze subresource)
    isrun{Is running/has VMI}
    
    Start-->backupchange
    backupchange --> reconcileVolumeSnapshots
    reconcileVolumeSnapshots --> isrun
    isrun-- yes -->hasga
    isrun-- no --> createvolsnap
    hasga-- yes --> putfrz
    hasga-- no --> createvolsnap
    putfrz-- fail --> Enqueue
    putfrz-- success --> createvolsnap
    Enqueue-.->backupchange
```

It should behave like this:

```mermaid
flowchart TD
    backupchange(OnBackupChange)
    createvolsnap(h.createVolumeSnapshot)
    recvolsnap(reconcileVolumeSnapshots)
    storega(persist fsfreeze strategy=true somewhere?)
    storenoga(persist fsfreeze strategy=false somewhere?)
    storenoga2(persist fsfreeze strategy=false somewhere?)
    hasga{Has guest agent?}
    putfrz(PUT freeze subresource)
    hasfrzstrat{Has fsfreeze strategy?}
    fetchfrzstrat(fetch persisted freeze strategy)
    isrun{Is running/has VMI}
    
    Start-->backupchange
    backupchange --> recvolsnap
    recvolsnap --> hasfrzstrat
    hasfrzstrat-- no -->isrun
    isrun-- yes -->hasga
    isrun-- no -->storenoga2
    storenoga2-->createvolsnap
    hasfrzstrat-- yes -->fetchfrzstrat
    fetchfrzstrat-- fsfreeze=true -->putfrz
    fetchfrzstrat-- fsfreeze=false -->createvolsnap
    putfrz-. error .-> Enqueue
    putfrz-- ok -->createvolsnap
    Enqueue-.->backupchange
    
    hasga-- no -->storenoga
    hasga-- yes -->storega
    storenoga-->fetchfrzstrat
    storega-->fetchfrzstrat
```

guangbochen: @connorkuehl the new flow looks correct to me, and I think we could just store it in the VMBackup annotations.

connorkuehl: Thanks, @guangbochen! I've pushed a revision that implements the new flow.

To my other reviewers, FYI: the code has changed in a nontrivial but not hugely substantial way since the revision you approved. (in case this changes your approval) CC: @w13915984028 @WebberHuang1118 @ibrokethecloud @markhillgit

pkg/controller/master/backup/backup.go
```diff
@@ -226,6 +249,118 @@ func (h *Handler) getBackupSource(vmBackup *harvesterv1.VirtualMachineBackup) (*
...
+	if sourceVMI.DeletionTimestamp != nil {
+		return nil, fmt.Errorf("vm %s/%s is being deleted", vmBackup.Namespace, vmBackup.Spec.Source.Name)
+	}
+
+	return sourceVMI, nil
+}
+
+func (h *Handler) tryFreezeFS(backup *harvesterv1.VirtualMachineBackup) error {
+	var err error
+	backup, err = h.withFreezeFSAnnotation(backup)
```
guangbochen: prefer direct assignment to merge them into one line

pkg/controller/master/backup/backup.go
```diff
@@ -226,6 +249,118 @@ func (h *Handler) getBackupSource(vmBackup *harvesterv1.VirtualMachineBackup) (*
...
+		logrus.WithError(err).Warn("qemu-guest-agent fs-freeze")
+		return errors.New("failed to freeze file system")
+	}
+
+	return nil
+}
+
+func (h *Handler) withFreezeFSAnnotation(backup *harvesterv1.VirtualMachineBackup) (*harvesterv1.VirtualMachineBackup, error) {
+	// Annotation already exists with valid bool text
+	val, ok := backup.Annotations[util.AnnotationSnapshotFreezeFS]
```
guangbochen, line 303: always check annotation is not nil first

pkg/controller/master/backup/backup.go
```diff
@@ -226,6 +249,118 @@ func (h *Handler) getBackupSource(vmBackup *harvesterv1.VirtualMachineBackup) (*
...
+			hasAgent = condition.Status == corev1.ConditionTrue
+			break
+		}
+	}
+
+	return h.saveFreezeFSAnnotation(backup, hasAgent)
+}
+
+func (h *Handler) saveFreezeFSAnnotation(backup *harvesterv1.VirtualMachineBackup, value bool) (*harvesterv1.VirtualMachineBackup, error) {
+	changed := backup.DeepCopy()
```
guangbochen: \s\changed\backupCopy

pkg/controller/master/backup/backup.go
```diff
@@ -226,6 +249,118 @@ func (h *Handler) getBackupSource(vmBackup *harvesterv1.VirtualMachineBackup) (*
...
+
+	v := "true"
+	if !value {
+		v = "false"
+	}
```
guangbochen: u may use `strconv.FormatBool` here.

guangbochen (approved):
Just some minor fixes, everything else is LGTM

guangbochen:
And per @w13915984028's comment: 
> it is difficult to define an unified timeout param, in certain environment, if fail, it may always fail, e.g., timeout is not enough, or low level fs / driver return error due to other reasons, the retry will block this feature

Can we file a separate issue to track it to allow timeout customization as a future enhancement? thanks  cc @bk201

pkg/controller/master/backup/backup.go
connorkuehl: Is this a stylistic preference? Only writing to a nil map will panic, but lookup, delete, len, and range loops are fine.

WebberHuang1118 (approved):
Based on how KubeVirt utilizes filesystem freeze and the timout:
-  https://kubevirt.io/user-guide/operations/snapshot_restore_api/#snapshot-a-virtualmachine
- https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-controller/watch/snapshot/snapshot.go#L121

The [Freeze's timeout parameter](https://github.com/kubevirt/kubevirt/blob/main/docs/freeze.md) is the deadline for operations (usually snapshot taking) between Filesystem freeze and unfreeze. 

Taking KubeVirt's vm snapshot as an example, KubeVirt freezes filesystem and takes the snapshot:
1. if snapshot completes before timeout, it explicitly calls filesystem unfreeze
2. if snapshot fails to complete before timeout, filesystem unfreeze automatically executed as the deadline is met, and update snapshot with "deadline exceed" fail status

IMHO, a tunable timeout and the snapshot status maybe is necessary,
aside from this, LGTM, thanks.

pkg/controller/master/backup/backup.go
guangbochen: then this should be fine in here and we can leave it as what it is, thanks.

bk201:
@mergifyio backport v1.2

Issue #3957, opened 2023-05-24
[BUG] No PCI device display after upgrade to v1.2.0 master and enabled the pcidevices-controller addons


TachunLin:
**Describe the bug**

After we upgrade from v1.1.2 to v1.2.0 master-b49e264e-head and **enable** the pcidevices-controller.

Open the PCI Devices page, there is no available pcidevices display. 

https://github.com/harvester/harvester/assets/29251855/0d004346-67f3-4913-a083-f08972eb730b

![image](https://github.com/harvester/harvester/assets/29251855/2f750f49-e529-4f0d-a57a-e82e3122268a)

![image](https://github.com/harvester/harvester/assets/29251855/54e2589d-df5e-49e1-aed8-2d67d357104d)

**To Reproduce**
1. Upgrade from v1.1.2 to v1.2.0 master 
2. After upgrade complete, open Addons page
3. Enable the `pcidevices-controller` addon
4. Check the addon in `Enabled` state
5. Open the PCI Devices page 
6. Check the content 

**Expected behavior**

After upgrade from v1.1.2 to v1.2.0 master and enable the pcidevices-controller.
We should be able to see the availabe PCI devices and use this feature

**Support bundle**

[supportbundle_ebea348e-b561-4725-8608-26223fe6f392_2023-05-24T05-51-02Z.zip](https://github.com/harvester/harvester/files/11550900/supportbundle_ebea348e-b561-4725-8608-26223fe6f392_2023-05-24T05-51-02Z.zip)

**Environment**
 - Harvester ISO version: master-b49e264e-head (05/22)
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 nodes bare machines

bk201:
@TachunLin need to clarify, did you enable the PCI devices add-on before or after the upgrade? Thanks.
The recording has the add-on enabled already.

TachunLin:
Before upgrade the PCI devices addon did not enabled, just as we fresh install v1.1.2 Harvester cluster. 
After upgrade to v1.2.0 master, I try to enable the addon and found no PCI devices list on the page.

In comparison, I fresh install a single node `v1.2.0-rc1` cluster and enable the PCI device addon on kvm machine
![image](https://github.com/harvester/harvester/assets/29251855/ae491af9-4c8e-4fd5-9d92-470880d4d3d2)

The pci devices can display correctly after enabled the addon
[...]

PR #3987, opened 2023-05-29
fix(upgrade): remove version check for addon upgrades

starbops:

**Problem:**

After upgrading from v1.1.2 to master-head, there are no PCI devices listed on the page when pcidevices-controller addon is turned on. The pcidevices didn't show up automatically.

**Solution:**

The version checking is no longer valid. Removing it to make sure every Harvester upgrade will upgrade the addons.

**Related Issue:**

#3957 

**Test plan:**

1. Install a v1.1.2 Harvester single-node cluster
2. Upgrade the cluster with an ISO image containing this fix
3. After the upgrade succeeded, check the version of the addons in the cluster. They should be upgraded (now only pcidevices-controller got newer release).

```
pcidevices-controller: 0.2.4 -> 0.2.5-rc1
```

4. Login Harvester UI and enable the pcidevices-controller addon
5. Wait for a few seconds; on the page "PCIDevices" there should be items pop up automatically

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -624,14 +624,11 @@ pause_all_charts() {
 
 upgrade_addons()
 {
- local is_upgrade_required=$(lower_version_check $UPGRADE_PREVIOUS_VERSION v1.1.1)
- if [ ! -z "$is_upgrade_required" ]; then
-   wait_for_addons_crd
+  wait_for_addons_crd
   addons="vm-import-controller pcidevices-controller"
   for addon in $addons; do
     upgrade_addon $addon "harvester-system"
   done
-  fi
 }
 
 reuse_vlan_cn() {


bk201 (approved):
lgtm! Thanks

ibrokethecloud (approved):
lgtm.

starbops:
@mergify backport v1.1

package/upgrade/upgrade_manifests.sh
```diff
@@ -624,14 +624,11 @@ pause_all_charts() {
 
 upgrade_addons()
 {
- local is_upgrade_required=$(lower_version_check $UPGRADE_PREVIOUS_VERSION v1.1.1)
- if [ ! -z "$is_upgrade_required" ]; then
-   wait_for_addons_crd
+  wait_for_addons_crd
```
w13915984028, line 627: `wait_for_addons_crd` is also only needed in specific version, but it seems harmless to wait again in later version, we may remove it in future.

Issue #3946, opened 2023-05-22
[BUG] We can set negative integer value to `Termination Grace Period` field and write into VM yaml


TachunLin:
**Describe the bug**

* When we create a vm and provide negative integer value `-1` to `Termination Grace Period` in advance options. 
* The UI can accept this value to create VM and write the negative value into the `spec.spec.terminationGracePeriodSeconds` 

![image](https://github.com/harvester/harvester/assets/29251855/c01baeef-2e0b-4eb7-a1aa-5dfcb64cd050)

![image](https://github.com/harvester/harvester/assets/29251855/99ffc29e-055b-4e10-8c94-998bbd616245)

* When we directly update the `spec.spec.terminationGracePeriodSeconds` of virtual machines setting to negative value. The webhook did not prompt message to block to proceed 

**To Reproduce**
1. Create a virtual machine
2. Expand Advanced Options 
3. Set `-1` to the `Termination Grace Period` field and save
4. Wait for VM in Running state
5. Check the yaml of VM

**Expected behavior**

When we set negative value to `Termination Grace Period` in advance options, the UI should prompt message to prevent.
And also when we update the `spec.spec.terminationGracePeriodSeconds` on virtual machine, the webhook should prompt message to prevent.

**Support bundle**

[supportbundle_3c320152-ecaf-4ca1-aed1-650eef8d3da8_2023-05-22T08-28-29Z.zip](https://github.com/harvester/harvester/files/11528820/supportbundle_3c320152-ecaf-4ca1-aed1-650eef8d3da8_2023-05-22T08-28-29Z.zip)

**Environment**
 - Harvester ISO version:  `master-b49e264e-head` 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Single nodes on local kvm machine

**Additional context**
Related to #2466

WuJun2016:
cc @bk201   @guangbochen   The backend may also need webhook validation.

irishgordo:
@n313893254 @WebberHuang1118  - thanks for the fix :smile: :+1: 
This looks great!
Tested with:
- Harvester Version: master-5557c692-head (06/05/23)
- Rancher v2.7.4


https://github.com/harvester/harvester/assets/5370752/33fb4030-cff6-415f-985a-13c64233fa63

I'll go ahead and close this out

PR #3991, opened 2023-05-29
Prevent VM's "TerminationGracePeriod" from negative value

WebberHuang1118:

**Problem:**

Harvester can create or update VM's `TerminationGracePeriod` with a negative value.

**Solution:**

Prevent negative value for VM's `TerminationGracePeriod` during creation or update via vm validator webhook.

In addition, vm mutator sets the default value for `TerminationGracePeriod` field during the update. Make the handling consistent between creation and update.

**Related Issue:**
Issue #3946

**Test plan:**

Case1: Setting negative `TerminationGracePeriod` for VM creation
- Create a virtual machine
- Expand Advanced Options
- Set a negative value to `TerminationGracePeriod` and save
- We should see the operation be prohibited by webhook with msg _`Termination grace period can't be negative`_ vm_create_grace_period

Case2: Setting negative `TerminationGracePeriod` value for VM update
- Create a virtual machine with default `TerminationGracePeriod` value
- Open the edit page for the newly created VM
- Set a negative value to `TerminationGracePeriod` and save
-  We should see the operation be prohibited by webhook vm_edit_grace_period

### DIFF
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -108,6 +108,11 @@ func (m *vmMutator) Update(request *types.Request, oldObj runtime.Object, newObj
 		return nil, err
 	}
 
+	patchOps, err = m.patchTerminationGracePeriodSeconds(newVM, patchOps)
+	if err != nil {
+		return nil, err
+	}
+
 	return patchOps, nil
 }
 
@@ -400,6 +405,7 @@ func (m *vmMutator) patchTerminationGracePeriodSeconds(vm *kubevirtv1.VirtualMac
 	if s.Value != "" {
 		value = s.Value
 	}
+
 	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/terminationGracePeriodSeconds", "value": %s}`, value))
 	return patchOps, nil
 }
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -108,6 +108,9 @@ func (v *vmValidator) checkVMSpec(vm *kubevirtv1.VirtualMachine) error {
 	if err := v.checkOccupiedPVCs(vm); err != nil {
 		return err
 	}
+	if err := v.checkTerminationGracePeriodSeconds(vm); err != nil {
+		return err
+	}
 	if err := v.checkReservedMemoryAnnotation(vm); err != nil {
 		return err
 	}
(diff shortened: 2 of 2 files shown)


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -239,6 +242,22 @@ func (v *vmValidator) checkVMBackup(vm *kubevirtv1.VirtualMachine) error {
 	return nil
 }
 
+func (v *vmValidator) checkTerminationGracePeriodSeconds(vm *kubevirtv1.VirtualMachine) error {
+	TerminationGracePeriodSeconds := vm.Spec.Template.Spec.TerminationGracePeriodSeconds
```
bk201: non-constant variable within local scope starts with lower case conventionally.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -239,6 +242,22 @@ func (v *vmValidator) checkVMBackup(vm *kubevirtv1.VirtualMachine) error {
...
+
+	//Even mutator prevents this from happening, we still check it for robustness
```
bk201: Can you explain the comment more, why mutator prevents this happening. Thanks.

pkg/webhook/resources/virtualmachine/validator.go
WebberHuang1118: Thanks, I will modify it.

pkg/webhook/resources/virtualmachine/validator.go
WebberHuang1118: From my understanding, if VM's `TerminationGracePeriodSeconds` is not specified, vmMutator's [`patchTerminationGracePeriodSeconds()`](https://github.com/harvester/harvester/blob/5a7cf30480868276009877186bf0326df05cf7c7/pkg/webhook/resources/virtualmachine/mutator.go#L383) function will set VM's `TerminationGracePeriodSeconds` as `default-vm-termination-grace-period-seconds` from settings. I will add more details in this comment. Thanks

bk201: Thanks for the explanation. The guard here makes sense, we should not assume there is always a mutation.

WebberHuang1118: I just found that the VM's `TerminationGracePeriodSeconds` still could be nil, if it's updated by config yaml directly without specifying the `TerminationGracePeriodSeconds` field. I will update the comment according to this, too. thanks.

bk201 (approved):
LGTM

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -400,6 +405,7 @@ func (m *vmMutator) patchTerminationGracePeriodSeconds(vm *kubevirtv1.VirtualMac
 	if s.Value != "" {
 		value = s.Value
 	}
+
```
Vicente-Cheng, line 408: Do we need this new line?

Vicente-Cheng (approved):
Just a nit, others LGTM.

pkg/webhook/resources/virtualmachine/mutator.go
WebberHuang1118: Just from habit, give some space for different code blocks. Do you suggest removing it?

Vicente-Cheng: That's ok for me. Thanks for the explanation.

guangbochen:
@mergify backport v1.2

Issue #4015, opened 2023-05-31
[BUG] Maintenance Mode of Node w/ SR-IOV VFs in Use Causes Cordon & Empty/`null` Error


irishgordo:
**Describe the bug**
When a node that is being used for its physical Network hardware NIC, that has Virtual Functions (VFs) created on it with SR-IOV, that are currently in use by one or more VMs, the node when put into maintenance mode is cordoned and then when attempting again to enable maintenance mode the error back is `null`

**To Reproduce**
1. Create a 3 node harvester bare metal cluster
2. Enable PCI Devices Add-on
3. Use one host for enabling SR-IOV 
4. Create a VM, adding the SR-IOV Virtual Function, keep it running
5. Attempt to migrate the VM, you will see error as expected
6. Then go to hosts -> enable maintenance mode on the host
7. You will be able to enable maintenance mode, with no error, the host will be in cordon state
8. Attempt to enable maintenance mode again, then error returns with `null` 

**Expected behavior**
Perhaps not have a `null` state on error - or not allow user to put host into maintenance mode until certain criteria is met?

**Environment**
 - Harvester ISO version: v1.2.0-rc1
 - Underlying Infrastructure: 3 node HP Proliant
 
**Additional context**
![Screenshot from 2023-05-31 15-20-44](https://github.com/harvester/harvester/assets/5370752/083ee298-12bc-44bd-a5a8-af4580e238d2)
![Screenshot from 2023-05-31 15-19-37](https://github.com/harvester/harvester/assets/5370752/e827fdaa-059f-4f05-9e76-827018adfea2)
![Screenshot from 2023-05-31 15-25-24](https://github.com/harvester/harvester/assets/5370752/6774b31d-3777-4caa-b32f-59578219d817)
![Screenshot from 2023-05-31 15-26-22](https://github.com/harvester/harvester/assets/5370752/80b1d8f2-1c46-4ced-955f-caa6b98b638d)

ibrokethecloud:
the `null` error may be transient. I will try and check if i can re-produce it.

We however do need to figure out how we want to handle forced maintenance on node with VM's having pcidevices.

These VM's are non migratable. May be we should treat them like VM's with last healthy replicas, which are shutdown or provide a warning to users that they should power these off.

@bk201 @guangbochen appreciate your feedback on this.

irishgordo:
@ibrokethecloud  I agree, the `null` error does indeed seem to be transient - I am not able to replicate seeing the `null` error.

[...]

PR #4018, opened 2023-06-01
maintenance mode api changes

ibrokethecloud:

**Problem:**

VM's with PCIDevices are not migratable.

As a result attempts to place nodes with VM's leveraging pcidevice passthrough is not going to be successful.

**Solution:**

Additional check in the maintenance mode api to query VM's on the target node containing pcidevices and provide details of the same.

User is expected to turn off these VM's before performing the maintenance operations.

**Related Issue:**
https://github.com/harvester/harvester/issues/4015

**Test plan:**

* create a VM with pcidevice attached
* attempt to place node the VM is scheduled to into maintenance mode.
* a warning should be displayed detailing the list of VM's on node in maintenance mode and blocking the operation.
* shutdown listed vm's
* attempt to place node into maintenance mode again.
* this should not be successful.

### DIFF
--- a/pkg/api/node/formatter.go
+++ b/pkg/api/node/formatter.go
@@ -203,6 +203,18 @@ func (h ActionHandler) listUnhealthyVM(rw http.ResponseWriter, node *corev1.Node
 		return json.NewEncoder(rw).Encode(&respObj)
 	}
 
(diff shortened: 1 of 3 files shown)


guangbochen (approved):
LGTM, thanks.

futuretea (approved):
LGTM, Thanks. The last step in the test plan should be `successful`

Issue #4049, opened 2023-06-08
[BUG] The VM needs at least 10 minutes to automatically recover after a host failure


chrisho:
**Describe the bug**
Create a VM in a cluster consisting of three nodes and run it. And I forced shut down the node running the VM and deleted the pod.
At this point, the VM hasn't immediately auto-restarted, which could take more than 10 minutes.

**To Reproduce**
1. Change the setting `vm-force-reset-policy` period to fewer seconds, `period: 30`.
2. Create a VM.
3. Force shutdown of the node where the VM is running.
4. Waiting for the node to change `not-ready` state.
5. While the VM automatically recovers, this takes more than 10 minutes.

**Expected behavior**
Except to use less time to recover the VM.

**Support bundle**
[longhorn-support-bundle_0654c6f1-8452-4cbf-becc-531c7a01703a_2023-06-07T03-17-55Z.zip](https://github.com/harvester/harvester/files/11683478/longhorn-support-bundle_0654c6f1-8452-4cbf-becc-531c7a01703a_2023-06-07T03-17-55Z.zip)

**Environment**
 - Harvester ISO version: v1.2.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
After investigation, we found that Kubernetes Controller-Manager must wait 6 minutes to delete the unavailable node VolumeAttachment and create a new node VolumeAttachment. 
Only after CSI detects the new VolumeAttachment will it request the longhorn Volume, fix the status, and remount it to the node where the Pod is located.
https://github.com/kubernetes/kubernetes/blob/da1f52034eb3c405de1ca2dc0e1a6163e8a48f4f/pkg/controller/volume/attachdetach/reconciler/reconciler.go#L216-L293

Kubevirt node's heartbeat doc link:  https://kubevirt.io/user-guide/operations/unresponsive_nodes/#virt-handler-heartbeat

**Workaround**
Reduce the Node CR's heartbeat annotation by more the 5 minutes.
Delete VolumeAttachment before recreating Pod, and it can restore running within 1 minute.

**Relate issue**
LH: https://github.com/longhorn/longhorn/issues/6075

chrisho:
Describe in the documentation more details when to perform a live migration or when to restart/reschedule.

albinsun:
Tested in `master-973a1a9a-head` with `vm-force-reset-policy`=30

[...]

PR #4054, opened 2023-06-09
Fix VM auto-restart post host failure

chrisho:
**Problem:**
https://github.com/harvester/harvester/issues/4049

**Solution:**
The following operate after removing the VM pod:
1. The controller will remove pod-related VolumeAttachments.
2. Reduce 6 minutes to the node's heartbeat annotation. 

**Related Issue:**
https://github.com/harvester/harvester/issues/4049

**Test plan:**
https://github.com/harvester/harvester/issues/4049


pkg/controller/master/node/node_down_controller.go
```diff
@@ -156,3 +174,89 @@ func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.No
...
+// due to controller-manager handle too long time, cause VM restart wait a long time.
+// then we delete VolumeAttachments directly.
+// ref: https://github.com/harvester/harvester/issues/4049
+func (h *nodeDownHandler) deleteVolumeAttachmentsOfPod(pod *corev1.Pod) error {
+	volumeAttachments, err := h.getVolumeAttachmentsOfPod(pod)
+	if err != nil {
+		return err
+	}
+	for _, va := range volumeAttachments {
+		if va.DeletionTimestamp == nil {
```
futuretea: ~Is this judgment necessary?~
reason: reduce unnecessary api request

pkg/controller/master/node/node_down_controller.go
```diff
@@ -156,3 +174,89 @@ func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.No
...
+			err := h.vas.Delete(va.Name, &metav1.DeleteOptions{})
+			if err != nil {
```
futuretea: these two lines can be joined

pkg/controller/master/node/node_down_controller.go
```diff
@@ -156,3 +174,89 @@ func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.No
...
+			continue
+		}
+		res = append(res, va)
+	}
+
+	return res, nil
+}
+
+// resetHeartbeat force reduce node's heartbeat to immediately rebuild VMI,
+// Ensure that can take effect here by reducing 6 minutes
```
futuretea, line 259: Can you add a comment to explain why does it need to be 6 minutes?

pkg/controller/master/node/node_down_controller.go
```diff
@@ -156,3 +174,89 @@ func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.No
...
+				pod.Name,
+				pod.Spec.NodeName)
+		}
+	}
+	return nil
+}
+
+func (h *nodeDownHandler) getVolumeAttachmentsOfPod(pod *corev1.Pod) ([]*storagev1.VolumeAttachment, error) {
+	res := make([]*storagev1.VolumeAttachment, 0)
+	volumeAttachments, err := h.vaCache.List(labels.Everything())
```
WebberHuang1118: Instead of listing all volumeAttachments in each pod iteration, how about getting volumeAttachments first, and checking which one should be deleted?

pkg/controller/master/node/node_down_controller.go
```diff
@@ -156,3 +174,89 @@ func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.No
...
+func (h *nodeDownHandler) resetHeartbeat(node *corev1.Node) (*corev1.Node, error) {
+	nodeCpy := node.DeepCopy()
+	if nodeCpy.Annotations == nil {
+		nodeCpy.Annotations = map[string]string{}
+	}
+
+	t := metav1.Now().Add(-6 * time.Minute).Format(time.RFC3339)
+	nodeCpy.Annotations[kubevirtv1.VirtHandlerHeartbeat] = t
+
+	return h.nodes.Update(nodeCpy)
```
WebberHuang1118, line 261: Should the update be checked with DeepEqual? In my environment, the heartbeat kept renewing.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -112,8 +126,12 @@ func (h *nodeDownHandler) OnNodeChanged(key string, node *corev1.Node) (*corev1.
 			return node, err
 		}
 
+		if err := h.deleteVolumeAttachmentsOfPod(&pod); err != nil {
+			return node, err
+		}
 	}
-	return node, nil
+
+	return h.resetHeartbeat(node)
```
WebberHuang1118, line 148: Should the heartbeat always need to update? In my environment, it kept renewing.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -156,3 +174,89 @@ func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.No
...
+	res := make([]*storagev1.VolumeAttachment, 0)
+	volumeAttachments, err := h.vaCache.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+
+	pvs := make(map[string]bool)
+
+	for _, vol := range pod.Spec.Volumes {
+		if vol.VolumeSource.PersistentVolumeClaim == nil {
```
w13915984028, line 195: @WebberHuang1118 's suggestion is practical,

The code could be optimized as:

(1) In the caller (NODE level), list `volumeAttachments` first, say `vas`, better to check if could list by nodeName;  after that, when `len==0`, no need to call this function,

(2) In this function, generate `pvs` first (most PODs do not have PVC), when  `len != 0`, then compare `pvs` and `vas`

You can also exchange the sequence of (1) and (2), get a POD list which do have PVC first, then get a list of `vas` is this NODE, finally cross check them.

pkg/controller/master/node/node_down_controller.go
futuretea: reduce unnecessary api request

pkg/controller/master/node/node_down_controller.go
chrisho: Add a judgment that returns if detect the heartbeat has been updated yet.

pkg/controller/master/node/node_down_controller.go
chrisho: In this function, filter out VM pods that should have one or more PVCs.

go.mod
```diff
@@ -62,6 +62,7 @@ require (
 	github.com/banzaicloud/logging-operator/pkg/sdk v0.8.16
 	github.com/banzaicloud/operator-tools v0.28.10
 	github.com/containernetworking/cni v1.1.2
+	github.com/deckarep/golang-set/v2 v2.3.0
```
futuretea: can we use "k8s.io/utils/set" instead ?

pkg/controller/master/node/node_down_controller.go
```diff
@@ -100,9 +112,12 @@ func (h *nodeDownHandler) OnNodeChanged(key string, node *corev1.Node) (*corev1.
...
 
+	pvcSet := mapset.NewSet[string]()
 	gracePeriod := int64(0)
-	for _, pod := range pods.Items {
-		logrus.Debugf("force delete pod %s/%s", pod.Namespace, pod.Name)
+	for i := range pods.Items {
+		pod := pods.Items[i]
+		logrus.Debugf("force delete pod %s/%s and VolumeAttachments", pod.Namespace, pod.Name)
+
 		if err := h.pods.Delete(
```
w13915984028, line 123: if any of the `h.pods.Delete` fail, the following `h.deleteVolumeAttachments` has no chance to execute;  when reconciler retry, the previous deleted PODs' vas have no chance to be cleared.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -112,11 +127,21 @@ func (h *nodeDownHandler) OnNodeChanged(key string, node *corev1.Node) (*corev1.
...
+	if err := h.deleteVolumeAttachments(node.Name, pvcSet); err != nil {
+		return node, err
 	}
-	return node, nil
+
+	return h.resetHeartbeat(node)
 }
 
-func (h *nodeDownHandler) OnVMForceResetPolicyChanged(key string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+func (h *nodeDownHandler) OnVMForceResetPolicyChanged(_ string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
```
w13915984028, line 151: please double check previous

```
// if we haven't waited for vmForceResetPolicy.Period seconds, we enqueue event again
```

When policy is changed, the alreay in queue NODE will not be executed quickly/slowly.

Should we adjust the code, e.g.  enqueue after only 1 minute, thus the NODE can adapt to the new param ?

w13915984028 (commented):
another 2 questions:

pkg/controller/master/node/node_down_controller.go
chrisho: Nice catch! It should add a logic that if occur an error, will return after the VolumeAttachments be removed.

pkg/controller/master/node/node_down_controller.go
chrisho: I have re-tested the above logic(quickly/slowly) because `OnVMForceResetPolicyChanged` will re-enqueue if the period changes and replace the existing node obj.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -156,3 +188,94 @@ func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.No
...
+	}
+	return nil
+}
+
+// resetHeartbeat force reduce node's heartbeat to immediately rebuild VMI,
+// the NodeController in Kubevirt force rebuilds the VMI if the node's heartbeat has no update in 5 minutes(hardcode)
+// Ensure that can take effect here by reducing 6 minutes
+// ref: https://kubevirt.io/user-guide/operations/unresponsive_nodes/#virt-handler-heartbeat
+func (h *nodeDownHandler) resetHeartbeat(node *corev1.Node) (*corev1.Node, error) {
+	nodeCpy := node.DeepCopy()
```
w13915984028: When should update, then DeepCopy it, put behind the following check. thanks.

w13915984028 (approved):
LGTM, thanks.

guangbochen (approved):
LGTM, please squash the commits before the PR merge; thanks.

futuretea (approved):
LGTM, thanks.

WebberHuang1118 (approved):
LGTM, thanks

guangbochen:
@mergify backport v1.2

chrisho:
@mergify backport v1.1

PR #4106, opened 2023-06-19
ntp/settings: do not add duplicated ntp server to nodeconfig

Vicente-Cheng:

**Problem:**
We should avoid adding duplicated NTP servers.

**Solution:**
Both frontend and backend need to add the related handle.

**Related Issue:**
Related: https://github.com/harvester/harvester/issues/3133#issuecomment-1595486084

**Test plan:**
TBD


pkg/controller/master/setting/ntp_server.go
```diff
@@ -42,6 +43,7 @@ func (h *Handler) syncNTPServer(setting *harvesterv1.Setting) error {
 				return err
 			}
 
+			ntpServers := reGenerateNTPServers(ntpSettings, []string{})
```
bk201: How about just refactoring `util.GetNTPServers`? This also doesn't make `util.GetNTPServers` an unused function. We can also consider moving the call out of the loop so it only run once.

Vicente-Cheng: Move to utils.
Also, reduce the duplicated re-generate operation when we do not have nodeconfig.

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM!

bk201:
@mergifyio backport v1.2

Issue #3867, opened 2023-05-08
[ENHANCEMENT] Changing storage network setting needs to consider user workload


bk201:
**Is your enhancement related to a problem? Please describe.**

We prohibit setting storage network settings when there is a running VM ([code](https://github.com/harvester/harvester/blob/b2de0f4b67b39f9a96b4cc86564ef71f93326f7f/pkg/webhook/resources/setting/validator.go#L640)).
We should also consider user workloads in v1.2.0.

**Describe the solution you'd like**

Check if there are any non-system attached volumes.

**Describe alternatives you've considered**

**Additional context**

irishgordo:
Tested on Harvester v1.2.0-rc3
This looks good :smile: :+1: , opened up #4217 - for an item discovered.
Thanks @WebberHuang1118 
![Screenshot from 2023-07-10 15-25-51](https://github.com/harvester/harvester/assets/5370752/7f3580ba-7bed-4d51-ba9c-0ebb964d5ed3)
![Screenshot from 2023-07-10 14-34-08](https://github.com/harvester/harvester/assets/5370752/75a1c028-2056-4897-a7e2-16ca71341578)
![Screenshot from 2023-07-10 14-31-08](https://github.com/harvester/harvester/assets/5370752/3abb9429-da08-4aae-9bcc-23a6d7c6ed49)
![Screenshot from 2023-07-10 14-22-53](https://github.com/harvester/harvester/assets/5370752/11165b6d-b2a5-4778-ba6f-8cad5fc13741)

Was able to see the ds/harvester-network-controller on harvester-system logs generate:
```yaml
I0710 21:31:42.777306       1 controller.go:46] nad configuration harvester-system/storagenetwork-2f52q has been changed: {"cniVersion":"0.3.1","type":"bridge","dns":{},"bridge":"additional-br","promiscMode":true,"vlan":1,"ipam":{"type":"whereabouts","range":"172.16.0.0/24","exclude":["172.168.0.1/24","172.168.0.1/24"]}}
I0710 21:31:42.789690       1 controller.go:46] nad configuration harvester-system/storagenetwork-spg8p has been changed: {"cniVersion":"0.3.1","type":"bridge","dns":{},"bridge":"additional-br","promiscMode":true,"vlan":1,"ipam":{"type":"whereabouts","range":"172.16.0.0/24","exclude":["172.168.0.1/24","172.168.0.1/24"]}}
[...]

PR #4124, opened 2023-06-21
Checking if any user workloads before making storage network change

WebberHuang1118:
Before making changes for storage network, webhook would check if there are any attached volumes other than system volume. If yes, webhook should prohibit the change from happening

**Problem:**

Once there is any attached volume (other than system volumes) accessed by the user workload, applying storage network changing would got stuck

**Solution:**

Webhook would first collect the system volumes list, and check if there is any attached volume not in such list. If yes, webhook should prohibit the change from happening

**Related Issue:**
#3867 

**Test plan:**
1. Apply the following yaml by `$kubectl apply -f pod_with_pvc.yaml` to create a pvc and user workload
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: longhorn-volv-pvc
  namespace: default
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: longhorn
  resources:
    requests:
      storage: 2Gi
---
apiVersion: v1
kind: Pod
metadata:
  name: volume-test
  namespace: default
spec:
  restartPolicy: Always
  containers:
  - name: volume-test
    image: nginx:stable-alpine
    imagePullPolicy: IfNotPresent
    livenessProbe:
      exec:
        command:
          - ls
          - /data/lost+found
      initialDelaySeconds: 5
      periodSeconds: 5
    volumeMounts:
    - name: volv
      mountPath: /data
    ports:
    - containerPort: 80
  volumes:
  - name: volv
    persistentVolumeClaim:
      claimName: longhorn-volv-pvc
```
2. Enable Addon -> rancher-logging and rancher-monitoring
3. Enable Addon -> vm-import-controllercontroller with reasonable volume size vm_importer
4. Confirm there are 4 system volumes and 1 user workload volume
    - System volumes: Prometheus volume, AlertManager volume, Grafana volume and VM importer volumevolumes
5. Apply storage network setting storage_network
6. We should find the operation is prohibited by Webhook with the message `please stop all workloads before configuring the storage-network setting` storage_network_webhook
7. Delete user workload and related pvc by `$kubectl delete -f pod_with_pvc.yaml` 
8. Apply storage network configuration again, the change would successfully applied


pkg/util/storage_network.go
```diff
@@ -0,0 +1,25 @@
...
+
+	RancherMonitoringPrometheus        = "rancher-monitoring-prometheus"
+	RancherMonitoringPrometheusSs      = "prometheus-" + RancherMonitoringPrometheus
+	RancherMonitoringPrometheusPvcPref = RancherMonitoringPrometheusSs + "-db-" + RancherMonitoringPrometheusSs + "-"
+
+	RancherMonitoringAlertmanager        = "rancher-monitoring-alertmanager"
+	RancherMonitoringAlertmanagerSs      = "alertmanager-" + RancherMonitoringAlertmanager
+	RancherMonitoringAlertmanagerPvcPref = RancherMonitoringAlertmanagerSs + "-db-" + RancherMonitoringAlertmanagerSs + "-"
+
+	FleetLocalNamespace = "fleet-local"
```
futuretea: already has `FleetLocalNamespaceName` in pkg/util/constants.go

pkg/util/storage_network.go
```diff
@@ -0,0 +1,25 @@
...
+	RancherMonitoring   = "rancher-monitoring"
+
+	RancherMonitoringGrafana    = "rancher-monitoring-grafana"
+	RancherMonitoringGrafanaPvc = RancherMonitoringGrafana
+
+	// VM import controller
+	HarvesterSystemNamespace       = "harvester-system"
```
futuretea: already has `HarvesterSystemNamespaceName` in pkg/util/constants.go

pkg/util/storage_network.go
```diff
@@ -0,0 +1,25 @@
+package util
+
+const (
+	// Rancher monitoring
+	CattleMonitoringSystemNamespace = "cattle-monitoring-system"
+
+	RancherMonitoringPrometheus        = "rancher-monitoring-prometheus"
+	RancherMonitoringPrometheusSs      = "prometheus-" + RancherMonitoringPrometheus
+	RancherMonitoringPrometheusPvcPref = RancherMonitoringPrometheusSs + "-db-" + RancherMonitoringPrometheusSs + "-"
```
futuretea: s/Pvc/PVC/g

pkg/util/storage_network.go
```diff
@@ -0,0 +1,25 @@
+package util
+
+const (
+	// Rancher monitoring
+	CattleMonitoringSystemNamespace = "cattle-monitoring-system"
+
+	RancherMonitoringPrometheus        = "rancher-monitoring-prometheus"
+	RancherMonitoringPrometheusSs      = "prometheus-" + RancherMonitoringPrometheus
```
futuretea: What is the meaning of `Ss`?

pkg/webhook/resources/setting/validator.go
```diff
@@ -35,6 +39,8 @@ import (
 	storagenetworkctl "github.com/harvester/harvester/pkg/controller/master/storagenetwork"
 	ctlv1beta1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	ctllhv1b2 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
+	monitoringv1 "github.com/harvester/harvester/pkg/generated/controllers/monitoring.coreos.com/v1"
```
futuretea: ctlmonitoringv1

pkg/webhook/resources/setting/validator.go
```diff
@@ -19,12 +19,16 @@ import (
 	// S3 Ref: https://github.com/longhorn/backupstore/blob/3912081eb7c5708f0027ebbb0da4934537eb9d72/s3/s3.go#L33-L37
 	_ "github.com/longhorn/backupstore/nfs" //nolint
 	_ "github.com/longhorn/backupstore/s3"  //nolint
+	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	"github.com/rancher/wharfie/pkg/registries"
+	appsv1 "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
```
futuretea: ctlappsv1

pkg/util/storage_network.go
```diff
@@ -0,0 +1,25 @@
...
+	RancherMonitoringPrometheusPvcPref = RancherMonitoringPrometheusSs + "-db-" + RancherMonitoringPrometheusSs + "-"
+
+	RancherMonitoringAlertmanager        = "rancher-monitoring-alertmanager"
+	RancherMonitoringAlertmanagerSs      = "alertmanager-" + RancherMonitoringAlertmanager
+	RancherMonitoringAlertmanagerPvcPref = RancherMonitoringAlertmanagerSs + "-db-" + RancherMonitoringAlertmanagerSs + "-"
```
futuretea: PVCPrefix

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +730,201 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+		systemVols[vmImportControllerVol] = struct{}{}
+	}
+
+	return systemVols, nil
+}
+
+func (v *settingValidator) checkOnlineVolume() error {
+	systemVols, err := v.getSystemVols()
+	if err != nil {
+		logrus.Infof("getSystemPvcs err:%v", err)
```
futuretea: s/Pvc/PVC/g

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +730,201 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+		return err
+	}
+
+	for _, vol := range volumes {
+		if _, found := systemVols[vol.Name]; found {
+			continue
+		}
+
+		if vol.Status.State != lhv1beta2.VolumeStateDetached {
+			logrus.Infof("vol %v in state %v", vol.Name, vol.Status.State)
```
futuretea: s/vol/volume/g

pkg/util/storage_network.go
WebberHuang1118: @futuretea may I know the meaning of `s/Pvc/PVC/g`?  Using PVC instead of Pvc? thanks

pkg/util/storage_network.go
WebberHuang1118: It means stateful set, recommend to use the full words?

futuretea: sts means statefulset.
```bash
kubectl get sts
```

pkg/util/storage_network.go
futuretea: replace Pvc to PVC globally

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +730,201 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
 }
 
+func (v *settingValidator) getPrometheusVols() ([]string, error) {
+	prometheus, err := v.prometheusCache.Get(util.CattleMonitoringSystemNamespace, util.RancherMonitoringPrometheus)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logrus.Infof("prometheus not found. skip")
+			return nil, nil
+		}
+		return nil, fmt.Errorf("prometheus get error %v", err)
```
yaocw2020: Please use %w instead of %v to wrap error. Check out other places.

https://go.dev/blog/go1.13-errors

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +730,201 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
 	return werror.NewMethodNotAllowed(fmt.Sprintf("Disallow delete setting name %s", settings.StorageNetworkName))
 }
 
+func (v *settingValidator) getPrometheusVols() ([]string, error) {
```
yaocw2020: It seems all the PVCs used by prometheus have a common label `app.kubernetes.io/managed-by=prometheus-operator`, so why not list by selector to get them? For example,
```
harvester:/home/rancher # kubectl get pvc -A --selector app.kubernetes.io/managed-by=prometheus-operator
NAMESPACE                  NAME                                                                                             STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS         AGE
cattle-monitoring-system   alertmanager-rancher-monitoring-alertmanager-db-alertmanager-rancher-monitoring-alertmanager-0   Bound    pvc-d3f2a72d-5fa6-4c0a-bf21-4b4be587019e   5Gi        RWO            harvester-longhorn   124d
cattle-monitoring-system   prometheus-rancher-monitoring-prometheus-db-prometheus-rancher-monitoring-prometheus-0           Bound    pvc-5f192ee3-693e-483f-97bf-d147eeacd413   50Gi       RWO            harvester-longhorn   124d

```

I think this way would be easier.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +730,201 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+			continue
+		}
+
+		prometheusVols = append(prometheusVols, pvc.Spec.VolumeName)
+	}
+
+	return prometheusVols, nil
+}
+
+func (v *settingValidator) getAlertManagerVols() ([]string, error) {
```
yaocw2020: AlertManager is a part of promethues. Its PVC is includes in PVCs listed by selector `app.kubernetes.io/managed-by=prometheus-operator`.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +730,201 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+			continue
+		}
+
+		alertmanagerVols = append(alertmanagerVols, pvc.Spec.VolumeName)
+	}
+
+	return alertmanagerVols, nil
+}
+
+func (v *settingValidator) getGrafanaVol() (string, error) {
```
yaocw2020: Listing PVCs directly by selector would be effetive for the grafana.

pkg/util/storage_network.go
WebberHuang1118: Fixed, thanks.

pkg/util/storage_network.go
WebberHuang1118: Fixed, thanks.

pkg/util/storage_network.go
WebberHuang1118: Fixed, thanks.

pkg/util/storage_network.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/util/storage_network.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Since for the PVC used by prometheus, its name is determined. So we can directly check PVC cache for its existence. And in case the PVC is manually created with the same name as what prometheus doses, I think we should check prometheus CR first. Thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: The same idea as https://github.com/harvester/harvester/pull/4124#discussion_r1250568879. Thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Since the name for Grafana PVC is determined, directly checking its existence wouldn't be more effective than listing all PVCs with selector? Thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +730,201 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+		systemVolumes[vmImportControllerVolume] = struct{}{}
+	}
+
+	return systemVolumes, nil
+}
+
+func (v *settingValidator) checkOnlineVolume() error {
+	systemVolumes, err := v.getSystemVolumes()
+	if err != nil {
+		logrus.Infof("getSystemPVCs err:%v", err)
```
bk201:
```suggestion
		logrus.Infof("getSystemPVCs err: %v", err)
```
nit.

WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +730,201 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+		}
+		return nil, fmt.Errorf("alertmanager get error %w", err)
+	}
+
+	if alertmanager.Spec.Replicas == nil {
+		return nil, nil
+	}
+
+	alertmanagerVolumes := make([]string, 0)
+	for i := 0; i < int(*alertmanager.Spec.Replicas); i++ {
```
bk201: Can you see if we have a better way to find alert manager volume? maybe via owner or labels. Thanks.

WebberHuang1118: As @yaocw2020 suggested https://github.com/harvester/harvester/pull/4124#discussion_r1250203133,
I think we could try to list PVCs with selector, thanks.

pkg/webhook/clients/clients.go
```diff
@@ -6,7 +6,7 @@ import (
 	ctlfleetv1 "github.com/rancher/rancher/pkg/generated/controllers/fleet.cattle.io"
 	rancherv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io"
 	"github.com/rancher/wrangler/pkg/clients"
-	corev1 "github.com/rancher/wrangler/pkg/generated/controllers/core"
+	"github.com/rancher/wrangler/pkg/generated/controllers/core"
```
futuretea: ctrlcorev1

pkg/util/storage_network.go
```diff
@@ -0,0 +1,17 @@
...
+	RancherMonitoringPrometheus = "rancher-monitoring-prometheus"
+
+	RancherMonitoringAlertmanager = "rancher-monitoring-alertmanager"
+
+	RancherMonitoring = "rancher-monitoring"
+
+	RancherMonitoringGrafana = "rancher-monitoring-grafana"
+
+	// VM import controller
+	HarvesterVMImportController = "vm-import-controller-harvester-vm-import-controller"
```
Vicente-Cheng: Typo here?
Should this string be `harvester-vm-import-controller`?

pkg/webhook/clients/clients.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,159 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+		systemVolumes[vmImportControllerVolume] = struct{}{}
+	}
+
+	return systemVolumes, nil
+}
+
+func (v *settingValidator) checkOnlineVolume() error {
+	systemVolumes, err := v.getSystemVolumes()
+	if err != nil {
+		logrus.Infof("getSystemPVCs err: %v", err)
```
Vicente-Cheng: s/Infof/Errorf/

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,159 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+		return err
+	}
+
+	for volume := range systemVolumes {
+		logrus.Infof("get getSystem volume %v", volume)
+	}
+
+	volumes, err := v.lhVolumeCache.List(util.LonghornSystemNamespaceName, labels.Everything())
+	if err != nil {
+		logrus.Infof("list longhorn volumes err %v", err)
```
Vicente-Cheng: ditto

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,159 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+		return err
+	}
+
+	for _, volume := range volumes {
+		if _, found := systemVolumes[volume.Name]; found {
+			continue
+		}
+
+		if volume.Status.State != lhv1beta2.VolumeStateDetached {
+			logrus.Infof("volume %v in state %v", volume.Name, volume.Status.State)
```
Vicente-Cheng: ditto

pkg/util/storage_network.go
WebberHuang1118: No, it's the actual name for the depolyment

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: https://github.com/WebberHuang1118/harvester/blob/121f387ad0ce6c83bfde536aeabd02bc6a02663b/pkg/webhook/resources/setting/validator.go#L726
Listing PVCs used by Prometheus via labels and selectors,  thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: https://github.com/WebberHuang1118/harvester/blob/121f387ad0ce6c83bfde536aeabd02bc6a02663b/pkg/webhook/resources/setting/validator.go#L748
Listing PVCs used by AlertManager via labels and selectors, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: https://github.com/WebberHuang1118/harvester/blob/121f387ad0ce6c83bfde536aeabd02bc6a02663b/pkg/webhook/resources/setting/validator.go#L770
Listing PVCs used by Grafana via labels and selectors, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: https://github.com/WebberHuang1118/harvester/blob/121f387ad0ce6c83bfde536aeabd02bc6a02663b/pkg/webhook/resources/setting/validator.go#L748
Listing PVCs used by AlertManager via labels and selectors, thanks.

w13915984028:
Just FYI:

If we are planning to backport to v1.1.2/v1.1.3, then please consider:

In v1.1.2/v1.1.3, the `rancher-monitoring` is a `managedchart` defined per https://github.com/harvester/harvester-installer/blob/v1.1/pkg/config/templates/rancherd-13-monitoring.yaml,  the `alertmanager` can be disabled on the fly, but `prometheus` and `grafana` can't.

It will mean those PVC/PV are always there, and it will block setting storagenetwork when this commit is merged.

---

v1.2.0 does not have this limitation, those are all `addons` and they can be disabled/enabled freely.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,159 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+
+	if len(pvcs) == 0 {
+		return "", nil
+	}
+
+	//grafana deployment only has one pvc
+	return pvcs[0].Spec.VolumeName, nil
+}
+
+func (v *settingValidator) getVMImportyControllerVolume() (string, error) {
```
yaocw2020: getVMImportyControllerVolume => getVMImportControllerVolume

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,159 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+	grafanaVolume, err := v.getGrafanaVolume()
+	if err != nil {
+		return nil, err
+	}
+
+	if grafanaVolume != "" {
+		systemVolumes[grafanaVolume] = struct{}{}
+	}
+
+	vmImportControllerVolume, err := v.getVMImportyControllerVolume()
```
yaocw2020: ditto

pkg/util/storage_network.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+const (
```
guangbochen: define those constants on the `util/constants.yaml` file is preferred.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,159 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+func (v *settingValidator) checkOnlineVolume() error {
+	systemVolumes, err := v.getSystemVolumes()
+	if err != nil {
+		logrus.Errorf("getSystemPVCs err: %v", err)
+		return err
+	}
+
+	for volume := range systemVolumes {
+		logrus.Infof("get getSystem volume %v", volume)
+	}
```
guangbochen: why do we need to loop it just for print?

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Fixed, thanks.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,159 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+
+	volumes, err := v.lhVolumeCache.List(util.LonghornSystemNamespaceName, labels.Everything())
+	if err != nil {
+		logrus.Errorf("list longhorn volumes err %v", err)
+		return err
+	}
```
guangbochen, line 800: this would be a cumbersome call for a large cluster since all ns volumes will be listed here, would prefer to either filter it by labels or get volumes one by one since there won't be too many `systemVolumes`.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,159 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+
+	for _, pvc := range pvcs {
+		if pvc.Spec.VolumeName == "" {
+			continue
+		}
+
+		prometheusVolumes = append(prometheusVolumes, pvc.Spec.VolumeName)
+	}
+
+	return prometheusVolumes, nil
```
guangbochen: Those are redundant codes, prefer to create a common method to pass the label and get PVCs result accordingly.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Just for logging information, if you think it's unnecessary I can remove it, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: It's for finding the online Longhorn volumes other than system volumes, so it seems needed to check all Longhorn volumes. Thanks.

pkg/webhook/resources/setting/validator.go
guangbochen: yes please or you can print the entire list using debug level log, thanks.

pkg/webhook/resources/setting/validator.go
guangbochen: I know we are looking for the LH volumes, and since the volume name is identical to the pv name, so instead of listing and looping the entire volume list, getting volume by pv and returning directly when it is attached is preferred from my point of view, thanks.

WebberHuang1118: But even without looping longhorn volumes, we still need to iterate all pvs and check the status of related longhorm volume. Do I misunderstand? thanks.

guangbochen: yes, but not iterate all pvc, only those system volumes that u want to check if I understand the code correctly.

WebberHuang1118: Actually, we want to check the status of all non-system volumes here, that's why it requires looping over PVs or Longhorn volumes. However, maybe we could consider iterating Longhorn volumes with pagination, I had an implementation with pagination and it works well so far. With pagination, it retrieves limited numbers of Longhorn volumes at a time, and if there is any non-system volume that meets certain conditions, we can break out of the loop. So we don't have to list all the Longhorn volumes most of the time. I will update this implementation with other fixes,  thanks.

guangbochen: I see, if it is about to check all non-system volume, then this LGTM, thanks.

pkg/util/storage_network.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Refactor per suggestion, thanks.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,133 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+	volumes, err := v.findVolumesByLabelAppName(util.CattleMonitoringSystemNamespace, labelAppNameValueGrafana)
+	if err != nil {
+		return "", err
+	}
+
+	if len(volumes) == 0 {
+		return "", nil
+	}
+
+	//grafana deployment only has one pvc
```
bk201: I'd suggest not assuming this. Considering the whole returned slice makes the code more robust.

pkg/webhook/resources/setting/validator.go
```diff
@@ -709,6 +723,133 @@ func (v *settingValidator) validateDeleteStorageNetwork(setting *v1beta1.Setting
...
+	volumes, err := v.findVolumesByLabelAppName(util.HarvesterSystemNamespaceName, labelAppNameValueImportController)
+	if err != nil {
+		return "", err
+	}
+
+	if len(volumes) == 0 {
+		return "", nil
+	}
+
+	//VM import contoller deployment only has one pvc
```
bk201: ditto

yaocw2020 (approved):
LGTM, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Refactor per suggestion, thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Refactor per suggestion, thanks.

guangbochen (approved):
LGTM, and please squash the commits before merge, thanks.

futuretea (approved):
LGTM, thanks.

Vicente-Cheng (approved):
LGTM!

bk201 (approved):
LGTM, thanks!

WebberHuang1118:
@Mergifyio backport v1.2

Issue #4118, opened 2023-06-20
[ENHANCEMENT] Try best to keep grafana PV when rancher-monitoring addon is disabled


w13915984028:
**Is your enhancement related to a problem? Please describe.**

The user may disable `rancher-monitoring` temporarily and enable it later. The potential customization in Grafana will be lost, as the PVC&PV are removed when addon is disabled.

**Describe the solution you'd like**

Try best to keep grafana PV when rancher-monitoring addon is disabled

**Describe alternatives you've considered**

Add `Caution` in the Harvester doc.

**Additional context**

The effort in upgrade to keep grafana pv:
Issue: https://github.com/harvester/harvester/issues/4071
PR: https://github.com/harvester/harvester/pull/4117

The effort to enhance the `disable` processing of addon, which will be the basis of this issue.
Issue https://github.com/harvester/harvester/issues/4108
PR https://github.com/harvester/harvester/pull/4110

lanfon72:
Verified this bug has been fixed.


Test Information
----
* Environment: **qemu/KVM 3 nodes**
* Harvester Version: **v1.2.0-rc3** (upgrade from **v1.1.2**)
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Login to dashboard, access embedded Rancher UI
3. Navigate to _Storage/PersistentVolumes_, the PV for PVC `rancher-monitoring-grafana` should be `Reclaim Policy: Retain`
4. In Harvester UI, navigate to _Advanced/Addons_, disable `rancher-monitoring`
5. Addon `rancher-monitoring` should be disabled successfully
6. In Rancher UI, navigate to _Storage/Volumes_, Volume `rancher-monitoring-grafana` should be removed
7. Navigate to _Storage/PersistentVolumes_, the PV with `Reclaim Policy: Retain` should not have PVC
8. In Harvester UI, navigate to _Advanced/Addons_, enable `rancher-monitoring`
9. In Rancher UI, navigate to _Storage/PersistentVolumes_, PVC `rancher-monitoring-grafana` should attached to the PV with `Reclaim Policy: Retain`
10. Access Grafana dashboard from **rancher-monitoring** addon
11. Panels should display normally and contain old data

PR #4110, opened 2023-06-20
additional logic to track addon while it is being disabled

ibrokethecloud:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**

When an Addon is disabled, it should become a new disabling status, and if user click enable again, the action should be denied, to make sure the enable will wait until the previous disable action is finished. To avoid potential race condition.

**Solution:**

Introduction of a DisablingAddon state, and additional validation logic in the harvester webhook.

**Related Issue:**
https://github.com/harvester/harvester/issues/4108

**Test plan:**

Issue #4108, opened 2023-06-19
[ENHANCEMENT] Harvester Addon needs to improve the processing of disable


w13915984028:
**Is your enhancement related to a problem? Please describe.**

Harvester addon needs to improve the processing of disable.

For large `addon` like `rancher-monitoring`, the `enable` and `disable` will take some time to finish the action

**Describe the solution you'd like**

When an `Addon` is disabled, it should become a new `disabling` status, and if user click `enable` again, the action should be denied, to make sure the `enable` will wait until the previous `disable` action is finished. To avoid potential race condition.

The `enable` -> `disable` will also be double checked.

**Describe alternatives you've considered**

**Additional context**

issue https://github.com/harvester/harvester/issues/4125 will also be solved in this enhancement,

In the enhancement of https://github.com/harvester/harvester/issues/4071, we need to wait the `addon` is disabled, and then remove the reference of PV toward PVC

cc @guangbochen @ibrokethecloud

w13915984028:
Encounter such issue:

disable/enable `rancher-monitoring` for first time, it is OK.

delete Harvester POD

disable/enable `rancher-monitoring` for second  time

(1)  Disable `rancher-monitoring`, but UI still shows `DeploySuccessful`

![image](https://github.com/harvester/harvester/assets/31133476/dcfc9743-63b5-4c66-ab6c-c51de1ec8019)

But the status is `""` .
```
harv41:~ # kk get addons.harvesterhci.io -n cattle-monitoring-system rancher-monitoring -ojsonpath={".status"}
{}
```

Similar with https://github.com/harvester/harvester/issues/4125

(2) The `delete job` was there for a short while
```
harv41:~ # kk get jobs -A
NAMESPACE                  NAME                               COMPLETIONS   DURATION   AGE
cattle-monitoring-system   helm-delete-rancher-monitoring     1/1           4s         4m54s
cattle-monitoring-system   helm-install-rancher-monitoring    1/1           40s        2m21s
```

(3) The `jobs` are gone, both `helm-install` and `helm-delete`
```
harv41:~ # kk get jobs -A
NAMESPACE     NAME                               COMPLETIONS   DURATION   AGE
kube-system   helm-install-rke2-canal            1/1           10s        19d
```

(4) All pods of `rancher-monitoring` are left untouched
```
harv41:~ # kk get pods -n cattle-monitoring-system
NAME                                                     READY   STATUS    RESTARTS        AGE
[...]

PR #4127, opened 2023-06-21
retain grafana PV when addon is disabled and refactor addon

w13915984028:
```

    retain grafana PV when addon is disabled and refactor addon
    
      refactor code to run per status, user operation will run in serial, one after another
      add validator to deny request when operation is ongoing
      add mutator to record user operation and timestamp
      add conditions to status to record more
      add test code
      add timeout
      add workaround to fix upstream helm-controller issue
      sync integration test code
    
    Signed-off-by: Jian Wang <w13915984028@gmail.com>
    
    Co-authored-by: Gaurav Mehta <gaurav@rancher.com>


```

joined with https://github.com/harvester/harvester/pull/4110 

TODO: Code will be rebased when PR https://github.com/harvester/harvester/pull/4163 is merged.

**Problem:**

Grafana PV is removed when addon is disabled.
Frequent enable/disable addon will cause addon into unknown status.

**Solution:**

Refactor addon, add status control; retain grafana PV

Add status machine:
(1) User operation is controlled to be in serial, any status in (2) will block (1)
(2) AddonEnabled, AddonDisabling, AddonUpdating  will run until finishi/fail/ timeout (TBD)
(3) For other status, will check (4) (5)
(4) If an valid user operation is there (via check annotations), and it is newer than the target (enable/disable/update) conditions' lastUpdateTime, then move status to (2)
(5) Based on Addons.Spec.Enabled to check helm chart, check if identical, then move status to (2) when necessary.

**Related Issue:**
#4118
#4108

**Test plan:**

test plan:

A. Addon status processing
(1) Normal enable/disable/update(when enabled) addon, it should work as normal
(2) When an operation is ongoing, try another operation, it will be denied by webhook
(3) Enable non-existing addon version, it will fail with timeout
     e.g., for `harvester-pcidevices-controller`, change the version from v0.2.5-rc1 to v0.2.6-rc1 or other version
     enable the addon, it will finally fail due to timeout, by default, the timeout is 15 minutes,
     
     to speed up the validation, you can set such an annotation, which change the timeout to 4 minutes
     harvesterhci.io/addon-operation-timeout: "4"
(4) If an addon like (3) is enabling and waiting for timeout, e.g., it has already been more than 5 minutes
    when change the annotation to 4 minutes, it will fail right now
(5) Disable the `failed` addon, it should success

B. Retain grafana PV
(1) Enable rancher-monitoring unitil is is deployed successfully
      there will be 3 PVC & 3 PV in rancher-monitoring, including grafana-pv, grafana PV has `RECLAIM POLICY` `Retain`

```
harv41:~ # kk get pv -A
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                                                                                                                     STORAGECLASS         REASON   AGE
pvc-81529c97-dd6e-424a-a3ca-705999755293   50Gi       RWO            Delete           Bound    cattle-monitoring-system/prometheus-rancher-monitoring-prometheus-db-prometheus-rancher-monitoring-prometheus-0           harvester-longhorn            7d14h
pvc-a1313e39-a7bb-4c16-8996-7b2f8f3a4800   5Gi        RWO            Delete           Bound    cattle-monitoring-system/alertmanager-rancher-monitoring-alertmanager-db-alertmanager-rancher-monitoring-alertmanager-0   harvester-longhorn            7d14h
pvc-a95cdf6b-26bc-4ed6-8aaf-5f5d6b0b6dbd   2Gi        RWO            Retain           Bound    cattle-monitoring-system/rancher-monitoring-grafana                                                                       harvester-longhorn            7d14h
```
      
(2) Disable rancher-monitoring until it is disabled
      the grafana related PVC will disappear; but grafana pv is there, and `STATUS` is `Available`
```
pvc-e2983b67-035a-485e-bb44-fbf6309389cb   2Gi        RWO            Retain           Available   cattle-monitoring-system/rancher-monitoring-grafana                                                                       harvester-longhorn            32d
```      
(3) Enable rancher-monitoring unitil is is deployed successfully
      the grafana pvc is back, and the PV is bounded to this new PVC, `STATUS`  is `Bound`


pkg/apis/harvesterhci.io/v1beta1/addon.go
```diff
@@ -10,6 +10,9 @@ const (
 	AddonEnabled  AddonState = "AddonEnabled"
 	AddonDeployed AddonState = "AddonDeploySuccessful"
 	AddonFailed   AddonState = "AddonDeployFailed"
+	//	AddonDisabled AddonState = "AddonDisabled"
+	AddonDisabling    AddonState = "AddonDisabling"
+	AddonDisableFaild AddonState = "AddonDisableFailed"
```
ibrokethecloud:
```suggestion
	AddonDisableFailed AddonState = "AddonDisableFailed"
```

pkg/controller/master/addon/addon.go
```diff
@@ -109,47 +121,191 @@ func (h *Handler) MonitorChanges(key string, aObj *harvesterv1.Addon) (*harveste
...
+	if strings.HasPrefix(hc.Status.JobName, "helm-delete-") {
+		logrus.Infof("the fetched job %s is from last time helmdelete, wait install job", hc.Status.JobName)
+		h.addon.EnqueueAfter(aObj.Namespace, aObj.Name, 5*time.Second)
+		return aObj, nil
+	}
+
+	job, err := h.job.Get(aObj.Namespace, hc.Status.JobName, metav1.GetOptions{})
 	if err != nil {
-		return a, fmt.Errorf("error fetching job %s: %v", job.Name, err)
+		logrus.Infof("waiting for job %s to start", job.Name)
```
ibrokethecloud: we should refine this check to requeue if job doesnt exist, and error if there was an actual error interacting with the apiserver

pkg/controller/master/addon/addon.go
```diff
@@ -109,47 +121,191 @@ func (h *Handler) MonitorChanges(key string, aObj *harvesterv1.Addon) (*harveste
...
+	}
+
+	hc, err := h.helm.Cache().Get(aObj.Namespace, aObj.Name)
+	if err != nil {
+		// disabled
+		if apierrors.IsNotFound(err) {
+			if aObj.Status.Status == "" {
+				logrus.Infof("addon %s: helm chart is gone, addon has already been in init status", aObj.Name)
+				return aObj, nil
+			}
```
ibrokethecloud: if `aObj.Status.Status == ""` then this condition will never be let since in L204-205 we already return check for this condition.

pkg/controller/master/addon/addon.go
```diff
@@ -361,3 +542,133 @@ func (h *Handler) PatchApps(key string, app *catalogv1.App) (*catalogv1.App, err
...
+	}
+
+	a := aObj.DeepCopy()
+	if a.Annotations == nil {
+		a.Annotations = make(map[string]string, 1)
+	}
+
+	a.Annotations[util.AnnotationGrafanaPVName] = pvname
+
+	return h.addon.Update(a)
```
ibrokethecloud: we could possibly simplify this as follows
```suggestion
    a := aObj.DeepCopy
	if aObj.Annotations != nil {
		a.Annotations = make(map[string]string, 1)
		}
	}

	a.Annotations[util.AnnotationGrafanaPVName] = pvname

    if !reflect.DeepEqual(aObj, a) {
        return h.addon.Update(a)
    }
	return aObj, nil
```

pkg/controller/master/addon/addon.go
```diff
@@ -282,14 +447,30 @@ func (h *Handler) ReconcileHelmChartOwners(_ string, _ string, obj runtime.Objec
...
-	return h.addon.UpdateStatus(a)
+	// delete or deleting, make sure status is AddonDisabling
+	if d {
+		if aObj.Status.Status != harvesterv1.AddonDisabling {
+			a := aObj.DeepCopy()
+
+			logrus.Infof("addon %s is being disabled now, update status from %s to %s", a.Name, a.Status.Status, harvesterv1.AddonDisabling)
+			a.Status.Status = harvesterv1.AddonDisabling
+			return h.addon.UpdateStatus(a)
+		}
```
ibrokethecloud, line 492: we should add more checks here.. 

if status is already Addon.Disabling, then helmChart object will have a new `helm-delete` job submitted. we should query the job status for the same, and accordingly update the addon status to `AddonDisableFailed` or clear status if job was successful.

ibrokethecloud:
we could just use this as the pr for addon enhancements.

The addition check needed to query delete job can be copied from the other [PR](https://github.com/harvester/harvester/pull/4110)

we can also copy the validating webhook changes to avoid disable/enable when another operation is already in progress.

w13915984028:
**solution: move the `type` definition away from uitl package.** 
---
The drone build complains
```
Running validation
Running: golangci-lint run
pkg/controller/master/addon/addon.go:474:84: AddonOperationRecord not declared by package util (typecheck)
func (h *Handler) deduceAddonOperation(aObj *harvesterv1.Addon, aor *harvesterutil.AddonOperationRecord) (*harvesterv1.Addon, error) {
                                                                                   ^
pkg/controller/master/addon/common.go:131:91: AddonOperationRecord not declared by package util (typecheck)
func (h *Handler) getUserOperationFromAnnotations(aObj *harvesterv1.Addon) *harvesterutil.AddonOperationRecord {
                                                                                          ^
```

but int local `make`, the `golangci-lint run` does not report such error.

An old issue link: https://github.com/golangci/golangci-lint/issues/658.

Harvester is using v1.49.0 of golangci-lint, a bit outdate. https://github.com/golangci/golangci-lint/releases/tag/v1.49.0

I tried the second latest version https://github.com/golangci/golangci-lint/releases/tag/v1.50.0, it reports more :)
```
Running: golangci-lint run
pkg/webhook/types/validator.go:57:35: unused-parameter: parameter 'request' seems to be unused, consider removing or renaming it as _ (revive)
func (v *DefaultValidator) Create(request *Request, newObj runtime.Object) error {
                                  ^
pkg/webhook/types/validator.go:61:35: unused-parameter: parameter 'request' seems to be unused, consider removing or renaming it as _ (revive)
func (v *DefaultValidator) Update(request *Request, oldObj runtime.Object, newObj runtime.Object) error {
                                  ^
pkg/webhook/types/validator.go:65:35: unused-parameter: parameter 'request' seems to be unused, consider removing or renaming it as _ (revive)
func (v *DefaultValidator) Delete(request *Request, oldObj runtime.Object) error {
                                  ^
pkg/util/crd/init.go:142:66: unused-parameter: parameter 'ready' seems to be unused, consider removing or renaming it as _ (revive)
func (f *Factory) createCRD(ctx context.Context, crdDef crd.CRD, ready map[string]*apiext.CustomResourceDefinition) (*apiext.CustomResourceDefinition, error) {
```

pkg/webhook/resources/addon/validator.go
```diff
@@ -83,12 +83,35 @@ func validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) erro
...
 	if newAddon.Name == vClusterAddonName && newAddon.Namespace == vClusterAddonNamespace && newAddon.Spec.Enabled {
 		return validateVClusterAddon(newAddon)
 	}
 	return nil
 }
 
+func isAddonSpecChanged(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) bool {
+	if newAddon.Spec != oldAddon.Spec {
+		return true
+	}
```
chrisho, line 108: prefer to use `reflect.DeepEqual` instead.
If not any other check, we could simplify it like this
```go
func isAddonSpecChanged(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) bool {
	return !reflect.DeepEqual(newAddon.Spec, oldAddon.Spec)
}
```

pkg/apis/harvesterhci.io/v1beta1/addon.go
```diff
@@ -10,6 +11,33 @@ const (
...
+	AddonInitState AddonState = "" // init status, when an addon is not enabled, or disabled successfully
+)
+
+type AddonOperation string
+
+const (
+	AddonUpdateOperation  AddonOperation = "update"
+	AddonEnableOperation  AddonOperation = "enable"
+	AddonDisableOperation AddonOperation = "disable"
+	AddonNullOperation    AddonOperation = "null"
```
guangbochen, line 33: what is null operation stands for?

pkg/webhook/resources/addon/mutator.go
```diff
@@ -0,0 +1,113 @@
...
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/sirupsen/logrus"
+
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
guangbochen, line 14: fix the import format.

pkg/apis/harvesterhci.io/v1beta1/addon.go
w13915984028: designed but not used, will remove it.

pkg/webhook/resources/addon/mutator_test.go
```diff
@@ -0,0 +1,158 @@
+package addon
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
guangbochen, line 8: ditto

pkg/webhook/resources/addon/validator.go
```diff
@@ -83,12 +83,35 @@ func validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) erro
 		return werror.NewBadRequest("chart field cannot be changed.")
 	}
 
+	// do not allow change Spec when an operation is onging, but allow change annotation, e.g. adjust the timeout
+	if oldAddon.Status.Status == v1beta1.AddonEnabled && isAddonSpecChanged(newAddon, oldAddon) {
+		return werror.NewBadRequest("addon is being deployed now, please wait and retry")
```
guangbochen, line 88: rephrase: the addon is being deployed, please try again later.

Additionally, this will block the upgrade case:
1. enable the addon chart in the previous version
2. apply the new crd changes and upgrade the harvester api and webhook image
3. click to disable the existing addon and it was blocked by the webhook here

```
status": {
"conditions": [
{
"error": false,
"lastUpdateTime": "",
"status": "",
"transitioning": false,
"type": ""
},
{
"error": false,
"lastUpdateTime": "2023-07-05T09:54:06Z",
"status": "",
"transitioning": false,
"type": "Enable"
}
],
"status": "AddonEnabled"
}
```

pkg/webhook/resources/addon/validator.go
```diff
@@ -83,12 +83,35 @@ func validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) erro
...
+	}
+
+	if oldAddon.Status.Status == v1beta1.AddonUpdating && isAddonSpecChanged(newAddon, oldAddon) {
+		return werror.NewBadRequest("addon is being updated now, please wait and retry")
+	}
+
+	if oldAddon.Status.Status == v1beta1.AddonDisabling && isAddonSpecChanged(newAddon, oldAddon) {
+		return werror.NewBadRequest("addon is being disabled now, please wait and retry")
```
guangbochen, line 96: ditto

pkg/webhook/resources/addon/validator.go
```diff
@@ -83,12 +83,35 @@ func validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) erro
...
+	}
+
 	if newAddon.Name == vClusterAddonName && newAddon.Namespace == vClusterAddonNamespace && newAddon.Spec.Enabled {
 		return validateVClusterAddon(newAddon)
 	}
 	return nil
 }
 
+func isAddonSpecChanged(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) bool {
+	if newAddon.Spec != oldAddon.Spec {
```
guangbochen, line 106: for spec compare use the deepEqual method instead.

pkg/webhook/resources/addon/mutator.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+
+		// timestamp is there
+		if _, ok := addon.Annotations[util.AnnotationAddonLastOperationTimestamp]; ok {
+			jsonOp2 = "replace"
+		}
+	}
+
+	// patch last operation, the key should be like harvesterhci.io~1addon-last-operation instead of harvesterhci.io/addon-last-operation
+	key := strings.Replace(util.AnnotationAddonLastOperation, "/", "~1", 1)
```
guangbochen, line 103: this is not a standard key name, any reason for the design choice here? and why we need to use a different key?

pkg/controller/master/addon/patch_grafana.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	harvesterutil "github.com/harvester/harvester/pkg/util"
+)
+
+// MonitorAddonRancherMonitoring will track status change of rancher-monitroing addon
+func (h *Handler) MonitorAddonRancherMonitoring(key string, aObj *harvesterv1.Addon) (*harvesterv1.Addon, error) {
```
guangbochen, line 15: \s\aObj\addon, using CR name is preferred for consistency.

pkg/controller/master/addon/patch_grafana.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// pv name format: volumeName: pvc-a95cdf6b-26bc-4ed6-8aaf-5f5d6b0b6dbd
+	pv, err := h.pv.Cache().Get(pvc.Spec.VolumeName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logrus.Debugf("grafana PV %s is not found, retry", pvc.Spec.VolumeName)
+			return false, "", fmt.Errorf("grafana PV %s is not found, retry", pvc.Spec.VolumeName)
+		}
+
+		return false, "", fmt.Errorf("error fetching grafana pv %s: %v", pvc.Spec.VolumeName, err)
```
guangbochen, line 70: we could return the error directly since it already contains the proper error message.

pkg/controller/master/addon/patch_grafana.go
```diff
@@ -0,0 +1,175 @@
...
+
+	logrus.Debugf("grafana PV %s is patched with new ReclaimPolicy %s", pvc.Spec.VolumeName, corev1.PersistentVolumeReclaimRetain)
+	return true, pvc.Spec.VolumeName, nil
+}
+
+// MonitorAddon will track the deployment of Addon
+func (h *Handler) annotateGrafanaPVName(pvname string, aObj *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+	if aObj == nil || aObj.DeletionTimestamp != nil {
+		return nil, nil
+	}
```
guangbochen, line 94: seems to be a redundant check.

pkg/controller/master/addon/patch_grafana.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if aObj.Annotations != nil {
+		// if p is not find, the return p is ""
+		if p, _ := aObj.Annotations[harvesterutil.AnnotationGrafanaPVName]; p == pvname {
```
guangbochen, line 98: don't skip the ok check here

pkg/controller/master/addon/patch_grafana.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	a.Annotations[harvesterutil.AnnotationGrafanaPVName] = pvname
+	return h.addon.Update(a)
+}
+
+func (h *Handler) patchGrafanaPVClaimRef(aObj *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+	if aObj == nil || aObj.DeletionTimestamp != nil {
+		return nil, nil
+	}
```
guangbochen, line 115: ditto

pkg/controller/master/addon/common.go
```diff
@@ -0,0 +1,402 @@
...
+		return false, true
+	}
+
+	timeout := time.Duration(addonTimeout)
+	if aObj.Annotations != nil {
+		if timeoutAnno, ok := aObj.Annotations[util.AnnotationAddonOperationTimeout]; ok {
+			i, err := strconv.ParseInt(timeoutAnno, 10, 32)
+			if err != nil {
+				logrus.Debugf("addon %s has invalid AnnotationAddonOperationTimeout, fallback to default value %v", aObj.Name, timeout)
+			} else if i >= addonTimeoutMin && i <= addonTimeoutMax {
```
chrisho, line 370: Also, add timeout range validation and validation of valid values to the webhook to let users know.

pkg/controller/master/addon/helmchart.go
```diff
@@ -0,0 +1,111 @@
...
+		if tm != nil && job.CreationTimestamp.Before(tm) {
+			logrus.Debugf("previous job %s/%s is to be deleted, wait", job.Namespace, job.Name)
+			if err := h.job.Delete(job.Namespace, job.Name, &metav1.DeleteOptions{}); err != nil {
+				return aObj, false, hc, owned, fmt.Errorf("error deleting helmchart related job %s/%s %v", job.Namespace, job.Name, err)
+			}
+			wait = true
+		}
+	}
+
+	return aObj, wait, hc, owned, nil
```
chrisho, line 110: Remove if the first return value unused.

pkg/controller/master/addon/addon.go
```diff
@@ -11,19 +10,19 @@ import (
 	ctlhelmv1 "github.com/k3s-io/helm-controller/pkg/generated/controllers/helm.cattle.io/v1"
 	"github.com/rancher/norman/types/slice"
 	catalogv1 "github.com/rancher/rancher/pkg/apis/catalog.cattle.io/v1"
+
```
guangbochen, line 13: remove the empty line here

pkg/controller/master/addon/helmchart.go
```diff
@@ -0,0 +1,111 @@
...
+	"strings"
+
+	helmv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
+	"github.com/rancher/wrangler/pkg/condition"
+
+	"github.com/rancher/wrangler/pkg/relatedresource"
+	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
```
guangbochen, line 14: ditto

pkg/controller/master/addon/helmchart.go
```diff
@@ -0,0 +1,111 @@
...
+const (
+	helmChartJobInstallPrefix = "helm-install"
+	helmChartJobDeletePrefix  = "helm-delete"
+)
+
+// downstream helmchart change will also trigger Addon OnChange
+func (h *Handler) ReconcileHelmChartOwners(_ string, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
+	if hc, ok := obj.(*helmv1.HelmChart); ok {
+		for _, v := range hc.GetOwnerReferences() {
+			if strings.ToLower(v.Kind) == "addon" {
```
guangbochen, line 30: define `addon` as constant name is preferred

pkg/controller/master/addon/helmchart.go
```diff
@@ -0,0 +1,111 @@
...
+
+// this file includes some workarounds on HelmChart
+
+const (
+	helmChartJobInstallPrefix = "helm-install"
+	helmChartJobDeletePrefix  = "helm-delete"
+)
+
+// downstream helmchart change will also trigger Addon OnChange
+func (h *Handler) ReconcileHelmChartOwners(_ string, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
```
guangbochen, line 27: seems error response it not needed.

pkg/controller/master/addon/common.go
```diff
@@ -0,0 +1,402 @@
...
+	"time"
+
+	helmv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
+	"github.com/rancher/wrangler/pkg/condition"
+
+	"github.com/sirupsen/logrus"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
guangbochen, line 16: ditto

pkg/webhook/resources/addon/validator.go
w13915984028: as the comment said, we may add more check here.

pkg/controller/master/addon/common.go
w13915984028: let validator do a check is OK, it could be added later.

the controller can tolerent the invalid value, as we have a default value to fallback.

pkg/controller/master/addon/helmchart.go
w13915984028: it does not matter too much, it may be used anytime

pkg/webhook/resources/addon/validator.go
w13915984028: use reflect.DeepEqual is a good idea; right now, the addon.Spec are all strings,  A == B check can work. but better to change

pkg/controller/master/addon/common.go
w13915984028: em, the timeout should be added to document rather than validator the annotations. in k8s, annotations are flexible to add/modify, it only works when annotations meets declare, but rarely using validator to validate it.  the addon may be added from bootstrap stage, or upgraded from previous;  validator to force check, may break something.

pkg/controller/master/addon/common.go
```diff
@@ -0,0 +1,402 @@
...
+	addonTimeoutMax = 60
+)
+
+type addonOperationRecord struct {
+	Operation                harvesterv1.AddonOperation `json:"operation,omitempty"`
+	OperationTimestamp       metav1.Time                `json:"operationTimestamp,omitempty"`
+	OperationTimestampString string                     `json:"operationTimestampstr,omitempty"`
+}
+
+func newAddonOperationRecord(operation string, tm string) (*addonOperationRecord, error) {
```
guangbochen, line 37: prefer to use abbreviation when the var name is too long or meaningful enough, e.g., use `time` instead of `tm` is preferred in here.

pkg/controller/master/addon/addon.go
```diff
@@ -39,290 +38,528 @@ var (
...
+	addonController.OnChange(ctx, "monitor-addon-rancher-monitoring", h.MonitorAddonRancherMonitoring)
 	appController.OnChange(ctx, "monitor-apps-catalog", h.PatchApps)
 	relatedresource.Watch(ctx, "watch-helmcharts", h.ReconcileHelmChartOwners, addonController, helmController)
 	return nil
 }
 
-// MonitorChanges will trigger updates to helm chart based on changes in values
-func (h *Handler) MonitorChanges(key string, aObj *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+// MonitorAddonPerStatus will track the OnChange on addon, route actions per addon status
+func (h *Handler) MonitorAddonPerStatus(key string, aObj *harvesterv1.Addon) (*harvesterv1.Addon, error) {
```
guangbochen, line 71: \s\aObj\addon is preferred for better project consistency.

pkg/webhook/resources/addon/mutator.go
```diff
@@ -0,0 +1,113 @@
...
+func NewMutator(addons ctlharvesterv1.AddonCache) types.Mutator {
+	return &addonMutator{
+		addons: addons,
+	}
+}
+
+// addonMutator injects last operation and timestamp
+type addonMutator struct {
+	types.DefaultMutator
+
```
guangbochen, line 31: remove the empty line.

pkg/apis/harvesterhci.io/v1beta1/addon.go
```diff
@@ -10,6 +11,33 @@ const (
 	AddonEnabled  AddonState = "AddonEnabled"
 	AddonDeployed AddonState = "AddonDeploySuccessful"
 	AddonFailed   AddonState = "AddonDeployFailed"
```
guangbochen, line 13: consider to move any deprecated state under a proper comment.

pkg/webhook/resources/addon/mutator.go
```diff
@@ -0,0 +1,113 @@
...
+	return patchLastOperation(newAddon, patchOps, string(addonOperation))
+}
+
+func patchLastOperation(addon *harvesterv1.Addon, patchOps types.PatchOps, addonOperation string) (types.PatchOps, error) {
+	jsonOp1 := "add"
+	jsonOp2 := "add"
+	if addon.Annotations == nil {
+		addon.Annotations = make(map[string]string, 2)
+	} else {
```
guangbochen, line 91: remove else here

pkg/controller/master/addon/addon.go
```diff
@@ -39,290 +38,528 @@ var (
...
-// MonitorAddon will track the deployment of Addon
-func (h *Handler) MonitorAddon(key string, aObj *harvesterv1.Addon) (*harvesterv1.Addon, error) {
-	if aObj == nil || aObj.DeletionTimestamp != nil {
-		return nil, nil
+// processAddonEnabling finish the ongoing enabling operation
+func (h *Handler) processAddonEnabling(aObj *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+	// status self recovering, generally it should not happen
+	if !aObj.Spec.Enabled {
+		msg := fmt.Sprintf("aborted, addon %s is enabling, but .Spec.Enabled is changed to false, check", aObj.Name)
+		logrus.Warnf(msg)
```
guangbochen, line 141: merge into one

pkg/controller/master/addon/common.go
```diff
@@ -0,0 +1,402 @@
...
+
+	aor.OperationTimestamp.Time = t
+	aor.OperationTimestampString = tm
+
+	return &aor, nil
+}
+
+func (h *Handler) updateAddonConditions(aObj *harvesterv1.Addon, cond condition.Cond, status corev1.ConditionStatus, reason, message string, updateLastTransitionTime bool) {
+	// wrangler's cond function is not very fit for here, any read/write needs to loop; play condition directly
+	for i := range aObj.Status.Conditions {
```
guangbochen, line 63: why can't we just update the condition via cond.SetMethods()? and prefer to retun the cond after update.

pkg/controller/master/addon/patch_grafana.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if aObj.Name != harvesterutil.RancherMonitoringName || aObj.Namespace != harvesterutil.CattleMonitoringSystemNamespaceName {
+		return aObj, nil
+	}
+
+	// deployed successfully
+	if aObj.Spec.Enabled && aObj.Status.Status == harvesterv1.AddonDeployed {
+		// patch pv ReclaimPolicy
+		_, pvname, err := h.patchGrafanaPVReclaimPolicy(harvesterutil.CattleMonitoringSystemNamespaceName, harvesterutil.GrafanaPVCName)
+
```
guangbochen, line 28: no need to add the line break here.

pkg/controller/master/addon/addon.go
```diff
@@ -39,290 +38,528 @@ var (
...
 	if aObj == nil || aObj.DeletionTimestamp != nil {
 		return nil, nil
 	}
 
-	a := aObj.DeepCopy()
-	if a.Status.Status != harvesterv1.AddonDeployed && a.Status.Status != harvesterv1.AddonFailed {
-		return a, nil
+	switch aObj.Status.Status {
+	// finish the onging enable
+	case harvesterv1.AddonEnabled:
```
ibrokethecloud, line 78: based on the workflow of `processAddonEnabling` we should be checking the status `harvesterv1.AddonEnabling`

pkg/controller/master/addon/common.go
```diff
@@ -0,0 +1,402 @@
...
+		if aObj.Status.Conditions[i].Type == cond {
+			aObj.Status.Conditions[i].LastUpdateTime = metav1.Now().UTC().Format(time.RFC3339)
+			return
+		}
+	}
+
+	// not find
+	if len(aObj.Status.Conditions) == 0 {
+		aObj.Status.Conditions = make([]harvesterv1.Condition, 1)
```
chrisho, line 186: Could we use the init condition instead of the empty condition? The empty condition seems like a bug.

pkg/webhook/resources/addon/mutator.go
w13915984028: when patching via json path, the /a/b/c/d has difficulty to express something like /a/b/(c/d), thats why use `~1` to escape `/`

to express a map key; otherwise, you can't patch the annotation with key in format of `harvesterhci.io/xyz`

pkg/controller/master/addon/patch_grafana.go
w13915984028: those are old codes;

the new refactor just follow the naming;  if replace, we need to change all those files

it looks not necessary

pkg/controller/master/addon/patch_grafana.go
w13915984028: for string type, there are many such kind of writing, to skip ok, as go will init string variable with `""` when it does not hit in map,  it is in our code base, if you search

pkg/controller/master/addon/helmchart.go
w13915984028: existing code ...

pkg/controller/master/addon/common.go
w13915984028: there is a package called `time`, use `time` as var name, sometimes is not OK

pkg/controller/master/addon/common.go
w13915984028: it can set one field at a time, and it will always loop to find this field; when you update 5 fields in a function, the wrangler function will blindly loop 5 times ...

pkg/controller/master/addon/patch_grafana.go
w13915984028: if go-lint does not complain,  we may skip those tiny aspects

pkg/webhook/resources/addon/mutator.go
w13915984028: if you check other mutator, the general way is leave one blank line after types

btw, if `go-lint` does not complain, could we skip those tiny aspects

pkg/controller/master/addon/common.go
w13915984028: could you explain a bit more? what do you mean `init` condition and `empty` condition

w13915984028:
for those naming, `else`, line breaks, could we simply follow `golang-lint`, there are already so many kinds of formatting check, if it does not complain, our review may skip those; we are hard to be better than those rules based check.

focus on functional bugs, logical errors, robustness etc.

pkg/webhook/resources/addon/mutator.go
w13915984028: strictly, the `else` is needed. the map looking is only needed when annotations is not nil; even though without else it can also wok.

pkg/controller/master/addon/addon.go
w13915984028: `msg` is also used in L142, thats why it is abstracted in L140

pkg/controller/master/addon/common.go
w13915984028: and, IIRC, wrangler does not have function to update the `LastTransitTime` (or `LastUpdateTime`), it only has one public function to update one of those 2 timestamps.

guangbochen:
> for those naming, `else`, line breaks, could we simply follow `golang-lint`, there are already so many kinds of formatting check, if it does not complain, our review may skip those; we are hard to be better than those rules based check.
> 
> focus on functional bugs, logical errors, robustness etc.

We should definitely focus on the functional bugs and logical ones first, but a good coding style and format(those are not covered by the golang-lint) should also be encouraged across the project for better maintenance and readability; unless it is a big effort otherwise I would suggest us to follow it.

pkg/webhook/resources/addon/validator.go
w13915984028: did not figure out reason yet, @ibrokethecloud any idea of this scenario?

the `AddonEnabled` should only be a middle status, it will finally be successful or failed (timeout)

@guangbochen , the action `3` seems to be expected, as the `addon` is being deployed now, disable is not allowed.

guangbochen: I would assume the current code didn't cover the migration/upgrade logic from the previous addon status; that's why step3 is getting blocked. (also talked with @ibrokethecloud, and he will help to continue on this PR, thanks.)

pkg/controller/master/addon/common.go
chrisho: `Addon.status.conditions` will appear as including an empty condition when enable for the first time, it should add a new type instead the empty conditions, e.g.: Initialized, as follow:
```yaml
status:
  conditions:
    # Empty condition
    - status: ''
      type: ''
    # Initialized condition
    - lastUpdateTime: '2023-07-06T06:25:23Z'
      status: 'True'
      type: 'Initialized'
    - lastUpdateTime: '2023-07-06T06:25:23Z'
      status: ''
      type: Enable
  status: AddonEnabled
```
The empty condition still exists whatever performs any operation.

chrisho:
Just a suggestion, notice the conditions, as long as they have been enabled once, even after disabling, the enable condition is True, do we want to update the conditions and change the enable condition to False/Unknown or even delete? 
```yaml
status:
  conditions:
    - status: ''
      type: ''
    - lastTransitionTime: '2023-07-06T06:30:36Z'
      lastUpdateTime: '2023-07-06T06:30:51Z'
      status: 'True'
      type: Enable
    - lastTransitionTime: '2023-07-06T06:34:04Z'
      lastUpdateTime: '2023-07-06T06:34:04Z'
      status: Unknown
      type: Failed
    - lastTransitionTime: '2023-07-06T06:34:04Z'
      lastUpdateTime: '2023-07-06T06:34:19Z'
      status: 'True'
      type: Disable
```

PR #4141, opened 2023-06-28
added validation for hostname field for rancher-vcluster addon

ibrokethecloud:

**Problem:**

1.2.0 introduces the experimental rancher-vcluster addon.

The PR introduces additional check in the validating webhook to ensure that a valid dns hostname is provided and IP based hostnames are rejected.

**Solution:**

**Related Issue:**
**Test plan:**

### DIFF
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -108,7 +108,8 @@ func validateVClusterAddon(newAddon *v1beta1.Addon) error {
 		if ipErrs := validationutil.IsValidIP(addonContent.Hostname); len(ipErrs) == 0 {
 			return werror.NewBadRequest(fmt.Sprintf("%s is not a valid hostname", addonContent.Hostname))
 		}
+		return nil
 	}
 
-	return nil
+	return werror.NewBadRequest(fmt.Sprintf("invalid hostname %s provided for %s addon", addonContent.Hostname, vClusterAddonName))
 }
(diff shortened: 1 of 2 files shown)


bk201 (approved):
LGTM

bk201 (approved):

pkg/webhook/resources/addon/validator.go
```diff
@@ -75,5 +83,33 @@ func validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) erro
...
+	if newAddon.Name == vClusterAddonName && newAddon.Namespace == vClusterAddonNamespace && newAddon.Spec.Enabled {
+		return validateVClusterAddon(newAddon)
+	}
 	return nil
 }
+
+func validateVClusterAddon(newAddon *v1beta1.Addon) error {
+	type contentValues struct {
+		Hostname string `yaml:"hostname"`
+	}
```
guangbochen, line 95: define this on the top is preferred.

guangbochen (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.2

w13915984028:
A general question: @ibrokethecloud @guangbochen 

If an addon, or any CRD object depends on the validator, is created in system bootstrap stage;  or in certain cases, the Harvester&Webhook are scaled down;  an `invalid` object is added into api-server.

Should we also do same checks in controller side ?, e.g., put addon into DeployFailed status. thanks.

guangbochen:
It is possible but scaling down the webhook will require manual operations. So prefer to leave it as a corner case and ask the users address it manually when it occurs.

PR #4143, opened 2023-06-28
Improve warning message if free system disk space is insufficient for upgrading

WebberHuang1118:

**Problem:**

Improve warning message if free system disk space is insufficient for upgrading

**Solution:**

In the error message, adding information about free system disk space

**Related Issue:**
https://github.com/harvester/harvester/issues/2935#issuecomment-1580231584

**Test plan:**

2. Create a new Harvester cluster. One of the nodes' disk space should be 250G.
3. If every node's `/usr/local` free space is more than 30G, you can use `dd` to write files to make sure one of the nodes is less than 30G. For example: `dd if=/dev/zero of=/usr/local/test.img bs=1G count=5`.
4. Create a test version.
    ```yaml
    kind: Version
    metadata:
      name: master
      namespace: harvester-system
    spec:
      isoURL: http://192.168.0.55:8080/iso/harvester-master-amd64.iso
    ```
4. Click upgrade on the dashboard. We should get an error message mentioning `/usr/local`.
upgrade_insufficient

### DIFF
--- a/pkg/webhook/resources/upgrade/validator.go
+++ b/pkg/webhook/resources/upgrade/validator.go
@@ -38,6 +38,7 @@ const (
 	rkeInternalIPAnnotation        = "rke2.io/internal-ip"
 	managedChartNamespace          = "fleet-local"
 	defaultMinFreeDiskSpace uint64 = 30 * 1024 * 1024 * 1024 // 30GB
+	freeSystemPartitionMsg         = "df -h '/usr/local/'"
 )
 
 func NewValidator(
@@ -282,7 +283,8 @@ func (v *upgradeValidator) checkDiskSpace(node *corev1.Node, minFreeDiskSpace ui
 	if *summary.Node.Fs.AvailableBytes < minFreeDiskSpace {
 		min := units.BytesSize(float64(minFreeDiskSpace))
 		avail := units.BytesSize(float64(*summary.Node.Fs.AvailableBytes))
-		return werror.NewBadRequest(fmt.Sprintf("Node %q has insufficient free space (%s). An upgrade requires at least %s of free space.", node.Name, avail, min))
+		return werror.NewBadRequest(fmt.Sprintf("Node %q has insufficient free system partition space %s (%s). The upgrade requires at least %s of free system partition space on each node.",
+			node.Name, avail, freeSystemPartitionMsg, min))
 	}
 	return nil
 }


w13915984028:
Please double check with doc PR https://github.com/harvester/docs/pull/332/files#diff-9b99ca129b131cc50e1c584be79715640745a08c667b20f0fa714558b2d3077aL37, we tend to to call it `free system disk space`, lets unify the code and doc, thanks.

WebberHuang1118:
> Please double check with doc PR https://github.com/harvester/docs/pull/332/files#diff-9b99ca129b131cc50e1c584be79715640745a08c667b20f0fa714558b2d3077aL37, we tend to to call it `free system disk space`, lets unify the code and doc, thanks.

Update error message as `"Node <node id> has insufficient free system disk space <current free system disk space>  ('df -h /ur/local/'). An upgrade requires at least <minimal free system disk space> of free system disk space."`

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -282,7 +283,8 @@ func (v *upgradeValidator) checkDiskSpace(node *corev1.Node, minFreeDiskSpace ui
 	if *summary.Node.Fs.AvailableBytes < minFreeDiskSpace {
 		min := units.BytesSize(float64(minFreeDiskSpace))
 		avail := units.BytesSize(float64(*summary.Node.Fs.AvailableBytes))
-		return werror.NewBadRequest(fmt.Sprintf("Node %q has insufficient free space (%s). An upgrade requires at least %s of free space.", node.Name, avail, min))
+		return werror.NewBadRequest(fmt.Sprintf("Node %q has insufficient free system disk space %s (%s). An upgrade requires at least %s of free system disk space.",
```
w13915984028: `free system disk space` or `free system partition space` ?  seems `partition` will be better. We can change code and doc to use `system partition`.   @bk201  WDYT ? thanks.

@WebberHuang1118 
`An upgrade requires at least %s of free system disk space.` -> `The upgrade requires at least %s of free system disk space on each node.`  seems better.

bk201: Both work. But partition is more accurate. The goal is to display the `/usr/local` info.

w13915984028: OK, lets use `free system partition space`, I will also update doc PR.

bk201 (approved):
LGTM

w13915984028 (approved):
LGTM, thanks.

pkg/webhook/resources/upgrade/validator.go
WebberHuang1118: Updated error message according to our discussion

bk201:
@mergifyio backport v1.2

WebberHuang1118:
@mergify backport v1.2

WebberHuang1118:
@Mergifyio backport v1.1

Issue #4095, opened 2023-06-16
[BUG] Upgrade stuck on post-drain job with the unresponsive VM


Vicente-Cheng:
**Describe the bug**
Upgrade stuck on the post-drin job, and you could observe these two situations.

1. on the post-drin job, we could see the similar log as follows
```
hvst-upgrade-79jcs-post-drain-harvester-node-0-2m8zh + curl -sSfL http://upgrade-repo-hvst-upgrade-79jcs.harvester-system/harvester-iso/rootfs.squashfs -o /host/usr/local/upgrade_tmp/tmp.ZqMu
5Yk811
hvst-upgrade-79jcs-post-drain-harvester-node-0-2m8zh curl: (7) Failed to connect to upgrade-repo-hvst-upgrade-79jcs.harvester-system port 80 after 3080 ms: Couldn't connect to server
hvst-upgrade-79jcs-post-drain-harvester-node-0-2m8zh + echo 'Failed to download the requested file from "http://upgrade-repo-hvst-upgrade-79jcs.harvester-system/harvester-iso/rootfs.squashfs"
 to "/host/usr/local/upgrade_tmp/tmp.ZqMu5Yk811" with error code: 7, retrying (6)...'
```
2. you could find the upgrade VM will have high CPU/Memory usage, which makes it not respond.
![截圖 2023-06-16 上午12 56 05](https://github.com/harvester/harvester/assets/1615081/09c35e95-48b8-4423-ae66-179fb1460ec2)

**To Reproduce**
1. install harvester v1.1.2
2. upgrade to v1.2.0-rc2
3. You will have a chance to see this issue.

**Expected behavior**
The post-drin job should be smooth.

**Support bundle**
None

**Environment**
 - Harvester ISO version: 1.1.2
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 VMs cluster

**Additional context**
We could force delete this VM. After this VM re-launch, the upgrade will continue.

bk201:
Seeing this in upgrade-vm console:
<img width="1022" alt="Screenshot 2023-06-16 at 11 25 55 AM" src="https://github.com/harvester/harvester/assets/1691518/586461ca-067f-47fd-b389-e72453ae8014">

bk201:
Analysis of the issue:
- `virsh reset the VM` doesn't work, the VM can't read the disk. The virt-launcher pod can't access the disk either.
[...]

PR #4149, opened 2023-06-29
fix(upgrade): skip restarting RKE2 servers

starbops:
**Problem:**

There will be machine plan (`custom-xxxxx` Secrets) changes after Rancher upgrades from v2.6.x to v2.7.x during Harvester upgrades. The RKE2 server/agent on the nodes will be restarted by `rancher-system-agent` while Rancher detects the changes and tries to reconcile them. This behavior brings down LH volumes and then causes various types of issues to the Pods and VMs associated with the volumes.

**Solution:**

Adding an environment variable `INSTALL_RKE2_SKIP_ENABLE=true` to make `rancher-system-agent`  to NOT restart `rke2-server`/`rke2-agent` triggered by machine plan reconciliation during Rancher upgrade (from v2.6.x to v2.7.x). This is kind of a workaround. We add a systemd service drop-in file for `rancher-system-agent` to feed an additional environment file, then reload and restart the service. The environment variable will take effect until we remove it at the node pre-drain phase.

**Related Issue:**

#4095 

**Test plan:**

1. Prepare a multi-node Harvester cluster in v1.1.2
2. Upgrade with the ISO containing the fix
3. Monitor the `rke2-server`/`rke2-agent` on the nodes during the apply-manifest phase with the following commands, to see if they get restarted (they should **NOT** be restarted)
```
# on the nodes, check the "Active" timestamp
sudo systemctl status rke2-server.service # if it's a worker node, use "rke2-agent.service" instead
```
4. The upgrade should be successful

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -1121,6 +1121,75 @@ pause_all_charts() {
   done
 }
 
+skip_restart_rancher_system_agent() {
+  # to prevent rke2-server/agent from restarting during the rancher upgrade.
+  # by adding an env var to temporarily make rancher-system-agent on each node skip restarting rke2-server/agent.
+  # issue link: https://github.com/rancher/rancher/issues/41965
+
+  # only versions before v1.2.0 that upgrading to v1.2.0 need this workaround
+  if [[ ! "${UPGRADE_PREVIOUS_VERSION%%-rc*}" < "v1.2.0" ]]; then
+    echo "Only versions before v1.2.0 need this patch."
+    return
+  fi
+
+  plan_manifest="$(mktemp --suffix=.yaml)"
+  plan_name="$HARVESTER_UPGRADE_NAME"-skip-restart-rancher-system-agent
+  plan_version="$(openssl rand -hex 4)"
+
+  cat > "$plan_manifest" <<EOF
+apiVersion: upgrade.cattle.io/v1
+kind: Plan
+metadata:
+  name: $plan_name
+  namespace: cattle-system
+spec:
+  concurrency: 10
+  nodeSelector:
+    matchLabels:
+      harvesterhci.io/managed: "true"
+  serviceAccountName: system-upgrade-controller
+  tolerations:
+  - operator: "Exists"
+  upgrade:
+    image: registry.suse.com/bci/bci-base:15.4
+    command:
+    - chroot
+    - /host
+    args:
+    - sh
+    - -c
+    - set -x && mkdir -p /run/systemd/system/rancher-system-agent.service.d && echo -e '[Service]\nEnvironmentFile=-/run/systemd/system/rancher-system-agent.service.d/10-harvester-upgrade.env' | tee /run/systemd/system/rancher-system-agent.service.d/override.conf && echo 'INSTALL_RKE2_SKIP_ENABLE=true' | tee /run/systemd/system/rancher-system-agent.service.d/10-harvester-upgrade.env && systemctl daemon-reload && systemctl restart rancher-system-agent.service
+  version: $plan_version
+EOF
+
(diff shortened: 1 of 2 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -370,6 +374,8 @@ EOF
 }
 
 command_pre_drain() {
+  recover_rancher_system_agent
```
bk201, line 384: pre_drain only works for 2+ node. We need to handle the single-node case as well.

bk201 (approved):
LGTM, thanks!

chrisho (approved):
LGTM and rke2-server have not restarted during in apply-manifest phase, thanks.

package/upgrade/upgrade_manifests.sh
```diff
@@ -1106,6 +1106,84 @@ pause_all_charts() {
...
+    operator: Equal
+    value: amd64
+  - effect: NoSchedule
+    key: kubernetes.io/arch
+    operator: Equal
+    value: arm64
+  - effect: NoSchedule
+    key: kubernetes.io/arch
+    operator: Equal
+    value: arm
```
guangbochen: Harvester does not support ARM arch ATM, so we can probably skip it.

package/upgrade/upgrade_manifests.sh
```diff
@@ -1106,6 +1106,84 @@ pause_all_charts() {
...
+  - effect: NoSchedule
+    key: kubevirt.io/drain
+    operator: Exists
+  - effect: NoExecute
+    key: node-role.kubernetes.io/control-plane
+    operator: Exists
+  - effect: NoSchedule
+    key: kubernetes.io/arch
+    operator: Equal
+    value: amd64
```
guangbochen: we may consider using the below config to tolerate all taints.
```
tolerations:
- operator: "Exists"
```

package/upgrade/upgrade_manifests.sh
```diff
@@ -1161,6 +1239,7 @@ detect_upgrade
 check_version
 pre_upgrade_manifest
 pause_all_charts
+skip_restart_rancher_system_agent
```
guangbochen, line 1248: let's add proper comments and issue link for reference, also prefer to check only run it if the previous Harvester version is less than `v1.2.0`.

package/upgrade/upgrade_manifests.sh
```diff
@@ -1106,6 +1106,84 @@ pause_all_charts() {
   done
 }
 
+skip_restart_rancher_system_agent() {
+  plan_manifest="$(mktemp --suffix=.yaml)"
+  plan_name="$HARVESTER_UPGRADE_NAME"-skip-restart-rancher-system-agent
+  plan_version="$(openssl rand -hex 4)"
```
guangbochen, line 1137: Just a suggestion, would it be preferred if we used a predictable plan version (e.g. upgrade name + target version)? So even if the upgrade pod is rebuilt for unexpected reasons, it will try to reuse the existing plan, and leaving no legacy plans behind.

starbops: IIRC, if reusing the plan version and the plan's content (the script itself) is not changed, no re-execution will happen because a hashed value for the plan was already added to the nodes' label if it has been executed before. This is just a one-time plan. We'll remove it after execution, so a random string for the plan version would be good enough.

guangbochen: In that case, we don't need to change the existing one, thanks.

package/upgrade/upgrade_manifests.sh
starbops: I have added the version check, PTAL. Thank you!

guangbochen (approved):
LGTM, and validated that the plan will help to prevent all rke2.services from restarting on the Rancher upgrade.

chrisho (approved):

guangbochen:
@mergify backport v1.2

PR #4158, opened 2023-06-30
add opt-in Longhorn network policy

guangbochen:

**Problem:**
Need to add opt-in Longhorn network policy to secure its backend API.

**Solution:**
Enable the opt-in Longhorn network policy

**Related Issue:**
https://github.com/longhorn/longhorn/discussions/3031

**Test plan:**
-  Cases1: Users from the other namespaces should not be able to access the longhorn backend API
   - get the longhorn-backed svc IP via `kubectl get svc -n longhorn-system | grep longhorn-backend`
   - start a VM in a namespace e.g., default and exec into the VM pod via `$kubectl exec -it $pod bash`
   - run `curl http://$ip:9500/v1` and you will get a JSON response
   - after applying the LH network policy rules, users are no longer able to curl the Longhorn backend API in any other pods except those pods defined by the network policy.
- Case 2: admin user can still access the Longhorn dashboard UI
- Case 3&4: users should be able to create and restore VM backups, VM snapshot,s and volume snapshots
- Case 5: users can still use the Harvester csi-driver in the guest k8s cluster as before.
   - create an nginx pod and add persistent pvc using the `harvester` storage class in the guest rke2 cluster. 
   - check if both the pod and its mount point is running successfully.
image
image

### DIFF
--- a/deploy/charts/harvester/templates/longhorn-network-policy.yaml
+++ b/deploy/charts/harvester/templates/longhorn-network-policy.yaml
@@ -165,4 +165,24 @@ spec:
     - ports:
         - protocol: TCP
           port: 9502
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: cattle-monitoring-prometheus
+  namespace: longhorn-system
+spec:
+  podSelector:
+    matchLabels:
+      app: longhorn-manager
+  policyTypes:
+    - Ingress
+  ingress:
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: cattle-monitoring-system
+        podSelector:
+          matchLabels:
+            app.kubernetes.io/instance: rancher-monitoring-prometheus
 {{- end -}}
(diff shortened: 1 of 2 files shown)


chrisho (approved):
LGTM and all the cases work fine, thanks.

futuretea (approved):
LGTM, thanks. Test Plan Passed

bk201 (approved):
Checked some components that communicate with Longhorn:
- Harvester CSI driver: use kube-api 
- NDM: use kube-api
- Harvester: backup/other LH settings: use kube-api too

So LGTM.

An exception is [Prometheus metrics](https://longhorn.io/docs/1.4.2/monitoring/prometheus-and-grafana-setup/), the prometheus pod need to access longhorn-manager. Do we want to address this together?

Screenshot 2023-07-03 at 10 49 23 PM

guangbochen:
rebased and [added network policy ingress for the Prometheus monitoring pod](https://github.com/harvester/harvester/pull/4158/commits/b7980d9683d57b8bb4a0bb516a74bc30becd97b2),  PTAL again, thanks.
image

futuretea (approved):

chrisho (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.2
and @bk201 PTAL tmr and leave us a comment if u see any problems here, thanks.

Issue #4113, opened 2023-06-20
[ENHANCEMENT] Bump elemental-toolkits


Vicente-Cheng:
**Is your enhancement related to a problem? Please describe.**
Bump the new elemental-toolkits with Harvester OS.
We use the elemental-cli from luet. That is an old version. We want to use newer elemental-toolkits on OBS.

**Describe the solution you'd like**
Bump elemental-toolkits

**Describe alternatives you've considered**
None

**Additional context**
Another similar issue: https://github.com/harvester/harvester/issues/3587

irishgordo:
@Vicente-Cheng this seems to look good :smile: :+1: 

I'm able to note:
```
t:/home/rancher # cat /etc/os-release 
NAME="SLE Micro"
VERSION="5.3"
VERSION_ID="5.3"
PRETTY_NAME="Harvester v1.2"
ID="sle-micro-rancher"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sle-micro-rancher:5.3"
VARIANT="Harvester"
VARIANT_ID="Harvester-v1.2-20230629"
GRUB_ENTRY_NAME="Harvester v1.2"
t:/home/rancher # env
HOSTTYPE=x86_64
LESSCLOSE=lessclose.sh %s %s
XKEYSYMDB=/usr/X11R6/lib/X11/XKeysymDB
LANG=POSIX
WINDOWMANAGER=xterm
LESS=-M -I -R
SUDO_GID=1000
HOSTNAME=t
KUBECONFIG=/etc/rancher/rke2/rke2.yaml
CSHEDIT=emacs
GPG_TTY=/dev/pts/0
LESS_ADVANCED_PREPROCESSOR=no
COLORTERM=1
SUDO_COMMAND=/usr/bin/su
MACHTYPE=x86_64-suse-linux
MINICOM=-c on
CONTAINERD_ADDRESS=/run/k3s/containerd/containerd.sock
OSTYPE=linux
XDG_SESSION_ID=2
USER=root
PAGER=less
MORE=-sl
PWD=/home/rancher
HOME=/root
LC_CTYPE=en_US.UTF-8
HOST=t
SUDO_USER=rancher
XNLSPATH=/usr/X11R6/lib/X11/nls
XDG_SESSION_TYPE=tty
XDG_DATA_DIRS=/usr/share
PROFILEREAD=true
[...]

PR #4164, opened 2023-07-03
Upgrade: wrap latest elemental CLI

Vicente-Cheng:
- We still need the latest elemental CLI for upgrading.
      That is because we need to change bootargs from COS_ACTIVE to
      COS_STATE.

    - Get back the wrap mechanism
    - Revert "Revert "Upgrade: wrap the elemental CLI for fixing the upgrade failure""
      This reverts commit 20ca0b3e737ede6b64787284d49e38aa2f9601e9.

**Problem:**
We still need the latest elemental CLI for upgrading.
That is because we need to change bootargs from COS_ACTIVE to COS_STATE. 

**Solution:**
Get back the wrap mechanism

**Related Issue:**
https://github.com/harvester/harvester/issues/4113

**Test plan:**
Upgrade to latest elemental-toolkit based ISO should work.

### DIFF
--- a/package/upgrade/Dockerfile
+++ b/package/upgrade/Dockerfile
@@ -1,3 +1,6 @@
+# source reference: https://build.opensuse.org/package/show/isv:Rancher:Elemental:Stable:Teal53/builder-image
+FROM registry.opensuse.org/isv/rancher/elemental/stable/teal53/15.4/rancher/elemental-builder-image/5.3:0.3.1 AS temp_elemental
+
 FROM registry.suse.com/bci/bci-base:15.4
 
 ARG ARCH=amd64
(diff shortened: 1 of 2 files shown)


Vicente-Cheng:
Tested with the upgrade.

bk201 (approved):
LGTM, thanks!

starbops (approved):
LGTM, thank you.

package/upgrade/Dockerfile
```diff
@@ -1,3 +1,5 @@
+FROM registry.opensuse.org/isv/rancher/harvester/toolbox/main/elemental-toolbox:latest AS temp_elemental
```
guangbochen: any concerns about the lint error @Vicente-Cheng, are we possible to use a minor tag version e,g, v1.2?

Vicente-Cheng: We do not maintain the elemental-toolkit on OBS.
If we want to maintain our elemental-toolkit version with the release, we need to re-originate this package and make it under the OS (e.g., for v1.2, we will create another toolbox image on the Harvester:OS:v1.2:toolbox).

Do we need to handle this release? We might not need to wrap the elemental again once we upgrade to the latest elemental-toolkit. WDUT?

Vicente-Cheng: Hi @guangbochen, after checking how OBS tags.
I found the elemental team already has the related image, so I changed the image to this.
Please recheck it. Thanks!

package/upgrade/upgrade_node.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash -ex
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+ELEMENTAL_DIR="elemental_cli"
```
guangbochen: seems this is not used.

package/upgrade/Dockerfile
guangbochen: thanks for the update, can u please help to share the related OBS repo just for reference, thanks.

guangbochen (approved):
Just a tiny comment, verified the build works and the elemental cli version is `0.3.1+gbab71d1`, thanks.

package/upgrade/upgrade_node.sh
Vicente-Cheng: good catch, removed!

package/upgrade/Dockerfile
Vicente-Cheng: Sure, add the source reference.

guangbochen:
@mergify backport v1.2

Issue #4119, opened 2023-06-20
[BUG] Upgrade stucks in `Waiting for RKE2 to be upgraded`, fleet-agent was scaled down, but failed to scale up


w13915984028:
**Describe the bug**

![image](https://github.com/harvester/harvester/assets/31133476/79fff965-8b02-4cc6-88a5-1ae6183fc4cf)

`apply-manifests` finished , even with some error messages https://github.com/harvester/harvester/issues/4095#issuecomment-1599484467, but it did not stuck there 

```
logs of apply-manifests
....
addon.harvesterhci.io/vm-import-controller patched
addon.harvesterhci.io/pcidevices-controller patched
upgrade addon rancher_monitoring
upgrade addon rancher_logging

Tue Jun 20 20:50:22 UTC 2023
NAMESPACE     NAME                                          BUNDLEDEPLOYMENTS-READY   STATUS
fleet-local   fleet-agent-local                             1/1                       
fleet-local   local-managed-system-agent                    1/1                       
fleet-local   mcc-harvester                                 1/1                       
fleet-local   mcc-harvester-crd                             1/1                       
fleet-local   mcc-local-managed-system-upgrade-controller   1/1                       
fleet-local   mcc-rancher-logging-crd                       1/1                       
fleet-local   mcc-rancher-monitoring-crd                    1/1                       
all bundles are ready
Tue Jun 20 20:57:48 UTC 2023
```

```
harv31:/home/rancher # kk get upgrade.harvesterhci.io -A
NAMESPACE          NAME                 AGE
harvester-system   hvst-upgrade-scnqb   35m
harv31:/home/rancher # kk get upgrade.harvesterhci.io -A -oyaml
apiVersion: v1
items:
- apiVersion: harvesterhci.io/v1beta1
  kind: Upgrade
  metadata:
    creationTimestamp: "2023-06-20T20:29:00Z"
    finalizers:
    - wrangler.cattle.io/harvester-upgrade-controller
    generateName: hvst-upgrade-
    generation: 14
    labels:
      harvesterhci.io/latestUpgrade: "true"
      harvesterhci.io/upgradeState: UpgradingNodes
    name: hvst-upgrade-scnqb
    namespace: harvester-system
    resourceVersion: "56088"
    uid: a36e81b9-898f-4e5a-ade1-918994277f2a
  spec:
    image: ""
    logEnabled: true
    version: v1.2.0
  status:
    conditions:
    - status: Unknown
      type: Completed
[...]

PR #4170, opened 2023-07-03
Upgrade: fix RKE2 is not upgraded in the single node case

bk201:
**Problem:**

RKE2 is pending for upgrade in single node upgrade case.

**Solution:**

The cluster is waiting for the `mcc-local-managed-system-upgrade-controller` bundle.  The bundle can't be ready due to no fleet-agent pod running. Previously, we scaled down fleet-agent to avoid hitting http:///github.com/rancher/fleet/issues/637. Since https://github.com/rancher/fleet/issues/637 is fixed, we can remove the workaround.

**Related Issue:**
https://github.com/harvester/harvester/issues/4119

**Test plan:**

- Create a single node 1.1.2 cluster.
- Upgrade to the PR version; the upgrade should succeed.
**note**: need to test with https://github.com/harvester/harvester/pull/4149

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -605,10 +605,6 @@ command_single_node_upgrade() {
   echo "wait for fleet bundles before upgrading RKE2"
   # wait all fleet bundles in limited time
   wait_for_fleet_bundles
-  # scale down fleet-agent, scale up will be done in upgrade controller
-  echo "scale down fleet-agent"
-  kubectl scale --replicas=0 deployment/fleet-agent -n cattle-fleet-local-system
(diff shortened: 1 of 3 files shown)


starbops (commented):
Script part LGTM, but we might also want to remove the corresponding part in the upgrade controller that scales fleet-agent back up, too.

Vicente-Cheng (approved):
LGTM, tested with single node upgrade.

w13915984028 (approved):
LGTM, thanks.

guangbochen (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.2

Issue #4208, opened 2023-07-10
[ENHANCEMENT] Upgrade: speed up image preloading time and space utilization


bk201:
**Is your enhancement related to a problem? Please describe.**

There are some optimizations we can do in the prepare script.

**Describe the solution you'd like**

- Chain commands with pipes, this reduces the image loading time and eliminates temporary space requirement.
- Do not unpack the image in containerd.

**Describe alternatives you've considered**

**Additional context**

irishgordo:
@bk201  this looks good :smile: :+1:  - I'll go ahead and close this out.

**Test-Results:**

**v1.1.2->master-head: PASS**
Artifacts:
![Screenshot from 2023-07-13 15-16-02](https://github.com/harvester/harvester/assets/5370752/aa0aeed3-d913-40f2-88b6-9016792bc915)
![Screenshot from 2023-07-13 12-13-10](https://github.com/harvester/harvester/assets/5370752/57bb1d22-01e1-43d6-a993-81f8913514d9)

**v1.1.2->v1.1-head: PASS**
Artifacts:
![Screenshot from 2023-07-13 16-18-33](https://github.com/harvester/harvester/assets/5370752/e3da4a76-85bc-4c29-8c39-f55d5bace811)
![Screenshot from 2023-07-13 15-54-33](https://github.com/harvester/harvester/assets/5370752/937b6609-14de-4439-88a4-f884c561c1b6)

**v1.1.2->v1.2-head: PASS**
Artifacts:
![Screenshot from 2023-07-13 17-16-26](https://github.com/harvester/harvester/assets/5370752/1d9d672d-f435-4534-bd78-bf172c7e71e1)
![Screenshot from 2023-07-13 16-38-20](https://github.com/harvester/harvester/assets/5370752/6d2e8501-9077-4780-84ec-b07af507e18d)

PR #4209, opened 2023-07-10
Upgrade: minor optimization of preloading phase

bk201:
**Problem:**

There are some optimizations we can do in the prepare script.

**Solution:**

- Chain commands with pipes, this reduces the image loading time and eliminates temporary space requirement.
  - We don't save the tarball to disk anymore. This reduce the chance to cause kubelet image GC on disk pressure.
- Do not unpack the image in containerd.
  - Images are unpacked only when containers use them. This reduces the disk pressure as well. It's a tradeoff of space saving to a little bit container starting overhead.
 
**Results**
(capture with running Harvester VMs in WD SN750 NVMe disk):

### Single node upgrade
| Before |      After     |
|:------:|:--------------:|
| 04m53s | 2m38s          |

### 3-node upgrade

| Before |      After     |
|:------:|:--------------:|
| 7m | 3m52s          |

**Related Issue:**

https://github.com/harvester/harvester/issues/4208

**Test plan:**

- Create a Harvester cluster.
- Patch the rancher/harvester-upgrade:<current_version> with the PR change.
- Upgrade to a newer version. The new ISO needs to contain different images in image tarballs for testing. Otherwise, the preload script will skip the preloading. One tip is to start a v1.2.0-rc3 cluster and upgrade to a master build ISO.
- The upgrade should succeed, and the `upgrade-preapre-node-xxx` job should be faster than previous upgrades.

### DIFF
--- a/package/upgrade/lib.sh
+++ b/package/upgrade/lib.sh
@@ -198,7 +198,6 @@ import_image_archives_from_repo() {
       image_list_url="$UPGRADE_REPO_BUNDLE_ROOT/$list"
       archive_url="$UPGRADE_REPO_BUNDLE_ROOT/$archive"
       image_list_file="${tmp_image_archives}/$(basename $list)"
-      archive_file="${tmp_image_archives}/${archive_name}.tar"
 
       # Check if images already exist
       curl -sfL $image_list_url | sort > $image_list_file
@@ -208,9 +207,7 @@ import_image_archives_from_repo() {
         continue
       fi
 
-      curl -sfL $archive_url | zstd -d -f --no-progress -o $archive_file
-      $CTR -n k8s.io image import $archive_file
-      rm -f $archive_file
+      curl -fL $archive_url | zstdcat | $CTR -n k8s.io images import --no-unpack -
     done
   rm -rf $tmp_image_archives
 }


starbops (approved):
LGTM! Tested with patched v1.2.0-rc3 upgrading to master. Even tested with patched v1.1.2 upgrading to master, the optimization works really well that it makes me upgrade my upgrade-impossible machines possible (decreased the preloading time to under 15 mins).

futuretea (approved):
LGTM, thanks. Test Plan passed

guangbochen (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.2

futuretea:
@bk201 Do we need to backport it to v1.1?

bk201:
@futuretea Yes, thanks for the reminder.

bk201:
@mergifyio backport v1.1

package/upgrade/lib.sh
```diff
@@ -208,9 +207,7 @@ import_image_archives_from_repo() {
         continue
       fi
 
-      curl -sfL $archive_url | zstd -d -f --no-progress -o $archive_file
-      $CTR -n k8s.io image import $archive_file
-      rm -f $archive_file
+      curl -fL $archive_url | zstdcat | $CTR -n k8s.io images import --no-unpack -
```
w13915984028, line 210: A question:

Normaly, for `harvester` or others, there may only be a few images which are changed when upgrading. 

I ever checked the upgrade log, even when 1 image from the tar ball is changed, all the images in this tar ball will be imported again, but not sure the `ctr` will skip those existing ones automatically.

will below option of `ctr image import` be helpful?

```
--base-name="": base image name for added images, when provided only images with this name prefix are imported
```

PR #4227, opened 2023-07-13
improved addon reconcilliation

ibrokethecloud:
PR introduces improved handling for addon reconciliation including detection of addon failures.

This builds on top of existing work from @w13915984028
**Problem:**

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/4108
**Test plan:**
https://github.com/harvester/harvester/pull/4227#issuecomment-1639170965


Dockerfile.dapper1518341745
```diff
@@ -0,0 +1,85 @@
+FROM quay.io/costoolkit/releases-green:luet-toolchain-0.21.2 as luet
```
guangbochen: to remove.

pkg/webhook/resources/addon/validator.go
```diff
@@ -83,12 +83,26 @@ func validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) erro
...
+		return werror.NewBadRequest(fmt.Sprintf("cannot perform operation, as an existing operation is in progress on addon %s", oldAddon.Name))
+	}
+
 	if newAddon.Name == vClusterAddonName && newAddon.Namespace == vClusterAddonNamespace && newAddon.Spec.Enabled {
 		return validateVClusterAddon(newAddon)
 	}
 	return nil
 }
 
+func isAddonSpecChanged(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) bool {
```
guangbochen: seems this func is not used.

pkg/webhook/resources/addon/validator_test.go
```diff
@@ -100,6 +101,197 @@ func Test_validateUpdatedAddon(t *testing.T) {
...
+			name: "user can disable deployed addon",
+			oldAddon: &harvesterv1.Addon{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "addon1",
+				},
+				Spec: harvesterv1.AddonSpec{
+					Repo:          "repo1",
+					Chart:         "chart1",
+					Version:       "version1",
+					Enabled:       false,
```
guangbochen: shoudn't this be `true`?

pkg/webhook/resources/addon/validator_test.go
```diff
@@ -100,6 +101,197 @@ func Test_validateUpdatedAddon(t *testing.T) {
...
+					ValuesContent: "sample",
+				},
+				Status: harvesterv1.AddonStatus{
+					Status: harvesterv1.AddonDisabling,
+				},
+			},
+			expectedError: true,
+		},
+		{
+			name: "user can disable deployed addon",
```
guangbochen, line 139: seems the bellowing case is not matched with the title.

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+	if owned && hc != nil && !hc.DeletionTimestamp.IsZero() {
+		logrus.Warnf("helmChart %s has a deletion timestamp. likely from an older install, wait for GC before re-installing", hc.Name)
+		return h.enqueueAfter(aObj)
+	}
+
+	if !owned && hc != nil {
+		logrus.Warnf("cannot create helmChart %s, as another un-owned chart with same name already exists. requeuing until this is removed", hc.Name)
+	}
+	// ensure condition is set first. This is to ensure the timestamp is before helm chart creation
+	// which allows us to be certain that the job creation timestamp should be after this condition
```
guangbochen, line 172: it is hard to ensure that the timestamp is always before the helmchart/job creation timestamp, it could same on init or late after disabling and re-enabling.
If so, what's te impact of this?

pkg/webhook/resources/addon/validator.go
```diff
@@ -83,12 +83,26 @@ func validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) erro
 		return werror.NewBadRequest("chart field cannot be changed.")
 	}
 
+	if v1beta1.AddonOperationInProgress.IsTrue(oldAddon) {
```
guangbochen, line 86: should allow allow deletion/update when the helmchart does not exist.

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+	hc, _, err := h.getAddonHelmChart(aObj)
+	if err != nil {
+		return aObj, err
+	}
+
+	if hc == nil {
+		// unable to find helmChart. Likely manually GC'd. Reset to initial state
+		// this will allow addon to be processed again and redeployed
+		markCompletedCondition(aObj)
+		aObj.Status.Status = harvesterv1.AddonDisabled
```
guangbochen, line 208: what if its `spec.enabled`` is equal to ture, do we need to change it?

pkg/controller/master/addon/common.go
```diff
@@ -0,0 +1,218 @@
...
+	lastUpdatedTime, err := time.Parse(time.RFC3339, harvesterv1.AddonOperationInProgress.GetLastUpdated(a))
+	if err != nil {
+		return nil, false, fmt.Errorf("error parsing last updated time for AddonOperationInProgress: %v", err)
+	}
+
+	metav1LastUpdatedTime := metav1.NewTime(lastUpdatedTime)
+	logrus.Debugf("last updated time on the addon: %s", lastUpdatedTime)
+	// job creation timestamp should be after the last updated time stamp
+	// on the inprogress condition
+	if j.CreationTimestamp.After(lastUpdatedTime) || j.CreationTimestamp.Equal(&metav1LastUpdatedTime) {
```
guangbochen, line 167: this is not valid if the addon is sucessfully eanbled and upgraded from the previous version. Cuz its job is already finished, but the condition was just patched.

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+
+func (h *Handler) reconcileCurrentInstallationJob(hc *helmv1.HelmChart, aObj *harvesterv1.Addon, state harvesterv1.AddonState) (*harvesterv1.Addon, error) {
+	j, ok, err := h.currentInstallationJob(hc, aObj)
+	if err != nil {
+		return aObj, err
+	}
+
+	// the current job on hc is from the previous deploy/update
+	// or job is not complete. then wait for this to be completed
+	if !ok || !isJobComplete(j) {
```
guangbochen, line 276: should be && ?

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+		return h.enqueueAfter(aObj)
+	}
+
+	if j.Status.Failed > 0 {
+		if harvesterv1.AddonOperationFailed.IsTrue(aObj) {
+			return aObj, nil // no further action possible, until the job is successful or hc is removed
+		}
+		markErrorCondition(aObj, fmt.Errorf("addon deployment job %s failed", j.Name))
```
guangbochen, line 284: get job failed condition reason and message would be more helpful.

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+			return h.addon.UpdateStatus(aObj)
+		}
+	}
+
+	return aObj, nil
+}
+
+// wait for AddonUpgrdes will wait for the helmChart upgrade to be completed
+// and accordingly reconcile state
+func (h *Handler) waitForAddonUpgrades(a *harvesterv1.Addon) (*harvesterv1.Addon, error) {
```
guangbochen: seems not used?

pkg/webhook/resources/addon/validator.go
```diff
@@ -83,12 +83,26 @@ func validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) erro
...
+	}
+
 	if newAddon.Name == vClusterAddonName && newAddon.Namespace == vClusterAddonNamespace && newAddon.Spec.Enabled {
 		return validateVClusterAddon(newAddon)
 	}
 	return nil
 }
 
+func isAddonSpecChanged(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) bool {
+	if newAddon.Spec != oldAddon.Spec {
```
chrisho, line 97: prefer to use `reflect.DeepEqual` instead.
If not any other check, we could simplify it like this

pkg/controller/master/addon/patch_grafana.go
```diff
@@ -37,7 +37,7 @@ func (h *Handler) MonitorAddonRancherMonitoring(key string, aObj *harvesterv1.Ad
 	}
 
 	// disabled successfully, the pv has already from Bound to Released/Available
-	if !aObj.Spec.Enabled && aObj.Status.Status == "" {
+	if !aObj.Spec.Enabled && aObj.Status.Status == harvesterv1.AddonInitState {
```
chrisho: When I continuously switch to enable and disable rancher monitoring, it only creates a new PV.
I think the addon status misses an init status.
image

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+	helmv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+)
+
+const (
+	chartDeletionPrefix     = "helm-delete-"
+	chartInstallationPrefix = "helm-install-"
```
chrisho: The two constants seem not used.

pkg/apis/harvesterhci.io/v1beta1/addon.go
```diff
@@ -1,15 +1,36 @@
...
+
+	AddonInitState AddonState = "" // init status, when an addon is not enabled, or disabled successfully
+)
+
+type AddonOperation string
+
+const (
+	AddonUpdateOperation  AddonOperation = "update"
+	AddonEnableOperation  AddonOperation = "enable"
+	AddonDisableOperation AddonOperation = "disable"
```
chrisho: The three constants seem not used.

pkg/controller/master/addon/common.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	// addon enqueue self interval, defaults to 5s
+	enqueueInterval = 5
+	// addon operation timeout, defaults to 15 min as cluster fisrt setup takes long time, range: [1 (min) ~ 60 (min)]
+	addonTimeout    = 15
+	addonTimeoutMin = 1
+	addonTimeoutMax = 60
```
chrisho: `addonTimeout` `addonTimeoutMin` `addonTimeoutMax` seem not used.

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+		return aObj, err
+	}
+
+	if owned && hc != nil && !hc.DeletionTimestamp.IsZero() {
+		logrus.Warnf("helmChart %s has a deletion timestamp. likely from an older install, wait for GC before re-installing", hc.Name)
+		return h.enqueueAfter(aObj)
+	}
+
+	if !owned && hc != nil {
+		logrus.Warnf("cannot create helmChart %s, as another un-owned chart with same name already exists. requeuing until this is removed", hc.Name)
```
chrisho, line 168: It does not match the warning, it seems to miss the call of the `h.requeueAfter` function.

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+		err := h.helm.Delete(hc.Namespace, hc.Name, &metav1.DeleteOptions{})
+		if err != nil {
+			return aObj, fmt.Errorf("error deleting helm chart: %v", err)
+		}
+		return h.addon.UpdateStatus(aObj)
+	}
+
+	// chart exists and helmChart deletion has been triggered
+	// wait for chart job to be completed
+	if !hc.DeletionTimestamp.IsZero() {
```
chrisho: I think the `!hc.DeletionTimestamp.IsZero()` shouldn't need since the above exists `hc.DeletionTimestamp.IsZero()`.

pkg/controller/master/addon/addon_processing.go
```diff
@@ -0,0 +1,301 @@
...
+			if j.Status.Failed > 0 {
+				markErrorCondition(aObj, fmt.Errorf("addon disable job %s failed", j.Name))
+			}
+
+			if j.Status.Succeeded > 0 {
+				markCompletedCondition(aObj)
+				aObj.Status.Status = harvesterv1.AddonDisabled
+			}
+
+			h.addon.UpdateStatus(aObj)
```
chrisho: It should handle the return error.

pkg/controller/master/addon/common.go
```diff
@@ -0,0 +1,218 @@
...
+	for _, v := range j.Status.Conditions {
+		if v.Type == batchv1.JobFailed && v.Reason == "BackoffLimitExceeded" {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (h *Handler) isAddonUpdateNeeded(aObj *harvesterv1.Addon) (bool, error) {
```
chrisho: It seems not used.

pkg/webhook/resources/addon/validator_test.go
ibrokethecloud: good spot. Old one needs to be `true` and new one needs to be `false`

pkg/webhook/resources/addon/validator_test.go
ibrokethecloud: yeah. fixed it now.

pkg/controller/master/addon/addon_processing.go
ibrokethecloud: the enable/disable job is scheduled by the helm controller after the helmchart object is created. 

the method https://github.com/harvester/harvester/pull/4227/files#diff-1ac4b8e62fbfeb387c7368f8e5d2b5ad6198ec61dfa08043d25d9c43142ee569R276 checks if the timestamp on the job and compares the same with the condition. Unfortunately the timestamp does not have enough resolution to be able to compare in micro-seconds.

I have thought about the possibility of introducing a fixed delay ( say 1 sec) before creating the helmChart object to ensure this condition can be reconciled correctly. 

Even if timestamp is before the condition timestamp, then we consider it as a previous processing attempt and requeue and ignore this job until the correct job is scheduled by the helm controller.

pkg/controller/master/addon/addon_processing.go
ibrokethecloud: that is the aim. if helm chart object is not found, because of manual cleanup, then resetting the status allows the addon to fall through the processing loop again.

This will result in the helm chart being created again since the spec is enabled.

pkg/controller/master/addon/addon_processing.go
ibrokethecloud: there is no easy way to get the job failed reason, as that is not populated in the job status. If we need this then I would need implement logic to fetch the pod logs.

pkg/controller/master/addon/addon_processing.go
ibrokethecloud: it is correct. The idea is to requeue the addon and check again if the following conditions are met:
* job doesnt exist 
* job does exist but is not is not complete.

pkg/webhook/resources/addon/validator.go
ibrokethecloud: If any operation fails the addon will be marked failed. enable/disable/deletion should be possible then. 

The aim is to avoid in-flight updates when a helm chart may already be in a install/upgrade phase.

pkg/controller/master/addon/addon_processing.go
ibrokethecloud: great spot. fixed it.

pkg/controller/master/addon/patch_grafana.go
ibrokethecloud: this is being caused due to install job being terminated by the helm controller when job is being patched.

as a result the pv gets used with the pvc claim but doesnt get cleaned up when `helm install` is terminated by helm controller mid way due to the patched job.

I have moved the patching logic into the controller as it cant be done later and doesnt neccessarily need to be done by a mutating webhook. This ensures that the helm install job is not terminated and pv is correctly used. 

the latest commit has the fixes.

pkg/controller/master/addon/common.go
ibrokethecloud: when addon is upgraded.. it will be moved from `Deployed` to `Updating` status.. and `markInProgress` will be set again. This will update the timestamps on conditions.

When HelmChart is updated with new values another helm install job is created with a newer timestamp which will be equal to or after the addon update is triggered.

guangbochen (commented):
Please help to address the CI error and mitigate the error logs on the Harvester API server; everything else is LGTM, thanks.
```
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868305       7 warnings.go:70] unknown field "metadata.fields"
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868332       7 warnings.go:70] unknown field "metadata.relationships"
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868340       7 warnings.go:70] unknown field "metadata.state"
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868347       7 warnings.go:70] unknown field "status.conditions[0].error"
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868354       7 warnings.go:70] unknown field "status.conditions[0].transitioning"
[harvester-66b78d7cdb-vgmch] time="2023-07-17T05:21:31Z" level=error msg="error syncing 'harvester-system/pcidevices-controller': handler monitor-addon-per-status: jobs.batch \"\" not found, requeuing"
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868361       7 warnings.go:70] unknown field "status.conditions[1].error"
[harvester-66b78d7cdb-vgmch] time="2023-07-17T05:21:31Z" level=error msg="error syncing 'harvester-system/pcidevices-controller': handler monitor-addon-per-status: jobs.batch \"\" not found, requeuing"
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868369       7 warnings.go:70] unknown field "status.conditions[1].transitioning"
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868376       7 warnings.go:70] unknown field "status.conditions[2].error"
[harvester-66b78d7cdb-mrv7h] W0717 05:21:31.868384       7 warnings.go:70] unknown field "status.conditions[2].transitioning"
[harvester-66b78d7cdb-vgmch] time="2023-07-17T05:21:31Z" level=error msg="error syncing 'harvester-system/pcidevices-controller': handler monitor-addon-per-status: jobs.batch \"\" not found, requeuing"
```

pkg/util/constants.go
```diff
@@ -30,43 +30,54 @@ const (
 	AnnotationInsufficientResourceQuota = prefix + "/insufficient-resource-quota"
 
 	AnnotationDefaultUserdataSecret = prefix + "/default-userdata-secret"
+
+	// Add to objects to record upgrade previous version
+	AnnotationUpgradePreviousVersion = prefix + "/upgrade-previous-version"
```
chrisho: seems not used

pkg/util/constants.go
```diff
@@ -30,43 +30,54 @@ const (
...
+
 	// Add to rancher-monitoring addon to record grafana pv name
 	AnnotationGrafanaPVName = prefix + "/grafana-pv-name"
 
+	// Add to addon objects to record user operation and timestamp, the mutator will do this work
+	AnnotationAddonLastOperation          = prefix + "/addon-last-operation"
```
chrisho: ditto

pkg/util/constants.go
```diff
@@ -30,43 +30,54 @@ const (
...
+	AnnotationAddonLastOperationTimestamp = prefix + "/addon-last-operation-timestamp"
```
chrisho: ditto

pkg/util/constants.go
```diff
@@ -30,43 +30,54 @@ const (
...
+
+	// Add to addon objects to adjust the operation timeout
+	AnnotationAddonOperationTimeout = prefix + "/addon-operation-timeout"
```
chrisho: ditto

chrisho (commented):
There are the only places, everything else is LGTM.
These cases are passed:
- Enable/Disable addons.
- Try Disable/Enable in the Enabling/Disabling.
- Update the addon when it is deployed.
- Disable Addons after upgrade from v1.1.2.

chrisho (approved):
LGTM, thanks.

guangbochen (approved):
LGTM, please add appropriate test plans and add the PR to the issue checklist, thanks.

ibrokethecloud:
## Regular addon processing
In terms of UX, there should be no change in enabling/disabling addons.

There will be a slightly longer delay when enabling/disabling addon, as the addon controller waits for the install/uninstall job to complete before marking the job accordingly.

There is a new `Failed` condition, when addon processing fails for any reason.

For example the following `fake-addon` when created and enabled will go to `AddonEnabling` state and have a `Failed` condition set which will be rendered in the UI as shown below:

```
kind: Addon
metadata:
  name: fake-addon
  namespace: harvester-system
spec:
  chart: fake-addon
  enabled: true
  repo: http://harvester-cluster-repo.cattle-system.svc/charts
  valuesContent: |
    image:
      tag: v0.1.0-rc2
    fullnameOverride: fake-addon
  version: 0.1.0-rc2
```

image

The status sub-resource show the following conditions:

```
status:
  conditions:
  - lastUpdateTime: "2023-07-17T11:00:39Z"
    status: "False"
    type: Completed
  - lastUpdateTime: "2023-07-17T11:03:54Z"
    status: "False"
    type: InProgress
  - lastUpdateTime: "2023-07-17T11:03:54Z"
    message: addon deployment job helm-install-fake-addon failed
    reason: Error
    status: "True"
    type: Failed
  status: AddonEnabling
```
## Changes to inflight addon processing
In addition while an addon operation is in progress like enabling/disabling/updating no further changes can be performed on the addon by the user. The new changes in the validating webhook, stops changes on the addon spec if `InProgress` condition is set to true. As a result of this, an addon cannot be disabled/enabled until the last operation has completed.

In the case of `fake-addon` the addon cannot be disabled, until the failure condition is set. At which point it can be disabled.

## Monitoring addon
No change to processing of monitoring addon from v1.2.0-rc3. The grafana pv should be retained and marked available when the monitoring addon is disabled. On subsequently enabling the monitoring addon the grafana pv will be re-used.

pkg/apis/harvesterhci.io/v1beta1/addon.go
```diff
@@ -1,15 +1,31 @@
...
 const (
-	AddonEnabled  AddonState = "AddonEnabled"
-	AddonDeployed AddonState = "AddonDeploySuccessful"
-	AddonFailed   AddonState = "AddonDeployFailed"
+	AddonEnabling  AddonState = "AddonEnabling"
+	AddonDeployed  AddonState = "AddonDeploySuccessful"
+	AddonDisabled  AddonState = "AddonDisabled"
+	AddonDisabling AddonState = "AddonDisabling"
+
+	// after successfully updating, addon will be AddonInitState when !Spec.Enaabled; AddonDeployed when Spec.Enaabled
```
bk201:
```suggestion
	// after successfully updating, addon will be AddonInitState when !Spec.Enabled; AddonDeployed when Spec.Enabled
```
nit: typo in comment.

bk201 (approved):

guangbochen:
@mergify backport v1.2

PR #3779, opened 2023-04-17
feat(setting): sync harvester rancher setting

FrankYang0529:
**Problem:**
Users can modify some settings on both Rancher and Harvester sides. We would like to sync this kind of setting between the two systems.

**Solution:**
Add a sync controller for it.

**Related Issue:**
https://github.com/harvester/harvester/issues/2679

**Test plan:**

### Case 1: Harvester `ui-index` maps to Rancher `ui-dashboard-index`

1. Change Harvester `ui-index`. Rancher `ui-dashboard-index` should also be changed.
2. Set default to Rancher `ui-dashboard-index`. Harvester `ui-index` should also be set to the default value.

### Case 2: Harvester `ui-source` maps to Rancher `ui-offline-preferred`

1. Update values on both sides and we can see the mapping between them.

Auto -> Dynamic
External -> Remote
Bundled -> Local

PR #4250, opened 2023-07-18
changes to remove rancher_manager_support setting

ibrokethecloud:
PR introduces changes to drop setting for `rancher-manager-support` setting.

This is not needed as it has been dropped in favor of the vcluster addon.

**Problem:**

**Solution:**

**Related Issue:**

**Test plan:**


pkg/webhook/resources/setting/validator.go
```diff
@@ -14,9 +14,6 @@ import (
 	"strings"
 
 	"github.com/longhorn/backupstore"
-	// Although we don't use following drivers directly, we need to import them to register drivers.
```
guangbochen: please keep this, otherwise the backup validation would fail

ibrokethecloud: 👍

guangbochen (approved):
LGTM, verified that the Harvester API server is running correctly and contains no outstanding error logs.

chrisho (approved):
LGTM, thanks.

bk201:
Should we also revert https://github.com/harvester/harvester/pull/3779? It's Ok to do it in another PR.

bk201 (approved):

guangbochen:
@mergify backport v1.2

Issue #4156, opened 2023-06-30
[BUG] containerd registry file is empty after upgrading the embedded Rancher to 2.7.5


bk201:
**Describe the bug**

The `/etc/rancher/rke2/registries.yaml` file become empty after upgrading the embedded Rancher to 2.7.5

**To Reproduce**
1. Create a 1.1.2 cluster.
2. Configure some registries with the [`containerd-registry` setting](https://docs.harvesterhci.io/dev/advanced/settings#containerd-registry). 
3. Upgrade to v1.2.0-rc2.
4. The file `/etc/rancher/rke2/registries.yaml` becomes empty.

**Expected behavior**

`/etc/rancher/rke2/registries.yaml` should preserve.

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

Previously, we managed the registry file with [a SUC Plan](https://github.com/harvester/harvester/blob/master/deploy/charts/harvester/templates/cattle-system.yaml#L27-L80) (`cattle-system/sync-containerd-registry`)
In Rancher 2.7.5, the registry file is managed by Rancher (see the [comment](https://github.com/harvester/harvester/issues/4095#issuecomment-1598989144) about Rancher plan change). Here is feedback from the Rancher developer:
> We would expect you to define the registries through your [clusters.provisioning.cattle.io](http://clusters.provisioning.cattle.io/) object and let Rancher designate them -- otherwise, you will need to override the registry with machineGlobalConfig (if possible, it may not actually be possible but I need to look at what we support overriding)

connorkuehl:
Since the clusters.provisioning.cattle.io object is now the source of truth for containerd registries, should we:

1. Update consumers to push/pull from clusters.provisioning.cattle.io instead of the containerd-registry setting? (for example, the Harvester UI operates on the containerd-registry setting but it should be updated to operate on the clusters object); OR
2. Have a controller monitor changes to clusters.provisioning.cattle.io containerd registries and update the containerd-registry setting and vice-versa to keep them in sync with each other? This way consumers would not have to be updated to push/pull from clusters.provisioning.cattle.io.

cc: @bk201 @guangbochen 

[...]

PR #4286, opened 2023-07-20
Propagate containerd-registry setting to Rancher

connorkuehl:

**Problem:**

Starting in Rancher 2.7.5, Rancher has taken over managing the /etc/rancher/rke2/registries.yaml file. The end result is that it ends up overwriting Harvester's changes to that file as Harvester handles updates to the containerd-registry setting.

**Solution:**

In order to avoid dataloss for the containerd-registry setting when upgrading to a version of Harvester with embedded Rancher >= v.2.7.5:

1. Update the settings controller to sync the containerd-registry secret into Kubernetes secrets and update the cluster.provisioning.cattle.io object so that Rancher can render the registries.yaml file.
2. "Replace" the containerd-registry setting in the upgrade path with its current value so that the new settings controller syncs the secrets that Rancher needs.
3. Remove the SUC plan that Harvester once used for syncing the registries.yaml.

**Related Issue:** https://github.com/harvester/harvester/issues/4156

**Test plan:**

1. Default/empty containerd-registry setting:
1.1. Starting from a Harvester cluster that is using the default value for setting/containerd-registry, upgrade to a build of Harvester including this change
1.2. `cat /etc/rancher/rke2/registries.yaml` -> file either doesn't exist or is empty

2. containerd-registry setting only has mirrors
2.1. Configure the containerd-registry setting to use a mirror (i.e., bogus values like "example.com" with endpoint "http://example.com:5000" is fine for the purpose of the test)
2.2. Ensure `cat /etc/rancher/rke2/registries.yaml` consists of the mirror `example.com` with endpoint `http://example.com:5000`.
2.3. Upgrade to a build of Harvester that includes this change
2.4. `cat /etc/rancher/rke2/registries.yaml` consists of the mirror `example.com` with endpoint `http://example.com:5000`.

3. containerd-registry setting has mirrors and configs
3.1. Configure the containerd-registry setting with some values:
3.1.1. mirror: `example.com`
3.1.1.1. endpoint: `http://example.com:5000`
3.1.2. config
3.1.2.1. FQDN: `example.com:5000`
3.1.2.2. User: `admin`
3.1.2.3. Pass: `1234`
3.1.2.4. Auth: `YWRtaW46MTIzNA==` (`echo -n "admin:1234" | base64`)
3.1.2.5. Identity token: `identity1234`
3.2. Ensure `cat /etc/rancher/rke2/registries.yaml` consists of the mirror and configs specified above
3.3. Upgrade to a build of Harvester that includes this change
3.4. Ensure `cat /etc/rancher/rke2/registries.yaml` consists of the mirror and configs specified above. Note that Rancher will probably serialize this as JSON instead of YAML (which Harvester serialized it as).
3.5. Check the Rancher Cluster object's .spec.rkeConfig.registries field (`kubectl get cluster.provisioning.cattle.io local -n fleet-local -o yaml | yq '.spec.rkeConfig.registries'`) and ensure the mirrors match what you entered prior to the upgrade. Check that the `configs` section includes the FQDN entered before the upgrade, and record the value from `authConfigSecretName` which should exist.

```console
# kubectl get cluster.provisioning.cattle.io local -n fleet-local -o yaml | yq '.spec.rkeConfig.registries'
configs:
  example.com:5000:
    authConfigSecretName: harvester-containerd-registry-6c0650f89de1ad48
mirrors:
  example.com:
    endpoint:
      - http://example.com:5000
```

3.6. Confirm the values stored in the secret referred to by `authConfigSecretName` are correct and what was entered prior to the upgrade:

```console
# for key in password username auth identityToken host ; do kubectl get secret harvester-containerd-registry-6c0650f89de1ad48 -o yaml -n fleet-local | yq ".data.$key" | base64 -d  && echo "" ; done
1234
admin
YWRtaW46MTIzNA==
identity1234
example.com:5000
```


package/upgrade/upgrade_manifests.sh
```diff
@@ -1239,6 +1239,28 @@ reuse_vlan_cn() {
...
+  fi
+
+  local patch_file="$UPGRADE_TMP_DIR/patch_containerd_registry.yaml"
+
+  # Note, the trailing whitespace is just to trick Kubernetes into thinking
+  # this is not a no-op. That said, it *will* trigger the settings onChange
+  # handler which will ensure secret/harvester-containerd-registry is properly
+  # synced to Rancher.
+  yq '.value | {"value": . += " "}' "$current" > "$patch_file"
+  kubectl patch setting.harvesterhci.io containerd-registry --patch-file="$patch_file" --type=merge
```
guangbochen: instead of patching the value, maybe add an annotation would be preferred. e.g.,
`kubectl annotate setting.harvesterhci.io containerd-registry harvesterhci.io/upgrade-patched=v1.2.0`

package/upgrade/upgrade_manifests.sh
```diff
@@ -1239,6 +1239,28 @@ reuse_vlan_cn() {
   kubectl get clusternetwork vlan -o yaml | yq '.metadata.finalizers = []' | kubectl apply -f -
 }
 
+sync_containerd_registry_to_rancher() {
+  echo "Sync containerd-registry setting to Rancher"
+
```
guangbochen, line 1244: Currently we do not have version controls in the upgrade script, so would prefer to check if the `clusters.provisioning` registry is not empty and returns directly.

guangbochen:
Just some feedback for the upgrade script, the rest is LGTM. Upgrade from v1.1.2 works perfectly for me; thanks for adding the detailed testing plans.

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM.
- Upgrade from 1.1.2 and my `registries.yaml` file stays.
- Edit the setting from GUI and it works.

starbops (approved):
LGTM! Tested with v1.1.2 upgrades. Registries settings are correctly synced.

guangbochen:
@mergify backport v1.2

Issue #4222, opened 2023-07-12
[BUG] Fail to pack upgrade log: no space left on device


lanfon72:
### Describe the bug
![image](https://github.com/harvester/harvester/assets/5169694/27c6c1ae-7f80-4d05-803a-622ce94b278d)
![image](https://github.com/harvester/harvester/assets/5169694/0d98369a-f5a4-49ad-8d61-82f33bdc53b5)
![image](https://github.com/harvester/harvester/assets/5169694/341ec10c-7a91-4c7b-beb3-56ca693f44fd)

### To Reproduce
1. Install Harvester with any nodes
2. Perform upgrade[^1]
3. Download upgrade log during upgrade

### Expected behavior
Logs should able to be downloaded

## Environment:
 - Harvester ISO version: **v1.1.2** -> **v1.2.0-rc3**
 - **ui-source** Option: **Auto**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **QEMU/KVM 3 nodes**

### Additional context
The upgrade started around 6 hours ago, and stuck in pre-draining status.

### Support bundle
[supportbundle_8365a58f-9f55-4480-a460-b46525cc2098_2023-07-12T12-47-59Z.zip](https://github.com/harvester/harvester/files/12028214/supportbundle_8365a58f-9f55-4480-a460-b46525cc2098_2023-07-12T12-47-59Z.zip)

[^1]: https://docs.harvesterhci.io/v1.1/upgrade/automatic#prepare-an-air-gapped-upgrade

bk201:
cc @starbops, should we increase the volume size? Or is there a way to set the maximum size of logs?

starbops:
The total size of logs generated in regular upgrades shouldn't exceed 1GB, though it's just an estimation. If something is wrong with the upgrade, most of the log lines are repeated, filling up the volume. Increasing the volume size seems not a good solution. Maybe we need to process the log files at a different volume, at least other than the `log-archive` volume.

bk201:
**update** Workaround in the doc PR: https://github.com/harvester/docs/pull/349

A temporary workaround is to expand the volume size to a bigger size (like 10G) in Longhorn GUI and re-download again.
Volume name is in the format `hvst-upgrade-xxxx-upgradelog-log-archive`.

Note if it's an upgrade from 1.1.2 to 1.2.0 and the upgrade doesn't finish yet, online resize is not supported. The user needs to delete these two pods to release the volume once:
[...]

PR #4289, opened 2023-07-21
fix(upgradelog): adjust format of collected logs

starbops:
This is to optimize the disk utilization, especially for the `log-archive` volume.

**Problem:**

Previously, the content of log files may include lots of unnecessary metadata, and we discard them at the log-packaging level. But those data still occupy precious disk space. If the upgrade gets into trouble, there will be bloated log files in the volume, causing the log-packager Job to fail to package the logs (due to no space left). 

**Solution:**

We relieve the issue to a certain degree by the following improvements:

- Trim down the redundant fields (`.kubernetes` metadata) in the log entries at the very beginning of the log pipeline.
- No more JSON processing in the log-packaging step
- Previously generated log archive files will be purged every time a new one is generated
- ~~Skip processing invalid JSON-format lines instead of giving up on the whole file. The problematic lines will be prepended with a `PARSE_ERROR` token.~~
- ~~Expand the default `log-archive` volume to **10Gi**~~

~~Also, we added a timestamp field for every log entry for ease to identify the order among log files.~~

**Related Issue:**

#4222

**Test plan:**

1. Provision a Harvester cluster (single-node is ok) with an ISO image which contains the PR
2. Upgrade the cluster with the same ISO image
3. Download the upgrade log archive by clicking the "Download Log" button on the upgrade dialog
4. Check if the content of the downloaded log files is correct
5. Download the upgrade log archive again
6. Find out the fluentd Pod and exec a shell to check the generated archive file. There should be only one archive file, which is the newest (check the timestamp on the filename):
```
$ kubectl -n harvester-system get pods -l harvesterhci.io/upgradeLogComponent=aggregator
NAME                                             READY   STATUS    RESTARTS   AGE
test-upgrade-master-upgradelog-infra-fluentd-0   2/2     Running   0          44s

$ kubectl -n harvester-system exec -it test-upgrade-master-upgradelog-infra-fluentd-0 -c fluentd -- sh

# ls /archive/
```

### DIFF
--- a/pkg/api/upgradelog/handler.go
+++ b/pkg/api/upgradelog/handler.go
@@ -30,28 +30,16 @@ const (
 	defaultJobBackoffLimit int32 = 5
 	logPackagingScript           = `
 #!/usr/bin/env sh
-set -ex
+set -e
+
+echo "clean up stale archives, if any"
+
+rm -vf /archive/*.zip
 
 echo "start to package upgrade logs"
 
-tmpdir=$(mktemp -d)
-mkdir $tmpdir/logs
-
-cd /archive/logs
-
-for f in *.log
-do
-	awk '{$1=$2=""; print $0}' $f | jq -r .message > $tmpdir/logs/$f || ret=$?
-	if [ -n "$ret" ]; then
-		echo "Failed to process the file $f with ret=$ret"
-		echo "Copy the original file to the destination"
-		cp $f $tmpdir/logs/$f
-		unset ret
-	fi
-done
-
-cd $tmpdir
-zip -r /archive/$ARCHIVE_NAME ./logs/
+cd /archive
+zip -r $ARCHIVE_NAME ./logs/
 echo "done"
 `
 )
--- a/pkg/controller/master/upgradelog/common.go
+++ b/pkg/controller/master/upgradelog/common.go
@@ -21,7 +21,10 @@ import (
 	"github.com/harvester/harvester/pkg/util"
 )
 
-const defaultDeploymentReplicas int32 = 1
+const (
+	defaultDeploymentReplicas   int32 = 1
+	defaultLogArchiveVolumeSize       = "1Gi"
+)
 
 func upgradeLogReference(upgradeLog *harvesterv1.UpgradeLog) metav1.OwnerReference {
 	return metav1.OwnerReference{
@@ -54,7 +57,7 @@ func preparePvc(upgradeLog *harvesterv1.UpgradeLog) *corev1.PersistentVolumeClai
 			},
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
-					"storage": resource.MustParse("1Gi"),
+					"storage": resource.MustParse(defaultLogArchiveVolumeSize),
 				},
 			},
 			StorageClassName: &upgradeLogStorageClassName,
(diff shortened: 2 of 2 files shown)


pkg/api/upgradelog/handler.go
```diff
@@ -41,13 +41,13 @@ cd /archive/logs
...
 for f in *.log
 do
-	awk '{$1=$2=""; print $0}' $f | jq -r .message > $tmpdir/logs/$f || ret=$?
-	if [ -n "$ret" ]; then
-		echo "Failed to process the file $f with ret=$ret"
-		echo "Copy the original file to the destination"
-		cp $f $tmpdir/logs/$f
-		unset ret
-	fi
+  jq -R -r '. as $line | try (fromjson | [.timestamp, .message] | @tsv) catch ("PARSE_ERROR\t" + $line)' $f > $tmpdir/logs/$f || ret=$?
```
w13915984028: L33 `-x` could be removed

w13915984028:
It may not be a good idea to expand the volume size to 10G simply. What will happen if the 10G volume cannot be created given the free space is to be used up?

By contract, we may filter those abnormal files (e.g. size > 20M) and only tail the last e.g. 1000 lines of such files.

The `jq` processing of each line of text is time-consuming, when a file is 100M+, it will be crazy.

The way too big log file will also be a problem for customers to download it, to send to us, to analyze and so on.

bk201:
> By contract, we may filter those abnormal files (e.g. size > 20M) and only tail the last e.g. 1000 lines of such files.

This also means we might lose some useful logs. Upgrade log feature is always the second helpful resource for debugging, the first one is the support bundle. I'll suggest we keep as many logs as possible.

w13915984028:
Suppose something is wrong, if we do not collect the log carefully, it may even taint upgradelog. Break the error chain will be more important than simply collect all of logs.

In most cases, the loop message expands the volume, 1000 lines of logs does not make too much differences with10000 lines. And the root cause may have already been flushed by those loop message, we can only deduce.

pkg/controller/master/upgradelog/common.go
```diff
@@ -249,7 +262,10 @@ func prepareClusterOutput(upgradeLog *harvesterv1.UpgradeLog) *loggingv1.Cluster
 		Spec: loggingv1.ClusterOutputSpec{
 			OutputSpec: loggingv1.OutputSpec{
 				FileOutput: &output.FileOutputConfig{
-					Path:   "/archive/logs/${tag}",
+					Path: "/archive/logs/${tag}",
+					Format: &output.Format{
+						Type: "json",
```
bk201: We can avoid the `jq` part by specifying the [`single_vlaue` format](https://docs.fluentd.org/formatter/single_value).
Here is my experimental ClusterOutput
```
spec:
  file:
    format:
      message_key: message
      type: single_value
```

And it's also possible to make the default output file gzipped ([ref](https://docs.fluentd.org/output/file#compress)):
```
spec:
  file:
    compress: gzip
    format:
      message_key: message
      type: single_value
```
Since pagers can work with gzip, perhaps we can compress files by default and live with 1G volume size. Please help check, thank you!

bk201:
@starbops Please also help check if zip files are clean after downloaded:
```
/archive $ ls /archive/
hvst-upgrade-vzxdd-upgradelog-archive-2023-07-24T09-36-00Z.zip  logs
hvst-upgrade-vzxdd-upgradelog-archive-2023-07-24T10-13-41Z.zip  lost+found
```

pkg/controller/master/upgradelog/common.go
starbops: Thanks for the tips. It works like a charm.

starbops:
Since we get rid of the `jq` part in the log-packaging step, it's less likely the Job will fail or consume lots of computing resources. And the old log archive files are being purged every time a new one is generated. So I changed back to 1GB for a lower system footprint.

I think most of the information we care about is at the front for large log files. The repeated lines that fill the space might flush out the messages that help us deduce the root cause. With that said, it's hard to determine the size cap of each log file.

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):
LGTM!

bk201:
@mergifyio backport v1.2 v1.1

Issue #4315, opened 2023-07-26
[BUG] The favicon icon is incorrect


WuJun2016:
**Describe the bug**

The newly installed harvester-rc4 has an incorrect favicon icon。
![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/8011f478-feb5-4164-8b5e-f25a71882c34)

According to this pr, we expect the value of `v1/management.cattle.io.settings/ui-brand` to be equivalent to harvester, which is now  '' . https://github.com/harvester/harvester/pull/3784
![image.png](https://images.zenhubusercontent.com/60345555ec1db310c78aa2b8/60c899be-8b34-4f58-8ed3-6a96552ee506)

**To Reproduce**

**Expected behavior**

The expected icon should be the harvester icon

**Support bundle**

**Environment**
 - Harvester ISO version:  harvester-rc4
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

noahgildersleeve:
Tested in Harvester `master-a3ee4dcb-head`. Validated that Harvester favicon is showing on the main pages and that it is also showing on the support page. Although on the support page it briefly goes to the Rancher icon then it changes back. 

If harvester is imported into upstream Rancher the icon stays as rancher even when in the Virtualization management pages. I assume this is by design. Closing as fixed.

PR #4322, opened 2023-07-27
Change the default value of ui-brand to harvester

chrisho:

**Problem:**
https://github.com/harvester/harvester/issues/4315

**Solution:**
Change the default value of ui-brand to "Harvester".

**Related Issue:**
https://github.com/harvester/harvester/issues/4315

**Test plan:**
Build and run the harvester image, watch the ui-brand setting, and its default value is "Harvester".

### DIFF
--- a/pkg/controller/master/rancher/embedded_rancher.go
+++ b/pkg/controller/master/rancher/embedded_rancher.go
@@ -14,7 +14,8 @@ import (
 )
 
 var UpdateRancherUISettings = map[string]string{
-	"ui-pl": "Harvester",
+	"ui-pl":    "Harvester",
+	"ui-brand": "harvester",
 }
 
 func (h *Handler) RancherSettingOnChange(key string, setting *rancherv3api.Setting) (*rancherv3api.Setting, error) {


pkg/controller/master/rancher/embedded_rancher.go
```diff
@@ -14,7 +14,8 @@ import (
 )
 
 var UpdateRancherUISettings = map[string]string{
-	"ui-pl": "Harvester",
+	"ui-pl":    "Harvester",
+	"ui-brand": "Harvester",
```
guangbochen: It seems the UI only only recognize the lower case `harvester`, can we make it compatible for both @w13915984028?

WuJun2016:
@chrisho   I tested the UI code and found that using capitalized `Harvester` for the UI was not very friendly, because it involved the case of the file and was used in many places, so I thought that this should continue to be the same as rancher and use lowercase `harvester`.    (for `ui-brand`).  cc @guangbochen

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
LGTM!

guangbochen:
@mergify backport v1.2

PR #4383, opened 2023-08-07
rancher-vcluster addon HEP [skip CI]

ibrokethecloud:

**Problem:**

HEP for `rancher-vcluster` addon.

**Solution:**

**Related Issue:**
**Test plan:**


w13915984028 (commented):
A few questions:

(1) Is the vcluster backed by PV ? When the Rancher is deployed in a vcluster, and the addon is disabled then eanbled, will the previous data still exist?

(2) For harvester system upgrading, will those vcluster affect system or be affected? In the upgrade flow, should any procssing be added?

ibrokethecloud:
the vcluster is deployed to a stateful set which has a pvc backing it for persistence.

i am not yet sure what is likely to happen when we upgrade a single node cluster. for clusters with multiple nodes i see no issues.

w13915984028:
when pvc is used by vcluster addon, could pvc be configured with it's size and the related storageclass?

we enhanced the storage network, when setting it, all pvcs are checked to be detached ( rancher-monitoring is stopped, vm is stopped), also disable vcluster ?

ibrokethecloud:
the default size is 5g. It is not configurable at the moment and only hosts the etcd backend.

w13915984028:
@guangbochen @ibrokethecloud 

We need to estimate the etcd size of a traditional Rancher deployment. Expand the PV size needs some additional work. A proper pre-defined PV size for vcluster to host Rancher, will be a good start.

The PV related info should also be added to HEP.

Meanwhile, storage-network is another point to check, as vCluster will be like `rancher-monitoring`, it uses PV but is not VM.

For upgrading in the single-node or two-node cluster, could there be potential conflicts? seems the related analysis is needed in HEP, and also please @bk201 double check.

w13915984028:
BTW: `vcluster` is new in Harvester, how to trouble-shooting the k8s in k8s, could be added to HEP or Harvester doc chapters like https://docs.harvesterhci.io/v1.0/troubleshooting/harvester .

enhancements/20230807-rancher-vcluster-addon.md
```diff
@@ -0,0 +1,78 @@
...
+
+When enabled a helm chart for vcluster is deployed in the background. 
+
+We leverage the advanced [init-manifests](https://www.vcluster.com/docs/operator/init-manifests) capability of vcluster to inject the following components into vcluster:
+* cert-manager
+* rancher
+
+At present the installed `rancher` leverages a self signed certificate, and this can be extended in the future to allow users to brign their own certificates.
+
+Once the addon is deployed, a rancher endpoint is available at the user defined `hostname`.
```
bk201, line 66: Can we add some info on why the user can access the endpoint via the `hostname`? It's always a mysterious part when I tested the implementation :)

ibrokethecloud: good point. I will add details on how ingress is setup.

ibrokethecloud:
@w13915984028 i tried a single node cluster upgrade with rancher-vcluster addon enabled and had no issues with the upgrade.

w13915984028:
@ibrokethecloud it may not be fully related to the HEP, but please check below code:

https://github.com/harvester/harvester/blob/master/pkg/controller/master/storagenetwork/storage_network.go#L730

enhancement: https://github.com/harvester/harvester/issues/4397

enhancements/20230807-rancher-vcluster-addon.md
```diff
@@ -0,0 +1,80 @@
...
+
+**NOTE** vcluster is deployed with ingress to underlying host cluster enabled. As a result of this `hostname` needs to a valid DNS record pointing to the harvester VIP.
+
+When enabled a helm chart for vcluster is deployed in the background. 
+
+We leverage the advanced [init-manifests](https://www.vcluster.com/docs/operator/init-manifests) capability of vcluster to inject the following components into vcluster:
+* cert-manager
+* rancher
+
+At present the installed `rancher` leverages a self signed certificate, and this can be extended in the future to allow users to brign their own certificates.
```
w13915984028, line 64: brign -> bring

enhancements/20230807-rancher-vcluster-addon.md
```diff
@@ -0,0 +1,80 @@
...
+
+Once the addon is deployed, a rancher endpoint is available at the user defined `hostname`.
+
+### Test plan
+
+* Enable `rancher-vcluster` addon.
+* In a few minutes a rancher installation should be accessible at the predefined `hostname`
+* Import underlying harvester into `rancher-vcluster` and perform MCM tests using this rancher.
+
```
w13915984028, line 73: Please add `Disable rancher-vcluster`  and `Re-enable rancher v-cluster`
When `Disabled`, the Rancher related data is stored in PV, will not be lost

When `Enable` again, the Rancher will recover from previous settings and saved data. 

At this case, does `Bootstrap Password` field still make sense? and can `Hostname` and `Rancher version` be changed when re-enabling?

In certain cases, the `rancher-vcluster` may be corrupted or user wants to throw it away, let's point a way to hardly reset the `rancher-vcluster`, namely remove the PV manually when disabled.

w13915984028 (commented):
please check, thanks.

enhancements/20230807-rancher-vcluster-addon.md
w13915984028: And, unlike other addons, it seems the `rancher-vcluster` may support multi-instances in the future, each `Hostname` will be the identification of the Rancher instance.

ibrokethecloud: i will include the steps for cleaning the PV in the docs PR: https://github.com/harvester/docs/pull/393

enhancements/20230807-rancher-vcluster-addon.md
```diff
@@ -0,0 +1,80 @@
...
+## Proposal
+
+### User Stories
+
+#### Simplify deployment of rancher for managing harvester
+
+##### User Experience In Detail
+
+Harvester leverages Rancher integration for the following key functionality:
+* fine grained rbac
```
bk201: The PR lgtm, can we make the first letter of every list upper case for consistency? Thanks.

bk201 (approved):
LGTM

w13915984028 (approved):
LGTM, thanks.

Issue #4221, opened 2023-07-12
[BUG] VIP is temporarily not accessible when upgrading


Vicente-Cheng:
**Describe the bug**
VIP is temporarily not accessible when upgrading.
The VIP will be unreachable when we upgrade with stage `apply manifest`.
That will cause some services which depend on VIP to go down.

**To Reproduce**
1. create a v1.1.2 cluster
2. upgrade to v1.2.0-rc3
3. Check the VIP connection, it should be down with stage `apply manifest`.

**Expected behavior**
VIP should not be down for more than 30 seconds.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.2.0-rc3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 VM cluster

**Additional context**

bk201:
I can reproduce this issue with these steps:
- Create a single node v1.1.2 cluster.
- Run [this script](https://raw.githubusercontent.com/bk201/misc/main/harvester/issue-4221/upgrade-manifest.sh) to bump the Harvester chart to v1.2.0-rc3. 

VIP is gone and back, but I can't ping it (even in the node):
```
4: mgmt-br: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default q
len 1000
    link/ether 52:54:00:6a:59:7c brd ff:ff:ff:ff:ff:ff
    inet 10.10.0.11/24 brd 10.10.0.255 scope global mgmt-br
       valid_lft forever preferred_lft forever
    inet 10.10.0.100/32 scope global mgmt-br
       valid_lft forever preferred_lft forever
```

```
4: mgmt-br: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default q
len 1000
    link/ether 52:54:00:6a:59:7c brd ff:ff:ff:ff:ff:ff
    inet 10.10.0.11/24 brd 10.10.0.255 scope global mgmt-br
       valid_lft forever preferred_lft forever
```

```
4: mgmt-br: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default q
len 1000
    link/ether 52:54:00:6a:59:7c brd ff:ff:ff:ff:ff:ff
    inet 10.10.0.11/24 brd 10.10.0.255 scope global mgmt-br
       valid_lft forever preferred_lft forever
[...]

PR #4393, opened 2023-08-09
Restart kube-vip pods after updating ingress-expose svc

yaocw2020:

**Problem:**

The VIP is temporarily not accessible during upgrading. 

**Solution:**

The Harvester pod adds the annotation `kube-vip.io/ignore-service-security: "true"` to the service ingress-expose, but the kube-vip can't handle the modification. Just delete the kube-vip pods to let the kube-vip watches service and handle the update even.

**Related Issue:** https://github.com/harvester/harvester/issues/4221

**Test plan:**

- Spin up a Harvester environment with version v1.2.0-rc5
- Delete annotation `kube-vip.io/ignore-service-security: "true"` of the service `kube-system/ingress-expose`.
- Delete kube-vip pod, and we will be unable to ping the VIP.
- Replace the Harvester image in the Harvester mangedchart. We'll be able to ping the VIP after the Harvester updates the service and restart the kube-vip.

### DIFF
--- a/pkg/controller/master/rancher/register.go
+++ b/pkg/controller/master/rancher/register.go
@@ -2,6 +2,7 @@ package rancher
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -11,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/harvester/harvester/pkg/config"
@@ -53,6 +55,7 @@ type Handler struct {
 	SecretCache              ctlcorev1.SecretCache
 	nodeController           ctlcorev1.NodeController
 	podCache                 ctlcorev1.PodCache
+	podClient                ctlcorev1.PodClient
 	Namespace                string
 }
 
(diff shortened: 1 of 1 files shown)


pkg/controller/master/rancher/register.go
```diff
@@ -110,75 +114,90 @@ func (h *Handler) registerExposeService() error {
...
+	svcCopy := svc.DeepCopy()
+	if svc.Annotations == nil {
+		svcCopy.Annotations = make(map[string]string)
+	}
+	svcCopy.Annotations[keyKubevipIgnoreServiceSecurity] = trueStr
+	if _, ok := svcCopy.Annotations[keyKubevipLoadBalancerIPs]; !ok {
+		svcCopy.Annotations[keyKubevipLoadBalancerIPs] = svc.Spec.LoadBalancerIP
+	}
+	if _, err := h.Services.Update(svcCopy); err != nil {
+		return err
```
w13915984028, line 197: when `Services.Update` is successful but `restartKubevipPod` is failed, seems `restartKubevipPod` has no chance to be rescheduled

yaocw2020: Yes, considering the time limit, the PR is just a workaround. I'll submit a PR to kube-vip to address the problem thoroughly.

guangbochen (approved):
LGTM, thanks for the fix, also @yaocw2020 can u please help to file an new to track the upstream VIP enhancement of this, thanks.

bk201 (approved):
Test upgrade and VIP is alive.

Vicente-Cheng (approved):
LGTM, tested with this PR. The VIP looks fine.

yaocw2020:
> LGTM, thanks for the fix, also @yaocw2020 can u please help to file an new to track the upstream VIP enhancement of this, thanks.

https://github.com/harvester/harvester/issues/4407

w13915984028 (approved):
LGTM, thanks.

guangbochen:
@mergify backport v1.2

Issue #4402, opened 2023-08-11
[BUG] Restart of Harvester pod stuck if addon is being processed


khushboo-rancher:
**Describe the bug**
Restart of Harvester pod stuck when any addon is being processed.

**To Reproduce**
1. Enable VM import and trigger an upgrade immediately.
2. The job will be stuck with error `Job has reached the specified backoff limit`

**Expected behavior**
The job should not stuck.

**Support bundle**
From @starbops :
In apply-manifest Pod’s log:
```
managedchart.management.cattle.io/rancher-logging-crd patched
Logging Event Audit: start to convert logging from managedchart to addon
logging has already been addon
Applying extra manifests
No extra manifests in /usr/local/share/extra_manifests/v1.2.0-rc5 to apply
Error from server (BadRequest): admission webhook "validator.harvesterhci.io" denied the request: cannot perform operation, as an existing operation is in progress on addon vm-import-controller
```
The status of the vm-import-controller Addon is still “In Progress”:
```
apiVersion: harvesterhci.io/v1beta1
kind: Addon
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"harvesterhci.io/v1beta1","kind":"Addon","metadata":{"annotations":{},"name":"vm-import-controller","namespace":"harvester-system"},"spec":{"chart":"harvester-vm-import-controller","enabled":false,"repo":"http://harvester-cluster-repo/.
cattle-system.svc/charts","valuesContent":"image:\n  tag: v0.1.4\nfullnameOverride: harvester-vm-import-controller\n","version":"0.1.4"}}
  creationTimestamp: "2023-07-18T19:21:41Z"
  generation: 3
  name: vm-import-controller
  namespace: harvester-system
  resourceVersion: "29899779"
  uid: ef7d7d4f-028d-4fbe-acfb-96480553d8b3
spec:
  chart: harvester-vm-import-controller
  enabled: true
  repo: http://harvester-cluster-repo.cattle-system.svc/charts
  valuesContent: '{"resources":{"requests":{"cpu":"0.5","memory":"4Gi"},"limits":{"cpu":"4","memory":"8Gi"}},"pvcClaim":{"enabled":true,"size":"200Gi","storageClassName":"nvme"}}'
  version: 0.1.4
status:
  conditions:
  - lastUpdateTime: "2023-08-08T17:15:01Z"
    status: "False"
    type: Completed
[...]

PR #4403, opened 2023-08-11
switch addon controller to use helmChart client

ibrokethecloud:

**Problem:**

Currently while an addon is being processed and if Harvester is restarted, the caches may not immediately be available on startup. As a result the addon controller cannot lookup the helm chart for addons currently being processed. 
This results in a scenario where addon controller assumes helm chart has been removed and it disables/enables the addon. This is accomplished by updating the Status on the addon.

Due to this re-processing the addon is `InProgress` and `Completed` condition timestamps are updated. No helm chart upgrade is performed as the spec for HelmChart has not changed.

The updated `InProgress` timestamps result in the install job never being reconcilled successfully as the start time for install job is earlier than the `InProgress` condition.

As a result of these updated timestamps the addon reconciliation never completes and it does not get to a successful or failed state.

**Solution:**

PR introduces minor changes to include the following:
* additional debug logging
* synchronization in controller setup, to ensure controllers are started post leader election before steve is booted.

**Related Issue:**
https://github.com/harvester/harvester/issues/4402
**Test plan:**

### DIFF
--- a/pkg/controller/master/setup.go
+++ b/pkg/controller/master/setup.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/rancher/steve/pkg/server"
 	"github.com/rancher/wrangler/pkg/leader"
-	"github.com/sirupsen/logrus"
 
 	"github.com/harvester/harvester/pkg/config"
 	"github.com/harvester/harvester/pkg/controller/master/addon"
@@ -66,7 +65,6 @@ func register(ctx context.Context, management *config.Management, options config
 func Setup(ctx context.Context, server *server.Server, controllers *server.Controllers, options config.Options) error {
 	scaled := config.ScaledWithContext(ctx)
 
-	startupComplete := make(chan bool, 1)
 	go leader.RunOrDie(ctx, "", "harvester-controllers", controllers.K8s, func(ctx context.Context) {
 		if err := register(ctx, scaled.Management, options); err != nil {
 			panic(err)
@@ -74,11 +72,8 @@ func Setup(ctx context.Context, server *server.Server, controllers *server.Contr
 		if err := scaled.Management.Start(options.Threadiness); err != nil {
 			panic(err)
 		}
-		startupComplete <- true
+		<-ctx.Done()
 	})
 
-	logrus.Debug("waiting for controller boot to complete")
-	<-startupComplete
-	logrus.Debug("controller boot complete")
 	return nil
 }
--- a/pkg/controller/master/setup.go
+++ b/pkg/controller/master/setup.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/rancher/steve/pkg/server"
 	"github.com/rancher/wrangler/pkg/leader"
+	"github.com/sirupsen/logrus"
 
 	"github.com/harvester/harvester/pkg/config"
 	"github.com/harvester/harvester/pkg/controller/master/addon"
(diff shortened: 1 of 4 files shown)


ibrokethecloud:
# to test

* deploy the following fake-addon
```
kind: Addon
metadata:
  annotations:
  name: fake-addon
  namespace: harvester-system
spec:
  chart: fake-addon
  enabled: false
  repo: http://harvester-cluster-repo.cattle-system.svc/charts
  valuesContent: |
    image:
      tag: v0.1.0-rc3
    fullnameOverride: fake-addon
  version: 0.1.0-rc3
```
* enable addon
* check time stamp on conditions `kubectl get addon.harvester fake-addon -n harvester-system -o yaml`
* restart harvester `kubectl rollout restart deploy/harvester -n harvester-system`
* check timestamp on conditions once harvester startup is complete `kubectl get addon.harvester fake-addon -n harvester-system -o yaml`

### before fix:
expect timestamp on conditions to be updated, and addon never gets to failed state. 
`kubectl get addon.harvester fake-addon -n harvester-system -o yaml`
should show `OperationFailed` condition as false.

### after fix:
timestamps on conditions do not change, and addon eventually gets to failed state in a few minutes. 
`kubectl get addon.harvester fake-addon -n harvester-system -o yaml`
should show `OperationFailed` condition as true.

pkg/controller/master/setup.go
```diff
@@ -65,15 +66,19 @@ func register(ctx context.Context, management *config.Management, options config
...
 		}
 		if err := scaled.Management.Start(options.Threadiness); err != nil {
 			panic(err)
 		}
-		<-ctx.Done()
+		startupComplete <- true
 	})
 
+	logrus.Debug("waiting for controller boot to complete")
+	<-startupComplete
```
w13915984028: question:

will all cached data be ready here ?

the leader election is backed by a special CRD object, I suspect this object is ready, does not mean all other CRD objects are ready, please correct me, thanks.

ibrokethecloud: yeah they are ready. only problem as we have found is that it blocks steve from starting up on all nodes in a HA cluster. So i need to fine tune it more or switch helm chart lookup to using the client rather than the cache.

ibrokethecloud: i have changed the controller to use the client and avoid this issue for now.

pkg/controller/master/setup.go
```diff
@@ -65,15 +66,19 @@ func register(ctx context.Context, management *config.Management, options config
...
+	logrus.Debug("controller boot complete")
```
futuretea: 3 harvester pods case,  the pods that are not the leader will always be stuck here?

ibrokethecloud: yeah i realized that during testing. Since looking up helm chart is critical I have switched to using the client in addon controller rather than depending on caches.
The setup change has been rolled back.

w13915984028 (approved):
LGTM, thanks.

futuretea (approved):
LGTM, thanks. Just need to update the solution part in the PR description
```
status:
  conditions:
  - lastUpdateTime: "2023-08-11T08:08:59Z"
    status: "False"
    type: Completed
  - lastUpdateTime: "2023-08-11T08:12:22Z"
    status: "False"
    type: InProgress
  - lastUpdateTime: "2023-08-11T08:12:22Z"
    message: addon deployment job helm-install-fake-addon failed
    reason: Error
    status: "True"
    type: OperationFailed
  status: AddonEnabling
```

starbops (approved):
LGTM, thanks for the speedy fix!

guangbochen:
@mergify backport v1.2

Issue #4360, opened 2023-08-01
[BUG] S3 Secret Access Key Is Exposed in Plain Text on Settings


irishgordo:
**Describe the bug**
Secret Access Key is exposed in plain text for backup settings w/ S3

**To Reproduce**
1. Create S3 Backup Endpoint
1. Click "show backup-target"
1. Secret Access Key is visible in plain text
 
**Expected behavior**
Secret Access Key to be obfuscated from user, not in plain text

**Environment**
 - Harvester ISO version: v1.2-head /  Version: v1.2-a40c4f2f-head 
 - Underlying Infrastructure: qemu/kvm

**Additional context**
![Screenshot from 2023-08-01 14-49-28](https://github.com/harvester/harvester/assets/5370752/7d34b6bc-fc8e-45a2-bcb2-edb917421fa9)

irishgordo:
MinIO seems to by default obfuscate it:
![Screenshot from 2023-08-01 14-55-48](https://github.com/harvester/harvester/assets/5370752/dd4a0f2a-82d4-4c6e-a75d-4509bc327441)
![Screenshot from 2023-08-01 14-55-17](https://github.com/harvester/harvester/assets/5370752/e86c2577-7192-4e01-9a58-fab2e5a78c6f)

irishgordo:
We obfuscate the Secret Access Key on the creation, but not on the viewing:

![Screenshot from 2023-08-01 14-57-04](https://github.com/harvester/harvester/assets/5370752/787d8388-c9a1-4427-bca5-0c07d5c58554)

irishgordo:
Error message also does not obfuscate:
![Screenshot from 2023-08-01 15-10-41](https://github.com/harvester/harvester/assets/5370752/0cc73e0a-a7be-4430-85a3-22e462607ceb)


*please ignore the https error, as this is expected, doing some negotiation of endpoint currently

irishgordo:
Harvester Version, v1.1.2 displays this, in contrast:
![Screenshot from 2023-08-02 14-31-16](https://github.com/harvester/harvester/assets/5370752/94f6e5e8-6063-4694-a798-7064d1a32bb4)

irishgordo:
@futuretea 
This seems to be most triggered in a few different scenarios:
(tested v1.2.0-rc5)

Scenario A:
- for the "first-time" / "original" time in adding a Minio Connection with Custom `additional-ca` cert, ensure that when pasting in the cert, add extra new lines
[...]

PR #4414, opened 2023-08-15
Trim space for all values in backup secret in order to pass the longhorn webhook

futuretea:

**Problem:**

Due to the AWS go client issue https://github.com/aws/aws-sdk-go/issues/689, 
Longhorn has a issue https://github.com/longhorn/longhorn/issues/811, 
so Longhorn add a fix https://github.com/longhorn/longhorn-manager/pull/970
Although aws/aws-sdk-go issue https://github.com/aws/aws-sdk-go/issues/689  has been fixed by https://github.com/aws/aws-sdk-go/pull/690, Longhorn's webhook is still retained. 

If the user add blank lines to the Harvester backup related settings, due to Harvester is using a newer version of the aws-sdk-go , it can pass the Harvester's own backup target setting webhook, but can not  pass the Longhorn's backup secret webhook in https://github.com/longhorn/longhorn-manager/pull/970

After failing to update longhorn's backup secret, method `OnBackupTargetChange`  returns prematurely in [here ](https://github.com/harvester/harvester/blob/a58ac82f3ace6907c9cda3fb4ff9278fc015dde0/pkg/controller/master/backup/backup_target.go#L99), without executing method [reUpdateBackupTargetSettingSecret](https://github.com/harvester/harvester/blob/a58ac82f3ace6907c9cda3fb4ff9278fc015dde0/pkg/controller/master/backup/backup_target.go#L142) which remove the sensitive fields in the `backup-target` setting

**Solution:**

When updating longhorn's backup secret, strim blank lines to avoid being rejected by the longhorn webhook

**Related Issue:**
https://github.com/harvester/harvester/issues/4360

**Test plan:**

1. prepare a https minio server, refer to https://github.com/minio/minio/blob/master/docs/tls/README.md#generate-use-self-signed-keys-certificates
2. for the "first-time" / "original" time in adding a Minio Connection with Custom additional-ca cert, ensure that when pasting in the cert, add extra new lines
3. then fill out the rest for the backup-target, the region, access key, secret access key
4. check the result
- no longhorn webhook error
- no sensitive fields leak

https://github.com/harvester/harvester/issues/4360#issuecomment-1678209639

### DIFF
--- a/pkg/controller/master/backup/backup_target.go
+++ b/pkg/controller/master/backup/backup_target.go
@@ -208,6 +208,11 @@ func getBackupSecretData(target *settings.BackupTarget) (map[string]string, erro
 	data[util.HTTPSProxyEnv] = httpProxyConfig.HTTPSProxy
 	data[util.NoProxyEnv] = util.AddBuiltInNoProxy(httpProxyConfig.NoProxy)
 
+	// trim spaces for all values in order to pass the Longhorn webhook, refer to https://github.com/longhorn/longhorn-manager/pull/970
+	for k, v := range data {
+		data[k] = strings.TrimSpace(v)
+	}
+
 	return data, nil
 }
 


guangbochen (approved):
LGTM, the LH webhook error will be gone after the patch and credentials are removed after a successful update.
image

chrisho (approved):
LGTM, it works expected. Thanks.

futuretea:
@mergify backport v1.2 v1.1

Issue #2260, opened 2022-05-11
[FEATURE] Support network policies for micro-segmentation of workloads for underlay networks


janeczku:
**Is your feature request related to a problem? Please describe.**
Use case: Need to enforce traffic policies for VM-VM and VM-External communication within a VLAN network.

**Implementation details**

Should leverage the native Kubernetes network policies subsystem.

w13915984028:
connect to:
https://github.com/harvester/harvester/issues/2236

yasker:
@janeczku The use case is Kubernetes guest cluster or just VMs on the VXLAN? Since if it's Kubernetes then the guest Kubernetes can enforce the network policy.

janeczku:
The use case is VMs on the VXLAN and VLAN.

innobead:
Relevant to https://github.com/harvester/harvester/issues/10154. 

cc @harvester/network

PR #4456, opened 2023-08-28
Security Groups for Virtual Machines

ibrokethecloud:
PR introduces the `SecurityGroup` CRD which is used to control network access to VM workloads using bridge interfaces.

**Problem:**

**Solution:**

The PR introduces a sidecar `vm-network-policy` which is injected into VM launcher pods where a `SecurityGroup` is defined by annotations.

The sidecar watches the SecurityGroup and updates the iptables chain on the tap interfaces in the launcher pod to control ingress to VM workloads.

**Related Issue:**
https://github.com/harvester/harvester/issues/2260

**Test plan:**


Faker523:
hello, btw ,this pr will release in v1.4?

ibrokethecloud:
we are currently targeting 1.5

Issue #4478, opened 2023-09-04
[FEATURE] Allow users to set count/size limit on snapshots


FrankYang0529:
**Is your feature request related to a problem? Please describe.**
Snapshots use extra storage. Currently, users don't have a way to control snapshots count/size and check how much size snapshot uses. Snapshots may use more storage than users expect.

**Describe the solution you'd like**
Leverage [LH snapshot space management](https://github.com/longhorn/longhorn/issues/6563) feature to achieve it.

tserong:
Does this need a "require/doc" label? Also, do we need GUI changes for this feature, or are users expected to set the annotations up by editing the annotations in the volume YAML?

m-ildefons:
Needs documentation update.
It's unclear what units are expected, e.g.:
`.spec.snapshotLimit.namespaceTotalSnapshotSizeQuota: 100` is this 100 snapshots, 100 bytes or 100 GiB?

TachunLin:
Verified **fixed** on Harvester `v1.4.0-rc3` with Rancher `v2.9.2`. Close this issue

### Result

<details><summary>$\color{green}{\textsf{PASS}}$ Check <strong>namespaceTotalSnapshotSizeUsage</strong> and <strong>vmTotalSnapshotSizeUsage</strong> values format $~~$</summary>

1. Can't set `namespaceTotalSnapshotSizeUsage` to negative value 
  ```
  # Please edit the object below. Lines beginning with a '#' will be ignored,
  # and an empty file will abort the edit. If an error occurs while saving this file will be
  # reopened with the relevant failures.
  #
  # resourcequotas.harvesterhci.io "default-resource-quota" was not valid:
  # * spec.snapshotLimit.namespaceTotalSnapshotSizeQuota: Invalid value: -1: spec.snapshotLimit.namespaceTotalSnapshotSizeQuota in body should be greater than or equal to 0
  #
  apiVersion: harvesterhci.io/v1beta1
  kind: ResourceQuota
  ...
  spec:
    snapshotLimit:
      namespaceTotalSnapshotSizeQuota: -1
      vmTotalSnapshotSizeQuota:
        vm: 32212254720
  ```

2. Can't set `vmTotalSnapshotSizeUsage` to negative value
  ```
  # Please edit the object below. Lines beginning with a '#' will be ignored,
  # and an empty file will abort the edit. If an error occurs while saving this file will be
  # reopened with the relevant failures.
  #
[...]

PR #4497, opened 2023-09-07
feat(hep): add snapshot space management

FrankYang0529:
**Proposal:**
Initial HEP to introduce snapshot space management.

**Related Issue:**
https://github.com/harvester/harvester/issues/4478

### DIFF
--- a/enhancements/20230907-snapshot-space-management.md
+++ b/enhancements/20230907-snapshot-space-management.md
@@ -1,0 +1,94 @@
+# Snapshot Space Management
+
+## Summary
+
+This proposal leverages [LH Snapshot Space Management](https://github.com/longhorn/longhorn/issues/6563) to control snapshot space usage in the system. LH introduces two new attributes `snapshotMaxCount` and `snapshotMaxSize` in Volume. Harvester can use these parameters to set upper bound for snapshot space usage.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/4478
+
+## Motivation
+
+### Goals
+
+- Set snapshot count and size limitations for each volume.
+
+## Proposal
+
+Use a PVC controller to watch PVCs. If there is `harvesterhci.io/snapshotMaxCount` or `harvesterhci.io/snapshotMaxSize` annotation on a PVC, the system uses the value to update the related LH Volume.
+
+### User Stories
+
+With snapshot space management, snapshot space usage is controllable. Users can set snapshot limitation and evaluate maximum space usage.
+
+### User Experience In Detail
+
+Before snapshot space management:
+The default maximum snapshot count for each volume is `250`. This is a constants value in the [source code](https://github.com/longhorn/longhorn-engine/blob/8b4c80ab174b4f454a992ff998b6cb1041faf63d/pkg/replica/replica.go#L33) and users don't have a way to control it. If a volume size is 1G, the maximum snapshot space usage will be 250G.
+
+After snapshot space management:
+There is a configurable default maximum snapshot count setting. Users can update it to overwrite the fixed value in the system.
+Users can set different maximum snapshot count and size on each volume. A more important volume can have more snapshot space usage.
+
+### API changes
+
+No harvester API changes are needed. The changes will only be in the pvc controller logic.
+
+## Design
+
+### Implementation Overview
+
+#### Settings
+
+Add a harvester setting `snapshot-max-count`. The range of the value is same as LH setting `snapshot-max-count`. It should be between `2` to `250`. The [Setting controller](https://github.com/harvester/harvester/blob/ea814927a14f99b0f0ef9fe74c37dbea3e6e95df/pkg/controller/master/setting/register.go#L66-L80) will overwrite the value to LH setting.
+
+#### PVC webhook
+
+Add annotations value check to [PVC webhook](https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/persistentvolumeclaim/validator.go).
+
+* `harvesterhci.io/snapshotMaxCount`: The default value is from the harvester setting `snapshot-max-count`. Users can overwrite the default value by setting the annotation.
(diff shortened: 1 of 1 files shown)


enhancements/20230907-snapshot-space-management.md
```diff
@@ -0,0 +1,94 @@
...
+With snapshot space management, snapshot space usage is controllable. Users can set snapshot limitation and evaluate maximum space usage.
+
+### User Experience In Detail
+
+Before snapshot space management:
+The default maximum snapshot count for each volume is `250`. This is a constants value in the [source code](https://github.com/longhorn/longhorn-engine/blob/8b4c80ab174b4f454a992ff998b6cb1041faf63d/pkg/replica/replica.go#L33) and users don't have a way to control it. If a volume size is 1G, the maximum snapshot space usage will be 250G.
+
+After snapshot space management:
+There is a configurable default maximum snapshot count setting. Users can update it to overwrite the fixed value in the system.
+Users can set different maximum snapshot count and size on each volume. A more important volume can have more snapshot space usage.
```
connorkuehl: With regards to user experience, could we also plan for the UI to warn the user if something like the following scenario occurs:

* Volume has 10 successful snapshots
* User changes `snapshot-max-count` to a value <= 10

The UI could display a warning banner stating that the new value prevents snapshots from being executed on volumes whose snapshot amount >= the new value?

enhancements/20230907-snapshot-space-management.md
```diff
@@ -0,0 +1,94 @@
...
+
+No harvester API changes are needed. The changes will only be in the pvc controller logic.
+
+## Design
+
+### Implementation Overview
+
+#### Settings
+
+Add a harvester setting `snapshot-max-count`. The range of the value is same as LH setting `snapshot-max-count`. It should be between `2` to `250`. The [Setting controller](https://github.com/harvester/harvester/blob/ea814927a14f99b0f0ef9fe74c37dbea3e6e95df/pkg/controller/master/setting/register.go#L66-L80) will overwrite the value to LH setting.
```
bk201: @innobead Do we know what's the max number of system snapshots? Is there a limitation? Thanks!

enhancements/20230907-snapshot-space-management.md
FrankYang0529: Yes, it's good suggestion to notify users when there is not enough count or size. At first version, I would like to forbid users from narrowing down `snapshot-max-count` & `snapshot-max-size`. In LH, system also can create snapshot for rebuilding a volume. If we allow to narrow down count or size, the system may become unstable.

FrankYang0529: After some implementations, we have to allow users to narrow down `snapshot-max-count` & `snapshot-max-size` in LH. The reason is that we don't operate LH Volume CR directly. We use PVC and StorageClass for it. If we use annotations on PVC to control `snapshot-max-count` & `snapshot-max-size` in LH Volume, we have to allow narrowing down operations.

In LH, it will use `Engine.Status.Snapshots` information to check whether new `snapshot-max-count` & `snapshot-max-size` are valid. However, the information from `Engine.Status.Snapshots` may be stale, because it's updated [every 5 seconds](https://github.com/longhorn/longhorn-manager/blob/e9f5eac31f56056649bc45a2d740a9d6d10cadd8/controller/engine_controller.go#L51). In some cases, the real snapshot usage may be more than `snapshot-max-count` & `snapshot-max-size`. For example, a user create a new snapshot and another user narrows down `snapshot-max-count`. At this moment, the new snapshot hasn't been reflected on Engine CR, so the system will allow the operation. In this case, users don't break the system, but the real snapshots usage is more than `snapshot-max-count` & `snapshot-max-size`.

connorkuehl (approved):

ibrokethecloud:
based on our discussion today
* do we allow this annotation at the vm level to ensure that all volumes attached with a vm get this setting applied to them?
* do we leverage snapshot crd, to count number of snapshots and ensure we can automatically gc older snapshots when limit is reached?
* we need UI changes to inform users it a snapshot limit is about to be reached
* the harvester setting for snapshot count needs to be considered the source of truth as most of our mutation logic is going to check default value from harvester settings before overriding the same from VM level / volume level setting if present.

FrankYang0529:
Hi @ibrokethecloud, thanks for the comments. After discussing with @bk201, we have some conclusion:

1. Discard count limitation in Harvester. From admin view, they care about how much size does snapshot use, not how many count. We don't need to set count limitation to each volume, because this may break data path in LH side. We just leave it as default value as `250`.
2. Use annotation to support size limitation at VM level. User can set annotation like `harvesterhci.io/maxTotalSnapshotSize` on VM. When there is a new VMBackup, webhook will check the annotation and sum snapshot usage in the VM. If total size is bigger than the annotation, deny the request.
3.  Use annotation to support size limitation at Namespace level. User can set annotation like `harvesterhci.io/maxTotalSnapshotSize` on Namespace.  When there is a new VMBackup / VolumeSnapshot, webhook will check the annotation and sum snapshot usage in each volume. If total size is bigger than the annotation, deny the request.

With this change, we don't set snapshot max size / count to LH volumes, so we can minimize risk like rebuilding volume without enough snapshot quota. Originally, we create a feature in LH side to make sure it's hard limit. However, from Harvester perspective, we can only do best effort to have soft limit, because we can only get status from CR. There is always time gap between status in CR and real status in engine process. From Harvester, we don't have a way to get real status in engine process.

ibrokethecloud (approved):
lgtm. thanks.

enhancements/20230907-snapshot-space-management.md
```diff
@@ -0,0 +1,117 @@
+# Snapshot Space Management
+
+## Summary
+
+This proposal introduces soft limitation about snapshot size. Users can set maximal total snapshot size at VM or Namespace level.
+Before creating a new VMBakcup or VolumeSnapshot, webhook will check snapshot size usage from LH Engine status. This proposal focuses on snapshot size usage in the system, not backup target.
```
a110605: typo `VMBakcup`

FrankYang0529: Fixed it. Thank you.

bk201 (approved):
LGTM, thanks!

FrankYang0529:
@mergify backport v1.4

Issue #4500, opened 2023-09-07
[BUG] Upgrade log packaging format error


starbops:
**Describe the bug**

When upgrading from v1.1.2 to v1.2.0-rc6, downloaded upgrade log files don't get handled correctly. For example:

```
...
2023-09-07T07:47:09+00:00	cattle-system.rancher-64f6b99dfb-w6h6c.rancher	{"stream":"stdout","logtag":"F","message":"2023/09/07 07:47:09 [INFO] Rancher version v2.7.5 (eef55334b) is starting","kubernetes":{"pod_name":"rancher-64f6b99dfb-w6h6c","namespace_name":"cattle-system","pod_id":"06762e27-89a8-4f54-9499-b74c11a69c5f","labels":{"app":"rancher","pod-template-hash":"64f6b99dfb","release":"rancher"},"annotations":{"cni-projectcalico-org/containerID":"eb8003158ff88c75ed625639ae05d2b13287719b3e3f6d8d50e1c87ea4d6032b","cni-projectcalico-org/podIP":"10.52.0.97/32","cni-projectcalico-org/podIPs":"10.52.0.97/32","harvesterhci-io/timestamp":"2023-09-07T07:00:37Z","k8s-v1-cni-cncf-io/network-status":"[{\n    \"name\": \"k8s-pod-network\",\n    \"ips\": [\n        \"10.52.0.97\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]","k8s-v1-cni-cncf-io/networks-status":"[{\n    \"name\": \"k8s-pod-network\",\n    \"ips\": [\n        \"10.52.0.97\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]","kubernetes-io/psp":"global-unrestricted-psp"},"host":"harvester-9v9tt","container_name":"rancher","docker_id":"4277f0873fb6a53218b927f657dfe93e0545e62fe89004409b69ca188f879c84","container_hash":"sha256:81ee0878ffcdc37e87f0c4a4b4d528763ad24442aaf07475370d15e1d887cb54","container_image":"docker.io/rancher/rancher:v2.7.5"}}
...
```

The log line is in JSON format with many unneeded metadata. However, the only log message we want is the following:

```
...
2023/09/07 07:47:09 [INFO] Rancher version v2.7.5 (eef55334b) is starting
...
```

[...]

PR #4501, opened 2023-09-07
fix(upgradelog): handle different log packaging cases

starbops:

**Problem:**

The log packaging script will vary according to different version of controllers. But the log collecting rules of fluentd are not. This introduces inconsistency in log packaging behaviors. This commit tries to take care of both cases.

**Solution:**

Add a flag to determine if the logs are in the plain text JSON format (`*.log`). If it's true, proceeded with the older version post-processing (`jq` then package). Otherwise, the packager script will archive the logs directly (log files are already in the gzip format while being collected by fluentd).

**Related Issue:**

#4500 

**Test plan:**

There are two cases that needs to be tested:

Upgrading from v1.1.2

1. Prepare a v1.1.2 Harvester cluster (single node would be fine)
2. Upgrade with the ISO containing the fix
3. Click the "Download Log" button on the upgrade dialog before entering **Upgrading System Service** phase
4. Check the downloaded log content, should not be in the JSON format
5. Wait until the upgrade passed the **Upgrading System Service** phase
6. Click the "Download Log" button again
7. Check the downloaded log content, should not be in the JSON format

Same version upgrade (simulating future upgrades)

1. Prepare a Harvester cluster provisioned with the ISO containing the fix
2. Upgrade with the same ISO
3. Click the "Download Log" button
4. Check the downloaded log content, should be in gzip format

### DIFF
--- a/pkg/api/upgradelog/handler.go
+++ b/pkg/api/upgradelog/handler.go
@@ -33,13 +33,42 @@ const (
 set -e
 
 echo "clean up stale archives, if any"
-
 rm -vf /archive/*.zip
 
 echo "start to package upgrade logs"
 
+tmpdir=$(mktemp -d)
+mkdir $tmpdir/logs
+
+# We used to collect logs in plain text until v1.2.0, the packaging script
+# should be able to determine what kind of post-processing to execute.
+# Starting from v1.2.0, the upgrade logs will be collected in .gzip format.
+# For the is_plain_log flag, 0 means the sampled file is in plain text format.
+is_plain_log=0
+
 cd /archive
-zip -r $ARCHIVE_NAME ./logs/
+sampled_file=$(find logs -type f -print | head -1)
+if [ "${sampled_file: -4}" != ".log" ]; then
+	is_plain_log=1
+fi
+
+if [ "$is_plain_log" -eq 0 ]; then
+	for f in logs/*.log; do
+		awk '{$1=$2=""; print $0}' $f | jq -r .message > $tmpdir/$f || ret=$?
+		if [ -n "$ret" ]; then
+			echo "Failed to process the file $(basename $f) with ret=$ret"
+			echo "Copy the original file to the destination"
+			cp $f $tmpdir/$f
+			unset ret
+		fi
+	done
+
+	cd $tmpdir
+	zip -r /archive/$ARCHIVE_NAME ./logs/
+else
+	zip -r $ARCHIVE_NAME ./logs/
+fi
+
 echo "done"
 `
 )


pkg/api/upgradelog/handler.go
```diff
@@ -33,13 +33,38 @@ const (
...
+# We used to collect logs in plain text until v1.2.0, the packaging script
+# should be able to determine what kind of post-processing to execute.
+# Starting from v1.2.0, the upgrade logs will be collected in .gzip format.
+is_plain_log="false"
+
 cd /archive
-zip -r $ARCHIVE_NAME ./logs/
+for f in logs/*.log
+do
+    is_plain_log="true"
```
w13915984028: few questions:
1. `is_plain_log` has leading space instead of tab, comparing to other lines
2. `is_plain_log` is set to `true` anyway in the do loop, is it wrong ?
3. L61 assumes that the *.log is either plain text, or in given format, there is no possibility that they are mixed,  then,  the code could be optimized as:

(1) check if any file is in `is_plain_log`  format, ant set the variable, break;  or simply check the first file to determin the potential format;  or simply use command like `file` to check the format of of
```
file sidebars.js 
sidebars.js: ASCII text

file sb_e040_20220510.zip 
sb_e040_20220510.zip: Zip archive data, at least v1.0 to extract

```
(2) if  `is_plain_log`, then `awk` and `jq` and copy all file to the tmpdir

bk201:
I got this when downloading after apply_manifest, not sure if I did something wrong:
```
clean up stale archives, if any
start to package upgrade logs
sh: -c: line 21: syntax error near unexpected token `)'
```

pkg/api/upgradelog/handler.go
starbops: 1. Good spot, thanks.
2. It is a stupid way of doing. I've changed the way we detect the format.
3. Same as above.

Thanks for the suggestion!

starbops: Sadly, we're not able to bring the `file` command into v1.1.2's `harvester-upgrade` image, so I'll just fallback to the original way.

bk201 (approved):
LGTM.
These paths look good.
- 1.1.2 -> 1.2.0-rc6+ 
- 1.2.0-rc6+ -> 1.2.0-rc6+

guangbochen (approved):
LGTM, thanks.
btw @w13915984028 can u PTAL of this again as we plan to add this to the v1.2.0 release, thanks.

guangbochen:
@mergify backport v1.2

w13915984028 (approved):
LGTM, thanks.

starbops:
@mergify backport v1.1

Issue #4545, opened 2023-09-19
[BUG] Unable to setup backup-target with error: There is a VM either in the process of creating or deleting


lanfon72:
### Describe the bug
https://github.com/harvester/harvester/assets/5169694/73ac1dd0-4e34-4d6d-bbfc-8163945a383d

### To Reproduce
1. Install Harvester with any nodes
2. Setup **backup-target**
3. Create an image for VM creation
4. Create VM `vm1` and take backup `vm1b`
5. Restore `vm1b` into new VM `VM1r`
6. Reset **backup-target** to default value
7. Setup **backup-target**

### Expected behavior
We should able to change backup-target, there haven't any VMs. 

## Environment:
 - Harvester ISO version: **v1.2-a9968541-head**
 - **ui-source** Option: **Auto**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 4 nodes**

### Additional context
I'm not 100% sure that the root cause is related to #4544, but hit the bug after it.  (so I labeled `reproduce/often` not `reproduce/always`)

### Support bundle
The SB size is too large (>25MB), separated into following:
[harvester-tchfx.zip](https://github.com/harvester/harvester/files/12662679/harvester-tchfx.zip)
[harvester-d5hzw.zip](https://github.com/harvester/harvester/files/12662680/harvester-d5hzw.zip)
[harvester-d2csm.zip](https://github.com/harvester/harvester/files/12662681/harvester-d2csm.zip)
[harvester-gv6lk.zip](https://github.com/harvester/harvester/files/12662682/harvester-gv6lk.zip)
[supportbundle_c6db0211-4e97-404e-bb78-72f4606f5a86_2023-09-19T14-55-03Z.zip](https://github.com/harvester/harvester/files/12662688/supportbundle_c6db0211-4e97-404e-bb78-72f4606f5a86_2023-09-19T14-55-03Z.zip)

FrankYang0529:
After some investigation, I think we may need to prevent users from giving uppercase letters.

```
VirtualMachineRestore encountered error
VirtualMachine.kubevirt.io "VM1r" is invalid: metadata.name: Invalid value: "VM1r": a lowercase RFC 1123 subdomain mus
t consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.
[...]

Issue #4544, opened 2023-09-19
[BUG] Unable to restore backup into new VM when the name starts with upper case


lanfon72:
### Describe the bug
https://github.com/harvester/harvester/assets/5169694/04ee7389-0049-4efb-8a29-9c0ae6ee6678

### To Reproduce
1. Install Harvester with any nodes
2. Setup backup-target
3. Create an image for VM creation
4. Create VM `vm1` and take backup `vm1b`
5. Restore `vm1b` into new VM `VM1r`

### Expected behavior
VM should be restored correctly, or error message should prompted.

## Environment:
 - Harvester ISO version: **v1.2-a9968541-head**
 - **ui-source** Option: **Auto**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 4 nodes**

### Additional context
N/A

### Support bundle
N/A

albinsun:
Provide test result on `master-645d9e2d-head`, close as fixed
Backport  `v1.1`  (harvester#4552) is not merge yet, let's verify it separatly later


## Environment
* Harvester
  * Version: `master-645d9e2d-head`
  * Profile: QEMU/KVM, single node (8C/48G/250G)
  * ui-source: `Auto`


## Test Cases
> **Note** Click "►" to expand details.

lower alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character

### Positive TCs
1. Single lower ${~~~\color{green}\textsf{V}}$
1. Lowers ${~~~\color{green}\textsf{V}}$
1. Lowers contains '.'  ${~~~\color{green}\textsf{V}}$
1. Lowers contains '-' ${~~~\color{green}\textsf{V}}$
1. Lowers contains '.' and '-' ${~~~\color{green}\textsf{V}}$

![image](https://github.com/harvester/harvester/assets/2773781/17fea11e-a266-484d-a9d4-3e3af1624d45)


### Negtive TCs
1. <details>
   <summary>
   Upper
   ${~~~\color{green}\textsf{V}}$
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/b2411e02-e0c1-4fef-b996-997c8c827862)
   </details> 

1. <details>
   <summary>
   Upper infront of valid
   ${~~~\color{green}\textsf{V}}$
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/15f47599-19d8-470e-8c1a-bea7e9c3a28b)
   </details> 

[...]

PR #4551, opened 2023-09-20
fix(vmrestore): check target vm name

FrankYang0529:
**Solution:**
Resource name should follow [RFC 1123](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names). Update VMRestore webhook to check it.

**Related Issue:**
https://github.com/harvester/harvester/issues/4545
https://github.com/harvester/harvester/issues/4544

**Test plan:**

### Setup
1. Create a VM `vm1`.
2. Setup backup target.
3. Create a VMBackup `vm1b`.

### 1. Empty name case
1. Apply following VMRestore. Webhook should return error.
```yaml
kind: VirtualMachineRestore
metadata:
  name: restore-vm1b-test
  namespace: default
spec:
  newVM: true
  target:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: ""
  virtualMachineBackupName: vm1b
```

### 2. Uppercase name
1. Restore `vm1b` to a new VM `VM1r`. Webhook should return error.

### 3. Valid name
1. Restore `vm1b` to a new VM `vm1r.pass`. A new VM should be created.
2. Try to update the target VM name to `vm1r.pass.edit`. Webhook should return error.

### DIFF
--- a/pkg/webhook/resources/virtualmachinerestore/validator.go
+++ b/pkg/webhook/resources/virtualmachinerestore/validator.go
@@ -2,12 +2,14 @@ package virtualmachinerestore
 
 import (
 	"fmt"
+	"reflect"
 
 	ctlv1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	validationutil "k8s.io/apimachinery/pkg/util/validation"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
@@ -24,6 +26,7 @@ import (
 )
 
 const (
+	fieldSpec                     = "spec"
 	fieldTargetName               = "spec.target.name"
 	fieldVirtualMachineBackupName = "spec.virtualMachineBackupName"
 	fieldNewVM                    = "spec.newVM"
@@ -72,6 +75,7 @@ func (v *restoreValidator) Resource() types.Resource {
 		ObjectType: &v1beta1.VirtualMachineRestore{},
 		OperationTypes: []admissionregv1.OperationType{
 			admissionregv1.Create,
+			admissionregv1.Update,
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/webhook/resources/virtualmachinerestore/validator.go
```diff
@@ -24,6 +26,7 @@ import (
 )
 
 const (
+	fieldSpec                     = "spec"
```
connorkuehl, line 29: Is this unused?

pkg/webhook/resources/virtualmachinerestore/validator.go
```diff
@@ -120,6 +124,15 @@ func (v *restoreValidator) Create(request *types.Request, newObj runtime.Object)
 	return v.handleExistVM(newVM, vm)
 }
 
+func (v *restoreValidator) Update(request *types.Request, oldObj, newObj runtime.Object) error {
+	oldRestore := oldObj.(*v1beta1.VirtualMachineRestore)
+	newRestore := newObj.(*v1beta1.VirtualMachineRestore)
+	if !reflect.DeepEqual(oldRestore.Spec, newRestore.Spec) {
+		return werror.NewInvalidError("VirtualMachineRestore spec is immutable", fieldTargetName)
```
connorkuehl: Would it be alright to just always return this error in the Update handler without running a comparison?

connorkuehl (changes requested):

pkg/webhook/resources/virtualmachinerestore/validator.go
ibrokethecloud: i guess status will likely change as request progresses to completion, and annotations may get updated during the processing of request.

ibrokethecloud: seems like we need to use `fieldSpec` here instead of `fieldTargetName`

pkg/webhook/resources/virtualmachinerestore/validator.go
FrankYang0529: I used wrong field in update webhook. Updated it. Thank you.

pkg/webhook/resources/virtualmachinerestore/validator.go
FrankYang0529: Yes, we need to allow controllers to update status, so we only check `spec` field.

FrankYang0529: Yes, udpated it. Thank you.

connorkuehl: Ah right, of course. Thanks, both!

connorkuehl (approved):

ibrokethecloud (approved):
LGTM. thanks.

WebberHuang1118 (approved):
Testing cases work as expectation,
LGTM, thanks.

WebberHuang1118:
@Mergifyio backport v1.2

WebberHuang1118:
https://github.com/Mergifyio backport v1.1

Issue #4520, opened 2023-09-12
[BUG] Unable to use image that has a name more than ~40 characters long created through terraform


jaevans:
**Describe the bug**

Any image that has a name more than 32 characters long ends up with a backingImage with a different name than Harvester expects

**To Reproduce**
1. Use terraform to create a new Image
```
resource "harvester_image" "ubuntu_22_04" {
  display_name = "Ubuntu Cloud 22.04"
  name         = "jammy-server-cloudimg-amd64"
  namespace    = "harvester-public"
  description  = "Ubuntu 22.04 LTE Offical Cloud Image"
  source_type  = "download"
  url          = "https://cloud-images.ubuntu.com/jammy/current/jammy-server-cloudimg-amd64.img"
}
```
2. Create a VM from the image

**Expected behavior**
VM is created and starts correctly

**Support bundle**

**Environment**
 - Harvester ISO version: 1.2.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Baremetal HP DL380

**Additional context**
The PVC reports an error:
```
3m9s        Warning   ProvisioningFailed       persistentvolumeclaim/test-disk-0-c0b7u                failed to provision volume with StorageClass "longhorn-jammy-server-cloudimg-amd64": rpc error: code = InvalidArgument desc = volume pvc-0c37e4c6-053c-4da2-93d7-fed52238a01b missing backing image harvester-public-jammy-server-cloudimg-amd64 unable to create volume
```
The `backingimage` name is truncated to 32 characters, and a random suffix is added. This seems to be a change in longhorn between harvester 1.1.2 and 1.2.0.
```
$ k get backingimage -n longhorn-system | grep jammy
harvester-public-jammy-server-c-32ce4c2c   f3dbe861   download     670302208    41h
```

This also causes restores of images created under 1.1.2 to fail. 

The same image created via the GUI results in a `backingimage` with a short, random name
```
longhorn-system   harvester-public-image-jgh7d               03d4c6d4   download     670302208    8m48s
```

WebberHuang1118:
Longhorn has a 40 characters name length limitation for backing images. If it exceeds the limitation, [webhook will automatically change the backing image name](https://github.com/longhorn/longhorn-manager/blob/432b6d40f02a6fdffd8389df9d779016775f1d3a/webhook/resources/backingimage/mutator.go#L50).

Considering Harvester's naming convention for `virtualimage`, `backingimage` and `storageclass`:
[...]

PR #4561, opened 2023-09-22
Rectify backing image name

WebberHuang1118:
**Problem:**

Longhorn has a 40 characters name length limitation for backing images. If it exceeds the limitation, [webhook will automatically change the backing image name](https://github.com/longhorn/longhorn-manager/blob/432b6d40f02a6fdffd8389df9d779016775f1d3a/webhook/resources/backingimage/mutator.go#L50).

And this violates Harvester's naming convention for `virtualimage`, `backingimage` and `storageclass`:
- `backingimage name` =` <virutalimage namsespace>`-`<virutalimage name>`
- `storageclass name` = `longhorn-<virutalimage name>`

**Solution:**
Rectify the backing image name to align Longhorn's webhook

**Related Issue:**
#4520 

**Test plan on Master Environment:**
- Prerequisites
  1. Applying the following yaml to create virtualimage:
      ```
      kind: VirtualMachineImage
      metadata:
        name: ubuntu-bionic-123456789012345678901234567890
        namespace: default
      spec:
        checksum: ""
        displayName: ubuntu-bionic
        pvcName: ""
        pvcNamespace: ""
        retry: 3
        sourceType: download
        storageClassParameters:
          migratable: "true"
          numberOfReplicas: "3"
          staleReplicaTimeout: "30"
        url: https://cloud-images.ubuntu.com/bionic/20230607/bionic-server-cloudimg-amd64.img
      ```
  2. Creating VM from this virtualimage
  
- Results without this patch:
  1. backingimage `default-ubuntu-bionic-123456789-xxxxxxxx` created
  2. storageclass `longhorn-ubuntu-bionic-123456789012345678901234567890`
      - `paramters.backingImage` points to `default-ubuntu-bionic-123456789012345678901234567890` (**mismatch**)
  3. VM creation appears volume provision issue `missing backing image default-ubuntu-bionic-123456789012345678901234567890 unable to create volume` 
  4. Backing image requires manual deletion to cleanup

- Results with this patch:
  1. backingimage `default-ubuntu-bionic-123456789-xxxxxxxx` created
  2. storageclass `longhorn-ubuntu-bionic-123456789012345678901234567890`
      -   `paramters.backingImage` points to `default-ubuntu-bionic-123456789-xxxxxxxx` (**match**)
  3. VM creation success

**Test plan for Backward Compatibility:**
- Prerequisites
  1. Build Harvester Cluster with v1.1.2
  2. Applying the following yaml to create virtualimage:
      ```
      kind: VirtualMachineImage
      metadata:
        name: ubuntu-bionic-123456789012345678901234567890
        namespace: default
      spec:
        checksum: ""
        displayName: ubuntu-bionic
        pvcName: ""
        pvcNamespace: ""
        retry: 3
        sourceType: download
        storageClassParameters:
          migratable: "true"
          numberOfReplicas: "3"
          staleReplicaTimeout: "30"
        url: https://cloud-images.ubuntu.com/bionic/20230607/bionic-server-cloudimg-amd64.img
      ```
  3. There should be a backingimage `default-ubuntu-bionic-123456789012345678901234567890` created
  4. storageclass `longhorn-ubuntu-bionic-123456789012345678901234567890`
      - `paramters.backingImage` points to `default-ubuntu-bionic-123456789012345678901234567890` (**match**)
   5. Build Harvester from this ISO, and upgrade to it
- Steps: 
  1.  VM creation with this image success
  2. Deleting the VM, related volume, and image success

### DIFF
--- a/pkg/util/image.go
+++ b/pkg/util/image.go
@@ -3,27 +3,69 @@ package util
 import (
 	"fmt"
 
+	lhdatastore "github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
+	lhutil "github.com/longhorn/longhorn-manager/util"
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctllhv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
 )
 
-func GetBackingImageName(image *harvesterv1.VirtualMachineImage) string {
+func backingImageLegacyName(image *harvesterv1.VirtualMachineImage) string {
 	return fmt.Sprintf("%s-%s", image.Namespace, image.Name)
+}
+
+func GetBackingImage(backingImageCache ctllhv1.BackingImageCache, image *harvesterv1.VirtualMachineImage) (*v1beta2.BackingImage, error) {
+	bi, err := backingImageCache.Get(LonghornSystemNamespaceName, backingImageLegacyName(image))
+	if err == nil {
+		return bi, nil
+	}
+
+	if !errors.IsNotFound(err) {
+		return nil, err
+	}
+
+	rectifyName := lhutil.AutoCorrectName(backingImageLegacyName(image), lhdatastore.NameMaximumLength)
+	return backingImageCache.Get(LonghornSystemNamespaceName, rectifyName)
+}
+
+func GetBackingImageName(backingImageCache ctllhv1.BackingImageCache, image *harvesterv1.VirtualMachineImage) (string, error) {
+	bi, err := GetBackingImage(backingImageCache, image)
+	if err == nil {
+		return bi.Name, nil
+	}
+
+	if !errors.IsNotFound(err) {
+		return "", err
+	}
+
+	return lhutil.AutoCorrectName(backingImageLegacyName(image), lhdatastore.NameMaximumLength), nil
+}
+
+func GetBackingImageDataSourceName(backingImageCache ctllhv1.BackingImageCache, image *harvesterv1.VirtualMachineImage) (string, error) {
+	//In LH design, backingimagedatasource name is identical with backingimage
+	return GetBackingImageName(backingImageCache, image)
 }
 
 func GetImageStorageClassName(imageName string) string {
 	return fmt.Sprintf("longhorn-%s", imageName)
 }
 
-func GetImageStorageClassParameters(image *harvesterv1.VirtualMachineImage) map[string]string {
+func GetImageStorageClassParameters(backingImageCache ctllhv1.BackingImageCache, image *harvesterv1.VirtualMachineImage) (map[string]string, error) {
+	biName, err := GetBackingImageName(backingImageCache, image)
+	if err != nil {
+		return nil, err
+	}
+
 	params := map[string]string{
-		LonghornOptionBackingImageName: GetBackingImageName(image),
+		LonghornOptionBackingImageName: biName,
 	}
 	for k, v := range image.Spec.StorageClassParameters {
 		params[k] = v
 	}
-	return params
+	return params, nil
 }
 
 func GetImageDefaultStorageClassParameters() map[string]string {
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -47,6 +47,7 @@ type Handler struct {
 	ImageCache                  v1beta1.VirtualMachineImageCache
 	BackingImageDataSources     ctllhv1.BackingImageDataSourceClient
 	BackingImageDataSourceCache ctllhv1.BackingImageDataSourceCache
+	BackingImageCache           ctllhv1.BackingImageCache
 }
 
 func (h Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
@@ -100,17 +101,21 @@ func (h Handler) downloadImage(rw http.ResponseWriter, req *http.Request) error 
 		return fmt.Errorf("failed to get VMImage with name(%s), ns(%s), error: %w", name, namespace, err)
 	}
 
+	biName, err := util.GetBackingImageName(h.BackingImageCache, vmImage)
+	if err != nil {
+		return fmt.Errorf("failed to get backing image name for VMImage %s/%s, error: %w", namespace, name, err)
+	}
+
 	targetFileName := vmImage.Spec.DisplayName
-	bkimgName := fmt.Sprintf("%s-%s", namespace, name)
-	downloadURL := fmt.Sprintf("%s/backingimages/%s/download", util.LonghornDefaultManagerURL, bkimgName)
+	downloadURL := fmt.Sprintf("%s/backingimages/%s/download", util.LonghornDefaultManagerURL, biName)
 	downloadReq, err := http.NewRequestWithContext(req.Context(), http.MethodGet, downloadURL, nil)
 	if err != nil {
-		return fmt.Errorf("failed to create the download request with backing Image(%s): %w", bkimgName, err)
+		return fmt.Errorf("failed to create the download request with backing Image(%s): %w", biName, err)
 	}
 
 	downloadResp, err := h.httpClient.Do(downloadReq)
 	if err != nil {
-		return fmt.Errorf("failed to send the download request with backing Image(%s): %w", bkimgName, err)
+		return fmt.Errorf("failed to send the download request with backing Image(%s): %w", biName, err)
 	}
 	defer downloadResp.Body.Close()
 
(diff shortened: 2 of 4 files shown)


FrankYang0529:
LH added `util.AutoCorrectName` in mutator since v1.3.0 [1]. However, Harvester v1.0 used LH v1.2.4 [2]. After this patch, not sure whether images created in Harvester v1.0 can work?

[1] https://github.com/longhorn/longhorn-manager/commit/5fc4895b0e53d8af1d9d85f25b48e6901b42e283#diff-a787b956487f2216a71535428b010bd048007a630f283169136ec20af7bfe68c
[2] https://github.com/harvester/harvester/tree/v1.0/deploy/charts/harvester/charts

w13915984028:
The 40 characters length limitation of name converting looks to be a breaking change, did LH add the original-name to certain annotaions for APP level to refer ?

When APP creates one backingimage by given name, it will not be possbile to retrieve it by the name, it looks brings trouble. If LH in validator refues the creating saying name is too long, that will let user know in early stage.

#4520 #4589 seem to be both related to it.

@bk201 @markhillgit @innobead 
It looks we need to check all Harvester related parts, to make sure the affecting.

w13915984028:
@FrankYang0529 

per your description:
> LH added util.AutoCorrectName in mutator since v1.3.0 [1]. However, Harvester v1.0 used LH v1.2.4 [2]. After this patch, not sure whether images created in Harvester v1.0 can work?

I did more tests to find it is slightly different:

Seems the bug in here https://github.com/harvester/harvester/issues/4589#issuecomment-1750645974 is fixed (Harvester v1.1.2 LH v1.3.2), but another bug is imported in Harvester v1.2.0, LH v1.4.3 #4589 

The LH `v1.3.2` used in Harvester v1.1.2, did not have this mutator to truncate the naming of backingimage, but in Harvester v1.2.0  LH `v1.4.3`, the backingimage name is really truncated.

FrankYang0529:
@WebberHuang1118, we may also need to take care storage class name in this PR. We only see backing image error is because we create storage class after backing image.

https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/controller/master/image/vm_image_controller.go#L145-L150

FrankYang0529:
> @WebberHuang1118, we may also need to take care storage class name in this PR. We only see backing image error is because we create storage class after backing image.
> 
> https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/controller/master/image/vm_image_controller.go#L145-L150

I think we can probably create another issue to track storage class name, because it's related to some parts.

In terraform, we use this builder to construct storage class name:
https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/builder/disk.go#L46-L51

In GUI, we use storage class name in annotation `harvesterhci.io/volumeClaimTemplates`.

WebberHuang1118:
> > @WebberHuang1118, we may also need to take care storage class name in this PR. We only see backing image error is because we create storage class after backing image.
> > https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/controller/master/image/vm_image_controller.go#L145-L150
> 
> I think we can probably create another issue to track storage class name, because it's related to some parts.
> 
> In terraform, we use this builder to construct storage class name:
> 
> https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/builder/disk.go#L46-L51
> 
> In GUI, we use storage class name in annotation `harvesterhci.io/volumeClaimTemplates`.

@FrankYang0529, do you mean terraform has a different naming convention for StorageClass? Is not based on [vm image controller's behavior](https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/controller/master/image/vm_image_controller.go#L145-L150)?

FrankYang0529:
> > > @WebberHuang1118, we may also need to take care storage class name in this PR. We only see backing image error is because we create storage class after backing image.
> > > https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/controller/master/image/vm_image_controller.go#L145-L150
> > 
> > 
> > I think we can probably create another issue to track storage class name, because it's related to some parts.
> > In terraform, we use this builder to construct storage class name:
> > https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/builder/disk.go#L46-L51
> > In GUI, we use storage class name in annotation `harvesterhci.io/volumeClaimTemplates`.
> 
> @FrankYang0529, do you mean terraform has a different naming convention for StorageClass? Is not based on [vm image controller's behavior](https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/controller/master/image/vm_image_controller.go#L145-L150)?

No, it just uses `BuildImageStorageClassName` to retrieve storage class name from vmimage name. I think we can have some refactor on that part.

https://github.com/harvester/terraform-provider-harvester/blob/f68b5a360a8c020dfdbf604c67f996bd444164b3/internal/provider/virtualmachine/resource_virtualmachine_constructor.go#L222

WebberHuang1118:
@FrankYang0529,
If my understanding is correct, even Longhorn doesn't have auto-correct on storage class names but terraform's expectations for storage class names are different from GUI's. As you mentioned, refactoring this part is a good idea.

FrankYang0529:
> @FrankYang0529, If my understanding is correct, even Longhorn doesn't have auto-correct on storage class names but terraform's expectations for storage class names are different from GUI's. As you mentioned, refactoring this part is a good idea.

No. Both terraform and GUI need to construct storage class name for PVC. They uses different utility functions to retrieve storage class name from vmimage name. It's better for them to get the vmimage object from kube-apiserver and use vmimage.status.storageClassName, so we don't need to maintain so many utility functions.

pkg/util/image.go
```diff
@@ -3,27 +3,68 @@ package util
...
+	bi, err := backingImageCache.Get(LonghornSystemNamespaceName, legacyName)
+	if err == nil {
+		return bi, nil
+	}
+
+	if !errors.IsNotFound(err) {
+		return nil, err
+	}
+
+	rectifyName := lhutil.AutoCorrectName(legacyName, lhdatastore.NameMaximumLength)
```
FrankYang0529: We can reuse code in `GetBackingImageName`. We will call `backingImageCache.Get` again for legacy name, but it's local cache, so the extra cost can be ignored.

WebberHuang1118: @FrankYang0529 
How about resue `GetBackingImage() `in `GetBackingImageName()` ? Then we don't have to call backingImageCache.Get() again like this:
```
func GetBackingImageName(backingImageCache ctllhv1.BackingImageCache, image *harvesterv1.VirtualMachineImage) (string, error) {
	bi, err := GetBackingImage(backingImageCache, image)
	if err == nil {
		return bi.Name, nil
	}

	if !errors.IsNotFound(err) {
		return "", err
	}

	legacyName := fmt.Sprintf("%s-%s", image.Namespace, image.Name)
	return lhutil.AutoCorrectName(legacyName, lhdatastore.NameMaximumLength), nil
}
```

FrankYang0529: I'm okay with it. However, do we want to create a local utility function for legacy name? We repeat it in both `GetBackingImage` and `GetBackingImageName`.

WebberHuang1118: Updated per suggestion, thanks.

FrankYang0529 (approved):
LGTM. Thanks you.

Test steps:

1. Install harvester v1.1.2.
2. Create the following image.
```yaml
kind: VirtualMachineImage
metadata:
  name: ubuntu-bionic-123456789012345678901234567890
  namespace: default
spec:
  checksum: ""
  displayName: ubuntu-bionic
  pvcName: ""
  pvcNamespace: ""
  retry: 3
  sourceType: download
  storageClassParameters:
    migratable: "true"
    numberOfReplicas: "3"
    staleReplicaTimeout: "30"
  url: https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-amd64.img
```
3. Use `ubuntu-bionic` to create a new VM. Check whether VM can work.
4. Build a new image from this PR and replace `harvester/harvester` deployment.
5. Check `ubuntu-bionic` is healthy.
6. Create a new image.
```yaml
kind: VirtualMachineImage
metadata:
  name: ubuntu-bionic-123456789012345678901234567890-12345
  namespace: default
spec:
  checksum: ""
  displayName: ubuntu-bionic-new
  pvcName: ""
  pvcNamespace: ""
  retry: 3
  sourceType: download
  storageClassParameters:
    migratable: "true"
    numberOfReplicas: "3"
    staleReplicaTimeout: "30"
  url: https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-amd64.img
```
7. Check the backing image name is `default-ubuntu-bionic-123456789-5048d59d` for vmimage `ubuntu-bionic-new`.
8. Use `ubuntu-bionic-new` to create a new VM without error.

connorkuehl (approved):

pkg/util/image.go
```diff
@@ -3,27 +3,70 @@ package util
...
 
+	"github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctllhv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
+
+	lhdatastore "github.com/longhorn/longhorn-manager/datastore"
+	lhutil "github.com/longhorn/longhorn-manager/util"
```
bk201: We can organize imports:
```
	"fmt"

	lhdatastore "github.com/longhorn/longhorn-manager/datastore"
	"github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
	longhorntypes "github.com/longhorn/longhorn-manager/types"
	lhutil "github.com/longhorn/longhorn-manager/util"
	"k8s.io/apimachinery/pkg/api/errors"

	ctllhv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
```

WebberHuang1118: Update per suggestion, thanks.

bk201 (approved):

WebberHuang1118:
@Mergifyio backport v1.2

Issue #4600, opened 2023-10-11
[BUG] Avoid to delete LH settings


FrankYang0529:
**Describe the bug**
LH prevents users from deleting LH settings. Harvester tries to delete the `backup-target-credential-secret` setting when users update the `backup-target` setting. We have to refactor this part.

LH Ref: https://github.com/longhorn/longhorn/issues/4984
Harv Ref: https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/controller/master/backup/backup_target.go#L110-L124

**To Reproduce**
1. Run Harvester with latest LH version.
2. Setup the backup target and we will get an error.

**Expected behavior**
1. Users can setup the backup target.
2. All old test cases pass.

lanfon72:
Verified NFS/S3 backup-target could be configured and spawn backups successfully.

(`test_restore_replace_with_delete_vols` is known issue in #4640)
![image](https://github.com/harvester/harvester/assets/5169694/ec26cf68-6a20-4dcd-92f6-cfad53dbafbe)


Test Information
----
* Environment: **Baremetal DL360G9 5 nodes**
* Harvester Version: **v1.3.0-dev-20231222**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Create VM `vm1`
3. Setup NFS/S3 backup-target
4. Backup `vm1` and restore `vm1`
5. VM should be backup and restore successfully
6. Backup-target should able to configured and switched successfully

PR #4605, opened 2023-10-12
fix(backup-target): don't remove LH settings

FrankYang0529:
**Problem:**
LH prevents users from deleting LH settings. Harvester tries to delete the `backup-target-credential-secret` setting when users update the `backup-target` setting. We have to refactor this part.

LH Ref: https://github.com/longhorn/longhorn/issues/4984
Harv Ref: https://github.com/harvester/harvester/blob/6da66dc6f7bdebf43bd4932eb5128642ec397e51/pkg/controller/master/backup/backup_target.go#L110-L124

**Solution:**
Remove content in LH setting instead of delete the setting.

**Related Issue:**
https://github.com/harvester/harvester/issues/4600

**Test plan:**
1. Create a new Harvester cluster without any error.
2. Set the `backup-target` setting to NFS. The `backup-target-credential-secret` LH setting is existent and content is empty.
3. Create a VM backup / restore without error.
4. Set the `backup-target` setting to default. The `backup-target-credential-secret` LH setting is existent and content is empty.
5. Set the `backup-target` setting to S3. The `harvester-backup-target-secret` secret is created.
6. Create a VM bakcup / restore without error.
7. Set the `backup-target` setting to default. The `backup-target-credential-secret` LH setting and `harvester-backup-target-secret` are existent and content is empty.

### DIFF
--- a/pkg/controller/master/backup/backup_target.go
+++ b/pkg/controller/master/backup/backup_target.go
@@ -106,8 +106,8 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
 			return h.setConfiguredCondition(setting, "", err)
 		}
 
-		// delete the may existing previous secret of S3
-		if err = h.deleteBackupTargetSecret(target); err != nil {
+		// reset previous secret of S3
+		if err = h.resetBackupTargetSecret(); err != nil {
 			return h.setConfiguredCondition(setting, "", err)
 		}
 
@@ -118,8 +118,8 @@ func (h *TargetHandler) OnBackupTargetChange(key string, setting *harvesterv1.Se
 				return h.setConfiguredCondition(setting, "", err)
 			}
 
-			// delete the may existing previous secret of S3
-			if err = h.deleteBackupTargetSecret(target); err != nil {
+			// reset previous secret of S3
+			if err = h.resetBackupTargetSecret(); err != nil {
 				return h.setConfiguredCondition(setting, "", err)
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/master/backup/backup_target.go
```diff
@@ -250,14 +250,30 @@ func (h *TargetHandler) updateBackupTargetSecret(target *settings.BackupTarget)
...
-func (h *TargetHandler) deleteBackupTargetSecret(target *settings.BackupTarget) error {
-	if err := h.secrets.Delete(util.LonghornSystemNamespaceName, util.BackupTargetSecretName, nil); err != nil && !apierrors.IsNotFound(err) {
+func (h *TargetHandler) resetBackupTargetSecret() error {
+	secret, err := h.secretCache.Get(util.LonghornSystemNamespaceName, util.BackupTargetSecretName)
+	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
+	if secret != nil {
+		secretCpy := secret.DeepCopy()
+		secretCpy.Data = map[string][]byte{}
```
w13915984028, line 260: please double check, should we check the .Data len == 0 to avoid unnecessary update ? following update is similar

thanks.

w13915984028 (approved):
LGTM, thanks.

pkg/controller/master/backup/backup_target.go
FrankYang0529: Yes, good catch. Added checking whether `secret.Data` and `setting.Value` is empty before udpate.

bk201 (approved):
LGTM!

FrankYang0529:
@Mergify backport v1.2

Issue #4606, opened 2023-10-12
[ENHANCEMENT] Bump LH to v1.5.x in Harv v1.3.0


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
LH v.14.x will be EOL in 2024/03/13. Also, LH only allows to upgrade 1 major or minor version after v1.5.x. If we don't bump LH to v1.5.x in Harv v1.2.x. We can't include LH v1.6 new features in Harv v1.3. For example:

* [[FEATURE] Longhorn snapshot space management](https://github.com/longhorn/longhorn/issues/6563#top)
* [[FEATURE] Restore BackingImage for BackupVolume in a new cluster](https://github.com/longhorn/longhorn/issues/4165#top)

lanfon72:
close as verified.

![image](https://github.com/harvester/harvester/assets/5169694/0788ad02-7037-464b-b54a-df4d78d6feb7)

Test Information
----
* Environment: **qemu/KVM single node**
* Harvester Version: **v1.3-f132d7fa-head**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Access embedded Longhorn UI

Issue #2237, opened 2022-05-09
[FEATURE] Restore vm backup to other harvester cluster


fanlix:
**Is your feature request related to a problem? Please describe.**

my harvester cluster crash down. 
so I reinstall a new harvester, 
import all vm backups (nfs), restore them, but fail.

the direct reason is `ProvisioningFailed: Storageclass.storage.k8s.io "longhorn-image-xxxxx" not found`

**Describe the solution you'd like**

* set old nfs backup dir to new harvester (or copy one backup file)
* restore this backup in harvester ui
* need to edit some config different from old harvester env: 
  * namespace/cpu/ram/sshkey/network  (already editable in current harvester version)
  * base-image/image-storageclass...

**Describe alternatives you've considered**

* export volume from old-harvester to image
* backup this image to nfs  (or download to local file)
* restore this image in new-harvester, create new vm

any idea is welcome.
we need a persistence backup method, not bound to a running cluster.

yasker:
There are a few complications regarding restoring a VM into a new Harvester cluster. We will investigate to understand the effort first.

dseralathan:
Backup target is set to a NFS on both clusters . 
I have created a "VMBackup" from cluster 1 and power off VM and try to restore to the cluster 2. 
On cluster 2 the there is no backup available to restore . any idea how we can get the backup sync from cluster1 to 2

albertkohl-monotek:
@dseralathan this is a known issue w/ v1.1.0, i believe it should have been fixed in v1.1.1 (i'm not running it yet, so i cant confirm that. 

what version of harvester are you running?

dseralathan:
Thanks Albert, I am running with v1.1.0 ..

albertkohl-monotek:
@dseralathan I’m about to test-upgrade a couple nodes tonight/tomorrow. I’ll let you know if the upgrade to v1.1.1 fixes it 😊

dseralathan:
I have upgraded my cluster to v1.1.1  and it works super cool . Thanks Albert . good  success in cross cluster  VM re-store  !

albertkohl-monotek:
[...]

Issue #1850, opened 2022-01-20
[FEATURE] Replication of VirtualMachines across Harvester Clusters - DR


ibrokethecloud:
**Is your feature request related to a problem? Please describe.**

Harvester should support replication of VMs across Harvester Clusters. 
This can be useful when looking to migrate VM's from one cluster to another, and also for DR scenarios.

**Describe the solution you'd like**

Longhorn already supports ability to backup and replication volumes across K8S clusters. Harvester needs to leverage the same to replicate the volumes and associated VirtualMachine metadata to another target cluster.

**Describe alternatives you've considered**

**Additional context**

asettle:
Related to Harvester support for DR Volumes

abonillabeeche:
Don't let this one die!!

innobead:
#10419

Vicente-Cheng:
We need to discuss #10419 and have the specific scope for the DR feature. 

cc @harvester/storage, @harvester/tpm, @ihcsim, @innobead

forbesguthrie:
This shouldn't be blocked by a broader DR strategy.  We need to be able to deliver cross cluster/namespace replication (and therefore migration) as an important operational workflow.  DR is a very specific use-case which may utilize this, but if we tie everything required for a sane DR solution, then this will continue to delayed.
We should also consider splitting this into storage agnostic migration, and Longhorn-backed migration/replication.  The first is more broadly useful, but the second could efficiencies for those VMs.  I presume there will be different mechanisms used, but the customer workflow should be the same.

innobead:
> This shouldn't be blocked by a broader DR strategy. We need to be able to deliver cross cluster/namespace replication (and therefore migration) as an important operational workflow. 

Yes, in fact, live migration is not identical to DR: one moves running VMs with their volumes, whereas the other is replication for an active backup.

PR #4614, opened 2023-10-13
Backup and restore VMImage

FrankYang0529:
**Problem:**
We backup volumes without images. If we set the backup target to another cluster, we can't really restore a VM.

**Solution:**
Integrate LH [[FEATURE] Restore BackingImage for BackupVolume in a new cluster](https://github.com/longhorn/longhorn/issues/4165#top) with Harvester. At Harvester side, we need to backup vmimage metadata, so we have information to restore vmimages in a new cluster.

**Related Issue:**
https://github.com/harvester/harvester/issues/2237

**Wait for following issues:**
- [x] [[BUG] Avoid to delete LH settings](https://github.com/harvester/harvester/issues/4600#top)
- [x] [[ENHANCEMENT] Bump LH to v1.5.x in Harv v1.2.x](https://github.com/harvester/harvester/issues/4606#top)
- [x] [[FEATURE] Restore BackingImage for BackupVolume in a new cluster](https://github.com/longhorn/longhorn/issues/4165#top)

**Test plan:**

### Case 1: Backup VMImage
1. Create a VMImage.
2. Setup backup target.
3. Create a VM.
4. Create a VMBackup for the VM.
5. Check there is metadata for the VMImage in the backup target with path `/opt/backupstore/harvester/vmimages/<vmimage-namespace>/<vmimage-name>`.

### Case 2: Restore VMImage
1. Create a new cluster or cleanup resources in current cluster with following order:
    a. backup-target
    b. vmbackup
    c. vm
    d. vmimage
2. Setup backup target.
3. Check whether VMImage and VMBackup are restored.
4. Restore VMBackup to a new VM. The new VM should be started without error and have all content.


pkg/controller/master/backup/backup_backing_image.go
```diff
@@ -0,0 +1,148 @@
...
+		vmImageCache:            vmImages.Cache(),
+		backingImageCache:       backingImages.Cache(),
+		backupBackingImageCache: backupBackingImages.Cache(),
+	}
+
+	backupBackingImages.OnChange(ctx, backupBackingImageControllerName, backupBackingImageController.OnBackupBackingImageChange)
+	return nil
+}
+
+// OnBackupTargetChange resync vm metadata files when backup target change
```
brandboat: The comment here seems to be inconsistent with the method itself.

pkg/controller/master/backup/backup_backing_image.go
```diff
@@ -0,0 +1,148 @@
...
+	secretCache             ctlcorev1.SecretCache
+	longhornSettingCache    ctllonghornv1.SettingCache
+	settingCache            ctlharvesterv1.SettingCache
+	vmImages                ctlharvesterv1.VirtualMachineImageClient
+	vmImageCache            ctlharvesterv1.VirtualMachineImageCache
+	backingImageCache       ctllonghornv1.BackingImageCache
+	backupBackingImageCache ctllonghornv1.BackupBackingImageCache
+}
+
+// RegisterBackupMetadata register the setting controller and resync vm backup metadata when backup target change
```
brandboat: The comment here seems to be inconsistent with the method itself.

pkg/controller/master/backup/backup_backing_image.go
```diff
@@ -0,0 +1,148 @@
...
+		vmImageMetadata.Namespace = metav1.NamespaceDefault
+	}
+
+	data, err := json.Marshal(vmImageMetadata)
+	if err != nil {
+		return nil, err
+	}
+
+	shouldUpload := true
+	destURL := getVMImageMetadataFilePath(vmImage.Namespace, vmImage.Name)
```
brandboat: Looks like the `getVMImageMetadataFilePath` output is not in  URL format. Maybe we can rename this variable to `filePath` ?

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -101,42 +116,112 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
 		return nil, nil
 	}
 
+	logrus.WithFields(logrus.Fields{
+		"target.type":     target.Type,
+		"target.endpoint": target.Endpoint,
+	}).Info("start syncing vm image metadata...")
```
brandboat: nit: Could we extract the code snippet to a variable ?
```go
logrus.Fields{
    "target.type":     target.Type,
    "target.endpoint": target.Endpoint,
}
```

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -101,42 +116,112 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
...
+			"target.endpoint": target.Endpoint,
+		}).Errorf("can't sync vm image metadata")
+		h.settings.EnqueueAfter(setting.Name, 5*time.Second)
+		return nil, nil
+	}
+
+	logrus.WithFields(logrus.Fields{
+		"target.type":     target.Type,
+		"target.endpoint": target.Endpoint,
+	}).Info("start syncing vm backup metadata...")
```
brandboat: ditto

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -101,42 +116,112 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
...
 	if err = h.syncVMBackup(target); err != nil {
-		logrus.Errorf("can't sync vm backup metadata, target:%s:%s, err: %v", target.Type, target.Endpoint, err)
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"target.type":     target.Type,
+			"target.endpoint": target.Endpoint,
```
brandboat: ditto

brandboat (approved):
LGTM % minor comments, many thanks :smiley:

pkg/controller/master/backup/backup_backing_image.go
```diff
@@ -0,0 +1,148 @@
...
+			shouldUpload = false
+		}
+	}
+
+	if shouldUpload {
+		logrus.Debugf("upload vm image metadata %s/%s to backup target %s", vmImage.Namespace, vmImage.Name, target.Type)
+		if err := bsDriver.Write(destURL, bytes.NewReader(data)); err != nil {
+			return nil, err
+		}
+	}
```
WebberHuang1118: Since we will upload the image metadata and the VM backup metadata, should we update both uploading statuses in the vmbackup CR to describe if the uploadings are successful or not? I think it also helps for the sync-up case from the backup target.

WebberHuang1118 (commented):
Generally LGTM, just please reorganize the commits, `util.GetBackupStoreDriver` is used [in the second commit ](https://github.com/harvester/harvester/pull/4614/commits/ccae2252a4a4fbdcdc52693fe48c98a28c975a76), however, it's defined in the [third commit](https://github.com/harvester/harvester/pull/4614/commits/007e470d36a5113d7b279b457cd0da00d8ff1bf8)

thanks

FrankYang0529:
Hi @brandboat and @WebberHuang1118, I have addressed all comments. Could you review again when you have time? Thank you.

pkg/controller/master/backup/backup_backing_image.go
FrankYang0529: I add condition to different CRDs:

* If VMBackup metadata is uploaded, set `MetadataReady` condition to VMBackup.
* If VMImage metadata is uploaded, set `MetadataReady` condition to VMImage.

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -89,12 +89,15 @@ func (v *virtualMachineImageValidator) CheckImageDisplayNameAndURL(newImage *v1b
...
+	if newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeDownload || newImage.Spec.SourceType == v1beta1.VirtualMachineImageSourceTypeRestore {
+		shouldHaveURL = true
 	}
 
-	if newImage.Spec.SourceType != v1beta1.VirtualMachineImageSourceTypeDownload && newImage.Spec.URL != "" {
-		return werror.NewInvalidError(`url should be empty when image source type is not "download"`, "spec.url")
+	if shouldHaveURL && newImage.Spec.URL == "" {
+		return werror.NewInvalidError(fmt.Sprintf(`url is required when image source type is "%s"`, newImage.Spec.SourceType), "spec.url")
+	} else if !shouldHaveURL && newImage.Spec.URL != "" {
+		return werror.NewInvalidError(`url should be empty when image source type is not "download" or "restore"`, "spec.url")
```
brandboat: nit, ignore this if you think not necessary.
```suggestion
		return werror.NewInvalidError(fmt.Sprintf(`url should be empty when image source type is "%s"`, newImage.Spec.SourceType), "spec.url")
```

brandboat (commented):
Thanks for the patch, I'll continue review tomorrow.

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -101,42 +116,104 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
...
-		os.Setenv(AWSAccessKey, string(secret.Data[AWSAccessKey]))
-		os.Setenv(AWSSecretKey, string(secret.Data[AWSSecretKey]))
-		os.Setenv(AWSEndpoints, string(secret.Data[AWSEndpoints]))
-		os.Setenv(AWSCERT, string(secret.Data[AWSCERT]))
+		for _, fileName := range fileNames {
+			imageMetadata, err := loadVMImageMetadataInBackupTarget(filepath.Join(vmImageMetadataFolderPath, namespaceFolder, fileName), bsDriver)
+			if err != nil {
+				return err
+			}
+			if imageMetadata.Namespace == "" {
```
brandboat, line 161: Will this happen ? In `pkg/controller/master/backup/backup_backing_image.go` L122, it force vmImageMetadata namespace to `default` if namespace is empty. 
```
	if vmImageMetadata.Namespace == "" {
		vmImageMetadata.Namespace = metav1.NamespaceDefault
	}
```

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -32,10 +31,21 @@ import (
 )
 
 const (
-	metadataFolderPath           = "harvester/vmbackups/"
+	vmBackupMetadataFolderPath   = "harvester/vmbackups/"
+	vmImageMetadataFolderPath    = "harvester/vmimages/"
 	backupMetadataControllerName = "harvester-backup-metadata-controller"
 )
 
+type VirtualMachineImageMetadata struct {
```
brandboat, line 39: Looks like we also need to refine the document, in https://docs.harvesterhci.io/v1.4/vm/backup-restore/#restore-a-new-vm-on-another-harvester-cluster, we have a prerequisite, 
> PREREQUISITES
You must manually configure the virtual machine images with the same name on the new cluster first, otherwise the virtual machines will be failed to recover.

Now users no longer have to manually upload all the VM images to the new cluster. 🎉

brandboat (approved):
LGTM! I've tested this PR and it works well. I just have a few questions left.

pkg/controller/master/backup/backup.go
```diff
@@ -884,6 +860,17 @@ func (h *Handler) uploadVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
 		}
 	}
 
+	vmBackupCopy := vmBackup.DeepCopy()
+	updateBackupCondition(vmBackupCopy, harvesterv1.Condition{
+		Type:               harvesterv1.BackupConditionMetadataReady,
+		Status:             corev1.ConditionTrue,
+		LastTransitionTime: currentTime().Format(time.RFC3339),
+	})
```
WebberHuang1118, line 868: Please reorder the commit (if you plan to squash all, just ignore this), 
this change is introduced in the commit https://github.com/harvester/harvester/pull/4614/commits/00b8b9d5f64c2de82f23a4efc8b6082cf673c4b5, however `BackupConditionMetadataReady` is declared in the later commit https://github.com/harvester/harvester/pull/4614/commits/ff4cfba5a8d60e52d9dbc24ca9af246ed6a0cb1c.

pkg/controller/master/backup/backup_backing_image.go
```diff
@@ -0,0 +1,156 @@
...
+
+	vmImageCopy := vmImage.DeepCopy()
+	harvesterv1.MetadataReady.True(vmImageCopy)
+	if !reflect.DeepEqual(vmImage, vmImageCopy) {
+		_, err = h.vmImages.Update(vmImageCopy)
+		return nil, err
+	}
```
WebberHuang1118, line 153: Should we also have a condition in VMBackup about the related images' upload status? 
IMHO, it would be easier to track.

WebberHuang1118 (commented):
LGTM, just a nit and a possible enhancement, thanks.

pkg/controller/master/backup/backup_backing_image.go
FrankYang0529: Hi @WebberHuang1118, thanks for the suggestion. I would prefer to put the condition in VMImage, because different Volumes may use same VMImage. In this situation, the VMImage backup metadata only need to be uploaded one time, so the condition on VMImage will be single source of truth.

pkg/controller/master/backup/backup.go
FrankYang0529: Fixed it. Thank you.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: In general, this should not happen, but I would like to do best effort to fill with default value.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: Yes, create a PR for it. Thank you.

https://github.com/harvester/docs/pull/613

pkg/controller/master/backup/backup_backing_image.go
WebberHuang1118: Thanks, like you said, let's keep the upload status in the image.

WebberHuang1118 (approved):
LGTM, thanks.

tserong (approved):
I feel like this is going to help with https://github.com/harvester/harvester/issues/1850 :-)

The code LGTM, and includes some nice refactoring. That said, I'm not yet 100% familiar with all the details of VM backup and restore. In such review situations I try to work backwards from where the feature intersects the existing codebase. For backup, the major work seems to ultimately come from a call to `syncVMImage()` in `OnBackupTargetChange()`. For restore, it looks like it comes down to  this rather neat thing in vm_image_controller.go:

```
case harvesterv1.VirtualMachineImageSourceTypeRestore:
  bi.Spec.SourceParameters[lhv1beta2.DataSourceTypeRestoreParameterBackupURL] = image.Spec.URL
```

FrankYang0529:
Hi @tserong, thanks for the review. The VMImage backup is relied on LH BackupBackingImage. If there is backup for a Volume which is from a BackingImage, LH will automatically upload it to backup target. In harvester, we only save VMImage metadata file to backup target.

When changing backup target, we check VMImage metadata file list on backup target and restore as VMImage if it's not existent.

bk201 (approved):
LGTM, I can restore a backup to a new cluster and create VMs.

bk201:
@mergifyio backport v1.4

Issue #4633, opened 2023-10-19
[BUG] VM got IO error after host restart


FrankYang0529:
**Describe the bug**
VM can't restart after a host accidentally restarts. VMI changes frequently between ready and unready and it has `VMI was paused, IO error` error message. At first, related Volume's robustness is fault and engine is stopped. After around 30 minutes, the volume is attached and engine is running, but VMI still can't get ready. We have to check why VM can't be ready after the LH engine is back.

```
> kubectl get vmi -w
NAME                          AGE   PHASE     IP             NODENAME           READY
rke2-3-pool1-47f7bd68-r99dx   44m   Running                  harvester-node-2   False
rke2-3-pool1-47f7bd68-r99dx   44m   Running   192.168.3.56   harvester-node-2   False
rke2-3-pool1-47f7bd68-r99dx   44m   Running                  harvester-node-2   False
rke2-3-pool1-47f7bd68-r99dx   44m   Running                  harvester-node-2   True
rke2-3-pool1-47f7bd68-r99dx   44m   Running                  harvester-node-2   False
rke2-3-pool1-47f7bd68-r99dx   44m   Running   192.168.3.56   harvester-node-2   False
rke2-3-pool1-47f7bd68-r99dx   44m   Running   192.168.3.56   harvester-node-2   True
rke2-3-pool1-47f7bd68-r99dx   44m   Running                  harvester-node-2   True
rke2-3-pool1-47f7bd68-r99dx   44m   Running                  harvester-node-2   False
```

**To Reproduce**
1. Create a 3-node harvester cluster.
2. Provision a 3-node RKE2 cluster on the harvester and wait for it's ready.
3. Choose one node to stop kubelet.
    a. Find the pid of kubelet process by `ps -aux | grep kubelet`.
    b. Freeze the kubelet process by `kill -STOP <kubelet-pid>`. 
4. After the node is unavailable. Run `shutdown now -r`.
5. After the node is running again, the VM can't be ready.

**Expected behavior**
The RKE2 VM can be migrated to other hosts and the VM can be ready after a while.

**Support bundle**
[supportbundle_f3b2af65-e81b-4c19-939a-14ff8688c5a1_2023-10-19T03-37-52Z.zip](https://github.com/harvester/harvester/files/13037503/supportbundle_f3b2af65-e81b-4c19-939a-14ff8688c5a1_2023-10-19T03-37-52Z.zip)

**Environment**
 - Harvester ISO version: v1.2.0
[...]

PR #4645, opened 2023-10-23
feat: fix flipping between Paused and Running

FrankYang0529:
**Problem:**
VM status flipping between Paused and Running when host accidentally restarts.

**Solution:**
Check whether VM is stuck in PasuedIOError when the related Volume is healthy.

**Related Issue:**
https://github.com/harvester/harvester/issues/4633

**Test plan:**
2. Provision a 3-node RKE2 cluster on the harvester and wait for it's ready.
3. Choose one node to stop kubelet.
    a. Find the pid of kubelet process by `ps -aux | grep kubelet`.
    b. Freeze the kubelet process by `kill -STOP <kubelet-pid>`. 
4. After the node is unavailable. Run `shutdown now -r`.
5. After the node is running again, the VM can't be ready.

### DIFF
--- a/pkg/controller/master/virtualmachine/vm_controller.go
+++ b/pkg/controller/master/virtualmachine/vm_controller.go
@@ -1,12 +1,15 @@
 package virtualmachine
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
 	"time"
 
+	longhornv1 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	v1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/slice"
 	"github.com/sirupsen/logrus"
@@ -14,6 +17,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	kubevirt "kubevirt.io/api/core"
 	kubevirtv1 "kubevirt.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


FrankYang0529:
This PR can work on both v1.2.0 and v1.2.1-rc1. It's ready. Wait for missing elemental image fix (`registry.opensuse.org/isv/rancher/elemental/stable/teal53/15.4/rancher/elemental-builder-image/5.3:0.3.1`).

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -435,6 +483,21 @@ func (h *VMController) cleanUpInsufficientResourceAnnotation(vm *kubevirtv1.Virt
 	return err
 }
 
+func (h *VMController) isPVCReady(namespace, name string) (bool, error) {
+	pvc, err := h.pvcCache.Get(namespace, name)
+	if err != nil {
+		return false, err
+	}
+	volume, err := h.longhornVolumeCache.Get(util.LonghornSystemNamespaceName, pvc.Spec.VolumeName)
```
w13915984028, line 519: check the pvc provider/driver is Longhorn first, as we support third-party CSI

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -398,6 +404,48 @@ func (h *VMController) SetHaltIfInsufficientResourceQuota(_ string, vm *kubevirt
...
+	if vm == nil || vm.DeletionTimestamp != nil || vm.Spec.Template == nil {
+		return vm, nil
+	}
+
+	// if the vm is not paused or it's under restarting, skip it.
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusPaused || len(vm.Status.StateChangeRequests) != 0 {
+		return vm, nil
+	}
+
+	vmi, err := h.vmiCache.Get(vm.Namespace, vm.Name)
```
w13915984028, line 432: skip the vmi none-existing error ?

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -398,6 +404,48 @@ func (h *VMController) SetHaltIfInsufficientResourceQuota(_ string, vm *kubevirt
...
+		} else if !ready {
+			logrus.Debugf("Volume %s is not healthy in VM %s/%s, wait for 5 seconds to check VM again", volume.Name, vm.Namespace, vm.Name)
+			h.vmController.EnqueueAfter(vm.Namespace, vm.Name, 5*time.Second)
+			return vm, nil
+		}
+	}
+
+	for _, condition := range vmi.Status.Conditions {
+		if condition.Type == kubevirtv1.VirtualMachineInstancePaused && condition.Status == corev1.ConditionTrue &&
+			condition.Reason == "PausedIOError" {
```
w13915984028, line 453: Is `PausedIOError` defined anywhere? let's refer it or define it. Reuse also in next line log.

w13915984028 (commented):
Few questions, please double check, thanks.

pkg/controller/master/virtualmachine/vm_controller.go
FrankYang0529: No, I traced KubeVirt source code. They don't define `PausedIOError`.

https://github.com/kubevirt/kubevirt/blob/85e9173c308434a6e2a9099f89eb631782e67bde/pkg/virt-handler/vm.go#L1465

w13915984028: Then let's define `PausedIOError` as a const in this go file.

bk201:
@FrankYang0529 Thanks for the great work of this PR. But I'm not a fan of doing the workaround within the controller.
Instead, we should pursue the proper fix either in Longhorn or kubelet.
The flipping state is caused by an I/O error. If we restart the VM silently, it might hide the problem and no one will notice it.
Also by doing this, it might make it even harder to debug because the virt-launcher pod is gone and states change. We had some support experience and found keeping the crime scene is always what we prefer most.
I'd suggest we add the workaround steps and let the user do it.

**Update**: we discussed offline and agree: since the issue depends on several upstream issues and those fixes are not expected to arrive in a short time, we'll have a flag to let users optionally enable the behavior in this PR.

pkg/controller/master/virtualmachine/vm_controller.go
FrankYang0529: Yes, we check whether VM is paused at first. In a happy path, a paused VM has related VMI CR. If we can't find VMI in this case, there must have some errors.

FrankYang0529:
Close the PR because https://github.com/longhorn/longhorn-manager/pull/2287 can solve the issue in upstream.

Issue #4493, opened 2023-09-06
[BUG] Post Draining `return code: 32` on removal of Elemental issues


irishgordo:
**Describe the bug**

**_NOTE: This does not impact the upgrade process, the cluster was able to upgrade successfully_**

It seems at post-draining, there is a point where at least from the logs:
```
kubectl logs jobs/hvst-upgrade-md6wr-post-drain-harvester-node-1 -n harvester-system
```
It struggles to remove `elemental` binary:
```
+ GRUBENV_FILE=/oem/grubenv
+ chroot /host /bin/bash -c 'if ! [ -f /oem/grubenv ]; then grub2-editenv /oem/grubenv create; fi'
+ umount /host/usr/bin/elemental
+ umount /host/tmp/tmp.G0W6LinMB6
+ rm -rf /host/usr/local/upgrade_tmp/tmp.u8YWut8e3W
+ reboot_if_job_succeed
+ cat
+ cat
+ chmod +x /host/tmp/upgrade-reboot.sh
+ cat
+ chroot /host systemctl daemon-reload
+ chroot /host systemctl start upgrade-reboot
+ clean_up_tmp_files
+ '[' -n /host/tmp/tmp.G0W6LinMB6 ']'
Try to unmount /host/tmp/tmp.G0W6LinMB6...
+ echo 'Try to unmount /host/tmp/tmp.G0W6LinMB6...'
+ umount /host/tmp/tmp.G0W6LinMB6
umount: /host/tmp/tmp.G0W6LinMB6: not mounted.
+ echo 'Umount /host/tmp/tmp.G0W6LinMB6 failed with return code: 32'
+ '[' -n /host/usr/bin/elemental ']'
+ echo 'Try to unmount /host/usr/bin/elemental...'
+ umount /host/usr/bin/elemental
Umount /host/tmp/tmp.G0W6LinMB6 failed with return code: 32
Try to unmount /host/usr/bin/elemental...
umount: /host/usr/bin/elemental: not mounted.
+ echo 'Umount /host/usr/bin/elemental failed with return code: 32'
+ echo 'Clean up tmp files...'
+ '[' -n '' ']'
+ '[' -n /host/usr/local/upgrade_tmp/tmp.u8YWut8e3W ']'
+ echo 'Try to remove /host/usr/local/upgrade_tmp/tmp.u8YWut8e3W...'
+ rm -vf /host/usr/local/upgrade_tmp/tmp.u8YWut8e3W
[...]

PR #4646, opened 2023-10-23
fix(upgrade): check mountpoint exist before umount

starbops:

**Problem:**

The rootfs and elemental cli may be unmounted before the trap handler gets called. In that case, the handler won't be able to unmount those directories because they're already unmounted. Therefore, the script generates horrifying messages like `Umount /host/usr/bin/elemental failed with return code: 32` which scares the user. In fact, the upgrade process is still in good shape.

**Solution:**

Check if the directories are mounted right before unmounting them.

**Related Issue:**

#4493

**Test plan:**

1. Prepare a v1.2.0 Harvester cluster (single node is fine)
1. Build an custom ISO image with the fix
1. Upgrade the cluster with the custom ISO image
1. Check the logs when the upgrade progress is at the post-draining phase, there should be no `Umount xxx failed with return code: 32` errors

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -28,13 +28,18 @@ cleanup_incomplete_state_file() {
   fi
 }
 
+is_mounted()
+{
+  mount | awk -v DIR="$1" '{ if ($3 == DIR) { exit 0 } } ENDFILE { exit 1 }'
+}
+
 clean_up_tmp_files()
 {
-  if [ -n "$tmp_rootfs_mount" ]; then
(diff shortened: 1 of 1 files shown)


w13915984028 (commented):
LGTM, thanks.

w13915984028 (approved):
LGTM, thanks.

starbops:
cc @bk201 @Vicente-Cheng
Might need your help reviewing this one since v1.1.3 also requires the fix. Thanks!

Vicente-Cheng (approved):
LGTM, tested upgrade on single-node.

starbops:
@mergify backport v1.1

starbops:
@mergify backport v1.2

Issue #4604, opened 2023-10-12
[BUG] Restore from snapshot not work if target VM is restore-replaced from backup


albinsun:
## Description
Restore from snapshot not work if target VM is restore-replaced from backup

## Environment
* Harvester
  * Version: `v1.2.1-rc1`
  * Profile: QEMU/KVM, 3 node (8C/16G/250G)
  * **ui-source**: Auto

## To Reproduce
> **Note** Click "►" to expand details.
1. <details>
   <summary>
   Config backup-target ${~~~\color{green}\textsf{V}}$
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/78ecbbf6-bfda-4fc2-b749-87985233422d)
   </details>

1. <details>
   <summary>
   Create VM, wait VM `Running` ${~~~\color{green}\textsf{V}}$
   </summary>
  
   ![image](https://github.com/harvester/harvester/assets/2773781/9146734b-4063-43b9-a737-20b3ef793a1d)
   ![image](https://github.com/harvester/harvester/assets/2773781/642898b7-5b67-48d7-b677-e91da17175ea)
   </details>

1. <details>
   <summary>
   Take Backup, wait backup `Ready` ${~~~\color{green}\textsf{V}}$
   </summary>

   Go to "Virtual Machines" -> VM -> "⋮" -> "Take Backup"
   ![image](https://github.com/harvester/harvester/assets/2773781/1995b671-ab46-4904-85c2-139b5b23a7d3)
   </details>

1. <details>
   <summary>
   Take VM Snapshot, wait snapshot `Ready` ${~~~\color{green}\textsf{V}}$
   </summary>

   Go to "Virtual Machines" -> VM -> "⋮" -> "Take VM Snapchot"
   ![image](https://github.com/harvester/harvester/assets/2773781/a11c5f1f-14dd-4469-9a22-fe94ee474e59)
   </details>

1. <details>
   <summary>
   Stop VM, wait VM `Off` ${~~~\color{green}\textsf{V}}$
   </summary>
   
   Go to "Virtual Machines" -> VM -> "Stop"
   ![image](https://github.com/harvester/harvester/assets/2773781/d91a00fc-0874-4ef8-8efb-66107928ca1c)
   </details>

1. <details>
   <summary>
   Restore (Replace) from backup, wait VM `Running` ${~~~\color{green}\textsf{V}}$
   </summary>
   
[...]

PR #4664, opened 2023-10-27
fix: prevent vmrestore with delete policy if there is vmbackup snapshot

FrankYang0529:
**Problem:**
Restore from snapshot not work if target VM is restore-replaced from backup

**Solution:**
Add webhook to prevent vmrestore with deletion policy if there is vmbackup snapshot with same pvcs.

**Related Issue:**
https://github.com/harvester/harvester/issues/4604

**Test plan:**
1. Config backup-target
2. Create VM, wait VM `Running`
3. Take Backup, wait backup `Ready`
4. Take VM Snapshot, wait snapshot `Ready`
5. Stop VM, wait VM `Off`
6. Replace source vm from backup with deletion policy, webhook denies the request
7. Replace source vm from backup with retain policy, wait VM `Running`
8. Stop VM again
9. Replace source vm from snapshot, wait VM `Running`
10. Restore (New) from snapshot

### DIFF
--- a/pkg/webhook/resources/virtualmachinerestore/validator.go
+++ b/pkg/webhook/resources/virtualmachinerestore/validator.go
@@ -224,7 +224,7 @@ func (v *restoreValidator) checkBackup(vmRestore *v1beta1.VirtualMachineRestore,
 		}
 
 		if len(vmBackupSnapshotNames) != 0 {
-			return fmt.Errorf("can't use delete policy, %s is used by VMBackup snapshots %s", pvcNamespaceAndName, strings.Join(vmBackupSnapshotNames, ","))
+			return fmt.Errorf("can't use delete policy, the volume %q is used by the VM Snapshot(s) %s", pvcNamespaceAndName, strings.Join(vmBackupSnapshotNames, ", "))
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


WebberHuang1118 (approved):
Testing cases work as expected, LGTM, thanks.

pkg/webhook/resources/virtualmachinerestore/validator.go
```diff
@@ -199,6 +201,35 @@ func (v *restoreValidator) checkSnapshot(vmRestore *v1beta1.VirtualMachineRestor
...
+			return err
+		}
+
+		var vmBackupSnapshotNames []string
+		for _, vmBackupSnapshot := range vmBackupSnapshots {
+			vmBackupSnapshotNames = append(vmBackupSnapshotNames, vmBackupSnapshot.Name)
+		}
+
+		if len(vmBackupSnapshotNames) != 0 {
+			return fmt.Errorf("can't use delete policy, %s is used by VMBackup snapshots %s", pvcNamespaceAndName, strings.Join(vmBackupSnapshotNames, ","))
```
bk201:
```suggestion
			return fmt.Errorf("can't use delete policy, the volume %q is used by the VM Snapshot(s) %s", pvcNamespaceAndName, strings.Join(vmBackupSnapshotNames, ", "))
```

bk201: We can just use the term VM snapshot here to avoid confusion.

bk201 (approved):
LGTM, thanks!

FrankYang0529:
@Mergify backport v1.2

Issue #4612, opened 2023-10-13
[BUG] VMBackup metadata file name conflicts


FrankYang0529:
**Describe the bug**
We use `<vmbackup-namespace>-<vmbackup-name>.cfg` as metadata file in backup target. If there is a vmbackup with namespace `a` and name `b-c`, it conflicts with namespace `a-b` and name `c`.

**To Reproduce**
1. Create namespaces `a-b` and `a`.
```
kubectl create ns a-b
kubectl create ns a
```
2. Create two VMs. One is in namespace `a-b` and another in `a`.
3. Take a backup for each VM. The backup name is `c` for VM in namespace `a-b`. Another name is `b-c` for VM in namespace `a`.
4. Run `ls /opt/backupstore/harvester/vmbackups` in backup target. We can only see one `a-b-c.cfg` file.

**Expected behavior**
There are two metadata files. One is for `a-b/c` and another is for `a/b-c` VMBackup.

**Environment**
 - Harvester ISO version: v1.2.0

bk201:
We need to add this information to the doc (e.g., https://docs.harvesterhci.io/v1.2/vm/backup-restore/)
- Path migration. Why it's needed and example.
- Best practice when using a backup store for multiple clusters.

Then mention the doc in the release note.

albinsun:
Provide test result on `v1.3.0-dev-20240105`, close as fixed.


## Environment
* Harvester
 * Version: `v1.3.0-dev-20240105`
 * Profile: QEMU/KVM, 2 nodes (8C/16G/500G)
 * ui-source: `Auto`


## Steps
1. Create namespace **n1** and **n1-n2**
   <details>
 
     ![image](https://github.com/harvester/harvester/assets/2773781/3366fc19-c472-4d6d-93f5-2690e2ac0189)
   </details>


1. Create 2 VMs
   <details>
 
     ![image](https://github.com/harvester/harvester/assets/2773781/9037468f-b382-4e9d-9956-68e2ae9e7455)
   </details>
   
   * In **n1** namespace, create VM **n2-n3**
   * In **n1-n2** namespace, create VM **n3**


1. Create VM backup
   <details>
 
     ![image](https://github.com/harvester/harvester/assets/2773781/838087d6-cb65-47a4-a065-5e25ffdad322)
   </details>
   
   * VM **n1** / **n2-n3** -> Backup name **n2-n3**
   * VM **n1-n2** / **n3** -> Backup name **n3**


[...]

PR #4667, opened 2023-10-30
refactor: move vmbackup metadata to namespace folder

FrankYang0529:
**Problem:**

We use `<vmbackup-namespace>-<vmbackup-name>.cfg` as metadata file in backup target. If there is a vmbackup with namespace `a` and name `b-c`, it conflicts with namespace `a-b` and name `c`.

**Solution:**
Make `namespace` as a folder.

**Related Issue:**
https://github.com/harvester/harvester/issues/4612

**Test plan:**
1. Create Harv v1.2.1 cluster.
2. Create a VM as `vm1`.
3. Setup backup target.
4. Create a VMBackup as `b1`.
5. Check the folder in backup target. There is a file `default-b1.cfg`.
6. Build a new image from this PR.
7. Replace `harvester/harvester` deployment image with this PR.
8. Check harvester pod log. There is a message like `move vm backup metadata default/b1 from harvester/vmbackups/default-b1.cfg to harvester/vmbackups/default/b1.cfg`.
9. Check the folder in backup target. There is no file `default-b1.cfg`, but there is a file path like `default/b1.cfg`.
10. Restore the `b1` VMBackup as a new VM `vm2`. The content in both VMs is same.
11. Stop `vm1`.
12. Replace the `vm1` VM with `b1` VMBackup. There is no error in restoring process.
13. Delete the `b1` VMBackup.
14. Check the folder in backup target.  There is a file `default/b1.cfg` is removed.

### DIFF
--- a/pkg/controller/master/backup/backup.go
+++ b/pkg/controller/master/backup/backup.go
@@ -12,7 +12,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 	"reflect"
 	"strconv"
 	"time"
@@ -804,7 +803,7 @@ func (h *Handler) deleteVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
 		return err
 	}
 
-	destURL := filepath.Join(metadataFolderPath, getVMBackupMetadataFileName(vmBackup.Namespace, vmBackup.Name))
+	destURL := getVMBackupMetadataFilePath(vmBackup.Namespace, vmBackup.Name)
 	if exist := bsDriver.FileExists(destURL); exist {
 		logrus.Debugf("delete vm backup metadata %s/%s in backup target %s", vmBackup.Namespace, vmBackup.Name, target.Type)
 		return bsDriver.Remove(destURL)
@@ -868,7 +867,7 @@ func (h *Handler) uploadVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
 	}
 
 	shouldUpload := true
-	destURL := filepath.Join(metadataFolderPath, getVMBackupMetadataFileName(vmBackup.Namespace, vmBackup.Name))
+	destURL := getVMBackupMetadataFilePath(vmBackup.Namespace, vmBackup.Name)
 	if bsDriver.FileExists(destURL) {
 		if remoteVMBackupMetadata, err := loadBackupMetadataInBackupTarget(destURL, bsDriver); err != nil {
 			return err
--- a/pkg/controller/master/backup/backup_metadata.go
+++ b/pkg/controller/master/backup/backup_metadata.go
@@ -1,6 +1,7 @@
 package backup
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -132,17 +133,38 @@ func (h *MetadataHandler) syncVMBackup(target *settings.BackupTarget) error {
 		return err
 	}
 
+	namespaceFolderSet := map[string]bool{} // ignore value of map, we only use it as a set
+	requiredMovingFilePaths := []string{}
 	for _, fileName := range fileNames {
-		backupMetadata, err := loadBackupMetadataInBackupTarget(filepath.Join(metadataFolderPath, fileName), bsDriver)
+		filePath := filepath.Join(metadataFolderPath, fileName)
+		if bsDriver.FileExists(filePath) {
+			requiredMovingFilePaths = append(requiredMovingFilePaths, filePath)
+			continue
+		}
+		namespaceFolderSet[filePath] = true
+	}
+
+	if err = h.moveFilePaths(requiredMovingFilePaths, bsDriver, namespaceFolderSet); err != nil {
+		return err
+	}
+
+	vmbackupMetadataFilePaths := []string{}
+	for namespaceFolder := range namespaceFolderSet {
+		fileNames, err := bsDriver.List(namespaceFolder)
 		if err != nil {
 			return err
 		}
-		if backupMetadata.Namespace == "" {
-			backupMetadata.Namespace = metav1.NamespaceDefault
+
+		for _, fileName := range fileNames {
+			filePath := filepath.Join(namespaceFolder, fileName)
+			if bsDriver.FileExists(filePath) {
+				vmbackupMetadataFilePaths = append(vmbackupMetadataFilePaths, filePath)
+			}
 		}
-		if err := h.createVMBackupIfNotExist(*backupMetadata, target); err != nil {
-			return err
-		}
+	}
+
+	if err = h.loadBackupMetadataAndCreateVMBackup(target, vmbackupMetadataFilePaths, bsDriver); err != nil {
+		return err
 	}
 	return nil
 }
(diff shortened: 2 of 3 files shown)


pkg/controller/master/backup/backup.go
```diff
@@ -804,7 +804,7 @@ func (h *Handler) deleteVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
 		return err
 	}
 
-	destURL := filepath.Join(metadataFolderPath, getVMBackupMetadataFileName(vmBackup.Namespace, vmBackup.Name))
+	destURL := filepath.Join(metadataFolderPath, vmBackup.Namespace, vmBackup.Name)
```
connorkuehl: deleteVMBackupMetadata seems to generate a destURL that is asymmetrical to uploadVMBackupMetadata.

deleteVMBackupMetadata constructs `.../metadataFolderPath/vmBackup.Namespace/vmBackup.Name` but uploadVMBackupMetadata calls getVMBackupMetadataFilePath which returns `.../metadataFolderPath/vmBackup.Namespace/vmBackup.Name.cfg` (with .cfg extension).

Is that intentional?

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -132,18 +133,40 @@ func (h *MetadataHandler) syncVMBackup(target *settings.BackupTarget) error {
...
+		} else {
+			namespaceFolders = append(namespaceFolders, filePath)
+		}
+	}
+
+	// collect all vmbackup metadata file paths in vmbackupMetadataFilePaths
+	vmbackupMetadataFilePaths := make([]string, len(requiredMovingFilePaths))
+	copy(vmbackupMetadataFilePaths, requiredMovingFilePaths)
+	for _, namespaceFolder := range namespaceFolders {
+		fileNames, err := bsDriver.List(filepath.Join(metadataFolderPath))
```
connorkuehl: This appears to be identical to the call on line 131. I don't think the original value from 131 has been mutated, so could it be used here instead of calling bsDriver.List inside the loop?

pkg/controller/master/backup/backup.go
FrankYang0529: Yeah, I forgot to change this one. Updated it. Thank you.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: Sorry, I used a wrong file path here. I would like to like files under namespace folder. Thanks for catching this!

connorkuehl (approved):

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -132,18 +133,40 @@ func (h *MetadataHandler) syncVMBackup(target *settings.BackupTarget) error {
...
+	var requiredMovingFilePaths []string
 	for _, fileName := range fileNames {
-		backupMetadata, err := loadBackupMetadataInBackupTarget(filepath.Join(metadataFolderPath, fileName), bsDriver)
+		filePath := filepath.Join(metadataFolderPath, fileName)
+		if bsDriver.FileExists(filePath) {
+			requiredMovingFilePaths = append(requiredMovingFilePaths, filePath)
+		} else {
+			namespaceFolders = append(namespaceFolders, filePath)
+		}
+	}
```
WebberHuang1118, line 145: Just a Nit, how about following for using less brackets?
```
	for _, fileName := range fileNames {
		filePath := filepath.Join(metadataFolderPath, fileName)
		if bsDriver.FileExists(filePath) {
			requiredMovingFilePaths = append(requiredMovingFilePaths, filePath)
                        continue;
		} 
		namespaceFolders = append(namespaceFolders, filePath)
	}
```

WebberHuang1118: Just a Nit, how about following for using less brackets?
```
	for _, fileName := range fileNames {
		filePath := filepath.Join(metadataFolderPath, fileName)
		if bsDriver.FileExists(filePath) {
			requiredMovingFilePaths = append(requiredMovingFilePaths, filePath)
                        continue;
		} 
		namespaceFolders = append(namespaceFolders, filePath)
	}
```

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -132,18 +133,40 @@ func (h *MetadataHandler) syncVMBackup(target *settings.BackupTarget) error {
...
+			}
 		}
 	}
+
+	if err = h.loadBackupMetadataAndCreateVMBackup(target, vmbackupMetadataFilePaths, bsDriver); err != nil {
+		return err
+	}
+	if err = h.moveFilePaths(requiredMovingFilePaths, bsDriver); err != nil {
+		return err
+	}
```
WebberHuang1118: If my understanding is correct, this would load metadata and create vmbackup CR first, then move existing metadata to the correct path in the backup target if needed ?

I'm considering if there is any power outage on the node or the backup target, and making incomplete metadata left on target. During next restart, would this bring conflict with vmbackup controller [handleBackupReady()](https://github.com/FrankYang0529/harvester/blob/bc7abe2d219b490b40b41736800326d3473ea293/pkg/controller/master/backup/backup.go#L979-L1000)?

WebberHuang1118 (commented):
Some minor questions, thanks.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: I think the situation is fine, because the vm metadata which is used to create VMBackup in a new cluster will be the same content in [uploadVMBackupMetadata](https://github.com/FrankYang0529/harvester/blob/bc7abe2d219b490b40b41736800326d3473ea293/pkg/controller/master/backup/backup.go#L999). Both `moveFilePaths` and `uploadVmBackupMetadata` are trying to save same content to a same path. The different order of `moveFilePaths` and `uploadVmBackupMetadata` will get same result.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: Updated it. Thanks.

pkg/controller/master/backup/backup_metadata.go
WebberHuang1118: Such interleaving seems avoidable? What if change the order to make the metadata operation sequential?

FrankYang0529: Do you mean that we only move files in `uploadVMBackupMetadata`?

WebberHuang1118: I mean like:
```
	if err = h.moveFilePaths(requiredMovingFilePaths, bsDriver); err != nil {
		return err
	}
	if err = h.loadBackupMetadataAndCreateVMBackup(target, vmbackupMetadataFilePaths, bsDriver); err != nil {
		return err
	}
```
Would it cost lots of effort? I think it may introduce some restructuring on `syncVMBackup`. Thanks.

WebberHuang1118 (approved):
LGTM, thanks.

bk201 (approved):
LGTM.

Confirm the path are migrated for both S3/NFS paths. And I can restore backup after upgrading from v1.2.1.

PR #4658, opened 2023-10-25
fix: missing honestbee/jobq

FrankYang0529:
**Problem:**
Run `go mod tidy` and we will get following response:

```
go: github.com/longhorn/longhorn-engine@v1.4.2-0.20230505032747-3343bb5fba06 requires
	github.com/honestbee/jobq@v1.0.2: reading github.com/honestbee/jobq/go.mod at revision v1.0.2: git ls-remote -q origin in /Users/poanyang/go/1.21.3/pkg/mod/cache/vcs/b7039093cfd5e54747139d9a392f45ffb733fb74488bd673022aa3b33e2ae79f: exit status 128:
	remote: Repository not found.
	fatal: repository 'https://github.com/honestbee/jobq/' not found
```

**Solution:**
Upgrade longhorn package dependence.

**Related Issue:**
https://github.com/harvester/harvester/issues/4651

Issue #4611, opened 2023-10-13
[ENHANCEMENT] Check free disk space percent before upgrade


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
From [container image lifecycle document](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#container-image-lifecycle), kubelet triggers grabage collection when disk free space percent is higher than `HighThresholdPercent`. This may make some risks in airgapped environment, because it can't get old images if we don't set container registry correctly.

For example, [missing vmdp image issue](https://github.com/harvester/harvester/issues/4534).

**Describe the solution you'd like**
Add a webhook to check free disk space percent like PR [Check disk space before upgrade](https://github.com/harvester/harvester/pull/3700#top).

TachunLin:
Verified **fixed** on `master-d3dc81ca-head` (24/01/03). Close this issue.

### Result
- [x] 1. The node without enough disk spaces in `/usr/local` can't upgrade

    <details><summary>Test Result $~~$ $\color{green}{\textsf{PASS}}$ </summary>

    When /usr/local did not have enough spaces
    
    ```
    rancher@n1-240104:~> df -h /usr/local
    Filesystem      Size  Used Avail Use% Mounted on
    /dev/vda5       147G   28G  112G  21% /usr/local
    rancher@n1-240104:~> sudo dd if=/dev/zero of=/usr/local/test.img bs=1G count=93
    93+0 records in
    93+0 records out
    99857989632 bytes (100 GB, 93 GiB) copied, 327.656 s, 305 MB/s
    rancher@n1-240104:~> df -h /usr/local
    Filesystem      Size  Used Avail Use% Mounted on
    /dev/vda5       147G  116G   24G  84% /usr/local
    ```
    
    If we trigger the upgrade, the webhook would prompt message to prevent to proceed
    
    ![image](https://github.com/harvester/harvester/assets/29251855/2e85353e-b05a-4c6c-b95b-b47680514d82)


    </details>

- [x] 2. When the cluster is ready, check Longhorn system-managed components have priority class set.

    <details><summary>Test Result $~~$ $\color{green}{\textsf{PASS}}$ </summary>

    When we free the /usr/local spaces
    
    ```
    rancher@n1-240104:/usr/local> sudo rm test.img
    rancher@n1-240104:/usr/local> df -h /usr/local
    Filesystem      Size  Used Avail Use% Mounted on
    /dev/vda5       147G   23G  118G  16% /usr/local
[...]

Issue #4534, opened 2023-09-15
[BUG] VMDP Image wrong after upgrade to Harvester 1.2.0


Martin-Weiss:
**Describe the bug**
Have a windows 2022 VM deployed on top of harvester 1.1.x and then upgrade to Harvester 1.2.0.
VM does NOT start anymore, because the VMDP image assigned is [registry.suse.com/harvester-beta/vmdp:latest](http://registry.suse.com/harvester-beta/vmdp:latest) which is not available air-gapped at all and which is not included in the list of harvester images for harvester 1.2.0.

**To Reproduce**
Create Windows 2022 VM on Harvester 1.1.x and then upgrade harvester to 1.2.0.
Stop and start the VM.

**Expected behavior**
- the tag :latest must not be used at all (requires registry connection and is not reproducible / immutable)
- the image used should be part of the harvester image list so we can add it to the on-premises registry
- we should have a central configuration place to manage the VMDP images used in the environment

**Environment**
 - Harvester ISO version: 1.2.0 (upgrade from 1.0.0 over all intermediate versions to 1.2.0)
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): not relevant

FrankYang0529:
Hi @Martin-Weiss, we've changed vmdp image from `latest` to a number version since v1.1.0 ([ref](https://github.com/harvester/harvester/commit/088a020fcd47f3e477a62bb725170c8bd3bf3aaa)). Currently, the vmdp on v1.2.0 harvester is `registry.suse.com/suse/vmdp/vmdp:2.5.4.2`. Could you try to update `virtio-container-disk` in your VM to see whether the VM can be started?

FrankYang0529:
Hi @guangbochen, do you know whether there is mapping between virt-operator and vmdp? For example, virt-operator `0.54.0-150400.3.16.1` needs to use vmdp `2.5.4.2`? If yes, I think we may need a controller to auto update vmdp version for windows VM.

cc @bk201

guangbochen:
VMDP [v2.5.4.2](https://github.com/SUSE/vmdp/releases) is the latest release from the SUSE and it is fully compatible with the current kubevirt v0.54.0 version. Because the VM template follows immutable design so we didn't manage to update the old template to a newer version during the Harvester upgrade, and users will only get the more recent template version in the newly installed Harvester cluster. 

[...]

PR #4701, opened 2023-11-03
Check free disk space percent before upgrade

FrankYang0529:
**Problem:**
From [container image lifecycle document](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#container-image-lifecycle), kubelet triggers grabage collection when disk free space percent is higher than `HighThresholdPercent`. This may make some risks in airgapped environment, because it can't get old images if we don't set container registry correctly.

For example, [missing vmdp image issue](https://github.com/harvester/harvester/issues/4534).

**Solution:**
Add a webhook to check free disk space percent like PR [Check disk space before upgrade](https://github.com/harvester/harvester/pull/3700#top).

**Related Issue:**
https://github.com/harvester/harvester/issues/4611

**Test plan:**
### Setup

1. Create a new Harvester cluster. Each node's disk space should be 250G.
2. If every node's `/usr/local` free space is more than 30G, you can use `dd` to write files to make sure one of the node's free disk space is 20G. For example: `dd if=/dev/zero of=/usr/local/test.img bs=1G count=93`.
3. Create a test version.
```yaml
kind: Version
metadata:
  annotations:
    harvesterhci.io/minFreeDiskSpaceGB: "5"
  name: 1.2.2
  namespace: harvester-system
spec:
  isoURL: http://192.168.0.181:8000/harvester-v1.2.2-amd64.iso
  minUpgradableVersion: 1.2.1
  releaseDate: "20231210"
  tags:
  - dev
  - test
```

Screenshot 2023-11-03 at 7 41 02 PM

### Case 2: a node with enough disk spaces can upgrade

1. Remove the `dd` file.
2. Click the upgrade again. There is no error message.

**Additional context**
* [X] Rely on https://github.com/harvester/harvester/pull/4658.


w13915984028 (commented):
A question:

would we allow adding an annotation to skip the new check? in none air-gapped cluster, or cluster with image registry, user will have chance to skip such situation, thanks.

FrankYang0529:
> A question:
> 
> would we allow adding an annotation to skip the new check? in none air-gapped cluster, or cluster with image registry, user will have chance to skip such situation, thanks.

Yeah, we already have it.

https://github.com/harvester/harvester/blob/7a7daa7e500eb0d2306fa05106106ccbcb936b10/pkg/webhook/resources/upgrade/validator.go#L126-L130

w13915984028 (approved):
LGTM, thanks.

Yu-Jack (approved):
LGTM.

BTW, because I'm curious whether I could register kubelet.config.k8s.io into factory like following, so we might use `Kubelet().V1beta1().KubeletConfiguration().Get()` to get config. 

```
kubeletconfigv1.SchemeGroupVersion.Group: {
	Types: []interface{}{
		kubeletconfigv1.KubeletConfiguration{},
	},
	GenerateTypes:   true,
	GenerateClients: true,
},
```

But, I found it lacks of `KubeletConfigurationList` in types, so it failed.

FrankYang0529:
> BTW, because I'm curious whether I could register kubelet.config.k8s.io into factory like following, so we might use `Kubelet().V1beta1().KubeletConfiguration().Get()` to get config.
> 
> ```
> kubeletconfigv1.SchemeGroupVersion.Group: {
> 	Types: []interface{}{
> 		kubeletconfigv1.KubeletConfiguration{},
> 	},
> 	GenerateTypes:   true,
> 	GenerateClients: true,
> },
> ```
> 
> But, I found it lacks of `KubeletConfigurationList` in types, so it failed.

If I understand correctly, `KubeletConfiguration` is not stored in etcd, so we can't get it from API server. We have to get it from each kubelet server.

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -252,34 +254,33 @@ func (v *upgradeValidator) checkDiskSpace(node *corev1.Node, minFreeDiskSpace ui
...
-	body, err := ioutil.ReadAll(resp.Body)
+	kubeletConfiguration, err := v.getKubeletConfigz(node.Name, kubeletURL)
 	if err != nil {
-		return werror.NewInternalError(fmt.Sprintf("node %s, can't read response from %s, err: %+v", node.Name, kubeletURL, err))
+		return err
 	}
-	if err = json.Unmarshal(body, &summary); err != nil {
-		return werror.NewInternalError(fmt.Sprintf("node %s, can't parse json response %s, err: %+v", node.Name, string(body), err))
+
+	imageGCHighThresholdPercent := 85.0 // default value in kubelet config
```
Yu-Jack: BTW, I think we could move default percent to top area as const?

FrankYang0529: Add `defaultImageGCHighThresholdPercent`. Thank you.

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -252,34 +255,33 @@ func (v *upgradeValidator) checkDiskSpace(node *corev1.Node, minFreeDiskSpace ui
...
+	imageGCHighThresholdPercent := defaultImageGCHighThresholdPercent
+	if kubeletConfiguration.ImageGCHighThresholdPercent != nil {
+		imageGCHighThresholdPercent = float64(*kubeletConfiguration.ImageGCHighThresholdPercent)
 	}
-	if summary.Node.Fs.AvailableBytes == nil {
-		return werror.NewInternalError(fmt.Sprintf("can't get node %s available bytes from %s, err: %+v", node.Name, kubeletURL, err))
+	usedPercent := (float64(*summary.Node.Fs.UsedBytes) / float64(*summary.Node.Fs.CapacityBytes)) * 100.0
+	logrus.Debugf("node %s uses %3f%% storage space, kubelet image garbage collection threshold is %3f%%", node.Name, usedPercent, imageGCHighThresholdPercent)
+
+	if usedPercent > imageGCHighThresholdPercent {
```
bk201, line 281: @FrankYang0529 I suggest we use the condition `"userPercent + (expected new image capacity)" > imageGCHighThresholdPercent`. The problem we saw before is that after/during preloading images, a GC removes new images. We can add a reasonable value of `(expected new image capacity)` by doing some experiment...

FrankYang0529: Updated it. Thank you.

bk201 (approved):
LGTM. The test plan needs an update. QA will need to calculate the a reasonable size which is near the threshold.

FrankYang0529:
@Mergify backport v1.2

Issue #4712, opened 2023-11-08
[EPIC] Kubernetes-native Elemental CloudInit objects


connorkuehl:
Stories:

- [x] https://github.com/harvester/harvester/issues/4709
- [x] https://github.com/harvester/harvester/issues/3902
- [x] https://github.com/harvester/harvester/issues/4710
- [x] https://github.com/harvester/harvester/issues/4708

irishgordo:
@connorkuehl 

Based around prior validation and additionally that:
- https://github.com/harvester/harvester/issues/5240

Will be in 1.4.0 milestone.

And that manual test docs have been merged in:
- https://github.com/harvester/tests/pull/1100

And that:
- https://github.com/harvester/harvester/issues/5110
has been created to track.

I'll go ahead and close this out :smile: :+1: 

Thanks for the awesome feature and docs!

Issue #4709, opened 2023-11-08
[ENHANCEMENT] Add a HEP for the CloudInit node-manager CRD


connorkuehl:
**Is your enhancement related to a problem? Please describe.**

In order to address https://github.com/harvester/harvester/issues/3902, a new CRD must be created and its corresponding controller behaviors should be well-understood so that the necessary changes to Harvester components can be carried out.

**Describe the solution you'd like**

Open a pull request with a HEP to begin these discussions more broadly.

Issue #4711, opened 2023-11-08
[ENHANCEMENT] Add CRUD operations for Elemental CloudInit CRs in the UI


connorkuehl:
**Details subject to change until the CloudInit CRD HEP is agreed-upon and merged** (TODO: add link to HEP when it exists)

**Is your enhancement related to a problem? Please describe.**

https://github.com/harvester/harvester/issues/3902

There is a new enhancement underway to distribute Elemental cloud-init documents to Harvester nodes.

**Describe the solution you'd like**

Add the ability to Create, Update, Read, Delete CloudInit CRs in the Harvester dashboard.

Perhaps a new page under Advanced, or a sub-page under Hosts (?).

TODO: Add a description of the CloudInit object once it is finally merged in the backend.

**Describe alternatives you've considered**

**Additional context**

markhillgit:
moving this to 1.4.0 (we are already at the code freeze date for 1.3.0)

PR #4713, opened 2023-11-08
Add CloudInit CRD HEP

connorkuehl:
**Related Issue:**

https://github.com/harvester/harvester/issues/4712
https://github.com/harvester/harvester/issues/4709

### DIFF
--- a/enhancements/20231108-elemental-cloudinit-crd.md
+++ b/enhancements/20231108-elemental-cloudinit-crd.md
@@ -1,0 +1,376 @@
+# Kubernetes-native Elemental cloud-init documents
+
+## Summary
+
+It is both an expected and supported behavior for Harvester users to
+persist certain configurations to their Harvester nodes. These are currently
+done with Elemental cloud-init documents under the `/oem` path on the system.
+
+Currently, Harvester users would have to write these cloud-init documents
+and manually push them to each node where they would like to effect
+the desired change.
+
+There is a cloud-native tooling gap with regards to this process, however.
+Harvester users might prefer to use a GitOps-style approach to configuring
+their Harvester clusters, where Kubernetes objects are version-controlled
+as YAML and each new revision is applied to the cluster.
+
+This HEP presents a new Custom Resource Definition (CRD) that is surfaced
+via the harvester-node-manager, where a Harvester user could apply a "CloudInit"
+CR to a Harvester cluster, like so:
+
+```yaml
+apiVersion: node.harvesterhci.io/v1beta1
+kind: CloudInit
+metadata:
+  name: ssh-access
+spec:
+  matchSelector: {}
+  filename: 99_ssh.yaml
+  contents: |
+    stages:
+      network:
+        - authorized_keys:
+            rancher:
+              - ssh-ed25519 AAAA...
+```
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/4712
+
+## Motivation
+
+Provide a GitOps-compatible way of persisting configuration changes to
+a Harvester cluster.
+
+### Goals
+
+* Elemental cloud-init documents are stored in a CloudInit object spec and
+   pulled down onto nodes that match the selector of this CloudInit object.
+* CloudInit object's status field describes how it is rolled out
+  across the cluster (for example: which nodes does it apply to,
+  are any out of sync, etc).
+
+### Non-goals [optional]
+
+* Automatic rollbacks. If a Harvester user wants to "undo" a previously applied
+  CloudInit object, they should expect to:
+
+  1. Revise the CloudInit object contents and reboot; or
+  2. Adjust the CloudInit object match selector to exclude a node that they
+     no longer want it to apply to and reboot; or
+  3. Delete the CloudInit object and reboot the nodes it applied to.
+
+* Automatic execution. Harvester users should expect to reboot a
+  Harvester node to allow the resulting cloud-init documents to be
+  applied at boot.
+
+* Stanza/block-level editing. The CloudInit resource's `.spec.contents`
+  field will fully overwrite the file that is referred to in the resource's
+  `.spec.filename` field.
+
+## Proposal
+
+### User Stories
+
+Harvester users may customize their Harvester nodes to the fullest extent
+that the [Elemental toolkit](
+https://rancher.github.io/elemental-toolkit/docs/reference/cloud_init/) allows.
+
+Because of this, there is a very broad and nuanced set of customizations a Harvester
+user can describe in a cloud-init-style file.
+
+#### Add/rotate rancher user authorized keys (API)
+
+For example, a user could distribute and/or rotate the `rancher` user's authorized
+keys by creating a CloudInit resource like so:
+
+```yaml
+apiVersion: node.harvesterhci.io/v1beta1
+kind: CloudInit
+metadata:
+  name: ssh-access
+spec:
+  matchSelector: {}
+  filename: 99_ssh.yaml
+  contents: |
+    stages:
+      network:
+        - authorized_keys:
+            rancher:
+              - ssh-ed25519 AAAA...
+```
+
+Then, the user could monitor the ssh-access CloudInit resource to know the status
+of its rollout to the Harvester cluster:
+
+```
+$ kubectl describe cloudinit ssh-access
+# Metadata and Spec omitted for brevity...
+Status:
+  Rollouts:
+    Conditions:
+      Last Transition Time:  2023-10-24T21:15:02Z
+      Message:               
+      Reason:                CloudInitApplicable
+      Status:                True
+      Type:                  Applicable
+      Last Transition Time:  2023-10-24T21:09:28Z
+      Message:               Local file checksum is the same as the CloudInit checksum
+      Reason:                CloudInitChecksumMatch
+      Status:                False
+      Type:                  OutOfSync
+      Last Transition Time:  2023-10-24T21:09:28Z
+      Message:               99_ssh.yaml is present under /oem
+      Reason:                CloudInitPresentOnDisk
+      Status:                True
+      Type:                  Present
+    Node Name:               harv1
+```
+
+In the case of this single-node cluster, the user can see that the
+cloud-init-style file has been copied down to `/oem/99_ssh.yaml` on
+node `harv1`.
+
+They can also see that the file is up-to-date since the rollout says
+it is not `OutOfSync`.
+
+The user probably already knows that it is `Applicable` to `harv` because
+the `ssh-access` CloudInit resource spec says `matchSelector: {}`, which is
+a label set that will match every node. That said, if they updated the
+`matchSelector` down the line, then the `Applicable` condition may help
+them troubleshoot why a node has or does not have a cloud-init-style file
+copied down to it.
+
+The user can now reboot the Harvester node so that their cloud-init-style
+file may be applied at boot time.
+
+#### Add/rotate rancher user authorized keys (UI/Dashboard)
+
+Alternatively, instead of creating the `ssh-access` resource via kubectl,
+there could be a new form added to the Harvester dashboard (perhaps as a
+sub-page under the "Hosts" section.) This page could have a table with
+a list of the cluster's CloudInit resources.
+
+At the far right of each row for each CloudInit resource, there could be a
+3-dot menu much like the Virtual Machines page (and others), where a Harvester
+user could click "Edit Config", "Edit YAML", or "delete."
+
+There could also be a button called "Create".
+
+The "Create" and "Edit CloudInit" pages could have a post form that looks
+a little something like this (though the "Edit CloudInit page" will have
+values pre-populated such as resource name, file name, labels, contents):
+
+1. CloudInit resource name (in the case of the above example, `ssh-access`.)
+1. The base filename (in the above example, `99_ssh.yaml`.)
+1. A `key` input and a `value` input (like the `Labels` form while editing
+a Host config). Another row of `key` and `value` inputs can be added by clicking
+the `Add Label` button. In the case of the `ssh-access` example above, these
+would be left empty so that the UI passes an empty match selector to match all
+nodes (`matchSelector: {}`.)
+1. A textbox where the cloud-init YAML contents go.
+1. A "save" button (or it could be labeled "Create" or "Update" depending on
+   if this is a new CloudInit object or if the user is editing a pre-existing one.)
+
+After clicking "save/create/update", the dashboard could redirect the user to
+a more detailed page (much like the one you see for a virtual machine when clicking
+on a virtual machine's name on the "Virtual Machines" page of the dashboard.)
+
+This page could present the object's spec and its status fields in a way that
+the user can understand the state of the `ssh-access` rollout as described above
+with the `kubectl` example.
+
+After the user is satisfied with the CloudInit object they've created,
+they can reboot the Harvester node(s) it applies to so that the changes
+may take effect.
+
+
+#### Create CloudInit resource that conflicts with a protected file under `/oem` (API)
+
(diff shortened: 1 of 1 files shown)


enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,176 @@
...
+  3. Delete the CloudInit object and then remove the underlying file from `/oem`
+     on affected nodes and reboot the nodes at their leisure.
+
+* Automatic execution. Harvester users should expect to reboot a
+  Harvester node to allow the resulting cloud-init documents to be
+  applied at boot.
+
+## Proposal
+
+### User Stories
```
w13915984028, line 75: Please add the detailed user stories, it is more at black-box's view to describe the input and output; then the `test-plan`,Harvester document will be writen accordingly.

e.g:
(1) User can configure ssh file from CLI
step 1: create CR to contain certain file content
step 2: apply/save the CR
step 3: check status to know CR is/not rolled out
step 4: reboot node to make the change be effective
step 5: trouble-shooting if the result is not as expected

(2) User can configure ssh file from WebUI
...
(3) User can remove xxx file from CLI
(4) User can remove xxx file from WebUI

If white/black list is used, add related story

(5) Do we consider generating k8s event to say, some files are changed? POD logs may be lost due to POD restart, but events are stored

w13915984028 (commented):
Please check this question, thanks.

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,176 @@
+# Kubernetes-native Elemental cloud-init documents
+
+## Summary
+
+It is both an expected and supported behavior for Harvester users to
+persist certain configurations to their Harvester nodes. These are currently
+done with Elemental cloud-init documents under the `/oem` path on the system.
```
Vicente-Cheng, line 7: Not sure about the term `cloud-init documents`.
I checked the elemental document (https://rancher.github.io/elemental-toolkit/docs/reference/cloud_init/). They are called cloud-init style file. (Maybe we could also call cloud-init file?)
WDYT?

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Motivation
+
+Provide a GitOps-compatible way of persisting configuration changes to
+a Harvester cluster.
+
+### Goals
+
+* Elemental cloud-init documents are stored in a CloudInit object spec and
+  pulled down onto nodes that match the CloudInit's match selector.
```
Vicente-Cheng:
```suggestion
  pulled down onto nodes that match the selector of this CloudInit object.
```

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,176 @@
...
+There is a cloud-native tooling gap with regards to this process, however.
+Harvester users might prefer to use a GitOps-style approach to configuring
+their Harvester clusters, where Kubernetes objects are version-controlled
+as YAML and each new revision is applied to the cluster.
+
+This HEP presents a new Custom Resource Definition (CRD) that is surfaced
+via the harvester-node-manager, where a Harvester user could apply a "CloudInit"
+CR to a Harvester cluster, like so:
+
+```yaml
```
Vicente-Cheng, line 22: Could this CRD example contain the status part?
That would consist of our CloudInit object definition.

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,176 @@
...
+  pulled down onto nodes that match the CloudInit's match selector.
+* CloudInit object's status field describes how it is rolled out
+  across the cluster (for example: which nodes does it apply to,
+  are any out of sync, etc).
+
+### Non-goals [optional]
+
+* Automatic rollbacks. If a Harvester user wants to "undo" a previously applied
+  CloudInit object, they should expect to:
+
```
Vicente-Cheng, line 59: We need to remove the related files on the `/oem/` because it is the persistent path.
For the reboot timing, I thought that would depend on the user. We might remind the user but the reboot operation should be done by themself.

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,176 @@
...
+  1. Revise the CloudInit object contents and reboot; or
+  2. Adjust the CloudInit object match selector to exclude a node that they
+     no longer want it to apply to and remove the file under `/oem` and reboot; or
+  3. Delete the CloudInit object and then remove the underlying file from `/oem`
+     on affected nodes and reboot the nodes at their leisure.
+
```
Vicente-Cheng, line 64: Ditto for the reboot timing.

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,176 @@
...
+  that value as a SHA256 hash in an annotation on the CloudInit object
+  to easily check for equality between a file under `/oem` and its
+  corresponding CloudInit object contents.
+
+A new webhook is required to protect other files under `/oem`, such as:
+
+* elemental.config
+* grubenv
+* harvester.config
+* install
```
Vicente-Cheng, line 324: There are current files:
`90_custom.yaml`
`99_settings.yaml`
`elemental.config`
`grubenv`
`harvester.config`
`install`

The above files should not be affected by this feature.

Vicente-Cheng (commented):
some nits and open discussion, overall LGTM

enhancements/20231108-elemental-cloudinit-crd.md
ibrokethecloud: I agree.. we should remove the files on deletion of CRD. 

How actual effect is rolled back needs to be left to the user.

enhancements/20231108-elemental-cloudinit-crd.md
ibrokethecloud: events may not be the best place to record this info, since the default TTL is 1hour.

w13915984028: We may keep both POD log and event.

For production deployment, many customers have third-party logging server, which filters&stores logging/event from Harvester cluster, then store them for longer time.

connorkuehl: Is it sufficient for this information to surface as Conditions?

```
Status:
  Rollouts:
    Conditions:
      Last Transition Time:  2023-10-24T21:15:02Z
      Message:               
      Reason:                CloudInitApplicable
      Status:                True
      Type:                  Applicable
      Last Transition Time:  2023-10-24T21:09:28Z
      Message:               Local file checksum is the same as the CloudInit checksum
      Reason:                CloudInitChecksumMatch
      Status:                False
      Type:                  OutOfSync
      Last Transition Time:  2023-10-24T21:09:28Z
      Message:               99_ssh.yaml is present under /oem
      Reason:                CloudInitPresentOnDisk
      Status:                True
      Type:                  Present
    Node Name:               harv1
```

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: Yes, I will add commentary on the Status to the HEP. I just omitted it from this snippet here specifically because I just wanted to showcase the part of the object that the user must write.

enhancements/20231108-elemental-cloudinit-crd.md
w13915984028: When user update the same CRD object many times, the Conditions normally only reflect the last (few) changes.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: I did incorporate their language to some parts but not the entire document. I read them as more or less as the same meaning.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: Thanks! Added to latest revision.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: Thanks! I've updated the HEP to indicate that the controller will remove the files when the resource is deleted or when it no longer applies to the node it is on (user has updated `matchSelector` field)

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: Updated phrasing in next revision

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: Thanks! Added.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: Thanks! I've updated the user stories with some examples. I could not be exhaustive about everything a user could possibly do to a Harvester node since they are only limited by their imagination and the cloud-init syntax.

I've also added a note indicating that the controller will create a kubernetes event whenever it modifies a file under `/oem` and it will also indicate which node it happened on.

ibrokethecloud (approved):
lgtm. thanks.

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,369 @@
...
+    // MatchSelector is used to select which Nodes this CloudInit object
+    // applies to. For Kubernetes resources, an empty map means "match everything."
+    MatchSelector map[string]string `json:"matchSelector"`
+    // Filename is the file's base name under `/oem`. For example, `a.yaml` would
+    // result in a file appearing as `/oem/a.yaml`. Note that it is the _base name_.
+    // Subdirectories are not allowed.
+    //
+    //   Invalid: `/oem/asdf/a.yaml`
+    //     Valid: `/oem/a.yaml`
+    Filename      string            `json:"filename"`
```
w13915984028, line 258: Please consider add a field called `Paused`, the idea is borrowed from `Managedchart` and others:
When this field is set to true, the controller will stop syncing data to nodes temporary, it is useful for upgrading and debugging.

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,369 @@
...
+CR to a Harvester cluster, like so:
+
+```yaml
+apiVersion: node.harvesterhci.io/v1beta1
+kind: CloudInit
+metadata:
+  name: ssh-access
+spec:
+  matchSelector: {}
+  filename: 99_ssh.yaml
```
w13915984028, line 29: Please add a `note` that:
The target file will be replaced by the `CloudInit` object fully, selectively updating a block of the file is not supported.

w13915984028 (approved):
Looks very good, thanks.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: This is a great idea! I've added it to the newest revision.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: I added a note to the non-goals section, thank you

Vicente-Cheng (approved):
LGTM!

connorkuehl:
@torchiaf :wave: Would you be willing to look at the user stories on this HEP with the "UI" or "Dashboard" label in their headings? I'm working on merging the backend code this week and I was hoping to add better details to the frontend-specific issue[1] so that it is easier to execute on.

[1] https://github.com/harvester/harvester/issues/4711

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,376 @@
...
+`matchSelector` down the line, then the `Applicable` condition may help
+them troubleshoot why a node has or does not have a cloud-init-style file
+copied down to it.
+
+The user can now reboot the Harvester node so that their cloud-init-style
+file may be applied at boot time.
+
+#### Add/rotate rancher user authorized keys (UI/Dashboard)
+
+Alternatively, instead of creating the `ssh-access` resource via kubectl,
```
torchiaf, line 151: We have a page, 'SSH Keys', under Advanced tab. Is it related in some way with the new CloudInit functionality ? If yes, we may consider to put the new page along with the SSH key page.

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,376 @@
...
+
+After clicking "save/create/update", the dashboard could redirect the user to
+a more detailed page (much like the one you see for a virtual machine when clicking
+on a virtual machine's name on the "Virtual Machines" page of the dashboard.)
+
+This page could present the object's spec and its status fields in a way that
+the user can understand the state of the `ssh-access` rollout as described above
+with the `kubectl` example.
+
+After the user is satisfied with the CloudInit object they've created,
```
torchiaf, line 185: Should we consider to add a 'Reboot' button when a new CloudInit is created/updated ?

enhancements/20231108-elemental-cloudinit-crd.md
```diff
@@ -0,0 +1,376 @@
...
+
+As above, the user may find themselves in a similar scenario if they are trying
+to create a CloudInit resource with a protected filename or editing a pre-existing
+resource's filename to match that of a protected file under `/oem`.
+
+In this case, after clicking "save/update/create", the button may turn red and error
+text can indicate that they cannot name their CloudInit resource's filename field after
+a protected file that resides under `/oem`.
+
+Alternatively, the UI could offer a richer experience and disable the "save/update/create"
```
torchiaf, line 224: I would prefer this option, but we need to be sure that it is consistent with other validations pattern applied to other pages.

torchiaf:
@connorkuehl left some comments, feel free to discard not relevant ones, otherwise LGTM.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: No, it's a sad naming coincidence. This is for "Elemental cloud init" for customizing the host Harvester OS, but the cloud-init that many people are familiar with is for customizing a virtual machine as it boots up. The SSH Keys page is for storing SSH public keys that can be injected into the VM as part of the VM-based cloud-init and not the Elemental cloud-init.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: This could be cool but I don't think the juice is worth the squeeze since there are some special cases that might require a disproportional amount of effort to deliver this.

For example, if someone makes a change to the document's formatting or ordering of fields, but the end result's semantic meaning is identical, like this:

```yaml
a: foo
b: bar
```

```yaml
b: bar
a: foo
```

then we'd want to avoid the confusion of showing a reboot prompt on the dashboard as it's not necessary, but the file on disk would be updated due to the change regardless, so in this situation, a changed file is not always a good signal for whether a reboot is required.

Furthermore, I don't think the backend has node power management capabilities (as of yet) and that would probably be out of scope of the node-manager component that this HEP is targeting.

enhancements/20231108-elemental-cloudinit-crd.md
connorkuehl: I must admit I'm not familiar with the frontend's validation patterns.

WRT this specific example, if the user is editing a CloudInit object on the frontend and they change the object's .spec.filename, it is possible to list all cloudinits in the cluster and walk that list to ensure the filename is not yet taken.

connorkuehl:
@mergify backport v1.3

Issue #3266, opened 2022-12-09
[FEATURE] Support 2 Node / Witness Configuration


rebeccazzzz:
**Describe the solution you'd like**
Enable Harvester to be deployed on 2 Nodes with HA, or allow a Witness server to be a tie-breaker or etcd node.

**Is your feature request related to a problem? Please describe.**
We can only support/allow 3 nodes due to quorum and the Kubernetes architecture.

We would like a two-node configuration, to minimize footprint and keep costs low while having an additional node for HA.

**Describe alternatives you've considered**
Workaround:
1. During the installation, allow the node to work the role of master, but not a worker.
2. Allow configuration after the cluster is up is also needed if the user doesn't configure during installation or changes their mind.
3. Allow manual pick and promote, instead always promote automatically.

Vicente-Cheng:
This issue has the same purpose as #4840, so we can test it together.

albinsun:
Test plan refer to HEP
https://github.com/harvester/harvester/blob/v1.3.0-rc1/enhancements/20231113-witness-node-support.md#note-optional.

## Environment
* Harvester
  * Version: `v1.3.0-rc1`
  * Profile: QEMU/KVM, 3 nodes (8C/16G/500G)
    * Default: 8C / 16G / 500G
    * Witness: 4C / 8G / 500G
  * ui-source: `Auto`
* Rancher
  * Version: `v2.7.6`
  * Profile: rancher-vcluster


## Checks
1. Node role and behavior
   * Should be only one witness node.
   * When user config more than 1 witness node, it will be regard as Mangement.
1. Basic functions: Creat a VM with some data
   1. Reboot -> Data check
   9. Restore -> Data check
1. `rancher-vcluster` can be created and works as well.
1. Can adding 4th node
3. Can drop witness node


## Scenarios

> Please refer following comments for detail test steps 

### Positive
1. D -> D -> W :heavy_check_mark: 
   * #5029
     -> Known UI issue but not relate to witness
2. D -> W -> D :heavy_check_mark: 
   * #5029
     -> Known UI issue but not relate to witness 
3. D -> W -> W :heavy_check_mark: 
   * #5070
     -> Fixed
   * #5071
     -> Can NOT reproduce, should be env. issue.

### Negtive
1. <details>
   <summary> First node should NOT be Witness :heavy_check_mark: 
   </summary>
   
   Config witness and system setup as management.
[...]

PR #4719, opened 2023-11-13
HEP: witness node support

Vicente-Cheng:

**Problem:**
HEP for witness node support

**Solution:**
No need

**Related Issue:**

**Test plan:**
No need

### DIFF
--- a/enhancements/20231113-witness-node-support.md
+++ b/enhancements/20231113-witness-node-support.md
@@ -1,0 +1,176 @@
+# Support witness node
+
+## Summary
+
+The witness node is a lightweight node that only runs the etcd components. The witness node is not schedulable and does not run any workload. The witness node forms a quorum with the other two nodes.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/3266
+
+## Motivation
+
+Some users do not have three powerful nodes for the edge uses case. Instead, they would have two powerful nodes and one lightweight node. The lightweight node is only used for the `etcd` role. Kubernetes need at least three `etcd` nodes to form a quorum. The witness node support will help the edge uses case. That will benefit both cost and high availability.
+
+Reference:
+  - https://etcd.io/docs/v3.5/faq/#why-an-odd-number-of-cluster-members
+  - https://etcd.io/docs/v3.5/faq/#what-is-failure-tolerance
+
+### Goals
+
+Create a three-node cluster with one witness node.
+
+## Proposal
+
+### User Stories
+
+For the edge use case, the witness node support benefits both cost and high availability. The classical scenario would be two nodes and one witness node.
+
+### User Experience In Detail
+
+#### Installation
+- Install with interactive mode:
+   1. There is no change for the first node (`create` mode). The first node must be the master node.
+   2. For the second/third node, we will select the `join` mode to join the current cluster. Then, we will move to the role selection panel. On this panel, we can choose the role of `Default` or `Witness`. If we select the `Witness` role, the node will be promoted to only the etcd role, which will be the witness node.
+
+- Install with non-interactive mode:
+   1. Create with different `install.role` determining the node role (master/witness). Note: the first node will be the master node regardless of the value of `install.role`.
+
+#### Running workloads
+After we create the cluster with the witness node, we can run the workloads as usual. The only difference is that the witness node does not run any workload. Only the etcd related workload and some necessary workload (like harvester-node-manager) will be scheduled for the witness node.
+
+#### Specify the node role (Optional)
+The users can specify the node role when the cluster is running. The minimal requirement is to keep the three etcd roles to form a quorum, so we can try to promote the node to etcd or master role. And we can also delete the node if we can satisfy the minimal requirement.
+
+#### Upgrade
+The upgrade should work as usual.
+
+### API changes
+
+None
+
+## Design
+
+We need to document the following limitations and requirements for the witness node feature.
+
+### Limitations
+
+
+- The replica number of the default StorageClass is `3`, but in the two master nodes + one witness node scenario, the replica number should be `2`. Users could manually create a new StorageClass and set it as default or use the installation configuration to change the replica number of the default StorageClass [link](https://docs.harvesterhci.io/v1.2/install/harvester-configuration#installharvesterstorage_classreplica_count).
+
+- Only allow one witness node in the cluster.
+
+#### Requirements
+
+Only one witness node is allowed in the cluster.
+
+### Implementation Overview
+
+#### Promote controller
+
+Harvester depends on the job `promote` to promote a node to related role after it joins (the default role is worker). We can only promote the node to `etcd` role with the label `node-role.harvesterhci.io/witness=true`. Also, we need to add taint for etcd role to prevent workload from scheduling to the node.
+
+After this feature, the harvester node will have the following roles:
+1. master node roles:
+   - node-role.kubernetes.io/control-plane: "true"
+   - node-role.kubernetes.io/etcd: "true"
+   - node-role.kubernetes.io/master: "true"
+
+2. etcd node roles:
+   - node-role.kubernetes.io/etcd: "true"
+
+3. worker node roles:沒沒
+
+#### Taint and Toleration
+
+Promote controller will taint the witness node with the following taint:
+```
+taints:
+- key: key: node-role.kubernetes.io/etcd
+  effect: NoExecute
+  value: "true"
+```
+
+Add the following toleration to the promote job and node manager:
+```
+tolerations:
+- effect: NoExecute
+  operator: Exists
+```
+
+**OPTIONAL**: Promote controller can change the node role dynamically.
+
+#### Harvester WebHook
+
+Add some checking mechanism to avoid adding the control-plane label, `node-role.kubernetes.io/control-plane: "true"`, to the witness node by mistake.
+
+#### Installation
+
+1. Interactive mode:
+   - We will move to the role selection panel when we select the join mode. We can choose the role of `Default` or `Witness`. If we select the `Witness` role, the node will be promoted to only the etcd role, which will be the witness node.
+
+2. Non-interactive mode:
+   - Use the `install.role=witness` to set the witness node.
+     - supported `default`, `witness`
+
+   Example:
+   ```yaml
+   scheme_version: 1
+   server_url: <server_url>
+   token: <join token>
+   os:
+     hostname: <hostname>
+     password: <password>
+     ntp_servers: <ntp servers>
+   install:
+     mode: join
+     role: witness <-- this would make the node as etcd role (witness node)
+     management_interface:
+     interfaces:
+       - name: <management interface name>
+     method: dhcp
+     device: <target disk>
+     iso_url: <iso url>
+     tty: ttyS0
+   ```
+
+#### Miscellaneous changes
+
+- **Longhorn**: 
+    
+   The default StorageClass `harvester-longhorn` will have three replicas, but it could not be satisfied with the witness node. We need to update the default StorageClass to two replicas.
+
+- **Harvester-Node-Manager**:
+
+   The node manager needs to run on the whole cluster because it handles the NTP synced and other node-related jobs. We need to add some toleration to the node manager to ensure it can be scheduled to the witness node.
+
+- **sync-additional-ca**:
+
+  The `sync-additional-ca` job will sync the additional CA to the node. We need to add some toleration to the job to ensure it can be scheduled to the witness node.
+
+- **kubelet**:
+
+  Because the witness node will taint with the `node-role.kubernetes.io/etcd: "true"` when provision, we need to run kubelet with `--register-with-taints=node-role.kubernetes.io/etcd=true:NoExecute` argument to ensure the kubelet can work.
+
+### Test plan
+
+1. Create a three-node cluster through both interactive and non-interactive modes.
+2. Check the node roles after the cluster is created. (Two master nodes and one witness node)
+3. Verify that all basic functions are working as expected.
+4. Also, check the `rancher-vcluster` can be created and worked as well.
+
+### Upgrade strategy
+
+Upgrade should not be affected.
+
+## Note [optional]
+
+Web UI should consider supporting the pormote operation. The user can select the node and specify the role for this node.
+
+There is the resource usage for the witness node. (harvester-node-2 is the witness node)
+```
+NAME↑                   STATUS       ROLE                             VERSION                     PODS        CPU         MEM       %CPU        %MEM        CPU/A        MEM/A AGE          
+harvester-node-0        Ready        control-plane,etcd,master        v1.25.9+rke2r1                71       1244       10081         12          63        10000        15976 24h          
+harvester-node-1        Ready        control-plane,etcd,master        v1.25.9+rke2r1                28        628        5794          6          36        10000        15976 24h          
+harvester-node-2        Ready        etcd                             v1.25.9+rke2r1                 7        223        3025          2          18        10000        15976 24h          
+``` 


enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,71 @@
...
+
+3. worker node roles:
+
+#### Installation
+
+1. Interactive mode:
+   - We need to add an extra panel for setting the witness node (only for join cluster mode).
+
+2. Non-interactive mode:
+   - We can simply the label `node-role.harvesterhci.io/etcd=true` to the node by yaml.
```
bk201: How about adding a new mode in https://docs.harvesterhci.io/v1.2/install/harvester-configuration#installmode? It's more consistent.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Proposal
+
+### User Stories
+
+For the edge use case, the witness node support benefits both cost and high availability. The classical scenario would be two nodes and one witness node.
+
+### User Experience In Detail
+
+1. Set the node label `node-role.harvesterhci.io/etcd=true` to the node.
```
bk201: Since it's user experience, suggest describing the installer change. Either interactive or iPXE way.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: How about?
```
install:
  - mode:
  - role:
```
Create only support `master` role
Join can support `worker`, `etcd` and `master`

enhancements/20231113-witness-node-support.md
Vicente-Cheng: OK, I will update once we conclude the above discussion.

abonillabeeche:
Should the node also carry a longhorn service that helps determine quorum or run basic, non data related services?

Vicente-Cheng:
> Should the node also carry a longhorn service that helps determine quorum or run basic, non data related services?

longhorn-manager is the `daemonset`, it is not affected by taint.
So it will still deploy on the witness node.

enhancements/20231113-witness-node-support.md
bk201: How about `mode: witness`? just add another role there.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,72 @@
...
+## Summary
+
+The witness node is a lightweight node that only runs the control plane components. The witness node is not schedulable and does not run any workload. The witness node forms a quorum with the other two nodes.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/3266
+
+## Motivation
+
```
w13915984028, line 12: For a new reader, there are some open questions against this HEP:
(1) The motivation/background of such requirement; then it comes to this idea that such a `witness node` can be used to address certain issues
(2) When `witness node` mode is selected, could this node use less resource like <250g disk, <8 core CPU ? do we try to set a new resource limitation for this node?
(3) As Harvester is adding all container images from ISO to each NODE, both new installation and upgrade; then for `witness node`, it seems only system level images/specific images are required, do we set an image list for it? to reduce the resource consumption.
(4) Given a cluster with 2-control and 1-witness, what's the default replica-count of longhorn we suggest? the default `3` seems not fit for such scenario.
(5) Do we allow role change of witness node ? e.g. with fewer resource as witness to join, and then remove this role?
(6) Do we allow more than one witness node?
(7) The potential cross-affect of `rancher-vcluster`, the `vcluster` is (partialy) sharing control-plane with the origin k8s cluster, but with more magics.

Those points need to be clarified and defined clearly in the HEP.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,72 @@
...
+### Goals
+
+Create a three-node cluster with one witness node.
+
+## Proposal
+
+### User Stories
+
+For the edge use case, the witness node support benefits both cost and high availability. The classical scenario would be two nodes and one witness node.
```
w13915984028, line 27: For user stories, they need to be defined, more from user pespective, e.g.
(1)  User can install new witness node
How is this node installed, resource requrirements;  detailed steps
(2)  User can remove witness node
How is this node removed,
(3) User can/cannot change witness node role
how/why
...

w13915984028 (commented):
From the view of a full HEP document, a few questions are listed, thanks.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: The `mode` is more like an operation because we use `create`/`join`.
So, I prefer to add `role` for the specific node role.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: Thanks @w13915984028,  for the HEP suggestion.

I updated the latest HEP for the above questions. Please help to recheck it!

enhancements/20231113-witness-node-support.md
Vicente-Cheng: I will update the user story once the installation steps are clear.

enhancements/20231113-witness-node-support.md
ibrokethecloud: This needs to be handled by the installer. Could be as simple as adding a new panel which in turn populates the additional node labels.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,93 @@
...
+1. Create a three-node cluster through both interactive and non-interactive modes.
+2. Check the node roles after the cluster is created. (Two master nodes and one etcd node)
+3. Verify that all basic functions are working as expected.
+4. Also, check the `rancher-vcluster` can be created and worked as well.
+
+### Upgrade strategy
+
+Upgrade should not be affected.
+
+## Note [optional]
```
ibrokethecloud, line 166: rancher currently watches resources and tries to scale deployments as nodes become available in the cluster using the `management.cattle.io/scale-available: "3"` annotation. We need to ensure when using a witness node is present in the cluster, the annotation is updated on harvester resources.

Vicente-Cheng: Yes, I just found that annotation and its tricks yesterday!
Thanks for the reminder!!

Also, we need to handle the replicas of harvester/harvester-webhook on the `managedcharts`.

w13915984028: BTW: 
Not all Harvester clusters are managed by Rancher.

For any cluster with `M Normal Node + N Witness Node` (suppose we allow such flexible setup?), we need to give detailed guide about the suggested replica count in Harvester document.  and utilize the configuration file https://docs.harvesterhci.io/v1.2/install/harvester-configuration/#installharvesterstorage_classreplica_count to set a proper value when the first node is installed.

Both Rancher-managed or none Rancher-managed will be in documents for reference.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,93 @@
...
+
+### API changes
+
+None
+
+## Design
+
+### Limitations
+
+- Each cluster only supports one witness node.
```
w13915984028: Seems in installation stage, it is not possbile/easy to check if there has been one witness node, to prevent other `witness node` from joining,  e.g. mis-behaviour to set `witness node` to a normal  node. Then let's add warning in the document , that this limitation should be met by user manually.

enhancements/20231113-witness-node-support.md
bk201: @w13915984028 We are discussing deployments replica count here, not the Longhorn replica count. Please refer to https://github.com/harvester/harvester/blob/b7350b147ceb919195f260214e6a80870481e890/deploy/charts/harvester/templates/deployment.yaml#L9C23-L9C23 for more information.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Implementation Overview
+
+#### Promote controller
+
+Harvester depends on the job `promote` to promote a node to related role after it joins (the default role is worker). We can only promote the node to `etcd` role with the label `node-role.harvesterhci.io/etcd=true`. Also, we need to add taint for etcd role to prevent workload from scheduling to the node.
```
w13915984028: Before the `promote` is finished, could some workload be scheduled to this `witness-node`? As it joins the cluster as `worker` from beginning, k8s may schedule any none-control related workload to this node at any time.

Namely, the `taint` needs to be done at very early stage.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+After this feature, the harvester node will have the following roles:
+1. master node roles:
+   - node-role.kubernetes.io/control-plane: "true"
+   - node-role.kubernetes.io/etcd: "true"
+   - node-role.kubernetes.io/master: "true"
+
+2. etcd node roles:
+   - node-role.kubernetes.io/etcd: "true"
```
w13915984028, line 80: Any way to prevent an role like `node-role.kubernetes.io/control-plane: "true"` from been added to `witness-node` (manually) ?

Is `witness` saved in some places in the NODE file? or any other CRD objects to get an additional check.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,93 @@
...
+#### Miscellaneous changes
+
+- **Longhorn**: 
+    
+   The default StorageClass `harvester-longhorn` will have three replicas, but it could not be satisfied with the witness node. We need to update the default StorageClass to two replicas.
+
+### Test plan
+
+1. Create a three-node cluster through both interactive and non-interactive modes.
+2. Check the node roles after the cluster is created. (Two master nodes and one etcd node)
```
w13915984028: Question:
When there is a two-node cluster with 1 master + 1 witness,  e.g., the third-node is not installed yet, or gone, or removed, anything needs to be cared?

w13915984028 (commented):
Some additional questions, thanks.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: Sure, I will also add some warning on the installer.
Thanks for the reminder.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: Yap, that's a good point.
In fact, as a worker node, we indeed deploy a few workloads on it after installation (like rancher/harvester pod)
We will set taint before promoting the operation, which will delete the unneeded pod on this node.

I do not think that will have too much impact on the above workloads, but if needed, we could try to add taint at a very early stage.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: Sure, we need to add some checking mechanisms to prevent this.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: Promote will be triggered on at least three nodes so we do not encounter this situation.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: After discussion, I thought we might also allow the user to choose the etcd nodes with join mode. (Means we do not limit only one etcd node)
Just satisfy the k8s quorum requirement, and we should let the user promote the etcd node to master node.

Vicente-Cheng:
Hi folks, I update the latest HEP.
Please help to recheck it. Thanks!

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,166 @@
...
+
+The witness node is a lightweight node that only runs the control plane components. The witness node is not schedulable and does not run any workload. The witness node forms a quorum with the other two nodes.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/3266
+
+## Motivation
+
+Some users do not have three powerful nodes for the edge uses case. Instead, they would have two powerful nodes and one lightweight node. The lightweight node is only used for the `etcd` role. Kubernetes need at least three nodes to form a quorum. The witness node support will help the edge uses case. That will benefit both cost and high availability.
```
w13915984028: please add a link to 
https://etcd.io/docs/v3.3/faq/#why-an-odd-number-of-cluster-members
or
https://etcd.io/docs/v3.3/faq/#what-is-failure-tolerance

it explains a minimum 3-etcd-node can tolarent 1 node failure, thus a basic HA is built.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,166 @@
...
+### User Stories
+
+For the edge use case, the witness node support benefits both cost and high availability. The classical scenario would be two nodes and one witness node.
+
+### User Experience In Detail
+
+#### Installation
+- Install with interactive mode:
+   1. There is no change for the first node (`create` mode). The first node must be the master node.
+   2. For the second/third node, we will select the `join` mode to join the current cluster. Then, we will move to the role selection panel. On this panel, we can choose the role of `Default` or `Etcd`. If we select the `Etcd` role, the node will be promoted to only the etcd role, which will be the witness node.
```
w13915984028: `Etcd` or `Etcd-only` ? the `Default` role includes `Etcd` implicitly.

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,166 @@
...
+
+- Install with non-interactive mode:
+   1. Create with different `install.role` determining the node role (master/etcd). Note: the first node will be the master node regardless of the `install.role` is.
```
w13915984028: the `install.role` is 
->
the value of `install.role`

enhancements/20231113-witness-node-support.md
```diff
@@ -0,0 +1,166 @@
...
+### Implementation Overview
+
+#### Promote controller
+
+Harvester depends on the job `promote` to promote a node to related role after it joins (the default role is worker). We can only promote the node to `etcd` role with the label `node-role.harvesterhci.io/etcd=true`. Also, we need to add taint for etcd role to prevent workload from scheduling to the node.
+
+After this feature, the harvester node will have the following roles:
+1. master node roles:
+   - node-role.kubernetes.io/control-plane: "true"
+   - node-role.kubernetes.io/etcd: "true"
```
w13915984028, line 76: again, `etcd` or `etcd-only` ?

w13915984028 (approved):
Looks very good, thanks.

A few remaining questions.

bk201 (approved):

enhancements/20231113-witness-node-support.md
ibrokethecloud: @Vicente-Cheng should we update the role to `Witness`?

ibrokethecloud: rest looks good and i can approve this once this is updated.

bk201: @w13915984028 We'll change the role to witness. Witness node is a commonly used term in the industry.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: Updated! Thanks for the suggestion!

enhancements/20231113-witness-node-support.md
Vicente-Cheng: yes, we use the witness node for the term

enhancements/20231113-witness-node-support.md
Vicente-Cheng: done

enhancements/20231113-witness-node-support.md
Vicente-Cheng: This key is not the custom-defined key. So keep it.

enhancements/20231113-witness-node-support.md
Vicente-Cheng: Updated!

ibrokethecloud (approved):
lgtm. thanks.

Issue #4334, opened 2023-07-31
[BUG] Harvester 1.1.2 OS stuck on boot with BOSS-S1 hardware RAID


bk201:
**Describe the bug**

The bootup procedure stuck on the "Welcome to GRUB!" screen for 30+ minutes. The node has a BOSS-S1 hardware RAID-card.
 
**To Reproduce**
1. Install Harvester v1.1.2.
2. Reboot
3. The OS hangs on "Welcome to GRUB!" for 30+ minutes.

**Expected behavior**

The OS should boot immediately.

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

- Related issue: https://github.com/harvester/harvester/issues/2115
- It's caused by a new search in the `/etc/cos/grub.cfg` file:
  ```
  # Load custom config file
  set custom="/grubcustom"
  search --no-floppy --file --set=custom_blk "${custom}"
  ```

bk201:
The workaround is:
- If the system manages to boot into the OS: After booting, create a grubcustom file and try reboot
  ```
  grub2-editenv /oem/grubcustom create
  ```
- If it takes too long to boot, boot the live ISO and mount the OEM partition to create the file.
  ```
  mkdir /tmp/oem
  mount -L COS_OEM /tmp/oem
  grub2-editenv /tmp/oem/grubcustom create
  ls -l /tmp/oem/grubcustom
  umount /tmp/oem
  ```

bk201:
The search is removed in 1.2.0, but we need to confirm again.

masteryyh:
Tested with Dell R6515 with a BOSS-S1 RAID card and v1.2.0-rc5 seems not stuck on boot menu.

bk201:
@masteryyh Thanks. Move the issue to v1.1.3.

bk201:
@masteryyh I think we also need to fix the upgrade path (1.1.2 -> 1.1.3). Like: https://github.com/harvester/harvester/blob/f77ef49db2cea24d934ef0a6db957e225d2b6f46/package/upgrade/upgrade_node.sh#L576-L582

TachunLin:
According to the above comment, move the issue to implement for investigation.

TachunLin:
Verifying **fixed** on `v1.1-340486a3-head` (11/20). Close this issue

### Result
[...]

PR #4720, opened 2023-11-13
upgrade: add `/oem/grubcustom` to prevent upgrade stuck

Vicente-Cheng:

**Problem:**
Prevent getting stuck on file searching when upgrading.

**Solution:**
Create an empty file for searching

**Related Issue:**
https://github.com/harvester/harvester/issues/4334

**Test plan:**
Test upgrade with specific hardware.

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -524,14 +524,21 @@ upgrade_os() {
   # stuck on the grub file search for about 30 minutes, this can be
   # mitigated by adding the `grubenv` file.
   #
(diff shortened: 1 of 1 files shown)


Vicente-Cheng:
Some context from https://github.com/harvester/harvester/issues/4334#issuecomment-1722888352

We do not need this trick on v1.2, so I directly create PR for v1.1

starbops (approved):
Tested on 2-node (VM) upgrade from v1.1.2 to the custom-built image containing the fix. The `/oem/grubcustom` file were generated. LGTM, thanks!

bk201 (approved):
lgtm, thanks!

PR #4747, opened 2023-11-21
bump images for 1.1.3rc1

markhillgit:
bump images for 1.1.3rc1

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -491,5 +491,5 @@ harvester-node-manager:
   image:
     imagePullPolicy: IfNotPresent
     repository: rancher/harvester-node-manager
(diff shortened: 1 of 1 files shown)


Vicente-Cheng:
Hi @markhillgit,
I thought we should also bump `node-manager` version here https://github.com/harvester/harvester/blob/v1.1/deploy/charts/harvester/values.yaml#L494

markhillgit:
> Hi @markhillgit, I thought we should also bump `node-manager` version here https://github.com/harvester/harvester/blob/v1.1/deploy/charts/harvester/values.yaml#L494

thank you - yes!

bk201 (approved):
LGTM, thanks!

ibrokethecloud (approved):
LGTM. thanks.

PR #4781, opened 2023-11-28
promote: make promote script can only promote the node to etcd role

Vicente-Cheng:

**Problem:**
we need to support witness node for 2+1 scenario (2 master node + 1 witness node)

**Solution:**
make promote controller can only promote the node to the `etcd`

**Related Issue:**

**Test plan:**
- How to build the ISO
We need this PR and https://github.com/harvester/harvester-installer/pull/604
You can change the `script/build-iso` to assign harvester-installer to the above PR
Then, build the ISO from the `harvester/harvester` as usual.

- Test scope
  - Installation: use both interactive and non-interactive mode
    - interactive mode: create 3 node cluster and select the etcd role with the third node
    - non-interactive mode: create 3 node cluster and use `install.role=etcd` with the third node
  - Cluster operations:
    - Whole basic operations need to verify and make sure there is no any workload be scheduled to the etcd node
    - Create downstream cluster and verify related operations. Also need to make sure there is no any workload be scheduled to the etcd node
    
  - Workloads on the etcd node
  The etcd nodes should only have the followings pods on it.
  ```
  harvester-system            harvester-node-manager-mjnjx                             1/1     Running     
  kube-system                 cloud-controller-manager-harvester-node-2                1/1     Running     
  kube-system                 etcd-harvester-node-2                                    1/1     Running     
  kube-system                 kube-proxy-harvester-node-2                              1/1     Running     
  kube-system                 rke2-canal-dxrbh                                         2/2     Running     
  kube-system                 rke2-multus-ds-zv98d                                     1/1     Running     
  ```

### DIFF
--- a/deploy/charts/harvester/templates/configmap.yaml
+++ b/deploy/charts/harvester/templates/configmap.yaml
@@ -20,6 +20,21 @@ metadata:
 data:
   promote.sh: |-
     {{`KUBECTL="/host/$(readlink /host/var/lib/rancher/rke2/bin)/kubectl"
+    ROLE_LABELS="rke.cattle.io/control-plane-role=true rke.cattle.io/etcd-role=true"
+    ETCD_ONLY=false
+    if [[ -n "$1" ]]; then
+      ETCD_ONLY=true
+      ROLE_LABELS=$1
+    fi
+
+    case $ROLE_LABELS in
+      "rke.cattle.io/control-plane-role=true rke.cattle.io/etcd-role=true" | "rke.cattle.io/etcd-role=true")
+    ;;
+      *)
+      echo "ROLE $ROLE_LABELS is not supported."
+      exit 1
+    ;;
+    esac
 
     get_machine_from_node() {
       $KUBECTL get node $HARVESTER_PROMOTE_NODE_NAME -o jsonpath='{.metadata.annotations.cluster\.x-k8s\.io/machine}'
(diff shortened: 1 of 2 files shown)


ibrokethecloud:
as discussed, we just need to make minor changes to the promote contorller to factor in the new etcd only node as a management node.

ibrokethecloud (approved):
LGTM. thanks.

bk201 (approved):
LGTM

deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -79,17 +94,29 @@ data:
...
 
     while true
     do
-      CONTROL_PLANE=$($KUBECTL get node $HOSTNAME -o go-template=$'{{index .metadata.labels "node-role.kubernetes.io/control-plane"}}\n' || true)
+      if [[ $ETCD_ONLY == true ]]; then
+        ETCD_STATE=$($KUBECTL get node $HOSTNAME -o go-template=$'{{index .metadata.labels "node-role.kubernetes.io/etcd"}}\n' || true)
+
+        if [ "$ETCD_STATE" = "true" ]; then
+          $KUBECTL taint nodes $HOSTNAME node-role.kubernetes.io/etcd=true:NoExecute
+          $KUBECTL patch managedchart harvester -n fleet-local --type=json -p='[{"op":"replace", "path":"/spec/values/replicas", "value": 2}]'
```
w13915984028, line 108: a question about following scenario:

When a `2 + 1` cluster runs some days, and the user removes the witness node, then add a new control-plane node, do we have chance to patch the managedchart back?

we may add this to limitations now

deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -20,6 +20,21 @@ metadata:
 data:
   promote.sh: |-
     {{`KUBECTL="/host/$(readlink /host/var/lib/rancher/rke2/bin)/kubectl"
+    ROLE_LABELS="rke.cattle.io/control-plane-role=true rke.cattle.io/etcd-role=true"
+    ETCD_ONLY=false
+    if [[ -n "$1" ]]; then
```
w13915984028, line 25: The code works now, but has chance to break:

if $1 == "rke.cattle.io/control-plane-role=true rke.cattle.io/etcd-role=true", the code will also assume `ETCD_ONLY`;  sure, the controller side does set such value at the moment.

Set `ETCD_ONLY` in following `case` statement will be better, uses exact match.

w13915984028 (commented):
Questions about not mentioned scenarios, thanks.

deploy/charts/harvester/templates/configmap.yaml
Vicente-Cheng: Hi @w13915984028,
Thanks for the nice reminder! You are correct about it.
I will fix it!

deploy/charts/harvester/templates/configmap.yaml
Vicente-Cheng: For now, we need to patch it manually, but maybe we could take a look to see if the promote controller can do it.

Issue #4758, opened 2023-11-24
[ENHANCEMENT] Bump KubeVirt to 1.1.0-150500.8.6.1


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
SLE 15 SP4 general ends on 31 Dec 2023. KubeVirt base image will change. We have to update to a new version.

The supported version for SLE 15 SP5 will be 1.0.0: registry.suse.com/suse/sles/15.5/virt-operator:1.0.0-150500.8.3.1

**Additional context**
https://www.suse.com/lifecycle/#suse-linux-enterprise-server-15

TachunLin:
### Test Plan
1. Check can upgrade from v1.2.1 to v1.2-head until v1.3.0-rc2 can work as expected 
2. Check on fresh install and upgrade path to v1.3.0 Harvester cluster contains the bumped KubeVirt image to
    - `registry.suse.com/suse/sles/15.5/virt-operator:1.1.0-150500.8.6.1`
3. Check the basic VM operation can work 
4. Check the basic Volume operation can work

TachunLin:
Verifying on `v1.3.0-rc1` (24/01/25).

### Result

<details><summary>$\color{Grey}{\textsf{Pending}}$ Check can  **upgrade** from v1.2.1 to v1.2-head until v1.3.0-rc2 $~~$</summary>

 * Wait until v1.3.0-rc2 released 



 </details>

<details><summary>$\color{Grey}{\textsf{Pending}}$ KubeVirt bumped to 1.1.0-150500.8.6.1 after **upgrade** from v1.2.1 to v1.2-head until v1.3.0-rc2 $~~$</summary>

 * Wait until v1.3.0-rc2 released 



 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ KubeVirt bumped to 1.1.0-150500.8.6.1 after **fresh install** Harvester cluster $~~$</summary>

1. In fresh install Harvester v1.3.0-rc1, Kubevirt already bumped to `1.1.0-150500.8.6.1`

```
node1:~ # ctr -n k8s.io image ls | grep virt-operator
registry.suse.com/suse/sles/15.5/virt-operator:1.1.0-150500.8.6.1                                      application/vnd.docker.distribution.manifest.v2+json sha256:6a1b32dd2e1732c587c15f3673ea97317a6210735425d50072f1f146eed6a398 181.1 MiB linux/amd64 io.cri-containerd.image=managed
```

 </details>

[...]

PR #4787, opened 2023-11-28
feat: bump KubeVirt to 1.0.0-150500.8.3.1

FrankYang0529:
SLE 15 SP4 general ends on 31 Dec 2023. KubeVirt base image will change. We have to update to a new version.

The supported version for SLE 15 SP5 will be 1.0.0: registry.suse.com/suse/sles/15.5/virt-operator:1.0.0-150500.8.3.1

**Related Issue:**
https://github.com/harvester/harvester/issues/4758

**Test plan:**

1. Test create/backup/snapshot/restore can work.
2. Test upgrade.


FrankYang0529:
@bk201 I can create a new cluster and do some basic operations without error. However, I'm not sure why UI always show `VirtualMachine generation is 2, but latest observed generation is 1`. Could you help to take a look? Thank you.

Screenshot 2023-11-28 at 5 09 30 PM

bk201:
@torchiaf We are bumping Kubevirt dependency, can you help check what's the condition to show those text? Thanks!

> @bk201 I can create a new cluster and do some basic operations without error. However, I'm not sure why UI always show `VirtualMachine generation is 2, but latest observed generation is 1`. Could you help to take a look? Thank you.
>

go.mod
```diff
@@ -7,85 +7,75 @@ replace (
...
+	github.com/gobuffalo/flect v1.0.2
 	github.com/gorilla/mux v1.8.0
 	github.com/guonaihong/gout v0.1.3
 	github.com/harvester/harvester-network-controller v0.3.1
 	github.com/harvester/node-manager v0.1.5-0.20230614075852-de2da3ef3aca
 	github.com/iancoleman/strcase v0.2.0
 	github.com/k3s-io/helm-controller v0.11.7
-	github.com/k8snetworkplumbingwg/network-attachment-definition-client v0.0.0-20200331171230-d50e42f2b669
+	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.3.0
+	github.com/kube-logging/logging-operator/pkg/sdk v0.9.1
```
FrankYang0529, line 66: @starbops It looks like `banzaicloud/logging-operator` has been changed to `kube-logging/logging-oeprator`. Could you check whether it's safe to do this change in `go.mod`? Thank you.

There is a function `loggingWorkaround`. Not sure whether we need do any update for it?

https://github.com/harvester/harvester/blob/acb6e69121a9843c8e7616de9f44401293c4d238/pkg/codegen/main.go#L200-L214

FrankYang0529:
Upgrading from `v1.2.1` to this branch also works in my cluster.


Screenshot 2023-11-29 at 3 49 33 PM

go.mod
starbops: The GroupVersion has not changed, so keeping it untouched is okay. Thank you.
https://github.com/kube-logging/logging-operator/blob/bc3a4727c2dee2a362e85277a76ccff061e5e804/pkg/sdk/logging/api/v1beta1/groupversion_info.go#L27

torchiaf:
> @torchiaf We are bumping Kubevirt dependency, can you help check what's the condition to show those text? Thanks!
> 
> > @bk201 I can create a new cluster and do some basic operations without error. However, I'm not sure why UI always show `VirtualMachine generation is 2, but latest observed generation is 1`. Could you help to take a look? Thank you.

I'm checking.

torchiaf:
@FrankYang0529 `VirtualMachine generation is 2, but latest observed generation is 1` is the state description of the Object in the Resources table, in this case, the Virtual Machine.
The source is the field `metadata.state.message` from the response of `/v1/harvester/kubevirt.io.virtualmachines` endpoint.
Here is an example, (`clusters` in this case):

image

bk201:
Rebased and remove non-changed charts in `deploy/charts/harvester/charts`

ibrokethecloud:
the UI error is being caused due to the additional fields `desiredGeneration` and `obeservedGeneration` in new version of kubevirt. 

image


https://github.com/kubevirt/api/blob/v1.0.0/core/v1/types.go#L1527-L1537

Unless we decide to ignore it in the dashboard, this can lead to a bad UX with each VM reporting this error.

ibrokethecloud:
Restarting the VM fixes the UI error.

The issue is being caused by vmi controller, which waits for VMI to be running, and then copies the interface mac addresses into the parent VM spec.

https://github.com/harvester/harvester/blob/master/pkg/controller/master/virtualmachine/vmi_network_controller.go#L19

Error will re-appear after a restart if an additional interface is added to the VM.

image

ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):
Thanks!

FrankYang0529:
@Mergify backport v1.2

Issue #4778, opened 2023-11-27
[BUG] Resizing Extra Volumes Fail if VM was shutdown via SSH `sudo shutdown -h now` (w/ 422: `admission webhook denied the request: please stop the VM before resizing volumes`)


irishgordo:
**Describe the bug**
Resizing an extra volume for the VM fails if the VM was shutdown by a user, via SSH with something like `sudo shutdown -h now`.

**To Reproduce**
1. Build a VM (I used Ubuntu based distro), in that VM add an extra disk (type disk)
2. Build a partition table with a single partition and mount that extra disk at somewhere like /mnt/extra, as if the disk were to be used by the os
3. Shutdown the VM with `sudo shutdown -h now` 
4. In the VM Index page on Harvester, you'll see the VM is off
5. Edit the config of the VM
6. Increase the extra-volume size
7. Attempt to save changes, you'll see that the admission webhook posts the error, `denied the request: please stop the VM before resizing volumes` - w/ 422

**Workaround**:
1. Start the VM from the UI
2. When the VM is running, just shut down the VM from the UI w/ "Stop"
3. Then you can resize the extra-volume successfully

**Expected behavior**
That it would be possible to resize an extra disk if the VM was shutdown via SSH command, like the UI "Stop" command.

**Environment**
 - Harvester ISO version: v1.2.1
 - Underlying Infrastructure: bare-metal 2 node

**Additional context**

![Screenshot from 2023-11-26 22-48-26](https://github.com/harvester/harvester/assets/5370752/6b007a55-1f7a-48d5-a02c-f99b5400802f)

Request:
```
[...]

PR #4794, opened 2023-11-29
fix: add one more condition to check vm is stopped or not

Yu-Jack:
**Problem:**
There is a difference between `Stopping vm inside vm` and `Stopping vm from GUI`
- `Stopping vm inside vm` remains the VMI and virt-launcher, and `VM.Spec.RunStrategy` is not changed.
- `Stopping vm from GUI` deletes the VMI and virt-launcher, and `VM.Spec.RunStrategy` is changed to `Halt`.

The webhook validates VM is stopped or not by `VM.Spec.RunStrategy: Halt` only.

**Solution:**
After discussing with @FrankYang0529 , I think we should add one more case to pass webhook validator. It would be better not to change resource status because it's existing kubevirt issue here https://github.com/harvester/harvester/issues/2265#issuecomment-1125758792 until fixed.

**Related Issue:**
https://github.com/harvester/harvester/issues/4778

**Test plan:**

These two test cases are already tested successfully.

### Stop inside VM case
1. Create VM in harvester UI and add one more disk like following example. Here I use alpinelinux to boot VM.
image
2. Start VM
3. Login to node and run `lsblk` to check additional disk size is 3G.
4. Run `poweroff` command to stop inside VM instead of GUI.
5. Resize additional disk size to 4G
6. Start VM
7. Login to node and run `lsblk` to check additional disk size is 4G or not.

### Stop from GUI case
1. Create VM in harvester UI and add one more disk like following example. Here I use alpinelinux to boot VM.
image
2. Start VM
3. Login to node and run `lsblk` to check additional disk size is 3G.
4. Stop VM from GUI
5. Resize additional disk size to 4G
6. Start VM
7. Login to node and run `lsblk` to check additional disk size is 4G.

### DIFF
--- a/pkg/util/virtualmachine/virtualmachine.go
+++ b/pkg/util/virtualmachine/virtualmachine.go
@@ -1,0 +1,52 @@
+package virtualmachine
+
+import (
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+)
+
+// IsVMStopped checks VM is stopped or not. It will check two cases
+// 1. VM is stopped from GUI
+// 2. VM is stopped inside VM
+// These two cases are all stopped case.
+func IsVMStopped(
+	vm *kubevirtv1.VirtualMachine,
+	vmiCache ctlkubevirtv1.VirtualMachineInstanceCache,
+) (bool, error) {
+	strategy, err := vm.RunStrategy()
+	if err != nil {
+		return false, fmt.Errorf("error getting run strategy for vm %s in namespace %s: %v", vm.Name, vm.Namespace, err)
+	}
+
+	// VM is stopped from GUI
+	if strategy == kubevirtv1.RunStrategyHalted {
+		return true, nil
+	}
+
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusStopped {
+		return false, nil
+	}
+
+	// When vm is stopped inside VM, the vmi will not be deleted.
+	// The status of vmi will be "Succeeded".
+	vmi, err := vmiCache.Get(vm.Namespace, vm.Name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logrus.Debugf("VM %s/%s is deleted", vm.Namespace, vm.Name)
+			return true, nil
+		}
+		return false, fmt.Errorf("error getting vmi %s in namespace %s: %v", vm.Name, vm.Namespace, err)
+	}
+
+	if vmi.IsFinal() {
+		logrus.Debugf("VM %s/%s is stopped inside VM", vm.Namespace, vm.Name)
+		return true, nil
+	}
+
+	return false, nil
+}
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -199,7 +199,7 @@ func (v *vmValidator) checkResizeVolumes(oldVM, newVM *kubevirtv1.VirtualMachine
 	if isResizeVolume {
 		stopped, err := vmUtil.IsVMStopped(newVM, v.vmiCache)
 		if err != nil {
-			return werror.NewInternalError(fmt.Sprintf("failed to get run strategy, err: %s", err.Error()))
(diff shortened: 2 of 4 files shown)


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -193,14 +197,16 @@ func (v *vmValidator) checkResizeVolumes(oldVM, newVM *kubevirtv1.VirtualMachine
 	}
 
 	if isResizeVolume {
-		runStrategy, err := newVM.RunStrategy()
+		stopped, err := vmUtil.IsVMStopped(newVM, v.vmiCache)
 		if err != nil {
 			return werror.NewInternalError(fmt.Sprintf("failed to get run strategy, err: %s", err.Error()))
```
FrankYang0529: Need to change error message here.

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return false, fmt.Errorf("error getting vmi %s in namespace %s: %v", vm.Name, vm.Namespace, err)
+	}
+
+	if vmi == nil {
+		logrus.Debugf("VM %s/%s is deleted", vm.Namespace, vm.Name)
+		return true, nil
+	}
+
+	if vmi.IsFinal() {
```
FrankYang0529, line 46: @Yu-Jack, `IsFinal` also checks whether VMI is `Failed`. Do you know when will VMI gets `Failed` and whether KubeVirt controller restarts it? Thanks.

Ref: https://github.com/kubevirt/kubevirt/blob/1b055aa06d0e899de1657143b27fe4199e7e16dc/staging/src/kubevirt.io/api/core/v1/types.go#L448

FrankYang0529 (changes requested):
Overall LGTM. Leave some minor suggestions.

pkg/util/virtualmachine/virtualmachine.go
Yu-Jack: @FrankYang0529  I'm not sure what I'm looking for is right place, but it seems KubeVirt doesn't support to start failed state vm. 
- virt-api
  - https://github.com/kubevirt/kubevirt/blob/v0.54.0/pkg/virt-api/rest/subresource.go#L535-L538
- virt-controller
  - https://github.com/kubevirt/kubevirt/blob/v0.54.0/pkg/virt-controller/watch/vmi.go#L556-L574
  - https://github.com/kubevirt/kubevirt/blob/v0.54.0/pkg/virt-controller/watch/vmi.go#L986-L992

In general, it says `Failed mean the vmi crashed, disappeared unexpectedly or got deleted from the cluster before it was ever started`. For one example, the vmi is created, but is not still scheduled. If it's deleted, it will become Failed state in https://github.com/kubevirt/kubevirt/blob/v0.54.0/pkg/virt-controller/watch/vmi.go#L435.

FrankYang0529: I found that we can make VMI as Failed if we update `kubevirt.io/heartbeat` annotation  to an error format in Node. In this case, VM is running but VMI is failed. However, this is an edge case. I think we can just skip it. I am good with `IsFinal` here. Thanks.

Ref: https://github.com/kubevirt/kubevirt/blob/e8e638edc22587ec7be2cc3d983b61763e33f973/pkg/virt-controller/watch/node.go#L186-L203

FrankYang0529 (approved):
LGTM. Thanks.

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	// When vm is stopped inside VM, the vmi will not be deleted.
+	// The status of vmi will be "Succeeded".
+	vmi, err := vmiCache.Get(vm.Namespace, vm.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return false, fmt.Errorf("error getting vmi %s in namespace %s: %v", vm.Name, vm.Namespace, err)
+	}
+
+	if vmi == nil {
```
bk201: I'd suggest we don't rely on checking vmi, but use the NotFound error as the condition. We can combine the statement to line 38.

Like
```
if err != nil {
        logrus.Debugf("VM %s/%s is deleted", vm.Namespace, vm.Name)
        return return true, nil
    }
    return false, fmt.Errorf("error getting vmi %s in namespace %s: %v", vm.Name, vm.Namespace, err)
}
```

bk201 (approved):
LGTM, thank you!

Yu-Jack:
@Mergifyio backport v1.2

Issue #3588, opened 2023-03-03
[ENHANCEMENT] Keep whether VM is graceful termination


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
The final event on VMI is different between graceful and non-graceful termination. We would like to find a way to keep it as debug information. If there is data corruption, we can know whether non-graceful termination is the root cause.

**Non-graceful termination**
```
0s          Warning   Stopped                  virtualmachineinstance/vm30                            The VirtualMachineInstance crashed.
```

**Graceful termination**
```
0s          Normal    Stopped                  virtualmachineinstance/vm30                            The VirtualMachineInstance was shut down
```

**Describe the solution you'd like**
Since the event is on VMI and VMI will be deleted if we stop a VM, we can try to cache the event and keep it on VM annotations.

**Additional context**
https://github.com/harvester/harvester/issues/2466

Yu-Jack:
After I tested two cases

- Stop VM from UI
```
default     68s         Normal   SuccessfulCreate   virtualmachine/testvm           Started the virtual machine by creating the new virtual machine instance testvm
default     6m42s       Normal   SuccessfulDelete   virtualmachine/testvm           Stopped the virtual machine by deleting the virtual machine instance 2fcf9d2c-44ff-43a6-8a8c-ab1152c4acd7
default     6m42s       Normal   SuccessfulDelete   virtualmachineinstance/testvm   Deleted virtual machine pod virt-launcher-testvm-hvwdl
default     6m42s       Normal   SuccessfulDelete   virtualmachineinstance/testvm   Deleted PodDisruptionBudget kubevirt-disruption-budget-8lqvs
default     5m12s       Normal   ShuttingDown       virtualmachineinstance/testvm   Signaled Graceful Shutdown
```
- Stop VM inside VM
```
default     68s         Normal   SuccessfulCreate   virtualmachineinstance/testvm   Created PodDisruptionBudget kubevirt-disruption-budget-fxzgm
default     68s         Normal   SuccessfulCreate   virtualmachineinstance/testvm   Created virtual machine pod virt-launcher-testvm-tghfr
default     53s         Normal   Created            virtualmachineinstance/testvm   VirtualMachineInstance defined.
default     53s         Normal   Started            virtualmachineinstance/testvm   VirtualMachineInstance started.
default     18s         Normal   Stopped            virtualmachineinstance/testvm   The VirtualMachineInstance was shut down.
[...]

PR #4818, opened 2023-12-05
feat: keep vmi events in vm annotations for debugging

Yu-Jack:
## Problem

Since there are very different reasons causes VM shutdown, we would like to keep this information as debugging information. Considering the event retention period, it would be better store these events in somewhere.

## Solution

I find out we could use event controller which I originally thought I can't use. That means we could save event by `event.InvolvedObject` into anywhere we would like.

We could keep these information in annotation like below demo screenshot. But, when I test my ideal, I found out there is an annotation limitation `metadata.annotations: Too long: must have at most 262144 bytes`. 

Although it seems to be abled to be addressed, but I think it would be better we just store latest X number events instead of all events. So, in this proposal, I use five as latest X number to store events, and it could be discussed surely.

## Related Issue

https://github.com/harvester/harvester/issues/3588

## Test plan

1. Create VM 
2. Stop VM from GUI

image

However, for some cases, it would notify multiple same events in short period of time. It might override too much events.
image

### Example events

- [normal events](https://gist.github.com/Yu-Jack/683e0ec7009034bdf22c56607dc7e449#file-example-events-js)
- [repeated events](https://gist.github.com/Yu-Jack/683e0ec7009034bdf22c56607dc7e449?permalink_comment_id=4783926#gistcomment-4783926)

## Additional DIscussion

I know it would be over-design at this moment. But, it would be large code base if we detect different kind and handle different behavior over time. We don't need to optimize it now surely, I just keep information about CoR pattern here for further investigation. The related branch is https://github.com/Yu-Jack/harvester/commit/84ba7af570a831156d557b25462e72aa75664357.

### DIFF
--- a/pkg/controller/master/event/event_controller.go
+++ b/pkg/controller/master/event/event_controller.go
@@ -2,9 +2,9 @@ package event
 
 import (
 	"encoding/json"
+	"fmt"
 
 	v1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
-	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
@@ -24,8 +24,7 @@ func (ctr *Controller) SyncEvent(_ string, event *corev1.Event) (*corev1.Event, 
 
 	if event.InvolvedObject.Kind == "VirtualMachineInstance" {
 		if err := ctr.updateVMEventRecord(event, event.InvolvedObject.Namespace, event.InvolvedObject.Name); err != nil {
-			logrus.Errorf("failed to update event record: %v", err)
-			return event, err
+			return event, fmt.Errorf("failed to update event record: %v", err)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


connorkuehl:
Have you seen https://github.com/harvester/eventrouter/? I only noticed it today, and so I don't know what its current state is or if it's used or maintained, but conceptually it seems like what you're after.

If the stdout sink is not durable enough, perhaps it could be extended to write to a file on a Longhorn PVC.

Yu-Jack:
> Have you seen https://github.com/harvester/eventrouter/? I only noticed it today, and so I don't know what its current state is or if it's used or maintained, but conceptually it seems like what you're after.
> 
> If the stdout sink is not durable enough, perhaps it could be extended to write to a file on a Longhorn PVC.

I didn't know that repository before. Currently, it's just a PoC in this PR, so I will take a look that repo. Thank for providing this information.

w13915984028:
FYI:

In production, users are suggested to deploy third party logging server to collect logs from Harvester cluster, per logging feature https://docs.harvesterhci.io/v1.2/logging/harvester-logging#event-log-format.

Harvester does not support in-cluster logging server by default, as such server will consume a big amount of disk spaces, and it also has cross-dependency issue.

Yu-Jack:
Need further investigation. I think recording those events on annotation is not a good option now.

Issue #4782, opened 2023-11-28
[TASK] Bump embedded Rancher to v2.7.9 and RKE2 to v1.26.x


bk201:
**Is your enhancement related to a problem? Please describe.**

Bump embedded Rancher and RKE2 to the next supported versions.

**Describe the solution you'd like**

- Rancher: v2.7.9
- RKE2: v1.26.x

**Describe alternatives you've considered**

**Additional context**

lanfon72:
Verified this change has been applied.
To be noticed that this bug been fixed on `master-head` but not `v1.3` branch yet.

![image](https://github.com/harvester/harvester/assets/5169694/8372b699-23f5-44ec-a138-075e6c0cf282)

Test Information
----
* Environment: **qemu/KVM single node**
* Harvester Version: **master-d3dc81ca-head**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Login to console, execute `sudo ranchered info`

PR #4868, opened 2023-12-14
minor change to patch capi-controller deployment

ibrokethecloud:
PR to patch capi-controller env variables. 

Needed for version upgrade to rancher 2.7.9 to work:

https://github.com/harvester/harvester-installer/pull/605

**Problem:**

A standalone capi-controller has been introduced in rancher v2.7.9.

The capi-controller is unable to reconcile the local rancher cluster.

capi-controller ties to generate a client for local cluster:

https://github.com/kubernetes-sigs/cluster-api/blob/main/controllers/remote/cluster_cache_tracker.go#L284-L349

It detects that the controller is running on the workload cluster being managed:

https://github.com/kubernetes-sigs/cluster-api/blob/main/controllers/remote/cluster_cache_tracker.go#L352-L375

As a result of this it patches the rest.Config to use the kubernetes default svc endpoint, however the `local-kubeconfig` token is generated by Rancher and not authenticated by rancher.

As a result the machines are never reconciled and controller keeps reporting the following error:

```
E1213 04:58:03.781942       1 controller.go:329] "Reconciler error" err="failed to create cluster accessor: error creating client for self-hosted cluster: error creating dynamic rest mapper for remote cluster \"fleet-local/local\": the server has asked for the client to provide credentials" controller="machine" controllerGroup="cluster.x-k8s.io" 
```

**Solution:**

The workaround is to patch the env variables for capi-controller.

This causes the `runningOnWorkloadCluster` check to fail, and local cluster is managed via the `local-kubeconfig` via the public rancher endpoint, and nodes are able to reconcile correctly.

**Related Issue:**
https://github.com/harvester/harvester/issues/4782
**Test plan:**

### DIFF
--- a/pkg/controller/master/rancher/embedded_rancher.go
+++ b/pkg/controller/master/rancher/embedded_rancher.go
@@ -2,12 +2,15 @@ package rancher
 
 import (
 	"encoding/json"
+	"reflect"
 	"strings"
 	"time"
 
 	rancherv3api "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	ranchersettings "github.com/rancher/rancher/pkg/settings"
 	"github.com/sirupsen/logrus"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/harvester/harvester/pkg/settings"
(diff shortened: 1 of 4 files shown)


FrankYang0529 (approved):
LGTM. Test with PR https://github.com/harvester/harvester-installer/pull/605. I can provision a new cluster without error.

bk201 (approved):
LGTM! Thanks!

bk201:
@mergifyio backport v1.2

Issue #4866, opened 2023-12-13
[ENHANCEMENT] Set default macaddress when vmi is deleted


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**

After we updated KubeVirt to v1.0.0, the actual generation is different from expected. The root cause is that we set macaddress to VM and don't restart the VM.

![image](https://github.com/harvester/harvester/assets/4927361/b76487ed-81d7-494c-a53b-d66413c81cba)

![image](https://github.com/harvester/harvester/assets/4927361/9dec64ff-3634-4a27-b2b2-3b8b197339a9)

**Describe the solution you'd like**
In backend, we can set the macaddress when VMI is removed, so VMI can have the same macaddress if users start the VM again.

In frontend, we can suggest users to restart the VM if they try to update vm.Spec and don't restart it.

**Additional context**
https://github.com/harvester/harvester/issues/4758

FrankYang0529:
Close this issue and use another https://github.com/harvester/harvester/issues/4909.

PR #4870, opened 2023-12-14
feat: set default macaddress when vmi is removed

FrankYang0529:
**Problem:**
After we updated KubeVirt to v1.0.0, the actual generation is different from expected. The root cause is that we set macaddress to VM and don't restart the VM.

image

image

**Solution:**
Set interfaces to VM when VMI is deleted.

**Related Issue:**
https://github.com/harvester/harvester/issues/4866
https://github.com/harvester/harvester/issues/4909

**Test plan:**

### Restart VM.
1. Create a VM. There is no warning message on the dashboard.
2. Check macaddress in `vmi.status.interfaces`.
3. Click `Restart` on the dashboard.
4. Check the macaddress is set to `vm.spec.template.spec.domain.devices.interfaces`.

### Use command to stop VM and start again on dashboard.
1. Create a VM. There is no warning message on the dashboard.
2. Run `sudo shutdown now` in the VM.
3. Check macaddress in `vmi.status.interfaces`. VMI status is Succeed in this step.
4. Click `Start` on the dashboard.
5. Check the macaddress is set to `vm.spec.template.spec.domain.devices.interfaces`.

### Migrate VM doesn't delete VMI, so the macaddress is not set to the VM, but it's still same in the new pod
1. Create a VM. There is no warning message on the dashboard.
2. Check macaddress in `vmi.status.interfaces`.
3. Click `Migrate` on the dashboard.
4. Check macaddress again in `vmi.status.interfaces`. They're same. Also, there is no warning message on the dashboard.

### DIFF
--- a/pkg/controller/master/virtualmachine/register.go
+++ b/pkg/controller/master/virtualmachine/register.go
@@ -87,7 +87,7 @@ func Register(ctx context.Context, management *config.Management, options config
 		vmCache:   vmCache,
 		vmiClient: virtualMachineInstanceClient,
 	}
-	virtualMachineInstanceClient.OnChange(ctx, vmControllerSetDefaultManagementNetworkMac, vmNetworkCtl.SetDefaultNetworkMacAddress)
+	virtualMachineInstanceClient.OnRemove(ctx, vmControllerSetDefaultManagementNetworkMac, vmNetworkCtl.SetDefaultNetworkMacAddress)
 
 	return nil
 }
--- a/pkg/controller/master/virtualmachine/vmi_network_controller.go
+++ b/pkg/controller/master/virtualmachine/vmi_network_controller.go
@@ -2,6 +2,7 @@ package virtualmachine
 
 import (
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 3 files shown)


starbops (approved):
Tested with all three cases. LGTM, thanks!

bk201:
截圖 2023-12-20 上午11 55 35
When a VM starts the first time, the GUI can't display its MAC address (The network has to be a VLAN network). The GUI seems to get the MAC address from the VirutalMachine resource.

ibrokethecloud:
this change will affect the UX. end users will not be able to view the mac address until the VM is stopped the first time.

would it not be better to have a mutating webhook to patch VM spec with a mac address if it is empty on creation. this will ensure mac address propogates to vmi and the vm and vmi generation stays in sync.

FrankYang0529:
> would it not be better to have a mutating webhook to patch VM spec with a mac address if it is empty on creation. this will ensure mac address propogates to vmi and the vm and vmi generation stays in sync.

In this case, we may need to have a way to prevent macaddress collision.

FrankYang0529:
Close the PR because we prefer https://github.com/harvester/harvester/pull/4948.

PR #4881, opened 2023-12-18
upgrade: create necessary RBAC rules for new CSI driver

Vicente-Cheng:

**Problem:**
The harvester CSI driver (v0.1.4 and later version) allows the user to specify the StorageClass they want. To validate it, we need to add new ClusterRole and ClusterRoleBinding for the RBAC.

**Solution:**
Create ClusterRole and ClusterRoleBinding when upgrading

**Related Issue:**
https://github.com/harvester/harvester/issues/5294

**Test plan:**
1. create 3 nodes harvester cluster
2. create guest cluster
3. execute upgrade
4. make sure both `ClusterRole` and `ClusterRoleBinding` are created

**ClusterRole**
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  creationTimestamp: "2023-12-18T11:01:06Z"
  labels:
    app.kubernetes.io/component: apiserver
    app.kubernetes.io/name: harvester
    app.kubernetes.io/part-of: harvester
  name: harvesterhci.io:csi-driver
  resourceVersion: "4411502"
  uid: e5736906-d080-4c3f-8f7a-affab4205a94
rules:
- apiGroups:
  - storage.k8s.io
  resources:
  - storageclasses
  verbs:
  - get
  - list
  - watch
```

**ClusterRoleBinding**
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  creationTimestamp: "2023-12-18T11:30:07Z"
  name: default-rke2-guest-01
  resourceVersion: "4462961"
  uid: 7a11d03a-090f-4b12-927f-f7a88ae061a3
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: harvesterhci.io:csi-driver
subjects:
- kind: ServiceAccount
  name: rke2-guest-01
  namespace: default
```
Please also check the `ServiceAccount` related info (name, namespace) should be the same as your guest cluster.

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -1299,6 +1299,77 @@ wait_for_deployment() {
   done
 }
 
+upgrade_rbac() {
+
+  # only versions before v1.2.2 that upgrading to v1.2.2 need this patch
+  if [[ ! "${UPGRADE_PREVIOUS_VERSION%%-rc*}" < "v1.2.2" ]]; then
+    echo "Only versions before v1.2.2 need this patch."
+    return
+  fi
+
+  full_names=$(upgrade-helper get-cloud-provider)
+  ret=$?
+  if [ -z "$full_names" ] || [ $ret != 0 ]; then
+    echo "No rolebinding harvesterhci.io:cloudprovider found or command failed, we can skip to patch rbac this moment."
+    return
+  fi
+
+  if kubectl get clusterrole harvesterhci.io:csi-driver 2> /dev/null; then
+    echo "ClusterRole harvesterhci.io:csi-driver already exists, skip to create it."
+  else
+    create_clusterrole_for_csi_driver
+  fi
+
+  # to handle multiple guest clusters
+  for item in $full_names
+  do
+    namespace=$(echo "$item" |awk -F '/' '{print $1}')
+    service_account=$(echo "$item" |awk -F '/' '{print $2}')
+
+    echo "Creating ClusterRoleBinding with ${namespace}/${service_account} ..."
+    cat <<EOF | kubectl create -f -
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: $namespace-$service_account
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: harvesterhci.io:csi-driver
+subjects:
+- kind: ServiceAccount
+  name: $service_account
+  namespace: $namespace
+EOF
+
+  done
+
+}
+
+create_clusterrole_for_csi_driver() {
+  echo "Creating ClusterRole harvesterhci.io:csi-driver ..."
+
+  cat <<EOF | kubectl create -f -
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/component: apiserver
+    app.kubernetes.io/name: harvester
+    app.kubernetes.io/part-of: harvester
+  name: harvesterhci.io:csi-driver
+rules:
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - watch
+EOF
+}
+
 wait_repo
 detect_repo
 detect_upgrade
@@ -1318,5 +1389,6 @@ upgrade_monitoring
 upgrade_logging_event_audit
 apply_extra_manifests
 upgrade_addons
+upgrade_rbac
 # wait fleet bundles upto 90 seconds
 wait_for_fleet_bundles 9
--- a/cmd/upgradehelper/get_cloud_providers.go
+++ b/cmd/upgradehelper/get_cloud_providers.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/harvester/harvester/pkg/util"
+)
+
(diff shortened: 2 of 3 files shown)


package/upgrade/upgrade_manifests.sh
```diff
@@ -1257,6 +1257,73 @@ sync_containerd_registry_to_rancher() {
...
 
+upgrade_rbac() {
+
+  # only versions before v1.3.0 that upgrading to v1.3.0 need this patch
+  if [[ ! "${UPGRADE_PREVIOUS_VERSION%%-rc*}" < "v1.3.0" ]]; then
+    echo "Only versions before v1.3.0 need this patch."
+    return
+  fi
+
+  full_name=$(kubectl get rolebinding -o wide |grep "harvesterhci.io:cloudprovider" |awk '{print $4}')
```
starbops: Do we need to add the `-A` flag here? Assuming the user created the guest cluster on a namespace different from the "default" NS.

The other one is, is it possible to have multiple results after grepping? That could cause the following logic to break. Thanks!

Vicente-Cheng: I think you are correct. I need to handle more than one guest cluster case.

Vicente-Cheng: updated! Add `-A` flags and also handle multiple items

starbops (approved):
LGTM, thanks!

package/upgrade/upgrade_manifests.sh
```diff
@@ -1257,6 +1257,79 @@ sync_containerd_registry_to_rancher() {
...
 
+upgrade_rbac() {
+
+  # only versions before v1.3.0 that upgrading to v1.3.0 need this patch
+  if [[ ! "${UPGRADE_PREVIOUS_VERSION%%-rc*}" < "v1.3.0" ]]; then
+    echo "Only versions before v1.3.0 need this patch."
+    return
+  fi
+
+  full_name=$(kubectl get rolebinding -A -o wide |grep "harvesterhci.io:cloudprovider" |awk '{print $5}')
```
bk201: Is it possible to make the search more reliable, like using the correct namespace/name or labels?

Vicente-Cheng: I think that because the guest cluster can configure the namespace on the creation, it seems we do not have a specific way to filter out what we need.

bk201: Or use yq to filter the output? e.g.:
```
kubectl get rolebinding -o yaml | yq '.items[] | select (.roleRef.name == "harvesterhci.io:cloudprovider" and .roleRef.kind == "ClusterRole") | .metadata.name'
```

We can also rely on https://github.com/harvester/harvester/pull/4984 to add a command if yq is too complicated here

bk201:
```suggestion
  full_names=$(kubectl get rolebinding -A -o wide |grep "harvesterhci.io:cloudprovider" |awk '{print $5}')
```
suggest change to names

Vicente-Cheng: I add a new command to get the full_names.
Please help to recheck, thanks!

package/upgrade/upgrade_manifests.sh
```diff
@@ -1256,6 +1256,80 @@ patch_local_cluster_details() {
   kubectl patch -n fleet-local cluster.provisioning local --subresource=status --type=merge --patch '{"status":{"fleetWorkspaceName": "fleet-local"}}'
 }
 
+upgrade_rbac() {
+
+  # only versions before v1.3.0 that upgrading to v1.3.0 need this patch
+  if [[ ! "${UPGRADE_PREVIOUS_VERSION%%-rc*}" < "v1.3.0" ]]; then
+    echo "Only versions before v1.3.0 need this patch."
```
bk201: I think we should make this PR in the v1.2 branch. The users have to upgrade from v1.2.1 to v1.2.x first.

Vicente-Cheng:
Hi folks, I backport the changes to the v1.2
Please help to review it again. Thanks!

package/upgrade/upgrade_manifests.sh
```diff
@@ -1299,6 +1299,80 @@ wait_for_deployment() {
   done
 }
 
+upgrade_rbac() {
+
+  # only versions before v1.3.0 that upgrading to v1.3.0 need this patch
```
bk201: need to refactor the comment as well.

package/upgrade/upgrade_manifests.sh
```diff
@@ -1299,6 +1299,80 @@ wait_for_deployment() {
...
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - watch
+EOF
+
+  echo "Creating clusterrole with ${clusterrole_manifest} ..."
+  kubectl create -f "$clusterrole_manifest"
```
bk201: Seeing the error when testing with https://github.com/harvester/harvester/pull/5558:
```
Creating clusterrole with /tmp/clusterrole.yaml ...
Error from server (AlreadyExists): error when creating "/tmp/clusterrole.yaml": clusterroles.rbac.authorization.k8s.io "harvesterhci.io:csi-driver" already exists
```

With #5558, the harvester pod creates the cluster role after the harvester pod is upgraded.

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: updated, thanks!

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: Good catch... the upgrade will apply the new template of the chart

Add another checking for this op.
Please help to recheck, thanks!

bk201 (approved):
LGTM.

starbops (approved):
LGTM and the fix works as expected with the test plan. Thank you.

Issue #4865, opened 2023-12-13
[ENHANCEMENT] Use general image for promote job


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
Currently, there're two image values in [promote_controller.go](https://github.com/FrankYang0529/harvester/blob/197012aa566c55469472e4c8970d7f6a0912bfb2/pkg/controller/master/node/promote_controller.go#L48) and [helm chart](https://github.com/FrankYang0529/harvester/blob/197012aa566c55469472e4c8970d7f6a0912bfb2/deploy/charts/harvester/values.yaml#L491-L492). We can use Rancher Apps to get the value from helm chart, so we don't need to maintain the same value in two places.

**Additional context**
https://github.com/harvester/harvester/pull/4852#pullrequestreview-1777529928

albinsun:
Doea this ticket relates/same to #4839 ?

Yu-Jack:
Yes, it's almost same. But, original PR https://github.com/harvester/harvester/pull/4852 doesn't fetch value from rancher app, so we open another enhancement issue (this one 4865) to track this. It's about implementation.

albinsun:
Provide test result on `v1.3.0-dev-20240105`, close as fixed.


## Environment
* Harvester
 * Version: `v1.3.0-dev-20240105`
 * Profile: QEMU/KVM, 3 nodes (8C/16G/500G)
 * ui-source: `Auto`


## Steps 
1. <details open>
     <summary> Check helm chart value: <code>registry.suse.com/bci/bci-base:15.5</code>
     </summary>
 
     ![image](https://github.com/harvester/harvester/assets/2773781/585597c5-19b0-486d-acd1-b0f6a9435671)
   </details>
   
1. Check promote jobs uses image same as helm chart value
   * <details open>
       <summary> harvester-promote-harvester-node-1 :ballot_box_with_check: 
       </summary>
       
       ![image](https://github.com/harvester/harvester/assets/2773781/8ed8e383-7b5f-4437-b075-9c05c3c6165f)
     </details>

   * <details open>
       <summary> harvester-promote-harvester-node-2 :ballot_box_with_check: 
       </summary>
       
       ![image](https://github.com/harvester/harvester/assets/2773781/e0c818d0-5bf8-4426-896c-15f0b1963cc5)
     </details>
     
   * <details open>
       <summary> Jobs are completed :ballot_box_with_check: 
       </summary>
       
[...]

PR #4896, opened 2023-12-20
feat: fetch app generalJob image information from rancher apps

Yu-Jack:
## **Problem:**

It would be annoying to manage repository name and tag in different places, so it would be better to have a single source of truth. Currently, the image of promote job exists in [promote_controller.go](https://github.com/FrankYang0529/harvester/blob/197012aa566c55469472e4c8970d7f6a0912bfb2/pkg/controller/master/node/promote_controller.go#L48) and [helm chart](https://github.com/FrankYang0529/harvester/blob/197012aa566c55469472e4c8970d7f6a0912bfb2/deploy/charts/harvester/values.yaml#L491-L492).

## **Solution:**

Because the chart value is in the Rancher APP CR, so we could retrieve from there.

However, in the Golang limitation, we couldn't parse like `"tag": 12.20` in map such as [this playground](https://go.dev/play/p/dfJQ9xSlPeo). So, if we have that kind of version, we should use `"tag": "12.20"` instead in the future.

## **Related Issue:**

https://github.com/harvester/harvester/issues/4865

## **Test plan:**

- Try to create one management harvester cluster, and create another two harvester to join. 
  Testing Image: `jk82421/harvester:v1.2.1.7`
  image
- Because I also modified the updateSupportBundleImage, so I tested it as well. I also added test case for this.
  Testing Image: `jk82421/harvester:v1.2.1.8`
  image
  image

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -489,6 +489,9 @@ generalJob:
   image:
     imagePullPolicy: IfNotPresent
     repository: registry.suse.com/bci/bci-base
+    # However, in the Golang limitation, we couldn't parse like "tag": 12.20.
+    # So, we must use "tag": "12.20" instead.
+    # More details in PR-4896 (https://github.com/harvester/harvester/pull/4896)
     tag: 15.5
 
 whereabouts:
@@ -503,6 +506,7 @@ harvester-node-manager:
     imagePullPolicy: IfNotPresent
     repository: rancher/harvester-node-manager
     tag: master-head
+
 
 snapshot-validation-webhook:
   enabled: true
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -507,7 +507,6 @@ harvester-node-manager:
     repository: rancher/harvester-node-manager
     tag: master-head
 
-
 snapshot-validation-webhook:
   enabled: true
   image:
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -489,7 +489,7 @@ generalJob:
   image:
     imagePullPolicy: IfNotPresent
     repository: registry.suse.com/bci/bci-base
-    # However, in the Golang limitation, we couldn't parse like "tag": 12.20.
+    # In the Golang limitation, we couldn't parse like "tag": 12.20.
     # So, we must use "tag": "12.20" instead.
     # More details in PR-4896 (https://github.com/harvester/harvester/pull/4896)
     tag: 15.5
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -11,6 +11,7 @@ import (
 	ctlbatchv1 "github.com/rancher/wrangler/pkg/generated/controllers/batch/v1"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/name"
+	"github.com/sirupsen/logrus"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -395,6 +396,7 @@ func isPromoteStatusIn(node *corev1.Node, statuses ...string) bool {
 func (h *PromoteHandler) createPromoteJob(node *corev1.Node) (*batchv1.Job, error) {
 	image, err := utilCatalog.FetchAppChartImage(h.appCache, h.namespace, releaseAppHarvesterName, []string{"generalJob", "image"})
 	if err != nil {
+		logrus.Errorf("failed to get harvester image (%s): %v", image.ImageName(), err)
 		return nil, err
 	}
 
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -74,7 +74,7 @@ type PromoteHandler struct {
 }
 
 // PromoteRegister registers the node controller
-func PromoteRegister(ctx context.Context, management *config.Management, options config.Options) (err error) {
+func PromoteRegister(ctx context.Context, management *config.Management, options config.Options) error {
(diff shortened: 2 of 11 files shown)


pkg/controller/master/node/promote_controller.go
```diff
@@ -68,18 +71,21 @@ type PromoteHandler struct {
 	jobCache  ctlbatchv1.JobCache
 	recorder  record.EventRecorder
 	namespace string
+	apps      catalogv1.AppController
```
FrankYang0529: Since we don't need to update `App` in this controller, could we just use `catalogv1.AppCache`? Thanks.

pkg/controller/master/node/promote_controller.go
```diff
@@ -91,6 +97,20 @@ func PromoteRegister(ctx context.Context, management *config.Management, options
 	return nil
 }
 
+func fetchPromoteImage(appClient catalogv1.AppClient, namespace, name string) error {
```
FrankYang0529: Is it better to return promote image name rather than storing in a variable which can be modified by code in `node` package?

Yu-Jack: Yes, it makes more sense. I could also remove `fetchPromoteImage` function because I could just directly call utilCatalog to retrieve image name.

Yu-Jack:
BTW, I also tested again. It seems well.
image

FrankYang0529 (approved):
LGTM.

```sh
> kubectl get job -n harvester-system harvester-promote-harvester-node-1 -o yaml | yq ".spec.template.spec.containers.[0].image"
registry.suse.com/bci/bci-base:15.5
```

pkg/controller/master/node/promote_controller.go
```diff
@@ -388,15 +393,20 @@ func isPromoteStatusIn(node *corev1.Node, statuses ...string) bool {
 }
 
 func (h *PromoteHandler) createPromoteJob(node *corev1.Node) (*batchv1.Job, error) {
-	job := buildPromoteJob(h.namespace, node)
+	image, err := utilCatalog.FetchAppChartImage(h.appCache, h.namespace, releaseAppHarvesterName, []string{"generalJob", "image"})
+	if err != nil {
+		return nil, err
+	}
+
```
FrankYang0529, line 400: NIT: can we add a log info to print image name here. It may help debug in the future. Thanks.

FrankYang0529:
> Because the chart value is in the Rancher APP CR, so we could retrieve from there.
> 
> However, in the Golang limitation, we couldn't parse like `"tag": 12.20` in map such as [this playground](https://go.dev/play/p/dfJQ9xSlPeo). So, if we have that kind of version, we should use `"tag": "12.20"` instead in the future.

May we add a comment in the chart, so other developers can catch it in the future? Thanks.

https://github.com/harvester/harvester/blob/0fb1e501ee27046fc4dbb9c9dcb192ebd58e3199/deploy/charts/harvester/values.yaml#L491-L492

pkg/controller/master/node/promote_controller.go
Yu-Jack: Surely, I will add some logs in catalog util as well.

Yu-Jack:
> > Because the chart value is in the Rancher APP CR, so we could retrieve from there.
> > However, in the Golang limitation, we couldn't parse like `"tag": 12.20` in map such as [this playground](https://go.dev/play/p/dfJQ9xSlPeo). So, if we have that kind of version, we should use `"tag": "12.20"` instead in the future.
> 
> May we add a comment in the chart, so other developers can catch it in the future? Thanks.
> 
> https://github.com/harvester/harvester/blob/0fb1e501ee27046fc4dbb9c9dcb192ebd58e3199/deploy/charts/harvester/values.yaml#L491-L492

Yeah, totally agree. I'll add it.

deploy/charts/harvester/values.yaml
```diff
@@ -489,6 +489,9 @@ generalJob:
   image:
     imagePullPolicy: IfNotPresent
     repository: registry.suse.com/bci/bci-base
+    # However, in the Golang limitation, we couldn't parse like "tag": 12.20.
+    # So, we must use "tag": "12.20" instead.
+    # More details in PR-4896 (https://github.com/harvester/harvester/pull/4896)
     tag: 15.5
```
Yu-Jack, line 495: @FrankYang0529 If there is anything you would like to add more explanation, please let me know.

FrankYang0529: LGTM. Thanks.

deploy/charts/harvester/values.yaml
```diff
@@ -489,6 +489,9 @@ generalJob:
   image:
     imagePullPolicy: IfNotPresent
     repository: registry.suse.com/bci/bci-base
+    # In the Golang limitation, we couldn't parse like "tag": 12.20.
+    # So, we must use "tag": "12.20" instead.
+    # More details in PR-4896 (https://github.com/harvester/harvester/pull/4896)
```
Yu-Jack, line 494: I removed `however` here, it seems weird in this context.

pkg/util/catalog/app.go
```diff
@@ -0,0 +1,47 @@
...
+		ok            bool
+		bytes         []byte
+		value         = harvesterApp.Spec.Chart.Values
+		previousValue map[string]interface{}
+	)
+
+	for _, key := range keyNames {
+		previousValue = value
+		if value, ok = value[key].(map[string]interface{}); !ok {
+			logrus.Debugf("cannot find key %s in layer(%s) %s/%s app chart value: %v", key, strings.Join(keyNames, ","), namespace, name, previousValue)
```
bk201: maybe just saying `fail to get value of path %s' % strings.Join(keyNames, ".")` ?

FWIW, this does a similar thing: https://github.com/harvester/harvester/blob/aacd8a0a3a30a84f02e48a98d2f8b2d183a176bd/vendor/github.com/rancher/norman/types/values/values.go#L69-L85, but we can keep your implementation here.

Yu-Jack: Okay, I think your message is more suitable.

bk201 (approved):
LGTM, please squash the commits.

Yu-Jack:
@Mergifyio backport v1.2

Issue #4907, opened 2023-12-22
[ENHANCEMENT] Take advantage of the attach/detach mechanism in Longhorn v1.5+


ejweber:
While looking into a few Harvester issues, I noticed https://github.com/harvester/harvester/pull/3670 and https://github.com/harvester/harvester/issues/3648. It appears from these (though I need to collect more context) that Harvester manually attached Longhorn volumes to nodes in order to ensure snapshots/backups can be taken.

This behavior makes sense for Longhorn <= `v1.4.x`. In those versions, Longhorn did not have any mechanism to attach the volume automatically to the node and perform the operation, so these operations were generally stuck until a volume was attached. However, in v1.5.x, Longhorn added a new attach/detach controller that automatically attaches volumes for these operations and then detaches them. See https://github.com/longhorn/longhorn/blob/master/enhancements/20221024-longhorn-volumeattachment.md and https://github.com/longhorn/longhorn/issues/3726.

I don't know the full extent of Harvester's need to manipulate Longhorn volumes, but I think it may be possible to remove or reduce this functionality in Harvester. It may even be safer to do so to avoid situations in which Harvester's attach/detach logic fights Longhorn's.

Copying @phanle1010 and @james-munson, who I was discussing this with.

bk201:
Thanks @ejweber for the heads-up. We'll evaluate the change.

TachunLin:
Verified **fixed** on `master-6ca6acfd-head` (24/01/23). Close this issue.

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ Case 1: snapshot can work on a stopped VM $~~$</summary>
* VM volumes are detached after we stop VM. 


https://github.com/harvester/harvester/assets/29251855/2e9bb93e-2dc0-432c-8884-f6bfb275e32b

  ```
  node1:~ # kubectl get volume -A
  NAMESPACE         NAME                                       STATE      ROBUSTNESS   SCHEDULED   SIZE          NODE   AGE
  longhorn-system   pvc-5a861225-920d-4059-b501-f02b2fd0ff27   detached   unknown                  10737418240          19m
  ```

* Take the vm snapshot, The snapshot of vm in off state can be ready.
![image](https://github.com/harvester/harvester/assets/29251855/35f3c4ed-e8cb-4492-b99e-98ca91648398)


 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Case 2: restore a snapshot from detached volumes can work. $~~$</summary>

[...]

PR #4910, opened 2023-12-25
refactor: remove mount logic in harvester

FrankYang0529:
**Solution:**
LH introduces VolumeAttachment CRD after v1.5. It can automatically attach/detach volume when there is a request from LH API / controller / CSI. We can just leverage LH mechanism and remove our implementation.

**Related Issue:**
https://github.com/harvester/harvester/issues/4907

**Test plan:**

### Setup

1. Create a 3-node harvester from this branch.
2. Setup backup-target.

### Case 1: snapshot can work on a stopped VM.

1. Create a VM.
2. After the VM is ready, stop the VM.
3. Check VM volumes are detached.
4. Take a snapshot on the VM. The snapshot can be ready.

### Case 2: restore a snapshot from detached volumes can work.

1. Follow Case 1.
2. Make sure VM volumes are detached.
3. Restore the snapshot to a new VM. The new VM can be ready.
4. Restore the snapshot to replace the old VM. The old VM can be ready.

### Case 3: backup can work on a stopped VM.

1. Create a VM.
2. After the VM is ready, stop the VM.
3. Check VM volumes are detached.
4. Take a backup on the VM. The backup can be ready.

### Case 4: race condition doesn't break VMBackup.

1. Create a VM.
2. After the VM is ready, stop the VM.
3. Check VM volumes are detached.
4. Take multiple backup on the VM in a short time. All backup can be ready.

### Case 5: restore a backup from detached volumes can work.

1. Follow Case 3.
2. Make sure VM volumes are detached.
3. Restore the backup to a new VM. The new VM can be ready.
4. Restore the backup to replace the old VM. The old VM can be ready.

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -15,7 +15,6 @@ import (
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	lhdatastore "github.com/longhorn/longhorn-manager/datastore"
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
-	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	lhutil "github.com/longhorn/longhorn-manager/util"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/name"
@@ -478,11 +477,6 @@ func (h *RestoreHandler) reconcileResources(
 	// reconcile restoring volumes and create new PVC from CSI volumeSnapshot if not exist
 	isVolumesReady, err := h.reconcileVolumeRestores(vmRestore, backup)
 	if err != nil {
-		return nil, false, err
-	}
-
-	// mount volumes after creating PVCs, so detaching volumes controller doesn't detach the volumes
-	if err := h.mountLonghornVolumes(backup); err != nil {
 		return nil, false, err
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/master/backup/restore.go
```diff
@@ -481,11 +480,6 @@ func (h *RestoreHandler) reconcileResources(
 		return nil, false, err
 	}
 
-	// mount volumes after creating PVCs, so detaching volumes controller doesn't detach the volumes
-	if err := h.mountLonghornVolumes(backup); err != nil {
-		return nil, false, err
-	}
-
```
WebberHuang1118, line 488: If my understanding is correct, since the restored pvc is from a backup volume's snapshot, hence during the new pvc creation we have to make sure the backup volume is attached.

[From the LH A/D implementation](https://github.com/longhorn/longhorn/blob/master/enhancements/20221024-longhorn-volumeattachment.md), we can make sure the volume would be attached during snapshot creation. However, in this case, can we make sure the volume would keep attached through new pvc provision?

WebberHuang1118 (commented):
LGTM, just a question for restore from VM snapshot, thanks.

pkg/controller/master/backup/restore.go
FrankYang0529: I think creating a PVC from a CSI Snapshot is also a CSI request. In `csi-provisioner` logs, I can see new PVC name. Following are my test steps:

1. In Harvester Volume page, create a volume `vol1`.
2. Take a snapshot for `vol1`, `vol1` state changes from `detached` to `attached`.
3. After the snapshot is finished, `vol1` state changes back to `detached`.
4. Provisioning a new PVC from the snapshot, `vol1` state changes from `detached` to `attached` again.
5. After a new PVC is provisioned, `vol1` state changes back to `detached`.

WebberHuang1118: That makes sense, thanks.

Vicente-Cheng (approved):
LGTM, also tested all test cases related to the snapshot/backup.

WebberHuang1118 (approved):
LGTM, all testing cases work as expected. Thanks.

bk201 (approved):
LGTM!!

FrankYang0529:
@mergify backport v1.3

FrankYang0529:
@mergify backport v1.2

Issue #4922, opened 2023-12-28
[ENHANCEMENT] Support bundle resource clean up


bk201:
**Is your enhancement related to a problem? Please describe.**

When creating a support bundle from GUI, the SupportBundle resources are only deleted after the bundle is fully downloaded ([ref](https://github.com/harvester/harvester/blob/fb1e5ae8fd817cdd879bb93a31966cbaf3749a3d/pkg/api/supportbundle/download_handler.go#L117-L122)). There are cases that downloads can't finish and there will be many unused SB and support-bundle-manager pods left in the cluster. This can occupy cluster resources and stop users from provisioning more VMs or services.

**Describe the solution you'd like**

Clean up unused support bundle resources after a certain period. We can consider if we should retain the bundle file.

**Describe alternatives you've considered**

**Additional context**

Yu-Jack:
## Add more context

When I test to create multiple support bundle in harvester at the same time, I observe following situation.

## Create multiple support bundle at the same time

```yaml
# Longhorn's apiVersion is `longhorn.io/v1beta2`, and namespace is `longhorn-system`

apiVersion: harvesterhci.io/v1beta1
kind: SupportBundle
metadata:
  name: bundle-a1
  namespace: harvester-system
spec:
  issueURL: "this is issueURL"
  description: "this is description"
---
apiVersion: harvesterhci.io/v1beta1
kind: SupportBundle
metadata:
  name: bundle-a2
  namespace: harvester-system
spec:
  issueURL: "this is issueURL"
  description: "this is description"
---
apiVersion: harvesterhci.io/v1beta1
kind: SupportBundle
metadata:
  name: bundle-a3
  namespace: harvester-system
spec:
  issueURL: "this is issueURL"
  description: "this is description"
```

### In Harvester

It causes the support bundle manager failed to start agent Daemonset, the reason is [this condition](https://github.com/rancher/support-bundle-kit/blob/4c92b3e4902ca6aa1339e4f35cb6422a4958a7f3/pkg/manager/manager.go#L370-L373) is always `false`. 

When you create three support bundle at the same time, it creates three Daemonset. But `len(pods.Items)` is 3, and the `daemonSet.Status.DesiredNumberScheduled` is 1, that's why it gets `false`.

### In Longhorn

[...]

PR #4939, opened 2024-01-03
feat: delete unused support bundle after a period of time

Yu-Jack:
**Problem:**

Support bundle CR isn't deleted due to refreshing window or something which terminates the downloading process. The leftover support bundle manager occupies the cluster resource.

**Solution:**

***We choose 30 minutes in the end***

Check the exist time with `SupportBundleInitialized` status (This status is in final state).

if this time is longer than 10 minutes, that means it might be broken or error, then delete the support bundle CR. 

I think 10 minutes is enough to wait after the support bundle is ready for downloading.

**Related Issue:**

https://github.com/harvester/harvester/issues/4922

**Test plan:**

1. Create support bundle in harvester dashboard.
2. Refresh the window to avoid the harvester dashboard to trigger download API.
3. Support bundle CR should be deleted after x seconds.
  Go to settings to change `support-bundle-expiration` to like 1 (mean 1 minute), support bundle will be delete in 1 minute. Then, use `k get supportbundles -n harvester-system` to check resource.

### DIFF
--- a/pkg/controller/master/supportbundle/controller.go
+++ b/pkg/controller/master/supportbundle/controller.go
@@ -11,6 +11,7 @@ import (
 	ctlappsv1 "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	settingctl "github.com/harvester/harvester/pkg/controller/master/setting"
@@ -18,6 +19,8 @@ import (
 	"github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/settings"
 )
+
+const supportBundleExistTimeLimit = 10 * time.Minute
 
 // Handler generates support bundles for the cluster
 type Handler struct {
@@ -70,10 +73,37 @@ func (h *Handler) OnSupportBundleChanged(key string, sb *harvesterv1.SupportBund
 	case types.StateGenerating:
 		logrus.Debugf("[%s] support bundle is being generated", sb.Name)
 		return h.checkManagerStatus(sb)
+	case types.StateError, types.StateReady:
+		return h.checkExistTime(sb)
 	default:
 		logrus.Debugf("[%s] noop for state %s", sb.Name, sb.Status.State)
 		return sb, nil
 	}
+}
+
+// checkExistTime checks if the support bundle has been updated for more than supportBundleExistTimeLimit minutes
+// If yes, that means the support bundle file is not be retrieved, and it should be deleted.
+func (h *Handler) checkExistTime(sb *harvesterv1.SupportBundle) (*harvesterv1.SupportBundle, error) {
+	t, err := time.Parse(time.RFC3339, sb.Status.Conditions[0].LastUpdateTime)
+	if err != nil {
+		logrus.Debugf("[%s] fail to parse %s", sb.Name, sb.Status.Conditions[0].LastUpdateTime)
+		return sb, err
+	}
+
+	existTime := time.Now().Sub(t)
+
+	logrus.Debugf("[%s] support bundle status: %s exist time is %s", sb.Name, sb.Status.State, existTime.String())
+	if existTime < supportBundleExistTimeLimit {
+		h.supportBundleController.EnqueueAfter(sb.Namespace, sb.Name, supportBundleExistTimeLimit)
+		return sb, err
+	}
+
+	if err := h.supportBundles.Delete(sb.Namespace, sb.Name, &metav1.DeleteOptions{}); err != nil {
+		return sb, err
+	}
+
+	logrus.Infof("[%s] support bundle is deleted", sb.Name)
+	return nil, nil
 }
 
 func (h *Handler) updateSupportBundleImageSetting() error {
--- a/pkg/controller/master/supportbundle/controller.go
+++ b/pkg/controller/master/supportbundle/controller.go
@@ -11,6 +11,7 @@ import (
 	ctlappsv1 "github.com/rancher/wrangler/pkg/generated/controllers/apps/v1"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
@@ -82,7 +83,7 @@ func (h *Handler) OnSupportBundleChanged(key string, sb *harvesterv1.SupportBund
 // checkExistTime checks if the support bundle has been updated for more than supportBundleExistTimeLimit minutes
 // If yes, that means the support bundle file is not be retrieved, and it should be deleted.
 func (h *Handler) checkExistTime(sb *harvesterv1.SupportBundle) (*harvesterv1.SupportBundle, error) {
-	t, err := time.Parse(time.RFC3339, sb.Status.Conditions[0].LastUpdateTime)
+	t, err := time.Parse(time.RFC3339, harvesterv1.SupportBundleInitialized.GetLastUpdated(sb))
 	if err != nil {
 		logrus.Debugf("[%s] fail to parse %s", sb.Name, sb.Status.Conditions[0].LastUpdateTime)
 		return sb, err
@@ -97,7 +98,7 @@ func (h *Handler) checkExistTime(sb *harvesterv1.SupportBundle) (*harvesterv1.Su
 		return sb, err
 	}
 
-	if err := h.supportBundles.Delete(sb.Namespace, sb.Name, &metav1.DeleteOptions{}); err != nil {
+	if err := h.supportBundles.Delete(sb.Namespace, sb.Name, &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
 		return sb, err
 	}
 
--- a/pkg/controller/master/supportbundle/controller.go
+++ b/pkg/controller/master/supportbundle/controller.go
@@ -20,8 +20,6 @@ import (
 	"github.com/harvester/harvester/pkg/settings"
 )
 
-const supportBundleExistTimeLimit = 10 * time.Minute
-
 // Handler generates support bundles for the cluster
 type Handler struct {
 	supportBundles          v1beta1.SupportBundleClient
@@ -91,10 +89,11 @@ func (h *Handler) checkExistTime(sb *harvesterv1.SupportBundle) (*harvesterv1.Su
 	}
 
 	existTime := time.Now().Sub(t)
+	limit := settings.SupportBundleExistTimeLimitSeconds.GetDuration() * time.Second
 
 	logrus.Debugf("[%s] support bundle status: %s exist time is %s", sb.Name, sb.Status.State, existTime.String())
-	if existTime < supportBundleExistTimeLimit {
-		h.supportBundleController.EnqueueAfter(sb.Namespace, sb.Name, supportBundleExistTimeLimit)
+	if existTime < limit {
+		h.supportBundleController.EnqueueAfter(sb.Namespace, sb.Name, limit)
 		return sb, err
 	}
 
--- a/pkg/controller/master/supportbundle/controller.go
+++ b/pkg/controller/master/supportbundle/controller.go
@@ -86,7 +86,7 @@ func (h *Handler) OnSupportBundleChanged(key string, sb *harvesterv1.SupportBund
 func (h *Handler) checkExistTime(sb *harvesterv1.SupportBundle) (*harvesterv1.SupportBundle, error) {
 	t, err := time.Parse(time.RFC3339, harvesterv1.SupportBundleInitialized.GetLastUpdated(sb))
 	if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/controller/master/supportbundle/controller.go
```diff
@@ -19,6 +20,8 @@ import (
 	"github.com/harvester/harvester/pkg/settings"
 )
 
+const supportBundleExistTimeLimit = 10 * time.Minute
```
bk201: Can we make the value longer to 30 minutes and configurable? We can create a setting for this.

Yu-Jack: surely, sounds good!

pkg/settings/settings.go
```diff
@@ -154,6 +157,10 @@ func (s Setting) GetInt() int {
 	return i
 }
 
+func (s Setting) GetDuration() time.Duration {
+	return time.Duration(s.GetInt())
+}
```
Yu-Jack: Just add helper function to convert

pkg/settings/settings.go
```diff
@@ -78,6 +80,7 @@ const (
 	HarvesterCSICCMSettingName                        = "harvester-csi-ccm-versions"
 	StorageNetworkName                                = "storage-network"
 	DefaultVMTerminationGracePeriodSecondsSettingName = "default-vm-termination-grace-period-seconds"
+	SupportBundleExistTimeLimitSecondsSettingName     = "support-bundle-exist-time-seconds"
```
bk201: How about `support-bundle-gc-timeout`? Thanks @ibrokethecloud

Yu-Jack: How about `support-bundle-expired-time` ? `gc-timeout` sounds like it encounters timeout while gcing.

Yu-Jack: @bk201 @ibrokethecloud How about `support-bundle-expired-time` ? `gc-timeout` sounds like it encounters timeout while gcing.

ibrokethecloud: `support-bundle-expiration-time` ?

Yu-Jack: I think `support-bundle-expiration-seconds` would be nice

bk201 (approved):
LGTM, thanks!

pkg/settings/settings.go
```diff
@@ -154,6 +157,10 @@ func (s Setting) GetInt() int {
 	return i
 }
 
+func (s Setting) GetDuration() time.Duration {
+	return time.Duration(s.GetInt())
```
w13915984028: If the setting is cleared, s.GetInt() returns 0, will this function return the default value? And, this value should have a minimal value, user can't set it as 1, 2 ...; for the max value, a limitation seems also required.

BTW, with a new `setting`, the Harvester document needs to updated, do we have another PR for it ? thanks.

pkg/controller/master/supportbundle/controller.go
```diff
@@ -70,12 +71,40 @@ func (h *Handler) OnSupportBundleChanged(key string, sb *harvesterv1.SupportBund
...
+	existTime := time.Now().Sub(t)
+	expiredTime := settings.SupportBundleExpirationSeconds.GetDuration() * time.Second
+
+	logrus.Debugf("[%s] support bundle status: %s exist time is %s", sb.Name, sb.Status.State, existTime.String())
+	if existTime < expiredTime {
+		h.supportBundleController.EnqueueAfter(sb.Namespace, sb.Name, expiredTime)
+		return sb, err
+	}
+
+	if err := h.supportBundles.Delete(sb.Namespace, sb.Name, &metav1.DeleteOptions{}); err != nil {
```
w13915984028: Skip the error like `NotFound`

pkg/controller/master/supportbundle/controller.go
```diff
@@ -70,12 +71,40 @@ func (h *Handler) OnSupportBundleChanged(key string, sb *harvesterv1.SupportBund
...
 	default:
 		logrus.Debugf("[%s] noop for state %s", sb.Name, sb.Status.State)
 		return sb, nil
 	}
 }
 
+// checkExistTime checks if the support bundle has been updated for more than supportBundleExistTimeLimit minutes
+// If yes, that means the support bundle file is not be retrieved, and it should be deleted.
+func (h *Handler) checkExistTime(sb *harvesterv1.SupportBundle) (*harvesterv1.SupportBundle, error) {
+	t, err := time.Parse(time.RFC3339, sb.Status.Conditions[0].LastUpdateTime)
```
w13915984028: Here and the test function, assume that the Conditions[0] is the target, it is better to use the provided function to get the target condition, when get, do check, if not, log something and return.

w13915984028 (commented):
Please check those robustness related questions, thanks.

pkg/settings/settings.go
Yu-Jack: Currently, `GetInt` is implemented like below.

```go
func (s Setting) GetInt() int {
	v := s.Get()
	i, err := strconv.Atoi(v)
	if err == nil {
		return i
	}
	logrus.Errorf("failed to parse setting %s=%s as int: %v", s.Name, v, err)
	i, err = strconv.Atoi(s.Default)
	if err != nil {
		return 0
	}
	return i
}
```

So, if the setting is cleared, `GetInt` returns default value and `GetDuration` also return default value. 

It seems there is no any existing mechanism to add this restriction of minimum and maximum, but I think it's good to add it. Maybe we could open another enhancement issue to track this? Because I think it would be better to check other settings's minimum and maximum if it's necessary.

Regarding document, thanks for reminding. I'll check this with @bk201.

w13915984028: The value limitation is more checked in webhook.
Assume someone test it with value 1, it will only bring issues to us.

w13915984028:
```
	v := s.Get()
	i, err := strconv.Atoi(v)
```

Check `v == ""` first; otherwise, it will print the error

pkg/settings/settings.go
```diff
@@ -141,19 +144,30 @@ func (s Setting) Get() string {
...
+		if i, err = strconv.Atoi(v); err == nil {
+			return i
+		}
+		logrus.Errorf("failed to parse setting %s=%s as int: %v", s.Name, v, err)
 	}
-	logrus.Errorf("failed to parse setting %s=%s as int: %v", s.Name, v, err)
-	i, err = strconv.Atoi(s.Default)
-	if err != nil {
+
+	if i, err = strconv.Atoi(s.Default); err != nil {
```
w13915984028, line 161: return the default value 1800 instead of 0, if this unlucky failure happens

Yu-Jack: if I failed to parse the default value, how do I return default value? 

If it could parse it correctly, it just return default value, but if it couldn't. The default value might be `1800ss`, I could't also return it as well.

BTW, I've added the webhook to validate the value here.

Yu-Jack: Or do you mean if I get the zero in controller, I just reassign 1800 to it?

Yu-Jack: Or do you mean if I get zero in controller, I just reassign 1800 to it?

w13915984028: This function, in any abnormal case, should return the none-zero default value, at the moment it is 1800. The `0` means the file will be deleted right after it is generated.

This odd situation is caused by below line, the default value can only be a string, and you set the 1800 a direct string.

Maybe you can define 2 consts in the uitl, "1800" as a string and referred in below, and another 1800 as an int, which is used as the fallback value `GetInt() int`

`SupportBundleExpirationSeconds = NewSetting(SupportBundleExpirationSecondsSettingName, "1800")`

Those 2 consts are put together, and with a comment saying they need to changed at mean time.

pkg/settings/settings.go
```diff
@@ -37,6 +38,7 @@ var (
 	SupportBundleImage                     = NewSetting(SupportBundleImageName, "{}")
 	SupportBundleNamespaces                = NewSetting("support-bundle-namespaces", "")
 	SupportBundleTimeout                   = NewSetting(SupportBundleTimeoutSettingName, "10") // Unit is minute. 0 means disable timeout.
+	SupportBundleExpirationSeconds         = NewSetting(SupportBundleExpirationSecondsSettingName, "1800")
```
w13915984028:
```
	SupportBundleTimeout                   = NewSetting(SupportBundleTimeoutSettingName, "10") // Unit is minute. 0 means disable timeout.
	SupportBundleExpirationSeconds         = NewSetting(SupportBundleExpirationSecondsSettingName, "1800")
```

Another question: what's the consideration to use different time units here ?

Yu-Jack: Different usage.

- SupportBundleTimeout
  It is used to check if support bundle manager execute too long, such as collecting logs. It might take some times if you have a lots of nodes.
- SupportBundleExpirationSeconds 
  It is used to remove useless support bundle CR after finished collecting logs.

pkg/settings/settings.go
Yu-Jack: I think it's weird to put two const here. I think we could change the `setting.Def` type into `interface{}`. Then you could put anything you want instead of putting two const due to the constraint of `setting.Def`.

pkg/settings/settings.go
w13915984028: Seems no clear reason that the first setting is in `minutes` but the second is in `seconds`, it may confuse the users.

Could `SupportBundleExpiration` also use minutes ?

Yu-Jack: Surely, then I will remove `Seconds` word.

pkg/settings/settings.go
w13915984028:
```
if i, err = strconv.Atoi(s.Default); err != nil {
  return 30 // minutes
}
```

should be fine in this case

Yu-Jack: If I understand correctly, it will affect other setting, right?

Yu-Jack: I think we just keep it returning 0 instead of returning 30 here. Otherwise, it affects other setting and makes some confusion about why we put 30 just only for support-bundle-expiration.

w13915984028: Indeed, 0 is an invalid value for this function, it should never be returned.

In any case, when return 0, the generated SB file will be removed right away, this breaks the feature.

Different settings have different default value and value range. This setting, should be at least e.g. 2 minutes, thus the SB file will have chance to be downloaded.

Recall the SB feature, the priority is:
(1) The SB file can be generated or fail
(2) The SB file can be downloaded or fail
(3) The SB file can be deleted after certain time.

Yu-Jack: But, if you return 30 in this function, it also break other features. For example, [support-bundle-timeout](https://docs.harvesterhci.io/v1.2/advanced/index#support-bundle-timeout) uses 0 as disabling. So, I suggest we need to enhance whole setting structure, otherwise it would be weird case here, like you said 0 is an invalid value. I think it's a different issue here. It's kind of existing setting struct issue.

w13915984028: OK, then please replace 0 in the controller side, revert 0 to 30.

Yu-Jack: Surely, it makes sense for me.

pkg/util/supportbundle/constants.go
```diff
@@ -0,0 +1,7 @@
+package supportbundle
+
+const (
+	// SupportBundleExpirationDefault and SupportBundleExpirationDefaultStr these two value should be changed together.
+	SupportBundleExpirationDefault    = 30
+	SupportBundleExpirationDefaultStr = "30"
+)
```
Yu-Jack, line 7: Due to import cycles, so I move out constants to another folder under utils. These two value should be removed after redesign the setting structure in the future.

w13915984028 (approved):
LGTM, thanks.

btw, please squash the commits before final merge.

pkg/controller/master/supportbundle/controller.go
```diff
@@ -70,12 +73,48 @@ func (h *Handler) OnSupportBundleChanged(key string, sb *harvesterv1.SupportBund
...
 		return sb, nil
 	}
 }
 
+// checkExistTime checks if the support bundle has been updated for more than supportBundleExistTimeLimit minutes
+// If yes, that means the support bundle file is not be retrieved, and it should be deleted.
+func (h *Handler) checkExistTime(sb *harvesterv1.SupportBundle) (*harvesterv1.SupportBundle, error) {
+	t, err := time.Parse(time.RFC3339, harvesterv1.SupportBundleInitialized.GetLastUpdated(sb))
+	if err != nil {
+		logrus.Debugf("[%s] fail to parse %s", sb.Name, sb.Status.Conditions[0].LastUpdateTime)
```
ibrokethecloud: we should probably use `harvesterv1.SupportBundleInitialized.GetLastUpdated(sb)` in the error message, rather than assuming this will always be the first condition. This should ensure that we future proof against any more conditions being added.

pkg/controller/master/supportbundle/controller.go
```diff
@@ -70,12 +73,48 @@ func (h *Handler) OnSupportBundleChanged(key string, sb *harvesterv1.SupportBund
...
+	)
+
+	if expiration := settings.SupportBundleExpiration.GetInt(); expiration == 0 {
+		expiredTime = time.Duration(supportBundleUtil.SupportBundleExpirationDefault) * time.Minute
+	} else {
+		expiredTime = time.Duration(expiration) * time.Minute
+	}
+
+	logrus.Debugf("[%s] support bundle status: %s exist time is %s", sb.Name, sb.Status.State, existTime.String())
+	if existTime < expiredTime {
```
ibrokethecloud, line 105: minor nit, we always requeue after `expiredTime` which will be the expiry duration coming from the setting. If we want to gc the crd earlier we should probably requeue after `expiredTime - existTime`

pkg/controller/master/supportbundle/controller.go
Yu-Jack: I forgot to modify usage in debug message.

ibrokethecloud (approved):
lgtm. thanks.

PR #4948, opened 2024-01-05
added check to reconcile controller revision

ibrokethecloud:

**Problem:**

**Solution:**

**Related Issue:**

**Test plan:**


pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -62,9 +70,62 @@ func (h *VMNetworkController) updateVMDefaultNetworkMacAddress(vmi *kubevirtv1.V
...
+		crObj.Data.Raw = patch
+		crObj.ResourceVersion = ""
+		_, err = h.crClient.Create(crObj)
+	}
+
+	return err
+}
+
+// copied from kubevirt.io/kubevirt/pkg/virt-controller/watch
+type VirtualMachineRevisionData struct {
```
FrankYang0529: We can update code like following and do `go mod tidy` and `go mod vendor`, so we don't need to declare the structure here:

```go
import (
  // ...
  kubevirtwatch "kubevirt.io/kubevirt/pkg/virt-controller/watch"
)

func (h *VMNetworkController) regenerateControllerRevision(vmi *kubevirtv1.VirtualMachineInstance, vm *kubevirtv1.VirtualMachine) error {
  // ...
  revisionSpec := &kubevirtwatch.VirtualMachineRevisionData{}
  // ...
}
```

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -62,9 +70,62 @@ func (h *VMNetworkController) updateVMDefaultNetworkMacAddress(vmi *kubevirtv1.V
 		}
 	}
 
+	if err := h.regenerateControllerRevision(vmi, vm); err != nil {
```
FrankYang0529, line 74: It's better to regenerate controller revision if VM is updated:

```go
	if !reflect.DeepEqual(vm, vmCopy) {
		if err := h.regenerateControllerRevision(vmi, vm); err != nil {
			return err
		}
		if _, err := h.vmClient.Update(vmCopy); err != nil {
			return err
		}
	}
```

FrankYang0529 (changes requested):
Overall LGTM. Thank you!

From KubeVirt source code, it looks like they just use controller revision, but not watching change for controller revision. I think it's safe to regenerate one, but we still need to be careful when we bump KubeVirt version.

pkg/controller/master/virtualmachine/vmi_network_controller.go
ibrokethecloud: this issue only happens when we patch back the mac address since the vmi does not need to be regenerated.

For any other updates for which VMI needs to be recreated the revision ends up being in sync

FrankYang0529 (approved):
LGTM. Thanks.

bk201 (approved):
LGTM, thanks!

ibrokethecloud:
@mergifyio backport v1.2

Issue #4954, opened 2024-01-08
[BUG] VMRestore to new VM doesn't work if there is VMSnapshot


FrankYang0529:
**Describe the bug**
After https://github.com/harvester/harvester/issues/4604, we deny VMRestore request with deletion policy if there is a VMSnapshot. It introduces a new issue - if we don't specify `DeletionPolicy`, the VMRestore can't work when there is a VMSnapshot. The root cause is that we only check whether deletion policy is `retain`, but not empty.

**To Reproduce**
1. Create a VM.
2. Create a VMBackup and VMSnapshot.
3. Restore VMBackup to a new VM and it can't work.

**Expected behavior**
VMBackup to a new VM can work.

**Environment**
 - Harvester ISO version: master-22d9e700-head

**Additional context**
Thanks @TachunLin for finding the issue.

TachunLin:
Verified **fixed** on `master-0c47319a-head` (24/01/10). Close this issue.

### Result
- [x] 1. Can restore vm backup to create a new vm when already setup vm backup and snapshot on it. 

    <details><summary>Test Result $~~$ $\color{green}{\textsf{PASS}}$ </summary>


    https://github.com/harvester/harvester/assets/29251855/e791dc5a-c011-4a2a-9c91-1b0c7d193d24
    
    ![image](https://github.com/harvester/harvester/assets/29251855/390b028a-7f0b-4326-979b-a935ab9159a9)


    </details>


- [x] 1. Can restore vm backup to replace existing vm (retain volume) when already setup vm backup and snapshot on it. 

    <details><summary>Test Result $~~$ $\color{green}{\textsf{PASS}}$ </summary>


    ![image](https://github.com/harvester/harvester/assets/29251855/72f9d169-778e-4df4-824a-93f4f688d098)
    
    
    https://github.com/harvester/harvester/assets/29251855/425347ea-7d23-4835-b14d-bbac73774f75


    </details>

### Test Information
* Test Environment: 3 nodes harvester on provo bare machine
* Harvester version: `master-0c47319a-head` (24/01/10)

### Verify Steps
<details><summary>Execution Steps</summary>

  1. Prepare a 3 nodes Harvester cluster
  2. Create a vm named vm1 
  3. Setup nfs backup target
  4. Create a backup for the vm1
  5. Create a snapshot for the vm1
  6. Restore the backup of vm1 to create a new VM
  7. Check can restore vm correctly
  8. Shutdown vm1, 
[...]

PR #4955, opened 2024-01-08
fix: ignore pvc checking when vmrestore to a new vm

FrankYang0529:
**Problem:**
After https://github.com/harvester/harvester/issues/4604, we deny VMRestore request with deletion policy if there is a VMSnapshot. It introduces a new issue - if we don't specify DeletionPolicy, the VMRestore can't work when there is a VMSnapshot. The root cause is that we only check whether deletion policy is retain, but not empty.

**Solution:**
Check whether it's restored to a new VM.

**Related Issue:**
https://github.com/harvester/harvester/issues/4954

**Test plan:**
1. Create a VM.
2. Create a VMBackup and VMSnapshot.
3. Stop the VM.
4. Restore VMBackup to a new VM.
5. Restore VMRestore to a new VM.

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -404,7 +404,7 @@ func (h *RestoreHandler) initVolumesStatus(vmRestore *harvesterv1.VirtualMachine
 		restoreCpy.Status.VolumeRestores = volumeRestores
 	}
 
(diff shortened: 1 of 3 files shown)


bk201 (approved):

WebberHuang1118 (approved):
LGTM, thanks.

FrankYang0529:
@Mergify backport v1.2

Issue #4467, opened 2023-08-30
[ENHANCEMENT] Add some checking mechanism for the single replica volume


Vicente-Cheng:
**Is your enhancement related to a problem? Please describe.**
In some specific environments, we will upgrade with the single replica volume.
That will cause the draining to hang with this volume (single replica volume).

**Describe the solution you'd like**
We could add some checking mechanisms on the harvester webhook.
Then, let the user know the potential risk before the upgrade.

**Describe alternatives you've considered**
Add the document for the upgrade.

**Additional context**
None

starbops:
We're going to add the corresponding check in the upgrade helper script first, and will eventually implement the check in the harvester-webhook before users initiate the upgrade.

TachunLin:
Verified **fixed** on `v1.3.0-rc1` (24/01/25). Close this issue

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ Confirm the message of single replica volume exists on vm when trying to upgrade $~~$</summary>

When we trigger the upgrade button while a single replicas exists on vm, it would prompt the error message to prevent to procedd
  ```
  admission webhook "validator.harvesterhci.io" denied the request: The backing volumes of the following PVCs have single replica configured, please consider shutdown the corresponding workload before upgrade: default/test-vm-disk-0-tabep
  ```
  ![image](https://github.com/harvester/harvester/assets/29251855/fd8e654c-482f-43e1-b682-f202921a9e12)

 </details>

### Test Information
* Test Environment: 3 nodes harvester on remote kvm machines
* Harvester version: `v1.3.0-rc1` (24/01/25)

### Verify Steps
<details><summary> Confirm the message of single replica volume exists on vm when trying to upgrade </summary>

1. Prepare a Harvester cluster with 2+ nodes (installed with the version including the commit of this PR)
2. Create a new VM, say, `test-vm`
3. Shutdown the newly created VM
4. Find the LH volume of the VM and adjust its replica count in the field `.spec.numberOfReplicas` to 1
  ```
  harvester-node-0:~ # kubectl get volumes -A
  NAMESPACE         NAME                                       DATA ENGINE   STATE      ROBUSTNESS   SCHEDULED   SIZE          NODE               AGE
  longhorn-system   pvc-c645dca6-268c-44c1-8f64-73aee7a17216   v1            attached   healthy                  10737418240   harvester-node-2   3m9s
[...]

PR #4956, opened 2024-01-08
feat(webhook): add single-replica volume check for upgrade

starbops:

**Problem:**

An LH volume's `.spec.numberOfReplicas` is sometimes configured as `1`. Either intentionally by the user or derived from a StorageClass with replicas set to 1. In such cases, the volumes are in a `healthy` state but are fragile, especially for the Harvester upgrade. Harvester upgrade will try to drain each node in the cluster in a later step. The replica instance manager Pod cannot be drained if the volume it manages has one and only one replica. The recommended way is to shut down the workload of those using the volume and try to start the upgrade again.

**Solution:**

Add the single-replica volume check in the validating admission webhook. If such circumstances are detected, reject the creation of the Upgrade object.

**Related Issue:**

#4467 

**Test plan:**

1. Prepare a Harvester cluster with 2+ nodes (installed with the version including the commit of this PR)
2. Create a new VM, say, `test-vm`
3. Shutdown the newly created VM
4. Find the LH volume of the VM and adjust its replica count in the field `.spec.numberOfReplicas` to `1`
5. Start the VM
6. Create the Version CR for upgrade (to whatever Harvester version, it does not matter)
7. When kick off the upgrade by clicking the "Upgrade" button, a webhook error message should show up like below:

image

The error message should specify the relevant PVCs' namespace and name.

### DIFF
--- a/pkg/webhook/resources/upgrade/validator.go
+++ b/pkg/webhook/resources/upgrade/validator.go
@@ -30,6 +30,7 @@ import (
 	ctllhv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
 	"github.com/harvester/harvester/pkg/util"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
+	"github.com/harvester/harvester/pkg/webhook/indexeres"
 	versionWebhook "github.com/harvester/harvester/pkg/webhook/resources/version"
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
@@ -163,7 +164,11 @@ func (v *upgradeValidator) checkResources(version *v1beta1.Version) error {
 		return err
 	}
 
-	return v.checkMachines()
+	if err := v.checkMachines(); err != nil {
+		return err
+	}
+
+	return v.checkSingleReplicaVolumes()
 }
 
 func (v *upgradeValidator) hasDegradedVolume() (bool, error) {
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/upgrade/validator.go
```diff
@@ -307,6 +311,38 @@ func (v *upgradeValidator) checkMachines() error {
...
+	}
+
+	for _, volume := range volumes {
+		if volume.Spec.NumberOfReplicas == 1 {
+			singleReplicaVolumeNames = append(singleReplicaVolumeNames, volume.Namespace+"/"+volume.Name)
+		}
+	}
+
+	if len(singleReplicaVolumeNames) > 0 {
+		return werror.NewInternalError(fmt.Sprintf("single-relica volumes (%s) running, please shutdown the correspondingn workload before upgrade", strings.Join(singleReplicaVolumeNames, ", ")))
```
bk201:
```suggestion
		return werror.NewInternalError(fmt.Sprintf("single-replica volumes (%s) running, please shutdown the corresponding workload before upgrade", strings.Join(singleReplicaVolumeNames, ", ")))
```

bk201: Instead of displaying Longhorn volumes here, I suggest we display pvc of volumes. It shows the correct namespace and volume name to users. Also, we can ignore detached volumes.

starbops: Updated, thank you.

starbops: Updated, thank you!

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -307,6 +312,44 @@ func (v *upgradeValidator) checkMachines() error {
...
+	// Find all single-replica volumes
+	singleReplicaVolumes, err := v.lhVolumes.GetByIndex(indexeres.VolumeByReplicaCountIndex, "1")
+	if err != nil {
+		return err
+	}
+
+	volumeNames := make([]string, 0, len(singleReplicaVolumes))
+	for _, volume := range singleReplicaVolumes {
+		// Ignore detached volumes
+		if volume.Status.State == "detached" {
```
Vicente-Cheng: nit: should we also skip the `deleting`?

Vicente-Cheng (approved):
just a nit, others lgtm.

pkg/webhook/resources/upgrade/validator.go
starbops: Good call! I guess we only need to take care of creating, attached, and attaching?

```go
const (
	VolumeStateCreating  = VolumeState("creating")
	VolumeStateAttached  = VolumeState("attached")
	VolumeStateDetached  = VolumeState("detached")
	VolumeStateAttaching = VolumeState("attaching")
	VolumeStateDetaching = VolumeState("detaching")
	VolumeStateDeleting  = VolumeState("deleting")
)
```

Vicente-Cheng: Yes, I think so. `detaching`, `detached`, and `deleting` could be ignored.

starbops: Updated and PTAL, thanks!

Vicente-Cheng: Looks good! Thanks!!

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -307,6 +312,43 @@ func (v *upgradeValidator) checkMachines() error {
...
+		switch volume.Status.State {
+		case lhv1beta2.VolumeStateCreating, lhv1beta2.VolumeStateAttached, lhv1beta2.VolumeStateAttaching:
+			pvcNamespace := volume.Status.KubernetesStatus.Namespace
+			pvcName := volume.Status.KubernetesStatus.PVCName
+			volumeNames = append(volumeNames, pvcNamespace+"/"+pvcName)
+		}
+	}
+
+	if len(volumeNames) > 0 {
+		return werror.NewInternalError(
```
bk201: Let's use NewInvalidError here, seems to make more sense here (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422)

starbops: Updated, thanks!

bk201 (approved):
LGTM, thanks!

Issue #3541, opened 2023-02-24
[BUG] backup restore does not carry the same MAC address


abonillabeeche:
**Describe the bug**
When restoring a VM from backup, the MAC address isn't the same.

**To Reproduce**
1. Perform a VM backup on a harvester cluster
2. Power off the VM
3. Restore the VM on another cluster
4. The VM will have a different MAC address

**Expected behavior**
The new VM definition should be exact to the original backup

w13915984028:
Together with https://github.com/harvester/harvester/issues/3682, https://github.com/harvester/harvester/issues/3602.

@FrankYang0529 any idea of this issue, is it lost or stripped ? thanks.

FrankYang0529:
The `MacAddress` field is kept in VMBackup object, but we remove it when we restore a VM: https://github.com/harvester/harvester/blob/398d492b81aa6fcc60bdea3d270f501bd5e1eea9/pkg/controller/master/backup/restore.go#L570-L573

We may need to add an option to use the same `MacAddress`.

w13915984028:
When the restored VM is targeting for an replacement of the original VM, the MAC address seems better to be kept.

But if the backup can be restored into a couple of VMs, then the MAC address cannot be identical.

An option to keep/drop MAC, will be a good enhancement.

albinsun:
Provide test plan.
Let's test after UI side backport.
Ref. dashboard/pull/957 [Add KeepMacAddress checkbox to Backup restore page](https://github.com/harvester/dashboard/pull/957#top)

# Test Plan
## Environment
* Harvester
  * Version: `v1.3-071de3aa-head`
  * Profile: QEMU/KVM, 3 nodes w/ witness (8C/16G/500G)
  * ui-source: `Auto`

## Test Suites
### Prerequisites
1. <details>
   <summary> Create a VM
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/289cf4b2-05cc-4492-a52e-6a081ddfe536)
   </details>

1. <details>
   <summary> Take VM backup
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/f676f7fc-c30e-4692-a206-ee9777173868)
   </details>

### Restore New
| Existence \ `KeepMac` | `True`                   |  `False`                 |
|:---------------------:|:------------------------:|:------------------------:|
|       `True`          |  Deny (TC1)              | Yes w/ **new** MAC (TC2) |
[...]

PR #4957, opened 2024-01-08
feat: add KeepMacAddress to VMRestore.Spec

FrankYang0529:
**Problem:**
The system removes macaddress by default when users restore a new VM.

**Solution:**
Add a option to support keeping macaddress.

**Related Issue:**
https://github.com/harvester/harvester/issues/3541

**Test plan:**
1. Create a VM.
2. Create a VMBackup.
3. Restore the VMBackup to a new VM with `keepMacAddress=true`. The webhook should deny the request because the source VM is still existent.
4. Remove the source VM.
5. Restore the VMBackup to a new VM with `keepMacAddress=true`. The new VM should have same macaddress as source VM.
6. Restore the VMBackup to a new VM with `keepMacAddress=true` again. The webhook should deny the request because the another VMRestore with `keepMacAddress=true` is existent.

### DIFF
--- a/pkg/webhook/resources/virtualmachinerestore/validator.go
+++ b/pkg/webhook/resources/virtualmachinerestore/validator.go
@@ -32,6 +32,7 @@ const (
 	fieldTargetName               = "spec.target.name"
 	fieldVirtualMachineBackupName = "spec.virtualMachineBackupName"
 	fieldNewVM                    = "spec.newVM"
+	fieldKeepMacAddress           = "spec.keepMacAddress"
 )
 
 func NewValidator(
@@ -85,10 +86,7 @@ func (v *restoreValidator) Resource() types.Resource {
 func (v *restoreValidator) Create(_ *types.Request, newObj runtime.Object) error {
 	newRestore := newObj.(*v1beta1.VirtualMachineRestore)
 
-	targetVM := newRestore.Spec.Target.Name
-	newVM := newRestore.Spec.NewVM
-
-	if errs := validationutil.IsDNS1123Subdomain(targetVM); len(errs) != 0 {
+	if errs := validationutil.IsDNS1123Subdomain(newRestore.Spec.Target.Name); len(errs) != 0 {
 		return werror.NewInvalidError(fmt.Sprintf("Target VM name is invalid, err: %v", errs), fieldTargetName)
 	}
 
@@ -101,29 +99,11 @@ func (v *restoreValidator) Create(_ *types.Request, newObj runtime.Object) error
 		return werror.NewInvalidError(err.Error(), fieldVirtualMachineBackupName)
 	}
 
-	if vmBackup.Spec.Type == v1beta1.Backup {
-		err = v.checkBackup(newRestore, vmBackup)
-	} else {
-		err = v.checkSnapshot(newRestore, vmBackup)
-	}
-	if err != nil {
-		return werror.NewInvalidError(err.Error(), fieldVirtualMachineBackupName)
+	if err := v.checkVMBackupType(newRestore, vmBackup); err != nil {
+		return err
 	}
 
-	vm, err := v.vms.Get(newRestore.Namespace, targetVM)
-	if err != nil {
-		if newVM && apierrors.IsNotFound(err) {
-			return v.handleNewVM(newRestore, targetVM, vmBackup)
-		}
-		return werror.NewInvalidError(err.Error(), fieldTargetName)
-	}
-
-	// restore a new vm but the vm is already exist
-	if newVM && vm != nil {
-		return werror.NewInvalidError(fmt.Sprintf("VM %s is already exists", vm.Name), fieldNewVM)
-	}
-
-	return v.handleExistVM(newVM, vm)
+	return v.checkNewVMField(newRestore, vmBackup)
 }
 
 func (v *restoreValidator) Update(_ *types.Request, oldObj, newObj runtime.Object) error {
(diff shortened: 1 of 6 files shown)


pkg/webhook/resources/virtualmachinerestore/validator.go
```diff
@@ -169,6 +172,27 @@ func (v *restoreValidator) handleNewVM(newRestore *v1beta1.VirtualMachineRestore
...
+			return werror.NewInvalidError(fmt.Sprintf("can't restore the new vm with the same macaddress as the vmrestore %s/%s", otherVMRestore.Namespace, otherVMRestore.Name), fieldKeepMacAddress)
+		}
+	}
+	sourceVM, err := v.vms.Get(vmBackup.Namespace, vmBackup.Spec.Source.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return werror.NewInternalError(fmt.Sprintf("failed to get the VM %s/%s, err: %+v", vmBackup.Namespace, vmBackup.Spec.Source.Name, err))
+	}
+	if sourceVM != nil && vmBackup.Status.SourceUID != nil && sourceVM.UID == *vmBackup.Status.SourceUID {
+		return werror.NewInvalidError("can't restore the new vm with the same macaddress because the source vm is still existent", fieldKeepMacAddress)
+	}
```
WebberHuang1118, line 201: In the real case, is `sourceVM.UID != *vmBackup.Status.SourceUID` possible?

WebberHuang1118: Just a suggestion for readability:
```
	sourceVM, err := v.vms.Get(vmBackup.Namespace, vmBackup.Spec.Source.Name)
		return nil
	}
	if err != nil {
		return werror.NewInternalError(fmt.Sprintf("failed to get the VM %s/%s, err: %+v", vmBackup.Namespace, vmBackup.Spec.Source.Name, err))
	}
	if sourceVM != nil && vmBackup.Status.SourceUID != nil && sourceVM.UID == *vmBackup.Status.SourceUID {
		return werror.NewInvalidError("can't restore the new vm with the same macaddress because the source vm is still existent", fieldKeepMacAddress)
	}
```

pkg/webhook/resources/virtualmachinerestore/validator.go
```diff
@@ -169,6 +172,27 @@ func (v *restoreValidator) handleNewVM(newRestore *v1beta1.VirtualMachineRestore
...
+
 	return nil
```
WebberHuang1118: Just a suggestion for readability:
```
	sourceVM, err := v.vms.Get(vmBackup.Namespace, vmBackup.Spec.Source.Name)
		return nil
	}
	if err != nil {
		return werror.NewInternalError(fmt.Sprintf("failed to get the VM %s/%s, err: %+v", vmBackup.Namespace, vmBackup.Spec.Source.Name, err))
	}
	if sourceVM != nil && vmBackup.Status.SourceUID != nil && sourceVM.UID == *vmBackup.Status.SourceUID {
		return werror.NewInvalidError("can't restore the new vm with the same macaddress because the source vm is still existent", fieldKeepMacAddress)
	}
```

pkg/webhook/resources/virtualmachinerestore/validator.go
FrankYang0529: Yes, like following case:

1. Create a VM.
2. Create a VMBackup.
3. Delete the VM.
4. Create a VM with the same name again. In this case, same VM names may have different UID.

pkg/webhook/resources/virtualmachinerestore/validator.go
FrankYang0529: Updated it. Thank you.

ibrokethecloud (approved):
lgtm. thanks.

WebberHuang1118:
After applying this PR, the restoring VM didn't start automatically, I'm still investigating it.

pkg/webhook/resources/virtualmachinerestore/validator.go
```diff
@@ -169,6 +172,30 @@ func (v *restoreValidator) handleNewVM(newRestore *v1beta1.VirtualMachineRestore
...
+	sourceVM, err := v.vms.Get(vmBackup.Namespace, vmBackup.Spec.Source.Name)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("failed to get the VM %s/%s, err: %+v", vmBackup.Namespace, vmBackup.Spec.Source.Name, err))
+	}
+	if sourceVM != nil && vmBackup.Status.SourceUID != nil && sourceVM.UID == *vmBackup.Status.SourceUID {
+		return werror.NewInvalidError("can't restore the new vm with the same macaddress because the source vm is still existent", fieldKeepMacAddress)
+	}
```
WebberHuang1118, line 201: I found during the VMRestore with `.spec.KeepMacAddress` is false, and the source VM exists, the VMRestore will be rejected by `can't restore the new vm with the same macaddress because the source vm is still existent`

FrankYang0529: Hi Webber, thanks for finding the bug. I add a check for `.spec.KeepMacAddress`. If it's `false`, skip the check.

FrankYang0529:
> After applying this PR, the restoring VM didn't start automatically, I'm still investigating it.

Probably, we miss some update in master-head like the new longhorn-manager package. I just rebased it. Thanks.

WebberHuang1118 (approved):
Testing cases work as expected,
LGTM, thanks.

connorkuehl (approved):

connorkuehl:
@mergify backport v1.3

Issue #4428, opened 2023-08-18
[FEATURE] VM network DHCP


starbops:
**Is your feature request related to a problem? Please describe.**

Currently, the network configuration of the VMs relies on external DHCP servers or the cloud-init configuration associated with the VM itself, which feels cluttered in terms of user experience and has limitations.

For example, creating a guest cluster on Harvester with Rancher integration requires all the backing VMs on the same VM network. Since we don't have a DHCP service on that VM network, it relies on users to run/leverage their own DHCP server outside of the Harvester cluster. It's possible that the IP addresses of the VMs change after a while if the configuration of the DHCP server isn't set carefully. This will undoubtedly break the guest cluster.

**Describe the solution you'd like**

Providing DHCP service on VM networks to the VMs. Defining new CRDs or leveraging existing ones to maintain the IP address configuration and leases. Implementing a custom controller reconciles the CRs and reflects changes to the DHCP service.

The scope should be restricted to:

- the VM networks with the DHCP feature enabled,
- the VMs associated with the DHCP-enabled VM networks running on the Harvester cluster, not the machines outside of the Harvester cluster

The above essentially means we should do it with the static configuration, a.k.a fixed IPs, which implies we must allocate IP addresses for each VM before the DHCP service can serve the IP addresses out.

**Describe alternatives you've considered**

We may integrate existing implementations of DHCP servers like [ISC Kea](https://www.isc.org/kea/) or [CoreDHCP](https://github.com/coredhcp/coredhcp). The custom controller is only responsible for communicating with them using API calls or updating the configurations and then reloading the daemon.

**Additional context**

Issues that might be (partially) solved if this feature is delivered:

#1470
#2099
#3682

starbops:
The current implementation can be found in [harvester/vm-dhcp-controller](https://github.com/harvester/vm-dhcp-controller).

starbops:
## Test Plans

- Prerequisites
    1. Prepare a Harvester cluster with 2+ nodes in v1.3.0-rc*
        - Each node should have two NICs (one for `mgmt`, the other for day-2 ClusterNetwork creation)
    2. Download the kubeconfig file (some operations need us to interact with the api server directly)
    3. Install the `harvester-vm-dhcp-controller` add-on with the following command
        ```
        kubectl apply -f https://raw.githubusercontent.com/harvester/experimental-addons/main/harvester-vm-dhcp-controller/harvester-vm-dhcp-controller.yaml
        ```
[...]

PR #4960, opened 2024-01-09
doc(hep): add managed dhcp hep [CI SKIP]

starbops:
The proposal of the [managed DHCP feature](https://github.com/harvester/vm-dhcp-controller)

**Related Issue:**

#4428


bk201:
Please check https://github.com/harvester/vm-dhcp-controller for the implementation.

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
+# Managed DHCP
+
+## Summary
+
+We want a mechanism for VMs attached to VM networks to have network connectivity out of the box, and such funtionality is provided by Harvester itself, not external services.
```
w13915984028: From the discussion with customer, if I am understanding correctly:

This feature is more an IP helper, it helps to allocate IP/gateway information to VMs and withdraw IPs quickly;  but the IP range allocation, the connectivity etc. are still sured by the infrastructure layer.
 
Harvester has no embedded router to route traffic between different L2 networks, the VxLan is also not supported yet. It is hard for Harvester to ensure the network connectivity independently now.
 
Due to those limitations, suggest to update the description of the functinality and motivation of this feature.

starbops: Thanks for the suggestion. I've updated the paragraph; I hope that helps clarify the intention of the feature and the problem we want to solve.

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+    ```
+
+4. Check the following:
+
+    - The `default-priv-net-cp-agent` agent Pod should run on one of the control plane nodes
+    - The two IPPool objects' `.status` field should contain the correct allocation records and statistics
+    - The VirtualMachineNetworkConfig object's `.status` field should show the two NICs have IP addresses allocated
+    - The VM `test-vm-1` should have the correct IP addresses configured automatically via DHCP from the agents respectively
+
+#### Kubernetes Guest Clusters (Rancher Integration)
```
ibrokethecloud, line 266: do we need changes to the machine driver to pass additional info to VM for leveraging the managed dhcp feature?

starbops: vm-dhcp-controller listens to VMI-related events and creates the corresponding vmnetcfg object; then, the IP address can be allocated from the pool by the vmnetcfg reconcile loop. This is only for the dynamic IP allocation scenario. As @ibrokethecloud suggested, we need other ways to support the static IP allocation if the user tends to configure the VM's IP address statically during the VM creation. This is tracked in #5052. Thank you!

ibrokethecloud (approved):
lgtm. thanks.

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+- Provide network information as a general DHCP service to machines outside of Harvester premises
+- NAT or route the internal traffic to the outside world
+
+## Proposal
+
+### User Stories
+
+#### Kubernetes Guest Clusters (Rancher Integration)
+
+Creating a Kubernetes guest cluster from Rancher using the Harvester node driver has a premise that the VMs consisting of the cluster should have network connectivity without human intervention. That is to say, those VMs need to be configured with valid IP addresses and other network configurations like the default gateway, DNS server, NTP server, etc. Usually, people set up an external DHCP server to provide network information to the VMs. This requires skilled users who are familiar with how to install and configure a DHCP server correctly, needless to say, a good understanding of the network configuration of their premises (how servers and switches are connected, what VLANs are provisioned for the Harvester nodes, etc.)
```
w13915984028: a good understanding of the network configuration of their premises
->
a good understanding of the network configuration of their premises is required.

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+
+Removing the VM will cause the associated VirtualMachineNetworkConfig object to be purged. Users can remove it directly without removing the VM. The static DHCP lease for the VM's network interfaces is removed. After the lease time, the VM loses the IP address. The same logic applies to the IPPool objects: removing the VM network (NetworkAttachmentDefinition) will cause the associated IPPool object to be purged. (Note: The IPPool object cannot be removed if there are still VirtualMachineNetworkConfig objects associated with it.) Users can remove it directly without removing the VM network. In that case, the VM network becomes a purely L2 network, just like we used to see.
```
w13915984028: In which situation, the `VitualMachineNetworkConfig` can be removed and `VM` is still there? Are there preconditions that, the VM changed it's network strategy from `managed dhcp` to others first?

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Goals
+
+- Provide network information (subnets, gateways, and IP addresses) to the VMs running on top of Harvester clusters
+- Improve the user experience of private networks in Harvester
+
+### Non-goals
+
+- Provide network information as a general DHCP service to machines outside of Harvester premises
+- NAT or route the internal traffic to the outside world
```
w13915984028, line 26: DHCP relay, proxy are both not supported.

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+	NTP []string `json:"ntp,omitempty"`
+	// +optional
+	LeaseTime *int `json:"leaseTime,omitempty"`
+}
+
+type Pool struct {
+	Start string `json:"start,omitempty"`
+	End   string `json:"end,omitempty"`
+	// +optional
+	Exclude []string `json:"exclude,omitempty"`
```
w13915984028, line 109: refer https://wiki.libvirt.org/VirtualNetworking.html#xml-format
```
<ip address="192.168.122.1" netmask="255.255.255.0">
 <dhcp>
  <range start="192.168.122.100" end="192.168.122.254" />
  <host mac="00:16:3e:e2:ed" name="foo.example.com" ip="192.168.122.10" />
 </dhcp>
</ip>
```

The static-ip binding, is normally configured in the `pool`, so that there won't be conflicts.

Now it is in `VirtualMachineNetworkConfig`, it will have such kind of races:
(1) mutli vm set same static ip binding
(2) the static IP is potentially conflict with the `pool`'s dynamic IP, even the webhook can't detect this effectively

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+
+    ```yaml
+    apiVersion: network.harvesterhci.io/v1alpha1
+    kind: IPPool
+    metadata:
+      name: priv-net-all
+      namespace: default
+    spec:
+      ipv4Config:
+        serverIP: 172.19.150.2
```
w13915984028, line 228: It assumes that the `serverIP` is in same range with next item `cidr`, is this checked by the webhook?

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+        start: 172.19.100.10
+        end: 172.19.100.14
+        leaseTime: 300
+      networkName: default/priv-net-cp
+    ```
+
+3. Create a VirtualMachineNetworkConfig object for the VM `test-vm-1`:
+
+    ```yaml
+    apiVersion: network.harvesterhci.io/v1alpha1
```
w13915984028: The `VirtualMachineNetworkConfig` looks to be redundant with kubevirt vm config `spec.domain.devices.interfaces`, does the user need to config vm MAC address in two places?

```
type Interface struct {
	// Logical name of the interface as well as a reference to the associated networks.
	// Must match the Name of a Network.
	Name string `json:"name"`
	// Interface model.
	// One of: e1000, e1000e, ne2k_pci, pcnet, rtl8139, virtio.
	// Defaults to virtio.
	// TODO:(ihar) switch to enums once opengen-api supports them. See: https://github.com/kubernetes/kube-openapi/issues/51
	Model string `json:"model,omitempty"`
	// BindingMethod specifies the method which will be used to connect the interface to the guest.
	// Defaults to Bridge.
	InterfaceBindingMethod `json:",inline"`
	// List of ports to be forwarded to the virtual machine.
	Ports []Port `json:"ports,omitempty"`
	// Interface MAC address. For example: de:ad:00:00:be:af or DE-AD-00-00-BE-AF.
	MacAddress string `json:"macAddress,omitempty"`
	// BootOrder is an integer value > 0, used to determine ordering of boot devices.
	// Lower values take precedence.
	// Each interface or disk that has a boot order must have a unique value.
	// Interfaces without a boot order are not tried.
	// +optional
```

A straight-forward idea is, config the `macaddress` in vm spec; config `mac-ip` binding in the pool, then the VM can get the desired ip from the pool.

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+
+## Design
+
+Main components:
+
+- Control plane
+  - `vm-dhcp-controller` is majorly responsible for reconciling IPPool and VirtualMachineNetworkConfg objects and maintaining the IPAM cache
+  - `vm-dhcp-webhook` guards the incoming requests for creating, updating, and deleting custom resources
+- Data plane
+  - `vm-dhcp-agent` is scattered in each VM network (defined by NAD), accounting for syncing the pool cache from the IPPool object it monitors and handling the underlying DHCP requests from VMs
```
w13915984028, line 184: When `vm-dhcp-agent` is running in an indepedent POD, please give a test data of the resource consumption of this POD. If many VM networks are defined, the resources consumption will be something worthy to notice in advance.

w13915984028 (commented):
Some questions, thanks.

enhancements/20230830-managed-dhcp.md
starbops: Yes, the current implementation is not tightly coupled with Harvester yet. If a VM is removed, the corresponding auto-generated vmnetcfg object will be removed, too (due to OwnerReference). But if a vmnetcfg object is removed, only the IP address will be released. This is somewhat like deleting the DHCP lease on the DHCP server. The VM will no longer get an IP address from the managed DHCP service.

I'm not quite sure if I understand your second question correctly. If the user wants to opt out of the managed DHCP service, they could **pause** it in the auto-generated vmnetcfg object or just simply remove the object.

enhancements/20230830-managed-dhcp.md
starbops: If a user deliberately crafted two vmnetcfg objects with the same designated IP address, only one of them could allocate it successfully (assuming the IP address was not in use beforehand). The other one will not get the same IP address; it will log down the error message in the condition field:

```
...
spec:
  networkConfig:
  - ipAddress: 192.168.48.82
    macAddress: 00:1a:2b:3c:4d:5e
    networkName: default/net-48
  vmName: test-vm-2
status:
  conditions:
  - lastUpdateTime: "2024-01-30T13:24:17Z"
    message: designated ip 192.168.48.82 is already allocated
    reason: Error
    status: "False"
    type: Allocated
  - lastUpdateTime: "2024-01-30T13:24:17Z"
    status: "False"
    type: Disabled
```

The IP allocation process happens on the control plane. The controller's mission is to determine what IP to allocate and ensure no double allocation. The allocation is recorded in the ippool object's status field. It could be taken as DHCP's lease database, and it will be the single source of truth of the allocation status of that specific subnet. So, it's possible to implement a validator in webhook to check whether an IP is allocated or not before creating the vmnetcfg object with a designated IP address (it hasn't been implemented yet).

enhancements/20230830-managed-dhcp.md
starbops: Good catch! It should be protected by the validator while creating. Also, the controller needs to handle the error case correctly. I will create an issue to track/fix it. Thank you.

Update: #5065 created

enhancements/20230830-managed-dhcp.md
starbops: Configuring the MAC address for both the VM and the vmnetcfg object is unnecessary, as the controller will monitor the VMI object to retrieve the VM's MAC address and use it to create the corresponding vmnetcfg object. This is the dynamic-bind scenario.

For the static-bind scenario, it will need a deeper integration with Harvester. This is tracked in #5052.

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+	Available int               `json:"available"`
+}
+
+type PodReference struct {
+	Namespace string `json:"namespace,omitempty"`
+	Name      string `json:"name,omitempty"`
+}
+```
+
+**VirtualMachineNetworkConfig** is the essentially the static lease of a individual VM.
```
w13915984028: Please add comment to this object:

`VirtualMachineNetworkConfig` is an intermediate object, it is automatically generated/removed by the controller, users do not need to operate it manually.

w13915984028 (commented):
After reading the source code (some questions about the code are in https://github.com/harvester/harvester/issues/5072), would suggest to clarify below in the document, for better understanding of the architecture.

General speaking, a VM can obtain IP via:
(1) Static IP mode
(2) DHCP mode

for static IP mode, it is via cloud-init, or even manually configuration from console of the VM.

for DHCP mode, we have further DHCP server option:
(I) Indepedent DHCP server, out of Harvester.
(II) Managed DHCP server, embedded in Harvester.

No matter which DHCP server, there are also mainly two ways:
(A) Static IP-MAC binding
(B) Dynamic allocation


For managed DHCP server, we can achive (B) at the moment,

The `networkv1.IPPool` is for a subnet config.
The `VirtualMachineNetworkConfig` is an intermediate object, it is generated after the `vm` object is created.

(A) seems not supportted yet, but it is quite common in such scenario:
User plans a couple of MAC/IP for a group of VM in advance, and/or the VMs are frequently created/detroied. This is useful in auto deployment.

The solution will be
(1) Like https://wiki.libvirt.org/VirtualNetworking.html#xml-format, allow user to configure the `MAC/IP` binding in the `networkv1.IPPool` object in advance.

(2) Allow user to configure `VirtualMachineNetworkConfig` manually in advance. The `VirtualMachineNetworkConfig` needs to extend to have information like `static/dynamic`.

When we stick to treat `VirtualMachineNetworkConfig` as passive objects, then (1) will be a choice.



note: https://github.com/harvester/harvester/issues/5052 already tracks new requirement of static mode.

enhancements/20230830-managed-dhcp.md
starbops: Yes, we hope that users don't bother to create the vmnetcfg objects for VMs but still have the chance to manipulate them manually. For example, if the user wants to temporarily opt out of managed DHCP service for a specific VM, they could "pause" the corresponding vmnetcfg object to make the embedded DHCP server not serve the IP address to the VM. Another example is the static IP-MAC binding case; the user could create the vmnetcfg object with IP-MAC pre-configured and then use the same MAC address to create the VM. The newly created VM will get the desired IP address from the pool (if no error occurs).

As you pointed out, as the integration with Harvester becomes more complete, the vmnetcfg objects will gradually become secondary. There will be no need to operate those objects directly.

enhancements/20230830-managed-dhcp.md
```diff
@@ -0,0 +1,318 @@
...
+
+Main components:
+
+- Control plane
+  - `vm-dhcp-controller` is majorly responsible for reconciling IPPool and VirtualMachineNetworkConfg objects and maintaining the IPAM cache
+  - `vm-dhcp-webhook` guards the incoming requests for creating, updating, and deleting custom resources
+- Data plane
+  - `vm-dhcp-agent` is scattered in each VM network (defined by NAD), accounting for syncing the pool cache from the IPPool object it monitors and handling the underlying DHCP requests from VMs
+
+![System Architecture](20230830-vm-network-dhcp/system-architecture.svg)
```
w13915984028, line 186: Please check https://github.com/harvester/harvester/issues/5072#issuecomment-1920653841

The controller local data `IPAllocator` and `CacheAllocator` are anti-k8s-controller patten.

w13915984028:
Per the concept of SDN, we can enhance and simplify the architecture:

When the producer generates a new dynamic VMNetworkConfig, it adds the timestamp and lease time, the consumer will check this timestamp continuously, if there is no update to renew/lease time, e.g. VMI does not boot, the consumer will withdraw the IP after lease. The control plane itself is safely working.

Agent only allocates an IP when it can get it from VMNetworkConfig, this will by nature hold the VM dhcp requesting. A few times of DHCP renew updates missing (Agent is not ready/ appdb is not ready) normally will not cause DHCP lease due right now.

image

starbops:
The architecture looks interesting. I will certainly take a look and see how we can proceed in the future. But please note that the renewal and lease timeout have not been implemented in the current version of vm-dhcp-controller. Thank you.

w13915984028 (approved):
LGTM, it is fine for v1.3.0, thanks for this important new feature.

bk201 (approved):

mingshuoqiu (approved):
LGTM.

Issue #4965, opened 2024-01-09
[BUG] Upgrade workaround for rancher-system-agent does not get cleaned up


starbops:
**Describe the bug**

In #4149, we added a workaround for **not** restarting the RKE2 server/agent on the nodes during the upgrade. Though we have thought of the cleanup procedure, there are still loose ends...

When an upgrade fails at a later phase (specifically, after the apply manifest phase has been completed), the workaround above is not cleaned up. This is because the cleanup code will be executed in the node upgrade phase for each node. So the workaround, i.e., the `override.conf` for the `rancher-system-agent` system service, remains on the nodes. That will cause the later upgrade to get into trouble.

**To Reproduce**

1. Upgrade from v1.1.2 to v1.2.1
2. Intentionally break the upgrade by removing the Upgrade CR when the upgrade progress goes beyond phase 3 (upgrading system service)
3. Verify that there are two files remain on the nodes under `/run/systemd/system/rancher-system-agent.service.d/`:
    - `override.conf`
    - `10-harvester-upgrade.env`

**Expected behavior**

The files generated by the workaround shouldn't exist after the upgrade, even if the upgrade failed.

**Support bundle**

**Environment**
 - Harvester ISO version: upgrade from v1.1.2 to v1.2.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

The removal of the workaround could be applied more widely, not just restricted to versions before v1.2.0:
https://github.com/harvester/harvester/blob/22d9e700e0e05231fd2b610584da43c0a3ce6fd9/package/upgrade/upgrade_node.sh#L279-L283

The workaround is to remove the `/run/systemd/system/rancher-system-agent.service.d/` directory, daemon reload, and restart `rancher-system-agent` on each node before start over an upgrade.

starbops:
The plan is to persist the workaround as a part of the regular upgrade flow. We don't want the RKE2 server/agent to accidentally restart during the Rancher upgrade, so this is a way to protect them. That is to say, we'll remove the version checking for installing and remove the `rancher-system-agent` system service files.

albinsun:
Provide test result on `v1.3-4372cbc4-head`, close as fixed.


## Environment
* Harvester
  * Version: `v1.3-4372cbc4-head`
  * Profile: QEMU/KVM, 2 nodes (8C/16G/500G)
  * ui-source: `Auto`


## Steps 
1. Setup 2 node harvester cluster

[...]

PR #4966, opened 2024-01-10
fix(upgrade): remove version check for rancher-system-agent workaround

starbops:
**Problem:**

Previously, the workaround applied to `rancher-system-agent` for not restarting RKE2 server/agent remains on the nodes if an upgrade failed after upgrading Harvester charts (in phase 3). This is problematic if an upgrade from v1.1.x to v1.2.x failed, and the user want to start over the upgrade again. This will effectively become an upgrade from v1.2.x to v1.2.x, and the workaround will not be removed in the pre-drain phase because of the version checking.

**Solution:**

We need RKE2 server/agent to not restart during Rancher upgrade. This is true no matter what version we're upgrading from/to, so changing the workaround applied to rancher-system-agent to be a part of the regular upgrade flow by removing the version checking.

**Related Issue:**

#4965 

**Test plan:**

1. Install Harvester v1.1.2
2. Upgrade to v1.2.1
3. Monitor the upgrade progress, when the Harvester version becomes v1.2.1, remove the Upgrade CR (to break the upgrade)
4. Clean up the mess (if any)
5. Start the upgrade again (upgrade from v1.2.1 to v1.2.1)
6. The upgrade ends up successfully
7. There should be **no** `/run/systemd/system/rancher-system-agent.service.d/` directory on the nodes

---

For QA engineers, the test plan would be:

1. Install Harvester v1.3-head
2. Upgrade to v1.3.0-rc3
3. Intentionally break the upgrade when the Harvester version becomes the new version (by removing the Upgrade CR)
4. Clean up the mess if there are any unsettled bundles
5. Start the upgrade again
6. The upgrade should end successfully
7. There should be **no** `/run/systemd/system/rancher-system-agent.service.d/` directory on the nodes

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -1126,12 +1126,6 @@ skip_restart_rancher_system_agent() {
   # by adding an env var to temporarily make rancher-system-agent on each node skip restarting rke2-server/agent.
   # issue link: https://github.com/rancher/rancher/issues/41965
 
-  # only versions before v1.2.0 that upgrading to v1.2.0 need this workaround
-  if [[ ! "${UPGRADE_PREVIOUS_VERSION%%-rc*}" < "v1.2.0" ]]; then
-    echo "Only versions before v1.2.0 need this patch."
-    return
-  fi
-
   plan_manifest="$(mktemp --suffix=.yaml)"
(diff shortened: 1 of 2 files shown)


bk201 (approved):

ibrokethecloud (approved):
lgtm. thanks.

starbops:
@mergifyio backport v1.2

Issue #4840, opened 2023-12-08
[ENHANCEMENT] various witness mode related enhancement


Vicente-Cheng:
**Is your enhancement related to a problem? Please describe.**
We will introduce the witness mode support in v1.3 sprint release.
This issue will be used to record anything we need to improve. 

**Describe the solution you'd like**
Will be described individually on below

**Describe alternatives you've considered**
None

**Additional context**

Related to the original issue: https://github.com/harvester/harvester/issues/3266

**UI enhancement**
We saw the following figures. It needs to do some specific handling with that.

- [x]  node-2 is the `etcd` node, it should not show cordon here. 
<img width="1676" alt="Screenshot 2023-12-07 at 6 36 31 PM" src="https://github.com/harvester/harvester/assets/1615081/7851cafd-32ce-4df2-9553-cf75e37cda1a">

- [x] role should not be compute node
<img width="1195" alt="Screenshot 2023-12-07 at 6 39 18 PM" src="https://github.com/harvester/harvester/assets/1615081/1ea56c05-dbe8-4bd8-bb36-10de44416a8b">

**Backend related enhancement**
- [x] Should use exact match on the promote.sh for checking the etcd role
reference: https://github.com/harvester/harvester/pull/4781#discussion_r1418836758

- [x] Add the limitation for only one etcd node to be promoted (for promote-controller).
We need to make sure the kube-apiserver is good. Two etcd nodes currently would make some trouble for it.

- [x] Kick the witness node out of Longhorn nodes; it cannot be scheduled. (For longhorn UI issue: https://github.com/longhorn/longhorn/issues/7450)

- [x] Add unittest for promote-controller

**Document enhancement**
- [ ] Write the document for this feature and its limitation

bk201:
<img width="1880" alt="截圖 2023-12-22 晚上8 05 38" src="https://github.com/harvester/harvester/assets/1691518/3be081a4-e4ce-4a7d-b66f-41adfdb3cc70">
The Longhorn GUI also displays the etcd node down. We might need to create a ticket in Longhorn.

Vicente-Cheng:
> The Longhorn GUI also displays the etcd node down. We might need to create a ticket in Longhorn.

~Opened the longhorn issue: https://github.com/longhorn/longhorn/issues/7450~

Update another ToDo item for this case

torchiaf:
@Vicente-Cheng what would be the status and the role labels for witness nodes ?

Vicente-Cheng:
[...]

PR #4967, opened 2024-01-10
controller/promote: misc witness enhancement

Vicente-Cheng:

**Problem:**
Do the enhancement for the witness feature.
- Should only have one witness node in the cluster
- Longhorn UI shows NotReady for the witness node. (Actually we can not deploy any longhorn workload on witness node)

**Solution:**
- limit the number of witness node
- remove witness node from `nodes.longhorn.io`

**Related Issue:**
https://github.com/harvester/harvester/issues/4840

**Test plan:**
1. Create a cluster and choose witness mode on the second and third nodes
2. Check only the second node would be promoted as the witness node
3. Make sure the nodes of longhorn UI should not contain the witness node 

** Optional Item: **
- [ ] add unittest of promote controller
- [ ] promote role on demand

### DIFF
--- a/deploy/charts/harvester/templates/configmap.yaml
+++ b/deploy/charts/harvester/templates/configmap.yaml
@@ -20,9 +20,10 @@ metadata:
 data:
   promote.sh: |-
     {{`KUBECTL="/host/$(readlink /host/var/lib/rancher/rke2/bin)/kubectl"
+    YQ="/host/usr/bin/yq"
     ROLE_LABELS="rke.cattle.io/control-plane-role=true rke.cattle.io/etcd-role=true"
     ETCD_ONLY=false
-    if [[ -n "$1" ]]; then
+    if [[ -n "$1" && $1 == "rke.cattle.io/etcd-role=true" ]]; then
       ETCD_ONLY=true
       ROLE_LABELS=$1
     fi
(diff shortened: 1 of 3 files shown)


Vicente-Cheng:
Tested with basic operations and guest cluster.
Please check this PR with the installer part https://github.com/harvester/harvester-installer/pull/630

deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -93,21 +93,38 @@ data:
     - rke2-snapshot-validation-webhook
     EOF
 
+    # make sure we should not have any related label/taint on the node
```
ibrokethecloud, line 97: this fails since kubectl returns a non zero respone when attempting to remote a taint when one is not present and promotion fails 
```
error: taint "node-role.kubernetes.io/etcd:NoExecute" not found 
```
This can be fixed by checking if taint exists before attempting to remove it.

```suggestion
      # make sure we should not have any related label/taint on the node
    if [[ $ETCD_ONLY == false ]]; then
      found=$($KUBECTL get node $HOSTNAME -o yaml | $YQ '.spec.taints[] | select (.effect == "NoSchedule" and .key == "node-role.kubernetes.io/etcd=true") | .effect')
      if [[ -n $found ]]
      then
        $KUBECTL taint nodes $HOSTNAME node-role.kubernetes.io/etcd=true:NoExecute-
      fi
      $KUBECTL label --overwrite nodes $HOSTNAME node-role.harvesterhci.io/witness-
    fi
```

deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -22,7 +22,7 @@ data:
     {{`KUBECTL="/host/$(readlink /host/var/lib/rancher/rke2/bin)/kubectl"
     ROLE_LABELS="rke.cattle.io/control-plane-role=true rke.cattle.io/etcd-role=true"
```
ibrokethecloud, line 24: define YQ:
```suggestion
    YQ="/host/usr/bin/yq"
    ROLE_LABELS="rke.cattle.io/control-plane-role=true rke.cattle.io/etcd-role=true"
```

deploy/charts/harvester/templates/configmap.yaml
Vicente-Cheng: Nice catch! Updated.
Please help to recheck.

ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):
LGTM, thanks!

Issue #3758, opened 2023-04-10
[BUG] User can delete the volume which is exporting image


lanfon72:
### Describe the bug
when the volume is exporting (to image), we can still delete it. 

### To Reproduce
1. Install Harvester with any nodes
2. Create a image `opensuse.img`
3. Create a volume `v1` from image `opensuse.img`
4. Export the volume `v1` into image `v1image`
5. Delete volume `v1`
6. Image `v1image` stuck in `Exporting` state

### Expected behavior
Image should be exported successfully.

## Environment:
 - Harvester ISO version: **v1.1.2-rc6**
 - **ui-source** Option: **Auto**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **qemu/KVM 3 nodes**

### Additional context
This could be reproduced both in API and UI.
We don't have any similar test case, so it not `not-require/test-plan`.

WuJun2016:
@guangbochen   After preliminary investigation, the UI status appears to be normal. It is necessary for a colleague on the backend team to determine whether the behavior is as expected.

albinsun:
Provide test result on `v1.3-5e35d640-head`, delete an exporting volume is still valid and cause exporting image stuck.


## Environment
* Harvester
  * Version: `v1.3-5e35d640-head`
  * Profile: QEMU/KVM, 2 nodes (8C/20G/500G)
  * ui-source: `Auto`


## Steps 
1. Go to **Images** -> **Create** -> Create **opensuse-leap-15.5.x86_64-nocloud.qcow2**
   <details>
 
     ![image](https://github.com/harvester/harvester/assets/2773781/8464f6c2-37eb-4065-94d9-cbc7ebab5934)
   </details>
   
1. Go to **Volumes** -> **Create** -> Create volume **v1** from **opensuse-leap-15.5.x86_64-nocloud.qcow2**
   <details>
 
     ![image](https://github.com/harvester/harvester/assets/2773781/fd9600d7-a027-4973-a7eb-e8bc8486ba96)
   </details>

1. Go to **Volumes** -> **v1** -> "⋮" -> **Export Image** to **v1image**
   <details>
 
     ![image](https://github.com/harvester/harvester/assets/2773781/f492c37f-7247-4637-b33a-aaa4a2069a88)
     ![image](https://github.com/harvester/harvester/assets/2773781/55406ae4-a63a-45bb-a6d6-448aed91417c)
   </details>

   * **v1image** show `Exporting`

[...]

PR #4976, opened 2024-01-15
Prevent Exporting PVC from Deletion

WebberHuang1118:
**Problem:**
Image stuck in exporting if the source volume is deleted

**Solution:**
Prevent exporting volume from deletion

**Related Issue:**
#3758 

**Test plan:**
- Creating a volume (both image-backed or not are fine)
- Exporting the volume to an image
- Deleting volume **before** image exporting complete
  - Harvester should receive webhook error `can not delete volume [pvc name] which is exporting for image [image namespace]/[image name]`
- Deleting volume **after** image exporting complete
  - The related PVC should be deleted sucessfully

### DIFF
--- a/pkg/webhook/indexeres/indexer.go
+++ b/pkg/webhook/indexeres/indexer.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	longhorntypes "github.com/longhorn/longhorn-manager/types"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/indexeres"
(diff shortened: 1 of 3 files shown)


FrankYang0529 (approved):
LGTM. The webhook denies the request if PVC hasn't finished exporting VMImage.

Screenshot 2024-01-17 at 1 08 54 PM

bk201 (approved):

pkg/webhook/indexeres/indexer.go
```diff
@@ -73,3 +78,12 @@ func VolumeByReplicaCount(obj *lhv1beta2.Volume) ([]string, error) {
 	replicaCount := strconv.Itoa(obj.Spec.NumberOfReplicas)
 	return []string{replicaCount}, nil
 }
+
+func imageByExportSourcePVC(obj *harvesterv1.VirtualMachineImage) ([]string, error) {
+	if obj.Spec.SourceType != longhorntypes.LonghornLabelExportFromVolume ||
+		obj.Spec.PVCNamespace == "" || obj.Spec.PVCName == "" {
+		return []string{}, nil
```
connorkuehl:
```suggestion
		return nil, nil
```

I usually prefer a nil slice here since I can never remember if the compiler ends up heap-allocating anything for `[]string{}`  and with `nil` I don't have to finally learn to remember :laughing:

connorkuehl (approved):

pkg/webhook/indexeres/indexer.go
WebberHuang1118: Updated per suggestion, thanks.

connorkuehl (approved):

WebberHuang1118:
@mergify backport v1.2

Issue #4784, opened 2023-11-28
[BUG] Harvester v1.2.1 stuck on boot with BOSS-S1 hardware RAID


bk201:
**Describe the bug**

Although we fixed the issue in v1.1.x, the customer still has an issue with 1.2.1.

**To Reproduce**
- Install Harvester v1.2.1.
- Reboot
- The OS hangs on "Welcome to GRUB!" for 30+ minutes.

**Expected behavior**

The OS should boot smoothly.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.2.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

bk201:
@megabreit Can you check if the issue description is still correct? And do you see the same issue in 1.2.0? Thanks.

megabreit:
@bk201 We don't see this issue in 1.2.0.
The description is basically correct. It's still only happening with machines which contain the BOSS-S1 controller. There are some small differences.
- This time we see this error in the middle of the 30 mins:
```
error: ../../grub-core/fs/fshelp.c:258:file '/grub2/i386-pc/zfs.mod not found.
```
It's probably not related to the actual error, but maybe it helps to better track the problem.
- This time the boot counter is missing, so the boot does not automatically continue. The user needs to press Enter on the console.

The workaround `grub2-editenv /oem/grubcustom create` is still working, but maybe the actual cause is different this time.

Vicente-Cheng:
> The workaround `grub2-editenv /oem/grubcustom create` is still working, but maybe the actual cause is different this time.

yap, that is what we are curious about. We are trying to reproduce it in our lab. And for sure I will backport this to v1.2.2.

megabreit:
One, maybe important, detail: Our Harvester nodes are booting in BIOS mode. Not because of a technical decision, but just because it's default.

TachunLin:
Verified **fixed** on `master-9e663418-head` (24/01/12).

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ On BOSS-S1 bare machine: Can install and reboot Harvester on BOSS-S1 RAID machine without stuck on boot $~~$</summary>

  1. Can install Harvester on BOSS-S1 card machine and without stuck on boot in GRUB
  
  https://github.com/harvester/harvester/assets/29251855/aff58090-097c-49d7-9887-eecc929a035d
  
  
[...]

PR #4980, opened 2024-01-15
upgrade: add `/oem/grubcustom` to prevent boot stuck

Vicente-Cheng:

**Problem:**
The boot stuck on the grub might be because of the search for a non-existent file.
(With specific Hardware, like BOSS-S1 raid card)
This is for the upgrade path, to prevent this situation when reboot.

**Solution:**
Manually create these files with the upgrade path

**Related Issue:**
https://github.com/harvester/harvester/issues/4784

**Test plan:**
Make sure the boot is not stuck on the bare metal with the BOSS-S1 raid card after upgrade


bk201 (approved):

starbops (approved):
LGTM, thanks!

Issue #3019, opened 2022-10-24
[Enhancement] Upgrade: enhance live-migration detection in the pre-drain phase


bk201:
**Is your feature request related to a problem? Please describe.**

We have trivial rules to detect non-migration VMs in the `upgrade_node.sh` script.
We need to adjust the rule after the v1.1.0 release.

**Describe the solution you'd like**

Some exceptions that we can still live-migrate VMs: 
- We add a mutator to schedule VMs on nodes with a network plane (PR: https://github.com/harvester/harvester/pull/3000).
- The user specifies an affinity that could contain multiple nodes.

We also need to shutdown VMs when passing through PCI devices to VMs.

**Describe alternatives you've considered**

**Additional context**

bk201:
Related: https://github.com/harvester/harvester/issues/2317

bk201:
@starbops Please help backport this to v1.2, it will help reveal  https://github.com/harvester/harvester/issues/5076 in v1.2.1 -> v1.2-head path. Thanks!

starbops:
As there is a problem migrating VMs during the upgrade from v1.2.1 to master (tracked in #5076), I suggest we postpone the testing til the issue is solved.

TachunLin:
Thanks for the checking, according to previous comment, we would move the issue to ready for testing and test til the rke2 have been bumped for the upcoming release.

noahgildersleeve:
Since this is an upgrade test and upgrading two versions of k8s isn't supported should we be running the upgrade from v1.2-head to v1.3.0-rcx for testing? CC @starbops

starbops:
Hi @noahgildersleeve thanks for bringing this up. Yes, we should validate this one by upgrading a cluster from v1.2-head to v1.3.0-rcX. And for the test against the webhook, please try to upgrade from v1.3.0-rcX to v1.3.0-rcX (same version upgrade). Thank you!

noahgildersleeve:
# Test Plan
Full Test plan outlined [here](https://github.com/harvester/harvester/pull/4984) and quoted below
**EDIT: THESE ARE OBSOLETE SCENARIOS AND ARE EXPANDED IN MORE RECENT COMMENT**
- [ ] VMs with node affinities
- [ ] VMs with PCIe devices
- [ ] VMs with node selector

> **Test plan:**
> 
> _VMs with node affinities_
> 
> 1. Prepare a 2-node Harvester cluster in v1.2.1
[...]

PR #4984, opened 2024-01-16
Enhance non-live migratable VMs detection in both webhook and upgrade script

starbops:
**Problem:**

The part handling non-live-migratable VMs in the upgrade script is a bit hard to read and maintain.

**Solution:**

This PR contains three part:

1. A command line interface called `vm-live-migrate-detector`, which is able to show the "non-live migratable VMs" given a node name as its positional argument. Plus a `--shutdown` flag that could help users shut off those non-migratable VMs directly with the tool at hand. The binary is targeted to replace the respective part in the upgrade shell script.
2. A webhook that does the validation work against Upgrade objects creation. When there are "non-live migratable VMs" running on the cluster, it will reject the request and return the VM name list.
3. A shared library code supporting the above two

**Related Issue:**

#3019 

**Test plan:**

*VMs with node affinities*

1. Prepare a 2-node Harvester cluster in v1.2.1
5. Create two VMs: one is live-migratable, the other is non-live-migratable. To achieve this, we could tweak the node scheduling option for the VMs.
SCR-20240116-okkg
SCR-20240116-okdu
6. Upgrade using the master ISO image with any custom-built image which includes this PR
7. Monitor the progress of the upgrade, the live-migratable VM should move between the two nodes and the non-live-migratable VM originally running on the control plane node should be shutoff when the control plane node is pre-draining.
8. The upgrade should end in success

*VMs with PCIe devices*

1. Prepare a 2-node Harvester cluster in v1.2.1
2. Enable `pcidevices-controller` addon
3. Enable the device you want for passthrough
4. Create a live-migratable VM with regular configs
5. Create another non-live-migratable VM with the PCI device attached
9. Start the upgrade
10. The non-live-migratable VM should be shutoff when the node it runs on is being pre-drained
11. The upgrade should end in success and the live-migratable VM is still running

*VMs with node selector*

1. Prepare a 2-node Harvester cluster in v1.2.1
2. Create a live-migratable VM with regular configs
3. Create another non-live-migratable VM with the node scheduling option set to a specific node
SCR-20240116-phzu
4. Start the upgrade
5. The VM should be shutoff when the node it runs on is being pre-drained
6. The upgrade should end in success and the live-migratable VM is still running

---

For the tests against the webhook, the above three test cases could be reused. The only difference is the cluster should be installed with the master build (or custom build containing this PR), not v1.2.1. The upgrade in all three cases should be rejected by the webhook with the names of the VM which are not live-migratable.

image


starbops:
The code conflicts have been resolved. PTAL, thanks!

starbops:
~~Hold on. I think I messed up the commits while solving the conflicts. I'm turning it into a draft until the commits are sorted out.~~

It's fixed now.

cmd/vm-live-migrate-detector/README.md
```diff
@@ -0,0 +1,54 @@
+# vm-live-migrate-detector
```
bk201: While the command is useful for detecting non-migratable VMs for now, can we rename the utility to something like upgrade-helper and put vm-live-migrate-detector as a sub-command for future extendibility? I'm thinking eventually we want to move all the upgrade_node.go logic into the command to save our time.

starbops: I separated the `vm-live-migrate-detector` from the root command so that other subcommands could be added to the `upgrade-helper` more easily in the future. Thanks for the suggestion!

FrankYang0529 (approved):
LGTM. Thanks.

We may have an edge case: a group of VM with anti affinity and the `topologyKey` is `kubernetes.io/hostname`.

However, I think it's okay for this version because we have `wait_vms_out_or_shutdown` in upgrade script.

cmd/upgradehelper/vmlivemigratedetector.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+		if err := run(ctx, options); err != nil {
+			fmt.Fprintf(os.Stderr, "%s\n", err.Error())
+			os.Exit(1)
+		}
+	},
+}
+
+func init() {
+	vmLiveMigrateDetectorCmd.Flags().BoolVar(&shutdown, "shutdown", false, "Do not shutdown non-migratable VMs")
```
bk201:
```suggestion
	vmLiveMigrateDetectorCmd.Flags().BoolVar(&shutdown, "shutdown", false, "Shut down non-migratable VMs")
```

make more sense in the help messsage:
```
Usage:
  upgrade-helper vm-live-migrate-detector NODENAME [flags]

Flags:
  -h, --help       help for vm-live-migrate-detector
      --shutdown   Do not shutdown non-migratable VMs
```

cmd/upgradehelper/vmlivemigratedetector.go
```diff
@@ -0,0 +1,53 @@
...
+
+	rootCmd.AddCommand(vmLiveMigrateDetectorCmd)
+}
+
+func run(ctx context.Context, options vmlivemigratedetector.DetectorOptions) error {
+	logrus.Info("Starting VM Live Migrate Detector")
+	detector := vmlivemigratedetector.NewVMLiveMigrateDetector(options)
+	detector.Init()
```
bk201: Check the return code in case init fails.

pkg/util/virtualmachineinstance/virtualmachineinstance.go
```diff
@@ -0,0 +1,78 @@
...
+	var nonLiveMigratableVMINames []string
+
+	// Skip for single-node clusters
+	if len(nodes) == 1 {
+		return nonLiveMigratableVMINames, nil
+	}
+
+	for _, vmi := range vmis {
+		// Node selectors
+		if vmi.Spec.NodeSelector != nil {
```
bk201, line 24: Can we log the reason why a VMI can't be migrated for these 3 cases? And optionally we can construct `vmi.Namespace+"/"+vmi.Name` to a variable first and reuse it.

starbops:
> We may have an edge case: a group of VM with anti affinity and the topologyKey is kubernetes.io/hostname.

As @FrankYang0529 pointed out, this PR could not detect VMs with such cases as non-live migratable. For example, if a user creates multiple VMs with the following VM scheduling configuration:

SCR-20240201-jllq

Which results in:

```
$ kubectl get vmis test-vm-01 -o yaml
...
podAntiAffinity:
    - labelSelector:
        matchExpressions:
          - key: group
            operator: In
            values:
              - ha
      topologyKey: kubernetes.io/hostname
...
```

Not just nodeAffinity, pod(Anti)Affinity should also be considered.

Update: tracked by #5077

pkg/util/virtualmachineinstance/virtualmachineinstance.go
starbops: Updated. PTAL, thank you!

bk201 (approved):
LGTM, thank you!

starbops:
I removed the force-shutdown part of the script at the node-upgrading phase, as we tend to let users decide whether to shut off such VMs.

bk201 (approved):

w13915984028:
@torchiaf A corner-case question about below warning message, it includes the vm name list, and this string may be very long, will it be shown in a scroll box, or with a `...` ? could it expand enough to hide the `close` and `upgrade` button? thanks.

image

torchiaf:
The modal dialog will show all the content without text ellipsis nor scroll bar:

image

I think the warning message will be reasonably shorted that this above, so the UI should be pretty much safe.

Issue #4786, opened 2023-11-28
[ENHANCEMENT] Manually configure Harvester roles for hosts


bk201:
**Is your enhancement related to a problem? Please describe.**

Harvester automatically promotes random nodes into the controller node, there are situations we don't want a node to be promoted.

**Describe the solution you'd like**

- Specify node's role while installing the node. We can have "management", "witness", "compute", etcd.

**Describe alternatives you've considered**

**Additional context**

Vicente-Cheng:
For this enhancement, we would like to do the following two things.
1. promote the specific node to mgmt node
2. prevent promote the specific node to mgmt node

We will define the following three role label
- management
- witness
- worker

The promote rule is as follows:
- without role definition: promote the ready worker node randomly
- with role definition:
  - promote the witness node to etcd node. (maximum: 1)
  - promote the mgmt node to mgmt node.
  - do not promote the worker node.

WDYT? cc @bk201, @ibrokethecloud

bk201:
That sounds sane to me.

albinsun:
## Test Plan

### Objective
1. Test the promotion behavior of each role when
   * Join a cluster which quorum is 1, 2, 3
   * Existed, but cluster quorum drops from 3 to 2 (one management down)
1. Worker will never be promoted
   * Setup initial condition as 1 D/M + 1 C
   * The C node-1 will never be promote.
1. Witness only promote when reach quorum.


### ExpeTest Matrix
We have 4 TCs each role, 16 total.

| **State When**     | **Join q=1** | **Join q=2** | **Join q=3** | **q=3->2 ** |
|:------------------:|:------------:|:------------:|:------------:|:-----------:|
| **management (M)** |  C (1M1C)    |   M (3M)     |   C (3M1C)   |  M (3M)     |
|   **default (D)**  |  C (1M1C)    |   M (3M)     |   C (3M1C)   |  M (3M)     |
|   **witness (W)**  |  W (1M1W)    |   M (3M)     |   W (3M1W)   |  M (3M)     |
|   **worker (C)**   |  C (1M1C)    |   C (2M1C)   |   C (3M1C)   |  C (2M1C)   |

> [!NOTE]
> Q: Quorum (3 here)
> q: Current number of management (candidates), namely `#M + #D + #W`

albinsun:
Provide test result for `management` suite.

## Environment
* Harvester
  * Version: `v1.3.0-rc2`
[...]

PR #4990, opened 2024-01-17
Add role definition for promote

Vicente-Cheng:

**Problem:**
We should give a way to make users promote the node that they prefer.
Also, give a way to make users prevent do not promote some nodes.

**Solution:**

We will define the following three role label
- management
- witness
- worker

The promote rule is as follows:
- without role definition: promote the ready worker node randomly
- with role definition:
  - promote the witness node to etcd node. (maximum: 1)
  - promote the mgmt node to mgmt node.
  - do not promote the worker node.

**Related Issue:**
https://github.com/harvester/harvester/issues/4786

**Test plan:**
Please refer to the test plan on https://github.com/harvester/harvester-installer/pull/634


ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):
LGTM, thanks!

Issue #4725, opened 2023-11-14
[BUG] Upgrade 1.2.0 -> 1.2.1 is stuck in “Waiting for VM live-migration or shutdown...(1 left)” even though there is NO VM running


Martin-Weiss:
**Describe the bug**
During the upgrade of Harvester 1.2.0 to 1.2.1 the upgrade gets stuck with the message "`Waiting for VM live-migration or shutdown...(1 left)`"

But no VM is running at all:
![grafik](https://github.com/harvester/harvester/assets/13362573/b8850acf-0ac4-4ffa-a1db-2646c8c2229f)

Status of the VM:
```
status:
  conditions:
    - lastProbeTime: '2023-09-15T09:54:53Z'
      lastTransitionTime: '2023-09-15T09:54:53Z'
      message: virt-launcher pod is terminating
      reason: PodTerminating
      status: 'False'
      type: Ready
    - lastProbeTime: null
      lastTransitionTime: null
      status: 'True'
      type: LiveMigratable
  created: true
  printableStatus: Stopped
  volumeSnapshotStatuses:
    - enabled: false
      name: cdrom-disk
      reason: 2 matching VolumeSnapshotClasses for longhorn-image-dzr67
    - enabled: false
      name: rootdisk
      reason: 2 matching VolumeSnapshotClasses for longhorn
    - enabled: false
      name: virtio-container-disk
      reason: >-
        Snapshot is not supported for this volumeSource type
        [virtio-container-disk]
    - enabled: false
      name: cloudinitdisk
      reason: Snapshot is not supported for this volumeSource type [cloudinitdisk]
```

**To Reproduce**
Upgrade a 3 node harvester cluster from 1.2.0 to 1.2.1 with all VMs stopped while one VM has a similar status as above.

**Expected behavior**
Upgrade does not get stuck based on "not running VMs"

**Environment**
 - Harvester ISO version: 1.2.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Baremetal

Martin-Weiss:
[1.20 - 1.12 hvst-upgrade-xdt69-upgradelog-archive-2023-11-13T13-10-19Z.zip](https://github.com/harvester/harvester/files/13346383/1.20.-.1.12.hvst-upgrade-xdt69-upgradelog-archive-2023-11-13T13-10-19Z.zip)

w13915984028:
Per discussion and investigation, below differences are related:

[...]

PR #4999, opened 2024-01-22
fix(upgrade): skip residual vmis during pre-draining

starbops:

**Problem:**

Residual VMIs, i.e., VMIs with phase equals "Succeeded" and their corresponding launcher Pods in a "Completed" state, will cause the waiting of live migrating VMs much longer.

**Solution:**

Skip the VMIs with the phase in "Succeeded" when counting the number of running VMs.

**Related Issue:**

#4725

**Test plan:**

1. Prepare a Harvester cluster in version 1.2.1
2. Create a VM and shut it down from within the OS (do not shut down the VM via dashboard)
SCR-20240122-ithh
3. Upgrade the cluster with the master build or custom-built ISO image containing the fix
4. The stopped VM should not not stall the upgrade

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -132,7 +132,7 @@ get_running_vm_count()
 {
   local count
 
-  count=$(kubectl get vmi -A -l kubevirt.io/nodeName=$HARVESTER_UPGRADE_NODE_NAME -ojson | jq '.items | length' || true)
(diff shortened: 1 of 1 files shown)


starbops:
Note: It seems `jq` could not return an empty array `[]` if the filtering result is empty, so I changed it to `yq`.

```
$ kubectl get vmi -A -l kubevirt.io/nodeName=harvester-xg8wq -ojson | jq '.items[] | [select(.status.phase!="Succeeded")] | length
# <-- no output instead of "zero"
```

bk201 (approved):
lgtm, thanks!

FrankYang0529 (approved):
LGTM. Thanks.

bk201:
@mergifyio backport v1.2

Issue #4970, opened 2024-01-11
[TASK] Bump Longhorn to v1.6.0 in v1.3.0


bk201:
**Is your enhancement related to a problem? Please describe.**

To provide better support, we need to ship Longhorn v1.6.x with Harvester v1.3.0.
Currently, Harvester v1.2.x has Longhorn v1.4.x, which will violate Longhorn's one-minor-version upgrade policy if we bump Longhorn. Longhorn team is working on the upgrade path validation and will support it: https://github.com/longhorn/longhorn/issues/7539

**Describe the solution you'd like**

Bump Longhorn to v1.6.x.

**Describe alternatives you've considered**

**Additional context**

innobead:
ref: https://github.com/longhorn/longhorn/issues/7539

Let's start the integration after RC2 is out.

lanfon72:
Verified Longhorn `v1.6.0` applied to Harvester  `v1.3.0-rc2`
Longhorn related (volumes, images, backups) automation tests all _**PASS**_ excepts known issues.

Waiting the case 2 described in https://github.com/harvester/harvester/pull/5087#issue-2113924381 that `v1.2.1` still using Longhorn `v1.4.3`, same as the latest _v1.2-head_: `v1.2-c5270e4d-head`, and the upgrade is blocked.

lanfon72:
Verified Harvester `v1.2-head` upgraded to `v1.3.0-rc2` successfully, and Longhorn related automation tests all _PASS_ excepts known issues.

PR #5002, opened 2024-01-22
feat: bump longhorn to v1.6.0-rc2

FrankYang0529:
**Solution:**
Bump LH to v1.6.0-rc2.

**Related Issue:**
https://github.com/harvester/harvester/issues/4970

**Test plan:**

### Case 1: New cluster

Make sure users can operate creating/migrating VM, creating and restoring VMBackup/VMSnapshot.

### Case 2: Upgrade cluster

1. Create a harvester v1.2.1 cluster.
2. Create a guest cluster.
3. Make sure the harvester can bump to this branch.


FrankYang0529:
Both test cases can pass in my local environment.

Screenshot 2024-01-23 at 3 55 43 PM

WebberHuang1118:
Do we need to align [Harvester's LH related packages ](https://github.com/FrankYang0529/harvester/blob/a265efd4fd832f7125ca330108e465e00bb53db7/go.mod#L69-L70) with [LH v.1.6.0-rc2](https://github.com/longhorn/longhorn-manager/blob/0c1a4942cd99432622d2ed375331be526eb74723/go.mod#L47)? Or we can wait until LH v1.6.0 officially release? Thanks.

FrankYang0529:
> Do we need to align [Harvester's LH related packages ](https://github.com/FrankYang0529/harvester/blob/a265efd4fd832f7125ca330108e465e00bb53db7/go.mod#L69-L70) with [LH v.1.6.0-rc2](https://github.com/longhorn/longhorn-manager/blob/0c1a4942cd99432622d2ed375331be526eb74723/go.mod#L47)? Or we can wait until LH v1.6.0 officially release? Thanks.

Yes, good catch. Updated it. Thank you.

tserong (approved):
LGTM, I just have three questions:

1. How is `deploy/charts/harvester/charts/longhorn-1.6.0-rc2.tgz` (and other chart tarballs) created? Are those files generated automatically, or is this something we have to do manually when updating chart versions or adding new charts? (I'm just curious)
2. The included longhorn 1.6.0-rc2 chart adds `.Values.namespaceOverride`, but this is not present in the upstream 1.6.0-rc2 longhorn chart (https://github.com/longhorn/longhorn/tree/v1.6.0-rc2/chart). It _is_ in longhorn 1.5.3, so my question is: is this a bug or omission in longhorn 1.6.0-rc2 that needs to be fixed there, or is it intentionally removed, and something we have to carry only in harvester?
3. I assume using RC2 now is an intermediate measure so we can get some testing done before the official LH 1.6.0 release, and then we will bump again after the LH release?

FrankYang0529:
Hi @tserong, thanks for reviewing!
 
> 1. How is `deploy/charts/harvester/charts/longhorn-1.6.0-rc2.tgz` (and other chart tarballs) created? Are those files generated automatically, or is this something we have to do manually when updating chart versions or adding new charts? (I'm just curious)

Yes, it's generated automatically. You can try to run `helm dependency update` to generate it.

> 2. The included longhorn 1.6.0-rc2 chart adds `.Values.namespaceOverride`, but this is not present in the upstream 1.6.0-rc2 longhorn chart (https://github.com/longhorn/longhorn/tree/v1.6.0-rc2/chart). It _is_ in longhorn 1.5.3, so my question is: is this a bug or omission in longhorn 1.6.0-rc2 that needs to be fixed there, or is it intentionally removed, and something we have to carry only in harvester?

Yes, in this version, I manually add it back. This field will be backup in upstream when v1.6.0-rc3 is released. Reference: https://github.com/longhorn/longhorn/issues/6527#issuecomment-1903514911.

> 3. I assume using RC2 now is an intermediate measure so we can get some testing done before the official LH 1.6.0 release, and then we will bump again after the LH release?

Yes, you're right. We will use official release when v1.6.0 is out.

FrankYang0529:
> > Do we need to align [Harvester's LH related packages ](https://github.com/FrankYang0529/harvester/blob/a265efd4fd832f7125ca330108e465e00bb53db7/go.mod#L69-L70) with [LH v.1.6.0-rc2](https://github.com/longhorn/longhorn-manager/blob/0c1a4942cd99432622d2ed375331be526eb74723/go.mod#L47)? Or we can wait until LH v1.6.0 officially release? Thanks.
> 
> Yes, good catch. Updated it. Thank you.

Hi @WebberHuang1118, I will only bump backupstore, but not longhorn-manager. The longhorn-manager v1.6.0-rc2 relies on `wait.PollUntilContextCancel` which was introduced in k8s.io/apimachinery after v0.27.0. We will use kubernetes v1.26 in Harv v1.3.0, so I prefer to keep k8s related packages as v0.26.x.

Screenshot 2024-01-24 at 2 16 35 PM

WebberHuang1118:
VM restore controller [couldn't update LH volume](https://github.com/FrankYang0529/harvester/blob/ed1b28f1e086f405113ac043327cfdbd74bf8550/pkg/controller/master/backup/restore.go#L294-L298), because it tries to update v1.6.0 LH volume with[ v1.5.3 LH volume data structure](https://github.com/FrankYang0529/harvester/blob/ed1b28f1e086f405113ac043327cfdbd74bf8550/go.mod#L70), however, there is a [newly added field image](https://github.com/longhorn/longhorn-manager/blob/0c1a4942cd99432622d2ed375331be526eb74723/k8s/pkg/apis/longhorn/v1beta2/volume.go#L244), and [LH webhook prevent update this field with empty value](https://github.com/longhorn/longhorn-manager/blob/0c1a4942cd99432622d2ed375331be526eb74723/webhook/resources/volume/validator.go#L209-L213).

As a result, VMRestore has no resotre progress, since it couldn't link restoring pvc with volume, thanks.

bk201 (approved):

WebberHuang1118 (approved):
LGTM, thanks.

Verify with fundamental operations, just found a front-end issue for VM snapshot restore https://github.com/harvester/harvester/issues/5029

Checked with:
- During Source VM running
   - Backup -> Restore New -> Check Data Integrity
- Stop Source VM
   - Backup -> Restore Existing -> Check Data Integrity
   - Backup -> Restore New -> Check Data Integrity
   - Backup -> Restore to another namesapce -> Check Data Integrity
   - Snapshot -> Restore New -> Check Data Integrity

Vicente-Cheng (approved):
Tested upgrade from v1.2.1
```
longhorn-manager  ●  longhornio/longhorn-manager:v1.6.0-rc2  true  Running
```

snapshot -> restore to new/current
backup -> restore to new/current
works well, LGTM!

PR #5021, opened 2024-01-24
controller/promote: make witness role with higher priority

Vicente-Cheng:

**Problem:**
We should have priority with witness role and default.

**Solution:**
Add a new preferred slice for selection

**Related Issue:**
https://github.com/harvester/harvester/issues/4786

**Test plan:**
Please refer to https://github.com/harvester/harvester-installer/pull/634


bk201 (approved):
LGTM!

ibrokethecloud (approved):
lgtm. thanks.

PR #5027, opened 2024-01-25
chore: fix node-manager pullPolicy and add webhook image in chart values

bk201:
- Fix pullPolicy: The field should be `pullPolicy` according to https://github.com/harvester/charts/blob/master/charts/harvester-node-manager/values.yaml
- Add webhook image to values.

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -488,9 +488,14 @@ whereabouts:
 
 harvester-node-manager:
   image:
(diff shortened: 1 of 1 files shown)


WebberHuang1118 (approved):
LGTM, thanks

FrankYang0529 (approved):
LGTM. Thanks.

Vicente-Cheng (approved):
thanks for fixing that!

PR #5035, opened 2024-01-26
CI: fix chart length detection and publish artifacts on release branches

bk201:
A "forward port" from the [`release-1.3`](https://github.com/harvester/harvester/commits/release-1.3/) branch:
- Fix the length comparison on the chart name.
- Allow publishing images and ISO on `release**` branches:
  - Currently, we are doing a release on `release-<version>` branch. I extend the syntax to support `release/<version>` in the future too.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -58,6 +58,7 @@ steps:
       ref:
         include:
           - "refs/heads/master"
+          - "refs/heads/release**"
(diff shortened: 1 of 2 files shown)


Vicente-Cheng (approved):
LGTM, thanks!

starbops (approved):
LGTM, thanks.

Issue #5044, opened 2024-01-29
[BUG] Improve Image Download Handling


WebberHuang1118:
**Describe the bug**
Downloaded images can't be uploaded again

**To Reproduce**
1. Creating a virtual machine image with upload from local
2. Downloading the image just uploaded
3. Creating another virtual image by uploading the newly downloaded image
4. Virtual image creation failed with `Upload failed: the uploaded file size 665368738 should be a multiple of 512 bytes since Longhorn uses directIO by default`

**Expected behavior**
Successfully uploading an image which is originally downloaded from Harvester 

**Environment**
 - Harvester ISO version: Harvester master-branch (commit `0861f5be4533418bbe148efc5f0bc32f07fe4220`)
 
 **Additional context**
After investigation, we found out LH changes the behavior to [compress the downloaded image](https://github.com/longhorn/backing-image-manager/pull/153), in the current Harvester implementation, the downloaded image file will be named with `.img` suffix by default, however, this file is compressed by LH, if users want to upload the downloaded image, it should be decompressed first.

Harvester should change the [attachment in the request header as LH](https://github.com/longhorn/backing-image-manager/blob/28b787c0ddfc95fb6af5573294e6366cdd0f87f2/pkg/sync/service.go#L246-L251) to name the downloaded image with `.gz` suffix, which could avoid confusing, and a document updating is also required.

**Workaround**
Rename the downloaded image as `.gz` suffix, and decompress it with `$ gzip -d <image name>`, then the image file should be eligible to upload.

irishgordo:
Hi, @WebberHuang1118 I'm noticing some issues with this still, but not sure if due to underlying configuration.

**Validation Failed:**

I'm still noticing the issue is present on:
- Harvester ISO Version: v1.3.0-rc1
- Rancher Dashboard Version: `commit c3ae9bc14ad1a306025c0882eefde7baef3d046d` ( containing `commit 65366810efe83db05512194589771b8c3a0a0914` ) (has fix for volumes)
- Underlying Infrastructure: qemu/kvm

![Screenshot from 2024-02-01 16-27-53](https://github.com/harvester/harvester/assets/5370752/8747744f-dfd3-4c34-aea6-64803690686c)

WebberHuang1118:
Hi @irishgordo,
[...]

PR #5053, opened 2024-01-30
Renaming the download image file with appropriate suffix

WebberHuang1118:
**Problem:**
Described in https://github.com/harvester/harvester/issues/5044#issue-2105104769

**Solution:**
Renaming the download image file with the appropriate suffix

**Related Issue:**
#5044 

**Test plan:**
- Creating a virtual machine image that is uploaded from local
- Downloading the image just uploaded
  -  The downloaded image should be named with `.gz` suffix by default
- Decompress the downloaded image with `$ gzip -d <image name>`
- Creating another virtual image by uploading the newly downloaded image
- The image should be uploaded successfully

### DIFF
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -106,7 +106,7 @@ func (h Handler) downloadImage(rw http.ResponseWriter, req *http.Request) error 
 		return fmt.Errorf("failed to get backing image name for VMImage %s/%s, error: %w", namespace, name, err)
 	}
 
-	targetFileName := vmImage.Spec.DisplayName
+	targetFileName := fmt.Sprintf("%s.gz", vmImage.Spec.DisplayName)
 	downloadURL := fmt.Sprintf("%s/backingimages/%s/download", util.LonghornDefaultManagerURL, biName)
 	downloadReq, err := http.NewRequestWithContext(req.Context(), http.MethodGet, downloadURL, nil)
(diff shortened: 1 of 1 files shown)


bk201:
~~do we need to change our header too?~~

https://github.com/longhorn/backing-image-manager/blob/28b787c0ddfc95fb6af5573294e6366cdd0f87f2/pkg/sync/service.go#L246-L251

NVM, it's updated later in the code.

pkg/api/image/formatter.go
```diff
@@ -106,7 +106,7 @@ func (h Handler) downloadImage(rw http.ResponseWriter, req *http.Request) error
 		return fmt.Errorf("failed to get backing image name for VMImage %s/%s, error: %w", namespace, name, err)
 	}
 
-	targetFileName := vmImage.Spec.DisplayName
+	targetFileName := fmt.Sprintf("%v.gz", vmImage.Spec.DisplayName)
```
bk201: any reason to use `%v` verb here?

WebberHuang1118: updated as `%s`, thanks.

bk201 (approved):
lgtm! thanks!

FrankYang0529 (approved):
LGTM. Thanks.

WebberHuang1118:
@Mergifyio backport v1.2

Issue #5070, opened 2024-01-31
[BUG] Second witness node's role is Management but NOT behave like management node


albinsun:
According to https://github.com/harvester/harvester/issues/4840#issuecomment-1883190132
> Even if we try to set up 2 witness nodes, we still have one witness node.

Suppose the 2nd witness node should become and behave like managemant node.
Please help to check it's expected or not.

## Environment
* Harvester
  * Version: `v1.3.0-rc1`
  * Profile: QEMU/KVM, 2 -> 3 nodes
    * Default (D) node: 8C/16G/500G
    * Witness (W) node: 4C/8G/500G
  * ui-source: `Auto`
  * Longhorn: `v1.6.0-rc2`

## Steps 
1. Setup 2 nodes cluster (Default + Witness)
   * <details>
     <summary> Cluster is Active :heavy_check_mark: 
     </summary>
   
     ![image](https://github.com/harvester/harvester/assets/2773781/45ccb0a3-cc07-47b0-ae29-cdfe93b06116)
     </details>

   * <details>
     <summary> <b>node-0</b> is Management and <b>node-1</b> is Witness :heavy_check_mark: 
     </summary>
     
     node-0
     ![image](https://github.com/harvester/harvester/assets/2773781/327fe6f6-5683-4002-aeca-64cc76c0cbcd)
     node-1
     ![image](https://github.com/harvester/harvester/assets/2773781/2c8f5600-2f19-4fbe-b87a-27ef5dce242d)
     </details>

   * <details>
     <summary> Longhorn should has only 1 node :heavy_check_mark: 
     </summary>
   
     ![image](https://github.com/harvester/harvester/assets/2773781/44d840f9-a2a7-4156-a175-0fb99f1b0592)
     </details>

2. Create a VM with some data
   * <details>
     <summary> VM is Running :heavy_check_mark: 
     </summary>
   
     ![image](https://github.com/harvester/harvester/assets/2773781/f1b09099-d1a7-4c39-bdea-c5ca95b7a20c)
     </details>

   * <details>
     <summary> Volume is Ready, replica should be 1 (LH has 1 node) :heavy_check_mark: 
     </summary>
   
     ![image](https://github.com/harvester/harvester/assets/2773781/8875d4b7-0406-4345-8746-08ac74a5985a)
     </details>

   
3. Adding Witness node-2
   * <details>
     <summary> Cluster is Active :heavy_check_mark: 
     </summary>
   
[...]

PR #5082, opened 2024-02-01
controller/promote: refine the promote mechanism with witness nodes

Vicente-Cheng:

**Problem:**
We only support one witness node in the cluster.
If we have more than one witness node, other witness nodes will not be calculated into the management node number.

**Solution:**
Update the promotion mechanism.

**Related Issue:**
https://github.com/harvester/harvester/issues/5070

**Test plan:**
1. Create 3 nodes cluster
2. 1st node as default, 2nd and 3rd node select the witness role.
3. Should not promote 2nd/3rd nodes, which means we only have one management node
The should should be 1 management node (1st node) and 2 worker node (2nd and 3rd nodes)

### DIFF
--- a/enhancements/20231113-witness-node-support.md
+++ b/enhancements/20231113-witness-node-support.md
@@ -58,7 +58,7 @@ We need to document the following limitations and requirements for the witness n
 
 - The replica number of the default StorageClass is `3`, but in the two master nodes + one witness node scenario, the replica number should be `2`. Users could manually create a new StorageClass and set it as default or use the installation configuration to change the replica number of the default StorageClass [link](https://docs.harvesterhci.io/v1.2/install/harvester-configuration#installharvesterstorage_classreplica_count).
 
-- Only allow one witness node in the cluster.
+- Only allow one witness node in the cluster. If we have more than one witness node, promote controller will not count the other witness nodes into the management node number.
 
 #### Requirements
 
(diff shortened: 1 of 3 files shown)


ibrokethecloud (approved):
lgtm. thanks.

enhancements/20231113-witness-node-support.md
albinsun, line 1: Does this mean we still have 2 Witness in this case?

Vicente-Cheng: Nope, we do not allow it.
Add this description for more specific behavior when the user defines two witness (or more) nodes.

pkg/controller/master/node/promote_controller.go
```diff
@@ -288,6 +290,8 @@ func selectPromoteNode(nodeList []*corev1.Node) *corev1.Node {
 			managementNumber++
 		}
 
+		witnessPromoted = witnessPromoted || isWitnessNode(node, isManagement)
```
Yu-Jack, line 293: if `witnessPromoted` is true, then it's always true, is that expected?

Vicente-Cheng: yes, this means this round, the selection will know we already have one witness node promoted.

Yu-Jack (approved):
lgtm

PR #5092, opened 2024-02-02
Remove disable-all in golangci-lint

FrankYang0529:

**Problem:**

**Solution:**

**Related Issue:**

**Test plan:**


FrankYang0529:
Close the issue since some fix is introduced in https://github.com/harvester/harvester/pull/5377. I will create another PR after it's merged.

PR #5103, opened 2024-02-04
patch local cluster during upgrade path to support v2.8.1

ibrokethecloud:

**Problem:**

Post upgrade to v1.2-head and v1.3-head, rancher still complains about `fleetWorkspaceName` not being setup.

For new installs this is already being setup by rancherd during the initialisation phase.

Currently the upgrade still completes successfully as the old fleet agent completes the updates to the bundles.

The fleet agent itself does not get updated as a result, and this is likely to cause issues in future upgrades.

This is visible as seen here: https://github.com/harvester/harvester-installer/pull/646#issuecomment-1925792176

**Solution:**

An extra step is available in the upgrade_manifests step which will perform the same annotation and status subresource updates as rancherd.

**Related Issue:**

**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -1241,6 +1241,14 @@ sync_containerd_registry_to_rancher() {
   kubectl annotate --overwrite=true setting.harvesterhci.io containerd-registry "harvesterhci.io/upgrade-patched=$REPO_HARVESTER_VERSION"
 }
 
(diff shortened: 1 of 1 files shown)


FrankYang0529 (approved):
LGTM. Thank you!

Vicente-Cheng (approved):
lgtm, get more info from https://github.com/rancher/rancherd/pull/35

ibrokethecloud:
@mergify backport v1.2

Vicente-Cheng:
@mergifyio backport v1.2

PR #5057, opened 2024-01-30
Bump longhorn to v1.5.3

FrankYang0529:
**Solution:**
Bump longhorn to v1.5.3

**Related Issue:**
https://github.com/harvester/harvester/issues/5056

**Test plan:**
### Case 1: New cluster

Make sure users can operate creating/migrating VM, creating and restoring VMBackup/VMSnapshot.

### Case 2: Upgrade cluster

1. Create a harvester v1.2.1 cluster.
2. Create a guest cluster.
3. Make sure the harvester can bump to this branch.
4. After the first bump is successful, bump the cluster again to master-head.

When bumping to master-head, the `rancher/harvester-cluster-repo` may not be reloaded, because the tag is same. This issue will not happened in production, because the tag will be different. For testing, users can load it manually after each node is in image-preloaded status. After doing it on each node, users can delete harvester-cluster-repo pod to restart it.

```bash
# the ip is cluster-ip of upgrade-repo-hvst-upgrade-xxx service.
curl -sfL http://<ip>/harvester-iso/bundle//harvester/images/harvester-repo-images-master.tar.zst | zstdcat | ctr -n k8s.io images import -
```

Issue #4879, opened 2023-12-18
[ENHANCEMENT] upgrade: adapt Longhorn im-e/im-r pod unification


bk201:
**Is your enhancement related to a problem? Please describe.**

In Longhorn 1.5.x, im-r and im-e pods are consolidated into one pod. We should adapt the change and fix the waiting logic here: https://github.com/harvester/harvester/blob/a46456a7cf1f4bafcc22a7b6960d9b7bb6732122/package/upgrade/upgrade_manifests.sh#L161

**Describe the solution you'd like**

Fix the wait logic.

**Describe alternatives you've considered**

**Additional context**

bk201:
Move back to implement. for upgrade from LH v1.5 -> v1.6, there will be 2 aio instance-manager pods if a node has any attached volumes.

lanfon72:
Verified the enhancement has been applied.

Test Information
----
* Environment: **baremetal DL360G9 5 nodes**
* Harvester Version: **v1.2-d542563e-head** -> **v1.3.0-rc2**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Create an image for VM creation
3. Create VMs with additional volumes
4. perform upgrade
5. Harvester should upgraded successfully
6. VM should keep running and operation normally

PR #5087, opened 2024-02-02
feat: bump longhorn to v1.6.0

FrankYang0529:
**Solution:**
Bump LH to v1.6.0.

**Related Issue:**
https://github.com/harvester/harvester/issues/4970

**Test plan:**

### Case 1: New cluster

Make sure users can operate creating/migrating VM, creating and restoring VMBackup/VMSnapshot.

### Case 2: Upgrade cluster

1. Create a harvester v1.2.1 cluster.
2. Create a guest cluster.
3. Make sure the harvester can bump to this branch.

Issue #5076, opened 2024-02-01
[BUG] VM can not migrate during upgrade


Vicente-Cheng:
**Describe the bug**
VM can not migrate during upgrade.
for example, I run the VM on node1. when node1 is draining, it should migrate to other nodes but fail.
That will impact the service, which should not be down.

**To Reproduce**
1. create 3 nodes cluster (version v1.2.1)
2. create guest cluster with rke2 cloud provider
3. upgrade to v1.3.0-rc1
4. Check the VM should running after upgrade.

**Expected behavior**
VM should migrate successfully

**Support bundle**
None

**Environment**
 - Harvester ISO version: v1.2.1 -> v1.3.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): kvm, 3 nodes cluster

**Additional context**
Some migrate failed logs from virt-handler.
```
2024-01-31T13:23:48.555760707Z stderr F {"component":"virt-handler","level":"info","msg":"re-enqueuing VirtualMachineInstance default/rke2-guest-01-pool1-b4649f31-9g2nx","pos":"vm.go:1588","reason":"syncing migration target failed: unknown error encountered sending command SyncMigrationTarget: rpc error: code = Unavailable desc = error reading from server: EOF","timestamp":"2024-01-31T13:23:48.555655Z"}
...
2024-01-31T13:25:33.067561695Z stderr F {"component":"virt-handler","level":"info","msg":"re-enqueuing VirtualMachineInstance default/rke2-guest-01-pool1-b4649f31-9g2nx","pos":"vm.go:1588","reason":"syncing migration target failed: unknown error encountered sending command SyncMigrationTarget: rpc error: code = Unavailable desc = error reading from server: EOF","timestamp":"2024-01-31T13:25:33.067501Z"}
...
2024-01-31T13:34:22.398539648Z stderr F {"component":"virt-handler","level":"info
","msg":"re-enqueuing VirtualMachineInstance default/rke2-guest-01-pool1-b4649f31-9g2nx","pos":"vm.go:1588","reason":"syncing migration target failed: unknown error encountered sending command SyncMigrationTarget: rpc error: code = Unavailable desc = error reading from server: EOF","timestamp":"2024-01-31T13:34:22.398483Z"}
...
[...]

PR #5105, opened 2024-02-05
fix(upgrade): replace instance manager count check with state check

FrankYang0529:

**Problem:**
Currently, the engine process can't be live migrated to a new longhorn instance manager, so there may be two IM at the same time. We have to improve our upgrade check for instance manager.

Ref: [Troubleshooting: Some old instance manager pods are still running after upgrade](https://longhorn.io/kb/troubleshooting-some-old-instance-manager-pods-are-still-running-after-upgrade/)

**Solution:**
Check whether new instance manager is running.

**Related Issue:**
https://github.com/harvester/harvester/issues/4879

**Test plan:**
1. Create a v1.2.1 Harvester cluster.
2. Create a VM.
3. Upgrade to v1.2-head with LH v1.5.3 (https://github.com/harvester/harvester/pull/5057).
4. Upgrade to this branch again. Instance manager check should pass without any error.

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -158,36 +158,43 @@ wait_longhorn_manager() {
   done
 }
 
-wait_longhorn_instance_manager_r() {
+wait_longhorn_instance_manager_aio() {
+  node_count=$(kubectl get nodes --selector=harvesterhci.io/managed=true -o json | jq -r '.items | length')
+  if [ $node_count -le 2 ]; then
+    echo "Skip waiting instance-manager (aio), node count: $node_count"
+    return
+  fi
+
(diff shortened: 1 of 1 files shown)


markhillgit (approved):
lgtm thanks

ibrokethecloud (approved):
lgtm. thanks.

Vicente-Cheng (approved):
LGTM, tested with https://github.com/harvester/harvester/pull/5087
upgrade path from v1.2.1 -> v1.2-head (with Longhorn v1.5.3) -> master-head (with this PR)

verified the VM can be migrated (w/o vlan due to https://github.com/harvester/harvester/issues/5076)
And all basic operations work well.

Also tested the snapshot/backup/restore work well.

FrankYang0529:
@mergify backport v1.2

Issue #4920, opened 2023-12-27
[BUG] `snapshot-validation-webhook` error in backup page


ne0YT:
**Describe the bug**
I try to restore Backups from a NFS Target on a new single node Cluster

**To Reproduce**
1. setup the backup target where the Backups are saved on.
2. see in Backup that it says:
```
Internal error occurred: failed calling webhook "snapshot-validation-webhook.csi.kubernetes.io": failed to call webhook: Post "https://harvester-snapshot-validation-webhook.kube-system.svc:443/volumesnapshot?timeout=2s": x509: certificate signed by unknown authority (possibly because of "x509: invalid signature: parent certificate cannot sign this kind of certificate" while trying to verify candidate authority certificate "harvester-snapshot-validation-webhook.kube-system.svc")
```

**Expected behavior**
I should see the backups normally and be able to restore them

**Support bundle**

**Environment**
 - Harvester ISO version: 1.2.1

ne0YT:
[supportbundle_9794a7ef-1936-46c1-b739-572848ed4c2b_2023-12-27T22-42-36Z.zip](https://github.com/harvester/harvester/files/13781941/supportbundle_9794a7ef-1936-46c1-b739-572848ed4c2b_2023-12-27T22-42-36Z.zip)

bk201:
@ne0YT I shorten the issue title.
@FrankYang0529 Can you help check the issue?

FrankYang0529:
Hi @ne0YT, may I know which Harvester version you used for the target before you set it to v1.2.1 cluster? Also, any setting did you modify for v1.2.1 cluster? 

I tried following test steps, but I can't reproduce the issue:
* Create the first cluster with harvester version v1.2.1 and setup backup-target.
* Create VMImage, VM, and VMBackup in the first cluster.
* Create the second cluster with harvester version v1.2.1  and setup a same backup-target.
* Check the VMBackup can be restored to the second cluster.

ne0YT:
Both were v1.2.1
The only thing I changed was configuring the storage to be single node Instead of three nodes (on both clusters).

I took the backups to change the Network-config by setting up the cluster from scratch.
On this site I only have one host available.

ne0YT:
@FrankYang0529 
Any Idea what to try?

FrankYang0529:
From the `kube-system/harvester-snapshot-validation` deployment, I can find log like:

```
I1227 22:37:17.283761       1 certwatcher.go:129] Updated current TLS certificate
[...]

PR #5107, opened 2024-02-05
feat(chart): bump snapshot-controller and snapshot-validation-webhook to v6.3.3

FrankYang0529:
## Test plan:

2. Check snapshot-controller and webhook is v6.3.3.


FrankYang0529:
Hi @ibrokethecloud, could you also help me check this PR? I can use this branch to create a cluster. At first, I also got error like following in the managedchart.

```
failed calling webhook \"snapshot-validation-webhook.csi.kubernetes.io\": failed to call webhook: Post \"https://harvester-snapshot-validation-webhook.kube-system.svc:443/volumesnapshot?timeout=2s\": dial tcp 10.96.21.231:443: connect: connection refused"
```

The reason is that `snapshot-validation-webhook` gets error like https://github.com/harvester/harvester/issues/4920#issuecomment-1871786968.

The error can be fixed when I removed all `snapshot-validation-webhook` pods at the same time (not deployment). I'm not sure why the first deployment can't work.

I follow the example like https://github.com/piraeusdatastore/helm-charts/tree/snapshot-validation-webhook-1.9.0/charts/snapshot-validation-webhook to change container port or add `--port=8443` to the argument, but it can't fix the error.

Could you help to take a look? Thank you!

ibrokethecloud:
there are 2 changes in the last commit:

* drop `volumesnapshotclasses` from list of resources being watched for by the snapshot webhook. This currently breaks the installation of the helm chart during CI as the harvester chart also creates a `volumesnapshotclass`
* renamed the deployment template file of snapshot webhook to `webhook-deployment` to overcome what seems to be the race condition, where the pods error out due `tls` error, likely because of order in which resources are created. deletion of snapshot webhook pods fixes the issue, likely because now a valid certificate exists. fleet creates resources based on name of file, and not like helm. https://github.com/rancher/fleet/issues/411. renaming the file works around the `tls` startup error as the cert is always created before the deployment.

w13915984028 (approved):
LGTM, thanks.

connorkuehl (approved):

FrankYang0529:
Test with last commit. The snapshot-controller and webhook is v6.3.3. Also, backup and restore can work. LGTM. Thanks.

deploy/charts/harvester/dependency_charts/snapshot-validation-webhook/templates/deployment.yaml
```diff
@@ -27,7 +27,8 @@ spec:
             {{- end }}
           ports:
             - name: https
-              containerPort: {{ .Values.service.port | default 443 }}
+              containerPort: {{ .Values.args.port | default 8443 }}
+              protocol: TCP
```
tserong: You mentioned:

> I follow the example like https://github.com/piraeusdatastore/helm-charts/tree/snapshot-validation-webhook-1.9.0/charts/snapshot-validation-webhook to change container port or add --port=8443 to the argument, but it can't fix the error.

Given Gaurav's subsequent fix, is this port change necessary?  If not, I'd suggest to revert it (unless there's some other reason I'm not yet aware of to change the port)

deploy/charts/harvester/dependency_charts/snapshot-validation-webhook/values.yaml
```diff
@@ -16,14 +16,15 @@ rbac:
...
-  tag: v5.0.1
+  tag: v6.3.3
   pullPolicy: IfNotPresent
 
 podAnnotations: {}
 
 args:
   tlsPrivateKeyFile: /etc/snapshot-validation-webhook/certs/tls.key
   tlsCertFile: /etc/snapshot-validation-webhook/certs/tls.crt
+  port: 8443
```
tserong: Same as my previous comment.

tserong (approved):
LGTM, I just have one comment/suggestion.

deploy/charts/harvester/dependency_charts/snapshot-validation-webhook/templates/deployment.yaml
FrankYang0529: Yeah, I think this update is not helpful. Reverted it. Thanks.

deploy/charts/harvester/dependency_charts/snapshot-validation-webhook/values.yaml
FrankYang0529: Reverted it. Thanks.

FrankYang0529:
@mergify backport v1.2

Issue #5121, opened 2024-02-06
[BUG] When you add a vGPU to a VM it won't boot


noahgildersleeve:
**Describe the bug**

When you add a vGPU to a VM it won't boot

**To Reproduce**

1. Enable both pcidevices-controller and nvidia-driver-toolkit addon on a system with a GPU installed
2. Navigate to SR-IOV GPU Devices in the left sidebar
3. Enable one of the GPUs and select a vGPU profile
![Greenshot 2024-02-06 12 47 15](https://github.com/harvester/harvester/assets/83787952/6cb633c9-6a3f-4cbf-8d1a-86581f9750fc)
![Greenshot 2024-02-06 14 25 41](https://github.com/harvester/harvester/assets/83787952/1983135d-aeff-4db3-b633-5a9dab7200ef)
5. Add the vGPU profile to a VM (I used a ubuntu focal VM)
![Greenshot 2024-02-06 14 25 57](https://github.com/harvester/harvester/assets/83787952/fe124e2a-b76f-4ae7-afbe-3fdaea4db3d8)
6. Save and restart the VM

**Expected behavior**

The vGPU should attach and the VM should boot

**Support bundle**

[supportbundle_b580a14b-9c0c-465c-ad49-dcdde8cfbb84_2024-02-06T23-44-11Z.zip](https://github.com/harvester/harvester/files/14187284/supportbundle_b580a14b-9c0c-465c-ad49-dcdde8cfbb84_2024-02-06T23-44-11Z.zip)

**Environment**
 - Harvester ISO version: v1.3.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 2 node bare metal DL360

**Additional context**
![Greenshot 2024-02-06 14 26 41](https://github.com/harvester/harvester/assets/83787952/0c011f97-89a9-478e-ab5a-19b27418a6e9)

noahgildersleeve:
This might be fixed with [this](https://github.com/harvester/harvester/pull/4987).

ibrokethecloud:
the mutating webhook fixes are now available in rc2

noahgildersleeve:
Tested in v1.3.0-rc2 in 2 node bare metal setup. Following the same steps as above the VM wouldn't start. It's worth noting that for this VM I'm using a custom storage class with 2 replicas on NVME with it being selected by disk tag.

[...]

PR #5131, opened 2024-02-09
update resource overcommit during GPU updates

ibrokethecloud:

**Problem:**

for host devices and vGPU to work with kubevirt v1.1.1 the resource overcommit needs to be disabled.

There is already a change in v1.3.0-rc2 which performs this action when a VM is created with host device / GPU devices defined. 

However the check for VM updates when host devices or GPU's are added to existing VM's does not take into account that a resource overcommit change is needed. 

As a result of this VM's where GPU is added after a VM has been created can not boot.

**Solution:**

The fix adds an extra check in the VM resource mutation to check if a memory patch is needed if host devices or GPUs are added.

**Related Issue:**
https://github.com/harvester/harvester/issues/5121
**Test plan:**


tserong (approved):

w13915984028 (approved):
LGTM, thanks.

Issue #5163, opened 2024-02-19
[Bug] Explicit behavior on multiple witness node


albinsun:
According to witness role design, there is a statement that 
> ...
> Requirement
> 
> Only one witness node is allowed in the cluster. [^1]
> ...

Per discussed[^2], if user deploy more than one witness node, currently we will just let it behave by it's design. If so we can have a cluster w/ 2 etcd nodes by speicific steps.
![image](https://github.com/harvester/harvester/assets/2773781/8d9ed5e7-78e0-40d5-a7f2-14a2a6b8339e)

If this is an expected behavior, shall we update the design document?
If it's NOT, and we should convert this ticket into bug.

## Environment
* Harvester
  * Version: `v1.3.0-rc2`
  * Profile: QEMU/KVM, 5 nodes (8C/10G/500G)

## Steps 
1. <details>
   <summary> Setup MCWWD cluster
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/e52a2b6f-a122-4875-a863-bf14d842d2ee)
   ![image](https://github.com/harvester/harvester/assets/2773781/33fcfa67-a7b2-4f59-b705-ab9a6158aedc)
   </details>
   
2. Delete node-4 (M) (q=3->2)
   * <details>
     <summary> node-0 (D) is <code>Management</code>
     </summary>
     
     ![image](https://github.com/harvester/harvester/assets/2773781/afaa25f9-a761-4669-8ff6-1426216a2797)
     </details>

   * <details>
     <summary> node-1 (C) is <code>Compute</code>
     </summary>
     
     ![image](https://github.com/harvester/harvester/assets/2773781/03d280c9-edc8-468e-bc23-6fc0582c2f79)
     </details>

   * <details>
     <summary> node-2 (W) is <code>Witness (etcd)</code> 
     </summary>
     
     ![image](https://github.com/harvester/harvester/assets/2773781/d4d1e1ef-b7b6-4c11-bf58-c04b2ef58df6)
     </details>

   * <details>
     <summary> node-3 (W) is <code>Witness (etcd)</code>
     </summary>
     
     ![image](https://github.com/harvester/harvester/assets/2773781/6f4e9605-bbf5-48cf-af80-d68183cbc7a7)
     </details>

   * <details>
     <summary> node-4 (M) is <code>n/a</code>
     </summary>
     
     </details>

   * <details open>
     <summary> Longhorn nodes is 2 (DC)
     </summary>
     
[...]

PR #5168, opened 2024-02-19
controller/promote: fix multiple witness nodes promotion

Vicente-Cheng:

**Problem:**
When we delete the management node, sometimes the witness node will be promoted again even if we already have one in the cluster.

**Solution:**
Ensure the promote controller do not promote another witness node if we already have one.

**Related Issue:**
https://github.com/harvester/harvester/issues/5163

**Test plan:**
See the steps from https://github.com/harvester/harvester/issues/5163


ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):
lgtm thanks!

Issue #5192, opened 2024-02-22
[BUG] Witness node timeout in UpgradingNode phase during upgrade


albinsun:
We found witness node will timeout fail when preloading image.
Please help to check, thanks.

## Environment
* Harvester
  * Version: `v1.3.0-rc2`
  * Profile: QEMU/KVM, 3 nodes (8C/16G/500G)
  * ui-source: `Auto`

## Steps 
1. Setup DDW cluster using `v1.3.0-rc2` :heavy_check_mark: 
   * Environment: QEMU/KVM, 3 nodes (8C/16G/500G)
   * <details>

     ![image](https://github.com/harvester/harvester/assets/2773781/54e6be4a-fddb-4603-947e-6223f84af584)
     </details>
3. Extending job time out to `7200`
   Ref. https://github.com/harvester/harvester/issues/2894#issuecomment-1274069690

2. Perform a airgapped upgrade to `v1.3.0-rc3` :x: ${\color{red} \text{Witness node preloading image timed out}}$
   * Reproducibility: 3/3
   * Upgrade log: [hvst-upgrade-jnhpk-upgradelog-archive-2024-02-22T00-51-53Z.zip](https://github.com/harvester/harvester/files/14367087/hvst-upgrade-jnhpk-upgradelog-archive-2024-02-22T00-51-53Z.zip)

   * Support bundle: [supportbundle_002bb4fc-a3f3-4a75-bae8-b045cbeaec97_2024-02-22T00-53-54Z.zip](https://github.com/harvester/harvester/files/14367088/supportbundle_002bb4fc-a3f3-4a75-bae8-b045cbeaec97_2024-02-22T00-53-54Z.zip)

   * <details open>
  
     ![image](https://github.com/harvester/harvester/assets/2773781/05c97aac-ef69-4218-a7de-9f6a65e1cbab)
     </details>

# Reference
1. #3266

albinsun:
Provide test result with `v1.3-7cb8315c-head` (same version upgrade), close as fixed.



## Environment
* Harvester
  * Version: `v1.3-7cb8315c-head`
  * Profile: QEMU/KVM, 3 nodes
    * 2 default nodes: 8C/16G/500G
    * 1 witness node: 4C/8G/500G 
  * ui-source: `Auto`


## Test Case
### 1. Fresh Cluster :ok: 
1. Setup Harvester cluster with witness node
2. <details>
   <summary> Create a 2-replica storage class and set as default :heavy_check_mark:
[...]

PR #5202, opened 2024-02-22
upgrade: misc enhancement for witness node

Vicente-Cheng:

**Problem:**
Upgrade stuck with witness node scenario

**Solution:**
improve the upgrade related jobs

**Related Issue:**
https://github.com/harvester/harvester/issues/5192

**Test plan:**
1. Create a cluster with one witness node
2. Check the upgrade result

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -140,7 +140,7 @@ wait_longhorn_manager() {
   lm_repo=$(kubectl get apps.catalog.cattle.io/harvester -n harvester-system -o json | jq -r .spec.chart.values.longhorn.image.longhorn.manager.repository)
   lm_tag=$(kubectl get apps.catalog.cattle.io/harvester -n harvester-system -o json | jq -r .spec.chart.values.longhorn.image.longhorn.manager.tag)
   lm_image="${lm_repo}:${lm_tag}"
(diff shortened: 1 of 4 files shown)


bk201:
I can see preparing job on witness role work now, but seems at the later stage, the `skip-restart-rancher-system-agent` can't finish: https://github.com/harvester/harvester/blob/db569a30aff50493ea7a178448f93adc09b38179/package/upgrade/upgrade_manifests.sh#L1131
Can you also help fix the plan? Thanks!

bk201:
The issue only happens in air-gap env, because registry.suse.com/bci/bci-base:15.4 doesn't exist anymore: https://github.com/harvester/harvester/blob/db569a30aff50493ea7a178448f93adc09b38179/package/upgrade/upgrade_manifests.sh#L1155
We can bump to 15.5 together

Vicente-Cheng:
> The issue only happens in air-gap env, because registry.suse.com/bci/bci-base:15.4 doesn't exist anymore:
> 
> https://github.com/harvester/harvester/blob/db569a30aff50493ea7a178448f93adc09b38179/package/upgrade/upgrade_manifests.sh#L1155
> 
> 
> We can bump to 15.5 together

Nice catch!
Updated to bci-base:15.5

bk201 (approved):
lgtm!

starbops (approved):
LGTM, thank you!

Issue #5175, opened 2024-02-19
[BUG] Scale down 3 node with-etcd cluster to 1 causes cluster out of service


albinsun:
We try to simulate an edge case that 2 nodes broken in a 3 nodes cluster via **Delete Node**
System looks fine when remaining 2 nodes but out of service when only 1 node left.

## Environment
* Harvester
  * Version: `v1.3.0-rc2`
  * Profile: QEMU/KVM, 3 nodes (8C/16G/500G)

## Steps 
1. Create MWM cluster
   * <details> 
     <summary> Cluster is Active :heavy_check_mark: 
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/ec67a63c-7e23-46de-8eb2-e80da45ee1be)
   </details>

   * node-0 is management :heavy_check_mark: 
   * node-1 is Witness (etcd) :heavy_check_mark: 
   * node-2 is management :heavy_check_mark: 
   
   
1. Delete node-2
   * <details> 
     <summary> Cluster is Active :heavy_check_mark: 
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/3737e59e-d6d1-4843-be32-79aa2e9038ae)
     </details>
   * node-0 is management :heavy_check_mark: 
   * node-1 is Witness (etcd) :heavy_check_mark: 

1. Delete node-1
   * <details open> 
       <summary> Cluster is Active :x: ${\color{red} \text{Error from server: etcdserver: request timed out}}$
       </summary>

       Delete
       ![image](https://github.com/harvester/harvester/assets/2773781/a03b7719-9a8f-409e-bb2c-93edd58ad1c9)
       Cluster our of service
       ![image](https://github.com/harvester/harvester/assets/2773781/0bcdce7e-9c69-4455-9d58-638e996387a6)
       ```
       harvester-node-0:~ # kubectl get nodes
       Error from server: etcdserver: request timed out
       ```
     </details>

## Supportbundle
1. Can not get support bundle from UI.
2. Provide /var/log tgz: [log.zip](https://github.com/harvester/harvester/files/14334386/log.zip)
1. `# tail -f /var/log/console.log`
   ```
   harvester-node-0:/var/log  
   time="2024-02-19T15:39:14Z" level=error msg="exit status 1Error from server: etcdserver: request timed out\n"
   time="2024-02-19T15:39:44Z" level=error msg="exit status 1Error from server: etcdserver: request timed out\n"
[...]

Issue #5158, opened 2024-02-17
[BUG] node-3 can NOT be promoted and stuck in Cordoned when shutdown and delete node-0


albinsun:
## Description
In a fresh installed 4 nodes cluster, when shutdown and delete node-0 (1st node), node-3 can NOT be promoted and stuck in Cordoned.

> [!NOTE]
> Test can pass if the target node is node-1.

## Environment
* Harvester
  * Version: `v1.3.0-rc2`
  * Profile: QEMU/KVM, 4 nodes (8C/10G/500G)

## Steps 
1. Setup 4 nodes Harvester cluster
   * All using default role (Do not specify role in yaml)
   * <details>
     <summary> Cluster is Active and with 3 management and 1 compute node. :heavy_check_mark: 
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/14fc55fa-f1b1-4103-a648-1780021993d2)
     </details>

   * <details>
     <summary> Longhorn show 4 Scheduable nodes too :heavy_check_mark: 
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/390c38ff-e9a0-411d-bae5-905ce9718730)
     </details>
   
1. Shutdown node-0 via shell CMD `shutdown -h now`
   * <details>
     <summary> node-0 becomes <code>Unavailable</code>, cluster still works (still 2 managements there)
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/86cf99de-4e1b-4d0e-8a39-d05862830727)
     </details>

   * <details>
     <summary> Longhorn show 4: 3 <code>Schedulable</code> and 1 <code>Down</code>
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/69073012-0005-4805-9dfd-8e83b8007411)
     </details>

1. Delete node-0 from UI
   * <details>
     <summary> node-0 gone, cluster still works (still 2 managements there)
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/17efa386-f86d-4325-8741-499487fa4be1)
     </details>

   * <details>
     <summary> node-3 should be promote to management :x: ${\color{red} \text{Stuck in Cordoned}}$
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/2657036f-3368-4b39-92ab-6b521b45917c)
     </details>

   * <details>
     <summary> Longhorn has 3 nodes: 2 <code>Schedulable</code> and 1 <code>Down</code>
[...]

PR #5204, opened 2024-02-22
feat: add cluster.x-k8s.io/control-plane label

FrankYang0529:

**Problem:**
A new node can't be promoted to control plane if the first node is removed.

**Solution:**
1. Add `cluster.x-k8s.io/control-plane` label to all control plane nodes, so CAPI doesn't fail in [reconcileDelete](https://github.com/kubernetes-sigs/cluster-api/blob/00dbf7b9f6322d7ebd06ae2efa703b23354dd37d/internal/controllers/machine/machine_controller.go#L306-L453) function.
2. Remove machine when users remove node, so Rancher doesn't use the first node as joining URL.

* https://github.com/rancher/rancher/blob/e8447834f3f5840c0a4dc0c7b9bd85787a082790/pkg/capr/planner/config.go#L120
* https://github.com/rancher/rancher/blob/e8447834f3f5840c0a4dc0c7b9bd85787a082790/pkg/capr/planner/planner.go#L899

**Related Issue:**
https://github.com/harvester/harvester/issues/5158

**Test plan:**

Case 1: Make sure all control plane nodes has `cluster.x-k8s.io/control-plane: "true"` label
2. Run following command and it should output 3 nodes.
```
kubectl get machine -n fleet-local -l cluster.x-k8s.io/control-plane=true
```

Case 2: Remove the first node and a new node can be promoted to control plane
1. Create a 4-node harvester cluster.
2. After all nodes are ready, make the first node into maintenance mode.
3. After the first node is in maintenance mode, remove the first node.
4. Check a new node can be promoted to control plane.
5. Check the first node related machine is removed.

### DIFF
--- a/pkg/controller/master/node/node_remove_controller.go
+++ b/pkg/controller/master/node/node_remove_controller.go
@@ -1,0 +1,68 @@
+package node
+
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
+
+	"github.com/harvester/harvester/pkg/config"
+	ctlclusterv1 "github.com/harvester/harvester/pkg/generated/controllers/cluster.x-k8s.io/v1alpha4"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	nodeRemoveControllerName = "node-remove-controller"
+)
+
+// nodeRemoveHandler force remove machine
+type nodeRemoveHandler struct {
+	machineClient ctlclusterv1.MachineClient
+}
+
+// RemoveRegister registers a controller to remove machine when node is removed
+func RemoveRegister(ctx context.Context, management *config.Management, _ config.Options) error {
+	nodes := management.CoreFactory.Core().V1().Node()
+	machineClient := management.ClusterFactory.Cluster().V1alpha4().Machine()
+	handler := &nodeRemoveHandler{
+		machineClient: machineClient,
+	}
+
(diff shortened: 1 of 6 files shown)


Vicente-Cheng:
the correct delete node behavior could also fix https://github.com/harvester/harvester/issues/5175.

FrankYang0529:
We will also update document to give an official way to remove node.

1. Go to the Hosts page
2. On the node you want to modify, click ⋮ > Enable Maintenance Mode
3. After the node is in Maintenance mode, click ⋮ > Delete

https://docs.harvesterhci.io/v1.2/host/#deleting-a-node

pkg/controller/master/node/node_remove_controller.go
```diff
@@ -0,0 +1,68 @@
...
+		machineClient: machineClient,
+	}
+
+	nodes.OnRemove(ctx, nodeRemoveControllerName, handler.OnNodeRemoved)
+
+	return nil
+}
+
+func (h *nodeRemoveHandler) OnNodeRemoved(_ string, node *corev1.Node) (*corev1.Node, error) {
+	if node == nil || node.Annotations == nil || node.Annotations[clusterv1.MachineAnnotation] == "" {
```
ibrokethecloud: we should also check if `DeletionTimestamp == nil` then return immediately.

The logic should only be run when DeletionTimestamp is set.

ibrokethecloud:
I ran a custom build with the extra DeletionTimestamp filter.

I was successfully able to scale down a 3 node HA cluster to a 1 node.

pkg/controller/master/node/node_remove_controller.go
FrankYang0529: Sorry for late. I just add the filter. I will test in my cluster again. Thank you.

ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):

Vicente-Cheng (approved):
LGTM! thanks for the enhancement!

Vicente-Cheng:
@mergifyio backport v1.3

PR #5213, opened 2024-02-23
Fix openapi spec

### LINKED ISSUE #5211: [BUG] Provide valid swagger docs
**Describe the bug**
Thank you for your work on this project. I am trying to generate a swagger client but I cannot because the document contains errors. The issues stem from an improper use of regex in the document. Instead you should use [swagger patterns](https://swagger.io/docs/specification/data-models/data-types/#pattern) (v3). Note: the current is invalid in both v2 and v3

For example:

[Current regex usage](https://raw.githubusercontent.com/harvester/harvester/master/api/openapi-spec/swagger.json)

```yaml
  paths:
    /api/v1/namespaces/{namespace:[a-z0-9][a-z0-9\\-]*}/persistentvolumeclaims: 
```

Should be
```yaml
paths: 
  /api/v1/namespaces/{namespace}/persistentvolumeclaims:
...
parameters:
- name: namespace
  in: path
  description: "Object name and auth scope, such as for teams and projects"
  required: true
  schema:
    pattern: '[a-z0-9][a-z0-9-]*'
    type: string
```

 

**To Reproduce**
```shell
$ swagger validate https://raw.githubusercontent.com/harvester/harvester/master/api/openapi-spec/swagger.json
```

**Confirming issues are fixed**

```shell
$ swagger generate client -f ./swagger.json -A harvester-go-sdk
```

I'm working on a PR that will also upgrade to oas v3

drewmullen:
Closes: https://github.com/harvester/harvester/issues/5211

Code gen now works:

```shell
$ openapi-generator generate -i https://raw.githubusercontent.com/drewmullen/harvester/master/api/openapi-spec/harvester.yaml -g go -o .
```


drewmullen:
my coworker pointed out that this doc is [generated](https://github.com/harvester/harvester/blob/master/scripts/generate-openapi) so we're gonna try and send over a fix of the generator so we dont get clobbered in future releases

Issue #5023, opened 2024-01-25
[BUG] Upgrade was stuck indefinitely in the "upgrade system services" phase due to fleet-agent unsuccessful registration


starbops:
**Describe the bug**

When upgrading from v1.2.1 to the master build, chances are you'll encounter a hurdle upgrading harvester charts. This is majorly due to Fleet, the one inside Harvester. I suspect there's a race condition in between.

The following are some observations:

*Before upgrade*

```
# the original ClusterRegistration object
name: request-4wj85
clientID: 8qtzwh6rkwtqgdnvg68k2qnw88v6jts6xfvtlxqwvz4sqh2w67vstd
clientRandom: xh2d9cvqk8gdrplf8tc7vdxbg5mgc4sxhmfkn6jd4gm9vbs4ddcw7v
---
# the generated Secret name (concatenate clientID and clientRandom, then sha256sum)
c-aede10a154549ac332492cd8c380ae5bfde238098ab3b980520d9ef5343cf7b1
```

*During Harvester charts upgrade*

Logs from cattle-fleet-local-system/fleet-agent:

```
...
create clusterregistration with id 8qtzwh6rkwtqgdnvg68k2qnw88v6jts6xfvtlxqwvz4sqh2w67vstd
Waiting for secret 'cattle-fleet-clusters-system/c-0d366abda81eeb6e0e2e4639bd37d7a816630bd361315709cae728abdb7b7 for request 'fleet-local/request-xkrps'
...
```

Logs from fleet-controller:

```
...
time="2024-01-24T02:57:59Z" level=info msg="Deleted old agent for cluster (fleet-local/local) in namespace cattle-fleet-local-system"
time="2024-01-24T02:57:59Z" level=info msg="Cluster import for 'fleet-local/local'. Deployed new agent"
W0124 02:57:59.626726       1 warnings.go:70] unknown field "status.apiServerCAHash"
W0124 02:57:59.626742       1 warnings.go:70] unknown field "status.apiServerURL"
time="2024-01-24T02:58:00Z" level=info msg="Waiting for service account token key to be populated for secret cluster-fleet-local-local-1a3d67d0a899/request-xkrps-5dce76d5-d54a-4b94-a795-9e29a5fe93bd-token"
time="2024-01-24T02:58:01Z" level=info msg="Namespace assigned to cluster 'fleet-local/local' enqueues cluster registration 'fleet-local/request-xkrps'"
[...]

Issue #5149, opened 2024-02-15
[BUG] node-manager-webhook has 3 replicas of POD in single-node cluster


w13915984028:
**Describe the bug**

node-manager-webhook has 3 replicas of POD in single-node cluster.

It seems such annotation is not in the replicaset definition:
```
management.cattle.io/scale-available: "3"
```

**To Reproduce**
1. Install a single-node v1.3.0 master-head cluster
2. Check pods and replicaset

```
NAMESPACE                         NAME                                                     READY   STATUS      RESTARTS      AGE

harvester-system                  harvester-node-manager-snqtr                             1/1     Running     0             25m
harvester-system                  harvester-node-manager-webhook-585c84c966-42h9s          1/1     Running     0             25m
harvester-system                  harvester-node-manager-webhook-585c84c966-9nds2          1/1     Running     0             25m
harvester-system                  harvester-node-manager-webhook-585c84c966-q7q5v          1/1     Running     0             25m

# kk get ReplicaSet -n harvester-system harvester-node-manager-webhook-585c84c966
NAME                                        DESIRED   CURRENT   READY   AGE
harvester-node-manager-webhook-585c84c966   3         3         3       28m
```

**Expected behavior**

1. Like `Harvester`, extend the replica when nodes are increasing.

**Support bundle**

**Environment**
 - Harvester ISO version:  Harvester v1.3.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM VM based

**Additional context**
Observed when validating https://github.com/harvester/harvester-installer/pull/658

---

- [x] https://github.com/harvester/charts/pull/221
- [x] Backport to release branch https://github.com/harvester/charts/pull/222
- [x] Vendor to harvester/harvester

Vicente-Cheng:
Thanks @w13915984028 for finding that.

I thought @connorkuehl should be more familiar with node-manager-webhook, so I changed the owner to him.

Hi @connorkuehl,
Please help to confirm that. Maybe we need the following annotation as @w13915984028 mentioned
```
management.cattle.io/scale-available
```

connorkuehl:
Fix should be available in 1.3.0-rc3, moving to Ready for Testing.

albinsun:
Hi @connorkuehl,

We found that **node-manager-webhook** will not deploy on witness node.
Please help to check the toleration.

cc @Vicente-Cheng 

## Environment
* Harvester
  * Version: `v1.3.0-rc3`
[...]

PR #5215, opened 2024-02-26
patch fleet cluster object to force redeploy of agent

ibrokethecloud:

**Problem:**

Users have randomly reported running into an issue during upgrade where the fleet-agent re-registration has not be successful

The current work around seems to be to delete the fleet-controller pod which forces a re-registration of the agents.

**Solution:**

Fix attempts to force a redeploy of the agent after the rancher upgrade to ensure that we do not hit this issue during the upgrade. Since the issue is hard to reproduce we may just have to wait until it is reported or not reported again.

**Related Issue:**
https://github.com/harvester/harvester/issues/5023

**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -281,6 +281,9 @@ upgrade_rancher() {
   wait_rollout cattle-system deployment rancher-webhook
   echo "Wait for cluster settling down..."
   wait_capi_cluster fleet-local local $pre_generation
+  patch_fleet_cluster
+  wait_for_deployment cattle-fleet-local-system fleet-agent
+  wait_rollout cattle-fleet-local-system deployment fleet-agent
 }
 
 update_local_rke_state_secret() {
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_manifests.sh
```diff
@@ -1256,6 +1258,23 @@ patch_local_cluster_details() {
...
+  local new_generation=$((generation+1))
+  patch_manifest="$(mktemp --suffix=.json)"
+  cat > "$patch_manifest" <<EOF
+{
+  "spec": {
+    "redeployAgentGeneration": $new_generation
+  }
+}
+EOF
+  echo $patch_string |  kubectl patch -n fleet-local cluster.fleet local  --type=merge --patch-file $patch_manifest
```
starbops: Maybe we could drop the `echo $patch_string` part?

starbops (commented):
Thanks for the fix! I just have one general question: do we need to revert the field? This seems to be a manner to force the agent to redeploy. I'm unsure if it will interfere with the fleet's standard upgrade procedure in the future.

package/upgrade/upgrade_manifests.sh
manno: `kubectl patch clusters.fleet.cattle.io -n fleet-local local --type=json -p '[{"op": "add", "path": "/spec/redeployAgentGeneration", "value": '"$new_generation"'}]'` works without a file, but it reads a bit more cryptic.

ibrokethecloud: i have no strong opinions about it, other than to force a redeploy of the agent 😆

ibrokethecloud:
during my testing, with clusters having 1 or 2 nodes only, the harvester bundle is reported as modified, which is fixed by:  https://github.com/harvester/charts/pull/226

For now to ensure the upgrade progresses the fix involves updating the harvester managed chart, using the instructions here: https://github.com/harvester/harvester/issues/5149#issuecomment-1960852535

package/upgrade/upgrade_manifests.sh
```diff
@@ -281,6 +281,9 @@ upgrade_rancher() {
   wait_rollout cattle-system deployment rancher-webhook
   echo "Wait for cluster settling down..."
   wait_capi_cluster fleet-local local $pre_generation
+  patch_fleet_cluster
+  sleep 15 #wait a few seconds to ensure that a new deployment is created. Without this wait_rollout fails as deployment is never found
```
starbops: Could we have something more robust to detect the existence of `fleet-agent` Deployment? I'm afraid that under heavy load, the rancher manager might have a chance of not responding in 15 seconds. Thanks!

bk201 (approved):
lgtm

starbops (approved):
LGTM, thanks!

FrankYang0529 (approved):
LGTM. Upgrade a cluster from v1.2.1 to this branch without error.

bk201:
@mergifyio backport v1.3

bk201:
@mergifyio backport v1.2

Issue #5190, opened 2024-02-21
[BUG] containerd-registry setting is no longer backwards compatible since 1.3.0-rc2


connorkuehl:
**Describe the bug**

Harvester's github.com/rancher/wharfie dependency was recently upgraded and it turns out that module introduced a [breaking change for the wharfie.Registry struct](https://github.com/harvester/harvester/commit/6687a6be3#diff-c9ae2ac8ca36c77313e61fb5a1bffcc789d55ef58fbb52bac2220aff81f8a3bc), which is what the Harvester settings controller uses to unmarshal the JSON payload from the Harvester dashboard where the containerd-registry settings are entered into.

As a result, now a containerd-registry setting from a version of Harvester prior to 1.3.0-rc2 is not forwards-compatible.

**To Reproduce**

1. Add a non-default containerd-registry setting value

```console
harv1:~ # kubectl get setting server-version
NAME             VALUE
server-version   v1.2.1
harv1:~ # cat /etc/rancher/rke2/registries.yaml 
{"configs":{},"mirrors":{"cr.example.com":{"endpoint":["https://cr.example.com:5000"]}}}
```

2. Upgrade to Harvester >= 1.3.0-rc2.

3. Touch the containerd-registry setting (either delete the hash annotation or just add a space somewhere in the containerd-registry value string (but leave it as syntactically correct JSON)

```console
harv1:~ # cat /etc/rancher/rke2/registries.yaml 
{"configs":{},"mirrors":{"cr.example.com":{}}}
```

As you can see, using the containerd-registry value carried forward from a release of Harvester < 1.3.0-rc2 resulted in losing part of the setting.

**Expected behavior**

/etc/rancher/rke2/registries.yaml contains semantically the same information as it did prior to upgrading Harvester >= 1.3.0-rc2.

**Support bundle**

**Environment**
 - Harvester ISO version: 1.2.1 upgrade to 1.3.0-rc3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): QEMU/KVM

**Additional context**

Related issue: https://github.com/harvester/harvester/issues/5187

connorkuehl:
Just waiting on the 1.2 backport to merge before moving to Ready for Testing.

albinsun:
Provide test result on `v1.2-a822a3a4-head` -> `v1.3-4372cbc4-head`, close as fixed.


## Environment
* Harvester
[...]

Issue #5187, opened 2024-02-21
[BUG] containerd-registry setting doesn't work


bk201:
**Describe the bug**

Setting the [containerd-registry](https://docs.harvesterhci.io/v1.2/advanced/index/#containerd-registry) setting doesn't work in V1.3.0-rc3.

**To Reproduce**
1. Go to settings.
2. Edit the `containerd_registry` setting:
    ![截圖 2024-02-21 下午5 13 26](https://github.com/harvester/harvester/assets/1691518/4a8a9818-22cc-4875-bbc4-3ecc1105d2a6)
3. Save.
4. log in to nodes and check the `/etc/rancher/rke2/registries.yaml` file, the endpoint is not set correctly:
    ```
    # cat /etc/rancher/rke2/registries.yaml
    {"configs":{},"mirrors":{"docker.io":{}}}
    ```

**Expected behavior**

mirror should be set correctly.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.3.0-rc3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
Not quite sure if this is related to Rancher bump (to v2.8.2).

bk201:
@connorkuehl Can you help check this issue?

connorkuehl:
Harvester's github.com/rancher/wharfie dependency was recently upgraded and it turns out that module introduced a breaking change for the wharfie.Registry struct[1], which is what the Harvester settings controller uses to unmarshal the JSON payload from the Harvester dashboard where the containerd-registry settings are entered into. As a result, the backend has now broken its API contract with the Harvester frontend.

Previously, it did not have any json field tags for the struct members, so the encoding/json package defaults to using the exported field names (`Endpoints` instead of `endpoints`).

The new version of wharfie has added json field tags for these structs. For example, `Endpoints` is now `endpoint`, which is not backwards compatible with previous versions of the module.

This is the payload that the Harvester dashboard sends when saving the value:

```
{\"Mirrors\":{\"cr.example.com\":{\"Endpoints\":[\"https://cr.example.com:5000\"],\"Rewrites\":{}}}}
```

Since the wharfie module changed the struct tags for JSON unmarshaling, it does not use `Endpoints` since it now expects only `endpoint`.

I converted the original HTTP request from the dashboard into a curl request, but this time I updated the JSON field names to use the ones that are in the wharfie module's Registry struct:

```
[...]

PR #5222, opened 2024-02-26
Vendor "frozen" wharfie Registry types

connorkuehl:
Harvester used to maintain and render /etc/rancher/rke2/registries.yaml, and so it relied on the structs in the wharfie/registries package for serializing/deserializing that information. These structs ultimately became the JSON interface between the Harvester dashboard and the settings controller (at least for the containerd-registry setting.)

A newer version of the embedded Rancher took over that responsibility, and now all Harvester has to do is transform the JSON from the frontend and pack it into the RKEConfig in the cluster.provisioning.cattle.io object.

**Problem:**

The wharfie module's registries package recently updated the struct fields with proper JSON tags, but up until this point, the interface that the setting controller shared with the frontend was predicated on those structs and the encoding/json package's default behavior of defaulting to the exported field name's spelling as the JSON key when one was not specified as it was previously.

**Solution:**

Since the wharfie module's only use in Harvester is for the registries package, just vendor those types with their old JSON key representations to avoid breaking the Harvester dashboard, PXE installs, scripts, etc.

**Related Issue:** https://github.com/harvester/harvester/issues/5190 https://github.com/harvester/harvester/issues/5187

**Test plan:**

1. Start on a release prior to 1.3.0-rc2
2. Add a non-default containerd-registry setting:

```console
harv1:~ # kubectl get setting server-version
NAME             VALUE
server-version   v1.2.1
harv1:~ # cat /etc/rancher/rke2/registries.yaml 
{"configs":{},"mirrors":{"cr.example.com":{"endpoint":["https://cr.example.com:5000"]}}}
```

3. Take note of /etc/rancher/rke2/registries.yaml, and the value shown in the Harvester dashboard
4. Upgrade to a release containing this commit
5. Touch the containerd-registry setting (either delete the hash annotation or just add a space somewhere in the containerd-registry value string (but leave it as syntactically correct JSON)
6. Check that /etc/rancher/rke2/registries.yaml and the value shown in the Harvester dashboard are the values prior to the upgrade

```console
harv1:~ # cat /etc/rancher/rke2/registries.yaml 
{"configs":{},"mirrors":{"cr.example.com":{"endpoint":["https://cr.example.com:5000"]}}}
```


FrankYang0529 (approved):
LGTM. I can set a registry and use it to download image in the harvester cluster.

bk201 (approved):
lgtm, verify it works again. Thanks!

bk201:
@mergifyio backport v1.3 v1.2

Issue #5230, opened 2024-02-27
[BUG] Cannot download upgrade logs due to the previous-version harvester-upgrade image missing


starbops:
**Describe the bug**

We introduced the image pruning feature for Harvester upgrades in #4995. However, in air-gapped environments, users are unable to download the upgrade logs from the dashboard after the upgrade ends successfully because the packager Pod is using the old `harvester-upgrade` container image, which no longer exists on the system.

![image](https://github.com/harvester/harvester/assets/1827717/b3508130-3f65-4247-9da7-576129c51236)

**To Reproduce**

1. Prepare a Harvester cluster with v1.2.1
2. Upgrade to v1.3.0-rc3
3. After the upgrade ends, try to download upgrade logs from the Harvester dashboard
4. The packaging Job fails to complete, and you won't get the tarball

**Expected behavior**

Upgrade logs can be downloaded successfully even after the upgrade ends.

**Support bundle**

N/A

**Environment**
 - Harvester ISO version: v1.2.1 to v1.3.0-rc3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

Retagging the `harvester-upgrade` image should did the trick (as a workaround).

TachunLin:
Verifying on `master-b227ca1d-head`  (24/03/04). Close this issue

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ Can download upgrade logs after upgrade from `v1.2.1` to `master-head` $~~$</summary>



https://github.com/harvester/harvester/assets/29251855/ffae9876-0d6c-4b57-8de6-42511e957a08


 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Can download upgrade logs after upgrade from `v1.2.1` to `v1.2-head` $~~$</summary>

https://github.com/harvester/harvester/assets/29251855/c8d8cf08-a4aa-4fb2-a144-d6f513dfb8b3



 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Can download upgrade logs after upgrade from `v1.2-head` to `master-head` $~~$</summary>


https://github.com/harvester/harvester/assets/29251855/9879d236-129b-4ee8-adb3-4a27b46c4dee


https://github.com/harvester/harvester/assets/29251855/ffae9876-0d6c-4b57-8de6-42511e957a08


 </details>

### Test Information
* Test Environment: Single nodes harvester on local kvm machine
[...]

PR #5235, opened 2024-02-27
fix(upgrade): add harvester-upgrade into the retain list

starbops:
**Problem:**

If an upgrade goes well and ends successfully, the old `harvester-upgrade` container image is gone. The upgrade log download request afterward will inevitably fail to be complete because the log packager Job depends on that container image.

**Solution:**

After the upgrade ends, the `harvester-upgrade` container image is immune to the image pruning procedure.

**Related Issue:**

#5230 

**Test plan:**

1. Prepare a Harvester cluster in v1.2.1
2. Upgrade to master-head
3. After the upgrade ends successfully, try to download the upgrade log from the dashboard
4. The upgrade log archive should be generated and downloaded successfully

### DIFF
--- a/pkg/controller/master/upgrade/common.go
+++ b/pkg/controller/master/upgrade/common.go
@@ -781,6 +781,16 @@ func upgradeReference(upgrade *harvesterv1.Upgrade) metav1.OwnerReference {
 	}
 }
 
+// removeItemFromSlice removes one element at index i from the slice. By
+// removing the element, it simply copies the last element in the slice to the
+// slot at index i and returns the same slice but excluding the last element.
+// That is to say, the order of elements in the slice might change, depending
+// on what element is going to be removed.
+func removeItemFromSlice(slice []string, i int) []string {
+	slice[i] = slice[len(slice)-1]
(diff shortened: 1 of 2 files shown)


FrankYang0529 (approved):
LGTM. Thanks.

pkg/controller/master/upgrade/common.go
```diff
@@ -781,6 +781,11 @@ func upgradeReference(upgrade *harvesterv1.Upgrade) metav1.OwnerReference {
 	}
 }
 
+func removeItemFromSlice(slice []string, i int) []string {
```
bk201, line 789: Can you help add a doc comment to the function and explain its feature and result order change

bk201: for long term we can add utility functions in https://github.com/harvester/go-common/blob/main/slice.go too

starbops: Updated, thank you!

bk201 (approved):

starbops:
@mergify backport v1.2 v1.3

PR #5241, opened 2024-02-28
support etcd taints in upgrade plans

ibrokethecloud:

**Problem:**

Currently upgrade from v1.3.0-rc2 to v1.3.0-rc3 fails if one of the nodes is etcd.

This is because the prepare plan which loads images cannot be scheduled on the etcd node due to missing taints.

The job eventually times out since the pod cannot be scheduled.
```
Events:
  Type     Reason            Age   From               Message
  ----     ------            ----  ----               -------
  Warning  FailedScheduling  14s   default-scheduler  0/3 nodes are available: 1 node(s) had untolerated taint {node-role.kubernetes.io/etcd: true}, 2 node(s) didn't match Pod's node
 affinity/selector. preemption: 0/3 nodes are available: 3 Preemption is not helpful for scheduling..
```

**Solution:**

Add additional taints for etcd only nodes in the prepare and clean up plans.

**Related Issue:**
**Test plan:**

To easily test the upgrade path without an RC the following can be done:

1. Install a HA cluster with v1.3.0-rc2
2. The cluster should have atleast one etcd role
3. Patch the harvester managedchart in fleet-local namespace as follows
```
values:
    containers:
      apiserver:
        env:
        # Use this to mock a source version to trigger an upgrade if needed
        - name: HARVESTER_SERVER_VERSION
          value: v1.3.0-rc2
        hciMode: true
        image:
          imagePullPolicy: Always
          repository: gmehta3/harvester 
          tag: upgrade-node-taints-head
```
4. Trigger upgrade to v1.3.0-rc3
5. Upgrade should be successful

### DIFF
--- a/pkg/controller/master/upgrade/common.go
+++ b/pkg/controller/master/upgrade/common.go
@@ -308,6 +308,11 @@ func preparePlan(upgrade *harvesterv1.Upgrade) *upgradev1.Plan {
 					Effect:   corev1.TaintEffectNoSchedule,
 					Value:    "arm",
 				},
+				{
+					Key:      node.KubeEtcdNodeLabelKey,
+					Operator: corev1.TolerationOpExists,
+					Effect:   corev1.TaintEffectNoExecute,
+				},
 			},
 			Upgrade: &upgradev1.ContainerSpec{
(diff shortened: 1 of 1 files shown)


ibrokethecloud:
already addressed by: https://github.com/harvester/harvester/pull/5202

Issue #5248, opened 2024-02-29
[BUG] Can't put a control node into maintenance if there is a witness node


bk201:
**Describe the bug**

Can't put a control node into maintenance if the witness node is presented.

**To Reproduce**
1. Create a 3-node cluster, one should be a witness node.
2. Put any non-witness node into maintenance mode and see this from the GUI:
![截圖 2024-02-29 上午9 26 08](https://github.com/harvester/harvester/assets/1691518/a8d07194-790d-4d9d-8f28-7375cdde4596)

**Expected behavior**

The node should be able to enter maintenance mode since there is one control node and witness node alive.
We need to find tune the selector here: https://github.com/harvester/harvester/blob/cd75150dbe642d12e1612509520cf1deb98241ec/pkg/util/drainhelper/helper.go#L84

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

albinsun:
# Test Plan
## Environment
* Harvester
  * Version: `v1.3-head`
  * Profile: QEMU/KVM, 3 nodes with witness
    * Default: 8C/16G/500G
    * Witness: 4C/8G/500G

## Test Suites
### Prerequisites
1. <details>
   <summary> Create a VM
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/289cf4b2-05cc-4492-a52e-6a081ddfe536)
   </details>

1. Create 3 node (MMW) cluster with witness node
   * <details> 
     <summary> Cluster is Active :question: 
     </summary>

     ![image](https://github.com/harvester/harvester/assets/2773781/63d288b2-2090-4895-b720-49acd2bd322a)
     </details>

   * <details>
     <summary> Create a VM (node-1) :question: 
     </summary>
     
     ![image](https://github.com/harvester/harvester/assets/2773781/a257493e-8d8e-472a-b9bd-4efffa62f366)
     </details>

   
1. Ensable maintenance on node-1
   * <details> 
     <summary> Cluster is Active :question: 
     </summary>

     </details>

   * <details> 
     <summary> node-0 and node-2 should NOT be able to maintenance :question: 
     </summary>

     </details>

   * <details> 
     <summary> Disable maintenance on node-1 :question: 
     </summary>

     </details>


1. Ensable maintenance on node-2
   * <details> 
     <summary> Cluster is Active :question: 
     </summary>

     </details>

[...]

PR #5249, opened 2024-02-29
change controlplane filter to look for etcd presence

ibrokethecloud:

**Problem:**

With the introduction of witness role for worker nodes, we have nodes with only etcd running and no other controlplane / workload components.
This breaks the current calculation logic when trying to place controlplane nodes into maintenance mode.

**Solution:**

The node role for etcd is labelled on all controlplane / witness nodes. As a result this would be a more effective label to filter for controlplane nodes.

**Related Issue:**
https://github.com/harvester/harvester/issues/5248
**Test plan:**

### DIFF
--- a/pkg/util/drainhelper/helper.go
+++ b/pkg/util/drainhelper/helper.go
@@ -6,11 +6,12 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/selection"
+
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/kubectl/pkg/drain"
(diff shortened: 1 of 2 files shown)


pkg/util/drainhelper/helper.go
```diff
@@ -77,30 +78,55 @@ func DrainNode(ctx context.Context, cfg *rest.Config, node *corev1.Node) error {
...
+	if err != nil {
+		return fmt.Errorf("error creating requirement: %v", err)
+	}
+
+	etcdSelector := labels.NewSelector()
+	etcdSelector = etcdSelector.Add(*etcdReq)
+
+	etcdNodeList, err := nodeCache.List(etcdSelector)
+	if err != nil {
+		return fmt.Errorf("error listing nodes matching selector %s: %v", cpSelector.String(), err)
```
connorkuehl:
```suggestion
		return fmt.Errorf("error listing nodes matching selector %s: %v", etcdSelector.String(), err)
```

connorkuehl (approved):
Just a minor copy-paste casualty, otherwise looks great!

pkg/util/drainhelper/helper.go
ibrokethecloud: lol.. good spot.

Vicente-Cheng (approved):
lgtm, thanks!

connorkuehl:
@mergify backport v1.3

Issue #5263, opened 2024-03-01
[BUG] Upgrade is stuck at the apply-manifest phase due to the "skip-rke2-restart" Plan failure


starbops:
**Describe the bug**

During the Harvester upgrade, precisely at the apply-manifest phase, a Plan will be created by the upgrade script to prevent RKE2 on the nodes from being restarted when Rancher Manager is upgrading. RKE2 should only be upgraded (thus restarted) at the upgrade-node phase.

This issue is caught when the `.spec.version` field of the Plan consists of all numeric characters. This could happen because we simply use the `openssl rand -hex 4` command to generate a random string and put it into a YAML template without quotes.

**To Reproduce**

1. Prepare a Harvester cluster (single node would be fine) with master-head
2. Upgrade to v1.3.0-rc3
3. The upgrade should be stuck at the apply-manifest phase when creating the Plan

**Expected behavior**

The "skip-rke2-restart" Plan should be created without issues and the whole upgrade should end successfully.

**Support bundle**

Refer to https://github.com/harvester/harvester/issues/5203#issuecomment-1971326706

**Environment**
 - Harvester ISO version: from master-head upgrading to v1.3.0-rc3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

**Additional context**

The issue was originally found in https://github.com/harvester/harvester/issues/5203#issuecomment-1971326706.

There seem to be no easily-applied workarounds. When users "luckily" hit the bug, the only way is to try to upgrade again.

albinsun:
Close along with #5203.
Thank you.

PR #5265, opened 2024-03-01
fix(upgrade): add quotes around the plan version

starbops:

**Problem:**

There's a chance that the randomly generated plan version consists of all numeric characters.

```shell
harvester-node-0:~ # openssl rand -hex 4
35964120
harvester-node-0:~ # openssl rand -hex 4
7cdd6c6e
```

And the variable is not surrounded by quotes:

```yaml
  ...
  version: $plan_version
  ...
```

**Solution:**

The Plan version `.spec.version` must be of type string. To prevent the randomly generated version consists of all numeric characters, quotes are added around the `plan_version` variable to force it to be parsed as string type.

**Related Issue:**

#5263

**Test plan:**

1. Prepare a Harvester cluster with master-head
2. Upgrade to a new version which contains the fix
3. The upgrade should end successfully


bk201 (approved):

Vicente-Cheng (approved):
nice catch!

Vicente-Cheng:
@mergifyio backport v1.3

bk201:
@mergifyio backport v1.2

PR #5312, opened 2024-03-07
upgrade/template: update nvidia-driver-toolkit tag

Vicente-Cheng:

**Problem:**
we should handle the upgrade part to use the same tag as the baseos image

**Solution:**
update the nvidia driver image tag as the baseos image

**Related Issue:**

**Test plan:**


bk201 (approved):

ibrokethecloud (approved):
lgtm. thanks.

Issue #5069, opened 2024-01-31
[FEATURE] Improving Harvester Maintenance Mode


votdev:
**Is your enhancement related to a problem? Please describe.**
When setting a Harvester host into maintenance mode it will (try to) migrate all virtual machines on that node to other nodes in the cluster.
I want to have the possibility to not migrate all virtual machines on that node to other nodes in the cluster.

**Describe the solution you'd like**
For instance, a radio button where you can prevent the migration of virtual machines in the "Enable Maintenance Mode" pop-up.
Having the option to trigger (via an annotation or webhook) the execution of a script/command inside the VM that then could i.e. remove/disable the application.

**Describe alternatives you've considered**
None

**Additional context**
None

votdev:
# Proposal
- The VMs that do not have to be migrated to other nodes when a node is set into maintenance mode must be labeled with `harvesterhci.io/maintain-force-shutdown`.
- When the maintenance mode is activated for a node, these labeled VMs are shut down to free their resources.
- The VMs are not started automatically if the maintenance mode is disabled for the node. At this point the VMs that have been shut down previously can not be identified because the reference to the owner node has been removed.
- Special actions that have to be done by the VM before shutting down can be implemented via systemd unit files in the VM itself or via the `PrePost` [container lifecycle hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/).
- If maintenance mode is forced via UI (by setting the checkbox in the dialog), this will override the custom settings per VM.

See https://github.com/harvester/harvester/pull/5984 for a up-to-date proposal.

w13915984028:
This will be a good enhancement.

When user plan to put a node to `Maintenance Mode`, if UI opens a new page with more details, it will be better:

Support is sunch a new page, it lists all VMs on this Node per their migration ability:
 - ready to migrate, 
 - cannot migrate e.g. with vGPU or pci-paththrough, with dedicated scheduling policy/label, or VM has only the last active replica
 
 User can click the specific vm to open vm page, and do related actions like stop or migration to a desired node.

And when user clicks the `Maintenance Mode` button, display the migration status and progress of each vm on this page.

votdev:
> This will be a good enhancement.
> 
> When user plan to put a node to `Maintenance Mode`, if UI opens a new page with more details, it will be better:
> 
> Support is sunch a new page, it lists all VMs on this Node per their migration ability:
> 
>     * ready to migrate,
> 
[...]

PR #5328, opened 2024-03-08
Improving Harvester Maintenance Mode

votdev:
VMs that should not be migrated to other nodes in maintenance mode have to labeled with `harvesterhci.io/maintain-mode-strategy`. These VMs are shut down properly to release the resources and allow the node to go into maintenance mode. There are several strategies:

- `Migrate`: The VM will be live-migrated to another node in the cluster. This is the default behavior if the label `harvesterhci.io/maintain-mode-strategy` is not set.
- `ShutdownAndRestartAfterEnable`: Restart the VM after the node has finally switched into maintenance mode. The VM will be scheduled on a different node.
- `ShutdownAndRestartAfterDisable`: Shut down when maintenance mode is enabled and restart the VM after maintenance mode is disabled. The VM will stay on the same node.
- `Shutdown`: Shut down when maintenance mode is enabled. Do NOT restart the VM, it remains switched off.

The HEP can be found [here](https://github.com/harvester/harvester/pull/5984).

You may also check the associated [documentation PR](https://github.com/harvester/docs/pull/523).

Peek 2024-03-19 14-42

Signed-off-by: Volker Theile <vtheile@suse.com>

**Problem:**
There are situations where it is necessary to shut down several VMs on a node instead of live-migrating them to another node when the maintenance mode is activated.

**Solution:**
Introduce a label that can be used to specify how a VM should behave when the node it is running on is going into maintenance mode.

**Related Issue:**
#5069

**Test plan:**
## Case 1:
1. Add the `harvesterhci.io/maintain-mode-strategy` label to a VM of your choice. Press `Edit YAML` and add `harvesterhci.io/maintain-mode-strategy: ShutdownAndRestartAfterDisable` to `metadata.labels`. Save the settings.
2. Make sure a second VM without the `harvesterhci.io/maintain-mode-strategy` label is running on the same node. Migrate it to the node on which the other VM is running if necessary.
3. Note the node on which the VMs are running.
4. Go to the `Hosts` page and select the `Enable Maintenance Mode` action in the actions dropdown menu of the node your VMs are running on.
5. After some time the `Cordon` state should be displayed for the host we are putting into maintenance mode.
6. Now go to the `Virtual Machines` page. The VM with the `harvesterhci.io/maintain-mode-strategy` label should show the `Stopping` status badge. The other VM should show the `Migrate` status badge.
7. After some time the VMs should be shut down or migrated to another node. Choose the `Edit YAML` action menu on the powered off VM and make sure the `harvesterhci.io/maintain-mode-strategy-node-name: <NODENAME>` is in the annotations.
8. Go back to the `Hosts` page. The host of choice should now (or after some time) display the `Maintenance` status badge.
9. Disable the maintenance mode via the action menu of the node. After some time the `Maintenance Mode` status badge should be gone away.
10. Go to the `Virtual Machines` page. The previously shut down VM with the `harvesterhci.io/maintain-mode-strategy: ShutdownAndRestartAfterDisable` label should be started again.

In the logs of the `harvester-xxx-xxx` pods you should find something like this:
```
time="2024-03-19T12:45:24Z" level=info msg="attempting to place node in maintenance mode" node_name=harvester-node-2
time="2024-03-19T12:45:24Z" level=info msg="force stopping VM" namespace=default virtualmachine_name=test01
time="2024-03-19T12:45:24Z" level=info msg="migration of pod owned by VM test01 is skipped because of the label harvesterhci.io/maintain-mode-strategy" namespace=default pod_name=virt-launcher-test01-gfgjp
time="2024-03-19T12:55:37Z" level=info msg="restarting the VM that was shut down in maintenance mode" namespace=default virtualmachine_name=test01
```

## Case 2:
1. Add the `harvesterhci.io/maintain-mode-strategy` label to a VM of your choice. Press `Edit YAML` and add `harvesterhci.io/maintain-mode-strategy: ShutdownAndRestartAfterEnable` to `metadata.labels`. Save the settings.
2. Make sure a second VM without the `harvesterhci.io/maintain-mode-strategy` label is running on the same node. Migrate it to the node on which the other VM is running if necessary.
3. Note the node on which the VMs are running.
4. Go to the `Hosts` page and select the `Enable Maintenance Mode` action in the actions dropdown menu of the node your VMs are running on.
5. After some time the `Cordon` state should be displayed for the host we are putting into maintenance mode.
6. Now go to the `Virtual Machines` page. The VM with the `harvesterhci.io/maintain-mode-strategy` label should show the `Stopping` status badge. The other VM should show the `Migrate` status badge.
7. Choose the `Edit YAML` action menu on the powered off VM and make sure the `harvesterhci.io/maintain-mode-strategy-node-name: <NODENAME>` is in the annotations.
8. Go back to the `Hosts`page. Wait until the `Maintenance`status badge is shown.
9. On the `Virtual Machines` page, the VM that was shut down should be restarted and running on a different node than before. Note, this VM was NOT migrated; it was shutdown and restarted.
10. On the `Hosts` page, disable the maintenance mode via the action menu of the node. After some time the `Maintenance Mode` status badge should be gone away.
11. Go to the `Virtual Machines` page. The VMs should still run and the node on which they run should not have changed again.

In the logs of the `harvester-xxx-xxx` pods you should find something like this:
```
time="2024-03-19T13:00:30Z" level=info msg="attempting to place node in maintenance mode" node_name=harvester-node-1
time="2024-03-19T13:00:30Z" level=info msg="force stopping VM" namespace=default virtualmachine_name=test01
time="2024-03-19T13:00:31Z" level=info msg="migration of pod owned by VM test01 is skipped because of the label harvesterhci.io/maintain-mode-strategy" namespace=default pod_name=virt-launcher-test01-rsvwf
time="2024-03-19T13:01:19Z" level=info msg="restarting the VM that was temporary shut down for maintenance mode" namespace=default virtualmachine_name=test01
```

## Case 3:
1. Add the `harvesterhci.io/maintain-mode-strategy` label to a VM of your choice. Press `Edit YAML` and add `harvesterhci.io/maintain-mode-strategy: Shutdown` to `metadata.labels`. Save the settings.
2. Make sure a second VM without the `harvesterhci.io/maintain-mode-strategy` label is running on the same node. Migrate it to the node on which the other VM is running if necessary.
3. Note the node on which the VMs are running.
4. Go to the `Hosts` page and select the `Enable Maintenance Mode` action in the actions dropdown menu of the node your VMs are running on.
5. After some time the `Cordon` state should be displayed for the host we are putting into maintenance mode.
6. Now go to the `Virtual Machines` page. The VM with the `harvesterhci.io/maintain-mode-strategy` label should show the `Stopping` status badge. The other VM should show the `Migrate` status badge.
7. Choose the `Edit YAML` action menu on the powered off VM and make sure the `harvesterhci.io/maintain-mode-strategy-node-name: <NODENAME>` is NOT in the annotations.
8. Go back to the `Hosts` page. Wait until the `Maintenance`status badge is shown.
9. On the `Virtual Machines` page, the VM with the `harvesterhci.io/maintain-mode-strategy` label should be off. The other VM should be migrated to another node.
10. On the `Hosts` page, disable the maintenance mode via the action menu of the node. After some time the `Maintenance Mode` status badge should be gone away.
11. On the `Virtual Machines` page, the VM with the `harvesterhci.io/maintain-mode-strategy` label should be still off.

In the logs of the `harvester-xxx-xxx` pods you should NOT find something like this:
```
time="2024-03-19T12:55:37Z" level=info msg="restarting the VM that was ... shut down in maintenance mode" namespace=default virtualmachine_name=test01
```

## Case 4:
1. Add the `harvesterhci.io/maintain-mode-strategy` label to a VM of your choice. Press `Edit YAML` and add `harvesterhci.io/maintain-mode-strategy: ShutdownAndRestartAfterDisable` to `metadata.labels`. Save the settings.
2. Make sure a second VM without the `harvesterhci.io/maintain-mode-strategy` label is running on the same node. Migrate it to the node on which the other VM is running if necessary.
3. Note the node on which the VMs are running.
4. Go to the `Volume` page in the embedded Longhorn UI and reduce the number of replicas to 1 for the volume that belongs to the VM that has no maintenance mode strategy. Make sure that you delete all replicas that are not located on the node that is to be put into maintenance mode.
5. Go to the `Hosts` page in the Harvester UI and select the `Enable Maintenance Mode` action in the actions dropdown menu of the node your VMs are running on. Check the `Force` checkbox. Note that this overrides the maintenance mode strategy of the VMs (which is to be tested here).
6. After some time the `Cordon` state should be displayed for the host we are putting into maintenance mode.
7. Now go to the `Virtual Machines` page. Both VMs should show the `Stopping` status badge and should be both `Off` after some time.
8. Disable the maintenance mode on the `Hosts` page via the action menu of the node. After some time the `Maintenance Mode` status badge should be gone away.
9. Go to the `Virtual Machines` page. Both VMs should be still off.
10. Choose the `Edit YAML` action menu on the VM that was labeled with `harvesterhci.io/maintain-mode-strategy: ShutdownAndRestartAfterDisable` and make sure the `harvesterhci.io/maintain-mode-strategy-node-name: <NODENAME>` is NOT in the annotations.
11. Note that in this test case, the node cannot switch to `Maintenance Mode` because several pods cannot be drained due to the deleted replicas of the Longhorn volume.

In the logs of the `harvester-xxx-xxx` pods you should NOT find something like this:
```
time="2024-03-19T12:55:37Z" level=info msg="restarting the VM that was ... shut down in maintenance mode" namespace=default virtualmachine_name=test01
```

## Case 5:
1. Add the `harvesterhci.io/maintain-mode-strategy` label to a VM of your choice. Press `Edit YAML` and add `harvesterhci.io/maintain-mode-strategy: ShutdownAndRestartAfterDisable` to `metadata.labels`. Save the settings.
2. Note the node on which the VM are running.
3. Go to the `Hosts` page and select the `Enable Maintenance Mode` action in the actions dropdown menu of the node your VM is running on.
4. After some time the `Cordon` state should be displayed for the host we are putting into maintenance mode.
5. Now go to the `Virtual Machines` page. The VM with the `harvesterhci.io/maintain-mode-strategy` label should show the `Stopping` status badge.
6. After some time the VM should be shut down. Choose the `Edit YAML` action menu on the powered off VM and make sure the `harvesterhci.io/maintain-mode-strategy-node-name: <NODENAME>` is in the annotations.
7. Go back to the `Hosts` page. The host of choice should now (or after some time) display the `Maintenance` status badge.
8. Delete this node. This might take some time. When it is done, go to the `Virtual Machines` page.
9. Choose the `Edit YAML` menu on the shut-down VM. The `harvesterhci.io/maintain-mode-strategy-node-name: <NODENAME>` annotation should not exist anymore.


FrankYang0529:
Hi @votdev, I take a look for the code and try the feature. I have a question: why we want to set `LabelMaintainForceShutdownStrategy` and `AnnotationMaintainForceShutdownNodeName` at different level?

The `LabelMaintainForceShutdownStrategy` is on VMI and `AnnotationMaintainForceShutdownNodeName` is on VM. I think we can set both key on VM annotation. If we want to use listing by filter, we can set them on VM labels.

votdev:
> Hi @votdev, I take a look for the code and try the feature. I have a question: why we want to set `LabelMaintainForceShutdownStrategy` and `AnnotationMaintainForceShutdownNodeName` at different level?
> 
> The `LabelMaintainForceShutdownStrategy` is on VMI and `AnnotationMaintainForceShutdownNodeName` is on VM. I think we can set both key on VM annotation. If we want to use listing by filter, we can set them on VM labels.

`LabelMaintainForceShutdownStrategy` is implemented as an instance label on `VirtualMachine` as the user can enter this in the Harvester UI.

It also has the advantage that these instance labels are automatically copied from `VirtualMachine` to the `VirtualMachineInstance` as labels. So i don't have to take care about it. See https://github.com/harvester/harvester/pull/5328/files#diff-668050f872b498d877736c7b008f2bdf89bbd48800f9ef83d2d3da8e57a1d872R101 where a query on VMIs is done.

I would like to use labels for this, as the user can easily set them to the corresponding VM objects using a command line tool, but there is no UI support for this in Harvester. For labels, I would also have to make sure that they are copied to the instances myself.

Finally, `VirtualMachineInstance` objects with specific labels are searched for in various places across the code. You cannot search for annotations in Kubernetes.

For these reasons above, I have implemented `LabelMaintainForceShutdownStrategy` as a label.

`AnnotationMaintainForceShutdownNodeName`is implemented as an annotation because I do not have to search for this information and it is only metadata/additional data. In other words, exactly what annotations are intended for.

votdev:
> > Hi @votdev, I take a look for the code and try the feature. I have a question: why we want to set `LabelMaintainForceShutdownStrategy` and `AnnotationMaintainForceShutdownNodeName` at different level?
> > The `LabelMaintainForceShutdownStrategy` is on VMI and `AnnotationMaintainForceShutdownNodeName` is on VM. I think we can set both key on VM annotation. If we want to use listing by filter, we can set them on VM labels.
> 
> `LabelMaintainForceShutdownStrategy` is implemented as an instance label on `VirtualMachine` as the user can enter this in the Harvester UI.
> 
> It also has the advantage that these instance labels are automatically copied from `VirtualMachine` to the `VirtualMachineInstance` as labels. So i don't have to take care about it. See [#5328 (files)](https://github.com/harvester/harvester/pull/5328/files#diff-668050f872b498d877736c7b008f2bdf89bbd48800f9ef83d2d3da8e57a1d872R101) where a query on VMIs is done.

> 
> I would like to use labels for this, as the user can easily set them to the corresponding VM objects using a command line tool, but there is no UI support for this in Harvester. For labels, I would also have to make sure that they are copied to the instances myself.
> 
> Finally, `VirtualMachineInstance` objects with specific labels are searched for in various places across the code. You cannot search for annotations in Kubernetes.
> 
> For these reasons above, I have implemented `LabelMaintainForceShutdownStrategy` as a label.
> 
> `AnnotationMaintainForceShutdownNodeName”`is implemented as an annotation because I do not have to search for this information and it is only metadata/additional data. In other words, exactly what annotations are intended for.

ibrokethecloud (approved):
lgtm. thanks.

bk201:
I suggest we can simply the labelling things by some UI widgets, create an issue to track so UI team can implement it: https://github.com/harvester/harvester/issues/5888

pkg/controller/master/node/maintain_controller.go
```diff
@@ -51,17 +61,87 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
...
+		return node, nil
+	}
+
+	vms, err := h.virtualMachineCache.List(corev1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return node, fmt.Errorf("failed to list VMs: %w", err)
+	}
+
+	for _, vm := range vms {
+		if vm.Annotations[util.AnnotationMaintainForceShutdownNodeName] != node.Name {
```
bk201: It would be nice to check whether `vm.Annotations` is null.

FrankYang0529: > `AnnotationMaintainForceShutdownNodeNameis` implemented as an annotation because I do not have to search for this information and it is only metadata/additional data. In other words, exactly what annotations are intended for.

Actually, you do search for it here, right? If you set it on VM label and we can do this part like:

```go
	vms, err := h.virtualMachineCache.List(corev1.NamespaceAll, labels.Set{"harvesterhci.io/maintain-force-shutdown-node-name": node.Name}.AsSelector())
	if err != nil {
		return node, fmt.Errorf("failed to list VMs: %w", err)
	}

	for _, vm := range vms {
		vmCopy := vm.DeepCopy()
		delete(vmCopy.Annotations, util.AnnotationMaintainForceShutdownNodeName)
		_, err = h.virtualMachineClient.Update(vmCopy)
		if err != nil {
			return node, err
		}
	}
```

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -50,3 +51,20 @@ func IsVMStopped(
...
+// ListByInstanceLabels gets a list of VMs whose instance labels match the
+// specified labels.
+func ListByInstanceLabels(namespace string, selector labels.Selector, cache ctlkubevirtv1.VirtualMachineCache) ([]*kubevirtv1.VirtualMachine, error) {
+	vmList, err := cache.List(namespace, labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+	vmMatchList := make([]*kubevirtv1.VirtualMachine, 0, len(vmList))
+	for _, vm := range vmList {
+		if !selector.Matches(labels.Set(vm.Spec.Template.ObjectMeta.Labels)) {
```
FrankYang0529: Why not set label on VM directly, so we don't need to search the label by ourselves? Also, original design takes O(N) time complexity to search all VMs in the namespace.

```go
vmList, err := cache.List(namespace, labels.Set{util.LabelMaintainForceShutdownStrategy: "RestartOnDisable"}.AsSelector())
```

BTW, if we design the label on VMI, users need to restart VM to let it take effect.

https://github.com/harvester/harvester/assets/4927361/2391270c-4030-4867-895d-c2ee4392a1ff

FrankYang0529 (commented):
Some design questions.

pkg/util/virtualmachine/virtualmachine.go
FrankYang0529: > I would like to use labels for this, as the user can easily set them to the corresponding VM objects using a command line tool, but there is no UI support for this in Harvester.

We can add an action for it like @bk201 mentioned in https://github.com/harvester/harvester/issues/5888.

> For labels, I would also have to make sure that they are copied to the instances myself.

We already have a controller for it.

https://github.com/harvester/harvester/blob/5ee0329f1a6acef48e52758d3caf1d19e3bac1aa/pkg/controller/master/virtualmachine/vm_controller.go#L167-L185

votdev: I'm currently adapting the code which will not use instance labels any more. Making sure that my special label is copied to VMI justifies the effort to then have the option of setting the label using kubectl and filters for the desired VMs.

> BTW, if we design the label on VMI, users need to restart VM to let it take effect.

I think this is not right because the labels are always synced on changes on the VM.
- https://github.com/harvester/harvester/blob/f9464f06692f046f07b5bd5d9cbea3ed182c1610/pkg/controller/master/virtualmachine/register.go#L67
- https://github.com/harvester/harvester/blob/f9464f06692f046f07b5bd5d9cbea3ed182c1610/pkg/controller/master/virtualmachine/vm_controller.go#L168

votdev: > BTW, if we design the label on VMI, users need to restart VM to let it take effect.

That is only true when using `Edit Config`. If you use `Edit YAML`, then the labels will be sync-ed instantly and no warning dialog is displayed.

FrankYang0529 (approved):
All test cases work good. However, the case 4 step 8 may be typo. The VM without label will be migrated to other node. Only VM with label should be still off. @votdev could you double check for it? Thank you.

votdev:
> All test cases work good. However, the case 4 step 8 may be typo. The VM without label will be migrated to other node. Only VM with label should be still off. @votdev could you double check for it? Thank you.

I've adapted the test case 4. 

The intention of this test case is to let the existing old code path apply to the unlabeled VM. To let the old `Force` maintenance mode code path shut down the VM, the replica count of the associated LH volume must be 1.

bk201 (approved):
LGTM, @votdev Please see if we need to squash the commits before merging.

votdev:
> LGTM, @votdev Please see if we need to squash the commits before merging.

Good idea. I've squashed the commits, rebased with `upstream/master` and adapted the commit message.

PR #5344, opened 2024-03-12
Fix README of vm-live-migrate-detector

votdev:
**Problem:**
The description of the `--shutdown` option is exactly the opposite of what the option actually does.

**Solution:**
Fix the README.

**Related Issue:**
https://github.com/harvester/harvester/pull/4984

**Test plan:**
n/a


starbops (approved):
Thanks for the fix! Maybe we could drop the "Do" word altogether.

bk201 (approved):

Issue #5137, opened 2024-02-12
[BUG] After deleting the only VM of LB, LB controller reports following error


w13915984028:
**Describe the bug**

**To Reproduce**
1. Create a IPPool
2. Create one VM
3. Create one LB with workload type VM and selecting the VM in step 2
4. After a few minutes, delete the VM
5. In LB POD, such errors are continuously reported

> time="2024-02-08T16:54:39Z" level=error msg="error syncing 'default/vm1': handler harvester-lb-vmi-controller: admission webhook \"virtualmachineinstances-update-validator.kubevirt.io\" denied the request: update of VMI object is restricted, requeuing"

**Expected behavior**

**Support bundle**

```
LB with selected VM

# kubectl get loadbalancer -A
NAMESPACE   NAME   DESCRIPTION   WORKLOADTYPE   IPAM   ADDRESS       AGE
default     lb1                  vm             pool   192.168.5.5   45s

# kubectl get loadbalancer -A -oyaml
apiVersion: v1
items:
- apiVersion: loadbalancer.harvesterhci.io/v1beta1
  kind: LoadBalancer
  metadata:
    annotations:
      loadbalancer.harvesterhci.io/namespace: default
      loadbalancer.harvesterhci.io/network: ""
      loadbalancer.harvesterhci.io/project: ""
    creationTimestamp: "2024-02-08T16:56:48Z"
    finalizers:
    - wrangler.cattle.io/harvester-lb-controller
    generation: 2
    name: lb1
    namespace: default
    resourceVersion: "80926"
    uid: 9ea1730d-e933-49c2-9e02-8f855ca8f206
  spec:
    backendServerSelector:
      harvesterhci.io/ippool:
      - test
    healthCheck: {}
    ipPool: test-pool
    ipam: pool
    listeners:
    - backendPort: 80
      name: http
      port: 8080
      protocol: TCP
    workloadType: vm
  status:
    address: 192.168.5.5
    allocatedAddress:
      gateway: 192.168.5.1
      ip: 192.168.5.5
      ipPool: test-pool
      mask: 255.255.255.0
    backendServers:
    - 10.52.0.135
    conditions:
    - lastUpdateTime: "2024-02-08T16:56:49Z"
      status: "True"
      type: Ready
kind: List
metadata:
  resourceVersion: ""

 # kubectl get ippool.loadbalancer.harvesterhci.io -A -oyaml
apiVersion: v1
items:
[...]

PR #5349, opened 2024-03-12
Bump harvester-load-balancer to v0.2.5

w13915984028:

MUST wait https://github.com/harvester/charts/pull/234 is merge at first.

**Problem:**

Fix issue https://github.com/harvester/harvester/issues/5137

**Solution:**

After https://github.com/harvester/charts/pull/234, we also need to sync it to harvester deployment

**BUT**, it seems harvester uses the `master-head` image tag, such sync looks not necessary when only bump image tag.

**Related Issue:**
https://github.com/harvester/harvester/issues/5137

**Test plan:**

### DIFF
--- a/deploy/charts/harvester/Chart.lock
+++ b/deploy/charts/harvester/Chart.lock
@@ -22,7 +22,7 @@ dependencies:
   version: 0.4.2
 - name: harvester-load-balancer
   repository: https://charts.harvesterhci.io
-  version: 0.2.0
+  version: 0.2.5
 - name: whereabouts
   repository: file://dependency_charts/whereabouts
   version: 0.1.1
--- a/deploy/charts/harvester/Chart.yaml
+++ b/deploy/charts/harvester/Chart.yaml
@@ -46,7 +46,7 @@ dependencies:
     version: 0.4.2
     repository: https://kube-vip.github.io/helm-charts
   - name: harvester-load-balancer
-    version: 0.2.0
+    version: 0.2.5
     repository: https://charts.harvesterhci.io
   - name: whereabouts
     version: 0.1.1


w13915984028:
This PR looks to be redundant, unless the chart itself is changed, new commits to controller master-head will be auto fetched by the new ISO.

@bk201 please help double confirm, thank.

Refer: https://github.com/harvester/harvester/issues/5137#issuecomment-1992664008

///
After PR https://github.com/harvester/load-balancer-harvester/pull/26 is merged, a new `master-head`  image of `harvester-load-balancer` is built.

The harvester/harvester-installer will always use the `master-head` image of `harvester-load-balancer` to build-iso. With a latest local build ISO, the issue is verifyed to be solved.

https://github.com/harvester/harvester/blob/a2c98e9606eec79c839c5bed095cc32fc1616b3c/deploy/charts/harvester/values.yaml#L453
```
harvester-load-balancer:
  enabled: false
  image:
    imagePullPolicy: Always
    repository: rancher/harvester-load-balancer
    tag: master-head
  webhook:
    image:
      imagePullPolicy: IfNotPresent
      repository: rancher/harvester-load-balancer-webhook
      tag: master-head
```

starbops (commented):
If we decided to bump the harvester-load-balancer chart to 0.2.5, shall we also pack and put the new chart `harvester-load-balancer-0.2.5.tgz` under the `deploy/charts/harvester/charts` directory?

w13915984028:
The `harvester-load-balancer` chart is not used out of Harvester cluster. Instead of releasing a new chart version, just bump the image tag in Harvester v1.3 branch. 

This PR is replaced by https://github.com/harvester/harvester/pull/5354

Issue #5288, opened 2024-03-04
[BUG] Download Upgrade Image Loops When `isoChecksum` is bad


irishgordo:
**Describe the bug**
When upgrading a Harvester cluster from v1.2-head -> v1.3-head with custom version.yaml object for Version CRD in harvester-system namespace - the Download Upgrade Image will continue to loop, not letting the user know the isoChecksum is bad.

**To Reproduce**
1. Build a Version CRD Object, but provide the isoChecksum of the net-install instead of the amd64 iso from v1.3-head's sha512sum
2. Upgrade the cluster
3. Download Upgrade Image loops multiple times, with no mention to user that the isoChecksum doesn't match

**Expected behavior**
The Download Upgrade Image, before "looping" will let the user know there is a discrepancy in the Version CRD Object's isoChecksum value.  And not just failing silently without notifying the user the checksums didn't match prior to another Download attempt.

**Environment**
 - Harvester ISO version: v1.2-head -> v1.3-head
 - Underlying Infrastructure: 3 HP DL160s Gen 9s

**Workaround:**
- Don't put in the improper value for the isoChecksum in the Version CRD Object that's created

albinsun:
Seems hit same issue at `v1.3-4372cbc4-head`
It stuck in `Downloading` and need about 10mins for checksum fail.
Furthermore, user can (but should not) use it create VM during `Downloading`. 
![image](https://github.com/harvester/harvester/assets/2773781/b345be99-bf2a-4d4b-ac15-4cb687f7e936)

brandboat:
> When upgrading a Harvester cluster from v1.2-head -> v1.3-head with custom version.yaml object for Version CRD in harvester-system namespace - the Download Upgrade Image will continue to loop, not letting the user know the isoChecksum is bad.

The "loop" in this context refers to the retry mechanism found within the code at https://github.com/harvester/harvester/blob/master/pkg/controller/master/image/vm_image_controller.go#L273. During each iteration of this retry process, it does not update the error message and status on the upgrade custom resource unless the number of retries surpasses the maximum limit. Consequently, users will not observe any error notifications in the UI until all retries have been exhausted and failed.

~A potential solution I've suggested in #5375 involves eliminating retries in cases where the ISO checksum verification fails.~ As Kiefer mentioned in https://github.com/harvester/harvester/pull/5375#issuecomment-1998997031, I'll try to use another approach to solve this

c.c. @FrankYang0529

lanfon72:
[...]

PR #5375, opened 2024-03-15
Do not retry vmImage when checksum is bad

brandboat:
**Problem:**
When backing image CR checksum is bad, virtualmachine image still retry 3 times

**Solution:**
Make virtualmachine image not retry to download when checksum is bad.

**Related Issue:**
#5288

**Test plan:**
1. Build a Version CRD Object, but provide the incorrect isoChecksum
2. Upgrade the cluster
3. Check that upgrade failed in first round and not retry again.


bk201:
From the issue statement, I think @irishgordo asks to display the checksum error when retrying. Not to eliminate the retries.
After all, a checksum error might indicate that there is something wrong during data transfer and worth retrying.

brandboat:
superseded by https://github.com/harvester/harvester/pull/5386

Issue #4905, opened 2023-12-22
[ENHANCEMENT] Harvester should display and honor the "virtual size" of uploaded qcow2 images


Martin-Weiss:
**Is your enhancement related to a problem? Please describe.**

In case we upload the SLE Micro 5.5 image (SLE-Micro.x86_64-5.5.0-Default-qcow-GM.qcow2) and create a new VM based on this image with default settings - the VM crashes during bootup.
Reason for this is that the default OS disk size in Harvester is 10 GB but the OS in this case requires 13 GB at least.
Because Harvester in this case "shrinks" below the minimum required - it causes the OS not to work.

` qemu-img info SLE-Micro.x86_64-5.5.0-Default-qcow-GM.qcow2`

```
image: SLE-Micro.x86_64-5.5.0-Default-qcow-GM.qcow2
file format: qcow2
virtual size: 20 GiB (21474836480 bytes)
disk size: 955 MiB
cluster_size: 65536
Format specific information:
    compat: 1.1
    compression type: zlib
    lazy refcounts: false
    refcount bits: 16
    corrupt: false
    extended l2: false

```

**Describe the solution you'd like**
Within a qcow2 image we can see the "virtual size" of an image - this should be taken as default for the OS image size at least. In case the admin selects something smaller - a warning should show up that there is a risk for problems if the OS disk is created smaller.

**Describe alternatives you've considered**
Manual adjustment of OS disk size during VM creation - this works but is not very user-friendly as Harvester does not give any idication of "more space required for the OS".

**Additional context**
Also the image overview in Harvester should show the "virtual size" in additon to the current "disk size" to help administrators.

Vogtinator:
> Because Harvester in this case "shrinks" below the minimum required - it causes the OS not to work.

That's put very mildly. Harvester litererally deletes OS data from the disk in this case.

Vicente-Cheng:
I thought that would be helpful and prevent the user from being confused about the VM crash.

WDYT?  cc @bk201, @markhillgit

tserong:
I hit the same issue experimenting with the SLE Minimal VM image (_not_ the Cloud one):

```
> qemu-img info SLES15-SP5-Minimal-VM.x86_64-kvm-and-xen-QU1.qcow2
image: SLES15-SP5-Minimal-VM.x86_64-kvm-and-xen-QU1.qcow2
file format: qcow2
virtual size: 24 GiB (25769803776 bytes)
disk size: 255 MiB
[...]

PR #5387, opened 2024-03-18
Expose virtual size of VM images

tserong:
This requires:

- https://github.com/longhorn/backing-image-manager/pull/208
- https://github.com/longhorn/longhorn-manager/pull/2680

With this change, we can do this:

```
# kubectl get vmimages
NAME          DISPLAY-NAME                                         SIZE          VIRTUALSIZE   AGE
image-26r4s   sles15-sp5-minimal-vm.x86_64-kvm-and-xen-qu1.qcow2   267845632     25769803776   6m36s
image-bgnhb   sle-micro.x86_64-5.5.0-default-qcow-gm.qcow2         1001652224    21474836480   20m
image-nmmvj   sle-15-sp5-full-x86_64-gm-media1.iso                 14548992000   14548992000   19m
```

Related issue: https://github.com/harvester/harvester/issues/4905

### DIFF
--- a/pkg/controller/master/image/backing_image_controller.go
+++ b/pkg/controller/master/image/backing_image_controller.go
@@ -37,7 +37,26 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
 	} else if err != nil {
 		return nil, err
 	}
-	if !harvesterv1beta1.ImageInitialized.IsTrue(vmImage) || !harvesterv1beta1.ImageImported.IsUnknown(vmImage) {
+	// There are two states that we care about here:
+	// - ImageInitialized
+	// - ImageImported
+	// If ImageInitialized isn't yet true, it means there's no backing
+	// image or storage class, so we've got nothing to work with yet and
+	// should return immediately.
+	if !harvesterv1beta1.ImageInitialized.IsTrue(vmImage) {
+		return nil, nil
+	}
+	// If ImageImported is not unknown, it means the backing image has
+	// been imported, and we think we know everything about it, i.e. we've
+	// now been through a series of progress updates during image download,
+	// and those are finally done, so let's not worry about further updates.
+	// The problem with this logic is that when we add new fields (e.g.
+	// VirtualSize), existing images won't pick up those newly added fields
+	// if we return here immediately.  So, now there's an additional check
+	// for that new field.  Another, simpler, alternative would be to just
+	// drop the ImageImported.IsUnknown check entirely, and let the following
+	// loop run through on every OnChanged event.
+	if !harvesterv1beta1.ImageImported.IsUnknown(vmImage) && vmImage.Status.VirtualSize == backingImage.Status.VirtualSize {
 		return nil, nil
 	}
 	toUpdate := vmImage.DeepCopy()
@@ -46,11 +65,16 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
 			toUpdate = handleFail(toUpdate, condition.Cond(harvesterv1beta1.ImageImported), fmt.Errorf(status.Message))
 			toUpdate.Status.Progress = status.Progress
 		} else if status.State == lhv1beta2.BackingImageStateReady || status.State == lhv1beta2.BackingImageStateReadyForTransfer {
-			harvesterv1beta1.ImageImported.True(toUpdate)
-			harvesterv1beta1.ImageImported.Reason(toUpdate, "Imported")
-			harvesterv1beta1.ImageImported.Message(toUpdate, status.Message)
 			toUpdate.Status.Progress = status.Progress
 			toUpdate.Status.Size = backingImage.Status.Size
+			toUpdate.Status.VirtualSize = backingImage.Status.VirtualSize
+			if toUpdate.Status.VirtualSize > 0 {
+				// We can't set ImageImported to True until we know the VirtualSize,
+				// which may happen one update after the image becomes ready.
+				harvesterv1beta1.ImageImported.True(toUpdate)
+				harvesterv1beta1.ImageImported.Reason(toUpdate, "Imported")
+				harvesterv1beta1.ImageImported.Message(toUpdate, status.Message)
+			}
 		} else if status.Progress != toUpdate.Status.Progress {
 			harvesterv1beta1.ImageImported.Unknown(toUpdate)
 			harvesterv1beta1.ImageImported.Reason(toUpdate, "Importing")
--- a/pkg/controller/master/image/backing_image_controller.go
+++ b/pkg/controller/master/image/backing_image_controller.go
@@ -64,17 +64,16 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
 		if status.State == lhv1beta2.BackingImageStateFailed {
 			toUpdate = handleFail(toUpdate, condition.Cond(harvesterv1beta1.ImageImported), fmt.Errorf(status.Message))
 			toUpdate.Status.Progress = status.Progress
-		} else if status.State == lhv1beta2.BackingImageStateReady || status.State == lhv1beta2.BackingImageStateReadyForTransfer {
+		} else if status.State == lhv1beta2.BackingImageStateReady {
+			// We can't set ImageImported to True until we know the VirtualSize,
+			// which will happen only after stats.State == lhv1beta2.BackingImageStateReady
+			// (it's not there yet for lhv1beta2.BackingImageStateReadyForTransfer)
+			harvesterv1beta1.ImageImported.True(toUpdate)
+			harvesterv1beta1.ImageImported.Reason(toUpdate, "Imported")
+			harvesterv1beta1.ImageImported.Message(toUpdate, status.Message)
 			toUpdate.Status.Progress = status.Progress
 			toUpdate.Status.Size = backingImage.Status.Size
 			toUpdate.Status.VirtualSize = backingImage.Status.VirtualSize
-			if toUpdate.Status.VirtualSize > 0 {
-				// We can't set ImageImported to True until we know the VirtualSize,
-				// which may happen one update after the image becomes ready.
-				harvesterv1beta1.ImageImported.True(toUpdate)
-				harvesterv1beta1.ImageImported.Reason(toUpdate, "Imported")
-				harvesterv1beta1.ImageImported.Message(toUpdate, status.Message)
-			}
 		} else if status.Progress != toUpdate.Status.Progress {
 			harvesterv1beta1.ImageImported.Unknown(toUpdate)
 			harvesterv1beta1.ImageImported.Reason(toUpdate, "Importing")
--- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -57,6 +57,37 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
 			return h.images.Update(toUpdate)
 		}
 
+		// sync virtualSize (handles the case for existing images that were
+		// imported before this field was added, because adding the field to
+		// the CRD triggers vmImageHandler.OnChanged)
+		if image.Status.VirtualSize == 0 {
+			toUpdate := image.DeepCopy()
+			bi, err := util.GetBackingImage(h.backingImageCache, toUpdate)
+			if err != nil {
+				// If for some reason we're unable to get the backing image,
+				// we set the BackingImageMissing condition on the image,
+				// including the error message, so it can be seen via
+				// `kubectl describe`.  The image is not marked failed or
+				// anything, in case this is some sort of transient error.
+				// In the unlikely event that we do hit this case, and whatever
+				// error was present is later fixed, the user can re-trigger
+				// this code by making a temporary change to the image (e.g.
+				// add/change the image description).  Once set, the
+				// BackingImageMissing condition is never automatically removed,
+				// but can be deleted manually with `kubectl edit`.
+				logrus.WithError(err).WithFields(logrus.Fields{
+					"namespace": toUpdate.Namespace,
+					"name":      toUpdate.Name,
+				}).Error("failed to get backing image for vmimage")
+				harvesterv1.BackingImageMissing.True(toUpdate)
+				harvesterv1.BackingImageMissing.Message(toUpdate,
+					fmt.Sprintf("Failed to get backing image: %s", err.Error()))
+			} else {
+				toUpdate.Status.VirtualSize = bi.Status.VirtualSize
+			}
+			return h.images.Update(toUpdate)
+		}
+
 		return image, nil
 	}
 
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -9547,6 +9547,10 @@
           },
           "storageClassName": {
             "type": "string"
+          },
+          "virtualSize": {
+            "type": "integer",
+            "format": "int64"
           }
         }
       },
--- a/deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml
+++ b/deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml
@@ -26,6 +26,9 @@ spec:
     - jsonPath: .status.size
       name: SIZE
       type: integer
+    - jsonPath: .status.virtualSize
+      name: VIRTUALSIZE
+      type: integer
     - jsonPath: .metadata.creationTimestamp
       name: AGE
       type: date
@@ -123,6 +126,9 @@ spec:
                 type: integer
               storageClassName:
                 type: string
+              virtualSize:
+                format: int64
+                type: integer
             type: object
         required:
         - spec
--- a/pkg/apis/harvesterhci.io/v1beta1/image.go
+++ b/pkg/apis/harvesterhci.io/v1beta1/image.go
@@ -10,6 +10,7 @@ var (
 	ImageInitialized        condition.Cond = "Initialized"
 	ImageImported           condition.Cond = "Imported"
 	ImageRetryLimitExceeded condition.Cond = "RetryLimitExceeded"
+	BackingImageMissing     condition.Cond = "BackingImageMissing"
 )
 
 // +genclient
--- a/pkg/apis/harvesterhci.io/v1beta1/image.go
+++ b/pkg/apis/harvesterhci.io/v1beta1/image.go
@@ -17,6 +17,7 @@ var (
 // +kubebuilder:resource:shortName=vmimage;vmimages,scope=Namespaced
 // +kubebuilder:printcolumn:name="DISPLAY-NAME",type=string,JSONPath=`.spec.displayName`
 // +kubebuilder:printcolumn:name="SIZE",type=integer,JSONPath=`.status.size`
+// +kubebuilder:printcolumn:name="VIRTUALSIZE",type=integer,JSONPath=`.status.virtualSize`
 // +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=`.metadata.creationTimestamp`
 
 type VirtualMachineImage struct {
@@ -79,6 +80,9 @@ type VirtualMachineImageStatus struct {
 
 	// +optional
 	Size int64 `json:"size,omitempty"`
+
+	// +optional
+	VirtualSize int64 `json:"virtualSize,omitempty"`
 
 	// +optional
 	StorageClassName string `json:"storageClassName,omitempty"`
--- a/pkg/apis/harvesterhci.io/v1beta1/openapi_generated.go
+++ b/pkg/apis/harvesterhci.io/v1beta1/openapi_generated.go
@@ -3967,6 +3967,12 @@ func schema_pkg_apis_harvesterhciio_v1beta1_VirtualMachineImageStatus(ref common
 							Format: "int64",
 						},
 					},
+					"virtualSize": {
+						SchemaProps: spec.SchemaProps{
+							Type:   []string{"integer"},
+							Format: "int64",
+						},
+					},
 					"storageClassName": {
 						SchemaProps: spec.SchemaProps{
 							Type:   []string{"string"},


tserong:
OK, the code here is (or should be) correct, but AFAICT we're not going to be able to get a proper build until we have a v1.7.0-whatever tag of longhorn with published images

tserong:
Rebased now that we've bumped to longhorn 1.6.2, so finally ready for review \o/

tserong:
Rebased on latest master

pkg/controller/master/image/backing_image_controller.go
```diff
@@ -46,11 +65,16 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
 			toUpdate = handleFail(toUpdate, condition.Cond(harvesterv1beta1.ImageImported), fmt.Errorf(status.Message))
 			toUpdate.Status.Progress = status.Progress
 		} else if status.State == lhv1beta2.BackingImageStateReady || status.State == lhv1beta2.BackingImageStateReadyForTransfer {
-			harvesterv1beta1.ImageImported.True(toUpdate)
-			harvesterv1beta1.ImageImported.Reason(toUpdate, "Imported")
-			harvesterv1beta1.ImageImported.Message(toUpdate, status.Message)
 			toUpdate.Status.Progress = status.Progress
 			toUpdate.Status.Size = backingImage.Status.Size
+			toUpdate.Status.VirtualSize = backingImage.Status.VirtualSize
```
m-ildefons, line 76: When upgrading an old version of Harvester, with existing images the image CRD will already be imported and this code will not be reached. Therefore the virtual image size will never be correctly displayed for existing image CRDs.

m-ildefons (changes requested):
I'm unsure about the behavior of this when upgrading from older Harvester versions. The existing (already initialized) images should have their virtual size updated as well. Otherwise images that are grandfathered in from before the upgrade of Harvester may result in VM disks with size 0 being created erroneously.

pkg/controller/master/image/backing_image_controller.go
tserong: The `if` on line 59 takes care of that. We only return up there if the image CRD is imported and the virtualSize is set. If the virtualSize isn't set (which it won't be for existing imported images), we land here.

m-ildefons: But the entire function will only be called when there is a change on the CRD, right? When an image is already imported, wouldn't the CRD always stay the same?

tserong: Dammit.

This `OnChanged()` function is called when the longhorn backing image CR changes, which it will, when a version of Longhorn which supports backing image virtualSize is installed/starts up. So for existing images, what _should_ happen (and what _did_ happen when I was originally testing this during development) is something like this:

- Longhorn's backing-image-manager determines virtualSize of existing backing image
- Longhorn's longhorn-manager, in `syncBackingImageFileInfo()` picks up the virtualSize from the backing image, and sets that in the lhbi CR.
- Harvester notices that change, propagates the change to the existing imported vmimage CR and everyone is happy.

Since then though, the virtualSize addition to longhorn has made it into LH version 1.5.5 (which is included in Harvester 1.2.2) and LH 1.6.2 (included in Harvester 1.3.1), which is why I started this comment with "dammit", because I fear you may be right -- if we're running Harvester 1.2.2 or 1.3.1, the lbhi CRs will have virtualSize set already so should have no reason to change when we update to a newer version of Harvester with this change included.

I'm gonna need to re-test this...

Thanks for the review!

tserong: OK, confirmed (dammit!)

Tested as follows:

- Install Harvester v1.3.0 (no virtualSize support anywhere).
- Add an image (https://download.opensuse.org/distribution/leap/15.5/appliances/openSUSE-Leap-15.5-Minimal-VM.x86_64-Cloud.qcow2)
- Check lhbi:
```
# kubectl get lhbi -A
NAMESPACE         NAME                  UUID       SOURCETYPE   SIZE        AGE
longhorn-system   default-image-kmvkf   006145f3   download     260465664   107s
```
- Upgrade to harvester v1.3.1 (LH supports virtualSize, Harvester does not yet)
- Check lhbi and vmimages:
```
# kubectl get lhbi -A
NAMESPACE         NAME                  UUID       SOURCETYPE   SIZE        VIRTUALSIZE   AGE
longhorn-system   default-image-kmvkf   006145f3   download     260465664   833617920     33m

# kubectl get vmimages
NAME          DISPLAY-NAME                                       SIZE        AGE
image-kmvkf   opensuse-leap-15.5-minimal-vm.x86_64-cloud.qcow2   260465664   58m
```
- Build a version of the v1.3.1 harvester container image with this PR applied, then run `kubectl -n harvester-system edit deployments/harvester` and replace `image: rancher/harvester:v1.3.1` with `image: tserong/harvester:wip-v1.3.1-virtualSize-head`. Also run `kubectl edit crd/virtualmachineimages.harvesterhci.io` to apply the CRD changes.
- Check vmimages and observe no virtualSize :-/ :
```
# kubectl get vmimages
NAME          DISPLAY-NAME                                       SIZE        VIRTUALSIZE   AGE
image-kmvkf   opensuse-leap-15.5-minimal-vm.x86_64-cloud.qcow2   260465664                 64m
```

If I subsequently add another image, that one of course works fine, but as you suspected the old one never gets updated:

```
# kubectl get lhbi -A
NAMESPACE         NAME                  UUID       SOURCETYPE   SIZE        VIRTUALSIZE   AGE
longhorn-system   default-image-4jjmc   7e8d55ee   download     21430272    117440512     46s
longhorn-system   default-image-kmvkf   006145f3   download     260465664   833617920     66m

# kubectl get vmimages
NAME          DISPLAY-NAME                                       SIZE        VIRTUALSIZE   AGE
image-4jjmc   cirros-0.6.2-x86_64-disk.img                       21430272    117440512     28s
image-kmvkf   opensuse-leap-15.5-minimal-vm.x86_64-cloud.qcow2   260465664                 66m
```

Guess I'll need to fix that...

tserong:
OK, 989087d picks up the virtualSize of existing images on upgrade in my testing (same manual test procedure as described above -- updating the vmimage CRD triggers `vmImageHandler.OnChanged`). The one thing I'm not certain about now is whether I still need the earlier attempt to handle the upgrade case with that adjusted `if !harvesterv1beta1.ImageImported.IsUnknown(vmImage)` in `backingImageHandler.OnChanged`. I doubt it hurts to keep it, I'm just not sure it's strictly necessary anymore, and if it's not necessary I'd rather not do it.

m-ildefons:
That's good.

I think the check still makes sense, because the Longhorn CRD 'abuses' the "Unknown" state of the "ImageImported" status condition to represent that an upload/download is in-progress. Therefore, if the "ImageImported" status condition is not in "Unknown" state, the backing image controller doesn't have to update the VM Image resource's status, since the upload/download has completed and there is no new progress information. This remains true, even if the backing image resource is updated for another reason (e.g. an upgrade).

The entire controller is a bit wonky in the sense that it watches a resource that it doesn't control (Longhorn Backing Images) and modifies its own resources based on what it sees. Usually for Kubernetes, resources should represent a _desired_ state and the controller should modify the actual state of the cluster to match that desired state as closely as possible.
There are cases, such as this, when the actual cluster state needs to be represented in the CRD. Often, this goes along with a controller that automatically re-queues its CRDs for reconciliation on a timer, so it can regularly update the CRD status if needed. In other cases, the controller interfaces with non-K8s APIs and _they_ might have triggers as well. But this isn't the case here.
Also, Kubernetes CRDs are versioned API objects for exactly the reason that if you need to make changes to the CRD, the version migration alone would provide a hook for dealing with this kind of stuff.

In any case, I think the way it is handled now is sufficient.
Great work on working out a reproducer for the update edge case, btw.

m-ildefons (approved):

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -57,6 +57,18 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
...
+		// sync virtualSize (handles the case for existing images that were
+		// imported before this field was added, because adding the field to
+		// the CRD triggers vmImageHandler.OnChanged)
+		if image.Status.VirtualSize == 0 {
+			bi, err := util.GetBackingImage(h.backingImageCache, image)
+			if err == nil {
+				toUpdate := image.DeepCopy()
+				toUpdate.Status.VirtualSize = bi.Status.VirtualSize
+				return h.images.Update(toUpdate)
+			}
```
WebberHuang1118: If we get an error from `util.GetBackingImage(h.backingImageCache, image)`, but do not return the error, it will make the image not be reconciled again, and be left with an empty `VirtualSize`

tserong: Ah, right. Maybe best I move this block into `processVMImage()` after that function's call to `util.GetBackingImage()` to handle that case, and avoid a bit of code duplication.

pkg/controller/master/image/backing_image_controller.go
```diff
@@ -46,11 +65,16 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
...
+			if toUpdate.Status.VirtualSize > 0 {
+				// We can't set ImageImported to True until we know the VirtualSize,
+				// which may happen one update after the image becomes ready.
+				harvesterv1beta1.ImageImported.True(toUpdate)
+				harvesterv1beta1.ImageImported.Reason(toUpdate, "Imported")
+				harvesterv1beta1.ImageImported.Message(toUpdate, status.Message)
+			}
```
WebberHuang1118: Just double confirm, does this mean bi could turn into a `BackingImageStateReady` state with `VirtualSize` still as zero?

That's not straightforward to me, can you help to point out the related LH source code snippet? Thanks.

pkg/controller/master/image/vm_image_controller.go
WebberHuang1118: `processVMImage()` is for images not imported yet, IIUC, this is for handling images that are already imported for the upgrade path?

pkg/controller/master/image/backing_image_controller.go
WebberHuang1118: I think is [this](https://github.com/longhorn/longhorn-manager/blob/77ac26d09a1bc18e743efa2e6bb94efc7d0fa4da/controller/backing_image_controller.go#L644-L700).

tserong: > I think is [this](https://github.com/longhorn/longhorn-manager/blob/77ac26d09a1bc18e743efa2e6bb94efc7d0fa4da/controller/backing_image_controller.go#L644-L700).

Yeah, that's it, from https://github.com/longhorn/longhorn-manager/pull/2680, which in turn relies on https://github.com/longhorn/backing-image-manager/pull/208

Unfortunately I don't have logs from when I first implemented/tested this, so I can't provide more detail right now on the comment in the code above. IIRC I observed `backingImageHandler.OnChanged()` being called twice, once when the backing image became ready(?), then again immediately after when the virtualSize was updated. I can re-test on Monday to confirm if that would help.

WebberHuang1118: No problem, thanks.

tserong: I added some logging and re-tested. This block is hit twice, because the `if` condition is `status.State == lhv1beta2.BackingImageStateReady || status.State == lhv1beta2.BackingImageStateReadyForTransfer`. The first time, we come in with `status.state == BackingImageStateReadyForTransfer`, and `virtualSize == 0`. The second time, it's `status.State == lhv1beta2.BackingImageStateReady` and `virtualSize` set to the actual size of the image. I'll update the comment...

pkg/controller/master/image/vm_image_controller.go
tserong: My reading is that `processVMImage()` is effectively a continuation of `OnChanged()`, and was split off the function as part of https://github.com/harvester/harvester/pull/5498. If I leave the virtualSize update in `OnChanged()`, and want to handle the error from `util.GetBackingImage()`, I have to duplicate the error handling code that's already in `processVmImage()`, for example:

```diff --git a/pkg/controller/master/image/vm_image_controller.go --- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -57,6 +57,26 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
                        return h.images.Update(toUpdate)
                }
 
+               // sync virtualSize (handles the case for existing images that were
+               // imported before this field was added, because adding the field to
+               // the CRD triggers vmImageHandler.OnChanged)
+               if image.Status.VirtualSize == 0 {
+                       bi, err := util.GetBackingImage(h.backingImageCache, image)
+                       if err != nil {
+                               if !errors.IsNotFound(err) {
+                                       logrus.WithError(err).WithFields(logrus.Fields{
+                                               "namespace": image.Namespace,
+                                               "name":      image.Name,
+                                       }).Error("failed to get backing image for vmimage")
+                                       return image, err
+                               }
+                               return h.handleRetry(image)
+                       }
+                       toUpdate := image.DeepCopy()
+                       toUpdate.Status.VirtualSize = bi.Status.VirtualSize
+                       return h.images.Update(toUpdate)
+               }
+
                return image, nil
        }
 ```

Whereas if I move the virtualSize update to `processVMImage()`, I should only have to do this:

```diff --git a/pkg/controller/master/image/vm_image_controller.go b/pkg/controller/master/image/vm_image_controller.go
index 5f832ccf..06ec076e 100644
--- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -88,6 +88,17 @@ func (h *vmImageHandler) processVMImage(image *harvesterv1.VirtualMachineImage)
                return h.handleRetry(image)
        }
 
+       if harvesterv1.ImageImported.IsTrue(image) {
+               // sync virtualSize (handles the case for existing images that were
+               // imported before this field was added, because adding the field to
+               // the CRD triggers vmImageHandler.OnChanged)
+               if image.Status.VirtualSize == 0 {
+                       toUpdate := image.DeepCopy()
+                       toUpdate.Status.VirtualSize = bi.Status.VirtualSize
+                       return h.images.Update(toUpdate)
+               }
+       }
+
        if bi.DeletionTimestamp != nil {
                h.imageController.EnqueueAfter(image.Namespace, image.Name, checkInterval)
                return image, nil
```

pkg/controller/master/image/backing_image_controller.go
WebberHuang1118: so if we find `status.State == lhv1beta2.BackingImageStateReady`, can we confirm `backingImage.Status.VirtualSize` must not be empty?

if yes, maybe we could just skip `status.State == lhv1beta2.BackingImageStateReadyForTransfer` ?

pkg/controller/master/image/vm_image_controller.go
WebberHuang1118:
```
func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
       ....

	if harvesterv1.ImageImported.IsTrue(image) {
		....
		return image, nil
	}

	return h.processVMImage(image)
}
```

If the image with `harvesterv1.ImageImported.IsTrue(image) `, it's not going to enter `processVMImage()`. Or do I miss anything?

tserong: My mistake, you are absolutely correct **>facepalm<**

To avoid duplicating the `util.GetBackingImage()` error handling, I'd need to move that entire `if harvesterv1.ImageImported.IsTrue(image)` block into `processVMImage`.

WDYT makes the most sense here? Duplicate the error handing, or move that whole block?

WebberHuang1118: > To avoid duplicating the util.GetBackingImage() error handling, I'd need to move that entire if harvesterv1.ImageImported.IsTrue(image) block into processVMImage.

I'm wondering this will introduce too much modification. Actually in the current implementation, we do not update `image.Status.VirtualSize` in `processVMImage()` (we update `image.Status.VirtualSize` in `func (h *backingImageHandler) OnChanged()`), right?

IMHO, maybe we could apply the approach you mentioned before

```diff --git a/pkg/controller/master/image/vm_image_controller.go --- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -57,6 +57,26 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
                        return h.images.Update(toUpdate)
                }
 
+               // sync virtualSize (handles the case for existing images that were
+               // imported before this field was added, because adding the field to
+               // the CRD triggers vmImageHandler.OnChanged)
+               if image.Status.VirtualSize == 0 {
+                       bi, err := util.GetBackingImage(h.backingImageCache, image)
+                       if err != nil {
+                               if !errors.IsNotFound(err) {
+                                       logrus.WithError(err).WithFields(logrus.Fields{
+                                               "namespace": image.Namespace,
+                                               "name":      image.Name,
+                                       }).Error("failed to get backing image for vmimage")
+                                       return image, err
+                               }
+                               return h.handleRetry(image)
+                       }
+                       toUpdate := image.DeepCopy()
+                       toUpdate.Status.VirtualSize = bi.Status.VirtualSize
+                       return h.images.Update(toUpdate)
+               }
+
                return image, nil
        }
 ```
 
 Or we can have the feedback from @bk201 or @FrankYang0529 ? Thanks.

File: pkg/controller/master/image/backing_image_controller.go (hunk 3 above)
Comment (tserong) [reply pkg/controller/master/image/backing_image_controller.go]: Yeah, good idea, that will make it a bit simpler.

File: pkg/controller/master/image/vm_image_controller.go (hunk 2 above)
Comment (tserong) [reply pkg/controller/master/image/vm_image_controller.go]: > I'm wondering this will introduce too much modification. Actually in the current implementation, we do not update `image.Status.VirtualSize` in `processVMImage()` (we update `image.Status.VirtualSize` in `func (h *backingImageHandler) OnChanged()`), right?

Correct. I was only thinking to move everything to `processVMImage()` to avoid duplicating that error handling code.

(My original reason for not handling the error case was just that I thought it extremely unlikely for `util.GetBackingImage` to fail for an already imported image)

Comment (tserong) [reply pkg/controller/master/image/vm_image_controller.go]: @bk201 @FrankYang0529 WDYT of the above?

TL;DR: we've discussed three possibilities for this little chunk of code:

1. Ignore error return from `util.GetBackingImage()` when updating virtualsize of existing imported images.
2. Handle possible error return by duplicating error handling code from `processVmImage()`, inside `OnChanged()`.
3. Move all imported image handling code from `OnChanged()` to `processVmImage()` to avoid duplicate error handling code.

Comment (bk201) [reply pkg/controller/master/image/vm_image_controller.go]: My 2 cents is option 2, it's more straightforward.
Swallowing errors is not a good idea. If an error happens here, it probably means there is a bug somewhere because a VirtualMachineimage can't get a corresponding BackingImage.

Comment (bk201) [reply pkg/controller/master/image/vm_image_controller.go]: BTW, is an existing backing image guaranteed to have a positive `bi.Status.VirtualSize` value after Longhorn is upgraded?

Comment (tserong) [reply pkg/controller/master/image/vm_image_controller.go]: > My 2 cents is option 2, it's more straightforward.

Alright, 2 it is. Thanks @bk201, will do, then a quick re-test for good measure, then squash and we're finally done :-)

> BTW, is an existing backing image guaranteed to have a positive `bi.Status.VirtualSize` value after Longhorn is upgraded?

Yep (see my manual test notes in https://github.com/harvester/harvester/pull/5387#discussion_r1654424863)

**Comment by tserong** [comment]
OK, that should be everything :-) do you mind having another look @WebberHuang1118 @bk201?

**Review by bk201 — APPROVED** [review]
LGTM, thanks.

File: pkg/controller/master/image/vm_image_controller.go [hunk 4]
```diff
@@ -57,6 +57,26 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
...
+						"name":      image.Name,
+					}).Error("failed to get backing image for vmimage")
+					return image, err
+				}
+				return h.handleRetry(image)
+			}
+			toUpdate := image.DeepCopy()
+			toUpdate.Status.VirtualSize = bi.Status.VirtualSize
+			return h.images.Update(toUpdate)
+		}
```
WebberHuang1118: IMHO, for an imported image, we should not make it go through `hanldeRetry()`. It will delete related bi and sc, and recreate all resources again. For the case bi is missing, maybe we can just log the error, and update the image condition. WDYT?

WebberHuang1118: Just a nit, I prefer the `nest-if` with lower depth (do not have functional change)
```
		if image.Status.VirtualSize == 0 {
			bi, err := util.GetBackingImage(h.backingImageCache, image)
			if err == nil {
				toUpdate := image.DeepCopy()
				toUpdate.Status.VirtualSize = bi.Status.VirtualSize
				return h.images.Update(toUpdate)
			}

			if !errors.IsNotFound(err) {
					"namespace": image.Namespace,
					"name":      image.Name,
				}).Error("failed to get backing image for vmimage")
				return image, err
			}

			return h.handleRetry(image)
		}
```

tserong: Oh, good catch. Yeah, I agree, given the backing image and storage class really _should_ be there already for an imported image, deleting those things in this case seems like a bad idea.

I definitely think we should log the error, but what do you think the condition should be updated to?

WebberHuang1118: https://github.com/harvester/harvester/blob/9c0a3a17f42831b590a92f3743ac027a66a25b8d/pkg/apis/harvesterhci.io/v1beta1/image.go#L9-L13

just my 2 cents, maybe a new condition like `BackingImageMiss`?

tserong: Alright, will see what I can do tomorrow!

tserong: code structure has changed with latest update

tserong: OK, the latest change will add a BackingImageMissing condition to the vmimage on error here, with a message set to the error, which means you can see this if something goes wrong:

```
# kubectl describe vmimages/image-kqcnt
[...]
Status:
  Conditions:
    Last Update Time:  2024-07-04T02:08:48Z
    Reason:            Imported
    Status:            True
    Type:              Imported
    Last Update Time:  2024-07-04T02:08:32Z
    Reason:            Initialized
    Status:            True
    Type:              Initialized
    Last Update Time:  2024-07-04T02:35:07Z
    Message:           Failed to get backing image: backingimages.longhorn.io "default-image-kqcnt" not found
    Status:            True
    Type:              BackingImageMissing
  Failed:              0
  Progress:            100
  Size:                260465664
  Storage Class Name:  longhorn-image-kqcnt
  Virtual Size:        833617920
Events:                <none>
```

As I mentioned in the comment in the code, the image isn't actually marked failed or anything in case this error is transient. I really think it's extremely unlikely to hit this, so I haven't added any code to automatically clear that condition if the image later comes good somehow (I don't know how such an event would be triggered anyway).

I tested this by:
- deleting the backing image
- upgrading the harvester container
- verifying that the error was logged and that the condition was set on the image

I then repaired the backing image by:
- using the longhorn GUI to upload a backing image with the same name as the missing image
- edited the image to tweak its description (which triggers the OnChange code, which in turn triggers the virtual size change, which worked this time because the backing image was present)
- manually deleted the BackingImageMissing condition with `kubectl edit`

WebberHuang1118 (approved):
LGTM, thanks for your effort @tserong .

tserong:
That last force push was just to squash a commit. Will merge once the build completes. Thanks all!

Issue #5383, opened 2024-03-18
[ENHANCEMENT] Refactor `harvesterhci.io/owned-by` annotation on PVC


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**

The `harvesterhci.io/owned-by` annotation is for checking PVC used by which VM. If there is resizing/deleting/attaching PVC request, the system uses it to check whether PVC is used on a running VM.

However, the annotation may be updated by users. In this case, the system can't do correct determination for the PVC.

**Describe the solution you'd like**

Introduce a new index `VMByPVCIndex`, this indexer can search PVC correspond to which VM by PVC name. It doesn't rely on annotation, so users don't have chance to update it.

**Additional context**

https://github.com/harvester/harvester/pull/5031

TachunLin:
Verified **PASS** on `master-49d172b-head` (24/07/28). Close this issue.

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ Case 1: a volume on a running VM can't be resized $~~$</summary>

1. When we try to resize the volume already attached to vm, the webhook would prevent and prompt message

  ```
  admission webhook "validator.harvesterhci.io" denied the request: resizing is only supported for detached volumes. The volume is being used by VM default/vm1. Please stop the VM first.
  ```

  ![image](https://github.com/user-attachments/assets/1aee818f-9013-4a45-9994-f73e5f894252)


 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Case 2: a volume on a stopped VM can be resized $~~$</summary>

1. Can resize volume when vm is off
  
    ![image](https://github.com/user-attachments/assets/221aa1d6-4909-4639-87e3-cb24f8b7f2cc)
    
    ![image](https://github.com/user-attachments/assets/a9aeff49-6c09-4317-8f38-743102227a2c)
    
    
    
    https://github.com/user-attachments/assets/0c777e12-5308-48c2-8af1-cae85e9bebf7




2. Can start the vm in running after volume resized


  https://github.com/user-attachments/assets/809f7dcc-0a8d-4d3e-9c39-eaa60b200b00



 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Case 3: a volume on a VM can't be deleted $~~$</summary>

1. Can't delete the volume already attached to a VM.

    ![image](https://github.com/user-attachments/assets/4204cabf-8635-457a-8e9d-795b752047f1)


 </details>

[...]

PR #5406, opened 2024-03-20
feat: remove harvesterhci.io/owned-by

FrankYang0529:
**Problem:**

The `harvesterhci.io/owned-by` annotation is for checking PVC used by which VM. If there is resizing/deleting/attaching PVC request, the system uses it to check whether PVC is used on a running VM.

However, the annotation may be updated by users. In this case, the system can't do correct determination for the PVC.

**Solution:**

Introduce a new index `VMByPVCIndex`, this indexer can search PVC correspond to which VM by PVC name. It doesn't rely on annotation, so users don't have chance to update it.

**Related Issue:**
https://github.com/harvester/harvester/issues/5383

**Additional context**

It looks like `Attached VM` on Volumes page is from `harvesterhci.io/owned-by` annotation. We need another UI PR to fix this.

Screenshot 2024-03-20 at 11 04 44 AM

**Test plan:**

### Case 1: a volume on a running VM can't be resized

1. Create a VM and make sure VM is running.
2. Resize the VM volume and webhook will deny the request.
Screenshot 2024-03-20 at 10 52 55 AM

### Case 2: a volume on a stopped VM can be resized
1. Follow Case1.
2. Stop the VM.
3. Resize the VM volume without any error.
4. Start the VM without any error.

### Case 3: a volume on a VM can't be deleted
1. Follow Case2.
2. Delete the volume and webhook will deny the request.
Screenshot 2024-03-20 at 11 01 47 AM

### Case 4: a volume without VM can be deleted
1. Follow Case 3.
2. Delete the VM but not the volume.
3. After VM is removed, delete the volume without any error.

### Case 5: delete a VM with volume can work
1. Create a new VM.
2. Delete the VM with volume.
3. Both VM and volume are removed.

### Case 6: A volume can't be attached to another VM

1. Create 2 VMs.
2. Create a data volume and attach to one VM.
3. Try to attach the data volume to another VM and webhook will deny the request.
Screenshot 2024-03-20 at 3 56 55 PM

### Case 7: Old VM can work with new controller.

1. Create a v1.3.0 harvester.
2. Create a VM. Check VM has `harvesterhci.io/VMController.UnsetOwnerOfPVCs` finalizer and VMI has `wrangler.cattle.io/VMIController.UnsetOwnerOfPVCs`.
3. Change the harvester and harvester-webhook image to this branch.
4. Delete the VM with volume.
5. Both resources will be removed.


pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -240,61 +164,38 @@ func (h *VMController) StoreRunStrategy(_ string, vm *kubevirtv1.VirtualMachine)
...
-			return fmt.Errorf("failed to get PVC(%s/%s): %w", pvcNamespace, pvcName, err)
-		}
-
-		if err := unsetBoundedPVCReference(h.pvcClient, pvc, vm); err != nil {
-			return fmt.Errorf("failed to revoke VitrualMachine(%s/%s) as PVC(%s/%s)'s owner: %w",
-				vm.Namespace, vm.Name, pvcNamespace, pvcName, err)
+// removePVCInRemovedPVCsAnnotation cleanup PVCs which users want to remove with the VM.
+func (h *VMController) removePVCInRemovedPVCsAnnotation(vm *kubevirtv1.VirtualMachine) error {
+	for _, pvcName := range getRemovedPVCs(vm) {
+		pvcName = strings.TrimSpace(pvcName)
```
connorkuehl: did getRemovedPVCs return strings with spaces in them? It might be better to call strings.TrimSpace on all entries in getRemovedPVCs before they're returned so that the caller doesn't have to remember to do this.

pkg/util/indexeres/pod.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+const (
+	PodByVMNameIndex = "harvesterhci.io/pod-by-vmname"
+)
+
+func PodByVMName(obj *corev1.Pod) ([]string, error) {
+	vmName, ok := obj.Labels[util.LabelVMName]
+	if !ok {
+		return []string{}, nil
```
connorkuehl: Feel free to ignore suggestion, but I prefer nil as it doesn't heap allocate.

```suggestion
		return nil, nil
```

pkg/util/indexeres/virtualmachine.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/harvester/harvester/pkg/ref"
+)
+
+// The file contains the indexers which are used by controller and webhook.
+const (
+	VMByPVCIndex = "harvesterhci.io/vm-by-pvc"
+)
+
+func VMByPVC(obj *kubevirtv1.VirtualMachine) ([]string, error) {
+	results := []string{}
```
connorkuehl: Feel free to ignore suggestion, but I prefer nil as it doesn't heap allocate.

```suggestion
	var results []string
```

connorkuehl (approved):
None of the comments are blockers! :+1:

connorkuehl:
> It looks like Attached VM on Volumes page is from harvesterhci.io/owned-by annotation. We need another UI PR to fix this.

BTW, is there a corresponding frontend ticket for this? I looked, but it's possible I just missed it. CC @torchiaf

pkg/controller/master/virtualmachine/vm_controller.go
FrankYang0529: Moved `TrimSpace` to `getRemovedPVCs` function.

We have to avoid error data from users. I will use https://github.com/harvester/harvester/issues/5456 to validate annotation value in webhook.

FrankYang0529:
> > It looks like Attached VM on Volumes page is from harvesterhci.io/owned-by annotation. We need another UI PR to fix this.
> 
> BTW, is there a corresponding frontend ticket for this? I looked, but it's possible I just missed it. CC @torchiaf

I add `require-ui/small` tag to the issue first. Feel free to create another issue to track this. Thank you.

pkg/util/indexeres/virtualmachine.go
```diff
@@ -0,0 +1,26 @@
...
+
+func VMByPVC(obj *kubevirtv1.VirtualMachine) ([]string, error) {
+	var results []string
+	if obj == nil || obj.Spec.Template == nil {
+		return results, nil
+	}
+
+	for _, volume := range obj.Spec.Template.Spec.Volumes {
+		if volume.PersistentVolumeClaim != nil && volume.PersistentVolumeClaim.ClaimName != "" {
+			results = append(results, ref.Construct(obj.Namespace, volume.PersistentVolumeClaim.ClaimName))
```
brandboat, line 22: Leave a noob question here, looks like we assume that the namespace of PVCs under the VM are all the same as VM.

I'm wondering if there is a scenario that
- namesapce_a contains volume_1 and vm_a
- namesapce_b contains volume_1
Can we bind volume_1 (under namespace_b) to vm_a (namespace_a) ? Is this legal ?

FrankYang0529: Yeah, from the source code, they are same namespaces.

https://github.com/kubevirt/kubevirt/blob/21b0183ea480f2084efd286b73dd5b0931c4a714/staging/src/kubevirt.io/api/core/v1/schema.go#L788-L792

brandboat: Thanks for the explanation ! :smiley:

brandboat (approved):
LGTM ! I've tested case 1~6 and all good :smiley:

WebberHuang1118 (approved):
Testing cases work as expected, LGTM, thanks.

bk201:
The e2e test case [`test_delete_volume_on_existing_vm`](https://github.com/harvester/tests/blob/main/harvester_e2e_tests/integrations/test_1_volumes.py#L140) fails because it uses the deprecated annotation to find a volume's attached VM. 
<details>
<summary>Expand to see the error message</summary>

```
self = <harvester_e2e_tests.integrations.test_1_volumes.TestVolumeWithVM object at 0x7faa7bf3eb80>
api_client = HarvesterAPI('https://192.168.0.131', 'Bearer token-68thc:xk44xbjzxjb2bt2m67zdd9kd72pg457mzpps8dhlflzjm6587htdgr')
ubuntu_image = {'display_name': 'img-07h53m05s183224-07-04', 'id': 'default/img-07h53m05s183224-07-04', 'ssh_user': 'ubuntu'}
ubuntu_vm = {'actions': {'addVolume': 'https://192.168.0.131/v1/harvester/kubevirt.io.virtualmachines/default/vm-07h53m05s183224-0...mplate', ...}, 'apiVersion': 'kubevirt.io/v1', 'id': 'default/vm-07h53m05s183224-07-04', 'kind': 'VirtualMachine', ...}
polling_for = <function polling_for.<locals>._polling_for at 0x7faa7bd28280>

    def test_delete_volume_on_existing_vm(self, api_client, ubuntu_image, ubuntu_vm, polling_for):
        """
        1. Create a VM with volume
        2. Delete volume should reply 422
        3. Pause VM
        4. Delete volume should reply 422 too
        5. Stop VM
        6. Delete volume should reply 422 too
        Ref. https://github.com/harvester/tests/issues/905
        """
        vol_name = (ubuntu_vm["spec"]["template"]["spec"]["volumes"][0]
                             ['persistentVolumeClaim']['claimName'])
    
        code, data = api_client.volumes.delete(vol_name)
        assert 422 == code, f"Should fail to delete volume\n{code}, {data}"
    
        self.pause_vm(api_client, ubuntu_vm, polling_for)
        code, data = api_client.volumes.delete(vol_name)
        assert 422 == code, f"Should fail to delete volume\n{code}, {data}"
    
        self.stop_vm(api_client, ubuntu_vm, polling_for)
        code, data = api_client.volumes.delete(vol_name)
        assert 422 == code, f"Should fail to delete volume\n{code}, {data}"
    
        # Check Volume
        code, data = api_client.volumes.get(vol_name)
        mdata, annotations = data['metadata'], data['metadata']['annotations']
        assert 200 == code, (code, data)
        assert mdata['name'] == vol_name, (code, data)
        # status
        assert not mdata['state']['error'], (code, data)
        assert not mdata['state']['transitioning'], (code, data)
        assert data['status']['phase'] == "Bound", (code, data)
        # source
        assert ubuntu_image["id"] == annotations['harvesterhci.io/imageId'], (code, data)
        # attachment
>       assert ubuntu_vm['id'] in annotations.get("harvesterhci.io/owned-by"), data
E       TypeError: argument of type 'NoneType' is not iterable

harvester_e2e_tests/integrations/test_1_volumes.py:176: TypeError
```


</details>

I don't feel removing an annotation breaks our API, but we must give a heads-up to QA.

The GUI uses the deprecated annotation to display a volume's corresponding VM on the volume page ([GUI ticket ](https://github.com/harvester/harvester/issues/5905)). Is there a fast way to find the corresponding VM of a volume? Besides building the relationship from all VMs.

FrankYang0529:
> The GUI uses the deprecated annotation to display a volume's corresponding VM on the volume page (https://github.com/harvester/harvester/issues/5905). Is there a fast way to find the corresponding VM of a volume? Besides building the relationship from all VMs.

Another way to find attached VM from PVC is that:

1. PVC.spec.volumeName -> LH Volume
2. Volume.status.kubernetesStatus.workloadsStatus.workloadName -> VM

However, the `workloadsStatus` is a list, so we have to filter `podStatus: Running` in it.

bk201:
> The e2e test case [`test_delete_volume_on_existing_vm`](https://github.com/harvester/tests/blob/main/harvester_e2e_tests/integrations/test_1_volumes.py#L140) fails because it uses the deprecated annotation to find a volume's attached VM.
> 
> Expand to see the error message
> I don't feel removing an annotation breaks our API, but we must give a heads-up to QA.
> 
> The GUI uses the deprecated annotation to display a volume's corresponding VM on the volume page ([GUI ticket ](https://github.com/harvester/harvester/issues/5905)). Is there a fast way to find the corresponding VM of a volume? Besides building the relationship from all VMs.

@lanfon72 @albinsun can you help check this?

torchiaf:
The e2e test case [test_delete_volume_on_existing_vm](https://github.com/harvester/tests/blob/main/harvester_e2e_tests/integrations/test_1_volumes.py#L140) should work after merging the UI part: https://github.com/harvester/dashboard/pull/1065

FrankYang0529:
Rebase master branch. I can run dashboard without error. Thank you.

bk201 (approved):

FrankYang0529:
@mergify backport v1.4

Issue #1710, opened 2021-12-20
[FEATURE] USB passthrough support


msnelling:
Will there be the option to pass through USB devices plugged into a Harvester node to a running VM? Example use-cases could be
* Security dongles
* Zigbee tranceivers
* HID
* External storage

Yabbo:
would like to add the ability to boot from a local USB drive (users desktop not the server) when installing VMs into the harvester cluster.

msnelling:
> would like to add the ability to boot from a local USB drive (users desktop not the server) when installing VMs into the harvester cluster.

This isn't really what I meant. I'm talking about passing through a USB device attached to the server running Harvester, similar to GPU/PCIe passthrough. Currently I'm using this functionality in ESXi

FrankYang0529:
@msnelling Not sure whether this can solve your problem? Thank you.

https://github.com/kubevirt/kubevirt/pull/4089

soderstrom-rikard:
> @msnelling Not sure whether this can solve your problem? Thank you.
> 
> [kubevirt/kubevirt#4089](https://github.com/kubevirt/kubevirt/pull/4089)

This looks very interesting and I assume that any client that wants to forward this will need to have the usbredirect package installed?

Sorry if I'm hijacking this issue slightly but I really want to highlight that this would help creating a cloud-native solution for developers in the embedded industry. So I'm very interested in trying this out with multiple local clients connecting external harvester nodes.

tlehman:
If you have a USB Controller that is a PCI device, you can use PCI passthrough to get direct access to the USB device in a VM. Is there some usecase for USB passthrough that doesn't involve PCI?

If so, I can look into making a separate feature for this.

tlehman:
@msnelling does PCI passthrough of a USB controller work?

msnelling:
@tlehman in my particular use case I don't believe so. In the case of a single USB controller, this would limit USB to a single VM correct?

tlehman:
That's right, each controller would be limited to a single VM. I'll look into USB passthrough without getting the PCI bus involved

frodriguezp:
Is possible redirect the usb from Web Console? In our case we have Jubikey security devices to login in windows servers in Vmware, so the usb redirect must be work from Web VNC Console.

hunghvu:
[...]

PR #5430, opened 2024-03-21
HEP: add usb host passthrough [CI SKIP]

Yu-Jack:

**Problem:**
This is new feature, please check this PR for more detail.

**Solution:**

- [ ] Bump kubevirt of pcidevices to v1.1.1 cause USB Passthrough need that.
- [ ] Expose KubeVirt USB Passthrough config - enable USB Passthrough feature.
- [ ] Create external device plugin - enable ability to choose specified USB device.

**Related Issue:**
https://github.com/harvester/harvester/issues/1710

**Test plan:**


Yu-Jack:
Just discussed with @ibrokethecloud  about this feature should be implemented on separated repository or kept on same repository. We both prefer same repository, and will rename the repository name properly.

enhancements/20240304-usb-passthrough.md
```diff
@@ -0,0 +1,163 @@
...
+### Related Issues
+
+https://github.com/harvester/harvester/issues/1710
+
+## Motivation
+
+### Goals
+
+- Allow virtual machine to use node's USB device without PCI Passthrough.
+- Users can select one USB device from two identical USB device instead of picked by KubeVirt.
```
bk201: Can we merge the second goal with the paragraph below: 
> the second one aim to fix random picked from two identical USB device.

enhancements/20240304-usb-passthrough.md
```diff
@@ -0,0 +1,163 @@
...
+
+When Bob sets up the A virtual machine, Bobs could select U2 from dropdown menu in the `USB Devices` tab. Here, Bob will see U1/U2 have a lot of same information, but only name is different. Then Bob does same things when creating B virtual machine.
+
+After creating all machine, Bob can successfully use U2 from A VM, and U1 from B VM.
+
+### User Experience In Detail
+
+We don't mention how many nodes and which node has dedicated USB device. It should be automatically scheduled to suitable node due to good user experience. Let's say Bob selects the U1 which attached on node N1. When bob select the U1 in creating page, the node's scheduling of vm should be set up to node N1 automatically.
+
+On the other hand, If Bob already set up the node scheduling to node N2 which attached U2, it would notify Bob it's not allowed when Bob select U1 from dropdown menu because UI belongs to node N1.
```
bk201: Is it possible we can also prohibit this from webhook/backend?

Yu-Jack: Yeah, after some investigation and PoC, it's possible.

ibrokethecloud: we should probably just delete the pcidevice crd when a usbclaim involves the pci address. This should ensure that parent pci device cannot be passed through to any other workload vm.

ibrokethecloud (approved):
lgtm. thanks.

enhancements/20240304-usb-passthrough.md
Yu-Jack: I think we could keep it, but don't allow it to pass through and show some error messages like:

image

Otherwise, I'm afraid of that users might misunderstand we don't detect all pcidevices.

bk201 (approved):
LGTM!

Issue #5433, opened 2024-03-21
[BUG] Harvester is flooded with Terminating VMs if namespace is out of storage quota


votdev:
**Describe the bug**
Harvester is flooded with Terminating VMs if the namespace is out of storage quota.

Harvester tries creating VM, fails, creates one another, fails again, and so on. The VMs are stuck in Terminating state, and the Rancher GUI becomes very slow quickly after hundreds of these VMs.

**To Reproduce**

- Create a cluster in a namespace with enough quota
- Adjust the namespace storage quota to have less available storage than the root disk of a new VM. CPU and memory quota must be plenty, enough to be extended with a VM.
- Scale up the pool

**Expected behavior**
To be clarified

**Environment**
- Harvester 1.2.1
- Underlying Infrastructure: Baremetal

TachunLin:
Verifying **failed** on `v1.4-eaf3009a-head` (24/11/05) and `v1.4.0-rc5` on Rancher v2.8.5 and v2.8.9. 
Move the issue to implement for further investigation. 

### Test Result on Rancher v2.9.2
<details><summary>$\color{green}{\textsf{PASS}}$ Test namespace resource quota limitation on virtual machine creation $~~$</summary>

1. When we create the second vm to exceed the namespace quota, it would prompt error message to prevent to proceed

  ```
  admission webhook "validator.harvesterhci.io" denied the request: storage insufficient resources due to resource quota
  ```
  
  ![image](https://github.com/user-attachments/assets/4d32b17e-b522-44f8-a353-3e448f9f88ee)


 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Test namespace resource quota limitation when clone virtual machine and change volume size $~~$</summary>

1. When we edit the volume size of the cloned vm to exceed the namespace quota, it would prompt error message to prevent to proceed 

  ![image](https://github.com/user-attachments/assets/001bb4c8-f510-4719-b681-ad4b10f716a0)

2. When we change the volume size to not exceed the namespace quota, it could change correctly

  ![image](https://github.com/user-attachments/assets/9a3946c7-98d6-42cb-81d4-b4e103aef955)


 </details>


<details><summary>$\color{green}{\textsf{PASS}}$ Test namespace resource quota limitation when clone virtual machine with clone volume size $~~$</summary>

1. When we clone a vm with volume to the new one but exceed the namespace quota, it would prompt error message to prevent to proceed

[...]

PR #5037, opened 2024-01-26
feat: patch service loadbalancers in resource quota

FrankYang0529:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**
Currently if there is a DHCP on the network, handing out load balancer IPs cannot be restricted in the guest kubernetes clusters, so users can use up too many IPs on the cluster with the DHCP solution.

**Solution:**
Specifying the corresponding values in the ResourceQuota object for the namespace in which users will create LB-type services or the Project containing the Kubernetes guest cluster.

**Related Issue:**
https://github.com/harvester/harvester/issues/5685

**Test plan:**

### Case 1: Default `patch-harvester-services-in-resource-quota` setting is `false`
1. Setup harvester and rancher cluster.
2. Import harvester to rancher.
3. Set Service Load Balancers and other resource types in the Default project.
Screenshot 2024-05-03 at 1 33 33 PM
4. Check resource quota in default namespace is "services.loadbalancers".

### Case 2: `patch-harvester-services-in-resource-quota` setting is `true`
1. Following case1.
2. Change the `patch-harvester-services-in-resource-quota` setting value to `true`.
3. Check resource quota in default namespace is not "services.loadbalancers". It's `count/loadbalancers.loadbalancer.harvesterhci.io`.

```
> kubectl get resourcequota -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: ResourceQuota
  metadata:
    creationTimestamp: "2024-05-03T03:40:43Z"
    generateName: default-
    labels:
      cattle.io/creator: norman
      resourcequota.management.cattle.io/default-resource-quota: "true"
    name: default-sk7dg
    namespace: default
    resourceVersion: "192431"
    uid: b2c995bb-6cd8-4976-bebe-91a904578451
  spec:
    hard:
      configmaps: 1k
      count/loadbalancers.loadbalancer.harvesterhci.io: 1k
      pods: 1k
  status:
    hard:
      configmaps: 1k
      count/loadbalancers.loadbalancer.harvesterhci.io: 1k
      pods: 1k
    used:
      configmaps: "1"
      count/loadbalancers.loadbalancer.harvesterhci.io: "0"
      pods: "0"
kind: List
metadata:
  resourceVersion: ""
```
4. Set Service Load Balancers and other resource types in another project. We can see the resource quota is `count/loadbalancers.loadbalancer.harvesterhci.io` in harvester.

### Case 3: change `patch-harvester-services-in-resource-quota` setting to `true`
[...]

PR #5434, opened 2024-03-21
Harvester is flooded with Terminating VMs if namespace is out of storage quota

votdev:
Improve the resource quota handling when creating VMs.

Check resource quotas when the VM is created via webhook. Prevent the VM from being created if the storage resource quotas are exceeded.

[Bildschirmaufzeichnung vom 2024-05-14, 16-20-08.webm](https://github.com/harvester/harvester/assets/1897962/1052944c-dfa8-4b60-93a8-74ef95c7d462)

[Bildschirmaufzeichnung vom 2024-05-14, 16-36-10.webm](https://github.com/harvester/harvester/assets/1897962/d9747a19-3b99-4798-8a19-5df21e9666cd)

**Related Issue:**
https://github.com/harvester/harvester/issues/5433

**Test plan:**

## Case 1:
1. Go to the UI of your Rancher standalone installation (NOT the embedded one in Harvester).
2. Make sure your cluster is imported in the `Virtualization Management`.
3. Click the Harvester cluster on the `Name` field of the datatable.
4. Go to `Projects/Namespaces` and create a new project by clicking `Create Project`.
5. In the `Resource Quotas` tab, add `Storage Reservation` and use `20000 MiB` as project limit and `12000 MiB` as namespace default limit.
6. Create a new namespace named `quota-test` in the previously created project. Take over the suggested resource quota of `12 GiB`.
7. In the Harvester UI, create a new VM with the namespace `quota-test`. In `Volumes`, make sure the volume occupies `10 GiB`.
8. Press `Create`. The VM should be created and started. All should be fine.
9. Create a second VM with the same parameters and press `Create`.
10. This time the VM shouldn't be created and a warning message should be displayed at the top of the form.
grafik

## Case 2:
1. Go to the UI of your Rancher standalone installation (NOT the embedded one in Harvester).
2. Make sure your cluster is imported in the `Virtualization Management`.
3. Click the Harvester cluster on the `Name` field of the datatable.
4. Go to `Projects/Namespaces` and create a new project by clicking `Create Project`.
5. In the `Resource Quotas` tab, add `Storage Reservation` and use `20000 MiB` as project limit and `12000 MiB` as namespace default limit.
6. Create a new namespace named `quota-test` in the previously created project. Take over the suggested resource quota of `12 GiB`.
7. In the Harvester UI, create a new VM named `test01` with the namespace `quota-test`. In `Volumes`, make sure the volume occupies `5 GiB`.
8. Press `Create`. The VM should be created and started. All should be fine.
9. Create a second VM named `test02` by cloning `test01`. Check the `clone volume data` checkbox. Press `Create`. The VM should be created and started. All should be fine.
10. Edit VM `test02` by choosing `Edit Config` in the action menu.
11. Go to the `Volumes` tab and change the `Size` field from `5 GiB` to `7 GiB`. Press `Save` An admission webhook error should be displayed at the top saying `admission webhook "validator.harvesterhci.io" denied the request: storage insufficient resources due to resource quota`. This is because the storage quota of `12 GiB` has been exceeded.
12. Change the `Size` to `6 GiB` and press `Save`. An error message `please stop the VM before resizing volumes` should be displayed at the top.
13. Press `Cancel` to switch back to the VM table. Stop VM `test02`.
14. Edit the VM `test02` again. Go to the `Volumes` tab and change `Size` to `6 GiB`. Press `Save`. This time the modification should be applied without an error.

## Case 3:
1. Go to the UI of your Rancher standalone installation (NOT the embedded one in Harvester).
2. Make sure your cluster is imported in the `Virtualization Management`.
3. Click the Harvester cluster on the `Name` field of the datatable.
4. Go to `Projects/Namespaces` and create a new project by clicking `Create Project`.
5. In the `Resource Quotas` tab, add `Storage Reservation` and use `20000 MiB` as project limit and `12000 MiB` as namespace default limit.
6. Create a new namespace named `quota-test` in the previously created project. Take over the suggested resource quota of `12 GiB`.
7. In the Harvester UI, create a new VM named `test01` with the namespace `quota-test`. In `Volumes`, make sure the volume occupies `7 GiB`.
8. Press `Create`. The VM should be created and started. All should be fine.
9. Stop the `test01`VM.
10. Select the `Clone` action menu on VM `test01`.
11. Enter `test02` as new VM name. Check the `clone volume data` checkbox in the displayed dialog. Press `Create`.
12. An error message should be displayed: `admission webhook "validator.harvesterhci.io" denied the request: storage insufficient resources due to resource quota`. This is because the storage quota of `12 GiB` has been exceeded (2 x 7GiB = 14 GiB).

grafik

### DIFF
--- a/pkg/util/resourcequota/error.go
+++ b/pkg/util/resourcequota/error.go
@@ -3,7 +3,7 @@ package resourcequota
 import "fmt"
 
 const (
-	ErrInsufficientResourcesFMT = "%s insufficient resources"
+	ErrInsufficientResourcesFMT = "%s insufficient resources due to resource quota"
 )
 
 type InsufficientResourceError struct {
@@ -26,6 +26,10 @@ func cpuInsufficientResourceError() error {
 
 func memInsufficientResourceError() error {
 	return newInsufficientResourceError("memory")
+}
+
+func storageInsufficientResourceError() error {
+	return newInsufficientResourceError("storage")
 }
 
 func IsInsufficientResourceError(err error) bool {
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -75,6 +75,11 @@ func (v *vmValidator) Create(_ *types.Request, newObj runtime.Object) error {
 	if err := v.checkVMSpec(vm); err != nil {
 		return err
 	}
+
+	if err := v.checkStorageResourceQuota(vm, nil); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -91,6 +96,10 @@ func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj run
 	oldVM := oldObj.(*kubevirtv1.VirtualMachine)
 	if oldVM == nil {
 		return nil
+	}
+
+	if err := v.checkStorageResourceQuota(newVM, oldVM); err != nil {
+		return err
 	}
 
 	// Prevent users to stop/restart VM when there is VMBackup in progress.
@@ -289,4 +298,8 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
 		return werror.NewInvalidError("reservedMemory cannot be less than 0", field)
 	}
 	return nil
+}
+
+func (v *vmValidator) checkStorageResourceQuota(vm *kubevirtv1.VirtualMachine, oldVM *kubevirtv1.VirtualMachine) error {
+	return v.rqCalculator.CheckStorageResourceQuota(vm, oldVM)
 }
--- a/pkg/util/resourcequota/calculator.go
+++ b/pkg/util/resourcequota/calculator.go
@@ -21,8 +21,9 @@ import (
 )
 
 var resourceQuotaConversion = map[string]string{
-	"limitsCpu":    "limits.cpu",
-	"limitsMemory": "limits.memory",
+	"limitsCpu":       string(corev1.ResourceLimitsCPU),
+	"limitsMemory":    string(corev1.ResourceLimitsMemory),
+	"requestsStorage": string(corev1.ResourceRequestsStorage),
 }
 
 type Calculator struct {
@@ -92,8 +93,7 @@ func (c *Calculator) CheckIfVMCanStartByResourceQuota(vm *kubevirtv1.VirtualMach
 	}
 
 	// get resource quota limits from ResourceQuota
-	selector := labels.Set{util.LabelManagementDefaultResourceQuota: "true"}.AsSelector()
-	rqs, err := c.rqCache.List(vm.Namespace, selector)
+	rqs, err := c.getResourceQuota(vm)
 	if err != nil {
 		return err
 	} else if len(rqs) == 0 {
@@ -119,11 +119,20 @@ func (c *Calculator) getNamespaceResourceQuota(vm *kubevirtv1.VirtualMachine) (*
 		return nil, err
 	}
 
-	if resourceQuota.Limit.LimitsCPU == "" && resourceQuota.Limit.LimitsMemory == "" {
+	if resourceQuota.Limit.LimitsCPU == "" && resourceQuota.Limit.LimitsMemory == "" && resourceQuota.Limit.RequestsStorage == "" {
 		return nil, nil
 	}
 
 	return resourceQuota, nil
+}
+
+func (c *Calculator) getResourceQuota(vm *kubevirtv1.VirtualMachine) ([]*corev1.ResourceQuota, error) {
+	selector := labels.Set{util.LabelManagementDefaultResourceQuota: "true"}.AsSelector()
+	rqs, err := c.rqCache.List(vm.Namespace, selector)
+	if err != nil {
+		return nil, err
+	}
+	return rqs, nil
 }
 
 // containsEnoughResourceQuotaToStartVM checks if the VM can be started based on the namespace resource quota limits
@@ -132,7 +141,7 @@ func (c *Calculator) containsEnoughResourceQuotaToStartVM(
 	namespaceResourceQuota *v3.NamespaceResourceQuota,
 	rq *corev1.ResourceQuota) error {
 	// get running migrations' used resource
-	vmimsCPU, vmimsMem, err := c.getRunningVMIMResources(rq)
+	vmimsCPU, vmimsMem, _, err := c.getRunningVMIMResources(rq)
 	if err != nil {
 		return err
 	}
@@ -192,15 +201,16 @@ func (c *Calculator) calculateVMActualOverhead(vm *kubevirtv1.VirtualMachine) *r
 	return &memoryOverhead
 }
 
-func (c *Calculator) getRunningVMIMResources(rq *corev1.ResourceQuota) (cpu, mem resource.Quantity, err error) {
+func (c *Calculator) getRunningVMIMResources(rq *corev1.ResourceQuota) (cpu, mem, storage resource.Quantity, err error) {
 	vms, err := GetResourceListFromMigratingVMs(rq)
 	if err != nil {
-		return cpu, mem, err
+		return cpu, mem, storage, err
 	}
 
 	for _, rl := range vms {
 		cpu.Add(*rl.Name(corev1.ResourceLimitsCPU, resource.DecimalSI))
 		mem.Add(*rl.Name(corev1.ResourceLimitsMemory, resource.BinarySI))
+		storage.Add(*rl.Name(corev1.ResourceRequestsStorage, resource.BinarySI))
 	}
 
 	return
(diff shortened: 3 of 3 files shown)


ibrokethecloud:
The current change will only slowdown the error messages and eventually slowdown of rancher.

We should probably have a validating webhook to validate if there are resource quotes on a namespace and if storage requirement can be met with the new VM, and accordingly allow a VM creation request to be accepted.

we can check if any storage resource quotas exist and decide accordingly if VM should be created or not: https://kubernetes.io/docs/concepts/policy/resource-quotas/#storage-resource-quota

ibrokethecloud:
we should also check with rancher, with the new webhook, the VM creation request would be denied, but rancher provisioning will keep trying to create new machines to reconcile with the desired state, resulting in jobs being created on rancher. This would not be different from say exhausting instances limits on your aws account.

votdev:
Note, the code of the [second commit](https://github.com/harvester/harvester/pull/5434/commits/2fad80c188b689f824ae57204c70a386190c5826) is taken into action by the VM validator webhook:

[Create](https://github.com/harvester/harvester/blob/27e4e3651fb4cb1a26ba4b5bfc210d652f66ccec/pkg/webhook/resources/virtualmachine/validator.go#L75) -> [checkVMSpec](https://github.com/harvester/harvester/blob/27e4e3651fb4cb1a26ba4b5bfc210d652f66ccec/pkg/webhook/resources/virtualmachine/validator.go#L107C23-L107C34) -> [CheckIfVMCanStartByResourceQuota](https://github.com/harvester/harvester/blob/27e4e3651fb4cb1a26ba4b5bfc210d652f66ccec/pkg/util/resourcequota/calculator.go#L61) -> [containsEnoughResourceQuotaToStartVM](https://github.com/votdev/harvester/blob/2fad80c188b689f824ae57204c70a386190c5826/pkg/util/resourcequota/calculator.go#L131)

w13915984028:
This PR and https://github.com/harvester/harvester/pull/5037 will have similar challenge:

The resource quota can protect resource from being over limit in api server side, but if the controller has no knowledge of this error and will reconciler to re-create the resources. Then it will cause kind of DDOS.

Either as @ibrokethecloud said, cooperate Rancher to use webhook to check and stop.

Or fail out the related resources to let the controller stop to reconciler.

cc @FrankYang0529 @starbops

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -66,6 +66,7 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
 		pvcAnno.Namespace = vm.Namespace
 		if pvc, err = h.pvcCache.Get(vm.Namespace, pvcAnno.Name); apierrors.IsNotFound(err) {
 			if _, err = h.pvcClient.Create(pvcAnno); err != nil {
+				h.vmController.EnqueueAfter(vm.Namespace, vm.Name, 5*time.Second)
```
w13915984028: what's the reason of creating PVC failure? delay 5s each failure will delay the creating & starting of vm in normal case.

pkg/util/resourcequota/error.go
```diff
@@ -28,6 +28,10 @@ func memInsufficientResourceError() error {
 	return newInsufficientResourceError("memory")
 }
 
+func storageInsufficientResourceError() error {
+	return newInsufficientResourceError("storage")
```
w13915984028, line 32: `storage insufficient resources` is somehow not straightforward, user may assume that the underlayer storage is insufficient, could we add `due to resource quota` to the error fmt `ErrInsufficientResourcesFMT = "%s insufficient resources"`

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -75,6 +75,11 @@ func (v *vmValidator) Create(_ *types.Request, newObj runtime.Object) error {
 	if err := v.checkVMSpec(vm); err != nil {
 		return err
 	}
+
```
w13915984028, line 78: 1. CPU and memory quota are checked in below path, what's the consideration to isolate `v.checkStorageResourceQuota` away from it?

```
func (v *vmValidator) checkVMSpec(vm *kubevirtv1.VirtualMachine) error {

	return v.rqCalculator.CheckIfVMCanStartByResourceQuota(vm)
```

2. CPU and memory check are done via `CheckIfVMCanStartByResourceQuota` which skips those stopped VMs; but for storage, it is a bit more complex:  the PVC has kind of `later binding` or `instant binding` strategy.

User can also create volumes independently, which are not binded to any VMs in advance.

Are only running VM's PVC computed or all? or taking care of the PVC binding strategy?

w13915984028 (approved):
Good enhancement, with a few questions.

BTW, about the integreation with Rancher when creating downstream RKE2 cluster which creates a few VMs to work as the cluster nodes, I will doble check where is the best place to add check.

pkg/controller/master/virtualmachine/vm_controller.go
votdev: The idea was to reduce noise in the log. See the comment in the PR summary:

> The PVCs creation will be delayed by 5s after failure. This will minimize the number of log entries. This improvement is implemented as a preventative measure.

But since that situation should not happen anymore, we can remove that if it does not meet with approval.

w13915984028: If the `pvcClient.Create` continuously fails, the root cause needs to be identified.

When it is predefined and inevitable, delay operation is ok; otherwise, the default reconciler behaviour seems better.

Generally, here `EnqueueAfter` should work on some kinds of identified `err`, not all `err`.

pkg/webhook/resources/virtualmachine/validator.go
votdev: > CPU and memory quota are checked in below path, what's the consideration to isolate v.checkStorageResourceQuota away from it?

I moved the storage check to a separate function because it has to be implemented differently for VM `Create` and `Update`. I did not want to pass the required parameters through dozens of functions to the corresponding `containsEnoughResourceQuotaToStartVM`. In this case, several functions would have had to be adapted. This way it is nicely separated by logic and no function headers had to be adapted just so that a function 3 call levels lower gets the required parameters.

votdev: > CPU and memory check are done via CheckIfVMCanStartByResourceQuota which skips those stopped VMs; 

That's right and that's an additional reason why I have also outsourced the functionality in a separate function. All VMs are taken into action in the storage quota assessment. See test case 3 which takes this into account.

votdev: > it is a bit more complex: the PVC has kind of later binding or instant binding strategy.

The assessment is based on the `harvesterhci.io/volumeClaimTemplates` annotation. It is therefore irrelevant when the storage is created.

If this is to be taken into account, then this must still be implemented; but then I assume that the quota check/assessment cannot be done BEFORE the VM is created.

The question here is how complex the whole thing should be and whether we want it to be.

pkg/controller/master/virtualmachine/vm_controller.go
votdev: OK, i've removed that as it is not really needed for this feature.

ibrokethecloud:
The webhook changes work as expected and VM creation is blocked when storage quota is exhausted. For traditional vm workloads this works fine.

There however seems to be a unexpected side effect when provisioning VM's for guest clusters. When the resource quote limit is hit, the webhook rejects the VM creation request, which results in the machine provisioning failing, and rancher retries with a new machine immediately. As a result it seems like rancher keeps making machine creation requests. There is a momentary transient error in the cluster which is not captured as the machine is deleted and recreated


image

Currently the machine creation hangs and eventually times out as the VM is unschedulable, since the pvc cannot be created due to resource quote exhaustion.

image

The current state of VM never being provisioned may actually be throttling the retries in rancher, but the change causes immediate retry. We should probably discuss with rancher team if there is a way to capture this better?

@bk201 not sure if you have any views on this? should we go ahead and merge it since the provisioning error is better than the hidden scheduling failure?

FrankYang0529 (approved):
All test cases work as expected. Thanks.

bk201:
>When the resource quote limit is hit, the webhook rejects the VM creation request, which results in the machine provisioning failing, and rancher retries with a new machine immediately. 

@ibrokethecloud Do you know if Rancher retries with a short-period tight loop or with exponential backoff? The initial issue we want to handle is not to flood terminating VMs in the GUI and crash the GUI. If the change doesn't make rancher backend busy on retrying, I guess it's also a way to fix the problem.

ibrokethecloud:
@bk201 seems almost instantaneous. 

From my testing.. 
```
2024/05/28 01:59:25 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 01:59:25 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 01:59:26 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 01:59:26 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 01:59:26 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 01:59:26 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 01:59:29 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 02:00:07 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 02:00:09 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 02:00:09 [INFO] [machineprovision] fleet-default/quota-test-pool1-3dd722b8-6f7gd: reconciling machine job
2024/05/28 02:00:09 [INFO] [machineprovision] Machine rke-machine.cattle.io/v1, Kind=HarvesterMachine quota-test-pool1-3dd722b8-6f7gd has already been deleted
2024/05/28 02:04:27 [ERROR] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: error getting machine by owner reference: no matching controller owner ref
2024/05/28 02:04:27 [ERROR] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: error getting machine by owner reference: no matching controller owner ref
2024/05/28 02:04:27 [ERROR] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: error getting machine by owner reference: no matching controller owner ref
2024/05/28 02:04:27 [ERROR] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: error getting machine by owner reference: no matching controller owner ref
2024/05/28 02:04:27 [ERROR] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: error getting machine by owner reference: no matching controller owner ref
2024/05/28 02:04:37 [INFO] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: reconciling machine job
2024/05/28 02:04:37 [INFO] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: reconciling machine job
2024/05/28 02:04:38 [INFO] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: reconciling machine job
2024/05/28 02:04:38 [INFO] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: reconciling machine job
2024/05/28 02:04:38 [INFO] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: reconciling machine job
2024/05/28 02:04:38 [INFO] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: reconciling machine job
2024/05/28 02:04:38 [INFO] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: reconciling machine job
2024/05/28 02:04:39 [INFO] [machineprovision] fleet-default/test-k3s-pool1-9e269de9-btg8q: reconciling machine job
``` 
machine provisioning fails, which triggers deletion and recreation:

https://github.com/rancher/rancher/blob/v2.8.2/pkg/controllers/capr/machineprovision/controller.go#L564-L574

w13915984028 (approved):
LGTM, thanks.

Regarding the concerns, we may contract Rancher to negotiate about checking the return error when calling node-driver, if error is like `insufficient resources`, then Rancher may slow down reconciler instead of tight loop.

votdev:
> LGTM, thanks.
> 
> Regarding the concerns, we may contract Rancher to negotiate about checking the return error when calling node-driver, if error is like `insufficient resources`, then Rancher may slow down reconciler instead of tight loop.

Question is raised in [discuss-rancher](https://suse.slack.com/archives/C02CX064EBX/p1718013920808399).

votdev:
@ibrokethecloud Is the PR now accepted and can we merge it?

ibrokethecloud (approved):
lgtm. thanks.

we should clearly document the impact of this change when provisioning with rancher, as the immediate failure / retry in rancher may cause rancher to be overwhelmed.

votdev:
> lgtm. thanks.
> 
> we should clearly document the impact of this change when provisioning with rancher, as the immediate failure / retry in rancher may cause rancher to be overwhelmed.

Please see https://github.com/harvester/docs/pull/645

Issue #5165, opened 2024-02-19
[BUG] storage class name conflicts for different namespace vmimages with a same name


FrankYang0529:
**Describe the bug**

We use `longhorn-<vmimage name>` as storage class name. The StorageClass doesn't have namesapce, but vmimage has. If there are same vmimage name in different namespace, storage class name will be conflicts.

https://github.com/harvester/harvester/blob/1e8c677c49005dfd312c7143f23a2632b83c42e5/pkg/controller/master/image/vm_image_controller.go#L214-L217

**To Reproduce**
1. Create a Harvester.
2. Using terraform to create two vmimage `default/test-image` and `harvester-public/test-image`.
3. After two vmimage are ready, deleting `default/test-image`.
4. Create a VM with `harvester-public/test-image` vmimage. The root disk PVC can't be provisioned.

**Expected behavior**
The storage class name doesn't conflict.

**Environment**
 - Harvester ISO version: v1.2.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

bk201:
We can only support this with a Harvester version plus some Rancher patch versions because the Rancher need to have the updated node driver included. For example.

~~Harvester v1.5.0 + Rancher v2.8.x~~
~~Harvester v1.5.0 + Rancher v2.9.3~~
Harvester v1.5.0 + Rancher v2.11.0

Update: Harvester v1.5.0 will only support Rancher v2.11

bk201:
The fix won't be backport to v1.3/v1.4 because the node drivers in supported Ranchers might not support this change.
In Harvester v1.5.0, we'll support only Rancher v2.11. In this way, we can assure Rancher has a correct node driver.

WebberHuang1118:
I will create a new PR after [VM Image Controller, Webhook and Download/Upload Refactor for Various Backend Support](https://github.com/harvester/harvester/pull/7170) is merged.

bk201:
move to 1.6.0 and target Rancher 2.12.0

Vicente-Cheng:
Hi @WebberHuang1118, 
Any update on this issue? I saw the above PRs were merged.
Thanks!

WebberHuang1118:
> Hi [@WebberHuang1118](https://github.com/WebberHuang1118), Any update on this issue? I saw the above PRs were merged. Thanks!

Hi @Vicente-Cheng 

[...]

PR #5459, opened 2024-03-26
feat: add new storage class name format

FrankYang0529:

**Problem:**
We use `longhorn-<vmimage name>` as storage class name. The StorageClass doesn't have namesapce, but vmimage has. If there are same vmimage name in different namespace, storage class name will be conflicts.

https://github.com/harvester/harvester/blob/1e8c677c49005dfd312c7143f23a2632b83c42e5/pkg/controller/master/image/vm_image_controller.go#L214-L217

**Solution:**
Use `namespace` and image `UID` as name format.

**Related Issue:**
https://github.com/harvester/harvester/issues/5165

**Additional content:**
~Dashboard doesn't use `vmimage.status.storageClassName` to construct value in `harvesterhci.io/volumeClaimTemplates` annotation. It uses format like `longhorn-<vmimage-name>` as storage class name, so we can't create VM by dashboard with this PR.~ Already fixed by https://github.com/harvester/dashboard/pull/986.

**Test plan:**
### Case 1: VM basic operations (create / backup /restore) without error.

1. Create a VMImage.
2. Create a VM without error.
3.. Write some data.
4. Setup backup-target.
5. Backup the VM.
6. Restore a new VM. New VM can start without error.

### Case 2: VMImage with a same name in different namespaces can work
1. Create a VMImage `thisistest` in `default` and `harvester-public` namespaces.
2. Both images can be created without error.

### Case 3: An old VMImage can work without error.
1. Create a Harvester cluster v1.3.0.
2. Create a VMImage.
3. Replace harvester and harvester-webhook deployment with this PR.
4. The VMImage can work without error.

### Case 4: Create VM Template without error
1. Follow case 1.
2. Generate Template with data.
3. After template is ready, launch a new VM and we can see default image name starts with `templateversion-`. New VM can start without error and data is in the new VM.
Screenshot 2024-04-10 at 3 36 17 PM

### DIFF
--- a/pkg/util/image.go
+++ b/pkg/util/image.go
@@ -7,6 +7,8 @@ import (
 	"github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	lhutil "github.com/longhorn/longhorn-manager/util"
+	ctlstoragev1 "github.com/rancher/wrangler/pkg/generated/controllers/storage/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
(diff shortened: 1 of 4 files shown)


WebberHuang1118:
@FrankYang0529 Do we need to verify the backward compatibility? Create an image without this PR and update Harvester with this PR to start a new VM from the existing image? WDYT? Thanks.

FrankYang0529:
> @FrankYang0529 Do we need to verify the backward compatibility? Create an image without this PR and update Harvester with this PR to start a new VM from the existing image? WDYT? Thanks.

Yes, added it as case 3. Thank you.

pkg/util/image.go
```diff
@@ -49,10 +51,43 @@ func GetBackingImageDataSourceName(backingImageCache ctllhv1.BackingImageCache,
...
 
-func GetImageStorageClassName(imageName string) string {
+func storageClassLegacyName(imageName string) string {
 	return fmt.Sprintf("longhorn-%s", imageName)
 }
 
+func GenerateStorageClassName(imageNamespace, imageName string) string {
+	return lhutil.AutoCorrectName(fmt.Sprintf("longhorn-%s-%s", imageNamespace, imageName), lhdatastore.NameMaximumLength)
+}
+
```
WebberHuang1118: @FrankYang0529 
I'm just wondering if we use the image's UUID to name the storage class, would it be more concise?

FrankYang0529: Update it to use namespace and UID. Thank you.

pkg/util/image.go
```diff
@@ -49,10 +51,43 @@ func GetBackingImageDataSourceName(backingImageCache ctllhv1.BackingImageCache,
...
 }
 
-func GetImageStorageClassName(imageName string) string {
+func storageClassLegacyName(imageName string) string {
 	return fmt.Sprintf("longhorn-%s", imageName)
 }
 
+func GenerateStorageClassName(imageNamespace, imageUID string) string {
+	return lhutil.AutoCorrectName(fmt.Sprintf("longhorn-%s-%s", imageNamespace, imageUID), lhdatastore.NameMaximumLength)
+}
```
WebberHuang1118:
```suggestion
func GenerateStorageClassName(imageNamespace, imageUID string) string {
	return lhutil.AutoCorrectName(fmt.Sprintf("longhorn-%s", imageUID), lhdatastore.NameMaximumLength)
}
```
As we plan to use image uid to name sc, maybe namespace is not necessary?

WebberHuang1118 (approved):
LGTM, just a nit, thanks.

WebberHuang1118:
`Dashboard doesn't use vmimage.status.storageClassName to construct value in harvesterhci.io/volumeClaimTemplates annotation. It uses format like longhorn-<vmimage-name> as storage class name, so we can't create VM by dashboard with this PR.`

Maybe front-end should have a fix at the same time, otherwise, merging this PR could make VM unable to be created from the template.

FrankYang0529:
> `Dashboard doesn't use vmimage.status.storageClassName to construct value in harvesterhci.io/volumeClaimTemplates annotation. It uses format like longhorn-<vmimage-name> as storage class name, so we can't create VM by dashboard with this PR.`
> 
> Maybe front-end should have a fix at the same time, otherwise, merging this PR could make VM unable to be created from the template.

@WebberHuang1118 Thanks for the review. I create another PR for it https://github.com/harvester/harvester/issues/5551.

pkg/util/image.go
FrankYang0529: Thank you. I update to use UID only.

bk201 (approved):

WebberHuang1118:
Verified all the testing plans with the latest version, and all work as expected, thanks.

FrankYang0529:
@mergify backport v1.2 v1.3

Issue #5402, opened 2024-03-19
[ENHANCEMENT] Structured logging for Harvester components


connorkuehl:
**Is your enhancement related to a problem? Please describe.**

There are a number of benefits to using structured logging in Harvester. Firstly, it would make Harvester support bundles more compatible with being fed into something like Elasticsearch (or any alternative). This would allow us to write queries, for example, to see all of the log messages emitted about a specific virtualmachine across all Harvester components.

Ideally, we would be able to feed an entire support bundle into Elasticsearch (or any alternative) to improve our diagnostic capability and establish root causes sooner.

This would improve many more things beyond diagnostic efforts. For example, this could open up more possibilities for Harvester users to customize monitoring/alerting based on queryable logs.

Most of the Harvester codebases already use a structured logging package called "logrus", but we are not taking full advantage of it.

For example, it's common to see logging calls like this in Harvester code:

```go
logrus.Infof("could not shut down %s/%s", vmi.Namespace, vmi.Name)
```

Which would result in a log message looking something like this:

```json
{"msg":"could not shut down default/test"}
```

This is very limiting, as we could only potentially query on strings or substrings in that message, for example:

```
msg == "could not shut down*"
```

Or, if we had the ability to write a query and we wanted to see everything about a specific VM:

```
msg == "test"
```

It was perhaps unfair of me to choose "test" as the VM's name, but there are a number of common words that might conflict with a VM's name, and a query like that would show us a lot of irrelevant logs.

**Describe the solution you'd like**

Instead, Harvester components should standardize on a cross-repo structured logging methodology. Adapting the above example, we would log like this:

```go
logrus.WithFields(logrus.Fields {
    "k8s_namespace": vmi.Namespace,
    "virtualmachine_name": vmi.Name,
}).Info("could not shut down VM")
```

This is a lot more flexible for writing queries against. For example, if we achieved a high level of consistency with logging and how we name our structured log keys, we could search for everything that happened to a VM:

```
virtualmachine_name == "test" AND k8s_namespace == "default"
```

Or query to see all of the VMs we failed to shutdown:

```
msg == "could not shut down VM"
```

**Describe alternatives you've considered**

**Additional context**

votdev:
Can we make a list of field names we want to use? E.g. i do not like the `k8s_namespace`, i suggest to use `namespace` instead.
[...]

PR #5474, opened 2024-03-27
HEP: Structured Logging

m-ildefons:
Add new HEP to introduce structured logging to Harvester.

**Related Issue:**
related-to: #5402

**Test plan:**
No tests, only HEP

### DIFF
--- a/enhancements/20240327-structured-logging.md
+++ b/enhancements/20240327-structured-logging.md
@@ -1,0 +1,109 @@
+# Structured Logging
+
+## Summary
+
+Harvester should emit only structured log messages. Existing non-structured log
+messages should be converted to structured log messages.
+
+### Related Issues
+
+- https://github.com/harvester/harvester/issues/5402
+
+## Motivation
+
+Unstructured logging is nice to read for humans, but the lack of structure makes
+it hard to apply tooling to it. This in turn makes it hard to manage large
+amounts of logs, apply automatic analysis and fault detection and filtering.
+Therefore we should utilize structured logging whenever possible.
+Log analysis is not only a concern for Harvester users, but also for Harvester
+developers and support engineers, when they need to analyze support bundles.
+
+### Goals
+
+- All log messages emitted by Harvester should be structured
+- The structure should utilize standardized field names
+- Implementation should require a minimal amount of new dependencies
+- A migration strategy for the existing code base to convert existing log
+  messages to structured logging
+
+### Non-Goals
+
+This HEP does not cover:
+
+- integration with log analysis tools
+- log storage, access, viewing, retention policies
+- when to log or what severity to log at
+
+## Proposal
+
+To implement this proposal, several steps should be taken. First, the developer
+guide should receive an update, describing how developers should implement
+logging. This includes the mention of which library to use for structured
+logging and how to implement structured logging with this library, including any
+standard fields and conventions.
+Second, all existing log messages should be converted to structured logging.
+Lastly, a CI check should be put in place to prevent new additions of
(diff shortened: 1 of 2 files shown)


enhancements/20240327-structured-logging.md
```diff
@@ -0,0 +1,109 @@
...
+## Proposal
+
+To implement this proposal, several steps should be taken. First, the developer
+guide should receive an update, describing how developers should implement
+logging. This includes the mention of which library to use for structured
+logging and how to implement structured logging with this library, including any
+standard fields and conventions.
+Second, all existing log messages should be converted to structured logging.
+Lastly, a CI check should be put in place to prevent new additions of
+unstructured logging, e.g. using the tool [Logcheck](https://github.com/kubernetes-sigs/logtools/tree/main/logcheck)
```
connorkuehl, line 46: This is a great find!

enhancements/20240327-structured-logging.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+These are possible field names to use:
+
+| Field Name | K8s Resource Field | Description |
+| ---------- | ------------------ | ----------- |
+| namespace | `.metadata.namespace` | K8s namespace of the subject of the message |
+| name | `.metadata.name` | Name of the K8s resource which is the subject of the message |
+| kind | `.kind` | Kind of the K8s resource which is subject of the log message |
+
```
connorkuehl:
```suggestion
| kind | `.kind` | Kind of the K8s resource which is subject of the log message |
| err | `err` | String representation of a `builtin.error` value |
```

connorkuehl (approved):
This looks great, thank you, Moritz! I don't think any of it is very controversial, so I've gone ahead and captured some of the action items described in the HEP as GitHub issues and linked them to this PR.

enhancements/20240327-structured-logging.md
starbops: Should we include `apiVersion` or `apiGroup` here? Since there might be multiple kinds of objects that share the same name in `kind`, users might need to differentiate among them. For example, the `clusters.provisioning.cattle.io` and `clusters.cluster.x-k8s.io` CRDs are both `Cluster` kinds but in different groups.

m-ildefons: Both are excellent suggestions. Keep in mind, that this table isn't meant to be exhaustive by any means, but just a suggestion for a possible convention. A complete list of standard fields should be maintained in the developer docs, since the likelihood of future new developers reading through old HEPs is fairly low.

enhancements/20240327-structured-logging.md
```diff
@@ -0,0 +1,111 @@
...
+To implement this proposal, several steps should be taken. First, the developer
+guide should receive an update, describing how developers should implement
+logging. This includes the mention of which library to use for structured
+logging and how to implement structured logging with this library, including any
+standard fields and conventions.
+Second, all existing log messages should be converted to structured logging.
+Lastly, a CI check should be put in place to prevent new additions of
+unstructured logging, e.g. using the tool [Logcheck](https://github.com/kubernetes-sigs/logtools/tree/main/logcheck)
+
+### Loggin infrastructure
```
tserong:
```suggestion
### Logging infrastructure
```

tserong (approved):
LGTM!

(also I found a tiny typo)

starbops (approved):
LGTM, thank you!

enhancements/20240327-structured-logging.md
m-ildefons: Good spot. Fixed

PR #5488, opened 2024-03-29
chore: switch UI version to release-harvester-v1.2

bk201:
We should use the release branch from UI, not a tag release.


brandboat (approved):
LGTM !

starbops (approved):

Issue #5480, opened 2024-03-28
[BUG] upgrade will stuck without any error meessage if the checksum is invalid 


lanfon72:
### Describe the bug
The upgrade will stuck if we input the checksum like `v`

### To Reproduce
1. Install Harvester with any nodes
2. Create upgrade version
3. Perform upgrade

### Expected behavior
Upgrade should failed with error message

## Environment:
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360G9 3 nodes**
 - Harvester ISO version: **v1.2.1**
 - **ui-source** Option: **Auto**

### Additional context
It would similar  #3123 but the check sum is invalid, in #3123, we will detect that the **inputted** is mismatch to the ISO's, but this issue is, the **inputted** checksum is not a **valid** checksum, maybe length is not enough, or something else.

starbops:
For detailed reproduced steps:

1. Create the Version CR with a **malformed** (not just mismatched) checksum string

```shell
$ cat <<EOF | kubectl apply -f -
apiVersion: harvesterhci.io/v1beta1
kind: Version
metadata:
  namespace: harvester-system
  name: master
spec:
  isoURL: http://123.123.123.123:8000/harvester/master/harvester-master-amd64.iso
  isoChecksum: 'asdf'
EOF
```

2. Check the created Version CR

```shell
$ kubectl -n harvester-system get versions master -o yaml
apiVersion: harvesterhci.io/v1beta1
kind: Version
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"harvesterhci.io/v1beta1","kind":"Version","metadata":{"annotations":{},"name":"master","namespace":"harvester-system"},"spec":{"isoChecksum":"asdf","isoURL":"http://123.123.123.123:8000/harvester/master/harvester-master-amd64.iso"}}
  creationTimestamp: "2024-03-28T08:18:45Z"
  generation: 1
  name: master
  namespace: harvester-system
  resourceVersion: "44990"
  uid: c70b4396-9c6c-47b2-b452-56c6eabdba14
spec:
  isoChecksum: asdf
  isoURL: http://123.123.123.123:8000/harvester/master/harvester-master-amd64.iso
```

3. Click the "Upgrade" button on the dashboard to start the upgrade
4. Check the created VMImage CR

```shell
$ kubectl -n harvester-system get vmimages hvst-upgrade-r2c77 -o yaml
[...]

PR #5498, opened 2024-04-02
Fix upgrade stuck when using incorrect isoChecksum in version CR

brandboat:
**Problem:**
The upgrade stuck when isoChecksum isn't in SHA512 format in Version CR. The root cause is in `createBackingImageAndStorageClass` we missed to invoke `handleFail` thus the max retry never reached to maximum count and lead to the infinite loop of vm_image_controller.go#onChanged

**Solution:**
Applied two things here.
1. Add Version isoChecksum validation, make it fail fast if the checksum isn't in SHA512 format.
2. Invoke handleFail when BackingImage, StorageClass creation fails.
  though 1. could avoid invalid isoChecksum passed to BackingImage, the VMImage required two things, StorageClass and BackingImage, and there might have other potential failures in BackingImage/StorageClass creation other than invalid isoChecksum.

**Related Issue:**
related to #5480

**Test plan:**
1. **Test version with incorrect isoChecksum** 
After applying the following version.yaml, the console should return error message like `admission webhook "validator.harvesterhci.io" denied the request: invalid isoChecksum asdf` since asdf isn't in sha512 format.
    ```yaml
    kind: Version
    metadata:
      name: master
      namespace: harvester-system
    spec:
      isoURL: http://192.168.100.1:8000/harvester-v1.3.0-amd64.iso
      isoChecksum: asdf
    ```
2. **Test version with correct isoChecksum and run upgrade** 
  - Apply the correct version.yaml to K8s
  - Click the upgrade button in harvester dashboard
  - Everything should works fine
3. **Test vm image**
  - Go to harvester UI `images` and create image with mismatch SHA512 checksum
  - The vm image should download 4 times and then failed with error message like `Failed to process sync file: the expected checksum ... doesn't match the file actual checksum ...`

### DIFF
--- a/pkg/controller/master/image/vm_image_controller.go
+++ b/pkg/controller/master/image/vm_image_controller.go
@@ -26,13 +26,14 @@ import (
 )
 
 const (
-	checkBackingImageInterval = 1 * time.Second
+	checkInterval = 1 * time.Second
 )
 
 // vmImageHandler syncs status on vm image changes, and manage a storageclass & a backingimage per vm image
 type vmImageHandler struct {
 	httpClient        http.Client
 	storageClasses    ctlstoragev1.StorageClassClient
+	storageClassCache ctlstoragev1.StorageClassCache
 	images            ctlharvesterv1.VirtualMachineImageClient
 	imageController   ctlharvesterv1.VirtualMachineImageController
 	backingImages     ctllhv1.BackingImageClient
@@ -59,32 +60,48 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
 		return image, nil
 	}
 
-	needRetry := false
+	return h.processVMImage(image)
+}
+
+func (h *vmImageHandler) processVMImage(image *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
 	bi, err := util.GetBackingImage(h.backingImageCache, image)
 	if err != nil {
 		if !errors.IsNotFound(err) {
-			logrus.Errorf("%v/%v failed to get backing image, err: %v", image.Namespace, image.Name, err)
+			logrus.WithError(err).WithFields(logrus.Fields{
+				"namespace": image.Namespace,
+				"name":      image.Name,
+			}).Error("failed to get backing image for vmimage")
 			return image, err
 		}
-		needRetry = true
-	} else if bi.DeletionTimestamp != nil {
-		h.imageController.EnqueueAfter(image.Namespace, image.Name, checkBackingImageInterval)
+		return h.handleRetry(image)
+	}
+
+	if bi.DeletionTimestamp != nil {
+		h.imageController.EnqueueAfter(image.Namespace, image.Name, checkInterval)
 		return image, nil
-	} else {
-		for _, status := range bi.Status.DiskFileStatusMap {
-			if status.State == lhv1beta2.BackingImageStateFailed {
-				needRetry = true
-				break
-			}
+	}
+
+	for _, status := range bi.Status.DiskFileStatusMap {
+		if status.State == lhv1beta2.BackingImageStateFailed {
+			return h.handleRetry(image)
 		}
 	}
 
-	if needRetry {
-		if image.Status.LastFailedTime == "" {
-			return h.initialize(image)
+	sc, scErr := h.storageClassCache.Get(util.GetImageStorageClassName(image.Name))
+	if scErr != nil {
+		if !errors.IsNotFound(scErr) {
+			logrus.WithError(scErr).WithFields(logrus.Fields{
+				"namespace": image.Namespace,
+				"name":      image.Name,
+			}).Error("failed to get storage class for vmimage")
+			return image, scErr
 		}
-
 		return h.handleRetry(image)
+	}
+
+	if sc != nil && sc.DeletionTimestamp != nil {
+		h.imageController.EnqueueAfter(image.Namespace, image.Name, checkInterval)
+		return image, nil
 	}
 
 	return image, nil
@@ -105,13 +122,17 @@ func (h *vmImageHandler) initialize(image *harvesterv1.VirtualMachineImage) (*ha
 		return image, err
 	}
 
-	image, err := h.checkImage(image)
+	toUpdate := image.DeepCopy()
+	toUpdate, err := h.checkImage(toUpdate)
 	if err != nil {
-		logrus.Errorf("failed to check image %s/%s, err: %v", image.Namespace, image.Name, err)
-		h.imageController.EnqueueAfter(image.Namespace, image.Name, 10*time.Second)
-		return h.images.Update(image)
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"namespace": toUpdate.Namespace,
+			"name":      toUpdate.Name,
+		}).Error("failed to check vmimage")
+		return h.images.Update(toUpdate)
 	}
-	return h.createBackingImageAndStorageClass(image)
+
+	return h.createBackingImageAndStorageClass(toUpdate)
 }
 
 func (h *vmImageHandler) checkImage(image *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
@@ -119,37 +140,46 @@ func (h *vmImageHandler) checkImage(image *harvesterv1.VirtualMachineImage) (*ha
 		return image, nil
 	}
 
-	toUpdate := image.DeepCopy()
-	toUpdate.Status.AppliedURL = image.Spec.URL
+	image.Status.AppliedURL = image.Spec.URL
 	resp, err := h.httpClient.Head(image.Spec.URL)
 	if err != nil {
-		toUpdate = handleFail(toUpdate, condition.Cond(harvesterv1.ImageInitialized), err)
-		return toUpdate, err
+		image = handleFail(image, condition.Cond(harvesterv1.ImageInitialized), err)
+		return image, err
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusBadRequest {
 		err = fmt.Errorf("got %d status code from %s", resp.StatusCode, image.Spec.URL)
-		toUpdate = handleFail(toUpdate, condition.Cond(harvesterv1.ImageInitialized), err)
-		return toUpdate, err
+		image = handleFail(image, condition.Cond(harvesterv1.ImageInitialized), err)
+		return image, err
 	}
 
 	if resp.ContentLength > 0 {
-		toUpdate.Status.Size = resp.ContentLength
+		image.Status.Size = resp.ContentLength
 	}
-	return toUpdate, nil
+	return image, nil
 }
 
 func (h *vmImageHandler) createBackingImageAndStorageClass(image *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
 	if err := h.createBackingImage(image); err != nil && !errors.IsAlreadyExists(err) {
-		return nil, err
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"namespace": image.Namespace,
+			"name":      image.Name,
+		}).Error("failed to create backing image for vmimage")
+		toUpdate := handleFail(image, condition.Cond(harvesterv1.ImageInitialized), err)
+		return h.images.Update(toUpdate)
 	}
 	if err := h.createStorageClass(image); err != nil && !errors.IsAlreadyExists(err) {
-		return nil, err
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"namespace": image.Namespace,
+			"name":      image.Name,
+		}).Error("failed to create storage class for vmimage")
+		toUpdate := handleFail(image, condition.Cond(harvesterv1.ImageInitialized), err)
+		return h.images.Update(toUpdate)
 	}
 
 	toUpdate := image.DeepCopy()
-	toUpdate.Status.AppliedURL = toUpdate.Spec.URL
+	toUpdate.Status.AppliedURL = image.Spec.URL
 	toUpdate.Status.StorageClassName = util.GetImageStorageClassName(image.Name)
 	toUpdate.Status.Progress = 0
 
@@ -165,6 +195,10 @@ func (h *vmImageHandler) createBackingImageAndStorageClass(image *harvesterv1.Vi
 }
 
 func (h *vmImageHandler) createBackingImage(image *harvesterv1.VirtualMachineImage) error {
+	if cachedBI, _ := util.GetBackingImage(h.backingImageCache, image); cachedBI != nil && cachedBI.DeletionTimestamp != nil {
+		return fmt.Errorf("backing image %s is being deleted", cachedBI.Name)
+	}
+
 	biName, err := util.GetBackingImageName(h.backingImageCache, image)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


brandboat:
This pr is based on master branch, while the issue also happens in v1.3 & v1.2 branch, so we should backport this pr after merged to master.

pkg/webhook/resources/version/validator.go
```diff
@@ -55,5 +60,25 @@ func (v *versionValidator) checkAnnotations(version *v1beta1.Version) error {
...
+}
+
+func checkISOChecksum(version *v1beta1.Version) error {
+	isoChecksum := version.Spec.ISOChecksum
+	// if an isoChecksum is provided, it must be in the SHA-512 format
+	// since Longhorn backing images only accept hashes in that format
+	if isoChecksum == "" || SHA512Pattern.MatchString(isoChecksum) {
+		return nil
+	}
+	return werror.NewBadRequest(fmt.Sprintf("invalid isoChecksum %s", isoChecksum))
```
starbops: I would suggest the following:

```suggestion
	if isoChecksum != "" && !SHA512Pattern.MatchString(isoChecksum) {
		return werror.NewBadRequest(fmt.Sprintf("invalid isoChecksum %s", isoChecksum))
	}
	return nil
```

It's more consistent and idiomatic.

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -142,10 +149,14 @@ func (h *vmImageHandler) checkImage(image *harvesterv1.VirtualMachineImage) (*ha
...
-		return nil, err
+		logrus.Errorf("failed to create backing image in image %s/%s, err: %v", image.Namespace, image.Name, err)
+		toUpdate := handleFail(image, condition.Cond(harvesterv1.ImageInitialized), err)
+		return toUpdate, err
 	}
 	if err := h.createStorageClass(image); err != nil && !errors.IsAlreadyExists(err) {
-		return nil, err
+		logrus.Errorf("failed to create storage class in image %s/%s, err: %v", image.Namespace, image.Name, err)
+		toUpdate := handleFail(image, condition.Cond(harvesterv1.ImageInitialized), err)
+		return toUpdate, err
```
starbops: You might need to deep-copy VMImage object first before updating it. Otherwise, you're manipulating the cache, which causes trouble.

pkg/webhook/resources/version/validator.go
brandboat: Sure, that make sense to me.

pkg/controller/master/image/vm_image_controller.go
brandboat: My bad :sweat:, thanks for the comment. Already address comments in the latest commit. Huge thanks @starbops

starbops: It would be nice if you could move the line `toUpdate := image.DeepCopy()` to the first line of this function so there won't be multiple `DeepCopy()` calls scattered around.

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -111,7 +113,12 @@ func (h *vmImageHandler) initialize(image *harvesterv1.VirtualMachineImage) (*ha
 		h.imageController.EnqueueAfter(image.Namespace, image.Name, 10*time.Second)
 		return h.images.Update(image)
 	}
-	return h.createBackingImageAndStorageClass(image)
+
+	image, err = h.createBackingImageAndStorageClass(image)
+	if err != nil {
+		return h.images.Update(image)
+	}
+	return h.images.Update(image)
```
FrankYang0529: It looks like no matter there is `err` or not, we do the same operation `h.images.Update(image)`. Do we want to call `toUpdate := handleFail(image.DeepCopy(), condition.Cond(harvesterv1.ImageInitialized), err)` here instead of in `createBackingImageAndStorageClass`?

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -111,7 +113,12 @@ func (h *vmImageHandler) initialize(image *harvesterv1.VirtualMachineImage) (*ha
 		h.imageController.EnqueueAfter(image.Namespace, image.Name, 10*time.Second)
 		return h.images.Update(image)
 	}
-	return h.createBackingImageAndStorageClass(image)
+
+	image, err = h.createBackingImageAndStorageClass(image)
```
FrankYang0529: Both `checkImage` and `createBackingImageAndStorageClass` call `toUpdate := image.DeepCopy()`. How about we do it in `initialize` and send the copied one to these two functions?

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -142,10 +149,14 @@ func (h *vmImageHandler) checkImage(image *harvesterv1.VirtualMachineImage) (*ha
 
 func (h *vmImageHandler) createBackingImageAndStorageClass(image *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
 	if err := h.createBackingImage(image); err != nil && !errors.IsAlreadyExists(err) {
-		return nil, err
+		logrus.Errorf("failed to create backing image in image %s/%s, err: %v", image.Namespace, image.Name, err)
```
FrankYang0529:
```suggestion
		logrus.WithError(err).WithFields(logrus.Fields{
			"namespace": image.Namespace,
			"name":      image.Name,
		}).Error("failed to create backing image for vmimage")
```

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -142,10 +149,14 @@ func (h *vmImageHandler) checkImage(image *harvesterv1.VirtualMachineImage) (*ha
...
+		toUpdate := handleFail(image.DeepCopy(), condition.Cond(harvesterv1.ImageInitialized), err)
+		return toUpdate, err
 	}
 	if err := h.createStorageClass(image); err != nil && !errors.IsAlreadyExists(err) {
-		return nil, err
+		logrus.Errorf("failed to create storage class in image %s/%s, err: %v", image.Namespace, image.Name, err)
```
FrankYang0529:
```suggestion
		logrus.WithError(err).WithFields(logrus.Fields{
			"namespace": image.Namespace,
			"name":      image.Name,
		}).Error("failed to create storage class for vmimage")
```

FrankYang0529 (commented):
Overall LGTM. Leave some minor changes.

pkg/controller/master/image/vm_image_controller.go
brandboat: Thanks for the comment ! We can do more as PoAn commented in https://github.com/harvester/harvester/pull/5498#discussion_r1560453534

pkg/controller/master/image/vm_image_controller.go
brandboat: Actually the err nil check here is redundant, I've updated the code in https://github.com/harvester/harvester/pull/5498/commits/8adb4be68f1e06dd65b012db3e3dca773138ec80, many thanks

pkg/controller/master/image/vm_image_controller.go
brandboat: Sure, addressed in https://github.com/harvester/harvester/pull/5498/commits/16bb45b048d75491a6006fa4e729278139069a70

pkg/controller/master/image/vm_image_controller.go
brandboat: Thanks, I've changed all error log format in vm_image_controller to what you have suggested in https://github.com/harvester/harvester/pull/5498/commits/3156aa6a01658b01bd673be6e429e2fce9733d4b

FrankYang0529 (approved):
All test cases can pass. Thanks.

starbops (approved):
LGTM, thank you!

bk201:
@brandboat Can you provide another test plan purely in the image path? Some changes have been made in the vm image controller, I'd prefer if there is also a test plan in the image path only.

brandboat:
> @brandboat Can you provide another test plan purely in the image path? Some changes have been made in the vm image controller, I'd prefer if there is also a test plan in the image path only.

@bk201, many thanks ! I've added the vm image test plan in pr description.

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -105,63 +113,77 @@ func (h *vmImageHandler) initialize(image *harvesterv1.VirtualMachineImage) (*ha
...
+	toUpdate, err := h.checkImage(toUpdate)
 	if err != nil {
-		logrus.Errorf("failed to check image %s/%s, err: %v", image.Namespace, image.Name, err)
-		h.imageController.EnqueueAfter(image.Namespace, image.Name, 10*time.Second)
-		return h.images.Update(image)
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"namespace": toUpdate.Namespace,
+			"name":      toUpdate.Name,
+		}).Error("failed to check vmimage")
+		h.imageController.EnqueueAfter(toUpdate.Namespace, toUpdate.Name, 10*time.Second)
```
w13915984028: Both the `EnqueueAfter` and follow `Update` will trigger `OnChange` again (suppose the `Update` really updates the object), which one is preferred?

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -105,63 +113,77 @@ func (h *vmImageHandler) initialize(image *harvesterv1.VirtualMachineImage) (*ha
...
+		return h.images.Update(toUpdate)
 	}
-	return h.createBackingImageAndStorageClass(image)
+
+	toUpdate = h.createBackingImageAndStorageClass(toUpdate)
```
w13915984028: In `initialize`, after the call to `deleteBackingImageAndStorageClass`, k8s will set the deletionTimestamp and it needs some time to finally delete the objects;  then the `createBackingImageAndStorageClass` is called.  

In several retries, will the backing image & storage classes objects use the same name or different generated-name?  if same name is used, could the create operaion hit the `isAlreadyExisting` but it is the to-be-deleted object ? (not sure)

w13915984028 (commented):
2 questions, thanks.

pkg/controller/master/image/vm_image_controller.go
brandboat: Thanks for the comment ! Like you said, currently the onChange trigger twice in the above code snippet. ~~One appropriate way to deal with this is check if checkImage works successfully, if not, we should invoke EnqueueAfter or just return err. Anyway, we should avoid trigger onChange twice here.  I'll address this later. :smiley:~~ (Updated: ) We should remove the EnqueueAfter here.

pkg/controller/master/image/vm_image_controller.go
brandboat: That's a good point ! Yes that might happen, I've discussed this with @FrankYang0529, we need to check if storageClass and backingImage deletionTimestamp exist. If exists, invoke EnqueueAfter to wait all resources cleaup.

pkg/webhook/resources/version/validator.go
```diff
@@ -55,5 +60,25 @@ func (v *versionValidator) checkAnnotations(version *v1beta1.Version) error {
...
+		return err
+	}
+	return nil
+}
+
+func checkISOChecksum(version *v1beta1.Version) error {
+	isoChecksum := version.Spec.ISOChecksum
+	// if an isoChecksum is provided, it must be in the SHA-512 format
+	// since Longhorn backing images only accept hashes in that format
+	if isoChecksum != "" && !SHA512Pattern.MatchString(isoChecksum) {
```
w13915984028, line 80:
```
There is SHA256, SHA384 and SHA512. There's no SHA1024 yet.
```

Someone mentioned in another issue, he only got SHA256 and needed to compute a SHA512 manually.

If we force it to be SHA512 format, we need to document it.

pkg/webhook/resources/version/validator.go
```diff
@@ -55,5 +60,25 @@ func (v *versionValidator) checkAnnotations(version *v1beta1.Version) error {
...
+		return werror.NewBadRequest(fmt.Sprintf("invalid isoChecksum %s", isoChecksum))
```
w13915984028, line 81: The error message could hint that the ` the provided isoChecksum is not in the expected SHA512 format`

And other question:

If the `isoCheckSum` is in correct format, but the value is wrong, will we have the exact error message from LH to show to user?

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -60,31 +60,38 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
...
-	if needRetry {
-		if image.Status.LastFailedTime == "" {
-			return h.initialize(image)
+	for _, status := range bi.Status.DiskFileStatusMap {
+		if status.State == lhv1beta2.BackingImageStateFailed {
+			return h.handleRetry(image)
 		}
+	}
 
+	_, scErr := util.GetStorageClass(h.storageClassCache, image)
```
w13915984028: L75 checks `bi.DeletionTimestamp != nil`, should  L97 do same?

w13915984028 (commented):
Few more questions, thanks.

pkg/webhook/resources/version/validator.go
brandboat: Sure, I'll file another issue about the documentation.

pkg/webhook/resources/version/validator.go
brandboat: Yes, we'll have something like `Failed to process sync file: the expected checksum ... doesn't match the file actual checksum ...` in both upgrade and image download page.

pkg/controller/master/image/vm_image_controller.go
brandboat: I thought you mean L87,  you are right, we need to do the same thing to storage class.

brandboat: I've addressed the comment in 8cc0ea0, please take another look. Thanks ! :smiley:

w13915984028 (approved):
LGTM, thanks. BTW, please squash those commits before final merging.

pkg/webhook/resources/version/validator.go
brandboat: gentle ping @w13915984028 , I've checked the documentation and found that we already mentioned that version CR should use SHA-512 checksum in https://docs.harvesterhci.io/v1.4/upgrade/index#prepare-an-air-gapped-upgrade.
```
    isoChecksum: <SHA-512 checksum of the ISO>
```

bk201:
@brandboat Due to the revert https://github.com/harvester/harvester/commit/18973f46329dd06587c3d98f0abae126e7bdb1fd#diff-271c0b1384d0424ab45c719482f3c867166e7c48319fee9754bf8938f56f67ba
The PR can't build correctly due to some missing fields, can you help rebase again, thanks!
you can check the output here: https://drone-pr.rancher.io/harvester/harvester/4726/1/2

brandboat:
> @brandboat Due to the revert https://github.com/harvester/harvester/commit/18973f46329dd06587c3d98f0abae126e7bdb1fd#diff-271c0b1384d0424ab45c719482f3c867166e7c48319fee9754bf8938f56f67ba
The PR can't build correctly due to some missing fields, can you help rebase again, thanks!
you can check the output here: https://drone-pr.rancher.io/harvester/harvester/4726/1/2

Thanks for the reminder, already rebased to the latest commit in master branch :smiley:

starbops (approved):

bk201:
@brandboat I'm trying to create a virtual machine image with the change, but it seems the field `status.storageClassName` is not set within an image. GUI and the Harvester Terraform provider are now relying on this field to get the correct storage class name.

FrankYang0529:
> @brandboat I'm trying to create a virtual machine image with the change, but it seems the field `status.storageClassName` is not set within an image. GUI and the Harvester Terraform provider are now relying on this field to get the correct storage class name.

I can reproduce the error. @brandboat, we have to call `h.images.Update(toUpdate)` if it's happy path. Currently, we only call `Update` when there is error.

brandboat:
Already fixed the errors... Thanks you @bk201 , @FrankYang0529 ! :smiley:

FrankYang0529 (approved):
LGTM. Thank you.

bk201 (approved):
LGTM

bk201:
@mergifyio backport v1.2 v1.3

Issue #5510, opened 2024-04-04
[BUG] VirtualMachineImageSourceType is an enum but implemented as string


m-ildefons:
**Describe the bug**

In the `VirtualMachineImageSpec` type definition, the field `SourceType` is of type string, but its behavior is that of an enum.

```
[...]
type VirtualMachineImageSpec struct {
  [...]
  // +kubebuilder:validation:Required
  // +kubebuilder:validation:Enum=download;upload;export-from-volume
  SourceType string `json:"sourceType"
  [...]
}
[...]
```

To correctly reflect the enum behavior, it should be a type alias. Additionally, it should not be relied on type inference for the constants defining the possible values of the enum, but the typing should be explicitly set.

```
[...]
type VirtualMachineImageSpec struct {
  [...]
  // +kubebuilder:validation:Required
  // +kubebuilder:validation:Enum=download;upload;export-from-volume
  SourceType VirtualMachineImageSourceType `json:"sourceType"
  [...]
}

// +enum
type VirtualMachineImageSourceType string

const (
  VirtualMachineImageSourceTypeDownload     VirtualMachineImageSourceType = "download"
  VirtualMachineImageSourceTypeUpload       VirtualMachineImageSourceType = "upload"
  VirtualMachineImageSourceTypeExportVolume VirtualMachineImageSourceType = "export-from-volume"
)
[...]
```

**Additional context**
The current typing situation is loose and relies solely on the webhooks to validate the inputs. This is fragile, and makes it impossible to utilize the typing system of the golang compiler to its full extend. Additionally, the generated openapi spec is wrong, since it doesn't contain the possible enum values, but rather just specifies the type as string. This in turn results in incomplete documentation.

https://github.com/kubernetes-sigs/kubebuilder/issues/1071

m-ildefons:
I've updated the swagger.json file for the v1.3 API docs [here](https://github.com/harvester/docs/pull/574) once that is merged, I think everything related to this issue is in and this can be closed.

m-ildefons:
:green_circle:  fixed on master branch
:green_circle: backported to v1.3 branch
:green_circle: documentation updated
This is done.

PR #5511, opened 2024-04-04
api: make VirtualMachineImageSourceType an enum

m-ildefons:
**Problem:**
VirtualMachineImageSourceType is implemented as string, but it actually is an enum. This results in inaccurate API specs and incomplete documentation

**Solution:**
Implement VirtualMachineImageSourceType as enum analogous to other enums in K8s APIs.

**Related Issue:**

- https://github.com/harvester/harvester/issues/5510

**Test plan:**
No behavior changes expected.

Api documentation for VirtualMachineImage should name the possible values for the enum type.

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -9556,7 +9556,12 @@
           },
           "sourceType": {
             "type": "string",
-            "default": ""
+            "default": "",
+            "enum": [
+              "download",
+              "export-from-volume",
+              "upload"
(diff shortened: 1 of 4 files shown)


connorkuehl (approved):

bk201 (approved):
LGTM, @WebberHuang1118 please take a took too!

WebberHuang1118 (approved):
LGTM, thanks.

Issue #4388, opened 2023-08-07
[BUG] API will stuck if delete the VM not existing


lanfon72:
### Describe the bug
API endpoint: `v1/harvester/kubevirt.io.virtualmachines/{namespace}/{vm_name}`

### To Reproduce
1. Install Harvester with any nodes
2. Call API endpoint to delete VM

### Expected behavior
API should return 404 error

## Environment:
 - Harvester ISO version: **v1.2.0-rc5**
 - **ui-source** Option: **Auto**
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 3 nodes**

### Additional context
API endpoint: `v1/harvester/persistentvolumeclaims/{namespace}/{vm_name}` will return 404 error.

guangbochen:
Is this urgent for e2e testing? If not will prefer to leave it for v1.2.1, thanks.

lanfon72:
It's not urgent, we can wait this to be fixed after `v1.2.0`.

bk201:
@Yu-Jack can you help check this, deleting a non-existing VM blocks the API call.

Yu-Jack:
I tried API endpoint `v1/harvester/kubevirt.io.virtualmachines/{namespace}/{vm_name}`, but I didn't get stuck. It returns http status code 500 error to me with following error message in v1.2.1 and v1.3 these two harvester versions.

```
{
    "type": "error",
    "links": {},
    "code": "ServerError",
    "message": "Failed to get vm default/testa, virtualmachines.kubevirt.io \"testa\" not found",
    "status": 500,
    "type": "error"
}
```

@lanfon72  Did I miss some steps to reproduce that? Or it just happened on RC version?

Here is my example API call:
```
# testa vm doesn't exist
curl --location --request DELETE 'https://192.168.122.61/v1/harvester/kubevirt.io.virtualmachines/default/testa' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'Authorization: Bearer token-xxxx'
```

lanfon72:
The result from Postman
![image](https://github.com/harvester/harvester/assets/5169694/974efecb-7acd-4695-ad32-b7df97328176)

curl (http2)
![image](https://github.com/harvester/harvester/assets/5169694/a7c91250-9a56-4e0a-a8a7-a499242f148f)

curl (http1.1)
[...]

PR #5532, opened 2024-04-10
fix: should return not found

Yu-Jack:

**Problem:**
Although we use steven to proxy all k8s resource, but we can still customize some logic with store of schema. In our custom logic, we didn't check not found error, we just returned any error with 500 error code.

**Solution:**
Return not found if vm doesn't exist.

**Related Issue:**
https://github.com/harvester/harvester/issues/4388

**Test plan:**
Delete a non-existing vm with this endpoint `/v1/harvester/kubevirt.io.virtualmachines/{namespace}/{name}`, it will return 404 not found http status code back.

image

### DIFF
--- a/pkg/api/vm/store.go
+++ b/pkg/api/vm/store.go
@@ -9,6 +9,7 @@ import (
 	v1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/schemas/validation"
 	"github.com/rancher/wrangler/pkg/slice"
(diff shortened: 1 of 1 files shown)


Yu-Jack:
@bk201  I think this one should be backported to v1.2 and v1.3. Let automation test work on those versions.

FrankYang0529 (approved):
LGTM. Thank you.

brandboat (approved):
LGTM ! :smiley:

Yu-Jack:
@Mergifyio backport v1.2 v1.3

Issue #5494, opened 2024-04-01
[BUG] User can do upgrade to old version


lanfon72:
### Describe the bug
The upgrade will started and fail if user use an older ISO to do the upgrade.
![image](https://github.com/harvester/harvester/assets/5169694/6d50c0fa-3e5c-40f5-a902-58a77c35fb2c)
![image](https://github.com/harvester/harvester/assets/5169694/37123a45-7bb9-4640-bb50-4e220900240f)

### To Reproduce
1. Install Harvester with any nodes
2. Create upgrade version
3. Perform upgrade

### Expected behavior
Upgrade should not start

## Environment:
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360G9 3 nodes**
 - Harvester ISO version: **v1.3.0**
 - **ui-source** Option: **Auto**

### Additional context
discovered when implementing test automation of #2431

starbops:
As we're targeting the v1.4 release, the test plan would be...

## Test Plan

1. Prepare a single-node cluster of version v1.4-head (should've contained this fix)
2. Fake the version as it's `v1.4.0` by adding `HARVESTER_SERVER_VERSION` env to the `harvester` deployment
   ```
         ...
         containers:
         - env:
           - name: HARVESTER_SERVER_VERSION
             value: v1.4.0
             ...
   ```
3. Upgrade to v1.3.2
4. The upgrade should be blocked because it's effectively a downgrade behavior

lanfon72:
Tested this bug not be fixed yet, move back to implement.

![image](https://github.com/user-attachments/assets/aa822782-366b-4d6a-9d49-dffb363e1160)
![image](https://github.com/user-attachments/assets/aaecc85b-cd5e-4745-89e5-d6c8a5a33272)


Test Information
----
* Environment: **qemu/KVM 2 nodes and single node**
* Harvester Version: **v1.4-97e60e1-head** (patch to `v1.4.0` by editing *harvester* deployment)
* **ui-source** Option: **Auto**

### Verify Steps
1. Follow **Steps to reproduce** in https://github.com/harvester/harvester/issues/5494#issue-2217383003
    - In `v1.4-head`, we would need to patch *harvester* deployment to `v1.4.0` as https://github.com/harvester/harvester/issues/5494#issuecomment-2375889345 mentioned.

starbops:
[...]

Issue #2877, opened 2022-10-05
[ENHANCEMENT] Upgrade path: tool to do semantic version comparison


bk201:
**Is your feature request related to a problem? Please describe.**

In the upgrade path, we need to compare semantic versions in many places. 
Neither the bash `<` operator nor `sort` command performs well.

**Describe the solution you'd like**

We need a handy tool to be called in the upgrade scripts.

**Describe alternatives you've considered**

Build our tool, leverage library like https://github.com/Masterminds/semver.

**Additional context**

starbops:
## The Ultimate Policy of Harvester Upgrade

### Glossary

- Formal release versions: a semantic version with no suffices, e.g. v1.2.2 and v1.3.0
- Prerelease versions: a semantic version with a suffix, e.g. v1.2.2-rc1
- Dev versions: any version that falls outside of the two above, e.g. v1.2-ab12cd34
- Minimum upgrade requirements: each formal release/prerelease version will incorporate an upgrade matrix file specifying the minimal upgradable version for such a version

### Rules

Since there are various upgrade paths, we explicitly enforce a set of rules to prohibit specific upgrade paths to ensure the stability of Harvester upgrades and reduce the maintenance burden for the Harvester team:

- Users are not allowed to upgrade from any version to a lower version (effectively downgrade)
- Users are not allowed to upgrade from any formal release version lower than the minimal upgrade requirement of the targeting formal release version
- Users are not allowed to upgrade from any dev version to a formal release version[^1]
- Users are not allowed to upgrade from any dev version to a prerelease version[^1]
- Users are allowed to upgrade from any lower prerelease version to a higher one within the same release version but not across different release versions

Upgrade paths other than the ones stated above are all legal. As a side note, we listed several peculiar upgrade paths that seem illegal but are actually doable below:

- Same-version upgrades are always allowed, no matter what the version is
- Upgrade from a formal release or prerelease version to a dev version is always allowed
- Upgrade from any dev version to another dev version is always allowed
- Upgrade from any prerelease version to a corresponding formal release version or another newer prerelease version within the same release version, e.g. v1.2.2-rc1 to v1.2.2-rc2

[^1]: By default, the strict mode is enabled to prevent any informal upgrades from happening. It could be disabled to loosen such a restriction.

Please feel free to add comments if the above rules are incomplete or need improvement. Thank you. cc @bk201  @khushboo-rancher

brandboat:
Test ok with v1.4.0-rc4, will close the issue

[...]

PR #5616, opened 2024-04-17
Refactoring the eligibility check regarding versions for Harvester upgrades

starbops:

**Problem:**

We have implemented the upgrade eligibility checks in several places. Fixing bugs or adding new checks causes headaches because we need to manage all of them. There should be a common Go library doing version-comparison stuff specific to Harvester extent that we can rely on (harvester/go-common#12). With that in place, we can update the controller's code and the upgrade scripts to adapt to the newly introduced library.

**Solution:**

Leveraging the `version` package from `go-common`, adapt the upgrade controller code and add a new subcommand in `upgrade-helper` for the upgrade script.

**Related Issue:**

#2877 #5494

~~Note: It's better to review this PR after harvester/go-common#12 is merged.~~
Update: harvester/go-common#12 was merged, and the PR has been updated with the latest dependency

**Test plan:**

Upgrade from the version built with this PR to v1.3.0
SCR-20240417-poxb

Upgrade from the version built with this PR to v1.3.0 (but pretend to be v1.2.1, which doesn't meet the minimum upgradable version of v1.3.0)
SCR-20240417-pqou

Upgrade from the version built with this PR to v1.3.0 (but pretend to be v1.2.2, which **DOES** meet the minimum upgradable version of v1.3.0)
SCR-20240417-psgk


brandboat:
gentle ping @starbops, there are conflicts here.

cmd/upgradehelper/cmd/root.go
```diff
@@ -35,9 +38,9 @@ func init() {
...
 
-	rootCmd.PersistentFlags().BoolVar(&logDebug, "debug", debug, "set logging level to debug")
-	rootCmd.PersistentFlags().BoolVar(&logTrace, "trace", trace, "set logging level to trace")
+	RootCmd.PersistentFlags().BoolVar(&logDebug, "debug", debug, "set logging level to debug")
+	RootCmd.PersistentFlags().BoolVar(&logTrace, "trace", trace, "set logging level to trace")
 
-	rootCmd.Flags().StringVar(&kubeConfigPath, "kubeconfig", os.Getenv("KUBECONFIG"), "Path to the kubeconfig file")
-	rootCmd.Flags().StringVar(&kubeContext, "kubecontext", os.Getenv("KUBECONTEXT"), "Context name")
+	RootCmd.Flags().StringVar(&KubeConfigPath, "kubeconfig", os.Getenv("KUBECONFIG"), "Path to the kubeconfig file")
+	RootCmd.Flags().StringVar(&KubeContext, "kubecontext", os.Getenv("KUBECONTEXT"), "Context name")
```
brandboat: nit: I'm a little bit confused about these two flags. There is no op when executing `./upgrade-helper`, perhaps we could remove those two flags ? Or move them into subcommands ?

pkg/upgradehelper/versionguard/version.go
```diff
@@ -0,0 +1,56 @@
...
+		if err != nil {
+			return err
+		}
+	} else {
+		repoInfo, err := getRepoInfo(upgrade)
+		if err != nil {
+			logrus.WithFields(logrus.Fields{
+				"namespace": upgrade.Namespace,
+				"name":      upgrade.Name,
+			}).Warn("failed to retrieve repo info")
```
brandboat: perhaps Error ?
```suggestion
			}).Error("failed to retrieve repo info")
```

pkg/upgradehelper/versionguard/version.go
```diff
@@ -0,0 +1,56 @@
...
+			return err
+		}
+
+		minUpgradableVersion, err = version.NewHarvesterVersion(repoInfo.Release.MinUpgradableVersion)
+		if err != nil && !errors.Is(version.ErrInvalidVersion, err) {
```
brandboat: Just a note, `ErrInvalidVersion` is ok here since this only happen when version is nil (or you can say empty string), and in HarvesterUpgradeVersion#IsUpgradable, nil minUpgradableVersion means there's no restriction.

cmd/upgradehelper/cmd/versionguard/versionguard.go
```diff
@@ -0,0 +1,125 @@
...
+
+	strictMode           bool
+	minUpgradableVersion string
+	upgradeName          string
+
+	harvClient *harv1type.HarvesterhciV1beta1Client
+}
+
+func init() {
+	versionGuardCmd.Flags().BoolVar(&strictMode, "strict", true, "The strict mode controls whether dev versions can be upgraded. If it is `true`, upgrading from dev versions is prohibited. Default to `true`.")
```
brandboat: Actually we can upgrade to dev versions no matter strict mode is true or false. Perhaps we can add this exception to the description.

brandboat (approved):
LGTM ! This look really gooood ! Just left some nits.

brandboat:
BTW, could you rebase the branch to the laster commit in master branch, we have bunch of upgrades these days.

pkg/upgradehelper/versionguard/version.go
starbops: I added some comments to explain the intention. Thanks for the heads-up.

pkg/upgradehelper/versionguard/version.go
starbops: Updated. Thanks.

cmd/upgradehelper/cmd/versionguard/versionguard.go
starbops: Updated the flag description. Thank you.

cmd/upgradehelper/cmd/root.go
starbops: They should've been the persistent flags. Thanks for pointing out.

```
./upgrade-helper --kubeconfig=/path/to/kubeconfig --kubecontext=test-context version-guard hvst-upgrade-xxxxx
```

The two subcommands will leverage them for the location of the desired kubeconfig file and the context (optional). We don't explicitly use the two flags because we inject the value via environment variables. They're convenient for us when doing the development and testing.

bk201 (approved):
LGTM, thanks!

Issue #5690, opened 2024-04-26
[BUG] upgrade stuck in waiting plan restart-rancher-system-agent to complete


lanfon72:
### Describe the bug
Upgrade pending on **Upgradeing System Service**, when check the log of pod _apply-manifests_, it shows
> Waiting for plan hvst-upgrade-fm29b-skip-restart-rancher-system-agent to complete...

### To Reproduce
1. Install Harvester with any nodes
2. Create an image for VM creation
3. Create Cluster network and VM network
4. Setup backup-target
5. Create new storageclass `sc1` with 3 replicas and set it as default
6. Create VM `vm1` with VM network's VLAN
7. Write data into `vm1` then take backup `vm1b`
8. Restore `vm1b` into `vm2`
9. Create VM `vm3` with additional volume
10. Perform upgrade

### Expected behavior
upgrade should success

## Environment:
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **Baremetal DL360 3 nodes**
 - Harvester ISO version: **v1.2.1** -> **v1.2.2-rc1**
 - **ui-source** Option: **Auto**

### Additional context
The upgrade is performed by test automation (locally)

### Support bundle
[supportbundle_229d2fc0-5c52-42e2-a10e-9cd6a437adb1_2024-04-26T07-43-07Z.zip](https://github.com/harvester/harvester/files/15127232/supportbundle_229d2fc0-5c52-42e2-a10e-9cd6a437adb1_2024-04-26T07-43-07Z.zip)

albinsun:
Hit in a 3 nodes upgrade test from `v1.2.1` -> `v1.2.2-rc2`.
Will test again after RC3 release.
![image](https://github.com/harvester/harvester/assets/2773781/e11d8b8f-729d-4e85-a4d0-4376c898b9a7)

w13915984028:
The `plan` object `hvst-upgrade-fm29b-skip-restart-rancher-system-agent` has no `status` field from the support-bundle; but other `plan` objects have.


```
- apiVersion: upgrade.cattle.io/v1
  kind: Plan
  metadata:
    creationTimestamp: "2024-04-25T20:55:07Z"
    generation: 1
    managedFields:
    - apiVersion: upgrade.cattle.io/v1
      fieldsType: FieldsV1
      fieldsV1:
        f:spec:
          .: {}
          f:concurrency: {}
          f:nodeSelector:
            .: {}
            f:matchLabels:
              .: {}
[...]

PR #5698, opened 2024-04-29
check timestamps before waiting for agent rollout

ibrokethecloud:

**Problem:**

PR attempts to address random upgrade fails where the upgrade fails due to the following error:

```
Wait for cluster settling down...
CAPI cluster fleet-local/local is provisioned (current generation: 838).
cluster.fleet.cattle.io/local patched
Error from server (NotFound): deployments.apps "fleet-agent" not found
```

This is caused by a delay in rollout of fleet-agent, where the rollout happens when the check for fleet-agent rollout status is running, and deployment gets recreated.

**Solution:**

The fix attempts to use original fleet-agent creation timestamp before fleet cluster object is patched, and ensures new fleet-agent deployment has a newer creation timestamp before rollout status is checked.

**Related Issue:**

**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -281,8 +281,10 @@ upgrade_rancher() {
   wait_rollout cattle-system deployment rancher-webhook
   echo "Wait for cluster settling down..."
   wait_capi_cluster fleet-local local $pre_generation
-  patch_fleet_cluster
   wait_for_deployment cattle-fleet-local-system fleet-agent
+  pre_patch_timestamp=$(fleet_agent_timestamp)
+  patch_fleet_cluster
+  wait_for_fleet_agent $pre_patch_timestamp
   wait_rollout cattle-fleet-local-system deployment fleet-agent
 }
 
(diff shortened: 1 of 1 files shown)


ibrokethecloud:
should address: https://github.com/harvester/harvester/issues/5690

bk201:
@starbops The fix avoid the failure of wait_rollout and https://github.com/harvester/harvester/blob/8958ebee30387c8591eeb17ddf60e178517e9de3/package/upgrade/upgrade_manifests.sh#L1310C1-L1311C1 will be called.

package/upgrade/upgrade_manifests.sh
```diff
@@ -1300,6 +1302,23 @@ wait_for_deployment() {
   done
 }
 
+fleet_agent_timestamp(){
+  wait_for_deployment cattle-fleet-local-system fleet-agent
```
starbops: Maybe we need to suppress the output of calling this `wait_for_deployment` function. Otherwise, the output log lines and the timestamp integer will be saved together in the variables used later for an integer comparison `-ge`.

```
/usr/local/bin/upgrade_manifests.sh: line 1314: [: waiting: integer expression expected
```

starbops (approved):
The upgrade went well on a 2-node cluster. LGTM, thank you.

FrankYang0529 (approved):

bk201 (approved):
LGTM, thanks!

bk201:
@mergifyio backport v1.2 v1.3

bk201:
@mergifyio backport v1.2 v1.3

Issue #1078, opened 2021-08-06
[FEATURE] Migrate harvester api actions to standardized k8s mechanism


guangbochen:
**Is your feature request related to a problem? Please describe.**
We will need to migrate the existing harvester API actions to the standardized k8s api mechanism, so users who deal with the vanilla k8s API can also use those customize API actions.  

**Describe the solution you'd like**
e.g using the k8s API subresources, but will need further investigation.

gitlawr:
Actions being used:
VM:
- start
- stop
- restart
- ejectCDRom
- pause
- unpause
- migrate
- abortmigration
- backup
- restore
- createTemplate

Node:
- enableMaintenanceMode
- disableMaintenanceMode
- cordon
- uncordon

Yu-Jack:
@bk201  Just spent some time checking this. To be more clear, we should use `apiservice` to replace. Let me use kubevirt as example to explain this.

Let's say, if we try to start or stop the VM, we could see there are requests printed in virt-api.
```
{"component":"virt-api","contentLength":0,"level":"info","method":"PUT","pos":"filter.go:47","proto":"HTTP/1.1","remoteAddress":"192.168.122.75","statusCode":202,"timestamp":"2024-04-26T06:32:28.770904Z","url":"/apis/subresources.kubevirt.io/v1/namespaces/default/virtualmachines/test/start","username":"-"}
{"component":"virt-api","contentLength":0,"level":"info","method":"PUT","pos":"filter.go:47","proto":"HTTP/1.1","remoteAddress":"192.168.122.75","statusCode":202,"timestamp":"2024-04-26T06:40:44.729912Z","url":"/apis/subresources.kubevirt.io/v1/namespaces/default/virtualmachines/test/stop","username":"-"}
```

The URL is `/apis/subresources.kubevirt.io/v1/namespaces/default/virtualmachines/test/start` or `/apis/subresources.kubevirt.io/v1/namespaces/default/virtualmachines/test/stop`, then KubeVirt will [handle these requests by different routes](https://github.com/kubevirt/kubevirt/blob/ca94b81c667ab4cee70e7c009a5a67e2386d2c1d/pkg/virt-api/api.go#L249-L271).

This uses Aggregated API to do that. In order to use that, we need to define a new resource called `apiservice`, here is an example KubeVirt defined. 

```yaml
apiVersion: apiregistration.k8s.io/v1
kind: APIService
metadata:
  labels:
    app.kubernetes.io/component: kubevirt
[...]

PR #5704, opened 2024-04-29
feat: migrate harvester action API to subresource

Yu-Jack:

**Problem:**
Currently, there are fews APIs don't follow kubernetes rules. In order to make use of kubernetes API format, we should migrate those APIs.

**Solution:**
HEP https://github.com/harvester/harvester/pull/5885

We could use Aggregated API Server (API Service) to solve this issue. Need to define a `subresources.harvesterhci.io` to proxy request to our harvester server. It's similar with [kubevirt subresource](https://kubevirt.io/api-reference/v1.2.0/operations.html#_v1start).

Create another common interface to deal with two different sources, this is a simple visual flow of objects.

image

**Related Issue:**
https://github.com/harvester/harvester/issues/1078

**Test plan:**

Before testing, should check there is a harvester apiservice:

```yaml
apiVersion: apiregistration.k8s.io/v1
kind: APIService
metadata:
  name: v1beta1.subresources.harvesterhci.io
spec:
  group: subresources.harvesterhci.io
  groupPriorityMinimum: 1000
  service:
    name: harvester
    namespace: harvester-system
    port: 8443
  version: v1beta1
  versionPriority: 15
  insecureSkipTLSVerify: true
```

Before testing, these are example action API lists we currently use:
- GET `/v1/harvester/harvesterhci.io.virtualmachineimages/default/image-g88lp/download`
- POST `/v1/harvester/kubevirt.io.virtualmachines/default/test?action=stop`
- POST `/v1/harvester/nodes/jacknode?action=cordon`
- GET `/v1/harvester/upgradeslogs/default/test/download`
- POST `/v1/harvester/persistentvolumeclaims/default/test-disk-0-rlnlk?action=snapshot`

We should test those APIs by using subresource APIs, for example:

- GET `/apis/subresources.harvesterhci.io/v1beta1/namespaces/default/virtualmachineimages/ubuntu-focal/download`
- POST `/apis/subresources.harvesterhci.io/v1beta1/namespaces/default/virtualmachines/test/stop`
- POST `/apis/subresources.harvesterhci.io/v1beta1/nodes/jacknode/uncordon`
- GET `/apis/subresources.harvesterhci.io/v1beta1/namespaces/default/upgradelogs/test/download`
- POST `/apis/subresources.harvesterhci.io/v1beta1/namespaces/default/persistentvolumeclaims/test/snapshot`
- `/apis/subresources.harvesterhci.io/v1beta1/namespaces/default/snapshot.storage.k8s.io.volumesnapshots/test/restore`

Then you could also use following REST config to invoke subresource:

```go
harvesterCopyConfig := rest.CopyConfig(server.RESTConfig)
harvesterCopyConfig.GroupVersion = &k8sschema.GroupVersion{Group: "subresources.harvesterhci.io", Version: "v1beta1"}
harvesterCopyConfig.APIPath = "/apis"
harvesterCopyConfig.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
harvesterSubresourceClient, err := rest.RESTClientFor(harvesterCopyConfig)

h.harvesterSubresourceClient.Post().Resource("nodes").SubResource("uncordon").Name(node.Name).Do(context.Background())
h.harvesterSubresourceClient.Post().Namespace("default").Resource("virtualmachines").SubResource("stop").Name("test").Do(context.Background())
```

When testing with wrong resource/name/subresource, will get different status code and message like:

- 422 - `InvalidAction 422: Unsupported virtualmachines/stopa`
- 422 - `InvalidAction 422: Unsupported virtualmachinessss/stop`
- 404 - `NotFound 404: Virtual machine default/testa not found`

Currently, I use rancher/apiserver to generate apierror like we did in action part. But, I think it's still discussable which ways we'd like to respond.

**Actions should migrate**

- [x] image
- [x] node
- [x] upgradelog
- [x] vm
- [x] volume
- [x] volumesnapshot

### DIFF
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -105,14 +105,17 @@ func (h Handler) do(rw http.ResponseWriter, req *http.Request) error {
 	}
 
 	// Need to convert the link and action to subresource
-	if req.Method == http.MethodGet {
+	switch req.Method {
+	case http.MethodGet:
 		resource.SubResource = vars["link"]
-	} else if req.Method == http.MethodPost {
+	case http.MethodPost:
 		resource.SubResource = vars["action"]
+	default:
+		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported method %s", req.Method))
 	}
 
 	if !h.IsMatchedResource(resource, req.Method) {
-		return apierror.NewAPIError(validation.InvalidAction, "Invalid resource handler")
+		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported %s action %s", req.Method, resource.SubResource))
 	}
 
 	return h.SubResourceHandler(rw, req, resource)
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -29,7 +29,6 @@ import (
 const (
 	actionUpload   = "upload"
 	actionDownload = "download"
-	imagesResource = "virtualmachineimages"
 )
 
 func Formatter(request *types.APIRequest, resource *types.RawResource) {
@@ -56,7 +55,7 @@ type Handler struct {
 }
 
 func (h Handler) IsMatchedResource(resource subresource.Resource, method string) bool {
-	if resource.Name != imagesResource {
+	if resource.Name != subresource.VirtualMachineImages.Resource {
 		return false
 	}
 
@@ -99,7 +98,7 @@ func (h Handler) do(rw http.ResponseWriter, req *http.Request) error {
 	vars := util.EncodeVars(mux.Vars(req))
 
 	resource := subresource.Resource{
-		Name:       imagesResource,
+		Name:       subresource.VirtualMachineImages.Resource,
 		ObjectName: vars["name"],
 		Namespace:  vars["namespace"],
 	}
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -118,24 +118,6 @@ func (h Handler) do(rw http.ResponseWriter, req *http.Request) error {
 	return h.SubResourceHandler(rw, req, resource)
 }
 
-func (h Handler) doGet(link string, rw http.ResponseWriter, req *http.Request) error {
-	switch link {
-	case actionDownload:
-		return h.downloadImage(rw, req)
-	default:
-		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported GET action %s", link))
-	}
-}
-
-func (h Handler) doPost(action string, rw http.ResponseWriter, req *http.Request) error {
-	switch action {
-	case actionUpload:
-		return h.uploadImage(rw, req)
-	default:
-		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported POST action %s", action))
-	}
-}
-
 func (h Handler) downloadImage(rw http.ResponseWriter, req *http.Request) error {
 	vars := util.EncodeVars(mux.Vars(req))
 	namespace := vars["namespace"]
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -31,6 +31,13 @@ const (
 	actionDownload = "download"
 )
 
+var (
+	subResourceMethod = map[string]string{
+		actionUpload:   http.MethodPut,
+		actionDownload: http.MethodPut,
+	}
+)
+
 func Formatter(request *types.APIRequest, resource *types.RawResource) {
 	resource.Actions = make(map[string]string, 1)
 	if request.AccessControl.CanUpdate(request, resource.APIObject, resource.Schema) != nil {
@@ -54,17 +61,13 @@ type Handler struct {
 	BackingImageCache           ctllhv1.BackingImageCache
 }
 
-func (h Handler) IsMatchedResource(resource subresource.Resource, method string) bool {
+func (h Handler) IsMatchedResource(resource subresource.Resource, httpMethod string) bool {
 	if resource.Name != subresource.VirtualMachineImages.Resource {
 		return false
 	}
 
-	if resource.SubResource == actionDownload && method == http.MethodGet {
-		return true
-	}
-
-	if resource.SubResource == actionUpload && method == http.MethodPost {
-		return true
+	if method, ok := subResourceMethod[resource.SubResource]; ok {
+		return method == httpMethod
 	}
 
 	return false
@@ -113,11 +116,7 @@ func (h Handler) do(rw http.ResponseWriter, req *http.Request) error {
 		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported method %s", req.Method))
 	}
 
-	if !h.IsMatchedResource(resource, req.Method) {
-		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported %s action %s", req.Method, resource.SubResource))
-	}
-
-	return h.SubResourceHandler(rw, req, resource)
+	return subresource.Execute(h, rw, req, resource)
 }
 
 func (h Handler) downloadImage(rw http.ResponseWriter, req *http.Request) error {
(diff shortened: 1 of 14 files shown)


pkg/api/image/formatter.go
```diff
@@ -68,31 +97,28 @@ func (h Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
-}
 
-func (h Handler) doPost(action string, rw http.ResponseWriter, req *http.Request) error {
-	switch action {
-	case actionUpload:
-		return h.uploadImage(rw, req)
-	default:
-		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported POST action %s", action))
+	if !h.IsMatchedResource(resource, req.Method) {
+		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported %s action %s", req.Method, resource.SubResource))
```
Yu-Jack: We should keep original action and method error message for old APIs.

pkg/api/image/formatter.go
```diff
@@ -53,6 +55,33 @@ type Handler struct {
...
+}
+
+func (h Handler) SubResourceHandler(rw http.ResponseWriter, req *http.Request, resource subresource.Resource) error {
+	switch resource.SubResource {
+	case actionDownload:
+		return h.downloadImage(rw, req)
+	case actionUpload:
+		return h.uploadImage(rw, req)
+	default:
+		return apierror.NewAPIError(validation.InvalidAction, fmt.Sprintf("Unsupported subresource %s", resource.SubResource))
```
Yu-Jack, line 83: And this error message is for new subresources APIs.

pkg/server/subresource/subresource.go
```diff
@@ -0,0 +1,129 @@
...
+
+var (
+	apiService = schema.GroupVersion{Group: "subresources.harvesterhci.io", Version: "v1beta1"}
+
+	VirtualMachineImages   = schema.GroupVersionResource{Group: apiService.Group, Version: apiService.Version, Resource: "virtualmachineimages"}
+	UpgradeLogs            = schema.GroupVersionResource{Group: apiService.Group, Version: apiService.Version, Resource: "upgradelogs"}
+	Node                   = schema.GroupVersionResource{Group: apiService.Group, Version: apiService.Version, Resource: "nodes"}
+	VirtualMachines        = schema.GroupVersionResource{Group: apiService.Group, Version: apiService.Version, Resource: "virtualmachines"}
+	PersistentVolumeClaims = schema.GroupVersionResource{Group: apiService.Group, Version: apiService.Version, Resource: "persistentvolumeclaims"}
+	VolumeSnapshots        = schema.GroupVersionResource{Group: apiService.Group, Version: apiService.Version, Resource: "snapshot.storage.k8s.io.volumesnapshot"}
```
Yu-Jack, line 23: I also collected all resource which contains action here, it would be more clear.

Yu-Jack, line 23: I collected all resource which contains action or link here, it'd be more clear. Then, we should consider what resources name we should use.

pkg/server/subresource/subresource.go
```diff
@@ -0,0 +1,129 @@
...
+			Namespace:   vars["namespace"],
+		}
+	)
+
+	if vars["namespace"] != "" {
+		resource.Namespaced = true
+		resource.Namespace = vars["namespace"]
+	}
+
+	logrus.Debug("Resource: ", resource)
```
m-ildefons, line 107: I think it would be a good idea to log any subresource requests. This would help tremendously when dealing with support cases as the service logs would then contain information on what actions happened and when.

pkg/api/image/formatter.go
```diff
@@ -53,6 +54,33 @@ type Handler struct {
...
+func (h Handler) IsMatchedResource(resource subresource.Resource, method string) bool {
+	if resource.Name != subresource.VirtualMachineImages.Resource {
+		return false
+	}
+
+	if resource.SubResource == actionDownload && method == http.MethodGet {
+		return true
+	}
+
+	if resource.SubResource == actionUpload && method == http.MethodPost {
```
m-ildefons: Shouldn't this be HTTP method `PUT`, since the action is idempotent?

pkg/api/node/formatter.go
```diff
@@ -102,15 +113,31 @@ func (h ActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	rw.WriteHeader(http.StatusNoContent)
 }
 
-func (h ActionHandler) do(rw http.ResponseWriter, req *http.Request) error {
-	vars := util.EncodeVars(mux.Vars(req))
-	action := vars["action"]
-	name := vars["name"]
+func (h ActionHandler) IsMatchedResource(resource subresource.Resource, method string) bool {
+	if resource.Name != subresource.Node.Resource || method != http.MethodPost {
```
m-ildefons: Again, shouldn't this be a HTTP method `PUT` for idempotent actions (such as cordon or uncordon)?
Additionally, list-actions like `listUnhealthyVM` should be a `GET`, right?

pkg/api/upgradelog/handler.go
```diff
@@ -82,6 +83,33 @@ type Handler struct {
...
+		return false
+	}
+
+	if resource.SubResource == downloadArchiveLink && method == http.MethodGet {
+		return true
+	}
+
+	if resource.SubResource == generateArchiveAction && method == http.MethodPost {
+		return true
+	}
```
m-ildefons: Is there ever a use-case where we'd need the generate action but not the download action? If not, the generate action would be superfluous.

pkg/api/vm/handler.go
```diff
@@ -95,11 +117,24 @@ func (h vmActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	rw.WriteHeader(http.StatusNoContent)
 }
 
-func (h *vmActionHandler) doAction(rw http.ResponseWriter, r *http.Request) error {
-	vars := util.EncodeVars(mux.Vars(r))
-	action := vars["action"]
-	namespace := vars["namespace"]
-	name := vars["name"]
+func (h *vmActionHandler) IsMatchedResource(resource subresource.Resource, method string) bool {
+	if resource.Name != subresource.VirtualMachines.Resource || method != http.MethodPost {
```
m-ildefons: Again, shouldn't we have HTTP method `PUT` for idempotent actions and HTTP method `GET` for listing-actions like `findMigratableNodes`?

m-ildefons (commented):
Looks generally quite good. I still found a few places where I'm not sure about the usage of `POST` vs `PUT` and I think we should add a logging hook to make it easier to recognize actions when dealing with support cases.
I found the upgradelog API particularly confusing since I'm not sure why generating and downloading are two separate actions.

pkg/api/image/formatter.go
Yu-Jack: What I'm doing is to migrate whole same functionality to sub resource handler, and original method is `POST`.  I didn't make too much changes to original behavior just in case, including HTTP method.

https://github.com/harvester/harvester/blob/bf8bdcd7423508f767bf389e43ee8e812639e01a/pkg/api/image/formatter.go#L71-L75

And another reason why it's `POST` because we're using rancher/steve to build our API server. In original documentation, `action` query string seems to only allow `POST` action ([reference](https://github.com/rancher/steve?tab=readme-ov-file#action)).

However, I was thinking and testing about this during these few days. I also think `PUT` would be better for those cases. So, I will use `PUT` method to replace all `POST` method if behavior is idempotent. I think this change would make our API look like more k8s standardized API. 

In short, I'll still check all behaviors are idempotent or not, and make it to `PUT` or `GET` method instead of all `POST` method.

**NOTE: For those people are using our action APIs, they might be confused about changing method. But, I think it's a reasonable change.**

pkg/api/node/formatter.go
Yu-Jack: Same reason in https://github.com/harvester/harvester/pull/5704#discussion_r1622033858.

Sure, list-actions should be `GET`.

pkg/api/vm/handler.go
Yu-Jack: Same reason in https://github.com/harvester/harvester/pull/5704#discussion_r1622033858.

pkg/api/image/formatter.go
Yu-Jack: Actually, subresource concept is the verb to original resource. It's that we use subresources (verbs) to manipulate existing resource instead of creating, so it's basically a `PUT` or `GET` method. There shouldn't be `POST` method here.

Issue #5749, opened 2024-05-07
[BUG] Some images should not be purged after the upgrade is completed


starbops:
**Describe the bug**

We introduced a new step, image cleanup, after the upgrade has completed (tracked in #4995). This saves lots of precious disk spaces. However, removing the images still used by some of the Pods could be a catastrophe in **air-gapped** environments. We've already added the `harvester-upgrade` image to the retaining list, but there are more.

For example:

- docker.io/longhornio/longhorn-engine
- docker.io/rancher/mirrored-fluent-fluent-bit

**To Reproduce**
1. Prepare an air-gapped 3-node Harvester cluster 
2. Upgrade the cluster
3. After the upgrade ends, observe the cluster state, there should be some Pods fail to run

**Expected behavior**

After the upgrade completed, all of the Pods should be running without problems.

**Support bundle**

[supportbundle_2687bb0c-dea1-4978-956d-bdbb2f5d927a_2024-05-07T03-28-20Z.zip](https://github.com/harvester/harvester/files/15233947/supportbundle_2687bb0c-dea1-4978-956d-bdbb2f5d927a_2024-05-07T03-28-20Z.zip)

**Environment**
 - Harvester ISO version: v1.2.1 (upgrading to v1.2.2-rc2)
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

**Additional context**

This is first found in [here](https://github.com/harvester/harvester/issues/5473#issuecomment-2097793487).

w13915984028:
This is kind of silent behaviour change.

(1) Is it mentioned in the Harvester docuement?

(2) Could we add a harvester-setting, or some labels/annotations for user to enable/disable it?  For some users, maybe system disk usage is not an important issue.

Or when the system disk reachs certain percent before kubelet trigger the purging, Harvester does it in advance.

Or we list all the container imges in Harvester NODE page, let user have chance to choose & remove those `old-versioned` images.

bk201:
>(1) Is it mentioned in the Harvester docuement?

@starbops Please add a section in the doc to mention the cleanup.

> (2) Could we add a harvester-setting, or some labels/annotations for user to enable/disable it? For some users, maybe system disk usage is not an important issue.
> 
> Or when the system disk reachs certain percent before kubelet trigger the purging, Harvester does it in advance.
> 
> Or we list all the container imges in Harvester NODE page, let user have chance to choose & remove those old-versioned images.

[...]

PR #5750, opened 2024-05-07
fix: add images which are needed post-upgrade into retaining list

starbops:

**Problem:**

Some of the pods like LH instance-manager will continue to use the old container image after the upgrade ended, such images should not be cleaned up. Otherwise, it could break the service especially for air-gapped environment.

**Solution:**

Find out those critical images and adding them into the retaining list so that they won't get purged.

**Related Issue:**

#5749

**Test plan:**

1. Build an ISO image containing this fix
2. Prepare a Harvester cluster (v1.3.0) in air-gapped environment
3. Start the upgrade
4. After the upgrade ends, check for the following images. Both the old and new versions should exist on the node
   - `rancher/harvester-upgrade`
   - `longhornio/longhorn-engine`
   - `longhornio/longhorn-instance-manager`
   - `rancher/mirrored-banzaicloud-fluentd`
   - `rancher/mirrored-fluent-fluent-bit`
5. Check if any pod fails to run

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_repo.go
+++ b/pkg/controller/master/upgrade/upgrade_repo.go
@@ -46,6 +46,10 @@ var (
 	// the end of an upgrade.
 	imageRetainList = []string{
 		"rancher/harvester-upgrade",
+		"longhornio/longhorn-engine",
+		"longhornio/longhorn-instance-manager",
+		"rancher/mirrored-banzaicloud-fluentd",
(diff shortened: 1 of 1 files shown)


bk201 (approved):
LGTM

FrankYang0529 (approved):
LGTM. Thanks.

bk201:
@mergifyio backport v1.2 v1.3

Issue #5528, opened 2024-04-09
[BUG] DeleteNamespacedVirtualMachine returns `VirtualMachine` not `Status` object


drewmullen:
**Describe the bug**

The delete vm action response is documented as `"#/components/schemas/k8s.io.v1.Status"`

https://github.com/harvester/harvester/blob/ba586eaf0a4eb5fe58b79c444eadc4ce18f9024b/api/openapi-spec/swagger.json#L7347

However, what comes back from the API is 

```json
{
    "apiVersion": "kubevirt.io/v1",
    "items": [...],
    "kind": "VirtualMachine",
}
```
or sometimes... trying to isolate what is different in 1 of my test cases where the return type is different and why
```json
    "kind": "VirtualMachineList",
```

**Expected behavior**

Docs should match output from the API

drewmullen:
Actually as i test this more, i usually get `"kind": "VirtualMachine"`. Trying to isolate why in some scenarios i get `VirtualMachineList`

drewmullen:
I would also be interested to know why `DeleteOptions` is required: 

https://github.com/harvester/harvester/blob/ba586eaf0a4eb5fe58b79c444eadc4ce18f9024b/api/openapi-spec/swagger.json#L7330

which requires:

https://github.com/harvester/harvester/blob/ba586eaf0a4eb5fe58b79c444eadc4ce18f9024b/api/openapi-spec/swagger.json#L10195-L10200

you can set the values to "" and the API still works fine

m-ildefons:
`DeleteOptions` are inherited from the K8s API and allow for advanced features, like foreground deletion, cascading deletion and dry-runs: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/delete-options/
There are reasonable defaults and if everything goes smoothly and quickly you probably don't notice a difference. But when freeing resources takes some time (or worse may block until an operator manually intervenes) the deletion options may matter a lot, especially in the context of automated pipelines.

There is both a regular `delete` and a `deletecollection` request in the K8s API and this is also implemented similarly for the KubeVirt CRDs in use in Harvester.
The regular `delete` request acts on a singular resource and usually returns the deleted resource upon successful completion of the request. The `deletecollection` request acts on multiple resources (e.g. filtered by a label selector) and returns a status upon successful completion:

- `delete` VirtualMachine: https://kubevirt.io/api-reference/v1.2.0/operations.html#_deletenamespacedvirtualmachine
[...]

PR #5782, opened 2024-05-09
[Doc] return object type on delete, not `status`

### LINKED ISSUE #5781: [BUG] Delete PVC should return PVC not status
**Describe the bug**

Deleting a `persistentvolumeclaim` should return the pvc object, not `Status`: 

https://github.com/harvester/harvester/blob/bf5847ad1bac8d63be93355559b85af4f86d1901/api/openapi-spec/swagger.json#L410-L425

This API comes from the k8s API which documents the appropriate return type: https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json

image

drewmullen:
Resolves: #5781, https://github.com/harvester/harvester/issues/5528

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -413,12 +413,12 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/k8s.io.v1.Status"
+                  "$ref": "#/components/schemas/k8s.io.v1.PersistentVolumeClaim"
                 }
               },
               "application/yaml": {
                 "schema": {
-                  "$ref": "#/components/schemas/k8s.io.v1.Status"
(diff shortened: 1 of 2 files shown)


m-ildefons (approved):
LGTM.

This fixes the DELETE operation, but keep in mind that Kubernetes also knows a DELETECOLLECTION operation.
The way this works is that a `delete` request is issued, but the URL path does not contain a specific name for the DELETECOLLECTION operation, but rather a selector (like a label-selector) is passed as URL parameter. All objects that match this selector are then deleted in bulk.
The DELETECOLLECTION operation does indeed return a Status object.
As far as I can tell, we don't have this operation documented at all though.

drewmullen:
Harvester API returns a list of VMs deleted when you call DELETECOLLECTION. But, as you said, that action isn’t documented

Yu-Jack (approved):
LGTM

Issue #5796, opened 2024-05-13
[BUG] When upgrading 1 node Harvester with PCI passthrough the VM won't come up and kube-vip crashes


noahgildersleeve:
**Describe the bug**

When you upgrade a single node Harvester cluster from v1.2.1 -> v1.2.2-rc3
**To Reproduce**
1. Create a Harvester cluster
2. Add PCI passthrough
3. Create a VM 
4. Turn on the VM
5. Add the PCI device to the VM and reboot
6. Validate that the device was added
7. Turn off VM
8. Upgrade to v1.2.2-rc3
9. Wait for upgrade to finish
10. Turn VM back on

**Expected behavior**

VM should turn on successfully.

**Support bundle**

[supportbundle_09ea7a32-1b8f-4996-80fa-845eb2a84415_2024-05-13T20-29-36Z.zip](https://github.com/harvester/harvester/files/15299972/supportbundle_09ea7a32-1b8f-4996-80fa-845eb2a84415_2024-05-13T20-29-36Z.zip)

**Environment**
 - Harvester ISO version: v1.2.1 -> v1.2.2-rc3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): DL360

**Additional context**
Possibly related to #5373 and #5374 .

It's worth noting I can still admin the cluster from the node IP and from the upstream Rancher cluster where it is imported. Just the VIP seems to be down.

The kubevip worked immediately after the upgrade, and stopped working after a while. `kube-vip` is getting this error then it's looping on the waiting error.
```
time="2024-05-13T20:27:11Z" level=info msg="Using existing macvlan interface for DHCP [vip-e8d7c773]"
time="2024-05-13T20:27:21Z" level=error msg="timeout to request the IP from DHCP server for service kube-system/ingress-expose"
time="2024-05-13T20:27:46Z" level=error msg="request failed, error: got an error while processing the request: no matching response packet received (waiting 10s)"
time="2024-05-13T20:28:28Z" level=error msg="request failed, error: got an error while processing the request: no matching response packet received (waiting 1m0s)"
```

The PCI passthrough VM is getting a memory setting error. I recall a similar issue in the past where when you edited a VM and added PCI passthrough it would not work because we were overcommiting memory by default and it seemed like kubevirt didn't like that with VMs that had PCI devices.

[...]

PR #5802, opened 2024-05-14
Additional checks for host devices in vm mutator

ibrokethecloud:

**Problem:**

During upgrade from v1.2.1 to v1.2.2 / v1.2-head, VMs turned off during upgrade cannot be powered due to issue: https://github.com/kubevirt/kubevirt/issues/10379

The mutator does not patch these VM's as there is no additional/removal of host devices or GPUs

**Solution:**

additional check to ensure domain memory and requested memory match when host devices or gpu devices are present in VM. This is only likely to be called during the upgrade path.

**Related Issue:**
https://github.com/harvester/harvester/issues/5796
**Test plan:**

### DIFF
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -174,12 +174,7 @@ func needUpdateResourceOvercommit(oldVM, newVM *kubevirtv1.VirtualMachine) bool 
 		return true
 	}
 
-	// if host devices or GPUs are added or removed then resource update is needed
-	if len(newVM.Spec.Template.Spec.Domain.Devices.HostDevices) != len(oldVM.Spec.Template.Spec.Domain.Devices.HostDevices) || len(newVM.Spec.Template.Spec.Domain.Devices.GPUs) != len(oldVM.Spec.Template.Spec.Domain.Devices.GPUs) {
-		return true
-	}
-
-	return false
+	return hostDevicesOvercommitNeeded(oldVM, newVM)
 }
 
 func (m *vmMutator) patchResourceOvercommit(vm *kubevirtv1.VirtualMachine) ([]string, error) {
@@ -241,7 +236,7 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity, 
 	guestMemory.Sub(reservedMemory)
 
 	// Needed to avoid issue due to overcommit when GPU or HostDevices are passed to a VM.
-	// Addresses issue: https://github.com/kubevirt/kubevirt/issues/10379
+	// Addresses issue: `https://github.com/kubevirt/kubevirt/issues/10379`
 	// This condition can be removed once a fix is available upstream
 	if hostDevicesPresent(vm) {
 		guestMemory.DeepCopyInto(quantity)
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (approved):
LGTM, thanks for the fixing!

bk201 (approved):
lgtm, thanks

bk201:
@mergifyio backport v1.2 v1.3

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -454,3 +449,19 @@ func hostDevicesPresent(vm *kubevirtv1.VirtualMachine) bool {
...
+func hostDevicesOvercommitNeeded(oldVM, newVM *kubevirtv1.VirtualMachine) bool {
+	// if host devices or GPUs are added or removed then resource update is needed
+	if len(newVM.Spec.Template.Spec.Domain.Devices.HostDevices) != len(oldVM.Spec.Template.Spec.Domain.Devices.HostDevices) || len(newVM.Spec.Template.Spec.Domain.Devices.GPUs) != len(oldVM.Spec.Template.Spec.Domain.Devices.GPUs) {
+		return true
+	}
+
+	// during upgrade path VMs with host devices are turned off. Post upgrade the memory needs to be patched
+	// to ensure devices with hostDevices/GPUs can be booted
+	if len(newVM.Spec.Template.Spec.Domain.Devices.HostDevices) > 0 || len(newVM.Spec.Template.Spec.Domain.Devices.GPUs) > 0 {
+		if newVM.Spec.Template.Spec.Domain.Memory != nil && (newVM.Spec.Template.Spec.Domain.Memory.Guest.Value() != newVM.Spec.Template.Spec.Domain.Resources.Requests.Memory().Value()) {
```
w13915984028, line 462: Question:

The `newVM.Spec.Template.Spec.Domain.Memory.Guest.Value()` seems always differ with `newVM.Spec.Template.Spec.Domain.Resources.Requests.Memory().Value()`,  

will the above code always run, or the `newVM.Spec.Template.Spec.Domain.Memory.Guest.Value()` will be equal to `newVM.Spec.Template.Spec.Domain.Resources.Requests.Memory().Value()` after patch ? 

```

            hostDevices:
            - deviceName: nvidia.com/TU104GL_TESLA_T4
              name: pci-test-01-000008000
              
              
          machine:
            type: q35
          memory:
            guest: 8092Mi
          resources:
            limits:
              cpu: "4"
              memory: 8Gi
            requests:
              cpu: 250m
              memory: 5461Mi
        evictionStrategy: LiveMigrate
        hostname: focal-cuda-01
 ...
  status:
    conditions:
    - lastProbeTime: "2024-05-13T20:29:10Z"
      lastTransitionTime: "2024-05-13T20:29:10Z"
      message: Guest VM is not reported as running
      reason: GuestNotRunning
      status: "False"
      type: Ready
    - lastProbeTime: "null"
      lastTransitionTime: "null"
      message: VMI uses a PCI host devices
      reason: HostDeviceNotLiveMigratable
      status: "False"
      type: LiveMigratable
    - lastProbeTime: "null"
      lastTransitionTime: "2024-05-13T20:29:27Z"
      message: 'server error. command SyncVMI failed: "LibvirtError(Code=1, Domain=10,
        Message=''internal error: Process exited prior to exec: libvirt:  error :
        cannot limit locked memory of process 81 to 9558818816: Operation not permitted'')"null'
      reason: Synchronizing with the Domain failed.
      status: "False"
````

ibrokethecloud: the check will only be applied when hostdevices/gpus are present in the vm definition. When vm meets this condition the resource limits will be recalculated. This is already being done in v1.3.0 and v1.2.2-rc when a new vm is launched or an existing one is updated to add/remove hostdevices/gpus.
The checks addresses the upgrade path from 1.2.1 for VM's with host devices which were turned off during the upgrade. Powering on the vm is missed by the check currently as there is no change in number of hostdevices/gpus or vm resource requirements and we end up with the VM being unable to start and the error we see above.

ibrokethecloud: in addition to above, this check will be run on each VM stop/start. This will trigger a calculation of memory overcommit and a patch will be generated if needed. Thanks @w13915984028 for pointing out.

Issue #2305, opened 2022-05-23
[FEATURE] Support CPU Pinning


janeczku:
Certain VM workloads, requiring a predictable latency and enhanced performance during its execution would benefit from obtaining dedicated CPU resources. KubeVirt, relying on the Kubernetes CPU manager, is able to pin guest's vCPUs to the host's pCPUs.
Harvester should expose this functionality to the user.

Ref: https://kubevirt.io/user-guide/virtual_machines/dedicated_cpu_resources/

gusraggio:
This is needed in telco use case. Also related to this but different is CPU prioritization some VMs, need to be prioritized to others but dont need a CPU assigned to a VM.

For example, if the server is overloaded prioritized VMs should be respected. Do we need a separeted issue?

torchiaf:
@brandboat I think Harvester server should return the new `enableCPUManager` and `disableCPUManager` nodes actions along with the already available actions in the `nodes` API response:

![image](https://github.com/user-attachments/assets/94e6d616-e462-4af4-92b8-e23b0b2443f7)

so that the UI can show the action buttons accordingly with current mechanism.
Should we also update the [HEP](https://github.com/harvester/harvester/pull/5803) as well ?
cc @bk201

brandboat:
> @brandboat I think Harvester server should return the new enableCPUManager and disableCPUManager nodes actions along with the already available actions in the nodes API response:
> 
> so that the UI can show the action buttons accordingly with current mechanism.
> Should we also update the https://github.com/harvester/harvester/pull/5803 as well ?

Thanks @torchiaf, the enableCPUManager and disableCPUManager apis are mentioned in  https://github.com/harvester/harvester/blob/master/enhancements/20240506-cpu-pinning.md#api-changes, about the API response, I'll update the HEP after implementation done.

TachunLin:
Related GUI issue: https://github.com/harvester/harvester/issues/6056

TachunLin:
### Related Test Plan

1. Test enable/disable cpu manager through API on Harvester
2. Test enable/disable cpu pinning through API on virtual machine
3. Test the cpu pinning vm can request specific core cpu on host to it's vgpu.
4. Verify only one master node could enable/disable cpu manager in the same time 
5. Verify there should be only one job enable/disable cpu manager per node (API only)
6. Verify we can't enable cpu manager when it is enabled, and we can't disable CPU manager when it's disabled (API only)
7. Test reboot node with cpu manager enabled
[...]

PR #5803, opened 2024-05-14
HEP: add cpu pinning [ci skip]

brandboat:
Add CPU Pinning HEP

**Related Issue:**
https://github.com/harvester/harvester/issues/2305

**Test plan:**
N/A

### DIFF
--- a/enhancements/20240506-cpu-pinning.md
+++ b/enhancements/20240506-cpu-pinning.md
@@ -30,8 +30,14 @@ This HEP does not cover:
 Enabling CPU pinning for KubeVirt requires setting the kubelet argument `--cpu-manager-policy` to [static](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy).
 However, this change affects the way CPU resources are utilized by all pods that meet the condition as follows, not just VMs. 
 - Both requests and limits must be configured in pods (Guaranteed QoS) and their values must be the same integer.
+And that means if a node use static cpu manager policy, there will have two scenarios, one is pods that meet the criteria as above, and the other one is not.
+For the former, CPUs inside the pods are not isolated and are all shared with each other, and the later will have isolated CPUs in each pod.
+Note that if a node use none cpu manager policy, all CPUs in the pods are not isolated and could be shared with each other.
 
-So we cannot set `--cpu-manager-policy` to static on all nodes initially. Instead, we allow users to decide which nodes they want to set `--cpu-manager-policy=static`.Subsequently, when creating VMs, they can deploy the VM desired CPU pinning to the corresponding nodes with `--cpu-manager-policy=static` enabled.
+To address this, the proposal is to allow users to set the CPU manager policy for each node individually, rather than forcing all nodes to use the same policy. This flexibility is crucial because switching the CPU manager policy from static to none causes CPUs in Guaranteed QoS pods to lose isolation. To prevent issues, an error should be thrown if any VMs with CPU pinning enabled exist on a node where the user wants to switch from static to none. The user must then either migrate these VMs to another node or stop them. Forcing a uniform policy across all nodes would be problematic, as turning the CPU manager policy to none would require migrating all CPU pinned VMs to other nodes. If no nodes are using the static policy, migration is impossible, leaving users with no option but to stop all VMs, which is inconvenient and reduces availability.
+
+So we cannot set `--cpu-manager-policy` to static on all nodes initially. Instead, we allow users to decide which nodes they want to set `--cpu-manager-policy=static`.
+To enable CPU pinning, we also need to specify `dedicatedCpuPlacement=true` to VirtualMachine CR, this will add 1-1 CPU mapping to libvirt xml.
 
 ### User Stories
 
@@ -158,7 +164,8 @@ To implement this proposal, there are several steps need to be taken.
   policy only reserved cpus when both cpu requests and limits values must be the same integer.
 
 In step1, we have to set up the reserved resource due to the k8s cpu manager requirement.
-We can utilize the formula as follows to set `system-reserved` CPU resources during install harvester node. This is inspired by [GKE CPU reservations](https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes).
+And we should ensure kube-reserved and system-reserved are both defined as rke2 is using static pods and we should ensure they do not get impacted in anyway.
+Here we utilize the formula as follows to set `system-reserved` and `kube-reserved` CPU resources during install harvester node and upgrade path. This is inspired by [GKE CPU reservations](https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes).
 
 ```text
 6% of the first core +
@@ -170,11 +177,12 @@ if the maximum number of Pods per node beyond the default of 110 (Currently harv
 reserves an extra 400 mCPU in addition to the preceding reservations.
 ```
 
-After calculating the cpu reserved resources, add the following content to `/etc/rancher/rke2/config.yaml/d/99-z00-harvester-system-reserved.yaml` during harvester installation.
+After calculating the cpu reserved resources, add the following content to `/etc/rancher/rke2/config.yaml/d/99-z00-harvester-reserved-resources.yaml` during harvester installation.
 
 ```yaml
 kubelet-arg+:
 - "system-reserved=cpu=490m"
+- "kube-reserved=cpu=490m"
 ```
 
 and add another config `/etc/rancher/rke2/config.yaml/d/99-z01-harvester-cpu-manager.yaml` which includes the cpu-manager-policy setting.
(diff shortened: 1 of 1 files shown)


enhancements/20240506-cpu-pinning.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### Web UI
+Regarding the Web UI, we should display the number of CPUs already pinned on each node,
+as well as the remaining number of CPUs available for allocation on each node.
+
+### Test plan
+TBD
+
+## Discussion
+- How much CPU resource allocation should be reserved? We have to set cpu resources in either `--kube-reserved` or `--system-reserved`
```
w13915984028: Per current design, the LH related PODs has a feature https://docs.harvesterhci.io/v1.4/install/harvester-configuration#installharvesterlonghorndefault_settingsguaranteedinstancemanagercpu which claims 12% of NODE's CPU for dedicated usage. Consider such kind of limitation:
reserve: max (4, 20% of CPU cores); others for pin-able usage. ?

The offical document also mentioned:

> Note:System services such as the container runtime and the kubelet itself can continue to run on these exclusive CPUs.  The exclusivity only extends to other pods.

enhancements/20240506-cpu-pinning.md
```diff
@@ -0,0 +1,219 @@
...
+Now, let's change the `cpu-manager-policy` to `none` in node1, and observe the CPU set in VM `test`, which remains `1-8,25-32`.
+Next, we create another VM named `test2` with 2 CPUs without enabling CPU pinning and deploy it to node1,
+the CPU set in VM `test2` is `0-47`. Although CPU pinning in `test` still work, CPUs inside `test` is no longer isolated
+and could be shared with other VMs. 
+
+Upon checking other pre-existing pod, such as the `harvester-node-manager` pod which deployed in node1, we find that
+the cpu set in it remains `0,9-24,33-47`, despite the `cpu-manager-policy` being set to none in node1.
+
+Now, if we restart VM `test`, it will be deployed to node2 since `cpu-manager-policy` is set to `none` in node1, node3.
+Subsequently, if we disable CPU pinning in node2 and restart VM test, it won't deploy successfully since no node has set
```
w13915984028: Does upstream aleady considered the `migration` of such pin'ed VM?

enhancements/20240506-cpu-pinning.md
```diff
@@ -0,0 +1,219 @@
...
+`cpu-manager-policy` to `static`.
+
+#### Story 3: users who don't want to deploy workload to nodes that set cpu-manager-policy to static
```
w13915984028, line 97: When a node with `cpu-manager-policy` static and has already deployed many VMs, e.g. 80% CPUs are allcated/used, then user add a pin'ed VM to this node, what will happen? will the new VM fail to deploy (say only 1 node pin-able), or some VMs are migrated to other nodes to give space for it?

enhancements/20240506-cpu-pinning.md
```diff
@@ -0,0 +1,219 @@
...
+          limits:
+            cpu: 2
+          requests:
+            cpu: 2
+[...]
+```
+
+#### Web UI
+Regarding the Web UI, we should display the number of CPUs already pinned on each node,
+as well as the remaining number of CPUs available for allocation on each node.
```
w13915984028: Will UI do pin-able check in advance when creating a VM ?

w13915984028 (commented):
Good HEP. 

A few open questions for discussion.

enhancements/20240506-cpu-pinning.md
brandboat: > Note:System services such as the container runtime and the kubelet itself can continue to run on these exclusive CPUs. The exclusivity only extends to other pods.

Thanks for the comment, @w13915984028! 😃 
Indeed, that's why I'm struggling with determining how much CPU resources should we reserved. If system services still use CPUs allocated to VMs with CPU pinning, the concept of “exclusive CPUs” becomes meaningless.

> reserve: max (4, 20% of CPU cores); others for pin-able usage. ?

Sounds good, while I'm wondering when should we setup the CPU resource reservation ?
1. Whenever we want to setup static policy in each harvester node.
2. During the installation of harvester.

I would lean to option2 since it is simpler. To me, CPU resource reservation is more of a setting that harvester should be aware of from the beginning, not just for this feature (CPU Pinning).

enhancements/20240506-cpu-pinning.md
brandboat: Did you mean Live Migration (https://kubevirt.io/user-guide/operations/live_migration/) ? If so, I tested this feature before. Will add user story about this scenario in HEP.

enhancements/20240506-cpu-pinning.md
brandboat: In such scenario, the new VM will fail to deploy. And raise error like `0/1 nodes are available: 1 Insufficient cpu...`.

enhancements/20240506-cpu-pinning.md
brandboat: I'm not sure if this check is necessary. Whether the VM can be successfully deployed will only be known after it is actually deployed to the node.

enhancements/20240506-cpu-pinning.md
w13915984028: Yes, the `live-migration` feature. Please add it, thanks.

brandboat:
gentle ping @ibrokethecloud @bk201 , could you please take a look ? Many thanks 😃

enhancements/20240506-cpu-pinning.md
```diff
@@ -0,0 +1,302 @@
...
+- "cpu-manager-policy=none"
+```
+
+> [!NOTE]
+> After upgrading harvester to a newer version, a new file `99-max-pods.yaml` shows up which override all the kubelet-args, that's why I have to 
+> create another two files `/etc/rancher/rke2/config.yaml/d/99-z00-harvester-system-reserved.yaml`, `/etc/rancher/rke2/config.yaml/d/99-z01-harvester-cpu-manager.yaml`
+> instead of adding these configs to `90-harvester-server.yaml` or `90-harvester-worker.yaml`. Another benefit is that if we want to modify cpu manager policy setting,
+> we only need to update one file `99-z01-harvester-cpu-manager.yaml`.
+
+To change cpu-manager policy on the fly, utilizing `Plan` CRD here to deploy jobs to all nodes. We need to modify the secret `cattle-system/cpu-manager`, and fill every policy settings in each node that we want to change in the `.stringdata.configs` field. It should be a json format like the following example.
```
brandboat: Discussed offline with @bk201, we identified two drawbacks with this solution:
- Jobs are sent to all nodes, even though we only need to change the CPU manager policy on specific nodes.
- The process is not straightforward; ideally, there should be a web API instead of requiring developers/users to upload secrets.

I will refine the HEP and redesign how we manage the CPU manager policy per node.

bk201 (approved):
LGTM, clear and good writing.

ibrokethecloud (approved):
we should ensure `kube-system-reserved` and `system-reserved` are both defined as rke2 is using static pods and we should ensure they do not get impacted in anyway.

We also need to document the CPU pinning behaviour clearly from the KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/3570-cpumanager#policy-2-static-cpuset-control as the actual explaining of how the shared cpu pool is calculated is not clear in the docs https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy

rest looks good to me

enhancements/20240506-cpu-pinning.md
brandboat: Sorry for the late response, in the latest HEP we use formula in https://cloud.google.com/kubernetes-engine/docs/concepts/plan-node-sizes  to calculate `kube-reserved` and `system-reserved` cpu resource.

enhancements/20240506-cpu-pinning.md
brandboat: Already added live-migration user story to the HEP

brandboat:
thanks for the comments @ibrokethecloud !

> we should ensure kube-system-reserved and system-reserved are both defined as rke2 is using static pods and we should ensure they do not get impacted in anyway.

OK, so we should add both `kube-reserved` and `system-reserved` to the kubelet config file, I add that to the HEP, and both configs leverage the GKE formula to calculate reserved cpu resource. 

> We also need to document the CPU pinning behaviour clearly from the KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/3570-cpumanager#policy-2-static-cpuset-control as the actual explaining of how the shared cpu pool is calculated is not clear in the docs https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy

filed issue https://github.com/harvester/harvester/issues/6148, I'll open a doc pr and add everything you mentioned after the cpu pinning implementation merged.

w13915984028 (approved):
LGTM, thanks.

Issue #5805, opened 2024-05-14
[BUG] Crash when writing error status on VM backup


nistei:
**Describe the bug**
When creating a VM backup that has an error (e.g. PVC not found) Harvester crashes when there is not already a status set on the VMBackup. In our case, the error is caused by a PVC not being found. 

**To Reproduce**
1. Create a VM backup that will error (in our case, attaching a rook-ceph volume to the VM causes an error)
2. Go to VM Backups
3. See the Error label on the Backup but no status set. Harvester is in a crash loop.

**Expected behavior**
Harvester does not crash and sets the status properly.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.3.0

**Additional context**
```
E0513 16:40:26.195095       7 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 2388 [running]:
github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime.logPanic({0x280f820?, 0x50d2b10})
    /go/src/github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:75 +0x99
github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc0063b6160?})
    /go/src/github.com/harvester/harvester/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:49 +0x75
panic({0x280f820, 0x50d2b10})
    /usr/local/go/src/runtime/panic.go:884 +0x213
github.com/harvester/harvester/pkg/controller/master/backup.(*Handler).setStatusError(0xc000f9ca00, 0xc000e94100?, {0x3434b20, 0xc002164000})
    /go/src/github.com/harvester/harvester/pkg/controller/master/backup/backup.go:756 +0x128
github.com/harvester/harvester/pkg/controller/master/backup.(*Handler).OnBackupChange(0x0?, {0x0?, 0x0?}, 0xc000e94100)
    /go/src/github.com/harvester/harvester/pkg/controller/master/backup/backup.go:166 +0x1da
time="2024-05-13T16:40:26Z" level=info msg="Starting harvesterhci.io/v1beta1, Kind=Upgrade controller"
[...]

PR #5806, opened 2024-05-14
fix: crash when setting backup status

nistei:

**Problem:**
harvester crashes when attempting to set an error status when no status exists yet.

**Solution:**
Set a status when it doesn't already exist to prevent nil pointer errors.

**Related Issue:**
Refs: #5805

**Test plan:**


Yu-Jack (approved):
LGTM, thanks!

FrankYang0529 (approved):
LGTM. Thank you.

Issue #4282, opened 2023-07-20
[ENHANCEMENT] add webhook validation of the http proxy config


guangbochen:
**Is your enhancement related to a problem? Please describe.**
If you configure `httpProxy` and `httpsProxy`, you must also put Harvester node's CIDR into noProxy, otherwise, the Harvester cluster will be broken. It would be nice to add proper validation to prevent misconfiguration.

**Describe the solution you'd like**
We can add a webhook to validate the user input; it should reject if the node IP is not included in the no_proxy CIDR range.

**Describe alternatives you've considered**
Add the `caution` text to the UI before implementing the webhook.

**Additional context**
doc ref: https://docs.harvesterhci.io/v1.1/advanced/index/#http-proxy

futuretea:
If user configure cluster-registration-url setting, user might need to put the host of cluster-registration-url into noProxy too, otherwise, the Harvester cluster cannot be accessed from upstream Rancher

irishgordo:
@votdev thanks for this enhancment :smile: :+1: 
This looks good on  Version: v1.4.0-dev-20240830 

With just the added context about the open issues/enhancments:
- https://github.com/harvester/harvester/issues/6539
- https://github.com/harvester/harvester/issues/6440

Everything looks good :smile: from all test-cases.
Some small artifacts:
![Screenshot from 2024-09-10 11-49-20](https://github.com/user-attachments/assets/9cf35e96-f76e-4b1f-8ed8-819854968bd8)
![Screenshot from 2024-09-10 11-48-16](https://github.com/user-attachments/assets/d634e3eb-b41d-4f57-a1a7-74b074bb32af)

I'll go ahead and close this out :+1:

PR #5824, opened 2024-05-16
Make sure the node's IP addresses is set in noProxy

votdev:
grafik

**Problem:**
If you configure httpProxy and httpsProxy, you must also put Harvester node's CIDR into noProxy, otherwise, the Harvester cluster will be broken.

**Solution:**
Add a webhook to validate the user input; it should reject if the node IP is not included in the no_proxy CIDR range.

**Related Issue:**
https://github.com/harvester/harvester/issues/4282

**Test plan:**
The test plan assumes that several hosts are available in the cluster. For the following test cases, the hosts with the following IP addresses are involved:

- 192.168.0.30
- 192.168.0.31
- 192.168.0.32

## Case 1:
1. Go to `Advanced` > `Settings` and then select `Edit Settings` for `http-proxy`.
2. Press `Save`.
3. The error message `noProxy should contain the node's IP addresses or CIDR. The node(s) 192.168.0.30, 192.168.0.31, 192.168.0.32 are not covered.` is displayed.

grafik

## Case 2:
1. Go to `Advanced` > `Settings` and then select `Edit Settings` for `http-proxy`.
2. Use `192.168.0.0/27` in the `no-proxy` form field.
3. Press `Save`.
4. The error message  `noProxy should contain the node's IP addresses or CIDR. The node(s) 192.168.0.32 are not covered.` is displayed.

## Case 3:
1. Go to `Advanced` > `Settings` and then select `Edit Settings` for `http-proxy`.
2. Use `192.168.0.0/24` in the `no-proxy` form field.
3. Press `Save`.
4. The setting is saved and you'll be redirected to the `Settings` page.

## Case 4:
1. Go to `Advanced` > `Settings` and then select `Edit Settings` for `http-proxy`.
2. Use `192.168.0.30, 192.168.0.31, 192.168.0.32` in the `no-proxy` form field.
3. Press `Save`.
4. The setting is saved and you'll be redirected to the `Settings` page.


Vicente-Cheng:
Hi @votdev, 
Could you move the common function (e.g. mpa/slice) to https://github.com/harvester/go-common?
That we could reuse this in various components!

Thanks!

votdev:
> Hi @votdev, Could you move the common function (e.g. mpa/slice) to [harvester/go-common](https://github.com/harvester/go-common)? That we could reuse this in various components!
> 
> Thanks!

See https://github.com/harvester/go-common/pull/14

pkg/webhook/resources/setting/validator.go
```diff
@@ -213,19 +223,81 @@ func validateDeleteSetting(oldObj runtime.Object) error {
...
+	// Make sure the node's IP addresses is set in 'noProxy'. These IP
+	// addresses can be specified individually or via CIDR address.
+	nodes, err := v.nodeCache.List(labels.Everything())
+	if err != nil {
+		return err
+	}
+	err = validateNoProxy(httpProxyConfig.NoProxy, nodes)
+	if err != nil {
+		return err
+	}
```
Yu-Jack: Should they use `werror.NewInvalidError` to create validation error as well?

votdev: > Should they use `werror.NewInvalidError` to create validation error as well?

The function `validateNoProxy` returns `NewInvalidError` errors. But you are right for the `nodes, err := v.nodeCache.List(labels.Everything())` error case.

FrankYang0529 (approved):
LGTM. Thanks for the PR.

Yu-Jack (approved):
LGTM, thanks!

mingshuoqiu:
LGTM.

votdev:
@Mergifyio backport v1.2 v1.3

Issue #1646, opened 2021-12-08
[FEATURE] Support Bundle timeout enhancement 


guangbochen:
I would suggest we revert the change: rancher/support-bundle-kit#16 (comment) to collect node bundles even some nodes don't push bundle ack in time.

Related to: https://github.com/harvester/harvester/issues/1585

Yu-Jack:
It's related to https://github.com/rancher/support-bundle-kit/issues/73

Yu-Jack:
Some findings from @albinsun , some minimum value should not be `30`. It's default. PR https://github.com/harvester/docs/pull/646 for this.

albinsun:
Test PASS, close as done.

Note that current [doc](https://docs.harvesterhci.io/v1.4/advanced/index/#support-bundle-node-collection-timeout) does not mention about the definition of `0` like [`support-bundle-timeout`](https://docs.harvesterhci.io/v1.4/advanced/index/#support-bundle-timeout) and should be an enhancement.
CC. @Yu-Jack 


### Environment
* Harvester
  * Version: v1.4.0-dev-20240918
  * Profile: 3 nodes QEMU/KVM (8C/16G/500G per node)
  * ui-source: Auto
 

### Verification
Define settings > support-bundle-node-collection-timeout (unit is minute)
Create support bundle
Check SUPPORT_BUNDLE_NODE_TIMEOUT of support bundle manager deployment.apps

#### Positive
1. :green_circle: Default value and successful donwload
   <details>
   <summary> <i>Given</i> <b>support-bundle-node-collection-timeout</b> is <code>30</code> (default)
   </summary>
   
   ![image](https://github.com/user-attachments/assets/f814575c-3c41-4bdc-a34a-638bd4358af6)
   </details>

   <details>
   <summary> <i>When</i> <b>Generate a Support Bundle</b>
   </summary>
   
   ![image](https://github.com/user-attachments/assets/a8323423-a51f-4f6a-8c59-929f4bf8ac86)
   </details>

   <details>
   <summary> <i>Then</i> SB deployment's <code>SUPPORT_BUNDLE_NODE_TIMEOUT</code> is <code>30</code>
   </summary>
   
   ![image](https://github.com/user-attachments/assets/589e7473-c0a4-46ac-93a6-55554febc6bd)
   </details>

   <details>
   <summary> <i>And</i> SB download successfully and contains all 3 nodes log
   </summary>
   
[...]

PR #5838, opened 2024-05-21
feat: add support bundle node collection timeout

Yu-Jack:

**Problem:**

Currently, there is no timeout for nodes when creating support bundle files, resulting in infinite waiting for node collection. That means we can't download support bundle file.

**Solution:**

- First, support node timeout in support-bundle-kit https://github.com/rancher/support-bundle-kit/pull/109.
- Second, provide custom setting of node timeout in harvester (this PR).

By this, we could still download support bundle file even node is timeout.

**Related Issue:**

https://github.com/harvester/harvester/issues/1646

**Test plan:**

Prepare:

1. Use new harvester with new support bundle kit image
2. Define `settings > support-bundle-node-collection-timeout` (unit is minute)
3. Create support bundle 
4. Check env fields of support bundle manager deployment

Two different cases:

- Case A: define large and enough minutes, it should get nodes folder successfully.
- Case B: define a small minute, like `"1"`. This one depends how long your environment takes. Just define a very small minute to make timeout happen. Although It won't get nodes folder, but It will get the support bundle file zip.

### DIFF
--- a/pkg/controller/master/supportbundle/manager.go
+++ b/pkg/controller/master/supportbundle/manager.go
@@ -54,8 +54,8 @@ func (m *Manager) Create(sb *harvesterv1.SupportBundle, image string, pullPolicy
 	deployName = m.getManagerName(sb)
 	logrus.Debugf("creating deployment %s with image %s", deployName, image)
 
-	if expiration := settings.SupportBundleNodeTimeout.GetInt(); expiration == 0 {
-		nodeTimeout = time.Duration(supportBundleUtil.SupportBundleNodeTimeoutDefault) * time.Minute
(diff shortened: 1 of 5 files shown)


Yu-Jack:
Cause https://github.com/longhorn/longhorn-manager/pull/2825 longhorn side uses `support-bundle-node-collection-timeout`, I think that one is more accurate, so I'll use it in harvester side, and for unite as well

bk201 (approved):
LGTM

ibrokethecloud (approved):
lgtm. thanks.

Yu-Jack:
@Mergifyio backport v1.3

PR #5864, opened 2024-05-23
fix: Harvester ManagedChart is modified after an upgrade

bk201:
Ignore the status fields in some post-modified CRDs in the upgrade path.

**Problem:**

**Solution:**

**Related Issue:**

**Test plan:**

### DIFF
--- a/package/upgrade/migrations/managed_charts/v1.2.1.sh
+++ b/package/upgrade/migrations/managed_charts/v1.2.1.sh
@@ -1,0 +1,19 @@
+#!/bin/bash -ex
+
+CHART_NAME=$1
+CHART_MANIFEST=$2
+
+patch_ignoring_resource()
+{
+	# add ignoring resources ()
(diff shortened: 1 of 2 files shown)


Vicente-Cheng (approved):
Thanks for fixing that!

w13915984028:
A question here: the `ComparePatches` is defined as an array. The multi run of current code will expand the array. Should we check if the target value is not in the array and then append it?

```
type DiffOptions struct {
	ComparePatches []ComparePatch `json:"comparePatches,omitempty"`
}

type ComparePatch struct {
	Kind         string      `json:"kind,omitempty"`
	APIVersion   string      `json:"apiVersion,omitempty"`
	Namespace    string      `json:"namespace,omitempty"`
	Name         string      `json:"name,omitempty"`
	Operations   []Operation `json:"operations,omitempty"`
	JsonPointers []string    `json:"jsonPointers,omitempty"`
}
```

bk201:
> A question here: the `ComparePatches` is defined as an array. The multi run of current code will expand the array. Should we check if the target value is not in the array and then append it?
> 

Good point, updated. The script is now idempotent. I'd keep the v1.2.1 one since it's already shipped.

bk201:
@mergifyio backport v1.3

Vicente-Cheng (approved):
LGTM, tested the upgrade from v1.3.0 to v1.3-head.

Works well.

starbops (approved):
Tested with upgrade from v1.3.0 to master containing this fix. LGTM, thank you.

Issue #5876, opened 2024-05-24
[BUG] Setting Harvester NTP Servers Is Only Additive


irishgordo:
**Describe the bug**
Harvester Dashboard -> Settings -> NTP Servers do not accurately update '/etc/systemd/timesyncd.conf`.

**To Reproduce**
1. Set up a single-node Harvester cluster v1.3-head
1. Ensure that you only implement 1 NTP Server at the Cluster Creation
1. Once Harvester is installed, check that the single NTP server is reflected on the node via `cat /etc/systemd/timesyncd.conf` , you should see your single NTP server there and present
1. In Harvester Dashboard -> Settings -> NTP Servers, modify that, removing your "old" ntp-server and then placing in a "new" ntp-server that is different than the original, for instance if your original NTP Server was "time1.google.com", change that to something like "time2.google.com" or really whatever you desire
1. Save those changes
1. Wait 20 minutes or less, (I waited 20), then on the Harvester node, again, audit the timesyncd.conf with `cat /etc/systemd/timesyncd.conf` you'll notice, that it has only "added" the other server that it is not "truly" reflective of what we set it in the Harvester settings.  

Example.

Original NTP Server:
```
harvester-node-0:/home/rancher # cat /etc/systemd/timesyncd.conf

[Time]
NTP = time1.google.com
```
After Modification (if let's say you modified the ntp-server in Harvester Dashboard -> Settings -> NTP Servers, from time1.google.com to replacing that with time2.google.com):
```
harvester-node-0:/home/rancher # cat /etc/systemd/timesyncd.conf

[Time]
NTP = time1.google.com time2.google.com
```

**Workaround**
- Manually modify the /etc/systemd/timesyncd.conf and remove the "legacy"/"old" NTP server from it's configuration, then additionally restart systemd-timesyncd.service

**Expected behavior**
Expectation would be that we would have parity reflected with `/etc/systemd/timesync.conf` - but it seems to only be "additive" in nature, and would not remove the original ntp-server from the `/etc/systemd/timesyncd.conf` even if in our Harvester Settings -> NTP Servers that server is nowhere to be found

**Environment**
 - Harvester ISO version: v1.3-head
 - Underlying Infrastructure: qemu/kvm

irishgordo:
Thanks for this fix @brandboat  :smile: :+1:  - this looks great on Harvester Version: v1.4.0-dev-20240705 

[...]

PR #5880, opened 2024-05-27
Fix incorrect NTP server settings behavior

brandboat:
**Problem:**
Harvester NTP server settings not correctly mapping to NodeConfig ntpConfigs.
https://github.com/harvester/harvester/blob/226f72aa58532071232ad548e27970c1cec55733/pkg/controller/master/setting/ntp_server.go#L54 retrieve ntp settings from NodeConfig and https://github.com/harvester/harvester/blob/226f72aa58532071232ad548e27970c1cec55733/pkg/util/ntp.go#L19-L21 append the ntp server if it not exist in the current net servers slice. This causes an issue where even if we clear the NTP server settings and add new ones, NodeConfig will append the new NTP servers to the existing configuration instead of overriding the original settings.

**Solution:**
Reflect what user set `ntp-servers` in settings CR to NodeConfig CR `ntpConfigs`

This pr also addressed the follows:
- Add validation rule to check if there is any duplicate NTP servers in settings CR
- Enhance error message when setup ntp server string start with https
- Add validateNTPServers unit tests

**Related Issue:**
https://github.com/harvester/harvester/issues/5876

**Test plan:**
Prerequisites:
- Set up a Harvester cluster

Test steps:
- Dashboard -> Settings -> NTP Servers, remove `0.suse.pool.ntp.org` `1.suse.pool.ntp.org` and add `time1.google.com` and `time2.google.com`
- Wait a couple seconds, go to each Harvester node and check `cat /etc/systemd/timesyncd.conf`. The content should be `time1.google.com` and `time2.google.com`
- Run `systemctl status systemd-timesyncd.service` and check service status is `active` and time server is sync from 216.239.35.0:123 (time1.google.com)
```
● systemd-timesyncd.service - Network Time Synchronization
     Loaded: loaded (/usr/lib/systemd/system/systemd-timesyncd.service; enabled; vendor preset: enabled)
     Active: active (running) since Mon 2024-05-27 09:41:14 UTC; 20s ago
       Docs: man:systemd-timesyncd.service(8)
   Main PID: 4053 (systemd-timesyn)
     Status: "Initial synchronization to time server 216.239.35.0:123 (time1.google.com)."
      Tasks: 2
     CGroup: /system.slice/systemd-timesyncd.service
             └─ 4053 /usr/lib/systemd/systemd-timesyncd

May 27 09:41:14 harvester-node-0 systemd[1]: Starting Network Time Synchronization...
May 27 09:41:14 harvester-node-0 systemd[1]: Started Network Time Synchronization.
May 27 09:41:15 harvester-node-0 systemd-timesyncd[4053]: Initial synchronization to time server 216.239.35.0:123 (time1.google.com).
```


pkg/webhook/resources/setting/validator_test.go
```diff
@@ -218,3 +218,86 @@ func Test_validateSSLProtocols(t *testing.T) {
...
+	testCases := []struct {
+		name        string
+		args        *v1beta1.Setting
+		expectedErr string
+	}{
+		{
+			name: "valid empty ntp servers",
+			args: &v1beta1.Setting{
+				ObjectMeta: v1.ObjectMeta{Name: settings.NTPServersSettingName},
+				Value:      "",
```
brandboat, line 508: Not sure if allowing empty NTP Servers is ok or not.
If we set empty NTP Servers. `systemd-timesyncd.service` fallback to `0.suse.pool.ntp.org`

```bash
harvester-node-0:~ # cat /etc/systemd/timesyncd.conf

[Time]
NTP =
```

```bash
harvester-node-0:~ # systemctl status systemd-timesyncd.service
● systemd-timesyncd.service - Network Time Synchronization
     Loaded: loaded (/usr/lib/systemd/system/systemd-timesyncd.service; enabled; vendor preset: enabled)
     Active: active (running) since Mon 2024-05-27 09:44:47 UTC; 22s ago
       Docs: man:systemd-timesyncd.service(8)
   Main PID: 11458 (systemd-timesyn)
     Status: "Initial synchronization to time server 125.229.162.223:123 (0.suse.pool.ntp.org)."
      Tasks: 2
     CGroup: /system.slice/systemd-timesyncd.service
             └─ 11458 /usr/lib/systemd/systemd-timesyncd

May 27 09:44:46 harvester-node-0 systemd[1]: Starting Network Time Synchronization...
May 27 09:44:47 harvester-node-0 systemd[1]: Started Network Time Synchronization.
May 27 09:44:52 harvester-node-0 systemd-timesyncd[11458]: Initial synchronization to time server 125.229.162.223:123 (0.suse.pool.ntp.org).
```

```bash
harvester-node-0:~ # timedatectl show-timesync
FallbackNTPServers=0.suse.pool.ntp.org 1.suse.pool.ntp.org 2.suse.pool.ntp.org 3.suse.pool.ntp.org
ServerName=0.suse.pool.ntp.org
ServerAddress=125.229.162.223
RootDistanceMaxUSec=5s
PollIntervalMinUSec=32s
PollIntervalMaxUSec=34min 8s
PollIntervalUSec=1min 4s
NTPMessage={ Leap=0, Version=4, Mode=4, Stratum=2, Precision=-23, RootDelay=8.697ms, RootDispersion=37.475ms, Reference=D8EF2304, OriginateTimestamp=Mon 2024-05-27 09:44:52 UTC, ReceiveTimestamp=Mon 2024-05-27 09:44:52 UTC, TransmitTimestamp=Mon 2024-05-27 09:44:52 UTC, DestinationTimestamp=Mon 2024-05-27 09:44:52 UTC, Ignored=no PacketCount=1, Jitter=0 }
Frequency=3391487
```

brandboat:
gentle ping @Vicente-Cheng , @bk201 , could you please take a look when you are available ? Thanks !

pkg/webhook/resources/setting/validator.go
```diff
@@ -1060,3 +1068,20 @@ func validateKubeConfigTTLSetting(newSetting *v1beta1.Setting) error {
 func validateUpdateKubeConfigTTLSetting(_ *v1beta1.Setting, newSetting *v1beta1.Setting) error {
 	return validateKubeConfigTTLSetting(newSetting)
 }
+
+func findDuplicates(arr []string) []string {
```
bk201: (optional) I feel we implement this somewhere or maybe worth moving to https://github.com/harvester/go-common/blob/main/slice.go

bk201 (commented):
LGTM, but let @Vicente-Cheng check if there was a reason to keep the old values.

pkg/webhook/resources/setting/validator.go
brandboat: Will do, thanks for the comment 😃

bk201 (approved):

pkg/webhook/resources/setting/validator.go
```diff
@@ -536,7 +536,7 @@ func validateNTPServers(setting *v1beta1.Setting) error {
 		logrus.Errorf(errMsg)
 		return fmt.Errorf(errMsg)
 	}
-	startWithHTTP, err := regexp.Compile("^http?://.*")
+	startWithHTTP, err := regexp.Compile("^https?://.*")
```
Vicente-Cheng, line 539: I'm just curious about that. Does the NTP server need `https`?

If we use the local NTP server, could it pass?

pkg/webhook/resources/setting/validator_test.go
Vicente-Cheng: Yeah, as you checked, we will use the Fallback NTP server, so I thought the empty should be fine.

Vicente-Cheng (commented):
Just a nit. Please help to check with that.

BTW, I cannot remember why we just append instead of replace, and it looks like it should be fine with replace. So let fix the old mechanism!

pkg/webhook/resources/setting/validator.go
brandboat: > I'm just curious about that. Does the NTP server need https?

Users should use the FQDN format for NTP server settings. The validator will throw an error if the server address starts with http or https.

> If we use the local NTP server, could it pass?

Did you mean setting up a local NTP server, passing its address to the settings, and then checking timedatectl timesync-status to ensure it works?

Vicente-Cheng: Ah..., yes, I misunderstood something here.
There is no issue.

Thanks for it.

Vicente-Cheng (approved):
LGTM, thanks.

brandboat:
@Mergifyio backport v1.2 v1.3

PR #5885, opened 2024-05-28
HEP: Add migration harvester to standardized k8s API [CI SKIP]

Yu-Jack:

**Problem:**

Harvester currently use rancher/steve to build custom api actions, such as stop vm with `?action=stop` which is not standard k8s API. It's not convenient practice to use custom actions in k8s API.

**Solution:**

We will leverage the Kubernetes API Aggregation Layer, which proxies client requests to a specified API server defined with a custom resource `APIService`. This allows us to use the native Kubernetes client to invoke subresource APIs. 

Please take a look at HEP for more details.

**Related Issue:**
https://github.com/harvester/harvester/issues/1078

**Processing implementation PR**
- https://github.com/harvester/harvester/pull/5704 

**TODO after 1st review**

- [x] Complete List APIs
- [x] More details about upgrade strategy
- [x] More details about test plan

### DIFF
--- a/enhancements/20240528-migrate-harvester-api-to-standardized-k8s-api.md
+++ b/enhancements/20240528-migrate-harvester-api-to-standardized-k8s-api.md
@@ -230,6 +230,8 @@ But, each subresource might contain different HTTP method. It depends on the beh
 - If it's not idempotent, use `POST` or more matched HTTP method.
 - For list API, should use `GET` method. For example, `findMigratableNodes` action, it should be `GET` cause it doesn't change anything.
 
+We should also check which action is not used anymore, and remove it on new APIs. 
+
 #### Related Documentation Issue
 
 Currently, we didn't have any documentation for the action APIs, we only have k8s resource API instead.
@@ -244,7 +246,7 @@ Please check [api list](#api-list), and we need to test old and new APIs.
 There are two ways to test it:
 
 1. Test Entry Point
-   - The server shouldn't return 404 not found error for old and new APIs.
+   - The server shouldn't return 404 not found error for old and new endpoints.
 2. Test Functionality
    - The server should correctly handle the request.
 
@@ -258,12 +260,12 @@ However, it's an effort to maintain two different APIs, so eventually we need to
 
 1. (v1.4.0) Introduce the new APIs, and still support old APIs. Here, we could also introduce the new feature flag of setting resource to enable/disable old APIs. By default, it should be enabled at this moment. 
 2. (v1.5.0) Encourage the users to use the new APIs, and internal services/tools should use the new APIs.
-3. (v1.6.0)Old APIs are disabled by default, and we could turn it on by feature flag of setting resource.
+3. (v1.6.0) Old APIs are disabled by default, and we could turn it on by feature flag of setting resource.
 4. (v1.7.0) After a period of time, we could remove the old APIs and feature flag. Only keep new APIs.
 
 Those releases are just for reference, we could adjust it based on the progress of the migration.
 
-> Thanks @m-ildefons for the suggestion.
+> Thanks @m-ildefons for the milestone suggestion.
 
 In v1.4.0 harvester, we should accomplish first step at least.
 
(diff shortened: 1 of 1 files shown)


enhancements/20240528-migrate-harvester-api-to-standardized-k8s-api.md
```diff
@@ -0,0 +1,159 @@
...
+
+It's basically done because we've had the action business logic in the controller, we just need to create a new interface to fulfill two different API sources, which one is from steve and the other is from subresource api.
+
+### Test Plan
+
+Should test old and new APIs path formats, both should work as expected.
+
+### Upgrade strategy
+
+We won't remove old APIs format immediately, we will support both old and new APIs at the same time for backward compatibility.
```
m-ildefons, line 301: This API migration requires a deliberate and well-defined plan for which versions of Harvester will support the new scheme, which will retain backwards compatibility with the old scheme and which will eventually drop the old scheme.
This needs to be communicated with customers as well, since they potentially hook into the API and need both time and instructions to migrate to the new scheme.
The K8s API version skew policies and API deprecation policies are a good examples of how such an API migration can be undertaken and be communicated to users.

I'd like to see a more detailed upgrade strategy for this HEP that also includes a rough timeline (in release versions) for the milestones of the migration.

As an example:

v1.4.0: introduction of new API scheme, old API scheme still supported, internal services still use old API scheme
v1.5.0: all Rancher- and Harvester-internal services utilize the new API scheme, old API still available
v1.6.0: old API scheme disabled by default, can be enabled via setting.
v1.7.0: old API scheme deprecated, can not be enabled anymore. Only the new API scheme remains

enhancements/20240528-migrate-harvester-api-to-standardized-k8s-api.md
```diff
@@ -0,0 +1,159 @@
...
+POST /v1/harvester/kubevirt.io.virtualmachines/default/test?action=stop
+POST /v1/harvester/nodes/jacknode?action=cordon
+GET  /v1/harvester/upgradeslogs/default/test/download
+POST /v1/harvester/persistentvolumeclaims/default/test-disk-0-rlnlk?action=snapshot
+```
+
+The new api will be like this:
+
+```yaml
+POST /apis/subresources.harvesterhci.io/v1beta1/namespaces/default/virtualmachines/test/stop
```
m-ildefons, line 56: In places where Harvester utilizes CRDs from other projects, e.g. KubeVirt, there are such API paths already in place: e.g. https://kubevirt.io/api-reference/v1.2.1/operations.html#_v1alpha3stop
We don't need to create duplicate API paths for such actions. It's still a good idea to follow this approach for the Harvester CRDs.

It would be a good thing to compile a complete list of new API routes for this HEP. It would help in estimating the implementation effort and we could double check to make sure we're not duplicating existing API routes.

enhancements/20240528-migrate-harvester-api-to-standardized-k8s-api.md
```diff
@@ -0,0 +1,159 @@
...
+    ```
+2. Define the new routes path to accept following request path formats:
+   - `/apis/subresources.harvesterhci.io/v1beta1/namespaces/{namespace}/{resource}/{name}/{subresource}` 
+   - `/apis/subresources.harvesterhci.io/v1beta1/{resource}/{name}/{subresource}`
+
+It's basically done because we've had the action business logic in the controller, we just need to create a new interface to fulfill two different API sources, which one is from steve and the other is from subresource api.
+
+### Test Plan
+
+Should test old and new APIs path formats, both should work as expected.
```
m-ildefons: The test plan should specify more accurately the new API routes and their functions.
Additionally it should specify which old API routes are expected to be removed in the future.
For the time period where both API schemes are available in parallel, the tests would need to ensure that both API schemes work as intended.

m-ildefons (changes requested):
This is a good HEP and I particularly like the initiative to employ a standardized approach in line with other K8s resources.

However, I think this HEP falls a bit short since it only covers adding the new API routes and no plan to migrate away from the old routes. Additionally it is missing a complete list of new API routes and it doesn't consider that CRDs inherited from projects like KubeVirt may already offer subresources.

enhancements/20240528-migrate-harvester-api-to-standardized-k8s-api.md
bk201: This is important; QA also needs time to migrate the existing codes.

Yu-Jack:
@m-ildefons According to the suggestion, I've advised the HEP. Thanks for reviewing.

enhancements/20240528-migrate-harvester-api-to-standardized-k8s-api.md
ibrokethecloud: we should definitely start with having a list of apis exposed by harvester and split them into ones which will be moved to harvester subresources and the ones which will be replaced with kubevirt native subresources. We will also need to document the impact of the same on the terraform provider and machine driver if there is any.

Issue #5874, opened 2024-05-24
[BUG] Harvester Kubeconfig Expires After 30 Days (No External Rancher Integration)


irishgordo:
**Describe the bug**
The Harvester Kubeconfig expires after 30 days.
When you download the Harvester Kubeconfig from the "support" portion of Harvester.

**To Reproduce**
1. Clone this repo: https://github.com/irishgordo/fake-ntp-thingy
1. Run `docker compose up` (or `-d` for detatched)
1. Check that it gives you back a date of 35 days prior from today, from your host machine to the container `ntpdate -q 192.168.9.163`, replace `192.168.9.163` with your host machine's ipv4 -> You can of course modify this date by setting in the Dockerfile `ENV FAKETIME="-Nd"` where "N" would represent the # of days we want to shift the date back, as we fake a date on the container running the ntp server `chrony` and cut `chrony` off from the outside world so it has no upstream ntp-servers
1. Install a Single-Node Harvester at v1.3-head (either ipxe-examples or iso) just be sure to set the NTP server to be your host machine
1. Once Harvester is installed check that it's using the NTP server via: `systemctl status systemd-timesyncd.service` - you should see it is pointing to the IPV4 address of your host machine
1. Goto Harvester Dashboard -> Support -> Download KubeConfig
1. Ensure the Kubeconfig works, eg `kubectl get nodes` when using the Kubeconfig
1. Then in Harvester, modify the setting of the ntp-server -> change it to `time1.google.com` or another NTP server of your desire
1. Ensure that when you look at on the single-node Harvester cluster's `cat /etc/systemd/timesync.conf` you should see that it has added the new "time server" to the file and would read something like `NTP = 192.168.9.163 time1.google.com` - you may need to query it a few times before the new time-server shows up 
1. Because it seems, that the NTP setting in Harvester is only additive, we need to modify the `/etc/systemd/timesync.conf`, we will modify it to remove the IPv4 address of your host machine so that it only reads `NTP = time1.google.com` 
1. Once modified we need to restart the systemd process, `systemctl restart systemd-timesyncd.service`
1. You should then see that it will resync, noting `systemctl status systemd-timesyncd.service`
1. Then refresh the GUI Harvester Dashboard, you'll see that it requires you to log in again
1. Then attempt, to use the Kubeconfig you downloaded, you will note, that it is broken when you try to do a `kubectl` operation:
```
kubectl get nodes
[...]

PR #5896, opened 2024-05-29
expose kubeconfig ttl setting in harvester

ibrokethecloud:

**Problem:**

Rancher 2.8.x introduced a new setting for kubeconfig ttl. The default value for the setting is 30 days. As a result admin kubeconfig generated from Harvester expires in 30 days since this leverages the embedded rancher.

**Solution:**

The PR exposes the same kubeconfig ttl setting via harvester, with the harvester value taking precedence over rancher setting.

**Related Issue:**
https://github.com/harvester/harvester/issues/5874
**Test plan:**

### DIFF
--- a/pkg/controller/master/setting/kubeconfigttl.go
+++ b/pkg/controller/master/setting/kubeconfigttl.go
@@ -7,18 +7,35 @@ import (
 	harvSettings "github.com/harvester/harvester/pkg/settings"
 )
 
+const (
+	AuthTokenMaxTTLSettinName = "auth-token-max-ttl-minutes"
+)
+
 func (h *Handler) syncKubeconfigTTL(setting *harvesterv1.Setting) error {
 	rancherKubeconfigTTLSetting, err := h.rancherSettingsCache.Get(harvSettings.KubeconfigDefaultTokenTTLMinutesSettingName)
 	if err != nil {
 		return fmt.Errorf("error fetching setting %s: %v", harvSettings.KubeconfigDefaultTokenTTLMinutesSettingName, err)
 	}
+	rancherAuthTokenMaxTTLSetting, err := h.rancherSettingsCache.Get(AuthTokenMaxTTLSettinName)
+	if err != nil {
+		return fmt.Errorf("error fetching setting %s: %v", AuthTokenMaxTTLSettinName, err)
+	}
 
 	// if a custom ttl is set in harvester
 	if len(setting.Value) > 0 {
 		rancherKubeconfigTTLSetting.Value = setting.Value
+		rancherAuthTokenMaxTTLSetting.Value = setting.Value
 	} else { // apply default setting
 		rancherKubeconfigTTLSetting.Value = setting.Default
+		rancherAuthTokenMaxTTLSetting.Value = setting.Default
 	}
-	_, err = h.rancherSettings.Update(rancherKubeconfigTTLSetting)
-	return err
+
+	if _, err := h.rancherSettings.Update(rancherKubeconfigTTLSetting); err != nil {
+		return fmt.Errorf("unable to update rancher setting %s: %v", rancherKubeconfigTTLSetting.Name, err)
+	}
+
+	if _, err := h.rancherSettings.Update(rancherAuthTokenMaxTTLSetting); err != nil {
+		return fmt.Errorf("unable to update rancher setting %s: %v", rancherAuthTokenMaxTTLSetting.Name, err)
+	}
+	return nil
 }
--- a/pkg/controller/master/setting/kubeconfigttl.go
+++ b/pkg/controller/master/setting/kubeconfigttl.go
@@ -8,9 +8,9 @@ import (
 )
 
 func (h *Handler) syncKubeconfigTTL(setting *harvesterv1.Setting) error {
-	rancherKubeconfigTTLSetting, err := h.rancherSettingsCache.Get(harvSettings.KubeconfigTTLSettingName)
+	rancherKubeconfigTTLSetting, err := h.rancherSettingsCache.Get(harvSettings.KubeconfigDefaultTokenTTLMinutesSettingName)
 	if err != nil {
-		return fmt.Errorf("error fetching setting %s: %v", harvSettings.KubeconfigTTLSettingName, err)
+		return fmt.Errorf("error fetching setting %s: %v", harvSettings.KubeconfigDefaultTokenTTLMinutesSettingName, err)
 	}
 
 	// if a custom ttl is set in harvester
--- a/pkg/controller/master/setting/kubeconfigttl.go
+++ b/pkg/controller/master/setting/kubeconfigttl.go
@@ -1,0 +1,24 @@
+package setting
+
+import (
+	"fmt"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	harvSettings "github.com/harvester/harvester/pkg/settings"
+)
+
+func (h *Handler) syncKubeconfigTTL(setting *harvesterv1.Setting) error {
+	rancherKubeconfigTTLSetting, err := h.rancherSettingsCache.Get(harvSettings.KubeconfigTTLSettingName)
+	if err != nil {
+		return fmt.Errorf("error fetching setting %s: %v", harvSettings.KubeconfigTTLSettingName, err)
+	}
+
+	// if a custom ttl is set in harvester
+	if len(setting.Value) > 0 {
+		rancherKubeconfigTTLSetting.Value = setting.Value
+	} else { // apply default setting
+		rancherKubeconfigTTLSetting.Value = setting.Default
+	}
+	_, err = h.rancherSettings.Update(rancherKubeconfigTTLSetting)
+	return err
+}
--- a/pkg/settings/settings.go
+++ b/pkg/settings/settings.go
@@ -52,7 +52,7 @@ var (
 	StorageNetwork                         = NewSetting(StorageNetworkName, "")
 	DefaultVMTerminationGracePeriodSeconds = NewSetting(DefaultVMTerminationGracePeriodSecondsSettingName, "120")
 	AutoRotateRKE2CertsSet                 = NewSetting(AutoRotateRKE2CertsSettingName, InitAutoRotateRKE2Certs())
-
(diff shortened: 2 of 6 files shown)


pkg/settings/settings.go
```diff
@@ -86,6 +86,7 @@ const (
 	SupportBundleExpirationSettingName                = "support-bundle-expiration"
 	NTPServersSettingName                             = "ntp-servers"
 	AutoRotateRKE2CertsSettingName                    = "auto-rotate-rke2-certs"
+	KubeconfigTTLSettingName                          = "kubeconfig-default-token-ttl-minutes"
```
votdev: I would suggest to use this name because it is more clear what it is about and it matches the naming schema of all other constants defined in that file.
```suggestion
	KubeconfigDefaultTokenTTLMinutesSettingName                          = "kubeconfig-default-token-ttl-minutes"
```

FrankYang0529: I think we need to add the setting, so we will have it by default.

```
KubeconfigTTL = NewSetting(KubeconfigTTLSettingName, "43200") // "43200" is default value in Rancher
```

FrankYang0529 (commented):
I think we also need to add webhook validation, so we don't send a wrong value to Rancher.

pkg/settings/settings.go
votdev: > I think we need to add the setting, so we will have it by default.
> 
> ```
> KubeconfigTTL = NewSetting(KubeconfigTTLSettingName, "43200") // "43200" is default value in Rancher
> ```

Exactly, i'm currently testing that.

votdev: This will auto-create the `Settings` resource, but it is still not visible in the Harvester `Settings` UI.

```
$ kubectl get settings kubeconfig-default-token-ttl-minutes -oyaml
default: "43200"
kind: Setting
metadata:
  annotations:
    harvesterhci.io/hash: d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f
  creationTimestamp: "2024-06-04T10:22:40Z"
  generation: 3
  name: kubeconfig-default-token-ttl-minutes
  resourceVersion: "5902896"
  uid: 73b81673-7c41-4d4f-851e-359f5aeebe57
status:
  conditions:
  - lastUpdateTime: "2024-06-04T10:23:26Z"
    status: "False"
    type: configured
```

m-ildefons (changes requested):
I think we should also expose the `auth-token-max-ttl-minutes` or have logic to adjust it automatically. Otherwise a user might try to set the `kubeconfig-default-token-ttl-minutes` to a value above 90 days but then they wouldn't see the desired effect, because the TTL of the token would still be bounded to 90 days by the `auth-token-max-ttl-minutes`.
This would be particularly annoying when a user decides that they want to set the `kubeconfig-default-token-ttl-minutes` to 0, because then they don't expect to have to deal with an expiring kubeconfig at all.

pkg/settings/settings.go
ibrokethecloud: the default setting of 0 is created on initialisation here: https://github.com/harvester/harvester/pull/5896/files#diff-f9454369fb0cd8a32b1a568916edfe94a77af272bf63bdc5dd63c976a200cf5dL38

ibrokethecloud: looks like setting got dropped while rebase of another PR 🤦 
i have added it back

ibrokethecloud:
@bk201 @m-ildefons should we just use this setting to sync both `auth-token-max-ttl-minutes` and `kubeconfig-default-token-ttl-minutes` together?

ibrokethecloud:
looks like the token ttl in the kubeconfig token is set to the lower of `auth-token-max-ttl-minutes` and `kubeconfig-default-token-ttl-minutes`. I have tested the following scenarios:

* scenario 1: `auth-token-max-ttl-minutes` is 10mins and `kubeconfig-default-token-ttl-minutes` is 0mins. 
The resulting kubeconfig token has ttl set to 10 mins and does stop working at 10 mins
```
apiVersion: management.cattle.io/v3
authProvider: local
current: false
description: Kubeconfig token
expired: false
expiresAt: ""
isDerived: true
kind: Token
lastUpdateTime: ""
metadata:
  creationTimestamp: "2024-06-05T01:37:08Z"
  generateName: kubeconfig-user-9lj8c
  generation: 1
  labels:
    authn.management.cattle.io/kind: kubeconfig
    authn.management.cattle.io/token-userId: user-9lj8c
    cattle.io/creator: norman
  name: kubeconfig-user-9lj8c88r4c
  resourceVersion: "1266865"
  uid: 1edfc983-31dd-419b-9a3e-b905947feb1c
token: -----
ttl: 600000
userId: user-9lj8c
userPrincipal:
  displayName: Default Admin
  loginName: admin
  me: true
  metadata:
    creationTimestamp: null
    name: local://user-9lj8c
  principalType: user
  provider: local
```

* scenario 2: `auth-token-max-ttl-minutes` is 0mins and `kubeconfig-default-token-ttl-minutes` is 90mins. 
The resulting token has a ttl of 90mins
```
apiVersion: management.cattle.io/v3
authProvider: local
current: false
description: Kubeconfig token
expired: false
expiresAt: ""
isDerived: true
kind: Token
lastUpdateTime: ""
metadata:
  creationTimestamp: "2024-06-05T01:57:07Z"
  generateName: kubeconfig-user-9lj8c
  generation: 1
  labels:
    authn.management.cattle.io/kind: kubeconfig
    authn.management.cattle.io/token-userId: user-9lj8c
    cattle.io/creator: norman
  name: kubeconfig-user-9lj8c5mrxc
  resourceVersion: "1282860"
  uid: a81ff594-3766-4bfc-b474-0b48fddbf034
token: -----
ttl: 5400000
userId: user-9lj8c
userPrincipal:
  displayName: Default Admin
  loginName: admin
  me: true
  metadata:
    creationTimestamp: null
    name: local://user-9lj8c
  principalType: user
  provider: local
```

* scenario 3: `auth-token-max-ttl-minutes` is 0mins and `kubeconfig-default-token-ttl-minutes` is 0mins. 
The resulting token has a ttl of 0, which is the desired result.

```
apiVersion: management.cattle.io/v3
authProvider: local
current: false
description: Kubeconfig token
expired: false
expiresAt: ""
isDerived: true
kind: Token
lastUpdateTime: ""
metadata:
  creationTimestamp: "2024-06-05T02:07:09Z"
  generateName: kubeconfig-user-9lj8c
  generation: 1
  labels:
    authn.management.cattle.io/kind: kubeconfig
    authn.management.cattle.io/token-userId: user-9lj8c
    cattle.io/creator: norman
  name: kubeconfig-user-9lj8cvnqmc
  resourceVersion: "1290901"
  uid: 7f1c510c-fd25-467d-a12a-bc831caa9444
token: --
ttl: 0
userId: user-9lj8c
userPrincipal:
  displayName: Default Admin
  loginName: admin
  me: true
  metadata:
    creationTimestamp: null
    name: local://user-9lj8c
  principalType: user
  provider: local
```

m-ildefons:
> @bk201 @m-ildefons should we just use this setting to sync both `auth-token-max-ttl-minutes` and `kubeconfig-default-token-ttl-minutes` together?

We could. There are two possible approaches here:

1) Expose both settings write some documentation or link to the Rancher docs and let the user deal with it.

2) Automatically adjust the maximum-TTL setting with some logic like:
```
if (kubeconfig-default-token-ttl-minutes > 0) {
  auth-token-max-ttl-minutes = max(kubeconfig-default-token-ttl-minutes, auth-token-max-ttl-minutes)
} else {
  auth-token-max-ttl-minutes = 0
}
```

ibrokethecloud:
I was more in favour of always keeping both in sync. The proposal was to keep the default to 0, as anything else would be a change from current behaviour.

pkg/controller/master/setting/kubeconfigttl.go
```diff
@@ -0,0 +1,41 @@
...
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	harvSettings "github.com/harvester/harvester/pkg/settings"
+)
+
+const (
+	AuthTokenMaxTTLSettinName = "auth-token-max-ttl-minutes"
+)
+
+func (h *Handler) syncKubeconfigTTL(setting *harvesterv1.Setting) error {
```
m-ildefons, line 14: If I understand the code correctly, this is only applied if the setting _changes_. Are we sure it's also applied during the installation/upgrade/setup correctly?

pkg/webhook/resources/setting/validator_test.go
```diff
@@ -446,3 +446,51 @@ func Test_validateNoProxy_2(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{Name: settings.KubeconfigDefaultTokenTTLMinutesSettingName},
+				Value:      "10",
+			},
+			expectedErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validateSupportBundleNodeCollectionTimeout(tt.args)
```
m-ildefons: Isn't this supposed to be `validateUpdateKubeConfigTTLSetting`?

m-ildefons (commented):
LGTM, but I have two nits/questions.

pkg/webhook/resources/setting/validator_test.go
ibrokethecloud: yes should be..

pkg/controller/master/setting/kubeconfigttl.go
ibrokethecloud: bootstrapSettings does the initial sync at start: https://github.com/harvester/harvester/pull/5896/files#diff-f9454369fb0cd8a32b1a568916edfe94a77af272bf63bdc5dd63c976a200cf5dR37-R40

votdev:
We need to document that behaviour somewhere that the modification of `kubeconfig-default-token-ttl-minutes` will also affect `auth-token-max-ttl-minutes` when the setting is changed, otherwise it will be confusing for the user.

bk201 (approved):
LGTM, works as expected.

m-ildefons (approved):
LGTM.

FrankYang0529 (approved):
LGTM. Thanks.

bk201:
@mergifyio backport v1.2 v1.3

votdev (approved):
LGTM

Issue #5900, opened 2024-05-29
[BUG] The `AddBuiltInNoProxy` function does not handle whitespace


votdev:
**Is your enhancement related to a problem? Please describe.**
When a user enters a `noProxy` setting like ` 10.0.0.0/8  ,  127.0.0.1 `, the [built-in noProxy configuration](https://github.com/harvester/harvester/blob/5ee0329f1a6acef48e52758d3caf1d19e3bac1aa/pkg/util/proxy.go#L9) will add duplicates, in this case `127.0.0.1`.

**Describe the solution you'd like**
When splitting the noProxy string for processing, whitespaces need to be removed beforehand. 

**Describe alternatives you've considered**
n/a

TachunLin:
@votdev, 

I tried first on `v1.3.0` Harvester to reproduce the original issue behavior. 
But after I set the noProxy setting in the Harvester proxy settings page  `10.0.0.0/8  ,  127.0.0.1`. 

There we did not find the setting have duplicate `127.0.0.1`.

  ![image](https://github.com/user-attachments/assets/abd37754-02bf-42e8-9d42-8d5d075913e5)
  
  ![image](https://github.com/user-attachments/assets/9012f525-8ed4-4ef7-b566-0a7c4712cda6)


May I know the correct place and steps to reproduce this issue. thank you.

votdev:
> @votdev,
> 
> I tried first on `v1.3.0` Harvester to reproduce the original issue behavior. But after I set the noProxy setting in the Harvester proxy settings page `10.0.0.0/8 , 127.0.0.1`.
> 
> There we did not find the setting have duplicate `127.0.0.1`.
> 
[...]

PR #5904, opened 2024-05-29
The `AddBuiltInNoProxy` function does not handle whitespace

votdev:
**Problem:**
When a user enters a noProxy setting like ` 10.0.0.0/8  ,  127.0.0.1 `, the [built-in noProxy configuration](https://github.com/harvester/harvester/blob/5ee0329f1a6acef48e52758d3caf1d19e3bac1aa/pkg/util/proxy.go#L9) will add duplicates, in this case `127.0.0.1`.

**Solution:**
Trim the parts of the split-ed `noProxy` string.

**Related Issue:**
https://github.com/harvester/harvester/issues/5900

**Test plan:**
The internal test cases have been adapted. 

It is not possible to test that from within the UI directly. You can only check it indirectly. See:

- https://github.com/harvester/harvester/issues/5900#issuecomment-2257951768
- https://github.com/harvester/harvester/issues/5900#issuecomment-2295886250

### DIFF
--- a/pkg/util/proxy.go
+++ b/pkg/util/proxy.go
@@ -3,6 +3,7 @@ package util
 import (
 	"strings"
 
+	gocommon "github.com/harvester/go-common"
 	"github.com/rancher/wrangler/pkg/slice"
 )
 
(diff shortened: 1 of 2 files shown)


mingshuoqiu (approved):
LGTM.

Yu-Jack (approved):
LGTM, thanks!

FrankYang0529 (approved):
LGTM. Thanks.

votdev:
@Mergifyio backport v1.2 v1.3

Issue #4935, opened 2024-01-02
[FEATURE] Create Support Toolkits


Vicente-Cheng:
**Is your feature request related to a problem? Please describe.**
Create support toolkits to enhance the progress of support.

Even though we already have support bundle kits, we still take some time to get the related information from different CRs.
We want to introduce the support toolkits to solve this point. We could easily know the problematic volume(VM), its VM(volume), and related replica status.
The main focus pointer would be
- upgrade
- storage related issues (this leads to IO issues or Non-healthy volume)

**Describe the solution you'd like**
This toolkit should be running in the live cluster or a support bundle.
And it should give us some quick advice like related issue numbers or doubt points.

**Describe alternatives you've considered**
None

**Additional context**
More details can be found on the HEP

Vicente-Cheng:
Should reference the document of [[DOC] Create troubleshooting diagrams](https://github.com/longhorn/longhorn/issues/7410#top). It will be helpful to generate the troubleshooting mechanism with tooling.

PR #5906, opened 2024-05-29
HEP: add support toolkit

Vicente-Cheng:

**Problem:**
Add a support toolkit to improve the support efficiency

**Solution:**
Add support toolkit

**Related Issue:**
https://github.com/harvester/harvester/issues/4935

**Test plan:**
No need


Yu-Jack (approved):
Overall, I think it's a good idea. 

I think we can combine this HEP and the ideals of https://github.com/harvester/harvester/issues/8436 and https://github.com/harvester/harvester/issues/6564 to make a comprehensive tool to analyze the cluster.

The picture in my mind is like:

- Some rules are used in admission webhook to prevent users from doing something wrong, like deleting default resources.
- Some rules are used in toolkit to detect, like the VM migration failed due to cd-rom VM during upgrades.
- A service health dashboard showing potential risk, like pre-check. Then users will know that before upgrading.

In short, there are two parts we can do

- Before the event - pre-check + admission webhook
- After the event- support toolkit

I think most rules are shared between these two parts.

Yu-Jack:
BTW, before implementing any features, we should have a discussion about which issues we want to prevent. Listing the tire 1 issues might help us get rid of a lot of issues from customers. I think this list will be a handbook for all team members.

PR #5922, opened 2024-05-31
ci: add daily build

bk201:
We no longer publish ISOs from the harvester-installer repo, so we do daily builds here.
Build ISOs twice a day to catch up the installer change earlier (in case there is no merge event in the harvester repo).


Vicente-Cheng (approved):
LGTM, thanks!

tserong (approved):

Issue #2108, opened 2022-04-01
[BUG] upgrade preload script won't preload the images again if it is already presented


guangbochen:
**Describe the bug**
The upgrade preload script will not preload the already displayed images again, e.g. the second upgrade uses the same image tag like `rancher/harvester-cluster-repo:v1.0`.

**To Reproduce**
1. A single-node upgrade is easier to test.
1. After done the image preloading process cancels the upgrade via deleting the related CR with `$kubectl delete upgrades.harvesterhc.io -n harvester-system --all`
2. patch the ISO image and upgrade it again with the same image tag, normally using the `master-head` or `v*-head` build.
3. the registry image should be reloaded correctly.

**Expected behavior**
Should check and load the image again if the image digest is different.

**Support bundle**

**Environment:**
 - Harvester ISO version: v1.0.1-rc2
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): All

bk201:
The reason is we only check the image name, for example, if `rancher/harvester-cluster-repo:v1.0` exists.
It's not an issue for normal upgrade paths, but might be a problem during development as the description says.

PR #5941, opened 2024-06-03
Add `harvesterhci.io/force-preload-import` to force import images in upgrade preload phase

brandboat:
**Problem:**
The upgrade preload script will not preload the already displayed images again, e.g. the second upgrade uses the same image tag like rancher/harvester-cluster-repo:master

**Solution:**
Introduce a new annotation `harvesterhci.io/force-preload-import`, set it to true to force all images loaded during upgrade pre-load phase. 

**Related Issue:**
https://github.com/harvester/harvester/issues/2108

**Test plan:**
1. Build harvester iso from local dev branch. e.g. `branch-foobar` and create harvester cluster with one node.
2. Use the following yaml to upgrade to a lastest version in master branch.
```yaml
kind: Version
metadata:
  name: master
  namespace: harvester-system
spec:
  isoURL: https://releases.rancher.com/harvester/master/harvester-master-amd64.iso
```

```yaml
kind: Upgrade
metadata:
  annotations:
    harvesterhci.io/force-preload-import: "true"
  generateName: hvst-upgrade-
  namespace: harvester-system
spec:
  version: "master"
```
3. Check the upgrade complete without any error.

### DIFF
--- a/package/upgrade/lib.sh
+++ b/package/upgrade/lib.sh
@@ -204,12 +204,14 @@ import_image_archives_from_repo() {
       archive_url="$UPGRADE_REPO_BUNDLE_ROOT/$archive"
       image_list_file="${tmp_image_archives}/$(basename $list)"
 
-      # Check if images already exist
-      curl -sfL $image_list_url | sort > $image_list_file
-      missing=$($CTR -n k8s.io images ls -q | grep -v ^sha256 | sort | comm -23 $image_list_file -)
-      if [ -z "$missing" ]; then
-        echo "Images in $image_list_file already present in the system. Skip preloading."
-        continue
+      if [ "$FORCE_PRELOAD_IMPORT" != "true" ]; then
+        # Check if images already exist
+        curl -sfL $image_list_url | sort > $image_list_file
+        missing=$($CTR -n k8s.io images ls -q | grep -v ^sha256 | sort | comm -23 $image_list_file -)
+        if [ -z "$missing" ]; then
+          echo "Images in $image_list_file already present in the system. Skip preloading."
+          continue
+        fi
       fi
 
       curl -fL $archive_url | zstdcat | $CTR -n k8s.io images import --no-unpack -
@@ -249,6 +251,7 @@ detect_upgrade()
 
   UPGRADE_PREVIOUS_VERSION=$(echo "$upgrade_obj" | yq e .status.previousVersion -)
   SKIP_VERSION_CHECK=$(echo "$upgrade_obj" | yq e '.metadata.annotations."harvesterhci.io/skip-version-check"' -)
+  FORCE_PRELOAD_IMPORT=$(echo "$upgrade_obj" | yq e '.metadata.annotations."harvesterhci.io/force-preload-import"' -)
 }
 
 # refer https://github.com/harvester/harvester/issues/3098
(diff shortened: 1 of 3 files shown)


starbops (commented):
Implementation details aside, I have two general questions:

1. I feel this is a bit overkill. IIRC, we only have one remaining workload with such an issue, and it is `harvester-cluster-repo`. Do we have to force-preload all the images?
2. I've tried manually patching the `harvester-cluster-repo` Deployment using a patch file with the same content (because the image tag does not change). It turned out the Deployment did not rollout:

   ```sh
   $ kubectl patch deployment harvester-cluster-repo -n cattle-system --patch-file ./cluster_repo.yaml --type merge
   deployment.apps/harvester-cluster-repo patched (no change)
   ```
   This is the [same step](https://github.com/harvester/harvester/blob/226f72aa58532071232ad548e27970c1cec55733/package/upgrade/upgrade_manifests.sh#L332-L340) that the apply-manifest Pod will take. As a result, the Deployment is still using the old image.

brandboat:
> I feel this is a bit overkill. IIRC, we only have one remaining workload with such an issue, and it is harvester-cluster-repo. Do we have to force-preload all the images?

The root cause is harvester-cluster-repo image version is always `master` in all dev branches under harvester repo, the ultimate way to solve this is to treat it like other images, add a specific dev tag to harvester-cluster-repo image instead of just use `master`. And this could be done after we merge `harvester` and `harvester-installer` into one repo. 

Back to the question, yes I think this is overkill, but I'm not sure if there is other places that also use the same image version across different dev branches, so I choose to force preload everything instead of adding a white list here. 
Also, there are too many annotations in the upgrade part.. perhaps we should not add this one, just go back to the workaround (though cumbersome).

> I've tried manually patching the harvester-cluster-repo Deployment using a patch file with the same content (because the image tag does not change). It turned out the Deployment did not rollout:

Actually it does roll out the old harvester-cluster-repo and create a new one ~~after applied the annotation (harvesterhci.io/force-preload-import)~~. And I'm not sure why... Will dig into the code and find out what did we do to make the rolling happen (Updated in https://github.com/harvester/harvester/pull/5941#issuecomment-2180939422)

brandboat:
I found that no matter I add `harvesterhci.io/force-preload-import` or not,  the code will rollout harvester-cluster-repo anyway. 

The original harvester-cluster-repo yaml is here

https://github.com/harvester/harvester-installer/blob/1e7572e3822c70adf5730d71ef94f071a15dde82/package/harvester-os/templates/91-harvester-bootstrap-repo.yaml

and [this line](https://github.com/harvester/harvester/blob/226f72aa58532071232ad548e27970c1cec55733/package/upgrade/upgrade_manifests.sh#L332-L340) will erase the ports in `harvester-cluster-repo` and trigger the deployment rolling... 
```
        ports:
        - containerPort: 80
          protocol: TCP
```

That's why even if the harvester-cluster-repo image version in the upgrade version is the same as current, the roll out still happen during upgrade manifests.

brandboat:
close this since this is not the right approach

starbops:
> I found that no matter I add harvesterhci.io/force-preload-import or not, the code will rollout harvester-cluster-repo anyway.

Good finding!

As a side note, [here](https://github.com/harvester/harvester/blob/226f72aa58532071232ad548e27970c1cec55733/package/upgrade/upgrade_manifests.sh#L340) we should use the strategic merge patch instead. Anyway, removing the container port harms nothing.

brandboat:
> As a side note, [here](https://github.com/harvester/harvester/blob/226f72aa58532071232ad548e27970c1cec55733/package/upgrade/upgrade_manifests.sh#L340) we should use the strategic merge patch instead. Anyway, removing the container port harms nothing.

Thanks, I'll file another pr to address this 👍

Issue #5724, opened 2024-05-02
[FEATURE] Host local storage support - Experimental 


bk201:
**Is your feature request related to a problem? Please describe.**

Applications require large volume or high-performance local disks. Live-migration is not a concern.

**Describe the solution you'd like**

As a HCI solution, we can have a build-in local storage support like leveraging LVM or files.
We have experimented lvm-csi-driver in a POC environment.

**Describe alternatives you've considered**

**Additional context**

SlavikCA:
I'm curious: how that feature will be different from simply using `hostPath`?

Is correct to say, that `hostPath` is file-level access and this feature will provide block-level access?

Or is it going to be file-level access, but the disc (volume group?) will be managed by the LVM CSI driver, so it will provide features such mirroring, striping, etc...?

Vicente-Cheng:
> I'm curious: how that feature will be different from simply using `hostPath`?
> 
> Is correct to say, that `hostPath` is file-level access and this feature will provide block-level access?
> 
> Or is it going to be file-level access, but the disc (volume group?) will be managed by the LVM CSI driver, so it will provide features such mirroring, striping, etc...?

Hi @SlavikCA,

This feature will provide the block device.
With LVM, we can easily use various LVM types like mirror and stripe. Also, we would benefit from various LVM features like expansion/snapshot.

abonillabeeche:
FYI - I had the LVM CSI driver installed on 1.3.0 and upgraded to 1.3.1 without an issue. Perhaps worth investigating is the ability to have snapshots and backup, as it is not available out of the box.
<img width="1172" alt="Screenshot 2024-06-19 at 9 51 06 AM" src="https://github.com/harvester/harvester/assets/5146963/880bd273-155a-4b45-aa37-ab8de43517a0">

abonillabeeche:
> Or is it going to be file-level access, but the disc (volume group?) will be managed by the LVM CSI driver, so it will provide features such mirroring, striping, etc...?

Correct. The LVM CSI provides stripping which could be used for large volumes.

albinsun:
Hi Guys,
Per discussed in [HEP PR](https://github.com/harvester/harvester/pull/5956#issuecomment-2228830404), Longhorn plans to support [LVM-based Local Volume](https://github.com/longhorn/longhorn/issues/4935) in `v1.8` (Nov. 2024) or `v1.9` (Mar. 2025) and we will use LH solution for local storage that time.  

[...]

PR #5956, opened 2024-06-04
HEP: introduce the LVM CSI driver

Vicente-Cheng:

**Problem:**
Introduce the LVM CSI Driver

**Solution:**
The HEP would help the design review

**Related Issue:**
https://github.com/harvester/harvester/issues/5724

**Test plan:**


enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,77 @@
...
+We would like to start with the existing LVM CSI driver and introduce it to the Harvester. We need to make some changes to make it more suitable for the Harvester. After that, we could use the LVM CSI driver to create the LVM volume in Harvester.
+
+### User Stories
+
+#### The user creates the PCI pass-through VM and wants a higher performance.
+When users use the vGPU feature with PCI pass-through, it means they will redirect the VM to a specific host. Also, the user wants a higher-performance disk for the VM. The LVM CSI driver could help users create the PVC with an LVM volume on a local disk. That could help users have better performance and latency.
+
+### User Experience In Detail
+
+The user creates a VM and starts with the vGPU feature. Then, the user creates the PVC with the LVM volume. The user attaches the PVC to the VM through hotplug and uses the LVM volume in the VM.
```
bk201: we can remove the vGPU here, the feature is not restricted to vGPU VMs.

enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,77 @@
...
+
+We do not have any current solution to support local volumes, which means directly using the local disk as the volume. The LVM CSI driver could help leverage the local disk as the volume. If the user could use the strict location VM, it would help us improve performance and latency.
+
+### Goals
+
+Introduce the LVM CSI driver to support the LVM volume in Harvester.
+
+### Non-goals [optional]
+
+Support volume expansion and snapshot for the LVM volume in the first version.
```
bk201: add guest pvc if not supported

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: updated!

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: Thanks for the reminder, updated!

enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Motivation
+
+We do not have any current solution to support local volumes, which means directly using the local disk as the volume. The LVM CSI driver could help leverage the local disk as the volume. If the user could use the strict location VM, it would help us improve performance and latency.
+
+### Goals
+
+Introduce the LVM CSI driver to support the LVM volume in Harvester.
+
+### Non-goals [optional]
```
tserong: Remove the word "[optional]" here, as we are explicitly stating some non-goals :-)

enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Proposal
+
+We would like to start with the existing LVM CSI driver and introduce it to the Harvester. We need to make some changes to make it more suitable for the Harvester. After that, we could use the LVM CSI driver to create the LVM volume in Harvester.
+
+### User Stories
+
+#### The user creates the PCI pass-through VM and wants a higher performance.
+When users use the features with PCI pass-through, it means they will redirect the VM to a specific host. Also, the user wants a higher-performance disk for the VM. The LVM CSI driver could help users create the PVC with an LVM volume on a local disk. That could help users have better performance and latency.
+
```
tserong, line 36: Is PCI pass-through required? The description in https://github.com/harvester/harvester/issues/5724 just says "Applications require large volume or high-performance local disks. Live-migration is not a concern." AFAICT it would be enough to create a VM with Node Scheduling set to "Run VM on specific node - (Live migration is not supported", then specify the node to run on, matching the node with the LVM volume.

enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,77 @@
...
+### Implementation Overview
+
+A overview of the design could be checked below:
+![lvm-csi-driver-arch](20240604-lvm-csi-driver/lvm-csi-driver-arch.png)
+
+### LVM CSI Driver
+At first, we consider two existing solutions: [metal-stack/lvm-csi-driver](https://github.com/metal-stack/csi-driver-lvm) and [topolvm/topolvm](https://github.com/topolvm/topolvm). Both solutions support the CSI topology feature, and other features like volume capability and snapshot are only supported in the topolvm. However, the topolvm is more complex than the lvm-csi-driver. We could try to add the capability and snapshot feature to the lvm-csi-driver and keep it as simple as we want.
+
+So, we chose the [metal-stack/lvm-csi-driver](https://github.com/metal-stack/csi-driver-lvm) as the base LVM CSI driver. This project is lightweight and already has the essential capability for our requirements. We want to introduce it in the first version. Then, we can add more features based on the requirements.
```
tserong:
```suggestion
So, we chose the [metal-stack/csi-driver-lvm](https://github.com/metal-stack/csi-driver-lvm) as the base LVM CSI driver. This project is lightweight and already has the essential capability for our requirements. We want to introduce it in the first version. Then, we can add more features based on the requirements.
```

enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Design
+
+### Implementation Overview
+
+A overview of the design could be checked below:
+![lvm-csi-driver-arch](20240604-lvm-csi-driver/lvm-csi-driver-arch.png)
+
+### LVM CSI Driver
+At first, we consider two existing solutions: [metal-stack/lvm-csi-driver](https://github.com/metal-stack/csi-driver-lvm) and [topolvm/topolvm](https://github.com/topolvm/topolvm). Both solutions support the CSI topology feature, and other features like volume capability and snapshot are only supported in the topolvm. However, the topolvm is more complex than the lvm-csi-driver. We could try to add the capability and snapshot feature to the lvm-csi-driver and keep it as simple as we want.
```
tserong:
```suggestion
At first, we consider two existing solutions: [metal-stack/csi-driver-lvm](https://github.com/metal-stack/csi-driver-lvm) and [topolvm/topolvm](https://github.com/topolvm/topolvm). Both solutions support the CSI topology feature, and other features like volume capability and snapshot are only supported in the topolvm. However, the topolvm is more complex than the csi-driver-lvm. We could try to add the capability and snapshot feature to the csi-driver-lvm and keep it as simple as we want.
```

enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,77 @@
...
+
+In the initial version, we propose a significant change that will enhance user flexibility. We plan to allow users to predefine the volume group. The provisioner will then use this predefined volume group and logical volume type (e.g., mirror, stripe) to create the volume. This change will enable users to group similar types of disks together, thereby ensuring optimal performance and capability.
+
+With this change, the StorageClass will now contain two additional parameters: `vgName` and `lvType`. The `vgName` refers to the volume group name, and the `lvType` specifies the logical volume type. These parameters are essential for the LVM provisioner to create the volume according to the user's specifications.
+
+### Node Disk Manager
+Although we can create a volume group manually, the node disk manager can help us do so. The node disk manager will create the volume group based on the disks the user chooses. Then, the user can use the volume group name to create StorageClass and PVC.
```
tserong: How does the user tell NDM which disks to use to create the VG?

tserong (commented):
Some comments/suggestions/questions inline

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: I gave a link to the NDM refactor, which contains more details.

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: updated, thanks!

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: updated, thanks!

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: Thanks for the suggestion. I rewrote this part for the feature with PCI passthrough.

But for the Detail experience, I still use that as an example. WDYT?

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: Cool, thanks!  Updated!

bk201:
@Vicente-Cheng can we design the feature as an add-on? For users who don't need this feature, they don't need to have the overhead in the cluster. And we can promote the feature to be enabled by default if it's widely adopted.

enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,81 @@
...
+
+In the initial version, we propose a significant change that will enhance user flexibility. We plan to allow users to predefine the volume group. The provisioner will then use this predefined volume group and logical volume type (e.g., mirror, stripe) to create the volume. This change will enable users to group similar types of disks together, thereby ensuring optimal performance and capability.
+
+With this change, the StorageClass will now contain two additional parameters: `vgName` and `lvType`. The `vgName` refers to the volume group name, and the `lvType` specifies the logical volume type. These parameters are essential for the LVM provisioner to create the volume according to the user's specifications.
+
+### Node Disk Manager
+Although we can create a volume group manually, the node disk manager can help us do so. The node disk manager will create the volume group based on the disks the user chooses. Then, the user can use the volume group name to create StorageClass and PVC. You can refer https://github.com/harvester/harvester/pull/6015 to find more details about the node disk manager how to create the volume group.
```
bk201, line 109: Does this mean when a user adds a disk to a VG, the VG is created (where should the user specify the VG's attributes)? And the sequential disk adds will extend the VG?

Or should we design a new CRD as a container for disks?

Vicente-Cheng:
> @Vicente-Cheng can we design the feature as an add-on? For users who don't need this feature, they don't need to have the overhead in the cluster. And we can promote the feature to be enabled by default if it's widely adopted.

I do not have any strong opinion about that.
We could make the first version available via the add-on feature.

For the additional effort, we need to add the checking mechanisms to the NDM to check the status of the add-on feature.

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: From the current behavior (see https://github.com/harvester/node-disk-manager/pull/109/commits/b2080f3179790384bfe806bf7cb3e43c8b7320c3).
If the VG does not exist, we will create the VG. If the VG is already here, we will extend the VG.
For the parameter, I will create a way for people who want to customize the VG.

> Or should we design a new CRD as a container for disks?

I do not think creating a new CRD is a good way because that will make things more complex.
We assume the user knows how VG/LV works. (Or rely on the document).

However, I agree that we should refine the UI flow to allow users to configure various storage backend disks.

bk201: Can you help clarify if extending the VG involves data movement or unexpected data placement?
Say a user has a VG with two disks and creates a PVC (a stripped LV). Its data are spread into two disks.
If the user later adds a new disk to the VG, will the existing PVC continue spreading data on the first two disks? And do all new PVCs created later spread data on 3 disks?

My point here is when users create PVCs from a storage class, they expect all volumes to have the same characteristics, like availability and performance.

The HEP has to clarify the supported LV types like linear/stripped/RAID/thin provision.

enhancements/20240604-lvm-csi-driver.md
```diff
@@ -0,0 +1,81 @@
...
+
+The LVM CSI driver includes the CSI plugin and the LVM provisioner. The CSI plugin handles all CSI requests, while the LVM provisioner manages LVM operations such as creation, deletion, expansion, and snapshotting. This division of responsibilities ensures efficient and effective management of our storage infrastructure.
+
+In the initial version, we propose a significant change that will enhance user flexibility. We plan to allow users to predefine the volume group. The provisioner will then use this predefined volume group and logical volume type (e.g., mirror, stripe) to create the volume. This change will enable users to group similar types of disks together, thereby ensuring optimal performance and capability.
+
+With this change, the StorageClass will now contain two additional parameters: `vgName` and `lvType`. The `vgName` refers to the volume group name, and the `lvType` specifies the logical volume type. These parameters are essential for the LVM provisioner to create the volume according to the user's specifications.
```
bk201, line 61: >and the `lvType` specifies the logical volume type.

Also type-specific parameters. We need to list all supported types and their parameter in the HEP.

bk201: Do we need to specify the `nodeName` here? Or do all nodes containing a VG with the name provide volume provisioning?

enhancements/20240604-lvm-csi-driver.md
WebberHuang1118: > Can you help clarify if extending the VG involves data movement or unexpected data placement? Say a user has a VG with two disks and creates a PVC (a stripped LV). Its data are spread into two disks. If the user later adds a new disk to the VG, will the existing PVC continue spreading data on the first two disks? And do all new PVCs created later spread data on 3 disks?
> 

I just checked with my env, TL;DR, extending VG doesn't involve data movement, so the existing PVC (LV) will only be allocated on the original two PVs.

Following are the experiment details:
- Three existing pvs: `vdb, vdc, vdd` and only `vdb, vbc` in vg `my_vg`
  - $ pvs
        ```
          PV         VG    Fmt  Attr PSize   PFree  
        /dev/vdb   my_vg lvm2 a--  <10.00g <10.00g
        /dev/vdc   my_vg lvm2 a--  <10.00g <10.00g
        /dev/vdd         lvm2 ---   10.00g  10.00g
        ```
- The existing lv `my_lv` is located on `vdb(252:16)` and `vdc(252:32)`
  - $ dmsetup ls --tree
      ```
      my_vg-my_lv (253:0)
       ├─ (252:32)
       └─ (252:16)
      ```
- Extend vg `my_vg` with pv `vdd`
  - $ vgextend my_vg /dev/vdd
- The existing lv my_lv is still located on the original disks
  - $ dmsetup ls --tree
      ```
      my_vg-my_lv (253:0)
       ├─ (252:32)
       └─ (252:16)
      ``` 
- The new LV could utilize the three pvs, if specify strip size as three
  - $ lvcreate -i 3 -I 4 -L 5G -n my_lv2 my_vg
  - $ dmsetup ls --tree
```
        my_vg-my_lv2 (253:1)
         ├─ (252:48)
         ├─ (252:32)
         └─ (252:16)
```

WebberHuang1118: If a new disk is added to the vg (e.q. by `vgextend`), will the parameter of `lvType` change accordingly? 

For example, if there are originally two disks, and add another new disk, will the strip number automatically change from two to three?

Vicente-Cheng: Thanks @WebberHuang1118,

Hi @bk201 
As @WebberHuang1118  mentioned, the old PVC still uses the number of PV on the creation instead of the new PV count (after the VG is extended).

The new LV will use the latest PV count of the corresponding VG. We will check the latest PV count of the VG when we create the LV. For the above example, the strip number will be three after extending the VG.

enhancements/20240604-lvm-csi-driver.md
Vicente-Cheng: >Also type-specific parameters. We need to list all supported types and their parameter in the HEP.

Updated, thanks!

>Do we need to specify the nodeName here? Or do all nodes containing a VG with the name provide volume provisioning?

I am still thinking about that. Currently, we will create the LV with the corresponding node affinity.
If the node had the same vgName, the LV could be created on there. But that will introduce another issue, we did not support PVC migrate (no matter online or offline).

If we did not specific the nodeName, I am afraid some user will get confused.

albinsun:
Hmm.. Local (affinity) storage seems a general requirement.
Does Longhorn plan to (or already) has similiar feature?
Should we have a sync with LH and avoid to rebuild the wheel for both projects.

Ref. https://longhorn.io/docs/1.6.2/best-practices/#disk-space-management

WebberHuang1118:
LGTM, Thanks for the HEP.

Just one point I want to double confirm,

we plan to provide LV type `linear`, `striped`, and `mirrored`,
I'm considering the data layout can be a little complicated.

For example:
- There are two 100G disks `sda` and `sdb` on one node
- User creates one `linear` `70G` LV
  - This LV is allocated on `sda`
- User adds a new 100G disk `sdc`
- User tries to create another `180G` `stripped` LV
- LVM CSI would try to create a stripped LV with `strip number 3`
  - LVM will try to allocate 60G data on each disk
  - The LV creation seems not to make it because there is only `30G` free data left

@Vicente-Cheng is my understanding correct? 
If yes, this case seems to be not easy for users to figure out. Thanks.

Vicente-Cheng:
> LGTM, Thanks for the HEP.
> 
> Just one point I want to double confirm,
> 
> we plan to provide LV type `linear`, `striped`, and `mirrored`, I'm considering the data layout can be a little complicated.
> 
> For example:
> 
> * There are two 100G disks `sda` and `sdb` on one node
> * User creates one `linear` `70G` LV
>   
>   * This LV is allocated on `sda`
> * User adds a new 100G disk `sdc`
> * User tries to create another `180G` `stripped` LV
> * LVM CSI would try to create a stripped LV with `strip number 3`
>   
>   * LVM will try to allocate 60G data on each disk
>   * The LV creation seems not to make it because there is only `30G` free data left
> 
> @Vicente-Cheng is my understanding correct? If yes, this case seems to be not easy for users to figure out. Thanks.

Hi @WebberHuang1118
Thanks for the suggestions and questions.
I updated the `Supported LV type` section to conclude our afternoon discussion.
Please help to recheck this part.

Hi folks, @WebberHuang1118 and I take some time to discuss the supported lv type.
We all agree that keeping the data layout simple is a good way. Also, we discuss the `dm-thin` and its pros and cons.

You can find more info in the `Supported LV type` section.
Thanks @WebberHuang1118 for providing the various and very helpful viewpoints.

Vicente-Cheng:
> Hmm.. Local (affinity) storage seems a general requirement. Does Longhorn plan to (or already) has similiar feature? Should we have a sync with LH and avoid to rebuild the wheel for both projects.
> 
> Ref. https://longhorn.io/docs/1.6.2/best-practices/#disk-space-management

Yeah, Longhorn plans to provide this capability, maybe v1.8 or v1.9.
In that time, we will use their solution for local storage.

Vicente-Cheng:
I would like to get more feedback on the user experience regarding the node affinity.

We rely on the node affinity when the Volume has the first consumer now.
After the PV records the node affinity, we will know which node to schedule.

For the StorageClass, nodeName is required or not?
We assume that the VM workload that wants to use the LVM volume will need to be fixed on a specific node.
That means the VM creation will have the node affinity, so we do not need the target node info on the StorageClass.
That's why we did not introduce the node info on the StorageClass.

So, the question is, if the VM workload is not fixed on the specific node, the LVM volume will have a chance of provision failure.  That scenario would be resolved if we had defined the nodeName on the StorageClass.

However, if we support nodeName on the StorageClass, that also introduces a potential conflict between VM node affinity and StorageClass parameters.
 
What did you think about the nodeName parameter with the StorageClass?
Or I got something misunderstanding here?
cc @WebberHuang1118, @ibrokethecloud, @tserong, @albinsun, @lanfon72

albinsun:
> Yeah, Longhorn plans to provide this capability, maybe v1.8 or v1.9. In that time, we will use their solution for local storage.

Thanks I saw it, Longhorn plans to have this in `v1.9` (Mar. 2025).
https://github.com/longhorn/longhorn/wiki/Roadmap#longhorn-v19-march-2025

If so, do we still need to make this feature by ourself as LH will support it?

WebberHuang1118:
> I would like to get more feedback on the user experience regarding the node affinity.
> 
> We rely on the node affinity when the Volume has the first consumer now. After the PV records the node affinity, we will know which node to schedule.
> 
> For the StorageClass, nodeName is required or not? We assume that the VM workload that wants to use the LVM volume will need to be fixed on a specific node. That means the VM creation will have the node affinity, so we do not need the target node info on the StorageClass. That's why we did not introduce the node info on the StorageClass.
> 
> So, the question is, if the VM workload is not fixed on the specific node, the LVM volume will have a chance of provision failure. That scenario would be resolved if we had defined the nodeName on the StorageClass.
> 
> However, if we support nodeName on the StorageClass, that also introduces a potential conflict between VM node affinity and StorageClass parameters.
> 
> What did you think about the nodeName parameter with the StorageClass? Or I got something misunderstanding here? cc @WebberHuang1118, @ibrokethecloud, @tserong, @albinsun, @lanfon72

Hi @Vicente-Cheng,

May I know how do you plan to leverage the `nodeName` parameter in the storageclass? By controlling the node to [deploy the `provisioner pod`](https://github.com/harvester/csi-driver-lvm/blob/840ec7c758ebc733a1c513aaec7c9e94a847edfb/pkg/lvm/controllerserver.go#L144-L147)?

lanfon72:
> I would like to get more feedback on the user experience regarding the node affinity.
> 
> We rely on the node affinity when the Volume has the first consumer now. After the PV records the node affinity, we will know which node to schedule.
> 
> For the StorageClass, nodeName is required or not? We assume that the VM workload that wants to use the LVM volume will need to be fixed on a specific node. That means the VM creation will have the node affinity, so we do not need the target node info on the StorageClass. That's why we did not introduce the node info on the StorageClass.
> 
> So, the question is, if the VM workload is not fixed on the specific node, the LVM volume will have a chance of provision failure. That scenario would be resolved if we had defined the nodeName on the StorageClass.
> 
> However, if we support nodeName on the StorageClass, that also introduces a potential conflict between VM node affinity and StorageClass parameters.
> 
> What did you think about the nodeName parameter with the StorageClass? Or I got something misunderstanding here? cc @WebberHuang1118, @ibrokethecloud, @tserong, @albinsun, @lanfon72

Consider the existing case of creating VM:
image

there would have 2 cases:
1. User create a VM with existing LV volume, the VM should be scheduled on the node serving LVM.
2. User create a VM with new volume (the storage class is for LVM), the VM should be scheduled on the node serving LVM.

As we can simply select the VM Network to make the VM scheduling on nodes, the UI could keep the same behavior for LV volumes.

My concept of the feature would like:
1. Host control PVs and VGs
2. Storage class control VG and LVs
3. LV volumes works like normal volumes but no replicas

Consider that in normal PC we can create multiple VGs and expand PVs/VGs, I thought the storage class will only map to a specific VG would make the concept simply to explain the feature to users.

Vicente-Cheng:
Hi guys, I updated the HEP on the implementation.
Please help to recheck it. Thanks!

For the above discussion. Now StorageClass for the LVM would leverage the nodeName with allowedTopologies fields. With that, if the VM starts with the volume from the LVM storageClass, it will be scheduled to the correct node because of the `allowedTopologies`. It would help us to use this volume more efficiently.

Issue #4888, opened 2023-12-19
[ENHANCEMENT] Improve the UX of non-migratable VMs in node maintenance mode


bk201:
**Is your enhancement related to a problem? Please describe.**

When a node is put into maintenance, Harvester will start to live-migrate VMs on it. In some cases, VMs can't migrate to other node:
- VMs have PCI passthrough devices.
- VMs have NodeSelector or zone constraints, but can't find another node to schedule.

And these VMs will stay migrating until timeout happens.

**Describe the solution you'd like**

Improving the UX. Some ideas:

- Display a non-migratable VM list and suggest those VMs will be shut down.
- Once the user confirms, we shut down the VMs and perform maintenance.

**Describe alternatives you've considered**

**Additional context**

megabreit:
Is this just an UI issue? Currently only 2 life migration are done in parallel. And if 2 migrations are in progress (and never finish), the rest of the VMs, even if they can be migrated, will not be moved to other hypervisor nodes.

bk201:
> Is this just an UI issue? Currently only 2 life migration are done in parallel. And if 2 migrations are in progress (and never finish), the rest of the VMs, even if they can be migrated, will not be moved to other hypervisor nodes.

No. the `require-ui` label is just for indicating there is UI work involved. So one proposal in the issue is to warn the user a VM can't live-migrate if the user tries to migrate it or put a node into maintenance.

torchiaf:
@bk201 @ibrokethecloud is there a chance that `force` option will shoutdown all non-migratable VMs ?
- If so, we could add a tooltip to the force checkbox or a message, to inform the users that they can use force checkbox to shutdown the VMs If first try fails.
- else we need a new checkbox, visible on errors, to shutdown the VMs.

What do you think ?

cc @a110605

a110605:
> @bk201 @ibrokethecloud is there a chance that `force` option will shoutdown all non-migratable VMs ?
> 
> * If so, we could add a tooltip to the force checkbox or a message, to inform the users that they can use force checkbox to shutdown the VMs If first try fails.
> * else we need a new checkbox, visible on errors, to shutdown the VMs.
> 
> What do you think ?
> 
> cc @a110605

I think we can always display force shutdown banner message in the modal rather than show it after getting the unhealtyVMs list.
It provides clear inform to users.

[...]

PR #5991, opened 2024-06-07
improved logic for identify VM's that are not migratable

ibrokethecloud:

**Problem:**

Improve UX during maintenance mode to easily identify VM's which cannot be migrated

**Solution:**

The PR introduces improvements to identify non-migeratable VM's when one of the following conditions is met:
* VM's with cdroms and container disks
* VM's where kubevirt has migrated instance non migeratable for any of the possible reasons as identified by virt-handler: https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-handler/vm.go#L1592
* VM is migeratable but no destination nodes exist which satisfy the VM node scheduling requirements.

**Related Issue:**
https://github.com/harvester/harvester/issues/4888
**Test plan:**

### DIFF
--- a/pkg/api/node/formatter.go
+++ b/pkg/api/node/formatter.go
@@ -268,41 +268,16 @@ func (h ActionHandler) retryMaintenanceModeUpdate(nodeName string, updateFunc ma
 
 func (h ActionHandler) listUnhealthyVM(rw http.ResponseWriter, node *corev1.Node) error {
 	ndc := nodedrain.ActionHelper(h.nodeCache, h.virtualMachineInstanceCache, h.longhornVolumeCache, h.longhornReplicaCache)
-
-	var respObj ListUnhealthyVM
-
-	nonMigrtableVMList, err := ndc.FindAndListNonMigratableVM(node)
+	nonMigrtableVMList, err := ndc.FindNonMigratableVMS(node)
 	if err != nil {
 		return err
 	}
-
-	if len(nonMigrtableVMList) > 0 {
-		respObj.Message = "Please detach CDROM or ContainerDisk from following VMs:"
-		respObj.VMs = nonMigrtableVMList
-		rw.WriteHeader(http.StatusOK)
-		return json.NewEncoder(rw).Encode(&respObj)
-	}
-
-	vmWithPCIDevicesList, err := ndc.FindAndListVMWithPCIDevices(node)
-	if err != nil {
-		return err
-	}
-
-	if len(vmWithPCIDevicesList) > 0 {
-		respObj.Message = "Following VMs have PCIDevices attached and are non-migratable. Please power these off:"
-		respObj.VMs = vmWithPCIDevicesList
-		rw.WriteHeader(http.StatusOK)
-		return json.NewEncoder(rw).Encode(&respObj)
-	}
-
-	vmList, err := ndc.FindAndListVM(node)
-	if err != nil {
-		return err
-	}
-
-	if len(vmList) > 0 {
-		respObj.Message = "Following unhealthy VMs will be impacted by the node drain:"
-		respObj.VMs = vmList
+	respObj := make([]ListUnhealthyVM, 0, len(nonMigrtableVMList))
+	for condition, vms := range nonMigrtableVMList {
+		respObj = append(respObj, ListUnhealthyVM{
+			Message: fmt.Sprintf("The following VMs cannot be migrated due to %s condition", condition),
+			VMs:     vms,
+		})
 	}
 
 	rw.WriteHeader(http.StatusOK)
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -3,6 +3,8 @@ package nodedrain
 import (
 	"context"
 	"fmt"
+	goslices "slices"
+	"strings"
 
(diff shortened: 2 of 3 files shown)


ibrokethecloud:
a UI change is needed since the response json has been changed, and UI will need to handle this.

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -259,10 +262,34 @@ func (ndc *ControllerHandler) FindAndListNonMigratableVM(node *corev1.Node) ([]s
...
+	}
+
+	if len(unschedulableVMs) > 0 {
+		result[NodeSchedulingRequirementsNotMetKey] = unschedulableVMs
+	}
+
+	return result, nil
+}
+
+// FindAndListNonMigratableVM is called by action handler to leverage caches to find VM's which may have a cdrom or container disk
```
tserong:
```suggestion
// findVMSwithCDROMOrContainerDisk is called by action handler to leverage caches to find VMs which may have a cdrom or container disk
```

pkg/api/node/formatter.go
```diff
@@ -219,41 +219,16 @@ func (h ActionHandler) retryMaintenanceModeUpdate(nodeName string, updateFunc ma
...
-		return err
-	}
-
-	if len(vmList) > 0 {
-		respObj.Message = "Following unhealthy VMs will be impacted by the node drain:"
-		respObj.VMs = vmList
+	respObj := make([]ListUnhealthyVM, 0)
+	for condition, vms := range nonMigrtableVMList {
+		respObj = append(respObj, ListUnhealthyVM{
+			Message: fmt.Sprintf("following VM's cannot be migrated due to %s condition", condition),
```
tserong:
```suggestion
			Message: fmt.Sprintf("The following VMs cannot be migrated due to %s condition", condition),
```

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -293,22 +320,80 @@ func vmContainsCDRomOrContainerDisk(vmi *kubevirtv1.VirtualMachineInstance) bool
...
+
+	for _, v := range node.Status.Conditions {
+		if v.Type == corev1.NodeReady && v.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+func findUniqueVMS(vms map[string][]string) []string {
```
tserong: When I first saw the name of this function, I thought "wait, how are the VMs not already unique?" then I realised we're being passed a map of conditions to VM lists, because of course it's possible for there to be several conditions which would prevent a given VM from being migratable. Anyway, the point is, the name of the function kinda threw me :-) so I'd suggest maybe adding a comment about what's happening here, or, change the name (uhmmm... `getVMSfromConditionMap`?)

tserong (approved):
I have a few small suggestions below, and one item for discussion.

If I'm reading everything correctly, the previous implementation was problematic in that it stopped once it found _one_ reason for VMs being non-migratable, which means the user could conceivably go through several iterations ("oh, I need to detach some CD-ROMs", "ok, did that", "oh, now I need to detach some PCI devices", etc.). The implementation in this PR fixes that, by listing all the non-migratable VMs by condition in one hit.

The thing I'm wondering about is that the previous implementation provided guidance on what to do ("Please detach CDROM or ContainerDisk from following VMs") whereas this one will tell you about the condition only ("following VM's cannot be migrated due to CDRomOrContainerDiskPresent condition"), so there's an extra mental step required for the user ("how do I get rid of that condition?")

Do we need to add those conditions to the docs, with suggestions of how to resolve them? Or, should the dashboard translate "CDRomOrContainerDiskPresent" into an action for the user ("Please detach the CDROM or ContainerDisk...")? Or something else?

ibrokethecloud:
@tserong since i am hoping to leverage the vmi conditions as well, it may be best to update our docs to include the additional context

pkg/api/node/formatter.go
```diff
@@ -219,41 +219,16 @@ func (h ActionHandler) retryMaintenanceModeUpdate(nodeName string, updateFunc ma
...
-
-	vmList, err := ndc.FindAndListVM(node)
-	if err != nil {
-		return err
-	}
-
-	if len(vmList) > 0 {
-		respObj.Message = "Following unhealthy VMs will be impacted by the node drain:"
-		respObj.VMs = vmList
+	respObj := make([]ListUnhealthyVM, 0)
```
Yu-Jack:
```suggestion
	respObj := make([]ListUnhealthyVM, 0, len(nonMigrtableVMList))
```

Yu-Jack (commented):
LGTM, just a one place.

tserong (approved):

Yu-Jack (approved):
LGTM, thanks!

ibrokethecloud:
@torchiaf we need a UI change to go with this PR.

Yu-Jack (approved):
LGTM, but just found some redundant codes?

```go
// In pkg/controller/master/nodedrain/nodedrain_controller.go
		return node, nil
	}
	_, ok := node.Annotations[drainhelper.DrainAnnotation]
	if !ok {
		return node, nil
	}
	_, forced := node.Annotations[drainhelper.ForcedDrain]
	if ok {
```

This `if ok {` seems redundant, but we don't need to change it in this PR, just noticed this one.

tserong (approved):

bk201:
@mergifyio backport v1.4 v1.3

Issue #2756, opened 2022-09-09
[FEATURE] Scheduling of VM Backups


Martin-Weiss:
**Is your feature request related to a problem? Please describe.**
It seems we can only create "manual" backups of VMs - but we need a way to schedule backups and retire / cleanup old backups.

Similar to longhorn where we can create backups per volume on a scheduled basis with the configuration option to keep them for a specific time or only keep a specific number of backups - we should have that on a Harvester / Namespace / VM level, too.

**Describe the solution you'd like**
Configuration option for global backups, namespaced backups and VM based backups to schedule.
This also should include alerting and monitoring for successful and failed backups.

There should also be an option to backup while the VM is running and include memory / snapshot / pause the VM or shutdown (gracefully), backup and then startup again..

**Describe alternatives you've considered**
Longhorn based scheduled backup - but that is per volume and not per VM so restore is a huge challenge.

P.S. the same should be there for VM snapshots once we support them.

Martin-Weiss:
While we are waiting on v1.2.0 is there a manual workaround we could use? I.e. create a kubernetes batch job?

bk201:
Some nice comments from @weihanglo from https://github.com/harvester/harvester/issues/572

A possible user experience may be:

- When a user want to create a recurring backup
    1. A user goes to "Cron Jobs" page on the web UI.
    2. Then the user selects "create backup" job type and a dialog shows up.
    3. The user inputs [cron format] and select a VM to backup.
    4. The backup will start in schedule.
- A user also can check job status from the "Cron Jobs" page
   - A user can read their statuses directly.
   - A user can also read information of past jobs
 
Of course, we can  leverage `Job` and `CronJob`, which are K8s resource, so we just need to add cronjob actions API. No additional CR needed. We can tweak those statuses in a user-friendly to hide some implementation details as well.

Since [CronJobs] GA in K8s 1.21, the problem is, how much flexibility do we provide to users?

[CronJobs]: https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/
[cron format]: https://en.wikipedia.org/wiki/Cron

---

That would be great if we can reuse Longhorn's [recurring job] feature by specifying longhorn volume `volume.spec.recurringJobs`. And because Longhorn already [validates recurring jobs input], we might want to patch the volume directly from Harvester web UI via RESTful API.

The questions are:

[...]

PR #6008, opened 2024-06-11
HEP for VM schedule backup/snapshot

WebberHuang1118:

**Problem:**
Add HEP for VM schedule backup

**Solution:**
Add HEP for VM schedule backup

**Related Issue:**
https://github.com/harvester/harvester/issues/2756


enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Controller Flow Chart
+
+![flow chart](./20240611-vm-schedule-backup/control-flow.png)
+
+---
+
+### Test plan
+
+1. Schdule VM backup for a running VM
```
votdev:
```suggestion
1. Schedule VM backup for a running VM
```

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,143 @@
...
+   - Create a `ScheduleVMBackup` resource
+     - With `spec.vmbackup` points to the running VM
+     - With `spec.retain = 2`
+     - With `spec.cron = * * * * ?`
+   - New VM backup keeps created in every minute
+     - The Harvester always maintains at least two VM backups
+     - The oldest VM backup will be replaced with the latest one
+     - The oldest Longhorn snapshot also be purged under the hood
+
+2. Schdule VM backup for a stopped VM
```
votdev:
```suggestion
2. Schedule VM backup for a stopped VM
```

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,143 @@
...
+     - With `spec.maxflunk = 2`
+   - Make the backup target unreachable
+   - After about three minutes
+     - `status.suspend` is set to true
+     - `status.conditions` has the message `Reach Max Flunk`
+     - no more new VM backups are created
+
+4. Resume suspended VM schedule backup
+   - Follow the above test plan
+   - Make the backup target health
```
votdev:
```suggestion
   - Make the backup target healthy
```

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,143 @@
...
+```
+type ScheduleVMBackupSpec struct {
+	// +kubebuilder:validation:Required
+	Cron string `json:"cron"`
+
+	// +kubebuilder:validation:Required
+	Retain int `json:"retain"`
+
+	// +kubebuilder:validation:Required
+	MaxFlunk int `json:"maxflunk"`
```
votdev: Maybe we should use a more common word here, e.g. `MaxFailure` or `FailureThreshold`.

votdev (changes requested):

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: It was updated per suggestion, thanks.

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: It was updated per suggestion, thanks.

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: It was updated per suggestion, thanks.

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: It was updated per suggestion, thanks.

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### CRD
+
+- Add new resource `ScheduleVMBackup` to control VM schedule backup jobs
+```
+type ScheduleVMBackupSpec struct {
+	// +kubebuilder:validation:Required
+	Cron string `json:"cron"`
+
+	// +kubebuilder:validation:Required
```
ibrokethecloud, line 44: may be we should add some additional kubebuilder annotations to offload validation of spec to apiserver like max count etc

ibrokethecloud:
we should document how the failed backups are counted for evaluating suspension of a scheduled backup. 

say a user has a backup retention count of 10, but max failure count of 5. In a scenario where scheduled backup status is of the order: fail, fail, fail, fail, pass, fail, then the max failure count will be measured as 5, and backup will be suspended.

I am of the opinion that this should be measured as 1, and not 5. we should may be only count number of failures from the last successful backup.

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,143 @@
+# Support VM Schdule Backup
```
albinsun: Typo, Schdule -> Schedule.

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,143 @@
...
+- `status.suspend` presents if the backup schedule is suspended
+
+#### ScheduleVMBackup Controller
+
+- ScheduleVMBackup reconcile:
+  - Create a cronjob with the name `svmbackup-<ScheduleVMBackup UID>` if it's not presented
+  - If `.spec.resumeRequest` is set, check the backup target's health (by webhook), remove failed backups, and resume the schedule
+  - Update the current backup list in `status.vmbackups` at the end of each reconcile iteration
+- On cronjob is scheduled according to the cron expression
+  - If the number of failed backups reaches `spec.maxflunk`, suspend the schedule, update the related status, and exit
```
albinsun: `spec.maxFailure`?

albinsun, line 85: spec.maxFailure?

albinsun (commented):
Thanks for the documentation, it's quite useful.
I have 2 questions:
1. Will backup check the remaining space on backup-target or we there is existing protection?
2. I'm not sure my understanding is correct or not, but when `spec.resumeRequest` set to `True`, does it possible to have some case fall in looping?
    For example one of VM volumes is corruptted always fail to backup. And everytime reaching `spec.maxFailure` we will clear those failed records and resume schedule backup cuz `spec.resumeRequest`  is `True`.
    Will this looping forever until user set `spec.resumeRequest` to `False`?

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: It was updated per suggestion, thanks.

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: It was updated per suggestion, thanks.

WebberHuang1118:
> Thanks for the documentation, it's quite useful. I have 2 questions:
> 
> 1. Will backup check the remaining space on backup-target or we there is existing protection?

No, as I know Longhorn [backupstore](https://github.com/longhorn/backupstore) doesn't support it.


> 2. I'm not sure my understanding is correct or not, but when `spec.resumeRequest` set to `True`, does it possible to have some case fall in looping?
>    For example one of VM volumes is corruptted always fail to backup. And everytime reaching `spec.maxFailure` we will clear those failed records and resume schedule backup cuz `spec.resumeRequest`  is `True`.
>    Will this looping forever until user set `spec.resumeRequest` to `False`?

As the Harvester controller finds that `spec.resumeRequest` is true, the controller will clear it, remove failed backups, and resume the schedule. So the controller will not loop until certain operation is applied.

Thanks.

WebberHuang1118:
> we should document how the failed backups are counted for evaluating suspension of a scheduled backup.
> 
> say a user has a backup retention count of 10, but max failure count of 5. In a scenario where scheduled backup status is of the order: fail, fail, fail, fail, pass, fail, then the max failure count will be measured as 5, and backup will be suspended.
> 
> I am of the opinion that this should be measured as 1, and not 5. we should may be only count number of failures from the last successful backup.

Thanks @ibrokethecloud, I think your idea is comprehensive.

I will add `status.failure` to record the #failure backups in the list, and `status.consecutiveFaiure` to record the #consecutive failure backups (from the last successful backup).

The schedule will be suspended if `status.consecutiveFaiure` equals or more than `spec.maxFailure`

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: It was updated per suggestion, and I also assigned default values for `spec.retain` and `.spec.maxFailure`, thanks.

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,173 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=8
+	// +kubebuilder:validation:Maximum=254
+	// +kubebuilder:validation:Minimum=2
+	Retain int `json:"retain"`
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=4
+	// +kubebuilder:validation:Minimum=2
+	MaxFailure int `json:"maxFailure"`
```
FrankYang0529, line 53: Since we use `ConsecutiveFailure` to determine whether to suspend the cronjob, do we want to change `MaxFailure` to `MaxConsecutiveFailure`? I think it's more clear for users to know when the cronjob will be suspended.

FrankYang0529 (commented):
Overall LGTM. Thanks for the HEP.

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: good idea, thanks @FrankYang0529

WebberHuang1118: Hi @FrankYang0529 

After discussing with @ibrokethecloud, I changed `status.failure` (remove `status.consecutiveFailure`) to count the consecutive failure backups since the last successful one. Because the #total failure backups could be obtained from `status.vmbackupInfo`, and it will not affect the controller's behavior, thanks.

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,171 @@
...
+- `status.conditions` presents the schedule's status, so far, it's only leveraged to describe the reason for the schedule is suspended
+
+#### ScheduleVMBackup Controller
+
+- ScheduleVMBackup reconcile:
+  - If `.spec.resumeRequest` not set
+    - Create a cronjob with the name `svmbackup-<ScheduleVMBackup UID>` if it's not presented
+    - If the number of the current VM backups goes beyond `spec.retain`
+      - Clean up the oldest VM backup(s)
+      - Delete the outdated Longhorn snapshots to trigger the snapshot purge
```
FrankYang0529: For this one, probably, we can leverage `auto-cleanup-when-delete-backup` setting from https://github.com/longhorn/longhorn/issues/8365.

FrankYang0529 (approved):
Overall LGTM. Thanks for the HEP.

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: Thanks, @FrankYang0529, I will check it with LH v1.7.0

FrankYang0529: I will also do this in snapshot space management. We can create a PR with LH v1.7.0-rc version and include the setting, so we don't do it twice.

abonillabeeche:
Can this also include snapshots, and not only Backups?

enhancements/20240611-vm-schedule-backup.md
FrankYang0529: https://github.com/harvester/harvester/pull/6236

a110605:
Do we consider let user manually stop and resume the schedule job ? 

Currenty design make the schedule job continually create backup until reach maxFailure threshold and become suspended.

But user is unable to stop a running schedule job unless delete the schedule job.

WebberHuang1118:
> Can this also include snapshots, and not only Backups?

Hi @abonillabeeche,
We will try to support Schedule VM Snapshot, it's under processing and testing.

WebberHuang1118:
> Do we consider let user manually stop and resume the schedule job ?
> 
> Currenty design make the schedule job continually create backup until reach maxFailure threshold and become suspended.
> 
> But user is unable to stop a running schedule job unless delete the schedule job.

Thanks @a110605,
It's a good idea, I will check and discuss with @bk201

bk201:
@WebberHuang1118 sounds like a very basic feature, please help implement it.

a110605:
For a created schedule job, should we allow user to modify below parameters ?
- cron
- retain 
- maxFailure 
- auto resume

bk201 (approved):
LGTM, thanks!

ibrokethecloud (approved):
lgtm. thanks.

enhancements/20240611-vm-schedule-backup.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+- `spec.cron` specifies the cron expression for the VM backup schedule
+- `spec.retain` specifies the number of up-to-date VM backups to retain
+- `spec.maxFailure` specifies the max number of consecutive failed backups that could be tolerated, if the number of consecutive failed backups reaches this setting, the Harvester controller will suspend the schedule
+- `spec.suspend` is used to suspend/resume the schedule proactively
+- `spec.vmbackup` specifies the source VM to backup/snapshot
+- `status.vmbackupInfo` records the current backup/snapshot list, including the status of all the volume backups/snapshots for the VM
+- `status.failure` number of consecutive failure backups after the last successful backups
+- `status.suspend` presents if the schedule is suspended
```
votdev: Shouldn't this named `suspended`? The `spec.suspend` describes an action (so present tense is correct), but `status.suspend` describes a status, so it needs to be named `status.suspended` (past tense).

votdev (changes requested):

enhancements/20240611-vm-schedule-backup.md
WebberHuang1118: Updated per suggestion, will update the related implementation in the controller, too. Thanks.

votdev (approved):

albinsun (approved):
LGTM

Issue #6014, opened 2024-06-12
[ENHANCEMENT] Node Disk Manager refactor


Vicente-Cheng:
**Is your enhancement related to a problem? Please describe.**
Node Disk Manager is used to manage the physical devices on the node. The current implementation is too tight to support other storage backends. We want to decompose some components (mainly the provision stage). It could be more flexible and easy to maintain and extend.

**Describe the solution you'd like**
Mainly decompose the provision stage with a specific provisioner

**Describe alternatives you've considered**

**Additional context**
We want to support at least three storage backends as below:
- [x] LonghornV1: Default
- [x] LonghornV2: https://github.com/harvester/node-disk-manager/pull/129
- [x] LVM: https://github.com/harvester/node-disk-manager/pull/128

NDM should have the corresponding provisioner.

Vicente-Cheng:
It's a code-level refactor that includes the various provisioners.
So, I closed it as the refactor had been down.

Issue #5059, opened 2024-01-30
[ENHANCEMENT] Support filter list on NDM


Vicente-Cheng:
**Is your enhancement related to a problem? Please describe.**
Now, NDM can filter out like `vendor`, `label`, and `path` by ENV or parameter
We could have the filter list to include the above. That would be more convenient.

**Describe the solution you'd like**
Add a filter list, and NDM can filter with this.

**Describe alternatives you've considered**
None

**Additional context**
None

innobead:
Discussed with @Vicente-Cheng . He will share the detailed idea and proposal for this improvement. I would expect @tserong can drive the implementation.

tserong:
I discussed with @Vicente-Cheng a bit last week - he suggested maybe using a ConfigMap for the filters. I've done some experimenting, and we could do something like this pretty easily:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: harvester-node-disk-manager
  namespace: harvester-system
data:
  label_filter: "COS_*,HARV_*"
  path_filter: "/mount/path/you/want/to/exclude"
  # also vendor_filter, auto_provision_filter
```

Then in the NDM daemonset:

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: harvester-node-disk-manager
  namespace: harvester-system
  # [...]
spec:
  # [...]
  template:
    spec:
      containers:
      - command:
        - node-disk-manager
        env:
        - name: NDM_LABEL_FILTER
          valueFrom:
            configMapKeyRef:
              key: label_filter
              name: harvester-node-disk-manager
        - name: NDM_PATH_FILTER
          valueFrom:
            configMapKeyRef:
              key: path_filter
              name: harvester-node-disk-manager
        # ...and so on for other filters...
```

The startup logic with environment variables could remain the same, but we'd of course need to add something to monitor for changes to the ConfigMap at runtime. Or, we get rid of the environment variables and just have NDM check the ConfigMap at startup (I'm not sure yet which would make the most sense).

Some further thoughts:

- The current implementation means the same set of filters are applied to all nodes. Are we likely to need to have filters that only apply to specific nodes? Or that are somehow different on a per-node basis?
- I can imagine a couple of ways in which the existing filter sets could be enhanced (e.g. we don't seem to currently have exclude by device path, but I can see that being potentially useful).
- It would be helpful to make the debug logging a bit more verbose, to say which filter was matched when something was included/excluded.

[...]

PR #6015, opened 2024-06-12
HEP: add node-disk-manager refactor

Vicente-Cheng:

**Problem:**
 The proposal for Node Disk Manager refactor

**Solution:**
Give the HEP for the whole story

**Related Issue:**
https://github.com/harvester/harvester/issues/6014

**Test plan:**
None


enhancements/20240612-node-disk-manager-refactor.md
```diff
@@ -0,0 +1,67 @@
...
+
+https://github.com/harvester/harvester/issues/6014
+
+## Motivation
+
+We need to refactor the current implementation to make the Node Disk Manager support various storage backends. This could help define the storage provisioner easily.
+
+### Goals
+
+Refactor the Node Disk Manager to make it more flexible and extensible to support various storage backends. The refactor will decompose the individual storage provisioner.
```
tserong: I would suggest removing the above sentence as it's pretty similar to the motivation section, and instead explicitly state the additional storage backends we wish to end up with, as goals, similar to what's in the related issue https://github.com/harvester/harvester/issues/6014. Maybe something like this?

- Add longhornv2 provisioner
- Add LVM provisioner
- Continue to support existing longhornv1 provisioning

enhancements/20240612-node-disk-manager-refactor.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Non-goals [optional]
+
+None
+
```
tserong, line 21: If there's no non-goals I think this section can be simply deleted.

enhancements/20240612-node-disk-manager-refactor.md
```diff
@@ -0,0 +1,67 @@
...
+#### The user wants to use the longhornV2 engine
+The user wants to try the LonghornV2 engine on the Harvester. The user can use the longhornv2 provisioner to provision the raw device to Longhorn for the v2 engine.
+
+#### The user wants to use LVM as the storage backend
+The user wants to use LVM as the storage backend for better performance. The node-disk-manager would help create the volume group, and the LVM CSI driver could depend on it to create the logical volume.
+
+### User Experience In Detail
+
+The user could use a specific provisioner like `longhornv1`, `longhornv2`, or `LVM` to provision the raw device to the corresponding storage backend. After that, we could easily create PVC from the specific storage backend.
+
```
tserong, line 44: Do we need to say anything here about specific steps in the GUI? e.g. when adding a disk to a node?

Also, we should probably mention the auto_provision_filter, to which I think we'll need to add a provisioner type (see my comments at https://github.com/harvester/harvester/issues/5059#issuecomment-2182056033 for more notes on this)

tserong (commented):
A few comments inline

enhancements/20240612-node-disk-manager-refactor.md
```diff
@@ -0,0 +1,67 @@
...
+The Node Disk Manager is responsible for managing the raw device on the node. The current implementation is not flexible enough to support various storage backends. This enhancement is to refactor the Node Disk Manager to make it more flexible and extensible.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/6014
+
+## Motivation
+
+We need to refactor the current implementation to make the Node Disk Manager support various storage backends. This could help define the storage provisioner easily.
```
Yu-Jack: I'd suggest that you could mention current limitation about how difficult if we'd like to add LVM and v2 provisioner, then it becomes our motivation to provide this proposal.

For example: 

If we don't use abstract coupling to decouple current implementation, it'd be hard to add new provisioner. 

Then this part can be different with Goals section. This one is similar with Tim, just a advice.

enhancements/20240612-node-disk-manager-refactor.md
```diff
@@ -0,0 +1,67 @@
...
+### API changes
+
+TBD
+
+## Design
+
+### Implementation Overview
+
+A overview of the design could be checked below:
+![ndm-refactor](20240612-node-disk-manager-refactor/ndm-refactor.png)
```
Yu-Jack, line 76: image is broken in https://github.com/harvester/harvester/blob/4a83da526b969b4e97ee033b78bdc7c78d105e9e/enhancements/20240612-node-disk-manager-refactor.md

enhancements/20240612-node-disk-manager-refactor.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Provisioner
+
+The refactor will try to decompose the provisioner from the controller. Each provisioner will be responsible for the specific storage backend. For example, the longhornv1 provisioner will handle the filesystem formatting on the Format stage and provision this raw device to `nodes.longhorn.io`. That longhorn could create volume on the provisioned raw device. The LVM provisioner only handles the provision stage to create the volume group.
+
+### Test plan
+
+1. Create Harvester cluster
+2. Create supported storage provisioner (e.g. longhornv1, longhornv2, and LVM)
```
Yu-Jack, line 87: I'd suggest make a fake provisioner to do unit test. It shows that you create a well interface that can be replaced by any new provisioners.

enhancements/20240612-node-disk-manager-refactor.md
Vicente-Cheng: Thanks @tserong,
I rewrote this part to specify the Goals of this refactor.
Please help to recheck!

enhancements/20240612-node-disk-manager-refactor.md
Vicente-Cheng: done!

enhancements/20240612-node-disk-manager-refactor.md
Vicente-Cheng: Sure, I add more detailed steps from the UI side here.
Please help to recheck, thanks!

enhancements/20240612-node-disk-manager-refactor.md
Vicente-Cheng: Thanks @Yu-Jack,
I also rewrote this part.
Hope that could give more clarity to the motivation!
Please help to recheck, thanks!

enhancements/20240612-node-disk-manager-refactor.md
Vicente-Cheng: fixed, nice catch!

enhancements/20240612-node-disk-manager-refactor.md
Vicente-Cheng: Thanks for the suggestion.
I added your suggestion on the unit test part of the design.

I thought the `Test plan` was used for integration tests or QA validation, so I prefer to keep that.
WDYT?

Yu-Jack: Oh, you're right. I misunderstood it. Sorry for that. I don't have problems now. Thanks!

Vicente-Cheng:
Hi folks, please review this HEP again.
I did some updates to align with the implementation.

Thanks!

bk201 (approved):

enhancements/20240612-node-disk-manager-refactor.md
```diff
@@ -0,0 +1,95 @@
...
+https://github.com/harvester/harvester/issues/6014
+
+## Motivation
+
+Currently, the Node Disk Manager can only provide the Longhorn v1 engine. Most codes of NDM are tight coupling with the Longhorn v1 engine and are hard to do to any extent. If we want to create various provisioners, we need to refactor the Node Disk Manager to make it more flexible and extensible. After that, we could easily create the provisioner for the Longhorn v2 engine or LVM.
+
+### Goals
+
+This refactor will introduce the provisioner concept to define various storage provisioners. We plan to create the following provisioner:
+- LonghornV1: The `LonghornV1` provisioner will work like the previous version to provision the longhorn v1 engine backend.
```
votdev, line 18:
```suggestion
- LonghornV1: The `LonghornV1` provisioner will work like the previous version to provision the Longhorn v1 engine backend.
```

enhancements/20240612-node-disk-manager-refactor.md
```diff
@@ -0,0 +1,95 @@
...
+- LonghornV2: The `LonghornV2` provisioner will provision the raw device to the longhorn v2 engine.
```
votdev, line 19:
```suggestion
- LonghornV2: The `LonghornV2` provisioner will provision the raw device to the Longhorn v2 engine.
```

PR #6018, opened 2024-06-13
Release v1.3.1

bk201:


Vicente-Cheng (approved):
LGTM, tested the upgrade path.

Also, checked the basic operations. Looks well

bk201:
smoke test result: [test-results (4).zip](https://github.com/user-attachments/files/15818704/test-results.4.zip)

re-test `test_create_vm_no_available_resources[disk]` and it passes, should be a false alarm.

PR #6075, opened 2024-06-26
test: check apparmor

Vicente-Cheng:
For Checking GHA apparmor


Vicente-Cheng:
close it because we already checked the GHA apparmor

Issue #6070, opened 2024-06-26
[BUG] harvester `make ci` failed when verify vm api in local


brandboat:
**Describe the bug**
execute `make ci` in local laptop under harvester master branch, got error as follows.
The verify vm api always failed, and in the kind I found both harvester deployment and harvester node manager not running.
![image](https://github.com/harvester/harvester/assets/4344302/8831d386-43ab-4c7c-a1e1-8e54ede7842b)
![image](https://github.com/harvester/harvester/assets/4344302/1902b7a2-c7b9-4af6-b1d6-463a51d1abcc)
![image](https://github.com/harvester/harvester/assets/4344302/859e01ee-5475-45cd-9cdc-abc828b04aa5)
![image](https://github.com/harvester/harvester/assets/4344302/dbf6fb19-21ea-4e0a-a5d0-89cd0c1d5060)

**To Reproduce**
1. check out harvester master branch
2. execute `make ci`

**Expected behavior**
no error

**Support bundle**
N/A

**Environment**
- harvester master branch
- local laptop

**Additional context**
`make ci` passed in github actions.

bk201:
The error is virt-launcher pod fails with:
```
compute {"component":"virt-launcher","kind":"","level":"error","msg":"Failed to sync vmi","name":"test-grsvz","namespace":"default","pos":"server.go:202","reason":"virError(Code=1, Domain=1
0, Message='internal error: Failed to start QEMU binary /usr/bin/qemu-system-x86_64 for probing: libvirt:  error : cannot execute binary /usr/bin/qemu-system-x86_64: Operation not permitted
\n')","timestamp":"2024-06-26T07:46:15.346012Z","uid":"e4ec1043-d9ce-4246-b672-f554e9a00169"}
```

Per @Vicente-Cheng's investigation, adding `Root` feature to kubevirt object is a workaround. We don't know why the CI pass in GHA runners yet.

Vicente-Cheng:
For now, we can disable the following apparmor for the local tests.

```
// on host side
# aa-disable /etc/apparmor.d/usr.sbin.libvirtd
# aa-disable /etc/apparmor.d/usr.sbin.virtqemud
```

But I am not sure if there is another way to give the correct permission, or if we need to disable that.

Vicente-Cheng:
I added the test to check our GHA configuration for AppArmor and looks like the AppArmor is disabled on that.
PR (only for checking): https://github.com/harvester/harvester/pull/6075 
[...]

PR #6079, opened 2024-06-27
README: add the appArmor related document [CI SKIP]

Vicente-Cheng:

**Problem:**
`make ci` failed in some local environment

**Solution:**
Give a clear hint for how to run the `make ci` locally

**Related Issue:**
https://github.com/harvester/harvester/issues/6070

**Test plan:**
No need

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -5,6 +5,23 @@ This path contains the basic testing framework and integration testing cases, th
 ### Integration test
 The integration test is performed on every pull request and is run by the Drone CI. The script is stored in the `scripts/test-integration` file.
 
+Before the integration test, you must check the AppArmor status to ensure the `usr.sbin.virtqemud` is not enabled.
+Or you might see the following error in virt-launcher pods when you test the VM API
+
+```
+compute {"component":"virt-launcher","kind":"","level":"error","msg":"Failed to sync vmi","name":"test-grsvz","namespace":"default","pos":"server.go:202","reason":"virError(Code=1, Domain=10, Message='internal error: Failed to start QEMU binary /usr/bin/qemu-system-x86_64 for probing: libvirt:  error : cannot execute binary /usr/bin/qemu-system-x86_64: Operation not permitted\n')","timestamp":"2024-06-26T07:46:15.346012Z","uid":"e4ec1043-d9ce-4246-b672-f554e9a00169"}
+```
+
+You can use the following command to check:
+```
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -5,6 +5,22 @@ This path contains the basic testing framework and integration testing cases, th
 ### Integration test
 The integration test is performed on every pull request and is run by the Drone CI. The script is stored in the `scripts/test-integration` file.
 
+Before the integration test, you need to check the apparmor status to ensure the `usr.sbin.virtqemud` is not enabled.
+Or you might see the following error when you test VM API
```
bk201:
```suggestion
Before the integration test, you must check the AppArmor status to ensure the `usr.sbin.virtqemud` is not enabled.
Or you might see the following error in virt-launcher pods when you test the VM API

```
A tiny suggestion to clarify where the log is.

bk201 (approved):

tests/README.md
```diff
@@ -5,6 +5,22 @@ This path contains the basic testing framework and integration testing cases, th
...
+```
+
+You can use the following command to check:
+```
+$ aa-status |grep virtqemud
+```
+
+And use the following command to disable:
+```
+$ aa-disable /etc/apparmor.d/usr.sbin.virtqemud
```
brandboat: So we don't need to disable libvirtd as well ?

BTW I need to use sudo to make `aa-disable /etc/apparmor.d/usr.sbin.virtqemud` work.

brandboat (approved):
LGTM, thank you 😃

tests/README.md
Vicente-Cheng: yeah, I tested, we only need to disable `virtqemud`.

I will add the sudo... thanks for the reminder!

brandboat: > yeah, I tested, we only need to disable virtqemud.

I see. Thanks for clarifying. 👍

Vicente-Cheng:
This is only a document change. Seems the [CI SKIP] is not working. I go ahead merge it because all reviewers approved.

brandboat:
> Seems the [CI SKIP] is not working

Yeah... it works only when adding `ci skip` in commit message
quoted from https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs
> Workflows that would otherwise be triggered using on: push or on: pull_request won't be triggered if you add any of the following strings to the commit message in a push, or the HEAD commit of a pull request:

Issue #4226, opened 2023-07-13
[BUG]Redundant NAD created during applying Storage-Network change


WebberHuang1118:
**Describe the bug**

Longhorn only requires one NAD for Storage Network. However, if applying Storage Network change, there are multiple identical NADs could be found on Harvester

**To Reproduce**
1. Enable Storage-Network and apply configuration ![storage-network](https://github.com/harvester/harvester/assets/19387129/47104b5a-11f6-42d2-90af-4745283eb49f)
2. After applying Storage-Network, there are multiple NADs could be found on Harvester, for example:
```
   kubectl -n harvester-system get network-attachment-definitions.k8s.cni.cncf.io -o wide
   NAME                   AGE
   storagenetwork-25c9h   10m
   storagenetwork-2wkbw   10m
   storagenetwork-9snqr   10m

```
3. Longhorn only leverages one of the NADs ![longhorn-storage-network](https://github.com/harvester/harvester/assets/19387129/0b18c243-8cee-4427-97f1-a77674012272)

**Expected behavior**

There should be only one NAD existing for Storage-Network

guangbochen:
can't reproducible this issue in my `v1.2.0-rc4` 3 nodes cluster with the above steps, wondering if any step is missing?
```
ks get network-attachment-definitions.k8s.cni.cncf.io -n harvester-system
NAME                   AGE
storagenetwork-jl9qj   31m
```

WebberHuang1118:
From my experience,
It does not always happen. I think it is related to the timing affecting on https://github.com/harvester/harvester/blob/2e0a773e2d21880442d605cce412dd6f5978a0e2/pkg/controller/master/storagenetwork/storage_network.go#L317-L326

lanfon72:
Verified this bug has been fixed.


Test Information
----
* Environment: **qemu/KVM 3 nodes**
* Harvester Version: **v1.4.0-rc3**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Create cluster network and configuration
3. enable storage network
4. Execute following script to test `storage-network` switching (update required info which quote with `<>`)
```python
from harvester_api import API  # install from https://github.com/harvester/tests
from time import sleep
from datetime import datetime, timedelta

import urllib3
urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
api = API('https://<cluster_VIP>')
api.session.verify = False
api.authenticate('admin', '<password>')

[...]

PR #6086, opened 2024-06-27
[Storage Network] Checking existing NAD if CR updating conflict occurs

WebberHuang1118:
**Problem:**
Redundant NAD created during applying Storage-Network

**Solution:**
Checking and reusing existing NAD if a conflict occurs during update setting the annotation

**Related Issue:**
#4226 

**Test plan:**

Applying storage network setting
1. Update Harvester deployment with this PR
2. Enable storage network
3. Check the number of NAD with the prefix `storagenetwork`
4. There should be only one NAD with this prefix
5. VM can start normally

Switching storage network setting
1. Continue with the above steps
2. Switch to another setting (e.q. `172.16.0.0/24` -> `192.16.0.0/24`)
3. Check the number of NAD with the prefix `storagenetwork`
4. There should be only one NAD with this prefix
5. VM can start normally

Controller upgrade path
1. Enable storage network in master-head branch
2. Update Harvester deployment with this PR
3. There should be only one NAD with this prefix
4. VM can start normally

### DIFF
--- a/pkg/controller/master/storagenetwork/storage_network.go
+++ b/pkg/controller/master/storagenetwork/storage_network.go
@@ -37,6 +37,8 @@ const (
 	NadStorageNetworkAnnotation     = StorageNetworkAnnotation + "/net-attach-def"
 	OldNadStorageNetworkAnnotation  = StorageNetworkAnnotation + "/old-net-attach-def"
 
+	HashStorageNetworkLabel = HashStorageNetworkAnnotation
+
 	StorageNetworkNetAttachDefPrefix    = "storagenetwork-"
 	StorageNetworkNetAttachDefNamespace = "harvester-system"
 
@@ -264,12 +266,12 @@ func (h *Handler) setNadAnnotations(setting *harvesterv1.Setting, newNad string)
 	return setting
 }
 
-func (h *Handler) createNad(setting *harvesterv1.Setting) (string, error) {
+func (h *Handler) createNad(setting *harvesterv1.Setting) (*nadv1.NetworkAttachmentDefinition, error) {
 	var config Config
 	bridgeConfig := NewBridgeConfig()
 
 	if err := json.Unmarshal([]byte(setting.Value), &config); err != nil {
-		return "", fmt.Errorf("parsing value error %v", err)
+		return nil, fmt.Errorf("parsing value error %v", err)
 	}
 
 	bridgeConfig.Bridge = config.ClusterNetwork + BridgeSuffix
@@ -286,7 +288,7 @@ func (h *Handler) createNad(setting *harvesterv1.Setting) (string, error) {
 
 	nadConfig, err := json.Marshal(bridgeConfig)
 	if err != nil {
-		return "", fmt.Errorf("output json error %v", err)
+		return nil, fmt.Errorf("output json error %v", err)
 	}
 
 	nad := nadv1.NetworkAttachmentDefinition{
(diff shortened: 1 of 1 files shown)


pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -303,35 +303,44 @@ func (h *Handler) createNad(setting *harvesterv1.Setting) (string, error) {
...
+	var nadNamespace, nadName string
+	nad := ""
 
 	if h.checkIsSameHashValue(setting) {
 		return setting, nil
 	}
 
 	if setting.Value != "" {
-		nadName, err = h.createNad(setting)
+		nadNamespace, nadName, err = h.createNad(setting)
```
bk201: Can we check the existence of the NAD and skip creating a new one, instead of deleting the NAD?

WebberHuang1118: It will lose track because the controller may fail to record the new NAD in the storage network setting annotation. 

IMHO, the safest way is to delete the NAD which can't be correctly recorded. Or do you have a better suggestion? @bk201 

Thanks.

starbops: Maybe we don't have to track the NAD for the storage-network setting because there will be only one NAD for the storage network. Is it possible to add a label to denote such a created NAD is for the storage network? With that, it should be possible to list NADs with a label filter. Just my two cents :P

WebberHuang1118: Good idea, I add the NAD's label with the hash from the setting value, and check if the NAD exists with the required hash in each reconcile, thanks.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -298,34 +300,55 @@ func (h *Handler) createNad(setting *harvesterv1.Setting) (string, error) {
...
+
+	return nadResult, nil
+}
+
+func (h *Handler) findOrCreateNad(setting *harvesterv1.Setting) (*nadv1.NetworkAttachmentDefinition, error) {
+	sets := labels.Set{
+		HashStorageNetworkLabel: h.sha1(setting.Value),
+	}
+
+	nads, err := h.networkAttachmentDefinitionsCache.List(StorageNetworkNetAttachDefNamespace, sets.AsSelector())
```
starbops: What if the cache is stale, and we still get no NAD in return? Maybe we can use the client to list. Though expensive, we won't do this often. It should be fine.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -298,34 +300,55 @@ func (h *Handler) createNad(setting *harvesterv1.Setting) (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(nads) > 0 {
```
starbops: Do we want to handle the case of more than one NAD? Or print a warning message because those excessive NADs should be safe enough to be purged manually.

starbops (commented):
We may need to consider the case when changing the storage network from A to B. The current version seems incapable of doing this.

WebberHuang1118:
> We may need to consider the case when changing the storage network from A to B. The current version seems incapable of doing this.

Do you mean to change the storage network between different settings? I've checked with it. 
Or are there other cases I didn't notice?

https://github.com/harvester/harvester/blob/2f6a78e963cadf669280003f0f181c1b989b0af1/pkg/controller/master/storagenetwork/storage_network.go#L318-L325

Harvester controller will check if matched NAD exists with the hash of the setting's value

starbops (approved):
This works pretty nice. LGTM, thank you!

pkg/controller/master/storagenetwork/storage_network.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/storagenetwork/storage_network.go
WebberHuang1118: Updated per suggestion, if there is more than one match NAD, the controller will leave a warning message and choose the first one. Thanks.

WebberHuang1118:
Hi @starbops,
I updated the PR per your suggestion and the test plan, please take a look, thanks.

starbops (approved):

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -298,34 +300,59 @@ func (h *Handler) createNad(setting *harvesterv1.Setting) (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(nads.Items) == 0 {
+		return h.createNad(setting)
+	}
+
+	if len(nads.Items) > 1 {
+		logrus.Warnf("storage network: found %d match nad, will use %s", len(nads.Items), nads.Items[0].Name)
```
bk201: nit. We are moving to [structure logging](https://github.com/harvester/harvester/blob/master/enhancements/20240327-structured-logging.md). Maybe we can consider adding the nad name as field here.

bk201 (approved):
LGTM, thanks!

pkg/controller/master/storagenetwork/storage_network.go
WebberHuang1118: Hi @bk201,
Updated per suggestion, please take a look, thanks.

bk201 (approved):

bk201:
@mergifyio backport v1.4 v1.3

Issue #4992, opened 2024-01-18
[ENHANCEMENT] setting.Default is not checked by webhook


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
Currently, we only validate setting.Value, but not setting.Default. In some cases, if setting.Value is empty, we use setting.Default instead. If it's broken, the system may get error.

**Describe the solution you'd like**
We should check setting.Default as well or deny update request for the field.

TachunLin:
Verified **fixed** on `master-8b80f82-head` (24/07/17). Close this issue

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ Check <strong>vm-force-reset-policy</strong> default setting $~~$</summary>

  1. Update the default value as a broken json string 
        <details><summary>$\color{green}{\textsf{PASS}}$: Can deny update request for the field  $~~$</summary>
                
        ```
        # Please edit the object below. Lines beginning with a '#' will be ignoredeny update request for the fieldd,
        # and an empty file will abort the edit. If an error occurs while saving this file will be
        # reopened with the relevant failures.
        #
        # settings.harvesterhci.io "vm-force-reset-policy" was not valid:
        # * default: unmarshal failed, error: invalid character '}' looking for beginning of value, value: {"enable":}
        #
        apiVersion: harvesterhci.io/v1beta1
        default: '{"enable":}'
        kind: Setting
        metadata:
          creationTimestamp: "2024-07-17T10:07:21Z"
          generation: 1
          name: vm-force-reset-policy
          resourceVersion: "6994"
          uid: fea02251-1b57-4ca5-a744-ddbbacfb70e7
        status: {}
        ```
      
      </details>

  2. Update the default value with non-boolean value for "enable" field 
        <details><summary>$\color{green}{\textsf{PASS}}$: Can deny update request for the field  $~~$</summary>
                
        ```
        # Please edit the object below. Lines beginning with a '#' will be ignored,
        # and an empty file will abort the edit. If an error occurs while saving this file will be
        # reopened with the relevant failures.
        #
        # settings.harvesterhci.io "vm-force-reset-policy" was not valid:
        # * default: unmarshal failed, error: json: cannot unmarshal number into Go struct field VMForceResetPolicy.enable of type bool, value: {"enable": 300,"period":300}
        #
        apiVersion: harvesterhci.io/v1beta1
        default: '{"enable": 300,"period":300}'
        kind: Setting
        metadata:
[...]

PR #6088, opened 2024-06-27
feat: check setting.Default

FrankYang0529:

**Problem:**
Currently, we only validate setting.Value, but not setting.Default. In some cases, if setting.Value is empty, we use setting.Default instead. If it's broken, the system may get error.

**Solution:**
We should check setting.Default as well.

**Related Issue:**
https://github.com/harvester/harvester/issues/4992

**Test plan:**

#### vm-force-reset-policy
1. Update the default value as a broken json string like '{"enable":}' and webhook should return error.
2. Update the default value with non-boolean value for "enable" field like '{"enable": 300,"period":300}' and webhook should return error.
3. Update the default value with negative value for "period" field like '{"enable": true,"period":-1}' and webhook 
should return error.
4. Update the default value with correct field value like '{"enable": false,"period":100}' and it should pass.

#### support-bundle-image
1. Update the default value as a broken json string like '{' and webhook should return error.
2. Update the default value with empty field value like '{"repository":"","tag":"","imagePullPolicy":""}' and webhook should return error.

#### support-bundle-timeout
1. Update the default value as a non-integer string like '1.1' and webhook should return error.
2. Update the default value as a negative integer string like '-1' and webhook should return error.

#### support-bundle-expiration
1. Update the default value as a non-integer string like '1.1' and webhook should return error.
2. Update the default value as a negative integer string like '-1' and webhook should return error.

#### support-bundle-node-collection-timeout
1. Update the default value as a non-integer string like '1.1' and webhook should return error.
2. Update the default value as a negative integer string like '-1' and webhook should return error.

#### overcommit-config
1. Update the default value as a broken json string like '{' and webhook should return error.
2. Update the default value with an integer less than 100 for "cpu" field like '{"cpu":99,"memory":150,"storage":200}' and webhook should return error.
3. Update the default value with an integer less than 100 for "memory" field like '{"cpu":1600,"memory":99,"storage":200}' and webhook should return error.
4. Update the default value with an integer less than 100 for "memory" field like '{"cpu":1600,"memory":150,"storage":99}' and webhook should return error.

#### vip-pools
1. Update the default value as a broken json string like '{' and webhook should return error.

#### ssl-certificates
1. Update the default value as a broken json string like '{' and webhook should return error.

#### ssl-parameters
1. Update the default value as a broken json string like '{' and webhook should return error.

#### containerd-registry
1. Update the default value as a broken json string like '{' and webhook should return error.

#### default-vm-termination-grace-period-seconds
1. Update the default value as a non-integer string like '1.1' and webhook should return error.
2. Update the default value as a negative integer string like '-1' and webhook should return error.

#### ntp-servers
1. Update the default value as a broken json string like '{' and webhook should return error.

#### auto-rotate-rke2-certs
1. Update the default value as a broken json string like '{' and webhook should return error.
2. Update the default value with non-boolean value for "enable" field like '{"enable":300,"expiringInHours":240}' and webhook should return error.
3. Update the default value with negative value for "expiringInHours" field like '{"enable":true,"period":-1}' and webhook 
should return error.
4. Update the default value with an integer value large than 8759 for "expiringInHours" field like '{"enable":true,"period":8760}' and webhook should return error.

#### kubeconfig-default-token-ttl-minutes
1. Update the default value as a non-integer string like '1.1' and webhook should return error.
2. Update the default value as a negative integer string like '-1' and webhook should return error.

#### backup-target
1. Update the default value as a broken json string like '{' and webhook should return error.

#### volume-snapshot-class
1. Update the default value with a non-existent volume snapshot class like 'longhorn-test' and webhook should return error.

#### storage-network
1. Update the default value as a broken json string like '{' and webhook should return error.

#### http-proxy
1. Update the default value as a broken json string like '{' and webhook should return error.

### DIFF
--- a/pkg/settings/settings.go
+++ b/pkg/settings/settings.go
@@ -267,6 +267,10 @@ func DecodeVMForceResetPolicy(value string) (*VMForceResetPolicy, error) {
 		return nil, fmt.Errorf("unmarshal failed, error: %w, value: %s", err, value)
 	}
 
+	if policy.Period <= 0 {
+		return nil, fmt.Errorf("period value should be greater than 0, value: %d", policy.Period)
+	}
+
 	return policy, nil
 }
 
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -24,6 +24,7 @@ import (
 	_ "github.com/longhorn/backupstore/nfs" //nolint
 	_ "github.com/longhorn/backupstore/s3"  //nolint
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	"github.com/rancher/lasso/pkg/log"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/pkg/slice"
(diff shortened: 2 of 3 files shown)


Yu-Jack (approved):
LGTM.

There is a suggestion could be HEP.

Since this file is too long and we mix all configuration here. Sometime it's hard to read. For example, because we need to setup multiple places to trigger this validation, so I forgot to add validation for Update once. I think we could split all setting into smaller files and define the interface.

Let's say we could refactor to this one. Anyway, it's just an example.

```go
package main

import (
	"fmt"
)

type Validation interface{
	Create(newObj string)
	Update(oldObj string, newObj string)
	Name() string
}

type Validator interface{
	Validate(setting string)
}

type validation struct{
	validator Validator
	name string
}

func (v *validation) Name() string {
	return v.name
}

func (v *validation) Create(newObj string) {
	v.validator.Validate(newObj)
}


func (v *validation) Update(_, newObj string) {
	v.validator.Validate(newObj)
}

// -----------------------

type SupportBundleTimeoutSetting struct {}

func NewSupportBundleTimeoutSetting() Validation {
	return &validation{
		validator: &SupportBundleTimeoutSetting{},
		name: "SupportBundleTimeoutSettingName",
	}
}

func (sb *SupportBundleTimeoutSetting) Validate(setting string) {
	fmt.Println("validate setting")
}


type SupportBundleExpirationSetting struct {}

func NewSupportBundleExpirationSetting() Validation {
	return &validation{
		validator: &SupportBundleExpirationSetting{},
		name: "SupportBundleExpirationSettingName",
	}
}

func (sb *SupportBundleExpirationSetting) Validate(setting string) {
	fmt.Println("validate setting")
}


// ----------------------------

func main() {
	m := make(map[string]Validation)

	settings := []Validation{
		NewSupportBundleTimeoutSetting(),
		NewSupportBundleExpirationSetting(),
	}

	for _, setting := range settings {
		m[setting.Name()] = setting
	}

	// check create or update
	if v, ok := m["SupportBundleTimeoutSettingName"]; ok {
		v.Create("setting")
		v.Update("oldSetting", "newSetting")
	}
}
```

FrankYang0529:
Hi @Yu-Jack, thanks for the great suggestion. I think we can create another issue and update it without HEP, because this refactor doesn't break any interface. 👍

bk201:
> Hi @Yu-Jack, thanks for the great suggestion. I think we can create another issue and update it without HEP, because this refactor doesn't break any interface. 👍

That's a good suggestion and we should refactor toward this direction.

bk201 (approved):

FrankYang0529:
@Mergify backport v1.4

Issue #5774, opened 2024-05-08
[BUG] The vGPU allocatable count for a node isn't updated after cluster creation


noahgildersleeve:
**Describe the bug**

When you create a RKE2 cluster from an upstream Rancher cluster it doesn't update the vGPU allocatable list `node.status.allocatable` after the cluster creation
**To Reproduce**
1. Create 4 vGPU profiles on a node
2. deploy a 3 node RKE2 cluster with that vGPU profile
3. Try to deploy out another 3 node RKE2 cluster and check allocatable 

**Expected behavior**

The allocatable vGPU count should update 
**Support bundle**

[supportbundle_314d015f-bac6-449f-a676-14fd3732e576_2024-05-08T22-16-49Z.zip](https://github.com/harvester/harvester/files/15255792/supportbundle_314d015f-bac6-449f-a676-14fd3732e576_2024-05-08T22-16-49Z.zip)

**Environment**
 - Harvester ISO version: v1.3.0
 - Rancher version: v2.8-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 2 node DL360

**Additional context**
This is the Harvester side bug for [10948](https://github.com/rancher/dashboard/issues/10948)

![Greenshot 2024-05-03 17 55 22](https://github.com/harvester/harvester/assets/83787952/7954c237-6a9c-4057-b41e-c448f7a345e1)

ibrokethecloud:
we should improve the device plugin pcidevices controller to monitor resource usage: https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#monitoring-device-plugin-resources

additional changes may be needed in the harvester api to federate plugin info from all nodes for rancher to leverage

bk201:
require-doc
- doc: update vGPU doc and mention the feature is available after Rancher version xxx
- release note: mention the feature is available after Rancher version xxx

noahgildersleeve:
Tested with Harvester v1.4.0-rc2 with Rancher versions

- v2.8-68e5d43e449fcd3ef73baf1caee98da6f0104eed-head
![Greenshot 2024-10-09 17 47 34](https://github.com/user-attachments/assets/e60d594d-0f23-4b0d-bde6-8caec0d0cff6)
![Greenshot 2024-10-09 17 46 08](https://github.com/user-attachments/assets/62da3e9b-f926-41fd-bbc9-69e034501f8c)
[...]

PR #6096, opened 2024-07-01
Improve tracking of device allocation

ibrokethecloud:

**Problem:**

Currently there is no way to track the device allocation across VM's in a cluster since the device plugin only reports the allocatable count based on health of devices and is not responsible for tracking the allocation.

Harvester v1.3.x introduced support for vgpu devices and the associated rancher release allows end users to define vGPU's directly during downstream cluster provisioning. However there is no easy way to track current device allocation on the underlying cluster. As a result end user may be able to create VMs with vGPU device requests which can no longer be satisfied by the cluster. As a result the VM is unschedulable, and eventually rancher keeps reprovisioning the VM.

**Solution:**

The PR introduces a new api endpoint for `/v1/harvester/cluster` which allows rancher to query the current device free list across all nodes from the cluster, and results the response as a json. This can then be used by the UI to filter on devices to gate creation of VM's

A sample json response from harvester looks as follows:

```
(⎈|local:harvester-system)➜  ~ curl -sk -u $APITOKEN "${URL}/v1/harvester/cluster/local?link=deviceCapacity"  | jq .
{
  "cpu": "32",
  "devices.kubevirt.io/kvm": "1k",
  "devices.kubevirt.io/tun": "1k",
  "devices.kubevirt.io/vhost-net": "1k",
  "ephemeral-storage": "573222064908",
  "hugepages-1Gi": "0",
  "hugepages-2Mi": "0",
  "intel.com/X540_ETHERNET_CONTROLLER_VIRTUAL_FUNCTION": "2",
  "memory": "263873540Ki",
  "nvidia.com/GA100_A100_PCIE_40GB": "0",
  "nvidia.com/GRID_A100-10C": "0",
  "nvidia.com/GRID_A100-20C": "-1",
  "nvidia.com/GRID_A100-40C": "0",
  "nvidia.com/GRID_A100-4C": "0",
  "pods": "200"
}
```
**Related Issue:**
https://github.com/harvester/harvester/issues/5774
**Test plan:**

### DIFF
--- a/pkg/api/cluster/formatter.go
+++ b/pkg/api/cluster/formatter.go
@@ -1,0 +1,93 @@
+package cluster
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	corev1 "k8s.io/api/core/v1"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/rancher/apiserver/pkg/apierror"
+	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
+	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+)
+
+type Handler struct {
+	vmCache   ctlkubevirtv1.VirtualMachineCache
+	nodeCache ctlcorev1.NodeCache
+}
+
+func (h Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	if err := h.do(rw, req); err != nil {
+		status := http.StatusInternalServerError
+		if e, ok := err.(*apierror.APIError); ok {
+			status = e.Code.Status
+		}
+		rw.WriteHeader(status)
+		_, _ = rw.Write([]byte(err.Error()))
+		return
+	}
+	rw.WriteHeader(http.StatusNoContent)
+}
+
+func (h Handler) do(rw http.ResponseWriter, _ *http.Request) error {
+
+	nodeDeviceAvailability, err := h.generateDeviceAvailability()
+	if err != nil {
+		return err
+	}
+	result, err := json.Marshal(nodeDeviceAvailability)
+	if err != nil {
+		return fmt.Errorf("unable to marshal node device capacity: %v", err)
+	}
+	_, err = rw.Write(result)
+	return err
+}
+
+func (h Handler) generateDeviceAvailability() (map[string]resource.Quantity, error) {
+	nodes, err := h.nodeCache.List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("error looking up nodes from nodeCache: %v", err)
+	}
+
+	vmList, err := h.vmCache.List(corev1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("error looking up vms from vmCache: %v", err)
+	}
+
+	return calculateAllocation(nodes, vmList), nil
+}
+
+func calculateAllocation(nodes []*corev1.Node, vms []*kubevirtv1.VirtualMachine) map[string]resource.Quantity {
+	nodeDeviceAvailability := make(map[string]resource.Quantity)
+	for _, node := range nodes {
+		for resourceName, quantity := range node.Status.Allocatable {
+			existingQuantity := nodeDeviceAvailability[resourceName.String()]
+			existingQuantity.Add(quantity)
+			nodeDeviceAvailability[resourceName.String()] = existingQuantity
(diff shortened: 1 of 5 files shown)


pkg/api/cluster/formatter.go
```diff
@@ -0,0 +1,93 @@
...
+	return err
+}
+
+func (h Handler) generateDeviceAvailability() (map[string]resource.Quantity, error) {
+	nodes, err := h.nodeCache.List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("error looking up nodes from nodeCache: %v", err)
+	}
+
+	vmList, err := h.vmCache.List("", labels.Everything())
```
tserong: I've noticed a few other places where we use `corev1.NamespaceAll` instead of `""` in this context, e.g.:

```suggestion
	vmList, err := h.vmCache.List(corev1.NamespaceAll, labels.Everything())
```

pkg/api/cluster/formatter_test.go
```diff
@@ -0,0 +1,118 @@
...
+	nodes = []*corev1.Node{node1, node2}
+	vms   = []*kubevirtv1.VirtualMachine{vm1, vm2, vm3, vm4}
+)
+
+func Test_calculateAllocation(t *testing.T) {
+	assert := require.New(t)
+	nodeDeviceAvailability := calculateAllocation(nodes, vms)
+	assert.Equal(nodeDeviceAvailability["nvidia.com/GRID_A100-20C"], *resource.NewQuantity(1, resource.DecimalSI), "expect to find 1 A100-20C device")
+	assert.Equal(nodeDeviceAvailability["nvidia.com/GRID_A100-10C"], *resource.NewQuantity(3, resource.DecimalSI), "expect to find 3 A100-10C device")
+	assert.Equal(nodeDeviceAvailability["intel.com/X540_ETHERNET_CONTROLLER_VIRTUAL_FUNCTION"], *resource.NewQuantity(30, resource.DecimalSI), "expect to find 12 X540_ETHERNET_CONTROLLER_VIRTUAL_FUNCTION VF's")
```
tserong:
```suggestion
	assert.Equal(nodeDeviceAvailability["intel.com/X540_ETHERNET_CONTROLLER_VIRTUAL_FUNCTION"], *resource.NewQuantity(30, resource.DecimalSI), "expect to find 30 X540_ETHERNET_CONTROLLER_VIRTUAL_FUNCTION VF's")
```

tserong (approved):
LGTM, just a couple of little nits

bk201 (approved):

pkg/api/cluster/schema.go
```diff
@@ -0,0 +1,78 @@
...
+			{
+				Type: "cluster",
+				ID:   "local",
+				Object: Cluster{
+					Name: "local",
+				},
+			},
+		},
+	}, nil
+}
```
Yu-Jack: we should keep schema clear, we can put this store into another file called `store.go` like other one does.

ibrokethecloud: sounds good..

WebberHuang1118 (approved):

Yu-Jack (approved):
LGTM, thanks!

w13915984028:
A question:

The `nodeDeviceAvailability` is a summary of the cluster, it will have a possible drawback:

`cluster level satisfaction != node level satisfaction`, the deployment of VM can still fail.

ibrokethecloud:
vGPU's are treated similar to CPU and if same while allocation we are not pinning the VM to a node, but letting scheduler decide the node where the requirement can be met.

w13915984028:
e.g. Node A has 2 vGPU, node B has 2 vGPU, the VM asks for 4 vGPU, from cluster level, it meets.

ibrokethecloud:
a VM cannot ask for more than 1 VGPU :D

w13915984028:
1. There are many different devices, not all of them can only be allocated 1.
2. Rancher needs to slow down itself when fail continue happens. (Not sure it already does)
3. Both k8s deployment or kubevirt VM, they fail on sheduling, not on object creation.
4. If Rancher gets a list of devices, then it or Harvester driver/plugin needs to understand the logic and try to fail earlier.

w13915984028:
Another case, a guest cluster Machine/node is asking for 64 cpu cores, and each Harvester node has 32 cores.

If Rancher trys to fail earler, it needs the list of resources on each node and then check.

ibrokethecloud:
@w13915984028 this api will only be currently used for tracking the pcidevice/vgpu allocation to allow overprovisioning as these resources cannot be overcommited. There is no plan to use this check for non device resources allocation requests.

ibrokethecloud:
@mergifyio backport v1.4

ibrokethecloud:
@mergifyio backport v1.3

ibrokethecloud:
@mergifyio backport v1.4

Issue #5695, opened 2024-04-26
[BUG] overcommit-config defaults are not applied


connorkuehl:
**To be honest, I am not 100% sure that this is a bug. I can see why it would be useful to err on the side of not overcommitting anything out of the box. Perhaps this is more of a UI/UX/docs issue regarding the behavior of this setting.**

---

**Describe the bug**

The overcommit-config setting's default settings that are shown on the dashboard are not automatically applied, as we can see here the Longhorn overcommit percentage is not the default 200:

![before](https://github.com/harvester/harvester/assets/5227338/b53ed6b7-fc31-4af9-9a60-85944d2c9bdd)

Unless you click "save", in which case, they are:

![after](https://github.com/harvester/harvester/assets/5227338/59006ea3-52e8-4523-bcd0-502102dd36f0)

The fact that these values are prepopulated makes it seem as though they should be applied out of the box.

**To Reproduce**
1. Install Harvester 1.3.0
2. Check the overcommit-config
3. Look at Longhorn dashboard, compare Longhorn's overprovisioning percentage to the overcommit-config storage value

**Expected behavior**

In this case, the default storage value for overcommit-config is 200, so I would expect Longhorn's overprovisioning percentage to be 200.

**Support bundle**

**Environment**
 - Harvester ISO version: Harvester 1.3.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

In the case of the Longhorn storage overcommit value not being propagated, that is because Harvester only pays attention to the setting.Value, and doesn't override it with the setting.Default if setting.Value is not set.

https://github.com/harvester/harvester/blob/8958ebee30387c8591eeb17ddf60e178517e9de3/pkg/controller/master/setting/overcommit_config.go#L15-36

janeczku:
It appears that Harvester 1.3.0 has storage over-commit disabled by default (since the Longhorn `StorageOverProvisioningPercentage` setting defaults to 100%).

This contradicts the documentation, which [clearly states](https://docs.harvesterhci.io/v1.3/vm/resource-overcommit) that Harvester uses a storage over commit ration of 200% by default.

This issue should be clearly marked in the documentation and/or release notes. @innobead

tserong:
[...]

PR #6113, opened 2024-07-03
Fix storage overcommit default not applied to Longhorn

brandboat:
**Problem:**
Harvester storage overcommit-config is set to 200% while the value not applied to Longhorn  `StorageOverProvisioningPercentage` which is still 100%.

**Solution:**
Take care `Default` value in  https://github.com/harvester/harvester/blob/6b79086f42d005c19070b760a877fe27ccb4767a/pkg/controller/master/setting/overcommit_config.go#L17
And sync storage overcommit-config during harvester bootstrap.

**Related Issue:**
https://github.com/harvester/harvester/issues/5695

**Test plan:**
1. Setup harvester with one node (easy for testing), and go to longhorn dashboard > setting tab, check `Storage Over Provisioning Percentage` is 200 instead of 100.
2. Setup harvester 1.3.1 with one node (easy for testing), upgrade to the harvester version contains this fix, and go to longhorn dashboard > setting tab, check `Storage Over Provisioning Percentage` is 200 instead of 100.
image

### DIFF
--- a/pkg/controller/master/setting/handler.go
+++ b/pkg/controller/master/setting/handler.go
@@ -37,6 +37,9 @@ var (
 	bootstrapSettings = []string{
 		settings.SSLCertificatesSettingName,
 		settings.KubeconfigDefaultTokenTTLMinutesSettingName,
+		// The Longhorn storage over-provisioning percentage is set to 100, whereas Harvester uses 200.
+		// This needs to be synchronized when Harvester starts.
+		settings.OvercommitConfigSettingName,
 	}
 	skipHashCheckSettings = []string{
 		settings.AutoRotateRKE2CertsSettingName,
--- a/pkg/controller/master/setting/overcommit_config.go
+++ b/pkg/controller/master/setting/overcommit_config.go
@@ -14,7 +14,13 @@ import (
 
 func (h *Handler) syncOvercommitConfig(setting *harvesterv1.Setting) error {
 	overcommit := &settings.Overcommit{}
-	if err := json.Unmarshal([]byte(setting.Value), overcommit); err != nil {
+	var value string
+	if setting.Value != "" {
+		value = setting.Value
+	} else {
+		value = setting.Default
+	}
+	if err := json.Unmarshal([]byte(value), overcommit); err != nil {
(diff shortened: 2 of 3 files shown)


bk201 (approved):

bk201:
@mergifyio backport v1.2 v1.3

w13915984028 (commented):
two questions about the overcommit, may out of the scope of this PR
(1) the validator only checks if the overcommit < 100, but no up limit check; if user sets an invalid high value, Harvester validator does not deny it.
(2) for upgrade path, can the default value also be synced to LH?
thanks.

brandboat:
> (1) the validator only checks if the overcommit < 100, but no up limit check; if user sets an invalid high value, Harvester validator does not deny it.

Sounds good !
This could be handled in separate issue, but how do we define "invalid high value" ? Any suggestions ?

> (2) for upgrade path, can the default value also be synced to LH?
thanks.

Yes, during the upgrade, the controller will restart and trigger `syncOvercommitConfig` to update Longhorn storage overcommit setting. And thanks, you remind me I need to add the upgrade part in test plan.

tserong:
> > (1) the validator only checks if the overcommit < 100, but no up limit check; if user sets an invalid high value, Harvester validator does not deny it.
> 
> Sounds good ! This could be handled in separate issue, but how do we define "invalid high value" ? Any suggestions ?

Good question. I thought to check the longhorn source, but that only specifies a minimum, not a maximum (and the minimum there is zero which seems weird to me): https://github.com/longhorn/longhorn-manager/blob/4d90cc71fb5fbb33008bb9de870f2e8cb018af58/types/setting.go#L524-L535

Anyway, as you both already said, that's really a separate issue :-)

tserong (approved):

FrankYang0529 (approved):
LGTM. Thanks for the fix.

bk201:
@w13915984028 Feel free to open a new issue for the highest value.

FrankYang0529:
> @w13915984028 Feel free to open a new issue for the highest value.

We can do this after https://github.com/harvester/harvester/pull/6088.

Issue #5618, opened 2024-04-17
[ENHANCEMENT] Exporting a volume to image API should respond with the image definition


drewmullen:
**Is your enhancement related to a problem? Please describe.**

I need to know the name of the image thats created when I export a volume to an image. Currently this API only generates a name using a `image-` prefix and accepts a display name (non unique). It gives me no information (empty `204` reply) on success. 

**There is no way for me to programmatically track the status of an exported image upon creation**

**Describe the solution you'd like**

`/v1/harvester/persistentvolumeclaims/drew/packer-build?action=export` should respond with the image definition, including the name of the new image.

Bonus: it would be nice to specify a `name` or the name-prefix used instead of relying on the default `image-`

bk201:
@WebberHuang1118 please help look into this.

WebberHuang1118:
I had an implementation [that could report the created VM image from volume exporting.](https://github.com/WebberHuang1118/harvester/commit/dc847746bf5c37461fd2ae9f4bb3faace0bc7e7b)
![export_volume](https://github.com/harvester/harvester/assets/19387129/10e05cb2-8bfe-4268-8da9-f5c71841c93a)

however, in the current design, volume actions [do not write any information on the ResponseWriter](https://github.com/harvester/harvester/blob/master/pkg/api/volume/handler.go#L45C1-L56C2).

If we don't want to treat `exporting volume` as a special case and keep the consistency as usual, certain refactor on all volume actions needs to be considered, thanks. @bk201

innobead:
> If we don't want to treat exporting volume as a special case and keep the consistency as usual, certain refactor on all volume actions needs to be considered, thanks. @bk201

If the original response of any volume API is empty, why would adding this cause some refactoring? I thought the response is never handled now since it's not generated.

WebberHuang1118:
Hi @innobead,
In the current design, volume actioners only write `no content status code` in responseWriter, and there is no interface to obtain additional information, [this is implemented in a shared function](https://github.com/harvester/harvester/blob/master/pkg/api/volume/handler.go#L45C1-L56C2) for all actions.

[...]

PR #6138, opened 2024-07-08
Feat: exporting volume returns the image data in response body

Yu-Jack:
## Problem
Some APIs doesn't return the response body, but some do L476 in example. In current design, we need to customize the logic to write the response body in the business logic.

https://github.com/harvester/harvester/blob/026d49fa55d465c43818068240c2ea8c2a48ea88/pkg/api/vm/handler.go#L458-L478

 It breaks the architecture consistency, and writes header twice. I think we need a consistent way to serve those APIs.

### Proposal 1 - This PR

Create a new interface and common function to serve all APIs like below.

```go
// 204 case
func businessLogic(w http.ResponseWriter, r *http.Request) (interface{}, error) {
	return nil, nil // -> 204
}

// 200 case-1
func businessLogic(w http.ResponseWriter, r *http.Request) (interface{}, error) {
	return obj1, nil // -> 200 with content
}

// 200 case-2
func businessLogic(w http.ResponseWriter, r *http.Request) (interface{}, error) {
	return struct{}{}, nil // -> 200 without content
}

// error case
func businessLogic(w http.ResponseWriter, r *http.Request) (interface{}, error) {
	return apierror.NewAPIError(validation.NotFound, "Not found resource: {resource name} ") // -> 404
}
```

If we'd like to return 200 status code with a empty response, we could return `struct{}{}`, then handle it, let it return status 200 with empty response body. (This one is for current `healthy_handler.go`.)

---

### Proposal 2 - https://github.com/Yu-Jack/harvester/commit/96e7cd3dd684e6d1e99a1047bc664681556bc000

Use simple ctx to wrap everything, inspired some Golang http framework.

```go
type harvesterCutomCtx struct {
	body       interface{}
	statusCode int
	rw         http.ResponseWriter
	req        *http.Request
}

// 204 case
func businessLogic(ctx *harvesterCutomCtx) (error) {
	ctx.SetStatus(204) -> return 204
	return nil
}

// 200 case-1
func businessLogic(ctx *harvesterCutomCtx) (error) {
	ctx.SetBody("hi") -> return 200 with content
	return nil
}

// 200 case-2
func businessLogic(ctx *harvesterCutomCtx) (error) {
	ctx.SetStatus(200) -> return 200 without content
	return nil
}

// error case
func businessLogic(ctx *harvesterCutomCtx) (error) {
	return apierror.NewAPIError(validation.NotFound, "Not found resource: {resource name} ") // -> 404
}
```

This one would be more flexible if we want more control.

---

## Discussion
Not sure we should change all APIs in this time, or we change it over time?

## Related Issue
https://github.com/harvester/harvester/issues/5618

## Test plan
Current all test cases should be successful like https://github.com/harvester/harvester/pull/6138#issuecomment-2219519324, including api and integration tests.

### DIFF
--- a/pkg/server/http/ctx.go
+++ b/pkg/server/http/ctx.go
@@ -1,0 +1,24 @@
+package http
+
+import "net/http"
+
+type Ctx struct {
+	statusCode int
+	rw         http.ResponseWriter
+	req        *http.Request
+}
+
+func newDefaultHarvesterServerCtx(rw http.ResponseWriter, req *http.Request) *Ctx {
+	return &Ctx{
+		statusCode: http.StatusNoContent, // default is 204
+		rw:         rw,
+		req:        req,
+	}
+}
+
+func (ctx *Ctx) SetStatus(statusCode int) { ctx.statusCode = statusCode }
+func (ctx *Ctx) StatusCode() int          { return ctx.statusCode }
+func (ctx *Ctx) Req() *http.Request       { return ctx.req }
+func (ctx *Ctx) RespWriter() http.ResponseWriter {
+	return ctx.rw
+}
--- a/pkg/server/http/handler.go
+++ b/pkg/server/http/handler.go
@@ -1,0 +1,46 @@
+package http
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/rancher/apiserver/pkg/apierror"
+
+	"github.com/harvester/harvester/pkg/util"
+)
+
+type HarvesterServerHandler interface {
+	Do(ctx *Ctx) (interface{}, error)
+}
+
+type harvesterServerHandler struct {
+	httpHandler HarvesterServerHandler
+}
+
+func NewHandler(httpHandler HarvesterServerHandler) http.Handler {
+	return &harvesterServerHandler{
+		httpHandler: httpHandler,
+	}
+}
+
+func (handler *harvesterServerHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	ctx := newDefaultHarvesterServerCtx(rw, req)
+	resp, err := handler.httpHandler.Do(ctx)
+	if err != nil {
+		status := http.StatusInternalServerError
+		var e *apierror.APIError
+		if errors.As(err, &e) {
+			status = e.Code.Status
+		}
+		rw.WriteHeader(status)
+		_, _ = rw.Write([]byte(err.Error()))
+		return
+	}
+
+	if resp == nil {
+		rw.WriteHeader(ctx.StatusCode())
+		return
+	}
+
+	util.ResponseOKWithBody(rw, resp)
+}
--- a/pkg/api/volume/handler.go
+++ b/pkg/api/volume/handler.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"net/http"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -26,6 +25,7 @@ import (
 	ctllhv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
 	ctlsnapshotv1 "github.com/harvester/harvester/pkg/generated/controllers/snapshot.storage.k8s.io/v1"
 	"github.com/harvester/harvester/pkg/ref"
+	harvesterServer "github.com/harvester/harvester/pkg/server/http"
 	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
 	indexeresutil "github.com/harvester/harvester/pkg/util/indexeres"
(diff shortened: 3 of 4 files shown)


Yu-Jack:
I pushed another version, which changed all APIs (https://github.com/Yu-Jack/harvester/commit/5f4e0e962c6ed3f8ebd2e107381e1bc69dc1599a). Then, I did some e2e api tests with skipping some test cases, like deleting hosts and integration tests. It ended up not bad. But, we still need full testing just in case.

image

innobead:
The flexibility of option 2 is good, but this implementation of option 1 seems better for us as it has a convention that can provide expected results for callers.

In general, LGTM. However, I would recommend having a consistent payload for CRUD. For example, for the create/update/patch of a resource, always reply with its metadata. Or, we can just simply refer to how K8s API usually does.

Yu-Jack:
@innobead I think we could combine two options. Use option 1 to have a convention to return response. Use option 2 to provide more flexibilities to customize some fields, like status code and headers etc.

Yu-Jack:
Gentle ping @bk201 @WebberHuang1118. Please help to check this ideal when you're free, thank you!

pkg/util/http.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 )
 
+func WriteResponseBody(rw http.ResponseWriter, obj interface{}) {
```
FrankYang0529: It looks like this function is unused?

pkg/server/http/ctx.go
```diff
@@ -0,0 +1,24 @@
+package http
+
+import "net/http"
+
+type Ctx struct {
+	body       interface{}
```
FrankYang0529: It looks like this variable is unused.

pkg/server/http/handler.go
```diff
@@ -0,0 +1,46 @@
...
+		if errors.As(err, &e) {
+			status = e.Code.Status
+		}
+		rw.WriteHeader(status)
+		_, _ = rw.Write([]byte(err.Error()))
+		return
+	}
+
+	if resp == nil {
+		rw.WriteHeader(ctx.statusCode)
```
FrankYang0529: If we use `Req` to get `ctx.req` and `RespWriter` to get `ctx.rw`, do we want to add a new function `StatusCode` to return `ctx.statusCode`?

FrankYang0529 (commented):
Hi @Yu-Jack, thanks for the PR. Will we apply common `ServerHTTP` function to all APIs?

Yu-Jack:
> Hi @Yu-Jack, thanks for the PR. Will we apply common `ServerHTTP` function to all APIs?

@FrankYang0529  Sure, I'll. This PR is a concept PR, which I just modified only one API for demo. If everyone is good with this solution, I'll modify all APIs.

FrankYang0529 (approved):
LGTM. Thanks.

bk201 (approved):
LGTM, thanks!

pkg/server/http/ctx.go
```diff
@@ -0,0 +1,24 @@
...
+		req:        req,
+	}
+}
+
+func (ctx *Ctx) SetStatus(statusCode int) { ctx.statusCode = statusCode }
+func (ctx *Ctx) StatusCode() int          { return ctx.statusCode }
+func (ctx *Ctx) Req() *http.Request       { return ctx.req }
+func (ctx *Ctx) RespWriter() http.ResponseWriter {
+	return ctx.rw
+}
```
WebberHuang1118, line 24: Hi @Yu-Jack ,
Do we need this? If yes, is there any example? thanks.

WebberHuang1118 (approved):
LGTM, jus a question, thanks.

pkg/server/http/ctx.go
Yu-Jack: @WebberHuang1118 

https://github.com/harvester/harvester/blob/3123cb9b11f7185e9218e954212bcb2ae2d945f7/pkg/api/image/formatter.go#L129-L137

Yu-Jack:
This is an example of response.

```json
{
    "kind": "VirtualMachineImage",
    "apiVersion": "harvesterhci.io/v1beta1",
    "metadata": {
        "name": "image-jjnlm",
        "generateName": "image-",
        "namespace": "default",
        "uid": "cd562ef7-7760-4912-9c5b-d75bd11aa9c2",
        "resourceVersion": "25252817",
        "generation": 1,
        "creationTimestamp": "2024-09-03T03:45:29Z",
        "annotations": {
            "harvesterhci.io/storageClassName": "harvester-longhorn"
        },
        "managedFields": [
            {
                "manager": "harvester",
                "operation": "Update",
                "apiVersion": "harvesterhci.io/v1beta1",
                "time": "2024-09-03T03:45:29Z",
                "fieldsType": "FieldsV1",
                "fieldsV1": {
                    "f:metadata": {
                        "f:annotations": {
                            ".": {},
                            "f:harvesterhci.io/storageClassName": {}
                        },
                        "f:generateName": {}
                    },
                    "f:spec": {
                        ".": {},
                        "f:checksum": {},
                        "f:displayName": {},
                        "f:pvcName": {},
                        "f:pvcNamespace": {},
                        "f:retry": {},
                        "f:sourceType": {},
                        "f:url": {}
                    },
                    "f:status": {
                        ".": {},
                        "f:failed": {}
                    }
                }
            }
        ]
    },
    "spec": {
        "displayName": "test",
        "sourceType": "export-from-volume",
        "pvcName": "test-disk-0-bwsso",
        "pvcNamespace": "default",
        "url": "",
        "checksum": "",
        "storageClassParameters": {
            "migratable": "true",
            "numberOfReplicas": "3",
            "staleReplicaTimeout": "30"
        },
        "retry": 0
    },
    "status": {
        "failed": 0
    }
}
```

PR #6149, opened 2024-07-10
Introduce scheduled VM backups/snapshots

WebberHuang1118:
**Description:**
Introduce scheduled VM backup, related HEP:https://github.com/harvester/harvester/pull/6008

**Solution:**
Implement schedule VM backup HEP

**Related Issue:**
#2756

**Test Plan: Backup Type**
- Prerequisites
  - Building Harvester Cluster with the master-head branch (`09f3ca` in my env)
  - Adding `ScheduleVMBackup` CRD
    - `$ kubectl apply -f deploy/charts/harvester-crd/templates/harvesterhci.io_schedulevmbackups.yaml` 
  - Updating `harvester` deployment from this PR
  - Updating `harvester-webhook` deployment from this PR
  - Configuring a valid S3 backup target
  - Creating and starting a VM `vm` in the default namespace

- Case0: Basic validation for CR `ScheduleVMBackup`
  - It will be rejected if the spec fields contain the following conditions
    - `spec.cron` with invalid cron format
    -  `spec.retain` < 2 or > 250
    - `spec.maxFailure` < 2 or >= `spec.retain` 
    - Backup target is not reachable (if the schedule is for backup type)

- Case1: Checking the backup rotation
  - Apply manifest to create a scheduled VM backup
    ```
    kind: ScheduleVMBackup
    metadata:
      annotations:
          harvesterhci.io/svmbackupSkipCronCheck: "enable"
      namespace: default
      name: svmbackup-vm
    spec:
      cron: '*/2 * * * *'
      retain: 3
      maxFailure: 2
      vmbackup:
        source:
          apiGroup: kubevirt.io
          kind: VirtualMachine
          name: vm
        type: backup
     ```
  - Checking new backups keep being created every two minutes
    - The Harvester always maintains at least three VM backups since the 6th minute
    - The oldest VM backup will be deleted as new VM backup completes
    - The oldest Longhorn snapshot should also be purged under the hood

- Case2: Scheduled backup for a stopped VM
  - Continuing the steps from Case1
  - Stopping VM `vm`
  - Checking new backups keep being created every two minutes
    - Related resources should keep being updated

- Case3: Scheduled backup keeps processing if temporal backup failure happens
  - Continuing the steps from Case1
  - Making the backup target unreachable
  - After a failure backup is discovered
    - `ScheduleVMBackup` CR `.status.failure = 1`
  - Making the backup target healthy
  - Checking new backups keep being created every two minutes
    - `ScheduleVMBackup` CR `.status.failure = 0`
    - Failure backup is removed
      - Not found in `ScheduleVMBackup` CR `.status.vmbackupInfo`

- Case4: #Failure backups reach threshold
  - Continuing the steps from Case3
  - Making the backup target unreachable
  - After around 8th minute
  - Schedules backup should be suspended
    -  `ScheduleVMBackup` CR `.status.failure = 2`
  - Checking new backups would not be created
    - `ScheduleVMBackup` CR `.status.suspend = true`
    - `ScheduleVMBackup` CR `.status.condition` contains the message `Reach Max Failure`
    - No more new VM backup is going to be created in this phase

- Case5: Resuming suspended backup schedule
  - Continuing the steps from Case4
  - Making the backup target healthy again
  - Setting `ScheduleVMBackup` CR `.spec.suspend = false`
  - Checking the backup schedule resume correctly
    - `ScheduleVMBackup` CR `.status.suspend = false`
    - `ScheduleVMBackup` CR `.status.failure = 0`
    - Checking the backup schedule resume correctly
      - New backups keep being created every two minutes

**Test Plan: Snapshot Type**
- Case1: Checking the snapshot rotation
  - Setting the backup target to default
  - Apply manifest to create a scheduled VM backup
    ```
    kind: ScheduleVMBackup
    metadata:
      annotations:
          harvesterhci.io/svmbackupSkipCronCheck: "enable"
      namespace: default
      name: svmbackup-snap
    spec:
      cron: '*/2 * * * *'
      retain: 3
      maxFailure: 2
      vmbackup:
        source:
          apiGroup: kubevirt.io
          kind: VirtualMachine
          name: vm
        type: snapshot
     ```
  - CR `ScheduleVMBackup` wouldn't be rejected even if the backup target is not reachable
  - Checking new snapshots keep being created every two minutes
    - The Harvester always maintains at least three VM snapshots since the 6th minute
    - The oldest VM snapshots will be deleted as new VM snapshot completes
    - The oldest Longhorn snapshots should also be purged under the hood

**Test Plan: Duplated  Schedule for the same VM**
- Creating two CR `ScheduleVMBackup` targets on the same VM
- The second one would be rejected by webhook
  - With the message `The request is invalid: spec.vmbackup: VM <VM> already has backup schedule` 

**Test Plan: Proactive Suspend/Resume the Schedule**
- Starting a scheduled VM backup or snapshot
- Editing the ScheduleVMBackup resource
  - As `spec.suspend` = true
  - No more new VM backups are created
- Editing the ScheduleVMBackup resource
  - As `spec.suspend` = false
  - New VM backup keeps being created in every two minutes

**Test Plan: Runtime update schedule cron expression**
- Creating a ScheduleVMBackup resource with backup or snapshot type
  - With `spec.vmbackup` points to the VM
  - With `spec.retain` = 3
  - With `spec.cron = */2 * * * *`
  - New VM backup keeps being created in every two minutes
- Editing the ScheduleVMBackup resource
  - As `spec.cron = */3 * * * *`
  - New VM backup keeps being created in every three minutes

**Test Plan: Schedule Granularity Check**
- Case1: If the schedule's granularity is less than one hour, it can't be applied
  - Apply manifest to create a scheduled VM backup
    ```
    kind: ScheduleVMBackup
    metadata:
      namespace: default
      name: svmbackup-vm1
    spec:
      cron: '*/2 * * * *'
      retain: 3
      maxFailure: 2
      vmbackup:
        source:
          apiGroup: kubevirt.io
          kind: VirtualMachine
          name: vm1
        type: backup
     ```
  - The creation would be ejected by webhook with `The request is invalid: spec.cron: schedule granularity 1m0s less than 1h0m0s`
- Case2: If the schedule's granularity is equal or more than one hour, it could be applied
  - Apply manifest to create a scheduled VM backup
    ```
    kind: ScheduleVMBackup
    metadata:
      namespace: default
      name: svmbackup-vm1
    spec:
      cron: '0 * * * *'
      retain: 3
      maxFailure: 2
      vmbackup:
        source:
          apiGroup: kubevirt.io
          kind: VirtualMachine
          name: vm1
        type: backup
     ```
  - The creation should be a success
- Case3: If two schedules have the same granularity and each iteration's timing offset is less than 10mins, the creation would be ejected
  - Following above plan
  - Apply manifest to create a new scheduled VM backup
    ```
    kind: ScheduleVMBackup
    metadata:
      namespace: default
      name: svmbackup-vm2
    spec:
      cron: '5 * * * *'
      retain: 3
      maxFailure: 2
      vmbackup:
        source:
          apiGroup: kubevirt.io
          kind: VirtualMachine
          name: vm2
        type: backup
     ```
  - The creation would be ejected by webhook with `The request is invalid: spec.cron: same granularity as schedule svmbackup-vm1 and offset less than 10m0s`
- Case4: If two schedules have the same granularity and each iteration's timing offset is *not* less than 10mins, the creation could be applied
  - Following above plan
  - Apply manifest to create a new scheduled VM backup
    ```
    kind: ScheduleVMBackup
    metadata:
      namespace: default
      name: svmbackup-vm2
    spec:
      cron: '10 * * * *'
      retain: 3
      maxFailure: 2
      vmbackup:
        source:
          apiGroup: kubevirt.io
          kind: VirtualMachine
          name: vm2
        type: backup
     ```
  - The creation should be a success
 
**Additional Context:**
- We can watch the VM backup list with:
  - `$ kubectl get schedulevmbackups.harvesterhci.io svmbackup-vm --watch -o json | jq -r .status`
- Since the GUI PR https://github.com/harvester/dashboard/pull/1103 is merged, the setting and getting operation for CR `ScheduleVMBackup` could be derived from GUI
- Annotation ` harvesterhci.io/svmbackupSkipCronCheck: "enable"` is used to skip granularity checking


pkg/controller/master/image/register.go
```diff
@@ -11,6 +11,7 @@ import (
 const (
 	vmImageControllerName      = "vm-image-controller"
 	backingImageControllerName = "backing-image-controller"
+	cronJobControllerName      = "cron-job-controller"
```
FrankYang0529: This variable is unused.

pkg/webhook/resources/schedulevmbackup/validator.go
```diff
@@ -0,0 +1,169 @@
...
+			return fmt.Errorf("can't get backup target secret: %s/%s, error: %w",
+				util.LonghornSystemNamespaceName, util.BackupTargetSecretName, err)
+		}
+		os.Setenv(backup.AWSAccessKey, string(secret.Data[backup.AWSAccessKey]))
+		os.Setenv(backup.AWSSecretKey, string(secret.Data[backup.AWSSecretKey]))
+		os.Setenv(backup.AWSEndpoints, string(secret.Data[backup.AWSEndpoints]))
+		os.Setenv(backup.AWSCERT, string(secret.Data[backup.AWSCERT]))
+	}
+
+	_, err = backupstore.GetBackupStoreDriver(backup.ConstructEndpoint(target))
```
FrankYang0529: We can use `util.GetBackupStoreDriver`, so we don't need to handle s3 backup type here.

https://github.com/harvester/harvester/blob/b28f8d79f8680f29bd9c5402c71c566837788c7a/pkg/util/backup.go#L26C27-L44

pkg/webhook/resources/schedulevmbackup/validator.go
```diff
@@ -0,0 +1,169 @@
...
+	if newSVMBackup.Spec.MaxFailure >= newSVMBackup.Spec.Retain {
+		return werror.NewInvalidError("max failure should be less than retain", fieldMaxFailure)
+	}
+
+	if _, err := cron.ParseStandard(newSVMBackup.Spec.Cron); err != nil {
+		return werror.NewInvalidError("invalid cron format", fieldCron)
+	}
+
+	//not updated to resume schedule
+	if !oldSVMBackup.Spec.Suspend || newSVMBackup.Spec.Suspend {
```
FrankYang0529, line 186: Could you elaborate more on this condition? Thanks.

pkg/controller/master/schedulevmbackup/common.go
```diff
@@ -0,0 +1,480 @@
...
+	return vmBackup, nil
+}
+
+func resolveSVMBackupRef(h *svmbackupHandler, obj metav1.Object) *harvesterv1.ScheduleVMBackup {
+	var annotations = obj.GetAnnotations()
+	if annotations == nil || annotations[util.AnnotationSVMBackupID] == "" {
+		return nil
+	}
+
+	namespace, name := ref.Parse(annotations[util.AnnotationSVMBackupID])
```
FrankYang0529: We may need to add webhook for this annotation to prevent users from updating it.

pkg/controller/master/schedulevmbackup/svmbackup.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+const (
+	releaseAppHarvesterName = "harvester"
+)
+
+func deleteCronJob(h *svmbackupHandler, svmbackup *harvesterv1.ScheduleVMBackup) error {
+	cronJob, err := getCronJob(h, svmbackup)
+	if err != nil {
+		return err
```
FrankYang0529, line 29: We can check whether cronjob is not found, so we can avoid a ScheduleVMBackup can't be removed.

pkg/controller/master/schedulevmbackup/common.go
WebberHuang1118: If the old `svmbackup`  CR is already in a resume state, the Harvester controller doesn't need to check backup target connectivity.

If the new `svmbackup` CR asks to suspend the schedule, the Harvester controller doesn't need to check backup target connectivity, either.

pkg/controller/master/image/register.go
WebberHuang1118: This variable is used here https://github.com/harvester/harvester/blob/a128dbfcdc2df4329ccd342ed4ae54f44e4cb432/pkg/controller/master/schedulevmbackup/register.go#L68

pkg/webhook/resources/schedulevmbackup/validator.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/webhook/resources/schedulevmbackup/validator.go
WebberHuang1118: If the old svmbackup CR is already in a resume state, the Harvester controller doesn't need to check backup target connectivity.

If the new svmbackup CR asks to suspend the schedule, the Harvester controller doesn't need to check backup target connectivity, either.

Thanks.

pkg/controller/master/schedulevmbackup/common.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/schedulevmbackup/svmbackup.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/image/register.go
FrankYang0529: There are two `cronJobControllerName` variables. One is in image and another is in schedulevmbackup. I use IDE reference check and can't find usage for the one in image package. Could you check it again? Thanks.

WebberHuang1118: Sorry, it's my fault, you're right, thanks for pointing this out. I will update it, thanks.

WebberHuang1118: Updated per suggestion, thanks.

FrankYang0529 (approved):
LGTM. Thanks for the PR.

pkg/controller/master/schedulevmbackup/common.go
```diff
@@ -0,0 +1,465 @@
...
+
+	return h.vmBackupClient.Create(vmBackup)
+}
+
+// Try our best to delete outdated VM Backup and LH snapshots
+// even error occurs during the process, log the error and continue to traverse all resources
+func cleanseVMBackup(h *svmbackupHandler, vmbackup *harvesterv1.VirtualMachineBackup) error {
+
+	// Following commentted code block should be used if LH setting AutoCleanupSnapshotWhenDeleteBackup is off
+	// TODO: remove this code block after confirm the default AutoCleanupSnapshotWhenDeleteBackup value
```
bk201: Let's remove the comment block if it's not needed.

pkg/webhook/resources/schedulevmbackup/validator.go
```diff
@@ -0,0 +1,203 @@
...
+			continue
+		}
+
+		offset, err := util.CalculateCronOffset(svmbackup, s)
+		if err != nil {
+			return err
+		}
+
+		if offset < minCronOffset {
+			return fmt.Errorf("same granularity as schedule %s and offset less than %s", s.Name, minCronOffset.String())
```
bk201, line 118:
```suggestion
			return fmt.Errorf("same granularity as schedule %s and offset less than %s minutes", s.Name, minCronOffset.String())
```

just a suggestion to make the error message more understandable.

deploy/charts/harvester-crd/templates/harvesterhci.io_schedulevmbackups.yaml
```diff
@@ -0,0 +1,196 @@
...
+  name: schedulevmbackups.harvesterhci.io
+spec:
+  group: harvesterhci.io
+  names:
+    kind: ScheduleVMBackup
+    listKind: ScheduleVMBackupList
+    plural: schedulevmbackups
+    shortNames:
+    - svmbackup
+    - svmbackup
```
tserong: I suspect one of the above shortNames is intended to be plural:

```suggestion
    - svmbackups
```

pkg/apis/harvesterhci.io/v1beta1/schedulebackup.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type ScheduleVMBackupSpec struct {
+	// +kubebuilder:validation:Required
+	Cron string `json:"cron"`
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=8
+	// +kubebuilder:validation:Maximum=250
+	// +kubebuilder:validation:Minimum=2
```
tserong, line 60: The HEP has maxium retained backups as 254, but it's 250 here.

pkg/controller/master/schedulevmbackup/common.go
```diff
@@ -0,0 +1,465 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return vmBackup, nil
+}
+
+func currentVMBackups(h *svmbackupHandler, svmbackup *harvesterv1.ScheduleVMBackup) (
+	[]*harvesterv1.VirtualMachineBackup, []*harvesterv1.VirtualMachineBackup,
+	*harvesterv1.VirtualMachineBackup, int, error) {
```
tserong: Because there's several return parameters here with the same or similar types, I'd suggest naming them, to make it clear what they're for when reading the function definition, something like:

```suggestion
func currentVMBackups(h *svmbackupHandler, svmbackup *harvesterv1.ScheduleVMBackup) (
	vmbackups []*harvesterv1.VirtualMachineBackup, errVMBackups []*harvesterv1.VirtualMachineBackup,
	lastVMBackup *harvesterv1.VirtualMachineBackup, failure int, err error) {
```

(If you take this change, it will mean also replacing a bunch of `:=` with just `=` in the function body)

pkg/controller/master/schedulevmbackup/svmbackup.go
```diff
@@ -0,0 +1,126 @@
...
+
+func createCronJob(h *svmbackupHandler, svmbackup *harvesterv1.ScheduleVMBackup) (*batchv1.CronJob, error) {
+	backoffLimit := int32(cronJobBackoffLimit)
+	jobImage, err := utilCatalog.FetchAppChartImage(h.appCache, h.namespace,
+		releaseAppHarvesterName, []string{"generalJob", "image"})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get harvester image (%s): %v", jobImage.ImageName(), err)
+	}
+
+	cronJob := &batchv1.CronJob{
```
tserong, line 46: Just to make sure I understand, the pod started by this cronjob doesn't actually do anything itself (it just sleeps for 10 seconds then terminates), but each time the job runs, the fact that it runs triggers the call to `OnCronjobChanged()`, which in turn is what creates the VM backup.

Can you please add a comment somewhere to explain this? I got a bit confused earlier when I started out looking at the job spec and was thinking "Wait, this thing just sleeps... What creates the backup?" :-)

pkg/controller/master/schedulevmbackup/common.go
```diff
@@ -0,0 +1,465 @@
...
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	scheduleVMBackupKindName = "ScheduleVMBackup"
+	timeFormat               = "20060102.1504"
+
+	reachMaxFailure = "Reach Max Failure"
+
+	proactiveSuspend = "Proactive Scheulde Suspend"
```
tserong: Small typo:

```suggestion
	proactiveSuspend = "Proactive Schedule Suspend"
```

tserong (approved):
LGTM in general, just a few small comments/suggestions/questions.

Also, the HEP doesn't seem to mention the granularity check, so that probably wants to be updated too.

pkg/webhook/resources/schedulevmbackup/validator.go
WebberHuang1118: Hi @bk201 
the error message will be like this `The request is invalid: spec.cron: same granularity as schedule svmbackup-vm1 and offset less than 10m0s`, maybe we don't need the additional word `minutes`?

bk201: no need my bad, I was not aware it's a time type.

pkg/controller/master/schedulevmbackup/common.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/apis/harvesterhci.io/v1beta1/schedulebackup.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/schedulevmbackup/common.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/schedulevmbackup/svmbackup.go
WebberHuang1118: Added the comment, thanks.

pkg/controller/master/schedulevmbackup/common.go
WebberHuang1118: Updated per suggestion, thanks.

bk201 (approved):
LGTM, thanks!

WebberHuang1118:
@mergify v1.4

WebberHuang1118:
@mergify backport v1.4

PR #6159, opened 2024-07-11
feat: add webhook for snapshot space management

FrankYang0529:

**Problem:**
Snapshots use extra storage. Currently, users don't have a way to control snapshots count/size and check how much size snapshot uses. Snapshots may use more storage than users expect.

**Solution:**
~Introduce `harvesterhcio.io/totalSnapshotSize` annotation to VM and Namespace. Users can set a value on it to limit total snapshot usage in a VM or Namespace.~

Introduce `ResourceQuota` to control total namespace snapshot size and VM snapshot size.

**Related Issue:**
https://github.com/harvester/harvester/issues/4478, https://github.com/harvester/harvester/pull/4497

**Test plan:**

1. `namespaceTotalSnapshotSizeUsage` and `vmTotalSnapshotSizeUsage` values format.
    - Deny value which can't be parse by [ParseQuantity](https://github.com/kubernetes/apimachinery/blob/e126c655b8146add9b9c19bf89dab6eb85578814/pkg/api/resource/quantity.go#L276-L277).
    - Deny negative value.
2. Set `30Gi` limit on a VM after it's created. The VM has two volumes. The first volume size is `10Gi` and the second volume size is `20Gi`.
    - Write data to make two volumes full. Create the first VMBackup. It should be allowed. The total snapshot size is `30Gi`.
    - Clear data and write data to make two volumes full again. Create the seoncd VMBackup. It should be allowed, because we haven't set the annotation. The total snapshot size is `60Gi`.
    - Set `30Gi` on the VM. It should be allowed.
    - Clear data and write data to make two volumes full again.  Create the third VMBackup. It should be denied, because the total snapshot size will be `90Gi` if the request is allowed.
    - Remove the first and the second VMBackup. The total snapshot size is `0Gi`.
    - Create a new VMBackup. It should be allowed. The total snapshot size is `30Gi`.
4. Set `10Gi` limit on a Namespace. Create a VM with size `10Gi`.
    - Write data to make the volume full. Create the first VMBackup. It should be allowed. The total snapshot size is `10Gi`.
    - Clear data and write data to make the volume full again. Create the second VMBackup. It should be denied, because the total snapshot size will be `20Gi` if the request is allowed.
    - Remove the first VMBackup. The total snapshot size is `0Gi`.
    - Create a new VMBackup. It should be allowed. The total snapshot size is `10Gi`.
5. Set `10Gi` limit on a Namespace. Create a volume with size `10Gi`.
    - Write data to make the volume full. Create the first VolumeSnapshot. It should be allowed. The total snapshot size is `10Gi`.
    - Clear data and write data to make the volume full again. Create the second VolumeSnapshot. It should be denied, because the total snapshot size will be `20Gi` if the request is allowed.
    - Remove the first VolumeSnapshot. The total snapshot size is `0Gi`.
    - Create a new VolumeSnapshot. It should be allowed. The total snapshot size is `10Gi`.
6. Users with `Harvester ResourceQuotas Read Only` role.
    - Apply `harvester-resourcequotas-admin` and `harvester-resourcequotas-read-only` to a Rancher.
    - Import Harvester to a Rancher.
    - Create a new user with `Harvester ResourceQuotas Read Only` role.
    - Add the user to a project as member.
    - Send requests to update Resource Quotas via API on Namespace / VM. Both requests should be denied.
7. Users with `Harvester ResourceQuotas Admin` role.
    - Apply `harvester-resourcequotas-admin` and `harvester-resourcequotas-read-only` to a Rancher.
    - Import Harvester to a Rancher.
    - Create a new user with `Harvester ResourceQuotas Admin` role.
    - Add the user to a project as readonly user.
    - Send requests to update Resource Quotas via API on Namespace / VM. Both requests should be allowed.


FrankYang0529:
This PR will need to wait for LH next release to include https://github.com/longhorn/longhorn/issues/8365.

pkg/webhook/resources/resourcequota/validator.go
```diff
@@ -0,0 +1,70 @@
...
+	resourceQuota := newObj.(*v1beta1.ResourceQuota)
+	if resourceQuota == nil {
+		return nil
+	}
+
+	// Now, only support "default-resource-quota" in each namespace
+	if resourceQuota.Name != util.DefaultResourceQuotaName {
+		return werror.NewInvalidError(fmt.Sprintf("name must be %s", util.DefaultResourceQuotaName), "name")
+	}
+	return nil
```
ibrokethecloud: apart from name, i think we should also be running `checkSnapshotLimit` here to ensure that the spec is valid

pkg/api/vm/handler.go
```diff
@@ -1367,3 +1390,53 @@ func convertNodeSelectorRequirementToSelector(req corev1.NodeSelectorRequirement
...
+		resourceQuotaCopy.Spec.SnapshotLimit.VMTotalSnapshotSizeQuota = map[string]int64{}
+	}
+	resourceQuotaCopy.Spec.SnapshotLimit.VMTotalSnapshotSizeQuota[name] = totalSnapshotSizeQuotaInt64
+	if resourceQuotaCopy.Spec.SnapshotLimit.VMTotalSnapshotSizeQuota[name] == 0 {
+		delete(resourceQuotaCopy.Spec.SnapshotLimit.VMTotalSnapshotSizeQuota, name)
+	}
+	if !reflect.DeepEqual(resourceQuota, resourceQuotaCopy) {
+		_, err = h.resourceQuotaClient.Update(resourceQuotaCopy)
+		return err
+	}
```
ibrokethecloud: should we have a separate action to deleteVMSnapshotResourceQuota? rather than deleting if the value is set to 0?

pkg/api/vm/formatter.go
```diff
@@ -332,3 +350,26 @@ func canDismissInsufficientResourceQuota(vm *kubevirtv1.VirtualMachine) bool {
 	}
 	return true
 }
+
+func (vf *vmformatter) canUpdateResourceQuota(namespace, user string) bool {
```
ibrokethecloud: should we leverage apiutil.CanUpdateResourceQuota to verify the same? This should make it easier to maintain in the long run

FrankYang0529:
Hi @ibrokethecloud, I addressed all comments. Thanks for your review.

ibrokethecloud (approved):
lgtm. thanks.

pkg/webhook/resources/resourcequota/validator.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil
+	}
+
+	return checkSnapshotLimit(resourceQuota.Spec.SnapshotLimit)
+}
+
+func checkSnapshotLimit(snapshotLimit v1beta1.SnapshotLimit) error {
+	if snapshotLimit.NamespaceTotalSnapshotSizeQuota < 0 {
+		return werror.NewInvalidError("must be greater than or equal to 0", "spec.snapshotLimit.namespaceTotalSnapshotSizeQuota")
+	}
```
WebberHuang1118: Maybe we can use Kubebuilder `// +kubebuilder:validation:Minimum=0` to replace this check

pkg/api/vm/handler.go
```diff
@@ -227,6 +239,19 @@ func (h *vmActionHandler) doAction(rw http.ResponseWriter, r *http.Request) erro
 		}
 	case dismissInsufficientResourceQuota:
 		return h.dismissInsufficientResourceQuota(name, namespace)
+	case updateResourceQuotaAction:
+		if ok, err := apiutil.CanUpdateResourceQuota(h.clientSet, namespace, user.GetName()); err != nil {
+			return apierror.NewAPIError(validation.ServerError, fmt.Sprintf("Failed to check permission: %v", err))
+		} else if !ok {
+			return apierror.NewAPIError(validation.PermissionDenied, "User does not have permission to update resource quota")
+		}
```
WebberHuang1118, line 247: Since `CanUpdateResourceQuota()` is already checked in https://github.com/FrankYang0529/harvester/blob/5b5235db6156776a16f18044c4a5015022bc5200/pkg/api/vm/formatter.go#L57-L65, is it necessary to check again here?

pkg/webhook/resources/volumesnapshot/validator.go
```diff
@@ -0,0 +1,80 @@
...
+			return nil
+		}
+		return werror.NewInternalError(fmt.Sprintf("failed to get resource quota %s/%s, err: %s", newVolumeSnapshot.Namespace, util.DefaultResourceQuotaName, err))
+	}
+
+	if err = webhookutil.CheckTotalSnapshotSizeOnNamespace(v.pvcCache, v.engineCache, newVolumeSnapshot.Namespace, resourceQuota.Spec.SnapshotLimit.NamespaceTotalSnapshotSizeQuota); err != nil {
+		return err
+	}
+	return nil
+}
```
WebberHuang1118: A Question: if take a snapshot of a VM's volume, will it only be limited by the namespace quota?

FrankYang0529: We check VM's PVCs and calculate all snapshot size of these PVCs, so a VolumeSnapshot is also calculated in VM quota.

pkg/api/vm/handler.go
FrankYang0529: Yeah, I see. I think `Formatter` is just to show which actions can be applied to this VM, not a real authentication. If an user has permission to operate VM, they still can do the operation even the action is not showed in the `Formatter`.

The steve server do VM permission check when users call a VM URL. However, `ResourceQuota` is a different object, so default steve server checks of VM permission is not enough. That's why we still have to do a check again here.

pkg/webhook/resources/resourcequota/validator.go
FrankYang0529: Good catch. Updated it. Thanks.

pkg/webhook/resources/volumesnapshot/validator.go
WebberHuang1118: I mean if the user directly using volume -> `Take Snapshot` via GUI, even the volume is the member disk for a certain VM, will it only be limited by the namespace quota? not VM quota?

FrankYang0529: No, it's in both namespace and VM quota.

pkg/webhook/resources/virtualmachinebackup/validator.go
```diff
@@ -177,3 +190,66 @@ func (v *virtualMachineBackupValidator) Delete(_ *types.Request, obj runtime.Obj
...
+			snapshotSize, err := longhornutil.ConvertSize(snapshot.Size)
+			if err != nil {
+				return werror.NewInternalError(fmt.Sprintf("failed to convert snapshot size %s to bytes, err: %s", snapshot.Size, err))
+			}
+			totalSnapshotUsage += snapshotSize
+		}
+	}
+
+	if totalSnapshotSizeQuota < totalSnapshotUsage {
+		return werror.NewInvalidError(fmt.Sprintf("total snapshot size limit %d is less than total snapshot usage %d", totalSnapshotSizeQuota, totalSnapshotUsage), "spec.source.name")
```
a110605: It would nice to show current usage and total snapshot quota in Gi.

Current UI only allows user to input quota in GB.

Screenshot 2024-08-21 at 3 31 12 PM

pkg/webhook/resources/volumesnapshot/validator.go
FrankYang0529: Hi @WebberHuang1118, I fix this. Could you take a look? Thank you.

pkg/webhook/resources/virtualmachinebackup/validator.go
FrankYang0529: Updated it.
Screenshot 2024-08-23 at 12 30 49 PM

pkg/webhook/resources/volumesnapshot/validator.go
WebberHuang1118: Thanks for updating the VM quota limit check for pure VolumeSnpashot (no owner reference to VM)

WebberHuang1118 (approved):
LGTM, thanks the PR,

pkg/api/namespace/formatter.go
```diff
@@ -0,0 +1,166 @@
...
+		if err := json.NewDecoder(req.Body).Decode(&updateResourceQuotaInput); err != nil {
+			return apierror.NewAPIError(validation.InvalidBodyContent, fmt.Sprintf("Failed to decode request body: %v ", err))
+		}
+		return h.updateResourceQuota(namespace, updateResourceQuotaInput)
+	case deleteResourceQuotaAction:
+		if ok, err := apiutil.CanUpdateResourceQuota(h.clientSet, namespace, user.GetName()); err != nil {
+			return apierror.NewAPIError(validation.ServerError, fmt.Sprintf("Failed to check permission: %v", err))
+		} else if !ok {
+			return apierror.NewAPIError(validation.PermissionDenied, "User does not have permission to update resource quota")
+		}
```
bk201, line 101: The code is the same as in the previous case, is there a way to remove the duplication?

pkg/webhook/resources/resourcequota/validator.go
```diff
@@ -0,0 +1,66 @@
...
+	resourceQuota := newObj.(*v1beta1.ResourceQuota)
+	if resourceQuota == nil {
+		return nil
+	}
+
+	return checkSnapshotLimit(resourceQuota.Spec.SnapshotLimit)
+}
+
+func checkSnapshotLimit(snapshotLimit v1beta1.SnapshotLimit) error {
+	for vmName, size := range snapshotLimit.VMTotalSnapshotSizeQuota {
```
bk201, line 60: do we need to check if `NamespaceTotalSnapshotSizeQuota < 0` too?

pkg/webhook/resources/virtualmachinebackup/validator.go
```diff
@@ -177,3 +190,66 @@ func (v *virtualMachineBackupValidator) Delete(_ *types.Request, obj runtime.Obj
...
+
+		pvcNamespace := vm.Namespace
+		pvcName := volume.PersistentVolumeClaim.ClaimName
+
+		pvc, err := v.pvcCache.Get(pvcNamespace, pvcName)
+		if err != nil {
+			return werror.NewInternalError(fmt.Sprintf("failed to get PVC %s/%s, err: %s", pvcNamespace, pvcName, err))
+		}
+
+		engine, err := webhookutil.GetLHEngine(v.engineCache, pvc.Spec.VolumeName)
```
bk201: We should check if a VM volume is a Longhorn volume first. The implementation counts Longhorn volume only and it's possible to attach 3rd party storage volume to a VM. Can you also check if the following code works with Longhorn v2 volumes? Thanks.

Same with the volumesnapshot validator, not 100% sure if the code works on non-Longhorn volumes.

pkg/api/namespace/formatter.go
```diff
@@ -0,0 +1,166 @@
...
+	resourceQuota, err := h.resourceQuotaClient.Get(namespace, util.DefaultResourceQuotaName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	resourceQuotaCopy := resourceQuota.DeepCopy()
+	resourceQuotaCopy.Spec.SnapshotLimit.NamespaceTotalSnapshotSizeQuota = 0
```
bk201, line 160: Wondering if we should remove the VM snapshot quota too.

bk201 (commented):
Thanks @FrankYang0529 the code looks complete, but I have some questions.

One side question is: is there a need to use actions? Can GUI manipulate the ResourceQuota resources via Kubernetes API directly?

pkg/api/namespace/formatter.go
FrankYang0529: This is API for namespace, so we only remove namespace quota here. For VM, we have another API.

pkg/webhook/resources/virtualmachinebackup/validator.go
FrankYang0529: Good catch. I update it to check whether PVC is from LH first. For now, this feature can only work with LH. We may need to have an adapter for different driver.

pkg/webhook/resources/resourcequota/validator.go
FrankYang0529: We have check in kubebuilder.

```go
type SnapshotLimit struct {
	// +kubebuilder:validation:Minimum=0
	NamespaceTotalSnapshotSizeQuota int64            `json:"namespaceTotalSnapshotSizeQuota,omitempty"`
	VMTotalSnapshotSizeQuota        map[string]int64 `json:"vmTotalSnapshotSizeQuota,omitempty"`
}
```

pkg/api/namespace/formatter.go
FrankYang0529: I would prefer to keep it, because these API will update `resourcequotas.harvesterhci.io`. The default authentication for VM/Namespace (by steve server) can't work for `resourcequotas.harvesterhci.io`.

Only changing `resourcequotas.harvesterhci.io` APIs need this. Other changing VM/Namespace APIs don't need it.

FrankYang0529:
Hi @bk201, thanks for the review.

> One side question is: is there a need to use actions? Can GUI manipulate the ResourceQuota resources via Kubernetes API directly?

Yeah, GUI can use Kubernetes API to change ResourceQuota. Also, the authentication still works, because we have `ClusterRoleBinding`. However, I would prefer UI can change ResourceQuota via steve server API. The reason is ResourceQuota stores data as int64, but UI shows `Gi` to users. I would like to let server side do string to int64 transformation. It's easier to debug.

bk201 (approved):
LGTM, thanks!

FrankYang0529:
@mergify backport v1.4

Issue #6155, opened 2024-07-11
[ENHANCEMENT] Enhance wording on the error message when insufficient kubelet image garbage collection when triggering upgrade


albinsun:
We should
1. Enhance error message to let user know it's a estimated value thus avoid confusion.
2. Provide document or hint about how to fix (eg. purge unused images)

Note that we hit 2 cases:
1. `higher than kubelet image garbage collection threshold`
   ![image](https://github.com/harvester/harvester/assets/2773781/52c1c59c-8bb0-496d-a5a3-ec1488b8b331)
1. `df -h /usr/local`
   ![image](https://github.com/harvester/harvester/assets/2773781/3544a06a-69f9-410a-bbd5-cd1b90edb593)

# DEPRECATED, PLEASE IGNORE FOLLOWING
**Describe the bug**
During reproduce harvesterharvester/issues/6041 and trigger `v1.2.2` -> `v1.3.1` upgrade but it 's denied by
```
admission webhook "validator.harvesterhci.io" denied the request: Node "harvester-4qqjh" uses 88.336050% storage space. It's higher than kubelet image garbage collection threshold 85.000000%.
```
![image](https://github.com/harvester/harvester/assets/2773781/edafcc15-4951-4db1-a95d-92383cb8a3b5)

But the node's partitions (also `/var/lib/kubelet` mounted as `/usr/local`) looks ok and does not over `85%` threshhold.
![image](https://github.com/harvester/harvester/assets/2773781/655ae136-243b-4690-ad90-2786d2f2a49b)

CC @TachunLin 

**To Reproduce**
1. Setup `v1.1.3` 4 nodes cluster
4. Join Harvester to Rancher `v2.8.5` (But not create any guest cluster)
5. Create 2 VMs, both with 2 volumes (1 default and 1 customized)
6. Upgrade to `v1.2.1`
7. Create a 3 nodes RKE2 guest cluster
8. Upgrade to `v1.2.2`
9. Upgrade to `v1.3.1` :x: 
   ![image](https://github.com/harvester/harvester/assets/2773781/edafcc15-4951-4db1-a95d-92383cb8a3b5)

**Expected behavior**
1. Can trigger upgrade cuz from user aspect the problem node has no partition over `85%` usage.
2. If the space calculation way is not simply checking `/usr/local` and perhaps we sould document or hint somewhere.

**Support bundle**
[...]

PR #6171, opened 2024-07-15
fix: enhance wording on insufficient space message

FrankYang0529:

**Problem:**

Original error message is not clear.

image

**Solution:**

Screenshot 2024-07-15 at 4 30 09 PM

**Related Issue:**
https://github.com/harvester/harvester/issues/6155

**Test plan:**
### Setup

1. Create a new Harvester cluster. Each node's disk space should be 250G.
2. If every node's `/usr/local` free space is more than 15%, you can use `dd` to write files to make sure one of the node's free disk space is lower than 15%. For example: `dd if=/dev/zero of=/usr/local/test.img bs=1G count=93`.
3. Create a test version.
```yaml
kind: Version
metadata:
  name: master-head
  namespace: harvester-system
spec:
  isoURL: http://192.168.0.181:8000/harvester-master-head-amd64.iso
  minUpgradableVersion: 1.3.1
  releaseDate: "20231210"
  tags:
  - dev
  - test
```

Screenshot 2024-07-15 at 4 30 09 PM

### Case 2: a node with enough disk spaces can upgrade

1. Remove the `dd` file.
2. Click the upgrade again. There is no error message.


albinsun (approved):
LGTM

Yu-Jack (approved):

FrankYang0529:
@mergify backport v1.4

Issue #5768, opened 2024-05-08
[ENHANCEMENT] A better reserved memory value for VMs


bk201:
**Is your enhancement related to a problem? Please describe.**

We reserve 100MB of memory since https://github.com/harvester/harvester/pull/1543, but we still see some VMs OOM. 

**Describe the solution you'd like**

- We can do some benchmarks and fine-tune a good default value.
- We should have a document or KB on adjusting the values based on VM workload.

**Describe alternatives you've considered**

**Additional context**

A customer is hitting the issue with the default value.
- It's a 64GB VM.
- The VM is running RKE2 guest cluster.
- The VM has many hot-plugging volumes. This should be taken into consideration, more volumes might increase memory usage.

bk201:
@w13915984028 Please help check this issue. cc @innobead

w13915984028:
From the investigation till now, there are still some doubts to be digged.

The main question is:
The Cgroup memory limit for a VM is set to: [(1) guest OS usable memory][(2)all kinds of reservation]

When OOM happens, it's the (1) + (2)  > limit, and normally, (1) >>>>> (2), it will mean (1) will use a big amount of memory to trigger the OOM.

tserong:
In case it helps with the investigation, someone on the Rancher Users slack ran into the OOM issue back in March when doing heavy IO on a database in a VM, and provided a support bundle there. The discussion is at https://rancher-users.slack.com/archives/C01GKHKAG0K/p1710208792865519 but note that we're probably getting pretty close to the 90 day message limit on Slack in that conversation.

bk201:
@w13915984028 Do you have any update on this issue?

w13915984028:
The current summary is on https://github.com/w13915984028/harvester-develop-summary/blob/main/oom-related-issue-investigation.md, we will optimize the VM to have options for user:

(1) Set a higher reserved memroy by default, update documents about why to do so and how

(2) Remove the memory limit of a VM, keep the wanted memroy of guest VM. e.g. A VM can have upto 16GB memory, but the cgroup of this VM's backing POD, has no memory uplimit. This is useful for mission-critical VM which should never be OOM killed.

User will decide to use which kind of otpion for each VM.

w13915984028:
@torchiaf Please also help remove the `Show More` from UI to expose the `Reserved Memory` option directly, it is very important to adjust this value accordingly. thanks.

[...]

PR #6177, opened 2024-07-16
[WIP]Set higher default reserved memory for VM

w13915984028:

**Problem:**

Harvester set a reserved memory, which is in such form, e.g.

Users specify 2Gi memory limit for a VM, Harvester sets the `memory.guest` to `1948Mi`, and `100Mi` memory is reserved for various of overhead, if users do not set it explicitly on the Harvester UI.

(1) Click the `Show More`
image

(2) Set the `Reserved Memory`
image

But in our test, the `100Mi` can't fit all scenarios.

```
        machine:
          type: q35
        memory:
          guest: 1948Mi
        resources:
          limits:
            cpu: "1"
            memory: 2Gi
          requests:
            cpu: 62m
            memory: 1365Mi
```

**Solution:**

https://github.com/harvester/harvester/issues/5768#issuecomment-2188232277

The current summary is on https://github.com/w13915984028/harvester-develop-summary/blob/main/oom-related-issue-investigation.md, we will optimize the VM to have options for user:

(1) Set a higher reserved memroy by default, update documents about why to do so and how

(2) Remove the memory limit of a VM, keep the wanted memroy of guest VM. e.g. A VM can have upto 16GB memory, but the cgroup of this VM's backing POD, has no memory uplimit. This is useful for mission-critical VM which should never be OOM killed. (not in this PR)

User will decide to use which kind of otpion for each VM.

The Rancher integration scenario will also benifit from this PR, there is no interface to set the reserved memory from Rancher UI yet, the auto-calculated reserved memory fits it in most cases.

**Related Issue:**
https://github.com/harvester/harvester/issues/5768

**Test plan:**

1. Create VM without setting reserved memory; Harvester computes a dynamic reserved memory based on user input.

```
4Gi limit, 3687Mi for guest

$ kubeclt  get vm vm21 -oyaml

        machine:
          type: q35
        memory:
          guest: 3687Mi
        resources:
          limits:
            cpu: "1"
            memory: 4Gi
          requests:
            cpu: 62m
            memory: 2730Mi

2Gi limit, 1792Mi for guest

        machine:
          type: q35
        memory:
          guest: 1792Mi
        resources:
          limits:
            cpu: "1"
            memory: 2Gi
          requests:
            cpu: 62m
            memory: 1365Mi


10Gi limit, 9Gi for guest

        machine:
          type: q35
        memory:
          guest: 9Gi
        resources:
          limits:
            cpu: "1"
            memory: 10Gi
          requests:
            cpu: 62m
            memory: 6826Mi

12Gi limit, 11Gi for guest

        machine:
          type: q35
        memory:
          guest: 11Gi
        resources:
          limits:
            cpu: "1"
            memory: 12Gi
          requests:
            cpu: 62m
            memory: 8Gi
```

2. Create VM with setting reserved memory

```
$ kubeclt  get vm vm21 -oyaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/reservedMemory: 400Mi


      memory:
        guest: 1648Mi
      resources:
        limits:
          cpu: "1"
          memory: 2Gi
        requests:
          cpu: 62m
          memory: 1365Mi

$ kubeclt get vm vm21 -oyaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/reservedMemory: 100Mi


        machine:
          type: q35
        memory:
          guest: 1948Mi
        resources:
          limits:
            cpu: "1"
            memory: 2Gi
          requests:
            cpu: 62m
            memory: 1365Mi
          
```


w13915984028:
@mergify backport v1.4

FrankYang0529 (approved):
LGTM. Run following test cases and get expected reserved memory:

* 1Gi -> 128Mi
* 2Gi -> 256Mi
* 3Gi -> 307Mi (10%)
* 11Gi -> 1Gi

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -215,6 +215,33 @@ func (m *vmMutator) patchResourceOvercommit(vm *kubevirtv1.VirtualMachine) ([]st
 	return patchOps, nil
 }
 
+// compute a higher default reserved memory for VM to tolerant
+// the potential more memory used by those backend components like qemu, kubevirt
+// refer: https://github.com/harvester/harvester/issues/5768
+func computeReservedMemory(mem, reservedMemory *resource.Quantity) {
```
bk201, line 238: Do you have a benchmark or evaluation result for these computed values?
nit: it would be nice to use constants for these values for readability.

w13915984028: constants update.

Benchmark, no direct data. More from community/jira ticket feedback, 1G reserved memory settled most cases even when VM's memory is about 64G, 512M is also quite good for VM with memory around 4~6G.  For specific senarios, user needs to do some tests and ajust this value accordingly.

There is an ongoing document PR to explain more details.

ibrokethecloud:
change looks good to me, however we need to consider the impact on the upgrade path for larger VM's where due to changes the users may experience unexpected results with 1G reserved for overhead, and subsequently being unavailable to the guest. 

should we look at may be changing how the overhead is reserved? rather than removing from the vm guest spec, we could add it to the vm guest spec, to ensure that the guest gets the memory it asked for?

w13915984028:
@ibrokethecloud Thanks for your reminder.

I am preparing another commit, then:
(1) User can set the option to disable/enable to reserve a fixed value or per formula, the default option will to keep the original 100M; in the furture, the option can even be at namespace/project level
(2) We will hightlight this in Harvester documents, guide user to set a proper option

(3) For the `Reserved Memory`, also considered to `add` instead of `reserve` from guest claim, but that may cause broken change, UI and update both affected. Will re-check a best option.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -231,9 +267,24 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity,
...
+		if reservedMemory.Value() >= mem.Value() {
+			return patchOps, fmt.Errorf("reservedMemory cannot be equal or greater than limits.memory %v %v", reservedMemory.Value(), mem.Value())
+		}
+	} else {
+		// if user does not specify the value, then compute it dynamically
+		computeReservedMemory(mem, &reservedMemory)
 	}
 	guestMemory := *mem
 	guestMemory.Sub(reservedMemory)
+	if guestMemory.Value() < 268435456 {
```
votdev: For better readability
```suggestion
	if guestMemory.Value() < 256*1024*1024 {
```
or
```suggestion
	if -1 == guestMemory.Cmp(resource.ParseQuantity("256Mi")) {
```

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -231,9 +267,24 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity,
...
+		// some test cases use a small value to test, but for production such VM make no sense, add a warning here
+		logrus.Warnf("vm %s/%s guest memory is under the suggested minimum requirement(256Mi), original: %v, reserved: %v, final: %v", vm.Namespace, vm.Name, mem.Value(), reservedMemory.Value(), guestMemory.Value())
+	}
+	if guestMemory.Value() < 10485760 {
```
votdev: For better readability
```suggestion
	if guestMemory.Value() < 10*1024*1024 {
```
or
```suggestion
	if -1 == guestMemory.Cmp(resource.ParseQuantity("10Mi")) {
```

votdev (changes requested):

w13915984028:
Update:

With more investigation, it looks exposing this setting `AdditionalGuestMemoryOverheadRatio` on Harvester will be a better solution:

I will test the feature, when it works well, our solution will be simplified.

```
// GetMemoryOverhead computes the estimation of total
// memory needed for the domain to operate properly.
// This includes the memory needed for the guest and memory
// for Qemu and OS overhead.
// The return value is overhead memory quantity
//
// Note: This is the best estimation we were able to come up with
//
//	and is still not 100% accurate
func GetMemoryOverhead(vmi *v1.VirtualMachineInstance, cpuArch string, additionalOverheadRatio *string) resource.Quantity {
```

https://github.com/kubevirt/kubevirt/blob/021e9b42f79b098fc50e76062e5388e13d31e235/pkg/virt-controller/services/renderresources.go#L303

https://kubevirt.io/api-reference/main/definitions.html
>AdditionalGuestMemoryOverheadRatio can be used to increase the virtualization infrastructure overhead. This is useful, since the calculation of this overhead is not accurate and cannot be entirely known in advance. The ratio that is being set determines by which factor to increase the overhead calculated by Kubevirt. A higher ratio means that the VMs would be less compromised by node pressures, but would mean that fewer VMs could be scheduled to a node. If not set, the default is 1.

Related issue:

Support control of memory guest overhead with InstanceTypes #10057
https://kubevirt.io/api-reference/main/definitions.html

w13915984028:
After adding `additionalGuestMemoryOverheadRatio:"1.2"` to `kubevirt` crd object, a `1Gi 1 CORE VM` will get ~48M more  memory on POD side limitation and finally on cgroup param, but the guest OS still gets ~1Gi memory.

This will be the base of the coming new commit.

```
spec:
  certificateRotateStrategy: {}
  configuration:
    additionalGuestMemoryOverheadRatio: "1.2"  

// roughly, after kubevirt computes a 240Mi value of overhead, it will further increase 240 to 240*1.2


```

w13915984028:
To simplify the review, the PR will be replaced by https://github.com/harvester/harvester/pull/6438 with new settings, test plans and so on. Thanks.

w13915984028:
Replaced by https://github.com/harvester/harvester/pull/6438

Issue #6028, opened 2024-06-18
[ENHANCEMENT] Skip the image preloading phase of Harvester upgrade


starbops:
**Is your enhancement related to a problem? Please describe.**

We've had all the required images published on the Internet, so presumably, all the needed images could be downloaded if the Harvester cluster has Internet connectivity, i.e., non-air-gapped clusters. This gives us the green light to skip the image-preloading phase of the Harvester upgrade to speed up the currently time-consuming node-by-node process.

**Describe the solution you'd like**

Respect a newly introduced annotation for the Upgrade CR: `harvesterhci.io/skip-image-preload`. If such an annotation is set to true, the upgrade controller will skip creating the image-preloading Plan CR and mark all the nodes as `Image Preloaded`.

**Describe alternatives you've considered**

Enabling image preloading parallelism is another way to speed up the Harvester upgrade, especially for the image preloading phase.

**Additional context**

w13915984028:
Validation: **PASS**

(1) Install a cluster from v140-20240918 ISO.
```
$kubectl get settings.harvesterhci.io upgrade-config
NAME             VALUE
upgrade-config   {"imagePreloadOption":{"strategy":{"type":"skip"}}, "restoreVM": true}
```
(2) Upgrade to a local built v140-branch master-head ISO at 20241001:

Upgrade stucks `ImagePullBackOff` as there seems no such image in the remote.

![image](https://github.com/user-attachments/assets/54fb36cd-8f98-4d9b-842a-45e1e2d9b1e9)

```
$kubectl get pods -A | grep upgrade
cattle-system                     system-upgrade-controller-6f86d6d4df-mg8p4               1/1     Running            2 (22m ago)    6d1h
harvester-system                  hvst-upgrade-tvs2p-apply-manifests-897cb                 0/1     ImagePullBackOff   0              5m19s
harvester-system                  virt-launcher-upgrade-repo-hvst-upgrade-tvs2p-hlc6b      2/2     Running            0              7m36s

$kubectl get pods -n harvester-system hvst-upgrade-tvs2p-apply-manifests-897cb -oyaml | grep image
    image: rancher/harvester-upgrade:aa4976fe
    imagePullPolicy: IfNotPresent
  - image: rancher/harvester-upgrade:aa4976fe
    imageID: ""
        message: Back-off pulling image "rancher/harvester-upgrade:aa4976fe"
```

(3) Upgrade to a v140-rc1 ISO

Upgrade is successful
[...]

Issue #3059, opened 2022-10-27
[FEATURE] Optimize upgrade `image preloading`


w13915984028:
**Is your feature request related to a problem? Please describe.**

Currently, the `image preloading` will happen in `NODEs` one by one, for massive-node cluster, it takes long time.

**Describe the solution you'd like**

Preload image to nodes in parallel.
Or allow user to set up an image registry outside of cluster, images will be fetched on demand.
Also consider to allow user to select  `one-by-one`, `multi-in-parallel`, `fetch-on-demand-from-registry` mode.

**Describe alternatives you've considered**

**Additional context**

example of long waiting list:
![image](https://user-images.githubusercontent.com/31133476/198296206-e1cef72d-a597-4133-adfb-2ad8a1f69131.png)

Martin-Weiss:
See also https://github.com/harvester/harvester/issues/3058

w13915984028:
duplicated with https://github.com/harvester/harvester/issues/3058

starbops:
I'd like to reopen this one since we're currently working on upgrade-related enhancement, and this is one of the essential bits. Plus, the scope of this issue is more specific in contrast to #3058.

starbops:
We're adding the ability to skip the image-preload phase during an upgrade (PR #6037 is still under review) by annotating the Upgrade CR with a specific key-value pair. However, as @w13915984028 suggested, we could consolidate these two and provide a new option/setting `image-preload-behavior` to instruct the controller about the image-preload behavior during upgrades (require the controller pods to restart):

- skip (the scope covered by #6028)
- sequential (like the good old days)
- parallel (the scope of this issue)

Or is just providing another annotation, say `harvesterhci.io/parallel-image-preload: true` for the Upgrade CR good enough? cc @bk201 @brandboat

brandboat:
> Or is just providing another annotation, say harvesterhci.io/parallel-image-preload: true for the Upgrade CR good enough? cc @bk201 @brandboat

Sorry for the late response, IMO the image-preload strategy should be a field in spec, add more annotations here is a bit mess... what if `harvesterhci.io/parallel-image-preload: true`, `harvesterhci.io/skip-image-preload: true` both exist, which one take precedence ? I propose we introduce a new field `image-preload-strategy` to upgrade CRD, doing this could avoid the problem mentioned as above.

starbops:
[...]

PR #6181, opened 2024-07-17
Introduce the upgrade-config setting to control the behavior of image preloading during upgrades

starbops:

**Problem:**

In the current Harvester upgrade implementation, image preloading will always happen (provided new container images are introduced) in a one-node-at-a-time fashion. It takes a significant portion of the upgrade time, no matter how powerful the given cluster nodes and network capacity. We shall provide the flexibility for users to customize the image preload strategy.

**Solution:**

Add a new Setting called `upgrade-config`. The upgrade controller will honor the setting value to decide whether to skip preload, preload sequentially, or preload in parallel for container images. This new Setting object will be where we store upgrade-relevant configs/options. For example:

```json
{
  "imagePreloadOption": {
    "strategy": {
      "type": "parallel",
      "concurrency": 2
    }
  }
}
```

This will allow the Harvester cluster to conduct future upgrades with parallel image preloading at a two-node-at-a-time pace. If the `.imagePreloadOption.strategy.concurrency` field is set to 0 (this is also the default value), all nodes will start preloading images simultaneously. Concurrency value is ignored when a strategy type other than "parallel" is chosen.

Available `.imagePreloadOption.strategy.type` field are:

- `skip`
- `sequential`
- `parallel`

**Related Issue:**

#3059 #6028

**Test plan:**

0. For reviewers, please use a custom-built ISO image that contains the PR; for QAs, please use the master ISO image.
1. Prepare a Harvester cluster consists of at lease two nodes
2. Configure the `upgrade-config` setting's `.imagePreloadOption.strategy.type` field to `skip`, `sequential`, or `parallel` respectively
   ```shell
   kubectl patch setting.harvesterhci upgrade-config --type=merge -p '{"value":"{\"imagePreloadOption\":{\"strategy\":{\"type\":\"skip\"}}}"}'
   kubectl patch setting.harvesterhci upgrade-config --type=merge -p '{"value":"{\"imagePreloadOption\":{\"strategy\":{\"type\":\"sequential\"}}}"}'
   kubectl patch setting.harvesterhci upgrade-config --type=merge -p '{"value":"{\"imagePreloadOption\":{\"strategy\":{\"type\":\"parallel\",\"concurrency\":2}}}"}'
   ```
4. Start the upgrade using the same ISO image
5. Observe the upgrade progress, when it enters the image-preload phase:
   - If `skip` is specified, there should be **no** `prepare` pods running in the cluster
   - If `sequential` is specified, there should be exact **one** `parepare` pod running at a time
   - If `parallel` is specified, there should be **two** (or other number depending on how many nodes are there) `prepare` pods running simultaneously
7. The upgrade ends successfully

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -32,8 +32,9 @@ import (
 )
 
 var (
-	upgradeControllerLock sync.Mutex
-	rke2DrainNodes        = true
+	upgradeControllerLock   sync.Mutex
+	rke2DrainNodes          = true
+	imagePreloadConcurrency = defaultImagePreloadConcurrency
 )
 
 const (
@@ -65,6 +66,7 @@ const (
 
 	imageCleanupPlanCompletedAnnotation = "harvesterhci.io/image-cleanup-plan-completed"
 	skipVersionCheckAnnotation          = "harvesterhci.io/skip-version-check"
+	defaultImagePreloadConcurrency      = 1
 )
 
 // upgradeHandler Creates Plan CRDs to trigger upgrades
@@ -74,6 +76,7 @@ type upgradeHandler struct {
 	nodeCache         ctlcorev1.NodeCache
 	jobClient         v1.JobClient
 	jobCache          v1.JobCache
+	settingCache      ctlharvesterv1.SettingCache
 	upgradeClient     ctlharvesterv1.UpgradeClient
 	upgradeCache      ctlharvesterv1.UpgradeCache
 	upgradeController ctlharvesterv1.UpgradeController
@@ -239,6 +242,7 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 		}
 		toUpdate.Status.SingleNode = singleNode
 
+		// Upgrade Repo Info Retrieval
 		backoff := wait.Backoff{
 			Steps:    30,
 			Duration: 10 * time.Second,
@@ -264,6 +268,7 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			return h.upgradeClient.Update(toUpdate)
 		}
 
+		// Upgrade Eligibility Check
 		isEligible, reason := upgradeEligibilityCheck(upgrade, repoInfo)
 
 		if !isEligible {
@@ -271,8 +276,34 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			return h.upgradeClient.Update(toUpdate)
 		}
 
-		logrus.Debug("Start preparing nodes for upgrade")
-		if _, err := h.planClient.Create(preparePlan(upgrade)); err != nil && !apierrors.IsAlreadyExists(err) {
+		// Node Preparation (Image Preloading)
+		imagePreloadStrategy := settings.ImagePreloadStrategy.Get()
+		logrus.WithFields(logrus.Fields{
+			"upgrade_namespace":      upgrade.Namespace,
+			"upgrade_name":           upgrade.Name,
+			"image_preload_strategy": imagePreloadStrategy,
+		}).Info("start preparing nodes for upgrade")
+
+		nodes, err := h.nodeCache.List(labels.Everything())
+		if err != nil {
+			return upgrade, err
+		}
+
+		switch imagePreloadStrategy {
+		case settings.SkipImagePreload:
+			for _, node := range nodes {
+				setNodeUpgradeStatus(toUpdate, node.Name, nodeStateImagesPreloaded, "", "")
+			}
+
+			toUpdate.Labels[upgradeStateLabel] = StatePreparingNodes
+			toUpdate.Status.RepoInfo = repoInfoStr
+			setNodesPreparedCondition(toUpdate, corev1.ConditionTrue, "", "")
+			return h.upgradeClient.Update(toUpdate)
+		case settings.ParallelImagePreload:
+			imagePreloadConcurrency = len(nodes)
+		}
+
+		if _, err := h.planClient.Create(preparePlan(upgrade, imagePreloadConcurrency)); err != nil && !apierrors.IsAlreadyExists(err) {
 			setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, err.Error(), "")
 			return h.upgradeClient.Update(toUpdate)
 		}
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -32,9 +32,8 @@ import (
 )
 
 var (
-	upgradeControllerLock   sync.Mutex
-	rke2DrainNodes          = true
-	imagePreloadConcurrency = defaultImagePreloadConcurrency
+	upgradeControllerLock sync.Mutex
+	rke2DrainNodes        = true
 )
 
 const (
@@ -76,7 +75,6 @@ type upgradeHandler struct {
 	nodeCache         ctlcorev1.NodeCache
 	jobClient         v1.JobClient
 	jobCache          v1.JobCache
-	settingCache      ctlharvesterv1.SettingCache
 	upgradeClient     ctlharvesterv1.UpgradeClient
 	upgradeCache      ctlharvesterv1.UpgradeCache
 	upgradeController ctlharvesterv1.UpgradeController
@@ -276,42 +274,7 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			return h.upgradeClient.Update(toUpdate)
 		}
 
-		// Node Preparation (Image Preloading)
-		imagePreloadStrategy := settings.ImagePreloadStrategy.Get()
-		logrus.WithFields(logrus.Fields{
-			"upgrade_namespace":      upgrade.Namespace,
-			"upgrade_name":           upgrade.Name,
-			"image_preload_strategy": imagePreloadStrategy,
-		}).Info("start preparing nodes for upgrade")
-
-		nodes, err := h.nodeCache.List(labels.Everything())
-		if err != nil {
-			return upgrade, err
-		}
-
-		switch imagePreloadStrategy {
-		case settings.SkipImagePreload:
-			for _, node := range nodes {
-				setNodeUpgradeStatus(toUpdate, node.Name, nodeStateImagesPreloaded, "", "")
-			}
-
-			toUpdate.Labels[upgradeStateLabel] = StatePreparingNodes
-			toUpdate.Status.RepoInfo = repoInfoStr
-			setNodesPreparedCondition(toUpdate, corev1.ConditionTrue, "", "")
-			return h.upgradeClient.Update(toUpdate)
-		case settings.ParallelImagePreload:
-			imagePreloadConcurrency = len(nodes)
-		}
-
-		if _, err := h.planClient.Create(preparePlan(upgrade, imagePreloadConcurrency)); err != nil && !apierrors.IsAlreadyExists(err) {
-			setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, err.Error(), "")
-			return h.upgradeClient.Update(toUpdate)
-		}
-
-		toUpdate.Labels[upgradeStateLabel] = StatePreparingNodes
-		toUpdate.Status.RepoInfo = repoInfoStr
-		harvesterv1.NodesPrepared.CreateUnknownIfNotExists(toUpdate)
-		return h.upgradeClient.Update(toUpdate)
+		return h.prepareNodesForUpgrade(toUpdate, repoInfoStr)
 	}
 
 	if harvesterv1.NodesPrepared.IsTrue(upgrade) && harvesterv1.SystemServicesUpgraded.GetStatus(upgrade) == "" {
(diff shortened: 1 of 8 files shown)


pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -264,15 +268,42 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+			for _, node := range nodes {
+				setNodeUpgradeStatus(toUpdate, node.Name, nodeStateImagesPreloaded, "", "")
+			}
+
+			toUpdate.Labels[upgradeStateLabel] = StatePreparingNodes
+			toUpdate.Status.RepoInfo = repoInfoStr
+			setNodesPreparedCondition(toUpdate, corev1.ConditionTrue, "", "")
+			return h.upgradeClient.Update(toUpdate)
+		case settings.ParallelImagePreload:
+			imagePreloadConcurrency = len(nodes)
```
w13915984028: Recall @bk201 ever mentioned, in parallel mode, the most challenge is that those nodes will get the ISO from the repo at the same time, that will be a pressure to the underlayer LH, and the CPU & network to the node where repo is running.

Did you have a test on multi-node cluster? Any plan to set a number like max[3, node-count] ?

The other solution is to set a max parallel accessing on repo nginx.

w13915984028 (commented):
A question about the parallel capacity.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: No, I haven't tried that in an environment with a considerable number of container images that need to be downloaded since the script simply skip preloading images in same-version upgrades.

We can enforce rate limiting on either side, with `curl --rate-limit` on the clients or once and for all, with Nginx's `limit_rate` directive on the server.

Ideally, clients will compete for their max bandwidth, so from the server's POV, the total uplink bandwidth is almost the same if no rate limitation is applied. The real problem, I guess, is when the client on the same node as the server starts to download and import images. Heavy read and write requests will happen simultaneously on that node, potentially causing instability.

With that said, the `parallel` option is not recommended in virtual environments. We'll have that noted in the setting's description and documents.

brandboat: I'm fine with the current approach, but if that is a problem when we are testing in the local environment,  perhaps we should add another setting to configure the number of parallel job.

brandboat (approved):
Thanks for the patch, LGTM !

pkg/settings/settings.go
```diff
@@ -57,6 +57,7 @@ var (
 	HarvesterCSICCMVersion = NewSetting(HarvesterCSICCMSettingName, `{"harvester-cloud-provider":">=0.0.1 <0.3.0","harvester-csi-provider":">=0.0.1 <0.3.0"}`)
 	NTPServers             = NewSetting(NTPServersSettingName, "")
 	WhiteListedSettings    = []string{"server-version", "default-storage-class", "harvester-csi-ccm-versions", "default-vm-termination-grace-period-seconds"}
+	ImagePreloadStrategy   = NewSetting(ImagePreloadStrategySettingName, SequentialImagePreload) // options are skip, sequential, and parallel
```
bk201: We can extend upgrade CRD to control the behavior of individual upgrades. But a setting is also OK since there will be only a running upgrade. Can you refactor the setting name to upgrade-config or something and make the value a json structure, it's easier to extend upgrade-related configs without adding more settings.

pkg/controller/master/upgrade/upgrade_controller.go
bk201: > the most challenge is that those nodes will get the ISO from the repo at the same time

It was about the image preloading behavior on each node (generating a large amount of I/Os). But the issue only happened in a virtualized environment (all Harvester nodes are VMs). Now we skip unpacking, so likely the preloading will be smooth on bare-metal machines.

bk201 (commented):
Thanks! Just leave some comments.

bk201:
I've discussed with @starbops about where we should set the default policy to `parallel` by default. But considering doing this might bring high traffic to the upgrade repo VM and pod network, we decided to keep the value to `serial` for now.

ibrokethecloud (approved):
lgtm. Do we want to bump up the upgrade repo vm specs from 1cpu/1G when parallel loading is requested?

starbops:
Dear reviewers, recent commits were added to replace the initial `image-preload-strategy` setting with a more general upgrade-relevant one called `upgrade-config` to ensure future-proofing. Also, specifying the concurrency with the parallel strategy is supported.

---

> Do we want to bump up the upgrade repo vm specs from 1cpu/1G when parallel loading is requested?

Though I agree we can bump the upgrade VM spec slightly for the parallel image preloading case, I'm unsure what the reasonable bump is. Is increasing the CPU/memory to a fixed value larger than the original 1cpu/1G enough? Or do we want it to be smart enough to increase the spec according to the concurrency set? But we must cap the resource bump to prevent any cluster node from being unable to accommodate the upgrade VM. This makes me start to think of horizontally scale the upgrade VM...

Do you have any ideas? cc @ibrokethecloud

ibrokethecloud:
@starbops I had a chat with @bk201 and we could may be leave it as is for now. The image download script has sufficient number of retries so we should be ok for most medium sized clusters.

pkg/settings/settings.go
```diff
@@ -57,6 +57,7 @@ var (
 	HarvesterCSICCMVersion = NewSetting(HarvesterCSICCMSettingName, `{"harvester-cloud-provider":">=0.0.1 <0.3.0","harvester-csi-provider":">=0.0.1 <0.3.0"}`)
 	NTPServers             = NewSetting(NTPServersSettingName, "")
 	WhiteListedSettings    = []string{"server-version", "default-storage-class", "harvester-csi-ccm-versions", "default-vm-termination-grace-period-seconds"}
+	UpgradeConfigSet       = NewSetting(UpgradeConfigSettingName, InitUpgradeConfig())
```
w13915984028: Suggest to use a formatted string to set the default value, to keep readability and formatting with other settings.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -264,15 +267,51 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+			"upgrade_name":      upgrade.Name,
+			"upgrade_config":    upgradeConfig,
+		}).Info("start preparing nodes for upgrade")
+
+		nodes, err := h.nodeCache.List(labels.Everything())
+		if err != nil {
+			return upgrade, err
+		}
+
+		imagePreloadConcurrency := defaultImagePreloadConcurrency
```
w13915984028: The OnChange is pretty long, please consider to remove this block to a function.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -264,15 +267,51 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+		case settings.SkipType:
+			for _, node := range nodes {
+				setNodeUpgradeStatus(toUpdate, node.Name, nodeStateImagesPreloaded, "", "")
+			}
+
+			toUpdate.Labels[upgradeStateLabel] = StatePreparingNodes
+			toUpdate.Status.RepoInfo = repoInfoStr
+			setNodesPreparedCondition(toUpdate, corev1.ConditionTrue, "", "")
+			return h.upgradeClient.Update(toUpdate)
+		case settings.ParallelType:
```
w13915984028: Please add a note to the document to recommend/suggest a possible value, and/or tell the potential risk.

w13915984028 (approved):
LGTM, thanks.

With a few suggestions.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: I'll create a doc PR for this. Thanks!

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Updated, PTAL.

pkg/settings/settings.go
starbops: Are you suggesting using the following?

```
UpgradeConfigSet = NewSetting(UpgradeConfigSettingName, `{"imagePreloadOption":{"strategy":{"type":"sequential"}}}`)
```

But if the upgrade config set becomes more extensive and complicated in the future, the string literal can be ugly and hard to maintain. This file seems to have two styles. I feel the current one is more flexible.

Thank you.

w13915984028: (1) At the moment, it is not so complex; when future it becomes really complex, we can discuss again when new PRs coming.

(2) InitUpgradeConfig can be further enhanced with params, and in the test code many hard-coded strings like 
>Value: `{"imagePreloadOption":{"strategy":{"type":"skip"}}}`,

 are good to be replaced with this handy function

pkg/settings/settings.go
```diff
@@ -324,3 +326,63 @@ func GetCSIDriverInfo(provisioner string) (*CSIDriverInfo, error) {
...
+	// Preloading multiple nodes starts at the same time
+	ParallelType StrategyType = "parallel"
+)
+
+type PreloadStrategy struct {
+	Type StrategyType `json:"type,omitempty"`
+
+	// Concurrency only takes effect when ParallelType is specified. Default to
+	// 0, which means "full scale." Any value larger than the number of the cluster
+	// nodes or smaller than 0 will be treated as 0.
```
brandboat: We have validator, why not just throw error if value < 0 ?

brandboat: Looks like you already make concurrently < 0 invalid in validator, we should modify the comment here.

starbops: Good point. Let me align the implementation between the validator and the controller (also the comment).

pkg/settings/settings.go
starbops: I've removed the InitUpgradeConfig function and assign the default value using string literal. Thank you.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Doc PR: harvester/docs#626

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -628,6 +623,55 @@ func upgradeEligibilityCheck(upgrade *harvesterv1.Upgrade, repoInfo *RepoInfo) (
 	return true, ""
 }
 
+func (h *upgradeHandler) prepareNodesForUpgrade(upgrade *harvesterv1.Upgrade, repoInfoStr string) (*harvesterv1.Upgrade, error) {
+	upgradeConfig, err := settings.DecodeConfig[settings.UpgradeConfig](settings.UpgradeConfigSet.Get())
+	if err != nil {
+		return upgrade, err
+	}
+	logrus.WithFields(logrus.Fields{
+		"upgrade_namespace": upgrade.Namespace,
```
bk201: Nit: Although this doesn't look wrong, should we use 'namespace' 'name' 'kind' as defined in https://github.com/harvester/harvester/blob/master/enhancements/20240327-structured-logging.md?

bk201 (approved):

pkg/controller/master/upgrade/upgrade_controller.go
starbops: Thanks for the heads-up. We should follow what we proposed :)

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -628,6 +623,55 @@ func upgradeEligibilityCheck(upgrade *harvesterv1.Upgrade, repoInfo *RepoInfo) (
...
+		"upgrade_config": upgradeConfig,
+	}).Info("start preparing nodes for upgrade")
+
+	nodes, err := h.nodeCache.List(labels.Everything())
+	if err != nil {
+		return upgrade, err
+	}
+
+	imagePreloadConcurrency := upgradeConfig.PreloadOption.Strategy.Concurrency
+	switch upgradeConfig.PreloadOption.Strategy.Type {
```
brandboat, line 642: Correct me if I'm wrong, looks like we didn't handle the `sequential` Strategy Type here, and the imagePreloadConcurrency is 0 (default concurrency value).  And that makes `sequential` the same behavior as `Parallel` (preload to all node concurrently)

starbops: Good catch. I missed that part during refactoring, meaning it doesn't have unit tests covered. I'll fix that.

brandboat (approved):
LGTM, thanks

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -628,6 +622,60 @@ func upgradeEligibilityCheck(upgrade *harvesterv1.Upgrade, repoInfo *RepoInfo) (
 	return true, ""
 }
 
+func (h *upgradeHandler) prepareNodesForUpgrade(upgrade *harvesterv1.Upgrade, repoInfoStr string) (*harvesterv1.Upgrade, error) {
+	upgradeConfig, err := settings.DecodeConfig[settings.UpgradeConfig](settings.UpgradeConfigSet.Get())
```
brandboat, line 626: Correct me if I'm wrong, but we need to get upgradeConfig through settingsCache, like https://github.com/harvester/harvester/blob/ccc71be7225ce76697da2bbe8aeb14507e1dbd5b/pkg/api/vm/handler.go#L606, otherwise this line always get default value.

brandboat: Please ignore this noise, the Provider inside settings already include SettingsCache, which means the Get method already done what I mentioned in the above code.

bk201:
@mergifyio backport v1.4

Issue #6011, opened 2024-06-11
[BUG] `kubeconfig-default-token-ttl-minutes` value of `99999999999999` breaks ability for embedded Rancher (non-external) Harvester Kubeconfig to download


irishgordo:
**Describe the bug**
Downloading of a Harvester Kubeconfig from embedded Rancher will fail when setting the `kubeconfig-default-token-ttl-minutes` to `99999999999999`.
Either Through the UI or via `kubectl edit` on the setting.

**To Reproduce**
1. Either through the UI or the kubectl edit - add `value` to the setting changing it from a default of `0` to `99999999999999`
1. Attempt to download a kubeconfig from Support -> Download Kubeconfig
1. The following error will be displayed:
```
failed to get default token TTL: failed to parse setting 'kubeconfig-default-token-ttl-minutes': error parsing token ttl: time: invalid duration "99999999999999m" 
```

**Workaround**
- Don't set that value to be that high, or else it breaks

**Expected behavior**
Perhaps an upper-bound limit, seems that the TTL breaks past a certain point.

**Environment**
 - Harvester ISO version: v1.3.1-rc2 & on feature-branch: https://github.com/harvester/dashboard/pull/1040
 - Underlying Infrastructure: qemu/kvm

**Additional context**
![Screenshot from 2024-06-10 14-58-32](https://github.com/harvester/harvester/assets/5370752/2a20de66-f22d-4413-897a-8425db538c20)
![Screenshot from 2024-06-10 14-57-28](https://github.com/harvester/harvester/assets/5370752/aad97dd9-1ddd-4ed0-81b6-20d3a274023a)

```
Name:         kubeconfig-default-token-ttl-minutes
Namespace:    
Labels:       <none>
Annotations:  harvesterhci.io/hash: 0e867c00b4100cac8fcc3c7570b3c659ff62d870b9acd6aedcf93fd3
API Version:  harvesterhci.io/v1beta1
Default:      0
Kind:         Setting
Metadata:
  Creation Timestamp:  2024-06-10T19:10:08Z
  Generation:          14
  Managed Fields:
    API Version:  harvesterhci.io/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:default:
    Manager:      kubectl-edit
    Operation:    Update
    Time:         2024-06-10T20:04:32Z
    API Version:  harvesterhci.io/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
[...]

PR #6182, opened 2024-07-17
limit kubeconfig-default-token-ttl-minutes value to 100 years

ibrokethecloud:

**Problem:**

harvester setting `kubeconfig-default-token-ttl-limit` is represented as a string, and there is no defined upper limit on the value passed. The string is eventually parsed as time.Duration, which has an upper limit of about 290 years.
As a result a value like `99999999999999` causes the duration parsing in rancher to break.

**Solution:**

The PR adds a simple check in the setting validator to limit max setting to 100 years or `52560000` minutes.

**Related Issue:**
https://github.com/harvester/harvester/issues/6011

**Test plan:**

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -59,6 +59,7 @@ const (
 	labelAppNameValueAlertManager     = "alertmanager"
 	labelAppNameValueGrafana          = "grafana"
 	labelAppNameValueImportController = "harvester-vm-import-controller"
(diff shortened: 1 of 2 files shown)


Vicente-Cheng (approved):
Thanks, that's enough for checking.

tserong (approved):

ibrokethecloud:
@mergifyio backport v1.3

ibrokethecloud:
@mergifyio backport v1.4

PR #6190, opened 2024-07-18
Support CPU Pinning

brandboat:
~~Blocked by https://github.com/harvester/harvester-installer/pull/756~~

**Problem:**
Implement CPU Pinning feature

**Solution:**
Introduce two endpoints:
- enable cpu manager: `/v1/harvester/nodes/{node_name}?action=enableCPUManager`, which turn cpu manager policy into `static` on the node. Then user could create vm with cpu pinning and deploy those vm to those nodes enable cpu manager.
- disable cpu manager: `/v1/harvester/nodes/{node_name}?action=disableCPUManager`, which turn cpu manager policy into `none`, note that if there is any vm with cpu pinning on the node that needed to disable cpu manager, an error raise `there should not be any running VMs with CPU pinning when disabling the CPU manager`.

for more details, please check the HEP: https://github.com/harvester/harvester/pull/5803

**Related Issue:**
related to https://github.com/harvester/harvester/issues/2305,

**Test plan:**
prerequisite: 
~~- build iso, note that you need to apply this patch
[0001-custom-installer.patch](https://github.com/user-attachments/files/16435446/0001-custom-installer.patch) after checkout to the branch which belongs to this PR. The patch change the harvester-installer version as I have to add some config files through harvester-installer while https://github.com/harvester/harvester-installer/pull/756 not merged yet.~~
- prepare 3 nodes harvester cluster
- create api key: click dashboard top right profile picture > Account & API Keys > Create API Key
- Use the following curl cmd to enable/disable cpu manager:
  a. enable cpu manager
  ```shell
      curl 'https://192.168.100.131/v1/harvester/nodes/{node_name}?action=enableCPUManager' \
      -X 'POST' \
      -H 'Authorization: Bearer <API_KEY>' \
      --insecure
  ```
  b. disable cpu manager
  ```shell
      curl 'https://192.168.100.131/v1/harvester/nodes/{node_name}?action=disableCPUManager' \
      -X 'POST' \
      -H 'Authorization: Bearer <API_KEY>' \
      --insecure
  ```

### Test enable/disable cpu pinning
1. enable cpu manager on harvester-node-0
2. Open k8s and there should have a job name start with `harvester-node-0-update-cpu-manager-` wait until the job done.
3. create vm named `test` with cpu pinning: go to dashboard and click `Virtual Machines` tab, click `Create` fill the Name, CPU, Memory, Volumes>Image field, and then click `Edit as YAML`. Add `dedicatedCpuPlacement: true` to `.spec.template.spec.domain.cpu`. And then click `Create`.
image

4. make sure the cpu pinning works by using the following cmd
  ```
  kubectl exec {virt-launcher-pod-id} -- virsh dumpxml default_test | awk "/<cputune>/,/<\/cputune>/"
  ```
  the cmd should output something like belows, depends on the cpu size, there should have the corresponding `vcpupin` xml fields.
  ```
    <cputune>
      <vcpupin vcpu='0' cpuset='1'/>
      ...
    </cputune>
  ```
5. disable cpu manager in harvester-node-0, and verify user can't disable cpu manager when there is a cpu pinned vm in harvester-node-0.
An error raise: `admission webhook "validator.harvesterhci.io" denied the request: there should not be any running VMs with CPU pinning when disabling the CPU manager`
6. stop the vm `test`
7. disable cpu manager in harvester-node-0

###  Verify only one master node could enable/disable cpu manager
1. enable cpu manager on harvester-node-0, no need to wait until cpu manager job done, go to step 2.
2. enable cpu manager on harvester-node-1
In step 2, an error raised: `admission webhook "validator.harvesterhci.io" denied the request: there is another master node updating the CPU manager policy`. Enabling or disabling the CPU manager requires restarting the kubelet, which in turn necessitates restarting the rke2-server. Since we are using a three-node setup where all nodes have the master role, restarting all rke2-servers simultaneously could lead to the entire cluster going offline. Therefore, the CPU manager update process allows only one master node to update at a time.

### Verify there should be only one job enable/disable cpu manager per node
1. enable cpu manager on harvester-node-0, no need to wait until cpu manager job done, go to step 2.
2. enable or disable cpu manager on harvester-node-0
An error raise: `admission webhook "validator.harvesterhci.io" denied the request: there is other ongoing job updating cpu manager policy`

### Verify we can't enable cpu manager when it is enabled, and we can't disable cpu manager when it's disabled 
1. enable cpu manager on harvester-node-0, wait until cpu manager job done, go to step 2.
2. enable cpu manager on harvester-node-0, an error raises
`admission webhook "validator.harvesterhci.io" denied the request: same cpu manager policy`

### Test reboot
1. Create vm with cpu pinning
2. Reboot harvester node
3. After reboot, check rke2-server or rke2-agent  (depends on the node role) start up successfully and vm is running.

### Test upgrade path
Loop through all tests mentioned above under these 2 scenarios:
1. Test single node upgrade path: Prepare a one node harvester and upgrade to the version that include this feature
2. Test multi node upgrade path: Prepare multi nodes harvester cluster and upgrade to the version that include this feature.
```yaml
kind: Version
metadata:
  name: master
  namespace: harvester-system
spec:
---
kind: Upgrade
metadata:
  annotations:
  generateName: hvst-upgrade-
  namespace: harvester-system
spec:
  version: "master"
  logEnabled: false
```

### DIFF
--- a/deploy/charts/harvester/templates/configmap.yaml
+++ b/deploy/charts/harvester/templates/configmap.yaml
@@ -145,3 +145,94 @@ data:
       sleep 2
     done
     `}}
+  cpu-manager.sh: |-
+    function cleanup_cpu_manager_state() {
+      if [ -f "$CPU_MANAGER_STATE_FILE" ]; then
+        mv "$CPU_MANAGER_STATE_FILE" "${CPU_MANAGER_STATE_FILE}.old"
+        echo "File $CPU_MANAGER_STATE_FILE has been renamed to ${CPU_MANAGER_STATE_FILE}.old"
+      else
+        echo "File $CPU_MANAGER_STATE_FILE does not exist."
+      fi
+    }
+
+    function manage_service() {
+      local service=$1
+
+      echo "Stopping ${service}."
+      if ! chroot $HOST_DIR systemctl stop $service; then
+        echo "Error: failed to stop ${service}."
+        exit 1
+      fi
+      echo "Stopped ${service}."
+
+      cleanup_cpu_manager_state
+
+      echo "Starting ${service}."
+      if ! chroot $HOST_DIR systemctl start $service; then
+        echo "Error: failed to start ${service}."
+        exit 1
+      fi
+      echo "Started ${service}."
+    }
+
+    function wait_for_label() {
+      local node_name=$1
+      local expect_label_value=$2
+      local interval=5
+      local elapsed=0
+
+      while [ $elapsed -lt $WAIT_LABEL_TIMEOUT ]; do
+        if ! labels=$($KUBECTL get node "$node_name" --show-labels); then
+          echo "Error: failed to get labels in $node_name"
+          exit 1
+        fi
+        if echo "$labels" | grep -q "cpumanager=$expect_label_value"; then
+          echo "End update cpu-manager-policy"
+          exit 0
+        fi
+        echo "Value in label cpumanager is not $expect_label_value, wait ${interval}s..."
+        sleep $interval
+        elapsed=$((elapsed + interval))
+      done
+      echo "Error: timeout, elapsed ${elapsed}s"
+      exit 1
+    }
+
+    echo "Start update cpu-manager-policy option..."
+    KUBECTL="$HOST_DIR/$(readlink $HOST_DIR/var/lib/rancher/rke2/bin)/kubectl"
+    CPU_MANAGER_CONFIG_FILE="$HOST_DIR/etc/rancher/rke2/config.yaml.d/99-z01-harvester-cpu-manager.yaml"
+    STATIC_POLICY="static"
+    NONE_POLICY="none"
+    CPU_MANAGER_STATE_FILE="$HOST_DIR/var/lib/kubelet/cpu_manager_state"
+    NODE_NAME="$1"
+    NODE_POLICY="$2"
+    EXIT_CODE=0
+
+    if [ "$NODE_POLICY" != "$STATIC_POLICY" ] && [ "$NODE_POLICY" != "$NONE_POLICY" ]; then
+      echo "Error: invalid cpu-manager-policy $NODE_POLICY"
+      exit 1
+    fi
+
+    if ! $KUBECTL get node "$NODE_NAME" --show-labels | grep -q "cpumanager="; then
+      echo "Error: There is no label cpumanager in node $NODE_NAME."
+      exit 1
+    fi
+
+    printf 'kubelet-arg+:\n- "cpu-manager-policy=%s"' "$NODE_POLICY" > $CPU_MANAGER_CONFIG_FILE
+
+    if chroot $HOST_DIR systemctl is-active --quiet rke2-server; then
+      manage_service "rke2-server"
+    elif chroot $HOST_DIR systemctl is-active --quiet rke2-agent; then
+      manage_service "rke2-agent"
+    else
+      echo "Error: Neither rke2-server nor rke2-agent are running."
+      exit 1
+    fi
+
+    if [ "$NODE_POLICY" = "$STATIC_POLICY" ]; then
+      EXPECT_LABEL_VALUE="true"
+    else
+      EXPECT_LABEL_VALUE="false"
+    fi
+
+    wait_for_label "$NODE_NAME" "$EXPECT_LABEL_VALUE"
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -5,6 +5,7 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 source $SCRIPT_DIR/lib.sh
 UPGRADE_TMP_DIR=$HOST_DIR/usr/local/upgrade_tmp
 STATE_DIR=$HOST_DIR/run/initramfs/cos-state
+MAX_PODS=200
 
 cleanup_incomplete_state_file() {
   STATE_FILE=${STATE_DIR}/state.yaml
@@ -490,8 +491,78 @@ EOF
 set_max_pods() {
 cat > /host/etc/rancher/rke2/config.yaml.d/99-max-pods.yaml <<EOF
 kubelet-arg:
-- "max-pods=200"
+- "max-pods=$MAX_PODS"
 EOF
+}
+
+# system-reserved/kube-reserved is required if we want to set cpu-manager-policy to static
+set_reserved_resource() {
+  local cores=$(chroot $HOST_DIR nproc)
+  local reserverdCPU=$(calculateCPUReservedInMilliCPU $cores $MAX_PODS)
+  # make system:kube cpu reservation ration 2:3
+  local systemReservedCPU=$((reservedCPU * 2 * 2 / 5))
+  local kubeReservedCPU=$((reservedCPU * 2 * 3 / 5))
+  local systemReserverConfig="$HOST_DIR/etc/rancher/rke2/config.yaml.d/99-z00-harvester-reserved-resources.yaml"
+  if [ ! -e $systemReserverConfig ]; then
+    cat > $systemReserverConfig << EOF
+kubelet-arg+:
+- "system-reserved=cpu=${systemReservedCPU}m"
+- "kube-reserved=cpu=${kubeReservedCPU}m"
+EOF
+  fi
+}
+
+# Delete the cpu_manager_state file during the initramfs stage. During a reboot, this state file is always reverted
+# because it was originally created during the system installation, becoming part of the root filesystem. As a result,
+# the policy in cpu_manager_state file is "none" (default policy) after reboot. If we've already set the cpu-manager-policy
+# to "static" before reboot, this mismatch can prevent kubelet from starting, and make the entire node unavailable.
+set_oem_cleanup_kubelet() {
+  cat > $HOST_DIR/oem/91_cleanup_kubelet.yaml << EOF
+name: Cleanup Kubelet
+stages:
+    initramfs:
+        - commands:
+            - rm -f /var/lib/kubelet/cpu_manager_state
+EOF
+}
+
+calculateCPUReservedInMilliCPU() {
+  local cores=$1
+  local maxPods=$2
+
+  # This shouldn't happen
+  if (( $cores <= 0 || $maxPods <= 0 )); then
+    echo 0
+    return
+  fi
+
+  local reserved=0
+
+  # 6% of the first core (60 milliCPU)
+  reserved=$(( $reserved + 6 * 1000 / 100 ))
+
+  # 1% of the next core (up to 2 cores) (10 milliCPU)
+  if (( cores > 1 )); then
+    reserved=$(( $reserved + 1 * 1000 / 100 ))
+  fi
+
+  # 0.5% of the next 2 cores (up to 4 cores) (5 milliCPU)
+  if (( cores > 2 )); then
+    reserved=$(( $reserved + 2 * 500 / 100 ))
+  fi
+
+  # 0.25% of any cores above 4 cores (2.5 milliCPU per core)
+  if (( cores > 4 )); then
+    reserved=$(( $reserved + (cores - 4) * 250 / 100 ))
+  fi
+
+  # If the maximum number of Pods per node is beyond the default of 110,
+  # reserve an extra 400 mCPU in addition to the preceding reservations.
+  if (( maxPods > 110 )); then
+    reserved=$(( $reserved + 400 ))
+  fi
+
+  echo $reserved
 }
 
 upgrade_os() {
@@ -576,8 +647,10 @@ command_post_drain() {
   wait_repo
   detect_repo
 
-  # update max-pods to 200 #
+  # update max-pods to 200
   set_max_pods
+  set_reserved_resource
+  set_oem_cleanup_kubelet
   # A post-drain signal from Rancher doesn't mean RKE2 agent/server is already patched and restarted
   # Let's wait until the RKE2 settled.
   wait_rke2_upgrade
@@ -616,8 +689,10 @@ command_single_node_upgrade() {
   # wait all fleet bundles in limited time
   wait_for_fleet_bundles
 
-  # update max-pods to 200 #
+  # update max-pods to 200
   set_max_pods
+  set_reserved_resource
+  set_oem_cleanup_kubelet
   # Upgarde RKE2
   upgrade_rke2
 
--- a/pkg/api/node/formatter.go
+++ b/pkg/api/node/formatter.go
@@ -12,12 +12,12 @@ import (
 	"github.com/rancher/apiserver/pkg/apierror"
 	"github.com/rancher/apiserver/pkg/types"
 	"github.com/rancher/norman/httperror"
+	ctlbatchv1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/batch/v1"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/rancher/wrangler/v3/pkg/schemas/validation"
 	"github.com/rancher/wrangler/v3/pkg/slice"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -49,6 +49,8 @@ const (
 	seederAddonName              = "harvester-seeder"
 	defaultAddonNamespace        = "harvester-system"
 	nodeReady                    = "inventoryNodeReady"
+	enableCPUManager             = "enableCPUManager"
+	disableCPUManager            = "disableCPUManager"
 )
 
 var (
@@ -61,6 +63,8 @@ func Formatter(request *types.APIRequest, resource *types.RawResource) {
 	resource.AddAction(request, listUnhealthyVM)
 	resource.AddAction(request, maintenancePossible)
 	resource.AddAction(request, powerActionPossible)
+	resource.AddAction(request, enableCPUManager)
+	resource.AddAction(request, disableCPUManager)
 
 	if request.AccessControl.CanUpdate(request, resource.APIObject, resource.Schema) != nil {
 		return
@@ -81,6 +85,7 @@ func Formatter(request *types.APIRequest, resource *types.RawResource) {
 }
 
 type ActionHandler struct {
+	jobCache                    ctlbatchv1.JobCache
 	nodeCache                   ctlcorev1.NodeCache
 	nodeClient                  ctlcorev1.NodeClient
 	longhornVolumeCache         ctllhv1.VolumeCache
@@ -99,6 +104,8 @@ func (h ActionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		status := http.StatusInternalServerError
 		if e, ok := err.(*apierror.APIError); ok {
 			status = e.Code.Status
+		} else if statusError, ok := err.(*apierrors.StatusError); ok {
+			status = int(statusError.ErrStatus.Code)
 		}
 		rw.WriteHeader(status)
 		_, _ = rw.Write([]byte(err.Error()))
@@ -145,9 +152,44 @@ func (h ActionHandler) do(rw http.ResponseWriter, req *http.Request) error {
 			return apierror.NewAPIError(validation.InvalidBodyContent, fmt.Sprintf("Failed to decode request body: %v ", err))
 		}
 		return h.powerAction(toUpdate, input.Operation)
+	case enableCPUManager:
+		return h.enableCPUManager(toUpdate)
+	case disableCPUManager:
+		return h.disableCPUManager(toUpdate)
 	default:
 		return apierror.NewAPIError(validation.InvalidAction, "Unsupported action")
 	}
+}
+
+func (h ActionHandler) enableCPUManager(node *corev1.Node) error {
+	return h.requestCPUManager(node, ctlnode.CPUManagerStaticPolicy)
+}
+
+func (h ActionHandler) disableCPUManager(node *corev1.Node) error {
+	return h.requestCPUManager(node, ctlnode.CPUManagerNonePolicy)
+}
+
+func (h ActionHandler) requestCPUManager(node *corev1.Node, policy ctlnode.CPUManagerPolicy) error {
+	newNode := node.DeepCopy()
+	if newNode.Annotations == nil {
+		newNode.Annotations = make(map[string]string)
+	}
+
+	updateStatus := &ctlnode.CPUManagerUpdateStatus{
+		Status: ctlnode.CPUManagerRequestedStatus,
+		Policy: policy,
+	}
+
+	bytes, err := json.Marshal(updateStatus)
+	if err != nil {
+		return err
+	}
+	newNode.Annotations[util.AnnotationCPUManagerUpdateStatus] = string(bytes)
+	if _, err = h.nodeClient.Update(newNode); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func (h ActionHandler) cordonUncordonNode(node *corev1.Node, actionName string, cordon bool) error {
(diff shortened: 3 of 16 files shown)


pkg/api/node/schema.go
```diff
@@ -88,6 +89,8 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, _ config.Optio
 				maintenancePossible:          nodeHandler,
 				powerAction:                  nodeHandler,
 				powerActionPossible:          nodeHandler,
+				enableCPUManager:             nodeHandler,
+				disableCPUManager:            nodeHandler,
```
FrankYang0529, line 95: We may need to add related action in `ResourceActions`.

```go
			s.ResourceActions = map[string]schemas.Action{
				enableMaintenanceModeAction: {
					Input: "maintenanceModeInput",
				},
				disableMaintenanceModeAction: {},
				cordonAction:                 {},
				uncordonAction:               {},
				listUnhealthyVM:              {},
				maintenancePossible:          {},
				powerAction: {
					Input: "powerActionInput",
				},
				powerActionPossible: {},
				enableCPUManager: {},
				disableCPUManager: {},
			}
```

pkg/api/node/formatter.go
```diff
@@ -145,11 +151,40 @@ func (h ActionHandler) do(rw http.ResponseWriter, req *http.Request) error {
 			return apierror.NewAPIError(validation.InvalidBodyContent, fmt.Sprintf("Failed to decode request body: %v ", err))
 		}
 		return h.powerAction(toUpdate, input.Operation)
+	case enableCPUManager:
+		return h.enableCPUManager(toUpdate)
```
FrankYang0529, line 156: Probably, we can use `apierror.NewAPIError(validation.InvalidAction, err.Error())` as returned error.

pkg/api/node/formatter.go
```diff
@@ -145,11 +151,40 @@ func (h ActionHandler) do(rw http.ResponseWriter, req *http.Request) error {
...
+	if newNode.Annotations == nil {
+		newNode.Annotations = make(map[string]string)
+	}
+
+	updateStatus := &ctlnode.CPUManagerUpdateStatus{
+		Status: ctlnode.CPUManagerRequestedStatus,
+		Policy: policy,
+	}
+
+	jsonStr, _ := json.Marshal(updateStatus)
```
FrankYang0529: It looks like we silent error here. It's better to return it.

package/upgrade/upgrade_node.sh
```diff
@@ -490,10 +491,72 @@ EOF
...
 kubelet-arg:
-- "max-pods=200"
+- "max-pods=$MAX_PODS"
 EOF
 }
 
+set_system_reserved_resource() {
+  local cores=$(chroot /host nproc)
+  local reserverdCPU=$(calculateCPUReservedInMilliCPU $cores $MAX_PODS)
+  local systemReserverConfig="/host/etc/rancher/rke2/config.yaml.d/99-z00-harvester-reserved-resources.yaml"
```
bk201: We can replace `/host` with `$HOST_DIR`. sometimes it's useful for tests.

pkg/controller/master/node/cpu_manager_controller.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+
+	node.OnChange(ctx, CPUManagerControllerName, cpuManagerNodeHandler.OnNodeChanged)
+	job.OnChange(ctx, CPUManagerControllerName, cpuManagerNodeHandler.OnJobChanged)
+
+	return nil
+}
+
+func (h *cpuManagerNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev1.Node, error) {
+	if node == nil || node.DeletionTimestamp != nil {
```
bk201: nice to guard annotation == nil here too

pkg/controller/master/node/cpu_manager_controller.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+func getScript(nodeName string, policy CPUManagerPolicy) string {
+	var label string
+	if policy == CPUManagerStaticPolicy {
+		label = "true"
+	} else {
+		label = "false"
+	}
+	return fmt.Sprintf(`
```
bk201: Alternatively, you can store the script in a config map and mount the conifg map to the job. It provides some flexibility. We do something similar in the promote job (https://github.com/harvester/harvester/blob/7eb6121940c35b6172f6698a5081d314209d427e/deploy/charts/harvester/templates/configmap.yaml#L21)

scripts/build-iso
```diff
@@ -7,9 +7,9 @@ cd $(dirname $0)/..
 
 echo "Start building ISO image"
 
-HARVESTER_INSTALLER_VERSION=master
+HARVESTER_INSTALLER_VERSION=add-cpu-pinning-conf
```
bk201: Please help remove the testing code. We can give the reviewers a hint of how to build a correct ISO in the PR description.

pkg/controller/master/node/cpu_manager_controller.go
ibrokethecloud: agree, we should store this in a configmap as it makes it easier to provide workarounds to handle edge cases

pkg/webhook/resources/node/validator.go
```diff
@@ -71,3 +95,137 @@ func validateCordonAndMaintenanceMode(oldNode, newNode *corev1.Node, nodeList []
...
+
+	return nil
+}
+
+func checkCPUManagerLabel(node *corev1.Node, policy ctlnode.CPUManagerPolicy) error {
+	cpuManagerLabel, err := strconv.ParseBool(node.Labels[kubevirtv1.CPUManager])
+	// means CPUManager feature gate not enabled
+	if err != nil {
+		return werror.NewBadRequest("label cpumanager not found")
+	}
```
ibrokethecloud, line 148: This check could indicate a condition where the job is still running or has failed. 
I am not sure if this may be the best place to run this check or if may even be needed since we are checking subsequently in `checkCurrentNodeCPUManagerJobs` if the job is still running?

pkg/webhook/resources/node/validator.go
```diff
@@ -71,3 +95,137 @@ func validateCordonAndMaintenanceMode(oldNode, newNode *corev1.Node, nodeList []
...
+	// check if the cpu manager policy is the same
+	if ((policy == ctlnode.CPUManagerStaticPolicy) && cpuManagerLabel) || ((policy == ctlnode.CPUManagerNonePolicy) && !cpuManagerLabel) {
+		return werror.NewBadRequest("same cpu manager policy")
+	}
```
ibrokethecloud: minor nit: can we please make the error more descriptive to indicate current policy value is alredy matching the requested value?

pkg/webhook/resources/node/validator.go
```diff
@@ -71,3 +95,137 @@ func validateCordonAndMaintenanceMode(oldNode, newNode *corev1.Node, nodeList []
...
+		return nil
+	}
+
+	isRunning, err := isCPUManagerJobRunningOnNodes(jobCache, masterNodeNames)
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+	if isRunning {
+		return werror.NewBadRequest("there is other master node updating cpu manager policy")
+	}
```
ibrokethecloud: nit: we should probably return the name of node to make it the error a bit more descriptive and ease subsequent debugging if needed.

pkg/api/node/formatter.go
```diff
@@ -61,6 +63,8 @@ func Formatter(request *types.APIRequest, resource *types.RawResource) {
 	resource.AddAction(request, listUnhealthyVM)
 	resource.AddAction(request, maintenancePossible)
 	resource.AddAction(request, powerActionPossible)
+	resource.AddAction(request, enableCPUManager)
+	resource.AddAction(request, disableCPUManager)
```
ibrokethecloud, line 67: can we please make this condition? please refer to usage below
we could only allow enableCPUManager when we know either last operation failed or cpu manager is disabled, and similarly allow disabling of manager if cpu manager is successfully enabled or failed?

pkg/webhook/resources/node/validator.go
brandboat: Thanks for the review @ibrokethecloud , the idea of this check is to make sure harvester have kubevirt feature gate `CPUManager` enabled, in this feature gate, kubevirt update the `cpumanager` label periodically (~1 min) to check if node cpu-manager-policy is static or none. 
And in cpu_manager_controller.go#getScript https://github.com/harvester/harvester/pull/6190/files#diff-83df6178dc387757db24fc13a1b2ce905d80b0632b55bac72a63fae0b8fc590bR258-R272 I use this label to check if job succeed or fail.

---
FYI, about the kubevirt update cpumanager label,
https://github.com/kubevirt/kubevirt/blob/81e62b66e79650fbbdfa177b570ab8a4f5c5d086/pkg/virt-handler/heartbeat/heartbeat.go#L132-L157
periodically update (~1min)
https://github.com/kubevirt/kubevirt/blob/81e62b66e79650fbbdfa177b570ab8a4f5c5d086/pkg/virt-handler/heartbeat/heartbeat.go#L71-L75

ibrokethecloud: makes sense @brandboat. thanks for the clarification.

pkg/api/node/formatter.go
brandboat: > we could only allow enableCPUManager when we know either last operation failed or cpu manager is disabled, and similarly allow disabling of manager if cpu manager is successfully enabled or failed?

Thanks for the comment, I'm not sure doing this is necessary, as we already have checks in validator to ensure the enable/disable cpu manager action is valid. If user triggers these actions in an unsupported scenario, the action will be no-op. So I'm not strongly feel we should make those action conditional, perhaps I missed something, is there something I should be aware of ?

brandboat:
gentle ping @ibrokethecloud , @bk201 , could you take another look when you are available ? Thank you.

pkg/api/node/formatter.go
brandboat: Thanks for the comment @FrankYang0529, I modify the error handling a bit in https://github.com/harvester/harvester/pull/6190/files#diff-eb153a88748df1fbf46481b843d26f5882036ceca1c94824bf9fecfcd649c862R107
Now the error response from apiserver will be take care, so the error status code returned by validator should be correctly handled and return to the client side (I mean the one who triggered harvester `action`). Do you think this approach is valid ?

ibrokethecloud (approved):
lgtm. thanks.

we should in future iterations may be improve the helper script to bring the steps more in line with the documentation https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#changing-the-cpu-manager-policy

deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -145,3 +145,76 @@ data:
...
+    fi
+
+    printf 'kubelet-arg+:\n- "cpu-manager-policy=%s"' "$NODE_POLICY" > $CPU_MANAGER_CONFIG_FILE
+
+    if [ -f "$CPU_MANAGER_STATE_FILE" ]; then
+      mv "$CPU_MANAGER_STATE_FILE" "${CPU_MANAGER_STATE_FILE}.old"
+      echo "File $CPU_MANAGER_STATE_FILE has been renamed to ${CPU_MANAGER_STATE_FILE}.old"
+    else
+      echo "File $CPU_MANAGER_STATE_FILE does not exist."
+    fi
```
brandboat: Discussed offline with @ibrokethecloud , the flow of restart kubelet should be
- stop kubelet
- rm cpu_manager_state file
- start kubelet

brandboat: > we should in future iterations may be improve the helper script to bring the steps more in line with the documentation https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#changing-the-cpu-manager-policy

Already addressed in the latest commit, also rebased the feature branch.

pkg/api/node/formatter.go
FrankYang0529: Yes, sounds good to me.

FrankYang0529 (approved):
LGTM. Thanks for the PR.

Vicente-Cheng (approved):
Checked the upgrade script part.
LGTM, thanks!

brandboat:
rebased and fix conflicts.

pkg/webhook/resources/node/validator.go
```diff
@@ -71,3 +97,139 @@ func validateCordonAndMaintenanceMode(oldNode, newNode *corev1.Node, nodeList []
...
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+	if len(jobNames) > 0 {
+		return werror.NewBadRequest(fmt.Sprintf("there is other job %s updating the cpu manager policy for this node %s", strings.Join(jobNames, ", "), node.Name))
+	}
+	return nil
+}
+
+func checkMasterNodeJobs(node *corev1.Node, nodeCache v1.NodeCache, jobCache ctlbatchv1.JobCache) error {
```
bk201, line 167: sorry one question: how about the witness node? not mean to calculate it or not, but do we allow changing this on the witness node (it's not supposed to run any VMs).

pkg/controller/master/node/cpu_manager_controller.go
```diff
@@ -0,0 +1,315 @@
...
+
+	job, err := h.submitJob(cpuManagerStatus, node)
+	logrus.WithFields(logrus.Fields{
+		"node_name": node.Name,
+		"job_name":  job.Name,
+		"policy":    cpuManagerStatus.Policy,
+	}).Info("Submit cpu manager job")
+	if err != nil {
+		logrus.WithField("node_name", node.Name).WithError(err).Error("Submit cpu manager job failed")
+		return h.nodeClient.Update(updateNode(node, toFailedStatus(cpuManagerStatus)))
```
bk201: (optional) can also considering checking if need to update by using DeepEqual

pkg/webhook/resources/node/validator.go
brandboat: Perhaps we should avoid user to change cpu manager policy on the witness node ? I can add another validation to avoid that thing happen.

pkg/controller/master/node/cpu_manager_controller.go
brandboat: Sounds good, thanks for the comment.

brandboat:
~~Not sure why ci failed... this pr didn't change anything related to vm image. I'll take a look.~~
(Updated: rerun ci and it passed, look like it is flaky) 

```
verify image APIs operate via steve API verify image initialization succeeds
/go/src/github.com/harvester/harvester/tests/integration/api/image_apis_test.go:315
  STEP: create cirros image @ 09/03/24 07:34:18.27
  STEP: then the Initialized condition is true @ 09/03/24 07:34:18.289
time="2024-09-03T07:34:24Z" level=error msg="error syncing 'ssl-certificates': handler harvester-setting-controller: helmchartconfigs.helm.cattle.io \"rke2-ingress-nginx\" not found, requeuing"
time="2024-09-03T07:34:24Z" level=error msg="error syncing 'kubeconfig-default-token-ttl-minutes': handler harvester-setting-controller: error fetching setting kubeconfig-default-token-ttl-minutes: settings.management.cattle.io \"kubeconfig-default-token-ttl-minutes\" not found, requeuing"
time="2024-09-03T07:34:24Z" level=error msg="error syncing 'overcommit-config': handler harvester-setting-controller: settings.longhorn.io \"storage-over-provisioning-percentage\" not found, requeuing"
time="2024-09-03T07:34:24Z" level=error msg="error syncing 'harvester-control-plane': handler harvester-setting-controller: failed to parse NTP settings: unexpected end of JSON input, handler node-drain-controller: single controlplane cluster, cannot place controlplane in maintenance mode, requeuing"
time="2024-09-03T07:34:24Z" level=error msg="error syncing 'harvester-worker2': handler harvester-setting-controller: failed to parse NTP settings: unexpected end of JSON input, requeuing"
time="2024-09-03T07:34:24Z" level=error msg="error syncing 'harvester-worker3': handler harvester-setting-controller: failed to parse NTP settings: unexpected end of JSON input, requeuing"
time="2024-09-03T07:34:24Z" level=error msg="error syncing 'harvester-worker': handler harvester-setting-controller: failed to parse NTP settings: unexpected end of JSON input, requeuing"
time="2024-09-03T07:34:33Z" level=error msg="failed to check vmimage" error="Head \"[https://download.cirros-cloud.net/0.5.1/cirros-0.5.1-x86_64-disk.img\](https://download.cirros-cloud.net/0.5.1/cirros-0.5.1-x86_64-disk.img/)": context deadline exceeded (Client.Timeout exceeded while awaiting headers)" name=6z7rw namespace=default
  [FAILED] in [It] - /go/src/github.com/harvester/harvester/tests/integration/api/image_apis_test.go:346 @ 09/03/24 07:34:33.375
• [FAILED] [15.105 seconds]
verify image APIs operate via steve API [It] verify image initialization succeeds
/go/src/github.com/harvester/harvester/tests/integration/api/image_apis_test.go:315

  [FAILED] Expected
      <bool>: true
  not to be true
  In [It] at: /go/src/github.com/harvester/harvester/tests/integration/api/image_apis_test.go:346 @ 09/03/24 07:34:33.375

  Full Stack Trace
    github.com/harvester/harvester/tests/integration/api_test.init.func1.2.8.2()
    	/go/src/github.com/harvester/harvester/tests/integration/api/image_apis_test.go:346 +0x128
    reflect.Value.call({0x2bfb500?, 0xc00066c1e0?, 0xc003dcba20?}, {0x335e927, 0x4}, {0x5ba6ee0, 0x0, 0xc003dcba72?})
    	/usr/local/go/src/reflect/value.go:596 +0xca6
    reflect.Value.Call({0x2bfb500?, 0xc00066c1e0?, 0x20?}, {0x5ba6ee0?, 0x1?, 0x3?})
    	/usr/local/go/src/reflect/value.go:380 +0xb9
    github.com/onsi/gomega/internal.(*AsyncAssertion).buildActualPoller.func3()
    	/go/src/github.com/harvester/harvester/vendor/github.com/onsi/gomega/internal/async_assertion.go:325 +0x11f
    github.com/onsi/gomega/internal.(*AsyncAssertion).match(0xc000df8000, {0x3c4c930, 0xc0027da050}, 0x1, {0x0, 0x0, 0x0})
    	/go/src/github.com/harvester/harvester/vendor/github.com/onsi/gomega/internal/async_assertion.go:540 +0x7cc
    github.com/onsi/gomega/internal.(*AsyncAssertion).Should(0xc000df8000, {0x3c4c930, 0xc0027da050}, {0x0, 0x0, 0x0})
    	/go/src/github.com/harvester/harvester/vendor/github.com/onsi/gomega/internal/async_assertion.go:145 +0x86
    github.com/harvester/harvester/tests/framework/dsl.MustFinallyBeTrue(0xc00066c1e0, {0xc00066c200?, 0x0?, 0x3?})
    	/go/src/github.com/harvester/harvester/tests/framework/dsl/action.go:40 +0x114
    github.com/harvester/harvester/tests/integration/api_test.init.func1.2.8()
    	/go/src/github.com/harvester/harvester/tests/integration/api/image_apis_test.go:343 +0x3a8
```

pkg/controller/master/node/cpu_manager_controller.go
brandboat: All comments are addressed in 60ce8aa, thanks !

bk201 (approved):
LGTM, please squash before merge

brandboat:
@Mergifyio backport v1.4

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -465,3 +479,7 @@ func hostDevicesOvercommitNeeded(oldVM, newVM *kubevirtv1.VirtualMachine) bool {
 	}
 	return false
 }
+
+func isDedicatedCPU(vm *kubevirtv1.VirtualMachine) bool {
+	return vm.Spec.Template.Spec.Domain.CPU != nil && vm.Spec.Template.Spec.Domain.CPU.DedicatedCPUPlacement
```
w13915984028, line 484: It seems the condition should be 

>...&& vm.Spec.Template.Spec.Domain.CPU.DedicatedCPUPlacement != nil && *vm.Spec.Template.Spec.Domain.CPU.DedicatedCPUPlacement

```
	// +optional
	DedicatedCPUPlacement *bool `json:"dedicatedCPUPlacement,omitempty"`
```

pkg/controller/master/node/cpu_manager_controller.go
```diff
@@ -0,0 +1,322 @@
...
+		logrus.WithField("node_name", node.Name).WithError(err).Error("Skip update cpu manager policy, failed to retrieve cpu-manager-update-status from annotation")
+		return node, nil
+	}
+	// do nothing if status is not requested
+	if cpuManagerStatus.Status != CPUManagerRequestedStatus {
+		return node, nil
+	}
+
+	newNode := node.DeepCopy()
+	job, err := h.submitJob(cpuManagerStatus, node)
```
w13915984028, line 112: There are scenarios: 

`submitJob` is successful but `h.nodeClient.Update` failed (e.g. node is updated by other controller on the meanwhile)

following code will re-create the job;  LB code ever hit similar issue and caused IP duplicated allocation

this is a general issue when a controller updates several CRD objects

because the job is created with generated name, there is no elegant solution:
either before submitJob, check if the label matched job has been there;
or in the job itself, check if a peer is running

```
	job, err := h.submitJob(cpuManagerStatus, node)
...
	return h.nodeClient.Update(newNode)
	```

File: pkg/controller/master/node/cpu_manager_controller.go [hunk 17]
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	if reflect.DeepEqual(newNode, node) {
+		return node, nil
+	}
+	return h.nodeClient.Update(newNode)
+}
+
+func (h *cpuManagerNodeHandler) OnJobChanged(_ string, job *batchv1.Job) (*batchv1.Job, error) {
+	if job == nil || job.Labels == nil || job.DeletionTimestamp != nil {
```
w13915984028, line 132: When a job is running but then deleted by user manually, are there any chance the node is updated with successful/failed?

deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -145,3 +145,94 @@ data:
...
+    }
+
+    function wait_for_label() {
+      local node_name=$1
+      local expect_label_value=$2
+      local interval=5
+      local elapsed=0
+
+      while [ $elapsed -lt $WAIT_LABEL_TIMEOUT ]; do
+        if ! labels=$($KUBECTL get node "$node_name" --show-labels); then
```
w13915984028, line 185: with a `time-out`, it is better to retry when `kubectl` returns error

deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -145,3 +145,94 @@ data:
...
+      fi
+    }
+
+    function manage_service() {
+      local service=$1
+
+      echo "Stopping ${service}."
+      if ! chroot $HOST_DIR systemctl stop $service; then
+        echo "Error: failed to stop ${service}."
+        exit 1
```
w13915984028, line 164: if each none-zero return can have a different return code (1,2,3...), it is better for troubleshooting

w13915984028 (commented):
Thanks for the wonderful PR to bring this new feature.

Sorry for the late reviewing.

From the robustness view of point, please consider to add a new PR to enhance those points, thanks.

pkg/webhook/resources/virtualmachine/mutator.go
brandboat: Good catch, many thanks

brandboat: Correct me if I'm wrong. Looks like the DedicatedCPUPlacement field is `bool`, not `*bool`. 
https://github.com/harvester/harvester/blob/73c3584d264228463648e7fe7ba5d40fbf22cda8/vendor/kubevirt.io/api/core/v1/schema.go#L328

w13915984028: You are right. I was confused by the kubevirt code, there was another `DedicatedCPUPlacement` defined as `*bool`...

```
vendor/kubevirt.io/api/instancetype/v1beta1/types.go:	DedicatedCPUPlacement *bool `json:"dedicatedCPUPlacement,omitempty"`

vendor/kubevirt.io/kubevirt/pkg/instancetype/instancetype.go:		vmiSpec.Domain.CPU.DedicatedCPUPlacement = *instancetypeSpec.CPU.DedicatedCPUPlacement
```

deploy/charts/harvester/templates/configmap.yaml
brandboat: gentle ping @w13915984028, all comments are addressed in https://github.com/harvester/harvester/pull/6517, please take a look, thanks.

Issue #5150, opened 2024-02-15
[FEATURE] Support for iscsi boot disks


ibrokethecloud:
**Is your feature request related to a problem? Please describe.**

Large customers running diskless systems may wish to use something like an iscsi disk as the boot device.

Most modern hardware allows definition of an iscsi target as a boot device. 

As we have tested internally when booting of an iscsi disk additional kernel arguments `rd.iscsi.firmware rd.iscsi.ibft` are needed to ensure that the OS can boot successfully.

**Describe the solution you'd like**

We should have an additional check in the installer / upgrade path to detect if the install disk is an iscsi disk, and accordingly append these additional kernel arguments in the default entry.

These arguments will need to be retained post upgrade as well.

**Describe alternatives you've considered**

**Additional context**

abonillabeeche:
We may need to include the support to use the OS and DATA mpio disks rather than the individual disk. This means the initramfs would need mpio support. See [here](https://documentation.suse.com/sles/15-SP5/html/SLES-all/cha-multipath.html#sec-multipath-planning)

Otherwise, we'd need to disable mpio for the [root/OS](https://documentation.suse.com/sles/15-SP5/html/SLES-all/cha-multipath.html#pr-multipath-disable-root) disk.

innobead:
Let's add HEP for this, even though we have a KB for reference. 

cc @bk201

ibrokethecloud:
For testing the changes:

The multipath support can be validated in a kvm setup as follows:

* Create a raw disk image of atleast 250GB
![image](https://github.com/user-attachments/assets/b60936cd-8c7f-421a-8fb7-a3a56c73777c)

* Add the same disk twice to an VM. Please ensure the following
   - The disk is marked sharable
   - A serial is provided, say disk1. The serial info needs to be same on both scsi disk additions
 
![image](https://github.com/user-attachments/assets/7008208c-7926-4b72-b8ea-d5780a219e02)
![image](https://github.com/user-attachments/assets/aaaf7767-776c-49bb-aee3-99a010a1a087)

* Boot from the iso.  Despite two paths being added to the same underlying disk, the harvester installer only shows the 1 installable disk. The serial info is important here as it is used to dedup disks with same serial info and/or wwn number
*  Proceed with installation as normal, and specify the following harvester config via a config url to enable multipathd and configure multipath.conf. The `additionalKernelArguments` are not needed for this example to work, but have been added to help validate the arguments being passed to kernel via grub post install
```
os:
[...]

PR #6229, opened 2024-07-26
HEP support remote disk installation [CI SKIP]

ibrokethecloud:

**Problem:**

Design proposal to detail changes needed to harvester to allow installation and boot from remote disks.

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/5150
**Test plan:**

### DIFF
--- a/enhancements/20240726-non-local-boot-disks.md
+++ b/enhancements/20240726-non-local-boot-disks.md
@@ -1,0 +1,97 @@
+# Install Harvester on diskless systems
+
+## Summary
+
+Support Harvester installation to diskless systems
+
+### Related Issues
+
+[#5150](https://github.com/harvester/harvester/issues/5150)
+
+## Motivation
+
+Enterprise users may wish to install Harvester on diskless systems which have boot lun's served from iscsi or san based disk arrays. The HEP details changes need to Harvester to support such possible scenarios
+
+### Goals
+
+Allows users to install and boot Harvester from non-local disks.
+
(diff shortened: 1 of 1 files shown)


bk201 (approved):
Thanks!

enhancements/20240726-non-local-boot-disks.md
```diff
@@ -0,0 +1,97 @@
...
+### Implementation Overview
+
+An end user would configure their host to leverage capable nic to boot off remote iscsi disks. 
+
+Once the hardware nic configuration is complete, the harvester installer should see the iscsi disks in the disk panel
+
+**NOTE:** in case of multipathing the same disk may appear twice
+
+User just needs to choose one of the disks only. In case data needs to be on the same disk please ensure to choose the same disk identifier for the data disk too
+
```
tserong, line 70: I think the above paragraph and note can be removed since we're deduping the devices.

tserong (approved):
LGTM, just one little tweak

enhancements/20240726-non-local-boot-disks.md
Vicente-Cheng: +1 I remember we already resolved this.
cc @ibrokethecloud

Vicente-Cheng: Hi @ibrokethecloud, 
Could we update that?
Then we can merge it HEP as completed. Thanks!

Issue #5102, opened 2024-02-04
[ENHANCEMENT] Manage upgrade-able versions in Harvester clusters


bk201:
**Is your enhancement related to a problem? Please describe.**

It's about the Harvester version sync.

**Describe the solution you'd like**

There were situations we wanted to revoke a published version due to sane reasons before. But currently, the version sync mechanism in Harvester can't do this even if we remove a version in the upgrade responder. We need a way to handle this.

Another need is to prohibit upgrading from a non-release previous minor version. For example, say we now release v1.3.0 and set the minimum version to v1.2.2. But later we released v1.2.3 and found it's not possible to upgrade from v1.2.3 to v1.3.0, we want the v1.3.0 version to vanish in the user's cluster (The user can still see v1.3.0 in the versions list because v1.2.3 meets the minimum version of v1.3.0).

**Describe alternatives you've considered**

**Additional context**

ibrokethecloud:
Test instructions

## Build upgrade-responder
* clone https://github.com/longhorn/upgrade-responder
* checkout branch v0.1.4
* run `make` to compile the responder
* a compiled binary will be available under ./bin directory

## Test version creation
* install harvester v1.4.0-rc3 to a cluster
* use the following response.json with the upgrade responder
```
{
    "versions": [
      {
        "Name": "v1.3.2",
        "ReleaseDate": "2024-09-07T10:00:00Z",
        "MinUpgradableVersion": "v1.3.1",
        "Tags": [
          "latest",
          "v1.3.2"
        ],
        "ExtraInfo": null
      },
      {
        "Name": "v1.4.0-rc3",
        "ReleaseDate": "2021-05-13T10:00:00Z",
        "MinUpgradableVersion": "v1.3.0",
        "Tags": [
          "v1.4.0-rc3",
          "dev"
        ],
        "ExtraInfo": null
      },
      {
        "Name": "v1.3.1",
        "ReleaseDate": "2024-06-14T10:00:00Z",
        "MinUpgradableVersion": "v1.2.2",
        "Tags": [
          "v1.3.1"
        ],
        "ExtraInfo": null
      }
    ],
    "requestIntervalInMinutes": 2
  }
```
* run the upgrade responder as follows
```
 ./bin/upgrade-responder --debug start --upgrade-response-config $path-to-response/response.json --geodb ./geodb/GeoLite2-City.mmdb --application-name harvester
```
[...]

PR #6242, opened 2024-07-31
harvester upgrade lookup enhancements

ibrokethecloud:

**Problem:**

Harvester upgrade responder currently does not have the logic to clean up Version CRD's if the ugprade responder version gets udpated. 
**Solution:**

The PR changes on version sync is handled in the harvester controller. This will allow for harvester team to update `MinUpgradableVersion` for an existing version, to trigger harvester clusters to update the version objects available for upgrade.
This will allow us to technically trigger rollback of version upgrade offers available from the harvester responder

**Related Issue:**
https://github.com/harvester/harvester/issues/5102
**Test plan:**

### DIFF
--- a/pkg/controller/master/upgrade/version_syncer.go
+++ b/pkg/controller/master/upgrade/version_syncer.go
@@ -149,17 +149,21 @@ func (s *versionSyncer) getExtraInfo() (map[string]string, error) {
 }
 
 func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion string) error {
-	if err := s.cleanupVersions(currentVersion); err != nil {
+	if err := s.cleanupVersions(currentVersion, resp.Versions); err != nil {
 		return err
 	}
 
+	// iterate over response and identify if a new version needs to be created
 	for _, v := range resp.Versions {
 		newVersion, err := s.getNewVersion(v)
 		if err != nil {
 			return err
 		}
 
-		if !canUpgrade(currentVersion, newVersion) {
+		// newVersion CRD is created from the version.yaml created from the release artifact
+		// this needs to be changed to use the min version coming from the upgrade responder
+		// as this could be dynamically update and will cause cleanup of older versions
+		if !canUpgrade(currentVersion, newVersion, v) {
 			continue
 		}
 
@@ -174,21 +178,28 @@ func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion s
 			}
 		}
 	}
+
 	return nil
 }
 
 // cleanupVersions remove version resources that's can't be upgraded to anymore
-func (s *versionSyncer) cleanupVersions(currentVersion string) error {
+func (s *versionSyncer) cleanupVersions(currentVersion string, remoteVersions []Version) error {
 	versions, err := s.versionClient.List(s.namespace, metav1.ListOptions{})
 	if err != nil {
 		return err
 	}
 
-	for i := range versions.Items {
-		v := versions.Items[i]
-		if !canUpgrade(currentVersion, &v) {
-			if err := s.versionClient.Delete(v.Namespace, v.Name, &metav1.DeleteOptions{}); err != nil {
-				return err
+	for _, version := range remoteVersions {
+		for i, v := range versions.Items {
+			versionObj := v
+			if !canUpgrade(currentVersion, &versionObj, version) {
+				if err := s.versionClient.Delete(v.Namespace, v.Name, &metav1.DeleteOptions{}); err != nil {
+					if apierrors.IsNotFound(err) {
+						continue // likely object has been removed so we can continue
+					}
+					return err
+				}
+				versions.Items = append(versions.Items[:i], versions.Items[i+1:]...)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/master/upgrade/version_syncer.go
```diff
@@ -174,21 +178,24 @@ func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion s
 			}
 		}
 	}
+
 	return nil
 }
 
 // cleanupVersions remove version resources that's can't be upgraded to anymore
-func (s *versionSyncer) cleanupVersions(currentVersion string) error {
+func (s *versionSyncer) cleanupVersions(currentVersion string, Versions []Version) error {
```
bk201:
```suggestion
func (s *versionSyncer) cleanupVersions(currentVersion string, remoteVersions []Version) error {
```

Can we rename the variable to something like this for readability?

pkg/controller/master/upgrade/version_syncer.go
```diff
@@ -174,21 +178,24 @@ func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion s
...
 		return err
 	}
 
-	for i := range versions.Items {
-		v := versions.Items[i]
-		if !canUpgrade(currentVersion, &v) {
-			if err := s.versionClient.Delete(v.Namespace, v.Name, &metav1.DeleteOptions{}); err != nil {
-				return err
+	for _, version := range remoteVersions {
+		for i := range versions.Items {
```
w13915984028: the second for loop looks a bit redundant:

the new `canUpgrade` mainly check the `current` and `remote`; mostly `!canUpgrade` will let second loop run and delete those item repeatly

pkg/controller/master/upgrade/version_syncer.go
```diff
@@ -149,17 +149,21 @@ func (s *versionSyncer) getExtraInfo() (map[string]string, error) {
 }
 
 func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion string) error {
-	if err := s.cleanupVersions(currentVersion); err != nil {
+	if err := s.cleanupVersions(currentVersion, resp.Versions); err != nil {
 		return err
 	}
```
w13915984028, line 154: it is not related to this PR,

but the `syncVersions` is not stopped when upgrade is ongoing; we had other PR to do not show the newly available versions after Harvester is just upgraded (with new currentVersion) but the upgrade is not finished

it looks the better solution is to stop sync & `cleanupVersions` to avoid potential effect against upgrade

w13915984028 (commented):
two questions, thanks.

pkg/controller/master/upgrade/version_syncer.go
ibrokethecloud: the upgrade controller does not depend on the version object to exist. The only time version object is referred is when the upgrade controller needs to create a repo image. If an upgrade has create the image then deleting the version will have no real impact. If version is deleted before that then upgrade will fail which may not be a bad idea since the version would have been recalled for a specific reason

pkg/controller/master/upgrade/version_syncer.go
ibrokethecloud: multiple versions may be possible for a current version based on the remoteVersions returned from upgrade responder. 

We need to run the inner loop to reconcile all local version CRD's with what is returned by upgrade responder. 

With this change we can rollback a version by bumping up the remoteVersion in upgrade responder to a really high semver say 1.99.x which would never be satisfied and trigger clean up version CRD's in clusters when they fetch the latest remote versions, there by preventing users from upgrading to the version.

starbops: Suppose a local version object is removed due to `!canUpgrade` in the first iteration. The removed version object will be iterated again with the following comparison round (because it's a nested loop and we don't remove the element from the list). The same local version object might fail the test again, causing the subsequent version deletion in a "version not found" error.

pkg/controller/master/upgrade/version_syncer.go
```diff
@@ -226,13 +233,13 @@ func (s *versionSyncer) getNewVersion(v Version) (*harvesterv1.Version, error) {
 	return &newVersion, nil
 }
 
-func canUpgrade(currentVersion string, newVersion *harvesterv1.Version) bool {
+func canUpgrade(currentVersion string, newVersion *harvesterv1.Version, responderVersion Version) bool {
 	switch {
 	case newVersion.Spec.ISOURL == "" || newVersion.Spec.ISOChecksum == "":
```
starbops, line 242: If these two cases of version objects should never exist in the cluster, maybe we can move this to the validator to simplify the logic of the `canUpgrade` function?

starbops (commented):
The functionality works as expected.

Just a suggestion: can we separate the logic of "syncing local version objects and responder versions" from "ensuring local version objects are upgradable according to responder versions"?

pkg/controller/master/upgrade/version_syncer.go
ibrokethecloud: thanks for the feedback, I have fined tuned the inner loop to drop deleted version objects and also ignore errors if a version object is not found during deletion

pkg/controller/master/upgrade/version_syncer.go
ibrokethecloud: we could do it later, it seems to have been around forever

pkg/controller/master/upgrade/version_syncer.go
```diff
@@ -174,21 +178,29 @@ func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion s
 			}
 		}
 	}
+
 	return nil
 }
 
 // cleanupVersions remove version resources that's can't be upgraded to anymore
-func (s *versionSyncer) cleanupVersions(currentVersion string) error {
+// cleanupVersions remove version resources that's can't be upgraded to anymore
```
starbops: Duplicated comments.

starbops (approved):
LGTM, thank you.

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

ibrokethecloud:
@mergifyio backport v1.4

Issue #3129, opened 2022-11-09
[FEATURE] Add volume encryption support for Harvester


nikolaishields:
**Is your feature request related to a problem? Please describe.**
 Presently volumes provisioned within harverster via the longhorn provisioner do not support the creation of encrypted volumes. It would be beneficial to security conscious users to have the ability to enable volume encryption at the storage class level. 

**Describe the solution you'd like**

By adding `volumeMode: block encryption` support to the harvester pvc request users could freely create both encrypted or non-encrypted volumes.

**Describe alternatives you've considered**

**Additional context**

https://longhorn.io/docs/1.4.0/advanced-resources/security/volume-encryption/

Yu-Jack:
HEP: https://github.com/harvester/harvester/pull/6270
Backend: https://github.com/harvester/harvester/pull/6348

lanfon72:
Verified this feature has been applied but with some minor problem:

1. There is not a way to verify VM image is encrypted or not. (or should we verify it?)
2. The way to verify the passphrase is not work, the pvc located in `/dev/longhorn` will only appear when it is attached to a running VM, but we can't open, if we stop the VM, the file will be removed. (as the snapshot)
![image](https://github.com/user-attachments/assets/9f37f11f-71e8-4a2f-8c5f-bc1297a10c38)

Both are not affect to the implementation, but we would need update HEP/Document to clarify it. 

move back to ready for testing.

Test Information
----
* Environment: **qemu/KVM 3 nodes**
* Harvester Version: **v1.4.0-rc1**
* **ui-source** Option: **Auto**
* Rancher Version: **v2.8.2** (VCluster)[^2]

### Verify Steps
#### Prerequisite
1. Install Harvester with any nodes
3. Create an image `img1` as the base image.
4. Create a secret `en-1` with data keys (in namespace `default`)
    - CRYPTO_KEY_CIPHER: aes-xts-plain64
    - CRYPTO_KEY_HASH: sha512
    - CRYPTO_KEY_PROVIDER: secret
    - CRYPTO_KEY_SIZE: 512
    - CRYPTO_KEY_VALUE: the encryption passphrase
    - CRYPTO_PBKDF: argon2i
5. Create a storage class `sc-en1` with volume encryption enabled by using secret `en-1`

#### Case 1 : image creation
1. Create encrypt image `en-img1` from `img1`, use storage class `en-1`
2. Image should created successfully
[...]

PR #6270, opened 2024-08-06
HEP: support block volume encryption and decryption

Yu-Jack:
**Summary:**
This HEP doesn't include the GUI changes, it focus on backend (It's v1.4.0 release goal). But, we're still able to use GUI to choose encrypted/decrypted boot volume and data volume.

**Related Issue:**
https://github.com/harvester/harvester/issues/3129

**Additional Notes**
- PR https://github.com/harvester/harvester/pull/6348


enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,333 @@
...
+
+Since Longhorn uses clone mechanism to encrypt/decrypt Boot volume, so Harvester will follow that. This means that there would be two virtual machine image after encryption. One is encrypted, and the other is source virtual machine image. 
+
+Data volumes are different from Boot volume. Because Longhorn supports block encryption/decryption, so Harvester could encrypt/decrypt data volume directly. 
+
+Regardless of volumes used, Longhorn will decrypt volume when it's attached to virtual machine. So, we'll be able to read the boot volumes and data volumes in the virtual machine. 
+
+### User Stories
+
+#### Story 1 - Encrypt/Decrypt Boot Volumes
```
innobead: Another story: Users should also be able to upload/reuse an encrypted image and create encrypted VMs based on it directly.

Yu-Jack: @innobead 

Could you elaborate more about "upload an encrypted image" ?

The original use case aims to prepare a image (upload image or use URL to download), then enable encryption. After that, users could reuse this encrypted image to create VM. I think original title was confused, so I changed it.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,335 @@
...
+
+Alice wants to encrypt the current boot volume for security, so Alice follows up the guideline:
+
+1. Prepare a virtual machine image, which you could upload or use URL to download.
+2. Create a secret and template storage class for encryption/decryption.
+3. Create a new virtual machine image with encryption enabled.
+
+After that, Alice gets an encrypted virtual machine image as boot volume. Then, Alice can start the virtual machine with selecting encrypted boot volume in GUI. For security, Alice decides to delete that original virtual machine image. For this moment, Alice only has this encrypted virtual machine image and Alice can reuse this boot volume.
+
+Over time, Alice wants to decrypt the encrypted virtual machine image. Alice follows up the guideline:
```
w13915984028: For security consideration, we may add option when encrypt an image:
With pass/key to decrypt
Always allow to decrypt

Then if some user (maybe not the owner) plans to decrypt certain image, he needs to provide proof.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,335 @@
...
+
+## Proposal
+
+Since Longhorn uses clone mechanism to encrypt/decrypt Boot volume, so Harvester will follow that. This means that there would be two virtual machine image after encryption. One is encrypted, and the other is source virtual machine image. 
+
+![](./20240805-encrypted-block-volume/lh-flow.png)
+
+Data volumes are different from Boot volume. Because Longhorn supports block encryption/decryption, so Harvester could encrypt/decrypt data volume directly. 
+
+Regardless of volumes used, Longhorn will decrypt volume when it's attached to virtual machine. So, we'll be able to read the boot volumes and data volumes in the virtual machine.
```
w13915984028, line 36: Can user further create new images/volumes from the running VM ? If user can freely do, it will be a possible security issue.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,335 @@
...
+
+Then, we could get the encrypted backing image in Longhorn, and use it as boot volume in Harvester.
+
+
+### Test plan
+
+- Verify that volumes can be encrypted and decrypted as specified. 
+- Verify that encrypted boot volume and data volume can be attached to virtual machines.
+- Verify that the encrypted data volume, when used as a hot plug volume, can be attached to virtual machines.
+- Ensure that encrypted volumes are not accessible without the correct decryption keys.
```
w13915984028: Do you have details abut this?

Can user create the decryption key when needed instead of initially.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,335 @@
...
+    spec:
+        displayName: decrypt-alpine-extended-3.20.2-x86_64.iso
+        retry: 3
+        sourceType: clone
+        storageClassParameters:
+            migratable: "true"
+            numberOfReplicas: "3"
+            staleReplicaTimeout: "30"
+        sourceParameters:
+            secretName: "longhorn-crypto"
```
w13915984028: Any plan to protect the related secrets from being removed accidently?

w13915984028 (commented):
Thanks for the clear and detailed HEP, a few questions regarding some security considerations.

enhancements/20240805-encrypted-block-volume.md
Yu-Jack: I think it's a complicated topic.

Since LH only provides k8s secret as key management, it's not enough for encryption/decryption at this moment. For more securities, we might need more different key managements.

Although we could force user to type the key in GUI each time when they'd like to encrypt/decrypt, but there are still issues by manually creating yaml. 

Besides, if we'd like to control the encrypt and decrypt permission, we probably need sub resource to define permission.

enhancements/20240805-encrypted-block-volume.md
Yu-Jack: Updated:

Although we can export image from running VMs, but it can't boot successfully. Cause it just attaches the boot volume without decryption,.

---
For now, yes. Continue with previous comment, we should encourage users to use RBAC to avoid restricted users from exporting images/volumes from running VMs.

enhancements/20240805-encrypted-block-volume.md
Yu-Jack: The logic behind the encryption/decryption uses LUKS. So yes, even If you know the key, you can still decrypt replicas to read the data on the node.

enhancements/20240805-encrypted-block-volume.md
Yu-Jack: Good idea, we should avoid this happening.

enhancements/20240805-encrypted-block-volume.md
w13915984028: No worries. We can claim cretain features are not covered in this HEP, and add related information and the possible workaround in the coming document. User will know we have considerations on such.

Yu-Jack: Yeah, it's a good point. Regardless of ways we choose, I think we should end up providing a full guideline about how to secure the volume based on my mechanism. I'll keep in mind while I'm doing this feature.

Yu-Jack: BTW, I'll add this part into enhancement as well, thanks.

enhancements/20240805-encrypted-block-volume.md
innobead: I meant, if users download an encrypted image, then upload it again to the same cluster or another cluster, can we just reuse it to create encrypted volumes if they provide the same secret.

Yu-Jack: @innobead No, they can't. We don't support it. If we'd like to do this, we need to add some fields on GUI to let users fill out some encryption related information. Otherwise, backend can't use the longhorn clone mechanism to encrypt image.

Yu-Jack:
Gentle ping to @innobead @w13915984028 , I've updated the HEP based on feedbacks from HEP meeting and other discussions in private.

@bk201 @Vicente-Cheng @WebberHuang1118 I also added a brief summary of our discussion in "Note" section.

w13915984028 (approved):
LGTM, thanks.

Yu-Jack:
Changed name from `spec.securityParameters.encryption` to `spec.securityParameters.cryptoOperation`. It's weird to put `decrypt` in `spec.SecurityParameters.encryption` field.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,440 @@
...
+        fromBackup: ""
+        migratable: "true"
+        encrypted: "true"
+        csi.storage.k8s.io/provisioner-secret-name: "longhorn-crypto"
+        csi.storage.k8s.io/provisioner-secret-namespace: "longhorn-system"
+        csi.storage.k8s.io/node-publish-secret-name: "longhorn-crypto"
+        csi.storage.k8s.io/node-publish-secret-namespace: "longhorn-system"
+        csi.storage.k8s.io/node-stage-secret-name: "longhorn-crypto"
+        csi.storage.k8s.io/node-stage-secret-namespace: "longhorn-system"
+    ```
```
WebberHuang1118, line 116: IIUC, the sc is not required for image encryption? Just want to make sure, thanks.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,440 @@
...
+        namespace: default
+    spec:
+        displayName: decrypt-alpine-extended-3.20.2-x86_64.iso
+        retry: 3
+        sourceType: clone
+        securityParameters:
+            encryption: "decrypt"
+            sourceImageName: "encrypted-alpine-extended" # not display name
+            sourceImageNamespace: "default"
+    ```
```
WebberHuang1118: Will there be a dedicated sc for the encrypted and decrypted image? Just make sure, thanks.

enhancements/20240805-encrypted-block-volume.md
Yu-Jack: In the first generation of solution, I bound the secret with `virtualmachineimage`. At that moment, we actually didn't need to create SC. 

However, after our discussion, I think we could directly use storage class from annotation to pass secret to `backingimage`. It makes `virtualmachineimage` not to bind with secret, but with SC. I think it makes more sense for our current flow and more flexible. (When we're trying to create `virtualmachineimage`, we need to select SC in GUI.)

enhancements/20240805-encrypted-block-volume.md
WebberHuang1118: For image encryption and decryption, can the source and target images reside in different namespaces?

WebberHuang1118 (approved):
LGTM, just a few questions, thanks for the HEP.

enhancements/20240805-encrypted-block-volume.md
Yu-Jack: > Will there be a dedicated sc for the encrypted and decrypted image? Just make sure, thanks.

No. Users need to create it by themselves.


> For image encryption and decryption, can the source and target images reside in different namespaces?

Yes.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,440 @@
...
+    annotations:
+        harvesterhci.io/storageClassName: harvester-longhorn-encryption
+    name: encrypted-alpine-extended
+    namespace: default
+spec:
+    displayName: encrypted-alpine-extended-3.20.2-x86_64.iso
+    retry: 3
+    sourceType: clone
+    securityParameters:
+        cryptoOperation: "encrypt"
```
a110605, line 267: kindly remind to update `encryption` -> `cryptoOperation` in all YAML.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,440 @@
...
+        annotations:
+          harvesterhci.io/storageClassName: harvester-longhorn-encryption
+        name: encrypted-alpine-extended
+        namespace: default
+    spec:
+        displayName: encrypted-alpine-extended-3.20.2-x86_64.iso
+        retry: 3
+        sourceType: clone
+        securityParameters:
+            encryption: "encrypt"
```
a110605: kindly remind to update `encryption` -> `cryptoOperation` in all YAML.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,440 @@
...
+
+Yes. Please follow answer of Q1 to use Rancher Integration to create different projects and namespaces. Storage class is available to everyone, normal users can still read it. But, they can't create it.
+
+The only way for normal users to create Storage Class is to request admin to create it.
+
+### Q3: How do we ***really*** know this volume is encrypted or not?
+
+For end-user, they never know. The CSI driver automatically decrypts the volume when volumes are attached to VMs.
+
+But, for admin which can log in to host, they could check `/dev/longhorn/pvc-xxxx` by command `cryptsetup luksDump {pvc_name}` after VMs start. Please check screenshots of [Test Plan](#test-plan).
```
a110605, line 441: Do we have any field in image & volume APIs response indicates the image & volume is encrypted or not?  

If have, we can show a lock icon before image/volume name or add a new column named `encrypted`.

See longhorn screenshot
Screenshot 2024-09-05 at 11 45 35 AM

Yu-Jack: About image, you could use securityParameters as condition, if `spec.securityParameters.cryptoOperation` is `encrypt`, then it's a encrypted image.

About volume, as I know, `v1/harvester/persistentvolumes` and `v1/harvester/longhorn.io.volumes` both contain a fields call `encrypted`. One is under `spec.csi.volumeAttributes.encrypted` of `v1/harvester/persistentvolumes`, another one is under `spec.encrypted` of `v1/harvester/longhorn.io.volumes`.

Yu-Jack: I agree with adding a lock icon as well.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,440 @@
...
+
+1. Prepare an encrypted virtual machine image.
+2. Create a new virtual machine image with decryption enabled.
+
+After that, Admin will have an encrypted virtual machine image and an unencrypted virtual machine image. Admin could select one of them to start the virtual machine in GUI.
+
+
+#### Story 2 - Use encrypted data volume in virtual machine
+
+Admin wants to encrypt a data volume for security, so Admin follows up the guideline:
```
bk201, line 64: Admin prepares the secret and storage class. But any user who creates volumes from the storage class can enjoy the encryption at rest feature. The paragraph can be clarified that a non-admin user still can use the encryption.

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,440 @@
...
+        encrypted: "true"
+        csi.storage.k8s.io/provisioner-secret-name: "longhorn-crypto"
+        csi.storage.k8s.io/provisioner-secret-namespace: "longhorn-system"
+        csi.storage.k8s.io/node-publish-secret-name: "longhorn-crypto"
+        csi.storage.k8s.io/node-publish-secret-namespace: "longhorn-system"
+        csi.storage.k8s.io/node-stage-secret-name: "longhorn-crypto"
+        csi.storage.k8s.io/node-stage-secret-namespace: "longhorn-system"
+    ```
+
+4. Create Encrypted Virtual Machine Image
```
bk201: Suggest adding "from a non-encrypted Virtual Machine Image"

enhancements/20240805-encrypted-block-volume.md
```diff
@@ -0,0 +1,440 @@
...
+After that, Admin gets an encrypted virtual machine image as boot volume. Then, Admin can start the virtual machine with selecting encrypted boot volume in GUI. 
+
+For security reason, Admin decides to delete that original virtual machine image. The encrypted image is independent of the original image. 
+
+For this moment, Admin only has this encrypted virtual machine image and Admin can reuse this boot volume.
+
+Over time, Admin wants to decrypt the encrypted virtual machine image. Admin follows up the guideline:
+
+1. Prepare an encrypted virtual machine image.
+2. Create a new virtual machine image with decryption enabled.
```
bk201: Maybe one step is missing here: decrypt the encrypted virtual machine image.

bk201 (commented):
Look great, just some minor suggestions.

enhancements/20240805-encrypted-block-volume.md
Yu-Jack: I added one sentence in the end 

> Furthermore, any non-admin users in Rancher Integration mode can still use this storage class to create an encrypted volume.

bk201 (approved):

Yu-Jack:
Gentle ping @a110605, thanks!

Issue #6258, opened 2024-08-05
[BUG] Upgrade can be deleted in the "Upgrade Node" phase


bk201:
**Describe the bug**

Some users delete an upgrade when the upgrade enters the "Upgrade node" phase.
This makes debugging extremely hard if the user starts over another upgrade, we should prohibit the behavior.

**To Reproduce**
1. Upgrade a cluster
2. After "Upgrading system services" phase complete, delete the upgrade.
3. The upgrade is deleted.

**Expected behavior**

The upgrade should not be deleted.

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

The webhook implemented in https://github.com/harvester/harvester/blob/e269dfe35b0cb822e2c2801c435fed4d9a3d2523/pkg/webhook/resources/upgrade/validator.go#L387-L405 is not enough to prevent the delete, we need more checks. And add an annotation to bypass the check.

khushboo-rancher:
@noahgildersleeve Can we test this when we will do the upgrade testing during release testing.

noahgildersleeve:
Validated with 2 node libvirt cluster upgrading from v1.3.2 -> v1.4-c6facee4-head. I waited for first node to upgrade then tried to delete upgrade. Verified as fixed.

![Screenshot_20241030_162147](https://github.com/user-attachments/assets/799171cb-9eb0-44e7-9083-9f234d8d98d0)

Issue #6203, opened 2024-07-21
[BUG] Upgrade from 1.2.2 to 1.3.1 stuck


johnwc:
**Describe the bug**
We attempted to upgrade to v1.3.1, the first node upgraded but the rest did not. Following one of your other issues, we attempted to restart the upgrade following the directions to delete the existing upgrade and start it again. Only after we did this, did we notice the known issue about this very specific thing where the fix is to restart a capi pod. So, we are stuck now with only one node upgraded, harvester UI displaying v1.3.1, longhorn UI displaying v1.6.2, one node on v1.27.13+rke2r1 and the rest on v1.26.15+rke2r1, and one node keeps getting set as cordoned like it is trying to continue the upgrade. I did see that the node that is being cordoned is waiting for a machine plan to be applied if that helps any. I will be creating a support bundle after this issue creation and emailing it to [harvester-support-bundle@suse.com](mailto:harvester-support-bundle@suse.com)

bk201:
@FrankYang0529 Please help check, a quick scan of SB the cluster is wating:
```
    - lastTransitionTime: "2024-07-20T19:22:07Z"
      message: configuring etcd node(s) custom-29716d46e185,custom-5a9f02b8304b,custom-c3b245516405
      reason: Waiting
      status: Unknown
      type: ControlPlaneReady
```

There are 9 nodes but 11 CAPI machines in the cluster.

johnwc:
> @FrankYang0529 Please help check, a quick scan of SB the cluster is wating:
> 
> ```
>     - lastTransitionTime: "2024-07-20T19:22:07Z"
>       message: configuring etcd node(s) custom-29716d46e185,custom-5a9f02b8304b,custom-c3b245516405
>       reason: Waiting
>       status: Unknown
>       type: ControlPlaneReady
> ```
> 
> There are 9 nodes but 11 CAPI machines in the cluster.

@bk201  Sorry, I don't know what SB stands for. Do you need me to run a command and share back with you? What do I need to do?

DovydasNavickas:
SB is support bundle 😄

johnwc:
@FrankYang0529 Do you need me to run a command and share back with you? What do I need to do?

johnwc:
@FrankYang0529 I found the 2 orphaned machine configs, removed them and restarted the upgrade manually. It is now stuck at Upgrading System Service 50%. I sent the logs to the same email address.

[...]

PR #6274, opened 2024-08-06
fix: deny removing upgrade request if it is in upgrade node phase

FrankYang0529:

**Problem:**
If there is NodeUpgraded condition, it means provisioning cluster is updated. In this case, if users remove upgrade CR, the cluster may have inconsistent node version.

**Solution:**
Deny removing upgrade CR if NodeUpgraded condition is unknown.

**Related Issue:**
https://github.com/harvester/harvester/issues/6258

**Test plan:**
1. Create a v1.3.1 harvester cluster.
2. Upgrade to this branch.
3. When NodeUpgraded condition is unknown, remove the upgrade CR. The operation should be denied.

```
kubectl delete upgrade.harvesterhci.io hvst-upgrade-gk9t6 -n harvester-system
Error from server (BadRequest): admission webhook "validator.harvesterhci.io" denied the request: node upgrade is in progressing, please wait for it to be provisioned
```

### DIFF
--- a/pkg/webhook/resources/upgrade/validator.go
+++ b/pkg/webhook/resources/upgrade/validator.go
@@ -401,6 +401,12 @@ func (v *upgradeValidator) Delete(_ *types.Request, oldObj runtime.Object) error
 		return werror.NewBadRequest(fmt.Sprintf("cluster %s/%s status is provisioning, please wait for it to be provisioned", util.FleetLocalNamespaceName, util.LocalClusterName))
 	}
 
+	// If fleet-local/local cluster.provisioning.cattle.io is upgrading, deny removing upgrade CR request.
+	// If upgrade is removed, the cluster may have different RKE2 version nodes. It will make next upgrade fail.
+	if v1beta1.NodesUpgraded.IsUnknown(oldUpgrade) {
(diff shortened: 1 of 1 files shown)


bk201 (approved):

starbops (commented):
I have one question:

We seem to have already preventing the user from removing the Upgrade CR during the `local` cluster.provisioning object is under provisioning. What does this new check cover? Does it cover more cases than the existing check?

FrankYang0529:
> I have one question:
> 
> We seem to have already preventing the user from removing the Upgrade CR during the `local` cluster.provisioning object is under provisioning. What does this new check cover? Does it cover more cases than the existing check?

Hi @starbops, sorry for the late reply. In issue https://github.com/harvester/harvester/issues/6203, we observe that CAPI cluster is ready, but provisioning cluster is still upgrading nodes. So we can't only rely on CAPI cluster status.

bk201:
@mergifyio backport v1.4 v1.3

starbops (approved):
> Hi @starbops, sorry for the late reply. In issue https://github.com/harvester/harvester/issues/6203, we observe that CAPI cluster is ready, but provisioning cluster is still upgrading nodes. So we can't only rely on CAPI cluster status.

I see. LGTM, thank you!

Issue #6266, opened 2024-08-05
[BUG] Single-replica volume may break Harvester maintenance mode


w13915984028:
**Describe the bug**

Single-replica volume may break Harvester maintenance mode

**To Reproduce**
1. Create a single-replica storageclass
2. Create a VM which uses this single-replica storageclass to create volume
3. After VM is running, observe the scheduled node (the single-replica volume is scheduled to this node), stop the VM
4. Put the observed node into `maintenance mode`
5. There are continuously reported such error on Harvester pod, it fails to evict LH pod

>time="2024-08-05T09:51:06Z" level=info msg="evicting pod longhorn-system/instance-manager-68cd2514dd3f6d59b95cbd865d5b08f7"
time="2024-08-05T09:51:06Z" level=info msg="error when evicting pods/\"instance-manager-68cd2514dd3f6d59b95cbd865d5b08f7\" -n \"longhorn-system\" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget."

6. The node will not enter `maintenance mode` as expected.

The root cause is that the LH will check there is at least one another PDB alive before allowing killing the pdb for the drained node, but for single-replica volume, it makes no sense. The global setting also does not solve all situations.

new debug log
>time="2024-08-05T15:07:41Z" level=info msg="Node instance-manager-68cd2514dd3f6d59b95cbd865d5b08f7 is marked unschedulable but removing harv2 PDB is blocked: no pdb on another node, wait " func="controller.(*InstanceManagerController).syncInstanceManagerPDB" file="instance_manager_controller.go:730" controller=longhorn-instance-manager node=harv2

The global setting can't solve all scenarios.

```
longhorn-system   node-drain-policy                                                 block-if-contains-last-replica                47d
```

**Expected behavior**

Enter maintenance node as expected.

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

Observed on PR https://github.com/longhorn/longhorn-manager/pull/3045

related issues:
https://github.com/longhorn/longhorn/issues/9183
https://github.com/longhorn/longhorn/issues/5910#issuecomment-2269316423
https://github.com/harvester/harvester/issues/6264

w13915984028:
[...]

Issue #6264, opened 2024-08-05
[BUG] A manual attached LH volume breaks Harvester maintenance mode


w13915984028:
**Describe the bug**

A manual attached LH volume breaks Harvester maintenance mode

**To Reproduce**

1. Create a volume manually on Harveter UI and attache this volume to a node from harvester embedded LH UI
![image](https://github.com/user-attachments/assets/4d7c8e47-7015-4160-a5c6-67072ea6c5fc)

after attaching
![image](https://github.com/user-attachments/assets/3e061328-9700-4fc0-990d-e9943716ebe0)

2. Put this node into `maintenance mode`

3. The node may stuck on `Cordoned` status, and the menu, `Enable maintenance mode` is available again

![image](https://github.com/user-attachments/assets/260a53b1-a840-4c25-a79a-143fa1baff11)

4. There are continuously reported such error on Harvester pod, it fails to evict LH pod

```
time="2024-08-05T09:51:06Z" level=info msg="evicting pod longhorn-system/instance-manager-68cd2514dd3f6d59b95cbd865d5b08f7"
time="2024-08-05T09:51:06Z" level=info msg="error when evicting pods/\"instance-manager-68cd2514dd3f6d59b95cbd865d5b08f7\" -n \"longhorn-system\" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget."
```

the error was also reported on https://github.com/longhorn/longhorn/issues/5910

LH fails silently https://github.com/longhorn/longhorn/issues/9183

**Expected behavior**

1. Put into `enable maintenance mode` should report proper error/emit event to show the error
3. It should fail after certain time

**Support bundle**

**Environment**
 - Harvester ISO version: v1.4.0 master-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

LH CRD object also have evidence:

The LH UI attached volume:

```
- apiVersion: longhorn.io/v1beta2
  kind: VolumeAttachment
...
  spec:
    attachmentTickets:
      longhorn-ui:
        generation: 0
        id: longhorn-ui
        nodeID: s37570031
        parameters:
          disableFrontend: "false"
          lastAttachedBy: "null"
        type: longhorn-api
    volume: pvc-9b35136c-f59e-414b-aa55-b84b9b21ff89
```

The CSI attached volume:

```
  spec:
    attachmentTickets:
[...]

PR #6276, opened 2024-08-06
Upgrade: change LH default settings to fix maintenance mode issues

w13915984028:

**Problem:**

The Harvester maintenance mode is blocked by manually attached volumes and single-replica volumes.

**Solution:**

Change LH default setting to allow node drain on such scenarios.

This is the upgrade PR of PR788, the installer PR needs to be merged first:
- [ ] https://github.com/harvester/harvester-installer/pull/788

**Related Issue:**
https://github.com/harvester/harvester/issues/6264
https://github.com/harvester/harvester/issues/6266

**Test plan:**

Per issue steps.
1. Before ugprade, don't set `detach-manually-attached-volumes-when-cordoned` and `node-drain-policy ` on LH and Harvester managedchart,
After upgrade, get such LH default settings:

```
default-data-path                                                 /var/lib/harvester/defaultdisk                    true      7m19s
...
detach-manually-attached-volumes-when-cordoned                    true                                              true      7m19s
...
node-drain-policy                                                 allow-if-replica-is-stopped                       true      7m19s
```

2. Before upgrade, set `detach-manually-attached-volumes-when-cordoned` and `node-drain-policy ` on LH using none-default values, after upgrade, they keep same.

3. Before upgrade,  `detach-manually-attached-volumes-when-cordoned` and `node-drain-policy ` on harvester managed-chart using any valid values, after upgrade, they keep same.

The upgrade will only set the target option when they are not set via harvester managedchart and still using the default values on LH, any other cases will be kept unchanged.

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -438,6 +438,44 @@ EOF
   kubectl -n cattle-system patch ingress rancher-expose --patch-file ./rancher-expose.yaml --type=merge
 }
 
+patch_longhorn_settings() {
+  # set Longhorn default settings with Harvester expected values, only when they are not set in managedchart and LH uses the default value
+  local target=$1
+  local EXIT_CODE=0
+  # yq returns 'Error: no matches found' if an item is not found
+  yq -e '.spec.values.longhorn.defaultSettings.nodeDrainPolicy' $target || EXIT_CODE=$?
+  if [ $EXIT_CODE != 0 ]; then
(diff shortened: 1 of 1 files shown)


w13915984028:
@bk201 @tserong The installer part has been merged, please also review this upgrade related PR, thanks.

tserong (approved):

w13915984028:
@mergify backport v1.4

bk201 (approved):

w13915984028:
code is rebased to solve merge conflicts

Issue #6312, opened 2024-08-12
[BUG] Restore from backup to replace existing vm would stuck in `Restoring` with `WaitingForVolumeBinding` 


TachunLin:
**Describe the bug**

When restore a vm backup to replace the existing vm (retain volume), the restore kept in `Restoring` and won't complete.

* Restore from clean backup to replace existing vm 
  ![image](https://github.com/user-attachments/assets/8c2ac109-49e7-4739-8772-db846708106e)
  
  
  https://github.com/user-attachments/assets/d71b0e18-f76b-4fec-82fc-7196f4209f77

* Restore from backup already restore to new before to replace existing vm 

  https://github.com/user-attachments/assets/367b5227-2c44-44c7-ad7e-8218735e56ee

* The vm display status `WaitingForVolumeBinding`
  ![image](https://github.com/user-attachments/assets/6d10192c-188d-46af-a4d1-4280ba2a10be)

**To Reproduce**

1. Create a vm names opensuse
2. Setup nfs backup target
3. Create a backup named opensuse-backup
4. Shutdown the vm
5. Select restore to existing vm
6. Select to retain (or delete) volume
7. Check the restore process complete
8. Check the restore column will disappear after restore process finish

**Expected behavior**

Can restore vm to replace existing vm correctly,the vm should be started in running state

**Support bundle**

* Restore to replace existing which already restore to new one before
[supportbundle_3c2a0704-abe7-41eb-894b-041b497c5bfc_2024-08-12T09-56-19Z.zip](https://github.com/user-attachments/files/16580321/supportbundle_3c2a0704-abe7-41eb-894b-041b497c5bfc_2024-08-12T09-56-19Z.zip)

* Restore to replace existing vm 
[supportbundle_3c2a0704-abe7-41eb-894b-041b497c5bfc_2024-08-12T11-20-43Z.zip](https://github.com/user-attachments/files/16580868/supportbundle_3c2a0704-abe7-41eb-894b-041b497c5bfc_2024-08-12T11-20-43Z.zip)

**Environment**
 - Harvester ISO version: `v1.4-4fe69de-head`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Single node on local kvm machine

TachunLin:
This issue also exists on the latest master `master-97b50dc-head` 

https://github.com/user-attachments/assets/f3ecec66-f289-4eed-a7b3-db716c6ae6c7

WebberHuang1118:
[...]

PR #6338, opened 2024-08-14
Fix Restored VM Stuck for the Replacing case (#6132)

WebberHuang1118:
**Problem:**
Restore from backup to replace existing vm would stuck in Restoring with WaitingForVolumeBinding

**Solution:**
To replace the existing case, set `RunStratedy` as `Halted` first, then the Harvester controller will try to kick off the VMI

**Related Issue:**
#6312 

**Test plan:**
- Creating a VM
- Setting the backup target (not required for snapshot case)
- Creating a backup/snapshot for the VM
- Shutting down the VM
- Restore backup/snapshot to existing VM as deletion/retain policy
- The restored VM should be created successfully

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -547,6 +547,12 @@ func (h *RestoreHandler) reconcileVM(
 
 	vmCpy := vm.DeepCopy()
 	vmCpy.Spec = backup.Status.SourceSpec.Spec
+
+	//if the source runStratedy is RerunOnFailure, Kubevirt will not start the new VMI
+	//set the VM runStrategy as Halted, VMI will be kicked off in startVM()
+	haltedRunStrategy := kubevirtv1.RunStrategyHalted
+	vmCpy.Spec.RunStrategy = &haltedRunStrategy
+
 	vmCpy.Spec.Template.Spec.Volumes = newVolumes
 	if vmCpy.Annotations == nil {
 		vmCpy.Annotations = make(map[string]string)


bk201 (approved):
lgtm, thanks.

Yu-Jack:
A quick question. 

Just trace back to the main issue and commit https://github.com/kubevirt/kubevirt/commit/b234354de9545e7c106dbeaf23510573350a09f5. 

As I know, did we use this annotation `harvesterhci.io/vmRunStrategy: RerunOnFailure` to  keep original strategy because kubevirt changed it to Halted? Then, we did some workarounds https://github.com/search?q=repo%3Aharvester%2Fharvester%20AnnotationRunStrategy&type=code.

I'm thinking If current kubevirt (after v1.2.2) decide to keep it, maybe someday we can remove those workarounds?

WebberHuang1118:
> A quick question.
> 
> Just trace back to the main issue and commit [kubevirt/kubevirt@b234354](https://github.com/kubevirt/kubevirt/commit/b234354de9545e7c106dbeaf23510573350a09f5).
> 
> As I know, did we use this annotation `harvesterhci.io/vmRunStrategy: RerunOnFailure` to keep original strategy because kubevirt changed it to Halted? Then, we did some workarounds https://github.com/search?q=repo%3Aharvester%2Fharvester%20AnnotationRunStrategy&type=code.
> 
> I'm thinking If current kubevirt (after v1.2.2) decide to keep it, maybe someday we can remove those workarounds?

Hi @Yu-Jack 
Thanks for pointing this out,
The workaround is for Kubevirt issue https://github.com/kubevirt/kubevirt/issues/7295, which still exists,

If a VM's runStrategy goes through `RerunOnFailure` -> `Halted` and rest API `start`, the VM's runStrategy will change to `Always`, from the following code snippet:
- https://github.com/kubevirt/kubevirt/blob/1466b658f78b9b8bb9517ffb6dafd4b777f33fe6/pkg/virt-api/rest/subresource.go#L560-L562
- https://github.com/kubevirt/kubevirt/blob/1466b658f78b9b8bb9517ffb6dafd4b777f33fe6/pkg/virt-api/rest/subresource.go#L275-L285

IMHO, we still need to keep the workaround.

Yu-Jack (approved):
LGTM, thanks.

FrankYang0529 (approved):
LGTM. Thanks for the fix.

WebberHuang1118:
@Mergifyio backport v1.4

Issue #6283, opened 2024-08-07
[ENHANCEMENT] [v1.3.2] minor updates


w13915984028:
**Is your enhancement related to a problem? Please describe.**

minor updates for v1.3.2 release, including:

- [x] https://github.com/harvester/harvester-installer/pull/790
- [x] https://github.com/harvester/harvester-installer/pull/813
- [x] https://github.com/harvester/harvester/pull/6347

**Describe the solution you'd like**

The whole Harvester cluster will does not depends on shell v0.0.24 relase; the image is not packed into Harvester ISO; only v0.0.26.

The air-gapped installed should success; the normall installation should have no v0.0.24 image accidently pulled from docker hub to local.

The upgrade path to v1.3.2 should also not depend on shell v0.024.

**Describe alternatives you've considered**

**Additional context**

manual cherry pick of
issue: https://github.com/harvester/harvester/issues/6166
pr: https://github.com/harvester/harvester-installer/pull/766

w13915984028:
From @mallardduck
https://github.com/harvester/harvester-installer/pull/766#issuecomment-2261336583

> From my understanding Harvester considers the last 2 minors as the maintained versions; so currently 1.3 and 1.2 - but when 1.4.0 releases it will be 1.3 and 1.4. If this is correct I am wondering if we can backport (at least just the shell changes) to Harvester 1.3 as well for security benefits?
> 
> This same version of Shell can be used in both Rancher 2.7 and 2.8 versions, so if the Harvester 1.3 branch uses similar k8s/rancher versions, then this can be safe there too.
> 
> Additionally, if the version of shell that Harvester embedded Rancher uses could be updated too that would be good. Not sure if there's a mechanism on the installer to do that, but changing the shell version that rancher uses can be done via:
> 
> `kubectl patch settings.management.cattle.io shell-image --type merge -p '{"value":"rancher/shell:v0.1.26"}'`

bk201:
@w13915984028 is the issue ready for testing?

w13915984028:
@bk201 
Harvester installer shell image is bumped, merged and can be tested now;

Patch Rancher is not started yet.

w13915984028:
Latest update: refer: https://github.com/harvester/harvester-installer/pull/813

<del>
Test on `patch rancher setting`:

(1) Drop the `docker.io/rancher/shell:v0.1.24` image and don't pack it into Harvester ISO

[...]

PR #6347, opened 2024-08-14
Upgrade: patch rancher shell-image to v0.1.26

w13915984028:

**Problem:**

https://github.com/harvester/harvester/issues/6283

The rancher shell default image v0.1.24 has CVE issues.

**Solution:**

Patch Rancher shell version to v0.1.26

**Related Issue:**
https://github.com/harvester/harvester/issues/6283

Related installer PR: https://github.com/harvester/harvester-installer/pull/813

**Test plan:**

```
1. not set, patch
Rancher shell-image is  , will be patched to v0.1.26
setting.management.cattle.io/shell-image patched
NAME          VALUE
shell-image   rancher/shell:v0.1.26


2. lower version, patch
Rancher shell-image is rancher/shell:v0.1.24, will be patched to v0.1.26
setting.management.cattle.io/shell-image patched
NAME          VALUE
shell-image   rancher/shell:v0.1.26

3. updated, no patch
Rancher shell-image is rancher/shell:v0.1.26, patch is not needed
```

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -245,6 +245,16 @@ upgrade_rancher() {
     exit 1
   fi
 
+  # When upgrading to Harvester v1.3.2, patch Rancher shell-image to v0.1.26 if necessary
+  local shellimage=$(kubectl get settings.management.cattle.io shell-image -ojsonpath='{.value}')
+  if [[ "$shellimage" = "" ]] || [[ "$shellimage" < "rancher/shell:v0.1.26" ]]; then
+    echo "rancher shell-image is $shellimage, will be patched to v0.1.26"
+    kubectl patch settings.management.cattle.io shell-image --type merge -p '{"value":"rancher/shell:v0.1.26"}'
+    kubectl get settings.management.cattle.io shell-image
+  else
+    echo "rancher shell-image is $shellimage, patch is not needed"
+  fi
+
   if [ "$RANCHER_CURRENT_VERSION" = "$REPO_RANCHER_VERSION" ]; then
     echo "Skip update Rancher. The version is already $RANCHER_CURRENT_VERSION"
     return


package/upgrade/upgrade_manifests.sh
```diff
@@ -245,6 +245,16 @@ upgrade_rancher() {
     exit 1
   fi
 
+  # When upgrading to Harvester v1.3.2, patch Rancher shell-image to v0.1.26 if necessary
+  local shellimage=$(kubectl get settings.management.cattle.io shell-image -ojsonpath='{.value}')
+  if [[ "$shellimage" = "" ]] || [[ "$shellimage" < "rancher/shell:v0.1.26" ]]; then
```
starbops, line 250: Is it possible to meet the following scenario?

```
shellimage=rancher/shell:v0.1.3
if [[ "$shellimage" = "" ]] || [[ "$shellimage" < "rancher/shell:v0.1.26" ]]; then 
  echo need patch
fi
# will not patch
```

I'm not sure if we need to handle all the cases. If this is just a quick fix, assuming the before-upgrade will always use `rancher/shell:v0.1.24`, then we're good.

w13915984028: This is a specific patch only on Harvester v1.3.2, suppose it is upgraded from v1.3.0/v1.3.1/v1.3.2, the shell image is  like `v0.1.22`, `v0.1.24`. 

A far previous version like `v0.1.14` was release on 2021. https://github.com/rancher/shell/releases/tag/v0.1.14, no need to worry about `v0.1.3`; and there is also no version like `v0.1.31`, the `v0.1.26` is the currently last version of `v0.1.2*`.

Recently, Rancher release new major version of shell, it starts from v0.2.1.

bk201: Similar question to the [new cluster case](https://github.com/harvester/harvester-installer/pull/813#issuecomment-2290897219). If we modify the setting to `v0.1.26` here, when the Rancher is upgraded I guess the setting is still v0.1.26 and will need some clean-up.

starbops (approved):
Tested, LGTM!

Could we have another issue to track this and harvester/harvester-installer#813 for future patches/improvements?

package/upgrade/upgrade_manifests.sh
w13915984028: Log a new issue https://github.com/harvester/harvester/issues/6352 to drop this potential patch on v1.3.2 cluster.

bk201 (approved):

w13915984028:
This patch will be dropped in Harvester v1.4.0

PR: https://github.com/harvester/harvester/pull/6353
ISSUE:  https://github.com/harvester/harvester/issues/6352

PR #6348, opened 2024-08-15
feat: support boot and data volume encryption

Yu-Jack:
## **Problem:**
Longhorn has supported boot and data volume encryption/decryption, Harvester should follow up.

## **Solution:**
Support boot and data volume encryption/decryption. More details in [HEP](https://github.com/harvester/harvester/pull/6270).

## **Related Issue:**
https://github.com/harvester/harvester/issues/3129

### Boot Volume

1. Create Secret (could be any namespace)
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: longhorn-crypto 
  namespace: longhorn-system # any namespace is okay
stringData:
  CRYPTO_KEY_VALUE: "Some secret value here"
  CRYPTO_KEY_PROVIDER: "secret"
  CRYPTO_KEY_CIPHER: "aes-xts-plain64"
  CRYPTO_KEY_HASH: "sha256"
  CRYPTO_KEY_SIZE: "256"
  CRYPTO_PBKDF: "argon2i"
```
2. Create Storage Class
```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: global-encryption
allowVolumeExpansion: true
parameters:
  numberOfReplicas: "2"
  staleReplicaTimeout: "2880" # 48 hours in minutes
  fromBackup: ""
  migratable: "true"
  encrypted: "true"
  csi.storage.k8s.io/provisioner-secret-name: "longhorn-crypto"
  csi.storage.k8s.io/provisioner-secret-namespace: "longhorn-system"
  csi.storage.k8s.io/node-publish-secret-name: "longhorn-crypto"
  csi.storage.k8s.io/node-publish-secret-namespace: "longhorn-system"
  csi.storage.k8s.io/node-stage-secret-name: "longhorn-crypto"
  csi.storage.k8s.io/node-stage-secret-namespace: "longhorn-system"
```
3. Prepare a virtual machine image as source image (Please wait for finishing)
```yaml
kind: VirtualMachineImage
metadata:
  name: alpine-extended
  namespace: default
spec:
  retry: 3
  displayName: alpine-extended-3.20.2-x86_64.iso
  sourceType: download
  url: https://dl-cdn.alpinelinux.org/alpine/v3.20/releases/x86_64/alpine-extended-3.20.2-x86_64.iso
```
4. Create an encrypted virtual machine image and use it to create virtual machine.
```yaml
kind: VirtualMachineImage
metadata:
  annotations:
      harvesterhci.io/storageClassName: global-encryption # Be aware that this should be the same as the previous one.
  name: encrypted-alpine-extended
  namespace: default
spec:
  displayName: encrypted-alpine-extended-harvester-3.20.2-x86_64.iso
  retry: 3
  sourceType: clone
  securityParameters:
    cryptoOperation: "encrypt"
    sourceImageName: "alpine-extended"
    sourceImageNamespace: "default"
```
5. For decryption, create a decrypted virtual machine image.
```yaml
kind: VirtualMachineImage
metadata:
  name: decrypted-alpine-extended
  namespace: default
spec:
  displayName: decrypted-alpine-extended-harvester-3.20.2-x86_64.iso
  retry: 3
  sourceType: clone
  securityParameters:
    cryptoOperation: "decrypt"
    sourceImageName: "encrypted-alpine-extended"
    sourceImageNamespace: "default"
```

### Data Volume

We'll reuse previous storage class to create PV on GUI like this:

image

> If you're interested in whether it's really encrypted or not, please check "Test Plan" part of [HEP](https://github.com/harvester/harvester/pull/6270).


pkg/apis/harvesterhci.io/v1beta1/image.go
```diff
@@ -62,6 +62,18 @@ type VirtualMachineImageSpec struct {
 	// +kubebuilder:validation:Maximum:=10
 	// +kubebuilder:validation:Optional
 	Retry int `json:"retry" default:"3"`
+
+	// +optional
+	SecurityParameters *VirtualMachineImageSecurityParameters `json:"securityParameters,omitempty"`
+}
+
+type VirtualMachineImageSecurityParameters struct {
```
ibrokethecloud, line 70: should all the fields not be mandatory when a user devices to define securityParamters.

this will ensure an additional validation by the apiserver as well

Yu-Jack: But, those three parameters are required when users define securityParameters. I was going to prevent users from defining `securityParameters: {}` in spec. Are there any concerns about making these parameters mandatory?

```yaml
# skip
        # +optional
        SecurityParameters *VirtualMachineImageSecurityParameters `json:"securityParameters,omitempty"`
# skip

type VirtualMachineImageSecurityParameters struct {
        // +kubebuilder:validation:Required
        // +kubebuilder:validation:Enum=encrypt;decrypt;qwe
        CryptoOperation VirtualMachineImageCryptoOperationType `json:"cryptoOperation"`

        // +kubebuilder:validation:Required
        SourceImageName string `json:"sourceImageName"`

        // +kubebuilder:validation:Required
        SourceImageNamespace string `json:"sourceImageNamespace"`
}
```

ibrokethecloud (approved):
lgtm. thanks.

pkg/apis/harvesterhci.io/v1beta1/image.go
Yu-Jack: NOTE: After discussion with @ibrokethecloud, we decide to make those three as required fields, and use pointer to prevent users from inputing `securityParameters: {}`.

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -272,6 +273,30 @@ func (h *vmImageHandler) createBackingImage(image *harvesterv1.VirtualMachineIma
 		bi.Spec.SourceParameters[lhmanager.DataSourceTypeExportFromVolumeParameterExportType] = lhmanager.DataSourceTypeExportFromVolumeParameterExportTypeRAW
 	case harvesterv1.VirtualMachineImageSourceTypeRestore:
 		bi.Spec.SourceParameters[lhv1beta2.DataSourceTypeRestoreParameterBackupURL] = image.Spec.URL
+	case harvesterv1.VirtualMachineImageSourceTypeClone:
+		bi.Spec.SourceParameters[lhv1beta2.DataSourceTypeCloneParameterEncryption] = string(image.Spec.SecurityParameters.CryptoOperation)
+
+		sourceImage, err := h.images.Get(image.Spec.SecurityParameters.SourceImageNamespace, image.Spec.SecurityParameters.SourceImageName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get source vmimage %s/%s, error: %s", image.Spec.SecurityParameters.SourceImageName, image.Spec.SecurityParameters.SourceImageNamespace, err.Error())
```
bk201:
```suggestion
			return fmt.Errorf("failed to get source vmimage %s/%s, error: %s", image.Spec.SecurityParameters.SourceImageNamespace, image.Spec.SecurityParameters.SourceImageName, err.Error())
```

Nit: just to be consistent with other log messages.

pkg/webhook/indexeres/indexer.go
```diff
@@ -114,3 +119,11 @@ func scheduleVMBackupByCronGranularity(obj *harvesterv1.ScheduleVMBackup) ([]str
 
 	return []string{granularity.String()}, nil
 }
+
+func imageByStorageClass(obj *harvesterv1.VirtualMachineImage) ([]string, error) {
+	sc, ok := obj.Annotations[util.AnnotationStorageClassName]
```
bk201, line 124: I'm not sure if the Annotations can be nil, if yes, I suggest checking if the Annotations is nil.

pkg/util/indexeres/storageclass.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/harvester/harvester/pkg/ref"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	StorageClassBySecretIndex = "harvesterhci.io/storageclass-by-secret"
+)
+
+func StorageClassBySecret(sc *storagev1.StorageClass) ([]string, error) {
+	encrypted, ok := sc.Parameters[util.LonghornOptionEncrypted]
```
bk201, line 15: will sc.Parameters be nil?

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -103,6 +114,78 @@ func (v *virtualMachineImageValidator) CheckImageDisplayNameAndURL(newImage *v1b
...
+	}
+
+	// For decryption, we don't need to check the storage class.
+	// We could use sourceImage to track back to get the storage class.
+	if newImage.Spec.SecurityParameters.CryptoOperation == v1beta1.VirtualMachineImageCryptoOperationTypeDecrypt {
+		return nil
+	}
+
+	// Check if storage class exits
+	scName, ok := newImage.Annotations[util.AnnotationStorageClassName]
```
bk201, line 170: check Annotations nil

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -103,6 +114,78 @@ func (v *virtualMachineImageValidator) CheckImageDisplayNameAndURL(newImage *v1b
...
+func (v *virtualMachineImageValidator) checkImageSecurityParameters(newImage *v1beta1.VirtualMachineImage) error {
+	if newImage.Spec.SourceType != v1beta1.VirtualMachineImageSourceTypeClone {
+		return nil
+	}
+
+	if newImage.Spec.SecurityParameters == nil {
+		return werror.NewInvalidError(`sourceParameters is required when image source type is "clone"`, "spec.sourceParameters")
+	}
+
+	if newImage.Spec.SecurityParameters.CryptoOperation == "" {
```
bk201: I am just curious, do we need to validate the fields on our own if the CRD defines these fields required?

bk201 (commented):
Thanks, some questions in the comment.

pkg/webhook/indexeres/indexer.go
Yu-Jack: We don't need to.

In Go, we can read a nil map, but can't write any value into nil map. 

reading nil map https://go.dev/play/p/AhYjSZ7DpQi
writing nil map https://go.dev/play/p/7TSdpOwxCvT

In our cases, we just need reading, so we can skip checking that.

pkg/util/indexeres/storageclass.go
Yu-Jack: same as above

pkg/webhook/resources/virtualmachineimage/validator.go
Yu-Jack: same as above

pkg/webhook/resources/virtualmachineimage/validator.go
Yu-Jack: It's a very good question. In my original code, I remove all empty checking. But, when I'm writing the unit test, I found the CRD check doesn't work (obviously). So, I added them back.

~~After testing order of checking, the admission webhook will check the fields first. Next CRD will check them. So, guarding some request in admission webhook might be a choice?~~ CRD is first, then webhook.

bk201 (approved):
LGTM, thanks!

pkg/webhook/indexeres/indexer.go
bk201: Sounds great, thanks for the check.

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -106,3 +133,69 @@ func (v *storageClassValidator) validateSetUniqueDefault(newObj runtime.Object)
...
+			return werror.NewInvalidError(fmt.Sprintf("storage class must contain %s", missingParam), "spec.parameters")
+		}
+
+		if secretName == "" && secretNamespace == "" {
+			secretName = name
+			secretNamespace = namespace
+		} else if secretName != name || secretNamespace != namespace {
+			return werror.NewInvalidError(fmt.Sprintf("secret names and namespaces in %s and %s are different from others", pair[0], pair[1]), "")
+		}
+	}
```
WebberHuang1118: how about this to get rid of unnecessary else if
```suggestion
		if secretName == "" && secretNamespace == "" {
			secretName = name
			secretNamespace = namespace
			continue
		}	
		if secretName != name || secretNamespace != namespace {
			return werror.NewInvalidError(fmt.Sprintf("secret names and namespaces in %s and %s are different from others", pair[0], pair[1]), "")
		}
	}
```

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -175,7 +260,11 @@ func (v *virtualMachineImageValidator) Update(_ *types.Request, oldObj runtime.O
 		return werror.NewInvalidError("url cannot be modified", "spec.url")
 	}
 
-	return v.CheckImageDisplayNameAndURL(newImage)
+	if err := v.CheckImageDisplayNameAndURL(newImage); err != nil {
+		return err
+	}
+
+	return v.checkImageSecurityParameters(newImage)
```
WebberHuang1118: Maybe we just need to check the security parameters can't be changed?

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -106,3 +133,69 @@ func (v *storageClassValidator) validateSetUniqueDefault(newObj runtime.Object)
...
+		name, nameExists := newSC.Parameters[pair[0]]
+		namespace, namespaceExists := newSC.Parameters[pair[1]]
+
+		if !nameExists || !namespaceExists {
+			missingParam := pair[0]
+			if !namespaceExists {
+				missingParam = pair[1]
+			}
+			return werror.NewInvalidError(fmt.Sprintf("storage class must contain %s", missingParam), "spec.parameters")
+		}
```
WebberHuang1118: Should we join the missing parameters?

WebberHuang1118 (approved):
LGTM, thanks for the PR, just some nits.

https://github.com/Yu-Jack/harvester/blob/b904fedebf2b52786c1b2fe4fabb7375e14ad706/pkg/controller/master/image/vm_image_controller.go#L314-L328

For decrypt image, it will use parameters from default storage class, maybe we can add some related comments

pkg/webhook/resources/storageclass/validator.go
Yu-Jack: what do you mean join?

Yu-Jack: image

Issue #6272, opened 2024-08-06
[BUG] UpgradeLog can't be ready cause of fluentd PVC doesn't give access mode


FrankYang0529:
**Describe the bug**
When testing upgrade with log enabled on master branch, the upgradelog can't be ready cause of fluentd statefulset is not ready. The reason is that PVC spec doesn't have access mode, so PVC can't be created.

hvst-upgrade-pgdtp-upgradelog-infra-fluentd event:
```
  Warning  FailedCreate  2m44s (x18 over 13m)  statefulset-controller  create Claim hvst-upgrade-pgdtp-upgradelog-infra-hvst-upgrade-pgdtp-upgradelog-log-archive-hvst-upgrade-pgdtp-upgradelog-infra-fluentd-0 for Pod hvst-upgrade-pgdtp-upgradelog-infra-fluentd-0 in StatefulSet hvst-upgrade-pgdtp-upgradelog-infra-fluentd failed error: PersistentVolumeClaim "hvst-upgrade-pgdtp-upgradelog-infra-hvst-upgrade-pgdtp-upgradelog-log-archive-hvst-upgrade-pgdtp-upgradelog-infra-fluentd-0" is invalid: [spec.accessModes: Required value: at least 1 access mode is required, spec.resources[storage]: Required value]
```

hvst-upgrade-pgdtp-upgradelog-infra-fluentd statefulset spec:
```yaml
  volumeClaimTemplates:
  - apiVersion: v1
    kind: PersistentVolumeClaim
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/component: fluentd
        app.kubernetes.io/managed-by: hvst-upgrade-pgdtp-upgradelog-infra
        app.kubernetes.io/name: fluentd
        harvesterhci.io/upgradeLog: hvst-upgrade-pgdtp-upgradelog
        harvesterhci.io/upgradeLogComponent: aggregator
      name: hvst-upgrade-pgdtp-upgradelog-infra-hvst-upgrade-pgdtp-upgradelog-log-archive
      namespace: harvester-system
      ownerReferences:
      - apiVersion: logging.banzaicloud.io/v1beta1
        controller: true
        kind: Logging
        name: hvst-upgrade-pgdtp-upgradelog-infra
        uid: 6d7089d5-3bf4-448a-be13-3630fd556a10
    spec:
      resources: {}
      volumeMode: Filesystem
    status:
      phase: Pending
```

**To Reproduce**
[...]

PR #6349, opened 2024-08-15
fix(upgradelog): adapt to new extravolume mechanics

starbops:

**Problem:**

The UpgradeLog feature is broken in the master build. The `fluentd` StatefulSet-populated PVC is malformed; therefore the StatefulSet could not start normally. The entire Harvester upgrade is jammed.

**Solution:**

As for logging-operator v4.4.0, the way ExtraVolume is handled is different. It's impossible to make the fluentd Pod associate with a pre-created PVC. The PVC has now been created along with the fluentd StatefulSet. The UpgradeLog mechanics adapt the new upstream behavior to reconcile the StatefulSet populated PVC instead of creating a new one.

An UpgradeLog OwnerReference will be added to the PVC to make it live "longer" than its original owner/creator, i.e., the Logging object.

**Related Issue:**

#6272 

**Test plan:**

Test the following two upgrade scenario (both with the upgradelog feature enabled):
- Upgrade a Harvester cluster from v1.3.1 to master (to make sure we don't break anything)
- Upgrade a Harvester cluster from master to master (to make sure the fix is effective)

For both scenario, please try to download the upgradelog at the following time point:
- During the upgrade
- After the upgrade success (or failed, either should work)

### DIFF
--- a/pkg/controller/master/upgradelog/common.go
+++ b/pkg/controller/master/upgradelog/common.go
@@ -2,6 +2,7 @@ package upgradelog
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/cisco-open/operator-tools/pkg/volume"
 	loggingv1 "github.com/kube-logging/logging-operator/pkg/sdk/logging/api/v1beta1"
@@ -15,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
@@ -32,35 +34,6 @@ func upgradeLogReference(upgradeLog *harvesterv1.UpgradeLog) metav1.OwnerReferen
 		Kind:       upgradeLog.Kind,
 		UID:        upgradeLog.UID,
 		APIVersion: upgradeLog.APIVersion,
-	}
-}
-
-func preparePvc(upgradeLog *harvesterv1.UpgradeLog) *corev1.PersistentVolumeClaim {
-	volumeMode := corev1.PersistentVolumeFilesystem
-
-	return &corev1.PersistentVolumeClaim{
-		ObjectMeta: metav1.ObjectMeta{
-			Labels: map[string]string{
-				util.LabelUpgradeLog:          upgradeLog.Name,
-				util.LabelUpgradeLogComponent: util.UpgradeLogArchiveComponent,
-			},
-			Name:      name.SafeConcatName(upgradeLog.Name, util.UpgradeLogArchiveComponent),
-			Namespace: util.HarvesterSystemNamespaceName,
-			OwnerReferences: []metav1.OwnerReference{
-				upgradeLogReference(upgradeLog),
-			},
-		},
-		Spec: corev1.PersistentVolumeClaimSpec{
-			AccessModes: []corev1.PersistentVolumeAccessMode{
-				corev1.ReadWriteOnce,
-			},
-			Resources: corev1.VolumeResourceRequirements{
-				Requests: corev1.ResourceList{
-					"storage": resource.MustParse(defaultLogArchiveVolumeSize),
-				},
-			},
-			VolumeMode: &volumeMode,
-		},
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/controller/master/upgradelog/common.go
```diff
@@ -416,6 +401,23 @@ func prepareLogDownloaderSvc(upgradeLog *harvesterv1.UpgradeLog) *corev1.Service
 	}
 }
 
+// Returns the name of the log-archive PVC, which is created by the fluentd StatefulSet.
+//
+//	The name will look like: <upgradelog>-infra-log-archive-<fluentd statefulset>-0
```
FrankYang0529: Is `<fluentd statefulset>-0` equal to `<upgradelog>-infra-fluentd-0`?

starbops: The name is kind of a mess. Yes you're right. Let me update it to match the code.

FrankYang0529 (approved):
LGTM. All test cases can work. Thanks.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -581,6 +587,54 @@ func (h *handler) OnStatefulSetChange(_ string, statefulSet *appsv1.StatefulSet)
...
+		APIVersion: upgradeLog.APIVersion,
+		UID:        upgradeLog.UID,
+		Kind:       upgradeLog.Kind,
+	}}
+
+	// Add UpgradeLog as an owner of the log-archive PVC because we want it to
+	// live longer than its original owner, i.e., Logging, so pods like the
+	// downloader and packager can still access the log-archive volume.
+	toUpdate := pvc.DeepCopy()
+	if len(pvc.OwnerReferences) == 0 {
```
w13915984028: can we avoid multi-times appending?

pkg/controller/master/upgradelog/controller.go
```diff
@@ -581,6 +587,54 @@ func (h *handler) OnStatefulSetChange(_ string, statefulSet *appsv1.StatefulSet)
...
+	if !logArchiveMatchingLabels.AsSelector().Matches(pvcLabels) {
+		return pvc, nil
+	}
+	upgradeLogName, ok := pvc.Labels[util.LabelUpgradeLog]
+	if !ok {
+		return pvc, nil
+	}
+
+	upgradeLog, err := h.upgradeLogCache.Get(util.HarvesterSystemNamespaceName, upgradeLogName)
+	if err != nil {
```
w13915984028, line 607: shall we skip the none-existing error?

if the pvc accidently has such a label

w13915984028 (approved):
Thanks for the fix.

pkg/controller/master/upgradelog/controller.go
starbops: Good point. We can log down the incident and skip it.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -581,6 +588,64 @@ func (h *handler) OnStatefulSetChange(_ string, statefulSet *appsv1.StatefulSet)
...
+		if errors.IsNotFound(err) {
```
w13915984028: `if errors.IsNotFound(err)`
->
`if !errors.IsNotFound(err)`

pkg/controller/master/upgradelog/controller.go
```diff
@@ -581,6 +588,64 @@ func (h *handler) OnStatefulSetChange(_ string, statefulSet *appsv1.StatefulSet)
...
+	if !ok {
+		return pvc, nil
+	}
+
+	upgradeLog, err := h.upgradeLogCache.Get(util.HarvesterSystemNamespaceName, upgradeLogName)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return pvc, err
+		}
+		logrus.WithFields(logrus.Fields{
```
bk201, line 611: nit: add a kind to indicate resource in the log message.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -581,6 +588,64 @@ func (h *handler) OnStatefulSetChange(_ string, statefulSet *appsv1.StatefulSet)
...
+	}
+
+	// Add UpgradeLog as an owner of the log-archive PVC because we want it to
+	// live longer than its original owner, i.e., Logging, so pods like the
+	// downloader and packager can still access the log-archive volume.
+	toUpdate := pvc.DeepCopy()
+	toUpdate.OwnerReferences = append(toUpdate.OwnerReferences, newOwnerRef)
+
+	if !reflect.DeepEqual(pvc, toUpdate) {
+		logrus.WithFields(logrus.Fields{
```
bk201, line 640: ditto. to add kind.

bk201 (commented):
LGTM. Just a minor suggestion to the log message.

bk201 (approved):

Issue #6207, opened 2024-07-22
[BUG] Deleting virtual machine restore CR makes restoring VM can't restart


WebberHuang1118:
**Describe the bug**
Directly deleting `virtualmachinerestore` CR makes resotring VM can't restart

**To Reproduce**
1. Creating backup for a VM
2. Restoring backup to a new VM
3. Deleting the related `virtualmachinerestore` CR with `kubectl`
4. Restarting the restored VM
5. The restarting will be rejected with `Error
start virtual machine <restored VM> failed, persistentvolumeclaims "<restored PVC>" not found`

**Environment**
 - Harvester ISO version: master-head (`8b80f821f9da22a678ecef018a01d0fe241715a4`)

**Additional context**
The restoring VM's pvc has `ownerReferences` points to the matched CR `virtualmahcinerestore`, for example:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  ....
  name: restore-vm-b1-434fd2c1-fd7a-4c0f-bda7-9ac31becef13-disk-0
  namespace: default
  ownerReferences:
  - apiVersion: harvesterhci.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: VirtualMachineRestore
    name: restore-vm-b1-5xcnx
    uid: 434fd2c1-fd7a-4c0f-bda7-9ac31becef13
  resourceVersion: "100298"
  uid: c7db72ed-d9c9-4fee-a424-14a3cefbb0bb
spec:
  accessModes:
  - ReadWriteMany
  ....
  storageClassName: longhorn-image-l8mlc
  volumeMode: Block
  volumeName: pvc-c7db72ed-d9c9-4fee-a424-14a3cefbb0bb
```

so once the `virtualmachinerestore` CR is deleted, the PVC will turn into `Terminating` status. If the restoring VM is stopped, the PVC will be completely deleted.

bk201:
set priority to 2 because the `virtualmachinerestore` resources are not available in the GUI.

brandboat:
Test OK on [v1.4.0-dev-20240918](https://github.com/harvester/harvester/releases/tag/v1.4.0-dev-20240918), close as fixed.

## Environment
* [`v1.4.0-dev-20240918`](https://github.com/harvester/harvester/releases/tag/v1.4.0-dev-20240918)
  * Profile: 2 nodes QEMU/KVM (8C/16G/250G per node)
  * ui-source: `Auto`

## Test Steps
[...]

Issue #6194, opened 2024-07-18
[Question] How can I clear "Completed" status after VM restore?


dawid10353:
Hello,

how can I clear "Completed" status after VM restore?
Harvester v1.3.1

Screenshot:
![image](https://github.com/user-attachments/assets/aac0a5c0-f167-444e-b6fa-cd62bd86565b)

devenkulkarni:
You may need to delete the completed restore from the harvester k8s cluster. We have identified it as a bug and as a workaround for now, you can delete the completed restore.

```
kubectl get virtualmachinerestores.harvesterhci -A
NAMESPACE   NAME                               TARGET_KIND      TARGET_NAME   COMPLETE   AGE     ERROR
default     restore-devendra-s3-backup-z5h5j   VirtualMachine   restored-vm   true       5m26s  
```

```
kubectl delete virtualmachinerestores.harvesterhci restore-devendra-s3-backup-z5h5 -n default
```

Note: You would need to change the name of restore as per your cluster and it shall help remove the `Restore` column from the UI.

dawid10353:
> virtualmachinerestores.harvesterhci

But.. When I deleted "completed restore" using that commad:
![image](https://github.com/user-attachments/assets/09cb540a-9636-4b35-bc5e-3a5a2d1909f4)
The PVC for VM is deleted too...
![image](https://github.com/user-attachments/assets/33a89d81-b37f-4816-b985-07bc908cdc79)

dawid10353:
It forces me to restore from backup again... And again I have "Completed" status.

WebberHuang1118:
Hi @dawid10353 

The restoring VM's pvc has `ownerReferences` points to the matched CR `virtualmahcinerestore`, for example:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  ....
  name: restore-vm-b1-434fd2c1-fd7a-4c0f-bda7-9ac31becef13-disk-0
  namespace: default
  ownerReferences:
  - apiVersion: harvesterhci.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: VirtualMachineRestore
    name: restore-vm-b1-5xcnx
    uid: 434fd2c1-fd7a-4c0f-bda7-9ac31becef13
  resourceVersion: "100298"
  uid: c7db72ed-d9c9-4fee-a424-14a3cefbb0bb
spec:
  accessModes:
  - ReadWriteMany
  ....
[...]

PR #6351, opened 2024-08-15
Reject VMRestore CR Deletion if Restored VM Existing (#6207)

WebberHuang1118:
**Problem:**
Removing VMRestore CR will make the restored VM's volume be deleted

**Solution:**
Reject VMRestore cr deletion if restored VM exists

**Related Issue:**
#6207

**Test plan:**
1. Creating a snapshot for a VM
2. Restoring the snapshot as a new VM
3. Deleting the related `virtualmachinerestore` CR with `kubectl`
4. The deletion will be rejected with the message `VM <VM name> is still in use`

### DIFF
--- a/pkg/webhook/resources/virtualmachinerestore/validator.go
+++ b/pkg/webhook/resources/virtualmachinerestore/validator.go
@@ -79,6 +79,7 @@ func (v *restoreValidator) Resource() types.Resource {
 		OperationTypes: []admissionregv1.OperationType{
 			admissionregv1.Create,
 			admissionregv1.Update,
+			admissionregv1.Delete,
 		},
 	}
 }
@@ -303,5 +304,23 @@ func (v *restoreValidator) checkVolumeSnapshotClass(vmBackup *v1beta1.VirtualMac
 			return fmt.Errorf("can't get volumeSnapshotClass %s for driver %s", volumeSnapshotClassName, csiDriverName)
 		}
 	}
+	return nil
+}
+
+func (v *restoreValidator) Delete(_ *types.Request, newObj runtime.Object) error {
+	vmRestore := newObj.(*v1beta1.VirtualMachineRestore)
+	vm, err := v.vms.Get(vmRestore.Namespace, vmRestore.Spec.Target.Name)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+
+	if err != nil {
+		return werror.NewInvalidError(err.Error(), fieldTargetName)
+	}
+
+	if vm.DeletionTimestamp == nil {
+		return werror.NewInvalidError(fmt.Sprintf("The restore can't be removed because the restored VM %s exists", vm.Name), fieldTargetName)
+	}
+
 	return nil
 }


FrankYang0529:
Hi @WebberHuang1118, thanks for the PR. If there is failed VMRestore, we would like cleanup related PVCs by clean up VMRestore. That is why we set owner reference as VMRestore. However, after VMRestore is finished, I think we should change owner reference to the VM, because VMRestore is just a process to create PVC. After the process is done, its lifecycle is finished.

If we change it, we can also allow users to remove VMRestore to cleanup progress bar on GUI. https://github.com/harvester/harvester/issues/6194

WDYT?

WebberHuang1118:
> Hi @WebberHuang1118, thanks for the PR. If there is failed VMRestore, we would like cleanup related PVCs by clean up VMRestore. That is why we set owner reference as VMRestore. However, after VMRestore is finished, I think we should change owner reference to the VM, because VMRestore is just a process to create PVC. After the process is done, its lifecycle is finished.
> 
> If we change it, we can also allow users to remove VMRestore to cleanup progress bar on GUI. #6194
> 
> WDYT?

Hi @FrankYang0529,

Thanks for the feedback, I'd like to double confirm do you mean set PVC's owner reference to VM as the VM restoring complete? Then the VMRestore CR is free to be removed.

WebberHuang1118:
> > Hi @WebberHuang1118, thanks for the PR. If there is failed VMRestore, we would like cleanup related PVCs by clean up VMRestore. That is why we set owner reference as VMRestore. However, after VMRestore is finished, I think we should change owner reference to the VM, because VMRestore is just a process to create PVC. After the process is done, its lifecycle is finished.
> > If we change it, we can also allow users to remove VMRestore to cleanup progress bar on GUI. #6194
> > WDYT?
> 

 Hi @FrankYang0529,
 
 Thanks for the feedback, I'd like to double confirm do you mean set PVC's owner reference to VM as the VM restoring complete? Then the VMRestore CR is free to be removed.

FrankYang0529:
> I'd like to double confirm do you mean set PVC's owner reference to VM as the VM restoring complete? Then the VMRestore CR is free to be removed.

Yes, that is what I want to say.

WebberHuang1118:
> > I'd like to double confirm do you mean set PVC's owner reference to VM as the VM restoring complete? Then the VMRestore CR is free to be removed.
> 
> Yes, that is what I want to say.

Hi @FrankYang0529,
After investigation, I found there are many resources' owner reference points `VMRestore`

PVC: https://github.com/harvester/harvester/blob/c5d790422fbf02fecbd7401113183cdf95bc8cfc/pkg/controller/master/backup/restore.go#L766-L794

VolumeSnapahot:
https://github.com/harvester/harvester/blob/c5d790422fbf02fecbd7401113183cdf95bc8cfc/pkg/controller/master/backup/restore.go#L867-L887

VolumeSnapshotContent:
https://github.com/harvester/harvester/blob/c5d790422fbf02fecbd7401113183cdf95bc8cfc/pkg/controller/master/backup/restore.go#L819-L845

Making the above CRs' get out of VMRestore is likely out of the scope of this ticket, so I'd like to keep the current solution.

How about creating another ticket for refactoring the `Owner Reference points to VMRestore`? WDYT

FrankYang0529:
> How about creating another ticket for refactoring the `Owner Reference points to VMRestore`? WDYT

Yeah, sounds good to me. Thank you.

FrankYang0529 (approved):

WebberHuang1118:
> > How about creating another ticket for refactoring the `Owner Reference points to VMRestore`? WDYT
> 
> Yeah, sounds good to me. Thank you.

Future enhancement ticket https://github.com/harvester/harvester/issues/6408

bk201 (approved):
Thanks!

WebberHuang1118:
@Mergifyio backport v1.4 v1.3

pkg/webhook/resources/virtualmachinerestore/validator.go
```diff
@@ -305,3 +306,21 @@ func (v *restoreValidator) checkVolumeSnapshotClass(vmBackup *v1beta1.VirtualMac
 	}
 	return nil
 }
+
+func (v *restoreValidator) Delete(_ *types.Request, newObj runtime.Object) error {
+	vmRestore := newObj.(*v1beta1.VirtualMachineRestore)
+	vm, err := v.vms.Get(vmRestore.Namespace, vmRestore.Spec.Target.Name)
```
w13915984028, line 312: @WebberHuang1118 

A question: if one un-restored VM happens to have the same ns & name, what will happen?

are there any other evidence like label/annotation/id... for further check? thanks.

WebberHuang1118: Hi @w13915984028,

> A question: if one un-restored VM happens to have the same ns & name, what will happen?

This is not going to happen for an existing un-restored VM that has the same ns/name as the restored VM.

If the restoring process starts first, the un-restored VM creation will be ejected, and vice versa.

w13915984028: @WebberHuang1118 Thanks. I have following question :)

Assume user cluster is v1.3.1:
(1) VM A was restored from VM restore.
(2) VM A was deleted later.  ( will the VM Restore CRD be deleted as well? when true, then it is good)
(3) A new VM A was created.

Then user upgrades to v1.4.1
(1) User try to remove the VM restore CRD object now.

In short, for many long-running and upgraded clusters, do we still have enough evidence about the VM and VM Restore relationship just by ns&name?

WebberHuang1118: > @WebberHuang1118 Thanks. I have following question :)
> 
> Assume user cluster is v1.3.1: (1) VM A was restored from VM restore. (2) VM A was deleted later. ( will the VM Restore CRD be deleted as well? when true, then it is good) (3) A new VM A was created.
> 
> Then user upgrades to v1.4.1 (1) User try to remove the VM restore CRD object now.
> 
> In short, for many long-running and upgraded clusters, do we still have enough evidence about the VM and VM Restore relationship just by ns&name?

Hi @w13915984028 

During step  `(2) VM A was deleted later. ( will the VM Restore CRD be deleted as well? when true, then it is good)`, the VMRestore CR will be deleted (as your expectation), since we have owner reference on it (which points to the restored VM).

w13915984028: Thanks for your kind reply.

Issue #6352, opened 2024-08-15
[ENHANCEMENT] Drop tho potential manual patch of Rancher shell image v0.1.26 on Harvester v1.3.2


w13915984028:
**Is your enhancement related to a problem? Please describe.**

Harvester v1.4.0 will use Rancher v2.9 version, and the later uses a new shell image `defaultShellVersion: rancher/shell:v0.2.1`

https://github.com/rancher/rancher/blob/3f7c25803c00e5311d89d1144b3c6684b288ed51/pkg/settings/setting.go#L128C78-L128C97
https://github.com/rancher/rancher/blob/a9946d267616beb1be27c3142a4ebaf6186ba3e7/build.yaml#L4

From issue https://github.com/harvester/harvester/issues/6352, the `shell-image` was patched to `v0.1.26` on Harvester v1.3.2 upgrade path, that may affect the further upgrade to v1.4.0.

**Describe the solution you'd like**

Drop to manual patch on v1.4.0 upgrade path

**Describe alternatives you've considered**

**Additional context**

Previous upgrade PR: https://github.com/harvester/harvester/pull/6347

w13915984028:
Test plan:
(1) Upgrade to v1.3.2, check below CRD object, notice the VALUE column has value like `v0.1.26`
(2) Further upgrade to v1.4.0, the VALUE column is empty

```
v132->v140 upgrade, Rancher should upgrade smoothly, and the value of below CRD object should be empty after upgrade

setting.management.cattle.io/shell-image
NAME          VALUE
shell-image   
```

brandboat:
Verified this enhancement has been applied.

Test Information
----
* Environment: **qemu/KVM 1 node**
* ui-source Option: **Auto**

### Test Plan
basically follow the test plan in https://github.com/harvester/harvester/issues/6352#issuecomment-2326958381
- Harvester Version: **v1.4.0-rc2** , `shell-image` is empty
  ```txt
  harvester-node-0:~ # kubectl get settings.management.cattle.io shell-image
  NAME          VALUE
  shell-image
  ```
- Harvester Version: **v1.3.2**, `shell-image` is v0.1.26
  ```txt
  harvester-node-0:~ # kubectl get settings.management.cattle.io shell-image
  NAME          VALUE
  shell-image   rancher/shell:v0.1.26
  ```
- Harvester Version: upgrade from **v1.3.2** to **v1.4.0-rc2**, `shell-image` is empty
  ```txt
[...]

PR #6353, opened 2024-08-15
Upgrade: drop the potential manual patch upon rancher shell-image

w13915984028:
Upgrade: drop the potential manual patch upon rancher shell-image to v0.1.26 on Harvester v1.3.2

**Problem:**

Harvester v1.4.0 will use Rancher v2.9 version, and the later uses a new shell image `defaultShellVersion: rancher/shell:v0.2.1`

https://github.com/rancher/rancher/blob/3f7c25803c00e5311d89d1144b3c6684b288ed51/pkg/settings/setting.go#L128C78-L128C97
https://github.com/rancher/rancher/blob/a9946d267616beb1be27c3142a4ebaf6186ba3e7/build.yaml#L4

From issue https://github.com/harvester/harvester/issues/6352, the `shell-image` was patched to `v0.1.26` on Harvester v1.3.2 upgrade path, that may affect the further upgrade to v1.4.0.

**Solution:**

Upgrade: drop the potential manual patch upon rancher shell-image

**Related Issue:**
https://github.com/harvester/harvester/issues/6352
v1.3.2 installer PR https://github.com/harvester/harvester-installer/pull/813
v1.3.2 upgrade PR https://github.com/harvester/harvester/pull/6347
v1.3.2 issue https://github.com/harvester/harvester/issues/6283

**Test plan:**

v132->v140 upgrade, Rancher should upgrade smoothly, and the value of below CRD object should be empty after upgrade
```
setting.management.cattle.io/shell-image patched
NAME          VALUE
shell-image   
```

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -245,6 +245,16 @@ upgrade_rancher() {
     exit 1
   fi
 
+  # drop the potential manual patch upon shell-image to v0.1.26 on Harvester v1.3.2
+  local shellimage=$(kubectl get settings.management.cattle.io shell-image -ojsonpath='{.value}')
+  if [[ "$shellimage" = "rancher/shell:v0.1.26" ]]; then
+    echo "rancher shell-image is $shellimage, will be reverted to empty"
+    kubectl patch settings.management.cattle.io shell-image --type merge -p '{"value":""}'
+    kubectl get settings.management.cattle.io shell-image
+  else
+    echo "rancher shell-image is $shellimage, patch is not needed"
+  fi
+
   if [ "$RANCHER_CURRENT_VERSION" = "$REPO_RANCHER_VERSION" ]; then
     echo "Skip update Rancher. The version is already $RANCHER_CURRENT_VERSION"
     return


w13915984028:
@mergify backport v1.4

bk201 (approved):

w13915984028:
code is rebased with master-head

starbops (commented):
I'm unsure if there is any possibility that a user sets this to `rancher/shell:v0.1.26` on purpose. If that's the case, it could be an unexpected behavior for them. Since the PR is targeted to the master branch, maybe we can add a version check for this, or just move the snippet to say `package/upgrade/migrations/upgrade_manifests/v1.3.2/pre-hook.sh` to limit the scope.

w13915984028:
> I'm unsure if there is any possibility that a user sets this to `rancher/shell:v0.1.26` on purpose. If that's the case, it could be an unexpected behavior for them. Since the PR is targeted to the master branch, maybe we can add a version check for this, or just move the snippet to say `package/upgrade/migrations/upgrade_manifests/v1.3.2/pre-hook.sh` to limit the scope.

At the moment, we seemed not seeing the scenario for a user to set it. The only occurance is PR https://github.com/harvester/harvester/pull/6347.

After v1.4.0 is released (which can deal with potential v1.3.2, v1.3.3...), we can drop this code from master-head.

starbops (approved):
LGTM, thank you.

Issue #6277, opened 2024-08-06
[ENHANCEMENT] Allow update the default value of Harvester settings from Chart


w13915984028:
**Is your enhancement related to a problem? Please describe.**

Observed this weird value from a SB:

```
- apiVersion: harvesterhci.io/v1beta1
  default: '{}'
  kind: Setting
  metadata:
    creationTimestamp: "2024-05-28T18:04:11Z"
    generation: 40
    managedFields:
    - apiVersion: harvesterhci.io/v1beta1
      fieldsType: FieldsV1
      fieldsV1:
        f:default: {}
        f:status: {}
      manager: harvester
      operation: Update
      time: "2024-07-26T14:21:10Z"
    name: support-bundle-image
    resourceVersion: "86812263"
    uid: 2662b6a9-17b0-4227-88e6-df8568c37e76
  status: {}
```

A good value looks:
```
settings.harvesterhci.io support-bundle-image -oyaml
apiVersion: harvesterhci.io/v1beta1
default: '{"repository":"rancher/support-bundle-kit","tag":"master-head","imagePullPolicy":"IfNotPresent"}'
kind: Setting
metadata:
  creationTimestamp: "2024-08-15T19:48:37Z"
  generation: 1
  name: support-bundle-image
  resourceVersion: "6646"
  uid: 01952165-f164-4edd-bc62-e5afa52fcfd8
status: {}
```

With the PR https://github.com/harvester/harvester/pull/1771, below log is also observed on Harvester cluster bootstrap stage, as the `App` object is earlier that `Settings.harvesterhci.io` object. But it does no harm as it will sccess finally.

```
time="2024-08-15T19:48:36Z" level=info msg="Starting monitoring.coreos.com/v1, Kind=Alertmanager controller"
time="2024-08-15T19:48:36Z" level=error msg="error syncing 'harvester-system/harvester': handler harvester-setting-controller: settings.harvesterhci.io \"support-bundle-image\" not found, requeuing"
time="2024-08-15T19:48:36Z" level=error msg="error syncing 'harvester-system/harvester': handler harvester-setting-controller: settings.harvesterhci.io \"support-bundle-image\" not found, requeuing"
time="2024-08-15T19:48:36Z" level=error msg="error syncing 'harvester-system/harvester': handler harvester-setting-controller: settings.harvesterhci.io \"support-bundle-image\" not found, requeuing"
```

**Describe the solution you'd like**

tasks
[...]

PR #6358, opened 2024-08-16
Update support-bundle-image default value per chart defnition

w13915984028:

**Problem:**

https://github.com/harvester/harvester/issues/6277

**Solution:**

Use chart and Pod ENV to update setting's default value when it is not fit for hard-coded on source code.

Avoid updatting empty value to settings.Dfault

Inspired from https://github.com/harvester/harvester-installer/pull/813, Harvester should allow to replace the default value and current value of settings from Pod ENV.
```
Further, the chart generate the ENV from Values.yaml, thus we don't need addtional code for update.

The current hard-coded settings, will update the Default automatically after upgrade.

For none-hard-coded settings, do this via the chart definition.

```

**Related Issue:**

https://github.com/harvester/harvester/issues/6277

**Test plan:**

1. Create a new cluster, it will have `support-bundle-image` with generation 1 in general; instead of 2 in the past

```
settings.harvesterhci.io support-bundle-image -oyaml
default: '{"repository":"rancher/support-bundle-kit","tag":"master-head","imagePullPolicy":"IfNotPresent"}'
kind: Setting
metadata:
  creationTimestamp: "2024-08-15T19:48:37Z"
  generation: 1
  name: support-bundle-image
  resourceVersion: "6646"
  uid: 01952165-f164-4edd-bc62-e5afa52fcfd8
status: {}
```

2. Upgrade to new version, the `default` value should always match the setting in chart.

### DIFF
--- a/deploy/charts/harvester/templates/_helpers.tpl
+++ b/deploy/charts/harvester/templates/_helpers.tpl
@@ -105,3 +105,21 @@ NB(thxCode): Use this value to unify the control tag and condition of KubeVirt.
 {{- .Values.tags.kubevirt | toString -}}
 {{- end -}}
 {{- end }}
+
+{{/*
+Get Support-bundle-kit image environment for updating the default values per current release.
+*/}}
+{{- define "harvester.supportBundleImageEnv" -}}
+{{- $result := dict -}}
+{{- range $k, $v := .Values -}}
+{{- if eq (toString $k) "support-bundle-kit" -}}
+{{- $result = $v -}}
+{{- end -}}
+{{- end -}}
+{{- with $result -}}
+{{- with .image -}}
+- name: HARVESTER_SUPPORT_BUNDLE_IMAGE_DEFAULT_VALUE
+  value: {{ printf "{\"repository\":\"%s\",\"tag\":\"%s\",\"imagePullPolicy\":\"%s\"}" .repository .tag .imagePullPolicy | squote }}
+{{- end -}}
+{{- end -}}
+{{- end }}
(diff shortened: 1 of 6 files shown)


w13915984028:
Later, we can drop this controller to update the setting default in such a dependency path. cc @FrankYang0529 @bk201 

https://github.com/harvester/harvester/blob/4e94f46e2f40fcf1cd6efd4530e63870b6945a5f/pkg/controller/master/setting/handler.go#L182

w13915984028:
@mergify backport v1.3 v1.4

w13915984028:
note: a tricky helm issue/limitation https://github.com/helm/helm/issues/12188 is escaped on this PR.

FrankYang0529 (approved):
LGTM.

votdev (approved):
LGTM.


Does it make sense to document that env var feature somewhere?

bk201 (commented):
overall lgtm, but it would be nice to follow https://github.com/harvester/harvester/blob/master/enhancements/20240327-structured-logging.md for new-added log messages.

w13915984028:
@bk201 The log messages are updated, please take a new look, thanks.

bk201 (approved):
Really appreciate!

w13915984028:
All 3 commits are merged into 1 commit for final committing.

w13915984028:
A wiki page is added to describe how to update the `Default` field via chart.

https://github.com/harvester/harvester/wiki/Harvester-Settings-and-Update-the-Default-via-Chart

Issue #6360, opened 2024-08-16
[BUG] Dashboard shows 404 Not Found after upgrading from v1.3.1 to v1.3.2-rc1


bk201:
**Describe the bug**

The dashboard shows 404 Not Found after upgrading from v1.3.1 to v1.3.2-rc1.

<img width="1914" alt="截圖 2024-08-16 下午4 10 19" src="https://github.com/user-attachments/assets/6b659af5-82ce-4d9c-8435-bbd88c7303af">

**To Reproduce**
1. Create a v1.3.1 cluster.
2. Upgrade to v1.3.2-rc1.
3. Access VIP with browser.

**Expected behavior**

Dashboard should display well.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.3.1 -> v1.3.2-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

With @starbops's help, we found the issue is caused by the argument change of `rke2-ingress-nginx-controller` pods:
- `--watch-ingress-without-class=true` argument is removed in RKE2 v1.28.12.
- The ingress `cattle-system/rancher-expose` has no `spec.ingressClassName` and it's ignored by the `rke2-ingress-nginx-controller`:
    ```
    2024-08-16T04:35:15.259654966Z I0816 04:35:15.259320       6 store.go:436] "Ignoring ingress because of error while validating ingress class" ingress="cattle-system/rancher-expose" error="ingress does not contain a valid IngressClass"
    ```
- The workaround is to edit the ingress and add an ingress class:
    ```
    $ kubectl edit ingress rancher-expose -n cattle-system
    
    <...>
    spec:
      ingressClassName: nginx  <--- add this
      rules:
      - http:
          paths:
          - backend:
              service:ot
                name: rancher
                port:
                  number: 80
            path: /
            pathType: Prefix
    <...>
    ```
- New v1.3.2-rc1 clusters are not affected. Even our bootstrap code creates the ingress without an ingress class, the resulting ingress has the `ingressClassName` field. The reason is unknown.
- This should affect v1.4-head too.

starbops:
From the `kubectl.kubernetes.io/last-applied-configuration` annotation in the Ingress object, we know that the `.spec.ingressClassName` field was added on the fly:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #6365, opened 2024-08-19
fix(upgrade): set ingress class name for rancher-expose

starbops:

**Problem:**

After upgraded to Harvester master build from v1.3.1, we notice that the dashboard becomes inaccessible. This is due to the removal of the `--watch-ingress-without-class` argument of rke2-ingress-nginx-controller and the addition of `ingressclass.kubernetes.io/is-default-class` annotation in the `nginx` IngressClass in newer versions. The result is the `rancher-expose` Ingress has no load-balancer configured.

**Solution:**

Ensure the rancher-expose Ingress object has the ingress class name configured. The value is determined from the default ingress class. This is to cope with the DefaultIngressClass admission controller's inability to handle existing Ingress objects.

**Related Issue:**

#6360 

**Test plan:**

Upgrade a Harvester cluster of version v1.3.1 to master build containing this PR.

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -415,6 +415,29 @@ modify_nad_bridge() {
       done
 }
 
+ensure_ingress_class_name() {
+  echo "Ensuring existing rancher-expose Ingress has ingress class name specified"
+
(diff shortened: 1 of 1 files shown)


starbops:
@mergify backport v1.3 v1.4

bk201 (approved):
LGTM, thanks!

FrankYang0529 (approved):
LGTM. Thanks.

Issue #6375, opened 2024-08-19
[BUG] VM created over 20 mins still shows `VirtualMachine generation is 2, but latest observed generation is 1`


albinsun:
**Describe the bug**
Even VM created over 30 mins, it still shows `VirtualMachine generation is 2, but latest observed generation is 1`.
Note that we had similiar issue in `v1.3.0`, see #4909.

**To Reproduce**
1. Create a 2 nodes `harvester-v1.3.2-rc1`
2. Create VM image `ubuntu-24.04-noble-server-cloudimg-amd64.img`
3. Create VM network `mgmt-vlan1`
4. Create VM `noble`

**Expected behavior**
According to previous behavior, `VirtualMachine generation...` message will gone in few minutes.

**Support bundle**
[supportbundle_generation_msg_2024-08-19T08-20-04Z.zip](https://github.com/user-attachments/files/16657712/supportbundle_generation_msg_2024-08-19T08-20-04Z.zip)

**Environment**
* `harvester-v1.3.2-rc1`
  * Profile: 2 nodes QEMU/KVM (8C/16G/500G per node)
  * ui-source: `Auto`

**Additional context**
1. Found during verifying #6321 
1. Seems not related to OS version cuz not only hit in ubuntu **noble** but also **jammy**.
   ![image](https://github.com/user-attachments/assets/b074a348-9ab6-4ce1-b9ff-ec9343a5fab2)

w13915984028:
@albinsun Did you see such on Harvester v1.4.0 / master-head cluster? thanks.

w13915984028:
I did not notice it on master-head; let me double check on v1.3.2.

w13915984028:
Easy to produce on v1.3.2.
![image](https://github.com/user-attachments/assets/57bafb3e-e8e5-4570-abe0-1339c0d16253)

vm
```

apiVersion: v1
items:
- apiVersion: kubevirt.io/v1
  kind: VirtualMachine
  metadata:
    annotations:
      harvesterhci.io/vmRunStrategy: RerunOnFailure
      harvesterhci.io/volumeClaimTemplates: '[{"metadata":{"name":"vm1-disk-0-fzg2n","annotations":{"harvesterhci.io/imageId":"default/image-mbhxj"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"4Gi"}},"volumeMode":"Block","storageClassName":"longhorn-image-mbhxj"}}]'
[...]

PR #6379, opened 2024-08-19
Fix vm regeneration issue

w13915984028:

**Problem:**

#6375 

**Solution:**

Check&patch vm generation in all cases, such log are seen in debug mode:

>time="2024-08-19T12:51:10Z" level=debug msg="set VM vm8 management network default macAddress to 66:97:f9:81:85:b7"
time="2024-08-19T12:51:10Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM default/vm8 has running pod, skip check"
time="2024-08-19T12:51:10Z" level=debug msg="set VM vm8 management network default macAddress to 66:97:f9:81:85:b7"
time="2024-08-19T12:51:10Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM default/vm8 has running pod, skip check"
time="2024-08-19T12:51:10Z" level=debug msg="VM default/vm8 delete and create new cr object to patch generation: {\"spec\":{\"runStrategy\":\"RerunOnFailure\",\"template\":{\"metadata\":{\"annotations\":{\"harvesterhci.io/sshNames\":\"[]\"},\"creationTimestamp\":null,\"labels\":{\"harvesterhci.io/vmName\":\"vm8\"}},\"spec\":{\"affinity\":{},\"architecture\":\"amd64\",\"domain\":{\"cpu\":{\"cores\":1,\"sockets\":1,\"threads\":1},\"devices\":{\"disks\":[{\"bootOrder\":1,\"disk\":{\"bus\":\"virtio\"},\"name\":\"disk-0\"},{\"disk\":{\"bus\":\"virtio\"},\"name\":\"cloudinitdisk\"}],\"inputs\":[{\"bus\":\"usb\",\"name\":\"tablet\",\"type\":\"tablet\"}],\"interfaces\":[{\"macAddress\":\"66:97:f9:81:85:b7\",\"masquerade\":{},\"model\":\"virtio\",\"name\":\"default\"}]},\"features\":{\"acpi\":{\"enabled\":true}},\"machine\":{\"type\":\"q35\"},\"memory\":{\"guest\":\"924Mi\"},\"resources\":{\"limits\":{\"cpu\":\"1\",\"memory\":\"1Gi\"},\"requests\":{\"cpu\":\"62m\",\"memory\":\"682Mi\"}}},\"evictionStrategy\":\"LiveMigrateIfPossible\",\"hostname\":\"vm8\",\"networks\":[{\"name\":\"default\",\"pod\":{}}],\"terminationGracePeriodSeconds\":120,\"volumes\":[{\"name\":\"disk-0\",\"persistentVolumeClaim\":{\"claimName\":\"vm8-disk-0-u5auq\"}},{\"cloudInitNoCloud\":{\"networkDataSecretRef\":{\"name\":\"vm8-p0oyv\"},\"secretRef\":{\"name\":\"vm8-p0oyv\"}},\"name\":\"cloudinitdisk\"}]}}}}"

**Related Issue:**

#6375 

**Test plan:**

1. Create new VM (after any kind of patch like patch MacAddr to vm, the generation is changed), no issue of `VirtualMachine generation is 2, but latest observed generation is 1`
2. Stop & start VM (to change vm object generation), no issue
3. Add new label to vm, then `Save and Restart` (to change vm object generation),  no issue

### DIFF
--- a/pkg/controller/master/virtualmachine/vmi_network_controller.go
+++ b/pkg/controller/master/virtualmachine/vmi_network_controller.go
@@ -3,7 +3,6 @@ package virtualmachine
 import (
 	"encoding/json"
 	"fmt"
-	"reflect"
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -71,13 +70,13 @@ func (h *VMNetworkController) updateVMDefaultNetworkMacAddress(vmi *kubevirtv1.V
 		}
 	}
 
-	if reflect.DeepEqual(vmCopy.Spec.Template.Spec.Domain.Devices, vm.Spec.Template.Spec.Domain.Devices) {
-		return nil
-	}
-
-	logrus.Debugf("update default network mac address of the vm: %s", vmi.Name)
+	// note: this function is related to vm/vmi, should always run on vmi change
 	if err := h.regenerateControllerRevision(vmi, vm); err != nil {
 		return err
+	}
+
+	if equality.Semantic.DeepEqual(vmCopy.Spec.Template.Spec.Domain.Devices, vm.Spec.Template.Spec.Domain.Devices) {
+		return nil
 	}
 	if _, err := h.vmClient.Update(vmCopy); err != nil {
 		return err
@@ -104,7 +103,7 @@ func (h *VMNetworkController) regenerateControllerRevision(vmi *kubevirtv1.Virtu
 		if deletionErr := h.crClient.Delete(vmi.Namespace, crObj.Name, &metav1.DeleteOptions{}); deletionErr != nil {
 			return fmt.Errorf("error during deletion of controllerRevision for vmi %s-%s: %v", vmi.Name, vmi.Name, deletionErr)
 		}
-
+		logrus.Debugf("VM %s/%s delete and create new cr object to patch generation: %s", vm.Namespace, vm.Name, patch)
 		crObj.Data.Raw = patch
 		crObj.ResourceVersion = ""
 		_, err = h.crClient.Create(crObj)


w13915984028:
@mergify backport v1.4 v1.3

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -71,14 +70,14 @@ func (h *VMNetworkController) updateVMDefaultNetworkMacAddress(vmi *kubevirtv1.V
...
-		return nil
-	}
-
-	logrus.Debugf("update default network mac address of the vm: %s", vmi.Name)
+	// note: this function is related to vm/vmi, should always run on vmi change
 	if err := h.regenerateControllerRevision(vmi, vm); err != nil {
 		return err
 	}
+
+	if equality.Semantic.DeepEqual(vmCopy.Spec.Template.Spec.Domain.Devices, vm.Spec.Template.Spec.Domain.Devices) {
```
brandboat, line 78: Just curious, why do we use `equality.Semantic.DeepEqual` instead of `reflect.DeepEqual` here ?

w13915984028: there are pointers in the struct, `equality.Semantic.DeepEqual` checks the value of the pointer

ibrokethecloud (approved):
lgtm. thanks.

pkg/controller/master/virtualmachine/vmi_network_controller.go
brandboat: Correct me if I'm wrong, but I tried reflect.DeepEqual. Looks like it also compares the value of the pointer.
https://go.dev/play/p/56N4QZIKOKZ?v=goprev

I'm not trying to debate something, just want to clarify the timing of using `equality.Semantic.DeepEqual` / `reflect.DeepEqual`. As in the harvester repo we mostly use `reflect.DeepEqual` for comparison.

brandboat: Just a quick survey, what I know is that there are differences in comparison for certain data structures. In Kubernetes, equality.Semantic.DeepEqual returns true when comparing nil and empty maps, meaning they are considered equal. https://pkg.go.dev/k8s.io/apimachinery/pkg/api/equality

However, this is not the case with reflect.DeepEqual, where nil and empty maps are not considered equal.
see https://go.dev/play/p/U8zt1G34ztR?v=goprev

I'm not sure if this is something we need to aware of

w13915984028: Good question and investigation. Thanks.

Honestly, I did not dive so much on this topic. The reason to use `equality.Semantic.DeepEqual` instead of `reflect` is: the `equality` package has been included in this file but `reflect` was added later. And it was used in another function to check as such `!equality.Semantic.DeepEqual(revisionSpec.Spec.Template, vm.Spec.Template)`

w13915984028: Further tested with following code, both `reflect` and `equality` work well on this conext. Preferring `equality` is mainly because it has been used in this file.

My above answer `there are pointers in the struct, equality.Semantic.DeepEqual checks the value of the pointer` was not fully accurate, sorry for the inconvenience.

image

Thanks again.

bk201 (approved):

FrankYang0529 (approved):
LGTM. Thanks.

brandboat (approved):
LGTM, thanks

Issue #5855, opened 2024-05-23
[BUG] Backup and snapshot can not have same name


albinsun:
**Describe the bug**
Backup and snapshot can not have same name, even from different VM.

**To Reproduce**
1. <details>
   <summary> Create a VM <code>johndoe</code>
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/f4c214c8-c676-400b-ba0c-85d5fd2ec93a)
   </details>

2. <details>
   <summary> Backup VM <code>johndoe</code> with name <code>john</code>
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/acb4a100-73fb-4894-af81-934d42ecb18d)
   </details>

3. <details>
   <summary> :red_circle: Snapshot <code>johndoe</code> with same name <code>john</code>
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/b5ea69d3-4eea-4847-ab1f-8f4dfc602361)
   </details>

   > `virtualmachinebackups.harvesterhci.io "john" already exists`

4. <details>
   <summary> :red_circle: Snapshot another VM with same name <code>john</code>
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/eff9e8ea-481b-4522-8b89-179c8f399897)
   </details>

   > `virtualmachinebackups.harvesterhci.io "john" already exists`

**Expected behavior**
If we regard **backup** and **snapshot** different features, and they should not affect each other. 

**Support bundle**
[supportbundle_BackupAndSnapshot.zip](https://github.com/harvester/harvester/files/15410516/supportbundle_BackupAndSnapshot.zip)

**Environment**
* Harvester
  * Version: `v1.2.2`
  * Profile: QEMU/KVM, 3 nodes (8C/16G/500G)
  * ui-source: `Auto`

**Additional context**
1. <details>
   <summary> :red_circle: Reversing test order, namely snapshot then backup, also gives similiar problem.
   </summary>

   ![image](https://github.com/harvester/harvester/assets/2773781/98cddc90-79b2-414b-a7e3-b350341b7647)
   </details>

   > `failed to create VM backup, error: virtualmachinebackups.harvesterhci.io "john" already exists`

Yu-Jack:
Add some contexts, when creating backup and snapshot, it created two `virtualmachinebackups.harvesterhci.io` with different type. So, it's not able to create same name backup and snapshot.

```yaml
[...]

PR #6384, opened 2024-08-20
feat: handle duplicate vm backup and snapshot

FrankYang0529:

**Problem:**
VM Backup and Snapshot use same CRD, but they're in different pages, so users may wonder why some names can't work.

**Solution:**
Give a better error message to point out VM Backup and Snapshot can't have same name.
Screenshot 2024-08-28 at 11 19 12 AM
Screenshot 2024-08-28 at 11 19 59 AM

**Related Issue:**
https://github.com/harvester/harvester/issues/5855

**Test plan:**

* Create a VM Backup with name `test`.
* Create a VM Snapshot with name `test`. Webhook should return an error message about VM Backup and Snapshot use same CRD.

### DIFF
--- a/pkg/webhook/resources/virtualmachinebackup/mutator.go
+++ b/pkg/webhook/resources/virtualmachinebackup/mutator.go
@@ -1,0 +1,57 @@
+package virtualmachinebackup
+
+import (
+	"fmt"
+
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	werror "github.com/harvester/harvester/pkg/webhook/error"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/virtualmachinebackup/mutator.go
```diff
@@ -0,0 +1,53 @@
...
+		APIGroup:   v1beta1.SchemeGroupVersion.Group,
+		APIVersion: v1beta1.SchemeGroupVersion.Version,
+		ObjectType: &v1beta1.VirtualMachineBackup{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+		},
+	}
+}
+
+func (m *BackupMutator) Create(_ *types.Request, newObj runtime.Object) (types.PatchOps, error) {
```
bk201, line 38: ~~How about handling the backup types individually? Say there is a snapshot with an identical name when the user creates a backup. And vice versa.~~

Forgive me. You do handle this in the error message with backup type, but it seems to always return the current creating backup type, just like there is no webhook case.

FrankYang0529: Updated it. Thanks for the suggestion.

WebberHuang1118 (approved):
LGTM, thanks for the PR.

bk201 (approved):
Thanks!

FrankYang0529:
@mergify backport v1.4

bk201:
@mergifyio backport v1.3

Issue #6158, opened 2024-07-11
[TASK] Bump Harvester component minor versions for v1.4


bk201:
Bump Harvester components' minor version x.y.z to x.y+1.z in v1.4.

cc @tserong 

### Charts

| Done |  | v1.3-head  | v1.4-head | Owner |
|---|---|---|---|---|
| 🟢 | harvester-network-controller  |  0.4.0 |  0.5.1 | @mingshuoqiu  |
| 🟢 | harvester-node-disk-manager | 0.6.0  | 0.7.2  | @Vicente-Cheng |
| 🟢 | harvester-load-balancer |  0.3.0 | 0.4.0 |  @w13915984028 |
| 🟢 | harvester-node-manager |  0.2.3 | 0.3.1 | @brandboat  | 
| 🟢 | (add-on) pcidevices-controller | 0.3.2 | 0.4.1 | @Yu-Jack   |
| 🟢 | (add-on) seeder | 0.3.0 | 0.4.0 | @starbops   |
| 🟢  | (add-on) vm-import-controller | v0.3.0 | 0.4.0 | @ibrokethecloud  |
|  | (add-on) nvidia-driver-runtime | v0.1.1 | 0.2.0 | @ibrokethecloud  |
|  | (add-on) vcluster | | | Experimental add-on |
|  | (add-on) harvester-vm-dhcp-controller | | | Skip, Experimental add-on |

> **How to bump a chart**
> - Fork and clone https://github.com/harvester/charts.
> - Check out a local branch from the `release` branch.
> - Modify a chart's version in the `Chart.yaml` file.
> - Create a PR to merge the change to the `release` branch.
> - A new chart version will be published once the PR is merged.
> - Sample PR: https://github.com/harvester/charts/pull/242

### Images

|Done|   | v1.3-head | v1.4-head  | Owner |
|---|---|---|---|---|
| 🟢 | harvester-network-controller  |  0.4.0 |  0.5.1 | @mingshuoqiu  |
| 🟢 | harvester-node-disk-manager | 0.6.0  | 0.7.1  | @Vicente-Cheng |
| 🟢 | harvester-load-balancer |  0.3.0 | 0.4.1 |  @w13915984028 |
| 🟢 | harvester-node-manager |  0.2.1 | 0.3.0 | @brandboat  | 
| 🟢 | (add-on) pcidevices-controller | 0.3.2 | 0.4.0 | @Yu-Jack   |
[...]

PR #6390, opened 2024-08-21
Bump node-manager to 0.3.0

brandboat:
related to https://github.com/harvester/harvester/issues/6158


brandboat:
close this as we should wait load balancer bump

Issue #6395, opened 2024-08-21
[BUG] The Harvester setting kubeconfig-default-token-ttl-minutes is only applied to Rancher once


w13915984028:
**Describe the bug**

The Harvester setting `kubeconfig-default-token-ttl-minutes` is only applied to Rancher once

**To Reproduce**
1. Set Harvester settings `kubeconfig-default-token-ttl-minutes` to value X, the Rancher setting `kubeconfig-default-token-ttl-minutes` is also updated to X
2. Manually change Rancher `kubeconfig-default-token-ttl-minutes` to Y; (This simulates the possible scenario where setting of Rancher was changed)
3. Restart Harvester POD, the Rancher `kubeconfig-default-token-ttl-minutes` is still X;

**Expected behavior**

If the backend setting `kubeconfig-default-token-ttl-minutes` was changed in various possibilities, Harvester should re-apply the setting to the backend at least when Harvester POD restarts.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.4.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

https://github.com/harvester/harvester/issues/6378

In general, for any Harvester settings, if it is forwarded to other components, there are 3 options:
(1) Set once
(2) Add a controller to watch and sync all the time
(3) When POD starts, check and sync

Most of the settings are using (1), but (2)/(3) will be better. This PR https://github.com/harvester/harvester/pull/6396 adopts (3).

w13915984028:
Test plan:

(1) From harvester UI, set `kubeconfig-default-token-ttl-minutes` setting to 36000

Check the Rancher settings from kubectl
```
$kubectl get settings.management.cattle.io | grep 36000
auth-token-max-ttl-minutes                36000
kubeconfig-default-token-ttl-minutes      36000
```

(2) Manual re-set Rancher setting to another value
```
$kubectl get settings.management.cattle.io kubeconfig-default-token-ttl-minutes
NAME                                   VALUE
kubeconfig-default-token-ttl-minutes   2600
```

(3) Kill the Harvester leader POD
```
$kubectl delete  pods -n harvester-system harvester-c8d679898-pzzkn
pod "harvester-c8d679898-pzzkn" deleted
```

Check the log of new Harvester leader POD, get such :
```
[...]

Issue #6378, opened 2024-08-19
[BUG] Harvester setting `log-level` has no effect after pod is restarted


w13915984028:
**Describe the bug**

**To Reproduce**
1. Set to `debug`
2. Restart the Harvester pod
3. The log-level reverts to `info`

**Expected behavior**

Persistant and/or update the document.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.4.0 master-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

w13915984028:
Test plan:
(1) From Harvester UI, set `log-level` to `debug`;
(2) Check Harvester POD log, should see many debug level logs and info/warning/error level logs;
(3) Kill the leader Harvester POD, the new leader/the replaced POD should still print debug logs;
(4) Reboot the cluster, after that the Harvester POD still print `debug` log
(5) Repeat (1)..(4), set `log-level` to `Info`, should only and always observe `Info` (and above like warning/error) level log.

w13915984028:
FYI: When testing with v1.4/master, the ISO needs to be after 2024.08.30, it will include the PR https://github.com/harvester/harvester/pull/6460 / https://github.com/harvester/harvester/pull/6396, thanks.

m-ildefons:
Failed to verify with `v1.4.0-dev-20240830`. Was this not included in that release?

w13915984028:
Need to wait a bit:

The `v1.4.0-dev-20240830` did not include the related PR.

https://github.com/harvester/harvester/commits/v1.4.0-dev-20240830/
https://github.com/harvester/harvester/commits/v1.4/

m-ildefons:
Testenvironment:
Harvester v1.4.0-rc4 on baremetal (falcons)

Test 1: Default Setting
:green_circle: Logs contain only `info` level or above messages

Test 2: Change Log-Level Setting to `debug`
:green_circle: Logs now also contain `debug` level messages

Test 3: Restart Harvester Pod
:green_circle: The new Harvester pod also writes `debug` level messages to its log

Test 4: Reboot Harvester Node
:green_circle: Logs still contain `debug` level messages

![Screenshot at 2024-10-25 11-04-24](https://github.com/user-attachments/assets/f567fb0d-e021-4174-a4f2-e48d3e5188ff)

Test 5: Change Log-Level Setting back to `info`
:green_circle: `debug` level log messages stop appearing in the log

[...]

PR #6396, opened 2024-08-21
Fix setting log-level and kubeconfig-default-token-ttl-minutes

w13915984028:

**Problem:**

Per issue describtion.

When settings are not added to `skipHashCheckSettings`, they will only be applied once when the change happens, and later they are not applied again when Harvester POD restarts.

In this context, it is not a `setting` but rather an `action`.

```
	skipHashCheckSettings = []string{
		settings.AutoRotateRKE2CertsSettingName,
		settings.LogLevelSettingName,
		settings.KubeconfigDefaultTokenTTLMinutesSettingName,
```

**Solution:**

Let setting rerun when pod starting.

No matter a change is changed or not, the controller should bring no harm to rerun.

**Related Issue:**
https://github.com/harvester/harvester/issues/6395
https://github.com/harvester/harvester/issues/6378

**Test plan:**

Refer issue: https://github.com/harvester/harvester/issues/6395#issuecomment-2302338402

### DIFF
--- a/pkg/controller/master/setting/handler.go
+++ b/pkg/controller/master/setting/handler.go
@@ -43,6 +43,8 @@ var (
 	}
 	skipHashCheckSettings = []string{
 		settings.AutoRotateRKE2CertsSettingName,
+		settings.LogLevelSettingName,
+		settings.KubeconfigDefaultTokenTTLMinutesSettingName,
 	}
 )
 
--- a/pkg/controller/master/setting/kubeconfigttl.go
+++ b/pkg/controller/master/setting/kubeconfigttl.go
@@ -3,6 +3,8 @@ package setting
 import (
 	"fmt"
 
+	"github.com/sirupsen/logrus"
+
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	harvSettings "github.com/harvester/harvester/pkg/settings"
 )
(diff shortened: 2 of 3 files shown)


w13915984028:
@mergify backport v.14

w13915984028:
cc @bk201 @ibrokethecloud @FrankYang0529 

Please double check: should this PR be backported to v1.3 as well? thanks.

FrankYang0529 (approved):
LGTM. Thank you.

bk201:
@w13915984028 I understand why we need the change for the `log-level` setting. But I still can't understand why we need this change for the `kubeconfig-default-token-ttl-minutes` setting.

From the description in https://github.com/harvester/harvester/issues/6395, you mention
> 2. Manually change Rancher kubeconfig-default-token-ttl-minutes to Y; (This simulates the possible scenario like in the upgrade path, the Rancher applies a new value)

If we override the Rancher setting, the value is stored in Rancher settings's `.value` field. If somehow Rancher want to change the setting default value, the value will change in the `.default` field. What we set in the `.value` field still takes effect.

w13915984028:
> @w13915984028 I understand why we need the change for the `log-level` setting. But I still can't understand why we need this change for the `kubeconfig-default-token-ttl-minutes` setting.
> 
> From the description in #6395, you mention
> 
> > 2. Manually change Rancher kubeconfig-default-token-ttl-minutes to Y; (This simulates the possible scenario like in the upgrade path, the Rancher applies a new value)
> 
> If we override the Rancher setting, the value is stored in Rancher settings's `.value` field. If somehow Rancher want to change the setting default value, the value will change in the `.default` field. What we set in the `.value` field still takes effect.

@bk201 The description is not full. The story could be updated to:

Harvester will check & amend the `kubeconfig-default-token-ttl-minutes` settings to Rancher when Harvester POD starts, if Rancher side value was not same, Harvester saved value will replace it.

In general, for any Harvester settings, if it is forwarded to other components, there are 3 options:
(1) Set once
(2) Add a controller to watch and sync all the time
(3) When POD starts, check and sync

Most of the settings are using (1), but (2)/(3) will be better. This PR adopts (3).

w13915984028:
Code is rebased to solve merge conflicts.

bk201 (approved):
lgm, sorry need to fix the conflict again.

w13915984028:
@mergify backport v1.4

PR #6397, opened 2024-08-22
Improve vm migration ux to include VM's with hostname specification

ibrokethecloud:

**Problem:**

Existing work done to address https://github.com/harvester/harvester/issues/4888 includes changes to identify VM's
* which are marked non migratable by kubevirt
* identify VM's which have their last healthy replica on the node being placed in toe maintenance mode 
* VM's which have node scheduling requirements which prevents it from being scheduled on other nodes in the cluster
* VM's which have container disk or a cdrom

This currently skips VM's which have node selector requriements

**Solution:**

PR introduces an additional check to find VM's with node selector requirements such as when hostname is added using `kubernetes.io/hostname` label

**Related Issue:**
**Test plan:**

### DIFF
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -456,6 +456,8 @@ func filterNodesForNodeSelector(possibleNodes []*corev1.Node, vmi *kubevirtv1.Vi
 				validNodes = append(validNodes, v)
 			}
 		}
+	} else {
+		return possibleNodes
 	}
 	return validNodes
 }
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -421,7 +421,7 @@ func (ndc *ControllerHandler) CheckVMISchedulingRequirements(originalNode *corev
 		return nil, fmt.Errorf("error listing nodes from nodeCache: %v", err)
 	}
 	for _, vmi := range vmiList {
-		var possibleNodes []string
+		var possibleNodes, validNodes []*corev1.Node
 		if vmi.Spec.Affinity != nil && vmi.Spec.Affinity.NodeAffinity != nil && vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -430,18 +430,36 @@ func (ndc *ControllerHandler) CheckVMISchedulingRequirements(originalNode *corev
...
+	if vmi.Spec.NodeSelector != nil && len(vmi.Spec.NodeSelector) > 0 {
+		vmiNodeSelector := labels.SelectorFromSet(vmi.Spec.NodeSelector)
+		for _, v := range possibleNodes {
+			nodeLabels := labels.Set(v.GetLabels())
+			if vmiNodeSelector.Matches(nodeLabels) {
+				validNodes = append(validNodes, v)
+			}
+		}
+	}
+	return validNodes
```
tserong, line 462: If the VM doesn't have a node selector, this looks like it'll return an empty slice, which means  CheckVMISchedulingRequirements() will think there's no valid nodes. In this case, should we return possibleNodes unaltered instead?

ibrokethecloud: good point, we should return possibleNodes in that case. i have made the change.

tserong (approved):

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -430,18 +430,38 @@ func (ndc *ControllerHandler) CheckVMISchedulingRequirements(originalNode *corev
...
+	var validNodes []*corev1.Node
+	// VM's may also have node selector, which is used when defining specific hostnames
+	if vmi.Spec.NodeSelector != nil && len(vmi.Spec.NodeSelector) > 0 {
+		vmiNodeSelector := labels.SelectorFromSet(vmi.Spec.NodeSelector)
+		for _, v := range possibleNodes {
+			nodeLabels := labels.Set(v.GetLabels())
+			if vmiNodeSelector.Matches(nodeLabels) {
+				validNodes = append(validNodes, v)
+			}
+		}
```
Vicente-Cheng, line 458: How about the following?
```suggestion
		}
		return validNodes
	}
	return possibleNodes
```

The `validNodes` will be filled and we can return if we get nodeSelector. Others return the `possibleNodes` 
WDYT?

Vicente-Cheng (approved):
just a nit, others lgtm!

Vicente-Cheng:
@mergifyio backport v1.4

ibrokethecloud:
@mergifyio backport v1.3

PR #6167, opened 2024-07-12
HEP: Support the Longhorn V2 Data Engine (v1.4.0, Experimental)

tserong:
**Related Issue:**
https://github.com/harvester/harvester/issues/5274

Issue #5274, opened 2024-03-02
[FEATURE] Support Longhorn v2 data engine - Experimental


Vicente-Cheng:
**Is your feature request related to a problem? Please describe.**
Longhorn v2 engine use the raw device as backend device.
node-disk-manager now only supports provision filesystem for longhorn backend
we should add a capability to provision raw devices and fill related fields to longhorn node.

**Describe the solution you'd like**
Add another provision behavior for the raw device

**Describe alternatives you've considered**

**Additional context**
https://longhorn.io/docs/1.6.0/v2-data-engine/quick-start/#enable-v2-data-engine

innobead:
1.5 will use Longhorn 1.8 or 1.9 which should have a more stable v2 engine.

Vicente-Cheng:
Briefly list the following items for this feature:
- [ ] OS side, we need hugepage and corresponding kernel module (REF: https://github.com/harvester/os2/pull/98)
- [ ] NDM side, this is an entry point because we need to format the physical backend for the Longhorn v2 volume (NDM will decompose the provisioned for `longhornv1`, `longhornv2` and `lvm` on v1.4.0)
- [ ] UI part
- [ ] MISC stuff (like the Longhorn v2 engine did not support live migration. We need to forbid this operation on the harvester side)

cc @Yu-Jack, @bk201, @innobead @derekbit @ibrokethecloud

w13915984028:
Besides the above list, a few questions:
(1) Will v1 and v2 are supported at the same time? When true, which one will be the default one on Harvester. Is this a global setting or local setting against volume/disk or both ?
(2) From user perspective, what's the change when creating images/volumes/storageclass/vms?
(3) Upgrade path.

The `require/HEP` label was added, guess the HEP will have detailed user stories.

Vicente-Cheng:
>(1) Will v1 and v2 are supported at the same time? When true, which one will be the default one on Harvester. Is this a global setting or local setting against volume/disk or both ?

Yeah, we should have capability provision v1 or v2 in the same time (but will use different physical disk)
We should not have any setting for it. It is determined when adding storage. 

>(2) From user perspective, what's the change when creating images/volumes/storageclass/vms?

It should not have any change.

>(3) Upgrade path.

It should not be affected.

I thought we would have more detailed information with HEP so we could discuss it later with HEP.

tserong:
> Yeah, we should have capability provision v1 or v2 in the same time (but will use different physical disk)
[...]

PR #6405, opened 2024-08-22
feat: Add ability to enable Longhorn V2 Data Engine

tserong:
**Problem:**
We need to make it easy for users to enable the Longhorn V2 Data Engine

**Solution:**
As described in the HEP (https://github.com/harvester/harvester/pull/6167)

**Related Issue:**
https://github.com/harvester/harvester/issues/5274

**Test plan:**
TBD

**Note:**
There's also a GUI PR: https://github.com/harvester/dashboard/pull/1108


tserong:
Force pushed to fix lint errors.

Not sure I agree with CodeFactor about the complex method. I didn't think it was _that_ complex...

pkg/controller/master/setting/register.go
```diff
@@ -74,24 +80,91 @@ func Register(ctx context.Context, management *config.Management, options config
...
+
+func (h *Handler) nodeOnChanged(_ string, node *corev1.Node) (*corev1.Node, error) {
+	if node == nil || node.DeletionTimestamp != nil {
+		return node, nil
+	}
+
+	// This ensures newly added cluster nodes have a node config created
+	// which is necessary to ensure the v2 data engine setting is correct
+	// on newly added nodes.
+	nodeConfig, err := h.getOrCreateNodeConfigForNode(node)
```
ibrokethecloud: the nodeConfig is not really needed until we have figured out if we need to enable/disable v2 engine, L148. 

may be we could move this check there and avoid having to runs this method and ops unless they are needed

tserong: There's one annoying case that we need this line here for, which is: when someone has already enabled the v2 engine, and then later they add a new node.  That node won't have any labels set (the label is only for if you want to _disable_ the engine on a node), so the code would never get past line 128 in this function, and the node config wouldn't be created for the new node, so it wouldn't have the engine enabled.

Or, is there another way I can catch the "new node was added" case, without doing this on every node change?

tserong: Oh, hey, what about this? When we create the node config the first time inside `getOrCreateNodeConfigForNode()`, we add an annotation to the node (something like "node.harvesterhci.io/node-config-created=true"), then in `nodeOnChanged()` we check for that annotation before the above call. That way in the normal case the only thing that executes is a quick lookup in the node's annotations map. It's possibly slightly ugly code-wise, but more efficient than calling `getOrCreateNodeConfigForNode()` on every damn node onchange event.

tserong:
Reworked after discussion with @ibrokethecloud to make `nodeOnChange()` do all the work of creating and updating nodeconfig CRDs, rather than splitting it between `nodeOnChange()` and `syncNodeConfig()`.

ibrokethecloud (approved):
lgtm. thanks.

tserong:
(Force pushed to manually resync with https://github.com/harvester/node-manager/pull/55)

tserong:
(Force pushed to cleanup import order)

bk201 (approved):

tserong:
Rebased on latest master

tserong:
Updated to pull in latest node-manager and go-common in go.mod

tserong:
I've added the node-manager chart bump to this PR, and I've also just merged the GUI PR, so if you build an ISO image of this PR now, you should get a Harvester install that lets you enable the V2 data engine.

pkg/controller/master/setting/nodeconfig.go
```diff
@@ -0,0 +1,156 @@
...
+	}).Info("Processing setting")
+
+	nodes, err := h.nodeClient.List(metav1.ListOptions{})
+	if err != nil {
+		logrus.WithError(err).Error("Failed to list all nodes")
+		return err
+	}
+
+	// Enqueue all the nodes so that nodeOnChanged can sync the node config.
+	for _, node := range nodes.Items {
```
w13915984028, line 37: Two questions here:
(1) Should we enqueue only after the successful processing of the following code? or always wakeup the nodes.

(2) From the context, it looks LH and each node takes time to enable/disable the V2 engine; how to prevent a user from enable & disable it on UI frequently? thanks.

pkg/controller/master/setting/handler.go
```diff
@@ -71,7 +71,7 @@ type Handler struct {
 	managedChartCache    ctlmgmtv3.ManagedChartCache
 	helmChartConfigs     ctlhelmv1.HelmChartConfigClient
 	helmChartConfigCache ctlhelmv1.HelmChartConfigCache
-	nodeClient           ctlcorev1.NodeClient
+	nodeClient           ctlcorev1.NodeController
```
w13915984028, line 74: Please also check 
```
	skipHashCheckSettings = []string{
		settings.AutoRotateRKE2CertsSettingName,
	}
```

As this setting will be processed only once by default. If LH config was maunally/accidently changed later; Harvester will not sync again.

Add the related controller here, to have at least another chance, when the POD is restarted, to sync again.

Refer: https://github.com/harvester/harvester/pull/6396#issuecomment-2314476254

w13915984028 (commented):
A few questions, thanks.

pkg/controller/master/setting/nodeconfig.go
```diff
@@ -6,83 +6,151 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	enableV2DataEngine := enableV2DataEngineSetting.Value == "true"
+	// Don't enable for witness nodes
+	if _, found := node.Labels[ctlnode.HarvesterWitnessNodeLabelKey]; found {
+		enableV2DataEngine = false
+	}
+	// Don't enable if explicitly disabled for this node
+	if value, found := node.Labels[longhorntypes.NodeDisableV2DataEngineLabelKey]; found && value == longhorntypes.NodeDisableV2DataEngineLabelKeyTrue {
```
Vicente-Cheng, line 98: Just curious, is this label set up manually by the user?
Or do we have any mechanism to set up this label?

Vicente-Cheng (approved):
just a nit, others LGTM. Thanks!

pkg/controller/master/setting/nodeconfig.go
tserong: 1. We have to enqueue the nodes first, because the subsequent setting of `lhs/v2-data-engine=true` won't succeed until all the nodes are in the correct state (i.e. have enough hugepages allocated). It may take a little while for the nodes to be in the right state, but the LH setting change will be requeued and it will stick in due course.
2. I hadn't considered a means of preventing users from enabling/disabling it frequently. As this is a preview feature I was planning to go into some detail in the documentation about the need to apply the setting, then wait for it to take effect, and possibly reboot nodes if necessary (see also the text in the GUI PR at https://github.com/harvester/dashboard/pull/1108)

pkg/controller/master/setting/handler.go
tserong: Thanks for that. In this case I think we're probably OK because we pick up settings changes in `nodeOnChanged`, which happens more frequently.

pkg/controller/master/setting/nodeconfig.go
tserong: It's to be set manually by the user - see https://longhorn.io/docs/1.7.0/v2-data-engine/features/selective-v2-data-engine-activation/

We can do this in the Harvester GUI by editing a node to add the label. I haven't proposed any extra special UI for it.

pkg/controller/master/setting/handler.go
w13915984028: Suppose this case:
(1) User enable V2 from harvester UI
(2) This setting is successfully synced and processed by LH
(3) Someone disable V2 from LH setting CRD object directly
anyway to correct LH again?

If we plan to leave this case blank, it is also ok to comment it on document.

tserong: Ah, OK, I understand :-) I'll have to test that, but I suspect you're right and I need to add this setting to skipHashCheckSettings and/or document it.

pkg/controller/master/setting/nodeconfig.go
w13915984028: (1)Suppose all nodes have enough resources for enabling V2
(2) User clicks `enable`, no error,
if it takes time to enable on all nodes, before it is succesfully done/failed, user can't click `disable`, or the `enable` menu is grayed.

The current code writes enable/disable to LH setting, does LH already have those check and protection, and will refuse to change in the middle? thanks.

w13915984028: Definitely, when it is not perfect, we can document it and enhance later.

w13915984028 (approved):
LGTM, thanks.

BTW: Though I input some questions, but they can be documented/enhanced later. They do not affect this PR to be merged.

bk201:
@mergifyio backport v1.4

Issue #4005, opened 2023-05-31
[BUG] When upgrading single node VMs are all turned off after upgrade


noahgildersleeve:
**Describe the bug**

When you upgrade a single node it will turn off all VMs
**To Reproduce**
1. Create a few VMs
2. Turn on VMs
3. Upgrade Harvester cluster
4. Check status of VMs
**Expected behavior**

I'm unsure, but there was some talk of having the VMs going back to previous state
**Support bundle**

[supportbundle_4561a7d5-0ffa-401b-80c4-37c54df69b98_2023-05-30T21-52-53Z.zip](https://github.com/harvester/harvester/files/11607375/supportbundle_4561a7d5-0ffa-401b-80c4-37c54df69b98_2023-05-30T21-52-53Z.zip)

**Environment**
 - Harvester ISO version: v1.1.2 -> v1.2.0 upgrade
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): tested on single node DL160

**Additional context**

w13915984028:
Test setting:
single-node v1.3.2 cluster, 3 VMs, 1 running, 1 paused, 1 stopped;  upgrade to harvester-v1.4.0-dev-20240918

When upgrade begins and is running in early stages, the VMs keep unchanged
![image](https://github.com/user-attachments/assets/732685ba-f84b-4dec-9975-1c5c1a6c03b1)

![image](https://github.com/user-attachments/assets/ca99c819-ae9f-4773-80ac-1bcd1666dab5)


Test report:
1. Upgrade without setting `"restoreVM": true`

After upgrade: All VMs are `Off`

![image](https://github.com/user-attachments/assets/d2ac5d12-a0d7-40d2-8b95-e39c7c836039)


2. Upgrade with setting `"restoreVM": true`

Upgrade path is: v1.3.2 -> harvester-v1.4.0-dev-20240918 -> harvester-master-head-20240920

Observed issue: https://github.com/harvester/harvester/issues/6646

after upgrade: not possible to check if VMs are recovered as upgrade failed.

3. Upgrade with setting `"restoreVM": true`
Report: https://github.com/harvester/harvester/issues/4005#issuecomment-2376372609

brandboat:
[...]

PR #6426, opened 2024-08-27
Restore running VMs after upgrade in single node cluster

brandboat:
**Problem:**
When you upgrade a single node it will turn off all VMs. And there was some talk of having the VMs going back to previous state.

**Solution:**
Add new setting `restoreVMs` to `UpgradeConfig`. This setting trigger harvester to store running vms namespace/name to secret before upgrade, and start those VMs based on the info stored in secret after upgrade. Note this setting only works when cluster is single node. **About the paused VMs, we leave them to stopped state.**

**Related Issue:**
https://github.com/harvester/harvester/issues/4005

**Test plan:**
1. Create iso based on this pr/branch
2. Use ipxe to create single node cluster, 
3. Create 4 VMs with 2 running VMs, 1 stopped VM, and 1 paused VM. 
  image
4. Apply the following yaml to update `restoreVM: true` in Setting upgrade-config
  ```yaml
  value: '{"imagePreloadOption":{"strategy":{"type":"sequential"}}, "restoreVM": true}'
  kind: Setting
  metadata:
    name: upgrade-config
  ```
5. Use the following yaml to upgrade harvester, note that the iso in isoURL should be the same one build in step1.
  ```yaml
  kind: Version
  metadata:
    name: master
    namespace: harvester-system
  spec:
  ---
  kind: Upgrade
  metadata:
    annotations:
    generateName: hvst-upgrade-
    namespace: harvester-system
  spec:
    version: "master"
    logEnabled: false
  ```
6. Verify all VMs were running before upgrade are still in running state after upgrade. And the paused vm before upgrade are stopped after upgrade.
image

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -2,6 +2,7 @@ package upgrade
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -12,6 +13,7 @@ import (
 	provisioningv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	provisioningctrl "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
+	"github.com/rancher/wrangler/v3/pkg/condition"
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/batch/v1"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
@@ -20,7 +22,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
+	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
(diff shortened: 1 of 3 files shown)


bk201 (approved):
lgtm. cc PoAn to take a look too.

FrankYang0529 (commented):
The code part LGTM. However, I need to wait for https://github.com/harvester/harvester/pull/6349 to do the test, because master to master upgrade doesn't work now.

FrankYang0529:
> The code part LGTM. However, I need to wait for #6349 to do the test, because master to master upgrade doesn't work now.

Sorry, ignore this. I can do the upgrade without logging.

FrankYang0529 (approved):

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -192,6 +213,9 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 
 	// clean upgrade repo VMs and images if a upgrade succeeds.
 	if harvesterv1.UpgradeCompleted.IsTrue(upgrade) {
+		// try to restore vm to the state before upgrade
+		h.restoreVMState(upgrade)
+
 		// try to clean up images before purging the repo VM
 		_, exists := upgrade.Annotations[imageCleanupPlanCompletedAnnotation]
 		if exists {
```
starbops: Just a suggestion: Could we move the `restoreVMState` call after the image-cleanup annotation check? That way, the function would only be called once. It is called twice currently, and there will be a "Get secret failed" message logged.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -723,3 +750,199 @@ func (h *upgradeHandler) setReplicaReplenishmentValue(value int) error {
...
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: harvesterSystemNamespace,
+			Name:      preUpgradeVMsSecretName,
+		},
+		Data: map[string][]byte{
+			"vms": runningVMsJSON,
+		},
+	}
```
starbops: Do we want to add an owner reference for the Secret? If the user removes the Upgrade object in the middle of an upgrade, the Secret without the Upgrade object as its owner will remain on the system.

starbops (approved):
The functionality LGTM. I have only two suggestions, both are optional since you already had error-prone paths covered. Thank you.

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Make sense to me, will add the owner reference. Thanks !

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Sure, thank you.

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Already addressed all comments in the latest commit.

brandboat:
@Mergifyio backport v1.4

PR #6438, opened 2024-08-28
Add setting AdditionalGuestMemoryOverheadRatio to tune guest os memory overhead

w13915984028:
PR changes:

(1) Add setting AdditionalGuestMemoryOverheadRatio to tune guest os memory overhead
(2) Split settings.go, which is mixed with global defintions and each setting related structures
(3) Replace the `default` and `value` string with consts

**Problem:**

The OOM happened from time to time on some VMs, a general reason is that the VM `overhead` memory is not enough. Refer: https://github.com/w13915984028/harvester-develop-summary/blob/main/oom-related-issue-investigation.md#oom-related-issue-investigation

This PR targets to add more options to set a proper memory overhead for VM.

--

Harvester set a reserved memory, which is in such form, e.g:

Users specify `2Gi` memory limit for a VM, Harvester sets the `memory.guest` to `1948Mi`, and `100Mi` memory is reserved for various of overhead, if users do not set it explicitly on the Harvester UI.

But in our test, the 100Mi can't fit all scenarios.
```
        machine:
          type: q35
        memory:
          guest: 1948Mi
        resources:
          limits:
            cpu: "1"
            memory: 2Gi
          requests:
            cpu: 62m
            memory: 1365Mi
```

**Solution:**

1. Add a new Harvester setting `additional-guest-memory-overhead-ratio`, user can set it in range [1..20]; this is a global setting, affects all new creating VMs.

It means, after the underlay kubevirt computes the overhead of a VM, it will further use this `additional-guest-memory-overhead-ratio` to adjust the value.

e.g. `additional-guest-memory-overhead-ratio` = 1.5

A VM is computed to have 200Mi overhead, then following function will do `200Mi * 15` to get a new value `300Mi` as the overhead.

```
// GetMemoryOverhead computes the estimation of total
// memory needed for the domain to operate properly.
// This includes the memory needed for the guest and memory
// for Qemu and OS overhead.
// The return value is overhead memory quantity
//
// Note: This is the best estimation we were able to come up with
//
//	and is still not 100% accurate
func GetMemoryOverhead(vmi *v1.VirtualMachineInstance, cpuArch string, additionalOverheadRatio *string) resource.Quantity {
```
https://github.com/kubevirt/kubevirt/blob/021e9b42f79b098fc50e76062e5388e13d31e235/pkg/virt-controller/services/renderresources.go#L303

2. The Harvester existing feature `Reserved Memory` still works, with a bit differences comparing to before:

- If `additional-guest-memory-overhead-ratio` is not set, and user does not specify a value for `Reserved Memory`, the default value `100Mi` will be used;
- If `additional-guest-memory-overhead-ratio` is set, and user does not specify a value for `Reserved Memory`, then the default `100Mi` will NOT be used.
- If `additional-guest-memory-overhead-ratio` is set, and user does specify a value for `Reserved Memory`, then this value is used. The related VM will get both overhead from `AdditionalGuestMemoryOverheadRatio` and `Reserved Memory`.

Those options bring more flexibility for user usage. And also keep aligned with upstream.

**Related Issue:**
https://github.com/harvester/harvester/issues/5768

The original PR https://github.com/harvester/harvester/pull/6177 is replaced by this PR.

Document PR: https://github.com/harvester/docs/pull/615
UI PR: https://github.com/harvester/dashboard/pull/1128

**Test plan:**

A tested form for reference:

image

(1) Clear setting.harvesterhci.io `additional-guest-memory-overhead-ratio`, let both `value` and `default` be empty. 

(1.1) Explicitly set a value of `Reserved Memory` on the UI when creating VM, the created VM object will have such information. e.g. 256 Mi
```
        memory:
          guest: 1792Mi  // the expected memory - reserved memory (256 Mi)
        resources:
          limits:
            cpu: "1"
            memory: 2Gi  // the expected  memory
```

Get the information from the backing pod of this vm:
```
virt-launcher-vm...

    resources:
      limits:
        cpu: "1"
        devices.kubevirt.io/kvm: "1"
        devices.kubevirt.io/tun: "1"
        devices.kubevirt.io/vhost-net: "1"
        memory: "2404034561" // this field is the memory limit upon the container which carries the VM
...
```

(1.2) Don't set `Reserved Memory` on the UI when creating VM, the created VM object will have such information
```
        memory:
          guest: 1948Mi  // the expected memory - default reserved memory (100Mi)
        resources:
          limits:
            cpu: "1"
            memory: 2Gi  // the expected  memory
```

Get the information from the backing pod of this vm:
...

(2) Set setting.harvesterhci.io `additional-guest-memory-overhead-ratio`, let `value`  be string "1.5".
```
$kubectl get settings.harvesterhci.io additional-guest-memory-overhead-ratio -oyaml
kind: Setting
...
value: "1.5"

After setting, this value is synced to kubevirt automatically
$ kubectl get kubevirt kubevirt -n harvester-system -oyaml | grep overheadratio -i
    additionalGuestMemoryOverheadRatio: "1.5"
```

(2.1) Explicitly set a value of `Reserved Memory` on the UI when creating VM, the created VM object will have such information. e.g. 256 Mi

```
        memory:
          guest: 1792Mi  // the expected memory - reserved memory (256 Mi)
        resources:
          limits:
            cpu: "1"
            memory: 2Gi  // the expected  memory
```

Get the information from the backing pod of this vm:
...

(2.2) Don't set `Reserved Memory` on the UI when creating VM, the created VM object will have such information
```
        memory:
          guest: 2Gi  // the expected memory 
        resources:
          limits:
            cpu: "1"
            memory: 2Gi  // the expected  memory
```

Get the information from the backing pod of this vm:
...

(3) Set setting.harvesterhci.io `additional-guest-memory-overhead-ratio`, let `value`  be string "2" and repeat the test.


w13915984028:
@ibrokethecloud @bk201 @FrankYang0529 @votdev 

The previous PR https://github.com/harvester/harvester/pull/6177 was replaced by this one. Please continue to review this new PR, thanks.

A new setting `additional-guest-memory-overhead-ratio` is added to sync with upstream kubevirt `Spec.Configuration.AdditionalGuestMemoryOverheadRatio`; with allowed range [1..10].

KubeVirt computes a value `~240Mi` as guest overhead based on many parameters of the VM.

Now we set the default value of `additional-guest-memory-overhead-ratio` to `1.5`; if user sets nothing, each VM will get an additional `~120Mi+` overhead, which replaces the Harvester original `100Mi` default reserved memory.  
One advantage is: the guest OS gets the exact requested memory. Not the `target memory -100Mi`.

Some test resultes are posted on a table: https://github.com/harvester/harvester/pull/6438#issue-2491832842

User can tune the system by:
(1) Set the (system level) setting `additional-guest-memory-overhead-ratio`, each VM benefits from it; defaults to "1.5", will suggest user to set a higher value if they encounter OOM;
(2) Set the `Reserved Memory` for some VMs to get more addtional overhead;

The effection of a VM's guest OS memory is:
- When `additional-guest-memory-overhead-ratio` is **NOT** set, everything works as before:
  - and user does **NOT** specify a value for `Reserved Memory`, `guest OS memory = requested memory - 100Mi`
  - and user does specify a value for `Reserved Memory`, `guest OS memory = requested memory - Reserved Memory`
- When `additional-guest-memory-overhead-ratio` is set, kubevirt adds ratio based `addtional overhead` memory to POD limit:
  - and user does **NOT** specify a value for `Reserved Memory`, `guest OS memory = requested memory`;  
  - and user does specify a value for `Reserved Memory`, `guest OS memory = requested memory - Reserved Memory`. 
  
  The document PR will be updated in the coming days.

w13915984028:
@mergify backport v1.4

w13915984028:
code is rebased to solve conflicts on settings

pkg/controller/master/setting/additional_guest_memory_overhead.go
```diff
@@ -0,0 +1,71 @@
...
+		return fmt.Errorf("fail to get kubevirt object %v/%v", util.HarvesterSystemNamespaceName, util.KubeVirtObjectName)
+	}
+	// kubevirt already has AdditionalGuestMemoryOverheadRatio
+	if kubevirt.Spec.Configuration.AdditionalGuestMemoryOverheadRatio != nil {
+		// clear current
+		if agmorc.IsEmpty() {
+			kubevirtCpy := kubevirt.DeepCopy()
+			kubevirtCpy.Spec.Configuration.AdditionalGuestMemoryOverheadRatio = nil
+			if _, err := h.kubeVirtConfig.Update(kubevirtCpy); err != nil {
+				return fmt.Errorf("fail to update %v as nil to kubevirt %w", overheadConfigName, err)
```
votdev:
```suggestion
				return fmt.Errorf("failed to update %v as nil to kubevirt %w", overheadConfigName, err)
```

pkg/controller/master/setting/additional_guest_memory_overhead.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+			return nil
+		}
+		// update with new value
+		if *kubevirt.Spec.Configuration.AdditionalGuestMemoryOverheadRatio != value {
+			kubevirtCpy := kubevirt.DeepCopy()
+			kubevirtCpy.Spec.Configuration.AdditionalGuestMemoryOverheadRatio = &value
+			if _, err := h.kubeVirtConfig.Update(kubevirtCpy); err != nil {
+				return fmt.Errorf("fail to update %v as %v to kubevirt %w", overheadConfigName, value, err)
```
votdev:
```suggestion
				return fmt.Errorf("failed to update %v as %v to kubevirt %w", overheadConfigName, value, err)
```

pkg/controller/master/setting/additional_guest_memory_overhead.go
```diff
@@ -0,0 +1,71 @@
...
+	// kubevirt has no AdditionalGuestMemoryOverheadRatio
+	// setting value is empty string or zero value, do nothing
+	if agmorc.IsEmpty() {
+		return nil
+	}
+	// create new AdditionalGuestMemoryOverheadRatio
+	kubevirtCpy := kubevirt.DeepCopy()
+	kubevirtCpy.Spec.Configuration.AdditionalGuestMemoryOverheadRatio = &value
+	if _, err := h.kubeVirtConfig.Update(kubevirtCpy); err != nil {
+		return fmt.Errorf("fail to update %v as %v to kubevirt %w", overheadConfigName, value, err)
```
votdev:
```suggestion
		return fmt.Errorf("failed to update %v as %v to kubevirt %w", overheadConfigName, value, err)
```

votdev (changes requested):

w13915984028:
FYI:
Document PR: https://github.com/harvester/docs/pull/615

document preview: https://66d61b2b7a08694d93628b0f--harvester-preview.netlify.app/v1.4/advanced/index#additional-guest-memory-overhead-ratio

UI PR: https://github.com/harvester/dashboard/pull/1128

pkg/controller/master/setting/additional_guest_memory_overhead.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// sync with  kubevirt configuration
+	kubevirt, err := h.kubeVirtConfigCache.Get(util.HarvesterSystemNamespaceName, util.KubeVirtObjectName)
+	if err != nil {
+		return fmt.Errorf("fail to get kubevirt object %v/%v", util.HarvesterSystemNamespaceName, util.KubeVirtObjectName)
+	}
+	// kubevirt already has AdditionalGuestMemoryOverheadRatio
+	if kubevirt.Spec.Configuration.AdditionalGuestMemoryOverheadRatio != nil {
+		// clear current
+		if agmorc.IsEmpty() {
```
ibrokethecloud, line 39: the default additional overhead will never be empty because when setting does have not a custom value, we switch to default of 1.5

a user who may have manually updated their configuration, will end up having this changed to 1.5 which may not be desirable.

may be we should leave the default value to an empty string?

w13915984028: Good questions & suggestions :)

(1) For newly installed cluster or upgraded cluster, if user does not touch this setting; the default "1.5" is designed to replace the legacy "100Mi" reserved memory.

(2) When the `Default` is `""`, this feature means to be disabled, and user needs to set it manually to get this feature.

Roughly, (1) will benefit more users.


(3) If a user doesn't want this setting but fully fallback to legacy `Reserved Memory` mode, he can set this setting as "0"; this is described in the document PR.

image

ibrokethecloud: in the kubevirt spec AdditionalGuestMemoryOverheadRatio defaults to 1.

If say a user had a kubevirt config modified with this set, the code block will find `kubevirt.Spec.Configuration.AdditionalGuestMemoryOverheadRatio` to be not nil.

In subsequent step `agmorc.IsEmpty()` is never going to be empty since we are using `Default` of 1.5 when a `Value` is not provided in the setting, which will trigger the update the wipe the user defined ration and set it to 1.5 which may not be desired.

image

w13915984028: When Harvester expose this option to setting, we generally assume the setting is the source of truth, changes to underlayer will be overwritten.

I will add this to document, if users has used to set this on kubevirt object directly, then user is suggested to set on Harvester settings later.

On the upgrade path, will also add another PR to convert the potential user set value to Harvester setting.

w13915984028: A third commit is added to convert the potential user configuration on kubevirt to Harvester setting; the document PR is also updated.

Sure, Kubevirt defaults it to `"1"`;  but Harvester defaults the setting to `"1.5"`, is mainly due to the `"1"` will have a higher chance of VM OOM; in the past we recommend user to configure the `Reserved Memory` with a big value; now we set it as `"1.5"` as default and recommend user to set it as `"2"`.

Some users tend to not reading the document until they hit issues :),  a higher `Default` value is also for those users.

And, for Harvester settings, the `Default` field can also be changed/deleted (revert to empty) via `kubectl`, unless the webhook deny it explicitly. At the moment, this trick is not documented.

bk201:
Thanks for the update. In the previous PR, I have [a question](https://github.com/harvester/harvester/pull/6177/files#r1687395966) on how we determine those values are proper (if there is benchmark data?). The change in this PR will make me think about how the user decides a proper `additional-guest-memory-overhead-ratio` value. Do we have a recommendation for the users?

w13915984028:
> Thanks for the update. In the previous PR, I have [a question](https://github.com/harvester/harvester/pull/6177/files#r1687395966) on how we determine those values are proper (if there is benchmark data?). The change in this PR will make me think about how the user decides a proper `additional-guest-memory-overhead-ratio` value. Do we have a recommendation for the users?

A table is attached at the beginning of this PR https://github.com/harvester/harvester/pull/6438#issue-2491832842. Which has tests on different scenarios.

In the document PR, it is described as: https://github.com/harvester/docs/pull/615/files#diff-c94ca771ce63719c3773b2b86d9a54cf438884580f371e8873d1be4424293ce6R354
```

To reduce the chance of hitting OOM, Harvester suggests to:

- Configure this setting with value `"2"` to give all the VMs more guest memory overhead.

- Configure the `Reserved Memory` for some important VMs to give them even more memory overhead.

- Avoid configuring the `spec.configuration.additionalGuestMemoryOverheadRatio` field of `kubevirt` object directly.
```

ibrokethecloud (commented):
overall the change looks good to me, we probably need to document the change in behaviour for existing users.

In my tests with a 2core/4GB VM, on an existing 1.3.1 cluster due to our patching of guest memory and kubevirts own reservation logic, we end up reserving approximately 355MB from the pod memory limit.

With the change enabled and setting defined to 1.5, reserved memory goes upto 382MB from the pod upper limit, while the guest sees the full 4G of memory since we do not tweak the guest memory setting

With the change enabled and setting defined to 1.0, reserved memory goes down to 255MB from the pod upper limit, while guest sees the full 4G of memory. 

Users may need to tweak the guest limit manually or unset it to ensure that the webhook mutation recalculates the guest memory definition.

bk201:
> To reduce the chance of hitting OOM, Harvester suggests to:
> - Configure this setting with value `"2"` to give all the VMs more guest memory overhead.

Why do we suggest the value `2` here? 
And do you have the overhead value if a guest VM has huge memory like 64G? I want to check the ratio works for both small-footprint VMs and large VMs.

@bk201 

(1) The guest OS memory size has little influence on the memory overhead. Roughly, 64G guest OS memory has `16Mi` more overhead; `512Gi -> 128 Mi` for pagetable.  ( I will add this to the document)

https://github.com/kubevirt/kubevirt/blob/aa117b4c9413eab9fef9680ca75546d613c0bfdf/pkg/virt-controller/services/renderresources.go#L310
```
// Add the memory needed for pagetables (one bit for every 512b of RAM size)
```

(2) The ratio `2` will let any VM has roughly >= 480 Mi of overhead, in production, most VM will have 2Gi or 4Gi or more memory; the `480 Mi` overhead can reduce OOM massively. If the cluster has no resource pressure, this setting will be better.

I will aslo add such data to the document.

```
vm1: 2G memory, 0 Mi reserved memory

overhead: 2401517569 - 2GI = 234 Mi

    annotations:
      harvesterhci.io/reservedMemory: 0Mi
          memory:
            guest: 2Gi
          resources:
            limits:
              cpu: "1"
              memory: 2Gi
            requests:
              cpu: 62m
              memory: 136Mi

pod1 - container limit: 2401517569

      name: compute
      resources:
        limits:
          cpu: "1"
          devices.kubevirt.io/kvm: "1"
          devices.kubevirt.io/tun: "1"
          devices.kubevirt.io/vhost-net: "1"
          memory: "2401517569"
        requests:
          cpu: 62m
          devices.kubevirt.io/kvm: "1"
          devices.kubevirt.io/tun: "1"
          devices.kubevirt.io/vhost-net: "1"
          ephemeral-storage: 50M
          memory: "396640257"



vm2: 64G memory, 0 Mi reserved memory

overhead: 67365410Ki - 64Gi = 250 Mi

    annotations:
      harvesterhci.io/reservedMemory: 0Mi
          resources:
            limits:
              cpu: "1"
              memory: 64Gi
            requests:
              cpu: 62m
              memory: 4369Mi        


pod1 - container limit: 67365410Ki

      name: compute
      resources:
        limits:
          cpu: "1"
          devices.kubevirt.io/kvm: "1"
          devices.kubevirt.io/tun: "1"
          devices.kubevirt.io/vhost-net: "1"
          memory: 67365410Ki
        requests:
          cpu: 62m
          devices.kubevirt.io/kvm: "1"
          devices.kubevirt.io/tun: "1"
          devices.kubevirt.io/vhost-net: "1"
          ephemeral-storage: 50M
          memory: 4730402Ki
```

w13915984028:
code is rebased to sync with master-head

w13915984028:
> overall the change looks good to me, we probably need to document the change in behaviour for existing users.
> 
> In my tests with a 2core/4GB VM, on an existing 1.3.1 cluster due to our patching of guest memory and kubevirts own reservation logic, we end up reserving approximately 355MB from the pod memory limit.
> 
> With the change enabled and setting defined to 1.5, reserved memory goes upto 382MB from the pod upper limit, while the guest sees the full 4G of memory since we do not tweak the guest memory setting
> 
> With the change enabled and setting defined to 1.0, reserved memory goes down to 255MB from the pod upper limit, while guest sees the full 4G of memory.
> 
> Users may need to tweak the guest limit manually or unset it to ensure that the webhook mutation recalculates the guest memory definition.

@ibrokethecloud Thanks. Yes, there is no easy & one-fit-all solution. We will try best to describe those as clear as possible and help the user to make the final decision.

w13915984028:
code is rebased to sync with master-head

ibrokethecloud (approved):
lgtm. thanks.

FrankYang0529 (approved):
LGTM.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -235,15 +248,44 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity,
...
+			return patchOps, fmt.Errorf("annotation %v can't be converted to memory unit", vm.Annotations[util.AnnotationReservedMemory])
+		}
+		// already checked on validator: reservedMemory < mem
+		if reservedMemory.Value() >= mem.Value() {
+			return patchOps, fmt.Errorf("reservedMemory can't be equal or greater than limits.memory  %v %v", reservedMemory.Value(), mem.Value())
+		}
+	} else {
+		// if user has set a valid AdditionalGuestMemoryOverheadRatio value
+		if agmorc != nil && !agmorc.IsEmpty() {
+			useReservedMemeory = false
```
bk201:
```suggestion
			useReservedMemory = false
```

bk201 (approved):
Thanks. The PR's mechanism LGTM.

> (2) The ratio 2 will let any VM has roughly >= 480 Mi of overhead, in production, most VM will have 2Gi or 4Gi or more memory; the 480 Mi overhead can reduce OOM massively

While it's nice to have the feature to set the overhead ratio globally, I still believe we need more data to support why the value 2 is recommended in the doc.

I understand a higher ratio increase the reserved memory and reduce the chance to hit OOM. But this doesn't clarify why the chance is reduce because there is no data to support it. I'd suggest having some benchmark and the benchmark could contain variables that influence the memory usage in the long run.

w13915984028:
Solve last comment https://github.com/harvester/harvester/pull/6438#discussion_r1753519053 and squash commits for final code commiting.

w13915984028:
> Thanks. The PR's mechanism LGTM.
> 
> > (2) The ratio 2 will let any VM has roughly >= 480 Mi of overhead, in production, most VM will have 2Gi or 4Gi or more memory; the 480 Mi overhead can reduce OOM massively
> 
> While it's nice to have the feature to set the overhead ratio globally, I still believe we need more data to support why the value 2 is recommended in the doc.
> 
> I understand a higher ratio increase the reserved memory and reduce the chance to hit OOM. But this doesn't clarify why the chance is reduce because there is no data to support it. I'd suggest having some benchmark and the benchmark could contain variables that influence the memory usage in the long run.

@bk201 I will add this to the document PR, nice to have benchmark, but honestly, OOM is not something we can easily to benchmark, instead, we suggest user to keep monitoring the cluster, and adjust this setting based on the observation. Users have a solution to solve some sort of un-expected issues. thanks.

votdev (approved):

Issue #6440, opened 2024-08-28
[BUG] http-proxy doesn't allow it to be reset w/ call-to-action/button "Use Default Value"


irishgordo:
**Describe the bug**
When setting the `http-proxy` in the Harvester Settings.  Once set.
Then when attempting to "unset" the "http-proxy" w/ "Use Default Value" (to empty out the fields) -> the admission webhook still says "must set node/clustrer ip in noProxy".
Using the call-to-action / "Use Default Value" --> causes admission webhook errors

**To Reproduce**
1. Go to http-proxy in settings
2. Set it to something for like http:// and then also set the node ip under "noProxy"
3. allow it to get set
4. go back to the setting
5. Click call-to-action / "use Default Value" button, to "empty out fields" in form, attempt to save -> error is thrown

**Expected behavior**
Being able to reset / empty out setting for http-proxy

**Environment**
 - Harvester ISO version: v1.3.2-rc2
 - Underlying Infrastructure: bare-metal

irishgordo:
Thanks for this fix @votdev :smile: :+1: 
This looks and works great in v1.4.0-rc2.
Validated http-proxy was working to acquire an image.
Then reset to default value with call-to-action/button.
And it all looks great :+1: 


https://github.com/user-attachments/assets/59f0e4e5-f1b5-4f73-b5b2-5caefeb58e9c

PR #6447, opened 2024-08-29
http-proxy doesn't allow it to be reset to its default value

votdev:
**Problem:**
The setting `http-proxy` in the Harvester Settings can not be reset to the configured defaults by pressing `Use the default value` in the UI.

**Solution:**
Check if the value to be set equals the configured default value. If yes, abort validation and exit without error.

**Related Issue:**
https://github.com/harvester/harvester/issues/6440

**Test plan:**
1. Go to the advanced settings page.
2. Go to `http-proxy` and select the `Edit Setting` action menu.
3. Add the CIDR that fits your cluster, e.g. `192.168.0.0/24`. Press `Save`.
4. Edit the setting again and press the `Use the default value` button. Press `Save`.
5. No error should appear and the browser should be redirected to the previous settings page.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -241,11 +241,14 @@ func validateHTTPProxyHelper(value string, nodes []*corev1.Node) error {
 		return err
 	}
 
-	// Make sure the node's IP addresses is set in 'noProxy'. These IP
-	// addresses can be specified individually or via CIDR address.
-	err := validateNoProxy(httpProxyConfig.NoProxy, nodes)
-	if err != nil {
-		return err
+	// Make sure the node's IP addresses are set in `NoProxy` if `HTTPProxy`
+	// or `HTTPSProxy` are configured. These IP addresses can be specified
+	// individually or via CIDR address.
+	if httpProxyConfig.HTTPProxy != "" || httpProxyConfig.HTTPSProxy != "" {
+		err := validateNoProxy(httpProxyConfig.NoProxy, nodes)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -262,8 +262,11 @@ func (v *settingValidator) validateHTTPProxy(setting *v1beta1.Setting) error {
 		}
 	}
 
-	if err := validateHTTPProxyHelper(setting.Value, nodes); err != nil {
-		return werror.NewInvalidError(err.Error(), settings.KeywordValue)
+	// Validate the value only if it is not the default value.
+	if setting.Value != setting.Default {
+		if err := validateHTTPProxyHelper(setting.Value, nodes); err != nil {
+			return werror.NewInvalidError(err.Error(), settings.KeywordValue)
+		}
 	}
 	return nil
 }
--- a/pkg/webhook/resources/setting/validator_test.go
+++ b/pkg/webhook/resources/setting/validator_test.go
@@ -534,6 +534,114 @@ func Test_validateNoProxy_2(t *testing.T) {
 	}
 }
 
+func Test_validateHTTPProxyHelper(t *testing.T) {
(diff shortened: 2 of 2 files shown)


votdev:
@Mergifyio backport v1.2 v1.3 v1.4

pkg/webhook/resources/setting/validator.go
```diff
@@ -260,8 +260,11 @@ func (v *settingValidator) validateHTTPProxy(setting *v1beta1.Setting) error {
 		}
 	}
 
-	if err := validateHTTPProxyHelper(setting.Value, nodes); err != nil {
-		return werror.NewInvalidError(err.Error(), "value")
+	// Validate the value only if it is not the default value.
```
w13915984028, line 268: Looks the current code does not handle `nil` well. Both "" and "{}" need to be cared.

The `value` is initialized as "", and `default` as "{}".

The controller side which utilize this setting, should also tolerant  `""`, the `default` can also be an empty string.

```
func validateHTTPProxyHelper(value string, nodes []*corev1.Node) error {
	if value == "" || value == "{}" {
		return nil
	}

	httpProxyConfig := &util.HTTPProxyConfig{}
	if err := json.Unmarshal([]byte(value), httpProxyConfig); err != nil {
		return err
	}
	// Make sure the node's IP addresses is set in 'noProxy'. These IP
	// addresses can be specified individually or via CIDR address.
	err := validateNoProxy(httpProxyConfig.NoProxy, nodes)
	if err != nil {
		return err
	}
	return nil
}

func (v *settingValidator) validateHTTPProxy(setting *v1beta1.Setting) error {
	nodes, err := v.nodeCache.List(labels.Everything())
	if err != nil {
		return werror.NewInternalError(err.Error())
	}

	if err := validateHTTPProxyHelper(setting.Default, nodes); err != nil {
		return werror.NewInvalidError(err.Error(), "default")
	}

	if err := validateHTTPProxyHelper(setting.Value, nodes); err != nil {
		return werror.NewInvalidError(err.Error(), "value")
	}

	return nil
}
```

votdev: Your suggestion does not work because it does not abort with an error for a `Value` with `{}`, since in this case `noProxy` does not meet the conditions that the node IPs must be specified.

This happens because in the case of 
```
func validateHTTPProxyHelper(value string, nodes []*corev1.Node) error {
	if value == "" || value == "{}" {
		return nil
	}
	...
```
the function immediately returns with `nil` instead of raising an error at
```
	err := validateNoProxy(httpProxyConfig.NoProxy, nodes)
	if err != nil {
		return err
	}
```

I see another problem, but it already exists before my changes. If `Default` and `Value` are an empty string no error is thrown. This is actually wrong because it is not a valid JSON. And since this setting contains JSON code, an error should be thrown if it is not. Or am I overlooking a general rule here that affects the implementation of settings and their behavior?

w13915984028: When `func (h *Handler) syncHTTPProxy(setting *harvesterv1.Setting) error {` treats `""` as `"{}"`, it will be more robust.

The setting itself is defined as `string`,`""` is the initial value;  user may also delete the validator hook to forcely update value/default as `""`.

votdev: > When `func (h *Handler) syncHTTPProxy(setting *harvesterv1.Setting) error {` treats `""` as `"{}"`, it will be more robust.
> 
> The setting itself is deined as `string`,`""` is the initial value; user may also delete the validator hook to forcely update value/default as `""`.

I will open a new issue for that.

votdev: The related enhancement issue is https://github.com/harvester/harvester/issues/6473 and the already merged PR is https://github.com/harvester/harvester/pull/6478.

w13915984028 (approved):
LGTM, thanks.

FrankYang0529 (commented):
Overall LGTM. There is conflict. Could you resolve it? Thanks.

FrankYang0529 (approved):

Issue #6432, opened 2024-08-28
[BUG]  v1.3.1 -> v1.3.2-rc2 upgrade fail on 2 nodes clusters with customized default storage class and VM


albinsun:
**Describe the bug**
v1.3.1 -> v1.3.2-rc2 upgrade fail on 2 nodes clusters with customized default storage class and VM

**To Reproduce**
1. <details>
   <summary> Fresh install 2 nodes <code>harvester-v1.3.1</code>
   </summary>
   
   ![image](https://github.com/user-attachments/assets/a4e8a5ce-4985-4cd0-9d69-20e99df55593)
   </details>

1. <details>
   <summary> Create storage class <code>sc2</code> and <b>Set as Default</b>
   </summary>
   
   * Create
     ![image](https://github.com/user-attachments/assets/6c9a9ce7-1c93-40df-b979-c6df5f0d8a8a)
   * Set as Default
     ![image](https://github.com/user-attachments/assets/d41bc4a4-bde8-4a4b-a20a-75e0638d6769)
   </details>

1. <details>
   <summary> Create a VM 
   </summary>
   
   * VM
     ![image](https://github.com/user-attachments/assets/42140442-ea67-4a70-8634-c3fb30673c4c)
     ![image](https://github.com/user-attachments/assets/77d17d1a-06b4-4f11-9304-41718fca4bbf)

   * Longhorn
     ![image](https://github.com/user-attachments/assets/16160d90-9734-46f3-8703-1bff3ac32955)

   </details>

1. <details>
   <summary> Trigger upgrade to <code>harvester-v1.3.2-rc2</code>
   </summary>
   
   * Create entry
     ```
     rancher@harvester-node-0:~> sudo -i
     harvester-node-0:~ # kubectl create -f http://192.168.123.1:8080/harvester-v1.3.2-rc2-version-amd64.yaml
     version.harvesterhci.io/v1.3.2-rc2 created
     ```
   * Trigger
     ![image](https://github.com/user-attachments/assets/2909c356-276a-41e2-874d-22d1edcb8543)
     ![image](https://github.com/user-attachments/assets/79f90b05-688a-4952-ae22-d161bbf7c979)

   </details>

1. <details open>
   <summary> Successfully upgrade to <code>harvester-v1.3.2-rc2</code> :x: 
   </summary>
   
   * Nginx down
     ![image](https://github.com/user-attachments/assets/8ee478e6-b7cc-463d-9d3f-204cb8f3bc5a)
[...]

PR #6461, opened 2024-08-30
Additional webhook taints

ibrokethecloud:

**Problem:**

With the bump to rke2 v1.27.12 2 node upgrades are currently failing. This is being caused due to the agent node in some cases loosing connectivity with the api server loadbalancer and not recovering from the same. 

The kubelet is unable to reconnect with the api server and messages similar to the one below are noticed

```
Trace[1756469586]: [10.002071408s] [10.002071408s] END
E0831 10:23:06.781356   16320 reflector.go:148] object-"kube-system"/"rke2-coredns-rke2-coredns": Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: Get "https://127.0.0.1:6443/api/v1/namespaces/kube-system/configmaps?fieldSelector=metadata.name%3Drke2-coredns-rke2-coredns&resourceVersion=80254": net/http: TLS handshake timeout
```

This loss of connectivity results in the node being tainted as unreachable. This eventually results in the harvester webhook pods being evicted from the agent node. Since there are no webhook pods running, the harvester-validator validatingwebhookconfiguration results in api calls to update node status to fail, resulting in a deadlock where the cluster is stuck in upgrade and unable to recover.

**Solution:**

Additional webhook taints to allow webhook to be scheduled during 2 node upgrade to v1.3.2

With this change if the upgrade is stuck due to agent being unable to contact the api server, a simple restart of rke2-agent will allow the upgrade to progress.

**Related Issue:**
https://github.com/harvester/harvester/issues/6432
**Test plan:**

### DIFF
--- a/deploy/charts/harvester/templates/deployment.yaml
+++ b/deploy/charts/harvester/templates/deployment.yaml
@@ -160,6 +160,13 @@ spec:
       serviceAccountName: harvester
       affinity:
 {{ include "harvester.apiAffinity" (dict "root" . "component" "webhook-server") | indent 8 }}
+      tolerations:
+      - effect: NoExecute
+        key: node.kubernetes.io/unreachable
+        operator: Exists
+      - effect: NoSchedule
+        key: node.kubernetes.io/unreachable
+        operator: Exists
       containers:
       - env:
         - name: HARVESTER_WEBHOOK_SERVER_HTTPS_PORT


starbops (approved):
LGTM, thanks!

bk201 (approved):

ibrokethecloud:
@mergifyio backport v1.3 v1.4

Issue #6473, opened 2024-09-02
[ENHANCEMENT] Hardening `syncHTTPProxy` function


votdev:
**Is your enhancement related to a problem? Please describe.**
The [syncHTTPProxy](https://github.com/harvester/harvester/blob/3123cb9b11f7185e9218e954212bcb2ae2d945f7/pkg/controller/master/setting/http_proxy.go#L18) function is decoding the setting string into JSON. The setting type is `string` and the `Value` property might be an empty string (which is no valid JSON). This circumstance might cause a decoding error.

**Describe the solution you'd like**
Handle empty strings as `{}`.

**Describe alternatives you've considered**
n/a

**Additional context**
See https://github.com/harvester/harvester/pull/6447#discussion_r1740554200

innobead:
@votdev don't forget to set the milestone. For every item we are working on, need the milestone set. Otherwise, this will be never covered in a release and verified by QA.

cc @bk201 @harvester/qa

albinsun:
Test PASS, try config settings which accept empty value and does not show error, close as done.


## Environment
* `harvester-v1.4.0-rc2`
  * Profile: 2 nodes AMD64 QEMU/KVM
  * ui-source: `Auto`


## Test Case
1. <details>
   <summary> :green_circle: <b>additional-ca</b>
   </summary>
   
   1. Configured
      ![image](https://github.com/user-attachments/assets/a27a0317-221c-4e15-8e6f-85ea1186e9f3)
   2. **Use the default value**
      ![image](https://github.com/user-attachments/assets/833a84dd-8d6b-4594-8365-1fbc42231fa9)
   </details>

1. <details>
   <summary> :green_circle: <b>additional-guest-memory-overhead-ratio</b>
   </summary>
   
   1. Configured
      ![image](https://github.com/user-attachments/assets/0404b59b-20ef-41fd-b10e-40c4303c9e96)
   2. Empty
      ![image](https://github.com/user-attachments/assets/b5ae014b-6958-417d-9484-d12ce16af430)
      adopt default value `1.5`
      ![image](https://github.com/user-attachments/assets/11ed2ee0-bd06-438d-bc77-450c2efa8e5f)

   </details>

1. <details>
   <summary> :green_circle: <b>auto-disk-provision-paths</b>
   </summary>
   
   1. Configured
      ![image](https://github.com/user-attachments/assets/383c9533-229d-4844-a835-2f68c5a7521e)
   2. Empty by **Use the default value**
[...]

PR #6478, opened 2024-09-02
Hardening the `syncHTTPProxy` function

votdev:
**Problem:**
The Harvester settings do not recognize a JSON type, so the setting `http-proxy` is a string. This string can be empty and therefore leads to an error when decoding the string assumed to be JSON. 

**Solution:**
To make the function more robust by using the settings `Default` if the `Value` is empty.

**Related Issue:**
Related to: https://github.com/harvester/harvester/issues/6473

**Test plan:**
n/a

### DIFF
--- a/pkg/controller/master/setting/http_proxy.go
+++ b/pkg/controller/master/setting/http_proxy.go
@@ -18,7 +18,16 @@ const (
 func (h *Handler) syncHTTPProxy(setting *harvesterv1.Setting) error {
 	// Add envs to the backup secret used by Longhorn backups
 	var httpProxyConfig util.HTTPProxyConfig
-	if err := json.Unmarshal([]byte(setting.Value), &httpProxyConfig); err != nil {
+	value := setting.Value
+	if value == "" {
+		value = setting.Default
+		// We need to check again because `Default` is allowed to be empty
+		// as well.
+		if value == "" {
+			value = "{}"
+		}
+	}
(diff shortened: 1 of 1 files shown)


pkg/controller/master/setting/http_proxy.go
```diff
@@ -18,7 +18,13 @@ const (
 func (h *Handler) syncHTTPProxy(setting *harvesterv1.Setting) error {
 	// Add envs to the backup secret used by Longhorn backups
 	var httpProxyConfig util.HTTPProxyConfig
-	if err := json.Unmarshal([]byte(setting.Value), &httpProxyConfig); err != nil {
+	// Convert an empty `Setting.Value` string to its JSON quasi-equivalent,
+	// an empty object, otherwise the decoding will fail.
+	var value = setting.Value
+	if value == "" {
+		value = "{}"
```
votdev: Or should it be like that?
```suggestion
		value = setting.Default
```

w13915984028:
```
// general setting processing
// take the target value from `Value` or `Default` field, depends on if `Value` field is empty
value := setting.Value
if value == "" {
	value = setting.Default
}

// for this setting, empty is equal to and further converted to "{}", avoid JSON Unmarshal failure
if value == "" {
	value = "{}"
}

// ....
```

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

FrankYang0529 (approved):
LGTM.

PR #6517, opened 2024-09-05
Enhance cpu pinning

brandboat:
This pr mainly address comments left in https://github.com/harvester/harvester/pull/6190#pullrequestreview-2280425931

- Do not recreate cpu manager job when update node failed.
- Set node cpu manager status to failed under scenario that job is
  deleted before it is completed/failed.
- In configmap cpu manager script, do not exit when doing kubectl
  get node --show-labels, if error raises, sleep 3s and retry again.
  Also, make all exit code unique
- If container exit code is not 0, leave log when pod terminated
  with exit code > 0 and add exit code to cpu manager job label
  harvesterhci.io/cpu-manager-exit-code

**Problem:**
N/A

**Related Issue:**
#2305 

**Test plan:**

### DIFF
--- a/pkg/controller/master/node/cpu_manager_controller.go
+++ b/pkg/controller/master/node/cpu_manager_controller.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 
 	"github.com/go-errors/errors"
 	catalogv1 "github.com/rancher/rancher/pkg/generated/controllers/catalog.cattle.io/v1"
@@ -16,7 +17,10 @@ import (
 	"github.com/sirupsen/logrus"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/utils/ptr"
 
 	"github.com/harvester/harvester/pkg/config"
@@ -41,6 +45,8 @@ const (
 	CPUManagerWaitLabelTimeoutInSec int64  = 300 // 5 min
 	CPUManagerJobTimeoutInSec       int64  = CPUManagerWaitLabelTimeoutInSec * 2
 	CPUManagerJobTTLInSec           int32  = 86400 * 7 // 7 day
+
+	CPUManagerJobPrefix string = "update-cpu-manager"
 )
 
 type CPUManagerPolicy string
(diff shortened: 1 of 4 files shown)


pkg/controller/master/node/cpu_manager_controller.go
```diff
@@ -152,9 +183,51 @@ func (h *cpuManagerNodeHandler) OnJobChanged(_ string, job *batchv1.Job) (*batch
 		return nil, h.updateCPUManagerStatus(node, job, CPUManagerFailedStatus)
 	}
 
+	// someone delete the cpu manager job
+	if job.DeletionTimestamp != nil {
+		logrus.WithField("job_name", job.Name).Warn("CPU Manager job job was deleted before it finished")
```
w13915984028: two `job` strings

pkg/controller/master/node/cpu_manager_controller.go
```diff
@@ -152,9 +183,51 @@ func (h *cpuManagerNodeHandler) OnJobChanged(_ string, job *batchv1.Job) (*batch
...
+	}
+	jobName, found := pod.Labels[batchv1.JobNameLabel]
+	exitCode := int32(0)
+	if len(pod.Status.ContainerStatuses) > 0 && pod.Status.ContainerStatuses[0].State.Terminated != nil {
+		exitCode = pod.Status.ContainerStatuses[0].State.Terminated.ExitCode
+	}
+	if found && strings.HasPrefix(jobName, CPUManagerJobPrefix) && exitCode > 0 {
+		// Update exit code to cpu manager job
+		job, err := h.jobCache.Get(h.namespace, jobName)
+		if err != nil {
```
w13915984028: For `IsNotFound` error, may consider to log message and return nil.

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

pkg/controller/master/node/cpu_manager_controller.go
brandboat: IMO, there is no need to specifically handle the IsNotFound case in Line206, if any error raise here, we should throw error. Job is expected to be there if pod still exist.

I'll modify the code to
```
		if err != nil {
			logrus.WithField("job_name", jobName).WithError(err).Error("Failed to get cpu manager job from cache")
			return nil, err
		}
```

w13915984028: When a Job is deleted, the related Pod is also deleted, both meet the `.DeletionTimestamp != nil`;  but there is no assurance that the Pod will be gone first and the Job will be gone then. (I did not check the k8s source code yet);  hence, 
```
pod.DeletionTimestamp != nil && IsNotFound(job)
```
is better to be escaped, otherwise, if above assumption is true, here it may stuck reporting error.

brandboat: Roger that, will handle the IsNotFound and DeletionTimestamp case, thanks.

brandboat: gentle ping @w13915984028 , I've addressed the comment in the latest commit, also rebased to master branch. PTAL, thanks.

FrankYang0529 (approved):

brandboat:
@mergify backport to v1.4

Issue #6521, opened 2024-09-06
[BUG] vGPU endpoint gets cleared on upgrade


noahgildersleeve:
**Describe the bug**

When you upgrade from v1.3.1 to v1.3.2 it clears out the driver endpoint for the vGPU driver in `nvidia-driver-toolkit`
**To Reproduce**
1. Enable PCI passthrough on a machine with a vGPU enabled GPU
2. Enable SR-IOV on the GPU
3. Create a vGPU profile
4. Create a VM and assign the vGPU profile
5. Upgrade to v1.3.2
6. Try to turn on the VM

**Expected behavior**

The VM should run and the vGPU should be available
**Support bundle**

[supportbundle_64c257d7-7081-4e4a-9873-c36b6c9955f5_2024-09-06T16-31-43Z.zip](https://github.com/user-attachments/files/16910862/supportbundle_64c257d7-7081-4e4a-9873-c36b6c9955f5_2024-09-06T16-31-43Z.zip)

**Environment**
 - Harvester ISO version: v1.3.1 -> v1.3.2-rc2
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): DL360 and nvidia A2

**Additional context**

[hvst-upgrade-gd7dg-upgradelog-archive-2024-09-06T16-30-19Z.zip](https://github.com/user-attachments/files/16910874/hvst-upgrade-gd7dg-upgradelog-archive-2024-09-06T16-30-19Z.zip)

### Error state
You can see if this is happening after the upgrade by trying to turn on the VM. You will get an error that it is unschedulable
![Greenshot 2024-09-06 09 39 15](https://github.com/user-attachments/assets/5347b1eb-cece-4364-89ff-16e3b2f23664)

## Fix
1. Check the config for the vGPU in advanced -> Addons
![Greenshot 2024-09-06 09 38 53](https://github.com/user-attachments/assets/349dde3c-f6e4-4e2e-ad46-4c86d5005082)
2. Edit the config for nvidia-driver-toolkit
![Greenshot 2024-09-06 09 31 05](https://github.com/user-attachments/assets/10563fbb-e64b-42db-92f3-401982d1c2e9)
3. You will see that the driver location is reset to the default. Change this to the location of your grid driver for KVM
4. Wait for the update to run
![Greenshot 2024-09-06 09 38 36](https://github.com/user-attachments/assets/42d67da1-53e0-4b59-8948-8cea4d46860a)
[...]

PR #6526, opened 2024-09-09
improve nvidia-driver-toolkit upgrade

ibrokethecloud:

**Problem:**

If nvidia-driver-toolkit is enabled, then during upgrade the driver endpoint gets cleared since the default addon upgrade process replaces the valuesContent and version info from the new addon spec packaged in the upgrade container.

As a result of this patch, the nvidia-driver-toolkit gets upgraded with the driverLocation set to place holder location in the  addon. This results in new driver never being reinstalled post upgrade, and causes VM's with vGPU's to fail.

**Solution:**

PR introduces a minor change to extract the `driverLocation` from existing addon and using the same to patch the addon upgrade manifest, before the actual addon upgrade is performed.

**Related Issue:**
https://github.com/harvester/harvester/issues/6521
**Test plan:**

Test the change is simple as we do not need an actual node which supports vGPUs

* Install a Harvester v1.3.2 cluster
* Post install edit the `driverLocation` and set it to a non default endpoint, say `http://fakeendpoint/vgpu.kvm`
* Trigger upgrade of cluster with new iso built from current change
* Post upgrade the addon should be upgraded, while retaining the `driverLocation`

### DIFF
--- a/package/upgrade/lib.sh
+++ b/package/upgrade/lib.sh
@@ -595,3 +595,15 @@ EOF
   # wait status only when enabled and already AddonDeploySuccessful
   wait_for_addon_upgrade_deployment $name $namespace $enabled $curstatus
 }
+
+
+upgrade_nvidia_driver_toolkit_addon()
+{
+  # patch nvidia-driver-toolkit with existing location before performing upgrade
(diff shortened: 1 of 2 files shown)


bk201 (approved):

tserong (approved):

ibrokethecloud:
@mergifyio backport v1.3 v1.4

Issue #2850, opened 2022-09-29
[FEATURE] add Harvester alerts


w13915984028:
**Is your feature request related to a problem? Please describe.**

After feature `alertmanager` is added into Harvester https://github.com/harvester/harvester/issues/2517, it is the turn for `harvester`, `node-disk-manager`, `network-controller` etc. to add related alert rules, thus `Harvester` will have well defined rules to monitor the running metrics of `Harvester` itself.

**Describe the solution you'd like**

Step by step, add alert rules into source code implementation.

**Describe alternatives you've considered**

**Additional context**

Code PR:
https://github.com/harvester/harvester/pull/2678

HEP PR:
HEP: Harvester supports alertmanager https://github.com/harvester/harvester/pull/2545
https://github.com/harvester/harvester/pull/2545

https://github.com/harvester/harvester/issues/2520

https://github.com/harvester/harvester/issues/1781

w13915984028:
out-dated

PR #6538, opened 2024-09-10
2995 Validate if cluster network is ready and vlan config spans all nodes during storage network configuration

### LINKED ISSUE #2995: [BUG] Verify ClusterNetwork & VLAN Config for storage network setting in backend
**Describe the bug**

Storage Network didn't verify cluster network and VLAN config are correct and expected.
Users must ensure the Cluster Network is configured and VLAN Config will cover all nodes and ensure the network connectivity is working and expected in all nodes.

Currently, Harvester cannot import Network Controller's module, so webhook cannot verify cluster network and vlan config.

**To Reproduce**

1. Create ClusterNetwork
2. Don't create VLAN Config
3. assign ClusterNetwork to the storage network setting

**Expected behavior**

get an error or warning message.

**Support bundle**

**Environment**
 - Harvester ISO version:  1.1.0-rc3
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

rrajendran17:
**Problem:**
1.storage network can be created even when vlan config is not present
2.storage network can be created without vlanconfig spanning all nodes in the cluster

**Solution:**
1.check if cluster network is ready (vlan config attached) before configuring storage network
2.check if vlanconfig attached to the cluster network spans all nodes before configuring storage network

Related issue:
https://github.com/harvester/harvester/issues/2995

**Test plan:**
steps to reproduce 
1.The harvester cluster has 3 nodes
2.Create cluster network "cluster-1" and create vlan network config "vlan-cluster". Chose node selector to specific node.
3.create storage network config - create storage network with vlan id and cluster-1 ((Use api to create storage network as front end already takes care of this and cluster-1 will be disabled in UI)

Expected behaviour:
storage network should not be created with error from webhook "vlan config "clan-cluster" did not span all nodes


pkg/webhook/resources/setting/validator.go
```diff
@@ -1142,6 +1167,71 @@ func (v *settingValidator) checkStorageNetworkValueVaild() error {
...
+	} else {
+		return fmt.Errorf("vlan config is not present for cluster network %s", config.ClusterNetwork)
+	}
+
+	nodes, err := v.nodeCache.List(labels.Everything())
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+
+	//check if vlanconfig contains all the nodes in the cluster
```
w13915984028, line 1240: The `witness-node` needs to be skipped https://docs.harvesterhci.io/v1.4/advanced/witness. It does not deploy LH related pods or VMs. 

Another point:

After storage-network is enabled, a new none-witness node is added, and this node is not covered by VC, how can we hint this?

w13915984028 (commented):
FYI.

pkg/webhook/resources/setting/validator.go
rrajendran17: Sure, will add the witness-node check to the changes.

I think pkg/webhook/resource/node/validator.go can handle checks during Update where a new node addition is handled.
The storage network nad cache could be validated to see if it exists but the challenge is webhook returns always an "error" or nil.Is there something where a message could be notified to the user ? Need to check with someone from UI if there is any other way any info from the cluster could be notified to the user/UI.

rrajendran17:
During testing I observed that, cluster network will move to "Ready state" even with 1 vlan config present under cluster network.So checking for cluster network in ready state will not be correct, so I modified my previous code and added changes to validate if all vlan status present under the cluster network is in "Ready" state.

pkg/webhook/resources/setting/validator.go
rrajendran17: Added the code changes to skip "witness-node" during the validation

w13915984028: One solution is to add metrics and alerts like https://kubevirt.io/monitoring/metrics.html#kubevirt_virt_controller_ready, https://kubevirt.io/monitoring/runbooks/NoReadyVirtOperator.html to indicate that certain nodes in the storagenetwork are not ready.

We may design and implement this feature in v1.5.0. https://github.com/harvester/harvester/issues/2850

w13915984028:
BTW, the `network-controller` has no test code upon webhook, I am adding them now.

pkg/webhook/resources/setting/validator.go
```diff
@@ -1142,6 +1169,85 @@ func (v *settingValidator) checkStorageNetworkValueVaild() error {
...
+
+	if len(vsList) > 0 {
+		for _, vs := range vsList {
+			if networkv1.Ready.IsFalse(vs.Status) {
+				return fmt.Errorf("vs %s status is not Ready", vs.Name)
+			}
+		}
+	} else {
+		return fmt.Errorf("vlan status is not present for cluster network %s", config.ClusterNetwork)
+	}
```
starbops: The functionality is sound, but please consider implementing it in the following way to avoid nested logic.

```suggestion
	if len(vsList) == 0 {
	    return fmt.Errorf("vlan status is not present for cluster network %s", config.ClusterNetwork)
	}
	
	for _, vs := range vsList {
	    if networkv1.Ready.IsFalse(vs.Status) {
		    return fmt.Errorf("vs %s status is not Ready", vs.Name)
	    }
	}
	
	return nil
```

pkg/webhook/resources/setting/validator.go
```diff
@@ -1142,6 +1169,85 @@ func (v *settingValidator) checkStorageNetworkValueVaild() error {
...
+			vnodes, err := getMatchNodes(vc)
+			if err != nil {
+				return err
+			}
+			vlanConfigNodes = append(vlanConfigNodes, vnodes...)
+		}
+		matchedNodes = mapset.NewSet[string](vlanConfigNodes...)
+	} else {
+		return fmt.Errorf("vlan config is not present for cluster network %s", config.ClusterNetwork)
+	}
```
starbops: Ditto.

starbops: Maybe we can just `matchedNodes.Append(vnodes...)` and get rid of `vlanConfigNodes` entirely.

starbops (commented):
Overall LGTM, just two nits. Please take a look.

pkg/webhook/resources/setting/validator.go
rrajendran17: Thanks @starbops taken care.Updated the PR.

Note:the build failure is to due to a latest breakage in harvester master due to the following commit
"https://github.com/harvester/harvester/commit/542f684a7d13b5558bc9125b636e5aab638b8c96"
failure is due to "fmt" package is not imported in that file

starbops: #6593 has been merged. Let's rebase the PR again. Thank you.

rrajendran17: @starbops sure done.The pipeline is green now

rrajendran17:
@starbops @w13915984028 @WebberHuang1118 @chrisho if it is all good, we can merge this.

pkg/webhook/clients/clients.go
```diff
@@ -12,6 +12,7 @@ import (
 	v1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/client-go/rest"
 
+	ctlnetwork "github.com/harvester/harvester-network-controller/pkg/generated/controllers/network.harvesterhci.io"
```
w13915984028: In your commit, there has been generated files like [pkg/generated/controllers/network.harvesterhci.io/factory.go](https://github.com/harvester/harvester/pull/6538/files#diff-6b91fccaf8eac94906aa45d10b15e5f726962e25dac7ddeee9a8cc975e3f49b9)
did you try to import them instead of from `harvester/harvester-network/controller`?

pkg/webhook/resources/setting/validator.go
```diff
@@ -1144,6 +1169,83 @@ func (v *settingValidator) checkStorageNetworkValueVaild() error {
...
 
+func (v *settingValidator) checkVlanStatusReady(config *storagenetworkctl.Config) error {
+	vsList, err := v.vsCache.List(labels.Set{
+		utils.KeyClusterNetworkLabel: config.ClusterNetwork,
+	}.AsSelector())
+	if err != nil {
+		return err
+	}
+
+	if len(vsList) == 0 {
```
w13915984028, line 1187: seems better to have a `checkClusterNetwork` function to sure the `ClusterNetwork` is existing at first, it will emit an exact error like `the ClusterNetwork xyz is not existing`

then the `checkVlanStatusReady`

w13915984028 (commented):
Please check those two questions, thanks.

pkg/webhook/clients/clients.go
rrajendran17: @w13915984028 yes it has to be imported from "github.com/harvester/harvester/pkg/generated/controllers/network.harvesterhci.io/" instead of "github.com/harvester/harvester-network-controller/pkg/generated/controllers/network.harvesterhci.io/".Changed in other places too.Thanks

pkg/webhook/resources/setting/validator.go
rrajendran17: @w13915984028 Added check for existence of cluster network as well.Thanks

w13915984028 (approved):
LGTM, thanks.

starbops (approved):
LGTM, thank you!

mingshuoqiu (approved):
LGTM.

WebberHuang1118 (approved):
LGTM, thanks for the PR.

rrajendran17:
@Mergifyio backport v1.4

Issue #6484, opened 2024-09-03
[BUG] go-common sys.(Try)RestartService doesn't close dbus connection, repeated calls eventually fail


tserong:
**Describe the bug**
When repeatedly changing ntp server settings, or (in harvester v1.4) repeatedly enabling and disabling the LH V2 data engine, eventually node-manager will fail to restart the underlying systemd-timesyncd and/or rke2 services due to the maximum number of dbus connections being exceeded.

The trivial workaround if you ever hit this is to restart node-manager.

**To Reproduce**
1. Open three terminals and ssh to the same harvester node in each
2. In the first terminal, run `journalctl -f`
3. In the second terminal, run `kubectl -n harvester-system logs -f -l app.kubernetes.io/name=harvester-node-manager`
4. In the third terminal, run this loop, but change the node name as appropriate for your environment (I'm using `harvester-node-1` in this example):

```
for i in $(seq 1 128) ; do
  for j in 0 1 ; do
    kubectl -n harvester-system patch --type merge -p "{\"spec\": {\"ntpConfigs\": {\"ntpServers\": \"$j.suse.pool.ntp.org\"}}}" nodeconfigs/harvester-node-1
    sleep 5
  done
done
```

5. After a few minutes (in my test, about 70 times through the loop) you'll see errors in the other two terminals, something like the following:

```
# journalctl -f
[...]
Sep 03 06:50:29 harvester-node-1 dbus-daemon[1502]: [system] The maximum number of active connections for UID 0 has been reached (max_connections_per_user=256)
```
```
# kubectl -n harvester-system logs -f -l app.kubernetes.io/name=harvester-node-manager
time="2024-09-03T06:48:27Z" level=error msg="Failed to create new connection for systemd. err: dial unix /run/systemd/private: connect: no such file or directory"
time="2024-09-03T06:48:27Z" level=error msg="Restart systemd-timesyncd fail. err: dial unix /run/systemd/private: connect: no such file or directory"
time="2024-09-03T06:48:27Z" level=error msg="error syncing 'harvester-system/harvester-node-1': handler harvester-node-config-controller: dial unix /run/systemd/private: connect: no such file or directory, requeuing"

```

**Expected behavior**
The services can be restarted indefinitely and we never hit the dbus connection limit.

**Environment**
 - Harvester ISO version: v1.4.0-dev-20240830
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): virtualized

tserong:
[...]

PR #6566, opened 2024-09-16
bump(chart): node-manager to v0.3.2

tserong:
**Problem:**
When repeatedly changing ntp server settings, or (in harvester v1.4) repeatedly enabling and disabling the LH V2 data engine, eventually node-manager will fail to restart the underlying systemd-timesyncd and/or rke2 services due to the maximum number of dbus connections being exceeded.

**Solution:**
Bump the node-manager chart to v0.3.2 which includes the fix for this issue

**Related Issue:**
https://github.com/harvester/harvester/issues/6484

**Test plan:**
See related issue


tserong:
Closing in favour of https://github.com/harvester/harvester/pull/6567

PR #6567, opened 2024-09-16
bump(chart): node-manager to v0.3.2, node-disk-manager to v0.7.3

tserong:
**Problem:**
Need to pick up the latest node-manager and node-disk manager for v1.4

**Solution:**
This PR

**Related Issue:**
https://github.com/harvester/harvester/issues/6484
https://github.com/harvester/harvester/issues/5724
https://github.com/harvester/harvester/issues/5274

**Test plan:**
See related issues

PR #6578, opened 2024-09-18
bump(chart): sync node-manager & node-disk-manager versions in values.yaml

tserong:
This was missed in https://github.com/harvester/harvester/pull/6567


Vicente-Cheng (approved):
lgtm!

bk201 (approved):

Issue #6627, opened 2024-09-24
[BUG] The new secret webhook may cause CI to fail


w13915984028:
**Describe the bug**

The CI may fail from time to time.

https://github.com/harvester/harvester/actions/runs/11008796171/job/30578651252

```
make ci log..
...
requested resource (get virtualmachines.meta.k8s.io)
Add/Modify event for harvester-pre-delete: MODIFIED
Starting delete for "harvester-pre-delete" Job
  uninstall: given cascade value: , defaulting to delete propagation background
Starting delete for "harvester-webhook" Service
Starting delete for "harvester-load-balancer-webhook" Service
Starting delete for "harvester-network-webhook" Service
Starting delete for "harvester-node-manager-webhook" Service
Starting delete for "harvester-snapshot-validation-webhook" Service
Starting delete for "harvester" Service
Starting delete for "harvester-network-webhook" Deployment
Starting delete for "harvester-snapshot-validation-webhook" Deployment
Starting delete for "harvester-node-manager-webhook" Deployment
Starting delete for "virt-operator" Deployment
Starting delete for "harvester" Deployment
Starting delete for "harvester-network-controller-manager" Deployment
Starting delete for "harvester-webhook" Deployment
...
Starting delete for "harvester-helpers" ConfigMap
Starting delete for "whereabouts-config" ConfigMap
Starting delete for "vip" ConfigMap

Starting delete for "snapshot-validation-webhook-tls" Secret

Starting delete for "harvester" ServiceAccount
Starting delete for "snapshot-controller" ServiceAccount
Starting delete for "harvester-load-balancer" ServiceAccount
Starting delete for "harvester-load-balancer-webhook" ServiceAccount
Starting delete for "harvester-network-controller" ServiceAccount
Starting delete for "harvester-network-helper" ServiceAccount
Starting delete for "harvester-network-webhook" ServiceAccount
Starting delete for "harvester-node-manager" ServiceAccount
Starting delete for "harvester-node-manager-webhook" ServiceAccount
Starting delete for "kube-vip" ServiceAccount
Starting delete for "kubevirt-operator" ServiceAccount
Starting delete for "snapshot-validation-webhook" ServiceAccount
Starting delete for "harvester-whereabouts" ServiceAccount
Starting delete for "harvester" NetworkPolicy
Starting delete for "backing-image-data-source" NetworkPolicy
Starting delete for "backing-image-manager" NetworkPolicy
Starting delete for "instance-manager" NetworkPolicy
[...]

PR #6629, opened 2024-09-24
Remove harvester validator in pre-delete

w13915984028:

**Problem:**

CI fails from time to time.

**Solution:**

Remove harvester webhook in pre-delete when helm uninstalling harvester.

**Related Issue:**
https://github.com/harvester/harvester/issues/6627

**Test plan:**

### DIFF
--- a/deploy/charts/harvester/templates/job.yaml
+++ b/deploy/charts/harvester/templates/job.yaml
@@ -66,6 +66,8 @@ spec:
               echo '[INFO] deleting kubevirt ...';
               kubectl delete kubevirt --wait --all;
               {{ end }}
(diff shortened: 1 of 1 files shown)


w13915984028:
@mergify backport v1.4

bk201 (approved):
@w13915984028 Do you know the root cause of why this is not always reproduced? Jack has some comments on https://github.com/harvester/harvester/issues/6627#issuecomment-2372858648

Deleting the harvester chart is not supposed to happen in non-ci clusters, we can merge the change to see if it fixes the CI failure.

Yu-Jack (approved):

Issue #6631, opened 2024-09-24
[BUG]  master-0bde441-head cannot build storage classes


irishgordo:
**Describe the bug**
Cannot build a storage class, error is:
> admission webhook "validator.harvesterhci.io" denied the request: storage class base is not for encryption or decryption

**To Reproduce**
1. Modify some hosts, add extra storage
1. Build a storage class either via Form or Edit As YAML, that targets the nodes new force-formatted & tagged storage
1. It will fail 

**Expected behavior**
The ability to add the storage class

**Workaround**
The workaround is to remove the "encrypted" property from the YAML in the `parameters` section:
```
__clone: true
metadata:
  annotations:
    field.cattle.io/description: base storage class to target secondary volumes
  name: base
  labels: {}
parameters:
  numberOfReplicas: '3'
  staleReplicaTimeout: '30'
  diskSelector: extra
  nodeSelector: hosts
  migratable: 'true'
provisioner: driver.longhorn.io
allowVolumeExpansion: true
reclaimPolicy: Delete
volumeBindingMode: Immediate
apiVersion: storage.k8s.io/v1
kind: StorageClass
```

**Other Attempted Workarounds, that didn't workout, so can ignore**
1. trying to change to `encrypted: false` -> will yield:  `StorageClass in version "v1" cannot be handled as a StorageClass: json: cannot unmarshal bool into Go struct field StorageClass.parameters of type string`
2. trying to just set it to `encrypted: 'true'` -> will yield: `admission webhook "validator.harvesterhci.io" denied the request: storage class must contain csi.storage.k8s.io/provisioner-secret-name, csi.storage.k8s.io/provisioner-secret-namespace, csi.storage.k8s.io/node-stage-secret-name, csi.storage.k8s.io/node-stage-secret-namespace, csi.storage.k8s.io/node-publish-secret-name, csi.storage.k8s.io/node-publish-secret-namespace`

**Environment**
 - Harvester ISO version: master-0bde441-head
 - Underlying Infrastructure: 4 node bare-metal hp dl160 gen9s

**Additional context**
![Screenshot from 2024-09-24 16-33-54](https://github.com/user-attachments/assets/d58f593f-cfef-4207-895e-b83b9517a30e)
![Screenshot from 2024-09-24 16-32-10](https://github.com/user-attachments/assets/0ef1f100-af01-478b-aa50-a99ac5e54eb4)
[...]

PR #6632, opened 2024-09-25
fix: storage class validation

Yu-Jack:

**Problem:**
Currently, storage class validation returns an error when `encrypted` field exists and its value is not `true`.

**Solution:**
We should only return an error when the value is `true` but required parameters are not provided.

**Related Issue:**
https://github.com/harvester/harvester/issues/6631

**Test plan:**
Please check [this unit test file](https://github.com/harvester/harvester/blob/968da91de16ba84cab5e9bd6c026ae7244ab8db1/pkg/webhook/resources/storageclass/validator_test.go) to check different test cases.

### DIFF
--- a/pkg/webhook/resources/storageclass/validator.go
+++ b/pkg/webhook/resources/storageclass/validator.go
@@ -2,6 +2,7 @@ package storageclass
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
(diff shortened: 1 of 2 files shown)


WebberHuang1118 (approved):
LGTM, thanks.

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -138,12 +138,33 @@ func (v *storageClassValidator) validateEncryption(newObj runtime.Object) error
 	newSC := newObj.(*storagev1.StorageClass)
 
 	// Use util.LonghornOptionEncrypted as the key to check if the storage class is encrypted
-	if value, ok := newSC.Parameters[util.LonghornOptionEncrypted]; !ok {
+	if value, ok := newSC.Parameters[util.LonghornOptionEncrypted]; !ok || value != "true" {
```
FrankYang0529: This field can only be boolean in LH volume. We should validate whether the value can be parsed as boolean in webhook.

https://github.com/longhorn/longhorn-manager/blob/3ac6b93c30e9f0839129e6fa3c205e171726340c/k8s/pkg/apis/longhorn/v1beta2/volume.go#L279-L280

FrankYang0529 (changes requested):
Overall LGTM. One change request. Thanks for the PR.

pkg/webhook/resources/storageclass/validator.go
Yu-Jack: I switch to using `strconv.ParseBool`.

FrankYang0529 (approved):
Thanks for the fix.

PR #6666, opened 2024-09-27
feat: update rancher and docker package

FrankYang0529:
**Test plan:**
Create a cluster without error.

**Behavior Change:**
1. Remove name missing cases in integration tests. When using harvester API calls, missing name will automatically get `generateName: h-`.
2. Remove uninstall harvester chart in integration tests. The harvester chart can't be uninstalled correctly in the kind cluster. The root cause may be that we don't really run harvester deployment, but there is finalizers for jobs, so helm can't remove harvester-pre-delete job. Remove the step since we don't support users to helm install harvester directly.


Yu-Jack:
@w13915984028  If this PR decides not to remove Harvester, I think we don't need to handle this issue https://github.com/harvester/harvester/issues/6627 ? Although it's closed, CI still failed. cc @bk201

bk201:
@w13915984028 this should also resolve https://github.com/harvester/harvester/issues/6627#issuecomment-2373764207, another case (https://github.com/harvester/harvester/actions/runs/11210615498/job/31157840560).
We don't expect a chart uninstall in real clusters.

w13915984028 (approved):
LGTM, thanks.

w13915984028:
@bk201 @Yu-Jack agree with your comments, the un-installation is not a published feature. I planed to investigate the root cause of #6627, but now it is done :), thanks @FrankYang0529

Yu-Jack (approved):
LGTM, thanks!

bk201 (approved):

FrankYang0529:
When both `name` and `generateName` are missing, steve server will use `schema.ID[0:1]+"-"` as `generateName`. That is why missing name cases fail.

commit: https://github.com/rancher/steve/commit/e930ae3691b2d645d3112b013d0bb558008811c3
code: https://github.com/rancher/steve/blob/1f21e5e5151682544157b24c49b9e4aa5e99bb39/pkg/stores/proxy/proxy_store.go#L431-L437

PR #6677, opened 2024-09-30
fix(upgradelog): support log archive pvc alt name

starbops:

**Problem:**

After upgrading from v1.3.2 to v1.4-head (or master-head), users is unable to download the upgrade log. This is because the previous fix #6349 is not backward-compatible.

**Solution:**

The log archive PVC alt name will be saved in the upgradelog object's annotation. If such an annotation exists, the downloader and packager pods will refer to that PVC instead of the default one. That is to say, the name of the PVC spawned by the Logging CR will be recorded in the UpgradeLog CR's annotation.

**Related Issue:**

#6272 

**Test plan:**

Test the following two upgrade scenario (both with the upgradelog feature enabled):

- Upgrade a Harvester cluster from v1.3.2 to master (to make sure we don't break anything)
- Upgrade a Harvester cluster from master to master (to make sure the fix is effective)

For both scenario, please try to download the upgradelog at the following time point:

- During the upgrade
- After the upgrade success (or failed, either should work)


bk201 (approved):
LGTM

pkg/controller/master/upgradelog/controller.go
```diff
@@ -616,6 +616,17 @@ func (h *handler) OnPvcChange(_ string, pvc *corev1.PersistentVolumeClaim) (*cor
...
 	}
 
+	// Add the PVC's name to the UpgradeLog annotation for tracking.
+	// This is useful for the log downloader to know which PVC to download logs from.
+	upgradeLogToUpdate := upgradeLog.DeepCopy()
+	if upgradeLogToUpdate.Annotations == nil {
+		upgradeLogToUpdate.Annotations = make(map[string]string, 1)
+	}
+	upgradeLogToUpdate.Annotations[util.AnnotationUpgradeLogLogArchiveAltName] = pvc.Name
+	if _, err := h.upgradeLogClient.Update(upgradeLogToUpdate); err != nil {
```
brandboat: nit: perhaps we don't need to upgrade if upgradeLogToUpdate is the same as the current upgradeLog.

pkg/controller/master/upgradelog/common.go
```diff
@@ -402,21 +401,14 @@ func prepareLogDownloaderSvc(upgradeLog *harvesterv1.UpgradeLog) *corev1.Service
 	}
 }
 
-// Returns the name of the log-archive PVC, which is created by the fluentd StatefulSet.
-//
-//	The name will look like: <upgradelog>-infra-log-archive-<upgradelog>-infra-fluentd-0
-//	For instance: hvst-upgrade-bczl4-upgradelog-infra-log-archive-hvst-upgrade-bczl4-upgradelog-infra-fluentd-0.
-//
-// TODO: As of rancher-logging v4.4.0, we use the PVC created by the fluentd StatefulSet, not making it ourselves. After v4.6.0, we need to revisit here and perhaps update the implementation because the upstream behavior changes.
+// Returns the name of the log-archive PVC, which is created by the fluentd StatefulSet. If an alternative name is set in the annotation, return it instead.
```
brandboat: nit: Would be great if we can add a comment here to clarify that we do this for backward compatibility.

brandboat (approved):
LGTM, all comments are optional, please feel free to ignore them.

test v1.3.2 -> master
✅  download log during upgrade
✅  download log after upgrade success

test master -> master
✅  download log during upgrade
✅  download log after upgrade success

FrankYang0529 (approved):
LGTM. Thanks.

FrankYang0529:
@mergify backport v1.4

Issue #6678, opened 2024-09-30
[ENHANCEMENT] Do not allow to delete harvester-longhorn storageclass


w13915984028:
**Is your enhancement related to a problem? Please describe.**

User may delete harvester-longhorn SC, which is monitoring by managedchart, and also used by rancher-monitoring stack by default.

**Describe the solution you'd like**

Deny to delete it.

**Describe alternatives you've considered**

**Additional context**

w13915984028:
Test plan:

(1) Install a v1.4.0 new cluster.
(2) Delete the `harvester-longhorn` storage-class, it will be denied.
(3) Upgrade a v1.3.* cluster to v1.4.0, repeat (2).

(4) On the newly installed v1.4.0 or upgrade to v1.4.0 cluster, create a new storage-class, add `harvesterhci.io/is-reserved-storageclass: "true"` to it's annotations manually, then delete this SC, it will be denied;  remove this annotation/set the value to "false", the SC can be deleted.

brandboat:
Test OK on [v1.4.0-rc2](https://github.com/harvester/harvester/releases/tag/v1.4.0-rc2), close as fixed.

## Environment
* [v1.4.0-rc2](https://github.com/harvester/harvester/releases/tag/v1.4.0-rc2)
  * Profile: 1 node QEMU/KVM (8C/16G/250G)
  * ui-source: `Auto`

## Test Steps
✅  Delete the harvester-longhorn storage-class, got error `The request is invalid: : storage class harvester-longhorn is reserved by Harvester and can't be deleted`
✅  Create storage-class `test` and add annotation `harvesterhci.io/is-reserved-storageclass: "true"` manually, then delete the SC, got the same error message as above.
✅  Create storage-class `test`, then I can delete the SC successfully.

These 3 checks all passed under 
- newly installed v1.4.0-rc2 harvester 
- cluster upgrade from v1.3.2 to v1.4.0-rc2

Issue #6564, opened 2024-09-16
[FEATURE] Prevent users from deleting default resources


Yu-Jack:
**Is your enhancement related to a problem? Please describe.**

For example, upgrading checks the bundle status. It's not allowed to upgrade if the status is not ready. The root cause is that users delete the default storage class. Users need to create it back with specific annotation and label because fleet would checks them. Related issue is https://github.com/harvester/harvester/issues/6561.

Besides, I believe there are similar cases where users might accidentally delete something, leading to issues such as failed upgrades and failed enabling of `rancher-monitoring` add-on.

**Describe the solution you'd like**

We could use admission webhook to prevent users from deleting default storage class. Beside that, we should also check whether there is any similar case in other resources.

- Storage Class : https://github.com/harvester/harvester/issues/6678

**Describe alternatives you've considered**

**Additional context**

PR #6679, opened 2024-09-30
Deny to delete the harvester-longhorn sc

w13915984028:

**Problem:**

The `harvester-longhorn` SC can be deleted.

**Solution:**

1. Allow to use annotation `harvesterhci.io/is-reserved-storageclass: "true"` to protect any SC
2. Add annotation `harvesterhci.io/is-reserved-storageclass: "true"` to harvester chart to protect the default harvester-longhorn SC.
3. For legacy `harvester-longhorn` SC, protect it as well when there is no `harvesterhci.io/is-reserved-storageclass: "true"` annotation.

**Related Issue:**
https://github.com/harvester/harvester/issues/6678

**Test plan:**

1. Create a new SC and set it as default
4. Delete the `harvester-longhorn` SC, which should be blocked by webhook.

image

5. Create a new SC, and add annotation `harvesterhci.io/is-reserved-storageclass: "true"` manually, the SC can't be deleted.
```
$kubectl get storageclass test -oyaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    harvesterhci.io/is-reserved-storageclass: "true"  // add this manually
  creationTimestamp: "2024-10-01T09:13:09Z"
  name: test
  resourceVersion: "111649"
  uid: 4cd9771f-a6d2-4734-8cad-11663c171994
parameters:
  dataEngine: v1
  encrypted: "false"
  migratable: "true"
  numberOfReplicas: "1"
  staleReplicaTimeout: "30"
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

image

### DIFF
--- a/pkg/webhook/resources/storageclass/validator.go
+++ b/pkg/webhook/resources/storageclass/validator.go
@@ -21,6 +21,10 @@ import (
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
+const (
+	errorMessageReservedStorageClass = "storage class %s is reserved by Harvester and can't be deleted"
+)
+
 var pairs = [][2]string{
 	{util.CSIProvisionerSecretNameKey, util.CSIProvisionerSecretNamespaceKey},
 	{util.CSINodeStageSecretNameKey, util.CSINodeStageSecretNamespaceKey},
@@ -79,11 +83,19 @@ func (v *storageClassValidator) Update(_ *types.Request, _ runtime.Object, newOb
 
 func (v *storageClassValidator) Delete(_ *types.Request, obj runtime.Object) error {
 	sc := obj.(*storagev1.StorageClass)
-	err := v.validateHarvesterLonghornSC(sc)
-	if err != nil {
-		return err
+
+	validators := []func(*storagev1.StorageClass) error{
+		v.validateReservedStorageClass,
+		v.validateVMImageUsage,
 	}
-	return v.validateVMImageUsage(sc)
+
+	for _, validator := range validators {
+		if err := validator(sc); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // Harvester rejects setting dataLocality as strict-local, because it makes volume non migrateable,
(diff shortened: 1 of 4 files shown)


w13915984028:
@mergify backport v1.4, v1.3

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -79,6 +79,10 @@ func (v *storageClassValidator) Update(_ *types.Request, _ runtime.Object, newOb
 
 func (v *storageClassValidator) Delete(_ *types.Request, obj runtime.Object) error {
 	sc := obj.(*storagev1.StorageClass)
+	err := v.validateHarvesterLonghornSC(sc)
+	if err != nil {
+		return err
+	}
 	return v.validateVMImageUsage(sc)
 }
```
Yu-Jack: NIT: we could put those validators into slice.

```go
sc := obj.(*storagev1.StorageClass)

validators := []func(*storagev1.StorageClass) error {
	v.validateHarvesterLonghornSC,
	v.validateVMImageUsage,
}

for _, validator := range validators {
	if err := validator(sc); err != nil {
		return err
	}
}

return nil
```

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -229,3 +233,15 @@ func (v *storageClassValidator) validateVMImageUsage(sc *storagev1.StorageClass)
...
+// It should not be deleted accidentally.
+func (v *storageClassValidator) validateHarvesterLonghornSC(sc *storagev1.StorageClass) error {
+	if sc.Name == util.StorageClassHarvesterLonghorn && sc.Annotations != nil {
+		if sc.Annotations[util.HelmReleaseNamespaceAnnotation] == util.HarvesterSystemNamespaceName && sc.Annotations[util.HelmReleaseNameAnnotation] == util.HarvesterChartReleaseName {
+			return werror.NewInvalidError(fmt.Sprintf("storage class %s is reserved by Harvester and can't be deleted", sc.Name), "")
+		}
+	}
+
+	return nil
+}
```
Yu-Jack: The whole annotations and labels are:

```yaml
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  labels:
```

Currently, users can't upgrade the Harvester due to losing `objectset.rio.cattle.io/id` and `objectset.rio.cattle.io/hash`. Besides checking helm installation, I suggest that we also need to check these two fields.

w13915984028: The current code is assuming that certain labels/annotations are set by the helm & managedchart; it is not the best idea to list all the known labels/annotations here from following considerations:

(0) The harvester-longhorn is defined in Harvester chart https://github.com/harvester/harvester/blob/606f485579d087def560e2a8a255a2b0f4bad019/deploy/charts/harvester/templates/harvester-storageclass.yaml#L5

(1) Harvester supports developer mode, install from helm directly, https://docs.harvesterhci.io/v1.3/developer/developer-mode-installation

(2) The managedchart is on-top of helm.

The helm related annotations will be there; managedchart related may be there. For webhook, check the helm related annotations works for most scenarios.

One drawback is that, no one is suring those annotations won't change.

(3) A better solution is to add a harvester releated label/annotation in the chart definition, thus the webhook can drop the dependency on helm/managedchart. I will add this.

It may not work for legacy deployments.

Yu-Jack (approved):
LGTM. I think it's covered:

1. New Harvester after this PR.
2. Old Harvester before this PR.

It seems that https://github.com/harvester/harvester/issues/6564 could use same mechanism.

bk201 (approved):
lgtm, @w13915984028 did you verify the change in upgrade path?

w13915984028:
@bk201 

I did a test, in the `v132->v140` upgrade path, the `managedchart/helm` will patch the newly added annotation automatically, and which will help the webhook to deny the deletion.

```
Before upgrade:

harv21:/home/rancher # kubectl get storageclass harvester-longhorn -oyaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: "2024-09-23T12:23:32Z"
  labels:
  name: harvester-longhorn
  resourceVersion: "3334"
  uid: 142208dc-1efa-4985-b52a-21c4e58492b5
...

After upgrade:

harv21:/home/rancher # kubectl get storageclass harvester-longhorn -oyaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    harvesterhci.io/is-reserved-storageclass: "true"  // after upgrade, auto added
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: "2024-09-23T12:23:32Z"
  labels:
  name: harvester-longhorn
  resourceVersion: "33566"
  uid: 142208dc-1efa-4985-b52a-21c4e58492b5
...
```

image

Issue #6670, opened 2024-09-27
[BUG] Those after-upgrade actions may conflict with a new upgrade


w13915984028:
**Describe the bug**

Upgrade controller will try to:
(1) Remove unused images (https://github.com/harvester/harvester/issues/5749)
(2) Start VMs which were running before the upgrade if it is in a single-node cluster (https://github.com/harvester/harvester/issues/4005)

It is based on the condition that the current upgrade is completed, either successful or failed.

https://github.com/harvester/harvester/blob/606f485579d087def560e2a8a255a2b0f4bad019/pkg/controller/master/upgrade/upgrade_controller.go#L217

```
h.cleanup(upgrade, harvesterv1.UpgradeCompleted.IsTrue(upgrade))
h.restoreVMState(upgrade)
repo.startVM()
cleanupImages(upgrade, repo)
```

The webhookk will deny new upgrade:

https://github.com/harvester/harvester/blob/606f485579d087def560e2a8a255a2b0f4bad019/pkg/webhook/resources/upgrade/validator.go#L122

if an existing upgrade is not completed.

https://github.com/harvester/harvester/blob/606f485579d087def560e2a8a255a2b0f4bad019/pkg/webhook/resources/upgrade/validator.go#L132

You will notice that, when the controller is running to do those last actions, another upgrade can be started now 
(https://github.com/harvester/harvester/issues/6144,  PR: https://github.com/harvester/dashboard/pull/1081), 

the webhook does not block it.

**To Reproduce**

**Expected behavior**

The new upgrade should be blocked until the privious upgrade has finished all of those last actions.

It may affect those continuous upgrades.

**Support bundle**

**Environment**
 - Harvester ISO version:  v1.4.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

This is checked via validating issues:
https://github.com/harvester/harvester/issues/4005

https://github.com/harvester/harvester/issues/6144
PR: https://github.com/harvester/dashboard/pull/1081

https://github.com/harvester/harvester/issues/5749

w13915984028:
cc @bk201 @starbops @tserong

bk201:
@w13915984028 I'd suggest fixing this in the next milestone, given it's not a very common case to start the upgrade again. Does this sound good to you?

@brandboat the secret name to keep VMs is re-used and can cause conflict in this situation.

brandboat:
[...]

PR #6683, opened 2024-10-01
Enhance single node upgrade restore vm operation

brandboat:
**Problem:**
address comments in https://github.com/harvester/harvester/issues/4005#issuecomment-2376379554
- Need to wait until kubevirt is ready to handle vm operation then start restoring vm.
- The information of vm before upgrade currently are placed in the same secret even if multiple upgrade CR exist. There are two drawbacks here, 1) Hard to know the vm info before upgrade in each upgrade 2) if error occurs, restoreVM delete secret, and in the next round of upgrade#onChange, restoreVM do nothing as the secret disappeared.

**Solution:**
- To make sure kubevirt is ready before `restoreVM`, this pr triggers the restoreVM method after repo VM startup successfully.
- Store the pre-upgrade vm info in upgrade CR annotation instead of secret.
- Also add a validator to make sure `restoreVM` setting should not be true under multi-node cluster.

**Related Issue:**
https://github.com/harvester/harvester/issues/4005

**Test plan:**
1. Create iso based on this pr/branch
3. Use ipxe to create single node cluster, 
4. Create 4 VMs with 2 running VMs, 1 stopped VM, and 1 paused VM. 
  image
5. Apply the following yaml to update `restoreVM: true` in Setting upgrade-config
  ```yaml
  value: '{"imagePreloadOption":{"strategy":{"type":"sequential"}}, "restoreVM": true}'
  kind: Setting
  metadata:
    name: upgrade-config
  ```
6. Use the following yaml to upgrade harvester, note that the iso in isoURL should be the same one build in step1.
  ```yaml
  kind: Version
  metadata:
    name: master
    namespace: harvester-system
  spec:
  ---
  kind: Upgrade
  metadata:
    annotations:
    generateName: hvst-upgrade-
    namespace: harvester-system
  spec:
    version: "master"
  ```
7. Verify all VMs were running before upgrade are still in running state after upgrade. And the paused vm before upgrade are stopped after upgrade.
image

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -73,7 +73,7 @@ const (
 	skipVersionCheckAnnotation          = "harvesterhci.io/skip-version-check"
 	defaultImagePreloadConcurrency      = 1
 
-	preUpgradeVMsSecretName = "pre-upgrade-vms"
+	preUpgradeVMsAnnotation = "harvesterhci.io/pre-upgrade-vms"
 
 	vmReady condition.Cond = "Ready"
 )
@@ -88,8 +88,6 @@ type upgradeHandler struct {
 	ctx               context.Context
 	namespace         string
 	nodeCache         ctlcorev1.NodeCache
-	secretClient      ctlcorev1.SecretClient
-	secretCache       ctlcorev1.SecretCache
 	jobClient         v1.JobClient
 	jobCache          v1.JobCache
 	upgradeClient     ctlharvesterv1.UpgradeClient
@@ -114,6 +112,8 @@ type upgradeHandler struct {
 	lhSettingClient ctllhv1.SettingClient
 	lhSettingCache  ctllhv1.SettingCache
 
+	kubeVirtCache kubevirtctrl.KubeVirtCache
+
 	vmRestClient rest.Interface
 }
 
@@ -130,16 +130,16 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 	if harvesterv1.UpgradeCompleted.GetStatus(upgrade) == "" {
 		logrus.Infof("Initialize upgrade %s/%s", upgrade.Namespace, upgrade.Name)
 
-		if err := h.storeVMState(upgrade); err != nil {
-			return nil, err
-		}
-
 		if err := h.resetLatestUpgradeLabel(upgrade.Name); err != nil {
 			return nil, err
 		}
 
 		toUpdate := upgrade.DeepCopy()
 		initStatus(toUpdate)
+
+		if err := h.storeVMState(toUpdate); err != nil {
+			return nil, err
+		}
 
 		if !upgrade.Spec.LogEnabled {
 			logrus.Info("Upgrade observability is administratively disabled")
@@ -221,16 +221,15 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			return nil, h.cleanup(upgrade, harvesterv1.UpgradeCompleted.IsTrue(upgrade))
 		}
 
-		// try to restore vm to the state before upgrade
-		// moved the restoreVMState call after the image-cleanup annotation check,
-		// ensuring the function is called only once
-		h.restoreVMState(upgrade)
-
 		// repo VM is required for the image cleaning procedure, bring it up if it's down
 		logrus.Info("Try to start repo VM for image pruning")
 		if err := repo.startVM(); err != nil {
 			return upgrade, err
 		}
+
+		// moved the restoreVMState call after the start repo vm as we need to make sure kubevirt is up and running
+		// thus we can bring up the vm to the state before upgrade
+		h.restoreVMState(upgrade)
 
 		if err := h.cleanupImages(upgrade, repo); err != nil {
 			logrus.Warningf("Unable to cleanup images: %s", err.Error())
@@ -246,6 +245,8 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 	// upgrade failed
 	if harvesterv1.UpgradeCompleted.IsFalse(upgrade) {
 		// try to restore vm to the state before upgrade
+		// we didn't wait for KubeVirt to reach the Deployed phase because the upgrade failed,
+		// as a result, some services might not be ready and may never fully start.
 		h.restoreVMState(upgrade)
 		// clean upgrade repo VMs.
 		return nil, h.cleanup(upgrade, harvesterv1.UpgradeCompleted.IsTrue(upgrade))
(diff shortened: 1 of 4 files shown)


brandboat:
@mergify backport v1.4

pkg/webhook/resources/setting/validator.go
```diff
@@ -167,6 +165,9 @@ func NewValidator(
 	validateSettingFuncs[settings.HTTPProxySettingName] = validator.validateHTTPProxy
 	validateSettingUpdateFuncs[settings.HTTPProxySettingName] = validator.validateUpdateHTTPProxy
 
+	validateSettingFuncs[settings.UpgradeConfigSettingName] = validator.validateUpgradeConfig
```
brandboat, line 168: Turn this into draft as https://github.com/harvester/harvester/issues/6670 reminds me a scenario:

Assume we have a single-node cluster and set `restoreVM` to true in the UpgradeConfig setting. Now, we're about to perform an upgrade:
1. Before the upgrade, vm1 is in the running state. The `storeVM` function adds vm1 to the upgrade annotation.
2. During the upgrade, vm1 is stopped.
3. Once the upgrade is successfully completed, restoreVM starts vm1, and vm1 is in the starting state.
4. If we start another upgrade immediately, storeVM might not add vm1 to the upgrade annotation since it's still in the starting state.

There are two ways to fix this:

1. Wait until all VMs in the annotation are in the running state inside the restoreVM method.
2. Add a validation in the upgrade webhook and verify when restoreVM set to true there is no VMs are in the starting state before initiate an upgrade.

I personally prefer option 2, as it's simpler to implement.

c.c. @bk201, @w13915984028, @starbops

w13915984028: As you said, a quick fix is to add an annotation to represent the last actions, and webhook check this. It can protect both vm image purging and vm restoring, and user can also manually delete it to stop if there is any kind of bugs.

A formal solution is to put those actions under a new state (upgrade crd already has many states) with time-out mechnism.

brandboat: > A formal solution is to put those actions under a new state (upgrade crd already has many states) with time-out mechnism.

As @w13915984028 pointed out, let's address this in [issue #6670](https://github.com/harvester/harvester/issues/6670) and keep the current behavior unchanged for now. I won’t be adding any further enhancements in this PR to avoid unnecessary complexity. Currently, if a user in a single-node cluster wants to perform a series of upgrades with restoreVM set to true, they will need to wait for the VMs to fully start before proceeding with the next upgrade.

> Currently, if a user in a single-node cluster wants to perform a series of upgrades with restoreVM set to true, they will need to wait for the VMs to fully start before proceeding with the next upgrade.

And I'll mention this in the doc, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -792,101 +795,58 @@ func (h *upgradeHandler) storeVMState(upgrade *harvesterv1.Upgrade) error {
...
-	for _, preUpgradeRunningVM := range preUpgradeRunningVMs {
-		vm, err := h.vmCache.Get(preUpgradeRunningVM.Namespace, preUpgradeRunningVM.Name)
+	for _, vmInfo := range preUpgradeRunningVMs {
+		vm, err := h.vmCache.Get(vmInfo.Namespace, vmInfo.Name)
 		if err != nil {
-			logrus.WithFields(logrus.Fields{
-				"namespace": preUpgradeRunningVM.Namespace,
-				"name":      preUpgradeRunningVM.Name,
-			}).WithError(err).Error("Failed to get VM from cache")
+			logrus.WithFields(logFields).WithError(err).Errorf("Failed to get VM %s/%s from cache", vmInfo.Namespace, vmInfo.Name)
```
Vicente-Cheng, line 842: If the error happens, we skip this VM. This means we might not ensure the Running VM can Run after the upgrade (somehow get VM failure). Is that right?

If yes, does this assumption meet our expectations?

pkg/webhook/resources/setting/validator.go
```diff
@@ -1483,3 +1493,15 @@ func validateAdditionalGuestMemoryOverheadRatio(newSetting *v1beta1.Setting) err
 func validateUpdateAdditionalGuestMemoryOverheadRatio(_ *v1beta1.Setting, newSetting *v1beta1.Setting) error {
 	return validateAdditionalGuestMemoryOverheadRatio(newSetting)
 }
+
+func (v *settingValidator) isSingleNode() (bool, error) {
+	nodes, err := v.nodeCache.List(labels.Everything())
+	if err != nil {
+		return false, err
+	}
+	if len(nodes) == 1 {
```
Vicente-Cheng:
```suggestion
	return len(nodes) == 1, nil
```
Is it more readable/clean?

bk201 (approved):
LGTM

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Ideally, there should be a retry mechanism in place. However, the current implementation lacks this, and the restoreVM process is intertwined with image cleanup, making it difficult to handle retries effectively. I'll refactor the entire post-upgrade process as part of [issue #6670](https://github.com/harvester/harvester/issues/6670).

As for your question—yes, this behavior is expected. If something goes wrong, the error is ignored, and the process moves on to start the next VM.

pkg/webhook/resources/setting/validator.go
brandboat: yes, thanks for the comment.

pkg/controller/master/upgrade/upgrade_controller.go
Vicente-Cheng: Thanks for the explanation!

brandboat:
rebased to master and manually test again, all good

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -792,101 +795,58 @@ func (h *upgradeHandler) storeVMState(upgrade *harvesterv1.Upgrade) error {
...
-			OwnerReferences: []metav1.OwnerReference{
-				upgradeReference(upgrade),
-			},
-		},
-		Data: map[string][]byte{
-			"vms": runningVMsJSON,
-		},
-	}
-	return h.createOrUpdateSecret(secret)
+	upgrade.Annotations[preUpgradeVMsAnnotation] = string(runningVMsJSON)
```
starbops, line 805: Is it possible, at this time point, that the annotation is still `nil`?

starbops (commented):
LGTM, just one nit. Thank you.

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: yeah that might happen, I'll fix that, thanks.

starbops (approved):

Vicente-Cheng (approved):
LGTM, thanks!

PR #6698, opened 2024-10-03
ci: pass fixed-length short revision

bk201:
The short revision has to be in fixed length (8) to let the harvester upgrade controller pick up the correct upgrade image.

**Problem:**

Harvester can't upgrade from a stable branch. 

**Solution:**

The cause is the mismatch between the server-version and the image tag. For example: in a crime scene we found the server version is v1.4-cd10d18-head, but the image preloaded on the node has the tag v1.4-cd10d18d-head (hash length different). 
- The server version is defined in https://github.com/harvester/harvester/blob/650dff194086acb3922a1575d284fbc7c8e0fd2b/.github/workflows/build.yml#L60, hash length is undetermined.
- The upgrade image tag is defined in https://github.com/harvester/harvester-installer/blob/f6d189bfc828879096e28786cfd175ec5a439bab/scripts/build-bundle#L333, hash length is 8.

We can specify the short hash length in the ci pipeline.

**Related Issue:**

https://github.com/harvester/harvester/issues/6699

**Test plan:**

The test seems to only be done after the PR is merged and the CI publishes a new ISO.
- Download the published ISO.
- Upgrade to other versions, the preload job should work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
 
     - name: Declare branch and sha_short
       run: |
-        echo "sha_short=$(git rev-parse --short "$GITHUB_SHA")" >> "$GITHUB_ENV"
+        echo "sha_short=$(git rev-parse --short=8 "$GITHUB_SHA")" >> "$GITHUB_ENV"
         echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> "$GITHUB_ENV"
         if [ "${{ matrix.arch }}" == "x64" ]; then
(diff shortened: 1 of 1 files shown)


Yu-Jack (approved):
LGTM.

starbops (approved):
LGTM, thank you.

tserong (approved):
~The change looks fine to me, but I don't understand how we ended up with mismatched hash lengths in the first place.~ Oh, wait, I found it! https://github.com/harvester/harvester/blob/fe5f3ebd0ec61bbd6c986283a8a535333dbd33b7/scripts/version#L11 :-)

Issue #6257, opened 2024-08-05
[ENHANCEMENT] Upgrade support of ARM64


bk201:
**Is your enhancement related to a problem? Please describe.**

To make arm64 port GA, we need to support upgrading an arm64 cluster.

**Describe the solution you'd like**

Some enhancements to do
- The upgraded responder should fetch correct version file: https://github.com/harvester/harvester/blob/e269dfe35b0cb822e2c2801c435fed4d9a3d2523/pkg/controller/master/upgrade/version_syncer.go#L201
- enable addons for arm. currently they are only packaged for amd64 based installs

**Describe alternatives you've considered**

**Additional context**

lanfon72:
is this means we can upgrade from `v1.3.2` to `v1.4.0` or support upgrade starting from `v1.4.0`?

ibrokethecloud:
The upgrade will be possible from 1.4.0 onwards only, as the upgrade related changes will only be made available in 1.4.0

TachunLin:
@ibrokethecloud 
Can we verify this issue with the upgrade path?  `v1.4.0-rc` -> `v1.4.0-rc` or `v1.4-head` -> `v1.4-head`

TachunLin:
Verified **failed** on `v1.4.0-rc4`. Move to implement.

### Result
<details><summary>$\color{red}{\textsf{Failed}}$ Single node upgrade from v1.4.0-rc4 to v1.4.0-rc4 stuck in <strong>image preloading</strong> stage $~~$</summary>

  1. On ARM based machine
  ![image](https://github.com/user-attachments/assets/bd82cbe8-67a6-4aca-bd75-d58d663a1617)

  2. The upgrade to v1.4.0-rc4 stuck in `image preloading` stage

  ![image](https://github.com/user-attachments/assets/25fa82f2-8253-4a75-beea-b1a6cc15d7fc)

  
  3. Check the job `apply-hvst-upgrade-j2484-prepare-on-arm133-140rc4-with-a2-tr59m`, it continues to bring up the upgrade repo VM but failed with message
  
  ```
   ++ kubectl get virtualmachines.kubevirt.io upgrade-repo-hvst-upgrade-j2484 -n harvester-system '-o=jsonpath={.status.printableStatus}'                     │
  │ /usr/local/bin/lib.sh: line 140: /usr/bin/kubectl: cannot execute binary file: Exec format error                                                           │
[...]

PR #6707, opened 2024-10-04
arm upgrade path fixes

ibrokethecloud:
fix up version lookup and minor changes to upgrade vm spec to use iscsi bus for iso devices to enable support across both arm64 and amd64 arch

**Problem:**

Currently when triggering arm upgrades via the version sync, the sync fails as older versions of harvester do not have a corresponding version-arm64.yaml published

**Solution:**

PR introduces following changes
* ignores version object if a corresponding version.yaml or version-arm64.yaml is not found
* minor changes to upgrade VM to drop q35 machine type so correct type can be identified based on underlying platform arch
* minor change to switch VM iso bus adapter to scsi, to ensure it is compatible across both amd64 and arm64 arch

**Related Issue:**
https://github.com/harvester/harvester/issues/6257
**Test plan:**

### DIFF
--- a/pkg/controller/master/upgrade/version_syncer.go
+++ b/pkg/controller/master/upgrade/version_syncer.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 
 	gversion "github.com/mcuadros/go-version"
@@ -158,6 +159,9 @@ func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion s
 	for _, v := range resp.Versions {
 		newVersion, err := s.getNewVersion(v)
 		if err != nil {
+			if strings.Contains(err.Error(), "failed to download version") {
+				continue
+			}
 			return err
 		}
 
--- a/pkg/controller/master/upgrade/upgrade_repo.go
+++ b/pkg/controller/master/upgrade/upgrade_repo.go
@@ -218,7 +218,7 @@ func (r *Repo) createVM(image *harvesterv1.VirtualMachineImage) (*kubevirtv1.Vir
 									BootOrder: &bootOrder,
 									DiskDevice: kubevirtv1.DiskDevice{
 										CDRom: &kubevirtv1.CDRomTarget{
-											Bus: "sata",
+											Bus: "scsi",
 										},
 									},
 									Name: "disk-0",
@@ -226,7 +226,7 @@ func (r *Repo) createVM(image *harvesterv1.VirtualMachineImage) (*kubevirtv1.Vir
 								{
 									DiskDevice: kubevirtv1.DiskDevice{
 										CDRom: &kubevirtv1.CDRomTarget{
-											Bus: "sata",
+											Bus: "scsi",
 										},
 									},
 									Name: "cloudinitdisk",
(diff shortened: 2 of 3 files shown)


pkg/controller/master/upgrade/version_syncer.go
```diff
@@ -158,6 +159,9 @@ func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion s
 	for _, v := range resp.Versions {
 		newVersion, err := s.getNewVersion(v)
 		if err != nil {
+			if strings.Contains(err.Error(), "failed to download version.yaml") {
```
tserong: This will match any HTTP failure code (`resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusBadRequest`), so it will just skip over things like 500 internal server error. Is that too loose, i.e. should we only be explicitly skipping 404s? Or am I being too picky? :-)

tserong (commented):
LGTM, just one question

pkg/controller/master/upgrade/version_syncer.go
ibrokethecloud: I do not have a strong opinion about this. If we error out the sync job fails and we retry again after the sync interval. In this scenario we just keep trying to find the next version and hopefully if the error resolves itself the next sync interval will pick it up.

tserong: Cool, sounds fine to me.

tserong (approved):

bk201 (approved):

pkg/controller/master/upgrade/version_syncer.go
Vicente-Cheng: If we have multiple upgrade versions, how do we make sure the upgrade version is we wanted if we do not error out here?

e.g. We have two versions, A and B. The A is newer than the B. But we got `failed to download version.yaml` with A, we chose B to upgrade. Is that what we want, or would this case not happen?

Vicente-Cheng (approved):
just a nit discussion, others LGTM

pkg/controller/master/upgrade/version_syncer.go
starbops: Do we also need to consider the case where the error message is `"failed to download version-arm64.yaml"`? I feel the same way as Tim; we only need to skip 4xx errors. For others, it might be a server-level issue, so there is no need to try other versions quickly.

ibrokethecloud: currently we do not retry immediately when there is an error, the default behaviour is to log the error and retry in an hour, so this behaviour would continue as is

```
func (s *versionSyncer) start() {
	ticker := time.NewTicker(syncInterval)
	for {
		select {
		case <-ticker.C:
			if err := s.sync(); err != nil {
				logrus.Warnf("failed syncing upgrade versions: %v", err)
			}
		case <-s.ctx.Done():
			ticker.Stop()
			return
		}
	}
}
```

ibrokethecloud: i should probably just change the message to `failed to download version` so it is a valid substring for both arm64 and amd64 arch's

ibrokethecloud:
@Vicente-Cheng the actual check if version is applicable for the currently running harvester cluster happens after fetching the versions in the block
```
// newVersion CRD is created from the version.yaml created from the release artifact
		// this needs to be changed to use the min version coming from the upgrade responder
		// as this could be dynamically update and will cause cleanup of older versions
		if !canUpgrade(currentVersion, newVersion, v) {
			continue
		}
```

so even if a version is fetched and is not valid it will be skipped.

pkg/controller/master/upgrade/version_syncer.go
starbops: I'm just being picky 😆
The updated version looks fine to me.

starbops (approved):

Issue #6727, opened 2024-10-08
[BUG] Improved Harvester Maintenance Mode broken


votdev:
**Describe the bug**
The [Improving Harvester Maintenance Mode](https://github.com/harvester/harvester/pull/5328) feature has been broken by another [PR](https://github.com/harvester/harvester/pull/5991) [by changing the logic](https://github.com/harvester/harvester/pull/5991/files#diff-668050f872b498d877736c7b008f2bdf89bbd48800f9ef83d2d3da8e57a1d872L147) which is responsible for shutting down VMs.

The for loop which iterates over the VMs to be shut down was probably mistakenly pulled in an if condition, which is unfortunately wrong and makes the previous logic inoperable.

**To Reproduce**
Replay the [test cases](https://github.com/harvester/harvester/issues/5069#issuecomment-2367154971).

**Expected behavior**
The Improving Harvester Maintenance Mode [should work as expected](https://github.com/harvester/harvester/pull/5328#issue-2176203298).

**Environment**
 - Harvester ISO version: v1.4.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): ixpe 3 node cluster

tserong:
(apologies, that close shouldn't have happened - my bad)

w13915984028:
Test report: **PASS**

test steps: refer: https://github.com/harvester/harvester/pull/5328#issue-2176203298

---
:green_circle: **Case 1: Label `harvesterhci.io/maintain-mode-strategy: ShutdownAndRestartAfterDisable`**

 Add label to VM and start VM
```
  labels:
    harvesterhci.io/creator: harvester
    harvesterhci.io/maintain-mode-strategy: ShutdownAndRestartAfterDisable
    harvesterhci.io/os: linux
```
![image](https://github.com/user-attachments/assets/c5bf976e-f551-4b5f-96b5-807640913172)

put node to `maintenance mode`:
![image](https://github.com/user-attachments/assets/db660227-f386-4afd-bb58-69a1cd85d843)

![image](https://github.com/user-attachments/assets/fe081af6-c3e9-4311-9e60-188bb454fe77)

vm becomes off:
![image](https://github.com/user-attachments/assets/2f327844-1f96-469f-bc9d-b65b307473af)

Disable `maintenance mode`
![image](https://github.com/user-attachments/assets/967ac925-6d29-481a-a486-e47640d7059f)

VM starts
[...]

PR #6729, opened 2024-10-08
Improved Harvester Maintenance Mode broken

votdev:
**Problem:**
The [Improving Harvester Maintenance Mode](https://github.com/harvester/harvester/pull/5328) feature has been broken by another [PR](https://github.com/harvester/harvester/pull/5991) [by changing the logic](https://github.com/harvester/harvester/pull/5991/files#diff-668050f872b498d877736c7b008f2bdf89bbd48800f9ef83d2d3da8e57a1d872L147) which is responsible for shutting down VMs.

The for loop which iterates over the VMs to be shut down was probably mistakenly pulled in an if condition, which is unfortunately wrong and makes the previous logic inoperable.

**Solution:**
Adapt the original code to the new conditions.

**Related Issue:**
https://github.com/harvester/harvester/issues/6727

**Test plan:**
See https://github.com/harvester/harvester/pull/5328

### DIFF
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -3,7 +3,7 @@ package nodedrain
 import (
 	"context"
 	"fmt"
-	goslices "slices"
+	"slices"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


ibrokethecloud (approved):

bk201 (approved):

ibrokethecloud:
@mergifyio backport v1.3

PR #6735, opened 2024-10-09
feat: add secret validator for encryption storage class 

Yu-Jack:

**Problem:**
1. Missing secret validator when creating an encrypted storage class.
2. Insufficient details on how to check encrypted volumes.

**Solution:**
1. Add secret validator for encryption storage class.
3. Provide more information on how to check encrypted volume.

**Related Issue:**
- https://github.com/harvester/harvester/issues/3129#issuecomment-2400756896
- https://github.com/harvester/harvester/issues/6493#issuecomment-2400758163

**Test plan:**
1. Create a secret that is missing some keys.
2. Create a storage class by selecting the previously created secret.

### DIFF
--- a/pkg/webhook/resources/storageclass/validator.go
+++ b/pkg/webhook/resources/storageclass/validator.go
@@ -5,10 +5,13 @@ import (
 	"strconv"
 	"strings"
 
+	lhcrypto "github.com/longhorn/longhorn-manager/csi/crypto"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhtypes "github.com/longhorn/longhorn-manager/types"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	ctlstoragev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/storage/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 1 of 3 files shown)


bk201 (approved):

WebberHuang1118 (commented):
LGTM, just a nit, thanks.

https://github.com/Yu-Jack/harvester/blob/bd5ac8e02f9d0fa4c315a90089d6abf594d27b81/pkg/webhook/resources/storageclass/validator.go#L232-L267

This function seems not only return the missing parameters ? I suggest to change the function name to match the return values.

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -269,3 +309,7 @@ func (v *storageClassValidator) validateReservedStorageClass(sc *storagev1.Stora
 
 	return nil
 }
+
+func (v *storageClassValidator) getEncryptionSecret(sc *storagev1.StorageClass) (secretName string, secretNamespace string) {
+	return sc.Parameters[pairs[0][0]], sc.Parameters[pairs[0][1]]
+}
```
WebberHuang1118: If we take this function alone, it doesn't check the length of pairs, which I consider a risk probably. What if combine this with the function `validateEncryptionParams()`? In the function `validateEncryptionParams()` it can still return the related secret.

Yu-Jack: I just removed it. I think it might be confused to return something (except error) when function name is `validate`.


This one should be enough.
```go
err = v.validateEncryptionParams(newSC)
if err != nil {
	return err
}

secretName, secretNamespace := newSC.Parameters[pairs[0][0]], newSC.Parameters[pairs[0][1]]
```

WebberHuang1118 (approved):

PR #6739, opened 2024-10-09
make VM/Volume snapshot works well on the LVM volume

Vicente-Cheng:

**Problem:**
If the VM attaches the LVM volume, it cannot take snapshot even if the LVM CSI driver supports the snapshot.

**Solution:**
Make sure the VM/Volume snapshot path works.

**Related Issue:**

**Test plan:**
1. create harvester cluster
2. enable LVM with addon (REF: https://github.com/harvester/experimental-addons/pull/18)

~3. create VolumeSnapshotClass by following yaml~
```
apiVersion: snapshot.storage.k8s.io/v1
deletionPolicy: Delete
driver: lvm.driver.harvesterhci.io
kind: VolumeSnapshotClass
metadata:
  name: lvm-snapshot
```

4. Modify the settings `csi-driver-config` as below:
```
default: '{"driver.longhorn.io":{"volumeSnapshotClassName":"longhorn-snapshot","backupVolumeSnapshotClassName":"longhorn"}}'
kind: Setting
metadata:
  name: csi-driver-config
status: {}
value: '{"lvm.driver.harvesterhci.io":{"volumeSnapshotClassName":"lvm-snapshot"}}'
```
5. make sure the snapshot feature works (for both VM/Volume)

~We would like to make steps 3 and 4 can be automatically applied when we enable the addon.
But for now, we need to manually config it.~

**UPDATE** Use the latest addon yaml to complete steps 3 when enable the addon.

### DIFF
--- a/pkg/controller/master/backup/backup.go
+++ b/pkg/controller/master/backup/backup.go
@@ -396,8 +396,18 @@ func (h *Handler) getVolumeBackups(backup *harvesterv1.VirtualMachineBackup, vm 
 // getCSIDriverMap retrieves VolumeSnapshotClassName for each csi driver
 func (h *Handler) getCSIDriverMap(backup *harvesterv1.VirtualMachineBackup) (map[string]string, map[string]snapshotv1.VolumeSnapshotClass, error) {
 	csiDriverConfig := map[string]settings.CSIDriverInfo{}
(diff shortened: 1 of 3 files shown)


WebberHuang1118 (approved):
LGTM, verified with VM Snapshot/Restore. Thanks for the extension.

FrankYang0529 (approved):
LGTM. Thank you.

Issue #2322, opened 2022-05-27
[FEATURE] Support VXLAN network


guangbochen:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Add VXLAN network with adding the following advanced features:
- Support network policies for micro-segmentation of workloads #2260
- Support multiple VXLANs with separated physical NICs.
- Support VM live migration and persistent IP address.
- Integrate with kube-vip to provide load balancer service for VMs based on vxlan network. 
- Allow provisioning guest k8s cluster and manage the cluster with Rancher through the vxlan network.

alejandroed:
It would be great to be able to use bgp unnumbered to do routing on the host and support full L3 fabrics, bonds are things of the past

yasker:
Due to the large level of effort, we won't be able to make this into v1.1.0.

Move it to v1.2.0.

moroniflores:
> Move it to v1.2.0.

When is v1.2.0 expected to be released?

abonillabeeche:
> > Move it to v1.2.0.
> 
> When is v1.2.0 expected to be released?

Early Summer 2023 perhaps.

yasker:
Yeah, H1 2023 is the planned date for v1.2.0

yasker:
The VXLAN feature has been postponed to v1.3.0 which is targeting H2 2023.

bk201:
Another request to ease the deployment on Hetzner https://github.com/harvester/harvester/issues/4622#issue-1945947885

> Hello together,
> I am currently playing with Harvester. My initial intent was to deploy Harvester in a production setting at a Hetzner datacenter. Unfortunately, I am seeing 2 show stopper for such a setup. Therefore, for now I switched to a Proxmox installation. Nevertheless, my hope is that soon missing pieces in Harvester come together, as it to my opinion is a very nice product. The first show stopper is described here https://github.com/harvester/harvester/issues/4608 , hence nothing to add.
> The second show stopper is related to the amount of MAC addresses I can use per switch port when using VLans (Hetzner vSwitch feature). The amount of MAC addresses is limited to 30 addresses per switch port. In other words, with one network card I can only support 30 VMs. In Proxmox I can simply create an overlay network with VxLan. Hence, this limitation is resolved with a little overhead. Would be great if VxLan could be made available in Harvester as well.

mingshuoqiu:
[...]

PR #6744, opened 2024-10-09
VxLAN HEP

mingshuoqiu:

**Problem:**
Enable the network-controller to support basic VxLAN configuration

**Solution:**
Use the current `ip` and `bridge` utility to attach the VM's NIC to the VxLAN and communicate to each other 

**Related Issue:**
https://github.com/harvester/harvester/issues/2322

**Test plan:**


enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+if the VM is on the Mgmt NODE, the following arp entry needs to be added on
+NODE_2/NODE_3:
+sudo ip neigh add 10.42.0.21 lladdr 02:11:11:11:11:11 dev vxlan0
+
+if the VM is on the NODE_2, then we need to add the ARP entry on
+Mgmt NODE/NODE_3:
+sudo ip neigh add 10.42.0.21 lladdr 02:22:22:22:22:22 dev vxlan0
+```
+
+(6) Use the VxLAN interface created in (3) to play as the router for all VMs attached to the same VxLAN network on the NODE.
```
w13915984028, line 158: When external access a VM/guest-cluster actively, how ?

w13915984028 (commented):
Thaks for the HEP, there are ome open scenarios to address:

(1) How can VMs on a VxLan be accessed from external. In K8s, the kubeclt/kube-proxy/service/load-balancer/... do the trick for PODs. For the Rancher node-driver scenario, Rancher needs to access at least the VIP of the guest cluster. For node IP (VM IP), accessing is required when troubleshooting.

(2) What's the impact for current Harvester VM type loadbalancer, it is backed by the ARP mode kube-vip now. This is also related to (1).

(3) Can storge-network also work on VxLan?

(4) Suppose it is similar to Vlan mode, user needs to set which nics are running for (if they are isolated from management network) which VxLan type of cluster-network, and also set the IP mode, static or DHCP. Are IP also required to be stable on the life-cycle of a cluster-network?

(5) Cross-VxLan communication (one guest-cluster to another guest-cluster), supported or not?

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+on NODE_3:	add Mgmt NODE and NODE_2 
+bridge fdb append 00:00:00:00:00:00 dev vxlan42 dst 192.168.5.11
+bridge fdb append 00:00:00:00:00:00 dev vxlan42 dst 192.168.5.22
+bridge fdb append 02:11:11:11:11:11 dev vxlan42 dst 192.168.5.11
+bridge fdb append 02:22:22:22:22:22 dev vxlan42 dst 192.168.5.22
+```
+
+(5) The VxLAN interface on each NODE should play as the ARP proxy for VMs attached to the same VxLAN network.
+Then the packet destined to the VM have the ARP response from the corresponding VTEP and correctly forwarded.
+Adding the ARP entry for each VM for other NODEs which has no information for it.
```
tjjh89017, line 147: I didn't get this.
Your `vxlan0` interface will be bridge slave in this setup.
The ARP entry will not be used in vxlan0 interface for tenant network only for VTEP.
If you want "arp suppression" feature, you will need to assign IP address on every vxlan bridge and enable "arp suppression".
Your bridge will learn ARP from ARP response and record in bridge's ARP table.
If your bridge have ARP entry, it will reply the ARP request directly next time.
But in most of time, if you didn't use BGP-EVPN, I will not suggest to enable any ARP suppression feature.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+
+if the VM is on the NODE_2, then we need to add the ARP entry on
+Mgmt NODE/NODE_3:
+sudo ip neigh add 10.42.0.21 lladdr 02:22:22:22:22:22 dev vxlan0
+```
+
+(6) Use the VxLAN interface created in (3) to play as the router for all VMs attached to the same VxLAN network on the NODE.
+The `vxlan` interface can connect the external network via the uplink/bonding interface of the guest ClusterNetwork.
+Thus all VMs which route via the vxlan interface can connect to the external network.
+The `iptables` rules which do the masquerade should be added and the `default` route entry should also be configured on the attached VM.
```
tjjh89017, line 161: You should not add your GW ip address on vxlan interface if that interface is a bridge slave.

Another issue, do you consider having two VNI with the same IP range and internet access?
If so, your host route table will be confused if you have multiple overlap subnet.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+(3) network-controller need to keep tracking the VTEP IP/MAC mapping information for all NODEs. 
+
+### Test plan
+
+
+### Upgrade strategy
+
+## Note [optional]
+
+(1) to enable cross VxLAN network connection (different VNI) on the same guest Cluster Network, the BGP EVPN is required to route for each VxLAN group. Need to figure out a way for it.
```
tjjh89017, line 190: No, you don't need any BGP-EVPN to get cross vxlan network connection.
BGP-EVPN only helps you to configure route table correctly.
What you need is L3VNI. (Please google it, you don't need BGP-EVPN to do this, just because configuring manually is complex)
You still can configure those route table and configuration on your own.
Even you can write a K8s Controller to replace BGP-EVPN's role.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+
+(2) Each `vxlan` interface requires an additional IP address assigned for the spcified VNI bound to the VNI. How to allocate that automatically from IPAM or DHCP?
```
tjjh89017, line 192: I'm not sure why you need IP address on vxlan interface.
Again, you should not assign any IP address on VXLAN interface but Bridge interface.

tjjh89017 (commented):
Overall, you will need to check Linux VXLAN configuration and how it works, some of your vxlan concept are incorrect.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+  - Support each Virtual Network is used for one VM Group
+  - Support VM live migration and persistent IP address, VM can freely migrate/move in the Harvester Cluster without change of IP address
+  - Support VM Group use VIP to communicate with outside
+- Support automatically establishing of VxLAN tunnel, all host NODEs in the specific Cluster Network are member of VxLAN network and should be able to build full mesh VxLAN tunnel connection
+- Enable efficient network segmentation and traffic management across different network segments.
+
+### Non-goals [optional]
+
+- Integrate with kube-vip to provide load balancer service for VMs based on vxlan network.
+- Allow provisioning guest k8s cluster and manage the cluster with Rancher through the vxlan network.
```
rrajendran17, line 39: 1.Are we supporting communication between VMs in different VNIs? 
2.Are we supporting tagged traffic from VMs ?
If not those can be added to the Non-goals section.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Proposal
+.
+
+### User Stories
+
+In Harvester, we assume the following:
+
+ - The user provides physical network connectivity to make sure echo Node are reachable to each other
```
rrajendran17, line 48: is "echo" typo ? Is it all nodes in the cluster

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+ - Multi networks may exist.
```
rrajendran17, line 49: what is "multi networks" in this context ?

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+A detailed explanation of Harvester networking is in the Harvester network samples.
+
+A user plans to deploy a Harvester cluster and assumes:
+
+ - Mgmt is a built-in cluster network and does not support any additional vxlanconfigs.
+ - Users are allowed to add a custom cluster network and add vxlanconfigs under it.
+ - Users should select cluster network when creating a VxLAN network.
+ - User configures multiple overlay networks using VxLAN VNI on the selected cluster network
+ - User has multiple nodes, each with multiple NICs.
+ - User requires strict traffic policies to be enforced between VMs and with external networks to ensure security and compliance.
```
rrajendran17, line 61: Is this "defining traffic policies" part of this enhancement ?

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+
+
+#### Story 1
+
+Adding Virtual Network
+
+![](./20240930-vxlan-enhancement/vxlan_vm_network_setup.jpg)
+User specifies the VNI and the related CIDR, VxLAN interfacde are created automatically and attached to the bridge which belongs to the Cluster Network..
```
rrajendran17, line 69: Referring to the UI images attached, I understand in the UI we will keep options for user to differentiate and use it either for vlan networks or vxlan networks,so that we could either create vlan network the usual way and vxlan network by differentiating its a virtual network for vxlan and IP is for the VTEP.

I see from the proposal, we are going to use VTEPs in same subnet for now.
If that is the case, can you add VTEPs in different subnet as non-goal.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+#### Story 3
+
+VM-internet communication
+
+First, make sure the guest Cluster Network is able to connect to the internet via the bonding interface.
+Then, the VxLAN interface attached to the bridge should play as a router of All VMs which attached to the same VM Network (the same VNI)
+The iptables rule for the routing should be added on the NODE, and the default route entry in the VM should be destined to the IP of the VxLAN interface
+All VMs attached to the internet can be done by the 
+
+
```
rrajendran17, line 96: We also need a user story for virtual network's VNI address range.
We need two IP addresses range.
1.For the VTEPs (story 1)
2.For the VMs (current design, VMs get IP address from DHCP server based on vlan network) 
How are we planning to get IP addresses for VM interfaces ?

Current harvester VMs, get address from DHCP server for secondary interfaces attached to them based on VM VLan Network.
In case of vxlan forwarding, VM interfaces once attached to vxlan virtual network should get an IP address range from the subnet provided. This will make sure to allocate IPs to VMs from same subnet pool assigned to a VNI.

we might need a change/enhancement in UI to provide this range.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+#### An example of VxLAN user operation
+
+The user will be able to use Harvester VxLAN network to create isolated Guest Kubernetes Cluster with following steps.
+![](./20240930-vxlan-enhancement/vxlan_three_nodes_setup_detail.png)
+(1) Create one guest Cluster Network on a dedicated NIC(different from the management NIC), IP addresses are static set/dynamic provisioned by DHCP from provider network, CIDR e.g. 192.168.5.0/24
+
+(2) Create one VxLAN overlay network on top of the guest Cluster Network, say VxLAN_1, VNI 42, the overlay network CIDR, e.g. 10.42.0.0/8. The default port number can be configured manually or automatically. The port 8472 is occupied for `flannel` by default. The port 4789 is used for demonstration in the following operations. The Local IP and remote peers of the VTEP in each NODE are fetched from APIServer. The vxlan interface will be created with the specified VNI on the NODE.
+```
+ip link add vxlan42 type vxlan id 42 dstport 4789 local 192.168.5.11 dev cn-vxlan_1-bo
```
rrajendran17, line 107: is it mandatory to provide destination port ?

rrajendran17, line 107: The example does not specify a remote VTEP. Is this expected ?
How are remote VTEPs learnt/configured on a node ?
Is it statically configured as remote peer during vxlan interface creation or all remote peers learnt dynamically ?

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+bridge fdb append 00:00:00:00:00:00 dev vxlan42 dst 192.168.5.33
+bridge fdb append 02:11:11:11:11:11 dev vxlan42 dst 192.168.5.11
+bridge fdb append 02:33:33:33:33:33 dev vxlan42 dst 192.168.5.33
+
+on NODE_3:	add Mgmt NODE and NODE_2 
+bridge fdb append 00:00:00:00:00:00 dev vxlan42 dst 192.168.5.11
+bridge fdb append 00:00:00:00:00:00 dev vxlan42 dst 192.168.5.22
+bridge fdb append 02:11:11:11:11:11 dev vxlan42 dst 192.168.5.11
+bridge fdb append 02:22:22:22:22:22 dev vxlan42 dst 192.168.5.22
+```
```
rrajendran17, line 143: are these macs belong to VMs?So we statically configure macs of VMs against VTEP, so that traffic to a particular vm will be sent to only those VTEPs?

Also we could use BGP EVPN here to populate mac to IP mapping on nodes whenever a VM is attached to a vxlan virtual network.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+Then the packet destined to the VM have the ARP response from the corresponding VTEP and correctly forwarded.
+Adding the ARP entry for each VM for other NODEs which has no information for it. 
+```
+if the VM is on the Mgmt NODE, the following arp entry needs to be added on
+NODE_2/NODE_3:
+sudo ip neigh add 10.42.0.21 lladdr 02:11:11:11:11:11 dev vxlan0
+
+if the VM is on the NODE_2, then we need to add the ARP entry on
+Mgmt NODE/NODE_3:
+sudo ip neigh add 10.42.0.21 lladdr 02:22:22:22:22:22 dev vxlan0
```
rrajendran17, line 155: Is this static arp entry needed at the host ?
My understanding is, neighbours will be learnt for the  VM Ips through vxlan interface.
Later vxlan needs only src vtep ip, dst vtep ip, src mac of vtep interface and dst mac of remote vtep intf (configured through static arp or dynamic) to communicate between VMs in same VNI.

enhancements/20240930-vxlan-enhancement.md
rrajendran17, line 190: BGP EVPN is used when we need to add vm mac to VTEP IP mapping, so that packet is not flooded to all VTEPs instead send only to required VTEPs.
So flood learning is avoided using BGP EVPN.

Vxlan L3 gateways, similar to vlan IRBs could be used to forward traffic between different VNIs.
We need to configure GWs so that it could decapsulate a VNI and encapsulate another VNI to be forwarded.

enhancements/20240930-vxlan-enhancement.md
rrajendran17, line 192: Is this about the overlay IP range for the VNI (from which ip address allocated to vms) ?
eg, from your config, 10.42.0.0/24 network

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+### Upgrade strategy
+
+## Note [optional]
+
+(1) to enable cross VxLAN network connection (different VNI) on the same guest Cluster Network, the BGP EVPN is required to route for each VxLAN group. Need to figure out a way for it.
+
+(2) Each `vxlan` interface requires an additional IP address assigned for the spcified VNI bound to the VNI. How to allocate that automatically from IPAM or DHCP?
+
+(3) Use the ovn-kubernetes for the overlay network deployment, which includes the VxLAN and more.
+
```
rrajendran17, line 195: You could also add detailed images with VMs, vtep mac mapping examples if possible.
And some example use case scenarios like same VTEPs with multiple VNIs, VNIs mapped to multiple VTEPs,
Broadcast and unicast scenario.

enhancements/20240930-vxlan-enhancement.md
rrajendran17: arp entry for remote VTEP IP, is it learnt dynamically or statically ? If so when will it be initiated/configured during the vxlan intf creation ?

enhancements/20240930-vxlan-enhancement.md
mingshuoqiu: 1. No. Currently we don't support VMs which belongs to different VNIs. It route require routing information based on the VxLAN interface.
2. Yes. The tagged VLAN can be encap in the VxLAN packet and communcate in the same VNI group

enhancements/20240930-vxlan-enhancement.md
mingshuoqiu: It refers to different `underlying network`.

enhancements/20240930-vxlan-enhancement.md
mingshuoqiu: No. The traffic policy would relies on the functions which is generically supported by Kubernetes.

enhancements/20240930-vxlan-enhancement.md
```diff
@@ -0,0 +1,197 @@
...
+Adding Virtual Network
+
+![](./20240930-vxlan-enhancement/vxlan_vm_network_setup.jpg)
+User specifies the VNI and the related CIDR, VxLAN interfacde are created automatically and attached to the bridge which belongs to the Cluster Network..
+
+
+![](./20240930-vxlan-enhancement/vxlan_vxlan_attched_to_bridge.png)
+![](./20240930-vxlan-enhancement/vxlan_three_nodes_setup.png)
+Harvester Host Cluster should have knowledge of all the existing VMs, VTEP and the created VxLAN interface.
+It should adds the MAC-IP mapping in the table for each VM and convert them into FDB entry/ARP etnry, so the VTEP can skip dynamic MAC learning
```
starbops, line 75: Would you elaborate on this design decision? What are the concerns for VTEPs and dynamic MAC learning? Thank you.

enhancements/20240930-vxlan-enhancement.md
mingshuoqiu: should be `each`.

enhancements/20240930-vxlan-enhancement.md
mingshuoqiu: Yes. It's the IP range (defined from the UI or attached IP pool)

enhancements/20240930-vxlan-enhancement.md
mingshuoqiu: My original thought is that it should be allocated dynamicall by IPAM but hidden from the User.

starbops (commented):
We might also need to support multiple VxLAN networks with the same or overlapping subnet defined. If we won't be able to do this, please note it in the Non-goals section. Thank you.

Issue #6754, opened 2024-10-11
[ENHANCEMENT] Prevent Upgrade if there is VM Backup (VM Backup Schedule) Under Processing


WebberHuang1118:
**Is your enhancement related to a problem? Please describe.**
While the VM Backup is under processing, it may occupy one LH VA ticket. This can cause the following VM operation (e.q. VM Live migration) to be blocked by LH VA webhook  https://github.com/longhorn/longhorn-manager/blob/5ff1f83aefb2c17365264d7cc111e04415c9cef8/webhook/resources/volumeattachment/validator.go#L121-L124

**Describe the solution you'd like**
Prevent upgrade if there is VM Backup (VM Backup Schedule) under Processing

w13915984028:
@WebberHuang1118 Another question:

When an upgrade is ongoing, it may take hours, will scheduled backup still process if the timer arrives?

WebberHuang1118:
Hi @w13915984028 

PR https://github.com/harvester/harvester/pull/6755 will make sure all schedules are suspended before starting an upgrade.

innobead:
> **Is your enhancement related to a problem? Please describe.** While the VM Backup is under processing, it may occupy one LH VA ticket. This can cause the following VM operation (e.q. VM Live migration) to be blocked by LH VA webhook [longhorn/longhorn-manager@`5ff1f83`/webhook/resources/volumeattachment/validator.go#L121-L124](https://github.com/longhorn/longhorn-manager/blob/5ff1f83aefb2c17365264d7cc111e04415c9cef8/webhook/resources/volumeattachment/validator.go#L121-L124)

Will the upgrade eventually happen after the backup is finished? What is the current result when encountering this issue?

WebberHuang1118:
> > **Is your enhancement related to a problem? Please describe.** While the VM Backup is under processing, it may occupy one LH VA ticket. This can cause the following VM operation (e.q. VM Live migration) to be blocked by LH VA webhook [longhorn/longhorn-manager@`5ff1f83`/webhook/resources/volumeattachment/validator.go#L121-L124](https://github.com/longhorn/longhorn-manager/blob/5ff1f83aefb2c17365264d7cc111e04415c9cef8/webhook/resources/volumeattachment/validator.go#L121-L124)
> 
> Will the upgrade eventually happen after the backup is finished? What is the current result when encountering this issue?

According to my discussion with @starbops, the upgrade may have a timeout for VM live migration if the backup spends too much time. cc @starbops

[...]

PR #6755, opened 2024-10-11
enhancement: Prevent Upgrade if VM Backup (Schedule) Under Processing

WebberHuang1118:
**Problem:**
While the VM Backup is under processing, it may occupy one LH VA ticket. This can cause the following VM operation (e.q. VM Live migration) to be blocked by LH VA webhook https://github.com/longhorn/longhorn-manager/blob/5ff1f83aefb2c17365264d7cc111e04415c9cef8/webhook/resources/volumeattachment/validator.go#L121-L124

**Solution:**
Prevent upgrade if there is VM Backup (VM Backup Schedule) under processing

**Related Issue:**
#6754 

**Test plan:**
- Building the Harvester cluster with master-head (`d4c699718434682dda46c201cbbd87e80d23db8d`)
- Setting ManageChart harvester's spec.paused = true
- Updating Harvester Webhook deployment with this PR
- Applying a version manifest upgrade to the same version, e.q. 
  ```
  kind: Version
  metadata:
    name: master
    namespace: harvester-system
  spec:
    isoURL: http://192.188.0.54:8080/iso/harvester-master-amd64.iso
  ```
- Creating a VM `vm1`
- Creating vmbackup for VM `vm1`, during the backup in progress, start the upgrade
- The upgrade CR creation will be rejected with message `vmbackup <vmabckup name> is under processing`
- Creating a backup schedule for VM `vm1`, and making sure the schedule is not suspended
- Starting the upgrade, the upgrade CR creation will be rejected with the message `schedule <scheduke name> is running`

### DIFF
--- a/pkg/webhook/indexeres/indexer.go
+++ b/pkg/webhook/indexeres/indexer.go
@@ -9,6 +9,7 @@ import (
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/controller/master/backup"
 	"github.com/harvester/harvester/pkg/util"
 	indexeresutil "github.com/harvester/harvester/pkg/util/indexeres"
 	"github.com/harvester/harvester/pkg/webhook/clients"
@@ -16,6 +17,7 @@ import (
 
 const (
 	VMBackupBySourceUIDIndex              = "harvesterhci.io/vmbackup-by-source-uid"
+	VMBackupByIsProgressing               = "harvesterhci.io/vmbackup-by-is-progressing"
 	VMRestoreByTargetNamespaceAndName     = "harvesterhci.io/vmrestore-by-target-namespace-and-name"
 	VMRestoreByVMBackupNamespaceAndName   = "harvesterhci.io/vmrestore-by-vmbackup-namespace-and-name"
 	VMBackupSnapshotByPVCNamespaceAndName = "harvesterhci.io/vmbackup-snapshot-by-pvc-namespace-and-name"
@@ -23,6 +25,7 @@ const (
 	ImageByExportSourcePVCIndex           = "harvesterhci.io/image-by-export-source-pvc"
 	ScheduleVMBackupBySourceVM            = "harvesterhci.io/svmbackup-by-source-vm"
 	ScheduleVMBackupByCronGranularity     = "harvesterhci.io/svmbackup-by-cron-granularity"
+	ScheduleVMBackupBySuspended           = "harvesterhci.io/svmbackup-by-suspended"
 	ImageByStorageClass                   = "harvesterhci.io/image-by-storage-class"
 	VMInstanceMigrationByVM               = "harvesterhci.io/vmim-by-vm"
 )
@@ -31,6 +34,7 @@ func RegisterIndexers(clients *clients.Clients) {
 	vmBackupCache := clients.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineBackup().Cache()
 	vmBackupCache.AddIndexer(VMBackupBySourceUIDIndex, vmBackupBySourceUID)
 	vmBackupCache.AddIndexer(VMBackupSnapshotByPVCNamespaceAndName, vmBackupSnapshotByPVCNamespaceAndName)
+	vmBackupCache.AddIndexer(VMBackupByIsProgressing, vmBackupByIsProgressing)
 
 	vmRestoreCache := clients.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineRestore().Cache()
 	vmRestoreCache.AddIndexer(VMRestoreByTargetNamespaceAndName, vmRestoreByTargetNamespaceAndName)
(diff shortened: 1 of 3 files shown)


bk201 (approved):
Good finding. We can improve further by
- An upgrade-config to enable automatically suspend al VM schedules (and restore them when an upgrade is done).
- A global setting to control all VM schedules.

WebberHuang1118:
> Good finding. We can improve further by
> 
> * An upgrade-config to enable automatically suspend al VM schedules (and restore them when an upgrade is done).
> * A global setting to control all VM schedules.

Open a ticket for these requirement https://github.com/harvester/harvester/issues/6759

pkg/webhook/indexeres/indexer.go
```diff
@@ -9,20 +9,23 @@ import (
...
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/controller/master/backup"
 	"github.com/harvester/harvester/pkg/util"
 	indexeresutil "github.com/harvester/harvester/pkg/util/indexeres"
 	"github.com/harvester/harvester/pkg/webhook/clients"
 )
 
 const (
 	VMBackupBySourceUIDIndex              = "harvesterhci.io/vmbackup-by-source-uid"
+	VMBackupByIsProgressing               = "harvesterhci.io/vmbackup-by-IsProgressing"
```
w13915984028: avoid mixed formating

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -224,6 +238,37 @@ func (v *upgradeValidator) checkManagedCharts() error {
...
+	}
+
+	return werror.NewBadRequest(fmt.Sprintf("vmbackup %s/%s is under processing", vmBackups[0].Namespace, vmBackups[0].Name))
+}
+
+// Since volume snapshot/backup will hold one additional LH VA tickets, it can block VM live migration,
+// and an active schedule could start a vmbackup at any time.
+// we should check if there is any running schedule before upgrade
+func (v *upgradeValidator) checkScheduleVMBackups() error {
+	svmbackups, err := v.svmbackupCache.GetByIndex(indexeres.ScheduleVMBackupBySuspended, strconv.FormatBool(false))
```
w13915984028, line 261: Is following case already prevented:

When an upgrade is on-going, the suspended status is changed from false to true.

Following function seems not checking.

```
func (v *scheuldeVMBackupValidator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) error {
```

w13915984028 (commented):
another question, thanks.

pkg/webhook/resources/upgrade/validator.go
WebberHuang1118: This PR is to prevent the unexpected VM backup from being triggered. Once the PR is merged, only the user manually resumes the schedule during the upgrade and can make a schedule back to running as an upgrade is on-the-fly.

However, the Harvester document already mentions `Do not operate the cluster during an upgrade` https://docs.harvesterhci.io/v1.4/upgrade/index/#start-an-upgrade

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -224,6 +238,37 @@ func (v *upgradeValidator) checkManagedCharts() error {
...
+	vmBackups, err := v.vmBackupCache.GetByIndex(indexeres.VMBackupByIsProgressing, strconv.FormatBool(true))
+	if err != nil {
+		return err
+	}
+
+	if len(vmBackups) == 0 {
+		return nil
+	}
+
+	return werror.NewBadRequest(fmt.Sprintf("vmbackup %s/%s is under processing", vmBackups[0].Namespace, vmBackups[0].Name))
```
starbops: Is there any reason we only show the first vmbackup to the user?

pkg/webhook/resources/upgrade/validator.go
```diff
@@ -224,6 +238,37 @@ func (v *upgradeValidator) checkManagedCharts() error {
...
+	svmbackups, err := v.svmbackupCache.GetByIndex(indexeres.ScheduleVMBackupBySuspended, strconv.FormatBool(false))
+	if err != nil {
+		return err
+	}
+
+	if len(svmbackups) == 0 {
+		return nil
+	}
+
+	return werror.NewBadRequest(fmt.Sprintf("schedule %s/%s is running", svmbackups[0].Namespace, svmbackups[0].Name))
```
starbops: ditto.

starbops (commented):
Thank you for making the upgrade more robust. Overall LGTM, just one nit. Please check the comment.

pkg/webhook/resources/upgrade/validator.go
WebberHuang1118: Because if the message contains all the running VMBackups, its length could be very long, and make it verbose.

In my opinion, showing the first VMBackup should provide enough information for the user to stop all running VMBackups. Even if the user doesn't stop all the VMBackup, the next running VMBackup will be popped up by the webhook message

pkg/webhook/resources/upgrade/validator.go
WebberHuang1118: Same reason as https://github.com/harvester/harvester/pull/6755#discussion_r1799168162

pkg/webhook/resources/upgrade/validator.go
starbops: Thank you. It makes sense to me. But I'm thinking maybe we can polish the error message a bit and give the user a hint that there might be more than this one to handle. So the user doesn't see similar errors again and then realizes they need to stop **all** the VMBackups/ScheduleVMBackups.

pkg/webhook/indexeres/indexer.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/webhook/resources/upgrade/validator.go
WebberHuang1118: Hi @starbops,
I updated the webhook message, please take a look, thanks.

starbops (approved):
LGTM, thank you.

bk201:
@w13915984028 I merged this for rc3, please let Webber know if you have further concerns, thanks!

Issue #6711, opened 2024-10-06
[ENHANCEMENT] Cancel previous in progress build when there is new commit in a same PR


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
If there is a new commit and previous in progress build has not finished yet, there will be multiple running workflows. We can cancel previous in progress build to save CI resources.

**Describe the solution you'd like**
Add this to `build.yml`:

```yml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: ${{ !contains(fromJson('["master", "v1.2", "v1.3", "v1.4", "v1.5"]'), github.ref) }}
```

FrankYang0529:
All PRs are merged. Close the issue.

PR #6765, opened 2024-10-14
feat(ci): cancel in progress CI if there is new commit in a same PR

FrankYang0529:

**Problem:**

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/6711

**Test plan:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,9 @@ on:
   pull_request:
   schedule:
     - cron: "10 0,12 * * *"
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' && !startsWith(github.ref, 'refs/heads/v') && !startsWith(github.ref, 'refs/tags/v') }}
 jobs:
   build-binaries:
     name: Build binaries
@@ -150,7 +153,7 @@ jobs:
 
     - name: generate addon manifests
       run: make generate-addons
-      
+
     # rancher/harvester-upgrade image
     - name: docker-publish-harvester-upgrade
       if: ${{ startsWith(github.ref, 'refs/heads/') }}


FrankYang0529:
Cancel in progress CI result: https://github.com/harvester/harvester/actions/runs/11322270751

bk201:
Will the change cancel the running job in another PR?

FrankYang0529:
> Will the change cancel the running job in another PR?

No, it uses PR number as name or `girhub.ref`. For branch CI, we skip canceling in progress CI.

bk201 (approved):

tserong (approved):
Nice!

tserong:
I know I just approved this, but I subsequently realised that we'll need to eventually update `cancel-in-progress: ${{ !contains(fromJson('["master", "v1.2", "v1.3", "v1.4", "v1.5"]'), github.ref) }}` once we get to harvester v1.6 (and v1.7, and ...).

It might be interesting to experiment with something like `github.ref != "master" && !startsWith(gihub.ref, "v")` to exclude master and all v-named branches instead. 

...and it might need to be `github.ref_name` which is the short ref name, vs `github.ref` which will be something like `refs/heads/...` according to https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

FrankYang0529:
Hi @tserong, updated it. Thanks for the suggestion.

tserong (approved):

.github/workflows/build.yml
```diff
@@ -10,6 +10,9 @@ on:
   pull_request:
   schedule:
     - cron: "10 0,12 * * *"
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' && !startsWith(github.ref, 'refs/tags/v') }}
```
bk201: Does this work for v1.x branches?

tserong: Oooh, good point, that means I think it'll need to be something like:

```
${{ github.ref != 'refs/heads/master' && !startsWith(github.ref, 'refs/heads/v') && !startsWith(github.ref, 'refs/tags/v') }}
```

That should cover both branches and tags.

FrankYang0529: Updated it. Thanks.

tserong (approved):

bk201 (approved):

PR #6767, opened 2024-10-14
fix(upgrade): source "upgrade to" version from repoinfo

starbops:

**Problem:**

A user may manually create a version CR with custom naming to trigger an upgrade. For instance:

```
cat > ./version.yaml << EOF
kind: Version
metadata:
  name: local-upgrade
  namespace: harvester-system
spec:
  isoChecksum: 2b99b0870d3363b393bb7900c53b8a1f529e150519558cea672e9ff8eab19ba0dac3f35688e49cdc2ac3ba0d253017fd3d8790cb229856735e30c524f56f22a7
  isoURL: http://192.168.1.1/harvester-amd64.iso
EOF
```

The name of the Version CR is then used to create the corresponding Upgrade CR. It will be in the `.spec.version` field. The upgrade eligibility check currently relies on that field to determine what version it will upgrade to. Apparently, it's not so reliable. We need to fix this.

**Solution:**

The `.spec.version` field in the Upgrade CR is not reliable. Instead of relying on that field to determine the "upgrade to" version, we read from the release version of the `.status.repoInfo` field. Those information is populated from the content in the ISO image.

**Related Issue:**

#5494 

**Test plan:**

1. Prepare a single-node Harvester cluster containing this fix
2. Fake the version to v1.4.0 by adding `HARVESTER_SERVER_VERSION=v1.4.0` environment variable to the `harvester` Deployment
3. Trigger an upgrade to v1.3.2 (simulating a downgrade)
   1. Create a Version CR with a custom name, e.g., `local-upgrade`
   2. Apply the Version CR
   3. Choose the version and click the upgrade button from the dashboard
4. The upgrade should end in the "downgrading is prohibited" error after the upgrade repo VM is up and running.

### DIFF
--- a/pkg/upgradehelper/versionguard/version.go
+++ b/pkg/upgradehelper/versionguard/version.go
@@ -11,11 +11,21 @@ import (
 
 func Check(upgrade *v1beta1.Upgrade, strictMode bool, minUpgradableVersionStr string) error {
 
-	currentVersion, err := version.NewHarvesterVersion(upgrade.Status.PreviousVersion)
+	repoInfo, err := getRepoInfo(upgrade)
+	if err != nil {
+		logrus.WithFields(logrus.Fields{
+			"namespace": upgrade.Namespace,
+			"name":      upgrade.Name,
+		}).Error("failed to retrieve repo info")
+		return err
+	}
+
+	upgradeVersion, err := version.NewHarvesterVersion(repoInfo.Release.Harvester)
 	if err != nil {
 		return err
 	}
-	upgradeVersion, err := version.NewHarvesterVersion(upgrade.Spec.Version)
+
+	currentVersion, err := version.NewHarvesterVersion(upgrade.Status.PreviousVersion)
 	if err != nil {
 		return err
 	}
@@ -27,15 +37,6 @@ func Check(upgrade *v1beta1.Upgrade, strictMode bool, minUpgradableVersionStr st
 			return err
 		}
 	} else {
-		repoInfo, err := getRepoInfo(upgrade)
-		if err != nil {
-			logrus.WithFields(logrus.Fields{
-				"namespace": upgrade.Namespace,
-				"name":      upgrade.Name,
-			}).Error("failed to retrieve repo info")
-			return err
-		}
-
 		minUpgradableVersion, err = version.NewHarvesterVersion(repoInfo.Release.MinUpgradableVersion)
 		// When the error is ErrInvalidVersion, let the nil minUpgradableVersion slip through the check since it's a
 		// valid scenario. It implies "upgrade with no restrictions."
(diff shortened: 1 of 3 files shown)


bk201 (approved):
LGTM! thanks!

pkg/upgradehelper/versionguard/version.go
```diff
@@ -11,11 +11,21 @@ import (
 
 func Check(upgrade *v1beta1.Upgrade, strictMode bool, minUpgradableVersionStr string) error {
 
-	currentVersion, err := version.NewHarvesterVersion(upgrade.Status.PreviousVersion)
+	repoInfo, err := getRepoInfo(upgrade)
```
brandboat, line 14: Not related to this PR, but I'm curious why UpgradeStatus.RepoInfo type is string
https://github.com/harvester/harvester/blob/848f704bbdbd577305b5500d1978bddd784b16b4/pkg/apis/harvesterhci.io/v1beta1/upgrade.go#L54

instead of type `RepoInfo`, is there any concern ?

https://github.com/harvester/harvester/blob/848f704bbdbd577305b5500d1978bddd784b16b4/pkg/controller/master/upgrade/repoinfo/repoinfo.go#L15-L17

bk201: I only have a vague impression. Maybe it's because I was not certain if the structure would change when I implemented it (or not familiar with GO enough).

pkg/upgradehelper/versionguard/version.go
```diff
@@ -27,15 +37,6 @@ func Check(upgrade *v1beta1.Upgrade, strictMode bool, minUpgradableVersionStr st
...
 	} else {
-		repoInfo, err := getRepoInfo(upgrade)
-		if err != nil {
-			logrus.WithFields(logrus.Fields{
-				"namespace": upgrade.Namespace,
-				"name":      upgrade.Name,
-			}).Error("failed to retrieve repo info")
-			return err
-		}
-
```
brandboat, line 38: Just go through the test plan, but I got error message `invalid version string` instead of `downgrading is prohibited`. 

image

brandboat: Didn't deep dive into the code, but looks like the `.Status.RepoInfo` could be empty in
https://github.com/harvester/harvester/blob/master/pkg/controller/master/upgrade/upgrade_controller.go#L302 and lead upgrade failed.

starbops: I also noticed that, looking into it.

starbops: It works now :D
image

pkg/upgradehelper/versionguard/version.go
brandboat: Thanks for the explanation ! I'm not sure if we can convert the type from string to RepoInfo, will try to do that if have time. If that's something we can improve, I'll create an issue later.

brandboat (approved):
LGTM 🎉

brandboat:
@mergify backport v1.4

Issue #6766, opened 2024-10-14
[BUG] VMBacup is Never Updated with the condition MetadataReady


WebberHuang1118:
**Describe the bug**
VMBackup CR is never updated with the condition `MetadataReady`, it should change to `True` after the Harvester controller uploads backup metadata https://github.com/harvester/harvester/blob/b3aee86ae14eeca73aed9647de244410f5f144f0/pkg/controller/master/backup/backup.go#L863-L868

**To Reproduce**
1. Setting a backup target
2. Creating a VM and having a VMBackup
3. VMBackup CR's `MetadataReady` condition never appears

**Expected behavior**
VMBackup CR's `MetadataReady` condition should display `True` once the VMBackup progress completes

**Environment**
 - Harvester ISO version: master-head branch (`b3aee86ae14eeca73aed9647de244410f5f144f0`)

**Additional context**
This condition is introduced by https://github.com/harvester/harvester/pull/4614 cc @FrankYang0529

FrankYang0529:
Thanks for creating the issue. I will check v1.4.0-rc2.

w13915984028:
Test report: **PASS**

(1) Set up a 2-node harvester cluster from v1.4.0-rc2 ISO;
(2) Set a NFS server as backup target
```
{
  "type": "nfs",
  "endpoint": "nfs://192.168.122.159:/nfs_server/nfs_share/"
}
```
(3) Take a backup of VM, the vmbackup CRD object has the `MetadataReady` condition updated as `True`.

```
$kubectl get vmbackup
NAME    SOURCE_KIND      SOURCE_NAME   TYPE     READY_TO_USE   AGE     ERROR
vm3-1   VirtualMachine   vm3           backup   true           3m23s   

...
    - lastTransitionTime: "2024-10-22T13:12:08Z"
      status: "True"
      type: MetadataReady

```

<details>
<summary>VMBackup object</summary>
<br>

```
$kubectl get vmbackup -oyaml
apiVersion: v1
items:
- apiVersion: harvesterhci.io/v1beta1
  kind: VirtualMachineBackup
  metadata:
    annotations:
      harvesterhci.io/snapshotFreezeFS: "false"
    creationTimestamp: "2024-10-22T13:11:59Z"
    finalizers:
    - wrangler.cattle.io/harvester-vm-backup-controller
    - wrangler.cattle.io/vm-backup-controller
    generation: 9
    name: vm3-1
    namespace: default
    resourceVersion: "175849"
[...]

PR #6773, opened 2024-10-14
fix(backup): update backup status if there is condition change

FrankYang0529:

**Problem:**
We didn't update VMBackup when we set a `MetadataReady` condition.

**Solution:**
Updated it if VMBackup status is different from previous one.

**Related Issue:**
https://github.com/harvester/harvester/issues/6766

**Test plan:**
1. Setting a backup target.
2. Creating a VM and having a VMBackup.
3. Check VMBackup CR's MetadataReady condition is true.


WebberHuang1118 (approved):
LGTM, thanks for the PR.

bk201 (approved):

Issue #6772, opened 2024-10-14
[BUG] All VMBackup couldn't be recovered if Remote Backup has any VMImage with the same DisplayName


WebberHuang1118:
**Describe the bug**
All VMBackup couldn't be recovered if Remote Backup has any VMImage with the same DisplayName

**To Reproduce**
1. Setting a remote backup target
2. Uploading an VMImage from the local file
3. Creating VM `vm1`, and having a related VMBackup `vm1-b`
4. Deleting VM `vm1`, VMBackup `vm1-b` and the VMImage
5. Uploading another VMImage from the same local file
6. Creating VM `vm2`, and having a related VMBackup `vm2-b`
7. Setting the remote backup target to default
8. Deleting VMBackup `vm2-b`
9. Setting the remote backup target back
10. Can't discover VMBackup `vm2-b`
      - Harvester controller with error message `can't sync vm image metadata" error="admission webhook \"validator.harvesterhci.io  \" denied the request: A resource with the same name exists" target.endpoint="http://192.188.0.54:9000" target.type=s3` 
      - This is brought by the legacy VMImage metadata. The legacy VMImage has the same displayName as an existing VMImage, although it has already been deleted. However, its metadata is still left on the backup target.
      - The above situation caused the conflict in `syncVMImage()` https://github.com/harvester/harvester/blob/b3aee86ae14eeca73aed9647de244410f5f144f0/pkg/controller/master/backup/backup_metadata.go#L124-L128, and make function `syncVMBackup()` be skipped

**Expected behavior**
VMBackup `vm2-b` could be recovered after setting back the remote backup target

**Environment**
 - Harvester ISO version: master-head branch (`b3aee86ae14eeca73aed9647de244410f5f144f0`)

**Additional context**
- This is introduced by https://github.com/harvester/harvester/pull/4614 cc @FrankYang0529
- The Harvester controller uploads VMImage metadata as LH creates backupbackingimage CR https://github.com/harvester/harvester/blob/b3aee86ae14eeca73aed9647de244410f5f144f0/pkg/controller/master/backup/backup_backing_image.go#L131-L146. However, when VMImage is deleted, the VMImage metadata is never touched on the backup target. This brings an inconsistency between the cluster and remote backup target, maybe we should consider the design for the backup image. WDYT? @FrankYang0529 
- Sync VMBackup to another cluster can have a similar situation

brandboat:
Test OK on v1.4.0-rc3, will close the issue

## Environment
[...]

PR #6780, opened 2024-10-15
fix(backup): skip vmimage restore if there is same display name

FrankYang0529:

**Problem:**
If there is same display name VM image in the cluster, the restoring metadata from backup target fails. The reason is that VM image can't be created and the whole process is stuck.

**Solution:**
Skip restoring VM image if there is another VM image which has same display name in the cluster.

**Related Issue:**
https://github.com/harvester/harvester/issues/6772

**Test plan:**
Case 1: Removing VM image can cleanup VM image metadata in backup target

1. Create a new cluster (this PR updates CRD, so please create a new cluster from this PR)
2. Setup backup target.
3. Create a VM image and VM.
4. Take a backup for the VM.
5. Check there is VM image metadata in the backup target.
6. Remove the backup, VM, and VM image.
7. Check the VM image metadata is removed in the backup target.

Case 2: Skip restoring VM image with same display name

1. Create a new cluster (this PR updates CRD, so please create a new cluster from this PR)
2. Setup backup target.
3. Create a VM image and VM.
4. Take a backup for the VM.
5. Set backup target to default value.
6. Remove the backup, VM, and VM image.
7. Create a new VM image with same display name and different CR name.
8. Set backup target back.
9. Check harvester pod has following log.
```
skip create vm image, because there is already an image with the same display name

skip creating vm backup, because the storage class is not found
```


FrankYang0529:
I updated the document PR about limitations of automatic sync vm images.

https://github.com/harvester/docs/pull/648

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -397,3 +399,45 @@ func handleFail(image *harvesterv1.VirtualMachineImage, cond condition.Cond, err
...
+	if exist := bsDriver.FileExists(destURL); exist {
+		logrus.WithFields(logrus.Fields{
+			"namespace":    image.Namespace,
+			"name":         image.Name,
+			"backupTarget": image.Status.BackupTarget,
+		}).Debugf("delete vm image metadata in backup target")
+		return bsDriver.Remove(destURL)
+	}
+	return nil
+}
```
WebberHuang1118, line 443: I'm not sure about this: if we delete the VMImage metadata on the remote target, will this behavior contradict the idea of backing up an image?

WebberHuang1118 (commented):
Thanks for the PR, I think some points may need further discussion.

WebberHuang1118:
https://github.com/FrankYang0529/harvester/blob/197f66ea10f8d931184b056cea986d81a3a551ba/pkg/controller/master/backup/backup_metadata.go#L308-L319

Maybe we shouldn't sync the VMBackup if the related VMImage can't be provided from https://github.com/FrankYang0529/harvester/blob/197f66ea10f8d931184b056cea986d81a3a551ba/pkg/controller/master/backup/backup_metadata.go#L183-L190. WDYT?

pkg/controller/master/image/vm_image_controller.go
FrankYang0529: Hi @WebberHuang1118, thanks for the review. I think removing VMImage metadata is a safe operation with PR https://github.com/harvester/harvester/pull/6786, because VMImage can't be removed if there is any VMBackup using it. That means if VMImage can be removed, there is no related VMBackup in the backup target.

FrankYang0529:
> https://github.com/FrankYang0529/harvester/blob/197f66ea10f8d931184b056cea986d81a3a551ba/pkg/controller/master/backup/backup_metadata.go#L308-L319
> 
> Maybe we shouldn't sync the VMBackup if the related VMImage can't be provided from https://github.com/FrankYang0529/harvester/blob/197f66ea10f8d931184b056cea986d81a3a551ba/pkg/controller/master/backup/backup_metadata.go#L183-L190. WDYT?

Thanks for the suggestion. I skip to restore vmbackup if related storage class is not in the cluster.

WebberHuang1118 (approved):
LGTM, thanks for the PR. 

Here are just some suggestions for logs:
https://github.com/FrankYang0529/harvester/blob/ca20e61dcf6df66afb72a789c0c62ea58fcb80d7/pkg/controller/master/backup/backup_metadata.go#L201-L216

https://github.com/FrankYang0529/harvester/blob/ca20e61dcf6df66afb72a789c0c62ea58fcb80d7/pkg/controller/master/backup/backup_metadata.go#L288-L307

How about adding logs if the Harvester creates VMImages or VMBackups from remote? This is only triggered when setting the backup target.

FrankYang0529:
> How about adding logs if the Harvester creates VMImages or VMBackups from remote? This is only triggered when setting the backup target.

Yes, updated it. Thank you.

bk201 (approved):

Issue #6782, opened 2024-10-15
[BUG] Prevent removing VM image if there is VM backup


FrankYang0529:
**Describe the bug**
We didn't prevent users from removing VM image when there is a VM backup using it. If users remove the VM image, the related VM backup can't work.

**To Reproduce**
1. Create a VM backup.
2. Remove VM and VM image.
3. Restore VM backup and it can't work.

**Expected behavior**
VM image can't be removed.

**Environment**
 - Harvester ISO version: master-head

w13915984028:
Test report: **PASS**

(1) Set up a 2-node cluster with v1.4.0-rc3 ISO
(2) Create a VM
(3) Setup backup target to NFS server
(4) Backup a VM
(5) Stop & Delete the VM
(6) Delete the image, it is blocked as expectation.
![image](https://github.com/user-attachments/assets/624fd5df-9428-4c38-b6ea-ce60f4c6941d)

check the NFS:
```
sudo cat harvester/vmimages/default/image-rr4hq.cfg
...
{"name":"image-rr4hq","namespace":"default","url":"nfs://192.168.122.159:/nfs_server/nfs_share/?backingImage=vmi-a22383cd-4515-4592-8c60-db4d014006f4","displayName":"cirros-0.5.2-x86_64-disk.img","checksum":"6b813aa46bb90b4da216a4d19376593fa3f4fc7e617f03a92b7fe11e9a3981cbe8f0959dbebe36225e5f53dc4492341a4863cac4ed1ee0909f3fc78ef9c3e869","storageClassParameters":{"migratable":"true","numberOfReplicas":"3","staleReplicaTimeout":"30"}}
```

(7) Related CRDs objects, the VMBackup refers to VMImage:
```
kk get vmimage
NAME          DISPLAY-NAME                   SIZE       VIRTUALSIZE   AGE
image-rr4hq   cirros-0.5.2-x86_64-disk.img   16300544   117440512     160m
```

```
...
- apiVersion: harvesterhci.io/v1beta1
  kind: VirtualMachineBackup
...
    volumeBackups:
    - csiDriverName: driver.longhorn.io
      longhornBackupName: backup-570a51cab7b44d98
      name: vm3-1-volume-vm3-disk-0-qfmxe
      persistentVolumeClaim:
        metadata:
          annotations:
            harvesterhci.io/imageId: default/image-rr4hq   // refers to above vmimage
            pv.kubernetes.io/bind-completed: "yes"
[...]

PR #6786, opened 2024-10-15
fix(vmimage): prevent removing vmimage if any vmbackup uses it

FrankYang0529:

**Problem:**
We didn't prevent users from removing VM image when there is a VM backup using it. If users remove the VM image, the related VM backup can't work.

**Solution:**
Prevent users from removing VM image if any vmbackup uses it.

**Related Issue:**
https://github.com/harvester/harvester/issues/6782

**Test plan:**
1. Setup backup target.
2. Create a VM image, VM, and VM backup.
3. Remove VM.
4. Remove VM image. Webhook should prevent this behavior.

Screenshot 2024-10-15 at 5 29 41 PM


FrankYang0529:
CodeFactor complex method is not related to this PR.

WebberHuang1118 (approved):
LGTM, thanks for the PR.

bk201 (approved):

Issue #6509, opened 2024-09-05
[BUG] improve identification of VM's pinned to hostnames during maintenance operations


ibrokethecloud:
**Describe the bug**

As part of the requirements for issue: https://github.com/harvester/harvester/issues/4888 we introduced additional logic to identify VM's which cannot be scheduled to other nodes due to
* HostDevices
* Scheduling or Zone constraints
* Unhealthy longhorn replicas

The change did not include identifying VM's where a specific node scheduling has been specified for the VM.  In such scenarios hostname is added using `kubernetes.io/hostname` label to VM node selector

**To Reproduce**
1. Create a VM in harvester which is scheduled to run on a specific host
2. Attempt to place the host into maintenance mode.

**Expected behavior**

Expect to see the VM reported in the list of VM's which cannot be migrated since the `hostname` based scheduling criteria cannot be met on another node in the cluster
**Support bundle**

**Environment**
 - Harvester ISO version: v1.3-head, v1.4-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

Additional info is available here: https://github.com/harvester/harvester/issues/5911#issuecomment-2302882601

lanfon72:
Verified this bug is not fix yet, move back to implement.


https://github.com/user-attachments/assets/a2711ad0-0d5b-423a-81d6-009f916688cc




Test Information
----
* Environment: **bare metal DL360G9 3 nodes**
* Harvester Version: **v1.4.0-rc3**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with at least 2 nodes having PCI devices
2. Create an image for VM creation
3. Create VM that scheduling on specific node (using `spec.spec.nodeSelector`) and start it
4. Enable maintenance mode without **force** on the node hosting VM
6. Error message displayed and VMs should NOT be shut down automatically
7. The node should NOT enter maintenance mode
8. Enable maintenance mode with **force** on the node hosting VMs
9. VMs should be shut down automatically
10. The node should enter maintenance mode

w13915984028:
@lanfon72 Your failure test case https://github.com/harvester/harvester/issues/6509#issuecomment-2427066405 looks to be possibly related to `KubeVirt` change. Please also check the yaml of your VM, if it is similar to following.

--

Harvester code has following assumption:

[...]

PR #6836, opened 2024-10-21
[WIP] Update the livemigration status check

w13915984028:

**Problem:**

The livemigratable status condition check needs to be amended.

**Solution:**

Update the condition check

**Special Node:**

`KubeVirt` seems also not function well:

The `migrationMethod: BlockMigration` marked VM can still start a migration, as long as the target node has the related label, and the migration is successful.

If the target node has no such label, the migration will fail.

**Related Issue:**

**Test plan:**

Per https://github.com/harvester/harvester/issues/6509


w13915984028:
@mergify backport v1.4

w13915984028:
@ibrokethecloud @bk201 According to the VM yaml output, I guess we also need this fix to improve the live migration status check on VMI object. The PR https://github.com/harvester/harvester/pull/6839 does not touch this condition. thanks.

w13915984028:
A customized nodeSelector builds the test case easily:

https://github.com/harvester/harvester/issues/6843#issuecomment-2428931928

w13915984028:
Not necessary as fixed.

PR #6839, opened 2024-10-22
minor tweak to how node scheduling requirements are calculated

ibrokethecloud:

**Problem:**

Original PR to identify VM's with hostname specification and mark them unschedulable https://github.com/harvester/harvester/pull/6397 seems to work only when VM's have a VM network in use.

This happens because when a VM network is added,  harvester injects additional node affinity requirements to ensure VM gets scheduled to nodes which support the VM network. This check subsequently also checks if additional nodeSelector is defined. 

When a VM is pinned to a host, a nodeSelector is added by Harvester to allow hostname matching to target node.

This check fails when a masquerade network is used as the check for node affinity requirements is skipped since no network specific affinity rules are added. As a result of this the subsequent nodeSelector based check is also skipped.

**Solution:**

The PR refactors the code to ensure nodeSelector based scheduling checks are run even no affinity requirements are specified for VM.

**Related Issue:**
**Test plan:**

### DIFF
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -441,25 +441,35 @@ func (ndc *ControllerHandler) CheckVMISchedulingRequirements(originalNode *corev
 	if err != nil {
 		return nil, fmt.Errorf("error listing nodes from nodeCache: %v", err)
 	}
+	var validNodes []*corev1.Node
+	for i, v := range nodeList {
+		if v.Name == originalNode.Name {
+			validNodes = append(nodeList[:i], nodeList[i+1:]...)
+			break
+		}
+	}
 	for _, vmi := range vmiList {
-		var possibleNodes, validNodes []*corev1.Node
+		var possibleNodes, matchingNodes []*corev1.Node
 		if vmi.Spec.Affinity != nil && vmi.Spec.Affinity.NodeAffinity != nil && vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
 			nodeAffinitySelector, err := nodeaffinity.NewNodeSelector(vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution)
 			if err != nil {
 				return nil, fmt.Errorf("error generating nodeAffinitySelector from node scheduling requirements: %v", err)
 			}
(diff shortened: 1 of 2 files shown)


pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -441,25 +441,35 @@ func (ndc *ControllerHandler) CheckVMISchedulingRequirements(originalNode *corev
 	if err != nil {
 		return nil, fmt.Errorf("error listing nodes from nodeCache: %v", err)
 	}
+	var validNodes []*corev1.Node
```
w13915984028, line 444: Yesterday when looking the previous PR, I was wondering why the `originalNode` is not skipped in certain filter.

Now, let's check `isNodeReady` in this block, filter any none-ready nodes, and the following `filterNodesForNodeSelector` implicitly needs this condition, but it was not checked before.

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -441,25 +441,35 @@ func (ndc *ControllerHandler) CheckVMISchedulingRequirements(originalNode *corev
...
 		if vmi.Spec.Affinity != nil && vmi.Spec.Affinity.NodeAffinity != nil && vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
 			nodeAffinitySelector, err := nodeaffinity.NewNodeSelector(vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution)
 			if err != nil {
 				return nil, fmt.Errorf("error generating nodeAffinitySelector from node scheduling requirements: %v", err)
 			}
 			// identify if nodeAffinity can be met by other nodes and node is ready
-			for _, v := range nodeList {
-				if nodeAffinitySelector.Match(v) && v.Name != originalNode.Name && isNodeReady(v) {
+			for _, v := range validNodes {
+				if nodeAffinitySelector.Match(v) && isNodeReady(v) {
```
w13915984028: `isNodeReady(v)` can be skipped via above change

w13915984028 (commented):
Few suggestions, thanks.

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

ibrokethecloud:
@mergifyio backport v1.3

Issue #6851, opened 2024-10-23
[BUG] Upgrade stuck on waiting harvester bundle (v1.3.2 -> v1.4 master build)


bk201:
**Describe the bug**

I see the upgrade is stuck on apply-manifest job when upgrading from v1.3.2 to master build:

```
enabled: true
managedchart.management.cattle.io/harvester configured
managedchart.management.cattle.io/harvester patched
managedchart.management.cattle.io/harvester-crd patched
Waiting for ManagedChart fleet-local/harvester from generation 2
Target version: 0.0.0-test-pr-6826-e086070e, Target state: ready
Current version: 0.0.0-test-pr-6826-e086070e, Current state: null, Current generation: 2
Sleep for 5 seconds to retry
Current version: 0.0.0-test-pr-6826-e086070e, Current state: WaitApplied, Current generation: 4
Sleep for 5 seconds to retry
Current version: 0.0.0-test-pr-6826-e086070e, Current state: WaitApplied, Current generation: 4
Sleep for 5 seconds to retry
...

Sleep for 5 seconds to retry
Current version: 0.0.0-test-pr-6826-e086070e, Current state: ErrApplied, Current generation: 4
Sleep for 5 seconds to retry
Current version: 0.0.0-test-pr-6826-e086070e, Current state: ErrApplied, Current generation: 4

```

The bundle status is:
```
  summary:
    desiredReady: 1
    errApplied: 1
    nonReadyResources:
    - bundleState: ErrApplied
      message: 'encountered 2 deletion errors. First is: admission webhook "validator.harvesterhci.io"
        denied the request: Internal error occurred: no route match found for DELETE
        /v1, Kind=Secret harvester-system/sh.helm.release.v1.harvester.v1'
      name: fleet-local/local
    ready: 0
  unavailable: 1
```

**To Reproduce**
1. Create a v1.3.2 cluster.
2. Upgrade to a maser build (I believe v1.4.0-rc3 is similar).
3. There is a chance to hit this issue.

**Expected behavior**

The upgrade should be smooth.

**Support bundle**

[supportbundle_19aaa896-fd99-4bec-8fe3-1b6ec00b690c_2024-10-22T15-49-19Z.zip](https://github.com/user-attachments/files/17484024/supportbundle_19aaa896-fd99-4bec-8fe3-1b6ec00b690c_2024-10-22T15-49-19Z.zip)

**Environment**
 - Harvester ISO version: d0d234b5d8f434301111acef85370cc41972e202
[...]

Issue #6664, opened 2024-09-26
[BUG] Upgrade stuck due to the unremove longhorn service


Vicente-Cheng:
**Describe the bug**
We are testing the upgrade from v1.3.2 to v1.4 (master branch with rancher upgrade).
Found that the upgrade is stuck with the below bundle error:
```
fleet-local   mcc-harvester                                 0/1                       Modified(1) [Cluster fleet-local/local]; service.v1 longhorn-system/longhorn-engine-manager extra; service.v1 longhorn-system/longhorn-replica-manager extra
fleet-local   mcc-harvester-crd                             1/1
fleet-local   mcc-hvst-upgrade-rzcdn-upgradelog-operator    0/1                       ErrApplied(1) [Cluster fleet-local/local: execution error at (rancher-logging/templates/validate-install-crd.yaml:24:7): Required CRDs are missing. Please install the corresponding CRD chart before installing this chart.]
```

After @FrankYang0529 investigated, we found the issue was caused by unremoved services `longhorn-engine-manager` and `longhorn-replica-manager`.
Longhorn has the corresponding mechanism to delete it. (REF: https://github.com/longhorn/longhorn-manager/blob/8bc7575fc3be2b72c72ba4899f465762a57c55ea/upgrade/v16xto170/upgrade.go#L238-L251)

But somehow, these services have not been deleted. I checked the longhorn-manager log and did not find any error like https://github.com/longhorn/longhorn-manager/blob/8bc7575fc3be2b72c72ba4899f465762a57c55ea/upgrade/v16xto170/upgrade.go#L240 

**To Reproduce**
1. create three nodes 1.3.2 cluster
2. create the downstream cluster
3. upgrade to master-head (master-head already included the rancher upgrade)

**Expected behavior**
upgrade should be smooth

**Support bundle**
None

**Environment**
 - Harvester ISO version: 1.3.2
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM

**Additional context**
We will try to reproduce this situation. Seems it does not always reproduce.
Also, the workaround is manually to delete these two services, `longhorn-engine-manager` and `longhorn-replica-manager`

PhanLe1010:
It is a bit weird that why the service is still in Harvester. I check the the helm chart in master-head Harvester and see that the services are already removed https://github.com/harvester/harvester/blob/master/deploy/charts/harvester/dependency_charts/longhorn/templates/services.yaml

[...]

PR #6857, opened 2024-10-23
fix(upgrade): remove secrets validation and unused longhorn services

FrankYang0529:

**Problem:**

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/6664

**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -88,14 +88,14 @@ wait_harvester_managed_chart() {
         echo "$updated_webhook_config" | kubectl apply -f -
       fi
 
-      if kubectl get service longhorn-engine-service -n longhorn-system > /dev/null 2>&1; then
-        echo "Remove longhorn-engine-service"
-        kubectl delete service longhorn-engine-service -n longhorn-system
+      if kubectl get service longhorn-engine-manager -n longhorn-system > /dev/null 2>&1; then
+        echo "Remove longhorn-engine-manager"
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_manifests.sh
```diff
@@ -44,6 +44,66 @@ wait_managed_chart() {
...
+      fi
+
+      if kubectl get service longhorn-engine-service -n longhorn-system > /dev/null 2>&1; then
+        echo "Remove longhorn-engine-service"
+        kubectl delete service longhorn-engine-service -n longhorn-system
+      fi
+
+      if kubectl get service longhorn-replica-service -n longhorn-system > /dev/null 2>&1; then
+        echo "Remove longhorn-replica-service"
+        kubectl delete service longhorn-replica-service -n longhorn-system
```
bk201:
```suggestion
      if kubectl get service longhorn-engine-manager -n longhorn-system > /dev/null 2>&1; then
        echo "Remove longhorn-engine-manager"
        kubectl delete service longhorn-engine-manager -n longhorn-system
      fi

      if kubectl get service longhorn-replica-manager -n longhorn-system > /dev/null 2>&1; then
        echo "Remove longhorn-replica-manager"
        kubectl delete service longhorn-replica-manager -n longhorn-system
```

FrankYang0529:
Close the PR. Another solution https://github.com/harvester/harvester/pull/6917 tries to solve root cause.

Issue #6646, opened 2024-09-25
[BUG] v140 upgrade encouter: Mounting partition COS_STATE  exit status 32


w13915984028:
**Describe the bug**

Upgrade from harvester-v1.4.0-dev-20240918 to master-head-20240920 failed with `Mounting partition COS_STATE...  exit status 32`

1. Upgrade to harvester-v1.4.0-dev-20240918-amd64.iso, it is successful
2. Upgrade to master-head-20240920, encounter the issue; the upgrade retried and failed finally.

```
INFO[2024-09-25T13:06:44Z] Upgrade called                               
DEBU[2024-09-25T13:06:44Z] Mounting partition COS_STATE                 
E0925 13:06:44.698763   22803 mount_linux.go:195] Mount failed: exit status 32
Mounting command: systemd-run
Mounting arguments: --description=Kubernetes transient mount for /run/cos/state --scope -- mount -t auto -o rw /dev/sda4 /run/cos/state
```

![image](https://github.com/user-attachments/assets/6ca1b18c-be7d-4b66-a328-84a052fb8ce6)

```
$kubectl get pods -A | grep upgrade -i

cattle-system                     system-upgrade-controller-6f86d6d4df-4rzsx               1/1     Running     2 (66m ago)   2d1h
harvester-system                  hvst-upgrade-9pxhh-apply-manifests-rxbsf                 0/1     Completed   0             39m
harvester-system                  hvst-upgrade-9pxhh-single-node-upgrade-harv21-2lmwp      0/1     Error       0             38m
harvester-system                  hvst-upgrade-9pxhh-single-node-upgrade-harv21-4kp5q      0/1     Error       0             20m
harvester-system                  hvst-upgrade-9pxhh-single-node-upgrade-harv21-fvfh5      0/1     Error       0             13m
harvester-system                  hvst-upgrade-9pxhh-single-node-upgrade-harv21-jrlws      0/1     Error       0             29m
harvester-system                  hvst-upgrade-9pxhh-single-node-upgrade-harv21-k6dnn      0/1     Error       0             32m
harvester-system                  hvst-upgrade-9pxhh-single-node-upgrade-harv21-s56lz      0/1     Error       0             35m
harvester-system                  hvst-upgrade-9pxhh-single-node-upgrade-harv21-zwbgq      0/1     Error       0             25m
```

```
error log from hvst-upgrade-9pxhh-single-node-upgrade-harv21* pod:
...
++ date +%Y%m%d%H%M%S
[...]

PR #6865, opened 2024-10-24
disable multipathd as default behaviour in upgrade to v1.4.0

ibrokethecloud:

**Problem:**

v1.4.0 builds enable multipathd by default in the iso. This is needed to help install to multipath devices where needed and users can subsequently enable multipathd via harvester.config

The side effect of this change is during upgrade when the v1.4 based image is replaced on the node multipath is enabled by default and kernel boots off the /dev/mapper devices even though multipathd gets disabled later in the boot sequence.

**Solution:**

The PR adds an extra check in the upgrade path to pass an additional kernel argument via /oem/grubenv file to add `multipath=off` to the kernel parameters.

This results in multipathd being disabled post boot and no /dev/mapper devices being created when they are not needed.

**Related Issue:**
https://github.com/harvester/harvester/issues/6646

**Test plan:**

* Install a harvester v1.3.2 cluster
* Upgrade cluster to v1.4-head or v1.4.0-rc5
* Post upgrade run `df -h` on the nodes and there should be no `/dev/mapper` based devices used for mount
* `dmesg |grep -i multipath` should show `multipath=off` in the kernel arguments

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -627,6 +627,18 @@ EOF
   GRUBENV_FILE="/oem/grubcustom"
   chroot $HOST_DIR /bin/bash -c "if ! [ -f ${GRUBENV_FILE} ]; then touch ${GRUBENV_FILE}; fi" 
 
+  multiPathEnabled=$(yq '.os.externalStorageConfig.enabled // false' ${HOST_DIR}/oem/harvester.config)
+  if [ ${multiPathEnabled} == false ]
+  then
+    thirdPartyArgs=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
+    if [[ ${thirdPartyArgs} != *"multipath=off"* ]]
+    then
+      thirdPartyArgs="${thirdPartyArgs} multipath=off"
+      thirdPartyArgs=$(echo ${thirdPartyArgs} | xargs)
+      chroot $HOST_DIR grub2-editenv /oem/grubenv set third_party_kernel_args="${thirdPartyArgs}"
+    fi
+  fi
+
   umount $tmp_rootfs_mount
   rm -rf $tmp_rootfs_squashfs
 


Vicente-Cheng (approved):
LGTM, thanks for the quick fix

package/upgrade/upgrade_node.sh
```diff
@@ -627,6 +627,15 @@ EOF
   GRUBENV_FILE="/oem/grubcustom"
   chroot $HOST_DIR /bin/bash -c "if ! [ -f ${GRUBENV_FILE} ]; then touch ${GRUBENV_FILE}; fi" 
 
+  multiPathEnabled=$(yq '.os.externalStorageConfig.enabled // false' ${HOST_DIR}/oem/harvester.config)
+  if [ ${multiPathEnabled} == false ]
+  then
+    thirdPartyArgs=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
+    thirdPartyArgs="${thirdPartyArgs} multipath=off"
```
w13915984028: check `multipath=off` is already there to avoid mutli appending it?

ibrokethecloud: new commit fixes this.

Vicente-Cheng (approved):

package/upgrade/upgrade_node.sh
```diff
@@ -627,6 +627,18 @@ EOF
...
 
+  multiPathEnabled=$(yq '.os.externalStorageConfig.enabled // false' ${HOST_DIR}/oem/harvester.config)
+  if [ ${multiPathEnabled} == false ]
+  then
+    thirdPartyArgs=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
+    if [[ ${thirdPartyArgs} != *"multipath=off"* ]]
+    then
+      thirdPartyArgs="${thirdPartyArgs} multipath=off"
+    fi
+    thirdPartyArgs=$(echo ${thirdPartyArgs} | xargs)
```
w13915984028: should we move L638, L639 to if/fi as well?

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

w13915984028:
@mergify backport v1.4

Issue #6895, opened 2024-10-27
[BUG] An Off VM in v1.3.2 with backup can not be restore replace after upgrade to v1.4.0-rc4


albinsun:
**Describe the bug**
![image](https://github.com/user-attachments/assets/c8206d93-eade-465b-9aaf-26d0b925030d)

**To Reproduce**
1. Setup single node Harvester `v1.3.2`
2. Create 2 VMs _vm-running_ and _vm-stopped_
3. Stop _vm-stopped_
4. Take VM Backup on both VM
   ![image](https://github.com/user-attachments/assets/6904cf91-447a-46f9-9872-c7b3afb5f30a)
5. Upgrade Harvester to `v1.4.0-rc4`
6. For now all VMs are `Off`, manually start _vm-running_
7. Restore backup _vm-stopped_
   * Restore Replace: VM stucking in `Restoring` and VirtualMachine status is still `Stopped`
   * Restore New: Successfully restored
   ![image](https://github.com/user-attachments/assets/b09c8c0c-b98a-43f7-b34e-ce68f24aa1be)

**Expected behavior**
Should be able to restore a VM no matter what its status when trigging upgrade 

**Support bundle**
[supportbundle_StuckRestoringAnOffVMBeforeUpgrade.zip](https://github.com/user-attachments/files/17534796/supportbundle_StuckRestoringAnOffVMBeforeUpgrade.zip)

**Environment**
 - Harvester ISO version: `v1.3.2` => `v1.4.0-rc4` single node
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): single node QEMU/KVM

**Additional context**
1. Can reproduce on both local QEMU/KVM and Bare-metal HPE ProLiant DL360
2. The original `Running` VM does not hit this problem

FrankYang0529:
The root cause is that when we try to start vm, its `runStrategy` is `RerunOnFailure`, but related VMI is not existent. From the source code, we do nothing in this case.

 We should fix it by sending a start request if VMI is not found.

https://github.com/harvester/harvester/blob/19bc22db86ec0c9d796ec1d9116c3ddce908082a/pkg/controller/master/backup/restore.go#L914-L938

khushboo-rancher:
Moved it back to review as the PR is not merged yet.

FrankYang0529:
> Moved it back to review as the PR is not merged yet.

Thanks for updating it. I intended to move to review, but not sure why it's in ready for testing 😅.

[...]

PR #6896, opened 2024-10-28
fix(restore): send start request if related vmi is not found

FrankYang0529:

**Problem:**
When volume resources is ready for a restoring VM, we try to send a start request. However, if `runStrategy` is `RerunOnFailure` and related VMI is not existent, we do nothing.

https://github.com/harvester/harvester/blob/19bc22db86ec0c9d796ec1d9116c3ddce908082a/pkg/controller/master/backup/restore.go#L914-L938

**Solution:**
We should send a start request if VMI is not found.

**Related Issue:**
https://github.com/harvester/harvester/issues/6895

**Test plan:**
1. Create a new cluster with this PR.
2. Create a VM with `RerunOnFailure`.
3. Create a backup target.
4. Create a backup.
5. SSH into the VM and use command to shutdown it. (Don't click `stop` button on the UI, because this behavior change the `runStrategy`.)
6. Make sure the stopped VM `runStrategy` is `RerunOnFailure`.
7. Use the backup to restore replace the VM.
8. Check the VM can be restarted.

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -925,7 +925,7 @@ func (h *RestoreHandler) startVM(vm *kubevirtv1.VirtualMachine) error {
 		vmi, err := h.vmiCache.Get(vm.Namespace, vm.Name)
 		if err != nil {
 			if apierrors.IsNotFound(err) {
-				return nil
+				return h.restClient.Put().Namespace(vm.Namespace).Resource("virtualmachines").SubResource("start").Name(vm.Name).Do(h.context).Error()
 			}
 			return err
 		}


bk201 (approved):

WebberHuang1118:
Hi @FrankYang0529 
https://github.com/harvester/harvester/blob/625f14e0c4b568272763f9fd74b87577edef7809/pkg/controller/master/backup/restore.go#L551-L555
We should already set `runStrategy` as `Halted`, why this PR is still required? Thanks

FrankYang0529:
> We should already set `runStrategy` as `Halted`, why this PR is still required?

This PR is to avoid edge case when the cluster is upgraded from v1.3.2 with off VM. Please take a look for reproduce steps in original issue: https://github.com/harvester/harvester/issues/6895

WebberHuang1118:
> > We should already set `runStrategy` as `Halted`, why this PR is still required?
> 
> This PR is to avoid edge case when the cluster is upgraded from v1.3.2 with off VM. Please take a look for reproduce steps in original issue: #6895

IIUC, the Harvester controller should always set restoring VM's `runStrategy` as `Halted`, then the VM will be started by https://github.com/harvester/harvester/blob/625f14e0c4b568272763f9fd74b87577edef7809/pkg/controller/master/backup/restore.go#L947-L948

FrankYang0529:
> > > We should already set `runStrategy` as `Halted`, why this PR is still required?
> > 
> > 
> > This PR is to avoid edge case when the cluster is upgraded from v1.3.2 with off VM. Please take a look for reproduce steps in original issue: #6895
> 
> IIUC, the Harvester controller should always set restoring VM's `runStrategy` as `Halted`, then the VM will be started by
> 
> https://github.com/harvester/harvester/blob/625f14e0c4b568272763f9fd74b87577edef7809/pkg/controller/master/backup/restore.go#L947-L948

Hi @WebberHuang1118, I noticed that the run strategy of VM from v1.3.2 is `Halted` at first. After VM restore replace, it will change to `RerunOnFailure`, because `startVM` sends a start request. However, its status is still stopped, so I think we still need this PR to send another start request to make VM be running.

WebberHuang1118 (approved):
LGTM, thanks for the PR.

PR #6907, opened 2024-10-29
fix: disable new version checks for arm64 arch

bk201:
**Problem:**

We'd like users to receive upgrades when the upgrade path is validated and supported.

**Solution:**

Disable new version checks for now.

**Related Issue:**

https://github.com/harvester/harvester/issues/6257

**Test plan:**


starbops (approved):
LGTM, thanks.

ibrokethecloud (approved):

PR #6912, opened 2024-10-30
upgrade: update harvester csi driver rbac

Vicente-Cheng:
- To support RWX volume feature, we need new rbac for csi-driver role

**Problem:**
`harvesterhci.io:csi-driver` needs new RBAC permissions to support RWX feature.

**Solution:**
Apply for needed permission when upgrade

**Related Issue:**

**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -903,6 +903,56 @@ wait_for_fleet_agent(){
   done
 }
 
+upgrade_harvester_csi_rbac() {
+
+  # only versions before v1.4.0 that upgrading to v1.4.0 need this patch
+  if [[ ! "${UPGRADE_PREVIOUS_VERSION%%-rc*}" < "v1.4.0" ]]; then
+    echo "Only versions before v1.4.0 need this patch."
+    return
+  fi
+
+  if kubectl get clusterrole harvesterhci.io:csi-driver 2> /dev/null; then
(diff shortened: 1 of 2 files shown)


bk201 (approved):

bk201:
Verified new permissions are added.

package/upgrade/upgrade_manifests.sh
```diff
@@ -903,6 +903,56 @@ wait_for_fleet_agent(){
...
+  - longhorn.io
+  resources:
+  - volumes
+  - volumes/status
+  verbs:
+  - get
+  - list
+EOF
+  else
+    echo "ClusterRole harvesterhci.io:csi-driver not found, skip updating."
```
starbops, line 952: Would you mind explaining why we are skipping the update for the ClusterRole when it does not exist? I thought we need the ClusterRole for the CSI driver's RWX function to work correctly. Thank you.

Vicente-Cheng: Hi @starbops, thanks for the review

Because only the user manually removes the ClusterRole, the ClusterRole will not exist.
And it only affects the downstream cluster. I don't want to block the upgrade in this situation.
For sure, we have a simple workaround for manually creating this ClusterRole.

That's why I thought we could skip here.

starbops: That makes sense. Thanks for the explanation.

starbops (approved):
LGTM, thank you.

bk201 (approved):

Issue #6894, opened 2024-10-27
[BUG] After upgrade to v1.4.0-rc4, one image display 0% progress and can't be used anymore 


TachunLin:
**Describe the bug**

After upgrade from v1.3.2 to v1.4.0-rc4, one of the image `sles15-sp4-minimal-vm.x86_64-openstack-cloud-gm.qcow2` 
display 0% progress with message

  ```
  Retry attempted 1/3 failed due to error: failed to process sync file: syncing file should be directly reused but failed
  ```
  ![image](https://github.com/user-attachments/assets/f3e96e61-7fc0-4d4e-8707-01a2b3961675)

* We can't find them in the image list 

  ![image](https://github.com/user-attachments/assets/6c15b696-599b-42ca-9162-a83fc2a08674)

**To Reproduce**
1. Prepare 4 nodes v1.3.2 Harvester cluster 
2. Create image `sles15-sp6-minimal-vm.x86_64-cloud-gm.qcow` and `sles15-sp4-minimal-vm.x86_64-openstack-cloud-gm.qcow2` from download URL
3. Create two virtual machines
4. Integrate with Rancher v2.7.15
5. Provision RKE2 downstream cluster
6. Upgrade Harvester to v1.4.0-rc4
7. Check the image state 

**Expected behavior**

After upgrade to v1.4.0-rc4, all images should keep the original state and can't become unusable. 

**Support bundle**

[supportbundle_93e674a6-935b-4b5b-a86e-815dbed50ba4_2024-10-27T12-23-07Z.zip](https://github.com/user-attachments/files/17533914/supportbundle_93e674a6-935b-4b5b-a86e-815dbed50ba4_2024-10-27T12-23-07Z.zip)

**Environment**
 - Harvester ISO version: v1.3.2 -> v1.4.0-rc4
 - Underlying Infrastructure: 4 nodes remote kvm machines

WebberHuang1118:
By original design, vmi controller doesn't reconcile bi state once bi is created successfully.

However, as introducing virtual size to vmi, after the upgrade, vmi controller would reconcile the bi state, because virtual size is not updated yet https://github.com/harvester/harvester/blob/19bc22db86ec0c9d796ec1d9116c3ddce908082a/pkg/controller/master/image/backing_image_controller.go#L59-L61.

[...]

PR #6915, opened 2024-10-30
fix: Virtual Machine Image should keep track with LH BackingImage's DiskFileStatusMap

WebberHuang1118:
**Problem:**
VirtualMachineImages do not keep in sync with LH BackingImage's DiskFileStatusMap

**Solution:**
VirtualMachineImages should update the status if there is no ready file in LH BackingImage's DiskFileStatusMap

**Related Issue:**
#6894 

**Test plan:**
1. Preparing a v1.3.2 Harvester cluster
2. Creating two images
3. Upgrading Harvester to v1.4.0-rc4
4. Existing images should be in an active state
5. Creating new images successfully

### DIFF
--- a/pkg/controller/master/image/backing_image_controller.go
+++ b/pkg/controller/master/image/backing_image_controller.go
@@ -6,6 +6,7 @@ import (
 
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	"github.com/rancher/norman/condition"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
 
 	harvesterv1beta1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
@@ -23,11 +24,21 @@ type backingImageHandler struct {
 	backingImageCache ctllhv1.BackingImageCache
 }
 
+func backingImageHasReadyDisk(backingImage *lhv1beta2.BackingImage) bool {
+	for _, status := range backingImage.Status.DiskFileStatusMap {
+		if status.State == lhv1beta2.BackingImageStateReady {
+			return true
+		}
+	}
+	return false
+}
+
 func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.BackingImage) (*lhv1beta2.BackingImage, error) {
 	if backingImage == nil || backingImage.DeletionTimestamp != nil {
 		return nil, nil
 	}
-	if backingImage.Annotations[util.AnnotationImageID] == "" || len(backingImage.Status.DiskFileStatusMap) != 1 {
+	// LH introduces image HA since v1.7, we should check if backingImage.Status.DiskFileStatusMap not exists yet
+	if backingImage.Annotations[util.AnnotationImageID] == "" || len(backingImage.Status.DiskFileStatusMap) < 1 {
 		return nil, nil
 	}
 	namespace, name := ref.Parse(backingImage.Annotations[util.AnnotationImageID])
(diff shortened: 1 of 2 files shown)


pkg/controller/master/image/backing_image_controller.go
```diff
@@ -80,6 +104,9 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
 			toUpdate.Status.Progress = status.Progress
 			toUpdate.Status.Size = backingImage.Status.Size
 			toUpdate.Status.VirtualSize = backingImage.Status.VirtualSize
+			toUpdate.Status.Failed = 0
+			// vmi is ready if there is one ready file in backing image file map
+			break
```
brandboat, line 109: I'm a bit confused about this, so backing image works fine even if not all disk files are in BackingImageStateReady state ?

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -397,6 +398,9 @@ func handleFail(image *harvesterv1.VirtualMachineImage, cond condition.Cond, err
 		harvesterv1.ImageRetryLimitExceeded.False(image)
 		harvesterv1.ImageRetryLimitExceeded.Message(image, errMsg)
 	}
+	harvesterv1beta1.ImageImported.Unknown(image)
+	harvesterv1beta1.ImageImported.Reason(image, err.Error())
+	harvesterv1beta1.ImageImported.Message(image, err.Error())
```
brandboat, line 403: Pardon me, why do we need these lines ?

pkg/controller/master/image/backing_image_controller.go
```diff
@@ -56,12 +67,25 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
...
-	if !harvesterv1beta1.ImageImported.IsUnknown(vmImage) && vmImage.Status.VirtualSize == backingImage.Status.VirtualSize {
+
+	// Found LH restruct DiskFileStatusMap in https://github.com/harvester/harvester/issues/6894,
+	// if Harvester finds the restruct, it should keep sync with LH backing image
+	if !harvesterv1beta1.ImageImported.IsUnknown(vmImage) && vmImage.Status.VirtualSize == backingImage.Status.VirtualSize &&
+		backingImageHasReadyDisk(backingImage) {
 		return nil, nil
 	}
+
+	// Todo: LH has bi HA from v1.7, Harvester should consider have the file map structure
```
brandboat, line 78: Do we have an issue for this ? If so, perhaps we can add github issue link here.

pkg/controller/master/image/backing_image_controller.go
```diff
@@ -80,6 +104,9 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
 			toUpdate.Status.Progress = status.Progress
 			toUpdate.Status.Size = backingImage.Status.Size
 			toUpdate.Status.VirtualSize = backingImage.Status.VirtualSize
+			toUpdate.Status.Failed = 0
```
brandboat, line 107: This might have a potential issue, and could make VirtualMachineImage trigger `handleRetry` forever, right ? Assume we have 3 disk files in backingimage, first in failed state, second & third in ready state. And each time we got the same DiskFileStatusMap (i.e. 1 failed, 2 ready), then VirtualMachineImage retry never ends as the failed count is reset to 0 every time, and controller keeps delete BackingImage and StorageClass in each retry.

I'm not sure if this is something we need to aware of, if that's not the case, please ignore this comment.

pkg/controller/master/image/backing_image_controller.go
WebberHuang1118: Yes, as there is one ready file, we can leverage the LH backingimage.

pkg/controller/master/image/backing_image_controller.go
WebberHuang1118: https://github.com/harvester/harvester/pull/6915#discussion_r1827182479
As you mentioned, we should not reset the imported condition, otherwise, it can make the image controller delete/recreate the backing image and storage class, thanks for pointing this out.

pkg/controller/master/image/vm_image_controller.go
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/6915#discussion_r1827285592

pkg/controller/master/image/backing_image_controller.go
WebberHuang1118: Issue link https://github.com/harvester/harvester/issues/6936, thanks.

FrankYang0529 (approved):
LGTM. Thanks for the PR.

WebberHuang1118:
This is not needed, as we apply another solution https://github.com/harvester/harvester/pull/6938

Issue #6936, opened 2024-11-04
[ENHANCEMENT] VM Image Multiple Backend Support


WebberHuang1118:
We should have improvement/refactoring for VM Image implementation:
- Clean interface to accommodate LH Backing image and CDI image
  - As preparation for https://github.com/harvester/harvester/issues/1199, in addition to supporting LH backing image, the VM Image controller should also be able to work with CDI image

votdev:
Verified this bug has been fixed.

Test Information
----
* Environment: **ipxe-examples/vagrant-pxe-harvester 1 node**
* Harvester Version: **v1.5.0-rc1**

### Verify Steps
- Case 1: Creating VM image with upload
![Image](https://github.com/user-attachments/assets/ef44a0fa-11e6-4ee0-837c-35e6aeaf5873)

- Case 2: Creating VM image with download
![Image](https://github.com/user-attachments/assets/045a34df-1823-41f5-81a0-b169f43f51b7)

- Case 3: Creating VM image with export
![Image](https://github.com/user-attachments/assets/1c6a628d-edbb-437a-a26d-d6bdc1df6ba2)

- Case 4: VM Image encryption
![Image](https://github.com/user-attachments/assets/743aec7f-67a1-443f-ad1a-4890ec9584b1)

- Case 5: VM Image decryption
![Image](https://github.com/user-attachments/assets/811c4d77-846a-4c65-85df-0900e62cb666)

- Case 6: VM Image Downloading to local
Compare MD5SUMs of the origin image and the downloaded image.

Overview of the images after the test:
![Image](https://github.com/user-attachments/assets/cbcd8dae-4849-4c6b-b051-1b7c4e8dbcd5)

PR #6938, opened 2024-11-04
fix(image): Once image is imported, not reconcile LH backing image an…

WebberHuang1118:
**Problem:**
To introduce image virtual size https://github.com/harvester/harvester/pull/5387, the VM image controller has a new behavior: [it will sync with the LH backing image even if the VM image is already imported](https://github.com/harvester/harvester/blob/5bf33f8410e88310f41b23ba56f01c14e9add62e/pkg/controller/master/image/backing_image_controller.go#L59-L61), this aims to update the virtual size for existing VM images, which is upgraded from previous Harvester version.

However, for issue #6894, the above behavior could leave the VM image stuck in a temporary state, as the controller does not always sync with the LH backing image.

Since https://github.com/harvester/harvester/blob/5bf33f8410e88310f41b23ba56f01c14e9add62e/pkg/controller/master/image/vm_image_controller.go#L53-L93 also handle the VM image virtual size both for the newly creation and upgrade case, we can rollback VM image controller's behavior: only sync with LH backing image during the create phase.

In addition, issue #6936 is for improving the sync logic between the VM image and LH backing image

**Solution:**
Only sync with LH backing image during the create phase

**Related Issue:**
#6894

**Test plan:**
- Preparing a v1.3.2 Harvester cluster
- Creating two images
- Upgrading Harvester to this PR
- Existing images should be in an active state with valid virtual size
- Creating new images, the images should transit to an active state with the valid virtual size

### DIFF
--- a/pkg/controller/master/image/backing_image_controller.go
+++ b/pkg/controller/master/image/backing_image_controller.go
@@ -50,13 +50,8 @@ func (h *backingImageHandler) OnChanged(_ string, backingImage *lhv1beta2.Backin
 	// been imported, and we think we know everything about it, i.e. we've
 	// now been through a series of progress updates during image download,
 	// and those are finally done, so let's not worry about further updates.
-	// The problem with this logic is that when we add new fields (e.g.
-	// VirtualSize), existing images won't pick up those newly added fields
-	// if we return here immediately.  So, now there's an additional check
-	// for that new field.  Another, simpler, alternative would be to just
-	// drop the ImageImported.IsUnknown check entirely, and let the following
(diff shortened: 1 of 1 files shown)


FrankYang0529 (approved):
LGTM. Thanks for the fix.

tserong (approved):
I went back through previous discussions and found https://github.com/harvester/harvester/pull/5387#issuecomment-2191534523 in support of doing exactly what you're doing here :-)

Thank you!

bk201 (approved):

brandboat:
@mergify backport v1.4

Issue #6972, opened 2024-11-08
[BUG] Harvester upgrade fails with witness node in cluster


SnakebiteEF2000:
**Describe the bug**
The upgrade failed from 1.3.1 to 1.3.2 on a 5 node harvester cluster.
2 physical master nodes
2 physical worker nodes
1 VM witness node

Tee upgrade get stuck at the step "upgrading system service"

After trying the trouble shooting steps in the docs this seems to be the problem:
instance-manager (aio) (image=longhornio/longhorn-instance-manager:v1.6.2) count is not 1 on node servername0000, will retry... 

I suspect the upgrade process somehow is not aware of the witness nodes inability to schedule this service.

**To Reproduce**
Start the upgrade from 1.3.1 to 1.3.2 with witness node in the cluster

**Expected behavior**
Successful upgrade

**Support bundle**
Sadly company policy does not allow the upload of the support bundle!

**Environment**
 - Harvester ISO version: 1.3.1
4 Poweredge R630 (within recommended specs)
1 vsphere based witness node
**Additional context**

Vicente-Cheng:
Hi @SnakebiteEF2000, 

>After trying the trouble shooting steps in the docs this seems to be the problem:
instance-manager (aio) (image=longhornio/longhorn-instance-manager:v1.6.2) count is not 1 on node servername0000, will retry...

Could you check the instance-manager pod number? It should only have one instance-manager pod for each node.

SnakebiteEF2000:
Yes it tries to startup 1 pod per node but the witness node does not start the pod as it is considered workload i guess. thats why the upgrade job just waits for the pod to be scheduled which will never happen.

ibrokethecloud:
would it be possible for you to please generate a support bundle while the cluster is stuck in this state?

bk201:
@SnakebiteEF2000 is `servername0000` the witness node and can you provide:
```
kubectl get instancemanager.longhorn.io -A
```

bk201:
We can reproduce this with a 5-node cluster (2 mgmt, 1 witness, 2 workers).

SnakebiteEF2000:
Hi @bk201,
Sadly I can't provide a support bundle due to company policy. The cluster also got rebuilt from scratch with the 1.3.2 image due to downtime and I am unable to retrieve new information from the cluster. Can I help in any other way like testing?

Edit: Yes `servername0000` is the witness node!

Vicente-Cheng:
[...]

PR #6988, opened 2024-11-12
upgrade: correct the instance-manager checking mechanism when upgrade

Vicente-Cheng:
- Also correct the condition for single node checking

**Problem:**
The instance-manager checking will hang in an environment with more than three nodes that contain the witness node because the witness node did not run any Longhorn pods.

**Solution:**
Should skip the witness node when checking the instance-manager pod

**Related Issue:**
https://github.com/harvester/harvester/issues/6972

**Test plan:**
1. create five node 1.3.2 cluster that contains the witness node
2. Try to upgrade to v1.4.0
3. Upgrade should be successful.


bk201 (approved):
lgtm.
Test on 2 nodes/3 nodes cluster and 5 nodes (1 witness) cluster.

hardcore75477812322423 (approved):

ibrokethecloud (approved):

PR #7004, opened 2024-11-14
fix(upgrade): move waiting apiServerURL before waiting fleet cluster

FrankYang0529:

**Problem:**
If `apiServerURL` is empty and it's filled with a new value, a new fleet-agent will start. We do harvester managed chart upgrade after the check. In this case, there may be a race condition.

**Solution:**

Move the check before waiting for a ready fleet cluster, so the new agent doesn't cause race conditions.

**Related Issue:**

**Test plan:**

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -110,10 +110,10 @@ wait_rollout_with_loop() {
 wait_cluster_local_and_fleet() {
   wait_new_fileds_in_cluster_fleet_crd
   wait_new_fileds_in_fleet_controller_configmap
+  wait_apiServerURL_in_fleet_controller_configmap
   wait_cluster_local_is_imported
   wait_fleet_agent_is_redeployed
   wait_cluster_local_is_ready
-  wait_apiServerURL_in_fleet_controller_configmap
 }
 
 debug_cluster_local_and_fleet() {
--- a/pkg/controller/master/mcmsettings/fleetcontroller.go
+++ b/pkg/controller/master/mcmsettings/fleetcontroller.go
@@ -36,7 +36,7 @@ func (h *mcmSettingsHandler) watchInternalEndpointSettings(_ string, setting *mg
 		return setting, nil
 	}
 
-	logrus.Debugf("reconcilling fleet-controller config for setting %s", setting.Name)
+	logrus.Infof("reconcilling fleet-controller config for setting %s", setting.Name)
 
 	fleetControllerConfig, err := h.configMapCache.Get(util.DefaultFleetControllerConfigMapNamespace, util.DefaultFleetControllerConfigMapName)
 	if err != nil {


pkg/controller/master/mcmsettings/fleetcontroller.go
```diff
@@ -36,7 +36,7 @@ func (h *mcmSettingsHandler) watchInternalEndpointSettings(_ string, setting *mg
 		return setting, nil
 	}
 
-	logrus.Debugf("reconcilling fleet-controller config for setting %s", setting.Name)
+	logrus.Infof("reconcilling fleet-controller config for setting %s", setting.Name)
```
bk201, line 39: @ibrokethecloud this is a suggestion from me just to make sure we didn't accidentally set the api URL again.

package/upgrade/upgrade_manifests.sh
```diff
@@ -110,10 +110,10 @@ wait_rollout_with_loop() {
 wait_cluster_local_and_fleet() {
   wait_new_fileds_in_cluster_fleet_crd
   wait_new_fileds_in_fleet_controller_configmap
+  wait_apiServerURL_in_fleet_controller_configmap
```
w13915984028, line 113: Sorry I did not find the fundmental differences on putting this function on L113 or L116.

The upgrade POD is running in parallel with Harvester POD, which has reconciller controller;  it seems after the upgrade of Rancher, the related Harvester old controller (v1.3.*) defected (it should also be notified about the change) https://github.com/harvester/harvester/blob/518d4436fcb755159384811b66dfe0d92bfe35c8/pkg/controller/master/mcmsettings/fleetcontroller.go#L46 

Another solution is to add a new function here, which uses `kubectl` to fulfill the abover controller's task, update the configmap directly

Those two settings of Rancher seem not empty:

```
internal-cacerts                          -----BEGIN CERTIFICATE-----...
internal-server-url                       https://10.53.198.41

apiVersion: management.cattle.io/v3
..
default: ""
kind: Setting
..
value: https://10.53.198.41

```

w13915984028 (commented):
It looks the mcm controller and fleet https://github.com/rancher/fleet/issues/3061 together cause a possible rollback of Harvester managedchart:

The upgarded Harvester POD resets the configmap; the fleet-controller then re-deploy the fleet-agent; the new fleet-agent rollbacks the Harvester managedchart if the previous upgrade is not finished.

This is a race condition and it does not happen always.

To break the race, the safe way is to:

in the upgrade scrip, update the configmap  after the Rancher is upgrade but before Harvester managedchart is upgraded.

package/upgrade/upgrade_manifests.sh
w13915984028: I am adding a PR https://github.com/harvester/harvester/pull/7022 to save & restore the configmap inbetween upgrade script.

FrankYang0529:
Close the PR since we prefer another one https://github.com/harvester/harvester/pull/7022

Issue #7005, opened 2024-11-14
[BUG] Upgrade  stuck with Pre-draining due to fail to migrate a Running VM with Container Disk


albinsun:
**Describe the bug**
`v1.3.2` -> `v1.4.0-rc5` witness cluster upgrade stuck with `Pre-draining` due to Windows VM can not migrate.
```
Warning  Failed                  5m46s (x3 over 6m30s)  kubelet                  Failed to pull image "registry.suse.com/suse/vmdp/vmdp@sha256:9056 │
│ 7499610d496f7973e5b483cefa18d99c2f3c8ef9771547c93ba476a7872e": rpc error: code = NotFound desc = failed to pull and unpack image "registry.suse.com/s │
│ use/vmdp/vmdp@sha256:90567499610d496f7973e5b483cefa18d99c2f3c8ef9771547c93ba476a7872e": failed to unpack image on snapshotter overlayfs: unexpected m │
│ edia type application/octet-stream for sha256:90567499610d496f7973e5b483cefa18d99c2f3c8ef9771547c93ba476a7872e: not found
```
![image](https://github.com/user-attachments/assets/07633e97-6033-4ff1-a7e2-4ae7bb24caa6)

**To Reproduce**
1. Setup 3 nodes witness `harvester-v1.3.2`
2. Create Linux and Windows 2022 VMs (Running)
4. Trigger upgrade to `harvester-v1.4.0-rc5`

**Expected behavior**
Windows VM should be able to migrate.

**Support bundle**
* Support Bundle: [supportbundle_be8be266-baf1-4923-85ea-82c65ec77cfe_2024-11-14T08-27-21Z.zip](https://github.com/user-attachments/files/17745553/supportbundle_be8be266-baf1-4923-85ea-82c65ec77cfe_2024-11-14T08-27-21Z.zip)
* Upgrade Log: [hvst-upgrade-k2jxv-upgradelog-archive-2024-11-14T08-26-24Z.zip](https://github.com/user-attachments/files/17745498/hvst-upgrade-k2jxv-upgradelog-archive-2024-11-14T08-26-24Z.zip)

**Environment**
 - Harvester ISO version: `v1.3.2` -> `v1.4.0-rc5`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 nodes baremetal 

**Additional context**

albinsun:
[...]

PR #7008, opened 2024-11-15
fix(webhook): check vm with container disks before upgrade

FrankYang0529:

**Problem:**
A VM with container disk can't be migrated. It will block upgrade in pre-drain status.

**Solution:**
Add webhook to check VM before starting upgrade.

**Related Issue:**
https://github.com/harvester/harvester/issues/7005

**Test plan:**
1. Create a new Harvester cluster.
2. Create a Windows VM and run it.
3. Create an upgrade. The webhook should deny the request.
4. Stop the Windows VM.
5. Create an upgrade again. It should pass.

### DIFF
--- a/pkg/util/virtualmachineinstance/virtualmachineinstance.go
+++ b/pkg/util/virtualmachineinstance/virtualmachineinstance.go
@@ -47,6 +47,13 @@ func GetAllNonLiveMigratableVMINames(vmis []*kubevirtv1.VirtualMachineInstance, 
 		if !migratable {
 			logrus.Infof("%s considered non-live migratable due to node affinities", vmiNamespacedName)
 			nonLiveMigratableVMINames = append(nonLiveMigratableVMINames, vmiNamespacedName)
+			continue
+		}
+
+		// container-disk or cdrom device
+		if VMContainsCDRomOrContainerDisk(vmi) {
+			nonLiveMigratableVMINames = append(nonLiveMigratableVMINames, vmiNamespacedName)
+			continue
 		}
 	}
 
@@ -103,4 +110,23 @@ func ListByNode(node *corev1.Node, selector labels.Selector, cache ctlkubevirtv1
 		return nil, fmt.Errorf("failed to list VMIs on node %s: %w", node.Name, err)
 	}
 	return list, nil
+}
+
+func VMContainsCDRomOrContainerDisk(vmi *kubevirtv1.VirtualMachineInstance) bool {
+	if vmi == nil {
+		return false
+	}
+
+	for _, disk := range vmi.Spec.Domain.Devices.Disks {
+		if disk.CDRom != nil {
+			return true
+		}
+	}
+
+	for _, volume := range vmi.Spec.Volumes {
+		if volume.VolumeSource.ContainerDisk != nil {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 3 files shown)


pkg/webhook/resources/upgrade/validator.go
```diff
@@ -238,6 +246,29 @@ func (v *upgradeValidator) checkManagedCharts() error {
 	return nil
 }
 
+// Since volume snapshot/backup will hold one additional LH VA tickets, it can block VM live migration.
```
bk201: how about extend the code in the function too: https://github.com/FrankYang0529/harvester/blob/9d1180fa081a5d367d56c8ad0a437862868688d0/pkg/webhook/resources/upgrade/validator.go#L409

FrankYang0529: Sounds good to me. Updated it. Thanks for the suggestion.

brandboat (approved):
LGTM, thanks!

bk201 (approved):

w13915984028:
Please wait a moment before merging.

pkg/util/virtualmachineinstance/virtualmachineinstance.go
```diff
@@ -47,6 +47,33 @@ func GetAllNonLiveMigratableVMINames(vmis []*kubevirtv1.VirtualMachineInstance,
 		if !migratable {
 			logrus.Infof("%s considered non-live migratable due to node affinities", vmiNamespacedName)
 			nonLiveMigratableVMINames = append(nonLiveMigratableVMINames, vmiNamespacedName)
+			continue
+		}
+
+		// container-disk
```
w13915984028: There are similar functions on 
https://github.com/harvester/harvester/blob/3b9a3e36125de9b4c500703803bd7807e89a7bf5/pkg/controller/master/nodedrain/nodedrain_controller.go#L390

and 
https://github.com/harvester/harvester/blob/3b9a3e36125de9b4c500703803bd7807e89a7bf5/pkg/api/vm/handler.go#L411

Could we refactor those 3 places to a common function? thanks.

starbops (approved):
LGTM, thank you.

pkg/util/virtualmachineinstance/virtualmachineinstance.go
FrankYang0529: Updated it. Thanks for the suggestion.

pkg/util/virtualmachineinstance/virtualmachineinstance.go
```diff
@@ -104,3 +111,18 @@ func ListByNode(node *corev1.Node, selector labels.Selector, cache ctlkubevirtv1
 	}
 	return list, nil
 }
+
+func VMContainsCDRomOrContainerDisk(vmi *kubevirtv1.VirtualMachineInstance) bool {
```
w13915984028, line 115: As a public function, suggest to add a first check if vmi is nil, then return false directly.

w13915984028 (approved):
LGTM, thanks for the refactor.

pkg/util/virtualmachineinstance/virtualmachineinstance.go
FrankYang0529: Updated it.

Issue #4565, opened 2023-09-25
[BUG] fleet-controller complains "missing apiServerURL" after the upgrade


bk201:
**Describe the bug**

Upgrade from v1.1.2 to v1.2.0 and fleet-controller shows:
```
time="2023-09-22T07:17:42Z" level=error msg="error syncing 'fleet-local/local': handler import-cluster: missing apiServerURL in fleet config for cluster auto registration, requeuing"
time="2023-09-22T07:17:42Z" level=error msg="error syncing 'fleet-local/local': handler import-cluster: missing apiServerURL in fleet config for cluster auto registration, requeuing"
time="2023-09-22T07:17:44Z" level=error msg="error syncing 'fleet-local/local': handler import-cluster: missing apiServerURL in fleet config for cluster auto registration, requeuing"
time="2023-09-22T07:20:26Z" level=error msg="error syncing 'fleet-local/local': handler import-cluster: missing apiServerURL in fleet config for cluster auto registration, requeuing"
time="2023-09-22T07:20:26Z" level=error msg="error syncing 'fleet-local/local': handler import-cluster: missing apiServerURL in fleet config for cluster auto registration, requeuing"
```

The messages seem to print every 10 minutes.

**To Reproduce**
1. Provision a 1.1.2 cluster.
2. Upgrade to v1.2.0.
3. Observe the fleet-controller pod log.

**Expected behavior**

There should be no error log in the fleet-controller pod.

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
- Note the fleet-agent pod still runs well, but it's not the "new agent" supposed to be redeployed by the fleet-controller.
- Related: https://github.com/rancher/rancherd/pull/31, we set API URL in the bootstrap path.

irishgordo:
@ibrokethecloud after testing this, I'm still seeing that issue present in the logs with an upgrade from:
v1.1.2 -> v1.2.1-rc1

**validation issue/failure info:**

It seems that that creeps up in the log during the [upgrading system services](https://docs.harvesterhci.io/v1.2/upgrade/troubleshooting#phase-3-upgrade-system-services) phase, if I'm auditing things correctly :sweat_smile: 

I apologize that the logs gathered may not be the cleanest, as I was appending a file with:
```bash
[...]

PR #7034, opened 2024-11-25
disable fleet-controller-configmap updates

ibrokethecloud:

**Problem:**

Rollback changes to reconcile fleet-controller configmap introduced as part of: https://github.com/harvester/harvester/issues/4565

This should not longer be need as the upgrade path is already address in v1.2.1 release.

The update to configmap seems to be causing the fleet-controller to be redeployed due to changes in fleet-controller.

**Solution:**

**Related Issue:**
**Test plan:**


w13915984028:
@ibrokethecloud Do we plan to remove this controller on v1.5.0 ?

ibrokethecloud:
@w13915984028 i was hoping to remove this as it was only added for 1.2.x to work around a known bug related to rancher + fleet. We have had to add more work arounds to keep this working so may be best to drop it and simplify the upgrade path.

w13915984028:
@ibrokethecloud That sounds great. BTW, could you create a new issue for this purpose, we will have enough time to cover all possible scenarios on v1.4.1/v1.5.0.

Issue #5835, opened 2024-05-20
[FEATURE] CPU / Memory hotplug support


innobead:
**Is your feature request related to a problem? Please describe.**

KubeVirt 1.1 supports CPU / Memory hotplug to dynamically update cpu / memory claims for VMs at runtime. Harvester should leverage this feature to have flexible and efficient resource management. 

**Additional context**

- https://kubevirt.io/user-guide/compute/cpu_hotplug/
- https://kubevirt.io/user-guide/operations/memory_hotplug/
- https://kubevirt.io/user-guide/operations/node_overcommit/

> Memory hotplug was introduced in KubeVirt version 1.1, enabling the dynamic resizing of the amount of memory available to a running VM.

FrankYang0529:
Some prerequisite for this feature:

* Bump KubeVirt to v1.2, because there is some redesign for hotplug feature (https://github.com/kubevirt/kubevirt/issues/11900#issuecomment-2111775208)
* UI Enhancement:
  * UI set memory to `spec.template.spec.domain.resources.limits.memory`, but this feature can't have value in this field.
  * If we manually remove `spec.template.spec.domain.resources.limits.memory`, the dashboard can't show correct memory.
  <img width="974" alt="Screenshot 2024-07-17 at 4 21 37 PM" src="https://github.com/user-attachments/assets/3c54c2bb-87a5-4978-b730-110db0104df9">

FrankYang0529:
To implement the feature, we may need to change following parts:

1. VM memory setting on UI.
When we set a VM memory on UI, it sets the value to the `resources.limits.memory` field. However, this blocks memory hotplug, because the limitation is set. Also, the `resources` field is not live upgradable, so we can't send a command with two patches for `resources.limits.memory` and `memory.guest`. We need to set the initial memory value to the `memory.guest` field.

2. Add two fields `Max CPU Hotplug` and `Max Memory Hotplug`.
The value in VM `resources.limits` is set to pod `resources.limits`. If we don't set `resources.limits` on VM, the pod may use whole node resources. We can have two new fields `Max CPU Hotplug` and `Max Memory Hotplug` on UI, so we can use these values as `resources.limits`.

3. Set `MacAddress` to annotation.
[...]

PR #7084, opened 2024-12-03
feat: add cpu / memory hotplug config

FrankYang0529:

**Solution:**
KubeVirt 1.2 supports CPU / Memory hotplug to dynamically update cpu / memory claims for VMs at runtime. Harvester should leverage this feature to have flexible and efficient resource management.

**Related Issue:**

**Test plan:**

### Mac Address

1. Create a VM.
2. Check there is `harvesterhci.io/mac-address` annotation on the VM and the value keeps mac address of different interfaces in VMI.

```yaml
# vm
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/mac-address: '{"default":"be:40:c3:88:75:7c"}'
```

```yaml
# vmi
status:
  interfaces:
  - infoSource: domain, guest-agent
    interfaceName: enp1s0
    ipAddress: 10.52.2.19
    ipAddresses:
    - 10.52.2.19
    mac: be:40:c3:88:75:7c
    name: default
    queueCount: 1
```

3. Stop / Restart VM and check the mac address is updated to VM devices.

```yaml
# vm
spec:
  template:
    spec:
      domain:
          interfaces:
          - macAddress: be:40:c3:88:75:7c
            masquerade: {}
            model: virtio
            name: default
```

### CPU / Memory hotplug

Since we need some update from UI, we can't create VM from UI currently.

1. Create a VMImage.

```yaml
kind: VirtualMachineImage
metadata:
  name: focal-image
  namespace: default
spec:
  checksum: ""
  displayName: focal-server-cloudimg-amd64.img
  pvcName: ""
  pvcNamespace: ""
  retry: 3
  sourceType: download
  storageClassParameters:
    migratable: "true"
    numberOfReplicas: "3"
    staleReplicaTimeout: "30"
    dataLocality: "strict-local"
  url: http://192.168.0.181:8000/focal-server-cloudimg-amd64.img
```

2. Create cloudinit secret.

```yaml
apiVersion: v1
data:
  networkdata: ""
  userdata: I2Nsb3VkLWNvbmZpZwpwYWNrYWdlX3VwZGF0ZTogdHJ1ZQpwYWNrYWdlczoKICAtIHFlbXUtZ3Vlc3QtYWdlbnQKcnVuY21kOgogIC0gLSBzeXN0ZW1jdGwKICAgIC0gZW5hYmxlCiAgICAtIC0tbm93CiAgICAtIHFlbXUtZ3Vlc3QtYWdlbnQuc2VydmljZQpwYXNzd29yZDogdGVzdApjaHBhc3N3ZDoKICBleHBpcmU6IGZhbHNlCnNzaF9wd2F1dGg6IHRydWUK
kind: Secret
metadata:
  labels:
    harvesterhci.io/cloud-init-template: harvester
  name: cloudinit-with-default-password
  namespace: default
type: secret
```

3. Create VM with 1 CPU and 1Gi memory.

```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/volumeClaimTemplates: '[{"metadata":{"name":"vm1-disk-0-tioib","annotations":{"harvesterhci.io/imageId":"default/focal-image"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"volumeMode":"Block","storageClassName":"longhorn-focal-image"}}]'
  labels:
  name: vm1
  namespace: default
spec:
  runStrategy: RerunOnFailure
  template:
    metadata:
      annotations:
      creationTimestamp: null
      labels:
        harvesterhci.io/vmName: vm1
    spec:
      affinity: {}
      architecture: amd64
      domain:
        cpu:
          cores: 1
          sockets: 1
          threads: 1
        memory:
          guest: 1Gi
        devices:
          disks:
          - bootOrder: 1
            disk:
              bus: virtio
            name: disk-0
          - disk:
              bus: virtio
            name: cloudinitdisk
          inputs:
          - bus: usb
            name: tablet
            type: tablet
          interfaces:
          - masquerade: {}
            model: virtio
            name: default
        features:
          acpi:
            enabled: true
        machine:
          type: q35
      evictionStrategy: LiveMigrate
      hostname: vm1
      networks:
      - name: default
        pod: {}
      terminationGracePeriodSeconds: 120
      volumes:
      - name: disk-0
        persistentVolumeClaim:
          claimName: vm1-disk-0-tioib
      - cloudInitNoCloud:
          networkDataSecretRef:
            name: cloudinit-with-default-password
          secretRef:
            name: cloudinit-with-default-password
        name: cloudinitdisk
```

4. Do CPU hotplug.

```
> kubectl patch vm vm1 --type='json' \
-p='[{"op": "replace", "path": "/spec/template/spec/domain/cpu/sockets", "value": 2}]'
```

5. Check the CPU sockets on VMI is `2`. The vCPU in KubeVirt is cores * sockets * threads.

```yaml
# vmi
spec:
  domain:
    cpu:
      cores: 1
      maxSockets: 4
      model: host-model
      sockets: 2
      threads: 1
```

6. Do memory hotplug.

```
> kubectl patch vm vm1 -p='[{"op": "replace", "path": "/spec/template/spec/domain/memory/guest", "value": "2Gi"}]' --type='json'
```

7. Check the Memory on VMI is `2Gi`.

```yaml
# vmi
spec:
  domain:
    memory:
      guest: 2Gi
      maxGuest: 4Gi
```

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -43,6 +43,7 @@ const (
 	dismissInsufficientResourceQuota = "dismissInsufficientResourceQuota"
 	updateResourceQuotaAction        = "updateResourceQuota"
 	deleteResourceQuotaAction        = "deleteResourceQuota"
+	cpuAndMemoryHotplug              = "cpuAndMemoryHotplug"
 )
 
 type vmformatter struct {
@@ -113,6 +114,10 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	if canMigrate(vmi) {
 		resource.AddAction(request, migrate)
 		resource.AddAction(request, findMigratableNodes)
+
+		if canCPUAndMemoryHotplug(vm) {
+			resource.AddAction(request, cpuAndMemoryHotplug)
+		}
 	}
 
 	if canAbortMigrate(vmi) {
@@ -399,4 +404,27 @@ func canDismissInsufficientResourceQuota(vm *kubevirtv1.VirtualMachine) bool {
 		return false
 	}
 	return true
+}
+
+func canCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
+		return false
+	}
+
+	hasRestartRequiredOrHotplugMigration := false
+	for _, condition := range vm.Status.Conditions {
+		if condition.Type == kubevirtv1.VirtualMachineRestartRequired && condition.Status == corev1.ConditionTrue {
+			hasRestartRequiredOrHotplugMigration = true
+			break
+		}
+		if string(condition.Type) == string(kubevirtv1.VirtualMachineInstanceVCPUChange) && condition.Status == corev1.ConditionTrue {
+			hasRestartRequiredOrHotplugMigration = true
+			break
+		}
+		if string(condition.Type) == string(kubevirtv1.VirtualMachineInstanceMemoryChange) && condition.Status == corev1.ConditionTrue {
+			hasRestartRequiredOrHotplugMigration = true
+			break
+		}
+	}
+	return !hasRestartRequiredOrHotplugMigration
 }
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -26,6 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/storage/names"
@@ -268,6 +269,12 @@ func (h *vmActionHandler) doAction(rw http.ResponseWriter, r *http.Request) erro
 			return apierror.NewAPIError(validation.PermissionDenied, "User does not have permission to update resource quota")
 		}
 		return h.deleteResourceQuota(namespace, name)
+	case cpuAndMemoryHotplug:
+		var input CPUAndMemoryHotplugInput
+		if err := json.NewDecoder(r.Body).Decode(&input); err != nil {
+			return apierror.NewAPIError(validation.InvalidBodyContent, "Failed to decode request body: %v "+err.Error())
+		}
+		return h.cpuAndMemoryHotplug(namespace, name, input)
 	default:
 		return apierror.NewAPIError(validation.InvalidAction, "Unsupported action")
 	}
@@ -1526,4 +1533,37 @@ func (h *vmActionHandler) deleteResourceQuota(namespace, name string) error {
 		return err
 	}
 	return nil
+}
+
+func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUAndMemoryHotplugInput) error {
+	if input.Sockets == 0 && input.Memory == "" {
+		return nil
+	}
+
+	patchOps := []string{}
+	if input.Sockets != 0 {
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/domain/cpu/sockets", "value": %d}`, input.Sockets))
+	}
+
+	if input.Memory != "" {
+		memory, err := resource.ParseQuantity(input.Memory)
+		if err != nil {
+			return fmt.Errorf("failed to parse memory quantity: %v", err)
+		}
+
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/domain/memory/guest", "value": "%s"}`, memory.String()))
+	}
+
+	if len(patchOps) == 0 {
+		return nil
+	}
+
+	logrus.WithFields(logrus.Fields{
+		"namespace": namespace,
+		"name":      name,
+		"patchOps":  patchOps,
+	}).Info("patch cpu and memory hotplug")
+	patchData := fmt.Sprintf("[%s]", strings.Join(patchOps, ","))
+	_, err := h.vms.Patch(namespace, name, k8stypes.JSONPatchType, []byte(patchData))
+	return err
 }
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -68,7 +68,9 @@ kubevirt:
       ## Specify kubevirt feature gates
       vmStateStorageClass: "vmstate-persistence"
       developerConfiguration:
-        featureGates: ["LiveMigration", "HotplugVolumes", "HostDevices", "GPU", "CPUManager", "VMPersistentState"]
+        featureGates: ["LiveMigration", "HotplugVolumes", "HostDevices", "GPU", "CPUManager", "VMPersistentState", "VMLiveUpdateFeatures"]
+
+      vmRolloutStrategy: "LiveUpdate"
 
       ## Specify the network configuration of VirtualMachineInstance.
       ##
@@ -108,6 +110,10 @@ kubevirt:
           resourceName: virt-handler
           resourceType: DaemonSet
           type: strategic
+
+    workloadUpdateStrategy:
+      workloadUpdateMethods:
+        - LiveMigrate
 
 ###########################################################################
 ###########################################################################
@@ -430,7 +436,7 @@ harvester-networkfs-manager:
 
   image:
     repository: rancher/harvester-networkfs-manager
-    pullPolicy: IfNotPresent 
+    pullPolicy: IfNotPresent
     # Overrides the image tag whose default is the chart appVersion.
     tag: main-head
 
@@ -447,7 +453,7 @@ harvester-node-disk-manager:
     pullPolicy: Always
     # Overrides the image tag whose default is the chart appVersion.
     tag: master-head
-  
+
   webhook:
     image:
       repository: rancher/harvester-node-disk-manager-webhook
--- a/pkg/api/vm/schema.go
+++ b/pkg/api/vm/schema.go
@@ -36,6 +36,7 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, options config
 	server.BaseSchemas.MustImportAndCustomize(AddVolumeInput{}, nil)
 	server.BaseSchemas.MustImportAndCustomize(RemoveVolumeInput{}, nil)
 	server.BaseSchemas.MustImportAndCustomize(CloneInput{}, nil)
+	server.BaseSchemas.MustImportAndCustomize(CPUAndMemoryHotplugInput{}, nil)
 
 	dataVolumeClient := scaled.CdiFactory.Cdi().V1beta1().DataVolume()
 	vms := scaled.VirtFactory.Kubevirt().V1().VirtualMachine()
@@ -138,6 +139,7 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, options config
 				dismissInsufficientResourceQuota: &actionHandler,
 				updateResourceQuotaAction:        &actionHandler,
 				deleteResourceQuotaAction:        &actionHandler,
+				cpuAndMemoryHotplug:              &actionHandler,
 			}
 			apiSchema.ResourceActions = map[string]schemas.Action{
 				startVM:    {},
@@ -179,6 +181,9 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, options config
 					Input: "updateResourceQuotaInput",
 				},
 				deleteResourceQuotaAction: {},
+				cpuAndMemoryHotplug: {
+					Input: "cpuAndMemoryHotplugInput",
+				},
 			}
 		},
 		Formatter: vmformatter.formatter,
--- a/pkg/api/vm/types.go
+++ b/pkg/api/vm/types.go
@@ -51,3 +51,8 @@ type FindMigratableNodesOutput struct {
 type UpdateResourceQuotaInput struct {
 	TotalSnapshotSizeQuota string `json:"totalSnapshotSizeQuota"`
 }
+
+type CPUAndMemoryHotplugInput struct {
+	Sockets uint32 `json:"sockets"`
+	Memory  string `json:"memory"`
+}


pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -109,3 +110,38 @@ func (h *VMIController) removeDeprecatedFinalizer(_ string, vmi *kubevirtv1.Virt
...
+	}
+
+	macAddresses := map[string]string{}
+	if err = json.Unmarshal([]byte(macAddressStr), &macAddresses); err != nil {
+		return vmi, err
+	}
+
+	vmCopy := vm.DeepCopy()
+	for i, nic := range vm.Spec.Template.Spec.Domain.Devices.Interfaces {
+		if mac, ok := macAddresses[nic.Name]; ok {
```
w13915984028: if vm template already has MAC (e.g. user set it maunally), the annotion should avoid overwrite it

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -109,3 +110,38 @@ func (h *VMIController) removeDeprecatedFinalizer(_ string, vmi *kubevirtv1.Virt
...
+		return vmi, err
+	}
+
+	macAddressStr := vm.Annotations[util.AnnotationMacAddressName]
+	if macAddressStr == "" {
+		return vmi, nil
+	}
+
+	macAddresses := map[string]string{}
+	if err = json.Unmarshal([]byte(macAddressStr), &macAddresses); err != nil {
```
w13915984028: For this error, maybe a better way is to log an error message, and then return (like there is no MAC annotation), otherwise, the VM will be blocked, reconciller won't solve such case

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -60,56 +54,23 @@ func (h *VMNetworkController) updateVMDefaultNetworkMacAddress(vmi *kubevirtv1.V
...
-
-	revisionSpec := &vmwatch.VirtualMachineRevisionData{}
-	if err = json.Unmarshal(crObj.Data.Raw, revisionSpec); err != nil {
-		return err
-	}
-	if !equality.Semantic.DeepEqual(revisionSpec.Spec.Template, vm.Spec.Template) && vm.Generation != vm.Status.ObservedGeneration {
-		patch, patchGenError := patchVMRevision(vm)
-		if patchGenError != nil {
-			return fmt.Errorf("error during patch generation for vmi %s-%s: %v", vmi.Name, vmi.Name, patchGenError)
+	if !reflect.DeepEqual(vm, vmCopy) {
```
w13915984028: in this case, compare the Annotations[util.AnnotationMacAddressName] and  string(vmiInterfaceBytes) seems to be enough

w13915984028 (commented):
There are also some other cases:

(1) When a VM deletes one interface, should the related MAC-annotation be removed? if it is left, it may bring side effects like when remove & re-add an interface later, the previous saved MAC is unwanted kept, but DHCP server has allocate this MAC to others

(2) When a VM is cloned, the original possible MAC-annotations

(3) MAC-spoof, the protection of MAC-annotation

Both this PR and https://github.com/harvester/harvester/pull/7402 has it's advantages and drawbacks.

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -109,3 +110,38 @@ func (h *VMIController) removeDeprecatedFinalizer(_ string, vmi *kubevirtv1.Virt
 	}
 	return vmi, nil
 }
+
+// ReconcileFromHostLabels handles the propagation of metadata from node labels to VirtualMachineInstance annotations.
```
ihcsim: Please update the method comment to match its implementation. Thanks.

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -60,56 +54,23 @@ func (h *VMNetworkController) updateVMDefaultNetworkMacAddress(vmi *kubevirtv1.V
...
 
-	if equality.Semantic.DeepEqual(vmCopy.Spec.Template.Spec.Domain.Devices, vm.Spec.Template.Spec.Domain.Devices) {
-		return nil
-	}
-	if _, err := h.vmClient.Update(vmCopy); err != nil {
-		return err
+	if vmCopy.Annotations == nil {
+		vmCopy.Annotations = make(map[string]string)
 	}
+	vmCopy.Annotations[util.AnnotationMacAddressName] = string(vmiInterfaceBytes)
```
ihcsim: Is there a limit to how many interfaces can be added to the VM? If it is unbounded, we might need to validate the size of this annotation map before attempting to persist it, using https://pkg.go.dev/k8s.io/apimachinery/pkg/api/validation#ValidateAnnotationsSize.

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -109,3 +110,38 @@ func (h *VMIController) removeDeprecatedFinalizer(_ string, vmi *kubevirtv1.Virt
...
+
+	macAddresses := map[string]string{}
+	if err = json.Unmarshal([]byte(macAddressStr), &macAddresses); err != nil {
+		return vmi, err
+	}
+
+	vmCopy := vm.DeepCopy()
+	for i, nic := range vm.Spec.Template.Spec.Domain.Devices.Interfaces {
+		if mac, ok := macAddresses[nic.Name]; ok {
+			vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces[i].MacAddress = mac
```
ihcsim: Looking at the code history, am I right in saying that this is an one-off that happens when the VM first starts? After that, the MAC addresses shouldn't change because they are already defined in the spec? If yes, it means that we only need to do this assignment when the VM's `metadata.generation` is `1`? Conversely, the annotation can be ignored in subsequent generations?

ihcsim (commented):
The hotplug code LGTM. If the MAC address issue becomes too complicated, we might want to consider splitting them into a different PR, so that we can approve and merge the hotplug feature first.

pkg/controller/master/virtualmachine/vmi_controller.go
w13915984028: This assumption is not true. The real cases are very complex:

(1) For the VM's spec, user has freedom to change the spec at any time, add/remove interface, change MAC...,  on a running VM, Harvester UI will hint user to `restart` the VM after certain change, but the user may choose `no`, then the VM will show a difference on the `generation` and `observedGeneration`, this is the initial design from kubevirt. User can even change the VM's spec many times, that will increase the one of them each time.

(2) For a stopped VM, it is similar, except there is no CR object to store the object, CR is only created when a VM is clicked the `Start`, it is a snapshot of current VM spec.

(3) Originally, to help VM to have fixed IP even after rebooting, Harvester tries to backfill the VM's dynamic MAC to VM's spec;  meanwhile, to clean the warning of `generation is X, but the observedGeneration is Y`, we tried to delete the CR and recreate it as well, to hack kubevirt.

those are described on my PR's test case https://github.com/harvester/harvester/pull/7402#issue-2798817503

(4) When the MACs are kept on VM's annotations instead of SPEC (original solution), we face another challenge: the truth of source. Especially for the stopped VM, if user manipulate the SPEC, the annotation MAC needs to cooperate with it:  when interfaces are removed, annotation needs to remove the related interface,  

the webhook looks to be the place to do this task, which has old & new VM object and if any interface has change can be figured out.

(5) For cloned VM, the anntations need to be removed.

(6) If on interface on the VM's spec has MAC, the annotations' MAC can't overwrite it. namely, annotation's data is in lower priority.

...

Anyway, no matter which solution, the test cases on https://github.com/harvester/harvester/pull/7402#issue-2798817503 need to be covered thoroughly.

w13915984028:
Leave the question as well: the cooperation between `hot plug` and `ResourceQuota`.

pkg/controller/master/virtualmachine/vmi_network_controller.go
FrankYang0529: Good catch. Thanks for the suggestion. I need to check whether KubeVirt limits this or whether there is limitation in Qemu.

pkg/controller/master/virtualmachine/vmi_controller.go
FrankYang0529: > (3) Originally, to help VM to have fixed IP even after rebooting, Harvester tries to backfill the VM's dynamic MAC to VM's spec; meanwhile, to clean the warning of generation is X, but the observedGeneration is Y, we tried to delete the CR and recreate it as well, to hack kubevirt.

Although we can fix the controller revision, we cannot fix `RestartRequired` condition in VM object. This condition blocks VM cpu/memory hotplug.

> the webhook looks to be the place to do this task, which has old & new VM object and if any interface has change can be figured out.

I tried to do this in webhook. However, it cannot cover all cases, because the value is between VMI and VM, not old and new VM.

> (5) For cloned VM, the anntations need to be removed.

Agree, we need to fix this. Thanks

> (6) If on interface on the VM's spec has MAC, the annotations' MAC can't overwrite it. namely, annotation's data is in lower priority.

Updated in this PR.

w13915984028: Yes `RestartRequired` is the side-effect of the original solution. Technically, I prefer the annotation based solution rather than hack kubevirt.

>I tried to do this in webhook. However, it cannot cover all cases, because the value is between VMI and VM, not old and new VM.

It is still needed in webhook: e.g.
(a) User create a VM with 2 interfaces without MAC, after VM starts, 2 MACs are generated and writen to VM's annotations
(b.1) User stops the VM
(b.2) User removes if1, 
(b.3) later adds if1;  if saved MAC is not cleaned, it will apply to new if1

(c.1) user does not stop VM
(c.2) User removes if1, 
(c.3) and later adds if1, but do't restart the VM either.

For both case, at step (b.2) and (c.2), the VM webhook needs to figure out the removed if1, and remove it from the existing annotation.

For case (c.3),  when adding MAC to annotation, maybe we can check VM's `generation` and `observedGeneration`, only when they are same (means user does not modify the VM manually), then add/update to annotation; otherweise, skip writing annotation

really, it is not easy to cover all cases.

pkg/controller/master/virtualmachine/vmi_network_controller.go
ihcsim: The entire annotation map [is limited to 256KB](https://github.com/kubernetes/apimachinery/blob/59e9003f02d6f0c8fff53719a7a0604ec82ee9a9/pkg/api/validation/objectmeta.go#L63). If exceeded, K8s will reject any update attempts. We had a similar problem with [labels in the machine driver](https://github.com/harvester/docker-machine-driver-harvester/pull/69).

It's one of those "_i don't think it matters_" until an user incident comes in type of issues, because who knows what other annotations might be added to the VM.

pkg/controller/master/virtualmachine/vmi_controller.go
ihcsim: I'm not asking about picking between this or #7402. I just notice that we moved from:

```
    // only set the network mac address when it has no existing value
    if ok && vmIface.MacAddress == "" {
        logrus.Debugf("set VM %s management network %s macAddress to %s", vm.Name, vmIface.Name, macAddress)
        vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces[i].MacAddress = macAddress
    }
```

to:

```
    // skip if mac address is already set
    if nic.MacAddress != "" {
        continue
    }
```

What I want to know is other than the initial VM creation, under what circumstances will `vmIface.MacAddress == ""` or `nic.MacAddress != "" ` be `true`. Won't all subsequent user changes be made directly to the VM spec? If yes, this new annotation is only used during VM creation?

----

UPDATE: answering my own question, the controller will need to lookup this annotation again during live migration.

FrankYang0529: > (5) For cloned VM, the anntations need to be removed.

We only clone annotations to new VM in allowed list.

https://github.com/harvester/harvester/blob/f8198b619e3904ed900b794a3ee7d00751493a92/pkg/api/vm/handler.go#L59-L61

https://github.com/harvester/harvester/blob/f8198b619e3904ed900b794a3ee7d00751493a92/pkg/api/vm/handler.go#L1376-L1390

> For both case, at step (b.2) and (c.2), the VM webhook needs to figure out the removed if1, and remove it from the existing annotation.
> For case (c.3), when adding MAC to annotation, maybe we can check VM's generation and observedGeneration, only when they are same (means user does not modify the VM manually), then add/update to annotation; otherweise, skip writing annotation

@w13915984028, The statement is correct when users want to change mac address for an interface with same name. However, what if some customers want to have the same mac address for a removed and added interface? I would prefer to use a document to let users know how to change mac address for an interface with same name. It's hard to cover all cases. If we try to cover some cases in the system, it may make the system messy.

w13915984028: The current behaviour: 

When one interface is removed from VM spec from UI, all info are gone; if user adds the same name interface again, it has no MAC unless user sets it explicitly.

This PR implements it in another solution, keep compatible is better, document the difference is also ok.

FrankYang0529: > When one interface is removed from VM spec from UI, all info are gone; if user adds the same name interface again, it has no MAC unless user sets it explicitly.

It depends on whether VMI is reconciliation. Following steps can have a same mac address interface. I think this is more like an undefined behavior in our system.

1. Create a VM.
2. Create VLAN1.
3. Change VM interface to VLAN1 without restart.
4. Change VM interface back without restart.
5. Check VM. There is no mac address for the interface.
6. Add annotation to the related VMI.
7. Check VM. There is mac address for the interface.

w13915984028:
cc @rrajendran17 

The duplicated MAC address check function might be partially not working after this PR is merged. VM's dynamic MAC is stored in annotation instead of spec. Add this to https://github.com/harvester/harvester/issues/8331 for double check.

pkg/api/vm/handler.go
```diff
@@ -1527,3 +1534,76 @@ func (h *vmActionHandler) deleteResourceQuota(namespace, name string) error {
...
+		if maxSockets != 0 && input.Sockets > maxSockets {
+			return fmt.Errorf("cpu hotplug value %d is greater than the max sockets value %d", input.Sockets, maxSockets)
+		}
+
+		sockets := vm.Spec.Template.Spec.Domain.CPU.Sockets
+		if input.Sockets < sockets {
+			return fmt.Errorf("cpu hotplug value %d is less than the current sockets value %d", input.Sockets, sockets)
+		}
+
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/domain/cpu/sockets", "value": %d}`, input.Sockets))
```
w13915984028, line 1545: in case of `resourcequota` set, could it exceed? memory is similar

pkg/api/vm/handler.go
```diff
@@ -1527,3 +1534,76 @@ func (h *vmActionHandler) deleteResourceQuota(namespace, name string) error {
...
+	}
+
+	patchOps := []string{}
+	if input.Sockets != 0 {
+		maxSockets := vm.Spec.Template.Spec.Domain.CPU.MaxSockets
+		if maxSockets != 0 && input.Sockets > maxSockets {
+			return fmt.Errorf("cpu hotplug value %d is greater than the max sockets value %d", input.Sockets, maxSockets)
+		}
+
+		maxSockets = vmi.Spec.Domain.CPU.MaxSockets
```
w13915984028: if `maxSockets` is not set or `0`, then vm can expand freely, right?

pkg/api/vm/formatter.go
```diff
@@ -113,6 +114,10 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	if canMigrate(vmi) {
 		resource.AddAction(request, migrate)
 		resource.AddAction(request, findMigratableNodes)
+
+		if canCPUAndMemoryHotplug(vm) {
```
w13915984028, line 118: When a hotplug is onging, how to avoid re-trigger hotplug;  like vm migration

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -109,3 +111,48 @@ func (h *VMIController) removeDeprecatedFinalizer(_ string, vmi *kubevirtv1.Virt
 	}
 	return vmi, nil
 }
+
+// PatchMacAddresses patch mac addresses to virtual machine when virtual machine instance is deleted
```
w13915984028: This function targets to restore the saved MAC on VM's annotation to VMI's interface, then it should be put in VMI's mutator, where when VMI is created then patch them;  

when controller does it, it will race with kubevirt

w13915984028 (commented):
As the VMI's MAC is saved to VM's annotation, the better place to patch the MAC back to VMI is mutator when VMI is created, to avoid race with kubevirt. WDYT? thanks.

FrankYang0529:
> the better place to patch the MAC back to VMI is mutator when VMI is created

@w13915984028, do you mean that we only need to keep MAC address on VM annotation and don't need to put the value back to devices field?

pkg/api/vm/formatter.go
ihcsim: if hotplugging is in-progress, i think the vmi `status.currentCPUTopology` and `status.memory` will be different from the values in its `spec`.

pkg/api/vm/formatter.go
```diff
@@ -400,3 +405,18 @@ func canDismissInsufficientResourceQuota(vm *kubevirtv1.VirtualMachine) bool {
...
 }
+
+func canCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
+		return false
+	}
+
+	hasRestartRequired := false
+	for _, condition := range vm.Status.Conditions {
+		if condition.Type == kubevirtv1.VirtualMachineRestartRequired && condition.Status == corev1.ConditionTrue {
```
ihcsim, line 416: Why is this condition necessary? Doesn't look like kubevirt imposed any restrictions on hotplugging. If user made a hotplug change request while restart is required, can we let kubevirt reconcile it?

pkg/api/vm/handler.go
```diff
@@ -1527,3 +1534,76 @@ func (h *vmActionHandler) deleteResourceQuota(namespace, name string) error {
...
+	if err != nil {
+		return err
+	}
+
+	patchOps := []string{}
+	if input.Sockets != 0 {
+		maxSockets := vm.Spec.Template.Spec.Domain.CPU.MaxSockets
+		if maxSockets != 0 && input.Sockets > maxSockets {
+			return fmt.Errorf("cpu hotplug value %d is greater than the max sockets value %d", input.Sockets, maxSockets)
+		}
```
ihcsim: shouldn't we just let kubevirt handle this situation?

pkg/api/vm/handler.go
```diff
@@ -1527,3 +1534,76 @@ func (h *vmActionHandler) deleteResourceQuota(namespace, name string) error {
...
+	}
+
+	if input.Memory != "" {
+		memory, err := resource.ParseQuantity(input.Memory)
+		if err != nil {
+			return fmt.Errorf("failed to parse memory quantity: %v", err)
+		}
+
+		maxGuest := vm.Spec.Template.Spec.Domain.Memory.MaxGuest
+		if maxGuest != nil && memory.Cmp(*maxGuest) > 0 {
```
ihcsim: Similar to comment on CPU, does kubevirt not know how to handle situation like this?

ihcsim (commented):
Is the plan to move the `PatchMacAddresses()` changes in the VMI controller to https://github.com/harvester/harvester/pull/8339? That will help make the PRs easier to review, comment and unit-test, especially since the MAC address backfill changes will affect other parts of the system. It's ok to merge this, even if hotplugging doesn't fully work until #8339 is reviewed, tested and merged, right?

pkg/controller/master/virtualmachine/vmi_controller.go
ihcsim: > Yes RestartRequired is the side-effect of the original solution. 

Strictly from the hotplug perspective, can we just force-restart the VM if the condition is true, before applying the hotplug changes? I am assuming that the `RestartRequired` condition isn't always true. So maybe only a few hotplug requests from the user would lead to force-restart. IOW, if kubevirt says the VM must be restarted before hotplugging, we restart the VM. If kubevirt says no restart required, we let kubevirt do its thing.

(I am not talking about controller revision changes, drifts etc. I just want to see if there is a way to unblock hotplug live migration without coupling this PR to #8331.)

pkg/api/vm/formatter.go
FrankYang0529: From KubeVirt source code, if there is `VirtualMachineRestartRequired`, we cannot do hotplug. Users need to restart the VM to cleanup the condition.

https://github.com/kubevirt/kubevirt/blob/590ef82628023c043b9b7c6e839aae1514f05129/pkg/virt-controller/watch/vm/vm.go#L3065-L3085

pkg/api/vm/handler.go
FrankYang0529: From KubeVirt source code, its admitter only validate whether input sockets is bigger than max sockets, but it doesn't validate whether it's smaller.

https://github.com/kubevirt/kubevirt/blob/590ef82628023c043b9b7c6e839aae1514f05129/pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go#L288-L298

If users give a smaller input, the request will be handled by vm controller and it sets `VirtualMachineRestartRequired` condition. This makes users need to restart VM to enable cpu / memory again.

https://github.com/kubevirt/kubevirt/blob/590ef82628023c043b9b7c6e839aae1514f05129/pkg/virt-controller/watch/vm/vm.go#L618-L621

The case you point out is "bigger" case which is handled by KubeVirt admitter. However, I would like to handle all cases ("bigger" / "smaller") here. It's more clear.

pkg/api/vm/handler.go
FrankYang0529: Same as above.

pkg/api/vm/handler.go
FrankYang0529: No, there is default max hotplug ratio `4`. If `maxSockets` is `0`, the KubeVirt apply `sockets * ratio` to the field.

https://github.com/kubevirt/kubevirt/blob/590ef82628023c043b9b7c6e839aae1514f05129/pkg/virt-config/virt-config.go#L92

https://github.com/kubevirt/kubevirt/blob/590ef82628023c043b9b7c6e839aae1514f05129/pkg/defaults/defaults.go#L151

pkg/api/vm/handler.go
FrankYang0529: Based on KubeVirt document, it can automatically set cpu / memory resource limit to the VM if following conditions are true:

- The namespace where the VMI will be created has a ResourceQuota containing cpu limits.
- The VMI has no manually set cpu limits.
- The VMI is not requesting dedicated CPU.

https://kubevirt.io/user-guide/compute/resources_requests_and_limits/#auto-cpu-limits

I think we don't need to handle this issue in ourside.

pkg/api/vm/formatter.go
FrankYang0529: Add two condition to check whether VM can do cpu / memory hotplug.

* HotVCPUChange
* HotMemoryChange

pkg/api/vm/handler.go
ihcsim: I am still not sure if we need all these extra logic. If kubevirt rejects the admission or marks the vm as restart required, can't we just propagate the result back to the user, and let the user handle it?

I think a good UI enhancement is to notify the user that the VM need to be restarted, if the `VirtualMachineRestartRequired` condition is `true`.

FrankYang0529: I'm okay to leave this to KubeVirt. Removed related logic in handler. Thanks.

pkg/api/vm/handler.go
ihcsim: It will be a good test case for QA so that we understand the failure behaviour. K8s should handle any resource quote violation for us, but user should be notified if hotplugging failed.

ihcsim (approved):

Issue #7083, opened 2024-12-03
[BUG] Backups not appearing on second cluster using shared backup storage


himslm01:
**Describe the bug**
This issue is about two Harvester clusters sharing the same backup storage. This is important when trying to migrate virtual machines between clusters, and for testing that a backed-up virtual-machine is consistent and can be restored successfully. After all, an untested backup is no backup. 

Imagine two Harvester clusters sharing one backup storage location. I expect that soon after the backup of a VM has completed on one cluster  I should be able to select that backup on the other cluster to restore the VM onto the second cluster. This is not happening.

Looking in the Longhorn user interface in both clusters it can be seen that the backed-up volumes are visible by both clusters. It seems that Harvester is not reading the config files and refreshing its user interface.

In my experience, after a very long wait of maybe 12 hours, a very small number of VMs backed up from the "source" Harvester cluster eventually appear ready for restoration into the "destination" Harvester cluster.

**To Reproduce**
1. In the “Setting: backup-target” screen of two separate Harvester clusters, configure the same shared storage’. In my specific use case the “source” Harvester cluster is v1.3.1 and the “destination” Harvester cluster is v1.4.0 and the storage is NFS.
1. In the “Virtual Machines” screen of the “source” Harvester cluster, complete a backup of several VMs
1. In the “Backup” tab of the Longhorn user interface of the “source” Harvester cluster, see that the volumes of the VMs are listed.
1. In the “Backup” tab of the Longhorn user interface of the “destination” Harvester cluster, after a while see that the volumes of the VMs appear
1. In the “Virtual Machine Backups” screen of the “destination” Harvester cluster, note that the backups are not listed

**Expected behavior**
I expect to see the VMs backed up from one Harvester cluster into storage shared by multiple Harvester clusters to be restorable onto any of the Harvester clusters in a timely manor.

**Environment**
- Harvester ISO version:
   - “source” Harvester cluster is v1.3.1
   - “destination” Harvester cluster is v1.4.0 
 - Underlying Infrastructure: Bare metal nodes and NFS storage.

**Additional context**

VM Backups screen on "source" Harvester cluster (highliting the only VM that is available in the "destination" Harvester cluster)
![image](https://github.com/user-attachments/assets/fcedaa84-dc5c-4fd3-b6c8-7a687971d7b7)

VM Backups screen on "destination" Harvester cluster
![image](https://github.com/user-attachments/assets/a56afcbc-ac45-42bf-9069-f98a296ad552)

[...]

PR #7114, opened 2024-12-06
feat: add refreshIntervalInSeconds to backup target

FrankYang0529:

**Problem:**
Currently, the system can sync data from backup target to the cluster when we setup backup target setting first time. It will not get any new data from backup target after it.

**Solution:**
Add a new field `refreshIntervalInSeconds` to backup target setting, so users can control when to get data from backup target to the cluster. The value `0` disables the feature.

**Related Issue:**
https://github.com/harvester/harvester/issues/7083

**Test plan:**
1. Setup two cluster - `cluster1` and `cluster2`.
2. Setup the same backup target on each cluster.
3. On `cluster2`, set `refreshIntervalInSeconds` as `60` seconds.
4. Create a VMImage, VM, and VMBackup on the `cluster1`.
5. After the data is on the backup target, wait around 1 minute. We can see the data is synced to `cluster2`.
6. Create a new VM and VMBakcup on the `cluster2`.
7.  After the data is on the backup target, wait around 1 minuter. The data is not in `cluster1`, because default `refreshIntervalInSeconds` is `0` which disables the feature.
8. Change `refreshIntervalInSeconds` from `0` to `60` on the `cluster1`. We can see new VMBackup is on the `cluster1`.

### DIFF
--- a/pkg/controller/master/backup/backup_metadata.go
+++ b/pkg/controller/master/backup/backup_metadata.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"reflect"
 	"time"
 
 	"github.com/longhorn/backupstore"
@@ -115,10 +116,12 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
 		return setting, err
 	}
 
-	logrus.Debugf("backup target change, sync vm backup:%s:%s", target.Type, target.Endpoint)
-
 	// when backup target is reset to default, do not trig sync
 	if target.IsDefaultBackupTarget() {
+		return nil, nil
+	}
+
+	if !h.shouldRefresh(setting, target.RefreshIntervalInSeconds) {
 		return nil, nil
 	}
 
@@ -140,7 +143,47 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
 		return nil, nil
 	}
 
-	return nil, nil
+	return h.renewBackupTarget(setting)
+}
+
+func (h *MetadataHandler) shouldRefresh(setting *harvesterv1.Setting, refreshIntervalInSeconds int64) bool {
+	if setting.Annotations == nil {
+		return true
+	}
+
+	var err error
+	lastTime := time.Unix(0, 0)
+	currentTime := time.Now()
+	if setting.Annotations[util.AnnotationLastRefreshTime] != "" {
+		lastTime, err = time.Parse(time.RFC3339, setting.Annotations[util.AnnotationLastRefreshTime])
+		if err != nil {
+			logrus.WithError(err).Errorf("failed to parse last refresh time")
+			lastTime = time.Unix(0, 0)
+			return true
+		}
+	}
+	if getBackupTargetHash(setting.Value) == setting.Annotations[util.AnnotationHash] &&
+		(refreshIntervalInSeconds == 0 ||
+			currentTime.Sub(lastTime).Seconds() < float64(refreshIntervalInSeconds)) {
+		h.settings.EnqueueAfter(setting.Name, lastTime.Add(time.Duration(refreshIntervalInSeconds)*time.Second).Sub(currentTime))
+		return false
+	}
+	return true
+}
+
+func (h *MetadataHandler) renewBackupTarget(setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	settingCopy := setting.DeepCopy()
+	if settingCopy.Annotations == nil {
+		settingCopy.Annotations = map[string]string{}
+	}
+
+	settingCopy.Annotations[util.AnnotationHash] = getBackupTargetHash(setting.Value)
+	settingCopy.Annotations[util.AnnotationLastRefreshTime] = time.Now().Format(time.RFC3339)
+	if !reflect.DeepEqual(setting, settingCopy) {
+		return h.settings.Update(settingCopy)
+	}
+	h.settings.Enqueue(setting.Name)
+	return setting, nil
 }
 
 func (h *MetadataHandler) syncVMImage(target *settings.BackupTarget) error {
--- a/pkg/controller/master/backup/backup_target.go
+++ b/pkg/controller/master/backup/backup_target.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
@@ -26,6 +27,10 @@ import (
 
 const (
 	backupTargetControllerName = "harvester-backup-target-controller"
+
+	longhornBackupTargetSettingName            = "backup-target"
+	longhornBackupTargetSecretSettingName      = "backup-target-credential-secret"
+	longhornBackupstorePollIntervalSettingName = "backupstore-poll-interval"
 )
 
 // RegisterBackupTarget register the setting controller and reconsile longhorn setting when backup target changed
@@ -95,7 +100,9 @@ func (h *TargetHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Sett
 			return h.setConfiguredCondition(setting, "", err)
 		}
 
-		return h.reUpdateBackupTargetSettingSecret(setting, target)
+		if setting, err = h.reUpdateBackupTargetSettingSecret(setting, target); err != nil {
+			return h.setConfiguredCondition(setting, "", err)
+		}
 
 	case settings.NFSBackupType:
 		if err = h.updateLonghornTarget(target); err != nil {
@@ -127,6 +134,12 @@ func (h *TargetHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Sett
 		}
 
 		return h.setConfiguredCondition(setting, "", fmt.Errorf("Invalid backup target type:%s or parameter", target.Type))
+	}
+
+	if target.RefreshIntervalInSeconds > 0 {
+		if err = h.updateLonghornPollIntervalSetting(target); err != nil {
+			return h.setConfiguredCondition(setting, "", err)
+		}
 	}
 
 	if len(setting.Status.Conditions) == 0 || harvesterv1.SettingConfigured.IsFalse(setting) {
@@ -285,4 +298,33 @@ func (h *TargetHandler) setConfiguredCondition(setting *harvesterv1.Setting, rea
 	// SetError with nil error will cleanup message in condition and set the status to true
 	harvesterv1.SettingConfigured.SetError(settingCpy, reason, err)
 	return h.settings.Update(settingCpy)
+}
+
+func (h *TargetHandler) updateLonghornPollIntervalSetting(backupTarget *settings.BackupTarget) error {
+	setting, err := h.longhornSettingCache.Get(util.LonghornSystemNamespaceName, longhornBackupstorePollIntervalSettingName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+
+		if _, err := h.longhornSettings.Create(&lhv1beta2.Setting{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      longhornBackupstorePollIntervalSettingName,
+				Namespace: util.LonghornSystemNamespaceName,
+			},
+			Value: strconv.FormatInt(backupTarget.RefreshIntervalInSeconds, 10),
+		}); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	settingCpy := setting.DeepCopy()
+	settingCpy.Value = strconv.FormatInt(backupTarget.RefreshIntervalInSeconds, 10)
+
+	if !reflect.DeepEqual(setting, settingCpy) {
+		_, err := h.longhornSettings.Update(settingCpy)
+		return err
+	}
+	return nil
 }
--- a/pkg/controller/master/backup/util.go
+++ b/pkg/controller/master/backup/util.go
@@ -1,9 +1,12 @@
 package backup
 
 import (
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"io"
 	"path/filepath"
+	"strings"
 	"time"
 
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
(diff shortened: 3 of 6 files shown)


bk201:
One unclear topic to me if Longhorn also syncs backups regularly, will this cause any issues?

innobead:
Can I also regard this as a part of the DR story?

ChanYiLin:
@bk201 
Longhorn syncs the backup target and creates the backup resources, e.g. `BackupVolume` and `BackupBackingImage`
Then the corresponding controllers' will keep updating the information from the backup target to the resources' status. 
I am not sure how Harvester handles `VMBackup`
But for `VMImage`, as I discussed with @FrankYang0529 , the VMImage will directly restore the BackingImage from the backup target, and that shouldn't have any problem.

That is, Longhorn syncs the backup target and manages the backup related CRs.
And those CRs are only the reflection(pointer) to the data in the backup target.
It won't have any race condition.

FrankYang0529:
> I am not sure how Harvester handles VMBackup

For VMBackup, we check whether related storage class is on the cluster first, so it will wait until VMImage is created and then resync back to the cluster.

https://github.com/harvester/harvester/blob/ef97c6c4fa49086a082c12682ef5287d4d3777f5/pkg/controller/master/backup/backup_metadata.go#L277-L289

FrankYang0529:
> Can I also regard this as a part of the DR story?

Yes, it's kind of DR for the cluster.

innobead:
ref: https://github.com/harvester/harvester/issues/1850

himslm01:
I think it's great to have an automatic refresh, but would it be possible for there to be a user accessible "Refresh" button and/or an API to call that just refreshes the list "now". An API could be used to automate / code the transfer of a VM from one cluster to another.

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -115,13 +119,34 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
...
+			lastTime = time.Unix(0, 0)
+		}
+	}
+	if currentHash == setting.Annotations[util.AnnotationHash] &&
+		(target.RefreshIntervalInSeconds == 0 ||
+			currentTime.Sub(lastTime).Seconds() < float64(target.RefreshIntervalInSeconds)) {
+		h.settings.EnqueueAfter(setting.Name, lastTime.Add(time.Duration(target.RefreshIntervalInSeconds)*time.Second).Sub(currentTime))
+		return nil, nil
+	}
+
```
WebberHuang1118: My two cents is to factor out the refresh interval checking as a new function, WDYT? @FrankYang0529

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -140,6 +165,13 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
 		return nil, nil
 	}
 
+	settingCopy := setting.DeepCopy()
+	settingCopy.Annotations[util.AnnotationHash] = currentHash
+	settingCopy.Annotations[util.AnnotationLastRefreshTime] = time.Now().Format(time.RFC3339)
+	if !reflect.DeepEqual(setting, settingCopy) {
+		return h.settings.Update(settingCopy)
+	}
```
WebberHuang1118: Ditto

WebberHuang1118 (commented):
LGTM in general, please help to check some questions, thanks.

pkg/controller/master/backup/backup_target.go
```diff
@@ -127,6 +128,12 @@ func (h *TargetHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Sett
 		return h.setConfiguredCondition(setting, "", fmt.Errorf("Invalid backup target type:%s or parameter", target.Type))
 	}
 
+	if target.RefreshIntervalInSeconds > 0 {
+		if err = h.updateLonghornPollIntervalSetting(target); err != nil {
+			return h.setConfiguredCondition(setting, "", err)
+		}
+	}
+
```
WebberHuang1118, line 144: From my code reading, the poll interval will not be updated if the [target type is S3](https://github.com/FrankYang0529/harvester/blob/bae83df24f6b37fcfd4c90b9ca2052aa71c93475/pkg/controller/master/backup/backup_target.go#L81-L97) ?

WebberHuang1118:
> I think it's great to have an automatic refresh, but would it be possible for there to be a user accessible "Refresh" button and/or an API to call that just refreshes the list "now". An API could be used to automate / code the transfer of a VM from one cluster to another.

Hi @himslm01,
IIUC, LH doesn't support such a scenario. So even if this case is implemented on the Harvester side, it still needs the support for LH resources (e.q. backupvolume, backup, backup, backupbackingimage ... etc).

FrankYang0529:
Hi @WebberHuang1118, I addressed all comments. Could you take a look when you have time? Thanks.

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -140,7 +143,46 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
...
+
+	var err error
+	lastTime := time.Unix(0, 0)
+	currentTime := time.Now()
+	if setting.Annotations[util.AnnotationLastRefreshTime] != "" {
+		lastTime, err = time.Parse(time.RFC3339, setting.Annotations[util.AnnotationLastRefreshTime])
+		if err != nil {
+			logrus.WithError(err).Errorf("failed to parse last refresh time")
+			lastTime = time.Unix(0, 0)
```
WebberHuang1118, line 161: Should we return `true` here?

pkg/controller/master/backup/backup_target.go
```diff
@@ -309,3 +318,32 @@ func (h *TargetHandler) setConfiguredCondition(setting *harvesterv1.Setting, rea
...
+		if _, err := h.longhornSettings.Create(&lhv1beta2.Setting{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      longhornBackupTargetSettingName,
+				Namespace: util.LonghornSystemNamespaceName,
+			},
+			Value: strconv.FormatInt(backupTarget.RefreshIntervalInSeconds, 10),
+		}); err != nil {
+			return err
+		}
```
WebberHuang1118: Should we directly return the error, if the LH backup setting doesn't exist? Since we already check the setting existence in [updateLonghornTarget()](https://github.com/FrankYang0529/harvester/blob/dda142f109791c1b6d28a101d173504166146013/pkg/controller/master/backup/backup_target.go#L165-L192)

WebberHuang1118 (commented):
LGTM, just some minor points, thanks.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: I think it's covered by the last line in this function. Let take a look for few cases:

* A new backup target: it doesn't have `"harvesterhci.io/lastRefreshTime"` annotation, so it doesn't get into this block.
* An updated backup target: it may have `"harvesterhci.io/lastRefreshTime"` annotation which is from an old backup target. In the next block, it doesn't have a same hash, so we will return true.
* A same backup target with broken time format: it has same hash, so it passes the first condition in the next block. However, it fails on the second condition, because `currentTime - 0 (lastTime)` is bigger than a sensible `refreshIntervalInSeconds`.

pkg/controller/master/backup/backup_target.go
FrankYang0529: I used a wrong name here. I would like to create `backupstore-poll-interval` if it's not existent. Fixed it. Thanks.

pkg/controller/master/backup/backup_metadata.go
WebberHuang1118: I mean if the controller has the time parsing error at
```
	if setting.Annotations[util.AnnotationLastRefreshTime] != "" {
		lastTime, err = time.Parse(time.RFC3339, setting.Annotations[util.AnnotationLastRefreshTime])
		if err != nil {
			logrus.WithError(err).Errorf("failed to parse last refresh time")
			lastTime = time.Unix(0, 0)
		}
	}
```
maybe we can have a direct return

FrankYang0529: Updated it.

WebberHuang1118:
> > Can I also regard this as a part of the DR story?
> 
> Yes, it's kind of DR for the cluster.

IIUC, it could be leveraged with the [VM backup schedule](https://docs.harvesterhci.io/v1.4/vm/backup-restore#scheduling-virtual-machine-backups-and-snapshots) to achieve cross-cluster VM DR.

WebberHuang1118 (approved):

bk201 (approved):

FrankYang0529:
@mergify backport v1.4 v1.5

pkg/controller/master/backup/backup_target.go
```diff
@@ -129,6 +136,12 @@ func (h *TargetHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Sett
 		return h.setConfiguredCondition(setting, "", fmt.Errorf("Invalid backup target type:%s or parameter", target.Type))
 	}
 
+	if target.RefreshIntervalInSeconds > 0 {
```
w13915984028, line 139: When `0` is a valid value of `RefreshIntervalInSeconds`, why it only update to LH when `RefreshIntervalInSeconds>0`,

if user sets the `RefreshIntervalInSeconds` to 10, then set back to 0, what will happen?

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -140,7 +143,47 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
...
```
w13915984028, line 161: when return true, `lastTime = time.Unix(0, 0)` seems redundant

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -140,7 +143,47 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
...
+			return true
+		}
+	}
+	if getBackupTargetHash(setting.Value) == setting.Annotations[util.AnnotationHash] &&
+		(refreshIntervalInSeconds == 0 ||
```
w13915984028, line 166: did some one notice that, when refreshIntervalInSeconds ==0,

h.settings.EnqueueAfter(setting.Name, lastTime.Add(time.Duration(refreshIntervalInSeconds)*time.Second).Sub(currentTime))

What value will be used to EnqueueAfter? a negative value?

Looks like the `refreshIntervalInSeconds == 0` should be processed indepedently, instead of call `EnqueueAfter`.

w13915984028 (commented):
Sorry for the later review, but I do have some questions, thanks.

pkg/controller/master/backup/backup_target.go
```diff
@@ -286,3 +299,32 @@ func (h *TargetHandler) setConfiguredCondition(setting *harvesterv1.Setting, rea
...
+
+	settingCpy := setting.DeepCopy()
+	settingCpy.Value = strconv.FormatInt(backupTarget.RefreshIntervalInSeconds, 10)
+
+	if !reflect.DeepEqual(setting, settingCpy) {
+		_, err := h.longhornSettings.Update(settingCpy)
+		return err
+	}
+	return nil
+}
```
WebberHuang1118, line 330: This LH setting is not valid since `LH v1.8.0`, we should update the poll interval in a similar way as https://github.com/harvester/harvester/pull/7536

FrankYang0529: Create PR for it https://github.com/harvester/harvester/pull/7684.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: Removed in https://github.com/harvester/harvester/pull/7684.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: Yes, good catch. Updated in https://github.com/harvester/harvester/pull/7684.

pkg/controller/master/backup/backup_target.go
FrankYang0529: The default poll interval in LH is 5 minutes. We update it when `RefreshIntervalInSeconds>0` to avoid that there is data in Harvester, but no data in LH.

> if user sets the RefreshIntervalInSeconds to 10, then set back to 0, what will happen?

It's ok. LH will keep refresh. However, in Harvester, it doesn't refresh backup target next time.

votdev:
Failed to fully verify this issue as solved.

### Setup
* Environment: **ipxe-examples/vagrant-pxe-harvester 1 node + one separate Vagrant box running Harvester**
* Harvester Version: **v1.5.0-rc1**

### Verify Steps
1. 🟢 Setup two cluster - `cluster1` and `cluster2`.
2. 🟢 Setup the same backup target on each cluster.
3. 🟢 On `cluster2`, set `refreshIntervalInSeconds` as `60` seconds.
4. 🟢 Create a VMImage, VM, and VMBackup on the `cluster1`.
5. 🟢 After the data is on the backup target, wait around 1 minute. We can see the data is synced to `cluster2`.
6. 🟢 Create a new VM and VMBakcup on the `cluster2`.
7. 🟢 After the data is on the backup target, wait around 1 minuter. The data is not in `cluster1`, because default `refreshIntervalInSeconds` is `0` which disables the feature.
8. 🔴 Change `refreshIntervalInSeconds` from `0` to `60` on the `cluster1`. We can see new VMBackup is on the `cluster1`.
```
time="2025-03-20T12:00:56Z" level=info msg="Loaded driver for nfs://192.168.0.1:/shared/" pkg=nfs
time="2025-03-20T12:00:56Z" level=warning msg="skip create vm image, because there is already an image with the same display name" displayName=cirros-0.6.2-x86_64-disk.img name=image-9nfbt namespace=default
time="2025-03-20T12:00:56Z" level=info msg="start syncing vm backup metadata..." target.endpoint="nfs://192.168.0.1:/shared/" target.type=nfs
time="2025-03-20T12:00:56Z" level=warning msg="Trying reading mount point /var/lib/longhorn-backupstore-mounts/192_168_0_1/shared to make sure it is healthy" pkg=nfs
time="2025-03-20T12:00:56Z" level=info msg="Loaded driver for nfs://192.168.0.1:/shared/" pkg=nfs
time="2025-03-20T12:00:56Z" level=warning msg="skip creating vm backup, because the storage class is not found" error="storageclasses.storage.k8s.io \"longhorn-image-9nfbt\" not found" name=vm-cirros-backup namespace=default storageClasss=longhorn-image-9nfbt
time="2025-03-20T12:00:56Z" level=warning msg="Trying reading mount point /var/lib/longhorn-backupstore-mounts/192_168_0_1/shared to make sure it is healthy" pkg=nfs
time="2025-03-20T12:00:56Z" level=info msg="Loaded driver for nfs://192.168.0.1:/shared/" pkg=nfs
``

### Conclusion
The sync is aborted because the used image already exists on the `cluster1`. The problem can already appear at step 5 if the image already exists on `cluster2`.
The VM backup sync can be continued and successfully finished when the already existing `Image` is deleted.

Issue #6256, opened 2024-08-04
[ENHANCEMENT] Provide User Info When Storage Network Is Configured Incorrectly 


irishgordo:
### Is your enhancement related to a problem? Please describe.
A user could configure a storage-network, but behind the scenes the `instance-manager-uuid` pod could be failing but that error is silent in the web-ui/dashboard/ui section.

### Describe the solution you'd like
Something that polls the new creation of instance-manager pod, looks at it's status and takes note of failures or rpc codes.

### Describe alternatives you've considered
Maybe a warning label with the storage-network creation screen to encourage users to look at the `instance-manager` in the `longhorn-system`.

### Additional Context:

**Pre-Reqs:**
Given a Network exists:
```
192.168.205.1/24
```
With VLAN 205
A DHCP Server operating from 192.168.205.6 - 192.168.205.254

Two NICs set up in Balance-ALB, on the Network Config, used in the storage network.


**Scenario A:**
```
{
  "vlan": 205,
  "clusterNetwork": "storage",
  "range": "192.168.205.0/32",
  "exclude": [
    "192.168.205.1/32",
    "192.168.205.2/32",
    "192.168.205.0/32",
    "192.168.205.3/32",
    "192.168.205.4/32",
    "192.168.205.5/32",
    "192.168.205.255/32"
  ]
}
```
As from the `instance-manager` pod, we'll note Errors / Failures of:
```
  Normal   AddedInterface          49s                multus   Add eth0 [10.52.0.246/32] from k8s-pod-network                                                                                                                                                                                                                                                                                                                         
  Warning  FailedCreatePodSandBox  48s                kubelet  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "81d2bd2fb23b3eafa33dda98f0d684eaecf75d28df4cf14054b8346d8a59d4dd": plugin type="multus" name="multus-cni-network" failed (add): [longhorn-system/backing-image-manager-c16a-46e0/f92b252f-7434-439e-8487-6d92237e2299:storagenetwork-zgmjd]: error adding container
 to network "storagenetwork-zgmjd": error at storage engine: net mask is too short, subnet {192.168.205.0 ffffffff} has no usable IP addresses, it is too small                                                                                                                                                                                                                                                                       
  Normal   AddedInterface          47s                multus   Add eth0 [10.52.0.248/32] from k8s-pod-network                                                                                                                                                                                                                                                                                                                         
[...]

PR #7126, opened 2024-12-09
6256 - Enhance storage network error handling

rrajendran17:
**Problem:**
During storage network configuration, a secondary network is created on LH pods like backing-image,instance-manager etc.
During this, if the IP address range provided by the user is not valid or no allocatable ip addresses derived,then LH pods are not assigned any ip address and they will not be in running status.User is not aware of this wrong config as the current code does not validate this or notify user.

**Solution:**
This PR takes care of the following.
1.After storage network post config, the controller logic checks if the whereabouts ip pools allocated count >= the number of allocatable addresses calculated.If its not >=, then a error msg is displayed on the storage network config settings
as in the image below.

original formula for min allocatable address is, 
Required Number of IPs = (Number of Nodes * 2) + (Number of nodes * Number of Disks * 2) + Number of Images to Download/Upload
which is changed to,
min allocatable address = Number of Nodes + Number of Nodes * Number of Disks 

since * 2 is for upgrade case and  Number of Images to Download/Upload is dynamic entity.

Attaching image of UI error when LH pods are not configured and in error state
wcni

2.The harvester-webhook validates for invalid ip addresses range considering both include and exclude range of ip addresses configured by user and throws error.This prevents user from configuring wrong subnet network range.
The validation checks if there is number of ips derived from user config is >= min allocatable ip addresses calculated using the formula as below.

Required Number of IPs = (Number of Nodes * 2) + (Number of nodes * Number of Disks * 2)

Number of Images to Download/Upload is skipped as it is dynamic entity.

**Related Issue:**
https://github.com/harvester/harvester/issues/6256

**Test plan:**
1.Configure storage network with following
   a.valid vlan id
   b.cluster network selected

case a:
Include - 192.168.2.0/32

Result : "no allocatable IP addresses, check IP range for include and exclude in storage network config"

case b:
Include - 192.168.2.0/31

Result : "no allocatable IP addresses, check IP range for include and exclude in storage network config"

case c:
Include - 192.168.2.0/30
Exclude - 192.168.2.0/31

Result : "no allocatable IP addresses, check IP range for include and exclude in storage network config"

case d:
Include - 192.168.2.0/24
Exclude - 192.168.2.9/24

Result : "no allocatable IP addresses, check IP range for include and exclude in storage network config"

case e:
Include - 192.168.2.0/24
Exclude - 192.168.2.0/28

Result : success

case f:
Include - 192.168.2.0/28

Exclude - 192.168.2.229/30
Exclude - 192.168.2.236/31

Result - webhook error as exclude range does not overlap include range

case g:
3 node cluster
minimum required IPs - 12

Include - 192.168.2.0/29
Exclude -  192.168.2.0/31

Result: webhook error as ip address range is < 12

Note:This PR does not handle dynamic change in entities after storage network configuration.
Example : new node added to the cluster, additional disks added (backing-image-ds pod created later)

Note for testing :The webhook validates and prevents user from configuring the invalid ranges,
                          So for testing harveser changes, the allocation failure has to be induced manually.


pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -397,6 +407,58 @@ func (h *Handler) removeOldNad(setting *harvesterv1.Setting) error {
 	return nil
 }
 
+// check if LH Pod backing-image,backing-image-ds, instance-manager status after configuring storage network
+func (h *Handler) checkLHPodStatus() error {
```
w13915984028: It seems not the best idea to check LH related PODs, which will enhance the coupling.

The SN is using Whereabouts to allocate IPs to LH PODs, could we check it, if none IP is allocated after certain time;  and if IPs are exhausted, add related condition/status as well.

pkg/webhook/resources/setting/validator.go
```diff
@@ -1306,6 +1307,29 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
 	if !network.IP.Equal(ip) {
 		return fmt.Errorf("Range should be subnet CIDR %v", network)
 	}
+
+	var ipAddrListExclude []string
+	ipAddrListInclude := webhookutil.GetUsableIPAddressesInclude(network)
+	//Minimum ips required is 2.
+	//single node with default LH pods(backing-image,instance-manager)
+	if len(ipAddrListInclude) < 2 {
+		return fmt.Errorf("no allocatable IP addresses, check IP range for include and exclude in storage network config")
```
w13915984028: ..only %v IPs are avaiable and they are less than the minimal %v IPs...

w13915984028 (commented):
two questions, thanks.

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: @w13915984028 Thanks for reviewing.
Can you please elaborate on "enhance the coupling" problem ?
when you say "check none IP allocated after sometime", is it to check on the pod or is it some other way ? 

Currently whereabouts does the allocatable ip addresses internally which is not known to the user unless it tries that on a pod and pod fails.

pkg/webhook/resources/setting/validator.go
rrajendran17: This check is basically to avoid conditions where user can provide /32 or /31 range ip address where /32 provides no allocatable addresses and /31 provides only 1 ip address.
(network and broadcast addresses are not included in allocatable address range by whereabouts)

pkg/controller/master/storagenetwork/storage_network.go
w13915984028: 1. Harvester does not need to couple the detailed depolyments of LH related PODS, and LH ever changed  at about v1.4.0->v1.5.0 or v1.5.0->v1.6.0;

2. The following object is used to handle out IPs to LH PODs, none allocations means something is wrong, IP exhausting as well.

Whereabouts is mentioned in the document https://docs.harvesterhci.io/v1.3/advanced/storagenetwork/#prerequisites

```
- apiVersion: whereabouts.cni.cncf.io/v1alpha1
  kind: IPPool
  metadata:
...
    name: 192.168.14.0-23
    namespace: kube-system
    resourceVersion: "91596"
    uid: ce519d61-ea63-40aa-a5fd-4e256d48f6ee
  spec:
    allocations:
      "3":
        id: 1e8de46994c3d9ef870da8e1a8caac0091ec0f1336670beb9a2be65a659a1262
        podref: longhorn-system/instance-manager-8949ce7090059632da2fd9d22c7e3300
      "4":
        id: dcbdc52e1a2e8c2a04d6cb110997d28badb7561dc740787a7a993e3a308d4a9f
        podref: longhorn-system/backing-image-manager-1e05-a178
    range: 192.168.14.0/23
```

pkg/webhook/resources/setting/validator.go
w13915984028: We may define the min as e.g. 2 or 16, and the log message shows clearly this error.

Min value 2 only makes sense for test; a roll-upgrade can also break single-node. We may define it as 8 or 16.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -55,6 +60,7 @@ const (
 	MsgStopPod               = "Stopping Pods"
 	MsgWaitForVolumes        = "Waiting for all volumes detached: %s"
 	MsgUpdateLonghornSetting = "Update Longhorn setting"
+	MsgLHPodStatus           = "Longhorn Pods not running,Check Storage network configuration for IP addr Range"
```
ihcsim: Given the ephemeral nature of pods, I think it's possible for the LH pods to not run due to reasons other than storage network IP range misconfiguration.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -411,6 +473,20 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
...
+		if _, updateConditionErr := h.setConfiguredCondition(setting, false, ReasonInProgress, MsgLHPodStatus); updateConditionErr != nil {
+			return setting, fmt.Errorf("update status error %v", updateConditionErr)
+		}
+		return setting, err
+	}
+
+	msg := harvesterv1.SettingConfigured.GetMessage(settingCopy)
+
+	//Requeue storage network config after a specified period to verify if LH pods are running
+	if msg == MsgUpdateLonghornSetting {
```
ihcsim: Is this meant to be `msg == MsgLHPodStatus`?

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -411,6 +473,20 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
...
+		h.settingsController.EnqueueAfter(setting.Name, PodStatusTime)
```
ihcsim: Shouldn't the enqueuing happen only if `h.checkLHPodStatus()` returns a non-nil error?

pkg/webhook/resources/setting/validator.go
```diff
@@ -1306,6 +1307,29 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
 	if !network.IP.Equal(ip) {
 		return fmt.Errorf("Range should be subnet CIDR %v", network)
 	}
+
+	var ipAddrListExclude []string
+	ipAddrListInclude := webhookutil.GetUsableIPAddressesInclude(network)
```
ihcsim: Is it possible to do the include/exclude XOR in one function? E.g., put `config.Exclude` into a map, then as you iterate through the `include` list, just exclude those in the exclude map including the broadcast address?

pkg/webhook/util/ip.go
```diff
@@ -0,0 +1,83 @@
+package util
```
ihcsim, line 1: What do you think of adding unit tests for some of these functions?

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: This check is needed to not repeat enqueue repeatedly, instead do only during msg MsgUpdateLonghornSetting which would be updated only after longhorn setting update, the enqueue happens after a specified period of time during which the LH pods would have been updated with new storage network.

If LH pods are not running/error returned, the controller logic takes care of re configuring it again and again until its success.So enqueue is done only once after storage network change LH settings update.

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: The LH pod status cannot be verified immediately as they will be restarted with storage network config after specified time.So enqueue after 10 seconds takes care of re triggering the controller logic after 10 seconds , during which LH pods status will be verified, if successful, its done, else the error returned from checkLHPodStatus() will make sure its checked repeatedly inside storage network controller logic

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: As we discussed, will see if there is a way to get the status from whereabouts CNI, so that we could skip checking LH pod status

rrajendran17: one thing I notice about wherabout ippools is, it does have only allocations list (not any other msg regarding failures), so there could be cases where it would have allocated partially, so getting a node count and if backing-image-ds is present,will make sure number of IPs required in the cluster and this can be compared to the number of ips allocated by whereabouts ippool to know if all IPs are allocated successfully.

Pros:
1.Use whereabouts ip allocation status instead of longhorn pod status.
2.IP count needed could be derived from above and validated against number of IPs available from the range given by user instead of validating against some minimum value.

Cons:
1.Makes the code little complex.

ihcsim: Yeah, this does sound a bit complicated and potentially brittle. Overall, is it the storage network controller's responsibility to watch for LH pods' health? The pods can fail for different reasons, potentially, not even related to storage network change. Ideally, degrading pods' health should be reported by the monitoring system, not a `setting` controller.

I think the webhook validation makes sense, but not sure about the LH pods' health and UI reporting part. Is it common for us to report errors in the setting UI? From an UX perspective, it's likely that the errors won't show up right away, right? If user navigated away from the settings page, they won't see any later errors.

pkg/webhook/util/ip.go
rrajendran17: Added unit tests

pkg/webhook/resources/setting/validator.go
rrajendran17: done

pkg/webhook/resources/setting/validator.go
```diff
@@ -64,6 +65,7 @@ const (
 	labelAppNameValueGrafana          = "grafana"
 	labelAppNameValueImportController = "harvester-vm-import-controller"
 	maxTTLDurationMinutes             = 52560000 //specifies max duration allowed for kubeconfig TTL setting, and corresponds to 100 years
+	MinallocatableIPAddrs             = 16
```
ihcsim:
```suggestion
	minAllocatableIPAddrs             = 16
```
Nit: use camel case for consistency and it doesn't look like this variable needs to be exported.

pkg/webhook/util/ip.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func GetUsableIPAddressesCount(includeRange string, excludeRange []string) (count int, err error) {
+	var includeRangeList []string
+	includeRangeList = append(includeRangeList, includeRange)
+
+	includeIPAddrList, err := getIPAddressesFromSubnet(includeRangeList, true)
+	if err != nil {
+		return count, err
+	}
```
ihcsim, line 27: I think the `net/netip` package offers a `Prefix.Bits()` function which we can utilize to calculate the number of included IP addresses without actually computing the actual IP addresses. Then we can still iterate the excluded IPs and use `Prefix.Contains(excludeIP)` to compute `count`.

In general, `net/netip` is more efficient than `net`.

pkg/webhook/util/ip.go
```diff
@@ -0,0 +1,77 @@
...
+		_, network, err := net.ParseCIDR(ipNetSubnet)
+		if err != nil {
+			return ipAddrList, err
+		}
+
+		// Get broadcast address (last address in the subnet)
+		broadcast := getBroadcastAddress(network)
+
+		// Iterate through all the IP addresses in the subnet
+		for ip := network.IP.Mask(network.Mask); network.Contains(ip); incrementIP(ip) {
```
ihcsim: Is `network.IP.Mask(network.Mask)` the same as the `ip` returned by `net.ParseCIDR()`?

pkg/webhook/util/ip.go
```diff
@@ -0,0 +1,77 @@
+package util
+
+import (
+	"net"
+)
+
+// incrementIP increments the IP address by 1
+func incrementIP(ip net.IP) net.IP {
```
ihcsim, line 8: The `net/netip` package offers the [`Addr.Next()` function](https://pkg.go.dev/net/netip#Addr.Next). In general `net/netip` is more efficient than `net`.

pkg/webhook/util/ip.go
```diff
@@ -0,0 +1,77 @@
...
+		if err != nil {
+			return ipAddrList, err
+		}
+
+		// Get broadcast address (last address in the subnet)
+		broadcast := getBroadcastAddress(network)
+
+		// Iterate through all the IP addresses in the subnet
+		for ip := network.IP.Mask(network.Mask); network.Contains(ip); incrementIP(ip) {
+			if include && (ip.Equal(network.IP) || ip.Equal(broadcast)) {
```
ihcsim, line 57: Do we need to check for `broadcast` in the `exclude` range?

pkg/webhook/resources/setting/validator.go
ihcsim: Thanks for the explanation on where the `16` comes from. Let me know when the new changes are ready.

pkg/webhook/util/ip.go
ihcsim: Sorry, I misunderstood this part. I guess do we really need to compute all the addresses within the include and exclude ranges to get the count? Can we just use the suffix bits to calculate the number of IPs in both lists, make sure they [`overlap`](https://pkg.go.dev/net/netip#Prefix.Overlaps), and count how many are usable?

pkg/webhook/util/ip.go
```diff
@@ -0,0 +1,77 @@
...
+	ipAddrList = make(map[string]struct{})
+
+	for _, ipNetSubnet := range ipNetSubnets {
+		_, network, err := net.ParseCIDR(ipNetSubnet)
+		if err != nil {
+			return ipAddrList, err
+		}
+
+		// Get broadcast address (last address in the subnet)
+		broadcast := getBroadcastAddress(network)
```
ihcsim, line 53: Do we have to do this ourselves or can user specify it in the exclude list, and we make this a documentation item?

rrajendran17:
summarizing here, the options available to provide status of storage network configuration and the
approach that will be taken.Thanks

1.Checking the status of LH pods to be running after the storage network configuration.
   pros : accurate because if storage network configuration failed, LH will not be in running state.
   cons: not the best one to include this in controller logic.

2.Check the IP pool whereabouts allocations list, if the list is empty, allocations failed.
   pros : trivial and common failures seen in customer setups for allocation failures.
   cons: does not handle partial allocations failure.

3.Check the IP pool whereabouts allocations list, compare this with the number of ip addresses required.
   If the number of allocations not same as number of ip addresses required, then there is an allocation failure.
  Pros: accurate
  cons : non trivial and number of nodes and number of disks available must be calculated.

Considering the above options with pros and cons, it is best to go with option 3.
cc @w13915984028 @ihcsim @starbops @mingshuoqiu @WebberHuang1118 @bk201

pkg/webhook/util/ip.go
rrajendran17: excludedIPs is again a list of subnet IPs(net.IPNet) and not single(net.IP) IP.so we cannot use Prefix.Contains(excludeIP)

pkg/webhook/util/ip.go
rrajendran17: true, yeah I could use the output ip from net.ParseCIDR(), will change it, thanks

pkg/webhook/util/ip.go
rrajendran17: we need the list of IPs to check overlap, so just calculating the number of ips will not be enough.
But instead of adding counting logic, I could use overlap function as you suggested.

pkg/webhook/util/ip.go
rrajendran17: users just specify subnet/IPNet in include, and wherabouts just calcuate the range excluding network and broadcast ip
for usable ip address.

It will not be correct for the user to specify this in exclude list.

The logic/functions added in pkg/webhook/util/ip.go is similar to what wherabouts CNI does it internally to know the ip addresses  that could be assigned to LH pods.

pkg/webhook/util/ip.go
ihcsim: I want to better understand this. The `netip.Bits()` function can be used to calculate how many IPs in the include and exclude ranges. Based on that, we can calculate how many are usable. Then the `netip.Overlaps()` function just make sure the exclude range is within the include range.

pkg/webhook/util/ip.go
ihcsim: I meant iterate through `excludedIPs` where each iteration calls `Prefix.Contains(excludeIP)`.

pkg/webhook/util/ip.go
rrajendran17: its possible , but I feel it becomes even more convoluted as there is no single or straightforward way to do this.
Overlap returns just true or false, we need to again check for each ip to check if its contained in include list.
The prefix length or Bits() could also be checked between include and exclude, but if exclude prefix length is greater we still need what its excluding to calculate the exact number of IPs.

rrajendran17: Writing a small pseudo code here of what could be possible from my perspective after going through netip package.

```

func getIPCount(prefixLength int) int {
    return numberofips (from given prefix length)
}

func getFinalIPCount(includePrefix,excludePrefix) int {
    //Range through excludePrefix.Next()
    if includePrefix.Contains(excludePrefix.IP) {
       count++
    }

    return count
}

func main() {
     //range this for list of excludePrefix
    // Calculate number of IPs
    includenumIPs := getIPCount(includePrefix.Bits())
    excludenumIPs := getIPCount(excludePrefix.Bits())

    if !includePrefix.Overlaps(excludePrefix) {
        return includenumIPs
    } else {
        return getFinalIPCount()
    }
} 
```

rrajendran17:
This PR now takes care of the following,

1.Min allocatable IP address calculated from the formula

Required Number of IPs = (Number of Nodes * 2) + (Number of Disks * 2) + Number of Images to Download/Upload

2.Harvester webhook validates the number of ip addresses possible from include/exclude range and it must be >= min allocatable ip address , if not throws error.

3.Harvester controller logic validates the length of ip pool allocated from whereabouts CNI >= min allocatable ip addresses, if not retries again till success and LH settings updated with error (displayed in GUI)

go.mod
```diff
@@ -77,6 +77,7 @@ require (
 	github.com/iancoleman/strcase v0.2.0
 	github.com/k3s-io/helm-controller v0.16.1
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.3.0
+	github.com/k8snetworkplumbingwg/whereabouts v0.8.0
```
WebberHuang1118: Harvester master currently uses [Whereabouts v0.7.0](https://github.com/harvester/harvester/blob/ec6b7901ec25597897fe231f4eca9dcd888f2b53/deploy/charts/harvester/values.yaml#L525-L530), IMO, this should be aligned with the deployed version

pkg/config/context.go
```diff
@@ -298,6 +308,13 @@ func setupManagement(ctx context.Context, restConfig *rest.Config, opts *generic
 	management.CniFactory = cni
 	management.starters = append(management.starters, cni)
 
+	whereaboutscni, err := whereaboutcniv1.NewFactoryFromConfigWithOptions(restConfig, opts)
+	if err != nil {
+		return nil, err
+	}
+	management.WhereaboutsCNIFactory = whereaboutscni
+	management.starters = append(management.starters, cni)
```
WebberHuang1118:
```suggestion
	management.starters = append(management.starters, whereaboutscni)
```
Should like this?

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -397,6 +407,51 @@ func (h *Handler) removeOldNad(setting *harvesterv1.Setting) error {
...
+	if setting.Value == "" {
+		return nil
+	}
+
+	lhnodes, err := h.longhornNodeCache.List(metav1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list longhorn node cache %v", err)
+	}
+
+	MinallocatableIPAddrs := 0
```
WebberHuang1118:
```suggestion
	MinAllocatableIPAddrs := 0
```
Just a nit

pkg/webhook/resources/setting/validator.go
```diff
@@ -1304,8 +1308,31 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
...
-		return fmt.Errorf("Range should be subnet CIDR %v", network)
+		return fmt.Errorf("range should be subnet CIDR %v", network)
 	}
+
+	lhnodes, err := v.lhNodeCache.List(metav1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+
+	MinallocatableIPAddrs := 0
```
WebberHuang1118: Ditto

pkg/webhook/resources/setting/validator.go
```diff
@@ -1304,8 +1308,31 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
...
+	nodeNum := 1
+	//Required Number of IPs = (Number of Nodes * 2) + (Number of Nodes * Number of Disks * 2) + Number of Images to Download/Upload
+	//Number of Images to download/upload is dynamic, so skipped in the formula calculated.
+	for _, lhNode := range lhnodes {
+		MinallocatableIPAddrs = MinallocatableIPAddrs + ((nodeNum * 2) + (nodeNum * len(lhNode.Spec.Disks) * 2))
+	}
+
```
WebberHuang1118:
```suggestion
	//Required Number of IPs = (Number of Nodes * 2) + (Number of Nodes * Number of Disks * 2) + Number of Images to Download/Upload
	//Number of Images to download/upload is dynamic, so skipped in the formula calculated.
	for _, lhNode := range lhnodes {
		MinAllocatableIPAddrs = MinAllocatableIPAddrs + ( 2 + len(lhNode.Spec.Disks) * 2)
	}
```
IIUC, the variable `nodeNum` should not be necessary?

WebberHuang1118 (changes requested):
Thanks for the enhancement, please help to check the comments.

pkg/config/context.go
rrajendran17: yes, changed, thanks

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: updated

pkg/webhook/resources/setting/validator.go
rrajendran17: sure, updated

pkg/webhook/resources/setting/validator.go
rrajendran17: yes, you are right, I have corrected and updated the review

go.mod
rrajendran17: do we have plans to upgrade whereabouts in master to v.0.8.0 ? And how often this is done ?
v0.8.0 is latest for whereabouts.

WebberHuang1118: Currently, the plan is:
- Harvester v1.4.1 -> Whereabout v0.7.0
- Harvester v1.4.2 -> Whereabout v0.8.0

As Whereabout v0.8.0 had [CRD changes](https://github.com/k8snetworkplumbingwg/whereabouts/pull/471), IMO, we should update the package version along with the deployed one.

pkg/webhook/util/ip.go
```diff
@@ -0,0 +1,77 @@
...
+		for ip := network.IP.Mask(network.Mask); network.Contains(ip); incrementIP(ip) {
+			if include && (ip.Equal(network.IP) || ip.Equal(broadcast)) {
+				continue
+			}
+			ipAddrList[ip.String()] = struct{}{}
+		}
+	}
+
+	return ipAddrList, nil
+}
```
WebberHuang1118: As mentioned in [Whereabout introduction](https://github.com/k8snetworkplumbingwg/whereabouts), the broadcast IP is not skipped
```
Whereabouts takes an address range, like 192.168.2.0/24 in CIDR notation, and will assign IP addresses within that range. In this case, it will assign IP addresses from 192.168.2.1 to 192.168.2.255. 
```

and IP ends with `.0` is invalid (IIUC, we already consider it)
```
In respect to the old equipment out there that doesn't think that IP addresses that end in .0 are valid -- Whereabouts will not assign addresses that end in .0.
```

rrajendran17: Not sure if the documentation in whereabouts is misleading, but while testing I observed whereabouts is not allocating ip addresses from network or broadcast.
Example, I created /30 storage network and it was able to allocate only .1 and .2 (.0 and .3 skipped)
and for /31, it was not able to allocate ip address to any pods.(.0 and .1 skipped).

for /30, the following log was printed which included range excluding broadcast.
"Could not allocate IP in range: ip: 192.168.12.1 / - 192.168.12.2 / range: 192.168.12.0/30 / "

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -397,6 +407,51 @@ func (h *Handler) removeOldNad(setting *harvesterv1.Setting) error {
...
+	}
+
+	lhnodes, err := h.longhornNodeCache.List(metav1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list longhorn node cache %v", err)
+	}
+
+	MinAllocatableIPAddrs := 0
+	nodeNum := 1
+	//Required Number of IPs = (Number of Nodes * 2) + (Number of Nodes * Number of Disks * 2) + Number of Images to Download/Upload
```
ihcsim:
```suggestion
	//Required Number of IPs = (Number of Nodes * 2) + (Number of Disks * 2) + Number of Images to Download/Upload
```

ihcsim: Also, looks like there is a discrepancy between the UI and the doc:

image

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -397,6 +407,51 @@ func (h *Handler) removeOldNad(setting *harvesterv1.Setting) error {
...
+	var config Config
+
+	if err := json.Unmarshal([]byte(setting.Value), &config); err != nil {
+		return fmt.Errorf("parsing value error %v", err)
+	}
+
+	ipprefix := strings.Split(config.Range, "/")
+	ippoolName := ipprefix[0] + "-" + ipprefix[1]
+
+	ippools, err := h.whereaboutsCNIIPPoolCache.List(util.KubeSystemNamespace, labels.Everything())
```
ihcsim: Can we just do a `Get()` on `ippoolName`? If it doesn't exist, we return an error. If it does, compare the `spec.Allocations` with `MinAllocatableIPAddrs`?

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -397,6 +407,51 @@ func (h *Handler) removeOldNad(setting *harvesterv1.Setting) error {
...
+		if ippool.Name != ippoolName {
+			continue
+		}
+
+		if len(ippool.Spec.Allocations) >= MinAllocatableIPAddrs {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("whereabouts cni IP allocation failure for IPPool %s retrying again", ippoolName)
```
ihcsim: I think we should make it more clear that there is an error here because `len(ippool.Spec.Allocations) < MinAllocatableIPAddrs`.

Consider something like:

```
return fmt.Errorf("whereabouts cni IP allocation failure for IPPool %s retrying again. 
    required: %d, 
    allocated: %d", 
    ippoolName, 
    MinAllocatableIPAddrs, 
    len(ippool.Spec.Allocations))
```

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -397,6 +407,51 @@ func (h *Handler) removeOldNad(setting *harvesterv1.Setting) error {
 	return nil
 }
 
+func (h *Handler) validateIPAddressesAllocations(setting *harvesterv1.Setting) error {
+	if setting.Value == "" {
+		return nil
+	}
+
+	lhnodes, err := h.longhornNodeCache.List(metav1.NamespaceAll, labels.Everything())
```
ihcsim, line 415: UPDATE: sorry, i think i mixed up the LH managers with the LH nodes.

---

~One thing to consider is that you can ask the LH manager daemonset how many nodes (hence, IPs) it wants and needs. Then we don't have to ask K8s for `Everything`. I am assuming that all LH nodes will have the same number of disks since they belong to the same parent daemonset.~

~It also saves us the trouble of taking into consideration the state of every node. E.g., what if some of the nodes are unhealthy or scaling down - does it still make sense to account for them?~

~The equivalent of this using `kubectl` looks like this:~

```
$ k -n longhorn-system get ds longhorn-manager -ojsonpath='{.status.currentNumberScheduled}'
1                                                                                                                                                                                           

$ k -n longhorn-system get ds longhorn-manager -ojsonpath='{.status.desiredNumberScheduled}'
1                                             
```

pkg/webhook/resources/setting/validator.go
```diff
@@ -1304,8 +1308,31 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
 		return err
 	}
 	if !network.IP.Equal(ip) {
-		return fmt.Errorf("Range should be subnet CIDR %v", network)
+		return fmt.Errorf("range should be subnet CIDR %v", network)
 	}
+
+	lhnodes, err := v.lhNodeCache.List(metav1.NamespaceAll, labels.Everything())
```
ihcsim, line 1392: UPDATE: sorry, i think i mixed up the LH managers with the LH nodes.

---

~As with the case in the controller, if `len(lhNode.Spec.Disks)` is the same for all the LH nodes, then we can just ask the parent daemonset for the desired number of LH manager pods (hence, IPs), without needing to ask K8s for `Everything` in all namespaces.~

pkg/webhook/resources/setting/validator.go
```diff
@@ -1304,8 +1308,31 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
...
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+
+	MinAllocatableIPAddrs := 0
```
ihcsim, line 1397: To confirm my understanding, 'allocatable' here means 'desired' or 'requested', right?

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: Yeah I notified this in an another thread to @torchiaf @a110605.This has to be corrected in the UI.

And also regarding the formula,
//Required Number of IPs = (Number of Nodes * 2) + (Number of Disks * 2) + Number of Images to Download/Upload

Its actually Number of disks per node, current documentation its understood implicitly.
I am not sure if document has to be modified to make it more clear.

pkg/webhook/util/ip.go
WebberHuang1118: Do we need a document update for this?

pkg/webhook/resources/setting/validator.go
rrajendran17: allocatable means "minimum needed"

pkg/webhook/util/ip.go
rrajendran17: Looks like it from my observation, I can send a PR for review for doc update

WebberHuang1118: Thank you @rrajendran17

go.mod
rrajendran17: ok, so you suggest this PR will have v0.7.0 whereabouts in go.mod and whenever deployed version of whereabouts is upgraded (to latest), the corresponding whereabouts version in go.mod to latest will also be taken care during that time ?

WebberHuang1118: Yes, I think so, WDYT?

rrajendran17: I am not sure exactly when can we do next upgrade of whereabouts in master.This issue is anyways targeted for v1.5.0, so we could upgrade masters to latest whereabouts and commit this PR with v0.8.0.

If there is no plan for whereabouts upgrade in master anytime soon, I will change this PR to v0.7.0 whereabouts.

rrajendran17:
@WebberHuang1118 @ihcsim @w13915984028 
I believe I have addressed all your review comments.
Please check if its all good and we can merge this whenever possible.

WebberHuang1118 (approved):

rrajendran17:
@ihcsim @w13915984028 Gentle Reminder.If there are no more concerns, can we merge this ?

pkg/webhook/util/ip.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return count, err
+	}
+
+	excludeIPAddrList, err := getIPAddressesFromSubnet(excludeRange, false)
+	if err != nil {
+		return count, err
+	}
+
+	for includeIP := range includeIPAddrList {
```
w13915984028, line 34: I did an extreme test, both of following call runs several seconds on my local PC in raw Linux shell. I suspect this could also lead to OOM in container env.  As both include and exclude lists are converted to IP address MAP.

```
	c, _ := GetUsableIPAddressesCount("192.168.123.100/24", []string{"1.1.1.1/8"})
	c1, _ := GetUsableIPAddressesCount("1.1.1.1/8", []string{"192.168.123.100/24"})
```
Maybe we need to further enhance:
(1) The include subnet should not be too large, /16 is fairly enough?
(2) The excluded network should be a sub-net of include network first.

w13915984028 (approved):
LGTM, please also check the last question, thanks.

pkg/webhook/util/ip.go
ihcsim: @w13915984028 @rrajendran17 Following slack conversation, I think we should try to calculate the number of include and exclude IP addresses within the provided ranges using the prefix bits, instead of stepping through every IP address within the range, on every admission request.

My main concern is that the admission webhook is configured to run with the [default (roundtrip) 10 seconds timeout](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts). If a timeout occurred, it will cause the K8s api server to reject admission requests of the storage-network `settings` resource. This can lead to flaky scenarios that are not immediately obvious and hard to reason about, especially since the amount of time for the storage network helper to finish is already dictated by the results of these two "list all" operations:

```
# line 1277
nodes, err := v.nodeCache.List(labels.Everything())

# line 1314
lhnodes, err := v.lhNodeCache.List(metav1.NamespaceAll, labels.Everything())
```

Essentially, when it comes to validating the storage network range, the validator should:
* if exclude range isn't within include range, validation should fail right away (i assume this is the intended behaviour)
* otherwise, exclude range is a subset of include range
* now compute:
    * `includeCount`: the number of IPs in the include range using its prefix bit
    * `excludeCount`: the number of IPs in the exclude range using its prefix bit
    * `desiredCount = includeCount - excludeCount - 1` 
    * `assertGreaterOrEqual(minAllocatable, desiredCount)`

WDYT?

rrajendran17: Yes my understanding is exclude must always be subset of include range.
I think if we make sure exclude is subset of include, then verifying the count of IP addresses is enough instead of getting all ip addresses.

Also, I tested for /8 with the loop changes, the webhook timed out after 10seconds.
Thanks @w13915984028 for pointing out and @ihcsim for the code suggestions.

rrajendran17:
@w13915984028 @ihcsim 
I am retaining the existing code in pkg/webhook/util/ip.go and have added checks to catch errors from user during webhook validation.
a.Validate include/exclude range prefix length for min /16
b.Validate include and exclude range overlap
c.Validate if overlap true, and exclude prefix length is not <= include prefix length (no allocatable ip addresses)
d.Validate if exclude does not contain set of overlapping subnet ranges.

The above fixes the problem of users giving some big and invalid ranges.

Just checking with prefix length to get allocatable address count is doable, but it has a bunch of edge cases to be handled and it is possible to miss some.So I have not considered that approach in this PR

w13915984028:
@rrajendran17 Thanks for your update, with those improvements, it is ok to move forward from my side.

ihcsim (approved):
LGTM - thanks for working on this

rrajendran17:
https://github.com/Mergifyio backport v1.5

Issue #7131, opened 2024-12-10
[BUG] Current free disk space check does not honor the settings for harvesterhci.io/minFreeDiskSpaceGB: "9"


Martin-Weiss:
We have a way to override the free disk space check with the annotation harvesterhci.io/minFreeDiskSpaceGB on the version.yaml. Unfortunately the validator does not honor this here:

https://github.com/harvester/harvester/blob/93523eab3d20719d688649798e26a83625645dde/pkg/webhook/resources/upgrade/validator.go#L295

-> could we get this fixed?

Hint: Harvester version upgrade 1.3.2 to 1.4.0 is where we had to realize this..

m-ildefons:
The check linked here ensures that the sum of the *currently used disk space*[1] and the *new container image tarballs*[2] does not exceed the kubelet image garbage collection threshold.

By default the kubelet image garbage collection threshold is 85%.
This check is independent of the `harvesterhci.io/minFreeDiskSpaceGB` annotation, because it is independent of the total free disk space.
It does however take the actual [kubelet image garbage collection threshold](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration) into account.

[1]: on the partition mounted at `/usr/local`
[2]: 13 GiB: https://github.com/harvester/harvester/blob/ef97c6c4fa49086a082c12682ef5287d4d3777f5/pkg/webhook/resources/upgrade/validator.go#L46

Martin-Weiss:
So does the setting harvesterhci.io/minFreeDiskSpaceGB have any sense, now? 

From my point of view we can set for harvesterhci.io/minFreeDiskSpaceGB whatever we want and it does not change any behavior in the situation that causes the upgrade blocker. 

So maybe we need to remove that functionality for harvesterhci.io/minFreeDiskSpaceGB or we need to adjust / combine it with the additional check that relates to the Kubernetes image garbage collection.

m-ildefons:
I think bottom line is you're right that the image garbage collection threshold can disturb upgrades that should work by all other documented prerequisites. This is a bug.

Let's do some math:

Assuming I have a node with the minimum disk size, 250GB[1], and exactly 30GB are free, which as you know is the default (documented) minimum requirement for starting an upgrade.
Then 30 / 250 = 12% of the disk are free, 88% of the disk are used, which is above the 85% garbage collection threshold.
[...]

PR #7134, opened 2024-12-11
feat: deprecated minFreeDiskSpaceGB and add skipGarbageCollectionThresholdCheck

FrankYang0529:

**Problem:**
The `minFreeDiskSpaceGB` is not useful now, because we have limit for disk to have at least 250GB when installing harvester. The default garbage collection threshold is 15%. That means the disk must have 37.5 GB to continue the upgrade. The default `minFreeDiskSpaceGB` is 30GB. It's lower then garbage collection threshold check.

**Solution:**
Deprecate `minFreeDiskSpaceGB`. Add `harvesterhci.io/skipGarbageCollectionThresholdCheck` annotation.

**Related Issue:**
https://github.com/harvester/harvester/issues/7131

**Test plan:**
2. Make one node disk full. Like `dd if=/dev/random of=/usr/local/test.img bs=1G count=100`
3. Create a test version.

```yaml
kind: Version
metadata:
  name: 1.5.0-head
  namespace: harvester-system
spec:
  isoURL: http://192.168.0.181:8000/harvester-1.5.0-head-amd64.iso
  minUpgradableVersion: 1.3.2
  releaseDate: "20241231"
```

Screenshot 2024-12-11 at 3 22 45 PM

5. Add annotation `harvesterhci.io/skipGarbageCollectionThresholdCheck: true` to the version.
6. Click upgrade on the dashboard. It should pass.

### DIFF
--- a/pkg/webhook/resources/upgrade/validator.go
+++ b/pkg/webhook/resources/upgrade/validator.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/docker/go-units"
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	fleetv1alpha1 "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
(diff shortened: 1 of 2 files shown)


bk201 (approved):

brandboat (approved):
LGTM, thanks for the PR. Mind that we need to update doc for this change. https://docs.harvesterhci.io/v1.5/upgrade/index#free-system-partition-space-requirement, the `harvesterhci.io/minFreeDiskSpaceGB` is removed after this change.

FrankYang0529:
@brandboat thanks for the reminder. Created a document PR for it.

https://github.com/harvester/docs/pull/706

Issue #6806, opened 2024-10-16
[BUG] USB Device Using VM Has Migrate As An Action Option


irishgordo:
**Describe the bug**
A VM Using VM Devices has the available action of Migration on the VM, when it the VM can never migrate.
See video:

https://github.com/user-attachments/assets/76911dd6-0966-48f9-97d1-c1e78bf8c9d3

**To Reproduce**
1. Have two avaialbe Nodes for a Harvester cluster
2. Try to migrate a VM with USB devices
3. It will fail the first time
4. But then the Migrate option disappears 

**Expected behavior**
The migrate option call-to-action shouldn't exist because the VM can not be migrated.

**Environment**
 - Harvester ISO version: v1.4.0-rc2
 - Underlying Infrastructure: Dell PowerEdge R720 & HP DL 160 Gen 9

**Additional Context**
- note, it's "reproduce/always", but "only" on the "first try" of the VM -> you'll need to build another VM to keep repeating the seen behavior

Yu-Jack:
It also happens in PCI Devices. After first try, the `nodeSelector` is updated. That's why it doesn't show `Migrate` button in second try.

bk201:
@Yu-Jack is this an GUI issue?

Yu-Jack:
@bk201 It's backend issue, we should add more different cases here.

https://github.com/harvester/harvester/blob/6b5070ae39f2689059102a538184b0fde6b20ee3/pkg/api/vm/formatter.go#L231-L251

w13915984028:
FYI, connect https://github.com/harvester/harvester/issues/6843, to show the `migrate` menu dynamically & accurately, it needs to  check a couple of scenarios.

Yu-Jack:
I'm thinking we might not need to remove the Migrate button if the VM is not allowed for migration. 

If we remove Migrate button, users don't have a chance understand why their migration failed. They would need to refer documentation to check what kind of VMs can be migrated.

Users might think of:

1. I'd like to migrate a VM to another node, but there is no Migrate button. Why can't I migrate it?
2. Search Harvester document to check what types of VMs can be migrated. Okay, VM with an attached host device or CD-ROM is not allowed.
3. Go back to Harvester dashboard to check which restriction my VM belongs to.

[...]

PR #7137, opened 2024-12-11
feat: use common func to detect vm is migratable or not

Yu-Jack:

**Problem:**
When we try to migrate a VM with an attached PCI or USB device, the GUI initially shows the Migrate button. However, after clicking it, an error message appears. After that, the Migrate button disappears from the menu.

**Solution:**
We should still display Migrate button, and show the error message after clicking it.

**Related Issue:**
https://github.com/harvester/harvester/issues/6806

**Test plan:**

#### Host Device

1. Start a VM attached with an attached PCI or USB device.
2. After clicking live migration, we should show the specific error message for pci/usb device.

image

#### CD-ROM

1. Start a VM with CD-ROM
2. After clicking live migration, we should show the specific error message for CD-ROM.

image

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -73,6 +73,8 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	resource.AddAction(request, addVolume)
 	resource.AddAction(request, removeVolume)
 	resource.AddAction(request, cloneVM)
+	resource.AddAction(request, migrate)
+	resource.AddAction(request, findMigratableNodes)
 
 	if canEjectCdRom(vm) {
 		resource.AddAction(request, ejectCdRom)
@@ -101,11 +103,6 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 
 	if vf.canUnPause(vmi) {
 		resource.AddAction(request, unpauseVM)
-	}
-
-	if canMigrate(vmi) {
-		resource.AddAction(request, migrate)
-		resource.AddAction(request, findMigratableNodes)
 	}
 
 	if canAbortMigrate(vmi) {
@@ -223,28 +220,6 @@ func (vf *vmformatter) canStop(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv1.Vi
 		default:
 			// skip to other condition
 		}
-	}
-
-	return true
-}
-
-func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
-	if vmi == nil {
-		return false
-	}
-
-	if !vmi.IsRunning() {
-		return false
-	}
-
-	// The VM is already in migrating state
-	if vmi.Annotations[util.AnnotationMigrationUID] != "" {
-		return false
-	}
-
-	// The VM is set to run on a specific node
-	if vmi.Spec.NodeSelector != nil && vmi.Spec.NodeSelector[corev1.LabelHostname] != "" {
-		return false
 	}
 
 	return true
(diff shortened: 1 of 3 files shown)


pkg/util/virtualmachineinstance/virtualmachineinstance.go
```diff
@@ -53,13 +53,44 @@ func GetAllNonLiveMigratableVMINames(vmis []*kubevirtv1.VirtualMachineInstance,
...
+		return fmt.Errorf("VM %s considered non-live migratable due to migration state", vmiNamespacedName)
+	}
+
+	// Node selectors
+	if vmi.Spec.NodeSelector != nil && vmi.Spec.NodeSelector[corev1.LabelHostname] != "" {
+		return fmt.Errorf("VM %s considered non-live migratable due to node selectors", vmiNamespacedName)
+	}
+
+	// PCIe devices
+	if vmi.Spec.Domain.Devices.HostDevices != nil {
```
ibrokethecloud: can we please also add a check for `Devices.GPUs` as that is being by vGPU feature

ibrokethecloud (commented):
is there a case when `migrate` action can be called directly? When the `migrate` action is called from the UI the first call is to `findMigratableNodes` to filter nodes. With this change `findMigratableNodes` should error out when the VM meets the condition of being non migratable in which case the call to `virtualmachineinstance.ValidateVMMigratable`  in `migrate` action is not needed.

Yu-Jack:
> is there a case when `migrate` action can be called directly? When the `migrate` action is called from the UI the first call is to `findMigratableNodes` to filter nodes. With this change `findMigratableNodes` should error out when the VM meets the condition of being non migratable in which case the call to `virtualmachineinstance.ValidateVMMigratable` in `migrate` action is not needed.

As far as I know, there are no cases where migrate is called directly. But, I think we could follow original logic, as validating migration logic (`canMigrate`) is used in both the `migrate` and the `findMigratableNodes`.

pkg/util/virtualmachineinstance/virtualmachineinstance.go
Yu-Jack: BTW, the condition `len(vmi.Spec.Domain.Devices.HostDevices) != 0` is better. We don't need to check `nil` anymore.

ibrokethecloud (approved):
thanks for including the vgpu changes.

lgtm.

bk201 (approved):

pkg/api/vm/formatter.go
```diff
@@ -103,11 +105,6 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 		resource.AddAction(request, unpauseVM)
 	}
 
-	if canMigrate(vmi) {
```
w13915984028, line 106: `migrate` and `abortMigration` menu are mutually exclusive, or both are not existing (e.g. vmi is not in running state)

after this PR, they may come together, did you notice when testing?

pkg/api/vm/formatter.go
```diff
@@ -228,28 +225,6 @@ func (vf *vmformatter) canStop(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv1.Vi
 	return true
 }
 
-func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
-	if vmi == nil {
-		return false
-	}
-
-	if !vmi.IsRunning() {
```
w13915984028, line 236: this condition is very important, only when vmi is running, it has the possibility to migrate

w13915984028 (commented):
need to recheck this PR, thanks.

pkg/api/vm/formatter.go
w13915984028: check https://github.com/harvester/harvester/issues/7140#issuecomment-2607710331, both pending starting VM and stopped VM have `migrate` menu.

Issue #7136, opened 2024-12-11
[ENHANCEMENT] Remove guest cluster in Rancher if related namespace is removed in Harvester


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
When removing a namespace in Harvester, no VM can be deployed in that namespace. We can have a setting for user to cascade remove related guest clusters.

w13915984028:
@FrankYang0529 

Is this still a valid ticket? 

Could you explain more about the user-story of this scenario? 

1. Is the namespace for guest-cluster managed by Rancher?  Rancher is responsible for create/delete the namespace.
2. If the namespace is accidentally deleted from Harvester, will Rancher re-create it?
3. In which case, Harvester needs to clean-up the related resources.

When Harvester is imported to Rancher as a node-driver, the Rancher should be the source of truth.

btw: I am also working on https://github.com/harvester/harvester/issues/7940 to add more details. thanks.

albinsun:
Hi @Yu-Jack 
Please help to verify this issue, thank you.

Yu-Jack:
🟢  Passed with v1.6-17d4fcea-head

<details><summary>Step1: Configure the rancher cluster in the GUI</summary>

<img width="2304" height="1261" alt="Image" src="https://github.com/user-attachments/assets/dca3e253-3cab-46fc-b181-badf63441fef" />

</details>

<details><summary>Step2: Create a guest cluster with a newly created namespace </summary>

<img width="2290" height="1260" alt="Image" src="https://github.com/user-attachments/assets/db565d19-152b-4d9c-a1ca-3b8b1114579f" />

<img width="2296" height="1259" alt="Image" src="https://github.com/user-attachments/assets/de394362-1c12-438e-9200-39fd0b9526f4" />

<img width="2294" height="1259" alt="Image" src="https://github.com/user-attachments/assets/81311e2f-dff2-4e47-a18a-0491c9085629" />

</details>

<details><summary>Step3: Remove the namespace in Harvester GUI. </summary>

<img width="2293" height="1261" alt="Image" src="https://github.com/user-attachments/assets/61aece38-cc82-4806-8bdb-8dde42d31fba" />

<img width="2290" height="1246" alt="Image" src="https://github.com/user-attachments/assets/c7636bd9-73e7-45d9-84da-c067c130a4a3" />

</details>

<details><summary>Step4: All resources are removed. </summary>

[...]

PR #7145, opened 2024-12-12
feat: remove guest cluster if related namespace is removed in Harvester

FrankYang0529:

Remove guest cluster in Rancher if related namespace is removed in Harvester

**Related Issue:**
https://github.com/harvester/harvester/issues/7136

**Test plan:**
2. Import the harvester cluster to a rancher cluster.
3. Copy rancher kubeconfig and save to a file like:

```yaml
apiVersion: v1
kind: Config
clusters:
- name: "local"
  cluster:
    server: "https://192.168.0.181:6443/k8s/clusters/local"
    certificate-authority-data: "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJ2VENDQ\
      VdPZ0F3SUJBZ0lCQURBS0JnZ3Foa2pPUFFRREFqQkdNUnd3R2dZRFZRUUtFeE5rZVc1aGJXbGoKY\
      kdsemRHVnVaWEl0YjNKbk1TWXdKQVlEVlFRRERCMWtlVzVoYldsamJHbHpkR1Z1WlhJdFkyRkFNV\
      GN6TXprdwpNRE0xTmpBZUZ3MHlOREV5TVRFd05qVTVNVFphRncwek5ERXlNRGt3TmpVNU1UWmFNR\
      Vl4SERBYUJnTlZCQW9UCkUyUjVibUZ0YVdOc2FYTjBaVzVsY2kxdmNtY3hKakFrQmdOVkJBTU1IV\
      1I1Ym1GdGFXTnNhWE4wWlc1bGNpMWoKWVVBeE56TXpPVEF3TXpVMk1Ga3dFd1lIS29aSXpqMENBU\
      VlJS29aSXpqMERBUWNEUWdBRTZFVHlrSnExQStwSgpoNXRzUno4Um9mbGVBZlJyaVBXaFd2V3Vhc\
      1loL0M2S0lCaFMrT0pFRmU2ZTFKZ2hBaUpNcmRwWUgyaUl6Y2xwCmhjTUtHUjlCRzZOQ01FQXdEZ\
      1lEVlIwUEFRSC9CQVFEQWdLa01BOEdBMVVkRXdFQi93UUZNQU1CQWY4d0hRWUQKVlIwT0JCWUVGR\
      XRWcGlIcDVxRmIwUys0cXY4RkNmeTc1RWRFTUFvR0NDcUdTTTQ5QkFNQ0EwZ0FNRVVDSVFEUgpLc\
      jFJekdHUWIzcFlnOURxaG9md0EzQTAranlUdkRwa0UwY29WNmZLeXdJZ1d4UmRQSnkrSlBZMGJLZ\
      EdmSzUyCm85ZE1sRUZZV2poa3orYU9YeXRJMjZVPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0t"

users:
- name: "local"
  user:
    token: "kubeconfig-user-qjg5lxq42q:9qttqjmkzf2n6t7jfbmswg6z6vbzcgbhlqws5nx7dsnndjfql49vpz"


contexts:
- name: "local"
  context:
    user: "local"
    cluster: "local"

current-context: "local"
```

4. Use base64 to encode the kubeconfig:

```
base64 test.config -w 0
```

5. Save the output to a secret `harvester-system/rancher-cluster-config` in harvester like:

```yaml
apiVersion: v1
data:
  kubeConfig: YXBpVmVyc2lvbjogdjEKa2luZDogQ29uZmlnCmNsdXN0ZXJzOgotIG5hbWU6ICJsb2NhbCIKICBjbHVzdGVyOgogICAgc2VydmVyOiAiaHR0cHM6Ly8xOTIuMTY4LjAuMTgxOjY0NDMvazhzL2NsdXN0ZXJzL2xvY2FsIgogICAgY2VydGlmaWNhdGUtYXV0aG9yaXR5LWRhdGE6ICJMUzB0TFMxQ1JVZEpUaUJEUlZKVVNVWkpRMEZVUlMwdExTMHRDazFKU1VKMlZFTkRRXAogICAgICBWZFBaMEYzU1VKQlowbENRVVJCUzBKblozRm9hMnBQVUZGUlJFRnFRa2ROVW5kM1IyZFpSRlpSVVV0RmVFNXJaVmMxYUdKWGJHb0tZXAogICAgICBrZHNlbVJIVm5WYVdFbDBZak5LYmsxVFdYZEtRVmxFVmxGUlJFUkNNV3RsVnpWb1lsZHNhbUpIYkhwa1IxWjFXbGhKZEZreVJrRk5WXAogICAgICBHTjZUWHByZHdwTlJFMHhUbXBCWlVaM01IbE9SRVY1VFZSRmQwNXFWVFZOVkZwaFJuY3dlazVFUlhsTlJHdDNUbXBWTlUxVVdtRk5SXAogICAgICBWbDRTRVJCWVVKblRsWkNRVzlVQ2tVeVVqVmliVVowWVZkT2MyRllUakJhVnpWc1kya3hkbU50WTNoS2FrRnJRbWRPVmtKQlRVMUlWXAogICAgICAxSTFZbTFHZEdGWFRuTmhXRTR3V2xjMWJHTnBNV29LV1ZWQmVFNTZUWHBQVkVGM1RYcFZNazFHYTNkRmQxbElTMjlhU1hwcU1FTkJVXAogICAgICBWbEpTMjlhU1hwcU1FUkJVV05FVVdkQlJUWkZWSGxyU25FeFFTdHdTZ3BvTlhSelVubzRVbTltYkdWQlpsSnlhVkJYYUZkMlYzVmhjXAogICAgICAxbG9MME0yUzBsQ2FGTXJUMHBGUm1VMlpURktaMmhCYVVwTmNtUndXVWd5YVVsNlkyeHdDbWhqVFV0SFVqbENSelpPUTAxRlFYZEVaXAogICAgICAxbEVWbEl3VUVGUlNDOUNRVkZFUVdkTGEwMUJPRWRCTVZWa1JYZEZRaTkzVVVaTlFVMUNRV1k0ZDBoUldVUUtWbEl3VDBKQ1dVVkdSXAogICAgICBYUldjR2xJY0RWeFJtSXdVeXMwY1hZNFJrTm1lVGMxUldSRlRVRnZSME5EY1VkVFRUUTVRa0ZOUTBFd1owRk5SVlZEU1ZGRVVncExjXAogICAgICBqRkpla2RIVVdJemNGbG5PVVJ4YUc5bWQwRXpRVEFyYW5sVWRrUndhMFV3WTI5V05tWkxlWGRKWjFkNFVtUlFTbmtyU2xCWk1HSkxaXAogICAgICBFZG1TelV5Q204NVpFMXNSVVpaVjJwb2Ezb3JZVTlZZVhSSk1qWlZQUW90TFMwdExVVk9SQ0JEUlZKVVNVWkpRMEZVUlMwdExTMHQiCgp1c2VyczoKLSBuYW1lOiAibG9jYWwiCiAgdXNlcjoKICAgIHRva2VuOiAia3ViZWNvbmZpZy11c2VyLXFqZzVsOGpoYmw6enFsY212bWJjcTV0YmJzamxidzVqdDY4OXFucHR4NWx6YmxqYzd0NmZybGZmNzJiamZmdnNuIgoKCmNvbnRleHRzOgotIG5hbWU6ICJsb2NhbCIKICBjb250ZXh0OgogICAgdXNlcjogImxvY2FsIgogICAgY2x1c3RlcjogImxvY2FsIgoKY3VycmVudC1jb250ZXh0OiAibG9jYWwiCg==
kind: Secret
metadata:
  name: rancher-cluster-config
  namespace: harvester-system
type: secret
```

6. Set `rancher-cluster` setting with following value:

```
'{"removeUpstreamClusterWhenNamespaceIsDeleted":true}'
```

7. Create a new namespace `thisistest` in harvester.
8. Create a new VMImage / VLAN network under `thisistest` namespace.
9. Create a guest cluster and set the pool namespace as `thisistest`.
10. After the guest cluster is ready, remove `thisistest` namespace in harvester. Check the guest cluster is removed.

### DIFF
--- a/pkg/controller/master/rancher/namespace.go
+++ b/pkg/controller/master/rancher/namespace.go
@@ -1,0 +1,108 @@
+package rancher
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sschema "k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/harvester/harvester/pkg/settings"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+func (h *Handler) onNamespaceRemoved(_ string, namespace *corev1.Namespace) (*corev1.Namespace, error) {
+	if namespace == nil {
+		return namespace, nil
+	}
+
+	value := settings.RancherCluster.Get()
+	if value == "" {
+		return namespace, nil
+	}
+
+	rancherClusterConfig, err := settings.DecodeConfig[settings.RancherClusterConfig](value)
+	if err != nil {
+		return namespace, err
+	}
+
+	if !rancherClusterConfig.RemoveUpstreamClusterWhenNamespaceIsDeleted {
+		return namespace, nil
+	}
+
+	secret, err := h.SecretCache.Get(util.HarvesterSystemNamespaceName, util.RancherClusterConfigSecretName)
+	if err != nil {
+		return nil, err
+	}
+
+	if secret.Data == nil || len(secret.Data["kubeConfig"]) == 0 {
+		return nil, fmt.Errorf("kubeConfig is empty in secret %s/%s", util.HarvesterSystemNamespaceName, util.RancherClusterConfigSecretName)
+	}
+
+	kc, err := clientcmd.RESTConfigFromKubeConfig(secret.Data["kubeConfig"])
+	if err != nil {
+		return nil, err
+	}
+
+	dynamicClient, err := dynamic.NewForConfig(kc)
+	if err != nil {
+		return nil, err
+	}
+
+	harvesterConfigsGroupVersion := k8sschema.GroupVersionResource{Group: "rke-machine-config.cattle.io", Version: "v1", Resource: "harvesterconfigs"}
+	provisioningClusterGroupVersion := k8sschema.GroupVersionResource{Group: "provisioning.cattle.io", Version: "v1", Resource: "clusters"}
+	harvesterConfigs, err := dynamicClient.Resource(harvesterConfigsGroupVersion).Namespace("fleet-default").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		logrus.WithError(err).Error("Fail to list harvesterConfigs")
+		return nil, err
+	}
+
+	for _, harvesterConfig := range harvesterConfigs.Items {
+		vmNamespace, ok := harvesterConfig.Object["vmNamespace"].(string)
+		if !ok {
+			continue
+		}
+		if vmNamespace == namespace.Name {
+			metadata, ok := harvesterConfig.Object["metadata"].(map[string]interface{})
+			if !ok {
+				continue
+			}
+			ownerReferences, ok := metadata["ownerReferences"].([]interface{})
+			if !ok {
+				continue
+			}
+			for _, o := range ownerReferences {
+				ownerReference, ok := o.(map[string]interface{})
+				if !ok {
+					continue
+				}
+				clusterName, ok := ownerReference["name"].(string)
+				if !ok {
+					continue
+				}
+
+				logrus.WithFields(logrus.Fields{
+					"harvester.namespace": namespace.Name,
+					"cluster":             clusterName,
+				}).Info("Namespace is removed, removing upstream cluster")
+				err = dynamicClient.Resource(provisioningClusterGroupVersion).Namespace("fleet-default").Delete(context.TODO(), clusterName, metav1.DeleteOptions{})
+				if err != nil {
+					if apierrors.IsNotFound(err) {
+						return namespace, nil
+					}
+					logrus.WithError(err).WithFields(logrus.Fields{
+						"harvester.namespace": namespace.Name,
+						"cluster":             clusterName,
+					}).Info("Fail to remove upstream cluster")
+					return nil, err
+				}
+			}
+		}
+	}
+	return namespace, nil
+}
--- a/pkg/controller/master/rancher/register.go
+++ b/pkg/controller/master/rancher/register.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	"github.com/harvester/harvester/pkg/config"
+	"github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	networkingv1 "github.com/harvester/harvester/pkg/generated/controllers/networking.k8s.io/v1"
 	"github.com/harvester/harvester/pkg/util"
 )
@@ -26,6 +27,7 @@ import (
 const (
 	appLabelName             = "app.kubernetes.io/name"
 	controllerRancherName    = "harvester-rancher-controller"
+	controllerNamespaceName  = "harvester-namespace-controller"
 	caCertsSetting           = "cacerts"
 	defaultAdminLabelKey     = "authz.management.cattle.io/bootstrapping"
 	defaultAdminLabelValue   = "admin-user"
@@ -65,6 +67,7 @@ type Handler struct {
 	Deployments              ctlappsv1.DeploymentClient
 	Namespace                string
 	RancherTokenController   rancherv3.TokenController
+	SettingCache             v1beta1.SettingCache
 }
 
 type VIPConfig struct {
@@ -87,7 +90,9 @@ func Register(ctx context.Context, management *config.Management, options config
 		configmaps := management.CoreFactory.Core().V1().ConfigMap()
 		nodes := management.CoreFactory.Core().V1().Node()
 		pods := management.CoreFactory.Core().V1().Pod()
+		namespaces := management.CoreFactory.Core().V1().Namespace()
 		deployments := management.AppsFactory.Apps().V1().Deployment()
+		settings := management.HarvesterFactory.Harvesterhci().V1beta1().Setting()
 		h := Handler{
 			RancherSettings:          rancherSettings,
 			RancherSettingController: rancherSettings,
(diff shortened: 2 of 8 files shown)


w13915984028:
@FrankYang0529 I still have the concern https://github.com/harvester/harvester/issues/7136#issuecomment-2753805807. thanks.

In which scenario, the namespace is removed but guest cluster is still there? Generally, Harvester should block user to do so if the guest cluster is still running, to avoid the battling with Rancher Manager.

FrankYang0529:
@w13915984028 Could you check the discussion in Jira? There has more context about this issue. Thanks.

pkg/controller/master/rancher/namespace.go
```diff
@@ -0,0 +1,104 @@
...
+	rancherClusterConfig, err := settings.DecodeConfig[settings.RancherClusterConfig](value)
+	if err != nil {
+		return namespace, err
+	}
+
+	if !rancherClusterConfig.RemoveUpstreamClusterWhenNamespaceIsDeleted {
+		return namespace, nil
+	}
+
+	secret, err := h.SecretCache.Get(util.HarvesterSystemNamespaceName, util.RancherClusterConfigSecretName)
```
w13915984028, line 38: If the secret was created before the setting of `rancher-cluster-*`, and later the secret is removed;  the removing of normal namespace will be blocked here, even if the namespace has no relationship with Rancher Manager.

pkg/controller/master/rancher/namespace.go
```diff
@@ -0,0 +1,104 @@
...
+				ownerReference, ok := o.(map[string]interface{})
+				if !ok {
+					continue
+				}
+				clusterName, ok := ownerReference["name"].(string)
+				if !ok {
+					continue
+				}
+
+				logrus.WithFields(logrus.Fields{
```
w13915984028, line 89: Let Harvester to remove Rancher Manager's resource is very anti-patten.

When Race conditions are there, they are totally out-of-control.

Harvester clean all this namespace related resource on Harvester cluster, is reasonable;  but reverse cleanning Rancher Manager's resource is not a good idea.

w13915984028 (commented):
Did not see a strong reason to reversely clean Rancher's resources from Harvester. 

Harvester is like a driver, Rancher Manager is the upper layer APP, driver should not hack into APP and do some cleanup work.

For general scenairo, when one Harvester namespace is deleted, clean all it's resources, need some more discussions, e.g.:

(1) If user delete the namesapce from Harvester, should it be allowed or denied?
(2) If user delete from Rancher Manager, how to identify.
(3) How to identify a namespace is used by Rancher or not...

The JIRA's case, needs to be converted to a general user scenario, that is also helpful for QA to test different cases.

I am working on https://github.com/harvester/harvester/issues/7940, and will add [this issue ](https://github.com/harvester/harvester/issues/7136) and PR https://github.com/harvester/harvester/pull/7145 into consideration.

FrankYang0529:
> Did not see a strong reason to reversely clean Rancher's resources from Harvester.

I don't know why you didn't see it. It's on last two messages from Radovan and me. Please check it again. We already get alignment to remove guest clusters if namespace is removed.

> (1) If user delete the namesapce from Harvester, should it be allowed or denied?
(2) If user delete from Rancher Manager, how to identify.
(3) How to identify a namespace is used by Rancher or not...

All these detailed scenario are not different. Once the namespace is removed from Harvester. Users don't have any chance to create VM in that namespace.

w13915984028:
@FrankYang0529 Did you consider another solution:

All VMs created via node-driver have the label like `harvesterhci.io/machineSetName`, from the VM object, we can further label a namespace like `guestClusterUsing`, when user delete this namespace from any tool, check if vm/vmis are still on this namespace and block it before vms are all removed.

https://github.com/harvester/docker-machine-driver-harvester/blob/1c0d7b2175c8a2a237f921c93805bf766724967d/harvester/create.go#L28

The current solution can work for the specific case, and might have side affects.

FrankYang0529:
@w13915984028 Please check the comment in Jira again. I already proposed the denying request method there, but they still prefer to remove guest cluster when the namespace is removed. Also, they mentioned we already have confirmation before removing namespace on UI, so this PR is acceptable.

pkg/controller/master/rancher/namespace.go
```diff
@@ -0,0 +1,104 @@
...
+					"harvester.namespace": namespace.Name,
+					"cluster":             clusterName,
+				}).Info("Namespace is removed, removing upstream cluster")
+				err = dynamicClient.Resource(provisioningClusterGroupVersion).Namespace("fleet-default").Delete(context.TODO(), clusterName, metav1.DeleteOptions{})
```
w13915984028, line 93: it might not always be `fleet-default`
```
vendor/github.com/rancher/rancher/pkg/fleet/const.go:	ClustersDefaultNamespace = "fleet-default"
vendor/github.com/rancher/rancher/pkg/settings/setting.go:	FleetDefaultWorkspaceName           = NewSetting("fleet-default-workspace-name", fleetconst.ClustersDefaultNamespace) // fleetWorkspaceName to assign to clusters with none
pkg/data/rancher.go:	FleetDefaultNamespace = "fleet-default"
```

pkg/controller/master/rancher/namespace.go
```diff
@@ -0,0 +1,104 @@
...
+				if err != nil {
```
w13915984028, line 94: skip `isNotExisting`?

pkg/controller/master/rancher/namespace.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/harvester/harvester/pkg/settings"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+func (h *Handler) onNamespaceRemoved(_ string, namespace *corev1.Namespace) (*corev1.Namespace, error) {
+	if namespace == nil {
+		return namespace, nil
+	}
+
```
w13915984028, line 23: allow additional label/annotation to skip the cascade-deleting quickly on L22? in case error blocking

w13915984028 (dismissed):

pkg/controller/master/rancher/namespace.go
FrankYang0529: I tried to update `fleet-default-workspace-name` in Rancher and create RKE2 guest cluster, but the cluster object is still in `fleet-default`. I will keep this and see whether need to do some enhancement in the future.

pkg/controller/master/rancher/namespace.go
FrankYang0529: Updated it.

pkg/controller/master/rancher/namespace.go
FrankYang0529: I'm not sure which label/annotation you mean here.

w13915984028: use a new annotation here, the main consideration is:

if remote Rancher is down/not reachable, then delete any namespace on Harvester is blocked, the current solution is to disable the `settings`

if we allow certain namespace to skip the check, it will be another solution for abovce case

it depends on your choice to add it or not

w13915984028 (approved):
LGTM, thanks.

Issue #4352, opened 2023-08-01
[FEATURE] Live migration progress awareness


Martin-Weiss:
**Is your enhancement related to a problem? Please describe.**
We have the problem over and over again, that the migration does get stuck and does not complete. In this situation the UI does not help at all to see where the problem might be.. what status the migration is in and how long it might take etc.

**Describe the solution you'd like**
In the UI when doing a migration indicate the steps that are done during a migration (create new vm, copy memory, show pogress of memory copy, ....)

**Describe alternatives you've considered**
did not find an alternative

**Additional context**
mayhe these metrics can help as well?

https://github.com/kubevirt/kubevirt/pull/7474
https://github.com/kubevirt/kubevirt/pull/7946

FrankYang0529:
After some investigation, we can't get migration progress from [VirtualMachineInstance.Status.MigrationState](https://github.com/kubevirt/kubevirt/blob/e4c7d4e53269d2c416a4daae761b262b668529ce/staging/src/kubevirt.io/api/core/v1/types.go#L637-L684). If we look into code, we can find [virt-launcher gets migration status](https://github.com/kubevirt/kubevirt/blob/e4c7d4e53269d2c416a4daae761b262b668529ce/pkg/virt-launcher/virtwrap/live-migration-source.go#L686-L699) from [Domain](https://gitlab.com/libvirt/libvirt-go-module/-/blob/master/domain.go?ref_type=heads#L3209-3243). If we can find a way to get connection through virt-api -> virt-handler -> virt-launcher, we can use `MemTotal` and `MemProcessed` to get the progress.

ibrokethecloud:
we should also leverage the conditions on the `vmim` object to indicate the various phases of the virtual machine migration.

For example in certain cases due to cluster specific defaults of 5 parallel migrations and 2 migrations per target node, a number of migrations may be queued.

We should ensure that we indicate that the migration is queued, so users know that the migration is scheduled but has not yet started.

innobead:
The backend effort was finished, and the current status is pending for UI improvement, https://github.com/harvester/harvester/issues/7339.  cc @harvester/ui

TachunLin:
### Test Scenarios

1. Migration success scenario
    - Check migration details
    - Check the Grafana dashboard 
  
2. Migration failed scenario
    - Check migration details
    - Check the Grafana dashboard 
  
3. Abort Migration scenario
    - Check migration details
    - Check the Grafana dashboard 

[...]

PR #7146, opened 2024-12-12
Add vm live migration enhancement HEP

brandboat:
HEP for https://github.com/harvester/harvester/issues/4352


w13915984028:
FYI, please also check issue https://github.com/harvester/harvester/issues/7180 and PR https://github.com/harvester/harvester/issues/7180 (WIP).

The current vm migration processing probably needs to be reviewed &/ refactored, to better cooperate with upstream vmim/vmi/vm objects, thanks.

brandboat:
> FYI, please also check issue https://github.com/harvester/harvester/issues/7180 and PR https://github.com/harvester/harvester/issues/7180 (WIP).
> 
> The current vm migration processing probably needs to be reviewed &/ refactored, to better cooperate with upstream vmim/vmi/vm objects, thanks.

Thanks for letting me know! I'll remove the "Pending Migration State" section from this HEP, as this will be addressed in https://github.com/harvester/harvester/issues/7180, this HEP will focus on exposing metrics to the dashboard.

enhancements/20241212-vm-migration-enhancement.md
```diff
@@ -0,0 +1,209 @@
...
+#### 2. Migration Metrics:
+
+KubeVirt's virt-handler provides a REST API endpoint (/metrics) to record VM migration metrics. Using the kubevirt-prometheus-metrics endpoint, we can resolve the IP addresses of virt-handler pods. Below is the procedure for retrieving and utilizing these metrics.
+
+**Retrieving virt-handler IPs**
+
+To resolve the virt-handler pod IPs:
+
+```sh
+kubectl get endpoints kubevirt-prometheus-metrics -n harvester-system -o yaml | yq eval '.subsets[].addresses[] | select(.targetRef.name | test("^virt-handler"))'
```
ihcsim: One thing to keep in mind is that this call doesn't provide us with the real time state of the virt-handler pods. If the pods are unhealthy or in the middle of scaling down, the subsequent HTTP call to its `/metrics` can fail. Perhaps consider asking the pods directly what their IPs are. It gives us the chance to assess the condition of the pods, before calling their `/metrics`.

E.g.,
```
$ k -n harvester-system get po -l"app.kubernetes.io/component=kubevirt" -l"kubevirt.io=virt-handler" -ojsonpath='{.items[*].status.podIP}'
172.16.2.34 172.16.1.79 172.16.0.111%                                   
```

enhancements/20241212-vm-migration-enhancement.md
```diff
@@ -0,0 +1,209 @@
...
+
+- [kubevirt_vmi_migration_data_remaining_bytes](https://github.com/kubevirt/monitoring/blob/main/docs/metrics.md#kubevirt_vmi_migration_data_remaining_bytes): The remaining guest OS data to be migrated to the new VM.
+
+- [kubevirt_vmi_migration_dirty_memory_rate_bytes](https://github.com/kubevirt/monitoring/blob/main/docs/metrics.md#kubevirt_vmi_migration_dirty_memory_rate_bytes): The rate of memory being dirty in the Guest OS.
+
+- [kubevirt_vmi_migration_disk_transfer_rate_bytes](https://github.com/kubevirt/monitoring/blob/main/docs/metrics.md#kubevirt_vmi_migration_disk_transfer_rate_bytes): The rate at which the memory is being transferred, unit is byte per second. Note that the name of this metrics is wrong and should be `kubevirt_vmi_migration_memory_transfer_rate_bytes` and will be fixed by https://github.com/kubevirt/kubevirt/pull/13500
+
+**Implementation Steps**
+
+We'll have a new controller `vmim_metrics_controller.go` to update metrics periodically to VirtualMachineInstanceMigration annotation, harvester dashboard can leverage on the annotation to show migration metrics on the UI, below are the implementationsteps:
```
ihcsim: When annotating the VMIM, we will likely need to utilize [`retry.RetryOnConflict()`](https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict) to avoid the infamous `Failed to update endpoint Operation cannot be fulfilled on endpoints the object has been modified; please apply your changes to the latest version and try again` error, to accommodate races with other controllers.

enhancements/20241212-vm-migration-enhancement.md
```diff
@@ -0,0 +1,209 @@
...
+      parallelMigrationsPerCluster: 5
+      parallelOutboundMigrationsPerNode: 2
+```
+For example, if a user attempts to migrate three VMs from harvester-node-1 to harvester-node-2, one of the migrations will be queued due to the parallelOutboundMigrationsPerNode limit of 2.
+
+**After**: Users can see a "Pending migration" state in the UI, clarifying that the migration is queued.
+
+#### Story 2
+
+**Before**: Users lack visibility into live migration metrics unless Prometheus is enabled, which can be resource-intensive.
```
ihcsim: Adding this as alternative consideration - it really depends on the out-of-box scrape interval and data retention settings of Prometheus. For users who already enabled monitoring (which I think they should for prod environments), adding another controller seems redundant. All else equal, there is no way our proposed controller can outperform Prometheus in terms of speed, resource efficiency and accuracy, especially since we are churning through this every minute.

Also, from the sound of it, the proposed controller is only reporting scalar values at every scrape interval, with no aggregation (to show metrics trend) possible. If user wants to see metrics trend (which is more useful when determining latency), they still end up needing to enable Prometheus.

brandboat: Thanks for the insightful comment! @ihcsim 

> For users who already enabled monitoring (which I think they should for prod environments), adding another controller seems redundant.

Initially, I heard that some users are running Harvester in edge environments with limited resources. That's why I chose to exclude Prometheus, as enabling such heavy services might be burdensome for them, especially during upgrades. However, I agree with you that enabling monitoring is always a good practice in production environments.

> the proposed controller is only reporting scalar values at every scrape interval, with no aggregation (to show metrics trend) 

You're right, this is a limitation of the current implementation. I've discussed the possibility of showing trend charts by tracking annotation changes in the dashboard with others, but that requires more consideration. The Harvester UI extension seems to lack the necessary libraries for this, and implementing it might be redundant, especially since we can simply enable monitoring addon and embed Grafana dashboards in the VM migration page.

Based on the two points above, perhaps we can allow users to enable the rancher-monitoring addon first if they want to see the VM live migration charts. We can then embed the corresponding VM live migration trend chart in the VM migration tab on the UI. But I'm unsure about this and would appreciate hearing more feedback. c.c. @bk201, @ibrokethecloud, @w13915984028

w13915984028: As @ihcsim has pointed, it is a bit complex and redundant to do those in a new controller directly.

The grafana panel/chart is a good candidate to show those data in UI.

For those without enabling monitoring stack or preferring API-calls, the vmi/vmim object has most of the detailed information, we may expand our documents to write down how to check and understand all of those params.

Kubevirt emits events e.g.
https://github.com/kubevirt/kubevirt/blob/950bb4b8c0ef317483e958dc2aceaaab4b85aca8/pkg/virt-controller/watch/migration/migration.go#L599
https://github.com/kubevirt/kubevirt/blob/950bb4b8c0ef317483e958dc2aceaaab4b85aca8/pkg/virt-controller/watch/migration/migration.go#L1044
 when migration state changes, Harvester vm page may monitor them.
 
 With the combination of them, a better user experience may come.

ihcsim: Do we know what's the minimum hardware spec for edge? The default prod installation already asked user for 16 CPU cores and 64 GB of memory, per node. I think that's more than enough to run Prometheus (with reasonable data retention and scrape interval settings). Using Grafana should be optional. As long as Prometheus is enabled, our dashboard can query it directly via the Prometheus JS client library.

I like the idea of just showing basic kubevirt events from the vmi/vmim if monitoring isn't enabled. Then if user wants more granular insights, they have to enable monitoring.

Also, I think it's going to be hard to track metric changes via annotations accurately and reliably. We're gonna end up with frequent update (retry) calls to the API server and I/Os to etcd. If we trigger enough churns, both client-go and the API server can block and throttle our API calls.

bk201: +1 to rely on the monitoring add-on. If the hardware doesn't meet the specs to run the monitoring stack, I probably will assume live migration is not a frequently-used operation.

brandboat: Thank you all, I'll redesign the HEP to address the comments mentioned above!

brandboat: Hi everyone, I just updated the HEP, please take a look when you are available, thank you!

ihcsim (dismissed):
LGTM - thanks :rocket:. Is it still possible to display some basic `status` info from the vmi/vmim objects if monitoring is not enabled?

brandboat:
>  Is it still possible to display some basic status info from the vmi/vmim objects if monitoring is not enabled?

Hi @ihcsim! Pardon me, did you mean vmim phase or the whole vmim status? vmim status contains lots of information, I'm not sure what's the best way to show the info to the UI.

ihcsim:
@brandboat Sorry, yeah, I meant the phase, per https://github.com/harvester/harvester/issues/4352#issuecomment-2533548608, and the kubevirt events per https://github.com/harvester/harvester/pull/7146#discussion_r1916307155. Thanks.

brandboat:
> @brandboat Sorry, yeah, I meant the phase, per https://github.com/harvester/harvester/issues/4352#issuecomment-2533548608, and the kubevirt events per https://github.com/harvester/harvester/pull/7146#discussion_r1916307155. Thanks.

@ihcsim Thanks for the explanation! I've updated the HEP slightly, adding vmim phase to the migration UI page.  Events were already included in the HEP, but this wasn't clear in the previous demo image. I've uploaded a new image to better highlight this.

ibrokethecloud (approved):

Issue #6835, opened 2024-10-21
[BUG] No validation on Maintenance Strategy when Edit as YAML


albinsun:
### Describe the bug
During validate #5069, we found that no validation on Maintenance Strategy when Edit as YAML.
And the VM with unexpected configuration will be created successfully...

> [!WARNING]
> If the wrongly configured VM locates on the first node _node-0_, **Enter Maintenance Mode** on _node-0_ will cause it stuck in `Cordoned`.
> See **Additional context**.

### To Reproduce
1. Setup a 3 nodes witness cluster
2. Create image `ubuntu-22.04-server-cloudimg-amd64.img`
3. Create network `mgmt-vlan1`
4. Virtual Machines => Create = Enter Name, CPU, Memory and Image => Edit as YAML => Add `harvesterhci.io/maintain-mode-strategy: xxx` in `metadata.labels` => Create
   ![image](https://github.com/user-attachments/assets/0017b6a8-b560-4348-a8e0-920eb780f619)
5. :x: Should FAIL to create VM cuz validation error
   * VM `Running` (Should NOT) and with empty **Maintenance Strategy**
     ![image](https://github.com/user-attachments/assets/9d80faf8-8b25-4168-bc94-665af595b18f)
   * Edit YAML
     ![image](https://github.com/user-attachments/assets/5468f541-2323-4c46-9c6c-a354790dd630)

### Expected behavior
Should FAIL to create VM cuz validation error since there are only 4 valid values:
1. `Migrate` (default)
2. `ShutdownAndRestartAfterEnable`
3. `ShutdownAndRestartAfterDisable`
4. `Shutdown`

### Support bundle
* Before **Enter Maintenance Mode** on _node-0_: [supportbundle_DoesNotValidateMaintenanceStrategy_2024-10-21T16-40-26Z.zip](https://github.com/user-attachments/files/17464387/supportbundle_DoesNotValidateMaintenanceStrategy_2024-10-21T16-40-26Z.zip)
* After **Enter Maintenance Mode** on _node-0_: [supportbundle_MigrateXXX_2024-10-21T17-05-00Z.zip](https://github.com/user-attachments/files/17464613/supportbundle_MigrateXXX_2024-10-21T17-05-00Z.zip)

### Environment
 - Harvester ISO version: `v1.4.0-rc3`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 nodes witness AMD64 QEMU/KVM includes 1 witness node.

[...]

PR #7160, opened 2024-12-13
Fix: Enforce consistent Maintenance Mode Strategy behavior

m-ildefons:
# Problem:

As described in harvester/harvester#6835, there are several problems to be considered here:

1. Ensuring that the maintenance-mode strategy is specified
2. Ensuring that only a valid maintenance-mode strategy can be specified
3. Ensuring that the specified maintenance-mode strategy label is propagated all the way to all resources where it's expected
4. Ensuring that default behavior is applied in all cases where a maintenance-mode strategy either isn't specified or isn't valid

# Solution:

The solution to the problem consists of a modification to the node drain helper and modifications both the mutating admission webhook and the validating admission webhook for the VirtualMachine resource

### The Node Drain Helper

The node drain helper was only marking Pods of VMs which had the maintenance-mode strategy label set to `Migrate` for deletion (i.e. migration off of the node).
This behavior would omit all Pods which didn't have this label set at all, or the ones which had it set to an invalid value. Similarly, Pods of VMs which didn't have one of the three other maintenance-mode strategies set would not be shut down. Thereby any VM without a maintenance-mode strategy, or with an invalid maintenance-mode strategy would be left running, preventing the node from entering the maintenance mode successfully. As a result the node would remain stuck in "Cordoned" state, waiting for an operator to intervene and deal with the VMs.
The new behavior of the drain helper is to migrate any VMs off of the node, if the associated Pod either has no maintenance-mode strategy label, the maintenance-mode strategy label is invalid, or the maintenance-mode strategy is specified as `Migrate`. This is the documented default behavior for VMs.
All other VMs will have a different maintenance-mode strategy specified and they will thus be shut down (and later restarted according to their maintenance-mode strategy).

### Admission Webhook

The admission webhook for the VirtualMachine resource plays a crucial role in making sure that VirtualMachiness have a sane maintenance-mode strategy specified.
First, the mutating webhook ensures that the maintenance-mode strategy label is set on the VirtualMachine resource. If it isn't, it will be set to `Migrate`, which is the default.
Then the mutating webhook ensures that the maintenance-mode strategy label that is set on the VirtualMachine resource is propagated into the `.spec.templates.metadata.labels`, which is to ensure that the label is propagated throughout both the VirtualMachineInstance as well as the Pod belonging to that VirtualMachine.
The mutating webhook will not modify exising maintenance-mode strategy labels on existing VirtualMachine resources. While the node drain helper will apply the default strategy to these virtual machines, this is to ensure that the user will not be faced with inexplicably changing maintenance-mode strategy labels, if they have an invalid value in their cluster.
The webhook will however log an appropriate message, allowing the user to identify VirtualMachine resources affected by invalid values in the maintenance-mode strategy label and correct it.

The validating webhook takes the role of a safe guard, ensuring that no VirtualMachine can be created or updated with an ill-specified maintenance-mode strategy.
It will reject the creation of VirtualMachine resources without a maintenance-mode strategy label or with an invalid value in the maintenance-mode strategy label.
This ensures that the label can not accidentally be filled with bogus data and that the user is informed.

# Related Issue:

- harvester/harvester#6835

# Test plan:

## Scenario 1

Ensure that new VirtualMachines have a valid maintenance-mode strategy. If it's not given, the default should be applied.

### Test 1

No maintenance-mode strategy given.

#### Setup:

Create a new VirtualMachine and specify no maintenance-mode strategy:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/os: linux
  name: test
[...]
```

#### Expected Result

The default strategy `Migrate` should have been added by the mutating admission webhook:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: Migrate
    [...]
  name: test
[...]
```

### Test 2

Valid maintenance-mode strategy given.

#### Setup:

Create a new VirtualMachine and specify a valid maintenance-mode strategy:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: Shutdown
    [...]
  name: test
[...]
```

#### Expected Result

The previously specified maintenance-mode strategy should be persisted in the VirtualMachine resource's metadata:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: Shutdown
    [...]
  name: test
[...]
```

## Scenario 2

If an invalid maintenance-mode strategy is specified, the VirtualMachine resource should be rejected.

### Test 1

Invalid maintenance-mode strategy given.

#### Setup:

Create a new VirtualMachine and specify an invalid maintenance-mode strategy:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: foobar
    [...]
  name: test
[...]
```

#### Expected Result

The creation should be rejected with a sensible error message

```
The request is invalid: metadata.labels[harvesterhci.io/maintain-mode-strategy]: invalid maintenance mode strategy: foobar
```

## Scenario 3

The specified maintenance-mode strategy should be propagated to all relevant resources.

### Test 1

The maintenance-mode strategy is propagated to the VirtualMachineInstance and the corresponding Pod, when the VirtualMachine is created

#### Setup

Create a new VirtualMachine and specify a valid maintenance-mode strategy:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: Shutdown
    [...]
  name: test
[...]
```

#### Expected Result

The VirtualMachine resource, the VirtualMachineInstance resource and the belonging Pod resource all have the specified maintenance-mode strategy label.

```
│ ~ │► kubectl get vm,vmi,pod -l harvesterhci.io/maintain-mode-strategy=Shutdown
NAME                              AGE   STATUS    READY
virtualmachine.kubevirt.io/test   10m   Running   True

NAME                                      AGE   PHASE     IP           NODENAME           READY
virtualmachineinstance.kubevirt.io/test   10m   Running   10.52.1.65   harvester-node-1   True

NAME                           READY   STATUS    RESTARTS   AGE
pod/virt-launcher-test-n5tf2   2/2     Running   0          10m
```

### Test 2

The maintenance-mode strategy is updated on the VirtualMachineInstance and on the corresponding Pod, when the maintenance-mode strategy is updated on the VirtualMachine resource.

//TODO

## Scenario 4

In case a VirtualMachine resource, VirtualMachineInstance resource, or a corresponding Pod exists and either has no maintenance-mode strategy label, or the maintenance-mode strategy label has an invalid value, the default behavior (`Migrate`) should be applied.

// TODO

### DIFF
--- a/pkg/controller/master/virtualmachine/vmi_controller.go
+++ b/pkg/controller/master/virtualmachine/vmi_controller.go
@@ -5,6 +5,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
@@ -14,11 +16,13 @@ import (
 
 	"github.com/harvester/harvester/pkg/builder"
 	kubevirtctrl "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/indexeres"
 	"github.com/harvester/harvester/pkg/util"
 )
 
 const (
 	VirtualMachineCreatorNodeDriver = "docker-machine-driver-harvester"
+	HarvesterLabelPrefix            = "harvesterhci.io"
 )
 
 // hostLabelsReconcileMapping defines the mapping for reconciliation of node labels to virtual machine instance annotations
@@ -27,12 +31,79 @@ var hostLabelsReconcileMapping = []string{
 }
 
 type VMIController struct {
+	podClient           ctlcorev1.PodClient
+	podCache            ctlcorev1.PodCache
 	vmClient            kubevirtctrl.VirtualMachineClient
 	virtualMachineCache kubevirtctrl.VirtualMachineCache
 	vmiClient           kubevirtctrl.VirtualMachineInstanceClient
 	nodeCache           ctlcorev1.NodeCache
 	pvcClient           ctlcorev1.PersistentVolumeClaimClient
 	recorder            record.EventRecorder
+}
+
+// SyncHarvesterVMILabelsToPod ensures that all Harvester labels (i.e. those
+// with the harvesterhci.io/ prefix) from the VirtualMachineInstance are synced
+// to the Pod
+func (h *VMIController) SyncHarvesterVMILabelsToPod(_ string, vmi *kubevirtv1.VirtualMachineInstance) (*kubevirtv1.VirtualMachineInstance, error) {
+	if vmi == nil || vmi.DeletionTimestamp != nil {
+		return vmi, nil
+	}
+
+	logrus.Debugf("Syncing labels %v for VMI %v to Pod", vmi.Labels, vmi.Name)
+
+	harvesterVMILabels := map[string]string{}
+	for label := range vmi.Labels {
+		if strings.HasPrefix(label, HarvesterLabelPrefix) {
+			harvesterVMILabels[label] = vmi.Labels[label]
+		}
+	}
+
+	activePods := vmi.Status.ActivePods
+	if len(activePods) < 1 {
+		logrus.Debugf("VMI %v does not have active Pods", vmi.Name)
+		return vmi, nil
+	}
+
+	nodeName := vmi.Status.NodeName
+	if nodeName == "" {
+		logrus.Debugf("Could not identify which node VMI %v runs on", vmi.Name)
+		return vmi, nil
+	}
+
+	pods, err := h.podCache.GetByIndex(indexeres.PodByNodeNameIndex, nodeName)
+	if err != nil {
+		return vmi, fmt.Errorf("failed to find pods for VMI %v, %v", vmi.Name, err)
+	}
+
+	var toUpdate = make([]*corev1.Pod, 0)
+	for _, pod := range pods {
+		for active := range activePods {
+			if pod.UID == active {
+				toUpdate = append(toUpdate, pod.DeepCopy())
+			}
+		}
+	}
+
+	for _, pod := range toUpdate {
+		// delete Harvester labels from Pod, if they are deleted from VMI
+		for podLabel := range pod.Labels {
+			_, ok := harvesterVMILabels[podLabel]
+			if strings.HasPrefix(podLabel, HarvesterLabelPrefix) && !ok {
+				delete(pod.Labels, podLabel)
+			}
+		}
+
+		// copy labels from VMI to Pod
+		for label := range harvesterVMILabels {
+			pod.Labels[label] = harvesterVMILabels[label]
+		}
+
+		_, err := h.podClient.Update(pod)
+		if err != nil {
+			return vmi, fmt.Errorf("failed to sync Harvester VMI labels to pod, %v", err)
+		}
+	}
+	return vmi, nil
 }
 
 // ReconcileFromHostLabels handles the propagation of metadata from node labels to VirtualMachineInstance annotations.
--- a/pkg/controller/master/virtualmachine/vmi_controller.go
+++ b/pkg/controller/master/virtualmachine/vmi_controller.go
@@ -2,7 +2,9 @@ package virtualmachine
 
 import (
 	"fmt"
+	"maps"
 	"reflect"
+	"slices"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -11,13 +13,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apitypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	"github.com/harvester/harvester/pkg/builder"
 	kubevirtctrl "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
-	"github.com/harvester/harvester/pkg/indexeres"
+	"github.com/harvester/harvester/pkg/ref"
 	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/util/indexeres"
 )
 
 const (
@@ -41,6 +45,17 @@ type VMIController struct {
 	recorder            record.EventRecorder
 }
 
+// Golang standard library's "maps" module contains the "Keys" function only for
+// Golang <v1.21 and >= v1.23
+// This function returns a sequence of keys of a map
+func mapKeys(m map[apitypes.UID]string) []apitypes.UID {
+	keys := make([]apitypes.UID, 0)
+	for k := range m {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
 // SyncHarvesterVMILabelsToPod ensures that all Harvester labels (i.e. those
 // with the harvesterhci.io/ prefix) from the VirtualMachineInstance are synced
 // to the Pod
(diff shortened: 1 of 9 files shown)


bk201 (dismissed):

w13915984028:
A question about the scenario:

Some VMs can‘t be migrated due to selector or dedicated devices. And sometimes user doesn’t want a VM to be migrated accidentally.

The original design respected this and only process it when a valid strategy is set and skip any unset/invalid marks.

m-ildefons:
> A question about the scenario:
> 
> Some VMs can‘t be migrated due to selector or dedicated devices. And sometimes user doesn’t want a VM to be migrated accidentally.
> 
> The original design respected this and only process it when a valid strategy is set and skip any unset/invalid marks.

In that case, the user should set one of the other maintenance-mode strategies. It is documented that VMs without a maintenance-mode strategy are migrated as well: https://docs.harvesterhci.io/v1.4/host/#node-maintenance
With the maintenance-mode label `ShutdownAndRestartAfterDisable`, they can even have the VMs be started again automatically, once the node is no longer in maintenance mode.

Users can also manually shut down VMs prior to enabling the maintenance mode, if they want to prevent migration.

I've re-worked the label synchronization between VMI resources and Pods, so now the maintenance-mode strategy can be updated without re-starting the VM as well.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -296,3 +307,41 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
 func (v *vmValidator) checkStorageResourceQuota(vm *kubevirtv1.VirtualMachine, oldVM *kubevirtv1.VirtualMachine) error {
 	return v.rqCalculator.CheckStorageResourceQuota(vm, oldVM)
 }
+
+func checkMaintenanceModeStrategyIsValid(newVM, oldVM *kubevirtv1.VirtualMachine) error {
+	labels := newVM.ObjectMeta.Labels
+	newStrategy, ok := labels[util.LabelMaintainModeStrategy]
+	if ok {
```
w13915984028: consider to use following to reduce the code nesting
```
if !ok {
  return ...
}
// continue with ok
```

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -35,6 +41,71 @@ type VMIController struct {
...
+		// delete Harvester labels from Pod, if they are deleted from VMI
+		for podLabel := range pod.Labels {
+			_, ok := harvesterVMILabels[podLabel]
+			if strings.HasPrefix(podLabel, HarvesterLabelPrefix) && !ok {
+				delete(pod.Labels, podLabel)
+			}
+		}
+
+		// copy labels from VMI to Pod
+		for label := range harvesterVMILabels {
```
w13915984028: Below is a bit faster
```
		for label, value := range harvesterVMILabels {
			pod.Labels[label] = value
		}
```

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -35,6 +41,71 @@ type VMIController struct {
...
+			pod.Labels[label] = harvesterVMILabels[label]
+		}
+
+		_, err := h.podClient.Update(pod)
```
w13915984028: vmi onChange happens many times, use deepEqual to check if pod is really changed then call update will be faster

the deepEqual runs in local go thread; the .Update runs on apiServer, the latter is much slower

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -540,3 +550,57 @@ func hostDevicesOvercommitNeeded(oldVM, newVM *kubevirtv1.VirtualMachine) bool {
...
+			},
+		)
+		if err != nil {
+			return patchOps, err
+		}
+
+		patchOps = append(patchOps, string(patch[:]))
+	}
+
+	// If the value under .spec.template.metadata.labels does not match the
```
w13915984028, line 579: we need to differentiate the `create` and `update` case:

for `create`, a wrong strategy should be rejected to tell user the clear error, then he will notice and adjust it;

for `update`, it may come from legacy vm, adjust the wrong value is user friendly

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -79,6 +79,11 @@ func (m *vmMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patch
 		return nil, err
 	}
 
+	patchOps, err = m.ensureMaintenanceModeStrategyIsSet(vm, patchOps)
```
w13915984028, line 82: as below comment, in `create` case, do not change user's input silently

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -35,6 +41,71 @@ type VMIController struct {
 	recorder            record.EventRecorder
 }
 
+// SyncHarvesterVMILabelsToPod ensures that all Harvester labels (i.e. those
+// with the harvesterhci.io/ prefix) from the VirtualMachineInstance are synced
+// to the Pod
+func (h *VMIController) SyncHarvesterVMILabelsToPod(_ string, vmi *kubevirtv1.VirtualMachineInstance) (*kubevirtv1.VirtualMachineInstance, error) {
```
w13915984028, line 62: this function may have performance issue

in which case, the strategy is not synced to to pod correctly? could we handle the specific one, instead of run this funcion on VMI change? VMI changes many times when a new VM is starting/migrating

w13915984028 (commented):
Please check those questions, thanks.

w13915984028:
btw, from 12.20, I will be offline and not review PRs.

please move the PR forward if it meets approval, do not wait for me, thanks.

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -35,6 +41,71 @@ type VMIController struct {
...
+	if vmi == nil || vmi.DeletionTimestamp != nil {
+		return vmi, nil
+	}
+
+	logrus.Debugf("Syncing labels %v for VMI %v to Pod", vmi.Labels, vmi.Name)
+
+	harvesterVMILabels := map[string]string{}
```
ibrokethecloud, line 69: we should skip the VMI if the maintenance strategy is `Migrate`. This will allow kubevirt to just handle the migration of VM's by watching for nodes being placed into maintenance mode via its own watchers. 
We do not need to do anything specific for virt-launcher pods in this case.

The additional benefit of this will be that we will not be patching the pod labels for every launcher pod. It is very likely that `Maintenance Stratergy` is only used by a subset of VM's among a large number of end user clusters. This greatly reduces the number of reconciles / update requests to the apiserver for the pods.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -296,3 +307,41 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
 func (v *vmValidator) checkStorageResourceQuota(vm *kubevirtv1.VirtualMachine, oldVM *kubevirtv1.VirtualMachine) error {
 	return v.rqCalculator.CheckStorageResourceQuota(vm, oldVM)
 }
+
+func checkMaintenanceModeStrategyIsValid(newVM, oldVM *kubevirtv1.VirtualMachine) error {
```
ibrokethecloud, line 311: we should fine tune this more to only ignore updates when oldVM and newVM migration strategy changes and the new strategy is not a valid strategy.

An old VM from say 1.4.0 which does not have the strategy applied as default will not have a strategy. Any update on this VM will be rejected by the validator, which may have unexpected results.

For example, I deployed the harvester / harvester-webhook to an existing 1.4.0 cluster to mimic the upgrade, and post upgrade, I still have VM's without the migration strategy setup.

As expected hot plug volume attachments are broken to this VM because the webhook is blocking the request.


```
node-1-dhcp:~ # virtctl addvolume test --volume-name=attachment --persist
You are using a client virtctl version that is different from the KubeVirt version running in the cluster
Client Version: v1.1.0
Server Version: v1.2.2
error adding volume, admission webhook "validator.harvesterhci.io" denied the request: label `harvesterhci.io/maintain-mode-strategy` not found
```

pkg/webhook/resources/virtualmachine/mutator.go
m-ildefons: in the `create` case, the user's input will never be changed. The mutator will only add the label if it doesn't exist.

pkg/webhook/resources/virtualmachine/mutator.go
m-ildefons: The mutator will not reject anything, since it's not its job.

The mutator will just make sure that the label is set (or it will set it itself to the default) and that the label in the `.spec.template` matches the label in `.metadata`.

If the label is set, but to an invalid value, the mutator will not change it, because this would be unexpected by the user.
However, the validator will later reject the operation, if the label is invalid.

pkg/webhook/resources/virtualmachine/validator.go
m-ildefons: Weird... seems like the mutating webhook didn't run, otherwise it would have added the label. But then the validating webhook was run.
I'm not sure if `virtctl addvolume` is an `update` operation or a patch.

pkg/controller/master/virtualmachine/vmi_controller.go
m-ildefons: This controller is not just for the maintenance-mode functionality. It syncs all labels with the `harvesterhci.io` prefix, which may come in handy later on, e.g. when we need to match Pods to their Rancher downstream clusters, or when we need to label Pods of node-pools, so load-balancers can identify them.

I'm also not a huge fan of implementing this functionality in a controller, but I don't see many other options.

pkg/controller/master/virtualmachine/vmi_controller.go
m-ildefons: I've made several changes that I think should improve the performance of this entire method.
Avoiding the `Update` as much as possible is one of them, but the `DeepCopy` also now only happens when the `Update` happens.
Additionally, I've replaced a few of the for-loops with functions from the standard library and I've gotten rid of the dynamically growing array.
The last big improvement in this department is that I'm now using a different Pod indexer, which only returns Pods associated with the VirtualMachine, not all Pods running on the node where the VirtualMachineInstance is running, so there are a lot less Pods to look through.

pkg/controller/master/virtualmachine/vmi_controller.go
m-ildefons: I haven't seen performance issues in my testing.
Regardless, I've applied the suggestions from the first round of review and made some more improvements to this method, which all should help with performance or at least robustness.

pkg/webhook/resources/virtualmachine/validator.go
m-ildefons: Good idea. Done

pkg/controller/master/virtualmachine/vmi_controller.go
votdev: Does it perhaps make sense to work on this functionality in a separate issue/PR as it is something that is not specific to the maintenance mode problem and has also required or triggered some discussion?

votdev:
I think the improvements to the `drainhelper` are good. The validator webhook is also quite good so that users get a feedback if they set a wrong value for the label, but the mutator is a bit too much of a good thing in my opinion. We have and use a lot of labels in Harvester and don't make such an effort for them. So if we make sure that the existing code does what it is supposed to do when the right values are in the label then fine, anything else is overengineering.

Issue #4893, opened 2023-12-20
[ENHANCEMENT] Examine for duplicated MAC addresses when creating new VMs in the Harvester cluster


starbops:
**Is your enhancement related to a problem? Please describe.**

Before v1.1.2, it was possible to create VMs with duplicated MAC addresses. We restricted the input for MAC addresses when creating VMs on the dashboard:

- For the single-instance option, we made the input field hidden by default. Users must click the **More** link to make it appear on the UI.
- For the multiple-instance option, we dropped the input field completely

However, the issue is only addressed on the frontend(see #3698); we still need to handle the input validation on the backend.

**Describe the solution you'd like**

Implement a validator on the webhook, iterating through the interfaces of all the VMs that exist in the cluster to see if there are duplicated MAC addresses every time a new VM is going to be created. If yes, reject the request; otherwise, accept the request.

**Describe alternatives you've considered**

**Additional context**

Martin-Weiss:
Please keep in mind that having duplicate MAC addresses is a valid scenario!

Duplicate MAC addresses can happen in case of restores (restore to a new VM name) - in this case the old VM is powered off and the new (restored VM) will have the same MAC address.

And in this scenario duplicate MAC addresses are even valid in case both VMs are running but are connected to different L2 networks. Usecase - restore VM to an isolated environment for testing etc.

The "only" problematic thing might be to run two VMs with identical MAC addresses in the same L2 network.

irishgordo:
@Martin-Weiss that's a great point :smile: 

_With:_
> The "only" problematic thing might be to run two VMs with identical MAC addresses in the same L2 network.

I would mention that for "some" network infrastructure/hardware - would likely get alerted / notified that it is happening - so Network Engineers (or other individuals) within an Org would be made aware from the switch/routing level that something is happening with the identical MAC addresses in the same L2 network.  Some systems do push notifications, emails, or other means (Slack, Teams, etc.).

Here's an example from my UniFi setup where I've had identical MACs running in VMs on the same L2 network (if I can recall correctly that this was happening when it did :sweat_smile: ) and the errors/notifications it would throw:

![Screenshot from 2023-12-20 10-07-31](https://github.com/harvester/harvester/assets/5370752/62202a8e-e491-447d-975c-a68d4f6c4ce8)

[...]

PR #7162, opened 2024-12-13
validate duplicate mac address during vm create/update

rrajendran17:
**Problem:**
VMs in the same L2 network cannot have same mac addresses as they cause network issues.

**Solution:**
Avoid creating VMs with same mac addresses

**Related Issue:**
https://github.com/harvester/harvester/issues/4893

**Test plan:**
Added unit tests for the following scenarios

1.Create the following VLAN VM Networks
 Name    namespace      vlan-id          cluster-network

 NAD1   default            vlan-1           cluster-1
 NAD2   default            vlan-2           cluster-1
 NAD3  default             vlan-3           cluster-1
 NAD4  non-default      vlan-1           cluster-1
 NAD5  non-default      vlan-2           cluster-2

2.create the following VMs attaching the VLAN VM Networks as VM's interface
 Name   NAD    MAC

  vm1     NAD1   MAC-1
  vm2     NAD2   MAC-2
  vm3     NAD3   MAC-3
  
3.create the following VMs
  a.different mac in same cluster network returns success
    - create vm and attach to NAD1 with mac MAC-2 
  b.duplicate mac in same cluster-network returns error
    - attach vm to NAD1 with mac MAC-1
  c.duplicate mac in same L2 with nad in different namespace returns error
    - attach vm to NAD4 with mac MAC-1
  d.duplicate mac in different L2 with nad in different namespace returns success
    - attach vm to NAD5 with mac MAC-1

4.Create any VMs and try to migrate them and it should succeed
5.Create any VMs and Restoring vms from backup to existing VMs should succeed.
6.Create any VMs and Restoring vms from backup to new VMs should return error

7.Creating multiple instances of vm using multiple instance template  should succeed.

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
+	"strings"
 
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
@@ -14,6 +16,7 @@ import (
 
 	ctlharvestercorev1 "github.com/harvester/harvester/pkg/generated/controllers/core/v1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
 	"github.com/harvester/harvester/pkg/ref"
 	"github.com/harvester/harvester/pkg/util"
@@ -21,6 +24,7 @@ import (
 	"github.com/harvester/harvester/pkg/util/resourcequota"
 	vmUtil "github.com/harvester/harvester/pkg/util/virtualmachine"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
+	indexerwebhook "github.com/harvester/harvester/pkg/webhook/indexeres"
 	"github.com/harvester/harvester/pkg/webhook/types"
 	webhookutil "github.com/harvester/harvester/pkg/webhook/util"
 )
@@ -34,12 +38,14 @@ func NewValidator(
 	vmimCache ctlkubevirtv1.VirtualMachineInstanceMigrationCache,
 	vmCache ctlkubevirtv1.VirtualMachineCache,
 	vmiCache ctlkubevirtv1.VirtualMachineInstanceCache,
+	nadCache ctlcniv1.NetworkAttachmentDefinitionCache,
 ) types.Validator {
 	return &vmValidator{
 		pvcCache:      pvcCache,
 		vmBackupCache: vmBackupCache,
 		vmCache:       vmCache,
 		vmiCache:      vmiCache,
+		nadCache:      nadCache,
 
 		rqCalculator: resourcequota.NewCalculator(nsCache, podCache, rqCache, vmimCache),
 	}
@@ -51,6 +57,7 @@ type vmValidator struct {
 	vmBackupCache ctlharvesterv1.VirtualMachineBackupCache
 	vmCache       ctlkubevirtv1.VirtualMachineCache
 	vmiCache      ctlkubevirtv1.VirtualMachineInstanceCache
+	nadCache      ctlcniv1.NetworkAttachmentDefinitionCache
 
 	rqCalculator *resourcequota.Calculator
 }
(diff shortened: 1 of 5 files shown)


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -69,12 +75,134 @@ func (v *vmValidator) Resource() types.Resource {
...
+
+	oldvmsInfo, err = v.getVMInterfaceInfo(oldvmsList)
+	if err != nil {
+		return err
+	}
+
+	for cn, macs := range newVMInfo {
+		if othermacs, exists := oldvmsInfo[cn]; exists {
+			if comparemacs(macs, othermacs) { //duplicate present
+				return fmt.Errorf("duplicate mac address present for vm %s for cluster network %s", vm.Name, cn)
```
w13915984028, line 194: just a question, will vm migration case hit this check? thanks.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -69,12 +75,134 @@ func (v *vmValidator) Resource() types.Resource {
...
+	}
+
+	if vm.Spec.Template == nil {
+		return fmt.Errorf("vm %s template is nil", vm.Name)
+	}
+
+	vmInterfaces := vm.Spec.Template.Spec.Domain.Devices.Interfaces
+
+	for _, vmInterface := range vmInterfaces {
+		oldvms, err := v.vmCache.GetByIndex(indexerwebhook.VMByMacAddress, vmInterface.MacAddress)
```
w13915984028: another question: the index VMByMacAddress is composed of MAC list, and use a single `vmInterface.MacAddress` to get, will it return?

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -69,12 +75,134 @@ func (v *vmValidator) Resource() types.Resource {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func comparemacs(macList1 map[string]string, macList2 map[string]string) bool {
+	for mac1, newvmIntfName := range macList1 {
+		if oldvmIntfName, exists := macList2[mac1]; exists && oldvmIntfName != newvmIntfName { //skip same vms during vm migration,restore cases
```
w13915984028: seems if `oldvmIntfName == newvmIntfName` indicates that the vms are from migration/restore cases,how is this ensured?

suppose use one vm's yaml to create another vm, and the only change is vmname, will above check identify it? thanks.

w13915984028 (commented):
a few questions, thanks.

pkg/webhook/resources/virtualmachine/validator.go
rrajendran17: No, there is a check for same vm name and vminterfacename to avoid hitting this during migration and restore cases.

pkg/webhook/resources/virtualmachine/validator.go
rrajendran17: vmByMacaddrs groups the vms based on mac address with mac address as key/index.
When user tries to give key as any mac address, the output returns list of vms matching that key if it matches or empty vm list.

pkg/webhook/resources/virtualmachine/validator.go
rrajendran17: yeah oldvmIntfName/newvmIntfName is a combination of both interface name and vm name.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -92,6 +220,10 @@ func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj run
 		return nil
 	}
 
+	if err := v.checkForDuplicateMacAddrs(newVM); err != nil {
```
w13915984028: the check of duplicated MAC is a heavy operation, and Harvester other controllers may update vm object other fileds like https://github.com/harvester/harvester/pull/7195 is doing, could you add an additional condition, only when VM's interface / MAC address are updating/adding upon the oldVM, then call this check, in most cases, there is no MAC change, thanks.

w13915984028 (approved):
LGTM, please check an additional question, thanks.

pkg/webhook/resources/virtualmachine/validator.go
rrajendran17: Thanks, done

rrajendran17:
@w13915984028 @starbops @mingshuoqiu Gentle Reminder.
The review comments are taken care.Please review whenever you get some time,Thanks

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -69,12 +75,148 @@ func (v *vmValidator) Resource() types.Resource {
...
+		}
+	}
+
+	return MacAddrPerCluster, nil
+}
+
+func (v *vmValidator) checkForDuplicateMacAddrs(vm *kubevirtv1.VirtualMachine) (err error) {
+	var newvm []*kubevirtv1.VirtualMachine
+	var oldvmsInfo map[string]map[string]string
+	var oldvmsList []*kubevirtv1.VirtualMachine
```
starbops:
```suggestion
	var newVM []*kubevirtv1.VirtualMachine
	var oldVMsInfo map[string]map[string]string
	var oldVMsList []*kubevirtv1.VirtualMachine
```

mingshuoqiu (approved):
LGTM. Thanks

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -69,12 +75,148 @@ func (v *vmValidator) Resource() types.Resource {
...
+	newvm = append(newvm, vm)
+	newVMInfo, err := v.getVMInterfaceInfo(newvm)
+	if err != nil {
+		return err
+	}
+
+	//not a mac update,skip duplicate mac address check
+	if len(newVMInfo) == 0 {
+		return nil
+	}
```
starbops: I'm not sure why `len(newVMInfo) == 0` implies "not a mac update". Could you elaborate on this?

rrajendran17: The idea behind this check is, skip further processing if the vm operation is not for mac update.
getVMInterfaceInfo() will populate the MacAddrPerCluster map only if any of the VM interface has a mac address configured.If its empty, the loop is continued and MacAddrPerCluster will also be empty.
getVMInterfaceInfo() will return empty map in this case and checking the len(newVMInfo) will tell us its not a vm operation for mac update rather some other vm operation.

This will make sure we are not validating for duplicate mac addresses if  its not a vm operation for mac configuration.
Please let me know if it makes sense.

starbops: I just feel the comment a bit weird because, IIRC, for the update requests, the received `oldObj` and `newObj` objects in the `Update()` function contain the full object content. Even the user does not update the MAC address, the object is populated with the original MAC address. So, under regular "update" cases, this chunk of short-circuit code is not that useful, the `newVMInfo` will almost always have element in it. But this can be useful when the user first "create" the virtual machine.

I'm not sure I fully catch your point. If I still understand it wrongly, please correct me. Thank you.

rrajendran17: Ok, I think I got your point now. During update case of VM, there could be other existing interfaces in VM with mac addresses and len(newVMInfo) will not be empty in this case even if there is no "new mac addresses configured by user".
Yeah looks like that line is not useful during vm update if the intention is to skip the processing for vm updates other than mac.

Just wondering if there is any trivial way to make sure  "duplicate mac validation" happens only if there is new mac configured by user.

We could compare oldobj and newobj during update for what has changed but its again looping through interfaces of both the sets.
Introducing a flag in UI for mac update would be easier to compare.

starbops: Maybe you can use `DeepEqual` to check whether there's any MAC address change between the new and old objects. For instance: https://github.com/harvester/harvester/blob/9609b9038f3724e584cfc432aaedd0274b6d8cc8/pkg/webhook/resources/schedulevmbackup/validator.go#L174-L176
(Thanks for @WebberHuang1118's insights)

rrajendran17: I think still we will not be able to use deepEqual in this case, as deepEqual of slices of Interfaces could be sorted and compared for equality, but there could be possibility that other params in Interface struct could be updated by user.
So my understanding is unless we compare each interface mac, we cannot skip this just for mac update.

So, I think its ok to have len(newVMsInfo) == 0, which is for create case and for update, its fine to compare the config with existing VM cache for duplicate mac.
Please let me know what you think about this ?

starbops: I didn't think of the update-with-only-nic-reordering cases. Then it makes sense to rely on sophisticated logic to filter the requests. But still, the relation between the comment and the piece of logic is not that intuitive. Could you add more context to the comment to lower the maintenance burden? Thank you very much.

starbops (approved):
Except for the suggested comment enhancement, others LGTM. Thank you for the great work.

pkg/webhook/resources/virtualmachine/validator.go
rrajendran17: Thanks @starbops I have rephrased the comment.Please let me if that works.

w13915984028: In update path, 

add
```
if oldvm.Spec.Template != nil && newvm.Spec.Template != nil && reflect.DeepEqual(oldvm.Spec.Template.Spec.Domain.Devices.Interfaces, newvm.Spec.Template.Spec.Domain.Devices.Interfaces) {
	return nil
}

before:
	if err := v.checkForDuplicateMacAddrs(vm); err != nil {
		return err
	}
```

still makes sense.

(1) In most VM update events (note: one VM object can be triggered many times of OnChange), the interfaces do not change, it helps to jump the MAC comparing.
(2) Without a more precise comparing of only MAC related filed, comparing `oldvm.Spec.Template.Spec.Domain.Devices.Interfaces` is meaningful either to improve performances.
(3) If unluckily the slices orders of `oldvm.Spec.Template.Spec.Domain.Devices.Interfaces` and `newvm.Spec.Template.Spec.Domain.Devices.Interfaces` are different but there are no real changes; or other params are changed, the following call of `checkForDuplicateMacAddrs` will finally figure out that there is no change about MAC.

This is the rational of my previous comment: https://github.com/harvester/harvester/pull/7162#discussion_r1893597191

rrajendran17: Sure, got it, thanks

rrajendran17: I have updated the PR addressing the comment.

rrajendran17:
Can I merge this if everything is good ? @w13915984028 @starbops

rrajendran17:
@Mergifyio backport v1.5

PR #7170, opened 2024-12-16
VM Image Controller, Webhook and Download/Upload Refactor for Various Backend Support

WebberHuang1118:
**Code Structure:**
- Backend interfaces:
  - `pkg/image/backend`
- Common VM Image operation package
  - `pkg/image/common`
- Backing image implementation
  - `pkg/image/backingimage`
- CDI implementation skeleton
  - pkg/image/cdi

**Problem:**
VM Image should support various source backend

**Solution:**
Enhance VM Image implementation to have a clear interface for different image backend

**Related Issue:**
#6936

**HEP:**
https://github.com/harvester/harvester/pull/7273

**Test plan:**
- Prerequisites
  - Building Harvester Cluster with the master-head branch ( `0cea310be669040bc682c75edd40fd6f81051bce` in my env) 
  - Updating `VirtualMachineImage` CRD
    - `$ kubectl apply -f deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml`
  - Updating `harvester` deployment with this PR
  - Updating `harvester-webhook` deployment with this PR
- Case 1:  Creating VM image with upload
  - Uploading an OS image file from local
  - Checking the VM image could progress to the completed status
  - Creating a VM with the VM image
- Case 2:  Creating VM image with download
  - Downloading an OS image file from a certain URL
  - Checking the VM image could progress to the completed status
  - Creating a VM with the VM image
- Case 3:  Creating VM image with export
  - Creating a volume with an existing OS VM Image
  - Exporting the volume as a new image
  - Checking the exported VM image could progress to the completed status
  - Creating a VM with the VM image
- Case 4:  VM Image encryption
  - Creating a [secret and sc for encryption](https://docs.harvesterhci.io/v1.4/image/image-security)
  - Creating an encrypted VM image from an existing VM image (non-encrypted)
  - Checking the encrypted VM image could progress to the completed status
  - Creating a VM with the VM image
- Case 5:  VM Image decryption
  - Continuing from the above steps 
  - Creating a VM image by decrypting the encrypted VM image
  - Checking the decrypted VM image could progress to the completed status
  - Creating a VM with the VM image
- Case 6:  VM Image Downloading
  - Triggering downloading an existing VM image
  - The VM image should downloaded as a compressed file

** Additional Context **
Since the VM image implementation has the new interfaces, the original unit testing is not applicable and removed. Creating a separate ticket to track the unit testing implementation #7204


pkg/image/backingimage/upload.go
```diff
@@ -0,0 +1,113 @@
...
+	var err error
+	defer func() {
+		if err != nil {
+			if updateErr := biu.vmio.FailUpload(vmi, err.Error()); updateErr != nil {
+				logrus.Error(err)
+			}
+		}
+	}()
+
+	//Wait for backing image data source to be ready. Otherwise the upload request will fail.
```
brandboat: nit:
```suggestion
	// Wait for backing image data source to be ready. Otherwise the upload request will fail.
```

pkg/image/common/operator.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+		if current.DeletionTimestamp != nil {
+			return nil
+		}
+		newVMI := current.DeepCopy()
+		harvesterv1.ImageImported.False(newVMI)
+		harvesterv1.ImageImported.Reason(newVMI, uploadFailReason)
+		harvesterv1.ImageImported.Message(newVMI, msg)
+		_, err = vmio.UpdateVMI(old, newVMI)
+		if err == nil || apierrors.IsConflict(err) {
```
brandboat: Is this a bug fix? I ask this since the original code is `if err == nil || !apierrors.IsConflict(err) {`

pkg/image/backingimage/upload.go
```diff
@@ -0,0 +1,113 @@
...
+		// Trim the "POST http://xxx" implementation detail for the error
+		// set the err var and it will be recorded in image condition in the defer function
+		err = errors.Unwrap(urlErr)
+		return err
+	} else if err != nil {
+		return fmt.Errorf("failed to send the upload request: %w", err)
+	}
+	defer uploadResp.Body.Close()
+
+	body, err := ioutil.ReadAll(uploadResp.Body)
```
brandboat: nit: ioutil.ReadAll is deprecated, could you please switch to `io.ReadAll`?
```suggestion
	body, err := io.ReadAll(uploadResp.Body)
```

brandboat (commented):
Thanks for the great refactor!

pkg/image/backingimage/upload.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/image/common/operator.go
WebberHuang1118: It's a miss, updated per suggestion, thanks for pointing it out.

pkg/image/backingimage/upload.go
WebberHuang1118: Updated per suggestion, thanks.

Yu-Jack (commented):
Nice try! There is a suggestion for folder structure:

----

I think our folder structure should match what it's , and `backend` might be a confused naming. 

For example, you could move `pkg/image/backend/backend.go` to `pkg/master/controller/image/interface.go`, and rename the interface name like `CSI` or `CSIInterface` etc.

Then we could use `import pkg/master/controller/image`, and use `image.CSIInterface` to present the meaning instead of `backend.Backend`.

The `pkg/image/backingimage/backing_image_controller.go` should be in `pkg/master/controller`. Spreading the controller on different places might not match our current folder structure, including mutator and validator.

But, I guess that you think putting them on same folder is more easier for finding and looking , So, there is another way to organize them. Create a new folder to present our core (real business) logic. That means we could put similar feature into this folder in the future if we need.

```
core
└── image
    ├── backingimage
    │   ├── backing_image_controller.go
    │   ├── backingimage.go
    │   ├── download.go
    │   ├── mutate.go
    │   ├── upload.go
    │   └── validate.go
    ├── cdi
    │   ├── cdi.go
    │   ├── download.go
    │   ├── mutate.go
    │   ├── upload.go
    │   └── validate.go
    ├── common
    │   ├── mutator.go
    │   ├── operator.go
    │   └── validator.god
    └── interface.go
```

However, I still think `backing_image_controller.go` should be `master/controller`, and  keep mutator.go and validator.go in original folder. We can wait for other people's feedbacks.

WebberHuang1118:
Thanks for the feedbacks @Yu-Jack 

> Nice try! There is a suggestion for folder structure:
> 
> I think our folder structure should match what it's , and `backend` might be a confused naming.
> 
> For example, you could move `pkg/image/backend/backend.go` to `pkg/master/controller/image/interface.go`, and rename the interface name like `CSI` or `CSIInterface` etc.
> 
> Then we could use `import pkg/master/controller/image`, and use `image.CSIInterface` to present the meaning instead of `backend.Backend`.
> 

It's a good idea, I will update it.


> The `pkg/image/backingimage/backing_image_controller.go` should be in `pkg/master/controller`. Spreading the controller on different places might not match our current folder structure, including mutator and validator.
> 

Because `pkg/image/backingimage/backing_image_controller.go` implements the sidecar handler for LH image, which is specific to LH backingimage, Harvester may support all the image features by CDI someday. IMO, placing the sidecar handler implementation at the general path `pkg/master/controller` seems not suitable.


> But, I guess that you think putting them on same folder is more easier for finding and looking , So, there is another way to organize them. Create a new folder to present our core (real business) logic. That means we could put similar feature into this folder in the future if we need.
> 
> ```
> core
> └── image
>     ├── backingimage
>     │   ├── backing_image_controller.go
>     │   ├── backingimage.go
>     │   ├── download.go
>     │   ├── mutate.go
>     │   ├── upload.go
>     │   └── validate.go
>     ├── cdi
>     │   ├── cdi.go
>     │   ├── download.go
>     │   ├── mutate.go
>     │   ├── upload.go
>     │   └── validate.go
>     ├── common
>     │   ├── mutator.go
>     │   ├── operator.go
>     │   └── validator.god
>     └── interface.go
> ```

I will consider it, thanks.


> 
> However, I still think `backing_image_controller.go` should be `master/controller`, and keep mutator.go and validator.go in original folder. We can wait for other people's feedbacks.

The VM image mutator and validator are still provided by the original path `pkg/webhook/resources/virtualmachineimage/mutator.go` and `pkg/webhook/resources/virtualmachineimage/validator.go`. 

`pkg/image/backingimage/mutate.go` and `pkg/image/backingimage/validate.go` aim to provide the common library for VM Image mutation and validation.

Yu-Jack:
> Because pkg/image/backingimage/backing_image_controller.go implements the sidecar handler for LH image, which is specific to LH backingimage, Harvester may support all the image features by CDI someday. IMO, placing the sidecar handler implementation at the general path pkg/master/controller seems not suitable.

You could organize like this. It's just an example. 

```
controller/master/image
├── register.go
├── third-party
│   ├── backingimage
│   │   ├── backing_image_controller.go
│   │   ├── backingimage.go
│   │   ├── download.go
│   │   ├── mutate.go
│   │   ├── upload.go
│   │   └── validate.go
│   └── cdi
│       ├── cdi.go
│       ├── download.go
│       ├── mutate.go
│       ├── upload.go
│       └── validate.go
└── vm_image_controller.go

```

> The VM image mutator and validator are still provided by the original path pkg/webhook/resources/virtualmachineimage/mutator.go and pkg/webhook/resources/virtualmachineimage/validator.go.
> 
> pkg/image/backingimage/mutate.go and pkg/image/backingimage/validate.go aim to provide the common library for VM Image mutation and validation.

Yeah, I also treat `pkg/webhook/resources/virtualmachineimage/mutator.go` and `pkg/webhook/resources/virtualmachineimage/validator.go` as a wrapper to wrap the core business logic. But, you could still do similar things like above to organize them if we're willing to do so. The point I'd like to highlight is the folder structure. Should we put all related business logic into same folder or organize by validator, mutator or controller like this? I just want to provide different perspectives to this topic.

```
webhook/resources/virtualmachineimage
├── mutator.go
├── third-party
│   ├── backingimage
│   │   ├── mutate.go
│   │   └── validate.go
│   └── cdi
│       ├── cdi.go
│       ├── mutate.go
│       └── validate.go
└── validator.go

controller/master/image
├── register.go
├── third-party
│   ├── backingimage
│   │   ├── backing_image_controller.go
│   │   └── backingimage.go
│   └── cdi
│       └── cdi.go
└── vm_image_controller.go

api/image
├── formatter.go
├── schema.go
└── third-party
    ├── backingimage
    │   ├── download.go
    │   └── upload.go
    └── cdi
        ├── cdi.go
        ├── download.go
        └── upload.go
```

WebberHuang1118:
> > Because pkg/image/backingimage/backing_image_controller.go implements the sidecar handler for LH image, which is specific to LH backingimage, Harvester may support all the image features by CDI someday. IMO, placing the sidecar handler implementation at the general path pkg/master/controller seems not suitable.
> 
> You could organize like this. It's just an example.
> 
> ```
> controller/master/image
> ├── register.go
> ├── third-party
> │   ├── backingimage
> │   │   ├── backing_image_controller.go
> │   │   ├── backingimage.go
> │   │   ├── download.go
> │   │   ├── mutate.go
> │   │   ├── upload.go
> │   │   └── validate.go
> │   └── cdi
> │       ├── cdi.go
> │       ├── download.go
> │       ├── mutate.go
> │       ├── upload.go
> │       └── validate.go
> └── vm_image_controller.go
> ```
> 
> > The VM image mutator and validator are still provided by the original path pkg/webhook/resources/virtualmachineimage/mutator.go and pkg/webhook/resources/virtualmachineimage/validator.go.
> > pkg/image/backingimage/mutate.go and pkg/image/backingimage/validate.go aim to provide the common library for VM Image mutation and validation.
> 
> Yeah, I also treat `pkg/webhook/resources/virtualmachineimage/mutator.go` and `pkg/webhook/resources/virtualmachineimage/validator.go` as a wrapper to wrap the core business logic. But, you could still do similar things like above to organize them if we're willing to do so. The point I'd like to highlight is the folder structure. Should we put all related business logic into same folder or organize by validator, mutator or controller like this? I just want to provide different perspectives to this topic.
> 
> ```
> webhook/resources/virtualmachineimage
> ├── mutator.go
> ├── third-party
> │   ├── backingimage
> │   │   ├── mutate.go
> │   │   └── validate.go
> │   └── cdi
> │       ├── cdi.go
> │       ├── mutate.go
> │       └── validate.go
> └── validator.go
> 
> controller/master/image
> ├── register.go
> ├── third-party
> │   ├── backingimage
> │   │   ├── backing_image_controller.go
> │   │   └── backingimage.go
> │   └── cdi
> │       └── cdi.go
> └── vm_image_controller.go
> 
> api/image
> ├── formatter.go
> ├── schema.go
> └── third-party
>     ├── backingimage
>     │   ├── download.go
>     │   └── upload.go
>     └── cdi
>         ├── cdi.go
>         ├── download.go
>         └── upload.go
> ```

My idea is if anyone wants to add another image solution (e.q. LH v2 image), we should make the task as easy as possible, so I encapsulate the details for VM Image, and it's provided as an interface/library.

By the current directory structure:
```
pkg/image/
├── backend
│   └── backend.go
├── backingimage
│   ├── backing_image_controller.go
│   ├── backingimage.go
│   ├── download.go
│   ├── mutate.go
│   ├── upload.go
│   └── validate.go
├── cdi
│   ├── cdi.go
│   ├── download.go
│   ├── mutate.go
│   ├── upload.go
│   └── validate.go
└── common
    ├── mutator.go
    ├── operator.go
    └── validator.go
```
If there is a requirement for new image solution, ideally, it only needs to add implementations under `pkg/image`. Compared to the structure:

 ```
 webhook/resources/virtualmachineimage
 ├── mutator.go
 ├── third-party
 │   ├── backingimage
 │   │   ├── mutate.go
 │   │   └── validate.go
 │   └── cdi
 │       ├── cdi.go
 │       ├── mutate.go
 │       └── validate.go
 └── validator.go
 
 controller/master/image
 ├── register.go
 ├── third-party
 │   ├── backingimage
 │   │   ├── backing_image_controller.go
 │   │   └── backingimage.go
 │   └── cdi
 │       └── cdi.go
 └── vm_image_controller.go
 
 api/image
 ├── formatter.go
 ├── schema.go
 └── third-party
     ├── backingimage
     │   ├── download.go
     │   └── upload.go
     └── cdi
         ├── cdi.go
         ├── download.go
         └── upload.go

```
the new implementation will be spreaded among different places.


> Should we put all related business logic into same folder or organize by validator, mutator or controller like this? I just want to provide different perspectives to this topic.

IMO, the business logic is still controlled by the original way, I mean:
- Controller: `pkg/controller/master/image/vm_image_controller.go`
- Webhook: `pkg/webhook/resources/virtualmachineimage/mutator.go` and `pkg/webhook/resources/virtualmachineimage/validator.go`
- API handler: `pkg/api/image/formatter.go`

Source codes under `pkg/image` are provided for the implementation of different image solutions, and are passively invoked by the original control path.

I can understand your concern, I think we can wait for other colleagues' feedback @bk201 @brandboat @Vicente-Cheng @FrankYang0529

File: pkg/image/common/operator.go [hunk 4]
```diff
@@ -0,0 +1,336 @@
...
+	uploadFailReason = "UploadFailed"
+)
+
+var (
+	ErrRetryable  = errors.New("retryable error")
+	ErrRetryLater = errors.New("retry later error")
+)
+
+func IsRetryable(err error) bool {
+	return errors.Is(err, ErrRetryable)
```
Vicente-Cheng: Could we use the IsRetryable() to check the error lists for retry capability instead of the new error `ErrRetryable`.
The new `ErrRetryable` might confuse.

e.g.
```
func IsRetryable(err error) bool {
    return apierrors.IsNotFound(err) || <something retryable error> || errors.Is(err, ErrRetryable)
}
```

brandboat:
I believe what Jack is trying to highlight is the importance of consistency. It seems that all modules currently follow the same pattern, with mutators and validators placed under the webhook directory, and controllers under the controller directory. Personally, I lean towards keeping things as they are, which means sticking to this pattern. 

However, the drawback is clear— as you mentioned, the backing image code is scattered across multiple locations. I believe this is a separate issue and might be worth opening a new discussion or issue. Additionally, as we plan to support third-party storage in more features, I believe this problem may not be limited to VM images and could affect other areas (like vm backup) as well.

WebberHuang1118:
> I believe what Jack is trying to highlight is the importance of consistency. It seems that all modules currently follow the same pattern, with mutators and validators placed under the webhook directory, and controllers under the controller directory. Personally, I lean towards keeping things as they are, which means sticking to this pattern.
> 
> However, the drawback is clear— as you mentioned, the backing image code is scattered across multiple locations. I believe this is a separate issue and might be worth opening a new discussion or issue. Additionally, as we plan to support third-party storage in more features, I believe this problem may not be limited to VM images and could affect other areas (like vm backup) as well.

Hi @brandboat,

Thanks for your feedback, however, I want to point out, that the pattern structure has not changed:
- Controller: pkg/controller/master/image/vm_image_controller.go
- Webhook: pkg/webhook/resources/virtualmachineimage/mutator.go and pkg/webhook/resources/virtualmachineimage/validator.go
- API handler: pkg/api/image/formatter.go

the control paths for the controller, mutator, validator, and API handlers are organized as before, they just access an abstracted layer/interface/implementation from `pkg/image/`, which is passively invoked.

IMO, if anyone wants to add a new image solution, he doesn't need to care about the control flow for the controller, webhook, API handler, and any details about CRD VMImage (VMImage operations are encapsulated in `pkg/image/common`), he just needs to focus on the interfaces exported in `pkg/image/`. This seems simpler to me.

brandboat:
> the control paths for the controller, mutator, validator, and API handlers are organized as before, they just access an abstracted layer/interface/implementation from pkg/image/, which is passively invoked.
> 
> IMO, if anyone wants to add a new image solution, he doesn't need to care about the control flow for the controller, webhook, API handler, and any details about CRD VMImage, he just needs to focus on the interfaces exported in pkg/image/. This seems simpler for me.

OK, that makes sense to me. Thanks for the clarification!

pkg/image/backingimage/backing_image_controller.go
```diff
@@ -0,0 +1,73 @@
...
+	if !h.vmio.IsInitialized(vmi) {
+		return nil, nil
+	}
+
+	// If ImageImported is not unknown, it means the backing image has
+	// been imported, and we think we know everything about it, i.e. we've
+	// now been through a series of progress updates during image download,
+	// and those are finally done, so let's not worry about further updates.
+	// TODO: Improve image to keep sync with LH backing image #6936
+	if h.vmio.IsImported(vmi) {
```
brandboat, line 49: The code doesn't match the comment as above, the comment says `If ImageImported is not unknown`, while this line is if ImageImported is true.

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -83,361 +61,78 @@ func (h *vmImageHandler) OnChanged(_ string, image *harvesterv1.VirtualMachineIm
...
 	}
 
-	if sc != nil && sc.DeletionTimestamp != nil {
-		h.imageController.EnqueueAfter(image.Namespace, image.Name, checkInterval)
-		return image, nil
+	if err != nil {
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"namespace": vmi.Namespace,
+			"name":      vmi.Name,
+		}).Error(err.Error())
```
brandboat: nit: perhaps change the error message in `.Error(err.Error())` to something else ? Two same err show up in the log seems a bit weird.

pkg/image/common/operator.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	return vmio.client.Update(newVMI)
+}
+
+func (vmio *vmiOperator) GetName(vmi *harvesterv1.VirtualMachineImage) string {
+	return vmi.Name
+}
+
+func (vmio *vmiOperator) GetNamespace(vmi *harvesterv1.VirtualMachineImage) string {
```
brandboat, line 105: It might be just my habit of taste, but I feel having a bunch of getters here is a bit overkill. Or do we need this for purpose?

pkg/webhook/resources/virtualmachineimage/mutator.go
```diff
@@ -43,87 +48,23 @@ func (m *virtualMachineImageMutator) Resource() types.Resource {
...
-		}
-		if storageClass.Provisioner != longhorntypes.LonghornDriverName {
-			return nil, fmt.Errorf("the provisioner of storageClass must be %s, not %s", longhorntypes.LonghornDriverName, storageClass.Provisioner)
-		}
-		if storageClass.Parameters[util.LonghornOptionBackingImageName] != "" {
-			return nil, errors.New("can not use a backing image storageClass as the base storageClass template")
+			return patchOps, err
 		}
-		return storageClass, nil
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backend", "value": %s}`, string(bytes)))
```
brandboat: Could we use `harvesterv1.VMIBackendBackingImage` in fmt.Sprintf?
```
patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backend", "value": "%s"}`, harvesterv1.VMIBackendBackingImage))
```

pkg/image/backingimage/backing_image_controller.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/image/vm_image_controller.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/image/common/operator.go
WebberHuang1118: > IMO, if anyone wants to add a new image solution, he doesn't need to care about the control flow for the controller, webhook, API handler, and any details about CRD VMImage (VMImage operations are encapsulated in pkg/image/common), he just needs to focus on the interfaces exported in pkg/image/. This seems simpler to me.

As mentioned, I'd like to encapsulate the VM Image's details for any image solution (including backing image and third-party), the image implementation uses the provided libraries to operate VM Images (rather than directly manipulate the object). 

That gives the integration a clear interface/concept and consistency to me. WDYT? @brandboat

brandboat (commented):
I've skimmed through the code, and overall seems good to me, only a few nits left. I'll test basic VMI operations later. Thanks for the PR!

pkg/image/common/operator.go
brandboat: My understanding is that VirtualMachineImage is similar to a DTO, and we can directly access elements like `vmi.name` or `vmi.namespace`. If there’s no additional logic in these getter methods, I’m wondering if there is much benefit in using `vmio.GetName(vmi)`. This leaves me a bit uncertain about which approach to choose: directly invoking the DTO field (e.g., vmi.Name) or using the GetName method? I’ve noticed that some places use getters while others don’t, and I’m curious if there’s a particular reason or preference for one approach over the other.

Yu-Jack: A quick question, are there any other objects that need to implement the `VMIOperator` interface? Since we treat the `VirtualMachineImage` as a simple object to accept different parameters to connect with `BackingImage` and `CDI`, do we still need the `VMIOperator` interface? From my perspective, maybe we don't need to?

WebberHuang1118: > My understanding is that VirtualMachineImage is similar to a DTO, and we can directly access elements like `vmi.name` or `vmi.namespace`. If there’s no additional logic in these getter methods, I’m wondering if there is much benefit in using `vmio.GetName(vmi)`. This leaves me a bit uncertain about which approach to choose: directly invoking the DTO field (e.g., vmi.Name) or using the GetName method? I’ve noticed that some places use getters while others don’t, and I’m curious if there’s a particular reason or preference for one approach over the other.

> A quick question, are there any other objects that need to implement the VMIOperator interface? Since we treat the VirtualMachineImage as a simple object to accept different parameters to connect with BackingImage and CDI, do we still need the VMIOperator interface? From my perspective, maybe we don't need to?

From the original design and my opinion, for any image solution integration in `pkg/image` (e.q. `pkg/image/backingimage` and `pkg/image/cdi`), it should just focus on implementing interfaces exported at `pkg/image/backend`. 

So in such a point of view, the particular image solution implementation/integration should not directly manipulate the resource VM Image. More generally speaking, it shouldn't have to care about any details in the resource VM image, instead, it only has to know the methods/interfaces (at `pkg/image/common`) that are provided to operate VM Images.

brandboat (approved):
LGTM, thanks!

Test below cases and all passed.
- Create image through URL download
- Create image through upload
- Failed to create image by using invalid URL and check image retry 3 times, error message pop up correctly.
- Download image from existing image

pkg/webhook/resources/virtualmachineimage/mutator.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/longhorn/backupstore"
-
```
Vicente-Cheng: Should not related to this PR?

pkg/image/backingimage/backing_image_controller.go
```diff
@@ -0,0 +1,72 @@
...
+	// If the VM image is imported, and we think we know everything about it, i.e. we've
+	// now been through a series of progress updates during image download,
+	// and those are finally done, so let's not worry about further updates.
+	// TODO: Improve image to keep sync with LH backing image #6936
+	if h.vmio.IsImported(vmi) {
+		return nil, nil
+	}
+
+	err = nil
+	for _, status := range bi.Status.DiskFileStatusMap {
```
Vicente-Cheng, line 54: I am curious that each `DiskFileStatusMap` means one backingimage replica.
For example, if we have two replicas for one backingimage, these two disk files have different statuses.
Which one should we use for the VM Image?

pkg/image/backingimage/upload.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}()
+
+	// Wait for backing image data source to be ready. Otherwise the upload request will fail.
+	dsName, err := util.GetBackingImageDataSourceName(biu.biCache, vmi)
+	if err != nil {
+		return fmt.Errorf("failed to get backing image name for VMImage %s/%s, error: %w", vmi.Namespace, vmi.Name, err)
+	}
+
+	if err := biu.waitForBackingImageDataSourceReady(dsName); err != nil {
```
Vicente-Cheng, line 78: Could we consider to use `wait.PollUntilContextTimeout`?
That might have good readability, and we could focus on the checking logic.

pkg/image/common/validator.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	if vmi.Spec.PVCNamespace == "" {
+		return werror.NewInvalidError(`pvcNamespace is required when image source type is "export-from-volume"`, "spec.pvcNamespace")
+	}
+	if vmi.Spec.PVCName == "" {
+		return werror.NewInvalidError(`pvcName is required when image source type is "export-from-volume"`, "spec.pvcName")
+	}
+
+	ssar, err := v.ssar.Create(request.Context, &authorizationv1.SelfSubjectAccessReview{
```
Vicente-Cheng, line 201: Assume we use the validator with a harvester webhook. In what situation will we not have permission?

Vicente-Cheng (commented):
some questions, generally lgtm

pkg/image/backingimage/upload.go
WebberHuang1118: Actually, in the bi-creation phase, there is only one replica of the backing image. Even if there is an iteration for the map, there should be only one element in the map,  this refactor PR follows the original implementation.

We will have a separate enhancement for tracking the bi-replicas' status post the creation phase.

pkg/image/common/validator.go
WebberHuang1118: This refactor follows the original implementation, I think we can have a separate ticket for tracking this.

pkg/image/common/operator.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/backup/backup_metadata.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/image/backingimage/backing_image_controller.go
WebberHuang1118: Actually, in the bi-creation phase, there is only one replica of the backing image. Even there is a iteration for the map, there should be only one element in the map,  this refactor PR follows the original implementation.

We will have a separate enhancement for tracking the bi replica status post the creation phase.

Vicente-Cheng:
For the folder (files) structure enhancement. I thought we could discuss that after v1.5; I agree currently, we have serval files in different folders for the same thing. However, I did not prefer to do the refactor and file structure changes simultaneously, as that would have more potential risk for the whole stability.

In my personal opinion, we could make the interface/logic in the `pkg/<component>`. The `API`/`controller`/`webhook` should only focus on the logic and mechanism.

 WDYT? cc @WebberHuang1118, @Yu-Jack, @brandboat, @FrankYang0529, @bk201

pkg/image/common/validator.go
Vicente-Cheng: Thanks, we can do that at the next milestone.

pkg/image/backingimage/upload.go
Vicente-Cheng: I thought you reply the comment (https://github.com/harvester/harvester/pull/7170#discussion_r1943191578).
That makes sense. Thanks.

What about the original comment above?
>Could we consider to use wait.PollUntilContextTimeout?
That might have good readability, and we could focus on the checking logic.

pkg/image/backingimage/backing_image_controller.go
Vicente-Cheng: Thanks for it!

pkg/image/backingimage/upload.go
WebberHuang1118: This refactor follows the original implementation, maybe we could have a separate enhancement ticket for all your suggestions. WDYT?

WebberHuang1118:
> For the folder (files) structure enhancement. I thought we could discuss that after v1.5; I agree currently, we have serval files in different folders for the same thing. However, I did not prefer to do the refactor and file structure changes simultaneously, as that would have more potential risk for the whole stability.
> 
> In my personal opinion, we could make the interface/logic in the `pkg/<component>`. The `API`/`controller`/`webhook` should only focus on the logic and mechanism.
> 
> WDYT? cc @WebberHuang1118, @Yu-Jack, @brandboat, @FrankYang0529, @bk201

Yes, considering the schedule and source code structure change will not bring a backward compatibility concern. I agree with your idea. We can discuss the source code structure after v1.5. Thanks.

pkg/image/common/operator.go
Vicente-Cheng: I thought some of the vmImgOperator might be overworked, as you mentioned above, `GetName` and `GetNamespace`, ... etc.
Usually, we use the general interface to protect some resources or make operations sequences simultaneously.
So for the resource we did not need to protect, we could directly access it.

But I thought the vmImgOperator gave the friendly/unified interface to update the VMImage status. That's very helpful if we have different backends to implement.

So, for me, keeping the vmImgOperator would prevent the same work for the vmImage status update.

pkg/image/backingimage/upload.go
Vicente-Cheng: That's Ok to me. We could do more in the next version.

pkg/image/common/operator.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func (vmio *vmiOperator) Initialized(old *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
+	return vmio.stateTransit(old, VMImageStateInitialized, "", 0, 0, 0)
+}
+
+func (vmio *vmiOperator) Importing(old *harvesterv1.VirtualMachineImage, msg string, progress int) (*harvesterv1.VirtualMachineImage, error) {
+	return vmio.stateTransit(old, VMImageStateImporting, msg, progress, 0, 0)
+}
+
```
Vicente-Cheng, line 329: Could we add some comments here? (for Imported func)
We should not add any msg when the image is imported. Or UI will mark this image as failed.

WebberHuang1118: Could you explain more?

for [Importing()](https://github.com/harvester/harvester/blob/77801103999caa4d5cbac867843aa9174cd20946/pkg/image/common/operator.go#L326-L328) and [Imported()](https://github.com/harvester/harvester/blob/77801103999caa4d5cbac867843aa9174cd20946/pkg/image/common/operator.go#L330-L332) functions, the messages are decided by the related image implementation. 

For example, in the case of LH backing image, we get the msg [from the backing image status](https://github.com/harvester/harvester/blob/77801103999caa4d5cbac867843aa9174cd20946/pkg/image/backingimage/backing_image_controller.go#L54-L69)

Vicente-Cheng: From the backingimage path. We updated the status of both `importing` or `imported` in the VMImage CR.
And the message comes from the backingImage status message.

But in CDI path, we need to handle the message ourselves.
I would like to say if we set the VMimage status to Imported, we cannot put any value to the `message.`
The UI will show the VM image failed once the message value is not the empty string.

pkg/image/backingimage/upload.go
bk201: @WebberHuang1118 let's create an enhancement ticket in case we forget, thanks!

pkg/image/common/operator.go
WebberHuang1118: IIUC, does it more like a cosmetics issue with the GUI?

For the backend part, there should be no limitation on the VM Image msg setting.

pkg/image/backingimage/upload.go
WebberHuang1118: @bk201 the enhancement ticket https://github.com/harvester/harvester/issues/7524

pkg/image/common/operator.go
Vicente-Cheng: Sure, I could discuss with the UI team once UI starts working on that.

Vicente-Cheng (approved):
lgtm, thanks

Vicente-Cheng:
Hi @FrankYang0529, @Yu-Jack,

Due to the v1.5 schedule, I merge this PR first.
If you have some suggestions for these, feel free to raise them at any time.

Issue #7180, opened 2024-12-17
[BUG] Fix the sync of migrate button on UI


w13915984028:
**Describe the bug**

Harvester render the UI `migrate` button via check vmi object annotation.

https://github.com/harvester/harvester/blob/a9cfd042a3b707108ff6337a933dc646bfbde64a/pkg/api/vm/formatter.go#L106

```
	if canMigrate(vmi) {
		resource.AddAction(request, migrate)
		resource.AddAction(request, findMigratableNodes)
	}
```

And, the vmim controller checks the vmim object then update the vmi object annotation.

https://github.com/harvester/harvester/blob/a9cfd042a3b707108ff6337a933dc646bfbde64a/pkg/controller/master/migration/vmim_controller.go#L45

```
...
	if vmim.Status.Phase == kubevirtv1.MigrationPending && !abortRequested {
		return vmim, h.scaleResourceQuota(vmi)
	} else if vmim.Status.Phase != kubevirtv1.MigrationFailed && abortRequested {
		return vmim, h.setVmiMigrationUIDAnnotation(vmi, string(vmim.UID), StateAbortingMigration)
	}
...
```

You may note that, when `vmim.Status.Phase == kubevirtv1.MigrationPending`, the `vmi` object is not update but the `resourcequota` object  (refer another issue: https://github.com/harvester/harvester/issues/7178)

In this phase, the UI still shows the `migrate` button, if user tries to trigger the migrate from UI, there is such error:

![image](https://github.com/user-attachments/assets/2bafc29d-9119-4e88-af43-93d2e8bd1733)

kubevirt denied the migration, as the `vmim` object is already there.

**To Reproduce**
1. Create 4 or more VMs on a same node
2. Maintain the node
3. 4 VM migrations are created
4. 2 of them starts ( by default 2 in parallel)
5. check UI, another 2 VMs sitll show `migrate` button, if you click it, the related error will be reported

**Expected behavior**

Render the UI properly

**Support bundle**

**Environment**
 - Harvester ISO version:  Harvester v1.3.1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

bk201:
@w13915984028 please add the require-ui labels if need UI help or change to `area/ui` if this is purely a UI issue.

w13915984028:
@bk201 At the moment, this is mainly a backend issue.

w13915984028:
[...]

PR #7195, opened 2024-12-18
[WIP] Ensure the vm is synced when vmim has changes

w13915984028:

**Problem:**

Migration UI menu is not properly rendered when VMIM has changes.

**Solution:**

Sync on all cases.

When a controller updates several objects in a row, there is always time-sequences related issues. Should be carefully processed.

BTW: The code is not fully finished & tested.

**Related Issue:**

https://github.com/harvester/harvester/issues/7180

**Test plan:**

(1) Setup a 2+ nodes cluster
(2) Create 4+ VMs, schedule them to one node
(3) Maintain this node, all VMs are put into migration automatically
(4) In general, 2 VMs will run migration at first, another 2 are on pending
(5) Those 2 pending migration VMs, should have `abort` menu, and allow to be aborted; the other running migration VMs should also be same.
...

TBD: Check the behaivour of UI and kubectl, if kubectl deletes a vmim successfully, UI should also allow, either from `Abort` menu, or from the `delete` menu.

TBD: Confirm the `Abort` allowed status from kubevirt source code.
...

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -73,7 +73,6 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	resource.AddAction(request, addVolume)
 	resource.AddAction(request, removeVolume)
 	resource.AddAction(request, cloneVM)
-	resource.AddAction(request, migrate)
 	resource.AddAction(request, findMigratableNodes)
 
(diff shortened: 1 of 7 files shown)


w13915984028:
on v1.6.0, this PR is replaced by https://github.com/harvester/harvester/pull/8711

PR #7203, opened 2024-12-19
fix(upgrade): clean up image with crictl

starbops:

**Problem:**

We used `ctr` to remove container images initially. But that's not thorough.

**Solution:**

Replace the use of `ctr i rm` with `crictl rmi`.

**Related Issue:**

#6691 

**Test plan:**

1. Prepare a Harvester cluster in v1.4.0
2. Prepare a customized build of Harvester ISO image containing this fix (for QAs, please use the master-head ISO image)
3. Get the before-upgrade image list on each node with `crictl images`
4. Upgrade the cluster with the ISO image
5. After the upgrade succeed, collect the after-upgrade image list on each node
6. Besides, the result of `crictl images` should not contain bunch of `<none>` lines (please refer to https://github.com/harvester/harvester/issues/6691#issue-2560988813 for example)
7. Compare the two lists. Images exist in v1.4.0 but not in the customized build should be removed.


bk201 (approved):

FrankYang0529 (approved):
LGTM. Thanks.

PR #7208, opened 2024-12-19
feat: use cronjob to update sprint issue

Yu-Jack:
Refer https://github.com/longhorn/longhorn/blob/master/.github/workflows/update-issue-sprint.yml

### DIFF
--- a/.github/workflows/update-issue-sprint.yml
+++ b/.github/workflows/update-issue-sprint.yml
@@ -1,0 +1,54 @@
+name: Update-Issue-Sprint
+
+on:
+  schedule:
+    # Trigger every Sunday at 20:00
+    - cron: '0 20 * * 0'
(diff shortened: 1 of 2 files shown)


.github/workflows/update-issue-sprint.yml
```diff
@@ -0,0 +1,54 @@
...
+  repository-projects: write
+  issues: write
+
+jobs:
+  move-to-next-iteration:
+    name: Move to next iteration
+    runs-on: ubuntu-latest
+
+    env:
+      GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
```
Yu-Jack, line 20: As I know, this secret has been set up before.

bk201 (approved):

PR #7228, opened 2024-12-23
Bump image version

brandboat:
**Problem:**

Bump image and chart version for harvester v1.3 branch, ~~blocked by https://github.com/harvester/node-disk-manager/pull/165~~

**Solution:**

Bump image and chart version for harvester v1.3 branch

**Related Issue:**

https://github.com/harvester/harvester/issues/6590

**Test plan:**
N/A


Vicente-Cheng:
Node-Disk-Manager is ready.
Please use tag v0.6.5. Thanks!

deploy/charts/harvester/values.yaml
```diff
@@ -462,12 +462,12 @@ harvester-load-balancer:
   image:
     imagePullPolicy: IfNotPresent
     repository: rancher/harvester-load-balancer
-    tag: v0.3.0
+    tag: v0.4.3
```
bk201: Isn't 0.4.x used for v1.4 branch?

brandboat: I had the same question initially, but after discussing offline with @starbops, it seems that this is the desired version...

starbops: According to https://github.com/harvester/harvester/issues/6590#issuecomment-2527861060, v0.4.3 is the desired version to bump to. cc @w13915984028

Vicente-Cheng (approved):
Thanks!

brandboat:
Update: pr rebased to v1.3 latest, update support-bundle-kit to v0.0.48, harvester-load-balancer & harvester-load-balancer-webhook to v0.4.4.

bk201 (approved):

Issue #7086, opened 2024-12-03
[BUG] Harvester VM Metrics Are Not Accurate


deiberts86:
**Describe the bug**
Prometheus VM Metrics are not accurate whatsoever where metrics at a percentage level are exceeding 200000% within the Harvester UI VM metrics. 

**To Reproduce**
1. Install Harvester Monitoring Add-On.
2. Create a few VMs and do some workloads.
3. Go to your "Dashboard > Virtual Machine Metrics"
4. Observed excessive percentages.

**Expected behavior**
Percentages shouldn't exceed 100% utilization

**Support bundle**
I'll send one if needed

**Environment**
OS: Harvester version 1.4.0 (FRESH) and tested with Harvester upgrade 1.3.2 - 1.4.0 (UPGRADE)
Hardware: Dell PowerEdge R640 (8-bay)

**Additional context**
Screenshot attached of the VM Metrics
![crazy-metrics-harvestesr](https://github.com/user-attachments/assets/59c758b3-c709-49ce-a321-626dcdf40c3c)

CamrynCarter:
SURE-9477

deiberts86:
[supportbundle_63682b4f-c84a-4f8c-9b23-8f489add4777_2024-12-05T17-12-05Z.zip](https://github.com/user-attachments/files/18027199/supportbundle_63682b4f-c84a-4f8c-9b23-8f489add4777_2024-12-05T17-12-05Z.zip)

noahgildersleeve:
This was reproduced in our lab as well. 

![image](https://github.com/user-attachments/assets/fb7a8b3a-67ea-4e24-98a4-58fbb2287370)
[supportbundle_37afb7f8-8815-43f9-a3eb-56e59d576053_2024-12-19T19-06-54Z.zip](https://github.com/user-attachments/files/18202237/supportbundle_37afb7f8-8815-43f9-a3eb-56e59d576053_2024-12-19T19-06-54Z.zip)

w13915984028:
Seems something is broken, let me take a look, thanks.

w13915984028:
I got this crazy result as well

![image](https://github.com/user-attachments/assets/27c3985f-3664-4bd5-b0d3-b5c684435370)

The `yaxes` needs to be adjusted to a trully percentage data.

[...]

PR #7250, opened 2024-12-25
Fix incorrect unit in cpu usage dashboard

brandboat:
**Problem:**
Prometheus VM Metrics are not accurate whatsoever where cpu usage metrics at a percentage level are exceeding 200000% within the Harvester UI VM metrics.

**Solution:**
fix incorrect cpu usage unit, this PR aim to fix issue in the upgrade path. harvester-installer pr is here: https://github.com/harvester/harvester-installer/pull/908

**Related Issue:**
https://github.com/harvester/harvester/issues/7086

**Test plan:**
- Prepare a single node v1.4.0 harvester  (easy to test)
- Upgrade harvestester to the version with this patch
- Enable rancher-monitoring addon
- Prepare some VMs (e.g. 3 VM)
- Goto Grafana UI and wait for a couple of minutes
- Check the value of cpu usage should fall in range [0, 100] in both `Harvester VM Dashboard` and `Harvester VM Info Detail` dashboards (during vm startup, the cpu usage may slightly go over 100%, after a while it should dramatically drop to a low value, like 2%)

<image src=https://github.com/user-attachments/assets/1b6e6d2c-bb8a-4364-aab5-a00a6f00910d width=500 />
<image src=https://github.com/user-attachments/assets/4df795d5-c996-4f97-9923-f68e8ae6d41a width=500 />

### DIFF
--- a/package/upgrade/migrations/upgrade_manifests/v1.4.0/pre-hook.sh
+++ b/package/upgrade/migrations/upgrade_manifests/v1.4.0/pre-hook.sh
@@ -1,0 +1,60 @@
+#!/bin/bash -x
+
+patch_harvester_vm_dashboard()
+{
+    echo "Patch configmap/harvester-vm-dashboard..."
+    local patch_json=$( \
+        kubectl get -n cattle-dashboards configmap harvester-vm-dashboard -o json | \
+            jq '.data."harvester_vm_dashboard.json" | fromjson | .panels |=
+                map(
+                    if .title == "CPU Usage" then
+                        .yaxes |= map(
+                            if .format == "percentunit" then
+                                .format = "percent"
+                            else
+                                .
+                            end
+                    )
+                    else
+                        .
+                    end
+                )' |
+        jq -r .
+    )
+
+    if [ -z "$patch_json" ]; then
+        echo "Something goes wrong, the patch json is empty... skip this patch"
+        return 1
+    fi
+
+    patch_json=$(echo '{"data":{"harvester_vm_dashboard.json": ""}}' | jq --arg arg "$patch_json" '.data."harvester_vm_dashboard.json"=$arg')
+    kubectl patch -n cattle-dashboards configmap harvester-vm-dashboard --type merge -p "$patch_json"
+}
+
+patch_harvester_vm_detail_dashboard()
+{
+    echo "Patch configmap/harvester-vm-detail-dashboard..."
+    local patch_json=$( \
+        kubectl get -n cattle-dashboards configmap harvester-vm-detail-dashboard -o json | \
+            jq '.data."harvester_vm_info_details.json" | fromjson | .panels |=
+                map(
+                    if .title == "CPU Usage" then
+                        .fieldConfig.defaults.unit = "percent"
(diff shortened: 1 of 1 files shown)


package/upgrade/migrations/upgrade_manifests/v1.4.0/pre-hook.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash -ex
```
bk201: I suggest not bailing out if these two patches are errors. The upgrade should continue, and it's OK to apply the patch manually

package/upgrade/migrations/upgrade_manifests/v1.4.0/pre-hook.sh
```diff
@@ -0,0 +1,60 @@
...
+        echo "Something goes wrong, the patch json is empty... skip this patch"
+        return 1
+    fi
+
+    patch_json=$(echo '{"data":{"harvester_vm_info_details.json": ""}}' | jq --arg arg "$patch_json" '.data."harvester_vm_info_details.json"=$arg')
+    kubectl patch -n cattle-dashboards configmap harvester-vm-detail-dashboard --type merge -p "$patch_json"
+}
+
+# Fix incorrect unit in CPU Usage dashboard https://github.com/harvester/harvester/issues/7086
+patch_harvester_vm_dashboard
```
bk201, line 59: Are there any side effects if this script is reentered again?

package/upgrade/migrations/upgrade_manifests/v1.4.0/pre-hook.sh
brandboat: Thanks, updated.

package/upgrade/migrations/upgrade_manifests/v1.4.0/pre-hook.sh
brandboat: I believe no, the script only replace unit in cpu usage dashboard from percentunit to percent. If there is no matches (e.g. no cpu usage dashboard, or the unit is already percent), it should remain the json as it is.

Below is my test that execute pre-hook.sh multiple times when two cpu usage dashboards unit are already `percent` (the expected unit). And everything is ok.
```
harvester-node-0:~ # sh pre-hook.sh
Patch configmap/harvester-vm-dashboard...
configmap/harvester-vm-dashboard patched (no change)
Patch configmap/harvester-vm-detail-dashboard...
configmap/harvester-vm-detail-dashboard patched (no change)
harvester-node-0:~ # sh pre-hook.sh
Patch configmap/harvester-vm-dashboard...
configmap/harvester-vm-dashboard patched (no change)
Patch configmap/harvester-vm-detail-dashboard...
configmap/harvester-vm-detail-dashboard patched (no change)
```

bk201 (approved):

FrankYang0529 (approved):
Thanks for the patch.

Before:
Screenshot 2024-12-30 at 11 44 56 AM

After:
Screenshot 2024-12-30 at 1 02 36 PM

PR #7273, opened 2024-12-27
Hep for vm image multiple backend support [ci skip]

WebberHuang1118:
HEP for VM Image multiple backend support

**Related Issue:**
https://github.com/harvester/harvester/issues/6936


innobead:
This is also related to Longhorn. @derekbit needs someone from Longhorn help with review as well.

derekbit:
@c3y1huang Can you help review?
@innobead I'm unable to add Shuo in this PR.

innobead:
Let me add Shuo to this org.

ihcsim:
Still relevant.

WebberHuang1118:
Still under review

bk201 (approved):

Vicente-Cheng (commented):
That is generally good for me.

Just one thing wants to discuss.
The timing of updating size/virtual size differs between CDI and BackingImage.

CDI updates the size/virtual size before any operation (like download/upload/exportTo image)
BackingImage updates the size/virtual size after the operation (like download/upload/exportTo image), and it comes from the longhorn side.

Should we align these two? Or do we need to change the flow on this HEP?

WebberHuang1118:
> That is generally good for me.
> 
> Just one thing wants to discuss. The timing of updating size/virtual size differs between CDI and BackingImage.
> 
> CDI updates the size/virtual size before any operation (like download/upload/exportTo image) BackingImage updates the size/virtual size after the operation (like download/upload/exportTo image), and it comes from the longhorn side.
> 
> Should we align these two? Or do we need to change the flow on this HEP?

@Vicente-Cheng 

In the case you mentioned, the VMI controller still calls `UpdateVirtualSize()` for `CDI`. However, as you pointed out, CDI implements it as a [no-op](https://github.com/harvester/harvester/blob/2a2e41eb53c277a3699cbdf8a8695368e35d8fa3/pkg/image/cdi/cdi.go#L98-L101), since it handles virtual size updates differently.

Although the backend solutions may differ in design (e.g., Longhorn BackingImage, CDI, etc.), the framework itself always follows the same flow. So I think it's reasonable to keep the current flow chart in the HEP. What do you think?

Vicente-Cheng:
> > That is generally good for me.
> > Just one thing wants to discuss. The timing of updating size/virtual size differs between CDI and BackingImage.
> > CDI updates the size/virtual size before any operation (like download/upload/exportTo image) BackingImage updates the size/virtual size after the operation (like download/upload/exportTo image), and it comes from the longhorn side.
> > Should we align these two? Or do we need to change the flow on this HEP?
> 
> @Vicente-Cheng
> 
> In the case you mentioned, the VMI controller still calls `UpdateVirtualSize()` for `CDI`. However, as you pointed out, CDI implements it as a [no-op](https://github.com/harvester/harvester/blob/2a2e41eb53c277a3699cbdf8a8695368e35d8fa3/pkg/image/cdi/cdi.go#L98-L101), since it handles virtual size updates differently.
> 
> Although the backend solutions may differ in design (e.g., Longhorn BackingImage, CDI, etc.), the framework itself always follows the same flow. So I think it's reasonable to keep the current flow chart in the HEP. What do you think?

Sure. We still use the same flow and add some extra behavior.

Vicente-Cheng (approved):
lgtm

PR #7286, opened 2025-01-02
misc: update copyright to 2025

Yu-Jack:
Update copyright


Yu-Jack:
Please ignore the CodeFactor, the error is from generated file.

bk201 (approved):

bk201:
Please also help fix https://github.com/harvester/harvester/blob/f374c203fc23f5123488c579664a22528298bb72/README.md?plain=1#L135

brandboat (approved):
LGTM, thanks.

One question, do we need to update the copyright in other repos ? e.g. node-manager, node-disk-manager, tests .etc. ?

Yu-Jack:
@brandboat Yes, we need to. I did that because I ran into CI issue in harvester/harvester.

Issue #7178, opened 2024-12-17
[BUG] Fix the race between Rancher and Harvester on resourcequota when VM is migrating


w13915984028:
**Describe the bug**

When a resourcequota on namespace is created from Rancher Manager upon Harvester, Rancher will sync the configuration. E.g., restart the `rancher-cluster-agent*` POD on Harvester.

On the other hand, when VM is migrating, Harvester will actively scale up the resourcequota to make rooms for migration target POD to run, and scale it down when migration is finished.

This works in most cases, but the race may cause some issue if following conditions are met:

-The ResourceQuota for a given namespace has been used 80+%.
-A couple of VMs from this namespace are running on a same node.
-The cattle-cluster-agent related PODs are also running on this node.

Detailed ananlysis: https://github.com/w13915984028/harvester-develop-summary/blob/main/v121-vm-migration-with-resources-quota.md#issue-resourcequota-was-changed-unexpected

**To Reproduce**
1. Import Harvester to Rancher
2. Create a project and namespace with resourcequota limitation
3. Create a couple of VMs on this namespace and use the resource to around 80%
4. Schedule/Migration all those VMs to one node (make sure the VMs can migrate)
5. Schedule the `cattle-cluster-agent*` to this node as well
6. Maintain this node
7. Harvester & kubevirt will auto migration all VMs to other node
8.There is chance to hit the issue and `virt-controller` report such error:

```
./harvester-system.virt-controller-6bc767746-qr5rm.virt-controller.20241210.log.gz:{"component":"virt-controller","level":"info","msg":"reenqueuing Migration ns1/kubevirt-evacuation-dfp6l","pos":"migration.go:233","reason":"failed to create vmi migration target pod: pods \"virt-launcher-ns1-4ae21d5d-n2mtw-qgtdp\" is forbidden:

 exceeded quota: default-4tmmk,

requested: limits.cpu=4,015m,  limits.memory= 8,940,039,936,
used:      limits.cpu=28,505m, limits.memory=62,900,279,552,
limited:   limits.cpu=26,      limits.memory=54,242,646Ki",

"timestamp":"2024-12-10T14:36:47.533782Z"}
```

**Expected behavior**

The migration should be successful.

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

https://github.com/harvester/harvester/issues/7161

w13915984028:
```
[...]

Issue #6594, opened 2024-09-20
[ENHANCEMENT] Preventing VM Backup if the VM is under Migration


WebberHuang1118:
**Is your enhancement related to a problem? Please describe.**

If the VM Backup is triggered during the VM is under Migration, there are multiple situations may occur:
- Filesystem freeze fail
  - It could happen if the VM Backup starts as two VM launcher pods are running
  - From compute log:
    ```
    compute {"component":"virt-launcher","kind":"","level":"error","msg":"Failed to freeze vmi","name":"vm1","namespace":"default","pos":"server.go:269","reason":"Failed to freeze VMI, VMI is currently during migration","timestamp":"2024-09-20T02:39:05.371600Z","uid":"16b24dd6-babb-4168-aa97-498e81cf3b12"}
    compute {"component":"virt-launcher","kind":"","level":"error","msg":"Failed to freeze vmi","name":"vm1","namespace":"default","pos":"server.go:269","reason":"Failed to freeze VMI, VMI is currently during migration","timestamp":"2024-09-20T02:39:05.444526Z","uid":"16b24dd6-babb-4168-aa97-498e81cf3b12"}
    compute {"component":"virt-launcher","kind":"","level":"error","msg":"Failed to freeze vmi","name":"vm1","namespace":"default","pos":"server.go:269","reason":"Failed to freeze VMI, VMI is currently during migration","timestamp":"2024-09-20T02:39:05.531953Z","uid":"16b24dd6-babb-4168-aa97-498e81cf3b12"}
    compute {"component":"virt-launcher","kind":"","level":"error","msg":"Failed to freeze vmi","name":"vm1","namespace":"default","pos":"server.go:269","reason":"Failed to freeze VMI, VMI is currently during migration","timestamp":"2024-09-20T02:39:05.668165Z","uid":"16b24dd6-babb-4168-aa97-498e81cf3b12"}
    compute {"component":"virt-launcher","kind":"","level":"error","msg":"Failed to freeze vmi","name":"vm1","namespace":"default","pos":"server.go:269","reason":"Failed to freeze VMI, VMI is currently during migration","timestamp":"2024-09-20T02:39:05.758315Z","uid":"16b24dd6-babb-4168-aa97-498e81cf3b12"}
    ```
[...]

PR #7315, opened 2025-01-07
feat: add CPU / Memory Hotplug HEP [SKIP CI]

FrankYang0529:

**Related Issue:**

### DIFF
--- a/enhancements/20250107-cpu-memory-hotplug.md
+++ b/enhancements/20250107-cpu-memory-hotplug.md
@@ -1,0 +1,98 @@
+# CPU / Memory Hotplug
+
+## Summary
+
+This enhancement focuses on leveraging KubeVirt's CPU and Memory hotplug feature to enable users to dynamically adjust the CPU and Memory resources of a VM.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/5835
+
+## Motivation
+
+### Goals
+
+- Users can adjust the CPU and Memory resources of a VM dynamically.
+- Users can set maximum CPU and Memory resources for a VM. The dynamic adjustment of CPU and Memory resources will be limited by the maximum resources.
+- Old VMs will not be affected by this change.
+
+### Non-goals
+
+N/A
+
+## Proposal
+
+The cpu / memory hotplug feature will be enabled by adding annotation `harvesterhci.io/enableCPUAndMemoryHotplug`. When creating a VM, users can set 4 fields: initial cpu / memory and maximum cpu / memory. The initial cpu / memory will be set to `domain.cpu.sockets` and `domain.memory.guest`. The maximum cpu / memory will be set to `domain.cpu.maxSockets`, `resources.limits.cpu` and `domain.memory.maxGuest`, `resources.limits.memory`.
+
+When the VM is running, users can adjust the CPU and Memory resources of the VM. The adjustment will be limited by the maximum resources.
+
+### User Stories
+
+#### Story 1 - Vertical Scaling a VM without downtime
+
+There is a legacy workload which is not actively developed. The workload was not designed as a distributed system, so it's hard to scale out. Recently, there are some users using this application, and the workload is getting heavier. The current resources allocated to the VM are not enough to handle the workload. The administration wants to adjust the CPU and Memory resources of the VM without downtime.
+
+### User Experience In Detail
+
+1. Prepare an Image.
+
+2. On the VM creation dashboard, click the `Enable CPU / Memory Hotplug` button and the dashboard will show another two fields: Maximum CPU and Maximum Memory.
+
+3. Fulfill the VM creation form and click the `Create` button.
+
+4. After the VM is created, administrators can adjust the CPU and Memory resources by clicking `CPU / Memory Hotplug` on the VM details page.
+
+### API changes
+
+#### Update CPU and Memomory resources of a VM [POST]
+
+Endpoint:  `/v1/harvester/vm/<vm>?action=cpuAndMemoryHotplug`
+
+Request (application/json):
+```json
+{
+  "sockets": 2,
+  "memory": "4Gi"
+}
+```
+
+## Design
+
+### Implementation Overview
+
+#### Initial CPU and Memory
+
+Currently, when creating a VM by UI, the initial CPU and Memory resources are set to `resources.limits`. If there is no annotation `harvesterhci.io/enableCPUAndMemoryHotplug`, the system should follow old pattern to set values. If there is annotation, the system sets the initial CPU to `domain.cpu.sockets` and the initial Memory to `domain.memory.guest`.
+
+There are 3 fields in `template.spec.domain.cpu`: `cores`, `sockets`, and `threads`. Set CPU to `sockets` instead of `cores`, is because KubeVirt only detects the `sockets` field when doing CPU hotplug ([ref](https://github.com/kubevirt/kubevirt/blob/4366d2810a6b59d1b9d4c3fdcc052cf78855a385/pkg/virt-controller/watch/vm/vm.go#L690-L706)).
+
+#### Maximum CPU and Memory
+
+The maximum CPU and Memory resources are set to `domain.cpu.maxSockets`, `resources.limits.cpu` and `domain.memory.maxGuest`, `resources.limits.memory` respectively. These fields are immutable. Adjust the maximum CPU or Memory needs to restart the VM.
+
+#### CPU and Memory Hotplug Action
+
+The API server sends a patch to adjust the CPU and Memory resources of a VM.
+
+Following cases makes a VM cannot do CPU and Memory hotplug. The VM will be marked with `RestartRequired` condition, which means the VM needs to be restarted to apply the changes.
+
+* new `sockets` is smaller than current `sockets`
+* new `memory` is smaller is bigger than current `memory`
+* modify `maxSockets` or `maxGuest`
+
+### Test plan
+
+* Use `cpuAndMemoryHotplug` action to adjust the CPU and Memory resources of a VM.
+  * If `sockets` is bigger than current `sockets` and smaller than `maxSockets`, the VM should be able to adjust the CPU resources dynamically.
+  * If `memory` is bigger than current `memory` and smaller than `maxGuest`, the VM should be able to adjust the Memory resources dynamically.
+  * If `sockets` is smaller than current `sockets`, the VM should be marked with `RestartRequired` condition.
+  * If `memory` is smaller than current `memory`, the VM should be marked with `RestartRequired` condition.
+* Modify `maxSockets` or `maxGuest` marks the VM with `RestartRequired` condition.
+
+### Upgrade strategy
+
+The CPU and Memory hotplug feature will be enabled for new VMs only, because the initial CPU and Memory resources were set to `resources.limits` in the old system.
+
+## Note
+
+N/A


enhancements/20250107-cpu-memory-hotplug.md
```diff
@@ -0,0 +1,98 @@
...
+
+### User Experience In Detail
+
+1. Prepare an Image.
+
+2. On the VM creation dashboard, click the `Enable CPU / Memory Hotplug` button and the dashboard will show another two fields: Maximum CPU and Maximum Memory.
+
+3. Fulfill the VM creation form and click the `Create` button.
+
+4. After the VM is created, administrators can adjust the CPU and Memory resources by clicking `CPU / Memory Hotplug` on the VM details page.
```
ihcsim, line 43: It will be really cool if the UI can present a slider with the initial and maximum values being the lower and upper bound :sunglasses:.

enhancements/20250107-cpu-memory-hotplug.md
```diff
@@ -0,0 +1,98 @@
...
+
+Request (application/json):
+```json
+{
+  "sockets": 2,
+  "memory": "4Gi"
+}
+```
+
+## Design
```
ihcsim, line 59: AIUI, hotplugging resources triggers live migration. Perhaps some visual cues in the UI to let the user know a live migration has started will be useful. In particular, these 2 issues come to mind:

* https://github.com/harvester/harvester/issues/7180
* https://github.com/harvester/harvester/pull/7146

enhancements/20250107-cpu-memory-hotplug.md
```diff
@@ -0,0 +1,98 @@
...
+
+The KubeVirt webhook doesn't handle these cases. It allows these requests but setting `RestartRequired` condition on the VM. This will make the VM can't do next CPU and Memory Hotplug, so we should handle these cases in the Harvester webhook server.
+
+### Test plan
+
+Integration test plan.
+
+### Upgrade strategy
+
+In old system, the initial CPU and Memory resources are set to `resources.limits`. We should migrate these fields to `template.spec.domain.cpu.sockets` and `template.spec.domain.memory.guest` when the VM is stopped or restarted.
```
w13915984028: Many customers limit the VM's CPU & Memory for resource management, if we migrate those existing limitations to `memory.guest`, it may break things.

The reasonable solution will be document this: the hot-plug only appies to new VM.

w13915984028 (commented):
Thanks for the investigation and HEP, it describes how to implement hot-plug on Harvester.

Meanwhile, seems some scenarios are not covered:

(1) The `hot-plug`  vs `migrate&abortMigration`

If `hot-plug` is backed by `live-migration`, then the `hot-plug` will affect / be affected by the coming/on-going migration, and vice versa.  (not sure `hot-plug` allows migration in same host)

We have many checks (usb/vgpu deivce/cpu-manager...) to deny `migrate`, are they applied to `hot-plug` as well?

UI menu rending. `hot-plug`, `migrate`, `abortMigration`, can user view and abort the `hot-plug` triggered migration? When a legacy migration is ongoing, can user trigger `hot-plug`?

Backup blocks `migrate`.


(2) VM resource management, ResourceQuota, OOM

In the past, VM's limits is converted to POD's carrying CGroup's limits, with certain amount of overhead.

With `hot-plug`, what's the rule? are there any potential side-influence on OOM?

In-between live-migration, Harvester scales up/down the ResourceQuota, will `hot-plug` also apply? But it seems a bit different.

e.g. one VM requests `2core, 2Gi memor`y, in live-migration, additional ~2core, 2Gi are temporary added to the RQ.

For `hot-plug`, it may become: max `2core, 2Gi`, initial `1core, 1Gi`, when `hot-plug` another `1core, 1Gi`, how?

Are there discussions / considrations on those? thanks.


context:

https://github.com/harvester/harvester/issues/7538
https://github.com/harvester/harvester/issues/7178
https://github.com/harvester/harvester/issues/6594
https://github.com/harvester/harvester/issues/6591
https://github.com/harvester/harvester/issues/6321
https://github.com/harvester/harvester/issues/5069
https://github.com/harvester/harvester/issues/4888
...

enhancements/20250107-cpu-memory-hotplug.md
FrankYang0529: Updated it. Thank you.

enhancements/20250107-cpu-memory-hotplug.md
```diff
@@ -0,0 +1,92 @@
+# CPU / Memory Hotplug
+
+## Summary
+
+This enhancement focues on leveraging KubeVirt's CPU and Memory hotplug feature to enable users to dynamically adjust the CPU and Memory resources of a VM.
```
ihcsim:
```suggestion
This enhancement focuses on leveraging KubeVirt's CPU and Memory hotplug feature to enable users to dynamically adjust the CPU and Memory resources of a VM.
```

enhancements/20250107-cpu-memory-hotplug.md
```diff
@@ -0,0 +1,92 @@
...
+
+#### CPU and Memory Hotplug Action
+
+The webhook server should deny following requests:
+* new `sockets` is bigger than `maxSockets`
+* new `sockets` is smaller than current `sockets`
+* new `memory` is bigger than `maxGuest`
+* new `memory` is smaller is bigger than current `memory`
+
+The KubeVirt webhook doesn't handle these cases. It allows these requests but setting `RestartRequired` condition on the VM. This will make the VM can't do next CPU and Memory Hotplug, so we should handle these cases in the Harvester webhook server.
```
ihcsim:
```suggestion
The KubeVirt webhook doesn't handle these cases. It allows these requests by setting `RestartRequired` condition on the VM. This will make the VM can't do next CPU and Memory Hotplug, so we should handle these cases in the Harvester webhook server.
```

enhancements/20250107-cpu-memory-hotplug.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Test plan
+
+Integration test plan.
+
+### Upgrade strategy
+
+The CPU and Memory hotplug feature will be enabled for new VMs only, because the initial CPU and Memory resources were set to `resources.limits` in the old system.
```
ihcsim, line 94: We should add this as a documentation item.

enhancements/20250107-cpu-memory-hotplug.md
ihcsim: If the VM needs to be restarted (whether due to hotplug or not), I think we should just notify the user on the UI. Let me know what you think of https://github.com/harvester/harvester/issues/8411.

> It allows these requests but setting `RestartRequired` condition on the VM.

Do we know what happens after the VM is restarted? Are the new values accepted or rejected?

FrankYang0529: Thanks for the suggestion. I mention #8411 in https://github.com/harvester/harvester/issues/8416.

enhancements/20250107-cpu-memory-hotplug.md
FrankYang0529: I will create a document PR after UI supports cpu / memory hotplug, so we can have some screenshot there.

enhancements/20250107-cpu-memory-hotplug.md
FrankYang0529: > Do we know what happens after the VM is restarted? Are the new values accepted or rejected?

They will be accepted after VM restarts.

enhancements/20250107-cpu-memory-hotplug.md
```diff
@@ -0,0 +1,97 @@
...
+  "sockets": 2,
+  "memory": "4Gi"
+}
+```
+
+## Design
+
+### Implementation Overview
+
+#### Initial CPU and Memory
```
w13915984028, line 63: Is `initial cpu and memroy` similar to existing `requests`?

As `cpu and memory overcommitting` is widely used, can use still get such flexibility via setting a small `init cpu and memory` & big `max cpu and memory`?

enhancements/20250107-cpu-memory-hotplug.md
```diff
@@ -0,0 +1,97 @@
...
+
+Currently, when creating a VM by UI, the initial CPU and Memory resources are set to `resources.limits`. However, these fields are immutable, so we need to set the initial CPU to `template.spec.domain.cpu.sockets` and the initial Memory to `template.spec.domain.memory.guest`.
+
+There are 3 fields in `template.spec.domain.cpu`: `cores`, `sockets`, and `threads`. Set CPU to `sockets`, but not `cores`, is because KubeVirt only detects the `sockets` field when doing CPU hotplug ([ref](https://github.com/kubevirt/kubevirt/blob/4366d2810a6b59d1b9d4c3fdcc052cf78855a385/pkg/virt-controller/watch/vm/vm.go#L690-L706)).
+
+#### Maximum CPU and Memory
+
+The maximum CPU and Memory resources are set to `template.spec.domain.cpu.maxSockets` and `template.spec.domain.memory.maxGuest` respectively. Same as above, we will not set these values to `requests.limits` because they are immutable.
```
w13915984028: How is https://docs.harvesterhci.io/v1.5/advanced/index#additional-guest-memory-overhead-ratio functioning on hot-plug mode?

w13915984028 (commented):
Thanks for the HEP.

Could we add some addtional stories about the resource control and scaling flexibility in production deployment:

(1) Is it possbile & how to covert from current overcommitting mode to hot-plug mode fully?  e.g. IaaS provider.

(2) In overcommitting mode, OOM only happens when upper limit is reached; but on hot-plug mode, if user sets a small initial value, will OOM happen easily? if a big initial value is set, it locks more resources from beginning.
```
e.g. a VM with 8core, 16g mem,
legacy:  requests 2 core,  2g,  limits  8 core, 16g
hot-plug: initial 4 core, 12g,    max  16 core, 32g  // more resources are locked ?
```

We have seen, many Harvester deployments are sensitive on resources usage and control.

(3) When `hot-plug` is an option, the choice is on user's hand, are there some best practices/suggestions for `hot-plug` type VMs?

(4) Is https://docs.harvesterhci.io/v1.5/advanced/index#additional-guest-memory-overhead-ratio still working as usual? how does hog-plug set the pod/cgroup limits.  https://github.com/harvester/harvester/issues/8416#issuecomment-2939017216

FrankYang0529:
> (1) Is it possbile & how to covert from current overcommitting mode to hot-plug mode fully? e.g. IaaS provider.

To do this, we have to update `generateMemoryPatch` function in VM mutator, because it patches `guest` value based on `resources.limits.memory`. We can set maximum value to `resources.limits.memory` and do nothing if `guest` value is smaller than (resources.limits.memory - reserved memory). However, this means the pod may occupy more resources (double or triple) than VM really uses.

https://github.com/harvester/harvester/blob/b0a59f5b6531bec2d70702f2b813f1056139814a/pkg/webhook/resources/virtualmachine/mutator.go#L245-L318 

> (2) In overcommitting mode, OOM only happens when upper limit is reached; but on hot-plug mode, if user sets a small initial value, will OOM happen easily? if a big initial value is set, it locks more resources from beginning.

The OOM happens when the VM usage is more than pod limitation. If we update the VM mutator like I mentioned in question 1. The OOM issue is like before. There is no new issue.

> (3) When `hot-plug` is an option, the choice is on user's hand, are there some best practices/suggestions for `hot-plug` type VMs?

It's tricky here. Users want to have cpu / memory hot-plug, because they don't want a VM occupy too much resources at first. The KubeVirt doesn't allow user to live update `resources.limits`, that means we should set `resources.limits` as maximum sockets / memory or don't set any `resources.limits`. For the first case, that mean a VM looks like using less resources. Actually, the VM pod occupies maximum resources. This doesn't match user expectation. For the second case, it means there is no resources limitation on a pod and the pod may use whole node resources.

> (4) Is https://docs.harvesterhci.io/v1.5/advanced/index#additional-guest-memory-overhead-ratio still working as usual? how does hog-plug set the pod/cgroup limits. [#8416 (comment)](https://github.com/harvester/harvester/issues/8416#issuecomment-2939017216)

No, it doesn't work if there is no `resources.limits`.

w13915984028:
It seems even upstream kubevirt is also not covering all cases well:

1. In VM default processing, `VM's resources.Limits.Memory and resources.Requests.Memory` can still be set; if not set, it get a value from `spec.Domain.Memory.Guest` and factor with the cluster level `clusterConfig.GetMemoryOvercommit`

https://github.com/kubevirt/kubevirt/blob/7a7199bbfa83b1f7585c607289644956cc186f66/pkg/defaults/defaults.go#L310-L347

2. When hot-plug, the `memoryDelta` is added to `resources.requests` and `resources.limits`;  looks inconsistant with above

https://github.com/kubevirt/kubevirt/blob/7a7199bbfa83b1f7585c607289644956cc186f66/pkg/virt-controller/watch/vm/vm.go#L3282-L3310


Back to Harvester case, we might consider mark a VM as `hot-plug`able explicitly or implicitly.

(1). when `maxSockets != sockets`, means this VM is hot-plugable, then the `requests` and `limits` should meet certain condition: e.g. it strictly equals to guest definition;

(2). when `maxSockets == sockets or maxSockets == 0 (?)`, means this VM is not hot-plugable,  the `requests` can still follow current overcommit https://docs.harvesterhci.io/v1.4/advanced/index/#overcommit-config

BTW, in kubevirt, the `hot-plug` is controlled by feature-gate, should we expose this as well? when `hot-plug` is enabled, then new VMs are default to the `hot-plug` mode; otherwise, in legacy mode. 

Seems to put the `hot-plug` to day 2 option is not so bad.

Why we take `overcommit` as a feature: 

In many customer deployments, the resources are limited;  somewhere single-node clusters are running. Even a `2 CP + 1 witness` cluster might also not able to run hot-plug all the time, as the migration room is quite restricted.

and some VMs are short-lived &/ disposable, like a VM for running build/test tasks; or runs as a guest cluster node,  the guest cluster itself has scaling flexibility and might not adapt to vm(machine) hot-plug feature quickly.



@FrankYang0529 @ihcsim WDYT? thanks.

FrankYang0529:
I reply my concept in https://github.com/harvester/harvester/issues/8416#issuecomment-2969546761. We can choose a thread for discussion.

FrankYang0529:
@w13915984028 I updated HEP according to discussion in the Slack. Please review it again when you're available. Thanks.

w13915984028 (approved):
LGTM, thanks.

Issue #5364, opened 2024-03-14
[BUG] VM's machine type options is not support


lanfon72:
### Describe the bug
![image](https://github.com/harvester/harvester/assets/5169694/fc02f439-bb71-4600-af5e-972af6812d31)

### To Reproduce
1. Install Harvester with any nodes
2. Create an image for VM creation
3. Create VM `vm1` and select one of *machine type* in _advanced options_
4. Update `vm1` to use another *machine type*

### Expected behavior
UI should only display supported machine types.

## Environment:
 - Harvester ISO version: **v1.3-2c8332b7-head** (after `v1.3.0-rc4`)
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): **MegaRAC SP-X 2 nodes** (arm-based)
 - **ui-source** Option: **Auto**

### Additional context
link to #371

albinsun:
Try on `v1.3.1-rc2`, hit same error.

1. <details>
   <summary> Install Harvester with any nodes
   </summary>

   ![image](https://github.com/harvester/harvester/assets/2773781/3df65856-c868-463b-b433-0e56850aed75)
   </details>

1. <details>
   <summary> Create an image for VM creation
   </summary>
   
   ![image](https://github.com/harvester/harvester/assets/2773781/3c8c3bd8-3f97-4948-a2d0-b4065987db6b)
   </details>

1. Create VM vm1 and select one of machine type in advanced options
   ![image](https://github.com/harvester/harvester/assets/2773781/120ce830-8a50-4234-afed-a56893b66570)

## Note
1. Machine Type `None` works fine
![image](https://github.com/harvester/harvester/assets/2773781/914f7f1b-28dd-48a4-ad74-833c8106d08b)

bk201:
Move to p0 for ARM GA. We need UI change to ask supported machine types from backend.

ibrokethecloud:
To test:
* Install harvester to a node
* When creating a VM go to Advanced options
* Check dropdown for machine type
   * On arm64 machines the dropdown should only show `virt`
   * On amd64 machines the dropdown should only show `q35` and `pc-q35`

asc5543:
Related to https://github.com/harvester/harvester/issues/5437, the will encounter the issue https://github.com/harvester/harvester/issues/8628 when trying to create a VM with machine type pc-q35

ibrokethecloud:
[...]

PR #7355, opened 2025-01-13
improve machine type lookup based on processor architecture

ibrokethecloud:

**Problem:**

Currently the `Machine Type` definition in harvester VM `Advanced Options` page is harcoded to `q35`

This breaks VM's on ARM clusters where the only supported type is `virt`

**Solution:**

The PR introduces a new api handler on the local cluster which returns valid machine types based on underlying node arch lookup. The response is an array of valid machine types.

On amd64 hosts the machine types returned are `q35` and `pc`

https://kubevirt.io/user-guide/compute/virtual_hardware/#machine-type

On arm64 hosts the only supported machine type is `virt`
https://kubevirt.io/user-guide/cluster_admin/virtual_machines_on_Arm64/#machine-type

A UI change is needed to ensure the `Machine Types` is populated from the api call to `"${URL}/v1/harvester/cluster/local?link=machineType"`

A similar lookup already exists to find `deviceCapacity` to track vGPU allocation details

**Related Issue:**
https://github.com/harvester/harvester/issues/5364
**Test plan:**

From backend to test
### AMD64 Cluster
* deploy new harvester build to an amd64 node.
* Run the following commands against newly provisioned cluster
```
export APITOKEN="token"
export URL="url"
curl -sk -u $APITOKEN "${URL}/v1/harvester/cluster/local?link=machineTypes"
```
* machine types returned should be `["pc","q53"]`
### ARM64 Cluster
* deploy new harvester build to an arm64 node.
* Run the following commands against newly provisioned cluster
```
export APITOKEN="token"
export URL="url"
curl -sk -u $APITOKEN "${URL}/v1/harvester/cluster/local?link=machineTypes"
```
* machine types returned should be `["virt"]`

### DIFF
--- a/pkg/api/cluster/formatter.go
+++ b/pkg/api/cluster/formatter.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"runtime"
 
 	corev1 "k8s.io/api/core/v1"
 	kubevirtv1 "kubevirt.io/api/core/v1"
@@ -14,7 +15,10 @@ import (
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
+	"github.com/gorilla/mux"
+
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/util"
 )
 
 type Handler struct {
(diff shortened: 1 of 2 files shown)


pkg/api/cluster/formatter.go
```diff
@@ -35,15 +39,21 @@ func (h Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 
-	nodeDeviceAvailability, err := h.generateDeviceAvailability()
-	if err != nil {
-		return err
+	vars := util.EncodeVars(mux.Vars(req))
+	link := vars["link"]
+	var result []byte
+	var err error
+	switch link {
+	case deviceCapacity:
```
ihcsim, line 49: So UI will need to be updated to include this query string, right?

ihcsim (approved):
LGTM - just one question below.

pkg/api/cluster/formatter.go
ibrokethecloud: yeah the UI will need to query the `local` cluster api to find support machine types

bk201 (approved):

Vicente-Cheng (approved):
lgtm thanks!

tserong (approved):

ibrokethecloud:
@mergifyio backport v1.5

pkg/api/cluster/formatter.go
ihcsim: do we have a ticket for the ui change?

a110605: UI PR is here : https://github.com/harvester/harvester-ui-extension/pull/307 @ihcsim

Issue #7323, opened 2025-01-08
[ENHANCEMENT] Improve Single-Replica Volume Detection for Upgrade


WebberHuang1118:
**Is your enhancement related to a problem? Please describe.**

In the upgrade webhook validator, we check [the single-replica condition only for the running volumes](https://github.com/harvester/harvester/blob/9609b9038f3724e584cfc432aaedd0274b6d8cc8/pkg/webhook/resources/upgrade/validator.go#L385-L398). 

However, for the detached volume with the single replica, this situation can lead to the node drain stuck or data integrity concern
- Before `Harvester v1.4.0`, the default [LH node drain policy](https://longhorn.io/docs/1.7.2/references/settings/#node-drain-policy) is `block-if-contains-last-replica`, a detached single-replica volume block the node drain
- Since `Harvester v1.4.0`, we changed the [LH node drain policy to `allow-if-replica-is-stopped`](https://github.com/harvester/harvester-installer/pull/788), even this will make the replica not block node drain, but it can still have potential data integrity concern during the upgrade

**Describe the solution you'd like**

We should detect the single-replica volume situation, both for attached and detached volumes. If the situation happens, we can reject the upgrade or provide the warning message via the [pre-check script](https://github.com/harvester/upgrade-helpers/tree/main/pre-check). cc @bk201 @starbops

WebberHuang1118:
As discussed with @starbops, we plan to have a checkbox, which indicates if the upgrade should be rejected by the single-replica volume (both attached and detached volume).

WebberHuang1118:
This ticket remains GUI integration https://github.com/harvester/harvester/issues/7328

hoo29:
Please could you explain what the `potential data integrity concern during the upgrade` is and if there are scenarios that might lead to data loss?


Edit - it is detailed at https://longhorn.io/docs/1.8.1/maintenance/maintenance/#node-drain-policy-recommendations

lanfon72:
Verified this enhancement has been applied.


Test Information
----
* Environment: **qemu/KVM 2 nodes**
* Harvester Version: **v1.5-a1ce6a16-head**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with at least 2 nodes
2. Create storageclass `lh-sc1` using _Longhorn V1_ with **1** replicas
3. Create an image with storageclass `lh-sc1`
4. Follow test plan in https://github.com/harvester/harvester/pull/7361#issue-2783171559

PR #7361, opened 2025-01-13
enhancement: improve single-replica volume detection for upgrade

WebberHuang1118:
**Problem:**
- Before `Harvester v1.4.0`, the default [LH node drain policy](https://longhorn.io/docs/1.7.2/references/settings/#node-drain-policy) is `block-if-contains-last-replica`, a detached single-replica volume block the node drain
- Since `Harvester v1.4.0`, we changed the [LH node drain policy to `allow-if-replica-is-stopped`](https://github.com/harvester/harvester-installer/pull/788), even this will make the replica not block node drain, but it can still have potential data integrity concern during the upgrade

**Solution:**

we can provide an option via the annotation `harvesterhci.io/skipSingleReplicaDetachedVol` to define how Harvester rejects upgrades or have the warning messages for single-replica volumes
- During the annotation `harvesterhci.io/skipSingleReplicaDetachedVol` not exists (default):
  - If there is any single-replica volume (no matter if it's active or detached), Harvester will reject the upgrade
- During the annotation `harvesterhci.io/skipSingleReplicaDetachedVol` exists:
  - If there is any single-replica volume in the `active` state (including `creating`, `attaching`, and `attached`), Harvester will reject the upgrade
  - Harvester will bypass single-replica volume in the `detached` state
  - This should require a related GUI checkbox

**Related Issue:**
#7323 

**Test plan:**

**Prerequisite**
- Upgrade `harvester-webhook` deployment with this PR
- Creating a VM `vm` with single-replica volume
- Applying a mock version resource for a mock upgrade
  ```
  kind: Version
  metadata:
    name: v1.4.0-mock
    namespace: harvester-system
  spec:
    isoURL: http://192.188.0.50:8080/iso/harvester-mock.iso
  ```

**Case 1: Checking Detached Single-Replica Volume**
- Stopping VM `vm`, and waiting until the related volume is detached
- Applying a mock upgrade
  ```
  curl -k --location 'https://<VIP>/v1/harvester/harvesterhci.io.upgrades/harvester-system' \
  --header 'Content-Type: application/json' \
  --header 'Cookie: <cookie>' \
  --data '{
      "type": "harvesterhci.io.upgrade",
      "metadata": {
          "generateName": "hvst-upgrade-",
          "namespace": "harvester-system"
      },
      "spec": {
          "version": "v1.4.0-mock",
          "logEnabled": true
      }
  }'
  ```
- The upgrade should be rejected with the message:
  - `Following PVCs with single-replica volume, even the volume is detached, upgrade may have potential data integrity concerns: <PVC namespace>/<PVC name>`

**Case 2: Checking Active Single-Replica Volume Only**
- Staring the VM `vm`
- Applying a mock upgrade with annotation `"harvesterhci.io/skipSingleReplicaDetachedVol": "true"`
  ```
  curl -k --location 'https://<VIP>/v1/harvester/harvesterhci.io.upgrades/harvester-system' \
  --header 'Content-Type: application/json' \
  --header 'Cookie: <cookie>' \
  --data '{
      "type": "harvesterhci.io.upgrade",
      "metadata": {
          "annotations": {
               "harvesterhci.io/skipSingleReplicaDetachedVol": "true"
           },
          "generateName": "hvst-upgrade-",
          "namespace": "harvester-system"
      },
      "spec": {
          "version": "v1.4.0-mock",
          "logEnabled": true
      }
  }'
  ```
- The upgrade should be rejected with the message:
`Following PVCs with active single-replica volume, please consider shutdown the corresponding workload before upgrade: <PVC namespace>/<PVC name>"`

### DIFF
--- a/pkg/webhook/resources/upgrade/validator.go
+++ b/pkg/webhook/resources/upgrade/validator.go
@@ -39,6 +39,7 @@ import (
 const (
 	upgradeStateLabel                         = "harvesterhci.io/upgradeState"
 	skipWebhookAnnotation                     = "harvesterhci.io/skipWebhook"
+	skipSingleReplicaDetachedVol              = "harvesterhci.io/skipSingleReplicaDetachedVol"
 	rkeInternalIPAnnotation                   = "rke2.io/internal-ip"
 	managedChartNamespace                     = "fleet-local"
 	defaultNewImageSize                uint64 = 13 * 1024 * 1024 * 1024 // 13GB, this value aggregates all tarball image sizes. It may change in the future.
@@ -143,10 +144,10 @@ func (v *upgradeValidator) Create(_ *types.Request, newObj runtime.Object) error
 		}
 	}
 
-	return v.checkResources(version)
+	return v.checkResources(version, newUpgrade)
 }
 
-func (v *upgradeValidator) checkResources(version *v1beta1.Version) error {
+func (v *upgradeValidator) checkResources(version *v1beta1.Version, upgrade *v1beta1.Upgrade) error {
 	hasDegradedVolume, err := v.hasDegradedVolume()
 	if err != nil {
 		return werror.NewInternalError(err.Error())
@@ -185,7 +186,7 @@ func (v *upgradeValidator) checkResources(version *v1beta1.Version) error {
 		return err
 	}
 
-	if err := v.checkSingleReplicaVolumes(); err != nil {
+	if err := v.checkSingleReplicaVolumes(upgrade); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


bk201 (approved):

bk201:
@WebberHuang1118 can you also add the check to https://github.com/harvester/upgrade-helpers/tree/main/pre-check/v1.1.x

WebberHuang1118:
> @WebberHuang1118 can you also add the check to https://github.com/harvester/upgrade-helpers/tree/main/pre-check/v1.1.x

sure, I'm working on the pre-check script

WebberHuang1118:
> @WebberHuang1118 can you also add the check to https://github.com/harvester/upgrade-helpers/tree/main/pre-check/v1.1.x

PR for pre-check script https://github.com/harvester/upgrade-helpers/pull/19

w13915984028:
A further discussion and review of this enhancement:

1. The main scenarios of single-replica volumes:
a. single-node cluster for trial, POC, test...
b. those VMs which have capacity/hardware... limitations and can only be attached to a specific node, users set a specific storage-class for it
c. those disposable VMs which are used for tasks like CI, users care more about performance/resources than reliability
d. by mistake or accidently create single-replica volume (rarely)
...

In most cases, we can assume that users know those and actively create storageclass for them to get the single-replica volumes, namely, reliability is in low priority.

2. Based on above, v1.4.0 changed `block-if-contains-last-replica` to `allow-if-replica-is-stopped`. 

3. If we still feel:
` it can still have potential data integrity concern during the upgrade`

we may convert this to an upgrade configuration option https://docs.harvesterhci.io/v1.4/advanced/index#upgrade-config; or just refer LH setting, the single-replica volume detection is based on LH setting/upgrade settings, rather than adding a new annotation; note: referring LH settings may not work for third-party CSI.

The annotation is needed for each upgrade CR object, when users are blocked by this, it takes time to find the reason and adopt the solution; but the setting persists.

starbops (approved):
LGTM, thank you!

WebberHuang1118:
> A further discussion and review of this enhancement:
> 
> 1. The main scenarios of single-replica volumes:
>    a. single-node cluster for trial, POC, test...
>    b. those VMs which have capacity/hardware... limitations and can only be attached to a specific node, users set a specific storage-class for it
>    c. those disposable VMs which are used for tasks like CI, users care more about performance/resources than reliability
>    d. by mistake or accidently create single-replica volume (rarely)
>    ...
> 
> In most cases, we can assume that users know those and actively create storageclass for them to get the single-replica volumes, namely, reliability is in low priority.
> 
> 2. Based on above, v1.4.0 changed `block-if-contains-last-replica` to `allow-if-replica-is-stopped`.
> 3. If we still feel:
>    ` it can still have potential data integrity concern during the upgrade`
> 
> we may convert this to an upgrade configuration option https://docs.harvesterhci.io/v1.4/advanced/index#upgrade-config; or just refer LH setting, the single-replica volume detection is based on LH setting/upgrade settings, rather than adding a new annotation; note: referring LH settings may not work for third-party CSI.
> 
> The annotation is needed for each upgrade CR object, when users are blocked by this, it takes time to find the reason and adopt the solution; but the setting persists.

Thanks for the feedback @w13915984028 

IMO, using the upgrade CR's annotation is more convenient and simple. If we put it in the setting, the user may forget he has to take the risk for single-replica volume, or even if the user configures the setting differs from the one who triggers the upgrade.

To put this in the annotation, we can link it with a GUI checkbox (for example `skip detached single-replica volume`), which is by default unchecked. 

If the user is aware of the risk, he can check the box and start the upgrade. In addition, the user can change this on-demand by simply checking/unchecking the box with each upgrade attempt. WDYT? cc @bk201 @starbops

WebberHuang1118:
@mergifyio v1.5

Issue #6844, opened 2024-10-22
[BUG] `VirtualMachine generation is 2, but latest observed generation is 1` is observed on a freshly installed cluster


w13915984028:
**Describe the bug**

After a Harvester v1.4.0-rc3 cluster is freshly installed, create VM, `VirtualMachine generation is 2, but latest observed generation is 1` is observed.

![image](https://github.com/user-attachments/assets/a8366147-f951-4088-ba5e-d192135ec9b2)

Stop VM2
![image](https://github.com/user-attachments/assets/16de9715-958b-47b2-841e-214c9d0afd5f)

and restart VM2, the message is gone.
![image](https://github.com/user-attachments/assets/21c9a4d6-dc21-47fb-bb24-bac5d7574e4c)

create a new vm, no such message.
![image](https://github.com/user-attachments/assets/7f365f17-d208-4254-b5d6-1f716d02e5a6)

It looks to be happen:

When a cluster is newly installed; create a few VMs in the early stage.

**To Reproduce**
1. Check above

**Expected behavior**

no such.

**Support bundle**

[supportbundle_cee1a458-8134-47a9-a1fb-a523c29215eb_2024-10-22T10-06-35Z.zip](https://github.com/user-attachments/files/17474353/supportbundle_cee1a458-8134-47a9-a1fb-a523c29215eb_2024-10-22T10-06-35Z.zip)

**Environment**
 - Harvester ISO version:  Harvester v1.4.0-rc3, 2-node cluster
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): KVM VM is used as Harvester Node.

**Additional context**

Observed when testing https://github.com/harvester/harvester/issues/6766

bk201:
cc @ibrokethecloud too

w13915984028:
When testing v132 upgrade issue https://github.com/harvester/harvester/issues/4005

It looks also happening on v132 cluster:

![image](https://github.com/user-attachments/assets/4a0e9257-dd58-4242-8776-1d92bba8ae2a)

bk201:
Move to v1.5.0 since it only happens in early stage

w13915984028:
The issue was reproduced with following debug logs, I will add a PR to fix it.

```
time="2024-12-10T15:41:10Z" level=debug msg="VM vm11 in namespace default is starting"
[...]

PR #7402, opened 2025-01-20
Fix6844 Backfill VMI MAC to VM and sync VM generation

w13915984028:

**Problem:**

vm, vmi, ControllerRevision are not well synced randomly

1. When VMI MAC is backfilled to VM, the Generation and ObservedGeneration are different;
2. In race case, the backfilled MAC may overwrite VM addtional changes (other than MAC) to controllerRevision.

**Solution:**

1. Make sure those vm, vmi and controllerRevision are well synced
2. Only backfill MAC, avoid to overwrite other data

**Related Issue:**
https://github.com/harvester/harvester/issues/6844

**Test plan:**

Test cases:
1. Create a VM without setting any MACs, VMI will get dynamic MACs and sync back to VM
3. Restore a saved VM, it is simlary to 1, VMI will get new dynamic MAC and sync back to VM
4. Create a VM with setting all interfaces with MACs, VMI inheriate them

vm Generation is same as ObservedGeneration, UI does not show generation difference

5. Create a VM witout setting any MACs, when VMI is starting, edit VM yaml to set one/all interfaces with MAC and do not restart VM, the still-empty MAC interface will be synced back, finally the Generation & ObservedGeneration are different

image
while MAC is backfilling, the VM spec is modified; VM shows both updates
image
controllerrevision is updated with the vmi's MAC
image

VMI MAC is backfilled, but the new annotation on VM is not synced to ControllerRevision; which requres a VM rebooting

6. Create  a VM witout setting any MACs, , when VMI is starting, edit VM yaml to set other fields and do not restart VM, the still-empty MAC interface will be synced back, finally the Generation & ObservedGeneration are different
7. Create a VM witout setting any MACs, when VMI is starting, edit VM yaml to set all interfaces with MAC and do not restart VM, finally the Generation & ObservedGeneration are different

### DIFF
--- a/pkg/controller/master/virtualmachine/register.go
+++ b/pkg/controller/master/virtualmachine/register.go
@@ -10,7 +10,8 @@ import (
 const (
 	vmControllerCreatePVCsFromAnnotationControllerName           = "VMController.CreatePVCsFromAnnotation"
 	vmiControllerReconcileFromHostLabelsControllerName           = "VMIController.ReconcileFromHostLabels"
-	vmControllerSetDefaultManagementNetworkMac                   = "VMController.SetDefaultManagementNetworkMacAddress"
+	vmiControllerSetDefaultManagementNetworkMac                  = "VMIController.SetDefaultManagementNetworkMacAddress"
+	vmiControllerSyncControllerRevision                          = "VMIController.SyncControllerRevision"
 	vmControllerStoreRunStrategyControllerName                   = "VMController.StoreRunStrategyToAnnotation"
 	vmControllerSyncLabelsToVmi                                  = "VMController.SyncLabelsToVmi"
 	vmControllerSetHaltIfInsufficientResourceQuotaControllerName = "VMController.SetHaltIfInsufficientResourceQuota"
@@ -88,15 +89,17 @@ func Register(ctx context.Context, management *config.Management, _ config.Optio
 	virtualMachineInstanceClient.OnChange(ctx, vmiControllerSetHaltIfOccurExceededQuotaControllerName, vmiCtrl.StopVMIfExceededQuota)
 	virtualMachineInstanceClient.OnChange(ctx, vmiControllerRemoveDeprecatedFinalizerControllerName, vmiCtrl.removeDeprecatedFinalizer)
 
-	// register the vm network controller upon the VMI changes
+	// register the vmi network controller upon the VMI changes
+	var virtualMachineInstanceController = management.VirtFactory.Kubevirt().V1().VirtualMachineInstance()
 	var vmNetworkCtl = &VMNetworkController{
-		vmClient:  vmClient,
-		vmCache:   vmCache,
-		vmiClient: virtualMachineInstanceClient,
-		crClient:  crClient,
-		crCache:   crClientCache,
+		vmClient: vmClient,
+		vmCache:  vmCache,
+		//vmiClient:     virtualMachineInstanceController,
+		crClient:      crClient,
+		crCache:       crClientCache,
+		vmiController: virtualMachineInstanceController,
 	}
-	virtualMachineInstanceClient.OnChange(ctx, vmControllerSetDefaultManagementNetworkMac, vmNetworkCtl.SetDefaultNetworkMacAddress)
+	virtualMachineInstanceController.OnChange(ctx, vmiControllerSetDefaultManagementNetworkMac, vmNetworkCtl.SyncMacAddressAndControllerRevision)
 
 	return nil
 }
--- a/pkg/controller/master/virtualmachine/register.go
+++ b/pkg/controller/master/virtualmachine/register.go
@@ -10,8 +10,7 @@ import (
 const (
 	vmControllerCreatePVCsFromAnnotationControllerName           = "VMController.CreatePVCsFromAnnotation"
 	vmiControllerReconcileFromHostLabelsControllerName           = "VMIController.ReconcileFromHostLabels"
-	vmiControllerSetDefaultManagementNetworkMac                  = "VMIController.SetDefaultManagementNetworkMacAddress"
-	vmiControllerSyncControllerRevision                          = "VMIController.SyncControllerRevision"
+	vmiControllerSyncMacAddressAndControllerRevision             = "VMIController.SyncMacAddressAndControllerRevision"
 	vmControllerStoreRunStrategyControllerName                   = "VMController.StoreRunStrategyToAnnotation"
 	vmControllerSyncLabelsToVmi                                  = "VMController.SyncLabelsToVmi"
 	vmControllerSetHaltIfInsufficientResourceQuotaControllerName = "VMController.SetHaltIfInsufficientResourceQuota"
@@ -92,14 +91,13 @@ func Register(ctx context.Context, management *config.Management, _ config.Optio
 	// register the vmi network controller upon the VMI changes
 	var virtualMachineInstanceController = management.VirtFactory.Kubevirt().V1().VirtualMachineInstance()
 	var vmNetworkCtl = &VMNetworkController{
-		vmClient: vmClient,
-		vmCache:  vmCache,
-		//vmiClient:     virtualMachineInstanceController,
+		vmClient:      vmClient,
+		vmCache:       vmCache,
 		crClient:      crClient,
 		crCache:       crClientCache,
 		vmiController: virtualMachineInstanceController,
 	}
-	virtualMachineInstanceController.OnChange(ctx, vmiControllerSetDefaultManagementNetworkMac, vmNetworkCtl.SyncMacAddressAndControllerRevision)
+	virtualMachineInstanceController.OnChange(ctx, vmiControllerSyncMacAddressAndControllerRevision, vmNetworkCtl.SyncMacAddressAndControllerRevision)
 
 	return nil
 }
--- a/pkg/controller/master/virtualmachine/vmi_network_controller.go
+++ b/pkg/controller/master/virtualmachine/vmi_network_controller.go
@@ -30,50 +30,51 @@ func (h *VMNetworkController) SyncMacAddressAndControllerRevision(id string, vmi
 	}
 
 	// ensure vmi mac is backfilled to vm
-	var retryvmi = false
-	var err error = nil
-	if vmi, err = h.syncNetworkDynamicMacAddress(vmi, &retryvmi); err != nil {
+	retryvmi := false
+	if vmi, err := h.syncNetworkDynamicMacAddress(vmi, &retryvmi); err != nil {
 		return vmi, err
 	}
 
 	// ensure vmi mac is backfilled to rc
-	var retryrc = false
-	if vmi, err = h.syncControllerRevisionFromVmi(vmi, &retryrc); err != nil {
+	retrycr := false
+	if vmi, err := h.syncControllerRevision(vmi, &retrycr); err != nil {
 		return vmi, err
 	}
 
 	// vmi may end onChange, but vm/rc is not synced yet
-	if retryvmi || retryrc {
+	if retryvmi || retrycr {
 		h.vmiController.EnqueueAfter(vmi.Namespace, vmi.Name, time.Second*1)
 		return vmi, nil
 	}
 
-	// last step: if vm did not sync with rc, update it
-
-	return h.updateVm(vmi)
+	// last step: if vm still did not sync with rc, update it
+	return h.updateVM(vmi)
 }
 
-// when user does not specify mac address when creating a VM
+// if user does not specify mac address when creating a VM
 // guest OS will use the initial allocated mac address from its DHCP config
-// when VM restarts that mac address may change, and related IP is not persistant
-// SyncNetworkDynamicMacAddress saves the dynamic mac address observed from the the VMI status,
+// when VM restarts that mac address may change, and related IP is not persistent
+// syncNetworkDynamicMacAddress saves the dynamic mac address observed from the the VMI status to vm
 func (h *VMNetworkController) syncNetworkDynamicMacAddress(vmi *kubevirtv1.VirtualMachineInstance, retry *bool) (*kubevirtv1.VirtualMachineInstance, error) {
 	vm, err := h.vmCache.Get(vmi.Namespace, vmi.Name)
 	if err != nil {
 		return vmi, err
 	}
 
+	if vm == nil || vm.DeletionTimestamp != nil {
+		return vmi, nil
+	}
+
 	vmCopy := vm.DeepCopy()
 	vmiInterfaces := getVmiInterfaceMacs(vmi)
-
 	updated := false
 	for i, vmIface := range vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces {
 		if vmIface.Name == "" || vmIface.MacAddress != "" {
 			continue
 		}
 		if macAddress := vmiInterfaces[vmIface.Name]; macAddress != "" {
-			// only set the network mac address when it is empty
-			logrus.Debugf("set VM %s management network %s macAddress to %s", vm.Name, vmIface.Name, macAddress)
+			// only backfill the network mac address when it is empty
+			logrus.Infof("set VM %s network %s macAddress to %s", vm.Name, vmIface.Name, macAddress)
 			vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces[i].MacAddress = macAddress
 			updated = true
 		}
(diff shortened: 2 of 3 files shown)


w13915984028:
@mergify backport v1.4 v1.5

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -12,107 +13,179 @@ import (
...
+	return vmi, nil
 }
 
-func (h *VMNetworkController) regenerateControllerRevision(vmi *kubevirtv1.VirtualMachineInstance, vm *kubevirtv1.VirtualMachine) error {
+func (h *VMNetworkController) regenerateControllerRevision(vmi *kubevirtv1.VirtualMachineInstance, vm *kubevirtv1.VirtualMachine, retry *bool) error {
 	crObj, err := h.crCache.Get(vmi.Namespace, vmi.Status.VirtualMachineRevisionName)
 	if err != nil {
-		return fmt.Errorf("error fetch controller revision object for vmi %s-%s: %v", vmi.Name, vmi.Namespace, err)
+		// controllerRevision has a different name than vm name, essential to log it's name in error
+		return fmt.Errorf("failed to fetch controllerRevision %s/%s: %w", vmi.Namespace, vmi.Status.VirtualMachineRevisionName, err)
```
ihcsim, line 139: IIUC, even with this PR, there is still a chance that we can encounter the `not found, requeuing` error if the cache isn't sync'ed. Is it worth retrying the `GET` here, in case the error is transient? The error also seems to imply some requeuing is happening.

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -12,107 +13,179 @@ import (
...
-	if !equality.Semantic.DeepEqual(revisionSpec.Spec.Template, vm.Spec.Template) && vm.Generation != vm.Status.ObservedGeneration {
-		patch, patchGenError := patchVMRevision(vm)
-		if patchGenError != nil {
-			return fmt.Errorf("error during patch generation for vmi %s-%s: %v", vmi.Name, vmi.Name, patchGenError)
+
+	revisionSpecCopy := revisionSpec
+	vmInterfaceMacs := getVMInterfaceMacs(vm)
+	vmiInterfaceMacs := getVmiInterfaceMacs(vmi)
+	updated := false
+	for i, rcIface := range revisionSpec.Spec.Template.Spec.Domain.Devices.Interfaces {
```
ihcsim, line 151: Is it possible for the vm or vmi to have interfaces that aren't present in the controller revision?

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -12,107 +13,179 @@ import (
...
 
-	for i, vmIface := range vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces {
-		macAddress, ok := vmiInterfaces[vmIface.Name]
-		// only set the network mac address when it has no existing value
-		if ok && vmIface.MacAddress == "" {
-			logrus.Debugf("set VM %s management network %s macAddress to %s", vm.Name, vmIface.Name, macAddress)
-			vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces[i].MacAddress = macAddress
-		}
+func (h *VMNetworkController) syncControllerRevision(vmi *kubevirtv1.VirtualMachineInstance, retry *bool) (*kubevirtv1.VirtualMachineInstance, error) {
+	vm, err := h.vmCache.Get(vmi.Namespace, vmi.Name)
```
ihcsim, line 114: Seems like we keep repeating this call in all the private methods. Is the intention here to ensure we always work with the latest vm? Otherwise, we can just call it once inside `SyncMacAddressAndControllerRevision()`, right?

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -12,107 +13,179 @@ import (
...
+	}
+
+	// ensure vmi mac is backfilled to rc
+	retrycr := false
+	if vmi, err := h.syncControllerRevision(vmi, &retrycr); err != nil {
+		return vmi, err
 	}
 
-	if len(vmi.Status.Interfaces) == 0 {
+	// vmi may end onChange, but vm/rc is not synced yet
```
ihcsim, line 44: I am not sure I understand why vm and rc won't be sync'ed. Both `h.syncNetworkDynamicMacAddress()` and `h.syncControllerRevision` return with `Update()` completed. Or are we waiting for kubevirt side to sync up?

pkg/controller/master/virtualmachine/vmi_network_controller.go
w13915984028: in rancher warngler & k8s controller mode,  most controller does not loop wait a cache to be synced in the normal run time, that will block current go thread and nobody knows exactly how long/how many times(e.g. try 3 times get) to wait;  instead, the controller reconciller mechanism ensure this, after return error, the controller will be retriggered. a returned & logged error like `not found` is normal in the pod, especially `query, delete, create & query` an object in a row

pkg/controller/master/virtualmachine/vmi_network_controller.go
w13915984028: controllerrevision is a copy of VM spec when VM is started, controllerrevision is (re)created right then , and it can't be modifed unless deleted&created again.   Harvester controller is hacking by deleting & recreating the CR to cheat kubevirt

when VM has new interface added or interface changed or other spce change, then they are different, and kubevirt/k8s will update VM.generation to new, but observedgeneration is kept old (or vice versa). that's how we get a hint from the UI that there are generation differences, only after re-start VM, the differences on VM spec can be copied to new VMI

pkg/controller/master/virtualmachine/vmi_network_controller.go
w13915984028: yes, the generation is updated by apiserver when backfill MAC;  observedgeneration is updated by kubevirt controller, we need to wait kubevirt finishs its work

refer https://github.com/harvester/harvester/issues/6844#issuecomment-2551020204,

pkg/controller/master/virtualmachine/vmi_network_controller.go
w13915984028: yes, work on the latest VM

besides harvester contrller backfills the MAC;  user may edit VM at the meantime (e.g. set a MAC, add a new volume), we can't use VM spec to overwirte controllerrevision wholely or vice versa;  strictly, original code does not handle this race case well. https://github.com/harvester/harvester/pull/7402#issue-2798817503  test case 5

w13915984028:
BTW: When testing https://github.com/harvester/harvester/pull/7504, the generation issue seems easy to occur:

image

image

w13915984028:
As the generation is quite easy to reproduce with the PR https://github.com/harvester/harvester/pull/7504, I merged https://github.com/harvester/harvester/pull/7504 with https://github.com/harvester/harvester/pull/7402 and tested again, the generation warnings are gone.

The log also shows that, with the PR 7402,  VMs are forcely to sync `"VM test3-ns/vm1 is forcely updated to sync vm.Generation 2, vm.Status.ObservedGeneration 1"` when gaps are existing.

image

```
time="2025-02-13T10:12:59Z" level=info msg="set VM vm1 network default macAddress to 8e:80:00:56:82:94"
time="2025-02-13T10:12:59Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test3-ns/vm1 has running pod, skip check"
time="2025-02-13T10:12:59Z" level=info msg="VM test3-ns/vm1 delete and create new controllerRevision revision-start-vm-ebd494a3-ae69-4f1c-8c19-d6eab043da92-1 to sync generation: vm.Generation 2, vm.Status.ObservedGeneration 1"

// forcely update to sync
time="2025-02-13T10:12:59Z" level=debug msg="VM test3-ns/vm1 is forcely updated to sync vm.Generation 2, vm.Status.ObservedGeneration 1"
time="2025-02-13T10:12:59Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test3-ns/vm1 has running pod, skip check"
time="2025-02-13T10:12:59Z" level=debug msg="VM test3-ns/vm1 is forcely updated to sync vm.Generation 2, vm.Status.ObservedGeneration 1"
time="2025-02-13T10:12:59Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test3-ns/vm1 has running pod, skip check"
time="2025-02-13T10:12:59Z" level=debug msg="Wrote ping"
...

time="2025-02-13T10:13:34Z" level=info msg="set VM vm5 network default macAddress to a6:e7:7f:0e:7e:04"
time="2025-02-13T10:13:34Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test6-ns/vm5 has running pod, skip check"
time="2025-02-13T10:13:34Z" level=info msg="set VM vm5 network default macAddress to a6:e7:7f:0e:7e:04"
time="2025-02-13T10:13:34Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test6-ns/vm5 has running pod, skip check"
time="2025-02-13T10:13:35Z" level=info msg="set VM vm5 network default macAddress to a6:e7:7f:0e:7e:04"
time="2025-02-13T10:13:35Z" level=info msg="set VM vm5 network default macAddress to a6:e7:7f:0e:7e:04"
time="2025-02-13T10:13:35Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test6-ns/vm5 has running pod, skip check"
time="2025-02-13T10:13:35Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test6-ns/vm5 has running pod, skip check"
time="2025-02-13T10:13:35Z" level=info msg="VM test6-ns/vm5 delete and create new controllerRevision revision-start-vm-6c2a1ea2-5c25-4346-9c19-3f4eef96352d-1 to sync generation: vm.Generation 2, vm.Status.ObservedGeneration 1"

// forcely update to sync
time="2025-02-13T10:13:35Z" level=debug msg="VM test6-ns/vm5 is forcely updated to sync vm.Generation 2, vm.Status.ObservedGeneration 1"
time="2025-02-13T10:13:35Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test6-ns/vm5 has running pod, skip check"
time="2025-02-13T10:13:35Z" level=debug msg="VM test6-ns/vm5 is forcely updated to sync vm.Generation 2, vm.Status.ObservedGeneration 1"
time="2025-02-13T10:13:35Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test6-ns/vm5 has running pod, skip check"
time="2025-02-13T10:13:36Z" level=debug msg="CheckIfVMCanStartByResourceQuota: VM test6-ns/vm5 has running pod, skip check"
```

w13915984028:
@bk201 @ibrokethecloud @ihcsim @FrankYang0529 

Please take a new look of above comparing tests, this PR is essential. Please help review PR, thanks.

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -12,107 +13,179 @@ import (
...
+	}
+
+	vmCopy := vm.DeepCopy()
+	vmiInterfaces := getVmiInterfaceMacs(vmi)
+	updated := false
+	for i, vmIface := range vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces {
+		if vmIface.Name == "" || vmIface.MacAddress != "" {
+			continue
+		}
+		if macAddress := vmiInterfaces[vmIface.Name]; macAddress != "" {
```
ibrokethecloud, line 75: do we need the extra check here to verify `macAddress` is not empty?
When we populate the map in `getVMInterfaces` we have already checked macAddress is not empty. It should not be possible to find an empty `macAddress` by the time the map lookup is performed.

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -12,107 +13,179 @@ import (
...
+	}
+
 	return vmi, nil
 }
 
-func (h *VMNetworkController) updateVMDefaultNetworkMacAddress(vmi *kubevirtv1.VirtualMachineInstance) error {
-	vm, err := h.vmCache.Get(vmi.Namespace, vmi.Name)
-	if err != nil {
-		return err
+func getVMInterfaceMacs(vm *kubevirtv1.VirtualMachine) map[string]string {
```
ibrokethecloud, line 93: we could refactor this into a common method which has input of `[]devices.Interfaces` and return the map[string]string.
This way it can be used across both VM's and VMI's

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -129,3 +202,65 @@ func patchVMRevision(vm *kubevirtv1.VirtualMachine) ([]byte, error) {
...
+	if !equality.Semantic.DeepEqual(revisionSpec.Spec, vm.Spec) {
+		return vmi, nil
+	}
+
+	// forcely update vm, then upstream controller will sync the Generation and ObservedGeneration
+	vmCopy := vm.DeepCopy()
+	if vmCopy.Annotations == nil {
+		vmCopy.Annotations = make(map[string]string, 1)
+	}
+	vmCopy.Annotations[util.AnnotationTimestamp] = time.Now().Format(time.RFC3339)
```
ibrokethecloud, line 259: is the annotationTimestamp used by the UI? I see no actual logic implemented in the controllers to actually perform additional actions when this annotation is set.

pkg/controller/master/virtualmachine/vmi_network_controller.go
w13915984028: That's for below case:

1.1 User creates a VM with 2 interfaces, and start the VM
1.2 User adds an additional interface to VM, but do not restart the VM right now
1.3 In this loop, with VM's interface to check VMI's interface map, it may get an empty one.

pkg/controller/master/virtualmachine/vmi_network_controller.go
w13915984028: this `annotationTimestamp` has two usages:

(1) Let kubevirt's controller be triggered, in this context

(2) UI has a websocket watching VM object (not all realted vmi, vmim...), when VM object has change, it will be notified, and then render the menu, e.g., the vmim controller call below function when vmim object status changes, it helps UI to update `migrate/abortMigration` button.

https://github.com/harvester/harvester/blob/b4b8aeb7ba260e8d01d306cf316d6611b3d01d53/pkg/controller/master/migration/vmim_controller.go#L86

https://github.com/harvester/harvester/blob/b4b8aeb7ba260e8d01d306cf316d6611b3d01d53/pkg/controller/master/migration/vmim_controller.go#L237

ibrokethecloud: should have known we were updating the vm object and not the vmi object. if we could please address the feedback rest looks good.

FrankYang0529:
Hi @w13915984028, I have another PR tries to solve mac address issue. My idea is to update mac address to VM when VM is stopped, so we don't need to touch controller revision.

https://github.com/harvester/harvester/pull/7084

pkg/controller/master/virtualmachine/register.go
```diff
@@ -88,15 +88,16 @@ func Register(ctx context.Context, management *config.Management, _ config.Optio
 	virtualMachineInstanceClient.OnChange(ctx, vmiControllerSetHaltIfOccurExceededQuotaControllerName, vmiCtrl.StopVMIfExceededQuota)
 	virtualMachineInstanceClient.OnChange(ctx, vmiControllerRemoveDeprecatedFinalizerControllerName, vmiCtrl.removeDeprecatedFinalizer)
 
-	// register the vm network controller upon the VMI changes
+	// register the vmi network controller upon the VMI changes
+	var virtualMachineInstanceController = management.VirtFactory.Kubevirt().V1().VirtualMachineInstance()
```
ihcsim, line 92: Does it make a difference by attaching the handler to this new instance vs. attaching it to the existing client at: https://github.com/harvester/harvester/blob/fce3808c7626f0db50faf328ce36c872f5a3576f/pkg/controller/master/virtualmachine/register.go#L78

ihcsim (approved):
LGTM. This is a tricky one. Thanks for working on it :+1:. 

Honestly, I don't know if our controller is the right place to handle this backfill mechanism, because in order to satisfy our use cases, we need to tiptoe around kubevirt's reconciliation logic. If their code change, we might be back to square one. But I also can't think of any alternative right now :shrug:. So far, the only seemingly reliable option is to utilize https://github.com/k8snetworkplumbingwg/kubemacpool. Maybe eventually, it's worth an investigation. 

Finally, I think this is orthogonal to the annotation approach in #7084 because they solve different problems.

ihcsim:
Maybe in a separate PR, we should look into adding integration or e2e tests, if not already there.

w13915984028:
BTW: I am also checking @FrankYang0529 's new solution on PR https://github.com/harvester/harvester/pull/7084. When most of the test cases can be covered by the annotation based solution, then we may not need this one. As @ihcsim already mentioned, this hack solution can break at any time in the future.

Issue #7375, opened 2025-01-14
[BUG] Fail to trigger v1.4.0 to v1.4.1-rc1 upgrade on witness cluster due to managed chart harvester is not ready


albinsun:
**Describe the bug**
`admission webhook "validator.harvesterhci.io" denied the request: managed chart harvester is not ready, please wait for it to be ready`

Similiar to
1. harvester/harvester/issues/6561
2. harvester/harvester/issues/5917

**To Reproduce**
1. <details>
   <summary> Fresh install a 3 nodes witness cluster
   </summary>
   
   ![Image](https://github.com/user-attachments/assets/bbddb0d8-33f0-4ba6-aeb5-18c7d1ca40c1)
   </details>

1. <details>
   <summary> Create a storage class <i>sc2</i> with replica 2 and <b>Set as Default</b>
   </summary>
   
   ![Image](https://github.com/user-attachments/assets/7cd4526d-2ad1-48c5-910f-c5f666f769ec)
   </details>

1. :x: Prepare and trigger an air-gapped upgrade
   ![image](https://github.com/user-attachments/assets/ae2bb50c-33e5-4292-b41d-e0560b14eef8)

**Expected behavior**
Witness cluster w/ default replica 2 storage class should be able to upgrade.
Should provide doc tf still need to manually patch configs.

**Support bundle**
[supportbundle-witness_upgrade_fail_managedchart.zip](https://github.com/user-attachments/files/18409421/supportbundle-witness_upgrade_fail_managedchart.zip)

**Environment**
 - Harvester ISO version: `v1.4.0` -> `v1.4.0-rc1`
 - Underlying Infrastructure: QEMU/KVM AMD64

**Additional context**
1. ManagedChart fleet-local/harvester
   `cannot patch "harvester-longhorn" with kind StorageClass: admission webhook "validator.harvesterhci.io" denied the request: default storage class %!s(MISSING) already exists, please reset it first`
   ![image](https://github.com/user-attachments/assets/897e6317-bcf9-4459-994a-13315865a345)
   ![image](https://github.com/user-attachments/assets/3ba55d6e-f532-4c77-bae9-98c0750d663a)

bk201:
@FrankYang0529 please help add a known issue page, the two PRs take effect in 1.4.1.

innobead:
Do we still need to add this to 1.4.1? @tserong @FrankYang0529 

Or it is good to postpone to 1.4.2? I think the fix wont affect  normal cases.

[...]

PR #7421, opened 2025-01-21
Selectively update SC harvester-longhorn default SC annotation

w13915984028:

**Problem:**

When harvester managedchart is updated, it may update the default sc, but denied by webhook

**Solution:**

1. If `storageclass.kubernetes.io/is-default-class: "true"` is not set on any sc, respect the chart value
2. If `storageclass.kubernetes.io/is-default-class: "true"` is on Harvester-longhorn, keep it true.
3. If `storageclass.kubernetes.io/is-default-class: "true"` is set on other SC, set it to false on harvester-longhorn, do not follow the chart value.

**Related Issue:**
https://github.com/harvester/harvester/issues/7375

**Test plan:**

See solution.

**Note**
There has been a block of code to process the default SC on upgrade script:
https://github.com/harvester/harvester/blob/41d3e1b5b1b5d3e9d3a91568638ab8c5b4071f8b/package/upgrade/upgrade_manifests.sh#L893C1-L896C5
```
  local sc=$(kubectl get sc -o json | jq '.items[] | select(.metadata.annotations."storageclass.kubernetes.io/is-default-class" == "true" and .metadata.name != "harvester-longhorn")')
  if [ -n "$sc" ] && [ "$UPGRADE_PREVIOUS_VERSION" != "v1.0.3" ]; then
      yq e '.spec.values.storageClass.defaultStorageClass = false' -i harvester.yaml
  fi
```

### DIFF
--- a/deploy/charts/harvester/templates/_helpers.tpl
+++ b/deploy/charts/harvester/templates/_helpers.tpl
@@ -123,3 +123,35 @@ Get Support-bundle-kit image environment for updating the default values per cur
 {{- end -}}
 {{- end -}}
 {{- end }}
+
+{{/*
+Get a dynamic storageclass.kubernetes.io/is-default-class value for harvester-longhorn storageclass.
+*/}}
+{{- define "harvester.annotations.defaultstorageclass" -}}
+{{- if .Values.storageClass.defaultStorageClass }}
+ {{- $allscs := (lookup "storage.k8s.io/v1" "StorageClass" "" "") -}}
+ {{- if eq (len $allscs ) 0 -}}
+storageclass.kubernetes.io/is-default-class: "true"
+ {{- else -}}
+ {{- $scname := "" -}}
+ {{- range $index, $cursc := $allscs.items -}}
+ {{- range $k, $v := $cursc.metadata.annotations -}}
+ {{- if eq $k "storageclass.kubernetes.io/is-default-class" -}}
+ {{- if eq $v "true" -}}
+ {{- $scname = $cursc.metadata.name -}}
+ {{- end -}}
+ {{- end -}}
+ {{- end -}}
+ {{- end -}}
+ {{- if eq $scname "" }}
+storageclass.kubernetes.io/is-default-class: "true"
+ {{- else if eq $scname "harvester-longhorn" -}}
+storageclass.kubernetes.io/is-default-class: "true"
+ {{- else -}}
+storageclass.kubernetes.io/is-default-class: "false"
+ {{- end }}
+ {{- end }}
+{{- else -}}
+storageclass.kubernetes.io/is-default-class: "false"
+{{- end }}
+{{- end }}
--- a/deploy/charts/harvester/templates/harvester-storageclass.yaml
+++ b/deploy/charts/harvester/templates/harvester-storageclass.yaml
@@ -5,9 +5,7 @@ metadata:
   name: harvester-longhorn
   annotations:
     harvesterhci.io/is-reserved-storageclass: "true"
-{{- if .Values.storageClass.defaultStorageClass }}
-    storageclass.kubernetes.io/is-default-class: "true"
-{{- end }}
+{{ include "harvester.annotations.defaultstorageclass" . | indent 4 }}
 provisioner: driver.longhorn.io
 allowVolumeExpansion: true
 reclaimPolicy: "{{ .Values.storageClass.reclaimPolicy }}"


w13915984028:
Note: The PR was tested on `master-head`, it works well. the `harvester-longhorn` won't set self as default if any other SC has that annotation.

But encountered another issue https://github.com/harvester/harvester/issues/7441, the issue was also confirmed on v1.4.0.

Use combination PR https://github.com/harvester/harvester-installer/pull/941 to build ISO and test the overall fix.

w13915984028:
@mergify backport v1.5

deploy/charts/harvester/templates/_helpers.tpl
```diff
@@ -123,3 +123,35 @@ Get Support-bundle-kit image environment for updating the default values per cur
...
+ {{- if eq $k "storageclass.kubernetes.io/is-default-class" -}}
+ {{- if eq $v "true" -}}
+ {{- $scname = $cursc.metadata.name -}}
+ {{- end -}}
+ {{- end -}}
+ {{- end -}}
+ {{- end -}}
+ {{- if eq $scname "" }}
+storageclass.kubernetes.io/is-default-class: "true"
+ {{- else if eq $scname "harvester-longhorn" -}}
```
ibrokethecloud, line 148: do we need to check if sc is `harvester-longhorn`? As long as there is another sc which is the default sc we should just let it be the default sc

FrankYang0529 (approved):
LGTM. I can upgrade with anther storage class without error.

deploy/charts/harvester/templates/_helpers.tpl
w13915984028: the check is essential, it identifies who has the `default` anno, if it is still `harvester-longhorn` then keep it, if it is others, `harvester-longhorn` needs to set `default` as `false` (not the managedchart input) to avoid denied by webhook

deploy/charts/harvester/templates/_helpers.tpl
```diff
@@ -123,3 +123,35 @@ Get Support-bundle-kit image environment for updating the default values per cur
 {{- end -}}
 {{- end -}}
 {{- end }}
+
+{{/*
+Get a dynamic storageclass.kubernetes.io/is-default-class value for harvester-longhorn storageclass.
+*/}}
+{{- define "harvester.annotations.defaultstorageclass" -}}
+{{- if .Values.storageClass.defaultStorageClass }}
+ {{- $allscs := (lookup "storage.k8s.io/v1" "StorageClass" "" "") -}}
```
ihcsim, line 132: I am both :smile: and :disappointed: that helm has something like `lookup`.

ihcsim: To confirm my understanding, the issue is that during upgrade, Helm would always try to reset the `harvester-longhorn` storage class to be the default one, right? Instead of looping through all the storage classes, can we just look up if the `harvester-longhorn` storage class already exists, and if yes, re-use the value of its `is-default-class` annotation?

w13915984028: more than upgrading;    user can also `kubectl edit managedchart harvester` to set the `Values.storageClass.defaultStorageClass` to `true` or `false` at any time

ihcsim: I'd think `k edit` scenarios like that should be checked by something like the admission webhook or policy. IMO, the chart should only be responsible for providing sane default value during install and upgrade. The logic to assess and enforce the global state (i.e. only one unique sc) should belong elsewhere.
I imagine enforcing this within the chart also makes it harder to reason about. E.g., if a keen user wanted to mod the managed chart, and our built-in logic decided to override the user's intention.

w13915984028: I agree with your point.

On the other hand, some users are not UI preferred, they may operate all those existing CRD objects from kubectl/restapi.  And, Harvester upgrade script itself is also widely using `kubectl` to patch/upgrade the related objects. e.g. https://github.com/harvester/harvester/blob/49f54b3f0e5c82578efb6e0634e1ad7bdbecd7cc/package/upgrade/upgrade_manifests.sh#L393

Use webhook to deny those changes on managedchart is also a solution, e.g. if another SC has the `is-default` annotation, then does not allow `harvester-longhorn` to set itself as default.   Of course very careful attention is needed, to avoid blocking the upgrade.

This PR, targets the specific issue, avoid upgrade is blocked by webhook.

WebberHuang1118: > I'd think `k edit` scenarios like that should be checked by something like the admission webhook or policy. IMO, the chart should only be responsible for providing sane default value during install and upgrade. The logic to assess and enforce the global state (i.e. only one unique sc) should belong elsewhere. I imagine enforcing this within the chart also makes it harder to reason about. E.g., if a keen user wanted to mod the managed chart, and our built-in logic decided to override the user's intention.

I have the same opinion as @ihcsim,

Can the chart only be responsible for providing the default value? and having the logic elsewhere? @w13915984028

w13915984028: @WebberHuang1118 @ihcsim 

The Harvester managedchart, in most scenarios, is the only source of truth of Harvester related resources, more than the default values. And, `kubectl` is the main tool to operate it manually & via upgrade script.  Please refer:

https://github.com/w13915984028/harvester-develop-summary/blob/main/harvester-managedchart-lifecycle-management.md

There has been a block of code to process the default SC on upgrade script:
https://github.com/harvester/harvester/blob/41d3e1b5b1b5d3e9d3a91568638ab8c5b4071f8b/package/upgrade/upgrade_manifests.sh#L893C1-L896C5

```
  local sc=$(kubectl get sc -o json | jq '.items[] | select(.metadata.annotations."storageclass.kubernetes.io/is-default-class" == "true" and .metadata.name != "harvester-longhorn")')
  if [ -n "$sc" ] && [ "$UPGRADE_PREVIOUS_VERSION" != "v1.0.3" ]; then
      yq e '.spec.values.storageClass.defaultStorageClass = false' -i harvester.yaml
  fi
```

ihcsim: Right, so the main concern is that user's actions (via UI, `kubectl annotate sc` etc.) can cause the values in the managed chart to go stale, correct? Once it's stale, the managed chart is no longer the source of truth. As far as a user is concerned, if their changes are accepted by Harvester, _that_ is the source of truth, especially when the UI and `kubectl` both confirm so. Do you agree that there needs to be a tighter reconciliation between what the user can do to these [special data](https://github.com/w13915984028/harvester-develop-summary/blob/main/harvester-managedchart-lifecycle-management.md#special-data) and what the managed chart holds?

w13915984028: Yeah. For those special data, we process them case by case.

The `default sc` option is exposed to UI, user's configurtion (may also from kubectl/restapi upon SC object) is more respected.

As written in the issue description, there are a couple of cases.

```
Solution:

If storageclass.kubernetes.io/is-default-class: "true" is not set on any sc, respect the chart value
If storageclass.kubernetes.io/is-default-class: "true" is on Harvester-longhorn, keep it true.
If storageclass.kubernetes.io/is-default-class: "true" is set on other SC, set it to false on harvester-longhorn, do not follow the chart value.
```

Of course there is another option:

Harvester controller watchs the SC, sync `harvester-longhorn`'s annotation back to managedchart object. 

Due to the `managedchart harvester` is a heavy big object, any change to it's spec will cause `fleet-agent` call helm to update the whole chart, it may have other side-effects.

At the moment, we normally do not write back to `managedchart harvester` from controller.

ihcsim: Good point on not writing back to the managed chart. I guess the hard thing for me to reconcile is that if a user sets the `Values.storageClass.defaultStorageClass` to `true`, it doesn't guarantee that it will always be `true`. The source of truth lies in the annotation of the SC, not the managed chart. What's the real value to a user of exposing `.Values.storageClass.defaultStorageClass` in the managed chart? Otherwise, LGTM. Thanks.

ihcsim (approved):
Thanks for working on this.

deploy/charts/harvester/templates/_helpers.tpl
w13915984028: The set value of `.Values.storageClass.defaultStorageClass` on managedchart is kept unchanged as user input, `true` or `false`; the `helm update` result on the `storageclass.kubernetes.io/is-default-class` annotation of `harvester-longhorn` SC is dynamic.   In another words, for this field, the priority of managedchart is lower than the existing SCs.

If a user happens to find that this specific setting on managedchart does not propgate to `harvester-longhorn` SC, we are happy to explain the details. It has much less surprise that the managedchart showing it's in error status and the upgrade is blocked.

WebberHuang1118 (approved):
LGTM, thanks for the PR.

w13915984028:
@mergify backport v1.4

Issue #4323, opened 2023-07-27
[ENHANCEMENT] Support untagged VLAN for storage network


starbops:
**Is your enhancement related to a problem? Please describe.**

The storage network configuration only supports tagged VLAN currently. The "VLAN ID" field on the UI is mandatory. Users might want to set up their storage network on an untagged VLAN network. This is not achievable now.

**Describe the solution you'd like**

For such use cases, users should be able to associate the storage network with a cluster network without specifying the VLAN ID.

**Describe alternatives you've considered**

**Additional context**

Original discussion thread: https://rancher-users.slack.com/archives/C01GKHKAG0K/p1690380258466139

We already supported [untagged VM networks](https://docs.harvesterhci.io/v1.1/networking/harvester-network#untagged-network); such capability is also needed for the storage network.

andyrue:
I ran into this issue while testing things out on a Proxmox host. By default Proxmox NICs don't allow passing of tagged traffic, and I was using a dedicated NIC and Switch, so I didn't need VLANs. So if anybody runs into this, it may seem obvious, but make sure your infrastructure is configured to handle whatever VLAN ID you give it, even if you don't need VLANs.

w13915984028:
@andyrue 

This is an interesting scenario, could you provide more context, thanks:

When Equinix supports machine(Harvester Node) to have mutli NICs and all those NICs are in plain network, how are they differentiated? Suppose each NIC is put in a separate subnet, then:

(1) Does Equinix allow machine to set IP itself  or machine should use the pre-allocated/DHCP based IP from equinix?

(2) How does Equinix price the IP?  The `StorageNetwork` needs a (big) subnet to work, this will be an additional cost to user if every IP costs.

The `StorageNetwork` adds a second IP to most of the Longhorn related PODs for dedicated storage usage, the packets go through the provider (e.g. Equinix infrastructure) network, but do not transfer the L3 subnet, and do not communicate with the external.

At the moment, the `StorageNetwork` depends on the VLAN Network for network isolation and IP management. If user does not have a pure L3 network infrastructure (indeed, most do not have), enabling the `StorageNetwork` on untagged network is error-prone.

Supporting the pure L3 network will be combined with the in-plan new feature https://github.com/harvester/harvester/issues/2322.

dfaltum:
[...]

Issue #7222, opened 2024-12-20
[DOC] Best practices for storage network configuration


rrajendran17:
Few questions arise as part of the following issues where we need to document the best practices with storage network configuration
1.https://github.com/harvester/harvester/issues/7212 - storage network using mgmt cluster
2.https://github.com/harvester/harvester/issues/6256 - user not providing sufficient IP range for storage network
3.https://github.com/harvester/harvester/issues/4323#issuecomment-2607809719

**Describe the solution you'd like**
The following are the best practices the user should keep in mind before configuring the storage network configuration.

1.Since new nodes and new disks could be added after the storage network configuration, it is the responsibility of users
to allocate a sufficient IP range for storage networks considering the scalability range of their clusters.

2.The users are allowed to configure storage networks using mgmt cluster, but it is not the best or recommended configuration
   as same physical nics are shared for both storage network and other traffic over mgmt network and communication between LH pods    using storage network is not isolated effectively.

consider using storage network on mgmt cluster only if you could segregate traffic using vlans effectively on same physical link and has some nic constraints in your cluster

3.Using untagged or vlan 1 with mgmt cluster (with no vlan) should be avoided as there is no segregation of network in this LH pod communication

khushboo-rancher:
Closing this as the doc update looks good.

PR #7422, opened 2025-01-21
untagged network support for storage network

rrajendran17:
**Problem:**
https://github.com/harvester/harvester/issues/4323#issuecomment-2568639037

**Solution:**
https://github.com/harvester/harvester/issues/4323#issuecomment-2568639037

**Related Issue:**
https://github.com/harvester/harvester/issues/4323

**Test plan:**
1.Select storage network with untagged vlan option from GUI
2.Fill in the ip address range from untagged vlan network and save
3.Verify the following
https://docs.harvesterhci.io/v1.5/advanced/storagenetwork/#after-applying-harvester-storage-network-setting

Refer following steps for GUI configuration
https://github.com/harvester/harvester-ui-extension/pull/119#issue-2841734801

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -1303,9 +1303,14 @@ func (v *settingValidator) checkVCSpansAllNodes(config *storagenetworkctl.Config
 }
 
 func (v *settingValidator) checkStorageNetworkVlanValid(config *storagenetworkctl.Config) error {
-	if config.Vlan < 1 || config.Vlan > 4094 {
-		return fmt.Errorf("The valid value range for VLAN IDs is 1 to 4094")
+	if config.Vlan < 0 || config.Vlan > 4094 {
+		return fmt.Errorf("The valid value range for VLAN IDs is 0 to 4094")
 	}
+
+	if config.Vlan <= 1 && config.ClusterNetwork == mgmtClusterNetwork {
+		return fmt.Errorf("storage network with vlan id %d not allowed on %s cluster", config.Vlan, config.ClusterNetwork)
+	}
+
 	return nil
 }
 
--- a/pkg/controller/master/storagenetwork/storage_network.go
+++ b/pkg/controller/master/storagenetwork/storage_network.go
@@ -277,9 +277,6 @@ func (h *Handler) createNad(setting *harvesterv1.Setting) (*nadv1.NetworkAttachm
 	bridgeConfig.Bridge = config.ClusterNetwork + BridgeSuffix
 	bridgeConfig.IPAM.Range = config.Range
 
-	if config.Vlan == 0 {
-		config.Vlan = DefaultPVID
-	}
 	bridgeConfig.Vlan = config.Vlan
 
 	if len(config.Exclude) > 0 {


rrajendran17:
There is a case where user might configure untagged with mgmt cluster and mgmt interface has no vlan configured.This applies to vlan with id 1 as well.

Currently there is no easy way getting the vlan of mgmt interface in harvester code for validating this.Is it worth adding a webhook change to prevent users configuring untagged vlan for storage network in mgmt cluster with mgmt interface with no vlan? Or can we add a tooltip or documentation to users, so that users should know the right way to configure storage network so that its used effectively ?

IMO, adding information to the doc is better as all network configurations need not to be validated at webhook.

starbops (approved):
LGTM, thanks.

pkg/webhook/resources/setting/validator.go
```diff
@@ -1303,8 +1303,8 @@ func (v *settingValidator) checkVCSpansAllNodes(config *storagenetworkctl.Config
 }
 
 func (v *settingValidator) checkStorageNetworkVlanValid(config *storagenetworkctl.Config) error {
-	if config.Vlan < 1 || config.Vlan > 4094 {
-		return fmt.Errorf("The valid value range for VLAN IDs is 1 to 4094")
+	if config.Vlan < 0 || config.Vlan > 4094 {
```
w13915984028, line 1306: Based on the cases,

```
b. Using vlan 1 or untagged could be supported with separate cluster network

c.Using vlan 1 or untagged could be supported with mgmt network in different vlan.
```

We could have following assumption:

If the mgmt network is vlan based, the storage network attached to mgmt network has no reason to be un-tagged; if mgmt network is untagged, the storage-network can't be un-tagged.

vlan id 0 is only supported when storage network is backed by a separate cluster-network, which is more like the equinix metal infra. all machine's NICs are put into different backend switch/router, that handles the traffic isolation.

WDYT? thanks.

rrajendran17: @w13915984028 Regarding "If the mgmt network is vlan based, the storage network attached to mgmt network has no reason to be un-tagged"

Can there be a case where ips allocated to the storage network will be in subnet1 and ips allocated to the mgmt network will be in subnet2 and external switch connected will have respective GW ips and packets are forwarded only within its each subnet ?I think physical nic with vlan configured (including untagged ) will be same for storage and mgmt network but there are virtual interfaces created for each vlan holding the gw ip of each subnet allocated from dhcp server.

This is something like the exact opposite usecase where we have mgmt interface untagged and storage network in vlan tag.
Please correct me if I am missing something.

Also FYI, I am also adding some info for best practices for configuring storage network which suggests user not to configure untagged with mgmt network.
https://github.com/harvester/harvester/issues/7222

w13915984028: The case `mgmt in vlan network but storage-network in none-vlan network` is possible in theory, but very rarely practical.

The storage-network aims to have a dedicated, high-bandwidth network for storage usage, the best option is exclusive NIC, better option is sharing with the second network with VM, last option is co-existing on mgmt network. 

The last option is mainly for test/PoC etc, not recommended for production. There is no QoS/bandwidth setting on mgmt network so that shared storage-network can have a [min, max] bandwidth, it is hardly to be better than fallbacking to the none storage-network mode, all LH pods just use the POD network.

The storge-network is only carrying private traffic, it does not send traffic to outside and vice versa, exposing the IPs & traffic to untagged network is not a secure option.

Additional considertaion is, a minor code change can get it works on mgmt network, before we get a specific requirement, we limit the storage-network on mgmt network must with a VLAN, but allow it on none-mgmt network. thanks.

rrajendran17: Thanks @w13915984028 for the details.Considering the practical usage its good to add a check so that we allow storage network with untagged only if the cluster network is non mgmt network as you quoted above.
Considering similar functionality we have with vlan 1 as well, I believe this is applicable to vlan id =1 also.

rrajendran17: updated the PR

w13915984028 (approved):
LGTM, thanks.

pkg/webhook/resources/setting/validator.go
starbops: Are we assuming the native VLAN is `1` here? What if the user has the native VLAN set to something other than `1`, do we still want to prohibit such kind of use cases?

rrajendran17: I believe native vlan applies to external switch configuration and we will not have control over it.Our current Harvester treats both untagged and vlan tag 1 in the same manner and the usecase we are trying to restrict here is using storage network of untagged/vlan 1 with mgmt network which is generally going to be untagged.
Please correct me if I understood your question otherwise.

starbops: Frankly speaking, I just don't get the idea of "not letting users create the storage network with untagged VLAN when the `mgmt` cluster network is chosen." I mean, for example, if the native VLAN of the external switch is configured as `100`, and the cluster nodes' management interface is configured to VLAN `200`, it seems valid for the user to create the storage network with untagged VLAN on the `mgmt` network.

Unless there are problems, we should be happy to provide users with as many options as possible.

Or maybe I misunderstood it, please correct me then.

rrajendran17: @starbops The decision taken is based on considering only the harvester cluster network and practical use cases currently we have with storage networks.
case 1:
storage network  tagged and mgmt network untagged/tagged with same vlan
case 2:
storage network untagged and  mgmt network tagged/untagged

I see both the above cases similar in case of segregation and performance as we are using the same physical link for traffic within Harvester.IMO, the infrastructure outside our cluster(how the traffic is carried in external network) is out of scope for this.
Ideally we should allow case 2 as you said if we are allowing case 1, but with the previous discussions we feel mgmt network to be untagged most of the times, and we could restrict user for storage network with untagged on mgmt cluster considering the common cases we have so far.
But I am open to suggestions and we can definitely revisit considering what is best.

WebberHuang1118 (approved):
LGTM, thanks for the PR.

rrajendran17:
@Mergifyio backport v1.5

pkg/webhook/resources/setting/validator.go
starbops: I get the idea. Thanks for the elaboration.

PR #7437, opened 2025-01-23
chore: add community sprint workflows

jbrockmeyer:
**Problem:**
chore: add community sprint workflows to help organize the process better. 

**Solution:**
Inspired by longhorn's process

**Related Issue:**

**Test plan:**

### DIFF
--- a/.github/workflows/add-to-projects.yml
+++ b/.github/workflows/add-to-projects.yml
@@ -1,0 +1,50 @@
+name: Add-To-Projects
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  community:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Is Harvester Member
+      id: is-harvester-member
+      continue-on-error: true
+      uses: tspascoal/get-user-teams-membership@v3
+      with:
+        username: ${{ github.event.issue.user.login }}
+        organization: harvester
+        GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+
+    - name: Get Issue
+      if: steps.is-harvester-member.outcome == 'success'
+      uses: octokit/request-action@v2.x
+      id: issue
+      with:
+        route: GET /repos/${{ github.repository }}/issues/${{ github.event.issue.number }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+
+    - name: Add Issue to Community Sprint Project
+      id: add-project
+      if: |
+        steps.is-harvester-member.outcome == 'success' &&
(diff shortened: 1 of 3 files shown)


ibrokethecloud (approved):

.github/workflows/add-to-projects.yml
```diff
@@ -0,0 +1,50 @@
...
+        GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+
+    - name: Get Issue
+      if: steps.is-harvester-member.outcome == 'success'
+      uses: octokit/request-action@v2.x
+      id: issue
+      with:
+        route: GET /repos/${{ github.repository }}/issues/${{ github.event.issue.number }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
```
ihcsim, line 26: I haven't tested this but I suspect we don't need this at all. I think `add-project` will auto-retrieve the issue at-hand.

.github/workflows/update-issue-sprint.yml
```diff
@@ -51,4 +51,16 @@ jobs:
...
+        uses: derekbit/move-to-next-iteration@master
+        if: steps.check_sprint_build_required.outcome == 'success'
+        with:
+          owner: harvester
+          number: 10
+          token: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+          iteration-field: Sprint
+          iteration: current
+          new-iteration: next
+          statuses: 'New'
```
ihcsim: Nit: Can we fix this line break issue? Most editors should auto-fix it if you 'save' the file again. Thanks.

.github/workflows/update-projects.yaml
```diff
@@ -0,0 +1,153 @@
...
+        steps.add-project.outputs.itemId != '' 
+        }}
+      run: |
+        teams=${{ steps.is-harvester-member.outputs.teams }}
+        teams=${teams:1:-1}
+        IFS=',' read -r -a team_array <<< "$teams"
+        team_count=${#team_array[@]}
+        echo "team_count: $team_count"
+
+        # If the issue is milestoned by harvester member, update to Resolved
```
ihcsim, line 71: Why does `milestoned` imply `resolved`? I thought that normally means `scheduled`.

.github/workflows/add-to-projects.yml
jbrockmeyer: i think we should leave it since its there from longhorn.

.github/workflows/update-projects.yaml
jbrockmeyer: i'll admit, its a term used from longhorn's process

ihcsim (approved):

Issue #7373, opened 2025-01-14
[ENHANCEMENT] Do not shutdown harvester upgrade-repo VMs in multiple-node cluster upgrade scenario


bk201:
**Is your enhancement related to a problem? Please describe.**

If the upgrade-repo vm lives in the pre-draining node, we'll shut down it.
And the post-drain job will bring it up again.

**Describe the solution you'd like**

This is not necessary and the VM can just be live-migrated out.

**Describe alternatives you've considered**

**Additional context**

w13915984028:
@bk201 @starbops @FrankYang0529 

It looks the issue https://github.com/harvester/harvester/issues/7096 shows that the repo VM can't migrate on some clusters due to CPU architecture differences.

The fix https://github.com/harvester/harvester/pull/7008 && https://github.com/harvester/harvester/pull/7440 indeed have no real relationship with #7096 

Before https://github.com/harvester/harvester/issues/7101 is added, the issue #7096  still has chance to occur on existing clusters.

Do we have a plan to fix https://github.com/harvester/harvester/issues/7096 before #7101 ?  Shutdown upgrade-repo instead of live-migrate it? thanks.

w13915984028:
Or set cpu model to `host-passthrough`  instead of default `host-model` for v14x/v15x?   For already released clusters, it is difficult to change existing code

https://github.com/harvester/harvester/blob/71f907dbdfb5898bcb93b18d0a9dfcd666efb21c/pkg/controller/master/upgrade/upgrade_repo.go#L203

```
Default CPU model is set when vmi doesn't have any cpu model. When vmi has cpu model set, then vmi's cpu model is preferred. When default cpu model is not set and vmi's cpu model is not set too, host-model will be set. Default cpu model can be changed when kubevirt is running. When CPUNodeDiscovery feature gate is enabled Kubevirt creates node selector with default cpu model.

CPU model special cases[¶](https://kubevirt.io/user-guide/compute/virtual_hardware/#cpu-model-special-cases)
As special cases you can set spec.domain.cpu.model equals to: - host-passthrough to passthrough CPU from the node to the VM
```

bk201:
+1 to update the cpu model.

brandboat:
Verify the upgrade repo VM is successfully migrated to other node, test OK, will close the issue.

### Test Environment
- harvester 2 node cluster, v1.4.2, 12C/16G/250GB for each node.

### Test steps
[...]

Issue #7101, opened 2024-12-05
[ENHANCEMENT] Deprecating Harvester upgrade-repo virtual machine


starbops:
**Is your enhancement related to a problem? Please describe.**

Harvester is released as ISO images. Whether installing a new Harvester cluster or upgrading an existing one, the ISO image is required. This is the main reason we designed the upgrade workflow surrounding an upgrade-repo virtual machine. The upgrade-repo virtual machine is booted from the ISO image, providing a central hub for various components to grab the data they need.

Enough introduction. Despite the fact that virtual machines can be directly booted from ISO images, the upgrade repo as a virtual machine imposes several limitations and inconveniences. This is because of the characteristics of how Harvester upgrade works.

1. Running a virtual machine itself is considered a heavy task, which involves BackingImage creation, handling the Volume's lifecycle, and maintaining the replicas; all of these take time
2. It could run into issues when live-migrating the upgrade-repo virtual machine. For instance, #7096 
3. Potential heavy traffic happens on the pod network during the image preloading phase (under the hood, it is the management network that is being affected)

In the end, we just need a place for the upgrade controller and upgrade-relevant scripts to grab what they want. We could have done it better. I think it's a good time to improve this.

**Describe the solution you'd like**

Deprecating the virtual machine way and use the pod instead. Pods cannot utilize ISO images inherently, so we have to take care of the ISO image ourselves, before spinning up the upgrade-repo pods. For files to be ready locally, we need a way to distribute the ISO image to every node. Therefore, we can mount the ISO image locally and import the container images and other image archives directly. Two methods that I can think of exist to have the ISO image ready on every node.:

- Leveraging Longhorn BackingImage to distribute the ISO file to every node by setting `minNumberOfCopies` to the cluster node count.
- Spread the pods to download the ISO image from the BackingImage using system-upgrade-controller's Plan

After that, spinning up upgrade-repo pods with a `hostPath` volume pointing to the mount point of the ISO image on the node.

**Describe alternatives you've considered**

Pre-processing the ISO image to form a Longhorn Volume with the necessary data by the upgrade controller. The upgrade-repo pod attaches to the Longhorn Volume. There will be only one upgrade-repo pod.

**Additional context**

[...]

PR #7440, opened 2025-01-23
fix(upgrade): exclude repo vm from live-migrate detection

starbops:

**Problem:**

Harvester upgrade repo VM will be shut down during the pre-drain phase if it happens to run on the node under upgrade.

**Solution:**

Make live-migrate detector no longer takes Harvester upgrade repo VMs into account.

**Related Issue:**

#7373 

**Test plan:**

1. Make an ISO image containing the fix (for QAs, please use the master-head ISO image)
2. Spin up a Harvester cluster (at least consists of two nodes) in v1.4.0
3. Kick-off the upgrade using the ISO image mentioned in step one
4. Monitor the upgrade repo VM. The upgrade procedure should not shut it down but rather live-migrate it.

### DIFF
--- a/pkg/upgradehelper/vmlivemigratedetector/detector.go
+++ b/pkg/upgradehelper/vmlivemigratedetector/detector.go
@@ -9,6 +9,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	kubevirtv1 "kubevirt.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


w13915984028:
There could be risks to add this enhancement, please check.

(1) VM migrations are processed in batch mode, e.g. 2 in parallel, when a node has many VMs, the repo VM will extend the overall time of finishing the migration.

(2) VM migration can be aborted from UI (https://github.com/harvester/harvester/issues/7140#issuecomment-2609293463), in upgrade when a VM takes too long time or due to other reason user can abort it to fix;  but repo VM is not shown on UI (IIRC), in error case, user can only use `kubectl` to delete the related vmim object, it may also not benefit from the enhancement https://github.com/harvester/harvester/pull/7146

(3) The repo VM is a read-only VM, it has no meaningful data to be live migrated, the advantage of live-migration may not over-beat the stop&start mode, but it depends on the real test.

For stop & start mode, a possible enhancement is to schedule the repo VM to an already upgraded node actively;  in live-migration mode, it may be migrated several times in unlucky case. thanks.

starbops:
@w13915984028 Thanks for the feedback. Live-migrating the upgrade repo VM has been a long tradition in the Harvester Upgrade procedure. It was not after #7008 was introduced that it became the stop-and-stop mode. This PR is mainly focused on restoring the original behavior. That being said, I agree with your viewpoints. Hopefully, they can be addressed with [this issue](https://github.com/harvester/harvester/issues/7101) by dropping the upgrade repo VM. If we still find something missing, we can track it in other new issues.

ihcsim (approved):

WebberHuang1118 (approved):
LGTM, thanks for the PR.

FrankYang0529 (approved):
LGTM. The upgrade-repo VM is lively migrated to another node.

Screenshot 2025-02-05 at 10 08 06 AM

Issue #7196, opened 2024-12-18
[ENHANCEMENT] Add Validating Admission Webhook/Policy To Reject RKE2 CIDRs Changes In The Config Map


ihcsim:
**Is your enhancement related to a problem? Please describe.**

PR #7156 updated the `harvester-helpers` config map to support overriding of the RKE2 CIDRs during node promotion, as part of the overall feature described at #4254. Despite the new comments in the chart's `values.yaml`, there is nothing prohibiting the user from manipulating Helm to override these settings, which must not be changed post-installation.

**Describe the solution you'd like**

Use a validating admission webhook or [validating admission policy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) to reject any post-installation changes to the RKE2 CIDRs in the config map.

**Describe alternatives you've considered**

**Additional context**

w13915984028:
Expect your investigation on [validating admission policy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/), it will be a handy tool.

albinsun:
Hi @starbops ,
Please help to verify this issue, thank you.

Vicente-Cheng:
Hi @martindekov, 
Since @starbops is working on the upgrade issue, could you help verify this issue? Thanks!

martindekov:
Yep @Vicente-Cheng on it

martindekov:
So I did the following:
* Basic test to validate webhook rejects changes to all values under `spec.values.promote` in the `managedchart/harvester` in namespace `fleet-local` as described ☑️  
* Install a `v1.5.1` cluster with default CIDR, and upgrade to `v1.6.0-rc5`, upgrade is successful and CIRD is kept ☑️  
* Install a `v1.5.1` cluster with customized CIDR, and upgrade to  `v1.6.0-rc5`, upgrade is successful and customized CIRD is kept ☑️   
* Post upgrade on v1.6.0-rc5 webhook still rejects all values under `spec.values.promote` in the `managedchart/harvester` ☑️   

I believe if those are sufficient and exhaust the testing, we can close the issue ?

cc @w13915984028 @Vicente-Cheng

innobead:
Moving to `testing`, since @martindekov is on it. BTW, the test result looks good to me. cc @ihcsim

ihcsim:
Closing this issue per QA verification.

Issue #4254, opened 2023-07-18
[FEATURE] Modify the default cluster-cidr of harvester


chloe6888:
The default cluster-cidr network segment of the harvester cluster conflicts with our production environment. Is it possible to modify the method in the following way?
```
# git clone https://github.com/harvester/harvester-installer.git
# cd harvester-installer
# git checkout v1.1.2

# vim ./pkg/config/templates/rke2-90-harvester-server.yaml
......
cluster-cidr: 30.0.0.0/16
service-cidr: 30.1.0.0/16
cluster-dns: 30.1.0.10
......

# make

# ls -al ./dist/artifacts/
total 5927012
drwxr-xr-x. 2 root root        259 Jun  7 07:15 .
drwxr-xr-x. 3 root root         23 Jun  7 06:55 ..
-rw-r--r--. 1 root root 5345443840 Jun  7 07:15 harvester-cf6d3b7-dirty-amd64.iso
-rw-r--r--. 1 root root        675 Jun  7 07:15 harvester-cf6d3b7-dirty-amd64.sha512
-rw-r--r--. 1 root root   90024588 Jun  7 07:10 harvester-cf6d3b7-dirty-initrd-amd64
-rw-r--r--. 1 root root  622329856 Jun  7 07:14 harvester-cf6d3b7-dirty-rootfs-amd64.squashfs
-rw-r--r--. 1 root root   11448480 Apr 13 08:04 harvester-cf6d3b7-dirty-vmlinuz-amd64
-rw-r--r--. 1 root root       1897 Jun  7 07:15 image-lists.tar.gz

```
Could the official provide a more convenient and simple way to modify the harvester cluster-cidr configuration?

masteryyh:
cc @guangbochen

chloe6888:
When deploying the Harvester cluster using a repackaged image, we encountered the following issue. Is it possible that there are some configurations that need to be adjusted?

I have prepared three virtual machines, hk-harvester-master-0-{1~3}, for testing cluster deployment using repackaged iso. After the first node was deployed, the configuration of CIDR in the file /etc/rancher/rke2/config.yaml.d/90-harvester-server.yaml matched our expectations. However, upon adding the third node to the Harvester cluster, the status of the second node hk-harvester-master-0-2 that had joined the cluster became notready. 
checking the logs of its RKE2 server, we have found the following error:
```
[...]

PR #7445, opened 2025-01-23
feat: add validating admission policy to prohibit changes to RKE2 CIDR within managed chart

ihcsim:

**Problem:**
Following the completion of #4254, user can now override the RKE2's default pod and service CIDRs. There is nothing prohibiting the user from manipulating Helm to override these settings within the `harvester` managed chart, to alter the values found in the `harvester-system/harvester-helpers` config map. These settings must not be changed post-installation.

Issue: #7196

**Solution:**
Introduce a validating admission policy to prevent the modification of the RKE2 CIDRs within the `harvester` managed chart. 

Per discussion at https://github.com/harvester/harvester/pull/7445#issuecomment-2638181603, the validating admission policy doesn't block attempts to alter the CIDRs that are propagated to the `harvester-system/harvester-helpers`  config map. Since these CIDRs are embedded in a big blob of string within the config map, any unforeseen string change patterns can easily trigger a false positive policy failures. For now, we continue to rely on the existing mechanism of letting the managed chart and bundle error out and block any upgrade, if the config map is changed. Further guardrails to protect the config map will be considered in a future PR.

**Related Issue:**

**Test plan:**

## Base Case - Current behaviour

Without the changes in this PR, user can use `k -n fleet-local edit managedchart harvester` to change the CIDRs defined within the `spec.values.promote` block. This causes the undesirable effect of changing the CIDRs defined in the `harvester-system/harvester-helpers` config map. E.g.,

```sh
# before changing the managed chart
✗ k -n harvester-system get cm harvester-helpers -ojsonpath='{.data.promote\.sh}' | grep -A2 -i "cluster-cidr"
cluster-cidr: 172.16.0.0/16
service-cidr: 172.20.0.0/16
cluster-dns: 172.20.0.10

# after changing the managed chart
✗ k -n harvester-system get cm harvester-helpers -ojsonpath='{.data.promote\.sh}' | grep -A2 -i "cluster-cidr"
cluster-cidr: 172.19.0.0/16
service-cidr: 172.31.0.0/16
cluster-dns: 172.31.0.10
```

## Test Case 1 - Prohibit changes to the promote CIDRs defined in the managed chart

Use `helm template` to render the updated Harvester chart which includes the new `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` resources:

```sh
helm -n harvester-system template harvester ./deploy/charts/harvester
```

Deploy these new Helm-rendered YAML to an existing Harvester cluster:

```sh
$ cat <<EOF | k apply -f -
# Source: harvester/templates/validating-admission-policy.yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: harvester-immutable-promote-cidr
  labels:
    app.kubernetes.io/managed-by: "Helm"
    helm.sh/chart: harvester-0.0.0-dev
    app.kubernetes.io/version: "0.1.x"
    helm.sh/release: harvester
    app.kubernetes.io/part-of: harvester
spec:
  failurePolicy: Fail
  matchConstraints:
    namespaceSelector:
      matchExpressions:
      - key: kubernetes.io/metadata.name
        operator: In
        values: ["fleet-local"]
    resourceRules:
    - apiGroups:   ["management.cattle.io"]
      apiVersions: ["v3"]
      operations:  ["UPDATE"]
      resources:   ["managedcharts"]
      resourceNames: ["harvester"]
      scope: Namespaced
  matchConditions:
  - name: "skip-if-upgrading-from-1.4-or-older"
    expression: "'promote' in oldObject.spec.values"
  variables:
  - name: newPodCIDR
    expression: "'clusterPodCIDR' in object.spec.values.promote ? object.spec.values.promote.clusterPodCIDR : ''"
  - name: oldPodCIDR
    expression: "'clusterPodCIDR' in oldObject.spec.values.promote ? oldObject.spec.values.promote.clusterPodCIDR : ''"
  - name: newServiceCIDR
    expression: "'clusterServiceCIDR' in object.spec.values.promote ? object.spec.values.promote.clusterServiceCIDR : ''"
  - name: oldServiceCIDR
    expression: "'clusterServiceCIDR' in oldObject.spec.values.promote ? oldObject.spec.values.promote.clusterServiceCIDR : ''"
  - name: newClusterDNS
    expression: "'clusterDNS' in object.spec.values.promote ? object.spec.values.promote.clusterDNS : ''"
  - name: oldClusterDNS
    expression: "'clusterDNS' in oldObject.spec.values.promote ? oldObject.spec.values.promote.clusterDNS : ''"
  validations:
  - expression: |-
      (variables.oldPodCIDR == "" || variables.newPodCIDR == variables.oldPodCIDR) && 
      (variables.oldServiceCIDR == "" || variables.newServiceCIDR == variables.oldServiceCIDR )&& 
      (variables.oldClusterDNS == "" || variables.newClusterDNS == variables.oldClusterDNS)
    messageExpression: |-
      'the promote CIDRs configuration is immutable. ' +
      'clusterCIDR: ' + variables.oldPodCIDR + ' => ' + variables.newPodCIDR + ', ' +
      'serviceCIDR:' + variables.oldServiceCIDR + ' => ' + variables.newServiceCIDR + ', ' +
      'clusterDNS: ' + variables.oldClusterDNS + ' => ' + variables.newClusterDNS
    reason: Invalid
---
# Source: harvester/templates/validating-admission-policy.yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: harvester-immutable-promote-cidr
  labels:
    app.kubernetes.io/managed-by: "Helm"
    helm.sh/chart: harvester-0.0.0-dev
    app.kubernetes.io/version: "0.1.x"
    helm.sh/release: harvester
    app.kubernetes.io/part-of: harvester
spec:
  policyName: harvester-immutable-promote-cidr
  validationActions: [Deny]
EOF
validatingadmissionpolicy.admissionregistration.k8s.io/harvester-immutable-promote-cidr created
validatingadmissionpolicybinding.admissionregistration.k8s.io/harvester-immutable-promote-cidr created
```

Use `k -n fleet-local edit managedchart harvester` to perform the following changes to the CIDRs within the `spec.values.promote` block:

* change the value of the `cluster-cidr` setting
* change the value of the `service-cidr` setting
* change the value of the `cluster-dns` setting
* delete any or all of these CIDRs settings

K8s will reject the changes with the following error:

```sh
#
# managedcharts.management.cattle.io "harvester" was not valid:
# * : ValidatingAdmissionPolicy 'harvester-immutable-promote-cidr' with binding 'harvester-immutable-promote-cidr' denied request: the promote CIDRs configuration is immutable. clusterCIDR: 172.16.0.0/16 => 172.16.0.0/16, serviceCIDR:172.22.0.0/16 => 172.22.0.0/16, clusterDNS: 172.22.0.10 => 172.20.0.10
#
```

## Test Case 2 - Skip policy validation when upgrading from 1.4 or older

Upgrading from Harvester 1.4 or older should not be blocked by the new admission policy. The policy object recognizes that older managed chart in 1.4 or older does not have the promote CIDRs. Hence, it would skip the rules validation.

### DIFF
--- a/deploy/charts/harvester/templates/validating-admission-policy.yaml
+++ b/deploy/charts/harvester/templates/validating-admission-policy.yaml
@@ -1,0 +1,58 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: harvester-immutable-promote-cidr
+  labels:
+{{ include "harvester.labels" . | indent 4 }}
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    namespaceSelector:
+      matchExpressions:
+      - key: kubernetes.io/metadata.name
+        operator: In
+        values: ["fleet-local"]
+    resourceRules:
+    - apiGroups:   ["management.cattle.io"]
+      apiVersions: ["v3"]
+      operations:  ["UPDATE"]
+      resources:   ["managedcharts"]
+      resourceNames: ["harvester"]
+      scope: Namespaced
+  matchConditions:
+  - name: "skip-if-upgrading-from-1.4-or-older"
+    expression: "'promote' in oldObject.spec.values"
+  variables:
+  - name: newPodCIDR
+    expression: "'clusterPodCIDR' in object.spec.values.promote ? object.spec.values.promote.clusterPodCIDR : ''"
+  - name: oldPodCIDR
+    expression: "'clusterPodCIDR' in oldObject.spec.values.promote ? oldObject.spec.values.promote.clusterPodCIDR : ''"
+  - name: newServiceCIDR
+    expression: "'clusterServiceCIDR' in object.spec.values.promote ? object.spec.values.promote.clusterServiceCIDR : ''"
+  - name: oldServiceCIDR
+    expression: "'clusterServiceCIDR' in oldObject.spec.values.promote ? oldObject.spec.values.promote.clusterServiceCIDR : ''"
+  - name: newClusterDNS
+    expression: "'clusterDNS' in object.spec.values.promote ? object.spec.values.promote.clusterDNS : ''"
+  - name: oldClusterDNS
+    expression: "'clusterDNS' in oldObject.spec.values.promote ? oldObject.spec.values.promote.clusterDNS : ''"
+  validations:
+  - expression: |-
+      (variables.oldPodCIDR == "" || variables.newPodCIDR == variables.oldPodCIDR) &&
+      (variables.oldServiceCIDR == "" || variables.newServiceCIDR == variables.oldServiceCIDR )&&
+      (variables.oldClusterDNS == "" || variables.newClusterDNS == variables.oldClusterDNS)
+    messageExpression: |-
+      'the promote CIDRs configuration is immutable. ' +
+      'clusterCIDR: ' + variables.oldPodCIDR + ' => ' + variables.newPodCIDR + ', ' +
+      'serviceCIDR:' + variables.oldServiceCIDR + ' => ' + variables.newServiceCIDR + ', ' +
+      'clusterDNS: ' + variables.oldClusterDNS + ' => ' + variables.newClusterDNS
+    reason: Invalid
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: harvester-immutable-promote-cidr
+  labels:
+{{ include "harvester.labels" . | indent 4 }}
+spec:
+  policyName: harvester-immutable-promote-cidr
+  validationActions: [Deny]


deploy/charts/harvester/templates/validating-admission-policy.yaml
```diff
@@ -0,0 +1,35 @@
...
+{{ include "harvester.labels" . | indent 4 }}
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    namespaceSelector:
+      matchExpressions:
+      - key: kubernetes.io/metadata.name
+        operator: In
+        values: ["harvester-system"]
+    resourceRules:
```
w13915984028, line 15: A question:

In Harvester upgrade path  (e.g. v1.4.0->v1.4.1), the configmap may be updated via :

managedchart harvester upgrade ->fleet-agent call ` helm upgrade ` (via source code)

Is this allowed? thanks.

ihcsim: Good point - that won't be allowed because the current policy can't distinguish between a user-initiated change vs. an upgrade-initiated change. Either the rules have to be updated to be more granular to protect these variables (instead of the whole script), or the config map has to be temporary labelled with an upgrade label to bypass the policy.

w13915984028: Block the upgrade will be a blocker, Harvester is 0-downtime upgraded😀.

ihcsim: I updated the policy to parse the string content. Not a big fan tbh, but I don't know if it is any worse than doing it in Go. LMKWYT.

ibrokethecloud:
CI is broken likely because we are using kind v0.14.0 and we are spinning k8s v1.24

```
(⎈|kind-harvester:default)@builder::➜  ~ k get nodes -o wide
NAME                      STATUS     ROLES           AGE   VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE       KERNEL-VERSION       CONTAINER-RUNTIME
harvester-control-plane   NotReady   control-plane   22s   v1.24.0   172.18.0.3    <none>        Ubuntu 21.10   5.15.0-124-generic   containerd://1.6.4
harvester-worker          NotReady   <none>          0s    v1.24.0   172.18.0.4    <none>        Ubuntu 21.10   5.15.0-124-generic   containerd://1.6.4
harvester-worker3         NotReady   <none>          1s    v1.24.0   172.18.0.5    <none>        Ubuntu 21.10   5.15.0-124-generic   containerd://1.6.4
```

ihcsim:
Yeah, `kind` 01.4 is at least 2 years old and K8s 1.24 is EOL. I was hoping to get some feedback on this PR itself first, before checking in to see if we want to bump `kind` to a newer K8s version that matches Rancher v2.9+. If we do bump `kind`, it will be a different PR, as there is a bit of `vendor` changes.

deploy/charts/harvester/templates/validating-admission-policy.yaml
w13915984028: There will be a hack solution to jump the check:

```
variables.oldData.indexOf('service-cidr:') > -1 ? 
```

```
    cat > /host/etc/rancher/rke2/config.yaml.d/90-harvester-server.yaml <<EOF
    cni: multus,canal
    cluster-cidr: {{ .Values.promote.clusterPodCIDR }}
    service-cidr: {{ .Values.promote.clusterServiceCIDR }}
    cluster-dns: {{ .Values.promote.clusterDNS }}
    tls-san:
      - $VIP
    audit-policy-file: /etc/rancher/rke2/config.yaml.d/92-harvester-kube-audit-policy.yaml
    EOF
```


If user edits the configmap to add a fake  originl `cluster-cidr: {{ .Values.promote.clusterPodCIDR }}` before/after (depends on the `indexOf`) the `cat` line, and edit the real `cidr` line..

The plain `indexOf` has no context about if the cidr is just under the `90-harvester-server.yaml` section.

ihcsim: Yeah, it's not hack proof. If a user tries hard enough, they can trigger all sort of workarounds to skip it. My preference is still to keep the entire `promote.sh` immutable, instead of trying to protect substrings of a big blob of string. I am not sure how `helm upgrade` will work with the `ManagedChart`. If it can perform a `DELETE` then `CREATE`, rather than an `UPDATE`, of the config map, that will be great.

Also, FWIW, user shouldn't be given direct RBAC permissions to mess with anything within the `harvester-system` namespace. I am not sure how/if we are enforcing that today.

w13915984028:
Additional thoughts about this PR:

The CIDR is initially added from below via installer,
https://github.com/harvester/harvester-installer/blob/6fa1df88616727e74b789a22440120787dc86732/pkg/config/templates/rancherd-10-harvester.yaml#L217

1.Do we consider to add webhook/policy upon `managedchart harvester` to deny further changes on those fields? 

2.Do we consider to separate the CIDR info to another configmap like VIP
https://github.com/harvester/harvester/blob/49f54b3f0e5c82578efb6e0634e1ad7bdbecd7cc/deploy/charts/harvester/templates/configmap.yaml#L2

The additonal configmap will also benifit other controllers, when they want to get this information. E.g., network related controller to check a user input network CIDR is in the range of clusterPoDCIDR.

3.If it is separated, the webhook can simply apply on the new configmap,disallow any changes; and the promote.sh uses `kubectl` to get values from the new configmap and inject to server.yaml, just like VIP

```
    VIP=$($KUBECTL get configmap vip -n harvester-system -o=jsonpath='{.data.ip}')
    cat > /host/etc/rancher/rke2/config.yaml.d/90-harvester-server.yaml <<EOF
    cni: multus,canal
    cluster-cidr: {{ .Values.promote.clusterPodCIDR }}
    service-cidr: {{ .Values.promote.clusterServiceCIDR }}
    cluster-dns: {{ .Values.promote.clusterDNS }}
    tls-san:
      - $VIP
    audit-policy-file: /etc/rancher/rke2/config.yaml.d/92-harvester-kube-audit-policy.yaml
    EOF
```

thanks.

ihcsim:
Good idea. For this PR, I am thinking of tackling just the managed chart, and making the `promote` CIDR variables immutable. I think it makes sense because these values should not ever change post-installation, unlike the entire `promote.sh` script. If someone mod that `promote.sh` directly, fleet will generate an error in the managed chart and bundle. Not 100% tamper safe, but it should be sufficient.

Moving them to a different config map sounds like a good idea. I am not sure what would be the scope of the code change in the promotion controller code at this point. Can we do it in a separate PR? LMKWYT.

w13915984028 (approved):
LGTM, thanks.

Use another issue&PR to enhance is good.

ibrokethecloud (approved):
lgtm

Issue #7474, opened 2025-02-03
[TASK] typo in `nodeStateWatingReboot` constant


tserong:
## What's the task? Please describe

The constant `nodeStateWatingReboot` in pkg/controller/master/upgrade/job_controller.go is missing the letter `i`. It should be `nodeStateWaitingReboot`

tserong:
PR is merged, no testing required.

PR #7475, opened 2025-02-03
fix: s/nodeStateWatingReboot/nodeStateWaitingReboot/g

tserong:
**Problem:**
The constant `nodeStateWatingReboot` in pkg/controller/master/upgrade/job_controller.go is missing the letter `i`. It should be `nodeStateWaitingReboot`.

**Solution:**
This PR

**Related Issue:**
https://github.com/harvester/harvester/issues/7474

**Test plan:**
N/A


tserong:
The CodeFactor error should be ignored in this instance (nothing I did here made that method any more complex than it already was).

starbops (approved):
LGTM, thanks.

Vicente-Cheng (approved):

Issue #7320, opened 2025-01-08
[ENHANCEMENT] Bump Rancher/RKE2 for v1.4.2


bk201:
**Is your enhancement related to a problem? Please describe.**

Bump RKE2 to v1.31.3/4+rke2r1
Bump Rancher to v2.9.x/v2.10.x (TBD)

**Describe the solution you'd like**

Bump RKE2 to v1.31.3/4+rke2r1
Bump Rancher to v2.9.x/v2.10.x  (TBD)

**Describe alternatives you've considered**

**Additional context**

lanfon72:
Verified this change has been applied.

![Image](https://github.com/user-attachments/assets/f93969f0-1f8c-4a93-a07b-0453a46551b6)

Test Information
----
* Environment: **baremetal DL360G9 3 nodes**
* Harvester Version: **v1.4.2-rc1** (and upgrade from **v1.4.1**)
* **ui-source** Option: **Auto**

### Verify Steps
- Install Harvester with any nodes
- Login to console, execute  `sudo ranchered info`
    - Or `kubectl get nodes` for RKE2 version, `kubectl get settings.management.cattle.io server-version` for Rancher version

PR #7479, opened 2025-02-03
feat: change to use helm get values -a to get image

FrankYang0529:

**Problem:**
After bumping Rancher to v2.10.1, we can't get helm values from App objects.

**Solution:**
Change to use `helm get values -a` directly, so we don't need to rely on Rancher code.

**Related Issue:**
https://github.com/harvester/harvester/issues/7320

**Test plan:**
1. Build an ISO with https://github.com/harvester/harvester-installer/pull/926.
2. Create a 3-node airgapped cluster without error.
3. Check all nodes are in control plane.
4. Create support bundle without error.


innobead:
> After bumping Rancher to v2.10.1, we can't get helm values from App objects.

I am curious about the reason behind this breaking change at Rancher. Do we have any information about it?

pkg/controller/master/setting/support_bundle_image.go
```diff
@@ -4,94 +4,46 @@ import (
...
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/settings"
+	"github.com/harvester/harvester/pkg/util"
+	utilHelm "github.com/harvester/harvester/pkg/util/helm"
 )
 
 const (
 	SupportBundleRepository = "support-bundle-kit"
```
innobead: Do we still need this const? as the old code has been removed.

innobead (approved):
LGTM

pkg/controller/master/setting/support_bundle_image.go
FrankYang0529: Removed it. Thanks.

FrankYang0529:
> I am curious about the reason behind this breaking change at Rancher. Do we have any information about it?

This is observed result. I can't find source code about it.

bk201 (approved):

ibrokethecloud (approved):

WebberHuang1118 (approved):
LGTM, thanks the PR.

Issue #1199, opened 2021-09-10
[FEATURE] Consume external/third-party storage within Harvester


jurrien-suse:
**Is your feature request related to a problem? Please describe.**
I got this question from a customer: I have a solution which I ship to my customers which is a blade chassis with blades and a storage appliance in it. It provides iSCSI storage to the blades. I would like to consume that iSCSI storage in Harvester.

**Describe the solution you'd like**
I would like to have the possibility, next to Longhorn providing storage, to have the option to add iSCSI and NFS network attached storage to Harvester.

**Describe alternatives you've considered**
Well technically it is already possible if you enable the Rancher UI and mount the iSCSI on the nodes and add the drives in Longhorn but it's a bit of a double situation when the storage appliance is also doing some kind of RAID config with the disks and you have to set the replication in Longhorn to 1. This is not really the way you want to go I guess having Longhorn in between the iSCSI/NFS storage and Harvester because this will create an extra layer and could not benefit the performance of the storage.

**Additional context**
So it would be nice to consume iSCSI and NFS storage directly without the use of Longhorn.
(This is a real customer question and with this feature the customer could choose for Harvester)

Mwogi:
We have an exact use case in our organization as well. We have iSCSI storage and we would like to easily add it to Harvester. Learning here that you can add it through Ranger? Any links to this?

mddeff:
Jumping on this bandwagon. I support multiple environments each with multiple 60-100T SANs, all serving up NFS for either VMWare or KVM. Would love for Harvester to support NFS/iSCSI for VM Storage.  

Could it be as simple as supporting the [nfs-subdir-external-provisioner](https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner)? At least, for NFS...

olljanat:
Btw. Do you guys mean case where those Blade servers/etc would still have local disks or real diskless setup where Harvester it selves would be installed to external storage?

I can see that documentation already describe how to boot Harvester installer from PXE but it still assumes that installation would be done to local disk.

But it should be possible load even rootfs from iSCSI https://github.com/dracutdevs/dracut/blob/c868feb1b1c118781050ecdff3c14a1a8f008c92/man/dracut.cmdline.7.asc#iscsi or from NFS https://github.com/dracutdevs/dracut/blob/c868feb1b1c118781050ecdff3c14a1a8f008c92/man/dracut.cmdline.7.asc#nfs which would works even on fully diskless nodes.

[...]

PR #7487, opened 2025-02-03
HEP: add Third-party storage support

Vicente-Cheng:

**Problem:**
Add HEP for Third-party storage support

**Solution:**
Add HEP for Third-party storage support

**Related Issue:**
https://github.com/harvester/harvester/issues/1199

**Test plan:**
No need


enhancements/20250203-third-party-storage-support.md
```diff
@@ -0,0 +1,98 @@
...
+### Test plan
+
+- Ensure the CDI is installed default in Harvester
+- Ensure the fleet did not complain anythin with the new CDI. (Both fresh install and upgrade)
+- Ensure the VM image can be created through download/upload with different third party storage solutions (e.g. LVM, Rook, etc.)
+- Ensure the VM can be create with the above VM image (means the root disk is created through the third party storage solution)
+
+### Upgrade strategy
+
+This enhancement will add extra fields to the current VirtualMachineImage CRD. So, the upgrade needs to handle these new fields.
```
ihcsim, line 328: Can we add a sample of what the new image YAML will look like?

enhancements/20250203-third-party-storage-support.md
```diff
@@ -0,0 +1,98 @@
+# Third Party Storage Support
+
+Third-party storage support is a feature that allows users to use third-party storage solutions in Harvester for root disks and data disks.
+
+## Summary
+
+This enhancement aims to provide a generic way to leverage third-party storage solutions which already have CSI driver. By introducing Containerized-Data Importer (CDI), Harvester can support third-party storage solutions with the current Virtual Machine Image path for the root and data disks.
```
ihcsim, line 7: AIUI, CDI is just an [enhanced volume populator](https://github.com/kubernetes-csi/lib-volume-populator). Theoretically, there is nothing stopping a user from creating a pre-populated PVC using whatever image data population mechanism their CSI supports, right? In a future enhancement, is it possible to support something like  `sourceType: predefined` in the image, to allow user to "bring their own root disk"?

enhancements/20250203-third-party-storage-support.md
```diff
@@ -0,0 +1,98 @@
...
+
+To introduce Containerized-Data Importer (CDI) to Harvester, we need to add the charts and the CRDs to Harvester. The CDI will be responsible for the VM Image and the root disk creation with the third-party storage solution. 
+
+#### VM Image Creation through CDI
+
+The VM Image creation design could be like the following:
+![vm-image-design](./20250203-third-party-storage-support/vm-image-design.png)
+
+Because the target PVC (golden image) size should be defined before DataVolume creation, we need to prefetch the size and virtual size.
+We have different mechanisms for different source types (download/upload) for size/virtual size prefetching, and progress updates.
```
ihcsim, line 84: Will download be done by CDI too? What are the supported download targets (ftp, http, oci etc.)?

ihcsim: Will we be supporting both [sync](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/upload.md#synchronous) and [async](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/upload.md#asynchronous) upload?

enhancements/20250203-third-party-storage-support.md
```diff
@@ -0,0 +1,98 @@
...
+For the golden image, we need to add webhook to protect the golden image from being deleted. The golden image should be protected until the VM Image is deleted.
+
+Also, we should not display the golden image on the volume page in the UI.
+
+#### Root Disk Creation through CDI
+
+The VM root disk creation design could be like the following:
+![root-disk-design](./20250203-third-party-storage-support/vm-root-disk-design.png)
+
+The root disk creation is similar to the VM Image creation. Now, with the golden image, we use the golden image as the source (type PVC), and the storage (target) is the root disk configuration. The root disk will be created with the CDI efficient cloning.
```
ihcsim, line 95: Looks like the efficient cloning only works if the underlying CSI driver supports snapshotting and/or volume cloning, right? Otherwise, CDI will fall back to what upstream calls host-assisted cloning, which is slower. If true, we will need to document this.

WebberHuang1118 (approved):
LGTM, thanks.

enhancements/20250203-third-party-storage-support.md
Vicente-Cheng: Currently, user can convert their root disk to qcow2 format and upload it as our VM image.
That might be the same purpose. WDYT?

We could try to allow them to directly import their root disk. Actually, CDI already has that capability (REF: https://github.com/kubevirt/containerized-data-importer/blob/main/doc/supported_operations.md)
we need to check how to integrate and the UI flow.

enhancements/20250203-third-party-storage-support.md
Vicente-Cheng: We now only support the download from http and sync upload.

CDI has better capability, but we are trying to align it with our previous version.
Maybe we could try to improve that in a later version.

enhancements/20250203-third-party-storage-support.md
Vicente-Cheng: update the storage profile part.

enhancements/20250203-third-party-storage-support.md
```diff
@@ -0,0 +1,225 @@
...
+  - Ensure the VM can generate the VM template with various disk configurations. (Like a mix of the Longhorn v1 and LVM)
+  - Ensure the above VM template can be used to create the VM.
+
+- VM snapshot
+  - Ensure the VM can take the snapshot with various disk configurations. (Like a mix of the Longhorn v1 and LVM)
+  - Ensure the above VM snapshot can be restored to the existing VM or new VM.
+
+- VM Backup
+  - We did not support the VM backup with any third-party storage solution. So, ensure this action item does not appear in the UI.
+
```
bk201, line 318: Some suggestions regarding the test plan
- SC creation.
- When an Image is created/deleted, corresponding volumes in 3rd party storage should be created/deleted. 
- VM live migration should work for RWX volumes.
- Harvester upgrade should work when there are VMs booted with 3rd party storage.

Vicente-Cheng: updated, thanks for the suggestion

ihcsim (approved):

enhancements/20250203-third-party-storage-support.md
```diff
@@ -0,0 +1,333 @@
...
+    ```shell
+    // find the golden image name
+    $ imageName=$(kubectl get vmimage -n <namespace> |grep <image display name on UI> |awk '{print $1}')
+    // check the golden image
+    $ kubectl get datavolume -n <namespace> |grep ${imageName}
+    $ kubectl get pvc |grep ${imageName}
+    ```
+    - The above command should not return any result.
+
+- VM Rook Disk
```
ihcsim:
```suggestion
- VM Root Disk
```

Vicente-Cheng: nice catch!
updated

bk201 (approved):
LGTM, thanks!

enhancements/20250203-third-party-storage-support.md
```diff
@@ -0,0 +1,333 @@
...
+- VM Template Generator
+  - Ensure the VM can generate the VM template with various disk configurations. (Like a mix of the Longhorn v1 and LVM)
+    - GoTo Virtual Machines Page
+    - Click the VM
+    - Click the `Generate Template` button and ensure the `With Data` checkbox is checked
+    - Create the VM with the above VM template
+    - Create and wait for the VM to be ready
+
+- VM Live Migration
+  - Ensure the VM can be live migrated with various disk configurations. (Like a mix of the Longhorn v1 and LVM)
```
innobead: Does LVM support live migration? I thought its specific per node.

innobead (approved):
In general, LGTM. Just one question.

enhancements/20250203-third-party-storage-support.md
Vicente-Cheng: nice catch. It's a typo.

Updated!

enhancements/20250203-third-party-storage-support.md
Vicente-Cheng: Already updated in the API change.

PR #7482, opened 2025-02-03
Add upgrade path processing of issue 7442

w13915984028:
**Problem:**

Harvester managedchart is marked as harvester-snapshot-validation-webhook modified

**Solution:**

Add those dynamic resources (updated each time chart is updated) to diff path of managedchart.

**Related Issue:**
https://github.com/harvester/harvester/issues/7442

**Test plan:**

Update harvester managedchart, observe the managedchart is not complained.

code PR:  https://github.com/harvester/harvester-installer/pull/939
backport v1.4 https://github.com/harvester/harvester-installer/pull/947
backport v1.5 https://github.com/harvester/harvester-installer/pull/946

note: this PR assumes there are v1.4.0, v1.4.1, v1.4.2, v1.4.3 releases, both v1.5 and v1.4.3(latest) can add this PR

PR #7528, opened 2025-02-06
Fix incorrect Grafana CPU Usage panel

brandboat:
**Problem:**
Incorrect Grafana CPU Usage panel

**Solution:**
Fix incorrect Grafana CPU Usage panel in upgrade path. installer pr is here: https://github.com/harvester/harvester-installer/pull/951

**Related Issue:**
https://github.com/harvester/harvester/issues/7086

**Test plan:**
https://github.com/harvester/harvester/issues/7086#issuecomment-2632801527, upgrade from v1.4.2 -> v1.5.0

### DIFF
--- a/package/upgrade/migrations/upgrade_manifests/v1.4.2/pre-hook.sh
+++ b/package/upgrade/migrations/upgrade_manifests/v1.4.2/pre-hook.sh
@@ -1,0 +1,62 @@
+#!/bin/bash -x
+
+patch_harvester_vm_dashboard()
+{
+    echo "Patch configmap/harvester-vm-dashboard..."
+    local patch_json=$( \
+        kubectl get -n cattle-dashboards configmap harvester-vm-dashboard -o json | \
+            jq '.data."harvester_vm_dashboard.json" | fromjson | .panels |=
+                map(
+                    if .title == "CPU Usage" then
+                        .yaxes |= map(
+                            if .format == "percent" then
+                                .format = "percentunit"
+                            else
+                                .
+                            end
+                        ) |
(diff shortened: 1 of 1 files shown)


package/upgrade/migrations/upgrade_manifests/v1.4.2/pre-hook.sh
```diff
@@ -0,0 +1,62 @@
...
+                    else
+                        .
+                    end
+                )' |
+            jq -r .
+    )
+
+    if [ -z "$patch_json" ]; then
+        echo "Something goes wrong, the patch json is empty... skip this patch"
+        return 1
```
w13915984028: if we skip, should 0 be returned?

w13915984028 (approved):
LGTM, thanks.

Please also take a look https://github.com/harvester/harvester/pull/7482

maybe we need to patch v1.4.0, v1.4.1, v1.4.2 if we allow v1.4.* to v1.5.0 upgrade. The code itself has no harm if the values have been patched.

brandboat:
> maybe we need to patch v1.4.0, v1.4.1, v1.4.2 if we allow v1.4.* to v1.5.0 upgrade. The code itself has no harm if the values have been patched.

I just confirmed with @bk201 , and the desired upgrade path is v1.4.0 -> v1.4.1 -> v1.4.2 -> v1.5.0.

package/upgrade/migrations/upgrade_manifests/v1.4.2/pre-hook.sh
brandboat: Updated, thanks!

bk201 (approved):

PR #7534, opened 2025-02-06
chore: bump kind to v0.24 to match RKE2 1.31/Rancher 2.10

ihcsim:

**Problem:**
The KinD version used to run the integration test is quite old. It provisioned K8s v1.24 which was EOL 2 years ago. Since then changes have been made to the K8s core API. We should bump it to match RKE2 1.31 and Rancher 2.10.

**Solution:**
This PR updates `go.mod` and `vendor` to use KinD v0.24 which provisions  K8s v1.30.

**Related Issue:**

**Test plan:**
CI integration test must pass.


FrankYang0529 (approved):

Vicente-Cheng (approved):
thanks for that!

Issue #7335, opened 2025-01-09
[ENHANCEMENT] Checking Backup Target Setting and Compatibility with LH v1.8


WebberHuang1118:
**Is your enhancement related to a problem? Please describe.**
Since LH v1.8.0, LH the settings for the backup target will change (some old settings are removed), for the Harvester side, we have to check if we have to modify the controller accordingly and the backward compatibility

LEP: https://github.com/longhorn/longhorn/pull/6630/files#diff-f66bf27d07bfda116e6936671071bf1ef0ce469653a32bdb0c1c3571f8a6c1b5R331

Documents: https://github.com/longhorn/website/pull/1035

Preview pages: https://deploy-preview-1035--longhornio.netlify.app/docs/1.8.0/references/helm-values/#other-settings

**Describe the solution you'd like**
As discussed with @mantissahz, for the upgraded path, the backup information will be kept in CR `BackupTarget`: `default`. 

However, for applying the backup setting in `LH v1.8.0`, the Harvester backup controller should change [the implementation](https://github.com/harvester/harvester/blob/9609b9038f3724e584cfc432aaedd0274b6d8cc8/pkg/controller/master/backup/backup_target.go#L79-L128)

cc @bk201 @FrankYang0529 @derekbit @mantissahz

noahgildersleeve:
Currently blocked b y #7819 . I was going to test this by installing `v1.4.2` with LH `1.7.2`, then upgrade Harvester to `v1.5.0-rc1` with LH `1.8.1`. However it looks like that is going to be blocked by the upgrade issue.

w13915984028:
@WebberHuang1118 Could you add a `test-plan`, I am going to validate this issue, thanks.

w13915984028:
Test result:  **PASS**

Test setup: v1.5.0-rc2 single-node cluster, NFS backup server

Test steps:
(1) Setup a NFS backup server, the refreshIntervalInSeconds is from default value
```
{
  "type": "nfs",
  "endpoint": "nfs://192.168.122.159:/nfs_server/nfs_share/",
  "refreshIntervalInSeconds": 0
}
```

(2) Create a VM and take a backup

```
$kubectl get vmbackup.harvesterhci
NAME          SOURCE_KIND      SOURCE_NAME   TYPE     READY_TO_USE   AGE   ERROR
vm1-backup1   VirtualMachine   vm1           backup   true           93s   
```

(3) Check backup

backup is successful

[...]

PR #7536, opened 2025-02-07
fix: backup target compatibility for LH v1.8

WebberHuang1118:
**Problem:**
Fix the Harvester backup-target controller to be compatible with LH v1.8 https://github.com/longhorn/longhorn/pull/6630/files#diff-f66bf27d07bfda116e6936671071bf1ef0ce469653a32bdb0c1c3571f8a6c1b5R331

**Solution:**
Fix the Harvester backup-target controller to be compatible with LH v1.8 

**Related Issue:**
#7335 

**Test plan:**
- Update LH CSI plugin daemonset with `phanle1010/longhorn-manager:v1.8.0-10303`
  - To avoid LH issue https://github.com/longhorn/longhorn/issues/10303
- Configuring an NFS server as a backup-target
- Starting a VM `vm1`, and having a VMBackup `vm1-b` for `vm1`
- Restoring the VMBackup `vm-b1` to a new VM or replacing the existing

**Note:**
Due to #7539, recovering an existing VMBackup from backup target or restoring VMBackup to another namespace is not in the scope of this PR


bk201 (approved):

FrankYang0529 (approved):
LGTM. Thanks for the fix.

WebberHuang1118:
@Mergifyio backport v1.5

pkg/controller/master/backup/backup_target.go
```diff
@@ -154,32 +156,20 @@ func (h *TargetHandler) reUpdateBackupTargetSettingSecret(setting *harvesterv1.S
...
-		}); err != nil {
-			return err
-		}
-		return nil
+		return err
 	}
 
-	targetCpy := target.DeepCopy()
-	targetCpy.Value = util.ConstructEndpoint(backupTarget)
+	lhBackupTargetCpy := lhBackupTarget.DeepCopy()
```
w13915984028, line 164: It looks to be a bit tricky.
Per below link
https://github.com/longhorn/longhorn/pull/6630/files#diff-f66bf27d07bfda116e6936671071bf1ef0ce469653a32bdb0c1c3571f8a6c1b5R345

There is a configmap `longhorn-default-backupstore`, which stores the chart default value, and it will be set to LH everytime configmap changes, or LH POD restarts.

When Harvester sets the user configure into this LH setting, the are potential ping-pong setting between Harvester and LH configmap controller.

It may also affect some auto-triggered sheduling backups.

Is this noticed?

This is like the `longhorn` default  LH StorageClass, Harvester creates a `harvester-longhorn` SC to avoid battling with LH.

WebberHuang1118: This is used to apply the default backup target if needed, which is not the case for Harvester.

Issue #7538, opened 2025-02-07
[BUG] Should drop Migrate UI option on Stopped VM if it is not intended


albinsun:
**Describe the bug**
Can **Migrate** a `Stopped` VM
![Image](https://github.com/user-attachments/assets/793bf0ba-c11b-49f3-b71e-899efc48b479)
And will have error message `virtualmachineinstances.kubevirt.io "xxx" not found`
![Image](https://github.com/user-attachments/assets/60b92313-7e1a-460a-af25-f6624a53fac9)

**Expected behavior**
If migrate a `Stopped` VM is not intended, may consider not to show the option on UI.

**Environment**
 - Harvester ISO version: `v1.5.0-dev-20250203`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): QEMU/KVM

w13915984028:
cc @Yu-Jack Your new PR https://github.com/harvester/harvester/pull/7548 may refer this issue as well and move the issue forward, thanks.

Yu-Jack:
This issue and this one https://github.com/harvester/harvester/issues/6806 are relevant. Testing them together would be convenient.

irishgordo:
<img width="2145" height="1916" alt="Image" src="https://github.com/user-attachments/assets/7170ffa8-b803-44ab-8b20-b10fd9925080" />

Thanks for the fix @Yu-Jack this looks good on 😄 👍 :
-  v1.6.0-dev-20250703

PR #7548, opened 2025-02-10
fix: migrate button incorrectly display

Yu-Jack:

**Problem:**
After https://github.com/harvester/harvester/pull/7137, the migrate and abortMigrate button always shows on the menu.

**Solution:**
We still need to consider the situations mentioned in https://github.com/harvester/harvester/issues/7140#issuecomment-2609293463. So, I prefer to add the canMigrate function back to prevent it from breaking basic function first.

**Related Issue:**
https://github.com/harvester/harvester/issues/7538

**Test plan:**

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -15,6 +15,7 @@ import (
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
 	"github.com/harvester/harvester/pkg/indexeres"
 	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/util/virtualmachineinstance"
 )
 
 const (
@@ -73,8 +74,6 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	resource.AddAction(request, addVolume)
 	resource.AddAction(request, removeVolume)
 	resource.AddAction(request, cloneVM)
-	resource.AddAction(request, migrate)
-	resource.AddAction(request, findMigratableNodes)
 
 	if canEjectCdRom(vm) {
 		resource.AddAction(request, ejectCdRom)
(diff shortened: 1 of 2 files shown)


pkg/api/vm/formatter.go
```diff
@@ -234,6 +238,13 @@ func isReady(vmi *kubevirtv1.VirtualMachineInstance) bool {
 	return false
 }
 
+func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
+	if err := virtualmachineinstance.ValidateVMMigratable(vmi); err != nil {
```
w13915984028, line 246: Would suggest to update `ValidateVMMigratable` to update all errors to a straightforward description

`non-live migratable due to running` is a bit difficult to understand

```
	if !vmi.IsRunning() {
		return fmt.Errorf("VM %s considered non-live migratable due to running", vmiNamespacedName)

-> VM %s is not live migratable as it is not running
	}

	// The VM is already in migrating state
	if vmi.Annotations[util.AnnotationMigrationUID] != "" {
		return fmt.Errorf("VM %s considered non-live migratable due to migration state", vmiNamespacedName)

-> VM %s is not live migratable as it already has a running migration instance.

... others are similar
	}
```

w13915984028 (approved):
LGTM, thanks.

pkg/api/vm/formatter.go
Yu-Jack: Sounds good. I'll update PR later.

Yu-Jack: @w13915984028  Done.

innobead (approved):

Yu-Jack:
This PR should be backport to v1.4 since v1.4 already contains wrong codes (Previous backport PR https://github.com/harvester/harvester/pull/7344).

bk201 (approved):

ibrokethecloud (approved):
lgtm

Issue #6732, opened 2024-10-08
[ENHANCEMENT] VMs with `dedicatedCpuPlacement=true` should be shut down at forced maintenance mode


votdev:
**Is your enhancement related to a problem? Please describe.**
The current maintenance mode (with `forced` checkbox is enabled) tries to migrate a VM with `spec.template.spec.cpu.dedicatedCpuPlacement = true` which is not possible when only one node has `CPU Manager` enabled. Because of that the node is stuck at `Cordoned` and the VM is tried to be migrated in a loop.

**Describe the solution you'd like**
The [FindNonMigratableVMS](https://github.com/harvester/harvester/blob/f7cfbb51deb2da0ca5dcdf8993bc6f3a9372be86/pkg/controller/master/nodedrain/nodedrain_controller.go#L146) function should identify those VMs with enabled CPU pinning which are not migratable because only one node has `CPU Manager` enabled.
Those VMs are automatically shut down at the beginning of the maintenance mode which seems to be the best solution here to do not block the node to go into maintenance mode.

**Describe alternatives you've considered**
n/a

**Additional context**

brandboat:
Test OK

Environment: 3 node, 8C/16G/250GB for each node, built on commit https://github.com/harvester/harvester/commit/cfc43722e5bde3df6a342459d39d17db66e681ad

***Case 1*** - **Test OK** ✅ 
- Create two VMs vm01 and vm02. Make sure they are running on the same node.
- Enable `CPU Manager` on the host where the two VMs are running. Make sure only one node has enabled this feature.
- Check the `Enable CPU Pinning` checkbox for vm01 on the `Advanced Settings` page.
- Enable `Maintenance Mode` on the node where the VMs are running. Check the `Force` checkbox and then `Apply`.
- Make sure the node affinity rule is added to the VM.
- The VM vm01 is shut down and NOT migrated. The VM vm02 is migrated to a different node.

![Image](https://github.com/user-attachments/assets/e70e87dd-b485-4929-b329-8d4e437df10c)
![Image](https://github.com/user-attachments/assets/ee750d62-64bd-484b-bf7d-52dc0f2f73aa)

***Case 2*** - **Test OK** ✅ 
- Create two VMs vm01 and vm02. Make sure they are running on the same node.
- Enable `CPU Manager` on the host where the two VMs are running. Make sure there is another node with `CPU Manager` enabled.
- Check the `Enable CPU Pinning` checkbox for vm01 on the `Advanced Settings` page.
[...]

PR #7559, opened 2025-02-10
Set node affinity when CPU pinning is set

votdev:
**Problem:**
The current maintenance mode (with `forced` checkbox is enabled) tries to migrate a VM with `spec.template.spec.cpu.dedicatedCpuPlacement = true` which is not possible when only one node has `CPU Manager` enabled. Because of that the node is stuck at `Cordoned` and the VM is tried to be migrated in a loop.

**Solution:**
~~The [FindNonMigratableVMS](https://github.com/harvester/harvester/blob/f7cfbb51deb2da0ca5dcdf8993bc6f3a9372be86/pkg/controller/master/nodedrain/nodedrain_controller.go#L146) function should identify those VMs with enabled CPU pinning which are not migratable because only one node has `CPU Manager` enabled.
Those VMs are automatically shut down at the beginning of the maintenance mode which seems to be the best solution here to do not block the node to go into maintenance mode.~~

Make sure the node affinity rule
```
spec:
  affinity:
    nodeAffinity:
        nodeSelectorTerms:
          - matchExpressions:
              - key: cpumanager
                operator: In
                values:
                  - true
```
is set via mutator when `spec.template.spec.domain.cpu.dedicatedCpuPlacement` is set.
In that case the existing code base will handle the automatic shutdown or re-scheduling of the VM correctly.

**Related Issue:**
https://github.com/harvester/harvester/issues/6732

**Test plan:**
A cluster with 3 nodes is required.

***Case 1***
- Create two VMs vm01 and vm02. Make sure they are running on the same node.
- Enable `CPU Manager` on the host where the two VMs are running. Make sure only one node has enabled this feature.
- Check the `Enable CPU Pinning` checkbox for vm01 on the `Advanced Settings` page.
- Enable `Maintenance Mode` on the node where the VMs are running. Check the `Force` checkbox and then `Apply`.
- Make sure the node affinity rule is added to the VM.
- The VM vm01 is shut down and NOT migrated. The VM vm02 is migrated to a different node.

In the logs of the harvester-xxx-xxx pods you should find something like this:
```
time="2025-02-10T10:28:24Z" level=info msg="force stopping VM" namespace=default ... virtualmachine_name=vm01
```

***Case 2***
- Create two VMs vm01 and vm02. Make sure they are running on the same node.
- Enable `CPU Manager` on the host where the two VMs are running. Make sure there is another node with `CPU Manager` enabled.
- Check the `Enable CPU Pinning` checkbox for vm01 on the `Advanced Settings` page.
- Enable `Maintenance Mode` on the node where the VMs are running. Check the `Force` checkbox and then `Apply`.
- Make sure the node affinity rule is added to the VM.
- The VM vm01 is NOT shut down; it is migrated to the other node with `CPU Manager` enabled. The VM vm02 is migrated to a different node.

***Case 3***
- Create a VM vm03. Check the `Enable CPU Pinning` checkbox on the `Advanced Settings` page before pressing the `Create` button.
- Press `Apply` to create the VM.
- Choose `Edit YAML` in the action menu.
- Make sure the node affinity rule `spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution` is added to the VM.
```
spec:
  affinity:
    nodeAffinity:
        nodeSelectorTerms:
          - matchExpressions:
              - key: cpumanager
                operator: In
                values:
                  - true
```
- Remove the `dedicatedCpuPlacement` line in the YAML editor.
- Press `Save` button.
- Choose `Edit YAML` in the action menu.
- Make sure the previously added node affinity rule has been removed.

***Case 4***
- Make sure that NO node has `CPU Manager` enabled.
- Create a VM vm05.
- Use the `Edit Config` action menu. Check the `Enable CPU Pinning` checkbox from the `Advanced Settings` page.
- Press `Save` and confirm to restart the VM after saving.
- The VM should not be started. See attached screenshot.

Bildschirmfoto vom 2025-02-11 17-15-08

### DIFF
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -362,7 +362,7 @@ func (m *vmMutator) patchAffinity(vm *kubevirtv1.VirtualMachine, patchOps types.
 	newVMNetworks := vm.Spec.Template.Spec.Networks
 	logrus.Debugf("newNetworks: %+v", newVMNetworks)
 
-	if err := m.addNodeSelectorTerms(vm.Namespace, requiredNodeSelector, newVMNetworks); err != nil {
+	if err := m.addNodeSelectorTerms(vm, requiredNodeSelector, newVMNetworks); err != nil {
 		return patchOps, err
 	}
 
@@ -436,12 +436,12 @@ func makeAffinityFromVMTemplate(template *kubevirtv1.VirtualMachineInstanceTempl
 		affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = &v1.NodeSelector{}
 	}
 
-	// clear node selector terms whose key contains the prefix "network.harvesterhci.io"
+	// clear node selector terms whose key contains the prefix "network.harvesterhci.io" or equals "cpumanager"
 	nodeSelectorTerms := make([]v1.NodeSelectorTerm, 0, len(affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms))
 	for _, term := range affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
 		expressions := make([]v1.NodeSelectorRequirement, 0, len(term.MatchExpressions))
 		for _, expression := range term.MatchExpressions {
-			if !strings.HasPrefix(expression.Key, networkGroup) {
+			if !strings.HasPrefix(expression.Key, networkGroup) && (expression.Key != kubevirtv1.CPUManager) {
 				expressions = append(expressions, expression)
 			}
 		}
@@ -455,9 +455,9 @@ func makeAffinityFromVMTemplate(template *kubevirtv1.VirtualMachineInstanceTempl
 	return affinity
 }
 
-func (m *vmMutator) addNodeSelectorTerms(defaultNamespace string, nodeSelector *v1.NodeSelector, incrementalNetworks []kubevirtv1.Network) error {
+func (m *vmMutator) addNodeSelectorTerms(vm *kubevirtv1.VirtualMachine, nodeSelector *v1.NodeSelector, incrementalNetworks []kubevirtv1.Network) error {
 	for _, network := range incrementalNetworks {
-		nodeSelectorRequirement, err := m.getNodeSelectorRequirementFromNetwork(defaultNamespace, network)
+		nodeSelectorRequirement, err := m.getNodeSelectorRequirementFromNetwork(vm.Namespace, network)
 		if err != nil {
 			return err
 		}
@@ -478,6 +478,27 @@ func (m *vmMutator) addNodeSelectorTerms(defaultNamespace string, nodeSelector *
 			nodeSelector.NodeSelectorTerms = []v1.NodeSelectorTerm{{
 				MatchExpressions: []v1.NodeSelectorRequirement{*nodeSelectorRequirement},
 			}}
+		}
+	}
+
+	if isDedicatedCPU(vm) {
+		requirement := v1.NodeSelectorRequirement{
+			Key:      kubevirtv1.CPUManager,
+			Operator: v1.NodeSelectorOpIn,
+			Values:   []string{"true"},
+		}
+
+		if len(nodeSelector.NodeSelectorTerms) == 0 {
+			nodeSelector.NodeSelectorTerms = []v1.NodeSelectorTerm{{
+				MatchExpressions: []v1.NodeSelectorRequirement{requirement},
+			}}
+		} else {
+			for i, term := range nodeSelector.NodeSelectorTerms {
+				if _, ok := isContainTargetNodeSelectorRequirement(term, requirement); !ok {
+					term.MatchExpressions = append(term.MatchExpressions, requirement)
+					nodeSelector.NodeSelectorTerms[i] = term
+				}
+			}
 		}
 	}
 
--- a/pkg/webhook/resources/virtualmachine/mutator_test.go
+++ b/pkg/webhook/resources/virtualmachine/mutator_test.go
@@ -985,6 +985,49 @@ func TestPatchAffinity(t *testing.T) {
 			},
 		},
 	}
+
+	vm8 := &kubevirtv1.VirtualMachine{
+		Spec: kubevirtv1.VirtualMachineSpec{
+			Template: &kubevirtv1.VirtualMachineInstanceTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{},
+				Spec: kubevirtv1.VirtualMachineInstanceSpec{
+					Domain: kubevirtv1.DomainSpec{
+						CPU: &kubevirtv1.CPU{
+							DedicatedCPUPlacement: true,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	vm9 := &kubevirtv1.VirtualMachine{
+		Spec: kubevirtv1.VirtualMachineSpec{
+			Template: &kubevirtv1.VirtualMachineInstanceTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{},
+				Spec: kubevirtv1.VirtualMachineInstanceSpec{
+					Affinity: &v1.Affinity{
+						NodeAffinity: &v1.NodeAffinity{
+							RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
+								NodeSelectorTerms: []v1.NodeSelectorTerm{
+									{
+										MatchExpressions: []v1.NodeSelectorRequirement{
+											{
+												Key:      kubevirtv1.CPUManager,
+												Operator: v1.NodeSelectorOpIn,
+												Values:   []string{"true"},
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
 	net1 := &cniv1.NetworkAttachmentDefinition{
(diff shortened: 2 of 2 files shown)


w13915984028:
cc @brandboat

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -513,3 +524,45 @@ func (ndc *ControllerHandler) listVMILabelMaintainModeStrategy(node *corev1.Node
...
+// are impacted.
+func (ndc *ControllerHandler) checkDedicatedCpuPlacementRequirements(node *corev1.Node, vmiList []*kubevirtv1.VirtualMachineInstance) ([]string, error) {
+	var impactedVMI []string
+
+	// Skip if the `CPU Manager` feature is disabled on this node. This
+	// indicates there is no running virtual machine with enabled CPU
+	// pinning.
+	value, ok := node.Labels[kubevirtv1.CPUManager]
+	if ok && value != "true" {
+		return nil, nil
```
w13915984028: `!ok || value != "true"`  or directly `value != "true"`

if node has no such label, the returned `value` is `""`

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -513,3 +524,45 @@ func (ndc *ControllerHandler) listVMILabelMaintainModeStrategy(node *corev1.Node
...
+	// Get number of nodes with enabled `CPU Manager` feature. Exit
+	// immediately if there are more than or less than 1 node with this
+	// feature enabled.
+	selector := labels.Set{kubevirtv1.CPUManager: "true"}.AsSelector()
+	nodeList, err := ndc.nodeCache.List(selector)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list nodes with labels %s: %v", selector.String(), err)
+	}
+
+	if len(nodeList) != 1 {
```
w13915984028: It may still not be enough to ensure that the VM can be migrated to another node.

e.g. peer node does not have enough CPU cores to pin, or CPU architecture mismatching ...

For those dynamic scheduling results directed by by `kubevirt` & `kubelet`, Harvester is hard to get a decision immediatly.

A further enhancement will be: the node-drain controller watches vmim instances, and if some vmims fail due to scheduling failure, then stop those VMs accordingly.

w13915984028 (approved):
LGTM, thanks.

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: > It may still not be enough to ensure that the VM can be migrated to another node.
> 
> e.g. peer node does not have enough CPU cores to pin, or CPU architecture mismatching ...
> 
> For those dynamic scheduling results directed by by `kubevirt` & `kubelet`, Harvester is hard to get a decision immediatly.
> 

Ok, I agree with you here. Nevertheless, I think that an optimistic prediction makes sense here.

> A further enhancement will be: the node-drain controller watches vmim instances, and if some vmims fail due to scheduling failure, then stop those VMs accordingly.

This should cushion the previous optimistic prediction in the event of an emergency.

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -513,3 +524,45 @@ func (ndc *ControllerHandler) listVMILabelMaintainModeStrategy(node *corev1.Node
 	return virtualmachineinstance.ListByNode(node, labels.NewSelector().Add(*req),
 		ndc.virtualMachineInstanceCache)
 }
+
+// checkDedicatedCpuPlacementRequirements checks if there are any VMs with
+// CPU pinning enabled. If the number of nodes with enabled `CPU Manager`
+// feature is not >= 2, then the function will return a list of VMs that
+// are impacted.
+func (ndc *ControllerHandler) checkDedicatedCpuPlacementRequirements(node *corev1.Node, vmiList []*kubevirtv1.VirtualMachineInstance) ([]string, error) {
```
ihcsim: I am just wondering if the placement logic here should be part of the `CheckVMISchedulingRequirements()` method where the [`validNodes`](https://github.com/harvester/harvester/blob/ec4b4cecaa29cefcdc29208fe8b1faed1219478e/pkg/controller/master/nodedrain/nodedrain_controller.go#L437-L441) is extended to include those with the `kubevirtv1.CPUManager: "true"` label, if `node` has CPU manager running. I am just thinking if we actually need to add a separate method that loops through all the nodes and VMI, whenever we need to make a new placement decision. 

Also looks like `CheckVMISchedulingRequirements()` already has some existing [unit tests](https://github.com/harvester/harvester/blob/ec4b4cecaa29cefcdc29208fe8b1faed1219478e/pkg/controller/master/nodedrain/nodedrain_controller_test.go#L628). It can be extended to ensure no regressions are introduced. WDYT?

pkg/controller/master/nodedrain/nodedrain_controller.go
brandboat: Did the conversation resolve unintentionally? I didn't see the comment being addressed.

pkg/controller/master/nodedrain/nodedrain_controller.go
ibrokethecloud: should we not filter `nodeList` for current node before running the check? 

A better check would be `nodeList >= 1` after applying the filter or `nodeList >= 2` without the filter before we return?

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: After doing some more research and code review i think the origin problem is that when the `Enable CPU Pinning` checkbox is selected on a VM in the `Advanced Options` tab, then no node affinity rule like 
```
spec:
  affinity:
    nodeAffinity:
        nodeSelectorTerms:
          - matchExpressions:
              - key: cpumanager
                operator: In
                values:
                  - true
```
is added to the VM. If those would be the case, the existing code will handle that correctly. If no additional node with CPU mannager enabled exists, the VM is shut down, otherwise it is migrated to the other node.
So the only change to fix this issue is to append the node affinity when the `Enable CPU Pinning` checkbox is set (and remove it when unchecked).

ihcsim: This is great! :heart: To confirm my understanding, this is now relying on the existing node drain controller's `CheckVMISchedulingRequirements()` to find another node with cpu manager enabled, right? If none exist, then the `FindNonMigratableVMS()` will shutdown the VM.

ihcsim (approved):

brandboat (approved):
LGTM, thanks for the fix!

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -481,6 +481,26 @@ func (m *vmMutator) addNodeSelectorTerms(defaultNamespace string, nodeSelector *
 		}
 	}
 
+	if isDedicatedCPU(vm) {
+		requirement := v1.NodeSelectorRequirement{
+			Key:      kubevirtv1.CPUManager,
+			Operator: v1.NodeSelectorOpExists,
```
brandboat: I have one question: Why this is `Exists` instead of `in (true)` as you mentioned in https://github.com/harvester/harvester/pull/7559#discussion_r1950677387? We can enable cpumanager then disable it, then the cpumanager label could be false, right?

votdev: I'm also a little undecided here. But I haven't found any clear documentation that it always has to be 'true'. You could also use `yes` or `1` as the value. I have found some code where it checks for `true`, but as long as I have not seen any Kubernetes documentation where it is explicitly specified that boolean values are always mapped with the string `true` I am skeptical.

brandboat: > You could also use yes or 1 as the value.

What I know is the `cpumanager` label is introduced by kubevirt, from the code
https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-handler/heartbeat/heartbeat.go#L145, looks like it convert a bool value to string, so it should be either "true" or "false" I thought.

w13915984028:
> Case 4
Make sure that NO node has CPU Manager enabled.
Create a VM vm05. Check the Enable CPU Pinning checkbox for vm05 on the Advanced Settings page before pressing the Create button.
The VM should not be started.

@brandboat Do we have webhook to report error in such case when user click `Create` button? thanks.

brandboat:
> @brandboat Do we have webhook to report error in such case when user click Create button? thanks.

No, but if there is no node enabled cpu manager, the UI will show a warning.
image

votdev:
> > @brandboat Do we have webhook to report error in such case when user click Create button? thanks.
> 
> No, but if there is no node enabled cpu manager, the UI will show a warning. image

Hmm, but i was able to create the VM. Would it make sense to deny/block the creation in that case?

brandboat:
> Hmm, but i was able to create the VM. Would it make sense to deny/block the creation in that case?

Sounds like a good idea! Yeah I think we can do that.

pkg/webhook/resources/virtualmachine/mutator.go
votdev: I've adapted the code. It is now using the `NodeSelectorOpIn` operator.

votdev:
> > Hmm, but i was able to create the VM. Would it make sense to deny/block the creation in that case?
> 
> Sounds like a good idea! Yeah I think we can do that.

I've opened https://github.com/harvester/harvester/issues/7598 to implement that validator.

bk201:
cc @votdev don't forget to backport PR to stable branch, thanks!

PR #7573, opened 2025-02-11
fix(upgrade): set repo vm cpu model to host-passthru

starbops:

**Problem:**

The default CPU model for virtual machines is `host-model`. It was used since we did not specify the CPU model for the upgrade repo VM.

**Solution:**

Per suggested in https://github.com/harvester/harvester/issues/7373#issuecomment-2636027816, explicitly specify the model to be `host-passthrough`.

**Related Issue:**

#7373

**Test plan:**

1. Prepare a Harvester cluster with at least two nodes using the built ISO image from this PR (for QAs, please use the master-head ISO image)
2. Trigger the upgrade with the same ISO image
3. Monitor the upgrade progress and the upgrade repo VM
4. The upgrade should succeed without the upgrade repo VM being shut down during the time

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_repo.go
+++ b/pkg/controller/master/upgrade/upgrade_repo.go
@@ -204,6 +204,7 @@ func (r *Repo) createVM(image *harvesterv1.VirtualMachineImage) (*kubevirtv1.Vir
 							Cores:   1,
 							Sockets: 1,
 							Threads: 1,
+							Model:   kubevirtv1.CPUModeHostPassthrough,
 						},
 						Firmware: &kubevirtv1.Firmware{
 							Bootloader: &kubevirtv1.Bootloader{


bk201 (approved):

pkg/controller/master/upgrade/upgrade_repo.go
```diff
@@ -204,6 +204,7 @@ func (r *Repo) createVM(image *harvesterv1.VirtualMachineImage) (*kubevirtv1.Vir
 							Cores:   1,
 							Sockets: 1,
 							Threads: 1,
+							Model:   "host-passthrough",
```
w13915984028: Refer the definition `CPUModeHostPassthrough` ?

```
vendor/kubevirt.io/api/core/v1/schema.go:	CPUModeHostPassthrough                 = "host-passthrough"
vendor/kubevirt.io/kubevirt/pkg/virt-controller/services/nodeselectorrenderer.go:	if nsr.cpuModelLabel != "" && nsr.cpuModelLabel != cpuModelLabel(v1.CPUModeHostModel) && nsr.cpuModelLabel != cpuModelLabel(v1.CPUModeHostPassthrough) {
vendor/kubevirt.io/kubevirt/pkg/defaults/arm64.go:	defaultCPUModelArm64 = v1.CPUModeHostPassthrough
```

w13915984028 (approved):
LGTM, thanks.

pkg/controller/master/upgrade/upgrade_repo.go
starbops: Updated. Thank you.

bk201: @starbops The latest code still seems to use a string.

starbops: Sorry I made a mistake. It should be fine now.

Issue #7518, opened 2025-02-06
[BUG] after initial installation, passive.img uses 3.1G of disk space, vs. active.img which only uses 1.7G


tserong:
**Describe the bug**
The passive OS image uses almost 2x the space of the active OS image, which reduces the amount of available free space in the COS_STATE partition, which in turn can cause problems for upgrades.

**To Reproduce**
1. Install Harvester
2. Run `du -sh /run/initramfs/cos-state/cOS/*`
3. Observe the following output:
```
1.7G    /run/initramfs/cos-state/cOS/active.img
3.1G    /run/initramfs/cos-state/cOS/passive.img
```

**Expected behavior**
active.img and passive.img use the same amount of disk space (they're meant to be identical at this point in time)

**Support bundle**
N/A

**Environment**
 - Harvester ISO version: v1.4.0, but confirmed the same on v1.3.2
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): any

**Additional context**
I've tracked this back to Elemental's `CopyFileImg()` function:

https://github.com/rancher/elemental-toolkit/blob/d16cafcb837b3537a376016514e5f7042f35c8ed/pkg/elemental/elemental.go#L430-L431

...which in turn uses golang's `io.Copy()`:

https://github.com/rancher/elemental-toolkit/blob/d16cafcb837b3537a376016514e5f7042f35c8ed/pkg/utils/common.go#L116

At this point, `active.img` is actually a sparse file (size ~3G, but disk usage 1.7G). Because `io.Copy()` just copies bytes, `passive.img` ends up being a full-size (i.e. non-sparse) file.

One hacky way to fix this would be to have the Harvester installer run `cp active.img passive.img` after elemental is done, because `cp` will actually create a sparse destination file, and then both active.img and passive.img will only use 1.7G each.

This was discovered while investigating https://github.com/harvester/harvester/issues/7457

ihcsim:
Tested ok in 1.5-rc3.

PR #7579, opened 2025-02-11
fix: upgrade: ensure passive.img is sparse before upgrading OS

tserong:
**Problem:**
If `passive.img` is not a sparse file, it can result in too much disk space being used on the COS_STATE partition which causes problems for upgrades (see related issues).

**Solution:**
Just before upgrading, run `fallocate --dig-holes` to make `passive.img` sparse if it's not already.

**Related Issue:**
https://github.com/harvester/harvester/issues/7457
https://github.com/harvester/harvester/issues/7518

**Test plan:**
- Install the newest version of Harvester that does not include this fix (v1.4.1 at the time of writing)
- Check disk usage of active.img and passive.img. active.img should be 1.7G, and passive.img should be 3.1G:
  ```
  # du -sh /run/initramfs/cos-state/cOS/*
  1.7G	/run/initramfs/cos-state/cOS/active.img
  3.1G	/run/initramfs/cos-state/cOS/passive.img
  ```
- Upgrade to a release that includes this fix (v1.4.2-rc1 at the time of writing).
- Once the upgrade is complete, check the logs of the single-node-upgrade pod (if testing single node upgrade), or the post-drain pod (if testing multiple node upgrade). You should find an indication that passive.img was made sparse and reduced to 1.7G e.g.:
  ```
  # kubectl -n harvester-system logs pod/hvst-upgrade-XXXXX-single-node-upgrade-NNNNNNN-YYYYYY
  [...]
  + sparsify_passive_img
  + PASSIVE_IMG=/host/run/initramfs/cos-state/cOS/passive.img
  + '[' -f /host/run/initramfs/cos-state/cOS/passive.img ']'
  + echo 'Ensuring /host/run/initramfs/cos-state/cOS/passive.img is sparse...'
  + mount -o rw,remount /host/run/initramfs/cos-state
  ++ du -h /host/run/initramfs/cos-state/cOS/passive.img
  + echo '  was: 3.1G	/host/run/initramfs/cos-state/cOS/passive.img'
  + fallocate --dig-holes /host/run/initramfs/cos-state/cOS/passive.img
    was: 3.1G	/host/run/initramfs/cos-state/cOS/passive.img
  ++ du -h /host/run/initramfs/cos-state/cOS/passive.img
  + echo '  now: 1.7G	/host/run/initramfs/cos-state/cOS/passive.img'
  + mount -o ro,remount /host/run/initramfs/cos-state
    now: 1.7G	/host/run/initramfs/cos-state/cOS/passive.img
  [...]
  ```

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -29,6 +29,22 @@ cleanup_incomplete_state_file() {
   fi
 }
 
+sparsify_passive_img()
+{
+  # See https://github.com/harvester/harvester/issues/7518
+  PASSIVE_IMG=${STATE_DIR}/cOS/passive.img
+  if [ -f $PASSIVE_IMG ]; then
+    echo "Ensuring $PASSIVE_IMG is sparse..."
+    mount -o rw,remount ${STATE_DIR}
+    echo "  was: $(du -h $PASSIVE_IMG)"
+    fallocate --dig-holes $PASSIVE_IMG
+    echo "  now: $(du -h $PASSIVE_IMG)"
+    mount -o ro,remount ${STATE_DIR}
+  else
+    echo "$PASSIVE_IMG does not exist"
+  fi
+}
+
 is_mounted()
 {
   mount | awk -v DIR="$1" '{ if ($3 == DIR) { exit 0 } } ENDFILE { exit 1 }'
@@ -597,6 +613,9 @@ EOF
 
   # we would like to clean up the incomplete state.yaml to avoid the issue of https://github.com/harvester/harvester/issues/4526
   cleanup_incomplete_state_file
+
+  # make sure the current passive image isn't using too much disk space
+  sparsify_passive_img
 
   elemental_upgrade_log="${UPGRADE_TMP_DIR#"$HOST_DIR"}/elemental-upgrade-$(date +%Y%m%d%H%M%S).log"
   local ret=0


tserong:
(Marked as Draft as I haven't actually built an ISO and tested this yet)

tserong:
So my thinking with this one is that if Harvester is upgraded (i.e. is running new images) before the post-drain step, then it should be the new images including this fix which run, and thus will make `passive.img` sparse before `elemental upgrade`, thereby avoiding the problem. Do I have this right? (I'm still working on trying to test this thing myself)

bk201:
yes, the post-drain job will run the new version script.

tserong:
Cool, looks like this is going to work, except of course there's a small variable substitution error in the code :-/ I'll fix that then do another test..

tserong:
OK, good for review. I ran through a single node upgrade (with an extra disk, so 8G COS_STATE) of v1.4.0 to a build of v1.4.1 with this applied, then after the upgrade was able to see this:

```
# kubectl -n harvester-system logs pod/hvst-upgrade-z89fr-single-node-upgrade-h140-extra-single-scv2p
[...]
+ sparsify_passive_img
+ echo 'Ensuring passive.img is sparse...'
+ mount -o rw,remount /host/run/initramfs/cos-state
Ensuring passive.img is sparse...
++ du -h /host/run/initramfs/cos-state/cOS/passive.img
  was: 3.1G     /host/run/initramfs/cos-state/cOS/passive.img
+ echo '  was: 3.1G     /host/run/initramfs/cos-state/cOS/passive.img'
+ fallocate --dig-holes /host/run/initramfs/cos-state/cOS/passive.img
++ du -h /host/run/initramfs/cos-state/cOS/passive.img
  now: 1.7G     /host/run/initramfs/cos-state/cOS/passive.img
+ echo '  now: 1.7G     /host/run/initramfs/cos-state/cOS/passive.img'
+ mount -o ro,remount /host/run/initramfs/cos-state
[...]
```

...and the upgrade succeeded, and the active.img was not corrupt, so no disk space errors.

Vicente-Cheng (approved):
lgtm, thanks!

bk201 (approved):

bk201:
confirm passive.img is spared after a upgrade.

starbops (approved):
LGTM, thanks.

package/upgrade/upgrade_node.sh
```diff
@@ -29,6 +29,17 @@ cleanup_incomplete_state_file() {
   fi
 }
 
+sparsify_passive_img()
+{
+  # See https://github.com/harvester/harvester/issues/7518
+  echo "Ensuring passive.img is sparse..."
+  mount -o rw,remount ${STATE_DIR}
+  echo "  was: $(du -h ${STATE_DIR}/cOS/passive.img)"
```
w13915984028: A similar question, test `${STATE_DIR}/cOS/passive.img` file is exsiting will be even secure.

tserong: It's extremely unlikely that passive.img will be missing here (someone would have had to intentionally delete it, as in normal operation it's only touched by elemental during `elemental install` or `elemental upgrade`). Still, unlike in the installer where passive.img is guaranteed to exist when we sparsify it (see https://github.com/harvester/harvester-installer/pull/970#discussion_r1953457151), it's possible the file might not be there at this point if something really weird has happened to the system in the meantime, so I'll add a check.

tserong: Done! Here's the relevant portion of the upgrade log:

```
+ sparsify_passive_img
+ PASSIVE_IMG=/host/run/initramfs/cos-state/cOS/passive.img
+ '[' -f /host/run/initramfs/cos-state/cOS/passive.img ']'
+ echo 'Ensuring /host/run/initramfs/cos-state/cOS/passive.img is sparse...'
+ mount -o rw,remount /host/run/initramfs/cos-state
++ du -h /host/run/initramfs/cos-state/cOS/passive.img
+ echo '  was: 3.1G	/host/run/initramfs/cos-state/cOS/passive.img'
+ fallocate --dig-holes /host/run/initramfs/cos-state/cOS/passive.img
  was: 3.1G	/host/run/initramfs/cos-state/cOS/passive.img
++ du -h /host/run/initramfs/cos-state/cOS/passive.img
+ echo '  now: 1.7G	/host/run/initramfs/cos-state/cOS/passive.img'
+ mount -o ro,remount /host/run/initramfs/cos-state
  now: 1.7G	/host/run/initramfs/cos-state/cOS/passive.img
```

Vicente-Cheng:
~I thought we should need this for v1.4.2?~
OOPS! already opened XD

package/upgrade/upgrade_node.sh
w13915984028: Thanks for the updating.

BTW, besides the file is removed, it is also possible that elemental happens to change the name or file path some day.

The shell script runs into `panic mode` by default if any error happens. Those additional checks cover the exteme cases.

Issue #7577, opened 2025-02-11
[BUG] Storage Network can't be Enabled


WebberHuang1118:
**Describe the bug**
After the master branch [bumps the NAD client from v1.3.0 -> v1.7.0](https://github.com/harvester/harvester/blob/a99cd628bf229f0c7b1f1b1de0d937cef873d058/go.mod#L70), configuring the storage network will fail with:
```
harvester-58c4d4bcbf-whnl2:time="2025-02-10T08:41:15Z" level=error msg="error syncing 'storage-network': handler harvester-storage-network-controller: create net-attach-def failed admission webhook \"validator.harvester-system.harvester-network-webhook\" denied the request: Internal error occurred: could not create nad harvester-system/storagenetwork-lwhj4 because the suffix of the brName should be -br, requeuing"
harvester-58c4d4bcbf-whnl2:time="2025-02-10T08:41:15Z" level=error msg="error syncing 'storage-network': handler harvester-storage-network-controller: create net-attach-def failed admission webhook \"validator.harvester-system.harvester-network-webhook\" denied the request: Internal error occurred: could not create nad harvester-system/storagenetwork-fgtr6 because the suffix of the brName should be -br, requeuing"
```

**To Reproduce**
1. Building a Harvester cluster
2. Enabling Storage Network

**Expected behavior**
Storage Network can enable successfully

**Support bundle**

**Environment**
 - Harvester ISO version: master (`a99cd628bf229f0c7b1f1b1de0d937cef873d058`)
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

WebberHuang1118:
Hi @rrajendran17 

I updated the PR https://github.com/harvester/harvester/pull/7405:
- Including bumping Whereabouts `v0.8.0` manifest 
- Including bumping Whereabouts `v0.8.0` module dependency
- Including flatting `bridgeConfig` to avoid `cniv1.NetConf`'s own marshaling
  - `containernetworking/cni` bumps from `v1.1.2` to `v1.2.0` automatically
[...]

PR #7405, opened 2025-01-20
bump(chart): bump Whereabouts to v0.8.0

WebberHuang1118:
**Problem:**
Bump Whereabout to v0.8.0

**Solution:**
Bump Whereabout to v0.8.0

**Additional Context:**
This PR can't be merged until the path of the [extra manifest is changed to `v1.4.1`](https://github.com/WebberHuang1118/harvester/tree/cve-62/package/upgrade/extra_manifests/v1.4.1-rc1) (this should happen after `v1.4.1` release)

**Related Issue:**
https://github.com/harvester/security/issues/62

**Test plan:**
**Deploy to running Harvester**
- Building Harvester cluster with v1.4.1-rc1
- Enabling Storage Network
- Creating a VM
- Applying new Whereabouts CRD
  - $ `kubectl apply -f  deploy/charts/harvester/dependency_charts/whereabouts/crds/whereabouts.cni.cncf.io_ippools.yaml`
  - $ `kubectl apply -f deploy/charts/harvester/dependency_charts/whereabouts/crds/whereabouts.cni.cncf.io_overlappingrangeipreservations.yaml`
- Updating Whereabouts from `v0.7.0` to `v0.8.0`
  - Adding mangedchart `harvester`'s `spec.values` with
  ```
  whereabouts:
    enabled: true
    image:
      pullPolicy: IfNotPresent
      repository: ghcr.io/k8snetworkplumbingwg/whereabouts
      tag: "v0.8.0"
  ```
- Checking if daemoset harvester-whereabouts successfully deployed image `ghcr.io/k8snetworkplumbingwg/whereabouts:v0.8.0`
- Checking if Longhorn `backing-image-manager-xxxx` pods, `instance-manager-xxxx` pods, and `backing-image-ds-xxxx` pods (only appearing during uploading) have interface lhnet1 and its IP in the assigned range
- The VM can stop/restart successfully

**Building New Harvester with this PR**
- Building ISO from this branch, and creating the Harvester cluster with the ISO
- After Harvester brought up, Check if daemonset harvester-whereabouts successfully deploying image `ghcr.io/k8snetworkplumbingwg/whereabouts:v0.8.0`
- Enabling Storage Network
- Checking if Longhorn `backing-image-manager-xxxx` pods, `instance-manager-xxxx` pods, and `backing-image-ds-xxxx` pods (only appearing during uploading) have interface lhnet1 and its IP in the assigned range
- Checking image upload
- Creating VM from uploaded image

[...]

PR #7591, opened 2025-02-11
Change min allocatable ip address formula for LH Pods

rrajendran17:
**Problem:**
https://github.com/harvester/harvester/pull/7405#issuecomment-2649705890

Min allocatable address formula considered backing-image-ds-manager also in the formula which will be dynamic.
For example,
In the issue state, it is single node with just instance-manager pod running so whereabouts allocated 1 IP address, but the formula expected 2 IP address including backing-ds-manager, hence the error.

**Solution:**
Consider only the min allocatable address to be  number of nodes (each running an instance-manager pod).
Whereabouts allocating >= minallocatable address is considered the success case

**Related Issue:**
https://github.com/harvester/harvester/issues/7577
https://github.com/harvester/harvester/pull/7405

**Test plan:**
1.Configure storage network from Advanced-setting Harvester UI
2.verify the following 
https://docs.harvesterhci.io/v1.5/advanced/storagenetwork/#after-applying-harvester-storage-network-setting

### DIFF
--- a/pkg/controller/master/storagenetwork/storage_network.go
+++ b/pkg/controller/master/storagenetwork/storage_network.go
@@ -18,6 +18,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
+	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/config"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
@@ -102,7 +104,6 @@ type Handler struct {
 	longhornSettings                  ctllhv1.SettingClient
 	longhornSettingCache              ctllhv1.SettingCache
 	longhornVolumeCache               ctllhv1.VolumeCache
-	longhornNodeCache                 ctllhv1.NodeCache
 	prometheus                        ctlmonitoringv1.PrometheusClient
 	prometheusCache                   ctlmonitoringv1.PrometheusCache
 	alertmanager                      ctlmonitoringv1.AlertmanagerClient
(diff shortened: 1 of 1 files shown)


WebberHuang1118:
Hi @rrajendran17 

The source code changes look good to me. 

Just I have a question about this description:
> In the issue state, it is single node with just instance-manager pod running so whereabouts allocated 1 IP address, but the formula expected 2 IP address including backing-ds-manager, hence the error.

why this is related to backing-ds or do you mean backing-image-manger? Thanks

rrajendran17:
> Hi @rrajendran17
> 
> The source code changes look good to me.
> 
> Just I have a question about this description:
> 
> > In the issue state, it is single node with just instance-manager pod running so whereabouts allocated 1 IP address, but the formula expected 2 IP address including backing-ds-manager, hence the error.
> 
> why this is related to backing-ds or do you mean backing-image-manger? Thanks

Yeah you are right.It is backing-image-manger and not backing-image-ds.
May be I confused the backing-image-manager and backing-image-ds.
From the formula,
Required Number of IPs = (Number of Nodes * 2) + (Number of Disks * 2) + Number of Images to Download/Upload

Number of Images to Download/Upload corresponds to backing-image-ds pods.
Number of Disks per node corresponds to backing-image-manger pods.

So ideally the formula now do not use the backing-image-manger which is also dynamic and the problem we saw was also backing-image-manager was not running as we did not have any images created.

WebberHuang1118:
> > Hi @rrajendran17
> > The source code changes look good to me.
> > Just I have a question about this description:
> > > In the issue state, it is single node with just instance-manager pod running so whereabouts allocated 1 IP address, but the formula expected 2 IP address including backing-ds-manager, hence the error.
> > 
> > 
> > why this is related to backing-ds or do you mean backing-image-manger? Thanks
> 
> Yeah you are right.It is backing-image-manger and not backing-image-ds. May be I confused the backing-image-manager and backing-image-ds. From the formula, Required Number of IPs = (Number of Nodes * 2) + (Number of Disks * 2) + Number of Images to Download/Upload
> 
> Number of Images to Download/Upload corresponds to backing-image-ds pods. Number of Disks per node corresponds to backing-image-manger pods.
> 
> So ideally the formula now do not use the backing-image-manger which is also dynamic and the problem we saw was also backing-image-manager was not running as we did not have any images created.

yes, backing image manager pods only exist if there is a backing image replica on such disk

WebberHuang1118 (approved):
LGTM, thanks.

bk201 (approved):

Issue #7598, opened 2025-02-12
[FEATURE] Add validator webhook to block the creation of a VM when CPU pinning is enabled but no CPU Manager is activated


votdev:
**Is your feature request related to a problem? Please describe.**
Add a validator webhook that throws an error when the `Enable CPU Pinning` checkbox is enabled in the `Advanced Options` tab of a VM resource but no node with enabled `CPU Manager` exists.

There is already a warning below the checkbox, but this is not enough.

**Describe the solution you'd like**
The validator webhook should deny the creation of the VM in that case.

**Describe alternatives you've considered**
Keep the current behavior that the VM can be created but it will stay in unscheduled state.

**Additional context**
Indirectly related to https://github.com/harvester/harvester/issues/6732.

votdev:
Do we want to have that backported to v1.5.1?

TachunLin:
Verified **fixed** on `v1.6.0-rc3` with Rancher `v2.12.0-rc2`. Close this issue.

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ Test 1: Disable CPU Manager, enable cpu pinning in yaml before creation $~~$</summary>

 1. The creation is aborted and with error message

  <img width="1591" height="833" alt="Image" src="https://github.com/user-attachments/assets/ab9cd02f-7ebf-403b-9007-5d79a1d4aa13" />

  https://github.com/user-attachments/assets/a40e8cfa-1cdb-4507-bec8-ce3f0a371f7f
  

 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Test 2: Enable CPU Manager, enable cpu pinning in yaml before creation $~~$</summary>

  1. Can create vm and scheduled on the node have cpu manager enabled 

  <img width="1600" height="792" alt="Image" src="https://github.com/user-attachments/assets/a19ed26e-e9ee-4691-a81e-a9db50e707a2" />
  
  https://github.com/user-attachments/assets/b7b8e465-6ea5-4951-a7e7-b64bf0e3b4d5

 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Test 3: Enable CPU Manager, vm created -> edit yaml to enable cpu pinning $~~$</summary>

  1. Can modify vm to enable cpu pinning after vm created

  <img width="1600" height="783" alt="Image" src="https://github.com/user-attachments/assets/fae24164-d983-4d09-b048-3a5855d4226d" />
  
[...]

PR #7619, opened 2025-02-13
Adding a webhook for the validation of CPU pinning

votdev:
**Problem:**
The UI is doing some checks related to CPU pinning/CPU Manager when creating a new VM, but they are only taken into action when the form view is used. The YAML editor does not do the validation. When an existing VM is modified, no validation is done.

**Solution:**
Add a validator webhook that throws an error when the `Enable CPU Pinning` checkbox is enabled in the Advanced Options tab of a VM resource but no node with enabled CPU Manager exists. The check is only done when the VM is running. This allows the user to create VMs with CPU pinning enabled in advance even if the CPU Manager is currently disabled. Trying to start a VM with enabled CPU pinning and no CPU Manager available will fail because the VM is not schedulable due to the node affinity/selector and also that no node is available for the preemption rights.

**Related Issue:**
https://github.com/harvester/harvester/issues/7598

**Test plan:**
A 3 node cluster is required.
***Case 1***
- Disable `CPU Manager` feature on all nodes.
- Create a new VM. Enter all necessary values. Click `Edit as YAML`.
- Add the following:
```
spec:
  template:
    spec:
      domain:
        cpu:
          dedicatedCpuPlacement: true
```
- Press `Create`.
- The creation is aborted and the error message `admission webhook "validator.harvesterhci.io" denied the request: no nodes with CPU Manager enabled found` is displayed.

***Case 2***
- Enable `CPU Manager` feature on one node.
- Create a new VM. Enter all necessary values. Click `Edit as YAML`.
- Add the following:
```
spec:
  template:
    spec:
      domain:
        cpu:
          dedicatedCpuPlacement: true
```
- Press `Create`.
- The VM is created.

***Case 3***
- Enable `CPU Manager` feature on one node.
- Create a new VM.
- Choose `Edit YAML` in the action menu of the VM.
- Add the following:
```
spec:
  template:
    spec:
      domain:
        cpu:
          dedicatedCpuPlacement: true
```
- Press `Save`.
- The VM is modified without errors.

***Case 4***
- Enable `CPU Manager` feature on one node.
- Create a new VM `vm-1`. Do not enable `CPU Pinning`.
- Disable `CPU Manager` feature on all nodes.
- Choose `Edit YAML` in the action menu of the VM `vm-1`.
- Add the following:
```
spec:
  template:
    spec:
      domain:
        cpu:
          dedicatedCpuPlacement: true
```
- Press `Save`.
- The VM is not modified. The error message `admission webhook "validator.harvesterhci.io" denied the request: no nodes with CPU Manager enabled found` is displayed.

***Case 5***
- Disable `CPU Manager` feature on all nodes.
- Create a new VM.
- The VM is started.
- Choose `Edit Config` in the action menu of the VM.
- Go to the `Advanced Options` tab and check the `Enable CPU Pinning` checkbox.
- Press `Save`.
- The save action is aborted and the error message `admission webhook "validator.harvesterhci.io" denied the request: no nodes with CPU Manager enabled found` is displayed.

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -12,7 +12,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
-	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	ctlharvestercorev1 "github.com/harvester/harvester/pkg/generated/controllers/core/v1"
@@ -41,6 +42,7 @@ func NewValidator(
 	vmiCache ctlkubevirtv1.VirtualMachineInstanceCache,
 	nadCache ctlcniv1.NetworkAttachmentDefinitionCache,
 	settingCache ctlharvesterv1.SettingCache,
+	nodeCache v1.NodeCache,
 ) types.Validator {
 	return &vmValidator{
 		pvcCache:      pvcCache,
@@ -48,6 +50,7 @@ func NewValidator(
 		vmCache:       vmCache,
 		vmiCache:      vmiCache,
 		nadCache:      nadCache,
+		nodeCache:     nodeCache,
 
 		rqCalculator: resourcequota.NewCalculator(nsCache, podCache, rqCache, vmimCache, settingCache),
 	}
@@ -60,6 +63,7 @@ type vmValidator struct {
 	vmCache       ctlkubevirtv1.VirtualMachineCache
 	vmiCache      ctlkubevirtv1.VirtualMachineInstanceCache
 	nadCache      ctlcniv1.NetworkAttachmentDefinitionCache
+	nodeCache     v1.NodeCache
 	settingCache  ctlharvesterv1.SettingCache
 	rqCalculator  *resourcequota.Calculator
 }
@@ -221,6 +225,10 @@ func (v *vmValidator) Create(_ *types.Request, newObj runtime.Object) error {
 		return err
 	}
 
+	if err := v.checkDedicatedCPUPlacement(vm); err != nil {
+		return err
+	}
+
 	if err := v.checkStorageResourceQuota(vm, nil); err != nil {
 		return err
 	}
@@ -239,6 +247,10 @@ func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj run
 	}
 
 	if err := v.checkVMSpec(newVM); err != nil {
+		return err
+	}
+
+	if err := v.checkDedicatedCPUPlacement(newVM); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -452,3 +464,22 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
...
+	}
+
+	selector := labels.Set{kubevirtv1.CPUManager: "true"}.AsSelector()
+	nodeList, err := v.nodeCache.List(selector)
+	if err != nil {
+		return fmt.Errorf("failed to list nodes with labels %s: %v", selector.String(), err)
+	}
+
+	if len(nodeList) == 0 {
+		return werror.NewInvalidError("no nodes with enabled CPU Manager found",
```
w13915984028: no nodes with enabled CPU Manager found
>>
VM requets CPU pinning, but no nodes with enabled CPU Manager are/were found
..., did not find any node with enabled CPU Manager

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -39,13 +40,15 @@ func NewValidator(
 	vmCache ctlkubevirtv1.VirtualMachineCache,
 	vmiCache ctlkubevirtv1.VirtualMachineInstanceCache,
 	nadCache ctlcniv1.NetworkAttachmentDefinitionCache,
+	nodeCache v1.NodeCache,
```
w13915984028, line 45: normally put the v1.* together, e.g. L36

w13915984028 (dismissed):
LGTM, thanks.

brandboat (dismissed):
LGTM, I have tested all the scenarios listed in the pr desc in my testing environment, and all good.

brandboat (dismissed):

w13915984028:
Just recalled another case:

If a VM claims CPU manager is newly created but is put to OFF status, even when there is no CPU manager enabled, it should still be allowed, as this VM is more like a place holder, user may provision many VMs and build the cluster in parallel;  later when some NODEs are enbled with CPU manager feature, user starts those VMs.

Either, the webhook is working on the `starting` VM, to report the exact error earlier; or do nothing, the k8s/kubevirt scheduller will put the VM into unscheduable automatically (not sure the error message is straightforward enough showing like `no matched cpu manager enabled node`);   another similar case: user may define customzed selector for scheduling, it works or not, can only be determined until the VM is put to run.

I have another PR https://github.com/harvester/harvester/pull/7743 to solve network related issues.

Short summary: for an `OFF` VM, Harvester is better to allow user to do any setting as more as possible.

votdev:
> Just recalled another case:
> 
> If a VM claims CPU manager is newly created but is put to OFF status, even when there is no CPU manager enabled, it should still be allowed, as this VM is more like a place holder, user may provision many VMs and build the cluster in parallel; later when some NODEs are enbled with CPU manager feature, user starts those VMs.
> 
> Either, the webhook is working on the `starting` VM, to report the exact error earlier; or do nothing, the k8s/kubevirt scheduller will put the VM into unscheduable automatically (not sure the error message is straightforward enough showing like `no matched cpu manager enabled node`); another similar case: user may define customzed selector for scheduling, it works or not, can only be determined until the VM is put to run.
> 
> I have another PR #7743 to solve network related issues.
> 
> Short summary: for an `OFF` VM, Harvester is better to allow user to do any setting as more as possible.

The error looks like
grafik

w13915984028:
The k8s source code returns error expressing that the pod is not schedulable due to node affinity/selector, and no node is available for preemption either.

And k8s will retry to schedule the pod via a backoff fashion. Later, if scheduling condition is met, the pod(vm) will be placed to running automatically.

On the other hand, if Harvester uses webhook to check all the possible conditions and deny them, it will mean users needs to take care the return error all the time.

A more flexiable solution is to have optioins for user to enable/disable those webhooks. If user prefers IaaS, disable them is better;  when user tends to have UI real-time check, then enable them.

@ibrokethecloud @ihcsim WDYT? thanks.

```
...
	// ErrReasonPod is the reason for Pod's node affinity/selector not matching.
	ErrReasonPod = "node(s) didn't match Pod's node affinity/selector"
...

// Filter checks if the Node matches the Pod .spec.affinity.nodeAffinity and
// the plugin's added affinity.
func (pl *NodeAffinity) Filter(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeInfo *framework.NodeInfo) *framework.Status {
...

	// Ignore parsing errors for backwards compatibility.
	match, _ := s.requiredNodeSelectorAndAffinity.Match(node)
	if !match {
		return framework.NewStatus(framework.UnschedulableAndUnresolvable, ErrReasonPod)
	}

	return nil
}

...
// nodesWherePreemptionMightHelp returns a list of nodes with failed predicates
// that may be satisfied by removing pods from the node.
func nodesWherePreemptionMightHelp(nodes []*framework.NodeInfo, m framework.NodeToStatusMap) ([]*framework.NodeInfo, framework.NodeToStatusMap) {
	var potentialNodes []*framework.NodeInfo
	nodeStatuses := make(framework.NodeToStatusMap)
	for _, node := range nodes {
		name := node.Node().Name
		// We rely on the status by each plugin - 'Unschedulable' or 'UnschedulableAndUnresolvable'
		// to determine whether preemption may help or not on the node.
		if m[name].Code() == framework.UnschedulableAndUnresolvable {
			nodeStatuses[node.Node().Name] = framework.NewStatus(framework.UnschedulableAndUnresolvable, "Preemption is not helpful for scheduling")
			continue
		}
		potentialNodes = append(potentialNodes, node)
	}
	return potentialNodes, nodeStatuses
}
...
```

w13915984028 (approved):
LGTM, thanks.

brandboat (approved):

Issue #7524, opened 2025-02-06
[Enhancement] Misc Enhancement for VM Image


WebberHuang1118:
**Describe the bug**

Based on the discussion in https://github.com/harvester/harvester/pull/7170, we'd like to have the following enhancement for VM Image
- Using `PollUntilContextTimeout` as waiting for LH bids to be ready https://github.com/harvester/harvester/pull/7170#discussion_r1943290105
- Checking if additional permission check is needed or not https://github.com/harvester/harvester/pull/7170#discussion_r1943316012
- Keeping sync bi status even if the bi is successfully imported https://github.com/harvester/harvester/pull/7170#discussion_r1944001162

PR #7640, opened 2025-02-15
Introduce CDI for third-party storage (VM Image/root disk)

Vicente-Cheng:

**Problem:**
To support third-party storage for VM Image/VM root disk

**Solution:**
Add new VMImage backend and VM root disk mechanism for third-party storage.

**Related Issue:**
https://github.com/harvester/harvester/issues/1199

**Test plan:**
- Try to create VM Image with 3rd-party storage (e.g. LVM)
  - Verify download/upload path
- Try to create VM with the above VM image
- ensure the VM Image PVC would have the annotation `harvesterhci.io/goldenImage: "true"`
- ensure this annotation cannot be modified/deleted
- ensure this volume cannot be removed directly (means you need to remove VMimage to remove the golden image PVC)
- ensure the golden image cannot be used with VM disks

For the download path, we should do the following steps until UI change.
1. apply the following yaml file
```
kind: VirtualMachineImageDownloader
metadata:
  name: <image name> NOTE: not display name
  namespace: <image namespace>
spec:
  imageName: <image name>
```

2. Wait until the above CR is ready, and click the download.


Vicente-Cheng:
~The failure of host_apis_tests should not be related to this PR change.
But I will check this.~
Don't know why the host_apis_tests failed in the previous time.
Anyway, it has passed now.

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml
```diff
@@ -105,6 +105,9 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              targetStorageClassName:
```
ihcsim, line 108: Will there be additional CDI-only fields in the future? If yes, I wonder if we should group them under a `cdi` field? E.g.,

```yaml
cdi:
   targetStorageClassName: foo
```

WDYT?

pkg/util/constants.go
```diff
@@ -86,6 +87,14 @@ const (
 	LonghornOptionEncrypted                  = "encrypted"
 	AddonValuesAnnotation                    = "harvesterhci.io/addon-defaults"
 
+	// CDI constants
+	CSIProvisionerLVM        = "lvm.driver.harvesterhci.io"
+	CSIProvisionerLonghorn   = "driver.longhorn.io"
+	AnnoCreatedForDataVolume = "cdi.kubevirt.io/createdForDataVolume"
+	AnnoPopulatorKind        = "cdi.kubevirt.io/storage.populator.kind"
```
ihcsim: Is it worth importing these annotations from `kubevirt.io/containerized-data-importer`, instead of hard-coding them? Or shall we have a `util` method to check PVCs for their `ownerRef` to determine if they are associated to CDI data volumes? WDYT?

pkg/api/vm/formatter.go
```diff
@@ -271,6 +273,23 @@ func (vf *vmformatter) canDoBackup(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv
 		return false
 	}
 
+	// additional check, we did not support backup with CDI volume
+	volumes := vm.Spec.Template.Spec.Volumes
```
ihcsim, line 278: I am assuming CDI doesn't use `vm.Spec.DataVolumeTemplates`?

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml
```diff
@@ -105,6 +105,9 @@ spec:
...
+                description: extra fields for CDI
```
ihcsim: Maybe a better description on the purpose of this field?

pkg/controller/master/pvc/pvc_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if pvc.Annotations[util.AnnoBindCompleted] != "yes" {
+		return pvc, fmt.Errorf("PVC %s/%s is not ready (for yes), with annotations: %v", pvc.Namespace, pvc.Name, pvc.Annotations)
+	}
+
+	// OK, let's sperate the PVC from the DataVolume
+	// 1. remove OwnerReference from PVC
+	// 2. remove DataVolume
+	pvcCpy := pvc.DeepCopy()
+	pvcCpy.Labels = map[string]string{"app.kubernetes.io/component": "storage"}
```
ihcsim: What does this label do?

pkg/controller/master/pvc/pvc_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	if _, find := pvc.Annotations[util.AnnoBindCompleted]; !find {
+		return pvc, fmt.Errorf("PVC %s/%s is not ready, with annotations: %v", pvc.Namespace, pvc.Name, pvc.Annotations)
+	}
+	if pvc.Annotations[util.AnnoBindCompleted] != "yes" {
+		return pvc, fmt.Errorf("PVC %s/%s is not ready (for yes), with annotations: %v", pvc.Namespace, pvc.Name, pvc.Annotations)
+	}
+
+	// OK, let's sperate the PVC from the DataVolume
+	// 1. remove OwnerReference from PVC
+	// 2. remove DataVolume
```
ihcsim: The thinking here is that once the temp PVC is removed, we no longer need the data volume. And we are removing the data volume because we don't want CDI to restore a user-deleted PVC, right?

My first thought is that I think we should just keep the data volume, and let CDI handle deletion the way it's supposed to. If we add a finalizer to the PVC and only remove the finalizer after CDI deleted the data volume, then we won't get into the situation where CDI tried to restore a user-deleted PVC. WDYT?

ihcsim (commented):
Here's the first set of comments and questions. More to come later. There's a lot in here. Maybe in the future, we can break it down into multiple smaller PRs for easier reviews and testing.

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml
Vicente-Cheng: Thanks @ihcsim, I agree with you if we would maintain both backingimage and cdi.
But we might remove the backingImage in the next version, which means at that time, the VMImage will only have one backend CDI so that the field would focus on the CDI. We could avoid adding an extra layer for CDI fields for just one version.

pkg/api/vm/formatter.go
Vicente-Cheng: No, we would like to keep the root disk volume for reuse. So, I prefer to split it from the VM template.
So, we create DV for root volume and attach it to the VM.

pkg/util/constants.go
Vicente-Cheng: Nice suggestion, I introduce the common libs from the CDI.
I use the annotation to check whether or not PVC is created by DataVolume. I thought checking the owner reference could also work, but annotation checking is more straightforward.

pkg/controller/master/pvc/pvc_controller.go
Vicente-Cheng: I would like to reserve the common label from the CDI. But I drop this code now.

pkg/controller/master/pvc/pvc_controller.go
Vicente-Cheng: Thanks @ihcsim,
That's good suggestion. I rewrite the PVC controller to add finalizer for removing the DataVolume with PVC deleting.

I do some tests for it. And looks like it works fine

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml
Vicente-Cheng: updated

pkg/api/image/schema.go
```diff
@@ -20,15 +20,17 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, _ config.Optio
...
 	vmio := common.GetVMIOperator(vmi, vmi.Cache(), http.Client{})
 	downloaders := map[harvesterv1.VMIBackend]backend.Downloader{
 		harvesterv1.VMIBackendBackingImage: backingimage.GetDownloader(bi.Cache(), http.Client{}, vmio),
-		harvesterv1.VMIBackendCDI:          cdi.GetDownloader(),
+		harvesterv1.VMIBackendCDI:          cdi.GetDownloader(http.Client{}, scaled.Management.ClientSet, vmio),
 	}
 	uploaders := map[harvesterv1.VMIBackend]backend.Uploader{
 		harvesterv1.VMIBackendBackingImage: backingimage.GetUploader(bi.Cache(), bids, http.Client{}, vmio),
-		harvesterv1.VMIBackendCDI:          cdi.GetUploader(),
+		harvesterv1.VMIBackendCDI:          cdi.GetUploader(vmi, ctlcdi, ctlcdiupload, http.Client{}, vmio),
```
WebberHuang1118: Based on the original design of https://github.com/harvester/harvester/pull/7170, I'd like to have all backend implementations know about vmimage's details.

So ideally, this better to be:
```suggestion
		harvesterv1.VMIBackendCDI:          cdi.GetUploader(ctlcdi, ctlcdiupload, http.Client{}, vmio),
```

and the backend implementation operators vmimage via `vmio`.

However, considering our schedule, we may defer it to the discussion in #7524

pkg/controller/master/pvc/register.go
```diff
@@ -0,0 +1,26 @@
...
+	ctlpvc := management.CoreFactory.Core().V1().PersistentVolumeClaim()
+
+	pvcHandler := &pvcHandler{
+		dataVolumeClient: dataVolume,
+		pvcClient:        ctlpvc,
+		pvcController:    ctlpvc,
+	}
+
+	ctlpvc.OnRemove(ctx, pvcControllerName, pvcHandler.cleanupDataVolume)
+
```
WebberHuang1118: Can we remove this empty line?

pkg/controller/master/storageprofile/register.go
```diff
@@ -0,0 +1,26 @@
...
+	ctlstorageprofile := management.CdiFactory.Cdi().V1beta1().StorageProfile()
+
+	storageProfileHandler := &storageProfileHandler{
+		scClient:                 sc,
+		storageProfileClient:     ctlstorageprofile,
+		storageProfileController: ctlstorageprofile,
+	}
+
+	ctlstorageprofile.OnChange(ctx, storageProfileControllerName, storageProfileHandler.OnChanged)
+
```
WebberHuang1118: ditto

pkg/controller/master/storageprofile/storage_profile_controller.go
```diff
@@ -0,0 +1,68 @@
...
+		return profile, err
+	}
+	profileSpec := h.generateProfileSpec(sc)
+
+	profileCpy := profile.DeepCopy()
+	profileCpy.Spec = profileSpec
+	if !reflect.DeepEqual(profile, profileCpy) {
+		return h.storageProfileController.Update(profileCpy)
+	}
+
```
WebberHuang1118: ditto

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -66,14 +74,53 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
...
+				if _, err = h.pvcClient.Create(pvcAnno); err != nil {
+					return nil, err
+				}
+				continue
+			} else if err != nil {
 				return nil, err
 			}
-			continue
-		} else if err != nil {
-			return nil, err
```
WebberHuang1118: Could we factor out this part as a function? IMO, the nested condition is kind of comlex

WebberHuang1118:
Hi @Vicente-Cheng 

For the test plan

> **Test plan:**
> - Try to create VM Image with 3rd-party storage (e.g. LVM)
>   - Verify download/upload path
> - Try to create VM with the above VM image
> - ensure the VM Image PVC would have the annotation `harvesterhci.io/goldenImage: "true"`
> - ensure this annotation cannot be modified/deleted
> - ensure this volume cannot be removed directly (means you need to remove VMimage to remove the golden image PVC)
> - ensure the golden image cannot be used with VM disks

IIUC, we should apply the vmimage and the vm by the manifest, could you provide an example? Thanks.

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
 	return nil
 }
 
-func (cb *Backend) AddSidecarHandler() {
+func (b *Backend) AddSidecarHandler() {
+}
+
+func (b *Backend) isDataVolumeCreated(vmImg *harvesterv1.VirtualMachineImage) (bool, error) {
+	targetDVNs := vmImg.ObjectMeta.Namespace
+	targetDVName := vmImg.ObjectMeta.Name
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+		return false, fmt.Errorf("failed to get DataVolume %s/%s: %v", targetDVNs, targetDVName, err)
+	}
+	if apierrors.IsNotFound(err) {
+		return false, nil
+	}
+	return true, nil
+}
+
+func (b *Backend) initializeDownload(vmImg *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
+	virtualSize, err := fetchImageVirtualSize(vmImg.Spec.URL)
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+	if err != nil {
+		return vmImg, fmt.Errorf("failed to fetch image virtual size: %v", err)
+	}
+
+	size, err := fetchImageSize(vmImg.Spec.URL)
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+	// means the image is not qcow format
+	if virtualSize == 0 {
+		virtualSize = size
+	}
+
+	if vmImg.Status.Size == 0 && vmImg.Status.VirtualSize == 0 {
+		logrus.Infof("Update VM Image size (%v) and virtual size (%v) before we create the DataVolume", size, virtualSize)
+		vmImgNew := vmImg.DeepCopy()
+		vmImgNew.Status.Size = size
+		vmImgNew.Status.VirtualSize = virtualSize
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/common.go
```diff
@@ -0,0 +1,175 @@
...
+	return len(p), nil
+}
+
+func (pu *ProgressUpdater) GetCurrentBytesNoLock() int64 {
+	return pu.totalBytes
+}
+
+func generateDVSource(vmImg *harvesterv1.VirtualMachineImage) (*cdiv1.DataVolumeSource, error) {
+	dvSource := &cdiv1.DataVolumeSource{}
+	sourceType := vmImg.Spec.SourceType
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+	// if dataVolume is already created, return
+	created, err := b.isDataVolumeCreated(vmImg)
+	if err != nil {
+		return vmImg, fmt.Errorf("failed to check DataVolume: %v", err)
+	}
+	if created {
+		return vmImg, nil
+	}
+
+	switch vmImg.Spec.SourceType {
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
-func (cb *Backend) UpdateVirtualSize(vmi *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
+func (b *Backend) UpdateVirtualSize(vmi *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
+	// CDI backend does not need this function, the VirtualSize is updated before the DataVolume is created
 	return vmi, nil
 }
 
-func (cb *Backend) Delete(_ *harvesterv1.VirtualMachineImage) error {
+func (b *Backend) Delete(vmImg *harvesterv1.VirtualMachineImage) error {
+	targetDVNs := vmImg.ObjectMeta.Namespace
+	targetDVName := vmImg.ObjectMeta.Name
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+		return b.initializeExportFromVolume(vmImg)
+	default:
+		return vmImg, fmt.Errorf("unsupported source type: %s", vmImg.Spec.SourceType)
+	}
 }
 
-func (cb *Backend) Check(_ *harvesterv1.VirtualMachineImage) error {
+func (b *Backend) Check(vmImg *harvesterv1.VirtualMachineImage) error {
+	targetDVNs := vmImg.ObjectMeta.Namespace
+	targetDVName := vmImg.ObjectMeta.Name
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/api/image/schema.go
Vicente-Cheng: Updated.

I would like to get the latest VM Image to ensure the progress update.
So, I add a new function to get the VM Image from vmi operator.

Vicente-Cheng:
> Hi @Vicente-Cheng
> 
> For the test plan
> 
> > **Test plan:**
> > 
> > * Try to create VM Image with 3rd-party storage (e.g. LVM)
> >   
> >   * Verify download/upload path
> > * Try to create VM with the above VM image
> > * ensure the VM Image PVC would have the annotation `harvesterhci.io/goldenImage: "true"`
> > * ensure this annotation cannot be modified/deleted
> > * ensure this volume cannot be removed directly (means you need to remove VMimage to remove the golden image PVC)
> > * ensure the golden image cannot be used with VM disks
> 
> IIUC, we should apply the vmimage and the vm by the manifest, could you provide an example? Thanks.

The UI part is almost done, and PRs are merged.
We could use the latest UI to complete the whole operation.

Could you try with the latest UI?

pkg/image/cdi/download.go
```diff
@@ -1,19 +1,299 @@
...
+	if err != nil {
+		return ""
+	}
+	targetImage := fmt.Sprintf("%s:%s", image.Repository, image.Tag)
+	return targetImage
+}
+
+func (cd *Downloader) getClusterRepoImage() string {
+	deployContent, err := cd.clientSet.AppsV1().Deployments("cattle-system").Get(context.TODO(), "harvester-cluster-repo", metav1.GetOptions{})
+	if err != nil {
```
ibrokethecloud: we absorb the error and return "" when lookup fails for any reason.

the pod creation will likely fail because container has no image name specified. are we ok with this? I do believe this is an edge case since the repo should be running before harvester initialises

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+		updatedVMImg, err := b.vmio.UpdateVMI(vmImg, vmImgNew)
+		if err != nil {
+			return vmImg, fmt.Errorf("failed to update VM Image size and virtual size: %v", err)
+		}
+		// use latest updated VM Image
+		vmImg = updatedVMImg
+	}
+
+	dvName := vmImg.ObjectMeta.Name
+	dvNamespace := vmImg.ObjectMeta.Namespace
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/common.go
```diff
@@ -0,0 +1,175 @@
...
+		dvSource.PVC = dvSourceExport
+		return dvSource, nil
+	default:
+		return dvSource, fmt.Errorf("unsupported source type: %s", sourceType)
+	}
+}
+
+func generateDVSourceHTTP(vmi *harvesterv1.VirtualMachineImage) *cdiv1.DataVolumeSourceHTTP {
+	return &cdiv1.DataVolumeSourceHTTP{
+		URL: vmi.Spec.URL,
```
WebberHuang1118, line 92: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/common.go
```diff
@@ -0,0 +1,175 @@
...
+		dvSource.HTTP = dvSourceHTTP
+		return dvSource, nil
+	case harvesterv1.VirtualMachineImageSourceTypeUpload:
+		dvSourceUpload := &cdiv1.DataVolumeSourceUpload{}
+		dvSource.Upload = dvSourceUpload
+		return dvSource, nil
+	case harvesterv1.VirtualMachineImageSourceTypeExportVolume:
+		dvSourceExport := &cdiv1.DataVolumeSourcePVC{}
+		dvSourceExport.Name = vmImg.Spec.PVCName
+		dvSourceExport.Namespace = vmImg.Namespace
```
WebberHuang1118, line 82: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+	dataVolumeTemplate := &cdiv1.DataVolume{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      dvName,
+			Namespace: dvNamespace,
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion:         common.HarvesterAPIV1Beta1,
+					Kind:               common.VMImageKind,
+					Name:               vmImg.Name,
+					UID:                vmImg.UID,
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+		return vmImg, fmt.Errorf("failed to get source PVC %s/%s: %v", vmImg.Spec.PVCNamespace, vmImg.Spec.PVCName, err)
+	}
+	size := sourcePVC.Status.Capacity.Storage().Value()
+	virtualSize := size
+
+	if vmImg.Status.Size == 0 && vmImg.Status.VirtualSize == 0 {
+		logrus.Infof("Update VM Image size (%v) and virtual size (%v) before we create the DataVolume", size, virtualSize)
+		vmImgNew := vmImg.DeepCopy()
+		vmImgNew.Status.Size = size
+		vmImgNew.Status.VirtualSize = virtualSize
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+		updatedVMImg, err := b.vmio.UpdateVMI(vmImg, vmImgNew)
+		if err != nil {
+			return vmImg, fmt.Errorf("failed to update VM Image size and virtual size: %v", err)
+		}
+		// use latest updated VM Image
+		vmImg = updatedVMImg
+	}
+
+	dvName := vmImg.ObjectMeta.Name
+	dvNamespace := vmImg.ObjectMeta.Namespace
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+	targetDVNs := vmImg.ObjectMeta.Namespace
+	targetDVName := vmImg.ObjectMeta.Name
+	targetDV, err := b.dataVolumeClient.Get(targetDVNs, targetDVName, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get DataVolume %s/%s: %v", targetDVNs, targetDVName, err)
+	}
+	if apierrors.IsNotFound(err) {
+		logrus.Info("DataVolume not found, waiting for the initialization")
+		return err
+	}
```
WebberHuang1118: Maybe can be compacted?
```suggestion
	if err != nil {
		if apierrors.IsNotFound(err) {
			logrus.Info("DataVolume not found, waiting for the initialization")
		}
		return fmt.Errorf("failed to get DataVolume %s/%s: %v", targetDVNs, targetDVName, err)
	}
```

pkg/image/cdi/download.go
```diff
@@ -1,19 +1,299 @@
...
+		httpClient: httpClient,
+		clientSet:  clientSet,
+		vmio:       vmio,
+	}
 }
 
-func (cd *Downloader) Do(_ *harvesterv1.VirtualMachineImage, _ http.ResponseWriter, _ *http.Request) error {
+func (cd *Downloader) Do(vmImg *harvesterv1.VirtualMachineImage, rw http.ResponseWriter, req *http.Request) error {
+	vmImgName := vmImg.Name
+	vmImgNamespace := vmImg.Namespace
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/upload.go
```diff
@@ -1,20 +1,337 @@
...
+	}
+	logrus.Debugf("Qcow header index: %v, virtual size: %v", qcowHeaderIndex, virtualSize)
+
+	dataContent := rawContent
+	if multipartFormat {
+		dataContent = rawContent[headerEndIndex+4:]
+	}
+	vmImgNew := vmImg.DeepCopy()
+	vmImgNew.Status.Size = fileSize
+	vmImgNew.Status.VirtualSize = virtualSize
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/upload.go
```diff
@@ -1,20 +1,337 @@
...
+
+	// get the latest VMImage
+	vmImg, err = cu.vmio.GetVMImageObj(vmImg.Namespace, vmImg.Name)
+	if err != nil {
+		return fmt.Errorf("failed to get VMImage: %v", err)
+	}
+
+	// create DataVolume
+	dvName := vmImg.ObjectMeta.Name
+	dvNamespace := vmImg.ObjectMeta.Namespace
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+	targetDVName := vmImg.ObjectMeta.Name
+	_, err := b.dataVolumeClient.Get(targetDVNs, targetDVName, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return false, fmt.Errorf("failed to get DataVolume %s/%s: %v", targetDVNs, targetDVName, err)
+	}
+	if apierrors.IsNotFound(err) {
+		return false, nil
+	}
+	return true, nil
+}
```
WebberHuang1118: Suggest this for readability
```suggestion
	if err != nil {
		if !apierrors.IsNotFound(err) {
			return false, fmt.Errorf("failed to get DataVolume %s/%s: %v", targetDVNs, targetDVName, err)
		}
		return false, nil
	}
	return true, nil
```

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,269 @@
...
+	if _, err := b.dataVolumeClient.Create(dataVolumeTemplate); err != nil {
+		return vmImg, fmt.Errorf("failed to create DataVolume %s/%s: %v", dvNamespace, dvName, err)
+	}
+	logrus.Infof("DataVolume %s/%s created", dvNamespace, dvName)
+	return vmImg, nil
+}
+
+func (b *Backend) initializeExportFromVolume(vmImg *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
+	// export from volume means we will create vm image from the raw volume
+	sourcePVC, err := b.pvcCache.Get(vmImg.Spec.PVCNamespace, vmImg.Spec.PVCName)
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

pkg/image/cdi/upload.go
```diff
@@ -1,20 +1,337 @@
...
+		vmImg, err = cu.vmio.GetVMImageObj(vmImg.Namespace, vmImg.Name)
+		if err != nil {
+			logrus.Errorf("[updateVMImageProgress] failed to get VMImage %s/%s: %v, ignore the latest", vmImg.Namespace, vmImg.Name, err)
+			cond.L.Unlock()
+			continue
+		}
+
+		// check if the upload is finished
+		targetDVNs := vmImg.ObjectMeta.Namespace
+		targetDVName := vmImg.ObjectMeta.Name
```
WebberHuang1118: The same as https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855

WebberHuang1118 (commented):
LGTM, generally, most of the comments are similar to those at https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640327855. Considering our schedule, we may defer this to the discussion at https://github.com/harvester/harvester/issues/7524.

In the meantime I'm building my env, there are a few questions:
- How the CR `StorageProfile` is created? and what's the default `CloneStrategy`, `clone` or `snapshot`?
- During a new PVC is created from the golden image with `snapshot` strategy, how the related `snapshotClass` is specified for each storage provider?

Thanks

pkg/image/cdi/download.go
```diff
@@ -1,19 +1,299 @@
...
+		return fmt.Errorf("failed to create the download server pod with VM Image(%s): %w", vmImgName, err)
+	}
+	defer func() {
+		if err := cd.clientSet.CoreV1().Pods(vmImgNamespace).Delete(context.TODO(), podName, metav1.DeleteOptions{}); err != nil {
+			logrus.Errorf("Failed to delete the download server pod(%s): %v", podName, err)
+		}
+	}()
+
+	// wait pod status running
+	if err := wait.PollUntilContextTimeout(context.Background(), tickPolling, tickTimeout, true, func(context.Context) (bool, error) {
```
ibrokethecloud: we may need to make this timeout configurable. currently 60 seconds may not be enough. on a slow environment or where larger images are involved, the conversion may not complete within 60 seconds and eventually the download request times out. should we read this off a setting? 

we may also need to change how the failure is handled as in my case I get a Gateway Timeout error

pkg/image/cdi/cdi.go
Vicente-Cheng: We cannot use vmio to fetchImageVirtualSize here.
At this moment, VM Image does not have any information about virtual size.

We fetch the virtual size here and update it to VM Image.

pkg/image/cdi/cdi.go
Vicente-Cheng: ditto, we don't have any size information here.
Here, we fetch the size from the request and update the VM image.

bk201:
I'm getting ImagePullBackOff in air-gapp env:

```
# pod
│ harvester-system                  harvester-cdi-namespace-modify-444dz                    ●   0/1            0 ImagePullBackOff      0    0      n/a      n/a      n/a      n/a 10.52.0.33    node1    11m     │

# container
│ harvester-cdi-namespace-modify   ●   rancher/kubectl:v1.28.7   false   ImagePullBackOff   false            0 off:off           0     0       0:0       0:0      n/a      n/a      n/a      n/a         11m     │
```

Need to manually include the kubectl image on the installer side.

pkg/image/cdi/upload.go
Vicente-Cheng: use new added update function to update virtual size and size

pkg/image/cdi/cdi.go
Vicente-Cheng: use the new added update function to update both size/virtual size

pkg/image/cdi/common.go
Vicente-Cheng: This is the simple helper function. Could we skip to passing the whole vmimage operator object here for just URL?

pkg/image/cdi/common.go
Vicente-Cheng: re-work the function parameter, now we will pass the sourceType and it comes from vmi operator.

pkg/image/cdi/cdi.go
Vicente-Cheng: I added new function to update both size/virtual size.

Vicente-Cheng:
> I'm getting ImagePullBackOff in air-gapp env:
> 
> ```
> # pod
> │ harvester-system                  harvester-cdi-namespace-modify-444dz                    ●   0/1            0 ImagePullBackOff      0    0      n/a      n/a      n/a      n/a 10.52.0.33    node1    11m     │
> 
> # container
> │ harvester-cdi-namespace-modify   ●   rancher/kubectl:v1.28.7   false   ImagePullBackOff   false            0 off:off           0     0       0:0       0:0      n/a      n/a      n/a      n/a         11m     │
> ```
> 
> Need to manually include the kubectl image on the installer side.

Updated the Value.yaml format to get this image.

pkg/image/cdi/cdi.go
WebberHuang1118: I mean the access to `vmImg.Spec.URL`

WebberHuang1118:
Just a finding if the storage provider is [LVM](https://github.com/harvester/harvester-csi-driver), the `cloneType` will always fallback to `copy`, because the volume expansion is not implemented:
- https://github.com/harvester/csi-driver-lvm/blob/7c60bcb3fe57f0985211b7ab817725d4be083f8c/pkg/lvm/controllerserver.go#L75
- https://github.com/harvester/csi-driver-lvm/blob/7c60bcb3fe57f0985211b7ab817725d4be083f8c/pkg/lvm/controllerserver.go#L524-L526

If using the CSI driver implement volume expansion for example [hostpath](https://github.com/kubernetes-csi/csi-driver-host-path), the `cloneType` will utilize `snapshot` (annotation `cdi.kubevirt.io/cloneType: snapshot`):
  ```
  kind: DataVolume
  metadata:
    annotations:
      cdi.kubevirt.io/cloneType: snapshot
      cdi.kubevirt.io/storage.clone.token: eyJhbGciOiJQUzI1NiJ9.eyJleHAiOjE3NDA2MjkyNjUsImlhdCI6MTc0MDYyODk2NSwiaXNzIjoiY2RpLWFwaXNlcnZlciIsIm5hbWUiOiJpbWFnZS16eHJoYiIsIm5hbWVzcGFjZSI6ImRlZmF1bHQiLCJuYmYiOjE3NDA2Mjg5NjUsIm9wZXJhdGlvbiI6IkNsb25lIiwicGFyYW1zIjp7InRhcmdldE5hbWUiOiJ2bTEtZGlzay0wLXljdTlwIiwidGFyZ2V0TmFtZXNwYWNlIjoiZGVmYXVsdCJ9LCJyZXNvdXJjZSI6eyJncm91cCI6IiIsInJlc291cmNlIjoicGVyc2lzdGVudHZvbHVtZWNsYWltcyIsInZlcnNpb24iOiJ2MSJ9fQ.yTZjIheka8dIMYFTWfjYv_kGOAjYEs_xgNAMwyzMhn7BHEJsDac0b9WpRMYt73FMrCe7XGo0VKC5wFkpNrDcKX2YGDHc5Rw2Ia6t5ytWwCHwb6wrpaFAO7mXfl60F7toyhfSWMK_IgL0rpSVAIekdaAIod32rICAFbrtgqzmq9Fn2YoOGLEN8pFHnFtwHud5UTNBON5QhbZCsD9UTLPHmHUMXno0BGljej2m51XE9URyEi2y2JvnYedPLMdMSN3YZp-90XzASTcceCdZ8iALcX9JQK8jty3X_8GNxufFvq9ut4z8bKNZmbM_W087SexhyAGRSSZcu6hhPCkFCp96AA
      cdi.kubevirt.io/storage.usePopulator: "true"
    name: vm1-disk-0-ycu9p
    namespace: default
    resourceVersion: "29970"
    uid: 17fa5812-f8c2-4bca-912e-99cc45355dd6
  spec:
    source:
      pvc:
        name: image-zxrhb
        namespace: default
    storage:
      accessModes:
      - ReadWriteMany
      resources:
        requests:
          storage: 10Gi
      storageClassName: csi-hostpath-sc
      volumeMode: Block
  status:
    claimName: vm1-disk-0-ycu9p
    conditions:
      ....
    - lastHeartbeatTime: "2025-02-27T04:03:05Z"
      lastTransitionTime: "2025-02-27T04:03:05Z"
      message: Clone Complete
      reason: Completed
      status: "False"
      type: Running
    phase: Succeeded
    progress: 100.0%
  
  ```

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml
ihcsim: Nit: can we use the same description as that in line 173? Thanks.

pkg/api/vm/formatter.go
```diff
@@ -271,6 +274,23 @@ func (vf *vmformatter) canDoBackup(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv
 		return false
 	}
 
+	// additional check, we did not support backup with CDI volume
```
ihcsim, line 277: What happens if user try to use LH (or other CSI providers) to backup the underlying PVC directly?

pkg/controller/master/storageprofile/storage_profile_controller.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	return profile, nil
+}
+
+func (h *storageProfileHandler) generateProfileSpec(sc *v1.StorageClass) cdiv1.StorageProfileSpec {
+	switch sc.Provisioner {
+	case util.CSIProvisionerLonghorn:
+		if sc.Parameters != nil {
+			if _, ok := sc.Parameters["dataEngine"]; ok {
+				if sc.Parameters["dataEngine"] == string(longhornv1.DataEngineTypeV2) {
```
ihcsim:
```suggestion
			if v, ok := sc.Parameters["dataEngine"]; ok {
				if v == string(longhornv1.DataEngineTypeV2) {
```

pkg/controller/master/storageprofile/storage_profile_controller.go
```diff
@@ -0,0 +1,67 @@
...
+
+	targetSCName := *profile.Status.StorageClass
+	sc, err := h.scClient.Get(targetSCName, metav1.GetOptions{})
+	if err != nil {
+		return profile, err
+	}
+	profileSpec := h.generateProfileSpec(sc)
+
+	profileCpy := profile.DeepCopy()
+	profileCpy.Spec = profileSpec
```
ihcsim: If the profile comes with existing properties like `spec.claimPropertySets`, won't we want to preserve them?

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -67,14 +75,53 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
...
+			} else if err != nil {
+				return nil, err
+			}
+			// found dv, try to get pvc for later checking
+			if pvc, err = h.pvcCache.Get(vm.Namespace, pvcAnno.Name); apierrors.IsNotFound(err) {
+				logrus.Warnf("DataVolume %s/%s is created, but the corresponding PVC is not found. Skip this one", pvcAnno.Namespace, pvcAnno.Name)
+				continue
+			} else if err != nil {
+				return nil, err
+			}
```
ihcsim: I think it might not be necessary to have this embedded in the `createPVCWithDataVolume` block, because we can handle the PVC in the next reconciliation loop.

Will something like this work:

```diff
diff --git a/pkg/controller/master/virtualmachine/vm_controller.go b/pkg/controller/master/virtualmachine/vm_controller.go
index 9d497598f..cffcbfc7e 100644
--- a/pkg/controller/master/virtualmachine/vm_controller.go
+++ b/pkg/controller/master/virtualmachine/vm_controller.go
@@ -70,10 +70,6 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
 		return nil, err
 	}
 
-	var (
-		pvc *corev1.PersistentVolumeClaim
-		err error
-	)
 	for _, pvcAnno := range pvcs {
 		imageName := ""
 		imageNS := ""
@@ -96,31 +92,26 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
 
 		// create PVC
 		if createPVCWithDataVolume {
-			if _, err := h.dataVolumeClient.Get(vm.Namespace, pvcAnno.Name, metav1.GetOptions{}); apierrors.IsNotFound(err) {
-				targetSCName := *pvcAnno.Spec.StorageClassName
-				if err := h.createDataVolume(pvcAnno.Name, vm.Namespace, imageName, imageNS, targetSCName, pvcAnno.Spec.Resources); err != nil {
-					return nil, err
+			if _, err := h.dataVolumeClient.Get(vm.Namespace, pvcAnno.Name, metav1.GetOptions{}); err != nil {
+				if apierrors.IsNotFound(err) {
+					targetSCName := *pvcAnno.Spec.StorageClassName
+					if err := h.createDataVolume(pvcAnno.Name, vm.Namespace, imageName, imageNS, targetSCName, pvcAnno.Spec.Resources); err != nil {
+						return nil, err
+					}
+					continue
 				}
-				continue
-			} else if err != nil {
 				return nil, err
 			}
-			// found dv, try to get pvc for later checking
-			if pvc, err = h.pvcCache.Get(vm.Namespace, pvcAnno.Name); apierrors.IsNotFound(err) {
-				logrus.Warnf("DataVolume %s/%s is created, but the corresponding PVC is not found. Skip this one", pvcAnno.Namespace, pvcAnno.Name)
-				continue
-			} else if err != nil {
-				return nil, err
-			}
-		} else {
-			pvcAnno.Namespace = vm.Namespace
-			if pvc, err = h.pvcCache.Get(vm.Namespace, pvcAnno.Name); apierrors.IsNotFound(err) {
+		}
+
+		pvcAnno.Namespace = vm.Namespace
+		pvc, err := h.pvcCache.Get(vm.Namespace, pvcAnno.Name)
+		if err != nil {
+			if apierrors.IsNotFound(err) && !createPVCWithDataVolume {
 				if _, err = h.pvcClient.Create(pvcAnno); err != nil {
 					return nil, err
 				}
 				continue
-			} else if err != nil {
-				return nil, err
 			}
 		}
 
```

pkg/image/cdi/cdi.go
```diff
@@ -1,32 +1,263 @@
...
+	"github.com/harvester/harvester/pkg/image/common"
 )
 
 type Backend struct {
+	ctx              context.Context
+	dataVolumeClient ctlcdiv1.DataVolumeClient
+	pvcCache         ctlcorev1.PersistentVolumeClaimCache
+	vmio             common.VMIOperator
+
+	dataVolumeCreated bool
```
ihcsim: Is this used anywhere else?

```
✗ grep -rnI "dataVolumeCreated" pkg
pkg/image/cdi/cdi.go:27:        dataVolumeCreated bool
pkg/image/cdi/cdi.go:36:                dataVolumeCreated: false,
```

pkg/image/cdi/download.go
```diff
@@ -1,19 +1,299 @@
...
+		clientSet:  clientSet,
+		vmio:       vmio,
+	}
 }
 
-func (cd *Downloader) Do(_ *harvesterv1.VirtualMachineImage, _ http.ResponseWriter, _ *http.Request) error {
+func (cd *Downloader) Do(vmImg *harvesterv1.VirtualMachineImage, rw http.ResponseWriter, req *http.Request) error {
+	vmImgName := cd.vmio.GetName(vmImg)
+	vmImgNamespace := cd.vmio.GetNamespace(vmImg)
+	podName := fmt.Sprintf("%s-%s-%s", vmImgNamespace, vmImgName, getRandomSuffix())
```
ihcsim: Instead of doing this ourselves, can we let K8s generate the name suffix by using [`ObjectMeta.GenerateName`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta) inside `cd.generateDownloadServerPod()`?

pkg/image/cdi/download.go
```diff
@@ -1,19 +1,299 @@
...
+			Containers: []corev1.Container{
+				{
+					Name:            "image-exporter",
+					Image:           clusterRepoImageStr,
+					ImagePullPolicy: corev1.PullIfNotPresent,
+					Ports: []corev1.ContainerPort{
+						{
+							ContainerPort: 80,
+							Protocol:      corev1.ProtocolTCP,
+						},
```
ihcsim: We might need a separate ticket to discuss if this needs to be secured. E.g., using K8s [`TokenReview` API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/) to make sure only the controller's service account can invoke this endpoint.

ihcsim (commented):
I got as far as the uploader today. Some comments and questions below. Thanks.

pkg/image/cdi/download.go
```diff
@@ -1,19 +1,299 @@
...
+	}
+
+	if _, err := io.Copy(rw, downloadResp.Body); err != nil {
+		return fmt.Errorf("failed to copy download content to target(%s), err: %w", targetFileName, err)
+	}
+
 	return nil
 }
+
+func (cd *Downloader) generateDownlodServerPod(podName, vmImageName string) *corev1.Pod {
```
ihcsim:
```suggestion
func (cd *Downloader) generateDownloadServerPod(podName, vmImageName string) *corev1.Pod {
```

pkg/image/cdi/download.go
```diff
@@ -1,19 +1,299 @@
...
+	vmImgName := cd.vmio.GetName(vmImg)
+	vmImgNamespace := cd.vmio.GetNamespace(vmImg)
+	podName := fmt.Sprintf("%s-%s-%s", vmImgNamespace, vmImgName, getRandomSuffix())
+
+	podTemplate := cd.generateDownlodServerPod(podName, vmImgName)
+	if _, err := cd.clientSet.CoreV1().Pods(vmImgNamespace).Create(context.TODO(), podTemplate, metav1.CreateOptions{}); err != nil {
+		return fmt.Errorf("failed to create the download server pod with VM Image(%s): %w", vmImgName, err)
+	}
+	defer func() {
+		if err := cd.clientSet.CoreV1().Pods(vmImgNamespace).Delete(context.TODO(), podName, metav1.DeleteOptions{}); err != nil {
```
ihcsim: Shall we delete the `Service` too?

pkg/image/cdi/download.go
```diff
@@ -1,19 +1,299 @@
...
+			logrus.Errorf("Failed to generate random number: %v", err)
+			return "qwert"
+		}
+		result.WriteByte(charset[randIdx])
+	}
+
+	return result.String()
+}
+
+func (cd *Downloader) waitPodRunning(podName, podNs string) (bool, error) {
```
ihcsim: Nit: technically, this is not a `wait` method because the `wait.PollUntilContextTimeout()` is called outside of this method.

pkg/image/cdi/upload.go
```diff
@@ -1,20 +1,334 @@
...
+	}
+
+	// create DataVolume
+	dvName := cu.vmio.GetName(vmImg)
+	dvNamespace := cu.vmio.GetNamespace(vmImg)
+
+	logrus.Infof("The VM Image Status updated, start to create DataVolume %s/%s", dvNamespace, dvName)
+
+	// generate DV source
+	dvSource, err := generateDVSource(vmImg, cu.vmio.GetSourceType(vmImg))
```
ihcsim, line 154: To confirm, this will always be upload source type, right?

pkg/image/cdi/upload.go
```diff
@@ -1,20 +1,334 @@
...
+	// wait data volume UploadReady
+	if err := wait.PollUntilContextTimeout(context.Background(), tickPolling, tickTimeout, true, func(context.Context) (bool, error) {
+		return cu.waitDataVolumeStatus(dvNamespace, dvName, cdiv1.UploadReady)
+	}); err != nil {
+		return fmt.Errorf("failed to wait for VMImage status: %v", err)
+	}
+
+	uploadTokenRequest := &uploadcdiv1.UploadTokenRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "token-for-upload",
```
ihcsim: We should use `GenerateName` to avoid collisions.

pkg/image/cdi/upload.go
```diff
@@ -1,20 +1,334 @@
...
+			Namespace: dvNamespace,
+		},
+		Spec: uploadcdiv1.UploadTokenRequestSpec{
+			PvcName: dvName,
+		},
+	}
+	retUploadTokenRequest, err := cu.cdiUploadClient.Create(uploadTokenRequest)
```
ihcsim, line 205: Does this get deleted later? We shouldn't leave it laying around after use.

pkg/image/cdi/upload.go
```diff
@@ -1,20 +1,334 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create UploadTokenRequest %s/%s: %v", dvNamespace, dvName, err)
+	}
+
+	newBody := io.MultiReader(bytes.NewReader(dataContent), req.Body)
```
ihcsim, line 210: IIUC, `dataContent` is the first N bytes of `req.Body` (multipart or not). Won't this `io.MultiReader()` end up reading redundant data from `req.Body`?

pkg/image/cdi/upload.go
```diff
@@ -1,20 +1,334 @@
...
+	cdiUploadURL := fmt.Sprintf("https://%s%s", CDIUploadURLRaw, UploadProxyURI)
+	uploadReq, err := http.NewRequestWithContext(req.Context(), http.MethodPost, cdiUploadURL, io.NopCloser(hookedReader))
+	if err != nil {
+		return fmt.Errorf("failed to wrap the upload request: %w", err)
+	}
+	uploadReq.Header = req.Header
+	uploadReq.Header.Add("Authorization", "Bearer "+token)
+
+	// create VMI progress updater
+	go cu.updateVMImageProgress(vmImg, updaterCond, progress, fileSize)
```
ihcsim, line 230: Although we start the goroutine here, it will remain in the waiting state until the call to `Signal()` at line 261, right? 

1. If any errors happened before line 261, won't this goroutine remain stuck indefinitely?
1. Instead of locking and waiting for `Signal()` inside the goroutine, can we just let the `for` loop iterates in the background with N seconds interval?

pkg/image/cdi/cdi.go
Vicente-Cheng: Updated

pkg/image/cdi/download.go
Vicente-Cheng: I rewrite the download mechanism.
Now, we will use the CRD to ensure the download server is ready. Then, the HTTP request will be executed after that.

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachineimages.yaml
Vicente-Cheng: Updated

pkg/api/vm/formatter.go
Vicente-Cheng: I am not sure what LH (or other CSI providers) means to back up.

The snapshot is the formal type for other CSI provisioners. (Backup is not the formal CSI interface)
LH provides the backup, but this is the special handling, so we cannot use that for other CSI provisioners.

Or could you elaborate more?

pkg/image/cdi/download.go
Vicente-Cheng: I rewrite the download part, so please recheck the new mechanism for the download.

pkg/image/cdi/download.go
Vicente-Cheng: ditto

pkg/image/cdi/download.go
Vicente-Cheng: ditto

pkg/image/cdi/download.go
Vicente-Cheng: ditto

pkg/image/cdi/download.go
Vicente-Cheng: ditto

pkg/image/cdi/upload.go
Vicente-Cheng: Yes, this is always the upload source.

pkg/image/cdi/upload.go
Vicente-Cheng: updated

pkg/image/cdi/upload.go
Vicente-Cheng: No, it will wrap these two contents again for later io reader.

pkg/image/cdi/upload.go
Vicente-Cheng: Thanks for the catching.
I added the defer func to handle the early error case.

BTW, we have another Single on the io writer (to trigger progress update every 1 seconds)

pkg/image/cdi/download.go
Vicente-Cheng: With the new mechanism for the downloader, the wrong repo name will not make the corresponding CRD become ready. So we will know what happened.

pkg/image/cdi/common.go
WebberHuang1118: I'm okay with that

pkg/image/cdi/download.go
Vicente-Cheng: Yeah, I thought you were correct. We could create another ticket to discuss authorization.

pkg/image/cdi/cdi.go
Vicente-Cheng: updated

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: Thanks for the suggestions. I rewrote this part to simplify the code logic here.

pkg/controller/master/storageprofile/storage_profile_controller.go
Vicente-Cheng: I enhanced the storage profile controller. Now, we only patch the `CloneStrategy` and `SnapshotClass`

pkg/controller/master/storageprofile/storage_profile_controller.go
Vicente-Cheng: updated

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: I rewrote this part. Please help to recheck it.

pkg/image/cdi/upload.go
Vicente-Cheng: I do some investigation, and thanks @bk201 for helping with that.
The uploadTokenRequest is not the usual CRD. It is used with API Aggregation Layer (REF: https://kubernetes.io/blog/2024/11/21/dynamic-kubernetes-api-server-for-cozystack/)

This is one-time resource (https://kubernetes.io/blog/2024/11/21/dynamic-kubernetes-api-server-for-cozystack/#one-time-resources)
The apiserver will handle the create (POST) and return the created CRD to user.

back to the UploadTokenRequest. The apiserver will hold the private key. When user wants to create the UploadTokenRequest, it will use this private key to generate the corresponding token. Then, validate when the user wants to upload.

TL;DR, this is not the usual CRD, so we could skip the delete after we create it.

pkg/controller/master/vmimagedownloader/vmimagedownloader_controller.go
```diff
@@ -0,0 +1,302 @@
...
+			Message:            "Waiting for the corresponding deployment to be ready",
+		}
+		updatedConditions := UpdateConds(downloaderCpy.Status.Conditions, conds)
+		downloaderCpy.Status.Conditions = updatedConditions
+		downloaderCpy.Status.Status = harvesterv1.ImageDownloaderStatusProgressing
+		if !reflect.DeepEqual(downloader, downloaderCpy) {
+			return h.vmImageDownloaders.UpdateStatus(downloaderCpy)
+		}
+		return downloader, fmt.Errorf("Deployment %s is not ready, wanted replicas: %d, ready replicas: %d", deployment.Name, deployment.Spec.Replicas, deployment.Status.ReadyReplicas)
+	}
```
ihcsim: Is this necessary? Feels like we can just create the `Service`, and then just call `h.vmImageDownloaders.UpdateStatus(downloaderCpy)` once at line 125. If anything, I think we should send a `GET` to port 80, to confirm it's live before calling line 125.

pkg/controller/master/vmimagedownloader/vmimagedownloader_controller.go
```diff
@@ -0,0 +1,302 @@
...
+	utilHelm "github.com/harvester/harvester/pkg/util/helm"
+)
+
+var (
+	kindDeployment = "Deployment"
+	boolTrue       = true
+)
+
+// storageProfileHandler dynamically manages storage profiles
+type vmImageDonwloaderHandler struct {
```
ihcsim:
```suggestion
type vmImageDownloaderHandler struct {
```

pkg/webhook/resources/datavolume/validator.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func dataVolumeIsAttachedToVM(vm *kubevirtv1.VirtualMachine, curDV *cdiv1.DataVolume) bool {
+	volumeClaimTemplates, ok := vm.Annotations[util.AnnotationVolumeClaimTemplates]
+	if !ok || volumeClaimTemplates == "" {
+		return false
+	}
+	var pvcs []*corev1.PersistentVolumeClaim
+	if err := json.Unmarshal([]byte(volumeClaimTemplates), &pvcs); err != nil {
+		logrus.Warnf("failed to unmarshal volumeClaimTemplates: %s, assume the dataVOlume is in use", err)
```
ihcsim:
```suggestion
		logrus.Warnf("failed to unmarshal volumeClaimTemplates: %s, assume the dataVolume is in use", err)
```

ihcsim (approved):
@Vicente-Cheng Overall, LGTM - thanks! I think we should get this PR merged soon to get QA started. We can handle any subsequent changes or bug fixes in smaller follow-up PRs, instead of churning through this big one.

pkg/controller/master/vmimagedownloader/vmimagedownloader_controller.go
Vicente-Cheng: updated

pkg/webhook/resources/datavolume/validator.go
Vicente-Cheng: updated

pkg/controller/master/vmimagedownloader/vmimagedownloader_controller.go
Vicente-Cheng: Yes, we need to update the `status` to `Progressing` here to let UI know the server is not ready.
Then, update the `status` to `Ready` once the server is ready.
That UI could depend on this status to execute the download link

Vicente-Cheng:
rebased to the latest

ibrokethecloud (approved):
lgtm. thanks for the enhacement.

bk201 (approved):
LGTM. Verified with NFS and Rook csi.

WebberHuang1118 (approved):
LGTM, checked with hostpath-csi.

Please help to answer it https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640371916 when you have time, thanks.

Vicente-Cheng:
> LGTM, checked with hostpath-csi.
> 
> Please help to answer it [#7640 (review)](https://github.com/harvester/harvester/pull/7640#pullrequestreview-2640371916) when you have time, thanks.

Thanks, I will add that info on the questions to HEP again.

Issue #7652, opened 2025-02-18
[BUG] Enhance upgradeLog to adapt with rancher-logging stack bumping


w13915984028:
**Describe the bug**

When testing PR https://github.com/harvester/harvester/pull/7621 for https://github.com/harvester/harvester/issues/7404, observe the upgradeLog also needs some enhancements:

**To Reproduce**
1. Upgrade v141 to v150 master-head

(1) The loggingRef field is a must, need to be added
(2) Add processing on upgrade path

```
$kubectl get logging -A
NAME                                  LOGGINGREF                     CONTROLNAMESPACE        WATCHNAMESPACES   PROBLEMS
hvst-upgrade-d6kpr-upgradelog-infra                                  harvester-system                          
rancher-logging-kube-audit            harvester-kube-audit-log-ref   cattle-logging-system                     
rancher-logging-root                                                 cattle-logging-system                     

$kubectl get logging -A -oyaml
apiVersion: v1
items:
- apiVersion: logging.banzaicloud.io/v1beta1
  kind: Logging
  metadata:
    creationTimestamp: "2025-02-18T19:39:47Z"
    generation: 1
    labels:
      harvesterhci.io/upgradeLog: hvst-upgrade-d6kpr-upgradelog
      harvesterhci.io/upgradeLogComponent: infra
    name: hvst-upgrade-d6kpr-upgradelog-infra
    ownerReferences:
    - apiVersion: harvesterhci.io/v1beta1
      kind: UpgradeLog
      name: hvst-upgrade-d6kpr-upgradelog
      uid: 4699a2eb-230a-4b5e-b589-749f17163c5a
    resourceVersion: "45315"
    uid: 98cfc653-9be1-4a3d-a76c-d697fa882da0
  spec:
    controlNamespace: harvester-system
    flowConfigCheckDisabled: true
    fluentbit:
      bufferStorage: {}
      bufferStorageVolume: {}
      bufferVolumeImage: {}
      filterKubernetes: {}
      image:
        repository: rancher/mirrored-fluent-fluent-bit
        tag: 2.2.0
      inputTail: {}
      labels:
        harvesterhci.io/upgradeLog: hvst-upgrade-d6kpr-upgradelog
        harvesterhci.io/upgradeLogComponent: shipper
      positiondb: {}
      resources: {}
      updateStrategy: {}
    fluentd:
      bufferStorageVolume: {}
      bufferVolumeImage: {}
      configCheckResources: {}
      configReloaderImage:
[...]

Issue #7404, opened 2025-01-20
[ENHANCEMENT] v1.5.0 Bump monitoring and logging


w13915984028:
**Is your enhancement related to a problem? Please describe.**

Bump monitoring & logging charts

Monitoring from 103.1.1+up45.31.1 to 105.1.2+up61.3.2
Logging from 103.1.0+up4.4.0 to 105.2.0+up4.10.0

**Describe the solution you'd like**

1. Bump charts
2. Review changes and potential adaption
3. Add upgrade processing
4. Tag new image of eventrouter
5. Add a fix for https://github.com/harvester/harvester/issues/7064 and consider to add a init contaier to check

PRs & issues:

- [x] installer https://github.com/harvester/harvester-installer/pull/980, 
- [x] upgrade: https://github.com/harvester/harvester/pull/7621, 
- [x] addon: https://github.com/harvester/addons/pull/47
- [x] eventrouter new tag https://github.com/harvester/eventrouter/releases/tag/v1.5.0-dev.0
- [x] upgradeLog, bug: https://github.com/harvester/harvester/issues/7652, needs enhancement
- [x] upgradeLog, bug: https://github.com/harvester/harvester/issues/7654, needs enhancement
- [x] upgradeLog fix PR: https://github.com/harvester/harvester/pull/7653
- [x] upgradeLog fix PR, on upgrade path: https://github.com/harvester/harvester/pull/7660
- [x] upgradeLog fix PR https://github.com/harvester/harvester/pull/7772, additional fix to bump vendor and adpat
- [x] code placement PR https://github.com/harvester/harvester/pull/7862
- [x] document PR https://github.com/harvester/docs/pull/744

**Describe alternatives you've considered**

**Additional context**

other issues observed when coding and testing:
- [x] https://github.com/harvester/harvester/issues/7891
- [x] https://github.com/harvester/harvester/pull/7892
- [x] https://github.com/harvester/harvester/issues/7895

w13915984028:
Test plan: per `test plan` on PR: https://github.com/harvester/harvester-installer/pull/980#issue-2851253070

w13915984028:
FYI: Wait a last PR (https://github.com/harvester/harvester/pull/7772) to address issues observed when testing PR https://github.com/harvester/harvester/pull/7660 and a more commit per https://github.com/harvester/harvester/pull/7772, then this issue can be moved to test stage.

[...]

PR #7653, opened 2025-02-18
Enhance upgradeLog

w13915984028:

**Problem:**

https://github.com/harvester/harvester/issues/7652
https://github.com/harvester/harvester/issues/7654

The controller checks `upgradeLog.DeletionTimestamp != nil`

https://github.com/harvester/harvester/blob/9b46493d505dfc4e9d0402ad57de43752dfcc9f8/pkg/controller/master/upgradelog/controller.go#L94

The `stopCollect` has no chance to be called here.

https://github.com/harvester/harvester/blob/9b46493d505dfc4e9d0402ad57de43752dfcc9f8/pkg/controller/master/upgradelog/controller.go#L298

The onRemove needs to call `stopCollect` anyway. (or check the related conditions first)

https://github.com/harvester/harvester/blob/9b46493d505dfc4e9d0402ad57de43752dfcc9f8/pkg/controller/master/upgradelog/controller.go#L317

**Solution:**

(1) Add the required loggingRef field
(2) Cleanup resources when upgrade is manually deleted

**Related Issue:**
https://github.com/harvester/harvester/issues/7652
https://github.com/harvester/harvester/issues/7654

**Test plan:**

1. without enabling rancher-logging,  trigger upgrade, enable upgradelog, delete upgrade after a while

All related resources are cleand up.

```
$kubectl get logging -A
NAME                                          LOGGINGREF             CONTROLNAMESPACE        WATCHNAMESPACES   PROBLEMS
hvst-upgrade-cjhgj-upgradelog-infra           harvester-upgradelog   harvester-system                          
hvst-upgrade-cjhgj-upgradelog-operator-root                          cattle-logging-system                     

...
$kubectl get clusterflow -A
No resources found
$kubectl  get clusteroutput -A
No resources found
$kubectl kk get logging -A
No resources found

$kubectl  get managedchart -A
NAMESPACE     NAME                                      AGE
fleet-local   harvester                                 28m
fleet-local   harvester-crd                             28m
fleet-local   local-managed-system-upgrade-controller   28m
fleet-local   rancher-logging-crd                       28m
fleet-local   rancher-monitoring-crd                    28m


...
time="2025-02-19T09:26:12Z" level=info msg="Update upgradeLog harvester-system/hvst-upgrade-cjhgj-upgradelog"
time="2025-02-19T09:26:12Z" level=info msg="Stop collecting logs"
time="2025-02-19T09:26:12Z" level=info msg="Tearing down the logging infrastructure for upgrade procedure"
time="2025-02-19T09:26:12Z" level=info msg="Delete UpgradeLog harvester-system/hvst-upgrade-cjhgj-upgradelog"
time="2025-02-19T09:26:12Z" level=info msg="Removing all other related resources"
time="2025-02-19T09:26:12Z" level=info msg="Tearing down the logging infrastructure for upgrade procedure"
```

2. with enabling rancher-logging,  trigger upgrade, enable upgradelog, delete upgrade after a while

No resources are left

```
...
$kubectl  get logging -A
NAME                                  LOGGINGREF                     CONTROLNAMESPACE        WATCHNAMESPACES   PROBLEMS
hvst-upgrade-9xjkb-upgradelog-infra   harvester-upgradelog           harvester-system                          
rancher-logging-kube-audit            harvester-kube-audit-log-ref   cattle-logging-system                     
rancher-logging-root                                                 cattle-logging-system                     


$kubectl get logging -A
NAME                         LOGGINGREF                     CONTROLNAMESPACE        WATCHNAMESPACES   PROBLEMS
rancher-logging-kube-audit   harvester-kube-audit-log-ref   cattle-logging-system                     
rancher-logging-root                                        cattle-logging-system                     
...
```

### DIFF
--- a/pkg/controller/master/upgradelog/common.go
+++ b/pkg/controller/master/upgradelog/common.go
@@ -25,6 +25,8 @@ import (
 const (
 	defaultDeploymentReplicas   int32 = 1
 	defaultLogArchiveVolumeSize       = "1Gi"
+
+	upgradeLogLoggingRef = "harvester-upgradelog"
 )
 
 func upgradeLogReference(upgradeLog *harvesterv1.UpgradeLog) metav1.OwnerReference {
@@ -84,6 +86,8 @@ func prepareLogging(upgradeLog *harvesterv1.UpgradeLog, images map[string]Image)
 			},
 		},
 		Spec: loggingv1.LoggingSpec{
+			// without this field, it may cause: "Other logging resources exist with the same loggingRef: rancher-logging-root"
+			LoggingRef:              upgradeLogLoggingRef,
 			ControlNamespace:        upgradeLog.Namespace,
 			FlowConfigCheckDisabled: true,
 			FluentbitSpec: &loggingv1.FluentbitSpec{
--- a/pkg/controller/master/upgradelog/common.go
+++ b/pkg/controller/master/upgradelog/common.go
@@ -26,6 +26,8 @@ const (
 	defaultDeploymentReplicas   int32 = 1
 	defaultLogArchiveVolumeSize       = "1Gi"
 
+	// this is used to differentiate separate fluentbit&fluentd group
+	// all the none-root logging/clusterflow/clusteroutput objects need this reference
 	upgradeLogLoggingRef = "harvester-upgradelog"
 )
 
@@ -170,6 +172,7 @@ func prepareClusterFlow(upgradeLog *harvesterv1.UpgradeLog) *loggingv1.ClusterFl
 			},
 		},
 		Spec: loggingv1.ClusterFlowSpec{
+			LoggingRef: upgradeLogLoggingRef,
 			Filters: []loggingv1.Filter{
 				{
 					TagNormaliser: &filter.TagNormaliser{},
@@ -264,6 +267,7 @@ func prepareClusterOutput(upgradeLog *harvesterv1.UpgradeLog) *loggingv1.Cluster
 		},
 		Spec: loggingv1.ClusterOutputSpec{
 			OutputSpec: loggingv1.OutputSpec{
+				LoggingRef: upgradeLogLoggingRef,
 				FileOutput: &output.FileOutputConfig{
 					Path:     "/archive/logs/${tag}",
 					Compress: "gzip",
--- a/pkg/controller/master/upgradelog/controller.go
+++ b/pkg/controller/master/upgradelog/controller.go
@@ -2,6 +2,7 @@ package upgradelog
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	loggingv1 "github.com/kube-logging/logging-operator/pkg/sdk/logging/api/v1beta1"
@@ -750,19 +751,20 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
 	var err error
 	err = h.clusterFlowClient.Delete(util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogFlowComponent), &metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("failed to delete clusterflow %s/%s", util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogFlowComponent))
 	}
 	err = h.clusterOutputClient.Delete(util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogOutputComponent), &metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("failed to delete clusteroutput %s/%s", util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogOutputComponent))
 	}
 	err = h.loggingClient.Delete(name.SafeConcatName(upgradeLog.Name, util.UpgradeLogInfraComponent), &metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("failed to delete logging client %s", name.SafeConcatName(upgradeLog.Name, util.UpgradeLogInfraComponent))
 	}
+	// new to wait a bit, let logging operator finish the task
 	err = h.managedChartClient.Delete(util.FleetLocalNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogOperatorComponent), &metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("failed to delete logging managedchart %s/%s", util.FleetLocalNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogOperatorComponent))
 	}
 
 	return nil
@@ -771,24 +773,48 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
 func (h *handler) cleanup(upgradeLog *harvesterv1.UpgradeLog) error {
 	// Cleanup the relationship from its corresponding Upgrade resource
 	upgradeName := upgradeLog.Spec.UpgradeName
-	upgrade, err := h.upgradeCache.Get(util.HarvesterSystemNamespaceName, upgradeName)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil
+	var err1, err2, err3 error
+	errorStr := fmt.Sprintf("upgradeLog %s/%s upgrade %s cleanup errors", upgradeLog.Namespace, upgradeLog.Name, upgradeName)
+
+	upgrade, err1 := h.upgradeCache.Get(util.HarvesterSystemNamespaceName, upgradeName)
+	for i := 0; i < 1; i++ {
+		if err1 != nil {
+			// can't return directly, need to clean other related resources
+			if apierrors.IsNotFound(err1) {
+				err1 = nil
+				break
+			}
+			errorStr = fmt.Sprintf("%s failed to get upgrade %s", errorStr, err1.Error())
+			break
 		}
-		return err
-	}
-	upgradeToUpdate := upgrade.DeepCopy()
-	upgradeToUpdate.Status.UpgradeLog = ""
-	if _, err = h.upgradeClient.Update(upgradeToUpdate); err != nil {
-		return err
+		// already updated
+		if upgrade.Status.UpgradeLog == "" {
+			break
+		}
+		upgradeToUpdate := upgrade.DeepCopy()
+		upgradeToUpdate.Status.UpgradeLog = ""
+		_, err2 = h.upgradeClient.Update(upgradeToUpdate)
+		if err2 != nil {
+			if apierrors.IsNotFound(err2) {
+				err2 = nil
+				break
+			}
+			errorStr = fmt.Sprintf("%s failed to update upgrade %s", errorStr, err2.Error())
+			break
+		}
 	}
 
-	// Remove the ManagedChart if the UpgradeLog resource is deleted before normal tear down
-	logrus.Info("Removing logging-operator ManagedChart if any")
-	err = h.managedChartClient.Delete(util.FleetLocalNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogOperatorComponent), &metav1.DeleteOptions{})
-	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+	// Remove all if the UpgradeLog resource is deleted before normal tear down
+	logrus.Info("Removing all other related resources")
+	err3 = h.stopCollect(upgradeLog)
+	if err3 != nil {
+		errorStr = fmt.Sprintf("%s failed to clean other resources %s", errorStr, err3.Error())
+	}
+
+	if err1 != nil || err2 != nil || err3 != nil {
+		// retry
+		logrus.Infof("Clean up upgradeLog related resources with error %s, retry", errorStr)
+		return fmt.Errorf("%s", errorStr)
 	}
 
 	return nil
--- a/pkg/controller/master/upgradelog/controller.go
+++ b/pkg/controller/master/upgradelog/controller.go
@@ -751,31 +751,34 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
 	var err error
 	err = h.clusterFlowClient.Delete(util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogFlowComponent), &metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		return fmt.Errorf("failed to delete clusterflow %s/%s", util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogFlowComponent))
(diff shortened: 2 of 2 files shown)


pkg/controller/master/upgradelog/controller.go
```diff
@@ -771,24 +773,48 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
...
-		return err
 	}
-	upgradeToUpdate := upgrade.DeepCopy()
-	upgradeToUpdate.Status.UpgradeLog = ""
-	if _, err = h.upgradeClient.Update(upgradeToUpdate); err != nil {
-		return err
+
+	// Remove all if the UpgradeLog resource is deleted before normal tear down
+	logrus.Info("Removing all other related resources")
+	err3 = h.stopCollect(upgradeLog)
```
w13915984028, line 811: @starbops please take a quick look of this on-going fix;  in general, if upgrade is deleted earlier, the upgradelog may leave many resources; check issue https://github.com/harvester/harvester/issues/7654

w13915984028:
@bk201 @starbops @tserong 

This PR is also needed on v1.4.2

For the v1.4.*->v1.5.0 upgrade path, I will add another PR to patch the upgradeLog resources with loggingRef.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -771,24 +773,48 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
...
 	upgradeName := upgradeLog.Spec.UpgradeName
-	upgrade, err := h.upgradeCache.Get(util.HarvesterSystemNamespaceName, upgradeName)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil
+	var err1, err2, err3 error
+	errorStr := fmt.Sprintf("upgradeLog %s/%s upgrade %s cleanup errors", upgradeLog.Namespace, upgradeLog.Name, upgradeName)
+
+	upgrade, err1 := h.upgradeCache.Get(util.HarvesterSystemNamespaceName, upgradeName)
+	for i := 0; i < 1; i++ {
```
m-ildefons, line 782: This loop runs once. Consider breaking the contents out in their own function.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -771,24 +773,48 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
...
-	upgradeToUpdate := upgrade.DeepCopy()
-	upgradeToUpdate.Status.UpgradeLog = ""
-	if _, err = h.upgradeClient.Update(upgradeToUpdate); err != nil {
-		return err
+
+	// Remove all if the UpgradeLog resource is deleted before normal tear down
+	logrus.Info("Removing all other related resources")
+	err3 = h.stopCollect(upgradeLog)
+	if err3 != nil {
+		errorStr = fmt.Sprintf("%s failed to clean other resources %s", errorStr, err3.Error())
```
m-ildefons: Consider wrapping the errors with `fmt.Errorf("%w", err)` rather than accumulating error strings that can not be checked for individual errors later.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -750,19 +751,20 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
 	var err error
 	err = h.clusterFlowClient.Delete(util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogFlowComponent), &metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("failed to delete clusterflow %s/%s", util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogFlowComponent))
```
m-ildefons: Consider propagating the actual error message up instead of swallowing it. In case there is an error having the root error in the log may help diagnosis.

pkg/controller/master/upgradelog/common.go
```diff
@@ -84,6 +86,8 @@ func prepareLogging(upgradeLog *harvesterv1.UpgradeLog, images map[string]Image)
 			},
 		},
 		Spec: loggingv1.LoggingSpec{
+			// without this field, it may cause: "Other logging resources exist with the same loggingRef: rancher-logging-root"
+			LoggingRef:              upgradeLogLoggingRef,
```
m-ildefons, line 92: Not sure if it's a good idea to use a constant name when the `loggingRef` must be unique.
In case UpgradeLog objects are leaked after an aborted upgrade or when the upgrade errors out, this could lead to the above EEXISTS problem.
We could use a generated unique name `harvester-upgradelog-XXXXXX` to avoid this.

pkg/controller/master/upgradelog/controller.go
w13915984028: This clean up function is special, it can't return on error simply, but need to try releasing resource as more as possible.

That's why we try to put those errors together.

pkg/controller/master/upgradelog/controller.go
w13915984028: o, the error is missing, needs to be added back

pkg/controller/master/upgradelog/common.go
w13915984028: a fixed name is needed, which can help trigger error when upgradeLog leaks resources, that requires a manual cleaning up

pkg/controller/master/upgradelog/controller.go
w13915984028: the for loop is to help reducing the nested `if else`, as it can't return quickly

m-ildefons: So this is a complicated way to do a goto?
https://go.dev/ref/spec#Goto_statements

pkg/controller/master/upgradelog/controller.go
m-ildefons: Yes, I understand, but the individual errors are swallowed this way.
By using `fmt.Sprintf("... %w", err)` you can create an error that still matches on the original error (e.g. when you later to a `errors.IsNotFound(err)` from the caller of this function).

pkg/controller/master/upgradelog/controller.go
w13915984028: similar, just for is more preferred personally here

pkg/controller/master/upgradelog/controller.go
w13915984028: Good idea, updated.

pkg/controller/master/upgradelog/controller.go
starbops: Do you think adding a finalizer to the Upgrade resource can prevent such a situation? In that case, the `cleanup` function can always find the Upgrade CR through UpgradeLog CR's `.spec.upgradeName`?
https://github.com/harvester/harvester/blob/a7e483894b31e0adccd71cb54f044ac5124d49cd/pkg/controller/master/upgradelog/controller.go#L773-L780

w13915984028: There are following considerations:

(1) When `upgradeLog` has been set the deletionTimestamp, no matter it can find `upgrade` or not, it needs to clean all sub resource as more as possible

(2) In existing code, it only cleans managedchart, but `clusterflow/clusteroutput ...`, seems are not cleaned like `stopCollect`
https://github.com/harvester/harvester/blob/c675476bf6704bc310b16fce64d875c0d6275237/pkg/controller/master/upgradelog/controller.go#L787

(3) In `cleanup`, the dependency with `upgrade` is to update `upgradeToUpdate.Status.UpgradeLog`, this makes little sense on a being deleted object

upgrade controller returns when it sees this stamp, please let me know if there are special processing on it

https://github.com/harvester/harvester/blob/c675476bf6704bc310b16fce64d875c0d6275237/pkg/controller/master/upgrade/upgrade_controller.go#L121

(4) With a `finalizer`, it means a tight coupling, we need to check the lifecycle management of both objects, to make sure all cases are covered, avoid to block the deletion of `upgrade` accidentally

(5) The upgraded object from v14*, won't have a `finalizer`, unless we patch it on the upgrade path

the `finalizer` is hard to solve all above

starbops: Thanks for the breakdown. It makes sense. I'll review the PR later.

w13915984028:
FYI:

If `rancher-logging` is not enabled, upgradeLog will create a managedchart, it looks also essential to add the `loggingRef`, I am checking if the source code can this.

In the upgrade path, it is done https://github.com/harvester/harvester/pull/7660/files#diff-cca301b3921d99450abbcd828d89bf81b8c0b7112d1b941efe0bf3d62e234dcfR615

tserong:
> This PR is also needed on v1.4.2

If this is needed on v1.4.2, please add a "backport-needed/1.4.2" label to the related issue. This should automatically make a matching issue against the v1.4.2 milestone, which will help with tracking/testing this for that release.

pkg/controller/master/upgradelog/controller.go
m-ildefons: Would it be an option to work with [owners and dependents](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) here, i.e. make the UpgradeLog object the owner of the subresources? And make the Upgrade resource owner of the UpgradeLog?
This way, cleanup of the dependents would be handled automatically.

w13915984028: It also becomes a bit more complex to test all at the moment, the upgradeLog utilizes the `managedchart` (if addon is not enabled) and the `logging-operator` to rollout `operator` &/ `fluentbit`,`fluentd`, the owner chain management and the upgrade path pathching will let this be a `refactor` than a `fix`.

starbops: The Upgrade and UpgradeLog CRs already have ownership set. However, some of the UpgradeLog CR's sub-resources are not set with an owner reference because they're in different namespaces. IIRC, the `.metadata.ownerReference` does not support cross-namespace references. The only way to clean up the mess is to do the cleanup by explicit deletion.

w13915984028:
After discussion with @bk201, this PR is not supposed to be included in v1.4.2, there is a workaround to solve the occured issue, thanks. cc @tserong 
thanks.

starbops (commented):
I feel the newly proposed change for the resource cleanup part is a bit complicated. Do you think moving the resource deletion part before the Upgrade CR update block can resolve the dangling resource issue?

For example:
```go
func (h *handler) cleanup(upgradeLog *harvesterv1.UpgradeLog) error {
	// Remove the Logging resource if the UpgradeLog resource is deleted before normal tear down
	logrus.Info("Removing the logging-infra Logging if any")
	err = h.loggingClient.Delete(name.SafeConcatName(upgradeLog.Name, util.UpgradeLogInfraComponent), &metav1.DeleteOptions{})
		return err
	}

	// Remove the ManagedChart resource if the UpgradeLog resource is deleted before normal tear down
	logrus.Info("Removing logging-operator ManagedChart if any")
	err = h.managedChartClient.Delete(util.FleetLocalNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogOperatorComponent), &metav1.DeleteOptions{})
		return err
	}

	// Cleanup the relationship from its corresponding Upgrade resource
	upgradeName := upgradeLog.Spec.UpgradeName
	upgrade, err := h.upgradeCache.Get(util.HarvesterSystemNamespaceName, upgradeName)
	if err != nil {
			return nil
		}
		return err
	}
	upgradeToUpdate := upgrade.DeepCopy()
	upgradeToUpdate.Status.UpgradeLog = ""
	if _, err = h.upgradeClient.Update(upgradeToUpdate); err != nil {
		return err
	}

	return nil
}
```
IIRC, there are only two CRs that are not in the `harvester-system` namespace or are cluster-scoped and need to be explicitly deleted:
- ManagedChart
- Logging

Other resources have owner references set and will be purged automatically.

w13915984028:
@starbops  Thanks. I would explain a bit more:


The issue and PR are OOB work of issue https://github.com/harvester/harvester/issues/7404. 

My principle is to use least effort, with least change to solve the issue, not targeting to change too much for a perfect solution.


---

The original code, tried to:
(1) Get & set upgrade object, to unmark self ( already explained, it makes little sense)
(2) Delete ManagedChart object.
note: here it seems to have a bug, the `logging` object is not deleted

---

This PR, tries to:
(1) Get & set upgrade object, to unmark self; but tolerant any error to go on next step
(2) Reuse `stopCollect`, which deletes all sub resources, more than `managedchart`, no matter it has been deleted or will be deleted by `ownerReferences`, it does no harm.
https://github.com/harvester/harvester/blob/f27750ade47d48d6127cf142e4f61a3b40bae917/pkg/controller/master/upgradelog/controller.go#L747

This solution, is in the same sequences as original code but is more robust, and has been well tested.

I agree that the code looks not so elegant, but in this context, it is not the first priority.

---

Your new suggestion:
(1) Delete `logging` object
(2) Delete `managedchart` object
(3) Get & set upgrade object

It has a different sequence with original code, more checks & tests are needed to make sure it works in all cases.

At the meantime, strictly speaking, reuse `stopCollect` in the `cleanup` unifies the logic. I did not see a strong reason that the `stopCollect` and `cleanup` are doing similar work but differentiate the sub resources.

tserong:
> After discussion with @bk201, this PR is not supposed to be included in v1.4.2, there is a workaround to solve the occured issue

Thanks for the confirmation @w13915984028 !

pkg/controller/master/upgradelog/controller.go
```diff
@@ -750,44 +751,72 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
...
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("failed to delete clusteroutput %s/%s error %w", util.HarvesterSystemNamespaceName, name.SafeConcatName(upgradeLog.Name, util.UpgradeLogOutputComponent), err)
 	}
 	err = h.loggingClient.Delete(name.SafeConcatName(upgradeLog.Name, util.UpgradeLogInfraComponent), &metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("failed to delete logging client %s error %w", name.SafeConcatName(upgradeLog.Name, util.UpgradeLogInfraComponent), err)
 	}
+	// new to wait a bit, let logging operator finish the task
```
starbops: What does this comment imply? I don't see any difference between the ManagedChart deletion and the others. Do we need to implement a waiting mechanism here?

pkg/controller/master/upgradelog/controller.go
```diff
@@ -750,44 +751,72 @@ func (h *handler) stopCollect(upgradeLog *harvesterv1.UpgradeLog) error {
...
-	upgrade, err := h.upgradeCache.Get(util.HarvesterSystemNamespaceName, upgradeName)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil
+	var err1, err2, err3 error
+	err := fmt.Errorf("upgradeLog %s/%s upgrade %s cleanup errors", upgradeLog.Namespace, upgradeLog.Name, upgradeName)
+
+	// when the upgrade object is deleted, it's DeletionTimestamp is set, and all ownering resources' DeletionTimestamp are also set
+	// but there is no promise that upgrade and other object disappear one by another
+	// the below function has high chance to return IsNotFoun error
```
starbops:
```suggestion
	// the below function has high chance to return IsNotFound error
```

starbops (commented):
@w13915984028 Thanks for the explanation! Now I understand the intention behind it. The logic LGTM. Just some nits, please see below.

pkg/controller/master/upgradelog/controller.go
w13915984028: updated, thanks

pkg/controller/master/upgradelog/controller.go
w13915984028: removed the comment, thanks.

starbops (approved):
LGTM, thanks.

m-ildefons (approved):
LGTM

w13915984028:
@mergify backport v1.5

Issue #7539, opened 2025-02-07
[BUG] Restoring VM Backup to another Namespace or another Cluster is Broken


WebberHuang1118:
**Describe the bug**
Restoring VM Bacup to another Namespace or another Cluster is Broken

Please check the LH issue https://github.com/longhorn/longhorn/issues/10366 for details

**Related LH issue**
https://github.com/longhorn/longhorn/issues/10366

WebberHuang1118:
Per discussion with @bk201 and @FrankYang0529 

We will update the VolumeSnapshotConten's `snapshotHandle` to the correct format `bak://<backupvolume name>/<backup name>` with the following scenarios:
- For the newly created `VolumeSnapshotContent`, the Harvester controller will deploy it with the right `snapshotHandle` content
- For the existing `VolumeSnapshotContent`, the Harvester will update it on-demand
  - Once the VM Backup Restore is requested, the Harvester will check if the `VolumeSnapshotContent`'s `snapshotHandle` is in the wrong format `bs://<pvc name>/<backup name>`, then apply the following operations:
    - Getting the `backupvolume` name from LH `backup` CR 
    - Creating a new `VolumeSnapshot` and `VolumeSnapshotContent`
    - Connecting the original `VMBackup` CR with the new `VolumeSnapshot` and `VolumeSnapshotContent`

WebberHuang1118:
I had the verification to confirm whether [our proposed method](https://github.com/harvester/harvester/issues/7539#issuecomment-2642316452) works

**Prerequisite**
- Building a Harvester cluster with the master branch
- Updating Harvester deployment with the image `webberhuang/harvester:issue-7539-head`
  - This contains the modification to the Harvester controller to match https://github.com/mantissahz/longhorn/blob/5972d80c833cc7fc7d7df3757f0b0d0558f14f7c/enhancements/20240926-multiple-backup-targets-support.md
  - This contains PR to have[ the correct VolumeSnapshotContent's snapshotHandle](https://github.com/WebberHuang1118/harvester/commit/487de8c15e7845b8c1881d6c1b8f9d1191d60aaf) for the case `restoring VMBackup to another namespace` 
-  Updating LH manager daemonset with the image `phanle1010/longhorn-manager:v1.8.0-10303`
   - This contains the fix for https://github.com/longhorn/longhorn/issues/10303
- Configuring backup target with an NFS server
- Starting a VM `vm1`
  - Creating a PVC `os-vol` from a VM image
[...]

PR #7658, opened 2025-02-19
Updating VolumeSnapshotContent at runtime if needed

WebberHuang1118:
**Problem:**
We use the wrong format in VolumeSnapshotContent's `snapshotHandle`:
- [Recovering an existing backup on the remote target](https://github.com/harvester/harvester/blob/f8198b619e3904ed900b794a3ee7d00751493a92/pkg/controller/master/backup/backup.go#L730)
- [Restoring the vmbackup to another namespace](https://github.com/harvester/harvester/blob/f8198b619e3904ed900b794a3ee7d00751493a92/pkg/controller/master/backup/restore.go#L815-L816)

**Solution:**
Updating the VolumeSnapshotConten's `snapshotHandle` to the correct format `bak://<backupvolume name>/<backup name>` with the following scenarios:
- For the newly created `VolumeSnapshotContent`, the Harvester controller will deploy it with the right `snapshotHandle` content
- For the existing `VolumeSnapshotContent`, the Harvester will update it on-demand
  - Once the VM Backup Restore is requested, the Harvester will check if the `VolumeSnapshotContent`'s `snapshotHandle` is in the wrong format `bs://<pvc name>/<backup name>`, then apply the following operations:
    - Getting the `backupvolume` name from LH `backup` CR 
    - Creating a new `VolumeSnapshot` and `VolumeSnapshotContent`
    - Connecting the original `VMBackup` CR with the new `VolumeSnapshot` and `VolumeSnapshotContent`

**Related Issue:**
#7539 

**Test plan:**
**Prerequisite:**
- Building a Harvester cluster with `v1.4.1`
- Creating a VMBackup, disconnecting the remote target, and deleting the VMBackup
- Re-connecting the remote target, the VMBackup should be recovered
- Upgrading Harvester to the ISO built from this PR

**Case1: Restore the existing VMBackup**
- Restoring the VMBakcup which is created before the upgrade
- The VMBakcup should be able to restore

**Case2: New VMBackup**
- Creating a new VMBackup
- The new VMBackup should be able to restore

**Case3: Restore VMBackup to another Namespace**
- Restoring any of the above VMBackup to another namespace
- The VMBackup should be able to restore


pkg/controller/master/backup/restore.go
```diff
@@ -728,6 +733,112 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+		*vmb.Status.VolumeBackups[i].Name = *vb.Name + snapRevice
+		logrus.WithFields(logrus.Fields{
+			"VMBackupNamespace": vmb.Namespace,
+			"VMBackupName":      vmb.Name,
+			"VolumeBackupName":  *vb.Name,
+		}).Info("updating volume backup name")
+		return nil
+	}
+
+	return fmt.Errorf("vmbackup %s/%s volumebackup %s has no mached entry %v",
```
ihcsim:
```suggestion
	return fmt.Errorf("vmbackup %s/%s volumebackup %s has no matched entry %v",
```

pkg/controller/master/backup/restore.go
```diff
@@ -728,6 +733,112 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+	vsc, err := h.snapshotContentCache.Get(getVolumeSnapshotContentName(vb))
+	if err != nil {
+		return "", err
+	}
+
+	if vsc.Spec.Source.SnapshotHandle == nil {
+		return "", fmt.Errorf("vsc %s missing SnapshotHandle", vsc.Name)
+	}
+
+	_, vol, backup := decodeSnapshotID(*vsc.Spec.Source.SnapshotHandle)
```
ihcsim, line 811: There's a chance that `backup` will be empty. Do we want to explicitly handle that? Or just let the next `Get()` fail? WDYT?

pkg/controller/master/backup/restore.go
```diff
@@ -728,6 +733,112 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+
+	if !apierrors.IsNotFound(err) {
+		return "", err
+	}
+
+	if err := h.reviseVolumeSnapshot(vb, vr); err != nil {
+		return "", err
+	}
+
+	return "", fmt.Errorf("volume backup %s needs to update VolumeSnapshot", *vb.Name)
```
ihcsim, line 845: Can we add a comment here to explain why we need to revise the volume snapshot if we can't find the LH volume? AIUI, the main reason is because we have old VSCs that could still be using the old handle format. All the new VSCs won't have this problem, right?

Also, instead of returning `fmt.Errorf("volume backup %s needs to update VolumeSnapshot", *vb.Name)`, can we log and enqueue? I imagine if people see these error in SBs, we will have to explain to them that these errors are benign.

pkg/controller/master/backup/backup.go
```diff
@@ -58,6 +58,7 @@ const (
 	backupBucketRegionAnnotation = "backup.harvesterhci.io/bucket-region"
 
 	defaultFreezeDuration = 1 * time.Second
+	snapRevice            = "-snap-revise"
```
ihcsim:
```suggestion
	snapRevise            = "-snap-revise"
```

pkg/controller/master/backup/restore.go
```diff
@@ -728,6 +733,112 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+}
+
+func (h *RestoreHandler) getDataSourceAnotherNs(vb harvesterv1.VolumeBackup, vr *harvesterv1.VirtualMachineRestore) (string, error) {
+	// create volumesnapshot if namespace is different
```
ihcsim, line 849: Can we create a separate issue to track the CSI's [cross-namespace data source](https://kubernetes-csi.github.io/docs/cross-namespace-data-sources.html) feature? Currently, it's still in alpha version, but eventually, it will make cross-namespace restores a lot easier.

pkg/controller/master/backup/restore.go
```diff
@@ -728,6 +733,112 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+	*vmBackupCpy.Status.ReadyToUse = false
+
+	if err := h.updateVolumeBackup(vmBackupCpy, vb); err != nil {
+		return err
+	}
+
+	if vmBackupCpy.Annotations == nil {
+		vmBackupCpy.Annotations = make(map[string]string)
+	}
+	vmBackupCpy.Annotations[util.AnnotationSnapshotRevise] = strconv.FormatBool(true)
```
ihcsim, line 779: What's the purpose of this annotation?

pkg/controller/master/backup/restore.go
```diff
@@ -728,6 +733,112 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+		}).Info("updating volume backup name")
+		return nil
+	}
+
+	return fmt.Errorf("vmbackup %s/%s volumebackup %s has no mached entry %v",
+		vmb.Namespace, vmb.Name, *vb.Name, vb.VolumeName)
+}
+
+// We will update VolumeBackup's name with `snap-revise` suffix, this will trigger
+// VMBackup controller to create new VolumeSnapshot/VolumeSnapshotContent
```
ihcsim, line 761: With the new ones being created, do we need to delete the old ones?

pkg/controller/master/backup/restore.go
```diff
@@ -728,6 +733,112 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+// and link VMBackup to these new resources
+func (h *RestoreHandler) reviseVolumeSnapshot(vb harvesterv1.VolumeBackup, vr *harvesterv1.VirtualMachineRestore) error {
+	vmBackup, err := h.backupCache.Get(vr.Spec.VirtualMachineBackupNamespace, vr.Spec.VirtualMachineBackupName)
+	if err != nil {
+		return err
+	}
+
+	vmBackupCpy := vmBackup.DeepCopy()
+	*vmBackupCpy.Status.ReadyToUse = false
```
ihcsim, line 770: I assume the vmbackup controller will eventually reset this to `true` after the new vs and vsc are created?

ihcsim: Is it necessary to set this to `false`? If yes, I feel like there needs to be something on the vm backup side to indicate that it's no longer ready, because the restore controller flipped this state. Otherwise, how would we know why all of the sudden, a vm backup is no longer ready?

pkg/controller/master/backup/restore.go
WebberHuang1118: It will not happen [in our use case](https://github.com/harvester/harvester/blob/a55a368596267618275d4638ffae3604d98e6b1f/pkg/controller/master/backup/backup.go#L746-L747). 

If this really happens unexpectedly, the reconciler will just return the error from `Get()`

pkg/controller/master/backup/restore.go
WebberHuang1118: > Can we add a comment here to explain why we need to revise the volume snapshot if we can't find the LH volume? AIUI, the main reason is because we have old VSCs that could still be using the old handle format. All the new VSCs won't have this problem, right?
> 

If the volume is not found, it indicates that the VSC is in the wrong format. Therefore, we will create a new VS/VSC and connect it to the current VMBackup.

I will add a comment for this.

> Also, instead of returning `fmt.Errorf("volume backup %s needs to update VolumeSnapshot", *vb.Name)`, can we log and enqueue? I imagine if people see these error in SBs, we will have to explain to them that these errors are benign.

If I understand correctly, a return error will trigger reconciliation, and the controller will log it error

pkg/controller/master/backup/restore.go
WebberHuang1118: It indicates the VMBackup's related VS/VSC has been updated.

pkg/controller/master/backup/restore.go
WebberHuang1118: The old VS/VSC should be okay to remove, however, the `deletion` operation will trigger the backup deletion for LH (even if this will not happen due to the wrong VSC format), it's kind of dangerous.

As discussed with @bk201 and @FrankYang0529, [we prefer to keep the old VS/VSC](https://github.com/harvester/harvester/issues/7539#issuecomment-2642316452).

pkg/controller/master/backup/restore.go
WebberHuang1118: > Is it necessary to set this to `false`? If yes, I feel like there needs to be something on the vm backup side to indicate that it's no longer ready, because the restore controller flipped this state. Otherwise, how would we know why all of the sudden, a vm backup is no longer ready?

Setting `false` will make the VMBackup to be reconciled, and create new VS/VSC
- https://github.com/harvester/harvester/blob/f8198b619e3904ed900b794a3ee7d00751493a92/pkg/controller/master/backup/backup.go#L148-L150
- https://github.com/harvester/harvester/blob/f8198b619e3904ed900b794a3ee7d00751493a92/pkg/controller/master/backup/backup.go#L176-L179

So I added the annotation `snapRevise` to indicate the VMBackup revise is triggered. 

In addition, the new VS/VSC creation is an `O(1)` operation, so the flipping is not obvious, and we can tell the reason from the annotation.

pkg/controller/master/backup/restore.go
WebberHuang1118: The ticket is created https://github.com/harvester/harvester/issues/7744

WebberHuang1118:
Hi @ihcsim 

Thanks for your suggestion, I've replied or updated accordingly, please take a look.

pkg/controller/master/backup/restore.go
ihcsim: Won't it cause the restore to fail, per https://github.com/harvester/harvester/blob/c0e0a9eaabfa0e0bb096ae69477f9dfcb719bba1/pkg/controller/master/backup/restore.go#L187-L190

unless the controller has built-in retries after updating the restore's `status` subresource.

pkg/controller/master/backup/restore.go
ihcsim: To confirm, so it's not possible for us to encounter a BI snapshot type here? https://github.com/harvester/harvester/blob/c0e0a9eaabfa0e0bb096ae69477f9dfcb719bba1/pkg/controller/master/backup/util.go#L260-L262

pkg/controller/master/backup/restore.go
ihcsim: The error handling here feels a bit odd to me. The `restore` will be updated with error conditions, and the controller is expected to enqueue and retry the restore again, via the returning of a non-nil `err`? I'd think once a restore failed, we don't want the controller to auto-retry.

pkg/controller/master/backup/restore.go
WebberHuang1118: Yes, we will only have this route while recovering an existing VM/Volume backup on the remote target.

pkg/controller/master/backup/restore.go
WebberHuang1118: Yes, returning the error will trigger the reconciliation.

It's the original design for VM Backup/Restore since day one. 

If the VM Backup/Restore can achieve the desired state, it will be just a temporary status.
If not, the related CR will maintain the error in the status.

ihcsim (approved):
LGTM - thanks for this PR!

pkg/controller/master/backup/restore.go
```diff
@@ -728,6 +733,114 @@ func (h *RestoreHandler) updateOwnerRefAndTargetUID(vmRestore *harvesterv1.Virtu
...
+		logrus.WithFields(logrus.Fields{
+			"name":           vsc.Name,
+			"snapshotHandle": *vsc.Spec.Source.SnapshotHandle,
+		}).Info("VolumeSnapshotContent get correct snapshotHandle")
+		return *vb.Name, nil
+	}
+
+	// If the volume is not found, it indicates that the VSC is in the wrong format.
+	// Therefore, we will create a new VS/VSC and connect it to the current VMBackup.
+	if !apierrors.IsNotFound(err) {
```
FrankYang0529: We need to check whether VMBackup has `harvesterhci.io/snapRevise` here. If the source PVC is removed, we keep getting error here and there will be many volume snapshot content.

Screenshot 2025-03-05 at 3 57 43 PM

Reproduce steps:
1. Follow Prerequisite.
2. Remove source VM.
3. Restore VMBackup.

FrankYang0529 (commented):
Hi @WebberHuang1118, I found an edge case. Could you take a look? Thank you.

pkg/controller/master/backup/restore.go
WebberHuang1118: Hi @FrankYang0529 
Thanks for pointing this out, I've fixed this by checking LH `backupvolume` instead of `volume`, please take a look.

FrankYang0529 (approved):
LGTM. All test cases can pass. Thanks.

WebberHuang1118:
@mergify backport v1.5

Issue #7622, opened 2025-02-13
[BUG] Harvester ISO installation is broken


w13915984028:
**Describe the bug**

After `pre-mount hook`, it show `format .. by-label/COS_OEM...`

When validating https://github.com/harvester/harvester-installer/pull/980, this issue is encountered. Then I rebuilt the ISO from master-head, the issue is still existing.

![Image](https://github.com/user-attachments/assets/facd4697-07cf-47b7-bc6c-543f8b6eceab)

A local built ISO from Harvester installer, with latest commit https://github.com/harvester/harvester-installer/commit/7d5c68dd136e07c at 2024.02.13

![Image](https://github.com/user-attachments/assets/ed47f4df-6477-4d64-b55a-7a7cfc603683)

the installer stuck

installer option:

![Image](https://github.com/user-attachments/assets/644ad1c3-dbff-4fd4-9e74-235357727243)

![Image](https://github.com/user-attachments/assets/da1d481d-b3b5-4fdc-83f0-0180c8ee54a5)

**To Reproduce**
1. Build an ISO
2. Install via ISO mode

A KVM VM is used to install, which runs v1.4.0 cluster before, it has only 1 disk.

**Expected behavior**

**Support bundle**

**Environment**
 - Harvester ISO version:  master-head 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

w13915984028:
After many testings, the commit

https://github.com/harvester/harvester-installer/commit/d63556f5ea and PR https://github.com/harvester/harvester-installer/pull/938 looks to be the cause; if I remove this PR, the installer works well.

w13915984028:
After revert this line of change, it works.

@ibrokethecloud Any idea this change breaks master-head ISO installation on KVM VM environment?

![Image](https://github.com/user-attachments/assets/a9f5a7de-8819-4abe-94b0-508030537d16)

![Image](https://github.com/user-attachments/assets/27807cfc-2f07-4cda-a923-30452420cae8)

bk201:
@w13915984028 can you provide your VM xml (`virsh dumpxml <domain>`)

w13915984028:
@bk201 Here:

```
<domain type='kvm' id='1'>
  <name>harv21</name>
  <uuid>cd3a995a-6414-45b4-9f61-6d7a1aa29238</uuid>
[...]

PR #7678, opened 2025-02-23
change default setup for multipathd - upgrade path

ibrokethecloud:
tweak multipath behaviour in upgrade path to allow users to subsequently enable multipath post upgrade

**Problem:**

PR https://github.com/harvester/harvester-installer/pull/938 changed kernel arguments to disable multipathd during boot. The original idea was to allow end users to still enable multipathd post boot for 3rd party CSI.

However the change seems to have broken boot in kvm when using ide, scsi and sata buses. The same issue has also been noticed on HP DL360's when using a raid controller.

**Solution:**

* PR applies an override to multipathd.service file via /etc/systemd/system, where we drop the condition check for multipath=off during the upgrade path.

**Related Issue:**
https://github.com/harvester/harvester/issues/7622
**Test plan:**

* Install harvester 1.4-head or 1.4.2-rc*
* Upgrade to an iso built with changes from this PR
* Post upgrade all the nodes should have a file /oem/99_disable_lh_multipathd.yaml which sets up multipath exceptions for longhorn volumes and override for multipathd service
* user should be able to enable multipathd on the nodes `systemctl enable multipathd && systemctl start multipathd`

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -650,11 +650,9 @@ EOF
   if [ ${multiPathEnabled} == false ]
   then
     thirdPartyArgs=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
-    if [[ ${thirdPartyArgs} != *"rd.multipath=0"* ]]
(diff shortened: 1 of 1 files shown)


tserong (approved):

w13915984028 (approved):
LGTM, thanks.

w13915984028:
@mergify backport v1.5

PR #7684, opened 2025-02-24
fix: update pollInterval instead of LH setting

FrankYang0529:

**Solution:**
In LH 1.8.0, it has a BackupTarget CR. We have to update poll interval to BackupTarget instead of LH setting.
Improve reconciliation time: if `refreshIntervalSeconds` is zero, the controller should not enqueue event to trigger the controller again.

**Related Issue:**
https://github.com/harvester/harvester/issues/7083, #7764 

**Test plan:**
1. Create a cluster.
2. Setup backup target with `refreshIntervalInSeconds`.
3. Check the `longhorn-system/default` BackupTarget has same value in `PollInterval`.

### DIFF
--- a/pkg/controller/master/backup/backup_metadata.go
+++ b/pkg/controller/master/backup/backup_metadata.go
@@ -165,15 +165,18 @@ func (h *MetadataHandler) shouldRefresh(setting *harvesterv1.Setting, refreshInt
 		lastTime, err = time.Parse(time.RFC3339, setting.Annotations[util.AnnotationLastRefreshTime])
 		if err != nil {
 			logrus.WithError(err).Errorf("failed to parse last refresh time")
-			lastTime = time.Unix(0, 0)
 			return true
 		}
 	}
-	if getBackupTargetHash(setting.Value) == setting.Annotations[util.AnnotationHash] &&
-		(refreshIntervalInSeconds == 0 ||
-			currentTime.Sub(lastTime).Seconds() < float64(refreshIntervalInSeconds)) {
-		h.settings.EnqueueAfter(setting.Name, lastTime.Add(time.Duration(refreshIntervalInSeconds)*time.Second).Sub(currentTime))
-		return false
+	if getBackupTargetHash(setting.Value) == setting.Annotations[util.AnnotationHash] {
+		if refreshIntervalInSeconds == 0 {
+			return false
+		}
+
+		if currentTime.Sub(lastTime).Seconds() < float64(refreshIntervalInSeconds) {
+			h.settings.EnqueueAfter(setting.Name, lastTime.Add(time.Duration(refreshIntervalInSeconds)*time.Second).Sub(currentTime))
+			return false
+		}
 	}
 	return true
 }
--- a/pkg/controller/master/backup/backup_target.go
+++ b/pkg/controller/master/backup/backup_target.go
@@ -7,8 +7,8 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 
-	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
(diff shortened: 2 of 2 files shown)


w13915984028:
@FrankYang0529 @WebberHuang1118 

Copy the comment to here: https://github.com/harvester/harvester/pull/7536#discussion_r1967323902

It looks we can't use `longhorn-default-backupstore` to avoid the battlle between Harvester controller and LH configmap controller.

Like default SC, Harvester needs to setup a Harvester default backupstore, and in the upgrade path, convert the existing user backup setting to the new configure.

There is a risk that user config is lost after upgrading. 

Please correct me if I miss something. thanks.

w13915984028:
Leave this topic for open discussion:  use the Longhorn internal default object, it may be fragile to break.

e.g. user sets the `managedchart harvester`  `spec.values.longhorn.defaultBackupStore.backupTarget` explicitly/accidentally

https://github.com/longhorn/longhorn/blob/af3331a8f67e7eedf9df4c21c6eb2ff23244915c/chart/templates/default-resource.yaml#L10

```
$kubectl get backuptarget -A
NAMESPACE         NAME      URL   CREDENTIAL   LASTBACKUPAT   AVAILABLE   LASTSYNCEDAT
longhorn-system   default                      5m0s           false       <no value>

$kubectl get backuptarget -A -oyaml
apiVersion: v1
items:
- apiVersion: longhorn.io/v1beta2
  kind: BackupTarget
  metadata:
    creationTimestamp: "2025-02-11T09:23:08Z"
    finalizers:
    - longhorn.io
    generation: 1
    name: default
    namespace: longhorn-system
    resourceVersion: "6245"
    uid: bd73d65d-5343-4480-9451-504a71059c8f
  spec:
    backupTargetURL: ""
    credentialSecret: ""
    pollInterval: 5m0s
    syncRequestedAt: null
  status:
    available: false
    conditions:
    - lastProbeTime: ""
      lastTransitionTime: "2025-02-11T09:24:08Z"
      message: backup target URL is empty
      reason: Unavailable
      status: "True"
      type: Unavailable
    lastSyncedAt: null
    ownerID: harv41
```

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -165,15 +165,16 @@ func (h *MetadataHandler) shouldRefresh(setting *harvesterv1.Setting, refreshInt
...
-			currentTime.Sub(lastTime).Seconds() < float64(refreshIntervalInSeconds)) {
-		h.settings.EnqueueAfter(setting.Name, lastTime.Add(time.Duration(refreshIntervalInSeconds)*time.Second).Sub(currentTime))
-		return false
+	if getBackupTargetHash(setting.Value) == setting.Annotations[util.AnnotationHash] {
+		if refreshIntervalInSeconds == 0 {
+			return false
+		} else if currentTime.Sub(lastTime).Seconds() < float64(refreshIntervalInSeconds) {
+			h.settings.EnqueueAfter(setting.Name, lastTime.Add(time.Duration(refreshIntervalInSeconds)*time.Second).Sub(currentTime))
+			return false
+		}
```
WebberHuang1118: Maybe we don't need the if and else if here?

WebberHuang1118 (approved):
LGTM, just a nit, thanks.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: This is to address @w13915984028's comment in https://github.com/harvester/harvester/pull/7114#discussion_r1966290355.

w13915984028: guess like belw?
```
		if refreshIntervalInSeconds == 0 {
			return false
		} 
		
		if currentTime.Sub(lastTime).Seconds() < float64(refreshIntervalInSeconds) {
			h.settings.EnqueueAfter(setting.Name, lastTime.Add(time.Duration(refreshIntervalInSeconds)*time.Second).Sub(currentTime))
			return false
		}
		...
```

FrankYang0529: Updated it.

w13915984028 (approved):
LGTM, thanks.

FrankYang0529:
@mergify backport v1.5

Issue #7695, opened 2025-02-25
[FEATURE] boot_order support for network interfaces


jainjj:
**Is your feature request related to a problem? Please describe.**
Harvester VMs booting up from a PXE server. Managed to make this work via the GUI changing the YAML file as per https://github.com/harvester/harvester/issues/3325. However trying to do the same via the terraform provider, the boot order attribute (boot_order in terraform) seems to be available only for the disks (https://registry.terraform.io/providers/harvester/harvester/latest/docs/resources/virtualmachine#nested-schema-for-disk), not for the interfaces.

**Describe the solution you'd like**
We would like boot order attribute (boot_order in terraform) to be available/supported via terraform for Harvester

asc5543:
Hi @m-ildefons ,
I checked the https://github.com/harvester/terraform-provider-harvester/pull/122 and the test plan is `Tests included.`.
Could you help to provide the test steps of this issue?
Thank you.

m-ildefons:
Hi Justin,

With "Tests included" I meant to say that the PR includes a test case in the acceptance tests:
https://github.com/harvester/terraform-provider-harvester/pull/122/files#diff-88529df1f3c90b034c8e930f13f8325e05cd36b3a62505ef0a8733a2767a5d4aR220

Vicente-Cheng:
Hi @m-ildefons, 
I know we already have the test on the code level.
But for the QA validation, we need to have the user view the test plan.

Could you help update the test plan?
Thanks!

cc @innobead

Vicente-Cheng:
Move back to `Ready for Testing` to wait for the test plan.

m-ildefons:
# Test Plan

The goal of this test plan is to demonstrate that the `boot_order` property of the `network_interface` block in the `harvester_virtualmachine` Terraform resource can be used to facilitate network booting for VMs.

## Setup

### 1. Set up single-node Harvester in ipxe-examples environment

It's not mandatory to use the ipxe-examples environment, but it's convenient, because it already includes everything needed for a PXE boot environment for VMs

### 2. Prepare PXE boot environment for VMs

1) Set up VM network

<img width="1740" height="829" alt="Image" src="https://github.com/user-attachments/assets/8e903d84-dab8-4606-bc34-0ad39c3c9fb7" />

2) Download Live System Image

I'll use GRML in this example, because it's small, it's made to be used as live-system and there are instructions online how to set it up with PXE

```
[...]

PR #7698, opened 2025-02-25
VMBuilder: expose network interface boot order

m-ildefons:
Expose a function that allows setting the boot order property of network interfaces in the VMBuilder.

The main use case for this is the Terraform provider, which utilizes the VMBuilder to generate VM structures.

Exposing this property as its own method rather than modifying the existing `Interface` method keeps compatibility for other users of the VMBuilder.

related-to: https://github.com/harvester/harvester/issues/7695

**Problem:**
The VMBuilder is used by the Harvester Terraform provider to generate the structure of a VirtualMachine object from the specifications in the Terraform manifest. To allow creating VMs, which can do network/PXE booting with the Terraform provider, the provider must be able to set the boot order property on a network interface of a VM structure.

**Solution:**
Add an additional method to the VMBuilder, which allows setting the boot order property of a specific network interface.

**Related Issue:**
https://github.com/harvester/harvester/issues/7695

**Test plan:**
Unit tests included

### DIFF
--- a/pkg/builder/network.go
+++ b/pkg/builder/network.go
@@ -44,7 +44,7 @@ func (v *VMBuilder) Network(interfaceName, networkName string) *VMBuilder {
 	return v
 }
 
-func (v *VMBuilder) Interface(interfaceName, interfaceModel, interfaceMACAddress string, interfaceType string) *VMBuilder {
+func (v *VMBuilder) Interface(interfaceName, interfaceModel, interfaceMACAddress, interfaceType string) *VMBuilder {
 	interfaces := v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces
 	networkInterface := kubevirtv1.Interface{
 		Name:       interfaceName,
@@ -66,6 +66,20 @@ func (v *VMBuilder) Interface(interfaceName, interfaceModel, interfaceMACAddress
 	}
 	interfaces = append(interfaces, networkInterface)
 	v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces = interfaces
+	return v
+}
+
+func (v *VMBuilder) SetNetworkInterfaceBootOrder(interfaceName string, bootOrder uint) *VMBuilder {
+	interfaces := v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces
+
+	for i, iface := range interfaces {
+		if iface.Name == interfaceName {
+			ifaceCopy := iface.DeepCopy()
+			ifaceCopy.BootOrder = func() *uint { v := uint(bootOrder); return &v }()
+			v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces[i] = *ifaceCopy
+		}
+	}
+
 	return v
 }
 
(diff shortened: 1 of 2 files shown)


pkg/builder/network.go
```diff
@@ -69,6 +69,20 @@ func (v *VMBuilder) Interface(interfaceName, interfaceModel, interfaceMACAddress
 	return v
 }
 
+func (v *VMBuilder) SetNetworkInterfaceBootOrder(interfaceName string, bootOrder uint) *VMBuilder {
+	interfaces := v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces
+
+	for i, iface := range interfaces {
+		if iface.Name == interfaceName {
```
w13915984028, line 76:
```
		if iface.Name == interfaceName {
			if iface.BootOrder == nil || *iface.BootOrder != bootOrder {
				ifaceCopy := iface.DeepCopy()
				bo := bootOrder
				ifaceCopy.BootOrder = &bo
				v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Interfaces[i] = *ifaceCopy
			}
			return // or break
		}
```

another question: are two interfaces allowed to have same boot order? if not, a check is needed.

m-ildefons: No, two interfaces can not have the same boot order value. As a matter of fact an interface can also not have the same boot order value as a disk (`v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Disks`) either.
But this is the wrong place to check.
The VMBuilder is just a constructor for the golang struct of a VirtualMachine object.
Whether the structure makes sense or is valid is for the validation webhook to determine.

https://kubevirt.io/api-reference/main/definitions.html#_v1_interface

ibrokethecloud (approved):
lgtm. thanks.

pkg/builder/network.go
```diff
@@ -69,6 +69,20 @@ func (v *VMBuilder) Interface(interfaceName, interfaceModel, interfaceMACAddress
...
+			ifaceCopy := iface.DeepCopy()
+			ifaceCopy.BootOrder = func() *uint { v := uint(bootOrder); return &v }()
```
votdev: This construction looks difficult. Would it make sense to simplify this, as @w13915984028 does in his code.

```suggestion
			bo := bootOrder
			ifaceCopy.BootOrder = &bo
```

m-ildefons: If you say so...

Golang (like most programming languages) allocates local variables in a function on the stack by default. But it does recognize when a variable should live beyond the scope of the function, in which case it allocates the variable on the heap instead.

This is a pretty standard way in to force the allocation on the heap and get a pointer to it, which is necessary here because `iface.BootOrder` is a pointer and needs to point to a valid location in the heap.

Since the lambda will be inlined, there is no performance penalty.

The method Jian proposed in his code snipped also works, but it splits the allocation and assignment over two lines, which I think makes the code less clear.
There is also a way to (ab)use arrays to do the same, which is more compact, but I find the type signature on that one harder to read.

innobead (approved):

PR #7704, opened 2025-02-26
fix: ignore VMBackup/VMImage if dependency is not existent

FrankYang0529:

**Problem:**
The backup target can be used among multiple clusters. If one cluster removes a VMBackup, it also removes data in backup target. If another cluster reconcile the VMBackup, it uploads the VMBackup metadata to the backup target again. In this case, there is VMBackup metadata in backup target without LH data, so the VMBackup can't be synced successfully again in a new cluster. Users will see a broken VMBackup in the new cluster.

**Solution:**

Check VMBackup dependencies are existent before syncing the VMBackup back to the cluster.

**Related Issue:**
https://github.com/harvester/harvester/issues/7083

**Test plan:**

1. Create two clusters A and B.
2. Set backup target on cluster A.
3. Create VM and VMBackup on cluster A.
4. Set backup target with `refreshIntervalInSeconds` on cluster B. We can see the VMImage and VMBackup are synced to cluster B.
5. Remove VMBackup on cluster B. We can see VMBackup metadata is removed in the backup target.
6. Modify the VMBackup on cluster A like adding an annotation to trigger reconciliation. We can see the VMBackup metadata is in the cluster again.
7. Check harvester pod log on cluster B. You can see log like following to skip syncing VMBakcup.
```
"skip creating vm backup, because the longhorn backup is not found" longhornBackup=backup-dca9304e20f7472f name=b1 namespace=default volume=disk-0
```

### DIFF
--- a/pkg/controller/master/backup/backup_metadata.go
+++ b/pkg/controller/master/backup/backup_metadata.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"path/filepath"
 	"reflect"
 	"time"
@@ -23,6 +24,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/utils/pointer"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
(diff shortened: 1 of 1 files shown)


WebberHuang1118:
Hi @FrankYang0529 

Is there a possible situation for a time window?
- CR bi, backup does exist on the remote, but they haven't been recovered as k8s CR
- The harvester controller finds the metadata in the remote and tries to bring up related resources

If this happens, we may miss certain VMBackups?

FrankYang0529:
> Is there a possible situation for a time window?
> 
> * CR bi, backup does exist on the remote, but they haven't been recovered as k8s CR
> * The harvester controller finds the metadata in the remote and tries to bring up related resources

Yes, that is why we check both data in backup target and in the cluster in `checkDependentLonghornBackupExist` funciton.

WebberHuang1118:
> > Is there a possible situation for a time window?
> > 
> > * CR bi, backup does exist on the remote, but they haven't been recovered as k8s CR
> > * The harvester controller finds the metadata in the remote and tries to bring up related resources
> 
> Yes, that is why we check both data in backup target and in the cluster in `checkDependentLonghornBackupExist` funciton.

If the mentioned situation happens, do we have a retry mechanism?
IIUC, if we find the resource exists in remote but not as CR in k8s, we will skip it?

FrankYang0529:
> > > Is there a possible situation for a time window?
> > > 
> > > * CR bi, backup does exist on the remote, but they haven't been recovered as k8s CR
> > > * The harvester controller finds the metadata in the remote and tries to bring up related resources
> > 
> > 
> > Yes, that is why we check both data in backup target and in the cluster in `checkDependentLonghornBackupExist` funciton.
> 
> If the mentioned situation happens, do we have a retry mechanism? IIUC, if we find the resource exists in remote but not as CR in k8s, we will skip it?

Yes, and it will be added in the next refresh interval.

WebberHuang1118 (approved):
LGTM, thanks.

bk201 (approved):
lgtm, thanks!

Issue #6961, opened 2024-11-07
[BUG] Fail to delete an Off VM if its network was deleted


albinsun:
**Describe the bug**
Fail to delete an Off VM if its network was deleted
![image](https://github.com/user-attachments/assets/604c31dd-4a87-4ba2-86df-dd39b8c9b154)

**Prerequisite**
1. VM network 
   ![image](https://github.com/user-attachments/assets/078b84df-f18d-49ff-a1d3-3e4de85f8c02)
1. Running VM using that network
   ![image](https://github.com/user-attachments/assets/23bff34d-5db7-493a-b6e7-bfc2d22b20b4)

**To Reproduce**
1. Network should not be able to delete if it's using by a `Running` VM 
   > Harvester suggest me to stop the VM first.
   ![image](https://github.com/user-attachments/assets/38eb04b4-b424-4421-81dd-453da53952ed)
1. Stop VM
   ![image](https://github.com/user-attachments/assets/2fe3151f-697f-45af-b3ed-0a3ee24267a2)
1. Now able to Delete network
   ![image](https://github.com/user-attachments/assets/20d06a75-926b-4c62-9b7e-f7d02f7c05c0)
1. But fail to delete VM with 500 error, and no GUI error message.
   ![image](https://github.com/user-attachments/assets/604c31dd-4a87-4ba2-86df-dd39b8c9b154)

**Expected behavior**
Depends on the delete validation, at least we should popup error message instead of 500 error in the dev consle. 

**Support bundle**
[supportbundle_ee4e61ad-a81c-4db9-930a-601f62cc6c35_2024-11-07T05-46-30Z.zip](https://github.com/user-attachments/files/17656668/supportbundle_ee4e61ad-a81c-4db9-930a-601f62cc6c35_2024-11-07T05-46-30Z.zip)

**Environment**
 - Harvester ISO version: v1.4.0-rc5
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Baremetal

**Additional context**
1. Start VM fails as expectation and has error message
   ![image](https://github.com/user-attachments/assets/b15e9a8d-8829-4d46-a0c4-f894c8ca3635)

albinsun:
First of all we should confirm the expected behavior:
Does Harvester really allow to delete a network if it still referenced by stopped VMs?
* Yes -> This issue 
* No -> Let's change title to "Network should not be able to delete if still referenced by VMs"

bk201:
[...]

PR #7743, opened 2025-03-03
Do not patch VM affinity when VM is stopped

w13915984028:

**Problem:**

Per issue description https://github.com/harvester/harvester/issues/6961

When a VM is stopped, it's nad may be deleted, and then the VM can't be deleted.

**Solution:**

Don't patch VM's affinity when VM is stopped/being deleted. For those VMs which has dangling interfaces, they are allowed to be deleted.

**Related Issue:**
https://github.com/harvester/harvester/issues/6961

**Test plan:**

create a VM, attache to NAD, start then stop it
image

delete NAD
image

the stopped VM can be deleted successfully
image

### DIFF
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -351,7 +351,9 @@ func (m *vmMutator) getAdditionalGuestMemoryOverheadRatioConfig() (*settings.Add
 }
 
 func (m *vmMutator) patchAffinity(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) (types.PatchOps, error) {
-	if vm == nil || vm.Spec.Template == nil {
+	// if vm is stopped/terminating/stopping, do not patch affinity, it makes no sense, and may block the deletion of vm
+	// e.g. the vm's nad has been deleted but vm's interface is dangling
+	if vm == nil || vm.Spec.Template == nil || vm.DeletionTimestamp != nil || vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopped || vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopping || vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusTerminating {
 		return patchOps, nil
 	}
 


w13915984028:
@mergify backport v1.5

starbops:
Still under review

FrankYang0529:
@w13915984028 Thanks for the PR. We encounter similar issue when removing a namespace. I think we can just skip the update request mutation if new VM has deletionTimestamp, because remove request mutation can be handled by `admissionregv1.Delete`. I already create another PR for it. Thanks.

https://github.com/harvester/harvester/pull/8110

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -351,7 +351,9 @@ func (m *vmMutator) getAdditionalGuestMemoryOverheadRatioConfig() (*settings.Add
 }
 
 func (m *vmMutator) patchAffinity(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) (types.PatchOps, error) {
-	if vm == nil || vm.Spec.Template == nil {
+	// if vm is stopped/terminating/stopping, do not patch affinity, it makes no sense, and may block the deletion of vm
+	// e.g. the vm's nad has been deleted but vm's interface is dangling
+	if vm == nil || vm.Spec.Template == nil || vm.DeletionTimestamp != nil || vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopped || vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusStopping || vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusTerminating {
```
FrankYang0529, line 356: Hi @w13915984028, since #8110 is merged, we can remove `vm == nil` and `vm.DeletionTimestamp != nil` here. Thanks.

w13915984028: @FrankYang0529 Thanks for your reminder, I will double check related PRs and update accordingly.

w13915984028:
Even when editing an `Off` VM and delete the dangling interface, there is still an webhook error reported.

image

image

booked via https://github.com/harvester/harvester/issues/8293

w13915984028:
With PR https://github.com/harvester/harvester/pull/8110, we can close this PR https://github.com/harvester/harvester/pull/7743.

Another issue https://github.com/harvester/harvester/issues/8293 is observed, I will add a new PR to fix it.

w13915984028:
https://github.com/harvester/harvester/pull/7743#issuecomment-2883402010

Issue #7739, opened 2025-03-03
[BUG] VM Snapshot Resotere can't Automatically Boot Up if the Source VM has 3rd party storage


WebberHuang1118:
**Describe the bug**
VM Snapshot Resotere can't Automatically Boot Up if the Source VM has 3rd party storage 

**To Reproduce**
1. Creating a VM with only one 3rd party storage
2. Stopping the VM
3. Taking a VM Snapshot
4. Restoring the VM Snapshot to a new VM
5. The VM will not start automatically

**Expected behavior**
The restored VM should boot up automatically

brandboat:
Test OK, will close the issue.

### Environment
- ipxe 3 node cluster
- harvester v1.5.0-rc1
- 8C/16G/250GB for each node

### Test Steps
- Follow the https://docs.harvesterhci.io/v1.5/advanced/addons/lvm-local-storage/#installing-and-enabling-harvester-csi-driver-lvm to setup harvester-csi-driver-lvm
- Create a VM `vm1` with 3 volumes, one with vm image, the other one is yet the same vm image, the last one is a empty LVM volume
![Image](https://github.com/user-attachments/assets/d0e13857-5e17-41f2-a5b4-390bf444c54c)
![Image](https://github.com/user-attachments/assets/4df0c160-bcb9-4e9d-a220-c749ade6486f)
- Run the cmd in VM, `sudo cp /dev/vdb /dev/vdc`, `sudo systctl -w vm.drop_caches=3`
- Stop `vm1`
- remove `disk2` (the LVM disk) from `vm1`
- Create vm2 with existing volume `disk2`.
- Taking `vm2` Snapshot
- Restoring the `vm2` Snapshot to a new VM, `vm3`
- Check `vm3` start up successfully.

PR #7749, opened 2025-03-04
fix: third party PVC should trigger VM Restore reconciliation

WebberHuang1118:
**Problem:**
VM Snapshot Resotere can't Automatically Boot Up if the Source VM has 3rd party storage

**Solution:**
While the third-party PVC is updated, we should trigger VM Restore reconciliation

**Related Issue:**
#7739

**Test plan:**
1. Creating a VM with only one 3rd party storage
2. Stopping the VM
3. Taking a VM Snapshot
4. Restoring the VM Snapshot to a new VM
5. The restoring VM should start automatically

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -15,6 +15,7 @@ import (
 	snapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	lhdatastore "github.com/longhorn/longhorn-manager/datastore"
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	"github.com/longhorn/longhorn-manager/types"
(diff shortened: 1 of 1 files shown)


pkg/controller/master/backup/restore.go
```diff
@@ -273,10 +273,17 @@ func (h *RestoreHandler) PersistentVolumeClaimOnChange(_ string, pvc *corev1.Per
 	}
 
 	volume, err := h.volumeCache.Get(util.LonghornSystemNamespaceName, volumeName)
-	if err != nil {
+	if apierrors.IsNotFound(err) {
```
bk201, line 293: Is it better to check the PV's `spec.csi.driver` field rather than just relying on getting LH volume? If the log message is important we can also make the level info.

WebberHuang1118: Updated per suggestion, thanks.

bk201 (approved):

FrankYang0529 (approved):

Issue #7679, opened 2025-02-23
[BUG] v1.4.1 -> v1.4.2-rc1 upgrade stuck in Pre-drained


albinsun:
**Describe the bug**
![Image](https://github.com/user-attachments/assets/ae55b2c5-a466-4d37-a72f-ef5dee7b8096)

**To Reproduce**
1. Setup 4 nodes `v1.4.1`
2. Create 3 VMs
   * 1 w/ some data written, take backup
   * 1 restored from the backup
   * 1 w/ additional volume w/ new storage class
3. Perform upgrade to `v1.4.2-rc1`

**Expected behavior**
Upgrade successfully

**Support bundle**
* [hvst-upgrade-vzwn4-upgradelog-archive-2025-02-23T13-17-05Z.zip](https://github.com/user-attachments/files/18930641/hvst-upgrade-vzwn4-upgradelog-archive-2025-02-23T13-17-05Z.zip)
* [supportbundle_154248fa-b60d-4574-8561-840598dd6d12_2025-02-23T13-17-38Z.zip](https://github.com/user-attachments/files/18930662/supportbundle_154248fa-b60d-4574-8561-840598dd6d12_2025-02-23T13-17-38Z.zip)

**Environment**
 - Harvester ISO version: `v1.4.1` -> `v1.4.2-rc1`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**
n/a

tserong:
Some notes:

```
# kubectl -n longhorn-system logs daemonsets/longhorn-manager 
[...]
time="2025-02-24T00:16:29Z" level=info msg="Node instance-manager-9f8399d93598184ecfe0f6fe3063319b is marked unschedulable but removing hp-113-tink-system PDB is blocked: some volumes are still attached InstanceEngines count 1 pvc-576ba86b-0980-42ae-a7f5-a21d34a8e858-e-1 " func="controller.(*InstanceManagerController).syncInstanceManagerPDB" file="instance_manager_controller.go:823" controller=longhorn-instance-manager node=hp-113-tink-system
```

So there's still a volume attached to `hp-113-tink-system`, but this problem is _not_ https://docs.harvesterhci.io/v1.4/upgrade/v1-4-0-to-v1-4-1#1-upgrade-is-stuck-in-the-pre-drained-state, because the engine is still running:

```
[...]

PR #7775, opened 2025-03-07
fix(upgrade): disable auto-cleanup-system-generated-snapshot

tserong:
**Problem:**
When live migrating VMs during upgrade, it's possible to hit a case where a replica is being rebuilt at _just_ the wrong time such that an automatic snapshot purge results in I/O errors for the VM on the target node.

**Solution:**
Disabling auto-cleanup-system-generated-snapshot for the duration of the upgrade prevents snapshot purge at that time and thus avoids the issue.

This commit also incidentally fixes a problem where the replica-replenishment-wait-interval setting value wasn't reverted after upgrade completion due to the upgrade CR's annotations being forcibly set to an empty map.

**Related Issue:**
https://github.com/harvester/harvester/issues/7679

**Test plan:**
- Install v1.4.1 with at least two nodes.
- Initiate an upgrade to a version which includes this fix.
- Check `kubectl -n longhorn-system get settings.longhorn.io/auto-cleanup-system-generated-snapshot` - it should show the value as `true`
- Once the upgrade gets through updating system services, but before the nodes are drained, the value of `auto-cleanup-system-generated-snapshot` should automatically change to `false` (try `watch -d 'kubectl -n longhorn-system get settings.longhorn.io/auto-cleanup-system-generated-snapshot'` in a terminal).
- Once the upgrade is complete, the value should return to `true`.

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -69,6 +69,9 @@ const (
 	replicaReplenishmentAnnotation           = "harvesterhci.io/" + replicaReplenishmentWaitIntervalSetting
 	extendedReplicaReplenishmentWaitInterval = 1800
 
+	autoCleanupSystemGeneratedSnapshotSetting    = "auto-cleanup-system-generated-snapshot"
+	autoCleanupSystemGeneratedSnapshotAnnotation = "harvesterhci.io/" + autoCleanupSystemGeneratedSnapshotSetting
+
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (approved):
lgtm, thanks for the quick enhancement.

bk201 (approved):
Thanks!

starbops (approved):
LGTM, thanks.

c3y1huang (approved):

derekbit:
LGTM

derekbit (approved):

ChanYiLin (approved):
LGTM!

PR #7788, opened 2025-03-07
fix: checking storage provider's ability before VM backup/snapshot creation

WebberHuang1118:
**Problem:**

VM Backup/Snapshot should have the following rules for different storage providers:
- VM with LH v1 volume:
   - Allowing VM snapshot and VM backup
- VM with LH v2 volume
  - Rejecting VM snapshot and VM backup
- VM with third-party storage
  - Allowing VM snapshots and rejecting VM backup

**Solution:**
Improve webhook with the above rules

**Related Issue:**
#7316 

**Test plan:**

**Prerequisite:**
- Building Harvester with master branch `a34d197e5067fc67dd51b5deb8d0c8778de34c6d`
- Updating `harvester-webhook` deployment with this PR

**Case LH v1 volume**:
- Creating a VM with LH v1 volume
- The VM should be able to have backups and snapshots

**Case LH v2 volume**:
- Creating a VM with LH v2 volume
- The VM Backup and snapshot will be rejected

**Case LVM volume**:
- Creating a VM with LH v1 volume
- The VM should be able to have snapshots, but the VM backups will be rejected

### DIFF
--- a/pkg/webhook/util/snapshot.go
+++ b/pkg/webhook/util/snapshot.go
@@ -3,14 +3,19 @@ package util
 import (
 	"fmt"
 
+	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhtypes "github.com/longhorn/longhorn-manager/types"
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	longhornutil "github.com/longhorn/longhorn-manager/util"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
+	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	ctllonghornv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
+	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 )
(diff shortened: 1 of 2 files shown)


pkg/webhook/util/snapshot.go
```diff
@@ -112,3 +117,94 @@ func CheckTotalSnapshotSizeOnNamespace(
...
+	// Determine which configuration value is required based on the backup type.
+	var requiredValue string
+	switch bt {
+	case v1beta1.Backup:
+		requiredValue = c.BackupVolumeSnapshotClassName
+	case v1beta1.Snapshot:
+		requiredValue = c.VolumeSnapshotClassName
+	}
+
+	if requiredValue == "" {
```
Vicente-Cheng, line 204: How about moving the non-supported case (means `requiredValue == ""`) to default case of switch-case?

Vicente-Cheng (approved):
lgtm, just a nit

pkg/webhook/util/snapshot.go
WebberHuang1118: Good idea, it's a legacy implementation, I will update it, thanks

WebberHuang1118: Hi @Vicente-Cheng 

`requiredValue` will be updated[ during the switch case for VMBackup type checking](https://github.com/harvester/harvester/blob/1bb727a5051636ce1ddb44191b0a975e24ff7f72/pkg/webhook/util/snapshot.go#L196-L202), so `requiredValue == ""` is not only specific to the switch-case default condition (for example a VM is required to have backup, but the VSC is not set for backup type ). I think we have to keep this implementation. Thanks.

pkg/webhook/util/snapshot.go
```diff
@@ -112,3 +117,94 @@ func CheckTotalSnapshotSizeOnNamespace(
...
+	if err != nil {
+		return err
+	}
+
+	ability := getAbilityForProvisioner(provisioner)
+	if !ability.IsSupported(bt) {
+		return fmt.Errorf("provisioner %s is not supported for type %s", provisioner, bt)
+	}
+
+	// Get origin provisoner and the CSI configuration
```
bk201:
```suggestion
	// Get origin provisioner and the CSI configuration
```

pkg/webhook/util/snapshot.go
```diff
@@ -112,3 +117,94 @@ func CheckTotalSnapshotSizeOnNamespace(
...
+	}
+
+	return string(engine.Spec.DataEngine) + "." + provisioner, nil
+}
+
+func ValidateProvisionerAndConfig(pvc *corev1.PersistentVolumeClaim,
+	engineCache ctllonghornv1.EngineCache, bt v1beta1.BackupType,
+	cdc map[string]settings.CSIDriverInfo) error {
+
+	// Get wrapper provisoner for checking snapshot/backup ability
```
bk201:
```suggestion
	// Get wrapper provisioner for checking snapshot/backup ability
```

bk201 (approved):

FrankYang0529 (approved):
LGTM.

WebberHuang1118:
@mergify backport v1.5

Issue #7791, opened 2025-03-07
[BUG] upgrade controller does not handle read-message well due to UI menu `Dismiss it` wipes upgrade CR's status


w13915984028:
**Describe the bug**

---

**update 2025.03.12: The root cause, is UI upgrade pop-up window menu action `dismiss it` wipes the upgrade CR's status, and the controller may re-start the state machine.**   Refer comment https://github.com/harvester/harvester/issues/7791#issuecomment-2717104180 and  PR: https://github.com/harvester/harvester-ui-extension/pull/201

---

When the upgrade is done, it has such log:
```
time="2025-03-07T12:28:02Z" level=info msg="Stop collecting logs"
time="2025-03-07T12:28:02Z" level=info msg="Tearing down the logging infrastructure for upgrade procedure"
```

After user clicks `dismiss` menu of upgrade pop-up window,

How ever, following log also comes:
```
time="2025-03-07T12:30:57Z" level=info msg="Initialize upgrade harvester-system/hvst-upgrade-sf594"
time="2025-03-07T12:30:57Z" level=info msg="Skip store VM state" name=hvst-upgrade-sf594 namespace=harvester-system
time="2025-03-07T12:30:57Z" level=info msg="Enabling upgrade observability"
time="2025-03-07T12:30:57Z" level=info msg="handle upgrade harvester-system/hvst-upgrade-sf594 with labels map[harvesterhci.io/latestUpgrade:true harvesterhci.io/read-message:true harvesterhci.io/upgradeState:PreparingLoggingInfra]"
```

This is due to the upgrade controller tries to create `upgradeLog` again

https://github.com/harvester/harvester/blob/68d8151a85b19d8ad2704831236e2ef7d74a6a70/pkg/controller/master/upgrade/upgrade_controller.go#L151

The `harvesterhci.io/upgradeState: PreparingLoggingInfra` label is on the upgrade object.

```
- apiVersion: harvesterhci.io/v1beta1
  kind: Upgrade
  metadata:
    annotations:
      harvesterhci.io/image-cleanup-plan-completed: "true"
    creationTimestamp: "2025-03-07T12:21:43Z"
    finalizers:
    - wrangler.cattle.io/harvester-upgrade-controller
    generateName: hvst-upgrade-
    generation: 17
    labels:
      harvesterhci.io/latestUpgrade: "true"
      harvesterhci.io/read-message: "true"
      harvesterhci.io/upgradeState: PreparingLoggingInfra
[...]

PR #7792, opened 2025-03-07
Add processing of read-message to upgrade controller

w13915984028:

**Problem:**

https://github.com/harvester/harvester/issues/7791

**Solution:**

Do not continue upgrade status machine when read-message label is there.

**Related Issue:**
https://github.com/harvester/harvester/issues/7791

**Test plan:**

1. Upgrade v150-v150, enable upgradeLog
2. After upgrade is done, click `dismiss` on UI
3. Check Harvester POD, the log should be like below; not like the ones on issue https://github.com/harvester/harvester/issues/7791

```
...
time="2025-03-10T10:26:38Z" level=info msg="upgrade harvester-system/hvst-upgrade-7xn6v has harvesterhci.io/read-messagelabel, no further processing"
time="2025-03-10T10:26:38Z" level=info msg="upgrade harvester-system/hvst-upgrade-7xn6v has harvesterhci.io/read-message label, no further processing"
....
apiVersion: v1
items:
- apiVersion: harvesterhci.io/v1beta1
  kind: Upgrade
  metadata:
    annotations:
      harvesterhci.io/image-cleanup-plan-completed: "true"
    creationTimestamp: "2025-03-10T10:19:53Z"
    finalizers:
    - wrangler.cattle.io/harvester-upgrade-controller
    generateName: hvst-upgrade-
    generation: 17
    labels:
      harvesterhci.io/latestUpgrade: "true"
      harvesterhci.io/read-message: "true"
      harvesterhci.io/upgradeState: Succeeded
    name: hvst-upgrade-7xn6v
    namespace: harvester-system
    resourceVersion: "28811"
    uid: 50644a47-8e88-43ea-8e0f-90a335663b95
  spec:
    image: ""
    logEnabled: true
    version: v150-fixupglog-upgrade
  status: {}
kind: List
metadata:
  resourceVersion: ""
```

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -131,6 +131,12 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 	repo := NewUpgradeRepo(h.ctx, upgrade, h)
 
 	if harvesterv1.UpgradeCompleted.GetStatus(upgrade) == "" {
+		// this label is set by UI when user clicks `dismiss`, the upgrade object is kept but status is reset
+		if upgrade.Labels[util.LabelUpgradeReadMessage] == "true" {
+			logrus.Infof("upgrade %v/%v has %v label, no further processing", upgrade.Namespace, upgrade.Name, util.LabelUpgradeReadMessage)
+			return upgrade, nil
+		}
+
 		logrus.Infof("Initialize upgrade %s/%s", upgrade.Namespace, upgrade.Name)
 
 		if err := h.resetLatestUpgradeLabel(upgrade.Name); err != nil {


w13915984028:
@mergify backport v1.5

starbops (commented):
The fix LGTM. However, the Upgrade CR's status is still sabotaged and erroneous. We still need to find the root cause.

w13915984028:
Lets validate the UI fix https://github.com/harvester/harvester-ui-extension/pull/201 first.

starbops (commented):
IMO, it's good to have this check as the controller can short-circuit CRs with read-message labeled. Do we want to make it more general by moving the check outside of the `if` statement? Like, before the line
```
if harvesterv1.UpgradeCompleted.GetStatus(upgrade) == "" {
```

w13915984028:
@starbops The current upgrade status-machine it not designed to be short-cutable via `read-message`, unless the status is nil it can skip the reconciller safely, in other status, jump via `read-message` may cause un-expected results, as the sub/dependency deployments/plans/jobs/... need to be taken care of and cleaned up accordingly.

if an action like `pause` is desired, then a new development is needed.

starbops:
You are right. I was previously confused, and I just remembered that the controller does not even do anything about the `read-message` label's occurrence in the original design. So, I think we are good to close this PR then.

w13915984028:
UI PR https://github.com/harvester/harvester-ui-extension/pull/201 will be the right direction.

Issue #7798, opened 2025-03-09
[ENHANCEMENT] 3rd-party storage support enhancement after 1.5.0-rc1


Vicente-Cheng:
## What's the task? Please describe

We still need some minor enhancements.
This issue is used to track those items.

## Describe the sub-tasks

- [x] Enhance the volume creation with filesystem mode (for filesystem based 3rd-party storage provisioner) - need UI change @a110605 
- [x] Add validator to deine export the volumes from the 3rd-party storage to longhorn v1 Image (because backingimage did not support that) 
- [x] Some codes cleanup
- [x] Improve the golden image download with the filesystem storage provisioner
- [x] Improve upgrade handle
- [x] Improve LVM storageprofile
- [x] Improve the `WaitForFirstConsumer` volumeBindingMode

Vicente-Cheng:
use the same test case with 3rd-party storage

innobead:
Changed to kind/`enahncement`.

Vicente-Cheng:
Close. Please use the original one (https://github.com/harvester/harvester/issues/1199) to track.

Issue #7736, opened 2025-03-03
[Enhancement] Improve the Robustness for Whereabout CRD Upgrade


WebberHuang1118:
**Describe the bug**

Whereabouts had CRD upgrade for `v0.8.0` (Harvester `v1.5.0`), as @w13915984028 [suggested](https://github.com/harvester/harvester/pull/7405#pullrequestreview-2648348692), we create this ticket to track the enhancements:
- Preventing downgrade
- Checking CRD path existence

PR #7799, opened 2025-03-09
misc enhancement for 3rd-party storage support feature

Vicente-Cheng:

**Problem:**
Some minor issues of 3rd-party storage support features need to be fixed

**Solution:**
Improve the 3rd-party storage support features for filesystem storage provisioner

**Related Issue:**
https://github.com/harvester/harvester/issues/7798

**Test plan:**
please use the same test case for 3rd-party storage support features


deploy/charts/harvester/templates/cdi-harvester-crd.yaml
```diff
@@ -0,0 +1,49 @@
...
+    helm.sh/hook: post-install
+    helm.sh/hook-weight: "4"
+    helm.sh/hook-delete-policy: hook-succeeded
+spec:
+  template:
+    spec:
+      serviceAccountName: harvester
+      containers:
+      - name: apply-cr
+        image: rancher/kubectl:v1.29.2
```
bk201: We need to rely on values.yaml for future proof. what's the default image pull policy?

deploy/charts/harvester/templates/cdi-harvester-job.yaml
```diff
@@ -0,0 +1,48 @@
...
+    helm.sh/hook: post-install
+    helm.sh/hook-weight: "2"
+    helm.sh/hook-delete-policy: hook-succeeded
+spec:
+  template:
+    spec:
+      serviceAccountName: harvester
+      containers:
+      - name: wait-for-crd
+        image: rancher/kubectl:v1.29.2
```
bk201: ditto

deploy/charts/harvester/templates/cdi-harvester-job.yaml
```diff
@@ -0,0 +1,48 @@
...
+    helm.sh/hook: post-install
+    helm.sh/hook-weight: "1"
+    helm.sh/hook-delete-policy: hook-succeeded
+spec:
+  template:
+    spec:
+      serviceAccountName: harvester
+      containers:
+      - name: wait-for-cdi-api
+        image: rancher/kubectl:v1.29.2
```
bk201: ditto

deploy/charts/harvester/templates/cdi-harvester-crd.yaml
Vicente-Cheng: updated

pkg/webhook/resources/volumeimportsource/validator.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func NewValidator() types.Validator {
+	return &dataVolumeValidator{}
+}
+
+type dataVolumeValidator struct {
+	types.DefaultValidator
+}
+
```
WebberHuang1118: `dataVolumeValidator` should be `volumeImportSourceValidator`?

WebberHuang1118 (approved):
LGTM, just a nit, thanks.

pkg/webhook/resources/volumeimportsource/validator.go
Vicente-Cheng: nice catch... updated.

deploy/charts/harvester/templates/cdi-harvester-crd.yaml
```diff
@@ -0,0 +1,49 @@
...
+          - /manifests/volumeimportsource.yaml
+        volumeMounts:
+        - name: cr-manifest
+          mountPath: /manifests
+      restartPolicy: Never
+      volumes:
+      - name: cr-manifest
+        configMap:
+          name: volumeimportsource-cr
+{{- end -}}
```
ihcsim: Nit: Fix line break. Same for `cdi-harvester-job.yaml`.

deploy/charts/harvester/templates/cdi-harvester-job.yaml
```diff
@@ -0,0 +1,48 @@
...
+  annotations:
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-weight: "1"
+    helm.sh/hook-delete-policy: hook-succeeded
+spec:
+  template:
+    spec:
+      serviceAccountName: harvester
+      containers:
+      - name: wait-for-cdi-api
```
ihcsim: Nit: You can probably use one job with multiple containers, instead of multiple jobs with one container each. So one config map hook and one job hook.

deploy/charts/harvester/templates/cdi-harvester-crd.yaml
```diff
@@ -0,0 +1,49 @@
...
+      name: filesystem-blank-source
+      namespace: default
+    spec:
+      source:
+        blank: {}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: apply-volumeimportsource-cr
```
ihcsim: Any reasons why we don't put these jobs and the config map in the `harvester-system` namespace? I don't know if it's safe to assume the `default` namespace is always available in the user environment.

package/upgrade/upgrade_manifests.sh
```diff
@@ -1288,6 +1288,14 @@ apply_whereabouts_manifests()
       kubectl apply -f "$manifest"
   done
 
+  echo "Applying cdi manifests"
+
+  for manifest in /usr/local/share/extra_manifests/cdi/*.yaml; do
+      echo "Applying $manifest"
+      kubectl apply -f "$manifest"
+  done
```
ihcsim, line 1296: Nit: I think you can do `kubectl apply -f /usr/local/share/extra_manifests/cdi/*.yaml`.

pkg/controller/master/vmimagedownloader/register.go
```diff
@@ -3,38 +3,36 @@ package vmimagedownloader
...
 		vmImageClient:               vmImage,
+		pvcCache:                    pvcCache,
 		deploymentClient:            deployment,
 		vmImageDownloaders:          vmImageDownloader,
 		vmImageDownloaderController: vmImageDownloader,
 	}
 
 	vmImageDownloader.OnChange(ctx, vmImageDownloaderControllerName, storageProfileHandler.OnChanged)
 	vmImageDownloader.OnRemove(ctx, vmImageDownloaderControllerName, storageProfileHandler.OnRemoved)
+	relatedresource.Watch(ctx, deploymentWatcherName, storageProfileHandler.ReconcileDeploymentOwners, vmImageDownloader, deployment)
```
ihcsim, line 35: To confirm my understanding of this `Watch`, every time the deployment changes, it enqueues the owner's VM image downloader, which will in turn, triggers `OnChanged()`?

pkg/webhook/resources/persistentvolumeclaim/mutator.go
```diff
@@ -47,32 +48,86 @@ func (m *pvcMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patc
...
 	}
 
 	annotations := pvc.GetAnnotations()
 
-	if _, find := annotations[util.AnnotationGoldenImage]; find {
-		if annotations[util.AnnotationGoldenImage] == "true" {
-			return nil, nil
-		}
+	if v, find := annotations[util.AnnotationGoldenImage]; find && v == "true" {
+		return "", nil
```
ihcsim: This will trigger `m.patchDataSource(pvc)` later. Do we need data source for the golden image?

pkg/webhook/resources/persistentvolumeclaim/mutator.go
```diff
@@ -47,32 +48,86 @@ func (m *pvcMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patc
...
+	if v, find := annotations[util.AnnotationVolForVM]; !find || v != "true" {
+		logrus.Debugf("PVC %s/%s is not for VM, skip patch", pvc.Namespace, pvc.Name)
+		return "", nil
+	}
+
+	cdiAPIGroup := cdicommon.AnnAPIGroup
+	dataSourceRef := corev1.TypedLocalObjectReference{
+		APIGroup: &cdiAPIGroup,
+		Kind:     util.KindVolumeImportSource,
+		Name:     util.ImportSrourceBlank,
```
ihcsim:
```suggestion
		Name:     util.ImportSourceBlank,
```

pkg/webhook/resources/volumeimportsource/validator.go
```diff
@@ -0,0 +1,69 @@
...
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+
+	werror "github.com/harvester/harvester/pkg/webhook/error"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+const (
+	BlankSourceName = "filesystem-blank-source"
```
ihcsim: Use `util.ImportSrourceBlank`?

ihcsim (approved):
LGTM - just some questions and minor suggestions.

deploy/charts/harvester/templates/cdi-harvester-job.yaml
ihcsim: FWIW, I don't think the config map is even necessary. In your job container, you should be able to run:

```
spec:
    containers:
    - # ...
      command:
      - |
        cat <<EOF | k apply -f -
        kind: VolumeImportSource
        metadata:
          name: filesystem-blank-source
          namespace: default
        spec:
          source:
            blank: {}
```

Vicente-Cheng:
Hello guys, please focus on the https://github.com/harvester/harvester/pull/7799/commits/08fe7f7b6aa6fb98edf6a1fa768eef6057bbd2b1 if you have already reviewed others.

Because the dataSourceRef need the same namespace, we need to create volumeImportSource on-demand.

deploy/charts/harvester/templates/cdi-harvester-crd.yaml
Vicente-Cheng: removed this file, please check the reason on following NS question

deploy/charts/harvester/templates/cdi-harvester-job.yaml
Vicente-Cheng: ditto, please check below NS question

deploy/charts/harvester/templates/cdi-harvester-crd.yaml
Vicente-Cheng: I thought the default namespace could allow all user use it as dataSourceRef. But looks like it must use the same namespace for the dataSourceRef. For example, if you create the PVC with namespace `longhorn-system`, you need to volumeImportSource in the `longhorn-system`. So I changed the mechanism to create volumeImportSource on demand.

Please check the commit https://github.com/harvester/harvester/pull/7799/commits/08fe7f7b6aa6fb98edf6a1fa768eef6057bbd2b1

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: I would like to keep that. 
Once we have several yaml files to update, we could know which we are working on.

pkg/controller/master/vmimagedownloader/register.go
Vicente-Cheng: Yes, you were correct. We will check the parent source and enqueues it if needed.

pkg/webhook/resources/persistentvolumeclaim/mutator.go
Vicente-Cheng: No, this is used for gold images. That path uses the datavolume instead of the populator.

pkg/webhook/resources/volumeimportsource/validator.go
Vicente-Cheng: I remove this file because the volumeImportSource now created by controller.

pkg/webhook/resources/persistentvolumeclaim/mutator.go
Vicente-Cheng: I updated this Name with the on-demand creation.

WebberHuang1118 (approved):
LGTM, thanks.

bk201 (approved):

pkg/image/cdi/common.go
```diff
@@ -93,6 +92,16 @@ func generateDVSourceHTTP(vmi *harvesterv1.VirtualMachineImage) *cdiv1.DataVolum
 	}
 }
 
+func GenerateDVAnnotations(sc *storagev1.StorageClass) map[string]string {
+	annotations := map[string]string{}
+	volBindingMode := *sc.VolumeBindingMode
```
w13915984028, line 102: need to check if VolumeBindingMode is not nil first
```
	// volumeBindingMode indicates how PersistentVolumeClaims should be
	// provisioned and bound.  When unset, VolumeBindingImmediate is used.
	// This field is only honored by servers that enable the VolumeScheduling feature.
	// +optional
	VolumeBindingMode *VolumeBindingMode `json:"volumeBindingMode,omitempty" protobuf:"bytes,7,opt,name=volumeBindingMode"`
```

package/upgrade/upgrade_manifests.sh
```diff
@@ -1277,7 +1277,7 @@ EOF
   fi
 }
 
-apply_whereabouts_manifests()
+apply_extra_nonversion_manifests()
```
w13915984028, line 1280: here strict version control is required, to avoid downgrade CRDs accidentally

package/upgrade/upgrade_manifests.sh
```diff
@@ -1288,6 +1288,14 @@ apply_whereabouts_manifests()
       kubectl apply -f "$manifest"
   done
 
+  echo "Applying cdi manifests"
+
+  for manifest in /usr/local/share/extra_manifests/cdi/*.yaml; do
```
w13915984028, line 1293: upgrade pod may run several times, the better solution is to check if CRD is existing, only apply when none-existing

pkg/controller/master/pvc/pvc_controller.go
```diff
@@ -1,21 +1,62 @@
...
+		return pvc, nil
+	}
+
+	annotations := pvc.GetAnnotations()
+	if v, find := annotations[util.AnnotationVolForVM]; !find || v != "true" {
+		logrus.Debugf("PVC %s/%s is not for VM, skip checking filesystem blank", pvc.Namespace, pvc.Name)
+		return pvc, nil
+	}
+
+	targetPopulatorName := fmt.Sprintf("filesystem-blank-source-%s", pvc.Namespace)
```
w13915984028: `filesystem-blank-source` is used in both controller and webhook, define it as a const?

pkg/controller/master/pvc/pvc_controller.go
```diff
@@ -1,21 +1,62 @@
...
+	if _, err := h.volImportSourceClient.Get(pvc.Namespace, targetPopulatorName, metav1.GetOptions{}); err != nil {
+		if apierrors.IsNotFound(err) {
+			blankSource := cdiv1.VolumeImportSource{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      targetPopulatorName,
```
w13915984028: the object name has 253 chars limitation, could targetPopulatorName exceed it?

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -94,8 +96,11 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
 		// create DataVolume for PVC (only for CDI backend)
 		if createPVCWithDataVolume {
 			if _, err := h.dataVolumeClient.Get(vm.Namespace, pvcAnno.Name, metav1.GetOptions{}); apierrors.IsNotFound(err) {
-				targetSCName := *pvcAnno.Spec.StorageClassName
-				if err := h.createDataVolume(pvcAnno.Name, vm.Namespace, imageName, imageNS, targetSCName, pvcAnno.Spec.Resources); err != nil {
+				targetSC, err := h.scCache.Get(*pvcAnno.Spec.StorageClassName)
```
w13915984028: In theory, `StorageClassName` could be nil

```
type PersistentVolumeClaimSpec struct {
...
	// storageClassName is the name of the StorageClass required by the claim.
	// More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#class-1
	// +optional
	StorageClassName *string `json:"storageClassName,omitempty" protobuf:"bytes,5,opt,name=storageClassName"`
```

pkg/controller/master/vmimagedownloader/vmimagedownloader_controller.go
```diff
@@ -281,6 +284,61 @@ func (h *vmImageDownloaderHandler) getClusterRepoImage() string {
...
+				DevicePath: targetImgVolPath,
+			},
+		}
+	} else if *mode == corev1.PersistentVolumeFilesystem {
+		initContainer.VolumeMounts = append(initContainer.VolumeMounts, corev1.VolumeMount{
+			Name:      "image-vol",
+			MountPath: targetImgVolPath,
+		})
+	}
+	convertCmd := fmt.Sprintf("qemu-img convert -t none -T none -W -m 8 -f raw %s -O qcow2 -c -S 4K /image-dir/%s.qcow2", convertSrcPath, vmImageName)
```
w13915984028, line 336: any command to check if the image has already met format, instead of alway run convert ?

w13915984028 (commented):
some questions, thanks.

pkg/controller/master/pvc/pvc_controller.go
```diff
@@ -1,21 +1,62 @@
...
+	pvcController         ctlcorev1.PersistentVolumeClaimController
+}
+
+func (h *pvcHandler) createFilesystemBlankSource(_ string, pvc *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+	if pvc == nil || pvc.Spec.VolumeMode == nil || *pvc.Spec.VolumeMode != corev1.PersistentVolumeFilesystem {
+		return pvc, nil
+	}
+
+	annotations := pvc.GetAnnotations()
+	if v, find := annotations[util.AnnotationVolForVM]; !find || v != "true" {
```
ihcsim, line 30: When does this annotation get added to the PVC? I can't find it in the code:

```
$ grep -rniI annotationvolforvm pkg
pkg/controller/master/pvc/pvc_controller.go:32: if v, find := annotations[util.AnnotationVolForVM]; !find || v != "true" {
pkg/util/constants.go:23:       AnnotationVolForVM                  = prefix + "/volumeForVirtualMachine"
pkg/webhook/resources/persistentvolumeclaim/mutator.go:113:     if v, find := annotations[util.AnnotationVolForVM]; !find || v != "true" {
```

pkg/controller/master/pvc/pvc_controller.go
ihcsim: The namespace suffix is unnecessary since `VolumeImportSource` is already scoped to the namespace. That will make sure we don't violate the naming chars limitation per @w13915984028.

pkg/webhook/resources/persistentvolumeclaim/mutator.go
ihcsim: To confirm, returning an empty string here will set `isGoldenImage` to `false`, even though the PVC is a golden image. Then the PVC will be patched with the `blank` data source. I guess I am not sure whether the golden image is supposed to have the `blank` data source or not.

pkg/webhook/resources/persistentvolumeclaim/mutator.go
```diff
@@ -47,32 +48,86 @@ func (m *pvcMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patc
...
 
 	annoVal, err := json.Marshal(annotations)
 	if err != nil {
 		logrus.Warnf("failed to marshal annotations: %v, err: %v", annotations, err)
-		return nil, err
+		return "", err
+	}
+
+	// patch annotation
+	return fmt.Sprintf(`{"op": "replace", "path": "/metadata/annotations", "value": %s}`, string(annoVal)), nil
```
ihcsim: Assuming that K8s call all mutators sequentially, this should be race free. Otherwise, we should append our annotation to the end of the list. See example in the [JSON patch RFC](https://datatracker.ietf.org/doc/html/rfc6902#appendix-A.16).

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: I thought the CRD we updated here was the non-handled CRD (which usually happened on the upgrade). I thought we would not encounter the downgrade issue here.
cc @starbops, @WebberHuang1118 WDYT?

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: If the CRD is the same, duplicated applies did not bring up any changes.
So, should we focus on the above discussion (for downgrade part)

pkg/controller/master/vmimagedownloader/vmimagedownloader_controller.go
Vicente-Cheng: We could ensure the source is in the raw format because this is created ourselves.
The source is the 3rd-party golden image pvc. This is the raw format.

pkg/controller/master/pvc/pvc_controller.go
Vicente-Cheng: This annotation only applied from UI.
The only case in which we need this annotation is when UI creates the filesystem volume for VM.

pkg/webhook/resources/persistentvolumeclaim/mutator.go
Vicente-Cheng: I am not sure what you mean.
I get the whole annotation and add (or do no-op) with the current PVC annotation.
So, it should not break any annotation.
Could you explain more about this?

pkg/webhook/resources/persistentvolumeclaim/mutator.go
Vicente-Cheng: In theory, it should not have happened because we only do this on the create. And the goldenImage is also controlled by us.
But you were correct: it will happen if users add the goldenImage flag themselves.
I will enhance this part.

pkg/webhook/resources/persistentvolumeclaim/mutator.go
ihcsim: Say if another CSI provider has a mutator that also modifies the annotations via JSON patch. If our mutator doesn't see these modified annotations, then when the K8s API Server sees this `replace` JSON patch, it will replace the whole annotations with what we submitted here, potentially, overriding the modified annotations from the other CSI provider. It has been a while since I last looked at K8s mutator invocation code, but I think they are called sequentially, meaning, it _should be_ safe to do a `replace` op here.

ihcsim (approved):

package/upgrade/upgrade_manifests.sh
w13915984028: The concern comes from the long-time version management, today we all remember the cdi has been put on 2 places. Later, we or others may forget to bump the two places in parallel (surely, with lower chance but it could happen), and it may took additional efforts to figure them out.

```
harvester version              v1.5        v1.6   v1.7   v1.8

upgrade path                   v1.0        v1.0   v1.0   ?
harvester chart path           v1.0        v1.0   v1.1   v1.2
```

w13915984028: On other parts, files are managed per version, e.g.

https://github.com/harvester/harvester/blob/ec25120bade7191e583a8298c600b1a803112824/package/upgrade/migrations/upgrade_manifests/v1.4.0/pre-hook.sh#L3

package/upgrade/upgrade_manifests.sh
w13915984028: the performance penalty

like in controller, usally call `reflect.DeepEqual` to compare, before really post a request to apiserver

pkg/webhook/resources/persistentvolumeclaim/mutator.go
Vicente-Cheng: Thanks for the detailed explanation. I know your concern, and now I only patch the annotation key we wanted.

pkg/controller/master/pvc/pvc_controller.go
Vicente-Cheng: udpated

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: thanks @w13915984028, 
I improved this. Now, if the StorageClassName is nil, I will try to find the default storageclass.

pkg/controller/master/pvc/pvc_controller.go
Vicente-Cheng: Improve that w/o namespace

pkg/controller/master/pvc/pvc_controller.go
Vicente-Cheng: updated

pkg/image/cdi/common.go
Vicente-Cheng: updated. The default volumeBindingMode is immediate. We could do nothing about this.

w13915984028 (approved):
LGTM, thanks for the confirmation and fix.

The version based crd file management can be improved via https://github.com/harvester/harvester/issues/7736 later.

pkg/api/volume/handler.go
```diff
@@ -89,19 +95,65 @@ func (h *ActionHandler) Do(ctx *harvesterServer.Ctx) (interface{}, error) {
 	}
 }
 
-func (h *ActionHandler) exportVolume(_ context.Context, imageNamespace, imageDisplayName, imageStorageClassName, pvcNamespace, pvcName string) (interface{}, error) {
+func (h *ActionHandler) validateExportVolume(storageClassName, pvcNamespace, pvcName string) string {
```
ibrokethecloud, line 98: minor nit, why not make the method return an error and return nil when conditions are met

pkg/controller/master/pvc/pvc_controller.go
```diff
@@ -6,16 +6,54 @@ import (
...
+		logrus.Debugf("PVC %s/%s is not for VM, skip checking filesystem blank", pvc.Namespace, pvc.Name)
+		return pvc, nil
+	}
+
+	if _, err := h.volImportSourceClient.Get(pvc.Namespace, util.ImportSourceFSBlank, metav1.GetOptions{}); err != nil {
```
ibrokethecloud, line 35: the object lookup and subsequent creation is using a constant value for name. Is this expected?

Vicente-Cheng: Yeah, this is the specific populator for blank filesystem PVC.
(used for creating the `disk.img` file for VM usage)

pkg/api/volume/handler.go
Vicente-Cheng: I am ok with using error for this function.
The reason to return string because the caller will return the error w string.
So, it is straightforward to return the string directly.

ibrokethecloud (approved):
lgtm. thanks for the enhancements.

pkg/controller/master/vmimagedownloader/register.go
```diff
@@ -3,38 +3,36 @@ package vmimagedownloader
...
 
-	deploymentHandler := &deploymentHandler{
-		vmImageDownloaderClient: vmImageDownloader,
-		deploymentCache:         deployment.Cache(),
-		deploymentController:    deployment,
-	}
-
-	deployment.OnChange(ctx, deploymentControllerName, deploymentHandler.OnChanged)
-	deployment.OnRemove(ctx, deploymentControllerName, deploymentHandler.OnRemoved)
```
w13915984028, line 38: Just for referrences:

The `deployment.OnRemove` also caused another issue observed when upgarding from v150-rc1->v150-rc2. Luckily, this PR deleted it.

the possible cause is: when managedchart is removed, the rbac, deployments... are deleted in a row;  when rbac went first but upgradelog deployment later, the finalizer can't work, then it stuck there

api-server log:
```
2025-03-27T02:42:13.968264952Z E0327 02:42:13.968141       1 authentication.go:74] "Unable to authenticate the request" err="[invalid bearer token, serviceaccounts \"hvst-upgrade-qvgbd-upgradelog-operator-rancher-logging\" not found]"
2025-03-27T02:42:14.086819891Z E0327 02:42:14.086660       1 authentication.go:74] "Unable to authenticate the request" err="[invalid bearer token, serviceaccounts \"hvst-upgrade-qvgbd-upgradelog-operator-rancher-logging\" not found]"
```
upgrade operator deployment:
    
```
- apiVersion: apps/v1
  kind: Deployment
  metadata:
    annotations:
      deployment.kubernetes.io/revision: "1"
      meta.helm.sh/release-name: hvst-upgrade-qvgbd-upgradelog-operator
      meta.helm.sh/release-namespace: cattle-logging-system
      objectset.rio.cattle.io/id: default-mcc-hvst-upgrade-qvgbd-upgradelog-operator-cattle-4729b
    creationTimestamp: "2025-03-26T22:46:49Z"
    deletionGracePeriodSeconds: 0
    deletionTimestamp: "2025-03-27T02:42:13Z" // 
    finalizers:
    - wrangler.cattle.io/deployment-controller  // blocked by it
    generation: 3
    labels:
      app.kubernetes.io/instance: hvst-upgrade-qvgbd-upgradelog-operator
      app.kubernetes.io/name: rancher-logging
      app.kubernetes.io/version: 4.10.0
      helm.sh/chart: rancher-logging-105.2.0_up4.10.0
      objectset.rio.cattle.io/hash: 2dc25db8c30ae8db473eb6c47b7c4dbc0b7b5ad3
    managedFields:
```

Issue #7381, opened 2025-01-16
[FEATURE] Support Network policies for micro segmentation of VMs using KubeOVN


rrajendran17:
We could achieve VM isolation using virtual switch concept in KubeOVN.
But implementing network policies facilitate further micro segmentation of VMs by isolating VMs within the same L2/Virtual Switch.

**Describe the solution you'd like**
a.KubeOVN makes use of Kubernetes Network Policy CRD to implement ACL rules allowing/restricting traffic between VMs.

(Or)
b.KubeOVN subnet acl can be used to isolate traffic between VMs within the same subnet.

**Describe alternatives you've considered**
OVN kubernetes.Ovn kubernetes uses MultiNetworkPolicy CRD to implement network policies per vswitch.

**Additional context**
Network policy yaml example:

apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: ip-block
spec:
  podSelector:
    matchLabels:
      name: access-control
  ingress:
    - from:
        - ipBlock:
            cidr: 10.10.0.0/16
  egress:
    - to:
        - ipBlock:
            cidr: 20.20.0.0/16

VMs could be labeled and used in network policy, so that the ACL rule is applied specific to those VMs.
ingress and egress rules allow traffic to and from particular ip block and rest of the traffic is dropped.
Using above VMs can be isolated within the same L2, by specifying the VMs and IPs which has to communicate.

subnet acl example:

apiVersion: kubeovn.io/v1
kind: Subnet
metadata:
  name: acl
spec:
  allowEWTraffic: false
  acls:
    - action: drop
      direction: to-lport
      match: ip4.dst == 10.10.0.2 && ip
      priority: 1002
    - action: allow-related
      direction: from-lport
      match: ip4.src == 10.10.0.3 && ip
      priority: 1002
  cidrBlock: 10.10.0.0/24

The above allows traffic from 10.10.0.3, but blocks traffic from 10.10.0.2

rrajendran17:
The following could be used to filter traffic per vswitch using acls added to subnet.
https://kubeovn.github.io/docs/stable/en/guide/subnet/#subnet-acl

rrajendran17:
Need for micro segmentation between VMs:

```
VM isolation could be achieved using VLANs traditionally or using virtual switches in KubeOVN.
Implementing network policies allow micro segmentation of VMs by isolating VMs within 
the same L2 network/Virtual Switch.
```

Proposal:
```
[...]

PR #7812, opened 2025-03-11
Support Network Policy in Harvester using kubeovn

rrajendran17:
**Problem:**
We could achieve VM isolation using virtual switch concept in KubeOVN.
But implementing network policies facilitate further micro segmentation of VMs by isolating VMs within the same L2/Virtual Switch.

**Solution:**
KubeOVN makes use of Kubernetes Network Policy CRD to implement ACL rules allowing/restricting traffic between VMs.
Subnet ACLs could be used to implement rules within subnet used by VMs

**Related Issue:**
https://github.com/harvester/harvester/issues/7381

**Test plan:**

### DIFF
--- a/enhancements/20250310-NetworkPolicy.md
+++ b/enhancements/20250310-NetworkPolicy.md
@@ -1,0 +1,165 @@
+# Support for Network Policies in Harvester
+
+Support Network Policies in Harvester using KubeOVN for microsegmentation of VM workloads
+
+## Summary
+
+- Traffic Isolation:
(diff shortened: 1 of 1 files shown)


starbops:
Please sign off the commits and perhaps add a `[skip ci]` suffix to the commit message to avoid triggering the CI for doc-only PRs. Thank you.

rrajendran17:
Added details to GH https://github.com/harvester/harvester/issues/7381
we do not require HEP

Issue #7856, opened 2025-03-14
[BUG] Reconnecting the Remote Target will not Discover Existing VMBackup Automatically


WebberHuang1118:
**Describe the bug**
Reconnecting the remote target will not discover the existing VMBackup 

**To Reproduce**
1. Creating a new VM `vm1`, and taking a VMBackup `vm1-b`
2. Disconnecting backup target
3. Deleting `vm1-b`
4. Reconnecting backup target
5. `vm1-b` is not discovered automatically

**Expected behavior**
`vm1-b` should be discovered automatically

**Environment**
 - Harvester ISO version: Harvester master branch `f0718cec621006688d051faad652453f3b0e4a3b`

**Additional context**
- Finding this issue when @m-ildefons had verification on #7539
- Setting backup target's annotation `harvesterhci.io/hash` is not updated while setting it back to default, if `Refresh Interval` is zero from the beginning, we will not try to recover existing VMBackups
  - https://github.com/harvester/harvester/blob/f0718cec621006688d051faad652453f3b0e4a3b/pkg/controller/master/backup/backup_metadata.go#L125-L128
  - https://github.com/harvester/harvester/blob/f0718cec621006688d051faad652453f3b0e4a3b/pkg/controller/master/backup/backup_metadata.go#L170-L178

cc @bk201 @FrankYang0529

votdev:
Verified this bug has been fixed.

Test Information
----
* Environment: **ipxe-examples/vagrant-pxe-harvester 1 node**
* Harvester Version: **v1.5.x master-4c350fda-head**

### Verify Steps
1. Create a VM `vm1`.
2. Go to `Advanced > Settings > backup-target`. Configure the NFS server. Press `Save`.
3. Check the settings YAML. The `harvesterhci.io/hash` and `harvesterhci.io/lastRefreshTime` annotation must be set.
```
$ k get settings backup-target -oyaml
apiVersion: harvesterhci.io/v1beta1
kind: Setting
metadata:
  annotations:
    harvesterhci.io/hash: cfe4f7ee25e4e2806fbf4d478ecd7ea7923518255ec1588bab2c666c
    harvesterhci.io/lastRefreshTime: "2025-03-18T08:27:25Z"
  creationTimestamp: "2025-03-18T06:50:17Z"
  generation: 6
  name: backup-target
  resourceVersion: "74050"
  uid: af48a191-36e5-4bca-989b-1b0cf1e3c365
status:
  conditions:
  - status: "True"
[...]

PR #7857, opened 2025-03-14
feat: remove backup target annotation if it's empty

FrankYang0529:

**Problem:**
When reset to empty backup target, we didn't clean up hash annotation. If user sets to a same backup target again and doesn't set refresh interval, the backup metadata will not refresh.

**Solution:**
Cleanup hash annotation if backup target is empty.

**Related Issue:**
https://github.com/harvester/harvester/issues/7856

**Test plan:**

1. Creating a new VM vm1, and taking a VMBackup vm1-b
2. Set backup target to default value. Check backup target annotation is empty
3. Deleting vm1-b
4. Set backup target again.
5. Check harvester log. It will try to rebuild vm1-b CR in the cluster. If CR is not rebuilt, users should see message like following:
```
level=warning msg="skip creating vm backup, because the longhorn backup is not found in the cluster" error="backups.longhorn.io \"backup-47c51e7bc8a64a98\" not found" longhornBackup=backup-47c51e7bc8a64a98 name=b1 namespace=default
```

### DIFF
--- a/pkg/controller/master/backup/backup_metadata.go
+++ b/pkg/controller/master/backup/backup_metadata.go
@@ -113,8 +113,11 @@ func RegisterBackupMetadata(ctx context.Context, management *config.Management, 
 // OnBackupTargetChange resync vm metadata files when backup target change
 func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
 	if setting == nil || setting.DeletionTimestamp != nil ||
-		setting.Name != settings.BackupTargetSettingName || setting.Value == "" {
+		setting.Name != settings.BackupTargetSettingName {
 		return nil, nil
+	}
+	if setting.Value == "" {
+		return h.resetBackupTarget(setting)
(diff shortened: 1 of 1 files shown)


WebberHuang1118 (approved):
LGTM, thanks

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -180,6 +180,20 @@ func (h *MetadataHandler) shouldRefresh(setting *harvesterv1.Setting, refreshInt
 	return true
 }
 
+func (h *MetadataHandler) resetBackupTarget(setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	settingCopy := setting.DeepCopy()
+	if settingCopy.Annotations == nil {
+		settingCopy.Annotations = map[string]string{}
```
ihcsim: Can return if annotation map is `nil`?

ihcsim (approved):
LGTM.

FrankYang0529:
@mergify backport v1.5

Issue #7860, opened 2025-03-14
[BUG] Remove the sepcial test code on upgradeLog controller


w13915984028:
**Describe the bug**

PR: https://github.com/harvester/harvester/pull/7772
A block of temparary code is added to upgradeLog controller to keep the whole test code running, but the related helm value get needs to be moved to test code.

https://github.com/harvester/harvester/blob/beb2d8d3258b474ecc866de8614d0788eb5f6360/pkg/controller/master/upgradelog/controller.go#L764

We did not figure out an easy mock solution to simulate `helm chart & version release (via secret)`.

Given the timeline of v1.5.0, the code is merged and we will solve this issue later.

```
func (h *handler) getConsolidatedLoggingImageListFromHelmValues(namespace, name string) (map[string]settings.Image, error) {
	images := make(map[string]settings.Image, len(loggingImagesList))

	// for the convenience of test
	if h.clientset == nil {
		images := map[string]settings.Image{
			imageConfigReloader: {
				Repository: "rancher/config-reload",
				Tag:        "default",
			},
			imageFluentbit: {
				Repository: "rancher/fluentbit",
				Tag:        "dev",
			},
			imageFluentd: {
				Repository: "test/fluentd",
				Tag:        "dev",
			},
		}
		return images, nil
	}

...
```

**To Reproduce**

**Expected behavior**

 Move the related helm value get needs to test code.

**Support bundle**

**Environment**
 - Harvester ISO version: v1.5.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

w13915984028:
Related helm code:

```
// Release describes a deployment of a chart, together with the chart
// and the variables used to deploy that chart.
type Release struct {
	// Name is the name of the release
	Name string `json:"name,omitempty"`
	// Info provides information about a release
	Info *Info `json:"info,omitempty"`
	// Chart is the chart that was released.
	Chart *chart.Chart `json:"chart,omitempty"`
	// Config is the set of extra Values added to the chart.
	// These values override the default values inside of the chart.
	Config map[string]interface{} `json:"config,omitempty"`
	// Manifest is the string representation of the rendered template.
	Manifest string `json:"manifest,omitempty"`
	// Hooks are all of the hooks declared for this release.
	Hooks []*Hook `json:"hooks,omitempty"`
	// Version is an int which represents the revision of the release.
	Version int `json:"version,omitempty"`
[...]

PR #7660, opened 2025-02-19
Patch upgradeLog on the upgrade path

w13915984028:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**

The upstream logging has some enhancements, the upgradeLog needs to adapt to them.

**Solution:**

This PR and https://github.com/harvester/harvester/pull/7653 enhance the upgradeLog in v1.5.0

**Related Issue:**
https://github.com/harvester/harvester/issues/7652
https://github.com/harvester/harvester/issues/7654

**Test plan:**

Upgrade v1.4.*->v1.5.0, ugpradeLog should work properly, no matter rancher-logging addon is enabled or not

local test log:
```
run the same test script several times

...
patch clusteroutput hvst-upgrade-b2f59-upgradelog-clusteroutput
clusteroutput.logging.banzaicloud.io/hvst-upgrade-b2f59-upgradelog-clusteroutput patched (no change)
patch clusterflow hvst-upgrade-b2f59-upgradelog-clusterflow
clusterflow.logging.banzaicloud.io/hvst-upgrade-b2f59-upgradelog-clusterflow patched (no change)
patch logging hvst-upgrade-b2f59-upgradelog-infra
logging.logging.banzaicloud.io/hvst-upgrade-b2f59-upgradelog-infra patched (no change)
patch logging hvst-upgrade-b2f59-upgradelog-operator-root
logging.logging.banzaicloud.io/hvst-upgrade-b2f59-upgradelog-operator-root patched


NAME                                          LOGGINGREF                           CONTROLNAMESPACE        WATCHNAMESPACES   PROBLEMS
hvst-upgrade-b2f59-upgradelog-infra           harvester-upgradelog                 harvester-system                          
hvst-upgrade-b2f59-upgradelog-operator-root   harvester-upgradelog-operator-root   cattle-logging-system            

```

v141->v150 upgrade test, with addon rancher-logging disabled, the upgradeLog works well

image

another test with updated code:
```
$ kubectl get upgrade.harvesterhci -A
NAMESPACE          NAME                 AGE
harvester-system   hvst-upgrade-ttggs   5m46s

$ kubectl get upgradelog -A
NAMESPACE          NAME                            UPGRADE
harvester-system   hvst-upgrade-ttggs-upgradelog   hvst-upgrade-ttggs



$ ./patch.sh 
patch clusteroutput hvst-upgrade-ttggs-upgradelog-clusteroutput
[...]

PR #7862, opened 2025-03-14
Remove upgradelog test code from controller

w13915984028:

**Problem:**

A block of test code was added to controller for the convenience of test code.

**Solution:**

It is not easy to simulate a helm chart and release, instead, the imageGetter return the wanted images directly.

Use ImageGetterInterface to abstract and consolidate

**Related Issue:**
https://github.com/harvester/harvester/issues/7860

**Test plan:**

1. go test on upgrade controller runs successfully
2. upgrade test on real cluster, is done together with PR https://github.com/harvester/harvester/pull/7660#issuecomment-2729515719

Thanks the suggestion from @m-ildefons


starbops (approved):

w13915984028:
@mergify backport v1.5

m-ildefons (commented):
LGTM.

m-ildefons (approved):

Issue #7876, opened 2025-03-17
[BUG] Error message when removing longhorn v2 disks after disabling


noahgildersleeve:
**Describe the bug**

When you try to remove disks from a host that are provisioned with Longhorn v2 after disabling the addon you get an error. It seems to hang until you re-enable the addon
**To Reproduce**
1. Enable Longhorn v2 in advanced settings
2. Add a disk to a host provisioned with LH v2
3. Wait for it to enable
4. I did some operations and VMs that were later deleted, but I"m not sure if that's necessary
5. Disable LH v2
6. Try to remove the disks from the hosts via edit settings in the hosts page

**Expected behavior**

You should be able to delete the disks

**Support bundle**

[supportbundle_6730e26e-199a-4ae3-ab37-4ba66374cd74_2025-03-17T19-59-51Z.zip](https://github.com/user-attachments/files/19297547/supportbundle_6730e26e-199a-4ae3-ab37-4ba66374cd74_2025-03-17T19-59-51Z.zip)

**Environment**
 - Harvester ISO version: v1.5.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Bare metal DL360 2 node cluster

**Additional context**

Found while testing #1199 
![Image](https://github.com/user-attachments/assets/82e1c5b1-dc5a-4c21-af51-090702fdeb19)

### Workaround
1. Enable Longhorn v2 again in advanced settings

Vicente-Cheng:
We should remove the whole disks for the v2 data engine before we disable the v2 data engine.

The reason @derekbit explained as below:
```
Removing a block-type disk involves the unbind operation, and the unbind operation is executed in an IM pod for v2 volume. Thus, users cannot disable v2 data engine first.
```

So, we need to add the webhook validator for it!

votdev:
Failed to fully verify this issue as solved.

### Setup
* Environment: **ipxe-examples/vagrant-pxe-harvester 1 nodes**
* Harvester Version: **master-4713cdcc-head**

### Verify Steps
1. 🟢 Enable Longhorn v2 in advanced settings
2. 🟢 Add a disk to a host provisioned with LH v2
3. 🟢 Wait for it to enable
4. 🟢 I did some operations and VMs that were later deleted, but I"m not sure if that's necessary
5. 🔴 Disable LH v2
6. 🔴 Try to remove the disks from the hosts via edit settings in the hosts page

[...]

Issue #7875, opened 2025-03-17
[Enhancement] UI should able to config the volumeMode when add the new volume


Vicente-Cheng:
## What's the task? Please describe

When we create the VM and try to add volume.
UI should be able to configure the volumeMode since we introduce the 3rd-party storage support.
For example, like the following figure, when we use the nfs-csi StorageClass, we should able to set the volumeMode to `Filesystem` instead of the `Block`

![Image](https://github.com/user-attachments/assets/a97b21c3-fa1b-4d0d-80d8-90af6ded2d1f)

cc @a110605 

## Describe the sub-tasks

- [x] Backend needs to add the populator to add volume because it created on demand - @Vicente-Cheng 
- [x] UI needs to add the volumeMode options (Like volume page)

ihcsim:
@Vicente-Cheng @a110605 this change will also need to be added to the image creation page. @clemenko
reported this on [slack](https://rancher-users.slack.com/archives/C01GKHKAG0K/p1743796075229099).

clemenko:
It would be amazing to get the image creation process the ability to set `volumeMode: Block`. The Portworx CSI needs this mode for RWX.

Vicente-Cheng:
The VMImage did not have any volumeMode concept. The volumeMode concept is linked to the corresponding StorageClass.
We don't need to bring this concept to VM Image. It will make things more complex.

If we were to use the 3rd-party storage feature, sometimes we might need to update the StorageProfile ourselves.
(REF: https://github.com/kubevirt/containerized-data-importer/blob/main/doc/storageprofile.md)

Because not every provisioner of CSI is defined in the CDI, we need to update it manually. 
The storageprofile will use the storageclass name for the PVC related to, so you can update the corresponding storageprofile.
For example, if we want to use Block as volumeMode, we can do that.
```
apiVersion: cdi.kubevirt.io/v1beta1
kind: StorageProfile
metadata:
  ...
  name: <your target StorageClass>
spec:
  claimPropertySets:
  - accessModes:
    - ReadWriteMany
    volumeMode: Block
```

We might redesign the UX experience for this after v1.5.0, but currently, we will keep the same flow about VM Image.

clemenko:
I tried updating and it immediately switched back. The PVC overrides this functionality. This is why I am asking for it here.

clemenko:
this is a known issue for the csi I am using https://docs.portworx.com/portworx-csi/release-notes.html#known-issues

[...]

Issue #7854, opened 2025-03-14
[BUG] Can't export volume to image with Longhorn v2 external storage


noahgildersleeve:
**Describe the bug**

When you try to export a volume to an image with Longhorn v2 it hangs at 0 percent exporting and never progresses
**To Reproduce**
1. Enable Longhorn v2 on cluster
2. Create a storage class with LH v2
3. Create a VM with the storage class
4. Export the volume to image with the storage LH v2 storage class selected
**Expected behavior**

Export should complete successfully.
**Support bundle**

[supportbundle_2f257f57-706b-459c-bd34-e0ca8e0c771d_2025-03-14T00-23-42Z.zip](https://github.com/user-attachments/files/19238625/supportbundle_2f257f57-706b-459c-bd34-e0ca8e0c771d_2025-03-14T00-23-42Z.zip)

**Environment**
 - Harvester ISO version: v1.5.0-rc1
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Single node bare metal DL360

**Additional context**

Found while testing #1199 / #7572 
![Image](https://github.com/user-attachments/assets/18f3893e-7931-4d45-94fe-21d58ee63371)

innobead:
Removed regression, as this is a new issue for V2.

Vicente-Cheng:
This is a generic CDI limitation. I will follow up and give some hints about my current behavior.

noahgildersleeve:
Tested in v1.5.0-rc1. Validated as working when the VM is turned off. I'm not sure if we are closing this or implementing a webhook validator first.

Vicente-Cheng:
I moved it back to v1.5.0. We should add validation before exporting volume to the image.
Also, write this limitation to the document. (Doc PR: https://github.com/harvester/docs/pull/728)

ihcsim:
@Vicente-Cheng is this a limitation on kubevirt's smart cloning?

Vicente-Cheng:
@ihcsim, 

Yes, this is the current limitation of CDI.
CDI will check any running workload with the source volume, if so, CDI will block the data fill.

brandboat:
@Vicente-Cheng, in LHv2, does export image only work when VM is stopped? I'm not sure if this behavior is expected, could you please take a look?

### Test Environment
- ipxe, single node, 8C/12G/250GB, v1.5.0-rc3
---
### 🔴  Test Scenario: export image with running VM
- Enable Longhorn v2 on cluster
[...]

PR #7897, opened 2025-03-18
Misc improvement/fixes for 3rd-party storage features

Vicente-Cheng:

**Problem:**
Serval issues need to be improved or fixed

**Solution:**
- Implement backend for https://github.com/harvester/harvester/issues/7875
- add validator for https://github.com/harvester/harvester/issues/7854
- add node affinity for https://github.com/harvester/harvester/issues/7899
- improve the webhook of golden PVC delete
- add validator for https://github.com/harvester/harvester/issues/7876

**Related Issue:**
See above

**Test plan:**
Refer to each issue

### DIFF
--- a/pkg/api/volume/handler.go
+++ b/pkg/api/volume/handler.go
@@ -36,6 +36,7 @@ import (
 
 type ActionHandler struct {
 	images      v1beta1.VirtualMachineImageClient
+	pods        ctlcorev1.PodCache
 	pvcs        ctlcorev1.PersistentVolumeClaimClient
 	pvcCache    ctlcorev1.PersistentVolumeClaimCache
 	pvs         ctlharvcorev1.PersistentVolumeClient
@@ -66,8 +67,8 @@ func (h *ActionHandler) Do(ctx *harvesterServer.Ctx) (interface{}, error) {
 		if input.Namespace == "" {
 			return nil, apierror.NewAPIError(validation.InvalidBodyContent, "Parameter `namespace` is required")
 		}
-		if errMsg := h.validateExportVolume(input.StorageClassName, pvcNamespace, pvcName); errMsg != "" {
-			return nil, apierror.NewAPIError(validation.InvalidBodyContent, errMsg)
+		if err := h.validateExportVolume(input.StorageClassName, pvcNamespace, pvcName); err != nil {
+			return nil, apierror.NewAPIError(validation.InvalidBodyContent, err.Error())
 		}
 		return h.exportVolume(req.Context(), input.Namespace, input.DisplayName, input.StorageClassName, pvcNamespace, pvcName)
 	case actionCancelExpand:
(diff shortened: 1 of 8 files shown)


pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -115,6 +115,15 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
 			if apierrors.IsNotFound(err) {
 				if !createPVCWithDataVolume {
 					pvcAnno.Namespace = vm.Namespace
+					// trigger to create the blank filesystem volume
+					if *pvcAnno.Spec.VolumeMode == corev1.PersistentVolumeFilesystem {
+						pvcAnnos := pvcAnno.GetAnnotations()
+						if pvcAnnos == nil {
+							pvcAnnos = make(map[string]string)
+						}
+						pvcAnnos[util.AnnotationVolForVM] = "true"
```
ihcsim, line 124: Per discussion from #7799, won't the UI already set this annotation at https://github.com/harvester/harvester-ui-extension/blob/9343813ace09bc3ba409c300745fa307ebf31cbd/pkg/harvester/edit/harvesterhci.io.volume.vue#L279-L280?

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -66,6 +69,22 @@ func (v *pvcValidator) Delete(request *types.Request, oldObj runtime.Object) err
...
+						if len(imageRaw) == 2 {
+							imageNS = imageRaw[0]
+							imageName = imageRaw[1]
+						}
+					}
+				}
+				// don't care other errors, just check if the image is still there
+				if _, err := v.imageCache.Get(imageNS, imageName); apierrors.IsNotFound(err) {
+					return nil
+				}
```
ihcsim: So this is saying, a golden image can be deleted if it's either:

* lost
* terminating or
* not associated to any vm images

correct?

Shouldn't this new check also be done outside of the `if oldPVC.Status.Phase != corev1.ClaimLost && oldPVC.Status.Phase != "Terminating"` conditions?

pkg/webhook/resources/setting/validator.go
```diff
@@ -978,6 +980,29 @@ func (v *settingValidator) validateUpdateVolumeSnapshotClass(_ *v1beta1.Setting,
...
+		return nil
+	}
+
+	// check if any disk is still using for v2 data engine
+	if oldSetting.Value == "true" && newSetting.Value == "false" {
+		lhnodes, err := v.lhNodeCache.List(util.LonghornSystemNamespaceName, labels.Everything())
+		if err != nil {
+			return werror.NewInternalError(err.Error())
+		}
+		for _, node := range lhnodes {
```
ibrokethecloud, line 994: if there are multiple block device type disks, do we want to aggergate all names and return them together?

pkg/api/volume/handler.go
```diff
@@ -95,32 +96,61 @@ func (h *ActionHandler) Do(ctx *harvesterServer.Ctx) (interface{}, error) {
 	}
 }
 
-func (h *ActionHandler) validateExportVolume(storageClassName, pvcNamespace, pvcName string) string {
+func (h *ActionHandler) checkPVCInUsed(pvcNamespace, pvcName string) error {
```
ibrokethecloud: would it make sense to conver the logic to an indexer that indexes pods by pvc names and then we can just query pod status if pvc is attached to any pod

WebberHuang1118 (approved):
LGTM, just the [indexes](https://github.com/harvester/harvester/pull/7897#pullrequestreview-2700621595) mentioned by @ibrokethecloud, thanks.

bk201 (approved):

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: This is a different case with the host disk path.
So, we need to add the annotation here, like the blank volume creation.

pkg/webhook/resources/setting/validator.go
Vicente-Cheng: Hmm, I won't go through all the nodes. 
So, return once the first block device type disk.
This should be a hint the user needs to handle it or he will reencounter it.

pkg/api/volume/handler.go
Vicente-Cheng: good suggestion, updated!

pkg/webhook/resources/persistentvolumeclaim/validator.go
Vicente-Cheng: Thanks for the suggestion!

I rewrite the goldenImageCheck mechanism.
Please help to recheck

pkg/api/volume/handler.go
```diff
@@ -95,32 +96,58 @@ func (h *ActionHandler) Do(ctx *harvesterServer.Ctx) (interface{}, error) {
...
-func (h *ActionHandler) validateExportVolume(storageClassName, pvcNamespace, pvcName string) string {
-	if storageClassName == "" {
-		return "Parameter `storageClassName` is required"
+func (h *ActionHandler) checkPVCInUsed(pvcNamespace, pvcName string) error {
+	// find any pod use this PVC (same validation on CDI)
+	index := fmt.Sprintf("%s-%s", pvcNamespace, pvcName)
+	if pods, err := h.pods.GetByIndex(indexPodByPVC, index); err == nil && len(pods) > 0 {
+		podList := []string{}
+		for _, pod := range pods {
+			indexedPod := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
```
votdev, line 105: We should really start implementing a helper function for that (e.g. https://github.com/harvester/vm-import-controller/pull/69/files#diff-988f959e35b780a27c5aede93f5acf9118827cac3d43da1a5cd648dfd78991c9R25)

pkg/api/volume/handler.go
```diff
@@ -95,32 +96,58 @@ func (h *ActionHandler) Do(ctx *harvesterServer.Ctx) (interface{}, error) {
 	}
 }
 
-func (h *ActionHandler) validateExportVolume(storageClassName, pvcNamespace, pvcName string) string {
-	if storageClassName == "" {
-		return "Parameter `storageClassName` is required"
+func (h *ActionHandler) checkPVCInUsed(pvcNamespace, pvcName string) error {
```
votdev: Change function name to

```suggestion
func (h *ActionHandler) checkPVCIsUsed(pvcNamespace, pvcName string) error {
```

or

```suggestion
func (h *ActionHandler) checkPVCInUse(pvcNamespace, pvcName string) error {
```

Actually a boolean value should be returned if the function is called like this.

Alternatively

```suggestion
func (h *ActionHandler) assertPVCNotInUse(pvcNamespace, pvcName string) error {
```

ihcsim (approved):

ibrokethecloud (approved):
lgtm. thanks.

pkg/api/volume/handler.go
Vicente-Cheng: Thanks for the suggestion. I use the `assertPVCNotInUse` to keep the return error.

pkg/api/volume/handler.go
Vicente-Cheng: nice suggestion! I will improve that later!

Vicente-Cheng:
Hi @votdev, I would like to merge it first once we need this commit for the rc-2.
I will try to improve with your comment (https://github.com/harvester/harvester/pull/7897#discussion_r2005649485) later.

If you have any suggestions for this PR, feel free to keep raising them here!
Thanks!

Issue #7692, opened 2025-02-25
[BUG] Improve VMBackup status when the data in backup target is removed


FrankYang0529:
**Describe the bug**

Currently, we support two clusters to use a same backup target. In original design, we only want to allow cluster A can migrate data to cluster B. It's about ADD & READ operations. However, we don't deny DELETE operation. If cluster B remove VMBackup, the data will be removed in backup target as well. In this situation, cluster A cannot  restore the VMBackup.

**To Reproduce**
1. Create two clusters A and B.
2. Set backup target to cluster A.
3. Create VM & VMBackup.
4. Set backup target to cluster B.
5. Remove the VMBackup on cluster B.
6. Restore VMBackup on cluster A.

**Expected behavior**

The VMBackup CR has a condition to show that data in backup target is gone, so users have chance to know the VMBackup can't be restored.

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

WebberHuang1118:
IIUC, this is similar with #7631

FrankYang0529:
Hi @WebberHuang1118, thanks for the information. I think #7631 focus more on retaining data in VMBackup instead of removing source data. This issue will focus more on if data in backup target is removed, we would like to let users know the VMBackup cannot work. For example, admin get into backup target and remove some data, then the VMBacup cannot work.

TachunLin:
Verified **failed** on `v1.6.0-rc3`. Move back to implement.


* When we set the backup target and given the `refreshIntervalInSeconds` to none zero value

  <img width="826" height="244" alt="Image" src="https://github.com/user-attachments/assets/be7b0835-ce91-48cd-8c5e-abef3e0c0571" />


* VM backup shows ready
  <img width="1595" height="785" alt="Image" src="https://github.com/user-attachments/assets/470c1b20-9bb1-411e-8a68-48493de7d0a9" />


* After we move backup volume out to other place in the longhorn-test-nfs container
  ```
  [root@longhorn-test-nfs-54b8774f8f-6d5j5 7c]# ls
  pvc-f667e848-4651-4bfe-92b5-3f05c1ef4f0e
  [root@longhorn-test-nfs-54b8774f8f-6d5j5 7c]# mv pvc-f667e848-4651-4bfe-92b5-3f05c1ef4f0e ../../../../
[...]

PR #7913, opened 2025-03-20
feat: check existing vmbackup

FrankYang0529:

**Problem:**
If data in backup-target is removed, the vmbackup cannot work as expected. Currently, users cannot know whether a VMBackup is healthy. They need to restore it and find whether it's broken.

**Solution:**
Check the existing VMBackup regularly. If data is missing, set it as non-ready.

**Related Issue:**
https://github.com/harvester/harvester/issues/7692

**Test plan:**
1. Create a cluster and setup backup target with non-zero `refreshIntervalInSeconds`.
2. Create a VMBackup.
3. Get into backup target and move LH data to other places. For example:
```
> cd /opt/backupstore/
> mv backupstore/volumes/97/e3/pvc-b1f62a1d-f364-4e5a-9406-7db6106e59ee/ ../../../../
```
4. Wait for refresh interval, we can see the VMBackup is marked as non-ready and `readyToUse` of related volume backup is false. 
5. Get into backup target and move the data back.
6. Wait for refresh interval, we can see the VMBackup is marked as ready again.

### DIFF
--- a/pkg/controller/master/backup/backup_metadata.go
+++ b/pkg/controller/master/backup/backup_metadata.go
@@ -24,6 +24,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/utils/pointer"
+	"k8s.io/utils/ptr"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/config"
@@ -152,6 +153,12 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
 		return nil, nil
 	}
 
+	contextLogger.Info("check existing vm backup...")
+	if err = h.checkExistingVMBackup(target); err != nil {
+		contextLogger.WithError(err).Errorf("can't check existing vm backup")
+		h.settings.EnqueueAfter(setting.Name, 5*time.Second)
+		return nil, nil
+	}
 	return h.renewBackupTarget(setting)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controller/master/backup/backup_metadata.go
```diff
@@ -152,6 +153,12 @@ func (h *MetadataHandler) OnBackupTargetChange(_ string, setting *harvesterv1.Se
 		return nil, nil
 	}
 
+	contextLogger.Info("check existing vm backup...")
+	if err = h.checkExistingVMBackup(target); err != nil {
+		contextLogger.WithError(err).Errorf("can't check existing vm backup")
```
w13915984028, line 158: It may have possible performance penalty:

When refreshIntervalInSeconds is set, in each refresh interval, all backups are checked, not in an incremental mode

w13915984028 (commented):
Some questions:

The check functions run in several nested loops. When there are many backups, the performance needs to be considered.

(1) Could we lazily check it until users use the `backup`, e.g. to restore from a backup ?

(2) Has a menu on individual backup, users trigger to check a specific backup.

(3) Or, has a setting option to `enable`/`disable` the backup object check.

thanks.

pkg/controller/master/backup/backup_metadata.go
FrankYang0529: We cannot do it in an incremental mode. If a VMBackup is heathy in this check and a data is removed unexpectedly after 5 seconds, how do we know this VMBackup is unhealthy? If using incremental mode, we skip this VMBackup check. So we finally need `refreshIntervalInSeconds` to check all VMBackup again.

WebberHuang1118 (commented):
[0001-refactor.txt](https://github.com/user-attachments/files/21094609/0001-refactor.txt)

I have a few suggestions to improve readability through some refactoring. Please take a look—thanks!

WebberHuang1118 (approved):
LGTM, thanks.

Issue #7923, opened 2025-03-21
[BUG] Upgrade a witness cluster with default storage class will stuck in Pre-draining


albinsun:
**Describe the bug**
* If we upgrade a 3 nodes witness cluster (2 default and 1 witness) with default storage class (replica is 3), it will stuck in Pre-draining.
* Same upgrade can PASS when using a custom storage class with replica 2
![Image](https://github.com/user-attachments/assets/c33c0d41-4b59-467d-b90f-83b7f9fda961)

**To Reproduce**
1. Provision a 3 nodes witness cluster (`v1.5-cb719430-head`)
2. Trigger a same version upgrade

**Expected behavior**
If this is expected, may consider to add an error message when user triggering upgrade on a witness cluster which only has 2 mgmt nodes.

**Support bundle**
[supportbundle_8860c991-776d-46fa-ab77-4b15b2e521fd_2025-03-21T08-44-59Z.zip](https://github.com/user-attachments/files/19385532/supportbundle_8860c991-776d-46fa-ab77-4b15b2e521fd_2025-03-21T08-44-59Z.zip)

**Environment**
 - Harvester ISO version: `v1.5-cb719430-head`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): QEMU/KVM

**Additional context**
n/a

w13915984028:
The `pre-drain` is waiting for `upgrade-repo` VM's volume to be `healthy`, it is `degraded` now. This looks to be a bug.


```
2025-03-21T08:30:53.422465112Z ++ jq -r '.items | length'
2025-03-21T08:30:53.529301110Z + node_count=3
2025-03-21T08:30:53.529437781Z + kubectl get engines.longhorn.io -n longhorn-system -o json
2025-03-21T08:30:53.529612536Z + jq -r '.items | map(select(.status.currentState == "running")) | map(.metadata.name + " " + .metadata.labels.longhornvolume) | .[]'
2025-03-21T08:30:53.529720263Z + read -r lh_engine lh_volume
2025-03-21T08:30:53.655561649Z + echo Checking running engine pvc-159093d9-a3ef-437d-a68e-4d1f3e3d621f-e-0...
2025-03-21T08:30:53.655586339Z + '[' true ']'
2025-03-21T08:30:53.655590726Z + '[' 3 -gt 2 ']'
[...]

PR #7924, opened 2025-03-21
Skip witness node when pre-drain waits for longhorn

w13915984028:

**Problem:**

3-node with 1 witness cluster, upgrade stucks on waiting for repo VM's volume to be healty, when SC replica is 3

**Solution:**

Exclude witness node in pre-drain, as what has been done on apply_manifest.

**Related Issue:**
https://github.com/harvester/harvester/issues/7923

**Test plan:**

Per issue descrition https://github.com/harvester/harvester/issues/7923


w13915984028:
@mergify backport v1.5

bk201 (approved):

bk201:
@w13915984028 can you help explain why making the variable local fix the issue in the commit?

w13915984028:
@bk201

Ideally, unless the shell variable is designed to be global, all vars inside a function should be claimed with `local`, it will not pollute/overwrite the global var accidentally, or vice versa. 

Unfortunately, many of the existing codes did not follow this rule. New code either. As we have no guide line / tool to enforce it.

Back to this PR, the `local` does not help fix the bug directly, but mainly follow the above practice.

Vicente-Cheng (approved):

Issue #3015, opened 2022-10-23
[FEATURE] CPU Masking (aka EVC): Live Migration cannot proceed since no node is suitable to run the required CPU model NoSuitableNodesForHostModelMigration


armagankaratosun:
**Describe the bug**

I have a Harvester cluster with 2 nodes, each of them has Ryzen CPUs but a different generation. 1 Host (ultron02) has the AMD Ryzen 7 5700G with Radeon Graphics while the other one (ultron01) has AMD Ryzen 7 3700X. During my tests with the cluster, I realized that I can live migrate VMs from ultron01 to ultron02 but not the other way around. I found that this is happening because some of the CPU features are missing the older gen CPU. 

I realized that the pod scheduled to migrate the VM is in the pending state with a warning `Warning  FailedScheduling  32s   default-scheduler  0/2 nodes are available: 2 node(s) didn't match Pod's node affinity/selector.` and In the VM events, there were some CPU features that are listed as missing.

```
Status: NoSuitableNodesForHostModelMigration	
VirtualMachineInstance: migrate-test

Migration cannot proceed since no node is suitable to run the required CPU model / required features: map[cpu-feature.node.kubevirt.io/amd-ssbd:true cpu-feature.node.kubevirt.io/arch-capabilities:true cpu-feature.node.kubevirt.io/cmp_legacy:true cpu-feature.node.kubevirt.io/erms:true cpu-feature.node.kubevirt.io/fsrm:true cpu-feature.node.kubevirt.io/hypervisor:true cpu-feature.node.kubevirt.io/invtsc:true cpu-feature.node.kubevirt.io/pschange-mc-no:true cpu-feature.node.kubevirt.io/skip-l1dfl-vmentry:true cpu-feature.node.kubevirt.io/tsc-deadline:true cpu-feature.node.kubevirt.io/tsc_adjust:true cpu-feature.node.kubevirt.io/vaes:true cpu-feature.node.kubevirt.io/vpclmulqdq:true cpu-feature.node.kubevirt.io/x2apic:true cpu-feature.node.kubevirt.io/xsaves:true host-model-cpu.node.kubevirt.io/EPYC-Rome:true]
```
When I run `kubeclt describe` on the Pending migration pod
```
ultron02:~ # kubectl describe po -n default virt-launcher-migrate-test-9v7j7
Name:           virt-launcher-migrate-test-9v7j7
Namespace:      default
Priority:       0
Node:           <none>
[...]

PR #7937, opened 2025-03-25
HEP: CPU models selection

Yu-Jack:

**Problem:**
There is no way to configure CPU model and features in Harvester Dashboard.

**Solution:**
The most changes are Harvester Dashboard because we plan to provide new pages to help users configure them.

**Related Issue:**
https://github.com/harvester/harvester/issues/3015


enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,402 @@
...
+
+## Proposal
+
+Due to multiple nodes, we can't show a big matrix for all the CPU models and features. Instead, we'll provide a dropdown selection menu with colored mark.
+
+### User Stories
+
+#### Story 1
+
+I have multiple nodes that have a common CPU model called `Nehalem`. For some reason, I'll need to migrate mv VMs to other nodes without manually shutting down. To ensure compatibility, I create my MVs with `Nehalem` as the CPU model.
```
ibrokethecloud: minor typo
```suggestion
I have multiple nodes that have a common CPU model called `Nehalem`. For some reason, I'll need to migrate mv VMs to other nodes without manually shutting down. To ensure compatibility, I create my VMs with `Nehalem` as the CPU model.
```

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,402 @@
...
+For certain reasons, I don’t want my VMs to have the `avx` CPU feature. Therefore, I configure the CPU feature `avx` with the `forbid` policy. This ensures that my VMs will not be scheduled to nodes that support the avx CPU feature.
+
+
+### User Experience In Detail
+
+Users need to understand their requirements and strategy of using virtual machine.
+
+- Do they need a specfic cpu model for VMs?
+- Do they need a specific cpu feature for VMs?
+- Do they need to migrate VM to different cluster without manually shutting down? They need to understand [how live migration works](https://docs.harvesterhci.io/v1.4/vm/live-migration/#how-migration-works).
```
ibrokethecloud: i assume this was meant to be `node`
```suggestion
- Do they need to migrate VM to different node without manually shutting down? They need to understand [how live migration works](https://docs.harvesterhci.io/v1.4/vm/live-migration/#how-migration-works).
```

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,402 @@
...
+Action Items:
+
+- [ ] Create a new tab in VM creation page.
+- [ ] Create a new tab in VM template page.
+- [ ] Provide a dropdown selection menu for selecting models and features.
+- [ ] Calculate common and uncommon the CPU models and features across all nodes.
+
+#### Backend
+
+Backend should reject the unreasonable request from frontend. When users try to migrate a VM, the `findMigratableNodes` action should return available nodes that matched the selected CPU model and features to avoid scheduling failure.
```
ibrokethecloud: the cpu model and features are already available as labels

if we convert the cpu feature requirements to labels used for vm scheduling then `findMigratableNodesByVMI` will already handle the scenario of finding nodes based on scheduling requirements

Yu-Jack:
BTW, about `CPUNodeDiscovery` feature gate, it had beed removed in this https://github.com/kubevirt/kubevirt/pull/7728. So, the default behavior is enabled CPUNodeDiscovery feature gate (https://github.com/kubevirt/user-guide/pull/762).


These feature gates are enabled by default (https://github.com/kubevirt/kubevirt/blob/352416fe7c2b944486d14e226e746f0051b2d0f7/pkg/virt-config/featuregate/inactive.go#L101-L119).

```
// GA represents features that reached General Availability.
// GA features are considered feature-gate enabled, with no option to disable them by an FG.
GA State = "General Availability"
```

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,402 @@
...
+
+- Scenario 3: Person requiring a specific color and shape  
+  A person requires an object with a specific color and shape (e.g., Red Circle).  
+  You must give the person an object that matches the color and shape (e.g., a Red Circle or a Red-Blue Circle).
+
+- Scenario 4: Person requiring a moving to a new location  
+  A person requires an object with a specific color and shape (e.g., Red Circle) and is moving to a new location.  
+  The new location must provide an object that matches the color and shape (e.g., a Red Circle or a Red-Blue Circle).
+
+A Node can provide objects with different shapes and colors. For example, a Node might provide a "Red Square", a "Blue Circle", and a "Red-Blue Circle".
```
ihcsim:
```suggestion
A node can provide objects with different shapes and colors. For example, a node might provide a "Red Square", a "Blue Circle", and a "Red-Blue Circle". 
```

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,402 @@
...
+
+![](20250407-cpu-model-features-selection/04-custom-model.png)
+
+> These are just examples. The real one will match our Harvester GUI style.
+
+### Implementation Overview
+
+#### Frontend
+
+Frontend needs to create a new tab in VM creation page and provide a dropdown selection menu for selecting models and features. This selection is also available in the VM template page. Then, calculate the common and uncommon CPU models and features across all nodes.
```
ihcsim: This can be quite memory-heavy for the frontend on large clusters. Essentially, every time the user navigates to the VM creation or the VM template page, the frontend has to iterate through all the nodes, and for each node, iterate through all its labels, in order to compute, deduplicate and sort (?) the CPU models and features lists.

Even after all these, if the user picks the wrong target node from the migration modal box, the migration still fails.

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,402 @@
...
+
+
+#### Story 3
+
+For certain reasons, I don’t want my VMs to have the `avx` CPU feature. Therefore, I configure the CPU feature `avx` with the `forbid` policy. This ensures that my VMs will not be scheduled to nodes that support the avx CPU feature.
+
+
+### User Experience In Detail
+
+Users need to understand their requirements and strategy of using virtual machine.
```
ihcsim: I bet most users don't know this level of details. When migration failed, they're gonna end up having to come back to the UI, shutdown the VM, and change the CPU model/feature to something that they don't fully understand.

ihcsim (commented):
Thanks for putting this together. It is annoying that kubevirt doesn't offer a way to [influence the new pod's node selectors](https://github.com/kubevirt/kubevirt/blob/04b1fda82c3b1ec3f01298e5dd5f7a6468cac0fc/pkg/virt-controller/watch/migration/migration.go#L760-L772). The only way to bypass those node selectors constraint is to change the model to non-`host-model`. That is not easy for our controller to do on the fly during migration because it requires the VM to be shutdown and restarted before migration can proceed. There is also the challenge that once a VM has been migrated once, the [pod's node selectors get copied to subsequent new pods](https://github.com/kubevirt/kubevirt/blob/04b1fda82c3b1ec3f01298e5dd5f7a6468cac0fc/pkg/virt-controller/watch/migration/migration.go#L2078-L2084) in future migrations.

If it's possible to modify pod's node selectors, then we can present the user with a checkbox option in the migration modal to "skip CPU model/features constraints" during migration.

Providing a user with ways to input the CPU models and features have its pro and cons. (See review comments below.) Even then, migration can still fail if the user inputs the wrong model/features or selects the wrong target node from the migration modal box. I don't know if we can fully fix the issue because at the end, the cluster admin has to decide whether the target node is suitable for migration or not.

One thing I think we need to do is to propagate the migration failure reasons from the migration `event` or the `virtualmachineinstancemigration` object  back to the UI, if not already.

I'd love to hear more thoughts on this.

Yu-Jack:
@ibrokethecloud  @ihcsim 

Here are the new updates after the latest commit:

- Goals: I've changed the goals.

- CPU Feature Input: The CPU feature drop-down has been removed and replaced with a text input field.

- Common CPU Model: We only calculate the common CPU models. We won't provide other uncommon CPU models. I think it'll be convenient for users to know which common CPU models they can use.

- POD Annotation Mechanism: A new annotation mechanism has been added to mutate PODs. This will remove all `*.kubevirt.io keys` in spec.nodeSelector, ensuring successful VM migration between different hosts and CPU models. This is an [example commit](https://github.com/Yu-Jack/harvester/commit/7d6e9f6a8ae58a532ead1bec26547e161e55b1dc). It's not an official way provided by the underlying system. However, I think it's still a option for users.

- GUI Error Propagation: Scheduling error messages should be propagated to the GUI.

enhancements/20250407-cpu-model-selection.md
```diff
@@ -0,0 +1,408 @@
...
+
+## Summary
+
+Harvester doesn't provide the options for selecting CPU models and features for VMs. This enhancement is to provide a way to select CPU models and features for VMs.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/3015
+
+## Motivation
```
ihcsim, line 11: Maybe a one liner here to describe why this is important to user. Primarily, this HEP addresses the limitation of the current default `host-model` CPU model used by Harvester, where it blocks migrations that are technically legit and possible, even though there is a mismatch between the source and target nodes' underlying CPU model and features.

ihcsim:
@Yu-Jack Thanks for updating the HEP. Instead of supporting per-VM CPU models/features configuration, can we just make it a global setting? The value in this setting (with `host-model` being the default) gets mapped to the kubevirt's `spec.configuration.cpuModel` property as described in https://kubevirt.io/user-guide/compute/virtual_hardware/#enabling-default-cluster-cpu-model. This is essentially the workaround solution described in #3015.

I like the suggested force-migration implementation. I think the option should be available in the live-migration dialog, making it part of the migration experience:

image

We will also need to confirm whether it will trigger a race between our mutator and [kubevirt's controller](https://github.com/kubevirt/kubevirt/blob/04b1fda82c3b1ec3f01298e5dd5f7a6468cac0fc/pkg/virt-controller/watch/migration/migration.go#L760-L772). My guess is that our mutator will run first, then kubevirt's controller will override our overrides. Hence, the force-migration logic might need to reside within our controller, not mutator.

Yu-Jack:
> Instead of supporting per-VM CPU models/features configuration, can we just make it a global setting? The value in this setting (with host-model being the default) gets mapped to the kubevirt's

@ihcsim Oops. I forgot to put that. In generally, in this HEP, I prefer to provide two ways to configure CPU model
- per-VM VPU model/features

Yu-Jack:
> Instead of supporting per-VM CPU models/features configuration, can we just make it a global setting? The value in this setting (with host-model being the default) gets mapped to the kubevirt's

@ihcsim Oops. I forgot to put that. In generally, I prefer to provide two ways to configure CPU model in the HEP.

- Per-VM CPU model/features (About this one, we can see what other people think.)
- Global VM CPU model/features (We must have this one)


> I think the option should be available in the live-migration dialog, making it part of the migration experience:

Yeah, it should be there like your screenshot.


> We will also need to confirm whether it will trigger a race between our mutator and [kubevirt's controller](https://github.com/kubevirt/kubevirt/blob/04b1fda82c3b1ec3f01298e5dd5f7a6468cac0fc/pkg/virt-controller/watch/migration/migration.go#L760-L772). My guess is that our mutator will run first, then kubevirt's controller will override our overrides. Hence, the force-migration logic might need to reside within our controller, not mutator.

This is a Interesting one. In some cases https://github.com/harvester/harvester/issues/8670#issuecomment-3083676872, we just confirmed KubeVirt removes the `/spec/nodeSelector` in the VM and VMI spec  with mutator. However, `*.kubevirt.io` keys in the `spec.nodeSelector` of POD isn't from VM or VMI spec. [The POD is created here](https://github.com/kubevirt/kubevirt/blob/54ebc9c4580e67714789e590b7e6a048344d7452/pkg/virt-controller/watch/migration/migration.go#L833). So, I guess we still need to use mutate the POD. Anyway, I'll take a look and try more different cases.

Yu-Jack:
Regarding of force migration, I'll check whether kubevirt supports. I prefer to use underlying features instead of crafting it ourselves, like we [did in nodeSelector before](https://github.com/harvester/harvester/pull/8707/files). Most features in the HEP are about GUI.

abonillabeeche:
Users may have hardware purchases in waves. Some might be AMD, some could be different generations of Intel. When we set a Gen5 Intel CPU, it allows the migration into newer servers. When we set an older AMD, then it allows such migration to those too. While the VMs should default to the lowest cluster-wide CPU generation in the cluster when the CPU compatibility is enabled, we should allow a per VM config.

abonillabeeche:
Users may have hardware purchases in waves. Some might be AMD, some could be different generations of Intel. When we set a Gen5 Intel CPU, it allows the migration into newer servers. When we set an older AMD, then it allows such migration to those too. While the VMs should default to the lowest cluster-wide CPU generation in the cluster when the CPU compatibility is enabled, we should allow a per VM config.

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,412 @@
...
+Imagine you have colored shapes representing CPU configurations:
+
+- Shapes: Represent CPU models (e.g., Circle, Square, Triangle).
+- Colors: Represent CPU features (e.g., Red, Blue, Green).
+
+Here's how they interact:
+
+- Scenario 1: Person requiring a specific color  
+    A person requires an object with a specific color (e.g., Red).  
+    You can give the person any shape that is red (e.g., a Red Square, a Red Circle, or a Red-Blue Circle).
```
ibrokethecloud: not neccessarily correct. some features might be cpu model specific only

ibrokethecloud:
I am not in favor of `forcedMigration` since the guest may fail to boot on the target node due to initialisation failures.

We have already noticed issues across AMD EPYC-Rome and AMD EPYC-Milan, where they different generations of the same architecture but AMD has changed how the registers operate for `xsaves`. So even though they share a similar generation, the migration fails because the domain does not boot successfully on target mode.

By removing the scheduling requirements due to CPU we are only exposing the guests to more random failures during migrations.

The better option may be to leverage say `kvm64` as cpu model when this flag is set as this would allow migration across vast majority of CPU models. We may need to ensure we accordingly update the `obsoleteCPUModels` in kubevirt CR.

ihcsim:
`kvm64` is not without its own quirks, depending on the guest OS CPU requirements. Maybe it's something that we can document, but I don't think we should implement it as some form of flag-enabled defaults, to avoid a different set of support issues.

Although force-migration would address https://github.com/harvester/harvester/issues/3015 (as agreed by the OP), from what I can see the implementation will be invasive, unless upstream is open to the idea of making the node selection constrain configurable (e.g., from a `hard` preference to `soft`).

Depending on the use case, as discussed in the issue, some form of user's decision will be required to address the live migration limitations, probably down to the per-VM config level per https://github.com/harvester/harvester/pull/7937#issuecomment-3243611961.

Yu-Jack:
From my point of view, I also hope Kubevirt can provide this feature. I don't want to hack our VMI node selector again.

I just opened a [upstream issue](https://github.com/kubevirt/kubevirt/issues/15623) to discuss force migration. To push forward the progress of this issue, we can focus on other parts for now. As for force migration, we can defer it to the future in case KubeVirt supports for it.

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: Okay, I'll add the concept like the red color might be only provide by a shape.

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,414 @@
...
+### Related Issues
+
+https://github.com/harvester/harvester/issues/3015
+
+## Motivation
+
+### Goals
+
+- Support per-VM ang global CPU model/features selections.
+- Support force migration between two different host and cpu models.
```
wheatdog: IIUC, as you mentioned in the HEP review meeting, this wouldn't be the goal of this HEP and instead be handled upstream (in KubeVirt).

I would like to check whether the user story 2 below would also be removed together?

Yu-Jack: Yes, the story 2 will be removed.

Yu-Jack: I'll update the HEP very soon to avoid any confusion.

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,385 @@
...
+### Non-goals
+
+- Change the underlying KubeVirt CPU model or migration logic.
+- Force migration. It should be done [by upstream](https://github.com/kubevirt/kubevirt/issues/15623).
+
+## Introduction
+
+Before this enhancement, we need to understand how the CPU model and features work in KubeVirt. To make the CPU model and migration logic even easier to grasp, let's use a visual analogy with colored shapes:
+
+Imagine you have colored shapes representing CPU configurations:
```
ihcsim: I think it might be clearer if you use say e.g., the output of `lscpu`  to explain the concepts of CPU model, flags/features, family etc. Then you can also compare the outputs from 2 virtual machines and explain that in order for migration to succeed, the CPU characteristics and supported features must be compatible etc.

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,385 @@
...
+### User Experience In Detail
+
+In general, there are two ways to migrate VMs between nodes.
+
+- Use a common CPU model
+- Use a Harvester special defined annotation
+
+Before users use the feature, we need to ensure they know what the difference is between these two ways.
+
+- If they know later joined nodes still have common CPU models, they can select a common CPU model for further migration.
```
ihcsim:
```suggestion
- If the cluster consists of mainly homogeneous nodes (including new nodes), users can define an universal common CPU model to ensure compatible migrations among the nodes.
```

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,385 @@
...
+
+In either case, we need to add a ["learn more for migration"](https://docs.harvesterhci.io/v1.5/vm/live-migration/#how-migration-works) link when users try to customize the CPU model and feature.
```
ihcsim:
```suggestion
In either case, we need to add a ["learn more for migration"](https://docs.harvesterhci.io/latest/vm/live-migration/#how-migration-works) link when users try to customize the CPU model and feature.
```

For the doc task, we should move this content to a new "How Migration Works" deep dive page, to avoid bloating the current page.

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,385 @@
...
+#### Frontend
+
+We'll have two ways to configure this.
+
+- Per-VM CPU model/features while creating the VM
+- Global VM CPU model/features in settings
+
+Frontend needs to create a new tab in the VM creation page and provide a dropdown selection menu for models and an input box for features. This selection is also available in the VM template page and global settings. Then, calculate the common CPU models across all nodes. 
+
+- If users select the common one, we'll show the option with green mark.
```
ihcsim: Is the idea to identify the common one by examining the labels of all the nodes in the cluster? What if there isn't a common one among _all_ the nodes, but only a common one among _some_ of the nodes? This computation also needs to happen every time the user navigates to the VM page whether to create a new VM or update an existing one, right?

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,385 @@
...
+#### Story 1
+
+I have multiple nodes that have a common CPU model called `Nehalem`. For some reason, I'll need to migrate my VMs to other nodes without manually shutting down. To ensure compatibility, I create my VMs with `Nehalem` as the CPU model.
+
+### User Experience In Detail
+
+In general, there are two ways to migrate VMs between nodes.
+
+- Use a common CPU model
+- Use a Harvester special defined annotation
```
ihcsim: What's the annotation?

Yu-Jack: This line should be removed because we don't need force migration.

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: Let me rewrite this section because we've removed force migration method.

enhancements/20250407-cpu-model-selection.md
Yu-Jack: I'll use `Primarily, this HEP addresses the limitation of the current default host-model CPU model used by Harvester, where it blocks migrations that are technically legit and possible.` and remove `even though there is a mismatch between the source and target nodes' underlying CPU model and features.` because source and target node's CPU model must have a common matched one, then VM can be migrated. I hope this is more clearer.

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: Maybe we can trigger the examining by clicking a button? Or we only examine once and save it in the annotation because the nodes are usually not changed until new node joins. We just need to examine it again when a new node joins.

The reason why I think helping users find the common one is important is because https://github.com/harvester/harvester/issues/3015#issuecomment-3024925763 mentioned this feature.

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: If using a realistic example makes more sense, I can change it later.

Yu-Jack:
HI @a110605, could you take a look when you're free? I need feedback from UI side. Thanks!

Yu-Jack:
Let me wrap up some points for this HEP.

We all agree with:

- Global wide CPU model configuration
- Per-VM CPU Model configuration (https://github.com/harvester/harvester/pull/7937#issuecomment-3243611961)
- Won't have force migration in this HEP, we [should discuss and implement it in the upstream](https://github.com/kubevirt/kubevirt/issues/15623).


We still have some discussions on:

- Calculate the common CPU models to help users select or input

@ihcsim  Could you help me take a look at this? Thanks!

a110605:
> HI @a110605, could you take a look when you're free? I need feedback from UI side. Thanks!

Rancher has [mutli-select](https://github.com/a110605/harvester-dashboard/blob/9ad30c413679d647ca9f537ba287dde2d576bd9a/pkg/eks/components/Networking.vue#L348) component that can allow user to select mutliple options.

For the option label, we can explicit list the text like
```
SSE4.2 (Supported by all nodes)
FMA (Supported by some nodes) -> FMA (supported by node1, node 2, node3)
```

But for the green / yellow color decroations, the rancher component doesn't provide such mechansim.

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,382 @@
...
+- Per-VM CPU model/features while creating the VM
+- Global VM CPU model/features in settings
+
+Frontend needs to create a new tab in the VM creation page and provide a dropdown selection menu for models and an input box for features. This selection is also available in the VM template page and global settings. Then, calculate the common CPU models across all nodes. 
+
+- If users select the common one, we'll show the option with a green mark.
+
+The data of CPU models are from node's labels:
+
+- The model is from `cpu-feature.node.kubevirt.io/{model}` of the node's labels.
```
ihcsim: I thought model comes from `cpu-model.node.kubevirt.io/{model}`

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,382 @@
...
+### Implementation Overview
+
+#### Frontend
+
+We'll have two ways to configure this.
+
+- Per-VM CPU model/features while creating the VM
+- Global VM CPU model/features in settings
+
+Frontend needs to create a new tab in the VM creation page and provide a dropdown selection menu for models and an input box for features. This selection is also available in the VM template page and global settings. Then, calculate the common CPU models across all nodes.
```
ihcsim: Sounds like a computationally expensive op for the UI to generate the list of common CPU models and features from all the nodes every time the user visits the VM page. Can the backend provide a cache?

enhancements/20250407-cpu-model-features-selection.md
```diff
@@ -0,0 +1,382 @@
...
+  - If `cpu.features` is provided, each item in the array must include a `cpu.features[].name` field. 
+  - `cpu.features[].policy` is optional. If omitted, the default value is `require`.  
+    Valid values are: `force`, `require`, `optional`, `disable`, `forbid`.
+
+Since VM spec in `virtualmachinetemplateversion` is the same, please use the same logic to fill the model and features.
+
+
+Action Items:
+
+- [ ] Create a new tab in the VM creation page.
```
ihcsim: Can we reuse the existing "Basic" or probably "Advanced Options" tab? I don't think most users will care about this or know what to do with this.

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: Good Catch.

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: Agree. Let's reuse `Advanced Options` tab.

enhancements/20250407-cpu-model-selection.md
```diff
@@ -0,0 +1,381 @@
...
+No.
+
+## Design
+
+The CPU models are like this:
+
+- Node-1 cpuModel: IvyBridge, Penryn
+- Node-2 cpuModel: IvyBridge, Westmere
+- Node-3 cpuModel: IvyBridge, SandyBridge
+- Node-4 cpuModel: IvyBridge, Westmere
```
ihcsim, line 222: What does the UI show the user given the following permutation?

```
- Node-1 cpuModel: IvyBridge, Penryn
- Node-2 cpuModel: IvyBridge
- Node-3 cpuModel: SandyBridge
- Node-4 cpuModel: Westmere
```

Do we still only show `IvyBridge` because that's the only migratable configuration? But what if a user has legit requirements to want to pick `SandyBridge`? That makes the VM non-migratable because there are no other compatible target nodes.

Maybe the UI needs to show all unique CPU models, but label some as migratable, while others not. WDYT?

Yu-Jack: We can do like this:

Selection List (Supported Node Num/Total Node Num):
- IvyBridge (2/4)
- Penryn (1/4)
- SandyBridge (1/4)
- Westmere (1/4)

I think that's more clear than this one:
- IvyBridge (Migratable)
- Penryn (Non-Migratable)
- SandyBridge (Non-Migratable)
- Westmere (Non-Migratable)

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: It's a good question. If we'd like the cache, the possible way is using annotation or local cache. Since our harvester controller has a leader mechanism (API request goes to the same controller), so we can use local cache as well. When a new node joins, we update local cache.

The key point depends on how much memory we'll use. 

```
# current resource object memory usage
$ k get nodes harvester1 -o json | wc -c | awk '{printf "%.2f KiB\n", $1/1024}'
39.18 KiB
```

If the memory requires not much, we can use annotation. If not, we might use local cache.

enhancements/20250407-cpu-model-selection.md
ihcsim: I am thinking more along the line of:

image

Maybe @a110605 can provide some suggestions here.

enhancements/20250407-cpu-model-features-selection.md
ihcsim: Maybe as a first pass, we can try without a cache. I think as long as the computation is done in the backend, it shouldn't have ~any~ much negative performance impact on the UI. (If I'm wrong, then we can add a cache :smile: - maybe store it in a config map or something.)

If we agreed that the backend should do the computation, then we should explore what the response payload to the UI will look like. WDYT?

Yu-Jack: > If we agreed that the backend should do the computation, then we should explore what the response payload to the UI will look like. WDYT?

A simple one is:

```js
{
  "totalNodes": 10,
  "models": {
    "Skylake-Client-IBRS": {
      "readyCount": 8,
      "migrationSafe": false
    },
    "EPYC-IBPB": {
      "readyCount": 10,
      "migrationSafe": true
    }
  },
  "features": {
    "avx2": { 
        "readyCount": 7,
        "migrationSafe": false 
    },
  },
}

// or

{
  "totalNodes": 10,
  "models": {
    "Skylake-Client-IBRS": 8,
    "EPYC-IBPB": 10
  },
  "features": {
    "avx2": 7,
  },
}

```

The "readyCount" present how many nodes support. That could also apply these two different UI cases.

enhancements/20250407-cpu-model-selection.md
Yu-Jack: When the case is:

```
- Node-1 cpuModel: IvyBridge, Penryn
- Node-2 cpuModel: IvyBridge,Penryn
- Node-3 cpuModel: IvyBridge,SandyBridge
- Node-4 cpuModel: IvyBridge,Westmere
```

It could also look like this:
```
--Fully Migratable--
IvyBridge
--Partial Migratable--
Penryn
--Non-Migratable--
SandyBridge
Westmere
```

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: About the feature part, we can do it next time, don't need to include it in this HEP.

enhancements/20250407-cpu-model-selection.md
ihcsim: From a UX perspective, I don't know if "partial migratable" will add much. As long as it's migratable, K8s can figure out the target node based on the node selector. From a user perspective, migration will just work.

enhancements/20250407-cpu-model-features-selection.md
ihcsim: Looks good - let's add the sample payload to the HEP. The `migrationSafe` field makes things clearer for the frontend.

Re: the feature part, I agreed that we don't need it now. I thought I added a comment on it but it's no way to be found.

enhancements/20250407-cpu-model-selection.md
ihcsim: Of course, there is always a chance that what is migratable now can become non-migratable later because e.g., the compatible target nodes were removed or died. I think we can just let KubeVirt handle these edge cases as is.

Yu-Jack: We just need to focus on providing a list and marking some cpu models for users. Other stuffs are handled by KubeVIrt.

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: updated hep

enhancements/20250407-cpu-model-selection.md
Yu-Jack: updated hep

enhancements/20250407-cpu-model-selection.md
```diff
@@ -0,0 +1,422 @@
...
+      "readyCount": 4,  
+      "migrationSafe": true  
+    },  
+    "Penryn": {  
+      "readyCount": 1,  
+      "migrationSafe": false  
+    }  ,
+    "Westmere": {  
+      "readyCount": 2,  
+      "migrationSafe": false
```
ihcsim: I thought this should be `migrationSafe: true` because there are 2 nodes with this model? From the live migration UI, the target node will be either Node-2 or Node-4, depending which one is the source node. In the case of live migration during upgrade, K8s can automatically figure out the target node based on the pod's node selector.

Similarly, in line 157 of Case 2.

enhancements/20250407-cpu-model-selection.md
```diff
@@ -0,0 +1,422 @@
...
+- Node-1 cpuModel: IvyBridge, Penryn
+- Node-2 cpuModel: IvyBridge, Westmere
+- Node-3 cpuModel: IvyBridge, SandyBridge
+- Node-4 cpuModel: IvyBridge, Westmere
+
+
+```json
+{  
+  "totalNodes": 10,  
+  "models": {
```
ihcsim: Maybe we should also include `host-model`, `host-passthrough` and whatever the global model is.

enhancements/20250407-cpu-model-selection.md
Yu-Jack: Yeah, it makes sense.

enhancements/20250407-cpu-model-selection.md
Yu-Jack: It's not guaranteed to migrate the VM with host-model sometime. So, how about we add a new filed called `globalModels: ["host-model", "host-passthrough"]`? The options will be `globalModels + model {migrationSafe: true}`.

ihcsim: That works. Also, to clarify, by "global model", I meant the one user defines in the `kubevirt` CR. Yes, we can't guarantee live migration would work with these system default; we're just surfacing them on the UI. As with everything else in this HEP, CPU model/feature is an advanced configuration that requires some user's knowledge and responsibility.

Yu-Jack: Do you mean this specific one when you say 'global model', right?

```
apiVersion: kubevirt.io/v1
kind: KubeVirt
metadata:
  name: kubevirt
  namespace: kubevirt
spec:
  configuration:
    cpuModel: "EPYC"
```

ihcsim: yup, that's it

Yu-Jack: Done!

enhancements/20250407-cpu-model-selection.md
```diff
@@ -0,0 +1,471 @@
...
+- `totalNodes`: Total number of ready nodes in the cluster
+- `globalModels`: Array of special CPU models that are always available
+  - `host-model`: Uses the host's CPU model (default behavior)
+  - `host-passthrough`: Passes through the host's CPU features directly
+  - KubeVirt configuration CPU model: The CPU model set in `kubevirt.spec.configuration.cpuModel` (if configured)
+- `models`: Object containing all available CPU models in the cluster
+  - `<cpu-model-name>`: The name of the CPU model (e.g., "IvyBridge", "Penryn")
+    - `readyCount`: Number of ready nodes that support this CPU model
+      - if `readyCount` is greater than zero, that cpu model can be migrated to some nodes in the cluster.
+      - if `readyCount` is equal to one, that cpu model can't be migrated.
```
ihcsim, line 110:
```suggestion
    - `readyCount`: Number of ready nodes that support this CPU model
      - if `readyCount` is greater than one, that cpu model can be migrated to some nodes in the cluster.
      - if `readyCount` is equal to one, that cpu model can't be migrated.
```

enhancements/20250407-cpu-model-selection.md
```diff
@@ -0,0 +1,471 @@
...
+    "<cpu-model-name>": {
+      "readyCount": "<number of ready nodes that support this CPU model>",
+      "migrationSafe": "<boolean: true if readyCount equals totalNodes, false otherwise>"
+    }
+  }
+}
+```
+
+**Field Descriptions:**
+- `totalNodes`: Total number of ready nodes in the cluster
```
ihcsim, line 101: From the example cases below, I am not sure we need this and `migrationSafe`. Technically, migration is safe only if ready count > 1. Are there any situations where `migrationSafe == false` but `readyCount > 1`?

Yu-Jack: > Are there any situations where migrationSafe == false but readyCount > 1?

No. It's just another way to show whether your node is migrated or not. In original thought, I considered migrationSafe as a status that you can migrate your VM to any nodes. Right now, the `readyCount` and `migrationSafe` are quite similar. 

But, I want to keep it flexible if we might need to show the how many nodes VM is able to be migrated to. And, `totalNodes` is just a convenient field for frontend. So, I think we can keep it now. Also, It's not much effort to remove or add those fields. If we need to improve it, then I'll do it during implementation.

ihcsim (dismissed):

a110605 (dismissed):

enhancements/20250407-cpu-model-selection.md
```diff
@@ -0,0 +1,471 @@
...
+
+I have multiple nodes that have a common CPU model called `Nehalem`. For some reason, I'll need to migrate my VMs to other nodes without manually shutting down. To ensure compatibility, I create my VMs with `Nehalem` as the CPU model.
+
+### User Experience In Detail
+
+If the cluster consists of mainly homogeneous nodes (including new nodes), users can define a universal common CPU model to ensure compatible migrations among the nodes.
+
+### API changes
+
+Backend should provide a API to let frontend geenrate a selection list.
```
votdev:
```suggestion
Backend should provide a API to let frontend generate a selection list.
```

Yu-Jack:
Just fix a typo. Nothing changes.

enhancements/20250407-cpu-model-features-selection.md
Yu-Jack: NOTE: [This issue](https://github.com/harvester/harvester/issues/8797) will take care of this.

enhancements/20250407-cpu-model-selection.md
```diff
@@ -0,0 +1,471 @@
...
+
+Action Items:
+
+- [ ] Create the new configurations in the "Advance Options" of the VM creation page and the VM template page.
+- [ ] Provide a dropdown selection menu for models.
+- [ ] Propagate the scheduling error on GUI.
+
+#### Backend
+
+Backend should reject unreasonable requests from the frontend. When users try to migrate a VM, the `findMigratableNodes` action should return available nodes that match the selected CPU model failure.
```
ibrokethecloud: might be typo, i assume we meant `CPU model feature`

ibrokethecloud (approved):
lgtm. thanks for the enhancement proposal

enhancements/20250407-cpu-model-selection.md
ihcsim: SGTM. Also, to make it easier for frontend to parse the payload, we can consider matching the backend implementation to the UI presentation layer. E.g.,

```json
{
    "totalNodes": "<total number of ready nodes in the cluster>",
    "global": [
        "host-model",
        "host-passthrough",
        "one from kubevirt spec.configuration.cpuModel"
    ],
    "migratable": [
        {
            "IvyBridge": {
                "readyCount": 4,
                "migrationSafe": true
            }
        },
        {
            "Westmere": {
                "readyCount": 2,
                "migrationSafe": true
            }
        }
    ],
    "nonMigratable": [
        {
            "Penryn": {
                "readyCount": 1,
                "migrationSafe": false
            }
        },
        {
            "SandyBridge": {
                "readyCount": 1,
                "migrationSafe": false
            }
        }
    ]
}
```

ihcsim (approved):

Issue #7944, opened 2025-03-26
[BUG] Should allow the VM take snapshot when CSI driver support snapshot


Vicente-Cheng:
### Describe the Bug

After v1.5.0, we block all the backup operations with any 3rd-party storage volume.
It affects the VM snapshot operation. The current code logic will block the VM snapshot once we do not support backup. This means we cannot take the VM snapshot even if the CSI driver has the capability to do so.

So, it seems we need to isolate the individual logic to handle the VM snapshot.

It also needs the UI changes.
cc @a110605, @WebberHuang1118, @bk201, @innobead

### To Reproduce

1. create harvester cluster
2. setup LVM CSI driver (LVM CSI driver support snapshot)
3. Create VM with LVM backend
4. Check the VM operations. We cannot take snapshot, but we can take snapshot for root volume.

### Expected Behavior

The VM should be able to take snapshot like volume.

### Support Bundle for Troubleshooting

no need.

### Environment

- Harvester version: v1.5.0-rc2
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

WebberHuang1118:
@Vicente-Cheng IIRC, the backend part should not block VM snapshot for third-party storage, do you mean it's blocked by GUI?

Vicente-Cheng:
Hi @WebberHuang1118, 

Yes, it's mainly the UI issue, but we need to provide some checking mechanisms (for snapshot capability) to make sure UI can fix it.

m-ildefons:
Did the fix for this go into v1.5.0-rc3?
Is there anything I should take into account when setting up the test environment? E.g. selecting a non-default UI source, or a specific setting for the VolumeSnapshotClass?

![Image](https://github.com/user-attachments/assets/b164b507-2924-4cd4-a8eb-f4fbe083995c)

Vicente-Cheng:
> Did the fix for this go into v1.5.0-rc3? Is there anything I should take into account when setting up the test environment? E.g. selecting a non-default UI source, or a specific setting for the VolumeSnapshotClass?
> 
> ![Image](https://github.com/user-attachments/assets/b164b507-2924-4cd4-a8eb-f4fbe083995c)

Hi @m-ildefons, 
Did you configure the csi-driver-config for lvm volumesnapshot class? (like the following figure)
[...]

PR #7950, opened 2025-03-27
api: add new checking mechanism for snapshot

Vicente-Cheng:

**Problem:**
The VM snapshot cannot be executed because we disable the backup with 3rd-party storage solution.
We need to redefine the flow of VM snapshots.

**Solution:**
We reuse the backup checking mechanism for the snapshot capability. After v1.5.0, we should split it out to ensure the snapshot can be executed once the corresponding CSI driver has the snapshot capability.

**Related Issue:**
https://github.com/harvester/harvester/issues/7944

**Test plan:**

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -4,6 +4,7 @@ import (
 	"github.com/rancher/apiserver/pkg/types"
 	"github.com/rancher/wrangler/v3/pkg/data/convert"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	ctlstoragev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/storage/v1"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -32,6 +33,7 @@ const (
 	abortMigration                   = "abortMigration"
 	findMigratableNodes              = "findMigratableNodes"
 	backupVM                         = "backup"
+	snapshotVM                       = "snapshot"
 	restoreVM                        = "restore"
 	createTemplate                   = "createTemplate"
 	addVolume                        = "addVolume"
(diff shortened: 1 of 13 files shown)


pkg/api/vm/formatter.go
```diff
@@ -294,6 +301,39 @@ func (vf *vmformatter) canDoBackup(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv
...
+				logrus.Errorf("Can't get PVC %s/%s, err: %+v", vm.Namespace, vol.PersistentVolumeClaim.ClaimName, err)
+				return false
+			}
+			provisioner := util.GetProvisionedPVCProvisioner(pvc, vf.scCache)
+			if find := util.GetCSIProvisionerSnapshotCapability(provisioner); !find {
+				return false
+			}
+		}
+	}
+	return true
```
Yu-Jack: Just a NIT.

```suggestion
	for _, vol := range volumes {
		if vol.PersistentVolumeClaim == nil {
			continue
		}
		pvc, err := vf.pvcCache.Get(vm.Namespace, vol.PersistentVolumeClaim.ClaimName)
		if err != nil {
			if apierrors.IsNotFound(err) {
				continue
			}
			logrus.Errorf("Can't get PVC %s/%s, err: %+v", vm.Namespace, vol.PersistentVolumeClaim.ClaimName, err)
			return false
		}
		provisioner := util.GetProvisionedPVCProvisioner(pvc, vf.scCache)
		if !util.GetCSIProvisionerSnapshotCapability(provisioner) {
			return false
		}
	}
	return true
```

Yu-Jack (dismissed):
LGTM, thanks!

pkg/api/vm/formatter.go
Vicente-Cheng: Thanks for the suggestions!
Updated!

WebberHuang1118 (approved):
LGTM, thanks.

Yu-Jack (approved):

Vicente-Cheng:
Hi @bk201, @ihcsim

I will merge it first for the RC-3.
Feel free to give any suggestions here. I can update them later.

Issue #7840, opened 2025-03-13
[BUG] Upgrade 1.4.0 -> 1.4.1 stuck: instance-manager ... count is not 1


SlavikCA:
**Environment**
 - Harvester ISO version: 1.4.0
 - Underlying Infrastructure:
   - 2 Dell Precision servers with dual Xeon Gold CPU and 192GB+ RAM
   - 1 witness VM

**Describe the bug**
- started upgrade from UI
- after about 24 hours, I see that UI shows it's 1.4.1 now:

![Image](https://github.com/user-attachments/assets/1fcc6e1b-59c8-49c2-85fb-2dc8771566cb)

- but list of nodes, shows it still at 1.4.0:
```
kubectl get nodes -o wide                                                    
NAME      STATUS   ROLES                       AGE    VERSION          INTERNAL-IP    EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION                 CONTAINER-RUNTIME
t7820     Ready    control-plane,etcd,master   60d    v1.29.9+rke2r1   192.168.0.75   <none>        Harvester v1.4.0   5.14.21-150500.55.83-default   containerd://1.7.21-k3s2
t7920     Ready    control-plane,etcd,master   178d   v1.29.9+rke2r1   192.168.0.83   <none>        Harvester v1.4.0   5.14.21-150500.55.83-default   containerd://1.7.21-k3s2
witness   Ready    etcd                        87d    v1.29.9+rke2r1   192.168.0.63   <none>        Harvester v1.4.0   5.14.21-150500.55.83-default   containerd://1.7.21-k3s2
```
nodes has not been rebooted.

I found this error:
```
kubectl logs -n harvester-system -l harvesterhci.io/upgradeComponent=manifest                               
  enabled: false
noDefaultAdmin: false
rancherImage: rancher/rancher
rancherImagePullPolicy: IfNotPresent
rancherImageTag: v2.9.3
replicas: -3
systemDefaultRegistry: ""
tls: external
useBundledSystemChart: true
Error: UPGRADE FAILED: context deadline exceeded
instance-manager (aio) (image=longhornio/longhorn-instance-manager:v1.7.2) count is not 1 on node t7820, will retry...
instance-manager (aio) (image=longhornio/longhorn-instance-manager:v1.7.2) count is not 1 on node t7820, will retry...
instance-manager (aio) (image=longhornio/longhorn-instance-manager:v1.7.2) count is not 1 on node t7820, will retry...
[...]

PR #7952, opened 2025-03-27
upgrade: improve the checking mechanism for waiting instance-manager

Vicente-Cheng:

**Problem:**
The upgrade will get stuck once we enable the v2 data engine.

**Solution:**
We might have two aio instance-manager for both v1 and v2 data engine so we need to improve the checking script to ensure the v1 and v2 data engine alive.

**Related Issue:**
https://github.com/harvester/harvester/issues/7840

**Test plan:**
1. create harvester cluster v1.4.0 or later version
2. enable v2 data engine
3. upgrade should work as usual

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -489,24 +489,39 @@ wait_longhorn_instance_manager_aio() {
   # Wait for instance-manager (aio) pods upgraded to new version first.
   kubectl get nodes.longhorn.io -n longhorn-system -o json | jq -r '.items[].metadata.name' | while read -r node; do
     echo "Checking instance-manager (aio) pod on node $node..."
-    while [ true ]; do
-      im_count=$(kubectl get instancemanager.longhorn.io --selector=longhorn.io/node=$node,longhorn.io/instance-manager-type=aio,longhorn.io/instance-manager-image=$im_image_checksum -n longhorn-system -o json | jq -r '.items | length')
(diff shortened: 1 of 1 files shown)


bk201 (approved):

starbops (approved):
LGTM, thank you!

package/upgrade/upgrade_manifests.sh
```diff
@@ -489,24 +489,39 @@ wait_longhorn_instance_manager_aio() {
...
-      fi
+    check_instance_manager $node $im_image $im_image_checksum "v1"
 
-      im_status=$(kubectl get instancemanager.longhorn.io --selector=longhorn.io/node=$node,longhorn.io/instance-manager-type=aio,longhorn.io/instance-manager-image=$im_image_checksum -n longhorn-system -o json | jq -r '.items[0].status.currentState')
-      if [ "$im_status" != "running" ]; then
-        echo "instance-manager (aio) (image=$im_image) state is not running on node $node, will retry..."
-        sleep 5
-        continue
-      fi
+    v2EngineEnabled=$(kubectl get settings.harvesterhci.io longhorn-v2-data-engine-enabled -o yaml | yq -e e '.value' -)
```
FrankYang0529, line 494: @Vicente-Cheng I'm testing v1.4.2 to v1.5-head. This line throws error like following if there is no value field:

```
null
Error: no matches found
```

I can fix this by removing `-e`. Before I filing a PR, I would like to know is it ok to remove it? Thanks.

Issue #2788, opened 2022-09-20
[BUG] Namespace harvester-public links.update is not shown correctly


futuretea:
**Describe the bug**

**To Reproduce**

Use admin user to

1. Create a standard user `test-member`
2. Add this user to an imported cluster as Cluster Member role

Use cluster member user to

3. view namespace API, we can see the namespace harvester-public doesn't have links.update
4. create a new project,
5. create a new namespace, and add it to the new project
6. view namespace API, we can see the namespace harvester-public have links.update (In fact, this user does not have permission to update namespace `harvester-public`)

**Expected behavior**

namespace `harvester-public` links.update is shown correctly

**Support bundle**

**Environment**
 - Harvester ISO version: 
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630):

**Additional context**

mantis-toboggan-md:
@futuretea I hit something very similar in cluster explorer and filed a rancher/rancher ticket here: https://github.com/rancher/rancher/issues/39129

guangbochen:
Temporarily moving this issue to the Harvester v1.3.0 milestone, as this issue will depend on an upstream fix from https://github.com/rancher/rancher/issues/39129, which is scheduled for [2023-Q2-v2.7x](https://github.com/rancher/rancher/milestone/300)

ericpromislow:
The underlying issue (39129) is fixed on 2.10.3 and 2.11.0-alpha*

TachunLin:
Verified fixed on `v1.6-c46cdb38-head` (25/07/23) with Rancher `v2.12.0-rc1`. Close this issue.

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ Cluster Member role can't see link.update in namespace API  $~~$</summary>

  1. Access Harvester with Rancher user `test-member` have Cluster member role
  2. When view api of the namespace `harvester-public`, there is no links.update shown  

  <img width="1600" height="831" alt="Image" src="https://github.com/user-attachments/assets/1584f527-9632-4b3a-b448-708f7dbb506c" />

  https://github.com/user-attachments/assets/987c55be-bbd9-4003-ae0b-1f13c0685e16
  
 </details>

<details><summary>$\color{green}{\textsf{PASS}}$ Cluster Member role can't see link.update in namespace API after add a new custom project and namespace $~~$</summary>

  1. Access Harvester with Rancher user `test-member` have Cluster member role
  2. And create a new project `test-project` and namespace `test-ns`
[...]

PR #7972, opened 2025-03-31
Cluster member should not be able to update harvester-public namespace

FrankYang0529:

**Problem:**
A cluster member user can see Edit Config button even if the user doesn't have permission to update a namespace.

**Solution:**
The root cause in in upstream. Update steve package version to fix the error.

https://github.com/rancher/steve/commit/d3269c91eb36b93798dd69aff8eeee2914fa7c3b

**Related Issue:**
https://github.com/harvester/harvester/issues/2788

**Test plan:**
Use admin user to

1. Create a standard user `test-member`
2. Add this user to an imported cluster as Cluster Member role

Use cluster member user to

3. view namespace API, we can see the namespace harvester-public doesn't have links.update
4. create a new project,
5. create a new namespace, and add it to the new project
6. view namespace API, user should not see the namespace harvester-public with links.update.


ibrokethecloud (approved):
lgtm. thanks.

ibrokethecloud (commented):
related change in steve: https://github.com/rancher/steve/pull/308

ibrokethecloud (approved):

bk201 (approved):

Issue #7835, opened 2025-03-13
[TASK] Bumping RKE2 and Rancher for 1.5


bk201:
## What's the task? Please describe

Bump RKE2 to v1.32 and Rancher to 2.11.0 for 1.5.0.

innobead:
This can be covered in upgrade testing to prevent duplicated efforts. Assigning to @noahgildersleeve first and see if this should be delegated to someone later. 

cc @harvester/qa @harvester/release

noahgildersleeve:
Tested in 2 node libvirt KVM cluster on Harvester `v1.5.0-rc4`. RKE2 showing as `v1.32.3+rke2r1`

```
Client Version: v1.29.9
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.32.3+rke2r1
WARNING: version difference between client (1.29) and server (1.32) exceeds the supported minor version skew of +/-1
```

![Image](https://github.com/user-attachments/assets/f915a376-4b93-40d5-9ea5-5764ebe55472)

PR #7977, opened 2025-04-01
fix(upgrade_manifest): fix getting longhorn-v2-data-engine-enabled if value is empty

FrankYang0529:

If value of longhorn-v2-data-engine-enabled setting is empty, the `yq -e` throws the error. Remove `-e`.

**Related Issue:**
https://github.com/harvester/harvester/issues/7835, https://github.com/harvester/harvester/issues/7840

**Test plan:**
Upgrade from v1.4.2 to v1.5-head without error.


Vicente-Cheng (approved):
thanks for fixing that.

starbops (approved):
LGTM, thanks!

Issue #2811, opened 2022-09-22
[FEATURE] Online resize of disks / volumes attached to virtual machines


Martin-Weiss:
**Is your feature request related to a problem? Please describe.**
We need to be able to increase the size of disks / volumes attached to virtual machines while they are up and running.
This also needs to be possible via terraform via harvester_virtualmachine.

Tried in the UI on edit volume on a virtual machine and then got this message:
`admission webhook “[validator.harvesterhci.io](http://validator.harvesterhci.io/)” denied the request: resizing is only supported for detached volumes. The volume is being used by VM default/rmt. Please stop the VM first.”`

**Describe the solution you'd like**
Incresing the size of attached volumes while the VM is running is required to resize without downtime.
When using terraform this should also be possible via changing disk size in harvester_virtualmachine (terraform import is not a feasable process).

**Describe alternatives you've considered**
Shutdown the VM - but that is not acceptable because it causes downtime.

yasker:
Depends on https://github.com/longhorn/longhorn/issues/1674 which is targeted for Longhorn v1.4 (Q4 2022).

After Longhorn incorporated the change, we should be able to support it in Harvester's later release.

haiwu:
Now that the above longhorn issue is already closed, when would Harvester be able to support this feature?

ibrokethecloud:
This is currently targeted for harvester v1.2.0

haiwu:
Thanks for the update!

hsinhoyeh:
add a few notes here for running an online resize of the disk.

background: we have a VM running with 200G storage but we are running out of space. 
so we need to resize the disk from 200 G to 500 G.

After making changes to the UI (steps like stopping VM, resizing the volume, and then restarting it again)

we run the `lsblk` to see the volume:

```
lsblk -a

vda    252:0    0  500G  0 disk
├─vda1 252:1    0    1M  0 part
└─vda2 252:2    0  200G  0 part /
```

it did show the total size is 500G but only 200G is used (as expected).

so I run fdisk to claim the remaining free space

```
fdisk /dev/vda

//Delete the previous partition
d
2
// then recreate a new one to claim all spaces
n
2

w
```

then check fdisk with `fdisk -l` 

```
Device     Start        End    Sectors  Size Type
/dev/vda1   2048       4095       2048    1M BIOS boot
/dev/vda2   4096 1048575966 1048571871  500G Linux filesystem
``` 

then finally, we run `resize2fs` to the target partition

```
[...]

PR #7978, opened 2025-04-01
Support volume online expand

WebberHuang1118:
**Problem:**
Online Resize of Disks/Volumes Attached to Virtual Machines

**Solution:**
Leverage the KubeVirt `ExpandDisks` feature gate to expose volume expansion capabilities to the VM. Since we already support [third-party storage starting from v1.5.0](https://docs.harvesterhci.io/v1.6/advanced/csidriver), it’s also necessary to handle various storage providers with different PVC modes.

**Related Issue:**
#2811

**Test Plan:**

**Prerequisite:**
- Build a Harvester cluster using the ISO from this PR.

**Case 1: Longhorn v1 Block Mode PVC**
- Create a VM with a Longhorn v1 PVC as the root disk.
- Expand the root disk PVC size.
- Verify that the corresponding block device inside the VM reflects the increased size.

**Case 2: Longhorn v2 Block Mode PVC**
- Create a VM with a Longhorn v2 PVC as the root disk.
- Attempt to expand the root disk PVC size.
- Verify that the PVC expansion is denied by the Harvester webhook.

**Case 3: NFS CSI PVC with Filesystem Mode**
- [Deploy the NFS CSI driver on Harvester](https://docs.harvesterhci.io/v1.6/advanced/csidriver/#how-to-deploy-the-nfs-csi-driver).
- Create a VM with an NFS CSI PVC as the root disk.
- Attempt to expand the root disk PVC size.
- Verify that the PVC expansion is denied by the Harvester webhook.
- Enable the CSI-Driver-Validation setting for the NFS CSI driver.
  ```
  default: '{"driver.longhorn.io":true}'
  kind: Setting
  metadata:
    name: csi-online-expand-validation
  status: {}
  value: '{"nfs.csi.k8s.io":true}'
  ```
- Attempt to expand the root disk PVC size again.
- Verify that the corresponding block device inside the VM reflects the increased size.

**Case 4: TopoLVM PVC as a Hotplug Volume**
- [Deploy the TopoLVM CSI driver on Harvester](https://github.com/topolvm/topolvm).
- Create a VM with the root disk provisioned by any CSI provider.
- Create a PVC using the TopoLVM CSI driver.
- Hotplug the TopoLVM PVC into the running VM.
- Verify that the VM detects the hotplugged volume.
- Enable the CSI-Driver-Validation setting for the TopoLVM CSI driver.
  ```
  default: '{"driver.longhorn.io":true}'
  kind: Setting
  metadata:
    name: csi-online-expand-validation
  status: {}
  value: '{"nfs.csi.k8s.io":true, "topolvm.io":true}'
  ```
- Expand the size of the TopoLVM PVC.
- Verify that the corresponding block device inside the VM reflects the increased size.

**Additional Context:**
- This PR will [deny online expansion for a filesystem mode PVC that is hotplugged to a running VM](https://github.com/WebberHuang1118/harvester/blob/6af58d85842595fdb03f7268cd9d336006d996ef/pkg/webhook/resources/persistentvolumeclaim/validator.go#L187-L200). 
- This is a corner case where PVC resizing may get stuck waiting for NodeExpand(). For more details, please refer to [this document](https://github.com/WebberHuang1118/harvester-study/blob/master/volume-online-resize/README.md#filesystem-mode-pvc-corner-case).

### DIFF
--- a/pkg/controller/global/settings/register.go
+++ b/pkg/controller/global/settings/register.go
@@ -5,9 +5,8 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/rancher/steve/pkg/server"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -36,21 +35,32 @@ type settingsProvider struct {
 }
 
 func (s *settingsProvider) Get(name string) string {
+	logrus.Debugf("Fetching setting: %s", name)
+
 	value := os.Getenv(settings.GetEnvKey(name))
 	if value != "" {
+		logrus.Debugf("Setting %s found in environment variable with value: %s", name, value)
 		return value
 	}
+
+	logrus.Infof("Attempting to fetch setting %s from cache", name)
 	obj, err := s.settingsLister.Get(name)
 	if err != nil {
+		logrus.Warnf("Failed to fetch setting %s from cache, attempting direct API call: %v", name, err)
 		val, err := s.settings.Get(name, v1.GetOptions{})
 		if err != nil {
+			logrus.Errorf("Failed to fetch setting %s from API, falling back to default: %v", name, err)
 			return s.fallback[name]
 		}
 		obj = val
 	}
+
 	if obj.Value == "" {
+		logrus.Debugf("Setting %s has no value, using default: %s", name, obj.Default)
 		return obj.Default
 	}
+
+	logrus.Infof("Setting %s found with value: %s", name, obj.Value)
 	return obj.Value
 }
 
--- a/pkg/settings/settings.go
+++ b/pkg/settings/settings.go
@@ -46,6 +46,7 @@ var (
 	VipPools                               = NewSetting(VipPoolsConfigSettingName, "")
 	AutoDiskProvisionPaths                 = NewSetting(AutoDiskProvisionPathsSettingName, "")
 	CSIDriverConfig                        = NewSetting(CSIDriverConfigSettingName, `{"driver.longhorn.io":{"volumeSnapshotClassName":"longhorn-snapshot","backupVolumeSnapshotClassName":"longhorn"}}`)
+	CSIOnlineExpandValidation              = NewSetting(CSIOnlineExpandValidationSettingName, `{"driver.longhorn.io":true}`)
 	ContainerdRegistry                     = NewSetting(ContainerdRegistrySettingName, "")
 	StorageNetwork                         = NewSetting(StorageNetworkName, "")
 	DefaultVMTerminationGracePeriodSeconds = NewSetting(DefaultVMTerminationGracePeriodSecondsSettingName, "120")
@@ -75,6 +76,7 @@ const (
 	DefaultDashboardUIURL                             = "https://releases.rancher.com/harvester-ui/dashboard/latest/index.html"
 	SupportBundleImageName                            = "support-bundle-image"
 	CSIDriverConfigSettingName                        = "csi-driver-config"
+	CSIOnlineExpandValidationSettingName              = "csi-online-expand-validation"
 	UIIndexSettingName                                = "ui-index"
 	UIPathSettingName                                 = "ui-path"
 	UISourceSettingName                               = "ui-source"
--- a/pkg/util/volume.go
+++ b/pkg/util/volume.go
@@ -103,3 +103,36 @@ func IndexPodByPVCFunc(pod *corev1.Pod) ([]string, error) {
 	}
 	return indexes, nil
 }
+
+// parseJSONSetting parses a JSON string into the provided target interface.
+func parseJSONSetting(setting string, target interface{}) error {
+	if setting == "" {
+		return fmt.Errorf("setting is empty")
+	}
+	return json.Unmarshal([]byte(setting), target)
+}
+
+func GetCSIOnlineExpandValidation(
+	provisioner string,
+	settingCache ctlharvesterv1.SettingCache,
+) (bool, error) {
+	if provisioner == "" {
+		return false, fmt.Errorf("provisioner is empty")
+	}
+
+	coevSetting, err := settingCache.Get(settings.CSIOnlineExpandValidationSettingName)
+	if err != nil {
+		return false, fmt.Errorf("failed to get %s setting: %w", settings.CSIOnlineExpandValidationSettingName, err)
+	}
+
+	coev := make(map[string]bool)
+	for _, data := range []string{coevSetting.Default, coevSetting.Value} {
+		if data != "" {
+			if err := parseJSONSetting(data, &coev); err != nil {
+				return false, err
+			}
+		}
+	}
+
+	return coev[provisioner], nil
+}
--- a/pkg/webhook/resources/persistentvolumeclaim/validator.go
+++ b/pkg/webhook/resources/persistentvolumeclaim/validator.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	ctlstoragev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/storage/v1"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -12,33 +13,48 @@ import (
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	cdicommon "kubevirt.io/containerized-data-importer/pkg/controller/common"
+	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 
 	harvesterv1beta1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlkv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	ctllonghornv1 "github.com/harvester/harvester/pkg/generated/controllers/longhorn.io/v1beta2"
 	"github.com/harvester/harvester/pkg/ref"
 	"github.com/harvester/harvester/pkg/util"
 	indexeresutil "github.com/harvester/harvester/pkg/util/indexeres"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/indexeres"
 	"github.com/harvester/harvester/pkg/webhook/types"
+	webhookutil "github.com/harvester/harvester/pkg/webhook/util"
 )
 
 func NewValidator(pvcCache v1.PersistentVolumeClaimCache,
 	vmCache ctlkv1.VirtualMachineCache,
-	imageCache ctlharvesterv1.VirtualMachineImageCache) types.Validator {
+	kubevirtCache ctlkv1.KubeVirtCache,
+	imageCache ctlharvesterv1.VirtualMachineImageCache,
+	engineCache ctllonghornv1.EngineCache,
+	scCache ctlstoragev1.StorageClassCache,
+	settingCache ctlharvesterv1.SettingCache) types.Validator {
 	return &pvcValidator{
-		pvcCache:   pvcCache,
-		vmCache:    vmCache,
-		imageCache: imageCache,
+		pvcCache:      pvcCache,
+		vmCache:       vmCache,
+		kubevirtCache: kubevirtCache,
+		imageCache:    imageCache,
+		engineCache:   engineCache,
+		scCache:       scCache,
+		settingCache:  settingCache,
 	}
 }
 
 type pvcValidator struct {
 	types.DefaultValidator
-	pvcCache   v1.PersistentVolumeClaimCache
-	vmCache    ctlkv1.VirtualMachineCache
-	imageCache ctlharvesterv1.VirtualMachineImageCache
+	pvcCache      v1.PersistentVolumeClaimCache
+	vmCache       ctlkv1.VirtualMachineCache
+	imageCache    ctlharvesterv1.VirtualMachineImageCache
+	kubevirtCache ctlkv1.KubeVirtCache
+	engineCache   ctllonghornv1.EngineCache
+	scCache       ctlstoragev1.StorageClassCache
+	settingCache  ctlharvesterv1.SettingCache
 }
 
 func (v *pvcValidator) Resource() types.Resource {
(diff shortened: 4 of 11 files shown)


package/upgrade/upgrade_manifests.sh
```diff
@@ -1305,6 +1305,7 @@ rules:
   resources:
   - networkfilesystems
   - networkfilesystems/status
+  - settings
```
ibrokethecloud: this will not run in the upgrade path 1.6.0 as we expect users to be atleast at v1.5.x. should we drop this?

pkg/util/volume.go
```diff
@@ -103,3 +103,36 @@ func IndexPodByPVCFunc(pod *corev1.Pod) ([]string, error) {
...
+		return false, fmt.Errorf("provisioner is empty")
+	}
+
+	coevSetting, err := settingCache.Get(settings.CSIOnlineExpandValidationSettingName)
+	if err != nil {
+		return false, fmt.Errorf("failed to get %s setting: %w", settings.CSIOnlineExpandValidationSettingName, err)
+	}
+
+	coev := make(map[string]bool)
+	for _, data := range []string{coevSetting.Default, coevSetting.Value} {
```
ibrokethecloud, line 129: should we merge the maps? the default will always have `longhorn` csi expansion enabled. Unless user explicitly adds `longhorn` provider to the list of values it will be dropped during validation

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -117,6 +132,91 @@ func (v *pvcValidator) Delete(request *types.Request, oldObj runtime.Object) err
 	return nil
 }
 
+func (v *pvcValidator) isOnlineExpandNeeded(pvc *corev1.PersistentVolumeClaim) (bool, error) {
+	vms, err := v.vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(pvc.Namespace, pvc.Name))
+	if err != nil {
+		return false, werror.NewInternalError(fmt.Sprintf("failed to get VMs by index: %s, PVC: %s/%s, err: %s", indexeresutil.VMByPVCIndex, pvc.Namespace, pvc.Name, err))
+	}
+	for _, vm := range vms {
```
ibrokethecloud, line 139: we are checking here if the VM is running to exit the validation early

there are a lot more conditions where we do not need to run this check, may be we just need to check if the VM is starting, running or migrating to check if online expansion is even needed
https://github.com/webberhuang1118/harvester/blob/5b5bea19729d1a751f368535a980f0e659ebe424/vendor/kubevirt.io/api/core/v1/types.go#L1593-L1594

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -117,6 +132,91 @@ func (v *pvcValidator) Delete(request *types.Request, oldObj runtime.Object) err
 	return nil
 }
```
ibrokethecloud, line 133: minor nit: we should may be add a comment that we are only checking for VM status to break validation early

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -117,6 +132,91 @@ func (v *pvcValidator) Delete(request *types.Request, oldObj runtime.Object) err
...
+		if f == virtconfig.ExpandDisksGate {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (v *pvcValidator) checkExpand(pvc *corev1.PersistentVolumeClaim) error {
+	if onlineExpand, err := v.isOnlineExpandNeeded(pvc); err != nil || !onlineExpand {
```
ibrokethecloud, line 175: in both `OnlineExpandNeeded` and `isHotpluggedFilesystemPVC` we fetch the vm from the cache using pvc name.

it might be better to make a single call in `checkExpand` and pass the vm object to both these methods

this way if VM is not using the pvc the check can exit early

pkg/util/volume.go
WebberHuang1118: @ibrokethecloud , do you mean that since the Longhorn setting is immutable, we should consider excluding Longhorn from this logic and treat it as a special case—especially because we need to handle the differences between the v1 and v2 engines?

pkg/webhook/resources/persistentvolumeclaim/validator.go
WebberHuang1118: We need to ensure the VM is fully quiesced; otherwise, we should treat it as an online expansion. Maybe checking whether the VM status is Stopped is sufficient—what do you think?

pkg/webhook/resources/persistentvolumeclaim/validator.go
WebberHuang1118: @ibrokethecloud Does this reference another part of the source code? It doesn’t seem to match what I’m seeing on my side.

I assume you're referring to the function call for `isOnlineExpandNeeded()`?

pkg/webhook/resources/persistentvolumeclaim/validator.go
WebberHuang1118: The function `isOnlineExpandNeeded()` uses the index `VMByPVCIndex`, while `isHotpluggedFilesystemPVC()` uses `VMByHotplugPVCIndex`, meaning they query different sets of VMs based on different scenarios.

package/upgrade/upgrade_manifests.sh
WebberHuang1118: Updated as suggested—thanks!

WebberHuang1118:
Hi @ibrokethecloud 
I've applied your suggestions and responded to your feedback. Please take a look, thanks!

pkg/webhook/resources/persistentvolumeclaim/validator.go
Vicente-Cheng: I am curious about the `VirtualMachineStatusMigrating`. What happens if we execute the volume expansion at this status?

pkg/controller/global/settings/register.go
```diff
@@ -36,21 +35,32 @@ type settingsProvider struct {
 }
 
 func (s *settingsProvider) Get(name string) string {
+	logrus.Debugf("Fetching setting: %s", name)
+
 	value := os.Getenv(settings.GetEnvKey(name))
 	if value != "" {
+		logrus.Infof("Setting %s found in environment variable with value: %s", name, value)
```
Vicente-Cheng: s/Infof/Debugf/

pkg/controller/global/settings/register.go
```diff
@@ -36,21 +35,32 @@ type settingsProvider struct {
...
 		val, err := s.settings.Get(name, v1.GetOptions{})
 		if err != nil {
+			logrus.Errorf("Failed to fetch setting %s from API, falling back to default: %v", name, err)
 			return s.fallback[name]
 		}
 		obj = val
 	}
+
 	if obj.Value == "" {
+		logrus.Infof("Setting %s has no value, using default: %s", name, obj.Default)
```
Vicente-Cheng: s/Infof/Debugf/

pkg/settings/settings.go
```diff
@@ -46,6 +46,7 @@ var (
 	VipPools                               = NewSetting(VipPoolsConfigSettingName, "")
 	AutoDiskProvisionPaths                 = NewSetting(AutoDiskProvisionPathsSettingName, "")
 	CSIDriverConfig                        = NewSetting(CSIDriverConfigSettingName, `{"driver.longhorn.io":{"volumeSnapshotClassName":"longhorn-snapshot","backupVolumeSnapshotClassName":"longhorn"}}`)
+	CSIOnlineExpandValidation              = NewSetting(CSIOnlineExpandValidationSettingName, `{"driver.longhorn.io":true}`)
```
Vicente-Cheng, line 49: I am curious why we disable the hotplug fs volume?

Do you mean the hotplug filesystem or the hotplug block volume, but actually is the filesystem pvc?
From the following validator, I thought you meant the block volume (for VM), but it is the filesystem pvc.
It should depend on the CSI side, why we need to disable it?

pkg/webhook/resources/persistentvolumeclaim/validator.go
Vicente-Cheng: Is the VMByHotplugPVCIndex a subset of VMByPVCIndex?
Could we consider using the one?

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -117,6 +132,92 @@ func (v *pvcValidator) Delete(request *types.Request, oldObj runtime.Object) err
...
+	// Check if the PVC is in Filesystem mode
+	if pvc.Spec.VolumeMode == nil || *pvc.Spec.VolumeMode != corev1.PersistentVolumeFilesystem {
+		return false, nil
+	}
+
+	// Check if the PVC is hotplugged to any VM
+	vms, err := v.vmCache.GetByIndex(indexeresutil.VMByHotplugPVCIndex, ref.Construct(pvc.Namespace, pvc.Name))
+	if err != nil {
+		return false, werror.NewInternalError(err.Error())
+	}
```
Vicente-Cheng, line 157: Just want to confirm, do we need to check the VM status here?

Vicente-Cheng:
BTW, please squash `d11ce2e9e79fc616ed0f18d379fb365d4cd87ce5` and `4cf82980bf5b0488ef20e419742897b22b036b3c`.

pkg/webhook/resources/persistentvolumeclaim/validator.go
WebberHuang1118: The PVC should still be referenced by at least one virt-launcher pod.
- If the CSI driver only reports `EXPAND_VOLUME` in `NodeGetCapabilities()` but not in `ControllerGetCapabilities()`, then `NodeExpandVolume()` will always be triggered by the `CSI Resizer`.
- If the CSI driver reports `EXPAND_VOLUME` in `ControllerGetCapabilities()`, the `CSI Resizer` will first trigger `ControllerExpandVolume()`, and then determine whether `NodeExpandVolume()` should be called.

In short, the `CSI Resizer` handles the controller flow, but the `CSI driver` determines what actually happens on the storage side.

I’ve provided more details in the related [HEP PR](https://github.com/harvester/harvester/blob/4fd55c53bf2973b59236e09cb593abf205f3d8ea/enhancements/20250612-volume-online-expansion.md#why-we-need-the-setting-csi-online-expand-validation).

You can also find additional notes from my study here:
https://github.com/WebberHuang1118/harvester-study/tree/master/volume-online-resize#csi-resizer-vs-volume-expand

pkg/settings/settings.go
WebberHuang1118: Yes, the block device is used for the VM, and it’s a filesystem mode PVC on the CSI side.

When a filesystem-mode PVC is hotplugged into the VM (i.e., added as a second volume), the existing `hp-volume` pod is replaced by a new one. This new `hp-volume `pod only handles the `NodeStageVolume()` and `NodePublishVolume()` calls for the second volume. As a result, the first hotplugged volume undergoes `NodeUnpublishVolume()` and `NodeUnstageVolume()`, but its image file remains bind-mounted in the `virt-launcher` pod.

When the `CSI Resizer` attempts to propagate a size expansion, the expected `Node*` gRPC calls (e.g., `NodeExpandVolume()`) are not triggered. Consequently, the resize operation for the PVC is incomplete, and the VM does not reflect the updated volume size for the hotplugged volume.

I’ve provided more details in the related [HEP PR](https://github.com/harvester/harvester/blob/4fd55c53bf2973b59236e09cb593abf205f3d8ea/enhancements/20250612-volume-online-expansion.md#why-we-prevent-hotplugged-filesystem-pvc-from-online-resizing).

Additional notes from my study can be found here:
https://github.com/WebberHuang1118/harvester-study/tree/master/volume-online-resize#filesystem-mode-pvc-corner-case

pkg/webhook/resources/persistentvolumeclaim/validator.go
WebberHuang1118: `VMByPVCIndex` is used to index VMs that have a specific PVC attached.
`VMByHotplugPVCIndex` is used to index VMs that hotplug a specific PVC.

In practice, we could just use `VMByPVCIndex` to find the VMs using a given PVC and then check whether the PVC is hotplugged. I introduced a separate index mainly to make the source code more compact and potentially improve readability (maybe?).

Do you have a strong preference for using only one index? If so, may I know the reason? Thanks!

pkg/webhook/resources/persistentvolumeclaim/validator.go
WebberHuang1118: IMO, we can stick with the current approach. The CSI Resizer will invoke the appropriate gRPC calls based on the status between the PVC and the VM (i.e., the `virt-launcher` pod).

If there’s an issue with the VM, it might cause the PVC resize process to get stuck at a certain phase. However, from my understanding, that would be due to the VM rather than the CSI logic. For example, restarting the VM can trigger the CSI Resizer to reconcile the PVC status.

pkg/controller/global/settings/register.go
WebberHuang1118: Updated as per suggestion—thanks.

pkg/controller/global/settings/register.go
WebberHuang1118: Updated as per suggestion—thanks.

WebberHuang1118:
Hi @Vicente-Cheng,
I’ve tried to address your feedback—please take a look when you have a moment. Thanks!

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -115,6 +131,92 @@ func (v *pvcValidator) Delete(request *types.Request, oldObj runtime.Object) err
 	return nil
 }
 
+func (v *pvcValidator) isOnlineExpandNeeded(pvc *corev1.PersistentVolumeClaim) (bool, error) {
```
ibrokethecloud, line 134: we need to fine tune this and also factor in pvc's attached to k8s native workloads like pods.

A common use case would be a pvc used for prometheus persistence, which may need to be resized.

In the current validation schema such a request will be rejected.

WebberHuang1118: > we need to fine tune this and also factor in pvc's attached to k8s native workloads like pods.
> 
> A common use case would be a pvc used for prometheus persistence, which may need to be resized.
> 
> In the current validation schema such a request will be rejected.

If I understand correctly, we don’t block volume expansion for PVCs used by native workloads. So I’ve kept the same behavior for PVCs not used by a VM.

ibrokethecloud (approved):
lgtm. thanks.

pkg/webhook/resources/persistentvolumeclaim/validator.go
Vicente-Cheng: Thanks for the explanation!

pkg/webhook/resources/persistentvolumeclaim/validator.go
Vicente-Cheng: OK, let's keep that for better readability. Thanks.

pkg/webhook/resources/persistentvolumeclaim/validator.go
Vicente-Cheng: That makes sense, thanks!

pkg/settings/settings.go
Vicente-Cheng: got it, very clear. Thanks!

Vicente-Cheng (approved):
lgtm, thanks!

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -115,6 +131,92 @@ func (v *pvcValidator) Delete(request *types.Request, oldObj runtime.Object) err
...
+	return false, nil
+}
+
+func (v *pvcValidator) isHotpluggedFilesystemPVC(pvc *corev1.PersistentVolumeClaim) (bool, error) {
+	// Check if the PVC is in Filesystem mode
+	if pvc.Spec.VolumeMode == nil || *pvc.Spec.VolumeMode != corev1.PersistentVolumeFilesystem {
+		return false, nil
+	}
+
+	// Check if the PVC is hotplugged to any VM
```
ihcsim, line 153: IIUC, it's not just whether a PVC is hotplugged or not. More importantly, it's whether the `hp-volume` pod is referencing it or not, right? Do you know if this is a known upstream issue or known kubevirt expected behaviour?

WebberHuang1118: Hi @ihcsim,

@Vicente-Cheng and I had a similar discussion here: [link](https://github.com/harvester/harvester/pull/7978#discussion_r2147758675). You might find it helpful to take a look.

It seems there’s no related upstream issue—possibly because it occurs in a corner case and involves Kubernetes’ built-in behavior.

ihcsim: Yeah, I saw that comment and the HEP. I just assume the code needs to check if the corresponding `hp-volume` pod still exists in order to determine if this is the VM's 1st or N-th hotplugged volume.

WebberHuang1118: Oh, I see your point. However, during the volume resizing process, it's still possible for another volume—especially one provisioned via the Harvester CSI driver—to be hotplugged into the VM. Even if we check whether the volume is the first hotplugged volume beforehand, there's still a potential race condition. As a result, the outcome of resizing a filesystem PVC cannot be fully guaranteed.

While we could consider blocking additional volume hotplug operations while a volume is being resized, this might introduce unnecessary complexity or impact performance. Given that online volume resizing is still a relatively new feature, I'd prefer to keep the current design for now and revisit safeguards if needed in the future.

Issue #7871, opened 2025-03-17
[FEATURE] Add option to change runStrategy  while cloning the VM


rajeshkio:
**Is your feature request related to a problem? Please describe.**

When cloning a VM, the user should be able to change the runStrategy to change the behaviour of the cloned VM. They should be able to create a VM clone not boot them.

**Describe the solution you'd like**

Give the option to change the runStrategy so that users can choose if they want to boot the clone or not.

rrajendran17:
@rajeshkio do you have a particular usecase for this scenario ?

ishimoto-331188:
@rrajendran17 
I would like to create  a VM with the same configuration as the source VM, except for the IP settings. However, if the cloned VM starts automatically before changing its IP settings, it will conflict with the existing VM's IP address.

rrajendran17:
Thanks @ishimoto-331188, that makes sense.
IIUC,if you could change the runStrategy on a cloned VM, then cloned VM will be in off state, then you would use cloud init to modify network settings and boot the cloned VM.

ishimoto-331188:
> Thanks [@ishimoto-331188](https://github.com/ishimoto-331188), that makes sense. IIUC,if you could change the runStrategy on a cloned VM, then cloned VM will be in off state, then you would use cloud init to modify network settings and boot the cloned VM.

How Can I do that?

albinsun:
Need to test with GUI implementation
* #7986

m-ildefons:
## Setup

- Setup Harvester v1.6.0-rc4
- Create new Virtual Machine

<img width="2402" height="1353" alt="Image" src="https://github.com/user-attachments/assets/c07ac07e-9afc-43f6-bb8b-6e6c48690eb4" />

## Verification

🟢 Clone Dialogue offers a possibility to change the run strategy

<img width="2434" height="1514" alt="Image" src="https://github.com/user-attachments/assets/2ece1602-8151-486a-85e5-aa3cb5cf045a" />

🟢 New run strategy correctly applied to cloned VM

<img width="2433" height="1511" alt="Image" src="https://github.com/user-attachments/assets/498717a0-6703-4ef3-8a09-73061cc155b8" />

## Conclusion

🟢 Successfully verified

PR #7985, opened 2025-04-02
feat: add runStrategy to clone api

FrankYang0529:

Add `runStrategy` field to clone API request body.

**Related Issue:**
https://github.com/harvester/harvester/issues/7871

**Test plan:**
Case 1: The valid value of `runStrategy` is emtpy, `Always`, `Halted`, `Manual`, `RerunOnFailure`, and `Once`. Other value will get error.

Case 2: Clone a VM with empty run strategy. New VM will start with run strategy from original VM.

Case 3: Clone a VM with `Halted` run strategy. New VM will not start automatically.

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -1143,7 +1143,7 @@ func (h *vmActionHandler) cloneVM(name string, namespace string, input CloneInpu
 	if err != nil {
 		return fmt.Errorf("cannot get vm %s/%s, err: %w", namespace, name, err)
 	}
-	newVM := getClonedVMYamlFromSourceVM(input.TargetVM, vm)
+	newVM := getClonedVMYamlFromSourceVM(input, vm)
 
 	newPVCs, secretNameMap, err := h.cloneVolumes(newVM)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/api/vm/handler.go
```diff
@@ -1143,7 +1143,7 @@ func (h *vmActionHandler) cloneVM(name string, namespace string, input CloneInpu
 	if err != nil {
 		return fmt.Errorf("cannot get vm %s/%s, err: %w", namespace, name, err)
 	}
-	newVM := getClonedVMYamlFromSourceVM(input.TargetVM, vm)
+	newVM := getClonedVMYamlFromSourceVM(input.TargetVM, vm, kubevirtv1.VirtualMachineRunStrategy(input.RunStrategy))
```
ihcsim: Nit: just pass in the entire `input` object?

ihcsim (dismissed):

pkg/api/vm/handler.go
FrankYang0529: Address it. Thanks.

pkg/api/vm/handler.go
```diff
@@ -1428,6 +1428,11 @@ func getClonedVMYamlFromSourceVM(newVMName string, sourceVM *kubevirtv1.VirtualM
 			newVM.Annotations[cloneVMAnnoKey] = sourceAnnoValue
 		}
 	}
+
+	runStrategy := kubevirtv1.VirtualMachineRunStrategy(input.RunStrategy)
```
w13915984028, line 1432: How to ensure this statemment ?
```
Case 1: The valid value of runStrategy is emtpy, Always, Halted, Manual, RerunOnFailure, and Once. Other value will get error.
```

FrankYang0529: @w13915984028 , kubevirt webhook handle for this part. Do we want to add related validation in API?

https://github.com/kubevirt/kubevirt/blob/f75b04afa5750369ea9f81a7693310db5c6313f6/pkg/virt-api/webhooks/validating-webhook/admitters/vms-admitter.go#L230-L264

ihcsim (approved):

w13915984028 (approved):
LGTM, thanks.

PR #7991, opened 2025-04-02
image/cdi: add validator for VMImage with CDI backend

Vicente-Cheng:

**Problem:**
We are lacking some validator for the VMImage with CDI backend

**Solution:**
Implement the needed validator for VMImage with CDI backend

**Related Issue:**
https://github.com/harvester/harvester/issues/7854

**Test plan:**

### DIFF
--- a/pkg/image/cdi/validate.go
+++ b/pkg/image/cdi/validate.go
@@ -3,21 +3,56 @@ package cdi
 import (
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/image/backend"
+	"github.com/harvester/harvester/pkg/image/common"
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
 type Validator struct {
+	vmiv common.VMIValidator
 }
 
-func GetValidator() backend.Validator {
-	return &Validator{}
+func GetValidator(vmiv common.VMIValidator) backend.Validator {
+	return &Validator{vmiv}
 }
 
-func (cv *Validator) Create(_ *types.Request, _ *harvesterv1.VirtualMachineImage) error {
+func (cv *Validator) Create(req *types.Request, vmImg *harvesterv1.VirtualMachineImage) error {
+	if err := cv.vmiv.CheckDisplayName(vmImg); err != nil {
+		return err
+	}
+
+	if err := cv.vmiv.CheckURL(vmImg); err != nil {
+		return err
+	}
+
+	if err := cv.vmiv.CheckPVCInUse(vmImg); err != nil {
+		return err
+	}
+
+	if err := cv.vmiv.CheckImagePVC(req, vmImg); err != nil {
+		return err
+	}
 	return nil
 }
 
-func (cv *Validator) Update(_, _ *harvesterv1.VirtualMachineImage) error {
+func (cv *Validator) Update(oldVMImg, newVMImg *harvesterv1.VirtualMachineImage) error {
+	if err := cv.vmiv.SourceTypeConsistency(oldVMImg, newVMImg); err != nil {
+		return err
+	}
+
+	if cv.vmiv.IsExportVolume(newVMImg) {
+		if err := cv.vmiv.PVCConsistency(oldVMImg, newVMImg); err != nil {
+			return err
+		}
+	}
+
+	if err := cv.vmiv.URLConsistency(oldVMImg, newVMImg); err != nil {
+		return err
+	}
+
+	if err := cv.vmiv.CheckDisplayName(newVMImg); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/pkg/image/common/validator.go
+++ b/pkg/image/common/validator.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"fmt"
+	"net/url"
 	"reflect"
 	"strings"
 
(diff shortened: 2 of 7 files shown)


pkg/image/cdi/validate.go
```diff
@@ -3,21 +3,56 @@ package cdi
...
 
-func GetValidator() backend.Validator {
-	return &Validator{}
+func GetValidator(vmiv common.VMIValidator) backend.Validator {
+	return &Validator{vmiv}
 }
 
-func (cv *Validator) Create(_ *types.Request, _ *harvesterv1.VirtualMachineImage) error {
+func (cv *Validator) Create(req *types.Request, vmImg *harvesterv1.VirtualMachineImage) error {
+	if err := cv.vmiv.CheckDisplayName(vmImg); err != nil {
```
ihcsim, line 19: Just want to confirm the current implementation of `CheckDisplayName()` at: https://github.com/harvester/harvester/blob/ce188325a591813e047d3bd5991839f704eda1f5/pkg/image/common/validator.go#L90-L95

Shouldn't it be:

```go
	for _, image := range sameDisplayNameImages {
		if vmi.Name == image.Name {
                    return werror.NewConflict("A resource with the same display name exists")
		}
                continue
	}
```

pkg/image/cdi/validate.go
```diff
@@ -3,21 +3,56 @@ package cdi
...
+		return err
+	}
+
+	if err := cv.vmiv.CheckURL(vmImg); err != nil {
```
ihcsim, line 23: What do we think of adding [`url.Parse(vmi.Spec.URL)`](https://pkg.go.dev/net/url#Parse) in `CheckURL()` to make sure it's well-formed URL?

pkg/util/volume.go
```diff
@@ -88,3 +89,17 @@ func LoadCSIDriverConfig(settingCache ctlharvesterv1.SettingCache) (map[string]s
 	}
 	return csiDriverConfig, nil
 }
+
+func IndexPodByPVCfunc(pod *corev1.Pod) ([]string, error) {
```
ihcsim:
```suggestion
func IndexPodByPVCFunc(pod *corev1.Pod) ([]string, error) {
```
Nit.

pkg/image/cdi/validate.go
Vicente-Cheng: Thanks, I do some enhancement for this checking function

pkg/image/cdi/validate.go
Vicente-Cheng: ditto, updated the enhancement

pkg/util/volume.go
Vicente-Cheng: updated

pkg/image/common/validator.go
```diff
@@ -87,10 +95,7 @@ func (v *vmiValidator) CheckDisplayName(vmi *v1beta1.VirtualMachineImage) error
 	if err != nil {
 		return err
 	}
-	for _, image := range sameDisplayNameImages {
-		if vmi.Name == image.Name {
-			continue
-		}
+	if len(sameDisplayNameImages) != 0 {
```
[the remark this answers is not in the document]
Vicente-Cheng: yes, this might happen in the update case.
We need to ignore the same object

WebberHuang1118 (dismissed):
LGTM, thanks.

ibrokethecloud (dismissed):
lgtm. thanks.

pkg/image/common/validator.go
```diff
@@ -186,6 +203,21 @@ func (v *vmiValidator) CheckSecurityParameters(vmi *v1beta1.VirtualMachineImage)
 	return nil
 }
 
+// CheckPVCInUse checks if the PVC is in use by any pods, this is only used to CDI backend
+func (v *vmiValidator) CheckPVCInUse(vmi *v1beta1.VirtualMachineImage) error {
+	index := fmt.Sprintf("%s-%s", vmi.Spec.PVCNamespace, vmi.Spec.PVCName)
+	if pods, err := v.podCache.GetByIndex(util.IndexPodByPVC, index); err == nil && len(pods) > 0 {
+		podList := []string{}
+		for _, pod := range pods {
+			logrus.Infof("[DEBUG]: ns: %v, name: %v", pod.GetNamespace(), pod.GetName())
```
brandboat: nit: could we remove this debug message?

brandboat (dismissed):
LGTM, thanks!

pkg/image/common/validator.go
Vicente-Cheng: yes, we should... nice catch! updated!

brandboat (dismissed):

WebberHuang1118 (dismissed):

pkg/image/common/validator.go
```diff
@@ -186,6 +202,20 @@ func (v *vmiValidator) CheckSecurityParameters(vmi *v1beta1.VirtualMachineImage)
 	return nil
 }
 
+// CheckPVCInUse checks if the PVC is in use by any pods, this is only used to CDI backend
```
ihcsim, line 205: > this is only used to CDI backend

this looks like a general function that can be re-used. if not, then consider reinforcing it in the code, with something like `if vmi.Spec.Backend != VMIBackendCDI { return nil }` (or `return err`)?

ihcsim (dismissed):
LGTM - just one observation.

bk201 (dismissed):

pkg/image/common/validator.go
Vicente-Cheng: Nice suggestion!

Updated!

bk201 (approved):

WebberHuang1118 (approved):

Issue #8019, opened 2025-04-08
[BUG] GetImageStorageClassName() is hardcoding the name `longhorn`


votdev:
### Describe the Bug

The helper function `GetImageStorageClassName` is hardcoding the prefix of the returned StorageClass name with `longhorn` which might cause confusion if you are using using a custom `StorageClass`. Note, we are talking about the backingimage code path of this helper function.

References:
- https://github.com/harvester/harvester/blob/bab4a2cff8676d6e3d3fb1c8f1c710a5be114a0d/pkg/util/image.go#L72
- https://github.com/harvester/harvester/blob/e51d7ac4db4b19d7adbef5ca08f1b18dda21abb4/pkg/image/common/operator.go#L292

### To Reproduce

Create a custom StorageClass which is named `my-storage-class`, import a VM with `storageClass: my-storage-class` in the `VirtualMachineImport` CR.
```
apiVersion: migration.harvesterhci.io/v1beta1
kind: VirtualMachineImport
metadata:
  name: gm-ubuntu-test
  namespace: default
spec: 
  virtualMachineName: "gm-ubuntu-test"
  networkMapping:
  - sourceNetwork: "VM Network"
    destinationNetwork: "default/vlan1"
  storageClass: my-storage-class
  sourceCluster: 
    name: vcsim
    namespace: default
    kind: VmwareSource
    apiVersion: migration.harvesterhci.io/v1beta1
```

During the import the `VirtualMachineImage` CR
```
apiVersion: harvesterhci.io/v1beta1
kind: VirtualMachineImage
metadata:
  annotations:
    harvesterhci.io/storageClassName: my-storage-class
  creationTimestamp: "2025-04-08T07:13:46Z"
  finalizers:
  - wrangler.cattle.io/vm-image-controller
  generateName: image-
  generation: 15
  labels:
    harvesterhci.io/imageDisplayName: vm-import-gm-ubuntu-test-gm-ubuntu-test-default-disk-0.img
  name: image-n2vp4
  namespace: default
  resourceVersion: "854392"
  uid: efcea13b-b05d-4846-b0fd-020bae0e8cf2
spec:
  backend: backingimage
  checksum: ""
  displayName: vm-import-gm-ubuntu-test-gm-ubuntu-test-default-disk-0.img
  pvcName: ""
  pvcNamespace: ""
  retry: 0
  sourceType: download
  storageClassParameters:
    encrypted: "false"
    migratable: "true"
    numberOfReplicas: "1"
[...]

PR #8020, opened 2025-04-08
Use storage class name from annotation `harvesterhci.io/storageClassName`

votdev:
... in helper function `GetImageStorageClassName` and use the previous value `longhorn` as fallback.

**Problem:**
The helper function GetImageStorageClassName is hardcoding the prefix of the returned StorageClass name with `longhorn` which might cause confusion if you are using using a custom StorageClass.

**Solution:**
Make use of the `harvesterhci.io/storageClassName` annotation if available. Alternatively fall back to `longhorn` or anything else that needs to be discussed.

**Related Issue:**
https://github.com/harvester/harvester/issues/8019

**Test plan:**
Check `Reproduce` in https://github.com/harvester/harvester/issues/8019.

### DIFF
--- a/pkg/util/image.go
+++ b/pkg/util/image.go
@@ -73,7 +73,11 @@ func GetImageStorageClassName(image *harvesterv1.VirtualMachineImage) string {
 	if image.Spec.Backend == harvesterv1.VMIBackendCDI {
 		return image.Spec.TargetStorageClassName
 	}
-	return fmt.Sprintf("longhorn-%s", image.Name)
+	prefix := "longhorn"
+	if annotationStorageClassName, ok := image.Annotations[AnnotationStorageClassName]; ok {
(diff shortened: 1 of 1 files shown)


pkg/util/image.go
```diff
@@ -73,7 +73,11 @@ func GetImageStorageClassName(image *harvesterv1.VirtualMachineImage) string {
 	if image.Spec.Backend == harvesterv1.VMIBackendCDI {
 		return image.Spec.TargetStorageClassName
 	}
-	return fmt.Sprintf("longhorn-%s", image.Name)
+	prefix := "longhorn"
```
votdev, line 76: I think we should replace `longhorn` with something else since this is not correct anymore since we support 3rd party storage.

votdev:
Will be closed for the time being until the issue has been discussed.

Issue #8055, opened 2025-04-11
[CI] Support auto-assignment for PR reviewer and assignee


innobead:
### Task request

Use the same approach of https://github.com/longhorn/website/blob/master/.github/auto-assign-config.yaml or similar to assign assignee and reviewers to a PR. 

Add this action to https://github.com/rancher/gh-issue-mgr.

Yu-Jack:
Hi @innobead 

Who can I assign? It's supposed to have a default reviewer. Is it harvester/dev ?

innobead:
Make the action template work first. We will set it for each repo later.

Yu-Jack:
We ended up using `workflow-templates`.

![Image](https://github.com/user-attachments/assets/70c2ba45-a9d8-49f4-9fa2-15454b78b9e5)

Yu-Jack:
@innobead Since PR has been merged, should we need to test it by adding a reviewer or should we just close this?

innobead:
This is a general task. Just close and fix if needed.

PR #8060, opened 2025-04-14
chore: add auto-assign workflow

Yu-Jack:
Related Issue: https://github.com/harvester/harvester/issues/8055
Template Workflow: https://github.com/harvester/.github/pull/2

### DIFF
--- a/.github/workflows/auto-assign.yaml
+++ b/.github/workflows/auto-assign.yaml
@@ -1,0 +1,16 @@
+name: "[PR Management] Auto Assign"
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  auto-assign:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


innobead (changes requested):
Suggest adding this common workflow to .github repo instead, which was recently created. 

Then, we can just reuse the same workflow over repos.

Yu-Jack:
@innobead Do you mean this repo `rancher/gh-issue-mgr`?

innobead (dismissed):

Yu-Jack:
I've tested it in my org [test4](https://github.com/JackTestHook/test/pull/219).

.github/workflows/auto-assign.yaml
```diff
@@ -0,0 +1,14 @@
+name: Auto Assign
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review]
+
+permissions:
+  pull-requests: write
```
Yu-Jack, line 8: After testing, I found out that it needs this. Otherwise, it got this `Warning: Resource not accessible by integration`.

innobead (approved):

Issue #7983, opened 2025-04-02
[CI] Sync issue management between harvester and longhorn


Yu-Jack:
For better understanding and action items, this is the current sprint project management flow between Harvester and Longhorn.

## General flow for a newly created issue added into a sprint project

```mermaid
flowchart LR
    A[Newly Created Issue] --> B{ Who is the creator? }
    B -- Community Users --> C[Added to Community Review Sprint project ]
    B -- Team Members --> G{Is milestone set?}
    G -- Yes --> D[Added to Harvester Sprint project]
    G -- No --> H[Add Planning milestone]
    H --> D
    C --> E{Marked with milestone?}
    E -- Yes --> D
    E -- No --> F[Back to Team Review Required]
```

### Harvester and Longhorn Sprint Project

| Issue Status           | Harvester Behavior                                | Longhorn Behavior                        |
|-----------------------|--------------------------------------------------|--------------------------------------------|
| **New**               | Sprint field will be **cleared**                  | Sprint field will be **cleared**          |
| **Backlog**           | Sprint field will be **cleared**                  | X (Don't have this field)                 |
| **Icebox**            | Sprint field will be **cleared**                  | X (Don't have this field)                  |
| **Analysis & Design** | Sprint field will be **cleared** (Analysis/Design)| Sprint field will be **cleared**          |
| **Implement**         | Sprint field will be **cleared**                  | Sprint field will be **cleared**          |
| **Review**            | Sprint field will be **moved to the next sprint** | Sprint field will be **moved to the next sprint** |
| **Ready For Testing** | Sprint field will **remain unchanged**            | Sprint field will **remain unchanged**     |
| **Testing**           | Sprint field will **remain unchanged**            | Sprint field will **remain unchanged**     |
| **Close**             | Sprint field will **remain unchanged** (Closed)   | Sprint field will **remain unchanged**     |

### Harvester and Longhorn QA Project

| Issue Status             | Longhorn Behavior                                          | 
|--------------------------|------------------------------------------------------------|
| **New**                  | Sprint field will be **cleared**                           |
| **Analysis and Design**  | Sprint field will be **cleared**                           |
| **Implement**            | Sprint field will be **cleared**                           |
| **Review**               | Sprint field will be **moved to the next Sprint**          |
| **Testing**              | Sprint field will **remain unchanged**                     |
| **Closed**               | Sprint field will **remain unchanged**                     |

| Issue Status      | Harvester Behavior                                   |
[...]

PR #8070, opened 2025-04-15
chore: increase frequency temporarily

Yu-Jack:
https://github.com/harvester/harvester/issues/7983

- ~~Temporarily Increase frequency of staling issues, it'll be reverted after checking all issues.~~
- Add new stale workflow to stale `Planning` issue for reminding us to review it.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,8 +1,14 @@
 name: 'Close stale issues and PRs'
 on:
   schedule:
-    - cron: '30 1 * * *'
+    - cron: '30 * * * *'
   workflow_dispatch:
+    inputs:
+      ascending:
+        description: 'Sort issues and PRs in ascending order'
+        required: true
+        type: boolean
+        default: true
 
 permissions:
   issues: write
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -1,8 +1,14 @@
 name: 'Close stale issues and PRs'
 on:
   schedule:
-    - cron: '30 1 * * *'
+    - cron: '30 * * * *'
```
innobead, line 4: This seems too frequent. The original daily cleanup is good enough for us. The previous issue is more about the limited processing number of issues instead of frequency.

.github/workflows/stale-planning-issue.yml
```diff
@@ -0,0 +1,54 @@
+name: 'Close stale Planning issues and PRs'
```
innobead: Planning is intended including PM tasks. It's more like a backlog, probably we should rename it.

We don't need to do any cleanup, as it's intended. However, reviewing issues in the backlog should be a regular process. This would be tackled by a different approach.

innobead (changes requested):

.github/workflows/stale-planning-issue.yml
Yu-Jack: okay, I'll remove this.

.github/workflows/stale.yml
Yu-Jack: Currently, the operation is not enough to go through all issues, but increasing operation might hit rate limit. In order to prevent rate limit, we can increase the frequency to check all issues.

After we check all issues and close most unactivated issues, we could tune the frequency back.

.github/workflows/stale.yml
```diff
@@ -1,8 +1,14 @@
...
   workflow_dispatch:
+    inputs:
+      ascending:
+        description: 'Sort issues and PRs in ascending order'
+        required: true
+        type: boolean
+        default: true
```
innobead, line 11: This needs to be reverted later after cleaning up all outdated issues.

innobead (approved):

Issue #5074, opened 2024-01-31
[BUG] `support-bundle-expiration` can be invalid value `000003333` 


irishgordo:
**Describe the bug**
Support Bundle Expiration setting either modifying through kubectl or harvester gui (harvester/dashboard) can be invalid value

**To Reproduce**
1. Modify setting to set to 000003333 or some variant

**Expected behavior**
Value will normalize to "0" 

**Workaround**
Don't set the value to that invalid value

**Environment**
 - Harvester ISO Version: v1.3.0-rc1
 - Rancher Dashboard Version: `commit c3ae9bc14ad1a306025c0882eefde7baef3d046d` ( containing `commit 65366810efe83db05512194589771b8c3a0a0914` )
 - Underlying Infrastructure: qemu/kvm

**Additional context**
![Screenshot from 2024-01-31 12-08-13](https://github.com/harvester/harvester/assets/5370752/1b7e9f08-e523-4992-b12b-8da1a9c904eb)
![Screenshot from 2024-01-31 12-07-28](https://github.com/harvester/harvester/assets/5370752/e484c4a3-943b-4fd7-8869-5eac6e45a011)

irishgordo:
This is still present in v1.5.0-rc4

![Image](https://github.com/user-attachments/assets/1d6d4214-df45-492a-aaee-50a73808dc83)


cc: @a110605

houhoucoop:
Hi @irishgordo 
based on common UX expectations, I recommend normalizing `000003333` to `3333`.

Yu-Jack:
For backend testing, you can edit yaml directly. It should show some validator error like this:

```
# settings.harvesterhci.io "support-bundle-expiration" was not valid:
# * value: leading zero in input: 00300
```

irishgordo:
Thanks for the fix @houhoucoop 😄 👍 
This looks good on Harvester v1.6-9d264654-head
I'll go ahead and close this out

![Image](https://github.com/user-attachments/assets/c7dc5865-186d-42d0-8ed8-085ac27a9ed3)

PR #8102, opened 2025-04-18
fix: use strictAtoi to check string leading with zero

Yu-Jack:
#### Problem:
The `strconv.Atoi` treats `"0003"` as normal integer, then convert it to `3`

#### Solution:
However, we don't want string leading with zero. So, we add a new method to validate it. 

The affected settings are:

- SupportBundleTimeout
- SupportBundleExpiration
- SupportBundleNodeCollectionTimeout
- KubeConfigTTLSetting

#### Related Issue(s):
https://github.com/harvester/harvester/issues/5074

#### Test plan:

Use string leading with zero to replace above settings.

image

image

image

image

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -12,7 +12,6 @@ import (
 	"os"
 	"regexp"
 	"slices"
-	"strconv"
 	"strings"
 
 	// Although we don't use following drivers directly, we need to import them to register drivers.
@@ -1465,7 +1464,7 @@ func validateDefaultVMTerminationGracePeriodSecondsHelper(value string) error {
 		return nil
 	}
 
-	num, err := strconv.ParseInt(value, 10, 64)
+	num, err := webhookUtil.StrictAtoi(value)
 	if err != nil {
 		return err
 	}
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -39,6 +39,7 @@ import (
 
 	networkv1 "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester-network-controller/pkg/utils"
+
 	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/containerd"
 	nodectl "github.com/harvester/harvester/pkg/controller/master/node"
@@ -55,7 +56,7 @@ import (
 	vmUtil "github.com/harvester/harvester/pkg/util/virtualmachine"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
-	webhookutil "github.com/harvester/harvester/pkg/webhook/util"
+	webhookUtil "github.com/harvester/harvester/pkg/webhook/util"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


FrankYang0529 (dismissed):

pkg/webhook/resources/setting/validator.go
```diff
@@ -775,7 +775,7 @@ func validateSupportBundleNodeCollectionTimeoutHelper(value string) error {
 		return nil
 	}
 
-	i, err := strconv.Atoi(value)
+	i, err := webhookUtil.StrictAtoi(value)
```
brandboat, line 778: We already have negative number validation in `webhookUtil.StrictAtoi` which make L782~L784 redundant

pkg/webhook/resources/setting/validator.go
```diff
@@ -1533,7 +1533,7 @@ func validateKubeConfigTTLSettingHelper(value string) error {
 		return nil
 	}
 
-	num, err := strconv.Atoi(value)
+	num, err := webhookUtil.StrictAtoi(value)
```
brandboat, line 1536: Same thing here, L1541~L1543 become redundant.

pkg/webhook/resources/setting/validator.go
```diff
@@ -715,7 +715,7 @@ func validateSupportBundleTimeoutHelper(value string) error {
 		return nil
 	}
 
-	i, err := strconv.Atoi(value)
+	i, err := webhookUtil.StrictAtoi(value)
```
brandboat, line 718: ditto, L722~L724 become redundant.

pkg/webhook/resources/setting/validator.go
```diff
@@ -745,7 +745,7 @@ func validateSupportBundleExpirationHelper(value string) error {
...
```
brandboat, line 748: ditto, L752~L754 become redundant.

brandboat (commented):
Could you make the same change to `DefaultVMTerminationGracePeriodSeconds` as well?

pkg/webhook/resources/setting/validator.go
Yu-Jack: Nice catch!

pkg/webhook/resources/setting/validator.go
Yu-Jack: @brandboat I should remove negative number validation in `webhookUtil.StrictAtoi`. It's just a stronger version of `strconv.Atoi`.

brandboat:
> Could you make the same change to DefaultVMTerminationGracePeriodSeconds as well?

@Yu-Jack , just one nit here, could you please refactor the `DefaultVMTerminationGracePeriodSeconds` as well?

pkg/webhook/util/string.go
```diff
@@ -0,0 +1,24 @@
...
+			return 0, fmt.Errorf("leading zero in input: %s", s)
+		}
+	}
+
+	i, err := strconv.Atoi(s)
+	if err != nil {
+		return 0, err
+	}
+	return i, nil
+}
```
Yu-Jack: The `strconv.Atoi` can parse `001` and `-001`.  In this stronger version, we shouldn't allow it.

brandboat: Does that mean we need to forbid the usage of `+001` ?

Yu-Jack: yes

Yu-Jack: I added a test file, and you can check it to see if it make sense.

brandboat (approved):
LGTM, thanks!

FrankYang0529 (approved):

Issue #8109, opened 2025-04-21
[BUG] VM with NAD network blocks namespace deletion


FrankYang0529:
### Describe the Bug

When removing a namespace, the controller automatically remove VMs in that namespace. If one of VM has net-attach-def network, the controller may show error message like:

```
time="2025-04-17T19:01:07Z" level=error msg="error syncing 'thisistest/svc-02': handler VMController.CleanupPVCAndSnapshot: admission webhook \"mutator.harvesterhci.io\" denied the request: network-attachment-definitions.k8s.cni.cncf.io \"thisistest\" not found, requeuing"
```

### To Reproduce

1. Create a namespace "thisistest".
2. Create a VMImage and vlan1 net-attach-def in "thisistest" namespace.
3. Create a VM with valn1 network.
4. Deelete the namespace "thisistest".

### Expected Behavior

There is no error message in the controller.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.4.0
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): KVM

albinsun:
Hi @pratikjagrut ,
Please help to verify this issue, thank you.

pratikjagrut:
All test cases pass as provided in the test plan.

# Environment
Harvester: 1.6.0-rc3

# Test Scenario
Steps:
<details>
<summary>Create a namespace "thisistest"</summary>

<img width="1252" height="597" alt="Image" src="https://github.com/user-attachments/assets/f4d8f8c8-56c8-428a-9585-9e61947558aa" />

</details>


<details>
<summary>Create a Virtual Machine Network in "thisistest" namespace</summary>

<img width="1252" height="597" alt="Image" src="https://github.com/user-attachments/assets/9cb832bd-f0b2-4e63-8e6b-84a7133acf43" />
</details>

<details>
<summary>Create a VM with the nic-1(virtual machine n/w) network.</summary>

<img width="1383" height="799" alt="Image" src="https://github.com/user-attachments/assets/e058b220-26e7-4005-b85b-883a4c156679" />
</details>

<details>
<summary>Delete the namespace "thisistest"</summary>

<img width="1383" height="799" alt="Image" src="https://github.com/user-attachments/assets/7dd6fcfc-a665-4f87-a51b-80bc1d6c77f5" />

[...]

PR #8110, opened 2025-04-21
feat: avoid to do mutation when deletion ts is not null

FrankYang0529:
#### Problem:
We didn't check whether the new VM has deletionTimestamp before doing update request mutation. The mutation may fail if related resource is gone.

#### Solution:
The delete request mutation can be handled with `admissionregv1.Delete`, so we can skip update request if VM has deletionTimestamp.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8109
https://github.com/harvester/harvester/pull/7743

#### Test plan:
1. Create a VM with NAD.
2. Stop the VM.
3. Remove the NAD.
4. Remove the VM successfully.

### DIFF
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -86,6 +86,10 @@ func (m *vmMutator) Update(_ *types.Request, oldObj runtime.Object, newObj runti
 	newVM := newObj.(*kubevirtv1.VirtualMachine)
 	oldVM := oldObj.(*kubevirtv1.VirtualMachine)
 
+	if newVM == nil || newVM.DeletionTimestamp != nil {
+		return nil, nil
+	}
+
 	logrus.Debugf("update VM %s/%s", newVM.Namespace, newVM.Name)
 
 	var patchOps types.PatchOps


w13915984028:
@FrankYang0529 

We need to review the `mutation` webhook, it should to be less active when the VM is `OFF`, or when it has a deletiontimestamp.

For issue like https://github.com/harvester/harvester/issues/6961, the user-friend solution is to edit the VM and delete the dangling network definition;  it is better than deleting the VM, which is the last step. 

This PR https://github.com/harvester/harvester/pull/8110 can't fully replace https://github.com/harvester/harvester/pull/7743, as it assumes that the only solution is to delete the original VM, but it seems unfair to the user.

There is another PR https://github.com/harvester/harvester/pull/7619, which also needs to cooperate with this case. cc @votdev

FrankYang0529:
> For issue like #6961, the user-friend solution is to edit the VM and delete the dangling network definition; it is better than deleting the VM, which is the last step.

If we allow to do this, we may need to update webhook PR to ignore stopped VM.

https://github.com/harvester/network-controller-harvester/pull/131

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -400,6 +404,9 @@ func (m *vmMutator) getNodeSelectorRequirementFromNetwork(defaultNamespace strin
 
 	nad, err := m.nad.Get(namespace, name)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
```
w13915984028: For issue https://github.com/harvester/harvester/issues/8109, changes on L407~409 are un-related.

could we only add L89~L92 for this PR? thanks.

w13915984028 (dismissed):
LGTM, thanks.

pkg/webhook/resources/virtualmachine/mutator.go
FrankYang0529: Updated it. Thanks.

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

votdev (approved):
LGTM

w13915984028:
@FrankYang0529 Suggest to backport this PR to v1.4.3 and v1.5.1, WDYT? thanks.

Issue #8049, opened 2025-04-11
[ENHANCEMENT] Reduce the downtime for VMs with PCI devices during Harvester upgrade


brandboat:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

During the Harvester 1.3.2 -> 1.4.0 upgrade, VMs with PCI Passthrough (GPU) must be powered off and remain off during the entire process, causing significant downtime for critical workloads.

### Describe the solution you'd like

VMs with PCI passthrough should only be powered off while the specific node is updated, allowing us to balance workloads between two nodes.

ihcsim:
To recap, it seems like the current implementation at 83b89e820f739a9333182cf22ca0c4ab41301226 triggers a reconciliation loop every time an annotation update is made per VM. This happens as the original upgrade reconciliation tries to complete. We explore putting the annotations either at the VM or node level, each has its pros and cons, and require us to retain and coordinate restart states throughout the whole upgrade process.

To simplify the logic and have a more reliable persistence mechanism, we decided to store the list of non-migratable VMs in a configmap, with an owner reference to the upgrade object.

@brandboat let me know if i missed anything from our discussion.

In pre-drain job, store the non-migratable VM information to a configmap.
* The configmap includes label to record the upgrade uuid to help the job in step 2 determine which upgrade it is associated with
* Set configmap owner reference to the upgrade object, once upgrade is deleted, the configmap is gone.
 
Here is the configmap example:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: upgrade-non-migratable-vms-xdkj #generated name
  namespace: harvester-system
  labels:
    upgrade-uuid: 123e4567-e89b-12d3-a456-426614174000
  ownerReferences: #the upgrade object
    ...
data:
  node-1: default/vm1,default/vm2
  node-2: default/vm3
  node-3: default/vm4,default/vm5,default/vm6
```

After the node reboots, we submit the VM restart job. We can do this in [here](https://github.com/harvester/harvester/blob/ccabfab00f0c97b898a27408fe730d5b55bd5702/pkg/controller/master/upgrade/node_controller.go#L84-L109).

This job is fire-and-forget, and won't block the upgrade process even if it failed.

We do not want to wait for vm status and delay the upgrade even more.

The restart is done on a best-effort basis. E.g., if vm failed to be restarted, we won't retry forever. If configmap is lost, no restart will happen. We can emit some `event`s to record these key events (so that they will be captured in SB):
* configmap is created
[...]

PR #8138, opened 2025-04-25
Restore non-migratible VMs back to running state after node upgrade

brandboat:
#### Problem:

In the current upgrade flow, if there is any non-migratable & running VMs exist, Harvester will ask the user to shut them down. After the upgrade completes, users need to manually start those VMs again. If users want to minimize the downtime of non-migratable VMs, they have to monitor the upgrade process for each node and manually restart the VMs as soon as their node finishes upgrading, which is quite inconvenient.

#### Solution:

This PR aims to provide a way to automatically bring non-migratable VMs back online once their node has been upgraded. We already have an option for this (`restoreVM`), as documented here: https://docs.harvesterhci.io/v1.6/advanced/index/#upgrade-config. However, currently this feature only works in a single-node environment. This PR extends the feature to support multi-node environments.

The idea is quite simple:
If restoreVM set to `true`
1. Add annotation `harvesterhci.io/restore-vm-{node_name}={vmnames_separated_by_comma}` to upgrade CR in pre-drain job (multi-node env) or single-node-upgrade job (single node env).
2. Non-migratable VMs will be shut down automatically when their node is about to be upgraded (see https://github.com/harvester/harvester/blob/master/package/upgrade/upgrade_node.sh#L292). And for migratable VMs, they should remain in running state during the whole upgrade process (if there is no trouble).
3. Once the node upgrade is "succeeded", `upgrade_controller` start the VMs in the annotation `harvesterhci.io/restore-vm-{node_name}`. Repeat the same process after each subsequent node upgrade. And also add another annotation `harvesterhci.io/restore-vm-finished-{node_name}` to make which node restore VM process is finished.

[Note]: paused non-migratable VM **_won't be started again_** after node upgrade finished.

#### Related Issue(s):

#8049 

#### Test plan:

**Test single node upgrade**
There are 2 scenarios here:
1. v1.5.0 upgrade to the version that includes this patch.
    - Prepare 3 VMs vm-stopped (in stopped state), vm-paused (in paused state), vm-running (in running state).
    - Enable `restoreVM` option
    - After the upgrade finished, only vm-running should be in running state, others should be in stopped state.
2. master -> master
    - Prepare 2 VMs, vm-stopped, vm-paused, vm-running.
    - Enable `restoreVM` option
    - After the upgrade finished, only vm-running should be in running state, others should be in stopped state.

**Test multi-node upgrade**
- Prepare a multi-node harvester cluster, create 5VMs
  vm-running, vm-stopped, vm0 (node-selector: node0), vm1 (node-selector: node1), vm2 (node-selector: node2)
- Enable the `restoreVM` option in upgradeConfig setting
- After upgrade finished, the VM state should be like picture in below.
image

What are non-migratable VMs?
- VM contains node selector
- VM contains PCIe devices
- VM contains node affinities while we can't migrate to those nodes in node-affinities settings 
- VM contains CD-ROM or container disks
https://github.com/harvester/harvester/blob/master/pkg/util/virtualmachineinstance/virtualmachineinstance.go#L17

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -17,6 +18,7 @@ import (
 	"github.com/rancher/wrangler/v3/pkg/condition"
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/batch/v1"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	"github.com/rancher/wrangler/v3/pkg/kv"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -77,15 +79,8 @@ const (
 	skipVersionCheckAnnotation          = "harvesterhci.io/skip-version-check"
 	defaultImagePreloadConcurrency      = 1
 
-	preUpgradeVMsAnnotation = "harvesterhci.io/pre-upgrade-vms"
-
 	vmReady condition.Cond = "Ready"
 )
-
-type PreUpgradeRunningVM struct {
-	Namespace string `json:"namespace"`
-	Name      string `json:"name"`
-}
 
 // upgradeHandler Creates Plan CRDs to trigger upgrades
 type upgradeHandler struct {
@@ -119,8 +114,6 @@ type upgradeHandler struct {
 	lhSettingClient ctllhv1.SettingClient
 	lhSettingCache  ctllhv1.SettingCache
 
-	kubeVirtCache kubevirtctrl.KubeVirtCache
-
 	vmRestClient rest.Interface
 }
 
@@ -143,10 +136,6 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 
 		toUpdate := upgrade.DeepCopy()
 		initStatus(toUpdate)
-
-		if err := h.storeVMState(toUpdate); err != nil {
-			return nil, err
-		}
 
 		if !upgrade.Spec.LogEnabled {
 			logrus.Info("Upgrade observability is administratively disabled")
@@ -220,6 +209,10 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 		return upgrade, nil
 	}
 
+	if err := h.restoreVM(upgrade); err != nil {
+		return nil, err
+	}
+
 	// clean upgrade repo VMs and images if a upgrade succeeds.
 	if harvesterv1.UpgradeCompleted.IsTrue(upgrade) {
 		// try to clean up images before purging the repo VM
@@ -243,10 +236,6 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			}
 		}
 
-		// moved the restoreVMState call after the start repo vm as we need to make sure kubevirt is up and running
-		// thus we can bring up the vm to the state before upgrade
-		h.restoreVMState(upgrade)
-
 		if err := h.cleanupImages(upgrade, repo); err != nil {
 			logrus.Warningf("Failed to cleanup images: %s", err.Error())
 			toUpdate := upgrade.DeepCopy()
@@ -264,10 +253,6 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 
 	// upgrade failed
 	if harvesterv1.UpgradeCompleted.IsFalse(upgrade) {
-		// try to restore vm to the state before upgrade
-		// we didn't wait for KubeVirt to reach the Deployed phase because the upgrade failed,
-		// as a result, some services might not be ready and may never fully start.
-		h.restoreVMState(upgrade)
 		// clean upgrade repo VMs.
 		return nil, h.cleanup(upgrade, harvesterv1.UpgradeCompleted.IsTrue(upgrade))
 	}
@@ -731,7 +716,7 @@ func (h *upgradeHandler) prepareNodesForUpgrade(upgrade *harvesterv1.Upgrade, re
 	logrus.WithFields(logrus.Fields{
 		"namespace":      upgrade.Namespace,
 		"name":           upgrade.Name,
-		"upgrade_config": upgradeConfig,
+		"upgrade_config": fmt.Sprintf("%+v", upgradeConfig),
 	}).Info("start preparing nodes for upgrade")
 
 	nodes, err := h.nodeCache.List(labels.Everything())
@@ -869,120 +854,114 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
 	return nil
 }
 
-// storeVMState stores VM state only in single-node environments.
-// this method stores information of VMs with Ready=true (i.e. running state) in the upgrade annotation
-func (h *upgradeHandler) storeVMState(upgrade *harvesterv1.Upgrade) error {
+// restoreVM attempts to restore all running non-migratable VMs after the node upgrade succeeds.
+// During the node upgrade, in a multi-node environment, annotation harvesterhci.io/restore-vm-{node_uid}={vm_names_separated_by_comma}
+// will be added to upgrade CR for each node pre-drain job.
+// In a single-node environment, this will occur during the single-node-upgrade job.
+// Once the node upgrade is complete, VMs will be started based on the annotation.
+// Note that this function waits until KubeVirt is ready before restoring the VMs.
+// If KubeVirt is not ready, it will retry every 5 seconds by enqueuing the upgrade.
+func (h *upgradeHandler) restoreVM(upgrade *harvesterv1.Upgrade) error {
 	logFields := logrus.Fields{
 		"namespace": upgrade.Namespace,
 		"name":      upgrade.Name,
 	}
-	isSingleNodeRestoreVM, err := h.isSingleNodeRestoreVM(upgrade)
+
+	restoreVM, err := util.IsRestoreVM()
 	if err != nil {
-		return err
-	}
-	if !isSingleNodeRestoreVM {
-		logrus.WithFields(logFields).Info("Skip store VM state")
+		logrus.WithFields(logFields).WithError(err).Errorf("Failed to get setting UpgradeConfig, enqueue upgrade for next check")
+		h.upgradeController.EnqueueAfter(upgrade.Namespace, upgrade.Name, time.Second*10)
 		return nil
 	}
 
-	if upgrade.Annotations != nil && upgrade.Annotations[preUpgradeVMsAnnotation] != "" {
-		logrus.WithFields(logFields).Infof("Skip store VM state since %s is already set", preUpgradeVMsAnnotation)
+	if !restoreVM {
 		return nil
 	}
 
-	vms, err := h.vmCache.List(corev1.NamespaceAll, labels.Everything())
+	err = h.checkKubeVirtHealth()
 	if err != nil {
-		logrus.WithFields(logFields).WithError(err).Error("Failed to list all VMs")
-		return err
+		logrus.WithFields(logFields).WithError(err).Warn("Kubevirt is not ready, enqueue upgrade for next check")
+		h.upgradeController.EnqueueAfter(upgrade.Namespace, upgrade.Name, time.Second*10)
+		return nil
 	}
 
-	runningVMs := make([]*PreUpgradeRunningVM, 0)
-	for _, vm := range vms {
-		if vmReady.IsTrue(vm) {
-			vm := &PreUpgradeRunningVM{
-				Namespace: vm.Namespace,
-				Name:      vm.Name,
+	for nodeName, nodeStatus := range upgrade.Status.NodeStatuses {
+		node, err := h.nodeCache.Get(nodeName)
+		if err != nil {
+			logrus.WithFields(logFields).WithError(err).Errorf("Failed to get node %s", nodeName)
+			return err
+		}
+
+		nodeUID := string(node.UID)
+		if nodeStatus.State == StateSucceeded &&
+			upgrade.Annotations[util.AnnotationRestoreVMFinishedPrefix+nodeUID] != "true" &&
+			upgrade.Annotations[util.AnnotationRestoreVMPrefix+nodeUID] != "" {
+			if err = h.restoreVMForNode(upgrade, node); err != nil {
+				return err
 			}
-			runningVMs = append(runningVMs, vm)
+		} else {
+			logrus.WithFields(logFields).Infof("VM restore for node %s is already done", nodeName)
 		}
 	}
-	runningVMsJSON, err := json.Marshal(runningVMs)
-	if err != nil {
-		logrus.WithFields(logFields).WithError(err).Error("Failed to marshal PreUpgradeRunningVM list to json")
-		return err
-	}
-
-	if upgrade.Annotations == nil {
-		upgrade.Annotations = make(map[string]string)
-	}
-	upgrade.Annotations[preUpgradeVMsAnnotation] = string(runningVMsJSON)
 	return nil
 }
 
-// restoreVMState attempts to restore all VMs to their pre-upgrade state.
-// This function is designed for single-node environments. There is no guarantee
-// that all VMs will successfully return to their previous state. If any VM fails to start,
-// the error is logged, but the process continues without interruption.
-func (h *upgradeHandler) restoreVMState(upgrade *harvesterv1.Upgrade) {
+// restoreVMForNode attempts to restore non-migratable VMs for specific node.
+// There is no guarantee that all VMs will successfully return to their previous state.
+// If any VM fails to start, the process continues without interruption.
+func (h *upgradeHandler) restoreVMForNode(upgrade *harvesterv1.Upgrade, node *corev1.Node) error {
 	logFields := logrus.Fields{
 		"namespace": upgrade.Namespace,
 		"name":      upgrade.Name,
 	}
-	// ignore the error here as the method should be no-op if there is
-	// something wrong in isSingleNodeRestoreVM, and we want to proceed
-	// to the next step: cleanup jobs, stop repo vm... after upgrade success/failed.
-	isSingleNodeRestoreVM, _ := h.isSingleNodeRestoreVM(upgrade)
-	if !isSingleNodeRestoreVM {
-		logrus.WithFields(logFields).Info("Skip restore VM")
-		return
-	}
 
-	if upgrade.Annotations == nil || upgrade.Annotations[preUpgradeVMsAnnotation] == "" {
-		logrus.WithFields(logFields).Infof("Skip restore VM state since %s is not set", preUpgradeVMsAnnotation)
-		return
-	}
+	nodeName := node.Name
+	nodeUID := string(node.UID)
 
-	preUpgradeRunningVMs := []PreUpgradeRunningVM{}
-	err := json.Unmarshal([]byte(upgrade.Annotations[preUpgradeVMsAnnotation]), &preUpgradeRunningVMs)
-	if err != nil {
-		logrus.WithFields(logFields).WithError(err).Error("Failed to unmarshal json to PreUpgradeRunningVM list")
-		return
-	}
+	vmsStr, _ := upgrade.Annotations[util.AnnotationRestoreVMPrefix+nodeUID]
+	for _, vmName := range strings.Split(vmsStr, ",") {
+		vmName := strings.TrimSpace(vmName)
+		if vmName == "" {
+			continue
+		}
 
-	for _, vmInfo := range preUpgradeRunningVMs {
-		vm, err := h.vmCache.Get(vmInfo.Namespace, vmInfo.Name)
+		namespace, name := kv.RSplit(vmName, "/")
+		vm, err := h.vmCache.Get(namespace, name)
 		if err != nil {
-			logrus.WithFields(logFields).WithError(err).Errorf("Failed to get VM %s/%s from cache", vmInfo.Namespace, vmInfo.Name)
+			logrus.WithFields(logFields).WithError(err).Errorf("Failed to get VM %s/%s", harvesterSystemNamespace, vmName)
 			continue
 		}
-		// the vm is already in running state
-		if vmReady.IsTrue(vm) {
-			continue
-		}
-		if err = h.startVM(context.Background(), vm); err != nil {
-			logrus.WithFields(logFields).WithError(err).Errorf("Failed to start vm %s/%s after upgrade", vmInfo.Namespace, vmInfo.Name)
-		}
-	}
-}
 
-func (h *upgradeHandler) isSingleNodeRestoreVM(upgrade *harvesterv1.Upgrade) (bool, error) {
-	singleNode, err := h.isSingleNodeCluster()
-	if err != nil {
-		logrus.WithFields(logrus.Fields{
-			"namespace": upgrade.Namespace,
-			"name":      upgrade.Name,
-		}).WithError(err).Error("Failed to check if cluster is single node")
-		return false, err
+		if vmReady.IsTrue(vm) {
+			logrus.WithFields(logFields).Infof("VM %s/%s is already running, skip restoring", vm.Namespace, vm.Name)
+			continue
+		}
+
+		logrus.WithFields(logFields).Infof("Restoring VM %s/%s", harvesterSystemNamespace, vmName)
+		startVMErr := retry.OnError(
+			retry.DefaultBackoff,
+			func(err error) bool {
+				return err != nil
+			},
+			func() error {
+				return h.startVM(context.Background(), vm)
+			},
+		)
+		// Don't return error if VM fails to start to avoid endless retry
+		if startVMErr != nil {
+			logrus.WithFields(logFields).WithError(startVMErr).Errorf("Failed to start VM %s/%s after retries", vm.Namespace, vm.Name)
+		}
 	}
-	upgradeConfig, err := settings.DecodeConfig[settings.UpgradeConfig](settings.UpgradeConfigSet.Get())
-	if err != nil {
-		logrus.WithFields(logrus.Fields{
-			"namespace": upgrade.Namespace,
-			"name":      upgrade.Name,
-		}).WithError(err).Error("Failed to get UpgradeConfig")
-		return false, err
+
+	// Add the annotation after restoring VMs finished
+	upgradeCopy := upgrade.DeepCopy()
+	upgradeCopy.Annotations[util.AnnotationRestoreVMFinishedPrefix+nodeUID] = "true"
+	if _, err := h.upgradeClient.Update(upgradeCopy); err != nil {
+		logrus.WithFields(logFields).WithError(err).Errorf("Failed to add annotation %s", util.AnnotationRestoreVMFinishedPrefix+nodeUID)
+		return err
 	}
-	return singleNode != "" && upgradeConfig.RestoreVM, nil
+	logrus.WithFields(logFields).Infof("Restored VMs for node %s", nodeName)
+	return nil
 }
 
 func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMachine) error {
@@ -998,5 +977,12 @@ func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMach
 		SubResource("start").
 		Body(body).
 		Do(ctx)
+	return res.Error()
+}
+
+func (h *upgradeHandler) checkKubeVirtHealth() error {
+	res := h.vmRestClient.Get().
+		Resource("healthz").
+		Do(context.Background())
 	return res.Error()
 }
--- a/pkg/upgradehelper/vmlivemigratedetector/detector.go
+++ b/pkg/upgradehelper/vmlivemigratedetector/detector.go
@@ -2,20 +2,33 @@ package vmlivemigratedetector
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"strings"
 
+	"github.com/rancher/wrangler/v3/pkg/condition"
 	"github.com/rancher/wrangler/v3/pkg/kv"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	types "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/client-go/util/retry"
+
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	"kubevirt.io/client-go/kubecli"
 
+	harvNodeController "github.com/harvester/harvester/pkg/controller/master/node"
+	harvclient "github.com/harvester/harvester/pkg/generated/clientset/versioned"
+	"github.com/harvester/harvester/pkg/util"
 	"github.com/harvester/harvester/pkg/util/virtualmachineinstance"
+)
+
+const (
+	vmPaused condition.Cond = "Paused"
 )
 
 type DetectorOptions struct {
@@ -23,16 +36,27 @@ type DetectorOptions struct {
 	KubeContext    string
 	Shutdown       bool
 	NodeName       string
+	RestoreVM      bool
+	Upgrade        string
 }
 
 type VMLiveMigrateDetector struct {
 	kubeConfig  string
 	kubeContext string
 
-	nodeName string
-	shutdown bool
+	nodeName  string
+	shutdown  bool
+	restoreVM bool
+	upgrade   string
 
 	virtClient kubecli.KubevirtClient
+	harvClient harvclient.Interface
+}
+
+type PatchStringValue struct {
+	Op    string      `json:"op"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value"`
 }
 
 func NewVMLiveMigrateDetector(options DetectorOptions) *VMLiveMigrateDetector {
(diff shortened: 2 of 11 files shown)


pkg/util/setting.go
```diff
@@ -61,3 +61,11 @@ func GetAdditionalGuestMemoryOverheadRatio(settingCache ctlharvesterv1.SettingCa
 	value = agmorc.Value()
 	return &value, nil
 }
+
+func IsRestoreVM() (bool, error) {
+	upgradeConfig, err := settings.DecodeConfig[settings.UpgradeConfig](settings.UpgradeConfigSet.Get())
```
w13915984028, line 66: The `DecodeConfig` function may return `nil, err`, all the caller should check the return first, then run it as a func; in fail case, the panic happens

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -130,6 +132,23 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+		logrus.WithFields(logrus.Fields{
+			"namespace": upgrade.Namespace,
+			"name":      upgrade.Name,
+		}).WithError(err).Error("Failed to get setting UpgradeConfig")
+		return nil, err
+	}
+	// restore VM state if the node upgrade is completed
+	if isRestoreVM && upgrade.Status.NodeStatuses != nil && len(upgrade.Status.NodeStatuses) > 0 {
+		for nodeName, nodeStatus := range upgrade.Status.NodeStatuses {
+			if nodeStatus.State == StateSucceeded || nodeStatus.State == StateFailed {
```
w13915984028: If node state is failed, start the VM may not be a good idea. The before-change code tries to restore when upgrade is failed, but it only works on single-node cluster.

For multi-node cluster, what's the consideration to do it also on the failed node?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -845,32 +858,25 @@ func (h *upgradeHandler) storeVMState(upgrade *harvesterv1.Upgrade) error {
...
-	vms, err := h.vmCache.List(corev1.NamespaceAll, labels.Everything())
+	vmis, err := h.vmiCache.List(corev1.NamespaceAll, labels.Everything())
 	if err != nil {
 		logrus.WithFields(logFields).WithError(err).Error("Failed to list all VMs")
 		return err
 	}
 
 	runningVMs := make([]*PreUpgradeRunningVM, 0)
-	for _, vm := range vms {
+	for _, vm := range vmis {
```
w13915984028: just save all running VMs ?

w13915984028 (commented):
Thanks for the enhancement, a few questions:

A main complex case is:

The running VMs info are all saved as pre-upgrade-vm, and when a node is upgraded, no matter successful/failed, start all none-running VMs originally hosted on this node.

I did not find any code to save the VMs info explicitly when the VMs are turned off by upgrade code. If a running VM is turned off by user in-between the upgrade due do various of reasons, the VM may still be powered on accidentally by the new code. (please correct me if I am wrong).  Could this be improved?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -130,6 +132,23 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 
 	repo := NewUpgradeRepo(h.ctx, upgrade, h)
 
+	isRestoreVM, err := util.IsRestoreVM()
+	if err != nil {
+		logrus.WithFields(logrus.Fields{
+			"namespace": upgrade.Namespace,
+			"name":      upgrade.Name,
+		}).WithError(err).Error("Failed to get setting UpgradeConfig")
+		return nil, err
```
ihcsim: Will this cause upgrade to abort or stuck in a loop if `err` doesn't goes away?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -140,8 +159,10 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 		toUpdate := upgrade.DeepCopy()
 		initStatus(toUpdate)
 
-		if err := h.storeVMState(toUpdate); err != nil {
-			return nil, err
+		if isRestoreVM {
+			if err := h.storeVMState(toUpdate); err != nil {
```
ihcsim: Instead of relying on annotations (which has a size limit), would it be possible/simpler to find and restart all the non-migratable previously shutdown VMs on this node? We already have the general logic in `virtualmachineinstance.GetAllNonLiveMigratableVMINames()`, as seen used by the [`vmlivemigratedetector`](https://github.com/harvester/harvester/blob/a0db5f850d09c4ee3a0b4101c520adb1219d66df/pkg/upgradehelper/vmlivemigratedetector/detector.go#L103-L106). The detector can also add a `restart-required` label to the VM before shutting it down. That way, we know what we touched. Together with the `kubevirt.io/nodeName` label, this will make the VMs more searchable. WDYT?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -130,6 +132,23 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+			"namespace": upgrade.Namespace,
+			"name":      upgrade.Name,
+		}).WithError(err).Error("Failed to get setting UpgradeConfig")
+		return nil, err
+	}
+	// restore VM state if the node upgrade is completed
+	if isRestoreVM && upgrade.Status.NodeStatuses != nil && len(upgrade.Status.NodeStatuses) > 0 {
+		for nodeName, nodeStatus := range upgrade.Status.NodeStatuses {
+			if nodeStatus.State == StateSucceeded || nodeStatus.State == StateFailed {
+				h.restoreVMState(upgrade, nodeName, repo)
```
ihcsim: Do we need to wait for kubevirt to be ready?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -130,6 +132,23 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 
 	repo := NewUpgradeRepo(h.ctx, upgrade, h)
 
+	isRestoreVM, err := util.IsRestoreVM()
```
ihcsim: Do we need to use the `restoreVM` option? Feels like the state restoration should be the default behaviour i.e., we shutdown user's VMs during upgrade, then we turn them back on when we are done.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -889,22 +895,13 @@ func (h *upgradeHandler) storeVMState(upgrade *harvesterv1.Upgrade) error {
 }
 
 // restoreVMState attempts to restore all VMs to their pre-upgrade state.
-// This function is designed for single-node environments. There is no guarantee
-// that all VMs will successfully return to their previous state. If any VM fails to start,
-// the error is logged, but the process continues without interruption.
-func (h *upgradeHandler) restoreVMState(upgrade *harvesterv1.Upgrade) {
+// There is no guarantee that all VMs will successfully return to their previous state.
+// If any VM fails to start, the process continues without interruption.
+func (h *upgradeHandler) restoreVMState(upgrade *harvesterv1.Upgrade, nodeName string, repo *Repo) {
```
ihcsim: Do we need to remove the annotation after the restoration?

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: In the beginning I want to align the behavior between single-node & multi-node. I'll change this to not restart VM if node upgrade failed. Thanks.

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Yes, it will stuck forever. Perhaps we can ignore the error here as we already have a validator for upgradeconfig to make sure the json format is valid. https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/setting/validator.go#L1620-L1630

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: You are right, but I'm uncertain what's the best way to check if KubeVirt is ready. Do you think check the `.status.phase` field is `Deployed` in kubevirt CR is enough?

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: One scenario is that user might not want us to automatically shut down non-migratable VMs—they might prefer to handle the shutdown themselves, or control when to stop and start those VMs.

But I might be overthinking. I don’t have a strong opinion against making it the default behavior, but I’d like to hear if others have any thoughts on this. c.c. @w13915984028, @starbops, @innobead

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Sure, I'll do that in the next commit.

pkg/controller/master/upgrade/upgrade_controller.go
w13915984028: Yes, the user scenarios are versatile, some VMs are dropable and some are strictly not.

We need to have flexible options for user to choose. 

Also another issue https://github.com/harvester/harvester/issues/8212 for info.

pkg/util/setting.go
brandboat: Sure, I'll update the if stmt in next line to

```go
	if err != nil || upgradeConfig == nil {
		return false, err
	}
```

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Thanks for the suggestion! This is a better approach, I'll address this in the next version.

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Not sure if I missed anything-would appreciate any hints you can share! 😃  
BTW @ihcsim left some suggestion in https://github.com/harvester/harvester/pull/8138#discussion_r2067418652, perhaps this is something you want to point out?

w13915984028: It was explained at above; and was similar to @ihcsim 's suggestion; Harvester needs to book exactly the VM operated by it and restore upon them.   The failure case https://github.com/harvester/harvester/issues/8212 is also thoughtful.

>A main complex case is:

>The running VMs info are all saved as pre-upgrade-vm, and when a node is upgraded, no matter successful/failed, start all none-running VMs originally hosted on this node.

>I did not find any code to save the VMs info explicitly when the VMs are turned off by upgrade code. If a running VM is turned off by user in-between the upgrade due do various of reasons, the VM may still be powered on accidentally by the new code. (please correct me if I am wrong). Could this be improved?
```

**Comment by hoo29** [comment]
How does this interact with the `harvesterhci.io/maintain-mode-strategy` annotations? If we set `ShutdownAndRestartAfterDisable` do we get similar behaviour?

**Comment by brandboat** [comment]
> How does this interact with the harvesterhci.io/maintain-mode-strategy annotations? If we set ShutdownAndRestartAfterDisable do we get similar behaviour?

Hi @hoo29, yes. If you set restoreVM to true, you will get similar behavior. Non-migratable VMs will be automatically shut down when the node is being upgraded and will be automatically started once the node upgrade is complete.

File: pkg/upgradehelper/vmlivemigratedetector/detector.go [hunk 9]
```diff
@@ -113,6 +123,34 @@ func (d *VMLiveMigrateDetector) Run(ctx context.Context) error {
...
+			if d.restoreVM {
+				logrus.Infof("Adding restoreVM label to VM %s", namespacedName)
+
+				payload, _ := json.Marshal([]util.PatchStringValue{
+					{
+						Op:    "add",
+						Path:  "/metadata/labels/" + strings.ReplaceAll(util.LabelRestoreVMAfterUpgrade, "/", "~1"), // JSON Patch format requires '/' to be replaced with '~1'
+						Value: d.nodeName,
+					},
+				})
```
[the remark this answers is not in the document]
brandboat: The json is a const, add err check here is redundant.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -838,120 +832,125 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
-		return false, err
-	}
-	upgradeConfig, err := settings.DecodeConfig[settings.UpgradeConfig](settings.UpgradeConfigSet.Get())
-	if err != nil {
-		logrus.WithFields(logrus.Fields{
-			"namespace": upgrade.Namespace,
-			"name":      upgrade.Name,
-		}).WithError(err).Error("Failed to get UpgradeConfig")
-		return false, err
+		if _, exist := vm.Labels[util.LabelRestoreVMAfterUpgrade]; exist {
```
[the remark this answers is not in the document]
brandboat: the vm is defined in each loop

brandboat:
gentle ping @w13915984028, @ihcsim. I’ve addressed the comments you mentioned. Could you please take a look when you have a chance? Thank you!

w13915984028 (commented):
Thanks for the new implementation, it covers almost all cases. I have two additional concerns:

(1) The `retry` is used in several places; when a controller goes into normal running stage, the `retry` should be used as less as possible, the `reconciller & enqueue` achieves the same target;  `retry` blocks the current go thread, it might slow the whole upgrade process

(2) The restore label is added to `vm`, the code tries to ensure label is added, processed, removed;  when it fails to remove the label in some cases, it might affect next round upgrade

instead, could we label it like `upgrade.harvesterhci.io/node-id/name-hash:[vm id slices]` (normally, one node can't have tons of VMs and cause the content exceed label/annotation limits) upon the upgrade CR, the benefits are:
```
all changes happen on upgrade CR;  
it is kept for better debugging;  
even if when error happens, remove label from upgrade CR will stop the reconciller quickly;
it won't affect next round upgrade
```

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -869,120 +858,106 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
 
-	runningVMs := make([]*PreUpgradeRunningVM, 0)
-	for _, vm := range vms {
-		if vmReady.IsTrue(vm) {
-			vm := &PreUpgradeRunningVM{
-				Namespace: vm.Namespace,
-				Name:      vm.Name,
+	for nodeName, nodeStatus := range upgrade.Status.NodeStatuses {
+		if nodeStatus.State == StateSucceeded &&
+			upgrade.Annotations[util.AnnotationRestoreVMFinishedPrefix+nodeName] != "true" &&
```
w13915984028: Did you check if the combined key length may exceed limitation?  use node UID?

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -119,3 +149,101 @@ func (d *VMLiveMigrateDetector) Run(ctx context.Context) error {
...
+	if err != nil {
+		return metav1.ListOptions{}, err
+	}
+	return metav1.ListOptions{
+		LabelSelector: labels.NewSelector().Add(*nodeReq).Add(*notUpgradeReq).String(),
+	}, nil
+}
+
+func getNonLiveMigratableVMNames(vmis []*kubevirtv1.VirtualMachineInstance, nodes []*v1.Node) ([]string, error) {
+	// All VMs are non-migratable if there is only one node
```
w13915984028, line 206: in case 1 control node + 1 witness node, all VMs are also not migratable; is this case covered by later call 
`return virtualmachineinstance.GetAllNonLiveMigratableVMINames(vmis, nodes)`

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -119,3 +149,101 @@ func (d *VMLiveMigrateDetector) Run(ctx context.Context) error {
...
+	return virtualmachineinstance.GetAllNonLiveMigratableVMINames(vmis, nodes)
+}
+
+func getRestoreVMNames(vmis []*kubevirtv1.VirtualMachineInstance, nonMigratableVMNames []string) []string {
+	restoreVMs := make([]string, 0, 0)
+
+	// filter out paused VMs and upgrade repo VM from the non-migratable VMs
+	excludeVMs := make(map[string]struct{})
+	for _, vmi := range vmis {
+		if vmPaused.IsTrue(vmi) || vmi.Labels["harvesterhci.io/upgrade"] != "" {
```
w13915984028, line 223: check `vmi.Labels == nil` first, though it rarely happens

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -119,3 +149,101 @@ func (d *VMLiveMigrateDetector) Run(ctx context.Context) error {
...
+			// JSON Patch format requires '/' to be replaced with '~1'
+			Path:  "/metadata/annotations/" + strings.ReplaceAll(util.AnnotationRestoreVMPrefix+d.nodeName, "/", "~1"),
+			Value: strings.Join(restoreVMNames, ","),
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
w13915984028, line 251: if `d.upgrade` was updated elsewhere, then `retry` here does not help; you need to fetch a new `upgrade` then update

w13915984028 (dismissed):
LGTM, thanks.

A few minor suggestions.

pkg/upgradehelper/vmlivemigratedetector/detector.go
brandboat: In the retry block, it uses PATCH instead of UPDATE. Is fetching a new object really necessary?

w13915984028: Need a double check, I did not test such before.

pkg/upgradehelper/vmlivemigratedetector/detector.go
brandboat: Yeah I've thought about it before. But in golang read from nil map won't cause NPE. Quite interesting.

https://go.dev/play/p/DwrwBdQ6lQI

w13915984028: yeah, read is ok, write is panic, quite brain-burning :)

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -869,120 +854,111 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
 	}
 
-	if upgrade.Annotations != nil && upgrade.Annotations[preUpgradeVMsAnnotation] != "" {
-		logrus.WithFields(logFields).Infof("Skip store VM state since %s is already set", preUpgradeVMsAnnotation)
+	if !restoreVM {
 		return nil
 	}
 
-	vms, err := h.vmCache.List(corev1.NamespaceAll, labels.Everything())
+	err = h.checkKubeVirtHealth()
```
brandboat, line 881: Checking the AllComponentsReady condition in KubeVirt object is not sufficient, as VM start failures can still occur with the error 'the server is currently unable to handle the request.' occasionally. A more reliable approach is to call the `/healthz` API.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -869,120 +854,111 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
+		namespace, name := kv.RSplit(vmName, "/")
+		vm, err := h.vmCache.Get(namespace, name)
+		if err != nil {
+			logrus.WithFields(logFields).WithError(err).Errorf("Failed to get VM %s/%s", harvesterSystemNamespace, vmName)
 			continue
 		}
-		if err = h.startVM(context.Background(), vm); err != nil {
-			logrus.WithFields(logFields).WithError(err).Errorf("Failed to start vm %s/%s after upgrade", vmInfo.Namespace, vmInfo.Name)
+
+		if !vmReady.IsTrue(vm) {
```
w13915984028: an alternative to reduce indent:
```
if vmReady.IsTrue(vm) {
  log
  continue
}
// start VM
...
```

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -119,3 +154,115 @@ func (d *VMLiveMigrateDetector) Run(ctx context.Context) error {
...
+	for _, name := range nonMigratableVMNames {
+		if _, exist := excludeVMs[name]; !exist {
+			restoreVMs = append(restoreVMs, name)
+		}
+	}
+	return restoreVMs
+}
+
+func (d *VMLiveMigrateDetector) patchUpgradeCR(ctx context.Context, restoreVMNames []string, nodeUID string) error {
+	if len(restoreVMNames) == 0 {
```
w13915984028: In this case, the patch might be still needed, if the ugprade CR contains a related annotation which was patched before.

is here a way like ` if !DeepEqual  { client.Update } ` ?

pkg/controller/master/upgrade/upgrade_controller.go
w13915984028: the `restoreVM` is out of the upgrade controller `OnChange` main state machine,  when error returns, it shortcuts the state machine, causes the latter can't move forward; if a better place can't be found, then log error, enqueue and continue following code will be the solution. WDYT ?

Just noticed the `upgrade.Labels[harvesterLatestUpgradeLabel] == "true"` is checked way after; but for `restoreVM`, it strictly depends on this label, to avoid confused by legacy upgrade objects

https://github.com/harvester/harvester/blob/202c56239f9b8440ee30d3f0474a1f6935aac9b0/pkg/controller/master/upgrade/upgrade_controller.go#L219
```
	// only run further operations for latest upgrade
	if upgrade.Labels == nil || upgrade.Labels[harvesterLatestUpgradeLabel] != "true" {
		return upgrade, nil
	}
```

w13915984028 (dismissed):
LGTM, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -869,120 +854,111 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
+				},
+			)
+			// Don't return error if VM fails to start to avoid endless retry
+			if startVMErr != nil {
+				logrus.WithFields(logFields).WithError(startVMErr).Errorf("Failed to start VM %s/%s after retries", vm.Namespace, vm.Name)
+			}
+		} else {
+			logrus.WithFields(logFields).Infof("VM %s/%s is already running, skip restoring", vm.Namespace, vm.Name)
 		}
 	}
```
ibrokethecloud: would it be better to just check if VM is halted, and in case it is halted to just update the run strategy and move on?

With a large number of non migratable VM's i worry this may slow down the upgrade process.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -869,120 +854,111 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
-	runningVMsJSON, err := json.Marshal(runningVMs)
-	if err != nil {
-		logrus.WithFields(logFields).WithError(err).Error("Failed to marshal PreUpgradeRunningVM list to json")
-		return err
-	}
 
-	if upgrade.Annotations == nil {
-		upgrade.Annotations = make(map[string]string)
+		nodeUID := string(node.UID)
+		if nodeStatus.State == StateSucceeded &&
```
ibrokethecloud, line 896: should we start the VMs back if the upgrade has failed?

ibrokethecloud (commented):
we should drop this as it is no longer needed https://github.com/brandboat/harvester/blob/27eab4d1157ddd712a6e9ebcacbeb4e270923806/pkg/webhook/resources/setting/validator.go#L1638-L1644

ibrokethecloud (commented):
we should also document how the user can leverage this capability. The default `restoreVM` option is set to false

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: We have this discussion before, https://github.com/harvester/harvester/pull/8138#discussion_r2066895485.
I have no strong opinion about this...

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: > would it be better to just check if VM is halted, and in case it is halted to just update the run strategy and move on?

Yeah, I actually tried that before. But later I realized the start vm REST API already does pretty much the same thing—plus it handles more edge cases. So I’d prefer sticking with that.

The StartVMRequestHandler https://github.com/kubevirt/kubevirt/blob/v1.4.0/pkg/virt-api/rest/subresource.go#L460 basically:

- Gets the VM and its VMI.
- Checks if it’s already running or needs manual recovery.
- Reads any extra options from the request (like paused start).
- Then, depending on the RunStrategy, sends the right patch to start the VM.

From what I’ve seen, the slowest parts could be:
- Fetching the VM and VMI from the API server.
- Sending the patch to update the VM—that's probably the most expensive step.

To me, performance here isn’t a big deal. I’d rather use the REST API since it already handles more cases and is less error-prone. WDYT?

brandboat:
> we should drop this as it is no longer needed https://github.com/brandboat/harvester/blob/27eab4d1157ddd712a6e9ebcacbeb4e270923806/pkg/webhook/resources/setting/validator.go#L1638-L1644

Thanks! I'll remove the code in next commit.

> we should also document how the user can leverage this capability. The default restoreVM option is set to false

Yeah, I'll file another doc pr after this one merged.

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: > the restoreVM is out of the upgrade controller OnChange main state machine, when error returns, it shortcuts the state machine, causes the latter can't move forward; if a better place can't be found, then log error, enqueue and continue following code will be the solution. WDYT ?

I assume you mean add `h.upgradeController.EnqueueAfter(upgrade.Namespace, upgrade.Name, time.Second*10)` if util.IsRestoreVM raise error, right?

> Just noticed the upgrade.Labels[harvesterLatestUpgradeLabel] == "true" is checked way after; but for restoreVM, it strictly depends on this label, to avoid confused by legacy upgrade objects

Good catch! Thanks!

@w13915984028, I've addressed all comments in the latest commit, hopefully I didn't get something wrong.

w13915984028 (approved):
LGTM, thanks.

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -119,3 +154,124 @@ func (d *VMLiveMigrateDetector) Run(ctx context.Context) error {
...
+	return virtualmachineinstance.GetAllNonLiveMigratableVMINames(vmis, nodes)
+}
+
+func getRestoreVMNames(vmis []*kubevirtv1.VirtualMachineInstance, nonMigratableVMNames []string) []string {
+	restoreVMs := make([]string, 0, 0)
+
+	// filter out paused VMs and upgrade repo VM from the non-migratable VMs
+	excludeVMs := make(map[string]struct{})
+	for _, vmi := range vmis {
+		if vmPaused.IsTrue(vmi) || vmi.Labels["harvesterhci.io/upgrade"] != "" {
```
ihcsim, line 223: I thought the list option will ensure the VMIs don't have the `harvesterhci.io/upgrade` label.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -869,120 +854,114 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
+	for _, vmName := range strings.Split(vmsStr, ",") {
+		vmName := strings.TrimSpace(vmName)
+		if vmName == "" {
+			continue
+		}
 
-	for _, vmInfo := range preUpgradeRunningVMs {
-		vm, err := h.vmCache.Get(vmInfo.Namespace, vmInfo.Name)
+		namespace, name := kv.RSplit(vmName, "/")
+		vm, err := h.vmCache.Get(namespace, name)
```
ihcsim, line 929: Instead of `Get`ting one VM at a time based on the upgrade object's annotations, why not just label the VMs in the detector and then select them here in one `List` call? The label can be something like `harvesterhci.io/post-upgrade-restart: "pending"`.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -869,120 +854,114 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
+			func(err error) bool {
+				return err != nil
+			},
+			func() error {
+				return h.startVM(context.Background(), vm)
+			},
+		)
+		// Don't return error if VM fails to start to avoid endless retry
+		if startVMErr != nil {
+			logrus.WithFields(logFields).WithError(startVMErr).Errorf("Failed to start VM %s/%s after retries", vm.Namespace, vm.Name)
```
ihcsim, line 952: If we use labels on the VMs instead of annotation, we can label the failed VM here with something like `harvesterhci.io/post-upgrade-restart: "failed"`.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -869,120 +854,114 @@ func (h *upgradeHandler) setAutoCleanupSystemGeneratedSnapshotValue(value string
...
-	upgradeConfig, err := settings.DecodeConfig[settings.UpgradeConfig](settings.UpgradeConfigSet.Get())
-	if err != nil {
-		logrus.WithFields(logrus.Fields{
-			"namespace": upgrade.Namespace,
-			"name":      upgrade.Name,
-		}).WithError(err).Error("Failed to get UpgradeConfig")
-		return false, err
+	// Add the annotation after restoring VMs finished
+	upgradeCopy := upgrade.DeepCopy()
+	upgradeCopy.Annotations[util.AnnotationRestoreVMFinishedPrefix+nodeUID] = "true"
```
ihcsim, line 958: Similarly, we can label the VMs with `harvesterhci.io/post-upgrade-restart: "completed"`.

pkg/controller/master/upgrade/upgrade_controller.go
ihcsim: I thought the [current upgrade behaviour](https://github.com/harvester/harvester/blob/83b89e820f739a9333182cf22ca0c4ab41301226/package/upgrade/upgrade_node.sh#L291-L292) is to auto-shutdown all the non-migratable VMs. Also, isn't this setting about restoring, not shutting down, the VMs? Would a user ever want a shutdown VM (for whatever reasons) to not be restored after an upgrade?

ihcsim (commented):
I think it's more efficient to use labels on the VMs/VMIs, than use annotations on the upgrade object because:

* One `List` API call on the controller side. Personally, I don't mind the detector making more `Update` API calls because it's an one-off client process
* Use just one label to reflect the restart status (`pending`, `completed`, `failed`) instead of 2 annotations to represent the same task

WDYT?

ihcsim:
Ah.. I just saw https://github.com/harvester/harvester/pull/8138#pullrequestreview-2842505844. My concern is still the N `Get` VM calls we have to make on the controller side. IMO, it's better to make them on the detector client side. With labels, one can do `k get [vm|vmi] -l harvesterhci.io/post-upgrade-restart=[pending|complete|failed]` post upgrade. On large cluster, we also don't to worry about the annotation size limit on the upgrade object.

Thoughts?

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Are you suggesting that users should be able to exclude certain VMs from being restored after an upgrade? That feels a bit unusual to me… but it’s doable. For example, we could introduce an additional option, such as `excludeVMs`, allowing users to specify a list of non-migratable VMs they don’t want to restore after the node upgrade completes.

However, this might add some complexity to the user experience. One challenge is: how would users know which VMs are non-migratable? We might need to add a dropdown menu in the Upgrade UI that lists all non-migratable VMs, so users can select which ones to exclude.

brandboat:
> My concern is still the N Get VM calls we have to make on the controller side

So originally, I thought it wouldn’t be a big performance issue since we’re using a cache to fetch VMs. Even though there are multiple Get calls, they shouldn’t all hit the API server directly. But maybe I’m missing something.

> On large cluster, we also don't to worry about the annotation size limit on the upgrade object.

I did try an approach before where we add annotations to the VMs during upgrade. But @w13915984028 pointed out a potential issue — if the annotation isn’t cleaned up properly after the upgrade (say the upgrade fails or something else goes wrong), it might mess up the next upgrade’s restoreVM behavior.
- One idea I had is to handle this in the upgrade validator. Like, if restoreVM is enabled, we check that no VMs still have the annotation. If any do, we return an error and ask the user to remove it.
- On top of that, we could also add a VM mutator that automatically removes the annotation whenever a VM is updated — but only if there’s no upgrade currently in progress.

ihcsim:
I think whether we are annotating/labeling the VMs or the upgrade object, there is always a chance for it to fail. It's hard to speculate which approach is more reliable in a distributed environment. We recently had an upgrade issue where the hook annotations were missing from the upgrade secret plans for who knows why. If we want a stronger guarantee, we need a different persistent mechanism.

If possible, I really don't want to make things more complicated with validator and mutator (not a fan of those things) changes.

The annotation change is introducing an additional single write for _every_ VM to the same upgrade object, which is already being regularly updated with different status `conditions`. Once the annotating is done (assuming it didn't fail), the controller has to re-do the whole reconciliation again using a new version of the upgrade object, but containing only an annotation change, with no _actual_ upgrade state change. This feels like a little "side quest" for the controller. With VM labeling, the controller can fire off a goroutine to patch the VM and continue with its original intention to reconcile the upgrade status.

Re: leftover annotations (or labels), I thought the only real effect they have is whether the VMs need to be restarted? It shouldn't complicate an upgrade restart. In my suggestion, when upgrade starts, the detector just label every VM with `harvesterhci.io/post-upgrade-restart=pending` regardless of what the previous value was. I think that makes sense (?) because a new run of the upgrade means every VM should be re-assessed for live migration eligibility (and restart condition). Then as upgrade progresses, we evolve the value of this label to `failed`, `completed` etc., but no need to remove them at the end.

Re: using cache to fetch the VMs, for a critical operation like upgrade, I'd think we want to fetch the latest version of the VMs from etcd, not from the cache just in case the cache doesn't see the latest status of the VMs.

pkg/controller/master/upgrade/upgrade_controller.go
ihcsim: Can we make this async (via goroutine) and best-effort, especially with the retries? An VM restart error shouldn't block the entire upgrade process. We do have to care about performance to the extent that it doesn't significantly slow down upgrade. There are cases where upgrade took N hours (~day) to complete.

pkg/controller/master/upgrade/upgrade_controller.go
ihcsim: What's gonna happen if we tried to restart the VMs on the failed node? Since they can't be live migrated, then restart requests will just fail, right?

brandboat:
superseded by https://github.com/harvester/harvester/pull/8526

PR #8141, opened 2025-04-28
chore: find tune some features in issue management flow [skip ci]

Yu-Jack:
Related Issue #7983 

What this PR did:

1. Use strict search.
2. Remove `not-require/test-plan` label.
3. Use `require/ui` instead of `require-ui/xxxx`
4. If we really need to add `wontfix` label, we can add it back in the future. But, we'll need to adjust the permission. So, I prefer not to add it.
  image

### DIFF
--- a/.github/workflows/issue-management-close-issue.yaml
+++ b/.github/workflows/issue-management-close-issue.yaml
@@ -31,20 +31,12 @@ jobs:
     - name: Find and Close Backport Issue
       run: |
         # Search for backport issues with matching title using major.minor version
-        BACKPORT_ISSUE=$(gh issue list --search '"[backport v${{ env.MAJOR_MINOR_VERSION }}] ${{ github.event.issue.title }}"' --json number,title --jq '.[0]')
-        BACKPORT_ISSUE_NUMBER=$(echo "$BACKPORT_ISSUE" | jq -r '.number')
-        BACKPORT_ISSUE_TITLE=$(echo "$BACKPORT_ISSUE" | jq -r '.title')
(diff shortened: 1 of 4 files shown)


.github/workflows/issue-management-close-issue.yaml
```diff
@@ -31,12 +31,20 @@ jobs:
...
-          gh issue close $BACKPORT_ISSUE --reason "not planned"
-          gh issue edit $BACKPORT_ISSUE --add-label "wontfix"
+        BACKPORT_ISSUE=$(gh issue list --search '"[backport v${{ env.MAJOR_MINOR_VERSION }}] ${{ github.event.issue.title }}"' --json number,title --jq '.[0]')
+        BACKPORT_ISSUE_NUMBER=$(echo "$BACKPORT_ISSUE" | jq -r '.number')
+        BACKPORT_ISSUE_TITLE=$(echo "$BACKPORT_ISSUE" | jq -r '.title')
+        if [ -n "$BACKPORT_ISSUE_NUMBER" ]; then
+          CURRENT_TITLE="${{ github.event.issue.title }}"
+          
+          # Prevent closing the original issue
+          if [ "$BACKPORT_ISSUE_TITLE" = "$CURRENT_TITLE" ]; then
```
innobead: Why does this happen? since you already use `[backport] ...` to search issue.

Yu-Jack: Good catch. With strict search, we don't this guard condition. Thanks.

innobead (approved):

Issue #8041, opened 2025-04-10
[Enhancement] Handle upgrade scenario for #7650


rrajendran17:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

This task is related to https://github.com/harvester/harvester/issues/7650

During Harvester boot, the system attempts to add all VLAN IDs to the management bond/bridge. With Mellanox ConnectX-6 NICs (which support 512 hardware-offloaded VLANs and have hardware VLAN offloading enabled by default), this causes boot log warnings and prevents management VLAN 2301 from being added to the hardware offloaded VLAN list.

https://github.com/harvester/harvester-installer/pull/1026 takes care of adding only user configured vlan to mgmt-br and mgmt-bo interface

when upgrading to v1.6.0 from older versions

- This task is created for handling the upgrade path where upgrade to any newer version with the fix from https://github.com/    
  harvester/harvester-installer/pull/1026  should patch only the vid configured by user for the mgmt-br(during installation)
  in wicked scripts of setup_bond.sh and setup_bridge.sh and 90_custom.yaml file.

### Describe the solution you'd like

This involves changing the files /etc/wicked/scripts/setup_bond.sh and /etc/wicked/scripts/setup_bridge.sh
and /oem/90_custom.sh during upgrade path.

albinsun:
Hi @Yu-Jack,
Please help to verify this issue.
As this relates to upgrade, should use `v1.6.0-rc5` or `v1.6.0-head` to avoid #8773.

thank you.

Yu-Jack:
🟢 Passed with upgrading from v1.5.0 to v1.6.0-rc5

```sh
node:/home/rancher # bridge vlan show
port              vlan-id
mgmt-br           1 PVID Egress Untagged
mgmt-bo           1 PVID Egress Untagged

node:/home/rancher # kubectl get upgrades -n harvester-system hvst-upgrade-l9lph -o yaml
apiVersion: harvesterhci.io/v1beta1
kind: Upgrade
metadata:
  creationTimestamp: "2025-08-06T05:15:54Z"
  finalizers:
  - wrangler.cattle.io/harvester-upgrade-controller
  generateName: hvst-upgrade-
  generation: 15
  labels:
    harvesterhci.io/latestUpgrade: "true"
    harvesterhci.io/upgradeState: UpgradingNodes
  name: hvst-upgrade-l9lph
  namespace: harvester-system
  resourceVersion: "49006"
[...]

Issue #7650, opened 2025-02-18
[ENHANCEMENT] Modify default behavior of adding all VLANs to the management bridge/bond


rrajendran17:
**Is your feature request related to a problem? Please describe.**
During Harvester boot, the system attempts to add all VLAN IDs to the management bond/bridge. With Mellanox ConnectX-6 NICs (which support 512 hardware-offloaded VLANs and have hardware VLAN offloading enabled by default), this causes boot log warnings and prevents management VLAN 2301 from being added to the hardware offloaded VLAN list.

**Describe the solution you'd like**
Add only configured vlans to the mgmt-br and mgmt-bo

**Describe alternatives you've considered**
a.During installation add only the vlan selected by user to mgmt-br and mgmt-bo interface.
b.Remove adding vlans 2-4094 for mgmt-br and mgmt-bo
c.During Network Attachment Definition creation, add only vlans to the mgmt bond interface that are part of mgmt cluster.

**Additional context**
https://github.com/harvester/harvester/issues/5678

RegisWeduLabs:
Hi,  I think I'm having exactly this issue - tough I'm getting a bit more complication...  It seems like all the traffic from the untagged network (or Vlan1) broadcast and multicasts ends up going in all vlans.  So is it safe to do something like:

bridge vlan del dev mgmt-br vid 2-4094
bridge vlan del dev mgmt-bo vid 2-4094

# Add back only your required VLANs
bridge vlan add dev mgmt-br vid 1 pvid untagged
bridge vlan add dev mgmt-br vid 2
bridge vlan add dev mgmt-br vid 3
bridge vlan add dev mgmt-br vid 4
....

# Do the same for mgmt-bo
bridge vlan add dev mgmt-br vid 1 pvid untagged
bridge vlan add dev mgmt-br vid 2
bridge vlan add dev mgmt-br vid 3
bridge vlan add dev mgmt-br vid 4
...

Also, when you add a second bond in the Harvester Interface, even if you don't assign/create new vlans (vm networks) on the 2nd bond, it still lists all active vlans, and the bond is also "member" of mgmt-bo, which seems very odd to me...

So running bridge vlan show, I get

mgmt-bo/br: 1 PVID Egress Untagged
2-4094...

new-bo-created-in-harvester: 1 PVID Egress Untagged
...actual vlans (vm networks) created with mgmt interface

On top of that, I end up with my new bridge (bond) created from the ui as part as mgmt-br like so:

[...]

PR #8153, opened 2025-04-28
patch mgmt vlan id to bridge vlan during upgrade

rrajendran17:
#### Problem:
This task is related to https://github.com/harvester/harvester/issues/7650

During Harvester boot, the system attempts to add all VLAN IDs to the management bond/bridge. With Mellanox ConnectX-6 NICs (which support 512 hardware-offloaded VLANs and have hardware VLAN offloading enabled by default), this causes boot log warnings and prevents management VLAN 2301 from being added to the hardware offloaded VLAN list.

https://github.com/harvester/harvester-installer/pull/1026 takes care of adding only user configured vlan to mgmt-br and mgmt-bo interface
when upgrading to v1.6.0 from older versions
This task is created for handling the upgrade path where upgrade to any newer version with the fix from 
https://github.com/harvester/harvester-installer/pull/1026 should patch only the vid configured by user for the mgmt-br(during installation) and mgmt-bo
in wicked scripts of setup_bond.sh and setup_bridge.sh and 90_custom.yaml file.

#### Solution:
This involves replacing the vlan range 2-4094 with vlan id configured during installation in the files /etc/wicked/scripts/setup_bond.sh and /etc/wicked/scripts/setup_bridge.sh
and /oem/90_custom.yaml with during upgrade path.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8041

#### Test plan:
1.Upgrade from v1.5.x to latest/v1.6.0
2.Check if wicked scripts and /oem/90_custom.yaml shows "bridge vlan ...." where vid is vlan configured for mgmt br during harvester installation
3.bridge vlan show - shows only the vlans configured for mgmt cluster along with the vlan id configured during installation
mgmt-br - will be part of vlan configured during installation
mgmt-bo - will be part of vlan configured during installation and other user configured vlans under mgmt cluster

Before upgrade:
mgmt-br and mgmt-bo will be part of 2-4094 vlans in "bridge vlan show" output

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -508,6 +508,64 @@ calculateCPUReservedInMilliCPU() {
   echo $reserved
 }
 
+replace_with_mgmtvlan() {
+  if [ -z "$UPGRADE_PREVIOUS_VERSION" ]; then
+    detect_upgrade
+  fi
+
+  if [[ ! "$UPGRADE_PREVIOUS_VERSION" =~ ^v1\.5\.[0-9]$ ]]; then
+    echo "version: $UPGRADE_PREVIOUS_VERSION does not require patching mgmt vlan"
+    return
+  fi
+
+  #files to update with mgmt vlan
+  local SETUPBOND_FILE="${HOST_DIR}/etc/wicked/scripts/setup_bond.sh"
+  local SETUPBRIDGE_FILE="${HOST_DIR}/etc/wicked/scripts/setup_bridge.sh"
+  local CUSTOM90_FILE="${HOST_DIR}/oem/90_custom.yaml"
+  local CONFIG_FILE="${HOST_DIR}/oem/harvester.config"
+
+  # Check if the harvester config file exists
+  if [[ ! -f "$CONFIG_FILE" ]]; then
+    echo "Error: Config file '$CONFIG_FILE' not found."
+    return
+  fi
+
+  # Search for the mgmt vlanid installed in the config file
+  vlan_id=$(yq '.install.managementinterface.vlanid' $CONFIG_FILE)
+
+  if [[ -z "$vlan_id" || "$vlan_id" -eq 0 ]]; then
+    echo "VLAN ID: $vlan_id Not found, assign vlan_id=1"
+    vlan_id=1
+  fi
+
+  # Check if the file exists
+  if [[ -f "$SETUPBOND_FILE" ]]; then
+    # Replace the range with the VLAN ID
+    sed -i "s/bridge vlan add vid 2-4094/bridge vlan add vid $vlan_id/" "$SETUPBOND_FILE"
+    echo "Updated $SETUPBOND_FILE with VLAN ID $vlan_id"
+  else
+    echo "File not found: $SETUPBOND_FILE"
+  fi
+
+  # Check if the file exists
+  if [[ -f "$SETUPBRIDGE_FILE" ]]; then
+    # Replace the range with the VLAN ID
+    sed -i "s/bridge vlan add vid 2-4094/bridge vlan add vid $vlan_id/" "$SETUPBRIDGE_FILE"
+    echo "Updated $SETUPBRIDGE_FILE with VLAN ID $vlan_id"
+  else
+    echo "File not found: $SETUPBRIDGE_FILE"
+  fi
+
+  # Check if the file exists
+  if [[ -f "$CUSTOM90_FILE" ]]; then
+    # Replace the range with the VLAN ID
+    sed -i "s/bridge vlan add vid 2-4094/bridge vlan add vid $vlan_id/" "$CUSTOM90_FILE"
+    echo "Updated $CUSTOM90_FILE with VLAN ID $vlan_id"
+  else
+    echo "File not found: $CUSTOM90_FILE"
+  fi
+}
+
 upgrade_os() {
   # The trap will be only effective from this point to the end of the execution
   trap clean_up_tmp_files EXIT
@@ -644,6 +702,9 @@ command_post_drain() {
 
   convert_nodenetwork_to_vlanconfig
 
+  #replace the range vlans with mgmt vlan in wicked scripts and 90_custom.yaml file before node reboot
+  replace_with_mgmtvlan
+
   upgrade_os
 }
 
@@ -682,6 +743,9 @@ command_single_node_upgrade() {
   clean_rke2_archives
 
   convert_nodenetwork_to_vlanconfig
+
+  #replace the range vlans with mgmt vlan in wicked scripts and 90_custom.yaml file before node reboot
+  replace_with_mgmtvlan
 
   # Upgrade OS
   upgrade_os


package/upgrade/upgrade_node.sh
```diff
@@ -508,6 +508,55 @@ calculateCPUReservedInMilliCPU() {
...
 
+replace_with_mgmtvlan() {
+  #files to update
+  SETUPBOND_FILE="${HOST_DIR}/etc/wicked/scripts/setup_bond.sh"
+  SETUPBRIDGE_FILE="${HOST_DIR}/etc/wicked/scripts/setup_bridge.sh"
+  CUSTOM90_FILE="${HOST_DIR}/oem/90_custom.yaml"
+  CONFIG_FILE="${HOST_DIR}/oem/harvester.config"
+
+  # Check if the harvester config file exists
+  if [[ ! -f "$CONFIG_FILE" ]]; then
```
w13915984028, line 528: return 1 will cause the ugprade to fail, and will always fail when the pod re-run

log an error/warning, then continue or return

package/upgrade/upgrade_node.sh
```diff
@@ -508,6 +508,55 @@ calculateCPUReservedInMilliCPU() {
   echo $reserved
 }
 
+replace_with_mgmtvlan() {
+  #files to update
+  SETUPBOND_FILE="${HOST_DIR}/etc/wicked/scripts/setup_bond.sh"
```
w13915984028: define vars as local to avoid pollute gloabl vars and vice versa

package/upgrade/upgrade_node.sh
```diff
@@ -508,6 +508,55 @@ calculateCPUReservedInMilliCPU() {
   echo $reserved
 }
 
+replace_with_mgmtvlan() {
```
w13915984028, line 511: This function is only needed when upgrade before version is v1.5.0 or v1.5.1 or v1.5.2, please add a check and return early if not required to run

w13915984028 (commented):
A few suggestions, thanks.

And, please post your local test output upon a mulit-node cluster on e.g. PR description, before and upgrade, the related files content and the upgrade script log, which is helpfu.

package/upgrade/upgrade_node.sh
rrajendran17: @w13915984028 I thought we do only incremental upgrades, so the previous version must be v1.5.x for upgrading to v1.6.0 and we cannot directly upgrade from 1.4.x or older version directly to v1.6.0.
Please correct me if I am wrong or missing something

w13915984028: Yeah, right, but not fully.

Look at the future, when a `v1.6` branch is forked, ideally, this block of code in `master` needs to be removed; and when `v1.6.0` is released, the code in `v1.6` branch (will become `v1.6.1`) needs also be removed.  When this option is chosen, a new ticket is needed to reminder you later.

If the code is controlled by before-upgrade version, it makes sure it only runs on specific version to avoid side-effects on un-related versions.

rrajendran17: sure got it, to avoid running this code on v1.6.0 or versions after that.

w13915984028 (dismissed):
LGTM, thanks for the fix.

mingshuoqiu (dismissed):
LGTM. Thanks

package/upgrade/upgrade_node.sh
```diff
@@ -508,6 +508,64 @@ calculateCPUReservedInMilliCPU() {
...
+  fi
+
+  if [[ ! "$UPGRADE_PREVIOUS_VERSION" =~ ^v1\.5\.[0-9]$ ]]; then
+    echo "version: $UPGRADE_PREVIOUS_VERSION does not require patching mgmt vlan"
+    return
+  fi
+
+  #files to update with mgmt vlan
+  local SETUPBOND_FILE="${HOST_DIR}/etc/wicked/scripts/setup_bond.sh"
+  local SETUPBRIDGE_FILE="${HOST_DIR}/etc/wicked/scripts/setup_bridge.sh"
```
starbops, line 523: Why do we need to update these two files? I thought they're immutable and can only be altered in `/oem/90_custom.yaml`. Maybe updating the `90_custom.yaml` is enough?

package/upgrade/upgrade_node.sh
```diff
@@ -508,6 +508,64 @@ calculateCPUReservedInMilliCPU() {
...
+  local CONFIG_FILE="${HOST_DIR}/oem/harvester.config"
+
+  # Check if the harvester config file exists
+  if [[ ! -f "$CONFIG_FILE" ]]; then
+    echo "Error: Config file '$CONFIG_FILE' not found."
+    return
+  fi
+
+  # Search for the mgmt vlanid installed in the config file
+  vlan_id=$(grep -E "vlanid[: ]" "$CONFIG_FILE" | awk -F':' '{print $2}' | tr -d ' ')
```
starbops: Any reason not using `yq`? e.g. `yq '.install.managementinterface.vlanid' /oem/harvester.config`

package/upgrade/upgrade_node.sh
```diff
@@ -613,6 +671,9 @@ EOF
   umount $tmp_rootfs_mount
   rm -rf $tmp_rootfs_squashfs
 
+  #replace the range vlans with mgmt vlan in wicked scripts and 90_custom.yaml file before node reboot
+  replace_with_mgmtvlan
```
starbops: Does it need to be inside the `upgrade_os` function? Can we move the function call outside one level to `command_post_drain` and `command_single_node_upgrade`?

package/upgrade/upgrade_node.sh
rrajendran17: Yes updating /oem/90_custom.yaml is enough, so that after reboot, it takes the contents of /oem/90_custom.yaml.But I think it will be good to update the wicked scripts as well so that their contents are in sync with what users expect and the 90_custom.yaml.

package/upgrade/upgrade_node.sh
rrajendran17: no particular reason, changed to yq, much simpler, thanks

package/upgrade/upgrade_node.sh
rrajendran17: Fixed it.

package/upgrade/upgrade_node.sh
rrajendran17: Yes updating /oem/90_custom.yaml is enough, so that after reboot, it takes the contents of /oem/90_custom.yaml.But I think it will be good to update the wicked scripts as well so that their contents are in sync with what users expect and the 90_custom.yaml.

rrajendran17:
<del>Also While revisiting the code changes, I observed I missed handling a case where in older version for example v1.5.0, users has configured vlans/NAD under mgmt cluster and after upgrade to v1.6.0 /oem/90_custom.yaml replaces "bridge vlan vid <> ...." with mgmt vlan id configured during installation.But the expectation is vlans configured by user part of mgmt cluster should also be part of "bridge vlan ...".</del>

<del>So I have included add_mgmtbridge_vlans() function which will create a systemd service to get the user configured NADs under mgmt cluster and add it to "bridge vlan ..." after node reboot.</del>

The new network controller running with changes "https://github.com/harvester/network-controller-harvester/pull/166/files" should take care of adding the vid to bridge vlan on mgmt cluster configured after update from 90_custom.yaml after node reboot.Additional changes done in the script for this is not needed.
@w13915984028 Thanks for pointing this out.

rrajendran17:
Testplan and logs:

1.1 node cluster running v1.5.0
2.Create VM Networks vlan2012 with vlan id 2012 and vlan2013 with vlan id 2013 under mgmt cluster
3.apply the following manifest having the custom image having the upgrade script changes
```
kind: Version
metadata:
  name: master
  namespace: harvester-system
spec:
  isoChecksum: bc692bb4278988bb34dc8568a3ee1d52a063fb9d234068c08dfdc5c234e6607dcef7c422abfbf190ec5422c2ab3cbeabc0ab9469de5f4514da3ec860568b9505
  isoURL: http://10.115.1.6/iso/renuka/upgrade1/harvester-9066893-dirty-amd64.iso

```
4.Click upgrade on the Harvester UI
5.Once the upgrade is completed successfully, check for the following output.
2021 is the mgmt vlan and 2012 and 2013 are VM networks created under mgmt cluster before upgrade.
They all should be part of mgmt-bo and mgmt vlan 2021 should be part of mgmt-br interface

```
bridge vlan show
port              vlan-id  
mgmt-br           1 PVID Egress Untagged
                  2021
mgmt-bo           1 PVID Egress Untagged
                  2012
                  2013
                  2021
harv-8:~ # kubectl get net-attach-def
NAME            AGE
vlan2012        12d
vlan2013        12d
harv-8:~ # kubectl get net-attach-def -o yaml vlan2012
apiVersion: k8s.cni.cncf.io/v1
metadata:
  annotations:
    network.harvesterhci.io/route: '{"mode":"auto","cidr":"10.115.8.0/21","gateway":"10.115.15.254","connectivity":"true"}'
  creationTimestamp: "2025-05-08T02:16:52Z"
  finalizers:
  - wrangler.cattle.io/harvester-network-nad-controller
  - wrangler.cattle.io/harvester-network-manager-nad-controller
  generation: 1
  labels:
    network.harvesterhci.io/clusternetwork: mgmt
    network.harvesterhci.io/ready: "true"
    network.harvesterhci.io/type: L2VlanNetwork
    network.harvesterhci.io/vlan-id: "2012"
  name: vlan2012
  namespace: default
  resourceVersion: "69915"
  uid: 71952c8d-dff2-4d1a-b220-2a1039db1d75
spec:
  config: '{"cniVersion":"0.3.1","name":"vlan2012","type":"bridge","bridge":"mgmt-br","promiscMode":true,"vlan":2012,"ipam":{}}'
harv-8:~ # kubectl get net-attach-def -o yaml vlan2013
apiVersion: k8s.cni.cncf.io/v1
metadata:
  annotations:
    network.harvesterhci.io/route: '{"mode":"auto","cidr":"10.115.16.0/21","gateway":"10.115.23.254","connectivity":"true"}'
  creationTimestamp: "2025-05-08T02:18:12Z"
  finalizers:
  - wrangler.cattle.io/harvester-network-nad-controller
  - wrangler.cattle.io/harvester-network-manager-nad-controller
  generation: 1
  labels:
    network.harvesterhci.io/clusternetwork: mgmt
    network.harvesterhci.io/ready: "true"
    network.harvesterhci.io/type: L2VlanNetwork
    network.harvesterhci.io/vlan-id: "2013"
  name: vlan2013
  namespace: default
  resourceVersion: "70866"
  uid: 85b169ca-206d-4987-93ca-df68ad2c34e9
spec:
  config: '{"cniVersion":"0.3.1","name":"vlan2013","type":"bridge","bridge":"mgmt-br","promiscMode":true,"vlan":2013,"ipam":{}}'
harv-8:~ # vim upgrade-manifest.yaml 
harv-8:~ # 


```
6.Upgrade logs from upgrade-hvst jobs in harvester-system namespace

```
+ replace_with_mgmtvlan
+ '[' -z v1.5.0 ']'
+ [[ ! v1.5.0 =~ ^v1\.5\.[0-9]$ ]]
+ local SETUPBOND_FILE=/host/etc/wicked/scripts/setup_bond.sh
+ local SETUPBRIDGE_FILE=/host/etc/wicked/scripts/setup_bridge.sh
+ local CUSTOM90_FILE=/host/oem/90_custom.yaml
+ local CONFIG_FILE=/host/oem/harvester.config
+ [[ ! -f /host/oem/harvester.config ]]
++ yq .install.managementinterface.vlanid /host/oem/harvester.config
+ vlan_id=2021
+ [[ -z 2021 ]]
+ [[ 2021 -eq 0 ]]
+ [[ -f /host/etc/wicked/scripts/setup_bond.sh ]]
+ sed -i 's/bridge vlan add vid 2-4094/bridge vlan add vid 2021/' /host/etc/wicked/scripts/setup_bond.sh
Updated /host/etc/wicked/scripts/setup_bond.sh with VLAN ID 2021
+ echo 'Updated /host/etc/wicked/scripts/setup_bond.sh with VLAN ID 2021'
+ [[ -f /host/etc/wicked/scripts/setup_bridge.sh ]]
+ sed -i 's/bridge vlan add vid 2-4094/bridge vlan add vid 2021/' /host/etc/wicked/scripts/setup_bridge.sh
+ echo 'Updated /host/etc/wicked/scripts/setup_bridge.sh with VLAN ID 2021'
+ [[ -f /host/oem/90_custom.yaml ]]
Updated /host/etc/wicked/scripts/setup_bridge.sh with VLAN ID 2021
+ sed -i 's/bridge vlan add vid 2-4094/bridge vlan add vid 2021/' /host/oem/90_custom.yaml
+ echo 'Updated /host/oem/90_custom.yaml with VLAN ID 2021'
Updated /host/oem/90_custom.yaml with VLAN ID 2021
+ upgrade_os
```


output of wicked scripts:

```
#!/bin/sh

ACTION=$1
INTERFACE=$2

case $ACTION in
        post-up)
                # inherit MAC address
                ip link set dev mgmt-br address $(ip -json link show dev $INTERFACE | jq -j '.[0]["address"]')

                # accept all vlan, PVID=1 by default
                bridge vlan add vid 2021 dev $INTERFACE
                ;;

esac
```

```
#!/bin/sh

ACTION=$1
INTERFACE=$2

case $ACTION in
        pre-up)
                # enable vlan-aware
                ip link set $INTERFACE type bridge vlan_filtering 1
                ;;

        post-up)
                # accept all vlan, PVID=1 by default
                bridge vlan add vid 2021 dev $INTERFACE self
                bridge vlan add vid 2021 dev mgmt-bo
                ;;
esac
```
Reboot the node and check if bridge vlan output stays intact

```
bridge vlan show
port              vlan-id  
mgmt-br           1 PVID Egress Untagged
                  2021
mgmt-bo           1 PVID Egress Untagged
                  2012
                  2013
                  2021
```

package/upgrade/upgrade_node.sh
```diff
@@ -508,6 +508,64 @@ calculateCPUReservedInMilliCPU() {
...
+  # Check if the file exists
+  if [[ -f "$CUSTOM90_FILE" ]]; then
+    # Replace the range with the VLAN ID
+    sed -i "s/bridge vlan add vid 2-4094/bridge vlan add vid $vlan_id/" "$CUSTOM90_FILE"
+    echo "Updated $CUSTOM90_FILE with VLAN ID $vlan_id"
+  else
+    echo "File not found: $CUSTOM90_FILE"
+  fi
+}
+
```
mingshuoqiu, line 568: The script seems to do the same thing on $CUSTOM90_FILE , $SETUPBRIDGE_FILE and $SETUPBOND_FILE. Should we use a common function with the file as parameter?
ex.
+  patch_vlan_file "$SETUPBOND_FILE"
+  patch_vlan_file "$SETUPBRIDGE_FILE"
+  patch_vlan_file "$CUSTOM90_FILE"

w13915984028 (approved):
LGTM, thanks.

package/upgrade/upgrade_node.sh
starbops: Thank you for clarifying.

starbops (approved):
LGTM, thank you.

mingshuoqiu (commented):
LGTM. Thanks

PR #8162, opened 2025-04-30
charts: bump harvester-node-disk-manager

Vicente-Cheng:
#### Problem:
We need to bump node-disk-manager for the issue https://github.com/harvester/harvester/issues/7706

#### Solution:
    - Bump chart v0.7.8
    - Bump image v0.7.10

#### Related Issue(s):

#7706


starbops (approved):
LGTM, thank you.

ibrokethecloud (approved):

Issue #8163, opened 2025-04-30
[BUG] KubeVirt CRD wasn't updated when upgrading


Yu-Jack:
### Describe the Bug

We have some users have been using Harvester for a long time, like using from Harvester v1.0. They might upgrade the Harvester step by step, like v1.0 -> v1.1 -> v1.2. Eventually, they upgraded it to v1.5.0.

During the latest upgrading, the cluster complained an error `decoding error: unknown field "spec.configuration.vmStateStorageClass"`. But, we actually have this field in deploy/charts.

https://github.com/harvester/harvester/blob/48a891b9015b8ce2376d52b26b1469d960c49c92/deploy/charts/harvester/dependency_charts/kubevirt-operator/crds/crd-kubevirt.yaml#L975-L977

Affected CRDs are under `deploy/charts/harvester/dependency_charts` like:

- csi-snapshotter
- kubevirt-operator
- cdi
- whereabouts

### To Reproduce

The most quick way to reproduce is to create a v1.1.3 Harvester, then upgrade it v1.2.1. The `vmStateStorageClass` is new field in v1.2.
 
- Install Harvester v1.1.3
- Upgrade it to v1.2.1
- Check `kubevirts.kubevirt.io` CRD.
- Search with this keyword `vmStateStorageClass`

### Expected Behavior

The CRD should be updated successfully.

### Support Bundle for Troubleshooting

This SB is generated from a cluster which is just upgraded from v1.1.3 to v1.2.1.

[supportbundle_80111823-d4fa-4324-a377-df83e8a459e7_2025-04-30T08-26-43Z.zip](https://github.com/user-attachments/files/19973779/supportbundle_80111823-d4fa-4324-a377-df83e8a459e7_2025-04-30T08-26-43Z.zip)

### Environment

I think all version have same problem, as long as CRDs needs upgrading. But, if you use Harvester v1.2 or v1.3, then upgrading to v1.5 should be fine.

### Additional context

Why this happens is because the `crds` folder position.

```
--templates
--crds

--templates
----crds
```

This produces different behaviors when running `helm install`

If you use first one

- It helps you install crd
- It **won't** help you upgrade crd (Even you add a new spec in CRD)
- It **won't** delete crd when uninstalling release
- No `app.kubernetes.io/managed-by: Helm` in labels

If you use second one

- It helps you install crd
- It helps you upgrade crd
- It deletes crd when uninstalling release
- Have `app.kubernetes.io/managed-by: Helm` in labels.

[...]

Issue #7912, opened 2025-03-20
[BUG] v1.5-c2e6602e-head CDI CRD version missing v1alpha1 compare to v1.5 branch head source code content.


albinsun:
**Describe the bug**
`v1.5-c2e6602e-head`'s CDI CRD version missing `v1alpha1` compare to `v1.5-head` content.
* `v1.5-c2e6602e-head`: `v1beta1`
* `v1.5` branch (source code): `v1alpha1`, `v1beta1`
  * Ref. https://raw.githubusercontent.com/harvester/harvester/refs/heads/v1.5/deploy/charts/harvester/dependency_charts/cdi/crds/cdi.yaml

**To Reproduce**
1. Provision a single node harvester and check cdi CRD version
   ![Image](https://github.com/user-attachments/assets/cb77701b-cdaa-40ff-a879-fff12aff840d)
2. Check YAML content
    ```
    harvester-node-0:~ # kubectl get Customresourcedefinitions/cdis.cdi.kubevirt.io -o yaml | grep v1
    apiVersion: apiextensions.k8s.io/v1
        name: v1beta1
                              of the ConfigMap (in yaml):\n\n\napiVersion: v1\nkind: ConfigMap\nmetadata:\n
      - v1beta1
    harvester-node-0:~ # 
    ```

**Expected behavior**
Expect CRD version the same as branch head, for example.
```
# ./check_dependency.sh https://raw.githubusercontent.com/harvester/harvester/refs/heads/v1.5/deploy/charts/harvester/dependency_charts/kubevirt-operator/crds/crd-kubevirt.yaml
Observation: v1 v1alpha3 
Expectation: v1 v1alpha3 
True
``` 

**Support bundle**
[supportbundle_2cbe8b45-501a-4bc1-a0b6-9f193a0262d5_2025-03-20T08-10-16Z.zip](https://github.com/user-attachments/files/19361839/supportbundle_2cbe8b45-501a-4bc1-a0b6-9f193a0262d5_2025-03-20T08-10-16Z.zip)

**Environment**
 - Harvester ISO version: `v1.5-c2e6602e-head`
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): Single node QEMU/KVM

**Additional context**
n/a

albinsun:
[...]

Issue #7958, opened 2025-03-28
[BUG] Backing images' on-disk data might be removed during upgrade


bk201:
### Describe the Bug

When upgrading from Harvester v1.3.2 to v1.4.0 (Longhorn v1.6.2 to v1.7.2), there is a chance the backing image's data will be deleted.

### To Reproduce

1. Upgrading from v1.3.2 to v1.40 and noticing some VMs can't start due to PVC can't be ready.

### Expected Behavior

Backing images should be available and ready.

### Environment

- Harvester version: v1.3.2 to v1.4.0
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Additional context

Longhorn upstream issue: https://github.com/longhorn/longhorn/issues/10644

### Workaround and Mitigation

Applying Longhorn v1.7.2 backing image CRD before the upgrade can avoid the issue.

bk201:
We'll need to verify the workaround and publish a known issue on the doc site (plus updating release note.)

WebberHuang1118:
I've tried the workaround to upgrade Harvester `v1.3.2` -> `v1.4.0` (LH `v1.6.2` -> `v1.7.2`)
- Patching harvester managechart to avoid complaints
  ```
  $ kubectl patch managedchart harvester \
  -n fleet-local \
  --type='json' \
  -p='[
    {
      "op":"add",
      "path":"/spec/diff/comparePatches/-",
      "value": {
        "apiVersion":"apiextensions.k8s.io/v1",
        "jsonPointers":["/spec","/metadata/annotations", "/metadata/labels", "/status"],
        "kind":"CustomResourceDefinition",
        "name":"backingimages.longhorn.io"
      }
    }
  ]'
  ```
- Applying [BI CRD from LH v1.7.2](https://github.com/longhorn/longhorn/blob/bd3b19984d8dcaa41e84edd207a385ef6befa907/deploy/longhorn.yaml#L486-L690) ahead of the upgrade
  ```
  apiVersion: apiextensions.k8s.io/v1
  kind: CustomResourceDefinition
  metadata:
    annotations:
      controller-gen.kubebuilder.io/version: v0.15.0
    labels:
      app.kubernetes.io/name: longhorn
      app.kubernetes.io/instance: longhorn
      app.kubernetes.io/version: v1.7.2
      longhorn-manager: ""
    name: backingimages.longhorn.io
  spec:
[...]

PR #8170, opened 2025-05-02
feat: move dependency crds to harvester-crd folder

Yu-Jack:
#### Problem:
The problem is pretty clear in original issue. So, I skip it here.

#### Solution:

We should consider two situations: installation and upgrading.

There are different solutions here:

- Use bindata to manage CRDs
- Create another managedChart to manage CRDs
- Apply CRDs independently [here](https://github.com/harvester/harvester/blob/71f3505c9bf9d956267065846642098850ebea23/package/upgrade/upgrade_manifests.sh#L1358)

Each Drawback:

- Order issue. It's hard to guarantee that other components can use latest CRDs. With this way, CRDs are updated after Harvester starts up. It might have potential bug while upgrading.
- Too many managedCharts. We've had a lot of workaround ways to deal with managedChart [here](https://github.com/harvester/harvester/blob/master/package/upgrade/upgrade_manifests.sh#L1358). I don't think we might want more.
- Those CRDs are still not in the control. They don't belong to any management flow.

The quick way is to put those dependency CRDs under harvester-crd. Then, with [takeOwnerShip field](https://github.com/harvester/harvester-installer/blob/90668931a78c860f309487fe100be5ef868c3f81/pkg/config/templates/rancherd-10-harvester.yaml#L235), it helps us add labels and annotations back and manages those CRDs.

***However, we won't put cdi under harvester-crd because cdi-operator will manage CRD status. For example, they will [delete old version here](https://github.com/kubevirt/containerized-data-importer/blob/e136558c999acce03291849aacc8154026e58445/pkg/operator/controller/callbacks.go#L61-L62). This is also mentioned in https://github.com/harvester/harvester/issues/7912#issuecomment-2743944344.***

#### Related Issue(s):
#8163 

#### Test plan:

1. Install new master-head or v1.5-head
2. Upgrade from v1.4.3 to v1.5-head
3. Upgrade from v1.5 to master-head or v1.5-head

You could add a new dummy field in those CRDs, then test it.

And before upgrade, please ensure that `cluster-repo` has been patched.

```
sudo mount -o loop ~/Downloads/harvester-master-cpu-pinning/harvester-master-amd64.iso /mnt/iso
scp -O /mnt/iso/bundle/harvester/images/harvester-repo-images-master.tar.zst rancher@xxxx:/home/rancher

# In node
docker load < /home/rancher/harvester-repo-images-master.tar.zst
# then, kill the cluster-repo pod
```

With this, your master-build iso can work.

Regardless of cases, those CRDs should contain specific labels and annotation like below:

- helm.sh/resource-policy: keep
- meta.helm.sh/release-name: harvester-crd
- meta.helm.sh/release-namespace: harvester-system
- objectset.rio.cattle.io/id: default-mcc-harvester-crd-cattle-fleet-local-system
- app.kubernetes.io/managed-by: Helm
- objectset.rio.cattle.io/hash: d4a83267ddde6a8769c04362d4a0e5605db9baa7

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    helm.sh/resource-policy: keep
    meta.helm.sh/release-name: harvester-crd
    objectset.rio.cattle.io/id: default-mcc-harvester-crd-cattle-fleet-local-system
  creationTimestamp: "2025-05-06T14:05:28Z"
  generation: 1
  labels:
    objectset.rio.cattle.io/hash: d4a83267ddde6a8769c04362d4a0e5605db9baa7
    operator.kubevirt.io: ""
  name: kubevirts.kubevirt.io
  resourceVersion: "623701"
  uid: ea688774-d544-4f5d-bc61-a4abf76175d9
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.14.0
    helm.sh/resource-policy: keep
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"apiextensions.k8s.io/v1","kind":"CustomResourceDefinition","metadata":{"annotations":{"controller-gen.kubebuilder.io/version":"v0.14.0"},"name":"overlappingrangeipreservations.whereabouts.cni.cncf.io"},"spec":{"group":"whereabouts.cni.cncf.io","names":{"kind":"OverlappingRangeIPReservation","listKind":"OverlappingRangeIPReservationList","plural":"overlappingrangeipreservations","singular":"overlappingrangeipreservation"},"scope":"Namespaced","versions":[{"name":"v1alpha1","schema":{"openAPIV3Schema":{"description":"OverlappingRangeIPReservation is the Schema for the OverlappingRangeIPReservations API","properties":{"apiVersion":{"description":"APIVersion defines the versioned schema of this representation of an object.\nServers should convert recognized schemas to the latest internal value, and\nmay reject unrecognized values.\nMore info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources","type":"string"},"kind":{"description":"Kind is a string value representing the REST resource this object represents.\nServers may infer this from the endpoint the client submits requests to.\nCannot be updated.\nIn CamelCase.\nMore info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds","type":"string"},"metadata":{"type":"object"},"spec":{"description":"OverlappingRangeIPReservationSpec defines the desired state of OverlappingRangeIPReservation","properties":{"containerid":{"type":"string"},"ifname":{"type":"string"},"podref":{"type":"string"}},"required":["podref"],"type":"object"}},"required":["spec"],"type":"object"}},"served":true,"storage":true}]}}
    meta.helm.sh/release-name: harvester-crd
    objectset.rio.cattle.io/id: default-mcc-harvester-crd-cattle-fleet-local-system
  creationTimestamp: "2025-05-06T14:05:29Z"
  generation: 1
  labels:
    objectset.rio.cattle.io/hash: d4a83267ddde6a8769c04362d4a0e5605db9baa7
  name: overlappingrangeipreservations.whereabouts.cni.cncf.io
  resourceVersion: "623709"
  uid: 19decf91-ad09-432f-abf3-22f227e01e66
```


innobead (dismissed):

w13915984028:
Please hold this PR a while for more thoughtful review and test, before final merge. I also notice another issue https://github.com/harvester/harvester/issues/8239, surprised it exists for such a long time.

And, this PR seems not touching LH's CRD,  what's the consideration to leave it?  thanks.

Even `cdi`, also needs another look:

Harvester values.yaml sets options upon cdi, if it relys on new cdi CRD but it was not upgrade in `harvester-crd`, it will either fail or miss silently.

https://github.com/harvester/harvester/blob/7f09341ba74b949ee05968078ed1cbc0dca1f0ab/deploy/charts/harvester/values.yaml#L114

```
cdi:
  enabled: true
  containers:
    version: &cdiVersion 1.61.0-150600.3.12.1
    operator:
```

Yu-Jack:
> I also notice another issue https://github.com/harvester/harvester/issues/8239, surprised it exists for such a long time.

Yeah, me too. We should ensure the upgrading keeps going when user changes some upgrade-related yaml, such as managedChart.

> And, this PR seems not touching LH's CRD, what's the consideration to leave it? thanks.

LH's CRDs has been under templates, so we don't need to handle it. And I don't think we can add takeOwnerShip field in harvester managedChart.

> Harvester values.yaml sets options upon cdi, if it relys on new cdi CRD but it was not upgrade in harvester-crd, it will either fail or miss silently.

So, we still need to apply the new cdi CRD during upgrading like this: https://github.com/harvester/harvester/blob/7f09341ba74b949ee05968078ed1cbc0dca1f0ab/package/upgrade/upgrade_manifests.sh#L1335-L1340

w13915984028:
@Yu-Jack Did you have a detailed test plan and test report? The influence of this PR might not be noticable directly.

I would suggest at least test those 3 scenarios, say the new version is v1.5.1
(1) New installation on v1.5.1, harvester-crd and harvester are rolled out correctly, managedchart has no errors, `vm, volume, storage-network, LH v2`, cdi related are all working smoothly.
(2) Upgrade from v142/v143 to v151, with configured `vm, volume, storage-network, LH v2`, and the upgrade goes well
(3) Upgrade from v151 to v151/master-head

thanks.

Yu-Jack:
@w13915984028  I've tested with case 1 and case 3. You just need to add some dummy fields in those CRDs. Regardless of cases, your CRDs should have specific labels and annotation like I mentioned in my Test Plan.  I'll add more detailed flow to my original post. For case 2, I'll test it later. Thanks!

w13915984028:
It is more than dummy fields to test.

managedchart `harvester` is a super-set, e.g, it sets params upon LH, if LH's CRD was not updated yet, and the apply may not be effective, it is same to cdi and others.

https://github.com/harvester/harvester/blob/7f09341ba74b949ee05968078ed1cbc0dca1f0ab/deploy/charts/harvester/values.yaml#L466
```
longhorn:
  ## Specify whether to install longhorn,
  ## defaults to "true"
  enabled: true

  createNamespace: true

  ## Specify the namespace to install longhorn.
  namespaceOverride: longhorn-system

  ## mark `longhorn` does not be default
  persistence:
    defaultClass: false
    migratable: true

  csi:
    ## Specify the kubelet root directory.
    kubeletRootDir: /var/lib/kubelet

  defaultSettings:
    ## Specify the concurrent automatic engine upgrade per node limit
    concurrentAutomaticEngineUpgradePerNodeLimit: 3
    priorityClass: &longhornPriorityClass system-cluster-critical
    autoCleanupSnapshotWhenDeleteBackup: true
```

currently, whereabouts and cdi are updated from `apply_extra_nonversion_manifests`, it ensures they are before Harvester.

```
apply_extra_nonversion_manifests
upgrade_harvester
```

Yu-Jack:
> It is more than dummy fields to test.
> 
> managedchart harvester is a super-set, e.g, it sets params upon LH, if LH's CRD was not updated yet, and the apply may not be effective, it is same to cdi and others.

LH puts CRDs under templates, which has been different from us. LH shouldn't have this kind of problem. For example, you could check this [new `backingimages` CRD](https://github.com/longhorn/longhorn/commit/7b36b870327afaa82aed67a291917364fea37ccf), then download SB from this issue https://github.com/harvester/harvester/issues/8203. You'll see the `backingimages` is still updated.

[The v1.7.3 `backingimages`](https://github.com/longhorn/longhorn/blob/76fb3e789ab6b221a4bdfee96ffff3800f614501/chart/templates/crds.yaml#L377-L578) doesn't contain `dataEngine`.
[The v1.8.1 `backingimages`](https://github.com/longhorn/longhorn/blob/30f817bba6038165dd213e79f3bb89e28aa89d54/chart/templates/crds.yaml#L380-L609) contains it.

The SB from #8203 also contains this one with following labels and annotations.

```
annotations:
  controller-gen.kubebuilder.io/version: v0.15.0
  helm.sh/resource-policy: keep
  sim.harvesterhci.io/creationTimestamp: "2023-04-25T07:14:23Z"
creationTimestamp: "2023-04-25T07:14:23Z"
generation: 1
labels:
  app.kubernetes.io/instance: harvester
  app.kubernetes.io/name: longhorn
  app.kubernetes.io/version: v1.8.1
  helm.sh/chart: longhorn-1.8.1
name: backingimages.longhorn.io
resourceVersion: "142"
uid: 4438125b-66c3-4e35-9fa9-94da2f878223
```

> currently, whereabouts and cdi are updated from apply_extra_nonversion_manifests, it ensures they are before Harvester.
> 
> apply_extra_nonversion_manifests
> upgrade_harvester


Yeah, with this PR, those CRDs are updated more early before `apply_extra_nonversion_manifests`. Actually, you could try it.

w13915984028:
BTW, Let's double check if issue like https://github.com/harvester/harvester/issues/7958 can be solved by this enhancement.

Yu-Jack:
> BTW, Let's double check if issue like https://github.com/harvester/harvester/issues/7958 can be solved by this enhancement.

well ... I didn't know LH had this issue before. If so, I think it's better put LH's CRDs under harvester-crd. But, those CRDs has contained specific labels and annotations. I'm not sure whether it works if we put them under harvester-crd with takeOwnerShip. 

Although LH's CRDs are managed by helm, but it might encounter some installation/upgrading errors in some situations. Like this, this is [my example testing repository](https://github.com/Yu-Jack/helmtest/tree/main/v3) to demo CRD installation error. I put this here for other people to try.

Vicente-Cheng:
>BTW, Let's double check if issue like https://github.com/harvester/harvester/issues/7958 can be solved by this enhancement.

IIUC, Longhorn side has already enhanced this behavior. (REF: https://github.com/longhorn/longhorn-manager/pull/3651)
With those changes, Longhorn major components (like lh-manager/instance-engine/...etc) will wait for the CRD to be applied.

The changes will be introduced with the following version:
- 1.6.5
- 1.7.4
- 1.8.2
- later version

So, we still need to take care of this case on v1.4.3
cc @WebberHuang1118, @tserong

And CDI is the tricky case, as @Yu-Jack mentioned on the description
>However, we won't put cdi under harvester-crd because cdi-operator will manage CRD status. For example, they will [delete old version here](https://github.com/kubevirt/containerized-data-importer/blob/e136558c999acce03291849aacc8154026e58445/pkg/operator/controller/callbacks.go#L61-L62). This is also mentioned in https://github.com/harvester/harvester/issues/7912#issuecomment-2743944344.

We still need the upgrade script for CDI CRD.

Vicente-Cheng (dismissed):
lgtm, thanks for the enhancement!

FrankYang0529 (dismissed):

tserong:
> So, we still need to take care of this case on v1.4.3

Just to clarify, do you mean we need to backport this PR to v1.4.3?

Yu-Jack:
LH uses strict way to detect fields ([Reference](https://kubernetes.io/blog/2023/04/24/openapi-v3-field-validation-ga/#server-side-field-validation)). It seems if fields are not updated, LH will raise an error and run upgrade again.  For v1.4.3 Harvester, I guess there is a chance that LH upgrade manager might fail. But, I'm surprised that we don't have this issue in Harvester side yet.

starbops (dismissed):
LGTM, thank you.

Yu-Jack:
I just updated the `tgz` of dependency chart.

starbops (dismissed):

w13915984028 (commented):
Another open question:

After this change, there is no version info directly reading from on `deploy/charts/harvester-crd/templates` 

The current `harvester` chart has.
image

We need a solution, either add the version info into the file name; or use another chart file to record the info. 

The version info helps the furture CRD bumping, validating and others.

WDYT @Yu-Jack , thanks.

Yu-Jack:
> We need a solution, either add the version info into the file name; or use another chart file to record the info.
>
> The version info helps the furture CRD bumping, validating and others.
>
> WDYT @Yu-Jack , thanks.

Good idea. Let me add version as suffix.

Yu-Jack:
I'll add application version.

Vicente-Cheng (approved):

starbops:
@Yu-Jack Could you help clarify what the steps/workflow look like if someone wants to update a dependency chart, e.g., whereabouts, after this change is introduced? Thank you.

w13915984028 (approved):
LGTM, thanks for the kind fix.

Harvester README.md needs an update, e.g.
https://github.com/harvester/harvester/blob/master/deploy/charts/harvester/README.md#chart-details

And harvester-crd needs a new README.md to describe the bump steps ...

Yu-Jack:
When everything is settled, I'll write a new README.

Yu-Jack:
I'll open another PR for that. I don't want to keep dismissing the reviewers in this PR ...

starbops (approved):

FrankYang0529 (approved):

ibrokethecloud (approved):
lgtm. thanks.

ihcsim (approved):

ihcsim:
Thanks for fixing this. My only thought on this is whether it's truly necessary to define these external charts as dependencies. Although it gives us the convenience of exposing all the children charts values in the parent's values.yaml, we are already managing these charts in the `dependency_charts` folder. We could have also just added some `override-values.yaml` in the children charts. Then provide doc instructions to install everything within the `dependency_charts` folder first.

That way, we won't have to manually manage those external CRDs.

Yu-Jack:
> Thanks for fixing this. My only thought on this is whether it's truly necessary to define these external charts as dependencies. Although it gives us the convenience of exposing all the children charts values in the parent's values.yaml, we are already managing these charts in the `dependency_charts` folder. We could have also just added some `override-values.yaml` in the children charts. Then provide doc instructions to install everything within the `dependency_charts` folder first.
> 
> That way, we won't have to manually manage those external CRDs.

It's really hard to say. Let's take KubeVirt as an example. KubeVirt doesn't manage its own CRDs, so we still need to manually download the CRD for a specific version.

This is different from Longhorn, which manages its CRDs internally, so we always get the correct version when installing it.

Issue #5848, opened 2024-05-22
[FEATURE] Live migration network


Vicente-Cheng:
**Is your enhancement related to a problem? Please describe.**
Currently, VM live migration uses the management network. It might bring some issues when many VM want to migrate simultaneously. We need to sync the memory dirty (through the mgmt network), and then the migration could be done.
The migration might take longer if the mgmt network has only 1GB capability.

**Describe the solution you'd like**
kubevirt already provides a way to config another interface for live migration.
Ref: https://kubevirt.io/user-guide/operations/live_migration/#using-a-different-network-for-migrations
We could check this and then try to introduce it in future releases.

**Describe alternatives you've considered**
None

**Additional context**
TBD

richevanscybermyte:
This is needed but I wrote up a workaround in case there is anyone that needs faster then 1G for live migrations and are willing to get into the cli (kubectl)

https://cybermyte.io/blog/rancher-harvester-live-migration-network/

It was something I was also struggling with in my production so I set out to solve it myself. I unfortunately have no GUI skillz (as of today) so I cannot help on that side but at least this solution works in the mean time. Let me know if there are any questions.

FrankYang0529:
Manual steps to setup live migration network:
1. Use ipxe to setup 3-node cluster. https://github.com/harvester/ipxe-examples/tree/main/vagrant-pxe-harvester
2. Add new network "Virtual network 'harvester'" on each node.
    <img width="760" alt="Image" src="https://github.com/user-attachments/assets/dd8de2f6-b209-47a7-b8cf-d3a18e6796c9" />
3. Run `ip a` on each node and there is `ens7` on each node.
4. Run `ip link set ens7 up` on each node.
5. Create `thisistest` Cluster Network.
6. Create Network with "Select all nodes" and and NIC "ens7".
7. Create net-attach-def.
```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  annotations:
    network.harvesterhci.io/route: '{"mode":"auto","cidr":"192.168.3.0/24","gateway":"192.168.3.1","connectivity":"true"}'
  labels:
    network.harvesterhci.io/clusternetwork: thisistest
    network.harvesterhci.io/ready: "true"
    network.harvesterhci.io/type: L2VlanNetwork
    network.harvesterhci.io/vlan-id: "1"
  name: vlan1
  namespace: harvester-system
spec:
[...]

PR #8174, opened 2025-05-02
feat: add vm-migration-network setting

FrankYang0529:
### Problem:
Currently, the harvester only supports VM migration on mgmt network.

### Solution:
Support to use other [migration network](https://kubevirt.io/user-guide/compute/live_migration/#using-a-different-network-for-migrations).

### Related Issue(s):
https://github.com/harvester/harvester/issues/5848

#### Setup
1. Use [ipxe vagrant](https://github.com/harvester/ipxe-examples/tree/main/vagrant-pxe-harvester) to create a 3-node harvester.
2. Add new network "Virtual network 'harvester'" on each node.
    Image
3. Run `ip a` on each node and there is `ens7` on each node.
4. Run `ip link set ens7 up` on each node.
5. Add new network "Virtual network 'default'" on each node.
6. Run `ip a` on each node and there is `ens8` on each node.
7. Run `ip link set ens8 up` on each node.
8. Create `test-1` and `test-2` Cluster Network.
9. Create Network under `test-1` with "Select all nodes" and and NIC "ens7".
10. Create Network under `test-2` with "Select all nodes" and and NIC "ens8".

#### Case 1: Change `vm-migration-network` from default value to `test-1` network
1. Modify `vm-migration-network` setting with value `{"clusterNetwork":"test-1","vlan":1,"range":"10.1.2.0/24"}`.
2. Check `harvester-system/kubevirt` KubeVirt resource, the `Available` type condition has reason `UpdateInProgress`.
3. Before the `Available` type condition has reason `AllComponentsReady`, users cannot migrate VM.
4. After the `Available` type condition has reason `AllComponentsReady`, migrate VM can work.

#### Case 2: Change `vm-migration-network` from `test-1` to `test-2` network
1. Modify `vm-migration-network` setting with value `{"clusterNetwork":"test-2","vlan":1,"range":"10.1.2.0/24"}`.
2. Check `harvester-system/kubevirt` KubeVirt resource, the `Available` type condition has reason `UpdateInProgress`.
3. Before the `Available` type condition has reason `AllComponentsReady`, users cannot migrate VM.
4. After the `Available` type condition has reason `AllComponentsReady`, migrate VM can work.

#### Case 3: Change `vm-migration-network` from `test-2` to default value
1. Modify `vm-migration-network` setting with value `""`.
2. Check `harvester-system/kubevirt` KubeVirt resource, the `Available` type condition has reason `UpdateInProgress`.
3. Before the `Available` type condition has reason `AllComponentsReady`, users cannot migrate VM.
4. After the `Available` type condition has reason `AllComponentsReady`, migrate VM can work.

#### Case 4: Webhook deny wrong setting
* Non-existent cluster network: `{"clusterNetwork":"non-existent","vlan":1,"range":"10.1.2.0/24"}`
* Range doesn't provide enough ip all nodes: `{"clusterNetwork":"test-2","vlan":1,"range":"10.1.2.0/32"}`
* (Range - Exclude) doesn't provide enough ip all nodes: `{"clusterNetwork":"test-2","vlan":1,"range":"10.1.2.0/30","exclude":["10.1.2.0/32","10.1.2.1/32"]}`

### DIFF
--- a/pkg/controller/master/kubevirt/vm_migration_network.go
+++ b/pkg/controller/master/kubevirt/vm_migration_network.go
@@ -1,0 +1,254 @@
+package kubevirt
+
+import (
+	"context"
+	"crypto/sha1"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+
+	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/config"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
+	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/settings"
+	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/util/network"
+)
+
+const (
+	ControllerName = "harvester-vm-migration-network-controller"
+
+	VMMigrationNetworkPrefix         = "vm-migration-network.settings.harvesterhci.io"
+	VMMigrationNetworkHashLabel      = VMMigrationNetworkPrefix + "/hash"
+	VMMigrationNetworkHashAnnotation = VMMigrationNetworkHashLabel
+	VMMigrationNetworkNadAnnotation  = VMMigrationNetworkPrefix + "/net-attach-def"
+
+	KubeVirtName                         = "kubevirt"
+	VMMigrationNetworkNetAttachDefPrefix = "vm-migration-network-"
+)
+
+type Handler struct {
+	ctx                               context.Context
+	settings                          ctlharvesterv1.SettingClient
+	networkAttachmentDefinitions      ctlcniv1.NetworkAttachmentDefinitionClient
+	networkAttachmentDefinitionsCache ctlcniv1.NetworkAttachmentDefinitionCache
+	kubevirts                         ctlkubevirtv1.KubeVirtClient
+	kubevirtCache                     ctlkubevirtv1.KubeVirtCache
+}
+
+// register the setting controller and reconsile longhorn setting when storage network changed
+func Register(ctx context.Context, management *config.Management, _ config.Options) error {
+	settings := management.HarvesterFactory.Harvesterhci().V1beta1().Setting()
+	networkAttachmentDefinitions := management.CniFactory.K8s().V1().NetworkAttachmentDefinition()
+	kubevirts := management.VirtFactory.Kubevirt().V1().KubeVirt()
+
+	controller := &Handler{
+		ctx:                               ctx,
+		settings:                          settings,
+		networkAttachmentDefinitions:      networkAttachmentDefinitions,
+		networkAttachmentDefinitionsCache: networkAttachmentDefinitions.Cache(),
+		kubevirts:                         kubevirts,
+		kubevirtCache:                     kubevirts.Cache(),
+	}
+
+	settings.OnChange(ctx, ControllerName, controller.OnVMMigrationNetworkChange)
+	return nil
+}
+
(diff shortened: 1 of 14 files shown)


w13915984028:
Those PRs are related, before this PR is merged, we need to do a cross check for potential further update/enhancement on them, thanks.

- [ ] https://github.com/harvester/docs/pull/640

- [ ] https://github.com/harvester/network-controller-harvester/pull/149

innobead:
@FrankYang0529 can you fix the conflicts? Thanks.

pkg/controller/master/kubevirt/vm_migration_network.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return setting, nil
+}
+
+func (h *Handler) OnVMMigrationNetworkChange(_ string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	if setting == nil || setting.DeletionTimestamp != nil || setting.Name != settings.VMMigrationNetworkSettingName {
+		return setting, nil
+	}
+	settingCopy := setting.DeepCopy()
+
```
ibrokethecloud: i am not sure if we need to explicitly check the hash here

the setting handler does this for all settings https://github.com/harvester/harvester/blob/master/pkg/controller/master/setting/handler.go#L88

ibrokethecloud (commented):
the PR attempts to create the NAD object from the settings. would it not be easier to just let user create their own custom ClusterNetwork/VM Network definition and leverage the same in the setting.

pkg/controller/master/kubevirt/vm_migration_network.go
```diff
@@ -0,0 +1,252 @@
...
+	if !reflect.DeepEqual(kubevirt.Spec.Configuration, kubevirtCopy.Spec.Configuration) {
+		if _, err = h.kubevirts.Update(kubevirtCopy); err != nil {
+			return setting, fmt.Errorf("update kubevirt error %v", err)
+		}
+	}
+
+	if removeOldNad {
+		oldNadNamespaceAndName := strings.Split(oldNadAnnotation, "/")
+		if len(oldNadNamespaceAndName) == 2 {
+			h.removeOldNad(oldNadNamespaceAndName[0], oldNadNamespaceAndName[1])
```
ibrokethecloud: we do not return an error and fail with a warning, should we return an error or just add more retry logic in the `removeOldNad` function before we log the error?

pkg/webhook/resources/setting/validator.go
```diff
@@ -1465,6 +1545,25 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
...
+	count, err := webhookUtil.GetUsableIPAddressesCount(config.Range, config.Exclude)
+	if err != nil {
+		return err
+	}
+
+	// 1 node has 1 virt-handler which needs 1 IP address.
+	if count < len(nodes) {
+		return fmt.Errorf("allocatable IP address range is < %d,allocate sufficient range", len(nodes))
+	}
+
```
ibrokethecloud, line 1586: should we have a check to ensure no VM migrations are in flight before accepting the change to the setting?

FrankYang0529:
Hi @ibrokethecloud, I addressed all comments. Could you review again when you have time? Thanks.

pkg/webhook/resources/setting/validator.go
```diff
@@ -1465,6 +1559,25 @@ func (v *settingValidator) checkStorageNetworkRangeValid(config *storagenetworkc
 	return nil
 }
 
+func (v *settingValidator) checkVMMigrationNetworkRangeValid(config *networkutil.Config) error {
+	nodes, err := v.nodeCache.List(labels.Everything())
```
starbops, line 1563: Probably not that important, but do we want to exclude witness nodes?

starbops (commented):
Thanks for the PR. Although I haven't tested it yet, a question came to mind immediately after reading the KubeVirt documentation: in their documentation, they use macvlan NADs, but in Harvester's context, we have always (at least before introducing Kube-OVN) used bridge NADs. Will that be a problem?

FrankYang0529:
@starbops, thanks for the review. Yes, in KubeVirt document, they use macvlan as example. The NAD in KubeVirt configuration is used to allocate IP to virt-handler on each node. The bridge can work on my environment.

pkg/webhook/resources/setting/validator.go
FrankYang0529: Addressed it. Thanks.

pkg/controller/master/kubevirt/vm_migration_network.go
FrankYang0529: Return error if the handler cannot remove old NAD. Thanks.

ibrokethecloud (approved):
lgtm. thanks.

WebberHuang1118:
Hi @FrankYang0529 

If I understand correctly (IIUC), this PR doesn't seem to check whether the IP range overlaps with the range used by the storage network. [Ensuring sufficient IPs for the storage network is critical](https://docs.harvesterhci.io/v1.6/advanced/storagenetwork#configuration-example) — otherwise, it could lead to I/O errors, VM image creation failures, and potentially impact Harvester upgrades.

cc @innobead @Vicente-Cheng

FrankYang0529:
@WebberHuang1118 You're correct. I will create a follow-up PR to validate it.

FrankYang0529:
@WebberHuang1118 thanks for the review. Follow-up PR is here: https://github.com/harvester/harvester/pull/8546

Issue #8166, opened 2025-05-01
[BUG] Images created with Terraform provider do not have storage classes labelled - Generate template gives error about storageclass.storage.k8s.io "" not found.


wstutt:
### Describe the Bug

After using terraform to create VM image resources, when you create a VM, and then attempt to generate a template from the resultant virtual machine, the UI returns the error:
 
storageclass.storage.k8s.io "" not found

### To Reproduce

1) use terraform to create a vm_image resource similar to this:

```
resource "harvester_image" "ubuntu20" {
  name      = "ubuntu20"
  namespace = "harvester-public"

  display_name = "ubuntu-20.04-server-cloudimg-amd64.img"
  source_type  = "download"
  url          = "http://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64.img"
}
```

2) Create a virtual machine from an image

3) click ellipses -> generate template -> with data

4) see error

Also note in the "images" tab that there are no storage classes associated with VM image.

### Expected Behavior

expect to be able to generate templates from virtual machines.

### Support Bundle for Troubleshooting

![Image](https://github.com/user-attachments/assets/7388a925-18ad-4500-b8f4-86872ad4d989)

### Environment

- Harvester version: 1.5.0
- Impacted VM: any kvm template
- Impacted volume (PV): longhorn v1 volume for virtual machine
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): baremetal 
- Rancher version: 2.11.1

### Workaround and Mitigation

On the virtual machines that I need to generate local templates, I have just used the UI to download the images and these virtual machines are behaving as expected.

Vicente-Cheng:
Hi @wstutt, 

I did the quick test with the following steps and it worked
1. create the VM image (ubuntu 22.04) with Longhorn v1
2. create a VM with the above image
3. generate template from the above VM

Could you generate the template again (see the error) and generate the SB for the investigation?

wstutt:
Tested on both 0.6.6 and 0.6.7 of the Terraform provider.

using the UI to download the same image shows longhorn-system as the storage class, as expected.

ihcsim:
Does it work if you add the `storage_class_name` field to your `harvester_image` resource? See examples [here](https://github.com/harvester/terraform-provider-harvester/blob/master/docs/resources/image.md#example-usage). If not, please generate and attach a support bundle.

wstutt:
[...]

PR #8181, opened 2025-05-04
webhook/image: ensure the StroageClass is set

Vicente-Cheng:
#### Problem:
If we create the VM image through terraform provider, we will lack the two fields `metadata/annotations/harvesterhci.io/storageClassName` and `spec.targetStorageClass`. The VM template generator depends on the `metadata/annotations/harvesterhci.io/storageClassName`. If we lack this field, the VM template will fail to generate.

#### Solution:
Ensure these two fields would be set even if we use the default storage class

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8166

#### Test plan:
1. create VM image through terraform provider with following tf content
```
resource "harvester_image" "ubuntu2204" {
  name      = "ubuntu22"
  namespace = "default"

  display_name = "ubuntu2204-server-cloudimg-amd64.img"
  source_type  = "download"
  url          = "https://free.nchc.org.tw/ubuntu-cloud-images/jammy/current/jammy-server-cloudimg-amd64-disk-kvm.img"
}
```
2. create VM with the above image
3. generate the VM template with the above vm

### DIFF
--- a/pkg/image/common/mutator.go
+++ b/pkg/image/common/mutator.go
@@ -8,8 +8,9 @@ import (
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	ctlstoragev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/storage/v1"
 	"github.com/rancher/wrangler/v3/pkg/slice"
+	"github.com/sirupsen/logrus"
 	storagev1 "k8s.io/api/storage/v1"
-	"k8s.io/apimachinery/pkg/labels"
+	"kubevirt.io/kubevirt/pkg/apimachinery/patch"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/util"
@@ -18,6 +19,7 @@ import (
 
 type VMIMutator interface {
 	PatchImageSCParams(vmi *harvesterv1.VirtualMachineImage) ([]string, error)
+	EnsureTargetSC(vmi *harvesterv1.VirtualMachineImage) ([]string, error)
 }
 
 type vmiMutator struct {
@@ -63,7 +65,8 @@ func (m *vmiMutator) getSC(scName string) (*storagev1.StorageClass, error) {
 			return nil, err
 		}
 		if storageClass.Provisioner != longhorntypes.LonghornDriverName {
-			return nil, fmt.Errorf("the provisioner of storageClass must be %s, not %s", longhorntypes.LonghornDriverName, storageClass.Provisioner)
+			logrus.Warnf("The provisioner of storageClass must be %s, not %s. We will use the default parameters.", longhorntypes.LonghornDriverName, storageClass.Provisioner)
+			return nil, nil
 		}
 		if storageClass.Parameters[util.LonghornOptionBackingImageName] != "" {
 			return nil, errors.New("can not use a backing image storageClass as the base storageClass template")
@@ -71,16 +74,12 @@ func (m *vmiMutator) getSC(scName string) (*storagev1.StorageClass, error) {
 		return storageClass, nil
 	}
 
-	storageClasses, err := m.scCache.List(labels.Everything())
-	if err != nil {
-		return nil, err
+	defaultSC := util.GetDefaultSC(m.scCache)
+	if defaultSC == nil {
+		return nil, fmt.Errorf("no default storageClass found for backingImage")
 	}
-
-	for _, storageClass := range storageClasses {
-		if storageClass.Annotations[util.AnnotationIsDefaultStorageClassName] == "true" &&
-			storageClass.Provisioner == longhorntypes.LonghornDriverName {
-			return storageClass, nil
-		}
+	if defaultSC.Provisioner == longhorntypes.LonghornDriverName {
+		return defaultSC, nil
 	}
 
 	return nil, nil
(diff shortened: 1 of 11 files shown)


Vicente-Cheng:
Hello folks, this is ready to review.

The main goal for this PR is to ensure these two things
- annotation `harvesterhci.io/storageClassName`
- spec.targetStorageClass

We now rely on the annotation `harvesterhci.io/storageClassName` for the VM template generator.
So, we need to ensure every VM Image has this annotation.

Another notable change is that if the default SC is not the Longhorn based, the backingimage will use the default SC parameter for the VM Image (These parameters will be used to create the SC associated with the PVC which comes from backingimage)

cc @ihcsim, @FrankYang0529, @WebberHuang1118

FrankYang0529 (commented):
Overall LGTM. Do we want to add validation for update VMImage operation as well? Users can change annotation value after VMImage is ready.

pkg/webhook/resources/virtualmachineimage/mutator.go
```diff
@@ -55,6 +59,44 @@ func (m *virtualMachineImageMutator) Create(_ *types.Request, newObj runtime.Obj
...
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/metadata/annotations/%s", "value": "%s"}`, patch.EscapeJSONPointer(util.AnnotationStorageClassName), targetSCName))
+		// set the annotation
+		vmi.Annotations[util.AnnotationStorageClassName] = targetSCName
+	}
+
+	// ensure the spec.targetStorageClassName is set and consist with the above annotation
+	if vmi.Spec.TargetStorageClassName == "" && targetSCName != "" {
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/targetStorageClassName", "value": "%s"}`, targetSCName))
+		vmi.Spec.TargetStorageClassName = targetSCName
+	}
```
WebberHuang1118: Since this logic would be shared across VMIs, I’d prefer to factor it out into [mutator.go](https://github.com/Vicente-Cheng/harvester/blob/ea769270ad3a33a2e601dcc67220acc4df20723e/pkg/image/common/mutator.go) as a common rule.

Both the [CDI](https://github.com/Vicente-Cheng/harvester/blob/ea769270ad3a33a2e601dcc67220acc4df20723e/pkg/image/cdi/mutate.go) and [BackingImage](https://github.com/Vicente-Cheng/harvester/blob/ea769270ad3a33a2e601dcc67220acc4df20723e/pkg/image/backingimage/mutate.go) backends can then reuse it, similar to how it's currently handled in [BackingImage](https://github.com/Vicente-Cheng/harvester/blob/ea769270ad3a33a2e601dcc67220acc4df20723e/pkg/image/backingimage/mutate.go#L18-L20).

This approach should better align with the current code structure.

pkg/webhook/resources/virtualmachineimage/validator.go
```diff
@@ -59,6 +62,9 @@ func (v *virtualMachineImageValidator) Resource() types.Resource {
 
 func (v *virtualMachineImageValidator) Create(request *types.Request, newObj runtime.Object) error {
 	vmi := newObj.(*v1beta1.VirtualMachineImage)
+	if vmi.Annotations[util.AnnotationStorageClassName] != vmi.Spec.TargetStorageClassName && vmi.Spec.TargetStorageClassName != "" {
+		return werror.NewInvalidError(fmt.Sprintf("The StorageClass of the VM Image (%v) should be consistent.", vmi.Name), "spec.targetStorageClassName")
+	}
```
WebberHuang1118: Ditto

WebberHuang1118 (changes requested):
Hi @Vicente-Cheng,

I have some thoughts on how to better align with the current source structure—please take a look when you have a chance. I'm happy to provide a draft implementation based on this PR if needed.

Thanks!

Vicente-Cheng:
>Overall LGTM. Do we want to add validation for update VMImage operation as well? Users can change annotation value after VMImage is ready.

Nice suggestion! Updated!

pkg/webhook/resources/virtualmachineimage/mutator.go
Vicente-Cheng: nice suggestion, updated!

pkg/webhook/resources/virtualmachineimage/validator.go
Vicente-Cheng: updated!

WebberHuang1118 (dismissed):
LGTM, thanks.

FrankYang0529 (dismissed):
LGTM. Thanks.

pkg/image/common/mutator.go
```diff
@@ -76,11 +80,12 @@ func (m *vmiMutator) getSC(scName string) (*storagev1.StorageClass, error) {
 		return nil, err
 	}
 
-	for _, storageClass := range storageClasses {
-		if storageClass.Annotations[util.AnnotationIsDefaultStorageClassName] == "true" &&
-			storageClass.Provisioner == longhorntypes.LonghornDriverName {
-			return storageClass, nil
-		}
+	defaultSC := util.GetDefaultSC(storageClasses)
```
ihcsim: Nit: consider moving the `List()` call to get `storageClasses` into `util.GetDefaultSC()` for re-use purposes. Otherwise, other callers (like `PatchDefaultSC()`) will have to repeat the same `List()` call before they can use `util.GetDefaultSC()`. `util.GetDefaultSC()` can be extended to accept a label argument.

pkg/image/common/mutator.go
```diff
@@ -109,3 +114,50 @@ func (m *vmiMutator) PatchImageSCParams(vmi *harvesterv1.VirtualMachineImage) ([
...
+			return patchOps, err
+		}
+		targetSCName = defaultSC.Name
+	}
+
+	// ensure the annotation `harvesterhci.io/storageClassName` is set
+	if vmi.Annotations[util.AnnotationStorageClassName] == "" && targetSCName != "" {
+		if vmi.Annotations == nil {
+			patchOps = append(patchOps, `{"op": "add", "path": "/metadata/annotations", "value": {}}`)
+			vmi.Annotations = map[string]string{}
```
ihcsim: Why do we need to modify the original `vmi` object?

pkg/image/common/mutator.go
```diff
@@ -109,3 +114,50 @@ func (m *vmiMutator) PatchImageSCParams(vmi *harvesterv1.VirtualMachineImage) ([
...
+		}
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/metadata/annotations/%s", "value": "%s"}`, patch.EscapeJSONPointer(util.AnnotationStorageClassName), targetSCName))
+		// set the annotation
+		vmi.Annotations[util.AnnotationStorageClassName] = targetSCName
+	}
+
+	// ensure the spec.targetStorageClassName is set and consist with the above annotation
```
ihcsim:
```suggestion
	// ensure the spec.targetStorageClassName is set and consistent with the above annotation
```

pkg/image/common/mutator.go
```diff
@@ -109,3 +114,50 @@ func (m *vmiMutator) PatchImageSCParams(vmi *harvesterv1.VirtualMachineImage) ([
 	patchOps = append(patchOps, fmt.Sprintf(`{"op": "%s", "path": "/spec/storageClassParameters", "value": %s}`, verb, string(valueBytes)))
 	return patchOps, nil
 }
+
+func (m *vmiMutator) PatchDefaultSC(vmi *harvesterv1.VirtualMachineImage) ([]string, error) {
```
ihcsim: Nit: I think this method is more like `EnsureTargetSC()`, per the inline comments at 145 and 156, because we want to make sure the target sc class is added to the annotation and the spec.

pkg/image/common/mutator.go
Vicente-Cheng: Thanks for the suggestion!

I separate the two functions `GetSCWithSelector` and `GetDefaultSC` for this case.

pkg/image/common/mutator.go
Vicente-Cheng: nice catch, we could skip the vmi update.
updated.

pkg/image/common/mutator.go
Vicente-Cheng: updated

pkg/image/common/mutator.go
Vicente-Cheng: yeah, the `EnsureTargetSC()` is more suitable. updated.

ihcsim (approved):

WebberHuang1118 (approved):

FrankYang0529 (approved):

Issue #8171, opened 2025-05-02
[CI] Add planned issued to `New Issues` column if they are added to `Harvester Sprint`


innobead:
### Task request

When a new issue is added to the Harvester Sprint, do the following:

- Add it to the `New Issues` column if it has no assignee.
- Assign it to the `Planning` milestone if it has not been assigned to a milestone.
- Move it to the `Backlog` column if it is assigned to a team member.

Yu-Jack:
Tested in these issues:

- Backlog
  - https://github.com/harvester/harvester/issues/8243
  - https://github.com/harvester/harvester/issues/8238
  - https://github.com/harvester/harvester/issues/6107
- Milestone and New Issues
  - https://github.com/harvester/harvester/issues/8239

PR #8209, opened 2025-05-06
chore: separate the update project flow [skip ci]

Yu-Jack:
#8171 

- Separate Harvester and Community Sprint Workflow
- Use actor instead of issue creator

### DIFF
--- a/.github/workflows/issue-management-update-community-projects.yaml
+++ b/.github/workflows/issue-management-update-community-projects.yaml
@@ -1,7 +1,7 @@
-name: "[Issue Management] Update Issue Status in Projects"
+name: "[Issue Management] Update Issue Status in Community Project"
 on:
   issues:
-    types: [labeled, milestoned, reopened, closed, assigned, unassigned]
+    types: [labeled, milestoned, reopened, closed]
   issue_comment:
     types: [created, edited]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/issue-management-update-community-projects.yaml
```diff
@@ -111,23 +27,14 @@ jobs:
       continue-on-error: true
       uses: rancher/gh-issue-mgr/get-user-teams-membership@main
       with:
-        username: ${{ github.event.issue.user.login }}
+        username: ${{ github.actor }}
```
Yu-Jack, line 30: It should be actor.

.github/workflows/issue-management-update-harvester-projects.yaml
```diff
@@ -0,0 +1,99 @@
...
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v4
+
+    - name: Is Harvester Member
+      id: is-harvester-member
+      continue-on-error: true
+      uses: rancher/gh-issue-mgr/get-user-teams-membership@main
+      with:
+        username: ${{ github.actor }}
```
Yu-Jack, line 27: It should be actor.

Yu-Jack (commented):
`github.actor` is the one who triggers the event. ([reference](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs))

`github.event.issue.user.login` is the creator ([reference](https://github.com/orgs/community/discussions/26029))

Take this [workflow](https://github.com/harvester/harvester/actions/runs/14852653727) as the example, the actor is Tim, but the username in workflow is issue creator.

image


cc @derekbit

innobead (approved):

Issue #8208, opened 2025-05-06
[TASK] Bump Longhorn to 1.7.3 for Harvester 1.4.3


innobead:
### Task request

As titled. 

### Additional context

cc @harvester/release @harvester/qa

lanfon72:
wait RC2 to include this change

noahgildersleeve:
Validated in 2 node libvirt/KVM cluster of Harvester running `v1.4.3-rc2`

![Image](https://github.com/user-attachments/assets/959df258-8634-48ae-b26b-f5ae01ee5868)
![Image](https://github.com/user-attachments/assets/d278c5ad-ba34-40df-9dae-754ea0e04e30)

PR #8210, opened 2025-05-06
Bump Longhorn v1.7.3

Vicente-Cheng:
#### Problem:
We need to use the latest Longhorn v1.7.x for Harvester v1.4.3

#### Solution:
Bump Longhorn v1.7.3

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8208

#### Test plan:
Ensure everything related to the Longhorn works as usual.


innobead (dismissed):

go.mod
```diff
@@ -11,6 +13,9 @@ replace (
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200521150516-05eb9880269c
 	github.com/operator-framework/operator-lifecycle-manager => github.com/operator-framework/operator-lifecycle-manager v0.0.0-20190128024246-5eb7ae5bdb7a
 
+	// keep lasso on old version to avoid the newer informer capability
```
innobead, line 16: ref: https://github.com/longhorn/longhorn/issues/10856

Dockerfile.dapper
```diff
@@ -45,7 +45,7 @@ RUN apt-get update -qq && apt-get install -y --no-install-recommends \
     docker-ce=5:20.10.* \
     && rm -rf /var/lib/apt/lists/*
 ## install golangci
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.57.1
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.64.8
```
innobead, line 48: Suggest using the latest version always, as it should be backward compatible. It doesn't support forward compatibility, so this is why sometimes we might encounter an incompatibility issue with golang-ci. 

Can handle this in another PR. 

ref: https://golangci-lint.run/welcome/faq/
> Basically, golangci-lint supports Go versions lower or equal to the Go version used to compile it.

innobead (approved):

Dockerfile.dapper
Vicente-Cheng: Yeah, I tried v2 golangci-lint (v2.1.6), and it looks like we need to put extra effort into using it. (Like golintci yaml, and many many lint warnings)

So, I thought we could do that on the master branch first.

Issue #8195, opened 2025-05-05
[ENHANCEMENT] Upgrade controller needs to resume those paused managedcharts


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

The `upgarde_manifest` pause all managedcharts, and if the ugprade CR is deleted due to some other reasons, those managedcharts are not resumed accordingly.

```
...
wait_repo
detect_repo
detect_upgrade
pre_upgrade_manifest
pause_all_charts
skip_restart_rancher_system_agent
upgrade_rancher
...
```

https://github.com/harvester/harvester/blob/44b1d5b33173c142d18050bc517374b5f8121570/package/upgrade/upgrade_manifests.sh#L1350

https://github.com/harvester/harvester/blob/44b1d5b33173c142d18050bc517374b5f8121570/pkg/controller/master/upgrade/upgrade_controller.go#L408

https://github.com/harvester/harvester/blob/44b1d5b33173c142d18050bc517374b5f8121570/pkg/controller/master/upgrade/upgrade_controller.go#L437

### Describe the solution you'd like

1. The upgrade controller resumes them.
2. Update the document https://docs.harvesterhci.io/v1.3/upgrade/troubleshooting/#start-over-an-upgrade

A workaround for referrence.

### Additional context

In same context with https://github.com/harvester/harvester/issues/8179

m-ildefons:
# Validation

#### Scenario

Upgrade v1.5.1 --> v1.6.0-rc5, aborted mid-upgrade.
The managed charts are paused during the upgrade. To ensure reliable recovery from the upgrade process, the managed charts must be unpaused when the the upgrade is terminated.

#### Expectation

Regardless of when the upgrade is terminated, and regardless of whether the upgrade is terminated regularly (i.e. completed normally) or irregularly (i.e. aborted), the managed charts must be unpaused again.

## Setup (all test cases)

- Single node Harvester v1.5.1 in ipxe-examples environment
- Add upgrade resource

```
apiVersion: harvesterhci.io/v1beta1
kind: Version
metadata:
  name: v1.6.0-rc5
  namespace: harvester-system
spec:
  isoChecksum: '66aab24212d1dcbb319457350b34ad787baeeeceb471541d277cf27033fa04f8c0a60948076e06e0926fd555f98a472c4477eadd8f0f4670c88636999161ca41'
[...]

PR #8220, opened 2025-05-07
Resume managedcharts when upgrade is finished or stopped

w13915984028:
#### Problem:

Upgrade script puased managedchart, and if the upgrade is stopped in some stages, the managedchart is not resumed.

#### Solution:

Resume them when ugprade is stopped or finished.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8195

#### Test plan:

Only when the current Harvester POD image includes the fix, it can resume all managedcharts. The v1.4.3 cluster does not have the fix.

```
                                                         abort the upgrade
                               before Harvester POD is upgraded          after Harvester POD is upgraded
                               
v1.4.2 -> v1.4.3          managedcharts might not be fully resumed          might not be fully resumed                   
v1.4.3 -> v1.5.1                        might not be fully resumed              fully resumed
v1.5.1 -> v1.6.0                            fully resumed                       fully resumed
```

Normal upgrade:
1. Upgrade from a cluster with this patch to a new ISO with this patch
1. Wait unit it is finished, check below managedcharts, if their `paused` are not existing or `false` (expected value)

Abnormal upgrade:
1. Start the upgrade, check below managedcharts, if they are paused, then delete the upgrade CR to stop the upgrade ( note: stop on this stage may cause other issue;  this PR targets to recover managedchart if the upgrade is aborted in the middle, other issues are not in the scope of this PR);   if deletion of the upgrade CR is blocked by validator, then wait and retry to find another chance.
1. wait some time,  check below managedcharts, if their `paused` are not existing or `false` (expected value)

The target managedchart:

"harvester harvester-crd rancher-monitoring-crd rancher-logging-crd"

Local test log:
1. Trigger an master-head to master-head upgrade
2. Delete the upgrade.harvesterhci in between
3. Observed following log, the still paused charts are resumed
```
time="2025-05-12T19:24:43Z" level=info msg="Stop collecting logs"
time="2025-05-12T19:24:43Z" level=info msg="Tearing down the logging infrastructure for upgrade procedure"
time="2025-05-12T19:24:43Z" level=info msg="managedchart rancher-monitoring-crd is resumed"
time="2025-05-12T19:24:43Z" level=info msg="managedchart rancher-logging-crd is resumed"
```

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -12,6 +12,7 @@ import (
 	semverv3 "github.com/Masterminds/semver/v3"
 	provisioningv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
+	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	provisioningctrl "github.com/rancher/rancher/pkg/generated/controllers/provisioning.cattle.io/v1"
 	"github.com/rancher/wrangler/v3/pkg/condition"
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/batch/v1"
@@ -101,6 +102,9 @@ type upgradeHandler struct {
 	versionCache      ctlharvesterv1.VersionCache
 	planClient        upgradectlv1.PlanClient
 	planCache         upgradectlv1.PlanCache
+
+	managedChartCache  mgmtv3.ManagedChartCache
+	managedChartClient mgmtv3.ManagedChartClient
 
 	vmImageClient ctlharvesterv1.VirtualMachineImageClient
 	vmImageCache  ctlharvesterv1.VirtualMachineImageCache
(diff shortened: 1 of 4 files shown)


pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -527,6 +531,33 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
...
+
+	for _, managedChart := range managedCharts {
+		if !managedChart.Spec.Paused {
+			continue
+		}
+		if _, ok := targetManagedcharts[managedChart.Name]; !ok {
+			continue
+		}
+		mc := managedChart.DeepCopy()
+		mc.Spec.Paused = true
```
FrankYang0529: I think this is `false`? So we can resume managed charts?

w13915984028: super, it definitely should be `false` here

FrankYang0529 (approved):
LGTM. Thanks for the PR.

starbops (approved):
LGTM, thank you!

w13915984028:
@mergify backport v1.5

Issue #8179, opened 2025-05-02
[ENHANCEMENT] Upgrade pre-flight checks the matching between node and machine objects


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Very rarely, the node is installed, but the related machine object is not created, the mis-matching may cause a couple of upgrade issues.

e.g., upgrade blocks `"Plan $plan_name completes."`
 https://github.com/harvester/harvester/blob/44b1d5b33173c142d18050bc517374b5f8121570/package/upgrade/upgrade_manifests.sh#L1112

### Describe the solution you'd like

The upgrade pre-flight check works upon this case.

### Describe alternatives you've considered

Review the waiting upon `plan` object.

### Additional context

Refer https://github.com/w13915984028/harvester-develop-summary/blob/main/one-v132-v140-upgrade-issue-analysis.md#why-upgrade-is-stucking-on-waiting-for-plan

albinsun:
Test PASS, close as fixed.


## Environment
* `harvester-v1.6.0-rc2`
  * Profile: 2 nodes AMD64 QEMU/KVM
  * ui-source: `Auto`


## Test Scenario
1. <details>
   <summary> Create a 2 nodes cluster
   </summary>

   <img width="1598" height="807" alt="Image" src="https://github.com/user-attachments/assets/23a7cdf6-93de-449d-812b-f7ed50bec270" />
   </details>

1. <details>
   <summary> Create a fake machine for testing
   </summary>

   machine.yaml
    ```
    apiVersion: cluster.x-k8s.io/v1beta1
    kind: Machine
    metadata:
      labels:
        cluster.x-k8s.io/cluster-name: local
        cluster.x-k8s.io/control-plane: "false"
        rke.cattle.io/cluster-name: local
        rke.cattle.io/control-plane-role: "false"
        rke.cattle.io/etcd-role: "false"
        rke.cattle.io/worker-role: "true"
      name: custom-4d279419a084
      namespace: fleet-local
    spec:
      bootstrap:
        configRef:
          apiVersion: rke.cattle.io/v1
          kind: RKEBootstrap
          name: custom-4d279419a084
          namespace: fleet-local
        dataSecretName: custom-4d279419a084-machine-bootstrap
      clusterName: local
      infrastructureRef:
        apiVersion: rke.cattle.io/v1
        kind: CustomMachine
        name: custom-4d279419a084
        namespace: fleet-local
[...]

PR #8222, opened 2025-05-07
Check node and machine are strictly matching before upgrade

w13915984028:
#### Problem:

When a cluster has incorrectly provisioned nodes/machines, the upgrade will encounter issues.

#### Solution:

Check all nodes and machines are well matched and referring to each other.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8179

#### Test plan:

1. It is not easy to produce the expected mismatching of node and machine on real clusters, the test code is added to ensure the check covers all cases.

2. Create a fake machine, trigger the upgrade, the webhook error:

image

https://github.com/harvester/harvester/issues/8195

### DIFF
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/utils/pointer"
 
 	"github.com/harvester/harvester/pkg/config"
+	"github.com/harvester/harvester/pkg/util"
 	utilHelm "github.com/harvester/harvester/pkg/util/helm"
 )
 
@@ -44,10 +45,9 @@ const (
 	HarvesterMgmtNodeLabelKey    = HarvesterNodeRoleLabelPrefix + "management"
 	HarvesterWorkerNodeLabelKey  = HarvesterNodeRoleLabelPrefix + "worker"
 
-	HarvesterLabelAnnotationPrefix      = "harvesterhci.io/"
-	HarvesterManagedNodeLabelKey        = HarvesterLabelAnnotationPrefix + "managed"
-	HarvesterPromoteNodeLabelKey        = HarvesterLabelAnnotationPrefix + "promote-node"
-	HarvesterPromoteStatusAnnotationKey = HarvesterLabelAnnotationPrefix + "promote-status"
+	HarvesterManagedNodeLabelKey        = util.HarvesterManagedNodeLabelKey
+	HarvesterPromoteNodeLabelKey        = util.HarvesterPromoteNodeLabelKey
+	HarvesterPromoteStatusAnnotationKey = util.HarvesterPromoteStatusAnnotationKey
 
 	PromoteStatusComplete = "complete"
 	PromoteStatusRunning  = "running"
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -49,7 +49,7 @@ const (
 	upgradeServiceAccount          = "system-upgrade-controller"
 	harvesterSystemNamespace       = util.HarvesterSystemNamespaceName
 	harvesterUpgradeLabel          = "harvesterhci.io/upgrade"
-	harvesterManagedLabel          = "harvesterhci.io/managed"
+	harvesterManagedLabel          = util.HarvesterManagedNodeLabelKey
 	harvesterLatestUpgradeLabel    = "harvesterhci.io/latestUpgrade"
 	harvesterUpgradeComponentLabel = "harvesterhci.io/upgradeComponent"
 	harvesterNodeLabel             = "harvesterhci.io/node"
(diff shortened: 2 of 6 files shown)


w13915984028:
@starbops @FrankYang0529 @votdev @m-ildefons Please help review, thanks.

pkg/webhook/resources/upgrade/validator_test.go
```diff
@@ -0,0 +1,387 @@
...
+func Test_isNodeMachineMatching(t *testing.T) {
+	tests := []struct {
+		name        string
+		nodes       []*corev1.Node
+		machines    []*clusterv1.Machine
+		expectError bool
+		errorKey    string
+	}{
+		{
+			name: "count mismatch",
```
m-ildefons: Good tests :+1: 

Can you add a tests for the case when there is a machine but there is no node?

pkg/controller/master/node/promote_controller.go
```diff
@@ -44,10 +45,9 @@ const (
 	HarvesterMgmtNodeLabelKey    = HarvesterNodeRoleLabelPrefix + "management"
 	HarvesterWorkerNodeLabelKey  = HarvesterNodeRoleLabelPrefix + "worker"
 
-	HarvesterLabelAnnotationPrefix      = "harvesterhci.io/"
-	HarvesterManagedNodeLabelKey        = HarvesterLabelAnnotationPrefix + "managed"
-	HarvesterPromoteNodeLabelKey        = HarvesterLabelAnnotationPrefix + "promote-node"
-	HarvesterPromoteStatusAnnotationKey = HarvesterLabelAnnotationPrefix + "promote-status"
+	HarvesterManagedNodeLabelKey        = util.HarvesterManagedNodeLabelKey
+	HarvesterPromoteNodeLabelKey        = util.HarvesterPromoteNodeLabelKey
+	HarvesterPromoteStatusAnnotationKey = util.HarvesterPromoteStatusAnnotationKey
```
m-ildefons, line 50: I don't see a point in redefining these constants. Can't we just use the `util.*` constants?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -49,7 +49,7 @@ const (
 	upgradeServiceAccount          = "system-upgrade-controller"
 	harvesterSystemNamespace       = util.HarvesterSystemNamespaceName
 	harvesterUpgradeLabel          = "harvesterhci.io/upgrade"
-	harvesterManagedLabel          = "harvesterhci.io/managed"
+	harvesterManagedLabel          = util.HarvesterManagedNodeLabelKey
```
m-ildefons, line 52: Same as above. Do we have to redefine these constants? Can't we just use the `util.*` constants?

m-ildefons (commented):
Generally LGTM.
Couple of small questions/comments.

pkg/controller/master/upgrade/upgrade_controller.go
w13915984028: I did not intend to change too many existing codes; in the past, consts are defined here and there, instead of a public file.

pkg/webhook/resources/upgrade/validator_test.go
w13915984028: It is checked on this line https://github.com/harvester/harvester/pull/8222/files#diff-b31f14ddefb333104c8045fe30364dd59957a3be3cccc8720dc5ea517665782eR387, and out of the none-fake-client related codes; 

ideally, we have a full test code which has all kinds of fake clients and can cover all cases

in this PR, it goes a short path, just test `isNodeMachineMatching`, and `isNodeMachineMatching` has the precondition that `node` has at least 1 node

m-ildefons: I think unit-testing `isNodeMachingMatching` on its own without fake clients is fine.

`isNodeMachineMatching` should return an error when there are more machines than nodes, since it checks that the count is equal in line 400. But the tests only cover the case where there are more nodes than machines.
The tests should make sure that any future modification to the code keeps the behavior that `isNodeMachineMatching` returns an error if the lengths of the two lists aren't equal, regardless which one is longer. Therefore both cases should be tested.

I'd move the `== 0` check from line 387 into `isNodeMachineMatching` and add another check testing `if len(machines) == 0`.
This way, we have a nice separation of concerns between the two functions. `checkNodeMachineMatching` is concerned with collecting the necessary data.
`isNodeMachineMatching` is only concerned with analyzing the two lists of objects for inconsistencies and can be unit tested without the need for fake clients and such. Since it's the only place that implements any actual logic for the validator, it makes testing the correct behavior of the validator in all cases fairly straightforward.

FrankYang0529 (dismissed):
LGTM.

pkg/webhook/resources/upgrade/validator_test.go
w13915984028: It is a good idea to make `isNodeMachineMatching` even independent. I have updated the code, please take a new look, thanks.

starbops (approved):
LGTM, thank you.

FrankYang0529 (approved):

w13915984028:
@mergify backport v1.5

Issue #6030, opened 2024-06-18
[BUG] Lost `status` field after Virtual Machine image update


torchiaf:
**Describe the bug**

I changed the `spec.displayName` field of a VM image. The field has been correctly updated but now the `status` field is missing from Image's object:

![image](https://github.com/harvester/harvester/assets/26394656/c9571945-9f36-4cdd-b64f-c7d03f11a554)

This is causing side effects in the UI, for instance

- The `Progress` column is showing an empty bar / percentage 0.
- The image is no more selectable in VM creation page, since the UI is not able to calculate the image status.

**To Reproduce**
1. Create a new Image, for example: Ubuntu, `https://releases.ubuntu.com/24.04/ubuntu-24.04-desktop-amd64.iso` ...
2. Go to Images tab
3. Select the new image, click Edit Config -> Edit YAML -> change the `spec.displayName` field.

**Expected behavior**

The `status` field should exist.

**Support bundle**

**Environment**
 - Harvester v1.3.0

**Additional context**

irishgordo:
Thanks for the fix @FrankYang0529 @torchiaf 
👍 😄 
I'll go ahead and close this out
This looks good on Harvester Version: v1.6-9d264654-head

![Image](https://github.com/user-attachments/assets/607044df-a3e9-4195-9d40-077be54efc3f)
![Image](https://github.com/user-attachments/assets/f7493557-758e-4336-ad3c-195e6a4caeb4)

PR #8278, opened 2025-05-14
feat: reject update VMI disply name request

FrankYang0529:
#### Problem:
In UI, we don't allow user to update VMIMage display name. However, users can do it via kubectl.

#### Solution:
The webhook should deny the update request.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/6030

#### Test plan:
1. Create a VMIMage.
2. Use kubectl to update `.spec.displayName`. The request is rejected.


WebberHuang1118 (approved):
LGTM, thanks.

Issue #8196, opened 2025-05-05
[FEATURE] PSA support for harvester


ibrokethecloud:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Add support for PSA in Harvester.

https://kubernetes.io/docs/concepts/security/pod-security-admission/

As part of PSA support we need to ensure
* Harvester specific system namespaced workloads are not impacted
* Upgrade from non PSA to PSA enabled images is not impacted by the change

ihcsim:
While Harvester can define PSS and enforce PSA on its system namespaces, K8s, by default, still allows users to create unenforced namespaces. Users have to decide what level of PSS is appropriate for their namespaces, which may differ based on the nature of the workloads.

For the system namespaces, I think the validating/mutating admission policies will be suitable for enforcing the PSS requirements.

megabreit:
What about Kubewarden, OPA Gatekeeper or Kyverno for maximum flexibility?

ihcsim:
Those are all great options. At this point, there is no plan to bundle them with Harvester, to avoid bloating the upgrade path. It gets complicated when users start to consume their CRD and API directly. Nothing stopping users from running them directly on Harvester though. E.g., I know one can use Kyverno to auto-assign PSS labels to their custom namespaces.

The validating/mutating admission policies are built into the k8s core, under the `admissionregistration.k8s.io/v1` api group. It should be able to satisfy our requirements for system namespaces.

ibrokethecloud:
A possible solution would be to have a combination of a controller and webhook in harvester which is driven of settings.

We could introduce the following settings:
* `enable-pod-security-standards`: a boolean option which enables / disables the pss controller within harvester. By default this will be set to `false` which will ensure existing workloads / clusters are not impacted by upgrade. When enabled the pss controller will apply the `baseline` pod security standard to all namespaces in the cluster. In addition it will watch new namespace creation and apply the `baseline` pss if enabled. The actual application of `pss` will be done by applying the `pod-security.kubernetes.io/enforce` label to all namespaces. When toggled off, is disabled after enabling the controller will remove the pss labels from the namespaces

* `privileged-pod-security-standards-namespace-list`: will allow cluster admins to override namespaces where privileged workloads etc are needed by applying `privileged` pss policy. The pss controller will watch this setting and apply the override pss level to all impacted namespaces.

[...]

PR #8302, opened 2025-05-16
introduce pss setting

ibrokethecloud:
PR introduces a new setting and associated handler to reconcile baseline setting against all non whitelisted namespaces on the cluster

https://github.com/harvester/harvester/issues/8303

#### Problem:

Harvester embedded RKE2 applies the default `privileged` pss on the underlying k8s cluster. 

This works for most users, however users wishing to leverage PSS for better workload security may need a more restrictive PSS such as baseline / restricted.

At present Harvester can only support `baseline` standard, and we should be able to support `restricted` in the future.

enforcing `baseline` standard as the default for all installations may result in issues for users who wish to have more relaxed policies to suit their use case.

#### Solution:

PR introduces a new setting `cluster-pod-security-standard` which allows a user to enable / disable the `baseline` pss at the cluster level.

Users can override the default `{"enabled":false,"whitelistedNamespacesList":""}` by changing the setting and also providing a comma separated list of namespaces in `whitelistedNamespacesList`. 

The controller ensures the correct pss enforcement level is applied to all existing and future namespaces, while allowing cluster admins to have whitelist namespaces.

The default behaviour of the setting would be to do nothing, which allows cluster admins to enable/disable the setting based on their use case,

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8303
https://github.com/harvester/harvester/issues/8196


pkg/webhook/resources/namespace/validator.go
```diff
@@ -116,3 +138,81 @@ func (v *namespaceValidator) checkIfNewResourceQuotaIsSufficient(rq *corev1.Reso
...
+	for _, key := range []string{util.PSSEnforcementKey, util.PSSEnforcementVersionKey, util.HarvesterManagedPSSKey} {
+		val, ok := newNamespace.Labels[key]
+		if ok {
+			oldNamespaceCopy.Labels[key] = val
+		}
+
+	}
+
+	// check if labels have changed from oldNamespace by running a reflect
+	if !reflect.DeepEqual(oldNamespaceCopy.Labels, oldNamespace.Labels) && req.Username() != harvesterSAName {
```
w13915984028: we could check `req.Username() == harvesterSAName` on L159 to return directly

pkg/webhook/resources/namespace/validator.go
```diff
@@ -116,3 +138,81 @@ func (v *namespaceValidator) checkIfNewResourceQuotaIsSufficient(rq *corev1.Reso
...
+		return nil
+	}
+
+	oldNamespaceCopy := oldNamespace.DeepCopy()
+	if oldNamespaceCopy.Labels == nil {
+		oldNamespaceCopy.Labels = make(map[string]string)
+	}
+	// verify pss label is not being changed
+	for _, key := range []string{util.PSSEnforcementKey, util.PSSEnforcementVersionKey, util.HarvesterManagedPSSKey} {
+		val, ok := newNamespace.Labels[key]
```
w13915984028: seems this could be simplified with
```
for key
if newNamespace.Labels[key] != oldNamespace.Labels[key] 
changed = true
break
````

oldNamespaceCopy and deepequal are not required

File: pkg/controller/master/setting/cluster_pod_security_standard.go [hunk 3]
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return fmt.Errorf("error listing namespaces: %v", err)
+	}
+
+	for _, ns := range nsList {
+		nsCopy := ns.DeepCopy()
+		delete(nsCopy.Labels, util.PSSEnforcementKey)
+		delete(nsCopy.Labels, util.PSSEnforcementVersionKey)
+		delete(nsCopy.Labels, util.HarvesterManagedPSSKey)
+		if !reflect.DeepEqual(nsCopy.Labels, ns.Labels) {
```
w13915984028: above list selector has ensured there must be wanted labels on objects, the DeepEqual could be skipped

pkg/controller/master/setting/cluster_pod_security_standard.go
```diff
@@ -0,0 +1,143 @@
...
+	// no further changes are needed
+	if reflect.DeepEqual(ns, nsCopy) {
+		return nil
+	}
+	_, err := h.namespaces.Update(nsCopy)
+	return err
+}
+
+// enqueue ClusterPodSecurityStandardSetting if there is a change to a namespace
+func (h *Handler) watchNamespaceChanges(_, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
```
w13915984028, line 85: the relatedResource seems to run every seconds, it is not like direct controller onChange

there could be a small delay to get the expected label applied or removed, but also do not matter too much

w13915984028 (commented):
Thanks for the PR, some questions.

pkg/util/constants.go
```diff
@@ -286,4 +286,14 @@ const (
...
 
 	HarvesterUpgradeComponentRepo = "repo"
+
+	// PSS controller labels
+	PSSEnforcementKey          = "pod-security.kubernetes.io/enforce"
+	BaselineEnforcementValue   = "baseline"
+	PSSEnforcementVersionKey   = "pod-security.kubernetes.io/enforce-version"
+	PSSEnforcementVersionValue = "latest"
+	RestrictedEnforcementValue = "restricted"
+	PrivilegedEnforcementValue = "privileged"
```
ihcsim: to avoid src drifts, can we point these constants to the upstream const [labels](https://pkg.go.dev/k8s.io/pod-security-admission@v0.34.0/api#pkg-constants) and [levels](https://pkg.go.dev/k8s.io/pod-security-admission@v0.34.0/api#Level)?

pkg/controller/master/setting/cluster_pod_security_standard_test.go
```diff
@@ -0,0 +1,174 @@
...
+	assert.NoError(err)
+	_, ok = nsObj.Labels[util.PSSEnforcementKey]
+	assert.False(ok, "expected to not find enforcement key")
+
+	// check restricted pss is applied
+	restrictedNSObj, err := h.namespaces.Get(restrictedNS.Name, metav1.GetOptions{})
+	assert.NoError(err)
+	assert.Equal(util.RestrictedEnforcementValue, restrictedNSObj.Labels[util.PSSEnforcementKey], "expected to find restricted enforcement key")
+
+	// check restricted pss is applied
```
ihcsim:
```suggestion
	// check privileged pss is applied
```

pkg/controller/master/setting/cluster_pod_security_standard_test.go
```diff
@@ -0,0 +1,174 @@
...
+	_, ok = nsObj.Labels[util.PSSEnforcementKey]
+	assert.False(ok, "expected to not find enforcement key")
+
+	// check restricted pss is skipped
+	restrictedNSObj, err := h.namespaces.Get(restrictedNS.Name, metav1.GetOptions{})
+	assert.NoError(err)
+	_, ok = restrictedNSObj.Labels[util.PSSEnforcementKey]
+	assert.False(ok, "expected to not find enforcement key")
+
+	// check restricted pss is skipped
```
ihcsim:
```suggestion
	// check privileged pss is skipped
```

pkg/controller/master/setting/cluster_pod_security_standard_test.go
```diff
@@ -0,0 +1,174 @@
...
+	_, ok = privilegedNSObj.Labels[util.PSSEnforcementKey]
+	assert.False(ok, "expected to not find enforcement key")
+
+	// check existing baseline policy is removed
+	nsObj, err = h.namespaces.Get(existingManagedPSSNS.Name, metav1.GetOptions{})
+	assert.NoError(err)
+	_, ok = nsObj.Labels[util.PSSEnforcementKey]
+	assert.False(ok, "expected to not find enforcement key")
+	_, ok = nsObj.Labels[util.PSSEnforcementKey]
+	assert.False(ok, "expected to not find enforcement key")
```
ihcsim: this assertion look redundant. did you mean to assert `util.HarvesterManagedPSSKey`?

pkg/webhook/resources/namespace/validator.go
```diff
@@ -116,3 +138,81 @@ func (v *namespaceValidator) checkIfNewResourceQuotaIsSufficient(rq *corev1.Reso
...
+	if slices.Contains(whitelistedNamespacesList, oldNamespace.Name) {
+		return nil
+	}
+
+	oldNamespaceCopy := oldNamespace.DeepCopy()
+	if oldNamespaceCopy.Labels == nil {
+		oldNamespaceCopy.Labels = make(map[string]string)
+	}
+	// verify pss label is not being changed
+	for _, key := range []string{util.PSSEnforcementKey, util.PSSEnforcementVersionKey, util.HarvesterManagedPSSKey} {
```
ihcsim: This doesn't stop a non-`harvesterSAName` user from adding a PSA label to reference older versions of the `audit`, `warn` policies. I think it's ok but I am not sure...

ihcsim: If we want to ensure complete immutability, upstream provides the [`api.PolicyToEvaluate`](https://pkg.go.dev/k8s.io/pod-security-admission/api#PolicyToEvaluate) and [`api.Equivalent`](https://pkg.go.dev/k8s.io/pod-security-admission/api#Policy.Equivalent) APIs, where it can assess a set of labels for PSA settings, and ignore non-existing ones accordingly.

pkg/webhook/resources/namespace/validator.go
ibrokethecloud: users may want to change other labels on the namespace, and those should be allowed. It is only the PSS ones which we do not allow to change.

pkg/controller/master/setting/cluster_pod_security_standard_test.go
ibrokethecloud: it was meant to check version key

pkg/webhook/resources/namespace/validator.go
ibrokethecloud: i could add this, but technically only enforcement mode and enforcement level are needed. based on a quick test if a user adds additional modes or levels for say audit/ warning ,the force one still takes effect

to test

```
kubectl create ns demo
kubectl label ns demo pod-security.kubernetes.io/enforce=restricted
kubectl label ns demo pod-security.kubernetes.io/enforce-version=latest
kubectl label ns demo pod-security.kubernetes.io/audit=restricted
kubectl label ns demo pod-security.kubernetes.io/audit-version=latest
```

attempt to create privileged pod still fails
```
apiVersion: v1
kind: Pod
metadata:
  name: privileged-pod-sample
spec:
  containers:
  - name: privileged-container
    image: busybox:stable
    command: ["sh", "-c", "while true; do sleep 3600; done"]
    securityContext:
      privileged: true
```

```
kubectl apply -f /tmp/pod.yaml -n demo
Error from server (Forbidden): error when creating "/tmp/pod.yaml": pods "privileged-pod-sample" is forbidden: violates PodSecurity "restricted:latest": privileged (container "privileged-container" must not set securityContext.privileged=true), allowPrivilegeEscalation != false (container "privileged-container" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "privileged-container" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "privileged-container" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "privileged-container" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

pkg/controller/master/setting/cluster_pod_security_standard.go
ibrokethecloud: the webhook will block any manual attempts to change this as well.

pkg/settings/settings_helper.go
```diff
@@ -280,3 +282,24 @@ func ValidateAdditionalGuestMemoryOverheadRatioHelper(value string) error {
...
+func GetPodSecuritySetting(setting *harvesterv1.Setting) (*PodSecuritySetting, error) {
+	pssSetting := &PodSecuritySetting{}
+	var value string
+	if setting.Value != "" {
+		value = setting.Value
+	} else {
+		value = setting.Default
+	}
+	if err := json.Unmarshal([]byte(value), pssSetting); err != nil {
+		return nil, fmt.Errorf("invalid JSON `%s`: %s", setting.Value, err.Error())
```
w13915984028: copilot is correct, use `value` to fmt

pkg/webhook/resources/namespace/validator.go
```diff
@@ -116,3 +138,87 @@ func (v *namespaceValidator) checkIfNewResourceQuotaIsSufficient(rq *corev1.Reso
...
+
+	return nil
+}
+
+// Create ensures user cannot apply PSS label during namespace creation there by ignoring the underlying cluster level
+// settings
+func (v *namespaceValidator) Create(req *types.Request, newObj runtime.Object) error {
+	newNamespace := newObj.(*corev1.Namespace)
+
+	settingObj, err := v.settingsCache.Get(settings.ClusterPodSecurityStandardSettingName)
```
w13915984028, line 208: The `IsNotFound` scenario needs to be considered, when a new cluster is installed, when setting is built and namespace is created in parallel, it might hit this error

pkg/controller/master/setting/cluster_pod_security_standard.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return err
+	}
+	// setting is not enabled, no further processing needed
+	if !pssSetting.Enabled {
+		return h.cleanupManagedPSSLabels()
+	}
+
+	// split comma separated whitelist
+	additionalNamespaces := strings.Split(pssSetting.WhitelistedNamespacesList, ",")
```
w13915984028: The suggestion from Copliot also makes sense
>The comma-separated namespace lists are split but not trimmed or filtered.

pkg/controller/master/setting/cluster_pod_security_standard.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// enqueue ClusterPodSecurityStandardSetting if there is a change to a namespace
+func (h *Handler) watchNamespaceChanges(_, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
+	if ns, ok := obj.(*corev1.Namespace); ok {
+		// ignore namespaces being deleted
+		if !ns.DeletionTimestamp.IsZero() {
+			return nil, nil
+		}
+		return []relatedresource.Key{
```
w13915984028, line 91: there is no namespace controller/mutator to apply the related pss setting, with this relatedresource watcher, 

>any namespace change -> wakeup pss-setting -> loop all namepsaces

when a system has many namespaces, the above logic will have some performance loss

when pss is enabled, a newly created namespace will wake up pss-setting  2+ times

w13915984028 (commented):
some last comments, thanks.

pkg/webhook/resources/namespace/validator.go
ibrokethecloud: if setting errors, it will just requeue until the cache is initialised.

pkg/webhook/resources/namespace/validator.go
w13915984028: guess we allow `harvesterSAName` to change any label, and non-`harvesterSAName` can't change pss related

put user check on L159 will let happy path run asap.

```
L159:

if req.Username() == harvesterSAName {
	return nil
}


L181:

if !reflect.DeepEqual(oldNamespaceCopy.Labels, oldNamespace.Labels) {
	return fmt.Errorf("....can't change pss")
}
```

pkg/webhook/resources/namespace/validator.go
w13915984028: user could also delete a setting, we did not use webhook to block, the missing setting is a corner case

ibrokethecloud: yeah, but it can only be done a cluster admin, and restarting the harvester pod will re-initialise the setting.
The aim is to block non cluster admin users from running privileged workloads on a cluster.

ihcsim: Can we log a warning instead of blocking the namespace creation/update? Given how fundamental namespaces are for all the major use cases, I don't think any errors (except for the request username violation in line 217) from an optional setting should block namespace creation/update.

pkg/webhook/resources/namespace/validator.go
ihcsim: My concern is that what if the `audit-version` and `warn-version` are downgraded to some obsolete version, to either loosen the standard definition or trigger some obsolete code path. While the `enforce` policy is still in-effect, the `audit` and `warn` policies are also still evaluated based on the version provided. Can we just outright reject any `warn` and `audit` changes submitted by a non-`harvesterSAName` user and let the controller handle everything?

pkg/webhook/resources/namespace/validator.go
w13915984028: we could consider to fallback the `IsNotFound` error to `pss-setting disabled`, do nothing when creating NS. When the setting is back/reconcilered, all NSs have chance to be handled again

the potential cluster-init conflicts could be skipped and if user deletes this setting by intention, webhook & controller tolerants it and assumes pss-setting is disabled

ibrokethecloud:
I have made the following changes in addition to the existing ones
* add a validating webhook handler to ensure setting for `cluster-pod-security-standard` cannot be deleted
* modified logic in namespace validation, to use the predefiend harvester default namespace whitelist, before looking up the setting, this should in case of cache lookup failures at start, allow the system workloads to initialise successfully. 
* there are minor changes in the namespace validation webhook to simplify validation logic and include feedback provided
* we also block requests if user is trying to define audit/warning pss levels on the `cluster-pod-security-standard` managed namespaces, to avoid any conflicts on unintended access

w13915984028 (approved):
LGTM, thanks.

ihcsim (approved):

Vicente-Cheng (approved):

pkg/webhook/resources/namespace/validator.go
```diff
@@ -116,3 +137,109 @@ func (v *namespaceValidator) checkIfNewResourceQuotaIsSufficient(rq *corev1.Reso
...
+		newNamespace.Labels = make(map[string]string)
+	}
+
+	// verify pss label is not being changed
+	for _, key := range []string{psaApi.EnforceLevelLabel, psaApi.EnforceVersionLabel, psaApi.AuditLevelLabel, psaApi.AuditVersionLabel, psaApi.WarnLevelLabel, psaApi.WarnVersionLabel, util.HarvesterManagedPSSKey} {
+		if newNamespace.Labels[key] != oldNamespace.Labels[key] {
+			return fmt.Errorf("%s is enabled, PSS level can only be changed by cluster admin via the setting", settings.ClusterPodSecurityStandardSettingName)
+		}
+
+	}
```
Yu-Jack, line 188: Just a Nit: Could we make the slice `[]string{psaApi ... }` as a variable?

pkg/webhook/resources/namespace/validator.go
```diff
@@ -116,3 +137,109 @@ func (v *namespaceValidator) checkIfNewResourceQuotaIsSufficient(rq *corev1.Reso
...
+	if !pssSetting.Enabled {
+		return nil
+	}
+
+	// verify if namespace is part of whiltelisted namespaces available from setting
+	if isNamespaceWhitelisted(util.GetWhitelistedNamespacesList(pssSetting), newNamespace.Name) {
+		return nil
+	}
+
+	for _, key := range []string{psaApi.EnforceLevelLabel, psaApi.EnforceVersionLabel, psaApi.AuditLevelLabel, psaApi.AuditVersionLabel, psaApi.WarnLevelLabel, psaApi.WarnVersionLabel, util.HarvesterManagedPSSKey} {
```
Yu-Jack, line 228: Same.

Yu-Jack (approved):
LGTM, thanks! Just few Nit.

PR #8329, opened 2025-05-21
docs: add component update task issue template [ci skip]

starbops:
#### Problem:

Since component updates happen regularly, it's better to have an issue template to reduce the burden.

#### Solution:

Add a new task issue template specifically for component updates.

#### Related Issue(s):

N/A

#### Test plan:

N/A

The author of the issue (typically the release captain) needs to determine which components to update for the upcoming minor/patch release of Harvester.

### DIFF
--- a/.github/ISSUE_TEMPLATE/09-component-update-task.yaml
+++ b/.github/ISSUE_TEMPLATE/09-component-update-task.yaml
@@ -1,0 +1,66 @@
+name: Task request (component update)
+description: Create a component update task targeting a specific Harvester release
+title: "[TASK] Component updates for Harvester vX.Y.Z"
+labels: [ "kind/task", "area/build-package" ]
+assignees:
+-
+body:
+- type: textarea
+  attributes:
+    label: Component update request
+    description: |
+      List the dependent components that need to be updated for a specific Harvester release.
+
+      Please note that for different Harvester releases, the component update strategies are different:
+      - For feature or minor releases, the minor version of the component should be updated.
+      - For patch releases, only the patch version of the component should be updated.
+      - For built-in components/add-ons, the version should be aligned with the Harvester version.
+    value: |
+      Components to be updated for Harvester version vX.Y.Z:
+        
+      - Main Components
+        - [ ] Embedded Rancher @FrankYang0529
+        - [ ] RKE2 @FrankYang0529
+        - [ ] Longhorn @Vicente-Cheng
+        - [ ] KubeVirt @Vicente-Cheng
+      - Libraries
+        - [ ] Go @harvester/dev
+        - [ ] K8s @harvester/dev
+      - Built-in Dependent Components
+        - [ ] * Network Controller @mingshuoqiu
+        - [ ] * Network FS Manager @Vicente-Cheng @WebberHuang1118
+        - [ ] * Node Disk Manager @tserong  @Vicente-Cheng
+        - [ ] * Node Manager @tserong
+        - [ ] * Load Balancer @starbops
+        - [ ] * Eventrouter @w13915984028
+        - [ ] CDI @Vicente-Cheng
+        - [ ] Whereabouts @rrajendran17
+        - [ ] Kube-VIP @starbops
+        - [ ] CSI Snapshotter @brandboat  (need to align with Longhorn version)
+        - [ ] snapshot-validation-webhook @brandboat
+      - Built-in Opt-In Add-Ons
+        - [ ] * VM Import Controller @votdev
+        - [ ] * PCI Device Controller @Yu-Jack
+        - [ ] * Seeder @ibrokethecloud
+        - [ ] Rancher Logging @w13915984028
+        - [ ] Rancher Monitoring @w13915984028
+        - [ ] * Nvidia Driver Runtime @ibrokethecloud
+      - User Interface Components
+        - [ ] * UI Extension @a110605
+        - [ ] * Terraform Provider @FrankYang0529
+      - Experimental Opt-In Add-Ons (OOB)
+        - [ ] vCluster @ibrokethecloud
+        - [ ] LVM CSI Driver @Vicente-Cheng
+        - [ ] VM DHCP Controller @starbops
+        
+      > [!Note]
+      > `*` means the component version (including the chart versions and/or the image tags) will align with Harvester releases.
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: Additional context
+    description: Please add any other context about the task here.
+  validations:
+    required: false


.github/ISSUE_TEMPLATE/09-component-update-task.yaml
```diff
@@ -0,0 +1,27 @@
...
+assignees:
+-
+body:
+- type: textarea
+  attributes:
+    label: Component update request
+    description: List the dependent component(s) that need to be updated for a specific Harvester version.
+    value: |
+      Components to be updated for Harvester version vx.y.z:
+      - Embedded Rancher vx.y.z
```
innobead: These are the primary underlying components. What about built-in add-ons?

Additionally, we can mention the following to differentiate the update strategy for different types of releases:
- For each feature or minor release, increment the next minor version of the dependent component.
- For patch releases, increment only the patch version of the dependent component.

For each update item, use `- [ ]`  instead of `-` to itemize tasks. For each item, include `@` followed by the corresponding owner's name to assign responsibility for those tasks. For example, RKE2 and Rancher will be Poan. The owner may change for the final execution, as there could be multiple owners for the components.

innobead (changes requested):

.github/ISSUE_TEMPLATE/09-component-update-task.yaml
innobead: cc @harvester/release

starbops: > These are the primary underlying components. What about built-in add-ons?

I'd imagine this template is for dependent components specifically outside of Harvester's control (we have little to no leverage over the release cycles of Rancher/RKE2/Longhorn/KubeVirt...). For our built-in add-ons, it's pretty straightforward since we have a strong-binding version policy that mandates those components follow the versioning of Harvester, e.g., network controller for Harvester v1.5.0 will be of version 1.5.0, so it's kind of a no-brainer and thus not require a GH issue to track, IMO.

But I get the idea, this is the place where we track all of the dependent components versions. I'll add a consolidated bullet point for all the built-in add-ons. Thank you for bringing this up.

innobead: Sounds good.

.github/ISSUE_TEMPLATE/09-component-update-task.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - For feature or minor releases, the minor version of the component should be updated.
+      - For patch releases, only the patch version of the component should be updated.
+      - For built-in components/add-ons, the version should be aligned with the Harvester version.
+    value: |
+      Components to be updated for Harvester version vx.y.z:
+      - [ ] Embedded Rancher vx.y.z @FrankYang0529
+      - [ ] RKE2 vx.y.z @FrankYang0529
+      - [ ] Longhorn vx.y.z @Vicente-Cheng
+      - [ ] KubeVirt vx.y.z @Vicente-Cheng
+      - [ ] Built-in components/add-ons (e.g., Network Controller, Node Disk Manager, VM Import Controller, etc.) vx.y.z @harvester/dev
```
innobead: If possible, let's make them more specific, for example:

> [!Note]
> `*` means the version will align with Harvester releases. 
> 
> I just noticed the versions of components aligning with the Harvester release seem a bit inconsistent. We might need to review this later, especially for experiential addons. 

- [ ] Built-in Dependent Components
  - [ ] * Network Controller @mingshuoqiu  
  - [ ] * Network FS Manager @Vicente-Cheng @WebberHuang1118 
  - [ ] * Node Disk Manager @tserong  @Vicente-Cheng 
  - [ ] * Node Manager @tserong 
  - [ ] * Load Balancer @starbops 
  - [ ] * Eventrouter @w13915984028
  - [ ] CDI @Vicente-Cheng  
  - [ ] Whereabouts @rrajendran17 
  - [ ] Kube-VIP @starbops  
  - [ ] CSI Snapshotter @brandboat  (need to align with Longhorn version)
  - [ ] snapshot-validation-webhook @brandboat 

- [ ] Built-in Opt-In Add-Ons
  - [ ] * VM Import Controller @votdev 
  - [ ] * PCI Device Controller @Yu-Jack 
  - [ ] * Seeder @ibrokethecloud 
  - [ ] Rancher Logging Controller @w13915984028 
  - [ ] Rancher Monitoring Controller @w13915984028 
  - [ ] Nvidia Driver Runtime @ibrokethecloud 

- [ ] Experimental Opt-In Add-Ons (OOB)
  - [ ] vCluster @ibrokethecloud 
  - [ ] LVM CSI Driver @Vicente-Cheng 
  - [ ] VM DHCP Controller @starbops

innobead (changes requested):

.github/ISSUE_TEMPLATE/09-component-update-task.yaml
```diff
@@ -0,0 +1,34 @@
...
+    label: Component update request
+    description: |
+      List the dependent components that need to be updated for a specific Harvester release.
+
+      Please note that for different Harvester releases, the component update strategies are different:
+      - For feature or minor releases, the minor version of the component should be updated.
+      - For patch releases, only the patch version of the component should be updated.
+      - For built-in components/add-ons, the version should be aligned with the Harvester version.
+    value: |
+      Components to be updated for Harvester version vx.y.z:
```
Vicente-Cheng: We should also notify the go/k8s version update. WDYT?

starbops:
@innobead @Vicente-Cheng I've updated the PR per the feedback. PTAL, thank you!

Vicente-Cheng (dismissed):
thanks!

ihcsim (dismissed):

.github/ISSUE_TEMPLATE/09-component-update-task.yaml
```diff
@@ -0,0 +1,63 @@
...
+        - [ ] * VM Import Controller @votdev
+        - [ ] * PCI Device Controller @Yu-Jack
+        - [ ] * Seeder @ibrokethecloud
+        - [ ] Rancher Logging @w13915984028
+        - [ ] Rancher Monitoring @w13915984028
+        - [ ] * Nvidia Driver Runtime @ibrokethecloud
+      - Experimental Opt-In Add-Ons (OOB)
+        - [ ] vCluster @ibrokethecloud
+        - [ ] LVM CSI Driver @Vicente-Cheng
+        - [ ] VM DHCP Controller @starbops
```
innobead, line 54: Please add a section for user interface components.

```
- User Interface Components
  - [ ] * UI Extension // andy
  - [ ] * Terraform provider //poan
```

innobead (changes requested):
In general, LGTM. 

We can improve this over time, as we may just add some updates to renovate. cc @FrankYang0529

innobead (approved):

Issue #8213, opened 2025-05-06
[ENHANCEMENT] Put upgrade related options together on upgrade object


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

A few annotations are added for upgarde control

Two on `version` object:

```
	SkipGarbageCollectionThreadholdCheckAnnotation = "harvesterhci.io/skipGarbageCollectionThresholdCheck"
	MinCertsExpirationInDayAnnotation              = "harvesterhci.io/minCertsExpirationInDay"
```

A few on `upgrade` object:
```
	skipWebhookAnnotation                     = "harvesterhci.io/skipWebhook"
	skipSingleReplicaDetachedVol              = "harvesterhci.io/skipSingleReplicaDetachedVol"
...
```

Those looks to be more related the `ugprade` object, not the `version` object. For those none-ui-preferred user, code the upgrade related object is also a bit complex.

### Describe the solution you'd like

1. Put all those annotations to `upgrade` object;  
2. UI provides tick box/input filed to allow user set those options clearly together with the final `upgrade` button.
3. For CaaS/integration test, set related annotations to `upgrade` object when creating it is enough.

### Additional context

When writing document PR https://github.com/harvester/docs/pull/773, those scattered options bring noticable complexicity.

https://github.com/harvester/harvester/issues/8179
https://github.com/harvester/harvester/issues/8195

w13915984028:
cc @starbops @FrankYang0529

Issue #8078, opened 2025-04-15
[FEATURE] Upgrade via ISO Upload


innobead:
Allow users to upgrade by uploading an ISO, in addition to the existing options of automatic version sync from the Harvester endpoint, or a user-created Version CR with a remote ISO endpoint. Provide a clear UX procedure for this upload-based upgrade.

starbops:
Is this similar to what #7690 described (though it did not work in v1.5.0)?

innobead:
> Is this similar to what [#7690](https://github.com/harvester/harvester/issues/7690) described (though it did not work in v1.5.0)?

Exactly, we might need to improve the current experience to make this feature an official path for users. 

cc @harvester/pm

a110605:
ui issues :
- https://github.com/harvester/harvester/issues/7690
- https://github.com/harvester/harvester/issues/8079

a110605:
Hi @brandboat, there is unknown webhook to block system upgrade via uploading an image method.
The issue is observed in harvester master and v1.5.0. 
But it's good on v1.4.3-rc4.

<img width="1489" alt="Image" src="https://github.com/user-attachments/assets/c9cb6955-288e-4155-8c29-5ab67cbfc2ed" />

a110605:
Test in ipxe environment, it's quite fast to upload a harvester image locally (within 1min).

But test against to a ECM machine, there is 408 request timeout (around 5 min) to abort the upload API (`/v1/harvester/harvesterhci.io.virtualmachineimages/harvester-system/image-zq6k5?action=upload&size=6184173568`) . Not sure it's caused Suse VPN.

@brandboat , @ibrokethecloud , do you think is it possible to fine tune the timeout from backend ? 
 
<img width="1489" alt="Image" src="https://github.com/user-attachments/assets/d1795b6e-06d7-45f9-b252-6927878f089e" />
<img width="1490" alt="Image" src="https://github.com/user-attachments/assets/ed288003-4561-4ee4-9985-0f2b0c013e04" />

brandboat:
> Hi @brandboat, there is unknown webhook to block system upgrade via uploading an image method.
The issue is observed in harvester master and v1.5.0.
But it's good on v1.4.3-rc4.

[...]

Issue #7690, opened 2025-02-25
[BUG][UI] Air-gapped OS upgrade page is not work


a110605:
**Describe the bug**
Air-gapped upgrade page is not work in v1.5.0.

<img width="1481" alt="Image" src="https://github.com/user-attachments/assets/5921c307-9ebc-4086-ab78-1e0742b4c9d7" />

**To Reproduce**
1. Enable `View in API` and `extension developer feature` in perference.
2. Go to setting page, find `server-version` and click `Upgrade`
3. The OS upgrade page should work.

**Expected behavior**
`Upload new image` and `Select extisting image` features should work as expected.

**Environment**
 - Harvester ISO version: v1.5.0
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): ipxe-example

**Additional context**

innobead:
@a110605 explained:
>  It's the bug that we haven't fixed during vue3 migration from harvester/dashboard to harvester-ui-extension.

brandboat:
This issue also involve a backend bug, and already fixed in 
master: https://github.com/harvester/harvester/pull/8333
v1.5: https://github.com/harvester/harvester/pull/8365

For other upload upgrade-iso enhancement, please refer to https://github.com/harvester/harvester/issues/8078 (targeting v1.6)

asc5543:
Test PASS, the Air-gapped OS upgrade page works

## Environment
### Lab device
* Harvester ISO version: v1.6-a0712164-head
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 1 node Baremetal

### ipxe-example
* Harvester ISO version: v1.6-9b417305-head
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 1 node with ipxe

## Test Scenario
There are three actions on Cluster Upgrade page
* Upload New Image
  * From Local (won't check Checksum)
  * From URL (will check Checksum)
* Select Existing Image
* Delete Existing Image

### Lab device
1. <details>
   <summary> ✔ Enable <b>View in API</b> and <b>extension developer feature</b> in perference.
   </summary>

   ![Image](https://github.com/user-attachments/assets/65e8afb7-16a9-4bd8-9657-ac1ab3089644)

   </details>

1. <details>
   <summary> ✔ Go to setting page, find server-version and click Upgrade
   </summary>

[...]

Issue #8226, opened 2025-05-08
[BUG] Upgrade failed with Version CR not found error


starbops:
### Describe the Bug

After the upgrade has been kicked off, it might fail with the `versions.harvesterhci.io "VERSION" not found` error. The Upgrade CR looks like this:

```
apiVersion: harvesterhci.io/v1beta1
kind: Upgrade
metadata:
  creationTimestamp: "2025-05-07T11:28:27Z"
  finalizers:
  - wrangler.cattle.io/harvester-upgrade-controller
  generateName: hvst-upgrade-
  generation: 6
  labels:
    harvesterhci.io/upgradeState: Failed
  name: hvst-upgrade-vd9bf
  namespace: harvester-system
  resourceVersion: "622429"
  uid: 3b805fea-eaf4-4afb-9c2d-34e94237059a
spec:
  image: ""
  logEnabled: true
  version: v1.5.0
status:
  conditions:
  - lastUpdateTime: "2025-05-07T20:16:12Z"
    reason: versions.harvesterhci.io "v1.5.0" not found
    status: "False"
    type: Completed
  - lastUpdateTime: "2025-05-07T20:16:12Z"
    status: "True"
    type: LogReady
  - lastUpdateTime: "2025-05-08T01:34:12Z"
    status: "True"
    type: ImageReady
  imageID: harvester-system/hvst-upgrade-vd9bf
  previousVersion: v1.4.2
  upgradeLog: hvst-upgrade-vd9bf-upgradelog
```

During the early stage of the upgrade process, the Version CR must exist to know the information of the ISO image. If the corresponding Version CR is not there at that specific time point, the upgrade fails immediately.

There is a piece of logic that manages the Version CRs, version-syncer. It will periodically check the eligibility of all the Version CRs on the cluster, including the cluster's current version and those downloaded from the upgrade responder. If there is any in-upgradable Version CR, the version-syncer deletes it. The default checking interval is an hour.

In my case, the upgrade was held off because of the Docker Hub rate limit when downloading the fluent-bit container image. After six hours, the download succeeded, and the upgrade should have been continued. However, the version-syncer had already deleted the relevant Version CR. The upgrade failed straight away.

### To Reproduce

There are many reasons why the Version CR is deleted, but the most reliable way to reproduce this is to directly delete it after the upgrade starts (enabling upgrade logging is recommended, as it can buy you some time to delete it before the upgrade process passes the checkpoint).

### Expected Behavior

[...]

PR #8333, opened 2025-05-21
Move annotations from Version to Upgrade

brandboat:
#### Problem:

Directly perform Upgrade via uploading ISO image without creating Version first case below error.
image
The root cause is we didn't check if the version object exists in ugprade/validator.go#checkCerts https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/upgrade/validator.go#L647, if we directly create upgrade object without first creating a version object, a NPE occurred, causing the upgrade creation process to fail.

#### Solution:

Now there are two validations in upgrade/validator.go rely on the version object:

- skipGarbageCollectionThresholdCheck https://docs.harvesterhci.io/v1.5/upgrade/index/#skipgarbagecollectionthresholdcheck-annotation
- minCertsExpirationInDay https://docs.harvesterhci.io/v1.5/upgrade/index/#mincertsexpirationinday-annotation

Move annotations harvesterhci.io/skipGarbageCollectionThresholdCheck and harvesterhci.io/minCertsExpirationInDay from Version CRD to Upgrade CRD could mitigate this issue

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8078
https://github.com/harvester/harvester/issues/8213
https://github.com/harvester/harvester/issues/7690

#### Test plan:
Since we move 2 annotations from Upgrade to Version object, we have to check if both 2 annotations works as expected.
Please Prepare 2+ nodes harvester cluster to perform below tests.

### minCertsExpirationInDay with value 0
1. Create a version.
```
kind: Version
metadata:
  name: master
  namespace: harvester-system
spec:
```
2. Create an upgrade
```
kind: Upgrade
metadata:
  annotations: 
    harvesterhci.io/minCertsExpirationInDay: "0"
  generateName: hvst-upgrade-
  namespace: harvester-system
spec:
  version: "master"
```
3. Expect harvester-webhook raise validation error `Error from server (BadRequest): error when creating "upgrade.yaml": admission webhook "validator.harvesterhci.io" denied the request: invalid value 0 for annotation harvesterhci.io/minCertsExpirationInDay, it should be greater than 0`

### minCertsExpirationInDay with negative value
Basically do the same thing as last test but this time add `harvesterhci.io/minCertsExpirationInDay: "-1"`, and we should get same error `Error from server (BadRequest): error when creating "upgrade.yaml": admission webhook "validator.harvesterhci.io" denied the request: invalid value -1 for annotation harvesterhci.io/minCertsExpirationInDay, it should be greater than 0`

## Test skipGarbageCollectionThresholdCheck annotation
1. Make one node disk full. Like `dd if=/dev/random of=/usr/local/test.img bs=1G count=100`
2. Create a test version.
```
kind: Version
metadata:
  name: master
  namespace: harvester-system
spec:
```
3. Create a Upgrade object
```
kind: Upgrade
metadata:
  generateName: hvst-upgrade-
  namespace: harvester-system
spec:
  version: "master"
```
4. Expect harvester-webhook raised validation error.
5. Add annotation `harvesterhci.io/skipGarbageCollectionThresholdCheck: true` to the Upgrade object.
```
kind: Upgrade
metadata:
  annotation:
    harvesterhci.io/skipGarbageCollectionThresholdCheck: true
  generateName: hvst-upgrade-
  namespace: harvester-system
spec:
  version: "master"
```
6. The upgrade should proceed.

## Test Upgrade via uploading ISO image
1. Click top-right corner > Click Preferences > Check "Enable View in API"
2. Go to settings page > Click server-version 3-vertical-dots button > Click Upgrade
3. Upload harvester ISO, and upgrade continue, no error pops up.

N/A

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -9,7 +9,6 @@ import (
 	"sync"
 	"time"
 
-	semverv3 "github.com/Masterminds/semver/v3"
 	provisioningv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
@@ -673,37 +672,6 @@ func getCachedRepoInfo(upgrade *harvesterv1.Upgrade) (*repoinfo.RepoInfo, error)
 		return nil, err
 	}
 	return repoInfo, nil
-}
-
-func isVersionUpgradable(currentVersion, minUpgradableVersion string) error {
-	if minUpgradableVersion == "" {
-		logrus.Debug("No minimum upgradable version specified, continue the upgrading")
-		return nil
-	}
-
-	// short-circuit the equal cases as the library doesn't support the hack applied below
-	if currentVersion == minUpgradableVersion {
-		logrus.Debug("Upgrade from the exact same version as the minimum requirement")
-		return nil
-	}
-	// to enable comparisons against prerelease versions
-	constraint := fmt.Sprintf(">= %s-z", minUpgradableVersion)
-
-	c, err := semverv3.NewConstraint(constraint)
-	if err != nil {
-		return err
-	}
-	v, err := semverv3.NewVersion(currentVersion)
-	if err != nil {
-		return err
-	}
-
-	if a := c.Check(v); !a {
-		message := fmt.Sprintf("The current version %s is less than the minimum upgradable version %s.", currentVersion, minUpgradableVersion)
-		return fmt.Errorf("%s", message)
-	}
-
-	return nil
 }
 
 func upgradeEligibilityCheck(upgrade *harvesterv1.Upgrade) (bool, string) {
--- a/pkg/controller/master/upgrade/upgrade_controller_test.go
+++ b/pkg/controller/master/upgrade/upgrade_controller_test.go
@@ -254,66 +254,6 @@ func TestUpgradeHandler_OnChanged(t *testing.T) {
 	}
 }
 
-func Test_isVersionUpgradable(t *testing.T) {
-	var testCases = []struct {
-		name                 string
-		currentVersion       string
-		minUpgradableVersion string
-		isUpgradable         bool
-	}{
-		{
-			name:                 "upgrade from versions above the minimal requirement",
-			currentVersion:       "v1.1.1",
-			minUpgradableVersion: "v1.1.0",
-			isUpgradable:         true,
-		},
-		{
-			name:                 "upgrade from the exact version of the minimal requirement",
-			currentVersion:       "v1.1.0",
-			minUpgradableVersion: "v1.1.0",
-			isUpgradable:         true,
-		},
-		{
-			name:                 "upgrade from rc versions lower than the minimal requirement",
-			currentVersion:       "v1.1.0-rc1",
-			minUpgradableVersion: "v1.1.0",
-			isUpgradable:         false,
-		},
-		{
-			name:                 "upgrade from rc versions above the minimal requirement (rc minUpgradableVersion)",
-			currentVersion:       "v1.1.0-rc2",
-			minUpgradableVersion: "v1.1.0-rc1",
-			isUpgradable:         true,
-		},
-		{
-			name:                 "upgrade from rc versions lower than the minimal requirement (rc minUpgradableVersion)",
-			currentVersion:       "v1.1.0-rc1",
-			minUpgradableVersion: "v1.1.0-rc2",
-			isUpgradable:         false,
-		},
-		{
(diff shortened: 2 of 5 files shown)


w13915984028:
This PR will solve https://github.com/harvester/harvester/issues/8213 as well.

Another question: There is a min-upgrade version check in normal upgrade, in case the directly ISO upgrade, how is this checked?

w13915984028:
The issue https://github.com/harvester/harvester/issues/8226 is also related.

Do we have an HEP/rough solution about the ISO based upgrade? Let's try to put those together and enhance them. thanks.

brandboat:
Thanks for the heads-up, @w13915984028! I actually wasn’t aware of those two issues until you mentioned them.

> Another question: There is a min-upgrade version check in normal upgrade, in case the directly ISO upgrade, how is this checked?

kudos to @starbops, we have a great tool called `versionguard` to handle this, see 
https://github.com/harvester/harvester/blob/7fce535f5dcb98e9895bc14542326ae498641351/pkg/controller/master/upgrade/upgrade_controller.go#L323

> Do we have an HEP/rough solution about the ISO based upgrade? Let's try to put those together and enhance them. thanks.

Nope, there is no HEP. and we already implemented this feature, but it's malfunctioned in v1.5.0. What I am doing here is try to make this feature work again. For web UI changes, please refer to https://github.com/harvester/harvester-ui-extension/pull/306.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -675,37 +674,6 @@ func getCachedRepoInfo(upgrade *harvesterv1.Upgrade) (*repoinfo.RepoInfo, error)
 	return repoInfo, nil
 }
 
-func isVersionUpgradable(currentVersion, minUpgradableVersion string) error {
```
brandboat, line 678: This is unrelated to the issue, but I came across some unused code while tracing the code.

w13915984028:
When those 2 annotations are moved to upgrade object, does UI have plan to expose them for user to set/select? On both normal upgrade and ISO based upgrade.

And the doc https://docs.harvesterhci.io/v1.6/upgrade/index#customize-the-version needs to be updated after this PR. thanks.

brandboat:
> When those 2 annotations are moved to upgrade object, does UI have plan to expose them for user to set/select? On both normal upgrade and ISO based upgrade.

AFAIK, currently we don't have plan to do this. c.c. @FrankYang0529, @a110605.

> And the doc https://docs.harvesterhci.io/v1.6/upgrade/index#customize-the-version needs to be updated after this PR. thanks.

Sure, I'll file another doc PR after this PR merged.

w13915984028:
> we already implemented this feature, but it's malfunctioned in v1.5.0

It might not be related to this PR, but still some test cases need to be disigned and validated to ensure this feature itself works well, and it can handle potential race with normal upgrade,  and various failure cases.

w13915984028:
Originally, user can customize those 2 annotations by create a version object manually;  when move to ugprade and no place for user to set/select, how can user know and use it?

brandboat:
> Originally, user can customize those 2 annotations by create a version object manually; when move to ugprade and no place for user to set/select, how can user know and use it?

@w13915984028, since these two annotations are not exposed through the UI, users will need to manually add them to the Upgrade CR, right? 
One annotation, `harvesterhci.io/minCertsExpirationInDay`, is intended for debugging purposes in the beginning, while the other, `harvesterhci.io/skipGarbageCollectionThresholdCheck`, is used to skip the disk remaining space check. These annotations are not commonly used by most users. If a user does wish to use them, I assume they have the necessary knowledge to create the Upgrade CR themselves.

w13915984028:
ATM, the upgrade has some inconsistencies in user exprience, and when `ISO based upgrade` is added, we need to keep/improve user experience as well.  This PR might not be the best places to dicsuss, forgive me.

Traditional Upgrade:
(1) UI, user can set `settings`, click `upgrade` menu and click enable `upgradeog` and start upgrade;  the missing part is above two annotations;
(2) CLI, user can create `version`, `upgrade` to start upgrade

ISO based upgrade:
(1) UI, similar to above
(2) CLI, create `image` by upload ISO/URL, create `upgrade` ?   any additional steps are needed/not feasible?  suppose user needs to customize it this way to outcome some issue,  a detailed step description is needed on the doc.

w13915984028:
BTW:  The webhook check of the image needs to be updated:

https://github.com/harvester/harvester/blob/b48df2b7abdf02536b851d9fc35ae0bf212776ae/pkg/webhook/resources/upgrade/validator.go#L120


if `newUpgrade.Spec.Image != ""`

The image needs to be existing and ready, and also has upgrade related label/annotation (UI add via upload image ?), to avoid an unrelated image/ un-ready image is set via CLI/UI. cc @a110605

brandboat:
> if newUpgrade.Spec.Image != ""
> 
> The image needs to be existing and ready, and also has upgrade related label/annotation (UI add via upload image ?), to > avoid an unrelated image/ un-ready image is set via CLI/UI. cc @a110605

I'll handle this in next PR. Thanks.

w13915984028 (approved):
LGTM, thanks.

FrankYang0529 (approved):

Issue #8334, opened 2025-05-21
[BUG] Can't Restore 3rd Party Storage Data Volumes Using CSI VolumeSnapshot


ihcsim:
### Describe the Bug

The PVC mutator overrides any existing data source of a 3rd party storage file system data volumes, with the `filesystem-blank-source` `VolumeImportSource` configuration. This wipes any data restored by CSI because the CSI `VolumeSnapshot` data source would be overwritten. This prevents 3rd party CSI-based backup/restore software like Velero from restoring the volumes correctly.

### To Reproduce

1. Set up the NFS CSI, NFS server and NFS storage class by following the instructions in the [NFS CSI doc](https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/deploy/example/README.md)
2. Enable volume snapshotting on the NFS CSI following the instructions in the [NFS CSI snapshot doc](https://github.com/kubernetes-csi/csi-driver-nfs/tree/master/deploy/example/snapshot)
3. Create a VM image based on the NFS CSI storage class 
4. Create a VM with NFS root volume and data volume
5. Take a CSI snapshot of the NFS data volume:

```sh
cat <<EOF | k -n <vm-ns> apply -f -
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: vm-00-snap
spec:
  volumeSnapshotClassName: csi-nfs-snapclass  # default name from nfs csi doc
  source:
    persistentVolumeClaimName: <data-volume-pvc-name>
EOF
```

6. Confirm that the snapshot is successfully created:

```sh
k -n demo-src get volumesnapshot
NAME         READYTOUSE   SOURCEPVC            SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS       SNAPSHOTCONTENT                                    CREATIONTIME   AGE
vm-00-snap   true         vm-00-disk-1-zfck3                           9902618       csi-nfs-snapclass   snapcontent-3e74d348-c122-4a55-a12f-fbce8d58775f   17s            46s
```

7. Restore the snapshot to a new PVC:

```sh
cat <<EOF | k -n <vm-ns> apply -f -
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: restore-pvc
  annotations:
    harvesterhci.io/volumeForVirtualMachine: "true" # label this pvc as a data volume
spec:
  accessModes:
  - ReadWriteMany
  dataSource:
    name: vm-00-snap
[...]

PR #8335, opened 2025-05-21
fix: update pvc mutator to not overwrite existing pvc data source

ihcsim:
#### Problem:

The PVC mutator overrides any existing data source of a 3rd party storage file system data volumes, with the `filesystem-blank-source` `VolumeImportSource` configuration. This wipes any data restored by CSI because the CSI `VolumeSnapshot` data source would be overwritten. This prevents 3rd party CSI-based backup/restore software like Velero from restoring the volumes correctly.

#### Solution:

Change the PVC mutator to skip patching (replacing) any existing data source of a file system PVC.

#### Related Issue(s):

#8334

##### Test Setup

1. Set up the NFS CSI, NFS server and NFS storage class by following the instructions in the [NFS CSI doc](https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/deploy/example/README.md)
2. Enable volume snapshotting on the NFS CSI following the instructions in the [NFS CSI snapshot doc](https://github.com/kubernetes-csi/csi-driver-nfs/tree/master/deploy/example/snapshot)
3. Create a VM image based on the NFS CSI storage class 

##### Test Case 1 - Regression Test

1. Create a VM with NFS root volume and data volume
2. Expect the VM to be successfully created and the additional disk of the data volume is mounted

    * SSH into the VM
    * Confirm the additional disk is attached: `sudo fdisk -l`
        * This example assumes it's `/dev/vdb`
    * Partition disk: `sudo fdisk /dev/vdb`
         * Create partition: `n`
         * Use default settings for the new partition
         * Save changes: `w`
    * Format disk: `sudo mkfs.ext4 /dev/vdb1`
    * Mount disk: `sudo mount /dev/vdb1 /mnt/data`
    * Write data to `/mnt/data`:

```sh
sudo touch /mnt/data/00.dat
sudo touch /mnt/data/01.dat
sudo touch /mnt/data/02.dat
sudo /bin/bash -c 'echo "hello world0" > /mnt/data/00.dat'
sudo /bin/bash -c 'echo "hello world1" > /mnt/data/01.dat'
sudo /bin/bash -c 'echo "hello world2" > /mnt/data/02.dat'
```

##### Test Case 2 - Restore Data Volume Snapshot

1. Re-use the VM from Test Case 1
2. Take a CSI snapshot of the NFS data volume:

```sh
cat <<EOF | k -n <vm-ns> apply -f -
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: vm-00-snap
spec:
  volumeSnapshotClassName: csi-nfs-snapclass  # default name from nfs csi doc
  source:
    persistentVolumeClaimName: <data-volume-pvc-name>
EOF
```

3. Confirm that the snapshot is successfully created:

```sh
k -n demo-src get volumesnapshot
NAME         READYTOUSE   SOURCEPVC            SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS       SNAPSHOTCONTENT                                    CREATIONTIME   AGE
vm-00-snap   true         vm-00-disk-1-zfck3                           9902618       csi-nfs-snapclass   snapcontent-3e74d348-c122-4a55-a12f-fbce8d58775f   17s            46s
```

4. Restore the snapshot to a new PVC:

```sh
cat <<EOF | k -n <vm-ns> apply -f -
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: restore-pvc
  annotations:
    harvesterhci.io/volumeForVirtualMachine: "true" # label this pvc as a data volume
spec:
  accessModes:
  - ReadWriteMany
  dataSource:
    name: vm-00-snap
    kind: VolumeSnapshot
    apiGroup: snapshot.storage.k8s.io
  resources:
    requests:
      storage: 10Gi
  storageClassName: nfs-csi
  volumeMode: Filesystem
EOF
```

5. Confirm that the PVC is successfully restored:

```sh
k -n demo-src get pvc restore-pvc
NAME          STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
restore-pvc   Bound    pvc-bb58ab7d-3ffa-4163-9eee-dffa188b696a   10Gi       RWX            nfs-csi        <unset>                 42s
```

6. Create a new VM to mount the `restore-pvc` as an extra disk:

image

7. Once the VM is ready, access it via SSH and confirm that the disk and its data is correctly restored:

```sh
ubuntu@vm-00-restored:~$ sudo fdisk -l /dev/vdb
Disk /dev/vdb: 9.45 GiB, 10146021376 bytes, 19816448 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x1fe9f7a3

Device     Boot Start      End  Sectors  Size Id Type
/dev/vdb1        2048 19816447 19814400  9.4G 83 Linux

ubuntu@vm-00-restored:~$ sudo mkdir /mnt/data
ubuntu@vm-00-restored:~$ sudo mount /dev/vdb1 /mnt/data
ubuntu@vm-00-restored:~$ ls -al /mnt/data
total 36
drwxr-xr-x 3 root root  4096 May 20 21:17 .
drwxr-xr-x 3 root root  4096 May 21 18:33 ..
-rw-r--r-- 1 root root    13 May 20 21:17 00.dat
-rw-r--r-- 1 root root    13 May 20 21:17 01.dat
-rw-r--r-- 1 root root    13 May 20 21:17 02.dat
drwx------ 2 root root 16384 May 20 21:17 lost+found

```

### DIFF
--- a/pkg/webhook/resources/persistentvolumeclaim/mutator.go
+++ b/pkg/webhook/resources/persistentvolumeclaim/mutator.go
@@ -113,6 +113,21 @@ func (m pvcMutator) patchDataSource(pvc *corev1.PersistentVolumeClaim) (string, 
 		return "", nil
 	}
 
+	// do not replace any existing data source, such as csi snapshot, with the
+	// filesystem-blank-source VolumeImportSource. if set, the 'blank'
+	// VolumeImportSource would wipe any restore or prepopulated data.
+	if pvc.Spec.DataSourceRef != nil || pvc.Spec.DataSource != nil {
+		dataSource := ""
+		if pvc.Spec.DataSourceRef != nil {
+			dataSource = fmt.Sprintf("%+v", pvc.Spec.DataSourceRef)
+		} else if pvc.Spec.DataSource != nil {
+			dataSource = fmt.Sprintf("%+v", pvc.Spec.DataSource)
+		}
+		pvcName := fmt.Sprintf("%s/%s", pvc.Namespace, pvc.Name)
+		logrus.Debugf("PVC %s has existing data source: %s, skip patch", pvcName, dataSource)
+		return "", nil
+	}
+
 	cdiAPIGroup := cdicommon.AnnAPIGroup
 	dataSourceRef := corev1.TypedLocalObjectReference{
 		APIGroup: &cdiAPIGroup,
--- a/pkg/webhook/resources/persistentvolumeclaim/mutator_test.go
+++ b/pkg/webhook/resources/persistentvolumeclaim/mutator_test.go
@@ -1,0 +1,123 @@
+package persistentvolumeclaim
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/harvester/harvester/pkg/util"
+)
+
+func TestPatchDataSource(t *testing.T) {
+	var (
+		blockMode = corev1.PersistentVolumeBlock
+		fsMode    = corev1.PersistentVolumeFilesystem
+	)
+	var testCases = []struct {
+		desc        string
+		pvc         *corev1.PersistentVolumeClaim
+		expected    string
+		expectedErr error
+	}{
+		{
+			desc: "skip if block volume mode",
+			pvc: &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-pvc",
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					VolumeMode: &blockMode,
+				},
+			},
+			expected:    "",
+			expectedErr: nil,
+		},
+		{
+			desc: "skip if no annotations",
+			pvc: &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "test-pvc",
+					Annotations: map[string]string{},
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					VolumeMode: &fsMode,
+				},
+			},
+			expected:    "",
+			expectedErr: nil,
+		},
+		{
+			desc: "skip if data source ref exists",
+			pvc: &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-pvc",
+					Annotations: map[string]string{
+						util.AnnotationVolForVM: "true",
+					},
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					VolumeMode: &fsMode,
+					DataSourceRef: &corev1.TypedObjectReference{
+						Name: "new-snapshot-test",
+						Kind: "VolumeSnapshot",
+					},
+				},
+			},
+			expected:    "",
+			expectedErr: nil,
+		},
+		{
+			desc: "skip if data source exists",
+			pvc: &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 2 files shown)


WebberHuang1118 (dismissed):
LGTM, thanks!

I just have one question:
For a data PVC created from scratch—if we skip [this section of code](https://github.com/harvester/harvester/blob/fcd33eddd85aa0ad72127fbf28497ddf1900a6cb/pkg/webhook/resources/persistentvolumeclaim/mutator.go#L116-L129), what would happen?

Will the PVC’s filesystem simply not contain the `disk.img` file?

Vicente-Cheng (dismissed):
nice catch!
Thanks for fixing that.

ihcsim:
@WebberHuang1118 @Vicente-Cheng I updated this PR to remove the calls to patch the data source reference, but kept the method for hotplug volumes enhancement. PTAL.

innobead:
> @WebberHuang1118 @Vicente-Cheng I updated this PR to remove the calls to patch the data source reference, but kept the method for hotplug volumes enhancement. PTAL.

Can we explain this a bit? Any context for hotplug volume in this case?

ihcsim:
From what I gathered, hotplug only works for either block volumes or fs volumes with `disk.img`. In the latter case, kubevirt does not auto-create the `disk.img`, so we have to do it ourselves via CDI. I decided to leave the implementation, even though my first thought is to remove it. Upstream support for hotplug fs is still pending, per https://github.com/kubevirt/kubevirt/issues/11699.

Vicente-Cheng:
> From what I gathered, hotplug only works for either block volumes or fs volumes with `disk.img`. In the latter case, kubevirt does not auto-create the `disk.img`, so we have to do it ourselves via CDI. I decided to leave the implementation, even though my first thought is to remove it. Upstream support for hotplug fs is still pending, per [kubevirt/kubevirt#11699](https://github.com/kubevirt/kubevirt/issues/11699).

This is a different case.

We still need to hotplug the block device (regardless of block/filesystem mode volume)
The issue is that when we treat the filesystem mode volume as the block device for the kubevirt, we need `disk.img`.

The hotplug filesystem is another story.

WebberHuang1118:
> > From what I gathered, hotplug only works for either block volumes or fs volumes with `disk.img`. In the latter case, kubevirt does not auto-create the `disk.img`, so we have to do it ourselves via CDI. I decided to leave the implementation, even though my first thought is to remove it. Upstream support for hotplug fs is still pending, per [kubevirt/kubevirt#11699](https://github.com/kubevirt/kubevirt/issues/11699).
> 
> This is a different case.
> 
> We still need to hotplug the block device (regardless of block/filesystem mode volume) The issue is that when we treat the filesystem mode volume as the block device for the kubevirt, we need `disk.img`.
> 
> The hotplug filesystem is another story.

Hi @Vicente-Cheng,

My understanding aligns with @ihcsim — KubeVirt does not generate `disk.img` for filesystem-mode hotplug volumes, which is why we need CDI's assistance. Could you clarify your concerns regarding @ihcsim's proposed design?

ihcsim:
Yeah, we will need `disk.img` as long as it's virtio. It's a matter whether kubevirt will auto-create it or not. I decided to not remove the original implementation of the blank import data source because I assume we will need it for fs hotplug. If that's not case, or the future implementation will be different, then I can remove it.

Vicente-Cheng (approved):
thanks!

ihcsim:
Merging this PR since it was already previously approved.

Issue #8331, opened 2025-05-21
[ENHANCEMENT] Ehancement on VM network & interface check


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Add additional webhook check on VM:

(1) Don't allow VM to refer the `storagenetwork` NAD, as https://github.com/harvester/harvester/issues/8317 experimented.  cc @rrajendran17 

(2) Don't allow VM to refer the `migrationnetwork` NAD, coming by PR https://github.com/harvester/harvester/pull/8174  cc @FrankYang0529 

(3) When a VM has multi network interfaces upon same NAD, their MACs can't be duplicated.    refer: https://github.com/harvester/harvester/pull/8325, https://github.com/harvester/harvester/issues/8293#issuecomment-2915480815

e.g.: in case on Harvester-on-Harvester, it is also reasonable that two interfaces/NICs are connected to same network, thus guest Harvester can create bond interface like bare-metal mode

(4) Recheck feature: duplicated MAC check after PR https://github.com/harvester/harvester/pull/7084 is merged.

w13915984028:
FYI: Move issue to v170, need some time to recheck all existing codes to ensure if some additional checks are still needed. Thanks.

PR #8339, opened 2025-05-22
feat: patch macaddress to vmi

FrankYang0529:
#### Problem:
In before, we backport VMI macaddress to VM, so VM can have same macaddress after restart. However, this makes VM bumps revision and has `VirtualMachineRestartRequired` condition.

#### Solution:

Backport VMI macaddress to VM annotation and use mutator to add macaddress data to new VMI.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8331

#### Test plan:

1. Create a VM and check there is `harvesterhci.io/mac-address` annotation.
2. Migrate / Restart / Stop and Start VM. Check VMI has same mac address.

### DIFF
--- a/pkg/webhook/resources/virtualmachineinstance/mutator.go
+++ b/pkg/webhook/resources/virtualmachineinstance/mutator.go
@@ -1,0 +1,79 @@
+package virtualmachineinstance
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+func NewMutator(
+	vm ctlkubevirtv1.VirtualMachineCache,
+) types.Mutator {
+	return &vmiMutator{
+		vm: vm,
+	}
+}
+
+type vmiMutator struct {
+	types.DefaultMutator
+	vm ctlkubevirtv1.VirtualMachineCache
+}
+
+func (m *vmiMutator) Resource() types.Resource {
+	return types.Resource{
+		Names:      []string{"virtualmachineinstances"},
+		Scope:      admissionregv1.NamespacedScope,
+		APIGroup:   kubevirtv1.SchemeGroupVersion.Group,
+		APIVersion: kubevirtv1.SchemeGroupVersion.Version,
+		ObjectType: &kubevirtv1.VirtualMachineInstance{},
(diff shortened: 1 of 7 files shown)


pkg/webhook/resources/virtualmachineinstance/mutator.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	vmiInterfaces := map[string]string{}
+	if err = json.Unmarshal([]byte(vm.Annotations[util.AnnotationMacAddressName]), &vmiInterfaces); err != nil {
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"name":      vm.Name,
+			"namespace": vm.Namespace,
+			"macs":      vm.Annotations[util.AnnotationMacAddressName],
+		}).Error("failed to unmarshal mac-address from vm annotation")
+		return nil, err
```
w13915984028: skip the error to allow vm creating?

pkg/webhook/resources/virtualmachineinstance/mutator.go
```diff
@@ -0,0 +1,79 @@
...
+	patchOps := types.PatchOps{}
+	for i, iface := range vmi.Spec.Domain.Devices.Interfaces {
+		if vm.Spec.Template.Spec.Domain.Devices.Interfaces[i].MacAddress != "" {
+			continue
+		}
+		ifaceMac, ok := vmiInterfaces[iface.Name]
+		if !ok || ifaceMac == "" {
+			continue
+		}
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "add", "path": "/spec/domain/devices/interfaces/%d/macAddress", "value": "%s"}`, i, ifaceMac))
```
w13915984028, line 80: could you add a test code to validate the mutator output?

w13915984028 (dismissed):
LGTM, thanks.

Need a double check on the original implementation:

When a VM is using the `default` network, the `MAC` is inheriated from pod network; both MAC and IP are allocated by k8s internal, is it essential to save & restore this MAC?

pkg/webhook/resources/virtualmachineinstance/mutator.go
FrankYang0529: Agree, it's better to not block VMI creation.

pkg/webhook/resources/virtualmachineinstance/mutator.go
FrankYang0529: Added it.

w13915984028 (approved):
LGTM, thanks.

itseyesack:
Encountered another problem caused by this change. When using the harvester-vm-dhcp-controller add-on to provide DHCP for VMs, it watches for changes to VM objects and creates/updates virtualmachinenetworkconfigs according to the spec.template.spec.domain.devices.interfaces.macAddress field on the virtualmachine. Now in Harvester v1.6.0 with this update to the way mac addresses are tracked, the VirtualMachine object is no longer patched to include the MAC at spec.template.spec.domain.devices.interfaces.macAddress, instead only in the annotation. Since the change was otherwise a clear improvement, and the dhcp plugin is considered experimental, this would likely best be solved with changes to that project.

Issue #8077, opened 2025-04-15
[FEATURE] Expose CDI settings in Harvester to allow fine tuning CDI settings for day 2 operations


innobead:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

Customize CDI settings.

Vicente-Cheng:
Keep on reviewing until the UI ready

Vicente-Cheng:
Test PASS, close as complete


## Environment
* `harvester-v1.6.0-rc5`
  * Profile: 2 nodes AMD64 QEMU/KVM
  * ui-source: `Auto`


## Test Scenario
1. <details>
   <summary> Create a 2 nodes cluster
   </summary>

   </details>

1. <details>
   <summary> Enable Longhorn v2/ NFS / LVM
   </summary>

     ```
     # kubectl get settings longhorn-v2-data-engine-enabled
     NAME                              VALUE
     longhorn-v2-data-engine-enabled   true
     ```

     ```
     # kubectl get csidriver
     NAME                         ATTACHREQUIRED   PODINFOONMOUNT   STORAGECAPACITY   TOKENREQUESTS   REQUIRESREPUBLISH   MODES        AGE
     driver.longhorn.io           true             true             false             <unset>         false               Persistent   16h
     lvm.driver.harvesterhci.io   false            true             false             <unset>         false               Persistent   9h
     nfs.csi.k8s.io               false            false            false             <unset>         false               Persistent   8h
     ```

   </details>

1. <details>
   <summary> Make sure we have default annotation values for LVM & Longhorn V2 storage classes
   </summary>


     LVM
     ```
     # cat lvm-sc.yaml
     allowVolumeExpansion: false
     allowedTopologies:
     - matchLabelExpressions:
       - key: topology.lvm.csi/node
         values:
         - harvester-node-0
     apiVersion: storage.k8s.io/v1
     kind: StorageClass
     metadata:
       name: lvm-sc-node-0
     parameters:
       type: striped
       vgName: vg001
     provisioner: lvm.driver.harvesterhci.io
     reclaimPolicy: Delete
     volumeBindingMode: WaitForFirstConsumer
     
     # kubectl apply -f lvm-sc.yaml
     storageclass.storage.k8s.io/lvm-sc-node-0 created
     
     # kubectl get sc lvm-sc-node-0 -o yaml
     allowVolumeExpansion: false
     allowedTopologies:
     - matchLabelExpressions:
       - key: topology.lvm.csi/node
         values:
[...]

PR #8340, opened 2025-05-23
Add HEP: expose cdi settings in ui

brandboat:
This is the HEP for https://github.com/harvester/harvester/issues/8077


enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,104 @@
...
+- Expose five CDI settings (FileSystemOverhead, CloneStrategy, VolumeSnapshotClass, AccessModes, VolumeMode) in the Harvester UI
+- Eliminate the need for users to directly modify CDI and StorageProfile CRDs
+- Provide intuitive UI controls with proper validation
+- Maintain consistency between UI configurations and underlying CRDs
+
+## Proposal
+
+### User Stories
+
+#### Story 1: User Configuring Filesystem Overhead
```
ibrokethecloud, line 26: filesystem overhead seems to be a global setting not tied to a specific storage class.
should we move the UI element to be a new harvester setting rather than leaving it in the storageclass section

@Vicente-Cheng

ibrokethecloud (commented):
rest looks good. we just need to get agreement on where to place the filesystem overhead

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: Hi @ibrokethecloud, the file system overhead config can be configured globally, but also supports per-storageclass overrides:

As noted in CDI doc https://github.com/kubevirt/containerized-data-importer/blob/main/doc/cdi-config.md

> How much of a Filesystem volume's space should be reserved for overhead related to the Filesystem. This is a composite value, that contains global and per-storageClass config. Please look below for details.

You can configure it via the CDIConfig spec:
https://github.com/kubevirt/containerized-data-importer/blob/7715443c66138d71a41fd835710bb65e59f09e1a/staging/src/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go#L992

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,104 @@
...
+- The new settings can be configured during the creation of a StorageClass and update a StorageClass.
+- The Volume Mode / Access Modes setting supports multiple combinations, where each combination must have a unique Volume Mode and one or more associated Access Modes.
+- The Volume Snapshot Class is presented as a dropdown menu listing all available VolumeSnapshotClass names.
+- The Clone Strategy is also a dropdown menu with the following fixed options: "copy", "snapshot", and "csi-clone".
+- The File System Overhead must be a value between 0 and 1, with a regex pattern of ^(0(?:\.\d{1,3})?|1)$ (i.e., up to three digits after the decimal point).
+#### Backend Implementation
+
+Introduce 4 annotations to StorageClass:
+
+- `cdi.harvesterhci.io/filesystemOverhead`: String value representing filesystem overhead percentage (0-1), regex pattern: `^(0(?:\.\d{1,3})?|1)$` (up to 3 digits after the decimal point)
```
a110605, line 69: Please provide the tooltip message that we can help user understand what these fields meaning.

enhancements/20250522-expose-cdi-settings-in-ui.md
Vicente-Cheng: thanks @brandboat, 

Yes, we could have the default filesystem overhead and it would be overridden with the filesystem overhead of storageclass

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,106 @@
...
+
+The CDI settings will be integrated into the existing Storage Classes page in Harvester UI.
+
+### API changes
+
+## Design
+
+### Implementation Overview
+
+#### UI Preview
```
Vicente-Cheng, line 52: Honestly, I prefer an extra panel instead of combining it with the customize panel.

But it is more related to the UI design, WDYT? cc @a110605

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,106 @@
...
+
+![storageclass-ui](./20250522-expose-cdi-settings-in-ui/storageclass-ui.png)
+
+- Note: These new settings are not available in the longhorn-v1 StorageClass.
+- The new settings can be configured during the creation of a StorageClass and update a StorageClass.
```
Vicente-Cheng:
```suggestion
- The new settings can be configured during the creation of a StorageClass or by updating a StorageClass.
```

Vicente-Cheng (commented):
some nits, others lgtm

Vicente-Cheng:
I would like to raise another issue here.

Currently, the storageprofile controller would help to complete some fields (like CloneStrategy, accessMode/volumeMode) of our validated CSI drivers (like LonghornV2/LVM).
Do we still need those behaviors after we introduce this feature? That might have some conflicts.

cc @brandboat, @WebberHuang1118, @ibrokethecloud, @innobead

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,106 @@
...
+- The Volume Mode / Access Modes setting supports multiple combinations, where each combination must have a unique Volume Mode and one or more associated Access Modes.
+  - The volume mode should be either `FileSystem` or `Block`.
+  - The access modes can have 0~4 unique mode, the access mode could be either `ReadWriteOnce`, `ReadOnlyMany`, `ReadWriteMany` or `ReadWriteOncePod`.
+- The Volume Snapshot Class is presented as a dropdown menu listing all available VolumeSnapshotClass names.
```
ihcsim: Does the snapshot class defined here takes precedence over that in the `csi-driver-config` setting https://docs.harvesterhci.io/v1.5/advanced/index/#csi-driver-config?

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: @Vicente-Cheng, did you mean expose global filesystem overhead to the UI?

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: Alright, I know this is a bit of a headache, but I’d like to first clarify a few things.

The csi-driver-config is used when creating a VolumeSnapshot. It tells us which VolumeSnapshotClass to use for a given provisioner.

On the other hand, the snapshotClass in CDI only takes effect during a [smart clone](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/smart-clone.md). When we create a DataVolume from a PVC, CDI looks up the snapshotClass defined in the corresponding StorageProfile and uses it to create the VolumeSnapshot.

So these two settings serve different purposes and don’t really affect each other.
Personally, I think the scopes and intents of these two parameters are quite different:
- csi-driver-config affects how we create VM/volume snapshots.
- snapshotClass in CDI affects the smart-clone flow.

This leads to the question: how should we interpret `csi-driver-config`? Should we expand its scope to also cover CDI’s smart clone process? I'm not sure about how do we handle this... There are two approaches here:
1. Treat the two settings separately: This means we don’t make any changes, but we’ll need to clearly document the difference, since both use the term VolumeSnapshotClass, which could be confusing.
2. Make csi-driver-config influence CDI’s smart clone flow as well: In this case, we’d need to define precedence rules — which config overrides which — and that might make things more complex.

I personally lean toward option 1, but I’m curious to hear what others think.
c.c. @Vicente-Cheng @WebberHuang1118, please correct me if I'm wrong.

ihcsim: For a given storage class, I think it's unlikely that the volume snapshot class will be different for the VM snapshot flow vs. the CDI flow. The latter is essentially the VM creation and image creation flows. (AIUI, storage profiles have no effects on volumes creation, as no CDI data volumes are involved, unless we want to enable the [CDI PVC mutator labeling](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/pvc-mutating-webhook-rendering.md) feature.) Hence, the volume snapshot class used to create the VM root disk (i.e. smart clone from the golden image) ~should be~ is likely the same as the one used to snapshot it later.

~I think CDI settings are more suitable for 3rd party storage usages than csi-driver-config. It should have precedence over csi-driver-config.~

UPDATE: I changed my mind. The two volume snapshot classes don't have to be the same even though in most cases, they are. It will be nice to clarify them though. If `csi-driver-config` is only used for VM snapshot, then maybe the UI can add some text for it.

Vicente-Cheng: I thought these two configurations had different purposes.

`StorageProfile` is the CRD used for the CDI
`csi-driver-config` is the general purpose for our Harvester things (for both snapshot/backup)

These two configurations did not conflict at all.
- For VM/VM Image creation, we rely on the CDI, which uses the `snapshotClass` of the corresponding StorageProfile (if the `cloneStrategy` is `snapshot`).
- For the purpose like VM/Volume snapshot(backup), we will check/use `csi-driver-config`.

So, it's fine to keep these two things, but as @ihcsim's concern, we need to have the detailed document for it.
The user would know how to configure the right value.

Thanks!

enhancements/20250522-expose-cdi-settings-in-ui.md
Vicente-Cheng: No, I just mentioned that we could have filesystem overhead with SC. It will override the global configuration.

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: I think it should be ok, I'll add a new tab named "CDI settings" to cover all cdi settings introduced in this HEP.

截圖 2025-06-04 中午12 05 54

brandboat:
> Currently, the storageprofile controller would help to complete some fields (like CloneStrategy, accessMode/volumeMode) of our validated CSI drivers (like LonghornV2/LVM).
Do we still need those behaviors after we introduce this feature? That might have some conflicts.

@Vicente-Cheng Yes, that'll bring conflicts, I think we should remove those control logic, will mention this part in the HEP, thanks for the reminder!

enhancements/20250522-expose-cdi-settings-in-ui.md
WebberHuang1118: I have the same question as @ibrokethecloud — does the annotation `cdi.harvesterhci.io/filesystemOverhead` apply globally or per StorageClass?

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,114 @@
...
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  annotations:
+    cdi.harvesterhci.io/filesystemOverhead: "0.85"
+    cdi.harvesterhci.io/cloneStrategy: "snapshot"
+    cdi.harvesterhci.io/volumeSnapshotClass: "csi-snapshot-class"
+    cdi.harvesterhci.io/volumeModeAccessModes: |
+      {"Block":["ReadWriteOnce","ReadOnlyMany"],"Filesystem":["ReadWriteOnce","ReadWriteMany"]}
```
WebberHuang1118, line 85: I have a question about the annotation `cdi.harvesterhci.io/volumeModeAccessModes`.
If the annotation contains multiple combinations, how does CDI determine which one to use?

For example, given the annotation:
```
{
  "Block": ["ReadWriteOnce", "ReadOnlyMany"],
  "Filesystem": ["ReadWriteOnce", "ReadWriteMany"]
}
```
Will CDI choose block or filesystem mode?
And if it selects block mode, will the access mode be `ReadWriteOnce` or `ReadOnlyMany`?

WebberHuang1118 (commented):
Hi @brandboat,

Thanks for the HEP! Could you please help clarify the questions I raised? Thanks!

WebberHuang1118:
> I would like to raise another issue here.
> 
> Currently, the storageprofile controller would help to complete some fields (like CloneStrategy, accessMode/volumeMode) of our validated CSI drivers (like LonghornV2/LVM). Do we still need those behaviors after we introduce this feature? That might have some conflicts.
> 
> cc @brandboat, @WebberHuang1118, @ibrokethecloud, @innobead

But some modifications to the current StorageProfile controller are still required, right? cc @Vicente-Cheng @brandboat

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: @WebberHuang1118, all annotations in this HEP are defined per storageclass.

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: CDI will use the first entry in the ClaimPropertySets as the default value.
https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/datavolume/util.go#L391-L396

e.g. in the example, CDI will use the first entry of claimPropertySets defined in storage profile.
```yaml
kind: StorageProfile
metadata:
  name: my-storage-profile
status:
  claimPropertySets:
    - accessModes:  # CDI choose this as default access modes.
        - ReadWriteOnce
        - ReadOnlyMany
      volumeMode: Filesystem # CDI choose this as default volume mode
    - accessModes:
        - ReadWriteMany
      volumeMode: Block
...
```

Note: the CDI priority rules are here: https://github.com/kubevirt/containerized-data-importer/blob/main/doc/storageprofile.md#priorities, 

> Priorities:
> 1. Overrides (for example cdi.Spec.CloneStrategyOverride)
> 2. Parameter defined on DataVolume
> 3. User provided parameters - defined on StorageProfile spec section.
> 4. Parameters provided by CDI.
> 5. Empty or kubernetes defaults (if available).

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,114 @@
...
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  annotations:
+    cdi.harvesterhci.io/filesystemOverhead: "0.85"
+    cdi.harvesterhci.io/cloneStrategy: "snapshot"
+    cdi.harvesterhci.io/volumeSnapshotClass: "csi-snapshot-class"
+    cdi.harvesterhci.io/volumeModeAccessModes: |
```
innobead: Consider utilizing the names below, as they can be directly linked to the fields in CDI-related CRDs. This can serve as a convention for standardizing annotation naming for configuring other CRDs. 

```
cdi.harvesterhci.io/fileSystemOverhead: // this is for CDI spec, so no need to have the prefix.
cdi.harvesterhci.io/storageProfileCloneStrategy: 
cdi.harvesterhci.io/storageProfileSnapshotClass: 
cdi.harvesterhci.io/storageProfileClaimPropertySets:
```

[/vendor/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go#L419-L429](https://github.com/harvester/harvester/blob/7a98d6231aad3af89650ea0c379bcd359b96efc1/vendor/kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1/types.go#L419-L429)
```go
type StorageProfileSpec struct {
	// CloneStrategy defines the preferred method for performing a CDI clone
	CloneStrategy *CDICloneStrategy `json:"cloneStrategy,omitempty"`
	// ClaimPropertySets is a provided set of properties applicable to PVC
	// +kubebuilder:validation:MaxItems=8
	ClaimPropertySets []ClaimPropertySet `json:"claimPropertySets,omitempty"`
	// DataImportCronSourceFormat defines the format of the DataImportCron-created disk image sources
	DataImportCronSourceFormat *DataImportCronSourceFormat `json:"dataImportCronSourceFormat,omitempty"`
	// SnapshotClass is optional specific VolumeSnapshotClass for CloneStrategySnapshot. If not set, a VolumeSnapshotClass is chosen according to the provisioner.
	SnapshotClass *string `json:"snapshotClass,omitempty"`
}
```

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,114 @@
...
+      {"Block":["ReadWriteOnce","ReadOnlyMany"],"Filesystem":["ReadWriteOnce","ReadWriteMany"]}
+```
+
+The implementation consists of three main components:
+
+1. StorageClass Webhook Validation
+Extend the existing StorageClass validation webhook to validate the five new annotations. The webhook will perform pattern matching for filesystemOverhead, enum validation for cloneStrategy, existence checks for volumeSnapshotClass, and JSON schema validation for volumeModeAccessModes.
```
innobead, line 91: nit: Does CDI have the corresponding validating admission? If so, we might be able to delegate the validation to it, as I assume it's validated synchronously during configuration. The downside could be that the error message can't be customized.

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,114 @@
...
+2. StorageProfile Controller
+Modify the existing storageprofile_controller to watch for StorageClass changes and synchronize cloneStrategy, volumeSnapshotClass, and volumeModeAccessModes annotations to the corresponding StorageProfile CRD.
```
innobead: Check if it's feasible to only monitor the annotated storage classes to avoid watching/reconciling all storage classes.

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,114 @@
...
+3. New StorageClass Controller
```
innobead: Have you considered renaming the StorageProfile controller to a CDI-specific controller? This way, we can consolidate all CDI-related resource handling into a single controller for better organization. This also prevents redundant resource reconciling costs.  

cc @Vicente-Cheng

innobead (commented):
In general, LGTM. Just some feedback.

For any configuration, it would be beneficial to have global settings alongside granular settings at the VM or volume level, where feasible.

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,114 @@
...
+Introduce a new storageclass_controller specifically for handling filesystemOverhead synchronization to the CDI object. This controller watches StorageClass events and updates the CDI object's filesystemOverhead configuration accordingly.
```
WebberHuang1118: Just wondering — why are two controllers needed here? From my understanding, reconciling the StorageClass annotation and updating the corresponding CDI settings or StorageProfile should be sufficient.

brandboat: Initially, I was thinking of handling all storage profile logic within the StorageProfile controller—when annotations are added to a StorageClass, we’d enqueue the corresponding StorageProfile event and let the storage profile controller pick it up to update the profile accordingly.

However, considering that we might not need the storage_profile_controller anymore (as discussed in https://github.com/harvester/harvester/pull/8340#issuecomment-2920169918), I agree with you—this logic can be handled directly in the StorageClass controller instead.

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: Yes, while CDI will validate these settings, we don’t allow users to modify the StorageProfile directly through the UI. Instead, users can add annotations to the StorageClass. However, any validation errors from the CDI side won’t surface in the UI, since the UI only captures validator errors from the StorageClass object. Therefore, we still need to implement our own validation logic for these annotations defined in StorageClass.

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: As far as I know, Wrangler doesn't support this scenario. What we can do is return early in the controller if the given object doesn't contain the specific annotation.
for example:
```go
ctrl.OnChange(ctx, "sc-handler", func(key string, obj *storagev1.StorageClass) (*storagev1.StorageClass, error) {
    if obj == nil {
        return nil, nil
    }
    if _, ok := obj.Annotations["my-annotation"]; !ok {
        return obj, nil
    }
    ...
```

brandboat:
Regarding this comment https://github.com/harvester/harvester/pull/8340#issuecomment-2944533811

> But some modifications to the current StorageProfile controller are still required, right? cc @Vicente-Cheng @brandboat

@WebberHuang1118 raised a point about whether we should consider adding default values for the annotations introduced in this HEP for known supported storage backends (Ceph, LVM, Longhorn v2, NFS). Additionally, should we validate the user-provided values for these supported storage types?

enhancements/20250522-expose-cdi-settings-in-ui.md
ihcsim: > The Volume Snapshot Class is presented as a dropdown menu listing all available VolumeSnapshotClass names.

Follow-up question: Do we have to display all available snapshot classes in the cluster? Or can we use the `csi-driver-config` to auto identify the snapshot class associated with this storage class provisioner? As discussed, it's very unlikely that user would use different snapshot class for VM snapshot/backup and vm image creation.

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,126 @@
...
+metadata:
+  annotations:
+    cdi.harvesterhci.io/filesystemOverhead: "0.85"
+    cdi.harvesterhci.io/storageProfileCloneStrategy: "snapshot"
+    cdi.harvesterhci.io/storageProfileVolumeSnapshotClass: "csi-snapshot-class"
+    cdi.harvesterhci.io/storageProfileVolumeModeAccessModes: |
+      {"Block":["ReadWriteOnce","ReadOnlyMany"],"Filesystem":["ReadWriteOnce","ReadWriteMany"]}
+```
+
+The implementation consists of three main components:
```
ihcsim:
```suggestion
The implementation consists of two main components:
```

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,126 @@
...
+- add annotations `cdi.harvesterhci.io/storageProfileCloneStrategy: "snapshot"` and `cdi.harvesterhci.io/storageProfileVolumeSnapshotClass: "longhorn-snapshot"` to LonghornV2 StorageClasses if both annotations are not set.
+- add `cdi.harvesterhci.io/storageProfileVolumeModeAccessModes: "{\"Block\":[\"ReadWriteOnce\"]}"` to LVM StorageClasses if the annotation is not already present.
+
+Note that Longhorn v2 will support CSI clone starting from v1.9.1, we need to set the default CDI clone strategy to `"clone"` in the storage class annotation `cdi.harvesterhci.io/storageProfileCloneStrategy`.
+
+### Test plan
+
+1. Create StorageClass with valid annotations and verify synchronization
+2. Update StorageClass annotations and verify changes propagate
+3. Delete StorageClass and verify cleanup behavior
```
ihcsim: Does cleanup involve deleting the storage profile even though it is created by CDI?

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,126 @@
...
+4. Test validation rejection for each annotation type
+5. Test partial annotation configurations (not all five annotations present)
+
+### Upgrade strategy
+
+Existing StorageClass, CDI, and StorageProfile configurations will continue to work unchanged. The new annotation-based approach is purely additive.
+
+### Notes
+- The `csi-driver-config` and `volume-snapshot-class` settings are unrelated to `cdi.harvesterhci.io/volumeSnapshotClass`. The former two are used when creating a VolumeSnapshot, specifying which VolumeSnapshotClass to use for a given provisioner—primarily in the context of volume snapshots or backups.
```
ihcsim:
```suggestion
- The `csi-driver-config` and `volume-snapshot-class` settings are unrelated to ` cdi.harvesterhci.io/storageProfileVolumeSnapshotClass`. The former two are used when creating a VolumeSnapshot, specifying which VolumeSnapshotClass to use for a given provisioner—primarily in the context of volume snapshots or backups.
```

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,126 @@
...
+In contrast, `cdi.harvesterhci.io/volumeSnapshotClass` is used during VM or VM image creation via third-party storage solutions that rely on CDI’s [smart-clone](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/smart-clone.md) mechanism. When the DataVolume source is a PVC and the CloneStrategy is set to snapshot, CDI creates a VolumeSnapshot behind the scenes. In this case, the value of `cdi.harvesterhci.io/volumeSnapshotClass` determines which VolumeSnapshotClass CDI uses.
```
ihcsim:
```suggestion
In contrast, ` cdi.harvesterhci.io/storageProfileVolumeSnapshotClass` is used during VM or VM image creation via third-party storage solutions that rely on CDI’s [smart-clone](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/smart-clone.md) mechanism. When the DataVolume source is a PVC and the CloneStrategy is set to snapshot, CDI creates a VolumeSnapshot behind the scenes. In this case, the value of ` cdi.harvesterhci.io/storageProfileVolumeSnapshotClass` determines which VolumeSnapshotClass CDI uses.
```

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: > Follow-up question: Do we have to display all available snapshot classes in the cluster? Or can we use the csi-driver-config to auto identify the snapshot class associated with this storage class provisioner? As discussed, it's very unlikely that user would use different snapshot class for VM snapshot/backup and vm image creation.

Perhaps we can default the snapshot class in storage profile to the one specified in csi-driver-config.
For example, if the user doesn’t set cdi.harvesterhci.io/storageProfileVolumeSnapshotClass in the storage class, but does set volumeSnapshotClassName in csi-driver-config for the same provisioner, the mutator can use that value as the default.

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: Sorry, this should be changed to `Delete the StorageClass annotations and ensure the corresponding setting in the StorageProfile is also cleared.`

ihcsim (dismissed):

brandboat:
Some updates here:

1. The VolumeSnapshotClass UI field has been changed from a dropdown menu to a text field, as suggested by @ihcsim. This avoids displaying unrelated VolumeSnapshotClass options that don't match the CSI driver.
2. If the user doesn’t set cdi.harvesterhci.io/storageProfileVolumeSnapshotClass in the storage class, but does specify volumeSnapshotClassName in the csi-driver-config for the same provisioner, the storage class mutator will use that value as the default—also suggested by @ihcsim.

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,127 @@
...
+      {"Block":["ReadWriteOnce","ReadOnlyMany"],"Filesystem":["ReadWriteOnce","ReadWriteMany"]}
+```
+
+The implementation consists of below main components:
+
+1. StorageClass Webhook Validation
+Extend the existing StorageClass validation webhook to validate the five new annotations. The webhook will perform pattern matching for filesystemOverhead, enum validation for cloneStrategy, existence checks for volumeSnapshotClass, and JSON schema validation for volumeModeAccessModes.
+2. New StorageClass Controller
+Introduce a new storageclass_controller specifically for watching for StorageClass changes and handling filesystemOverhead synchronization to the CDI object and synchronize cloneStrategy, volumeSnapshotClass, and volumeModeAccessModes annotations to the corresponding StorageProfile CRD.
+3. If the user doesn’t set `cdi.harvesterhci.io/storageProfileVolumeSnapshotClass` in the storage class, but does set `volumeSnapshotClassName` in `csi-driver-config` for the same provisioner, the storage class mutator can use that value as the default.
```
ihcsim: Sorry for the confusion. My previous suggestion was pertaining to the UI where instead of displaying a dropdown containing all the volume snapshot classes, the UI can auto-fill the text field with what it can find in the `csi-driver-config` (if the storage class doesn't already have the `cdi.harvesterhci.io/storageProfileVolumeSnapshotClass` annotation). Is it possible to make this a frontend logic? It's not necessary for the mutator to implement this.

brandboat: Thanks for the explanation, I've discussed with @a110605, and it's ok to add the logic in frontend.

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: We can provide the tooltip as follows:
  - `cdi.harvesterhci.io/filesystemOverhead`: Specifies the percentage of filesystem overhead to consider when calculating PVC size.
  - `cdi.harvesterhci.io/storageProfileCloneStrategy`: Defines the clone strategy to use for volumes created with this StorageClass.
  - `cdi.harvesterhci.io/storageProfileVolumeSnapshotClass`: Sets the Volume Snapshot Class name to be used when taking snapshots of PVCs under this StorageClass.
  - `cdi.harvesterhci.io/storageProfileVolumeModeAccessModes`: Specifies the default volume mode and access modes.
  
  c.c. @Vicente-Cheng, @WebberHuang1118, feel free to correct me if I missed anything, many thanks.

Vicente-Cheng (approved):
lgtm, thanks!

ihcsim (approved):

enhancements/20250522-expose-cdi-settings-in-ui.md
innobead: >Therefore, we still need to implement our own validation logic for these annotations defined in StorageClass.

minor: I see, but the validation will be the same as what CDI does. Not sure if we can use their validation using CDI as a lib to prevent duplicated implementation or future change in validation.

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,126 @@
...
+
+For LVM, the default values are:
+- volume mode is set to `Block`
+- access modes is set to `["ReadWriteOnce"]`
+
+This HEP preserves the same logic, but shifts it to the StorageClass mutator. The mutator will:
+- add annotations `cdi.harvesterhci.io/storageProfileCloneStrategy: "snapshot"` and `cdi.harvesterhci.io/storageProfileVolumeSnapshotClass: "longhorn-snapshot"` to LonghornV2 StorageClasses if both annotations are not set.
+- add `cdi.harvesterhci.io/storageProfileVolumeModeAccessModes: "{\"Block\":[\"ReadWriteOnce\"]}"` to LVM StorageClasses if the annotation is not already present.
+
+Note that Longhorn v2 will support CSI clone starting from v1.9.1, we need to set the default CDI clone strategy to `"clone"` in the storage class annotation `cdi.harvesterhci.io/storageProfileCloneStrategy`.
```
innobead, line 110: Unfortunately, the team can't make it work in 1.9.1 and will postpone this to 1.10. 

cc @derekbit @Vicente-Cheng

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,126 @@
...
+- clone strategty is set to `copy`
+- volume snapshot class is set to `longhorn-snapshot`
+
+For LVM, the default values are:
+- volume mode is set to `Block`
+- access modes is set to `["ReadWriteOnce"]`
+
+This HEP preserves the same logic, but shifts it to the StorageClass mutator. The mutator will:
+- add annotations `cdi.harvesterhci.io/storageProfileCloneStrategy: "snapshot"` and `cdi.harvesterhci.io/storageProfileVolumeSnapshotClass: "longhorn-snapshot"` to LonghornV2 StorageClasses if both annotations are not set.
+- add `cdi.harvesterhci.io/storageProfileVolumeModeAccessModes: "{\"Block\":[\"ReadWriteOnce\"]}"` to LVM StorageClasses if the annotation is not already present.
```
innobead, line 108: If an unsupported value has been set to the annotation, I assume it will be reverted to the default value supported by Longhorn v2 and LVM. For example, when setting RWX access mode to a LVM storage class, the value should be reverted to RWO. 

For any mutation change, to prevent silent change, it's better to at least have logs or even better to have event audits.

enhancements/20250522-expose-cdi-settings-in-ui.md
innobead: So we will have a global setting for filesystemOverhead this time?

enhancements/20250522-expose-cdi-settings-in-ui.md
```diff
@@ -0,0 +1,126 @@
...
+
+Note that Longhorn v2 will support CSI clone starting from v1.9.1, we need to set the default CDI clone strategy to `"clone"` in the storage class annotation `cdi.harvesterhci.io/storageProfileCloneStrategy`.
+
+### Test plan
+
+1. Create StorageClass with valid annotations and verify synchronization
```
innobead, line 114: I assume the synchronization should be one-way, starting from the annotations set to the storage class. If anything is set to the storage profile by users, it won't be synchronized back to the corresponding storage class.

innobead (approved):
In general, LGTM. A few minor comments.

innobead (approved):
In general, LGTM. A few minor comments.

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: We’ll block all unsupported values in the webhook (validator). In case something goes wrong—such as the webhook not functioning properly—the controller will also validate all values. If any unsupported value is detected, the controller will return an error and log the details accordingly.

enhancements/20250522-expose-cdi-settings-in-ui.md
brandboat: Yes, you're right.

enhancements/20250522-expose-cdi-settings-in-ui.md
innobead: No silent mutation here, sounds good.

PR #8390, opened 2025-05-30
Update storagenetwork check

w13915984028:
#### Problem:

When a cluster is newly installed and has no storagenetwork, if VM starts before Harvester POD creates the empty storagenetwork default setting, then Harvester POD stucks.

before this pr:

```
0. run any vm on a newly installed running cluster which has no storage-network setting

1. remove current webhook

kubectl delete ValidatingWebhookConfiguration harvester-validator

2. remove current sn setting
kubectl delete settings.harvesterhci.io storage-network

3. replace harvester-webhook pod to let it re-add the webhook
kubectl delete pod -n harvester-system harvester-webhook-8f77d684d-hn9b6

4. replace harvester pod, it will stuck on CrashLoopBackOff 
kubectl delete pod -n harvester-system harvester-5c4f7f555c-xt2kg

with error:

time="2025-05-30T08:26:31Z" level=fatal msg="failed to create harvester server: admission webhook \"validator.harvesterhci.io\" denied the request: Please stop all VMs before configuring the storage-network setting"
```

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8407

#### Test plan:

```
0. run any vm on a newly installed running cluster which has no storage-network setting

1. remove current webhook

kubectl delete ValidatingWebhookConfiguration harvester-validator

2. remove current sn setting
kubectl delete settings.harvesterhci.io storage-network

3. replace harvester-webhook pod to let it re-add the webhook
kubectl delete pod -n harvester-system harvester-webhook-8f77d684d-hn9b6

4. replace harvester pod, it will run smoothly

5. check webhook and storage-network sn

 kubectl get ValidatingWebhookConfiguration harvester-validator 
NAME                  WEBHOOKS   AGE
harvester-validator   1          6s  // newly created by new harvester-webhook pod



kubectl get settings.harvesterhci.io storage-network -oyaml
kind: Setting
metadata:
  creationTimestamp: "2025-05-30T10:30:09Z"  // newly created by harvester pod
  generation: 1
  name: storage-network
  resourceVersion: "415790"
  uid: 0e9f8bc6-40b3-4275-ae22-709c7e0ac629
status: {}

```

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -1343,7 +1343,7 @@ func (v *settingValidator) checkVCSpansAllNodes(config *storagenetworkctl.Config
 }
 
 func (v *settingValidator) checkStorageNetworkVlanValid(config *storagenetworkctl.Config) error {
-	if config.Vlan < 0 || config.Vlan > 4094 {
+	if config.Vlan > 4094 {
 		return fmt.Errorf("The valid value range for VLAN IDs is 0 to 4094")
 	}
 
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -1037,7 +1037,7 @@ func (v *settingValidator) validateStorageNetworkHelper(value string) error {
 
 	var config storagenetworkctl.Config
 	if err := json.Unmarshal([]byte(value), &config); err != nil {
-		return fmt.Errorf("failed to unmarshal the setting value, %v", err)
+		return fmt.Errorf("failed to unmarshal the setting value %v, %w", value, err)
 	}
 
 	if err := v.checkStorageNetworkVlanValid(&config); err != nil {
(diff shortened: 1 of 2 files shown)


w13915984028:
@mergify backport v1.5

votdev (dismissed):
LGTM

pkg/webhook/resources/setting/validator.go
```diff
@@ -1072,7 +1072,12 @@ func (v *settingValidator) validateStorageNetwork(setting *v1beta1.Setting) erro
 		return werror.NewInvalidError(err.Error(), settings.KeywordValue)
 	}
 
-	return v.checkStorageNetworkValueVaild()
+	// When a new setting is created, it's value and default are both empty, do not check storage-network usage
+	if setting.Value != "" || setting.Default != "" {
```
ihcsim, line 1076: How does `setting.Value` or `setting.Default` trigger the [`!isStopped` condition](https://github.com/harvester/harvester/blob/7a2ebdaf8521453e29efeefbcbaa07c678e2f5b2/pkg/webhook/resources/setting/validator.go#L1241-L1243)? Any chance that the cache might be stale?

w13915984028: There are two validators here:

OnChange, when setting.Value changes (non-empty to empty or vice versa), it triggers vm stopped check, where their root disk is backed by LH v1. The `cache stale` is not a big problem, it only triggers the webhook when value has changes, that should only happen when a new value is synced from apiserver to local cache. It might be a bit later, but won't push false value to local.

OnCreate, this PR is working on, skips vm stopped check when `setting.Value=="" && setting.Default == "" `, which is created by the Harvester POD. It is similar, onCreate doesn't suffer from cache stale case.

ihcsim (approved):

Issue #8383, opened 2025-05-29
[CI] Correlate backport PRs to the backport issue


innobead:
### Task request

Currently, people need to make a reference comment to associate a backport PR with its backport issue. Since backport PRs and issues follow certain conventions to understand their relationship, we should automate this human operation. 

Similar to https://github.com/longhorn/longhorn/blob/master/.github/workflows/backport-pr.yml

innobead:
cc @harvester/tpm @harvester/dev

Yu-Jack:
this one should backport, otherwise it won't take effect.

Yu-Jack:
Tested in:

- https://github.com/harvester/harvester/pull/9250#issuecomment-3359106221
- https://github.com/harvester/harvester/issues/9249#issuecomment-3359106187

PR #8400, opened 2025-06-03
chore: add link between backport issue and backport PR [skip ci]

Yu-Jack:
https://github.com/harvester/harvester/issues/8383

The original GA also has escape string issue. So, I use EOF to replace it as well.

Test Action: https://github.com/JackTestHook/test/actions/runs/15409902431/job/43359604325


innobead (approved):
LGTM.

Please also apply this workflow to other active repos including lib, component and addons, which are branched based. Might need to check if the branch naming convention is consistent. 

Can skip the UI ext repo, since the branch naming doesn't follow our general pattern. (cc @a110605 @houhoucoop )

PR #8412, opened 2025-06-03
Bump Longhorn v1.9.0 and golang v1.24.2

Vicente-Cheng:
#### Problem:
Bump Longhorn v1.9.0 for various issues.

#### Solution:
Bump Longhorn v1.9.0 for various issues.


Vicente-Cheng:
Please ignore the codeFactor, the error comes from the `TODO` comment and the complex method.

Vicente-Cheng:
Please ignore the codeFactor, the error comes from the TODO comment and the complex method.

And for the library bump, we now faced the issue with the conflict between kubevirt 1.5 and k8s v1.32.
We will check the library things later.

So, please help to review this PR, Thanks!

pkg/controller/master/backup/util.go
```diff
@@ -236,7 +236,9 @@ func getVMBackupMetadataFilePath(vmBackupNamespace, vmBackupName string) string
 
 func getBackupTargetHash(value string) string {
 	hash := sha256.New224()
-	io.Copy(hash, io.MultiReader(strings.NewReader(value)))
+	if _, err := io.Copy(hash, io.MultiReader(strings.NewReader(value))); err != nil {
+		return ""
```
ihcsim: This feels a bit dangerous. Since this method isn't widely used, can we consider changing the signature to return `(string, error)`? If not, at least log the error as warning.

ihcsim (commented):
Thanks for the cleanup. Just one comment. Also, what do you mean by the "library bump"? Are you referring to go 1.24?

pkg/controller/master/backup/util.go
WebberHuang1118: This was introduced in this PR: https://github.com/harvester/harvester/pull/7114. 

Maybe we can create a separate ticket to track this—WDYT? @ihcsim @Vicente-Cheng @FrankYang0529 @innobead

Vicente-Cheng: I thought I could address that in another commit.
Good suggestion @ihcsim

Vicente-Cheng:
> Thanks for the cleanup. Just one comment. Also, what do you mean by the "library bump"? Are you referring to go 1.24?

That means we should bump Longhorn/Kubevirt/K8s library.
But we encounter some conflicts between kubevirt/k8s, it depends on the upstream.

ihcsim (dismissed):

pkg/controller/master/backup/util.go
Vicente-Cheng: updated, please recheck

ihcsim (dismissed):

FrankYang0529 (dismissed):

WebberHuang1118 (dismissed):
LGTM, thanks for the PR.

ihcsim (approved):

FrankYang0529 (approved):

Issue #6855, opened 2024-10-23
[ENHANCEMENT] Set timeout for upgrade log infra scaffolding


starbops:
**Is your enhancement related to a problem? Please describe.**

If the user enables the upgrade log function on the upgrade dialog, a stage called "upgrade-log infra scaffolding" occurs at the very beginning of the Harvester upgrade. Such a stage may last a very long or, even worse, be stuck. This will block the entire upgrade flow and cause frustration.

For example, when the user had the `rancher-logging` add-on enabled and configured problematic (Cluster)Flows/Outputs, the logging operator broke and did not process further requests. [This was a known upstream issue](https://github.com/kube-logging/logging-operator/issues/1567) (no worries, it was fixed in newer releases). When the user tries to upgrade their Harvester cluster, the logging operator is unable to scaffold the infra that the upgrade log function needs, hence the aforementioned stuck.

**Describe the solution you'd like**

Add a default timeout value for the infra creation and continue to the next stage if it does timeout (for whatever reason). The controller should also reflect such an incident on the Upgrade/UpgradeLog custom resources.

**Describe alternatives you've considered**

Leave it jammed and wait for eventually human intervention.

**Additional context**

N/A

starbops:
@dharmit, do you have the capacity to devise a test plan for the enhancement? Thank you.

dharmit:
I surely have the capacity, but might need help reviewing the test plan. Will update here with my thoughts.

dharmit:
### Test plan

- Trigger a Harvester upgrade
- Ensure `UpgradeLog` infrastructure can't get ready (@starbops help me elaborate how this can be done)
- Default timeout to wait is 5 minutes, which can be adjusted as documented in https://github.com/harvester/docs/pull/808. Acceptable range is 5 to 20 minutes
  - Set a value outside the acceptable range and observe the default 5 minutes timeout being used by Harvester
- Upgrade would fail if the upgrade log infrastructure fails to scaffold within the timeout.

irishgordo:
this all looks good on v1.7.0-rc3 -> master-head
we can see when the timeout is adjusted down below the acceptable range everything does timeout properly and we see the upgrade failure
thanks again for this enhancement @dharmit ! 😄 

<img width="2664" height="2001" alt="Image" src="https://github.com/user-attachments/assets/81f39265-ca5e-4b64-8753-9bc7b6135e58" />
[...]

PR #8421, opened 2025-06-04
 Adds timeout for UpgradeLog's logging infrastructure 

dharmit:
#### Problem:

Explained in #6855 

#### Solution:

The timeout ensures that Upgrade is not stuck indefinitely on waiting
for logging infrastructure, i.e. fluentd and fluentbit, to come up. At
the end of the timer, we fail the Upgrade process by appropriately setting its `LogReady` condition.

User decides what they want to do next: fix the cause of failing setup of UpgradeLog infrastructure, or disable upgrade logging.
#### Related Issue(s):

#6855


pkg/controller/master/upgradelog/controller.go
```diff
@@ -108,6 +108,54 @@ type Image struct {
...
+			// check if UpgradeLog CR's InfraReady condition is still Unknown
+			if harvesterv1.InfraReady.IsUnknown(upgradeLog) {
+				logrus.Info("Timed out creating logging infrastructure")
+				setInfraReadyCondition(upgradeLog, corev1.ConditionFalse, "Timeout", "Timed out creating logging infrastructure")
+				_, err = h.upgradeLogClient.Update(upgradeLog)
+				if err != nil {
+					logrus.Errorf("Failed to update upgradeLog: %v", err)
+				}
+
+				// TODO: should we use upgradeClient instead of upgradeCache?
```
dharmit: @starbops please suggest. The reason I'm wondering about `upgradeClient` is that maybe `upgradeCache` would have outdated info after the 10 minutes ticker? Unless, of course, the cache is getting updated during this timeframe because other tasks might be running.

pkg/controller/master/upgradelog/controller.go
```diff
@@ -108,6 +108,54 @@ type Image struct {
...
+				upgrade, err = h.upgradeCache.Get(util.HarvesterSystemNamespaceName, upgradeLog.Spec.UpgradeName)
+				if err != nil {
+					if apierrors.IsNotFound(err) {
+						// TODO: is this even possible if UpgradeLog was found except manual deletion? What to do in such a case?
```
dharmit: @starbops please suggest.

w13915984028 (commented):
@dharmit Thanks for the PR, a better solution is:
1. write down the timestamp of starting logging infra onto the upgrade object, if it is not there
2. enqueue upgrade object, and compare the timestamp, if outdated, mark upgrade is fail
3. the timeout should be configurable

the benefit is:
when upgrade is deleted, you don't need to handle the dangling go thread which checks the logging infra ready

dharmit:
@w13915984028 thanks for the suggestion. I have pushed another commit that adheres to your recommendation. However, I haven't tested it yet and 30 seconds' timeout is set for the sake of testing purpose only.

dharmit:
@w13915984028 

> 1. write down the timestamp of starting logging infra onto the upgrade object, if it is not there
> 
> 2. enqueue upgrade object, and compare the timestamp, if outdated, mark upgrade is fail

I'm lost as to how to trigger the upgrade. I understand that a change needs to be made on upgrade CR for it, but where in the upgradelog flow do you suggest doing this?

> the benefit is: when upgrade is deleted, you don't need to handle the dangling go thread which checks the logging infra ready

The goroutine I had proposed in the earlier commit checked for existence of Upgrade CR and exited if it didn't find one. Hence, it won't dangle, no?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -162,9 +162,46 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+			harvesterv1.InfraReady.Reason(toUpdateUpgradeLog, "Timeout")
+			harvesterv1.InfraReady.Message(toUpdateUpgradeLog, "Timed out creating logging infrastructure")
+			_, err = h.upgradeLogClient.Update(toUpdateUpgradeLog)
+			if err != nil {
+				logrus.Errorf("Failed to update upgradeLog: %v", err)
+				return upgrade, nil
+			}
+		}
+
+		logrus.Info("Waiting for LogReady condition to be set")
```
w13915984028: add `EnqueueAfter` to loop self;  the controller is triggered by an `update` event, if no others update the upgrade CR, it will stuck here

dharmit: Thanks, Jian. I was unaware of `EnqueueAfter`. It makes more sense now.
I have pushed the change and still kept the old approach. If we are absolutely certain about Approach 3, I would remove the extra commit and open this up for review.

dharmit:
@w13915984028 @starbops I have opened the PR for review. PTAL when you can.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -162,9 +163,45 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			toUpdate.Status.UpgradeLog = upgradeLog.Name
 		}
 		harvesterv1.LogReady.CreateUnknownIfNotExists(toUpdate)
+		harvesterv1.LogReady.LastUpdated(toUpdate, time.Now().UTC().Format(time.RFC3339))
 		return h.upgradeClient.Update(toUpdate)
 	}
 
+	if harvesterv1.LogReady.IsUnknown(upgrade) {
```
w13915984028: @dharmit  Thanks.

Could you put this new block of code into an independent function? the current `OnChanged` is too big and needs to be splited step by step.

And, could `upgradeLogCondLogReadTimeout` be defined from the https://docs.harvesterhci.io/v1.5/advanced/index#upgrade-config, also with a valid range check like [1..20] minutes.

w13915984028: upgradeLogCondLogReadTimeout
->
upgradeLogCondLogReadyTimeout ?

dharmit: Thanks for the review. Pushed changes.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -968,3 +974,45 @@ func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMach
...
+	if err != nil {
+		logrus.Errorf("Failed to parse last updated time: %v", err)
+		return upgrade, nil
+	}
+	if time.Since(t) >= timeout {
+		toUpdate := upgrade.DeepCopy()
+		upgradeLog, err := h.upgradeLogCache.Get(util.HarvesterSystemNamespaceName, upgrade.Status.UpgradeLog)
+		if err != nil {
+			logrus.Errorf("Failed to get upgrade log: %v", err)
+			return upgrade, nil
```
w13915984028: Like L1011, set upgrade to infra false, avoid stucking here

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -968,3 +974,45 @@ func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMach
 		Do(ctx)
 	return res.Error()
 }
+
+// checkLogReadyCondition times out LogReady condition, thus moving on to the next step if upgradeLogCondLogReadyTimeout
+// time has passed since UpgradeLog's logging infrastructure setup started
+func (h *upgradeHandler) checkLogReadyCondition(upgrade *harvesterv1.Upgrade) (*harvesterv1.Upgrade, error) {
+	upgradeConfig, err := settings.DecodeConfig[settings.UpgradeConfig](settings.UpgradeConfigSet.Get())
+	if err != nil {
+		return upgrade, err
```
w13915984028, line 846: log an error, and use the default timeout value to continue

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -968,3 +974,45 @@ func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMach
...
+		_, err = h.upgradeLogClient.Update(toUpdateUpgradeLog)
+		if err != nil {
+			logrus.Errorf("Failed to update upgradeLog: %v", err)
+			return upgrade, nil
+		}
+		setLogReadyCondition(toUpdate, corev1.ConditionFalse, "Timeout", "Timed out creating logging infrastructure")
+		return h.upgradeClient.Update(toUpdate)
+	}
+
+	logrus.Info("Waiting for LogReady condition to be set")
```
w13915984028: change to `logrus.Debug` to reduce the log

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -968,3 +974,45 @@ func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMach
...
 }
+
+// checkLogReadyCondition times out LogReady condition, thus moving on to the next step if upgradeLogCondLogReadyTimeout
+// time has passed since UpgradeLog's logging infrastructure setup started
+func (h *upgradeHandler) checkLogReadyCondition(upgrade *harvesterv1.Upgrade) (*harvesterv1.Upgrade, error) {
+	upgradeConfig, err := settings.DecodeConfig[settings.UpgradeConfig](settings.UpgradeConfigSet.Get())
+	if err != nil {
+		return upgrade, err
+	}
+	timeout := upgradeConfig.LogReadyTimeout
```
w13915984028: add a range check again, if out of range, use the default value

for e.g. the webhook was removed/defected, an out of range value was injected

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -162,9 +163,14 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
 		}
 		harvesterv1.LogReady.CreateUnknownIfNotExists(toUpdate)
+		harvesterv1.LogReady.LastUpdated(toUpdate, time.Now().UTC().Format(time.RFC3339))
 		return h.upgradeClient.Update(toUpdate)
 	}
 
+	if harvesterv1.LogReady.IsUnknown(upgrade) {
+		return h.checkLogReadyCondition(upgrade)
+	}
+
```
w13915984028, line 159: Did you check: when logReady is timeouted, what's the upgrade (UI) status?

Will the upgrade be stopped or continued? if continuing, is the UI menu `download upgrade log` still available?

L174 needs to be refactored to add a check of `upgrade.Spec.LogEnabled` first, and handle the success, failure case

w13915984028 (commented):
Some questions, please also add test code on `upgrade_controller_test.go` to cover those cases you added. thanks.

pkg/controller/master/upgrade/upgrade_controller.go
dharmit: > use the default timeout value to continue

Is the rationale here to let the user/admin make necessary changes so that this step doesn't fail, and then after timeout duration check again if things are fixed?

w13915984028: as a minor enhancement, would prefer to block the upgrade main process as less as possbile;
it can return error and wait user to sovle it; or use the default value to continue;   the (default value) 5 mintues looks to be enough for most cases to get either success or failure result

pkg/controller/master/upgrade/upgrade_controller.go
dharmit: Done. Can you PTAL if I understood you correctly?

pkg/controller/master/upgrade/upgrade_controller.go
dharmit: Done. Again, can you PTAL if I understood you correctly?

pkg/controller/master/upgrade/upgrade_controller.go
dharmit: Done.

pkg/controller/master/upgrade/upgrade_controller.go
dharmit: Done.

pkg/controller/master/upgrade/upgrade_controller.go
dharmit: The upgrade does move further to the step of preloading images. But I didn't check the UI. I'll check this and report here.

dharmit:
@w13915984028 I might be overthinking/over-engineering here.

What's the preferred way of setting a default value? E.g., if the `UpgradeConfig.LogReadyTimeout` is not set, we want to default it to 5mins. Can you share an example from Harvester code that could help me understand how to accomplish this?

One way I know of is to implement a custom `UnmarshalJSON` that can be found in d6f4c24f5ace870b7138e67657d1f6db6098033b. Copy-pasting here for the record as that commit will be removed in future:
```go
func (u *UpgradeConfig) UnmarshalJSON(bytes []byte) error {
	uc := struct {
		preloadOption   ImagePreloadOption
		restoreVM       bool
		logReadyTimeout time.Duration
	}{}
	err := json.Unmarshal(bytes, &uc)
	if err != nil {
		return err
	}
	u.PreloadOption = uc.preloadOption
	u.RestoreVM = uc.restoreVM
	if uc.logReadyTimeout == 0*time.Second {
		u.LogReadyTimeout = 5 * time.Minute
	} else {
		u.LogReadyTimeout = uc.logReadyTimeout
	}
	return nil
}
```

w13915984028 (commented):
@dharmit 

Before this PR, the upgrade have two options:
(1) continue with `LogReady.IsTrue` which is set when log is ready
(2) continue with `LogReady.IsFalse` which is set when log is disabled/created failed

if log stucks, then it stucks

With your PR,
if log is time-outed, the better choice is to fail the whole upgrade (no real change upon the cluster happens), let user have chance to fix the failure, or disable the log and do upgrade again.

if you continue with `log enabled but failed`, all possible cases (e.g. log pvc is kept until user click `dismiss` from UI...) need to be added and even UI might need some changes.

WDYT?

w13915984028:
> @w13915984028 I might be overthinking/over-engineering here.
> 
> What's the preferred way of setting a default value? E.g., if the `UpgradeConfig.LogReadyTimeout` is not set, we want to default it to 5mins. Can you share an example from Harvester code that could help me understand how to accomplish this?
> 
> One way I know of is to implement a custom `UnmarshalJSON` that can be found in [d6f4c24](https://github.com/harvester/harvester/commit/d6f4c24f5ace870b7138e67657d1f6db6098033b). Copy-pasting here for the record as that commit will be removed in future:
...

@dharmit Please refer the apis defined on below links, they are purpose-based.

https://github.com/harvester/harvester/blob/ab75ccb7539d85ab321de1370c99683aa783c922/pkg/util/setting.go#L11

https://github.com/harvester/harvester/blob/ab75ccb7539d85ab321de1370c99683aa783c922/pkg/util/setting.go#L40

dharmit:
> if log stucks, then it stucks

In my understanding, it was THE problem that #6855 wanted to address.

> With your PR, if log is time-outed, the better choice is to fail the whole upgrade (no real change upon the cluster happens), let user have chance to fix the failure, or disable the log and do upgrade again.

IMO, it's a valid point! Instead of silently failing when logging infra fails to come up, we could inform the user who could then either:
1. Choose to re-trigger an upgrade with logging disabled
2. Fix the problem that caused logging infra to fail from coming up
    1. Reach out to fix the problem if they can't figure it out themselves.

> if you continue with `log enabled but failed`, all possible cases (e.g. log pvc is kept until user click `dismiss` from UI...) need to be added and even UI might need some changes.

Tbh, I'm not 100% aware of everything that happens under the hood. But, I could imagine dangling resources being left behind with this approach. Although I do need to do a proper testing of failure scenario as asked/discussed in https://github.com/harvester/harvester/pull/8421#discussion_r2174627463.

@starbops please chime in for failing the entire upgrade vs. silently putting it in the Upgrade CR. 😄

dharmit:
> @dharmit Please refer the apis defined on below links, they are purpose-based.
> 
> https://github.com/harvester/harvester/blob/ab75ccb7539d85ab321de1370c99683aa783c922/pkg/util/setting.go#L11
> 
> https://github.com/harvester/harvester/blob/ab75ccb7539d85ab321de1370c99683aa783c922/pkg/util/setting.go#L40

@w13915984028 can you PTAL if latest changes to validator.go and validator_test.go make sense? I'm skeptical because the validator is also mutating the value. https://github.com/harvester/harvester/blob/176a2a7e6653f7bf2c6e89455aee907cc10bb688/pkg/webhook/resources/setting/validator.go#L1742-L1745

pkg/webhook/resources/setting/validator.go
```diff
@@ -1737,6 +1738,15 @@ func validateUpgradeConfigFields(setting *v1beta1.Setting) error {
 		return fmt.Errorf("invalid image preload concurrency: %d", concurrency)
 	}
 
+	timeout := upgradeConfig.LogReadyTimeout
+	if timeout == 0*time.Second {
+		// this would be true if user didn't specify any value for UpgradeConfig.LogReadyTimeout
```
w13915984028: normally, validator does not change value;

here it could be:
```
if timeout == 0 {
  return nil
}
```

instead, when featching value from setting, the `0` is treated as the default value

if this field is not an int/int64, but *int/int64, then the `0` does not represent both empty value and normly input value

but take 0 as the default value here is not a big problem, just write down in the doc, when user never set it, or set it to 0 explcitly, the default value is used.

also refer:
https://docs.harvesterhci.io/v1.4/advanced/index#additional-guest-memory-overhead-ratio

`Valid values: "", "0" and from "1.0" to "10.0".`

dharmit: Thanks! Makes sense.

I made some more changes because letting `LogReadyTimeout` be of `time.Duration` type would have led to bad UX. User would have to specify a long ass value for it because it's in nanoseconds by default.

Taking cue from [support-bundle-timeout](https://docs.harvesterhci.io/v1.4/advanced/index/#support-bundle-timeout) I have changed it to `string` type and the conversion from `string` to `time.Duration` happens in the code.

starbops:
> @starbops please chime in for failing the entire upgrade vs. silently putting it in the Upgrade CR. 😄

I initially thought that silently failing the upgradelog and letting the upgrade continue was a good idea, but turns out what @w13915984028 suggested was sane. We should fail the entire upgrade procedure since no actual changes have been applied cluster/system-wide yet.

dharmit:
@starbops @w13915984028 I have added latest commit that fails the Upgrade process if timeout is reached. PTAL.

@w13915984028 I wonder how relevant https://github.com/harvester/harvester/pull/8421/#discussion_r2174627463 is now because we fail things when/if LogReadyTimeout is reached. What are your thoughts?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -831,3 +836,59 @@ func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMach
...
+		return upgrade, nil
+	}
+	if time.Since(t) >= timeoutDuration {
+		toUpdate := upgrade.DeepCopy()
+		upgradeLog, err := h.upgradeLogCache.Get(util.HarvesterSystemNamespaceName, upgrade.Status.UpgradeLog)
+		if err != nil {
+			logrus.Errorf("Failed to get upgrade log: %v", err)
+			setLogReadyCondition(toUpdate, corev1.ConditionFalse, "UpgradeLogGetFailed", "Failed to get upgrade log")
+			return h.upgradeClient.Update(toUpdate)
+		}
```
dharmit: This doesn't feel like the right thing to do for the situation. Can you PTAL and suggest any changes/improvement? Should we also do `setUpgradeCompletedCondition`? I could be totally wrong, but wonder if this will go into a loop of trying to get upgradeLog.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -831,3 +836,59 @@ func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMach
...
+
+		toUpdateUpgradeLog := upgradeLog.DeepCopy()
+		harvesterv1.InfraReady.SetStatus(toUpdateUpgradeLog, string(metav1.ConditionFalse))
+		harvesterv1.InfraReady.Reason(toUpdateUpgradeLog, "Timeout")
+		harvesterv1.InfraReady.Message(toUpdateUpgradeLog, "Timed out creating logging infrastructure")
+		_, err = h.upgradeLogClient.Update(toUpdateUpgradeLog)
+		if err != nil {
+			logrus.Errorf("Failed to update upgradeLog: %v", err)
+			return upgrade, nil
+		}
```
dharmit: Similar question here. Should we do `setUpgradeCompletedCondition` here as well?

dharmit (commented):
@w13915984028 @starbops couple of questions for you folks.

pkg/controller/master/upgrade/upgrade_controller.go
w13915984028: As you have already waited until the timeout, even if unluckily the upgradeLog was deleted by user or due to other reason, you can't get it, it is ok to fail the upgrade.

pkg/controller/master/upgrade/upgrade_controller.go
w13915984028: `return error` is better here, the reconciller will help you retry

the update failure is basically caused by two cases:
(1) the object was removed by others, when again, L874 will fail out;
(2) the object was updated by others, when again, it goes through

pkg/settings/settings_helper.go
```diff
@@ -168,11 +169,15 @@ type ImagePreloadOption struct {
 	Strategy PreloadStrategy `json:"strategy,omitempty"`
 }
 
+const UpgradeLogReadyTimeoutDefault = 5 * time.Minute
```
w13915984028: please put this const to pkg/util/peakvalues.go

and also define 1, 20 as const in above file, refer them

pkg/webhook/resources/setting/validator_test.go
```diff
@@ -1010,6 +1010,22 @@ func Test_validateUpgradeConfig(t *testing.T) {
...
+		{
+			name: "LogeReadyTimeout not in expected range",
+			args: &v1beta1.Setting{
+				ObjectMeta: metav1.ObjectMeta{Name: settings.UpgradeConfigSettingName},
+				Value:      `{"imagePreloadOption":{"strategy":{"type":"parallel","concurrency":2}},"logReadyTimeout": "1"}`,
+			},
+			expectedErr: false,
+		},
+		{
+			name: "LogReadyTimeout is in expected range",
```
w13915984028: L1022: is `21` in expected range?

please also add a new case where a value like `10` is in range

w13915984028 (commented):
A few questions, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -149,9 +149,14 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			toUpdate.Status.UpgradeLog = upgradeLog.Name
 		}
 		harvesterv1.LogReady.CreateUnknownIfNotExists(toUpdate)
+		harvesterv1.LogReady.LastUpdated(toUpdate, time.Now().UTC().Format(time.RFC3339))
 		return h.upgradeClient.Update(toUpdate)
 	}
 
+	if harvesterv1.LogReady.IsUnknown(upgrade) {
```
w13915984028: L156-L160 should be updated as:
```
	if upgrade.Spec.LogEnabled && harvesterv1.LogReady.IsUnknown(upgrade) {
			return h.checkLogReadyCondition(upgrade)
	}

	if harvesterv1.UpgradeCompleted.IsUnknown(upgrade) && harvesterv1.ImageReady.GetStatus(upgrade) == "" {
	
	}
```

In current code, if upgrade is failed, L160 seems will still run.

dharmit: > if upgrade.Spec.LogEnabled && harvesterv1.LogReady.IsUnknown(upgrade)

Thanks for catching that bug right away! 🙇🏽‍♂️

pkg/webhook/resources/setting/validator_test.go
dharmit: Nope, it isn't. Fixed. Also added the test for 10, but removed for 1.

pkg/controller/master/upgrade/upgrade_controller.go
dharmit: Makes sense. Thanks for validating. Done.

pkg/controller/master/upgrade/upgrade_controller.go
dharmit: Thanks for explaining. Makes sense. Done.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -831,3 +836,61 @@ func (h *upgradeHandler) startVM(ctx context.Context, vm *kubevirtv1.VirtualMach
...
+	}
+
+	timeoutStr := upgradeConfig.LogReadyTimeout
+	timeout, err := strconv.Atoi(timeoutStr)
+	if err != nil {
+		return upgrade, fmt.Errorf("invalid value for image preload timeout: %s", timeoutStr)
+	}
+	timeoutDuration := time.Duration(timeout) * time.Minute
+
+	if timeoutStr == "" || timeoutDuration < 1*time.Minute || timeoutDuration > 20*time.Minute {
```
w13915984028: 1, 20 are still hard-coded, the logrus.Warnf can also use const to format

pkg/settings/settings.go
```diff
@@ -59,7 +59,7 @@ var (
 	HarvesterCSICCMVersion = NewSetting(HarvesterCSICCMSettingName, `{"harvester-cloud-provider":">=0.0.1 <0.3.0","harvester-csi-provider":">=0.0.1 <0.3.0"}`)
 	NTPServers             = NewSetting(NTPServersSettingName, "")
 	WhiteListedSettings    = []string{ServerVersionSettingName, DefaultStorageClassSettingName, HarvesterCSICCMSettingName, DefaultVMTerminationGracePeriodSecondsSettingName}
-	UpgradeConfigSet       = NewSetting(UpgradeConfigSettingName, `{"imagePreloadOption":{"strategy":{"type":"sequential"}}, "restoreVM": false}`)
+	UpgradeConfigSet       = NewSetting(UpgradeConfigSettingName, `{"imagePreloadOption":{"strategy":{"type":"sequential"}}, "restoreVM": false, "logReadyTimeout": "5"}`)
```
w13915984028, line 62: use fmt.Sprintf... to refer the defined default value

pkg/webhook/resources/setting/validator.go
```diff
@@ -1834,6 +1835,22 @@ func validateUpgradeConfigFields(setting *v1beta1.Setting) error {
...
+		return nil
+	}
+	timeout, err := strconv.Atoi(timeoutStr)
+	if err != nil {
+		return fmt.Errorf("invalid value for image preload timeout: %s", timeoutStr)
+	}
+	timeoutDuration := time.Duration(timeout) * time.Minute
+	// If the user set LogReadyTimeout to a value not in the permissible range [1-20 minutes], we return an error
+	if timeoutDuration < util.MinUpgradeLogReadyTimeout || timeoutDuration > util.MaxUpgradeLogReadyTimeout {
+		return fmt.Errorf("invalid logReadyTimeout must be between 1 to 20 minutes, given: %d", timeoutDuration)
```
w13915984028: update fmt.Errorf

pkg/webhook/resources/setting/validator_test.go
```diff
@@ -1010,6 +1010,22 @@ func Test_validateUpgradeConfig(t *testing.T) {
 			},
 			expectedErr: false,
 		},
+		{
+			name: "LogeReadyTimeout not in expected range",
```
w13915984028: name: ... is in expected range

pkg/webhook/resources/setting/validator_test.go
w13915984028: name: ... is not in expected range

w13915984028 (commented):
A few last updates, thanks.

pkg/settings/settings.go
dharmit: That's leading to cyclic imports. 😕

pkg/webhook/resources/setting/validator_test.go
dharmit: Weird, I thought I already did https://github.com/harvester/harvester/pull/8421#discussion_r2190304745.

pkg/settings/settings.go
w13915984028: Oops, then skip it now.

pkg/webhook/resources/setting/validator.go
dharmit: PTAL if I understood your feedback correctly.

w13915984028: cool, please squash the commits to prepare for final merging, thanks.

dharmit: Done.

w13915984028 (dismissed):

dharmit:
@starbops do you want to take a look at this, or should I go ahead and merge it?

dharmit:
> @starbops do you want to take a look at this, or should I go ahead and merge it?

@innobead do we need extra review on this beyond what @w13915984028 already did? His was an exhaustive review, and we made a bunch of changes to the code and approach based on it.

If that seems enough, I request we merge this PR. Thanks.

starbops (approved):
LGTM! Thanks for the great enhancement 👍 @dharmit

PR #8424, opened 2025-06-05
feat: use sockets stands for cpu

FrankYang0529:
#### Problem:
Our client implementation sets cpu value to `cores`. However, the cpu hot-plug feature needs to patch value on `sockets`, not `cores`. It's better to mutate cpu value to `sockets` in backend, so we don't need to change each client side.

#### Solution:
Use mutator to patch cpu value to sockets.

#### Related Issue(s):

#### Test plan:

* On UI, create a new VM with 2 CPU, the mutator change the value to sockets.
```
          cores: 1
          sockets: 2
          threads: 1
```
* On UI, update the VM to 3 CPU, the mutator change the value to sockets.
```
          cores: 1
          sockets: 3
          threads: 1
```

The UI shows the CPU based on `cores`, so you have to use YAML to validate it.

### DIFF
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -79,7 +79,7 @@ func (m *vmMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patch
 		return nil, err
 	}
 
-	return patchOps, nil
+	return patchCPUSocket(vm, patchOps), nil
 }
 
 func (m *vmMutator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) (types.PatchOps, error) {
@@ -121,7 +121,7 @@ func (m *vmMutator) Update(_ *types.Request, oldObj runtime.Object, newObj runti
 		return nil, err
 	}
 
-	return patchOps, nil
+	return patchCPUSocket(newVM, patchOps), nil
 }
 
 func needUpdateRunStrategy(oldVM, newVM *kubevirtv1.VirtualMachine) (bool, error) {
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -534,6 +534,44 @@ func (m *vmMutator) patchTerminationGracePeriodSeconds(vm *kubevirtv1.VirtualMac
...
+	}
+	if sockets == 0 {
+		sockets = 1
+	}
+	if threads == 0 {
+		threads = 1
+	}
+
+	patchSocket := false
+	if cores != 1 {
```
w13915984028, line 553: if original cores == "2", here it is also replaced to "1", is it expected? similar question on others

or:
```
if vm.Spec.Template.Spec.Domain.CPU.Cores == 0 {
    patchOps = append
}
```

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -534,6 +534,44 @@ func (m *vmMutator) patchTerminationGracePeriodSeconds(vm *kubevirtv1.VirtualMac
...
+
+	if threads != 1 {
+		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/template/spec/domain/cpu/threads", "value": 1}`)
+		patchSocket = true
+	}
+
+	if patchSocket {
+		// Calculate sockets based on cores and threads.
+		// If client sets sockets directly, we don't need to patch it.
+		// If client sets cores or threads, the client expects the cpu is same as cores or threads.
```
w13915984028, line 566: could we convert those examples to test cases? thanks.

w13915984028 (commented):
Please check the questions, thanks.

pkg/webhook/resources/virtualmachine/mutator.go
FrankYang0529: This is expected, because we want to use `sockets` to represent all CPU.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -534,6 +534,44 @@ func (m *vmMutator) patchTerminationGracePeriodSeconds(vm *kubevirtv1.VirtualMac
...
+
+	if cores == 0 {
+		cores = 1
+	}
+	if sockets == 0 {
+		sockets = 1
+	}
+	if threads == 0 {
+		threads = 1
+	}
```
ihcsim, line 550: I think KubeVirt can already do this for us [here](https://github.com/kubevirt/kubevirt/blob/7a7199bbfa83b1f7585c607289644956cc186f66/pkg/defaults/defaults.go#L364-L387).

E.g., I run `k apply -f https://raw.githubusercontent.com/kubevirt/kubevirt/refs/heads/main/examples/vm-cirros.yaml` without setting any CPU topology in the VM, the VMI will be started with core, socket and thread set to 1.

pkg/webhook/resources/virtualmachine/mutator.go
ihcsim: I thought the only place where the CPU is mapped to the socket is the UI. If user wants to change the CPU topology via YAML, I don't see any reasons to disallow that. Or if we really don't want user to mess with cores and threads, then we can consider using the validator to reject the request. I think the validator is easier to reason about compared to trying to reset the user-provided values to 1.

w13915984028: There is `DedicatedCPUPlacement` case, if the cores is replaced to `1` all the time, will it break CPU pinning?  In kubevirt code, there is also processing related to CPU pinning. cc @brandboat, please also double check, thanks.

```
func isDedicatedCPU(vm *kubevirtv1.VirtualMachine) bool {
	return vm.Spec.Template.Spec.Domain.CPU != nil && vm.Spec.Template.Spec.Domain.CPU.DedicatedCPUPlacement
}
```

brandboat: > There is DedicatedCPUPlacement case, if the cores is replaced to 1 all the time, will it break CPU pinning? In kubevirt code, there is also processing related to CPU pinning. cc @brandboat, please also double check, thanks.

No it won't, I just test the VM with the following settings
```
      domain:
        cpu:
          cores: 1
          dedicatedCpuPlacement: true
          sockets: 2
          threads: 1
```
everything is ok.
```
qemu@t:/> virsh dumpxml default_t | awk "/<cputune>/,/<\/cputune>/"
  <cputune>
    <vcpupin vcpu='0' cpuset='1'/>
    <vcpupin vcpu='1' cpuset='2'/>
  </cputune>
```

w13915984028: BTW, In the hot-plug + cpu pin scenario, there is a chance that when expanding vCPUs and there is no target node meeting the required dedicated cpus, then at which stage will it fail? trigger hot-plug, trigger migration, migration is not scheduable? thanks.

brandboat: > BTW, In the hot-plug + cpu pin scenario, there is a chance that when expanding vCPUs and there is no target node meeting the required dedicated cpus, then at which stage will it fail? trigger hot-plug, trigger migration, migration is not scheduable? thanks.

I'm not familiar with the hot-plug part, but from what I know, if the target node doesn't have enough dedicated CPUs, the virt-launcher pod will just stay pending until the node has enough resources to continue.

FrankYang0529:
@ihcsim @w13915984028 @brandboat Thanks for the review. I will close this PR and leave this to UI.

BTW, I will add another validation to hotplug API check. If `cores` or `threads` are not `1`, users cannot use the `cpuAndMemoryHotplug` action, because the result is not expected. For example, a VM with `{"cores": 2, "sockets": 1, "threads": 1}` and would like to use `cpuAndMemoryHotplug` action to update CPU to 4. However, only `sockets` can be changed, so the result will be `{"cores": 2, "sockets": 4, "threads": 1}`. The real vCPU will be `2*4*1`, not 4.

FrankYang0529:
Followup PR: https://github.com/harvester/harvester/pull/8472

Issue #8298, opened 2025-05-15
[BUG] upgrade stuck in Pre-draining which VM not migrating to other nodes


lanfon72:
### Describe the Bug

the node stuck in pre-draining status, it's waiting VM migrate to other node, but there haven't any migration triggered.

![Image](https://github.com/user-attachments/assets/3003f3b5-68e5-4bff-9bb5-868415904967)
![Image](https://github.com/user-attachments/assets/790a8234-ce80-40f9-85a5-61fc31c04353)

![Image](https://github.com/user-attachments/assets/83ed3ca9-feab-44ad-a141-3350a5ed6749)
![Image](https://github.com/user-attachments/assets/4a07402f-a956-43cc-a61b-771eb9d62e7b)

### To Reproduce

upgrade automation of 2 staging upgrade: 1.4.1 -> 1.4.2 -> 1.4.3-RC3

### Expected Behavior

VM should be migrated to other nodes

### Support Bundle for Troubleshooting

envrionment kept for debugging.

### Environment

- Harvester version: `v1.4.3-rc3` (upgrade from `v1.4.2`)
- Impacted VM: `upgrade-repo-hvst-upgrade-s2mk7`
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Baremtal DL360G9 4 nodes
- Rancher version: N/A

lanfon72:
add `severity/2` as I think it could be resolved by manually migrate the VM to other nodes (workaround).

Vicente-Cheng:
Share the information from @brandboat,

Caused by the logs
```
internal error: QEMU unexpectedly closed the monitor (vm='harvester-system_upgrade-repo-hvst-upgrade-ffv4b'): 2025-05-15T03:51:45.401119Z qemu-system-x86_64: error: failed to set MSR 0x38f to 0x7000000ff","pos":"qemuProcessReportLogError:1926","subcomponent":"libvirt","thread":"83","timestamp":"2025-05-15T03:51:45.403000Z"
```

Related bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1439078
Workaround: Restart the affected VM

Vicente-Cheng:
Some additional informations here.

We saw the VM CPU information like below:
```
      domain:
        cpu:
          cores: 1
          model: host-passthrough
          sockets: 1
          threads: 1
```

and it will make the qemu use the CPU parameter as below:
```
[...]

PR #8425, opened 2025-06-05
Fix upgrade stuck caused by upgrade repo VM

brandboat:
#### Problem:
The upgrade repo VM uses the CPUModeHostPassthrough setting, which may cause live migration to fail if the source and target nodes have different CPU settings. In such cases, QEMU might raise an error during migration, leading to the upgrade process getting stuck. See https://bugzilla.redhat.com/show_bug.cgi?id=1439078 for more details.

#### Solution:

This patch introduces a new flag `--exclude-repo-vm` to the VM live migration detector (default: false). During the upgrade process, the detector will no longer skip upgrade repo VMs. Instead, it will shut it down before the node is upgraded to avoid migration failure.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8298

#### Test plan:
1. Prepare a Harvester cluster with at least two nodes using the built ISO image from this PR (for QAs, please use the master-head ISO image)
1. Trigger the upgrade with the same ISO image
1. Monitor the upgrade progress and the upgrade repo VM
1. The upgrade should succeed with the upgrade repo VM being shut down during the time.

### DIFF
--- a/cmd/upgradehelper/cmd/vmlivemigratedetector/vmlivemigratedetector.go
+++ b/cmd/upgradehelper/cmd/vmlivemigratedetector/vmlivemigratedetector.go
@@ -13,7 +13,8 @@ import (
 )
 
 var (
-	shutdown bool
+	shutdown      bool
+	excludeRepoVM bool
 )
 
 var vmLiveMigrateDetectorCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


brandboat:
This was included in https://github.com/harvester/harvester/pull/8526

PR #8428, opened 2025-06-06
change machine type order 

ibrokethecloud:
#### Problem:

PR https://github.com/harvester/harvester/pull/8336 returns different machine types based on underlying host architecture.

The harvester UI picks up the first machine type being returned as the default, which causes issues as pc-q35 seems to be less common as the q35 machine type.

#### Solution:

The PR introduces a minor fix to change the order in which machine types are returned.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/5364

### DIFF
--- a/pkg/api/cluster/formatter.go
+++ b/pkg/api/cluster/formatter.go
@@ -116,7 +116,7 @@ func generateMachineTypes() ([]byte, error) {
 	var machineTypes []string
 	switch runtime.GOARCH {
 	case "amd64":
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (approved):
thanks for helping with that!

a110605 (approved):
cc @houhoucoop

innobead (approved):

ibrokethecloud:
@mergifyio backport v1.5

Issue #5619, opened 2024-04-17
[BUG] Harvester/kubevirt and high mem gpus with a 32 bit BAR memory slot


mitchty:
**Describe the bug**

This may or may not be specific to Nvidia A30 Gpus with 32GiB of ram.

Tested with Harvester 1.2.1 but the bug likely exists in 1.3.0. Note this could be bugfixed in 1.3.x not 1.2.1 as there is a workaround I'll describe.

The bug however is when you pass through the aforementioned gpu to a vm and for 1.2.1 ensure limits == requests on memory, the nvidia kernel module will fail to attach like so:

The device in question and note, it is the only device in this iommu group as well so no other devices can nor should be attached to the vm.

![image](https://github.com/harvester/harvester/assets/611923/4afa8910-6138-416c-ba0f-15a6670578e7)

```
3b:00.0 3D controller [0302]: NVIDIA Corporation GA100GL [A30 PCIe] [10de:20b7] (rev a1)
	Subsystem: NVIDIA Corporation Device [10de:1532]
	Flags: bus master, fast devsel, latency 0, IRQ 255, NUMA node 0
	Memory at ab000000 (32-bit, non-prefetchable) [size=16M]
	Memory at 38a800000000 (64-bit, prefetchable) [size=32G]
	Memory at 38b810000000 (64-bit, prefetchable) [size=32M]
	Capabilities: [60] Power Management version 3
	Capabilities: [68] #00 [0080]
	Capabilities: [78] Express Endpoint, MSI 00
	Capabilities: [c8] MSI-X: Enable- Count=6 Masked-
	Capabilities: [100] Virtual Channel
	Capabilities: [250] Latency Tolerance Reporting
	Capabilities: [258] L1 PM Substates
	Capabilities: [128] Power Budgeting <?>
	Capabilities: [420] Advanced Error Reporting
	Capabilities: [600] Vendor Specific Information: ID=0001 Rev=1 Len=024 <?>
	Capabilities: [900] #19
	Capabilities: [bb0] #15
	Capabilities: [bcc] Single Root I/O Virtualization (SR-IOV)
	Capabilities: [c14] Alternative Routing-ID Interpretation (ARI)
	Capabilities: [c1c] #26
	Capabilities: [d00] #27
	Capabilities: [e00] #25
	Kernel driver in use: vfio-pci
```

So basically this gpu appears to be getting mapped only as the first 32 bit iommu address space you see in the lspci -nnk output and that is plainly not going to work.

[...]

PR #8429, opened 2025-06-06
introduce HEP for sidecar support [SKIP CI]

ibrokethecloud:
#### Problem:

Harvester by default does not enable the Kubevirt sidecar feature gate.
Users manually enabling this gate run into upgrade issues where the harvester managed chart cannot be updated since it is out of sync due to the extra feature gate.

#### Solution:

PR introduces the HEP for sidecar enablement.

#### Related Issue(s):

https://github.com/kubevirt/kubevirt/issues/11093

### DIFF
--- a/enhancements/20250606-sidecar-support.md
+++ b/enhancements/20250606-sidecar-support.md
@@ -1,0 +1,92 @@
+# Sidecar support
+
+Enable Kubvirt sidecar support to allow fine tuning of the Kubevirt generated domain definitions
+
+## Summary
+
+This enhancement aims to provide a generic way to tweak Kubevirt generated domain definitions to allow fine tuning of limited set of settings.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/5619
+
+## Motivation
+
+### Goals
+
+To provide a generic way to leverage Harvester built and 3rd party sidecars with Harvester
+
+### Non-goals [optional]
+
+* Instructions for building sidecars
+* Distribution mechanism for 3rd party sidecars
+
+## Proposal
+
+We will introduce the following changes
+* enable the sidecar feature gate
+* setting to allow cluster administrators to whitelist sidecars for usage within the cluster
+
+### User Stories
+
+#### User wants to tweak kubevirt generated domain definition
+
+Users attempting to passthrough GPUs/vGPUs with more than 24G of memory to VM's booted in EFI mode, run into BAR resizing issues. 
+The default mmio aperture on in the firmware is not large enough to map the GPU/vGPU memory.
+This results in the OS being unable to leverage the hardware. This can be addressed by passing additional firmware arguments to tweak the mmio aperture size in the guest.
+
+Kubevirt does not directly expose this setting, and the only possible solution is to leverage a sidecar to patch the VM domain definition to add additional arguments.
+
+### User Experience In Detail
+Users will be expected to edit the VM definition as a yaml and apply the sidecar specfic annotations to their VM.
+
+For example the following annotations are needed to inject a sidecar and any additional arguments that the sidecar looks up on boot
+```
+    hooks.kubevirt.io/hookSidecars: '[{"args": ["--version", "v1alpha2"], "image":
+      "registry:5000/kubevirt/example-hook-sidecar:devel"}]'
+    smbios.vm.kubevirt.io/baseBoardManufacturer: Radical Edward
+```
+The VM will boot with the sidecar and apply the sidecar will apply the VM domain definition changes.
+
+### API changes
+
+#### Setting
+We will have a new setting `whitelisted-sidecars`
+
+```
+apiVersion: harvesterhci.io/v1beta1
+kind: Setting
+metadata:
+  name: whitelisted-sidecars
+default: "sidecarimage1:tag,sidecarimage2:tag"
+value: ""  
+```
+
+Cluster admins can define additional sidecars by passing them via the `value` field.
+
+## Design
+
+### Implementation Overview
+
+The implementation will consist of the following:
+* A new setting `whitelisted-sidecars` which the cluster admins can control at a cluster level which sidecar images can be used to edit the cluster
+
+* Sidecar validating webhook in harvester. The webhook will watch the VM objects and check if the vm contains the sidecar injection annotation `hooks.kubevirt.io/hookSidecars`. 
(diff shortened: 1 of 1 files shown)


enhancements/20250606-sidecar-support.md
```diff
@@ -0,0 +1,92 @@
...
+```
+The VM will boot with the sidecar and apply the sidecar will apply the VM domain definition changes.
+
+### API changes
+
+#### Setting
+We will have a new setting `whitelisted-sidecars`
+
+```
+apiVersion: harvesterhci.io/v1beta1
```
w13915984028, line 57: add an `eanbled` field? some users might not plan to control with while-list

enhancements/20250606-sidecar-support.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/5619
+
+## Motivation
+
+### Goals
+
+To provide a generic way to leverage Harvester built and 3rd party sidecars with Harvester
```
w13915984028, line 17: upstream also supports to run shell/python from configmap, it is easier; do we plan to add it?
https://kubevirt.io/user-guide/user_workloads/hook-sidecar/#go-python-shell-pick-any-one

enhancements/20250606-sidecar-support.md
```diff
@@ -0,0 +1,92 @@
...
+
+* Sidecar validating webhook in harvester. The webhook will watch the VM objects and check if the vm contains the sidecar injection annotation `hooks.kubevirt.io/hookSidecars`. 
+  It will then extract the image name specified in the annotation, and validate the same against the whitelist provided by the `whitelisted-sidecars` setting.
+
+
+### Test plan
+* Install new build of harvester
+* Edit the settign `whitelisted-sidecars` to provide value of a custom image
+* Create a VM with the `hooks.kubevirt.io/hookSidecars: '[{"image":"registry:5000/kubevirt/example-hook-sidecar:devel"}]'` referring to the whitelisted sidecar
+* Start VM, and check it has 3 contains in the virt-launcher pod. One of these containers should be the sidecar container using the image specificed via the `hooks.kubevirt.io/hookSidecars` annotation.
```
w13915984028: contains -> containers

w13915984028 (commented):
Thanks for introducing this feature, a few questions.

enhancements/20250606-sidecar-support.md
ibrokethecloud: the plan is to use a `whitelist` to ensure users can control sidecars used cluster wide. With the sidecar the user will have access to core libvirt functionality not to mention devices attached to the launcher pod.

enhancements/20250606-sidecar-support.md
ibrokethecloud: The expectation is that the user will whitelist the generic image and then consume the configmap via the generic image

enhancements/20250606-sidecar-support.md
w13915984028: the `whitelist` could be an option, to enable/disable whilelist control with a list;  if disabled, user can use any image

the consideration is:

some users tend to use terraform or others to operate the cluster, the explicit `whilelist` means addtional coding for them

ibrokethecloud: the setting is a means to ensure secure way of consuming sidecars. The idea is only a cluster admin can control which sidecar is allowed to operate. This will ensure only trusted sidecars are allowed in the cluster. This also helps from a supportability perspective since we will be providing a predefined list of sidecars available for consumption.

w13915984028: the idea does not conflict with the option to disable it, when users have the full control of cluster on hand  e.g. in an air-gapped env, the cluster users are from internal, the admin is free to disable this list and allow the trusted users to run the expected images.

starbops (commented):
So, by default, we will enable the feature gate and provided a white list of sidecar images with an empty value, which means, by default, there will be no images allowed. Am I understand it correctly?

ibrokethecloud:
@starbops we will have a sidecar to allow mmio aperture resizing to go with this feature. This is needed to ensure GPUs/vGPUs with more than 24G memory can work correctly in the VM.

WebberHuang1118:
> @starbops we will have a sidecar to allow mmio aperture resizing to go with this feature. This is needed to ensure GPUs/vGPUs with more than 24G memory can work correctly in the VM.

Hi @ibrokethecloud,

Could you provide an example of the sidecar annotation used to ensure BAR resizing for GPUs/vGPUs?

ibrokethecloud:
@WebberHuang1118 would be something like 

```
annotations:
  hooks.kubevirt.io/hookSidecars: '["image":"rancher/harvester-vgpu-support:dev"]'
  harvester.mmio64mb: '65535'

```

The sidecar looks up the vmi object, and if it finds the key `harvester.mmio.value` would patch the libvirt domain with qemu extra args `-fw_cfg name=opt/ovmf/X-PciMmio64Mb,string=$valueFromAnnotation`

ibrokethecloud:
not needed anymore since the issue has been identified to a efi/bios firmware bug

dsmithbauer:
> not needed anymore since the issue has been identified to a efi/bios firmware bug

@ibrokethecloud can you provide more information here? Link to the firmware bug maybe? I see that #8757 is back porting a bump to the KubeVirt version in Harvester 1.5.2. Does the newer KubeVirt version fix the EFI/PCI MMIO issue #5619 - without adding full sidecar option support?

ibrokethecloud:
@dsmithbauer yes the issue was ovmf/seabios firmware being repackaged in the SLES repos

This has been fixed in our SLES repos and Kubevirt 1.5.2 and 1.4.1 images being used by Harvester will have the fix. 

Fix is available in v1.6-head images in case you wish to test and provide feedback

dsmithbauer:
> @dsmithbauer yes the issue was ovmf/seabios firmware being repackaged in the SLES repos
> 
> This has been fixed in our SLES repos and Kubevirt 1.5.2 and 1.4.1 images being used by Harvester will have the fix.
> 
> Fix is available in v1.6-head images in case you wish to test and provide feedback

Thanks for the additional information! I will test as soon as I am able to upgrade the environment.

innobead:
Closing this first, as we don't need this for now.

PR #8452, opened 2025-06-10
feat: add max-hotplug-ratio setting

FrankYang0529:
#### Solution:
There is a `maxHotplugRatio` configuration to control all VM's `maxSockets` and `maxGuest` based on `sockets` and `guest`. The default value is 4. It's better to let user to control this.

#### Related Issue(s):

#### Test plan:
1. Update the `max-hotplug-ratio` to `-1` and it cannot be applied.
2. Update the `max-hotplug-ratio` to `0` and it cannot be applied.
3. Update the `max-hotplug-ratio` to `non-int` and it cannot be applied.
4. Update the `max-hotplug-ratio` to `1` and it can be applied.
5. Update the `max-hotplug-ratio` to `4` and it can be applied.
6. Remove value in `max-hotpug-ratio` and it uses default value `4` in KubeVirt.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -1676,7 +1676,7 @@ func validateMaxHotplugRatioHelper(field, value string) error {
 	if value == "" {
 		return nil
 	}
-	num, err := strconv.Atoi(value)
+	num, err := strconv.ParseUint(value, 10, 32)
 	if err != nil {
 		return fmt.Errorf("failed to parse %s: %v", field, err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -1656,3 +1659,26 @@ func (v *settingValidator) isSingleNode() (bool, error) {
 	}
 	return len(nodes) == 1, nil
 }
+
+func validateMaxHotplugRatio(setting *v1beta1.Setting) error {
+	if err := validateMaxHotplutRatioHelper(settings.KeywordDefault, setting.Default); err != nil {
```
ihcsim:
```suggestion
	if err := validateMaxHotplugRatioHelper(settings.KeywordDefault, setting.Default); err != nil {
```

ihcsim (commented):
LGTM - just one minor typo.

ihcsim (dismissed):

ihcsim (approved):

Issue #7358, opened 2025-01-13
[FEATURE] Volume resizing inside guest clusters


bk201:
**Is your feature request related to a problem? Please describe.**

Related to https://github.com/harvester/harvester/issues/2811, but the resizing request is from the guest clusters. Only for RWO. 

**Describe the solution you'd like**

Providing volume resizing via `harvester-csi-driver`.

**Describe alternatives you've considered**

**Additional context**
ref: https://github.com/harvester/harvester/issues/8076

innobead:
Added to `planning` milestone w/ P0.

hoo29:
~Whilst this gets released, is there anyway currently to do resizing of PVCs without having to delete them?~

edit: scale down workload to zero, resize PVC, scale up workload, run resize2fs on VM which has the volume mounted

lanfon72:
Verified this feature been partially applied as following, move back to implement.

- [x] Able to resize volume when PVC is created by Longhorn v1
- [x] Unable to resize volume when PVC is created by Longhorn v2
- [ ] Able to resize volume when PVC is created by 3rd party LVM CSI

1. Tested with **TopoLVM**, the volume stuck in `Pending` status as following
<img width="1583" height="753" alt="Image" src="https://github.com/user-attachments/assets/05d48035-1291-4bea-b629-daea2a4146f2" />

The volume showing **RWO** on guest cluster (what we going to create)
<img width="909" height="712" alt="Image" src="https://github.com/user-attachments/assets/2536409e-2f6a-4d88-ac51-e42ec300ad77" />

But showing **RWM** on Harvester
<img width="967" height="452" alt="Image" src="https://github.com/user-attachments/assets/a24491a8-9dab-4736-bfda-f62c7d6694c1" />

The default of **volumeBindingMode** is _WaitForFirstConsumer_
<img width="741" height="754" alt="Image" src="https://github.com/user-attachments/assets/ea76a451-46f6-4d78-b954-8bdd1c342d12" />

2. Tested with **harvester-csi-driver-lvm**[^1], same result as **TopoLVM**


Test Information
----
* Environment: **Baremetal DL360G9 4 nodes / MegaRAC SP-X 3 nodes **
* Harvester Version: **v1.6.0-RC5**
* **ui-source** Option: **Auto**
* Rancher Version: **v2.12.0**

### Verify Steps
1. Install Harvester with at least 2 nodes
2. Import Harvester into Rancher
3. Create RKE2 guest cluster `rke2`
[...]

PR #8469, opened 2025-06-12
Introduce HEP for volume online expansion

WebberHuang1118:
#### Problem:
This feature enables the expansion of volumes while they are attached to a running VM or when the PVC is attached to a running pod in the guest cluster.

#### Solution:
PR introduces the HEP for volume online expansion

#### Related Issue(s):
#2811
#7358


enhancements/20250612-volume-online-expansion.md
```diff
@@ -0,0 +1,297 @@
...
+            - CPUManager
+            - VMPersistentState
+            - ExpandDisks
+        network:
+          ...
+        vmStateStorageClass: vmstate-persistence
+    ```
+    - This enables online resize support on the KubeVirt side. See [KubeVirt PR #5981](https://github.com/kubevirt/kubevirt/pull/5981).
+
+2. **Introduce a New Harvester Setting: `csi-online-expand-validation`**
```
ihcsim, line 59: Can we re-use the `csi-driver-config` to avoid scattering csi configuration in multiple places?

enhancements/20250612-volume-online-expansion.md
```diff
@@ -0,0 +1,297 @@
...
+          - settings
+        verbs:
+          - '*'
+    ```
+    - In the guest cluster, the Harvester CSI driver will verify the existence of the `csi-online-expand-validation` setting to determine if the PVC supports online expansion.
+
+
+#### PVC Webhook Validator
+
+- PVC Webhook Validator for resizing:
```
ihcsim, line 106: Shrinking is also not supported. I don't know if we need to validate this ourselves, or K8s or the CSI drivers will do it for us.

enhancements/20250612-volume-online-expansion.md
```diff
@@ -0,0 +1,297 @@
...
+
+Under these circumstances, the volume will not undergo online expansion. This issue arises because the CSI `NodeExpand()` function is not invoked, as the PVC is not mounted to either the `virt-launcher` pod or the `hotplug` pod.
+
+##### Behavior with Multiple Hotplugged Volumes
+
+When a second volume is hotplugged into the VM, the old `hp-volume` pod is replaced by a new one. The new `hp-volume` pod only handles the `NodeStage()` and `NodePublish()` operations for the second volume. Consequently, the first hotplugged volume undergoes `NodeUnPublish()` and `NodeUnstage()` operations, but its image file remains bind-mounted in the `virt-launcher` pod.
+
+This behavior explains why the online expansion for the first volume is not completed. Since the first volume is no longer managed by the `hp-volume` pod, the necessary CSI operations to propagate the size expansion (e.g., `NodeExpand()`) are not triggered. As a result, the VM does not reflect the volume size expansion for the first hotplugged volume.
+
+#### Harvester CSI Volume Expand Flow
```
ihcsim, line 167: We need to detect expansion failures by watching the PVC's `.status.allocatedResourceStatuses`, and report the errors to the users. User intervention seems required for recovery. Othewise, the K8s PVC controller retries the resizing indefinitely, which is not what we want.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#recovering-from-failure-when-expanding-volumes

enhancements/20250612-volume-online-expansion.md
```diff
@@ -0,0 +1,297 @@
...
+     metadata:
+       name: blk-vol-pod
+       namespace: default
+     spec:
+       containers:
+         - name: volume-test
+           image: ubuntu
+           imagePullPolicy: IfNotPresent
+           securityContext:
+             privileged: true
```
ihcsim: Is `privileged` context required for volume expansion to work?

enhancements/20250612-volume-online-expansion.md
```diff
@@ -0,0 +1,297 @@
...
+     blockdev --getsize64 /dev/blk-vol
+     ```
+
+2. Longhorn v2 PVC
+   - On Harvester, set the default StorageClass to use Longhorn v2.
+   - Apply the same PVC manifest as in Case 1 on the guest cluster.
+   - Deploy the same pod (as in Case 1) to attach the PVC.
+   - Expand the PVC size (e.g., from 10Gi to 20Gi).
+     - The PVC size will not expand.
+     - The PVC will remain in the `Resizing` condition indefinitely.
```
ihcsim, line 277: This will cause continuous churns in the K8s PVC controller and Harvester-side CSI drivers. At the minimum, the user needs to be notified so that they can stop the resizing request. I don't know if there is a safe way to short-circuit this in order to protect the Harvester control plane from retries storm.

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: The csi-driver-config is intended for configuring snapshot-related settings for CSI providers.

So, I’d prefer to have a separate setting to control volume expansion validation. This way, users can explicitly confirm that their CSI provider supports online expansion before enabling it.

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: Yes, Kubernetes will prevent the PVC from shrinking

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: > We need to detect expansion failures by watching the PVC's .status.allocatedResourceStatuses

I didn’t find `.status.allocatedResourceStatuses`. Instead, when the PVC is stuck in the resizing phase, I can see a condition with `.status.conditions[type] = "Resizing"`.

For example, this occurs with a PVC using `harvester-csi-driver v0.2.2`, which doesn’t support online resizing.
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: blk-vol
  namespace: default
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: "9223372036854775807"
  storageClassName: harvester
  volumeMode: Block
  volumeName: pvc-6f2998e3-a6d5-4846-b3a5-d29f8349ab3b
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 10Gi
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2025-06-18T03:32:25Z"
    status: "True"
    type: Resizing
  phase: Bound
```

> the K8s PVC controller retries the resizing indefinitely, which is not what we want

Based on the information above, this issue can also occur with the latest `harvester-csi-driver` (`v0.2.2`), which represents the default Day 1 setup and isn't specifically configured to support online resizing.

From my understanding, the retry behavior is part of the internal logic of the `csi-resizer`. It's not something defined by the CSI driver itself. The CSI driver implements the required gRPC functions, while the actual control flow is handled by `csi-resizer`.

Additionally, in my environment, when a PVC is stuck in the resizing phase, I haven't observed any retry storms—this seems to be independent of the CSI driver's behavior.
```
...
time="2025-06-18T03:53:36Z" level=info msg="GRPC call: /csi.v1.Controller/ControllerExpandVolume request: {\"capacity_range\":{\"required_bytes\":236223201280},\"volume_capability\":{\"AccessType\":{\"Mount\":{\"fs_type\":\"ext4\"}},\"access_mode\":{\"mode\":1}},\"volume_id\":\"pvc-c944091e-497b-4334-93c9-6857b867b8cd\"}"
time="2025-06-18T03:53:36Z" level=error msg="GRPC error: rpc error: code = FailedPrecondition desc = Volume pvc-c944091e-497b-4334-93c9-6857b867b8cd is in use. Online volume expansion is not supported."
```

At the very least, users can manually follow a process similar to Harvester’s [volume expand cancel action](https://github.com/harvester/harvester/blob/be9de3ac020904bea4136bd53ece445a078ccc97/pkg/api/volume/handler.go#L232-L333). This can help get the PVC out of the resizing state while preserving the data content.

IMO, we can include the limitation and its workaround in the documentation update that introduces this feature.

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: This follows the same logic as discussed in [this comment](https://github.com/harvester/harvester/pull/8469#discussion_r2153583166).

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: Updated per suggestion, thanks.

enhancements/20250612-volume-online-expansion.md
```diff
@@ -0,0 +1,295 @@
...
+- PVC Webhook Validator for resizing:
+  - If the volume is not attached to a running VM:
+    - Return `Permitted`
+  - If the KubeVirt `expandDisks` feature gate is not enabled:
+    - Return `Rejected`
+  - If the volume is hotplugged in filesystem mode:
+    - Return `Rejected`
+  - If the `csi-online-expand-validation` setting is not enabled:
+    - Return `Rejected`
+  - If the CSI provider is not Longhorn:
```
ibrokethecloud, line 115: i believe this is meant to be `Longhorn v1`

enhancements/20250612-volume-online-expansion.md
```diff
@@ -0,0 +1,295 @@
...
+    ```
+    - In the guest cluster, the Harvester CSI driver will verify the existence of the `csi-online-expand-validation` setting to determine if the PVC supports online expansion.
+
+
+#### PVC Webhook Validator
+
+- PVC Webhook Validator for resizing:
+  - If the volume is not attached to a running VM:
+    - Return `Permitted`
+  - If the KubeVirt `expandDisks` feature gate is not enabled:
```
ibrokethecloud, line 109: do we really need this check? the kubevirt configuration is being managed by the harvester managedchart and `ExpandDisks` feature gate would have already been enabled as part of the install / upgrade

enhancements/20250612-volume-online-expansion.md
```diff
@@ -0,0 +1,295 @@
...
+    metadata:
+      name: harvesterhci.io:csi-driver
+    rules:
+      ...
+      - apiGroups:
+          - harvesterhci.io
+        resources:
+          - networkfilesystems
+          - networkfilesystems/status
+          - settings
```
ibrokethecloud: we need to change this clusterrole to only allow access to named settings needed to validate volume expansion using `resourceNames` field. Based on the csi PR we only need access to `csi-online-expand-validation`.

With the current settings, any user could use the csi clusterrole to fetch cluster wide settings in the guest cluster which would be vulnerability.

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: Hi @ibrokethecloud,

This condition applies when the CSI provider is **not** Longhorn.
For PVCs backed by Longhorn, it follows the code path below.

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: Yes, you're correct.

However, it serves as a safeguard to ensure double confirmation—maybe not strictly necessary, but it adds an extra layer of safety.

If you don’t mind, I’d prefer to keep the check in place. WDYT?

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: Updated per suggestion, thanks.

enhancements/20250612-volume-online-expansion.md
ihcsim: The `settings.CSIDriverInfo` struct seems generic enough to store any driver configuration, which in a way reminds me of the upstream `storage.k8s.io/v1/csidrivers` API. Are there any reasons why we can't re-purpose it to be a more generic configuration?

enhancements/20250612-volume-online-expansion.md
ihcsim: Hmm.. yeah, I can't trigger `.status.allocatedResourceStatuses` too even though the feature gate is enabled in RKE2 v1.32.3+rke2r1. I think the status is supposed to appear only if the resizing actually started and then failed due to capacity limit.

Playing around with LH a bit (by disabling the LH webhook), a resize failure does cause the resizer to churn indefinitely:
```
E0618 22:06:02.086972       1 controller.go:316] "Error syncing PVC" err="resize volume \"pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624\" by resizer \"driver.longhorn.io\" failed: rpc error: code = Internal desc = Bad response statusCode [500]. Status [500 Internal Server Error]. Body: [code=Internal Server Error, detail=, message=unable to expand volume pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624: error while CheckReplicasSizeExpansion for volume pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624: cannot schedule 531502202880 more bytes to disk 849644d6-4db8-4446-915a-96b77f9c28bf with &{DiskUUID:849644d6-4db8-4446-915a-96b77f9c28bf StorageAvailable:101607014400 StorageMaximum:101862989824 StorageReserved:30558896947 StorageScheduled:5614129152 OverProvisioningPercentage:200 MinimalAvailablePercentage:25}] from [http://longhorn-backend:9500/v1/volumes/pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624?action=expand]"
I0618 22:06:02.087011       1 event.go:389] "Event occurred" object="demo-src/data-vol-lh" fieldPath="" kind="PersistentVolumeClaim" apiVersion="v1" type="Warning" reason="VolumeResizeFailed" message="resize volume \"pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624\" by resizer \"driver.longhorn.io\" failed: rpc error: code = Internal desc = Bad response statusCode [500]. Status [500 Internal Server Error]. Body: [code=Internal Server Error, detail=, message=unable to expand volume pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624: error while CheckReplicasSizeExpansion for volume pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624: cannot schedule 531502202880 more bytes to disk 849644d6-4db8-4446-915a-96b77f9c28bf with &{DiskUUID:849644d6-4db8-4446-915a-96b77f9c28bf StorageAvailable:101607014400 StorageMaximum:101862989824 StorageReserved:30558896947 StorageScheduled:5614129152 OverProvisioningPercentage:200 MinimalAvailablePercentage:25}] from [http://longhorn-backend:9500/v1/volumes/pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624?action=expand]"
```

There are also many events around it:
```
k -n demo-src get events | grep -i expand
46m         Normal    ExternalExpanding        persistentvolumeclaim/data-vol-lh   waiting for an external controller to expand this PVC
42m         Warning   VolumeResizeFailed       persistentvolumeclaim/data-vol-lh   resize volume "pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624" by resizer "driver.longhorn.io" failed: rpc error: code = Internal desc = Bad response statusCode [500]. Status [500 Internal Server Error]. Body: [code=Internal Server Error, detail=, message=unable to expand volume pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624: error while CheckReplicasSizeExpansion for volume pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624: cannot schedule 531502202880 more bytes to disk 849644d6-4db8-4446-915a-96b77f9c28bf with &{DiskUUID:849644d6-4db8-4446-915a-96b77f9c28bf StorageAvailable:101607014400 StorageMaximum:101862989824 StorageReserved:30558896947 StorageScheduled:5614129152 OverProvisioningPercentage:200 MinimalAvailablePercentage:25}] from [http://longhorn-backend:9500/v1/volumes/pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624?action=expand]
45m         Warning   VolumeResizeFailed       persistentvolumeclaim/data-vol-lh   resize volume "pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624" by resizer "driver.longhorn.io" failed: rpc error: code = Internal desc = Bad response statusCode [500]. Status [500 Internal Server Error]. Body: [message=unable to expand volume pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624: error while CheckReplicasSizeExpansion for volume pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624: cannot schedule 531502202880 more bytes to disk 849644d6-4db8-4446-915a-96b77f9c28bf with &{DiskUUID:849644d6-4db8-4446-915a-96b77f9c28bf StorageAvailable:101607014400 StorageMaximum:101862989824 StorageReserved:30558896947 StorageScheduled:5614129152 OverProvisioningPercentage:200 MinimalAvailablePercentage:25}, code=Internal Server Error, detail=] from [http://longhorn-backend:9500/v1/volumes/pvc-3fc94b36-3f8e-47cb-81e2-55660f39f624?action=expand]
```

The cancel action is the right thing to do, as user should be the one deciding on the right capacity to set. At the moment, the UI will show that the volume remains in the `Resizing` state. It will be nice if those events can be propagated to the UI.

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: The `csi-online-expand-validation` setting comes from internal discussions with @innobead and @Vicente-Cheng.

As mentioned in [this section of the enhancement proposal](https://github.com/harvester/harvester/blob/76097110da8f8ec67c85534d79f712ca6974383c/enhancements/20250612-volume-online-expansion.md#why-we-need-the-setting-csi-online-expand-validation), it addresses the need for validating CSI support for online expansion.

> The PersistentVolumeClaimResize admission controller allows size changes when the PVC’s StorageClass has allowVolumeExpansion: true. This is the only built-in gate for volume expansion. Kubernetes does not check CSI driver capabilities when a PVC's size is updated. The request is admitted (or denied) before any CSI sidecars act.

Users should confirm that their third-party storage provider supports online expansion and explicitly enable it. Keeping this as a separate setting helps make the semantics clearer. I’m also trying to avoid overcomplicating the existing `csi-driver-config` structure.

Additionally, the `csi-online-expand-validation` setting serves as a flag for `harvester-csi-driver` to check whether the current Harvester version supports online expansion.

So, if possible, I’d prefer to keep this setting in place. WDYT? @innobead @Vicente-Cheng @ibrokethecloud

enhancements/20250612-volume-online-expansion.md
WebberHuang1118: Yes, that’s a good point. I just checked the latest Harvester GUI, and it doesn’t reflect the resizing status indicated by `.status.conditions[type] = "Resizing"` in the PVC.

I’ll include this as part of the requirements in the related GUI ticket: https://github.com/harvester/harvester/issues/8314

ihcsim (approved):

ibrokethecloud (approved):
lgtm. thanks.

enhancements/20250612-volume-online-expansion.md
houhoucoop: After syncing with @WebberHuang1118 , we confirmed that the code displays `Resizing` correctly for drivers like Longhorn and TopoLVM.

PR #8473, opened 2025-06-13
chore: make stale bot ignore require/investigate label [skip ci]

Yu-Jack:
#7983

since some issues require more time for investigation, so we decided to make the stale bot ignore `require/investigate` label.


w13915984028 (approved):
LGTM, thanks for the quick fix.

innobead (approved):

PR #8481, opened 2025-06-16
allow vm-import-controller managed vm's to directly used golden pvc

ibrokethecloud:
#### Problem:

Harvester validating webhooks block VM's from currently using the golden pvc (source datavolume) directly in the VMs

This works fine in most scenarios except when used with vm-import-controller, which is being enhanced to support 3rd party csi. In this scenario we intend to import the source disk via virtual machine images and cdi combination. These disks are different from traditional vm images as we do not intend to make these the imported image available for subsequently cloning/snapshotting across multiple vm's. In addition these disks can be large, from a few hundred GB's to TB's, as a result it could become inefficient to keep multiple copies of data via the current golden pvc design.

To work around this we need to allow VM's to boot directly off golden pvc in certain scenarios.

#### Solution:

PR introduces a minor change to harvester webhook to skip pvc validation when the VM is being created by the vm-import-controller.

The change depends on an annotation https://github.com/harvester/vm-import-controller/blob/main/pkg/controllers/migration/virtualmachine.go#L43 added by the controller to all imported VM's

#### Related Issue(s):

https://github.com/harvester/harvester/pull/7487
#### Test plan:

No specific changes are needed to validate harvester.

The test plan will be part of the vm-import-controller 3rd party csi support enhancement.

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -315,6 +315,10 @@ func (v *vmValidator) checkVolumeReq(newVM *kubevirtv1.VirtualMachine) error {
 		return nil
 	}
 
+	// if vm is being created by vm import controller then we need to skip pvc golden image validation
+	// this case needs to be treated differently from a standaard image scenario as user will boot directly from disk
+	imported := checkIsVirtualMachineImported(newVM)
+
 	for _, volReq := range newVM.Status.VolumeRequests {
 		if volReq.AddVolumeOptions == nil {
 			continue
@@ -331,6 +335,11 @@ func (v *vmValidator) checkVolumeReq(newVM *kubevirtv1.VirtualMachine) error {
 			}
 			return werror.NewInternalError(fmt.Sprintf("failed to get PVC %s/%s, err: %s", pvcNS, pvcName, err))
 		}
+
+		if imported {
+			continue
+		}
+
 		if _, ok := pvc.Annotations[util.AnnotationGoldenImage]; ok {
 			if pvc.Annotations[util.AnnotationGoldenImage] == "true" {
 				return werror.NewInvalidError(fmt.Sprintf("PVC %s/%s is a golden image, it can't be used as a hotplug volume in VM", pvcNS, pvcName), "status.volumeRequests")
@@ -434,6 +443,10 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
 		return nil
 	}
 
+	// if vm is being created by vm import controller then we need to skip pvc golden image validation
+	// this case needs to be treated differently from a standaard image scenario as user will boot directly from disk
+	imported := checkIsVirtualMachineImported(vm)
+
 	for _, volume := range vm.Spec.Template.Spec.Volumes {
 		if volume.PersistentVolumeClaim == nil {
 			continue
@@ -449,6 +462,11 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
 		if targetPVC.Annotations == nil {
 			continue
 		}
+
+		if imported {
+			continue
+		}
+
 		if _, ok := targetPVC.Annotations[util.AnnotationGoldenImage]; ok {
 			if targetPVC.Annotations[util.AnnotationGoldenImage] == "true" {
 				return werror.NewInvalidError(fmt.Sprintf("PVC %s/%s is a golden image, it can't be used as a volume in VM", targetPVC.Namespace, targetPVC.Name), "spec.templates.spec.volumes")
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -315,6 +315,10 @@ func (v *vmValidator) checkVolumeReq(newVM *kubevirtv1.VirtualMachine) error {
 		return nil
 	}
 
+	// if vm is being created by vm import controller then we need to skip pvc golden image validation
+	// this case needs to be treated differently from a standaard image scenario as user will boot directly from disk
+	_, imported := newVM.Annotations[util.AnnotationVMImportController]
+
```
WebberHuang1118: Have we considered wrapping this up as a utility function?

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -434,6 +443,10 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
 		return nil
 	}
 
+	// if vm is being created by vm import controller then we need to skip pvc golden image validation
+	// this case needs to be treated differently from a standaard image scenario as user will boot directly from disk
+	_, imported := vm.Annotations[util.AnnotationVMImportController]
+
```
WebberHuang1118: Ditto

WebberHuang1118 (commented):
Thanks for the PR, just a minor points.

Aside from the source code implementation, will the golden PVC created by the VM importer be displayed as a VM Image in the Harvester GUI?

If so, it seems the GUI should enforce certain limitations on that item.

ibrokethecloud:
> Thanks for the PR, just a minor points.
> 
> Aside from the source code implementation, will the golden PVC created by the VM importer be displayed as a VM Image in the Harvester GUI?
> 
> If so, it seems the GUI should enforce certain limitations on that item.

yes the image will appear as a regular virtualmachineimage in the UI. I do not think we have any specific annotations / label to distinguish vm-imported managed images from others

WebberHuang1118:
> > Thanks for the PR, just a minor points.
> > Aside from the source code implementation, will the golden PVC created by the VM importer be displayed as a VM Image in the Harvester GUI?
> > If so, it seems the GUI should enforce certain limitations on that item.
> 
> yes the image will appear as a regular virtualmachineimage in the UI. I do not think we have any specific annotations / label to distinguish vm-imported managed images from others

So this is a new case introduced by third-party storage integration:

With Longhorn backing images, the image is read-only, and once the VM importer completes, the resulting VM image can be referenced by any VM.

But with third-party storage, the VM image (which is actually a PVC) created by the VM importer can't be used by other VMs.

Do I understand that correctly? @ibrokethecloud cc @innobead @Vicente-Cheng

ibrokethecloud:
the behaviour is the similar.. with longhorn v1 engine the virtualmachineimage is a backing image. It is still writeable but via a copy-on-write mechanism which is inherited from the backing image. With 3rd party csi the datavolume is using a raw image so making a copy from a golden pvc will double the storage usage for the vm.

in addition there is a change in vm-import-controller which will add an extra label on the vmimage `migration.harvesterhci.io/hideImportedImage=true` which can be used by the UI to filter the vmimage.

WebberHuang1118:
> in addition there is a change in vm-import-controller which will add an extra label on the vmimage `migration.harvesterhci.io/hideImportedImage=true` which can be used by the UI to filter the vmimage.

So just to confirm—we’ll have a related GUI issue to handle the filtering, right?

Vicente-Cheng:
> > in addition there is a change in vm-import-controller which will add an extra label on the vmimage `migration.harvesterhci.io/hideImportedImage=true` which can be used by the UI to filter the vmimage.
> 
> So just to confirm—we’ll have a related GUI issue to handle the filtering, right?

No, I thought we needed to create the one. @ibrokethecloud could you help create the one for the UI?

Vicente-Cheng (approved):
lgtm, thanks.

Could you also link the vm-importer changes for `migration.harvesterhci.io/hideImportedImage=true` with this one?

Then, we will know why we need this annotation.

WebberHuang1118 (approved):

ibrokethecloud:
Related UI issue: https://github.com/harvester/harvester/issues/8504

innobead:
Can we also let users know if VM images are created by the VM importer when using kubectl, possibly by adding an extra column? Additionally, we need to prevent newly created VMs from utilizing those images.

innobead (approved):

ihcsim (approved):

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -331,6 +335,11 @@ func (v *vmValidator) checkVolumeReq(newVM *kubevirtv1.VirtualMachine) error {
 			}
 			return werror.NewInternalError(fmt.Sprintf("failed to get PVC %s/%s, err: %s", pvcNS, pvcName, err))
 		}
+
+		if imported {
```
w13915984028, line 339: A question: 

do we have related import mark on golden pvc as well?

suppose an imported VM later refer to other golden pvc, could it break things? looks like a cross check is required.

also, any other parts to block the clonning of the imported golden pvc?

let's discuss via PR https://github.com/harvester/harvester/pull/8537

Issue #8486, opened 2025-06-16
[ENHANCEMENT] Log version and commit id when Harvester starts


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Harvester has built the version and commit into binary, but it is not shown when container starts up.

### Describe the solution you'd like

Add following to pod log:

```
time="2025-06-16T10:27:46Z" level=info msg="Name: Harvester API Server, Version:7f01ea28 (7f01ea28)"
time="2025-06-16T10:27:46Z" level=info msg="Options: &{Debug:false Trace:false LogFormat: ProfilerAddress: KubeConfig:}"
time="2025-06-16T10:27:46Z" level=info msg="Starting Harvester API Server and controllers"
```

Note: this ticket is used to enhance the pod log, it is verified directly, not require QA testing.

w13915984028:
Enhancement is done.

PR #8487, opened 2025-06-16
Log app info when it starts

w13915984028:
#### Problem:

Harvester POD log does not include the version and commit info, you need to cross check the pod/container info to confirm.

It has been linked to binary
https://github.com/harvester/harvester/blob/0ab8afd0a19946a681fa7228cd8458129220d803/scripts/build#L23

#### Solution:

Log the info when app starts.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8486

#### Test plan:

1. Check harvester pod log when it starts
2. There are info like
```
time="2025-06-17T09:10:52Z" level=info msg="Starting Harvester API Server version f9ccd696 (f9ccd696)"
time="2025-06-17T09:10:53Z" level=info msg="Do not update the existing CRD KeyPair"
```

### DIFF
--- a/pkg/cmd/app.go
+++ b/pkg/cmd/app.go
@@ -71,10 +71,13 @@ func NewApp(name string, usage string, flags []cli.Flag, action Action) *App {
 		return action(&options)
 	}
 
-	return &App{
+	app := &App{
 		app:     cliApp,
 		Options: &options,
 	}
+	// log app basic info
+	app.info()
+	return app
 }
 
 func initProfiling(options *config.CommonOptions) {
@@ -116,4 +119,9 @@ func (a *App) Run() {
 	if err := a.app.Run(os.Args); err != nil {
 		logrus.Fatal(err)
 	}
+}
+
+func (a *App) info() {
+	logrus.Infof("Starting %v version %v", a.app.Name, a.app.Version)
+	logrus.Debugf("Options: %+v", a.Options)
 }
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -5,7 +5,6 @@ import (
 	_ "net/http/pprof"
 
 	"github.com/rancher/wrangler/v3/pkg/signals"
-	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 
 	"github.com/harvester/harvester/pkg/cmd"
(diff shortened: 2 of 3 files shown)


w13915984028:
cc @FrankYang0529 added https://github.com/harvester/harvester/pull/8491 to enhance https://github.com/harvester/harvester/pull/8452

FrankYang0529 (dismissed):

pkg/cmd/app.go
```diff
@@ -117,3 +120,8 @@ func (a *App) Run() {
 		logrus.Fatal(err)
 	}
 }
+
+func (a *App) info() {
+	logrus.Infof("Name: %v, Version: %v", a.app.Name, a.app.Version)
+	logrus.Infof("Options: %+v", a.Options)
```
ihcsim: Do we know if the `kubeconfig` within `a.Options` would contain sensitive info?

ihcsim: Also, I thought I saw some new linter that makes lower-case log messages a requirement or something.

w13915984028: It looks the Harvester deployment starts with the default `KubeConfig` as empty.

```
time="2025-06-17T08:37:57Z" level=info msg="Name: Harvester API Server, Version: ffde3fed (ffde3fed)"
time="2025-06-17T08:37:57Z" level=info msg="Options: &{Debug:false Trace:false LogFormat: ProfilerAddress: KubeConfig:}"
time="2025-06-17T08:37:57Z" level=info msg="Starting Harvester API Server and controllers"
```

w13915984028: Updated with: log the `a.Options` only in `debug` mode

```
time="2025-06-17T09:10:52Z" level=info msg="Starting Harvester API Server version f9ccd696 (f9ccd696)"
time="2025-06-17T09:10:53Z" level=info msg="Do not update the existing CRD KeyPair"
```

w13915984028:
Now Harvester, network-controller, LB log similar info at beginning:
```
time="2025-06-17T09:10:52Z" level=info msg="Starting Harvester API Server version f9ccd696 (f9ccd696)"

I0617 08:44:51.210302       1 main.go:93] Starting harvester-network-controller version 15448977

time="2025-06-17T08:25:40Z" level=info msg="Starting harvester-load-balancer-webhook version 12e0e9e1"

```

ihcsim (approved):

pkg/cmd/app.go
ihcsim: Thanks - digging a bit deeper, looks like `kubeconfig` is the file path, not the actual access credentials.

Issue #7385, opened 2025-01-16
[ENHANCEMENT] SB Enhancements (NS selection and timeout)


bk201:
**Is your enhancement related to a problem? Please describe.**

A support bundle only collects selected namespaces unless the user configures additional namespaces in the settings. It's not intuitive in the UX. Also the customer wants to be able to customize the timeout value.

**Describe the solution you'd like**

Enhancement support bundle collection with:
- Add the harvester cluster name and version to the support bundle name
- Make it possible to select the namespaces when a support bundle is created. Even add "all namespaces" button. This will reduce the the load on both sides, because it will reduce the repeated creation of "wrong" support bundles.
- The dynamic choice should not change the default namespaces configured in settings.
- Increase the default support bundle timeout to something bigger, e.g. 30 min

**Describe alternatives you've considered**

**Additional context**

SURE-9519

Yu-Jack:
Current flow is that GUI creates a `supportbundle` resource with description and URL in spec field. 

Note environments used in SB for future usage:

```yaml
# request from frontend
{
    "apiVersion": "harvesterhci.io/v1beta1",
    "type": "harvesterhci.io.supportbundle",
    "kind": "SupportBundle",
    "metadata": {
        "name": "bundle-1kzct",
        "namespace": "harvester-system"
    },
    "spec": {
        "issueURL": "",
        "description": "test"
    }
}
```

```yaml
# deployment 
image: rancher/support-bundle-kit:v0.0.46
env: 
- name: SUPPORT_BUNDLE_TARGET_NAMESPACES
  value: cattle-dashboards,cattle-fleet-local-system,cattle-fleet-system,cattle-fleet-clusters-system,cattle-monitoring-system,fleet-local,harvester-system,local,longhorn-system,cattle-logging-system,cattle-provisionin    g-capi-system
- name: SUPPORT_BUNDLE_NAME
  value: bundle-0hl85
- name: SUPPORT_BUNDLE_DESCRIPTION
  value: {input from UI}
- name: SUPPORT_BUNDLE_ISSUE_URL
  value: {input from UI}
- name: SUPPORT_BUNDLE_DEBUG
  value: "true"
- name: SUPPORT_BUNDLE_MANAGER_POD_IP
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: status.podIP
- name: SUPPORT_BUNDLE_IMAGE
  value: rancher/support-bundle-kit:v0.0.46
[...]

PR #8497, opened 2025-06-17
feat: support inputing timeout and expiration value from frontend to generate support bundle

Yu-Jack:
#### Problem:
Currently, users can't define which namespaces they'd like to collect. Regarding timeout and expiration, they need to configure it in the setting when they need. It's not convenient.

#### Solution:

- Add new fields in support bundle to support different value per request.
- Provide `/v1/harvester/namespaces?link=supportbundle` to serve support bundle used namespaces.

#### Related Issue(s):
#7385 

#### Test plan:

Please refer the [automation test cases](https://github.com/harvester/tests/pull/2068).


pkg/controller/master/supportbundle/controller.go
```diff
@@ -114,15 +108,12 @@ func (h *Handler) checkExistTime(sb *harvesterv1.SupportBundle) (*harvesterv1.Su
...
-		if timeout, err = strconv.Atoi(timeoutStr); err != nil {
-			return nil, err
-		}
+	timeout, err := determineTimeout(sb)
+	if err != nil {
+		return nil, err
 	}
 
-	if timeout != 0 && time.Now().After(sb.CreationTimestamp.Add(time.Duration(timeout)*time.Minute)) {
+	if timeout != 0 && time.Now().After(sb.CreationTimestamp.Add(timeout)) {
```
innobead, line 116: Suggest adding the timeout in the error message.

pkg/controller/master/supportbundle/manager.go
```diff
@@ -262,3 +282,41 @@ func GetManagerPodIP(podCache ctlcorev1.PodCache, sb *harvesterv1.SupportBundle)
...
+		sb.Spec.Expiration,
+		settings.SupportBundleExpiration.GetInt(),
+		supportBundleUtil.SupportBundleExpirationDefault,
+	)
+}
+
+// determineTimeout determines the timeout based on priority:
+func determineTimeout(sb *harvesterv1.SupportBundle) (time.Duration, error) {
+	// Priority 1: If spec.Timeout is set, use it directly
+	if sb.Spec.Timeout != nil && *sb.Spec.Timeout >= 0 {
```
innobead, line 291: Why don't we use reuse supportBundleUtil.DetermineDurationWithDefaults here?

pkg/webhook/resources/supportbundle/validator.go
```diff
@@ -0,0 +1,64 @@
...
+	return types.Resource{
+		Names:      []string{v1beta1.SupportBundleResourceName},
+		Scope:      admissionregv1.NamespacedScope,
+		APIGroup:   v1beta1.SchemeGroupVersion.Group,
+		APIVersion: v1beta1.SchemeGroupVersion.Version,
+		ObjectType: &v1beta1.SupportBundle{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+			admissionregv1.Update,
+		},
```
innobead, line 38: Do we need `update` here? I assume support bundle should be immutable after creation.

innobead (changes requested):
In general, LGTM. Just a few feedback.

pkg/controller/master/supportbundle/manager.go
Yu-Jack: > support-bundle-timeout
> When the value is 0, the timeout feature is disabled.
>
> support-bundle-expiration, support-bundle-node-collection-timeout
> When the value is 0, Harvester uses the default value.

It has slightly different logic. The `determineTimeout` accepts zero, but others doesn't. I want to make it clearly, so I separated the function.

pkg/webhook/resources/supportbundle/validator.go
Yu-Jack: I found support bundle mutable at the beginning. Let me add blocking update. It should be immutable.

innobead (dismissed):

deploy/charts/harvester-crd/templates/harvesterhci.io_supportbundles.yaml
```diff
@@ -52,8 +52,30 @@ spec:
             properties:
               description:
                 type: string
+              expiration:
+                description: Number of minutes Harvester waits before deleting a support
+                  bundle that has been packaged but not downloaded (either deliberately
+                  or unsuccessfully) or retained.
+                minimum: 0
```
Yu-Jack, line 59: I change the minimum to match documentation.

deploy/charts/harvester-crd/templates/harvesterhci.io_supportbundles.yaml
```diff
@@ -52,8 +52,30 @@ spec:
...
+                items:
+                  type: string
+                type: array
               issueURL:
                 type: string
+              nodeTimeout:
+                description: Number of minutes Harvester allows for collection of
+                  logs and configurations (Harvester) on the nodes for the support
+                  bundle.
+                minimum: 0
```
Yu-Jack, line 71: I change the minimum to match documentation.

Yu-Jack (commented):
I added following changes after first approval.

pkg/api/namespace/store.go
```diff
@@ -0,0 +1,52 @@
+package namespace
```
Yu-Jack, line 1: this is new file to provide support bundle used namespaces.

FrankYang0529 (approved):
LGTM. Thanks for the patch.

PR #8509, opened 2025-06-18
feat: patch guest value if there is no enableCPUAndMemoryHotplug annotation

FrankYang0529:
* Patch guest value if there is no enableCPUAndMemoryHotplug annotation.
* Validate guest + reserved memory <= limits memory

#### Related Issue(s):

#### Test plan:

* Case 1: VM without `harvesterhci.io/enableCPUAndMemoryHotplug` can work as usual. Using `resources.limits` as memory input.
* Case 2: VM with `harvesterhci.io/enableCPUAndMemoryHotplug` use domain guest memory as memory input.
  * Create a VM without reserved memory, the guest input should not be updated.
  * Create a VM with reserved memory + guest memory > limits memory, the request should be denied.
  * Create a VM with 1 socket, 1Gi guest and 4Gi limits memory. Use `cpuAndMemoryHotplug` action to patch sockets to 2 and guest to 2Gi. The action should be successful.

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -18,6 +18,7 @@ import (
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
 	"github.com/harvester/harvester/pkg/indexeres"
 	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/util/virtualmachine"
 	"github.com/harvester/harvester/pkg/util/virtualmachineinstance"
 )
 
@@ -411,6 +412,9 @@ func canCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
 		return false
 	}
 	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
+		return false
+	}
+	if !virtualmachine.SupportCPUAndMemoryHotplug(vm) {
 		return false
 	}
 
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -17,6 +17,7 @@ import (
 	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/util/virtualmachine"
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
@@ -248,6 +249,7 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity, 
 	var err error
 	newRequest := mem.Value() * int64(100) / int64(overcommit.Memory) / 1048576 * 1048576
 	quantity := resource.NewQuantity(newRequest, mem.Format)
+	enableCPUAndMemoryHotplug := virtualmachine.SupportCPUAndMemoryHotplug(vm)
 
 	// Reserve 100MiB (104857600 Bytes) for QEMU on guest memory
 	// Ref: https://github.com/harvester/harvester/issues/1234
@@ -275,7 +277,14 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity, 
 	// when AnnotationReservedMemory is set, or AdditionalGuestMemoryOverheadRatioConfig is not set
 	// if AdditionalGuestMemoryOverheadRatioConfig and AdditionalGuestMemoryOverheadRatioConfig are both set
 	// then the VM will benefit from both
+
+	// if hotpluggable guest memory is set by the user, use it for subsequent calculations.
+	// otherwise, set the initial guest memory to be the same as the vm's memory limit.
 	guestMemory := *mem
+	if enableCPUAndMemoryHotplug && vm.Spec.Template.Spec.Domain.Memory != nil && vm.Spec.Template.Spec.Domain.Memory.Guest != nil {
+		guestMemory = *vm.Spec.Template.Spec.Domain.Memory.Guest
+	}
+
 	if useReservedMemory {
 		guestMemory.Sub(reservedMemory)
 	}
@@ -310,7 +319,7 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity, 
 
 	if vm.Spec.Template.Spec.Domain.Memory == nil {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/domain/memory", "value": {"guest":"%s"}}`, &guestMemory))
-	} else {
+	} else if !enableCPUAndMemoryHotplug {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/domain/memory/guest", "value": "%s"}`, &guestMemory))
 	}
 
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -540,6 +540,14 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
 	if reservedMemory.Cmp(*mem) >= 0 {
 		return werror.NewInvalidError("reservedMemory cannot be equal or greater than limits.memory", field)
 	}
+	if vm.Spec.Template.Spec.Domain.Memory != nil && vm.Spec.Template.Spec.Domain.Memory.Guest != nil {
+		guestMemory := vm.Spec.Template.Spec.Domain.Memory.Guest
+		totalMemory := *guestMemory
+		totalMemory.Add(reservedMemory)
+		if totalMemory.Cmp(*mem) > 0 {
+			return werror.NewInvalidError(fmt.Sprintf("guest memory %v plus reservedMemory %v cannot be greater than limits.memory %v", guestMemory.Value(), reservedMemory.Value(), mem.Value()), field)
+		}
+	}
 	if reservedMemory.CmpInt64(0) == -1 {
 		return werror.NewInvalidError("reservedMemory cannot be less than 0", field)
 	}
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -34,6 +34,7 @@ const (
 	AnnotationIsDefaultStorageClassName = "storageclass.kubernetes.io/is-default-class"
 	AnnotationLastRefreshTime           = prefix + "/lastRefreshTime"
 	AnnotationMacAddressName            = prefix + "/mac-address"
+	AnnotationEnableCPUAndMemoryHotplug = prefix + "/enableCPUAndMemoryHotplug"
 
 	AnnotationSkipRancherLoggingAddonWebhookCheck = prefix + "/skipRancherLoggingAddonWebhookCheck"
 
--- a/pkg/util/virtualmachine/virtualmachine.go
+++ b/pkg/util/virtualmachine/virtualmachine.go
@@ -2,12 +2,14 @@ package virtualmachine
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/util"
 )
 
 // IsVMStopped checks VM is stopped or not. It will check two cases
@@ -49,4 +51,12 @@ func IsVMStopped(
 	}
 
 	return false, nil
+}
+
+func SupportCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
+	if vm == nil || vm.Annotations == nil {
+		return false
+	}
+
+	return strings.ToLower(vm.Annotations[util.AnnotationEnableCPUAndMemoryHotplug]) == "true"
 }


pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -310,7 +310,9 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity,
 
 	if vm.Spec.Template.Spec.Domain.Memory == nil {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/domain/memory", "value": {"guest":"%s"}}`, &guestMemory))
-	} else {
+	} else if useReservedMemory {
```
ihcsim: Technically, we only need to reduce the guest memory if `guest_memory + reserved_memory > limits_memory`. E.g., if the pod's memory limit is 4GiB, and the guest asks for 3GiB and reserved memory is set to 256MiB. There is no need to reduce the guest value. With this PR, the validator would have rejected such request. I wonder if we actually still need this conditional.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -540,6 +540,13 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
 	if reservedMemory.Cmp(*mem) >= 0 {
 		return werror.NewInvalidError("reservedMemory cannot be equal or greater than limits.memory", field)
 	}
+	if vm.Spec.Template.Spec.Domain.Memory != nil && vm.Spec.Template.Spec.Domain.Memory.Guest != nil {
+		totalMemory := *vm.Spec.Template.Spec.Domain.Memory.Guest
+		totalMemory.Add(reservedMemory)
+		if totalMemory.Cmp(*mem) > 0 {
+			return werror.NewInvalidError("guest memory + reservedMemory cannot be greater than limits.memory", field)
```
ihcsim: can we include the actual values of the guest, reserved and limit in the error message?

ihcsim (commented):
If provided, the mutator needs to use the user-provided guest memory value to ensure that it matches the pod's memory request as described in the `else if hostDevicesPresent(vm)` case:

```diff
diff --git a/pkg/webhook/resources/virtualmachine/mutator.go b/pkg/webhook/resources/virtualmachine/mutator.go
index 72e160851..1cd31ae2a 100644
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -275,10 +275,18 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity,
 	// when AnnotationReservedMemory is set, or AdditionalGuestMemoryOverheadRatioConfig is not set
 	// if AdditionalGuestMemoryOverheadRatioConfig and AdditionalGuestMemoryOverheadRatioConfig are both set
 	// then the VM will benefit from both
+
+	// if hotpluggable guest memory is set by the user, use it for subsequent calculations.
+	// otherwise, set the initial guest memory to be the same as the pod's memory limit.
 	guestMemory := *mem
+	if vm.Spec.Template.Spec.Domain.Memory != nil && vm.Spec.Template.Spec.Domain.Memory.Guest != nil {
+		guestMemory = &vm.Spec.Template.Spec.Domain.Memory.Guest
+	}
+
 	if useReservedMemory {
 		guestMemory.Sub(reservedMemory)
 	}
+
 	if guestMemory.Value() < memory256M {
 		// some test cases use a small value to test, but for production such VM makes no sense, add a warning here
 		logrus.WithFields(logrus.Fields{

```

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -275,7 +275,14 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity,
 	// when AnnotationReservedMemory is set, or AdditionalGuestMemoryOverheadRatioConfig is not set
 	// if AdditionalGuestMemoryOverheadRatioConfig and AdditionalGuestMemoryOverheadRatioConfig are both set
 	// then the VM will benefit from both
+
+	// if hotpluggable guest memory is set by the user, use it for subsequent calculations.
+	// otherwise, set the initial guest memory to be the same as the pod's memory limit.
```
w13915984028: pod's memory limit.
->
vm's memory limit

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -310,8 +317,6 @@ func generateMemoryPatch(vm *kubevirtv1.VirtualMachine, mem *resource.Quantity,
 
 	if vm.Spec.Template.Spec.Domain.Memory == nil {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/domain/memory", "value": {"guest":"%s"}}`, &guestMemory))
-	} else {
-		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/template/spec/domain/memory/guest", "value": "%s"}`, &guestMemory))
```
w13915984028, line 314: this change might break existing VMs.

1. for `legacy` mode VM, the guest is decided by `domain.resources`, UI sets `domain.resources`, mutator patchs guest per reserved label, always replace the guest memory with a computed value;

2. for `hot-plug` mode vm, the guest is the source of truth

without an annotation/label on VM, we can't identify the vm's mode, both have the `guest` value

with this PR, if we change legacy VM, expand the memory from `domian.resources`, the `guest` stucks on old value, esp. the `rancher manager` -> docker machine interface.

https://github.com/harvester/docker-machine-driver-harvester/blob/1c0d7b2175c8a2a237f921c93805bf766724967d/harvester/create.go#L43

https://github.com/harvester/harvester/blob/96302ffb12623c08a309a399162df68ca2c852aa/pkg/builder/vm.go#L53

we have docker-machine-driver, terraform provider;  user also have their code to configure VM's resources widely...

harvester supports a new `hot-plug` mode VM; and users also need to adapt step by step

e.g. The adpation of `hot-plug` in Rancher Manager:
(1) A new docker-machine-driver
(2) Adds new UI elements on Rancher, adds backend processing ...
(3) When & how to use `hot-plug`? Rancher-manager needs some stories for user to set it from initial and use it on the run-time

w13915984028 (commented):
those additional PRs enhanced my thought: differentiating the mode of vm is essential to avoid breaking legacy VMs

There are also a category of `non-migratable VMs` (vGPU, passthrough...), for them, the current:

additional-guest-overhead  + reseved memory is a flexiable combination to avoid OOM issues. Stick to current mode avoids any potential risks.

When user toggles a VM as `hot-plug` mode but selects a vGPU etc., UI/webhook can quickly deny the request.

pkg/webhook/resources/virtualmachine/mutator.go
ihcsim: Can this done in 1.7? Otherwise, the change and QA scope will be too big. For 1.6, hotplug updates can only be done from Harvester UI. For Rancher integration, the mutator retains the logic to fill in the guest resources if they aren't provided in the request.

ihcsim:
> additional-guest-overhead + reseved memory is a flexiable combination to avoid OOM issues. Stick to current mode avoids any potential risks.

The intention is that if reserved memory is used, the mutator does what it's doing now.

pkg/webhook/resources/virtualmachine/mutator.go
w13915984028: What I am thinking is, the v1.6.0 hot-plug mode vm needs to have a label;  for any VM without this label, fallback to legacy mode.

With an explicit label on `hot-plug` mode VM and the webhook/controller process VM per this label, all others are not affected, and rancher/terraform... can adapt to this new mode step by step.

This way, the sope and QA validation can be limited.

w13915984028:
> > additional-guest-overhead + reseved memory is a flexiable combination to avoid OOM issues. Stick to current mode avoids any potential risks.
> 
> The intention is that if reserved memory is used, the mutator does what it's doing now.

With `additional-guest-overhead`, the `reserved memory` https://docs.harvesterhci.io/v1.5/vm/index#reserved-memory is only set when the general overhead is still less than the real used; assume `reserved memory` can't cover all cases.  And Rancher->Harvester, the `reserved memory` is even not exposed on UI, an ongoing requirment for it is on list.

pkg/webhook/resources/virtualmachine/mutator.go
ihcsim: Rancher integration will affect both legacy and hotplug VMs. If a Rancher integration user enabled hotplug on an existing legacy VM, then the guest auto-configuration would stop working. If possible, we should find a better way to make this a more consistent UX. Up till now, Rancher integration users don't have to worry about domain resource topology configuration. Either it stays that way or we add new UI elements to Rancher. The latter increases the scope but we can have a separate discussion on timeline and milestone.

w13915984028:
@FrankYang0529 @ihcsim

I did following test:

(1) Before replace the webhook image with this PR, there is one existing VM:

User claims 4Gi memory, the final VM spec is:
```
          memory:
            guest: 4Gi    // `reserved memory` is 0 by default
          resources:
            limits:
              cpu: "2"
              memory: 4Gi  // UI setting is mapped to this field
            requests:
              cpu: 125m  
              memory: 2730Mi  // overhead re-balance
```

From UI, change the memory to 6Gi
image

The final VM spec is:

```
        memory:
          guest: 4Gi  // stucks on old value;  not the expected 6Gi
        resources:
          limits:
            cpu: '2'
            memory: 6Gi  // adjust to 6Gi from UI,  note: third-party code runs upon this field.
          requests:
            cpu: 125m  
            memory: 4Gi
```

This PR is OK for newly created VM;  but for existing VMs, it might not work, as all exsting VMs (before upgrade to v1.6.0) have the `memory.guest` field set by the mutator in the past;  the PR runs below check and set, then is affected by existing values when user changes the VM memory.

```
        // when AnnotationReservedMemory is set, or AdditionalGuestMemoryOverheadRatioConfig is not set
        // if AdditionalGuestMemoryOverheadRatioConfig and AdditionalGuestMemoryOverheadRatioConfig are both set
        // then the VM will benefit from both
+
+       // if hotpluggable guest memory is set by the user, use it for subsequent calculations.
+       // otherwise, set the initial guest memory to be the same as the pod's memory limit.
        guestMemory := *mem
+       logrus.Infof("Guest memory initial: %v", guestMemory)
+       if vm.Spec.Template.Spec.Domain.Memory != nil && vm.Spec.Template.Spec.Domain.Memory.Guest != nil {
+               guestMemory = *vm.Spec.Template.Spec.Domain.Memory.Guest
+               logrus.Infof("Guest memory updated with guest: %v", guestMemory)
+       }

```

debug log:
```
time="2025-06-23T08:04:36Z" level=info msg="Guest memory initial: {{6442450944 0} {<nil>} 6Gi BinarySI}"
time="2025-06-23T08:04:36Z" level=info msg="Guest memory updated with guest: {{4294967296 0} {<nil>} 4Gi BinarySI}"
```

This means: the PR will break any existing VMs, if user change the VM's memory spec.

FrankYang0529:
> This PR is OK for newly created VM; but for existing VMs, it might not work, as all exsting VMs (before upgrade to v1.6.0) have the `memory.guest` field set by the mutator in the past; the PR runs below check and set, then is affected by existing values when user changes the VM memory.

@w13915984028 This is why we also need UI change. As @ihcsim mentioned in https://github.com/harvester/harvester/issues/8416#issuecomment-2986517561, there will be 4 fields on VM creation/edition page. The required fields `cpu/memory` are mapping to `cores*sockets*threads/guest`. The optional fields are `resources.limits.cpu` and `resources.limits.memory`. For the example VM you mentioned, the UI page shows 4Gi memory and 4Gi `resources.limits.memory`. If users try to update `resources.limits.memory` to 6Gi, they will get error from KubeVirt webhook and they can see the error message from the UI.

w13915984028:
@FrankYang0529 Harvester's resources are not only managed by UI, theare are Rancher driver, terraform, third-party api, kubectl..., for none-UI operations, the above assumption is not always true, very careful processing is needed to avoid breaking existing things.

```
func (m *vmMutator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) (types.PatchOps, error) {
	newVM := newObj.(*kubevirtv1.VirtualMachine)
	oldVM := oldObj.(*kubevirtv1.VirtualMachine)

...

	if needUpdateResourceOvercommit(oldVM, newVM) {
		patchOps, err = m.patchResourceOvercommit(newVM)
	}
```

The `patchResourceOvercommit` can further include the `oldVM` to make decision:

e.g. when oldVM's guest and newVM's guest are same  && resources.limits are not same, then they are updated from legacy mode; if not, from hot-plug/new mode ...

FrankYang0529:
@w13915984028 I'm not sure whether we promise compatibility among tools you said. For Rancher driver and  terraform, we have support matrix in document. We can update it in v1.6. For third-party API, actually, I don't known which one relies on our API.

> e.g. when oldVM's guest and newVM's guest are same && resources.limits are not same, then they are updated from legacy mode; if not, from hot-plug/new mode ...

This statement is not always true. In new UI, users also can update `resources.limits.memory` only, right?

w13915984028:
@FrankYang0529 

In legacy-mode, VM's memory spec is controlled via `resources.limits`, `resources.requests` and `memory.guest` are computed accordingly.

In hot-plug mode, `memory.guest`, `memory.maxguest` will be the source spec; and others fields like `resources.limits` should be computed by Harvester (mutator) or Kubevirt accordingly, let user edit the `resources.limits` will have risks like the limits are not matching guest.

We can either label all VMs from the upgrade path with `legacy-mode`;  or add `hot-plug-mode` to new VMs defined on v1.6.0. Then we know which field is the source of truth. WDYT?

`hot-plug-mode` will give us more time to expand it on docker-driver, terraform ...

```
          memory:
            guest: 4Gi     // computed
          resources:
            limits:
              cpu: "2"
              memory: 4Gi  // legacy-mode: UI setting is mapped to this field
            requests:
              cpu: 125m  
              memory: 2730Mi // computed
```

FrankYang0529:
> In hot-plug mode, `memory.guest`, `memory.maxguest` will be the source spec; and others fields like `resources.limits` should be computed by Harvester (mutator) or Kubevirt accordingly, let user edit the `resources.limits` will have risks like the limits are not matching guest.

The `resources.limits` cannot be computed by Harvester (mutator), because changing it in mutator will cause `RestartRequriedCondition` condition and the VM cannot run hot-plug.

w13915984028:
It is even better that the `resources.limits` can't be modified on `hot-pulg` mode VM.

Suppose there is a new `hot-plug-mode` label on VM's defined by Harvester v1.6.0 UI
(1) Only guest related fieds are modified via UI, if user changes other field like `resources.limits` via edit yaml/kubectl, validator denies it;
(2) For legacy mode, UI auto switches to legacy mode; if user updates hot-plug related field, also refuses it;
(3) Existing docker-machie-driver, terraform fallback to legacy mode;  extend them to support both mode on new version;
(4) UI can even have an option for user to convert `legacy-mode` VM to `hot-plug-mode` when VM is off.

FrankYang0529:
> It is even better that the `resources.limits` can't be modified on `hot-pulg` mode VM.

Then we need to accept there may be some VMs don't have `resources.limits`, or we need to let them create a large `resources.limits` in VM creation like this PR suggests.

w13915984028:
> > It is even better that the `resources.limits` can't be modified on `hot-pulg` mode VM.
> 
> Then we need to accept there may be some VMs don't have `resources.limits`, or we need to let them create a large `resources.limits` in VM creation like this PR suggests.

When we add detailed description on the UI and document, user can make decison upon their requirements:
(1) For flexibiblity to hot-plug more resources on VMs, then select `hot-plug` mode;
(2) For fine-grained control on VM resources allocation, then select `legacy/traditional` mode;

(3) As mentioned above, some VMs like with vGPU can't do migration, `hot-plug` makes little sense to them. When webhook can detect them and report error, it is more user-friendly.

I guess users also need time to be familiar with the new `hot-plug` mode, understand the benefits and drawbacks, then make decision and plan their adaption.

ihcsim:
> let user edit the resources.limits will have risks like the limits are not matching guest.

The validator can help enforce this. Currently, the mutator sets the guest memory to either the same as the memory limit or the memory limit less the reserved memory. The validator can be updated to enforce these rules:

* `guest + reserved_memory <= domain.resources.limits.memory` (if reserved memory is configured), or
* `guest <= domain.resources.limits.memory` (if reserved memory is not configured) or
* for workloads requiring guaranteed QoS, `guest == domain.resources.limits.memory`

> (3) As mentioned above, some VMs like with vGPU can't do migration, hot-plug makes little sense to them. When webhook can detect them and report error, it is more user-friendly.

:+1: I think we have some util functions to determine if a VM is migratable.

FrankYang0529:
@w13915984028 I updated the PR as we discussed in the Slack. Could you check it again? Thanks.

pkg/api/vm/formatter.go
```diff
@@ -413,6 +415,11 @@ func canCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
 	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
 		return false
 	}
+	if vm.Annotations != nil {
```
w13915984028: Could this be encapsulated as a public API?

pkg/webhook/resources/virtualmachine/mutator_test.go
```diff
@@ -136,23 +136,6 @@ func TestPatchResourceOvercommit(t *testing.T) {
 			},
 			setting: `{"cpu":1000,"memory":1000,"storage":800}`,
 		},
-		{
-			name: "replace old guest memory",
```
w13915984028: Should this test case be reverted and some new test cases upon hotplug mode be added?

w13915984028 (dismissed):
LGTM, thanks.

w13915984028 (approved):

PR #8526, opened 2025-06-20
Extend restoreVM to support multi-node environments

brandboat:
#### Problem:

In the current upgrade flow, if there is any non-migratable & running VMs exist, Harvester will ask the user to shut them down. After the upgrade completes, users need to manually start those VMs again. If users want to minimize the downtime of non-migratable VMs, they have to monitor the upgrade process for each node and manually restart the VMs as soon as their node finishes upgrading, which is quite inconvenient.

#### Solution:

This PR aims to provide a way to automatically bring non-migratable VMs back online once their node has been upgraded. We already have an option for this (`restoreVM`), as documented here: https://docs.harvesterhci.io/v1.6/advanced/index/#upgrade-config. However, currently this feature only works in a single-node environment. This PR extends the feature to support multi-node environments.

The idea is quite simple:
If restoreVM set to `true`
1. Record all non-migratable VMs to configmap in pre-drain job (multi-node env) or single-node-upgrade job (single node env).
2. Non-migratable VMs will be shut down automatically when their node is about to be upgraded. And for migratable VMs, they should remain in running state during the whole upgrade process (if there is no trouble).
3. Once the node upgrade finished (i.e. os image upgraded), `upgrade/job_controller` submit the restore-vm job to restart VMs recorded in the configmap created in step1. Repeat the same process after each subsequent node upgrade.

[Note]: 
- paused non-migratable VM **_won't be started again_** after node upgrade finished.
- if upgrade failed, non-migratable VMs won't be restarted. 

#### Related Issue(s):

#8049 

#### Test plan:

**Test single node upgrade**
There are 2 scenarios here:
1. v1.5.0 upgrade to the version that includes this patch.
    - Prepare 3 VMs vm-stopped (in stopped state), vm-paused (in paused state), vm-running (in running state).
    - Enable `restoreVM` option
    - After the upgrade finished, only vm-running should be in running state, others should be in stopped state.
3. master -> master
    - Prepare 2 VMs, vm-stopped, vm-paused, vm-running.
    - Enable `restoreVM` option
    - After the upgrade finished, only vm-running should be in running state, others should be in stopped state.

**Test multi-node upgrade**
- **WARNING: this can only be done by upgrading from master to master** as restore-vm option is only supported in single node before this PR merged. 
- Prepare a multi-node harvester cluster, create 5VMs
  vm-running, vm-stopped, vm0 (node-selector: node0), vm1 (node-selector: node1), vm2 (node-selector: node2)
- Enable the `restoreVM` option in upgradeConfig setting
- After upgrade finished, the VM state should be like picture in below.
image

What are non-migratable VMs?
- VM contains node selector
- VM contains PCIe devices
- VM contains node affinities while we can't migrate to those nodes in node-affinities settings 
- VM contains CD-ROM or container disks
https://github.com/harvester/harvester/blob/master/pkg/util/virtualmachineinstance/virtualmachineinstance.go#L17

### DIFF
--- a/cmd/upgradehelper/cmd/vmlivemigratedetector/vmlivemigratedetector.go
+++ b/cmd/upgradehelper/cmd/vmlivemigratedetector/vmlivemigratedetector.go
@@ -14,6 +14,7 @@ import (
 
 var (
 	shutdown bool
+	upgrade  string
 )
 
 var vmLiveMigrateDetectorCmd = &cobra.Command{
@@ -31,6 +32,7 @@ If there is no place to go, it can optionally shut down the VMs.
 			KubeConfigPath: cmd.KubeConfigPath,
 			KubeContext:    cmd.KubeContext,
 			Shutdown:       shutdown,
+			Upgrade:        upgrade,
 			NodeName:       args[0],
 		}
 		if err := run(ctx, options); err != nil {
@@ -42,6 +44,9 @@ If there is no place to go, it can optionally shut down the VMs.
 
 func init() {
 	vmLiveMigrateDetectorCmd.Flags().BoolVar(&shutdown, "shutdown", false, "Shutdown non-migratable VMs")
+	vmLiveMigrateDetectorCmd.Flags().StringVar(&upgrade, "upgrade", "",
+		"Upgrade CR name. If non-empty, it's used to generate the name of the configmap that stores all "+
+			"non-migratable VMs stopped by Harvester during the upgrade")
 
 	cmd.RootCmd.AddCommand(vmLiveMigrateDetectorCmd)
 }
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -289,7 +289,7 @@ command_pre_drain() {
   wait_longhorn_engines
 
   # Shut down non-live migratable VMs
-  upgrade-helper vm-live-migrate-detector "$HARVESTER_UPGRADE_NODE_NAME" --shutdown
+  upgrade-helper vm-live-migrate-detector "$HARVESTER_UPGRADE_NODE_NAME" --shutdown --upgrade "$HARVESTER_UPGRADE_NAME"
 
   # Live migrate VMs
   kubectl taint node $HARVESTER_UPGRADE_NODE_NAME --overwrite kubevirt.io/drain=draining:NoSchedule
@@ -723,8 +723,8 @@ command_single_node_upgrade() {
   NEW_OS_SQUASHFS_IMAGE_FILE=$(mktemp -p $UPGRADE_TMP_DIR)
   download_file "$UPGRADE_REPO_SQUASHFS_IMAGE" "$NEW_OS_SQUASHFS_IMAGE_FILE"
 
-  # Stop all VMs
-  shutdown_all_vms
+  # Shut down non-live migratable VMs
+  upgrade-helper vm-live-migrate-detector "$HARVESTER_UPGRADE_NODE_NAME" --shutdown --upgrade "$HARVESTER_UPGRADE_NAME"
   wait_vms_out
 
   echo "wait for fleet bundles before upgrading RKE2"
--- a/pkg/controller/master/upgrade/job_controller.go
+++ b/pkg/controller/master/upgrade/job_controller.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 
+	jobv1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/batch/v1"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	batchv1 "k8s.io/api/batch/v1"
@@ -18,6 +20,7 @@ import (
 	ctlclusterv1 "github.com/harvester/harvester/pkg/generated/controllers/cluster.x-k8s.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	upgradev1 "github.com/harvester/harvester/pkg/generated/controllers/upgrade.cattle.io/v1"
+	"github.com/harvester/harvester/pkg/util"
 )
 
 const (
@@ -45,6 +48,7 @@ const (
 	upgradeJobTypeLabel             = "harvesterhci.io/upgradeJobType"
 	upgradeJobTypePreDrain          = "pre-drain"
 	upgradeJobTypePostDrain         = "post-drain"
+	upgradeJobTypeRestoreVM         = "restore-vm"
 	upgradeJobTypeSingleNodeUpgrade = "single-node-upgrade"
 )
 
@@ -55,10 +59,13 @@ type jobHandler struct {
 	upgradeClient ctlharvesterv1.UpgradeClient
 	upgradeCache  ctlharvesterv1.UpgradeCache
 
-	machineCache ctlclusterv1.MachineCache
-	secretClient ctlcorev1.SecretClient
-	nodeClient   ctlcorev1.NodeClient
-	nodeCache    ctlcorev1.NodeCache
+	machineCache   ctlclusterv1.MachineCache
+	secretClient   ctlcorev1.SecretClient
+	nodeClient     ctlcorev1.NodeClient
+	nodeCache      ctlcorev1.NodeCache
+	jobClient      jobv1.JobClient
+	jobCache       jobv1.JobCache
+	configMapCache ctlcorev1.ConfigMapCache
 }
 
 func (h *jobHandler) OnChanged(_ string, job *batchv1.Job) (*batchv1.Job, error) {
@@ -135,6 +142,9 @@ func (h *jobHandler) syncNodeJob(job *batchv1.Job) (*batchv1.Job, error) {
 			} else if jobType == upgradeJobTypePostDrain && nodeState == nodeStatePostDraining {
 				logrus.Debugf("Post-drain job %s is done.", job.Name)
 				if repoInfo.Release.OS == node.Status.NodeInfo.OSImage {
+					if err = h.sendRestoreVMJob(upgrade, node, repoInfo); err != nil {
+						return job, err
+					}
 					setNodeUpgradeStatus(toUpdate, nodeName, StateSucceeded, "", "")
 					postDrained = true
 				} else {
@@ -147,6 +157,9 @@ func (h *jobHandler) syncNodeJob(job *batchv1.Job) (*batchv1.Job, error) {
 			} else if jobType == upgradeJobTypeSingleNodeUpgrade {
 				logrus.Debugf("Single-node-upgrade job %s is done.", job.Name)
 				if repoInfo.Release.OS == node.Status.NodeInfo.OSImage {
+					if err = h.sendRestoreVMJob(upgrade, node, repoInfo); err != nil {
+						return job, err
+					}
 					setNodeUpgradeStatus(toUpdate, nodeName, StateSucceeded, "", "")
 				} else {
 					setNodeUpgradeStatus(toUpdate, nodeName, nodeStateWaitingReboot, "", "")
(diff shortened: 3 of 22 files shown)


hoo29:
Sorry if I am missing the obvious but why is this needed in addition to the `harvesterhci.io/maintain-mode-strategy` labels? It would be great to get that annotation added as an exclusion to upgrader checker so we are not forced to shutdown VMs before an upgrade.

package/upgrade/upgrade_node.sh
```diff
@@ -723,8 +727,12 @@ command_single_node_upgrade() {
   NEW_OS_SQUASHFS_IMAGE_FILE=$(mktemp -p $UPGRADE_TMP_DIR)
   download_file "$UPGRADE_REPO_SQUASHFS_IMAGE" "$NEW_OS_SQUASHFS_IMAGE_FILE"
 
-  # Stop all VMs
-  shutdown_all_vms
+  # Shut down non-live migratable VMs
```
ibrokethecloud, line 710: should `--restore-vm` not be added by default for single node upgrades to preserve the current upgrade behaviour?

package/upgrade/upgrade_node.sh
```diff
@@ -723,8 +727,12 @@ command_single_node_upgrade() {
...
+  local upgrade_helper_args="--shutdown"
+  if [ "$HARVESTER_UPGRADE_RESTORE_VM" = "true" ]; then
+    upgrade_helper_args="$upgrade_helper_args --restore-vm --upgrade $HARVESTER_UPGRADE_NAME"
+  fi
+  upgrade-helper vm-live-migrate-detector "$HARVESTER_UPGRADE_NODE_NAME" $upgrade_helper_args
```
ibrokethecloud, line 735: on a single node upgrade this is going to be a noop

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -57,65 +99,247 @@ func (d *VMLiveMigrateDetector) Init() (err error) {
...
+		return names, nil
+	}
+	return virtualmachineinstance.GetAllNonLiveMigratableVMINames(vmis, nodes)
+}
+
+func getRestoreVMNames(vmis []*kubevirtv1.VirtualMachineInstance, nonMigratableVMNames []string) []string {
+	restoreVMs := make([]string, 0, 0)
+
+	// filter out paused VMs and upgrade repo VM from the non-migratable VMs
+	excludeVMs := make(map[string]struct{})
```
ibrokethecloud, line 255: is there a specific reason to include `paused` vms?

the upgrade vm should be migratable since it would be using the default storage class

cmd/upgradehelper/cmd/vmlivemigratedetector/vmlivemigratedetector.go
```diff
@@ -42,6 +46,8 @@ If there is no place to go, it can optionally shut down the VMs.
 
 func init() {
 	vmLiveMigrateDetectorCmd.Flags().BoolVar(&shutdown, "shutdown", false, "Shutdown non-migratable VMs")
+	vmLiveMigrateDetectorCmd.Flags().BoolVar(&restoreVM, "restore-vm", false, "Add non-migratable VMs info to ConfigMap, must be used with --upgrade {upgrade-name}")
```
ibrokethecloud: is there a case here `restoreVM` can be true but `shutdown` is false?

we could just drive all logic via `shutdown` flag, where it `shutdown` is true we just identify all non migratable vm's and store them in the configmap

pkg/controller/master/upgrade/node_controller.go
```diff
@@ -27,6 +28,8 @@ type nodeHandler struct {
 	upgradeClient ctlharvesterv1.UpgradeClient
 	upgradeCache  ctlharvesterv1.UpgradeCache
 	secretClient  ctlcorev1.SecretClient
+	jobClient     jobv1.JobClient
```
ibrokethecloud: does the nodeHandler need the job client/cache?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -263,10 +246,6 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 
 	// upgrade failed
 	if harvesterv1.UpgradeCompleted.IsFalse(upgrade) {
```
ibrokethecloud, line 248: should we schedule restoreVM job when an upgrade fails?

package/upgrade/upgrade_node.sh
brandboat: Thanks for the review :) 
Yes, you're right. The current behavior is as you described: `--restore-vm` is only added when the settings upgrade-config.restore-vm is set to true. By default, restore-vm is set to false.

cmd/upgradehelper/cmd/vmlivemigratedetector/vmlivemigratedetector.go
brandboat: > we could just drive all logic via shutdown flag, where it shutdown is true we just identify all non migratable vm's and store them in the configmap

Did you mean store all non-migratable VMs to configmap no matter the setting `restore-vm` is true or false?

In the beginning, I thought there is a case where `restore-vm` is set to false, but we still need to shut down all non-migratable VMs to make upgrade proceed. That's why I didn't put all the restore VMs logic behind the shutdown flag. But I agree with you that consolidating the logic under `shutdown` flag would make simplify things and also make debugging easier, since we can check the configmap to see which VMs are non-migratable. By the way this approach removes the need to pass extra `HARVESTER_UPGRADE_RESTORE_VM` env variable to the job.

pkg/controller/master/upgrade/node_controller.go
brandboat: Oops, leftover code, I'll clean it up, thanks for the heads-up!

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: Haha, looks like we’ve had this discussion before:
https://github.com/harvester/harvester/pull/8138#discussion_r2066895485
https://github.com/harvester/harvester/pull/8138#discussion_r2106369759

That said, I think this is a good starting point for another round of discussion.
c.c. @ihcsim @w13915984028

ibrokethecloud: thanks for the reminder, we can skip it for now until we can figure out a better way to identify likely failure scenarios

cmd/upgradehelper/cmd/vmlivemigratedetector/vmlivemigratedetector.go
ibrokethecloud: we only need to restore when we trigger a shutdown of vm's during the upgrade. if we shut them down we likely will attempt to start them unless there was a failed upgrade. so i am wondering if one flag is enough to trigger the entire behaviour

brandboat: Sounds great, I'll put all restore-vm logic behind the shutdown flag in the next commit, many thanks!

pkg/upgradehelper/restorevm/restorevm.go
```diff
@@ -0,0 +1,181 @@
...
+			return false, nil // keep retrying
+		}
+		return true, nil
+	})
+}
+
+func (h *RestoreVMHandler) getVMNamesFromConfigMap(ctx context.Context) ([]string, error) {
+	cmName := name.SafeConcatName(h.upgradeName, util.RestoreVMConfigMap)
+	cm, err := h.k8sClient.CoreV1().ConfigMaps(util.HarvesterSystemNamespaceName).Get(ctx, cmName, metav1.GetOptions{})
+	if err != nil {
```
w13915984028, line 165: check `IsNotFound` and log it?

if configmap is unluckily removed, log and return is ok

pkg/controller/master/upgrade/job_controller.go
```diff
@@ -283,3 +299,34 @@ func (h *jobHandler) setNodeWaitRebootLabel(node *v1.Node, repoInfo *repoinfo.Re
...
+	if err != nil {
+		logrus.WithFields(logrus.Fields{"name": upgrade.Name, "node": node.Name}).WithError(err).
+			Errorf("Failed to get setting UpgradeConfig, skip restore VM job")
+		return nil
+	}
+	// if restoreVM is false, we don't need to create restore VM job
+	if !restoreVM {
+		return nil
+	}
+	restoreVMJob := applyNodeJob(upgrade, repoInfo, node.Name, upgradeJobTypeRestoreVM)
```
w13915984028: the applyNodeJob is scheduled to the target node in general;  it can be used here, but might not be the best fit

need to skip the `witness` node;  or check the configmap to ensure this job is really needed for any given node

pkg/upgradehelper/restorevm/restorevm.go
```diff
@@ -0,0 +1,181 @@
...
+		return nil, fmt.Errorf("failed to get configmap: %w", err)
+	}
+	vmNamesStr, ok := cm.Data[h.nodeName]
+	if !ok {
+		return nil, fmt.Errorf("no VM names found for node %s", h.nodeName)
```
w13915984028: not every node has to-be-restored vm, `!ok` is a normal case

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -57,65 +99,247 @@ func (d *VMLiveMigrateDetector) Init() (err error) {
...
 
+	d.recordUpgradeEvent(corev1.EventTypeNormal, VMShutdownCompleted,
+		fmt.Sprintf("Shutdown completed for %d VM(s) on node %s, success: %d, failed: %d ", len(nonLiveMigratableVMNames), d.nodeName, vmSuccessCnt, vmFailedCnt))
+
 	return nil
 }
+
+func (d *VMLiveMigrateDetector) getAllNodes(ctx context.Context) ([]*v1.Node, error) {
+	// filter out witness nodes as they are not available for vm migration
+	nodeReq, err := labels.NewRequirement(harvNodeController.HarvesterWitnessNodeLabelKey, selection.DoesNotExist, nil)
```
w13915984028: accordingly, the restoreVM job should also skip witness node

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -57,65 +99,247 @@ func (d *VMLiveMigrateDetector) Init() (err error) {
...
+		}
+	}
+	return restoreVMs
+}
+func (d *VMLiveMigrateDetector) getConfigMapName() string {
+	return name.SafeConcatName(d.upgradeName, util.RestoreVMConfigMap)
+}
+
+func (d *VMLiveMigrateDetector) createOrUpdateConfigMap(ctx context.Context, restoreVMNames []string) error {
+	vmNames := strings.Join(restoreVMNames, ",")
```
w13915984028, line 231: vmNames can be an empty string when restoreVMNames is an empty slice

pkg/upgradehelper/vmlivemigratedetector/detector.go
```diff
@@ -57,65 +99,247 @@ func (d *VMLiveMigrateDetector) Init() (err error) {
...
 	if err != nil {
 		return err
 	}
-	nodes := make([]*v1.Node, 0, nodeList.Size())
-	for i := range nodeList.Items {
-		nodes = append(nodes, &nodeList.Items[i])
-	}
 
-	nonLiveMigratableVMNames, err := virtualmachineinstance.GetAllNonLiveMigratableVMINames(vmis, nodes)
+	nonLiveMigratableVMNames, err := getNonLiveMigratableVMNames(vmis, nodes)
```
w13915984028: it looks all none-livemigratable VMs are listed, from where, this list is further splited per the `namename`?

the `createOrUpdateConfigMap` bookes the list of VMs under the current node

w13915984028 (commented):
Thanks for the new commits, a few questions:

(1) The whole cluster might have 0 non-migrate-able VMs, the configmap data[nodename] can return empty string for any node;  I did not find where the listed non-migratable VMs are further splited per node

(2) Before fire the restore-vm job, a corresponding check is better.

(3) The current applyNode job looks not the best: with node-selector, priority, let's see if a better one can be figure out later.
```
								Privileged: &privileged,
								Capabilities: &corev1.Capabilities{
									Add: []corev1.Capability{
										corev1.Capability("CAP_SYS_BOOT"),
```

pkg/upgradehelper/vmlivemigratedetector/detector.go
brandboat: `getVMIListOptions` list VMIs for specific node, but the code here is quite misleading as there is an argument nodes... I'll refactor then. 
```go
nodeReq, err := labels.NewRequirement("kubevirt.io/nodeName", selection.Equals, []string{d.nodeName})
```

brandboat:
> (3) The current applyNode job looks not the best: with node-selector, priority, let's see if a better one can be figure out later.

@w13915984028, thanks for the comment, I'll do this in the next commit, like applyManifestsJob, I'll customize the restore-vm job.

cmd/upgradehelper/cmd/vmlivemigratedetector/vmlivemigratedetector.go
```diff
@@ -42,6 +44,7 @@ If there is no place to go, it can optionally shut down the VMs.
 
 func init() {
 	vmLiveMigrateDetectorCmd.Flags().BoolVar(&shutdown, "shutdown", false, "Shutdown non-migratable VMs")
+	vmLiveMigrateDetectorCmd.Flags().StringVar(&upgrade, "upgrade", "", "Upgrade CR name, speified this flag will store non-migratable VMs info to the ConfigMap")
```
ihcsim:
```suggestion
	vmLiveMigrateDetectorCmd.Flags().StringVar(&upgrade, "upgrade", "", "Upgrade CR name. If non-empty, it's used to generate the name of the configmap that stores all non-migratable VMs stopped by Harvester during the upgrade")
```

ihcsim:
@hoo29 We try to keep logic of upgrade support tasks outside of the Harvester controller. That way, we don't have to rely on the Harvester controller's availability, to decide, e.g., if the embedded node maintenance controller is back online, what version is at etc. Also, node maintenance doesn't work for single-node cluster.

pkg/upgradehelper/restorevm/restorevm.go
```diff
@@ -0,0 +1,197 @@
...
+		// wait for events to be flushed
+		time.Sleep(10 * time.Second)
+	}()
+
+	if err := h.factory.Sync(ctx); err != nil {
+		return fmt.Errorf("failed to sync factory: %w", err)
+	}
+
+	vmNames, err := h.getVMNamesFromConfigMap(ctx)
+	if err != nil && !errors.IsNotFound(err) {
```
w13915984028: L106 and L107 seems conflicting

w13915984028 (dismissed):
LGTM, thanks.

pkg/upgradehelper/restorevm/restorevm.go
brandboat: My bad, thanks for pointing that out.

w13915984028 (dismissed):
LGTM, thanks.

brandboat:
I've updated the PR a little bit, here is the diff link:
https://github.com/harvester/harvester/compare/56847df309239d54389170ee164a1e7fa7f5a4de..779a33e1a2b97d4ab884b927cf79aacc8f2bf70c

Removed the "harvesterhci.io/upgrade" filter from the VMI list options. This filter was causing single-node upgrades to get stuck because it excluded the upgrade repo VM from the VMI list at the start. 
And in https://github.com/harvester/harvester/issues/8298, we've decided to stop and restart the upgrade repo VM to avoid migration issues, making the filter unnecessary. Since this is a relatively small change, I think we can do this one shot (in this PR)

ihcsim (commented):
Code change LGTM. Just a suggestion on the `cmd` flag description.

ibrokethecloud (commented):
lgtm. thanks.

ibrokethecloud (approved):

PR #8533, opened 2025-06-24
chore: add new pr-backport capability for v1.6

Vicente-Cheng:
- Also remove the EoL v1.3 capability

#### Problem:
We need that capability because we will have the v1.6 branch for the upcoming release.

#### Solution:
Add those capability for mergify configuration


Yu-Jack (approved):
Thanks!

Issue #8504, opened 2025-06-18
[ENHANCEMENT] UI change to hide vmimages created by vm-import-controller


ibrokethecloud:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Users can import VM's from vmware/openstack installations using the vm-import-controller.

The vm-import-controller does this by creating `virtualmachineimages` for each disk on the source VM.

The `virtualmachineimages` appear as regular images in the Harvester UI, which means the user can use them directly with other VM's too. 

We need a way to prevent users from using these images like regular images with multiple vm's.

### Describe the solution you'd like

The latest release of vm-import-controller will add a new label `migration.harvesterhci.io/hideImportedImage=true` to the newly created `virtualmachineimages`

We need a UI change to filter out `virtualmachineimages` containing this label to hide them from the users.

a110605:
We need filter the vm image with label `migration.harvesterhci.io/hideImportedImage=true` in 
- image list page
- image count in dashboard
- image options in create VM page

houhoucoop:
@ibrokethecloud 
If we hide images from the image page, users might be confused when they still see them in the VM detail page?
Hiding also impacts counts across the UI, might easily cause UI defects, do we consider disabling instead of hiding it? e.g.

Disable image when creating VM
![Image](https://github.com/user-attachments/assets/2e798ca2-9ef6-491f-8eef-5461bc6d9d66)

Users can see the image in the list with all actions disabled, and optionally a tooltip indicating that the image was generated by the VM importer?
![Image](https://github.com/user-attachments/assets/50d0f85f-18d2-4c2b-a293-25c1317a665f)

ibrokethecloud:
@Vicente-Cheng i do like the idea of disabling the image, we could add a check in the vm validator if the pvc is not used in a vm managed by the vm-import-controller then vm creation would be rejected?

votdev:
Would it make sense to rename the label `migration.harvesterhci.io/hideImportedImage=true` to `migration.harvesterhci.io/hideImage=true` so that it can also be used universal for other scenarios? Perhaps a user also wants to hide other non-imported images.

ibrokethecloud:
Based on discussions, we will tweak the vm-import-controller to apply an new label `migration.harvesterhci.io/import:true` to the vm-import-controller managed images

There will be a change in the vm images api to hide actions for imported images and only allow the following actions:
* download
* download yaml
* delete 

remaining actions will be hidden from the user
[...]

PR #8537, opened 2025-06-24
Use `migration.harvesterhci.io/imported` label to check if VM is imported by vm-import-controller

votdev:
Make use of the label `migration.harvesterhci.io/imported` instead of the annotation `migration.harvesterhci.io/virtualmachineimport`.

There is no need to migrate existing annotations `migration.harvesterhci.io/virtualmachineimport` to `migration.harvesterhci.io/imported` label because the annotation was not used previously in Harvester code. It was only used by vm-import-controller itself.

Related to: https://github.com/harvester/harvester/pull/8481

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -24,7 +24,7 @@ import (
 	"github.com/harvester/harvester/pkg/util"
 	indexeresutil "github.com/harvester/harvester/pkg/util/indexeres"
 	"github.com/harvester/harvester/pkg/util/resourcequota"
-	vmUtil "github.com/harvester/harvester/pkg/util/virtualmachine"
+	vmutil "github.com/harvester/harvester/pkg/util/virtualmachine"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	indexerwebhook "github.com/harvester/harvester/pkg/webhook/indexeres"
 	"github.com/harvester/harvester/pkg/webhook/types"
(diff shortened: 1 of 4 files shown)


votdev:
Note, the `IsImportedByVMIC` is relocated into a separate file and is public because it was previously used in different locations. Check the commits for more information. We can cleanup if necessary and only change the annotation to label in the origin function.

votdev:
We need to backport this PR to v1.6 branch after merging.

ibrokethecloud (approved):

Vicente-Cheng (approved):

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -314,9 +314,9 @@ func (v *vmValidator) checkVolumeReq(newVM *kubevirtv1.VirtualMachine) error {
 		return nil
 	}
 
-	// if vm is being created by vm import controller then we need to skip pvc golden image validation
-	// this case needs to be treated differently from a standaard image scenario as user will boot directly from disk
-	imported := checkIsVirtualMachineImported(newVM)
+	// if vm is being created by vm import controller, then we need to skip pvc golden image validation
+	// this case needs to be treated differently from a standard image scenario as user will boot directly from disk
+	imported := util.IsImportedByVMIC(newVM)
```
w13915984028, line 319: a question posted here https://github.com/harvester/harvester/pull/8481#discussion_r2171471720

when basic usage of golden image is changed, looks more changes are needed to cover all cases.

e.g.
imported vm referes to other golden images, where to block;  
imported vm's golden image is cloned via related cdi annotation, where to block

w13915984028 (commented):
A question, thanks.

pkg/webhook/resources/virtualmachine/validator.go
ibrokethecloud: we cannot stop an end user from doing something stupid if they wish to. The only reason for basic checks is to ensure vm image create by vm import controller is not treated as a normal image. we are trying to address some of this via https://github.com/harvester/harvester/issues/8504

votdev:
@Mergifyio backport v1.6

pkg/webhook/resources/virtualmachine/validator.go
votdev: @ibrokethecloud is working on that.

w13915984028: suppose this case:

on a cluster, many VMs are imported and running,  later, user change one VM and selects one another golden image pvc accidentally/by mistake as there are many VMs and images, the above source code allows vm with imported label/annotation to set any golden image

ibrokethecloud: we could refine the logic to only allow attaching golden pvc's if they are also imported disks

Issue #8292, opened 2025-05-15
[BUG] VM Image Is Unexpectedly Recovered After Deletion


WebberHuang1118:
### Describe the Bug

After a VM image is recovered from a remote backup, deleting the image does not remove it as expected; the image reappears unexpectedly.

### To Reproduce

1. Create a VM image.
2. Create a VM using that VM image.
3. Back up the VM to a remote target.
4. After the VM backup completes, disconnect the remote backup target.
5. Delete the related VM, VM backup, and VM image.
6. Change the [Backupstore Poll Interval](https://longhorn.io/docs/1.7.2/references/settings/#backupstore-poll-interval) to 10.
7. Reconnect the same remote backup target. The VM image will be recovered with the `restore` type (along with the VM backup), for example:
    ```
    apiVersion: harvesterhci.io/v1beta1
    kind: VirtualMachineImage
    metadata:
      name: image-29t7d
      namespace: default
    spec:
      checksum: f7fdcbfa32b337c0ee7573b5abbf4ae5356def81f4656bde64124d83d32a9484476fb3c900f49ddca1d672af9067d1c7cfb42208338a01a89e413851b12c7d5f
      displayName: jammy-server-cloudimg-amd64.img
      pvcName: ""
      pvcNamespace: ""
      retry: 0
      sourceType: restore
      storageClassParameters:
        encrypted: "false"
        migratable: "true"
        numberOfReplicas: "1"
        staleReplicaTimeout: "30"
      url: s3://mybucket@pcloud/?backingImage=vmi-55dd2b3a-b005-4ea2-b57c-a91d477c84a6
    status:
      appliedUrl: s3://mybucket@pcloud/?backingImage=vmi-55dd2b3a-b005-4ea2-b57c-a91d477c84a6
      conditions:
      - lastUpdateTime: "2025-05-15T07:17:04Z"
        reason: Imported
        status: "True"
        type: Imported
      - lastUpdateTime: "2025-05-15T07:15:55Z"
        reason: Initialized
        status: "True"
        type: Initialized
      - lastUpdateTime: "2025-05-15T07:17:04Z"
        status: "False"
        type: RetryLimitExceeded
      failed: 0
      progress: 100
      size: 672464896
      storageClassName: longhorn-image-29t7d
      virtualSize: 2361393152
    ```
  8. Delete the recovered VM image and VM backup again.
[...]

PR #8544, opened 2025-06-25
fix: ensure VM image and corresponding remote metadata are deleted when the image type is `restore`

WebberHuang1118:
Ensure VM image and corresponding remote metadata are deleted when the image type is `restore`

#### Problem:
VM Image Is Unexpectedly Recovered After Deletion

#### Solution:
Ensure VM image and corresponding remote metadata are deleted when the image type is `restore`

#### Related Issue(s):
#8292 

#### Test plan:
1. Build Harvester with this PR
2. Create a VM image.
3. Create a VM using that VM image.
4. Back up the VM to a remote target.
5. After the VM backup completes, disconnect the remote backup target.
6. Delete the related VM, VM backup, and VM image.
7. Reconnect the same remote backup target and set the `refresh interval` to 10 seconds. The VM image will be recovered with the `restore` type (along with the VM backup)
8. Delete the recovered VM image and VM backup again.
9. The VM image will not be recovered again.

### DIFF
--- a/pkg/image/backingimage/backing_image_controller.go
+++ b/pkg/image/backingimage/backing_image_controller.go
@@ -6,9 +6,11 @@ import (
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	"k8s.io/apimachinery/pkg/api/errors"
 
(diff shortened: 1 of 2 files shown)


FrankYang0529 (approved):
LGTM. Thanks for the PR.

Vicente-Cheng (approved):
thanks for the fix!

PR #8546, opened 2025-06-25
feat: check storage and vm-migration network overlap

FrankYang0529:
#### Related Issue(s):
https://github.com/harvester/harvester/issues/5848

#### Test plan:
1. storage-network cannot overlap with vm-migration-network setting.
* Set storage-network setting as `{"clusterNetwork":"test-1","vlan":1,"range":"10.1.2.0/24"}` successfully.
* Set vm-migration-network setting as `{"clusterNetwork":"test-1","vlan":1,"range":"10.1.2.0/24"}` and get error like "vm-migration-network: the network configuration {''ClusterNetwork'': 'test-1', ''Vlan'': 1, ''Range'': '10.1.2.0/24',  ''Exclude'': []} is overlapped with another {''ClusterNetwork'': 'test-1', ''Vlan'': 1, ''Range'': '10.1.2.0/24',     ''Exclude'': []}.
* Set vm-migration-network setting as `{"clusterNetwork":"test-1","vlan":1,"range":"10.1.3.0/24"}` successfully.

2. vm-migration-network cannot overlap with storage-network setting.
* Set storage-network and vm-migration-network to default value.
* Set vm-migration-network setting as `{"clusterNetwork":"test-1","vlan":1,"range":"10.1.2.0/24"}` successfully.
* Set storage-network setting as `{"clusterNetwork":"test-1","vlan":1,"range":"10.1.2.0/24"}` and get error like "storage-network: the network configuration {''ClusterNetwork'': 'test-1', ''Vlan'': 1, ''Range'': '10.1.2.0/24',  ''Exclude'': []} is overlapped with another {''ClusterNetwork'': 'test-1', ''Vlan'': 1, ''Range'': '10.1.2.0/24',     ''Exclude'': []}".
* Set storage-network setting as `{"clusterNetwork":"test-1","vlan":1,"range":"10.1.3.0/24"}` successfully.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -32,6 +32,7 @@ import (
 	"golang.org/x/net/http/httpproxy"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -1036,7 +1037,7 @@ func validateUpdateContainerdRegistry(_ *v1beta1.Setting, newSetting *v1beta1.Se
 	return validateContainerdRegistry(newSetting)
 }
 
-func (v *settingValidator) validateNetworkHelper(value string) (*networkutil.Config, error) {
+func (v *settingValidator) validateNetworkHelper(name, value string) (*networkutil.Config, error) {
 	if value == "" {
 		// harvester will create a default setting with empty value
 		return nil, nil
@@ -1063,40 +1064,48 @@ func (v *settingValidator) validateNetworkHelper(value string) (*networkutil.Con
 		return nil, err
 	}
 
+	// Add a map of setting names to their specific network range validation callbacks
+	networkRangeValidators := map[string]func(*networkutil.Config) error{
+		settings.StorageNetworkName:            v.checkStorageNetworkRangeValid,
+		settings.VMMigrationNetworkSettingName: v.checkVMMigrationNetworkRangeValid,
+	}
+	if validator, ok := networkRangeValidators[name]; ok {
+		if err := validator(&config); err != nil {
+			return nil, err
+		}
+	}
 	return &config, nil
 }
 
-func (v *settingValidator) validateStorageNetworkHelper(value string) error {
-	config, err := v.validateNetworkHelper(value)
+func (v *settingValidator) getNetworkConfig(settingName string) (*networkutil.Config, error) {
+	if settingName != settings.StorageNetworkName && settingName != settings.VMMigrationNetworkSettingName {
+		return nil, nil
+	}
+
+	networkConfigSetting, err := v.settingCache.Get(settingName)
 	if err != nil {
-		return err
-	} else if config == nil {
-		// harvester will create a default setting with empty value
-		// storage-network will be the same in Longhorn, just skip check, don't require all VMs are stopped.
-		return nil
+		if apierrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get %s setting, err: %v", settingName, err)
 	}
 
-	if err := v.checkStorageNetworkRangeValid(config); err != nil {
-		return err
+	if networkConfigSetting.Value != "" {
+		var config networkutil.Config
+		if err := json.Unmarshal([]byte(networkConfigSetting.Value), &config); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal the %s setting value %v, %w", settingName, networkConfigSetting.Value, err)
+		}
+		return &config, nil
 	}
 
-	return nil
-}
-
-func (v *settingValidator) validateVMMigrationNetworkHelper(value string) error {
-	config, err := v.validateNetworkHelper(value)
-	if err != nil {
-		return err
-	} else if config == nil {
-		// harvester will create a default setting with empty value
-		return nil
+	if networkConfigSetting.Default != "" {
+		var config networkutil.Config
+		if err := json.Unmarshal([]byte(networkConfigSetting.Default), &config); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal the %s setting default value %v, %w", settingName, networkConfigSetting.Default, err)
+		}
+		return &config, nil
 	}
-
-	if err := v.checkVMMigrationNetworkRangeValid(config); err != nil {
-		return err
-	}
-
-	return nil
+	return nil, nil
 }
 
 func (v *settingValidator) validateStorageNetwork(setting *v1beta1.Setting) error {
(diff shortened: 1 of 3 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -1587,6 +1676,31 @@ func (v *settingValidator) checkVMMigrationNetworkRangeValid(config *networkutil
 	return nil
 }
 
+func checkStorageAndVMMigrationNetwokrOverlap(storageConfig, vmMigration *networkutil.Config) error {
```
WebberHuang1118: I think we can make the function signature more general and flexible — for example:
`func checkNetworkOverlap(c1, c2 *networkutil.Config)`

pkg/webhook/resources/setting/validator.go
```diff
@@ -1587,6 +1676,31 @@ func (v *settingValidator) checkVMMigrationNetworkRangeValid(config *networkutil
...
+		return err
+	}
+
+	for storageIP := range storageUsableIPAddresses {
+		for vmMigrationIP := range vmMigrationUsableIPAddresses {
+			if storageIP == vmMigrationIP {
+				return fmt.Errorf("storage-network and vm-migration-network have overlapping IP addresses: %s", storageIP)
+			}
+		}
+	}
```
WebberHuang1118: Would it be feasible to use a map to detect overlaps, rather than relying on O(n²) loop comparisons?

WebberHuang1118 (commented):
The functions `validateStorageNetwork()` and `validateVMMigrationNetwork()` are nearly identical — if I understand correctly, the only difference is the order in which the configs are checked.

The same pattern applies to the following function pairs:
- `validateUpdateStorageNetwork()` and `validateUpdateVMMigrationNetwork()`
- `validateDeleteStorageNetwork()` and `validateDeleteVMMigrationNetwork()`
- `validateStorageNetworkHelper()` and `validateVMMigrationNetworkHelper()`

Would it be feasible to factor out the common logic, and use function parameters or callbacks to handle the differing parts?

w13915984028:
To avoid cross affection, at this stage, we might just disallow the `storage-network` and `migration-network` to use same nad, namely, the vlan ID should be different when clusternetwork is same. It also has the benefit that with different VID, user can easily set hardware based bandwidth/QoS for them.

there are also storagenetwork related checks on networkcontroller side, the `migration-network` needs to be added accordingly;  or we use a general label like `protectednetwork` to hadle all the similar dedicated network nad.

https://github.com/harvester/network-controller-harvester/blob/651dcf2b297192990247f80aa2afcc40a1a81a1a/pkg/webhook/nad/validator.go#L107


`storage-network` nad is hide from UI selection when attaching it to VM, is `migration-network` also fltered?


BTW, below example is not accurate, normally, one VID is mapped to one subnet, map both vid 1/any to `10.1.2.0/24` and `10.1.3.0/24` is not a good practice. Spliting one subnet into two ranges for both of them to share in parallel increases the complexity.

```
Set storage-network setting as {"clusterNetwork":"test-1","vlan":1,"range":"10.1.2.0/24"} successfully.
Set vm-migration-network setting as {"clusterNetwork":"test-1","vlan":1,"range":"10.1.2.0/24"} and get error like "vm-migration-network: storage-network and vm-migration-network have overlapping IP addresses: 10.1.2.31".
Set vm-migration-network setting as {"clusterNetwork":"test-1","vlan":1,"range":"10.1.3.0/24"} successfully.
```

FrankYang0529:
@WebberHuang1118, I address all comments. Thanks for the review. 

> The functions validateStorageNetwork() and validateVMMigrationNetwork() are nearly identical.

Yeah, they're similar, so we encapsulate similar part in `validateNetworkHelper`.  The IP range check for storage-network and vm-migration-network are different. Also, the storage-network needs to do `checkStorageNetworkUsage`, but vm-migration-network doesn't.

WebberHuang1118 (commented):
`validateStorageNetworkHelper()` and `validateVMMigrationNetworkHelper()` are nearly identical:

```
func (v *settingValidator) validateStorageNetworkHelper(value string) (*networkutil.Config, error) {
	config, err := v.validateNetworkHelper(value)
	if err != nil {
		return nil, err
	} else if config == nil {
		// harvester will create a default setting with empty value
		// storage-network will be the same in Longhorn, just skip check, don't require all VMs are stopped.
		return nil, nil
	}

	if err := v.checkStorageNetworkRangeValid(config); err != nil {
		return nil, err
	}

	return config, nil
}

func (v *settingValidator) validateVMMigrationNetworkHelper(value string) (*networkutil.Config, error) {
	config, err := v.validateNetworkHelper(value)
	if err != nil {
		return nil, err
	} else if config == nil {
		// harvester will create a default setting with empty value
		return nil, nil
	}

	if err := v.checkVMMigrationNetworkRangeValid(config); err != nil {
		return nil, err
	}

	return config, nil
}
```

IMO, we could factor out the common logic into a utility function, passing in the validation function (e.g., `checkStorageNetworkRangeValid` or `checkVMMigrationNetworkRangeValid`) as a parameter.

Also, what @w13915984028 mentioned is worth considering:

> BTW, below example is not accurate, normally, one VID is mapped to one subnet, map both vid 1/any to 10.1.2.0/24 and 10.1.3.0/24 is not a good practice. Spliting one subnet into two ranges for both of them to share in parallel increases the complexity.

@w13915984028 — just to clarify, are you saying that two non-overlapping subnets can share the same VLAN ID? Or is it discouraged even if the subnets don’t overlap? cc @starbops

FrankYang0529:
> are you saying that two non-overlapping subnets can share the same VLAN ID? Or is it discouraged even if the subnets don’t overlap?

If it's just discouraged setting, but not error setting, we should not deny it in webhook validation.

w13915984028:
@WebberHuang1118 

the general practice is e.g.:
```
vlan 100 -> 192.168.100.0/24
vlan 110 -> 192.168.110.0/24
```

when saying, the `storage-network` and `migration-network` share vlan id 100, 
storage-network uses `192.168.100.0/24 range 1..128`,  migration-network uses `192.168.100.0/24 range 129..253`,  

I would prefer deny they use same VID (of course, different subnet) to simplify the network plan and operation

unless some day customers specifically require that those two should be able to share a vid/subnet and split the ip range

w13915984028:
Also noticed the https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/virtualmachine/validator.go did not check if VM refers to storage-network nad, even though from UI it is filtered, but non-UI has no check; needs to be added. as well the `migration-network`.

I have an ongoing PR https://github.com/harvester/harvester/pull/8360, will continue later.

w13915984028:
BTW:  The `storage-network` IP usages is related to the nodes/workloads, it can exhaust the IP range and needs user intervention

https://docs.harvesterhci.io/v1.5/advanced/storagenetwork#step-2

I did not check the source code yet: if expanding the IP ranges also requires the stopping of VMs; when true, we might need to enhance it via another tickets.

The `migration-network` is different, the usage is on demand, shortly and can fallback to the mgmt network much easily.

When migration is happending, both `storage-network` and `migration-network` might occupy high network bandwidth on writing, sharing them has potential risks. Spliting them is also a kind of protection upon `storage-network`.

Also refer https://docs.harvesterhci.io/v1.5/advanced/storagenetwork
> ... Users can allocate a dedicated storage network for Longhorn replication traffic to get better network bandwidth and performance.

WebberHuang1118:
> > are you saying that two non-overlapping subnets can share the same VLAN ID? Or is it discouraged even if the subnets don’t overlap?
> 
> If it's just discouraged setting, but not error setting, we should not deny it in webhook validation.

I agree with @FrankYang0529’s point — if this practice is merely discouraged (but not technically invalid), then we should document it as a best practice rather than outright reject it.

For users with limited physical resources, it wouldn’t be surprising if they want the storage network and migration network to share the same VLAN and physical interfaces.

In my opinion, the most critical concern is to ensure that the storage network has sufficient available IPs and is not exhausted by IP usage from the migration network.

pkg/util/network/common.go
```diff
@@ -15,6 +17,14 @@ type Config struct {
 	Exclude        []string `json:"exclude,omitempty"`
 }
 
+func (c *Config) ToString() string {
+	if c == nil {
+		return ""
+	}
+	return fmt.Sprintf("{''ClusterNetwork'': '%s', ''Vlan'': %d, ''Range'': '%s', ''Exclude'': %v}",
```
votdev: Wouldn't it better to use `json.Marshal` here?

FrankYang0529: Remove this function.

pkg/webhook/resources/setting/validator.go
```diff
@@ -1063,53 +1064,75 @@ func (v *settingValidator) validateNetworkHelper(value string) (*networkutil.Con
...
 
+	if name == settings.StorageNetworkName {
+		if err := v.checkStorageNetworkRangeValid(&config); err != nil {
+			return nil, err
+		}
+	} else if name == settings.VMMigrationNetworkSettingName {
+		if err := v.checkVMMigrationNetworkRangeValid(&config); err != nil {
+			return nil, err
+		}
+	}
```
WebberHuang1118: Here’s a suggestion to improve flexibility:

```suggestion
	// Add a map of setting names to their specific network range validation callbacks
	networkRangeValidators := map[string]func(*networkutil.Config) error{
		settings.StorageNetworkName:            v.checkStorageNetworkRangeValid,
		settings.VMMigrationNetworkSettingName: v.checkVMMigrationNetworkRangeValid,
	}

	if validator, ok := networkRangeValidators[name]; ok {
		if err := validator(&config); err != nil {
			return nil, err
		}
	}
```

WebberHuang1118 (commented):
LGTM. Please help check whether the suggested code restructuring is feasible. Thanks!

WebberHuang1118 (approved):
LGTM, thanks.

Issue #7601, opened 2025-02-13
[FEATURE] block registration of new nodes while upgrade is in progress


ibrokethecloud:
**Is your feature request related to a problem? Please describe.**

Harvester should block addition of new nodes while an upgrade is in progress.

This can have unwanted side effects if a user adds a node during the upgrade process.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

innobead:
Are there any other areas where we need to prevent this type of concurrent operation? cc @bk201 @starbops @tserong

bk201:
Would be nice to work this one together with
- https://github.com/harvester/harvester/issues/1860 
- https://github.com/harvester/harvester/issues/5278

ihcsim:
Do we know what are the risks of adding new nodes during an upgrade? I could think of 2 (or 3) reported cases where clusters were left in limbo, due to quorum loss (deleted nodes), kubelet/api server version mismatch (broken nodes) etc. It feels like one way to recover from these incidents was to add some form of rescue node to, either guarantee or re-establish the quorum, so that we can e.g., delete the broken nodes.

Felipalds:
Studying this issue and investigating further. Thinking in creating a guard [here](https://github.com/harvester/harvester/blob/b623da0fe73c75d7a2fed577ff4d6528c62b6e90/pkg/webhook/resources/node/validator.go).

I am also studying how Harvester does the upgrades and the CRD about the [upgrades](https://github.com/harvester/harvester/blob/507e99e446753c56ae064ed45825979f77974d8c/pkg/controller/master/node/node_volume_detach_controller.go#L125-L129).

```
1. Use the CRD that says that the cluster is being upgraded.
2. Add a admissionregv1.Create in this Webhook file
3. And then add the logic to block adding a node.
```

irishgordo:
@mingshuoqiu should this be tested on:
v1.7.0 -> master-head
Since no RC is cut yet?

irishgordo:
@mingshuoqiu , I've moved this back to implement


### I'm noticing an issue with this:
```
╭─mike at suse-workstation-team-harvester in ~/Projects/v1.7.1-stuff
╰─○ kubectl get nodes
NAME                 STATUS   ROLES                       AGE     VERSION
hp-103-tink-system   Ready    control-plane,etcd,master   2d21h   v1.33.5+rke2r1
[...]

PR #8550, opened 2025-06-25
feat: add new machine validator under webhook

Felipalds:
Issue https://github.com/harvester/harvester/issues/7601
Problem:
Adding a new node to a cluster can lead to unknown problems and even put the cluster into a limbo state

Solution:
Create a machine webhook to validate the intetions of the user to creating a new Node
Related Issue(s):
Test plan:
Additional documentation or context

### DIFF
--- a/pkg/webhook/resources/machine/validator.go
+++ b/pkg/webhook/resources/machine/validator.go
@@ -1,0 +1,40 @@
+package machine
+
+import (
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+func NewValidator(client client.Client) types.Validator {
+	return &machineValidator{
+		client: client,
+	}
+}
+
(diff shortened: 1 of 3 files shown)


pkg/webhook/resources/machine/validator.go
```diff
@@ -37,6 +45,28 @@ func (v *machineValidator) Resource() types.Resource {
...
+	if err != nil {
+		return false, err
+	}
+	for _, upgrade := range upgrades {
+		if !harvesterv1.UpgradeCompleted.IsFalse(upgrade) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
```
dharmit, line 72: IIUC, `isClusterUpgrading()` would return `true` even for an Upgrade that failed to complete in the past. Might be better off checking only the last upgrade in the `upgrades`, and check if it's in `Unknown` state because that, AFAIK, is the indicator of an upgrade in progress.

And, of course, I'm not sure if this is the right resource. My comment is only for the logic. :)

starbops: Yes, if any Upgrade CRs have a `Completed` condition of `Unknown`, it indicates an ongoing upgrade. Ideally, there should be only one Upgrade CR that is `Unknown` in the `Completed` condition.

Felipalds:
Closed accidentaly, will reopen this one today.

Issue #7931, opened 2025-03-24
[TASK] remove usage of Longhorn's `namespaceOverride` option


m-ildefons:
## What's the task? Please describe

Longhorn is deprecating the `namespaceOverride` option in its helm chart.
Audit Harvester code for use of said option and replace it with the corresponding hard-coded value.

## Additional context

https://github.com/longhorn/longhorn/issues/6527
https://github.com/longhorn/longhorn/issues/7311

Vicente-Cheng:
Hi @m-ildefons, 
Could you have the status update for this issue?
Thanks!

m-ildefons:
The current status of this issue is that Longhorn is *not* going to deprecate the `namespaceOverride` setting in its Helm chart anymore. See discussions here:
https://github.com/longhorn/longhorn/issues/6527
https://github.com/longhorn/longhorn/issues/7311

Therefore we can safely continue using this option to install Longhorn into the `longhorn-system` namespace.

The TL;DR for why this option wasn't removed from the Longhorn chart is that it doesn't work. When Longhorn is installed as dependency chart (as is the case for Harvester) it will inherit the install namespace. This creates several problems. To avoid this, Longhorn's Helm chart must support the `namespaceOverride` setting.

I've audited our charts for consistent use of either `.Values.longhorn.namespaceOverride` and hardcoded `longhorn-system` respectively and found that the Harvester Helm chart installs auxiliary resources for Longhorn (e.g. Prometheus rules) and some of these resources are templated with hardcoded namespace and others respect the `.Values.longhorn.namespaceOverride` setting.
To fix this [I've created a PR](https://github.com/harvester/harvester/pull/8575)

While this shouldn't change default behavior, I think it's important to make sure that the options offered in the `values.yaml` can't inadvertently break a Harvester installation when they are changed - even for options that users shouldn't ever need to change.

Vicente-Cheng:
> The current status of this issue is that Longhorn is _not_ going to deprecate the `namespaceOverride` setting in its Helm chart anymore. See discussions here: [longhorn/longhorn#6527](https://github.com/longhorn/longhorn/issues/6527) [longhorn/longhorn#7311](https://github.com/longhorn/longhorn/issues/7311)
> 
> Therefore we can safely continue using this option to install Longhorn into the `longhorn-system` namespace.

There is nothing needed to change. Close

PR #8575, opened 2025-06-30
make use of `namespaceOverride` consistent

m-ildefons:
Make consistent use of .Values.longhorn.namespaceOverride to specify install location of Longhorn.

related-to: harvester/harvester#7931

#### Problem:

Usage of `.Values.longhorn.namespaceOverride` needs to be consistent or replaced with hardcoded value. Since Longhorn requires this option or else it will be installed in the `harvester-system` namespace as well (which isn't good), the Harvester helm chart must make consistent use of the `.Values.longhorn.namespaceOverride` setting in every auxiliary resource that it installs into the Longhorn namespace.

The inconsistent use of this setting makes it impossible for users to customize the namespace in which Longhorn is installed in a Harvester cluster.

#### Solution:

Replace all hard-coded instances of `longhorn-system` in the templates installed by the Harvester Helm chart by `{{ .Values.longhorn.namespaceOverride }}`.

#### Related Issue(s):

#7931

#### Test plan:

Regression testing:
- Install Harvester
- Observe no errors, Longhorn should work normally, Prometheus rules should be correctly in place

N/A

### DIFF
--- a/deploy/charts/harvester/templates/configmap.yaml
+++ b/deploy/charts/harvester/templates/configmap.yaml
@@ -19,6 +19,7 @@ metadata:
 {{ include "harvester.labels" . | indent 4 }}
 data:
   promote.sh: |-
+    LONGHORN_NAMESPACE={{ .Values.longhorn.namespaceOverride | quote }}
     {{`KUBECTL="/host/$(readlink /host/var/lib/rancher/rke2/bin)/kubectl"
     YQ="/host/usr/bin/yq"
     ROLE_LABELS="rke.cattle.io/control-plane-role=true rke.cattle.io/etcd-role=true"
@@ -115,10 +116,10 @@ data:
     kickout_longhorn_node()
     {
       target=$1
-      found=$($KUBECTL get nodes.longhorn.io -n {{ .Values.longhorn.namespaceOverride }} |grep -q $target && echo true || echo false)
+      found=$($KUBECTL get nodes.longhorn.io -n $LONGHORN_NAMESPACE |grep -q $target && echo true || echo false)
       if [[ $found == true ]]; then
         echo "Found longhorn node $target, kicking it out..."
-        $KUBECTL delete nodes.longhorn.io $target -n {{ .Values.longhorn.namespaceOverride }}
+        $KUBECTL delete nodes.longhorn.io $target -n $LONGHORN_NAMESPACE
       fi
     }
 
--- a/deploy/charts/harvester/templates/configmap.yaml
+++ b/deploy/charts/harvester/templates/configmap.yaml
@@ -115,10 +115,10 @@ data:
     kickout_longhorn_node()
     {
       target=$1
-      found=$($KUBECTL get nodes.longhorn.io -n longhorn-system |grep -q $target && echo true || echo false)
+      found=$($KUBECTL get nodes.longhorn.io -n {{ .Values.longhorn.namespaceOverride }} |grep -q $target && echo true || echo false)
       if [[ $found == true ]]; then
         echo "Found longhorn node $target, kicking it out..."
-        $KUBECTL delete nodes.longhorn.io $target -n longhorn-system
+        $KUBECTL delete nodes.longhorn.io $target -n {{ .Values.longhorn.namespaceOverride }}
       fi
     }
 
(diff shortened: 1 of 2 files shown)


deploy/charts/harvester/templates/configmap.yaml
```diff
@@ -115,10 +115,10 @@ data:
     kickout_longhorn_node()
     {
       target=$1
-      found=$($KUBECTL get nodes.longhorn.io -n longhorn-system |grep -q $target && echo true || echo false)
+      found=$($KUBECTL get nodes.longhorn.io -n {{ .Values.longhorn.namespaceOverride }} |grep -q $target && echo true || echo false)
       if [[ $found == true ]]; then
         echo "Found longhorn node $target, kicking it out..."
-        $KUBECTL delete nodes.longhorn.io $target -n longhorn-system
+        $KUBECTL delete nodes.longhorn.io $target -n {{ .Values.longhorn.namespaceOverride }}
```
ihcsim: The variable interpolation won't work here because the embedded script is interpreted as a literal string (due to the enclosing escaping context).

Compare the templated output of this function with that of the RKE2 CIDRs (which are also `values` variables):
```
✗ helm template harvester  . | grep -m1 -A8 "kickout_longhorn_node"
    kickout_longhorn_node()
    {
      target=$1
      found=$($KUBECTL get nodes.longhorn.io -n {{ .Values.longhorn.namespaceOverride }} |grep -q $target && echo true || echo false)
      if [[ $found == true ]]; then
        echo "Found longhorn node $target, kicking it out..."
        $KUBECTL delete nodes.longhorn.io $target -n {{ .Values.longhorn.namespaceOverride }}
      fi
    }
```

```sh
✗ helm template harvester  . | grep -A2 cluster-cidr
    cluster-cidr: 10.52.0.0/16
    service-cidr: 10.53.0.0/16
    cluster-dns: 10.53.0.10
```

You might need to split the escape context into something like:

```diff
diff --git a/deploy/charts/harvester/templates/configmap.yaml b/deploy/charts/harvester/templates/configmap.yaml
index 108e8eef1..208b93b5b 100644
--- a/deploy/charts/harvester/templates/configmap.yaml
+++ b/deploy/charts/harvester/templates/configmap.yaml
@@ -111,14 +111,16 @@ data:
     $KUBECTL label --overwrite -n fleet-local machines.cluster.x-k8s.io $CUSTOM_MACHINE cluster.x-k8s.io/control-plane=true
     $KUBECTL label --overwrite -n fleet-local secret $PLAN_SECRET $ROLE_LABELS
     $KUBECTL label --overwrite -n fleet-local rkebootstraps.rke.cattle.io $CUSTOM_MACHINE $ROLE_LABELS
-
+    `}}
+    longhorn_namespace={{ .Values.longhorn.namespaceOverride | quote }}
+    {{`
     kickout_longhorn_node()
     {
       target=$1
-      found=$($KUBECTL get nodes.longhorn.io -n {{ .Values.longhorn.namespaceOverride }} |grep -q $target && echo true || echo false)
+      found=$($KUBECTL get nodes.longhorn.io -n $longhorn_namespace  |grep -q $target && echo true || echo false)
       if [[ $found == true ]]; then
         echo "Found longhorn node $target, kicking it out..."
-        $KUBECTL delete nodes.longhorn.io $target -n {{ .Values.longhorn.namespaceOverride }}
+        $KUBECTL delete nodes.longhorn.io $target -n $longhorn_namespace
       fi
     } 
```

This will force `helm template` to render:

```sh
 ✗ helm template harvester  . | grep -m1 -B3 -A9 "kickout_longhorn_node"
    
    longhorn_namespace="longhorn-system"
    
    kickout_longhorn_node()
    {
      echo "longhorn namespace is $longhorn_namespace"
      target=$1
      found=$($KUBECTL get nodes.longhorn.io -n $longhorn_namespace  |grep -q $target && echo true || echo false)
      if [[ $found == true ]]; then
        echo "Found longhorn node $target, kicking it out..."
        $KUBECTL delete nodes.longhorn.io $target -n $longhorn_namespace
      fi
    }
```

m-ildefons: Good spot. I hadn't seen that there were escaped sections in the script. I've fixed it now.

ihcsim (approved):

Vicente-Cheng (approved):
thanks

WebberHuang1118 (approved):
LGTM, thanks.

PR #8579, opened 2025-07-01
Expose CDI settings for day2 operations

brandboat:
#### Problem:
expose CDI settings for day 2 operations

#### Solution:
see HEP https://github.com/harvester/harvester/pull/8340

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8077

#### Test plan:
1. Make sure we have default annotation values for LVM & Longhorn V2 storage classes
   - For LVM, make sure we have annotation `cdi.harvesterhci.io/storageProfileVolumeModeAccessModes: '{"Block":["ReadWriteOnce"]}'`
   - For LHv2, make sure we have 2 annotations `cdi.harvesterhci.io/storageProfileCloneStrategy: copy`, and `cdi.harvesterhci.io/storageProfileVolumeSnapshotClass: longhorn-snapshot`.
2. Test all annotations, including both valid (positive) and invalid (negative) cases, covering scenarios for creating and updating a StorageClass. Below lists the newly added StorageClass annotations and their corresponding fields in the StorageProfile. When these annotations are added to a StorageClass, the backend updates the associated StorageProfile and CDI spec accordingly. Note that each StorageClass has a one-to-one mapping with a StorageProfile, and both share the same name.
    - `cdi.harvesterhci.io/filesystemOverhead`: backend will update the CDI object  `.spec.config.filesystemOverhead.storageClass`
    - `cdi.harvesterhci.io/storageProfileCloneStrategy`: backend will update the storage profile `.spec.cloneStrategy`
    - `cdi.harvesterhci.io/storageProfileVolumeSnapshotClass`: backend will update the storage profile `.spec.snapshotClass`
    - `cdi.harvesterhci.io/storageProfileVolumeModeAccessModes`: backend will update the storage profile `.spec.claimPropertySets`, e.g. providing `{"Block":["ReadWriteOnce"]}`, we'll have below yaml in storage profile
        ```yaml
        spec:
          claimPropertySets:
          - accessModes:
            - ReadWriteOnce
            volumeMode: Block
        ```
    Below are items we need to verify:
        - When annotations are added, the corresponding StorageProfile fields are correctly updated.
        - When annotations are removed, the corresponding StorageProfile fields are properly cleared.
        - When invalid values are provided in the annotations, the backend returns an appropriate error.


pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+	}
+	return nil
+}
+
+// updateCloneStrategy updates the CloneStrategy in the StorageProfile based on the StorageClass annotation.
+func updateCloneStrategy(sc *storagev1.StorageClass, profile *cdiv1.StorageProfile) {
+	if v, ok := sc.Annotations[util.AnnotationStorageProfileCloneStrategy]; ok {
+		if profile.Spec.CloneStrategy == nil || string(*profile.Spec.CloneStrategy) != v {
+			cloneStrategy, err := util.ToCloneStrategy(v)
```
Vicente-Cheng, line 184: we should do an associated check with snapshotclass if cloneStrategy is snapshot.

So, maybe you can change the order, to update the snapshotclass first, then use the result for the checking with cloneStrategy

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			if err != nil {
+				logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
+					WithError(err).
+					Warnf("Invalid clone strategy %s", v)
+			}
+			profile.Spec.CloneStrategy = cloneStrategy
+		}
+	} else if profile.Spec.CloneStrategy != nil {
+		profile.Spec.CloneStrategy = nil
```
Vicente-Cheng, line 191: I am wondering if we should have the default value of cloneStrategy, or the 3rd-party storage feature will fail.

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+		if err != nil {
+			logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
+				WithError(err).
+				Warnf("Invalid JSON for %s", util.AnnotationStorageProfileVolumeModeAccessModes)
+		}
+		if !equalsClaimPropertySets(profile.Spec.ClaimPropertySets, parsed) {
+			profile.Spec.ClaimPropertySets = parsed
+		}
+	} else if profile.Spec.ClaimPropertySets != nil {
+		profile.Spec.ClaimPropertySets = nil
```
Vicente-Cheng, line 230: the same situation of cloneStrategy or we could do some validation here?
I mean to check if there is any default value of CDI hard-coded. If not, perhaps consider sending an event or giving the default?

pkg/webhook/resources/storageclass/mutator.go
```diff
@@ -0,0 +1,91 @@
...
+						patchOps = append(patchOps, fmt.Sprintf(
+							patchAnnotation,
+							patch.EscapeJSONPointer(util.AnnotationStorageProfileSnapshotClass),
+							"longhorn-snapshot",
+						))
+					}
+				}
+			}
+		}
+	case util.CSIProvisionerLVM:
```
Vicente-Cheng, line 61: iiuc, we need to patch the cloneStrategy and snapshotclass on our lvm csi driver also?

brandboat: > iiuc, we need to patch the cloneStrategy and snapshotclass on our lvm csi driver also?

@Vicente-Cheng, I don't see that in storage_profile_controller.go. https://github.com/harvester/harvester/blob/master/pkg/controller/master/storageprofile/storage_profile_controller.go#L63-L75
So the cloneStrategy should be "snapshot" & snapshotclass should be lvm-snapshot, right?

pkg/controller/master/storageclass/storage_class_controller.go
brandboat: From what I understand, if cloneStrategy is not specified (nil), CDI will default it to "copy", see https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/controller/storageprofile-controller.go#L187-L203

We still have the option to explicitly provide a default value if needed. What are your thoughts on this?

pkg/controller/master/storageclass/storage_class_controller.go
brandboat: You're right, thanks for the comment!

pkg/controller/master/storageclass/storage_class_controller.go
brandboat: Discuss offline with @Vicente-Cheng, we can leverage on https://github.com/kubevirt/containerized-data-importer/blob/v1.61.1/pkg/storagecapabilities/storagecapabilities.go#L35, if user don't provide volume mode & access modes and there is no default value in the above link, we'll return error in the validator to notify user to fill in values.

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+func (h *storageClassHandler) syncCDI(sc *storagev1.StorageClass) error {
+	// If the storage class is Longhorn v1, we don't need to sync CDI settings or
+	// storage profile, as we don't support CDI for Longhorn v1.
+	if sc.Provisioner == util.CSIProvisionerLonghorn {
+		if v, ok := sc.Parameters["dataEngine"]; ok {
+			if v == string(longhornv1.DataEngineTypeV1) {
+				return nil
+			}
+		}
+	}
```
WebberHuang1118: Suggest wrapping this as a utility function to improve readability

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
+				WithError(err).
+				Errorf("Failed to update CDI %s", util.CDIObjectName)
+			return fmt.Errorf("failed to update CDIConfig: %w", err)
+		}
+	}
+
+	return nil
+}
+
```
WebberHuang1118: Suggestion to make the logic more concise and easier to read:
```suggestion
// syncCDISettings updates the CDI based on the annotations of the StorageClass.
func (h *storageClassHandler) syncCDISettings(sc *storagev1.StorageClass) error {
	cdi, err := h.cdiCache.Get(util.CDIObjectName)
	if err != nil {
		if apierrors.IsNotFound(err) {
			logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).Warnf("CDI %s not found, skipping update", sc.Name)
			return nil
		}
		return err
	}

	toUpdate := cdi.DeepCopy()

	v, hasAnno := sc.Annotations[util.AnnotationCDIFSOverhead]
	if hasAnno {
		if !regexp.MustCompile(util.FSOverheadRegex).MatchString(v) {
			logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).Warnf("Invalid filesystem overhead %s, not match %s, skipping update", v, util.FSOverheadRegex)
			return nil
		}
		setFilesystemOverhead(toUpdate, sc.Name, v)
	} else {
		removeFilesystemOverhead(toUpdate, sc.Name)
	}

	if reflect.DeepEqual(toUpdate, cdi) {
		return nil
	}

	if _, err := h.cdiClient.Update(toUpdate); err != nil {
		logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).WithError(err).Errorf("Failed to update CDI %s", util.CDIObjectName)
		return fmt.Errorf("failed to update CDIConfig: %w", err)
	}

	return nil
}

// setFilesystemOverhead sets the filesystem overhead for a storage class in the CDI config.
func setFilesystemOverhead(cdi *cdiv1.CDI, scName, value string) {
	if cdi.Spec.Config == nil {
		cdi.Spec.Config = &cdiv1.CDIConfigSpec{}
	}
	if cdi.Spec.Config.FilesystemOverhead == nil {
		cdi.Spec.Config.FilesystemOverhead = &cdiv1.FilesystemOverhead{}
	}
	if cdi.Spec.Config.FilesystemOverhead.StorageClass == nil {
		cdi.Spec.Config.FilesystemOverhead.StorageClass = make(map[string]cdiv1.Percent, 1)
	}
	if string(cdi.Spec.Config.FilesystemOverhead.StorageClass[scName]) != value {
		cdi.Spec.Config.FilesystemOverhead.StorageClass[scName] = cdiv1.Percent(value)
	}
}

// removeFilesystemOverhead removes the filesystem overhead entry for a storage class if it exists.
func removeFilesystemOverhead(cdi *cdiv1.CDI, scName string) {
	if cdi.Spec.Config != nil && cdi.Spec.Config.FilesystemOverhead != nil && cdi.Spec.Config.FilesystemOverhead.StorageClass != nil {
		if _, ok := cdi.Spec.Config.FilesystemOverhead.StorageClass[scName]; ok {
			delete(cdi.Spec.Config.FilesystemOverhead.StorageClass, scName)
		}
	}
}
```

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+				WithError(err).
+				Errorf("Failed to update StorageProfile %s", profileName)
+			return err
+		}
+	}
+	return nil
+}
+
+// updateCloneStrategy updates the CloneStrategy in the StorageProfile based on the StorageClass annotation.
+func updateCloneStrategy(sc *storagev1.StorageClass, profile *cdiv1.StorageProfile) {
```
WebberHuang1118: For the function signature, I suggest using:
```
func updateCloneStrategy(sc *storagev1.StorageClass, profile *cdiv1.StorageProfile) (*cdiv1.StorageProfile, error)
```
This makes the return values clear and explicit.

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+		_, err := h.storageProfileClient.Update(toUpdate)
+		if err != nil {
+			logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
+				WithError(err).
+				Errorf("Failed to update StorageProfile %s", profileName)
+			return err
+		}
+	}
+	return nil
+}
```
WebberHuang1118: Suggestion to make the logic more concise and easier to read:

```suggestion
func (h *storageClassHandler) syncStorageProfile(sc *storagev1.StorageClass) error {
    profileName := sc.Name
    profile, err := h.storageProfileCache.Get(profileName)
    if err != nil {
        if apierrors.IsNotFound(err) {
            logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
                Warnf("StorageProfile %s not found, skipping update", profileName)
            return nil
        }
        return err
    }

    toUpdate := profile.DeepCopy()
    updateCloneStrategy(sc, toUpdate)

    if err := h.updateSnapshotClass(sc, toUpdate); err != nil {
        logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
            WithError(err).
            Errorf("Failed to update StorageProfile %s", profileName)
        return err
    }

    updateClaimPropertySets(sc, toUpdate)

    if reflect.DeepEqual(toUpdate, profile) {
        return nil
    }

    if _, err := h.storageProfileClient.Update(toUpdate); err != nil {
        logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
            WithError(err).
            Errorf("Failed to update StorageProfile %s", profileName)
        return err
    }
    return nil
}
```

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
+				Warnf("StorageProfile %s not found, skipping update", profileName)
+			return nil
+		}
+		return err
+	}
+
+	toUpdate := profile.DeepCopy()
+
+	updateCloneStrategy(sc, toUpdate)
```
WebberHuang1118: We likely don’t have a silent failure in this case.

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+	if err := h.updateSnapshotClass(sc, toUpdate); err != nil {
+		logrus.WithFields(logrus.Fields{"storageclass": sc.Name}).
+			WithError(err).
+			Errorf("Failed to update StorageProfile %s", profileName)
+		return err
+	}
+	updateClaimPropertySets(sc, toUpdate)
```
WebberHuang1118: We likely don’t have a silent failure in this case.

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			profile.Spec.SnapshotClass = &v
+		}
+	} else if profile.Spec.SnapshotClass != nil {
+		profile.Spec.SnapshotClass = nil
+	}
+	return nil
+}
+
+// updateClaimPropertySets updates the ClaimPropertySets in the StorageProfile based on the StorageClass annotation.
+func updateClaimPropertySets(sc *storagev1.StorageClass, profile *cdiv1.StorageProfile) {
```
WebberHuang1118: ditto

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -327,3 +351,162 @@ func (v *storageClassValidator) validateReservedStorageClass(sc *storagev1.Stora
...
+type cdiAnnotations struct {
+	fsOverhead                  string
+	fsOverheadExists            bool
+	cloneStrategy               string
+	cloneStrategyExists         bool
+	snapshotClass               string
+	snapshotClassExists         bool
+	volumeModeAccessModes       string
+	volumeModeAccessModesExists bool
+}
```
WebberHuang1118: Do we really need an additional structure here? Can we use sc directly in validateCDIAnnotations() instead?
Introducing a new structure would require us to ensure synchronization between the related resources, which could add unnecessary complexity.

pkg/webhook/resources/storageclass/mutator.go
```diff
@@ -0,0 +1,105 @@
...
+			patchOps = append(patchOps, fmt.Sprintf(
+				patchAnnotation,
+				patch.EscapeJSONPointer(util.AnnotationStorageProfileSnapshotClass),
+				strconv.Quote("lvm-snapshot"),
+			))
+		}
+	}
+
+	return patchOps
+}
```
WebberHuang1118: Can we refactor the multiple switch/if conditions into smaller functions? This would improve readability and make the code easier to maintain.

WebberHuang1118 (commented):
Hi @brandboat,

Thanks for the PR! Please take a look at my feedback when you have a moment. Thanks

pkg/controller/master/storageclass/storage_class_controller.go
brandboat: Thanks for the suggestion! Since the method directly modifying the passed-in storageprofile, IMO there's no need to return it. Perhaps changed it to the below code?
```
func updateCloneStrategy(sc *storagev1.StorageClass, profile *cdiv1.StorageProfile) (error)
```

pkg/webhook/resources/storageclass/validator.go
brandboat: OK, I'll refactor out cdiAnnotations, thanks!

pkg/webhook/resources/storageclass/mutator.go
brandboat: Thanks for the comments! I've addressed everything in the latest commit, PTAL.

pkg/controller/master/storageclass/storage_class_controller.go
WebberHuang1118: This adds implicit behavior to the function: the **profile** variable passed as a parameter will be mutated within the function and used afterward.

As a result, anyone using this function has to dig into its implementation to understand the behavior, rather than getting a clear idea from the function’s signature.

WebberHuang1118 (commented):
Hi @brandboat,

Regarding [this comment](https://github.com/harvester/harvester/pull/8579#discussion_r2194169493):

IMO, it would be better for the following functions to return the updated profile:
- `updateCloneStrategy()`
- `updateSnapshotClass()`
- `updateClaimPropertySets()`

While it's acceptable to pass the same variable as both input and to receive modifications through return values, having an explicit return in the function signature makes the behavior clearer and easier to reason about.

pkg/controller/master/storageclass/register.go
```diff
@@ -0,0 +1,30 @@
+package storageclass
+
+import (
+	"context"
+
+	"github.com/harvester/harvester/pkg/config"
+)
+
+const (
+	storageClassControllerName = "storage-class-controller"
```
Vicente-Cheng: could we have harvester- prefix? e.g. harvester-storage-class-controller

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,259 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func isLonghornV1(sc *storagev1.StorageClass) bool {
+	// Check if the storage class is Longhorn v1 by checking the provisioner and parameters
+	return sc.Provisioner == util.CSIProvisionerLonghorn &&
+		sc.Parameters["dataEngine"] == string(longhornv1.DataEngineTypeV1)
```
Vicente-Cheng, line 65: `sc.Parameters` could be nil. We need to check with that.

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -327,3 +351,174 @@ func (v *storageClassValidator) validateReservedStorageClass(sc *storagev1.Stora
...
+	if !hasCDIAnnotations(sc) {
+		return nil
+	}
+
+	if sc.Provisioner != util.CSIProvisionerLonghorn {
+		return nil
+	}
+
+	// Longhorn v1 does not support CDI annotations
+	if dataEngine, ok := sc.Parameters["dataEngine"]; ok && dataEngine == string(longhorn.DataEngineTypeV1) {
```
Vicente-Cheng, line 388: check parameters as above suggestion

Vicente-Cheng (commented):
some nits

pkg/controller/master/storageclass/storage_class_controller.go
brandboat: Thanks! In Go, I found  it's safe to read from a nil map — it just returns the empty value from nil map, so we don’t need to explicitly check for nil in this case.”

Here is an example: https://go.dev/play/p/eJ-XT1RjktD

pkg/webhook/resources/storageclass/validator.go
brandboat: ditto https://github.com/harvester/harvester/pull/8579#discussion_r2206493574

WebberHuang1118 (approved):
LGTM, thanks.

pkg/controller/master/storageclass/storage_class_controller.go
Vicente-Cheng: Got it, Thanks for the explanation!

Vicente-Cheng (approved):
thanks!

Vicente-Cheng:
Hi @ihcsim, 
Could you check with this one?

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,259 @@
...
+func (h *storageClassHandler) syncCDI(sc *storagev1.StorageClass) error {
+	// If the storage class is Longhorn v1, we don't need to sync CDI settings or
+	// storage profile, as we don't support CDI for Longhorn v1.
+	if isLonghornV1(sc) {
+		return nil
+	}
+
+	if err := h.syncCDISettings(sc); err != nil {
+		return err
+	}
```
ihcsim, line 54: Do we know if this update to the `cdi` resource is race-free? If not, will a failure from the `Update()` call (e.g., due to the `object has been modified` error) at line 92 cause the controller to requeue at the request, or do we need to [retry-on-conflict](https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict)?

pkg/controller/master/storageclass/storage_class_controller.go
```diff
@@ -0,0 +1,259 @@
...
+			profile.Spec.ClaimPropertySets = parsed
+		}
+	} else if profile.Spec.ClaimPropertySets != nil {
+		profile.Spec.ClaimPropertySets = nil
+	}
+	return profile, nil
+}
+
+// equalsClaimPropertySets compares two slices of ClaimPropertySet.
+// It returns true if both slices contain the same ClaimPropertySets, regardless of order.
```
ihcsim, line 236: Entirely optional, we can also consider using the [`k8s.io/apimachinery/pkg/conversion#Equalities`](https://pkg.go.dev/k8s.io/apimachinery@v0.33.2/pkg/conversion#Equalities) type to perform semantic comparison on K8s resources or CRDs.

E.g., with this diff containing an implementation of semantic comparison on the `ClaimPropertySet` type, the unit tests will still pass:

```diff
diff --git a/pkg/controller/master/storageclass/storage_class_controller.go b/pkg/controller/master/storageclass/storage_class_controller.go
index 37130c351..b1dd83926 100644
--- a/pkg/controller/master/storageclass/storage_class_controller.go
+++ b/pkg/controller/master/storageclass/storage_class_controller.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/conversion"
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 
 	ctlcdiv1 "github.com/harvester/harvester/pkg/generated/controllers/cdi.kubevirt.io/v1beta1"
@@ -232,28 +233,43 @@ func updateClaimPropertySets(sc *storagev1.StorageClass, profile *cdiv1.StorageP
 	return profile, nil
 }
 
+func claimPropertySetSemanticallyEqual(a, b cdiv1.ClaimPropertySet) bool {
+	equalities := conversion.EqualitiesOrDie(
+		func(a, b *corev1.PersistentVolumeMode) bool {
+			return a == b
+		},
+		func(a, b []corev1.PersistentVolumeAccessMode) bool {
+			sort.Slice(a, func(i, j int) bool {
+				return a[i] < a[j]
+			})
+			sort.Slice(b, func(i, j int) bool {
+				return b[i] < b[j]
+			})
+			return slices.Equal(a, b)
+		},
+	)
+
+	return equalities.DeepEqual(a, b)
+}
+
 // equalsClaimPropertySets compares two slices of ClaimPropertySet.
 // It returns true if both slices contain the same ClaimPropertySets, regardless of order.
 func equalsClaimPropertySets(a, b []cdiv1.ClaimPropertySet) bool {
-	normalize := func(list []cdiv1.ClaimPropertySet) []cdiv1.ClaimPropertySet {
-		normalized := make([]cdiv1.ClaimPropertySet, len(list))
-		for i, item := range list {
-			accessModes := append([]corev1.PersistentVolumeAccessMode(nil), item.AccessModes...)
-			// Sort access modes and add to normalized
-			slices.Sort(accessModes)
-			normalized[i] = cdiv1.ClaimPropertySet{
-				AccessModes: accessModes,
-				VolumeMode:  item.VolumeMode,
-			}
-		}
-		// Sort volume modes
-		sort.Slice(normalized, func(i, j int) bool {
-			return *normalized[i].VolumeMode < *normalized[j].VolumeMode
-		})
-		return normalized
+	if len(a) != len(b) {
+		return false
 	}
 
-	normA := normalize(a)
-	normB := normalize(b)
-	return reflect.DeepEqual(normA, normB)
+	sort.Slice(a, func(i, j int) bool {
+		return *(a[i].VolumeMode) < *(a[j].VolumeMode)
+	})
+	sort.Slice(b, func(i, j int) bool {
+		return *(b[i].VolumeMode) < *(b[j].VolumeMode)
+	})
+
+	for i := range a {
+		if !claimPropertySetSemanticallyEqual(a[i], b[i]) {
+			return false
+		}
+	}
+	return true
 }
```

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -327,3 +351,174 @@ func (v *storageClassValidator) validateReservedStorageClass(sc *storagev1.Stora
...
+	}
+
+	value, exists := sc.Annotations[util.AnnotationCDIFSOverhead]
+	if !exists {
+		return nil
+	}
+
+	if !regexp.MustCompile(util.FSOverheadRegex).MatchString(value) {
+		return werror.NewInvalidError(
+			fmt.Sprintf("invalid filesystem overhead %s in annotation %s, must be in range [0, 1]",
```
ihcsim, line 427:
```suggestion
			fmt.Sprintf("invalid filesystem overhead %s in annotation %s, must be in the range [0.00, 1.00] (up to 3 decimal places)",
```

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -327,3 +351,174 @@ func (v *storageClassValidator) validateReservedStorageClass(sc *storagev1.Stora
...
+
+	value, exists := sc.Annotations[util.AnnotationStorageProfileVolumeModeAccessModes]
+	if exists {
+		return v.validateVolumeModeAccessModesAnnotation(value)
+	}
+
+	// Check if annotation is required for this provisioner
+	if _, ok := cdicaps.CapabilitiesByProvisionerKey[sc.Provisioner]; !ok {
+		return werror.NewInvalidError(
+			fmt.Sprintf("missing annotation %s", util.AnnotationStorageProfileVolumeModeAccessModes), "")
```
ihcsim, line 507:
```suggestion
			fmt.Sprintf("missing annotation %s. volume access modes are required for CDI integration to work with storage class provisioner %s.", util.AnnotationStorageProfileVolumeModeAccessModes, sc.Provisioner), "")
```

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -327,3 +351,174 @@ func (v *storageClassValidator) validateReservedStorageClass(sc *storagev1.Stora
...
+		}
+	}
+
+	return nil
+}
+
+func (v *storageClassValidator) validateLonghornV1(sc *storagev1.StorageClass) error {
+	if !hasCDIAnnotations(sc) {
+		return nil
+	}
```
ihcsim, line 381: Doesn't it make sense to move this to the parent `validateCDIAnnotations()` method? If there are no CDI annotations, there is no need to proceed to the other `validators`.

pkg/webhook/resources/storageclass/mutator.go
```diff
@@ -0,0 +1,146 @@
...
+			cdiv1.CloneStrategyHostAssisted,
+		))
+	}
+
+	// Add snapshot class annotation
+	if shouldAddAnnotation(sc, util.AnnotationStorageProfileSnapshotClass) {
+		patchOps = append(patchOps, fmt.Sprintf(
+			patchAnnotation,
+			patch.EscapeJSONPointer(util.AnnotationStorageProfileSnapshotClass),
+			"longhorn-snapshot",
```
ihcsim, line 89: Is this a reliable default? Or shall we read it from the `csi-driver-config` setting? Same question for LVM CSI.

pkg/controller/master/storageclass/storage_class_controller.go
brandboat: When a conflict error pops up, Wrangler just triggers another update event. That sends it back to the reconciler, which triggers the storage_class_controller again. This makes me wonder if we even need retryOnConflict. I've seen other Harvester controllers skip this too, and I'm honestly not sure if that's okay...

pkg/controller/master/storageclass/storage_class_controller.go
brandboat: Much cleaner, thanks for the comment!

pkg/webhook/resources/storageclass/validator.go
brandboat: Sadly we can't, in validateVolumeModeAccessModes, we have a validation rule that requires the volume access modes annotation to be present when the storage class provisioner is not found in the CDI capabilities map. This ensures CDI integration works correctly with unknown provisioners. btw, this idea comes from https://github.com/harvester/harvester/pull/8579#discussion_r2180628451

```
	if _, ok := cdicaps.CapabilitiesByProvisionerKey[sc.Provisioner]; !ok {
		return werror.NewInvalidError(
			fmt.Sprintf("missing annotation %s. volume access modes are required for CDI integration to work with storage class provisioner %s.",
				util.AnnotationStorageProfileVolumeModeAccessModes, sc.Provisioner), "")
	}
```

Vicente-Cheng:
Hi @ihcsim, 
Consider the rc2 schedule, @brandboat will create another PR for your suggestion.
We will merge this first. Thanks!

pkg/webhook/resources/storageclass/mutator.go
brandboat: You're right, I missed that.

brandboat:
gentle ping @ihcsim, the follow-up PR is here: https://github.com/harvester/harvester/pull/8679

Issue #8571, opened 2025-06-30
[ENHANCEMENT] protect the witness label and taint


Vicente-Cheng:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

If the end user removes the witness label, it would introduce some issues as follows:
- upgrade hang
- workload will be scheduled to witness node

### Describe the solution you'd like

Block the operation for the witness nodes related to label/taint

dharmit:
I'd like to work on this. Can I pick it up?

Vicente-Cheng:
Sure!

Thanks @dharmit

Vicente-Cheng:
PR was merged into the master branch, we could verify on v1.7.0 sprint release

m-ildefons:
@dharmit could you please fill out the ready-for-testing checklist?

m-ildefons:
## Setup

Set up Harvester v1.7.0-dev-20250914 with two control-plane and one witness nodes
```
│ ~ │► kubectl get settings.harvesterhci.io server-version 
NAME             VALUE
server-version   v1.7.0-dev-20250914
```

<img width="2021" height="149" alt="Image" src="https://github.com/user-attachments/assets/b97dde07-a3ce-4621-9489-a96f5201b572" />

## Test 1 - Remove Taint from Witness Node

🟢 Removal is blocked

<img width="1578" height="85" alt="Image" src="https://github.com/user-attachments/assets/fb517c25-a8f0-49a8-892a-79082d8f9bfc" />

🟢 Witness node still has taint

<img width="775" height="164" alt="Image" src="https://github.com/user-attachments/assets/c56880b5-91b3-4d57-bf3e-ea58bba07234" />

## Test 2 - Remove Label `node-role.harvesterhci.io/witness` from witness node

🟢 Removal is blocked

<img width="1713" height="63" alt="Image" src="https://github.com/user-attachments/assets/a26020e5-3114-4922-b9dd-89c7a548477a" />

🟢 Witness node still has label `node-role.harvesterhci.io/witness`

<img width="2574" height="996" alt="Image" src="https://github.com/user-attachments/assets/1bc61960-ed11-439e-8177-4e5d440370d5" />

## Test 3 - Remove label `node-role.kubernetes.io/etcd` from witness node

🔴 Removal is not blocked

<img width="722" height="40" alt="Image" src="https://github.com/user-attachments/assets/e0ec7e63-c27a-46d1-b8e0-106995c2a809" />

[...]

PR #8587, opened 2025-07-02
Protect Witness node taint and label

dharmit:
#### Problem:

Explained in #8571 

#### Solution:

Adds a method to check if user is removing either the required witness node label or taint.

#### Related Issue(s):

#8571 
#### Test plan:

- Set up a Harvester cluster with a Witness node
- Try removing the taint `node-role.kubernetes.io/etcd` or the label `node-role.harvesterhci.io/witness` from it
- Removal should fail

### DIFF
--- a/pkg/webhook/resources/node/validator.go
+++ b/pkg/webhook/resources/node/validator.go
@@ -65,6 +65,9 @@ func (v *nodeValidator) Update(_ *types.Request, oldObj runtime.Object, newObj r
 	if err := v.validateCPUManagerOperation(newNode); err != nil {
 		return err
 	}
+	if err := validateWitnessRoleChange(oldNode, newNode); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/node/validator.go
```diff
@@ -134,6 +137,43 @@ func (v *nodeValidator) validateCPUManagerOperation(node *corev1.Node) error {
...
+	}
+
+	var witnessTaintPersist, witnessLabelPersist bool
+	for _, taint := range newNode.Spec.Taints {
+		if taint.Key == "node-role.kubernetes.io/etcd" && taint.Value == "true" && taint.Effect == corev1.TaintEffectNoExecute {
+			witnessTaintPersist = true
+			break
+		}
+	}
+	if val, ok := newNode.ObjectMeta.Labels[util.HarvesterWitnessNodeLabelKey]; ok || val == "true" {
```
brandboat: `ok || val == "true"` this condition looks weird, if label doesn't exist, `ok` should be `false`, and `val` is always empty string.

pkg/webhook/resources/node/validator.go
```diff
@@ -134,6 +137,43 @@ func (v *nodeValidator) validateCPUManagerOperation(node *corev1.Node) error {
...
+		return nil
+	}
+
+	var witnessTaintPersist, witnessLabelPersist bool
+	for _, taint := range newNode.Spec.Taints {
+		if taint.Key == "node-role.kubernetes.io/etcd" && taint.Value == "true" && taint.Effect == corev1.TaintEffectNoExecute {
+			witnessTaintPersist = true
+			break
+		}
+	}
```
brandboat: This is duplicated with lines 144–148. Could you please extract the shared logic into a new method?

brandboat (commented):
Thanks for the PR, left some comments below

pkg/webhook/resources/node/validator.go
dharmit: Thanks for pointing this out. I assumed it would have `true` value, but you're right about it being an empty string. Fixed, PTAL.

pkg/webhook/resources/node/validator.go
dharmit: Ack, done, PTAL.

brandboat (approved):
LGTM!
gentle ping @Vicente-Cheng, could you take a final look? Thanks.

Vicente-Cheng:
Thanks @brandboat, 
I will check tomorrow!

Vicente-Cheng (approved):
thanks, please squash the commits

dharmit:
> thanks, please squash the commits

Thanks for reviewing. Squashed.

dharmit:
All checks passed. Squashed the commits. Received two LGTMs. Merging.

Issue #8148, opened 2025-04-28
[FEATURE] Longhorn v1 data engine - volume encryption GA


innobead:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

The volume encryption, along with VM image encryption, has been implemented in #3129. To achieve GA, at least need to achieve the following tasks. 

- [ ] Feature completeness: review feature completeness and fix the missing parts
- [ ] Validation readness: review test coverage in QA
- [ ] Consistent testing via auto: implement e2e test cases to ensure consistent testing

Yu-Jack:
Issue Tracker

Related Issues:

- https://github.com/harvester/harvester/issues/7248 and upstream issue https://github.com/longhorn/longhorn/issues/11262

TODO:

- Support Terraform https://github.com/harvester/terraform-provider-harvester/pull/125

Yu-Jack:
Example Integration Test: https://github.com/harvester/tests/blob/main/harvester_e2e_tests/integrations/test_2_security_images.py

Yu-Jack:
We can skip the #7248 issue. It should rarely happen because the use case is unusual.

hoo29:
To add our use case, we use terraform to create the images in harvester for multiple (3+) different operating systems and would likely hit this. Why is it unusual?

Yu-Jack:
Hi @hoo29  May I know more about your case? Do you mean you'll create three identical encrypted images at the same time?

hoo29:
Sorry I was going from the title of the linked bug. We are creating multiple images at the same time (or nearly) with terraform but they will be from different images. If it's for identical images we won't do that, sorry for the noise.

Yu-Jack:
Support provisioning a guest cluster with an encrypted image.

<img width="1028" height="301" alt="Image" src="https://github.com/user-attachments/assets/24313dd2-2cab-42f4-aa26-d806b3a1f18f" />

---

<img width="1506" height="468" alt="Image" src="https://github.com/user-attachments/assets/2d06fd8b-8195-4714-b9e0-9b982e83c8ab" />

asc5543:
# Ensure [image encryption/decryption test cases](https://github.com/harvester/tests/blob/main/harvester_e2e_tests/integrations/test_2_security_images.py) runs well.
## Test result
Test PASS, the test cases pass.

## Environment
* Harvester ISO version: `v1.6.0-rc2`
[...]

PR #8592, opened 2025-07-03
fix: securityParameters error message

Yu-Jack:
#### Problem:

The error message is wrong while testing terraform. It shouldn't be `url`.

```
╷
│ Error: admission webhook "validator.harvesterhci.io" denied the request: url cannot be modified
│
│   with harvester_image.encrypted_image,
│   on main.tf line 46, in resource "harvester_image" "encrypted_image":
│   46: resource "harvester_image" "encrypted_image" {
│
```

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8148


WebberHuang1118 (approved):
LGTM, thanks.

PR #8599, opened 2025-07-03
minor changes for 3rd party csi support and vm importer

ibrokethecloud:
#### Problem:

Currently with the vm-import-controller enhancements the VM can boot directly off the golden pvc. When a user deletes a VM, the UI allows users to delete attached disks. This can cause issues as the golden pvc is owned by the virtualmachineimage object and that blocks pvc deletion. This can result in a scenario when VM is stuck in deleting phase.

Additional changes in vm-import-controller and vm mutator, result in a situation where when a longhorn v1 imported pvc matches the image name. This results in a situation when the pvc cannot be deleted since it has the golden pvc annotation. This in turn blocks the longhorn based vm image deletion since the backing image storage class is being used for a pvc

#### Solution:

* Minor change in vm api to skip pvc's with golden images from disk deletion request
* Minor change to pvc mutator to skip annotating longhorn backed images as golden pvc when image and pvc name are the same
 
#### Related Issue(s):

https://github.com/harvester/harvester/issues/8504
https://github.com/harvester/harvester/issues/8634

### DIFF
--- a/pkg/api/vm/store.go
+++ b/pkg/api/vm/store.go
@@ -48,8 +48,18 @@ func (s *vmStore) Delete(request *types.APIRequest, _ *types.APISchema, id strin
 		}
 	}
 
+	var filteredPVCs []string
+	for _, v := range removedPVCs {
+		ok, err := s.isPVCGoldenImage(vm.Namespace, v)
+		if err != nil {
+			return types.APIObject{}, apierror.NewAPIError(validation.ServerError, fmt.Sprintf("Failed to lookup pvc %s for vm %s/%s, %v", v, request.Namespace, request.Name, err))
+		}
+		if !ok {
+			filteredPVCs = append(filteredPVCs, v)
+		}
+	}
 	// Set removed PVCs in annotations. VMController is in charge of the cleanup.
-	if err = s.setRemovedPVCs(vm, removedPVCs); err != nil {
+	if err = s.setRemovedPVCs(vm, filteredPVCs); err != nil {
 		return types.APIObject{}, apierror.NewAPIError(validation.ServerError, fmt.Sprintf("Failed to set removedPersistentVolumeClaims to virtualMachine %s/%s, %v", request.Namespace, request.Name, err))
 	}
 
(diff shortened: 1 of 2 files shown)


WebberHuang1118 (approved):
LGTM, thanks. Just a suggestion for comment:

```
Additional changes in vm-import-controller and vm mutator, result in a situation where when a longhorn v1 imported pvc matches the image name. This results in a situation when the pvc cannot be deleted since it has the golden pvc annotation. This in turn blocks the longhorn based vm image deletion since the backing image storage class is being used for a pvc
```
This scenario is a bit complex—I spent some time understanding the entire flow. Perhaps we could add comments to the relevant code sections?

Vicente-Cheng (approved):
lgtm

pkg/api/vm/store.go
```diff
@@ -67,3 +77,22 @@ func (s *vmStore) setRemovedPVCs(vm *kubevirtv1.VirtualMachine, removedPVCs []st
 	_, err := s.vms.Update(vmCopy)
 	return err
 }
+
+// vm-import-controller allows golden images to be used natively as pvc with a vm
+// during a delete operation we need to skip such pvc's as they are managed by the
+// vmimage, and deletion of the vmimage will remove the pvc
```
ihcsim, line 83: Any reasons why we can't just delete the vm image? That will automatically clean up the associated data volume and PVC.

ibrokethecloud: The users can delete the vm image explicitly. Ideally we should not be cleaning it up if VM is being deleted.

ihcsim: Once the VM is removed, does the vm image has any other uses? Can it be re-used by other imported VMs?

ibrokethecloud: yes you could create another vm using the imported vm image. Based on some annotation logic if the vm is imported you could use the golden pvc directly or if this is a normal vm then we will create another pvc from the golden pvc, as is a normal vm creation workflow

ihcsim (approved):

ibrokethecloud:
@mergify backport v1.6

Issue #8625, opened 2025-07-09
[BUG] Volume online resizing is blocked when configured from the VM page


WebberHuang1118:
### Describe the Bug

Volume online resizing is blocked when configured from the VM page

### To Reproduce

- Create a VM and start it.
- Change the volume size from the VM details page.
- Volume resizing is blocked by the webhook.

<img width="1902" height="961" alt="Image" src="https://github.com/user-attachments/assets/cba439cd-4be6-425b-afd0-88e6ff78c2c0" />

### Expected Behavior

The volume should be able to be resized.

### Support Bundle for Troubleshooting

none

### Environment

- Harvester version: master branch (commit `e85a0f142cc1b8ec512ffeb95559bd777538d29c`)
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Workaround and Mitigation

Navigate to the Volume page and expand the PVC.

WebberHuang1118:
After applying the commit to allow volume online expansion from the VM page:  https://github.com/harvester/harvester/pull/8629/commits/2dc5dfce633f663aa5288f0c84357b30abc705ff

There’s an important detail to note: 
- When resizing is configured from the VM page, the Harvester GUI actually updates the VM annotation. The VM controller then updates the size of all related PVCs accordingly:  [vm_controller.go#L142-L148](https://github.com/harvester/harvester/blob/e85a0f142cc1b8ec512ffeb95559bd777538d29c/pkg/controller/master/virtualmachine/vm_controller.go#L142-L148)

This leads to two notable situations:
1. Misleading GUI Message When **Online Expand Criteria Are Met** [Relevant enhancement doc](https://github.com/harvester/harvester/blob/master/enhancements/20250612-volume-online-expansion.md#pvc-webhook-validator)  (for example, `csi-online-expand-validation` is enabled)
    - However, the GUI still displays the message: `Restart the virtual machine for configuration to take effect.`
    - This message is shown for any VM manifest change, even though a VM restart is not required for the volume expansion to take effect.
    - Recording: https://github.com/user-attachments/assets/c11ad312-e79b-42e8-a932-622aee1bc7fc
    
2. Silent Failure and Retry Storm When **Online Expand Criteria Are Not Met** (for example, `csi-online-expand-validation` is disabled)
   - However, the GUI does **not** surface the webhook error to the user.
[...]

PR #8629, opened 2025-07-10
fix: allow volume online resizing when configured from the VM page

WebberHuang1118:
#### Problem:
Volume online resizing is blocked when configured from the VM page

#### Solution:
allow volume online resizing when configured from the VM page

#### Related Issue(s):
#8625 

#### Test plan:
- Update the Harvester webhook deployment with this PR.
- Create a VM and start it.
- Change the volume size from the VM details page.
- Verify that the volume resizing is successfully applied.


Vicente-Cheng (dismissed):
lgtm, thanks for the enhancement

FrankYang0529 (dismissed):
LGTM. Thanks.

FrankYang0529 (approved):

Issue #8639, opened 2025-07-10
[BUG] crash in upgrade version syncer


m-ildefons:
### Describe the Bug

Harvester Pod crashes due to out of bound bug in version syncer

### To Reproduce

- Install Harvester
- Make sure there are at least two versions that the Harvester cluster can be upgraded to:
```
> k get versions --all-namespaces
NAMESPACE          NAME     ISO-URL                                                                    RELEASEDATE            MINUPGRADABLEVERSION
harvester-system   v1.4.3   https://releases.rancher.com/harvester/v1.4.3/harvester-v1.4.3-amd64.iso   2025-05-26T10:00:00Z   v1.4.1
harvester-system   v1.5.1   https://releases.rancher.com/harvester/v1.5.1/harvester-v1.5.1-amd64.iso   2025-06-30T10:00:00Z   v1.4.2
```

### Expected Behavior

Harvester Pod should not crash.
RKE2 guest cluster should reliably become ready

### Support Bundle for Troubleshooting

Harvester pod logs:

```
Container: apiserver
Disconnected
W0710 14:10:52.554270       7 warnings.go:70] kubevirt.io/v1 VirtualMachineInstancePresets is now deprecated and will be removed in v2.
2025-07-10T14:15:54.555605625Z W0710 14:15:54.555491       7 warnings.go:70] kubevirt.io/v1 VirtualMachineInstancePresets is now deprecated and will be removed in v2.
2025-07-10T14:20:57.557650868Z W0710 14:20:57.557461       7 warnings.go:70] kubevirt.io/v1 VirtualMachineInstancePresets is now deprecated and will be removed in v2.
panic: runtime error: slice bounds out of range [2:1]

goroutine 330 [running]:
github.com/harvester/harvester/pkg/controller/master/upgrade.(*versionSyncer).cleanupVersions(0xc007918ae0, {0xc000058199, 0x6}, {0xc0097e5408, 0xe, 0x18?})
	/go/src/github.com/harvester/harvester/pkg/controller/master/upgrade/version_syncer.go:211 +0x446
github.com/harvester/harvester/pkg/controller/master/upgrade.(*versionSyncer).syncVersions(0xc007918ae0, {{0xc0097e5408?, 0xe?, 0x3183caf?}}, {0xc000058199, 0x6})
	/go/src/github.com/harvester/harvester/pkg/controller/master/upgrade/version_syncer.go:158 +0x65
[...]

PR #8640, opened 2025-07-10
fix version GC logic

m-ildefons:
The version GC logic used a mechanism to remove an elemet from a slice that didn't do bounnds checking. As a result, on edge cases the mechanism might cause a crash - e.g. when there are two possible versions that can be targeted for an upgrade.

This change fixes that by utilizing a mechanism that doesn't rely on indices into slices.

related-to: harvester/harvester#8639

#### Problem:
This code removes an element from a slice without changing the order of the remaining elements in the slice. However it only works if `i` is a valid index in the slice. It does not do bounds checking or validate that `i` is indeed a valid index.

```
append(versions.Items[:i], versions.Items[i+1:]...)
```

To quote from https://stackoverflow.com/questions/37334119/how-to-delete-an-element-from-a-slice-in-golang:

> This answer does not perform bounds-checking. It expects a valid index as input. This means that negative values or indices that are greater or equal to the initial len(s) will cause Go to panic.

Example:
Assuming `versions.Items` contains 2 elements and we're in the second iteration of the loop, i.e. `i` equals `1`.
Then `versions.Items[:i]` will be equal to `versions.Items[0:1]` and contain the entire slice. While `versions.Items[i+1:]` will be equal to `versions.Items[2:1]`, which is outside of the bounds of the slice.

#### Solution:

Use an algorithm that doesn't rely on indices.
By copying each element that shall remain into a new slice and then exchanging the slices, index based access can be avoided.

#### Related Issue(s):

harvester/harvester#8639

#### Test plan:

The Harvester pod shall not crash, regardless of the number of Versions in the cluster.

### DIFF
--- a/pkg/controller/master/upgrade/version_syncer.go
+++ b/pkg/controller/master/upgrade/version_syncer.go
@@ -189,22 +189,27 @@ func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion s
 
 // cleanupVersions remove version resources that's can't be upgraded to anymore
 func (s *versionSyncer) cleanupVersions(currentVersion string, remoteVersions []Version) error {
-	versions, err := s.versionClient.List(s.namespace, metav1.ListOptions{})
+	versionList, err := s.versionClient.List(s.namespace, metav1.ListOptions{})
 	if err != nil {
 		return err
 	}
 
-	for _, version := range remoteVersions {
-		for i, v := range versions.Items {
-			versionObj := v
-			if !canUpgrade(currentVersion, &versionObj, version) {
-				if err := s.versionClient.Delete(v.Namespace, v.Name, &metav1.DeleteOptions{}); err != nil {
+	remainingVersions := versionList.Items
+
+	for _, remote := range remoteVersions {
+		versions := remainingVersions
+		remainingVersions = []harvesterv1.Version{}
+
+		for _, version := range versions {
+			if !canUpgrade(currentVersion, &version, remote) {
+				if err := s.versionClient.Delete(version.Namespace, version.Name, &metav1.DeleteOptions{}); err != nil {
 					if apierrors.IsNotFound(err) {
 						continue // likely object has been removed so we can continue
 					}
 					return err
 				}
-				versions.Items = append(versions.Items[:i], versions.Items[i+1:]...)
+			} else {
+				remainingVersions = append(remainingVersions, version)
 			}
 		}
 	}


ibrokethecloud (approved):

starbops (approved):
LGTM, thank you.

pkg/controller/master/upgrade/version_syncer.go
```diff
@@ -189,22 +189,27 @@ func (s *versionSyncer) syncVersions(resp CheckUpgradeResponse, currentVersion s
...
 	if err != nil {
 		return err
 	}
 
-	for _, version := range remoteVersions {
-		for i, v := range versions.Items {
-			versionObj := v
-			if !canUpgrade(currentVersion, &versionObj, version) {
-				if err := s.versionClient.Delete(v.Namespace, v.Name, &metav1.DeleteOptions{}); err != nil {
+	remainingVersions := versionList.Items
```
w13915984028, line 197: technically, this function could also be optimized as following

```
func (s *versionSyncer) cleanupVersions(currentVersion string, remoteVersions []Version) error {
	versionList, err := s.versionClient.List(s.namespace, metav1.ListOptions{})
	if err != nil {
		return err
	}

	if len(versionList.Items) == 0 {
		return nil
	}

	versionListRemovedFlag := make([]bool, len(versionList.Items)) // initial as false

	for _, remote := range remoteVersions {
		for i, _ := range versionList {
			if versionListRemovedFlag[i] == true {
				continue
			}
			if !canUpgrade(currentVersion, &version, remote) {
				if err := s.versionClient.Delete(versionList.Items[i].Namespace, versionList.Items[i].Name, &metav1.DeleteOptions{}); err != nil {
						versionListRemovedFlag[i] = true
						continue // likely object has been removed so we can continue
					}
					return err
				}
				versionListRemovedFlag[i] = true
			}
		}
	}

	return nil
}	
```

m-ildefons: You're right. A slice of booleans would save a few bytes of memory compared to copying entire objects around.

And I'll gladly admit that the version I wrote isn't optimized at all. Not for speed and not for space efficiency either.

But if we really want to optimize with a noticeable effect, then we'd need to change this function such that it minimizes API calls. In particular, instead of deleting each `Version` object individually, the function should rather pre-compute a list of `Version` objects to delete and then use a single deletecollection API call. This saves on network round-trips, HTTP request parsing, error checks and most importantly this saves on transactions for the etcd store.

PR #6129, opened 2024-07-05
feat: change status from optional to required

FrankYang0529:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/6126

**Test plan:**

#### New cluster

1. Check backup and restore can work.
2. Check when restoring in progress, delete related vmbackup. In this case, webhook should return error.

#### Upgrade path

1. Create a v1.5.0 cluster.
2. Create VM and backup and restored VM.
3. Upgrade to this branch.
4. Check backup still can work.

PR #8668, opened 2025-07-15
Revert "feat: change status from optional to required"

FrankYang0529:
Reverts harvester/harvester#6129

Since the original PR change some VMBackup logic path and we cannot create VMBackup via the API (but not kubectl), it's safer to revert the change and do this enhancement in the next version.


FrankYang0529:
Close the PR and use https://github.com/harvester/harvester/pull/8671.

Issue #6126, opened 2024-07-05
[REFACTOR] Change VirtualMachineBackupStatus from optional to required


FrankYang0529:
**Is your enhancement related to a problem? Please describe.**
In most of Harvester CRD, we don't use pointer for status field.

Without pointer, we can assign condition like this:

https://github.com/harvester/harvester/blob/3274578739026db4699b5018d695dca77e2d5bb7/pkg/controller/master/image/vm_image_controller.go#L82-L83

However, we use pointer for VirtualMachineBackup status. If we assign condition directly, it shows pointer error. That is why we have lot of utility functions for it:

https://github.com/harvester/harvester/blob/3274578739026db4699b5018d695dca77e2d5bb7/pkg/controller/master/backup/util.go#L70-L113

**Describe the solution you'd like**
Change VirtualMachineBackupStatus from optional to required. Make sure upgrade path can work.

innobead:
@FrankYang0529 please move this to the right status, since this is ready.

FrankYang0529:
@innobead this is not ready yet. I need to create another PR to refactor this. The refactor should consider how to handle `status != nil` check. I will work on it.

asc5543:
# New cluster
## Test result
Test Fail, the deletion isn't blocked 

## Environment
* Harvester ISO version: `v1.7.0-rc3`
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): `3 node baremetal`

## Test Scenario
### Preparation
Configure the backup target

1. <details> <summary> Go to Virtual Machine Backups, click setting </summary>

    <img width="1896" height="1057" alt="Image" src="https://github.com/user-attachments/assets/aa8ae024-44a9-4ef7-b3c9-067644eec3da" />
    
   </details>

1. <details> <summary> Configure the back-up target </summary>

    <img width="1896" height="1057" alt="Image" src="https://github.com/user-attachments/assets/f42d8d32-3b70-4c1c-a90c-7a85fcf674fe" />

   </details>

### Test Steps
1. <details>
   <summary> ✔ Check backup and restore can work.
   </summary>

    * <details><summary> ✔ Create file on original vm </summary>

        ```
        ubuntu@first-vm:~$ cat file.txt 
        file
        ubuntu@first-vm:~$ sha512sum file.txt 
[...]

PR #8671, opened 2025-07-15
Revert "feat: change status from optional to required"

FrankYang0529:
#### Problem:
Reverts https://github.com/harvester/harvester/pull/6129

Since the original PR change some VMBackup logic path and we cannot create VMBackup via the API (but not kubectl), it's safer to revert the change and do this enhancement in the next version.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/6126


Vicente-Cheng (approved):
thanks!

PR #8679, opened 2025-07-16
Expose cdi settings for day 2 operations follow-up

brandboat:
this is a review follow-up for https://github.com/harvester/harvester/pull/8579
original issue:
https://github.com/harvester/harvester/issues/8077

### DIFF
--- a/pkg/webhook/resources/storageclass/mutator.go
+++ b/pkg/webhook/resources/storageclass/mutator.go
@@ -15,6 +15,8 @@ import (
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
+const emptyAnnotationsPatch = `{"op": "add", "path": "/metadata/annotations", "value": {}}`
+
 var patchAnnotation = `{"op": "add", "path": "/metadata/annotations/%s", "value": %s}`
 
 func NewMutator() types.Mutator {
@@ -44,15 +46,15 @@ func (m *storageClassMutator) Resource() types.Resource {
 }
 
 func (m *storageClassMutator) Create(_ *types.Request, newObj runtime.Object) (types.PatchOps, error) {
-	return generatePatchOps(newObj.(*storagev1.StorageClass)), nil
+	return generateCDIAnnoPatchOps(newObj.(*storagev1.StorageClass)), nil
 }
 
 func (m *storageClassMutator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) (types.PatchOps, error) {
-	return generatePatchOps(newObj.(*storagev1.StorageClass)), nil
+	return generateCDIAnnoPatchOps(newObj.(*storagev1.StorageClass)), nil
 }
 
-// generatePatchOps generates patch operations for the storage class
-func generatePatchOps(sc *storagev1.StorageClass) types.PatchOps {
+// generateCDIAnnoPatchOps generates cdi annotations patch for the storage class
+func generateCDIAnnoPatchOps(sc *storagev1.StorageClass) types.PatchOps {
 	var patchOps types.PatchOps
 
 	switch sc.Provisioner {
@@ -60,6 +62,11 @@ func generatePatchOps(sc *storagev1.StorageClass) types.PatchOps {
 		patchOps = append(patchOps, generateLonghornPatchOps(sc)...)
 	case util.CSIProvisionerLVM:
 		patchOps = append(patchOps, generateLVMPatchOps(sc)...)
+	}
+
+	if len(patchOps) > 0 && sc.Annotations == nil {
+		// append empty annotations map as the first operation
+		patchOps = append([]string{emptyAnnotationsPatch}, patchOps...)
 	}
 
 	return patchOps
@@ -77,7 +84,7 @@ func generateLonghornPatchOps(sc *storagev1.StorageClass) types.PatchOps {
 		patchOps = append(patchOps, fmt.Sprintf(
 			patchAnnotation,
 			patch.EscapeJSONPointer(util.AnnotationStorageProfileCloneStrategy),
-			cdiv1.CloneStrategyHostAssisted,
+			strconv.Quote(string(cdiv1.CloneStrategyHostAssisted)),
 		))
 	}
 
@@ -86,7 +93,7 @@ func generateLonghornPatchOps(sc *storagev1.StorageClass) types.PatchOps {
 		patchOps = append(patchOps, fmt.Sprintf(
 			patchAnnotation,
 			patch.EscapeJSONPointer(util.AnnotationStorageProfileSnapshotClass),
-			"longhorn-snapshot",
+			strconv.Quote("longhorn-snapshot"),
 		))
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/webhook/resources/storageclass/mutator.go
```diff
@@ -117,14 +131,54 @@ func generateLVMPatchOps(sc *storagev1.StorageClass) types.PatchOps {
...
+	}
+
+	return patchOps, nil
+}
+
+func (m *storageClassMutator) getSnapshotClass(provisioner string, defaultClass string) (string, error) {
+	csiDriverConfig, err := util.LoadCSIDriverConfig(m.settingCache)
+	if err != nil {
+		return defaultClass, fmt.Errorf("failed to load CSI driver config: %w", err)
+	}
```
ihcsim: Sorry, I forgot about our previous HEP discussion where we agreed that the snapshot class in CDI and the one in `csi-driver-config` are different :man_facepalming:. 

I am thinking about the error handling around missing `longhorn-snapshot` or `lvm-snapshot` snapshot classes. After a closer look, I think the previous implementation in #8579 is fine, where we used the hardcoded default snapshot classes. If the LHv2 or LVM snapshot class is missing, CDI should handle it during image creation. I assume the errors would be visible to the user. Then to recover from it, the user should just re-create the missing snapshot class.

Unless we want to provide a user with a way to override the default snapshot class, either through annotations or by looking up the `csi-driver-config` setting.

WDYT?

brandboat: > Sorry, I forgot about our previous HEP discussion where we agreed that the snapshot class in CDI and the one in csi-driver-config are different 🤦‍♂️.

Forgot that too, 🤣 

> I am thinking about the error handling around missing longhorn-snapshot or lvm-snapshot snapshot classes. After a closer look, I think the previous implementation in https://github.com/harvester/harvester/pull/8579 is fine

👍 

> If the LHv2 or LVM snapshot class is missing, CDI should handle it during image creation. I assume the errors would be visible to the user. Then to recover from it, the user should just re-create the missing snapshot class.

AFAIK, if a user provide the wrong snapshot class, CDI will fallback to host-assisted cloning (the slower clone), I believer there are events/log recorded somewhere, but I can't remember the exact code location.

> Unless we want to provide a user with a way to override the default snapshot class, either through annotations or by looking up the csi-driver-config setting.

I'm not quite sure if we need this override behavior right now, but if it turns out we do, we can always add it later. c.c. @Vicente-Cheng, @WebberHuang1118

brandboat:
Turn this into WIP as I need to fix other issues.

pkg/webhook/resources/storageclass/mutator.go
ihcsim: SGTM. Everything else LGTM.

Re: https://github.com/harvester/harvester/pull/8579#discussion_r2209118325, we can consider performing the CDI storage capabilities check earlier, outside of the `validateVolumeModeAccessModes()` method.

E.g.,

```diff
diff --git a/pkg/webhook/resources/storageclass/validator.go b/pkg/webhook/resources/storageclass/validator.go
index 69a6f8ae1..49077843d 100644
--- a/pkg/webhook/resources/storageclass/validator.go
+++ b/pkg/webhook/resources/storageclass/validator.go
@@ -355,6 +355,16 @@ func (v *storageClassValidator) validateReservedStorageClass(sc *storagev1.Stora
 func (v *storageClassValidator) validateCDIAnnotations(newObj runtime.Object) error {
 	sc := newObj.(*storagev1.StorageClass)
 
+	if !hasCDIAnnotations(sc) {
+		// If the provisioner is not in the CDI capabilities map, the volume access modes annotation is required
+		if _, hasCDICapababilities := cdicaps.CapabilitiesByProvisionerKey[sc.Provisioner]; !hasCDICapababilities {
+			return werror.NewInvalidError(
+				fmt.Sprintf("missing annotation %s. volume access modes are required for CDI integration to work with storage class provisioner %s.",
+					util.AnnotationStorageProfileVolumeModeAccessModes, sc.Provisioner), "")
+		}
+		return nil
+	}
+
 	if err := v.validateLonghornV1(sc); err != nil {
 		return err
 	}
@@ -376,10 +386,6 @@ func (v *storageClassValidator) validateCDIAnnotations(newObj runtime.Object) er
 }
 
 func (v *storageClassValidator) validateLonghornV1(sc *storagev1.StorageClass) error {
-	if !hasCDIAnnotations(sc) {
-		return nil
-	}
-
 	if sc.Provisioner != util.CSIProvisionerLonghorn {
 		return nil
 	}
@@ -413,10 +419,6 @@ func hasCDIAnnotations(sc *storagev1.StorageClass) bool {
 }
 
 func (v *storageClassValidator) validateFilesystemOverhead(sc *storagev1.StorageClass) error {
-	if sc.Annotations == nil {
-		return nil
-	}
-
 	value, exists := sc.Annotations[util.AnnotationCDIFSOverhead]
 	if !exists {
 		return nil
@@ -432,10 +434,6 @@ func (v *storageClassValidator) validateFilesystemOverhead(sc *storagev1.Storage
 }
 
 func (v *storageClassValidator) validateCloneStrategy(sc *storagev1.StorageClass) error {
-	if sc.Annotations == nil {
-		return nil
-	}
-
 	value, exists := sc.Annotations[util.AnnotationStorageProfileCloneStrategy]
 	if !exists {
 		return nil
@@ -454,10 +452,6 @@ func (v *storageClassValidator) validateCloneStrategy(sc *storagev1.StorageClass
 }
 
 func (v *storageClassValidator) validateSnapshotClass(sc *storagev1.StorageClass) error {
-	if sc.Annotations == nil {
-		return nil
-	}
-
 	snapshotClass, snapshotClassExists := sc.Annotations[util.AnnotationStorageProfileSnapshotClass]
 	cloneStrategy, cloneStrategyExists := sc.Annotations[util.AnnotationStorageProfileCloneStrategy]
 
@@ -492,22 +486,11 @@ func (v *storageClassValidator) validateSnapshotClass(sc *storagev1.StorageClass
 }
 
 func (v *storageClassValidator) validateVolumeModeAccessModes(sc *storagev1.StorageClass) error {
-	if sc.Annotations == nil {
-		return nil
-	}
-
 	value, exists := sc.Annotations[util.AnnotationStorageProfileVolumeModeAccessModes]
 	if exists {
 		return v.validateVolumeModeAccessModesAnnotation(value)
 	}
 
-	// If the provisioner is not in the CDI capabilities map, the annotation is required
-	if _, ok := cdicaps.CapabilitiesByProvisionerKey[sc.Provisioner]; !ok {
-		return werror.NewInvalidError(
-			fmt.Sprintf("missing annotation %s. volume access modes are required for CDI integration to work with storage class provisioner %s.",
-				util.AnnotationStorageProfileVolumeModeAccessModes, sc.Provisioner), "")
-	}
-
 	return nil
 }
```

brandboat: Thanks, looks much better!

ihcsim (approved):

Vicente-Cheng (approved):
thanks!

WebberHuang1118 (commented):
LGTM, thanks! The implementation is very clean and highly readable.

PR #8686, opened 2025-07-17
remove unused function

isimluk:
Unused since 7f82b84056a950e674bcaa5e08e9e8f7a2f1c75a

#### Problem:
There is no problem, I just randomly opened a file within a project and something felt off, so I searched and it turned out this code is not used for a good while.

#### Solution:
Delete all the things!

#### Related Issue(s):
N/A

#### Test plan:
Let's run the CI.


ihcsim:
AFAICT, this isn't used anymore in `org:harvester` nor `org:rancher`. Looks safe to remove.

cc @Vicente-Cheng

ihcsim (approved):

Vicente-Cheng (approved):

Issue #8703, opened 2025-07-21
[ENHANCEMENT] Migrate VMs with KubeVirt underlying way


Yu-Jack:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Currently, we directly patch the nodeSelector in vmi spec.

https://github.com/harvester/harvester/blob/4a3f06d17dce8eef1cef1a45e6812638f6119765/pkg/api/vm/handler.go#L480-L481

However, due to https://github.com/harvester/harvester/issues/8670#issuecomment-3083676872, it's not a official way to migrate VMs.

### Describe the solution you'd like

In v1.6.0 KubeVirt, it starts to support adding nodeSelector in `VirtualMachineInstanceMigrationSpec`.

https://github.com/kubevirt/kubevirt/blob/9a793044b701647b3705cd40f5b423a513f1a83c/staging/src/kubevirt.io/api/core/v1/types.go#L1618-L1635

### Describe alternatives you've considered

Before v1.6.0 KubeVirt, we still need a way to migrate VM to the target node. We track this issue in https://github.com/harvester/harvester/issues/8670.

lanfon72:
Verified this enhancement has been applied.

Test Information
----
* Environment: **Baremetal DL360 4 nodes**
* Harvester Version: **v1.7.0-dev-20250914**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with at least 2 nodes
2. Create an image for VM creation
3. Create a VM `vm1` and set _node scheduling_ to a specific host
4. Trigger VM `vm1` migrate to another host (not available in UI)
5. VM `vm1` should not start migration and error message should displayed.
6. Run VM migrate automation tests

Issue #8670, opened 2025-07-15
[BUG] VM can't actually migrate to the correct node we specify in the list


TachunLin:
### Describe the Bug

With a three nodes `v1.6.0-rc1` Harvester cluster. 

When we tried to migrate a vm and select the target node in the list.

After migration finished, we found that the vm **is not actually migrated to the correct node** we selected.

* In this video, we tried to migrate to `node1` but it actually migrate to `node3`
  - rancher-monitoring `disabled`

  https://github.com/user-attachments/assets/b05226a3-3b90-44a7-8889-2a744d74fe26

* In this video, In the video, we tried to migrate to `node1` but it still migrate to `node3` 
  - rancher-monitoring `enabled`

  https://github.com/user-attachments/assets/b05226a3-3b90-44a7-8889-2a744d74fe26

On another 3 nodes kvm environment also encounter this issue

* We tried to migrate to `harvester-node-0` but actually migrate to `harvester-node-1`

  https://github.com/user-attachments/assets/9f3117b7-5dad-41ae-9e7e-8a4be5ab1171

### To Reproduce

1. Prepare a three nodes Harvester cluster
2. Import Harvester to Rancher 
3. Use stand alone or integrated mode to access Harvester dashboard
4. Create a vm (e.g `vm1`)
5. Migrate `vm1` and select a target node
6. Check vm can actually migrate to the node we specified.

### Expected Behavior

VM should always been migrated to the correct node we select from the list. 

### Support Bundle for Troubleshooting

* Disable rancher-monitoring: [supportbundle_78b9afdf-52b8-4905-8f08-3ddb97924277_2025-07-14T06-17-59Z.zip](https://github.com/user-attachments/files/21212144/supportbundle_78b9afdf-52b8-4905-8f08-3ddb97924277_2025-07-14T06-17-59Z.zip)

* Enable rancher-monitoring: [supportbundle_78b9afdf-52b8-4905-8f08-3ddb97924277_2025-07-14T06-29-13Z.zip](https://github.com/user-attachments/files/21212167/supportbundle_78b9afdf-52b8-4905-8f08-3ddb97924277_2025-07-14T06-29-13Z.zip)

### Environment

- Harvester version: v1.6.0-rc1
- Impacted VM: vm1
- Underlying Infrastructure: three nodes bare-metal and kvm machines
- Rancher version: v2.12.0-alpha17

### Additional context

None

### Workaround and Mitigation

None

w13915984028:
[...]

PR #8707, opened 2025-07-21
fix: update vm node select instead of vmi

Yu-Jack:
#### Problem:
Check this https://github.com/harvester/harvester/issues/8670 for more details.

#### Solution:
Since the KubeVirt checks the nodeSelect between vm and vmi spec, we can just update the vm spec instead of vmi. 

However, this fix is temporary. We will use more formal way to migrate the VM in https://github.com/harvester/harvester/issues/8703.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8670

#### Test plan:

Prepare a three nodes cluster to migrate the VM.

Please check this one.

https://github.com/user-attachments/assets/50dc6a35-fbd5-444f-a913-b5f69454f4d9

---

This is the old one. Please just ignore this. I keep it because other people had some comments about the old one.

https://github.com/user-attachments/assets/2c45d022-596d-4d5c-bf3b-5963f9839a91

Upgrades use eviction concept to migrate the VM, so this shouldn't affect the upgrade. Thanks @starbops.

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -478,9 +478,24 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
 			toUpdateVmi.Spec.NodeSelector = make(map[string]string)
 		}
 		toUpdateVmi.Annotations[util.AnnotationMigrationTarget] = nodeName
-		toUpdateVmi.Spec.NodeSelector[corev1.LabelHostname] = nodeName
 
 		if err := util.VirtClientUpdateVmi(ctx, h.virtRestClient, h.namespace, namespace, vmName, toUpdateVmi); err != nil {
+			logrus.Errorf("failed to update vmi %s/%s with migration target %s: %v", namespace, vmName, nodeName, err)
+			return err
+		}
+
+		vm, err := h.vmCache.Get(namespace, vmName)
+		if err != nil {
+			return fmt.Errorf("failed to get virtual machine %s/%s: %v", namespace, vmName, err)
+		}
+		toUpdateVM := vm.DeepCopy()
+		if toUpdateVM.Spec.Template.Spec.NodeSelector == nil {
+			toUpdateVM.Spec.Template.Spec.NodeSelector = make(map[string]string)
+		}
+
+		toUpdateVM.Spec.Template.Spec.NodeSelector[corev1.LabelHostname] = nodeName
+		if _, err := h.vms.Update(toUpdateVM); err != nil {
+			logrus.Errorf("failed to update VM %s/%s with migration target %s: %v", namespace, vmName, nodeName, err)
 			return err
 		}
 	}
--- a/pkg/controller/master/migration/vmi_controller.go
+++ b/pkg/controller/master/migration/vmi_controller.go
@@ -5,7 +5,6 @@ import (
 
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
-	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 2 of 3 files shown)


w13915984028:
@Yu-Jack 

I guess the original implementation has such consideration:

When patching vm's spec, it might cause the vm has a info like `the vm generaiton is x and the observed generation is y`, in your video, right after selecting the target node, it appered very quicky.

After migration, Harvester will drop the nodeSelector, which brings the above info again. What's the real case on your test?

Yu-Jack:
> What's the real case on your test?

@w13915984028  What do you mean? I'm testing to migrate the VM to the specified node.

w13915984028:
@Yu-Jack After VM is migrated, and the migration target nodeSelector is dropped from vm's spec,

Harvester UI will hint a message `vm generaiton is x and the observed generation is y`

refer: https://github.com/harvester/harvester/issues/6844

Yu-Jack:
> When patching vm's spec, it might cause the vm has a info like the vm generaiton is x and the observed generation is y, in your video, right after selecting the target node, it appered very quicky.

Let my upload the demo again. It seems I forgot to change the way of updating the VM. I used Impersonate (`VirtClientUpdateVmi`) to do it like we did in vmi before.

This is the new demo for this PR.

https://github.com/user-attachments/assets/50dc6a35-fbd5-444f-a913-b5f69454f4d9

Yu-Jack:
> After VM is migrated, and the migration target nodeSelector is dropped from vm's spec,

We can't update the nodeSelector vmi spec, so the only choice is vm spec. However, we still need to inherit behavior we did before. We actually use vmi to check something about migration in the code base. 

So, I think the best choice here is to only update and drop the nodeSelector in the vm spec.

brandboat (approved):
Thanks for the clever workaround. We should remove this hack once we've upgraded to KubeVirt v1.6.

Yu-Jack:
> We should remove this hack once we've upgraded to KubeVirt v1.6.

Yeah, we must do it. Otherwise, it might be awful in the future.

w13915984028:
@Yu-Jack From my local test, it seems following spec is patched to vm for a very short while, and then seems kubevirt drops it automatically, it is not done by Harvester vmim_controller.

Did you observe this?  a difference is that my vm is attached to pod network, not a bridge network.

```
		toUpdateVM.Spec.Template.Spec.NodeSelector[corev1.LabelHostname] = nodeName

run a shell script:
while true; do
  kubectl get vm vm1 -oyaml | grep hostname -i
  sleep 0.1
done


      hostname: vm1
      hostname: vm1
        kubernetes.io/hostname: harv2  // the migrate patched NodeSelector, consists a short while
      hostname: vm1
      hostname: vm1


```

w13915984028:
Below are observed from my loca test:

(1) After clicking `migrate`, 6844 is sometimes observed for a very short while, as vm's spec is patched, but kubevirt has a controller-revision object to snapshot a vm when it is started https://github.com/harvester/harvester/issues/6844

(2) the patched `NodeSelector` is used for calculating target node, but it is removed very quickly by kubevirt

it also answers the doubt, why after ugprade, removing the `NodeSelector` (by Harvester controller ) from vm's spec does not cause complain like 6844

(3) suspect there could be kind of race between the controller-revision controller and vm controller/others, to be confirmed.

Yu-Jack:
I think it's this one. If the vmi doesn't get the node selector from vm spec yet and this line triggers, it might drop the node selector in vm and vmi.

https://github.com/kubevirt/kubevirt/blob/4ec4e55cd893d0fe8530b72333d0abf7e2a98e73/pkg/virt-controller/watch/vm/vm.go#L742-L766

I think we should patch and remove them at the same time.

Yu-Jack:
To clarify, it's a bit awkward because it's not a official way to implement it. I think the race condition is expected. Just like I mentioned in the original issue. The best way to remove the message is to use the latest approach described in the https://github.com/harvester/harvester/issues/8703.

Yu-Jack:
To clarify, it's a bit awkward because it's not a official way to implement it. I think the race condition is expected. Just like I mentioned in the original issue. The best way to remove the message and sync between vm and vmi spec is to use the latest approach described in the https://github.com/harvester/harvester/issues/8703. I think we just need to ensure that the feature works and the message doesn't show up eventually in this version. WDYT?

w13915984028 (approved):
LGTM, thanks.

It looks to be the best effort we can offer for now.

Vicente-Cheng:
sorry for the late review, but thanks @Yu-Jack.
This is exactly what we can offer at this time.

We should move the native way next version.

PR #8711, opened 2025-07-21
Fix the sync of migrate button on UI

w13915984028:
#### Problem:

Per issue desc on https://github.com/harvester/harvester/issues/7180

#### Solution:

Add better sync between vm, vmi, vmim

#### Related Issue(s):

https://github.com/harvester/harvester/issues/7180

#### Test plan:

1. Create 4 or more VMs on a same node
1.1 Maintain the node
1.2  4 VM migrations are created
1.3 2 of them starts ( by default 2 in parallel)
1.4 check UI, the still migrating vms have UI menu `Abort Migration`

VM shows `Pending` when it is in the waiting list to be migrated.
image

It looks `Pending migration` is better than `Pending`, the latter is hard to connect with `migration`.

image

image

Note: kubevirt allows 2 VMs to be migrated in parallel when it evacuats vms, additional vm is in a new UI state `Pending Migration`

```
kubevirt-evacuation-2l2qn 
```

note the issue desc https://github.com/harvester/harvester/issues/7180

2. Create 4 or more VMs on a same node
2.1 Click `migrate` on each vm
2.2 When a vm is in `migrating`/`pending`, you can click `Abort Migration`
image

When a vm is in migration, below log is alway seen
```
time="2025-07-21T12:16:24Z" level=debug msg="Fetching setting: csi-driver-config"

time="2025-07-21T12:16:24Z" level=info msg="Attempting to fetch setting csi-driver-config from cache"

time="2025-07-21T12:16:24Z" level=debug msg="Setting csi-driver-config has no value, using default: {\"driver.longhorn.io\":{\"volumeSnapshotClassName\":\"longhorn-snapshot\",\"backupVolumeSnapshotClassName\":\"longhorn\"}}"
```

Solution: Lower info `Attempting to fetch setting csi-driver-config from cache` to debug level.

The original planned PR https://github.com/harvester/harvester/pull/7195 is replaced by this one.

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -255,7 +255,7 @@ func isReady(vmi *kubevirtv1.VirtualMachineInstance) bool {
 }
 
 func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
-	if vmi == nil {
+	if vmi == nil || vmi.DeletionTimestamp != nil || vmi.Annotations[util.AnnotationMigrationState] != "" {
 		return false
 	}
 
@@ -267,11 +267,10 @@ func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
 }
 
 func canAbortMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
-	if vmi != nil &&
-		vmi.Annotations[util.AnnotationMigrationState] == migration.StateMigrating {
-		return true
+	if vmi == nil {
+		return false
 	}
-	return false
+	return vmi.Annotations[util.AnnotationMigrationState] == migration.StateMigrating || vmi.Annotations[util.AnnotationMigrationState] == migration.StatePending
 }
 
 func (vf *vmformatter) canDoBackup(vm *kubevirtv1.VirtualMachine, vmi *kubevirtv1.VirtualMachineInstance) bool {
--- a/pkg/controller/master/migration/vmim_controller.go
+++ b/pkg/controller/master/migration/vmim_controller.go
@@ -15,14 +15,17 @@ import (
 )
 
 const (
+	// StateMigrating represents kubevirt MigrationScheduling, MigrationScheduled, MigrationPreparingTarget, MigrationTargetReady, MigrationRunning
 	StateMigrating         = "Migrating"
 	StateAbortingMigration = "Aborting migration"
+
+	// StatePending represents kubevirt MigrationPhaseUnset, MigrationPending
+	StatePending = "Pending"
 )
 
 // The handler adds the AnnotationMigrationUID annotation to the VMI when vmim starts.
 // This is mainly for the period when vmim is created but VMI.status.migrationState is not updated before
 // the target pod is running.
-
 func (h *Handler) OnVmimChanged(_ string, vmim *kubevirtv1.VirtualMachineInstanceMigration) (*kubevirtv1.VirtualMachineInstanceMigration, error) {
 	if vmim == nil {
 		return nil, nil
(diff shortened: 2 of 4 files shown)


w13915984028:
@mergify backport v1.6

w13915984028:
cc @FrankYang0529  This PR does not touch hot-plug.

If hot-plug just creates `vmim` instances as usual, then when `vmim` is existing, the `Abort Migration` is visiable.

Maybe we need some further thinking on the UI menu:  `Migrate`, `Hotplug` and `Abort Migration`, to allow/avoid the `Abort Migration` to abort the `hot-plug` triggered `vmim`. We can use other issue/PR to enhance. thanks.

pkg/api/vm/formatter.go
```diff
@@ -255,7 +255,7 @@ func isReady(vmi *kubevirtv1.VirtualMachineInstance) bool {
 }
 
 func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
-	if vmi == nil {
+	if vmi == nil || vmi.DeletionTimestamp != nil || vmi.Annotations[util.AnnotationMigrationState] != "" {
```
brandboat, line 258: nit: perhaps we can remove these check to ValidateVMMigratable? Regarding `vmi.Annotations[util.AnnotationMigrationState] != ""`, we already have this check in `ValidateVMMigratable`.

pkg/controller/master/migration/vmim_controller.go
```diff
@@ -39,16 +43,24 @@ func (h *Handler) OnVmimChanged(_ string, vmim *kubevirtv1.VirtualMachineInstanc
...
+		err := h.setVmiMigrationUIDAnnotation(vmi, string(vmim.UID), StatePending)
+		if err != nil {
+			return nil, err
+		}
 		return vmim, h.scaleResourceQuota(vmi)
-	} else if vmim.Status.Phase != kubevirtv1.MigrationFailed && abortRequested {
+	} else if phase != kubevirtv1.MigrationFailed && abortRequested {
 		return vmim, h.setVmiMigrationUIDAnnotation(vmi, string(vmim.UID), StateAbortingMigration)
-	} else if vmim.Status.Phase == kubevirtv1.MigrationScheduling {
+	} else if phase == kubevirtv1.MigrationScheduling || phase == kubevirtv1.MigrationScheduled || phase == kubevirtv1.MigrationPreparingTarget || phase == kubevirtv1.MigrationTargetReady || phase == kubevirtv1.MigrationRunning {
```
Yu-Jack, line 61: I think using a function to replace this long statement would be better for reading like this https://github.com/kubevirt/kubevirt/blob/54ebc9c4580e67714789e590b7e6a048344d7452/staging/src/kubevirt.io/api/core/v1/types.go#L810-L820.

Yu-Jack (commented):
A Nit.

pkg/controller/master/migration/vmim_controller.go
w13915984028: The whole function `OnVmimChanged` needs a refactor, with the bump on kubevirt v160 and supporting migration target node directly.

Let's optimize then.

w13915984028:
Code is rebased with master-head latest.

Yu-Jack (approved):
LGTM, thanks.

pkg/controller/master/migration/vmim_controller.go
brandboat, line 61: Correct me if I'm wrong, I noticed that we didn't use `reflect.DeepEqual` to compare the updated vmi with original vmi in `setVmiMigrationUIDAnnotation` or `syncVM`, which may result in updating the VMI multiple times, even if the migration state annotation has already been set to StateMigrating. Could we also fix that?

w13915984028: The tricky looks to be on the beginning of the function, which won't update multi times if UID & state do not change.

```
func (h *Handler) setVmiMigrationUIDAnnotation(vmi *kubevirtv1.VirtualMachineInstance, UID string, state string) error {
	if vmi.Annotations[util.AnnotationMigrationUID] == UID &&
		vmi.Annotations[util.AnnotationMigrationState] == state {
		return nil
	}
```

brandboat (approved):
LGTM, thanks!

pkg/controller/master/migration/vmim_controller.go
brandboat: Thanks for the info!

pkg/api/vm/formatter.go
w13915984028: it tries to return false asap if a migration is onging

I will leave this for v170 migration refactor after https://github.com/harvester/harvester/issues/8719, more changes are needed

PR #8717, opened 2025-07-22
charts: bump Longhorn v1.9.1-rc1

Vicente-Cheng:
#### Problem:
Bump Longhorn v1.9.1-rc1

#### Solution:
Bump Longhorn v1.9.1-rc1

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8718


brandboat (approved):
LGTM!

I did the following tests:
- create VMImage
- Volume > export image, take snapshot
- create a VM
- VM live migration

innobead (changes requested):
https://github.com/longhorn/charts/commit/43dea24e79455e141179355d212c828ecbf1244a

1.9.1 is ready, even though it's not officially released yet. We can use it instead.

WebberHuang1118 (approved):
LGTM. Just one question — don’t we need to update the package reference in `go.mod` as well?

Vicente-Cheng:
> LGTM. Just one question — don’t we need to update the package reference in `go.mod` as well?

There is some conflict between Kubevirt client-go and k8s client-go.
So, we keep the longhorn package on the old version until the newer kubevirt version (maybe kubevirt v1.6.0)

Vicente-Cheng:
After discussion, we will use the Longhorn v1.9.1-rc1 for Harvester v1.6.0-rc3.
Then, use Longhorn v1.9.1 for Harvester v1.6.0-rc4

FrankYang0529 (approved):

innobead (approved):

Issue #8719, opened 2025-07-22
[TASK] Bump KubeVirt to v1.6


Yu-Jack:
### Task request

Since #8703 will use the new filed in KubeVirt v1.6, we need to bump it fisrt.

### Additional context

When bumping, it raises a type mismatch with generated clientset when running tests. There are upstream issues:

1. https://github.com/kubevirt/containerized-data-importer/issues/3878

    `github.com/harvester/harvester/pkg/generated/clientset/versioned/typed/cdi.kubevirt.io/v1beta1/fake
pkg/generated/clientset/versioned/typed/cdi.kubevirt.io/v1beta1/fake/fake_volumeuploadsource.go:44:12: cannot use gentype.ToPointerSlice(list.Items) (value of type []*"kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1".VolumeImportSource) as []*"kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1".VolumeUploadSource value in return statement`

2. https://github.com/kubevirt/kubevirt/issues/15515, so we can't bump following dependencies. Instead, we need to keep k8s.io/code-generator in v0.31.5 at Harvester.

    From
    ```
    k8s.io/code-generator => k8s.io/code-generator v0.31.5
    k8s.io/gengo/v2 => k8s.io/gengo/v2 v2.0.0-20240228010128-51d4e06bde70
    k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20240228011516-70dd3763d340
    ```

    To
    ```
    k8s.io/code-generator => k8s.io/code-generator v0.32.5
    k8s.io/gengo/v2 => k8s.io/gengo/v2 v2.0.0-20250207200755-1244d31929d7
    k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20250318190949-c8a335a9a2ff
    ```

    The reason is [0.31.5 controller gen ](https://github.com/kubernetes/code-generator/blob/v0.32.5/cmd/client-gen/generators/fake/generator_fake_for_type.go) vs. [0.32.5 controller gen](https://github.com/kubernetes/code-generator/blob/v0.32.5/cmd/client-gen/generators/fake/generator_fake_for_type.go)

    `github.com/harvester/harvester/pkg/generated/clientset/versioned/typed/monitoring.coreos.com/v1/fake
[...]

PR #8721, opened 2025-07-22
bump: kubevirt to v1.6 (go.mod)

Yu-Jack:
https://github.com/harvester/harvester/issues/8719

1. Due to the reason in the https://github.com/harvester/harvester/issues/8719#issue-3251235542, we still can't bump the code generator until upstream code is updated. 

2. Also, we still don't bump  kubevirt version in `deploy/charts/harvester/values.yaml` until it's released. I think it will be release in few weeks. So, I only bump the go.mod here. Chart will be in another PR.


ibrokethecloud (dismissed):

FrankYang0529 (dismissed):
LGTM.

Vicente-Cheng:
Generally, LGTM, but could we add the reason why we downgrade some component versions?
No matter whether it's a commit message or a code-level comment, it would be fine.

I know the reason, but we should have the record so that others would know when he took over the bump behavior.

Vicente-Cheng (approved):
thanks for the big effort

FrankYang0529 (approved):

PR #8734, opened 2025-07-23
fix: migrate VM should have multiple nodes

FrankYang0529:
#### Problem:

A cluster with single node can migrate VM, but it cannot find another node to do it.

#### Solution:

Check node count before migrate VM.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/5835, https://github.com/harvester/harvester/issues/8774

#### Test plan:

Create a single node cluster. There should not have migrate VM or cpu / memory hotplug button.

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -8,6 +8,7 @@ import (
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	cdicommon "kubevirt.io/containerized-data-importer/pkg/controller/common"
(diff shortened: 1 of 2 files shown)


pkg/api/vm/formatter.go
```diff
@@ -254,11 +256,21 @@ func isReady(vmi *kubevirtv1.VirtualMachineInstance) bool {
 	return false
 }
 
-func canMigrate(vmi *kubevirtv1.VirtualMachineInstance) bool {
+func canMigrate(nodeCache ctlcorev1.NodeCache, vmi *kubevirtv1.VirtualMachineInstance) bool {
 	if vmi == nil || vmi.DeletionTimestamp != nil || vmi.Annotations[util.AnnotationMigrationState] != "" {
 		return false
 	}
 
+	nodes, err := nodeCache.List(labels.Everything())
```
w13915984028, line 264: There are some cases which cause this PR will not always work as expected:

(1) Even when there are 2 nodes on the cluster, if another node is witness, or put to cordorn/maintenence mode, `migrate` still does not work

(2) A cluster starts with single node, but later a new node is added, the VM's UI does not watch all related resources, `migrate` button does not appear automatically, a force refresh is needed.

(3) In a multi-cluster env, the `canMigrate` might still not able to cover all cases, click `migrate` and then no target it selectable, or failed/blocked by kubevirt validator

On the other hand, write this as a know limitation on the doc might also be acceptable on v160:

Harvester tries to filter out those non-migratable scenarios, but there are still cases where `migrate` button is visiable, but when clicking it, no `migrate target` is selectable on the drop donw list.

FrankYang0529: I know there are many cases we cannot handle now. Please create another issue to track it. This PR only focuses on single node case.

Vicente-Cheng: HI @w13915984028, 
Could we create another ticket to improve the canMigrate checking for more cases?

For now, we just simply disable this capability for only one node.

Vicente-Cheng (approved):
lgtm, please rebase

Issue #7193, opened 2024-12-18
[BUG] Harvester node driver does not propagate machine pool labels


m-ildefons:
# Description

When creating an RKE2 cluster in Rancher, using the Harvester node driver, labels are not propagated to the VirtualMachine objects or the Pods running the VM processes. This potentially causes problems when the user wants to user labels to identify the VMs belonging to a guest cluster or to a node pool of a guest cluster later on.

# Expected Behavior

The user should be able to set labels to be applied to the VirtualMachine objects.
While users can of course manually label the VMs, it can be hard to identify the correct VMs and this action would need to be repeated e.g. when the guest cluster is scaled.

# Related Issues:

- [SURE-9641](https://jira.suse.com/browse/SURE-9641)
- https://github.com/harvester/harvester/issues/8391

ibrokethecloud:
we need to update the builder https://github.com/harvester/harvester/blob/master/pkg/builder/vm.go#L124-L125 to also apply labels to `vm.spec.template.metadata.labels`

The labels in the `virtualmachineinstancetemplate` are passed on to the launcher pod

Vicente-Cheng:
duplicated with https://github.com/harvester/harvester/issues/8391

PR #8735, opened 2025-07-23
builder: method for setting labels on template

m-ildefons:
Add a method for setting the labels on the VMI template of a VM. Unlike labels on the VM object itself, the labels in the VMI template will propagate to the VMI and launcher pod.

related-to: harvester/harvester#7193

#### Problem:

To propagate labels from Rancher (i.e. the Harvester node driver) all the way to the launcher Pod, it's necessary to be able to set the labels on the VMI template spec.

#### Solution:

Add a method to the VMBuilder which allows setting labels for the VMI template spec

#### Related Issue(s):

related-to: harvester/harvester#7193

### DIFF
--- a/pkg/builder/vm.go
+++ b/pkg/builder/vm.go
@@ -131,6 +131,24 @@ func (v *VMBuilder) Labels(labels map[string]string) *VMBuilder {
 	return v
 }
 
+// VirtualMachineInstanceTemplateLabels is similar to the Labels method, but it
+// sets the labels in the template spec. Unlike lables on the VirtualMachine
+// object, labels in the template spec will propagate to the VMI and the
+// launcher pod.
+func (v *VMBuilder) VirtualMachineInstanceTemplateLabels(labels map[string]string) *VMBuilder {
+	vmiTemplate := v.VirtualMachine.Spec.Template
+
+	if vmiTemplate.ObjectMeta.Labels == nil {
+		vmiTemplate.ObjectMeta.Labels = labels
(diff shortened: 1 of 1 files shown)


ibrokethecloud (approved):

pkg/builder/vm.go
```diff
@@ -131,6 +131,24 @@ func (v *VMBuilder) Labels(labels map[string]string) *VMBuilder {
...
+
+	if vmiTemplate.ObjectMeta.Labels == nil {
+		vmiTemplate.ObjectMeta.Labels = labels
+		return v
+	}
+
+	for key, value := range labels {
+		vmiTemplate.ObjectMeta.Labels[key] = value
+	}
+	return v
```
ibrokethecloud, line 149:
```suggestion
v.VirtualMachine.Spec.Template = vmiTemplate
return v
```

m-ildefons: Is this really necessary?

To my understanding. `vmiTemplate` is a reference to the same object which `v.VirtualMachine.Spec.Template` references. It's in that sense just a short name.
Therefore assigning to `vmiTemplate` already changes the object referenced by `v.VirtualMachine.Spec.Template` and this assignment isn't necessary anymore.

ibrokethecloud: you are right, i should have the vmbuilder spec

ibrokethecloud (approved):
lgtm. thanks.

Issue #8728, opened 2025-07-23
[TASK] Bump go v1.24


Vicente-Cheng:
### Task request

Bump go v1.24
Bump k8s at least v1.31

Update repositories to GO 1.24 in `go.mod`/`Dockerfile`/`Dockerfile.dapper`. If a repository builds many images, please remember to update each `Dockerfile`.

| Done | Repository | go version | k8s version |
| --- | --- | --- | --- |
| 🟢 | https://github.com/harvester/cloud-provider-harvester | 1.24.5 | v1.31.5  |
| 🟢 | https://github.com/harvester/docker-machine-driver-harvester | 1.24.2 | v1.31.5 |
| 🟢 | https://github.com/harvester/eventrouter | 1.24.5 | v1.32.6 |
| 🟢 | https://github.com/harvester/go-common | 1.24.5 | N/A | 
| 🟢 | https://github.com/harvester/harvester | 1.24.2 | v1.31.5 |
| 🟢 | https://github.com/harvester/harvester-csi-driver | 1.24.2 | v1.31.6 |
| 🟢 | https://github.com/harvester/networkfs-manager | 1.24.2 | v1.32.6 |
| 🟢 | https://github.com/harvester/csi-driver-lvm | 1.24.5 | v1.32.6 |
| 🟢 | https://github.com/harvester/harvester-installer | 1.24.5 | v1.32.6 |
| 🟢 | https://github.com/harvester/load-balancer-harvester | 1.24.5 | v1.31.5 | 
| 🟢 | https://github.com/harvester/network-controller-harvester | 1.24.3  | v1.31.5 |
| 🟢 | https://github.com/harvester/node-disk-manager | 1.24.2 | 1.32.6 |
| 🟢 | https://github.com/harvester/node-manager | 1.24.0 | v1.32.1 |
| 🟢 | https://github.com/harvester/pcidevices | 1.24.5 | v1.31.1 |
| 🟢 | https://github.com/harvester/seeder | 1.24.5 | v1.32.2 |
| 🟢 | https://github.com/rancher/rancherd | 1.24.5 | *note1* |
| 🟢 | https://github.com/rancher/support-bundle-kit | 1.24.0 | v1.33.1 | 
| 🟢 | https://github.com/harvester/vm-dhcp-controller | 1.24.4 | v1.31.5 |
| 🟢 | https://github.com/harvester/vm-import-controller | 1.24.5 | v1.32.6 |
[...]

PR #8755, opened 2025-07-25
bump harvester network controller dependency

ibrokethecloud:
PR bumps the version of harvester-network-controller in go.mod to ensure that NodeNetwork client is not being generated.

NodeNetwork was deprecated [in](https://github.com/harvester/network-controller-harvester/commit/24bb90b8214d44f5b3d7f4df3d894848b1e274e4)

The PR ensures that this is not use as this causes issues where other projects like pcidevices where harvester and harvester-network-controller packages are being used.
#### Related Issue(s):

https://github.com/harvester/harvester/issues/8728


Vicente-Cheng (dismissed):

tserong (commented):
I just noticed we still also have github.com/harvester/node-manager v0.3.1 -- does that need updating as well?

ibrokethecloud:
i could while we are at it

tserong (approved):

Vicente-Cheng (approved):

ibrokethecloud:
@mergify backport v1.6

PR #8779, opened 2025-07-31
fix: ignore backup not found error in shouldSkipVolumeSnapshotUpdate

FrankYang0529:
#### Problem:
When a VMBackup is changed back to non ready, the backup controller reconcile it. In some cases, there is no LH Backup and the controller returns error which changing the VMBackup condition message (from `Change back to non-ready` to `Not Ready`). In the next reconciliation, the `shouldSkipVolumeSnapshotUpdate` cannot work, because the condition message is changed and it always returns `false`. The controller update `volumeBackup.readyToUse` from `VolumeSnapshot` and the VMBackup will be ready again finally.

#### Solution:
Ignore LH Backup not found error or always check LH Backup in VMBackup controller.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/7692#top

#### Test plan:

1. Create a cluster and setup backup target with non-zero `refreshIntervalInSeconds`.
2. Create a VMBackup.
3. Get into backup target and move LH data to other places. For example:
```
> cd /opt/backupstore/
> mv backupstore/volumes/97/e3/pvc-b1f62a1d-f364-4e5a-9406-7db6106e59ee/ ../../../../
```
4. Wait for refresh interval, we can see the VMBackup is marked as non-ready and `readyToUse` of related volume backup is false. 
5. Get into backup target and move the data back.
6. Wait for refresh interval, we can see the VMBackup is marked as ready again.

### DIFF
--- a/pkg/controller/master/backup/backup.go
+++ b/pkg/controller/master/backup/backup.go
@@ -1042,19 +1042,25 @@ func (h *Handler) configureCSIDriverVolumeSnapshotClassNames(vmBackup *harvester
 // We would like to avoid the controller change the VMBackup to ready again blindly,
 // because related volume snapshot status doesn't really reflect backup state.
 func (h *Handler) shouldSkipVolumeSnapshotUpdate(vmBackup *harvesterv1.VirtualMachineBackup, volumeBackup harvesterv1.VolumeBackup) (bool, error) {
-	isChangedBackupToNonReady := false
+	isChangedBackToNonReady := false
 	for _, condition := range vmBackup.Status.Conditions {
 		if condition.Type == harvesterv1.BackupConditionReady && condition.Message == changeToNonReadyMessage {
-			isChangedBackupToNonReady = true
+			isChangedBackToNonReady = true
 			break
 		}
 	}
 
-	if volumeBackup.LonghornBackupName == nil || !isChangedBackupToNonReady {
+	if volumeBackup.LonghornBackupName == nil || !isChangedBackToNonReady {
 		return false, nil
 	}
 
 	conditionMsg, err := checkLHBackup(h.lhbackupCache, *volumeBackup.LonghornBackupName)
+	if apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/controller/master/backup/backup.go
```diff
@@ -1062,6 +1062,11 @@ func (h *Handler) shouldSkipVolumeSnapshotUpdate(vmBackup *harvesterv1.VirtualMa
 			"volumeBackup":   *volumeBackup.Name,
 			"longhornBackup": *volumeBackup.LonghornBackupName,
 		}).Warn("cannot check longhorn backup")
+		if apierrors.IsNotFound(err) {
+			// Don't return not found error here, because it changes the VMBackup status message and there will not have "Change back to non-ready" message.
+			// In the next reconcile, the shouldSkipVolumeSnapshotUpdate will return false, so the volume backup will get updated from VolumeSnapshot.
+			return true, nil
+		}
```
WebberHuang1118:
```suggestion
	conditionMsg, err := checkLHBackup(h.lhbackupCache, *volumeBackup.LonghornBackupName)
	if apierrors.IsNotFound(err) {
		// Don't return not found error here, because it changes the VMBackup status message and there will not have "Change back to non-ready" message.
		// In the next reconcile, the shouldSkipVolumeSnapshotUpdate will return false, so the volume backup will get updated from VolumeSnapshot.
		return true, nil
	}

	if err != nil {
		logrus.WithError(err).WithFields(logrus.Fields{
			"name":           vmBackup.Name,
			"namespace":      vmBackup.Namespace,
			"volumeBackup":   *volumeBackup.Name,
			"longhornBackup": *volumeBackup.LonghornBackupName,
		}).Warn("cannot check longhorn backup")
		return true, err
	}
```

WebberHuang1118 (dismissed):
LGTM, just a nit, thanks.

WebberHuang1118 (approved):

PR #8795, opened 2025-08-03
charts: bump containerized-data-importer v1.62.0

Vicente-Cheng:
#### Problem:
Bump containerized-data-importer v1.62.0

#### Solution:
Bump containerized-data-importer v1.62.0

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8796


deploy/charts/harvester/Chart.lock
```diff
@@ -34,9 +34,9 @@ dependencies:
   version: 0.1.1
 - name: harvester-node-manager
   repository: https://charts.harvesterhci.io
-  version: 1.6.0-dev.0
+  version: v1.6.0-dev.0
```
WebberHuang1118: Is this change expected? If I understand correctly, it shouldn’t be related to this PR.

Vicente-Cheng: nice catch... updated!

Vicente-Cheng:
I will manually backport this PR. We only need the CDI part for the v1.6 branch

brandboat (approved):

WebberHuang1118 (approved):
LGTM, thanks.

Issue #8778, opened 2025-07-31
[BUG] Can't add volume, delete and backup vm when vm enabled cpu pinning while cpu manager is `disabled`


TachunLin:
### Describe the Bug

Given the cpu manager enabled on node (e.g `n2-hp10`). 

* When we create a new vm in running then edit yaml to enable the cpu pinning feature (`dedicatedCpuPlacement: true`)

  https://github.com/user-attachments/assets/61b5379a-c868-4bbd-b046-345caa8dd2eb
  
* After we disable the cpu manager on the node

* We find we can't do further action on vm like `add volume`, `Delete vm`, `Take backup` and etc... 

  https://github.com/user-attachments/assets/458891a4-44d0-49fb-8248-827b7966b93d

  <img width="1597" height="776" alt="Image" src="https://github.com/user-attachments/assets/102dd728-fab2-441a-be44-eda83620d55a" />
  
  <img width="1598" height="781" alt="Image" src="https://github.com/user-attachments/assets/7764f7cc-ffb8-41d6-83a4-a670b37effbf" />

### To Reproduce

1. Enable CPU manager on one of the node
2. Create a new vm (e.g `vm-yaml-add-pinning`)
3. Check vm started in running state
4. Edit yaml of the vm to enable cpu pinning

    ```
    spec:
      template:
        spec:
          domain:
            cpu:
              dedicatedCpuPlacement: true
    ```

5. Check cpu pinning is enabled on that vm
6. Disable the cpu manager on the node
7. Try to add volume to the vm
8. Try to delete the vm
9. Try to backup vm

### Expected Behavior

Should do basic operation on vm when cpu pinning enabled while cpu manager is disabled.

If this action is not allowed, we can considering prevent user to disable cpu manager when there is still have vm with cpu pinning enabled.

### Support Bundle for Troubleshooting

Suuport bundle size too large to exceed upload size limitation

### Environment

- Harvester version: `v1.6.0-rc3`
- Impacted VM: `vm-yaml-add-pinning`
- Impacted volume (PV): `vm-yaml-add-pinning-disk-0-grlkj`
- Underlying Infrastructure: Three nodes bare-metal machines
- Rancher version: v2.12.0-rc2
- Harvester UI extension: v1.6.0-rc3

### Additional context

Found when verify issue https://github.com/harvester/harvester/issues/7598

### Workaround and Mitigation

Enable CPU Manager again could mitigate this issue.

albinsun:
[...]

PR #8800, opened 2025-08-04
Ensure one node has CPU Manager when there is vm has cpu pinning

brandboat:
#### Problem:
Can't add volume, delete and backup vm when vm enabled cpu pinning while cpu manager is disabled.

The root cause is we currently check for running VMIs with CPU pinning, but we didn't check the corresponding VMs before disabling CPU the Manager. Once the CPU Manager is disabled, if no node has it enabled, the validator is triggered. https://github.com/harvester/harvester/blob/22470dab30aa3981079cb5689e7d63f1cbd668f3/pkg/webhook/resources/virtualmachine/validator.go#L252 This blocks any further operations (e.g. add volume, delete vm...), as at least one node with CPU Manager enabled is required.

#### Solution:

Prevent disabling CPU Manager on the last enabled node when pinned VMs exist

https://github.com/user-attachments/assets/a9a57209-fd7b-435a-b0ab-378a2dbc6952

#### Related Issue(s):

#8778

#### Test plan:
- Enable CPU Manager in one node (e.g. harvester-node-0)
- Create a VM `test`
- Click Edit Config, enable CPU Pinning **WITHOUT** restart the VM.
- Disable CPU Manager in `harvester-node-0`, error pops up: `At least one node must have CPU Manager enabled when a VM is using CPU pinning. Please remove CPU pinning from the following VM(s) to proceed: default/test`.

N/A

### DIFF
--- a/pkg/webhook/indexeres/indexer.go
+++ b/pkg/webhook/indexeres/indexer.go
@@ -56,6 +56,7 @@ func RegisterIndexers(clients *clients.Clients) {
 	vmInformer := clients.KubevirtFactory.Kubevirt().V1().VirtualMachine().Cache()
 	vmInformer.AddIndexer(indexeresutil.VMByPVCIndex, indexeresutil.VMByPVC)
 	vmInformer.AddIndexer(indexeresutil.VMByHotplugPVCIndex, indexeresutil.VMByHotplugPVC)
+	vmInformer.AddIndexer(indexeresutil.VMByCPUPinningIndex, indexeresutil.VMByCPUPinning)
 
 	svmBackupCache := clients.HarvesterFactory.Harvesterhci().V1beta1().ScheduleVMBackup().Cache()
 	svmBackupCache.AddIndexer(ScheduleVMBackupBySourceVM, scheduleVMBackupBySourceVM)
@@ -70,7 +71,6 @@ func RegisterIndexers(clients *clients.Clients) {
 
 	vmCache := clients.KubevirtFactory.Kubevirt().V1().VirtualMachine().Cache()
 	vmCache.AddIndexer(VMByMacAddress, vmByMacaddrs)
-	vmInformer.AddIndexer(indexeresutil.VMByCPUPinningIndex, indexeresutil.VMByCPUPinning)
 }
 
 func vmBackupBySourceUID(obj *harvesterv1.VirtualMachineBackup) ([]string, error) {
--- a/pkg/webhook/indexeres/indexer.go
+++ b/pkg/webhook/indexeres/indexer.go
@@ -70,6 +70,7 @@ func RegisterIndexers(clients *clients.Clients) {
 
 	vmCache := clients.KubevirtFactory.Kubevirt().V1().VirtualMachine().Cache()
 	vmCache.AddIndexer(VMByMacAddress, vmByMacaddrs)
+	vmInformer.AddIndexer(indexeresutil.VMByCPUPinningIndex, indexeresutil.VMByCPUPinning)
 }
 
 func vmBackupBySourceUID(obj *harvesterv1.VirtualMachineBackup) ([]string, error) {
--- a/pkg/webhook/resources/node/validator.go
+++ b/pkg/webhook/resources/node/validator.go
@@ -137,7 +137,7 @@ func (v *nodeValidator) validateCPUManagerOperation(node *corev1.Node) error {
 		return err
 	}
 	// check if there is at least one node has cpu manager enabled if there is any vm with cpu pinning
-	if err := checkCPUManagerEnabledForPinnedVMs(policy, v.nodeCache, v.vmCache); err != nil {
+	if err := checkCPUManagerEnabledForPinnedVMs(node, policy, v.nodeCache, v.vmCache); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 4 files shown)


FrankYang0529 (dismissed):

pkg/webhook/indexeres/indexer.go
```diff
@@ -70,6 +70,7 @@ func RegisterIndexers(clients *clients.Clients) {
 
 	vmCache := clients.KubevirtFactory.Kubevirt().V1().VirtualMachine().Cache()
 	vmCache.AddIndexer(VMByMacAddress, vmByMacaddrs)
+	vmInformer.AddIndexer(indexeresutil.VMByCPUPinningIndex, indexeresutil.VMByCPUPinning)
```
innobead: Move this next to the other `vmInformer.AddIndexer` statements. Group them together.

pkg/webhook/resources/node/validator.go
```diff
@@ -261,3 +268,75 @@ func getCPUManagerRunningJobNamesOnNodes(jobCache ctlbatchv1.JobCache, nodeNames
...
+	if policy == ctlnode.CPUManagerStaticPolicy {
+		return nil
+	}
+
+	cpuManagerNodeCount, err := countNodesWithCPUManager(nodeCache)
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+
+	if cpuManagerNodeCount != 1 {
```
innobead: Should check the `cpuManagerNodeCount >= 1` excluding the updating node?

innobead (changes requested):
After this, need to remove the following limitation from the doc.

[/docs/vm/cpu-pinning.md#L72](https://github.com/harvester/docs/blob/adab876002cf3e5c6ea8e90a3a526eb9d272e412/docs/vm/cpu-pinning.md#L72)
```md
- VMs with CPU pinning enabled must be stopped before CPU Manager is disabled on the corresponding node.

```

pkg/webhook/resources/node/validator.go
brandboat: I want to skip the further checking when this is the last node that has cpu manager enabled and user intends to disable it. However, I think my initial implementation was a bit misleading and unclear, I've refactored the method a bit. Please take another look when you're available, thanks!

brandboat:
> - VMs with CPU pinning enabled must be stopped before CPU Manager is disabled on the corresponding node.

ping @innobead, this is still needed, we have a validator to prevent user from disabling CPU Manager if that node already have VM with CPU Pinning enabled. see https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/node/validator.go#L235-L251

w13915984028:
@brandboat 

I suspect we need to remove/refactor below check, instead of current PR:

https://github.com/harvester/harvester/blob/76d7250f10ff63824abcc107e2d93d1b76da4d83/pkg/webhook/resources/virtualmachine/validator.go#L547

**When webhook check is removed:**
Given a VM is started with CPU pinning enabled and there is not suitable node ( cpu manager disabled or other cases), the VM is finally `unscheduable`;  this can be explained by our newly updated document:
https://docs.harvesterhci.io/v1.5/troubleshooting/vm#unschedulable-virtual-machine

Note: there are many `unschedulable` cases, the webhook check just covers a very basic one.
e.g.:
```
vm1 is pointed to schedule to node1 with cpu pinning enabled while there is only node2 is cpu manager enabled
vm1 is set with cpu pinning, and only node2 is cpu manager enabled, but node2 is put to maintenance
vm1 is set with cpu pinning, and node2 is cpu manager enabled, but node2's CPUs have been used up by other VMs, no room for vm1
...
```

**When webhook check is refactored:**
Precisely detects the change is related to `cpu pinning`:

- when `cpu pinning` is changed to `enabled`, but there is not node suitable for scheduling; block it
- when `cpu pinning` is changed to `disabled`, allow anyway

the change is un-related to `cpu pinning`: pass always

ATM, I would prefer to remove the webhook check, it won't harm too much and can avoid above cases.

pkg/webhook/resources/node/validator.go
```diff
@@ -129,10 +132,14 @@ func (v *nodeValidator) validateCPUManagerOperation(node *corev1.Node) error {
 	if err := checkMasterNodeJobs(node, v.nodeCache, v.jobCache); err != nil {
 		return err
 	}
-	// check if there is any vm that enable cpu pinning while cpu manager is going to be disabled
+	// check if there is any running vmis that enable cpu pinning while cpu manager is going to be disabled
```
ihcsim, line 135: Do we still need to check the VMIs, if we're going to check the VMs?

ihcsim (commented):
I think this change makes sense for enabling/disabling the CPU Manager.

@w13915984028 https://github.com/harvester/harvester/pull/8800#issuecomment-3164167868 also makes sense to me. If kubevirt knows how to handle the error case, we should let kubevirt do it.

brandboat:
> ATM, I would prefer to remove the webhook check, it won't harm too much and can avoid above cases.

@w13915984028 @ihcsim, I haven't been able to come up with a better solution that covers all potential unschedulable cases. As you mentioned, there are too many scenarios to account for. Even with the cpumanager enabled on a node, a VM may still become unschedulable due to resource exhaustion.

I’ll revert [this commit](https://github.com/harvester/harvester/commit/079cee711f0d3e91966cd4e97039920e346a6c4c), unless we can come up with a more effective approach to prevent unschedulable cases. If we do, we can reintroduce this validator.

brandboat:
superseded by https://github.com/harvester/harvester/pull/8853

ihcsim:
@brandboat Thanks for the updates. If we are retaining the node validator check at https://github.com/brandboat/harvester/blob/152b2b578174da4d9d0aca91dacc5867720ef331/pkg/webhook/resources/node/validator.go#L242C5-L258, I think it makes sense to check VMs instead of the VMIs.

brandboat:
> @brandboat Thanks for the updates. If we are retaining the node validator check at https://github.com/brandboat/harvester/blob/152b2b578174da4d9d0aca91dacc5867720ef331/pkg/webhook/resources/node/validator.go#L242C5-L258, I think it makes sense to check VMs instead of the VMIs.

@ihcsim, the `checkCPUPinningVMIs` method is intended to check if there are any running VMs on the node. From my perspective, checking the VMI directly seems simpler since a VMI represents a VM that is either running or paused, but not stopped. Did I misunderstand something? If I missed anything, please let me know. Thanks.

ihcsim:
What are the reasons behind skipping stopped VM? I was thinking the stopped VM might fail to start, if the CPU manager was disabled.

brandboat:
> What are the reasons behind skipping stopped VM?

The original intention of `checkCPUPinningVMIs` is prevent user from disabling CPU Manager on the node that still has CPU Pinned VMs, so that's the reason I skipped stopped VM.

> I was thinking the stopped VM might fail to start, if the CPU manager was disabled.

That's what I've addressed in this PR - This PR addresses the scenario where CPU Manager could be disabled on all nodes while CPU-pinned VMs are still running. 
Before proceeding further, we might need a broader discussion in https://github.com/harvester/harvester/issues/8859. I’m not entirely sure what the best long-term approach is.

ihcsim:
I was hoping it could just be as simple as changing `checkCPUPinningVMIs` to iterate through VMs, instead of VMIs. To clarify, I wasn't talking about a global check like what was did in `checkCPUManagerEnabledForPinnedVMs`.

PR #8821, opened 2025-08-06
ci: use env variable to aviod injection [skip ci]

Yu-Jack:
Use [intermediate environment variable](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable) to store user-defined parameters and avoid using `cat <<` to store string.


innobead (approved):

Vicente-Cheng:
@mergifyio backport v1.5 v1.6

Issue #8689, opened 2025-07-17
[BUG] Upgrade v1.4.2 --> v1.5.1 stuck Post-draining when node disks use multipathing (mpath)


danielbjornadal:
When upgrading from v1.4.2 (only one tested) to v1.5.1, the upgrade process will stop on Post-draining on the first node. 
All nodes use disks from multipathing.

```
t-hvst-1:~ # lsblk
NAME                                   MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINTS
loop0                                    7:0    0     3G  1 loop  /
sda                                      8:0    0 446.6G  0 disk  
├─sda1                                   8:1    0    64M  0 part  
├─sda2                                   8:2    0    50M  0 part  
├─sda3                                   8:3    0     8G  0 part  
├─sda4                                   8:4    0    15G  0 part  
├─sda5                                   8:5    0   300G  0 part  
├─sda6                                   8:6    0 123.5G  0 part  
└─362cea7f0d36b81002d4907d476c34076    254:0    0 446.6G  0 mpath 
  ├─362cea7f0d36b81002d4907d476c34076-part1
  │                                    254:1    0    64M  0 part  
  ├─362cea7f0d36b81002d4907d476c34076-part2
  │                                    254:2    0    50M  0 part  /oem
  ├─362cea7f0d36b81002d4907d476c34076-part3
  │                                    254:3    0     8G  0 part  
  ├─362cea7f0d36b81002d4907d476c34076-part4
  │                                    254:4    0    15G  0 part  /run/initramfs/cos-state
  ├─362cea7f0d36b81002d4907d476c34076-part5
  │                                    254:5    0   300G  0 part  /var/lib/longhorn
  │                                                               /var/crash
  │                                                               /var/lib/third-party
  │                                                               /var/lib/cni
  │                                                               /var/lib/wicked
  │                                                               /var/lib/kubelet
  │                                                               /var/lib/rancher
  │                                                               /var/log
  │                                                               /usr/libexec
  │                                                               /root
  │                                                               /opt
  │                                                               /home
  │                                                               /etc/pki/trust/anchors
  │                                                               /etc/cni
  │                                                               /etc/iscsi
  │                                                               /etc/ssh
  │                                                               /etc/rancher
  │                                                               /etc/systemd
  │                                                               /usr/local
  └─362cea7f0d36b81002d4907d476c34076-part6
                                       254:6    0 123.5G  0 part  /var/lib/harvester/defaultdisk
[...]

PR #8839, opened 2025-08-07
address multipath upgrade failures

ibrokethecloud:
#### Problem:

currently upgrades on hosts booting off external disks and leveraging multipath fail because elemental generates an upgrade spec which leverages the underlying multipath device and attempts to mount the same to perform the upgrade.

This fails and results in upgrade failing during the OS upgrade

#### Solution:

elemental-cli v1.1.7 has a fix to allow it to identify a mapper device during an upgrade and use the same to generate the upgrade spec.

minor change to upgrade path to package latest elemental binary from current builds, to ensure new binary is used during upgrade. In addition if mpath is being used mount COS_STATE into /run/cos/state to ensure upgrade can work smoothly

#### Test plan:

https://github.com/harvester/harvester/issues/8689

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -45,6 +45,30 @@ sparsify_passive_img()
   fi
 }
 
+# elemental scans /dev for blockdevices and checks for their mount points
+# if the disk is already mounted then the current logic remounts them as read/write partition
+# and if a partition is not mounted elemental tries to just mount it as a read/write partition
+# in case of multipath devices, the /dev/mapper device is not detected, and elemental
+# finds the path making the disk as not mounted. 
+# due to change introduced by https://github.com/rancher/elemental-toolkit/pull/2302 elemental reconciles
+# the actual disk partition in use and finds the correct mapper device related to the cos partition
+# however the logic still scans the underlying disk for mount path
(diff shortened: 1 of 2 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -45,6 +45,15 @@ sparsify_passive_img()
   fi
 }
 
+check_and_mount_state(){
+  MAPPER_IN_USE=$(chroot ${HOST_DIR} blkid -L COS_STATE |grep mapper)
+  if [ -n "$MAPPER_IN_USE" ]; then
+    echo "mapper devices in use, performing mount of COS_STATE partition"
```
Vicente-Cheng, line 66: could we add a little bit more context on this? (through comment)

Vicente-Cheng (dismissed):
just a nit. thanks for fixing that!

tserong (dismissed):

Vicente-Cheng (dismissed):
Thanks!

tserong (approved):

Vicente-Cheng (approved):

ibrokethecloud:
@mergify backport v1.6

ibrokethecloud:
@mergify backport v1.5

PR #8853, opened 2025-08-08
Revert "Adding a webhook for the validation of CPU pinning"

brandboat:
#### Problem:
Can't add volume, delete and backup vm when vm enabled cpu pinning while cpu manager is disabled.

#### Solution:

revert https://github.com/harvester/harvester/commit/079cee711f0d3e91966cd4e97039920e346a6c4c

see https://github.com/harvester/harvester/pull/8800#issuecomment-3164167868 and https://github.com/harvester/harvester/pull/8800#issuecomment-3166406230

#### Related Issue(s):

#8778

#### Test plan:
- Enable CPU Manager in one node (e.g. harvester-node-0)
- Create a VM `test`
- Click Edit Config, enable CPU Pinning **WITHOUT** restart the VM.
- Disable CPU Manager in `harvester-node-0`
- Test add volume, delete and backup vm `test`, all operation should be successful.

N/A

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -12,8 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime"
+	runtime "k8s.io/apimachinery/pkg/runtime"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	ctlharvestercorev1 "github.com/harvester/harvester/pkg/generated/controllers/core/v1"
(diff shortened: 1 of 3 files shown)


w13915984028 (approved):
LGTM, thanks.

Meanwhile, the document needs some updates:
Mention a VM might be unscheduable and add a link to troubleshooting https://docs.harvesterhci.io/v1.5/vm/cpu-pinning#enable-cpu-pinning-on-a-new-vm

Add some examples which are related to cpu-pinning
 https://docs.harvesterhci.io/v1.5/troubleshooting/vm#unschedulable-virtual-machine
 ...

w13915984028:
BTW: @brandboat I also agree with @ihcsim https://github.com/harvester/harvester/pull/8800#pullrequestreview-3098792776 that when `disable cpuManager` action is triggered, it is better to check if some running VMs with CPU pinning enabled are hosted by current node. Not sure it is checked in the internal processing.

Sure, this can be done on next release if it is not there.

```
func (h ActionHandler) requestCPUManager(node *corev1.Node, policy ctlnode.CPUManagerPolicy) error {
```

brandboat:
> BTW: @brandboat I also agree with @ihcsim https://github.com/harvester/harvester/pull/8800#pullrequestreview-3098792776 that when disable cpuManager action is triggered, it is better to check if some running VMs with CPU pinning enabled are hosted by current node. Not sure it is checked in the internal processing.

@w13915984028, not sure if I understand correctly, we already have a such validator in https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/node/validator.go#L133

ihcsim (approved):

Vicente-Cheng (approved):
thanks, @brandboat will create an issue to discuss the UX in the next version

brandboat:
> thanks, @brandboat will create an issue to discuss the UX in the next version

Filed https://github.com/harvester/harvester/issues/8859

votdev (approved):
LGTM

Issue #8841, opened 2025-08-07
[BUG] Upgrading clusters with witness nodes from v1.5.1 to v1.6.0-rc5 failed resulting in Rancher flapping between versions


starbops:
### Describe the Bug

This is a clean install 3-node cluster with one witness node in v1.5.1. Upgrading the cluster to v1.6.0-rc5 failed with the following error messages:

<img width="1507" height="869" alt="Image" src="https://github.com/user-attachments/assets/6210f5d7-b85a-4c84-b19a-a1f07b9dff4c" />

The system service upgrade phase has already completed, as suggested in the Upgrade CR's status condition:

```yaml
status:
  conditions:
  - lastUpdateTime: "2025-08-07T07:59:38Z"
    reason: 'Internal error occurred: failed calling webhook "rancher.cattle.io.clusters.provisioning.cattle.io":
      failed to call webhook: Post "https://rancher-webhook.cattle-system.svc:443/v1/webhook/mutation/clusters.provisioning.cattle.io?timeout=10s":
      proxy error from 127.0.0.1:9345 while dialing 10.52.0.34:9443, code 502: 502
      Bad Gateway'
    status: "False"
    type: Completed
  - lastUpdateTime: "2025-08-07T07:12:09Z"
    status: "True"
    type: LogReady
  - lastUpdateTime: "2025-08-07T07:13:36Z"
    status: "True"
    type: ImageReady
  - lastUpdateTime: "2025-08-07T07:20:15Z"
    status: "True"
    type: RepoReady
  - lastUpdateTime: "2025-08-07T07:43:29Z"
    status: "True"
    type: NodesPrepared
  - lastUpdateTime: "2025-08-07T07:59:37Z"
    status: "True"
    type: SystemServicesUpgraded
  imageID: harvester-system/hvst-upgrade-5fqph
  nodeStatuses:
    node-0:
      state: Images preloaded
    node-1:
      state: Images preloaded
    node-witness:
      state: Images preloaded
```

Though the `SystemServicesUpgraded` condition was `True`, the embedded Rancher was not fully upgraded. The previously helm upgrade command failed:

```
$ helm history rancher -n cattle-system
REVISION        UPDATED                         STATUS          CHART           APP VERSION     DESCRIPTION
1               Thu Aug  7 04:23:06 2025        deployed        rancher-2.11.2  d0cd9e1-dirty   Install complete
[...]

Issue #5421, opened 2024-03-21
[FEATURE] Allow Harvester Node Promotion from GUI or Kubectl


dipakparmar:
**Is your feature request related to a problem? Please describe.**
Harvester version 1.2.1 automatically promotes nodes to management and compute roles, but in a DR scenario, this promotion does not happen automatically, causing cluster breakage even with VIP setup.

**Describe the solution you'd like**
Implement automatic node promotion to management and compute roles in DR scenarios to prevent cluster breakdown and ensure seamless failover. Also, documentation on how to manually do the node promotion and button from GUI.

**Describe alternatives you've considered**
Tried removing  node-role annotations and adding them back, as well as managing HCI node roles, but encountered issues like status staying in restarting and cluster breakage.

**Additional context**
The current manual workaround of restoring backup nodes is cumbersome and disrupts cluster operations. Automating node promotion in DR scenarios will enhance cluster resilience and simplify recovery processes.

notsrch:
This is very important for us.

We had a master node hardware failure (non-recoverable) on a master node.   While troubleshooting, a second master node went down.

bk201:
@Vicente-Cheng I believe we can specify the role now, but not via GUI.

Vicente-Cheng:
Yes, we introduce the role management on v1.3.0. Reference as below:
https://docs.harvesterhci.io/v1.3/install/harvester-configuration#installrole
https://docs.harvesterhci.io/v1.3/host/#role-management

Hi @dipakparmar, @notsrch
Could you check the document above and give some feedback?
Or would you like to share more about your scenario?

Thanks!

notsrch:
@Vicente-Cheng 
We had an 8 node harvester cluster running v1.3.1.   Nodes 1-3 were the master nodes.   Node#2 had a hardware failure.   3 days later, while troubleshooting Node#2, Node#3 failed.   Both were hardware failures and unrecoverable.

At least in my mind, a failure of only 2 nodes should not lead to a down and non-recoverable cluster.   It should be possible to promote a new node to master without deleting the failed one.   Ideally it should also happen automatically.

kfehrenbach:
Because of split brain scenarios i would suggest not to promote a node to master automatically. But there needs to be a solution to do it manually!

notsrch:
Another option that would also help for our use case is having 5 master nodes instead of 3.   Anything that helps survive through a double hardware failure or allow for quicker recovery from it is a net gain.

[...]

PR #8857, opened 2025-08-08
fix: update rancher chart value replicas to -2 for witness

starbops:
Originally, clusters with witness nodes has the `management.cattle.io/scale-available` annotation set to 2 (done by the node promotion controller). However, it will be overwritten to 3 after chart upgrade. This commit ensures clusters with witness nodes can still have the annotation set to 2 even after upgrade, by setting the chart value `replicas` to -2.

#### Related Issue(s):

#8841

#### Test plan:

1. Prepare a v1.5.1 3-node cluster with one witness node
2. Upgrade the cluster to master-head including the fix
3. The upgrade should complete successfully
4. Check the embedded Rancher Helm release by running `helm history rancher -n cattle-system`:
   ```
   REVISION        UPDATED                         STATUS          CHART           APP VERSION     DESCRIPTION
   1               Mon Aug 11 05:00:19 2025        superseded      rancher-2.11.2  d0cd9e1-dirty   Install complete
   2               Mon Aug 11 07:06:57 2025        deployed        rancher-2.12.0  8815e66-dirty   Upgrade complete
   ```
5. Check the values of the Rancher Helm release by running `helm get values rancher -n cattle-system`, the `replicas` should be `-2`:
   ```
   agentTLSMode: system-store
   antiAffinity: required
   bootstrapPassword: admin
   features: multi-cluster-management=false,multi-cluster-management-agent=false,managed-system-upgrade-controller=false
   global:
     cattle:
       psp:
         enabled: false
   hostPort: 0
   ingress:
     enabled: false
   noDefaultAdmin: false
   rancherImage: rancher/rancher
   rancherImagePullPolicy: IfNotPresent
   rancherImageTag: v2.12.0
   replicas: -2
   systemDefaultRegistry: ""
   tls: external
   useBundledSystemChart: true
   ```

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -573,6 +573,15 @@ upgrade_rancher() {
     exit 1
   fi
 
+  # Clusters with witness node should have rancher's replicas set to -2 if the total number of nodes is 3.
+  local total_nodes_count=$(kubectl get nodes -o json 2>/dev/null | jq -r '.items | length' || echo 0)
+  local witness_nodes_count=$(kubectl get nodes -l "node-role.harvesterhci.io/witness=true" -o json 2>/dev/null | jq -r '.items | length' || echo 0)
+  # Here we don't consider the case of multiple witness nodes, as we prohibit it in Harvester.
+  if [[ "$witness_nodes_count" -gt 0 && "$total_nodes_count" -eq 3 ]]; then
+      echo "3-node cluster with witness node detected, setting Rancher replicas to -2"
+      RANCHER_REPLICAS=-2 yq e '.replicas = env(RANCHER_REPLICAS)' values.yaml -i
+  fi
+
   # drop the potential manual patch upon shell-image to v0.1.26 on Harvester v1.3.2
   local shellimage=$(kubectl get settings.management.cattle.io shell-image -ojsonpath='{.value}')
   if [[ "$shellimage" = "rancher/shell:v0.1.26" ]]; then


package/upgrade/upgrade_manifests.sh
```diff
@@ -573,6 +573,13 @@ upgrade_rancher() {
     exit 1
   fi
 
+  # Clusters with witness node should have rancher's replicas set to -2
+  local witness_node_count=$(kubectl get nodes -l "node-role.harvesterhci.io/witness=true" -o json 2>/dev/null | jq -r '.items | length' || echo 0)
+  if [ "$witness_node_count" -gt 0 ]; then
+      echo "Cluster has witness nodes"
+      RANCHER_REPLICAS=-2 yq e '.replicas = env(RANCHER_REPLICAS)' values.yaml -i
```
w13915984028, line 582: can this handle the case `1 control + 1 witness` cluster?

or `2 control + 1 witness + n worker node`

Rancher does not request itself to be deployed on `control-plane` nodes

https://github.com/rancher/rancher/blob/5cafe3312fac9c5b5508b10e645d11037d24f872/chart/templates/deployment.yaml#L39

w13915984028: the smaller value between  `non-witness node count` and `3`

refer https://github.com/harvester/harvester/issues/8841#issuecomment-3167284571

starbops: Good ask. Though the `1 control + 1 witness` configuration makes little sense, we still need to take care of it, as we can't forbid users from doing it.

We can also get the total node number to reason about what value the `replicas` field should be.

w13915984028: Try this https://github.com/rancher/rancher/blob/5cafe3312fac9c5b5508b10e645d11037d24f872/chart/templates/deployment.yaml#L66 to to exclude witness,

it gives us a way to patch it from installer and upgrade

innobead: We may adopt the suggestion from @w13915984028 and improve this later. 

cc @Vicente-Cheng

starbops: For a `1 control + 1 witness` setup, the witness node will not be promoted, resulting in it becoming a semi-worker node. By semi, I mean that there will still be an etcd taint for that node; as a result, pods without proper tolerations cannot be scheduled to it. This is no exception for the Rancher pod. Under such a circumstance, the Rancher deployment still has the `management.cattle.io/scale-available` annotation set to `3`, but its `replicas` will be updated to `2` . One of the Rancher pods will remain pending. In this situation, the upgrade cannot even be started.

So, for this fix, I'd rather focus on N-node cluster (N >= 3) with one witness node cases.
- N < 3: do not touch the value (should remain in `-3`)
- N == 3: set replicas to `-2`
- N > 3: do not touch the value (should remain in `-3`)

WDYT?

w13915984028: For `1 control + 1 witness`, it might come from: `2 control + 1 witness` and later 1 control is deleted (e.g. the hardware is broken/the server is to be replaced, a intermediate state...);  but sure, it is rare case

as mentioned above, the current code seems not handling the node deletion

innobead: > For a `1 control + 1 witness` setup, the witness node will not be promoted, resulting in it becoming a semi-worker node. 

I believe what @starbops said is the current design. The min number of node for promotion is 3. 

[/pkg/controller/master/node/promote_controller.go#L114-L127](https://github.com/harvester/harvester/blob/76d7250f10ff63824abcc107e2d93d1b76da4d83/pkg/controller/master/node/promote_controller.go#L114-L127)
```go
	nodeList, err := h.nodeCache.List(labels.Everything())
	if err != nil {
		return nil, err
	}

	// early return if the node number not enough
	if len(nodeList) < defaultSpecManagementNumber {
		return node, nil
	}

	promoteNode := selectPromoteNode(nodeList)
	if promoteNode == nil {
		return node, nil
	}

```
I think we are good for the current resolution, but we can improve the promotion logic later including the node management at runtime. 

So, for this fix, I'd rather focus on N-node cluster (N >= 3) with one witness node cases.

> N < 3: do not touch the value (should remain in -3)
> N == 3: set replicas to -2
> N > 3: do not touch the value (should remain in -3)

@w13915984028 or @starbops, can you help create a ticket for a future improvement? This shouldn't only be for Rancher; any application replica change during promotion should be included in the future. 

cc @harvester/maintainer

innobead (dismissed):

package/upgrade/upgrade_manifests.sh
```diff
@@ -573,6 +573,14 @@ upgrade_rancher() {
     exit 1
   fi
 
+  # Clusters with witness node should have rancher's replicas set to -2 if the total number of nodes is 3.
+  local total_nodes_count=$(kubectl get nodes -o json 2>/dev/null | jq -r '.items | length' || echo 0)
+  local witness_nodes_count=$(kubectl get nodes -l "node-role.harvesterhci.io/witness=true" -o json 2>/dev/null | jq -r '.items | length' || echo 0)
+  if [[ "$witness_nodes_count" -gt 0 && "$total_nodes_count" -eq 3 ]]; then
```
innobead, line 579: nit: 
```suggestion
      echo "Updating the replica of embedded Rancher deployment to 2, since the Cluster has witness nodes"
```

package/upgrade/upgrade_manifests.sh
```diff
@@ -573,6 +573,14 @@ upgrade_rancher() {
...
+      echo "3-node cluster with witness node detected, setting Rancher replicas to -2"
```
innobead, line 581: Having a log is good.

innobead (dismissed):

package/upgrade/upgrade_manifests.sh
starbops: Sure, I can help create one for the node promotion enhancement.

starbops: A new ticket has been created #8865. Please feel free to update the context, thank you.

innobead (approved):

Vicente-Cheng:
Sorry for the late update.

Validate upgrade successfully
From v1.5.1 to v1.6-head with this PR.

Thanks for fixing that!

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: We have the related issue https://github.com/harvester/harvester/issues/5421.
We could consider that together.

Vicente-Cheng:
And the code logic looks good to me.
Thanks for the detailed comment and discussion!

Issue #8888, opened 2025-08-13
[BUG] Upgrading to v1.6-head gets stuck at the post-draining phase if not using multipath disks


starbops:
### Describe the Bug

A 3-node cluster with one witness node in v1.5.1 upgrading to v1.6-head got stuck at the post-draining phase indefinitely. The log of the post-drain pod `hvst-upgrade-rckrk-post-drain-node-0-xr5hm` showed that:

```
...
+ check_and_mount_state
++ chroot /host blkid -L COS_STATE
++ grep mapper
+ MAPPER_IN_USE=
+ clean_up_tmp_files
+ '[' -n /host/tmp/tmp.cKzJqpdTVL ']'
+ is_mounted /host/tmp/tmp.cKzJqpdTVL
+ mount
+ awk -v DIR=/host/tmp/tmp.cKzJqpdTVL '{ if ($3 == DIR) { exit 0 } } ENDFILE { exit 1 }'
+ echo 'Try to unmount /host/tmp/tmp.cKzJqpdTVL...'
+ umount /host/tmp/tmp.cKzJqpdTVL
Try to unmount /host/tmp/tmp.cKzJqpdTVL...
+ echo 'Clean up tmp files...'
+ '[' -n '' ']'
+ '[' -n /host/usr/local/upgrade_tmp/tmp.1MdNHNpy9W ']'
+ echo 'Try to remove /host/usr/local/upgrade_tmp/tmp.1MdNHNpy9W...'
+ rm -vf /host/usr/local/upgrade_tmp/tmp.1MdNHNpy9W
Clean up tmp files...
Try to remove /host/usr/local/upgrade_tmp/tmp.1MdNHNpy9W...
removed '/host/usr/local/upgrade_tmp/tmp.1MdNHNpy9W'
+ is_mounted /run/cos/state
+ mount
+ awk -v DIR=/run/cos/state '{ if ($3 == DIR) { exit 0 } } ENDFILE { exit 1 }'
[Wed Aug 13 06:37:34 UTC 2025] Running "upgrade_node.sh post-drain" errors, will retry after 10 minutes (12 retries)..
```

Note that the nodes of the cluster did not use any multipath disks.

### To Reproduce

1. Prepare a v1.5.1 3-node cluster with one witness node
2. Trigger upgrade to v1.6-head
3. The upgrade gets stuck at the post-draining phase of the first node (if it's a single-node cluster, it will stuck at the single-node-upgrade job)

### Expected Behavior

[...]

PR #8889, opened 2025-08-13
fix(upgrade): ensure not exiting immaturely for non-mpath nodes

starbops:
#### Related Issue(s):

#8888

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -49,7 +49,7 @@ sparsify_passive_img()
 # if the disk is already mounted then the current logic remounts them as read/write partition
 # and if a partition is not mounted elemental tries to just mount it as a read/write partition
 # in case of multipath devices, the /dev/mapper device is not detected, and elemental
-# finds the path making the disk as not mounted. 
+# finds the path making the disk as not mounted.
 # due to change introduced by https://github.com/rancher/elemental-toolkit/pull/2302 elemental reconciles
 # the actual disk partition in use and finds the correct mapper device related to the cos partition
 # however the logic still scans the underlying disk for mount path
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -61,8 +61,9 @@ sparsify_passive_img()
 # remount option is select. Mounting the partition under /run/cos/state works around this,
 # as elemental now finds a disk mounted under /run/cos/state and uses the `remount` option to mount the partition
 check_and_mount_state(){
-  MAPPER_IN_USE=$(chroot ${HOST_DIR} blkid -L COS_STATE |grep mapper)
-  if [ -n "$MAPPER_IN_USE" ]; then
+  local exit_code=0
+  chroot ${HOST_DIR} blkid -L COS_STATE 2>/dev/null | grep -q mapper || exit_code=$?
```
w13915984028: use `local` to escape the grep error

```
harv21:/home/rancher # ./a1.sh 
1
0
harv21:/home/rancher # cat a1.sh 
test() {
  MAPPER_IN_USE=$(blkid -L COS_STATE |grep mapper)
  echo $?
  local MAPPER_IN_USE1=$(blkid -L COS_STATE |grep mapper)
  echo $?
}

test
```

starbops: I think it makes sense. Let me update the PR. Thank you.

w13915984028 (approved):
LGTM, thanks for the quick fix.

Vicente-Cheng (approved):
thanks for the quick fix!

ibrokethecloud (approved):

ibrokethecloud:
@mergify backport v1.5

Issue #8890, opened 2025-08-13
[BUG] Clone vm with cpu/memory hotplug enabled, the cloned vm will use Maximum memory value


TachunLin:
### Describe the Bug

When we clone a vm that already enabled the CPU and memory hotplug and set the maximum values.

The cloned new vm will use the original `Maximum Memory` as it's vm memory. 

* Original vm, have 1 core and `1Gi` memory.

  <img width="1600" height="785" alt="Image" src="https://github.com/user-attachments/assets/11a9605e-87d1-40d4-b039-931990f84967" />
  
* Cloned vm, have 1 core and `4Gi` memory.

  <img width="1600" height="782" alt="Image" src="https://github.com/user-attachments/assets/82bf0ab1-dd2c-402b-a4d8-3370dd2b182a" />

  https://github.com/user-attachments/assets/7fc6567a-f780-4972-9a15-017426de4b2f

   

### To Reproduce

1. Create a new vm
2. Enable the CPU and memory hotplug option
3. Set the maximum cpu and memory value 
4. Clone the vm to a new one
5. Check the CPU, memory value of the cloned vm

### Expected Behavior

Cloned new vm should have the same memory size with the original one

### Support Bundle for Troubleshooting

[supportbundle_aef7daa8-6e33-42b4-95d6-a70df5abc0dc_2025-08-13T07-32-31Z.zip](https://github.com/user-attachments/files/21749194/supportbundle_aef7daa8-6e33-42b4-95d6-a70df5abc0dc_2025-08-13T07-32-31Z.zip)

### Environment

- Harvester version: v1.6.0-rc5
- Impacted VM: cloned-vm
- Underlying Infrastructure: Single node kvm machine
- Rancher version: v2.12.0
- Harvester UI extension: v1.6.0-rc5

### Additional context

Also exists on three nodes Harvester cluster.

### Workaround and Mitigation

N/A

TachunLin:
Verified **fixed** on `v1.6.0-rc6`. Close this issue.

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ The cloned vm will use the same memory size as original vm $~~$</summary>

  1. Original vm have 1 core and 1Gi memory 

  <img width="1595" height="787" alt="Image" src="https://github.com/user-attachments/assets/9ca5fd5a-6ba2-4190-9ebc-4e891bb0ee56" />

  2. The cloned vm also have the same 1Gi memory and the 4Gi maximum memory

[...]

PR #8892, opened 2025-08-13
fix: clone harvesterhci.io/enableCPUAndMemoryHotplug annotation

FrankYang0529:
#### Problem:
We didn't clone `harvesterhci.io/enableCPUAndMemoryHotplug` to the new VM, so mutator uses `resources.limits.memory` as VM memory.

#### Solution:
Clone `harvesterhci.io/enableCPUAndMemoryHotplug` annotation.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8890

#### Test plan:
1. Create a new vm
2. Enable the CPU and memory hotplug option
3. Set the maximum cpu and memory value 
4. Clone the vm to a new one
5. Check the CPU, memory value of the cloned vm

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -61,6 +61,7 @@ const (
 var (
 	cloneVMAnnotationKeys = []string{
 		util.AnnotationReservedMemory,
+		util.AnnotationEnableCPUAndMemoryHotplug,
 	}
 )
 


Vicente-Cheng (approved):

w13915984028 (approved):
LGTM, thx.

Issue #8847, opened 2025-08-07
[BUG] DHCP in OverlayNetwork VMs doesn't work


bcdurden:
### Describe the Bug

When using DHCP within a Subnet attached to an OverlayNetwork NAD for a VM, DHCP is not working correctly.

The VM has no default route and the supplied DNS of `10.53.0.10` is unreachable. This DNS is on the service CIDR range and not a real endpoint unless RKE2's CoreDNS is configured with nodelocal, and even then it is still not reachable due to routing rules with the localdns pods.

Here is a VM using default DHCP settings within a Subnet with a 10.20.0.0/24 cidr range attached to the ovn-cluster VPC (the default vpc). The IP assigned is reflected in the UI but the VM is not reachable over SSH unless you are within a VM on the same subnet already.

Once in the VM, we can see the assigned IP:

```console
$ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute 
       valid_lft forever preferred_lft forever
2: enp1s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc fq_codel state UP group default qlen 1000
    link/ether 3a:74:8f:c9:e6:be brd ff:ff:ff:ff:ff:ff
    inet 10.20.0.4/24 metric 100 brd 10.20.0.255 scope global dynamic enp1s0
       valid_lft 86312354sec preferred_lft 86312354sec
    inet6 fe80::3874:8fff:fec9:e6be/64 scope link 
       valid_lft forever preferred_lft forever
ubuntu@dhcp-test-uplink:~$ ip route
10.20.0.0/24 dev enp1s0 proto kernel scope link src 10.20.0.4 metric 100 
```

But the DNS is defaulted to `10.53.0.10` and connects upstream to CoreDNS:
```console
$ resolvectl status
Global
         Protocols: -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
  resolv.conf mode: stub

Link 2 (enp1s0)
    Current Scopes: DNS
         Protocols: +DefaultRoute -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
Current DNS Server: 10.53.0.10
[...]

PR #8914, opened 2025-08-15
add networking binding configuration for managedtap

ibrokethecloud:
#### Problem:

DHCP configuration is incomplete / fails when using overlay networks
#### Solution:

This can be addressed by editing the VM interface to use `managedtap` binding

To use this the kubevirt configuration needs to contain network binding settings for type `managedTap`.

The PR addresses by adding it to the kubevirt configuration. 

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8847

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -91,6 +91,10 @@ kubevirt:
         ## and then the user can focus on the above "defaultNetworkInterface".
         ##
         permitBridgeInterfaceOnPodNetwork: true
+        ## enable managedTap binding so allow vxlan based VM's to get DHCP correctly
+        binding:
+          managedtap:
+            domainAttachmentType: managedTap
 
     customizeComponents:
       patches:


Vicente-Cheng (approved):
thanks!

deploy/charts/harvester/values.yaml
```diff
@@ -91,6 +91,10 @@ kubevirt:
         ## and then the user can focus on the above "defaultNetworkInterface".
         ##
         permitBridgeInterfaceOnPodNetwork: true
+        ## enable managedTap binding so allow vxlan based VM's to get DHCP correctly
```
innobead, line 94: ref: https://kube-ovn.readthedocs.io/zh-cn/stable/en/kubevirt/dual-stack/

> In KubeVirt's Bridge network mode, the DHCP service is provided by virt-launcher. However, KubeVirt currently only implements DHCP for IPv4 single-stack, which prevents KubeVirt VMs in Bridge mode from dynamically obtaining IPv6 addresses through the RA (Router Advertisement) protocol. Although Kube-OVN provides DHCP and RA capabilities, these features are ineffective because KubeVirt intercepts DHCP/RA requests in advance.
> 
> In versions of KubeVirt after 1.4.0, a new Network Binding Plugin introduces a Bridge-like network mode called managedTap. In this mode, KubeVirt does not intercept DHCP requests. Therefore, by combining the new managedTap mode with Kube-OVN's DHCP/RA capabilities, it is possible to automatically obtain dual-stack network addresses for VMs.

innobead (approved):

ibrokethecloud:
@mergify backport v1.6

Issue #8581, opened 2025-07-01
[FEATURE] Expose kubevirt cluster wide migration settings via a harvester setting


ibrokethecloud:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

Kubevirt allows users to configure a number of cluster wide vm migration settings via the kubevirt cr: https://kubevirt.io/user-guide/compute/live_migration/#changing-cluster-wide-migration-configuration

A user needs to manually edit these settings via kubectl or any other k8s native tool to apply the changes in setting

### Describe the solution you'd like

We should expose these kubevirt cr fields via a harvester setting, and allow a user to configure these via the Harvester settings.

We should also have a UI enhacement to simplify configuration of these settings.

lanfon72:
Verified this feature has been applied.


Test Information
----
* Environment: **Baremetal DL360 4 nodes**
* Harvester Version: **v1.7.0-rc1**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Makes sure the setting could be fetch in the endpoint `apis/harvesterhci.io/v1beta1/settings/kubevirt-migration`
3. Update one onf the setting value (either via UI or API)
4. The change should be applied into kubevirt. (verify by executing `kubectl get kubevirt -nharvester-system -ojson | jq -r '.items[0].spec.configuration.migrations'`)

PR #8920, opened 2025-08-18
feat: add kubevirt-migration setting

FrankYang0529:
Use "kubevirt-migration" configuration to config `migrationConfiguration` in KubeVirt.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8581

#### Test plan:

Case 1: Use kubevirt-migration setting to update following fields
* parallelOutboundMigrationsPerNode
* parallelMigrationsPerCluster
* allowAutoConverge
* bandwidthPerMigration
* completionTimeoutPerGiB
* progressTimeout
* unsafeMigrationOverride
* allowPostCopy
* allowWorkloadDisruption
* disableTLS
* matchSELinuxLevelOnMigration

Case 2: Use kubevirt-migration setting to update following fields and will get webhook error
* nodeDrainTaintKey
* network

### DIFF
--- a/pkg/controller/master/setting/kubevirt_migration.go
+++ b/pkg/controller/master/setting/kubevirt_migration.go
@@ -1,0 +1,54 @@
+package setting
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+func (h *Handler) syncKubeVirtMigration(setting *harvesterv1.Setting) error {
+	var value string
+	if setting.Value != "" {
+		value = setting.Value
+	} else {
+		value = setting.Default
+	}
+
+	var err error
+	migrationConfiguration := &kubevirtv1.MigrationConfiguration{}
+
+	if value != "" {
+		err = json.Unmarshal([]byte(value), migrationConfiguration)
+		if err != nil {
+			return fmt.Errorf("invalid value: `%s`: %w", value, err)
+		}
+	}
+
+	kubevirt, err := h.kubeVirtConfigCache.Get(util.HarvesterSystemNamespaceName, util.KubeVirtObjectName)
+	if err != nil {
+		return fmt.Errorf("failed to get kubevirt object %v/%v %w", util.HarvesterSystemNamespaceName, util.KubeVirtObjectName, err)
+	}
+
+	kubevirtCpy := kubevirt.DeepCopy()
+	if kubevirtCpy.Spec.Configuration.MigrationConfiguration == nil {
+		kubevirtCpy.Spec.Configuration.MigrationConfiguration = &kubevirtv1.MigrationConfiguration{}
+	}
+	// ignore nodeDrainTaintKey and network field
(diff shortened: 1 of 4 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -1933,3 +1937,49 @@ func (v *settingValidator) validateRancherCluster(newSetting *v1beta1.Setting) e
...
+
+	vmims, err := v.vmimCache.List(metav1.NamespaceAll, labels.Everything())
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("Failed to list VM Migrations, err: %v", err.Error()))
+	}
+	for _, vmim := range vmims {
+		if vmim.DeletionTimestamp != nil || vmim.Status.MigrationState.Completed {
+			continue
+		}
+		return werror.NewBadRequest("There is a VM Migration in progress, please wait until it is completed before updating the kubevirt-migration setting")
```
Vicente-Cheng, line 1971: just curious, if we applied the new VM migration settings with the running vmim.
Does it encounter any errors?

If no, we could let the running vmim keep the current status, and the newly created vmim could use the new settings, right?

Vicente-Cheng (commented):
just a nit, others lgtm

Vicente-Cheng:
BTW, please rebased

pkg/webhook/resources/setting/validator.go
FrankYang0529: Currently, there is no error. However, I would like to avoid unexpected result here.

FrankYang0529: When updating migration configuration, the KubeVirt status has `DeploymentInProgress` condition. Although not all configurations really make redeployment, it's better to make sure there is no running migration.

pkg/controller/master/setting/kubevirt_migration.go
```diff
@@ -0,0 +1,54 @@
...
+
+	kubevirtCpy := kubevirt.DeepCopy()
+	if kubevirtCpy.Spec.Configuration.MigrationConfiguration == nil {
+		kubevirtCpy.Spec.Configuration.MigrationConfiguration = &kubevirtv1.MigrationConfiguration{}
+	}
+	// ignore nodeDrainTaintKey and network field
+	// The default nodeDrainTaintKey is "kubevirt.io/drain" and it's used in upgrade script.
+	// The network field is handled by "vm-migration-network" setting.
+	migrationConfiguration.NodeDrainTaintKey = nil
+	migrationConfiguration.Network = kubevirtCpy.Spec.Configuration.MigrationConfiguration.Network
```
Yu-Jack, line 45: Is it possible that users might have their own setting?

Yu-Jack (approved):
LGTM. Only one point.

pkg/webhook/resources/setting/validator.go
Vicente-Cheng: Thanks! Let's keep the current design

Vicente-Cheng (approved):
lgtm, thanks

pkg/controller/master/setting/kubevirt_migration.go
FrankYang0529: We already have vm-migration-network for this setting. It's better to avoid unexpected value here and just used value from vm-migration-network setting.

https://docs.harvesterhci.io/v1.6/advanced/vm-migration-network

PR #8925, opened 2025-08-19
minor change to fix check for external storage in the upgrade path

ibrokethecloud:
#### Problem:

`upgrade_node.sh` script checks existing harvester.config to verify if `externalStorageConfig` is enabled and if not, it patches `multipath` kernel arguments. This has been done in the past to allow us to migrate from various multipath disabling options when external storage support is not needed.

We are using `gopkg.in/yaml.v3` to marshal the configuration to yaml, which results in the following field

https://github.com/harvester/harvester-installer/blob/master/pkg/config/config.go#L213

Being encoded as follows in the saved `harvester.config`

```
os:
    externalstorage:
        enabled: true
```

This results in the field being `os.externalstorage.enabled` and not `os.externalStorageConfig.enabled`

#### Solution:

The PR fixes the check in the upgrade script to ensure field is accessed correctly, there by avoiding setting `multipath=off` when external storage is in use.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8689

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -650,7 +650,7 @@ EOF
   GRUBENV_FILE="/oem/grubcustom"
   chroot $HOST_DIR /bin/bash -c "if ! [ -f ${GRUBENV_FILE} ]; then touch ${GRUBENV_FILE}; fi"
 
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (approved):

tserong (approved):

ibrokethecloud:
@mergify backport v1.6

ibrokethecloud:
@mergify backport v1.5

Issue #6975, opened 2024-11-11
[FEATURE] Multipath device support for node-disk-manager


ibrokethecloud:
**Is your feature request related to a problem? Please describe.**

Currently in v1.4.0 multipath devices are skipped by node-disk-manager.

With introduction of disk less systems with 1.4.0 users running harvester on diskless systems like SAN arrays will become more common. Users may wish to leverage SAN storage lun's with Harvester and Longhorn for a seamless HCI experience.

**Describe the solution you'd like**

Currently the node-disk-manager code skips multipath devices while trying to reconcile blockdevice crds with underlying view of the OS. We should include additional logic to identify devices representing the multipath device, and prefer the multipath device over individual devices where possible.

**Describe alternatives you've considered**

**Additional context**

ibrokethecloud:
Related to: https://github.com/harvester/harvester/issues/6968

doccaz:
Describing the issue further... when LUNs are delivered via fibre-channel HBAs from an external storage, multiple devices are delivered with redundant paths (multipath), so either for balancing load and/or redundancy, the kernel drivers + device mapper take care of writing to the proper channels. It's an established Linux feature.  

The problem is, whenever a LUN is delivered to Harvester, and one tries to "Add Disk", it usually will try and take one of the multipath devices, instead of the device mapper one. This causes an immediate error as node-disk-manager tries to initialize the wrong device:


![Image](https://github.com/user-attachments/assets/42ef4615-bc5b-490f-b245-8841371edf27)
![Image](https://github.com/user-attachments/assets/00f80546-cff1-4163-b2cd-050a60f5b218)

A multipath setup will combine the constituent devices in a device mapper-created device:
```
harvester:~ # multipath -ll
33063000000123456 dm-0 QEMU,QEMU HARDDISK
size=20G features='0' hwhandler='0' wp=rw
|-+- policy='service-time 0' prio=1 status=active
| `- 0:0:0:0 sda 8:0  active ready running
`-+- policy='service-time 0' prio=1 status=enabled
  `- 0:0:0:1 sdb 8:16 active ready running
harvester:~ # ls -l /dev/disk/by-id/
total 0
lrwxrwxrwx 1 root root  9 Feb 13 04:50 ata-QEMU_DVD-ROM_QM00005 -> ../../sr0
lrwxrwxrwx 1 root root 10 Feb 13 04:50 dm-name-33063000000123456 -> ../../dm-0
[...]

PR #8936, opened 2025-08-20
HEP: support multipath device [skip ci]

Yu-Jack:
#### Problem:
Harvester doesn't support multipath device.

#### Solution:
Check this HEP for more details. 

Also, this one is the [node-disk-manager PR](https://github.com/harvester/node-disk-manager/pull/221), and this one is the [harvester-installer PR](https://github.com/harvester/harvester-installer/pull/1117). You can refer to those PRs for better understanding. 

#### Related Issue(s):
https://github.com/harvester/harvester/issues/6975

#### Other contexts

TODO are https://github.com/harvester/harvester/pull/8936#issuecomment-3244007850 after HEP meeting.

### DIFF
--- a/enhancements/20250820-support-multipath.md
+++ b/enhancements/20250820-support-multipath.md
@@ -4,12 +4,12 @@ Support for multipath disk devices in Harvester to enable reliable storage acces
 
 ## Summary
 
-This enhancement introduces multipath device recognition and management in Harvester's node-disk-manager component, allowing users to configure and utilize multipath storage devices safely while ensuring proper identification and avoiding conflicts with existing Longhorn v2 and LVM dm-devices.
+This enhancement introduces multipath device recognition and management in Harvester's node-disk-manager component, allowing users to configure and utilize multipath storage devices safely while ensuring proper identification and avoiding conflicts with existing Longhorn v2 and LVM DM-devices.
 
 ### Related Issues
 
 - Primary Issue: https://github.com/harvester/harvester/issues/6975
-- Non-wwn Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
+- Non-WWN Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
 
 ## Motivation
 
@@ -17,7 +17,7 @@ This enhancement introduces multipath device recognition and management in Harve
 
 - Enable Harvester to recognize and manage multipath disk devices properly
 - Support multipath device configuration in Harvester installer with extended configuration options
-- Distinguish between multipath devices and Longhorn v2 dm-devices to avoid conflicts
+- Distinguish between multipath devices and Longhorn v2 DM-devices to avoid conflicts
 - Ensure multipath devices are back after starting multipathd service when multipathd service is disabled.
 
 ### Non-goals
@@ -56,7 +56,7 @@ No direct API changes to existing Harvester APIs. The enhancement extends the in
 
 The multipath support is implemented through modifications to Harvester's node-disk-manager component:
 
-1. **Enhanced Device Recognition:** Added support for multipath device identification by extending device property extraction to handle `DM_SERIAL` and `DM_WWN` attributes specific to multipath devices. For example, we'll retrevive WWN value in this order from `udevadm info`: `ID_WWN_WITH_EXTENSION` -> `ID_WWN` -> `DM_WWN` and SERIAL value in this order: `ID_SERIAL_SHORT` -> `ID_SERIAL` -> `DM_SERIAL`
+1. **Enhanced Device Recognition:** Added support for multipath device identification by extending device property extraction to handle `DM_SERIAL` and `DM_WWN` attributes specific to multipath devices. For example, we'll retrieve WWN value in this order from `udevadm info`: `ID_WWN_WITH_EXTENSION` -> `ID_WWN` -> `DM_WWN` and SERIAL value in this order: `ID_SERIAL_SHORT` -> `ID_SERIAL` -> `DM_SERIAL`
 
 2. **Device Filtering:** Implemented logic to distinguish multipath devices from other dm-devices (such as Longhorn v2 volumes) using `multipath -C dm-x` command-line for verification. It will return an exit code of 0 if the dm-x device belongs to multipath service; otherwise, it will return 1. Besides, we also need to check whether /dev/xxx is managed by multipath service, using `multipath -c /dev/xxx`. It will also return an exit code of 0 if the /dev/xxx is managed by multipath service; otherwise, it will return 1. Besides, we should have a way to filter out other multipath devices that are used in other CSI drivers. We have this issue https://github.com/harvester/harvester/issues/5059 to filter out block device. We can make use of that mechanism to filter out multipath devices.
 
--- a/enhancements/20250820-support-multipath.md
+++ b/enhancements/20250820-support-multipath.md
@@ -90,7 +90,8 @@ For the second cluster, after the cluster is up, we need to run `systemctl enabl
 Additionally, you need to prepare two LUNs pointing to the same disk. If you're not sure about the setup, it should look like this in the `virt-install` command line:
 
 ```
-# Same serial numbers, and same physical disk
+# 1. Same serial numbers, and same physical disk
+# 2. Same serial numbers, same WWN, and same physical disk
 --disk path=/var/lib/libvirt/images/mydisk1.img,bus=scsi,cache=none,shareable=on,serial=disk1 \
 --disk path=/var/lib/libvirt/images/mydisk1.img,bus=scsi,cache=none,shareable=on,serial=disk1 \
 ```
--- a/enhancements/20250820-support-multipath.md
+++ b/enhancements/20250820-support-multipath.md
@@ -10,6 +10,7 @@ This enhancement introduces multipath device recognition and management in Harve
 
 - Primary Issue: https://github.com/harvester/harvester/issues/6975
 - Non-WWN Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
+- Support filter list on NDM: https://github.com/harvester/harvester/issues/5059 (out of scope for this enhancement)
 
 ## Motivation
 
@@ -17,12 +18,13 @@ This enhancement introduces multipath device recognition and management in Harve
 
 - Enable Harvester to recognize and manage multipath disk devices properly
 - Support multipath device configuration in Harvester installer with extended configuration options
-- Distinguish between multipath devices and Longhorn v2 DM-devices to avoid conflicts
+- Distinguish between multipath devices and deliberate vendor devices (ex: Longhorn v2 DM-devices) to avoid conflicts
 - Ensure multipath devices are back after starting multipathd service when multipathd service is disabled.
 
 ### Non-goals
 
 - Implementing multipath configuration without DM_WWN support (this will be addressed in a separate issue #6261)
+- Guaranteeing optimal multipathd performance
 
 ## Proposal
 
--- a/enhancements/20250820-support-multipath.md
+++ b/enhancements/20250820-support-multipath.md
@@ -1,0 +1,290 @@
+# Support Multipath Disk
+
+Support for multipath disk devices in Harvester to enable reliable storage access through multiple paths for improved fault tolerance and performance.
+
+## Summary
+
+This enhancement introduces multipath device recognition and management in Harvester's node-disk-manager component, allowing users to configure and utilize multipath storage devices safely while ensuring proper identification and avoiding conflicts with existing Longhorn v2 and LVM dm-devices.
+
+### Related Issues
+
+- Primary Issue: https://github.com/harvester/harvester/issues/6975
+- Non-wwn Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
+
+## Motivation
+
+### Goals
+
+- Enable Harvester to recognize and manage multipath disk devices properly
+- Support multipath device configuration in Harvester installer with extended configuration options
+- Distinguish between multipath devices and Longhorn v2 dm-devices to avoid conflicts
+- Ensure multipath devices are back after starting multipathd service when multipathd service is disabled.
+
+### Non-goals
+
+- Implementing multipath configuration without DM_WWN support (this will be addressed in a separate issue #6261)
+
+## Proposal
+
+This enhancement adds multipath device support to Harvester by implementing device identification capabilities in the node-disk-manager component and extending the installer configuration options.
+
+### User Stories
+
+#### Story 1: Be able to use external storage like SAN, iSCSI, or Fibre Channel in NDM
+
+Those disks expose different accessible paths to the node. However, even we aggregate those paths in to a single device with multipath service, NDM doesn't still support recognizing them. Right now, it's only able to be configured in Longhorn. That's why we need to support configuaring in Harvester.
+
+After NDM is able to identify those aggregated disks, users can directly select those disk in Harvester instead of crating the settings in Longhorn.
+
+#### Story 2: High availability and fault tolerance
+
+Without multipath support, users cannot configure Harvester with multipath storage devices. This limitation results in single points of failure in storage connectivity and prevents effective use of enterprise SAN storage.
+
+With multipath, storage paths can be aggregated into a single logical device, enabling high availability and fault tolerance. In this setup, users can configure multipath devices during Harvester installation, and the system will automatically fail over to alternative paths if one becomes unavailable.
+
+### User Experience In Detail
+
+Multipath devices appear in the Harvester UI as available storage devices and can be used for Longhorn storage configuration.
+
+### API changes
+
+No direct API changes to existing Harvester APIs. The enhancement extends the installer configuration schema to accept additional fields for multipath configuration.
+
+## Design
+
+### Implementation Overview
+
+The multipath support is implemented through modifications to Harvester's node-disk-manager component:
+
+1. **Enhanced Device Recognition:** Added support for multipath device identification by extending device property extraction to handle `DM_SERIAL` and `DM_WWN` attributes specific to multipath devices. For example, we'll retrevive WWN value in this order from `udevadm info`: `ID_WWN_WITH_EXTENSION` -> `ID_WWN` -> `DM_WWN` and SERIAL value in this order: `ID_SERIAL_SHORT` -> `ID_SERIAL` -> `DM_SERIAL`
+
+2. **Device Filtering:** Implemented logic to distinguish multipath devices from other dm-devices (such as Longhorn v2 volumes) using `multipath -C dm-x` command-line for verification. It will return an exit code of 0 if the dm-x device belongs to multipath service; otherwise, it will return 1. Besides, we also need to check whether /dev/xxx is managed by multipath service, using `multipath -c /dev/xxx`. It will also return an exit code of 0 if the /dev/xxx is managed by multipath service; otherwise, it will return 1. Besides, we should have a way to filter out other multipath devices that are used in other CSI drivers. We have this issue https://github.com/harvester/harvester/issues/5059 to filter out block device. We can make use of that mechanism to filter out multipath devices.
+
+3. **Improved Mount Handling:** Enhanced mount point detection to properly handle `/dev/mapper/xx` symlinks used by multipath devices. Because there isn't a `/dev/dm-x` mounting in `/proc/mounts`, there is only `/dev/mapper/xx` ([Ref](#Multipath-Device-Example)). Hence, we can only use `/dev/mapper/xx` to check.
+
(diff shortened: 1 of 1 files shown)


enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,285 @@
...
+    - Longhorn V2 Image with Longhorn V2 extra disk
+6. After starting the VMs, reboot the node. Verify that the virtual machines come back up successfully after the reboot
+
+### Upgrade strategy
+
+Some users might have already enabled multipathd and are using it. In such cases, the disks should be configured and added in the Longhorn UI by the users themselves. 
+
+If those disks are available in the Harvester UI, adding them in the Harvester UI will cause failure. Therefore, NDM or GUI should filter out disks that are already mounted and added in the Longhorn UI.
+
+By the way, if you add a normal disk and configure it in the Longhorn UI, it's still available in the Harvester UI. Adding it also causes failure. So, the behavior has been like this for a long time. The suitable way to migrate it from Longhorn to Harvester is to delete replica and backing images before adding the disks on the Harvester side.
```
ibrokethecloud, line 117: is there a separate issue tracking this behaviour?

ibrokethecloud (commented):
HEP looks good. 

Would it be worth while adding more context on details of how we filter out longhorn v2 disks from device manager.

The actual multipath device detection is easy to follow since we just use info for dm_wwn and dm_serial to identify disks.

Yu-Jack:
> Would it be worth while adding more context on details of how we filter out longhorn v2 disks from device manager.

Sure, let me add it.

enhancements/20250820-support-multipath.md
Yu-Jack: I haven't opened it. Let me open it later.

Yu-Jack: this one https://github.com/harvester/harvester/issues/8990

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,285 @@
...
+
+No direct API changes to existing Harvester APIs. The enhancement extends the installer configuration schema to accept additional fields for multipath configuration.
+
+## Design
+
+### Implementation Overview
+
+The multipath support is implemented through modifications to Harvester's node-disk-manager component:
+
+1. **Enhanced Device Recognition:** Added support for multipath device identification by extending device property extraction to handle `DM_SERIAL` and `DM_WWN` attributes specific to multipath devices.
```
Vicente-Cheng: Could we describe how we handle both WWN and DM_WWN?

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,285 @@
...
+
+2. **Device Filtering:** Implemented logic to distinguish multipath devices from other dm-devices (such as Longhorn v2 volumes) using `multipath -C dm-x` command-line for verification. It will return an exit code of 0 if the dm-x device belongs to multipath service; otherwise, it will return 1. Besides, we also need to check whether /dev/xxx is managed by multipath service, using `multipath -c /dev/xxx`. It will also return an exit code of 0 if the /dev/xxx is managed by multipath service; otherwise, it will return 1.
```
Vicente-Cheng: Looks like the DM_WWN is related to the ID_SERIAL of the slave device. Could we directly filter out the Longhorn device?

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,285 @@
...
+
+1. Cluster with multipath service enabled by default.
+2. Cluster with multipath service disabled by default.
+
+For the second cluster, after the cluster is up, we need to run `systemctl enable multipathd` and `systemctl start multipathd`. After that, continue with the following test steps.
+
+Additionally, you need to prepare two LUNs pointing to the same disk. If you're not sure about the setup, it should look like this in the `virt-install` command line:
+
+```
+# Same serial numbers, and same physical disk
```
Vicente-Cheng: Adding another test with the same serial and WWN would be great.

Vicente-Cheng (commented):
others lgtm, thanks for the HEP!

enhancements/20250820-support-multipath.md
Yu-Jack: From my point of view, It's better use `multipath` command to check device. It's a source of truth.

Yu-Jack:
TODO

- LVM test case

- Filter out multipath devices that are already mounted. Use status.deviceStatus.fileSystem.mountPoint, but consider the case where users unmount them and add them back in Harvester. These disks should become available again in the Harvester UI once unmounted.

- Documentation about migration from Longhorn to Harvester — users should evict volumes first if they truly want to migrate.

- Some iSCSI and SAN environments expose multiple physical paths, and the CSI driver cannot recognize them without multipathd. This is another reason we need the multipath service.

- Filtering out other vendors and products in NDM, since other CSI drivers might already support multipath devices. (We can use mechanism in https://github.com/harvester/harvester/issues/5059)

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+Support for multipath disk devices in Harvester to enable reliable storage access through multiple paths for improved fault tolerance and performance.
+
+## Summary
+
+This enhancement introduces multipath device recognition and management in Harvester's node-disk-manager component, allowing users to configure and utilize multipath storage devices safely while ensuring proper identification and avoiding conflicts with existing Longhorn v2 and LVM dm-devices.
+
+### Related Issues
+
+- Primary Issue: https://github.com/harvester/harvester/issues/6975
+- Non-wwn Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
```
votdev:
```suggestion
- Non-WWN Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
```

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
+# Support Multipath Disk
+
+Support for multipath disk devices in Harvester to enable reliable storage access through multiple paths for improved fault tolerance and performance.
+
+## Summary
+
+This enhancement introduces multipath device recognition and management in Harvester's node-disk-manager component, allowing users to configure and utilize multipath storage devices safely while ensuring proper identification and avoiding conflicts with existing Longhorn v2 and LVM dm-devices.
```
votdev:
```suggestion
This enhancement introduces multipath device recognition and management in Harvester's node-disk-manager component, allowing users to configure and utilize multipath storage devices safely while ensuring proper identification and avoiding conflicts with existing Longhorn v2 and LVM DM-devices.
```

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+- Primary Issue: https://github.com/harvester/harvester/issues/6975
+- Non-wwn Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
+
+## Motivation
+
+### Goals
+
+- Enable Harvester to recognize and manage multipath disk devices properly
+- Support multipath device configuration in Harvester installer with extended configuration options
+- Distinguish between multipath devices and Longhorn v2 dm-devices to avoid conflicts
```
votdev:
```suggestion
- Distinguish between multipath devices and Longhorn v2 DM-devices to avoid conflicts
```

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+
+### User Stories
+
+#### Story 1: Be able to use external storage like SAN, iSCSI, or Fibre Channel
+
+Since Longhorn doesn't support the multipath service, it will recognize different paths as separate disks, even though all the paths actually point to the same disk.
+
+In this case, we can't use Longhorn as our CSI driver directly. However, with the multipath service, the paths can be aggregated into a single device, which Longhorn can then use as the entry point. This means users can use the Longhorn CSI driver together with the multipath service.
+
+#### Story 2: Add multipath disk as Harvester backend storage.
```
votdev:
```suggestion
#### Story 2: Add multipath disk as Harvester backend storage
```

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+
+No direct API changes to existing Harvester APIs. The enhancement extends the installer configuration schema to accept additional fields for multipath configuration.
+
+## Design
+
+### Implementation Overview
+
+The multipath support is implemented through modifications to Harvester's node-disk-manager component:
+
+1. **Enhanced Device Recognition:** Added support for multipath device identification by extending device property extraction to handle `DM_SERIAL` and `DM_WWN` attributes specific to multipath devices. For example, we'll retrevive WWN value in this order from `udevadm info`: `ID_WWN_WITH_EXTENSION` -> `ID_WWN` -> `DM_WWN` and SERIAL value in this order: `ID_SERIAL_SHORT` -> `ID_SERIAL` -> `DM_SERIAL`
```
votdev:
```suggestion
1. **Enhanced Device Recognition:** Added support for multipath device identification by extending device property extraction to handle `DM_SERIAL` and `DM_WWN` attributes specific to multipath devices. For example, we'll retrieve WWN value in this order from `udevadm info`: `ID_WWN_WITH_EXTENSION` -> `ID_WWN` -> `DM_WWN` and SERIAL value in this order: `ID_SERIAL_SHORT` -> `ID_SERIAL` -> `DM_SERIAL`
```

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+
+2. **Device Filtering:** Implemented logic to distinguish multipath devices from other dm-devices (such as Longhorn v2 volumes) using `multipath -C dm-x` command-line for verification. It will return an exit code of 0 if the dm-x device belongs to multipath service; otherwise, it will return 1. Besides, we also need to check whether /dev/xxx is managed by multipath service, using `multipath -c /dev/xxx`. It will also return an exit code of 0 if the /dev/xxx is managed by multipath service; otherwise, it will return 1. Besides, we should have a way to filter out other multipath devices that are used in other CSI drivers. We have this issue https://github.com/harvester/harvester/issues/5059 to filter out block device. We can make use of that mechanism to filter out multipath devices.
+
+3. **Improved Mount Handling:** Enhanced mount point detection to properly handle `/dev/mapper/xx` symlinks used by multipath devices. Because there isn't a `/dev/dm-x` mounting in `/proc/mounts`, there is only `/dev/mapper/xx` ([Ref](#Multipath-Device-Example)). Hence, we can only use `/dev/mapper/xx` to check.
```
votdev, line 65: `/dev/mapper/xx` are symlinks to `/dev/dm-x`.
```
# ls -alh /dev/mapper/vg0-lv0
lrwxrwxrwx 1 root root 7 Sep  2 10:13 /dev/mapper/vg0-lv0 -> ../dm-0
# realpath /dev/mapper/vg0-lv0
/dev/dm-0
```
Keeping that in mind might help to implement the check.

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,292 @@
...
+
+2. **Device Filtering:** Implemented logic to distinguish multipath devices from other dm-devices (such as Longhorn v2 volumes) using `multipath -C dm-x` command-line for verification. It will return an exit code of 0 if the dm-x device belongs to multipath service; otherwise, it will return 1. Besides, we also need to check whether /dev/xxx is managed by multipath service, using `multipath -c /dev/xxx`. It will also return an exit code of 0 if the /dev/xxx is managed by multipath service; otherwise, it will return 1. Besides, we should have a way to filter out other multipath devices that are used in other CSI drivers. We have this issue https://github.com/harvester/harvester/issues/5059 to filter out block device. We can make use of that mechanism to filter out multipath devices.
```
votdev, line 63: The DM (Device Mapper) framework keeps the information about which devices are assigned to a device in `sysfs`.
Here an example from a LVM device:
```
# ls -alh /sys/block/dm-0/slaves/
total 0
drwxr-xr-x 2 root root 0 Sep  2 10:13 .
drwxr-xr-x 9 root root 0 Sep  2 10:13 ..
lrwxrwxrwx 1 root root 0 Sep  2 10:13 sdb -> ../../../../pci0000:00/0000:00:03.0/host2/target2:0:2/2:0:2:0/block/sdb
```
Maybe this information can be used somehow instead of executing an userland app which is costly.

votdev (changes requested):

enhancements/20250820-support-multipath.md
Yu-Jack: Yeah. That's what I used in the implementation. Thanks.

ibrokethecloud:
Based on our HEP review we need to handle the following scenarios
* skip multipath devices which may already be mounted as we do not want to end up formatting in use disks via the `BlockDevice` interface
* we also need to provide a means to allow users to filter multipath devices based on vendor etc, this can be exposed as a setting. Some 3rd party CSI may already be using multipath, and with this option we can allow users to filter these devices from being managed as `BlockDevices`

Yu-Jack:
> * we also need to provide a means to allow users to filter multipath devices based on vendor etc, this can be exposed as a setting. Some 3rd party CSI may already be using multipath, and with this option we can allow users to filter these devices from being managed as `BlockDevices`

FYI, this is tracked on https://github.com/harvester/harvester/issues/5059. So, this HEP only focuses on identifying multipath device.

enhancements/20250820-support-multipath.md
WebberHuang1118: Using `dmsetup table` is also an option; the output will describe the DM target type.

WebberHuang1118 (dismissed):
LGTM, thanks.

enhancements/20250820-support-multipath.md
WebberHuang1118: We can also consider using `dmsetup ls --tree` to view the multipath device architecture.

Yu-Jack: Actually, if we have a wrapped command, I prefer to use that. At least, it helps us do the searching in different cases.

In my testing, if we use LVM with two disk in a same group. The result is like below. You can see there are devices under slaves folder. Obviously, checking slaves folder is not enough. So, using `multipath` command is easier for us.  I also think `dmsetup ls -tree` might be a bit complicated if we try to parse it and retrieve the results from the command. But, I'm glad to know we have those methods to check information.

image

image

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
+# Support Multipath Disk
+
+Support for multipath disk devices in Harvester to enable reliable storage access through multiple paths for improved fault tolerance and performance.
+
+## Summary
+
+This enhancement introduces multipath device recognition and management in Harvester's node-disk-manager component, allowing users to configure and utilize multipath storage devices safely while ensuring proper identification and avoiding conflicts with existing Longhorn v2 and LVM DM-devices.
+
+### Related Issues
```
innobead, line 9: Add https://github.com/harvester/harvester/issues/5059

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+- Primary Issue: https://github.com/harvester/harvester/issues/6975
+- Non-WWN Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
+
+## Motivation
+
+### Goals
+
+- Enable Harvester to recognize and manage multipath disk devices properly
+- Support multipath device configuration in Harvester installer with extended configuration options
+- Distinguish between multipath devices and Longhorn v2 DM-devices to avoid conflicts
```
innobead:
```suggestion
- Distinguish between multipath devices and deliberate vendor devices (ex: Longhorn v2 DM-devices) to avoid conflicts
```

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+- Ensure multipath devices are back after starting multipathd service when multipathd service is disabled.
+
+### Non-goals
```
innobead, line 24: nit: Worth mentioning that performance guarantee is not a non-goal, as Longhorn prefers local disks for better performance.

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+
+Those disks expose different accessible paths to the node. However, even we aggregate those paths in to a single device with multipath service, NDM doesn't still support recognizing them. Right now, it's only able to be configured in Longhorn. That's why we need to support configuaring in Harvester.
+
+After NDM is able to identify those aggregated disks, users can directly select those disk in Harvester instead of crating the settings in Longhorn.
+
+#### Story 2: High availability and fault tolerance
+
+Without multipath support, users cannot configure Harvester with multipath storage devices. This limitation results in single points of failure in storage connectivity and prevents effective use of enterprise SAN storage.
+
+With multipath, storage paths can be aggregated into a single logical device, enabling high availability and fault tolerance. In this setup, users can configure multipath devices during Harvester installation, and the system will automatically fail over to alternative paths if one becomes unavailable.
```
innobead, line 45: Besides day 1 (installation time), how about day 2 at runtime?

enhancements/20250820-support-multipath.md
```diff
@@ -0,0 +1,290 @@
...
+- Non-WWN Device Handling Issue: https://github.com/harvester/harvester/issues/6261 (out of scope for this enhancement)
+
+## Motivation
+
+### Goals
+
+- Enable Harvester to recognize and manage multipath disk devices properly
+- Support multipath device configuration in Harvester installer with extended configuration options
+- Distinguish between multipath devices and Longhorn v2 DM-devices to avoid conflicts
+- Ensure multipath devices are back after starting multipathd service when multipathd service is disabled.
```
innobead, line 22:
```suggestion
- Ensure multipath devices are back after starting multipathd service when multipathd service is enabled.
```

enhancements/20250820-support-multipath.md
innobead: Will the failure to add a wrongly recognized device cause data loss? Need to consider data protection if needed.

enhancements/20250820-support-multipath.md
innobead: Ensuring that multipath devices are available should be solely the responsibility of multipathd, so this may not need to be listed here as a goal?

innobead (commented):
In general, LGTM, and thanks for preparing the HEP review.

enhancements/20250820-support-multipath.md
Yu-Jack: Yeah, good point. It's multipath service stuff. We can remove this.

enhancements/20250820-support-multipath.md
Yu-Jack: I don't get your meaning. multipathd service is supposed to run normally from day 1. I just mentioned what mulitpath can do in this section. This HEP only focuses on identifying the multipath devices in Harvester UI.

innobead: I meant below. In fact, users should be able to configure devices at runtime after installation?

>  In this setup, users can configure multipath devices during Harvester installation,

enhancements/20250820-support-multipath.md
Yu-Jack: Well, I don't think we should allow this happen. so we definitely need to filter out those disks that are already mounted. Just like @ibrokethecloud said, we can simply filter out those disks by checking mount point.

But, I still did some tests about duplicate provision for the evidence.

### Case 1 empty replicas and backing images:

It show this error in GUI.

image

image

```
# in longhorn node diskStatus
disk-1:
  conditions:
  - lastProbeTime: ""
    lastTransitionTime: "2025-09-04T04:08:33Z"
    message: 'Disk disk-1(/home/rancher/data2) on node harvester1 is not ready:
      record diskUUID doesn''t match the one on the disk '
    reason: DiskFilesystemChanged
    status: "False"
    type: Ready
  - lastProbeTime: ""
    lastTransitionTime: "2025-09-04T04:08:33Z"
    message: Disk disk-1 (/home/rancher/data2) on the node harvester1 is not ready
    reason: DiskNotReady
    status: "False"
    type: Schedulable
```

### Case 2 there are replicas or backing images:

Before duplicate provision:

image


After:

Besides, the mount point was reset first time, but wasn't reset at second attempt. It might need more tests. But, I'm still able to rebuild the replicas in longhorn side.

image

Yu-Jack: I just tested it again. It seems if there are still replicas in the disk, we can still schedule replicas on the disk even disk is added in the Harvester UI. It just show "target is busy".

enhancements/20250820-support-multipath.md
Yu-Jack: > I meant below. In fact, users should be able to configure devices at runtime after installation?

Actually, I'm not sure how they did at runtime before (Before this HEP, some users might have already used multipath service). But, in my case, I'll edit `/oem/90_custom.yaml` for permanent usage like Harvester document mentioned here https://docs.harvesterhci.io/v1.5/install/update-harvester-configuration#configuration-persistence and `/etc/multipath.conf` at run time.

Yu-Jack:
@ibrokethecloud @innobead @tserong @Vicente-Cheng @votdev @WebberHuang1118 Gentle ping on this. Thanks!

Vicente-Cheng (approved):
lgtm, let's work on the corresponding PRs

WebberHuang1118 (approved):

votdev (approved):

enhancements/20250820-support-multipath.md
innobead: Filter out disks that are being used beforehand, or simply fail without issues afterward. All works for us; at least from your tests, data loss is not an issue here.

innobead (approved):

Issue #8934, opened 2025-08-19
[BUG] Replica Replenishment Wait Interval from longhorn settings will be reset to the value stored in Upgrade CR annotation


wheatdog:
### Describe the Bug

When restarting the harvester controller, it will pick up the latest completed Upgrade CR and `Replica Replenishment Wait Interval` from longhorn settings will be reset to the value stored in the annotation of Upgrade CR.

### To Reproduce

1. Provision a harvester cluster (at least 2 nodes)
2. Perform an upgrade (either to the same version or newer version) and wait until it finish successfully
3. Modify Replica Replenishment Wait Interval 600 -> 60. (This can be set from embedded Longhorn UI settings)
4. Restart harvester controller
    ```
    kubectl rollout -n harvester-system restart deployment/harvester
    ```
6. Replica Replenishment Wait Interval will be reset to 600 (the value stored in the annotation of Upgrade CR)

### Expected Behavior

Replica Replenishment Wait Interval should not be modified after restarting harvester controller

### Support Bundle for Troubleshooting

n/a

### Environment

- Harvester version: v1.4.2
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Workaround and Mitigation

Here are the potential workarounds:
1. remove this annotation `harvesterhci.io/replica-replenishment-wait-interval` from the completed Upgrade CR
2. delete the completed Upgrade CR -- but right after delete it, `replica-replenishment-wait-interval` will be reset. We must set it to the correct value manually again

The idea is to prevent harvester controller accidentally reset the setting.

wheatdog:
Confirmed that this can also be reproduced on the current latest release `v1.6.0-rc6`.

wheatdog:
Another Upgrade CR lifecycle related issue: #8442

TachunLin:
When upgrade from `v1.7.0-dev-20250914` to the same version.

  <img width="1598" height="783" alt="Image" src="https://github.com/user-attachments/assets/5222f321-0e35-486f-839d-bba44390dcd1" />


* Testing was blocked since the upgrade stuck in `Creating Upgrade Repository`

    <img width="1598" height="793" alt="Image" src="https://github.com/user-attachments/assets/decd00dd-09b8-4d97-a40b-fb218ebfcc7c" />

* The virt-launcher-upgrade-repo keep in `init` state with warning
[...]

Issue #8442, opened 2025-06-09
[BUG] Potential upgrade flow control bug when post-drain job finishes


starbops:
### Describe the Bug

During the Harvester Upgrade process, particularly in the node upgrade phase, the upgrade controller reconciles the node upgrade jobs and waits for the nodes to reboot and come back online.

Precisely, the post-drain job will upgrade the node's operating system. After the OS is upgraded, it will reboot the node. On the upgrade controller side, whenever it detects the post-drain job finished where its corresponding node is in the `Post-draining` node upgrade status, it will do the following two things:
1. Change the node upgrade status to `Waiting reboot` in the Upgrade CR
2. Set the `harvesterhci.io/pendingOSImage` annotation to the node resource for later comparison
https://github.com/harvester/harvester/blob/ccabfab00f0c97b898a27408fe730d5b55bd5702/pkg/controller/master/upgrade/job_controller.go#L141-L144

Chances are, the node label update request fails, but it is never retried again, introducing a state inconsistency between the Upgrade CR and the Node resource, hence stall the entire upgrade process.

### To Reproduce

It's hard to reproduce. Essentially, you need to make the node's update requests fail at that specific time point (due to a conflict). But it does happen on the field, with low odds.

### Expected Behavior

If the node update request fails, the controller should at least try it again, rather than leaving it and introducing an inconsistency.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version:
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Additional context

We should probably avoid using `return nil, err` in the `OnChange()` control loops. Most of the time, we'll want the controller to reconcile the resource again when something is wrong; otherwise, critical time points might pass and never come back. This is rather important for the Harvester Upgrade process, since it's a big state machine.

### Workaround and Mitigation

Manually annotate the node that should've been annotated with the `harvesterhci.io/pendingOSImage` key and value. The upgrade progress should proceed after that.

Vicente-Cheng:
Due to the issue was hard to reproduce, verified as the logger added. Close it.

PR #8941, opened 2025-08-21
fix(upgrade): prevent longhorn settings be reset to the value stored in Upgrade CR unexpectedly

wheatdog:
#### Problem:

When restarting the harvester controller, it will pick up the latest completed Upgrade CR and `Replica Replenishment Wait Interval` from longhorn settings will be reset to the value stored in the annotation of Upgrade CR. But the value stored in the annotation might already be stale.

#### Solution:

To prevent unexpected restoring again, a new annotation `harvesterhci.io/longhorn-settings-restored` is introduced to be set right after the corresponding settings are restored. If this annotation is set, the longhorn settings value restoration would be skipped.

#### Related Issue(s):

#8934 

#### Test plan:

1. Provision a harvester cluster (at least 2 nodes)
2. Perform an upgrade (2 scenarios)
    1. from version **without this fix** to version **with this fix**
    2. from version **with this fix** to version **with this fix**
3. Modify Replica Replenishment Wait Interval 600 -> 60.
4. Restart harvester controller
    ```
    kubectl rollout -n harvester-system restart deployment/harvester
    ```
5. Replica Replenishment Wait Interval should still be 60
6. The new annotation should be set from Upgrade CR

n/a


pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -461,55 +465,65 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
...
+			return nil, err
 		}
 	}
 
 	// tear down logging infra if any
 	if harvesterv1.LogReady.IsTrue(upgrade) && upgrade.Status.UpgradeLog != "" {
 		upgradeLog, err := h.upgradeLogCache.Get(upgradeNamespace, upgrade.Status.UpgradeLog)
 		if err != nil {
-			if apierrors.IsNotFound(err) {
-				return nil
```
wheatdog, line 496: Previously, we have this early return when upgradeLog CR couldn't be fetched. I think it is a little bit weird since this would skip the `h.resumeManagedCharts()` part below.

So I remove the early return part in this PR.

wheatdog: @w13915984028 @starbops I saw you have added `resumeManagedCharts` and `upgradeLog` related changes before. Can you help me review this logic? Thanks.

w13915984028: @wheatdog Did you consider to improve below code, after restoring saved data to LH, then remove the annotation from upgrade object, thus the change is limited.

```
	// restore Longhorn replica-replenishment-wait-interval and
	// auto-cleanup-system-generated-snapshot settings (multi-node cluster only)
	if upgrade.Status.SingleNode == "" {
		if err := h.loadReplicaReplenishmentFromUpgradeAnnotation(upgrade); err != nil {
			return err
		}
		if err := h.loadAutoCleanupSystemGeneratedSnapshotFromUpgradeAnnotation(upgrade); err != nil {
			return err
		}
	}
```

The `upgrade` controller itself has an issue atm.: it does not check the `latest-upgrade` related mark on the beginning of `OnChange`,  after an upgrade CR is fully processed, it should be put to a stale status, and the controller should avoid to process it again. But it needs a more complicated processing than your PR.

wheatdog: @w13915984028 Got it, thanks for the suggestion. I'll revise the PR as your suggestion to limit the change scope.

starbops: @wheatdog IIRC, the UpgradeLog update part used to be the last step of the entire Upgrade cleanup operation. Anyways, we can fix this in the PR. Thank you.

wheatdog: @w13915984028 btw, is there a ticket to track this issue you mentioned?

> The upgrade controller itself has an issue atm.: it does not check the latest-upgrade related mark on the beginning of OnChange, after an upgrade CR is fully processed, it should be put to a stale status, and the controller should avoid to process it again.

w13915984028: There is one https://github.com/harvester/harvester/issues/8442, but is not dedicatedly for the upgrade controller state machine refactor when it was filed.

pkg/codegen/main.go
```diff
@@ -179,6 +180,13 @@ func main() {
 				GenerateTypes:   false,
 				GenerateClients: true,
 			},
+			provisioningv1.SchemeGroupVersion.Group: {
```
w13915984028, line 183: Is this generated client used only for test code?

w13915984028:
BTW: The `imageCleanupPlanCompletedAnnotation` is another solution, which keeps the original value for reference, in addtion, add a new annotation to mark the sync/action has been done, avoid running again.

pkg/codegen/main.go
wheatdog: Yes, this is only used for fake cluster client. Is there other preferred way to do so?

wheatdog:
> BTW: The `imageCleanupPlanCompletedAnnotation` is another solution, which keeps the original value for reference, in addtion, add a new annotation to mark the sync/action has been done, avoid running again.

Good point, preserving annotations might be better for troubleshooting. I can try that solution.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -754,7 +764,10 @@ func (h *upgradeHandler) loadReplicaReplenishmentFromUpgradeAnnotation(upgrade *
 	if err != nil {
 		return err
 	}
-	return h.setReplicaReplenishmentValue(value)
+	if err := h.setReplicaReplenishmentValue(value); err != nil {
+		return err
+	}
+	return nil
```
FrankYang0529: Is this different from `return h.setReplicaReplenishmentValue(value)`?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -798,7 +811,10 @@ func (h *upgradeHandler) loadAutoCleanupSystemGeneratedSnapshotFromUpgradeAnnota
 		logrus.Warnf("no original %s value set", autoCleanupSystemGeneratedSnapshotSetting)
 		return nil
 	}
-	return h.setAutoCleanupSystemGeneratedSnapshotValue(value)
+	if err := h.setAutoCleanupSystemGeneratedSnapshotValue(value); err != nil {
+		return err
+	}
+	return nil
```
FrankYang0529: ditto

FrankYang0529 (commented):
Over all LGTM. Leave some minor comments.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -461,55 +462,64 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
 			}
 			jobs, err := h.jobCache.List(plan.Namespace, set.AsSelector())
 			if err != nil {
-				return err
+				return nil, err
```
w13915984028, line 465: not specifically related to this PR,

but it deserves to skip the `IsNotFound` error when delete an object (list does not return such error);  

the deletion of an object in k8s takes time, and the controller might run a tight loop, repeatly delete an object makes little sense,

it is better to check if an object has `deletionTimestamp` before file a deletion operation; and skip `IsNotFound`  return error, it will make controller more efficiency

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -461,55 +462,64 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
...
 			}
 			for _, job := range jobs {
 				logrus.Debugf("Deleting job %s/%s", job.Namespace, job.Name)
 				if err := h.jobClient.Delete(job.Namespace, job.Name, &metav1.DeleteOptions{}); err != nil {
-					return err
+					return nil, err
```
w13915984028: it deserves to skip the `IsNotFound` error

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -461,55 +462,64 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
...
 				}
 			}
 		}
 
 		logrus.Debugf("Deleting plan %s/%s", plan.Namespace, plan.Name)
 		if err := h.planClient.Delete(plan.Namespace, plan.Name, &metav1.DeleteOptions{}); err != nil {
-			return err
+			return nil, err
```
w13915984028: it deserves to skip the `IsNotFound` error

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -461,55 +462,64 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) e
...
-			return err
+		_, exists := upgrade.Annotations[longhornSettingsRestoredAnnotation]
+		if !exists {
+			if err := h.loadReplicaReplenishmentFromUpgradeAnnotation(upgrade); err != nil {
+				return nil, err
+			}
+			if err := h.loadAutoCleanupSystemGeneratedSnapshotFromUpgradeAnnotation(upgrade); err != nil {
+				return nil, err
+			}
+			toUpdate.Annotations[longhornSettingsRestoredAnnotation] = strconv.FormatBool(true)
```
w13915984028, line 497: the `toUpdate` will be write back on L522, but there are other possible failures on L502, L512, L519,  in those cases, LH setting is amended, but upgrade CR itself does not book it;  in next retry, it will try to write to LH again if LH's value is different with current (e.g., set by user)

two options:
(1) Write to upgrade CR on L494, and let next OnChange to run following upgradeLog, managedchart
(2) Change L502, L512, L519, which will write back anyway even when error happens, but sill return the err, to let reconciller retry

w13915984028 (commented):
A few additional questions, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
wheatdog: Yes, they are redundant. Fixed. Thanks for pointing them out!

pkg/controller/master/upgrade/upgrade_controller.go
wheatdog: Fixed.

pkg/controller/master/upgrade/upgrade_controller.go
wheatdog: Thanks @w13915984028 for pointing out this potential issue. I prefer the first option you proposed. For the behavior of wrangler reconciller, is there any material I can refer to? Or I should just dive into the code of wrangler directly.

pkg/controller/master/upgrade/upgrade_controller.go
wheatdog: Addressed. Check `deletionTimestamp` and `IsNotFound` error before deleting.

pkg/controller/master/upgrade/upgrade_controller.go
wheatdog: Addressed. Check `deletionTimestamp` and `IsNotFound` error before deleting.

pkg/controller/master/upgrade/upgrade_controller.go
wheatdog: I fixed the the suggestion below of `XXXClient.Delete` -> Check deletionTimestamp and IsNotFound error before deleting.

For this line though, would we get `IsNotFound` error while listing? I think we don't need to handle it if not.
> 			jobs, err := h.jobCache.List(plan.Namespace, set.AsSelector())

pkg/controller/master/upgrade/upgrade_controller.go
wheatdog: fixed by https://github.com/harvester/harvester/pull/8941/commits/37329dc4cb2d017a40089b4a29da85415607b2d1

w13915984028 (approved):

wheatdog:
Thanks @w13915984028! I'll build the image for this enhanced version and test migration locally before merging.

Issue #8772, opened 2025-07-29
[ENHANCEMENT] Add webhook to prevent editing volume StorageClass in VM YAML


houhoucoop:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)
SURE-10457

On the VM page, the `Edit Config` UI prevents users from changing the StorageClass of an attached volume, but users can still successfully modify it through `Edit YAML`.

Edit Config
<img width="2992" height="1626" alt="Image" src="https://github.com/user-attachments/assets/4201aa28-142a-4a00-adfd-7b53c5c1e77e" />

Edit YAML
<img width="2992" height="1628" alt="Image" src="https://github.com/user-attachments/assets/4acc7a3c-8982-459f-897c-a3af442bf48e" />

### Describe the solution you'd like

Add webhook to disallow editing volume StorageClass via YAML to align with UI behavior.

### Describe alternatives you've considered

N/A

### Additional context

N/A

m-ildefons:
## Setup

<details>
<summary>Install Harvester v1.7.0-dev-20250831</summary>
<img width="1103" height="379" alt="Image" src="https://github.com/user-attachments/assets/240e2695-aad4-43f5-bc2c-2b07653421d5" />
</details>

<details>
<summary>Upload VM image and create test VM with extra volume</summary>
<img width="3834" height="1998" alt="Image" src="https://github.com/user-attachments/assets/d740f9e7-c3d6-438c-bb4e-e2aca77d5803" />

</details>

## 🟢 Test 1 -- Can't edit volume storage class from UI dialogue

<details>
<summary>Volume's storage class selector menu is inactive and greyed out</summary>
<img width="3444" height="1801" alt="Image" src="https://github.com/user-attachments/assets/b6650cae-5726-49e6-9474-22829d46785c" />
</details>

## 🟢 Test 2 -- Can't edit volume storage class from YAML editor

<details>
<summary>Changing YAML manifest</summary>
<img width="3432" height="1891" alt="Image" src="https://github.com/user-attachments/assets/92688c06-8677-4697-9534-5dd886a1981f" />
</details>

<details>
<summary>Error message shown when applying changes</summary>
<img width="2399" height="259" alt="Image" src="https://github.com/user-attachments/assets/d4cd23b9-121e-4d4b-afa3-b115e64992ca" />
</details>

## Conclusion

🟢 Validation passed

PR #8944, opened 2025-08-21
Add webhook which reject storage class name change

martindekov:
Adding webhook which rejects storage class name changes as part of the `harvesterhci.io/volumeClaimTemplates` annotation.

Creating, deleting and adding to the annotaion is allowed, but changing the storage class name itself is rejected. Also setting empty storage class name is allowed as well - following the omitempty flag of the field in the object.

#### Problem:
Changes to the volume's storage class is rejected through UI, but allowed through API - direct edits of the raw yaml.

#### Solution:
Introduce webhook which will detect such changes to already existing attached volumes and reject changes to the storage class name.

#### Related Issue(s):
Issue #8772

#### Test plan:
1. Build iso with the change
2. Create local harvester with the iso (single node)
3. Create storage class - `default-sc` with replication 1 and set it as default storage class
4. Create VM
5. Create volume and attach it as additional volume to the VM
6. Try to edit the original attached volume through kubectl and UI the value of key `storageClassName` - expect rejection (need to scroll a little):
```
      [{"metadata":{"name":"test-disk-0-kgsuo","creationTimestamp":null,"annotations":{"harvesterhci.io/imageId":"default/image-wd7kd"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"storageClassName":"default-sc","volumeMode":"Block"},"status":{}},{"metadata":{"name":"testtest","creationTimestamp":null},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"storageClassName":"default-sc","volumeMode":"Block"},"status":{}}]
                                                                                                                                                                                                                                                                                                                                                                                                                                              ### from this here ###########
```
             
to (need to scroll a little):
```
      [{"metadata":{"name":"test-disk-0-kgsuo","creationTimestamp":null,"annotations":{"harvesterhci.io/imageId":"default/image-wd7kd"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"storageClassName":"default-sc","volumeMode":"Block"},"status":{}},{"metadata":{"name":"testtest","creationTimestamp":null},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"storageClassName":"longhorn","volumeMode":"Block"},"status":{}}]
                                                                                                                                                                                                                                                                                                                                                                                                                                              ### to this here ###########
```
7. Try to edit the additional attached volume as above

Initially logic was not triggered because of [this check](https://github.com/harvester/harvester/blob/abce991f78d6c4020cf2dff6f72c660d7124e098/pkg/webhook/resources/virtualmachine/validator.go#L268) always returning nil:
```
	if oldVM.Spec.Template != nil && newVM.Spec.Template != nil && reflect.DeepEqual(oldVM.Spec.Template.Spec.Domain.Devices.Interfaces, newVM.Spec.Template.Spec.Domain.Devices.Interfaces) {
		return nil
	}
```
Added a comment above it so all new validation should come before the comment and the check.

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -265,6 +265,12 @@ func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj run
 		return err
 	}
 
+	if err := v.storageClassOfVolumeChanged(oldVM, newVM); err != nil {
+		return err
+	}
+
+	// This logic will return in case interfaces are existing and not changed
+	// make sure new validation logic is added above this comment
 	if oldVM.Spec.Template != nil && newVM.Spec.Template != nil && reflect.DeepEqual(oldVM.Spec.Template.Spec.Domain.Devices.Interfaces, newVM.Spec.Template.Spec.Domain.Devices.Interfaces) {
 		return nil
 	}
@@ -529,4 +535,41 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
 
 func (v *vmValidator) checkStorageResourceQuota(vm *kubevirtv1.VirtualMachine, oldVM *kubevirtv1.VirtualMachine) error {
 	return v.rqCalculator.CheckStorageResourceQuota(vm, oldVM)
+}
+
+func (v *vmValidator) storageClassOfVolumeChanged(oldVM, newVM *kubevirtv1.VirtualMachine) error {
+	oldVMvolumeClaimTemplates := oldVM.Annotations[util.AnnotationVolumeClaimTemplates]
+	newVMvolumeClaimTemplates := newVM.Annotations[util.AnnotationVolumeClaimTemplates]
+
+	if oldVMvolumeClaimTemplates == "" ||
+		newVMvolumeClaimTemplates == "" ||
+		oldVMvolumeClaimTemplates == newVMvolumeClaimTemplates {
+		return nil
+	}
+
+	var oldVMpvcs []*corev1.PersistentVolumeClaim
+	if err := json.Unmarshal([]byte(oldVMvolumeClaimTemplates), &oldVMpvcs); err != nil {
+		return fmt.Errorf("invalid %v annotation, error: %w", util.AnnotationVolumeClaimTemplates, err)
+	}
+	var newVMpvcs []*corev1.PersistentVolumeClaim
+	if err := json.Unmarshal([]byte(newVMvolumeClaimTemplates), &newVMpvcs); err != nil {
+		return fmt.Errorf("invalid %v annotation, error: %w", util.AnnotationVolumeClaimTemplates, err)
+	}
+
+	for _, oldVMpvc := range oldVMpvcs {
+		for _, newVMpvc := range newVMpvcs {
+			if oldVMpvc.Name == newVMpvc.Name {
+				if *newVMpvc.Spec.StorageClassName != *oldVMpvc.Spec.StorageClassName {
+					return fmt.Errorf("storage class %v of volume %v in annotation %v cannot be changed to %v",
+						*oldVMpvc.Spec.StorageClassName,
+						oldVMpvc.Name,
+						util.AnnotationVolumeClaimTemplates,
+						*newVMpvc.Spec.StorageClassName)
+				}
+				continue
+			}
+		}
+	}
+
+	return nil
 }
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -260,12 +260,8 @@ func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj run
 		}
 	}
 
-	// Check resize volumes
-	if err := v.checkResizeVolumes(oldVM, newVM); err != nil {
-		return err
-	}
-
-	if err := v.storageClassOfVolumeChanged(oldVM, newVM); err != nil {
+	// Check volume annotations
+	if err := v.checkVolumeAnnotations(oldVM, newVM); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


w13915984028 (dismissed):
LGTM, thanks for the fix.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -530,3 +536,40 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
 func (v *vmValidator) checkStorageResourceQuota(vm *kubevirtv1.VirtualMachine, oldVM *kubevirtv1.VirtualMachine) error {
 	return v.rqCalculator.CheckStorageResourceQuota(vm, oldVM)
 }
+
+func (v *vmValidator) storageClassOfVolumeChanged(oldVM, newVM *kubevirtv1.VirtualMachine) error {
```
ihcsim: This functions looks almost identical to `checkResizeVolumes()`. Consider reconciling them by:

1. Moving the check criteria from line 562 to 568 to that function
2. Rename that function from `checkResizeVolumes()` to `checkAnnotationsUpdate()` or something

I think your test cases will still pass with the change.

martindekov: Thanks for the input Ivan, good catch! Did not realize that already existing method is doing most of the heavy lifting of the new one.

Renamed the method to: `checkVolumeAnnotations` to better reflect that we still work on the volume annotations inside. And yes the testing coverts the new method:
image

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -407,6 +409,21 @@ func (v *vmValidator) checkResizeVolumes(oldVM, newVM *kubevirtv1.VirtualMachine
 		}
 	}
 
+	for _, oldVMpvc := range oldPvcs {
+		for _, newVMpvc := range newPvcs {
+			if oldVMpvc.Name == newVMpvc.Name {
+				if *newVMpvc.Spec.StorageClassName != *oldVMpvc.Spec.StorageClassName {
+					return fmt.Errorf("storage class %v of volume %v in annotation %v cannot be changed to %v",
```
olblak: You exist after the first error, wouldn't it be better to use an array of error, to detect all potential issues and to return an error if your array of error is greater than 0?

martindekov: This is very good suggestion, the only downside I can think of is if you have multiple changes to storageclassname webhook error might get a big message. Otherwise I agree that it can be a better UX to get all errors in 1 shot as opposed to:
1. Change a,b,c storage class names
2. Apply; Get rejected on `a`; Fix `a`
3. Apply; Get rejected on `b`; Fix `b`
...

wdyt @ihcsim @w13915984028 do we have some guideline on what the webhook size/content should be?

If not I can go ahead and refactor

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -407,6 +409,21 @@ func (v *vmValidator) checkResizeVolumes(oldVM, newVM *kubevirtv1.VirtualMachine
 		}
 	}
 
+	for _, oldVMpvc := range oldPvcs {
+		for _, newVMpvc := range newPvcs {
+			if oldVMpvc.Name == newVMpvc.Name {
+				if *newVMpvc.Spec.StorageClassName != *oldVMpvc.Spec.StorageClassName {
```
martindekov: A question regarding nil pointers as there are some semantics on how we handle them, I don't check them here and this is a problem which I will fix in incoming commit. I see 3 approaches:
1. Either reject nil pointer storage class name - enforcing storage class name always present even though omitempty stated in json struct
2. Allow nil pointer on both new and old, hackable to allow transitions, oldname -> nil -> newName -> nil -> newnewname -> so on
3. Allow nil pointer on creation, then on change from nil (oldpvc) to value (newpvc), but reject from value (oldpvc) to nil (newpvc) 

WDYT would be the best approach?

pkg/webhook/resources/virtualmachine/validator.go
w13915984028: ATM, most Harvester code returns from the first detected error. Return all errors is an alternative soltution. It is hard to say which one is even better.

martindekov: Ok, will stick with the standard for now, and will leave the discussion open if someone insists on certain format

pkg/webhook/resources/virtualmachine/validator.go
martindekov: As per offline discussion with Jian, will go with 2 to avoid breaking and for simplicity, but keep in mind the nil hop

pkg/webhook/resources/virtualmachine/validator.go
olblak: You can have a standard error message, but having a way to identify all the potential errors at one time provides a better UX IMHO

pkg/webhook/resources/virtualmachine/validator.go
olblak: I am wondering if it's even possible to have a nil pointer at this time.

martindekov: My initial assumption was that we can't, but added a test and validated that we actually can and that was the reason the integration test was failing with the EOF, Jian pointed out there is no `StorageClassName` given to`builder.PersistentVolumeClaimOption` here:
https://github.com/harvester/harvester/blob/abce991f78d6c4020cf2dff6f72c660d7124e098/tests/integration/api/vm_apis_test.go#L120

And on restart vm - changes are applied, logic is triggered and failure appears. So far on local integration test run this part of the test passed, unlike before. When whole suit passes I will push a commit with the fix.

pkg/webhook/resources/virtualmachine/validator.go
martindekov: ok, I can try building a string so final result can be shorter:
```
storage class names for the volumes in the harvesterhci.io/volumeClaimTemplates annotation cannot be changed; name %s in volume %s;name %s in volume %s; name %s in volume %s;...
```
I think we that format can hit both all issues and smaller message, will update PR shortly

martindekov: The error message now looks like:
```
storage class names for the volumes in the harvesterhci.io/volumeClaimTemplates annotation cannot be changed; new name longhorn-replicated in volume test-disk-0; new name longhorn-new in volume test-disk-1;
```
We can discuss further fixes if necessary a follow up discussion.

pkg/webhook/resources/virtualmachine/validator.go
martindekov: Resolving as latest commit contains fix 2

ihcsim (approved):

olblak (approved):

PR #8964, opened 2025-08-26
[Enhancement] Better handling of failed maintenance mode

votdev:
#### Problem:
- Annotations are not cleaned up correctly in case of an error.
- There is a useless if query which can be removed because it is candled some lines above.

#### Solution:
- Remove annotations in case draining the node is not possible.
- Remove `if` query and correctly indent code.
- Additionally several improvements are applied with this PR as well.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8985

##### Case 1 ####
- Setup a 3-node cluster.
- Set node `harvester-node-2` into maintenance mode.
- Use `Edit YAML` action menu on node `harvester-node-1`.
- Apply `harvesterhci.io/drain-requested: 'true'` in `metadata.annotations` and press `Save`.
- The node should not go into maintenance mode.
- In the logs of the `harvester` pod the following line should appear:
```
time="2025-09-02T10:04:14Z" level=info msg="Attempting to place node in maintenance mode" forced=false node_name=harvester-node-1
time="2025-09-02T10:04:14Z" level=warning msg="Draining is not possible: another controlplane is already in maintenance mode, cannot place current node in maintenance mode" node_name=harvester-node-1
```
- Use `Edit YAML` action menu on node `harvester-node-1` again.
- Make sure there is no `harvesterhci.io/drain-requested` or `harvesterhci.io/maintain-status` annotation.

n/a

### DIFF
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -2,6 +2,7 @@ package nodedrain
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"strings"
@@ -10,6 +11,7 @@ import (
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/rest"
@@ -26,10 +28,8 @@ import (
 )
 
 const (
-	nodeDrainController  = "node-drain-controller"
-	defaultWorkloadType  = "VirtualMachineInstance"
-	defaultSingleCPCount = 1
-	defaultHACPCount     = 3
+	nodeDrainController = "node-drain-controller"
+	defaultWorkloadType = "VirtualMachineInstance"
 )
 
 // ControllerHandler to drain nodes.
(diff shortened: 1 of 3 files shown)


votdev:
For easier code review simply click the `Hide whitespace changes` in the files view.

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -77,121 +77,120 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
-	if ok {
-		// still running a check in the background to avoid maintenance issues when using object annotations
-		// directly
-		err := drainhelper.DrainPossible(ndc.nodeCache, node)
-		if err != nil {
-			return node, err
-		}
+	// still running a check in the background to avoid maintenance issues when using object annotations
+	// directly
+	err := drainhelper.DrainPossible(ndc.nodeCache, node)
```
w13915984028, line 95: It is not directly related to this PR, but `DrainPossible` seems to have issues:
(1) If it failed to list things, then return error is okay for retry
(2) if it failed due to error like `errSingleControlPlaneNode` and `errHAControlPlaneNode`, the whole drain (or the maintenance mode) should be put to failure instead of return error to reconciller
 
 `DrainPossible` is more like `IsDrainPossible`,  return `bool, err` will not lead such

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -77,121 +77,120 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
-				}
+			// Append the VM to the list of VMs that need to be shut down.
+			maintainModeStrategyVMs = append(maintainModeStrategyVMs, fmt.Sprintf("%s/%s", vmi.Namespace, vmName))
+
+			// Skip and do not annotate VMs that do not have to be restarted
+			// at several stages of the maintenance mode. These are VMs with
+			// the label values:
+			// - Shutdown
+			if !slices.Contains([]string{
+				util.MaintainModeStrategyShutdownAndRestartAfterEnable,
```
w13915984028: could this tmp slice be put outside of the for loop for reuse?

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -77,121 +77,120 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
-		}
-
-		// Shutdown ALL VMs on that node forcibly? This is activated by a
-		// checkbox in the maintenance mode dialog in the UI.
-		if forced {
-			shutdownVMs, err = ndc.FindNonMigratableVMS(node)
+			vm, err := ndc.virtualMachineCache.Get(vmi.Namespace, vmName)
 			if err != nil {
-				return node, fmt.Errorf("error listing VMIs in scope for shutdown: %v", err)
+				return node, fmt.Errorf("error looking up VM %s/%s: %w", vmi.Namespace, vmName, err)
```
w13915984028, line 169: should `IsNotFound` be skipped? in case the vm is being deleted

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -77,121 +77,120 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
 			}
-		}
 
-		for _, v := range getUniqueVMSfromConditionMap(shutdownVMs) {
-			// Fetch VMI again in case it has been modified.
-			err := ndc.findAndStopVM(v)
+			vmCopy := vm.DeepCopy()
+			vmCopy.Annotations[util.AnnotationMaintainModeStrategyNodeName] = node.Name
```
w13915984028: check annotation is there, to avoid duplicated updating

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -77,121 +77,120 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
+			return node, fmt.Errorf("error listing VMIs in scope for shutdown: %v", err)
 		}
+	}
 
-		// run node drain
-		nodeCopy := node.DeepCopy()
-		err = drainhelper.DrainNode(ndc.context, ndc.restConfig, nodeCopy)
+	for _, v := range getUniqueVMSfromConditionMap(shutdownVMs) {
+		// Fetch VMI again in case it has been modified.
+		err := ndc.findAndStopVM(v)
```
w13915984028, line 197: findAndStopVM seems should also skip IsNotFound

w13915984028 (commented):
some suggestions, thanks.

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: IMO this does not make sense because it will only relocate the error to the next call of [findAndStopVM](https://github.com/harvester/harvester/blob/89798197afc87b2231f922c9c6397098a21e0fda/pkg/controller/master/nodedrain/nodedrain_controller.go#L203) which is done some lines below.

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: When it comes to conserving resources, I don't think we really gain anything by first checking whether the key already exists. If the key already exists, we should still overwrite it because, for whatever reason, it contains something incorrect and we need to ensure that it contains what we expect.

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: Hmmm, in this case https://github.com/harvester/harvester/pull/8964#discussion_r2303173713 is obsolete.

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: Done

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: Done

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: Done

pkg/controller/master/nodedrain/nodedrain_controller.go
w13915984028: additional conext:

when `vmCopy.Annotations[util.AnnotationMaintainModeStrategyNodeName] == node.Name`, the next update could be skipped.  or don't copy vm when `vm.Annotations[util.AnnotationMaintainModeStrategyNodeName] == node.Name`

the OnChange of one node might happen multi times due to the errors from here or there/ or reconciled by other changes on the node, update vm annoation might have been done in previous try, and even though the non-change update to kube-apiserver does not harm, but the cross-node RPC call is still time-sensitive

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: Done

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: Oh right, i didn't realize the `ndc.virtualMachineClient.Update(vmCopy)` which is not a cheap action. So your suggestion will make absolute sense.

votdev: Done

votdev:
Any thoughts about backporting this to 6.x and maybe 5.x as well?

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -77,121 +78,134 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
-		// directly
-		err := drainhelper.DrainPossible(ndc.nodeCache, node)
-		if err != nil {
-			return node, err
+	// still running a check in the background to avoid maintenance issues when using object annotations
+	// directly
+	drainPossible, err := drainhelper.DrainPossible(ndc.nodeCache, node)
+	if err != nil {
+		if !drainPossible {
+			// Fail hard, cannot drain this node.
```
w13915984028: when error is returned, the controller will retry

need a way to express the `failed to drain` and stop;  e.g. remove the `DrainAnnotation` and put `DrainFailedAnnotation`

but it might belong to another refactor PR

w13915984028 (dismissed):
LGTM, thanks.

w13915984028:
> Any thoughts about backporting this to 6.x and maybe 5.x as well?

As per current rule, it is not necessary to backport to v1.5.2 / v1.6.1,  sticks to master-head (v1.7.0).

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: I've opened https://github.com/harvester/harvester/issues/8985 for that.

votdev: Will be fixed in this PR as well.

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -91,6 +91,16 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
 	drainPossible, err := drainhelper.DrainPossible(ndc.nodeCache, node)
 	if err != nil {
 		if !drainPossible {
+			// Cleanup annotations to avoid recurring approaches to drain the node.
+			nodeCopy := node.DeepCopy()
+			delete(nodeCopy.Annotations, ctlnode.MaintainStatusAnnotationKey)
+			delete(nodeCopy.Annotations, drainhelper.DrainAnnotation)
+			delete(nodeCopy.Annotations, drainhelper.ForcedDrain)
+
```
w13915984028, line 108: could add an info/warn message like "drain is impossible, remove related annotations" for better debug

w13915984028 (dismissed):
LGTM, thanks.

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: Done

w13915984028 (dismissed):
LGTM, thanks.

w13915984028 (dismissed):
LGTM, thanks.

w13915984028 (dismissed):
LGTM, thanks

w13915984028 (dismissed):
LGTM, thanks.

pkg/util/drainhelper/helper.go
```diff
@@ -76,40 +76,46 @@ func DrainNode(ctx context.Context, cfg *rest.Config, node *corev1.Node) error {
...
-// to identify if it is possible to place the current mode in maintenance mode
-func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) error {
+// DrainPossible is a helper method to check a node object and query remaining
+// nodes in the cluster to identify if it is possible to place the current mode
+// in maintenance mode.
+// Returns true if it is possible to drain the node, false if not possible. If
+// true and an error are returned, then the drain operation can be reconciled
+// again since the error does not rule out the possibility of a drain. If false
+// and an error are returned, then the conditions for a drain are not met and
+// reconciling does not make sense.
```
ihcsim, line 87: Instead of having to reason about a bool/error returned tuple, maybe consider returning a wrapped "irrecoverable" error, to make it clear to the caller whether to retry the drain operation. E.g.,

```diff
diff --git a/pkg/controller/master/nodedrain/nodedrain_controller.go b/pkg/controller/master/nodedrain/nodedrain_controller.go
index 77f40dff7..63787fd14 100644
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -2,6 +2,7 @@ package nodedrain
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"strings"
@@ -91,9 +92,8 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
 
 	// still running a check in the background to avoid maintenance issues when using object annotations
 	// directly
-	drainPossible, err := drainhelper.DrainPossible(ndc.nodeCache, node)
-	if err != nil {
-		if !drainPossible {
+	if err := drainhelper.DrainPossible(ndc.nodeCache, node); err != nil {
+		if errors.Is(err, drainhelper.ErrNodeDrainImpossible) {
 			logrus.WithFields(logrus.Fields{
 				"node_name": node.Name,
 			}).Warnf("Draining is not possible: %v", err)
diff --git a/pkg/util/drainhelper/helper.go b/pkg/util/drainhelper/helper.go
index d1f03383e..326528103 100644
--- a/pkg/util/drainhelper/helper.go
+++ b/pkg/util/drainhelper/helper.go
@@ -37,8 +37,9 @@ const (
 )
 
 var (
-	errSingleControlPlaneNode = errors.New("single controlplane cluster, cannot place controlplane in maintenance mode")
-	errHAControlPlaneNode     = errors.New("another controlplane is already in maintenance mode, cannot place current node in maintenance mode")
+	ErrNodeDrainImpossible    = errors.New("draining of this node is not possible")
+	errSingleControlPlaneNode = fmt.Errorf("%w. single controlplane cluster, cannot place controlplane in maintenance mode", errNodeDrainImpossible)
+	errHAControlPlaneNode     = fmt.Errorf("%w. another controlplane is already in maintenance mode, cannot place current node in maintenance mode", errNodeDrainImpossible)
 )
 
 func defaultDrainHelper(ctx context.Context, cfg *rest.Config) (*drain.Helper, error) {
@@ -84,17 +85,17 @@ func DrainNode(ctx context.Context, cfg *rest.Config, node *corev1.Node) error {
 // again since the error does not rule out the possibility of a drain. If false
 // and an error are returned, then the conditions for a drain are not met and
 // reconciling does not make sense.
-func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) (bool, error) {
+func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) error {
 	_, cpLabelOK := node.Labels["node-role.kubernetes.io/control-plane"]
 	_, etcdLabelOK := node.Labels["node-role.kubernetes.io/etcd"]
 
 	if !cpLabelOK && !etcdLabelOK { // not a controlplane node. no further action needed
-		return true, nil
+		return nil
 	}
 
 	cpReq, err := labels.NewRequirement("node-role.kubernetes.io/control-plane", selection.Exists, nil)
 	if err != nil {
-		return true, fmt.Errorf("error creating requirement: %v", err)
+		return fmt.Errorf("error creating requirement: %v", err)
 	}
 
 	cpSelector := labels.NewSelector()
@@ -102,12 +103,12 @@ func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) (bool, erro
 
 	cpNodeList, err := nodeCache.List(cpSelector)
 	if err != nil {
-		return true, fmt.Errorf("error listing nodes matching selector %s: %v", cpSelector.String(), err)
+		return fmt.Errorf("error listing nodes matching selector %s: %v", cpSelector.String(), err)
 	}
 
 	etcdReq, err := labels.NewRequirement("node-role.kubernetes.io/etcd", selection.Exists, nil)
 	if err != nil {
-		return true, fmt.Errorf("error creating requirement: %v", err)
+		return fmt.Errorf("error creating requirement: %v", err)
 	}
 
 	etcdSelector := labels.NewSelector()
@@ -115,7 +116,7 @@ func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) (bool, erro
 
 	etcdNodeList, err := nodeCache.List(etcdSelector)
 	if err != nil {
-		return true, fmt.Errorf("error listing nodes matching selector %s: %v", etcdSelector.String(), err)
+		return fmt.Errorf("error listing nodes matching selector %s: %v", etcdSelector.String(), err)
 	}
 
 	// add items to map since controlplane nodes will have etcd labels
@@ -130,7 +131,7 @@ func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) (bool, erro
 	}
 	// only controlplane node which we are trying to place into maintenance
 	if len(nodeMap) == defaultSingleCPCount {
-		return false, errSingleControlPlaneNode
+		return errSingleControlPlaneNode
 	}
 
 	var availableNodes int
@@ -143,10 +144,10 @@ func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) (bool, erro
 	}
 
 	if availableNodes != defaultHACPCount {
-		return false, errHAControlPlaneNode
+		return errHAControlPlaneNode
 	}
 
-	return true, nil
+	return nil
 }
 
 func maintainModeStrategyFilter(pod corev1.Pod) drain.PodDeleteStatus {
```

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -77,121 +76,152 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
+	if err != nil {
+		if !drainPossible {
+			logrus.WithFields(logrus.Fields{
+				"node_name": node.Name,
+			}).Warnf("Draining is not possible: %v", err)
+
+			nodeCopy := node.DeepCopy()
+
+			// Cleanup annotations to avoid recurring approaches to drain the node.
+			delete(nodeCopy.Annotations, ctlnode.MaintainStatusAnnotationKey)
```
ihcsim, line 105: do we need to indicate to the maintenance controller that node maintenance is not possible, and disable maintenance mode? e.g., resetting `node.Spec.Unschedulable` back to `false`, similar to [what the formatter is doing](https://github.com/harvester/harvester/blob/8841ea4ff79c3940a0ee28cc16700064e5eea155/pkg/api/node/formatter.go#L234-L245)?

votdev: `node.Spec.Unschedulable` is not set at that point, it will be set at a [later date](https://github.com/harvester/harvester/blob/8841ea4ff79c3940a0ee28cc16700064e5eea155/pkg/controller/master/nodedrain/nodedrain_controller.go#L184).

pkg/util/drainhelper/helper.go
votdev: Good idea.

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -77,121 +77,152 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
+			delete(nodeCopy.Annotations, drainhelper.DrainAnnotation)
+			delete(nodeCopy.Annotations, drainhelper.ForcedDrain)
+
+			_, errUpdate := ndc.nodes.Update(nodeCopy)
+			if errUpdate != nil {
+				return node, fmt.Errorf("failed to update node to clean up annotations: %w", errUpdate)
```
```suggestion
				return node, errors.Join(fmt.Errorf("failed to update node to clean up annotations: %w", errUpdate), err)
```
I think we should include the `err` in this returned error so that the original reason for the drain failures is retained.

votdev: Right, but i suggest to use a different order of the errros: `return node, errors.Join(err, fmt.Errorf("failed to update node to clean up annotations: %w", errUpdate))`.
`err` is the origin error that triggered the whole thing, so it has to be the first mentioned in the concatenated error.

ihcsim (approved):

Issue #8962, opened 2025-08-26
[BUG] pod mutator check for label app=rancher which can block vm from starting


wheatdog:
### Describe the Bug

According to [this implementation](https://github.com/harvester/harvester/blob/6f5a48197f562e3fe85d8ddfb815ef2cc537dc59/pkg/webhook/resources/pod/mutator.go#L29), pods with label `app = rancher` would be patched with "additional-ca-volume". This label can be added to any VM as an instance label in any namespace not just cattle-system, harvester-system, and longhorn-system where the corresponding secret harvester-additional-ca get propagated to.

### To Reproduce

1. Go to `Advanced > Settings` set `additional-ca` ([reference](https://docs.harvesterhci.io/v1.5/advanced/index/#additional-ca))
2. Create a VM with one of the `Instance Labels` to be `app = rancher`

It would stuck at `Starting` phase.

<img width="1293" height="997" alt="Image" src="https://github.com/user-attachments/assets/6176dda7-a17e-445f-a7c9-338ed9b0b973" />

### Expected Behavior

VMs with label `app = rancher` should be created successfully.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.5.0, v1.6.0-rc6
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Additional context

<details>
<summary>kubectl get vm</summary>

```
❯ kubectl get vm vm1-with-app-rancher -o yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/default-userdata-secret: raw-image-template-userdata
    harvesterhci.io/mac-address: '{"default":"06:4a:ba:71:d5:5f"}'
    harvesterhci.io/vmRunStrategy: RerunOnFailure
    harvesterhci.io/volumeClaimTemplates: '[{"metadata":{"name":"vm1-with-app-rancher-rootdisk-uhkfo","annotations":{"harvesterhci.io/imageId":"default/image-fmfmr"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"volumeMode":"Block","storageClassName":"longhorn-image-fmfmr"}}]'
    kubevirt.io/latest-observed-api-version: v1
    kubevirt.io/storage-observed-api-version: v1
    network.harvesterhci.io/ips: '[]'
[...]

PR #8973, opened 2025-08-27
fix(webhook): pod mutator should also consider namespace

wheatdog:
#### Problem:

When a VM is created with the `app=rancher` instance label, it gets stuck indefinitely in the "Starting" phase. The root cause is that the corresponding pod fails to start due to a missing Kubernetes secret.

The pod's event logs show the following error:
```
failed for volume "additional-ca-volume" : secret "harvester-additional-ca" not found
```

#### Solution:

Limit the mutator to handle only system pods by making the pod filter check the namespace. It now only affects pods that are in the right namespace and have the right labels.

#### Related Issue(s):

#8962 

#### Test plan:

1. Go to Advanced > Settings set additional-ca ([reference](https://docs.harvesterhci.io/v1.5/advanced/index/#additional-ca))
2. Create a VM with one of the Instance Labels to be app = rancher
3. It should be created successfully and eventually at "Running" phase

N/A

### DIFF
--- a/pkg/webhook/resources/pod/mutator.go
+++ b/pkg/webhook/resources/pod/mutator.go
@@ -17,16 +17,22 @@ import (
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
-var matchingLabels = []labels.Set{
-	{
-		"longhorn.io/component": "backing-image-data-source",
+var matchingLabelsForNamespace = map[string][]labels.Set {
+	"longhorn-system": {
+		{
+			"longhorn.io/component": "backing-image-data-source",
+		},
 	},
-	{
-		"app.kubernetes.io/name":      "harvester",
-		"app.kubernetes.io/component": "apiserver",
+	"harvester-system": {
+		{
+			"app.kubernetes.io/name":      "harvester",
+			"app.kubernetes.io/component": "apiserver",
+		},
 	},
-	{
-		"app": "rancher",
+	"cattle-system": {
+		{
+			"app": "rancher",
+		},
 	},
 }
 
@@ -62,16 +68,7 @@ func (m *podMutator) Resource() types.Resource {
 
 func (m *podMutator) Create(_ *types.Request, newObj runtime.Object) (types.PatchOps, error) {
 	pod := newObj.(*corev1.Pod)
-
-	podLabels := labels.Set(pod.Labels)
-	var match bool
-	for _, v := range matchingLabels {
-		if v.AsSelector().Matches(podLabels) {
-			match = true
-			break
-		}
-	}
-	if !match {
+	if !shouldPatch(pod) {
 		return nil, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/pod/mutator.go
```diff
@@ -17,16 +17,22 @@ import (
...
+			"app.kubernetes.io/name":      "harvester",
+			"app.kubernetes.io/component": "apiserver",
+		},
 	},
-	{
-		"app": "rancher",
+	"cattle-system": {
+		{
+			"app": "rancher",
+		},
```
ihcsim: Can we use the `const` namespaces defined in the `util` package? E.g., https://github.com/harvester/harvester/blob/90bac098aef17a8eea1fd7fc40341a85ae86538e/pkg/util/constants.go#L88

wheatdog: Fixed, thanks for the suggestion.

pkg/webhook/resources/pod/mutator.go
```diff
@@ -236,3 +233,18 @@ func volumeMountPatch(target []corev1.VolumeMount, path string, volumeMount core
...
+		if pod.Namespace != namespace {
+			continue
+		}
+		for _, v := range ls {
+			if v.AsSelector().Matches(podLabels) {
+				return true
+			}
+		}
+	}
+	return false
```
ihcsim: I think we should be able to access the map directly via its key, without needing two `for` loops. Will the following work?

```diff
diff --git a/pkg/webhook/resources/pod/mutator.go b/pkg/webhook/resources/pod/mutator.go
index d62e6104f..be0befbde 100644
--- a/pkg/webhook/resources/pod/mutator.go
+++ b/pkg/webhook/resources/pod/mutator.go
@@ -17,7 +17,7 @@ import (
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
-var matchingLabelsForNamespace = map[string][]labels.Set {
+var matchingLabelsForNamespace = map[string][]labels.Set{
 	util.LonghornSystemNamespaceName: {
 		{
 			"longhorn.io/component": "backing-image-data-source",
@@ -235,15 +235,15 @@ func volumeMountPatch(target []corev1.VolumeMount, path string, volumeMount core
 }
 
 func shouldPatch(pod *corev1.Pod) bool {
+	matchingLabels, exists := matchingLabelsForNamespace[pod.Namespace]
+	if !exists {
+		return false
+	}
+
 	podLabels := labels.Set(pod.Labels)
-	for namespace, ls := range matchingLabelsForNamespace {
-		if pod.Namespace != namespace {
-			continue
-		}
-		for _, v := range ls {
-			if v.AsSelector().Matches(podLabels) {
-				return true
-			}
+	for _, label := range matchingLabels {
+		if label.AsSelector().Matches(podLabels) {
+			return true
 		}
 	}
 	return false

```

wheatdog: You are right. That is cleaner and more efficient. Thanks!

Yu-Jack (approved):
LGTM, thanks!

FrankYang0529 (approved):

PR #9011, opened 2025-08-31
scripts: ignore minimum upgradable version with sprint release

Vicente-Cheng:
#### Problem:
ignore minimum upgradable version with sprint release

#### Solution:
ignore minimum upgradable version with sprint release


Vicente-Cheng:
@mergifyio backport v1.6

Issue #9029, opened 2025-09-03
[BUG] Should rollout kube-system/rke2-ingress-nginx-controller daemonset when reset ssl-certificates setting


FrankYang0529:
### Describe the Bug

When updating `ssl-certificates` from default to a new value, it updates `tls-ingress` secret and `kube-system/rke2-ingress-nginx` helmconfig, and restarts rancher deployment. However, when user changes `ssl-certificates` from a value to a new value, the update for `kube-system/rke2-ingress-nginx` helmconfig is no changed. The `kube-system/rke2-ingress-nginx-controller` doesn't restart, so the new `ssl-certificates` doesn't work.

### To Reproduce

1. Create a harvester cluster.
2. Set `ssl-certificates` settings.
3. Set `ssl-certificates` settings again.
4. The second change doesn't restart `kube-system/rke2-ingress-nginx-controller`.

### Expected Behavior

The harvester controller should restart `kube-system/rke2-ingress-nginx-controller`.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: master
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): KVM
- Rancher version:

Vicente-Cheng:
Hi @FrankYang0529, 
Since the backport PR was just merged, could you help update the status tomorrow?
Thanks!

Vicente-Cheng:
Hi @Yu-Jack, 
Could you help verify this issue? Thanks!

Yu-Jack:
### 🟢 Test Passed with v1.7.0-rc3

---

#### Step 1: nginx pod and ssl-certificate status

<details>
<summary>Click to expand Step 1 Screenshots</summary>

<img width="2475" height="915" alt="Image" src="https://github.com/user-attachments/assets/e9e363c0-233e-437f-a4c3-a857e633b951" />

</details>

---

#### Step 2: Generate and Attach Certificate (First Time)

#### Step 3: nginx controller Restart (First Time)

<details>
<summary>Click to expand Step 3 Screenshots</summary>

<img width="2473" height="942" alt="Image" src="https://github.com/user-attachments/assets/aed38dea-f863-4989-b92a-297062375c7f" />

</details>

---

#### Step 5: Generate and Re-attach Certificate (Second Time)

You can tell the cert difference between the first and second time.

#### Step 6: nginx controller Restart (Second Time)

<details>
<summary>Click to expand Step 6 Screenshots (2 images total)</summary>

[...]

PR #9034, opened 2025-09-03
fix: redeploy kube-system/rke2-ingress-nginx-controller daemonset

FrankYang0529:
#### Related Issue(s):

https://github.com/harvester/harvester/issues/9029

#### Test plan:

2. Set `ssl-certificates` settings.
3. Set `ssl-certificates` settings again.
4. The second change restarts `kube-system/rke2-ingress-nginx-controller`.

### DIFF
--- a/pkg/controller/master/setting/ssl_certificate.go
+++ b/pkg/controller/master/setting/ssl_certificate.go
@@ -37,8 +37,7 @@ func (h *Handler) resetCertificates() error {
 	if err := h.updateIngressDefaultCertificate(util.CattleSystemNamespaceName, util.InternalTLSSecretName); err != nil {
 		return err
 	}
-
-	return h.redeployDeployment(util.CattleSystemNamespaceName, rancherDeploymentName)
+	return h.redeploySSLCertificateWorkload()
 }
 
 func (h *Handler) updateCertificates(sslCertificate *settings.SSLCertificate) error {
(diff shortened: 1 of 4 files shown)


Vicente-Cheng (dismissed):
lgtm, thanks!

pkg/controller/master/setting/ssl_certificate.go
```diff
@@ -50,6 +50,10 @@ func (h *Handler) updateCertificates(sslCertificate *settings.SSLCertificate) er
 		return err
 	}
 
+	if err := h.redeployDaemonset(util.KubeSystemNamespace, util.Rke2IngressNginxControllerName); err != nil {
```
w13915984028: Is it also required on reset cert path? thx.


https://github.com/harvester/harvester/blob/0d07cda45dc903a9528407297b9a45a3fff9f31f/pkg/controller/master/setting/ssl_certificate.go#L36-L54

FrankYang0529: Thanks for the review. Updated it.

pkg/controller/master/setting/ssl_certificate.go
```diff
@@ -89,3 +87,11 @@ func (h *Handler) updateIngressDefaultCertificate(namespace, secretName string)
 	}
 	return nil
 }
+
+func (h *Handler) redeploySSLCertificateWorkload() error {
+	if err := h.redeployDaemonset(util.KubeSystemNamespace, util.Rke2IngressNginxControllerName); err != nil {
```
w13915984028, line 92: we could add a log about the redeployment

w13915984028 (approved):
LGTM, thanks.

ibrokethecloud (approved):
lgtm. thanks.

Issue #3418, opened 2023-02-07
[FEATURE] Change OS network tool from wicked to the NetworkManager


guangbochen:
**Is your enhancement related to a problem? Please describe.**
Consider changing the OS network tool from wicked to NetworkManager, this is because NetworkManager is the default network tool in SLE-Micro-for-Rancher 5.3 and may provide the following pros.
- https://documentation.suse.com/sle-micro/5.3/html/SLE-Micro-all/cha-nm-vs-wicked.html 

**Describe the solution you'd like**
Replace wicked with NetworkManager, and it should be compatible with upgrade changes.

**Describe alternatives you've considered**
N/A

**Additional context**
- Related to https://github.com/harvester/os2/pull/54 change

bk201:
@mingshuoqiu Please check if we need this change for vxlan.

mingshuoqiu:
Not really necessary. The VxLAN won't rely on NetworkManger

tserong:
Assigning to myself for now as I'm investigating as part of #7025

irishgordo:
@albinsun , net-install w/ 802.3AD was looked at here: https://github.com/harvester/harvester/issues/9393

albinsun:
> [@albinsun](https://github.com/albinsun) , net-install w/ 802.3AD was looked at here: [#9393](https://github.com/harvester/harvester/issues/9393)

Hi @irishgordo 
For different bond modes, if I understand it correctly it depends on the env. network setup (`balance-tlb` in the lab), any way we can test this?
<img width="406" height="313" alt="Image" src="https://github.com/user-attachments/assets/50fa850a-462b-4325-b5b5-f67575530810" />

cc. @tserong

tserong:
I'm not sure how to physically test this in the lab (sorry!) but one thing to check on the configuration side, is that the selected mode is correctly set in the `[bond]` section of the `/etc/NetworkManager/system-connections/bond-mgmt.nmconnection` file. For example, by default with `active-backup`, you should see something like this:

```
# cat /etc/NetworkManager/system-connections/bond-mgmt.nmconnection
[connection]
id=bond-mgmt
type=bond
interface-name=mgmt-bo
master=mgmt-br
slave-type=bridge

[ethernet]

[bond]
miimon=100
mode=active-backup

[bridge-port]
```

If you select a different mode, you should see a change in the `mode=` line above.

tserong:
[...]

PR #9039, opened 2025-09-03
HEP: Migrate from Wicked to NetworkManager [skip ci]

tserong:
#### Problem:
We need to migrate from Wicked to NetworkManager in order to be able to update our base OS to SL Micro 6.x.

#### Solution:

Described in this HEP

#### Related Issue(s):

https://github.com/harvester/harvester/issues/3418
https://github.com/harvester/harvester/issues/9581
#### Test plan:

Included in this HEP

### DIFF
--- a/enhancements/20250902-networkmanager.md
+++ b/enhancements/20250902-networkmanager.md
@@ -1,0 +1,137 @@
+# Migrate from Wicked to NetworkManager
+
+## Summary
+
+Harvester up to and including version v1.6.x is built on SLE Micro 5.x. While SLE Micro 5.x includes both Wicked and NetworkManager network managment stacks, Harvester has always used Wicked. With Harvester v1.7.0 we need to bump the base OS to SL Micro 6.x, which drops support for Wicked and uses NetworkManager exclusively. Consequently we need to teach Harvester how to configure NetworkManager instead of Wicked during installation, and we also need to make sure correct NetworkManager configuration is generated during upgrades.
+
+### Related Issues
+
+- https://github.com/harvester/harvester/issues/3418
+- https://github.com/harvester/harvester/issues/9300
+- https://github.com/harvester/harvester/issues/9301
+- https://github.com/harvester/harvester/issues/9581
+
+## Motivation
+
+### Goals
+
+- Generate Network Manager connection profiles instead of Wicked `ifcfg-*` files during initial installation.
+- Generate appropriate Network Manager connection profiles during upgrade from Harvester v1.6.x.
+- All currently supported networking options (DHCP, static IP, VLAN, bond options, etc.) must continue to operate correctly.
+
+### Non-goals
+
+- Actually replacing Harvester's base OS with SL Micro 6.x. This will be done as a separate work item for Harvester v1.7 once the NetworkManager configuration is implemented and known to be working correctly against SLE Micro 5.5.
+
+## Proposal
+
+Ideally, from the users's perspective, nothing really changes, provided the user isn't manually customizing the network configuration. Installation should still work as it does now, with the same UI and configuration options. The difference is that behind the scenes we'll have a different network management stack installed, and will be writing out NetworkManager connection profiles directly, instead of going through YAML files in `/oem/90_custom.yaml`.
+
+If the user does need to make manual changes to the network configuration, this can be done using the `nmcli` tool rather than editing YAML.
+
+### User Stories
+
+#### Story 1
+
+I'm installing Harvester v1.7.0 for the first time. I expect networking to operate correctly.
+
+#### Story 2
+
+I'm upgrading from Harvester v1.6.x to v1.7.0. I have _not_ made any post-installation changes to network configuration by manually editing `/oem/*.yaml` files. I expect networking to continue to operate correctly after the upgrade.
+
+#### Story 3
+
+I'm upgrading from Harvester v1.6.x to v1.7.0, but I _have_ made some post-installation changes to network configuration by manually editing `/oem/*.yaml` files. I expect networking to continue to operate correctly after the upgrade.
+
+#### Story 4
+
+I've made a mistake with some manual changes to network configuration, and want to revert back to the settings as they were at install time.
+
+### User Experience In Detail
+
+#### Initial Installation / Upgrades Without Post-installation Configuration Changes
+
+No special user action is required. Everything should just appear to work as it currently does.
+
+#### Upgrades Where Post-installation Configuration Changes Have Been Made
+
+Upgrade should proceed as it currently does, but there may be additional manual steps required depending on what post-installation configuration changes were made.
+
+For example, if bonding options have been changed as described in [the docs], there will have been manual edits made to the `/etc/sysconfig/network/ifcfg-mgmt-bo` file in `/oem/90_custom.yaml`. Those `ifcfg-*` file aren't used by NetworkManager, so the user will need to either:
+
+1. Before upgrade, make appropriate changes to `/oem/harvester.config`, so the upgrade process will generate the expected NetworkManager configuration automatically, or,
+2. After upgrade, use the `nmcli` tool to apply the desired configuration to the NetworkManager connection profiles that were generated during upgrade.
+
+[the docs]: https://docs.harvesterhci.io/v1.6/install/update-harvester-configuration/#configuration-persistence-2
+
+#### Reverting Manual Changes to NetworkManager Connection Profiles
+
+The user can run `harvester-installer generate-network-config` to re-generate NetworkManager connection profiles based on the contents of `/oem/harvester.config`.
+
+### API changes
+
+Add a new subcommand to the `harvester-installer` binary to generate NetworkManager connection profiles during upgrade.
+
+## Design
+
+### Implementation Overview
+
+#### Changes to [baseos](https://build.opensuse.org/package/show/isv:Rancher:Harvester:OS:Dev/baseos)
+
+- Don't install `wicked`
+- Don't remove `NetworkManager` (it's installed by default in SL Micro, but we're removing it in current Harvester releases)
+
+#### Changes to [os2](https://github.com/harvester/os2)
+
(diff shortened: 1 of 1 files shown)


enhancements/20250902-networkmanager.md
```diff
@@ -0,0 +1,116 @@
...
+  - Static IP
+  - DHCP
+  - VLAN
+- Install Harvester v1.6.x then upgrade to v1.7.0, and verify that the upgrade succeeds and networking continues to operate correctly.
+
+### Upgrade strategy
+
+With new installs, `/oem/90_custom.yaml` will include NetworkManager connection profiles instead of `ifcfg-*` files.
+
+With upgrades, the existing `/oem/90_custom.yaml` file will still include the old `ifcfg-*` files, which will be ignored by NetworkManager.
```
w13915984028: harvester.config might be outdated after several upgrades and user's potential changes; the general  upgrade practice is to generate the target `networkmanager.yaml` from the active `90_custom.yaml`

request user to manually change before/after upgrade is not acceptable in most cases

enhancements/20250902-networkmanager.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### Changes to [docs](https://github.com/harvester/docs)
+
+- Anywhere we describe how to tweak network configuration must be updated to talk about NetworkManager instead of wicked.
+- We need to add a special section about possible manual changes required during upgrade (see "Upgrade strategy" below).
+
+### Test plan
+
+- Install Harvester and verify that the various different ways of configuring networking all work:
+  - Static IP
```
w13915984028, line 113: mgmt MAC, bond options, MTU are also supported params atm.

w13915984028 (commented):
The HEP looks have covered the broad areas, thanks.

enhancements/20250902-networkmanager.md
tserong: I understand the concern here about the requesting manual changes, but there are some difficulties with generating NetworkManager config based on what's in `90_custom.yaml`.

What we need, to generate NetworkManager config, is essentially the data in `install.management_interface` from `harvester.config` (i.e. whether it's DHCP or static IP, the list of interfaces to be bonded, the bond options, the mtu, the VLAN ID). This is all explicitly stated in `harvester.config` and can be reliably read, at least since Harvester v1.1 (earlier in v1.0 and v0.3 the format was different, so actually if someone's got a system _that_ old, that would be another problem with the approach I'm suggesting here :-/) 

The data we need is not explicit or trivial to extract from `90_custom.yaml`. Rather, it's spread over a bunch of `ifcfg-*` files embedded in that YAML. So we would have to read `90_custom.yaml`, look for all the files directives, check those to see which ones were ifcfg files, parse them all and figure out what the settings are based on that. This strikes me as complicated and potentially error prone, especially if we have to handle every possible thing that someone could have added to an `ifcfg-*` file. If any of this were to break, the user would be forced to fix it manually.

Note that there's _not_ a 1:1 mapping between `ifcfg-*` files and NetworkManager connection profiles, and SLE doesn't include any tools to perform this migration.

In the happy case (Harvester was originally installed with v1.1 or newer, and no manual changes have been made to management interface configuration via `90_custom.yaml`), it's going to be much simpler to read `harvester.config` than trying to parse `90_custom.yaml`.

In the unhappy case (Harvester originally installed with v1.0 or older, or manual changes made to management interface configuration via `90_custom.yaml`), reading `harvester.config` may not give the correct configuration. But OTOH, trying to parse `90_custom.yaml` can't be guaranteed to give a reliable result either.

ibrokethecloud: we may need to document that users ensure harvester.config needs to match the existing network definition before the user initiates the upgrade. This would only be needed in case the users have made post install networking changes by editing the harvester-installer generated elemental cloudinit.

We also need to handle the schema change in harvester.config which was introduced in Harvester v1.1. This has been stable since.

However in harvester v1.0 our networking configuration looked like

```
install:
  mode: create
  networks:
    harvester-mgmt:
      interfaces:
      - name: ens5
        hwAddr: "B8:CA:3A:6A:64:7C"
      method: dhcp
```

compared to post v1.1 

```
install:
  mode: create
  management_interface:
    interfaces:
    - name: ens5
      hwAddr: "B8:CA:3A:6A:64:7C"
    method: dhcp
```

We need to handle this as part of the upgrade as we may have users running since Harvester v1.0 and the harvester.config has remained untouched.

A possible option would be to provide a utility/cli which leverages the cloudinit generation logic and can be run in advance on each node in the cluster. This would help users review the generated network configuration before the actual upgrade. This will allow them the opportunity to fine tune the harvester.config to ensure the generate network configuration is matching the current network setup.

irishgordo:
Just as a small mention on the Docs aspect :smile: , thinking:
- [Harvester "post-install" Runtime Change, updating DNS Servers](https://docs.harvesterhci.io/v1.6/install/update-harvester-configuration#runtime-change) would be great to get updates:
    - not sure if `netconfig update` would be a reality anymore or not :thinking: , guessing the DNS STATIC SERVERS list, won't be stashed in `/etc/sysconfig/network/config` anymore? 

Since:
>Story 2
>I'm upgrading from Harvester v1.6.x to v1.7.0. I have not made any post-installation changes to network configuration by manually editing /oem/*.yaml files. I expect networking to continue to operate correctly after the upgrade.

The user may also likely have persisted that change, in `/oem` so they likely wouldn't meet the qualifications of `Story 2` -> but Fresh Installs of Harvester Post-Install, a user may still want to change / update DNS

enhancements/20250902-networkmanager.md
```diff
@@ -0,0 +1,116 @@
...
+
+I'm upgrading from Harvester v1.6.x to v1.7.0, but I _have_ made some post-installation changes to network configuration by manually editing `/oem/*.yaml` files. I expect networking to continue to operate correctly after the upgrade.
+
+### User Experience In Detail
+
+#### Initial Installation / Upgrades Without Post-installation Configuration Changes
+
+No special user action is required. Everything should just appear to work as it currently does.
+
+#### Upgrades Where Post-installation Configuration Changes Have Been Made
```
ihcsim, line 56: I want to better understand the failure domains and possible scenarios. Basically, what is the worst thing that can happen and how do we recover? Presumably, if the switch over from wicked to nm failed, the problem will be very visible, right? The host loses connectivity, RKE2 reports the node status as `NotReady` then `Unknown`. Upgrade is halted. Likely, the admin can still access the host (via serial console?) to diagnose and repair.

If the host loses connectivity, does its role play a role (no pun intended) here? Will role promotion happen while the upgrade is happening? If yes, will the next promoted host be promoted to the correct version? Do we need to ask user to add another management host to ensure quorum in case of partition? Should upgrade be done on management hosts first? (I assume we are already doing this, but can't remember.)

Taking it a step further, is it possible for the (networking stack) upgrade to complete successfully, but user applications to be completely broken, say because some customization translation was amiss, they failed silently etc.? If yes, then what can user do to detect the problem mid-upgrade, before it takes down the entire user application layer? Is it possible to pause the upgrade?

tserong:
@irishgordo yeah, instead of tweaking `/etc/sysconfig/network/config`, it'll be something like running `nmcli con modify bridge-mgmt ipv4.dns 8.8.8.8,1.1.1.1 && nmcli device reapply mgmt-br` to apply the change live. To persist it, you'll be looking for that `nmcli con modify` command in `/oem/90_custom.yaml`, rather than the `sed -i 's/^NETCONFIG_DNS_STATIC_SERVERS` line.

enhancements/20250902-networkmanager.md
tserong: Complete failure would leave the host without network connectivity as you say, and the admin would need to access it via remote console to diagnose and repair.

Role promotion should only happen if a host is actually removed. AFAIK that won't happen automatically if a host just becomes uncontactable for an extended period.

Could user apps be broken by a messed up network config?  Uh.. I'm not sure. All we do statically (i.e. the stuff we're changing the configuration of here) is the management interface and how it comes up on boot. Other networks that might affect workloads (extra cluster networks, vm networks, storage network) are all created by harvester dynamically at runtime and so thus shouldn't(!) be affected by this change.

irishgordo:
@tserong for [Harvester ISO the "Net Install" variant](https://docs.harvesterhci.io/v1.7/install/net-install) will there be any extra considerations needed from the network manager switch?

enhancements/20250902-networkmanager.md
tserong: > Complete failure would leave the host without network connectivity as you say, and the admin would need to access it via remote console to diagnose and repair.

Just to add to that -- it turns out this is (or should be) _really easy_ to do if necessary since https://github.com/harvester/harvester-installer/pull/1150 went in:

- Login via remote console and become root
- Edit `/oem/harvester.config` and make sure `install: managementinterface: ....` specifies the network configuration you want
- Run `harvester-installer generate-network-yaml`. By default (i.e. with no other options specified) this will create `/oem/91_networkmanager.yaml` based on ` oem/harvester.config`.
- Reboot and enjoy your now functional network

tserong:
> for [Harvester ISO the "Net Install" variant](https://docs.harvesterhci.io/v1.7/install/net-install) will there be any extra considerations needed from the network manager switch?

I don't believe so @irishgordo but we should probably give it a quick test run to make sure.

tserong:
BTW there's a WIP docs PR at https://github.com/harvester/docs/pull/897, which between the finished bits and the bits still marked "DOCS TODO" should hopefully cover all the relevant changes.

enhancements/20250902-networkmanager.md
tserong: I've opened a couple of enhancement issues to cover the above:

https://github.com/harvester/harvester/issues/9300
https://github.com/harvester/harvester/issues/9301

tserong:
Rebased and updated to reflect changes from https://github.com/harvester/harvester/issues/9581

tserong:
rebased

ibrokethecloud (approved):

Issue #9022, opened 2025-09-02
[ENHANCEMENT] Make use of status conditions when activating maintenance mode


votdev:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Right now there are only set several annotations on the node resource when the maintenance mode is activated. There is not much information available for users and the UI to see what is going on.

### Describe the solution you'd like

Make use of `status.conditions` when the maintenance mode is activated or failing. This way the user and the UI can make use of this information to better understand what is going. Beside that it can be helpful in situations when things go wrong.

The condition `type` will be named `MaintenanceMode`. The following `reasons` will be available:
- Initializing
- Running
- Completed
- Error

The `status.condition` `MaintenanceMode` will be removed when the maintenance mode is deactivated on the node.

### Describe alternatives you've considered

n/a

### Additional context

Related to:
- https://github.com/harvester/harvester/issues/8985
- https://github.com/harvester/harvester/issues/8966

innobead:
cc @harvester/release @Vicente-Cheng

votdev:
The PR has been switched to draft/WIP because my gut feeling tells me that I'm not yet sure whether the implementation is perfect. Using status messages is a cool thing and makes background activities a little more visible, but I don't like the approach that displays the message in the UI for ~2 minutes. It just doesn't feel right somehow.

olblak:
Moving to 1.10

PR #9041, opened 2025-09-03
[ENHANCEMENT] Make use of status conditions when activating maintenance mode

votdev:
#### Problem:
Right now there are only set several annotations on the node resource when the maintenance mode is activated. There is not much information available for users and the UI to see what is going on.

#### Solution:
Make use of `status.conditions` when the maintenance mode is activated or failing. This way the user and the UI can make use of this information to better understand what is going. Beside that it can be helpful in situations when things go wrong.

The `status.condition` `MaintenanceMode` will be kept until the maintenance mode is deactivated. An error condition is reset at the start of a new maintenance mode cycle. This means that the error message will remain displayed in the `Hosts` view of the UI.

The condition `type` will be named `MaintenanceMode`. The following `reasons` will be available:
- Running
- Completed
- Error

By giving the condition that is set in case of a failure the name `Error`, the [Wrangler code](https://github.com/rancher/wrangler/blob/73fdb33d6a7529e14e2f769b66f717468a09201d/pkg/summary/summarizers.go#L320) will recognize it and the error is displayed in the Harvester `Hosts` page below the node that is going to be put into maintenance mode.

#### Problem:
There's a catch with this PR that I don't know how to resolve: resetting an error. The error remains displayed until maintenance mode is successfully activated. K8s does not provide a way to manually reset status conditions.
Automatic deletion of the error condition was not desired and was removed from the code in the latest commit of the PR.
In the event of an error, the annotations `harvesterhci.io/drain-requested` and `harvesterhci.io/drain-forced` are removed. If these were retained, the system would constantly attempt to activate maintenance mode until the process could be completed. If this behavior is desired, deleting the status condition could be accompanied by removing the `harvesterhci.io/drain-requested` annotation. With this approach, however, the error message is only displayed for a short time—until the next reconciliation run is started (which, in the current implementation, causes the condition to be reset).

No matter how you look at it, the maintenance mode feature just doesn't integrate very well into the UI.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9022

#### Test plan:
Prepare a 3 node test cluster.

***Case 1***:
- Go to the `Hosts` page.
- Choose `Enable Maintenance Mode` in the action menu on `harvester-node-0`.
- Do not check the `Force` checkbox.
- During the time `Cordoned` is displayed in the UI, a `kubectl get node harvester-node-0 -o jsonpath='{.status.conditions[*]}' | jq` will not contain any condition of type `MaintenanceMode`.
- After some time the node should be in maintenance mode.

Bildschirmfoto vom 2025-12-15 16-12-24

- The node `harvester-node-0` should have the annotation `harvesterhci.io/maintain-status=completed`.
- Run `k get node harvester-node-0 -o jsonpath='{.status.conditions[*]}' | jq`. The output should look like:

```
...
{
  "lastHeartbeatTime": "2025-12-15T15:12:05Z",
  "lastTransitionTime": "2025-12-15T15:12:05Z",
  "message": "Maintenance mode enabled",
  "reason": "Completed",
  "status": "True",
  "type": "MaintenanceMode"
}
```
- Choose `Disable Maintenance Mode` in the action menu on `harvester-node-0`.
- Run `kubectl get node harvester-node-0 -o jsonpath='{.status.conditions[*]}' | jq`. The output should not contain any condition of type `MaintenanceMode`.
- The annotation `harvesterhci.io/maintain-status` should have been removed.

***Case 2***:
- Go to the `Hosts` page.
- Add annotation to node: `kubectl annotate node harvester-node-1 harvesterhci.io/maintain-status='completed'`.
- Add annotation to node: `kubectl annotate node harvester-node-2 harvesterhci.io/drain-requested='true'`.
- The error message will remain displayed until maintenance mode is disabled on `harvester-node-1` and enabled on `harvester-node-2`.

image

***Case 3***:
- Create a VM named `vm01`.
- Make sure the VM is running on node `harvester-node-0`. Migrate the VM to this node if necessary.
- Add annotation to node: `kubectl annotate node harvester-node-0 harvesterhci.io/drain-requested='true'`.
- After some time an error message should appear:

image

- Choose the `Enable Maintenance Mode` action menu on `harvester-node-0`. Click the `Force` checkbox and press `Save`.
- The error message will disappear shortly.
- The `YAML` of the resource should contain the annotations:
```
    harvesterhci.io/drain-forced: 'true'
    harvesterhci.io/drain-requested: 'true'
```
- Run `kubectl get node harvester-node-0 -o jsonpath='{.status.conditions[*]}' | jq`. The `type=MaintenanceMode` condition should look like this.
```
{
  "lastHeartbeatTime": "2026-05-20T11:58:55Z",
  "lastTransitionTime": "2026-05-20T11:58:55Z",
  "message": "Maintenance mode enabled",
  "reason": "Completed",
  "status": "True",
  "type": "MaintenanceMode"
}
```
- The annotations `harvesterhci.io/drain-forced` and `harvesterhci.io/drain-requested`  need to be removed from the resource. The following annotation need to exist:
```
    harvesterhci.io/maintain-status: completed
```

### DIFF
--- a/pkg/controller/master/node/maintain_controller.go
+++ b/pkg/controller/master/node/maintain_controller.go
@@ -3,6 +3,7 @@ package node
 import (
 	"context"
 	"fmt"
+	"time"
 
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
@@ -17,16 +18,17 @@ import (
 )
 
 const (
-	maintainNodeControllerName  = "maintain-node-controller"
-	MaintainStatusAnnotationKey = "harvesterhci.io/maintain-status"
-	MaintainStatusComplete      = "completed"
-	MaintainStatusRunning       = "running"
+	maintainNodeControllerName        = "maintain-node-controller"
+	maintainErrorRetentionTimeSeconds = 120
+	MaintainStatusAnnotationKey       = "harvesterhci.io/maintain-status"
+	MaintainStatusComplete            = "completed"
+	MaintainStatusRunning             = "running"
 )
 
-// maintainNodeHandler updates maintenance status of a node in its annotations, so that we can tell whether the node is
-// entering maintenance mode(migrating VMs on it) or in maintenance mode(VMs migrated).
+// maintainNodeHandler updates the maintenance status of a node in its annotations, so that we can tell whether the node is
+// entering maintenance mode (migrating VMs on it) or in maintenance mode(VMs migrated).
 type maintainNodeHandler struct {
-	nodes                       ctlcorev1.NodeClient
+	nodes                       ctlcorev1.NodeController
 	nodeCache                   ctlcorev1.NodeCache
 	virtualMachineClient        v1.VirtualMachineClient
 	virtualMachineCache         v1.VirtualMachineCache
@@ -57,7 +59,39 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
 	if node == nil || node.DeletionTimestamp != nil {
 		return node, nil
 	}
+
 	if maintenanceStatus, ok := node.Annotations[MaintainStatusAnnotationKey]; !ok || maintenanceStatus != MaintainStatusRunning {
+		// Make sure the maintenance mode status condition is cleaned up when
+		// there is no "harvesterhci.io/maintain-status" annotation (e.g., the
+		// maintenance mode has been disabled).
+
+		// If it is an error status condition, then reconcile again for a short
+		// period (~ 2 minute) so that there is some time to display the error
+		// on the `Hosts` page (Wrangler collects such status conditions and
+		// the UI displays them). If the status is older than the configured
+		// retention time, proceed with the cleanup.
+		condition := util.FindNodeStatusCondition(node.Status.Conditions, util.NodeConditionTypeMaintenanceMode)
+		if condition != nil && condition.Reason == util.NodeConditionReasonError {
+			retentionTime := time.Duration(maintainErrorRetentionTimeSeconds) * time.Second
+			if time.Since(condition.LastHeartbeatTime.Time) <= retentionTime {
+				h.nodes.EnqueueAfter(node.Name, 5*time.Second)
+				return node, nil
+			}
+		}
+
+		// Clean up the maintenance mode status condition if the annotation is
+		// not present; otherwise leave it as is.
+		if !ok {
+			nodeCopy := node.DeepCopy()
+			removed := util.RemoveNodeStatusCondition(nodeCopy, util.NodeConditionTypeMaintenanceMode)
+			if removed {
+				if _, err := h.nodes.UpdateStatus(nodeCopy); err != nil {
+					return node, fmt.Errorf("failed to clean up %q condition: %w", util.NodeConditionTypeMaintenanceMode, err)
+				}
+				return nodeCopy, nil
+			}
+		}
+
 		return node, nil
 	}
 
@@ -70,48 +104,21 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
 		return node, nil
 	}
 
-	// Restart those VMs that have been labeled to be shut down before
-	// maintenance mode and that should be restarted when the node has
-	// successfully switched into maintenance mode.
-	selector := labels.Set{util.LabelMaintainModeStrategy: util.MaintainModeStrategyShutdownAndRestartAfterEnable}.AsSelector()
-	vmList, err := h.virtualMachineCache.List(node.Namespace, selector)
+	// Restart VMs that were shut down for maintenance and should be restarted.
+	if err = h.restartShutdownVMs(node); err != nil {
+		return node, err
+	}
+
+	nodeCopy := node.DeepCopy()
+	nodeCopy.Annotations[MaintainStatusAnnotationKey] = MaintainStatusComplete
+	nodeCopy, err = h.nodes.Update(nodeCopy)
 	if err != nil {
-		return node, fmt.Errorf("failed to list VMs with labels %s: %w", selector.String(), err)
-	}
-	for _, vm := range vmList {
-		// Make sure that this VM was shut down as part of the maintenance
-		// mode of the given node.
-		if vm.Annotations[util.AnnotationMaintainModeStrategyNodeName] != node.Name {
-			continue
-		}
-
-		logrus.WithFields(logrus.Fields{
-			"namespace":           vm.Namespace,
-			"virtualmachine_name": vm.Name,
-		}).Info("restarting the VM that was temporary shut down for maintenance mode")
-
-		// Update the run strategy of the VM to start it and remove the
-		// annotation that was previously set when the node went into
-		// maintenance mode.
-		// Get the running strategy that is stored in the annotation of the
-		// VM when it is shut down. Note, in general this is automatically
-		// patched by the VM mutator in general.
-		runStrategy := kubevirtv1.VirtualMachineRunStrategy(vm.Annotations[util.AnnotationRunStrategy])
-		if runStrategy == "" {
-			runStrategy = kubevirtv1.RunStrategyRerunOnFailure
-		}
-		vmCopy := vm.DeepCopy()
-		vmCopy.Spec.RunStrategy = &[]kubevirtv1.VirtualMachineRunStrategy{runStrategy}[0]
-		delete(vmCopy.Annotations, util.AnnotationMaintainModeStrategyNodeName)
-		_, err = h.virtualMachineClient.Update(vmCopy)
-		if err != nil {
-			return node, fmt.Errorf("failed to start VM %s/%s: %w", vm.Namespace, vm.Name, err)
-		}
+		return node, err
 	}
 
-	toUpdate := node.DeepCopy()
-	toUpdate.Annotations[MaintainStatusAnnotationKey] = MaintainStatusComplete
-	return h.nodes.Update(toUpdate)
+	util.SetNodeStatusCondition(nodeCopy, util.NodeConditionTypeMaintenanceMode, corev1.ConditionTrue, util.NodeConditionReasonCompleted, "Maintenance mode enabled")
+
+	return h.nodes.UpdateStatus(nodeCopy)
 }
 
 // OnNodeRemoved Ensure that all "harvesterhci.io/maintain-mode-strategy-node-name"
@@ -143,4 +150,50 @@ func (h *maintainNodeHandler) OnNodeRemoved(_ string, node *corev1.Node) (*corev
 	}
 
 	return node, nil
+}
+
+// restartShutdownVMs restarts those VMs that have been labeled to be shut down
+// before maintenance mode, and that should be restarted when the node has
+// successfully switched into maintenance mode.
+func (h *maintainNodeHandler) restartShutdownVMs(node *corev1.Node) error {
+	selector := labels.Set{util.LabelMaintainModeStrategy: util.MaintainModeStrategyShutdownAndRestartAfterEnable}.AsSelector()
+	vmList, err := h.virtualMachineCache.List(node.Namespace, selector)
+	if err != nil {
+		return fmt.Errorf("failed to list VMs with labels %s: %w", selector.String(), err)
+	}
+
+	for _, vm := range vmList {
+		// Make sure that this VM was shut down as part of the maintenance
+		// mode of the given node.
+		if vm.Annotations == nil || vm.Annotations[util.AnnotationMaintainModeStrategyNodeName] != node.Name {
+			continue
+		}
+
+		logrus.WithFields(logrus.Fields{
+			"namespace":           vm.Namespace,
+			"virtualmachine_name": vm.Name,
+		}).Info("Restarting the VM that was temporarily shut down for maintenance mode")
+
+		// Update the run strategy of the VM to start it and remove the
+		// annotation that was previously set when the node went into
+		// maintenance mode.
+		// Get the running strategy that is stored in the annotation of the
+		// VM when it is shut down. Note, in general this is automatically
+		// patched by the VM mutator in general.
+		runStrategy := kubevirtv1.VirtualMachineRunStrategy(vm.Annotations[util.AnnotationRunStrategy])
+		if runStrategy == "" {
+			runStrategy = kubevirtv1.RunStrategyRerunOnFailure
+		}
+
+		vmCopy := vm.DeepCopy()
+		vmCopy.Spec.RunStrategy = &[]kubevirtv1.VirtualMachineRunStrategy{runStrategy}[0]
+		delete(vmCopy.Annotations, util.AnnotationMaintainModeStrategyNodeName)
+
+		_, err = h.virtualMachineClient.Update(vmCopy)
+		if err != nil {
+			return fmt.Errorf("failed to start VM %s/%s: %w", vm.Namespace, vm.Name, err)
+		}
+	}
+
+	return nil
 }
--- a/pkg/controller/master/node/maintain_controller.go
+++ b/pkg/controller/master/node/maintain_controller.go
@@ -3,7 +3,6 @@ package node
 import (
 	"context"
 	"fmt"
-	"time"
 
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
@@ -18,17 +17,13 @@ import (
 )
 
 const (
-	maintainNodeControllerName        = "maintain-node-controller"
-	maintainErrorRetentionTimeSeconds = 120
-	MaintainStatusAnnotationKey       = "harvesterhci.io/maintain-status"
-	MaintainStatusComplete            = "completed"
-	MaintainStatusRunning             = "running"
+	maintainNodeControllerName = "maintain-node-controller"
 )
 
 // maintainNodeHandler updates the maintenance status of a node in its annotations, so that we can tell whether the node is
 // entering maintenance mode (migrating VMs on it) or in maintenance mode(VMs migrated).
 type maintainNodeHandler struct {
-	nodes                       ctlcorev1.NodeController
+	nodes                       ctlcorev1.NodeClient
 	nodeCache                   ctlcorev1.NodeCache
 	virtualMachineClient        v1.VirtualMachineClient
 	virtualMachineCache         v1.VirtualMachineCache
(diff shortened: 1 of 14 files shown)


pkg/controller/master/node/maintain_controller.go
```diff
@@ -57,7 +59,20 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
 	if node == nil || node.DeletionTimestamp != nil {
 		return node, nil
 	}
+
 	if maintenanceStatus, ok := node.Annotations[MaintainStatusAnnotationKey]; !ok || maintenanceStatus != MaintainStatusRunning {
+		if !ok {
```
w13915984028: Double check: does the code targets below?

```
	maintenanceStatus, ok := node.Annotations[MaintainStatusAnnotationKey];
	if !ok {
		// Make sure the maintenance mode status condition is cleaned up
		// when there is no "harvesterhci.io/maintain-status" annotation.
		nodeCopy := node.DeepCopy()
		util.RemoveConditionFromNode(nodeCopy, util.NodeConditionTypeMaintenanceMode)
		if !reflect.DeepEqual(node.Status.Conditions, nodeCopy.Status.Conditions) {
			if _, err := h.nodes.UpdateStatus(nodeCopy); err != nil {
				return node, fmt.Errorf("failed to clean up %q condition: %w", util.NodeConditionTypeMaintenanceMode, err)
			}
			return nodeCopy, nil
		}

		// condition has already been removed
		return node, nil
	}
	
	if maintenanceStatus != MaintainStatusRunning {
		return node, nil
	}
```

pkg/util/node.go
```diff
@@ -13,3 +13,31 @@ func ExcludeWitnessNodes(nodes []*corev1.Node) []*corev1.Node {
...
+	found := false
+	for i := range node.Status.Conditions {
+		c := &node.Status.Conditions[i]
+		if c.Type == newCondition.Type {
+			found = true
+			c.Status = newCondition.Status
+			c.LastHeartbeatTime = newCondition.LastHeartbeatTime
+			c.LastTransitionTime = newCondition.LastTransitionTime
+			c.Reason = newCondition.Reason
+			c.Message = newCondition.Message
```
w13915984028: add a break or return when found

pkg/util/node.go
```diff
@@ -13,3 +13,31 @@ func ExcludeWitnessNodes(nodes []*corev1.Node) []*corev1.Node {
...
+		}
+	}
+	if !found {
+		node.Status.Conditions = append(node.Status.Conditions, newCondition)
+	}
+}
+
+func RemoveConditionFromNode(node *corev1.Node, condType corev1.NodeConditionType) {
+	newConditions := make([]corev1.NodeCondition, 0, len(node.Status.Conditions))
```
w13915984028: the target condition might not be in the slice

an even effective solution is to find the index first, and then use slice.Delete if index > 0

https://pkg.go.dev/golang.org/x/exp/slices#Delete

current solution will copy n-1/n elements all the way

w13915984028 (commented):
some questions

pkg/controller/master/node/maintain_controller.go
```diff
@@ -19,16 +20,17 @@ import (
...
 const (
-	maintainNodeControllerName  = "maintain-node-controller"
-	MaintainStatusAnnotationKey = "harvesterhci.io/maintain-status"
-	MaintainStatusComplete      = "completed"
-	MaintainStatusRunning       = "running"
+	maintainNodeControllerName        = "maintain-node-controller"
+	MaintainStatusAnnotationKey       = "harvesterhci.io/maintain-status"
+	MaintainStatusComplete            = "completed"
+	MaintainStatusRunning             = "running"
+	maintainErrorRetentionTimeSeconds = 60
```
w13915984028: consider to emit an event on failure case for better trouble shooting, or there have been such events?

pkg/controller/master/node/maintain_controller.go
```diff
@@ -57,7 +61,33 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
 	if node == nil || node.DeletionTimestamp != nil {
 		return node, nil
 	}
+
 	if maintenanceStatus, ok := node.Annotations[MaintainStatusAnnotationKey]; !ok || maintenanceStatus != MaintainStatusRunning {
+		// Make sure the maintenance mode status condition is cleaned up when
+		// there is no "harvesterhci.io/maintain-status" annotation.
+		if !ok {
+			// If it is an error status condition, delay the cleanup so that
```
ihcsim: Is it necessary to do the garbage collection? Any issues with just retaining it until the next maintenance cycle, and let `util.AddOrUpdateConditionToNode()` at line 150 manage it? Since the condition has its own heartbeat timestamp, there shouldn't be any ambiguity.

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -210,19 +223,56 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
+		return node, err
+	}
+
+	util.AddOrUpdateConditionToNode(nodeCopy, corev1.NodeCondition{
+		Type:               util.NodeConditionTypeMaintenanceMode,
+		Status:             corev1.ConditionTrue,
+		LastHeartbeatTime:  metav1.Now(),
+		LastTransitionTime: metav1.Now(),
+		Reason:             util.NodeConditionReasonRunning,
+		Message:            "Draining the node is running",
```
```suggestion
		Message:            "Draining the node",
```

pkg/controller/master/node/maintain_controller.go
```diff
@@ -109,9 +139,24 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
...
+		return node, err
+	}
+
+	util.AddOrUpdateConditionToNode(nodeCopy, corev1.NodeCondition{
+		Type:               util.NodeConditionTypeMaintenanceMode,
+		Status:             corev1.ConditionTrue,
+		LastHeartbeatTime:  metav1.Now(),
+		LastTransitionTime: metav1.Now(),
+		Reason:             util.NodeConditionReasonCompleted,
+		Message:            "Draining the node is completed",
```
ihcsim:
```suggestion
		Message:            "Maintenance mode enabled",
```
Nit: I think technically this controller is more about enabling/disabling maintenance mode than about draining the node, right?

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -210,19 +223,56 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
 
-	return ndc.nodes.Update(nodeCopy)
+	nodeCopy, err = ndc.nodes.Update(nodeCopy)
+	if err != nil {
+		return node, err
+	}
+
+	util.AddOrUpdateConditionToNode(nodeCopy, corev1.NodeCondition{
+		Type:               util.NodeConditionTypeMaintenanceMode,
+		Status:             corev1.ConditionTrue,
```
ihcsim: If it is still running, it means maintenance mode isn't fully enabled yet. I'd think status should be `false`?

pkg/util/node_test.go
```diff
@@ -0,0 +1,109 @@
...
+					LastTransitionTime: metav1.Now(),
+					Reason:             NodeConditionReasonCompleted,
+					Message:            "Draining the node is completed",
+				},
+			},
+		},
+	}
+	condition := GetConditionFromNode(node, NodeConditionTypeMaintenanceMode)
+	assert.NotNil(t, condition)
+	assert.Equal(t, condition.Type, NodeConditionTypeMaintenanceMode)
```
ibrokethecloud: incorrect comparison types

this is currently causing CI to fail since test fails

pkg/util/constants.go
```diff
@@ -237,3 +237,10 @@ const (
 	FSOverheadRegex                               = `^(0(?:\.\d{1,3})?|1)$`
 	PVCExpandErrorPrefix                          = "PVC_EXPAND"
 )
+
+const (
+	NodeConditionTypeMaintenanceMode = "MaintenanceMode"
```
ibrokethecloud: should defined this to be of type https://pkg.go.dev/k8s.io/api/core/v1#NodeConditionType 
this will make us having to avoid type assertions later on in the tests

pkg/controller/master/node/maintain_controller.go
ibrokethecloud: i feel 60 seconds may be too small a window for users to pick up the error in the UI. Do we want to start with say 300 seconds?

pkg/controller/master/node/maintain_controller.go
ibrokethecloud: wrangler summarisers will look at error conditions in status subresource and generate a UI error. The idea with this change was to provide a visual notification in harvester ui when the maintinance operation fails. if we do not remove the error condition the error will stay in the UI which could be confusing for the user since there is nothing actually wrong with the underlying node

More information is available here https://github.com/rancher/wrangler/blob/main/pkg/summary/summarizers.go

pkg/util/node.go
votdev: Kubernetes API says: Conditions should be represented as a list of objects, where each condition type appears at most once.

BUT: There is no validation that there is only ONE entry per condition type.

My implementation tries to be on the safe side and update ALL entries, even though this is conceptually incorrect.

pkg/controller/master/node/maintain_controller.go
votdev: Yes, IMO the code is handling that.

pkg/util/node.go
votdev: Same as [here](https://github.com/harvester/harvester/pull/9041#discussion_r2523195190).

The API says that there should be one condition per type, but it does not force or validate it. My implementation takes a more conservative approach and is more fault-tolerant IF the situation should arise where there are several conditions of one type. Since we are not talking about actions that are performed thousands of times per second, and a condition list does not have millions of entries, looping through all elements should not take much more time and resources.

pkg/controller/master/node/maintain_controller.go
votdev: We need to clean it up as it affects the display in the UI.

pkg/util/node.go
w13915984028: that's a good point,

on the other hand, if other contrllers made mistakes, and this PR removes one of the duplicated ones, it might break others

keep unrelated as is brings less risk

pkg/controller/master/node/maintain_controller.go
votdev: Increased to 300 seconds.

pkg/util/node.go
votdev: The helper function in the meta package iterates over all conditions as well: https://github.com/kubernetes/apimachinery/blob/b72d93d174332f952a8d431419fece5e6f044bcb/pkg/api/meta/conditions.go#L73

go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/harvester/harvester
 
 go 1.25
+toolchain go1.25.2
```
w13915984028: There is another PR
https://github.com/harvester/harvester-installer/pull/1188 to avoid set toolchain specific minor version,  here it could be similar

pkg/controller/master/node/maintain_controller.go
```diff
@@ -57,7 +60,36 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
...
+		// Make sure the maintenance mode status condition is cleaned up when
+		// there is no "harvesterhci.io/maintain-status" annotation.
+		if !ok {
+			// If it is an error status condition, delay the cleanup so that
+			// there is some time to display the error on the `Hosts` page
+			// (Wrangler collects such status conditions and the UI displays
+			// them).
+			condition := util.GetConditionFromNode(node, util.NodeConditionTypeMaintenanceMode)
+			if condition != nil && condition.Reason == util.NodeConditionReasonError {
+				retentionTime := time.Duration(maintainErrorRetentionTimeSeconds) * time.Second
```
w13915984028: seems it will wait 300 seconds anyway, even when Harvester related annotation is removed, need to double check, if it should recover asap e.g., user stop the maintenance mode.

w13915984028 (commented):
A few last concerns.

And as the `test steps` is TODO, need to figure out the UI, controller, condition status transferring inbetween different cases, thanks.

pkg/controller/master/node/maintain_controller.go
ihcsim: Does the controller remove the error condition if the user re-triggers maintenance mode after the previous failed attempt? AIUI, the condition is removed only after the timeout at line 75. It looks like the condition will remain even though subsequent attempts to activate maintenance mode succeeded, before the timeout expired.

go.mod
votdev: Removed.

pkg/controller/master/node/maintain_controller.go
votdev: Improved

pkg/controller/master/node/maintain_controller.go
votdev: Improved

pkg/controller/master/node/maintain_controller.go
votdev: Changed to 120 seconds because 5 minutes is too long.

votdev:
Switching to draft/WIP because i feel like the feature isn't running smoothly yet. Using status messages is a cool thing and makes background activities a little more visible, but I don't like the workaround that displays the message in the UI for ~2 minutes. It just doesn't feel right somehow.

pkg/controller/master/node/maintain_controller.go
```diff
@@ -57,7 +59,39 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
...
+		// Clean up the maintenance mode status condition if the annotation is
+		// not present; otherwise leave it as is.
+		if !ok {
+			nodeCopy := node.DeepCopy()
+			removed := util.RemoveNodeStatusCondition(nodeCopy, util.NodeConditionTypeMaintenanceMode)
+			if removed {
+				if _, err := h.nodes.UpdateStatus(nodeCopy); err != nil {
+					return node, fmt.Errorf("failed to clean up %q condition: %w", util.NodeConditionTypeMaintenanceMode, err)
+				}
+				return nodeCopy, nil
```
w13915984028: Could we leave a logrus.Infof to indicate the user removal or auto removal of (failure) maintenance

pkg/controller/master/node/maintain_controller.go
```diff
@@ -57,7 +59,39 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
 	if node == nil || node.DeletionTimestamp != nil {
 		return node, nil
 	}
+
 	if maintenanceStatus, ok := node.Annotations[MaintainStatusAnnotationKey]; !ok || maintenanceStatus != MaintainStatusRunning {
+		// Make sure the maintenance mode status condition is cleaned up when
```
w13915984028: it is a bit ambiguous here:

if there is an error status, the disable-maintenance won't clear the status promptly but still wait until timeout?

the !ok and failure are different cases

pkg/controller/master/node/maintain_controller.go
```diff
@@ -57,7 +59,39 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
...
+		// period (~ 2 minute) so that there is some time to display the error
+		// on the `Hosts` page (Wrangler collects such status conditions and
+		// the UI displays them). If the status is older than the configured
+		// retention time, proceed with the cleanup.
+		condition := util.FindNodeStatusCondition(node.Status.Conditions, util.NodeConditionTypeMaintenanceMode)
+		if condition != nil && condition.Reason == util.NodeConditionReasonError {
+			retentionTime := time.Duration(maintainErrorRetentionTimeSeconds) * time.Second
+			if time.Since(condition.LastHeartbeatTime.Time) <= retentionTime {
+				h.nodes.EnqueueAfter(node.Name, 5*time.Second)
+				return node, nil
```
w13915984028: the error case is processed before `!ok`, and the latter is normally triggered by user operation

pkg/util/drainhelper/helper.go
```diff
@@ -173,3 +168,10 @@ func maintainModeStrategyFilter(pod corev1.Pod) drain.PodDeleteStatus {
 	}
 	return drain.MakePodDeleteStatusOkay()
 }
+
+func injectErrorFilter(pod corev1.Pod) drain.PodDeleteStatus {
```
w13915984028: doubt if we should leave such code on the formal PR, can we achive the test purpose by e.g. set vm to run on specific node

w13915984028 (commented):
Some questions:

As the PR has been standing for so long, it takes time to recall all scenarios.

Could we add full test cases and local test report on a new thread/reply to the PR? e.g. if a maintain failed or is still running, if user disable the maintenance, will it be rejected, or the condtion is cleaned at now or others.  thanks.

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -217,19 +237,41 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
 
 	// run node drain
 	err = drainhelper.DrainNode(ndc.context, ndc.restConfig, nodeCopy)
 	if err != nil {
-		return node, err
+		logrus.WithFields(logrus.Fields{
+			"node_name": node.Name,
+		}).Errorf("Failed to drain the node: %v", err)
+
+		util.SetNodeStatusCondition(nodeCopy, util.NodeConditionTypeMaintenanceMode, corev1.ConditionTrue, util.NodeConditionReasonError, err.Error())
```
ihcsim, line 258: Do we need to call `ndc.cleanupMaintenanceModeAnnotations()` before calling `util.SetNodeStatusCondition()` here, as seen in line 102 and 136? Otherwise, the maintenance controller won't auto-remove the error condition after the retention duration expired.

pkg/controller/master/node/maintain_controller.go
```diff
@@ -57,7 +59,39 @@ func (h *maintainNodeHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev
...
 	if maintenanceStatus, ok := node.Annotations[MaintainStatusAnnotationKey]; !ok || maintenanceStatus != MaintainStatusRunning {
+		// Make sure the maintenance mode status condition is cleaned up when
+		// there is no "harvesterhci.io/maintain-status" annotation (e.g., the
+		// maintenance mode has been disabled).
+
+		// If it is an error status condition, then reconcile again for a short
+		// period (~ 2 minute) so that there is some time to display the error
+		// on the `Hosts` page (Wrangler collects such status conditions and
+		// the UI displays them). If the status is older than the configured
+		// retention time, proceed with the cleanup.
```
ihcsim: I think we should retain the error conditions and dismiss them only when the user tries to [re-enable/disable maintenance mode](https://github.com/harvester/harvester/blob/d7b94aa6abfbcbaca1b3fe546af0c6e70818e9fc/pkg/api/node/formatter.go#L114-L117). This auto-remove mechanism assumes that the user sees and fixes the errors within 2 minutes of the last failed  attempt. Once the error condition is auto-dismissed, there won't be any records of previous failures.

pkg/controller/master/node/maintain_controller.go
ihcsim: AIUI, if there is an error condition, the annotation can't be removed [via the user workflow](https://github.com/harvester/harvester/blob/d7b94aa6abfbcbaca1b3fe546af0c6e70818e9fc/pkg/api/node/formatter.go#L231) because the "disable maintenance mode" UI action won't be available to the user. The only way to recover from it is if the next "enable maintenance mode" attempt works flawlessly, and let the node drain controller and maintenance controller reconcile the annotations and conditions.

I find this annotation hard to reason about due to the way it's shared between the maintenance controller, the node drain controller and the frontend-facing formatter.

w13915984028: As this feature touches node maintenance, node-drain and UI operations, we need a defined user-stroies (and later on the document) about in which case what will happen/could be done. (black box behaviour)

Without such a defined user-stroies, it is not easy to review the code, do QA validations and write documents. thanks.

votdev:
Thanks for the comments. I've moved the issue back to `Implement` because it seems this PR needs an overall rewrite.

pkg/controller/master/node/maintain_controller.go
votdev: A comprehensive revision has been made. The handling of status conditions has been changed, so this comment is no longer relevant.

pkg/controller/master/node/maintain_controller.go
votdev: A comprehensive revision has been made. The handling of status conditions has been changed, so this comment is no longer relevant.

pkg/controller/master/node/maintain_controller.go
votdev: A comprehensive revision has been made. The handling of status conditions has been changed, so this comment is no longer relevant.

pkg/util/drainhelper/helper.go
votdev: This code was introduced because it is difficult to reproduce this error in test cases in any other way.

But I'm on your side; I'd actually prefer a different approach. However, I removed the code again in the latest commit.

pkg/controller/master/nodedrain/nodedrain_controller.go
votdev: I think this comment is outdated after the refactoring of the PR.

pkg/controller/master/node/maintain_controller.go
votdev: Done with the refactoring of the PR.

votdev:
What do you think about completely removing the `harvesterhci.io/maintain-status` annotation and handling everything solely through status conditions? We’ll keep the `harvesterhci.io/drain-requested` and `harvesterhci.io/drain-forced` annotations, though. I haven’t fully worked through this yet, but I wanted to get your thoughts on it in advance.

ihcsim:
@votdev Depending on the scope of the change, we might need a HEP for it to give others a chance to chime in and more importantly, for QA to understand the test scope. Last I looked at it, the annotation logic is not easy to reason about, because 2 (or 3?) controllers are using it to share state. Can changing it from annotation to condition help to simplify the state coordination logic?

votdev:
> @votdev Depending on the scope of the change, we might need a HEP for it to give others a chance to chime in and more importantly, for QA to understand the test scope. Last I looked at it, the annotation logic is not easy to reason about, because 2 (or 3?) controllers are using it to share state. Can changing it from annotation to condition help to simplify the state coordination logic?

The HEP can be found at https://github.com/harvester/harvester/pull/10705

votdev:
Superseded by https://github.com/harvester/harvester/pull/11606

PR #9096, opened 2025-09-08
feat: use addedNodeSelector from virtualmachineinstancemigrations

Yu-Jack:
#### Problem:
Currently, we hack the VMI node selector during migration ([Ref. 1](https://github.com/harvester/harvester/commit/0f960117c906262d534eb5b4c412bf61c5ca21a4) and [Ref. 2](https://github.com/harvester/harvester/commit/719a5960fff7f73a69e14f1bb731c8f30e5c1ec8)), which is not a official way.

#### Solution:
Our kubevirt chart has been updated ([Ref.](https://github.com/harvester/harvester/pull/9094)).

Now, kubevirt supports adding node selector in virtualmachineinstancemigrations. Hence, we can use native way to migrate the VM. But, I still keep the annotations in the VMI.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8703

#### Test plan:

Prepare a three node clusters

Case 1: Migrate the VM to another node.
Case 2: Abort migration after starting migration.
Case 3: Start a virtual machine on a specific node, the VM is not allowed for migration.

None.

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -475,9 +475,6 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
 		if toUpdateVmi.Annotations == nil {
 			toUpdateVmi.Annotations = make(map[string]string)
 		}
-		if toUpdateVmi.Spec.NodeSelector == nil {
-			toUpdateVmi.Spec.NodeSelector = make(map[string]string)
-		}
 		toUpdateVmi.Annotations[util.AnnotationMigrationTarget] = nodeName
 
 		if err := util.VirtClientUpdateVmi(ctx, h.virtRestClient, h.namespace, namespace, vmName, toUpdateVmi); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/api/vm/handler.go
```diff
@@ -475,29 +475,15 @@ func (h *vmActionHandler) migrate(ctx context.Context, namespace, vmName string,
...
-		toUpdateVM := vm.DeepCopy()
-		if toUpdateVM.Spec.Template.Spec.NodeSelector == nil {
-			toUpdateVM.Spec.Template.Spec.NodeSelector = make(map[string]string)
-		}
-
-		toUpdateVM.Spec.Template.Spec.NodeSelector[corev1.LabelHostname] = nodeName
-		if _, err := h.vms.Update(toUpdateVM); err != nil {
-			logrus.Errorf("failed to update VM %s/%s with migration target %s: %v", namespace, vmName, nodeName, err)
-			return err
+		vmim.Spec.AddedNodeSelector = map[string]string{
```
Vicente-Cheng, line 485: Just curious, if the first vmim failure. Would the node selector be inherent to the next vmim?

Yu-Jack: From [source code](https://github.com/kubevirt/kubevirt/blob/4aedcf79a0b064c0a530e59eb2ce93f2f45bb85b/pkg/virt-controller/watch/migration/migration.go#L1451), I didn't see any vmim recreation. It seems they only re-queue the vmim key in some cases. Maybe it's a old behavior in old kubevirt version? Also, as we discussed offline, it should use the same vmim object if failure happens.

brandboat: > Just curious, if the first vmim failure. Would the node selector be inherent to the next vmim?

The VMIM is created by us (see line L490), and if my understanding is correct, KubeVirt doesn’t provide retry behavior—meaning that if a migration fails, KubeVirt won’t automatically create a new VMIM. Harvester doesn’t implement such a mechanism either. That means whenever a migration fails, we need to manually create another VMIM to trigger the VM migration. This also implies if we want to migrate a VM to a specific node, we’ll have to set AddedNodeSelector each time.

brandboat (approved):
LGTM, thanks!

Vicente-Cheng (approved):
thanks!

pkg/api/vm/handler.go
Vicente-Cheng: Thanks guys, very clear!

Issue #9056, opened 2025-09-04
[BUG] Clone VM only should support secret volume


FrankYang0529:
### Describe the Bug

Currently, we only support PVC, Cloud init, and container disk in clone VM action. We should also support secret volumes.

### To Reproduce

Clone a VM with secret volumes will get following error.
<img width="590" height="294" alt="Image" src="https://github.com/user-attachments/assets/70e972bc-dc0d-40d3-8cb1-72da13e0542a" />

### Expected Behavior

Support secret volumes.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: master-head
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): KVM

noahgildersleeve:
Validated in Harvester `v1.7.0-dev-20250914` with 2 node libvirt/KVM cluster. Validated as fixed with migration and clone.

<img width="356" height="130" alt="Image" src="https://github.com/user-attachments/assets/aab71530-8073-42f1-b98e-672b2d7d2fd9" />
<img width="615" height="347" alt="Image" src="https://github.com/user-attachments/assets/5f11ec0b-4328-4a9b-aa0f-93a0a2b87e48" />
<img width="600" height="276" alt="Image" src="https://github.com/user-attachments/assets/e3642cb2-cf7c-4bb8-b06b-309321453cf0" />
<img width="2262" height="374" alt="Image" src="https://github.com/user-attachments/assets/1d66057e-c440-42d7-9255-b6d4aa0d5267" />

PR #9097, opened 2025-09-08
feat: support clone secret disk

FrankYang0529:
#### Related Issue(s):

https://github.com/harvester/harvester/issues/9056

#### Test plan:
1. Create a VM.
2. Add a secret. `kubectl create secret generic my-secret --from-literal=username=admin --from-literal=password='S!B\*d$zDsb='`.
3. Add the secret as secret disk to the VM.
```yaml
# ...
           disks:
           # other disks ...
           - disk:
               bus: virtio
             name: secret-disk
# ...
       volumes:
       # other volumes ...
       - name: secret-disk
         secret:
           secretName: my-secret
```
4. Migrate the VM to another node successfully.

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -1275,14 +1275,23 @@ func (h *vmActionHandler) cloneVolumes(newVM *kubevirtv1.VirtualMachine) ([]core
 				}
 				volume.CloudInitNoCloud.NetworkDataSecretRef.Name = secretNameMap[volume.CloudInitNoCloud.NetworkDataSecretRef.Name]
 			}
+		} else if volume.Secret != nil {
+			cloneSecretVolume(&volume, secretNameMap)
 		} else if volume.ContainerDisk != nil {
 			continue
 		} else {
-			return nil, nil, fmt.Errorf("invalid volume %s, only support PersistentVolumeClaim, CloudInitNoCloud, and ContainerDisk", volume.Name)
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (dismissed):
thanks

pkg/api/vm/handler.go
```diff
@@ -1289,10 +1289,15 @@ func (h *vmActionHandler) cloneVolumes(newVM *kubevirtv1.VirtualMachine) ([]core
 				}
 				volume.CloudInitNoCloud.NetworkDataSecretRef.Name = secretNameMap[volume.CloudInitNoCloud.NetworkDataSecretRef.Name]
 			}
+		} else if volume.Secret != nil {
```
WebberHuang1118, line 1278: Maybe wrap this as a function for better readability?

WebberHuang1118 (dismissed):
LGTM, just a nit, thanks.

pkg/api/vm/handler.go
FrankYang0529: We already have `cloneVolumes`. Do you mean add a new `cloneVolume` for single volume?

WebberHuang1118: I mean the nested if statements aren’t very readable. We might consider handling the new part about `volume’s secret` with a utility function.

FrankYang0529: Updated it.

WebberHuang1118 (approved):

Vicente-Cheng (approved):
thanks

Issue #8257, opened 2025-05-12
[CI] Create a Sprint Release by the End of Sprint


innobead:
### Task request

In the past, we did not make sprint releases on a periodic basis. 

- [x] Release a sprint release by the end of the sprint. @harvester/qa can use the build to verify `read-for-testing` items from the previous engineering sprint.
- [x] Only enable the master,  n, and n-1 sprint releases. Currently (May 15, 2025), it's master (1.6), 1.5, and 1.4. 
- [x] Disable the sprint release when doing the official release process. 

Not in this Scope:

- [ ] 100% automate the release with an automatic smoke test. Release it regardless of whether the test succeeds or fails. If there is no existing smoke test, create a following task and plan it in a latter milestone. 

cc @harvester/release @harvester/maintainer

Vicente-Cheng:
Because of the newly added installer image, I need to make some changes to the current implementation.
We can validate on this sprint. If everything goes fine, we can close it.

Vicente-Cheng:
We can now create a sprint release by the end of the sprint automatically. 
Will improve the release note content in later versions.

PR #9123, opened 2025-09-11
ci: add build factory for build/build-for-tag

Vicente-Cheng:
#### Problem:
Remove the duplicated build code

#### Solution:
Remove the duplicated build code

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8257


.github/workflows/build-factory.yml
```diff
@@ -0,0 +1,366 @@
...
+    permissions:
+      contents: read
+      id-token: write # for reading credential https://github.com/rancher-eio/read-vault-secrets
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Declare branch
+      run: |
+        echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> "$GITHUB_ENV"
```
tserong: I don't think this `Declare branch` step needs to be here anymore - the things that were using it in `docker-pull-*` all seem to have been replaced with `inputs.image_tag`.

.github/workflows/build-factory.yml
```diff
@@ -0,0 +1,366 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Declare branch and sha_short
```
tserong: Probably should be named `Set arch` or similar

.github/workflows/build-factory.yml
```diff
@@ -0,0 +1,366 @@
...
+    permissions:
+      contents: read
+      id-token: write # for reading credential https://github.com/rancher-eio/read-vault-secrets
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Declare branch
+      run: |
+        echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> "$GITHUB_ENV"
```
tserong: I think this `Declare branch` step can also be removed (same reason as the other one).

tserong (commented):
I've found a couple of little things, otherwise LGTM, I hope! (these GHAs always hurt my brain)

.github/workflows/build-factory.yml
Vicente-Cheng: nice catch, updated!

.github/workflows/build-factory.yml
Vicente-Cheng: updated!

.github/workflows/build-factory.yml
Vicente-Cheng: updated!

tserong (approved):

starbops (approved):
LGTM, thank you!

Issue #9060, opened 2025-09-05
[ENHANCEMENT] Relax webhook validation for CDI annotations in storage class creation


brandboat:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

When trying to install a CSI driver (e.g. democratic-csi for TrueNAS iSCSI), the admission webhook validator.harvesterhci.io blocks the StorageClass creation if the required CDI annotation is missing:

```
admission webhook "validator.harvesterhci.io" denied the request: 
missing annotation cdi.harvesterhci.io/storageProfileVolumeModeAccessModes. 
volume access modes are required for CDI integration to work with storage class provisioner org.democratic-csi.iscsi.
```

For example, we had to manually add the following annotation to make the StorageClass pass validation:

```
storageClasses:
- name: truenas-iscsi
  defaultClass: false
  reclaimPolicy: Delete
  volumeBindingMode: Immediate
  allowVolumeExpansion: true
  annotations:
    cdi.harvesterhci.io/storageProfileVolumeModeAccessModes: '{"Block":["ReadWriteMany"]}'
  parameters:
    fsType: xfs
  mountOptions: []
```

While this workaround unblocks the installation, the current behavior makes Helm installs fail because the webhook blocks the StorageClass creation before the user is even aware of this requirement.

### Describe the solution you'd like

Instead of blocking storage class creation:

- Consider allowing the StorageClass to be created without the CDI annotation and surface a warning in the Harvester UI (under Storage > StorageClasses > CDI Settings).

- Alternatively, clearly document in [CDI Settings docs](https://docs.harvesterhci.io/v1.6/advanced/storageclass?utm_source=chatgpt.com#containerized-data-importer-cdi-settings)
 that custom CSI drivers need this annotation, and provide example values (like the one above).

ihcsim:
Honestly, if I can get away with it, I will just make this a documentation item. Most (if not all) CSI Helm charts allow user to insert annotations into their storage classes during installation, including [democratic-csi](https://github.com/democratic-csi/charts/blob/91e1c02acbea399700a277bbdd1f9fb5e7833b17/stable/democratic-csi/templates/storage-classes.yaml#L16-L17). If there really are charts that don't support overriding storage class annotations, one can still use commands like `helm template` and tools like `kustomize` to override them. CDI is essential for 3rd-party storage solutions to work on Harvester. By removing the webhook validation, we are only shifting the issue further down the workflow and opening up more gaps and edge cases.

[...]

PR #9127, opened 2025-09-12
Relax CDI annotation checks in StorageClass and add VM image creation validation

brandboat:
#### Problem:

When trying to install a CSI driver (e.g. democratic-csi for TrueNAS iSCSI), the admission webhook validator.harvesterhci.io blocks the StorageClass creation if the required CDI annotation is missing:
```
admission webhook "validator.harvesterhci.io" denied the request: 
missing annotation cdi.harvesterhci.io/storageProfileVolumeModeAccessModes. 
volume access modes are required for CDI integration to work with storage class provisioner org.democratic-csi.iscsi.
```

For example, we had to manually add the following annotation to make the StorageClass pass validation:
```
storageClasses:
- name: truenas-iscsi
  defaultClass: false
  reclaimPolicy: Delete
  volumeBindingMode: Immediate
  allowVolumeExpansion: true
  annotations:
    cdi.harvesterhci.io/storageProfileVolumeModeAccessModes: '{"Block":["ReadWriteMany"]}'
  parameters:
    fsType: xfs
  mountOptions: []
```
While this workaround unblocks the installation, the current behavior makes Helm installs fail because the webhook blocks the StorageClass creation before the user is even aware of this requirement.

#### Solution:

Relax webhook validation for CDI annotations in storage class, and add volume mode access mode validator in vm image creation.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9060

#### Test plan:
test case 1: create/update a storage class that the provisioner is not in the [cdi storagecapabilities](https://github.com/kubevirt/containerized-data-importer/blob/99bcc36a931be02418192135dff7776f8a74471f/pkg/storagecapabilities/storagecapabilities.go#L37-L145)), make sure validator doesn't complain.
example:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: foo
provisioner: foo.csi.k8s.io
```

test case 2: create a sample storage class that the provisioner is not in the [cdi storagecapabilities], and create a vm image that use the sample sc, validator should return error "no default volume mode or access mode, please specify ..."
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: sample-storageclass
provisioner: kubernetes.io/no-provisioner
```
image

test case 3: create vm image with longhorn v1, v2, lvm, and 3-rd storage class that are inside [storagecapibilities](https://github.com/kubevirt/containerized-data-importer/blob/99bcc36a931be02418192135dff7776f8a74471f/pkg/storagecapabilities/storagecapabilities.go#L37-L145), and vm image should create successfully 
e.g. trident.
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: trident
provisioner: csi.trident.netapp.io
parameters:
  backendType: ontap-nas
```
Note: There’s no need to install Trident in Kubernetes, since our goal is only to verify that the VM image creation validator works as expected.


WebberHuang1118 (commented):
@brandboat

Since this PR will skip the CDI annotation check during StorageClass creation to avoid CSI driver installation failures, the storage provider may encounter PVC creation failures due to missing annotations. Should we provide a hint or warning to users, letting them know about the necessary steps to apply before provisioning volumes?

brandboat:
> Should we provide a hint or warning to users, letting them know about the necessary steps to apply before provisioning volumes?

Perhaps we can add a hint in the UI? But currently harvester UI don't support create 3-rd party storage class.

WebberHuang1118:
> > Should we provide a hint or warning to users, letting them know about the necessary steps to apply before provisioning volumes?
> 
> Perhaps we can add a hint in the UI? But currently harvester UI don't support create 3-rd party storage class.

Maybe the backend could handle this situation when a PVC is created from a third-party storage provider that lacks the necessary CSI annotation. cc @Vicente-Cheng

pkg/image/cdi/validate.go
```diff
@@ -67,3 +87,30 @@ func (cv *Validator) Update(oldVMImg, newVMImg *harvesterv1.VirtualMachineImage)
...
+		}
+		if _, found := cdicaps.GetCapabilities(cv.client, sc); !found {
+			return werror.NewInvalidError(
+				fmt.Sprintf("no default volume mode or access mode, please specify %s annotation in storage class %q",
+					util.AnnotationStorageProfileVolumeModeAccessModes, vmImg.Spec.TargetStorageClassName),
+				"spec.targetStorageClassName")
+		}
+	}
+
+	return nil
```
WebberHuang1118: suggest removing the nested if statement:
```suggestion
	if len(sp.Spec.ClaimPropertySets) != 0 {
		return nil
	}

	sc, err := cv.scCache.Get(vmImg.Spec.TargetStorageClassName)
	if err != nil {
		return werror.NewInvalidError(fmt.Sprintf("failed to get storage class %q: %v", vmImg.Spec.TargetStorageClassName, err), "spec.targetStorageClassName")
	}
	if _, found := cdicaps.GetCapabilities(cv.client, sc); !found {
		return werror.NewInvalidError(
			fmt.Sprintf("no default volume mode or access mode, please specify %s annotation in storage class %q",
				util.AnnotationStorageProfileVolumeModeAccessModes, vmImg.Spec.TargetStorageClassName),
			"spec.targetStorageClassName")
	}

	return nil
```

pkg/image/cdi/validate.go
```diff
@@ -67,3 +87,30 @@ func (cv *Validator) Update(oldVMImg, newVMImg *harvesterv1.VirtualMachineImage)
...
+		return werror.NewInvalidError("targetStorageClassName is required", "spec.targetStorageClassName")
+	}
+
+	// storage profile name is always the same as storage class name
+	sp, err := cv.spCache.Get(vmImg.Spec.TargetStorageClassName)
+	if err != nil {
+		return werror.NewInvalidError(fmt.Sprintf("failed to get storage profile %q: %v", vmImg.Spec.TargetStorageClassName, err), "spec.targetStorageClassName")
+	}
+
+	if len(sp.Spec.ClaimPropertySets) == 0 {
```
WebberHuang1118: Question: Does `len(sp.Spec.ClaimPropertySets) != 0` imply that CDI has sufficient information about the respective StorageClass?

WebberHuang1118 (commented):
LGTM—just one question and a minor suggestion. Thanks.

pkg/image/cdi/validate.go
brandboat: That means it have at least one volume mode access modes pair.

WebberHuang1118: Does Harvester need to check the pair content of the mode, or does CDI already handle this check?

brandboat: https://github.com/kubevirt/containerized-data-importer-api/blob/main/pkg/apis/core/v1beta1/types.go#L470-L480

yes, cdi already handled that.

WebberHuang1118 (approved):

brandboat:
as offline discussed, we will file a doc pr to warn user to add cdi annotations instead.

Issue #8769, opened 2025-07-28
[ENHANCEMENT] Review the daemonset longhorn-loop-device-cleaner


Vicente-Cheng:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

This daemonset was a workaround for [[BUG] Block device volume failed to unmount when it is detached unexpectedly](https://github.com/harvester/harvester/issues/2137#top), corresponding Longhorn issue: [[BUG] Block device volume failed to unmount when it is detached unexpectedly](https://github.com/longhorn/longhorn/issues/3778#top).

And this issue was fixed in k8s >= v1.27 (REF: https://github.com/longhorn/longhorn/issues/3778#issuecomment-1572999855).

### Describe the solution you'd like

We should retest to see if we still need this daemonset.

wheatdog:
I followed the steps in https://github.com/longhorn/longhorn/issues/3778#issuecomment-1077089425 but only changed the image in use (from `k8s.gcr.io/nginx-slim:0.8` to `registry.k8s.io/nginx-slim:0.27`) and confirmed that the behavior is correct even without `longhorn-loop-device-cleaner`.

Therefore, I'll proceed with the removal of this damonset.


# Steps
1. Prepare a 2-node Harvester cluster based on v1.6.0 with `longhorn-loop-device-cleaner` removed. (1-node Harvester cluster should also be fine)
2. Confirm that there is no running `longhorn-loop-device-cleaner` pods
```
$ kubectl get pods -A | grep longhorn-loop-device-cleaner | wc -l
0
```
3. Apply the manifest
```
$ cat << EOF > /tmp/manifest.yaml
apiVersion: v1
kind: Service
metadata:
  name: test-bv
  labels:
    app: test-bv
spec:
  ports:
  - port: 80
    name: test-bv
  selector:
    app: test-bv
  type: NodePort
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: test-bv
  namespace: default
spec:
  selector:
    matchLabels:
      app: test-bv
  serviceName: "test-bv"
  template:
    metadata:
      labels:
        app: test-bv
    spec:
      restartPolicy: Always
      terminationGracePeriodSeconds: 10
      containers:
      - name: test-bv
        image: registry.k8s.io/nginx-slim:0.27
        command:
        - sleep
        - infinity
        ports:
        - containerPort: 80
          name: test-bv
        volumeDevices:
[...]

PR #9128, opened 2025-09-12
charts: remove longhorn-loop-device-cleaner due to upstream fix

wheatdog:
#### Problem:

Remove the daemonset longhorn-loop-device-cleaner since this workaround is no longer needed. It's resolved by upstream.

#### Solution:

Remove the manifest of longhorn-loop-device-cleaner.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8769

#### Test plan:

Follow https://github.com/harvester/harvester/issues/8769#issuecomment-3284939497


wheatdog:
@FrankYang0529 @Vicente-Cheng Would you help me take a look? Thanks!

FrankYang0529 (approved):

Vicente-Cheng:
Sorry for the late review.
Thanks for the testing, and remove the out-of-date service.

Issue #9033, opened 2025-09-03
[BUG] Network setup error during upgrade from v1.5.1 to v1.6.0


W0n9:
### Describe the Bug

In upgrade progress, I found that some nodes stuck at `Waiting Reboot`, and it lost its network.
When I drill down into the log of `element-setup-network.service`, I found it raise an error about config file parsing.
But I have never edit this file, and it was parsed succeessfully at element v1.1.6.

<img width="1024" height="768" alt="Image" src="https://github.com/user-attachments/assets/e55cca0f-b8f2-4828-a3be-359d3d313651" />

![Image](https://github.com/user-attachments/assets/79182bed-2fdb-453e-8ce4-62d0ec8f5b7d)

### To Reproduce

Upgrade from v1.5.1 to v1.6.0

### Expected Behavior

Upgrade successfully

### Support Bundle for Troubleshooting

no support bundle file

### Environment

- Harvester version: v1.6.0
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Baremetal with Dell PowerEdge R730
- Rancher version: v2.12.1

### Workaround and Mitigation

Edit the /oem/99_custom.yaml manually

w13915984028:
@W0n9 Could you take a picture to show the full content of `setup_bond.sh`? And did you have the `upgrade log`?

and also run `yq -e /oem/90_custom.yaml` under the linux shell, thanks.

And, in your cluster, did you set `mgmt vid` before, or mgmt was in un-tag mode (when install first node, did not set vid)? 

From the upgrade script, the `sed` operation should not break the yaml format in general.

https://github.com/harvester/harvester/blob/b85b952b4936632595e57df082e37bd745d5ec57/package/upgrade/upgrade_node.sh#L570-L578

cc @rrajendran17

W0n9:
Hi @w13915984028 
Here is full content of `setup_bond.sh`
```yaml
 - path: /etc/wicked/scripts/setup_bond.sh
      permissions: 493
      owner: 0
      group: 0
      content: "#!/bin/sh\n\nACTION=$1\nINTERFACE=$2\n\ncase $ACTION in\n\tpost-up)\n\t\t#
        inherit MAC address\n\t\tip link set dev mgmt-br address $(ip -json link show
        dev $INTERFACE | jq -j '.[0][\"address\"]')\n\n\t\t# accept all vlan, PVID=1
      encoding: ""
[...]

PR #8939, opened 2025-08-20
Remove unused legacy upgrade code

w13915984028:
#### Problem:

Some legacy codes are out-dated.

#### Solution:

**This PR targets for v170**, in v1.6.x->v1.7.0 upgrade path, those unused legacy code (esp. they are called on specific previous version like v1.0.3, v1.5.x) will get removed.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8938

#### Test plan:

1. Install v160 cluster
2. Upgrade to master-head, it should not block

#### Additional documentation or context

note: Harvester upgrade is controlled in defined paths, e.g. v1.3.x->v1.4.x->v1.5.x->v1.6.x->v1.7.x, you can't ugprade from v1.4.x to v1.7.x directly.

PR #9129, opened 2025-09-12
Fix bridge vlan removal for both yaml-v2 and yam-v3 format

rrajendran17:
#### Problem:
https://github.com/harvester/harvester/blob/b85b952b4936632595e57df082e37bd745d5ec57/package/upgrade/upgrade_node.sh#L578 assumed the contents of /oem/90_custom.yaml to be in format as generated from yaml/v3 but if the contents are generated from old harvester-installer, then it is generated using yaml/v2 and contents looks different and the sed operation above can truncate the full line during upgrade from v1.5.x to v1.6.x when mgmt interface is installed with no vid.

yaml/v2 generated /oem/90_custom.yaml

image

image

yaml/v3 generated /oem/90_custom.yaml

```
 - path: /etc/wicked/scripts/setup_bond.sh
              permissions: 493
              owner: 0
              group: 0
              content: |
                #!/bin/sh

                ACTION=$1
                INTERFACE=$2

                case $ACTION in
                        post-up)
                                # inherit MAC address
                                ip link set dev mgmt-bo address $(ip -json link show dev $INTERFACE | jq -j '.[0]["address"]')

                                #skip bridge vlan setting when no vlan id or vlan id=1 specified by user
                                if [ 2021 -eq 0 ] || [ 2021 -eq 1 ]; then
                                    exit 0
                                fi
                                #assign user configured vlan,PVID=1 by default
                                bridge vlan add vid 2021 dev $INTERFACE
                                ;;

                esac
```
There could be cases where older nodes were installed using harvester-installer version (v1.1.2) and harvester version probably v1.2.x and later upgraded to v1.3.x->v1.4.x->v1.5.x, during the upgrade process the harvester-installer versions would have upgraded to newer versions, but the contents of the /oem/90_custom.yaml never changed as we do not update the contents of /oem/90_custom.yaml file during upgrade until v1.5.x.Due to this there is possibility that Harvester nodes could still have 90_custom.yaml using contents generated from old yaml/v2 format.This will cause issue when user upgrades from v1.5.x to v1.6.x and mgmt interface is installed with no vlan id.

#### Solution:
Handle removal of command "bridge vlan add vid 2-4094 ....." for both yaml/v2 and yaml/v3 format

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9033

#### Test plan:
case 1:
1.Harvester node installed v1.5.x on mgmt interface with no vid
2.Contents of /oem/90_custom.yaml in yaml/v2 format generated from older harvester-installer version (or copy older contents to this 90_custom.yaml)
2.Upgrade from v1.5.x to v1.6.x
3.Check if upgrade is successful

case 2:
1.Harvester node installed v1.5.x on mgmt interface with no vid
2.Upgrade from v1.5.x to v1.6.x
3.Check if upgrade is successful

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -575,7 +575,10 @@ replace_with_mgmtvlan() {
   else
     echo "VLAN ID: $vlan_id remove bridge vlan"
     sed -i "s/accept all vlan, PVID=1 by default/PVID=1 by default/" $CUSTOM90_FILE
-    sed -i "/bridge vlan add vid/d" $CUSTOM90_FILE
+    # match contents of bridge vlan add vid 2-4094 using yaml/v3 format
+    sed -i '/^[[:space:]]*bridge vlan add vid 2-4094 dev \(\$INTERFACE\|mgmt-bo\)\( self\)\?[[:space:]]*$/d' $CUSTOM90_FILE
+    # match contents of bridge vlan add vid 2-4094 using yaml/v2 format (escape characters and command spanning multiple lines)
+    perl -0777 -i -pe 's/bridge vlan add vid 2-4094.*?(?:\\n|\\t)+//gs' $CUSTOM90_FILE
   fi
 }
 


w13915984028 (approved):
LGTM, thanks, 
please also double confirm those questions:
(1) Is the code just needed on v1.6 branch or also on master?
(2) Is `perl` packed in the upgrade container image by default?
(3) It looks necessary to add the issue to `known issues` on v160 upgrade via another doc PR.
https://docs.harvesterhci.io/v1.6/upgrade/v1-5-0-to-v1-6-0#known-issues

Vicente-Cheng (approved):
Thanks for the quick fix. (perl messes my head, but verified with the problematic configuration)

rrajendran17:
> LGTM, thanks, please also double confirm those questions: (1) Is the code just needed on v1.6 branch or also on master? (2) Is `perl` packed in the upgrade container image by default? (3) It looks necessary to add the issue to `known issues` on v160 upgrade via another doc PR. https://docs.harvesterhci.io/v1.6/upgrade/v1-5-0-to-v1-6-0#known-issues

1.This code is needed only for v1.6, https://github.com/harvester/harvester/pull/8939 should take care of removing the changes from master.
2.yes verified the fix with upgrade from v1.5.x to v1.6.x (mgmt with vid 1) with contents generated from yaml.v2 and upgrade completed successfully with perl commands removing the bridge commands from v1.5.x
3.https://github.com/harvester/docs/pull/880 - document changes

rrajendran17:
Verified the changes by upgrading from v1.5.1 to custom image (master) with fix and upgrade completed successfully replacing the contents of /oem/90_custom.yaml using the perl command.

Before upgrade:
```
 - path: /etc/wicked/scripts/setup_bond.sh
              permissions: 493
              owner: 0
              group: 0
              content:
                "#!/bin/sh\n\nACTION=$1\nINTERFACE=$2\n\ncase $ACTION in\n\tpost-up)\n\t\t#
                inherit MAC address\n\t\tip link set dev mgmt-br address $(ip -json link show
                dev $INTERFACE | jq -j '.[0][\"address\"]')\n\n\t\t# accept all vlan, PVID=1
                by default\n\t\tbridge vlan add vid 2-4094 dev $INTERFACE\n\t\t;;\n\nesac\n"
```

```
 - path: /etc/wicked/scripts/setup_bridge.sh
              permissions: 493
              owner: 0
              group: 0
              content:
                "#!/bin/sh\n\nACTION=$1\nINTERFACE=$2\n\ncase $ACTION in\n\tpre-up)\n\t\t#
                enable vlan-aware\n\t\tip link set dev $INTERFACE type bridge vlan_filtering 1\n\t\t\t;;\n\n\tport-up)\n\t\t#
                accept all vlan, PVID=1 by default\n\t\tbridge vlan add vid 2-4094 dev $INTERFACE
                self\n\t\tbridge vlan add vid 2-4094 dev mgmt-bo\n\t\t;;\n\nesac\n"
```

After upgrade:
``` - path: /etc/wicked/scripts/setup_bond.sh
              permissions: 493
              owner: 0
              group: 0
              content:
                "#!/bin/sh\n\nACTION=$1\nINTERFACE=$2\n\ncase $ACTION in\n\tpost-up)\n\t\t#
                inherit MAC address\n\t\tip link set dev mgmt-br address $(ip -json link show
                dev $INTERFACE | jq -j '.[0][\"address\"]')\n\n\t\t# accept all vlan, PVID=1
                by default\n\t\t;;\n\nesac\n"
```

```  - path: /etc/wicked/scripts/setup_bridge.sh
              permissions: 493
              owner: 0
              group: 0
              content:
                "#!/bin/sh\n\nACTION=$1\nINTERFACE=$2\n\ncase $ACTION in\n\tpre-up)\n\t\t#
                enable vlan-aware\n\t\tip link set dev $INTERFACE type bridge vlan_filtering 1\n\t\t\t;;\n\n\tport-up)\n\t\t#
                PVID=1 by default\n\t\t;;\n\nesac\n"
```

rrajendran17:
@Mergifyio backport v1.6

ihcsim:
Thanks for the fix. Sorry for the late post, for posterity, instead of using `sed` on the entire file, it might be safer (if possible) to use `yq` to identify the relevant blocks, clear and replace the content with what we want it to be.

w13915984028:
BTW: Need to be very careful about using `yq` to operate (write) a persistent yaml file in the shell script, it can silently break things like stripping blank lines, use a different indent. The `oem` file itself is also very complex.

https://github.com/harvester/addons/blob/6726254bc209ad6a5802ddf180a1044da5c0e1c3/scripts/hack/patch-rancher-monitoring#L36-L40

another example to operate the `multi line content` field on yaml file:

https://github.com/harvester/harvester/blob/6cacbe6f029b5294784ac308f8c521f0f051a00b/package/upgrade/patch_99_custom.sh#L40-L136

Issue #9045, opened 2025-09-04
[TASK] Should not disable auto-cleanup-system-generated-snapshot during upgrade


Vicente-Cheng:
### Task request

Currently, we disable the longhorn settings `auto-cleanup-system-generated-snapshot` during upgrade to avoid the https://github.com/longhorn/longhorn/issues/10495.

This issue was resolved with Longhorn v1.9.0. We should not disable this parameter to ensure we could clean up the system snapshot during the upgrade. That would help us reduce the extra disk usage by the system snapshot.

cc @WebberHuang1118, @derekbit, @innobead

m-ildefons:
## Validation

- <details><summary>🟢 Harvester v1.6.1-rc1 (single-node, ipxe-examples based)</summary><img width="3819" height="1991" alt="Image" src="https://github.com/user-attachments/assets/a759f19d-3ac0-44f9-9811-d6a1c656e651" /></details>
- <details><summary>🟢 Confirm Longhorn >= v1.9.0</summary><img width="1129" height="173" alt="Image" src="https://github.com/user-attachments/assets/7ef15942-6ec1-4500-ba38-4adf49fff87a" /></details>
- <details><summary>🟢 Confirm Setting `auto-cleanup-system-generated-snapshot` has value `"true"`</summary><img width="1182" height="411" alt="Image" src="https://github.com/user-attachments/assets/433933f5-13d3-4916-9475-8de2b64b296c" /></details>
- <details><summary>🟢 Create custom `version.harvesterhci.io` resource for v1.7.0-dev-20250928</summary><img width="1322" height="200" alt="Image" src="https://github.com/user-attachments/assets/b15efe00-6245-46b3-a545-bffacab97c47" /></details>
- <details><summary>🟢 Initiate upgrade to v1.7.0-dev-20250928</summary><img width="1306" height="839" alt="Image" src="https://github.com/user-attachments/assets/a8384563-f704-47f1-ab1c-e4204266a8bd" /></details>
- <details><summary>🫖 Make cup of tea while the upgrade progresses. This may take some time.</summary><img width="920" height="1142" alt="Image" src="https://github.com/user-attachments/assets/0627cacf-9459-4d41-aff6-23496598af0f" /></details>
- <details><summary>🔴 Upgrade failed to progress after more than an hour</summary><img width="897" height="1138" alt="Image" src="https://github.com/user-attachments/assets/0ccb2df8-c8df-4fa7-86e0-6a36ef42d6a7" /></details>

[...]

PR #9136, opened 2025-09-15
upgrade: do not disable the auto cleanup system snapshot

Vicente-Cheng:
we disable this capability
    `auto-cleanup-system-generated-snapshot` because of the
    corresponding longhorn issue:
    https://github.com/longhorn/longhorn/issues/10495.

    Since this issue was fixed on v1.9.0, we should enable the
    auto cleanup for system snapshot to reduce the disk pressure
    with system snapshot

#### Problem:
Since this issue was fixed on v1.9.0, we should enable the auto cleanup for system snapshot to reduce the disk pressure with system snapshot

#### Solution:
Ensure the `auto-cleanup-system-generated-snapshot` would not be disabled during upgrade.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9045

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -356,22 +356,6 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 				}
 			}
 
-			// Disable auto-cleanup-system-generated-snapshot to avoid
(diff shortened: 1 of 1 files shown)


Vicente-Cheng:
@mergifyio backport v1.6

starbops (approved):
LGTM, thank you.

tserong (approved):

innobead (approved):

Issue #2311, opened 2022-05-24
[FEATURE] Auto Live Migration (Rebalancing VM workloads)


janeczku:
**Is your feature request related to a problem? Please describe.**
TODO

unnamedjk:
There is a decent script i have used with Proxmox, its similar to the weight system that VMware DRS uses.
https://github.com/kcl-nmssys/python-proxmoxbalancer/blob/master/proxmoxbalancer/proxmoxbalancer.py

iwilltry42:
I guess this is about rebalancing the cluster in cases of high load on a node?
KubeVirt proposes using the Descheduler for that :thinking: 
https://kubevirt.io/user-guide/operations/node_assignment/#node-balancing-with-descheduler
I'm definitely following this issue :)

abonillabeeche:
We need this one guys - top 5 of requests to allow the dynamic movement of certain VMs based on utilization.

asc5543:
# Enable/Disable descheduler add-ons, and upgrade test
## Test result
Test Fail, the `descheduler` add-ons is still disabled after upgrade, even it is enabled before upgrade

## Environment
* Harvester ISO version: `v1.7.0-rc3`
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): `2 node via ipxe-example`

## Test Scenario
### Preparation
Prepare the version.yaml for upgrade

1. Prepare the v1.7.0-rc3 ISO file
1. <details> <summary> Prepare the version.yaml file </summary>

    ```
    apiVersion: harvesterhci.io/v1beta1
    kind: Version
    metadata:
      name: v1.7.0-rc3
      namespace: harvester-system
    spec:
      isoChecksum: '1cbd5ef881663656c9bf0d79f46007a386d0f2f6d180791864e8d7e50e3216229599ba2636cbc51eebe50df3829dfe8923d63a519ecb5f807ad710374afc135a'
      isoURL: http://<lab_url>/harvester-v1.7.0-rc3-amd64.iso
      releaseDate: '20251121'
    ```

   </details>

1. <details> <summary> Create a version object. </summary>

    ```
    kubectl create -f version.yaml
    version.harvesterhci.io/v1.7.0-rc3 created
    ```

   </details>

### Test Steps
1. <details>
   <summary> ✔ On Add-ons page, enable <b>descheduler</b>, check the add-on can be enable
   </summary>

[...]

PR #9137, opened 2025-09-15
feat: add descheduler HEP [SKIP CI]

FrankYang0529:
#### Related Issue(s):

### DIFF
--- a/enhancements/20250910-descheduler.md
+++ b/enhancements/20250910-descheduler.md
@@ -1,0 +1,64 @@
+# Descheduler
+
+## Summary
+
+The Descheduler is a Kubernetes component that helps maintain optimal pod distribution across nodes by evicting pods that violate certain policies. By integrating the Descheduler into Harvester, we can enhance the cluster's resilience and performance, ensuring that workloads are balanced and resources are utilized efficiently.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/2311
+
+## Motivation
+
+### Goals
+
+- Integrate the Descheduler into Harvester to improve pod distribution and resource utilization.
+- VM pods can be evicted.
+- User can enable or disable the Descheduler as needed.
+- User can configure the node utilization threshold for the Descheduler.
+- Support DefaultEvictor and LowNodeUtilization strategies.
+
+### Non-goals [optional]
+
+- Support more strategies other than DefaultEvictor and LowNodeUtilization.
+
+## Proposal
+
+### User Stories
+
+- User finds that some nodes are overutilized while others are underutilized, leading to performance bottlenecks.
+- By enabling the Descheduler, the system can automatically redistribute pods across nodes, improving overall cluster performance and resource utilization.
+
+### User Experience In Detail
+
+1. Enable the Descheduler via Harvester UI.
+2. Configure the node utilization threshold for the Descheduler.
+3. Select included and excluded namespaces for the Descheduler.
+4. Monitor the Descheduler's activity and the resulting pod distribution across nodes.
+
+### API changes
+
+N/A
+
+## Design
+
+### Implementation Overview
+
+- The Descheduler will be added as an AddOn in Harvester.
+- When user enables the Descheduler, the AddOn controller will create a Descheduler deployment in the `kube-system` namespace.
+
+### Test plan
+
+- Create a 2-node Harvester cluster.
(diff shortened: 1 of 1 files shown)


enhancements/20250910-descheduler.md
```diff
@@ -0,0 +1,64 @@
...
+
+- The Descheduler will be added as an AddOn in Harvester.
+- When user enables the Descheduler, the AddOn controller will create a Descheduler deployment in the `kube-system` namespace.
+
+### Test plan
+
+- Create a 2-node Harvester cluster.
+- Create several VMs and migrate them to the control-plane node. Don't select the node selector on VMs.
+- Enable the Descheduler and set the node utilization threshold to 20%.
+- Verify that some VMs are evicted from the control-plane node and scheduled to the worker node.
```
w13915984028, line 55: two questions, thanks.

(1) in such case, is the VM migrated to other node, or simply stopped & started ?

(2) how does this feature collaborate with the upgrade or node maintenance case, or the `migrate` action where use user can set the target node?

FrankYang0529: > (1) in such case, is the VM migrated to other node, or simply stopped & started ?

This is determined by `evictionStrategy`. By default, it's `LiveMigrate`, so the VM is migrated to another node.

> (2) how does this feature collaborate with the upgrade or node maintenance case, or the migrate action where use user can set the target node?

Like the above question, the behavior is defined by `evictionStrategy`.

ibrokethecloud:
Found this runbook in kubevirt documentation which is for leveraging openshift kube deschedular operator
https://kubevirt.io/monitoring/runbooks/HCOMisconfiguredDescheduler.html

This recommends the extra profileCustomization

```
spec:
   profileCustomizations:
      devEnableEvictionsInBackground: true
```

Based on operator code this customization maps to 

https://github.com/openshift/cluster-kube-descheduler-operator/blob/main/pkg/operator/target_config_reconciler.go#L1552-L1561

This is in turn enabling the descheduler feature gate `EvictionsInBackground`

We should ensure we leverage this when we package the addon.

FrankYang0529:
@ibrokethecloud Thanks for the suggestion. I create a PR for it. https://github.com/harvester/addons/pull/169

wheatdog:
I'll repeat the questions I have during review meeting here for the record and further discussion. Thanks.

Question 1: Is it safe to let the descheduler run while upgrading Harvester?
Question 2: Is it ok to disable the descheduler addon after enabling it? Do we need to revert or clean up some configurations? Do we need to check whether the descheduler is in the middle of moving something around before disabling it?

FrankYang0529:
> Question 1: Is it safe to let the descheduler run while upgrading Harvester?

Create PR https://github.com/harvester/harvester/pull/9389 to disable descheduler before upgrade and enable it again after upgrade.

> Question 2: Is it ok to disable the descheduler addon after enabling it? Do we need to revert or clean up some configurations? Do we need to check whether the descheduler is in the middle of moving something around before disabling it?

Yes, it's okay. It's a stateless deployment.

Vicente-Cheng (approved):
lgtm, thanks!

enhancements/20250910-descheduler.md
```diff
@@ -0,0 +1,64 @@
...
+### User Stories
+
+- User finds that some nodes are overutilized while others are underutilized, leading to performance bottlenecks.
+- By enabling the Descheduler, the system can automatically redistribute pods across nodes, improving overall cluster performance and resource utilization.
+
+### User Experience In Detail
+
+1. Enable the Descheduler via Harvester UI.
+2. Configure the node utilization threshold for the Descheduler.
+3. Select included and excluded namespaces for the Descheduler.
```
innobead, line 36: I assume harvester components and other underlying components (e.g., Longhorn, RKE2, addons, etc.) should be excluded by default and not eligible for the descheduler, correct?

@FrankYang0529 @Vicente-Cheng @ibrokethecloud

FrankYang0529: Yes, we exclude these namespaces by default.

https://github.com/harvester/addons/blob/4a4837a25aff69c55896f924897bf3ceb5ba5f89/pkg/templates/rancherd-22-addons.yaml#L389-L402

innobead: Sounds good, but don't forget namespaces for addons.

Issue #8338, opened 2025-05-22
[ENHANCEMENT] support node level driver override for nvidia-driver-toolkit


ibrokethecloud:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Current implementation of nvidia driver toolkit only allows using one version of driver across the cluster.

In scenarios where users have multiple GPU types spread across the nodes, users may need a different driver to be installed on specific node.

User should be able to override the driver on a per node basis.

### Describe the solution you'd like

Currently downward api does not allow us to fetch node level annotations. however the node name is available

https://kubernetes.io/docs/concepts/workloads/pods/downward-api/

A possible option would be to tweak the entrypoint script to query the node annotations and override the driver endpoint based during install.

Vicente-Cheng:
We could have the HEP for both #8320 and #8338.

Yu-Jack:
After getting node name, we should be able use token in `/var/run/secrets/kubernetes.io/serviceaccount/token` to call k8s API to fetch node information.

albinsun:
Test PASS, close as done.



## Environment
* `harvester-v1.7.0-rc1`
  * Profile: 3 nodes AMD64 Bare-metal HPE ProLiant DL360 
  * ui-source: `Auto`


## Test Scenario
### Prerequisites
1. <details>
   <summary> Setup 3 nodes <code>harvester-v1.7.0-rc1</code> with GPU on each nodes
   </summary>

   <img width="1566" height="828" alt="Image" src="https://github.com/user-attachments/assets/90eda01d-b6fe-4e78-a5fa-fb2fd2c769b0" />
   </details>

1. <details>
   <summary> Enable <b>pcidevices-controller</b> then <b>nvidia-driver-toolkit</b> using <code>570</code> driver
   </summary>

   * add-on
     <img width="1566" height="774" alt="Image" src="https://github.com/user-attachments/assets/35c3f705-349c-4891-9b43-4cf88eb52a5e" />

   * PCI Devices
     <img width="1566" height="480" alt="Image" src="https://github.com/user-attachments/assets/46fd5164-7add-44bf-9510-28a57f15be41" />

   * SR-IOV GPU Devices
     <img width="1566" height="480" alt="Image" src="https://github.com/user-attachments/assets/3c3ca4b3-4fd8-492b-812d-4758ec1deb7b" />

   * vGPU Devices
[...]

PR #9146, opened 2025-09-16
hep: support installing different nvidia driver per node

Yu-Jack:
#### Problem:
Users have multiple GPU types spread across the nodes, users may need a different driver to be installed on specific node.

#### Solution:
User should be able to install specific the driver on a per node.

- https://github.com/harvester/charts/pull/415
- https://github.com/harvester/os2/pull/211

#### Related Issue(s):
#8338 

- [ ] Require documentation.

### DIFF
--- a/enhancements/20250916-support-installing-specific-nvidia-driver.md
+++ b/enhancements/20250916-support-installing-specific-nvidia-driver.md
@@ -1,0 +1,97 @@
+# Support Installing Specific NVIDIA Driver
+
+Enable Harvester to install specific NVIDIA drivers per node by reading node annotations from the Kubernetes API.
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


innobead (commented):
In general, LGTM.

There are a few minor points that need to be considered later:

- Currently, when accessing HTTPS endpoints of drivers, certificate validation is just ignored. We need to improve this with a secure option for users.
- In the doc, we need to let users know that updating the annotation after enabling the addon won't trigger the NVIDIA driver reload. The affected pods need to be restarted.

ibrokethecloud (approved):
lgtm. thanks for the proposal.

innobead (approved):

Issue #8859, opened 2025-08-11
[ENHANCEMENT] Enhance UX to guide users on CPU-pinned VM creation without CPU Manager


brandboat:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Currently, if a user creates a VM with CPU Pinning enabled and there is no node with CPU Manager enabled, the VM will remain in the Unschedulable state. Although our documentation mentions this limitation, we still need to address it in the product.

We previously attempted to fix this issue in [commit](https://github.com/harvester/harvester/commit/079cee711f0d3e91966cd4e97039920e346a6c4c), which prohibited users from performing any VM operation if the VM had CPU Pinning enabled while no node had CPU Manager enabled.

However, this approach introduced another problem ([#8778](https://github.com/harvester/harvester/issues/8778)). We hadn’t considered a scenario where a user might directly update the VM YAML without restarting the VM, then disable CPU Manager. In this case, the newly added validator would block further VM operations, even though the VM was already running.

### Describe the solution you'd like

Prevent users from provisioning CPU-pinned VMs when no nodes have CPU Manager enabled.

### Describe alternatives you've considered

1. [Commit](https://github.com/harvester/harvester/commit/079cee711f0d3e91966cd4e97039920e346a6c4c) — Added a VM validator to block any VM operation if the VM has CPU Pinning enabled while no node has CPU Manager enabled.
2. [PR #8800](https://github.com/harvester/harvester/pull/8800) — Added a node validator to prevent situations where a CPU-pinned VM exists but no node has CPU Manager enabled.

1 + 2 => Having both a VM validator and a node validator feels redundant and potentially overkill.

ihcsim:
I am under the impression that k8s/kubevirt already knows how to handle erroneous dedicated cpu requests when the cpu manager isn't enabled i.e., the pod can't be scheduled due to unsatisfied [node affinity constraint](https://docs.harvesterhci.io/v1.6/vm/index#related-cpu-pinning-concepts). We just need to show the error messages and/or `events` in the UI for subsequent user intervention.

[...]

PR #9151, opened 2025-09-16
Add validator to block disabling CPU manager node with cpu pinned VMs

brandboat:
#### Problem:
Currently, if a user creates a VM with CPU Pinning enabled and there is no suitable nodes with CPU Manager enabled, the VM will remain in the "Unschedulable" state. Although our documentation mentions this limitation, we still need to address it in the product.

#### Solution:
Add validator to block disabling CPU manager node with cpu pinned VMs with the following considerations:
1. VM Binding Check: Identifies VMs with CPU pinning that are bound to the current node via
    NodeSelector constraints. These VMs cannot be rescheduled to other nodes that not meet the constraints, so disabling
    CPU Manager would make them unschedulable.
2. Last Node Protection: Prevents disabling CPU Manager on the last enabled node when any
    CPU-pinned VMs exist in the cluster, ensuring at least one node remains available for
    scheduling these VMs.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8859

#### Test plan:

Test plan1:
1. prepare a multi-node harvester cluster.
1. enable cpu pinning on one node, harvester-node-0.
1. create test vm
1. edit test vm, enable cpu pining, click save, “DO NOT” click save and restart.
1. create test2 vm, stop the vm
1. edit test2 vm, enable cpu pinning, click save.
1. disable cpu manager on harvester-node-0, you should see error as follows:
    image
1. disable cpu pinning for test, test2 vms.
1. disable cpu manager successfully.

Test plan2:
1. prepare a multi-node harvester cluster.
1. enable cpu pinning on two nodes, harvester-node-0, harvester-node-1.
1. create 2 cpu pinning vm with additional scheduling rules
  1. test1: click node scheduling > Run virtual machine on node(s) matching scheduling rules > add rule "kubernetes.io/hostname" "in" "harvester-node-1"
  2. test2: click node scheduling > Run virtual machine on specific node - (Live migration is not supported) > select "harvester-node-0"
1. stop test1 and test2.
1. disable cpu manager on harvester-node-0, you should see error as follows: `admission webhook "validator.harvesterhci.io" denied the request: Cannot disable CPU Manager on this node because the following VM(s) with CPU pinning are bound to this node via Node Scheduling rules and cannot be scheduled on other CPU Manager enabled nodes. Please remove CPU pinning or update Node Scheduling rules for these VM(s): default/test1`
1. disable cpu manager on harvester-node-1, you should see error as follows: `admission webhook "validator.harvesterhci.io" denied the request: Cannot disable CPU Manager on this node because the following VM(s) with CPU pinning are bound to this node via Node Scheduling rules and cannot be scheduled on other CPU Manager enabled nodes. Please remove CPU pinning or update Node Scheduling rules for these VM(s): default/test2`

### DIFF
--- a/pkg/util/indexeres/virtualmachine.go
+++ b/pkg/util/indexeres/virtualmachine.go
@@ -10,6 +10,9 @@ import (
 const (
 	VMByPVCIndex        = "harvesterhci.io/vm-by-pvc"
 	VMByHotplugPVCIndex = "harvesterhci.io/vm-by-hp-pvc"
+	VMByCPUPinningIndex = "harvesterhci.io/vm-by-cpu-pinning"
+
+	CPUPinningEnabled = "enabled"
 )
 
 func VMByPVC(obj *kubevirtv1.VirtualMachine) ([]string, error) {
@@ -42,6 +45,18 @@ func VMByHotplugPVC(obj *kubevirtv1.VirtualMachine) ([]string, error) {
 		if isVolumeHotplugged(volume) {
 			results = append(results, ref.Construct(obj.Namespace, volume.PersistentVolumeClaim.ClaimName))
 		}
+	}
+	return results, nil
+}
+
+func VMByCPUPinning(obj *kubevirtv1.VirtualMachine) ([]string, error) {
+	if obj == nil || obj.Spec.Template == nil {
+		return nil, nil
+	}
+
+	var results []string
+	if obj.Spec.Template.Spec.Domain.CPU != nil && obj.Spec.Template.Spec.Domain.CPU.DedicatedCPUPlacement {
+		return []string{CPUPinningEnabled}, nil
 	}
 	return results, nil
 }
--- a/pkg/webhook/resources/node/validator.go
+++ b/pkg/webhook/resources/node/validator.go
@@ -7,25 +7,29 @@ import (
 
 	ctlbatchv1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/batch/v1"
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	schedulinghelper "k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
 
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	ctlnode "github.com/harvester/harvester/pkg/controller/master/node"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
 	"github.com/harvester/harvester/pkg/util"
+	indexeresutil "github.com/harvester/harvester/pkg/util/indexeres"
 	"github.com/harvester/harvester/pkg/util/virtualmachineinstance"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
 )
 
-func NewValidator(nodeCache v1.NodeCache, jobCache ctlbatchv1.JobCache, vmiCache ctlkubevirtv1.VirtualMachineInstanceCache) types.Validator {
+func NewValidator(nodeCache v1.NodeCache, jobCache ctlbatchv1.JobCache, vmCache ctlkubevirtv1.VirtualMachineCache, vmiCache ctlkubevirtv1.VirtualMachineInstanceCache) types.Validator {
 	return &nodeValidator{
 		nodeCache: nodeCache,
 		jobCache:  jobCache,
+		vmCache:   vmCache,
 		vmiCache:  vmiCache,
 	}
 }
@@ -34,6 +38,7 @@ type nodeValidator struct {
 	types.DefaultValidator
 	nodeCache v1.NodeCache
 	jobCache  ctlbatchv1.JobCache
+	vmCache   ctlkubevirtv1.VirtualMachineCache
 	vmiCache  ctlkubevirtv1.VirtualMachineInstanceCache
 }
 
@@ -129,8 +134,12 @@ func (v *nodeValidator) validateCPUManagerOperation(node *corev1.Node) error {
 	if err := checkMasterNodeJobs(node, v.nodeCache, v.jobCache); err != nil {
 		return err
 	}
-	// check if there is any vm that enable cpu pinning while cpu manager is going to be disabled
+	// check if there is any running vmis that enable cpu pinning while cpu manager is going to be disabled
 	if err := checkCPUPinningVMIs(node, policy, v.vmiCache); err != nil {
+		return err
+	}
+	// check if there is at least one node has cpu manager enabled if there is any vm with cpu pinning
+	if err := checkCPUManagerEnabledForPinnedVMs(node, policy, v.nodeCache, v.vmCache); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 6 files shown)


brandboat:
> As for https://github.com/harvester/harvester/pull/8800#issuecomment-3180955206, I wonder if we can refactor checkCPUPinningVMIs to use the implementation in checkCPUManagerEnabledForPinnedVMs such that via the VMByCPUPinning index, we can find running VMs on the affected node, and also find any stopped VMs within the cluster. That will allow us to decide whether to allow or reject the request to disable the CPU manager in one function.

@ihcsim, regarding this comment - I tried refactoring the method, but from the VM CR I can’t determine which node the VM is on. We still need to list the VMI to know the actual node. For stopped VMs, there’s no VMI at all, so in the end we need to list both VMs and VMIs. I’m not sure if there’s a better approach — please let me know if I missed something. Thanks!

pkg/webhook/resources/node/validator.go
```diff
@@ -261,3 +269,58 @@ func getCPUManagerRunningJobNamesOnNodes(jobCache ctlbatchv1.JobCache, nodeNames
 	}
 	return jobNames, nil
 }
+
+// checkCPUManagerEnabledForPinnedVMs validates that at least one node maintains CPU Manager enabled
```
w13915984028: A question: do you consider the case that:

(1) When CPU pinning VM is started, but there is no node supporting it, the VM is just left unscheduable;  later if any node meets the requirement enables the CPU manager, then the VM has chance to be re-scheduled;

this behaviour meets k8s general principle

it is friendly to API based orchestrator, it deploys things in it's sequences and finally everything works; no strong code sequences are binded.

(2) When any VMI with CPU pinning is running on a specific node, this node can't disable CPU manager, the webhook denies it

The current PR seems not able to handle the case: the VM has both CPU manager and other selector, and the protected last node with CPU manager is not it's target, the VM will still be in unscheduable status.

brandboat: Thanks for the comment @w13915984028!

(1): I agree with your point, and I also acknowledge that we can’t cover every case. Still, I think we should do our best to provide more useful information about the consequences of disabling the CPU manager when there are VMs/VMIs with CPU pinning. While documentation can warn users not to do this, the “unscheduled” VM status isn’t very explicit. If we can block the action and show an error message instead, it would make the situation much clearer to the user.

(2): Thanks, I'll update the PR.

brandboat: All comments are addressed, @w13915984028, please take another look when you are available. Thank you!

pkg/util/indexeres/virtualmachine.go
```diff
@@ -45,3 +48,15 @@ func VMByHotplugPVC(obj *kubevirtv1.VirtualMachine) ([]string, error) {
...
+func VMByCPUPinning(obj *kubevirtv1.VirtualMachine) ([]string, error) {
+	if obj == nil || obj.Spec.Template == nil {
+		return nil, nil
+	}
+
+	var results []string
+	if obj.Spec.Template.Spec.Domain.CPU != nil && obj.Spec.Template.Spec.Domain.CPU.DedicatedCPUPlacement {
+		return []string{CPUPinningEnabled}, nil
+	}
+	return results, nil
```
w13915984028, line 61: `return []string{}, nil` and remove L57

pkg/webhook/resources/node/validator.go
w13915984028, line 257: Could we also optimize L257 to output this VM's namespace/name for better handling?

pkg/webhook/resources/node/validator.go
```diff
@@ -261,3 +271,143 @@ func getCPUManagerRunningJobNamesOnNodes(jobCache ctlbatchv1.JobCache, nodeNames
...
+	cpuPinnedVMs, err := vmCache.GetByIndex(indexeresutil.VMByCPUPinningIndex, indexeresutil.CPUPinningEnabled)
+	if err != nil {
+		return werror.NewInternalError(err.Error())
+	}
+
+	if len(cpuPinnedVMs) == 0 {
+		return nil
+	}
+
+	// Check if VMs are bound to this specific node via NodeSelector
```
w13915984028, line 314: It becomes a bit complex in following case:

Suppose a node with CPU pin enabled and some VMs are bound to this VM, now user suspects the node has potential issues on underlayer and plans to do a `stop` and `start` CPU pin to check if it can recover. 

User stops all those affected VMs, and then click `disable` CPU manager. Per new validator, it is not allowed, unless user changes VMs' spec to remove CPU pin first, and then `disable` and `enable` CPU manager again, and finally amend VM's spec.

w13915984028 (commented):
The last concern, thanks.

pkg/webhook/resources/node/validator.go
brandboat: Hmm... if that’s the case, we probably shouldn’t proceed with merging this PR.

Yu-Jack: I think when there is VM spec that requests the CPU pinning (not vmi), either preventing users from disabling the CPU manager or letting users be able to disable the CPU manager both are reasonable.  But, I think the point is that we should let user know what they're doing. If they forget to remove the cpu pinning in the VM spec, and something bad happens, they eventually raise a ticket to ask us to block this case. If there are not users complaining one of these cases, we can try cooper's idea first. Or, keep it as is until someone complains.

brandboat: @Vicente-Cheng, I think what @w13915984028 mentioned makes sense. Perhaps we can move this issue to the ice-box for now. I also don’t think my current approach is the right direction. The original proposal is to try out best to prevent user from encountering the unschedulable VM experience. While adding more validator here might just introduce unnecessary complexity and make user feel troublesome in cases like Jian mention above.

And we’ve already covered the unscheduable VM multiple times in the Harvester documentation, perhaps that's enough.

Vicente-Cheng: I agree with @Yu-Jack's opinion above.

>If they forget to remove the cpu pinning in the VM spec, and something bad happens, they eventually raise a ticket to ask us to block this case.

That's also my concern. However, I would like to know more about the details, specifically when the user did not remove the CPU pinning in the VM spec but instead disabled the CPU manager. What happens after they start VM w/o a CPU manager? If there's no harm, we could drop the validator if any side effects occur. We should consider the validator depending on the result.

brandboat: > what happens after they start VM w/o a CPU manager?

The vm will stuck in "unschedulable" state forever if there is no suitable node. And this is mentioned in doc
https://docs.harvesterhci.io/v1.6/vm/cpu-pinning#enable-cpu-pinning-on-a-new-vm

> important
> Ensure that at least one node has CPU Manager enabled. Otherwise, the VM will be stuck in the unschedulable state after startup. For more information, see [Node Affinity Rules](https://docs.harvesterhci.io/v1.6/vm/index#related-cpu-pinning-concepts).

w13915984028: the `unscheduable` is more k8s way, when any condition is not met, the pod remains on this state, and k8s controller will reconcille it, later if the condition is met, the pod get scheduled

this benefits to things like place-holder, when different users operate on VM and NODE in parallel, they don't need to wait for each other

and, like before mentioned, even with this PR, some VMs can still stuck on unscheduable, as VM can have CPU pin and other scheduler condtions

what we can further optimize, it to generate addtional annotation on POD, when k8s can't schedule it, the annotation explains the more detailed reason of unscheduable (e.g. target node has no cpu manager enabled), and UI shows this if it is existing

w13915984028: validator upon a `stopped` VM, might bring side effects of blocking user to do some further maintenance related operations

VM is not scheduable brings few harm, it can be checked and explaind;   but blocking other operations, will take more efforts to recover

w13915984028: BTW, this PR is a very good implementation, thanks @brandboat .

Our discussion is more about the balance between `user-firendly` and `controller-friendly`.

Issue #8391, opened 2025-05-30
[ENHANCEMENT] Harvester Node Driver labels for machine pools


m-ildefons:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

The Harvester node driver for Rancher adds a label with the cluster name to VMs, which belong to a cluster in Rancher.
It also adds a label with the name of the node pool the VM belongs to, but for some reason it only does so when an (anti-) affinity is set.

The labels are useful when dealing with load balancers, since they provide an identification mechanism for which VM is to be used as backend by the load balancer. However at the moment, if no affinity is set, it is only possible to select all or no VMs of a cluster as backends for the load balancer. This is insufficiently granular.
If the machine pool labels are added regardless of the affinity, then the load balancer can select VMs only belonging to a specific machine pool of a cluster as backends.

### Describe the solution you'd like

The Harvester node driver should add the machine pool label in all cases to enable the load balancer to select the VMs for the backends more granularly.

### Additional context

SURE-9641

### Tasks

- [x] Add missing setter methods to VMBuilder (https://github.com/harvester/harvester/pull/9153)
- [ ] Add command line flag to node driver (https://github.com/harvester/docker-machine-driver-harvester/pull/83)
- [ ] Add corresponding settings to Rancher (https://github.com/rancher/dashboard/issues/16018)

innobead:
@m-ildefons, as you are working on this, assigned it to you (next time, you can self-assign). Added this to the current sprint and 1.6.0, although it's independent from the Harvester release. Please also move this to the right status. 

cc @harvester/release @Vicente-Cheng

Vicente-Cheng:
Moved to the next version.
It seems we need more time for this enhancement.

Vicente-Cheng:
Hi @m-ildefons, 
For the UI part, could you help create the corresponding Rancher UI issue (on here https://github.com/rancher/dashboard/issues) and assign it to @a110605 first?

Thanks

innobead:
This issue has been postponed several times. Let's ensure this is ready in 1.8. 

cc @harvester/tpm

Vicente-Cheng:
We still missed some related implementation here.
Need to move to the next milestone. 
@m-ildefons could we prioritize this issue on the next milestone?

Vicente-Cheng:
We are wait for the https://github.com/rancher/dashboard/issues/16018, moved to v1.10.0

PR #9153, opened 2025-09-16
VM builder: add setter methods for two fields

m-ildefons:
Add setter methods for two more fields:
- DedicatedCPUPlacement
- IsolateEmulatorThread and tests

#### Problem:

The VMBuilder is a factory module that makes it easy to generate a `VirtualMachine` structure in Golang programs. It is used e.g. in the node driver.

The lack of some methods in the VMBuilder leads to an accumulation of fragile custom logic in the various repositories that consume the VMBuilder.

#### Solution:

Add missing methods and ensure complete feature coverage of the VMBuilder library.

#### Related Issue(s):

related-to: https://github.com/harvester/harvester/issues/8391

#### Test plan:

Unit tests included.

### DIFF
--- a/pkg/builder/vm.go
+++ b/pkg/builder/vm.go
@@ -240,6 +240,16 @@ func (v *VMBuilder) RunStrategy(runStrategy kubevirtv1.VirtualMachineRunStrategy
 	return v
 }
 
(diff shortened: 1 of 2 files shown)


ibrokethecloud (approved):
lgtm. thanks.

PR #9161, opened 2025-09-17
docs(hep): harvester upgrade v2

starbops:
#### Related Issue(s):

#7101 
#7112

### DIFF
--- a/enhancements/20241211-upgrade-enhancement.md
+++ b/enhancements/20241211-upgrade-enhancement.md
@@ -404,6 +404,133 @@ The challenge of switching to SUC Plans is that the node-upgrade phase is no lon
 
 Overview on how the enhancement will be implemented.
 
+```mermaid
+sequenceDiagram
+    participant User as User
+    participant K8s as Kubernetes API (Control Plane)
+    participant HARV as Harvester Controllers
+    participant US as Upgrade Shim
+    participant UM as Upgrade Manager
+    participant SUC as System Upgrade Controller
+    participant UR as Upgrade Repository
+    participant Nodes as Node Agents/Jobs
+
+    Note over User,Nodes: Phase 1: Preparation
+    User->>K8s: Create Version
+
+    Note over User,Nodes: Phase 2: Upgrade Initiation (Shim is in charge)
+    User->>K8s: Create UpgradePlan
+    K8s-->>US: Watch: UpgradePlan created
+    US->>K8s: Create VirtualMachineImage
+    K8s-->>HARV: Watch: VirtualMachineImage created
+    activate HARV
+    HARV->>K8s: Update VirtualMachineImage status (Imported)
+    deactivate HARV
+    K8s-->>US: Watch: VirtualMachineImage imported
+    US->>K8s: Create Upgrade Repository
+    K8s-->>Nodes: Schedule pods
+    Nodes->>UR: Create
+    UR->>UR: Download ISO image
+    activate UR
+    deactivate UR
+
+    Note over User,Nodes: Phase 3: Manager Bootstrap (Shim is in charge)
+    UR->>Nodes: Preload Manager container image
+    activate Nodes
+    Nodes->>K8s: Update pod status (Ready)
+    deactivate Nodes
+    K8s-->>US: Watch: Repository ready
+    US->>K8s: Create Upgrade Manager
+    K8s-->>Nodes: Schedule pods
+    Nodes->>K8s: Update pod status (Ready)
+    US->>K8s: Update UpgradePlan status
+    K8s-->>UM: Watch: UpgradePlan status [repository created]
+
+    Note over User,Nodes: Phase 4: Image Preloading (Manager takes over)
+    UM->>K8s: Create Plan (image-preload)
+    K8s-->>SUC: Watch: Plan created
+    loop For each node batch (node-by-node by default)
+        SUC->>K8s: Create Jobs (image-preload)
+        K8s-->>Nodes: Schedule Jobs
+        Nodes->>UR: Get container image tarballs
+        activate Nodes
+        Nodes->>Nodes: Preload container images
+        Nodes->>K8s: Update Job status
+        deactivate Nodes
+        K8s-->>SUC: Watch: Jobs complete
+        SUC->>K8s: Update Plan status
+    end
+    K8s-->>UM: Watch: Plans complete
+
+    Note over User,Nodes: Phase 5: Cluster Upgrade
+    UM->>K8s: Ensure Cluster CR deleted (decouple from Rancher)
+    UM->>K8s: Create Job (cluster-upgrade)
+    K8s->>Nodes: Schedule Job
+    Nodes->>Nodes: Upgrade charts
+    activate Nodes
+    Nodes->>K8s: Update Job status
+    deactivate Nodes
+    K8s-->>UM: Watch: Job complete
+
+    Note over User,Nodes: Phase 6: Node Upgrade - RKE2
+    UM->>K8s: Create Plan (RKE2 upgrade)
+    loop For each node group
+        K8s-->>SUC: Watch: Plan CR created
+        SUC->>K8s: Create Jobs
+        K8s-->>Nodes: Schedule Jobs
+        Nodes->>Nodes: VM evacuation
+        activate Nodes
+        Nodes->>K8s: Request self-drain
+        deactivate Nodes
+        K8s->>Nodes: Drain pods
+        Nodes->>Nodes: RKE2 upgrade
+        activate Nodes
+        Nodes->>K8s: Update Job/Node status
+        deactivate Nodes
+        K8s-->>SUC: Watch: Jobs complete
+        SUC->>K8s: Update Plan status
+    end
+    K8s-->>UM: Watch: Plan complete
+
+    Note over User,Nodes: Phase 7: Node Upgrade - OS (Optional)
+    opt OS Upgrade requested
+        UM->>K8s: Create Plan (OS upgrade)
+        loop For each node group
+            K8s-->>SUC: Watch: Plan CR created
+            SUC->>K8s: Create Jobs
+            K8s-->>Nodes: Schedule Jobs
+            Nodes->>Nodes: VM evacuation
+            activate Nodes
+            Nodes->>K8s: Request self-drain
+            deactivate Nodes
+            K8s->>Nodes: Drain pods
+            Nodes->>Nodes: OS upgrade
+            activate Nodes
+            Nodes->>Nodes: Reboot
+            Nodes->>K8s: Update Job/Node status
+            deactivate Nodes
+            K8s-->>SUC: Watch: Jobs complete
+            SUC->>K8s: Update Plan status
+        end
+        K8s-->>UM: Watch: Plan complete
+    end
+
+    Note over User,Nodes: Phase 8: Cleanup
+    UM->>K8s: Delete Plans
+    UM->>K8s: Delete Jobs
+    UM->>K8s: Update UpgradePlan status
+    K8s-->>US: Watch: UpgradePlan status [resource cleaned up]
+
+    Note over User,Nodes: Phase 9: Finalization (Shim takes over)
+    US->>K8s: Delete Upgrade Manager
+    US->>K8s: Delete Upgrade Repository
+    UM->>K8s: Update UpgradePlan status (Succeeded)
+    K8s-->>US: Watch: UpgradePlan succeeded
+
+    User->>K8s: Get UpgradePlan status
+    K8s-->>User: Upgrade complete
+```
+
 #### Upgrade Shim
 
 Upgrade Shim reconciles the Version and Upgrade CRs.
--- a/enhancements/20241211-upgrade-enhancement.md
+++ b/enhancements/20241211-upgrade-enhancement.md
@@ -174,7 +174,7 @@ type ReleaseMetadata struct {
 
 // UpgradePlanSpec defines the desired state of UpgradePlan
 type UpgradePlanSpec struct {
-	// version refers to the corresponding version resource in the same namespace.
+	// version refers to the corresponding version resource.
 	// +required
 	Version string `json:"version"`
 
@@ -228,7 +228,7 @@ type UpgradePlanStatus struct {
 	// +optional
 	ReleaseMetadata *ReleaseMetadata `json:"releaseMetadata,omitempty"`
 
-	// version is the snapshot of the associated Version resource
+	// version is the snapshot of the associated Version resource.
 	// +optional
 	Version *VersionSpec `json:"version,omitempty"`
 }
--- a/enhancements/20241211-upgrade-enhancement.md
+++ b/enhancements/20241211-upgrade-enhancement.md
@@ -318,6 +318,77 @@ type Version struct {
 
 The overall architecture of Harvester Upgrade V2 is as follows.
 
+```mermaid
+graph TD
+  %% Objects
+  subgraph Harvester_Main["Harvester Main"]
+    US[Upgrade Shim]
+  end
+
+  subgraph Upgrade_Toolkit["Upgrade Toolkit"]
+    VMIMAGE[(VirtualMachineImage: harvester.iso)]
+    subgraph Upgrade_Repo["Upgrade Repository"]
+      PVC[(RWX PVC)]
+      SVC[Service]
+      UR1[(Repo Pod #1)]
+      URn[(Repo Pod #n)]
+    end
+    UM[Upgrade Manager]
+    ClusterUpgradeJob[Job: cluster-upgrade]
+  end
+
+  subgraph Cattle_System_NS["cattle-system ns"]
+    SUC[System Upgrade Controller]
+    PlanPreload[[Plan: image-preload]]
+    PlanRKE2[[Plan: node-upgrade - RKE2]]
+    PlanOS[[Plan: node-upgrade - OS]]
+    Rancher[Rancher]
+    ClusterCR[Cluster CR]
+  end
+
+  subgraph Cluster_Nodes["All Cluster Nodes"]
+    direction LR
+    N1{{Node A}}
+    N2{{Node B}}
+    Nn{{Node N}}
+    subgraph Supply_Path["Image Distribution"]
+      Spegel[Spegel]
+    end
+  end
+
+  %% Relations
+  US -- watches --> VersionCR[Version CR]
+  US -- reconciles --> UpgradePlanCR[UpgradePlan CR]
+  US -- creates --> VMIMAGE
+  US -- creates --> Upgrade_Repo
+  PVC -- download/extract --> VMIMAGE
+  PVC --> UR1
+  PVC --> URn
+  SVC -. load balance .-> UR1
+  SVC -. load balance .-> URn
+
+  UR1 -- preload Upgrade Manager image on nodes --> N1
+  URn -- preload Upgrade Manager image on nodes --> Nn
+
+  US -- creates --> UM
+  UM -- reconciles --> UpgradePlanCR
+
+  UM -. (optional) creates .-> Upgrade_Repo
+  UM -- creates/reconciles --> PlanPreload
+  UM -- deletes --> ClusterCR
+  UM -- creates --> ClusterUpgradeJob
+  UM -- creates/reconciles --> PlanRKE2
+  UM -- (optional) creates/reconciles --> PlanOS
+
+  SUC -- executes --> PlanRKE2
+  SUC -- executes --> PlanOS
+
+  PlanRKE2 -- prepare/drain/upgrade --> Cluster_Nodes
+  PlanOS -- prepare/drain/upgrade --> Cluster_Nodes
+
+  Rancher -. reconciles .-> ClusterCR
+```
+
 ### Upgrade Shim
 
 Upgrade Shim should be the least frequently changed and the most stable part of the entire Harvester Upgrade mechanics. This implies it should be as simple as possible, avoiding any sophisticated approaches. This part also lives with other controllers in the Harvester main program. Upgrade Shim is responsible for the following tasks:
--- a/enhancements/20241211-upgrade-enhancement.md
+++ b/enhancements/20241211-upgrade-enhancement.md
@@ -212,8 +212,8 @@ type UpgradePlanStatus struct {
 	// +optional
 	NodeUpgradeStatuses map[string]NodeUpgradeStatus `json:"nodeUpgradeStatuses,omitempty"`
 
-	// phase shows what overall phase the UpgradePlan resource is in.
-	Phase UpgradePlanPhase `json:"phase,omitempty"`
+	// currentPhase shows what overall phase the UpgradePlan resource is in.
+	CurrentPhase UpgradePlanPhase `json:"currentPhase,omitempty"`
 
 	// phaseTransitionTimestamp is the timestamp of when the last phase change occurred.
 	// +listType=atomic
@@ -237,7 +237,7 @@ type UpgradePlanStatus struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:scope=Cluster,shortName=up;ups
 // +kubebuilder:printcolumn:name="VERSION",type="string",JSONPath=`.spec.version`
-// +kubebuilder:printcolumn:name="PHASE",type="string",JSONPath=`.status.phase`
+// +kubebuilder:printcolumn:name="CURRENTPHASE",type="string",JSONPath=`.status.currentPhase`
 // +kubebuilder:printcolumn:name="AVAILABLE",type=string,JSONPath=`.status.conditions[?(@.type=='Available')].status`
 // +kubebuilder:printcolumn:name="PROGRESSING",type=string,JSONPath=`.status.conditions[?(@.type=='Progressing')].status`
 // +kubebuilder:printcolumn:name="DEGRADED",type=string,JSONPath=`.status.conditions[?(@.type=='Degraded')].status`
--- a/enhancements/20241211-upgrade-enhancement.md
+++ b/enhancements/20241211-upgrade-enhancement.md
@@ -1,0 +1,631 @@
+# Harvester Upgrade Enhancement
+
+## Summary
+
+This enhancement aims to improve the Harvester Upgrade experience by deprecating the Upgrade Repository virtual machine and replacing it with a set of Upgrade Repository pods running on each node behind a unified endpoint. The default image-preloading behavior will be changed to a local-first manner, and all nodes can do the preload independently and simultaneously. Also, a stand-alone Harvester Upgrade Manager will be introduced to manage the (almost) entire upgrade process.
+
+The Harvester Upgrade enhancement, denoted as *Upgrade V2* hereafter, introduced here will involve the following main components:
+
+- Upgrade Shim: A small portion of code embedded in the Harvester main program, i.e., the controller manager, that is responsible for bootstrapping the upgrade environment
+- Upgrade Repository: A set of pods running on each node behind a unified endpoint to replace the original Upgrade Repository functionalities
+- Upgrade Manager: A dedicated controller manager released separately that is responsible for the Harvester Upgrade mechanics
+
+### Related Issues
+
+- https://github.com/harvester/harvester/issues/7101
+- https://github.com/harvester/harvester/issues/7112
+
+## Motivation
+
+**Clunky Upgrade Repository Virtual Machine**
+
+Every Harvester Upgrade instance runs an Upgrade Repository virtual machine (hereafter referred to as *the upgrade-repo VM*) for different upgrade-relevant components to query data. This VM is necessary because Harvester is released as ISO images. Booting up a virtual machine from the ISO image is the most straightforward way. So far, no known way to spin up pods that mount an ISO image directly exists.
+
+However, the VM-based approach has several drawbacks:
+
+- Virtual machines typically have a larger footprint than pods regarding startup time and resource usage.
+  - Populating a volume (and its replicas) from the backing image is a time-consuming process
+  - Starting up a virtual machine means going through a complete OS boot-up process, which is more than necessary for the upgrade-repo use case
+  - The already-precious resources are consumed by the OS stack of the upgrade-repo VM
+- Hard to scale and ensure resiliency during the upgrade. If the only upgrade-repo VM is down, the upgrade process will be blocked.
+- Though virtual machines can be live-migrated, they still have limitations (for instance, https://github.com/harvester/harvester/issues/7096). Besides, under the upgrade-repo context, it's an overkill.
+
+The Upgrade Repository, in its essence, is just an HTTP server serving files. It can be simplified to make the entire workflow streamlined.
+
+**Self-upgrade Paradox of the Embedded Upgrade Controller Manager**
+
+The current embedded upgrade controller manager has a bootstrap problem. Let's say the user initiates the upgrade from v1.5.2 to v1.6.0:
+
+- Early upgrade phases, e.g., the image-preload phase, execute with the current version (v1.5.2)
+- Controller manager binary gets replaced (upgraded to v1.6.0) at the apply-manifests phase
+- Later phases run with the new controller version (v1.6.0)
+
+As a result, bugs in early-phase logic require a two-release fix cycle. With the example above, the error-prone behaviors can only be rectified when upgrading from v1.6.0.
+
+**Cumbersome Maintenance Burden Imposed by Rancher-managed Self-import RKE2 Cluster**
+
+Harvester leverages Rancher Provisioning V2 framework (v2prov) to manage the local RKE2 cluster. The framework requires the Rancher System Agent to be installed and running on each node. Instructions are then distributed to the system agents through machine plan Secrets. Rancher utilizes both the annotations and data sections of the machine plan Secret to facilitate cluster node life-cycle management, including upgrades, in a cloud-native way. Specifically, the Harvester Upgrade relies on the hook mechanism exposed by the v2prov functionality to realize pre- and post-draining tasks, which include evacuating user workloads and upgrading the node operating system.
+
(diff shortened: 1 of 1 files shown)


Vicente-Cheng:
Hey folks,
Please help review the upgrade v2 design.

cc @ibrokethecloud, @ihcsim, @w13915984028

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+
+The introduction of Upgrade Manager is the most crucial part of the Harvester Upgrade V2 enhancement. It not only separates the upgrade reconcile loops from the bulky Harvester controller manager, but also changes how node-upgrade used to work in Upgrade V1. It encompasses two stages to give us enough time to incorporate all the changes and make them take effect.
+
+#### Stage 1 - Transition Period (version N-1 to N upgrades, where N is the debut minor version of Harvester Upgrade V2)
+
+The inner workings of Upgrade Manager in this stage will be essentially the same as before; the significant difference is that it is built and packaged separately from the primary Harvester artifact. It will also have its own Deployment in contrast to the main Harvester controller manager Deployment. The main concerns will be:
```
w13915984028, line 436: it seems to import another requriement:

an upgrade matrix management upon Harvester/UpgradeShim <-> Upgrade Manager, to ensure the versions are matching between them

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+
+> [!NOTE]
+> Spegel cannot replace the Upgrade Repository entirely because, in the end, container images and other artifacts have to somehow appear on one of the nodes to be distributed among others through Spegel.
+
+#### Plan 1 - Static Pods with BackingImage Disk Files
+
+The main idea is to leverage static pods, Longhorn BackingImage disk files, and hostPath volumes. The `minNumberOfCopies` field of the BackingImage for the ISO image is set to the number of nodes of the cluster (subject to change) to distribute the disk files to all nodes. By mounting the BackingImage disk file directly to a specific path on the host filesystem for each node, static pods on each node can access the ISO image content using the hostPath volume.
```
w13915984028, line 412: `The `minNumberOfCopies` field`: ... not sure it works effectively on massive cluster with say 100 nodes.

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+Upgrade Shim should be the least frequently changed and the most stable part of the entire Harvester Upgrade mechanics. This implies it should be as simple as possible, avoiding any sophisticated approaches. This part also lives with other controllers in the Harvester main program. Upgrade Shim is responsible for the following tasks:
+
+1. Downloading the ISO image
+1. (?) Checking the eligibility of the upgrade in terms of the to and from versions
+1. Deploying Upgrade Repository
+1. Bootstrapping Upgrade Manager
+   1. (Optional) Preloading the Upgrade Manager container image
+   1. Deploying Upgrade Manager
+
+### Upgrade Repository
```
w13915984028, line 403: both `plan 1` and `plan 2` seem not robust enough

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+
+Harvester Upgrade V2 is fully unlocked in this stage. The cluster-upgrade phase largely remains the same. In the node-upgrade phase, Rancher v2prov will be dropped, and Harvester Upgrade will no longer rely on the embedded Rancher and Rancher System Agent on each node. Instead, System Upgrade Controller becomes a crucial companion.
+
+The highlights in the original node-upgrade phase are:
+
+1. User workloads, i.e., VMs, evacuation, corresponds to the **pre-drain** stage
+1. Node drain and RKE2 upgrade, corresponds to the **drain** stage
+1. OS upgrade, corresponds to the **post-drain** stage
+
+Since the hook mechanism in Rancher v2prov is no longer usable, Upgrade Manager relies on Plan CRs backed by System Upgrade Controller to execute node-upgrade tasks. One of the hidden advantages of this change is that it allows the operating system upgrade to be separated from the overall upgrade, thus enabling true zero-downtime (for nodes) upgrades. Another one is that it unblocks the way for users to have granular control over the node-upgrade order through node selectors. The new node-upgrade phase looks like the following:
```
w13915984028, line 460: >it allows the operating system upgrade to be separated from the overall upgrade

how to, Harvester or user makes the decision?

simply skip when it is not necessary or selectively do it on another upgrade?

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+  1. User workloads, i.e., VMs, evacuation, achieved by `prepare`
+  1. Node drain and RKE2 upgrade, achieved by `upgrade`
+- **OS-upgrade** stage (skippable)
+  1. User workloads, i.e., VMs, evacuation, achieved by `prepare`
+  1. Node drain and OS upgrade, achieved by `upgrade`
+
+The challenge of switching to SUC Plans is that the node-upgrade phase is no longer executed as a control loop, but as a one-time task. Upgrade Manager needs to reconcile Plans instead of machine-plan Secrets, and there will be no other entities, such as the embedded Rancher, which abstracts the lifecycle management of downstream clusters, to organize the upgrade nuances.
+
+> [!IMPORTANT]
+> The infinite-retry vs. fail-fast will be a key topic for discussion.
```
w13915984028, line 472: it seems no an easy/straightforward decision, we might need to copy those hacks or patches from Rancher to cover various cases

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+
+Another significant aspect is installation. Do we rely on Rancher System Agent to bootstrap clusters? What are the impacts if we drop it entirely? Do we need to develop a new installation method to fill the gap left by the decommissioning of the Rancher System Agent?
+
+It appears that we do rely on the Rancher System Agent to bootstrap nodes except for the initial one. If we remove the Rancher System Agent, there might not be issues in day 2 management; however, we cannot live without it when it comes to cluster bootstrapping. If we leave it as is, the Rancher System Agent generates numerous logs containing error messages (because it is no longer able to communicate with the Rancher Manager) and does nothing.
```
w13915984028, line 977: >because it is no longer able to communicate with the Rancher Manager

The embeded `rancher` deployment is still inside Harvester, what's the reason that it can't communicate, due to below?
```
Decouple the cluster itself with the embedded Rancher Manager by removing the kubernetesVersion and rkeConfig fields from the local provisioning cluster CR
```

w13915984028 (commented):
Thanks for the big and well writen HEP.

A few concerns for open discussion
(1) The alternative solution of upgrade-repo.
(2) The decoupling with Rancher, some hidden dependencies might only be observed on developing/testing stage.
(3) Given Harvester's 4-month release plan, as said in the HEP, it needs more thoughts about the feature delivery stage by stage.

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+	Rancher              string `json:"rancher,omitempty"`
+	MonitoringChart      string `json:"monitoringChart,omitempty"`
+	MinUpgradableVersion string `json:"minUpgradableVersion,omitempty"`
+}
+
+// UpgradePlanSpec defines the desired state of UpgradePlan
+type UpgradePlanSpec struct {
+	// version refers to the corresponding version resource in the same namespace.
+	// +required
+	Version string `json:"version"`
```
ihcsim, line 179: Maybe consider a `Version` type (instead of just `string`) to allow for versions comparison and syntax validation. See e.g., github.com/hashicorp/go-version. Similarly, for `ReleaseMetadata.[Harvester|Kubernetes|Rancher]`.

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+	// +optional
+	PreviousVersion *string `json:"previousVersion,omitempty"`
+
+	// releaseMetadata reflects the essential metadata extracted from the artifact.
+	// +optional
+	ReleaseMetadata *ReleaseMetadata `json:"releaseMetadata,omitempty"`
+
+	// version is the snapshot of the associated Version resource
+	// +optional
+	Version *VersionSpec `json:"version,omitempty"`
```
ihcsim, line 233: Can we use the [`corev1.ObjectReference`](https://pkg.go.dev/k8s.io/api@v0.34.1/core/v1#ObjectReference) type here?

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+// UpgradePlanStatus defines the observed state of UpgradePlan.
+type UpgradePlanStatus struct {
+	// conditions represent the current state of the UpgradePlan resource.
+	// Each condition has a unique type and reflects the status of a specific aspect of the resource.
+	//
+	// The status of each condition is one of True, False, or Unknown.
+	// +listType=map
+	// +listMapKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
ihcsim, line 204: What kind of conditions will be captured here?

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+	// +optional
+	NodeUpgradeStatuses map[string]NodeUpgradeStatus `json:"nodeUpgradeStatuses,omitempty"`
+
+	// phase shows what overall phase the UpgradePlan resource is in.
+	Phase UpgradePlanPhase `json:"phase,omitempty"`
+
+	// phaseTransitionTimestamp is the timestamp of when the last phase change occurred.
+	// +listType=atomic
+	// +optional
+	PhaseTransitionTimestamps []UpgradePlanPhaseTransitionTimestamp `json:"phaseTransitionTimestamps,omitempty"`
```
ihcsim: `Phase` is already captured in `PhaseTransitionTimestamps`.

enhancements/20241211-upgrade-enhancement.md
ihcsim: I assume it should just follow the guideline described in https://github.com/harvester/docs/pull/898/files#diff-9b99ca129b131cc50e1c584be79715640745a08c667b20f0fa714558b2d3077aR52-R53. i.e.:

* Allowing upgrade from one minor version to the next
* Allowing upgrade to a later patch version

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+
+            echo "Download completed, extracting harvester.iso..."
+            gzip -dc harvester.iso.gz > "$WORK_DIR"/harvester.iso
+
+            echo "harvester.iso is ready"
+            touch "$WORK_DIR"/"$READY_FLAG"
+
+            trap - EXIT
+          else
+            echo "$POD_NAME is a follower, waiting for harvester.iso downloaded..."
```
ihcsim, line 755: If you use a statefulset, then you can designate the pod with ordinal index `0` as the leader.

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+              echo "harvester.iso is ready"
+            fi
+          fi
+        env:
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: BI_IMAGE_NAME
+          value: vmi-43ba8c47-5963-4287-902d-7e09d00d9462
```
ihcsim, line 773: Does the upgrade shim provides this value?

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+        - -c
+        - |
+          set -e
+
+          WORK_DIR="/iso"
+          LOCK_FILE="leader.lock"
+          READY_FLAG="harvester.iso.ready"
+
+          if mkdir "$WORK_DIR"/"$LOCK_FILE" 2>/dev/null; then
+            trap "rmdir $WORK_DIR/$LOCK_FILE; rm -vf $WORK_DIR/$READY_FLAG; exit 1" EXIT
```
ihcsim, line 742: Why `exit 1`?

ihcsim (commented):
I'll need to give this another pass. Meanwhile, some preliminary feedback and questions - thanks.

enhancements/20241211-upgrade-enhancement.md
starbops: The overall upgrade policy for Harvester remains intact. What @w13915984028 raised is valid—since the old upgrade controller is being decoupled into two parts, it inevitably introduces a new compatibility requirement between the Upgrade Shim and Manager. The initial plan is to have it the same old way, i.e., Harvester 1.7.0 works with Upgrade Manager v1.7.0. We can bump the version by adding suffices to the Upgrade Manager version, e.g. v1.7.0-r1, just in case we need to fix anything after GA. The new UpgradePlan API allows the user to specify a different Upgrade Manager image.

enhancements/20241211-upgrade-enhancement.md
starbops: The entire Upgrade Repository section is just a preliminary proposal and some experiments we've explored so far. Please refer to #9217 for the formal PR. The implementation works in a different way and should be more robust. Thank you.

enhancements/20241211-upgrade-enhancement.md
starbops: I'd say it provides flexibility, but does not necessarily offer a supported API to the user at this time. In the previous architecture, it was not possible to separate the two because they're both part of the Rancher v2prov framework (pre- and post-hooks). We still need to discuss whether or not it's meaningful and valuable for users to upgrade clusters but not node OSes in the context of Harvester. What I can tell is that I've heard people demand the ability to upgrade clusters without node reboots, which implies no OS upgrades are involved.

enhancements/20241211-upgrade-enhancement.md
starbops: It depends on LH's implementation. Probably not the best choice here. See #9217 for a better way to do this.

enhancements/20241211-upgrade-enhancement.md
starbops: Sorry I'm not quite sure what you mean. Mind sharing some examples?

enhancements/20241211-upgrade-enhancement.md
starbops: IIRC, if the `kubernetesVersion` and `rkeConfig` fields from the `local` provisioning cluster CR are removed, it implies the embedded Rancher Manager can no longer manage the cluster it runs on. The Machine CR and machine-plan Secrets are gone. The communication between Rancher System Agent and Rancher Manager is broken.

enhancements/20241211-upgrade-enhancement.md
starbops: The comment at L177 is misleading. I'll update it. The field is used to refer to the other CRD, Version. The name of a Version CR can be anything, not necessarily a semantic version. Also, we don't rely on this field to validate upgrade eligibility because it's entirely freestyle and can be anything specified by the user, e.g., `dev`, `test`, etc., depending on what the corresponding Version CRs are named.

enhancements/20241211-upgrade-enhancement.md
w13915984028: When looking into Rancher's plan processing code, it is very complex; if we do the related processing on Harvester independently, some or many (edge) scenarios from Rancher plan are also valid for Harvester, we need to summarize those scenarios &/ borrow some implementations.

enhancements/20241211-upgrade-enhancement.md
starbops: There was a case where the Version CR referenced by the Upgrade CR had already been GC'd while the upgrade controller still needed it. This is because another piece of logic handles GC for Version CRs. Since Version CRs only record static information and serve as candidates for upgrades, and there is no actual reconciliation built on them, I feel storing a snapshot of them in the Upgrade CRs is acceptable. WDYT?

enhancements/20241211-upgrade-enhancement.md
w13915984028: we could update the HEP to simplify the reviewing, follow the design on #9217 for open discussion

enhancements/20241211-upgrade-enhancement.md
starbops: There are three conditions currently:

```go
const (
	// Available means the UpgradePlan is ready to be reconciled.
	UpgradePlanAvailable string = "Available"
	// Progressing means the cluster is currently applying the UpgradePlan.
	UpgradePlanProgressing string = "Progressing"
	// Degraded means the progress of the upgrade is stalled due to issues.
	UpgradePlanDegraded string = "Degraded"
)
```

enhancements/20241211-upgrade-enhancement.md
starbops: Exactly. Though it looks like a duplicate, a dedicated `Phase` field helps construct the `PhaseTransitionTimestamps` slice. BTW, these fields are cloned from KubeVirt's VirtualMachineInstanceMigration CRD and work similarly.

enhancements/20241211-upgrade-enhancement.md
starbops: Yes, with a StatefulSet, it can be achieved easily. However, it also defeats the idea of using one single RWX volume. IIUC, pods of a StatefulSet cannot use the same PVC.

enhancements/20241211-upgrade-enhancement.md
starbops: To make the init container respawn and rerun the logic in case of errors, such as download failures.

enhancements/20241211-upgrade-enhancement.md
starbops: Yes, Upgrade Shim needs to provide such information.

BTW, the proposed implementation probably won't be used and should be superseded by #9217.

enhancements/20241211-upgrade-enhancement.md
starbops: Some of the reconciliation logic will be offloaded to the System Upgrade Controller since we rely on SUC Plans instead. From Harvester's viewpoint (specifically, Upgrade Manager), it only cares about the status of SUC Plans (and probably the underlying Jobs for node upgrade status updates). The main idea is to decouple from Rancher's v2prov framework (the complex plan-processing code you mentioned above) and avoid reimplementing parts of it.

enhancements/20241211-upgrade-enhancement.md
ihcsim: For upgrade policy to remain intact, won't it mean the shim and the manager versions need to follow the same requirements? E.g., if upgrading to Harvester 1.6.1, shim v1.6.0 should work with manager v1.6.1, shim v1.5.2 should work with manager v1.6.1? Granted, the logic of the shim and manager shouldn't change often between releases. And if possible, we should avoid a separate versioning scheme/matrix for the manager.

enhancements/20241211-upgrade-enhancement.md
ihcsim: Try use `template.spec.volumes` in the STS to mount the RWX PVC, instead of `spec.volumeClaimTemplates`.

enhancements/20241211-upgrade-enhancement.md
ihcsim: oh.. timestamp is a slice - gotcha! So `Phase` is more like `CurrentPhase`?

enhancements/20241211-upgrade-enhancement.md
ihcsim: So this is the spec of the `Version` specified in `UpgradePlanSpec`? Storing a snapshot works, or just don't GC the version (I know we are doing it now), and use other mechanism to determine upgradable versions, prevent downgrade etc.

starbops: Yes, it's simply a snapshot of the referenced Version CR's spec.

To my understanding, the Version CRD serves as a means for us to distribute the upgradable versions through a public channel. Otherwise, we could've moved all the fields from the Version CRD to the Upgrade CRD, and it would make no difference. I currently have no plan to update the GC logic for Version CRs, so a rough idea is to add a finalizer to the Version CR that is referenced as a workaround. Any suggestions?

enhancements/20241211-upgrade-enhancement.md
starbops: Exactly. Or do you suggest renaming it to `CurrentPhase`? I simply followed the naming from the VirtualMachineInstanceMigration CRD.

enhancements/20241211-upgrade-enhancement.md
starbops: That works. I totally overlooked that option 😂 

That said, the entire upgrade-repo section just shows some POCs. The proper implementation is at #9217. I will update the HEP accordingly later.

enhancements/20241211-upgrade-enhancement.md
ihcsim: makes sense to keep it as-is - thanks.

enhancements/20241211-upgrade-enhancement.md
ihcsim: yeah, i think `CurrentPhase` is clearer - thanks.

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+	// Progressing means the cluster is currently applying the UpgradePlan.
+	UpgradePlanProgressing string = "Progressing"
+	// Degraded means the progress of the upgrade is stalled due to issues.
+	UpgradePlanDegraded string = "Degraded"
+)
+
+const (
+	NodeStateImagePreloading     string = "ImagePreloading"
+	NodeStateImagePreloaded      string = "ImagePreloaded"
+	NodeStateKubernetesUpgrading string = "KubernetesUpgrading"
```
Vicente-Cheng, line 123: Do we need another status for the charts?
Or we unify it on `KubernetesUpgrading`?

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+	Version string `json:"version"`
+
+	// upgrade can be specified to opt for any other specific upgrade image. If not provided, the version resource name is used.
+	// For instance, specifying "dev" for the field can go for the "rancher/harvester-upgrade:dev" image.
+	// +optional
+	Upgrade *string `json:"upgrade,omitempty"`
+
+	// force indicates the UpgradePlan will be forcibly applied, ignoring any pre-upgrade check failures. Default to "false".
+	// +optional
+	Force *bool `json:"force,omitempty"`
```
Vicente-Cheng, line 188: should we really need the field for it (force upgrade, ignore all checks)? or just use the annotation?

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+  SUC -- executes --> PlanRKE2
+  SUC -- executes --> PlanOS
+
+  PlanRKE2 -- prepare/drain/upgrade --> Cluster_Nodes
+  PlanOS -- prepare/drain/upgrade --> Cluster_Nodes
+
+  Rancher -. reconciles .-> ClusterCR
+```
+
+### Upgrade Shim
```
Vicente-Cheng, line 392: could we summarize our all pre-check (no matter on precheck script or webhook currently) on this stage?
We now did not have the centrialize place to do the whole checks.

enhancements/20241211-upgrade-enhancement.md
Vicente-Cheng: Do we discuss the `not robust enough` on #9217?

enhancements/20241211-upgrade-enhancement.md
Vicente-Cheng, line 460: I am not really sure about that. If we cannot move to SUC plan, we can only use hook mechanism like current upgrade flow, right?

enhancements/20241211-upgrade-enhancement.md
starbops: Chart upgrades are conducted in the ClusterUpgrade phase, so they won't appear here, as this is for reflecting node-specific upgrade states.

enhancements/20241211-upgrade-enhancement.md
starbops: Some of the checks are not suitable to be conducted in the webhook, e.g., node-level checks, so we put those into the pre-check script that can be run directly on the node; some are impossible to do then we push over to a later phase, such as upgrade eligibility check (the trustworthy version information is only packed in the ISO image. The name of the Version CR can be anything thus we shouldn't rely on it.)

Do you want to put all the checks in the Upgrade Shim? I'm unsure about the question here.

enhancements/20241211-upgrade-enhancement.md
starbops: IMO, if we really want a feature to be used on rare occasions, we put it in the annotation; otherwise, making it available in the API (CRD) is better because

1. The information about the feature is more organized at the code level.
2. Users can utilize tools like `kubectl explain` to understand its functionality.

It really depends. If we want to highlight specific checks that are skippable, annotations are the way to go, because no one wants their CRD to become bloated.

enhancements/20241211-upgrade-enhancement.md
starbops: Yes, that's the backup plan if the SUC path is not feasible.

starbops:
Due to Harvester's unique use of Rancher (the local cluster is neither fully imported nor provisioned), it is currently impossible to opt out of the Rancher v2prov framework. This will cause the proposed node-upgrade SUC plans to conflict with the v2prov mechanism. We've pivoted to stay in for the node-upgrade phase in Upgrade V2 after some discussions. I will update the HEP accordingly.

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+
+Upgrade Shim should be the least frequently changed and the most stable part of the entire Harvester Upgrade mechanics. This implies it should be as simple as possible, avoiding any sophisticated approaches. This part also lives with other controllers in the Harvester main program. Upgrade Shim is responsible for the following tasks:
+
+1. Downloading the ISO image
+1. (?) Checking the eligibility of the upgrade in terms of the to and from versions
+1. Deploying Upgrade Repository
+1. Bootstrapping Upgrade Manager
+   1. (Optional) Preloading the Upgrade Manager container image
+   1. Deploying Upgrade Manager
```
ihcsim, line 401: Is it possible to move all these out of the upgrade shim like e.g., bundle them into an upgrade helm chart? In my mind, I think the upgrade shim should be so thin that it's only responsibility is to kickstart upgrade by e.g., downloading and installing the suggested upgrade helm chart, and maybe aggregating and sending upgrade errors to a source or something.

Helm also makes it easier to upgrade, run hooks and cleanup the upgrade components. E.g., upgrading Harvester from 1.7 to 1.8 can just be initiated by upgrading the upgrade helm chart from 1.7 to 1.8.

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+
+### Upgrade Repository
+
+Upgrade Repository is the warehouse that is built on the shipped ISO image. It serves other upgrade-relevant components, not just artifacts, but also metadata. The proposed design is to replace the upgrade-repo VM with a set of upgrade-repo pods that are behind a consistent service endpoint, providing a highly available in-cluster upgrade-repo service. This can be further enhanced by increasing speed and saving traffic through the introduction of Spegel.
+
+> [!NOTE]
+> Spegel cannot replace the Upgrade Repository entirely because, in the end, container images and other artifacts have to somehow appear on one of the nodes to be distributed among others through Spegel.
+
+#### Plan 1 - Static Pods with BackingImage Disk Files
```
ihcsim, line 410: Something for consideration is whether we can just bypass CSI and CDI altogether. The rationale is to reduce as many variables as possible. On a typical cluster, both CSI and CDI are heavily consumed by user workloads. We can't always count on their availability and stability.

One alternative is to start the upgrade-repo as a DaemonSet where each pod mounts a `hostpath` volume, and it downloads and writes the ISO directly to the nodes, or start upgrade-repo as a static pod running on a management node and rsync the downloaded ISO to other nodes on the cluster.

enhancements/20241211-upgrade-enhancement.md
```diff
@@ -0,0 +1,997 @@
...
+
+Plan 2 does not mess with the host filesystem and is more robust even under the chaos of an upgrading cluster. The drawback is unnecessary compression/decompression back and forth overhead, resulting in a longer preparation time compared to the previous VM-based implementation. Also, the share manager pod of the RWX volume becomes the SPOF (Single Point of Failure) of the entire upgrade-repo architecture.
+
+### Upgrade Manager Preloader
```
ihcsim, line 422: as suggested above, can we package the upgrade manager in a helm chart and let upgrade shim download and install the helm chart?

enhancements/20241211-upgrade-enhancement.md
starbops: That was one of the original plans. Setting up and tearing down static pods requires running node-level maintenance jobs. In this case, we usually leverage SUC plans. However, we need to take the following into account:

1. Not breaking existing ways to initiate an upgrade
2. Image availability to each node

A group of users relies on initiating the upgrade by uploading the ISO image directly from their computer. It would be hard to preserve the entrypoint if we change the implementation to the way you mentioned.
OTOH, storing a copy of the ISO image on every node is straightforward and most efficient, but we cannot assume that every node has enough disk space for it. Moreover, if the ISO image is corrupted or missing, the node still needs a way to access that data from other nodes. This makes the design more complicated. Lastly, the repository needs to survive node reboots, as some post-actions require it during upgrade completion. Though static pods survive reboots, mount points do not.
As a result, we still used the centralized approach: downloading the ISO image to a single location in a single shot, setting up the repository, and consumers get what they want on demand.

enhancements/20241211-upgrade-enhancement.md
starbops: > Is it possible to move all these out of the upgrade shim like e.g., bundle them into an upgrade helm chart? In my mind, I think the upgrade shim should be so thin that it's only responsibility is to kickstart upgrade by e.g., downloading and installing the suggested upgrade helm chart, and maybe aggregating and sending upgrade errors to a source or something.
> 
> Helm also makes it easier to upgrade, run hooks and cleanup the upgrade components. E.g., upgrading Harvester from 1.7 to 1.8 can just be initiated by upgrading the upgrade helm chart from 1.7 to 1.8.

I like the idea, and I would love to do that. The only question is whether we want to break the existing ways users can kick off upgrades.
The main consideration is the air-gapped environment. As before, users only need to prepare one thing, i.e., the new version ISO image, and they can kick off the upgrade. Should we choose to do it, then it would be different. Let's say we want to upgrade a cluster from v1.7.0 to v1.8.0. The v1.8.0 Upgrade Manager chart will be installed by the shim first. But the v1.8.0 Upgrade Manager is bundled in the v1.8.0 ISO image, which hasn't been downloaded yet, and the container image hasn't been preloaded to each node; this will be an issue. Overall, if we choose to let the shim install the chart, we must also distribute the chart and the container image somewhere, separately from the ISO image we used to distribute. Also, users need to host the chart and images themselves, along with the ISO image. This would require them to set up an OCI registry in their environment. Only if we can convince ourselves and users that this does not complicate the upgrade procedure can we do it.

ihcsim: I am not convinced that setting up an oci registry poses much of a hurdle to air-gapped users. In exchange, they get the benefits of v2 upgrade. It also becomes easier for users and maintainers to scrutinize, scan, test, and reproduce upgrade-related bugs. If breaking the existing ways yield more benefits and stability, maybe it's worth it. There are always ways to verify with existing users if we are open to this idea.

I guess, fundamentally, I am not sure how we decide the logical boundary for what stays in the shim and what stays in the upgrade manager. In the current proposal, can the steps to "Checking the eligibility of the upgrade in terms of the to and from versions" and "Deploying Upgrade Repository" be moved to the upgrade manager?

pohanhuang:
Hi @starbops,
I’m not sure where to raise this, so I’m leaving it here.

Issue: Our Helm upgrade/release flow doesn’t have a clear timeout/escape hatch in some phases. PR #9934 improves the rollback behavior, which helps a lot, but the overall upgrade can still get stuck in certain edge cases (i.e., it neither completes nor cleanly rolls back).

Do you think we should open an issue to track adding phase-level timeouts / guardrails (and documenting the recommended manual recovery steps)? Happy to help draft the issue with concrete scenarios/logs if needed.

Thanks

starbops:
> Hi @starbops, I’m not sure where to raise this, so I’m leaving it here.
> 
> Issue: Our Helm upgrade/release flow doesn’t have a clear timeout/escape hatch in some phases. PR #9934 improves the rollback behavior, which helps a lot, but the overall upgrade can still get stuck in certain edge cases (i.e., it neither completes nor cleanly rolls back).
> 
> Do you think we should open an issue to track adding phase-level timeouts / guardrails (and documenting the recommended manual recovery steps)? Happy to help draft the issue with concrete scenarios/logs if needed.
> 
> Thanks

@pohanhuang Thanks for the input. Given the nature of the Harvester Upgrade, it is generally not good practice, and may not even be possible, to roll back every step or bail out the process when it gets stuck. Our current approach is to retry (endlessly), which might sound trivial, but it at least does not push the cluster to a more complicated situation. This can get us through because most steps are idempotent; retrying can overcome transient errors. Cluster stability and user workload availability are the most concerned topics when it comes to upgrade.
I agree that we can have an issue tracking your proposal. At least we can discuss the applicability phase by phase. Thanks again!

starbops:
The HEP still requires updates to reflect the current design.

Issue #8770, opened 2025-07-28
[ENHANCEMENT] Review the volume detacher behavior


Vicente-Cheng:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

We introduce the volume detacher mechanism during the upgrade.
Reference code section:
https://github.com/harvester/harvester/blob/master/pkg/controller/master/node/node_volume_detach_controller.go

Reference issues as below:
- [[BUG] Upgrade 3 nodes Harvester stuck in Pre-drained of the first node (create volume with snapshot)](https://github.com/harvester/harvester/issues/3681#top)
- [[BUG] VM stuck in *Starting* with _FailedMapVolume_ event](https://github.com/harvester/harvester/issues/3727#top)
That would help us to detach the non-reference volume (comes from backup/snapshot) during the upgrade to avoid the upgrade hang.

However, after Longhorn introduced the volume attachment ticket mechanism, Longhorn can now automatically attach/detach volumes with backups and snapshots.

### Describe the solution you'd like

We should review the volume detacher logic to determine if it is still necessary.

WebberHuang1118:
During internal testing of the upgrade from `v1.5` to `v1.6`, we encountered [Harvester issue #8904](https://github.com/harvester/harvester/issues/8904).

Related files:
- [volume-flapping.part-aa.zip](https://github.com/user-attachments/files/21982579/voume-flapping.part-aa.zip)
- [volume-flapping.part-ab.zip](https://github.com/user-attachments/files/21982581/voume-flapping.part-ab.zip)

This time, we may have identified how Harvester’s forced volume detachment—specifically in the logic [here in the node_volume_detach_controller](https://github.com/harvester/harvester/blob/dd8bad999d80b7891fe754ae4e222f20dea87b91/pkg/controller/master/node/node_volume_detach_controller.go#L85-L104) —triggers the volume flapping behavior described in [Longhorn issue #11479](https://github.com/longhorn/longhorn/issues/11479)

Here are some details: the focus is on volume `pvc-15433fbc-f325-4009-a45c-608da1e96bcd`, which was initially attached to node `93`. A migration was then initiated from node `93` to node `96`

`ControllerPublish()` for node `96`:
```
[...]

PR #9169, opened 2025-09-17
fix: remove forced volume detachment in the node controller

WebberHuang1118:
#### Problem:
Harvester controller incorrectly forces volume detachment, as noted [here](https://github.com/harvester/harvester/issues/8770#issuecomment-3222987264).

#### Solution:
Remove this unnecessary forced volume detachment behavior during upgrades.

#### Related Issue(s):
#8770

#### Test plan:
* Install Harvester `v1.6.0`.
* Create a running VM.
* Upgrade Harvester to the version from this PR.
* The upgrade should complete successfully.

### DIFF
--- a/pkg/controller/master/node/node_volume_detach_controller.go
+++ b/pkg/controller/master/node/node_volume_detach_controller.go
@@ -92,13 +92,7 @@ func (c *nodeVolumeDetachController) OnNodeChanged(_ string, node *corev1.Node) 
 			if err != nil {
 				return node, err
 			}
-			logrus.Debugf("pvc: %s/%s, wait: %t", pvc.Namespace, pvc.Name, wait)
-			if !wait {
-				logrus.Infof("detach the volume %s of pvc %s/%s on node %s", volume.Name, pvc.Namespace, pvc.Name, node.Name)
-				if err := c.detachVolumeFromNode(pvc, node); err != nil {
-					return node, err
-				}
-			} else {
+			if wait {
 				waitVolumeNames = append(waitVolumeNames, volume.Name)
 			}
 		}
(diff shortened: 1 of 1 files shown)


ihcsim:
With https://github.com/longhorn/longhorn/issues/11479, is it still possible to find non-migrating volumes without pod references during upgrade?

WebberHuang1118:
> With [longhorn/longhorn#11479](https://github.com/longhorn/longhorn/issues/11479), is it still possible to find non-migrating volumes without pod references during upgrade?

Harvester’s behavior, as described [here](https://github.com/harvester/harvester/issues/8770#issuecomment-3222987264), is the reason for the Longhorn volume becoming dangling [here](https://github.com/longhorn/longhorn/issues/11479).

If forced detachment occurs (such as [here](https://github.com/harvester/harvester/blob/master/pkg/controller/master/node/node_volume_detach_controller.go)), [#11479](https://github.com/longhorn/longhorn/issues/11479) aims to keep the volume in a detached state rather than dangling. However, manual intervention is still required, as explained [here](https://github.com/longhorn/longhorn/issues/11479#issue-3320596636), to make the volume functional again.

Therefore, this PR is still needed to remove forced volume detachment in the Harvester controller.

ihcsim:
Essentially, we are letting LH handle all the "non-waiting" volumes during the upgrade pre-drain phase, right? Looking at the implementation of `c.checkDetachVolume(pvc)`, previously force detachment happened when [`return false, nil`](https://github.com/harvester/harvester/blob/ce6583e4eff244afd1bf7ed66f095838af897221/pkg/controller/master/node/node_volume_detach_controller.go#L196), which looks like a "catch all" for anything that don't fits the criteria before the `return` statement.

WebberHuang1118:
> Essentially, we are letting LH handle all the "non-waiting" volumes during the upgrade pre-drain phase, right? Looking at the implementation of `c.checkDetachVolume(pvc)`, previously force detachment happened when [`return false, nil`](https://github.com/harvester/harvester/blob/ce6583e4eff244afd1bf7ed66f095838af897221/pkg/controller/master/node/node_volume_detach_controller.go#L196), which looks like a "catch all" for anything that don't fits the criteria before the `return` statement.

Since this check is triggered when a node is in the `pre-draining` state (as shown [here](https://github.com/harvester/harvester/blob/55cb7c4a13ffc90f21b264c999f8a743a04ec602/pkg/controller/master/node/node_volume_detach_controller.go#L70-L74), for volumes requested to be attached to the upgrading node, we only need to wait for:

* Volumes that are still in a transition state ([reference](https://github.com/harvester/harvester/blob/55cb7c4a13ffc90f21b264c999f8a743a04ec602/pkg/controller/master/node/node_volume_detach_controller.go#L83-L84))
  * The volume should reach a stable state eventually
 
* Attached volumes:

  * If attached because of a running VM, we should wait until the volume or VM is migrated to other nodes.
  * If attached due to a Longhorn snapshot or backup, we should wait until the operation completes (Longhorn will detach the volume automatically).

ihcsim (approved):

Vicente-Cheng (approved):
thanks for the revisit this

Vicente-Cheng:
@mergifyio backport v1.6

Issue #9154, opened 2025-09-16
[BUG] Maintenance mode can be enabled by adding annotation whereas the UI blocks the action


votdev:
### Describe the Bug

Enabling the maintenance mode can be done on nodes where the UI is blocking the action in the maintenance mode modal dialog by adding the annotation `harvesterhci.io/drain-requested: true`. This is because [the UI](https://github.com/harvester/harvester-ui-extension/blob/a694288162ed75a20b8760eb470f377b7c81fef3/pkg/harvester/dialog/HarvesterMaintenanceDialog.vue#L78) is calling an [additional function](https://github.com/harvester/harvester/blob/a9bd2b13e6fdfa5bc44786fd3dbcce7b059472a2/pkg/api/node/formatter.go#L136) which gets a list of VM that can not be migrated. The code path in the related controller is not performing this check, therefore the maintenance mode can be triggered on nodes which normally should deny this action.

The [code of interest](https://github.com/harvester/harvester/blob/29d8acc6b30c4a153322a2f4cd74877654db823a/pkg/controller/master/nodedrain/nodedrain_controller.go#L188) should look like:

*incorrect*
```
	// Shutdown ALL VMs on that node forcibly? This is activated by a
	// checkbox in the maintenance mode dialog in the UI.
	if forced {
		shutdownVMs, err = ndc.FindNonMigratableVMS(node)
		if err != nil {
			return node, fmt.Errorf("error listing VMIs in scope for shutdown: %w", err)
		}
	}
```

*correct*
```
    nonMigratableVMs, err = ndc.FindNonMigratableVMS(node)
	if err != nil {
		return node, fmt.Errorf("error listing VMIs in scope for shutdown: %w", err)
	}
	if !forced {
		if len(nonMigratableVMs) > 0 {
			<Raise error here>
		}
	}
```
This is the equivalent to the [code that is preformed in the UI](https://github.com/harvester/harvester-ui-extension/blob/a694288162ed75a20b8760eb470f377b7c81fef3/pkg/harvester/dialog/HarvesterMaintenanceDialog.vue#L78).

### To Reproduce

- Create a VM and pin it to a specific node.
- Try to put that node into maintenance mode. Use the `Enable Maintenance Mode` action menu in the `Hosts` page in the UI. There should be an error message when pressing the `Apply` button.
[...]

PR #9170, opened 2025-09-17
Maintenance mode can be enabled by adding annotation whereas the UI blocks the action

votdev:
#### Problem:
Enabling the maintenance mode can be done on nodes where the UI is blocking the action in the maintenance mode modal dialog by adding the annotation `harvesterhci.io/drain-requested: 'true'`. This is because [the UI](https://github.com/harvester/harvester-ui-extension/blob/a694288162ed75a20b8760eb470f377b7c81fef3/pkg/harvester/dialog/HarvesterMaintenanceDialog.vue#L78) is calling an [additional function](https://github.com/harvester/harvester/blob/a9bd2b13e6fdfa5bc44786fd3dbcce7b059472a2/pkg/api/node/formatter.go#L136) which gets a list of VM that can not be migrated. The code path in the related controller is not performing this check, therefore the maintenance mode can be triggered on nodes which normally should deny this action.

#### Solution:
The action should be blocked regardless if it is triggered via UI or by adding the annotation to the node resource. To do so, the controller that is responsible for the maintenance mode should perform the additional check that is already done in the UI, so that both paths. triggered via UI and by manually adding annotations, should behave the same.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9154

#### Test plan:
***Prereqs***
Setup a multi-node cluster.

***Case 1***
- Create VM `test01` and schedule it to node `harvester-node-2`.
- Create VM `test02` and schedule it to node `harvester-node-2` as well.
- Go to `Hosts` page and use the action menu `Edit YAML` on node `harvester-node-2`.
- Append the annotation `harvesterhci.io/drain-requested: 'true'`. Press `Save`.
- In the logs of the `harvester-system/harvester-xxxx` pod there should be these lines:
```
time="2025-09-22T07:43:07Z" level=info msg="Attempting to place node in maintenance mode" forced=false node_name=harvester-node-2
time="2025-09-22T07:43:07Z" level=error msg="error syncing 'harvester-node-2': handler node-drain-controller: enabling maintenance mode is impossible. Non-migratable VMs found: default/test01,default/test02 cannot be migrated due to NodeSchedulingRequirementsNotMet. Use 'force drain' to perform a collective shutdown, requeuing"
```
- The node should not be cordoned nor in maintenance mode.
- Make sure the `harvesterhci.io/maintain-status`, `harvesterhci.io/drain-requested` and `harvesterhci.io/drain-forced` annotations are non-existent.

***Case 2***
- Delete the previously created VMs.
- Append the annotation `harvesterhci.io/drain-requested: 'true'` to the node `harvester-node-2`. Press `Save`.
- The node should be cordoned and after a while the maintenance mode label should be seen for `harvester-node-2`.

### DIFF
--- a/pkg/controller/master/nodedrain/nodedrain_controller.go
+++ b/pkg/controller/master/nodedrain/nodedrain_controller.go
@@ -95,27 +95,19 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
 	err := drainhelper.DrainPossible(ndc.nodeCache, node)
 	if err != nil {
 		if errors.Is(err, drainhelper.ErrNodeDrainNotPossible) {
-			logrus.WithFields(logrus.Fields{
-				"node_name": node.Name,
-			}).Warnf("Draining is not possible: %v", err)
-
-			nodeCopy := node.DeepCopy()
-
-			// Cleanup annotations to avoid recurring approaches to drain the node.
-			delete(nodeCopy.Annotations, ctlnode.MaintainStatusAnnotationKey)
-			delete(nodeCopy.Annotations, drainhelper.DrainAnnotation)
-			delete(nodeCopy.Annotations, drainhelper.ForcedDrain)
-
-			_, errUpdate := ndc.nodes.Update(nodeCopy)
+			nodeUpdate, errUpdate := ndc.cleanupMaintenanceModeAnnotations(node)
 			if errUpdate != nil {
-				return node, errors.Join(err, fmt.Errorf("failed to update node to clean up annotations: %w", errUpdate))
+				return node, errors.Join(err, errUpdate)
 			}
-
-			// Do not reconcile again.
-			return nil, err
+			return nodeUpdate, fmt.Errorf("enabling maintenance mode is impossible: %w", err)
 		}
 
 		return node, err
+	}
+
+	nonMigratableVMs, err := ndc.FindNonMigratableVMS(node)
+	if err != nil {
+		return node, fmt.Errorf("error getting non-migratable VMs: %w", err)
 	}
 
 	// List of VMs that need to be forcibly shutdown before maintenance
@@ -124,6 +116,24 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
 
 	if !forced {
 		var maintainModeStrategyVMs []string
+
+		// Make sure there are no VMs that are non-migratable. Abort the
+		// approach to place a node into maintenance mode if there are
+		// such VMs.
+		if len(nonMigratableVMs) > 0 {
+			nodeUpdate, err := ndc.cleanupMaintenanceModeAnnotations(node)
+			if err != nil {
+				return node, err
+			}
+
+			reasons := make([]string, 0, len(nonMigratableVMs))
+			for condition, vms := range nonMigratableVMs {
+				reasons = append(reasons, fmt.Sprintf("%s cannot be migrated due to %s", strings.Join(vms, ","), condition))
+			}
+
+			return nodeUpdate, fmt.Errorf("enabling maintenance mode is impossible. Non-migratable VMs found: %s. Use 'force drain' to perform a collective shutdown",
+				strings.Join(reasons, "; "))
+		}
 
 		// Get the list of VMs that are labeled to forcibly shut down
 		// before maintenance mode.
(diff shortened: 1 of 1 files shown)


pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -97,34 +97,56 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
+		// such VMs.
+		if len(nonMigratableVMs) > 0 {
+			reasons := make([]string, 0, len(nonMigratableVMs))
+			for condition, vms := range nonMigratableVMs {
+				reasons = append(reasons, fmt.Sprintf("%s cannot be migrated due to %s", strings.Join(vms, ","), condition))
+			}
+
+			msg := fmt.Sprintf("non-migratable VMs found: %s. Use the %q annotation to bypass this situation", strings.Join(reasons, "; "), drainhelper.ForcedDrain)
+
+			logrus.WithFields(logrus.Fields{
```
votdev: Does it make sense to log this? The reconciler will log this error as well.
```
time="2025-09-17T09:39:16Z" level=error msg="Enabling maintenance mode is impossible: non-migratable VMs found: default/test01, default/test02 cannot be migrated due to NodeSchedulingRequirementsNotMet. Use the \"harvesterhci.io/drain-forced\" annotation to bypass this situation" node_name=harvester-node-2
time="2025-09-17T09:39:16Z" level=error msg="error syncing 'harvester-node-2': handler node-drain-controller: enabling maintenance mode is impossible: non-migratable VMs found: default/test01, default/test02 cannot be migrated due to NodeSchedulingRequirementsNotMet. Use the \"harvesterhci.io/drain-forced\" annotation to bypass this situation, requeuing"
```
Do we have a rule/guideline in the Harv project how to handle that?

pkg/controller/master/nodedrain/nodedrain_controller.go
```diff
@@ -97,34 +97,56 @@ func (ndc *ControllerHandler) OnNodeChange(_ string, node *corev1.Node) (*corev1
...
+		// Make sure there are no VMs that are non-migratable. Abort the
+		// approach to place a node into maintenance mode if there are
+		// such VMs.
+		if len(nonMigratableVMs) > 0 {
+			reasons := make([]string, 0, len(nonMigratableVMs))
+			for condition, vms := range nonMigratableVMs {
+				reasons = append(reasons, fmt.Sprintf("%s cannot be migrated due to %s", strings.Join(vms, ","), condition))
+			}
+
+			msg := fmt.Sprintf("non-migratable VMs found: %s. Use the %q annotation to bypass this situation", strings.Join(reasons, "; "), drainhelper.ForcedDrain)
```
```suggestion
			msg := fmt.Sprintf("non-migratable VMs found: %s. Use 'force drain' to perform a collective shutdown", strings.Join(reasons, "; "), drainhelper.ForcedDrain)
```
This can also be done via the UI.

pkg/controller/master/nodedrain/nodedrain_controller.go
ihcsim: I don't see any reasons to log it twice.

w13915984028:
@votdev Did you consider to enhance the webhook

https://github.com/harvester/harvester/blob/153edff2d8adc458b18f59564001ce3e87c454b8/pkg/webhook/resources/node/validator.go#L53-L62

where we utilize the `_ *types.Request`, which only allow maitenance related label to be added by the harvester controller via RBAC related info  (e.g. kubevirt has such check, some operations are only allowed to be triggered by controller)

if user operates via API, then post upon the `maitenance` action is required to trigger the operation; and Harvester adds similar check on the action processing code (if it does not include the similar logic as UI does);  controller is free

https://github.com/harvester/harvester/blob/153edff2d8adc458b18f59564001ce3e87c454b8/pkg/api/node/formatter.go#L196-L199

votdev:
> @votdev Did you consider to enhance the webhook
> 
> https://github.com/harvester/harvester/blob/153edff2d8adc458b18f59564001ce3e87c454b8/pkg/webhook/resources/node/validator.go#L53-L62
> 
> where we utilize the `_ *types.Request`, which only allow maitenance related label to be added by the harvester controller via RBAC related info (e.g. kubevirt has such check, some operations are only allowed to be triggered by controller)
> 
> if user operates via API, then post upon the `maitenance` action is required to trigger the operation; and Harvester adds similar check on the action processing code (if it does not include the similar logic as UI does); controller is free
> 
> https://github.com/harvester/harvester/blob/153edff2d8adc458b18f59564001ce3e87c454b8/pkg/api/node/formatter.go#L196-L199

My goal with this PR is to fix the current code the most simple way (KISS).

ihcsim (approved):
LGTM. Not sure why frontend needs to do this check to begin with. It makes more sense for this check to happen in the backend.

Issue #8794, opened 2025-08-02
[BUG] A Harvester node failure blocks all Rancher clusters provisioned with the Harvester node driver


davidepasquero:
#### Context and Critical Importance

The provisioning and management of Rancher clusters via the Harvester node driver is only truly automated when using a DHCP solution where IP leases are tightly coupled to the actual existence of the requesting Rancher node's network interface.

The vm-dhcp-controller perfectly implements this requirement. Alternative solutions, such as using very long leases or "reserved" leases, involve manual intervention that defeats the purpose of the automated provisioning and management offered by the Harvester node driver.

Given the strategic importance of the vm-dhcp-controller in this ecosystem, its fault tolerance is not an option, but a fundamental requirement.

#### Bug Description

The vm-dhcp-controller's architecture deploys a single agent pod for each IPPool object. This singleton nature creates a single point of failure. If the node hosting an agent pod experiences a catastrophic event (e.g., a kubelet or full node crash), the DHCP service for dependent VMs is effectively blocked, leaving them unable to obtain or renew their IP addresses.

#### Steps to Reproduce

Deploy a vm-dhcp-controller and create an IPPool object. An agent pod will be scheduled on a specific node.
Create one or more VMs configured to get their IP from this agent.
Simulate a failure of the node hosting the agent pod (e.g., by stopping the kubelet service or shutting down the machine).
Observe that the agent pod's status becomes 'Terminating' but it is not removed, and therefore a new agent pod is not created.
Verify that the VMs served by that agent can no longer obtain or renew their IP addresses.

#### Expected Behavior

The system should be resilient to node failures. In the event of a node crash, the agent pod should be quickly rescheduled onto a healthy node, or an HA (High Availability) mechanism should ensure DHCP service continuity. VMs should be able to maintain or renew their IP addresses with minimal to no disruption.

#### Actual Behavior

When the node fails, the agent pod enters a Terminating state and is not replaced. As a result, the DHCP service for that IP pool is completely paralyzed. This leads to catastrophic consequences if a downstream cluster depends on the connectivity of those VMs.

### Support Bundle for Troubleshooting

Xxx

### Environment

- Harvester version: 1.5.1

- Rancher version: All

#### Proposed Solution

A solution has been proposed in Pull Request #64:

https://github.com/harvester/vm-dhcp-controller/pull/64

tserong:
Set milestone to v1.7.0 per https://github.com/harvester/vm-dhcp-controller/pull/64#issuecomment-3139113765

Issue #8205, opened 2025-05-06
[ENHANCEMENT] Increase robustness for the Managed DHCP add-on under disastrous events


starbops:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

The VM DHCP controller deploys an agent pod on a node for each IPPool object. If a node encounters disastrous events, e.g., the kubelet stops for no reason or the entire node crashes, the cluster cannot know the agent pods' status on that specific node. Pods without appropriate tolerations will be marked with a deletion timestamp and enter the terminating state due to the `NoExecute` node taint. This behavior is the same for all pods, not just the VM DHCP agents. However, the singleton characteristics of the VM DHCP agent pods result in service discontinuity under such circumstances. The VMs are thus affected and can no longer get their original IP addresses from the agent. Even worse, if a downstream cluster consists of those affected VMs, it will be a catastrophe.

### Describe the solution you'd like

There are two directions to enhance the robustness of the DHCP service:
- Make the VM DHCP controller manage agent Deployments instead of Pods, so the responsibility of pod lifecycle management is passed to the deployment controller. This will give us more leverage, like fine-tuning the rolling update strategy and tolerance.
- Modify the agent pod reconciling logic, making it more paranoid. Whenever the agent pod turns into the terminating state, create a new one to ensure service continuity.

That said, it's unclear whether there exists any possibility that the DHCP agent pod is marked as terminating, but inside the pod, the DHCP server process is still serving requests, causing conflicts with the newly created one.

### Describe alternatives you've considered

The other possible solution is to adopt the DHCP failover mode or load-balance mode, which allows multiple DHCP server instances serving the same network to exist simultaneously. This requires much more work and needs prior evaluation.

### Additional context

It's reported that the only way to bring back the VM DHCP agent pod in such circumstances is to delete the stuck-at-terminating one forcibly. In that case, a new agent pod will then be created and scheduled to the other healthy cluster nodes. Otherwise, users will need to wait until the node comes back.

starbops:
Another way is to leverage leader election among the agent pods so that only one agent takes charge at any given time. Hence, the agent pods can be under a Deployment. That said, the upgrade plan for the transition needs to be considered. Kudos to @ibrokethecloud for providing the idea.

PR #9173, opened 2025-09-17
hep: Introduce Leader Election and use Deployments for DHCP Controller agents

davidepasquero:
https://github.com/harvester/vm-dhcp-controller/pull/64#issuecomment-3247930712

**Related Issues**

#8205 
#8794


starbops:
Still relevant.

starbops (commented):
Hi @davidepasquero, thank you for the HEP! Could you reformat the proposal by following the [template](https://github.com/harvester/harvester/blob/master/enhancements/YYYYMMDD-template.md)?

starbops (commented):
I quickly scanned through it, and one question came to my mind: are you proposing a fixed Helm-installed agent deployment instead of per-IPPool agent deployments? In other words, is it still in the operator pattern?

davidepasquero:
@starbops The proposal replaces the old “one agent pod per IPPool” model with a single, shared agent Deployment installed by the Helm chart and configured through JSON env vars, so agents are no longer created per pool.

It remains an operator-style flow: the controller still watches IPPools, aggregates their active networks, and patches that fixed agent Deployment with the Multus annotations and AGENT_NETWORK_CONFIGS/IPPOOL_REFS_JSON payloads, so runtime behavior is still driven by reconciliation rather than manual pods.

davidepasquero:
@starbops The new "single agent" architecture: the Helm chart always creates a shared Deployment for the agent, while the controller continues to operate with operator logic, reconciling IPPools and updating that Deployment (Multus and env JSON annotations) rather than creating per-pool pods. The pod agent reads aggregated configurations via environment variables.

Fixed Helm Deployment
The chart generates a single Deployment called <release>-agent with a configurable replica count, a dedicated service account, and the *-agent container. The AGENT_NETWORK_CONFIGS and IPPOOL_REFS_JSON variables are initialized to [] and will then be patched by the controller; there are no longer any arguments to specify individual NICs/IPPools.

Operator Flow on the Controller
The controller, in reconcileAgentDeployment, reads all IPPools from the cache, filters those that are active and have essential fields, constructs an AgentNetConfig entry and an ipPoolRefs list for each, and calculates the Multus annotation (k8s.v1.cni.cncf.io/networks) by concatenating the networks of the active pools.

It retrieves the Agent Deployment (name and container derived from the AGENT_DEPLOYMENT_NAME/AGENT_CONTAINER_NAME env), updates the Multus annotation, and replaces obsolete arguments/variables with AGENT_NETWORK_CONFIGS and IPPOOL_REFS_JSON; if anything changes, it executes an Update on the Deployment. This maintains the "operator" behavior because any IPPool changes automatically reconcile the shared Deployment, without creating pods for each pool.

Agent Behavior
Upon startup, the vm-dhcp-agent binary reads the JSON from AGENT_NETWORK_CONFIGS and IPPOOL_REFS_JSON (passed from the controller via env) and uses it to configure the interfaces; the old flags for individual NICs/IPPools have been removed, indicating that the agent now relies solely on the aggregate configuration provided by the operator-patched deployment.

In summary: the deployment is a fixed deployment installed with Helm, but the behavior remains operator-driven because the controller observes the IPPools and dynamically updates that shared deployment with the necessary configurations and annotations.

davidepasquero:
@starbops 
Agent Operation
The vm-dhcp-agent process reads the AGENT_NETWORK_CONFIGS and IPPOOL_REFS_JSON environment variables from the Deployment, configures logging/flags, and creates runtime options: all interfaces, pools, and flags are therefore provided by the controller via env, not from historical CLI arguments.

With these options, the agent builds AgentNetConfig structures (interface, ServerIP, CIDR, IPPool/NAD references) and a list of pools, initializes a shared DHCPAllocator, and prepares a handler for each active IPPool, avoiding duplicates and logging any missing or malformed JSON.

Interface Preparation and Cycle Start
At startup, if not in a dry run, the agent configures each interface: it validates the IP/CIDR, cleans up any existing IPs, adds the calculated prefix address, and brings the link up. Incomplete or incorrect configurations are simply warned, so the agent continues on the other NICs.

Before serving DHCP, the main thread waits for each EventHandler to complete its first cache synchronization across IPPools. If this fails or the context is deleted, it continues, but warns that leases may be incomplete.

Listening to IPPools
Each EventHandler watches the individual IPPool assigned to it (selector by name/namespace), enqueues ADD/UPDATE/DELETE events for that pool only, and delivers them to a dedicated controller with a rate-limited queue. Each handler exposes InitialSyncDone to signal the main loop that the first snapshot of leases is ready.

The per-pool controller updates the poolCache cache and the DHCPAllocator: it deletes no longer present leases, inserts/updates those in the Allocated state, and clears everything if the pool is deleted, so each DHCP reply is consistent with the CRD state.

Multi-Pool DHCP Service
DHCPAllocator maintains separate leases and servers per IPPool (map[ipPoolRef]map[hwAddr]DHCPLease) and rejects entries without a valid pool or MAC address. It adds leases with full parameters (IP server, router, DNS, domain, NTP, lease time), and allows targeted reclamation/deletion per pool.

In response to Discover/Request, the per-pool manager searches for the lease on the correct pool, populates the offer/ack with options (ServerIdentifier, subnet mask, router, DNS, domain, NTP, lease time), and sends the response on the associated NIC.

The agent starts a DHCP server for each configured interface, bound to the specific NIC but listening on 0.0.0.0:67. Each server goroutine terminates when the context is cleared, and the shutdown is orchestrated by Cleanup/stopAll to close sockets in an orderly manner.

Process Management and Leader Election
The agent ingress builds the Kubernetes client, instantiates the Agent, and runs it under leader election (which can be disabled). Only the leader pod processes the pools and serves DHCP, while an optional HTTP server exposes debug/cachedump handlers using the same DHCPAllocator. Everything is governed by an errgroup for clean termination on server signals or errors.

davidepasquero:
@starbops 
Overall DHCP ↔ Agent Flow
The agent receives the network configurations for each IPPool (interface, IP server, CIDR, IPPool reference) from the controller and converts them to []dhcp.DHCPNetConfig; it then passes the list to the DHCPAllocator to start the DHCP servers or simulate them in dry-run mode.

An EventHandler/controller is activated for each IPPool that monitors the IPPool CRD and synchronizes the leases in the DHCPAllocator, so the DHCP server always has an up-to-date snapshot of official leases.

Lease Management in the DHCPAllocator
Leases are indexed per IPPool (map[ipPoolRef]map[hwAddr]DHCPLease); each entry verifies essential parameters (pool, valid MAC, CIDR) and constructs the DHCP options (router, DNS, domain, NTP, lease time) before saving them and logging the operation.

The IPPool controller updates this structure: it removes MAC addresses no longer present in the status, then adds/updates current leases using AddLease, while also maintaining a local cache to track what has already been synchronized.

Per-Pool DHCP Response
Each incoming request is handled by dhcpHandlerPerPool, which searches for the lease in the correct pool (GetLease), constructs the response with consistent options (ServerIdentifier, subnet mask, router, DNS, domains, NTP, lease time), and sends Offer/Ack only if the lease exists.

Startup and Service Cycle
DHCPAllocator.Run starts a DHCP server for each received configuration, binding UDP 67 on the indicated interface and capturing the IPPoolRef in a closure to direct the handler to the correct pool. The servers remain active until the context is cleared, while stopAll neatly closes all sockets during cleanup.

The agent waits for the first IPPool controller synchronization before launching Run, so the servers start with their leases already populated; afterward, the event listeners continue to propagate changes to the same shared DHCPAllocator, keeping the DHCP service aligned with the CRDs.

davidepasquero:
@starbops 

How dhcpHandlerPerPool Works
Initial validation and logging: Each incoming packet is handled in a read-lock protected context; if the frame is null or not a BootRequest, the handler logs the error and stops processing; otherwise, it logs the entire packet for the specific pool (ipPoolRef).

Response construction and lease lookup: The handler creates a basic reply with dhcpv4.NewReplyFromRequest, then looks up the lease associated with the MAC and pool with GetLease. If the lease is missing or lacks a ClientIP, it issues a warning and does not respond, preventing offers to unrecognized clients.

Population of IP and option fields: When the lease exists, the reply copies addresses (server, client, YourIP, Gateway, TransactionID, flags) and updates the main DHCP options: ServerIdentifier, subnet mask, router, DNS, domain name/search, NTP, and lease time (default 1 year if absent).

Response type selection: Based on the request's MessageType, the DHCPOFFER option is set for Discover or DHCPACK for Request; other types are simply logged as unhandled and do not produce a response.

Sending on the interface socket: The reply is sent via the PacketConn bound to the interface/pool; any write errors are logged. Informational logging accompanies both the incoming packet and the outgoing response.

How the handler is connected to DHCP servers
DHCPAllocator.Run creates a DHCP server for each DHCPNetConfig, binding UDP/67 to 0.0.0.0 but bound to the target interface. For each server, it constructs a closure (handlerWrapper) that captures the correct ipPoolRef and forwards each packet to the dhcpHandlerPerPool with the correct pool reference.

Each server runs in a goroutine and uses the same handler to maintain separate leases per pool while sharing the agent process. Context management ensures that servers shut down cleanly upon deletion.

davidepasquero:
> I quickly scanned through it, and one question came to my mind: are you proposing a fixed Helm-installed agent deployment instead of per-IPPool agent deployments? In other words, is it still in the operator pattern?

@starbops @joy717 @tlehman @fsdaniel 

If instead we want an architecture in which each IPpool is associated with a distribution, is this the right solution?
https://github.com/harvester/vm-dhcp-controller/pull/78

Leader election between pods has not yet been implemented so replication must remain at 1, but this already solves our problems when an agent is on a node that crashes

Issue #7902, opened 2025-03-19
[BUG] Upgrade Stuck in Pre-Draing if a Running Volume transit to Detached


WebberHuang1118:
**Describe the bug**
Harvester upgrade may stuck in `pre-draining` if a running volume transit to detached during the process

**To Reproduce**
1. Attaching a degraded volume `v1` on node `n1`
2. Upgrading a Harvester cluster
3. During `n1` start pre-draining phase, detach `v1`
4. Node `n1` will stuck in the `pre-draining` phase

**Expected behavior**
The upgrade process can continue

**Workaround**
Adding `v1` to the skip list

**Environment**
 - Harvester ISO version:  Harvester master branch `9829175dfc813c094bcce2f440a1d1252f6901f3`

**Additional context**
This is because we don't renew the checking volume list for the loop in `wait_longhorn_engines()`
- https://github.com/harvester/harvester/blob/4c350fda088f9ea7b86b351106056b7e7b0c0a9f/package/upgrade/upgrade_node.sh#L245-L257

cc @bk201 @starbops

TachunLin:
We also encounter this issue while upgrading a three nodes Harvester with one witness node from v1.5.1 to v1.6.1-rc2.
There the upgrade was stuck in `Pre-draining` on the witness node. 

  <img width="720" height="336" alt="Image" src="https://github.com/user-attachments/assets/68d63aeb-510c-4290-b073-b8c726425c46" />


Thanks @WebberHuang1118 for checking and applying the **workaround** by adding this to the pre-draining pod:

  ```
  $ touch /tmp/skip-<volume name>
  ```

The upgrade process can proceed to complete.

DamiaSan:
### Test plan

**Scenario 1**
This scenario force the upgrade process to wait for a volume to become healthy during pre-draining phase, and during this waiting time an already processed volume change its state:
1. Create an Harvester cluster with at least 3 nodes
2. Create 2 volumes, v1 and v2, with 3 replicas
3. Attach v1 on node 1 and write some data over it
4. Start the upgrade of the Harvester cluster
5. Delete one replica of v1 (this must be done before the start of the pre-draining phase of node1)
6. Attach v2 on some node
7. Ensure that after the replica rebuilding of v1 the upgrade process can continue without error.

**Scenario 2**
This scenario is the one that caused the opening of this issue:
1. Create an Harvester cluster with at least 3 nodes
2. Attaching a degraded volume v1 on node n1
[...]

PR #9178, opened 2025-09-18
fix(upgrade): renew volume list in wait_longhorn_engines

DamiaSan:
#### Problem:
During pre-drain phase, upgrade process wait for all longhorn volumes to have enough replicas. If an already processed volume change its status, the upgrade process may stuck if the volume transit into a state not ready for draining.

#### Solution:
The volume list is renewed for every loop iteration .

#### Related Issue(s):
https://github.com/harvester/harvester/issues/7902

#### Test plan:
Described in the issue description

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -270,46 +270,57 @@ recover_rancher_system_agent() {
 wait_longhorn_engines() {
   local node_count=$(kubectl get nodes --selector=harvesterhci.io/managed=true,node-role.harvesterhci.io/witness!=true -o json | jq -r '.items | length')
 
-  # For each running engine and its volume
-  kubectl get engines.longhorn.io -n longhorn-system -o json |
-    jq -r '.items | map(select(.status.currentState == "running")) | map(.metadata.name + " " + .metadata.labels.longhornvolume) | .[]' |
+  while true; do
+    local unhealthy_found=0
+
+    # For each running engine and its volume
     while read -r lh_engine lh_volume; do
       echo Checking running engine "${lh_engine}..."
 
+      if [ -f "/tmp/skip-$lh_volume" ]; then
+        echo "Skip $lh_volume."
+        continue
+      fi
+
       # Wait until volume turn healthy (except two-node clusters)
-      while [ true ]; do
-        if [ $node_count -gt 2 ];then
-          robustness=$(kubectl get volumes.longhorn.io/$lh_volume -n longhorn-system -o jsonpath='{.status.robustness}')
-          if [ "$robustness" = "healthy" ]; then
-            echo "Volume $lh_volume is healthy."
-            break
-          fi
-        else
-          # two node situation, make sure maximum two replicas are healthy
-          expected_replicas=2
-
-          # replica 1 case
-          volume_replicas=$(kubectl get volumes.longhorn.io/$lh_volume -n longhorn-system -o jsonpath='{.spec.numberOfReplicas}')
-          if [ $volume_replicas -eq 1 ]; then
-            expected_replicas=1
-          fi
-
-          ready_replicas=$(kubectl get engines.longhorn.io/$lh_engine -n longhorn-system -o json |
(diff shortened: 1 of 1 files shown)


WebberHuang1118:
Hi @DamiaSan 

Is there a test plan for the reviewers?

DamiaSan:
> Hi @DamiaSan
> 
> Is there a test plan for the reviewers?

These are the test I have made on a 2 node harvester's setup:

- create a script that call the function `wait_longhorn_engines`
- create one volume with 3 replicas (only 2 will run)
- create one volume with 2 replicas
- attach the volume with 3 replicas, meanwhile the other volume is detached
- delete one of the replica of the 3-replica volume
- start the script (in the loop there is only the 3-replica volume), it wait for the 3-replica volume to become "ready" again (a new replica is rebuilt)
- attach the 2-replica volume (before the rebuilding above complete)
- check that in the loop now there are both the volumes 
- the script exit successfully

WebberHuang1118 (approved):
LGTM, thanks.

By the way, suggest including an upgrade operation in the test plan, since the target issue ([#7902](https://github.com/harvester/harvester/issues/7902)) occurs during an upgrade.

Vicente-Cheng (approved):
thanks for the enhancement

Vicente-Cheng:
Hi @DamiaSan, 

Sorry, I rechecked the test plan.
> - attach the volume with 3 replicas, meanwhile the other volume is detached
> - delete one of the replica of the 3-replica volume

If you create 2-node cluster, I thought the 3-replica volume would not get healthy.
And the original issue is that we did not renew the whole volume list after we stuck in the while loop.

Could you recheck your test plan?
I thought we could simplify and refine it.

Thanks

WebberHuang1118:
Waiting for the test plan to be rechecked as well.

DamiaSan:
Hi @Vicente-Cheng , thanks for the recheck.

> If you create 2-node cluster, I thought the 3-replica volume would not get healthy.

The 3-replica volume isn't healthy, but it is ready for the upgrade because it has 2 ready replicas: https://github.com/harvester/harvester/blob/8369decbf8a83f7354ca084fd8cefc7beaf0b949/package/upgrade/upgrade_node.sh#L295



> I thought we could simplify and refine it.

Do you have some suggestion? 

The aim of this test plan is to ensure that the script, in every external loop iteration and so while it is waiting for some volume to become ready, recheck the entire volume list:
- deleting one of the replica of the 3-replica volume is the way I found to have the script waiting for a volume to become ready.
- attaching the 2-replica volume is the way to have in the volume list a volume that that wasn't present before.

WebberHuang1118:
IMO, for the test plan, a two-node cluster is rare in real-world cases, and the Harvester controller treats it as a special case. It’s better to use at least a three-node cluster.

I have no strong opinion; however, if possible, including the upgrade context in the test plan would be ideal, so the same plan can serve both PR reviewers and QA verification. We usually create one test plan that’s adaptable to different scenarios. cc @Vicente-Cheng @olblak

mingshuoqiu:
LGTM. Some suggestions may worth consideration
1. Consider adding a timeout or retry limit to avoid indefinite loops if volumes keep flipping states.
2. Wrap `kubectl` calls with retries to handle transient API failures.

DamiaSan:
> LGTM. Some suggestions may worth consideration
> 
> 1. Consider adding a timeout or retry limit to avoid indefinite loops if volumes keep flipping states.
> 2. Wrap `kubectl` calls with retries to handle transient API failures.

Good considerations, thanks!
However I think it is better to address these 2 enhancements in a different PR, the scope of this one is to refresh the volume list.

DamiaSan:
> IMO, for the test plan, a two-node cluster is rare in real-world cases, and the Harvester controller treats it as a special case. It’s better to use at least a three-node cluster.
> 
> I have no strong opinion; however, if possible, including the upgrade context in the test plan would be ideal, so the same plan can serve both PR reviewers and QA verification. We usually create one test plan that’s adaptable to different scenarios. cc @Vicente-Cheng @olblak

The PR have already a couple of approvals, so I am going to merge it.
But I will surely add the test case you mentioned in the test plan for the root issue, thanks.

Issue #9191, opened 2025-09-23
[BUG] Validator does not consider SC parameters misidentifying CDI default capabilities


brandboat:
### Describe the Bug

The validator only checks sc.Provisioner against cdicaps.CapabilitiesByProvisionerKey:

```
if _, ok := cdicaps.CapabilitiesByProvisionerKey[sc.Provisioner]; 
    !ok && sc.Provisioner != util.CSIProvisionerLonghorn && sc.Provisioner != util.CSIProvisionerLVM {
    // ...
}
```

This ignores StorageClass parameters (sc.Parameters). CDI’s default capabilities may be keyed by provisioner + params (see CDI [storagecapabilities.go#L330](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/storagecapabilities/storagecapabilities.go#L330)), so classes that rely on a param-qualified capability are missed. As a result, the validator mistakenly requires users to set the volumeMode/accessMode annotation even when CDI already defines defaults for that StorageClass.

### To Reproduce

Create a StorageClass whose CDI default capability depends on SC params (e.g., Trident with backendType: ontap-nas):

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: trident
provisioner: csi.trident.netapp.io
parameters:
  backendType: ontap-nas
```

Observe: the validator errors, claiming the user must set volumeMode/accessMode annotations.

### Expected Behavior

The validator should derive the capability key including SC params (consistent with CDI’s CapabilityKeyFromSCParams logic) and successfully match CDI’s default capabilities.

- If a param-qualified default exists: do not require volumeMode/accessMode annotations.
- Only require annotations when no matching CDI default capability is found.

### Support Bundle for Troubleshooting

n/a

### Environment

- Harvester version: v1.6.0
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Workaround and Mitigation

follow https://docs.harvesterhci.io/v1.6/advanced/storageclass/#containerized-data-importer-cdi-settings and add correct `cdi.harvesterhci.io/storageProfileVolumeModeAccessModes` value to storage class.

TachunLin:
Verified **fixed** on `v1.7.0-dev-20251026`. Close this issue.

### Result
[...]

PR #9193, opened 2025-09-23
Fix validator does not consider SC parameters misidentifying CDI default capabilities

brandboat:
#### Problem:
see https://github.com/harvester/harvester/issues/9191

#### Solution:
as title

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9191

#### Test plan:
- Positive case: create a storage class with parameters that the is defined in the [cdi storagecapabilities](https://github.com/kubevirt/containerized-data-importer/blob/99bcc36a931be02418192135dff7776f8a74471f/pkg/storagecapabilities/storagecapabilities.go#L37-L145)), make sure validator doesn't complain.
example:
  ```yaml
  apiVersion: storage.k8s.io/v1
  kind: StorageClass
  metadata:
    name: trident
  provisioner: csi.trident.netapp.io
  parameters:
    backendType: ontap-nas
  ```

- Negative case: create a storage class with parameters that the is NOT defined in the [cdi storagecapabilities], make sure validator raise error: `admission webhook "validator.harvesterhci.io" denied the request: missing annotation cdi.harvesterhci.io/storageProfileVolumeModeAccessModes. volume access modes are required for CDI integration to work with storage class provisioner csi.trident.netapp.io`.
e.g. trident.
  ```yaml
  apiVersion: storage.k8s.io/v1
  kind: StorageClass
  metadata:
    name: trident
  provisioner: csi.trident.netapp.io
  parameters:
    backendType: foo
  ```
  image

Note: There’s no need to install Trident in Kubernetes, since our goal is only to verify that the VM image creation validator works as expected.

### DIFF
--- a/pkg/webhook/resources/storageclass/validator.go
+++ b/pkg/webhook/resources/storageclass/validator.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	cdicaps "kubevirt.io/containerized-data-importer/pkg/storagecapabilities"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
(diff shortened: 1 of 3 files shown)


Vicente-Cheng (approved):
lgtm, thanks!

WebberHuang1118 (approved):
LGTM, thanks.

Vicente-Cheng:
@mergifyio backport v1.6

Issue #6183, opened 2024-07-17
[ENHANCEMENT] Enable selection of the upgrade Image


starbops:
**Is your enhancement related to a problem? Please describe.**

The current upgrade image selection is hardcoded in the upgrade controller's code. For instances,

- The prepare plan uses the previous version's `harvester-upgrade` container image: https://github.com/harvester/harvester/blob/1de4c2f8fd6f9214b313d39fab5e568c535fcd42/pkg/controller/master/upgrade/common.go#L316
- The upgrade-node job uses the new version's `harvester-upgrade` container image: https://github.com/harvester/harvester/blob/1de4c2f8fd6f9214b313d39fab5e568c535fcd42/pkg/controller/master/upgrade/common.go#L368
- The apply-manifest job uses the new version's `harvester-upgrade` container image: https://github.com/harvester/harvester/blob/1de4c2f8fd6f9214b313d39fab5e568c535fcd42/pkg/controller/master/upgrade/common.go#L464
- The container-image-cleanup plan uses the previous version's `harvester-upgrade` container image: https://github.com/harvester/harvester/blob/1de4c2f8fd6f9214b313d39fab5e568c535fcd42/pkg/controller/master/upgrade/common.go#L237

From developers' perspectives, sometimes it's difficult to verify issues or work on feature development related to upgrades. We're eager for an easy way to specify what upgrade image to use during Harvester upgrades.

**Describe the solution you'd like**

Add a new field called `.spec.upgradeImage` in the Upgrade CRD for users to choose what upgrade image they want to use. If such a field is not provided with any value, the upgrade controller falls back to the original logic.

**Describe alternatives you've considered**

Fake the upgrade image with the same name and tag, and then manually import the image on each node.

**Additional context**

#4926

votdev:
I've labeled the issue with `not-require/test-plan` because this issue is hard to test and is only meant for developers. The added unit tests should ensure that existing code does not break and the feature is working as expected.

@Vicente-Cheng Any thoughts or comments?

w13915984028:
The PR has been merged, I put further open questions here:

(1) It seems there is no local manual made upgrade test to show that the feature itself runs smoothly on the happy path.

(2) Let's talk a bit more about this feature:

Upgrade has a strong control on the versions (upgrade matrix), and a lot of checks are done to really start an upgrade. The upgrade is hop-by-hop, and single way, you can't down-grade a cluster.

[...]

PR #9196, opened 2025-09-23
Enable configuration of the upgrade image via CR

votdev:
#### Problem:
From developers' perspectives, sometimes it's difficult to verify issues or work on feature development related to upgrades. We're eager for an easy way to specify what upgrade image to use during Harvester upgrades.

#### Solution:
Add a new field called .spec.upgradeImage in the Upgrade CRD for users to choose what upgrade image they want to use. If such a field is not provided with any value, the upgrade controller falls back to the original logic.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/6183

#### Test plan:
n/a


pkg/apis/harvesterhci.io/v1beta1/upgrade.go
```diff
@@ -67,3 +75,10 @@ type NodeUpgradeStatus struct {
 	Reason  string `json:"reason,omitempty"`
 	Message string `json:"message,omitempty"`
 }
+
+// GetUpgradeImage returns the upgrade image as a Docker image reference, e.g. `rancher/harvester-upgrade:latest`.
+// If the `spec.upgradeImage` field is not provided, it constructs the image reference using the provided repository and tag.
+func (u *Upgrade) GetUpgradeImage(repository, tag string) string {
+	defaultValue := fmt.Sprintf("%s:%s", repository, tag)
```
w13915984028, line 82: given this feature is more for the convenience of develop,

when the target tag is a formal released version, e.g. v1.6.0, then it is better to skip the input

pkg/apis/harvesterhci.io/v1beta1/upgrade.go
```diff
@@ -43,6 +46,11 @@ type UpgradeSpec struct {
 	// +optional
 	// +kubebuilder:default:=true
 	LogEnabled bool `json:"logEnabled" default:"true"`
+
+	// Specifies the image to be used for upgrade operations.
+	// Please note that this field is intended for development and testing purposes only.
+	// +optional
+	UpgradeImage *string `json:"upgradeImage,omitempty"`
```
w13915984028, line 53: it is mainly for developing purpose, would prefer to use an label/annotation to to overwrite

normally when building on `harvester/harvester`, it generates:

`harvester,harvester-webhook, update the harvester,harvester-crd managedchart version, upgrade-repo, upgrade`,  sololy to replace the `upgrade` image is in high risk of inconsistency, only fits for dev mode

w13915984028 (approved):
a few questions, thanks

pkg/apis/harvesterhci.io/v1beta1/upgrade.go
votdev: > when the target tag is a formal released version, e.g. v1.6.0, then it is better to skip the input

Can you explain this please?

Does that mean we should parse the tag variable here and check whether it complies with semantic versioning? Don't you think we're over-engineering this?

w13915984028: There are alway users trying to use every possible options and report bugs; adding this to configuration might bring some potential troubles to us.

Even for dev engineers, if the installer-ISO and the customized image has big differences, the upgrade might also fail and needs more time to dig.

pkg/apis/harvesterhci.io/v1beta1/upgrade.go
votdev: On the one hand, I completely agree with you; annotations are very useful for things like this. But what's better about the current implementation is that it's documented. With annotations, at some point no one knows what they're for anymore because we don't document them anywhere.
Hopefully, anyone who sets the field in their CR is also aware of the implications.

Perhaps @starbops, as the creator of the issue, can comment on this.

starbops: I'm more inclined to the current implementation. Although annotations are a handy and easy way for us to add features to existing APIs, the drawback is that we need to maintain documentation for them separately. Unless we expose this toggle on the UI, either approach will not interfere with the existing user experience of upgrades. It can only be done directly via kubectl, which kind of aligns with the definition of a dev-only feature.

starbops (approved):
LGTM, thank you.

pkg/apis/harvesterhci.io/v1beta1/upgrade.go
starbops: Suppose a user create an Upgrade CR with `upgradeImage: v1.6.0` using the v1.7.0 ISO image and runs into issues. In that case, we can ask them "do you ever experience upgrade issues using the default v1.7.0 image?" and still focus on the supported upgrade path. So it's not a problem IMO.

Issue #8669, opened 2025-07-15
[ENHANCEMENT] Online Volume Resizing from the VM Manifest


WebberHuang1118:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

We will introduce online volume expansion starting with `v1.6.0` [HEP](https://github.com/harvester/harvester/blob/master/enhancements/20250612-volume-online-expansion.md)

By design, volumes can be expanded either from the Volume page or the VM page. However, we’ve identified a user experience issue when performing volume resizing from the VM page:
- https://github.com/harvester/harvester/issues/8625#issuecomment-3056246057
- https://github.com/harvester/harvester/issues/8625#issuecomment-3060254186

Therefore, the UI disables (gray out) volume resizing while the VM is running in `v1.6.0`.

To improve clarity and behavior, we propose the following enhancements:
- **Backend**: Add validation in the VM webhook to detect annotation changes related to volume expansion before applying them
- **GUI**: Suppress the message `Restart the virtual machine for configuration to take effect` when performing online volume resizing

@a110605 @houhoucoop please help with the GUI part, thanks.

cc @Vicente-Cheng

albinsun:
Test PASS via __Edit YAML__, close this backend ticket.
Frontend (__Edit Config__) is implementing and should be test in harvester/harvester/issues/9205.



## Environment
* `harvester-v1.7.0-dev-20250928`
  * Profile: Single node AMD64 Bare-metal HPE ProLiant DL360 
  * ui-source: `Auto`


## Test Steps
1. <details>
   <summary> Install the NFS CSI driver and create a third-party volume 
   </summary>

   Ref. https://github.com/WebberHuang1118/harvester-study/blob/master/nfs-in-cluster/README.md

   * `kubectl apply -f nfs-server.yaml`
      ```
      hp-7-tink-system:~/nfs-in-cluster # kubectl apply -f nfs-server.yaml
      persistentvolumeclaim/nfs-data created
      service/nfs-server created
      deployment.apps/nfs-server created
      ```

   * `curl -skSL https://raw.githubusercontent.com/kubernetes-csi/csi-driver-nfs/master/deploy/install-driver.sh | bash -s master -`
      ```
      hp-7-tink-system:~/nfs-in-cluster # curl -skSL https://raw.githubusercontent.com/kubernetes-csi/csi-driver-nfs/master/deploy/install-driver.sh | bash -s master -
      Installing NFS CSI driver, version: master ...
[...]

PR #9204, opened 2025-09-24
fix: enhance the VM webhook for volume expansion

WebberHuang1118:
#### Problem:
Volume can’t be expanded online from the VM page.

#### Solution:
Enhance the VM webhook for volume expansion

#### Related Issue(s):
#8669 

#### Test plan:
* Install the NFS CSI driver or another third-party storage driver.
* Create a VM with the third-party volume as the root disk.
* After the VM is running, change the volume size using a VM annotation.
  - e.q. 
    ```
    harvesterhci.io/volumeClaimTemplates: '[{"metadata":{"name":"vm-nfs-disk-0-hq6rz","annotations":{"harvesterhci.io/imageId":"default/image-lvsrb"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"8Gi"}},"volumeMode":"Block","storageClassName":"nfs-csi"}}]'
    ``` 
    to
    ```
    harvesterhci.io/volumeClaimTemplates: '[{"metadata":{"name":"vm-nfs-disk-0-hq6rz","annotations":{"harvesterhci.io/imageId":"default/image-lvsrb"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"16Gi"}},"volumeMode":"Block","storageClassName":"nfs-csi"}}]'
    ```
* This update will be rejected by the Harvester webhook.
* Enable `csi-online-expand-validation` as described [here](https://docs.harvesterhci.io/v1.7/advanced/index#csi-online-expand-validation) for the third-party driver.
* Change the volume size using a VM annotation again.
* The volume should be expanded.

### DIFF
--- a/pkg/webhook/resources/persistentvolumeclaim/validator.go
+++ b/pkg/webhook/resources/persistentvolumeclaim/validator.go
@@ -13,7 +13,6 @@ import (
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	cdicommon "kubevirt.io/containerized-data-importer/pkg/controller/common"
-	kvirtfeatures "kubevirt.io/kubevirt/pkg/virt-config/featuregate"
 
 	harvesterv1beta1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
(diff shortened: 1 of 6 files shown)


pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -131,90 +130,8 @@ func (v *pvcValidator) Delete(request *types.Request, oldObj runtime.Object) err
...
-				util.PVCExpandErrorPrefix+": pvc %s/%s is not online expandable with its provider",
-				pvc.Namespace,
-				pvc.Name,
-			),
-			"",
-		)
-	}
-
-	return nil
+	return webhookutil.CheckExpand(pvc, v.vmCache, v.kubevirtCache, v.engineCache, v.scCache, v.settingCache)
```
FrankYang0529: Do we want to use this function directly? It looks like `pvcValidator#checkExpand` only has one line.

FrankYang0529 (dismissed):
Overall LGTM. Thanks for the patch.

pkg/webhook/util/volume.go
```diff
@@ -60,3 +66,102 @@ func checkLonghornExpandability(pvc *corev1.PersistentVolumeClaim, engineCache c
...
+
+func isKubevirtExpandEnabled(kubevirt *kubevirtv1.KubeVirt) bool {
+	featureGates := kubevirt.Spec.Configuration.DeveloperConfiguration.FeatureGates
+	for _, f := range featureGates {
+		if f == kvirtfeatures.ExpandDisksGate {
+			return true
+		}
+	}
+
+	return false
```
brandboat: nit:
```suggestion
    return slices.Contains(featureGates, kvirtfeatures.ExpandDisksGate)
```

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -419,10 +458,24 @@ func (v *vmValidator) checkVolumeAnnotations(oldVM, newVM *kubevirtv1.VirtualMac
...
 }
 
+func (v *vmValidator) checkPVCStorageRequestChange(newPvc, oldPvc *corev1.PersistentVolumeClaim) error {
+	cmp := newPvc.Spec.Resources.Requests.Storage().Cmp(*oldPvc.Spec.Resources.Requests.Storage())
+	if cmp == -1 {
+		return werror.NewInvalidError(
+			fmt.Sprintf("%s PVC requests storage can't be less than previous value", newPvc.Name),
+			fmt.Sprintf("metadata.annotations.%s", util.AnnotationVolumeClaimTemplates),
+		)
+	}
```
brandboat, line 468: dumb question: do we need to add this check to pvc validator too (pkg/webhook/resources/persistentvolumeclaim/validator.go#Update)? we can't decrease the existing pvc size, right?

pkg/webhook/resources/persistentvolumeclaim/validator.go
WebberHuang1118: Updated as suggested, thank you.

pkg/webhook/util/volume.go
WebberHuang1118: Updated as suggested, thank you.

pkg/webhook/resources/virtualmachine/validator.go
WebberHuang1118: Because in this context, we only update the VM annotation, and the VM controller handles the volume size update under the hood. If the webhook doesn’t check this condition, the UI won’t surface the PVC update error, leaving users unaware of what happened. Therefore, we need to enforce this check in the VM webhook.

brandboat (approved):
LGTM, thanks!

FrankYang0529 (approved):

Issue #9187, opened 2025-09-22
[TASK] Bump Longhorn v1.10.0 for Harvester v1.7.0


Vicente-Cheng:
### Task request

Bump Longhorn v1.10.0 for Harvester v1.7.0 with the following tasks
- [x] Chart bump
- [x] ~Library bump~ - blocked until kubevirt bump client-go 0.33.x

mcerveny:
Long awaited higlighted feature "interrupt mode" is not compatible with NVMe disks. https://github.com/longhorn/longhorn/issues/11816

Vicente-Cheng:
PR merged, we will test the new Longhorn with any generic path. Close it directly

Issue #9102, opened 2025-09-09
[TASK] Component updates for Harvester v1.7.0


Vicente-Cheng:
### Component update request

Components to be updated for Harvester version v1.7.0:

- System Components
  - [x] OBS Projects @tserong 
  - [x] OS2 @tserong 
- Main Components
  - [x] Embedded Rancher @FrankYang0529
  - [x] RKE2 @FrankYang0529
  - [x] Longhorn @Vicente-Cheng - Target version 1.10.1
  - [x] KubeVirt @Vicente-Cheng (also virtctl) (@Yu-Jack helped with that)
- Libraries/Base Image (https://github.com/harvester/harvester/issues/9279)
  - [x] BCI Image @harvester/dev - Target version 15.7
  - [x] Go @harvester/dev - Target version v1.25
  - [x] K8s @harvester/dev - Target version v1.32/v1.33 (depends on the upstream, e.g. kubevirt)
- Built-in Dependent Components
  - [x] * Network Controller @mingshuoqiu (will be tagged by bot automatically)
  - [x] * Network FS Manager @Vicente-Cheng @WebberHuang1118 (will be tagged by bot automatically)
  - [x] * Node Disk Manager @tserong  @Vicente-Cheng (will be tagged by bot automatically)
  - [x] * Node Manager @tserong (will be tagged by bot automatically)
  - [x] * Load Balancer @starbops (will be tagged by bot automatically)
  - [x] * Eventrouter @w13915984028 (will be tagged by bot automatically)
  - [x] CDI @Vicente-Cheng (Already v1.62.0, we don't have any new version like v1.63.1 image on SLES15SP7)
  - [x] Whereabouts @rrajendran17
  - [x] Kube-VIP @starbops (by @w13915984028  https://github.com/harvester/charts/pull/404 Bump kube-vip chart to 0.8.0 with app image tag v0.9.2)
  - [x] CSI Snapshotter @brandboat  (need to align with Longhorn version) https://github.com/harvester/harvester/pull/9443 https://github.com/harvester/harvester-installer/pull/1180
  - [x] snapshot-validation-webhook @brandboat https://github.com/harvester/harvester/pull/9443 https://github.com/harvester/harvester-installer/pull/1180
- Built-in Opt-In Add-Ons
  - [x] * VM Import Controller @votdev (will be tagged by bot automatically)
  - [x] * PCI Device Controller @Yu-Jack (will be tagged by bot automatically)
[...]

PR #9189, opened 2025-09-22
Bump Longhorn v1.10.0-rc3 for Harvester v1.7.0

Vicente-Cheng:
#### Problem:
Bump Longhorn v1.10.0-rc3

#### Solution:
Bump Longhorn v1.10.0-rc3

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9187
https://github.com/harvester/harvester/issues/9102

PR #9215, opened 2025-09-25
charts: bump Longhorn v1.10.0

Vicente-Cheng:
#### Problem:
Bump Longhorn v1.10.0

#### Solution:
Bump Longhorn v1.10.0

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9187


WebberHuang1118:
Do we need to update this [go.mod entry](https://github.com/harvester/harvester/blob/5a6f574e57c8471a33303f6725ec254a8055783b/go.mod#L90)? 

The backupstore used by Longhorn v1.10.0 is [v0.0.0-20250804022317-794abf817297](https://github.com/longhorn/longhorn-manager/blob/22af8fe21683753eb380348eb512a17ee87473a9/go.mod#L69). Should we align with Longhorn Manager?

Vicente-Cheng:
> Do we need to update this [go.mod entry](https://github.com/harvester/harvester/blob/5a6f574e57c8471a33303f6725ec254a8055783b/go.mod#L90)?
> 

Here is the reason https://github.com/harvester/harvester/pull/9189#issuecomment-3322155125
`k8s.io/client-go` v0.33 (for longhorn-manager) would conflict with `kubevirt.io/client-go`

> The backupstore used by Longhorn v1.10.0 is [v0.0.0-20250804022317-794abf817297](https://github.com/longhorn/longhorn-manager/blob/22af8fe21683753eb380348eb512a17ee87473a9/go.mod#L69). Should we align with Longhorn Manager?

Do we use the newer backupstore API, which is only provided on Longhorn v1.10.0?

WebberHuang1118:
> > Do we need to update this [go.mod entry](https://github.com/harvester/harvester/blob/5a6f574e57c8471a33303f6725ec254a8055783b/go.mod#L90)?
> 
> Here is the reason [#9189 (comment)](https://github.com/harvester/harvester/pull/9189#issuecomment-3322155125) `k8s.io/client-go` v0.33 (for longhorn-manager) would conflict with `kubevirt.io/client-go`
> 
> > The backupstore used by Longhorn v1.10.0 is [v0.0.0-20250804022317-794abf817297](https://github.com/longhorn/longhorn-manager/blob/22af8fe21683753eb380348eb512a17ee87473a9/go.mod#L69). Should we align with Longhorn Manager?
> 
> Do we use the newer backupstore API, which is only provided on Longhorn v1.10.0?

In fact, we’ve been using the same backupstore API since day one, so the latest version shouldn’t be mandatory. I think we can stick with the current policy outlined [here](https://github.com/harvester/harvester/pull/9189#issuecomment-3322155125).

WebberHuang1118 (approved):
LGTM, thanks.

Issue #8422, opened 2025-06-04
[ENHANCEMENT] Remove validation that prevents RWX block volumes from being shared among VMs


innobead:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Remove the validation logic to allow RWX block volumes to be attached to multiple VMs if the underlying storage solution supports it. Harvester should delegate this validation or behavior to KubeVirt instead of implementing restrictive prevention.

[/pkg/webhook/resources/virtualmachine/validator.go#L461-L478](https://github.com/harvester/harvester/blob/7a98d6231aad3af89650ea0c379bcd359b96efc1/pkg/webhook/resources/virtualmachine/validator.go#L461-L478)
```go
func (v *vmValidator) checkOccupiedPVCs(vm *kubevirtv1.VirtualMachine) error {
	for _, volume := range vm.Spec.Template.Spec.Volumes {
		if volume.PersistentVolumeClaim != nil {
			vms, err := v.vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(vm.Namespace, volume.PersistentVolumeClaim.ClaimName))
			if err != nil {
				return werror.NewInternalError(fmt.Sprintf("failed to get VMs by index: %s, PVC: %s/%s, err: %s", indexeresutil.VMByPVCIndex, vm.Namespace, volume.PersistentVolumeClaim.ClaimName, err))
			}
			for _, otherVM := range vms {
				if otherVM.Namespace != vm.Namespace || otherVM.Name != vm.Name {
					message := fmt.Sprintf("the volume %s is already used by VM %s/%s", volume.PersistentVolumeClaim.ClaimName, otherVM.Namespace, otherVM.Name)
					return werror.NewInvalidError(message, "spec.templates.spec.volumes")
				}
			}
		}
	}

	return nil
}
```

### Additional context

cc @harvester/storage

abonillabeeche:
Requesting for this to please be changed to 1.6.1

forbesguthrie:
x.y.1 releases should only contain critical fixes or security patches (CVEs). There should be no feature enhancements. This is part of the new 4-month release cycle.

lanfon72:
@Vicente-Cheng I need to confirm that:
> Test plan:
Ensure the RWX volume of 3rd-party storage solution can attach to two different VMs

[...]

PR #9216, opened 2025-09-25
webhook/vm: mitigate the limitation for RWX volume

Vicente-Cheng:
#### Problem:
Some RWX volume (block mode) could support IO with different node. For example, support SCSI PR Command. Since we support 3rd-party Storage solution, we should mitigate this limitation.

#### Solution:
Now we only check the RWO volume and RWX volume but from the Longhorn.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8422

#### Test plan:
Ensure the RWX volume of 3rd-party storage solution can attach to two different VMs

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -517,7 +517,30 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
 }
 
 func (v *vmValidator) checkOccupiedPVCs(vm *kubevirtv1.VirtualMachine) error {
+	// only block two scenarios:
+	// - RWO volume
+	// - Longhorn volume
+	// other scenarios like RWX volume, user should know what they are doing
 	for _, volume := range vm.Spec.Template.Spec.Volumes {
+		// only check PVC volumes
+		if volume.PersistentVolumeClaim == nil {
+			continue
+		}
+		name := volume.PersistentVolumeClaim.ClaimName
+		pvc, err := v.pvcCache.Get(vm.Namespace, name)
+		if apierrors.IsNotFound(err) {
+			// means runtime creation, no need to check
+			continue
+		}
+		if err != nil {
+			// any error here should be raised
+			return werror.NewInternalError(fmt.Sprintf("failed to get PVC %s/%s, err: %s", vm.Namespace, name, err))
+		}
+		targetAccessMode := pvc.Spec.AccessModes
+		targetProvisioner := util.GetProvisionedPVCProvisioner(pvc, v.scCache)
+		if volumeSupportRWXForVM(targetAccessMode, targetProvisioner) {
+			continue
+		}
 		if volume.PersistentVolumeClaim != nil {
 			vms, err := v.vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(vm.Namespace, volume.PersistentVolumeClaim.ClaimName))
 			if err != nil {
@@ -597,4 +620,18 @@ func (v *vmValidator) checkReservedMemoryAnnotation(vm *kubevirtv1.VirtualMachin
 
 func (v *vmValidator) checkStorageResourceQuota(vm *kubevirtv1.VirtualMachine, oldVM *kubevirtv1.VirtualMachine) error {
 	return v.rqCalculator.CheckStorageResourceQuota(vm, oldVM)
+}
+
+func volumeSupportRWXForVM(accessModes []corev1.PersistentVolumeAccessMode, provisioner string) bool {
+	if provisioner == util.CSIProvisionerLonghorn {
+		// Longhorn provisioner does not support RWX volume for VM
+		return false
+	}
+
+	for _, mode := range accessModes {
+		if mode == corev1.ReadWriteMany {
+			return true
+		}
+	}
+	return false
 }


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -462,7 +465,27 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
 }
 
 func (v *vmValidator) checkOccupiedPVCs(vm *kubevirtv1.VirtualMachine) error {
+	// only block two scenarios:
+	// - RWO volume
+	// - Longhorn volume
```
brandboat, line 522: dumb question: where do we block rwo+longhorn volume?

brandboat (dismissed):
Just one question, overall LGTM, thanks!

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -462,7 +465,27 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
...
+		pv, err := v.pvCache.Get(name)
+		if err != nil {
+			// any error here should be raised
+			return werror.NewInternalError(fmt.Sprintf("failed to get PV %s/%s, err: %s", vm.Namespace, name, err))
+		}
+		targetAccessMode := pv.Spec.AccessModes
+		targetCSIDriver := ""
+		if pv.Spec.CSI != nil {
+			targetCSIDriver = pv.Spec.CSI.Driver
+		}
```
WebberHuang1118: Should we use the logic in [pkg/util/volume.go (L42–L63)](https://github.com/Vicente-Cheng/harvester/blob/e3375e6276c61a5e8a9c74da0d1088878f1fce72/pkg/util/volume.go#L42-L63) to determine the storage provider for implementation consistency?

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -462,7 +465,27 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
...
+		for _, mode := range targetAccessMode {
+			if mode == corev1.ReadWriteMany && targetCSIDriver != util.CSIProvisionerLonghorn {
+				continue
```
WebberHuang1118: Suggest factoring out this part for better readability

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -462,7 +465,27 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
 }
 
 func (v *vmValidator) checkOccupiedPVCs(vm *kubevirtv1.VirtualMachine) error {
+	// only block two scenarios:
+	// - RWO volume
+	// - Longhorn volume
+	// other scenarios like RWX volume, user should know what they are doing
```
WebberHuang1118, line 523: Does this mean users need to ensure they are using a storage provider that supports RWX?

WebberHuang1118 (commented):
LGTM overall—just a few questions and one suggestion, thanks.

pkg/webhook/resources/virtualmachine/validator.go
WebberHuang1118: For RWO volumes and Longhorn volumes, we need to check whether the volume is attached to more than one VM. This is the basic principle for RWO volumes. In the case of Longhorn volumes, since we use block-mode PVCs, each can also only be attached to a single VM.

pkg/webhook/resources/virtualmachine/validator.go
Vicente-Cheng: Good suggestion, I missed that we have those helper functions.
Updated.

pkg/webhook/resources/virtualmachine/validator.go

pkg/webhook/resources/virtualmachine/validator.go
Vicente-Cheng: yes, we will have another document update for this change.

pkg/webhook/resources/virtualmachine/validator.go
Vicente-Cheng: like @WebberHuang1118 mentioned.
For RWO, sure, we need to block.
For Longhorn volume, because they did not support SCSI PR with their frontend (iSCSI), concurrent writes across multiple nodes are not allowed. We need to block it.

WebberHuang1118 (dismissed):
LGTM, thanks.

Vicente-Cheng:
fix the CI failure.
Now, we will skip two cases.
- non pvc disks
- on-demand pvc disks (no multiple attachment case)

brandboat (dismissed):

ihcsim (dismissed):

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -517,7 +517,30 @@ func (v *vmValidator) checkGoldenImage(vm *kubevirtv1.VirtualMachine) error {
...
+		name := volume.PersistentVolumeClaim.ClaimName
+		pvc, err := v.pvcCache.Get(vm.Namespace, name)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				// means runtime creation, no need to check
+				continue
+			}
+			// any error here should be raised
+			return werror.NewInternalError(fmt.Sprintf("failed to get PVC %s/%s, err: %s", vm.Namespace, name, err))
+		}
```
WebberHuang1118, line 538:
```suggestion
        if apierrors.IsNotFound(err) {
           // means runtime creation, no need to check
           continue;
        }
       if err != nil {
	   // any error here should be raised
	  return werror.NewInternalError(fmt.Sprintf("failed to get PVC %s/%s, err: %s", vm.Namespace, name, err))
	}
```

WebberHuang1118 (dismissed):
LGTM, just a nit, thanks.

brandboat (approved):

WebberHuang1118 (approved):

PR #9217, opened 2025-09-26
Deprecating Harvester upgrade-repo virtual machine

brandboat:
#### Problem:

see issue desc

#### Solution:

This change replaces the VM-based upgrade repository with a Deployment-based approach starting from v1.7.0, while maintaining backward compatibility with v1.6.x VM-based upgrades.

###### Implementation Details:

- **Deployment Creation**: nginx container with 3 replicas serving ISO via HTTP, mounted from PVC using `mount -o loop,ro`, includes liveness/readiness probes
- **VM Image**: Uses `longhorn-static` storage class and `cdi` backend, annotated with `harvesterhci.io/os-upgrade-image: "True"` to trigger vm_image_controller auto-creating data volume (PVC) with RWX and Filesystem mode.
- **Controller Layer Changes**: Added `deployment_controller.go` to manage repo Deployment lifecycle and removed `pod_controller.go` as now we are using deployment-based repo instead of VM which is a pod.

**Important:**

- **v1.6.x → v1.7.x**: Still uses **VM-based** repo (old Harvester code creates the repo)
- **v1.7.x → v1.8.x**: Uses **Deployment-based** repo (new Harvester code creates the repo)

This is because the upgrade repo is created **before** `apply-manifests`, so the previous Harvester version's code is used. Therefore, dual resource type support is mandatory throughout the codebase.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/7101

#### Test plan:
1. Upgrade from v1.6.x to master
    - The upgrade should complete successfully (verify on both single-node and multi-node setups).
    - During this process, the upgrade repo is still VM-based. In the preload stage, you should see the repo VM running as usual. (VM name: `harvester-system/upgrade-repo-<UPGRADE_NAME>`)
    - Images from the previous upgrade (e.g., v1.6.0) should be cleaned up afterward. Once cleanup is finished, the Upgrade CR will include the annotation: `harvesterhci.io/image-cleanup-plan-completed: "true"`
2.  Upgrade from master to master
    - The upgrade should also complete successfully (verify on both single-node and multi-node setups).
    - In this process, the upgrade repo is Deployment-based. During the preload stage, confirm that a Deployment named `upgrade-repo-<UPGRADE_NAME>` exists.
    - For master-to-master upgrades, post-upgrade image cleanup does not apply, since image lists are only defined for official release versions (e.g., v1.5.0, v1.6.0), not for development versions like master. Therefore, the Upgrade CR will show the annotation: `harvesterhci.io/image-cleanup-plan-completed: "false"`
3. Test 1. & 2. with image upload/download process (ui : settings>version>3-dots button>upgrade).
 
N/A


pkg/controller/master/upgrade/upgrade_repo.go
```diff
@@ -556,3 +383,115 @@ func (r *Repo) getImagesDiffList() ([]string, error) {
...
+					Labels: map[string]string{
+						harvesterUpgradeLabel:          r.upgrade.Name,
+						harvesterUpgradeComponentLabel: upgradeComponentRepo,
+					},
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:            "nginx-iso-server",
+							Image:           fmt.Sprintf("%s:%s", util.HarvesterUpgradeImageRepository, r.upgrade.Status.PreviousVersion),
```
w13915984028, line 478: what's the context to use `PreviousVersion`, not `.spec.version` of current ISO version

pkg/controller/master/upgrade/upgrade_repo.go
```diff
@@ -556,3 +383,115 @@ func (r *Repo) getImagesDiffList() ([]string, error) {
...
+							ImagePullPolicy: corev1.PullIfNotPresent,
+							Command:         []string{"sh", "-c"},
```
w13915984028, line 480: use `.spec.affinity` to avoid all 3 replicas pods are running on same host?

pkg/controller/master/upgrade/upgrade_repo.go
```diff
@@ -556,3 +383,115 @@ func (r *Repo) getImagesDiffList() ([]string, error) {
...
+			Labels: map[string]string{
+				harvesterUpgradeLabel:          r.upgrade.Name,
+				harvesterUpgradeComponentLabel: upgradeComponentRepo,
+			},
+			OwnerReferences: []metav1.OwnerReference{
+				upgradeReference(r.upgrade),
+			},
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: ptr.To(int32(3)),
```
w13915984028: is `3` hard coded? could we convert it to a param of the CreateDeployment and allow the customization,

e.g. for single-node upgrade, spawn a single replica deployment

w13915984028 (commented):
Thanks for the optimization upon upgrade-repo, a few questions.

BTW, some other questions:
(1) Do we need to add a check upon the repo, controller assumes the deployment ready is repo ready;  but the best option is to do a curl upon the service endpoint to ensure it really serves

(2) Do we need to protect the repo deployment from being deleted in-between the upgrade

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -231,16 +234,32 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+		}
+
+		// Check if both VM and Deployment are missing for backward compatibility:
+		// - versions before v1.7.0: Uses VM for preload stage
+		// - v1.7.0+: Uses Deployment only
+		// - v1.6.x to v1.7.x upgrade: Still requires VM during transition
+		vmNotFound := errVM != nil && apierrors.IsNotFound(errVM)
+		deploymentNotFound := errDeployment != nil && apierrors.IsNotFound(errDeployment)
+		if vmNotFound && deploymentNotFound {
+			logrus.Errorf("Both repo VM and Deployment %s/%s not found", repo.GetVMNamespace(), repo.GetVMName())
```
Vicente-Cheng: If the error is raised in the upper layer, do we still need a logger here?

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -231,16 +234,32 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+			return nil, errVM
+		}
+		if errDeployment != nil && !apierrors.IsNotFound(errDeployment) {
+			logrus.Warnf("Failed to get repo deployment %s/%s from cache, error %s", repo.GetVMNamespace(), repo.GetVMName(), errDeployment.Error())
+			return nil, errDeployment
+		}
+
+		// Check if both VM and Deployment are missing for backward compatibility:
+		// - versions before v1.7.0: Uses VM for preload stage
+		// - v1.7.0+: Uses Deployment only
```
Vicente-Cheng, line 268: I am curious, if this code just exists on the v1.7.0+, why do we still need to handle the VM case?
Seems the v1.7.0+ should move to deployment, why do we just drop the VM code?

pkg/controller/master/upgrade/upgrade_repo.go
Vicente-Cheng: Yes, even though we have 3 CP nodes, maybe two replicas would be enough. 2 replicas could keep us alive with an upgrade repo. The single node looks like a special case, but yes, we need to fix it to 1 with that case.

Vicente-Cheng (commented):
generally lgtm, just some nits

pkg/controller/master/upgrade/upgrade_controller.go
brandboat: This is because the upgrade repo is created before apply-manifests, if user upgrade from v.1.6.x, the upgrade-repo is still vm based, that's why we still need to take care the VM part.

pkg/controller/master/upgrade/upgrade_repo.go
brandboat: The newer rancher/harvester-upgrade image hasn’t been imported yet, so we’re still using the old one.

brandboat: I'll add some comment here.

pkg/controller/master/upgrade/upgrade_repo.go
brandboat: Good points, thanks!

brandboat:
@w13915984028, thanks for the input!

> (1) Do we need to add a check upon the repo, controller assumes the deployment ready is repo ready; but the best option is to do a curl upon the service endpoint to ensure it really serves

We already have ReadinessProbe, perhaps that's enough?
```
	ReadinessProbe: &corev1.Probe{
		ProbeHandler: corev1.ProbeHandler{
			HTTPGet: &corev1.HTTPGetAction{
				Path:   "/harvester-iso/harvester-release.yaml",
				Port:   intstr.FromInt(80),
				Scheme: corev1.URISchemeHTTP,
			},
		},
	},
```

> (2) Do we need to protect the repo deployment from being deleted in-between the upgrade

Sure, I'll add another deployment validator.

pkg/controller/master/upgrade/upgrade_repo.go
starbops: Or maybe set the `management.cattle.io/scale-available` annotation to `3` if we don't mind relying on Rancher to handle the auto-scaling.

pkg/controller/master/upgrade/upgrade_controller.go
starbops: IIUC, we can only come back and clean up the code specifically for the VM after we start to work on v1.8.0. Is that correct?

brandboat: Yep, 1.8.0 is the version that we can totally get rid of repo vm.

brandboat: The assumption is that user won't upgrade from v1.6.x to v1.8.x

pkg/controller/master/upgrade/upgrade_repo.go
```diff
@@ -556,3 +383,115 @@ func (r *Repo) getImagesDiffList() ([]string, error) {
...
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:            "nginx-iso-server",
+							Image:           fmt.Sprintf("%s:%s", util.HarvesterUpgradeImageRepository, r.upgrade.Status.PreviousVersion),
+							ImagePullPolicy: corev1.PullIfNotPresent,
+							Command:         []string{"sh", "-c"},
+							Args: []string{
+								`echo "Mounting ISO and starting Nginx..."
```
w13915984028, line 482: did you test, if any of the shell command fails, what's status the pod will be?

pkg/controller/master/upgrade/upgrade_repo.go
w13915984028: I did not get the point that we need to use the old one to start the upgrade-reo.

The `repo.Bootstrap` is called in below case, it shows `ImageReady`.

```
if harvesterv1.ImageReady.IsTrue(upgrade) && harvesterv1.RepoProvisioned.GetStatus(upgrade) == "" {
		logrus.Info("Starting upgrade repo VM")
		if err := repo.Bootstrap(); err != nil && !apierrors.IsAlreadyExists(err) {
```

w13915984028 (commented):
thx.

pkg/controller/master/upgrade/upgrade_repo.go
brandboat: We can't access new image until we finished the preload job on all nodes, right?
The repo deployment is happen before the preload stage, so I can't use the new image as they are not even loaded in the harvest node.
Or did I misunderstand anything 😕?

w13915984028: got it.

pkg/controller/master/upgrade/upgrade_repo.go
brandboat: Just tested, the pod will stay in `CrashLoopBackOff`

pkg/controller/master/upgrade/deployment_controller.go
```diff
@@ -0,0 +1,48 @@
...
+	if deployment == nil || deployment.DeletionTimestamp != nil || deployment.Labels == nil || deployment.Namespace != upgradeNamespace || deployment.Labels[harvesterUpgradeLabel] == "" {
+		return deployment, nil
+	}
+
+	upgradeControllerLock.Lock()
+	defer upgradeControllerLock.Unlock()
+
+	upgrade, err := h.upgradeCache.Get(upgradeNamespace, deployment.Labels[harvesterUpgradeLabel])
+	if err != nil {
+		return nil, err
```
Vicente-Cheng, line 32: should we take care this err?
I mean will return nil trigger another reconcile?

Vicente-Cheng (dismissed):
just a nit, overall lgtm. Thanks!

pkg/controller/master/upgrade/deployment_controller.go
brandboat: I think there is no need to reconcile if the upgrade CR is not found, which means the upgrade is either not created (which shouldn't happen) or it was deleted unexpectedly while the deployment still exists. In both cases, there is nothing actionable in the controller, we can just leave an error log and move on here.

For any other error, we should trigger reconcile since it indicates a real issue in cache#get.

@Vicente-Cheng, please correct me if I misunderstand anything, thanks!

starbops (commented):
What's the final decision on the replicas number? I saw there is only one repo pod running in a 3-node cluster setup.

brandboat:
> What's the final decision on the replicas number? I saw there is only one repo pod running in a 3-node cluster setup.

In single node, the replica should be 1, in multi-node, the replica should be 2. But from you description, it seems something went wrong... I'll test again.

brandboat:
Hi @starbops, I fixed the incorrect replica count issue you mentioned above, I shouldn't use singleNode field in upgrade CR as singleNode remains in empty string until upgrade repo deployment is provisioned. Fixed in 0269fe6.

starbops (approved):
LGTM, thank you for the great enhancement!

PR #9220, opened 2025-09-29
Skip cron job if node is in or going into maintenance mode

votdev:
#### Related Issue(s):
SURE-10547

### DIFF
--- a/pkg/controller/master/node/maintain_controller.go
+++ b/pkg/controller/master/node/maintain_controller.go
@@ -18,7 +18,7 @@ import (
 
 const (
 	maintainNodeControllerName  = "maintain-node-controller"
-	MaintainStatusAnnotationKey = "harvesterhci.io/maintain-status"
+	MaintainStatusAnnotationKey = util.AnnotationMaintainStatus
 	MaintainStatusComplete      = "completed"
 	MaintainStatusRunning       = "running"
 )
@@ -121,7 +121,7 @@ func (h *maintainNodeHandler) OnNodeRemoved(_ string, node *corev1.Node) (*corev
 		return node, nil
 	}
 
-	if _, ok := node.Annotations[MaintainStatusAnnotationKey]; !ok {
+	if !util.IsNodeInMaintenanceMode(node) {
 		return node, nil
 	}
 
--- a/pkg/controller/master/schedulevmbackup/common.go
+++ b/pkg/controller/master/schedulevmbackup/common.go
@@ -9,6 +9,7 @@ import (
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	"go.uber.org/multierr"
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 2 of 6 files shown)


WebberHuang1118:
Hi @votdev,

In my opinion, this shouldn’t be necessary. The Scheduled VM Backup controller creates a job to trigger the creation of the VMBackup resource (along with the Longhorn snapshot and backup). The Kubernetes scheduler will then assign the job to suitable nodes. As long as the cluster is functional and the Longhorn volume has at least one replica on a working node, the Scheduled VM Backup should work as expected.

votdev:
Thanks for your comment @WebberHuang1118. As you mentioned, it should work in theory, but it doesn't work in a user's error report. That's why i'm trying to fix this behavior as simply as possible with this code.

It now appears that the cron job was activated during draining. If the VM had not yet been moved to another node at that moment, the backup would have been started. Therefore, the `status.ownerID` of the backup is still that of the node that was actually supposed to be set to [maintenance mode](https://github.com/harvester/harvester/blob/0073cca8fa27743f325458717dddf1dcbaacdcb9/pkg/controller/master/nodedrain/nodedrain_controller.go#L223) and [marked as unscheduled](https://github.com/harvester/harvester/blob/47633a88313069b527e306ca1fb3e7fcbee4adaa/vendor/k8s.io/kubectl/pkg/drain/cordon.go#L89).

WebberHuang1118:
If we’re talking about [[Scheduling Virtual Machine Backups and Snapshots](https://docs.harvesterhci.io/v1.7/vm/backup-restore#scheduling-virtual-machine-backups-and-snapshots)](https://docs.harvesterhci.io/v1.7/vm/backup-restore#scheduling-virtual-machine-backups-and-snapshots):

The cronjob itself is only responsible for sleeping 10 seconds:

* [[common.go (L38–L39)](https://github.com/harvester/harvester/blob/a20c4beb39238267ab3d941d89ba2b74f160206d/pkg/controller/master/schedulevmbackup/common.go#L38-L39)](https://github.com/harvester/harvester/blob/a20c4beb39238267ab3d941d89ba2b74f160206d/pkg/controller/master/schedulevmbackup/common.go#L38-L39)
* [[svmbackup.go (L36–L83)](https://github.com/harvester/harvester/blob/a20c4beb39238267ab3d941d89ba2b74f160206d/pkg/controller/master/schedulevmbackup/svmbackup.go#L36-L83)](https://github.com/harvester/harvester/blob/a20c4beb39238267ab3d941d89ba2b74f160206d/pkg/controller/master/schedulevmbackup/svmbackup.go#L36-L83)

Then the Harvester controller reconciles the cronjob CR to create a new VMBackup:

* [[cronjob.go (L38–L40)](https://github.com/harvester/harvester/blob/a20c4beb39238267ab3d941d89ba2b74f160206d/pkg/controller/master/schedulevmbackup/cronjob.go#L38-L40)](https://github.com/harvester/harvester/blob/a20c4beb39238267ab3d941d89ba2b74f160206d/pkg/controller/master/schedulevmbackup/cronjob.go#L38-L40)

Regarding the situation you mentioned:

> It now appears that the cron job was activated during draining. If the VM had not yet been moved to another node at that moment, the backup would have been started. Therefore, the status.ownerID of the backup is still that of the node that was actually supposed to be set to [maintenance mode](https://github.com/harvester/harvester/blob/0073cca8fa27743f325458717dddf1dcbaacdcb9/pkg/controller/master/nodedrain/nodedrain_controller.go#L223) and [marked as unscheduled](https://github.com/harvester/harvester/blob/47633a88313069b527e306ca1fb3e7fcbee4adaa/vendor/k8s.io/kubectl/pkg/drain/cordon.go#L89)

Since the cronjob only creates a job that sleeps for 10 seconds, it won’t interfere with node draining. If a Longhorn backup is already in progress, Longhorn will use a PDB to delay the draining until the backup is complete.

votdev:
Closing because the Jira issue has been closed as well.

Issue #9223, opened 2025-09-30
[BUG] deploying v1.7.0-dev-20250928 fails: longhorn defaultSettings must be strings


tserong:
### Describe the Bug

Deploying Harvester v1.7.0-dev-20250928 gets stuck due to errors like:

```
# kubectl get bundle -A
NAMESPACE     NAME                         BUNDLEDEPLOYMENTS-READY   STATUS
fleet-local   fleet-agent-local            1/1                       
fleet-local   mcc-harvester                0/1                       ErrApplied(1) [Cluster fleet-local/local: execution error at (harvester/charts/longhorn/templates/default-setting.yaml:50:8): defaultSettings.storageReservedPercentageForDefaultDisk must be a string]
fleet-local   mcc-harvester-crd            1/1                       
fleet-local   mcc-kubeovn-operator-crd     1/1                       
fleet-local   mcc-rancher-logging-crd      1/1                       
fleet-local   mcc-rancher-monitoring-crd   1/1                       
```

```
# kubectl get bundle -A
NAMESPACE     NAME                         BUNDLEDEPLOYMENTS-READY   STATUS
fleet-local   fleet-agent-local            1/1                       
fleet-local   mcc-harvester                0/1                       ErrApplied(1) [Cluster fleet-local/local: execution error at (harvester/charts/longhorn/templates/default-setting.yaml:236:8): defaultSettings.concurrentAutomaticEngineUpgradePerNodeLimit must be a string]
fleet-local   mcc-harvester-crd            1/1                       
fleet-local   mcc-kubeovn-operator-crd     1/1                       
fleet-local   mcc-rancher-logging-crd      1/1                       
fleet-local   mcc-rancher-monitoring-crd   1/1                  
```

### To Reproduce

Install harvester v1.7.0-dev-20250928, wait a bit, run `kubectl get bundle -A` and observe the above errors.

### Expected Behavior

Harvester deploys successfully with no errors

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.7.0-dev-20250928
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Virtualized
- Rancher version:

### Additional context

This is caused by a recent change in Longhorn's value validation https://github.com/longhorn/longhorn/commit/d41740d18a51eb3084a88276a10867fc53e1644d

### Workaround and Mitigation

[...]

PR #9203, opened 2025-09-24
Upgrade processing of monitoring and logging bumping

w13915984028:
#### Problem:

Monitoring and logging are bumped, related upgrade path should be updated too.

#### Solution:

Add processing on upgrade path

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8930
https://github.com/harvester/harvester/issues/9201
https://github.com/harvester/harvester/issues/4958

#### Test plan:

Upgrade without enabling rancher-monitoring and rancher-logging addon
1. Install v160/v161 cluster
2. Upgrade to v170
3. Should have no issue to enable rancher-monitoring and rancher-logging addon

Upgrade without enabled rancher-monitoring and rancher-logging addon
1. Install v160/v161 cluster
2. Should have no issue to enable rancher-monitoring and rancher-logging addon
3. Upgrade to v170

#### Additional documentation or context

test log when run related script directly:

```
try to patch addon rancher-monitoring in cattle-monitoring-system to 107.1.0+up69.8.2-rancher.15, with patchs upon rancherMonitoring and kube-state-metrics
check rancherMonitoring.enabled
Error: no matches found
null
check kube-state-metrics.metricLabelsAllowlist
Error: no matches found
null
current valuesContent of the addon rancher-monitoring:
prometheus:
...
  service:
    port: 9093
new patch of addon rancher-monitoring:
spec:
  version: 107.1.0+up69.8.2-rancher.15
  valuesContent: |
    prometheus:
...
      service:
        port: 9093
    rancherMonitoring:  // patched
      enabled: false // patched
    kube-state-metrics: // patched
      metricLabelsAllowlist: // patched
        - nodes=[*] // patched
```

local test:  with monitoring and logging enabled on v160, upgrade to v170 master-head

image

existing addon was patched
image

test log when upgrade:

```
...
try to patch addon rancher-monitoring in cattle-monitoring-system to 107.1.0+up69.8.2-rancher.15, with patches upon rancherMonitoring and kube-state-metrics
...
...
    rancherMonitoring:
      enabled: false
    kube-state-metrics:
      metricLabelsAllowlist:
        - nodes=[*]
addon.harvesterhci.io/rancher-monitoring patched
Wed Sep 24 20:17:32 UTC 2025
wait addon status to be AddonDeploySuccessful
...
current status is AddonUpdating, continue wait: 10
Wed Sep 24 20:18:28 UTC 2025
addon status is AddonDeploySuccessful
upgrade addon rancher-logging
[...]

PR #9236, opened 2025-10-01
fix: stringify longhorn numeric defaultValues on upgrade

tserong:
#### Problem:
Since https://github.com/longhorn/longhorn/commit/d41740d1, longhorn's defaultValues must be strings.

#### Solution:
Force concurrentAutomaticEngineUpgradePerNodeLimit to string on upgrade if set

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9223

#### Test plan:
- Install harvester v1.6.0
- Run `kubectl -n fleet-local edit managedchart/harvester` and edit `.spec.values.longhorn.defaultSettings` to set `concurrentAutomaticEngineUpgradePerNodeLimit = 3` (i.e. make sure it's an unquoted number)
- Upgrade to a build that includes this fix
- During the upgrade, the value above should be patched to a quoted string, e.g.:
```
> kubectl -n fleet-local get managedchart/harvester -o yaml | grep -A 5 defaultSettings:
      defaultSettings:
        concurrentAutomaticEngineUpgradePerNodeLimit: "3"
        defaultDataPath: /var/lib/harvester/defaultdisk
        detachManuallyAttachedVolumesWhenCordoned: true
        nodeDrainPolicy: allow-if-replica-is-stopped
        taintToleration: kubevirt.io/drain:NoSchedule
```
- The upgrade should succeed as usual.

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -785,6 +785,14 @@ patch_longhorn_settings() {
     echo "longhorn detachManuallyAttachedVolumesWhenCordoned has been set in managedchart, do not patch again"
   fi
 
+  EXIT_CODE=0
+  local value=$(yq -e -r '.spec.values.longhorn.defaultSettings.concurrentAutomaticEngineUpgradePerNodeLimit' $target) || EXIT_CODE=$?
+  if [ $EXIT_CODE = 0 ]; then
+    # this works even if it's already a string, because the `yq` above gives us the unquoted value
+    echo "patch longhorn concurrentAutomaticEngineUpgradePerNodeLimit to string"
+    yq ".spec.values.longhorn.defaultSettings.concurrentAutomaticEngineUpgradePerNodeLimit = \"$value\"" -i $target
+  fi
+
   echo "longhorn related config"
   yq -e '.spec.values.longhorn' $target || echo "fail to get info .spec.values.longhorn"
 }


package/upgrade/upgrade_manifests.sh
```diff
@@ -785,6 +785,14 @@ patch_longhorn_settings() {
     echo "longhorn detachManuallyAttachedVolumesWhenCordoned has been set in managedchart, do not patch again"
   fi
 
+  EXIT_CODE=0
+  local value=$(yq -e -r '.spec.values.longhorn.defaultSettings.concurrentAutomaticEngineUpgradePerNodeLimit' $target) || EXIT_CODE=$?
```
w13915984028, line 789: for `local var`, any error is escaped and the var is set to nil; 
hence we can simplify via
```
if [[ ! -z var ]]; then
fi
```

but for none-local var, the `|| EXIT_CODE...` is used to escape error

w13915984028 (approved):
LGTM, thanks.

package/upgrade/upgrade_manifests.sh
tserong: Thanks for the hint. I just did a bit of testing, and it looks like `yq` will send "null" to stdout even in the error case(!), so if I did the above then `value` would end up being the string `"null"`, so the `-z` check wouldn't work properly in that case :-/

I think I'd better leave it as-is...

w13915984028: sorry, I forgot it, yes, yq outputs `null` when not finding the key

a similar processing on https://github.com/harvester/harvester/pull/9203/files#diff-cca301b3921d99450abbcd828d89bf81b8c0b7112d1b941efe0bf3d62e234dcfR612

Issue #9162, opened 2025-09-17
[FEATURE] Support Longhorn V2 volume expansion


innobead:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

Currently, Harvester supports volume expansion for Longhorn V1 and third-party storage if they support the feature. Besides VM workloads, workloads in guest clusters are also supported.

Longhorn v2 volume expansion is released in Longhorn 1.10. It needs to be made available in Harvester along with the existing volume expansion and VM images based on v2. 

### Additional context

Added to 1.7.0 first. @Vicente-Cheng, we may need to review the backlog again to review the capacity. 

cc @harvester/pm  @harvester/storage

albinsun:
Test PASS, close as done.


## Environment
* `harvester-v1.7.0-dev-20251026`
  * Profile: 3 nodes AMD64 Bare-metal HPE ProLiant DL360 
  * ui-source: `Auto`


## Test Scenario
### Prerequisite
1. <details>
   <summary> Enable LHv2 Data Engine
   </summary>

   Ref. https://docs.harvesterhci.io/v1.7/advanced/longhorn-v2/

   * Harvester
     <img width="1566" height="769" alt="Image" src="https://github.com/user-attachments/assets/ec21ca71-8d34-4b67-97b0-0aa3c425cd7f" />
   * Longhorn
     <img width="1566" height="769" alt="Image" src="https://github.com/user-attachments/assets/1413af7b-946d-48f9-8d47-775e05d92a42" />
   </details>

1. <details>
   <summary> Add extra disks with <b>Longhorn V2 (CSI)</b> provisioner
   </summary>

   Ref. https://docs.harvesterhci.io/v1.7/host/#multi-disk-management

   * node-0
     <img width="1566" height="775" alt="Image" src="https://github.com/user-attachments/assets/e564a91d-1523-445b-b4cf-99205f8c40e7" />

   * node-1
     <img width="1566" height="775" alt="Image" src="https://github.com/user-attachments/assets/c19ea4fe-55c6-4b62-8ea2-bb185e79c99e" />

   * node-2
     <img width="1566" height="769" alt="Image" src="https://github.com/user-attachments/assets/d2143714-0c99-4f89-be3d-e8179c28ec95" />
   </details>

1. <details>
   <summary> add a new StorageClass with <b>Longhorn V2 (CSI)</b> provisioner
   </summary>

[...]

PR #9243, opened 2025-10-01
feat: support LH v2 online expansion

WebberHuang1118:
#### Problem:
Support online expansion for Longhorn v2 volumes

#### Solution:
Adjust the Harvester webhook to allow online expansion for Longhorn v2 volumes.

#### Related Issue(s):
#9162 

#### Test plan:
* Start a VM with an LHv2 volume
* Expand the size of the related PVCs
  * By editing the manifest, if operating via GUI, the Harvester GUI blocks expansion for the LHv2 volumes
* The volume should be expanded


brandboat (approved):
LGTM, thanks!

Vicente-Cheng (approved):
Thanks!

Issue #9059, opened 2025-09-05
[BUG] PCI Passthrough VMs Will Not Boot with Default Harvester Settings in 1.6.0


ParadoxGuitarist:
### Describe the Bug

There's [a known issue upstream with libvirt](https://github.com/kubevirt/kubevirt/issues/12433), where if Memory Overcommit and PCI Passthrough are enabled, VMs will not start. The upstream bug has not yet been patched, and this bug report is simply for community awareness and for the Harvester Devs to be able to work with upstream, or advise the community when the patched version will get into Harvester. 

** EDIT ** Working with the devs, the reproducibility did change a bit. Initially this was only able to be reproduced in the first install/boot, which produced the same error messages as the upstream bug. (`exec: libvirt:  error : cannot limit locked memory of process $foo to $bar : Operation not permitted'`) Later we discovered that setting the cpus to 64 reproduced that same `locked memory` error when the GPU cuda device and related audio device were added via PCI passthrough. Removing the audio device changed the error output to: `LibvirtError(Code=1, Domain=10, Message='internal error: QEMU unexpectedly closed the monitor: qemu-system-x86_64: -accel kvm: warning: Number of hotpluggable cpus requested (256) exceeds the recommended cpus supported by KVM`. Which was then related to the hotplug feature gate. Lowering the core count to less than 64 but over 32, the VM would *occasionally* throw similar/additional errors, but not every time. Lowering it to below 32 and it would boot consistently. Due to the nature of the changing error text, it's feasible that these might be separate bugs, so I'm leaving a lot of the language still in here to help with searches. 

### To Reproduce

* Install harvester with the default Memory Overcommit settings (150%)
* Enable PCI Passthrough addon. 
* Create a VM with an enabled passthrough GPU Device.
* VM fails to start. 

### Expected Behavior

VMs start with the Overcommit and PCI passthrough enabled.  

### Support Bundle for Troubleshooting

Available upon request, but seeing as this is already verified as a bug upstream already, I suspect it's not currently needed. 

### Environment

- Harvester version: 1.6.0

### Additional context

https://github.com/harvester/harvester/issues/8972 - 1.5.1 this bug was not observed. 

### Workaround and Mitigation

[...]

PR #9246, opened 2025-10-02
fix: patch maxSockets if enableCPUAndMemoryHotplug is false

FrankYang0529:
#### Solution:
If users don't enable cpu and memory hotplug, patch `maxSockets` as `1`, so we don't get default `maxHotplugRatio` from the KubeVirt CR.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9059

#### Test plan:

1. Create a VM without enabling cpu and memory hotplug. The `maxSockets` in the VM spec must be 1.
2. Create a VM with enabling cpu and memory hotplug. The `maxSockets` in the VM spec should be same as user definition.


ibrokethecloud (approved):
lgtm. thanks for the fix.

Vicente-Cheng (approved):

Vicente-Cheng:
@mergifyio backport v1.6

Issue #9247, opened 2025-10-02
[BUG] Harvester CDI backend crashes with OOM when HTTP server doesn't support range requests


brandboat:
### Describe the Bug

The CDI backend in Harvester crashes with Out-of-Memory (OOM) errors when the HTTP server hosting VM images doesn't properly support HTTP Range requests. This happens during VMImage creation when `fetchImageVirtualSize()` attempts to read the first 128 bytes to detect qcow2 format, but ends up reading the entire large image file into memory.

In `pkg/image/cdi/common.go`, the `fetchImageVirtualSize()` function:

1. Sends a `Range: bytes=0-128` header to request partial content.
2. Calls `io.ReadAll(rsp.Body)` assuming the server honors the Range request
3. If the server doesn't support Range requests (like Python's `SimpleHTTPServer`), it returns the entire file
4. `io.ReadAll()` attempts to read the entire file (potentially GBs) into memory, causing OOM crash

### To Reproduce

1. Host an ISO file (e.g. harvester v1.6.0 ISO) using `python3 -m http.server 8000` (doesn't support Range requests). 
2. Create a VMImage with CDI backend:
   ```yaml
   apiVersion: harvesterhci.io/v1beta1
   kind: VirtualMachineImage
   metadata:
     name: test-image
     namespace: harvester-system
   spec:
     backend: "cdi"
     sourceType: download
     url: "http://<HTTP_SERVER_IP>:8000/large-image.iso"
     targetStorageClassName: "longhorn-static"
   ```
3. Harvester crashes with OOM during image processing

### Expected Behavior

The CDI backend should gracefully handle HTTP servers that don't support Range requests without crashing.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.6.0
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Workaround and Mitigation

Host the iso file with formal HTTP server, normally a formal HTTP server should handle Range requests properly (e.g. nginx), but some development HTTP servers don't implement this feature.

asc5543:
## Test result
Test PASS, harvester won't crash even download the large image.

## Environment
* Harvester ISO version: `v1.7.0-dev-20251012`
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): `1 node via ipxe-example`


## Test Scenario
### Preparation
[...]

PR #9248, opened 2025-10-02
Fix harvester OOM when HTTP server doesn't support range requests

brandboat:
#### Related Issue(s):
https://github.com/harvester/harvester/issues/9247

#### Test plan:

Test with python http server
- Host ISO files (e.g. harvester v1.6.0 ISO) using python3 -m http.server 8000 (doesn't support Range requests).
- Test vm image with cdi backend and large iso file, expect Harvester works properly without crashing.
   ```yaml
  kind: VirtualMachineImage
  metadata:
    name: test-upgrade-repo
    namespace: harvester-system
  spec:
    displayName: "test-upgrade-repo"
    backend: "cdi"
    sourceType: download
    url: "http://192.168.100.1:8000/harvester-master/harvester-master-amd64.iso"
    targetStorageClassName: "longhorn-static"
   ```

Test with formal http server (e.g. nginx, or other formal image provider like centos in the example) and qcow2 image

- Test vm image with qcow2 image and cdi backend, the vm image SIZE should be `1716977664`, VIRTUALSIZE should be `10737418240`.
   ```yaml
  kind: VirtualMachineImage
  metadata:
    name: test-qcow2
    namespace: harvester-system
  spec:
    displayName: "test-qcow2"
    backend: "cdi"
    sourceType: download
    url: "https://cloud.centos.org/centos/10-stream/x86_64/images/CentOS-Stream-GenericCloud-10-latest.x86_64.qcow2"
    targetStorageClassName: "longhorn-static"
   ```
N/A

### DIFF
--- a/pkg/image/cdi/common.go
+++ b/pkg/image/cdi/common.go
@@ -147,7 +147,7 @@ func fetchImageVirtualSize(url string) (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	req.Header.Set("Range", "bytes=0-128")
+	req.Header.Set("Range", "bytes=0-127")
 
 	rsp, err := http.DefaultClient.Do(req)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


WebberHuang1118 (approved):
LGTM, thanks.

pkg/image/cdi/common.go
```diff
@@ -147,15 +147,18 @@ func fetchImageVirtualSize(url string) (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	req.Header.Set("Range", "bytes=0-128")
+	req.Header.Set("Range", "bytes=0-127")
```
Vicente-Cheng, line 150: nice catch :p

Vicente-Cheng (approved):
thanks for fixing that!

Vicente-Cheng:
@mergifyio backport v1.6

Issue #9253, opened 2025-10-02
[CI] Adjust scan and notify testing message format


Yu-Jack:
### Task request

The format from #8106 is hard to read, so we'll use Slack API to organize the format to be like:

```
summary message
    thread message for details
```

asc5543:
Current format looks great.
Thanks Jack!

Yu-Jack:
@asc5543 Thanks! Actually, I'm testing the one lol. I'll close it when I'm done.

Yu-Jack:
This one will be tested in 2 weeks, most of codes and tests are done, so I'll keep it closed until we find some bugs.

PR #9254, opened 2025-10-02
ci: fix scan-and-notify-testing message format [skip ci]

Yu-Jack:
Secrets are set. Related Issue: https://github.com/harvester/harvester/issues/9253.

- [x] Separate summary and details
- [x] Prevent showing `read more` hint with two blocks as a group. It was five blocks as a group before.

image

### DIFF
--- a/.github/workflows/scan-and-notify-testing-items.py
+++ b/.github/workflows/scan-and-notify-testing-items.py
@@ -1,12 +1,41 @@
 import requests
 import os
 import sys
-import time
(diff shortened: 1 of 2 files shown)


Yu-Jack:
Don't merge it. I need to request more permissions.

Yu-Jack:
Permission issue is solved.

Vicente-Cheng (approved):
thanks for the enhancement!

PR #9275, opened 2025-10-08
feat: add new machine validator under webhook

mingshuoqiu:
#### Problem:

Harvester should block addition of new nodes while an upgrade is in progress.

This can have unwanted side effects if a user adds a node during the upgrade process.

#### Solution:

add new machine validator to prevent node registration during upgrade.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/7601

#### Test plan:

1. Install an older version Harvester to create a single node cluster.
2. Trigger upgrade process on the created management node
3. Install Harvester on another machine to create a worker/witness node to join the cluster created in step 1.
4. Check if the node creation can succeed or not.

### DIFF
--- a/pkg/webhook/resources/machine/validator.go
+++ b/pkg/webhook/resources/machine/validator.go
@@ -1,0 +1,107 @@
+package machine
+
+import (
+	"fmt"
+
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/webhook/types"
+)
+
+func NewValidator(client client.Client, upgradeCache ctlharvesterv1.UpgradeCache) types.Validator {
+	return &machineValidator{
+		client:       client,
+		upgradeCache: upgradeCache,
+	}
+}
+
+type machineValidator struct {
+	types.DefaultValidator
+	client       client.Client
+	upgradeCache ctlharvesterv1.UpgradeCache
+}
+
+func (v *machineValidator) Resource() types.Resource {
+	return types.Resource{
+		Names:    []string{"machines"},
+		Scope:    admissionregv1.ClusterScope,
+		APIGroup: clusterv1.GroupVersion.Group,
+		APIVersion: clusterv1.GroupVersion.Version,
+		ObjectType: &clusterv1.Machine{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+		},
+	}
+}
+
+func (v *machineValidator) Create(_ *types.Request, newObj runtime.Object) error {
+	machine, ok := newObj.(*clusterv1.Machine)
+	if !ok {
+		return fmt.Errorf("expected Machine, but got %T", newObj)
+	}
+
+	upgrading, err := v.isClusterUpgrading()
+	if err != nil {
+		logrus.WithFields(logrus.Fields{
+			"machine":   machine.Name,
+			"namespace": machine.Namespace,
+			"error":     err,
+		}).Error("failed to check cluster upgrade status")
+		return nil
+	}
+
+	if upgrading {
+		fields := logrus.Fields{
+			"machine":   machine.Name,
+			"namespace": machine.Namespace,
+		}
+
+		if machine.Spec.ProviderID != nil && *machine.Spec.ProviderID != "" {
+			fields["providerID"] = *machine.Spec.ProviderID
+		}
+
+		if len(machine.Status.Addresses) > 0 {
+			addrParts := []string{}
+			for _, addr := range machine.Status.Addresses {
+				// MachineAddressType gives type (HostName, ExternalIP, InternalIP…)
+				addrParts = append(addrParts, fmt.Sprintf("%s=%s", addr.Type, addr.Address))
+			}
+			fields["addresses"] = addrParts
+		}
+
+		logrus.WithFields(fields).Warn("blocking creation of new Harvester Machine: upgrade in progress")
+
+		return fmt.Errorf("cluster upgrade in progress: cannot create machine %s", machine.Name)
+	}
+
+	return nil
+}
+
+func (v *machineValidator) isClusterUpgrading() (bool, error) {
+	selector := labels.Set{
+		util.LabelUpgradeLatest: "true",
+	}.AsSelector()
+
+	upgrades, err := v.upgradeCache.List(util.HarvesterSystemNamespaceName, selector)
+	if err != nil {
+		return false, err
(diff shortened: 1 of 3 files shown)


mingshuoqiu:
@starbops Could you help review? Thanks

starbops (commented):
The implementation seems incomplete. I saw there is a new machine validator introduced but not registered. Also, the node validator now checks on "create" events, but there is no "Create" method for node validator.

mingshuoqiu:
> The implementation seems incomplete. I saw there is a new machine validator introduced but not registered. Also, the node validator now checks on "create" events, but there is no "Create" method for node validator.
 
Fixed the validator registration. Please help review. Thanks

pkg/webhook/resources/machine/validator.go
```diff
@@ -0,0 +1,74 @@
...
+	if upgrading {
+		return fmt.Errorf("cluster upgrade is in progress: node registration is temporarily blocked")
+	}
+
+	return nil
+}
+
+func (v *machineValidator) isClusterUpgrading() (bool, error) {
+
+	upgrades, err := v.upgradeCache.List(util.HarvesterSystemNamespaceName, labels.NewSelector())
```
starbops: We can only list the "latest" upgrade object(s) by specifying the label selector.

pkg/webhook/resources/node/validator.go
starbops, line 1: If we're not going to implement any check upon node creation, we can drop the changes in this file.

pkg/webhook/resources/machine/validator.go
mingshuoqiu: Fixed. Please help check if it's correct or not.

pkg/webhook/resources/node/validator.go
mingshuoqiu: Removed.

w13915984028:
BTW, the `machine` is the core CRD of the embedded Rancher, need to investigate and observe logs/object status/... on the embedded Rancher, to avoid blocking the upgrade of Rancher, which in turn blocks the whole upgrade.

mingshuoqiu:
> BTW, the `machine` is the core CRD of the embedded Rancher, need to investigate and observe logs/object status/... on the embedded Rancher, to avoid blocking the upgrade of Rancher, which in turn blocks the whole upgrade.

I’m trying to refine the validation logic, but based on the current state of the Machine objects in Harvester, I’m not able to find any reliable identifiers that distinguish Harvester-managed machines from other embedded Rancher/CAPI machines.

What criteria would you recommend we use to determine:
  - Which Machine objects should be protected during Harvester upgrade, and
  - Which Machine objects must always be allowed, to avoid interfering with Rancher/CAPI’s own controllers?
Thanks

w13915984028:
@mingshuoqiu 

On Harvester, all the `machine` objects are 1:1 mapping to Harvester cluster nodes, there is no other machines. The concern is, the `machine` life cycle is managed by the embedded `Rancher (Manager)`,  not Harvester, if we block the `create` of it, could this bring some side-effect to Rancher &/ other objects?  We need to investigate and ensure.

starbops (dismissed):
LGTM, thank you.

According to the proposed implementation, the machine creation will only be blocked during the upgrade period, so it should be fine. @w13915984028 Do you have any suspicious cases that you think are worth exploring?

w13915984028:
As it is difficult to evalute the influence statically, please prepare a test report, thanks.

(1) Setup a >= 2 nodes cluster with v160/v161 ( in a multi-node cluster upgrade, the node upgrade is delegated to Rancher)
(2) Upgrade to a v170 version with this PR
(3) Join a new node in-between the upgrade
(4) Check below objects, if any is abnormal;  and the logs from Rancher, CAPI pod
(5) Check if upgrade is successful
(6) After upgrade is successful, check if the pending node can join, is it with old version or new version?

BTW: not sure Rancher/RKE2 checks the join node must have the same version with the running cluster; in the upgrade scenario, it is a big tricky:
e.g. when a v160 cluster is upgrading, the join node is installed with v160 ISO, after v160 is upgraded to v170, then, can the pending v160 node join later?   sure, this is out of the scope of this PR.

image

pkg/webhook/resources/machine/validator.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if upgrading {
+		return fmt.Errorf("cluster upgrade is in progress: node registration is temporarily blocked")
+	}
+
+	return nil
+}
+
+const upgradeLatestLabel = "harvesterhci.io/latestUpgrade"
```
w13915984028: BTW: please define a public label on constants, thanks.

https://github.com/harvester/harvester/blob/5b3a1bbea8ba07928f5725b34f41ccc5d8a6a9e7/pkg/util/constants.go#L147

and redefine below to the public one

https://github.com/harvester/harvester/blob/5b3a1bbea8ba07928f5725b34f41ccc5d8a6a9e7/pkg/controller/master/upgrade/upgrade_controller.go#L55

pkg/webhook/resources/machine/validator.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (v *machineValidator) Create(_ *types.Request, newObj runtime.Object) error {
+
+	upgrading, err := v.isClusterUpgrading()
+	if err != nil {
+		return fmt.Errorf("failed to check cluster upgrade status: %w", err)
+	}
+
+	if upgrading {
+		return fmt.Errorf("cluster upgrade is in progress: node registration is temporarily blocked")
```
w13915984028: when blocking, please also log a warning message with the new machine's info like name, version, ip (if feasible), for better troubleshooting, thanks.

mingshuoqiu: I've added the warning message per the suggestion. Please help check if it's sufficient or not. Thanks

pkg/webhook/resources/machine/validator.go
mingshuoqiu: Added.

mingshuoqiu:
Tested the new image of 1.7rc2+this PR

Test date: 2025-11-25
Tester: Chris Chiu
Build: v1.7.0-rc2 + PR #7601 (validator patch)
Cluster under test: 3-node Harvester on bare-metal (AMD64)
Upgrade origin: v1.6.1 → v1.7.0-rc2 

    **Environment preparation**

Step 1.1  Deploy 3-node Harvester v1.6.1 (nodes: hv-01, hv-02, hv-03)
Step 1.2  Verify all pods Ready, RKE2 version identical on every node
Step 1.3  Create dummy VM “vm-fio” to obtain baseline I/O metrics

Screenshot from 2025-11-26 06-10-52


    **Trigger upgrade to v1.7.0-rc2**

Step 2.1  Apply Version CR harvesterhci.io/v1beta1 Version object v1.7.0-rc2
Step 2.2  Rancher upgrades first control-plane (hv-01) → cordon → drain → replace RKE2 image
Step 2.3  Upgrade status: In-Progress
Screenshot from 2025-11-26 06-12-35


    **Attempt to join a new node while upgrade is running**

Step 3.1  PXE-boot new bare-metal node hv-04 (still on v1.6.1 image)
Step 3.2  Harvester validating webhook invoked → blocks creation
Step 3.3  Machine hv-04 remains in Pending state; no RKE2 provisioning started

Screenshot from 2025-11-26 06-16-00


    **Upgrade finalisation**

Step 4.1  hv-02 and hv-03 upgraded sequentially
Step 4.2  RKE2 version now with new version on all 3 nodes

Screenshot from 2025-11-26 06-17-34

w13915984028 (approved):
LGTM, thanks for the detailed test, it is safe to merge now.

BTW, per offline discussion, the old versioned node (v16*) finally joined into the upgraded cluster (v17*),  please file a new issue to track.

starbops (approved):

Issue #9269, opened 2025-10-07
[ENHANCEMENT] checking the Longhorn CRD version before upgrade


Vicente-Cheng:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

To upgrade Longhorn v1.10.0, we need to ensure we migrate from v1beta1 to v1beta2 and remove all the v1beta1 resource CRD.

We need the corresponding checks for the above stuff to ensure the upgrade would not be stuck.

### Describe the solution you'd like

- Add the checking mechanism through webhook
- Add HowTo on the document like (https://longhorn.io/docs/1.10.0/important-notes/#upgrade)

innobead:
cc @harvester/storage

asc5543:
## Test result
Test PASS in `v1.6.0` -> `v1.7.0-rc3` upgrade path

## Environment
* Harvester ISO version: `v1.6.0` -> `v1.7.0-rc3`
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): `1 node via ipxe-example`


## Test Scenario
### Preparation
Prepare the v1.7.0-rc3 for upgrade

1. Prepare the v1.7.0-rc3 ISO file
1. <details> <summary> Prepare the version.yaml file </summary>

    ```
    apiVersion: harvesterhci.io/v1beta1
    kind: Version
    metadata:
      name: v1.7.0-rc3
      namespace: harvester-system
    spec:
      isoChecksum: '1cbd5ef881663656c9bf0d79f46007a386d0f2f6d180791864e8d7e50e3216229599ba2636cbc51eebe50df3829dfe8923d63a519ecb5f807ad710374afc135a'
      isoURL: http://<lab-ip>/harvester-v1.7.0-rc3-amd64.iso
      releaseDate: '20251121'
    ```

   </details>

1. <details> <summary> Create a version object. </summary>

    ```
    # kubectl create -f version.yaml
    version.harvesterhci.io/v1.7.0-rc3 created
    ```

   </details>

### Test Steps
1. <details>
   <summary> ✔ Create a VM with file, and check the sha512sum
   </summary>
  
    <img width="1896" height="1057" alt="Image" src="https://github.com/user-attachments/assets/731ad970-9033-40ed-a128-af9d591bd031" />

    ```
    ubuntu@testvm:~$ cat testfile 
    testfile
    ubuntu@testvm:~$ sha512sum testfile 
[...]

PR #9284, opened 2025-10-09
upgrade: add Longhorn CRD migrations during upgrade

Vicente-Cheng:
- we need to take care the dropped v1beta1 resource.
    - This is the corner case only if user starts Harvester from v1.0, Longhorn v1.2.4 (v1beta1)

#### Problem:
Need to take care of the Longhorn v1beta1 resource

#### Solution:
Migrate/Cleanup the v1beta1 resource if needed

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9269

#### Test plan:
Upgrade from v1.6.x and ensure everything is OK.
We need to ensure the common path is not affected

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -1245,6 +1245,43 @@ apply_extra_nonversion_manifests()
   shopt -u nullglob
 }
 
+migrate_longhorn_v1beta1_crds() {
+
+  # reference: https://longhorn.io/docs/1.10.0/important-notes/#upgrade
+  # Harvester v1.6.x -> v1.7.x, Longhorn is upgraded from v1.9.x to v1.10.x
+  echo "Checking if migrating Longhorn v1beta1 CRDs to v1beta2 is required, it only occurs from v1.6.x to v1.7.x upgrade"
+  if [[ ! "$UPGRADE_PREVIOUS_VERSION" =~ ^v1\.6\.[0-9]$ ]]; then
+    echo "Skip migrate longhorn v1beta1 CRDs with Harvester version $UPGRADE_PREVIOUS_VERSION"
+    return
+  fi
+
+  # Temporarily disable the Longhorn webhook for CR validation
+  echo "Temporarily disabling the Longhorn webhook validator..."
+  kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+      --type=merge \
+      -p "$(kubectl get validatingwebhookconfiguration longhorn-webhook-validator -o json | jq '.webhooks[0].rules |= map(if .apiGroups == ["longhorn.io"] and .resources == ["settings"] then .operations |= map(select(. != "UPDATE")) else . end)')"
+
+  # Find and migrate CRDs with v1beta1 stored versions
+  echo "Starting migration of CRDs that store v1beta1 resources to v1beta2"
+  migration_time="$(date +%Y-%m-%dT%H:%M:%S)"
+  crds=($(kubectl get crd -l app.kubernetes.io/name=longhorn -o json | jq -r '.items[] | select(.status.storedVersions | index("v1beta1")) | .metadata.name'))
+  for crd in "${crds[@]}"; do
+    echo "Migrating ${crd} ..."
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_manifests.sh
```diff
@@ -1244,6 +1244,43 @@ apply_extra_nonversion_manifests()
   shopt -u nullglob
 }
 
+migrate_longhorn_v1beta1_crds() {
+
+  # reference: https://longhorn.io/docs/1.10.0/important-notes/#upgrade
+  # Harvester v1.6.x -> v1.7.x, Longhorn is upgraded from v1.9.x to v1.10.x
+  echo "Checking version, it only happened on v1.6.1 -> v1.7.x upgrade"
+  if [[ $UPGRADE_PREVIOUS_VERSION != "v1.6.0" && $UPGRADE_PREVIOUS_VERSION != "v1.6.1" ]]; then
```
innobead: Should do this for any 1.6.x upgrade to 1.7.x, since v1beta1 is removed from Longhorn 1.10.

innobead (changes requested):

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: Updated for the general condition

ibrokethecloud (dismissed):

package/upgrade/upgrade_manifests.sh
```diff
@@ -1244,6 +1244,43 @@ apply_extra_nonversion_manifests()
   shopt -u nullglob
 }
 
+migrate_longhorn_v1beta1_crds() {
+
+  # reference: https://longhorn.io/docs/1.10.0/important-notes/#upgrade
+  # Harvester v1.6.x -> v1.7.x, Longhorn is upgraded from v1.9.x to v1.10.x
+  echo "Checking version, it only happened on v1.6.x -> v1.7.x upgrade"
```
innobead:
```suggestion
  echo "Checking if migrating Longhorn v1beta1 CRDs to v1beta2 is required, it only occurs from v1.6.x to v1.7.x upgrade"
```

package/upgrade/upgrade_manifests.sh
```diff
@@ -1244,6 +1244,43 @@ apply_extra_nonversion_manifests()
...
+  fi
+
+  # Temporarily disable the Longhorn webhook for CR validation
+  echo "Temporarily disabling the Longhorn webhook validator..."
+  kubectl patch validatingwebhookconfiguration longhorn-webhook-validator \
+      --type=merge \
+      -p "$(kubectl get validatingwebhookconfiguration longhorn-webhook-validator -o json | jq '.webhooks[0].rules |= map(if .apiGroups == ["longhorn.io"] and .resources == ["settings"] then .operations |= map(select(. != "UPDATE")) else . end)')"
+
+  # Find and migrate CRDs with v1beta1 stored versions
+  echo "Starting migration of CRDs that store v1beta1 resources..."
```
innobead:
```suggestion
  echo "Starting migration of CRDs that store v1beta1 resources to v1beta2"
```

innobead (dismissed):
LGTM, just a minor suggestion for messages.

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: udpated

package/upgrade/upgrade_manifests.sh
Vicente-Cheng: updated

ibrokethecloud (approved):

Vicente-Cheng:
rebased to the latest.
@starbops, @ibrokethecloud, @innobead could you help review it again?

ibrokethecloud (approved):

starbops (approved):
LGTM, thank you.

Vicente-Cheng:
@mergifyio backport v1.7

w13915984028:
Sorry for later reply, as the PR will shortly disable/enable LH webhook two times no matter it really needs to patch CRD, and apply manifest pod might rerun. Could we check if it really needs to patch, then do those „write“ operations? Thx.

PR #9285, opened 2025-10-09
fix: steve server does not check nil extension server api issue

Yu-Jack:
#### Problem:

After bumping to v0.6.17 steve server in https://github.com/harvester/harvester/issues/8719, steve server has a bug that doesn't check nil extension api server. It's fixed in v0.6.20.

v0.6.17 steve serve https://github.com/rancher/steve/blob/c32995dba487dfde38008e2a2f830e54b9282e53/pkg/server/server.go#L258-L265

Fixed in v0.6.20 steve server https://github.com/rancher/steve/blob/6495e7879c3e768835d099e764a2fc4e1b38edaf/pkg/server/server.go#L263-L275

But we can't bump it because kubevirt is still tied to k8s v1.32. There will be dependency hell if we bump it.

#### Solution:
Pass a fake extension api server to bypass.

#### Related Issue(s):
we bumped steve server in https://github.com/harvester/harvester/issues/8719.

#### Test plan:

Build Harvester binary in local and run it with [this guideline.](https://github.com/harvester/harvester/wiki/Setting-Up-Harvester-Development-Environment#set-up-goland).

### DIFF
--- a/pkg/server/types.go
+++ b/pkg/server/types.go
@@ -194,6 +194,31 @@ func (s *HarvesterServer) Scaled() *config.Scaled {
 	return config.ScaledWithContext(s.Context)
 }
 
+// noopExtensionAPIServer is a no-op implementation of ExtensionAPIServer
+type noopExtensionAPIServer struct {
+	registered chan struct{}
+}
+
+func newNoopExtensionAPIServer() *noopExtensionAPIServer {
+	ch := make(chan struct{})
+	close(ch) // immediately close the channel to indicate "registered"
+	return &noopExtensionAPIServer{
+		registered: ch,
+	}
+}
+
+func (n *noopExtensionAPIServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	http.NotFoundHandler().ServeHTTP(rw, req)
+}
+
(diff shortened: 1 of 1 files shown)


pkg/server/types.go
```diff
@@ -234,10 +259,11 @@ func (s *HarvesterServer) generateSteveServer(options config.Options) error {
...
 	s.steve, err = steveserver.New(s.Context, s.RESTConfig, &steveserver.Options{
-		Controllers:     s.controllers,
-		AuthMiddleware:  steveauth.ExistingContext,
-		Router:          router.Routes,
-		AccessSetLookup: s.ASL,
+		Controllers:        s.controllers,
+		AuthMiddleware:     steveauth.ExistingContext,
+		Router:             router.Routes,
+		AccessSetLookup:    s.ASL,
+		ExtensionAPIServer: newNoopExtensionAPIServer(),
```
innobead, line 269: Since this is a workaround, in addition to the commit message, suggest adding a FIXME comment for future reference.

innobead (changes requested):

DamiaSan (approved):
LGTM, thanks.

innobead (approved):

Issue #7042, opened 2024-11-26
[FEATURE] Support for HotPlugNICs from Kubevirt into Harvester


rrajendran17:
**Is your feature request related to a problem? Please describe.**
Currently in Harvester, interfaces cannot be added/removed from running VMs dynamically.
The vm has to be restarted after a network interface is added/removed for the configuration to take effect.
This will disrupt the applications running or traffic loss happens on other network interfaces attached to the vm.

**Describe the solution you'd like**
Bring in support for HotPlugNics from Kubevirt into Harvester.
https://kubevirt.io/user-guide/network/hotplug_interfaces/

**Describe alternatives you've considered**
Migration based hotplug[¶](https://kubevirt.io/user-guide/network/hotplug_interfaces/#migration-based-hotplug)

In case cluster doesn't run Multus as [thick plugin](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/thick-plugin.md) and [Multus Dynamic Networks controller](https://github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller), it's possible to hotplug an interface by migrating the VM.
The actual attachment won't take place immediately, and the new interface will be available in the guest once the migration is completed.

**Additional context**
1.The feature HotplugNICs has to be enabled (from v1.1 until v1.3)
2. Manifests dynamic-network-controller.yaml and multus-daemonset-thick.yaml  from k8snetworkplumbingwg has to be modified and applied 
3.Current harvester does have only multus thin plugin, it has to add support for multus thick plugin.

Also, currently when storage networks are configured, the secondary networks are created on LH instance manager and backing image pods.Without the support for dynamic-network-controller and multus-thick plugin, the pods has to be recreated with new networks addition and deletion.This can be avoided if dynamic networks and multus thick plugin are supported.

thomasferrandiz:
We just added support for multus thick plugin and dynamic network controller.
It will be available in the July release.

https://github.com/rancher/rke2/issues/8533

wheatdog:
Make this issue a sub-issue of #8283.

a110605:
@wheatdog , please add UI change requirement in https://github.com/harvester/harvester/issues/9322. thanks

albinsun:
Multi-nodes cases passes but not single node and upgrade cases. 
Let's wait further version and will test upgrade cases again.

Opened issues:
1. harvester/harvester/issues/9619 (Plan in `v1.8.0`)
[...]

Issue #8283, opened 2025-05-14
[EPIC] Hotplug resources


forbesguthrie:
Priority 1.  Add Support for the items KubeVirt supports today (CPU, Mem, NIC) 
Priority 2.  Support removing a cdrom and container (try to live migrate a Windows VM after an install) 
Priority 3.  Change NIC network without needing a reboot

abonillabeeche:
🙌🙌🙌

innobead:
> Priority 1. Add Support for the items KubeVirt supports today (CPU, Mem, NIC) Priority 2. Support removing a cdrom and container (try to live migrate a Windows VM after an install) Priority 3. Change NIC network without needing a reboot

@forbesguthrie can you explain a bit about case 2? Or do we have a JIRA or discussion context somewhere? For case 3, I assume it's just part of the hotplug NIC.

wheatdog:
#7042 is related to NICs hotplugging. Link it here as a sub-issue.

Vicente-Cheng:
We will handle the NIC hotplug on v1.7.0.
Other items would be handled in a later version.

innobead:
Changed this to EPIC, as it includes several sub-features. The remaining one should be https://github.com/harvester/harvester/issues/9261. @wheatdog Please correct me if anything is missing here, but it's available upstream.

cc @harvester/tpm

wheatdog:
@innobead Yes, #9261 should be the only one remaining.

innobead:
> [@innobead](https://github.com/innobead) Yes, [#9261](https://github.com/harvester/harvester/issues/9261) should be the only one remaining.

Sounds good. Let's close this when #9261 is finished.

innobead:
Closing, as all is finished. 

cc @harvester/release @harvester/tpm

PR #9295, opened 2025-10-13
docs(hep): NIC Hotplug [skip ci]

wheatdog:
#### Problem:

Please refer to the HEP.

#### Solution:

Please refer to the HEP.

#### Related Issue(s):

#7042 

#### Test plan:

Please refer to the HEP.

Please refer to the HEP.

### DIFF
--- a/enhancements/20251009-nic-hotplug.md
+++ b/enhancements/20251009-nic-hotplug.md
@@ -95,9 +95,17 @@ Content-Type: application/json
 
 ### Implementation Overview
 
+#### Check whether the VM is hot-pluggable / hot-unpluggable
+
+If the VM is a guest cluster VM (with label `harvesterhci.io/creator: docker-machine-driver-harvester`), it would be not hot-pluggable / hot-unpluggable before the integration with Rancher Manager is done. This protection ensures the stability of the guest cluster.
+
+If the VM is not live-migratable, then it isn't hot-pluggable / hot-unpluggable.
+
+If the VM is currently in transitional state, then it isn't hot-pluggable / hot-unpluggable until it reaches the stable running state.
+
 #### Hot-pluggable VM Networks
 
-In this milestone, NetworkAttachmentDefinitions with `bridge` type in their cni configs are hot-pluggable and hot-unpluggable. Namely, `UntaggedNetwork`, `L2VlanNetwork`, and `L2VlanTrunkNetwork` are supported while `OverlayNetwork` isn't.
+In the current milestone, NetworkAttachmentDefinitions with `bridge` type in their cni configs are hot-pluggable and hot-unpluggable. Namely, `UntaggedNetwork`, `L2VlanNetwork`, and `L2VlanTrunkNetwork` are supported while `OverlayNetwork` isn't.
 
 #### Network Interface Hot-plug Actions
 
--- a/enhancements/20251009-nic-hotplug.md
+++ b/enhancements/20251009-nic-hotplug.md
@@ -1,0 +1,133 @@
+# NIC Hotplug
+
+## Summary
+
+This enhancement leverages KubeVirt's network interface hot-plug feature to enable users to add or remove specific types of Network Interface Controllers (NICs) from a running Virtual Machine (VM).
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/7042
+
+## Motivation
+
+### Goals
+
+- Enable users to hot-plug a `virtio` interface connected via `bridge` or `SR-IOV` binding to a running VM without downtime.
+- Enable users to hot-unplug an interface connected via `bridge` binding from a running VM without downtime.
+
+### Non-goals
+
+- Support for hot-plugging interfaces with models not supported by KubeVirt.
+- Support migration based hotplug; the LiveUpdate roll-out strategy is preferred.
+- Addressing existing Virtio limitations.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Hot-plug an interface using the `virtio` model
+
+A user needs to add a new network interface to a running VM to connect it to a new network. Previously, this required scheduling downtime to shut down the VM, add the interface, and restart it. With this enhancement, the user can add a `virtio` interface using `bridge` or `SR-IOV` binding to the running VM, making the new network available immediately without service interruption.
+
+#### Story 2: Hot-unplug an interface connected through `bridge` binding
+
+A user needs to reconfigure a VM's network topology by removing an obsolete network interface. Instead of shutting down the VM, the user can now hot-unplug the `bridge`-bound interface from the running VM, simplifying network management and avoiding downtime.
+
+### User Experience In Detail
+
+The following steps outline the user workflow for managing network interfaces on a running VM:
+
+1.  Navigate to the details page of a running VM.
+2.  On the **Networks** tab, interfaces that support hot-unplug will display a corresponding action button (e.g., "Delete" or "Unplug").
+3.  To hot-plug an interface, the user can click the kebab menu (three vertical dots) on either the VM details page or the VM list page and select the **Add Network Interface** option. A dialog will appear, prompting the user to specify an interface name and select a network from a dropdown list.
+
+### API changes
+
+Two new API actions will be introduced for the `VirtualMachine` resource.
+
+**Add NIC**
+```
+POST /v1/harvester/vm/<vm>?action=addNic
+Content-Type: application/json
+
+{
+  "interfaceName": "dynif1",
+  "networkName": "default/vmnet1",
+  "bindingMethod": "bridge"  // or "sriov"
+}
+```
+
+**Remove NIC**
+```
+POST /v1/harvester/vm/<vm>?action=removeNic
+Content-Type: application/json
+
+{
+  "interfaceName": "dynif1"
+}
+```
+
+## Design
+
+
+### Implementation Overview
+
+#### Satisfy KubeVirt Hot-plug Network Interface Requirements
+
+According to the KubeVirt documentation[^1], two requirements must be met:
+
+1.  Multus CNI must be deployed using a [thick plugin](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/thick-plugin.md) architecture.
+2.  The Multus Dynamic Networks Controller must be installed.
+
+As per the RKE2 documentation, these requirements can be met by applying the appropriate `HelmChartConfig` via the Harvester installer.
+
+```yaml
+# pkg/config/templates/rancherd-23-multus-config.yaml
+resources:
+- apiVersion: helm.cattle.io/v1
+  kind: HelmChartConfig
+  metadata:
+    name: rke2-multus
+    namespace: kube-system
+  spec:
+    valuesContent: |-
+      thickPlugin:
+        enabled: true
+      dynamicNetworksController:
+        enabled: true
+```
+
+#### Network Interface Hot-plug Actions
+
+The API server will provide two new actions, `addNic` and `removeNic`, for `VirtualMachine` resources.
+
+-   `addNic` will patch the `VirtualMachine` resource by adding new elements to `spec.template.spec.domain.devices.interfaces` and `spec.template.spec.networks`. The interface `model` will be hardcoded to `virtio`, and only `bridge` and `SR-IOV` binding methods will be permitted.
+
+-   `removeNic` will patch the `VirtualMachine` resource by setting the `state` of the specified interface in `spec.template.spec.domain.devices.interfaces` to `absent`. This action will only be permitted for interfaces using the `bridge` binding method.
+
+### Test plan
+
+#### Regression tests
+
+Since this enhancement migrates the Multus CNI from a thin to a thick plugin architecture, all existing network functionalities, including the storage network, must be thoroughly tested to prevent regressions.
+
+#### Functional tests
+
+1.  Verify that a NIC with the `bridge` binding method can be added to a running VM.
+2.  Verify that a NIC with the `SR-IOV` binding method can be added to a running VM.
+3.  Verify that attempts to add a NIC with any other binding method are blocked.
+4.  Verify that a NIC with the `bridge` binding method can be removed from a running VM.
+5.  Verify that attempts to remove a NIC with any other binding method are blocked.
+
+### Upgrade strategy
+
+The NIC hot-plug feature must be compatible with VMs created prior to the upgrade. Additionally, the migration of Multus CNI from the thin to the thick plugin architecture must be verified to ensure a seamless Harvester upgrade process.
+
+## Note
+
+The KubeVirt documentation[^1] specifies a maximum number of interfaces that can be hot-plugged. This limitation should be clearly communicated in the Harvester user documentation.
+
+A `RestartRequired` condition has been observed on the `VirtualMachine` resource after a NIC hot-plug operation is followed by a VM reboot. This appears to be incorrect behavior, and we will coordinate with the KubeVirt community to investigate and resolve the issue.
+
+
+[^1]: https://kubevirt.io/user-guide/network/hotplug_interfaces/
--- a/enhancements/20251009-nic-hotplug.md
+++ b/enhancements/20251009-nic-hotplug.md
@@ -54,7 +54,6 @@ Content-Type: application/json
 {
   "interfaceName": "dynif1",
   "networkName": "default/vmnet1",
-  "bindingMethod": "bridge",
   "macAddress": "da:59:6c:5e:8b:fb"  // empty string or missing field indicates the Mac address is auto-generated
 }
 
@@ -75,26 +74,9 @@ Content-Type: application/json
 - status code: 204
 ```
 
-**List Hotpluggable VM Networks** which returns bridge vlan based VM Networks that can be used to create hot plug NICs for the VM
+**Find hot-pluggable NICs** which returns interfaces that can be hot-unplugged for the VM
 ```
-POST /v1/harvester/vm/<vm>?action=listHotpluggableVmNetworks
-Content-Type: application/json
-
-# Response
-- status code: 200
-- body:
-{
-  "networks": [
-    "ns1/vmnet1",
-    "ns1/vmnet2"
-    "ns2/vmnet3"
-  ]
-}
-```
-
-**List hot-pluggable NICs** which returns interfaces that can be hot-unplugged for the VM
-```
-POST /v1/harvester/vm/<vm>?action=listHotunpluggableNics
+POST /v1/harvester/vm/<vm>?action=findHotunpluggableNics
 Content-Type: application/json
 
 # Response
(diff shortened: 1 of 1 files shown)


enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Goals
+
+- Enable users to hot-plug a `virtio` interface connected via `bridge` or `SR-IOV` binding to a running VM without downtime.
+- Enable users to hot-unplug an interface connected via `bridge` binding from a running VM without downtime.
+
+### Non-goals
+
+- Support for hot-plugging interfaces with models not supported by KubeVirt.
+- Support migration based hotplug; the LiveUpdate roll-out strategy is preferred.
```
wheatdog: According to [KubeVirt's document](https://kubevirt.io/user-guide/network/hotplug_interfaces/#migrate-the-vm_2),

> Due to limitation of Kubernetes device plugin API to allocate resources dynamically, the SR-IOV device plugin cannot allocate additional SR-IOV resources for Kubevirt to hotplug. Thus, SR-IOV interface hotplug is limited to migration based hotplug only, regardless of Multus "thick" version.

the migration based hotplug is the only way that we can support SR-IOV interfaces.

I'm trying to sort out our current usage of SR-IOV interfaces in Harvester before the decision on which method we would like to use.

wheatdog: Consulted with @ibrokethecloud and [our document](https://docs.harvesterhci.io/v1.6/advanced/addons/pcidevices#sriov-network-devices), for SR-IOV Network interfaces, instead of using the off-the-shelf [device plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin) and [CNI](https://github.com/k8snetworkplumbingwg/sriov-cni) as mentioned in [KubeVirt's document](https://kubevirt.io/user-guide/network/interfaces_and_networks/#sriov), it's managed by our own device plugin framework (https://github.com/harvester/pcidevices) and can be assigned to VM via the `hostDevices` field [like this](https://kubevirt.io/user-guide/compute/host-devices/#starting-a-virtual-machine).

With this implementation, we might need more study on how to support SR-IOV interfaces hotplug in Harvester. For now, I prefer we support `bridge` binding method in this milestone first.

wheatdog: BTW, according to the implementation of [ui-extension](https://github.com/harvester/harvester-ui-extension/blob/main/pkg/harvester/edit/kubevirt.io.virtualmachine/VirtualMachineNetwork/base.vue#L121-L137), it's not possible to choose SR-IOV binding method while configuring network of VMs.

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+### Related Issues
+
+https://github.com/harvester/harvester/issues/7042
+
+## Motivation
+
+### Goals
+
+- Enable users to hot-plug a `virtio` interface connected via `bridge` or `SR-IOV` binding to a running VM without downtime.
+- Enable users to hot-unplug an interface connected via `bridge` binding from a running VM without downtime.
```
rrajendran17: Since kubevirt supports model virtio and bridge type for hotplugging, both for hotplug and hotunplug scenario you can mention model virtio and type bridge  so that both statements will be in sync

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+**Remove NIC**
+```
+POST /v1/harvester/vm/<vm>?action=removeNic
+Content-Type: application/json
+
+{
+  "interfaceName": "dynif1"
+}
+```
+
```
rrajendran17, line 76: We do not need new APIs, once we have multus thick plugin and dynamic network controller resource installed in Harvester, we should be able to edit the VM config(interfaces under domain and networks)  to add or remove interfaces through kubectl (UI needs changes to just save the config and no restart required).
This will update the VM pod's annotations.Dynamic network controller pod will listen to these change in pod annotations and  will send the request to multus thick plugin to attach/remove new interface to/from the pod.

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+  metadata:
+    name: rke2-multus
+    namespace: kube-system
+  spec:
+    valuesContent: |-
+      thickPlugin:
+        enabled: true
+      dynamicNetworksController:
+        enabled: true
+```
```
rrajendran17: We need to also enable the featureGate for HotplugNICs in kubevirt CRD.
Add HotplugNICs under featureGates of kubectl edit kubevirt.kubevirt.io/kubevirt -n harvester-system.

https://github.com/harvester/harvester/blob/14ee9af0c8034d7a133e82254dc183803a63395f/deploy/charts/harvester/values.yaml#L71

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Regression tests
+
+Since this enhancement migrates the Multus CNI from a thin to a thick plugin architecture, all existing network functionalities, including the storage network, must be thoroughly tested to prevent regressions.
+
+#### Functional tests
+
+1.  Verify that a NIC with the `bridge` binding method can be added to a running VM.
+2.  Verify that a NIC with the `SR-IOV` binding method can be added to a running VM.
+3.  Verify that attempts to add a NIC with any other binding method are blocked.
```
rrajendran17: Going forward, if all our VM/pod nics support hotpluggable nics by default, then I think we should still allow other types but a restart will be required.

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+
+-   `addNic` will patch the `VirtualMachine` resource by adding new elements to `spec.template.spec.domain.devices.interfaces` and `spec.template.spec.networks`. The interface `model` will be hardcoded to `virtio`, and only `bridge` and `SR-IOV` binding methods will be permitted.
+
+-   `removeNic` will patch the `VirtualMachine` resource by setting the `state` of the specified interface in `spec.template.spec.domain.devices.interfaces` to `absent`. This action will only be permitted for interfaces using the `bridge` binding method.
+
+### Test plan
+
+#### Regression tests
+
+Since this enhancement migrates the Multus CNI from a thin to a thick plugin architecture, all existing network functionalities, including the storage network, must be thoroughly tested to prevent regressions.
```
rrajendran17: I remember the LH pods like instance manager and backing image will also be not restarted using hotpluggable feature when storage networks is configured

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+4.  Verify that a NIC with the `bridge` binding method can be removed from a running VM.
+5.  Verify that attempts to remove a NIC with any other binding method are blocked.
+
+### Upgrade strategy
+
+The NIC hot-plug feature must be compatible with VMs created prior to the upgrade. Additionally, the migration of Multus CNI from the thin to the thick plugin architecture must be verified to ensure a seamless Harvester upgrade process.
+
+## Note
+
+The KubeVirt documentation[^1] specifies a maximum number of interfaces that can be hot-plugged. This limitation should be clearly communicated in the Harvester user documentation.
```
rrajendran17, line 159: For normal virtio interfaces, we can support upto 256 maximum as per my observation.
Since multi function is on and with this, multiple interfaces can use single PCI slot, and in our case upto 8 interfaces are using 1 PCI slot, With this we will be able to extend up to 32 * 8 = 256 interfaces as best case.

It will be good if we can test this and document the best case limit with hotplugsnics as well.

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+
+1.  Verify that a NIC with the `bridge` binding method can be added to a running VM.
+2.  Verify that a NIC with the `SR-IOV` binding method can be added to a running VM.
+3.  Verify that attempts to add a NIC with any other binding method are blocked.
+4.  Verify that a NIC with the `bridge` binding method can be removed from a running VM.
+5.  Verify that attempts to remove a NIC with any other binding method are blocked.
+
+### Upgrade strategy
+
+The NIC hot-plug feature must be compatible with VMs created prior to the upgrade. Additionally, the migration of Multus CNI from the thin to the thick plugin architecture must be verified to ensure a seamless Harvester upgrade process.
```
rrajendran17: May be we can add cases like what can be affected by upgrade.
Since creation/addition of interfaces is done after the node and pods are running the newer version, ideally there should not be any issues with hotplugnics.

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Note
+
+The KubeVirt documentation[^1] specifies a maximum number of interfaces that can be hot-plugged. This limitation should be clearly communicated in the Harvester user documentation.
+
+A `RestartRequired` condition has been observed on the `VirtualMachine` resource after a NIC hot-plug operation is followed by a VM reboot. This appears to be incorrect behavior, and we will coordinate with the KubeVirt community to investigate and resolve the issue.
```
rrajendran17: I think this can become a blocker later on where network config changes on VM will require a reboot which defeats the purpose of hotplugnics.

enhancements/20251009-nic-hotplug.md
wheatdog: Good point, my original UX idea is more like what we already have for volume hotplug -- buttons of "Add Volume" and "Detach Volume". I agree that it's better if we could just edit the VM config and save it without restarting.

enhancements/20251009-nic-hotplug.md
wheatdog: I'll follow up closely with KubeVirt via [this issue](https://github.com/kubevirt/kubevirt/issues/15881)

enhancements/20251009-nic-hotplug.md
wheatdog: Agree. I'll test this and document it.

enhancements/20251009-nic-hotplug.md
wheatdog: I'm not sure whether there would be some caveats for migrating Mutlus from the thin to the thick plugin. I'll check their document and do some tests.

As for the interface hot-plug for a VM created before migration, I would also test it and see if it works.

enhancements/20251009-nic-hotplug.md
wheatdog: > Hot-unplug is supported only for interfaces connected through [bridge binding](http://kubevirt.io/api-reference/main/definitions.html#_v1_interfacebridge).

As mentioned [here](https://kubevirt.io/user-guide/network/hotplug_interfaces/), since `virtio` model only is not explicitly stated, I'm not sure whether it's possible to hot-unplug other model. I'll do some tests.

enhancements/20251009-nic-hotplug.md
wheatdog: As suggested by @ihcsim and @ibrokethecloud during the meeting, I would try it on a vanilla KubeVirt environment and see if this issue remains.

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+
+-   `removeNic` will patch the `VirtualMachine` resource by setting the `state` of the specified interface in `spec.template.spec.domain.devices.interfaces` to `absent`. This action will only be permitted for interfaces using the `bridge` binding method.
+
+### Test plan
+
+#### Regression tests
+
+Since this enhancement migrates the Multus CNI from a thin to a thick plugin architecture, all existing network functionalities, including the storage network, must be thoroughly tested to prevent regressions.
+
+#### Functional tests
```
wheatdog, line 145: As suggested by @ihcsim during meeting, we could try hot-plug and hot-unplug multiple times and observe the behavior.

enhancements/20251009-nic-hotplug.md
wheatdog: revisit: @rrajendran17 @ibrokethecloud as discussed during the meeting about the UX, if we have addition validations need to make for hot-plug and hot-unplug, do you think it's a better idea to provide them as separate actions/buttons instead of letting user directly change VM config?

ibrokethecloud:
We should call out explicitly that in the current implementation we will not be supporting `kubeovn` based VM networks as they use a custom binding to allow dhcp capability to work correctly.

From an implementation perspective, we should have a few more api calls
* `isHotPluggingPossible` which checks if VM is live migrate-able based on conditions on VM status, and if we have not already hit the upper limit on allowable hot plug count for nics.
* `VMNetworkList` which returns bridge vlan based VM Networks that can be used to create hot plug nics for the VM by filter VMNetworks from the namespace of the VM
*

enhancements/20251009-nic-hotplug.md
wheatdog: This issue doesn't occur with vanilla KubeVirt. The problem stemmed from the VMI mutating webhook in Harvester, where mac addresses were set in the VMI spec but weren't backported to the VM spec. This behavior (https://github.com/harvester/harvester/pull/8339, https://github.com/harvester/harvester/issues/6844#issuecomment-2551020204, https://github.com/harvester/harvester/pull/7402, https://github.com/harvester/harvester/pull/7084) is introduced to fix another issue.

I'm trying to come up with a solution for it.

enhancements/20251009-nic-hotplug.md
```diff
@@ -0,0 +1,133 @@
...
+
+1.  Verify that a NIC with the `bridge` binding method can be added to a running VM.
+2.  Verify that a NIC with the `SR-IOV` binding method can be added to a running VM.
+3.  Verify that attempts to add a NIC with any other binding method are blocked.
+4.  Verify that a NIC with the `bridge` binding method can be removed from a running VM.
+5.  Verify that attempts to remove a NIC with any other binding method are blocked.
+
```
mingshuoqiu: Maybe we can add corner test case.
 -  Add test for invalid cases (unsupported binding, removing non-existent NIC).

What is the expected VM condition or event for confirming success?

w13915984028:
Thanks for the PR,  connect with main issue https://github.com/harvester/harvester/issues/8283, and also a few comments:

(1) For guest cluster VMs(with label `harvesterhci.io/creator: docker-machine-driver-harvester`), they should/might need to be protected to hot-plug nic before the integration with Rancher Manager is done.

(2) For a running VM, are there identifications to show which NICs were ever hot-pluged? Is it allowed to un-plug any existing NICs (bridge or sriov type) or the original initial NIC is protected? 

the concern is: the initial first nic sometimes is more like a mgmt nic, if it is un-plugged accidentally, the VM might lose it's mgmt IP and can't be connected.

(3) Cross check with managed-dhcp https://docs.harvesterhci.io/v1.6/advanced/addons/managed-dhcp#limitations

enhancements/20251009-nic-hotplug.md
wheatdog: I've further tested hot-plug / hot-unplug using different models like e1000, e1000e, ne2k_pci, pcnet, rtl8139. To my surprise that most of them actually behave correctly within the guest OS. I can see them popped up or disappeared by inspecting the output of `ip a`. This could be the guest OS of my choice compensates the limitation of legacy NIC models. 

Given that `virtio` is often the most recommended model for modern guests, we can limit the goal to `virtio` model only for both hot-plug and hot-unplug. This could also reduce the QA efforts.

enhancements/20251009-nic-hotplug.md
wheatdog: To reduce the scope of this HEP, we'll focus on hot-plug and hot-unplug of `virtio` NIC with `bridge` binding method.

wheatdog:
I've revised the HEP with https://github.com/harvester/harvester/pull/9295/commits/44ba1f24fc3715a913085020aeb00399efb38732

- limit the scope to virtio model and bridge binding method
- use migration-based method
- add new APIs to listHotpluggableVmNetworks and listHotunpluggableNics
- allow MAC addresses be specified during hot-plugging
- adjust the MAC addresses preserving strategy to prevent unpexected RestartRequired
- refine the test plan

wheatdog:
@ibrokethecloud thanks for the comments. I've added those suggestion to the revised HEP.

@w13915984028 thanks for the comments. I'll also add them back to HEP when we have conclusion.

> (1) For guest cluster VMs(with label harvesterhci.io/creator: docker-machine-driver-harvester), they should/might need to be protected to hot-plug nic before the integration with Rancher Manager is done.

Good point, I would add this to HEP.

> (2) For a running VM, are there identifications to show which NICs were ever hot-pluged? Is it allowed to un-plug any existing NICs (bridge or sriov type) or the original initial NIC is protected?
> 
> the concern is: the initial first nic sometimes is more like a mgmt nic, if it is un-plugged accidentally, the VM might lose it's mgmt IP and can't be connected.
>
> (3) Cross check with managed-dhcp https://docs.harvesterhci.io/v1.6/advanced/addons/managed-dhcp#limitations

Sounds good. It seems that there are a few evidences that the initial set of interfaces, decided during the VM creation, are quite special:
- the initial NIC is sometimes used for management
- managed-dhcp: IP addresses are not allocated or deallocated when you add or remove network interfaces after the virtual machine is created. The actual MAC addresses are recorded in the VirtualMachineNetworkConfig CRs.

We could prevent them from being hot-unplugged if we can confirm that there is no such use case or it's rarely useful.

On the other hand, those interfaces could still be removed / modified after stopping the VM. I couldn't decide whether this prevention is really worthy.

w13915984028:
@wheatdog 
>We could prevent them from being hot-unplugged if we can confirm that there is no such use case or it's rarely useful.

>On the other hand, those interfaces could still be removed / modified after stopping the VM. I couldn't decide whether this prevention is really worthy.

Both cases are at least valid. We could consider to add a setting to toggle the protection upon initial NICs.

When a VM is stopped, user can change any spec. For a running VM, hot-unplug it's mgmt NIC might be kind of mis-opration, and also affects the `migration` opertion potentially.

Just recalled this case: Harvester supports the migration-network https://docs.harvesterhci.io/v1.6/advanced/vm-migration-network, the NIC attached to this network should be protected. Please double check, thanks.

wheatdog:
@w13915984028 Thanks. Making it a VM setting sounds good. I'll also check the migration network.

enhancements/20251009-nic-hotplug.md
wheatdog: The solution is to generate MAC Addresses during VM creation and save it to the VM spec. Additionally, backfill MAC addresses from annotation to spec while stopping or restarting VM.


https://github.com/harvester/harvester/pull/9395

wheatdog:
@w13915984028 

> For a running VM, hot-unplug it's mgmt NIC might be kind of mis-opration, and also affects the migration opertion potentially.
> 
> Just recalled this case: Harvester supports the migration-network https://docs.harvesterhci.io/v1.6/advanced/vm-migration-network, the NIC attached to this network should be protected. Please double check, thanks.

I have tested this feature. The `migration0` NIC mentioned in the document wouldn't be the target for hot-unplugging since its on the `virt-handler` pods not `virt-launcher` pods.

Additionally, hot-plugging and hot-unplugging is forbidden during transient states like `migrating`. This is handled in my implementation (https://github.com/harvester/harvester/pull/9395).


> > We could prevent them from being hot-unplugged if we can confirm that there is no such use case or it's rarely useful.
>
> > On the other hand, those interfaces could still be removed / modified after stopping the VM. I couldn't decide whether this prevention is really worthy.
> 
> Both cases are at least valid. We could consider to add a setting to toggle the protection upon initial NICs.

We might need more time to cook for both FE and BE. Given that the schedule is a little tight for this feature, I would like to postpone this enhancement to a future enhancement. What do you think? Thanks.

w13915984028:
@wheatdog Thanks for your validation, it is fine to move ahead with your suggested steps.

enhancements/20251009-nic-hotplug.md
wheatdog: The settled UI/UX flow can be referred to in [this UI PR](https://github.com/harvester/harvester-ui-extension/pull/582).

wheatdog:
@rrajendran17 @ibrokethecloud @w13915984028 could you help me review the revised HEP again? Thanks!

w13915984028 (approved):
LGTM, thanks.

wheatdog:
@mergify backport v1.7

enhancements/20251009-nic-hotplug.md
wheatdog: > Kubevirt reserves resources for 4 interface to allow later hotplug operations. The actual maximum amount of available resources depends on the machine type (e.g. q35 adds another PCI slot). For more information on maximum limits, see [libvirt documentation](https://libvirt.org/pci-hotplug.html).

The quote above is from https://kubevirt.io/user-guide/network/hotplug_interfaces/#virtio-limitations. Turns out the actual behavior is complex. https://github.com/kubevirt/kubevirt/pull/14754 implemented a new way to determine the number of reserved PCI ports.

Following is my test result on Harvester

1 vCPU 2Gi Memory: 4 free PCI slots
```
$ lspci -t -v
-[0000:00]-+-00.0  Intel Corporation 82G33/G31/P35/P31 Express DRAM Controller
           +-01.0  Device 1234:1111
           +-02.0-[01]----00.0  Red Hat, Inc. Virtio 1.0 network device
           +-02.1-[02]--
           +-02.2-[03]--
           +-02.3-[04]--
           +-02.4-[05]----00.0  Red Hat, Inc. Virtio 1.0 SCSI
           +-02.5-[06]----00.0  Red Hat, Inc. Virtio 1.0 console
           +-02.6-[07]----00.0  Red Hat, Inc. Virtio 1.0 block device
           +-02.7-[08]----00.0  Red Hat, Inc. Virtio 1.0 block device
           +-03.0-[09]----00.0  Red Hat, Inc. Virtio 1.0 memory balloon
           +-03.1-[0a]--
           +-1f.0  Intel Corporation 82801IB (ICH9) LPC Interface Controller
           +-1f.2  Intel Corporation 82801IR/IO/IH (ICH9R/DO/DH) 6 port SATA Controller [AHCI mode]
           \-1f.3  Intel Corporation 82801I (ICH9 Family) SMBus Controller
``` 

1 vCPU 3Gi memory: 10 free PCI slots
```
$ lspci -t -v
-[0000:00]-+-00.0  Intel Corporation 82G33/G31/P35/P31 Express DRAM Controller
           +-01.0  Device 1234:1111
           +-02.0-[01]----00.0  Red Hat, Inc. Virtio 1.0 network device
           +-02.1-[02]--
           +-02.2-[03]--
           +-02.3-[04]--
           +-02.4-[05]--
           +-02.5-[06]--
           +-02.6-[07]--
           +-02.7-[08]--
           +-03.0-[09]--
           +-03.1-[0a]--
           +-03.2-[0b]----00.0  Red Hat, Inc. Virtio 1.0 SCSI
           +-03.3-[0c]----00.0  Red Hat, Inc. Virtio 1.0 console
           +-03.4-[0d]----00.0  Red Hat, Inc. Virtio 1.0 block device
           +-03.5-[0e]----00.0  Red Hat, Inc. Virtio 1.0 block device
           +-03.6-[0f]----00.0  Red Hat, Inc. Virtio 1.0 memory balloon
           +-03.7-[10]--
           +-1f.0  Intel Corporation 82801IB (ICH9) LPC Interface Controller
           +-1f.2  Intel Corporation 82801IR/IO/IH (ICH9R/DO/DH) 6 port SATA Controller [AHCI mode]
           \-1f.3  Intel Corporation 82801I (ICH9 Family) SMBus Controller
```

I'll try to add this to our document

PR #9299, opened 2025-10-13
feat: deny empty memory and patch default cpu

FrankYang0529:
#### Problem:
Currently, we allowed users to create following VMs:
* VM without memory in `resources.limits.memory` and `memory.guest`. This kind of VM cannot start.
* VM with null value in `cpu`. This kind of VM can start with default value `1`.

#### Solution:
* VM without memory in `resources.limits.memory` and `memory.guest`. Deny the request.
* VM with null value in `cpu`. Use mutator to patch default value to it.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9302

#### Test plan:
* Create a VM without memory in `resources.limits.memory` and `memory.guest`. Webhook should deny the request.

Screenshot 2025-10-13 at 3 06 16 PM

* Create a VM with null value in `cpu`. Webhook should patch default value to it.

### DIFF
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -70,6 +70,8 @@ func (m *vmMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patch
 		return patchOps, err
 	}
 
+	patchOps = patchDefaultCPU(vm, patchOps)
+
 	patchOps, err = m.patchAffinity(vm, patchOps)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


innobead:
Do we have a corresponding issue for this?

FrankYang0529:
> Do we have a corresponding issue for this?

I just created https://github.com/harvester/harvester/issues/9302 for it.

Vicente-Cheng (dismissed):
lgtm, thanks for the quick fix

Yu-Jack (approved):

Issue #7025, opened 2024-11-20
[FEATURE] Bump base OS to SL Micro 6.1


bk201:
**Is your enhancement related to a problem? Please describe.**

Bump base OS to SLE Micro 6.

**Describe the solution you'd like**

Evaluate bumping the base OS to SLE Micro 6.

**Describe alternatives you've considered**

**Additional context**

innobead:
@tserong Suggest starting to do some evaluation first during 1.5 development, but the priority is less than your items planned in 1.5 already.

tserong:
For evaluation purposes, I branched our `isv:Rancher:Harvester:OS:Dev/baseos` on OBS to https://build.opensuse.org/project/show/home:tserong:branches:isv:Rancher:Harvester:OS:Dev, and updated the `Dockerfile` in `baseos` to use `suse/sl-micro/6.0/baremetal-os-container:latest`. After some fiddling around I got that to build, and in the process discovered the following items we're going to need to investigate further:

- SL Micro 6.0 does not support AppArmor. Instead it runs SELinux in enforced mode as a default (see https://www.suse.com/releasenotes/x86_64/SLE-Micro/6.0/index.html#v60-securitysecurity-framework). I'm not yet sure what implications this has for Harvester.
- SL Micro 6.0 uses NetworkManager instead of wicked. We have a bunch of code in various places in harvester which expects networking to be managed by wicked, so this would all need to be checked/changed/migrated to NetworkManager.
- elemental-toolkit is updated from v1.1.6 to v2.1.3.

mcerveny:
I am not happy that upgrade was postponed to 1.7.0 (~ December 2025)  I need 6.x kernel to better support: 
- Thunderbold/PCIe bridges (unusable in 5.x kernel). :-(
- longhorn ublk does not work https://github.com/longhorn/longhorn/issues/9456#issuecomment-2811508087

Please start porting at least to "openSUSE Leap Micro 6.2" (6.12 LTS kernel) ([slack](https://rancher-users.slack.com/archives/C01GKHKAG0K/p1756492483166309)).

mcerveny:
Did you rethink to replace [own installer](https://github.com/harvester/harvester-installer) with participation and integration of [Agama](https://agama-project.github.io/) installer ?

lkocman:
[...]

PR #9303, opened 2025-10-13
 Switch to SL Micro 6.1

tserong:
#### Problem:
Need to bump the base OS to SL Micro 6.1

#### Solution:
The installer changes in https://github.com/harvester/harvester-installer/pull/1163, plus this PR which pulls in the elemental-toolkit build from registry.opensuse.org/isv/rancher/harvester/os/sl-micro-6.1/main/baseos:latest.

The scary thing here is that elemental-toolkit built against SL Micro 6.1 needs a newer glibc than is available in the SLE Micro 5.5 host. I've worked around this by bind mounting `/lib64` from the upgrade container over the host environment for the duration of `elemental upgrade`. Incredibly, this works.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/7025

#### Test plan:
TBD (basically: upgrade Harvester and make sure it works)

This PR plus https://github.com/harvester/harvester-installer/pull/1163 will let you upgrade an existing v1.6.x system to v1.7.0-dev based on SL Micro 6.1, provided you don't run into https://github.com/harvester/harvester/issues/9349 while trying to upgrade :-/

As mentioned in https://github.com/harvester/harvester-installer/pull/1163, my plan is to merge the os2 sl-micro-6.1 branch back to the sle-micro branch, but I thought it might be least disruptive to daily builds and QA and other folks, to merge this and the related harvester PR, including the switch to the os2 sl-micro-6 branch image, then do the os2 and OBS merges, then open another PR to switch harvester-installer back to using the sle-micro branch once that's settled.

Please ignore the CodeFactor complaint. I'll fix that once we have a separate v1.7 project on OBS and can set tags appropriately.

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -675,6 +675,12 @@ stages:
 EOF
   fi
 
+  # SLE Micro 5.5 uses /usr/lib/ssh/sftp-server
+  # SL Micro 6.1 uses /usr/libexec/ssh/sftp-server
+  if [ -e ${HOST_DIR}/etc/ssh/sshd_config.d/sftp.conf ]; then
+    sed -i 's%/usr/lib/ssh/sftp-server%/usr/libexec/ssh/sftp-server%' ${HOST_DIR}/etc/ssh/sshd_config.d/sftp.conf
+  fi
+
   umount $tmp_rootfs_mount
   rm -rf $tmp_rootfs_squashfs
 
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -594,11 +594,26 @@ EOF
 
   elemental_upgrade_log="${UPGRADE_TMP_DIR#"$HOST_DIR"}/elemental-upgrade-$(date +%Y%m%d%H%M%S).log"
   local ret=0
+  # elemental-toolkit built for SL Micro 6.1 needs a newer glibc than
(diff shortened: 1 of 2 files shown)


tserong:
Rebased to pick up https://github.com/harvester/harvester/pull/9378 so this PR can be tested properly

package/upgrade/upgrade_node.sh
```diff
@@ -660,6 +675,10 @@ stages:
 EOF
   fi
 
+  # SLE Micro 5.5 uses /usr/lib/ssh/sftp-server
+  # SL Micro 6.1 uses /usr/libexec/ssh/sftp-server
+  sed -i 's%/usr/lib/ssh/sftp-server%/usr/libexec/ssh/sftp-server%' ${HOST_DIR}/etc/ssh/sshd_config.d/sftp.conf
```
ibrokethecloud: unless SFTP is enabled in harvester config the directory and sftp config file will not exist.
we should gate this `sed` in a check which verifies that file exists

ibrokethecloud (approved):
lgtm. was able to upgrade a single node v1.6.0 install to an iso build with this PR and associated installer PR https://github.com/harvester/harvester-installer/pull/1163

Vicente-Cheng (approved):
lgtm!
verified with the fresh installation and upgrade.

basic function works well!
Thanks for the huge effort!!

Issue #9289, opened 2025-10-09
[BUG] Fail to enable rancher-logging addon during a multi-stages upgrade test


albinsun:
### Describe the Bug
During a multi-stages upgrade test 151 -> 160 -> 161.
We hit following issue when preparing prerequsites before upgrade to 161.

```
Error: INSTALLATION FAILED: Unable to continue with install: ClusterRole "logging-admin" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "rancher-logging": current value is "hvst-upgrade-md54b-upgradelog-operator"
```

<img width="1566" height="1036" alt="Image" src="https://github.com/user-attachments/assets/d9347515-caee-4d1c-b133-768fad818559" />

### To Reproduce

1. Setup 4 nodes harvester-v1.5.1
3. Create VMs
4. enable rancher-logging
6. Upgrade to v1.6.0
7. disable rancher-logging
8. Delete VMs
10. enable rancher-logging

### Expected Behavior

Should be able to enable rancher-logging addon

### Support Bundle for Troubleshooting

[supportbundle_eec15770-25ba-46a0-8628-abab28837547_2025-10-09T07-19-50Z.zip](https://github.com/user-attachments/files/22792000/supportbundle_eec15770-25ba-46a0-8628-abab28837547_2025-10-09T07-19-50Z.zip)

### Environment

- Harvester version: v1.6.0 (Just upgrade from v1.5.1)
- Impacted VM:
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Baremetal
- Rancher version:

### Additional context
1. Manual disable and reenable hits same problem.
2. Found `hvst-upgrade-md54b-upgradelog-clusterflow` and `hvst-upgrade-md54b-upgradelog-clusteroutput` in **Logging** -> **Cluster Flows** and **Cluster Outputs**
   <img width="1566" height="547" alt="Image" src="https://github.com/user-attachments/assets/08043c2f-54dd-4ecb-84df-dc98855812c3" />
   <img width="1566" height="547" alt="Image" src="https://github.com/user-attachments/assets/eef08f50-e55c-4c99-9e22-465b63761d4a" />

ibrokethecloud:
When logging is enabled during upgrade an `upgradelog` object is created for the associated `upgrade`

```
(⎈|default:harvester-system)➜  ~ k get upgrade.harvester
[...]

PR #9304, opened 2025-10-14
block enabling of rancher-logging addon if an upgradelog object exists

ibrokethecloud:
#### Problem:

If an upgrade is run with logging enabled, then the upgrade controller creates an associated `upgradelog` object.

The upgradelog controller checks if logging addon is enabled, and if not it creates a managedchart using the logging addon and deploys the chart as a managedchart. 

If a user attempts to enable rancher logging addon before the upgrade has been completed, the addon deployment fails since the rancher logging chart contains cluster scoped objects which already exist in the cluster as part of the managedchart deployment of rancher-logging

#### Solution:

Add a minor webhook validation which blocks a user from enabling rancher-logging addon if an upgrade is running.

#### Test plan:

https://github.com/harvester/harvester/issues/9289


tserong (dismissed):

ihcsim (approved):

w13915984028 (approved):
LGTM, thanks fro the fix.

Issue #9279, opened 2025-10-08
[TASK] Bump golang/k8s/BCI Image


Vicente-Cheng:
### Task request

Bump the following components:

- golang -> v1.25.2
- k8s -> v1.33 or v1.32 (depends on the kubevirt/client-go)
- BCI image -> 15.7

| Done | Repository | Owner |
| --- | --- | --- |
| ✅ | https://github.com/harvester/cloud-provider-harvester | @FrankYang0529  |
| ✅ | https://github.com/harvester/docker-machine-driver-harvester | @FrankYang0529  |
| ✅ | https://github.com/harvester/eventrouter | @martindekov  |
| ✅ | https://github.com/harvester/go-common | @Vicente-Cheng   |
| ✅ | https://github.com/harvester/harvester | @FrankYang0529 |
| ✅ | https://github.com/harvester/harvester-csi-driver | @Vicente-Cheng |
| ✅ | https://github.com/harvester/networkfs-manager | @Vicente-Cheng |
| ✅ | https://github.com/harvester/csi-driver-lvm | @Vicente-Cheng |
| ✅ | https://github.com/harvester/harvester-installer | @tserong |
| ✅ | https://github.com/harvester/load-balancer-harvester | @martindekov   |
| ✅ | https://github.com/harvester/network-controller-harvester | @mingshuoqiu  |
| ✅ | https://github.com/harvester/node-disk-manager | @Vicente-Cheng |
| ✅ | https://github.com/harvester/node-manager | @brandboat  |
| ✅ | https://github.com/harvester/pcidevices | @Yu-Jack  |
| ✅ | https://github.com/harvester/seeder | @ibrokethecloud |
| ✅ | https://github.com/harvester/rancherd | @FrankYang0529 |
| ✅ | https://github.com/rancher/support-bundle-kit | @Yu-Jack  |
| ✅| https://github.com/harvester/vm-dhcp-controller | @rrajendran17  |
| ✅ | https://github.com/harvester/vm-import-controller | @votdev   |
| ✅ | https://github.com/harvester/terraform-provider-harvester | @FrankYang0529 |
| ✅ | https://github.com/harvester/storage-validator | @ibrokethecloud  |
| ✅ | https://github.com/harvester/kubeovn-operator | @ibrokethecloud  |

Vicente-Cheng:
All components were updated. Close as resolved.

PR #9306, opened 2025-10-14
feat: bump golang to v1.25

FrankYang0529:
#### Related Issue(s):

https://github.com/harvester/harvester/issues/9279


Vicente-Cheng (approved):
thanks!

Issue #9313, opened 2025-10-15
[BUG] VM Image upload issue with 3rd-party storage


Vicente-Cheng:
### Describe the Bug

There are two symptoms we observed when uploading the image to a third-party storage.
- stuck on 0%
- stuck on 99% and then see the error "context canceled"

### To Reproduce

1. create harvester cluster
2. configure 3rd-party storage (using LHv2 would be easy for setup)

For 0% issue: 
1. upload the VM image with the above 3rd-party storage
  1.1. cancel the upload when you saw the progress
2. upload another VM Image with the above 3rd-party storage
3. you will see the progress is stuck on 0% 

For 99% -> context canceled issue
1. upload the VM image with larger qcow2 image (the source qcow2 image should be more 4GB)
2. you will see the progress is stuck on 99% then report the error context canceled

 

### Expected Behavior

Upload without any issues.

### Support Bundle for Troubleshooting

None

### Environment

- Harvester version: v1.6.0
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Additional context

The upload is actually running.
You can check the corresponding pod, like harvester/cdi-upload-xxx, to ensure it is still running.

And the VM image would eventually be ready.

WebberHuang1118:
> For 99% -> context canceled issue
> - upload the VM image with larger qcow2 image (the source qcow2 image should be more 4GB)
> - you will see the progress is stuck on 99% then report the error context canceled

This is similar to [Harvester issue #3450](https://github.com/harvester/harvester/issues/3450#issuecomment-1443937746).

By default, Harvester’s NGINX ingress has a timeout of **60 seconds**.
```
$ kubectl exec -n kube-system -i -t rke2-ingress-nginx-controller-kfs7k -- /bin/bash
www-data@rke2-ingress-nginx-controller-kfs7k:/etc/nginx> grep 60 nginx.conf
# Configuration checksum: 6054469776559397366
        client_header_timeout           60s;
        client_body_timeout             60s;
                keepalive_timeout  60s;
                        proxy_send_timeout                      60s;
                        proxy_read_timeout                      60s;
```

During the final phase of uploading an image to third-party storage, CDI needs time to convert the image if it’s in **qcow2** format. When this conversion takes longer than 60 seconds, the request may time out with a `context canceled` error.

[...]

PR #9315, opened 2025-10-15
fix vm image upload issues

Vicente-Cheng:
#### Problem:
The VM Image will be stuck on 0%/99% and will see the context canceled error.

#### Solution:
improve the upload behavior

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9313

#### Test plan:

1. create harvester cluster
2. configure 3rd-party storage (using LHv2 would be easy for setup)

case 1:
1. upload the VM image with the above 3rd-party storage
  1.1. cancel the upload when you saw the progress
2. upload another VM Image with the above 3rd-party storage

case 2
1. upload the VM image with larger qcow2 image (the source qcow2 image should be more 4GB)

The above cases should pass w/o any errors

### DIFF
--- a/pkg/image/cdi/upload.go
+++ b/pkg/image/cdi/upload.go
@@ -220,6 +220,8 @@ func (cu *Uploader) DoUpload(vmImg *harvesterv1.VirtualMachineImage, req *http.R
 	progress := &ProgressUpdater{
 		targetBytes:       fileSize,
 		lastTime:          time.Now(),
+		imageNS:           vmImg.Namespace,
+		imageName:         vmImg.Name,
 		vmImgUpdateLocker: updaterLocker,
 		vmImgCond:         updaterCond,
 	}
@@ -266,13 +268,9 @@ func (cu *Uploader) DoUpload(vmImg *harvesterv1.VirtualMachineImage, req *http.R
 		return err
 	}
 
-	// final wake up, we need to wait the DataVolume status to be succeeded
-	// ensure the VMImage status is updated, wake up the updater
-	if err := wait.PollUntilContextTimeout(context.Background(), tickPolling, tickTimeout, true, func(context.Context) (bool, error) {
-		return cu.waitDataVolumeStatus(dvNamespace, dvName, cdiv1.Succeeded)
-	}); err != nil {
-		return fmt.Errorf("failed to wait for VMImage status: %v", err)
-	}
+	// we don't wait the DataVolume to be Succeeded here.
+	// the controller will help with that.
+	// for the progress, we set it to 99% here on progress updater
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


Vicente-Cheng:
@mergifyio backport v1.6

pkg/image/cdi/upload.go
```diff
@@ -351,13 +351,19 @@ func (cu *Uploader) updateVMImageProgress(vmImg *harvesterv1.VirtualMachineImage
...
+			uploadCompleted = true
 		}
 		logrus.Debugf("DataVolume %s/%s upload progress: %d/%d (%v)", targetDVNs, targetDVName, currentBytes, targetSize, progress)
 		_, err = cu.vmio.Importing(vmImg, "Image Importing", int(progress))
 		if err != nil {
 			logrus.Errorf("failed to update VMImage status: %v", err)
 		}
 		cond.L.Unlock()
+		if uploadCompleted {
+			logrus.Infof("Upload completed, stop the progress updater")
```
brandboat: could you please also print the vmimage namespace/name here?

would be great if you can help modify the same thing in https://github.com/harvester/harvester/blob/master/pkg/image/cdi/common.go#L53 for debug purpose, thanks!

pkg/image/cdi/upload.go
```diff
@@ -303,13 +297,19 @@ func (cu *Uploader) waitDataVolumeStatus(namespace, name string, targetState cdi
...
+	vmImgNS := vmImg.Namespace
+	vmImgName := vmImg.Name
 	for {
 		cond.L.Lock()
 		cond.Wait()
 
-		if cu.foundError != nil {
-			logrus.Debugf("Found error, stop the progress updater")
+		if *anyErr != nil {
+			logrus.Errorf("Found error (%v), stop the progress updater", *anyErr)
```
brandboat: ditto.

brandboat (dismissed):
LGTM, many thanks!

WebberHuang1118 (dismissed):
LGTM, thanks.

Just the functions [upload](https://github.com/Vicente-Cheng/harvester/blob/9841f7524f8a6cddd6ef0192fcacd33443a68e98/pkg/image/cdi/upload.go#L65-L274) and [updateVMImageProgress](https://github.com/Vicente-Cheng/harvester/blob/9841f7524f8a6cddd6ef0192fcacd33443a68e98/pkg/image/cdi/upload.go#L300-L370) are a bit complex. We might consider refactoring them in the future.

WebberHuang1118 (approved):

pkg/image/cdi/upload.go
Vicente-Cheng: updated

pkg/image/cdi/upload.go
Vicente-Cheng: updated

brandboat (approved):

Issue #6696, opened 2024-10-02
[BUG] when you restore a backup and replace existing it doesn't remove previous volumes


noahgildersleeve:
**Describe the bug**

When you restore a backup and select replace existing and select delete previous volumes it doesn't delete them.
**To Reproduce**
1. Create a VM
2. Set a backup target
3. Create 3 chained backups with files changed in between
4. Restore to backup-1
5. Check for volumes
6. Restore to backup-2

**Expected behavior**

It should remove the old volumes

**Support bundle**

[supportbundle_8292c761-b4a1-4a28-b63d-4645a3b55689_2024-10-02T22-04-14Z.zip](https://github.com/user-attachments/files/17236914/supportbundle_8292c761-b4a1-4a28-b63d-4645a3b55689_2024-10-02T22-04-14Z.zip)

**Environment**
 - Harvester ISO version: v1.4-aa4976f-head
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 node libvirt setup

**Additional context**

Found while testing #6371 and #2237 

It looks like the deletion policy is being set in the payload
```
{"apiVersion":"harvesterhci.io/v1beta1","kind":"VirtualMachineRestore","metadata":{"name":"","namespace":"default","generateName":"restore-backupoff3-1-"},"type":"harvesterhci.io.virtualmachinerestore","spec":{"target":{"apiGroup":"kubevirt.io","kind":"VirtualMachine","name":"opensuse-03"},"virtualMachineBackupName":"backupoff3-1","deletionPolicy":"delete","virtualMachineBackupNamespace":"default"}}
```

![Screenshot_20241002_151045](https://github.com/user-attachments/assets/d7804fd5-ba2e-480e-a8e6-cd899568b7f1)
![Screenshot_20241002_150851](https://github.com/user-attachments/assets/1a3a2a62-2dfd-4b07-be2f-180650f88047)

This was tested against both NFS and S3 backup scenarios

WebberHuang1118:
It should not be a regression issue, it's an enhancement from my side.

[...]

PR #9324, opened 2025-10-16
fix: Track latest VM volume for deletion when restoring with replace-and-delete policy

WebberHuang1118:
#### Problem:
When restoring a VM with the replace-and-delete policy, the old PVC isn’t deleted.

#### Solution:
Track the latest VM volume for deletion when restoring with the replace-and-delete policy

#### Related Issue(s):
#6696 

#### Test plan:
1. Update the Harvester deployment with this PR.
2. Create a VM and write some data to it.
3. Create a VMBackup for the VM.
4. Stop the VM after the VMBackup completes.
5. Restore the backup to the original VM, using the delete policy for the original volumes.
6. After the VMRestore completes, the original VM should be deleted.
7. Stop the VM.
8. Repeat steps 5–7 several times. Each time a new VMRestore completes, the previous VM’s volumes should be deleted.

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -421,26 +421,51 @@ func (h *RestoreHandler) initStatus(restore *harvesterv1.VirtualMachineRestore) 
 	return nil
 }
 
-func (h *RestoreHandler) initVolumesStatus(vmRestore *harvesterv1.VirtualMachineRestore, backup *harvesterv1.VirtualMachineBackup) error {
-	restoreCpy := vmRestore.DeepCopy()
+// getDeletedVolumesFromCurrentVM extracts PVC volume names from the current VM referenced in backup's spec.source
+func (h *RestoreHandler) getDeletedVolumesFromCurrentVM(backup *harvesterv1.VirtualMachineBackup) ([]string, error) {
+	if backup.Spec.Source.Kind != kubevirtv1.VirtualMachineGroupVersionKind.Kind {
+		return nil, fmt.Errorf("unsupported backup source kind: %s, expected %s", backup.Spec.Source.Kind, kubevirtv1.VirtualMachineGroupVersionKind.Kind)
+	}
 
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (approved):
lgtm, thanks for fixing that!

FrankYang0529 (approved):
Overall LGTM. I have a question about an edge case.

* Create a VM `vm1`.
* Create a VMBackup for `vm1`.
* Remove the VM `vm1`.
* Create a VM `vm1`.
* Stop the VM `vm1`.
* Use "Replace Existing" to restore the VMBackup.

For this case, the second `vm1` is not the first `vm1`. Should we use `sourceUID` in VMBackup status to avoid this?

WebberHuang1118:
> Overall LGTM. I have a question about an edge case.
> 
> * Create a VM `vm1`.
> * Create a VMBackup for `vm1`.
> * Remove the VM `vm1`.
> * Create a VM `vm1`.
> * Stop the VM `vm1`.
> * Use "Replace Existing" to restore the VMBackup.
> 
> For this case, the second `vm1` is not the first `vm1`. Should we use `sourceUID` in VMBackup status to avoid this?

Hi @FrankYang0529,

Thanks for your feedback. If I understand correctly, this behavior has existed since day one and isn’t directly related to this issue or PR, right?

FrankYang0529:
Yes, ignore it.

WebberHuang1118:
> Overall LGTM. I have a question about an edge case.
> 
> * Create a VM `vm1`.
> * Create a VMBackup for `vm1`.
> * Remove the VM `vm1`.
> * Create a VM `vm1`.
> * Stop the VM `vm1`.
> * Use "Replace Existing" to restore the VMBackup.
> 
> For this case, the second `vm1` is not the first `vm1`. Should we use `sourceUID` in VMBackup status to avoid this?

Created [ticket #9357](https://github.com/harvester/harvester/issues/9357) to track this scenario.

Issue #8451, opened 2025-06-10
[ENHANCEMENT] Set default value for LHv2 clone strategy to csi-clone in CDI annotations


brandboat:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

In https://github.com/harvester/harvester/issues/8077, we introduced storage class annotations to sync CDI settings to the corresponding storage profile. Since Longhorn v2 will support CSI clone starting from v1.9.1, we need to set the default CDI clone strategy to `"csi-clone"` in the storage class annotation `cdi.harvesterhci.io/storageProfileCloneStrategy`.

lh issue: https://github.com/longhorn/longhorn/issues/7794

### Describe the solution you'd like

Set `cdi.harvesterhci.io/storageProfileCloneStrategy: "csi-clone"` as default value in the lhv2 Storage classses.

Vicente-Cheng:
The fast clone of Longhorn v2 volume was moved to v1.10.0.
So, moved this issue to Harvester v1.7.0

innobead:
Longhorn 1.10 will be released soon with the smart/fast-clone feature. 

cc @harvester/storage @PhanLe1010

brandboat:
During testing we found that under a **_single node_** harvester setup, creating a vm with longhorn v2 vm image with storage class number of replicas configured to 3 causes the VM creation process to get stuck.

The root cause is related to the cdi csi-clone flow which include three phases:
- CSIClonePhase - Creates temporary PVC via CSI clone
- PrepClaimPhase - Handles volume expansion if needed
- RebindPhase - Rebinds PV to target namespace

In the PrepClaimPhase, cdi performs volume expansion if the requested PVC size is larger than the VM image’s PVC size. However, this stage cannot proceed unless the volume’s ROBUSTNESS is in a healthy state. This issue can be mitigated by setting the SC number of replicas to 1.

In UI volumes page, you'll see 
<img width="511" height="121" alt="image" src="https://github.com/user-attachments/assets/1a445b26-22a6-4fa1-a01f-9545447eee17" />

And In pvc tmp-pvc-963e7dbc-77df-4485-94f7-ca116f5ecd04, below warning events pops up:

```
[...]

PR #9329, opened 2025-10-17
Set default value for LHv2 clone strategy to csi-clone in CDI annotations

brandboat:
#### Problem:

lhv2 now support csi-clone (snapshot based), we need to set cdi clonestrategy annotation default to csi-clone.

#### Solution:

lhv2 now support csi-clone (snapshot based), we need to set cdi clonestrategy annotation default to csi-clone.

#### Related Issue(s):

#8451 

#### Test plan:

1. setup single node lhv2 harvester environment.
2. create lhv2 storage class, mind that the number of replicas should be set to "1", no need to do further setting as we want to check  if the default values is right.
  image

3. check lhv2 storage class again, you'll see clone strategy become "csi-clone".
  image

4. create a vm image with lhv2 sc.
5. create a vm with vm image created in previous step.
---

Note: 
In the last step, if your lhv2 storage class number of replica is 3, vm creation will stuck. 
The reason is in lhv2 csi-clone, we firstly create a snapshot based on the vm image volume, and then do volume expansion, and vm creation stucks in this stage. 
In UI, you'll see 
image

In pvc tmp-pvc-963e7dbc-77df-4485-94f7-ca116f5ecd04, below warning events pops up:

```
Warning  VolumeResizeFailed     50s (x2 over 4m34s)    external-resizer                          driver.longhorn.io resize volume "pvc-36c778a5-a062-4fe4-86b3-b20c6be65307" by resizer "driver.longhorn.io" failed: rpc error: code = Internal desc = Bad response statusCode [500]. Status [500 Internal Server Error]. Body: [code=Internal Server Error, detail=, message=unable to expand volume pvc-36c778a5-a062-4fe4-86b3-b20c6be65307: cannot expand volume before replica scheduling success] from [http://longhorn-backend:9500/v1/volumes/pvc-36c778a5-a062-4fe4-86b3-b20c6be65307?action=expand]
```


Vicente-Cheng (approved):

ihcsim (approved):

WebberHuang1118:
> In the last step, if your lhv2 storage class number of replica is 3, vm creation will stuck.
> The reason is in lhv2 csi-clone, we firstly create a snapshot based on the vm image volume, and then do volume expansion, and vm creation stucks in this stage.

This part might need a documentation ticket for tracking, as it could be important for users.

We might also consider creating a Longhorn ticket to check whether online volume expansion should be allowed for degraded volumes. cc @Vicente-Cheng @derekbit

Vicente-Cheng:
> > In the last step, if your lhv2 storage class number of replica is 3, vm creation will stuck.
> > The reason is in lhv2 csi-clone, we firstly create a snapshot based on the vm image volume, and then do volume expansion, and vm creation stucks in this stage.
> 
> This part might need a documentation ticket for tracking, as it could be important for users.
> 
> We might also consider creating a Longhorn ticket to check whether online volume expansion should be allowed for degraded volumes. cc @Vicente-Cheng @derekbit

Thanks for the reminder.
I did not notice this part. Did this mean we cannot use clone for the current flow? (snapshot -> resize)

brandboat:
> Did this mean we cannot use clone for the current flow? (snapshot -> resize)

I think what Webber means is we need to warn user that before they perform lhv2 vm images, they need to make sure they have enough disk space and healthy volumes to prevent lhv2 vm image creation stuck. If all prerequisites are meet, using csi-clone is fine here.

brandboat:
> This part might need a documentation ticket for tracking, as it could be important for users.

@WebberHuang1118, thanks for the reminder, I already added `require-doc` label to the issue, will file doc pr later.

brandboat:
If there is no further comments, I'll merge the pr tomorrow. Thanks!

WebberHuang1118 (approved):
LGTM, thanks.

Vicente-Cheng (approved):
thanks for the explanation.

brandboat:
Please hold off this pr, currently csi-clone strategy is much slower than the copy mode due to the volume resizing stage. We may consider stick to the copy mode for now until we found out why volume resizing is so slow.  see https://github.com/harvester/harvester/issues/9344

brandboat:
The csiclone stuck issue has been resolved in Longhorn v1.12.0 ([commit 22fe84f](https://github.com/longhorn/longhorn-manager/commit/22fe84fc6952c6db8433f209515d4f604f6fd33f)), so we can move forward with this.

I ran a few experiments in a single-node environment comparing the csiclone and host-assisted copy strategies:
- 3.5 GB VM image (creating a 10 GB VM root disk):
  - csiclone: VM startup takes ~1.5 mins.
  - host-assisted: Takes ~1 min.
- 100 GB VM image (creating a 100 GB VM root disk):
  - csiclone: VM startup takes ~5 mins.
  - host-assisted: Takes ~15 mins.

Although csiclone can be slightly slower than host-assisted copy for smaller VM images, it is significantly faster for larger ones. Therefore, making it the default is definitely worth it.

Additionally, we are currently discussing ways to improve the csiclone workflow with the Longhorn team. There is currently a potential delay of 0 to 30 seconds before Longhorn creates the snapshot, and we are working on a solution to eliminate this delay.

WebberHuang1118 (approved):

brandboat:
@Mergify backport v1.9

Issue #8101, opened 2025-04-18
[FEATURE] Possibility to assign IP addresses to the Hosts in the Virtual Machine Networks


laralar:
### Describe the solution you'd like
 
I don't know if it would be possible to assign an IP address directly to the NAD instance for each host, so that each host has an IP address in that Virtual Machine Network. Or to be able , maybe through the dashboard, to assign, maybe via annotations, an IP address for the nodes for each  Virtual Machine Network

For example, in Ceph Rook , we can give an annotation to each node, network.rook.io/mon-ip: < IPAddress >, so that the Ceph monitor is published in this IP address instead of the Kubernetes interface (management in Harvester)

We could do some annotation like 
network.harvester.io/vlan.203-ip: < IPAddress >
network.harvester.io/storage.601-ip: < IPAddress >

And this will create the host interface vlan-br.203 and storage-br.203

Thanks.

### Describe alternatives you've considered

The workaround that I found is to 
1. Create the storage and vlan Cluster Networks
2. Add the storage-bo, storage-br, storage-br.601, vlan-bo, vlan-br, vlan-br.203 entries in /oem/90_custom.yaml with their respective Host IP Addresses in order to generate the respective wicked files in /etc/sysconfig/network

It is important to create the cluster networks first, otherwise the devices are not available later on while creating the cluster networks.

### Additional context

Currently I have a 4 node harvester cluster with 3 network cards:

1x 1 Gbps connection used in VLAN 200 for management
1x 10 Gbps connection used for Longhorn (let's suppose VLAN 600 for longhorn and VLAN 601 for Ceph cluster network)
1x 10 Gbps connection used for VLAN routing (let's suppose VLAN 203 for Ceph public network)

I can't add more cards right now.

For the management network the following devices are created in /oem/90_custom.yaml
- mgmt-bo 
- mgmt-br
- mgmt-br.200

and they generate the respective wicked files in /etc/sysconfig/network

I want to attach CSI Ceph RBD and CephFS volumes as well as CSI ISCSI volumes as secondary to VMs; however, during the installation of Ceph Rook in Harvester ( https://harvesterhci.io/kb/use_rook_ceph_external_storage/ ) Monitors are created by default in the Kubernetes Management network (1 x Gbps)

If I want to connect to either Ceph of an iSCSI provider, I would need to add either a dedicated card in each server and configure it accordingly (ideal) but since I don't have more cards, I need to share the longhorn and vlan 10 Gbps network cards

[...]

Issue #7834, opened 2025-03-12
[FEATURE] Support Layer 3 connectivity on user defined cluster network to act as underlay


rrajendran17:
**Is your enhancement request related to a problem? Please describe.**

When using kubeovn, underlay Layer 3 connectivity is required for communication between VMs running on different nodes.
**Currently Harvester has IP addresses configured only on mgmt interfaces, so kubeovn uses the mgmt-bo as underlay for communication between VMs running on different nodes.
Using same physical nic/nics part of mgmt-bo for VM traffic will impose traffic overhead as there is no segregation of VM traffic on a separate physical link.**

**Describe the solution you'd like**
Allow configuration of IP Addresses on user configured cluster networks and utilising them as underlay for Layer 3 connectivity between nodes when using kubeovn

1.Create a cluster network
2.Create a vlanconfig spanning all nodes and select uplink
3.Allow users to provide ip range for underlay for the cluster network.
4.IP addresses must be assigned from the ip range to all the nodes in the cluster.
5.Allow users to select the user configured cluster network for the VM attached NAD and tweak kubeovn to use the user selected cluster network when using vxlan or communication between VMs running on different nodes.
(This has to be tested for feasibility)

By default br-int ovs bridge using mgmt-bo is used as ovs bridge for underlay connectivity

With this enhancement/feature, ovs bridge must be created for every user defined cluster network and vswitches/VMs lports must be attached to these ovs bridges based on the cluster network user defines for the NAD/subnet

rrajendran17:
How to achieve an underlay on different interface other than mgmt when using vxlan with kubeovn

```
1.Create a cluster network cluster-1 and network config under it
2.Create a vm network 2013 under cluster-1
3.Create vlan L3 interface on all the 3 nodes in the cluster
vxlan-dev-1:

sudo ip link add link cluster-1-bo name br.2013 type vlan id 2013
sudo ip link set br.2013 up
sudo ip address add 10.115.18.10/21 dev br.2013


vxlan-dev-2

sudo ip link add link cluster-1-bo name br.2013 type vlan id 2013
sudo ip link set br.2013 up
sudo ip address add 10.115.18.11/21 dev br.2013

vxlan-dev-3


sudo ip link add link cluster-1-bo name br.2013 type vlan id 2013
sudo ip link set br.2013 up
sudo ip address add 10.115.18.12/21 dev br.2013

4.Verify ping is successful between all three nodes using the new underlay ip address
  10.115.16.0/21 network

[...]

Issue #7500, opened 2025-02-04
[EPIC] Load balancers on secondary networks


starbops:
The load balancing feature in Harvester consists of two parts:

1. The front-end IP address (so-called VIP/FIP address). This is the entry/access point of the load-balanced workloads users want to connect to.
2. The load-balanced back-end workloads (could be virtual machines or a set of pods behind an LB-type Service)

It is known that Harvester supports load balancers for the following two types of workloads:

- With the `VM`-type load balancers, the created LoadBalancer CRs are relevant to virtual machines running on the Harvester cluster. The front-end IP addresses are announced on the Harvester cluster's `mgmt-br` interface, and the back-end workloads are the virtual machines running on the Harvester cluster.
- With the `Cluster`- type load balancers, the created LoadBalancer CRs are merely intermediary/reference resources relevant to the actual entities - the LB-type Service resources on the downstream guest clusters. The front-end IP addresses are announced on the downstream guest clusters' management network interface (which one has the default route associated with). The back-end workloads are the pods on the downstream guest cluster.

Besides the workload types, there are two IPAMs for the front-end IP address for each of the LoadBalancer CR:

- The `Pool` IPAM means the LoadBalancer CRs are associated with user-defined IP pools, and the front-end IP addresses are picked from the pools by Harvester Load Balancer. The retrieved IP addresses are handed to kube-vip to be assigned to the network interface.
- The `DHCP` IPAM means the LoadBalancer CRs are not associated with any IP pool but delegate the responsibility of getting the front-end IP addresses through DHCP to kube-vip (Harvester Load Balancer does not deal with DHCP directly)

---

The current design and implementation apparently have limitations: the front-end IP addresses are all on the management network interface, regardless of the workload type and IPAM. **We want to extend the ability of the load balancing feature to cover networks other than the management one.** This will make the workload deployment more versatile and aid performance and security.

Adding secondary network support will make the load-balancing feature more challenging due to its complexity. Here, we will divide the feature by use cases:

- [ ] Support load balancers on secondary networks for VM workloads https://github.com/harvester/harvester/issues/9479
  - ClusterNetwork-aware
  - NAD-aware (including VLAN)
- [ ] #5486 
- [ ] #7327 is also somewhat relevant.

innobead:
LB support for container workloads on the underlay Harvester cluster (https://github.com/harvester/harvester/issues/5820) should be considered as the third type of support for our built-in LB. cc @starbops 

[...]

PR #9335, opened 2025-10-18
HEP for configuring network config on cluster networks

rrajendran17:
#### Problem:
1.This feature supports creation of vlan sub interfaces on mgmt and user defined cluster networks and configuration of IP addresses on these sub interfaces.

2.This feature allows users to select a vlan interface to be used as underlay for overlay networks used by VMs.

#### Solution:

case 1:
- Allow users to configure/unconfigure Layer 3 network Information using a new resource called hostnetworkconfig
- Harvester network controller implemets the new crd hostnetworkconfig, webhook validations and controller logic
to create a vlan interface, assign ip address given by user (static) or via DHCP.
- new hostnetworkconfig controller starts a lease manager to periodically renew ip addresses on interfaces acquired by DHCP.

case 2:
- set underlay as "true" on the hostnetworkconfig which user wants to act as underlay
- Network controller hostnetworkconfig controller handles the change, and update the annotations `ovn.kubernetes.io/tunnel_interface` on each node in the cluster.
- kubeovn takes care of updating the remote tunnel endpoints for vxlan in ovs bridges on each node to act as underlay.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8101
https://github.com/harvester/harvester/issues/7834


enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,413 @@
...
+Harvester cluster with 3 nodes with `mgmt` interface on vlan 2021 having the following IPs.
+- Node 1: `10.115.252.135/23`
+- Node 2: `10.115.252.136/23`
+- Node 3: `10.115.252.137/23`
+
+- Default cluster network `mgmt` spanning all 3 nodes.
+- Create cluster network `cluster-1` with networkconfig spanning all 3 nodes
+- Create cluster network `cluster-2` with networkconfig spanning all node 1 and node 2.
+
+#### Story 1: Create VM Network with cluster network `cluster-1`, vid `2012`,`L3 Connectivity` field set and mode `DHCP`
```
w13915984028: did you consider to define a new type of network like `l3-sub-network` to differentiate with VM Network, the `l3-sub-network` is dedicated for createing sub-interfaces on specific cluster-network, and set it's IP via dhcp/pool

`l3-sub-network` can't be used by VM;  it is not a NAD, but rather a local defined new CRD

it will not cross-affect with existing NAD, and also the following solutions like label/annotation to represent many concepts

this new `l3-sub-network` can then also be referred by OVN or third-party apps like storage

If a NAD defines with same VID with the `l3-sub-network`, then it means the VMs on this NAD can talk with `l3-sub-network` IPs, and can be further contrlled to allow or deny

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,413 @@
...
+
+- bridge vlan vid `2012` added to `cluster-1-br` on node 1,node 2 and node 3
+- vlan sub interface `cluster-1-br.2012` is created on node 1,node 2 and node 3 in the cluster and ip address is assigned from DHCP.
+
+#### Story 2: Create VM Network with cluster network `cluster-2`, vid `2013`,`L3 Connectivity` field set and mode `DHCP`
+
+- bridge vlan vid `2013` added to `cluster-1-br` on node 1,node 2
+- vlan sub interface `cluster-1-br.2013` is created on node 1,node 2 in the cluster and ip address is assigned from DHCP.
+
+#### Story 3: Update VM Network with cluster network `cluster-1`, vid `2012`,`L3 Connectivity` field set and mode `Static`
```
w13915984028: a sub-story:  if OVN is running on-top of this network, some checks are needed to block the operation until the uplayer is stopped

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,413 @@
...
+
+- bridge vlan vid `2012` removed from `cluster-1-br` on node 1,node 2 and node 3
+- vlan sub interface `cluster-1-br.2012` is removed from node 1,node 2 and node 3
+
+#### Story 6: Delete network config under cluster network `cluster-1`
+
+- Remove all vlan associated with `cluster-1` on node 1, node 2 and node 3
+- Remove all vlan sub interfaces associated with `cluster-1 from node 1,node 2 and node 3
+
+#### Story 7: New node node 4 added to the cluster
```
w13915984028: it assumes the underlay cluster-network spans to all nodes by default

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,413 @@
...
+
+- bridge vlan vid `2012` added to `cluster-1-br` on node 4
+- vlan sub interface `cluster-1-br.2012` is created on node 4 in the cluster and ip address is assigned from DHCP.
+
+#### Story 8: `Node 3` removed from network config of `cluster-1` or `Node 3` shutdown
```
w13915984028: >`or `Node 3` shutdown`

it does not trigger such operation

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,413 @@
...
+- bridge vlan vid `2012` added to `cluster-1-br` on node 4
+- vlan sub interface `cluster-1-br.2012` is created on node 4 in the cluster and ip address "192.168.1.12/24" is assigned to `cluster-1-br.2012`
+
+#### Story 10: Reboot node 1, node 2 and node 3
+
+After the nodes come up,
+- bridge vlan vid `2012` added to `cluster-1-br` on node 1,node 2 and node 3
+- vlan sub interface `cluster-1-br.2012` is created on node 1,node 2 and node 3 in the cluster and ip address is assigned from DHCP or static
+
+#### Story 11: Update node selector in network config under cluster network `cluster-1` to node-1
```
w13915984028: it depends on the upper layer, if OVN is sitting on this network, the change might break VMs

w13915984028 (commented):
Some initial feedbacks, thanks.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: Right!! Added a webhook case to the test plan.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: The underlay tunnel interface config resource will make sure the cluster network selected spans all nodes in the cluster and have no specific node selectors.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: Any config changes with node selectors on network config will be taken care by OnChange vlanconfig controller.

And, We have a handler for OnRemove() on node controller, does it not handle node shutdown ?

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: Yeah Right!!! Added a webhook validation to the testplan.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: Since we have all the required info and handling done by the cluster network agent, it will be easy to use the existing resources.Cluster Network agent is already adding bridge vlans to the cluster bond, we need additional steps like adding bridge vlan to cluster bridge and nmconnection file creation.

My thought is may be additional crd will be an overhead while it can be done with existing resources.

And VMs using the NAD will communicate with the other VMs over L2 in the usual way and if L3 connectivity is required for any vlan, the users specifies mode and a vlan interface is created on demand.

I am not sure if we have a use case where VMs want to communicate with vlan interface on the host.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
w13915984028: `Shut down` is a different concept than `node deletion`, do you mean the latter here?

rrajendran17: Oh ok, so shutdown will it make a node still present in cluster but not in ready state or something ? can this be handled on OnChange() of node controller ?

enhancements/20251015-support-ipconfig-on-clusternetworks.md
w13915984028: I have below considerations:

A. Recall those concepts:

The `VM Network` is mapped to `NAD`, which in turn guides the CNI to create  `interface/nic` for POD/VM, it is more a Layer 2 concept.

This HEP, targets to create addition `vlan sub interfance &/ (allocate) IPs`, it aims at those scenarios:
(1) The https://github.com/harvester/harvester/issues/7834 mentioned, adds fixed IP on `mgmt` or other cluster-network, then OVN can sit on top of it
(2) For third party to run applications like distributed storage.

The concept is different with `NAD` & `CNI`.

B. For such a `l3-sub-network`, it has things like VID, IP/dhcp/... listed in HEP,  put them on NAD's annotation/label will be difficult to manage, but a new CRD is more flexiable to define and extend.

C. The network-contoller has defined those deployments/daemonsets,  write new controller can highly utilize existing framework, but doesn't increase the complexity of NAD related handling.  NAD has strong checks upon VMs. The `l3-sub-network` has less relationship with VMs.

D. The OOB management (via wicked/network-manager) should be as less as possible

D.1 When user needs additional sub-interfaces on `mgmt` when installing cluster, those are done via `network-manager`, we call it OOB.  The ISO installation could add a new page/configuration to do it.

D.2 All other cases (sub-interface on `mgmt` after cluster is up;  sub-interface on secondary-cluster-network) can be taken care by Harvester controllers.

e.g. 
1. OVN sits on mgmt:
 1.1 install cluster
 1.2 create sub-interfaces after cluster is up
 1.3 enable ovn on top of it

2. OVN sits on cluster network cn2:
 2.1 install cluster
 2.2 create cluster-network cn2 (related cn2-br, cn2-bo are done by controller, not network-manager)
 2.3 create sub-interfaces after cn2 is ready  
 2.4 enable ovn on top of cn2

enhancements/20251015-support-ipconfig-on-clusternetworks.md
w13915984028: The `cluster-1-br.2012` is created on each node, when `node3` is down, everything on `node3` (network-controller agent, bridge, kubelet, ovn damenon, vswitch...) are gone.   The `down` might be temporary (e.g. node rebooting) or long-term (e.g. powerd off).

Normally, `network-controller` does not care about this event. The `vlanconfig` can still select `node3` if it select all, as `node3` CRD object is existing.

What's your concern here? worry about the IP address of `node3.cluster-1-br.2012.ip` is released by remote DHCP server?

Even when OVN is sitting on `*cluster-1-br.2012.ip`, it shows that one of the VTEP is down. This is very similar to the current OVN when one of the Harvester host node is down.

rrajendran17: Like a node loses connectivity or it is no more part of the cluster due to network reasons.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: > I have below considerations:
> 
> A. Recall those concepts:
> 
> The `VM Network` is mapped to `NAD`, which in turn guides the CNI to create `interface/nic` for POD/VM, it is more a Layer 2 concept.
> 
> This HEP, targets to create addition `vlan sub interfance &/ (allocate) IPs`, it aims at those scenarios: (1) The #7834 mentioned, adds fixed IP on `mgmt` or other cluster-network, then OVN can sit on top of it (2) For third party to run applications like distributed storage.
> 
> The concept is different with `NAD` & `CNI`.
> 
> B. For such a `l3-sub-network`, it has things like VID, IP/dhcp/... listed in HEP, put them on NAD's annotation/label will be difficult to manage, but a new CRD is more flexiable to define and extend.
> 
> C. The network-contoller has defined those deployments/daemonsets, write new controller can highly utilize existing framework, but doesn't increase the complexity of NAD related handling. NAD has strong checks upon VMs. The `l3-sub-network` has less relationship with VMs.
> 
> D. The OOB management (via wicked/network-manager) should be as less as possible
> 
> D.1 When user needs additional sub-interfaces on `mgmt` when installing cluster, those are done via `network-manager`, we call it OOB. The ISO installation could add a new page/configuration to do it.
> 
> D.2 All other cases (sub-interface on `mgmt` after cluster is up; sub-interface on secondary-cluster-network) can be taken care by Harvester controllers.
> 
> e.g.
> 
>     1. OVN sits on mgmt:
>        1.1 install cluster
>        1.2 create sub-interfaces after cluster is up
>        1.3 enable ovn on top of it
> 
>     2. OVN sits on cluster network cn2:
>        2.1 install cluster
>        2.2 create cluster-network cn2 (related cn2-br, cn2-bo are done by controller, not network-manager)
>        2.3 create sub-interfaces after cn2 is ready
>        2.4 enable ovn on top of cn2

Thanks @w13915984028 
C is a valid point, VM Network has strong connection to the VMs while l3-sub-network has less or no relation to the underlying VMs, but the problem is even l3-sub-network is strongly interconnected to the VM Network and the cluster network.Any changes to them will affect the l3-sub-network.
l3-sub-network in actual dealing with the same set of underlying network config as cluster networks crd doing.

We could allow L3 network config changes on VM Network with VMs attached to it.

I do not see bigger advantage of using one over the other (separate crd vs VM Network) except may be having a separate CRD would be more cleaner and we could use this as host Network instead of VM Network.

For D1, yes we could enhance user experience during installation to add multiple vlan interfaces to the mgmt-br.
May be we can track this during separate issue.

And ovn uses single underlay for the full cluster, either mgmt + vlan-id or user defined cluster + vlan-id and it must be created before VMs uses the overlay networks.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
w13915984028: It sounds like a more generic topic on k8s cluster, when a node lose its connectivity, and the `none-goal` seems to fit for it for now, to limit the range of HEP.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,416 @@
...
+## Motivation
+
+- Some data‑plane or storage networks must be reachable as routed L3 subnets (for example: dedicated storage network with static or dynamic IP addressing)
+
+- Operators may want to isolate management traffic from application traffic using different routed subnets on separate physical NICs (different gateways, different next‑hop routers, or different physical uplinks).
+
+- Cloud/edge environments sometimes require that a non‑management NIC has its own IP address on a different L3 network to integrate with existing routing/OSPF/BGP or external services.
+
+Harvester currently configures a single VLAN sub-interface on the management NIC by default during installation. However, users need the ability to configure multiple VLAN sub-interfaces on both the management NIC and other user defined cluster networks with full L3 support.These configurations must be automated and persist across reboots and upgrades. Manual setup today is error-prone and not sustainable for production use.
+
```
mingshuoqiu: Consider adding:
1. “Persistent network configuration is required to survive host reboots and upgrades.”
2. “Manual creation of VLAN sub-interfaces today is inconsistent across nodes.”

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,416 @@
...
+[Unit]
+Description=Dispatch wicked ifreload based on ifcfg-* changes
+
+[Service]
+Type=oneshot
+ExecStart=/tmp/nm-reload-dispatch.sh
+```
+
+3.Create a reload script which executes `nmcli connection reload` from the dispatch service
+
```
mingshuoqiu: Maybe we should consider Multiple NADs created simultaneously will trigger one reload per file?
Use `sleep x` as debounce before `nmcli connection reload`

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,416 @@
...
+#### Story 8: `Node 3` removed from network config of `cluster-1` or `Node 3` shutdown
+
+- bridge vlan vid `2012` removed from `cluster-1-br` on node 3
+- vlan sub interface `cluster-1-br.2012` is removed on node 3
+
+#### Story 9: New node node 4 added to the cluster (L3 connectivity mode `static`)
+
+- User reconfigures VM Network `2012` with node IP "192.168.1.13/24" for node 4
+- bridge vlan vid `2012` added to `cluster-1-br` on node 4
+- vlan sub interface `cluster-1-br.2012` is created on node 4 in the cluster and ip address "192.168.1.12/24" is assigned to `cluster-1-br.2012`
```
starbops: May I know what happens when the node's IP address is not given for a newly joined node?

starbops (commented):
Do we have a way to ensure that there are no default route conflicts when the mode is DHCP? The cluster will break if there are multiple default routes exist in the nodes.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,416 @@
...
+3.Create a reload script which executes `nmcli connection reload` from the dispatch service
+
+```
+#!/bin/bash
+
+SERVICE_NAME="nm-path-reload"
+
+logger -t "$SERVICE_NAME" "Change detected in`L3 Connecitivy Required` system-connections — reloading NetworkManager connections"
+
+nmcli connection reload
```
starbops: To make sure: will this command introduce any network interruption?

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,416 @@
...
+  namespace: default
+spec:
+  config: >-
+    {"cniVersion":"0.3.1","name":"vlan2012","type":"bridge","bridge":"cluster-1-br","promiscMode":true,"vlan":2012,"ipam":{}}
+```
+
+- Harvester network controller handles the nad change and updates the annotations of the cluster network with the following.
+  - VLAN-ID (2 - 4094)
+  - Mode (Static or DHCP)
+  - List of nodenames to IP address mapping (if Mode = Static)
```
starbops: Based on my understanding, is it correct that the only thing that matters is the ClusterNetwork but not NAD? It looks like we chose NAD as the entry point for this feature because it contains VLAN information. So technically, the user can bypass updating the NAD and instead annotate the ClusterNetwork with sufficient information (mode, VLAN ID, etc), and it still works as expected. Please correct if I misunderstood.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,416 @@
...
+
+- Supporting L3 Connectivity on VM Network with type `L2UntaggedNetwork`,`OverlayNetwork` and `L2VlanTrunkNetwork`.
+- Supporting IPPools or external IPAM integration for IP allocation to the vlan sub interfaces on Harvester hosts.
+- Support IPv6 addresses on the vlan sub interfaces of Harvester Hosts.
+
+## Proposal
+
+- Allow users to configure/unconfigure Layer 3 network Information on the VM Network resource.
+- Harvester network controller NAD handles the change and updates the annotations of the cluster network resource.
+- Harvester network controller cluster network agent handles the bridge vlan and creation/update/deletion of nmconnection profiles on the host.
```
starbops: What's the reason behind the decision to let NetworkManager, but not the network controller, manage bridge VLANs and sub-interfaces? We used to do this ourselves entirely only except for the `mgmt` network.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
starbops: I have a similar concern of putting these in the NAD annotations. It feels semantically weird to configure a "VM Network" to manipulate "Host Network". Maybe "Network Config (VlanConfig)" is a better place to do this? I'm unsure. With that said, I understand it's for the VLAN ID information stored in the NAD.

starbops (commented):
Thank you for the well-written HEPs. This also unlocks an use case for VM load balancers on secondary networks at some degree. #7500

We should also consider the security implications: having more externally accessible IP addresses on the nodes increases their exposure to potential access by others. In my opinion, having documentation that explains this should be the baseline requirement.

rrajendran17:
> Do we have a way to ensure that there are no default route conflicts when the mode is DHCP? The cluster will break if there are multiple default routes exist in the nodes.

sorry I did not get your question right during our last discussion.
I think I understand now that you are talking about the "DHCLIENT_SET_DEFAULT_ROUTE='yes'" if set by user when using DHCP mode.

Wicked:
Wicked by default ignores the default route from newer clients if there is already a default route installed.

default route from dhcp server:
```<lease>
  <family>ipv4</family>
  <type>dhcp</type>
  <uuid>cc3b0a69-b108-0800-b208-000017000000</uuid>
  <state>granted</state>
  <acquired>1762284446</acquired>
  <update>0x000037ff</update>
  <ipv4:dhcp>
    <client-id>ff:15:30:46:80:00:01:00:01:2f:6f:d1:2c:8c:dc:d4:b5:ec:08</client-id>
    <server-id>10.115.8.1</server-id>
    <sender-hw-address>f6:a6:be:01:bb:21</sender-hw-address>
    <lease-time>600</lease-time>
    <renewal-time>300</renewal-time>
    <rebind-time>525</rebind-time>
    <address>10.115.14.243</address>
    <netmask>255.255.248.0</netmask>
    <boot>
      <server-address>10.115.8.1</server-address>
    </boot>
    <routes>
      <route>
        <nexthop>
          <gateway>10.115.15.254</gateway>
        </nexthop>
      </route>
    </routes>
    <dns>
      <domain>harvester.rancherlabs.com</domain>
      <server>10.115.8.1</server>
      <server>8.8.8.8</server>
      <search>harvester.rancherlabs.com</search>
    </dns>
  </ipv4:dhcp>
</lease>

```
dhclient uses only one default route as mentioned in /etc/sysconfig/network/dhcp and I did not see any additional default route installed with option DHCLIENT_SET_DEFAULT_ROUTE set to yes.
```
# When multiple copies of dhcp client run, it would make sense that only
# one of them does it. 
# 
DHCLIENT_SET_DEFAULT_ROUTE="yes"
```

Network Manager:
Network Manager if set with the following option for using auto(DHCP) and nerver-default as "false" (install default route), it installs another default route but with higher metric than the one already installed.
This will make sure that always the default route with lower priority over mgmt interface configured during harvester installer is chosen.

```
[ipv4]
dns=8.8.8.8;
method=auto
never-default=false
```

```
ip addr show mgmt-br.2021
9: mgmt-br.2021@mgmt-br: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 8c:dc:d4:b5:7a:04 brd ff:ff:ff:ff:ff:ff
    inet 10.115.252.137/23 brd 10.115.253.255 scope global dynamic noprefixroute mgmt-br.2021
       valid_lft 51653sec preferred_lft 51653sec
    inet 10.115.252.5/32 scope global mgmt-br.2021
       valid_lft forever preferred_lft forever
hp-65:/etc/NetworkManager/system-connections # ip addr show rrrr-br.2012
183: rrrr-br.2012@rrrr-br: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether ea:50:43:b0:fd:44 brd ff:ff:ff:ff:ff:ff
    inet 10.115.14.246/21 brd 10.115.15.255 scope global dynamic noprefixroute rrrr-br.2012
       valid_lft 331sec preferred_lft 331sec
    inet6 fe80::baee:4fdf:f127:583e/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever

ip route show
default via 10.115.253.254 dev mgmt-br.2021 proto dhcp src 10.115.252.137 metric 400 
default via 10.115.15.254 dev rrrr-br.2012 proto dhcp src 10.115.14.246 metric 401 

ip route get 8.8.8.8
8.8.8.8 via 10.115.253.254 dev mgmt-br.2021 src 10.115.252.137 uid 0 
    cache 
```

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: The node handler in network controller should take care of updating the annotations on the cluster network with mode: empty (or DHCP as default). The cluster network change should not handle any updates to this changes if mode is empty.
We need to document that user has update the VM Network with new node and its ip if mode is static for it to configure nmconnection profile on the new node joined the cluster.

rrajendran17: I believe it will be users` responsibility to take care of reconfiguration in case of node changes if the ip address retrieval mode is static

rrajendran17:
> Thank you for the well-written HEPs. This also unlocks an use case for VM load balancers on secondary networks at some degree. #7500
> 
> We should also consider the security implications: having more externally accessible IP addresses on the nodes increases their exposure to potential access by others. In my opinion, having documentation that explains this should be the baseline requirement.

We could also check if we could impose some security rules in kernel to protect them.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: nmcli connection reload will act only on connection which needs change.The existing active connections are not disturbed using this command.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: Harvester Network controller is taking care of bridge vlan addition to cluster network bridge in the usual way.But the interface creation and ip address retrieval using DHCP and ip address assignment is taken care by Network Manager.

Since network manager can maintain the connection profile and already taking care of the logic of creating the interface, getting address from dhcp server and assigning ip address to the interface it is much simpler if network controller could delegate this to network manager.

If done in network controller, then network controller needs to have a dhclient job initiating ipaddress request per host for every vlan and use ip commands to create interface and set ip address on the host.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: you are right, that still works.
cluster network change itself cannot configure anything unless we have vlanconfig changes calls "wakeUpClusterNetwork" to invoke a cluster network change on each node.We need to also make sure that the VM Network is created for the vlan id specified.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: I do not have a strong opinion about the CRD to configure this.
We could still do this with cluster network or Network Config under cluster network with vlan-id specified.
The webhook must make sure to validate if VM Network exists for the vlan-id configured.

rrajendran17:
I am summarizing some of the key discussions of this review and Please let me know if we can conclude on the design part or any other concerns.

1.Default route installed over the secondary vlan sub interface if DHCLIENT_SET_DEFAULT_ROUTE='yes' in the nmconnection
   - Using Network Manager any default route installed from DHCP over additional vlan sub interfaces will be installed with priority 
     greater than the existing default route over mgmt interface installed during harvester installation.So default route over mgmt 
     interface is not disturbed and all traffic go over this until it is overwritten manually or with route policies.

2.CRD for the L3 connectivity configuration.One of the below options.
   - VM Network 
   - Network Config(vlanconfig CRD) under cluster network
   - New CRD for configuration

Since vlanconfig CRD is closer to cluster network configuration and it holds the host related config we can use vlan config CR instead of NAD CR.

3.Network controller creating the nmconnection profiles on host instead of handling this inside network controller code.
   - since nmconnection profile automatically takes care of interface creation, especially DHCP and ip address assignment, it     
     becomes simpler to delegate this to NM.
 
  cc: @w13915984028 @starbops @mingshuoqiu

w13915984028:
@rrajendran17 

> 2 VlanConfig CRD

The `VlanConfig` can be 1:N or 1:1 or n:m mapping to nodes, it depends on users' configuration, and there is another object `VlanStatus` which is created by controller automatically to get a 1:1 mapping with node object to show the actual status of vlanconfig on each node.

This new L3-subnetwork, can be 1:N to utilize DHCP or 1:1 to set via static IP; it might be a different mode comparing to `VlanConfig`. 

I would prefer to add a new CRD to carry this information, to avoid breaking any existing objects and logics. User can either create one object (for [secondnetwork-vid] pair) to set dhcp mode; or multi objects to set 1:1 static IPs, the controller and webhook have clear rule to work. Any configuration/dhcp/controller error, do not affect the basic cluster-network/vlanconfig.

BTW, changes on `VlanConfig` might trigger the `stop vm to continue` webhook check.


> 3.Network controller creating the nmconnection profiles on host instead of handling this inside network controller code.

Besides the mgmt and related NICs, all additional cluster networks are created by network-controller until network-controller POD is up and fetches related configuration information (uplink NICs, bond params, MTU, vid set ...) from api-server;  when `nmconnection profile` is auto taken care by network-manager, could there be race cases like `cn2-br`, `cn2-bo` interfaces are not created/up but nm tries to fetch and set ip? It also means the network-controller needs to convert CRD objects to nmconnection file, and continuously check & revert the differences between them.  Plus, controller nees a way to get the failure information from network-manager and reconcile properly when error happens. Network-controller already has the feature to do DHCP on VM-network. Do as more tasks as possible on the controller side, could have more benefits.   thanks.

rrajendran17:
> @rrajendran17
> 
> > 2 VlanConfig CRD
> 
> The `VlanConfig` can be 1:N or 1:1 or n:m mapping to nodes, it depends on users' configuration, and there is another object `VlanStatus` which is created by controller automatically to get a 1:1 mapping with node object to show the actual status of vlanconfig on each node.
> 
> This new L3-subnetwork, can be 1:N to utilize DHCP or 1:1 to set via static IP; it might be a different mode comparing to `VlanConfig`.
> 
> I would prefer to add a new CRD to carry this information, to avoid breaking any existing objects and logics. User can either create one object (for [secondnetwork-vid] pair) to set dhcp mode; or multi objects to set 1:1 static IPs, the controller and webhook have clear rule to work. Any configuration/dhcp/controller error, do not affect the basic cluster-network/vlanconfig.
> 
> BTW, changes on `VlanConfig` might trigger the `stop vm to continue` webhook check.
> 
> > 3.Network controller creating the nmconnection profiles on host instead of handling this inside network controller code.
> 
> Besides the mgmt and related NICs, all additional cluster networks are created by network-controller until network-controller POD is up and fetches related configuration information (uplink NICs, bond params, MTU, vid set ...) from api-server; when `nmconnection profile` is auto taken care by network-manager, could there be race cases like `cn2-br`, `cn2-bo` interfaces are not created/up but nm tries to fetch and set ip? It also means the network-controller needs to convert CRD objects to nmconnection file, and continuously check & revert the differences between them. Plus, controller nees a way to get the failure information from network-manager and reconcile properly when error happens. Network-controller already has the feature to do DHCP on VM-network. Do as more tasks as possible on the controller side, could have more benefits. thanks.

Thanks @w13915984028 

"Any configuration/dhcp/controller error, do not affect the basic cluster-network/vlanconfig."
This makes more sense and if having a new CRD will be more cleaner then we can go ahead with a new CRD approach.I am thinking more like the following having one object per (clusternetwork + vlan-id), something like below for example

```apiVersion: network.harvesterhci.io/v1beta1
kind: HostNetworkConfig
metadata:
  annotations:
    network.harvesterhci.io/clusternetwork: cluster-1
    network.harvesterhci.io/vlan-id: "2012"
    network.harvesterhci.io/mode: "static"
    network.harvesterhci.io/ips: |
      {
        "node1": "192.168.1.10/24",
        "node2": "192.168.1.11/24",
        "node3": "192.168.1.12/24"
      }
  name: hostnetworkconfig1

apiVersion: network.harvesterhci.io/v1beta1
kind: HostNetworkConfig
metadata:
  annotations:
    network.harvesterhci.io/clusternetwork: cluster-1
    network.harvesterhci.io/vlan-id: "2013"
    network.harvesterhci.io/mode: "dhcp"
  name: hostnetworkconfig2
```
Also in case of node updates (new node addition), user has to be explicitly add the node name and mode and static ip to the CRD.

And regarding network manager, the idea as mentioned in the HEP is to act on nm connection profiles only after it is created by the network controller and network controller makes sure to create these files only if the cluster network uplink exists.
But regarding any error from network manager, currently we do not have a way to track this.And it becomes cumbersome to get a status from network manager and update network controller.
May be initiating a DHCP within the controller could be a right option, could be done synchronously and interface could be updated with the ip received from dhcp server using netlink package.

Yu-Jack:
How about moving those annotations to the spec fields? Because annotations doesn't have definition generally, it makes us hard to manage them. Also, you can reduce parsing effort in the codebase. 

```yaml
apiVersion: network.harvesterhci.io/v1beta1
kind: HostNetworkConfig
metadata:
  name: hostnetworkconfig1
spec:
  clusterNetwork: cluster-1
  vlanID: 2012
  mode: static
  ips:
    node1: 192.168.1.10/24
    node2: 192.168.1.11/24
    node3: 192.168.1.12/24
---
apiVersion: network.harvesterhci.io/v1beta1
kind: HostNetworkConfig
metadata:
  name: hostnetworkconfig2
spec:
  clusterNetwork: cluster-1
  vlanID: 2013
  mode: dhcp
```

w13915984028 (dismissed):
LGTM, thanks.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
```diff
@@ -0,0 +1,235 @@
...
+### Test plan
+
+- Webhook should reject host network config if `cluster network` is not created or not in `ready` state.
+- Create/Update HostNetworkConfig resource with vlan-id,mode,ips should create a vlan interface and assign IP address on Harvester hosts.
+- Remove vlanconfig on the cluster network should remove all bridge vlan and vlan interfaces associated with the cluster network on Harvester hosts selected by the network config.
+- Update vlanconfig to specific node selectors should take care adding/removing bridge vlan and vlan interfaces on required Harvester hosts.
+- Remove HostNetworkConfig  should remove bridge vlan from the cluster network bridge and remove the vlan interface on Harvester hosts.
+- Adding new node to the cluster (if network config spans all node in cluster and mode is `DHCP`) should add bridge vlan and vlan interfaces on the new Harvester host.
+- Removing node from the cluster (if network config spans all node in cluster) should remove bridge vlan and vlan interfaces on Harvester host being removed.
+- Webhook should reject Host Network config with vlan-id `0` or `1`.
```
ibrokethecloud: what would be the scenario which we are trying to address here? say mgmt interface is running on a tagged vlan.. and we also want a second host network on untagged vlan will it cause issues?

ibrokethecloud:
after reviewing the PR and the HEP, would it not have been better if the `HostNetworkConfig` object was 1-to-1 mapping for underlying node.

```
type HostNetworkConfig struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
	// +kubebuilder:validation:XValidation:rule="self.mode != 'static' || (has(self.ips) && size(self.ips) > 0)",message="spec.ips must be specified and non-empty when mode is static"
	Spec []HostNetworkConfigSpec `json:"spec"`
	// +optional
	Status HostNetworkConfigStatus `json:"status"`
}

// +kubebuilder:validation:MaxLength=50
// +kubebuilder:validation:XValidation:rule="isCIDR(self)",message="Invalid CIDR format"
type IPAddr string

type HostNetworkConfigSpec struct {
	// +optional

	// Required, non-empty
	// +kubebuilder:validation:MinLength=1
	ClusterNetwork string `json:"clusterNetwork"`

	// +optional
	Underlay bool `json:"underlay"`

	// Required, must be 1-4094
	// +kubebuilder:validation:Minimum=1
	// +kubebuilder:validation:Maximum=4094
	VlanID uint16 `json:"vlanID"`

	// Required, non-empty, only 'static' or 'dhcp'
	// +kubebuilder:validation:Enum=static;dhcp
	// +kubebuilder:validation:MinLength=1
	Mode string `json:"mode"`

	// validated if mode=static
	// +optional
	// +kubebuilder:validation:MaxProperties=500
	HostIPs IPAddr`json:"ips,omitempty"`
}
```

The spec would be an array of `HostNetworkConfigSpec`. This way we would just append multiple `HostNetworkConfigSpec` entries to the same `HostNetworkConfig` object.

The network controller could init an empty object for each node in the cluster, and user could just customise the entries in the `HostNetworkConfigSpec` to meet their multiple cluster network / vlan id requirements.

The reconcile would be easy, since agent would only check the object matching its underlying hostname.

It would also make validation easier since you can only have 1 HostNetworkConfig object per node, so there is less chance of duplicates across the cluster pointing to same cluster network / vlan id combination.

We will probably need nothing in the `manager` as the agent could just check if underlying `clusternetwork` is ready before applying changes.

The agent could continuously reconcile if spec matched the bridge interface definitions on host, and add / remove interface definitions as needed.

enhancements/20251015-support-ipconfig-on-clusternetworks.md
rrajendran17: vlan 1 and untagged vlan share the same subnet in Harvester and both are egress untagged.So idea is to allow only vlan 1 to 4094 as sub interfaces on any cluster network br including mgmt-br.Webhook ideally should reject just the vlan id as 0. I will update the doc.

But this also reminds me of a case what if mgmt-br with mgmt ip is on untagged network.Then should users not be allowed to configure a sub interface with vlan id 1 ?

rrajendran17:
@w13915984028 @ibrokethecloud I have updated the HEP considering all the comments and updates from the PR.
This HEP is up to date.Please check. We can merge this once you approve. Thanks

w13915984028 (approved):
LGTM, thanks.

Issue #9340, opened 2025-10-21
[DOC] Minor updates upon resource quota and VM memory overhead


w13915984028:
### Document request

Highlight that when adjusting the RQ, take overhead into account.

w13915984028:
Both PRs are merged, close as finished. thanks.

PR #9345, opened 2025-10-21
Minor update on overhead-ratio and RQ processing

w13915984028:
#### Problem:

Minor update on overhead-ratio and RQ processing

1. Return the default value in abnormal case
2. Add test case to show how overhea-ratio affects VM with RQ

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9340

#### Test plan:

**Not-required.**

doc PR: https://github.com/harvester/docs/pull/903

### DIFF
--- a/pkg/util/resourcequota/calculator_test.go
+++ b/pkg/util/resourcequota/calculator_test.go
@@ -12,6 +12,8 @@ import (
 	corefake "k8s.io/client-go/kubernetes/fake"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/generated/clientset/versioned/fake"
 	"github.com/harvester/harvester/pkg/util"
 	"github.com/harvester/harvester/pkg/util/fakeclients"
 )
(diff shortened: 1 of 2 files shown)


martindekov:
Is this ready for review? Seems pipeline is failing and builds are canceled

w13915984028:
@martindekov It is ready for review, the CI is just slow and failed from time to time.

pkg/util/resourcequota/calculator_test.go
```diff
@@ -464,3 +466,203 @@ func Test_vmValidator_containsEnoughResourceQuotaToStartVM(t *testing.T) {
...
+		name          string
+		vm            *kubevirtv1.VirtualMachine
+		nsrq          *v3.NamespaceResourceQuota
+		rq            *corev1.ResourceQuota
+		overheadRatio string
+		wantErr       bool
+	}{
+		{
+			name: "VM can start as quota is enough with default overhead-ratio 1.5",
+			vm: &kubevirtv1.VirtualMachine{
```
martindekov, line 499: I'd extract the long VM creation in function similar to what you did with generateRQ to reduce the strain on going through the object and the lines of code for the test

pkg/util/resourcequota/calculator_test.go
```diff
@@ -464,3 +466,203 @@ func Test_vmValidator_containsEnoughResourceQuotaToStartVM(t *testing.T) {
...
+						Spec: kubevirtv1.VirtualMachineInstanceSpec{
+							Domain: kubevirtv1.DomainSpec{
+								Resources: generateRQ(4, 4), // 4core, 4Gi
+							},
+						},
+					},
+				},
+			},
+			nsrq: &nsrq,
+			rq: &corev1.ResourceQuota{
```
martindekov, line 515: Similar to the VM suggestion we can extract and put as parameters to function the things that change in the spec, reducing the strain on understanding the test, reading function header is easier than going through object body IMHO

martindekov (approved):
LGTM Jian, thanks for the work, will approve, left comments as they are related with test refactoring, the core of the fix I think is already there and the comments left are optional

pkg/util/resourcequota/calculator_test.go
w13915984028: it is a balance of time and effort, and will be a good entry for AI to assist us optimizing such code :)

Issue #9338, opened 2025-10-20
[ENHANCEMENT] Replace klog with logrus on network-controller-harvester


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Most of Harvester repos are using logrus, which has better formatting and dynamic log level adjustment.

Network-controller-harvester is using both of them.

### Describe the solution you'd like

Replace klog with logrus.

Support change loglevel via deployment env.

**QA validation** is skipped as this is a more dev related change and can be validated by dev directly.

w13915984028:
Code PR is merged, dev tests are all passed, close.

PR #9355, opened 2025-10-22
Fix codegen lint error

w13915984028:
#### Problem:

In PR https://github.com/harvester/network-controller-harvester/pull/217, it copied code from Harvester code gen, and following errors are encountered

```

first:

pkg/codegen/main.go:77:11: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	if err = ioutil.WriteFile(absPath, output, 0644); err != nil {
	         ^
pkg/codegen/main.go:5:2: SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package [io] or package [os], and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
	"io/ioutil"
	^
2 issues:
* gosec: 1

then:

Running validation
Running: golangci-lint
pkg/codegen/main.go:78:11: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	if err = os.WriteFile(absPath, output, 0644); err != nil {
	         ^
1 issues:
* gosec: 1
FATA[0118] exit status 1                                
make: *** [Makefile:11: ci] Error 1


```

#### Solution:

Fix

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9359
https://github.com/harvester/harvester/issues/9338

#### Test plan:

Both `make generate` and `make ci` are successful.

QA validation is not required.

### DIFF
--- a/pkg/codegen/main.go
+++ b/pkg/codegen/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"bytes"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


ibrokethecloud (approved):
lgtm. thanks for the fix.

PR #9361, opened 2025-10-23
Ignore resource version changes

ibrokethecloud:
#### Problem:

VM Hot plug operations result in VM objects generation vs observed generation being no longer in sync.

As a result when the steve executes wrangler summarizers, it finds a discrepancy between the objects generation and observedGeneration fields and generates a warning in the steve state.

Though the message is benign, it just results in a poor UX, since it may confuse users about the health of the VM or if a restart is needed.

#### Solution:

The PR introduces a customizer which patches the vm state to clear the state of generation related warning messages.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9629
#### Test plan:

To test we need a minimal 2 node cluster to ensure VM can be migrated to verify hot plug operations:
* Install a build of harvester which contains this change.
* Run a VM
* Once VM is running, hot plug cpu/disk/nic to this VM
* The VM should successfully migrate without any `observed generation` messages in the UI

### DIFF
--- a/pkg/server/customizers/vm_test.go
+++ b/pkg/server/customizers/vm_test.go
@@ -1,0 +1,18 @@
+package customizers
+
+import (
+	"testing"
+
+	"github.com/rancher/apiserver/pkg/types"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func Test_DropRevisionStateIfNeededWithEmptyState(t *testing.T) {
+	resource := &types.RawResource{
+		APIObject: types.APIObject{
+			Object: &corev1.ConfigMap{},
+		},
+	}
+
+	DropRevisionStateIfNeeded(nil, resource)
+}
--- a/pkg/server/customizers/vm.go
+++ b/pkg/server/customizers/vm.go
@@ -1,0 +1,42 @@
+package customizers
+
+import (
+	"strings"
+
+	"github.com/rancher/apiserver/pkg/types"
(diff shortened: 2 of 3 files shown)


ihcsim (dismissed):

Vicente-Cheng (approved):

pkg/server/customizers/vm_test.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func Test_DropRevisionStateIfNeededWithEmptyState(t *testing.T) {
+	resource := &types.RawResource{
+		APIObject: types.APIObject{
+			Object: &corev1.ConfigMap{},
+		},
+	}
+
+	DropRevisionStateIfNeeded(nil, resource)
```
ihcsim, line 17: this test looks incomplete. should it assert the `state` within `resource`?

ibrokethecloud: there is no assertion because its an empty configmap. The idea was to ensure we catch any change in behaviour of core wrangler summarizers or data lookups if object is empty. This will hopefully fail if something changes and will avoid nil pointer errors and pods crashing because of the same.

ihcsim (approved):

Yu-Jack (approved):

wheatdog (approved):

ibrokethecloud:
@mergify backport v1.8

ibrokethecloud:
@mergify backport v1.7

Issue #8636, opened 2025-07-10
[BUG] Uploading images via CDI sometimes just stops / gets stuck / doesn't start properly


tserong:
### Describe the Bug

Sometimes, when uploading an image file to external storage via CDI, the upload will never start correctly - it just gets stuck on "Uploading 0%" as shown for the second image in the screenshot below (opensuse-leap-15.5-net-x86_64-build491.1-media.iso )

<img width="1078" height="534" alt="Image" src="https://github.com/user-attachments/assets/22139a5e-b7d4-4421-89d9-885ec0d64e3b" />

In my testing this seems to only happen after some other upload has failed. In the above screenshot I first uploaded two ISOs and a qcow2 image, than I uploaded ext4.raw, which failed for some reason (I don't know why). After that ext4.raw upload failed, any subsequent uploads just get stuck, until I run `kubectl -n harvester-system rollout restart deployment/cdi-uploadproxy` which seems to get things going again.

Also, ext4.raw really should have uploaded successfully, so there's probably actually two issues here.

### To Reproduce

- Configure Longhorn V2, so there's some external storage for CDI to work with.
- Verify everything works as normal by uploading an ISO and a qcow2 image, ensuring you use an LHv2 storage class
- Upload something that will break. This is probably tricky, but in my case I was able to create a raw image which breaks the upload by running `qemu-img create ext4.raw 100M ; /usr/sbin/mkfs.ext4 ext4.raw`
- After that upload breaks, try uploading a previously working ISO or qcow2 image. It should get stuck.
- Run `kubectl -n harvester-system rollout restart deployment/cdi-uploadproxy`
- Try uploading an ISO or qcow2 again. It should work now.

### Expected Behavior

Uploads should always either succeed, or, if they're going to fail, they should fail cleanly with some sort of suitable error message.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.5.0 and v1.6.0-dev-20250703
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Virtualized single node 
- Rancher version:

jmcshane:
I have been testing this bug further and have additional diagnostics. From the cdi-uploadproxy deployment, this error coincides with the following log:

```
[...]

PR #9362, opened 2025-10-23
Fix concurrency between cdi and upload components

yulken:
#### Problem:

This PR has two purposes:
Improve error logging in the CDI image upload flow.
Address concurrency issues between the cdi.go and upload.go components.

#### Solution:

Checking if `sourceType` is `upload` when the DataVolume is not found on `cdi.go`. This is to prevent instant retries from causing conflict with the functions being executed in `upload.go`. If that's the case, retry with a short delay.
#### Related Issue(s):

Issue #8636

### DIFF
--- a/pkg/image/cdi/cdi.go
+++ b/pkg/image/cdi/cdi.go
@@ -65,11 +65,7 @@ func (b *Backend) Check(vmImg *harvesterv1.VirtualMachineImage) error {
 	targetDVName := b.vmio.GetName(vmImg)
 	targetDV, err := b.dataVolumeClient.Get(targetDVNs, targetDVName, metav1.GetOptions{})
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			logrus.Infof("DataVolume %s/%s not found, waiting for the initialization", targetDVNs, targetDVName)
-			return err
-		}
-		return fmt.Errorf("failed to get DataVolume %s/%s: %v", targetDVNs, targetDVName, err)
+		return b.handleDataVolumeError(err, vmImg, targetDVNs, targetDVName)
 	}
 
 	// upload source type will update the progress on the upload handler
@@ -100,6 +96,17 @@ func (b *Backend) Check(vmImg *harvesterv1.VirtualMachineImage) error {
 	}
 
 	return nil
+}
+
+func (b *Backend) handleDataVolumeError(err error, vmImg *harvesterv1.VirtualMachineImage, targetDVNs, targetDVName string) error {
+	if apierrors.IsNotFound(err) {
+		if vmImg.Spec.SourceType == harvesterv1.VirtualMachineImageSourceTypeUpload {
+			return common.ErrRetryLater
+		}
+		logrus.Infof("DataVolume %s/%s not found, waiting for initialization", targetDVNs, targetDVName)
+		return err
+	}
+	return fmt.Errorf("failed to get DataVolume %s/%s: %v", targetDVNs, targetDVName, err)
 }
 
 func (b *Backend) UpdateVirtualSize(vmi *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
--- a/pkg/api/image/formatter.go
+++ b/pkg/api/image/formatter.go
@@ -8,6 +8,7 @@ import (
 	"github.com/rancher/apiserver/pkg/apierror"
 	"github.com/rancher/apiserver/pkg/types"
 	"github.com/rancher/wrangler/v3/pkg/schemas/validation"
+	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


votdev (dismissed):
LGTM

pkg/image/cdi/cdi.go
```diff
@@ -66,6 +66,9 @@ func (b *Backend) Check(vmImg *harvesterv1.VirtualMachineImage) error {
 	targetDV, err := b.dataVolumeClient.Get(targetDVNs, targetDVName, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			if vmImg.Spec.SourceType == harvesterv1.VirtualMachineImageSourceTypeUpload {
+				return common.ErrRetryLater
+			}
```
WebberHuang1118: Suggest factoring out this part to reduce nested conditions.

yulken: @WebberHuang1118  I'm not sure I understand. By "factoring out," do you mean moving this if statement outside of the parent condition? 
```go
       if vmImg.Spec.SourceType == harvesterv1.VirtualMachineImageSourceTypeUpload {
				return common.ErrRetryLater
		}
			logrus.Infof("DataVolume %s/%s not found, waiting for the initialization", targetDVNs, targetDVName)
			return err
		}
``` 
If so, that affects the behavior. Or perhaps you mean creating another condition, like:
```go
       if apierrors.IsNotFound(err) && vmImg.Spec.SourceType == harvesterv1.VirtualMachineImageSourceTypeUpload {
				return common.ErrRetryLater
		}
			logrus.Infof("DataVolume %s/%s not found, waiting for the initialization", targetDVNs, targetDVName)
			return err
		}
``` 
Please let me know

WebberHuang1118: No, I mean when `apierrors.IsNotFound(err)` occurs, we should wrap the succeeding steps into a utility function rather than keeping the current design:
 
```
	if err != nil {
			if vmImg.Spec.SourceType == harvesterv1.VirtualMachineImageSourceTypeUpload {
				return common.ErrRetryLater
			}Expand commentComment on lines R69 to R71Code has comments. Press enter to view.
			logrus.Infof("DataVolume %s/%s not found, waiting for the initialization", targetDVNs, targetDVName)
			return err
		}
```

There are three nested `if` statements, which hurts readability and will make future maintenance harder if we need to add more conditions.

WebberHuang1118 (approved):

pkg/image/cdi/cdi.go
```diff
@@ -102,6 +98,17 @@ func (b *Backend) Check(vmImg *harvesterv1.VirtualMachineImage) error {
 	return nil
 }
 
+func (b *Backend) handleDataVolumeError(err error, vmImg *harvesterv1.VirtualMachineImage, targetDVNs, targetDVName string) error {
+	if apierrors.IsNotFound(err) {
+		if vmImg.Spec.SourceType == harvesterv1.VirtualMachineImageSourceTypeUpload {
+			return common.ErrRetryLater
```
brandboat, line 104: After this change, we can no longer create a VM from an uploaded VM Image. In the UI, VM creation fails with: `disk-0: "Image" is required`.

This is because the UI validator treats a VM Image as invalid when `.status.storageClassName` is  undefined[^1].

Now `.status.storageClassName` never gets populated because we return ErrRetryLater and the uploaded VM Image never reaches the Initialized status[^2]. As a result, `.status.storageClassName` remains unset[^3] and the UI considers the image invalid.

c.c. @Vicente-Cheng, @WebberHuang1118. Perhaps we should figure other ways to deal with the concurrency update issue.

[^1]: https://github.com/harvester/harvester-ui-extension/blob/ad9fef63c093986f03cf2f560fb28ae006aaf76b/pkg/harvester/validators/vm.js#L97-L102 
[^2]: https://github.com/harvester/harvester/blob/dedd291efd8c659a9f05e13b62b615477e708993/pkg/image/common/operator.go#L310  
[^3]: https://github.com/harvester/harvester/blob/dedd291efd8c659a9f05e13b62b615477e708993/pkg/controller/master/image/vm_image_controller.go#L86-L92

WebberHuang1118: @yulken @Vicente-Cheng 

This PR introduces a regression that prevents third-party VM images from being used to start a VM, which has a noticeable impact on users.

I revisited the PR, and it doesn’t appear to address the scenario described in the original issue. In my opinion, we may need to revert this PR.

yulken: We can revert this, I'll try developing a better solution as soon as I'm available.
Thanks for detecting this @brandboat

brandboat: I think it's fine, no need to revert the commit as this PR also includes some useful debug log, regarding the fix, it's covered in https://github.com/harvester/harvester/pull/9868

Issue #9325, opened 2025-10-16
[BUG] Skip witness node when considering minallocatableipaddr count for LH instance manager pods


rrajendran17:
### Describe the Bug

https://github.com/harvester/harvester/blob/c537e12b7f09a02e315220ba4605bc973d31c3ea/pkg/controller/master/storagenetwork/storage_network.go#L373-L394 

The above code validates if whereabouts has allocated atleast the min addresses required.
Example, the count of instance manager pods in the cluster will be equal to the number of nodes in the cluster.
So the above code calculates number of allocatable ip addr >= number of nodes.
But the above did not consider witness node into account which will not run instance manager pod.
This caused error msg in storage network settings, though ips are allocated successfully to the instance manager pods.

```
IP allocation failure for Longhorn Pods.
This message can be found in:
 Harvester UI → Advanced → Settings → Storage-Network.
```

kubectl get settings.harvesterhci.io storage-network -o yaml

```
status:
  conditions:
  - lastUpdateTime: "2025-10-06T10:40:44Z"
    message: IP allocation failure for Longhorn Pods
    reason: In Progress
    status: "False"
    type: configured
value: '{"vlan":45,"clusterNetwork":"storage","range":"172.18.0.0/16"}'
```

### To Reproduce

1.Have a harvester cluster with atleast 1 witness node
2.Make sure to have only instance manager pods running and no backing image pods.
3.Create storage network
4.Check the IP allocation failure for Longhorn Pods as mentioned above in Harvester UI or kubectl storage network settings

Example: 1 control plane node and 1 witness node in the cluster

The above cluster will be running only 1 instance manager pod, so whereabouts would have allocated 1 ip address.
But during the validation, number of wherabouts allocated ip addr (1) >= number of nodes(2) fails and error msg is seen.Instead the witness node has to be excluded from number of nodes and the validation will succeed.

### Expected Behavior

No error msg in the storage network settings.

### Support Bundle for Troubleshooting

.

### Environment

- Harvester version:v1.5.1
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:N/A

albinsun:
Test FAIL, still see error message in UI under storage network settings with the above config (1 default + 1 witness).
Please help to check, thx.


## Test Scenario
### :red_circle: 1st Trial
1. <details>
   <summary> ✔️ Setup 2 nodes witness <code>harvester-v1.7.0-rc1</code>
   </summary>

[...]

PR #9364, opened 2025-10-23
skip witness node during node count calculation running LH Pods

rrajendran17:
#### Problem:
"IP allocation failure for Longhorn Pods" error msg seen on Harvester UI under storage network settings.
Since witness node does not run LH pod and it is also taken into account for node count calculation and validated against the number of ips allocated from whereabouts, error msg is seen.

#### Solution:
Exclude witness node during node count calculation

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9325

#### Test plan:

To reproduce:
1.Have a harvester cluster with atleast 1 witness node
2.Make sure to have only instance manager pods running and no backing image pods.
3.Create storage network
4.Check the IP allocation failure for Longhorn Pods as mentioned above in Harvester UI or kubectl storage network settings

Example: 1 control plane node and 1 witness node in the cluster

The above cluster will be running only 1 instance manager pod, so whereabouts would have allocated 1 ip address.
But during the validation, number of wherabouts allocated ip addr (1) >= number of nodes(2) fails and error msg is seen.Instead the witness node has to be excluded from number of nodes and the validation will succeed.

With fix:
No error msg should be seen in UI under storage network settings with the above config.

### DIFF
--- a/pkg/controller/master/storagenetwork/storage_network.go
+++ b/pkg/controller/master/storagenetwork/storage_network.go
@@ -373,7 +373,7 @@ func (h *Handler) validateIPAddressesAllocations(setting *harvesterv1.Setting) e
 	//Formula - https://docs.harvesterhci.io/v1.4/advanced/storagenetwork/
 	//Dynamic parameters like number of images download/upload, backing-image-manager and backing-image-ds are skipped
 	//and only the number of nodes each running an instance-manager pod is used
-	MinAllocatableIPAddrs := len(nodes)
+	MinAllocatableIPAddrs := util.CountNonWitnessNodes(nodes)
 
 	var config network.Config
 
--- a/pkg/util/node.go
+++ b/pkg/util/node.go
@@ -4,6 +4,18 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	KubeNodeRoleLabelPrefix      = "node-role.kubernetes.io/"
+	KubeMasterNodeLabelKey       = KubeNodeRoleLabelPrefix + "master"
+	KubeControlPlaneNodeLabelKey = KubeNodeRoleLabelPrefix + "control-plane"
+	KubeEtcdNodeLabelKey         = KubeNodeRoleLabelPrefix + "etcd"
+
+	PromoteStatusComplete = "complete"
+	PromoteStatusRunning  = "running"
+	PromoteStatusUnknown  = "unknown"
+	PromoteStatusFailed   = "failed"
+)
+
 func ExcludeWitnessNodes(nodes []*corev1.Node) []*corev1.Node {
 	nonWitnessNodes := make([]*corev1.Node, 0, len(nodes))
 	for _, node := range nodes {
@@ -12,4 +24,66 @@ func ExcludeWitnessNodes(nodes []*corev1.Node) []*corev1.Node {
 		}
 	}
 	return nonWitnessNodes
+}
+
+func IsPromoteStatusIn(node *corev1.Node, statuses ...string) bool {
+	status, ok := node.Annotations[HarvesterPromoteStatusAnnotationKey]
+	if !ok {
+		return false
+	}
+
+	for _, s := range statuses {
+		if status == s {
+			return true
+		}
+	}
+
+	return false
+}
+
+func IsWitnessNode(node *corev1.Node, isManagement bool) bool {
+	_, found := node.Labels[HarvesterWitnessNodeLabelKey]
+	if !found {
+		return false
+	}
+
+	// promotion has already been run for this node
+	if found && (isManagement || IsPromoteStatusIn(node, PromoteStatusComplete, PromoteStatusRunning, PromoteStatusFailed, PromoteStatusUnknown)) {
+		return true
+	}
+
+	return false
+}
+
+// IsManagementRole determine whether it's an management node based on the node's label.
+// Management Role included: master, control-plane, etcd
+func IsManagementRole(node *corev1.Node) bool {
+	if value, ok := node.Labels[KubeMasterNodeLabelKey]; ok {
+		return value == "true"
+	}
+
+	// Related to https://github.com/kubernetes/kubernetes/pull/95382
+	if value, ok := node.Labels[KubeControlPlaneNodeLabelKey]; ok {
+		return value == "true"
+	}
+
+	// Now we have the witness node, we need to count it as a management node
+	if value, ok := node.Labels[KubeEtcdNodeLabelKey]; ok {
+		return value == "true"
+	}
+
+	return false
+}
+
+// count the number of nodes running instance manager pod
+func CountNonWitnessNodes(nodes []*corev1.Node) int {
+	count := 0
+
+	for _, node := range nodes {
+		if !IsWitnessNode(node, IsManagementRole(node)) {
+			count++
+		}
+	}
+
+	return count
 }
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -29,11 +29,6 @@ import (
 const (
 	promoteControllerName = "promote-node-controller"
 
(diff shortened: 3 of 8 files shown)


pkg/controller/master/storagenetwork/storage_network.go
w13915984028, line 420: Due to above failure case, L427 is better to optimize, to avoid tight looping, LH takes time to apply SN

```
		h.settingsController.Enqueue(setting.Name)
->
		h.settingsController.EnqueueAfter(setting.Name, 1*time.Second)
```

w13915984028, line 424: note: the `ReasonCompleted` is set after the stoage-network was successfully rolled-out

please double check, if it is better to check this condition at the beginning of handleLonghornSettingPostConfig or it's caller

```
func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
```

the above functions calls `checkPodStatusAndStart` anyway, and that might affect other pods if they are operated in other situation


everytime, when the leader harvester controller starts to run, it gets the setting `onChange` event

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -373,7 +374,13 @@ func (h *Handler) validateIPAddressesAllocations(setting *harvesterv1.Setting) e
 	//Formula - https://docs.harvesterhci.io/v1.4/advanced/storagenetwork/
 	//Dynamic parameters like number of images download/upload, backing-image-manager and backing-image-ds are skipped
 	//and only the number of nodes each running an instance-manager pod is used
-	MinAllocatableIPAddrs := len(nodes)
+	MinAllocatableIPAddrs := 0
+	for _, node := range nodes {
+		//skip witness nodes which do not run LH Pods
+		if !nodectl.IsWitnessNode(node, nodectl.IsManagementRole(node)) {
```
w13915984028: if some nodes on a big cluster are down when user sets the storage-network, could this error also be set ?  the `ippool.Spec.Allocations` could be less than `MinAllocatableIPAddrs`

```
if len(ippool.Spec.Allocations) >= MinAllocatableIPAddrs
```

do we have better ways to do? e.g.

list all the live `instance-manager` PODs, and check their annotations, the successful POD has below annotation to show that it gets the SN IP

https://docs.harvesterhci.io/v1.4/advanced/storagenetwork/#step-3

```
    k8s.v1.cni.cncf.io/network-status: |-
      [{
          "name": "k8s-pod-network",
          "ips": [
              "10.52.2.122"
          ],
          "default": true,
          "dns": {}
      },{
          "name": "harvester-system/storagenetwork-95bj4",
          "interface": "lhnet1",
          "ips": [
              "192.168.0.3"
          ],
          "mac": "2e:51:e6:31:96:40",
          "dns": {}
      }]
```

w13915984028 (commented):
some questions about the SN setting, thanks.


btw: it is safe to remove blow function, `rancher-monitoring` is not a managedchart on v170
```

func (h *Handler) checkRancherMonitoringStatusAndStart() error {
	// check managedchart fleet-local/rancher-monitoring paused
	monitoring, err := h.managedChartCache.Get(util.FleetLocalNamespaceName, util.RancherMonitoring)

```

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: I do not get your point correctly.
The  len(ippool.Spec.Allocations) will be same as the ips allocated to the instance manager pods.
so for the nodes which are down, SN cannot be configured on those LH pods(no secondary network annotations) and ips also cannot be allocated by whereabouts.

We can hit the problem if ip allocated(whereabouts or instance manager pod) < number of nodes if nodes are not removed from nodeCache when node is down/lose connectivity from rest of the cluster.

Do yo think we need something at a broader level to cleanup or remove nodes from nodeCache after a time period if we lose connectivity to that node(like a keep alive) ?I believe majority of our code relies on nodeCache for checking presence of nodes in the cluster.

rrajendran17:
> some questions about the SN setting, thanks.
> 
> btw: it is safe to remove blow function, `rancher-monitoring` is not a managedchart on v170
> 
> ```
> 
> func (h *Handler) checkRancherMonitoringStatusAndStart() error {
> 	// check managedchart fleet-local/rancher-monitoring paused
> 	monitoring, err := h.managedChartCache.Get(util.FleetLocalNamespaceName, util.RancherMonitoring)
> ```

I am not sure about this, do you want to remove checkRancherMonitoringStatusAndStart() and checkRancherMonitoringStatusAndStop().If this requires a different QA scope, we can track this separately.

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: I am not sure if I am getting this right.
You mean set the Reason as completed on post config and then validate the rest of the functions and in case of any error, reasonCompleted will be modified to specific error msg in case of any failure from checkPodStatusAndStart() or validateIPAddressesAllocations() ?

I am not sure if it will be right to move L431 to beginning of function handleLonghornSettingPostConfig() as we still have other validations pending.

rrajendran17:
May be we need to change the description and broaden the QA testplan scope if we make additional changes to this PR.
Or will create a new PR for storage network error handling improvements.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -373,7 +374,13 @@ func (h *Handler) validateIPAddressesAllocations(setting *harvesterv1.Setting) e
...
+			MinAllocatableIPAddrs++
+		}
+	}
```
WebberHuang1118: Suggest factoring this part out into a utility function.

WebberHuang1118 (dismissed):
LGTM, just a suggestion, thanks.

WebberHuang1118:
> May be we need to change the description and broaden the QA testplan scope if we make additional changes to this PR. Or will create a new PR for storage network error handling improvements.

IMO, we might create a separate ticket to track the storage network controller enhancement. It doesn’t necessarily need to be tied to this issue.

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: created a new function and moved some of the dependent functions under pkg/util/node.go

rrajendran17:
Created GH issue https://github.com/harvester/harvester/issues/9375 for any further enhancements.

pkg/util/node.go
```diff
@@ -13,3 +25,65 @@ func ExcludeWitnessNodes(nodes []*corev1.Node) []*corev1.Node {
...
+	// Now we have the witness node, we need to count it as a management node
+	if value, ok := node.Labels[KubeEtcdNodeLabelKey]; ok {
+		return value == "true"
+	}
+
+	return false
+}
+
+// count the number of nodes running instance manager pod
+func CountAllocatableNodes(nodes []*corev1.Node) int {
```
WebberHuang1118: The current function name is a bit confusing — for example, renaming it to **`CountNonWitnessNode`** might make it clearer.

WebberHuang1118 (dismissed):
LGTM, just a nit, thanks.

pkg/controller/master/storagenetwork/storage_network.go
w13915984028: In K8s, most CRD object has `spec` and `status` field, the controller works per `spec` and update `status` accordingly.

When a node is down/offline, K8s controller updates it's status and say like `kubelet stops to update node...`.

The code here, use len(node) to make further decision, the assumption is not correct all the time. As a consumer of the  K8s `node` object, check `status` will make the code even robust.

We can use this PR fix it for now, but better to refactor on issue https://github.com/harvester/harvester/issues/9375

WebberHuang1118 (approved):

Issue #8561, opened 2025-06-27
[ENHANCEMENT] Both `vmstate-persistence` and `longhorn-static` StorageClasses should be treated as internal and not be used directly by users


starbops:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

We found that creating volumes from the two StorageClasses, `vmstate-persistence` and `longhorn-static`, and attaching them to existing VMs causes the VMs to run incorrectly.

As clarified by @Vicente-Cheng ([ref](https://github.com/harvester/harvester/issues/8543#issuecomment-3011137289)), these two SCs lack the essential parameter `migratable: "true"`, thus subsequently created volumes will be of type Filesystem RWX volume.

### Describe the solution you'd like

We may need to do either of the following (or both):

- Prevent users from creating volumes from both the `vmstate-persistence` and `longhorn-static` StorageClasses (require UI changes and additional validating webhook at the backend)
- Document the reason why these two SCs should not be used directly (and perhaps explain a bit more on the purposes of different default SCs)

### Describe alternatives you've considered

N/A

### Additional context

This is a follow-up ticket for #8543.

Vicente-Cheng:
Hi @brandboat, 
Could you help with this issue?

albinsun:
Test PASS, volume can not be create base on these 2 reserved storages classes. Close as done.


## Environment
* `harvester-v1.7.0-rc1`
  * Profile: Single node AMD64 Bare-metal HPE ProLiant DL360 
  * ui-source: `Auto`


## Test Scenario
### 🟢 longhorn-static
1. <details>
   <summary> Create a <code>pvc-my-vol.yaml</code> using reserved storage class <b>longhorn-static</b>
   </summary>


   <img width="1566" height="1011" alt="Image" src="https://github.com/user-attachments/assets/886d47de-b75a-45b3-b27e-a692f0fd12e2" />
   </details>

1. <details>
   <summary> Should be fail to apply yaml due to <code>can not create volume with the reserved storage class</code>
   </summary>

   <img width="1566" height="808" alt="Image" src="https://github.com/user-attachments/assets/a0a2001d-6c1f-4205-826a-4b639ba0290c" />
   </details>


### :green_circle: vmstate-persistence
1. <details>
   <summary> Create a <code>pvc-my-vol.yaml</code> using reserved storage class <b>vmstate-persistence</b>
   </summary>

[...]

PR #9366, opened 2025-10-24
enhancement: check internal storageclasses when creating new volumes

DamiaSan:
#### Problem:
Storageclasses longhorn-static and vmstate-persistence must be used only for internal purpose and so can't be used directly by users when creating new volumes.

#### Solution:
Added a check in the pvc webhook validator to check if the new pvc to be created is related to one one of these internal storageclasses.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8561

#### Test plan:

1. Create a `volume.yaml` file like the following one
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    volume.beta.kubernetes.io/storage-provisioner: driver.longhorn.io
    volume.kubernetes.io/storage-provisioner: driver.longhorn.io
  name: my-vol
  namespace: default
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
  storageClassName: longhorn-static
  volumeMode: Block
  volumeName: pvc-my-vol
 ```

2. Apply the yaml with the command
```
kubectl apply -f volume.yaml
```
3. Check that the output of the command is the following one
```
The request is invalid: spec.storageClassName: can not create volume with the reserved storage class longhorn-static
```

4. Check that the pvc has not been created with the command
```
kubectl get pvc my-vol
```

5. Try again setting the `storageClassName`  to `vmstate-persistence`, the result must be same

### DIFF
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -190,8 +190,11 @@ const (
 	LablelVClusterAppNameKey   = "app"
 	LablelVClusterAppNameValue = "vcluster"
 
-	StorageClassHarvesterLonghorn = "harvester-longhorn" // the initial & default storageclass
-	HarvesterChartReleaseName     = "harvester"          // the release name
+	StorageClassHarvesterLonghorn  = "harvester-longhorn"  // the initial & default storageclass
+	StorageClassLonghornStatic     = "longhorn-static"     // internal storageclass used for management of existing Longhorn volumes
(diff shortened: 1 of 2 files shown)


w13915984028 (approved):
LGTM, thanks.

m-ildefons (approved):
LGTM. Would be nice to have unit tests though.

Issue #8797, opened 2025-08-04
[FEATURE] Include vm cpu model in filtering nodes during vm migration


ibrokethecloud:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

Kubevirt allows users to specify cpu-model at a per VM basis or a cluster level using the kubevirt CR.

When a cpu-model is set, the cpu model is used in the `nodeSelector` for the launcher pod

```
  nodeSelector:
    cpu-model.node.kubevirt.io/EPYC-Rome: "true"
    kubernetes.io/arch: amd64
    kubevirt.io/schedulable: "true"
```

Harvester currently does not apply a cpu model filter to the list of nodes available as a target for VM migration

### Describe the solution you'd like

Improve the target node identification logic to include cpu-model from the VM or cluster wide settings.

martindekov:
@ibrokethecloud I wanted to make sure I got things right. So (my env is 1.6 that's why I reference the 1.6 branch) when trying to migrate VM I see this is being called:
https://github.com/harvester/harvester/blob/8713e69da7b296b5f5b7da8de07def89c083b631/pkg/api/vm/handler.go#L109

And this calls:
https://github.com/harvester/harvester/blob/8713e69da7b296b5f5b7da8de07def89c083b631/pkg/api/vm/handler.go#L109

So we want to add filtering logic for both 1.6.1 and 1.7 in this method:
https://github.com/harvester/harvester/blob/89c87bb356428ac8b10913ab019dac6f4f2f7609/pkg/api/vm/handler.go#L607-L625

In case our virtual machine - which we want to migrate has specific requirements for CPU through those node selectors:
[cpu-model.node.kubevirt.io](https://github.com/kubevirt/kubevirt/blob/9c6b929caf22d7acfc035f9e064ab956326371c9/staging/src/kubevirt.io/api/core/v1/types.go#L1213)   
[kubevirt.io/schedulable](https://github.com/kubevirt/kubevirt/blob/9c6b929caf22d7acfc035f9e064ab956326371c9/staging/src/kubevirt.io/api/core/v1/types.go#L1148)
[kubernetes.io/arch](https://github.com/kubernetes/kubernetes/blob/37f067778e9ef714d02e5e901c25e83170144dc1/staging/src/k8s.io/api/core/v1/well_known_labels.go#L46
)

We want to only return nodes fitting those labels. Does this summarize the solution and the right place where I need to inject the changes? Thank you in advance for the input!

[...]

Issue #7583, opened 2025-02-11
[BUG] Node Scheduling tab of VMs does not work


votdev:
**Describe the bug**
The node scheduling page of VMs has several issues:

- You can not delete existing rules.
- The value field of a rule cuts the last character.
- If you configure the `is set` operator, it is displayed as `in list` when the page is displayed the next time after saving.

https://github.com/user-attachments/assets/bef54cf1-a814-4bc8-a434-55495243f2e0

https://github.com/user-attachments/assets/678fa968-6383-4ff5-88a1-f9cccc7f0bc9

![Image](https://github.com/user-attachments/assets/506892a5-f123-47f0-b546-6b80d3dc76ed)

**To Reproduce**

***Case 1****
- Create a VM.
- Choose `Edit Config`. Go to the `Node Scheduling` tab.
- Select `Run virtual machine on node(s) matching scheduling rules`.
- Add a new Rule. Use
```
Priority: Required
Key: cpumanager
Operator: is set
```
- Press `Save`button.

***Case 2****
- Create a VM.
- Choose `Edit Config`. Go to the `Node Scheduling` tab.
- Select `Run virtual machine on node(s) matching scheduling rules`.
- Add a new Rule. Use
```
Priority: Required
Key: cpumanager
Operator: in list
Value: yes
```
- Press `Save`button.

**Expected behavior**
It should work.

**Support bundle**
n/a

**Environment**
 - Harvester ISO version: master-head (v1.5)
 - Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): 3 node cluster (ipxe-examples)

houhoucoop:
When updating the VM, I included `spec.template.spec.affinity` in the request payload, but the `spec.template.spec.affinity` in the response is empty.

**The request payload**
![Image](https://github.com/user-attachments/assets/c871675d-94a1-4e51-a9e2-6180c14bc4a4)

**The response**
![Image](https://github.com/user-attachments/assets/d2823955-686f-487b-bd48-1eff50e20234)

starbops:
This is a little bit tricky. It seems that only when the key is `cpumanager`, no matter what the operator or value is, the rule will be dropped by the backend.

@brandboat Do you happen to know any behavior that causes this result? Since it's the CPU manager, I think we or KubeVirt might do something about it.

brandboat:
> @brandboat Do you happen to know any behavior that causes this result? Since it's the CPU manager, I think we or KubeVirt might do something about it.

[...]

PR #9372, opened 2025-10-27
Filter migration nodes based on VMI pod's node selectors

martindekov:
Filtering nodes for migration based on VMI pod's nodeSelection field. We construct label selector based on pod's node selector when picking up available nodes for migration

#### Problem:
Currently we omit filtering of nodes based on criteria like whether node is schedulable, what is the CPU model of the node and what is the CPU which the VM requires if any. We don't respect VMI's pod nodeSelectors when figuring out compatible nodes for VM migration.

#### Solution:
When listing possible nodes for migration we use label selector built using the virtual machine instance pod's nodeSelectors.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8797

#### Test plan:
Unit tests - added new nodes and new test case with the fields

Basic E2E test validation on virtualization:
1. Create 2 node harvester cluster `node-1` and `node-2`
2. Create VM - `vm-1` on the cluster
3. By default since nodes are similar - created on virtualization you should be able to see the other node and be able to migrate to it
4. Adding affinity rules to vm should make it unschedulable, and then once you label the nodes the vm will be scheduled and each node which contains the labels should be seen as migration target as well

Validate Node Scheduling works:
1. Create a VM
2. By default VM is on - `Run virtual machine on any available node` so migration button was present
3. Enabled VM to be hosted only on the current node - `Run virtual machine on specific node - (Live migration is not supported)`
4. Migration button dissapeared
5. Reverted VM to original state of - `Run virtual machine on any available node`
6. Migrated VM to another node
7. After migration enabled - `Run virtual machine on specific node - (Live migration is not supported)`
8. Migration button dissapeared

Results for Basic E2E test validation on virtualization:

<details>
<summary>Built cluster with this change and custom image to replace the controller and used it on 2 node cluster</summary>
<br>

```
node-1:/home/rancher # kubectl get deployment harvester -n harvester-system -o yaml | grep image
          value: '{"repository":"rancher/support-bundle-kit","tag":"master-head","imagePullPolicy":"IfNotPresent"}'
        image: rancher/harvester:cpufilter8797-head-1
        imagePullPolicy: IfNotPresent
```
</details>

<details>
<summary>Virtual machine on one node allows migration to another node</summary>
<br>
image
image
</details>

<details>
<summary>Adding affinity while vm is on node-1 makes node-2 dissapear as potential target</summary>
<br>
image
image
image
image
image
image
</details>

<details>
<summary>While migration is running you migration button dissapears</summary>
<br>
image
</details>

Results for Validate Node Scheduling works:
<details>
<summary>Create VM by default schedulable to all nodes so migratable</summary>
<br>

image
image

```
node-1:/home/rancher # kubectl get pod virt-launcher-vm-1-rrgzp -n default -o jsonpath='{.spec.nodeSelector}' | jq
{
  "kubernetes.io/arch": "amd64",
  "kubevirt.io/schedulable": "true",
  "machine-type.node.kubevirt.io/q35": "true"
}
```
</details>

<details>
<summary>Make VM schedulable on specific node - current one and migration button dissapears</summary>
<br>
image
image

```
node-1:/home/rancher # kubectl get pod virt-launcher-vm-1-cbjpd -n default -o jsonpath='{.spec.nodeSelector}' | jq
{
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "node-2",
  "kubevirt.io/schedulable": "true",
  "machine-type.node.kubevirt.io/q35": "true"
}
```
</details>

<details>
<summary>Revert vm schedulable to all nodes, migrate it and see migration back and forth is allowed</summary>
<br>
image

```
node-1:/home/rancher # kubectl get pod virt-launcher-vm-1-cbx4r -n default -o jsonpath='{.spec.nodeSelector}' | jq
{
  "cpu-feature.node.kubevirt.io/arch-capabilities": "true",
  "cpu-feature.node.kubevirt.io/cmp_legacy": "true",
  "cpu-feature.node.kubevirt.io/flush-l1d": "true",
  "cpu-feature.node.kubevirt.io/flushbyasid": "true",
  "cpu-feature.node.kubevirt.io/gds-no": "true",
  "cpu-feature.node.kubevirt.io/hypervisor": "true",
  "cpu-feature.node.kubevirt.io/ibpb-brtype": "true",
  "cpu-feature.node.kubevirt.io/lbrv": "true",
  "cpu-feature.node.kubevirt.io/mds-no": "true",
  "cpu-feature.node.kubevirt.io/overflow-recov": "true",
  "cpu-feature.node.kubevirt.io/pause-filter": "true",
  "cpu-feature.node.kubevirt.io/perfmon-v2": "true",
  "cpu-feature.node.kubevirt.io/pfthreshold": "true",
  "cpu-feature.node.kubevirt.io/pschange-mc-no": "true",
  "cpu-feature.node.kubevirt.io/rdctl-no": "true",
  "cpu-feature.node.kubevirt.io/rfds-no": "true",
  "cpu-feature.node.kubevirt.io/sbpb": "true",
  "cpu-feature.node.kubevirt.io/skip-l1dfl-vmentry": "true",
  "cpu-feature.node.kubevirt.io/ssbd": "true",
  "cpu-feature.node.kubevirt.io/stibp": "true",
  "cpu-feature.node.kubevirt.io/succor": "true",
  "cpu-feature.node.kubevirt.io/tsc-deadline": "true",
  "cpu-feature.node.kubevirt.io/tsc-scale": "true",
  "cpu-feature.node.kubevirt.io/tsc_adjust": "true",
  "cpu-feature.node.kubevirt.io/vgif": "true",
  "cpu-feature.node.kubevirt.io/virt-ssbd": "true",
  "cpu-feature.node.kubevirt.io/vmcb-clean": "true",
  "cpu-feature.node.kubevirt.io/x2apic": "true",
  "cpu-model-migration.node.kubevirt.io/EPYC-Genoa": "true",
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "node-1",
  "kubevirt.io/schedulable": "true",
  "machine-type.node.kubevirt.io/pc-q35-9.2": "true"
}
node-1:/home/rancher # kubectl get pod virt-launcher-vm-1-dwjrf -n default -o jsonpath='{.spec.nodeSelector}' | jq
{
  "kubernetes.io/arch": "amd64",
  "kubevirt.io/schedulable": "true",
  "machine-type.node.kubevirt.io/q35": "true"
}
```
</details>

<details>
<summary>Enabling vm to be ran on specific node after migration removes ability to migrate</summary>
<br>
image
image

```
node-1:/home/rancher # kubectl get pods
NAME                       READY   STATUS    RESTARTS   AGE
virt-launcher-vm-1-nxqn4   2/2     Running   0          2m23s
node-1:/home/rancher # kubectl get pod virt-launcher-vm-1-nxqn4 -n default -o jsonpath='{.spec.nodeSelector}' | jq
{
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "node-1",
  "kubevirt.io/schedulable": "true",
  "machine-type.node.kubevirt.io/q35": "true"
}
```
</details>

N/A

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -653,6 +653,9 @@ func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.Virt
 
 	nodeFilter := labels.NewSelector()
 	for key, value := range latestPod.Spec.NodeSelector {
+		if key == corev1.LabelHostname {
+			continue
+		}
 		requirement, err := labels.NewRequirement(key, selection.Equals, []string{value})
 		if err != nil {
 			return nil, fmt.Errorf("failed to create requirement for %s=%s: %w", key, value, err)
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -652,6 +652,7 @@ func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.Virt
 	}
 
 	nodeFilter := labels.NewSelector()
+
 	for key, value := range latestPod.Spec.NodeSelector {
 		if key == corev1.LabelHostname {
 			continue
@@ -662,7 +663,69 @@ func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.Virt
 		}
 		nodeFilter = nodeFilter.Add(*requirement)
 	}
+
+	if isRequiredAffinityFilterPresent(latestPod) {
+		required := latestPod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+		var err error
+		nodeFilter, err = addNodeAffinityFilters(nodeFilter, required.NodeSelectorTerms)
+		if err != nil {
+			return nil, err
+		}
+
+	}
+
 	return nodeFilter, nil
+}
+
+func isRequiredAffinityFilterPresent(pod *corev1.Pod) bool {
+	return pod.Spec.Affinity != nil &&
+		pod.Spec.Affinity.NodeAffinity != nil &&
+		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil
+}
+
+func addNodeAffinityFilters(nodeFilter labels.Selector, terms []corev1.NodeSelectorTerm) (labels.Selector, error) {
+	for _, term := range terms {
+		for _, expr := range term.MatchExpressions {
+			if expr.Key == corev1.LabelHostname {
+				continue
+			}
+
+			requirement, err := convertNodeSelectorRequirementToSelector(expr)
+			if err != nil {
+				return nil, err
+			}
+			if requirement != nil {
+				nodeFilter = nodeFilter.Add(*requirement)
+			}
+		}
+	}
+	return nodeFilter, nil
+}
+
+func convertNodeSelectorRequirementToSelector(req corev1.NodeSelectorRequirement) (*labels.Requirement, error) {
+	var op selection.Operator
+	switch req.Operator {
+	case corev1.NodeSelectorOpIn:
+		op = selection.In
+	case corev1.NodeSelectorOpNotIn:
+		op = selection.NotIn
+	case corev1.NodeSelectorOpExists:
+		op = selection.Exists
+	case corev1.NodeSelectorOpDoesNotExist:
+		op = selection.DoesNotExist
+	case corev1.NodeSelectorOpGt, corev1.NodeSelectorOpLt:
+		logrus.Debugf("Skipping unsupported node selector operator %s for key %s", req.Operator, req.Key)
+		return nil, nil
+	default:
+		logrus.Warnf("Unknown node selector operator %s for key %s", req.Operator, req.Key)
+		return nil, nil
+	}
+
+	requirement, err := labels.NewRequirement(req.Key, op, req.Values)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create requirement for %s %v %v: %w", req.Key, op, req.Values, err)
+	}
+	return requirement, nil
 }
 
 func (h *vmActionHandler) getLatestVMIPodOnCurrentNode(vmi *kubevirtv1.VirtualMachineInstance) (*corev1.Pod, error) {
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -48,7 +48,6 @@ import (
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
-	"github.com/harvester/harvester/pkg/indexeres"
 	harvesterServer "github.com/harvester/harvester/pkg/server/http"
 	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
@@ -643,7 +642,7 @@ func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.Virt
 		return labels.Everything(), nil
 	}
 
-	latestPod, err := h.getLatestVMIPodOnCurrentNode(vmi)
+	latestPod, err := h.getVMIPod(vmi)
 	if err != nil {
 		return labels.Everything(), err
 	}
@@ -728,23 +727,30 @@ func convertNodeSelectorRequirementToSelector(req corev1.NodeSelectorRequirement
 	return requirement, nil
 }
 
-func (h *vmActionHandler) getLatestVMIPodOnCurrentNode(vmi *kubevirtv1.VirtualMachineInstance) (*corev1.Pod, error) {
-	vmiPods, err := h.podCache.GetByIndex(indexeres.PodByVMIUIDIndex, string(vmi.UID))
+func (h *vmActionHandler) getVMIPod(vmi *kubevirtv1.VirtualMachineInstance) (*corev1.Pod, error) {
+	if len(vmi.Status.ActivePods) == 0 {
+		return nil, fmt.Errorf("there are no active pods for VMI: %s/%s, migration target cannot be picked unless only 1 pod is active", vmi.Namespace, vmi.Name)
+	}
+	if len(vmi.Status.ActivePods) > 1 {
+		return nil, fmt.Errorf("there are multiple active pods for VMI: %s/%s, migration target can be picked only when at most 1 pod is active", vmi.Namespace, vmi.Name)
+	}
+
+	selector := labels.SelectorFromSet(labels.Set{
+		kubevirtv1.CreatedByLabel: string(vmi.UID),
+	})
+
+	vmiPods, err := h.podCache.List(vmi.Namespace, selector)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get pods for VMI %s/%s: %w", vmi.Namespace, vmi.Name, err)
 	}
 
-	var latestPod *corev1.Pod
 	for _, pod := range vmiPods {
-		if vmi.Status.NodeName != "" && pod.Spec.NodeName != vmi.Status.NodeName {
-			continue
-		}
-
-		if latestPod == nil || pod.CreationTimestamp.After(latestPod.CreationTimestamp.Time) {
-			latestPod = pod
+		if pod.Status.Phase != corev1.PodSucceeded && pod.Status.Phase != corev1.PodFailed {
+			return pod, nil
 		}
 	}
-	return latestPod, nil
+
+	return nil, fmt.Errorf("no active pod found for VMI %s/%s", vmi.Namespace, vmi.Name)
 }
 
 func (h *vmActionHandler) createVMBackup(vmName, vmNamespace string, input BackupInput) error {
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -728,13 +728,6 @@ func convertNodeSelectorRequirementToSelector(req corev1.NodeSelectorRequirement
 }
 
 func (h *vmActionHandler) getVMIPod(vmi *kubevirtv1.VirtualMachineInstance) (*corev1.Pod, error) {
-	if len(vmi.Status.ActivePods) == 0 {
-		return nil, fmt.Errorf("there are no active pods for VMI: %s/%s, migration target cannot be picked unless only 1 pod is active", vmi.Namespace, vmi.Name)
-	}
-	if len(vmi.Status.ActivePods) > 1 {
-		return nil, fmt.Errorf("there are multiple active pods for VMI: %s/%s, migration target can be picked only when at most 1 pod is active", vmi.Namespace, vmi.Name)
-	}
-
 	selector := labels.SelectorFromSet(labels.Set{
 		kubevirtv1.CreatedByLabel: string(vmi.UID),
 	})
@@ -744,13 +737,21 @@ func (h *vmActionHandler) getVMIPod(vmi *kubevirtv1.VirtualMachineInstance) (*co
 		return nil, fmt.Errorf("failed to get pods for VMI %s/%s: %w", vmi.Namespace, vmi.Name, err)
 	}
 
+	var activePods []*corev1.Pod
 	for _, pod := range vmiPods {
 		if pod.Status.Phase != corev1.PodSucceeded && pod.Status.Phase != corev1.PodFailed {
-			return pod, nil
+			activePods = append(activePods, pod)
 		}
 	}
 
-	return nil, fmt.Errorf("no active pod found for VMI %s/%s", vmi.Namespace, vmi.Name)
+	if len(activePods) == 0 {
+		return nil, fmt.Errorf("there are no active pods for VMI: %s/%s, migration target cannot be picked unless only 1 pod is active", vmi.Namespace, vmi.Name)
+	}
+	if len(activePods) > 1 {
+		return nil, fmt.Errorf("there are multiple active pods for VMI: %s/%s, migration target can be picked only when at most 1 pod is active", vmi.Namespace, vmi.Name)
+	}
+
+	return activePods[0], nil
 }
 
 func (h *vmActionHandler) createVMBackup(vmName, vmNamespace string, input BackupInput) error {
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -637,21 +637,48 @@ func isDrained(node *corev1.Node) bool {
 }
 
 func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.VirtualMachineInstance) (labels.Selector, error) {
-	if vmi == nil || vmi.Spec.Affinity == nil || vmi.Spec.Affinity.NodeAffinity == nil || vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
-		return labels.Everything(), nil
-	}
-
 	nodeSelector := labels.NewSelector()
-	terms := vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
-	for _, term := range terms {
-		for _, e := range term.MatchExpressions {
-			r, err := convertNodeSelectorRequirementToSelector(e)
-			if err != nil {
-				return nil, err
+
+	if vmi != nil && vmi.Spec.Affinity != nil && vmi.Spec.Affinity.NodeAffinity != nil && vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+		terms := vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+		for _, term := range terms {
+			for _, e := range term.MatchExpressions {
+				req, err := convertNodeSelectorRequirementToSelector(e)
+				if err != nil {
+					return nil, err
+				}
+				nodeSelector = nodeSelector.Add(*req)
 			}
-			nodeSelector = nodeSelector.Add(*r)
 		}
 	}
+
+	if vmi != nil && vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
+		key := fmt.Sprintf("%s%s", kubevirtv1.CPUModelLabel, vmi.Spec.Domain.CPU.Model)
+		req, err := labels.NewRequirement(key, selection.Equals, []string{"true"})
+		if err != nil {
+			return nil, err
+		}
+		nodeSelector = nodeSelector.Add(*req)
+	}
+
+	if vmi != nil && vmi.Spec.Domain.CPU != nil {
+		for _, feature := range vmi.Spec.Domain.CPU.Features {
+			if feature.Policy == "" || feature.Policy == "require" {
+				key := fmt.Sprintf("%s%s", kubevirtv1.CPUFeatureLabel, feature.Name)
+				req, err := labels.NewRequirement(key, selection.Equals, []string{"true"})
+				if err != nil {
+					return nil, err
+				}
+				nodeSelector = nodeSelector.Add(*req)
+			}
+		}
+	}
+
+	req, err := labels.NewRequirement(kubevirtv1.NodeSchedulable, selection.Equals, []string{"true"})
+	if err != nil {
+		return nil, err
+	}
+	nodeSelector = nodeSelector.Add(*req)
 
 	return nodeSelector, nil
 }
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -644,7 +644,7 @@ func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.Virt
 
 	vmiPod, err := h.getVMIPod(vmi)
 	if err != nil {
-		return labels.Everything(), err
+		return labels.Nothing(), err
 	}
 
 	nodeFilter := labels.NewSelector()
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -652,9 +652,14 @@ func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.Virt
 		}
 	}
 
-	if vmi != nil && vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
-		key := fmt.Sprintf("%s%s", kubevirtv1.CPUModelLabel, vmi.Spec.Domain.CPU.Model)
-		req, err := labels.NewRequirement(key, selection.Equals, []string{"true"})
+	cpuModel, err := h.getCPUModel(vmi)
+	if err != nil {
+		return nil, err
+	}
+
+	if cpuModel != "" {
+		cpuModelKey := fmt.Sprintf("%s%s", kubevirtv1.CPUModelLabel, cpuModel)
+		req, err := labels.NewRequirement(cpuModelKey, selection.Equals, []string{"true"})
 		if err != nil {
 			return nil, err
 		}
@@ -681,6 +686,24 @@ func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.Virt
 	nodeSelector = nodeSelector.Add(*req)
 
 	return nodeSelector, nil
+}
+
+func (h *vmActionHandler) getCPUModel(vmi *kubevirtv1.VirtualMachineInstance) (string, error) {
+	if vmi != nil && vmi.Spec.Domain.CPU != nil && vmi.Spec.Domain.CPU.Model != "" {
+		return vmi.Spec.Domain.CPU.Model, nil
+	}
+	return h.getClusterDefaultCPUModel()
+}
+
+func (h *vmActionHandler) getClusterDefaultCPUModel() (string, error) {
+	kubevirt, err := h.kubevirtCache.Get(util.HarvesterSystemNamespaceName, util.KubeVirtObjectName)
+	if err != nil {
+		return "", err
+	}
+	if kubevirt == nil || kubevirt.Spec.Configuration.CPUModel == "" {
+		return "", nil
+	}
+	return kubevirt.Spec.Configuration.CPUModel, nil
 }
 
 func (h *vmActionHandler) createVMBackup(vmName, vmNamespace string, input BackupInput) error {
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -48,6 +48,7 @@ import (
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/indexeres"
 	harvesterServer "github.com/harvester/harvester/pkg/server/http"
 	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
@@ -98,6 +99,7 @@ type vmActionHandler struct {
 	storageClassCache         ctlstoragev1.StorageClassCache
 	resourceQuotaClient       ctlharvesterv1.ResourceQuotaClient
 	clientSet                 kubernetes.Clientset
+	podCache                  ctlcorev1.PodCache
 }
 
 func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.ResponseBody, error) {
@@ -589,12 +591,12 @@ func (h *vmActionHandler) findMigratableNodes(rw http.ResponseWriter, namespace,
 }
 
 func (h *vmActionHandler) findMigratableNodesByVMI(vmi *kubevirtv1.VirtualMachineInstance) ([]string, error) {
-	nodeSelector, err := h.getNodeSelectorRequirementFromVMI(vmi)
+	nodeFilter, err := h.getNodeSelectorRequirementFromVMI(vmi)
 	if err != nil {
 		return nil, err
 	}
 
-	nodes, err := h.nodeCache.List(nodeSelector)
+	nodes, err := h.nodeCache.List(nodeFilter)
 	if err != nil || len(nodes) == 0 {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


ibrokethecloud:
PR in general looks good.

Can we please also check if a cluster wide default CPU model is set: https://kubevirt.io/user-guide/compute/virtual_hardware/#enabling-default-cluster-cpu-model

When one is set, even though the VM does not have a cpu model set, one is applied to the launcher pod, and will eventually be used for migration as well.

martindekov:
Thanks for the input @ibrokethecloud added in the latest commmit - when model is missing, look up default one, and if set use the selector from it, fixed the tests as well

pkg/api/vm/handler.go
```diff
@@ -622,25 +622,75 @@ func isDrained(node *corev1.Node) bool {
 }
 
 func (h *vmActionHandler) getNodeSelectorRequirementFromVMI(vmi *kubevirtv1.VirtualMachineInstance) (labels.Selector, error) {
-	if vmi == nil || vmi.Spec.Affinity == nil || vmi.Spec.Affinity.NodeAffinity == nil || vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
-		return labels.Everything(), nil
+	nodeSelector := labels.NewSelector()
+
+	if vmi != nil && vmi.Spec.Affinity != nil && vmi.Spec.Affinity.NodeAffinity != nil && vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
```
ihcsim: The CPU model and feature labels are added to the `virt-launcher` pod's `.spec.nodeSelector`, not the VMI's node affinity. Ultimately, it's the pod that gets scheduled to the node, not the VMI. Every time a VM is restarted, the node selector is wiped from the pod, while the VM spec remains unchanged.

pkg/api/vm/handler.go
```diff
@@ -622,25 +622,75 @@ func isDrained(node *corev1.Node) bool {
...
-			r, err := convertNodeSelectorRequirementToSelector(e)
-			if err != nil {
-				return nil, err
+	cpuModel, err := h.getCPUModel(vmi)
+	if err != nil {
+		return nil, err
+	}
+
+	if cpuModel != "" {
+		cpuModelKey := fmt.Sprintf("%s%s", kubevirtv1.CPUModelLabel, cpuModel)
```
ihcsim: You'll also have to account for the labels relevant to the `host-model` configuration. E.g., if I don't specify a named model in the VM's spec, KubeVirt picks the `host-model` setting, which leads to node selector that looks like:

```
k get po virt-launcher-vm-00-4wtx9 -ojsonpath='{.spec.nodeSelector}' | jq .
{
  "cpu-feature.node.kubevirt.io/arch-capabilities": "true",
  "cpu-feature.node.kubevirt.io/cmp_legacy": "true",
  "cpu-feature.node.kubevirt.io/gds-no": "true",
  "cpu-feature.node.kubevirt.io/hypervisor": "true",
  "cpu-feature.node.kubevirt.io/mds-no": "true",
  "cpu-feature.node.kubevirt.io/overflow-recov": "true",
  "cpu-feature.node.kubevirt.io/pschange-mc-no": "true",
  "cpu-feature.node.kubevirt.io/rdctl-no": "true",
  "cpu-feature.node.kubevirt.io/rfds-no": "true",
  "cpu-feature.node.kubevirt.io/skip-l1dfl-vmentry": "true",
  "cpu-feature.node.kubevirt.io/spec-ctrl": "true",
  "cpu-feature.node.kubevirt.io/ssbd": "true",
  "cpu-feature.node.kubevirt.io/stibp": "true",
  "cpu-feature.node.kubevirt.io/succor": "true",
  "cpu-feature.node.kubevirt.io/tsc-deadline": "true",
  "cpu-feature.node.kubevirt.io/tsc_adjust": "true",
  "cpu-feature.node.kubevirt.io/virt-ssbd": "true",
  "cpu-feature.node.kubevirt.io/x2apic": "true",
  "cpu-model-migration.node.kubevirt.io/EPYC-Genoa": "true",
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "sand-2.isim.dev",
  "kubevirt.io/schedulable": "true",
  "machine-type.node.kubevirt.io/pc-q35-9.2": "true"
}
```

If I specified a named model, the pod's node selector looks like:

```
k get po virt-launcher-vm-01-247bb -ojsonpath='{.spec.nodeSelector}' | jq .
{
  "cpu-model.node.kubevirt.io/EPYC": "true",
  "kubernetes.io/arch": "amd64",
  "kubernetes.io/hostname": "sand-2.isim.dev",
  "kubevirt.io/schedulable": "true",
  "machine-type.node.kubevirt.io/pc-q35-9.2": "true"
}
```

martindekov: I though subset of all the labels as per what the original issue described would cut it, meaning subset of the labels are truly relevant, isn't that the case?

If I have to encode all the above without reusing logic from kubevirt, directly porting methods where the logic is internal, this can become really fragile and we need to make sure we support proper format for bigger amount of labels. Do you suggest encoding all of the above is necessary to properly pick compatible hosts for migration?

pkg/api/vm/handler.go
martindekov: This simply inverted the original logic where in that case
```
if vmi == nil || vmi.Spec.Affinity == nil || vmi.Spec.Affinity.NodeAffinity == nil || vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil
```
we returned and in case it's not nil we effectively did: `convertNodeSelectorRequirementToSelector(e)`

This was changed in order to not return early, but rather fill up the node selectors. Why was it there in the first place if it's not relevant, I mean we did check the affinity originally, do you suggest the original logic of adding node selectors based on those fields was flawed and we need to revisit it as part of this PR?

pkg/api/vm/handler.go
martindekov: @ibrokethecloud Did I assume by mistake that the original issue mentioned only the three labels as per the description, or were those simply examples, but not the full list?

pkg/api/vm/handler.go
martindekov: So looked through the kubevirt codebase in attempt to find helper method to do this for me they have:
`controller.CurrentVMIPod(vmi, h.podIndexer)` but we don't have podIndexer we have incompatible cache.

Plan B would be to simply get the pod which is related to VMI and in case th epod exists:
```
    for key, value := range currentPod.Spec.NodeSelector {
        req, err := labels.NewRequirement(key, selection.Equals, []string{value})
        if err != nil {
            return nil, err
        }
        nodeSelector = nodeSelector.Add(*req)
    }
```

I think that way we simply copy paste the requirements from `pod.spec.nodeSelector` to seek the desired `nodeSelector` that way we don't care about the internals (specific constant key) and we seek what the pod wants. I think that way we can fix both this feedback and the below @ihcsim WDYT, would this be sufficient to cover all cases?

pkg/api/vm/handler.go
martindekov: I think this suggestion:
https://github.com/harvester/harvester/pull/9372#discussion_r2610476531
Would solve this concern as well, let me know

pkg/api/vm/handler.go
martindekov: Also if it's safe to directly translate all of the pod's nodeSelectors in attempt to filter potential nodes for migration

ihcsim: Yes, it makes sense to me to just copy the pod's node selector. All of them must be satisfied for a node to be deemed a valid migration target. You can find the pod via its `harvesterhci.io/vmName` label (the vmi and vm names are the same).

pkg/api/vm/handler.go
Yu-Jack: I think you can just copy the original node selector of the pod.

martindekov:
Updated the PR with the new logic where we simply bulid the label selector based on VMI pod's nodeSelector. Also updated the Description with the PR, how I tested the change e2e and updated the unit tests. PTAL, thanks!

cc @ibrokethecloud @ihcsim @Yu-Jack

pkg/api/vm/handler.go
```diff
@@ -622,23 +624,46 @@ func isDrained(node *corev1.Node) bool {
...
+
+	nodeFilter := labels.NewSelector()
+	for key, value := range latestPod.Spec.NodeSelector {
+		requirement, err := labels.NewRequirement(key, selection.Equals, []string{value})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create requirement for %s=%s: %w", key, value, err)
 		}
+		nodeFilter = nodeFilter.Add(*requirement)
 	}
+	return nodeFilter, nil
```
Yu-Jack, line 698: I just tested and realized that we should skip the `"kubernetes.io/hostname"`. Otherwise, it always returns the node where it's from.

For example:

This is pod node selector.

```yaml
nodeSelector:
  cpu-feature.node.kubevirt.io/amd-ssbd: "true"
  cpu-feature.node.kubevirt.io/amd-stibp: "true"
  cpu-feature.node.kubevirt.io/flush-l1d: "true"
  cpu-feature.node.kubevirt.io/gds-no: "true"
  cpu-feature.node.kubevirt.io/hypervisor: "true"
  cpu-feature.node.kubevirt.io/ibpb: "true"
  cpu-feature.node.kubevirt.io/ibrs: "true"
  cpu-feature.node.kubevirt.io/md-clear: "true"
  cpu-feature.node.kubevirt.io/movdir64b: "true"
  cpu-feature.node.kubevirt.io/movdiri: "true"
  cpu-feature.node.kubevirt.io/ss: "true"
  cpu-feature.node.kubevirt.io/stibp: "true"
  cpu-feature.node.kubevirt.io/tsc_adjust: "true"
  cpu-feature.node.kubevirt.io/vmx: "true"
  cpu-feature.node.kubevirt.io/vmx-activity-wait-sipi: "true"
  cpu-feature.node.kubevirt.io/vmx-tsc-scaling: "true"
  cpu-model-migration.node.kubevirt.io/SierraForest: "true"
  kubernetes.io/arch: amd64
  kubernetes.io/hostname: harvester2
  kubevirt.io/schedulable: "true"
  machine-type.node.kubevirt.io/pc-q35-9.2: "true"
```

When I was trying to migrate it, the option list was empty because it also used `kubernetes.io/hostname` as a filter.

This video did two things

- Created a new VM and migrated it.
- Migrated it again. -> The list was empty.

https://github.com/user-attachments/assets/1ea9fbb8-3218-4cc2-8819-a9e4b9cf3c56

Yu-Jack (changes requested):

pkg/api/vm/handler.go
martindekov: ok I will skip the  add if the filter is the hostname, nice catch Jack, thanks !

martindekov: Rebased and skipped this key `kubernetes.io/hostname`, you can resolve if the latest commit makes sense regarding this feedback

martindekov: I think leaving this open might be misleading, so will resolve the discussion as I fixed the feedback

w13915984028:
A question:

The current POD's nodeSelector has a couple of `kubevirt` related KV, this PR will use them to filter the target node, it has the assumpation that `kubevirt` will continue to use them as selector for migration target pod. But it might not be true all the time, `kubevirt` could migrate/add/remove the selector internally in the stage like `MigrationScheduling`/`MigrationPreparingTarget`.

Namely, Harvester is hard to make the same sheduling decision before forwarding the migration to kubevirt/k8s for final processing. There are two cases:

A. Harvester decides a node is target, but kubevirt/k8s refuses.

e.g. when kubevirt related feature gate is enabled/disabled on the fly, a new migration can apply different selector on the POD.

B. Harvester filters out one node, however kubevirt/k8s allows the vm to run on it.

more details: Harvester v170 supports the pausable-upgrade, user can set the node upgrade to be paused before drain and upgrade a node and do migration manually.

Suppose: A VM from v160 can have 10 selectors, but on v170, kubevirt merges them to 5. If user tries to migrate a VM on v160  pausable-upgrade stage manually, Harvester says none of other nodes are fit for. 

This apporach has potential risk to defeat the pausable-upgrade. It is one case I can think of atm.



The harm of B is more than A.

How do we consider this case? thanks.

```
nodeSelector:
  cpu-feature.node.kubevirt.io/amd-ssbd: "true"
  cpu-feature.node.kubevirt.io/amd-stibp: "true"
  cpu-feature.node.kubevirt.io/flush-l1d: "true"
  cpu-feature.node.kubevirt.io/gds-no: "true"
  cpu-feature.node.kubevirt.io/hypervisor: "true"
  cpu-feature.node.kubevirt.io/ibpb: "true"
...
```

w13915984028:
As Kubevirt's code is out of our control, Harvester might add a new stting like `UI VM Migration target node selection strategy` with value `strict` and `loose`.

The `strict` mode goes into new code; `loose` mode uses the original code.

No matter which one, Harvester is hard to predict precisely the final decision from kubevirt, we might add this to document.

martindekov:
I though we don't have situations where kubevirt's assigned VMI pod labels conflict with nodes and might not be listed if we adopt this new logic in this PR even though the node might be compatible for the vm to be migrated onto it. Thanks for the example @w13915984028 I was not aware that we can end up in such case.

@ihcsim @Yu-Jack Do you think we can go with the setting to toggle strict/lose as per what Jian suggested or as alternative pick specific selectors from the pod to be valued like in the original issue mentioning subset of 3 fields as per what @ibrokethecloud suggested:
```
  nodeSelector:
    cpu-model.node.kubevirt.io/EPYC-Rome: "true"
    kubernetes.io/arch: amd64
    kubevirt.io/schedulable: "true"
```

WDYT?

Yu-Jack:
I'm wondering if there's an existing pre-check script or method on the KubeVirt side that we could leverage, then we don't need to have `loose` or `strict` way.

> Suppose: A VM from v160 can have 10 selectors, but on v170, kubevirt merges them to 5. If user tries to migrate a VM on v160 pausable-upgrade stage manually, Harvester says none of other nodes are fit for.

This case should not happen. While upgrading to v170 from v160, those VMs will be evicted and live migrated with newer KubeVIrt version to another node, which the selectors will be 5. After the upgrade, the pre-check in this PR only check the 5 selectors.

> Harvester is hard to predict precisely the final decision from kubevirt, we might add this to document.

I also agree with this. So, it would be better to have a standard for live migration if KubeVirt can provide something like pre-check or function. If not, maybe we just choose not to make any changes and avoid us from getting more troubles of live migration.

w13915984028:
It looks using a new setting, which defaults with following KVs and allows users to tune on the fly, will be a good balance for now. The target node selector appends below KVs after the original processing.

```
    cpu-model.node.kubevirt.io/EPYC-Rome: "true"
    kubernetes.io/arch: amd64
    kubevirt.io/schedulable: "true"
```

ihcsim:
I think the issue on the node's CPU model and feature labels changing during upgrade doesn't nullify the solution this PR proposes. The issue already exists today. 

We know that after the first live migration, KubeVirt retains the node selectors on the `virt-launcher` pods of the same virtual machine during subsequent live migrations, until the virtual machine is restarted. The labels on the nodes may change only if some low level changes happen on the node or KubeVirt is upgraded (e.g., during upgrade). The only way for user to make these nodes suitable for live migration again is to restart the virtual machine so that KubeVirt can reset the node selectors on the `virt-launcher` pod of the virtual machine.

Harvester can only discover eligible target nodes for a virtual machine based on the node selector configuration of the _current_ `virt-launcher` pod of that virtual machine. These labels reflect the CPU host/feature configuration of the last known successful live migration. From that perspective, I think it's acceptable for the migration suitability of a target node to change, if changes are made to the node and/or KubeVirt. E.g., a node has 10 CPU feature labels in Harvester 1.6, but after upgrading to 1.7, `libvirt` reports to KubeVirt that this node now only have 5 CPU feature labels (for whatever reasons). Then the node is no longer an suitable target node for live migration of virtual machines that require the previous 10 CPU features.

@martindekov You will also need to account for the labels generated by the `host-model` setting as I described [above](https://github.com/harvester/harvester/pull/9372#discussion_r2608371730). This is the default setting that most Harvester users use. If a `virt-launcher` pod has some `cpu-feature.node.kubevirt.io` node selectors and we don't account for them during our node discovery, live migration to some of these discovered nodes can still fail even though they have `cpu-model-migration.node.kubevirt.io` .

ibrokethecloud:
in general the PR looks good however we need to fine tune it a bit more, the current implementation drops the original check where we calculate VMI node affinity criteria

https://github.com/harvester/harvester/pull/9372/files#diff-6a02f49ff2d9e428071b3a2d7c39eea15917d24b5e67997a714a7cb6535eec09L625

The VMI node affinity gets mapped to launcher pod's nodeAffinity.

We need to apply both the node affinity and node selectors to ensure correct filtering is applied to identify target nodes.

For example consider a sample VM which has a user defined node scheduling criteria

image

This maps to `.spec.template.spec.affinity.nodeAffinity` field

```
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/mac-address: '{"default":"b2:e1:ff:3f:52:ed"}'
    harvesterhci.io/volumeClaimTemplates: '[{"metadata":{"name":"migration-test-zone-disk-0-vymfh","annotations":{"harvesterhci.io/imageId":"default/image-6sb88"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"10Gi"}},"volumeMode":"Block","storageClassName":"longhorn-image-6sb88"}}]'
    network.harvesterhci.io/ips: '[]'
  creationTimestamp: "2026-01-15T23:20:39Z"
  finalizers:
  generation: 1
  labels:
    harvesterhci.io/os: linux
  name: migration-test-zone
  namespace: default
  resourceVersion: "1781131"
  uid: 7a16cb99-d38a-47e3-a3ab-445f6b060ea8
spec:
  runStrategy: RerunOnFailure
  template:
    metadata:
      annotations:
      creationTimestamp: null
      labels:
        harvesterhci.io/vmName: migration-test-zone
    spec:
      affinity:
        nodeAffinity:
            nodeSelectorTerms:
            - matchExpressions:
              - key: zone
                operator: In
                values:
                - dell
 ```

which maps to the pods `.spec.affinity.nodeAffinity` field

```
k get pod virt-launcher-migration-test-zone-7gj5v -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cni.projectcalico.org/containerID: 83f8996d591a2a827c882584dbec0636357107406481de17c831664472ec4174
    cni.projectcalico.org/podIP: 10.52.1.60/32
    cni.projectcalico.org/podIPs: 10.52.1.60/32
    descheduler.alpha.kubernetes.io/request-evict-only: ""
    istio.io/reroute-virtual-interfaces: k6t-eth0
    k8s.v1.cni.cncf.io/network-status: |-
      [{
          "name": "k8s-pod-network",
          "ips": [
              "10.52.1.60"
          ],
          "default": true,
          "dns": {}
      }]
    kubectl.kubernetes.io/default-container: compute
    kubevirt.io/domain: migration-test-zone
    kubevirt.io/migrationTransportUnix: "true"
    kubevirt.io/vm-generation: "1"
    post.hook.backup.velero.io/command: '["/usr/bin/virt-freezer", "--unfreeze", "--name",
      "migration-test-zone", "--namespace", "default"]'
    post.hook.backup.velero.io/container: compute
    pre.hook.backup.velero.io/command: '["/usr/bin/virt-freezer", "--freeze", "--name",
      "migration-test-zone", "--namespace", "default"]'
    pre.hook.backup.velero.io/container: compute
    traffic.sidecar.istio.io/kubevirtInterfaces: k6t-eth0
  creationTimestamp: "2026-01-15T23:20:39Z"
  generateName: virt-launcher-migration-test-zone-
  generation: 1
  labels:
    harvesterhci.io/vmName: migration-test-zone
    kubevirt.io: virt-launcher
    kubevirt.io/created-by: 0265a7d1-94a3-4c2c-96f8-73be31b99749
    kubevirt.io/nodeName: dell-140-tink-system
    vm.kubevirt.io/name: migration-test-zone
  name: virt-launcher-migration-test-zone-7gj5v
  namespace: default
  ownerReferences:
  - apiVersion: kubevirt.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: VirtualMachineInstance
    name: migration-test-zone
    uid: 0265a7d1-94a3-4c2c-96f8-73be31b99749
  resourceVersion: "1780934"
  uid: c6139f5d-df45-420d-ae46-8e68ff975cdd
spec:
  affinity:
    nodeAffinity:
        nodeSelectorTerms:
        - matchExpressions:
          - key: zone
            operator: In
            values:
            - dell
          - key: node-labeller.kubevirt.io/obsolete-host-model
            operator: DoesNotExist
```

In the current implementation we ignore this and only use the labels from `.spec.nodeSelector`.

This will result in nodes not meeting the original VM scheduling criterial being available in the list of nodes available from migration.

if would be nice to also have a unit test that covers this scenario.

**Comment by martindekov** [comment]
Thanks for the input folks I appreciate it! I will try to incorporate both comments and let you know if there is something I am missing.

File: pkg/api/vm/handler.go [hunk 4]
```diff
@@ -637,23 +639,112 @@ func isDrained(node *corev1.Node) bool {
...
+
+	requirement, err := labels.NewRequirement(req.Key, op, req.Values)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create requirement for %s %v %v: %w", req.Key, op, req.Values, err)
+	}
+	return requirement, nil
+}
+
+func (h *vmActionHandler) getLatestVMIPodOnCurrentNode(vmi *kubevirtv1.VirtualMachineInstance) (*corev1.Pod, error) {
+	vmiPods, err := h.podCache.GetByIndex(indexeres.PodByVMIUIDIndex, string(vmi.UID))
```
w13915984028: vmi object has `status.activePods`, could it be used to replace the current index?

```
  status:
    activePods:
      a5e9c672-e3ad-4f13-8ffe-ec1afe0ee2ba: xyz-03
    conditions:
```

martindekov: Yes, I can skip the index and use this and enter the logic of picking latest pod if map has more than 1 value. Good suggestion for optimization

w13915984028: Given the issue description https://github.com/harvester/harvester/issues/8797, I guess the scenario is on Harvester UI, where when user clicks `migrate` then a `migration target` window is poped.

Per current source code, only when a VM is running, there is no pending/ongoing migrations and some other conditions are met, then the related menus are rendered.

It could be even simplified: when `findMigratableNodesByVMI` is called, if the `len(vmi.status.activePods) != 1`, report error.

martindekov: Actually reverting my statement looking at how we set that field in kubevirt:
https://github.com/kubevirt/kubevirt/blob/release-1.6/pkg/virt-controller/watch/vmi/lifecycle.go#L879-L898

this is UID, hence this indexer would be used, it's not a pod name which we could call get on for each element in this map.

Regarding:
>It could be even simplified: when findMigratableNodesByVMI is called, if the len(vmi.status.activePods) != 1, report error.

You are right, if you have 2 pods, migration is currently either about to happen or did not finished yet so migration is not an option at this moment so directly returning error as opposed to "picking latest pod" is the better approach. I can go ahead and do that.

One question is, is that field rejectable on a webhook level? Should we do that rejection there ? I have seen the migraton field pop in and out but not sure the conditions of this appearance

w13915984028: BTW: Use k8s list from target vm's  `namespace` and with vm name label `vm.kubevirt.io/name` / `harvesterhci.io/vmName` / `kubevirt.io/created-by`  as selector, to filter vmi's PODs directly in a single API call, is an alternative.

It is expensive to maintain a new pod index which is only called when user clicks UI menu `migrate`. The above `list` operation on demand gives a better overall performance.

```
    labels:
...
      harvesterhci.io/vmName: ...-q8p9x
...
      kubevirt.io: virt-launcher
      kubevirt.io/created-by: 2ed7c971-15c8-4385-abb3-37158c0458cd
      kubevirt.io/nodeName: ...-03
      nodepool.harvesterhci.io/name: pool1-vxm55
      vanguard.tridentcloud.dev/control-plane: "true"
      vm.kubevirt.io/name: ...-q8p9x  // exact copy of vm's name
```

martindekov: Will remove the indexer and list with label as selector using `kubevirt.io/created-by` as I currently do in the indexer, effectively moving the logic from indexer to locally in this method. Would need some time to build and test e2e will push the last commit with the suggestions above and tests.

martindekov: So using:
```
  status:
    activePods:
      a5e9c672-e3ad-4f13-8ffe-ec1afe0ee2ba: xyz-03
    conditions:
```

Won't cut it simply as after migration this is what's left:
```
  activePods:
    c59bc636-7f09-49ef-a769-6ffbc5e74e6f: node-1
    f7e5e033-1acd-4e8e-abde-9084e99a0163: node-2
```
So 2 pods in that state:
```
node-1:/home/rancher # kubectl get pods -A | grep virt-launcher
default                     virt-launcher-vm-1-4xnv6                                               2/2     Running     0             11m
default                     virt-launcher-vm-1-djjzk                                               0/2     Completed   0             11m
```

So I won't rely on active pods amount, but rather will list pods and check for their statuses.

martindekov:
@w13915984028 @ibrokethecloud @ihcsim @Yu-Jack Team if you can please do a second review, updated testing done section where I validated:
1. Migration does not appear while migration is running
2. Affinity is respected 

Attached pictures and explained basic flow of the change. Let me know your thoughts

pkg/api/vm/handler.go
martindekov: I addressed all the comments here so resolving as per: https://github.com/harvester/harvester/pull/9372#issuecomment-3773214760

pkg/api/vm/handler.go
```diff
@@ -637,23 +638,120 @@ func isDrained(node *corev1.Node) bool {
...
 		return labels.Everything(), nil
 	}
 
-	nodeSelector := labels.NewSelector()
-	terms := vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+	latestPod, err := h.getVMIPod(vmi)
+	if err != nil {
+		return labels.Everything(), err
+	}
+	if latestPod == nil {
```
w13915984028: per definition of getVMIPod, the `latestPod == nil` seems not possible, after the above error case processing

w13915984028 (dismissed):
Thanks @martindekov

pkg/api/vm/handler.go
martindekov: Good catch Jian, removed the check and renamed to vmiPod I think it suits better

w13915984028 (dismissed):
LGTM, thanks.

pkg/api/vm/handler.go
```diff
@@ -637,23 +638,117 @@ func isDrained(node *corev1.Node) bool {
...
-	if vmi == nil || vmi.Spec.Affinity == nil || vmi.Spec.Affinity.NodeAffinity == nil || vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+	if vmi == nil {
 		return labels.Everything(), nil
 	}
 
-	nodeSelector := labels.NewSelector()
-	terms := vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+	vmiPod, err := h.getVMIPod(vmi)
+	if err != nil {
+		return labels.Everything(), err
```
ihcsim: i think returning `labels.Nothing()` here makes more sense.

Yu-Jack (dismissed):
Thanks!

ihcsim (dismissed):

pkg/api/vm/handler.go
```diff
@@ -637,23 +638,117 @@ func isDrained(node *corev1.Node) bool {
...
+	}
+
+	nodeFilter := labels.NewSelector()
+
+	for key, value := range vmiPod.Spec.NodeSelector {
+		if key == corev1.LabelHostname {
```
ibrokethecloud, line 653: if the pod as a hostName node selector, then it is becuase the VM was pinned to a host
image
it may be better to return `labels.Nothing()` in this case as the VM cannot be live migrated.

ibrokethecloud:
lgtm apart from the minor nit. if we find hostname in node selector then VM is not live migratable and we should return no valid nodes matching the migration criteria.

pkg/api/vm/handler.go
Yu-Jack: Just need to be careful this case https://github.com/harvester/harvester/pull/9372#discussion_r2608371730. When you migrate the VM to other node, the hostname will appear in node selector. I'm not sure whether it's different from this case.

martindekov: So I performed some testing.
1. Created VM
2. By default VM is on - `Run virtual machine on any available node` so migration button was present
3. Enabled VM to be hosted only on the current node - `Run virtual machine on specific node - (Live migration is not supported)`
4. Migration button dissapeared
5. Reverted VM to original state of `Run virtual machine on any available node`
6. Migrated VM to another node (Jack's original comment which drove this logic migration will add this hostname)
7. After migration enabled -  `Run virtual machine on specific node - (Live migration is not supported)`
8. Migration button dissapeared

I guess with continue as opposed to labels.Nothing() the feature of binding vm to specific node works as intended and migration is disbled with that case. I will add testing section with that test as well.

Do you think the test validates the functionality and resolves the concern without changes to the current logic?

martindekov: Added to the testing done section - `Validate Node Scheduling works`

pkg/api/vm/handler.go
martindekov: Good catch Ivan, returning labels.Nothing now

martindekov: Also test section covered this change I build image with it inside - `Validate Node Scheduling works`

martindekov:
Found bug with Node Scheduling tab while testing the change: https://github.com/harvester/harvester/issues/7583#issuecomment-3790256917

pkg/api/vm/handler.go
```diff
@@ -637,23 +638,117 @@ func isDrained(node *corev1.Node) bool {
...
-	if vmi == nil || vmi.Spec.Affinity == nil || vmi.Spec.Affinity.NodeAffinity == nil || vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+	if vmi == nil {
 		return labels.Everything(), nil
 	}
 
-	nodeSelector := labels.NewSelector()
-	terms := vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+	vmiPod, err := h.getVMIPod(vmi)
+	if err != nil {
+		return labels.Nothing(), err
```
ihcsim, line 647: I noticed the previous `labels.Everything()` too during my first review, but figure it won't matter because the caller ends up returning the `err` anyway. Thanks for changing it.

pkg/api/vm/handler.go
ihcsim: So yeah, there are two ways that a pod ends up with the hostname selector:

* user adds it via the vm ui section (i.e. edit as yaml, node scheduling)
* kubevirt [copies it from the vmim to the target pod](https://github.com/kubevirt/kubevirt/blob/3d3ad57915a769a3f48bc9004f6a1f8d82ca022e/pkg/virt-controller/watch/migration/migration.go#L931)

In the first case, our current handler wouldn't have reached this code path during migration.

ihcsim (approved):

pkg/api/vm/handler.go
martindekov: During my testing I hit errors before some commits and it just stops the migration with an error in the corner of the UI, so I think indeed might not be too relevant, but did it anyway. Good catch!

pkg/api/vm/handler.go
martindekov: Thanks for the input all, I will leave this open as it's more general question and if it make sense and as per the testing Gaurav can resolve the discussion if no other concerns

Issue #9373, opened 2025-10-27
[BUG] Modify ManagedChart break fleet bundle


FrankYang0529:
### Describe the Bug

When modifying ManagedChart, the bundle shows error like: `failed to get options secret for bundledeployment cluster-fleet-local-local-xxx/mcc-harvester in fleet bundle`. The root cause is that we deny delete request for BundleDeployment in https://github.com/harvester/harvester/issues/2107. When updating fleet Bundle, it will remove BundleDeployment and Secret. However, the BundleDeployment cannot be removed. When fleet-controller reconciles fleet Bundle, it finds there is `valuesHash` without related Secret, so reporting the error.

Related issues: https://github.com/harvester/harvester/issues/9349, https://github.com/rancher/fleet/issues/4144

### To Reproduce

1. Install a v1.7.0-head cluster.
2. Update values in `fleet-local/harvester` ManagedChart.
3. See error in related fleet Bundle.

### Expected Behavior

The fleet can reconcile Bundle without error.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.7.0-head
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): KVM
- Rancher version: v2.13.0-alpha2

#### Remove valuesHash

1. Find the BundleDeployment and remove `valuesHash`.
2. Add a random annotation to Bundle.
3. Wait few seconds, the fleet-controller will reconcile BundleDeployment and we can see a new secret.

#### Create the secret

1. Find the Bundle and get base64 encoded values. (`kg bundle -n fleet-local mcc-harvester -o json | jq .spec.helm.values -c -j | base64 -w 0`)
2. Create a secret and use step 1 output in `stagedValues`.

TachunLin:
Verified **failed** when upgrade from `v1.6.1` to `v1.7.0-rc1`. Move back to implement


### Result

* When upgrade from v1.6.1 to v1.7.0-rc1
  <img width="1598" height="781" alt="Image" src="https://github.com/user-attachments/assets/e6c12799-8df7-462c-99ff-7bb9b18f6f2d" />

* The upgrade process failed, stuck in the `apply manifest` stage
  <img width="1597" height="786" alt="Image" src="https://github.com/user-attachments/assets/08173b69-ce60-4cfa-9a53-a86b0799aeab" />

* Node1 is Not ready and kept in `cordened` state
[...]

PR #9377, opened 2025-10-28
feat: remove bundledeployment webhook

FrankYang0529:
#### Problem:
In current design, the BundleDeployment cannot be removed. This makes ManagedChart upgrade fails.

#### Solution:
Remove BundleDeployment to unblock ManagedChart upgrade.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9373

#### Test plan:
1. Create a v1.6.1 cluster.
2. Upgrade to this branch without error.


ihcsim (approved):

FrankYang0529:
Close the PR since we prefer another fix https://github.com/harvester/harvester/pull/9378.

PR #9378, opened 2025-10-28
feat: patch fleet image tag

FrankYang0529:
#### Problem:
In fleet v0.14.0, it introduces secret to keep helm values. However, if the secret is missing, we cannot upgrade ManagedChart.

#### Solution:
Set fleet to v0.13.4 to avoid secret missing error.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9373

#### Test plan:
1. Create a v1.6.1 cluster.
2. Upgrade to this branch without error.

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -571,6 +571,23 @@ upgrade_rancher() {
     exit 1
   fi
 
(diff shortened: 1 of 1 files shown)


ibrokethecloud (approved):

Vicente-Cheng (approved):
verified upgrade with 3 node cluster with workload

everything works fine during upgrade.

PR #9385, opened 2025-10-29
feat: patch prefer-no-eviction if vmi is non-migratable

FrankYang0529:
#### Problem:
Currently, we use `LiveMigrateIfPossible` for all VMs. Users cannot update it via UI. If we evict VM with `LiveMigrateIfPossible`, it will be shutdown.

#### Solution:
Use `descheduler.alpha.kubernetes.io/prefer-no-eviction=true` annotation to avoid evict non-migratable VM.

#### Related Issue(s):

#### Test plan:
1. Create a VM with PCI device.
2. Check there is `descheduler.alpha.kubernetes.io/prefer-no-eviction=true` on the VM.

### DIFF
--- a/pkg/controller/master/virtualmachine/vmi_descheduler_controller.go
+++ b/pkg/controller/master/virtualmachine/vmi_descheduler_controller.go
@@ -1,0 +1,50 @@
+package virtualmachine
+
+import (
+	"reflect"
+
+	"github.com/sirupsen/logrus"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
+	vmv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+)
+
+const (
+	deschedulerPreferNoEvictionAnnotationKey = "descheduler.alpha.kubernetes.io/prefer-no-eviction"
+)
+
+type VMIDeschedulerController struct {
+	vmCache  vmv1.VirtualMachineCache
+	vmClient vmv1.VirtualMachineClient
+}
+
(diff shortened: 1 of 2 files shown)


Vicente-Cheng (dismissed):
lgtm, thanks!

wheatdog (dismissed):

Vicente-Cheng (dismissed):

ibrokethecloud:
we also need to remove the annotation if VM becomes migratable, for example removal of a host device etc. In the current logic the VM will stay non migratable even though kubevirt will mark the VM as migratable

pkg/controller/master/virtualmachine/vmi_descheduler_controller.go
```diff
@@ -0,0 +1,50 @@
...
+			vmCopy.Annotations = make(map[string]string)
+		}
+		vmCopy.Annotations[deschedulerPreferNoEvictionAnnotationKey] = "true"
+		if !reflect.DeepEqual(vm.Annotations, vmCopy.Annotations) {
+			_, err = h.vmClient.Update(vmCopy)
+			if err != nil {
+				return vmi, err
+			}
+		}
+	}
```
ibrokethecloud, line 53:
```suggestion
	if !vmi.IsMigratable() {
		logrus.Infof("VM %s/%s is non-migratable, skipping descheduling", vm.Namespace, vm.Name)
		vmCopy := vm.DeepCopy()
		if vmCopy.Annotations == nil {
			vmCopy.Annotations = make(map[string]string)
		}
		vmCopy.Annotations[deschedulerPreferNoEvictionAnnotationKey] = "true"
	} else {
        delete(vmCopy.Annotations, deschedulerPreferNoEvictionAnnotationKey)
    }

    if !reflect.DeepEqual(vm.Annotations, vmCopy.Annotations) {
        return h.vmClient.Update(vmCopy)
    }
```

ibrokethecloud (approved):

PR #9389, opened 2025-10-30
feat: disable descheduler during upgrade

FrankYang0529:
#### Solution:
Disable descheduler during upgrade to avoid unexpected eviction.

#### Related Issue(s):

#### Test plan:
2. Enable descheduler addon.
3. Upgrade to this branch.
4. Check the descheduler addon is disabled before node upgrade.
5. Check the descheduler addon is enabled after upgrade complete.

### DIFF
--- a/pkg/controller/master/upgrade/addon_controller.go
+++ b/pkg/controller/master/upgrade/addon_controller.go
@@ -17,7 +17,7 @@ type addonHandler struct {
 }
 
 func (h *addonHandler) OnChanged(_ string, addon *harvesterv1.Addon) (*harvesterv1.Addon, error) {
-	if addon == nil || addon.DeletionTimestamp != nil || addon.Name != util.DeschedulerName || addon.Labels == nil {
+	if addon == nil || addon.DeletionTimestamp != nil || addon.Name != util.DeschedulerName || addon.Labels == nil || !addon.Spec.Enabled {
 		return addon, nil
 	}
 
--- a/pkg/controller/master/upgrade/addon_controller.go
+++ b/pkg/controller/master/upgrade/addon_controller.go
@@ -1,0 +1,42 @@
+package upgrade
+
+import (
+	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+type addonHandler struct {
+	namespace    string
+	upgradeCache ctlharvesterv1.UpgradeCache
+	addonClient  ctlharvesterv1.AddonClient
+	addonCache   ctlharvesterv1.AddonCache
+}
+
+func (h *addonHandler) OnChanged(_ string, addon *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+	if addon == nil || addon.DeletionTimestamp != nil || addon.Name != util.DeschedulerName || addon.Labels == nil {
+		return addon, nil
+	}
+
+	upgradeControllerLock.Lock()
+	defer upgradeControllerLock.Unlock()
+
+	upgrade, err := h.upgradeCache.Get(upgradeNamespace, addon.Labels[harvesterUpgradeLabel])
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+
+	if upgrade != nil && upgrade.Annotations != nil && upgrade.Annotations[reenableDeschedulerAddonAnnotation] == "true" {
+		toUpdate := addon.DeepCopy()
+		logrus.Infof("Disabling %s addon during upgrade", toUpdate.Name)
+		toUpdate.Spec.Enabled = false
+		if _, err := h.addonClient.Update(toUpdate); err != nil {
+			return nil, err
+		}
+	}
+
+	return addon, nil
+}
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -74,7 +74,9 @@ const (
 	longhornSettingsRestoredAnnotation  = "harvesterhci.io/longhorn-settings-restored"
 	imageCleanupPlanCompletedAnnotation = "harvesterhci.io/image-cleanup-plan-completed"
 	skipVersionCheckAnnotation          = "harvesterhci.io/skip-version-check"
-	defaultImagePreloadConcurrency      = 1
+	reenableDeschedulerAddonAnnotation  = "harvesterhci.io/reenable-descheduler-addon"
+
+	defaultImagePreloadConcurrency = 1
 
 	vmReady condition.Cond = "Ready"
 )
@@ -94,6 +96,8 @@ type upgradeHandler struct {
 	versionCache      ctlharvesterv1.VersionCache
 	planClient        upgradectlv1.PlanClient
 	planCache         upgradectlv1.PlanCache
+	addonClient       ctlharvesterv1.AddonClient
+	addonCache        ctlharvesterv1.AddonCache
 
 	managedChartCache  mgmtv3.ManagedChartCache
 	managedChartClient mgmtv3.ManagedChartClient
@@ -258,6 +262,10 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			return h.upgradeClient.Update(toUpdate)
 		}
 
+		if err := h.reenableAddons(upgrade); err != nil {
+			return nil, err
+		}
+
 		return upgrade, nil
 	}
 
@@ -383,6 +391,9 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 				}
 			}
 
+			if err := h.addUpgradeLabelToDeschedulerAddons(toUpdate); err != nil {
+				return nil, err
+			}
 			// go with RKE2 pre-drain/post-drain hooks
 			logrus.Infof("Start upgrading Kubernetes runtime to %s", info.Release.Kubernetes)
 			if err := h.upgradeKubernetes(info.Release.Kubernetes); err != nil {
@@ -528,6 +539,9 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) (
 		}
 	}
 
+	if err := h.reenableAddons(upgrade); err != nil {
+		return nil, err
+	}
 	return upgrade, h.resumeManagedCharts()
 }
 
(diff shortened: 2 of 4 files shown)


pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -864,3 +873,44 @@ func (h *upgradeHandler) checkLogReadyCondition(upgrade *harvesterv1.Upgrade) (*
...
+	}
+	if reenableStr == "true" {
+		addon, err := h.addonCache.Get(util.KubeSystemNamespace, util.DeschedulerAddon)
+		if err != nil {
+			return err
+		}
+
+		if !addon.Spec.Enabled {
+			logrus.Info("Re-enabling descheduler addon after upgrade")
+			toUpdate := addon.DeepCopy()
```
w13915984028: remove the annotation `AnnotationReenableDeschedulerAddon`

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -864,3 +873,44 @@ func (h *upgradeHandler) checkLogReadyCondition(upgrade *harvesterv1.Upgrade) (*
...
+		toUpdate := addon.DeepCopy()
+		toUpdate.Spec.Enabled = false
+		if _, err := h.addonClient.Update(toUpdate); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (h *upgradeHandler) reenableAddons(upgrade *harvesterv1.Upgrade) error {
```
w13915984028, line 900: reenableAddons seems need to be called on `onDelete` path in function `cleanup`

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -864,3 +873,44 @@ func (h *upgradeHandler) checkLogReadyCondition(upgrade *harvesterv1.Upgrade) (*
...
+	reenableStr, ok := upgrade.Annotations[util.AnnotationReenableDeschedulerAddon]
+	if !ok {
+		return nil
+	}
+	if reenableStr == "true" {
+		addon, err := h.addonCache.Get(util.KubeSystemNamespace, util.DeschedulerAddon)
+		if err != nil {
```
w13915984028: skip isNotFound error

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -864,3 +873,44 @@ func (h *upgradeHandler) checkLogReadyCondition(upgrade *harvesterv1.Upgrade) (*
...
+
+func (h *upgradeHandler) disableAddons(upgrade *harvesterv1.Upgrade) error {
+	addon, err := h.addonCache.Get(util.KubeSystemNamespace, util.DeschedulerAddon)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	if addon != nil && addon.Spec.Enabled {
+		logrus.Info("Descheduler addon is enabled, disabling it for upgrade")
+		upgrade.Annotations[util.AnnotationReenableDeschedulerAddon] = "true"
```
w13915984028: the annotation `AnnotationReenableDeschedulerAddon` seems need to be put on `addon` object with the value of upgrade name/UID, at least more than `upgrade` object, as there is case:

`addon object is disabled successfully, then upgrade object is failed to update` (a couple of controllers might update the `upgrade` object in parallel, and when reconciller comes again, upgrade controller only finds the addon is disabled

updating 2+ objects on one controller logic, will face such races

w13915984028 (commented):
A few questions, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: I think it's fine to keep `AnnotationReenableDeschedulerAddon`.

w13915984028: upgrade controller should only do once `enable` action if the annotation is there, and then remove annotation

otherweise, if user diable it, the controller will enable again to battle with user's operation

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: We already have lock for upgrade related controllers, so race condition should not exist here.

https://github.com/harvester/harvester/blob/3e533b9c62af83c1275079805ac5db96ccb57ace/pkg/controller/master/upgrade/upgrade_controller.go#L122-L123

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: Updated it.

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: Updated it.

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: Updated it.

pkg/controller/master/upgrade/upgrade_controller.go
w13915984028: even with the locker, it still does not mean both CRD object updating are always successful in the mean time

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -864,3 +876,49 @@ func (h *upgradeHandler) checkLogReadyCondition(upgrade *harvesterv1.Upgrade) (*
...
+	}
+	return nil
+}
+
+func (h *upgradeHandler) reenableAddons(upgrade *harvesterv1.Upgrade) error {
+	reenableStr, ok := upgrade.Annotations[util.AnnotationReenableDeschedulerAddon]
+	if !ok {
+		return nil
+	}
+	if reenableStr == "true" {
```
w13915984028:
```
	if reenableStr != "true" {
		return nil
	}

// following code in reenable path

```

pkg/controller/master/upgrade/upgrade_controller.go
w13915984028: btw, image cleaner uses `imageCleanupPlanCompletedAnnotation` to mark the action is done

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: Updated it.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -864,3 +878,60 @@ func (h *upgradeHandler) checkLogReadyCondition(upgrade *harvesterv1.Upgrade) (*
...
+		return err
+	}
+
+	if addon != nil && addon.Spec.Enabled {
+		logrus.Info("Adding upgrade label to descheduler addon")
+		upgrade.Annotations[reenableDeschedulerAddonAnnotation] = "true"
+
+		toUpdate := addon.DeepCopy()
+		if toUpdate.Labels == nil {
+			toUpdate.Labels = make(map[string]string)
```
w13915984028, line 890: As here the addon labels has been updated with upgradeLabel, could we set the `add.Spec.Enabled = false` directly? Thus the `addon_controller` is not necessary; and it ensures the `disable` action is kicked only once.  

The `addon_controller` might reconcile if user clicks `eanble` addon in the upgrade.

w13915984028 (commented):
The doubt on `addon_controller`, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: As you mentioned in https://github.com/harvester/harvester/pull/9389#discussion_r2480665329, even if the controller updates addon successfully, it doesn't mean update upgrade successfully. For example,

1. Addon is turned off cause of the upgrade.
2. Upgrade doesn't update successfully with `"harvesterhci.io/reenable-descheduler-addon"` annotation.
3. In the next reconciliation, the upgrade cannot add the annotation again.

In current implementation, I make sure the addon and upgrade have required label/annotation before turning off the addon.

w13915984028: Give me some time to write an explanation of the differences between  upgrade controller and addon controller, before merging this PR. thanks.

w13915984028: key differences:

upgrade name/id is writen to addon CR together with enable/disable to avoid the case where addon CR is updated but upgrade CR is failed to update, and upgrade controller writes to addon object with the status machine control, enable/disable action is executed at most once.  

the addon controller has no context of upgrade status machine, it might race with user operation

image
image

FrankYang0529: During upgrade, we don't want user to enable the descheduler. This is to avoid unexpected migration.

pkg/controller/master/upgrade/addon_controller.go
```diff
@@ -0,0 +1,42 @@
...
+
+type addonHandler struct {
+	namespace    string
+	upgradeCache ctlharvesterv1.UpgradeCache
+	addonClient  ctlharvesterv1.AddonClient
+	addonCache   ctlharvesterv1.AddonCache
+}
+
+func (h *addonHandler) OnChanged(_ string, addon *harvesterv1.Addon) (*harvesterv1.Addon, error) {
+	if addon == nil || addon.DeletionTimestamp != nil || addon.Name != util.DeschedulerAddon || addon.Labels == nil {
```
w13915984028: add one more condition:

`|| !addon.Spec.Enabled`

w13915984028 (dismissed):
LGTM, thanks.

w13915984028 (dismissed):
LGTM, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -258,6 +262,10 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 			return h.upgradeClient.Update(toUpdate)
 		}
 
+		if err := h.reenableAddons(upgrade); err != nil {
```
Vicente-Cheng: I noticed we have two times to re-enable the descheduler.
Is this not enough if we only re-enable with the corresponding operation during the upgrade successed?

If so, maybe we could remove the re-enable on the cleanup caller.

Vicente-Cheng (commented):
just one nits, overall lgtm

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: I think we can keep the one in `cleanup` function and remove another one in `OnChange` function. For `cleanup`, it's the last step for both successful or failure path, so we can cover all cases. When users remove Upgrade after it finishes, the controller calls `cleanup` function again. This step is safe, because we remove `harvesterhci.io/reenableDeschedulerAddon` annotation in `reenableAddons` function. Even if the function is called again, it's no side effect.

ibrokethecloud (dismissed):
lgtm. thanks.

pkg/controller/master/upgrade/upgrade_controller.go
Vicente-Cheng: makes sense. Thanks for the explaination.

Vicente-Cheng (approved):

PR #9392, opened 2025-10-30
fix: storage class name conflict

DamiaSan:
#### Problem:
We use `longhorn-<vmimage name>` as storage class name. The StorageClass doesn't have namesapce, but vmimage has. If there are same vmimage name in different namespace, storage class name will be conflicts.

#### Solution:
Use `namespace` and image `UID` as name format.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/5165

- - -

# Test plan:

### Case 1: Reproducer from #5165
1. Create two VirtualMachineImages: `default/test-image` and `harvester-public/test-image`
2. After the two VirtualMachineImages are ready, delete `default/test-image`
3. Create a VirtualMachine from VirtualMachineImage `harvester-public/test-image`

### Case 1: VM basic operations (create / backup /restore) without error.
1. Create a VMImage.
2. Create a VM without error.
3.. Write some data.
4. Setup backup-target.
5. Backup the VM.
6. Restore a new VM. New VM can start without error.

### Case 2: VMImage with a same name in different namespaces can work
1. Create a VMImage `thisistest` in `default` and `harvester-public` namespaces.
2. Both images can be created without error.

### Case 3: An old VMImage can work without error.
1. Create a Harvester cluster v1.4.0.
2. Create a VMImage.
3. Replace harvester and harvester-webhook deployment with this PR.
4. The VMImage can work without error.

### Case 4: Create VM Template without error
1. Follow case 1.
2. Generate Template with data.
3. After template is ready, launch a new VM and we can see default image name starts with `templateversion-`. New VM can start without error and data is in the new VM.

### Case 1: Guest Cluster with Reproducer from #5165
1. Install Rancher, register Harvester cluster in Rancher
2. Create two VirtualMachineImages: `default/test-image` and `harvester-public/test-image`
3. After the two VirtualMachineImages are ready, delete `default/test-image`
4. Create a new guest cluster on Harvester from VirtualMachineImage `harvester-public/test-image`

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -48,6 +48,7 @@ import (
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	vmicommon "github.com/harvester/harvester/pkg/image/common"
 	harvesterServer "github.com/harvester/harvester/pkg/server/http"
 	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
@@ -74,6 +75,7 @@ type vmActionHandler struct {
 	clientSet                 kubernetes.Interface
 	virtRestClient            rest.Interface
 	virtSubresourceRestClient rest.Interface
+	vmio                      vmicommon.VMIOperator
 
 	backupClient            ctlharvesterv1.VirtualMachineBackupClient
 	datavolumeClient        ctlcdiv1.DataVolumeClient
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -28,7 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	k8stypes "k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/kubernetes"
@@ -37,7 +36,6 @@ import (
 	k8svolumehelpers "k8s.io/cloud-provider/volume/helpers"
 	"k8s.io/utils/ptr"
 	kubevirtv1 "kubevirt.io/api/core/v1"
-	cdicommon "kubevirt.io/containerized-data-importer/pkg/controller/common"
 	kubevirtmultus "kubevirt.io/kubevirt/pkg/network/multus"
 	kubevirtutil "kubevirt.io/kubevirt/pkg/virt-operator/util"
 
@@ -912,173 +910,6 @@ func getMigrationUID(vmi *kubevirtv1.VirtualMachineInstance) string {
 	return ""
 }
 
-// createTemplate creates a template and version that are derived from the given virtual machine.
-func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTemplateInput) (err error) {
-	var (
-		vm         *kubevirtv1.VirtualMachine
-		keyPairIDs []string
-		vmt        *harvesterv1.VirtualMachineTemplate
-		vmtv       *harvesterv1.VirtualMachineTemplateVersion
-	)
-	defer func() {
-		if err != nil {
-			logrus.WithError(err).WithFields(logrus.Fields{
-				"namespace": namespace,
-				"name":      name,
-			}).Error("Failed to create template")
-
-			if vmtv != nil {
-				// other resources have owner reference to vmtv, so we only need to delete vmtv
-				if cleanupErr := h.vmTemplateVersionClient.Delete(vmtv.Namespace, vmtv.Name, &metav1.DeleteOptions{}); cleanupErr != nil {
-					logrus.WithError(cleanupErr).WithFields(logrus.Fields{
-						"namespace": vmtv.Namespace,
-						"name":      vmtv.Name,
-					}).Error("Failed to delete vm template version")
-					err = fmt.Errorf("failed to delete vm template version, cleanup error: %s, original error: %w", cleanupErr.Error(), err)
-				}
-			}
-		}
-	}()
-	vm, err = h.vmCache.Get(namespace, name)
-	if err != nil {
-		return err
-	}
-
-	keyPairIDs, err = getSSHKeysFromVMITemplateSpec(vm.Spec.Template)
-	if err != nil {
-		return err
-	}
-
-	vmtvName := fmt.Sprintf("%s-%s", input.Name, rand.String(5))
-	vmSourceSpec := h.sanitizeVirtualMachineForTemplateVersion(vmtvName, vm)
-
-	vmt, err = h.vmTemplateClient.Create(
-		&harvesterv1.VirtualMachineTemplate{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      input.Name,
-				Namespace: namespace,
-			},
-			Spec: harvesterv1.VirtualMachineTemplateSpec{
-				Description: input.Description,
-			},
-		})
-	if err != nil {
-		return err
-	}
-
-	vmID := fmt.Sprintf("%s/%s", vmt.Namespace, vmt.Name)
-
-	vmtv, err = h.vmTemplateVersionClient.Create(
-		&harvesterv1.VirtualMachineTemplateVersion{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      vmtvName,
-				Namespace: namespace,
-			},
-			Spec: harvesterv1.VirtualMachineTemplateVersionSpec{
-				TemplateID:  vmID,
-				Description: fmt.Sprintf("Template drived from virtual machine [%s]", vmID),
-				VM:          vmSourceSpec,
-				KeyPairIDs:  keyPairIDs,
-			},
-		})
-	if err != nil {
-		return err
-	}
-
-	if input.WithData {
-		var (
-			pvcMap             map[string]corev1.PersistentVolumeClaim
-			pvcStorageClassMap map[string]string
-			vmImageMap         map[string]harvesterv1.VirtualMachineImage
-		)
-		pvcMap, pvcStorageClassMap, err = h.getPVCStorageClassMap(vm)
-		if err != nil {
-			return err
-		}
-		vmImageMap, err = h.createVMImages(vmtv, vm, pvcStorageClassMap)
-		if err != nil {
-			return err
-		}
-		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			vmtv, err = h.vmTemplateVersionClient.Get(vmtv.Namespace, vmtv.Name, metav1.GetOptions{})
-			if err != nil {
-				return err
-			}
-			return h.sanitizeVolumes(vmtv, pvcMap, vmImageMap)
-		})
-		if err != nil {
-			return err
-		}
-	}
-
-	return h.createSecrets(vmtv, vm)
-}
-
-func (h *vmActionHandler) createSecrets(templateVersion *harvesterv1.VirtualMachineTemplateVersion, vm *kubevirtv1.VirtualMachine) error {
-	for index, credential := range vm.Spec.Template.Spec.AccessCredentials {
-		if sshPublicKey := credential.SSHPublicKey; sshPublicKey != nil && sshPublicKey.Source.Secret != nil {
-			toCreateSecretName := getTemplateVersionSSHPublicKeySecretName(templateVersion.Name, index)
-			if err := h.copySecret(sshPublicKey.Source.Secret.SecretName, toCreateSecretName, templateVersion); err != nil {
-				return err
-			}
-		}
-		if userPassword := credential.UserPassword; userPassword != nil && userPassword.Source.Secret != nil {
-			toCreateSecretName := getTemplateVersionUserPasswordSecretName(templateVersion.Name, index)
-			if err := h.copySecret(userPassword.Source.Secret.SecretName, toCreateSecretName, templateVersion); err != nil {
-				return err
-			}
-		}
-	}
-	for _, volume := range vm.Spec.Template.Spec.Volumes {
-		if volume.CloudInitNoCloud == nil {
-			continue
-		}
-		if volume.CloudInitNoCloud.UserDataSecretRef != nil {
-			toCreateSecretName := getTemplateVersionUserDataSecretName(templateVersion.Name, volume.Name)
-			if err := h.copySecret(volume.CloudInitNoCloud.UserDataSecretRef.Name, toCreateSecretName, templateVersion); err != nil {
-				return err
-			}
-		}
-		if volume.CloudInitNoCloud.NetworkDataSecretRef != nil {
-			toCreateSecretName := getTemplateVersionNetworkDataSecretName(templateVersion.Name, volume.Name)
-			if err := h.copySecret(volume.CloudInitNoCloud.NetworkDataSecretRef.Name, toCreateSecretName, templateVersion); err != nil {
-				return err
-			}
-		}
-	}
-	return nil
-}
-
-func (h *vmActionHandler) copySecret(sourceName, targetName string, templateVersion *harvesterv1.VirtualMachineTemplateVersion) error {
-	secret, err := h.secretCache.Get(templateVersion.Namespace, sourceName)
-	if err != nil {
-		logrus.WithError(err).WithFields(logrus.Fields{
-			"namespace":       templateVersion.Namespace,
-			"templateVersion": templateVersion.Name,
-			"secret":          sourceName,
-		}).Error("Failed to get secret")
-		return err
-	}
-	toCreate := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      targetName,
-			Namespace: secret.Namespace,
-			OwnerReferences: []metav1.OwnerReference{
-				{
-					APIVersion: templateVersion.APIVersion,
-					Kind:       templateVersion.Kind,
-					Name:       templateVersion.Name,
-					UID:        templateVersion.UID,
-				},
-			},
-		},
-		Data: secret.Data,
-	}
-	_, err = h.secretClient.Create(toCreate)
-	return err
-
-}
-
 func (h *vmActionHandler) updateVMVolumeClaimTemplate(vm *kubevirtv1.VirtualMachine, updateVolumeClaimTemplate func([]corev1.PersistentVolumeClaim) ([]corev1.PersistentVolumeClaim, bool)) error {
 	var volumeClaimTemplates []corev1.PersistentVolumeClaim
 	vmCopy := vm.DeepCopy()
@@ -1102,137 +933,11 @@ func (h *vmActionHandler) updateVMVolumeClaimTemplate(vm *kubevirtv1.VirtualMach
 	anno[util.AnnotationVolumeClaimTemplates] = string(volumeClaimTemplatesJSON)
 	vmCopy.SetAnnotations(anno)
 	if !reflect.DeepEqual(vm, vmCopy) {
-		if _, err = h.vms.Update(vmCopy); err != nil {
+		if _, err = h.vmClient.Update(vmCopy); err != nil {
 			return fmt.Errorf("failed to update vm %s/%s, error: %v", vm.Namespace, vm.Name, err)
 		}
 	}
 	return nil
-}
-
-func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (map[string]corev1.PersistentVolumeClaim, map[string]string, error) {
-	pvcMap := map[string]corev1.PersistentVolumeClaim{}
-	pvcStorageClassMap := map[string]string{}
-	for _, volume := range vm.Spec.Template.Spec.Volumes {
-		if volume.PersistentVolumeClaim == nil {
-			continue
-		}
-
-		pvc, err := h.pvcCache.Get(vm.Namespace, volume.PersistentVolumeClaim.ClaimName)
-		if err != nil {
-			return pvcMap, pvcStorageClassMap, err
-		}
-		pvcMap[pvc.Name] = *pvc
-
-		pvcStorageClass, err := h.storageClassCache.Get(*pvc.Spec.StorageClassName)
-		if err != nil {
-			return pvcMap, pvcStorageClassMap, err
-		}
-		var sc *storagev1.StorageClass
-		if _, ok := pvcStorageClass.Parameters["backingImage"]; ok {
-			// If "backingImage" is set, the storage class is from a VM image.
-			// We can't use it directly. We need to find storageClassName annotation.
-			if imageID, ok := pvc.Annotations[util.AnnotationImageID]; ok {
-				imageIDSplit := strings.Split(imageID, "/")
-				if len(imageIDSplit) == 2 {
-					vmImage, err := h.vmImageCache.Get(imageIDSplit[0], imageIDSplit[1])
-					if err != nil {
-						return pvcMap, pvcStorageClassMap, err
-					}
-					if storageClassName, ok := vmImage.Annotations[util.AnnotationStorageClassName]; ok {
-						sc, err = h.storageClassCache.Get(storageClassName)
-						if err != nil {
-							return pvcMap, pvcStorageClassMap, err
-						}
-					}
-				}
-			}
-		} else {
-			sc = pvcStorageClass
-		}
-
-		if sc == nil {
-			pvcStorageClassMap[pvc.Name] = ""
-			continue
-		}
-		pvcStorageClassMap[pvc.Name] = sc.Name
-	}
-	return pvcMap, pvcStorageClassMap, nil
-}
-
-func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMachineTemplateVersion, vm *kubevirtv1.VirtualMachine, pvcStorageClassMap map[string]string) (map[string]harvesterv1.VirtualMachineImage, error) {
-	var err error
-	vmImageMap := map[string]harvesterv1.VirtualMachineImage{}
-	for index, volume := range vm.Spec.Template.Spec.Volumes {
-		if volume.PersistentVolumeClaim == nil {
-			continue
-		}
-		targetSCName := pvcStorageClassMap[volume.PersistentVolumeClaim.ClaimName]
-		// 3 cases here:
-		// Root volume with backingImage, check SC name starts with "longhorn-templateversion-", use backingImage backend
-		// volume with non longhorn provisioner, use CDI backend
-		// volume with longhorn provisioner, use backingImage backend
-		// we use CDI as default, so we need to check other two cases
-		vmImageBackend := harvesterv1.VMIBackendCDI
-		if strings.HasPrefix(targetSCName, "longhorn-templateversion-") {
-			vmImageBackend = harvesterv1.VMIBackendBackingImage
-		} else {
-			// checking SC
-			targetSC, err := h.storageClassCache.Get(targetSCName)
-			if err != nil {
-				return nil, fmt.Errorf("failed to get storage class %s, error: %v", targetSCName, err)
-			}
-			if targetSC.Provisioner == util.CSIProvisionerLonghorn {
-				if dataEngineVers, find := targetSC.Parameters["dataEngine"]; find {
-					if dataEngineVers == string(longhorn.DataEngineTypeV1) {
-						vmImageBackend = harvesterv1.VMIBackendBackingImage
-					}
-				} else {
-					vmImageBackend = harvesterv1.VMIBackendBackingImage
-				}
-			}
-		}
-		if vmImageBackend == "" {
-			return nil, fmt.Errorf("failed to configure vm image backend for volume %s", volume.Name)
-		}
-		vmImageName := getTemplateVersionVMImageName(templateVersion.Name, index)
-		vmImage := &harvesterv1.VirtualMachineImage{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      vmImageName,
-				Namespace: vm.Namespace,
-				Annotations: map[string]string{
-					util.AnnotationStorageClassName: targetSCName,
-				},
-				OwnerReferences: []metav1.OwnerReference{
-					{
-						APIVersion: templateVersion.APIVersion,
-						Kind:       templateVersion.Kind,
-						Name:       templateVersion.Name,
-						UID:        templateVersion.UID,
-					},
-				},
-			},
-			Spec: harvesterv1.VirtualMachineImageSpec{
-				Backend:                vmImageBackend,
-				DisplayName:            vmImageName,
-				SourceType:             harvesterv1.VirtualMachineImageSourceTypeExportVolume,
-				PVCName:                volume.PersistentVolumeClaim.ClaimName,
-				PVCNamespace:           vm.Namespace,
-				TargetStorageClassName: targetSCName,
-			},
-		}
-
-		vmImage, err = h.vmImageClient.Create(vmImage)
-		if err != nil {
-			logrus.WithError(err).WithFields(logrus.Fields{
-				"namespace":       templateVersion.Namespace,
-				"templateVersion": templateVersion.Name,
-				"pvc":             volume.PersistentVolumeClaim.ClaimName,
-			}).Error("Failed to create VM image")
-			return nil, err
-		}
-		vmImageMap[vmImage.Name] = *vmImage
-	}
-	return vmImageMap, nil
 }
 
 // Since the LH volume creation and replica scheduling are asynchronous
@@ -1695,64 +1400,6 @@ func replaceSecrets(templateVersionName string, vm *kubevirtv1.VirtualMachine) *
 		}
 	}
 	return sanitizedVM
-}
-
-func (h *vmActionHandler) sanitizeVolumes(templateVersion *harvesterv1.VirtualMachineTemplateVersion, pvcMap map[string]corev1.PersistentVolumeClaim, vmImageMap map[string]harvesterv1.VirtualMachineImage) error {
-	templateVersionCopy := templateVersion.DeepCopy()
-	vmSource := &templateVersionCopy.Spec.VM
-	volumeClaimTemplates := []corev1.PersistentVolumeClaim{}
-	for index, volume := range vmSource.Spec.Template.Spec.Volumes {
-		if volume.PersistentVolumeClaim == nil {
-			continue
-		}
-
-		// generate new volume template
-		// - longhorn v1, we need to use the new storageclass Name (for backingImage)
-		// - others, use the pvc StorageClassName
-		vmImageName := getTemplateVersionVMImageName(templateVersion.Name, index)
-		vmImage := vmImageMap[vmImageName]
-		pvc := pvcMap[volume.PersistentVolumeClaim.ClaimName]
-		targetStorageClassName := util.GenerateStorageClassName(string(vmImage.UID))
-		if _, find := pvc.Annotations[cdicommon.AnnCreatedForDataVolume]; find {
-			targetStorageClassName = *pvc.Spec.StorageClassName
-		}
-		annoImageID := fmt.Sprintf("%s/%s", templateVersion.Namespace, vmImageName)
-		pvcName := getTemplateVersionPvcName(templateVersion.Name, index)
-		volumeClaimTemplates = append(volumeClaimTemplates, corev1.PersistentVolumeClaim{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: pvcName,
-				Annotations: map[string]string{
-					util.AnnotationImageID: annoImageID,
-				},
-			},
-			Spec: corev1.PersistentVolumeClaimSpec{
-				AccessModes:      pvc.Spec.AccessModes,
-				Resources:        pvc.Spec.Resources,
-				VolumeMode:       pvc.Spec.VolumeMode,
-				StorageClassName: &targetStorageClassName,
-			},
-		})
-		vmSource.Spec.Template.Spec.Volumes[index].PersistentVolumeClaim.ClaimName = pvcName
-	}
-
-	volumeCliamTemplatesJSON, err := json.Marshal(volumeClaimTemplates)
-	if err != nil {
-		logrus.WithError(err).WithFields(logrus.Fields{
-			"namespace":       templateVersion.Namespace,
-			"templateVersion": templateVersion.Name,
-			"volumeClaim":     volumeClaimTemplates,
-		}).Error("Failed to json.Marshal volume claim templates")
-		return err
-	}
-
-	if vmSource.ObjectMeta.Annotations == nil {
-		vmSource.ObjectMeta.Annotations = map[string]string{}
-	}
-	vmSource.ObjectMeta.Annotations[util.AnnotationVolumeClaimTemplates] = string(volumeCliamTemplatesJSON)
-	if _, err := h.vmTemplateVersionClient.Update(templateVersionCopy); err != nil {
-		return err
-	}
-	return nil
 }
 
 func (h *vmActionHandler) dismissInsufficientResourceQuota(name, namespace string) error {
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -1337,43 +1337,6 @@ func cloneSecretVolume(volume *kubevirtv1.Volume, secretNameMap map[string]strin
 	volume.Secret.SecretName = secretNameMap[volume.Secret.SecretName]
 }
 
-// sanitizeVirtualMachineForTemplateVersion removes mac addresses and changes secrets names.
-// The PVCs are not sanitized in the function because users may want to create template with data.
-// This will change storage class name in PVCs, so we handle it with VMImages.
-func (h *vmActionHandler) sanitizeVirtualMachineForTemplateVersion(templateVersionName string, vm *kubevirtv1.VirtualMachine) harvesterv1.VirtualMachineSourceSpec {
-	sanitizedVM := removeMacAddresses(vm)
-	sanitizedVM = replaceSecrets(templateVersionName, sanitizedVM)
-
-	return harvesterv1.VirtualMachineSourceSpec{
-		ObjectMeta: sanitizedVM.ObjectMeta,
-		Spec:       sanitizedVM.Spec,
-	}
-}
-
-func replaceSecrets(templateVersionName string, vm *kubevirtv1.VirtualMachine) *kubevirtv1.VirtualMachine {
-	sanitizedVM := vm.DeepCopy()
-	for index, credential := range sanitizedVM.Spec.Template.Spec.AccessCredentials {
-		if sshPublicKey := credential.SSHPublicKey; sshPublicKey != nil && sshPublicKey.Source.Secret != nil {
-			sanitizedVM.Spec.Template.Spec.AccessCredentials[index].SSHPublicKey.Source.Secret.SecretName = getTemplateVersionSSHPublicKeySecretName(templateVersionName, index)
-		}
-		if userPassword := credential.UserPassword; userPassword != nil && userPassword.Source.Secret != nil {
-			sanitizedVM.Spec.Template.Spec.AccessCredentials[index].UserPassword.Source.Secret.SecretName = getTemplateVersionUserPasswordSecretName(templateVersionName, index)
-		}
-	}
-	for index, volume := range sanitizedVM.Spec.Template.Spec.Volumes {
-		if volume.CloudInitNoCloud == nil {
-			continue
-		}
-		if volume.CloudInitNoCloud.UserDataSecretRef != nil {
-			sanitizedVM.Spec.Template.Spec.Volumes[index].CloudInitNoCloud.UserDataSecretRef.Name = getTemplateVersionUserDataSecretName(templateVersionName, volume.Name)
-		}
-		if volume.CloudInitNoCloud.NetworkDataSecretRef != nil {
-			sanitizedVM.Spec.Template.Spec.Volumes[index].CloudInitNoCloud.NetworkDataSecretRef.Name = getTemplateVersionNetworkDataSecretName(templateVersionName, volume.Name)
-		}
-	}
-	return sanitizedVM
-}
-
 func (h *vmActionHandler) dismissInsufficientResourceQuota(name, namespace string) error {
 	vm, err := h.vmCache.Get(namespace, name)
 	if err != nil {
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -913,32 +913,46 @@ func getMigrationUID(vmi *kubevirtv1.VirtualMachineInstance) string {
 }
 
 // createTemplate creates a template and version that are derived from the given virtual machine.
-func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTemplateInput) error {
-	vm, err := h.vmCache.Get(namespace, name)
+func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTemplateInput) (err error) {
+	var (
+		vm         *kubevirtv1.VirtualMachine
+		keyPairIDs []string
+		vmt        *harvesterv1.VirtualMachineTemplate
+		vmtv       *harvesterv1.VirtualMachineTemplateVersion
+	)
+	defer func() {
+		if err != nil {
+			logrus.WithError(err).WithFields(logrus.Fields{
+				"namespace": namespace,
+				"name":      name,
+			}).Error("Failed to create template")
+
+			if vmtv != nil {
+				// other resources have owner reference to vmtv, so we only need to delete vmtv
+				if cleanupErr := h.vmTemplateVersionClient.Delete(vmtv.Namespace, vmtv.Name, &metav1.DeleteOptions{}); cleanupErr != nil {
+					logrus.WithError(cleanupErr).WithFields(logrus.Fields{
+						"namespace": vmtv.Namespace,
+						"name":      vmtv.Name,
+					}).Error("Failed to delete vm template version")
+					err = fmt.Errorf("failed to delete vm template version, cleanup error: %s, original error: %w", cleanupErr.Error(), err)
+				}
+			}
+		}
+	}()
+	vm, err = h.vmCache.Get(namespace, name)
 	if err != nil {
 		return err
 	}
 
-	keyPairIDs, err := getSSHKeysFromVMITemplateSpec(vm.Spec.Template)
+	keyPairIDs, err = getSSHKeysFromVMITemplateSpec(vm.Spec.Template)
 	if err != nil {
 		return err
 	}
 
 	vmtvName := fmt.Sprintf("%s-%s", input.Name, rand.String(5))
-	vmSourceSpec, err := h.sanitizeVirtualMachineForTemplateVersion(vmtvName, vm, input.WithData)
-	if err != nil {
-		return err
-	}
+	vmSourceSpec := h.sanitizeVirtualMachineForTemplateVersion(vmtvName, vm)
 
-	var pvcStorageClassMap map[string]string
-	if input.WithData {
-		pvcStorageClassMap, err = h.getPVCStorageClassMap(vm)
-		if err != nil {
-			return err
-		}
-	}
-
-	vmt, err := h.vmTemplateClient.Create(
+	vmt, err = h.vmTemplateClient.Create(
 		&harvesterv1.VirtualMachineTemplate{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      input.Name,
@@ -954,7 +968,7 @@ func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTem
 
 	vmID := fmt.Sprintf("%s/%s", vmt.Namespace, vmt.Name)
 
-	vmtv, err := h.vmTemplateVersionClient.Create(
+	vmtv, err = h.vmTemplateVersionClient.Create(
 		&harvesterv1.VirtualMachineTemplateVersion{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      vmtvName,
@@ -972,7 +986,27 @@ func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTem
 	}
 
 	if input.WithData {
-		if err := h.createVMImages(vmtv, vm, pvcStorageClassMap); err != nil {
+		var (
+			pvcMap             map[string]corev1.PersistentVolumeClaim
+			pvcStorageClassMap map[string]string
+			vmImageMap         map[string]harvesterv1.VirtualMachineImage
+		)
+		pvcMap, pvcStorageClassMap, err = h.getPVCStorageClassMap(vm)
+		if err != nil {
+			return err
+		}
+		vmImageMap, err = h.createVMImages(vmtv, vm, pvcStorageClassMap)
+		if err != nil {
+			return err
+		}
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			vmtv, err = h.vmTemplateVersionClient.Get(vmtv.Namespace, vmtv.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			return h.sanitizeVolumes(vmtv, pvcMap, vmImageMap)
+		})
+		if err != nil {
 			return err
 		}
 	}
@@ -1018,6 +1052,11 @@ func (h *vmActionHandler) createSecrets(templateVersion *harvesterv1.VirtualMach
 func (h *vmActionHandler) copySecret(sourceName, targetName string, templateVersion *harvesterv1.VirtualMachineTemplateVersion) error {
 	secret, err := h.secretCache.Get(templateVersion.Namespace, sourceName)
 	if err != nil {
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"namespace":       templateVersion.Namespace,
+			"templateVersion": templateVersion.Name,
+			"secret":          sourceName,
+		}).Error("Failed to get secret")
 		return err
 	}
 	toCreate := &corev1.Secret{
@@ -1040,7 +1079,38 @@ func (h *vmActionHandler) copySecret(sourceName, targetName string, templateVers
 
 }
 
-func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (map[string]string, error) {
+func (h *vmActionHandler) updateVMVolumeClaimTemplate(vm *kubevirtv1.VirtualMachine, updateVolumeClaimTemplate func([]corev1.PersistentVolumeClaim) ([]corev1.PersistentVolumeClaim, bool)) error {
+	var volumeClaimTemplates []corev1.PersistentVolumeClaim
+	vmCopy := vm.DeepCopy()
+	anno := vmCopy.GetAnnotations()
+	if volumeClaimTemplatesJSON, ok := anno[util.AnnotationVolumeClaimTemplates]; ok {
+		if err := json.Unmarshal([]byte(volumeClaimTemplatesJSON), &volumeClaimTemplates); err != nil {
+			return fmt.Errorf("failed to unserialize %s, error: %v", util.AnnotationVolumeClaimTemplates, err)
+		}
+	}
+
+	var changed bool
+	volumeClaimTemplates, changed = updateVolumeClaimTemplate(volumeClaimTemplates)
+	if !changed {
+		return nil
+	}
+
+	volumeClaimTemplatesJSON, err := json.Marshal(volumeClaimTemplates)
+	if err != nil {
+		return fmt.Errorf("failed to serialize payload %v, error: %v", volumeClaimTemplates, err)
+	}
+	anno[util.AnnotationVolumeClaimTemplates] = string(volumeClaimTemplatesJSON)
+	vmCopy.SetAnnotations(anno)
+	if !reflect.DeepEqual(vm, vmCopy) {
+		if _, err = h.vms.Update(vmCopy); err != nil {
+			return fmt.Errorf("failed to update vm %s/%s, error: %v", vm.Namespace, vm.Name, err)
+		}
+	}
+	return nil
+}
+
+func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (map[string]corev1.PersistentVolumeClaim, map[string]string, error) {
+	pvcMap := map[string]corev1.PersistentVolumeClaim{}
 	pvcStorageClassMap := map[string]string{}
 	for _, volume := range vm.Spec.Template.Spec.Volumes {
 		if volume.PersistentVolumeClaim == nil {
@@ -1049,11 +1119,13 @@ func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (
 
 		pvc, err := h.pvcCache.Get(vm.Namespace, volume.PersistentVolumeClaim.ClaimName)
 		if err != nil {
-			return pvcStorageClassMap, err
+			return pvcMap, pvcStorageClassMap, err
 		}
+		pvcMap[pvc.Name] = *pvc
+
 		pvcStorageClass, err := h.storageClassCache.Get(*pvc.Spec.StorageClassName)
 		if err != nil {
-			return pvcStorageClassMap, err
+			return pvcMap, pvcStorageClassMap, err
 		}
 		var sc *storagev1.StorageClass
 		if _, ok := pvcStorageClass.Parameters["backingImage"]; ok {
@@ -1064,12 +1136,12 @@ func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (
 				if len(imageIDSplit) == 2 {
 					vmImage, err := h.vmImageCache.Get(imageIDSplit[0], imageIDSplit[1])
 					if err != nil {
-						return pvcStorageClassMap, err
+						return pvcMap, pvcStorageClassMap, err
 					}
 					if storageClassName, ok := vmImage.Annotations[util.AnnotationStorageClassName]; ok {
 						sc, err = h.storageClassCache.Get(storageClassName)
 						if err != nil {
-							return pvcStorageClassMap, err
+							return pvcMap, pvcStorageClassMap, err
 						}
 					}
 				}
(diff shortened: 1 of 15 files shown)


pkg/util/image.go
```diff
@@ -72,11 +72,47 @@ func GetBackingImageDataSourceName(backingImageCache ctllhv1.BackingImageCache,
...
+	return lhutil.AutoCorrectName(fmt.Sprintf("longhorn-%s", imageUID), lhdatastore.NameMaximumLength)
+}
+
+func GetStorageClass(storageClassCache ctlstoragev1.StorageClassCache, image *harvesterv1.VirtualMachineImage) (*storagev1.StorageClass, error) {
+	// For backward compatibility, try to get the storage class with legacy name first.
+	// If it exists, return it directly.
+	// If not, try with a new name format based on the image UID, which can avoid name conflict. Because
+	// the legacy name is based on image name, so two images with the same name but in different namespaces
+	// would generate the same storage class name (StorageClass doesn't have namespace).
+	sc, err := storageClassCache.Get(legacyStorageClassName(image.Name))
```
w13915984028: should we try to get new SC name first, and finally fallback to legacy for last try?

a new image might also hit legacy SC as the old naming has more chances to conflict

the consideration is put to https://github.com/harvester/harvester/pull/9392#issuecomment-3472045861

pkg/image/backingimage/backingimage.go
```diff
@@ -83,7 +83,12 @@ func (bib *Backend) deleteBackingImage(vmi *harvesterv1.VirtualMachineImage) err
 }
 
 func (bib *Backend) deleteStorageClass(vmi *harvesterv1.VirtualMachineImage) error {
-	return bib.scClient.Delete(util.GetImageStorageClassName(vmi), &metav1.DeleteOptions{})
+	storageClassName, err := util.GetImageStorageClassName(bib.scCache, vmi)
+	if err != nil {
+		return err
+	}
```
w13915984028: skip isNotFound ?

pkg/image/backingimage/backingimage.go
```diff
@@ -174,7 +179,12 @@ func (bib *Backend) createBackingImage(vmi *harvesterv1.VirtualMachineImage) err
 }
 
 func (bib *Backend) createStorageClass(vmi *harvesterv1.VirtualMachineImage) error {
-	if cachedSC, _ := bib.scCache.Get(util.GetImageStorageClassName(vmi)); cachedSC != nil && cachedSC.DeletionTimestamp != nil {
+	storageClassName, err := util.GetImageStorageClassName(bib.scCache, vmi)
+	if err != nil {
+		return err
+	}
+
+	if cachedSC, _ := bib.scCache.Get(storageClassName); cachedSC != nil && cachedSC.DeletionTimestamp != nil {
```
w13915984028, line 179: should we return earlier if `cachedSc != nil && cachedSC.DeletionTimestamp == nil` ?

w13915984028 (commented):
Some questions, thanks.

w13915984028:
@DamiaSan Did you consider to `mutator` the newly created vmi, to add an annotation/lablel indicating the new SC naming, and later when controller creating/fetching/deleting the SC, prefer the mark first.  
Or update the `VMI CRD` to add a new filed like `scNamingVersion`. new version to use new naming.

The retry between new naming and old naming, still has chance to be messed. thanks.

https://github.com/harvester/harvester/blob/3e533b9c62af83c1275079805ac5db96ccb57ace/pkg/webhook/resources/virtualmachineimage/mutator.go#L61

pkg/image/backingimage/backingimage.go
DamiaSan: The `IsNotFound` error is handled inside the function `GetImageStorageClassName`, do you suggest something different?

w13915984028: skip the `isNotFound` returned from `bib.scClient.Delete(storageClassName, &metav1.DeleteOptions{}` ?

pkg/image/backingimage/backingimage.go
DamiaSan: In this case, how should we return, error or nil?

w13915984028: The reconciller might run to this function multi times.

Either use the existing one, or compare parameters and if not same, update it or delete/create new one and log related message/event. WDYT?

pkg/image/backingimage/backingimage.go
DamiaSan: This error wasn't skipped before, why should we skip it now? Making so we are going to change the behavior of the function ...

pkg/image/backingimage/backingimage.go
DamiaSan: I think that a `create` function should not update an existing storage class, so we should use the existing one.
But are we sure that it isn't better to leave to the specific implementation of the `backingimage` to handle this case (in the `bib.scClient.Create` function)? Maybe they could handle this case in different ways ...

Vicente-Cheng:
Hi @FrankYang0529, 
Please also check this PR. I will take a look at it, too.

pkg/util/image.go
DamiaSan: If we check the new SC before the legacy one, we should create a new `scNamingVersion` fiels as you suggest, but what benefit will we have? Briefly, which is the case where checking the legacy one before the new one could fail? Can you make an example?

w13915984028: a legacy VMI named `abc` in namespace `ns1` has SC name  `longhorn-abc`, 
a new VMI `abc` in namespace `ns2` will have SC name `longhorn-af3069-....` in new mode

the `GetStorageClass` tries legacy first, `ns2/abc` will get `longhorn-abc`

by contrast, try new mode first,  `ns1/abc` won't get `longhorn-af3069-...`

DamiaSan: > ns2/abc will get longhorn-abc

Is this a problem? Because we should simply avoid conflicts ...

w13915984028: From the PR desc, I assumed we allow user to configure as such (user on different namespaces might not have context of each other) and this PR aims to solve it.

The new naming is conflict-free and legacy naming is conflict-vulnerable,  and it is possible that newly created VMI has same name with existing VMI.

```
Problem:
We use longhorn-<vmimage name> as storage class name. The StorageClass doesn't have namesapce, but vmimage has. If there are same vmimage name in different namespace, storage class name will be conflicts.

Solution:
Use namespace and image UID as name format.
```

DamiaSan: What do you think @FrankYang0529 @Vicente-Cheng ?

m-ildefons:
Can anyone enlighten me why we're creating a separate storage class for each VMImage?
Why can't we just re-use the default storage class unless the user specifies otherwise?

Vicente-Cheng:
> Can anyone enlighten me why we're creating a separate storage class for each VMImage? Why can't we just re-use the default storage class unless the user specifies otherwise?

This is the special behavior of Longhorn backingimage. We need a specific field, like `parameters/backingImage`, to point out how we create the PVC from the backing image. 

That's why we need to do that.

pkg/api/vm/handler.go
```diff
@@ -717,32 +717,46 @@ func getMigrationUID(vmi *kubevirtv1.VirtualMachineInstance) string {
...
+				if cleanupErr := h.vmTemplateVersionClient.Delete(vmtv.Namespace, vmtv.Name, &metav1.DeleteOptions{}); cleanupErr != nil {
+					logrus.WithError(cleanupErr).WithFields(logrus.Fields{
+						"namespace": vmtv.Namespace,
+						"name":      vmtv.Name,
+					}).Error("Failed to delete vm template version")
+					err = fmt.Errorf("failed to delete vm template version, cleanup error: %s, original error: %w", cleanupErr.Error(), err)
+				}
+			}
+		}
+	}()
```
FrankYang0529: We need to remove `vmt` as well if it is not nil.

pkg/util/image.go
FrankYang0529: I agree that we should get new mode first. A potential issue is like following:

* User creates a new VMImage `default/test`.
* User creates a new VMIMage `new-ns/test`. In current logic, it will remove `longhorn-test` StorageClass and cause issue.

m-ildefons:
> > Can anyone enlighten me why we're creating a separate storage class for each VMImage? Why can't we just re-use the default storage class unless the user specifies otherwise?
> 
> This is the special behavior of Longhorn backingimage. We need a specific field, like `parameters/backingImage`, to point out how we create the PVC from the backing image.
> 
> That's why we need to do that.

How does having a separate storage class for each VMImage affect that? I don't understand.

w13915984028:
@m-ildefons 

The `de facto` is that the `StorageClass.parameters.backingImage` points to the image UUID. It caused the SC & image became a <1:1> pair.

I did not check the history why it was imple..d as such.

Backingimage CRD seems no field to book the SC name.

```
$ kubectl get storageclass longhorn-image-f2vpr -oyaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2025-11-18T18:34:04Z"
  name: longhorn-image-f2vpr
  resourceVersion: "718159"
  uid: 555a7a11-6200-4c11-bec1-0f2a4f903f39
parameters:
  backingImage: vmi-98ceb77e-da79-4cd6-8c4c-25e631114aa9  // note
  migratable: "true"
  numberOfReplicas: "3"
  staleReplicaTimeout: "30"
reclaimPolicy: Delete
volumeBindingMode: Immediate

$ kubectl get storageclass longhorn-image-666qx -oyaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2025-11-17T14:16:27Z"
  name: longhorn-image-666qx
  resourceVersion: "279553"
  uid: f4dcf318-f15a-4ce6-8dab-a4dde3204f09
parameters:
  backingImage: vmi-6892e1d3-4eb3-4344-bfe0-2f6cfb2fceb2  // note
  migratable: "true"
  numberOfReplicas: "3"
  staleReplicaTimeout: "30"
reclaimPolicy: Delete
volumeBindingMode: Immediate


$ kubectl get backingimage -n longhorn-system vmi-98ceb77e-da79-4cd6-8c4c-25e631114aa9 -oyaml
apiVersion: longhorn.io/v1beta2
kind: BackingImage
metadata:
  annotations:
    harvesterhci.io/imageId: default/image-f2vpr
  creationTimestamp: "2025-11-18T18:34:04Z"
  finalizers:
  - longhorn.io
  generation: 2
  labels:
    longhorn.io/component: backing-image
    longhorn.io/managed-by: longhorn-manager
  name: vmi-98ceb77e-da79-4cd6-8c4c-25e631114aa9
  namespace: longhorn-system
  resourceVersion: "841602"
  uid: 8de70ef6-89af-4a1f-a4e2-070c8a19ee00
spec:
  checksum: ""
  dataEngine: v1
  diskFileSpecMap:
    1855ea8b-e0cc-4f01-9acd-5bf7f382f8b8:
      dataEngine: v1
      evictionRequested: false
  disks: {}
  minNumberOfCopies: 3
  secret: ""
  secretNamespace: ""
  sourceParameters:
    export-type: raw
    volume-name: pvc-97293b02-ac49-469d-9ae4-a4fcd939da0a
  sourceType: export-from-volume
status:
  checksum: 0c40c18b92b740aabaf0c2c572834972155ab85c2472f04ffc00513e368ffe8c6dc254b2272e48364eb0587ecba3dd345e59f822860dc9f76b2f652de0c42cbe
  diskFileStatusMap:
    1855ea8b-e0cc-4f01-9acd-5bf7f382f8b8:
      dataEngine: v1
      lastStateTransitionTime: "2025-11-19T12:48:08Z"
      message: ""
      progress: 100
      state: ready
  diskLastRefAtMap:
    1855ea8b-e0cc-4f01-9acd-5bf7f382f8b8: "2025-11-18T21:42:14Z"
  ownerID: harv41
  realSize: 4294971392
  size: 4294967296
  uuid: 13405d4e
  v2FirstCopyDisk: ""
  v2FirstCopyStatus: ""
  virtualSize: 4294967296

```

m-ildefons:
Ok, so the StorageClass contains the information which BackingImage is to be used when initializing a PVC.
Therefore if I want a PVC initialized from a specific VMImage, I need a StorageClass that is configured to use the corresponding BackingImage.

Thanks @w13915984028 now I think I understand why we need a separate storage class for each VMImage.

---

It's a weird construction, because it means that the parameters used for a volume (e.g. replica count, stale replica timeout, etc) have to be known when the VMImage is created and if I want multiple volumes with different parameters, but with the same content, I need to create multiple different VMImages with the same content but different parameters.
It seems like it would be better to have an annotation on the PVC which defines which BackingImage (if any) is to be used to initialize the PVC and otherwise let the StorageClass only handle the parameters for the storage engine. That way it would be possible to instantiate multiple VMs from the same VMImage but with different parameters for their root disk PVCs, where as now we would need several copies of the VMImage, each with different parameters for the storage engine...
But this would require changes in Longhorn.

pkg/util/image.go
WebberHuang1118: Agreed — we should get the new mode first, because the new mode has a one-to-one mapping between the VMI and the StorageClass.

pkg/api/vm/handler.go
WebberHuang1118: There are nested `if` conditions, which hurts readability. Suggest refactoring this into smaller, more focused functions.

pkg/api/vm/handler.go
```diff
@@ -776,7 +790,27 @@ func (h *vmActionHandler) createTemplate(namespace, name string, input CreateTem
...
+			return err
+		}
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			vmtv, err = h.vmTemplateVersionClient.Get(vmtv.Namespace, vmtv.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			return h.sanitizeVolumes(vmtv, pvcMap, vmImageMap)
+		})
+		if err != nil {
```
WebberHuang1118: This function has become too long; Suggest factoring this section out into a separate function.

pkg/api/vm/handler.go
```diff
@@ -874,7 +913,8 @@ func (h *vmActionHandler) updateVMVolumeClaimTemplate(vm *kubevirtv1.VirtualMach
 	return nil
 }
 
-func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (map[string]string, error) {
+func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (map[string]corev1.PersistentVolumeClaim, map[string]string, error) {
```
WebberHuang1118: This function’s readability isn’t great — there are nested `for` and `if` blocks. Please extract the logic inside the main loop into a separate function.

pkg/api/vm/handler.go
```diff
@@ -918,10 +960,12 @@ func (h *vmActionHandler) getPVCStorageClassMap(vm *kubevirtv1.VirtualMachine) (
 		}
 		pvcStorageClassMap[pvc.Name] = sc.Name
 	}
-	return pvcStorageClassMap, nil
+	return pvcMap, pvcStorageClassMap, nil
 }
 
-func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMachineTemplateVersion, vm *kubevirtv1.VirtualMachine, pvcStorageClassMap map[string]string) error {
+func (h *vmActionHandler) createVMImages(templateVersion *harvesterv1.VirtualMachineTemplateVersion, vm *kubevirtv1.VirtualMachine, pvcStorageClassMap map[string]string) (map[string]harvesterv1.VirtualMachineImage, error) {
```
WebberHuang1118: ditto

pkg/api/vm/handler.go
```diff
@@ -1492,29 +1539,27 @@ func replaceSecrets(templateVersionName string, vm *kubevirtv1.VirtualMachine) *
 	return sanitizedVM
 }
 
-func (h *vmActionHandler) replaceVolumes(templateVersionName string, vm *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
-	sanitizedVM := vm.DeepCopy()
+func (h *vmActionHandler) sanitizeVolumes(templateVersion *harvesterv1.VirtualMachineTemplateVersion, pvcMap map[string]corev1.PersistentVolumeClaim, vmImageMap map[string]harvesterv1.VirtualMachineImage) error {
```
WebberHuang1118: ditto

WebberHuang1118 (changes requested):
In `pkg/api/vm/handler.go`, the change introduces nested `if` and `for` statements, which reduces readability and may make future maintenance or extensions more difficult. 

Please factor out the logic and wrap it into smaller, more focused functions.

pkg/api/vm/handler.go
m-ildefons: Done.

pkg/api/vm/handler.go
m-ildefons: I've refactored and simplified this functions quite a bit. I've also moved it and all related functions into their own file, which should make them much easier to review.
I haven't yet gone through all of the related functions to refactor them as well, but as far as I can tell most of them aren't quite as convoluted.

pkg/api/vm/handler.go
m-ildefons: Ok, done.

WebberHuang1118:
@m-ildefons
Thanks for the improvements [here](https://github.com/harvester/harvester/pull/9392#discussion_r2571998805) and [here](https://github.com/harvester/harvester/pull/9392#discussion_r2571997465).

Regarding [`getPVCStorageClassMap`](https://github.com/DamiaSan/harvester/blob/e5f232775c4ef9d1a529e1c31ad8667b4d2f1050/pkg/api/vm/template.go#L106-L154), [`sanitizeVolumes`](https://github.com/DamiaSan/harvester/blob/e5f232775c4ef9d1a529e1c31ad8667b4d2f1050/pkg/api/vm/template.go#L156-L212), and [`createVMImages`](https://github.com/DamiaSan/harvester/blob/e5f232775c4ef9d1a529e1c31ad8667b4d2f1050/pkg/api/vm/template.go#L279-L353), the implementations are still quite dense. The complex nested conditions and loops hinder readability and maintainability. Please refactor these functions. Thanks.

BTW, please expand the test plan to include guest cluster creation, as [Rancher's node driver is also related](https://github.com/harvester/harvester/issues/5165#issuecomment-2594411682). cc @FrankYang0529

pkg/api/vm/handler.go
m-ildefons: Done

pkg/api/vm/handler.go
m-ildefons: Done

pkg/api/vm/handler.go
m-ildefons: Done

pkg/util/image.go
m-ildefons: Ok, done.

w13915984028 (commented):
Could we add some unit test code to cover the target cases as more as possible? thanks.

m-ildefons:
While going through a few test scenarios, I noticed that the original issue does not reproduce under Harvester v1.6.1 when using the web UI.
The root cause might be in the Terraform provider, making this PR superfluous.
This makes sense, because the existing code base differentiates between name and display name for VM Images. The display name can be the same and the name (i.e. `.metadata.name`) is generated at creation time without user input. There is a webhook which prevents VMImages with the same `.metadata.name` from being created as well.
The Terraform provider subverts the name generation and lets users directly specify the `.metadata.name`.

m-ildefons:
@WebberHuang1118 I believe all code issues have been addressed. Could you please take another look?

WebberHuang1118:
> @WebberHuang1118 I believe all code issues have been addressed. Could you please take another look?

Oh, I thought you were addressing the comment from @w13915984028. I’ll check your update — thanks.

WebberHuang1118:
> While going through a few test scenarios, I noticed that the original issue does not reproduce under Harvester v1.6.1 when using the web UI. The root cause might be in the Terraform provider, making this PR superfluous. This makes sense, because the existing code base differentiates between name and display name for VM Images. The display name can be the same and the name (i.e. `.metadata.name`) is generated at creation time without user input. There is a webhook which prevents VMImages with the same `.metadata.name` from being created as well. The Terraform provider subverts the name generation and lets users directly specify the `.metadata.name`.

This can also happen when directly applying manifests, e.g., using kubectl.

m-ildefons:
> Oh, I thought you were addressing the comment from @w13915984028. I’ll check your update — thanks.

Not yet, I will do that. But I have addressed your findings.

> This can also happen when directly applying manifests, e.g., using kubectl.

Yes, I noticed. That's why I'm pushing this PR forward and not trying to fix the Terraform provider only.

pkg/util/image.go
```diff
@@ -72,11 +72,47 @@ func GetBackingImageDataSourceName(backingImageCache ctllhv1.BackingImageCache,
...
+	if err == nil {
+		return sc.Name, nil
 	}
-	return fmt.Sprintf("longhorn-%s", image.Name)
+
+	if !errors.IsNotFound(err) {
+		return "", err
+	}
+
+	return GenerateStorageClassName(string(image.UID)), nil
```
WebberHuang1118: We should encapsulate `GetImageStorageClassName()` within the `VMIOperator` interface. 

In this PR's design, it directly accesses the VMImage member fields.

Instead, the StorageClass name should be retrievable through the `VMIOperator` interface, and the backing image backend, `bib` should obtain it via `VMIOperator` as well.

WebberHuang1118 (dismissed):
The current implementation violates the `VMIOperator` interface design. We should follow the clean, encapsulated interface it provides. 

Please check the comments. Thanks.

WebberHuang1118:
Re: [discussion](https://github.com/harvester/harvester/pull/9392#discussion_r2692911126)

To support VM root disks with third-party storage, we established `VMImage` as an [abstraction layer](https://github.com/harvester/harvester/pull/7170) in `v1.5.0` to accommodate different backends.

As a result, instead of directly accessing `VMImage` member fields, e.g.,

```
return GenerateStorageClassName(string(image.UID)), nil
```

```
sc, err := storageClassCache.Get(legacyStorageClassName(image.Name))
```

We should access it via the `VMIOperator`, which is intended to provide a clean and encapsulated interface. I realize some legacy code does not follow this principle, but we should gradually remove such implementations. Since this is a new enhancement, IMO, we must adhere to the interface design.

I envision extending `VMIOperator` with a new method, `GetSCName()`. This would allow any part of the source code holding a `VMIOperator` stub to retrieve the StorageClass name.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+
+			if cleanupErr = h.cleanupVMTemplateVersion(vmtv); cleanupErr != nil {
+				err = fmt.Errorf("failed to clean up VM template version while dealing with error %w: %s", err, cleanupErr.Error())
+			}
+
+			if cleanupErr = h.cleanupVMTemplate(vmt); cleanupErr != nil {
+				err = fmt.Errorf("failed to clean up VM template while dealing with error %w: %s", err, cleanupErr.Error())
+			}
+		}
+	}()
```
WebberHuang1118: Suggest using an early return to reduce nesting.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+		if err != nil {
+			return "", err
+		}
+
+		if scName, ok := vmImage.Annotations[util.AnnotationStorageClassName]; ok {
+			return scName, nil
+		}
+		return "", fmt.Errorf("VMImage %s/%s does not have an annotation %s", vmImage.Namespace, vmImage.Name, util.AnnotationStorageClassName)
+	}
+	return "", fmt.Errorf("malformed image ID: %s", imageId)
```
WebberHuang1118: Suggest using an early return to reduce nesting.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+			VolumeMode:       claim.Spec.VolumeMode,
+			StorageClassName: &targetStorageClassName,
+		},
+	}
+	return pvc
+}
+
+func (h *vmActionHandler) createSecrets(templateVersion *harvesterv1.VirtualMachineTemplateVersion, vm *kubevirtv1.VirtualMachine) error {
+	for index, credential := range vm.Spec.Template.Spec.AccessCredentials {
+		if sshPublicKey := credential.SSHPublicKey; sshPublicKey != nil && sshPublicKey.Source.Secret != nil {
```
WebberHuang1118: This `if` expression is too long; please wrap it in a separate function.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+			toCreateSecretName := getTemplateVersionSSHPublicKeySecretName(templateVersion.Name, index)
+			if err := h.copySecret(sshPublicKey.Source.Secret.SecretName, toCreateSecretName, templateVersion); err != nil {
+				return err
+			}
+		}
+		if userPassword := credential.UserPassword; userPassword != nil && userPassword.Source.Secret != nil {
```
WebberHuang1118: ditto

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+	if targetSC.Provisioner == util.CSIProvisionerLonghorn {
+		if dataEngineVers, ok := targetSC.Parameters["dataEngine"]; ok {
+			if dataEngineVers == string(longhorn.DataEngineTypeV1) {
+				return harvesterv1.VMIBackendBackingImage, nil
+			}
+		}
+	}
+
+	return harvesterv1.VMIBackendCDI, nil
+}
```
WebberHuang1118: Use `util.GetProvisionedPVCProvisioner()` to retrieve the CSI provisioner. The caller should pass the PVC to utilize this existing utility function.

Suggest using an early return to reduce nesting. Additionally, please wrap the logic to retrieve the data engine version instead of using plain text, consistent with the rest of the codebase.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+		if err = h.vmTemplateClient.Delete(vmt.Namespace, vmt.Name, &metav1.DeleteOptions{}); err != nil {
+			logrus.WithFields(logrus.Fields{
+				"namespace": vmt.Namespace,
+				"name":      vmt.Name,
+			}).Errorf("failed to clean up template version: %v", err)
+			return err
+		}
+	}
+	return nil
+}
```
WebberHuang1118: Suggest using an early return to reduce nesting.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+		if err = h.vmTemplateVersionClient.Delete(vmtv.Namespace, vmtv.Name, &metav1.DeleteOptions{}); err != nil {
+			logrus.WithFields(logrus.Fields{
+				"namespace": vmtv.Namespace,
+				"name":      vmtv.Name,
+			}).Errorf("failed to clean up template version: %v", err)
+			return err
+		}
+	}
+	return nil
+}
```
WebberHuang1118: Suggest using an early return to reduce nesting.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+		}
+		pvcStorageClassMap[pvc.Name] = scName
+	}
+	return pvcMap, pvcStorageClassMap, nil
+}
+
+// If the PVC has an annoation, which associates it with a BackingImage, then
+// this function will return the StorageClass name of the BackingImage. Otherwise it
+// will return the StorageClass name of the PVC itself.
+func (h *vmActionHandler) getStorageClassNameFromPVCOrBackingImage(pvc *corev1.PersistentVolumeClaim) (scName string, err error) {
```
WebberHuang1118: The function name is too long. Suggest shortening it.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+	if imageId, ok := pvc.Annotations[util.AnnotationImageID]; ok {
+		return h.getStorageClassNameFromImageID(imageId)
+	}
+	return *pvc.Spec.StorageClassName, nil
+}
+
+// Given an ImageID, find the corresponding VMImage and figure out it's
+// StorageClass name
+func (h *vmActionHandler) getStorageClassNameFromImageID(imageId string) (scName string, err error) {
```
WebberHuang1118: The function name is too long. Suggest shortening it.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,449 @@
...
+
+func generateVolumeClaimTemplate(index int, tvName, tvNamespace, claimName string, pvcMap map[string]corev1.PersistentVolumeClaim, vmImageMap map[string]harvesterv1.VirtualMachineImage) (pvc corev1.PersistentVolumeClaim) {
+	// generate new volume template
+	// - longhorn v1, we need to use the new storageclass Name (for backingImage)
+	// - others, use the pvc StorageClassName
+	vmImageName := getTemplateVersionVMImageName(tvName, index)
+	vmImage := vmImageMap[vmImageName]
+
+	claim := pvcMap[claimName]
+	targetStorageClassName := util.GenerateStorageClassName(string(vmImage.UID))
```
WebberHuang1118: The variable name is too long. Suggest shortening it.

WebberHuang1118:
@m-ildefons, is the PR ready for review again? It looks like some of the comments still haven’t been addressed.

m-ildefons:
> is the PR ready for review again? It looks like some of the comments still haven’t been addressed.

Sorry for the delay, I was out sick last week.
I've made changes addressing your comments specifically.

The `GetImageStorageClassName` functionality is now implemented as part of the `VMIOperator` interface and all callers have been updated to utilize such an operator instance.

I've also addressed a log-spamming issue that I noticed while testing the changes myself:
The naming-schema `longhorn-{{ IMAGE_UID }}` will reliably produce storage class names longer than 40 characters, which need to be truncated to fit Longhorn's 40 character limit for names, since the IMAGE_UID itself is already 37 characters.
This was remedied by using a utility function from the Longhorn library to truncate the name, but this generates two log messages on each individual access to a storage class. The log messages are emitted from Longhorn library code, so I don't want to remove them.
To address this, I've changed the naming-schema to be `lh-{{ IMAGE_UID }}`, so it stays under the 40 character limit by default.

pkg/image/common/operator.go
```diff
@@ -172,6 +177,23 @@ func (vmio *vmiOperator) GetDisplayName(vmi *harvesterv1.VirtualMachineImage) st
...
+	if err == nil {
+		return legacySCName
+	}
+
+	return lhutil.AutoCorrectName(
+		fmt.Sprintf("lh-%s", vmio.GetUID(vmi)),
+		lhdatastore.NameMaximumLength,
+	)
+}
+
```
WebberHuang1118, line 206: As [discussed](https://github.com/harvester/harvester/pull/9392#discussion_r2480509607), we should get the new mode first.

pkg/util/storageclass.go
```diff
@@ -0,0 +1,18 @@
...
+	ParamDataEngine = "dataEngine"
+)
+
+func GetLonghornDataEngineType(sc *storagev1.StorageClass) longhorn.DataEngineType {
+	v, ok := sc.Parameters[ParamDataEngine]
+	if ok {
+		return longhorn.DataEngineType(v)
+	}
+	return longhorn.DataEngineTypeAll
+}
```
WebberHuang1118, line 18: This doesn’t seem to be used—am I missing something?

m-ildefons: This is used in `pkg/api/vm/template.go` line 387.

WebberHuang1118: thanks

pkg/api/vm/template.go
```diff
@@ -0,0 +1,472 @@
...
+	if imageId, ok := pvc.Annotations[util.AnnotationImageID]; ok {
+		return h.x_bcdb3011(imageId)
+	}
+	return *pvc.Spec.StorageClassName, nil
+}
+
+// Get Storage Class Name from ImageID
+// Given an ImageID, find the corresponding VMImage and figure out it's
+// StorageClass name
+func (h *vmActionHandler) x_bcdb3011(imageId string) (scName string, err error) {
```
WebberHuang1118: What does this function name mean? Is there any “behind the scenes” context or rationale for it?

pkg/api/vm/template.go
```diff
@@ -0,0 +1,472 @@
...
+	// - others, use the pvc StorageClassName
+	vmImageName := getTemplateVersionVMImageName(tvName, index)
+	vmImage := vmImageMap[vmImageName]
+
+	claim := pvcMap[claimName]
+	// target storage class name
+	x_4a322c04 := h.vmio.GetStorageClassName(&vmImage)
+	if _, ok := pvc.Annotations[cdicommon.AnnCreatedForDataVolume]; ok {
+		x_4a322c04 = *claim.Spec.StorageClassName
+	}
```
WebberHuang1118: ditto

pkg/api/vm/template.go
```diff
@@ -0,0 +1,472 @@
...
+			Resources:        claim.Spec.Resources,
+			VolumeMode:       claim.Spec.VolumeMode,
+			StorageClassName: &x_4a322c04,
+		},
+	}
+	return pvc
+}
+
+// Copy Credentials Secret
+func (h *vmActionHandler) x_a0fc0b55(templateVersion *harvesterv1.VirtualMachineTemplateVersion, index int, credential *kubevirtv1.AccessCredential) (err error) {
```
WebberHuang1118: ditto

pkg/api/vm/template.go
```diff
@@ -0,0 +1,472 @@
...
+		vmtv, err = h.createTemplateWithData(vm, vmtv)
+		if err != nil {
+			return err
+		}
+	}
+
+	return h.createSecrets(vmtv, vm)
+}
+
+func (h *vmActionHandler) createTemplateWithData(vm *kubevirtv1.VirtualMachine, vmtvIn *harvesterv1.VirtualMachineTemplateVersion) (vmtv *harvesterv1.VirtualMachineTemplateVersion, err error) {
```
WebberHuang1118: The second parameter isn’t used?

WebberHuang1118 (commented):
Thanks for the enhancement. The implementation looks clear and compact from my perspective—please address or reply to the comments.

I think it’s close to merge. As discussed, please make sure all test cases are covered before merging. I’ll also have testing when I am available.

pkg/api/vm/template.go
m-ildefons: This function used to be called `getStorageClassNameFromImageID`, which was a nice and descriptive name for what the function does.
However, in the last round of review, you asked to shorten the function name: https://github.com/harvester/harvester/pull/9392#discussion_r2693198705
I didn't find a short _and_ descriptive name for the function, so I just took the CRC32 sum of the old name, prefixed with `x_` since CRC32 can start with numbers, which isn't valid in Go.

pkg/api/vm/template.go
m-ildefons: Name shortened similar to https://github.com/harvester/harvester/pull/9392#discussion_r2792559075

pkg/api/vm/template.go
m-ildefons: Name shortened similar to https://github.com/harvester/harvester/pull/9392#discussion_r2792559075

pkg/api/vm/template.go
WebberHuang1118: For the previous comment in [function naming suggestion](https://github.com/harvester/harvester/pull/9392#discussion_r2693198705), that’s my suggestion and it’s open for discussion.

We need a meaningful function name, `getSCNameFromImg` for example.

pkg/api/vm/template.go
WebberHuang1118: same as https://github.com/harvester/harvester/pull/9392#discussion_r2792574338

We need a meaningful variable name.

pkg/api/vm/template.go
WebberHuang1118: same as https://github.com/harvester/harvester/pull/9392#discussion_r2792574338

We need a meaningful function name.

pkg/api/vm/template.go
m-ildefons: > We need a meaningful function name.

Yes. Would `getStorageClassNameFromImageID` work ok?

pkg/api/vm/template.go
m-ildefons: How about `targetStorageClassName`?

pkg/api/vm/template.go
m-ildefons: Any suggestions? Maybe `copyCredentialsSecret`?

pkg/api/vm/template.go
m-ildefons: Fixed. In several places in the body of this method `vmtv` (which is a return value) is used in place of `vmtvIn`.
To avoid further confusion, I've re-named the second parameter to `vmtv` and the return value is now named `out` and declared as variable in the variable block instead of the function signature.

pkg/api/vm/template.go
WebberHuang1118: LGTM for `copyCredentialsSecret`.

pkg/api/vm/template.go
Vicente-Cheng: Shortening the function naming would help the whole thing look clean, but not lose readability.
The function name `x_bcdb3011` looks cannot present `getStorageClassNameFromImageID`.

Could you maybe try `getSCNameFromImgID`, something like that?
It would not lose readability and have shorter naming.

WDYT?

pkg/api/vm/template.go
Vicente-Cheng: ditto, maybe `targetSCName`?

pkg/api/vm/template.go
olblak: Could we stop arguing on function name and stay focus to land this pull request?
Naming is hard, and we all agree that `x_bcdb3011` from @m-ildefons is a bad one.
To be honest, I don't understand why `targetStorageClassName` is a bad one neither.
I don't know what too long, really means.

@WebberHuang1118 Feel free to make a suggestion for a better name if you have a strong opinion on the matter, but to me, it seems nit picking.

This kind of [comment](https://github.com/harvester/harvester/pull/9392#discussion_r2791315245) are more constructive  IMHO

pkg/api/vm/template.go
olblak: Please see my comment [above](https://github.com/harvester/harvester/pull/9392#discussion_r2798045181)

pkg/api/vm/template.go
olblak: Please see my comment [above](https://github.com/harvester/harvester/pull/9392#discussion_r2798045181)

pkg/api/vm/template.go
WebberHuang1118: > To be honest, I don't understand why targetStorageClassName is a bad one neither.
> I don't know what too long, really means.
> This kind of https://github.com/harvester/harvester/pull/9392#discussion_r2791315245 are more constructive IMHO

IMO, if it’s an internal variable in the code (e.g., a local variable or an internal struct field), using `targetSCName` is totally reasonable.

And the other two names I mentioned are `getStorageClassNameFromPVCOrBackingImage` and `getStorageClassNameFromImageID`. From my perspective, both are a bit of a mouthful.

However, as I mentioned in [this comment](https://github.com/harvester/harvester/pull/9392#discussion_r2792574338), that’s just my suggestion and it’s open for discussion.

During the earlier conversation, I thought we were still discussing it, so out of respect I didn’t want to come across as telling the author what to do.

That said, if there were any follow-up feedback or further discussion, I’d be happy to share my thoughts.

I don’t have a strong opinion on it—my only point is that the naming should remain meaningful.

m-ildefons: I've renamed the functions/variables in question accordingly.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,474 @@
...
+
+	v := util.GetLonghornDataEngineType(targetSC)
+	if v == longhorn.DataEngineTypeV1 {
+		return harvesterv1.VMIBackendBackingImage, nil
+	}
+	return harvesterv1.VMIBackendCDI, nil
+}
+
+func (h *vmActionHandler) createVMTemplate(namespace string, input CreateTemplateInput) (vmt *harvesterv1.VirtualMachineTemplate, err error) {
+	vmt, err = h.vmTemplateClient.Create(
```
Vicente-Cheng, line 421: Should we consider if the vm template exists?

Think about the following case:
1. calling createTemplate
2. failed with createTemplateWithData
3. defer func failed with cleanupVMTemplate
4. Next calling createTemplate must fail on createVMTemplate (assume they use the same configuration) because the CR was already here

It might be okay with multiple reconciliations, because each time the deferred function will try to clean up.
Just wondering if we can do more checks to create the template instead of doing it directly.
WDYT?

pkg/api/vm/template.go
```diff
@@ -0,0 +1,474 @@
...
+		return nil, err
+	}
+	return vmt, nil
+}
+
+func (h *vmActionHandler) createVMTemplateVersion(namespace string, vm *kubevirtv1.VirtualMachine, vmt *harvesterv1.VirtualMachineTemplate, keyPairIDs []string) (vmtv *harvesterv1.VirtualMachineTemplateVersion, err error) {
+	name := fmt.Sprintf("%s-%s", vmt.Name, rand.String(6))
+	vmtID := fmt.Sprintf("%s/%s", vmt.Namespace, vmt.Name)
+	vmSourceSpec := h.sanitizeVirtualMachineForTemplateVersion(name, vm)
+	vmtv, err = h.vmTemplateVersionClient.Create(
```
Vicente-Cheng, line 445: ditto (similar case)

pkg/api/vm/template.go
```diff
@@ -0,0 +1,474 @@
...
+		vmImageMap         map[string]harvesterv1.VirtualMachineImage
+		out                *harvesterv1.VirtualMachineTemplateVersion
+		err                error
+	)
+
+	pvcMap, pvcStorageClassMap, err = h.getPVCStorageClassMap(vm)
+	if err != nil {
+		return vmtv, err
+	}
+	vmImageMap, err = h.createVMImages(vmtv, vm, pvcStorageClassMap)
```
Vicente-Cheng: Do we need to create the defer function to clean up the create VMImage once the later steps fail?

or we can add the existing checking on the createVMImages?
WDYT?

Vicente-Cheng (commented):
overall lgtm, just some order issues I would like to discuss.

Vicente-Cheng:
BTW, please fix the linter once the PR is ready.

pkg/api/vm/template.go
m-ildefons: If a VM Template with the same name in the same namespace already exists, then the `h.vmTemplateClient.Create(...)` function call at line 397 will return an error. `createTemplate` already handles this case by aborting and propagating the error upwards.
We could of course check whether the error from `h.vmTemplateClient.Create(...)` is a HTTP 409 Conflict error. In that case we could possibly return the existing VM Template without creating a new one. Then the handler can continue with creating a new VM Template Version for that existing VM Template. Would that work?

IMO, in general, there is too much logic in the VM Template handler. A lot of this logic should probably exist in the VM Template controller instead.
However, I think re-designing the VM Template handler and controller is out of scope for this PR.

pkg/api/vm/template.go
m-ildefons: This is a bit tricky, a conflict is unlikely, because of the randomly generated name. But I'll add a check an let the handler retry with a different randomly generated name in case the name is already occupied.

pkg/api/vm/template.go
m-ildefons: We should do a cleanup, since there can be multiple images and it can fail on an image that isn't the first image. In that case all images need to be cleaned up.

m-ildefons:
@Vicente-Cheng @WebberHuang1118 I fixed the linter complaints and I believe I also fixed all review comments. Could you please have another look?

pkg/api/vm/template.go
```diff
@@ -0,0 +1,474 @@
...
+	targetSC, err := h.storageClassCache.Get(targetSCName)
+	if err != nil {
+		return "", err
+	}
+	provisioner := util.GetProvisionedPVCProvisioner(pvc, h.storageClassCache)
+	if provisioner != util.CSIProvisionerLonghorn {
+		return harvesterv1.VMIBackendCDI, nil
+	}
+
+	v := util.GetLonghornDataEngineType(targetSC)
```
w13915984028, line 413: The existing code seems to use `VMIBackendBackingImage` when Provisioner is LH and `dataEngine` is missing, is it a bug and fixed on new PR?

```
			if targetSC.Provisioner == util.CSIProvisionerLonghorn {
				if dataEngineVers, find := targetSC.Parameters["dataEngine"]; find {
					if dataEngineVers == string(longhorn.DataEngineTypeV1) {
						vmImageBackend = harvesterv1.VMIBackendBackingImage
					}
				} else {
					vmImageBackend = harvesterv1.VMIBackendBackingImage
				}
			}
```

Existing code seems to have no chance to hit below block, is it a bug? @Vicente-Cheng 
https://github.com/harvester/harvester/blob/b2c3cbff76859693a98583a8db48c03e5dddcdce/pkg/api/vm/handler.go#L1120-L1122

pkg/api/vm/template.go
```diff
@@ -0,0 +1,474 @@
...
+	return vmtv, nil
+}
+
+// cleanup the VirtualMachineTemplate in case we hold a non-nil reference
+func (h *vmActionHandler) cleanupVMTemplate(vmt *harvesterv1.VirtualMachineTemplate) (err error) {
+	if vmt == nil {
+		return nil
+	}
+
+	if err = h.vmTemplateClient.Delete(vmt.Namespace, vmt.Name, &metav1.DeleteOptions{}); err != nil {
```
w13915984028, line 478: the cleanup needs to skip `IsNotFound` error, as code is cleaning up a couple of objects, some might not be created before

pkg/api/vm/template.go
```diff
@@ -0,0 +1,474 @@
...
+	return nil
+}
+
+// cleanup the VirtualMachineTemplateVersion in case we hold a non-nil reference
+func (h *vmActionHandler) cleanupVMTemplateVersion(vmtv *harvesterv1.VirtualMachineTemplateVersion) (err error) {
+	if vmtv == nil {
+		return nil
+	}
+
+	if err = h.vmTemplateVersionClient.Delete(vmtv.Namespace, vmtv.Name, &metav1.DeleteOptions{}); err != nil {
```
w13915984028, line 498: skip IsNotFound

w13915984028 (commented):
some questions,

and the new file `template.go` looks to refactor some existing code logic, anyway to add some test codes to cover the different cases as more as possible? thanks.

@m-ildefons Do you have a detailed local test report?

The main concern is: the refactored code touchs many normal and abnormal running paths, we need enough tests to ensure they are well covered.

pkg/api/vm/template.go
Vicente-Cheng: In the old format, we did not have `dataEngine` field.
So, 
- `dataEngine` == DataEngineTypeV1
- no `dataEngine` field

These two also present the Longhorn v1 (use backingimage)

Vicente-Cheng: >Existing code seems to have no chance to hit below block, is it a bug? @Vicente-Cheng

yes, this should be redundant. Thanks for pointing that out.

pkg/api/vm/template.go
Vicente-Cheng: sure, fair enough, in theory, it will work with multiple reconciles. We can leave here and wait for the refactor.

pkg/api/vm/template.go
Vicente-Cheng: hmm, which means we might not be able to clean the old `VirtualMachineTemplateVersion`?
If we failed with `cleanupVMTemplate()`, we need to manually clean up, right?

pkg/api/vm/template.go
m-ildefons: I'm not sure I fully understand what you want here.
If I do an explicit lookup with a `h.vmTemplateClient.Get(...)` then I still have a brief period of time between the return of that function and the `h.vmTemplateClient.Delete` call where the object might disappear from the API server.

But if you meant that I should check if `err` contains an `IsNotFound` error after the call to `h.vmTemplateClient.Delete`, that is checked just below this line.

Besides, the VM Template object passed as argument to this function is only non-nil if it has been created before. Not sure why you think it might not exist?

pkg/api/vm/template.go
m-ildefons: see above

m-ildefons:
Thanks for review, Jian. I'll add more unit-test cases and make the code simpler.

I did run the manual tests a while ago and at the time the PR worked as expected. However it has seen quite a few changes since, so I'll re-do the testing and report back here.

pkg/api/vm/template.go
```diff
@@ -0,0 +1,547 @@
...
+		name := getTemplateVersionUserPasswordSecretName(templateVersion.Name, index)
+		err = h.copySecret(userPassword.Source.Secret.SecretName, name, templateVersion)
+		return err
+	}
+	return nil
+}
+
+func (h *vmActionHandler) createSecrets(templateVersion *harvesterv1.VirtualMachineTemplateVersion, vm *kubevirtv1.VirtualMachine) error {
+	for index, credential := range vm.Spec.Template.Spec.AccessCredentials {
+		err := h.copyCredentialsSecret(templateVersion, index, &credential)
```
w13915984028, line 271: if any of copyCredentialsSecret, copySecret fails, how to clean up?

pkg/api/vm/template.go
w13915984028: Your code has no problem, I reviewed too fast to notice the `vm, vmt` are local vars to save them, check non `nil` and then cleanup is good.

pkg/api/vm/template.go
Vicente-Cheng: I saw the cleanup, thanks

pkg/api/vm/template.go
Vicente-Cheng: Hi @m-ildefons, 
just want to confirm the above. Thanks

m-ildefons: `cleanupVMTemplate()` is called _after_ `cleanupVMTemplateVersion()` in case something went wrong during the template creation.
Should `cleanupVMTemplateVersion()` fail, then we will indeed leak a resource and require manual intervention. However that case is very unlikely and the only scenario I can think of is if the Kubernetes API server becomes unavailable between the API calls. However in that case we can't do any cleanup anyways.

m-ildefons:
Note that on the current `master` branch, the implementation of the VM template action doesn't clean up or recover from errors at all: https://github.com/harvester/harvester/blob/3de9a7875ad445bbff0a93e34528e4671c887ce9/pkg/api/vm/handler.go#L916

As I've stated above already, I'm not happy with how much logic the VM action handler manages here. I believe it would be better to implement it in the VirtualMachineTemplate and VirtualMachineTemplateVersion controllers. However the redesign of the VM template logic isn't in scope for this PR.

The original goal of this PR is to deal with a specific bug that results from VM images being namespaced resources and storage classes being non-namespaced resources and storage classes being generated with names solely based on the names of VM images.
The only reason this PR touches the VM action handler at all is because during the creation of a VM template, the names of the storage classes for the images of the volumes of the source VM need to be generated with the new naming scheme.

To make it easier for me to deal with the amount of VM action handler code, I've pulled all functions that are specific to the VM template action out into their own file. As you can see, roughly a quarter of the entire VM action handler is just the "create VM template" action.

IMO, the best course of action to move forward is:
- Focus this PR on the resolution of https://github.com/harvester/harvester/issues/5165 only.
- Create an issue and a separate PR for a redesign of the VM action API handler, particularly the "create VM template" action.

Otherwise we will get bogged down by scope-creep and we it will become a nightmare to do QA for this PR.

pkg/api/vm/template.go
m-ildefons: The `copySecret` function sets an ownership reference. This is used by the Kubernetes API server to do cleanup of the dependents (i.e. the secrets) when the owner (i.e. the VM Template Version) is deleted.

The `copyCredentialSecret` function internally calls the `copySecret` function and thereby takes advantage of the ownership mechanism as well.

See: https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

w13915984028:
@m-ildefons Agree with your points. Restrict the scope of this PR will make it easy to deliver and validate. The broad-range rafactor needs more efforts to test. thanks.

pkg/api/vm/template.go
m-ildefons: Thanks for pointing this out. I'll correct the logic such that it also uses the backingimage if the provisioner is Longhorn, but no `dataEngine` is specified.

pkg/util/image.go
m-ildefons: Done.

pkg/api/vm/template.go
m-ildefons: Done

pkg/api/vm/template.go
m-ildefons: Done

pkg/api/vm/template.go
m-ildefons: Fixing the dead code is out of scope for this PR. It'll be a //TODO for a general refactor of the create-VM-template action handler.

pkg/api/vm/template.go
Vicente-Cheng: Yeah, I agree. Just want to confirm that. Thanks!

m-ildefons:
## Reproducer #5165

<details><summary> :green_circle: Install Harvester v1.8.0-dev-20260301 </summary>Screenshot at 2026-03-02 11-38-07</details>

<details><summary> :green_circle: Replace Harvester container image with built from HEAD of `damiasan/5165`</summary>Screenshot at 2026-03-02 11-47-38</details>

<details><summary> :green_circle: Create two images with the same name in two different namespaces</summary>Screenshot at 2026-03-02 11-49-28
Screenshot at 2026-03-02 11-49-52</details>

<details><summary> :green_circle: Delete image `harvester-public/test-image`</summary>
Screenshot at 2026-03-02 11-51-02</details>

<details><summary> :green_circle: Instantiate a VM from the image `default/test-image`</summary>Screenshot at 2026-03-02 11-58-50</details>

## Regression Tests

#### Case 1: VM basic operations (create / snapshot / backup / restore)

<details><summary> :green_circle: Install Harvester v1.8.0-dev-20260301</summary>Screenshot at 2026-03-02 15-32-24</details>

<details><summary> :green_circle: Replace Harvester container image</summary>Screenshot at 2026-03-02 15-35-08</details>

<details><summary> :green_circle: Configure backup target</summary>Screenshot at 2026-03-02 15-50-24</details>

<details><summary> :green_circle: Create VM Image</summary>Screenshot at 2026-03-02 15-36-11</details>

<details><summary> :green_circle: Create VM</summary>Screenshot at 2026-03-02 15-37-45</details>

<details><summary> :green_circle: Create VM snapshot</summary>Screenshot at 2026-03-02 15-57-06</details>

<details><summary> :green_circle: Restore from snapshot to new VM</summary>Screenshot at 2026-03-02 16-01-02</details>

<details><summary> :green_circle: Create VM backup</summary>Screenshot at 2026-03-02 16-01-54</details>

<details><summary> :green_circle: Restore from backup to new VM</summary>Screenshot at 2026-03-02 16-03-44</details>

#### Case 2: An old VMImage can work without error

<details><summary> :green_circle: Set up Harvester v1.4.0</summary>Screenshot at 2026-03-02 13-12-46</details>

<details><summary> :green_circle: Create a VMImage</summary>Screenshot at 2026-03-02 13-16-08</details>

<details><summary> :green_circle: Replace Harvester container image with one built from `damiasan/5165`</summary>Screenshot at 2026-03-02 13-18-10</details>

<details><summary> :green_circle: Instantiate a VM from the image</summary>Screenshot at 2026-03-02 13-29-40</details>

#### Case 3: Create VM Template

<details><summary> :green_circle: Install Harvester v1.8.0-dev-20260301</summary>Screenshot at 2026-03-02 15-32-24</details>

<details><summary> :green_circle: Replace Harvester container image</summary>Screenshot at 2026-03-02 15-35-08</details>

<details><summary> :green_circle: Create VM Image</summary>Screenshot at 2026-03-02 15-36-11</details>

<details><summary> :green_circle: Create VM</summary>Screenshot at 2026-03-02 15-37-45</details>

<details><summary> :green_circle: Create VM Template (with data) from that VM</summary>Screenshot at 2026-03-02 15-39-37</details>

<details><summary> :green_circle: Create new VM from VM Template</summary>Screenshot at 2026-03-02 15-45-04</details>

## Summary

:green_circle: #5165 does not reproduce with this PR applied
:green_circle: No regression found in VM creation, VM template, VM snapshot or VM backup functions

pkg/api/vm/template.go
```diff
@@ -0,0 +1,580 @@
...
+
+	defer func() {
+		var cleanupErr error
+
+		logrus.WithFields(logrus.Fields{
+			"namespace": vm.Namespace,
+			"name":      vm.Name,
+		}).Errorf("failed to create VM images while trying to create VM Template with data: %v", err)
+
+		if err == nil {
```
w13915984028: should L96 be moved to L90?

pkg/api/vm/template.go
```diff
@@ -0,0 +1,580 @@
...
+	pvcMap = make(map[string]corev1.PersistentVolumeClaim)
+
+	for _, volume := range vm.Spec.Template.Spec.Volumes {
+		if volume.PersistentVolumeClaim == nil {
+			continue // Skip if the volume does not reference a PVC
+		}
+
+		pvc, err = h.pvcCache.Get(vm.Namespace, volume.PersistentVolumeClaim.ClaimName)
+		if err != nil {
+			return pvcMap, err
```
w13915984028: to double check: in the for loop, if `h.pvcCache.Get` on one volume fails, the returned `pvcMap` is not empty, it contains the already successful fetched pvcs, is this expected? normally, the return is `nil, err` in error case

e.g.  another map return `nil, err`
https://github.com/harvester/harvester/pull/9392/changes#diff-ec4341351bd981bf249ec7f83755ef6a5c353c3d3e9186886cfd87297eea10fbR336

pkg/api/vm/template.go
```diff
@@ -0,0 +1,580 @@
...
+
+// Get Storage Class name from PVC or Backing Image
+// If the PVC has an annoation, which associates it with a BackingImage, then
+// this function will return the StorageClass name of the BackingImage. Otherwise it
+// will return the StorageClass name of the PVC itself.
+func (h *vmActionHandler) getSCNameFromPVC(pvc *corev1.PersistentVolumeClaim) (scName string, err error) {
+	if imageId, ok := pvc.Annotations[util.AnnotationImageID]; ok {
+		return h.getSCNameFromImgID(imageId)
+	}
+	return *pvc.Spec.StorageClassName, nil
```
w13915984028, line 159: many other code checks this pointer when use it, should it be followed here, or is it checked before?

```
pkg/controller/master/backup/backup.go:	if pvc.Spec.StorageClassName == nil {
...
pkg/util/volume.go:		if pvc.Spec.StorageClassName == nil {
```

pkg/api/vm/template.go
```diff
@@ -0,0 +1,580 @@
...
+	return pvcMap, nil
+}
+
+// Get Storage Class name from PVC or Backing Image
+// If the PVC has an annoation, which associates it with a BackingImage, then
+// this function will return the StorageClass name of the BackingImage. Otherwise it
+// will return the StorageClass name of the PVC itself.
+func (h *vmActionHandler) getSCNameFromPVC(pvc *corev1.PersistentVolumeClaim) (scName string, err error) {
+	if imageId, ok := pvc.Annotations[util.AnnotationImageID]; ok {
+		return h.getSCNameFromImgID(imageId)
```
w13915984028, line 152: it might be better to wrap the error message here, in erro case,

```
fmt.Errorf("pvc %s/%s has annotation %s=%s, but get SCName from it failed")...
```

the `getSCNameFromImgID` returns multi possible errors, and `getSCNameFromPVC` also has two branches

pkg/api/vm/template.go
```diff
@@ -0,0 +1,580 @@
...
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"namespace":       templateVersion.Namespace,
+			"templateVersion": templateVersion.Name,
+			"volumeClaim":     volumeClaimTemplates,
+		}).Error("Failed to json.Marshal volume claim templates")
+		return err
+	}
+
+	if vmSource.ObjectMeta.Annotations == nil {
+		vmSource.ObjectMeta.Annotations = map[string]string{}
```
w13915984028: general usage is to make a map

```
toUpdate.Annotations = make(map[string]string)
```

pkg/api/vm/template.go
```diff
@@ -0,0 +1,580 @@
...
+	// - longhorn v1, we need to use the new storageclass Name (for backingImage)
+	// - others, use the pvc StorageClassName
+	vmImageName := getTemplateVersionVMImageName(tvName, index)
+	vmImage := vmImageMap[vmImageName]
+
+	claim := pvcMap[claimName]
+	// target storage class name
+	targetSCName := h.vmio.GetStorageClassName(&vmImage)
+	if _, ok := pvc.Annotations[cdicommon.AnnCreatedForDataVolume]; ok {
+		targetSCName = *claim.Spec.StorageClassName
```
w13915984028: similar: is `claim.Spec.StorageClassName != nil` checked before?

w13915984028 (commented):
Some last questions/suggestions, thanks.

Vicente-Cheng (dismissed):
lgtm, thanks for the huge efforts

pkg/api/vm/template.go
m-ildefons: 🤦‍♂️ yes, you're right, the log message should only be emitted in case of failure.

pkg/api/vm/template.go
m-ildefons: This originated from the existing code.
https://github.com/harvester/harvester/blob/e892272f8ea20eb320d9fdcc3e7d4b03ac666fe6/pkg/api/vm/handler.go#L1050

I agree with you that this should not return a partially constructed map. However the caller discards the results in case of an error anyways, so there is no functional defect here.

pkg/api/vm/template.go
m-ildefons: Why would it be better? The caller never compares the error, so it always does the same, no matter from which branch in `getSCNameFromImgID` we return.

I think extracting the error, wrapping it and returning just adds unnecessary verbosity to the code here.

w13915984028: it is more for easy troubleshooting, the function name is `getSCNameFromPVC`, but the returned error is only about `image id...`, with above `pvc %s/%s has annotation %s=%s, but get SCName from it failed`, it describes the necessary hierarchy info

sure, I don't mind it keeps unchanged, it is just a suggestion

pkg/api/vm/template.go
m-ildefons: This is again a leftover from the original code, which blindly trusts that all PVCs have a storage class name set:
https://github.com/harvester/harvester/blob/e892272f8ea20eb320d9fdcc3e7d4b03ac666fe6/pkg/util/volume.go#L54

This assumption isn't too terrible in the context of Harvester, where manually created volumes are rare, but it's not 100% watertight.
I'll add a nil-check to prevent crashes, but the canonically correct logic would also have to deal with the case that the `.Spec.StorageClassName` can be `""` and that there can be a default storage class in case the pointer is nil.

However fixing this really isn't related to #5165 and we should deal with it in a separate PR.

w13915984028: it could be optimized on another PR/issues

as those code blocks are shown to be `modified` on github PR review window, reviewer is hard to know if it is copied or new

the best practice is still to split bug fix and refator

w13915984028 (dismissed):
LGTM, thanks for the massive fix and improvements.

pkg/api/vm/template.go
m-ildefons: This function just copies the existing settings from the PVC to a PVC template.
In case `.Spec.StorageClassName` is `nil`, the PVC is set up to fall back to the default storage class. Therefore the PVC template also needs to receive the nil pointer.

I'll rewrite the function to make this more obvious.

w13915984028 (approved):
LGTM, thanks.

WebberHuang1118 (commented):
> #### Case 3: Create VM Template
> 
> <details><summary> :green_circle: Install Harvester v1.8.0-dev-20260301</summary>Screenshot at 2026-03-02 15-32-24</details>
> 
> <details><summary> :green_circle: Replace Harvester container image</summary>Screenshot at 2026-03-02 15-35-08</details>
> 
> <details><summary> :green_circle: Create VM Image</summary>Screenshot at 2026-03-02 15-36-11</details>
> 
> <details><summary> :green_circle: Create VM</summary>Screenshot at 2026-03-02 15-37-45</details>
> 
> <details><summary> :green_circle: Create VM Template (with data) from that VM</summary>Screenshot at 2026-03-02 15-39-37</details>
> 
> <details><summary> :green_circle: Create new VM from VM Template</summary>Screenshot at 2026-03-02 15-45-04</details>

From my perspective, for this case we should further verify whether the source VM contains secrets from SSH credentials, cloud-init, and network data, and whether a new VM created from the template correctly carries over the secret contents. For example:

* Create a VM with SSH credentials, cloud-init, and network data.
* Write some data to the VM.
* Generate a template from that VM.
* Create a new VM from the template.
* Verify the new VM has the correct SSH credentials, cloud-init, network data, and the data you wrote.

m-ildefons:
<details><summary> :green_circle: Install Harvester v1.8.0-dev-20260301</summary>Screenshot at 2026-03-06 11-04-17</details>

<details><summary> :green_circle: Replace Harvester container image</summary>Screenshot at 2026-03-06 11-12-29</details>

<details><summary> :green_circle: Create SSH Keys</summary>Screenshot at 2026-03-06 11-14-05</details>

<details><summary> :green_circle: Create VM Image</summary>Screenshot at 2026-03-06 11-15-59</details>

<details><summary> :green_circle: Create VM with SSH credentials, clout-init and network data</summary>
Screenshot at 2026-03-06 11-18-32
Screenshot at 2026-03-06 11-19-01
Screenshot at 2026-03-06 11-19-10
Screenshot at 2026-03-06 11-23-36
</details>

<details><summary> :green_circle: Write a canary to the VM disk</summary>Screenshot at 2026-03-06 11-25-11</details>

<details><summary> :green_circle: Create new VM Template (with data) from that VM</summary>
Screenshot at 2026-03-06 11-25-59
Screenshot at 2026-03-06 11-28-10
Screenshot at 2026-03-06 11-28-22
</details>

<details><summary> :green_circle: Create new VM from that VM template</summary>
Screenshot at 2026-03-06 11-29-39
Screenshot at 2026-03-06 11-30-31
</details>

<details><summary> :green_circle: Confirm SSH credentials</summary>
Screenshot at 2026-03-06 11-32-25
</details>

<details><summary> :green_circle: Confirm cloud-init data</summary>
Screenshot at 2026-03-06 11-33-16
</details>

<details><summary> :green_circle: Confirm network data</summary>
Screenshot at 2026-03-06 11-35-06
</details>

<details><summary> :green_circle: Confirm canary file</summary>
Screenshot at 2026-03-06 11-35-45
</details>

WebberHuang1118 (approved):

PR #9395, opened 2025-10-31
feat: add migration-based hotplug NIC

wheatdog:
#### Problem:

Add migration-based NIC hot-plugging / hot-unplugging to Harvester

#### Solution:

Follow the method from [KubeVirt's documentation -  Migration based hotplug](https://kubevirt.io/user-guide/network/hotplug_interfaces/#migration-based-hotplug). Update the VM spec and do the migration.

These 3 actions are added for VM Custom Resource:
- addNic
- removeNic
- findHotunpluggableNics

To prevent unexpected RestartRequired condition [caused by our current MAC address preserving mechanism](https://github.com/harvester/harvester/pull/9295#discussion_r2443964405) (https://github.com/harvester/harvester/pull/8339), MAC addresses would be generated automatically during VM creation. This ensures that VM and VMI would both have the MAC addresses in spec.

For existing VMs, Their MAC addresses would be backfilled from annotation to spec while stopping or restarting. Until then, those VMs would be considered un-hot-pluggable.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8283
https://github.com/harvester/harvester/issues/7042

#### Test plan:

1. MAC addresses should be generated successfully during VM creation
2. Hotplug NICs to a VM that has restart before shouldn't trigger unexpected RestartRequired condition
3. `addNic`, `removeNic` and `findHotunpluggableNics` shouldn't be allowed when the VM is not live-migratable or is in transition state like `migrating`, `starting`.
4. Can hot-plug to a running VM
5. Can hot-unplug to a running VM

Additionally, please refer to the test plan in HEP: https://github.com/harvester/harvester/pull/9295

https://kubevirt.io/user-guide/network/hotplug_interfaces/

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -129,6 +129,9 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 
 		if canHotplugNic(vm) {
 			resource.AddAction(request, addNic)
+		}
+
+		if canHotUnplugNic(vm) {
 			resource.AddAction(request, removeNic)
 			resource.AddAction(request, findHotunpluggableNics)
 		}
@@ -477,4 +480,11 @@ func canHotplugNic(vm *kubevirtv1.VirtualMachine) bool {
 		return false
 	}
 	return virtualmachine.SupportHotplugNic(vm)
+}
+
+func canHotUnplugNic(vm *kubevirtv1.VirtualMachine) bool {
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
+		return false
+	}
+	return virtualmachine.SupportHotUnplugNic(vm)
 }
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -41,6 +41,9 @@ const (
 	createTemplate                   = "createTemplate"
 	addVolume                        = "addVolume"
 	removeVolume                     = "removeVolume"
+	addNic                           = "addNic"
+	removeNic                        = "removeNic"
+	findHotunpluggableNics           = "findHotunpluggableNics"
 	cloneVM                          = "clone"
 	forceStopVM                      = "forceStop"
 	dismissInsufficientResourceQuota = "dismissInsufficientResourceQuota"
@@ -122,6 +125,12 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 
 		if canCPUAndMemoryHotplug(vm) {
 			resource.AddAction(request, cpuAndMemoryHotplug)
+		}
+
+		if canHotplugNic(vm) {
+			resource.AddAction(request, addNic)
+			resource.AddAction(request, removeNic)
+			resource.AddAction(request, findHotunpluggableNics)
 		}
 	}
 
@@ -461,4 +470,11 @@ func canCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
 		}
 	}
 	return !hasRestartRequiredOrHotplugMigration
+}
+
+func canHotplugNic(vm *kubevirtv1.VirtualMachine) bool {
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
+		return false
+	}
+	return virtualmachine.SupportHotplugNic(vm)
 }
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -476,15 +476,9 @@ func canCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
 }
 
 func canHotplugNic(vm *kubevirtv1.VirtualMachine) bool {
-	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
-		return false
-	}
 	return virtualmachine.SupportHotplugNic(vm)
 }
 
 func canHotUnplugNic(vm *kubevirtv1.VirtualMachine) bool {
-	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
-		return false
-	}
 	return virtualmachine.SupportHotUnplugNic(vm)
 }
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -476,9 +476,11 @@ func canCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
 }
 
 func canHotplugNic(vm *kubevirtv1.VirtualMachine) bool {
-	return virtualmachine.SupportHotplugNic(vm)
+	ok, _ := virtualmachine.SupportHotplugNic(vm)
+	return ok
 }
 
 func canHotUnplugNic(vm *kubevirtv1.VirtualMachine) bool {
-	return virtualmachine.SupportHotUnplugNic(vm)
+	ok, _ := virtualmachine.SupportHotUnplugNic(vm)
+	return ok
 }
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -1254,6 +1254,8 @@ func (h *vmActionHandler) addNic(ctx context.Context, namespace, name string, in
 		return err
 	}
 
+	logrus.Infof("VM %s/%s has interface %s backed by network %s hot-plugged", namespace, name, input.InterfaceName, input.NetworkName)
+
 	// https://kubevirt.io/user-guide/network/hotplug_interfaces/#migration-based-hotplug
 	// Although it's not required to manually migrate the VM as mentioned in the document,
 	// we still immediately call the migration here for better UX instead of waiting KubeVirt to discover and reconcile.
@@ -1294,6 +1296,8 @@ func (h *vmActionHandler) removeNic(ctx context.Context, namespace, name string,
 	if err != nil {
 		return err
 	}
+
+	logrus.Infof("VM %s/%s has interface %s hot-unplugged", namespace, name, input.InterfaceName)
 
 	// https://kubevirt.io/user-guide/network/hotplug_interfaces/#migration-based-hotplug
 	// Although it's not required to manually migrate the VM as mentioned in the document,
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -1243,7 +1243,7 @@ func (h *vmActionHandler) addNic(ctx context.Context, namespace, name string, in
 	if err != nil {
 		return err
 	}
-	return h.migrate(ctx, namespace, name, "")
+	return nil
 }
 
 func (h *vmActionHandler) getHotunpluggableNetworks(vm *kubevirtv1.VirtualMachine) (map[string]struct{}, error) {
@@ -1312,7 +1312,7 @@ func (h *vmActionHandler) removeNic(ctx context.Context, namespace, name string,
 	if err != nil {
 		return err
 	}
-	return h.migrate(ctx, namespace, name, "")
+	return nil
 }
 
 // findHotunpluggableNics return a list of NIC names that could be hot-unplugged
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -19,7 +19,6 @@ import (
 	"github.com/rancher/wrangler/v3/pkg/schemas/validation"
 	"github.com/rancher/wrangler/v3/pkg/slice"
 	"github.com/sirupsen/logrus"
-	cnitypes "github.com/containernetworking/cni/pkg/types"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -53,6 +52,7 @@ import (
 	"github.com/harvester/harvester/pkg/util"
 	"github.com/harvester/harvester/pkg/util/drainhelper"
 	"github.com/harvester/harvester/pkg/util/virtualmachineinstance"
+	"github.com/harvester/harvester/pkg/util/virtualmachine"
 	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 )
 
@@ -1191,12 +1191,12 @@ func (h *vmActionHandler) getVmNetwork(name, vmNamespaceFallback string) (*cniv1
 }
 
 func isVmNetworkHotpluggable(nad *cniv1.NetworkAttachmentDefinition) (bool, error) {
-	var conf cnitypes.PluginConf
-	if err := json.Unmarshal([]byte(nad.Spec.Config), &conf); err != nil {
+	conf, err := util.DecodeNadConfigToNetConf(nad)
+	if err != nil {
 		return false, err
 	}
 
-	if conf.Type != "bridge" {
+	if !conf.IsBridgeCNI() {
 		return false, nil
 	}
 
@@ -1246,37 +1246,6 @@ func (h *vmActionHandler) addNic(ctx context.Context, namespace, name string, in
 	return nil
 }
 
-func (h *vmActionHandler) getHotunpluggableNetworks(vm *kubevirtv1.VirtualMachine) (map[string]struct{}, error) {
-	namespace := vm.ObjectMeta.Namespace
-	validNetworks := make(map[string]struct{})
-	for _, network := range vm.Spec.Template.Spec.Networks {
-		if network.Multus != nil {
-			nad, err := h.getVmNetwork(network.Multus.NetworkName, namespace)
-			if err != nil {
-				return nil, err
-			}
-
-			ok, err := isVmNetworkHotpluggable(nad)
-			if err != nil {
-				return nil, err
-			}
-			if ok {
-				validNetworks[network.Name] = struct{}{}
-			}
-		}
-	}
-
-	return validNetworks, nil
-}
-
-func isInterfaceHotunpluggable(iface kubevirtv1.Interface, hotunpluggableNetworks map[string]struct{}) bool {
-	_, isNetworkHotunpluggable := hotunpluggableNetworks[iface.Name]
-	if iface.Model == "virtio" && iface.Bridge != nil && iface.State != kubevirtv1.InterfaceStateAbsent && isNetworkHotunpluggable {
-		return true
-	}
-	return false
-}
-
 // removeNic remove a hotplug NIC by its interface name
 func (h *vmActionHandler) removeNic(ctx context.Context, namespace, name string, input RemoveNicInput) error {
 	vm, err := h.vmCache.Get(namespace, name)
@@ -1284,9 +1253,8 @@ func (h *vmActionHandler) removeNic(ctx context.Context, namespace, name string,
 		return err
 	}
 
-	hotunpluggableNetworks, err := h.getHotunpluggableNetworks(vm)
-	if err != nil {
-		return err
+	if !virtualmachine.SupportHotUnplugNic(vm) {
+		return fmt.Errorf("%s/%s doesn't support HotUnplugNic", namespace, name)
 	}
 
 	vmCopy := vm.DeepCopy()
@@ -1295,7 +1263,7 @@ func (h *vmActionHandler) removeNic(ctx context.Context, namespace, name string,
 	interfaces := vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces
 	for idx := range interfaces {
 		if interfaces[idx].Name == input.InterfaceName {
-			if isInterfaceHotunpluggable(interfaces[idx], hotunpluggableNetworks) {
+			if virtualmachine.IsInterfaceHotUnpluggable(interfaces[idx]) {
 				interfaces[idx].State = kubevirtv1.InterfaceStateAbsent
 				found = true
 				break
@@ -1322,16 +1290,13 @@ func (h *vmActionHandler) findHotunpluggableNics(rw http.ResponseWriter, namespa
 		return err
 	}
 
-	hotunpluggableNetworks, err := h.getHotunpluggableNetworks(vm)
-	if err != nil {
-		return err
-	}
-
-	iterfaces := vm.Spec.Template.Spec.Domain.Devices.Interfaces
-	ifaceNames := make([]string, 0, len(iterfaces))
-	for _, iface := range iterfaces {
-		if isInterfaceHotunpluggable(iface, hotunpluggableNetworks) {
-			ifaceNames = append(ifaceNames, iface.Name)
+	ifaceNames := make([]string, 0)
+	if virtualmachine.SupportHotUnplugNic(vm) {
+		iterfaces := vm.Spec.Template.Spec.Domain.Devices.Interfaces
+		for _, iface := range iterfaces {
+			if virtualmachine.IsInterfaceHotUnpluggable(iface) {
+				ifaceNames = append(ifaceNames, iface.Name)
+			}
 		}
 	}
 	resp := FindHotunpluggableNicsOutput{
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -1238,7 +1238,7 @@ func (h *vmActionHandler) addNic(ctx context.Context, namespace, name string, in
 	newIface := kubevirtv1.Interface{
 		Name:                   input.InterfaceName,
 		MacAddress:             input.MacAddress,
-		Model:                  "virtio",
+		Model:                  kubevirtv1.VirtIO,
 		InterfaceBindingMethod: kubevirtv1.InterfaceBindingMethod{Bridge: &kubevirtv1.InterfaceBridge{}},
 	}
 	vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces = append(vmCopy.Spec.Template.Spec.Domain.Devices.Interfaces, newIface)
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -19,6 +19,7 @@ import (
 	"github.com/rancher/wrangler/v3/pkg/schemas/validation"
 	"github.com/rancher/wrangler/v3/pkg/slice"
 	"github.com/sirupsen/logrus"
+	cnitypes "github.com/containernetworking/cni/pkg/types"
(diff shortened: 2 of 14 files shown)


pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -606,3 +617,33 @@ func patchDefaultCPU(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) []s
...
+func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) (types.PatchOps, error) {
+	if vm == nil || vm.Spec.Template == nil {
+		return patchOps, nil
+	}
+
+	MacAddressInAnnotation := make(map[string]string)
+	if raw, exists := vm.ObjectMeta.Annotations[util.AnnotationMacAddressName]; exists {
+		err := json.Unmarshal([]byte(raw), &MacAddressInAnnotation)
+		if err != nil {
+			logrus.Warnf("unable to unmarshal %s: %s", util.AnnotationMacAddressName, raw)
```
w13915984028, line 630: return earlier

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -606,3 +617,33 @@ func patchDefaultCPU(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) []s
...
+		}
+	}
+
+	for idx, iface := range vm.Spec.Template.Spec.Domain.Devices.Interfaces {
+		if iface.MacAddress == "" {
+			if mac, exists := MacAddressInAnnotation[iface.Name]; exists {
+				logrus.Infof("MAC address (%s) has already been observed in the annotation for %s", mac, iface.Name)
```
w13915984028: logrus.Debugf

pkg/api/vm/formatter.go
```diff
@@ -123,6 +126,12 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 		if canCPUAndMemoryHotplug(vm) {
 			resource.AddAction(request, cpuAndMemoryHotplug)
 		}
+
+		if canHotplugNic(vm) {
+			resource.AddAction(request, addNic)
+			resource.AddAction(request, removeNic)
+			resource.AddAction(request, findHotunpluggableNics)
```
w13915984028, line 136: When a VM is in migrating, `hot-plug add/remove` is better to be blocked, as you can't start two migrations on a same VM in parallel,

the `removeNic` could be even strict, iterate the NICs to ensure at least one removable `bridged` interface is there?

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,167 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+		Name:          input.InterfaceName,
+		NetworkSource: kubevirtv1.NetworkSource{Multus: &kubevirtv1.MultusNetwork{NetworkName: input.NetworkName}},
+	}
+	vmCopy.Spec.Template.Spec.Networks = append(vmCopy.Spec.Template.Spec.Networks, newNetwork)
+
+	_, err = h.vms.Update(vmCopy)
+	if err != nil {
+		return err
+	}
+	return h.migrate(ctx, namespace, name, "")
```
w13915984028: as we have several sources of `migration source` now (cpu/memory hot-plug, nic hog-plug, original VM migration action), the `migrate()` function seems need to check if any existing migration is still running

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,167 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+	return nad, nil
+}
+
+func isVmNetworkHotpluggable(nad *cniv1.NetworkAttachmentDefinition) (bool, error) {
+	var conf cnitypes.PluginConf
+	if err := json.Unmarshal([]byte(nad.Spec.Config), &conf); err != nil {
+		return false, err
+	}
+
+	if conf.Type != "bridge" {
```
w13915984028: suggest to encap a public function if possible

https://github.com/harvester/network-controller-harvester/blob/e58fad96f48ad838f8f8f1a207e85967d1a2e32a/pkg/utils/nad.go#L482

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,167 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+			if isInterfaceHotunpluggable(interfaces[idx], hotunpluggableNetworks) {
+				interfaces[idx].State = kubevirtv1.InterfaceStateAbsent
+				found = true
+				break
+			}
+			return fmt.Errorf("interface %s is not hot-unpluggable", input.InterfaceName)
+		}
+	}
+
+	if !found {
```
w13915984028, line 1291: if the `found` nic is the last nic of the VM, would suggest to block it

w13915984028 (commented):
Some questions, thanks.

w13915984028:
btw, please check below code, which is about the UI menu rendering, the `migrate` and `abortMigration` occurs at most one.

the `addnic`, `removenic` might also need some `mutually exclusive` considerations, to improve the UI

https://github.com/harvester/harvester/blob/363164e7ed3665333808272fc12bfb4fbba816a4/pkg/api/vm/formatter.go#L119-L130

pkg/api/vm/handler.go
ibrokethecloud: do we need to explicitly call migrate here? my understanding was kubevirt will trigger the migration when a nic is hot plugged to a VM

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,167 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+}
+
+// removeNic remove a hotplug NIC by its interface name
+func (h *vmActionHandler) removeNic(ctx context.Context, namespace, name string, input RemoveNicInput) error {
+	vm, err := h.vmCache.Get(namespace, name)
+	if err != nil {
+		return err
+	}
+
+	hotunpluggableNetworks, err := h.getHotunpluggableNetworks(vm)
```
ibrokethecloud: during a nic removal we do not really need to check if the underlying network is bridge. we just need to check if we are using the bridge interface or not using a custom binding, and just remove the interface

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,167 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+
+	if !found {
+		return fmt.Errorf("interface %s doesn't exist", input.InterfaceName)
+	}
+
+	_, err = h.vms.Update(vmCopy)
+	if err != nil {
+		return err
+	}
+	return h.migrate(ctx, namespace, name, "")
```
ibrokethecloud, line 1305: same as before, this may not be needed.

pkg/api/vm/handler.go
wheatdog: Explicit migration is needed since we are using the [migration-based hotplug](https://kubevirt.io/user-guide/network/hotplug_interfaces/#migration-based-hotplug).

pkg/api/vm/handler.go
wheatdog: ditto. Explicit migration is needed since we are using the [migration-based hotplug](https://kubevirt.io/user-guide/network/hotplug_interfaces/#migration-based-hotplug).

pkg/api/vm/handler.go
w13915984028: the `migration/abortMigration` is added to vm pop-up menu;  `cpu memory hot-(un)plug`  is triggered via `edit vm` menu;  the `nic hot-(un)plug` menu looks to be similar with `migrate` menu

if/not explicitly call `migrate` is a good question, @wheatdog did you test and notice the differences?

pkg/api/vm/formatter.go
wheatdog: > When a VM is in migrating, hot-plug add/remove is better to be blocked, as you can't start two migrations on a same VM in parallel,

Agree. I think this should already be checked inside `canMigrate` in L122
```
func canMigrate(nodeCache ctlcorev1.NodeCache, vmi *kubevirtv1.VirtualMachineInstance) bool {
	if vmi == nil || vmi.DeletionTimestamp != nil || vmi.Annotations[util.AnnotationMigrationState] != "" {
		return false
	}

```

> the removeNic could be even strict, iterate the NICs to ensure at least one removable bridged interface is there?

Good catch. Let me add it.

pkg/api/vm/handler.go
ibrokethecloud: https://kubevirt.io/user-guide/network/hotplug_interfaces/#migration-based-hotplug points out that it is not needed

pkg/controller/master/virtualmachine/vmi_network_controller.go
```diff
@@ -85,3 +98,52 @@ func (h *VMNetworkController) updateVMDefaultNetworkMacAddress(vmi *kubevirtv1.V
 
 	return nil
 }
+
+func (h *VMNetworkController) updateVMMacAddressFromAnnotation(vmi *kubevirtv1.VirtualMachineInstance) error {
+	vm, err := h.vmCache.Get(vmi.Namespace, vmi.Name)
+	if err != nil {
+		return err
+	}
+
```
ibrokethecloud, line 107: we should have a check if vm has a deletion timestamp set, in which we can return immediately.

This will handle the case when a VMI deletion is triggered due to deletion of the VM

pkg/api/vm/handler.go
wheatdog: Oops, I miss that. Thanks for the note! I'll remove the explicit migration call and try it.

pkg/api/vm/handler.go
wheatdog: @ibrokethecloud has pointed out that the document mention that we don't need need the explicit migration call in our setup (`LiveUpdate` roll-out strategy, version >= v1.6).

I'll try it and see if it's working.

wheatdog: I have tested it without explicit migration call and it works. I'll remove it.

pkg/api/vm/handler.go
wheatdog: Thanks, it seems that that function is not visible yet in the vendor folder. Do you think I should [bump the dependency](https://github.com/harvester/harvester/blob/e209e335f6f929ac99eb66417fea993fa45b2982/go.mod#L80) `harvester-network-controller`? Or I can just create a similar function here myself.

w13915984028: create a similar one :)

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,167 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
 	})
 }
 
+func (h *vmActionHandler) getVmNetwork(name, vmNamespaceFallback string) (*cniv1.NetworkAttachmentDefinition, error) {
+	nadNamespacedName := kubevirtmultus.NetAttachDefNamespacedName(vmNamespaceFallback, name)
```
w13915984028, line 1185: Harvester webhook tries below solution, are they doing same job?

https://github.com/harvester/harvester/blob/363164e7ed3665333808272fc12bfb4fbba816a4/pkg/webhook/resources/virtualmachine/mutator.go#L413-L422

pkg/webhook/resources/virtualmachine/mutator.go
wheatdog: It's intended to not return earlier if the mac address annotation is malformed. In such case, mutator can still generate mac addresses and patch it.

pkg/api/vm/handler.go
Yu-Jack: Just a thought. If we don't really need explicit migration here, we can put a comment/link for the note because it's an implicit migration. I think it's helpful for future debug.

pkg/api/vm/handler.go
wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/9e42071238bf8e96c534364828e398b03e2b8794

pkg/api/vm/handler.go
wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/9e42071238bf8e96c534364828e398b03e2b8794

pkg/api/vm/handler.go
wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/9e42071238bf8e96c534364828e398b03e2b8794

pkg/api/vm/handler.go
wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/56a511c55aa3e631a61dc91ae12c3874aa591166

pkg/api/vm/handler.go
wheatdog: Thanks @Yu-Jack, addressed in https://github.com/harvester/harvester/commit/56a511c55aa3e631a61dc91ae12c3874aa591166 and https://github.com/harvester/harvester/pull/9395/commits/aeff31f8872b7387a69d10cde656c1fe399cf739 (for comment/link)

pkg/webhook/resources/virtualmachine/mutator.go
wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/81a398bdd6c031ad3f2305e8a43bf7b758519cd3

wheatdog:
> btw, please check below code, which is about the UI menu rendering, the `migrate` and `abortMigration` occurs at most one.
> 
> the `addnic`, `removenic` might also need some `mutually exclusive` considerations, to improve the UI
> 
> https://github.com/harvester/harvester/blob/363164e7ed3665333808272fc12bfb4fbba816a4/pkg/api/vm/formatter.go#L119-L130

@w13915984028 I think `addnic` and `removenic` are not mutually exclusive. Actually, only `addnic` would be present in the menu of kebab button while `removenic` is shown in the network tab of view mode VM page. You can refer to [this UI PR](https://github.com/harvester/harvester-ui-extension/pull/582).

pkg/api/vm/formatter.go
wheatdog: addressed in https://github.com/harvester/harvester/commit/9e42071238bf8e96c534364828e398b03e2b8794

pkg/controller/master/virtualmachine/vmi_network_controller.go
wheatdog: Sounds good. Addressed in https://github.com/harvester/harvester/pull/9395/commits/02522d77f047db1a6a7501277c095897aedda353

pkg/api/vm/handler.go
wheatdog: Yes. I reused the helper `kubevirtmultus.NetAttachDefNamespacedName` from KubeVirt.

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,140 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+	nad, err := h.nadCache.Get(nadNamespacedName.Namespace, nadNamespacedName.Name)
+	if err != nil {
```
w13915984028, line 1187: to be even secure, please check if nad.DeletionTimestamp != nil

plus, please ensure some special nads like `stoarge-network`, `migration-network` nad are not used  (I guess they might be blocked by webhook), from here or `isVmNetworkHotpluggable` or thers

UI can filter such special nad, but the directly api operation might be able to pass them


network-controller has checks upon special nads, and `migration-network` is missing;

harvester webhook seems not checking them, let me double check

https://github.com/harvester/network-controller-harvester/blob/e58fad96f48ad838f8f8f1a207e85967d1a2e32a/pkg/webhook/nad/validator.go#L113-L116

pkg/api/vm/handler_test.go
```diff
@@ -801,3 +802,77 @@ func Test_vmActionHandler_findMigratableNodesByVMI(t *testing.T) {
 		})
 	}
 }
+
+func Test_isVmNetworkHotpluggable(t *testing.T) {
```
w13915984028, line 807: add test cases to filter out `storage-network`, `migration-network` nads.

w13915984028 (commented):
some additional questions, thanks.

pkg/api/vm/handler.go
w13915984028: It seems harvester webhook does not check the special nads, we might need to handle it on another PR

https://github.com/harvester/harvester/blob/363164e7ed3665333808272fc12bfb4fbba816a4/pkg/webhook/resources/virtualmachine/validator.go#L221-L235

wheatdog: I'll also check how we distinguish VM Networks from special nads.

w13915984028: refer; if required, we can copy or refer code from network-controller

https://github.com/harvester/network-controller-harvester/blob/e58fad96f48ad838f8f8f1a207e85967d1a2e32a/pkg/utils/nad.go#L625-L641

w13915984028: BTW, from the context of VM, filter out NADs from namespace `harvester-system` seems to be fine, those are reserved NADs.

wheatdog: I see. Sounds good. Let me fix it. Thanks.

wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/bf265c0420ce1d6cc0d52ba70cdde040cc51401b

pkg/api/vm/handler_test.go
wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/bf265c0420ce1d6cc0d52ba70cdde040cc51401b

pkg/api/vm/formatter.go
```diff
@@ -462,3 +474,17 @@ func canCPUAndMemoryHotplug(vm *kubevirtv1.VirtualMachine) bool {
 	}
 	return !hasRestartRequiredOrHotplugMigration
 }
+
+func canHotplugNic(vm *kubevirtv1.VirtualMachine) bool {
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
```
w13915984028: Is it duplicated with the check `canMigrate->ValidateVMMigratable` ?


https://github.com/harvester/harvester/blob/f16f02037000971c3a4fa3bd46a6c80593e183ae/pkg/util/virtualmachineinstance/virtualmachineinstance.go#L73-L75

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,150 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+	if !ok {
+		return fmt.Errorf("network %s is not hot-pluggable", input.NetworkName)
+	}
+
+	vmCopy := vm.DeepCopy()
+
+	newIface := kubevirtv1.Interface{
+		Name:                   input.InterfaceName,
+		MacAddress:             input.MacAddress,
+		Model:                  "virtio",
```
w13915984028: define `virtio` as a const ? it occurs a couple of times

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,150 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+		return err
+	}
+	if !ok {
+		return fmt.Errorf("network %s is not hot-pluggable", input.NetworkName)
+	}
+
+	vmCopy := vm.DeepCopy()
+
+	newIface := kubevirtv1.Interface{
+		Name:                   input.InterfaceName,
```
w13915984028, line 1239: what happens if user add an already binded `InterfaceName`, namely bind the same network twice, this is generally not allowed

w13915984028 (commented):
Thanks for the update, a few questions.

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,150 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+		Name:          input.InterfaceName,
+		NetworkSource: kubevirtv1.NetworkSource{Multus: &kubevirtv1.MultusNetwork{NetworkName: input.NetworkName}},
+	}
+	vmCopy.Spec.Template.Spec.Networks = append(vmCopy.Spec.Template.Spec.Networks, newNetwork)
+
+	_, err = h.vms.Update(vmCopy)
+	if err != nil {
+		return err
+	}
+
```
w13915984028, line 1256: add log/event to indicate the hotplug action; hotunplug is similar

also consider to add an annotation to the vm to record the last hotplug & hot-unplug operations (two annotations), when, which nic, it will be helpful for troubleshooting

pkg/api/vm/handler.go
```diff
@@ -1164,6 +1181,150 @@ func (h *vmActionHandler) removeVolume(ctx context.Context, namespace, name stri
...
+}
+
+// removeNic remove a hotplug NIC by its interface name
+func (h *vmActionHandler) removeNic(ctx context.Context, namespace, name string, input RemoveNicInput) error {
+	vm, err := h.vmCache.Get(namespace, name)
+	if err != nil {
+		return err
+	}
+
+	if !virtualmachine.SupportHotUnplugNic(vm) {
```
w13915984028: As there is a couple of causes can trigger `SupportHotUnplugNic` to return false, it is better to return `bool, error`, and here L1270 encap the error to return message for better troubleshooting

the return message will be like `default/vm1 doesn't support HotUnplugNic as it is a guest node vm/it has no bridged nic/it has only one nic/...`

w13915984028 (commented):
thanks.

wheatdog:
> https://github.com/harvester/harvester/pull/9395#discussion_r2493468272
> https://github.com/harvester/harvester/pull/9395#discussion_r2493580524

@w13915984028 @ibrokethecloud @Yu-Jack  regarding to the explicit call to `migrate`, I have second thought after playing around with UI. You can refer to the following two clips. I think the UX is better with explicit call to `migrate`. I prefer having it back. What do you think? Thanks.

### With explicit call to `migrate`
The VM status can be quickly updated to `migrating` or `pending migration`. IMO, it's better.

https://github.com/user-attachments/assets/f72fce16-10c6-4874-b3b6-829ada9f3e55


### Without explicit call to `migrate`

(Please ignore the storage network in the dropdown list from the clip, that would be fixed and it's irrelevant to this discussion.)

The VM would keep as `running` status for a while. This feels slow and could be error-prone if users do additional actions within this period since they aren't blocked like in migrating status. 

https://github.com/user-attachments/assets/87c6e741-9458-4207-9cb7-f6ac3ce3f545

w13915984028:
Good observation. Without the explicit calling to `migrate`, kubevirt seems to take time to response, it is a bit difficult for user, unless we do some additional cosmetics on the UI.

pkg/api/vm/formatter.go
wheatdog: They are different (`vm` vs `vmi`) but could be related. Though I think it should be safe to remove this check in both `canHotplugNic` and `canHotUnplugNic`.

pkg/api/vm/handler.go
wheatdog: Yes, this would be blocked by admission webhook from KubeVirt. No addition code needed from our side.

image

pkg/api/vm/formatter.go
wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/b2e43a9cd9b7c565e39442a9937e5127037761d2

pkg/api/vm/handler.go
wheatdog: Sure. I'll reuse the const `VirtIO` from KubeVirt. Addressed in https://github.com/harvester/harvester/pull/9395/commits/07c5aed3e337c7560f409671a602f32e8cc9a061

pkg/api/vm/handler.go
wheatdog: addressed in https://github.com/harvester/harvester/pull/9395/commits/85275a4b6c4285b5d57790d1fabd5187bf1ca356

pkg/api/vm/handler.go
wheatdog: As [the observation](https://github.com/harvester/harvester/pull/9395#issuecomment-3501334788), I have added it back in https://github.com/harvester/harvester/pull/9395/commits/d551f90695a69631c7896bd410368c642eaaf2d8

pkg/api/vm/handler.go
wheatdog: As [the observation](https://github.com/harvester/harvester/pull/9395#issuecomment-3501334788), I have added it back in https://github.com/harvester/harvester/pull/9395/commits/d551f90695a69631c7896bd410368c642eaaf2d8

pkg/api/vm/handler.go
wheatdog: Thanks, I think adding some logs some be enough for now. Addressed in https://github.com/harvester/harvester/pull/9395/commits/dc226c5705e7dfcce175c10b6bfb78b6a3f36ed3

w13915984028 (approved):
LGTM, thanks.

wheatdog:
@mergify backport v1.7

wheatdog:
As suggested by @Vicente-Cheng, I dug into KubeVirt's code to find out why it took a while for KubeVirt to trigger the auto migration.

According to the 2nd clip in https://github.com/harvester/harvester/pull/9395#issuecomment-3501334788, it took 25 seconds (from 0:15 to 0:40). I believe this is related to this enhancement https://github.com/kubevirt/kubevirt/pull/14259

Those 25 seconds might be contributed by
- 15 seconds of `DynamicNetworkControllerGracePeriod`: https://github.com/kubevirt/kubevirt/pull/14259/files#diff-513cdb230d096b30280705ead0601d957bbdd658cf85a1017a4ea7759763feadR973
- 10 seconds of `pendingMigrationReEvalPeriod`: https://github.com/kubevirt/kubevirt/pull/14259/files#diff-bb2848a511d9143139b4c3b6f718cd9d3f1b057ab0873740814e4c94880381a2R35

In our current use case, since we don't have Multus Dynamic Network Controller installed, I think it should be safe to directly call migration before auto-migration kick in.

cc @w13915984028 @ibrokethecloud

PR #9406, opened 2025-11-03
upgrade: add logger to clarify if we did not set the WaitRebootLabel

Vicente-Cheng:
#### Problem:
We need logger to ensure we encounter the edge case.

#### Solution:
Add the logger to the upgrade controller. 

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8442


starbops (approved):
LGTM, thank you.

Issue #9431, opened 2025-11-05
[BUG] `descheduler` replica set for addon when enabled on single node harvester v1.7.0-rc1 sticks around in CrashLoopBackOff


irishgordo:
### Describe the Bug

Single node Harvester v1.7.0-rc1, enable addon descheduler, the descheduler fails in crashloopbackoff.

### To Reproduce

1. v1.7.0-rc1, single node (had mgmt set to 802.3AD w/ 4 links, but I doubt that is necessary for reproduce) (also had harvester.install.wipeAllDisks=true set at GRUB, but again, doubt that is necessary for reproduce)
2. enable descheduler

### Expected Behavior

not be in crash loop backoff

### Support Bundle for Troubleshooting

[descheduler-support-bundle.zip](https://github.com/user-attachments/files/23375592/descheduler-support-bundle.zip)

### Environment

- Harvester version: v1.7.0-rc1

### Additional context

<img width="2007" height="2002" alt="Image" src="https://github.com/user-attachments/assets/8de84be6-5147-4a06-92b3-8e6192bf29d5" />
<img width="2009" height="1879" alt="Image" src="https://github.com/user-attachments/assets/4088111e-0d71-426c-b4b2-c32070dba0ba" />

### Workaround and Mitigation

n/a

irishgordo:
<img width="1477" height="423" alt="Image" src="https://github.com/user-attachments/assets/c103e25b-d91c-4000-9206-7c038698a2b9" />

Single Node, denies as expected ^


Multi node looks good too

<img width="2222" height="1957" alt="Image" src="https://github.com/user-attachments/assets/dabd777e-62b3-4f5b-99eb-852a7c5e38c9" />

All on v1.7.0-rc2

PR #9434, opened 2025-11-06
feat: add webhook for descheduler addon

FrankYang0529:
#### Problem:
Single node cannot enable descheduler. It will show error like `"The cluster size is 0 or 1 meaning eviction causes service disruption or degradation. So aborting.."`.

#### Solution:
Add webhook to deny the request.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9431

#### Test plan:
1. Use single node to enable descheduler addon. It should be denied.
2. Use multiple nodes to enable descheduler addon. There should not have error.

### DIFF
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -18,6 +18,7 @@ import (
 	"github.com/harvester/harvester/pkg/util/logging"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
+	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/webhook/resources/addon/validator.go
```diff
@@ -112,6 +115,14 @@ func (v *addonValidator) validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon
 		return v.validateRancherLoggingAddon(newAddon)
 	}
 
+	if newAddon.Name == util.DeschedulerName && newAddon.Spec.Enabled {
+		if newAddon.Annotations != nil && newAddon.Annotations[util.AnnotationSkipRancherLoggingAddonWebhookCheck] == "true" {
```
wheatdog: it seems that this check is incorrect. 

```
newAddon.Annotations[util.AnnotationSkipRancherLoggingAddonWebhookCheck] == "true" 
```

FrankYang0529: You're right. Updated it. Thank you.

Vicente-Cheng (approved):

Issue #8930, opened 2025-08-19
[TASK] v170 Bump rancher-monitoring and rancher-logging


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Bump those two addons to the matched versions.

1. Bump monitoring, logging addon version on addon repo
2. Update image list on harvester-installer
3. Handle upgrade path on harvester
4. Other minor fix

Refer the steps and test plans on https://github.com/harvester/harvester/issues/7404

### Additional context

tests/tasks:

- [x] Versions and Related patches are updated on addon repo https://github.com/harvester/addons/pull/158, https://github.com/harvester/addons/pull/159, https://github.com/harvester/addons/pull/162
- [x] All images are included in installer repo https://github.com/harvester/harvester-installer/pull/1147
- [x] Monitoring works on Harvester, cluster level and VM level metrics,  test resultes updated on 
- [x] Monitoring works on Rancher Manager -> Virtualization Management, cluster level and VM level metrics, test result updated on https://github.com/harvester/harvester-installer/pull/1147
- [x] Logging works on Harvester, simple http server can receive logs, test results are posted on https://github.com/harvester/harvester-installer/pull/1147#issue-3433555689
- [x] Upgrade runs smoothly, need to patch issue https://github.com/harvester/harvester/issues/4958 ( fixed by https://github.com/harvester/addons/pull/129 & https://github.com/harvester/harvester/pull/9203 )
- [x] Other issues https://github.com/harvester/harvester/issues/9201 ( fixed by https://github.com/harvester/addons/pull/160 & https://github.com/harvester/harvester/pull/9203 )

_No response_

Sub-task of https://github.com/harvester/harvester/issues/8882

w13915984028:
Test plan, refer task description https://github.com/harvester/harvester/issues/8930#issue-3333271241

```
Enable/disable monitoring & logging addon on v170 are sucessfuly

Monitoring works on Harvester, cluster level and VM level metrics

Monitoring works on Rancher Manager -> Virtualization Management, cluster level and VM level metrics

Logging works on Harvester, simple http server &/ other log servers can receive logs

Upgrade runs smoothly, with/without monitoring,logging enabled on v160/v161, upgrade to v170 is successfuly

```

noahgildersleeve:
- [x] Enable/disable monitoring & logging addon on v170 are sucessfuly
- [x] Monitoring works on Harvester, cluster level and VM level metrics
[...]

PR #9438, opened 2025-11-06
Webhooks to protect addon and managedchart

w13915984028:
#### Problem:

https://github.com/harvester/harvester/issues/9289 had a fix https://github.com/harvester/harvester/pull/9304, but a scenario is missing https://github.com/harvester/harvester/issues/8930#issuecomment-3499021408

When `rancher-logging` is enabled before upgrade;  while `apply_manifest`, the addon is updated to new version, and the `enabled` is kept, this will hit the current webhook and is blocked.

#### Solution:

1. If rancher-logging is enabled and upgradeLog uses it, block disabling
2. If rancher-logging is disabled and upgradeLog is running, block enabling
3. Protect `rancher-monitoring`, `rancher-logging` and other non-experimental addons from being deleted
4. Protect additional managedchart `rancher-logging-crd` and `rancher-monitoring-crd` from being deleted

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9289

#### Test plan:

Per test plan on https://github.com/harvester/harvester/issues/9289 and
https://github.com/harvester/harvester/issues/8930

Deletion protection upon `managedchart`, can on be validated via kubectl:

```
$ kubectl delete managedchart -n fleet-local harvester
The request is invalid: : Delete managedchart fleet-local/harvester is prohibited

$ kubectl delete managedchart -n fleet-local harvester-crd
The request is invalid: : Delete managedchart fleet-local/harvester-crd is prohibited

$ kubectl delete managedchart -n fleet-local rancher-monitoring-crd
The request is invalid: : Delete managedchart fleet-local/rancher-monitoring-crd is prohibited

$ kubectl delete managedchart -n fleet-local rancher-logging-crd
The request is invalid: : Delete managedchart fleet-local/rancher-logging-crd is prohibited
```

Deletion protection upon `addon`, can only be validated via kubectl:

`non-experimental` addons cannot be deleted.

```
$ kubectl delete addons.harvesterhci.io -n cattle-monitoring-system rancher-monitoring
Error from server (BadRequest): admission webhook "validator.harvesterhci.io" denied the request: cattle-monitoring-system/rancher-monitoring addon cannot be deleted

$ kubectl delete addons.harvesterhci.io -n cattle-logging-system rancher-logging
Error from server (BadRequest): admission webhook "validator.harvesterhci.io" denied the request: cattle-logging-system/rancher-logging addon cannot be deleted

$ kubectl delete addons.harvesterhci.io -n harvester-system pcidevices-controller
Error from server (BadRequest): admission webhook "validator.harvesterhci.io" denied the request: harvester-system/pcidevices-controller addon cannot be deleted
```

Test the `enable/disable` upon `rancher-logging` w/wo upgrade:

Local test:

(1) Enable an empty rancher-logging, pass

(2) Create a dangling clusterflow before enable it, get such error, pass (existing features)
image

Test when upgrading v170->v170 cluster;  or v16*->v170 cluster after `harvester` has been upgraded.

(3.1) Enable `rancher-logging`, enable `upgradeLog`

During the upgrade,  disable rancher-logging addon when upgradeLog has already been there, webhook blocks it, if rancher-logging is disabled, then the `upgradeLog` will break.

image

(3.2) Disable `rancher-logging`, enable `upgradeLog`

During the upgrade,  enable the `rancher-logging` addon when upgradeLog has already been there, webhook blocks it, if rancher-logging is enabled, then the `helm install` job will fail as a managedchart based logging-operator has been deployed.

image

### DIFF
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -62,6 +62,7 @@ func (v *addonValidator) Resource() types.Resource {
 		OperationTypes: []admissionregv1.OperationType{
 			admissionregv1.Create,
 			admissionregv1.Update,
+			admissionregv1.Delete,
 		},
 	}
 }
@@ -109,12 +110,23 @@ func (v *addonValidator) validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon 
 		return validateVClusterAddon(newAddon)
 	}
 
-	if newAddon.Name == util.RancherLoggingName && newAddon.Spec.Enabled {
-		if newAddon.Annotations != nil && newAddon.Annotations[util.AnnotationSkipRancherLoggingAddonWebhookCheck] == "true" {
+	if newAddon.Name == util.RancherLoggingName {
+		if newAddon.Annotations[util.AnnotationSkipRancherLoggingAddonWebhookCheck] == "true" {
 			return nil
 		}
 
-		return v.validateRancherLoggingAddon(newAddon)
+		// check when addon is `enabled`
+		//   block if upgradeLog has deployed a managedchart as logging-operator
+		if !oldAddon.Spec.Enabled && newAddon.Spec.Enabled {
+			return v.validateEnableRancherLoggingAddon(newAddon)
+		}
+
+		// check when addon is `disabled`
+		//   block if upgradeLog sits on top of rancher-logging addon
+		if oldAddon.Spec.Enabled && !newAddon.Spec.Enabled {
+			return v.validateDisableRancherLoggingAddon(newAddon)
+		}
+
 	}
 
 	if newAddon.Name == util.DeschedulerName && newAddon.Spec.Enabled {
@@ -163,7 +175,7 @@ func validateVClusterAddon(newAddon *v1beta1.Addon) error {
 	return werror.NewBadRequest(fmt.Sprintf("invalid fqdn %s provided for %s addon", addonContent.Hostname, vClusterAddonName))
 }
 
-func (v *addonValidator) validateRancherLoggingAddon(newAddon *v1beta1.Addon) error {
+func (v *addonValidator) validateEnableRancherLoggingAddon(newAddon *v1beta1.Addon) error {
 	loger := logging.NewLogging(v.flowCache, v.outputCache, v.clusterFlowCache, v.clusterOutputCache)
 
 	if err := loger.FlowsDangling(newAddon.Namespace); err != nil {
@@ -174,17 +186,60 @@ func (v *addonValidator) validateRancherLoggingAddon(newAddon *v1beta1.Addon) er
 		return werror.NewBadRequest(fmt.Sprintf("%s, fix or delete it before enabling addon", err.Error()))
 	}
 
+	// when rancher-logging is disabled, then upgradeLog deploys a managedchart as the logging operator
+	// block the enabling until upgradeLog is gone to avoid issues during addon helm install
+	upgradeLogRunning, namespacedName, err := v.isUpgradeLogRunning()
+	if err != nil {
+		return err
+	}
+
+	if upgradeLogRunning {
+		return werror.NewBadRequest(fmt.Sprintf("%v addon cannot be enabled as upgradeLog %v exists in the cluster, fix or delete before enabling addon", util.RancherLoggingName, namespacedName))
+	}
+
+	return nil
+}
+
+func (v *addonValidator) validateDisableRancherLoggingAddon(newAddon *v1beta1.Addon) error {
+	// if rancher-logging is enabled, then upgradeLog utilizes it
+	// block the disabling until upgradeLog is gone
+	upgradeLogRunning, namespacedName, err := v.isUpgradeLogRunning()
+	if err != nil {
+		return err
+	}
+
+	if upgradeLogRunning {
+		return werror.NewBadRequest(fmt.Sprintf("%v addon cannot be disabled as upgradeLog %v exists in the cluster, fix or delete before disabling addon", util.RancherLoggingName, namespacedName))
+	}
+
+	return nil
+}
+
+func (v *addonValidator) isUpgradeLogRunning() (bool, string, error) {
 	// validate no `upgradeLog` CRs exist as they deployed rancher-logging as a managedchart
 	// so we need to block enablement to avoid issues during addon helm install
 	upgradeLogList, err := v.upgradeLogCache.List(util.HarvesterSystemNamespaceName, labels.Everything())
 	if err != nil {
-		return werror.NewBadRequest(fmt.Sprintf("error list upgradelog objects: %v", err.Error()))
+		return false, "", werror.NewBadRequest(fmt.Sprintf("error list upgradeLog objects: %v", err.Error()))
 	}
 
 	if len(upgradeLogList) > 0 {
-		return werror.NewBadRequest("rancher-logging addon cannot be enabled as upgradeLog objects exist in the cluster, fix or delete before enabling addon")
+		return true, fmt.Sprintf("%s/%s", upgradeLogList[0].Namespace, upgradeLogList[0].Name), nil
 	}
 
+	return false, "", nil
+}
+
+func (v *addonValidator) Delete(_ *types.Request, oldObj runtime.Object) error {
+	oldAddon := oldObj.(*v1beta1.Addon)
+	if oldAddon == nil {
+		return nil
+	}
+	// don't allow delete non-experimental addons
+	//  strictly protect rancher-monitoring and rancher-logging
+	if oldAddon.Name == util.RancherLoggingName || oldAddon.Name == util.RancherMonitoringName || oldAddon.Labels[util.AddonExperimentalLabel] != "true" {
+		return werror.NewBadRequest(fmt.Sprintf("%v/%v addon cannot be deleted", oldAddon.Namespace, oldAddon.Name))
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


w13915984028:
@mergify backport v1.7

pkg/webhook/resources/addon/validator.go
```diff
@@ -104,7 +104,10 @@ func (v *addonValidator) validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon
 		return validateVClusterAddon(newAddon)
 	}
 
-	if newAddon.Name == util.RancherLoggingName && newAddon.Spec.Enabled {
+	// check when addon is `enabled`
+	//  skip the `enabling` when addon controller updates addon
+	//  skip the `re-enable` on the upgrade path, the already enabled addon is updated with new version
```
ibrokethecloud: minor nit with regards to the comment, technically we are skipping the check because addon is already enabled so the upgrade log condition is not really valid as upgrade log will use the underlying rancher monitoring stack.
As a result update to addon should be allowed.

ibrokethecloud (dismissed):
thanks for the fix. lgtm.

w13915984028:
I am checking, we need some further enhancements, will be added.

(1) When `rancher-logging` is enabled, and `upgradeLog` is enabled too, the latter sits on top of `rancher-logging`, then `rancher-logging` can't be disabled on the upgrade phase.

(2) Those initial addons like `rancher-logging`, `rancher-monitoring` should be protected from being deleted. They are not in the webhook now.

https://github.com/harvester/harvester/blob/f16f02037000971c3a4fa3bd46a6c80593e183ae/pkg/webhook/resources/addon/validator.go#L58-L59

w13915984028:
Code is ready for final review.

Now, the enhanced webhook protections are:

- If rancher-logging is enabled and upgradeLog uses it (upgrade is ongoing, upgradeLog is enabled), block disabling
- If rancher-logging is disabled and upgradeLog is running, block enabling
- Protect `rancher-monitoring`, `rancher-logging` and other `non-experimental addons` from being deleted
- Protect additional managedcharts `rancher-logging-crd` and `rancher-monitoring-crd` from being deleted

case 1
image

image

case 2

image



When an upgrade is running, with rancher-logging & upgradeLog enabled, it has:

```
$ kubectl get upgradelog -A
NAMESPACE          NAME                            UPGRADE
harvester-system   hvst-upgrade-v4b7f-upgradelog   hvst-upgrade-v4b7f

$ kubectl get upgrade.harvesterhci -A
NAMESPACE          NAME                 AGE
harvester-system   hvst-upgrade-v4b7f   2m4s


$ kubectl get logging -A
NAME                                  LOGGINGREF                     CONTROLNAMESPACE        WATCHNAMESPACES   PROBLEMS
hvst-upgrade-v4b7f-upgradelog-infra   harvester-upgradelog           harvester-system                          // benefits from `rancher-logging`
rancher-logging-kube-audit            harvester-kube-audit-log-ref   cattle-logging-system                     
rancher-logging-root                                                 cattle-logging-system        
```

pkg/webhook/resources/addon/validator.go
```diff
@@ -174,17 +194,60 @@ func (v *addonValidator) validateRancherLoggingAddon(newAddon *v1beta1.Addon) er
...
+}
+
+func (v *addonValidator) Delete(_ *types.Request, oldObj runtime.Object) error {
+	oldAddon := oldObj.(*v1beta1.Addon)
+	if oldAddon == nil {
+		return nil
+	}
+	// don't allow delete non-experimental addons
+	//  strictly protect rancher-monitoring and rancher-logging
+	if oldAddon.Name == util.RancherLoggingName || oldAddon.Name == util.RancherMonitoringName || oldAddon.Labels[util.AddonExperimentalLabel] != "true" {
```
martindekov, line 248: Similar to enable/disable I though delete should be rejected in specific cases, here we just prohibit deletion once created for the monitor/loggin. Do we think there is scenario where user/process might want to delete those and lower the restriction a bit?

pkg/webhook/resources/addon/validator.go
```diff
@@ -174,17 +194,60 @@ func (v *addonValidator) validateRancherLoggingAddon(newAddon *v1beta1.Addon) er
...
+	}
+
+	if upgradeLogRunning {
+		return werror.NewBadRequest(fmt.Sprintf("%v addon cannot be enabled as upgradeLog %v exists in the cluster, fix or delete before enabling addon", util.RancherLoggingName, namespacedName))
+	}
+
+	return nil
+}
+
+func (v *addonValidator) validateDisableRancherLoggingAddon(newAddon *v1beta1.Addon) error {
```
martindekov, line 211: Is the general idea to reject user intervention to enable/disable of the addon/s in case upgrade is running?

Would there be a case where upgrade itself attempts to enable/disable those addons? I see we don't distinguish some sort of token with which the API is called?

pkg/webhook/resources/addon/validator.go
```diff
@@ -109,12 +112,29 @@ func (v *addonValidator) validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon
...
 	}
 
-	if newAddon.Name == util.RancherLoggingName && newAddon.Spec.Enabled {
-		if newAddon.Annotations != nil && newAddon.Annotations[util.AnnotationSkipRancherLoggingAddonWebhookCheck] == "true" {
+	if newAddon.Name == util.RancherLoggingName {
+		if oldAddon.Spec.Enabled == newAddon.Spec.Enabled {
+			// spec `enabled` is not changed
 			return nil
 		}
+		skip := newAddon.Annotations[util.AnnotationSkipRancherLoggingAddonWebhookCheck] == "true"
```
martindekov, line 120: What sets this annotation? Is this some way to distinguish that upgrade process triggered the enable/disable and not a user?

martindekov (commented):
The implementation looks good Jian, left some questions to understand what we try to achieve

pkg/webhook/resources/addon/validator.go
w13915984028: this is an escalation annotation for user, for debugging/troubleshooting, to enable/disable it any way, even though it might bring issues;  

even with webhook, user can still delete the webhook to do some operaitons, this annotation let the operation be easier

pkg/webhook/resources/addon/validator.go
w13915984028: addon is enabled/disabled by user;  upgrade does not enable/disable it, but check if it is there then use it; or deploy another managedchart for temporary usage, it avoids the conflict with user operation as more as possible

in general, we suggest user to avoid `configurating` the cluster like create VM, enable/disable addons inbetween the upgrade, and those webhooks protect the conflicting points.

pkg/webhook/resources/addon/validator.go
w13915984028: monitoring, logging addons are pre-installed with some `rendered` values by installer, if user deletes and then re-creates it, it is difficult for humen to `render` the values correctly

the `deletion` and `re-creation` on those pre-installed addons, might cause unnecessary issues, hence block the deletion atm.

pkg/webhook/resources/addon/validator.go
```diff
@@ -174,17 +194,60 @@ func (v *addonValidator) validateRancherLoggingAddon(newAddon *v1beta1.Addon) er
...
 
+	// when rancher-logging is disabled, then upgradeLog deploys a managedchart as the logging operator
+	// block the enabling until upgradeLog is gone to avoid issues during addon helm install
+	upgradeLogRunning, namespacedName, err := v.isUpgradeLogRunning()
+	if err != nil {
+		return err
+	}
+
+	if upgradeLogRunning {
+		return werror.NewBadRequest(fmt.Sprintf("%v addon cannot be enabled as upgradeLog %v exists in the cluster, fix or delete before enabling addon", util.RancherLoggingName, namespacedName))
```
martindekov: What does fix or delete mean? Can we change with something like:
```
addon cannot be enabled as upgradeLog %v exists in the cluster wait for upgrade to finish
```
or
```
addon cannot be enabled as upgradeLog %v exists in the cluster delete the object to proceed or wait for upgrade to finish
```

I was thinking that the webhook message should provide with something meaningful and actionable to the user

pkg/webhook/resources/addon/validator.go
```diff
@@ -174,17 +194,60 @@ func (v *addonValidator) validateRancherLoggingAddon(newAddon *v1beta1.Addon) er
...
+func (v *addonValidator) validateDisableRancherLoggingAddon(newAddon *v1beta1.Addon) error {
+	// if rancher-logging is enabled, then upgradeLog utilizes it
+	// block the disabling until upgradeLog is gone
+	upgradeLogRunning, namespacedName, err := v.isUpgradeLogRunning()
+	if err != nil {
+		return err
+	}
+
+	if upgradeLogRunning {
+		return werror.NewBadRequest(fmt.Sprintf("%v addon cannot be disabled as upgradeLog %v exists in the cluster, fix or delete before disabling addon", util.RancherLoggingName, namespacedName))
```
martindekov: Same as above

martindekov (dismissed):
LGTM, Jian small comments regarding the webhook message content and I think it's good to go. I think we can improve the message with more actionable suggestions for the user.

pkg/webhook/resources/addon/validator.go
w13915984028: Updated the message to:

>addon cannot be enabled as upgradeLog %v exists in the cluster, wait until the Harvester upgrade is finished or removed

martindekov (approved):

Issue #9414, opened 2025-11-04
[BUG] Second node cannot join with Rancher v2.13.0-alpha4


FrankYang0529:
### Describe the Bug

In Harvester, the second node cannot join the cluster. There is error in rancher-system-agent v0.3.14-rc.4 (e725216).

```
Nov 04 02:06:20 harvester-node-1 rancher-system-agent[4111]: time="2025-11-04T02:06:20Z" level=info msg="Rancher System Agent version v0.3.14-rc.4 (e725216) is starting"
Nov 04 02:06:20 harvester-node-1 rancher-system-agent[4111]: time="2025-11-04T02:06:20Z" level=info msg="Using directory /var/lib/rancher/agent/work for work"
Nov 04 02:06:20 harvester-node-1 rancher-system-agent[4111]: time="2025-11-04T02:06:20Z" level=info msg="Starting remote watch of plans"
Nov 04 02:06:20 harvester-node-1 rancher-system-agent[4111]: time="2025-11-04T02:06:20Z" level=info msg="Starting /v1, Kind=Secret controller"
Nov 04 02:08:20 harvester-node-1 rancher-system-agent[4111]: I1104 02:08:20.213189    4111 reflector.go:568] "Warning: watch ended with error" reflector="pkg/mod/k8s.io/client-go@v0.34.1/tools/cache/reflector.go:290" type="*v1.Secret" err="an error on the server (\"unable to decode an event from the watch stream: stream error: stream ID 5; INTERNAL_ERROR; received from peer\") has prevented the request from succeeding"
```

ref: https://github.com/rancher/rancher/issues/52592

### To Reproduce

Use https://github.com/harvester/harvester-installer/pull/1173 to start 2-node Harvester cluster.

### Expected Behavior

The second node can join the cluster.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: https://github.com/harvester/harvester-installer/pull/1173
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): KVM
- Rancher version: v2.13.0-alpha4

TachunLin:
Verified **fixed** on  on v1.7-head (25/11/12). Close this issue.

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ ISO installation: Can join two nodes into Harvester cluster $~~$</summary>

1. Create node
[...]

PR #9444, opened 2025-11-10
fix(capi): update namespace to cattle-capi-system

FrankYang0529:
#### Problem:
The CAPI namespace changes from `cattle-provisioning-capi-system` to `cattle-capi-system`, so the Machine / Provisioning Cluster CR cannot be ready.

#### Solution:

Fix the namespace.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9414

#### Test plan:
Create a 3-node cluster successfully.


FrankYang0529:
Close the PR because we prefer the fix in https://github.com/harvester/harvester-installer/pull/1173/commits/2bc0506a1de6121301a06abc7e30dfa24cd93aea.

Issue #8980, opened 2025-08-28
[FEATURE] Pause-ability of node upgrades


starbops:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

Users have requested that we provide more control over the Harvester zero-downtime upgrade. Specifically, the ability to move user workloads among nodes during an upgrade event is crucial. In Harvester's current implementation, user workloads on a to-be-upgraded node are evacuated without allowing users to specify which nodes to live-migrate to. Furthermore, it is likely that the evacuated user workloads cannot run correctly on other nodes due to various reasons, such as CPU model issues. This is a reasonable request, as it allows users to take action before a node is put under the upgrade process, so they have a chance to move the workloads to the desired target nodes.

### Describe the solution you'd like

It consists of three stages in the process of each node upgrade:
1. Pre-draining: Shutting down the impossible-to-live-migrate VMs and evacuating remaining VMs on the node
2. Draining: Draining pods on the node and then upgrade RKE2 on the node
3. Post-draining: Upgrading OS for the node

For each node upgrade, pausing at the beginning of the pre-drain stage has little impact, as it already takes time to check the health of the volumes. However, how this pause-ability is presented and maintains a good user experience is crucial and worth discussing.

The initial plan is to
- Introduce a new upgrade config setting: `mode`. Either being `auto` (default) or `interactive`.
- If the `mode` is `interactive`, make a pause at the pre-draining stage of each node upgrade
- The pause has a default timeout window, say, 10 minutes, for the user to confirm they want to pause the node upgrade and start off any manual tasks, e.g., shut down or live-migrating VMs
- The user continues the node upgrade by actively consent it after they complete the tasks

It's infeasible to pause a node upgrade in the middle of the process, so we go for the pre-pause way.

### Describe alternatives you've considered

Let users to decide the order of node upgrades. This requires exploration Rancher's cluster upgrade capabilities.

### Additional context

Additionally, with the feature being available, it should be allowed for users to upgrade clusters that contain VMs that can't be live-migrated. Which means, we also need to adjust the validating webhook accordingly, depending on the `mode` of the upgrade.

#6145
#8764

forbesguthrie:
Both this and https://github.com/harvester/harvester/issues/5077 are targeted for the 1.7.0 - we should consider the impact on the workflow as they should really compliment each other.  Run https://github.com/harvester/harvester/issues/5077 as a step 0 before pre-draining.

try2codesecure:
+1 for this feature
[...]

PR #9446, opened 2025-11-10
docs(hep): introduce pausable node upgrades [ci skip]

starbops:
#### Related Issue(s):

#8980

### DIFF
--- a/enhancements/20251105-pausable-node-upgrades.md
+++ b/enhancements/20251105-pausable-node-upgrades.md
@@ -1,0 +1,158 @@
+# Pausable Node Upgrades
+
+## Summary
+
+This enhancement allows users to manage upgrades during the node upgrade phase of the Harvester Upgrade process. Specifically, in Harvester Upgrade V1, when a cluster node is transitioning to the pre-draining state, the pre-drain job can only be created after the user actively approves it. Within the time period, the user can perform various operations, such as manually live-migrating the virtual machines from the to-be-pre-drained node to other suitable nodes, or simply turning off the virtual machines at the last minute (they no longer need to be shut off at the beginning of an upgrade).
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/8980
+
+## Motivation
+
+There are different types of user workloads. Some of these workloads are not eligible for live migration. In such cases, they will inevitably be shut down during the Harvester Upgrade process, particularly during the node upgrade phase. Such behavior/requirement is less acceptable to users, as they cannot know the order of node upgrades in advance and don't have enough time to shut down the virtual machines that will be affected gracefully. Moreover, many users prefer to handle an upgrade event manually rather than let it execute autonomously. They would like to nudge the workload migration based on node-specific criteria. For instance, moving workloads to a more capable node. All of this stems from the inability to manually intervene in node upgrade operations and the overly simplistic virtual machine eviction mechanism.
+
+Note: In recent Harvester versions, virtual machines that cannot be live-migrated must be turned off before an upgrade can start, resulting in significant downtime for user workloads and services.
+
+### Goals
+
+- Enable per-upgrade opt-in for pausable node upgrades
+- Automatically pause node upgrade before pre-drain job creation
+- Allow per-node manual approval to proceed with pre-drain
+
+### Non-goals [optional]
+
+- Provide the ability to pause node upgrades at any time
+- Support pausing during other upgrade phases
+- Provide an interface specifying the node upgrade order explicitly or skipping the upgrade for any node
+- Modify the pre-drain/drain/upgrade logic itself
+- Implement automated user workload placement decisions
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Manual Virtual Machine Migration During Pause
+
+As a cluster administrator, I want to manually migrate virtual machines while an upgrade for a node is paused so that I can move workloads to more capable nodes based on my knowledge and preference.
+
+#### Story 2: Last-Minute VM Shutdown
+
+As a cluster administrator with virtual machines requiring graceful shutdown, I want to shut down virtual machines right before node pre-drain so that I minimize downtime (virtual machines don't need to be off during entire upgrade).
+
+### User Experience In Detail
+
+The user can set the node upgrade **mode** to be `auto` or `manual`. The default value is `auto`, which suggests the node upgrade behavior we're all familiar with: nodes are upgraded one by one autonomously. If the mode is set to `manual`, the user can further set the pausability of upgrades for specific nodes in the **pauseNodes** field. Nodes not in the **pauseNodes** field will still undergo node upgrades automatically; if no nodes are specified in **pauseNodes**, all nodes will be paused before entering the pre-draining state, provided the mode is `manual`. When the **mode** is set to `auto` or not being set at all, the values in the **pauseNodes** are completely ignored.
+
+For instance, the following setting will pause upgrades for nodes `node-2`, `node-5`, and `node-9`. Other nodes in the cluster will still be upgraded autonomously.
+
+```json
+{
+  "nodeUpgradeOption": {
+    "strategy": {
+      "mode": "manual",
+      "pauseNodes": [
+        "node-2",
+        "node-5",
+        "node-9"
+      ]
+    }
+  }
+}
+```
+
+Another example, the following setting will pause all the nodes in the cluster before they enters the pre-draining state.
+
+```json
+{
+  "nodeUpgradeOption": {
+    "strategy": {
+      "mode": "manual"
+    }
+  }
+}
+```
+
+When the user triggers an upgrade for the cluster by clicking the "Upgrade" button on the dashboard UI or manually creating the Upgrade custom resource, the relevant setting will be automatically applied. Note that it will not take effect if the user configures the setting after the upgrade has started. During the node upgrade phase, the user can observe each node's upgrade progress in the Upgrade custom resource and in the upgrade dialog.
+
+After the user completes the tasks and is ready to continue with the node upgrade, they can remove the annotation from the node or click the "Resume" button on the upgrade dialog to proceed.
+
+### API changes
+
+There are no actual API changes involved.
+
+## Design
+
+### Implementation Overview
+
+The implementation includes two major parts: a set of new setting fields and the node annotation manipulation. Users configure the `nodeUpgradeOption` fields in the `upgrade-config` setting to specify their node upgrade preferences. The upgrade controller then annotates the nodes marked for pause according to the setting when the upgrade is initiated.
+
+During the node upgrade phase, the secret controller reconciles the machine-plan secrets. When the time comes (when the machine-plan secret for a node is annotated as eligible to upgrade), it checks whether the node has the annotation before creating the pre-drain job. When the annotation is removed from the node, the secret controller is triggered again and reconciles the machine-plan secrets to create the corresponding pre-drain job, thus continuing the original node upgrade flow.
+
+In theory, a successful upgrade should have no node-upgrade-pause annotations left on the node objects (users should have removed them because they consented to proceed with the node upgrade). For cases such as failed or aborted upgrades (where the Upgrade custom resource is removed in the middle of an upgrade), the annotation must be cleaned up automatically by the upgrade controller.
+
+### Test plan
+
+Due to the Harvester Upgrade implementation, single-node cluster upgrades are executed differently from other cluster configurations; it's required to test at least both.
+
+The following assumes the master-head version already includes the proposed enhancement.
+
+#### Single-node cluster upgrade
+
+1. Prepare a single-node cluster with the master-head version
+1. Configure the `upgrade-config` setting:
+   ```json
+   {
+     "nodeUpgradeOption": {
+       "mode": "manual"
+     }
+   }
+   ```
+1. Trigger an upgrade using the same master-head version ISO image
+1. The upgrade should be paused when entering the node upgrade phase
+1. Consent the node upgrade to proceed by removing the node-upgrade-pause annotation on the node
+1. The entire upgrade should finish successfully
+
+#### Multi-node cluster upgrade
+
+1. Prepare a three-node cluster with the master-head version
+1. Configure the `upgrade-config` setting:
+   ```json
+   {
+     "nodeUpgradeOption": {
+       "mode": "manual",
+       "pauseNodes": [
+         "node-2",
+         "node-3"
+       ]
+     }
+   }
+   ```
+1. Trigger an upgrade using the same master-head version ISO image
+1. The upgrade should be paused for `node-2` and `node-3` (the order is not guaranteed) before they enter the pre-draining state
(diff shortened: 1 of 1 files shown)


enhancements/20251105-pausable-node-upgrades.md
```diff
@@ -0,0 +1,158 @@
...
+
+### API changes
+
+There are no actual API changes involved.
+
+## Design
+
+### Implementation Overview
+
+The implementation includes two major parts: a set of new setting fields and the node annotation manipulation. Users configure the `nodeUpgradeOption` fields in the `upgrade-config` setting to specify their node upgrade preferences. The upgrade controller then annotates the nodes marked for pause according to the setting when the upgrade is initiated.
```
w13915984028: Do we allow user to manually annotate the node object to get this ability after the upgrade is triggered?

enhancements/20251105-pausable-node-upgrades.md
```diff
@@ -0,0 +1,158 @@
...
+
+During the node upgrade phase, the secret controller reconciles the machine-plan secrets. When the time comes (when the machine-plan secret for a node is annotated as eligible to upgrade), it checks whether the node has the annotation before creating the pre-drain job. When the annotation is removed from the node, the secret controller is triggered again and reconciles the machine-plan secrets to create the corresponding pre-drain job, thus continuing the original node upgrade flow.
```
w13915984028: Does Harvester or Rancher have related time-out, if this step does not move ahead?

enhancements/20251105-pausable-node-upgrades.md
```diff
@@ -0,0 +1,158 @@
...
+## Motivation
+
+There are different types of user workloads. Some of these workloads are not eligible for live migration. In such cases, they will inevitably be shut down during the Harvester Upgrade process, particularly during the node upgrade phase. Such behavior/requirement is less acceptable to users, as they cannot know the order of node upgrades in advance and don't have enough time to shut down the virtual machines that will be affected gracefully. Moreover, many users prefer to handle an upgrade event manually rather than let it execute autonomously. They would like to nudge the workload migration based on node-specific criteria. For instance, moving workloads to a more capable node. All of this stems from the inability to manually intervene in node upgrade operations and the overly simplistic virtual machine eviction mechanism.
+
+Note: In recent Harvester versions, virtual machines that cannot be live-migrated must be turned off before an upgrade can start, resulting in significant downtime for user workloads and services.
+
+### Goals
+
+- Enable per-upgrade opt-in for pausable node upgrades
+- Automatically pause node upgrade before pre-drain job creation
```
w13915984028, line 20: UI needs the review / sub-story, to selectively show the `unpause` menu.

w13915984028 (dismissed):
LGTM, thanks.

enhancements/20251105-pausable-node-upgrades.md
starbops: Yes, the user can skip configuring the `upgrade-config` setting entirely and annotate the node objects accordingly to achieve the same result, even after the upgrade is triggered. The newly introduced setting fields are just a way for the user to configure everything in a single place and provide a more consistent user experience as a whole. Whether or not the **mode** being `manual`, the secret controller only cares about the existence of the node annotation, and the upgrade controller always tries to delete the annotation from the node objects when cleaning things up.

enhancements/20251105-pausable-node-upgrades.md
starbops: AFAIK, there are no timeouts, as we sometimes see people stuck in the pre-draining state for extended periods when issues arise, even for days. Are you suggesting that we should implement a timeout here?

w13915984028: Timeout is not required for now.

We can add this to the document, controller waits without timeout. Meanwhile, some Harvester webhooks might disallow user's operation inbetween the upgrade.

enhancements/20251105-pausable-node-upgrades.md
w13915984028: It could be a bit arguable for testing. Which object is the source of truth.

A possbile candidate is to copy the user setting of `manual/auto` into `upgrade` CR, it is the feature gate, and for the node, annotation can be dynamically added/removed.

I am open to both solutions.

starbops: I don't see actual benefits from replicating the setting to the Upgrade CR. It can be a bit daunting for users to create a customized Upgrade CR unless they're happy with using `kubectl` to initiate the upgrade. Otherwise, we will need to consider how many nodes are in the cluster if we want to make the upgrade dialog support configuring the pause node list. Not to mention the user will have to do this again for the next upgrade. Putting them on the settings page can solve this once and for all, IMO. We also did similar things (put the options into a Setting CR) to support the parallel image preloading in #6181.

Would you mind elaborate on the testing point? Thank you.

w13915984028: The consideration is:

HEP proposal:
(1) User can set the `upgrade global setting` at any time
(2) After upgrade is triggered, user can still set annotation on node & upgrade global setting at any time, but only the annotation is respected.

user might think that change the global to `auto` can resume the current upgrade automatically, or other cases

The suggested solution:
(1) User can set the upgrade global setting at any time
(2) When upgrade is `clicked`, the `paused` related setting is copied to `upgrade` object automatically, inbetween then upgrade:
(3.1) The copied `manual/auto` is respected, if user changes on global setting (`manual/auto`, node list), `upgrade` will adapts it on the next time upgrade
(3.2) If the copied value is `manual`, then annotations on object can still be updated inbetween the upgrade, annotations are respected

For validation/document, new solution will be a bit more clear.

starbops: In the original proposed design, the pause setting won't be propagated to node annotations if no Upgrade CRs are created. In practice, the user can safely change the pause upgrade setting without causing any actual behavior changes in the cluster. Even after the upgrade has been initiated, the same is also true since the setting is only referenced at the beginning of an upgrade. Moreover, the node annotations are unique to each Upgrade CR, so the upgrade process cannot be altered after an upgrade has been initiated by updating the setting.

I feel you disagree with the design that included node objects, which is a valid point. I started to notice that it might not be a good idea to ask the user to remove the node annotations directly to unpause the node upgrade process. Making the Upgrade CR the single point of control may be better.

starbops: @w13915984028 I've updated the proposal as per our discussion above. PTAL, many thanks!

w13915984028 (approved):
LGTM, thanks.

enhancements/20251105-pausable-node-upgrades.md
```diff
@@ -0,0 +1,144 @@
...
+
+There are no actual API changes involved.
+
+## Design
+
+### Implementation Overview
+
+Users configure the `nodeUpgradeOption` fields in the `upgrade-config` setting to specify their node upgrade preferences. When the upgrade is created, the mutator webhook references the setting and replicates its value as an annotation on the creating Upgrade custom resource. The annotation acts as the single source of truth of the entire pausable node upgrade function.
+
+During the node upgrade phase, the secret controller reconciles the machine-plan secrets. When the time comes (when the machine-plan secret for a node is annotated as eligible to upgrade), it loads up the annotation and checks whether the node is marked as pausable before creating the pre-drain job. When the annotation is updated to unpause the node, the corresponding machine-plan secret is put on the workqueue so the secret controller can reconcile it again to create the pre-drain job for the unpaused node, thus continuing the original node upgrade flow.
```
brandboat, line 90: So if a user (i.e. UI) wants to unpause a node, they need to remove that node from the annotation in the Upgrade CR, right?
e.g. `harvester.hci.io/xxxAnno: "node1,node2"` → `harvester.hci.io/xxxAnno: "node2"`
This means node1 is now unpaused and ready to proceed?

brandboat (commented):
LGTM, just one question, thanks!

enhancements/20251105-pausable-node-upgrades.md
starbops: Yes, the idea is correct. My current implementation is adding per-node annotations while the Upgrade CR is created by the mutator. The format looks like:

```
harvesterhci.io/node-0: pause
harvesterhci.io/node-1: pause
```

By changing them to `unpause` will unblock the node-upgrade for respective nodes.

brandboat: Thanks for the explanation!

brandboat (approved):
LGTM, thanks!

Vicente-Cheng (approved):
thanks

enhancements/20251105-pausable-node-upgrades.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/8980
+
+## Motivation
+
+There are different types of user workloads. Some of these workloads are not eligible for live migration. In such cases, they will inevitably be shut down during the Harvester Upgrade process, particularly during the node upgrade phase. Such behavior/requirement is less acceptable to users, as they cannot know the order of node upgrades in advance and don't have enough time to shut down the virtual machines that will be affected gracefully. Moreover, many users prefer to handle an upgrade event manually rather than let it execute autonomously. They would like to nudge the workload migration based on node-specific criteria. For instance, moving workloads to a more capable node. All of this stems from the inability to manually intervene in node upgrade operations and the overly simplistic virtual machine eviction mechanism.
+
+Note: In recent Harvester versions, virtual machines that cannot be live-migrated must be turned off before an upgrade can start, resulting in significant downtime for user workloads and services.
```
ihcsim, line 15: I guess I am just not sure what other options are available, besides shutting down these virtual machines. Not like user can add a new migration-compatible node at this point. However, I can see this gives the user a chance to make a conscious decision to incur a workload outage, and that they will remember to bring the workload back online post node upgrade.

enhancements/20251105-pausable-node-upgrades.md
```diff
@@ -0,0 +1,144 @@
...
+
+#### Story 1: Manual Virtual Machine Migration During Pause
+
+As a cluster administrator, I want to manually migrate virtual machines while an upgrade for a node is paused so that I can move workloads to more capable nodes based on my knowledge and preference.
+
+#### Story 2: Last-Minute VM Shutdown
+
+As a cluster administrator with virtual machines requiring graceful shutdown, I want to shut down virtual machines right before node pre-drain so that I minimize downtime (virtual machines don't need to be off during entire upgrade).
+
+### User Experience In Detail
```
ihcsim, line 43: Per our slack discussion, I feel like the flow will be more intuitive if we make the UI pause/unpause button the main UX entrypoint. As I see it, the limitation of this setting mechanism doesn't justifiy its complexity.

From a UX perspective, in the current implementation, user has to configure the setting before upgrade starts. Once upgrade started, this setting has no effects. User will also have to revisit this setting before every upgrade because the workloads on the nodes will likely be different.

With the UI pause/unpause button, we give user the on-the-fly control to manage the nodes.

enhancements/20251105-pausable-node-upgrades.md
```diff
@@ -0,0 +1,144 @@
...
+
+### API changes
+
+There are no actual API changes involved.
+
+## Design
+
+### Implementation Overview
+
+Users configure the `nodeUpgradeOption` fields in the `upgrade-config` setting to specify their node upgrade preferences. When the upgrade is created, the mutator webhook references the setting and replicates its value as an annotation on the creating Upgrade custom resource. The annotation acts as the single source of truth of the entire pausable node upgrade function.
```
ihcsim, line 88: Per slack discussion, storing the annotation in the upgrade resource means every time a user interacts with the pause/unpause button, a reconciliation loop is triggered in the upgrade controller. Once the UI is ready, we should test to see what happens when a user clicks through a series of pause/unpause/pause node sequence.

enhancements/20251105-pausable-node-upgrades.md
```diff
@@ -0,0 +1,144 @@
...
+
+During the node upgrade phase, the secret controller reconciles the machine-plan secrets. When the time comes (when the machine-plan secret for a node is annotated as eligible to upgrade), it loads up the annotation and checks whether the node is marked as pausable before creating the pre-drain job. When the annotation is updated to unpause the node, the corresponding machine-plan secret is put on the workqueue so the secret controller can reconcile it again to create the pre-drain job for the unpaused node, thus continuing the original node upgrade flow.
+
+### Test plan
+
+Due to the Harvester Upgrade implementation, single-node cluster upgrades are executed differently from other cluster configurations; it's required to test at least both.
+
+The following assumes the master-head version already includes the proposed enhancement.
+
+#### Single-node cluster upgrade
```
ihcsim, line 98: Do we need to support this? Again realistically, what else can a user do besides shutting down the problematic virtual machines?

ihcsim (commented):
Thanks for the HEP. Sorry, I completely missed this one before looking at the implementation PR.

enhancements/20251105-pausable-node-upgrades.md
starbops: Users can do whatever they want to the VMs running on the to-be-upgraded node, such as backing up important data or shutting them down gracefully.

enhancements/20251105-pausable-node-upgrades.md
starbops: We don't show a pause button for each node during the upgrade period. We can, however, display a list of nodes and let the user toggle which node to pause before the upgrade starts. It's totally feasible and plays nicely with the setting path (direct toggle takes priority over the setting).

We don't show pause buttons because the pause feature only takes effect when the node hasn't been upgraded, i.e., before pre-draining. Once the node enters the pre-draining state, the pause is meaningless. But again, it's doable to show the pause buttons to the UI during the upgrade period. Under the hood, it just updates the pause map annotation to the Upgrade object. The backend already supports the operation.

enhancements/20251105-pausable-node-upgrades.md
starbops: Yes, it provides a finer-grained control over the upgrade procedure, allowing the user to perform maintenance on a specific node and virtual machine before an (potential) outage happens to the user workload.

enhancements/20251105-pausable-node-upgrades.md
ihcsim: When I initially read https://github.com/harvester/harvester/issues/9493#issuecomment-3545917183, I thought we are going for something like this, where user can control everything from the upgrade dialog:

image

I think this will be more intuitive than using a setting.

starbops: Sorry for the misleading screenshot I posted in the GH issue. That's only for demonstrating where the unpause buttons can be. I initially did not put the pause buttons there because I'm afraid it could confuse users, leading them to think they can **pause** a node upgrade immediately when they click the pause button, which is not true.

starbops: To some extent, I think your suggestion is correct. In the end, we already stated that the pause map annotation is user-mutable, so why not provide them with an easier way to do the same in the UI? It's just that we need to polish the user experience and avoid confusion.

ihcsim (approved):

PR #9447, opened 2025-11-10
improve validation for vcluster addon 

ibrokethecloud:
#### Problem:

PR https://github.com/harvester/experimental-addons/pull/39 bumps vcluster addon to use vcluster v0.30.0

Due to introduction of values schema validation and a new config type within vcluster, we can no longer pass randon values such as `hostname`, `rancherVersion` and `bootstrapPassword` via the values content directly.

These values need to be embedded in a `global` field in values.yaml.

As a result rancher-vcluster addon validation fails since there is no `hostname` field anymore in the valuesContent but rather `global.hostname`

#### Solution:

The PR introduces a minor change in how we parse valuesContent and evaluate hostname so we can continue to support both vcluster v0.19.0 and vcluster v0.30.0

#### Related Issue(s):

#### Test plan:

To test:
* Install the new rancher-vcluster addon once the PR https://github.com/harvester/experimental-addons/pull/39 is merged
* Edit the rancher-vcluster addon as yaml, and change the following fields
  * global.hostname
  * global.bootstrapPassword
* Enable the addon
* Addon should be successfully deployed and user able to login to Rancher

### DIFF
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -3,7 +3,7 @@ package addon
 import (
 	"fmt"
 
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/addon/validator.go
```diff
@@ -117,22 +117,32 @@ func (v *addonValidator) validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon
...
-
 	// valuesContent contains a yaml string
 	if err := yaml.Unmarshal([]byte(newAddon.Spec.ValuesContent), addonContent); err != nil {
 		return werror.NewInternalError(fmt.Sprintf("unable to parse contentValues: %v for %s addon", err, vClusterAddonName))
 	}
 
+	// currently we only support v0.19.0 and v0.30.0 of vcluster
+	// the parsing is designed to handle only these two versions for now
+	var hostname string
+	if newAddon.Spec.Version == "v0.19.0" {
```
w13915984028: it will be nice to define v0.19.0 as a local const

w13915984028 (dismissed):
LGTM, thanks.

BTW, CI is complaining the test code

w13915984028 (approved):
LGTM, thanks.

Vicente-Cheng (approved):
thanks for the enhancement

tserong (approved):

ibrokethecloud:
@mergify backport v1.7

Issue #9141, opened 2025-09-15
[ENHANCEMENT] Check storage-network is fully stopped on LH pods before re-enable storage-network


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Such log was observed when re-enable storagenetwork
>2025-09-11T14:49:43.737064823Z 2025-09-11T14:49:43Z [debug] pod's network status: {Name:harvester-system/storagenetwork-wrnkd Interface:lhnet1 IPs:[172.19.0.9] Mac:6a:fd:af:cd:64:b9 Mtu:0 Default:false DNS:{Nameservers:[] Domain: Search:[] Options:[]} DeviceInfo:<nil> Gateway:[]}
2025-09-11T14:49:43.737071343Z 2025-09-11T14:49:43Z [verbose] result of garbage collecting pods: failed to get network-attachment-definition for iface harvester-system/storagenetwork-wrnkd: networkattachmentdefinition.k8s.cni.cncf.io "storagenetwork-wrnkd" not found

### Describe the solution you'd like

Besides following check, Harvester webhook might check all LH pods, if any has storage-network related annotations/labels, it means the previous storage-networks `disable` action was not finished yet, new `enable` action should be blocked to wait until it is fully done.

https://github.com/harvester/harvester/blob/016e070d64c22c1ed7f7d35f621f0cdb89c775e1/pkg/webhook/resources/setting/validator.go#L1394-L1416

martindekov:
As per suggestion by @WebberHuang1118 we can add to the issue rejection of storage network changes during upgrade as well WDYT @w13915984028 ?

WebberHuang1118:
> it means the previous storage-networks disable action was not finished yet, new enable action should be blocked to wait until it is fully done.

If I understand correctly, the storage network can be updated directly without going through `enable → disable → enable`. In this case, blocking storage network changes in the webhook may not be suitable. We might need to handle this in the setting controller—in other words, wait until Longhorn syncs the latest storage network setting and then remove the outdated NAD. @martindekov @Vicente-Cheng @rrajendran17 @w13915984028

w13915984028:
by default (say newly installed), LH pods do not have second network related annotations/labels, webhook just does noting; but if user clicks `enable` and the pods already have second network annotation, it might be kind of issues, if we can check further about the storagenetwork related nad naming, it can be even secure.

rrajendran17:
[...]

PR #9453, opened 2025-11-10
Ensure storage network configured from clean state

martindekov:
Ensuring storage network is configured from clean state by adding the following:
* Validation on webhook level that no pods with the longhorn storage network (lhnet1) exist when network is enabled for the first time
* On NAD controller remove all IP pools just before NAD is enabled, so we make sure that whereabouts will create only 1 new IP pool

Through above we make sure that each enablement of storage network starts from clean state. There are no pods assigned IPs from previous network enablement and no more than 1 IP pool, the current one, exists preventing ip pool exhaustion.

#### Problem:
When we configure storage network multiple times over time - enable -> disable -> enable -> disable if we use different ranges, whereabouts does not do cleanup of its IP pools. We allow only 1 storage network configured meaning 1 IP pool to lock IPs for the network from whereabouts.

This is fragile as we might hit a bug where one of the old and no longer used IP pools for whatever reason cannot recollect IPs and the allocation stays.

#### Solution:
The change makes sure that each time we enable a storage network:
* No pods containing IPs from previous network enablement exist
* Only one (the current) IP pool allocating IPs exist at the time

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9141

#### Test plan:
Unit tests:
- [x] Webhook
- [x] Controller

Below are positive E2E Tests for:
* Controller maintains only 1 IP Pool when storage network is configured - no leftover storage pools
* Webhook rejects enabling of storage network in case of bad system state

### Test 1 - controller makes sure it works on new IP pool each time

Tested the following:
* Created management network and storage network
* Storage network configured with range: `192.168.50.0/25` - validate clean slate works fine
* Disabling the storage network with range `192.168.50.0/25`
* Enabling new storage network with new range  `192.168.50.0/24`
* Validating that both old ip pool (`192.168.50.0/25`) and new ip pool (`192.168.50.0/24`) storage networks are present but the new one is used while old one is empty
* Disabled the new (`192.168.50.0/24`) and enabled the old (`192.168.50.0/25`) storage networks - through timestamp on the IP pool validated that controller deleted the IP pool with range `192.168.50.0/25` and re-created it. Now ip pool `192.168.50.0/24` is empty while `192.168.50.0/25` was recreated and used.
* Validate storage network nad is there and with valid range

<details>
<summary>On 2 node harvester cluster configuring migration network first</summary>
<br>

```
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io -n kube-system 
NAME               AGE
192.168.100.0-24   5m24s
192.168.50.0-24    4m8s
192.168.50.0-25    4m43s
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-14T14:10:11Z"
  generation: 3
  name: 192.168.50.0-24
  namespace: kube-system
  resourceVersion: "18374"
  uid: dbec4df3-819e-459e-872a-f2a9e8feda81
spec:
  allocations:
    "1":
      id: 29873c53508bf5e42b41fb63ae36a37fb1a1f3349695460926a1aae2f9aed6dc
      ifname: lhnet1
      podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
    "2":
      id: ed70d69a073d1762267a2ace95848488ec3e9740c831f460bf9ff8918bbe6896
      ifname: lhnet1
      podref: longhorn-system/instance-manager-74ec8021feeb76cf6e8cce35c5a81c01
  range: 192.168.50.0/24
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io -n kube-system 
NAME               AGE
192.168.100.0-24   5m56s
192.168.50.0-24    4m40s
192.168.50.0-25    5s
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-25 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-14T14:14:46Z"
  generation: 3
  name: 192.168.50.0-25
  namespace: kube-system
  resourceVersion: "21631"
  uid: d69657d0-0b5b-4897-8462-136ef316ae93
spec:
  allocations:
    "1":
      id: f4547d1bba22eb485fd3f01db8fa67b3e5f3643247826ec8dc6b59b24855156f
      ifname: lhnet1
      podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
    "2":
      id: 78162ca463ee47bc59a9cd6c44c4aef07d2fd6dd59604ef5eebcb5064bf555ec
      ifname: lhnet1
      podref: longhorn-system/instance-manager-74ec8021feeb76cf6e8cce35c5a81c01
  range: 192.168.50.0/25
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-14T14:10:11Z"
  generation: 5
  name: 192.168.50.0-24
  namespace: kube-system
  resourceVersion: "21549"
  uid: dbec4df3-819e-459e-872a-f2a9e8feda81
spec:
  allocations: {}
  range: 192.168.50.0/24
harv-1:/home/rancher # kubectl get net-attach-def -n harvester-system
NAME                         AGE
storagenetwork-tk4k5         5m31s
vm-migration-network-xdv5l   12m
harv-1:/home/rancher # kubectl get net-attach-def/storagenetwork-tk4k5 -o yaml -n harvester-system
apiVersion: k8s.cni.cncf.io/v1
metadata:
  annotations:
    storage-network.settings.harvesterhci.io: "true"
  creationTimestamp: "2025-11-14T14:14:45Z"
  finalizers:
  - wrangler.cattle.io/harvester-network-manager-nad-controller
  generateName: storagenetwork-
  generation: 1
  labels:
    network.harvesterhci.io/clusternetwork: storage
    network.harvesterhci.io/ready: "true"
    network.harvesterhci.io/type: L2VlanNetwork
    network.harvesterhci.io/vlan-id: "50"
    storage-network.settings.harvesterhci.io/hash: 11bf1b0fff1696149ad393c2ce37e943253dfedc
  name: storagenetwork-tk4k5
  namespace: harvester-system
  resourceVersion: "21522"
  uid: bb4988d8-6dfd-471e-bb01-801cc4016922
spec:
  config: '{"cniVersion":"0.3.1","type":"bridge","bridge":"storage-br","promiscMode":true,"vlan":50,"ipam":{"type":"whereabouts","range":"192.168.50.0/25"}}'
```
</details>

### Test 2 - validation webhook suggest cleanup in case leftover pods in bad state

Following experiment above we end up with configured storage network.
<details>
<summary>Turn off one of the nodes and disable storage network to make pods in bad Terminating state</summary>
<br>

```
harv-1:/home/rancher # kubectl get settings -n harvester-system | grep network
storage-network                               
vm-migration-network                          {"clusterNetwork":"mgmt","vlan":50,"range":"192.168.100.0/24","exclude":[]}
harv-1:/home/rancher # kubectl get net-attach-def -n harvester-system
NAME                         AGE
vm-migration-network-qfngv   149m
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io -n kube-system
NAME                AGE
192.168.100.0-24    147m
192.168.50.128-25   19m
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.128-25 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-12T14:05:10Z"
  generation: 4
  name: 192.168.50.128-25
  namespace: kube-system
  resourceVersion: "114612"
  uid: 714686e8-b7ba-4c9f-9704-795f38f08ca3
spec:
  allocations:
    "2":
      id: c5cdd9922594c4b1d37677f4d77d821ec1295d057952bccd3fe5f9e8c0de92cb
      ifname: lhnet1
      podref: longhorn-system/instance-manager-f22861418d9476c0a8ffc683e29d279c
  range: 192.168.50.128/25
harv-1:/home/rancher # kubectl get pods -n longhorn-system | grep Terminating
instance-manager-f22861418d9476c0a8ffc683e29d279c   1/1     Terminating   0              20m
```
</details>

<details>
<summary>Attempting to enable storage network from bad state results in request being rejected by the webhook</summary>
<br>
image
</details>

<details>
<summary>Deleting the resource enables us to configure the storage network again</summary>
<br>

```
harv-1:/home/rancher # kubectl delete pod/instance-manager-f22861418d9476c0a8ffc683e29d279c -n longhorn-system --grace-period 0 --force
Warning: Immediate deletion does not wait for confirmation that the running resource has been terminated. The resource may continue to run on the cluster indefinitely.
pod "instance-manager-f22861418d9476c0a8ffc683e29d279c" force deleted from longhorn-system namespace
harv-1:/home/rancher # kubectl get settings -n harvester-system | grep network
storage-network                               {"vlan":50,"clusterNetwork":"storage","range":"192.168.50.128/25"}
vm-migration-network                          {"clusterNetwork":"mgmt","vlan":50,"range":"192.168.100.0/24","exclude":[]}
harv-1:/home/rancher # kubectl get net-attach-def -n harvester-system
NAME                         AGE
storagenetwork-4rldc         47s
vm-migration-network-qfngv   154m
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io -n kube-system
NAME                AGE
192.168.100.0-24    153m
192.168.50.128-25   38s
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.128-25 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-12T14:29:40Z"
  generation: 2
  name: 192.168.50.128-25
  namespace: kube-system
  resourceVersion: "118016"
  uid: 4051a62c-b5e0-4fe4-8d39-928fd0e2702e
spec:
  allocations:
    "1":
      id: 6945c9d8aaf1f2841a4d19ffb3b0a3fd40578f89a9ca985583dbcde7ede9fea3
      ifname: lhnet1
      podref: longhorn-system/instance-manager-357c5150c97be5e63e0bb6c64ad2f6a2
  range: 192.168.50.128/25
harv-1:/home/rancher # kubectl get pod/instance-manager-357c5150c97be5e63e0bb6c64ad2f6a2 -o yaml -n longhorn-system | grep 192.168.50.
              "192.168.50.129"
```
</details>

### Test 3 - validating we already reject same ranges

<details>
<summary>Attempting to create storage network with same range as migration network is rejected</summary>
<br>
image

N/A - in the issue


pkg/webhook/indexeres/indexer.go
```diff
@@ -45,6 +45,7 @@ func RegisterIndexers(clients *clients.Clients) {
 
 	podCache := clients.CoreFactory.Core().V1().Pod().Cache()
 	podCache.AddIndexer(indexeresutil.PodByVMNameIndex, indexeresutil.PodByVMName)
+	podCache.AddIndexer(indexeresutil.PodNetworkInterfaceIndex, indexeresutil.PodByNetworkInterface)
```
w13915984028: SN is rarely enabled/disabled in production, it looks not necessary to build an index for it, iterate the PODs from LH namespace is ok

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -813,3 +819,19 @@ func (h *Handler) updateLonghornStorageNetwork(storageNetwork string) error {
...
 	return nil
 }
+
+func (h *Handler) cleanUpOldIPPool() error {
+	ipPools, err := h.whereaboutsCNIIPPoolCache.List(util.KubeSystemNamespace, labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	for idx := range ipPools {
```
w13915984028: the `migration-network` seems also using similar NAD and IPPools, does IPPool have annotation/lable to identify it is from SN? if not, add one from the SN controller creation operation, and here uses it to match

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -813,3 +819,19 @@ func (h *Handler) updateLonghornStorageNetwork(storageNetwork string) error {
...
+		err := h.whereaboutsCNIIPPoolClient.Delete(util.KubeSystemNamespace, ipPools[idx].Name, &metav1.DeleteOptions{})
+		if err != nil {
```
w13915984028: skip `IsNotFound` error

w13915984028 (commented):
Thanks. OK to do more local tests.

pkg/controller/master/storagenetwork/storage_network.go
martindekov: Done :heavy_check_mark:

pkg/webhook/indexeres/indexer.go
martindekov: Replaced with list and filter :heavy_check_mark:

pkg/controller/master/storagenetwork/storage_network.go
martindekov: That one was interesting, so yes the migration NAD has annotation:
```
apiVersion: k8s.cni.cncf.io/v1
metadata:
  annotations:
    vm-migration-network.settings.harvesterhci.io: "true"
```
So now I basically list all nads and IP pools. FRom the nads belonging to migraiton network I construct IP pool name and before deleting I make sure to skip calling delete when the ip pool belongs to the migration network

Very good catch Jian, I didn't know we use the same approach for the migration network as well! Leaving open discussion if it makes sense to resolve it yourself

pkg/webhook/resources/setting/validator.go
```diff
@@ -1451,6 +1460,66 @@ func (v *settingValidator) checkVlanStatusReady(config *networkutil.Config) erro
...
+	podsFromStorageNetwork := filterPodsByStorageNetwork(pods)
+	if len(podsFromStorageNetwork) > 0 {
+		podNames := extractPodNamesFromObj(podsFromStorageNetwork)
+		return fmt.Errorf("the following pods: %s still contain IPs from previous storage network; clean them up before re-enabling", strings.Join(podNames, ", "))
+	}
+
+	return nil
+}
+
+func filterPodsByStorageNetwork(pods []*corev1.Pod) []*corev1.Pod {
```
w13915984028: seems return the pod name list `[]string` directly is better

pkg/webhook/resources/setting/validator.go
```diff
@@ -1451,6 +1460,66 @@ func (v *settingValidator) checkVlanStatusReady(config *networkutil.Config) erro
...
+	}
+	return podsFromStorageNetwork
+}
+
+func isPartOfStorageNetwork(networkInterface string) bool {
+	var networkStatuses []nettypes.NetworkStatus
+	if err := json.Unmarshal([]byte(networkInterface), &networkStatuses); err != nil {
+		return false
+	}
+	for _, status := range networkStatuses {
```
w13915984028, line 1510: check len of `networkStatuses` first, if it <=1, it can only be pod network and skip it

that's a basic assumption of k8s pod, it always has an interface to default pod network

w13915984028 (commented):
some optimization suggestions

pkg/webhook/resources/setting/validator.go
martindekov: Good suggestion, addressed, reducing lines of code

pkg/webhook/resources/setting/validator.go
martindekov: Yep, optimized, neat

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -813,3 +820,56 @@ func (h *Handler) updateLonghornStorageNetwork(storageNetwork string) error {
 	}
 	return nil
 }
+
+func (h *Handler) cleanUpOldSNIPPool() error {
+	ipPools, err := h.whereaboutsCNIIPPoolCache.List(util.KubeSystemNamespace, labels.Everything())
```
w13915984028: if len(ipPools) == 0, return directly

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -813,3 +820,56 @@ func (h *Handler) updateLonghornStorageNetwork(storageNetwork string) error {
...
+	}
+
+	migrationPoolNames := h.getMigrationPoolNames(nads.Items)
+	for _, ipPool := range ipPools {
+		if _, keep := migrationPoolNames[ipPool.Name]; keep {
+			continue
+		}
+		err := h.whereaboutsCNIIPPoolClient.Delete(util.KubeSystemNamespace, ipPool.Name, &metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			return err
```
w13915984028: we might need a `enqueueAfter` action after successfully putting the `deletion` on the existing IPPool, and wait unti l it is gone

if the new creation operation works upon the existing being deleted object, the resut is unexpected

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -813,3 +820,56 @@ func (h *Handler) updateLonghornStorageNetwork(storageNetwork string) error {
...
+func (h *Handler) getMigrationPoolNames(nads []nadv1.NetworkAttachmentDefinition) map[string]bool {
+	migrationPools := make(map[string]bool)
+	for _, nad := range nads {
+		annotations := nad.GetAnnotations()
+
+		if val, ok := annotations[kubevirt.VMMigrationNetworkPrefix]; !ok || val != "true" {
+			continue
+		}
+		var config network.BridgeConfig
+		if err := json.Unmarshal([]byte(nad.Spec.Config), &config); err != nil {
```
w13915984028: log a message to indicate the error, as we skip the error

w13915984028 (commented):
some questions, 

and also I forgot the details, when `storage-network` and `migration-network` are sharing the same sub-net, are there two IPPools or one? thanks.

pkg/controller/master/storagenetwork/storage_network.go
martindekov: You are right, I will requeue the setting after 5 seconds before returning the error here

martindekov:
>and also I forgot the details, when storage-network and migration-network are sharing the same sub-net, are there two IPPools or one? thanks.

Overlapping ranges are blocked by a webhook:
image

pkg/controller/master/storagenetwork/storage_network.go
martindekov: Good optimization, done :heavy_check_mark:

pkg/controller/master/storagenetwork/storage_network.go
martindekov: Requeue after 2 secs :heavy_check_mark:

pkg/controller/master/storagenetwork/storage_network.go
martindekov: logged error :heavy_check_mark:

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -813,3 +819,36 @@ func (h *Handler) updateLonghornStorageNetwork(storageNetwork string) error {
 	}
 	return nil
 }
+
+func (h *Handler) cleanUpOldSNIPPool(setting *harvesterv1.Setting) error {
+	ipPools, err := h.whereaboutsCNIIPPoolCache.List(util.KubeSystemNamespace, labels.Everything())
```
w13915984028: When we work upon a specific IPPool, the `list` could be replaced by `get` via the target name

Also add a comment that the `webhook` ensures the `migration-network` and `storage-network` have different IPPool name even when they share but split a same subnet

w13915984028 (commented):
Last optimization, thanks.

pkg/controller/master/storagenetwork/storage_network.go
martindekov: I can actually call directly delete and in case not found return nil, I think this is even better optimization reducing everything to 1 API Call

martindekov: Calling delete directly :heavy_check_mark:

w13915984028 (dismissed):
LGTM, thanks.

.gitignore
```diff
@@ -13,3 +13,4 @@ package/harvester
 package/harvester-webhook
 package/upgrade/upgrade-helper
 package/upgrade/addons
+.vscode
```
Vicente-Cheng: redundant
I believe we can add it, but it would not be related to this PR.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -813,3 +819,28 @@ func (h *Handler) updateLonghornStorageNetwork(storageNetwork string) error {
...
+	ipPoolName := h.cidrToIPPoolName(config.Range)
+
+	logrus.Warnf("Deleting old IPPool: %s", ipPoolName)
+	err := h.whereaboutsCNIIPPoolClient.Delete(namespace, ipPoolName, &metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if err != nil && apierrors.IsNotFound(err) {
+		return nil
+	}
```
WebberHuang1118, line 838: If we clean up the `ippool`, don’t we also need to clean up the `overlappingrangeipreservation` resources?

pkg/webhook/resources/setting/validator.go
```diff
@@ -1451,6 +1460,60 @@ func (v *settingValidator) checkVlanStatusReady(config *networkutil.Config) erro
...
+		return err
+	}
+
+	podNames := filterPodsByStorageNetwork(pods)
+	if len(podNames) > 0 {
+		return fmt.Errorf("the following pods: %s still contain IPs from previous storage network; clean them up before re-enabling", strings.Join(podNames, ", "))
+	}
+
+	return nil
+}
```
WebberHuang1118, line 1483: Suggest making the syntax a bit clearer:


```suggestion
func (v *settingValidator) checkLeftoverPodIPs(oldSetting, newSetting *v1beta1.Setting) error {
	if oldSetting.Value == newSetting.Value {
		return nil
	}

	// Only check for leftover pod IPs when enabling the setting (transitioning from empty to non-empty)
	if !isSettingBeingEnabled(oldSetting.Value, newSetting.Value) {
		return nil
	}

	pods, err := v.podCache.List(util.LonghornSystemNamespaceName, labels.Everything())
	if err != nil {
		return err
	}

	podNames := filterPodsByStorageNetwork(pods)
	if len(podNames) > 0 {
		return fmt.Errorf("the following pods: %s still contain IPs from previous storage network; clean them up before re-enabling", strings.Join(podNames, ", "))
	}

	return nil
}

// isSettingBeingEnabled checks if a setting is transitioning from unset (empty) to set (non-empty).
// This is useful for validating conditions that should only apply when initially enabling a feature.
func isSettingBeingEnabled(oldValue, newValue string) bool {
	return oldValue == "" && newValue != ""
}
```

pkg/webhook/resources/setting/validator.go
```diff
@@ -1451,6 +1460,60 @@ func (v *settingValidator) checkVlanStatusReady(config *networkutil.Config) erro
...
+
+func filterPodsByStorageNetwork(pods []*corev1.Pod) []string {
+	podsFromStorageNetwork := make([]string, 0)
+	for _, pod := range pods {
+		networkInterface := pod.Annotations[string(lhmtypes.CNIAnnotationNetworkStatus)]
+		if networkInterface != "" && isPartOfStorageNetwork(networkInterface) {
+			podsFromStorageNetwork = append(podsFromStorageNetwork, pod.Name)
+			continue
```
WebberHuang1118: The original `continue` isn’t needed, or we could rewrite it as:

```suggestion
		networkInterface := pod.Annotations[string(lhmtypes.CNIAnnotationNetworkStatus)]
		if networkInterface == "" || !isPartOfStorageNetwork(networkInterface) {
			continue
		}
		podsFromStorageNetwork = append(podsFromStorageNetwork, pod.Name)
```

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -288,6 +290,10 @@ func (h *Handler) findOrCreateNad(setting *harvesterv1.Setting) (*nadv1.NetworkA
 	}
 
 	if len(nads.Items) == 0 {
+		err := h.cleanUpOldSNIPPool(setting, util.KubeSystemNamespace)
+		if err != nil {
+			return nil, err
+		}
```
WebberHuang1118: Suggest refactoring it to avoid nested `if` statements:

```
func (h *Handler) cleanUpAndCreateNad(setting *harvesterv1.Setting) (*nadv1.NetworkAttachmentDefinition, error) {
	err := h.cleanUpOldSNIPPool(setting, util.KubeSystemNamespace)
	if err != nil {
		return nil, err
	}
	return h.createNad(setting)
}

func (h *Handler) findOrCreateNad(setting *harvesterv1.Setting) (*nadv1.NetworkAttachmentDefinition, error) {
	nads, err := h.networkAttachmentDefinitions.List(util.StorageNetworkNetAttachDefNamespace, metav1.ListOptions{
		LabelSelector: labels.Set{
			util.HashStorageNetworkLabel: h.sha1(setting.Value),
		}.String(),
	})
	if err != nil {
		return nil, err
	}

	if len(nads.Items) == 0 {
		return h.cleanUpAndCreateNad(setting)
	}

	if len(nads.Items) > 1 {
			"num_of_nad":          len(nads.Items),
			"storage_network_nad": nads.Items[0].Name,
		}).Info("storage network: found more than one match nad")
	}

	return &nads.Items[0], nil
}
```

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -813,3 +819,28 @@ func (h *Handler) updateLonghornStorageNetwork(storageNetwork string) error {
...
+
+	logrus.Warnf("Deleting old IPPool: %s", ipPoolName)
+	err := h.whereaboutsCNIIPPoolClient.Delete(namespace, ipPoolName, &metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if err != nil && apierrors.IsNotFound(err) {
+		return nil
+	}
+	return fmt.Errorf("IPPool: %s is being deleted; requeue", ipPoolName)
```
WebberHuang1118, line 842: We return an error here even though the deletion succeeded? If I understand correctly, we should return `nil` in this case.

pkg/controller/master/storagenetwork/storage_network.go
WebberHuang1118, line 838:
```suggestion
	err := h.whereaboutsCNIIPPoolClient.Delete(namespace, ipPoolName, &metav1.DeleteOptions{})
	if apierrors.IsNotFound(err) {
		return nil
	}
	if err != nil {
		return err
	}
```

WebberHuang1118 (changes requested):
Do we really need to modify the Whereabouts CR in [this section of the code](https://github.com/martindekov/harvester/blob/8589556cd00b76779d9223e5b50be47993bd190a/pkg/controller/master/storagenetwork/storage_network.go#L823-L839)? As far as I know, Whereabouts already has a garbage-collection mechanism to clean up leaked IPs.

pkg/controller/master/storagenetwork/storage_network.go
martindekov: if error is nil this will throw panic that's why I checked nil not empty before checking whether error is of type found/notfound:
https://github.com/kubernetes/apimachinery/blob/v0.34.2/pkg/api/errors/errors.go#L528-L537

Also we always have upper level check in other places that err is not nil before checking type:
https://github.com/martindekov/harvester/blob/8589556cd00b76779d9223e5b50be47993bd190a/pkg/controller/master/storagenetwork/storage_network.go#L750

pkg/controller/master/storagenetwork/storage_network.go
martindekov: We force reconcile by the error returned here, until object is not found - this is the condition where we exit

pkg/controller/master/storagenetwork/storage_network.go
martindekov: Yes whereabouts have garbage collection, but it works only in case all resources are up in running.

I could replicate situations where there are leftover reservations in the ippool, when pod (instance manager) is shut down (related node is down) and you do some enabling/disabling of storage network. If I never turn this on, it will be forever stuck in there and the reservation won't be GCd I haven't looked into `overlappingrangeipreservation` objects cc @w13915984028

martindekov:
>Do we really need to modify the Whereabouts CR in [this section of the code](https://github.com/martindekov/harvester/blob/8589556cd00b76779d9223e5b50be47993bd190a/pkg/controller/master/storagenetwork/storage_network.go#L823-L839)? As far as I know, Whereabouts already has a garbage-collection mechanism to clean up leaked IPs.

Replied here: https://github.com/harvester/harvester/pull/9453#discussion_r2549487851

pkg/controller/master/storagenetwork/storage_network.go
martindekov: This is a good thing to keep in mind I just saw there are places in the same file where we have 2 nests as well 1 if and inside another if check for the error, can refactor as part of other change in same codebase

pkg/webhook/resources/setting/validator.go
martindekov: Good suggestion, can change in a follow up, also didn't extract originally as I couldn't find other places where we need this check `isSettingBeingEnabled` so decided to extract when needed

pkg/webhook/resources/setting/validator.go
martindekov: You are right, originally the logic was diffident, but forgot to remove continue good catch

martindekov:
@WebberHuang1118 apart from the overlappingranges we can rewrite the suggestions as part of: https://github.com/harvester/harvester/issues/9501 this implementation is well tested if we want to get this in today, otherwise if we don't I can reference it right away, but testing the code after I made the refactoring will take some time and we won't be able to get this in for today in that case,

cc @Vicente-Cheng

pkg/controller/master/storagenetwork/storage_network.go
martindekov: Rewrote per suggestion :heavy_check_mark:

pkg/webhook/resources/setting/validator.go
martindekov: Done :heavy_check_mark:

pkg/webhook/resources/setting/validator.go
martindekov: Done :heavy_check_mark:

martindekov:
>Do we really need to modify the Whereabouts CR in [this section of the code](https://github.com/martindekov/harvester/blob/8589556cd00b76779d9223e5b50be47993bd190a/pkg/controller/master/storagenetwork/storage_network.go#L823-L839)? As far as I know, Whereabouts already has a garbage-collection mechanism to clean up leaked IPs.

Ran the following experiment cc @WebberHuang1118 

For the overlapping ranges performed the following experiment:
1. 2 nodes 1 worker and 1 management node with configured storage network
<details>
<summary>2. turn off the worker node and disable the storage network forcing the pod on the turned off node to be in Terminating state</summary>
<br>

```
harv-1:/home/rancher # kubectl get pods -n longhorn-system --watch
NAME                                                READY   STATUS        RESTARTS   AGE
csi-attacher-7b996ff85b-gkn2s                       1/1     Running       0          95m
csi-attacher-7b996ff85b-j7xgj                       1/1     Running       0          95m
csi-attacher-7b996ff85b-q4kmj                       1/1     Running       0          95m
csi-provisioner-b5c5f5cbf-crgwn                     1/1     Running       0          95m
csi-provisioner-b5c5f5cbf-rg5th                     1/1     Running       0          95m
csi-provisioner-b5c5f5cbf-xz9nq                     1/1     Running       0          95m
csi-resizer-85fb9c8c9f-dvx7x                        1/1     Running       0          95m
csi-resizer-85fb9c8c9f-slnbn                        1/1     Running       0          95m
csi-resizer-85fb9c8c9f-tjjvd                        1/1     Running       0          95m
csi-snapshotter-867b4445d6-dgpcr                    1/1     Running       0          95m
csi-snapshotter-867b4445d6-qkn7m                    1/1     Running       0          95m
csi-snapshotter-867b4445d6-qv674                    1/1     Running       0          95m
engine-image-ei-3154f3aa-9pfbk                      1/1     Running       0          95m
engine-image-ei-3154f3aa-jrht9                      1/1     Running       0          4m30s
instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4   1/1     Terminating   0          3m10s
instance-manager-74ec8021feeb76cf6e8cce35c5a81c01   1/1     Running       0          5s
longhorn-csi-plugin-2bx5h                           3/3     Running       0          4m30s
longhorn-csi-plugin-s9jms                           3/3     Running       0          95m
longhorn-driver-deployer-785c6548d-qcdjb            1/1     Running       0          96m
longhorn-manager-74bxb                              2/2     Running       0          96m
longhorn-manager-pwrmc                              2/2     Running       0          4m30s
longhorn-ui-677666b49d-64kxt                        1/1     Running       0          95m
longhorn-ui-677666b49d-f6th9                        1/1     Running       0          96m
harv-1:/home/rancher # kubectl get overlappingrangeipreservations -n kube-system
NAME            AGE
192.168.100.1   80m
192.168.100.2   68m
192.168.100.3   4m42s
192.168.50.2    83m
harv-1:/home/rancher # kubectl get overlappingrangeipreservations/192.168.50.2 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: OverlappingRangeIPReservation
metadata:
  creationTimestamp: "2025-11-21T13:51:54Z"
  generation: 1
  name: 192.168.50.2
  namespace: kube-system
  resourceVersion: "16247"
  uid: d93482fc-a588-4ff8-b8a3-e33c45dd142d
spec:
  ifname: lhnet1
  podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-21T15:11:41Z"
  generation: 4
  name: 192.168.50.0-24
  namespace: kube-system
  resourceVersion: "76573"
  uid: e47a776c-b666-4b5a-a9e2-af9fade52995
spec:
  allocations:
    "2":
      id: 5363ccf27a794b2d1d1692939501490b228753387b2e30493c27b2f06a829870
      ifname: lhnet1
      podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
  range: 192.168.50.0/24
```
</details>

<details>
<summary>3. Simply turning on (fixing) the node, makes the whereabouts clear the overlapping ranges from the storage network but NOT clear the ip pool</summary>
<br>

```
harv-1:/home/rancher # kubectl get overlappingrangeipreservations -n kube-system
NAME            AGE
192.168.100.1   86m
192.168.100.2   74m
192.168.100.3   2m41s
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io -n kube-system 
NAME               AGE
192.168.100.0-24   86m
192.168.50.0-24    9m42s
```
</details>
     
4. Hence that's why we clear up the ip pools on next enablement only, and leave the overlappingranges, whereabouts leave the former, while it clears up the latter

pkg/controller/master/storagenetwork/storage_network.go
w13915984028: @WebberHuang1118 @martindekov 

The key point here is:

As the `ippool` will be created with the same name, if the old IPPool is not truely deleted ( note: when it is first deleted, it has a deletionTimestamp, butk8s takes time to finally remove the object from api-server), we can't create the new object before the old object really disappears.  If run `create` upon a same named, but with deletionTimestamp object, the result is un-expected.

There are two solutions:

(1) Actively enqueueAfter, until it can't detect any existing leftover same named object, then create one;

(2) Simply reuse the wrangler framework, return `error` to let reconciller calls it;

This PR adopts solution (2) due to it is simple.

Indeed, this is a general race for any module, if it deletes old object and creates new object in a tight loop, if the object has the same name, then the new object might surprisingly disappeared/the old object is accidentally picked up and reused.

w13915984028 (dismissed):
LGTM, thanks.

pkg/controller/master/storagenetwork/storage_network.go
w13915984028: @martindekov Please add a comment on the code, to avoid confusing, thanks.

WebberHuang1118: @martindekov @w13915984028 

This just improves readability a bit — there’s no change in logic.

WebberHuang1118:
> > Do we really need to modify the Whereabouts CR in [this section of the code](https://github.com/martindekov/harvester/blob/8589556cd00b76779d9223e5b50be47993bd190a/pkg/controller/master/storagenetwork/storage_network.go#L823-L839)? As far as I know, Whereabouts already has a garbage-collection mechanism to clean up leaked IPs.
> 
> Ran the following experiment cc @WebberHuang1118
> 
> For the overlapping ranges performed the following experiment:
> 
> 1. 2 nodes 1 worker and 1 management node with configured storage network
> 
> 2. turn off the worker node and disable the storage network forcing the pod on the turned off node to be in Terminating state
> 
> ```
> harv-1:/home/rancher # kubectl get pods -n longhorn-system --watch
> NAME                                                READY   STATUS        RESTARTS   AGE
> csi-attacher-7b996ff85b-gkn2s                       1/1     Running       0          95m
> csi-attacher-7b996ff85b-j7xgj                       1/1     Running       0          95m
> csi-attacher-7b996ff85b-q4kmj                       1/1     Running       0          95m
> csi-provisioner-b5c5f5cbf-crgwn                     1/1     Running       0          95m
> csi-provisioner-b5c5f5cbf-rg5th                     1/1     Running       0          95m
> csi-provisioner-b5c5f5cbf-xz9nq                     1/1     Running       0          95m
> csi-resizer-85fb9c8c9f-dvx7x                        1/1     Running       0          95m
> csi-resizer-85fb9c8c9f-slnbn                        1/1     Running       0          95m
> csi-resizer-85fb9c8c9f-tjjvd                        1/1     Running       0          95m
> csi-snapshotter-867b4445d6-dgpcr                    1/1     Running       0          95m
> csi-snapshotter-867b4445d6-qkn7m                    1/1     Running       0          95m
> csi-snapshotter-867b4445d6-qv674                    1/1     Running       0          95m
> engine-image-ei-3154f3aa-9pfbk                      1/1     Running       0          95m
> engine-image-ei-3154f3aa-jrht9                      1/1     Running       0          4m30s
> instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4   1/1     Terminating   0          3m10s
> instance-manager-74ec8021feeb76cf6e8cce35c5a81c01   1/1     Running       0          5s
> longhorn-csi-plugin-2bx5h                           3/3     Running       0          4m30s
> longhorn-csi-plugin-s9jms                           3/3     Running       0          95m
> longhorn-driver-deployer-785c6548d-qcdjb            1/1     Running       0          96m
> longhorn-manager-74bxb                              2/2     Running       0          96m
> longhorn-manager-pwrmc                              2/2     Running       0          4m30s
> longhorn-ui-677666b49d-64kxt                        1/1     Running       0          95m
> longhorn-ui-677666b49d-f6th9                        1/1     Running       0          96m
> harv-1:/home/rancher # kubectl get overlappingrangeipreservations -n kube-system
> NAME            AGE
> 192.168.100.1   80m
> 192.168.100.2   68m
> 192.168.100.3   4m42s
> 192.168.50.2    83m
> harv-1:/home/rancher # kubectl get overlappingrangeipreservations/192.168.50.2 -o yaml -n kube-system
> apiVersion: whereabouts.cni.cncf.io/v1alpha1
> kind: OverlappingRangeIPReservation
> metadata:
>   creationTimestamp: "2025-11-21T13:51:54Z"
>   generation: 1
>   name: 192.168.50.2
>   namespace: kube-system
>   resourceVersion: "16247"
>   uid: d93482fc-a588-4ff8-b8a3-e33c45dd142d
> spec:
>   ifname: lhnet1
>   podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
> harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system
> apiVersion: whereabouts.cni.cncf.io/v1alpha1
> kind: IPPool
> metadata:
>   creationTimestamp: "2025-11-21T15:11:41Z"
>   generation: 4
>   name: 192.168.50.0-24
>   namespace: kube-system
>   resourceVersion: "76573"
>   uid: e47a776c-b666-4b5a-a9e2-af9fade52995
> spec:
>   allocations:
>     "2":
>       id: 5363ccf27a794b2d1d1692939501490b228753387b2e30493c27b2f06a829870
>       ifname: lhnet1
>       podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
>   range: 192.168.50.0/24
> ```
> 
> 3. Simply turning on (fixing) the node, makes the whereabouts clear the overlapping ranges from the storage network but NOT clear the ip pool
> 
> ```
> harv-1:/home/rancher # kubectl get overlappingrangeipreservations -n kube-system
> NAME            AGE
> 192.168.100.1   86m
> 192.168.100.2   74m
> 192.168.100.3   2m41s
> harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io -n kube-system 
> NAME               AGE
> 192.168.100.0-24   86m
> 192.168.50.0-24    9m42s
> ```
> 
> 4. Hence that's why we clear up the ip pools on next enablement only, and leave the overlappingranges, whereabouts leave the former, while it clears up the latter

@martindekov

I have one question:
- After **step 3**, does the IPPool `192.168.50.0-24` have an empty `spec.allocations`, or does it retain leftover IPs? And if you reuse the storage network with `192.168.50.0-24` again, can Whereabouts still use the existing IPPool CR and works as expected?

My main concern is that the `ippool` and `overlappingrangeipreservation` CRs are fully managed by the Whereabouts controller, not by Harvester. The way Whereabouts manages its CRs may change across versions. Even if we only need to care about `ippool` right now, this may not hold for future versions—for example, a new CR has already been introduced in Whereabouts `v0.9.2`. If the Harvester controller deletes Whereabouts-managed CRs, as seen [here](https://github.com/martindekov/harvester/blob/8589556cd00b76779d9223e5b50be47993bd190a/pkg/controller/master/storagenetwork/storage_network.go#L830-L838), it can influence the behavior of the Whereabouts controller. 

It happens to work today, but it could introduce issues in future releases and reduces flexibility and generality in Harvester’s controller design.

In my opinion, if possible, Harvester should avoid modifying internal Whereabouts resources.

cc @Vicente-Cheng @rrajendran17 @w13915984028

martindekov:
@WebberHuang1118 If you fix the nodes and you've disabled the network, whereabouts will garbage collect the IP pool, but it relies on node going back to a working state. In real life however if you have outage, fixing the node to a running state is not always possible.

pkg/controller/master/storagenetwork/storage_network.go
martindekov: Tested the nil won't panic out also the commment states that in case of nil err not found returns false so changed :heavy_check_mark:

martindekov:
So to elaborate on the above ran the experiment

<details>
<summary>1. From step 3 where pod ended in terminating state as its node went off I enabled storage network the controller deleted the ip pool, but the overlapping ranges was left intact</summary>
<br>
```
harv-1:/home/rancher # kubectl get pods -n longhorn-system
NAME                                                READY   STATUS        RESTARTS        AGE
csi-attacher-7b996ff85b-gkn2s                       1/1     Running       0               2d21h
csi-attacher-7b996ff85b-j7xgj                       1/1     Running       0               2d21h
csi-attacher-7b996ff85b-q4kmj                       1/1     Running       1 (3h34m ago)   2d21h
csi-provisioner-b5c5f5cbf-crgwn                     1/1     Running       0               2d21h
csi-provisioner-b5c5f5cbf-rg5th                     1/1     Running       1 (3h34m ago)   2d21h
csi-provisioner-b5c5f5cbf-xz9nq                     1/1     Running       0               2d21h
csi-resizer-85fb9c8c9f-dvx7x                        1/1     Running       0               2d21h
csi-resizer-85fb9c8c9f-slnbn                        1/1     Running       0               2d21h
csi-resizer-85fb9c8c9f-tjjvd                        1/1     Running       1 (3h34m ago)   2d21h
csi-snapshotter-867b4445d6-dgpcr                    1/1     Running       0               2d21h
csi-snapshotter-867b4445d6-qkn7m                    1/1     Running       0               2d21h
csi-snapshotter-867b4445d6-qv674                    1/1     Running       1 (3h34m ago)   2d21h
engine-image-ei-3154f3aa-9n7mv                      1/1     Running       0               2d20h
engine-image-ei-3154f3aa-9pfbk                      1/1     Running       0               2d21h
instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4   1/1     Terminating   0               3m34s
instance-manager-74ec8021feeb76cf6e8cce35c5a81c01   1/1     Running       0               37s
longhorn-csi-plugin-lwmld                           3/3     Running       0               2d20h
longhorn-csi-plugin-s9jms                           3/3     Running       0               2d21h
longhorn-driver-deployer-785c6548d-qcdjb            1/1     Running       0               2d21h
longhorn-manager-74bxb                              2/2     Running       0               2d21h
longhorn-manager-gg8wk                              2/2     Running       0               2d20h
longhorn-ui-677666b49d-64kxt                        1/1     Running       0               2d21h
longhorn-ui-677666b49d-f6th9                        1/1     Running       0               2d21h
harv-1:/home/rancher # kubectl delete pod instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4 --grace-period=0 --force  -n longhorn-system
Warning: Immediate deletion does not wait for confirmation that the running resource has been terminated. The resource may continue to run on the cluster indefinitely.
pod "instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4" force deleted from longhorn-system namespace
harv-1:/home/rancher # kubectl get overlappingrangeipreservations -n kube-system
NAME            AGE
192.168.100.1   2d21h
192.168.100.3   2d20h
192.168.50.2    3m58s
harv-1:/home/rancher # kubectl get overlappingrangeipreservations/192.168.50.2 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: OverlappingRangeIPReservation
metadata:
  creationTimestamp: "2025-11-24T11:20:03Z"
  generation: 1
  name: 192.168.50.2
  namespace: kube-system
  resourceVersion: "239342"
  uid: c6371f4f-0e48-47c7-90d5-ffc0a17a8727
spec:
  ifname: lhnet1
  podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io -n kube-system 
NAME               AGE
192.168.100.0-24   2d21h
192.168.50.0-24    4m22s
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -n kube-system 
NAME              AGE
192.168.50.0-24   4m27s
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system 
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-24T11:20:03Z"
  generation: 4
  name: 192.168.50.0-24
  namespace: kube-system
  resourceVersion: "241390"
  uid: a9a3fd38-def6-4d14-a166-1cf8919a37b7
spec:
  allocations:
    "2":
      id: 6cfbc5f1aa75bcf44fde747bd5eede817e132ba0494e1fe9b8b619b9887ee763
      ifname: lhnet1
      podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
  range: 192.168.50.0/24
harv-1:/home/rancher # kubectl get overlappingrangeipreservations -n kube-system
NAME            AGE
192.168.100.1   2d21h
192.168.100.3   2d20h
192.168.50.2    4m43s
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system 
Error from server (NotFound): ippools.whereabouts.cni.cncf.io "192.168.50.0-24" not found
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system 
Error from server (NotFound): ippools.whereabouts.cni.cncf.io "192.168.50.0-24" not found
```
</details>

<details>
<summary>2. IP pools were recreated with 1 allocation but overlapping range with IP from 2nd node was left</summary>
<br>

```
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system 
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-24T11:25:11Z"
  generation: 2
  name: 192.168.50.0-24
  namespace: kube-system
  resourceVersion: "243268"
  uid: f9d0ea55-9054-4756-989b-1ad1f8165ab2
spec:
  allocations:
    "1":
      id: c70b629771a1da5e4b287c2244ba1755951f07026d6011d900e08382734800e1
      ifname: lhnet1
      podref: longhorn-system/instance-manager-74ec8021feeb76cf6e8cce35c5a81c01
  range: 192.168.50.0/24
harv-1:/home/rancher # kubectl get overlappingrangeipreservations -n kube-system
NAME            AGE
192.168.100.1   2d21h
192.168.100.3   2d20h
192.168.50.1    5s
192.168.50.2    5m13s
```
</details>

<details>
<summary>3. Adding 2nd node to the storage network took the IP and pod ID from previous run</summary>
<br>

```
arv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system 
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-24T11:25:11Z"
  generation: 2
  name: 192.168.50.0-24
  namespace: kube-system
  resourceVersion: "243268"
  uid: f9d0ea55-9054-4756-989b-1ad1f8165ab2
spec:
  allocations:
    "1":
      id: c70b629771a1da5e4b287c2244ba1755951f07026d6011d900e08382734800e1
      ifname: lhnet1
      podref: longhorn-system/instance-manager-74ec8021feeb76cf6e8cce35c5a81c01
  range: 192.168.50.0/24
harv-1:/home/rancher # kubectl get pods -n longhorn-system
NAME                                                READY   STATUS    RESTARTS        AGE
csi-attacher-7b996ff85b-gkn2s                       1/1     Running   0               2d21h
csi-attacher-7b996ff85b-j7xgj                       1/1     Running   0               2d21h
csi-attacher-7b996ff85b-q4kmj                       1/1     Running   1 (3h40m ago)   2d21h
csi-provisioner-b5c5f5cbf-crgwn                     1/1     Running   0               2d21h
csi-provisioner-b5c5f5cbf-rg5th                     1/1     Running   1 (3h40m ago)   2d21h
csi-provisioner-b5c5f5cbf-xz9nq                     1/1     Running   0               2d21h
csi-resizer-85fb9c8c9f-dvx7x                        1/1     Running   0               2d21h
csi-resizer-85fb9c8c9f-slnbn                        1/1     Running   0               2d21h
csi-resizer-85fb9c8c9f-tjjvd                        1/1     Running   1 (3h40m ago)   2d21h
csi-snapshotter-867b4445d6-dgpcr                    1/1     Running   0               2d21h
csi-snapshotter-867b4445d6-qkn7m                    1/1     Running   0               2d21h
csi-snapshotter-867b4445d6-qv674                    1/1     Running   1 (3h40m ago)   2d21h
engine-image-ei-3154f3aa-9pfbk                      1/1     Running   0               2d21h
engine-image-ei-3154f3aa-c6tlc                      1/1     Running   0               96s
instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4   1/1     Running   0               33s
instance-manager-74ec8021feeb76cf6e8cce35c5a81c01   1/1     Running   0               4m4s
longhorn-csi-plugin-58xzw                           3/3     Running   0               96s
longhorn-csi-plugin-s9jms                           3/3     Running   0               2d21h
longhorn-driver-deployer-785c6548d-qcdjb            1/1     Running   0               2d21h
longhorn-manager-74bxb                              2/2     Running   0               2d21h
longhorn-manager-bzp4s                              2/2     Running   0               96s
longhorn-ui-677666b49d-64kxt                        1/1     Running   0               2d21h
longhorn-ui-677666b49d-f6th9                        1/1     Running   0               2d21h
harv-1:/home/rancher # kubectl get overlappingrangeipreservations -n kube-system
NAME            AGE
192.168.100.1   2d21h
192.168.100.2   22s
192.168.50.1    4m7s
192.168.50.2    9m15s
harv-1:/home/rancher # kubectl get ippools.whereabouts.cni.cncf.io/192.168.50.0-24 -o yaml -n kube-system 
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: IPPool
metadata:
  creationTimestamp: "2025-11-24T11:25:11Z"
  generation: 3
  name: 192.168.50.0-24
  namespace: kube-system
  resourceVersion: "246105"
  uid: f9d0ea55-9054-4756-989b-1ad1f8165ab2
spec:
  allocations:
    "1":
      id: c70b629771a1da5e4b287c2244ba1755951f07026d6011d900e08382734800e1
      ifname: lhnet1
      podref: longhorn-system/instance-manager-74ec8021feeb76cf6e8cce35c5a81c01
    "2":
      id: f22e3ec031201629759999c0b2b4d02eaae401306001a7843c025fc94d79b00f
      ifname: lhnet1
      podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
  range: 192.168.50.0/24
harv-1:/home/rancher # kubectl get overlappingrangeipreservations/192.168.50.2 -o yaml -n kube-system
apiVersion: whereabouts.cni.cncf.io/v1alpha1
kind: OverlappingRangeIPReservation
metadata:
  creationTimestamp: "2025-11-24T11:20:03Z"
  generation: 1
  name: 192.168.50.2
  namespace: kube-system
  resourceVersion: "239342"
  uid: c6371f4f-0e48-47c7-90d5-ffc0a17a8727
spec:
  ifname: lhnet1
  podref: longhorn-system/instance-manager-4d3cfa1d987a8b31f3968c378fc4c1b4
```
</details>

Problem is the IDs coincide in real life and there is no update timestamp, simply the old overlapping range was successfully reused.

pkg/webhook/resources/setting/validator.go
```diff
@@ -1189,6 +1194,10 @@ func (v *settingValidator) validateUpdateStorageNetwork(oldSetting *v1beta1.Sett
 		return werror.NewInvalidError(err.Error(), settings.StorageNetworkName)
 	}
 
+	if err := v.checkLeftoverPodIPs(oldSetting, newSetting); err != nil {
+		return werror.NewInvalidError(err.Error(), settings.StorageNetworkName)
+	}
+
```
rrajendran17, line 1200: If this is handled by webhook to send error to users of this inconsistent state, then do we still need a logic in harvester storage network controller for cleaning up ipool resources ?
If the cleanup has to be handled by user, then it has to be documented as well.
(force deleting the terminating pod, updating the IPPool spec with empty allocations and updating the corresponding overlapping resource)

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -277,6 +279,14 @@ func (h *Handler) createNad(setting *harvesterv1.Setting) (*nadv1.NetworkAttachm
...
 }
 
+func (h *Handler) cleanUpAndCreateNad(setting *harvesterv1.Setting) (*nadv1.NetworkAttachmentDefinition, error) {
+	err := h.cleanUpOldSNIPPool(setting, util.KubeSystemNamespace)
+	if err != nil {
+		return nil, err
+	}
+	return h.createNad(setting)
+}
+
```
rrajendran17, line 289: 1.Do we need to delete the IPPool or instead just update the IPPool spec allocations as empty, this way it will be in sync with whereabouts logic to keep the IPPool intact once created.

2.We also need to update/delete the corresponding overlapping object when IPPool is deleted. (case where there is a failure after IPPool deletion from Harvester)

martindekov:
Closing this PR in favor of the bigger effort in the Epic: https://github.com/harvester/harvester/issues/9628

Issue #9454, opened 2025-11-10
[BUG] VM Volume expand failing with PX-CSI


clemenko:
### Describe the Bug

Customers of both RGS and Pure are running into an issue with PX-CSI `25.8.1` and Harvester `v1.6.1`. They are using a FlashArray storage device for the VMs with fiber channel. I have confirmed it is an issue with iscsi. 

issue : VM can't start after a powered off vm volume expansion.

![Image](https://github.com/user-attachments/assets/f11f69b4-90c1-4829-ac1f-5aa7a7227643)

![Image](https://github.com/user-attachments/assets/7d54f188-ebaa-4363-8a1c-ea29dcf87d05)

### To Reproduce

from https://github.com/clemenko/px-harvester/blob/main/readme.md

High level
- New install Harvester v1.6.1. 
- Update multipathd on Harvester node
- Install PX-CSI 25.8.1
- Add image on new storageclass
- boot vm with new sc
- stop vm
- expand volume
- start vm

### Expected Behavior

The volume can start after volume expansion

### Support Bundle for Troubleshooting

[supportbundle_9d8576e2-fa9e-4daa-bf52-1bf65495be8f_2025-11-10T15-39-56Z.zip](https://github.com/user-attachments/files/23457446/supportbundle_9d8576e2-fa9e-4daa-bf52-1bf65495be8f_2025-11-10T15-39-56Z.zip)

### Environment

- Harvester version:
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

ibrokethecloud:
When a PVC is resized, the csi expands the PV. The PVC will not reflect the correct size until its attached to a pod so the associated filesystem can be resized. Till then the PVC has the following condition

```
  - lastProbeTime: null
    lastTransitionTime: "2025-10-07T07:18:46Z"
    message: Waiting for user to (re-)start a pod to finish file system resize of
      volume on node.
    status: "True"
    type: FileSystemResizePending
```
We have a check in harvester which is verifying of any PVC is being resized. 

Unfortunately this results in a deadlock since the PVC will not have its FS resized until the VM is booted.

We need to change the check to validate the size of PV after a resizing operation.

https://github.com/harvester/harvester/blob/master/pkg/api/vm/handler.go#L321

For now the work around would be to use `virtctl start $VM` to boot the VM which will help you get around this issue.

clemenko:
[...]

PR #9475, opened 2025-11-11
minor changes to check evaluating if volume is resizing

ibrokethecloud:
#### Problem:

When a PVC is resized, the csi expands the PV. The PVC will not reflect the correct size until its attached to a pod so the associated filesystem can be resized. Till then the PVC has the following condition

```
  - lastProbeTime: null
    lastTransitionTime: "2025-11-10T15:38:04Z"
    status: "True"
    type: Resizing
  - lastProbeTime: null
    lastTransitionTime: "2025-11-10T15:38:04Z"
    message: Waiting for user to (re-)start a pod to finish file system resize of
      volume on node.
    status: "True"
    type: FileSystemResizePending
```

We have a check in harvester which is verifying of any PVC is being resized.

Unfortunately this results in a deadlock since the PVC will not have its FS resized until the VM is booted.

We need to change the check to validate the size of PV after a resizing operation.

#### Solution:

The PR changes the `IsResizing` check to only perform validation if provisioner is Longhorn.

The check was originally added in case users defined a pvc size which could not be satisifed by longhorn disks, and aim was to stop the vm from consuming the volume and avoid data corruption.

With 3rd party CSI, we can leave this for the CSI to manage.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9454

#### Test plan:

To test:

For Longhorn backed pvc:
* Create a VM booting off a longhorn based vmimage
* Stop the VM
* Edit the VM and expand the disk
* After saving the changes try and start the VM
* VM start action should fail with an error "can not start the VM $VMName which has a resizing volume $VolumeName"
* Once the disk is expanded successfully, the VM will start fine

For 3rd party csi backed pvc:
* Create a VM booting off a 3rd party csi based vmimage
* Stop the VM
* Edit the VM and expand the disk
* After saving the changes try and start the VM
* VM should start successfully

### DIFF
--- a/pkg/api/volume/formatter.go
+++ b/pkg/api/volume/formatter.go
@@ -31,7 +31,7 @@ func (f *volFormatter) Formatter(request *types.APIRequest, resource *types.RawR
 		return
 	}
 
-	if IsResizing(pvc) {
+	if IsResizing(pvc, f.scCache) {
 		resource.AddAction(request, actionCancelExpand)
 		return
 	}
@@ -51,8 +51,16 @@ func (f *volFormatter) Formatter(request *types.APIRequest, resource *types.RawR
 	}
 }
 
-func IsResizing(pvc *corev1.PersistentVolumeClaim) bool {
+func IsResizing(pvc *corev1.PersistentVolumeClaim, scCache ctlstoragev1.StorageClassCache) bool {
 	if pvc == nil {
+		return false
+	}
+
+	// check ensures we only apply IsReszing check when provisioner is longhorn
+	// for 3rd party CSI we ignore this validation
+	// related to https://github.com/harvester/harvester/issues/9454
+	provisioner := util.GetProvisionedPVCProvisioner(pvc, scCache)
+	if provisioner != util.CSIProvisionerLonghorn {
 		return false
 	}
 
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -51,8 +51,8 @@ import (
 	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
 	"github.com/harvester/harvester/pkg/util/drainhelper"
-	"github.com/harvester/harvester/pkg/util/virtualmachineinstance"
 	"github.com/harvester/harvester/pkg/util/virtualmachine"
+	"github.com/harvester/harvester/pkg/util/virtualmachineinstance"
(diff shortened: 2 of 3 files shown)


pkg/api/volume/formatter.go
```diff
@@ -56,6 +58,13 @@ func IsResizing(pvc *corev1.PersistentVolumeClaim) bool {
 		return false
 	}
 
+	// check ensures we only apply IsReszing check when provisioner is longhorn
+	// for 3rd party CSI we ignore this validation
+	// related to https://github.com/harvester/harvester/issues/9454
+	if val, ok := pvc.Annotations[provisionerAnnotationKey]; !ok || val != longhornProvisionerValue {
+		return false
+	}
```
WebberHuang1118: We can reuse the utility function [GetProvisionedPVCProvisioner](https://github.com/harvester/harvester/blob/f16f02037000971c3a4fa3bd46a6c80593e183ae/pkg/util/volume.go#L42-L63), as shown in [this example](https://github.com/harvester/harvester/blob/f16f02037000971c3a4fa3bd46a6c80593e183ae/pkg/api/vm/formatter.go#L364).

WebberHuang1118 (commented):
Thanks for the PR, just a nit, PTAL, thanks.

WebberHuang1118 (dismissed):
LGTM, thanks.

pkg/api/volume/formatter.go
```diff
@@ -10,10 +10,11 @@ import (
...
 const (
-	actionExport       = "export"
-	actionCancelExpand = "cancelExpand"
-	actionClone        = "clone"
-	actionSnapshot     = "snapshot"
+	actionExport             = "export"
+	actionCancelExpand       = "cancelExpand"
+	actionClone              = "clone"
+	actionSnapshot           = "snapshot"
+	longhornProvisionerValue = "driver.longhorn.io"
```
Vicente-Cheng: we have the existing const on https://github.com/harvester/harvester/blob/master/pkg/util/constants.go#L120
maybe we can use that?

Vicente-Cheng (dismissed):
just a nit, others lgtm
thanks for fixing that!

WebberHuang1118 (approved):

Vicente-Cheng (approved):

ibrokethecloud:
@mergify backport v1.7

Vicente-Cheng:
@mergifyio backport v1.6

PR #9477, opened 2025-11-12
chore: unify go version to match harvester-installer repo

tserong:
#### Problem:
The harvester and harvester-installer repos need to work together to build our ISO images, so it's important that they have the exact same golang version to avoid occasional weird build issues, for example as reported in https://github.com/harvester/harvester-installer/pull/1170#issuecomment-3516260269

#### Solution:
This commit switches from golang:1.25-bookworm to registry.suse.com/bci/golang:1.25 as is used in the harvester-installer repo. go.mod already specifies `go 1.25` so this means it will build using the latest in the v1.25 series as provided by the registry.suse.com/bci/golang:1.25 image (currently v1.25.3).

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9279

#### Test plan:
N/A

See also https://github.com/harvester/harvester-installer/pull/1188

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,7 +1,6 @@
 FROM quay.io/costoolkit/releases-teal:grub2-live-0.0.4-2 AS grub2-mbr
 FROM quay.io/costoolkit/releases-teal:grub2-efi-image-live-0.0.4-2 AS grub2-efi
-
-FROM golang:1.25-bookworm
+FROM registry.suse.com/bci/golang:1.25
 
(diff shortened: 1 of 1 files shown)


tserong:
Ugh, wait, this breaks installation of a bunch of other stuff via apt-get. Let's leave this alone for the moment (the other PR https://github.com/harvester/harvester-installer/pull/1188) should still be fine though.

w13915984028:
@tserong I close this PR for now to eliminate the bot reminder, thanks.

Issue #8945, opened 2025-08-21
[TASK] Revisit the `deleteVolumeAttachments` on the Harvester node controller


Vicente-Cheng:
### Task request

We introduce this mechanism with [[BUG] The VM needs at least 10 minutes to automatically recover after a host failure](https://github.com/harvester/harvester/issues/4049).

And we need to revisit this behavior because we should not delete the volume attachment directly. Longhorn has its own volume attachmentticket also. We should revisit this issue to see if we need to update for the latest mechanism.

cc @WebberHuang1118, @brandboat, @innobead

ibrokethecloud:
we can simplify this by having the `node_down_controller` just tainting the node with taint `node.kubernetes.io/out-of-service:NoExecute`

This should trigger clean up resources and we do not need to manually perform the cleanup.

 https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#non-graceful-node-shutdown

Vicente-Cheng:
We now will handle the following three major cases:
  - Disk pressure: do-no op
  - Kubelet temporary: trigger migration with VMForceResetPolicy
  - Node is not reachable: trigger migration with VMForceResetPolicy

albinsun:
Test PASS, close as done.


## Environment
* `harvester-v1.7.0-rc3`
  * Profile: 3 nodes AMD64 QEMU/KVM
  * **ui-source**: `Auto` (https://releases.rancher.com/harvester-ui/dashboard/release-harvester-v1.7/index.html)



## Test Scenario
### 🟢 3 Nodes
1. <details>
   <summary> Setup 3 nodes <code>harvester-v1.7.0-rc3</code> and create a VM
   </summary>

   <img width="1565" height="1037" alt="Image" src="https://github.com/user-attachments/assets/ba3af633-5725-489f-a1e4-50d9e6e624d1" />

   <img width="1565" height="474" alt="Image" src="https://github.com/user-attachments/assets/12f4f2db-8bd3-43b4-80fc-1bd90a5b2129" />
   </details>

1. <details>
   <summary> Check there are no taints on the node which VM is <code>Running</code>
   </summary>

   <img width="1565" height="1036" alt="Image" src="https://github.com/user-attachments/assets/9b8acf0c-b96b-4c79-83b2-cb382c620f6b" />
   </details>

   > vm on **node-1**

1. <details>
[...]

PR #9480, opened 2025-11-12
Improve node failure

Vicente-Cheng:
#### Problem:
Improve our mechanism when node failed

#### Solution:
Improve our mechanism when node failed

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8945

#### Test plan:
1. create 3 node cluster
2. create VM
3. disable the network on the host which the VM is running
4. ensure the VM can be migrated smoothly
5. ensure node which network was disabled have the following taints
`kubevirt.io/drain:NoSchedule`
`node.kubernetes.io/out-of-service`
6. enable the network. Ensure the above taints were removed after the node was back.

### DIFF
--- a/pkg/controller/master/node/node_down_controller.go
+++ b/pkg/controller/master/node/node_down_controller.go
@@ -3,23 +3,17 @@ package node
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"slices"
 	"time"
 
-	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
-	"go.uber.org/multierr"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/utils/set"
-	kubevirtv1 "kubevirt.io/api/core/v1"
+	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 
-	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/config"
-	v1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
-	ctlstoragev1 "github.com/harvester/harvester/pkg/generated/controllers/storage.k8s.io/v1"
 	"github.com/harvester/harvester/pkg/settings"
 )
 
@@ -27,154 +21,59 @@ const (
 	nodeDownControllerName = "node-down-controller"
 )
 
+// kubevirtDrainTaint is used to trigger migration
+var kubevirtDrainTaint = corev1.Taint{
+	Key:    virtconfig.NodeDrainTaintDefaultKey,
+	Effect: corev1.TaintEffectNoSchedule,
+}
+
+var nonGracefulTaint = corev1.Taint{
+	Key:    corev1.TaintNodeOutOfService,
+	Effect: corev1.TaintEffectNoExecute,
+}
+
+var (
+	// vars for HarvesterNodeDrained condition
+	HarvesterNodeCondDrained               = corev1.NodeConditionType("Drained")
+	HarvesterNodeDrainedCondReasonDraining = "HarvesterNodeIsDraining"
+	HarvesterNodeDrainedCondReasonDrained  = "HarvesterNodeIsDrained"
+	HarvesterNodeDrainedCondMsg            = "Node is draining due to kubelet/node not ready."
+)
+
 // nodeDownHandler force deletes VMI's pod when a node is down, so VMI can be reschduled to anothor healthy node
 type nodeDownHandler struct {
-	nodes                       ctlcorev1.NodeController
-	nodeCache                   ctlcorev1.NodeCache
-	pods                        ctlcorev1.PodClient
-	pvcCache                    ctlcorev1.PersistentVolumeClaimCache
-	vas                         ctlstoragev1.VolumeAttachmentClient
-	vaCache                     ctlstoragev1.VolumeAttachmentCache
-	virtualMachineInstanceCache v1.VirtualMachineInstanceCache
+	nodes     ctlcorev1.NodeController
+	nodeCache ctlcorev1.NodeCache
 }
 
 // DownRegister registers a controller to delete VMI when node is down
 func DownRegister(ctx context.Context, management *config.Management, _ config.Options) error {
 	nodes := management.CoreFactory.Core().V1().Node()
-	pods := management.CoreFactory.Core().V1().Pod()
-	setting := management.HarvesterFactory.Harvesterhci().V1beta1().Setting()
-	vmis := management.VirtFactory.Kubevirt().V1().VirtualMachineInstance()
-	pvcs := management.CoreFactory.Core().V1().PersistentVolumeClaim()
-	vas := management.HarvesterStorageFactory.Storage().V1().VolumeAttachment()
 	nodeDownHandler := &nodeDownHandler{
-		nodes:                       nodes,
-		nodeCache:                   nodes.Cache(),
-		pods:                        pods,
-		pvcCache:                    pvcs.Cache(),
-		vas:                         vas,
-		vaCache:                     vas.Cache(),
-		virtualMachineInstanceCache: vmis.Cache(),
+		nodes:     nodes,
+		nodeCache: nodes.Cache(),
 	}
 
 	nodes.OnChange(ctx, nodeDownControllerName, nodeDownHandler.OnNodeChanged)
-	setting.OnChange(ctx, nodeDownControllerName, nodeDownHandler.OnVMForceResetPolicyChanged)
 
 	return nil
 }
 
-// OnNodeChanged monitors whether a node is ready or not
-// Force delete a pod when all the below conditions are meet:
-// 1. VMForceResetPolicy is enabled.
-// 2. A node has been down for more than VMForceResetPolicy.Period seconds
-// 3. The owner of Pod is VirtualMachineInstance.
-// 4. The Pod is on a down node.
+// OnNodeChanged monitors we focus the following 3 things
+// 2. Kubelet not ready -> after vmForceResetPolicy.Period, add annotation to kickoff VM migration
+// 3. Node not reachable -> add annotation to kickoff VM migration directly
 func (h *nodeDownHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev1.Node, error) {
 	if node == nil || node.DeletionTimestamp != nil {
 		return node, nil
 	}
 
-	// get Ready condition
-	cond := getNodeCondition(node.Status.Conditions, corev1.NodeReady)
-	if cond == nil {
-		return node, fmt.Errorf("can't find %s condition in node %s", corev1.NodeReady, node.Name)
+	// we skip DiskPressure check here because kubelet already have the basic audit.
+	// if you want to do more thing when the node is under disk pressure, you can implement it.
+	if err := h.checkNodeReady(node); err != nil {
+		return node, fmt.Errorf("failed to check node ready for node %s: %w", node.Name, err)
 	}
 
-	// check whether node is healthy
-	if cond.Status == corev1.ConditionTrue {
-		return node, nil
-	}
-
-	// get VMForceResetPolicy setting
-	vmForceResetPolicy, err := settings.DecodeVMForceResetPolicy(settings.VMForceResetPolicySet.Get())
-	if err != nil {
-		return node, err
-	}
-
-	if !vmForceResetPolicy.Enable {
-		return node, nil
-	}
-
-	// if we haven't waited for vmForceResetPolicy.Period seconds, we enqueue event again
-	if time.Since(cond.LastTransitionTime.Time) < time.Duration(vmForceResetPolicy.Period)*time.Second {
-		deadline := cond.LastTransitionTime.Add(time.Duration(vmForceResetPolicy.Period) * time.Second)
-		logrus.Debugf("Enqueue node event again at %v", deadline)
-		h.nodes.EnqueueAfter(node.Name, time.Until(deadline))
-		return node, nil
-	}
-
-	// get VMI pods on unhealthy node
-	pods, err := h.pods.List(corev1.NamespaceAll, metav1.ListOptions{
-		LabelSelector: labels.Set{
-			kubevirtv1.AppLabel: "virt-launcher",
-		}.String(),
-		FieldSelector: "spec.nodeName=" + node.Name,
-	})
-	if err != nil {
-		return node, err
-	}
-
-	var errs error
-	pvcSet := set.New[string]()
-	gracePeriod := int64(0)
-	for i := range pods.Items {
-		pod := pods.Items[i]
-		logrus.Debugf("force delete pod %s/%s and VolumeAttachments", pod.Namespace, pod.Name)
-
-		if err := h.pods.Delete(
-			pod.Namespace,
-			pod.Name,
-			&metav1.DeleteOptions{
-				GracePeriodSeconds: &gracePeriod,
-			}); err != nil {
-			errs = multierr.Append(errs, fmt.Errorf("failed to delete pod %s/%s: %w", pod.Namespace, pod.Name, err))
-			continue
-		}
-
-		names, err := h.getPVCsName(&pod)
-		if err != nil {
-			errs = multierr.Append(errs, fmt.Errorf("failed to get PVCs name of pod %s/%s: %w", pod.Namespace, pod.Name, err))
-			continue
-		}
-		pvcSet.Insert(names...)
-	}
-
-	if err := h.deleteVolumeAttachments(node.Name, pvcSet); err != nil {
-		errs = multierr.Append(errs, fmt.Errorf("failed to delete VolumeAttachments: %w", err))
-	}
-	if errs != nil {
-		return node, errs
-	}
-
-	return h.resetHeartbeat(node)
-}
-
-func (h *nodeDownHandler) OnVMForceResetPolicyChanged(_ string, setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
-	if setting == nil || setting.DeletionTimestamp != nil ||
-		setting.Name != settings.VMForceResetPolicySettingName || setting.Value == "" {
-		return setting, nil
-	}
-
-	vmForceResetPolicy, err := settings.DecodeVMForceResetPolicy(setting.Value)
-	if err != nil {
-		return setting, err
-	}
-
-	if !vmForceResetPolicy.Enable {
-		return setting, nil
-	}
-
-	nodes, err := h.nodeCache.List(labels.Everything())
-	if err != nil {
-		return setting, err
-	}
-
-	for _, node := range nodes {
-		cond := getNodeCondition(node.Status.Conditions, corev1.NodeReady)
-		if cond != nil && cond.Status != corev1.ConditionTrue {
-			h.nodes.Enqueue(node.Name)
-		}
-	}
-	return setting, nil
+	return node, nil
 }
 
 func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.NodeConditionType) *corev1.NodeCondition {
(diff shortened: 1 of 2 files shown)


pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
+			break
+		}
 	}
+	return cond
+}
 
-	if !vmForceResetPolicy.Enable {
-		return node, nil
+// just logger when we detect node is under disk pressure, maybe we can do more things in the future
+func (h *nodeDownHandler) check_disk_pressure(node *corev1.Node) error {
```
brandboat: nit:
```suggestion
func (h *nodeDownHandler) checkDiskPressure(node *corev1.Node) error {
```

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
-			kubevirtv1.AppLabel: "virt-launcher",
-		}.String(),
-		FieldSelector: "spec.nodeName=" + node.Name,
-	})
-	if err != nil {
-		return node, err
+// here we need to check node ready condition, two cases here:
+// 1. node not reachable -> we can handle it directly
+// 2. kubelet not ready -> we need to wait for vmForceResetPolicy.Period seconds before handling
+func (h *nodeDownHandler) check_node_ready(node *corev1.Node) error {
```
brandboat: nit:
```suggestion
func (h *nodeDownHandler) checkNodeReady(node *corev1.Node) error {
```

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
+			break
 		}
-		if va.Spec.Source.PersistentVolumeName == nil {
+	}
+	return taint
+}
+
+// get_ip_from_other_nodes tries to get an reachable IP from other nodes
+// we directly get this annotation `etcd.rke2.cattle.io/node-address` set by RKE2
+func get_ip_from_other_nodes(currentNode *corev1.Node, nodes []*corev1.Node) string {
```
brandboat: nit:
```suggestion
func getIpFromOtherNodes(currentNode *corev1.Node, nodes []*corev1.Node) string {
```

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
+			return err
 		}
-
-		names, err := h.getPVCsName(&pod)
+		targetIP := get_ip_from_other_nodes(node, nodes)
+		if targetIP == "" {
+			logrus.Warnf("can't get reachable IP from other nodes for node %s, skip this time", node.Name)
+			return nil
+		}
+		connected, err := pingNode(targetIP)
```
brandboat: Could you please add some explanation here? I'm not quite understand why we need to ping other nodes here.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
+	default:
+		return fmt.Errorf("unknown status %s for condition %s in node %s", cond.Status, corev1.NodeReady, node.Name)
 	}
+	return nil
+}
 
-	vmForceResetPolicy, err := settings.DecodeVMForceResetPolicy(setting.Value)
+func (h *nodeDownHandler) tryWaitForKubelet(node *corev1.Node) (bool, error) {
+	// get VMForceResetPolicy setting
+	targetInterival := defaultTimeInterval
```
brandboat: nit:
```suggestion
	targetInterval := defaultTimeInterval
```

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
+	vmForceResetPolicy, err := settings.DecodeVMForceResetPolicy(settings.VMForceResetPolicySet.Get())
 	if err != nil {
-		return setting, err
+		return false, fmt.Errorf("failed to decode VMForceResetPolicy setting: %w", err)
 	}
 
-	if !vmForceResetPolicy.Enable {
-		return setting, nil
+	if vmForceResetPolicy.Enable {
+		targetInterival = time.Duration(vmForceResetPolicy.Period) * time.Second
```
brandboat: Now vmForceResetPolicy.Enable seems redundant, perhaps we can drop it?
I'm thinking we can move the `defaultTimeInterval` (15s) to setting - `VMForceResetPolicySet` as the default value. Then we can get rid of the if condition here.

And please ignore this comment if you don't have bandwidth, this is a nitpick I think.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
 
-	nodes, err := h.nodeCache.List(labels.Everything())
-	if err != nil {
-		return setting, err
+	// check if we wait enough (targetInterival) after the kubelet becomes not ready
+	cond := getNodeCondition(node.Status.Conditions, corev1.NodeReady)
+	if cond != nil && time.Since(cond.LastTransitionTime.Time) < targetInterival {
+		deadline := cond.LastTransitionTime.Add(targetInterival)
+		logrus.Debugf("Enqueue node event again at %v", deadline)
+		h.nodes.EnqueueAfter(node.Name, time.Until(deadline))
```
brandboat: We need to return false here, right? Otherwise, we won’t wait long enough until the deadline.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
-// causing the VM to restart to wait a long time.
-// Then we delete VolumeAttachments directly.
-// ref: https://github.com/harvester/harvester/issues/4049
-func (h *nodeDownHandler) deleteVolumeAttachments(nodeName string, pvcSet set.Set[string]) error {
-	if pvcSet.Len() == 0 {
-		return nil
+func (h *nodeDownHandler) addTaints(node *corev1.Node, taints ...corev1.Taint) error {
+	nodeCpy := node.DeepCopy()
+	for _, taint := range taints {
+		nodeCpy.Spec.Taints = append(nodeCpy.Spec.Taints, taint)
```
brandboat: ~Should we add a check here to prevent adding duplicate taints, so we can avoid unnecessary updates?~

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
-		if err != nil {
-			return node, err
-		}
-		if time.Since(hbt) > 5*time.Minute {
-			return node, nil
-		}
+	pinger, err := ping.NewPinger(ip)
+	if err != nil {
+		// create Pinger failed, skip testing, assume node is reachable
+		return connected, fmt.Errorf("create pinger failed, error: %s", err.Error())
```
brandboat: nit:
```suggestion
		return connected, fmt.Errorf("create pinger failed, error: %w", err)
```

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,315 @@ package node
...
+			taintKeys = append(taintKeys, kubevirtDrainTaintKey)
+		}
+		if taint := getNodeTaint(node.Spec.Taints, nonGracefulTaintKey); taint != nil {
+			taintKeys = append(taintKeys, nonGracefulTaintKey)
+		}
+		if len(taintKeys) > 0 {
+			if err := h.removeTaints(node, taintKeys...); err != nil {
+				return fmt.Errorf("failed to remove taints %v from node %s: %w", taintKeys, node.Name, err)
+			}
+		}
```
brandboat: nit: we already take care the existence checks in removeTaints.
```suggestion
		if err := h.removeTaints(node, kubevirtDrainTaintKey, nonGracefulTaintKey); err != nil {
			return fmt.Errorf("failed to remove taints from node %s: %w", node.Name, err)
		}
```

pkg/controller/master/node/node_down_controller.go
brandboat: Please ignore this comment, I saw we have done such existence check in before we invoke addTaints

pkg/settings/settings_helper.go
```diff
@@ -67,7 +67,7 @@ func (target *BackupTarget) IsDefaultBackupTarget() bool {
 func InitVMForceResetPolicy() string {
 	policy := &VMForceResetPolicy{
 		Enable: true,
-		Period: 5 * 60, // 5 minutes
+		Period: 30, // 30 seconds
```
ibrokethecloud: the default originally was 300 seconds. just wanted to make sure if this is intentional?

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: updated, thanks for the suggestion!

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: we drop pinger

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: nice catch!
updated

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: I prefer to keep that default on the nodeDown controller once we change the default value for `vmForceResetPolicy`.
WDYT?

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: udpated, thanks!

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: we drop piner

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: updated

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: updated

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: updated

pkg/settings/settings_helper.go
Vicente-Cheng: I don't think so.. 5 mins is too long, it close to Kubevirt internal heartbeat timeout.

pkg/controller/master/node/node_down_controller.go
brandboat: Did you mean set defaultTimeInterval (15s) as default value in vmForceResetPolicy.Period?

Vicente-Cheng: No, I mean keep the defaultTimeInterval on controller if user set `vmForceResetPolicy.Enabled` to false, we can still have the default timeout value here.

brandboat: Got it, thanks for the explanation. I'm fine with the change.

brandboat (dismissed):
LGTM, thanks!

brandboat:
gentle ping @Vicente-Cheng, now we dropped the go-ping lib, could you please remove the related libs under vendor? Thanks. ([e0f3732] not required anymore)

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,239 @@ package node
...
-	if errs != nil {
-		return node, errs
-	}
+		// we wait enough or node is unreachable, add taints to trigger migration
+		if getNodeTaint(node.Spec.Taints, kubevirtDrainTaintKey) == nil {
+			if err := h.triggerMigration(node); err != nil {
+				return fmt.Errorf("failed to trigger migration for node %s: %w", node.Name, err)
+			}
+			logrus.Infof("triggered migration for node %s", node.Name)
+			h.nodes.EnqueueAfter(node.Name, 5*time.Second)
```
WebberHuang1118: Just a question — is the purpose of the requeuing behavior to handle situations where the node becomes ready again?

WebberHuang1118 (dismissed):
LGTM, just a question, thanks.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,239 @@ package node
...
 	"github.com/harvester/harvester/pkg/config"
-	v1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
-	ctlstoragev1 "github.com/harvester/harvester/pkg/generated/controllers/storage.k8s.io/v1"
 	"github.com/harvester/harvester/pkg/settings"
 )
 
 const (
 	nodeDownControllerName = "node-down-controller"
+	kubevirtDrainTaintKey  = "kubevirt.io/drain"
+	nonGracefulTaintKey    = "node.kubernetes.io/out-of-service"
```
ihcsim: can we use the `corev1.TaintNodeOutOfService` as the value, instead of the literal?

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,239 @@ package node
...
 func (h *nodeDownHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev1.Node, error) {
 	if node == nil || node.DeletionTimestamp != nil {
 		return node, nil
 	}
 
-	// get Ready condition
-	cond := getNodeCondition(node.Status.Conditions, corev1.NodeReady)
-	if cond == nil {
-		return node, fmt.Errorf("can't find %s condition in node %s", corev1.NodeReady, node.Name)
+	if err := h.checkDiskPressure(node); err != nil {
```
ihcsim: is this necessary? i think K8s already sends out `event`s. or if we want to keep it, consider making it more generic (like say `h.checkNodeConditions()`) to include the `NodeMemoryPressure`, `NodePIDPressure`, `NodeNetworkUnavailable` [conditions](https://pkg.go.dev/k8s.io/api/core/v1#NodeConditionType)?

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,239 @@ package node
...
-		if va.Spec.Source.PersistentVolumeName == nil {
-			continue
-		}
-		if !pvcSet.Has(*va.Spec.Source.PersistentVolumeName) {
+func (h *nodeDownHandler) removeTaints(node *corev1.Node, taintKeys ...string) error {
+	nodeCpy := node.DeepCopy()
+	taints := nodeCpy.Spec.Taints
+	newTaints := make([]corev1.Taint, 0, len(taints))
+	for _, taint := range taints {
+		if slices.Contains(taintKeys, taint.Key) {
```
ihcsim: fwiw, you should be able to just use `slices.DeleteFunc()` to remove `taintKeys` from `taints`:

```diff
@@ -213,14 +211,15 @@ func (h *nodeDownHandler) addTaints(node *corev1.Node, taints ...corev1.Taint) e
 func (h *nodeDownHandler) removeTaints(node *corev1.Node, taintKeys ...string) error {
        nodeCpy := node.DeepCopy()
        taints := nodeCpy.Spec.Taints
-       newTaints := make([]corev1.Taint, 0, len(taints))
-       for _, taint := range taints {
-               if slices.Contains(taintKeys, taint.Key) {
-                       continue
+       slices.DeleteFunc(taints, func(key string) bool {
+               for _, remove := range taintKeys {
+                       if key == remove {
+                               return true
+                       }
                }
-               newTaints = append(newTaints, taint)
-       }
-       nodeCpy.Spec.Taints = newTaints
+               return false
+       })
+       nodeCpy.Spec.Taints = taints
        var err error
        if !reflect.DeepEqual(node.Spec.Taints, nodeCpy.Spec.Taints) {
                _, err = h.nodes.Update(nodeCpy)
```

pkg/controller/master/node/node_down_controller.go
ihcsim: Consider using the upstream consts instead of literals:

* `kubevirt.io/kubevirt/pkg/virt-config.NodeDrainTaintDefaultKey`
* `corev1.TaintNodeOutOfService`

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,239 @@ package node
...
 
-	if err := h.deleteVolumeAttachments(node.Name, pvcSet); err != nil {
-		errs = multierr.Append(errs, fmt.Errorf("failed to delete VolumeAttachments: %w", err))
-	}
-	if errs != nil {
-		return node, errs
-	}
+		// we wait enough or node is unreachable, add taints to trigger migration
+		if getNodeTaint(node.Spec.Taints, kubevirtDrainTaintKey) == nil {
+			if err := h.triggerMigration(node); err != nil {
```
ihcsim: I think it's not necessary to trigger migration because the VMs are configured with `.spec.evictionStrategy: LiveMigrateIfPossible`. From my brief testing, the kubevirt auto-starts the live migration the moment when the node is marked as out-of-service:

* create 2-node cluster
* start a vm on the 2nd non-primary node
* stop kubelet on the node where vm is running: `sudo systemctl stop rke2-agent`
* watch `kubectl get node` output and see 2nd node transitions to `NotReady` state
* add out-of-service taint to 2nd node - live migration happens immediately

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,239 @@ package node
...
-				GracePeriodSeconds: &gracePeriod,
-			}); err != nil {
-			errs = multierr.Append(errs, fmt.Errorf("failed to delete pod %s/%s: %w", pod.Namespace, pod.Name, err))
-			continue
-		}
+// We treat the node not ready (False/Unknown) as node down
+// In this situation, we will wait for the extra timeout and then add taints to trigger migration and cleanup
+func (h *nodeDownHandler) checkNodeReady(node *corev1.Node) error {
+	// get Ready condition
+	cond := getNodeCondition(node.Status.Conditions, corev1.NodeReady)
```
ihcsim, line 95: This looks reasonable to me. From my testing, the moment kubelet is down, all the major DaemonSets like `virt-handler` and `longhorn-manager` stop working, which leads me to think the chances of workloads continue operating in the background with a dead kubelet, are very unlikely.

pkg/controller/master/node/node_down_controller.go
brandboat: I have the same observation in my local env.

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: What we requeue is to give a chance for the migration. In most case, we can ignore the migration, but if only the kubelet temporary down, the migration might complete.

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: Thanks, I added the comment and drop the diskpressure operation

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: nice suggestion!
updated!

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: updated

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: I do some tests on it, if we directly taint the out-of-service, the pod would be delete directly instead of migration.
So we still need kubevirt drain taint.

And I thought your case might be related to the kubevirt heartbeat, once kubevirt heartbeat timeout, it also trigger migration.

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: Now, kubelet/node down will wait for the VMForceResetPolicy.Period. Then taint the kubevirt drain.
After that, we will wait another timeout for the VM migration (also user configurable), then taint the out-of-service.

Vicente-Cheng:
@mergifyio backport v1.7

pkg/controller/master/node/node_down_controller.go
ihcsim: Ah.. so putting the node out-of-service actual _deletes_, not _evicts_ the `virt-launcher` pod. Migration happens when the pod is evicted (read as _node drained_). Deleting the pod causes the VM to be restarted (not live-migrated because no VMIM was ever created), which was probably what I saw.

In light of that, yes, we will need a separate kubevirt drain step to trigger live migration. If live migration failed, then the out-of-service taint nukes all the stuck pods.

I haven't tested this (fully) yet, but is it safe to assume live migration works if kubelet on the node is down?

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,379 @@ package node
...
+			nodeCpy.Status.Conditions[id] = newCond
 		}
-		names = append(names, pvc.Spec.VolumeName)
 	}
-	return names, nil
+	if _, err := h.nodes.UpdateStatus(nodeCpy); err != nil {
+		return err
+	}
+	return nil
 }
```
WebberHuang1118: Can we combine `addNodeCondition()` and `updateNodeCondition()` into a single function that handles both the add and update cases for node conditions internally? This would make [this code snippet](https://github.com/Vicente-Cheng/harvester/blob/6a7f1bc931b8db599c333a400ce0043eebeefca9/pkg/controller/master/node/node_down_controller.go#L219-L227) more concise.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,379 @@ package node
...
+		logrus.Debugf("Enqueue node event again at %v", time.Now().Add(targetInterval))
+		h.nodes.EnqueueAfter(node.Name, time.Until(time.Now().Add(targetInterval)))
+		return false, nil
+	}
+	if time.Since(cond.LastTransitionTime.Time) < targetInterval {
+		deadline := cond.LastTransitionTime.Add(targetInterval)
+		logrus.Debugf("Enqueue node event again at %v", deadline)
+		h.nodes.EnqueueAfter(node.Name, time.Until(deadline))
+		return false, nil
+	}
```
WebberHuang1118, line 286: Do we need separate code paths for these two cases? As far as I understand, regardless of whether the condition is nil or not, we can set the node condition anyway and then requeue with an appropriate timeout, correct?

WebberHuang1118 (commented):
LGTM generally, just some suggestions, thanks.

pkg/controller/master/node/node_down_controller.go
ihcsim: What I'm trying to say is that we can count on this condition to decide whether to migrate/evict VM or not. When kubelet is down and the node is marked not ready, critical components like `virt-handler` and `longhorn-manager` on that node will be down (even before we do anything with the VM or the node). Hence, we don't need to worry about e.g., "is it safe to move this vm, is it still doing disk I/O ops etc.".

pkg/controller/master/node/node_down_controller.go
ihcsim: What i am seeing is that if kubelet is down, live migration won't work. Eventually, KubeVirt detects the node is down, live migration fails and the VM is restarted. This makes me wonder whether tainting/draining the node is even necessary when the node is down.

```sh
# on isim-dev2, run systemctl stop rke2-agent to kill kubelet
✗ k get node 
NAME        STATUS     ROLES                       AGE   VERSION
isim-dev    Ready      control-plane,etcd,master   25h   v1.33.3+rke2r1
isim-dev2   NotReady   <none>                      25h   v1.33.3+rke2r1

✗ k taint node isim-dev2 kubevirt.io/drain:NoSchedule
node/isim-dev2 tainted

# the vmim remains in scheduling state indefintely
✗ k get vm,vmi,po,vmim
NAME                               AGE   STATUS    READY
virtualmachine.kubevirt.io/vm-00   24h   Running   False

NAME                                       AGE     PHASE     IP           NODENAME    READY
virtualmachineinstance.kubevirt.io/vm-00   5m49s   Running   10.52.1.21   isim-dev2   False

NAME                            READY   STATUS      RESTARTS   AGE
pod/virt-launcher-vm-00-b6mxc   0/2     Init:0/1    0          90s
pod/virt-launcher-vm-00-xtt6n   2/2     Running     0          4m35s

NAME                                                                    PHASE        VMI
virtualmachineinstancemigration.kubevirt.io/kubevirt-evacuation-26xzg   Scheduling   vm-00

# because the virt-handler on the source node is down
✗ k -n harvester-system logs virt-handler-tj697 
Defaulted container "virt-handler" out of: virt-handler, virt-launcher (init)
Error from server: Get "https://192.168.122.20:10250/containerLogs/harvester-system/virt-handler-tj697/virt-handler": proxy error from 127.0.0.1:9345 while dialing 192.168.122.20:10250, code 502: 502 Bad Gateway

# migration timed out, kubevirt detects node is down... vm is restarted
✗ k get vm,vmi,po
NAME                               AGE   STATUS     READY
virtualmachine.kubevirt.io/vm-00   24h   Starting   False

NAME                                       AGE   PHASE        IP    NODENAME   READY
virtualmachineinstance.kubevirt.io/vm-00   21s   Scheduling                    False

NAME                            READY   STATUS     RESTARTS   AGE
pod/virt-launcher-vm-00-rrw52   0/2     Init:0/1   0          21s
```

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,379 @@ package node
...
+	// check if we wait enough (vmMigrationTimeout) after the kubevirtDrainTaint is added
+	cond := getNodeCondition(node.Status.Conditions, HarvesterNodeCondDrained)
+	if cond != nil && cond.Status == corev1.ConditionTrue {
+		// already drained
+		return true, nil
+	}
+	targetInterval := time.Duration(vmMigrationTimeout) * time.Second
+	if cond == nil {
+		// means we need to change the condition to unreachable (add a new one or update the existing one)
+		if err := h.setHarvesterNodeDrainedCondition(node, corev1.ConditionFalse, HarvesterNodeFailureCondReasonUnreachable, HarvesterNodeFailureCondMsgUnreachable); err != nil {
```
ihcsim: Re: the `reason` for this condition (here and in line 131), normally, we specify the reason (not the cause) for the true/false status of a condition. In our case, the node drained condition is false because draining is still in progress or the timeout hasn't expired. Conversely in line 131, the node drained condition is true because draining is completed or the timeout has expired. 

Node failure is the cause that triggers this condition (which can be the `message`), not the reason why the condition is in a certain state.

pkg/settings/settings_helper.go
```diff
@@ -78,8 +81,19 @@ func InitVMForceResetPolicy() string {
 
 func DecodeVMForceResetPolicy(value string) (*VMForceResetPolicy, error) {
 	policy := &VMForceResetPolicy{}
+	type oldPolicy struct {
+		Enable bool `json:"enable"`
+		Period int  `json:"period"`
+	}
```
ihcsim: Is this necessary? The un-marshalling should still work even if the `vmMigrationTimeout` field is missing in the JSON object. See https://go.dev/play/p/gif9EvNfedw.

ibrokethecloud: we can make the `VMMigrationTimeout` as optional field, and then same object can be used to unmarshal new or old settings, and we can apply default if one does not exist

pkg/settings/settings_helper.go
```diff
@@ -32,6 +32,8 @@ type VMForceResetPolicy struct {
 	Enable bool `json:"enable"`
 	// Period means how many seconds to wait for a node get back.
 	Period int64 `json:"period"`
+	// VMMigrationTimeout means how many seconds to wait for a VM to migrate when a node is reported not ready.
+	VMMigrationTimeout int64 `json:"vmMigrationTimeout"`
```
ibrokethecloud:
```suggestion
	VMMigrationTimeout int64 `json:"vmMigrationTimeout,omitempty"`
```

ibrokethecloud (commented):
the two methods `tryWaitForExtraTimeout` and `tryWaitForVMMigration` have a lot of common logic which could be factored into a common method say `waitForConditionTimeout(node *corev1.Node, conditionType corev1.NodeConditionType, timeout int64)` and all this method does is requeue and return if condition has not expired.

This way object keeps getting requeued until timeout has expired else code flow continues, and will make it easier to track how the node object is being requeued to ensure we are waiting for timeout to occur.

we should also call the method `setHarvesterNodeDrainedCondition`  as soon as kubevirt drain taint is called. 

this will ensure `addOutOfServiceTaintAfterMigrationTimeout` is only looking for timeout of our custom node drain condition before marking node out of service.

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: Yeah, I think most cases might be that. But I still want to follow the formal path, we cannot verify every cases. The extra drain taint would not affect too much.

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: A new function called `createOrUpdateNodeCondition` replaces these two.

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: restruct here. we use the same waiting function with different condition

pkg/controller/master/node/node_down_controller.go
Vicente-Cheng: thanks for the suggestion.
Now 
HarvesterNodeDrainedCondReasonDraining -> for status = False
HarvesterNodeDrainedCondReasonDrained  -> for status = True
HarvesterNodeDrainedCondMsg -> for dedicated msg

pkg/settings/settings_helper.go
Vicente-Cheng: thanks for the tips, updated

pkg/settings/settings_helper.go
Vicente-Cheng: updated

pkg/controller/master/node/node_down_controller.go
brandboat: Thanks for the explanation. Just to confirm—if a node goes down, we’ll have to wait until the migration timeout expires before the VM can come back online, correct?

Vicente-Cheng: Yeah, the default value is much longer (3 mins), but users can set their own.
For example, if they set the period to 15 seconds, the VM migration timeout to 15 seconds, which means it takes 30 seconds to ensure everything is cleaned up and the VM is starting on the new node.

WebberHuang1118 (dismissed):
LGTM, thanks.

ibrokethecloud (approved):
thanks for the enhancement

pkg/controller/master/node/node_down_controller.go
```diff
@@ -3,278 +3,343 @@ package node
...
 
-		names, err := h.getPVCsName(&pod)
+		// try to add Kubevirt drain taint after waiting extra timeout
+		if waitEnough, err := h.addKubevirtTaintAfterExtraTimeout(node, cond, vmForceResetPolicy.Period); err != nil {
+			return err
+		} else if !waitEnough {
+			return nil
+		}
+
+		node, err = h.createOrUpdateNodeCondition(node, HarvesterNodeCondDrained, corev1.ConditionFalse, HarvesterNodeDrainedCondReasonDraining, HarvesterNodeDrainedCondMsg)
```
brandboat, line 126: Will this cause busy loop when node down?  There is no reflect.deepEqual check in `createOrUpdateNodeCondition`, when we invoke `createOrUpdateNodeCondition` it will trigger another reconcile loop in this controller, right?

Vicente-Cheng: We have checked both taints to see if we have already tainted. We will return
```
        if getNodeTaint(node.Spec.Taints, virtconfig.NodeDrainTaintDefaultKey) != nil && getNodeTaint(node.Spec.Taints, corev1.TaintNodeOutOfService) != nil {
			// already added both taints, no need to process again
			return nil
		}
```

So, it would not cause the busy loop here.

ihcsim (approved):

brandboat (approved):
LGTM, thanks!

Vicente-Cheng:
@mergifyio backport v1.7

PR #9512, opened 2025-11-14
Pausable node upgrade

starbops:
#### Problem:

In the current Harvester Upgrade design, it's impossible to pause a node entering the node-upgrade state beforehand. This causes headaches, such as more extended downtime for user workloads.

#### Solution:

Let users specify which nodes should be paused before entering the node-upgrade state —specifically, before the pre-draining state, by configuring the newly introduced `nodeUpgradeOption` under the `upgrade-config` setting.

Such a setting will be translated to a node-to-action map in an annotation and attached to the Upgrade CR by the mutator upon creation. This intermediate representation is the source of truth that the controllers honor. An example is like below:

```yaml
...
metadata:
  harvesterhci.io/node-upgrade-pause-map: '{"bravo-1-tink-system": "pause", "bravo-2-tink-system": "pause"}'
  ...
```

When the upgrade proceeds to the node-upgrade phase, the embedded Rancher will pick a node to conduct the node-upgrade tasks. If such a node is marked as `pause` in the Upgrade CR's annotation, the pre-drain job won't be created by the Harvester upgrade controller; otherwise, the job is created as usual. When any node is being paused, the per-node upgrade status field will show the pause state. For instance, the following indicates that node `bravo-1-tink-system` is paused for node upgrade:

```yaml
status:
  nodeStatuses:
    bravo-1-tink-system:
      state: Node-upgrade paused
      reason: AdministrativelyPaused
      message: Node upgrade paused as requested by the user
    bravo-2-tink-system:
      state: Images preloaded
```

To un-pause a paused node, annotate the corresponding annotation in the Upgrade CR to `unpause`. The upgrade process will proceed until the next request-to-pause node (if any).

```yaml
...
metadata:
  harvesterhci.io/node-upgrade-pause-map: '{"bravo-1-tink-system": "unpause", "bravo-2-tink-system": "pause"}'
  ...
```

> [!NOTE]
> It's hard to know which node will be upgraded next from Harvester's viewpoint (the embedded Rancher decides the order). Harvester, as a consumer, can at best only pause a node from undergoing an upgrade by not creating the pre-drain job. However, the node will still be put into the cordon state (but not drained yet, fortunately). In the end, users should still be able to move things out from that node.

> [!NOTE]
> For experienced users (or automation, probably), it's entirely feasible to add the pause node annotations directly to the Upgrade CR without configuring the `upgrade-config` setting. As long as the node hasn't ever entered the node-upgrade state, the annotation is deemed valid and will take effect. The `nodeUpgradeOption` in the `upgrade-config` setting provides a simple way for users to configure without having to annotate the Upgrade CR every time.

#### Related Issue(s):

#8980

#### Test plan:

**Single-node upgrade**

1. Prepare a single-node cluster provisioned using a custom-built ISO image containing the enhancement
2. Configure the `upgrade-config` setting, adding the following options:
   ```
   ...
   "nodeUpgradeOption": {
     "strategy": {
       "mode": "manual"
     }
   },
   ...
   ```
3. Trigger an upgrade with the same ISO image
4. Observe the created Upgrade CR, it should contain the rendered annotation `harvesterhci.io/<node_name>: pause`
5. The upgrade should be paused at the node-upgrade phase until the above annotation is updated to `unpause`
   - The `NodesUpgraded` condition should remain `Unknown`
   - The `nodeStatuses` for the node should becomes `AdministrativelyPaused`
6. Unpause the node by updating the Upgrade CR's annotation for the node to `unpause`
7. The upgrade should immediately continue
8. The upgrade should complete successfully

**Multi-node upgrade**

1. Prepare a 3-node cluster provisioned using a custom-built ISO image containing the enhancement
2. Configure the `upgrade-config` setting, adding the following options:
   ```
   ...
   "nodeUpgradeOption": {
     "strategy": {
       "mode": "manual",
       "pauseNodes": [
         "<node-0>",
         "<node-2>"
       ]
     }
   },
   ...
   ```
   Only two of the three nodes specified in the `pauseNodes` list will be paused during node upgrades.
3. Trigger an upgrade with the same ISO image
4. Observe the created Upgrade CR, it should contain the rendered annotations `harvesterhci.io/<node_name>: pause`
5. The upgrade should be paused when one of the above two nodes entering the node-upgrade state, until the annotation is updated to `unpause`
   - The `NodesUpgraded` condition should remain `Unknown`
   - The `nodeStatuses` for the node should becomes `AdministrativelyPaused`
6. Unpause the node by updating the Upgrade CR's annotation for the node to `unpause`
7. The upgrade should immediately continue
8. Repeat step 5 (if all the pause nodes annotations are updated to unpause, go to step 9)
9. The upgrade should complete successfully


pkg/controller/master/upgrade/secret_controller.go
```diff
@@ -75,9 +76,17 @@ func (h *secretHandler) OnChanged(_ string, secret *v1.Secret) (*v1.Secret, erro
...
 	}
 
+	upgradeCpy := upgrade.DeepCopy()
+
 	switch upgrade.Status.NodeStatuses[nodeName].State {
 	case nodeStateImagesPreloaded:
 		if secret.Annotations[rke2PreDrainAnnotation] != secret.Annotations[preDrainAnnotation] {
+			if shouldPauseNodeUpgrade(upgrade, nodeName) {
+				logrus.Infof("Pause creating pre-drain job on %s", nodeName)
+				setPausedCondition(upgradeCpy, v1.ConditionTrue, "NodeUpgrade", fmt.Sprintf("node upgrade for %s is administratively paused as requested", nodeName))
```
brandboat: I was just thinking — do we have any plan to support upgrading multiple nodes at the same time in the future?
If yes, maybe it would make more sense to put the pause status under .status.nodeStatuses[node_name], since the current condition might get overwritten when another node gets paused.

But for now, since we only upgrade nodes one by one, I’m totally fine with the current approach.

pkg/controller/master/upgrade/secret_controller.go
```diff
@@ -75,9 +76,17 @@ func (h *secretHandler) OnChanged(_ string, secret *v1.Secret) (*v1.Secret, erro
...
+				break
+			}
+			setPausedCondition(upgradeCpy, v1.ConditionFalse, "", "")
```
brandboat: I was just thinking — do we have any plan to support upgrading multiple nodes at the same time in the future?
If yes, maybe it would make more sense to put the pause status under .status.nodeStatuses[node_name], since the current condition might get overwritten when another node gets paused.

But for now, since we only upgrade nodes one by one, I’m totally fine with the current approach.

pkg/controller/master/upgrade/common.go
```diff
@@ -53,6 +53,18 @@ fi
...
+func shouldPauseNodeUpgrade(upgrade *harvesterv1.Upgrade, nodeName string) bool {
+	key := fmt.Sprintf("%s/%s", util.Prefix, nodeName)
+	val, ok := upgrade.Annotations[key]
+	if !ok {
+		return false
+	}
+	if val != "pause" {
+		return false
+	}
+	return true
```
brandboat: nit: 
```suggestion
    return val == "pause"
```

pkg/controller/master/upgrade/register.go
```diff
@@ -21,6 +28,43 @@ const (
...
+				continue
+			}
+			node, err := h.nodeCache.Get(nodeName)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					continue
+				}
+				return nil, err
+			}
+			machineName, ok := node.Annotations["cluster.x-k8s.io/machine"]
```
brandboat: perhaps use `MachineAnnotation` under package `"sigs.k8s.io/cluster-api/api/v1beta1"`?

pkg/webhook/resources/setting/validator.go
```diff
@@ -1814,7 +1814,47 @@ func validateUpgradeConfigHelper(setting *v1beta1.Setting) (*settings.UpgradeCon
 	return config, nil
 }
 
-func validateUpgradeConfigFields(setting *v1beta1.Setting) error {
+func (v *settingValidator) checkNodesExist(nodeNames []string) error {
```
brandboat, line 1827: Do we need to add a check to ensure all node names are unique, just to be safe?

pkg/controller/master/upgrade/secret_controller.go
brandboat: L89 won't take effect if `checkEligibleToDrain` is true, in L91 we directly return without invoking `h.upgradeClient.Update`

brandboat (commented):
Thanks for the patch!

pkg/controller/master/upgrade/register.go
```diff
@@ -21,6 +28,43 @@ const (
...
+	nodeCache ctlcorev1.NodeCache
+}
+
+func (h *handler) NotifyUnpausedMachinePlanSecret(_ string, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
+	if upgrade, ok := obj.(*harvesterv1.Upgrade); ok {
+		for key, val := range upgrade.Annotations {
+			if val == "pause" {
+				continue
+			}
+			nodeName, ok := extractNodeName(key)
```
ihcsim: This means we can only unpause one node at a time, right?

pkg/controller/master/upgrade/secret_controller.go
ihcsim: can we put the pre-drain job in `suspend` mode, instead of `break`-ing at line 87? it will make future support of multi-nodes upgrade easier.

starbops: Thank you for the suggestion. Besides reflecting node upgrade pauses in node-specific upgrade status fields, I will also replace the `Paused` condition with `Degraded` to support future multi-node upgrade cases better. At least the term makes semantic sense. WDYT?

brandboat: +1, sounds great!

brandboat (dismissed):
LGTM, thanks!

pkg/apis/harvesterhci.io/v1beta1/upgrade.go
```diff
@@ -22,6 +22,10 @@ var (
 	NodesUpgraded condition.Cond = "NodesUpgraded"
 	// SystemServicesUpgraded is true when Harvester chart is upgraded
 	SystemServicesUpgraded condition.Cond = "SystemServicesUpgraded"
+
+	// Degraded is true when the upgrade process is jammed due to any node upgrade pause made by the user
+	// The condition is purely representational and does not effectively interfere with any of the above upgrade phases
+	Degraded condition.Cond = "Degraded"
```
w13915984028: Could `AdminPaused` or similar be more straightforward?

pkg/controller/master/upgrade/register.go
```diff
@@ -21,6 +29,43 @@ const (
...
+				return nil, err
+			}
+			machineName, ok := node.Annotations[capiv1.MachineAnnotation]
+			if !ok {
+				return nil, fmt.Errorf("machine name not found on node %s", node.Name)
+			}
+			return []relatedresource.Key{
+				{
+					Name:      fmt.Sprintf("%s-machine-plan", machineName),
+					Namespace: "fleet-local",
```
w13915984028: there is a defined const  `fleet-local`

pkg/util/constants.go
```diff
@@ -1,86 +1,86 @@
 package util
 
 const (
-	prefix                              = "harvesterhci.io"
-	RemovedPVCsAnnotationKey            = prefix + "/removedPersistentVolumeClaims"
+	Prefix                              = "harvesterhci.io"
```
w13915984028: define a new `func GetHarvesterPrfix()` to reduce the change scope?

pkg/webhook/resources/upgrade/mutator.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func addPauseAnnotations(upgrade *harvesterv1.Upgrade, patchOps types.PatchOps, nodeNames []string) types.PatchOps {
+	if upgrade.Annotations == nil {
+		patchOps = append(patchOps, `{"op": "add", "path": "/metadata/annotations", "value": {}}`)
+	}
+
+	for _, nodeName := range nodeNames {
+		patchOps = append(patchOps,
+			fmt.Sprintf(`{"op": "add", "path": "/metadata/annotations/harvesterhci.io~1%s", "value": "pause"}`, nodeName))
```
w13915984028: the annotation key has a name field length limit 63 chars, need to double check if `nodeName` can exceed this limit

pkg/controller/master/upgrade/common.go
```diff
@@ -53,6 +53,15 @@ fi
 `
 )
 
+func shouldPauseNodeUpgrade(upgrade *harvesterv1.Upgrade, nodeName string) bool {
+	key := fmt.Sprintf("%s/%s", util.Prefix, nodeName)
```
w13915984028: replace `util.Prefix` with `util.GetHarvesterPrefix()`

pkg/controller/master/upgrade/register.go
```diff
@@ -21,6 +29,43 @@ const (
...
 )
 
+type handler struct {
+	nodeCache ctlcorev1.NodeCache
+}
+
+func (h *handler) NotifyUnpausedMachinePlanSecret(_ string, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
+	if upgrade, ok := obj.(*harvesterv1.Upgrade); ok {
+		for key, val := range upgrade.Annotations {
+			if val == "pause" {
```
w13915984028: could `pause` be defined as a const?

pkg/controller/master/upgrade/register.go
```diff
@@ -21,6 +29,43 @@ const (
 	nodeControllerName       = "harvester-upgrade-node-controller"
 )
 
+type handler struct {
+	nodeCache ctlcorev1.NodeCache
+}
+
+func (h *handler) NotifyUnpausedMachinePlanSecret(_ string, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
+	if upgrade, ok := obj.(*harvesterv1.Upgrade); ok {
```
w13915984028: need to check the `upgrade` object, if it has latest upgrade mark, or it is in the `pause` status, and also has no deletionTimeStamp, to avoid working on legacy/fake objects

w13915984028 (commented):
Some questions, thanks.

pkg/controller/master/upgrade/register.go
```diff
@@ -21,6 +29,43 @@ const (
...
+		for key, val := range upgrade.Annotations {
+			if val == "pause" {
+				continue
+			}
+			nodeName, ok := extractNodeName(key)
```
w13915984028: Some questions on this function:

It looks to find the first annotation with `harvesterhci.io/node-name: unpause` and locate the target node, enqueue the related machine.

(1) If the first one on the annotations with `unpause` was not removed, it will not handle the next one;

(2) If user directly remove one annotation with `harvesterhci.io/node-name` from upgrade object, it seems not detecting this case

Seems a more robust solution is needed:

The upgrade CR books which node(s) is/are paused for now;  and here check the annotations & nodes lists, compute a dynamic `unpaused node list`, enqueue the required ones

pkg/apis/harvesterhci.io/v1beta1/upgrade.go
starbops: The condition was originally named `Paused`. Previous discussion around the change: https://github.com/harvester/harvester/pull/9512#discussion_r2532822745

pkg/controller/master/upgrade/register.go
starbops: This can be reworked along with the [above comment](https://github.com/harvester/harvester/pull/9512#discussion_r2533581288), which concerns whether the length of the node name plus the prefix might exceed 63 characters.

How about we change the annotation key to a constant, say `harvesterhci.io/node-upgrade-pause-map`, and its value is a JSON string recording the desired status of each node?

```json
{
  "node-0": "pause",
  "node-1": "pause",
  "node-3": "unpause"
}
```

WDYT?

pkg/controller/master/upgrade/secret_controller.go
w13915984028: I have following concerns:

The `degraded` is a bit scaring and confusing, the term like `AdminPaused` is more specific to this case. 

On K8s object, the condition is flexiblely extended, each condition tries to reflect one specific phase/cause. The generic condtion will increase the difficulties of troubleshooting, as people needs to check more other conditions/fields to know the root cause. thanks.

pkg/controller/master/upgrade/register.go
w13915984028: I am open to the solution.

BTW, do we have check to skip/deny user to add an already upgraded node to the `pause` map, avoid causing the controller reworks. thanks.

starbops: If the node has already passed the checkpoint, the annotation for that node is ignored, regardless of whether it's `pause` or anything else.

pkg/webhook/resources/upgrade/mutator.go
```diff
@@ -0,0 +1,143 @@
...
+		Scope:      admissionregv1.NamespacedScope,
+		APIGroup:   harvesterv1.SchemeGroupVersion.Group,
+		APIVersion: harvesterv1.SchemeGroupVersion.Version,
+		ObjectType: &harvesterv1.Upgrade{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+		},
+	}
+}
+func (m *upgradeMutator) Create(_ *types.Request, newObj runtime.Object) (types.PatchOps, error) {
```
ihcsim: Since we are watching only `create` events, susbsequent additions to the upgrade annotations will have no effects, right?

pkg/settings/settings_helper.go
```diff
@@ -168,13 +166,38 @@ type ImagePreloadOption struct {
...
+	ManualType ModeType = "manual"
+)
+
+type NodeUpgradeStrategy struct {
+	Mode *ModeType `json:"mode,omitempty"`
+
+	// PauseNodes contains the node names that should be paused before entering the pre-draining state during the node
+	// upgrade phase. It is only honored when the Mode is "manual". If the Mode is "manual" but no node names are
+	// provided, all nodes will be paused before entering the pre-draining state.
+	PauseNodes []string `json:"pauseNodes,omitempty"`
```
ihcsim: what do you think of using node labels to allow more flexibility with grouping and selections? for per node identification, user can also use the `kubernetes.io/hostname` label.

pkg/webhook/resources/setting/validator.go
```diff
@@ -1814,7 +1814,61 @@ func validateUpgradeConfigHelper(setting *v1beta1.Setting) (*settings.UpgradeCon
 	return config, nil
 }
 
-func validateUpgradeConfigFields(setting *v1beta1.Setting) error {
+func checkNodesDuplicates(nodeNames []string) error {
+	duplicates := ds.SliceFindDuplicates(nodeNames)
```
ihcsim, line 1818: why not just dedupe it with `ds.SliceDedupe`?

pkg/util/constants.go
```diff
@@ -77,6 +77,10 @@ const (
 	// airgap upgrades when uploading images.
 	AnnotationUpgradeImage = prefix + "/os-upgrade-image"
 
+	AnnotationNodeUpgradePauseMap = prefix + "/node-upgrade-pause-map"
+	NodePause                     = "pause"
+	NodeUnpause                   = "unpause"
```
ihcsim, line 82: i don't think this is being used

pkg/controller/master/upgrade/register.go
```diff
@@ -21,6 +30,57 @@ const (
...
+	if err := json.Unmarshal([]byte(value), &pauseMap); err != nil {
+		return nil, err
+	}
+
+	// Ideally, there will be only one "unpause" node left in the annotation.
+	machinePlanSecretKeys := make([]relatedresource.Key, 1)
+	for nodeName, state := range pauseMap {
+		if state == util.NodePause {
+			continue
+		}
```
ihcsim: will we find any unpaused nodes here, since we are using `upgrade.clearNodePause()` to remove them in the controller logic?

pkg/controller/master/upgrade/common.go
```diff
@@ -53,6 +54,59 @@ fi
...
 
+func getNodeUpgradePauseMap(upgrade *harvesterv1.Upgrade) (map[string]string, error) {
+	if upgrade.Annotations == nil {
+		return nil, nil
+	}
+
+	value, ok := upgrade.Annotations[util.AnnotationNodeUpgradePauseMap]
+	if !ok {
+		return nil, nil
+	}
```
ihcsim, line 65: I am wondering if it is possible to not use the annotation map to track the pause states. Can the `unpause` state be tracked in the upgrade's `status.nodeStatuses`? The secret controller should be able to handle that. My concern about storing states in the annotation map is that, AIUI, there is nothing preventing the it from being tampered. I think the annotation map should only track the list of (remaining) nodes user wants to pause.

pkg/webhook/resources/upgrade/mutator.go
ihcsim, line 50: AIUI, `Create` here means changing the pause-node setting will have no effect on an in-progress upgrade, right? It might be something worth documenting to manage user expectation. E.g., they changed their mind on which nodes they want to pause after upgrade already started.

pkg/controller/master/upgrade/secret_controller.go
ihcsim: Sorry for being unclear earlier, what i meant was create the job anyway, but in [`suspend` mode](https://kubernetes.io/docs/concepts/workloads/controllers/job/#suspending-a-job) so the pod won't run, until the node is unpaused. What you have works too - thanks.

pkg/controller/master/upgrade/common.go
starbops: The annotation map stores the desired action for specific nodes upon node upgrade. The actual state of the nodes are recorded in the `status.nodeStatuses`. The secret controller will factor in the `status.nodeStatuses` (current node state), the secret annotations, and the annotation map (desired node upgrade action). For instance, if the current node state is "image preloaded", the secret has "pre-drain" annotated, and the map says the node should be paused, the node upgrade for the node will be paused. Another example, if the current node state is "node-upgrade paused", the secret has "pre-drain" annotated, and the map says the node should be unpaused, the node upgrade for the node will be unpaused.

pkg/webhook/resources/upgrade/mutator.go
starbops: Yes, I will create a documentation update describing how the setting is applied when starting an upgrade. Any further changes to the setting can only affect the next upgrade.

pkg/controller/master/upgrade/register.go
starbops: When the user updates the Upgrade CR with the node `unpause`, this should be triggered. It will enqueue the corresponding machine-plan secret (the one that is unpaused), and then the secret controller will unpause the node upgrade. Finally, the `unpause` entry is removed by `clearNodePause()`. If we don't do this, the machine-plan secret might no longer be reconciled.

It's fine not calling the `clearNodePause()` function. The drawback is that every unpause node is enqueued whenever the Upgrade CR is updated, even though many of them are unnecessary.

pkg/webhook/resources/setting/validator.go
starbops: The returned error message clearly shows what is duplicated, which provides a better user experience.

pkg/controller/master/upgrade/register.go
starbops: After testing, your concern is right. I should've re-enqueued the secret object; otherwise, the upgrade couldn't proceed.

Update: fixed in caa5e93

pkg/controller/master/upgrade/secret_controller.go
starbops: @ihcsim sorry for not getting the idea at first place. That works as well :)

@w13915984028 In fact, we can drop the `Degraded` condition entirely. I added it because UI needs the information to know what node to unpause. Now that it's of no use and more of an informational item to the user.

pkg/webhook/util/node.go
```diff
@@ -0,0 +1,14 @@
+package util
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewNodes(nodeNames ...string) []*corev1.Node {
```
w13915984028: This function looks to be a helper of the test code, it returns a list of partially/specially initialized object, it might not be reusable for other test code,

would suggest to put it on the test code as a private function.

put it under util will have the un-used function compiled on controller binary

pkg/webhook/resources/upgrade/mutator_test.go
```diff
@@ -0,0 +1,161 @@
...
+			},
+			expected: output{
+				patchOps: types.PatchOps{
+					`{"op": "add", "path": "/metadata/annotations", "value": {}}`,
+					`{"op": "add", "path": "/metadata/annotations/harvesterhci.io~1node-upgrade-pause-map", "value": "{\"node-0\":\"pause\",\"node-1\":\"pause\",\"node-2\":\"pause\"}"}`,
+				},
+			},
+		},
+		{
+			name: "manual node upgrade mode with pause nodes specified explicitly",
```
w13915984028, line 112: please add some additional cases:

1. the setting has a node, but later the node is removed (setting is not updated by controller automatically), the upgrade object should have zero/nil patch ops

2. the fetched data from setting is failed to (un)marshal
https://github.com/harvester/harvester/pull/9512/files#diff-9ffe98c7fcbfd31871cff0c565ba9b2c92686eded58089f855892feb867d827fR134

pkg/webhook/resources/upgrade/mutator.go
```diff
@@ -0,0 +1,143 @@
...
+
+	return settings.DecodeConfig[settings.UpgradeConfig](value)
+}
+
+func (m *upgradeMutator) getNodesToPause(config *settings.UpgradeConfig) ([]string, error) {
+	if *config.NodeUpgradeOption.Strategy.Mode != settings.ManualType {
+		return nil, nil
+	}
+
+	// Use explicitly configured pause nodes if provided
```
w13915984028: note: as we don't have `node onRemove` controller to handle the  node removing case, the setting could be stale, the saved values   have some non-existing nodes

seems we can still filter it again via following node list

pkg/webhook/resources/upgrade/mutator.go
```diff
@@ -0,0 +1,143 @@
...
+		config.NodeUpgradeOption.Strategy.Mode != nil &&
+		*config.NodeUpgradeOption.Strategy.Mode != settings.AutoType
+}
+
+func addPauseAnnotations(upgrade *harvesterv1.Upgrade, patchOps types.PatchOps, nodeNames []string) types.PatchOps {
+	if upgrade.Annotations == nil {
+		patchOps = append(patchOps, `{"op": "add", "path": "/metadata/annotations", "value": {}}`)
+	}
+
+	nodeMap := make(map[string]string, len(nodeNames))
```
w13915984028, line 135: add below before L125, as the passed nodeNames could have zero valid elements
```
if len(nodeNames) == 0 {
	return nil
}
```

pkg/webhook/resources/upgrade/mutator.go
```diff
@@ -0,0 +1,143 @@
...
+	for _, nodeName := range nodeNames {
+		nodeMap[nodeName] = util.NodePause
+	}
+
+	nodeMapJSON, err := json.Marshal(nodeMap)
+	if err != nil {
+		logrus.Errorf("failed to marshal node map for upgrade %s/%s: %v", upgrade.Namespace, upgrade.Name, err)
```
w13915984028: "failed to marshal node map for upgrade %s/%s: %v"
->
"failed to marshal node map for upgrade %s/%s: %v, the patch is skipped"

w13915984028 (commented):
Some additional questions, thanks.

pkg/controller/master/upgrade/common.go
```diff
@@ -53,6 +54,59 @@ fi
...
+
+	var pauseMap map[string]string
+	if err := json.Unmarshal([]byte(value), &pauseMap); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal node upgrade pause map: %w", err)
```
w13915984028, line 69: as we allow user to change this map on the runtime to add/remove nodes to/from the map

seems we need to add webhook to check this map, to avoid the `Unmarshal` failure case, otherwise, the controller might stuck

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -397,6 +397,16 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 		toUpdate := upgrade.DeepCopy()
 		singleNodeName := upgrade.Status.SingleNode
 		if singleNodeName != "" {
+			if shouldPauseNodeUpgrade(upgrade, singleNodeName) {
```
w13915984028, line 410: when user does not enable `pause` ability at all, it will still have log like 
`Unpause pre-drain job ...`

It seems like the `upgrade` object needs a `AdminPaused` spec, and it is set on `Upgrade create` mutator, and doesn't allow change in-between the upgrade

And the pause/unpause controller feature should check this spec first

w13915984028 (commented):
Additional question about the `AdminPaused` switch,

as described in the HEP, the enable/disable `pause` is copied from the setting when upgrade is created, and if setting is changed in between the upgrade, it is not followed;  

meanwhile, the `upgrade` object does not book the `automaic/manual`, and user seems to be able to set the pause map to pause the upgrade, is this intended? thanks.

pkg/webhook/resources/upgrade/mutator.go
starbops: Nice catch!

pkg/controller/master/upgrade/upgrade_controller.go
starbops: This piece of code is for the single-node cluster upgrade only. We can change it to something similar to the secret controller's 2-step handling, without introducing additional spec or condition.

pkg/webhook/resources/upgrade/mutator.go
starbops: I'll add a filter to return the intersection of the existing nodes and pause nodes.

pkg/controller/master/upgrade/secret_controller.go
```diff
@@ -75,9 +79,18 @@ func (h *secretHandler) OnChanged(_ string, secret *v1.Secret) (*v1.Secret, erro
...
 	}
 
+	upgradeCpy := upgrade.DeepCopy()
+
 	switch upgrade.Status.NodeStatuses[nodeName].State {
 	case nodeStateImagesPreloaded:
 		if secret.Annotations[rke2PreDrainAnnotation] != secret.Annotations[preDrainAnnotation] {
+			if shouldPauseNodeUpgrade(upgrade, nodeName) {
+				logrus.Infof("Pause pre-drain job creation for node %s", nodeName)
+				setNodeUpgradeStatus(upgradeCpy, nodeName, nodeStateUpgradePaused, "AdministrativelyPaused", "Node upgrade paused as requested by the user")
```
w13915984028: skip this comment, not a valid review point

pkg/controller/master/upgrade/secret_controller.go
```diff
@@ -75,9 +79,18 @@ func (h *secretHandler) OnChanged(_ string, secret *v1.Secret) (*v1.Secret, erro
 		return secret, nil
 	}
 
+	upgradeCpy := upgrade.DeepCopy()
+
 	switch upgrade.Status.NodeStatuses[nodeName].State {
 	case nodeStateImagesPreloaded:
 		if secret.Annotations[rke2PreDrainAnnotation] != secret.Annotations[preDrainAnnotation] {
```
w13915984028: encap 
`secret.Annotations[rke2PreDrainAnnotation] != secret.Annotations[preDrainAnnotation]` to a local func for resuing on L103

pkg/controller/master/upgrade/common.go
```diff
@@ -53,6 +54,59 @@ fi
...
+	pauseMap, err := getNodeUpgradePauseMap(upgrade)
+	if err != nil || pauseMap == nil {
+		return false
+	}
+
+	return pauseMap[nodeName] == util.NodePause
+}
+
+// clearNodePause removes the specified node from the node upgrade pause map annotation.
+func clearNodePause(upgrade *harvesterv1.Upgrade, nodeName string) error {
```
w13915984028: this func, is a bit different with `shouldPauseNodeUpgrade`, 

`shouldPauseNodeUpgrade`: only cares the target node has `NodePause` value

`clearNodePause`  directly removes the related field, it can also set the value to `Unpause` for kind of future log

The controller works per the condition `shouldPauseNodeUpgrade`, the `clearNodePause` (from the map) seems not necessary, just updates the object status and leave the pause map for reference (which node(s) had been ever paused on the upgrade)

pkg/controller/master/upgrade/secret_controller.go
```diff
@@ -86,6 +99,27 @@ func (h *secretHandler) OnChanged(_ string, secret *v1.Secret) (*v1.Secret, erro
...
+			if shouldPauseNodeUpgrade(upgrade, nodeName) {
+				logrus.Debugf("Continue pausing pre-drain job creation for node %s", nodeName)
+				return secret, nil
+			}
+			logrus.Infof("Unpause pre-drain job creation for node %s", nodeName)
+			if err := clearNodePause(upgradeCpy, nodeName); err != nil {
+				return secret, err
+			}
+			setNodeUpgradeStatus(upgradeCpy, nodeName, nodeStateImagesPreloaded, "", "")
+			if reflect.DeepEqual(upgradeCpy, upgrade) {
```
w13915984028: at this case, the deepequal seems redundant, as in the `switch case nodeStateUpgradePaused, the status is set to a new value nodeStateImagesPreloaded`

w13915984028 (dismissed):
LGTM, thanks.

pkg/controller/master/upgrade/common.go
starbops: That's a good idea. So the controller does not alter the pause map and keeps it as a user-defined intention for the desired node upgrade action on specific nodes. I'll adjust the `NotifyUnpausedMachinePlanSecret` function to handle the change so that whenever the pause map annotation is updated, only the machine-plan secrets whose corresponding nodes happen to be in the paused state and requested to be unpaused are notified.

pkg/controller/master/upgrade/register.go
```diff
@@ -21,6 +30,66 @@ const (
...
+	upgrade, ok := obj.(*harvesterv1.Upgrade)
+	if !ok {
+		return nil, nil
+	}
+
+	if upgrade.DeletionTimestamp != nil || upgrade.Labels[harvesterLatestUpgradeLabel] != "true" {
+		return nil, nil
+	}
+
+	value, ok := upgrade.Annotations[util.AnnotationNodeUpgradePauseMap]
```
w13915984028: are L47~L55 same/similar with getNodeUpgradePauseMap?

w13915984028 (dismissed):
LGTM, thanks.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -397,6 +397,22 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+				logrus.Infof("Unpause node-upgrade job creation for node %s", singleNodeName)
+				setNodeUpgradeStatus(toUpdate, singleNodeName, nodeStateImagesPreloaded, "", "")
+				return h.upgradeClient.Update(toUpdate)
+			}
+
+			if shouldPauseNodeUpgrade(upgrade, singleNodeName) {
+				logrus.Infof("Pause node-upgrade job creation for node %s", singleNodeName)
+				setNodeUpgradeStatus(toUpdate, singleNodeName, nodeStateUpgradePaused, "AdministrativelyPaused", "Node upgrade paused as requested by the user")
+				return h.upgradeClient.Update(toUpdate)
+			}
```
ihcsim, line 414: Since the annotation map is the source of truth, can the code be made simpler by checking the annotation map first, instead of the node status? E.g.,

```diff
diff --git a/pkg/controller/master/upgrade/upgrade_controller.go b/pkg/controller/master/upgrade/upgrade_controller.go
index fca81db6f..0197f8493 100644
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -397,20 +397,20 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
 		toUpdate := upgrade.DeepCopy()
 		singleNodeName := upgrade.Status.SingleNode
 		if singleNodeName != "" {
-			if upgrade.Status.NodeStatuses[singleNodeName].State == nodeStateUpgradePaused {
-				if shouldPauseNodeUpgrade(upgrade, singleNodeName) {
-					logrus.Debugf("Continue pausing node-upgrade job creation for node %s", singleNodeName)
-					return upgrade, nil
-				}
-				logrus.Infof("Unpause node-upgrade job creation for node %s", singleNodeName)
-				setNodeUpgradeStatus(toUpdate, singleNodeName, nodeStateImagesPreloaded, "", "")
-				return h.upgradeClient.Update(toUpdate)
-			}
-
 			if shouldPauseNodeUpgrade(upgrade, singleNodeName) {
-				logrus.Infof("Pause node-upgrade job creation for node %s", singleNodeName)
-				setNodeUpgradeStatus(toUpdate, singleNodeName, nodeStateUpgradePaused, "AdministrativelyPaused", "Node upgrade paused as requested by the user")
-				return h.upgradeClient.Update(toUpdate)
+				if upgrade.Status.NodeStatuses[singleNodeName].State != nodeStateUpgradePaused {
+					logrus.Infof("Pause node-upgrade job creation for node %s", singleNodeName)
+					setNodeUpgradeStatus(toUpdate, singleNodeName, nodeStateUpgradePaused, "AdministrativelyPaused", "Node upgrade paused as requested by the user")
+					return h.upgradeClient.Update(toUpdate)
+				}
+				logrus.Debugf("Continue pausing node-upgrade job creation for node %s", singleNodeName)
+				return upgrade, nil
+			} else {
+				if upgrade.Status.NodeStatuses[singleNodeName].State == nodeStateUpgradePaused {
+					logrus.Infof("Unpause node-upgrade job creation for node %s", singleNodeName)
+					setNodeUpgradeStatus(toUpdate, singleNodeName, nodeStateImagesPreloaded, "", "")
+					return h.upgradeClient.Update(toUpdate)
+				}
 			}
 
 			logrus.Info("Start single node upgrade job")
@@ -625,7 +625,6 @@ func (h *upgradeHandler) isSingleNodeCluster() (string, error) {
 	}
 	if len(nodes) == 1 {
 		return nodes[0].Name, nil
-
 	}
 	return "", nil
 }
```

starbops: I don't see any difference between the two in terms of LoC and logic paths. The original style aligns with the counterpart in the secret controller (state dominant) and is more idiomatic in Go. Could you elaborate on the suggested change? Thank you.

pkg/controller/master/upgrade/secret_controller.go
brandboat: ~~gentle ping @starbops, could you please fix this? Thank you!~~ already addressed

w13915984028 (dismissed):
LGTM, thanks.

brandboat (dismissed):
LGTM, thanks for the great feature! I tested the manual mode on both a single-node setup and a two-node cluster in my local environment, and everything looks good.

pkg/controller/master/upgrade/secret_controller.go
```diff
@@ -75,17 +78,43 @@ func (h *secretHandler) OnChanged(_ string, secret *v1.Secret) (*v1.Secret, erro
...
 	case nodeStateImagesPreloaded:
-		if secret.Annotations[rke2PreDrainAnnotation] != secret.Annotations[preDrainAnnotation] {
-			if err := checkEligibleToDrain(upgrade, nodeName); err != nil {
-				return nil, err
-			}
-			logrus.Debugf("Create pre-drain job on %s", nodeName)
-			if err := h.createHookJob(upgrade, nodeName, upgradeJobTypePreDrain, nodeStatePreDraining); err != nil {
-				return nil, err
-			}
+		if isPreDrained(secret) {
```
Vicente-Cheng: Looks like the pause will be stuck before the pre-drain job creation.
The isPreDrained checks if the job is finished or if the job exists?

From the code level, it should return true once we are in the pre-drain stage instead of the pre-drained. So I would like to confirm that.

pkg/controller/master/upgrade/common.go
ihcsim: Per slack discussion, every time the user clicks on the pause/unpause button, a reconciliation loop is triggered in the upgrade controller. Once the UI is ready, we need to test what happens if the user clicks through a series of pause/unpause/pause actions on the same node.

pkg/controller/master/upgrade/upgrade_controller.go
ihcsim: It's a TIOLI optimization to reduce duplicate code with the `shouldPauseNodeUpgrade()` 😄. Yeah, I did notice it's trying to follow the same pattern as the `switch` in the secret controller (but I don't think it needs to be). It's fine.

pkg/controller/master/upgrade/common.go
starbops: There won't be **pause** buttons in the upgrade dialog during the upgrade; only **unpause** buttons will appear. In UI's workflow, for this feature specifically, the underlying pause map is determined when the "Upgrade" button is clicked. If we have time and capacity to work on the "pause nodes selection" part on the UI, the user will have a chance to specify what nodes should be paused for node upgrades right before they click the "Upgrade" button, which only applies to the current upgrade activity.

> Once the UI is ready, we need to test what happens if the user clicks through a series of pause/unpause/pause actions on the same node.

UI aside, the behavior resembles directly modifying the pause map annotation on the Upgrade object using `kubectl`. This is the core part of the feature implementation, and it is appropriately handled. Or do you have concerns regarding the watcher's ripple effect?

Vicente-Cheng (dismissed):
just a nit question, overall lgtm

pkg/controller/master/upgrade/secret_controller.go
starbops: Actually, the function only checks the machine-plan secret's annotations. In reality, there will be three possible combinations:

1. Both the `pre-drain` and `pre-hook` annotations are missing; this implies the node is not under upgrade.
2. Only the `pre-drain` annotation exists; this means Rancher notifies us (Harvester) to initiate the pre-drain job, and it is waiting for us to annotate the `pre-hook` annotation once the job is done.
3. Both the `pre-drain` and `pre-hook` annotations exist and have the same value; this represents that we have already notified Rancher to continue with the drain, which implies the pre-drain job has been completed.

The secret controller does not care whether the pre-/post-drain job is running or finished. It only focuses on the machine-plan secrets.

So, according to the above cases, the `isPreDrained` function should return:

1. `true`
2. `false`
3. `true`

The function name can be misleading because in case 1, it's actually "not-yet-pre-drained". Only cases 2 & 3 fit in the context of the function name. We can rename it to something like `isUnderPreDrain` and negate the condition.

w13915984028 (approved):
LGTM, thanks.

brandboat (approved):

Vicente-Cheng (approved):

Vicente-Cheng:
Hi @ihcsim, 
I know there are still some discussions about where we should put the annotation. 
But consider the v1.7.0 schedule, I prefer to let everything work here.

We can discuss it at the upgrade v2 review meeting.
Thanks!

pkg/controller/master/upgrade/secret_controller.go
Vicente-Cheng: Thanks for the detailed explanation!

starbops:
I've created task #9575 to track the unsettled discussions so we can work on them in upgrade v2. Appreciate all the reviewers!

PR #9518, opened 2025-11-14
Check if a node is witnessnode without promotion status

rrajendran17:
#### Problem:
The existing logic in harvester code returns true for a node as witness node only if both witness label ( node-role.harvesterhci.io/witness) and promote-status is set.
Ref: https://github.com/harvester/harvester/blob/bdd8042958e2bb7b09b3524b5d4a2b849a724b33/pkg/util/node.go#L44

But this is not the case always.A node can be created with just witness label set to true

#### Solution:
Introduce a new function to return true for a node as witness node only if witness node label is set to true

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9325

#### Test plan:
case 1:
1.Have a harvester cluster 2 nodes (1 witness node and 1 master node)
2.Make sure to have only instance manager pods running and no backing image pods.
3.Create storage network
4.Check if "IP allocation failure for Longhorn Pods" msg not seen in Harvester UI under storage network settings and storage network creation and allocation of secondary IP to instance manager pod is successful.

case 2:
1.Have a harvester cluster 2 nodes ( 1 witness node and 1 master node)
2.Create a cluster network cluster-1
3.Create a vlan config vlancluster-1 spanning all nodes
4.3.Create storage network
4.Check if there are no webhook errors seen and storage network creation and allocation of secondary IP to instance manager pod is successful

### DIFF
--- a/pkg/util/node.go
+++ b/pkg/util/node.go
@@ -41,6 +41,15 @@ func IsPromoteStatusIn(node *corev1.Node, statuses ...string) bool {
 	return false
 }
 
+func IsWitnessNodeWithoutPromotionStatus(node *corev1.Node) bool {
+	val, found := node.Labels[HarvesterWitnessNodeLabelKey]
+	if found && val == "true" {
+		return true
+	}
+
+	return false
+}
+
 func IsWitnessNode(node *corev1.Node, isManagement bool) bool {
 	_, found := node.Labels[HarvesterWitnessNodeLabelKey]
 	if !found {
@@ -80,7 +89,7 @@ func CountNonWitnessNodes(nodes []*corev1.Node) int {
 	count := 0
 
 	for _, node := range nodes {
-		if !IsWitnessNode(node, IsManagementRole(node)) {
+		if !IsWitnessNodeWithoutPromotionStatus(node) {
 			count++
 		}
 	}
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -1499,8 +1499,7 @@ func (v *settingValidator) checkVCSpansAllNodes(config *networkutil.Config) erro
 	//check if vlanconfig contains all the nodes in the cluster
 	for _, node := range nodes {
 		//skip witness nodes which do not run LH Pods
-		isManagement := util.IsManagementRole(node)
-		if util.IsWitnessNode(node, isManagement) {
+		if util.IsWitnessNodeWithoutPromotionStatus(node) {
 			continue
 		}
 


pkg/util/node.go
```diff
@@ -80,7 +89,7 @@ func CountNonWitnessNodes(nodes []*corev1.Node) int {
 	count := 0
 
 	for _, node := range nodes {
-		if !IsWitnessNode(node, IsManagementRole(node)) {
+		if !IsWitnessNodeWithoutPromotionStatus(node) {
```
WebberHuang1118, line 92: What if the witness node is under promotion? Would that affect the counting?

pkg/util/node.go
```diff
@@ -41,6 +41,15 @@ func IsPromoteStatusIn(node *corev1.Node, statuses ...string) bool {
 	return false
 }
 
+func IsWitnessNodeWithoutPromotionStatus(node *corev1.Node) bool {
+	val, found := node.Labels[HarvesterWitnessNodeLabelKey]
+	if found && val == "true" {
+		return true
+	}
+
+	return false
```
WebberHuang1118, line 50: The functions `IsWitnessNodeWithoutPromotionStatus` and `IsWitnessNode` seem quite similar. Is `IsWitnessNode` still needed, or is it no longer correct?

WebberHuang1118 (commented):
Please take a look at the comments. Thanks.

pkg/util/node.go
rrajendran17: Good question.My understanding is, a witness node is never eligible for promotion, so a storage network configuration on a witness node under promotion does not have to strictly validate for promote-status as promote-status is more applicable in the context of promotion controller who is comparing these promote status annotations.

pkg/util/node.go
rrajendran17: My understanding is, promotion controller needs to validate if promote-status is set in case of tracking the promotion statuses on a node whatever the role it may be which is not strictly required during storage-network configurations.

And I am not very clear on the logic of IsWitnessNode() on why it requires a check for isManagement.My assumption is that witness node was expected to have etcd role set earlier , but that is not true in recent cases. @Vicente-Cheng  will have more insights on this.

rrajendran17:
Overall I feel we might need two different functions for checking witness node, one for promotion controller and another one for other parts of the harvester code and also since etcd label is no more updated to node-role annotations of the witness, we need to also reconsider having isManagement field in the IsWitnessRole() function used by promotion controller.But if that is removed then it must be carefully validated for all promotion scenarios.
cc: @Vicente-Cheng

pkg/util/node.go
Vicente-Cheng: If we don't want to count the witness node, no matter whether it promotes or not, the annotation `HarvesterWitnessNodeLabelKey` was there.

So, this might not be the issue.

pkg/util/node.go
Vicente-Cheng: I checked the code.
The isWitnessNode is like @rrajendran17 mentioned,
`only if both witness label ( node-role.harvesterhci.io/witness) and promote-status is set.`

So, this function `IsWitnessNodeWithoutPromotionStatus` can simply check the witness node (even if it was not promoted).

Vicente-Cheng (approved):
lgtm, thanks

WebberHuang1118 (approved):
LGTM, thanks.

Vicente-Cheng:
@mergifyio backport v1.7

Issue #9509, opened 2025-11-14
[BUG] enabling multipathd fails on v1.7.0-rc2


ibrokethecloud:
### Describe the Bug

Attempt to start multipathd on v1.7.0-rc2 fails with the following error

```
× multipathd.service - Device-Mapper Multipath Device Controller
     Loaded: loaded (/etc/systemd/system/multipathd.service; disabled; preset: disabled)
     Active: failed (Result: exit-code) since Thu 2025-11-13 22:36:59 UTC; 1s ago
TriggeredBy: ○ multipathd.socket
    Process: 945832 ExecStart=/sbin/multipathd -d -s (code=exited, status=1/FAILURE)
   Main PID: 945832 (code=exited, status=1/FAILURE)
     Status: "configure"
        CPU: 42ms

Nov 13 22:36:59 harvester-node-1 systemd[1]: Starting Device-Mapper Multipath Device Controller...
Nov 13 22:36:59 harvester-node-1 multipathd[945832]: multipathd v0.10.0+103+suse.0fc97cd: start up
Nov 13 22:36:59 harvester-node-1 multipathd[945832]: DM multipath kernel driver not loaded
Nov 13 22:36:59 harvester-node-1 multipathd[945832]: reconfigure: setting up paths and maps
Nov 13 22:36:59 harvester-node-1 systemd[1]: multipathd.service: Main process exited, code=exited, status=1/FAILURE
Nov 13 22:36:59 harvester-node-1 systemd[1]: multipathd.service: Failed with result 'exit-code'.
Nov 13 22:37:00 harvester-node-1 systemd[1]: Failed to start Device-Mapper Multipath Device Controller.
```

This is caused by the a combination of SLE Micro bump and custom multipath service written by the installer

### To Reproduce

* Install harvester v1.7.0-rc2
* ssh to node once install is complete and run `systemctl enable multipathd` and `systemctl start multipathd`
* service will fail to start

### Expected Behavior

multipath should work correctly

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.7.0-rc2
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): N/A
- Rancher version: N/A

### Workaround and Mitigation

run `modprobe dm_multipath` before enabling the service

ibrokethecloud:
To test

## New install
* Install harvester v1.7-head or v1.7.0-rc*
[...]

PR #9521, opened 2025-11-17
updated multipathd service definition as part of upgrade

ibrokethecloud:
#### Problem:

multipathd does fails to enable after upgrade to v1.7-head or rc.

This is caused as `dm_multipath` kernel module is not loaded.

#### Solution:

Update the definition of `multipathd.service` as part of the upgrade to load the `dm_multipath` kernel module.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9509

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -668,8 +668,10 @@ EOF
       thirdPartyArgs=$(echo ${thirdPartyArgs} | xargs)
       chroot $HOST_DIR grub2-editenv /oem/grubenv set third_party_kernel_args="${thirdPartyArgs}"
     fi
-    # add cloud-init directive to disable multipathing for longhorn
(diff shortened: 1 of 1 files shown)


tserong (dismissed):
File content change LGTM.

One other issue though which I realised while reviewing this, which I'd presumably somehow missed when this was first implemented: I suspect this file creation should possibly be moved outside the enclosing `if [ ${multiPathEnabled} == false ]` block, so that it happens all the time, regardless of whether external storage is enabled, because that's what the installer code does (`disableLonghornMultipathing()` appears to be always called here https://github.com/harvester/harvester-installer/blob/947d134c51158e57c70edd027089b8fc664d0b5e/pkg/config/cos.go#L185 regardless of the externalstorage enabled setting).

tserong (approved):

Vicente-Cheng (approved):
lgtm, thanks!

Vicente-Cheng:
@mergifyio backport v1.7

PR #9522, opened 2025-11-17
feat: use config map to track cpu model

Yu-Jack:
#### Solution:
Create a config map to collect all cpu model information. This can ensure cache. This design also makes sense in controller pattern.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/3015

#### Test plan:

<details>

<summary> kubectl get configmap node-cpu-model-configuration -n harvester-system -o yaml </summary>

```yaml
apiVersion: v1
data:
  cpuModels: |
    checksum: 6b8dd5cb2aa0f103be2737120d7bf781f9bb4975200f21cb460fbfe27e61053d
    globalModels:
    - host-model
    - host-passthrough
    models:
      Denverton-v2:
        migrationSafe: false
        readyCount: 1
      Denverton-v3:
        migrationSafe: false
        readyCount: 1
      IvyBridge:
        migrationSafe: false
        readyCount: 1
      IvyBridge-IBRS:
        migrationSafe: false
        readyCount: 1
      IvyBridge-v1:
        migrationSafe: false
        readyCount: 1
      IvyBridge-v2:
        migrationSafe: false
        readyCount: 1
      Nehalem:
        migrationSafe: false
        readyCount: 1
      Nehalem-IBRS:
        migrationSafe: false
        readyCount: 1
      Nehalem-v1:
        migrationSafe: false
        readyCount: 1
      Nehalem-v2:
        migrationSafe: false
        readyCount: 1
      Penryn:
        migrationSafe: false
        readyCount: 1
      Penryn-v1:
        migrationSafe: false
        readyCount: 1
      SandyBridge:
        migrationSafe: false
        readyCount: 1
      SandyBridge-IBRS:
        migrationSafe: false
        readyCount: 1
      SandyBridge-v1:
        migrationSafe: false
        readyCount: 1
      SandyBridge-v2:
        migrationSafe: false
        readyCount: 1
      Westmere:
        migrationSafe: false
        readyCount: 1
      Westmere-IBRS:
        migrationSafe: false
        readyCount: 1
      Westmere-v1:
        migrationSafe: false
        readyCount: 1
      Westmere-v2:
        migrationSafe: false
        readyCount: 1
    totalNodes: 1
    updatedAt: "2026-01-28T07:14:11Z"
kind: ConfigMap
metadata:
  name: node-cpu-model-configuration
  namespace: harvester-system
```

</details>

HEP: https://github.com/harvester/harvester/pull/7937
UI: https://github.com/harvester/harvester-ui-extension/pull/702
Guest Cluster: https://github.com/harvester/harvester/issues/9744


ibrokethecloud (commented):
we are calculating cluster wide cpu models, would it not make sense to have this be available in the cluster object? we already do something similar for vGPU profiles available across the cluster

Yu-Jack:
> we are calculating cluster wide cpu models, would it not make sense to have this be available in the cluster object? we already do something similar for vGPU profiles available across the cluster

I think it's a good idea. We just need to create a new CRD, then modify the status when a node joins or is removed. The frontend just needs to fetch this resource. Also, this can ensure cache. Besides, this design makes more sense here because we're running k8s controller.

w13915984028:
cc @martindekov FYI.

Yu-Jack:
@ibrokethecloud @ihcsim Just a gentle ping for this. It would be better to have your feedback. Thanks!

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachinecpumodels.yaml
```diff
@@ -0,0 +1,76 @@
...
+              Servers may infer this from the endpoint the client submits requests to.
+              Cannot be updated.
+              In CamelCase.
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+            type: string
+          metadata:
+            type: object
+          spec:
+            type: object
+          status:
```
ihcsim: I don't think it's necessary to always have `.spec` and `.status`, especially, if `.spec` is going to be empty. If this is going to serve as a data structure, we can just use `.data`.

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachinecpumodels.yaml
```diff
@@ -0,0 +1,76 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    {}
+  name: virtualmachinecpumodels.harvesterhci.io
```
ihcsim: I think the CPU model is more like a node feature. Can we consider naming it something like `NodeCPUModels` or `NodeCPUModelConfiguration`? I am using plural because the resource will have more than one models.

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachinecpumodels.yaml
```diff
@@ -0,0 +1,76 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
ihcsim: What's the trade-off between using a CRD and a configmap?

pkg/controller/master/virtualmachinecpumodel/controller.go
```diff
@@ -0,0 +1,206 @@
...
+
+	h := &Handler{
+		nodeCache:            nodes.Cache(),
+		kubevirtCache:        kubevirt.Cache(),
+		vmCpuModelClient:     vmCpuModel,
+		vmCpuModelController: vmCpuModel,
+		vmCpuModelCache:      vmCpuModel.Cache(),
+	}
+
+	if err := h.createSingleton(); err != nil {
```
ihcsim: Can the resource be included in the Helm chart? Then we don't need additional logic here to create it.

pkg/controller/master/virtualmachinecpumodel/controller.go
```diff
@@ -0,0 +1,206 @@
...
+	toUpdate.Status = status
+	return h.vmCpuModelClient.Update(toUpdate)
+}
+
+func collectNodeModels(node *corev1.Node) map[string]struct{} {
+	models := make(map[string]struct{})
+	for labelKey, labelValue := range node.Labels {
+		if labelValue != "true" {
+			continue
+		}
```
ihcsim: I don't think this is necessary as it won't speed things up by much. Also, we have no idea what value upstream might use in the future.

pkg/controller/master/virtualmachinecpumodel/controller.go
```diff
@@ -0,0 +1,206 @@
...
+
+	for _, node := range nodes {
+		if !isNodeReady(node) {
+			continue
+		}
+
+		nodeModels := collectNodeModels(node)
+		for model := range nodeModels {
+			modelCounts[model]++
+		}
```
ihcsim: Nit: an alternative is to pass `modelCounts` into `collectNodeModels(node)` and update the counts in the map within the function, to avoid copying an extra map for every node iteration.

E.g.,

```diff
diff --git a/pkg/controller/master/virtualmachinecpumodel/controller.go b/pkg/controller/master/virtualmachinecpumodel/controller.go
index 519e5014d..312459332 100644
--- a/pkg/controller/master/virtualmachinecpumodel/controller.go
+++ b/pkg/controller/master/virtualmachinecpumodel/controller.go
@@ -122,11 +122,7 @@ func (h *Handler) reconcile(obj *harvesterv1.VirtualMachineCPUModel) (*harvester
 		if !isNodeReady(node) {
 			continue
 		}
-
-		nodeModels := collectNodeModels(node)
-		for model := range nodeModels {
-			modelCounts[model]++
-		}
+		countNodeCPUModels(modelCounts, node)
 	}
 
 	globalModels, err := h.collectGlobalModels()
@@ -157,8 +153,7 @@ func (h *Handler) reconcile(obj *harvesterv1.VirtualMachineCPUModel) (*harvester
 	return h.vmCpuModelClient.Update(toUpdate)
 }
 
-func collectNodeModels(node *corev1.Node) map[string]struct{} {
-	models := make(map[string]struct{})
+func countNodeCPUModels(modelCounts map[string]int, node *corev1.Node) {
 	for labelKey, labelValue := range node.Labels {
 		if labelValue != "true" {
 			continue
@@ -171,10 +166,9 @@ func collectNodeModels(node *corev1.Node) map[string]struct{} {
 			if model == "" {
 				continue
 			}
-			models[model] = struct{}{}
+			modelCounts[model]++
 		}
 	}
-	return models
 }
 
 func (h *Handler) collectGlobalModels() ([]string, error) {
```

deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachinecpumodels.yaml
Yu-Jack: I think it's too much with a CRD, configmap is more suitable and easy for this. 

Usually, there is only one resource to present the node cpu models. I think It's not necessary to create a CRD for this case actually. 

In my original idea, this resource shouldn't  be changed by anyone, it should be generated by the controller when a node joins or leaves. That's why I put it into `.status`.

With configmap, we can use Immutable configmap. So the flow will become:

- When Harvester controller starts
- Create a immutable configmap based on all nodes
- If a node joins or leaves
- Delete the original configmap and re-create it again

However, with this, we might not be able to put the resource in the Helm chart. WDYT?

ihcsim: I'm ok with using a configmap. I don't think we need to make it immutable, as it will lead to more delete/re-create churns during cluster init and setup.

With a configmap, there is also the opportunity for the pod to mount it as file. If the dashboard can access the pod's file system, it will save us tons of API calls. Kubelet automatically propagates any data change to the file within the pod, after a brief sync period. (The sync isn't instantaneous.)

Finally, a configmap can hold about 1MiB of data and a CRD can hold about 1.5MiB. Either one should be more than enough. We can think about whether we want to handle this scenario in this PR or in another enhancement ticket.

Yu-Jack:
@ihcsim I changed it to use config map and fixed the codes based on the feedback.

Yu-Jack:
I'm considering removing the global model like host-model and host-passthrough because we can't guarantee live migration must work with those model. If users don't select any models, we use default as current behavior.

Besides, do we really need to show the option from kubevirt configuration?

```yaml
spec:
  configuration:
    cpuModel: IvyBridge
```

Because if you set up like the above, the virtual machine will boot with that model as default. It seems unnecessary for users to select it.

So, my thought is that we just need to show users the values from `models` field if we have the below example. Then we can remove the globalModels filed.

```yaml
cpuModels: |
  globalModels:
  - host-model
  - host-passthrough
  - IvyBridge
  models:
    IvyBridge-IBRS:
      migrationSafe: true
      readyCount: 2
    IvyBridge:
      migrationSafe: true
      readyCount: 2
```

ibrokethecloud:
we need to move the configmap setup out of the helm chart. the config will be part of the managedchart, and updates to configmap will result in the managedchart being marked as modified. this will block upgrades.

```
 partitions:
  - count: 1
    maxUnavailable: 1
    name: All
    summary:
      desiredReady: 1
      modified: 1
      nonReadyResources:
      - bundleState: Modified
        modifiedStatus:
        - apiVersion: v1
          kind: ConfigMap
          name: node-cpu-model-configuration
          namespace: harvester-system
          patch: '{"data":{"cpuModels":"totalNodes: 0\nglobalModels: []\nmodels: {}\n"}}'
        name: fleet-local/local
      ready: 0
  resourcesSha256Sum: 0647554c83d2923a7b3494418612bfb3d169c7bba0dcc31da296c9588d2d0648
  summary:
    desiredReady: 1
    modified: 1
    nonReadyResources:
    - bundleState: Modified
      modifiedStatus:
      - apiVersion: v1
        kind: ConfigMap
        name: node-cpu-model-configuration
        namespace: harvester-system
        patch: '{"data":{"cpuModels":"totalNodes: 0\nglobalModels: []\nmodels: {}\n"}}'
      name: fleet-local/local
```

we should setup the configmap using `pkg/data`, similar to how other objects are being setup.

Yu-Jack:
@ibrokethecloud Done!

ihcsim:
Overall, LGTM.

> I'm considering removing the global model like host-model and host-passthrough because we can't guarantee live migration must work with those model. If users don't select any models, we use default as current behavior.
> Besides, do we really need to show the option from kubevirt configuration?

I think users may want to pick certain CPU model for reasons other than live migration. E.g., they might pick `host-passthrough` to prioritize performance over live migration capability.

I am under the impression that the goal of this feature is to provide an UI way for user:

* to discover CPU models on nodes, and 
* configure CPU model of VMs

without manipulating the YAML. 

IMO, this feature can never guarantee live migration will work. The only way to guarantee that live migration can work seamlessly and not fail due to CPU model constraints, is that the nodes are and remain homogeneous.

Hence, my current thinking is that we just present all available CPU model options to the user in the list, and let user decide what they want to do. If we are motivated, we can also present node name alongside CPU model in the UI (but feels like more work)? WDYT?

ibrokethecloud:
for large GPU passthrough, users need to use `host-passthrough` to allow EFI to calculate mmio aperture based on memory addressable size from cpu

ibrokethecloud:
PR in general looks good to me. Do we need to have an `OnRemove` handler, since this will add a finalizer on the node which may be unnecessary. We already list nodes from the cache and filter them from readiness and update the configmap, so there is nothing specific we are doing to actually update the node itself.

In addition, we should probably filter the `isNodeReady` method to return false if DeletionTimestamp is set on the node to ensure quicker filtering of cpu models.

Yu-Jack:
> for large GPU passthrough, users need to use host-passthrough to allow EFI to calculate mmio aperture based on memory addressable size from cpu
> 
> Hence, my current thinking is that we just present all available CPU model options to the user in the list, and let user decide what they want to do. If we are motivated, we can also present node name alongside CPU model in the UI (but feels like more work)? WDYT?

That makes the point. Maybe they need a specific cpu model to do something although that cpu model might not be used for live migration. I'm thinking maybe we list all cpu models with node number, and add a document on the field to teach users what it means like:

image


Although we're working on live migration issue, cpu model is also related to other issue, like EFI. @ibrokethecloud  @ihcsim  WDYT?  If so, that's UI work, and we can discuss on UI PR (https://github.com/harvester/harvester-ui-extension/pull/702).

Yu-Jack:
> PR in general looks good to me. Do we need to have an `OnRemove` handler, since this will add a finalizer on the node which may be unnecessary. We already list nodes from the cache and filter them from readiness and update the configmap, so there is nothing specific we are doing to actually update the node itself.
> 
> In addition, we should probably filter the `isNodeReady` method to return false if DeletionTimestamp is set on the node to ensure quicker filtering of cpu models.

Done!

Yu-Jack:
Also, with this current format, we can add more different cases in the future when requirement comes out. So, I think we can keep `migrationSafe` field.

```
cpuModels: |
    models:
      Denverton-v2:
        migrationSafe: true
        canDoA: true -> this one
        readyCount: 2
      Denverton-v3:
        migrationSafe: true
        canDoA: true -> this one 
        readyCount: 2
      IvyBridge:
        migrationSafe: true
        canDoA: false -> this one
        readyCount: 2
      IvyBridge-IBRS:
        canDoA: false -> this one
        readyCount: 2
```

ibrokethecloud (dismissed):
lgtm. thanks for the PR.

pkg/controller/master/node/cpu_model_config_controller.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to list nodes: %w", err)
+	}
+
+	modelCounts := make(map[string]int)
+
+	for _, node := range nodes {
+		if !isNodeReady(node) {
+			continue
+		}
```
ihcsim, line 103: I think we will need to remove the CPU models of the non-ready node from the config map. One option is to move this check into `collectNodeModels()` so that only ready nodes are accounted for. Then the config map always gets rebuilt to match reality.

Yu-Jack: I thought I've filtered out the non-ready node?

Yu-Jack: Or is this just a refactor request?

pkg/data/configmap.go
```diff
@@ -0,0 +1,33 @@
...
+		Data: map[string]string{
+			"cpuModels": `totalNodes: 0
+globalModels: []
+models: {}`,
+		},
+	}
+
+	if _, err := configMaps.Create(nodeCPUModelConfigMap); err != nil && !apierrors.IsAlreadyExists(err) {
+		return errors.Wrapf(err, "Failed to create configmap %s/%s", nodeCPUModelConfigMap.Namespace, nodeCPUModelConfigMap.Name)
+	}
```
Yu-Jack, line 30: I changed it to use creation, and skip it when the resource exists. Original `Apply` will modify the existing data.

ibrokethecloud (approved):

pkg/controller/master/node/cpu_model_config_controller.go
ihcsim: Let's say the config map has some CPU models for node N. During the next reconciliation, node N is broken and not ready. Shouldn't we remove the CPU models of node N from the config map?

ihcsim: Sorry my bad - somehow, I mistaken that the `h.nodeCache.List(labels.Everything())` happened within the `collectNodeModels()`.

ihcsim (approved):

PR #9528, opened 2025-11-18
Fix: Enforce consistent Maintenance Mode Strategy behavior (rebase of #7160)

m-ildefons:
Rebase of https://github.com/harvester/harvester/pull/7160

#### Problem:
As described in https://github.com/harvester/harvester/issues/6835, there are several problems to be considered here:

 -  Ensuring that the maintenance-mode strategy is specified
 -  Ensuring that only a valid maintenance-mode strategy can be specified
 -  Ensuring that the specified maintenance-mode strategy label is propagated all the way to all resources where it's expected
 -  Ensuring that default behavior is applied in all cases where a maintenance-mode strategy either isn't specified or isn't valid

#### Solution:

The solution to the problem consists of a modification to the node drain helper and modifications both the mutating admission webhook and the validating admission webhook for the VirtualMachine resource

## The Node Drain Helper

The node drain helper was only marking Pods of VMs which had the maintenance-mode strategy label set to Migrate for deletion (i.e. migration off of the node).
This behavior would omit all Pods which didn't have this label set at all, or the ones which had it set to an invalid value. Similarly, Pods of VMs which didn't have one of the three other maintenance-mode strategies set would not be shut down. Thereby any VM without a maintenance-mode strategy, or with an invalid maintenance-mode strategy would be left running, preventing the node from entering the maintenance mode successfully. As a result the node would remain stuck in "Cordoned" state, waiting for an operator to intervene and deal with the VMs.
The new behavior of the drain helper is to migrate any VMs off of the node, if the associated Pod either has no maintenance-mode strategy label, the maintenance-mode strategy label is invalid, or the maintenance-mode strategy is specified as Migrate. This is the documented default behavior for VMs.
All other VMs will have a different maintenance-mode strategy specified and they will thus be shut down (and later restarted according to their maintenance-mode strategy).

## Admission Webhook

The admission webhook for the VirtualMachine resource plays a crucial role in making sure that VirtualMachiness have a sane maintenance-mode strategy specified.
First, the mutating webhook ensures that the maintenance-mode strategy label is set on the VirtualMachine resource. If it isn't, it will be set to Migrate, which is the default.
Then the mutating webhook ensures that the maintenance-mode strategy label that is set on the VirtualMachine resource is propagated into the .spec.templates.metadata.labels, which is to ensure that the label is propagated throughout both the VirtualMachineInstance as well as the Pod belonging to that VirtualMachine.
The mutating webhook will not modify exising maintenance-mode strategy labels on existing VirtualMachine resources. While the node drain helper will apply the default strategy to these virtual machines, this is to ensure that the user will not be faced with inexplicably changing maintenance-mode strategy labels, if they have an invalid value in their cluster.
The webhook will however log an appropriate message, allowing the user to identify VirtualMachine resources affected by invalid values in the maintenance-mode strategy label and correct it.

The validating webhook takes the role of a safe guard, ensuring that no VirtualMachine can be created or updated with an ill-specified maintenance-mode strategy.
It will reject the creation of VirtualMachine resources without a maintenance-mode strategy label or with an invalid value in the maintenance-mode strategy label.
This ensures that the label can not accidentally be filled with bogus data and that the user is informed.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/6835

#### Test plan:
See https://github.com/harvester/harvester/pull/7160

### DIFF
--- a/pkg/controller/master/virtualmachine/vmi_controller.go
+++ b/pkg/controller/master/virtualmachine/vmi_controller.go
@@ -22,11 +22,6 @@ import (
 	"github.com/harvester/harvester/pkg/util/indexeres"
 )
 
-const (
-	VirtualMachineCreatorNodeDriver = "docker-machine-driver-harvester"
-	HarvesterLabelPrefix            = "harvesterhci.io"
-)
-
 // hostLabelsReconcileMapping defines the mapping for reconciliation of node labels to virtual machine instance annotations
 var hostLabelsReconcileMapping = []string{
 	corev1.LabelTopologyZone, corev1.LabelTopologyRegion, corev1.LabelHostname,
@@ -43,21 +38,16 @@ type VMIController struct {
 	recorder            record.EventRecorder
 }
 
-// SyncHarvesterVMILabelsToPod ensures that all Harvester labels (i.e. those
-// with the harvesterhci.io/ prefix) from the VirtualMachineInstance are synced
-// to the Pod
+// SyncHarvesterVMILabelsToPod syncs some specific labels from the VMI to the
+// Pod to make sure that e.g. the maintenance mode strategy is correctly applied
 func (h *VMIController) SyncHarvesterVMILabelsToPod(_ string, vmi *kubevirtv1.VirtualMachineInstance) (*kubevirtv1.VirtualMachineInstance, error) {
 	if vmi == nil || vmi.DeletionTimestamp != nil {
 		return vmi, nil
 	}
 
-	logrus.Debugf("Syncing labels %v for VMI %v to Pod", vmi.Labels, vmi.Name)
-
-	harvesterVMILabels := map[string]string{}
-	for label, value := range vmi.Labels {
-		if strings.HasPrefix(label, HarvesterLabelPrefix) {
-			harvesterVMILabels[label] = value
-		}
+	vmiLabel, ok := vmi.Labels[util.LabelMaintainModeStrategy]
+	if !ok {
+		vmiLabel = util.DefaultMaintainModeStrategy
 	}
 
 	activePodUIDs := vmi.Status.ActivePods
@@ -81,21 +71,10 @@ func (h *VMIController) SyncHarvesterVMILabelsToPod(_ string, vmi *kubevirtv1.Vi
 			continue
 		}
 
-		newLabels := maps.Clone(pod.Labels)
-		// delete Harvester labels from Pod, if they are deleted from VMI
-
-		maps.DeleteFunc(newLabels, func(k, _ string) bool {
-			if _, ok := harvesterVMILabels[k]; !ok && strings.HasPrefix(k, HarvesterLabelPrefix) {
-				return true
-			}
-			return false
-		})
-
-		maps.Copy(newLabels, harvesterVMILabels)
-
-		if !maps.Equal(pod.Labels, newLabels) {
+		podLabel, ok := pod.Labels[util.LabelMaintainModeStrategy]
+		if !ok || podLabel != vmiLabel {
 			newPod := pod.DeepCopy()
-			newPod.Labels = newLabels
+			newPod.Labels[util.LabelMaintainModeStrategy] = vmiLabel
 			_, err := h.podClient.Update(newPod)
 			if err != nil {
 				return vmi, fmt.Errorf("failed to sync Harvester VMI labels to pod, %v", err)
--- a/pkg/controller/master/virtualmachine/vmi_controller.go
+++ b/pkg/controller/master/virtualmachine/vmi_controller.go
@@ -5,6 +5,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -13,7 +15,13 @@ import (
 
 	"github.com/harvester/harvester/pkg/builder"
 	kubevirtctrl "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/indexeres"
 	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	VirtualMachineCreatorNodeDriver = "docker-machine-driver-harvester"
+	HarvesterLabelPrefix            = "harvesterhci.io"
 )
 
 // hostLabelsReconcileMapping defines the mapping for reconciliation of node labels to virtual machine instance annotations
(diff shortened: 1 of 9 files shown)


pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -13,7 +17,14 @@ import (
 
 	"github.com/harvester/harvester/pkg/builder"
 	kubevirtctrl "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/ref"
 	"github.com/harvester/harvester/pkg/util"
+	"github.com/harvester/harvester/pkg/util/indexeres"
+)
+
+const (
+	VirtualMachineCreatorNodeDriver = "docker-machine-driver-harvester"
```
w13915984028: VirtualMachineCreatorNodeDriver has been defined on
https://github.com/harvester/harvester/blob/70aa161bb60453c773788dde599f1805cb44802c/pkg/util/constants.go#L254

the `harvesterhci.io` is defined on above consts file and if need to refer it, define a pulic function GetHarvesterPrefix() on constants.go

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -644,6 +654,59 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
...
+	templateStrategy, ok := templateLabels[util.LabelMaintainModeStrategy]
+	if ok && strategy != templateStrategy {
+		logrus.WithFields(logrus.Fields{
+			"namespace": vm.Namespace,
+			"name":      vm.Name,
+			"strategy":  strategy,
+		}).Infof("maintenance mode strategy label in template does not match maintenance mode strategy label, applying %v", strategy)
+	}
+
+	if !ok || strategy != templateStrategy {
```
w13915984028, line 697: this patch, might cause existing running VM shows complains like `The VM generation is x, but the observed generation is y`

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -644,6 +654,59 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
...
 		}
 	}
+	return patchOps, nil
+}
+
+func (m *vmMutator) ensureMaintenanceModeStrategyIsSet(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) (types.PatchOps, error) {
+	// If the maintenante-mode strategy label is not set, add it with the default
+	// value of `Migrate`.
+	labels := vm.ObjectMeta.Labels
+	strategy, ok := labels[util.LabelMaintainModeStrategy]
```
w13915984028, line 664: forcely patch the default value `MaintainModeStrategyMigrate` might complaining of generation mismatch

w13915984028 (commented):
A main concern is: forcely patch the original allowed empty value (equal to the default MaintainModeStrategyMigrate), might cause some complains on UI.

In the `Rancher-manager` -> Harvester deploying guest-cluster case, the VM does not have this label;  after upgrade to new Harvester version, if we forcely patch the label to vm object, I suspect the Rancher UI might complain the generation, and the worst result is that it might hint user to reboot VM to sovle the generation gap (to be confirmed, but ever see similar case).

As a result, better to leave existing VMs as-is.

pkg/controller/master/virtualmachine/vmi_controller.go
```diff
@@ -30,6 +43,69 @@ type VMIController struct {
 	recorder            record.EventRecorder
 }
 
+// SyncHarvesterVMILabelsToPod ensures that all Harvester labels (i.e. those
+// with the harvesterhci.io/ prefix) from the VirtualMachineInstance are synced
+// to the Pod
+func (h *VMIController) SyncHarvesterVMILabelsToPod(_ string, vmi *kubevirtv1.VirtualMachineInstance) (*kubevirtv1.VirtualMachineInstance, error) {
```
w13915984028, line 43: This sync function could suffer performance from below case, as it always compares the labels.

Due to a know IP flapping bug on vmi, the vmi object generation is keeping increasing.

Image


As a trade-off, ensure non-default maintain strategy is synced to vmi object would be fine, if some legacy VM/POD does not have the desired label, the default value can still work, even it is not the best.

We could avoid keeping checking and syncing vmi/pod label.

pkg/webhook/resources/virtualmachine/mutator.go
m-ildefons: Not if done in the mutating webhook. The mutating webhook generates patches that are applied to the request _before_ processing by the API server. Therefore the generation of the object will not be affected.

pkg/webhook/resources/virtualmachine/mutator.go
m-ildefons: See above

pkg/webhook/resources/virtualmachine/mutator.go
m-ildefons: Note that applying default values to empty fields is one of the express purposes of a mutating webhook: https://book.kubebuilder.io/reference/admission-webhook

pkg/controller/master/virtualmachine/vmi_controller.go
m-ildefons: Ok

pkg/controller/master/virtualmachine/vmi_controller.go
m-ildefons: I don't think that there is a big performance impact, since this function utilizes the cached data to compare the labels maps and only makes a modification to the Pod when it finds a difference.
Therefore, most updates to the VMI will only cause a comparison of the two labels maps in the local controller cache. And this doesn't take a long time.

Nevertheless, I've modified the function to only compare and update the maintenance mode strategy label - other labels may be also added in the future as needed.

w13915984028: Before we have a full feature to sync lablels/annotations between vm, vmi, pod dynamically without causing UI complains like below, would suggest not adding it as this time; 

 the maintain controller should anyway handle the empty label & invalid label case effectively
 
 vm's label, annotation;  vm.spec.template.label, annotation are treated differently and it is a bit complex
 
image

w13915984028: The concern is: it is very difficult to test out the potential impacts on all existing features in a short time.

The missing of maintain label, should not in turn import a controller to forcely do this task and waste 99.99% of the runs

pkg/webhook/resources/virtualmachine/mutator.go
w13915984028: I did a test on v170:
(1) Start a cirros VM
(2) Add `test.io/test: test` to `vm.metadata.annotations`;  UI shows no complain
(3) Add `test.io/test: test` to `vm.spec.template.metadata.annotations`,  UI shows complain


Given this PR, if we do operation like (2) on an existing running VM (indeed, some controller updates the vm annotations on existing running VMs), it might cause complains like (3), you can test as this 

```
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/mac-address: '{"default":"02:78:80:cb:ec:f8"}'
      [{"metadata":{"name":"vm1-disk-0-zbggq","annotations":{"harvesterhci.io/imageId":"default/image-fzk7j"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"2Gi"}},"volumeMode":"Block","storageClassName":"longhorn-image-fzk7j"}}]
    network.harvesterhci.io/ips: '[]'
    test.io/test: test                                                               // added first time via `edit config`
  creationTimestamp: '2025-11-20T11:35:17Z'
  finalizers:
  generation: 2
  labels:
    harvesterhci.io/os: ''
      manager: virt-controller
      operation: Update
      subresource: status
      time: '2025-11-20T11:36:34Z'
  name: vm1
  namespace: default
  resourceVersion: '1023415'
  uid: 786c02ca-c647-42a1-9401-e91ad514a403
spec:
  runStrategy: RerunOnFailure
  template:
    metadata:
      annotations:
        test.io/test: test                                         // added second time via `edit config`
      creationTimestamp: null
      labels:
        harvesterhci.io/vmName: vm1
```

image

pkg/controller/master/virtualmachine/vmi_controller.go
m-ildefons: 1) 
The UI shows VirtualMachine objects.
This sync function does a unidirectional sync of lables from VirtualMachineInstance objects to the corresponding Pod objects. The UI will not show updates to the Pod with the warning you speak of.

2)
This function doesn't sync labels between VirtualMachine and VirtualMachineInstance objects and it never did.

3)
This function doesn't sync annotations and it never did.

4)
Buggy behavior of other parts of the system is no excuse for incorrect behavior of this part.

5)
As I've explained in my previous comment, the function only applies an update when it finds that the labels differ. Therefore it will not apply an update every time the object generation is increased.
Still, the controller needs to check if the label has changed whenever the object changes. That is just how controllers work.

pkg/webhook/resources/virtualmachine/mutator.go
m-ildefons: If you update the VirtualMachine's `.spec` then the controller needs to update the respective VirtualMachineInstance. There is no way around it, that is just how controllers work. In fact, it's necessary to make anything in Kubernetes work.

The update that causes the complaint doesn't necessarily originate from a mutating webhook and it also doesn't necessarily originate from _this_ mutating webhook.

Since you've demonstrated that this warning appears also in a situation that doesn't have anything to do with this webhook, it's just fair to assume that this webhook isn't the cause of that warning.

w13915984028: Here is the PR's side-effect, I did a demo code on another PR https://github.com/harvester/harvester/pull/9564

The `ensureMaintenanceModeStrategyIsSet` is called `OnUpdate` upon vm object https://github.com/harvester/harvester/pull/9528/files#diff-1bde4c748f58e474d67fd449ce94739fe11076c2ba5c63f79cfda2f94c4cb0a3R145

The `OnUpdate` can be triggered when user only adds an annotation to VM object.

For those legacy VMs which have no maintenance mode strategy label, if any controller/human updates VM's annotation, the `OnChange` happens, the webhook `ensureMaintenanceModeStrategyIsSet` runs, and patches the vm's related label, which in turn causes UI warning.

In my demo PR https://github.com/harvester/harvester/pull/9564, the VM's annotation `testtest: test6` is updated from `test5` to `test6`, and the webhook updates the template of vm object, then the UI shows:

image


Forcely append the `/spec/template/metadata/labels/harvesterhci.io~1maintain-mode-strategy"` label can happen in those unexpected path, and user is the victim.

https://github.com/harvester/harvester/pull/9528/files#diff-1bde4c748f58e474d67fd449ce94739fe11076c2ba5c63f79cfda2f94c4cb0a3R685-R686

In k8s, when a filed is empty, it is treated with default value in general. The `MaintainModeStrategyMigrate` is the default, if an object does not include this annotation, use the default value. It is very rarely that the default value should be forcely set upon object.

m-ildefons: Screenshot at 2025-11-20 16-43-06

If you edit `.spec.template.metadata.annotations` then there will always be the message "VirtualMachine generation is x, but latest observed generation is x+1".
This happens regardless of whether this webhook implements `ensureMaintenanceModeStrategyIsSet` as in this PR, or not.

This message indicates that the template for the VirtualMachineInstance has been updated, but the VirtualMachineInstance object has not yet been re-created to receive this update.
But that is correct if the maintenance mode strategy has been changed and the VM has not been restarted to apply this change.

But this webhook also serves an important function:
If the user updates only the `.metadata.labels` of the VirtualMachine object and does not update the `.spec.template.metadata.labels`, in particular with a non-default maintenance mode strategy, then this webhook ensures that the desired strategy is propagated correctly. Otherwise there could be a discrepancy between the `.metadata.labels` of the VirtualMachine object, the template and subsequently the VirtualMachineInstance

w13915984028:
For v170, would suggest we only do below code change:

https://github.com/harvester/harvester/blob/38979d6eeea24f0a7619a35135df00b627776320/pkg/util/drainhelper/helper.go#L153-L166

`if ok && value != util.MaintainModeStrategyMigrate`
->
`if value == util.MaintainModeStrategyShutdownAndRestartAfterEnable ||  value == util.MaintainModeStrategyShutdownAndRestartAfterDisable || value == util.MaintainModeStrategyShutdown`

the logic is same with:

https://github.com/harvester/harvester/blob/38979d6eeea24f0a7619a35135df00b627776320/pkg/controller/master/nodedrain/nodedrain_controller.go#L544-L549
and
https://github.com/harvester/harvester/blob/38979d6eeea24f0a7619a35135df00b627776320/pkg/controller/master/nodedrain/nodedrain_controller.go#L153-L155

And, update the document: update VM's instance label will cause `generatiton mismatch`.

patch the default stratey to VM object has high chance to import `generaiton mismatch` warnings on UI, can be further discussed on v180

Issue #9558, opened 2025-11-20
[BUG] terraform apply same logic storage-network setting cannot pass


FrankYang0529:
### Describe the Bug

In current implementation, following two setting are different in `storage-network` setting webhook.

```
{"clusterNetwork":"new-cn","exclude":["172.16.0.4/32","172.16.0.3/32"],"range":"172.16.0.0/24","vlan":1}
```

```
{"clusterNetwork":"new-cn","exclude":["172.16.0.3/32","172.16.0.4/32"],"range":"172.16.0.0/24","vlan":1}
```

However, this two value should be logically same, so we don't need to do further check.

In current version, if there is any running VM, the webhook rejects this change.

#### Why not fix this in Harvester webhook?

It's easy to update the check in Harvester webhook. However, it's complicated to update related controller logic. The controller uses sha1 hash of setting value to identify resources. In controller side, we only get new setting value. Without the old one, we have to try all kinds of combination to check whether the sha1 hash is same.

### To Reproduce

1. Create a Harvester cluster.
2. Set up a new cluster network `new-cn`.
3. Apply following terraform.
```tf
 locals {
   storage_network = {
     "vlan" = 1
     "clusterNetwork" = "new-cn"
     "range" = "172.16.0.0/24"
     "exclude" = ["172.16.0.4/32","172.16.0.3/32"]
   }
 }

 resource "harvester_setting" "storage_network" {
   name = "storage-network"
   value = jsonencode(local.storage_network)
 }
```
4. Create a VM.
5. Apply following terraform. The webhook rejects it.
```tf
 locals {
   storage_network = {
     "vlan" = 1
     "clusterNetwork" = "new-cn"
     "range" = "172.16.0.0/24"
     "exclude" = ["172.16.0.3/32","172.16.0.4/32"]
   }
 }

 resource "harvester_setting" "storage_network" {
   name = "storage-network"
   value = jsonencode(local.storage_network)
 }
```
### Expected Behavior

The terraform provider should check whether the value in terraform is logical same.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.5.x, v1.6.x, v1.7.0-rcx

m-ildefons:
## Validation Test

<details><summary> 🟢 Install Harvester v1.8.0-dev-20260201</summary><img width="3829" height="1991" alt="Image" src="https://github.com/user-attachments/assets/5e90fdf3-c45f-4f57-af0f-b254f563ba41" /></details>

[...]

PR #9559, opened 2025-11-20
fix(storage-network): check whether string function are same

FrankYang0529:
#### Problem:
In previous implementation, following two setting are different in `storage-network` setting webhook.

```
{"clusterNetwork":"new-cn","exclude":["172.16.0.4/32","172.16.0.3/32"],"range":"172.16.0.0/24","vlan":1}
```

```
{"clusterNetwork":"new-cn","exclude":["172.16.0.3/32","172.16.0.4/32"],"range":"172.16.0.0/24","vlan":1}
```

However, this two value should be logically same, so we don't need to do further check.

#### Solution:
Webhook should not reject the change if the change just has different order.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9558

#### Test plan:
2. Set up a new cluster network `new-cn`.
3. Set the `storage-setting` as `{"clusterNetwork":"new-cn","exclude":["172.16.0.4/32","172.16.0.3/32"],"range":"172.16.0.0/24","vlan":1}`.
4. Create a VM.
5. Set the `storage-setting` as `{"clusterNetwork":"new-cn","exclude":["172.16.0.3/32","172.16.0.4/32"],"range":"172.16.0.0/24","vlan":1}`. The webhook should not reject it.

### DIFF
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -1167,9 +1167,24 @@ func (v *settingValidator) validateUpdateStorageNetwork(oldSetting *v1beta1.Sett
 	}
 
 	var (
-		config *networkutil.Config
-		err    error
+		oldConfig networkutil.Config
+		config    *networkutil.Config
+		err       error
 	)
+	if oldSetting.Default != "" {
+		if err = json.Unmarshal([]byte(oldSetting.Default), &oldConfig); err != nil {
+			logrus.WithError(err).
+				WithField("oldSetting.Default", oldSetting.Default).
+				Warn("Failed to unmarshal old storage-network setting")
+		}
+	}
+	if oldSetting.Value != "" {
+		if err = json.Unmarshal([]byte(oldSetting.Value), &oldConfig); err != nil {
+			logrus.WithError(err).
+				WithField("oldSetting.Value", oldSetting.Value).
+				Warn("Failed to unmarshal old storage-network setting")
+		}
+	}
 	if config, err = v.validateNetworkHelper(settings.StorageNetworkName, newSetting.Default); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -1167,9 +1167,24 @@ func (v *settingValidator) validateUpdateStorageNetwork(oldSetting *v1beta1.Sett
...
+				Warn("Failed to unmarshal old storage-network setting")
+		}
+	}
+	if oldSetting.Value != "" {
+		if err = json.Unmarshal([]byte(oldSetting.Value), &oldConfig); err != nil {
+			logrus.WithError(err).
+				WithField("oldSetting.Value", oldSetting.Value).
+				Warn("Failed to unmarshal old storage-network setting")
+		}
+	}
```
WebberHuang1118, line 1187: Suggest wrapping this into a separate function.

WebberHuang1118 (commented):
Overall LGTM - just a nit. Thanks for the improvement.

Vicente-Cheng (approved):
thanks for the improvement!

rrajendran17:
LGTM,Thanks

rrajendran17 (approved):

w13915984028:
We did not mention anything about the controller side:

From below code, such change will cause the controller to re-sync SN, could it bring some issues? and it might defeat another ongoing PR https://github.com/harvester/harvester/pull/9453

https://github.com/harvester/harvester/blob/32c65348a312b033192ee08898036e68df34f34f/pkg/controller/master/storagenetwork/storage_network.go#L228

https://github.com/harvester/harvester/blob/32c65348a312b033192ee08898036e68df34f34f/pkg/controller/master/storagenetwork/storage_network.go#L309-L324

PR #9593, opened 2025-11-24
fix(ci): change the condition while checking content in the issue

Yu-Jack:
https://github.com/harvester/harvester/issues/9253
https://github.com/harvester/harvester/actions/runs/19610896417/job/56156546445


Vicente-Cheng (approved):
thanks for fixing that!

Issue #9595, opened 2025-11-24
[BUG] Upgrade might stuck with waiting for the VM migration with pre-draining stage


Vicente-Cheng:
### Describe the Bug

Upgrading to the master-head (with Rancher v2.13.0-rc3) may result in getting stuck waiting for the VM migration.

### To Reproduce

1. create v1.6.1 cluster
2. upgrade to master-head
3. you might see the upgrade stuck on waiting for VM migration (not always reproduce)

### Expected Behavior

can upgrade smoothly w/o any issues.

### Support Bundle for Troubleshooting

No need

### Environment

- Harvester version:
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Additional context

RCA:
We improved the node-down controller with this PR https://github.com/harvester/harvester/pull/9480. Now, if we find the node is ready, we will try to remove the Kubevirt drain and out-of-service taint.
If the taint of kubevirt drain was removed during the upgrade, the VM migration might not be triggered

### Workaround and Mitigation

Manually adding the kubevirt drain will trigger the VM migration again.

Vicente-Cheng:
We need to wait for rc4 or v1.7-head to test

asc5543:
# Descheduler Function Test
## Test result
Test PASS, the upgrade won't stuck and the VM imgration success.

## Environment
* Harvester ISO version: `v1.6.1` -> `v1.7.0-rc5`
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): `2 node Baremetal`

## Test Scenario
### Preparation

1. <details> <summary> Create 2 nodes cluster</summary>

    <img width="1907" height="1039" alt="Image" src="https://github.com/user-attachments/assets/3817831f-fc3a-46fa-ad50-87c47a25db66" />

   </details>

1. Prepare the v1.7.0-rc5 ISO file
1. <details> <summary> Prepare the version.yaml file </summary>

    ```
    apiVersion: harvesterhci.io/v1beta1
    kind: Version
    metadata:
      name: v1.7.0-rc5
      namespace: harvester-system
    spec:
      isoChecksum: 'f2a38b078c9af22414713959a3a349db30a88265a8967ee7ad2bc2b1d0bb03075383315518206af6abf758f018e697a4d859fdb913d55003ea5df9dc07a29c64'
[...]

PR #9596, opened 2025-11-24
controller: should not remove the node drain taint while upgrading

Vicente-Cheng:
- We rely on this node drain taint to trigger the migration during upgrade, so skip this mechanism during upgrade.

#### Problem:
The upgrade might stuck because we remove the taint of kubevirt drain

#### Solution:
Suspend this behavior durning the upgrade

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9595

#### Test plan:
- Create 3 node cluster
- execute the upgrade with these changes
- the taint of kubevirt drain should not be removed on pre-draining stage

### DIFF
--- a/pkg/controller/master/node/node_down_controller.go
+++ b/pkg/controller/master/node/node_down_controller.go
@@ -11,10 +11,15 @@ import (
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 
 	"github.com/harvester/harvester/pkg/config"
+	"github.com/harvester/harvester/pkg/controller/master/upgrade"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/settings"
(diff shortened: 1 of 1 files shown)


WebberHuang1118 (dismissed):
LGTM, thanks.

starbops (dismissed):
LGTM, thank you.

Vicente-Cheng:
Improve the logger, please help recheck!

pkg/controller/master/node/node_down_controller.go
```diff
@@ -11,14 +11,23 @@ import (
...
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/settings"
+	"github.com/harvester/harvester/pkg/util"
 )
 
 const (
 	nodeDownControllerName = "node-down-controller"
+
+	// labels for upgrade state
+	upgradeStateLabel     = "harvesterhci.io/upgradeState"
```
ibrokethecloud: do we want to use the existing constants https://github.com/harvester/harvester/blob/master/pkg/util/constants.go#L33

https://github.com/harvester/harvester/blob/master/pkg/controller/master/upgrade/job_controller.go#L36
https://github.com/harvester/harvester/blob/master/pkg/controller/master/upgrade/job_controller.go#L37

Vicente-Cheng: good suggestion!
updated

WebberHuang1118 (approved):

ibrokethecloud (approved):

Vicente-Cheng:
@mergifyio backport v1.7

Issue #9467, opened 2025-11-11
[ENHANCEMENT] Exclude Harvester VM workloads from Longhorn's `auto-delete-pod-when-volume-detached-unexpectedly` setting


innobead:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

- [ ] https://github.com/longhorn/longhorn/issues/12120, the blacklist feature is available on Longhorn 1.10.1. Exclude VM workloads from this setting by adding relevant resources to the setting. 
- [ ] Since baremetal container workloads are supported on the Harvester cluster, we need to re-enable the setting, which has been disabled in https://github.com/harvester/harvester/pull/8748, but use the new blacklist setting from Longhorn in version 1.8 to exclude VM workloads.

### Additional context

cc @harvester/storage  @derekbit

innobead:
@Vicente-Cheng @WebberHuang1118: Moving this to 1.8.0 and backporting to 1.7.1 due to time constraints. WDYT?

WebberHuang1118:
> [@Vicente-Cheng](https://github.com/Vicente-Cheng) [@WebberHuang1118](https://github.com/WebberHuang1118): Moving this to 1.8.0 and backporting to 1.7.1 due to time constraints. WDYT?

It’s okay for me, since this setting is currently disabled on the Longhorn side as noted in [this PR](https://github.com/harvester/harvester/pull/8748). I can hold the PR and postpone it to v1.7.1 and v1.8.0.

WebberHuang1118:
The related PR has been merged—please verify with the 2026/03/02 daily build. Thanks.

m-ildefons:
## Validation

<details><summary> 🟢 Create Harvester v1.8.0-rc1 cluster in local ipxe-examples based environment</summary>
<img width="1545" height="821" alt="Image" src="https://github.com/user-attachments/assets/f56db2bc-a507-4849-a7fa-fed1f0c55dea" />
</details>

<details><summary> 🟢 Check Longhorn setting `auto-delete-pod-when-volume-detached-unexpectedly`</summary>
<img width="1104" height="328" alt="Image" src="https://github.com/user-attachments/assets/569f9151-58b5-4f43-8423-0cf37a2be0bb" />
</details>

<details><summary> 🟢 Check Longhorn setting `blacklist-for-auto-delete-pod-when-volume-detached-unexpectedly`</summary>
[...]

PR #9609, opened 2025-11-25
enhancement: exclude VM workload from LH's automatic deletion mechanism

WebberHuang1118:
#### Problem:
Exclude VM workload from LH's automatic deletion mechanism

#### Solution:
Exclude VM workload from LH's automatic deletion mechanism via the blacklist

#### Related Issue(s):
#9467 

#### Test plan:
1. Build Harvester using the ISO from this PR.
2. Check the LH setting `auto-delete-pod-when-volume-detached-unexpectedly` is set to `true`.
3. Check the LH setting `blacklist-for-auto-delete-pod-when-volume-detached-unexpectedly` is set to `kubevirt.io/VirtualMachineInstance`.
4. Create a VM with a Longhorn volume root disk and set the RunStrategy to manual.
5. Delete the root disk VM's Instance Manager (IM) pod. The VM should remain in a Paused state, and the corresponding virt-launcher pod will not be deleted.

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -498,7 +498,8 @@ longhorn:
     priorityClass: &longhornPriorityClass system-cluster-critical
     autoCleanupSnapshotWhenDeleteBackup: true
     orphanResourceAutoDeletion: instance
-    autoDeletePodWhenVolumeDetachedUnexpectedly: false
+    autoDeletePodWhenVolumeDetachedUnexpectedly: true
+    blacklistForAutoDeletePodWhenVolumeDetachedUnexpectedly: "kubevirt.io/VirtualMachineInstance"
 
   # after upgrade to longhorn 1.6.0, we need to set the priorityClass for longhorn-manager, longhorn-driver and longhorn-ui
   # or it will be default one longhorn-critical


deploy/charts/harvester/values.yaml
```diff
@@ -498,7 +498,8 @@ longhorn:
     priorityClass: &longhornPriorityClass system-cluster-critical
     autoCleanupSnapshotWhenDeleteBackup: true
     orphanResourceAutoDeletion: instance
-    autoDeletePodWhenVolumeDetachedUnexpectedly: false
+    autoDeletePodWhenVolumeDetachedUnexpectedly: true
+    blacklistForAutoDeletePodWhenVolumeDetachedUnexpectedly: "kubevirt.io/VirtualMachineInstance"
```
brandboat, line 502: do we need to take care this for the upgrade flow?

WebberHuang1118: It shouldn’t be necessary—the management chart will apply this manifest.

brandboat: Yeah, you're right, I missed this part, thanks for clarification.

brandboat (approved):
LGTM

votdev (approved):
LGTM

Issue #9594, opened 2025-11-24
[BUG] Fail to same-version upgrade on v1.7.0-rc3


albinsun:
### Describe the Bug

`retry attempted 3/3 failed due to error: failed to fetch image size: Content-Length is not found in the header`
<img width="1566" height="887" alt="Image" src="https://github.com/user-attachments/assets/041d8213-f7ff-435b-be52-574558f9ab02" />
<img width="1566" height="560" alt="Image" src="https://github.com/user-attachments/assets/f585c991-cc6d-489b-98dd-8de6fdfeb29e" />

```
│ time="2025-11-24T05:48:35Z" level=error msg="error syncing 'cattle-system/sync-additional-ca': handler system-upgrade-controller: secrets \"harvester-additional-ca\" not found, requeuing"                                                                                                                                                    
| time="2025-11-24T06:04:00Z" level=error msg="error syncing 'cattle-system/apply-sync-additional-ca-on-hp-5-tink-system-with-2c27af4-dbee8': handler system-upgrade-controller: jobs.batch \"apply-sync-additional-ca-on-hp-5-tink-system-with-2c27af4-dbee8\" not found, requeuing" 
```
<img width="1566" height="560" alt="Image" src="https://github.com/user-attachments/assets/ed5091ae-7cf6-42d3-9f81-01b73b12179c" />

### To Reproduce

1. Setup single node `v1.7.0-rc3`
2. Trigger upgrade with `https://releases.rancher.com/harvester/v1.7.0-rc3/version.yaml`

### Expected Behavior

Should be able to upgrade

### Support Bundle for Troubleshooting

[supportbundle_f3be46fd-a27e-4499-ab98-af1a8922abfe_2025-11-24T06-34-19Z.zip](https://github.com/user-attachments/files/23706464/supportbundle_f3be46fd-a27e-4499-ab98-af1a8922abfe_2025-11-24T06-34-19Z.zip)

[hvst-upgrade-nwqc7-upgradelog-archive-2025-11-24T06-33-51Z.zip](https://github.com/user-attachments/files/23706481/hvst-upgrade-nwqc7-upgradelog-archive-2025-11-24T06-33-51Z.zip)

### Environment

- Harvester version: `v1.7.0-rc3`
- Impacted VM: 
- Impacted volume (PV):
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

brandboat:
[...]

PR #9611, opened 2025-11-25
Support x-goog-stored-content-length header as fallback for image size detection

brandboat:
#### Problem:

see https://github.com/harvester/harvester/issues/9594

#### Solution:

Support x-goog-stored-content-length header as fallback for image size detection

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9594

#### Test plan:

well described in https://github.com/harvester/harvester/issues/9594

When the HTTP server doesn't provide a Content-Length header, CDI cannot track the download progress because it doesn't know the total file size. In this case, [NewFormatReaders](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/importer/format-readers.go#L82-L94) skips creating the ProgressReader wrapper, which means no progress updates are sent during the entire HTTP download phase.

The progress reporting only starts after the download completes and qemu-img conversion begins ([qemu.go:279-291](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/image/qemu.go#L279-L291)). Since both HTTP download and qemu conversion report progress using the same DataVolume .status.progress field, users see the progress stuck at 0% for an extended period (during the entire download) until conversion starts.

**In short, user will see vm image progress stays in 0% during the entire http download until the import pod starts to perform qemu-conversion.**

### DIFF
--- a/pkg/image/cdi/common.go
+++ b/pkg/image/cdi/common.go
@@ -140,8 +140,14 @@ func fetchImageSize(url string) (int64, error) {
 	}
 	defer rsp.Body.Close()
 
-	contentLengthStr := rsp.Header.Get("Content-Length")
-	if contentLengthStr == "" {
+	var contentLengthStr string
+	if cl := rsp.Header.Get("Content-Length"); cl != "" {
+		contentLengthStr = cl
+	} else if xgcl := rsp.Header.Get("x-goog-stored-content-length"); xgcl != "" {
+		// Fallback to x-goog-stored-content-length if Content-Length is not available
+		contentLengthStr = xgcl
+	} else {
+		// if both headers are missing, return an error
 		return 0, ErrHeaderContentLengthNotFound
 	}
 
--- a/pkg/image/cdi/common.go
+++ b/pkg/image/cdi/common.go
@@ -140,22 +140,27 @@ func fetchImageSize(url string) (int64, error) {
 	}
 	defer rsp.Body.Close()
 
-	var contentLengthStr string
-	if cl := rsp.Header.Get("Content-Length"); cl != "" {
-		contentLengthStr = cl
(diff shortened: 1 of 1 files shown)


brandboat:
> In short, user will see vm image progress stays in 0% during the entire http download until the import pod starts to perform qemu-conversion.

quoted from pr description, this is something we need to aware of, but I don't think there is any quick fix because this is a limitation from the upstream. IMO, we can update the documentation to clearly inform users about this limitation.

c.c @Vicente-Cheng, @WebberHuang1118

pkg/image/cdi/common.go
```diff
@@ -140,8 +140,14 @@ func fetchImageSize(url string) (int64, error) {
...
-	contentLengthStr := rsp.Header.Get("Content-Length")
-	if contentLengthStr == "" {
+	var contentLengthStr string
+	if cl := rsp.Header.Get("Content-Length"); cl != "" {
+		contentLengthStr = cl
+	} else if xgcl := rsp.Header.Get("x-goog-stored-content-length"); xgcl != "" {
+		// Fallback to x-goog-stored-content-length if Content-Length is not available
+		contentLengthStr = xgcl
+	} else {
+		// if both headers are missing, return an error
```
WebberHuang1118: Suggest wrapping this logic into a separate function.

WebberHuang1118 (commented):
LGTM, just a nit, thanks.

pkg/image/cdi/common.go
brandboat: updated in [243638c](https://github.com/harvester/harvester/pull/9611/commits/243638c51e562bcf7c570b70e77396b959592fc5)

WebberHuang1118 (approved):

wheatdog (approved):

Vicente-Cheng (approved):
thanks for the quick enhancement!

PR #9613, opened 2025-11-25
fix(restore): set InProgress=true if VM is not ready

FrankYang0529:
#### Problem:
When we restoring a VM, we set InProgress=false during volumes are ready and VM is not ready. The VM can run. However, if user removes the VMBackup, the VMRestore cannot reconcile correctly.

#### Solution:
Set InProgress=true even if the volumes are ready.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/6126

#### Test plan:
1. Create a VMBackup.
2. Create a VMRestore.
3. Wait for message "Waiting for target vm to be ready" and remove the VMBackup. The webhook should deny the request.
Screenshot 2025-11-25 at 3 56 41 PM

We found the issue during testing a refactor item. However, the issue is not from the refactor PR, because we don't change logic in https://github.com/harvester/harvester/commit/f192270c2bfc04d3fc99fe8051a6d2b929654c8d#diff-3255f37987feb6224175b7683e961f0e2b786f289cff7457f334ea16b0abbed7.
Screenshot 2025-11-25 at 3 54 22 PM

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -1258,7 +1258,7 @@ func (h *RestoreHandler) updateStatus(
 	if !vm.Status.Ready {
 		h.recifyProgressBeforeVMStart(restoreCpy)
 		message := "Waiting for target vm to be ready"
-		setCondition(restoreCpy, harvesterv1.BackupConditionProgressing, false, "", message)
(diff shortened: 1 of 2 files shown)


WebberHuang1118 (dismissed):
LGTM, thanks for the fix.

WebberHuang1118 (approved):

Vicente-Cheng (approved):
lgtm, thanks!

PR #9624, opened 2025-11-25
consistent maintenance mode strategy behavior

m-ildefons:
The Pod deletion filter identifies the Pods belonging to VMs, which need to be deleted during a node drain by looking at their maintenance-mode strategy label. The default value for this label is `Migrate`, which indicates that the Pod should be deleted during the node drain. Pods with an invalid label, i.e. where the value of the maintenance-mode strategy label is not one of:
  - Migrate
  - ShutdownAndRestartAfterEnable
  - ShutdownAndRestartAfterDisable
  - Shutdown are now also being treated with the default behavior. This fixes the problem that if a VM contains an invalied value in this label, nodes can become stuck in `Cordoned` state when transitioning to maintenance mode, as the node drain controller won't shut down the VM, but it's also not migrated away from the node. Therefore the VM keeps running, preventing the node from completely transitioning into maintenance mode.

Also adds a check to the validating webhook for the VirtualMachine resources.
This check also ensures that updates to virtual machines can not update the label to an invalid value that is none of the above mentioned.
However the check still alows VirtualMachines to be created without the maintenance mode strategy label, since external integrators like Rancher don't always set this label.

For backwards compatibility this webhook allows VirtualMachine updates to a VirtualMachine if the old VirtualMachine had an invalid maintenance mode strategy label and the new VirtualMachine does not change this label.
This case is crucial to allow existing VirtualMachines to continue working as expected. The webhook will emit a suitable log message.

related-to: #6835

#### Problem:

The scope of https://github.com/harvester/harvester/pull/9528 makes review too hard to complete in time for v1.7.0

#### Solution:

This PR contains a smaller and less controversial subset of the changes of PR#9528. This should make review easier and implement at least most of the necessary functionality.

#### Related Issue(s):

#6835

---

# Test plan:

## Scenario 1

Ensure that new VirtualMachines have a valid maintenance-mode strategy or none at all.

### Test 1

No maintenance-mode strategy given.

#### Setup:

Create a new VirtualMachine and specify no maintenance-mode strategy:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/os: linux
  name: test
[...]
```

#### Expected Result

The creation of the VM should be accepted.

### Test 2

Valid maintenance-mode strategy given.

#### Setup:

Create a new VirtualMachine and specify a valid maintenance-mode strategy:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: Shutdown
    [...]
  name: test
[...]
```

#### Expected Result

The previously specified maintenance-mode strategy should be persisted in the VirtualMachine resource's metadata:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: Shutdown
    [...]
  name: test
[...]
```

## Scenario 2

If an invalid maintenance-mode strategy is specified, the VirtualMachine resource should be rejected.

### Test 1

Invalid maintenance-mode strategy given.

#### Setup:

Create a new VirtualMachine and specify an invalid maintenance-mode strategy:
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: foobar
    [...]
  name: test
[...]
```

#### Expected Result

The creation should be rejected with a sensible error message

```
The request is invalid: metadata.labels[harvesterhci.io/maintain-mode-strategy]: invalid maintenance mode strategy: foobar
```

## Scenario 3

Updates to a VirtualMachine should be allowed if the new maintenance mode strategy is valid.

### Test 1

Valid maintenance-mode strategy given.
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: Migrate
    [...]
  name: test
[...]
```

#### Expected Result

The maintenance mode strategy should be successfully updated to the new value

### Test 2

Invalid maintenance-mode strategy given.
```
type: kubevirt.io.virtualmachine
metadata:
  namespace: default
  labels:
    harvesterhci.io/maintain-mode-strategy: foobar
    [...]
  name: test
[...]
```

#### Expected Result

The update to the VirtualMachine should be rejected on the basis that the new maintenance mode strategy is invalid

## Scenario 4

Ensure that existing VMs don't break.

Setup:
- Install Harvester without this change
- Create one VirtualMachine (`testvm-no-label`) without a maintenance mode strategy label
- Create one VirtualMachine (`testvm-invalid-label`) with an invalid maintenance mode strategy label
- Update to Harvester with this change

### Test 1

Make an update to both VMs (`testvm-no-label`, `testvm-invalid-label`) that does not modify the maintenance mode strategy label

#### Expected Result

The updates to the VirtualMachines should be successfully accepted and applied

### Test 2

Update the maintenance mode strategy label on both VMs (`testvm-no-label`, `testvm-invalid-label`)  to a new invalid value.

#### Expected Result

The updates to the VirtualMachines should be rejected because the new maintenance mode strategy is invalid

### Test 3

Update the maintenance mode strategy label on both VMs (`testvm-no-label`, `testvm-invalid-label`)  to a new valid value.

#### Expected Result

The updates to the VirtualMachines should be successfully accepted and applied

### DIFF
--- a/pkg/util/drainhelper/helper.go
+++ b/pkg/util/drainhelper/helper.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	"golang.org/x/exp/slices"
 	"k8s.io/apimachinery/pkg/selection"
 
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
@@ -151,15 +152,22 @@ func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) error {
 }
 
 func maintainModeStrategyFilter(pod corev1.Pod) drain.PodDeleteStatus {
-	// Ignore VMs that should not be migrated in maintenance mode. These
+	// If this label is set, the Pod belongs to a VM. Otherwise we don't need to
+	// skip the Pod.
+	vmName, vmOk := pod.Labels[util.LabelVMName]
+
+	// Ignore Pods of VMs that should not be migrated in maintenance mode. These
 	// VMs are forcibly shut down when maintenance mode is activated.
 	value, ok := pod.Labels[util.LabelMaintainModeStrategy]
-	if ok && value != util.MaintainModeStrategyMigrate {
+	if vmOk && ok && slices.Contains(util.MaintainModeStrategyShutdownValues, value) {
 		logrus.WithFields(logrus.Fields{
-			"namespace": pod.Namespace,
-			"pod_name":  pod.Name,
-		}).Infof("migration of pod owned by VM %s is skipped because of the label %s",
-			pod.Labels[util.LabelVMName], util.LabelMaintainModeStrategy)
+			"kind":                         "pod",
+			"namespace":                    pod.Namespace,
+			"pod_name":                     pod.Name,
+			util.LabelMaintainModeStrategy: value,
+			util.LabelVMName:               vmName,
+		}).Infof("migration of pod owned by VM %s is skipped because of the label %s with value %s",
+			vmName, util.LabelMaintainModeStrategy, value)
 		return drain.MakePodDeleteStatusSkip()
 	}
 	return drain.MakePodDeleteStatusOkay()
--- a/pkg/util/drainhelper/helper.go
+++ b/pkg/util/drainhelper/helper.go
@@ -155,20 +155,21 @@ func maintainModeStrategyFilter(pod corev1.Pod) drain.PodDeleteStatus {
 	// If this label is set, the Pod belongs to a VM. Otherwise we don't need to
 	// skip the Pod.
 	vmName, vmOk := pod.Labels[util.LabelVMName]
-
-	// Ignore Pods of VMs that should not be migrated in maintenance mode. These
-	// VMs are forcibly shut down when maintenance mode is activated.
-	value, ok := pod.Labels[util.LabelMaintainModeStrategy]
-	if vmOk && ok && slices.Contains(util.MaintainModeStrategyShutdownValues, value) {
-		logrus.WithFields(logrus.Fields{
-			"kind":                         "pod",
-			"namespace":                    pod.Namespace,
-			"pod_name":                     pod.Name,
-			util.LabelMaintainModeStrategy: value,
-			util.LabelVMName:               vmName,
-		}).Infof("migration of pod owned by VM %s is skipped because of the label %s with value %s",
-			vmName, util.LabelMaintainModeStrategy, value)
-		return drain.MakePodDeleteStatusSkip()
+	if vmOk {
+		// Ignore Pods of VMs that should not be migrated in maintenance mode. These
+		// VMs are forcibly shut down when maintenance mode is activated.
+		value, ok := pod.Labels[util.LabelMaintainModeStrategy]
+		if ok && slices.Contains(util.MaintainModeStrategyShutdownValues, value) {
+			logrus.WithFields(logrus.Fields{
+				"kind":                         "pod",
+				"namespace":                    pod.Namespace,
+				"pod_name":                     pod.Name,
+				util.LabelMaintainModeStrategy: value,
+				util.LabelVMName:               vmName,
+			}).Infof("migration of pod owned by VM %s is skipped because of the label %s with value %s",
+				vmName, util.LabelMaintainModeStrategy, value)
+			return drain.MakePodDeleteStatusSkip()
+		}
 	}
 	return drain.MakePodDeleteStatusOkay()
 }
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -9,6 +9,8 @@ import (
 
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	ctlstoragev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/storage/v1"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/exp/slices"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -235,6 +237,10 @@ func (v *vmValidator) Create(_ *types.Request, newObj runtime.Object) error {
 		return err
 	}
 
+	if err := v.checkMaintenanceModeStrategyIsValid(vm, nil); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -280,6 +286,10 @@ func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj run
 	}
 
 	if err := v.checkForDuplicateMacAddrs(newVM); err != nil {
+		return err
+	}
+
+	if err := v.checkMaintenanceModeStrategyIsValid(newVM, oldVM); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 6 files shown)


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -646,3 +656,65 @@ func (v *vmValidator) checkEmptyMemory(vm *kubevirtv1.VirtualMachine) error {
 	}
 	return nil
 }
+
+func (v *vmValidator) checkMaintenanceModeStrategyIsValid(newVM, oldVM *kubevirtv1.VirtualMachine) error {
+	newLabels := newVM.ObjectMeta.Labels
+	newStrategy, newOk := newLabels[util.LabelMaintainModeStrategy]
+	// Creating a new VM without maintenance mode strategy is not ok
```
w13915984028: In `rancher-manager` -> `Harvester` scenario, the created/updated VM, does not include this label

The `empty means default` might be acceptable

```
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  annotations:
    harvesterhci.io/mac-address: '{"nic-0":"52:c1:db:cb:43:a2"}'
      [{"metadata":{"name":"gc5-pool1-sq6g8-q6z8l-disk-0-9tcd4","creationTimestamp":null,"annotations":{"harvesterhci.io/imageId":"default/image-q2rth"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"20Gi"}},"storageClassName":"longhorn-image-q2rth","volumeMode":"Block"},"status":{}}]
  creationTimestamp: '2025-11-13T19:44:51Z'
  finalizers:
  generation: 1
  labels:
    guestcluster.harvesterhci.io/name: gc5
    harvesterhci.io/creator: docker-machine-driver-harvester
    harvesterhci.io/machineSetName: default-gc5-pool1
    harvesterhci.io/vmName: gc5-pool1-sq6g8-q6z8l
    nodepool.harvesterhci.io/name: pool1-sq6g8
  name: gc5-pool1-sq6g8-q6z8l
  namespace: default
```

w13915984028 (commented):
The first question is still about the empty label, esp. on Rancher/Terraform or third party api scenario, forcely requires this label might break them.

pkg/webhook/resources/virtualmachine/validator.go
m-ildefons: Updated to accept creation of VMs which don't have the label at all.
Invalid values are still rejected, as well as updates which contain either invalid values or remove the label.

pkg/util/drainhelper/helper.go
```diff
@@ -151,15 +152,22 @@ func DrainPossible(nodeCache ctlcorev1.NodeCache, node *corev1.Node) error {
 }
 
 func maintainModeStrategyFilter(pod corev1.Pod) drain.PodDeleteStatus {
-	// Ignore VMs that should not be migrated in maintenance mode. These
+	// If this label is set, the Pod belongs to a VM. Otherwise we don't need to
+	// skip the Pod.
+	vmName, vmOk := pod.Labels[util.LabelVMName]
```
w13915984028, line 157: return early, and no need to check vmOk in following proces
```
if !vmOk {
	return drain.MakePodDeleteStatusOkay()
}
```

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -646,3 +656,49 @@ func (v *vmValidator) checkEmptyMemory(vm *kubevirtv1.VirtualMachine) error {
...
+				util.MaintainModeStrategyMigrate,
+			)
+			return nil
+		}
+	}
+
+	// New maintenance mode strategy is invalid, not ok
+	if newOk && !slices.Contains(util.MaintenanceModeStrategyValidValues, newStrategy) {
+		return werror.NewInvalidError(
+			fmt.Sprintf("invalid maintenance mode strategy: %v", newStrategy),
```
w13915984028, line 664: could we merge two Sprintf into one?

w13915984028 (dismissed):
LGTM, thanks.

The tests cases are excellent.

pkg/webhook/resources/virtualmachine/validator.go
m-ildefons: No, because the function signature of `werror.NewInvalidError` is:
```golang
func NewInvalidError(message string, field string) AdmitError
```
i.e. it needs two separate strings. One for the error message and one for the path of the parameter field that caused it.
The two strings do not get concatenated.

pkg/util/drainhelper/helper.go
m-ildefons: The log message includes the value of `vmName`. Therefore we must make sure that it was actually found in the map, otherwise a nil-dereference may occur.
I'm going to turn the logic around and pull the `if vmOk {}` out and move the read of the maintenance mode strategy label and the rest of the logic inside it. This makes it explicit that we need `vmOk` to be true for the rest of the logic.

w13915984028: when `vmOk is not true`, it has returned at beginning, no chance to run following and log it
```
if !vmOk {
	return drain.MakePodDeleteStatusOkay()
}

// do following

vs 

if vmOk {
	// do following
}
```
the first solution has less indent;  keep the second is also good.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -646,3 +656,49 @@ func (v *vmValidator) checkEmptyMemory(vm *kubevirtv1.VirtualMachine) error {
 	}
 	return nil
 }
+
+func (v *vmValidator) checkMaintenanceModeStrategyIsValid(newVM, oldVM *kubevirtv1.VirtualMachine) error {
+	newLabels := newVM.ObjectMeta.Labels
+	newStrategy, newOk := newLabels[util.LabelMaintainModeStrategy]
```
w13915984028, line 629: seems both newLabels and oldLabels only occur once, they can be merged with next line

w13915984028 (dismissed):
LGTM, thanks.

martindekov (dismissed):
LGTM Moritz, my understanding is that now we allow only from the list of known and valid labels and reject bad formats and properly first taking pods part of the vm for migration and from those pods pick the strategy and skip if pod is not part of the vm.

ihcsim (dismissed):

Vicente-Cheng:
Thanks @m-ildefons for this PR.
If #9528 was too complex for v1.7.1, we can backport this to v1.7.1.

m-ildefons:
Why is this now marked DNM? Please explain.

Vicente-Cheng:
> Why is this now marked DNM? Please explain.

The 1.7 was in code freeze stage.
Based on your description
> The scope of https://github.com/harvester/harvester/pull/9528 makes review too hard to complete in time for v1.7.0

I assume we should go back to review #9528 instead of this one.
Or you thought these changes were independent of #9528?
Please update the issue description.

m-ildefons:
This PR is independent of #9528

This PR contains a subset of changes from #9528. The changes in this PR had been identified by @w13915984028 as suitable for inclusion in v1.7.0, as discussed here: https://github.com/harvester/harvester/pull/9528#issuecomment-3572757277

Meanwhile #9528 contains further changes, which enforce the maintenance mode strategy behavior more aggressively. @w13915984028 raised concerns about being able to ensure correctness in time for the v1.7.0 release.

Therefore we decided that it would be best not to merge #9528 before the v1.7.0 release.
Instead I made this PR with only the changes that we are sure about, so we can at least have them for v1.7.0
All other changes can then receive the necessary work later.

w13915984028:
I confirm that the current change ensures to fix the existing bug: when the expected label is wrong/missing, the maintain/drain controller might go in to wrong state. But allow the empty label to avod breking cases like docker-machine-driver which does not apply the maintain label at all on VM.

Please @votdev also double check the fix, thanks.

Vicente-Cheng:
Got it, Thanks for the explanation @m-ildefons, 

But as I mentioned, currently we only merge the blocker or high severity/priority issues at this moment.
We can recheck this one after the v1.7.0 release.
Thanks!

cc @olblak, @innobead

m-ildefons:
@Vicente-Cheng Now that v1.7.0 is out, can we resume work on this?

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -646,3 +656,49 @@ func (v *vmValidator) checkEmptyMemory(vm *kubevirtv1.VirtualMachine) error {
...
+	// integrators (e.g. Rancher) may omit it. Similarly removing the label is ok.
+	// Therefore just log that the label is missing
+	if !newOk {
+		logrus.WithFields(logrus.Fields{
+			"name":      newVM.Name,
+			"namespace": newVM.Namespace,
+		}).Warnf(
+			"no maintenance-mode strategy for VM, behavior will be equivalent to default `%v`",
+			util.MaintainModeStrategyMigrate,
+		)
```
Vicente-Cheng, line 640: Can we return early here?

Vicente-Cheng (dismissed):
Just one nit, also nest if could be improved, but it did not affect the readability.
So it depends on you. 

LGTM thanks.

pkg/webhook/resources/virtualmachine/validator.go
m-ildefons: Good spot. We can indeed return early here, because all other cases are guarded behind a `if newOk` condition.

Vicente-Cheng (approved):
lgtm, thanks for the enhancement!

votdev (commented):
Even though the PR has already been merged; LGTM.

m-ildefons:
Should we backport this for v1.7.1 @Vicente-Cheng ?

Issue #9620, opened 2025-11-25
[BUG] Failed to upload file to create upgrade image in server-version setting


TachunLin:
### Describe the Bug

* When we open the server-version settings page.

* And upload a new image for upgrade (e.g harvester-v1.7.0-rc3-amd64.iso)

* The upload failed with error message

  ```
  {"data":"failed to update VM Image size and virtual size: Operation cannot be fulfilled on virtualmachineimages.harvesterhci.io \"image-sfj64\": the object has been modified; please apply your changes to the latest version and try again"}
  ```
  
  <img width="1598" height="786" alt="Image" src="https://github.com/user-attachments/assets/6665cc5f-80bc-4e2b-a70d-9bf9f50c75b1" />
  

  https://github.com/user-attachments/assets/61c7afe0-63c9-42f8-a3a0-5e11296d1c9d
  
* The upgrade image related storage class have created

  <img width="1599" height="784" alt="Image" src="https://github.com/user-attachments/assets/a644db37-c6ba-41e7-8edb-95013d2cae40" />

* But the image is actually can't be used

  <img width="1600" height="785" alt="Image" src="https://github.com/user-attachments/assets/2e9b1fb9-3d27-4434-8a42-d979dd22cfc1" />
  

### To Reproduce

1. Download the Harvester iso image file on your local (e.g harvester-v1.7.0-rc3-amd64.iso) 
2. Reach the settings -> server-version
3. Click the upgrade menu item
4. On the Upload New Image page, click the Upload file
5. Select the harvester-v1.7.0-rc3-amd64.iso on your local 
6. Check the upload file state

### Expected Behavior

Should be able to create new upgrade image from uploading file without failure

### Support Bundle for Troubleshooting

[supportbundle_e7abd598-3fed-4d03-8f1c-ef564e67edcb_2025-11-25T13-13-44Z.zip](https://github.com/user-attachments/files/23747562/supportbundle_e7abd598-3fed-4d03-8f1c-ef564e67edcb_2025-11-25T13-13-44Z.zip)

### Environment

- Harvester version: v1.7.0-rc3
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Single node bare-metal and kvm machine

### Additional context

N/A

### Workaround and Mitigation

N/A

TachunLin:
Verified **fixed** on `v1.7.0-rc5`. Close this issue.

### Result
[...]

PR #9627, opened 2025-11-26
Add retry mechanism in VMIOperator#UpdateVirtualSizeAndSize

brandboat:
#### Problem:

see https://github.com/harvester/harvester/issues/9620, during the vm image upload process, the vm image cr is updated at the moment and cause upload failed.

#### Solution:

Add retry mechanism in VMIOperator#UpdateVirtualSizeAndSize

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9620

#### Test plan:

well described in https://github.com/harvester/harvester/issues/9620

### DIFF
--- a/pkg/image/common/operator.go
+++ b/pkg/image/common/operator.go
@@ -8,8 +8,11 @@ import (
 	"time"
 
 	"github.com/rancher/norman/condition"
+	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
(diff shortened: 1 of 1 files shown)


pkg/image/common/operator.go
```diff
@@ -367,3 +372,45 @@ func (vmio *vmiOperator) Imported(old *harvesterv1.VirtualMachineImage, msg stri
...
+		}
+
+		newVMI, err := updateFunc(latestVMI.DeepCopy())
+		if err != nil {
+			return err
+		}
+
+		updatedVMI, err = vmio.UpdateVMI(latestVMI, newVMI)
+		if err != nil {
+			logrus.Warnf("Failed to update VMImage size: %v, will retry if retriable", err)
```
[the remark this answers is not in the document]
brandboat: Though the updateWithRetry is only used in update vmimage size at the moment, that doesn't mean we won't use it in other place.

WebberHuang1118 (dismissed):
LGTM, thanks for the fix, the PR has clean and compacted logic.

pkg/image/common/operator.go
Vicente-Cheng: It looks like this might be the general retry function for any VM Image update. 
So maybe the logger should not narrow down to the `Failed to update VMImage size`.
Could be something like `Failed to update VM Image again in retry mechanism: %v, will retry if retriable`.
WDYT?

Vicente-Cheng (dismissed):
a nit logger discussion, thanks for the PR!

pkg/image/common/operator.go
brandboat: Looks like I didn't push the latest change to the remote branch, my bad... Thanks for the reminder!

WebberHuang1118 (approved):

Vicente-Cheng (approved):

PR #9429, opened 2025-11-05
bump whereabouts to 0.9.2

rrajendran17:
#### Problem:
bump whereabouts to 0.9.2 (latest)

#### Solution:
bump whereabouts to 0.9.2 (latest)

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9102

PR #9631, opened 2025-11-26
use image value instead of hard coded image for new whereabouts manifest

rrajendran17:
#### Problem:
During bump-whereabouts https://github.com/harvester/harvester/pull/9429, a new deployment manifest from whereabouts was introduced which has hardcoded image latest.

#### Solution:
change to dynamically get the image value from charts value.yaml

#### Related Issue(s):

#### Test plan:
1.deploy/charts/harvester directory, run helm dependency update .
2.helm template .
3.Check if the deployment of whereabouts-controller shows whereabouts image with version 0.9.2


ibrokethecloud (approved):

Vicente-Cheng (approved):

Vicente-Cheng:
@mergifyio backport v1.7

Issue #9581, opened 2025-11-21
[ENHANCEMENT] add /etc/NetworkManager to persistent paths, avoid using /oem YAML for network config


tserong:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Harvester v1.7.0 switches from using wicked to NetworkManager (see https://github.com/harvester/harvester/issues/3418). The current implementation essentially takes the same approach as was used with wicked, i.e. config files for networking are written via `/oem/90_custom.yaml`, and are then applied on boot. This made the switch relatively easy in terms of implementation, but has some downsides, notably it's a pain to update network configuration after installation: you can make runtime configuration changes easily using the `nmcli` tool, but if you want those changes to persist after reboot you have to also edit either `/oem/90_custom.yaml` (for new installs) or, for older systems upgraded to v1.7.0, you need to edit `/oem/91_networkmanager.yaml`. This is messy, confusing, and error prone.

### Describe the solution you'd like

- Add `/etc/NetworkManager` to the list of persistent state paths.
- Update the installer so that instead of writing NM connection profiles to `/oem/90_custom.yaml`, it writes them directly to `/etc/NetworkManager/system-connections/*nmconnection`.
- On upgrade, instead of generating `/oem/91_networkmanager.yaml`, we again just write out NM connection profiles to the host in the (newly persistent) `/etc/NetworkManager/system-connections/` directory.

This makes subsequent network configuration changes really, really easy.

For example, if you want to change the DNS server, all you need to do is run these commands:

```
nmcli con modify bridge-mgmt ipv4.dns 8.8.8.8
nmcli device reapply mgmt-br
```

If you want to add a new slave to a bond interface, it's just:

```
nmcli con add type bond-slave ifname ens7 master mgmt-bo
```

There's no need to go messing around editing or adding connection profiles in `/oem/90_custom.yaml` or `/oem/91_networkmanager.yaml`, the above just works at runtime, and also the `nmcli` command will have already persisted these changes to the relevant files in `/etc/NetworkManager/system-connections/` so they'll be there after reboot.

### Describe alternatives you've considered

[...]

PR #9639, opened 2025-11-27
feat: persist NetworkManager connection profiles on upgrade

tserong:
#### Problem:
See https://github.com/harvester/harvester-installer/pull/1201

#### Solution:
See https://github.com/harvester/harvester-installer/pull/1201

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9581

#### Test plan:
- Install Harvester v1.6.0
- Upgrade to a version with this PR applied
- Make a network configuration change with `nmcli`
- Reboot, verify the change persists

Note: in order to test this, you need to be build harvester using the rancher/harvester-installer image that's built from https://github.com/harvester/harvester-installer/pull/1201. My process for local testing looks like this:

- Do a build of harvester-installer
- Tag your rancher/harvester-installer:master image, and push it to a local repo
- Tweak the installer FROM line in package/upgrade/Dockerfile to point to your harvester-installer image
- Build harvester
- Build the ISO from harvester-installer, making sure it pulls in your local harvester-upgrade image

(This is all a PITA, there must be a better way)

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -539,7 +539,7 @@ generate_networkmanager_config() {
     if [ -n "$current_paths_line" ]; then
       if ! [[ "$current_paths_line" =~ "/var/lib/NetworkManager" ]]; then
         echo "Adding /var/lib/NetworkManager to PERSISTENT_STATE_PATHS in $CUSTOM90_FILE"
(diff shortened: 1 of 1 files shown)


tserong:
Rebased

Vicente-Cheng (approved):

ibrokethecloud (approved):

PR #9643, opened 2025-11-27
fix(upgrade): only do early return when there is an error

FrankYang0529:
#### Problem:
The upgrade controller does early return in `cleanup` function for multiple nodes cluster. After `cleanup` function, the controller sets label `"harvesterhci.io/upgradeCleanup=Succeeded"` and it will not get into `cleanup` function in next reconciliation. This makes controller cannot update upgrade-log and re-enable descheduler addon.

#### Solution:
Only return when there is error.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8934

#### Test plan:
1. Create a 2-node v1.6.1 Harvester cluster.
2. Enable descheduler addon.
3. Upgrade to this PR.
4. Check the descheduler can be re-enable after upgrade.

https://github.com/harvester/harvester/pull/8941

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -560,7 +560,9 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) (
 				toUpdate.Annotations = make(map[string]string)
 			}
 			toUpdate.Annotations[longhornSettingsRestoredAnnotation] = strconv.FormatBool(true)
-			return h.upgradeClient.Update(toUpdate)
+			if upgrade, err = h.upgradeClient.Update(toUpdate); err != nil {
+				return upgrade, err
+			}
 		}
 	}
 
@@ -585,8 +587,8 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) (
 		}
 	}
 
-	if err := h.reenableAddons(upgrade); err != nil {
-		return nil, err
+	if upgrade, err = h.reenableAddons(upgrade); err != nil {
+		return upgrade, err
 	}
 	return upgrade, h.resumeManagedCharts()
 }
@@ -947,19 +949,19 @@ func (h *upgradeHandler) addUpgradeLabelToDeschedulerAddons(upgrade *harvesterv1
 	return nil
 }
 
-func (h *upgradeHandler) reenableAddons(upgrade *harvesterv1.Upgrade) error {
+func (h *upgradeHandler) reenableAddons(upgrade *harvesterv1.Upgrade) (*harvesterv1.Upgrade, error) {
 	reenableStr, ok := upgrade.Annotations[reenableDeschedulerAddonAnnotation]
 	if !ok {
-		return nil
+		return upgrade, nil
 	}
 
 	if reenableStr != "true" {
-		return nil
+		return upgrade, nil
 	}
 
 	addon, err := h.addonCache.Get(util.KubeSystemNamespace, util.DeschedulerName)
 	if err != nil && !apierrors.IsNotFound(err) {
-		return err
+		return upgrade, err
 	}
 
 	if addon != nil && !addon.Spec.Enabled {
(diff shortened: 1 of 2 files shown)


Vicente-Cheng (dismissed):
thanks for the quick fix

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -560,7 +560,9 @@ func (h *upgradeHandler) cleanup(upgrade *harvesterv1.Upgrade, cleanJobs bool) (
 				toUpdate.Annotations = make(map[string]string)
 			}
 			toUpdate.Annotations[longhornSettingsRestoredAnnotation] = strconv.FormatBool(true)
-			return h.upgradeClient.Update(toUpdate)
+			if upgrade, err = h.upgradeClient.Update(toUpdate); err != nil {
+				return nil, err
+			}
```
wheatdog, line 565: I found the review thread with @w13915984028 at the original PR: https://github.com/harvester/harvester/pull/8941/files#r2306489191

> @w13915984028: the toUpdate will be write back on L522, but there are other possible failures on L502, L512, L519, in those cases, LH setting is amended, but upgrade CR itself does not book it; in next retry, it will try to write to LH again if LH's value is different with current (e.g., set by user)
> 
> two options:
> (1) Write to upgrade CR on L494, and let next OnChange to run following upgradeLog, managedchart
> (2) Change L502, L512, L519, which will write back anyway even when error happens, but sill return the err, to let reconciller retry

> @wheatdog: Thanks @w13915984028 for pointing out this potential issue. I prefer the first option you proposed. For the behavior of wrangler reconciller, is there any material I can refer to? Or I should just dive into the code of wrangler directly.

The basic idea then is to write the upgrade CR back first with the annotations, and let the next OnChange to run the following upgradeLog, managedchart.

With the change from this PR, is there any chance / risk that they would be handled twice concurrently (by the current one and the next OnChange handle)?

wheatdog (commented):
Thanks for catching this! `harvesterhci.io/upgradeCleanup=Succeeded` was introduced in https://github.com/harvester/harvester/pull/9264 to address https://github.com/harvester/harvester/issues/6670. And the early return here (implemented in https://github.com/harvester/harvester/pull/8941) doesn't play well with the enhancement.

pkg/controller/master/upgrade/upgrade_controller.go
wheatdog: I briefly look up its behavior. I think the OnChange handler should handle it sequentially. It should be a safe change since the second handle should be blocked by the `"harvesterhci.io/upgradeCleanup=Succeeded"` annotation.

wheatdog (dismissed):

pkg/controller/master/upgrade/upgrade_controller.go
FrankYang0529: Hi @wheatdog, if we don't apply this change, the controller will set `harvesterhci.io/upgradeCleanup= Succeeded` and we don't have chance to do further operations.

https://github.com/harvester/harvester/blob/ca97f60367d0d9d1cd4f4cac6ce2882652a69868/pkg/controller/master/upgrade/upgrade_controller.go#L241-L255

For your concern, do you mean that: if L556 succeeds, but L563 failed. During this very small moment before next reconciliation, users update LH setting `replica-replenishment-wait-interval` again. After reconciliation, the value from user will be reset again?

https://github.com/harvester/harvester/blob/ca97f60367d0d9d1cd4f4cac6ce2882652a69868/pkg/controller/master/upgrade/upgrade_controller.go#L550-L565

wheatdog: I see. Would it make sense to remove `toUpdate := upgrade.DeepCopy()` in L555 and just reuse variable `upgrade` if we don't do early return?

wheatdog: My original concern is that the part below `the original early return line` might be called twice or even concurrently. But given that the OnChange handle change sequentially, I think it should be fine. 

https://github.com/harvester/harvester/blob/ca97f60367d0d9d1cd4f4cac6ce2882652a69868/pkg/controller/master/upgrade/upgrade_controller.go#L567-L586

FrankYang0529: > Would it make sense to remove toUpdate := upgrade.DeepCopy() in L555 and just reuse variable upgrade if we don't do early return?

I prefer to align the pattern with other places. Use `toUpdate` to assign updated value.

> My original concern is that the part below `the original early return line` might be called twice or even concurrently. But given that the OnChange handle change sequentially, I think it should be fine.

Yes, we also have a lock here, so I think this is fine.

https://github.com/harvester/harvester/blob/ca97f60367d0d9d1cd4f4cac6ce2882652a69868/pkg/controller/master/upgrade/upgrade_controller.go#L132-L133

wheatdog: I see. `upgrade` in L563 would be the updated `upgrade`.

wheatdog (dismissed):

Vicente-Cheng (approved):
thanks!

Vicente-Cheng:
@mergifyio backport v1.7

Issue #9604, opened 2025-11-24
[ENHANCEMENT] UI should able to attach RWX block volume for multi VMs


lanfon72:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

The enhancement comes from #8422, we did remove the restriction, but UI still have filter for volumes.

### Describe the solution you'd like

Let RWX volumes available in **Add Volume**

### Describe alternatives you've considered

Remove the filter. (which means for all volumes, could be selected in **Add Volume**

### Additional context

N/A

Vicente-Cheng:
cc @a110605, 
Could you help to relax this filter?

Thanks!

a110605:
After discussed with @lanfon72 , we will allow ALL volumes be able to to attach to multiple VMs from harvester UI. That the backend webhook to decide which volume can be attached. cc @Vicente-Cheng

lanfon72:
Verified this enhancement has been applied.

For 3rd party storage-class:
<img width="1470" height="529" alt="Image" src="https://github.com/user-attachments/assets/cacb503c-912a-440d-96d2-80f68cfcfecc" />

For built-in volumes (the volume been attached to other VM):
<img width="759" height="467" alt="Image" src="https://github.com/user-attachments/assets/16f86734-12ec-47ef-900c-57304732f976" />

Test Information
----
* Environment: **Baremetal DL360 3 nodes**
* Harvester Version: **v1.7.0-rc5**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Refer to https://github.com/harvester/harvester/issues/8422#issuecomment-3573143857 but test with UI.

PR #9646, opened 2025-11-27
Align the checking for addVolume operation

Vicente-Cheng:
#### Problem:
While UI relaxes the add volume filter with issue https://github.com/harvester/harvester/issues/9604, we need to align the webhook validator to have a better UX.

#### Solution:
Use the same validator for both webhook/api side

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9604

#### Test plan:
1. create 3 node cluster with v1.7.0-rc3
2. update the UI settings to use latest UI
3. create two VM and one volume
4. The addVolume can find the volume, but it would fail with the second try.

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -54,6 +54,7 @@ const (
 
 type vmformatter struct {
 	vmiCache      ctlkubevirtv1.VirtualMachineInstanceCache
+	vmCache       ctlkubevirtv1.VirtualMachineCache
 	pvcCache      ctlcorev1.PersistentVolumeClaimCache
 	nodeCache     ctlcorev1.NodeCache
 	scCache       ctlstoragev1.StorageClassCache
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/gorilla/mux"
+	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	longhorntypes "github.com/longhorn/longhorn-manager/types"
 	"github.com/pkg/errors"
@@ -53,7 +54,6 @@ import (
 	"github.com/harvester/harvester/pkg/util/drainhelper"
 	"github.com/harvester/harvester/pkg/util/virtualmachine"
 	"github.com/harvester/harvester/pkg/util/virtualmachineinstance"
-	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 )
 
 const (
@@ -1032,6 +1032,7 @@ func (h *vmActionHandler) checkAttachable(pvc *corev1.PersistentVolumeClaim) err
 }
 
 // addVolume add a hotplug volume with given volume source and disk name.
+// name -> VM name, namespace -> VM namespace, input.VolumeSourceName -> PVC name
 func (h *vmActionHandler) addVolume(ctx context.Context, namespace, name string, input AddVolumeInput) error {
 	// We only permit volume source from existing PersistentVolumeClaim at this moment.
 	// KubeVirt won't check PVC existence so we validate it on our own.
@@ -1041,6 +1042,10 @@ func (h *vmActionHandler) addVolume(ctx context.Context, namespace, name string,
 	}
 
 	if err := h.checkAttachable(pvc); err != nil {
+		return err
+	}
+
+	if err := virtualmachine.CheckBlockRWXVolumeForVM(h.pvcCache, h.storageClassCache, h.vmCache, namespace, input.VolumeSourceName, namespace, name); err != nil {
 		return err
 	}
 
--- a/pkg/util/virtualmachine/virtualmachine.go
+++ b/pkg/util/virtualmachine/virtualmachine.go
@@ -2,14 +2,24 @@ package virtualmachine
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
+	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	ctlstoragev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/storage/v1"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
+	"github.com/harvester/harvester/pkg/ref"
 	"github.com/harvester/harvester/pkg/util"
+	indexeresutil "github.com/harvester/harvester/pkg/util/indexeres"
+)
+
+var (
+	ErrVolumeIsUsedByOtherVM = fmt.Errorf("the volume is non-shareable and already used by other VMs")
 )
 
 // IsVMStopped checks VM is stopped or not. It will check two cases
@@ -115,7 +125,7 @@ func SupportHotUnplugNic(vm *kubevirtv1.VirtualMachine) (bool, error) {
 
 	errMsgs := make([]string, 0)
 	for _, iface := range ifaces {
-		ok, err := IsInterfaceHotUnpluggable(iface);
+		ok, err := IsInterfaceHotUnpluggable(iface)
 		if ok {
 			// as long as there is at least one hot-unpluggable interface
 			return true, nil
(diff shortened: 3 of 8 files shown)


pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -127,3 +132,45 @@ func SupportHotUnplugNic(vm *kubevirtv1.VirtualMachine) (bool, error) {
...
+	if volumeSupportRWXForVM(targetAccessMode, targetProvisioner) {
+		return true, nil
+	}
+	vms, err := vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(pvcNS, pvcName))
+	if err != nil {
+		return false, fmt.Errorf("failed to get VMs by index: %s, PVC: %s/%s, err: %s", indexeresutil.VMByPVCIndex, pvcNS, pvcName, err)
+	}
+	for _, otherVM := range vms {
+		if otherVM.Namespace != targetVMNS || otherVM.Name != targetVMName {
+			return true, fmt.Errorf("the volume %s is already used by VM %s/%s", pvcName, otherVM.Namespace, otherVM.Name)
```
ibrokethecloud: should return here not be `false`? 
based on the logic if the PVC is not RWX or Longhorn as checked in `volumeSupportRWXForVM` then it means PVC cannot be attached to any more VM's anyways

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -525,32 +524,11 @@ func (v *vmValidator) checkOccupiedPVCs(vm *kubevirtv1.VirtualMachine) error {
...
-			for _, otherVM := range vms {
-				if otherVM.Namespace != vm.Namespace || otherVM.Name != vm.Name {
-					message := fmt.Sprintf("the volume %s is already used by VM %s/%s", volume.PersistentVolumeClaim.ClaimName, otherVM.Namespace, otherVM.Name)
-					return werror.NewInvalidError(message, "spec.templates.spec.volumes")
-				}
+		if validated, err := vmutil.CheckBlockRWXVolumeForVM(v.pvcCache, v.scCache, v.vmCache, vm.Namespace, volume.PersistentVolumeClaim.ClaimName, vm.Namespace, vm.Name); err != nil {
+			if !validated {
+				return werror.NewInternalError(err.Error())
 			}
+			return werror.NewInvalidError(err.Error(), "spec.templates.spec.volumes")
```
WebberHuang1118: Can you explain more about the return values in `CheckBlockRWXVolumeForVM()`? It seems that the only meaningful distinction is whether the returned error is nil, and the `validated` flag is used solely to determine the final error message?

If I understand correctly, the `validated` return value does not seem strictly necessary?

WebberHuang1118 (commented):
Some questions, please take a look, thanks.

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -127,3 +132,45 @@ func SupportHotUnplugNic(vm *kubevirtv1.VirtualMachine) (bool, error) {
...
+		}
+	}
+	return true, nil
+}
```
ibrokethecloud:
```suggestion
var VolumeIsAlreadyInUse = errors.New("volume is already in use")
func CheckBlockRWXVolumeForVM(pvcCache v1.PersistentVolumeClaimCache, scCache ctlstoragev1.StorageClassCache, vmCache ctlkubevirtv1.VirtualMachineCache, pvcNS, pvcName, targetVMNS, targetVMName string) error {
	pvc, err := pvcCache.Get(pvcNS, pvcName)
	if apierrors.IsNotFound(err) {
		// means runtime creation, no need to check
		return nil
	}
	if err != nil {
		// any error here should be raised
		return fmt.Errorf("failed to get PVC %s/%s, err: %s", pvcNS, pvcName, err)
	}
	targetAccessMode := pvc.Spec.AccessModes
	targetProvisioner := util.GetProvisionedPVCProvisioner(pvc, scCache)
	if volumeSupportRWXForVM(targetAccessMode, targetProvisioner) {
		return nil
	}
	vms, err := vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(pvcNS, pvcName))
	if err != nil {
		return fmt.Errorf("failed to get VMs by index: %s, PVC: %s/%s, err: %s", indexeresutil.VMByPVCIndex, pvcNS, pvcName, err)
	}
	for _, otherVM := range vms {
		if otherVM.Namespace != targetVMNS || otherVM.Name != targetVMName {
			return fmt.Errorf("%s %w by VM %s/%s", pvcName, otherVM.Namespace, otherVM.Name)
		}
	}
	return nil
}
```
We can change the method signature to only return an error. 

If need to verify anything specific we can use `errors.Is` to identify if error is due to api call or because volume is in use else where

Based on the current logic in validator webhook, we may not need to even check error type since if there is no error.. we return else wrap and return the error

```
		name := volume.PersistentVolumeClaim.ClaimName
		pvc, err := v.pvcCache.Get(vm.Namespace, name)
			// means runtime creation, no need to check
			continue
		}
		if err != nil {
			// any error here should be raised
			return werror.NewInternalError(fmt.Sprintf("failed to get PVC %s/%s, err: %s", vm.Namespace, name, err))
		}
		targetAccessMode := pvc.Spec.AccessModes
		targetProvisioner := util.GetProvisionedPVCProvisioner(pvc, v.scCache)
		if volumeSupportRWXForVM(targetAccessMode, targetProvisioner) {
			continue
		}
		if volume.PersistentVolumeClaim != nil {
			vms, err := v.vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(vm.Namespace, volume.PersistentVolumeClaim.ClaimName))
			if err != nil {
				return werror.NewInternalError(fmt.Sprintf("failed to get VMs by index: %s, PVC: %s/%s, err: %s", indexeresutil.VMByPVCIndex, vm.Namespace, volume.PersistentVolumeClaim.ClaimName, err))
			}
			for _, otherVM := range vms {
				if otherVM.Namespace != vm.Namespace || otherVM.Name != vm.Name {
					message := fmt.Sprintf("the volume %s is already used by VM %s/%s", volume.PersistentVolumeClaim.ClaimName, otherVM.Namespace, otherVM.Name)
					return werror.NewInvalidError(message, "spec.templates.spec.volumes")
				}
			}
```

pkg/api/vm/handler.go
```diff
@@ -1044,6 +1045,10 @@ func (h *vmActionHandler) addVolume(ctx context.Context, namespace, name string,
 		return err
 	}
 
+	if _, err := virtualmachine.CheckBlockRWXVolumeForVM(h.pvcCache, h.storageClassCache, h.vmCache, namespace, input.VolumeSourceName, namespace, name); err != nil {
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -127,3 +132,45 @@ func SupportHotUnplugNic(vm *kubevirtv1.VirtualMachine) (bool, error) {
...
+
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -127,3 +132,45 @@ func SupportHotUnplugNic(vm *kubevirtv1.VirtualMachine) (bool, error) {
 
 	return false, fmt.Errorf("%s/%s doesn't support HotUnplugNic as none of its interfaces is hot-unpluggable: %s", vm.Namespace, vm.Name, strings.Join(errMsgs, ", "))
 }
+
+// CheckBlockRWXVolumeForVM checks whether the given PVC is a RWX volume (exclude Longhorn) and whether it is already used by other VMs.
```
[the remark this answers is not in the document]
Vicente-Cheng: no need

pkg/util/virtualmachine/virtualmachine.go
Vicente-Cheng: use new mechanism, no need

pkg/webhook/resources/virtualmachine/validator.go
Vicente-Cheng: I update the `CheckBlockRWXVolumeForVM()` as @ibrokethecloud suggested. please recheck

pkg/util/virtualmachine/virtualmachine.go
Vicente-Cheng: nice suggestion!
updated

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -525,32 +524,14 @@ func (v *vmValidator) checkOccupiedPVCs(vm *kubevirtv1.VirtualMachine) error {
...
-			}
-			for _, otherVM := range vms {
-				if otherVM.Namespace != vm.Namespace || otherVM.Name != vm.Name {
-					message := fmt.Sprintf("the volume %s is already used by VM %s/%s", volume.PersistentVolumeClaim.ClaimName, otherVM.Namespace, otherVM.Name)
-					return werror.NewInvalidError(message, "spec.templates.spec.volumes")
-				}
+		if err := vmutil.CheckBlockRWXVolumeForVM(v.pvcCache, v.scCache, v.vmCache, vm.Namespace, volume.PersistentVolumeClaim.ClaimName, vm.Namespace, vm.Name); err != nil {
+			if err == vmutil.ErrVolumeIsUsedByOtherVM {
+				return werror.NewInvalidError(
+					fmt.Sprintf("PVC %s/%s is already used by other VM, please use a RWX volume or Longhorn volume", vm.Namespace, volume.PersistentVolumeClaim.ClaimName),
```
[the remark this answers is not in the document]

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -127,3 +136,45 @@ func SupportHotUnplugNic(vm *kubevirtv1.VirtualMachine) (bool, error) {
...
+		// Longhorn provisioner does not support RWX volume for VM
+		return false
+	}
+
+	for _, mode := range accessModes {
+		if mode == corev1.ReadWriteMany {
+			return true
+		}
+	}
+	return false
```
brandboat: nit:

```suggestion
return slices.Contains(accessModes, corev1.ReadWriteMany)
```

ibrokethecloud (dismissed):
lgtm

pkg/api/vm/handler.go
```diff
@@ -1044,6 +1045,10 @@ func (h *vmActionHandler) addVolume(ctx context.Context, namespace, name string,
 		return err
 	}
 
+	if err := virtualmachine.CheckBlockRWXVolumeForVM(h.pvcCache, h.storageClassCache, h.vmCache, namespace, input.VolumeSourceName, namespace, name); err != nil {
+		return err
+	}
```
brandboat, line 1050: Just curious why we need to add a extra validation here? This will be check again in the validator, right?

pkg/util/virtualmachine/virtualmachine.go
Vicente-Cheng: nice suggestion, updated

brandboat (dismissed):
LGTM, thanks!

pkg/api/vm/handler.go
Vicente-Cheng: On the UI side, we update the annotation. But the original webhook would work after the controller syncs the annotation to the VMI template.

If we did not check on api action, user will see the operation success, but later VM status move to error (due to the validator)

brandboat: I see, thanks for the explanation!

pkg/api/vm/formatter.go
```diff
@@ -54,6 +54,7 @@ const (
 
 type vmformatter struct {
 	vmiCache      ctlkubevirtv1.VirtualMachineInstanceCache
+	vmCache       ctlkubevirtv1.VirtualMachineCache
```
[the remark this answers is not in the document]
Vicente-Cheng: we use it on handler.go

WebberHuang1118 (dismissed):
LGTM, thanks.

ibrokethecloud (dismissed):

brandboat (approved):

Vicente-Cheng:
rebased and added the missing indexer initializer

ibrokethecloud (approved):

PR #9661, opened 2025-12-01
Fix harvester crash when performing cdi vmimage download

brandboat:
#### Problem:

https://github.com/harvester/harvester/blob/master/pkg/controller/master/image/register.go#L28 send nil vmi cache and cause harvester crash when entering the retry mechanism implemented in https://github.com/harvester/harvester/commit/feef4a1e1e787da814af7570521a989b62745ff0

#### Test plan:
- test lhv2 download vmiage
- test lhv2 upload vmimage
截圖 2025-12-01 晚上11 35 43

### DIFF
--- a/pkg/api/image/schema.go
+++ b/pkg/api/image/schema.go
@@ -26,7 +26,10 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, _ config.Optio
 	vmImageDownloader := scaled.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineImageDownloader()
 	scClient := scaled.StorageFactory.Storage().V1().StorageClass()
 
-	vmio := common.GetVMIOperator(vmi, vmi.Cache(), http.Client{})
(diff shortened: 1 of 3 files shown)


WebberHuang1118:
Hi @brandboat

Maybe we can simply enforce that [GetVMIOperator()](https://github.com/harvester/harvester/blob/cd39687c3b1a95339b0a412068a8f145b7d0aa7f/pkg/image/common/operator.go#L98-L100) must not be called when the VMI cache is nil; in that case, it should just return an error.

With that mechanism in place, we can address the current issue with the one-line fix in [this line](https://github.com/harvester/harvester/blob/cd39687c3b1a95339b0a412068a8f145b7d0aa7f/pkg/controller/master/image/register.go#L28).

WebberHuang1118 (approved):
LGTM, thanks for the fix.

Vicente-Cheng (approved):

Issue #9664, opened 2025-12-03
[BUG] Harvester v1.7.0-rc4 cannot be installed in airgapped networks


brinnjoyce:
### Describe the Bug

Harvester v1.7.0-rc4 includes the `longhornio/longhorn-manager:v1.10.1` image - but the chart was updated to use `v1.10.1-hotfix-1` tag in https://github.com/harvester/harvester/commit/ddd4b324cbb32fa29ee961240eea3923f65b5413 

### To Reproduce

Try to install Harvester v1.7.0-rc4 on a host with no internet access

### Expected Behavior

All of the required images should be included in the ISO

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.7.0-rc4
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Lenovo SR665 V3
- Rancher version: 2.13.0

irishgordo:
@brinnjoyce **_thank you for reporting this 😄 🙏 - we really appreciate it!_**

I've been able to reproduce this and the entire v1.7.0-rc4 cluster in an airgapped environment, is indeed entirely dead in the water with the missing image.

**_I'm marking this as a regression._**

### screen grabs

<img width="1889" height="1961" alt="Image" src="https://github.com/user-attachments/assets/5c5c2189-58dd-4d50-a5b8-85f3f18fc2c4" />
<img width="2058" height="1948" alt="Image" src="https://github.com/user-attachments/assets/45d14fa4-a1d0-4b2d-ac7c-bd99c86d908b" />
<img width="2466" height="2021" alt="Image" src="https://github.com/user-attachments/assets/e11e37b2-e765-47f5-8549-969e7ba479ef" />
<img width="2480" height="2027" alt="Image" src="https://github.com/user-attachments/assets/ecf519f8-3f1d-43f0-89bb-0168322eb332" />

### support bundle

[broken-airgap-support-bundle-v170rc4.zip](https://github.com/user-attachments/files/23897872/broken-airgap-support-bundle-v170rc4.zip)

### support bundle methodology for loadout/acquisition/image-manually-sideloaded/standalone-mode

I got a support bundle by modifying this:
- https://github.com/rancher/support-bundle-kit/blob/master/docs/standalone.md

To this (support-bundle-manager.yaml):
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: support-bundle-manager
    rancher/supportbundle: sample
[...]

PR #9668, opened 2025-12-04
charts: add `repository` to patch image

Vicente-Cheng:
- that we can ensure the image can be packaged into ISO

#### Problem:
Missing the Longhorn hotfix image.

#### Solution:
Add the `repository` to patched image

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9664

#### Test plan:
Ensure we have `longhornio/longhorn-manager:v1.10.1-hotfix-1` on image list


Vicente-Cheng:
@mergifyio backport v1.7

starbops (approved):

Issue #9685, opened 2025-12-08
[ENHANCEMENT] Harvester VM webhook should skip check if VM is onDeleting


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

In general, the webhook skips `update` check when object is onDeleting

https://github.com/harvester/harvester/blob/51b8b75d5c1211a3cba535464b5fd0c39b96f8b4/pkg/webhook/resources/virtualmachine/mutator.go#L94-L100

but Harvester VM webhook is not adopting this way, it still checks VM spec even when VM is onDeleting

https://github.com/harvester/harvester/blob/51b8b75d5c1211a3cba535464b5fd0c39b96f8b4/pkg/webhook/resources/virtualmachine/validator.go#L241-L287

### Describe the solution you'd like

The check should skip when VM is onDeleting

### Additional context

v170 has an enhancement on webhook https://github.com/harvester/harvester/pull/9216

on v160, the related functions are:
https://github.com/harvester/harvester/blob/43d55d03c61c004b361576e6aa4096cfa3ec42eb/pkg/webhook/resources/virtualmachine/validator.go#L451-L468

At least, those checks on deleting VM make little sense:
```
checkOccupiedPVCs
checkReservedMemoryAnnotation
CheckIfVMCanStartByResourceQuota
checkStorageResourceQuota
```

https://github.com/harvester/harvester/issues/9500
_No response_

w13915984028:
close as we need to solve to root cause observed from https://github.com/harvester/harvester/issues/9761

PR #9686, opened 2025-12-08
Webhook skip checking VM when it is on deleting

w13915984028:
#### Problem:

Webhook check upon on deleting VM might cause unnecessary issues.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9685

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -240,7 +240,7 @@ func (v *vmValidator) Create(_ *types.Request, newObj runtime.Object) error {
 
 func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) error {
 	newVM := newObj.(*kubevirtv1.VirtualMachine)
-	if newVM == nil {
+	if newVM == nil || newVM.DeletionTimestamp != nil {
 		return nil
 	}
 
--- a/pkg/webhook/resources/virtualmachine/validator_test.go
+++ b/pkg/webhook/resources/virtualmachine/validator_test.go
@@ -226,6 +226,55 @@ func Test_virtualMachineValidator_duplicateMacAddress(t *testing.T) {
 			expectError: true,
 		},
 		{
+			name: "Update case - add a new interface to an existing vm with conflicting mac address, but vm is on deleting, returns success",
+			vm: &kubevirtv1.VirtualMachine{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:              "vm1",
+					Namespace:         "default",
+					DeletionTimestamp: &metav1.Time{},
+				},
+				Spec: kubevirtv1.VirtualMachineSpec{
+					Template: &kubevirtv1.VirtualMachineInstanceTemplateSpec{
+						Spec: kubevirtv1.VirtualMachineInstanceSpec{
+							Networks: []kubevirtv1.Network{
+								{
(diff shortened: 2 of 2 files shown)


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -240,7 +240,7 @@ func (v *vmValidator) Create(_ *types.Request, newObj runtime.Object) error {
 
 func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) error {
 	newVM := newObj.(*kubevirtv1.VirtualMachine)
-	if newVM == nil {
+	if newVM == nil || newVM.DeletionTimestamp != nil {
```
ihcsim, line 243: Generally speaking, I think as long as an object still exists, it shouldn't be allowed to bypass the validator, to prevent it from entering into an invalid state and/or trigger an invalid/malicious reconciliation on the controller side. E.g., if I add a custom finalizer to a VM and try to delete the VM, while the VM remains stuck in the terminating state indefinitely due to my custom finalizer, I can still modify the VM specs. Without the validator, some of these invalid inputs would have been persisted. (The good thing is that all of our VM controllers already skip resources with deletion timestamps.)

I think it is better to either keep the code as-is or return an error, instead of letting it bypass the validator. LMKWYT.

ihcsim: Following slack discussion, one thing we can do is to narrow the scope of the deletion timestamp conditional to just [`checkOccupiedPVCs()`](https://github.com/w13915984028/harvester/blob/01f573222d43ee9f8b8d026c69163b1ebe845c75/pkg/webhook/resources/virtualmachine/validator.go#L297-L299) i.e., don't check for contending VM, if the current VM is pending deletion.

WebberHuang1118: @ihcsim

For the original VM that holds the PVC, if it becomes unexpectedly stuck in deletion, would it be safe to bypass this check and allow the volume to be attached to a new VM?

If I understand correctly, in Longhorn this would trigger a volume live migration, which would never complete because the Kubernetes CSI resources for the original VM might not be torn down.

ihcsim: I am not familiar with the scenario you described. I tested VM deletion on Harvester side, and node deletion and node scale down from Rancher side. What I saw was, while the VM was pending deletion (blocked indefinitely by my custom finalizer), the volumes were either deleted immediately (for root disks like `disk-0`) or they remain attached to the VM (for non-root disks like `disk-1`, if user chooses to retain them). I can't reproduce a scenario where the VM was pending deletion, its volumes get detached completely and re-attached to another VM.

On Rancher side, the `MachineDeployment` would always create new volumes for new replacement machine replicas, instead of trying to detach/attach existing volumes. I think @w13915984028 mentioned about a machine image upgrade scenario; I haven't looked into that yet.

I also tested VM migration but that would only delete the pods, not the VM. The VM would not have a deletion timestamp, and any attempt to create a new VM to "hijack" the volumes would be blocked by the validators.

Overall, yes, we should investigate this further before making any code change.

w13915984028:
Note:  

This PR aims to solve https://github.com/harvester/harvester/issues/9761 before some puzzles are not fully figured out, now we will focus on the root cause of https://github.com/harvester/harvester/issues/9761 and fix it accordingly.

close this PR.

Issue #9666, opened 2025-12-04
[BUG] fqdn used after upgrade to 1.7.0-rc4, breaking etcd and harvester


abonillabeeche:
### Describe the Bug

After upgrading to 1.7.0-rc4, the system hostname changed from short-name to fqdn. 

### To Reproduce

Install 1.5.0 or a previous release using the host fqdn, in this case r640-1.5glinux.com then upgrade to 1.7.0-rc4, the fqdn will now be used instead of the short-name. Creating a new host, failing to load etcd and the hostname specific services.

> r640-1:~ # kubectl get nodes
> NAME                 STATUS     ROLES                       AGE     VERSION
> r640-1               NotReady   control-plane,etcd,master   239d    v1.34.2+rke2r1
> r640-1.5glinux.com   Ready      <none>                      6m33s   v1.34.2+rke2r1

### Expected Behavior

The same hostname should have been respected.

### Support Bundle for Troubleshooting
```
Dec 01 02:24:02 r640-1.5glinux.com rke2[3572]: time="2025-12-01T02:24:02Z" level=error msg="error syncing 'r640-1.5glinux.com': handler managed-etcd-metadata-controller: Internal error occurred: failed calling webhook \"validator.harvesterhci.io\": failed to call webhook: Post \"[https://harvester-webhook.harvester-system.svc:443/v1/webhook/validation?timeout=10s](https://harvester-webhook.harvester-system.svc/v1/webhook/validation?timeout=10s)\": no endpoints available for service \"harvester-webhook\", handler managed-etcd-controller: Internal error occurred: failed calling webhook \"validator.harvesterhci.io\": failed to call webhook: Post \"[https://harvester-webhook.harvester-system.svc:443/v1/webhook/validation?timeout=10s](https://harvester-webhook.harvester-system.svc/v1/webhook/validation?timeout=10s)\": no endpoints available for service \"harvester-webhook\", handler node: Internal error occurred: failed calling webhook \"validator.harvesterhci.io\": failed to call webhook: Post \"[https://harvester-webhook.harvester-system.svc:443/v1/webhook/validation?timeout=10s](https://harvester-webhook.harvester-system.svc/v1/webhook/validation?timeout=10s)\": no endpoints available for service \"harvester-webhook\", requeuing"
Dec 01 02:24:03 r640-1.5glinux.com rke2[3572]: time="2025-12-01T02:24:03Z" level=info msg="Reconciling ETCDSnapshotFile resources"
[...]

PR #9703, opened 2025-12-10
added an additional step to copy hostname into /oem/92_hostname_override.yaml

ibrokethecloud:
#### Problem:

Clusters installed prior to v1.6.0, where fqdn was used for node names, the hostname was set to only shortname, and k8s nodes were registered with the shortname.

Even though the harvester.config and generated cloudinit contained the fqdn this was ignored and nodes always ended up using the shortname.

The PR https://github.com/harvester/harvester-installer/pull/1011 introduced in v1.6.0 allowed fqdn to be used for k8s node names as well.

With the upgrade to NetworkManager in v1.7.x if a fqdn was used to install on a v1.5x or older install, post OS upgrade, the hostname changes from shortname to fqdn.

This results in another node with fqdn registering in k8s cluster and breaks the upgrade.

#### Solution:

Ensure original hostname is copied across to network stage to ensure hostname is not changed post switch to NetworkManager

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9666

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -570,6 +570,49 @@ generate_networkmanager_config() {
   # when the node comes back up after reboot, networking may be broken?
 }
 
+generate_hostname_persistance() {
+   if [ -z "$UPGRADE_PREVIOUS_VERSION" ]; then
+    detect_upgrade
+  fi
+
+  # NetworkManager is new in Harvester v1.7.0
+  # nodes installed via v1.5.x or lower which may have used fqdn hostnames
+  # only render short hostname in the k8s node name however the
+  # harvester.config and /oem/90_custom.yaml set hostname to fqdn 
+  # the /oem/90_custom.yaml sets hostname to fqdn as part which was ignored
+  # we need to ensure the hostname does not change after bump to NetworkManager
+  # as this will cause the node to be re-registered with apiserver using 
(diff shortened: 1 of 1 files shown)


package/upgrade/upgrade_node.sh
```diff
@@ -570,6 +570,42 @@ generate_networkmanager_config() {
...
+    echo "version: $UPGRADE_PREVIOUS_VERSION does not require generating Hostname override"
+    return
+  fi
+
+  # Just in case Hostname override has already been generated
+  # and/or potentially modified by the user, let's not overwrite it.
+  if [ -e ${HOST_DIR}/oem/92_hostname_override.yaml ]; then
+    echo "skipping hostname override config generation (${HOST_DIR}/oem/92_hostname_override.yaml already exists)"
+    return
+  fi
```
Yu-Jack, line 597: Should we also check whether the hostname from `hostname` command is equal to `/oem/harvester.config` or not? If it's the same, we can also skip creating cloud init.

Yu-Jack (commented):
Thanks! Just one question.

package/upgrade/upgrade_node.sh
ibrokethecloud: i have added an additional check

tserong (approved):

ihcsim (approved):

Yu-Jack (approved):

Issue #9614, opened 2025-11-25
[BUG] Click `View in API` of resource always direct to empty page


TachunLin:
### Describe the Bug

* On v1.7.0-rc3, after enabled the `View in API` feature in Preferences

<img width="1905" height="924" alt="Image" src="https://github.com/user-attachments/assets/a3a04c91-253f-4cab-992e-16e8339687c2" />

* When we click on the `View API` menu item of the selected item like virtual machine, image or vm network

* It always direct to an empty page

<img width="1249" height="590" alt="Image" src="https://github.com/user-attachments/assets/4cb1bb47-b731-4a6b-8837-b15a61cf7ae4" />

  https://github.com/user-attachments/assets/1ccf5161-e837-40d4-9ea2-a0063fae5452
  

### To Reproduce

1. Enable `View in API` feature in Preferences page
2. Create some resource like image, vm network and virtual machine
3. Click the `View API` 
4. Check the directed page content

### Expected Behavior

Should be able to display full API content after clicking the `View API` of the selected resource 

### Support Bundle for Troubleshooting

[supportbundle_fc2335e3-33ff-4aa9-9046-52fee5196224_2025-11-25T08-41-36Z.zip](https://github.com/user-attachments/files/23742391/supportbundle_fc2335e3-33ff-4aa9-9046-52fee5196224_2025-11-25T08-41-36Z.zip)

### Environment

- Harvester version: v1.7.0-rc3
- Underlying Infrastructure: Single node bare-metal machine

### Workaround and Mitigation

N/A

a110605:
Can't reproduce using 
- ui-index:  https://releases.rancher.com/harvester-ui/dashboard/release-harvester-v1.7/index.html
- ui-source : external

Please re-verify using v1.7.0-rc4

<img width="1470" height="881" alt="Image" src="https://github.com/user-attachments/assets/89111dff-12ed-4deb-bd13-d0b5e7eb74a0" />

TachunLin:
Verified **failed** on `v1.7.0-rc5`. Move to implement for investigation.

With the default setting couple with the release:
* ui-index: https://releases.rancher.com/harvester-ui/dashboard/release-harvester-v1.7/index.html
* ui-source: Auto


This issue still exists on `v1.7.0-rc5`. 

* Click the `View in API` of the image object, it shows empty page

[...]

PR #9711, opened 2025-12-11
fix(api-ui): add api-ui folder path

FrankYang0529:
#### Problem:
When clicking "View in API", the web page shows that `/api-ui/ui.min.js` is not found. This path is served as file server on `ui-path` setting (the default value is `/usr/share/harvester/harvester`). Files in this path are packaged in harvester image.

https://github.com/harvester/harvester/blob/0c47c9c8905e2e2982de564e762a444d58facf92/package/Dockerfile#L33-L42

I check the running container. There're files in `/usr/share/harvester/harvester`. However, the API shows 404 not found. That means the error is in server handler. The last change for `pkg/server/ui/ui.go` was three years ago. I'm not sure why it's broken in Harvester v1.7.0.

#### Solution:
When users get `/api-ui/ui.min.js`, the server handler only retrieves `ui.min.js`, not `api-ui/ui.min.js`. Change the API handler to serve files on `/usr/share/harvester/harvester/api-ui` to fix the error.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9614

#### Test plan:
Use Harvester dashboard to check "View in API" on any resource.

### DIFF
--- a/pkg/server/router.go
+++ b/pkg/server/router.go
@@ -94,7 +94,7 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
 	vueUI := ui.Vue
 	m.Handle("/dashboard/", vueUI.IndexFile())
 	m.PathPrefix("/dashboard/").Handler(vueUI.IndexFileOnNotFound())
-	m.PathPrefix("/api-ui").Handler(vueUI.ServeAsset())
+	m.PathPrefix("/api-ui").Handler(vueUI.ServeAsset("api-ui"))
 
 	if r.options.RancherURL != "" {
 		host, scheme, err := parseRancherServerURL(r.options.RancherURL)
--- a/pkg/server/ui/ui.go
+++ b/pkg/server/ui/ui.go
@@ -3,7 +3,6 @@ package ui
 import (
 	"crypto/tls"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


Yu-Jack:
interesting .. I tried my local built image base on master branch, it didn't show not found error. Does master also have this issue?

Vicente-Cheng:
Since the issue was targeted to v1.7.1, set DNM now

Yu-Jack:
maybe newer wrangler did something we don't know

FrankYang0529:
@Yu-Jack Did you try to access "View in API" or  `/api-ui/ui.min.js`? For "View in API", it doesn't show error, but the page is white without content. For `/api-ui/ui.min.js`, it shows 404 error.
Screenshot 2025-12-12 at 2 47 06 PM

Yu-Jack:
@FrankYang0529 In latest master ISO, it access `https://releases.rancher.com/api-ui/1.1.9/ui.min.js`. Is it the same thing as you mentioned about "View in API"?

image

image

Yu-Jack:
What's your ui-source/index/path setting? All are default?

FrankYang0529:
> In latest master ISO, it access `https://releases.rancher.com/api-ui/1.1.9/ui.min.js`. Is it the same thing as you mentioned about "View in API"?

No, it's different. The `/api-ui/ui.min.js` is file from local harvester image. The `https://releases.rancher.com/api-ui/1.1.9/ui.min.js` is external resource. Could you change the `ui-source` setting to `Bundled` and try again? Thanks.

Yu-Jack:
But, in my 170rc-4, it doesn't work like you mentioned.

Yu-Jack:
Okay, after using `ui-source: bundled`, it also shows white page.

Yu-Jack (approved):
Thanks!

Issue #9630, opened 2025-11-26
[BUG] Can not hot-plugging on a VM has management network with bridge-bond and virtio


albinsun:
### Describe the Bug

Hit webhook error when hot-plugging on a VM has `management network` with `bridge`-bond and `virtio`
Note that when _nic0_ is non-`management` Networks like `L2VlanNetwork`, it's ok to hot-plugging NIC.

### To Reproduce

1. Create a VM with `management network` using `bridge`-bond and `virtio`
   <img width="1565" height="772" alt="Image" src="https://github.com/user-attachments/assets/ba3e212d-dc90-48f3-a789-27595f8bddbe" />

1. Hot-plug a L2VlanNetwork
   <img width="1565" height="772" alt="Image" src="https://github.com/user-attachments/assets/46168205-ffce-4d38-b91c-a35f6f51ea06" />
   Denied by `{"data":"admission webhook \"migration-create-validator.kubevirt.io\" denied the request: Cannot migrate VMI, Reason: InterfaceNotLiveMigratable, Message: cannot migrate VMI which does not use masquerade, bridge with kubevirt.io/allow-pod-bridge-network-live-migration VM annotation or a migratable plugin to connect to the pod network"}`
   <img width="1565" height="772" alt="Image" src="https://github.com/user-attachments/assets/9490649f-d199-4b87-b3b5-6c173ec02e85" />
   But it do configured
   <img width="1565" height="772" alt="Image" src="https://github.com/user-attachments/assets/97868f1b-2f08-4787-b544-c5a5a08d260a" />

1. Try to Detach
   Denied by `admission webhook "migration-create-validator.kubevirt.io" denied the request: Cannot migrate VMI, Reason: InterfaceNotLiveMigratable, Message: cannot migrate VMI which does not use masquerade, bridge with kubevirt.io/allow-pod-bridge-network-live-migration VM annotation or a migratable plugin to connect to the pod network`
   <img width="1565" height="772" alt="Image" src="https://github.com/user-attachments/assets/faefe108-39b5-4c32-a633-91abb47301a8" />

### Expected Behavior

Do we allow VM attach a `management network` with `bridge`-bond and `virtio`?
* If yes, then this issue is a bug of NIC hot-plugging.
* If No, perhaps need to drop `bridge` type when **Network** is `management`.

### Support Bundle for Troubleshooting

[...]

PR #9713, opened 2025-12-11
fix: should also check KubeVirt's condition for LiveMigratable

wheatdog:
#### Problem:

VMs with bridge-bonded management network should not be able to be migrated/hotplugged.

#### Solution:

Since KubeVirt provides the LiveMigratable condition in the status of VM/VMI CRD, we should check it in `ValidateVMMigratable`.

#### Related Issue(s):

#9630 

#### Test plan:

- Follow https://github.com/harvester/harvester/issues/9630#issue-3667712868, and the both actions `Hotplug Network Interface` and `Migrate` shouldn't be there.
- Follow the recording in https://github.com/harvester/harvester/issues/9630#issuecomment-3584430925, and the both actions `Hotplug Network Interface` and `Migrate` shouldn't be there.

https://github.com/harvester/harvester/issues/9630#issuecomment-3584564806

### DIFF
--- a/pkg/util/virtualmachineinstance/virtualmachineinstance.go
+++ b/pkg/util/virtualmachineinstance/virtualmachineinstance.go
@@ -99,6 +99,13 @@ func ValidateVMMigratable(vmi *kubevirtv1.VirtualMachineInstance) error {
 		return fmt.Errorf("VM %s is not live migratable as CD-ROM or container disk is set", vmiNamespacedName)
 	}
 
+	// Lastly, check the condition reported by KubeVirt
+	for _, cond := range vmi.Status.Conditions {
+		if cond.Type == kubevirtv1.VirtualMachineInstanceIsMigratable && cond.Status == corev1.ConditionFalse {
+			return fmt.Errorf("VM %s is not live migratable as the condition reported with reason: %s", vmiNamespacedName, cond.Reason)
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/util/virtualmachineinstance/virtualmachineinstance.go
```diff
@@ -99,6 +99,13 @@ func ValidateVMMigratable(vmi *kubevirtv1.VirtualMachineInstance) error {
 		return fmt.Errorf("VM %s is not live migratable as CD-ROM or container disk is set", vmiNamespacedName)
 	}
 
+	// Lastly, check the condition reported by KubeVirt
+	for _, cond := range vmi.Status.Conditions {
+		if cond.Type == kubevirtv1.VirtualMachineInstanceIsMigratable && cond.Status == corev1.ConditionFalse {
```
w13915984028: when the target condition is found, return error or nil, the loop can be finished; and also could you add a test code to cover this case? thanks.

wheatdog: Sure. Sounds good!

wheatdog: fixed by https://github.com/harvester/harvester/pull/9713/commits/cb6542249871c1652fdd6a1a14eaf2c604f2ad12

w13915984028 (approved):
LGTM, thanks.

Yu-Jack (approved):
Thanks!

Vicente-Cheng:
Hold this PR for a while, we need to discuss the scope

Vicente-Cheng:
keep DNM until we release v1.7.0

wheatdog:
Since we have released v1.7.0, I am going to merge this PR back to master and open a backport PR for v1.7.1.

Issue #9260, opened 2025-10-03
[BUG] Upgrade from v1.6.x to v1.7.x may result in host IP address change when using DHCP


tserong:
### Describe the Bug

Harvester v1.7.0 uses NetworkManager instead of Wicked as used in earlier versions (see https://github.com/harvester/harvester/issues/3418). These two network stacks have different defaults for DHCP client IDs, which means that after upgrading, the DHCP server may return a different IP address. If this happens, the host in question will be unable to join the cluster because its IP address has changed.

This problem will not occur if your DHCP server is configured to allocate fixed IP addresses based on MAC address, as is done in https://github.com/harvester/ipxe-examples. It will occur however if the DHCP server is allocating IP addresses based solely on DHCP client ID.

### To Reproduce

- Deploy a single node Harvester v1.6.0 in a libvirt VM, configured to use DHCP. Once deployed, check the IP address of the host.
- Upgrade to latest v1.7.x dev version (note: this requires https://github.com/harvester/harvester/pull/9200).
- Once upgrade is complete, check the IP address of the host again. It will almost certainly have changed (e.g. initially it may have been 192.168.122.247, but after upgrade it will be 192.168.122.248)

### Expected Behavior

The IP address of the host remains the same as it was before the upgrade.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.6.0/v1.7.0-dev-*
- Impacted VM: N/A
- Impacted volume (PV): N/A
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): All
- Rancher version: N/A

### Additional context

See https://github.com/harvester/harvester/pull/9200#issuecomment-3364308840 for some investigation

### Workaround and Mitigation

Workarounds include:
- Configure the DHCP server to allocate IP addresses based on MAC address, not client ID (we could, and probably should, document this)
- Force the upgraded host to use the same client ID as was used before- `grep client-id /var/lib/wicked/lease-mgmt-br-dhcp-ipv4.xml` then apply that specific value e.g.: `nmcli con modify bridge-mgmt ipv4.dhcp-client-id "ff:00:90:8d:45:00:01:00:01:30:72:1e:f6:52:54:00:90:8d:45" && nmcli device reapply mgmt-br` (hopefully we can automate this as part of the upgrade)

Vicente-Cheng:
Hi @tserong, 
Before we fixed this issue, could we have the document for the workaround first?

Thanks!

[...]

PR #9732, opened 2025-12-12
fix: Update DHCP client ID on upgrade

tserong:
#### Problem:
Harvester v1.7.x uses NetworkManager instead of Wicked as used in earlier versions.  These two network stacks have different defaults for DHCP client IDs, which means that after upgrading, the DHCP server may return a different IP address.  If this happens, the host in question will be unable to join the cluster because its IP address has changed.

#### Solution:
This commit extracts the wicked DHCP client ID on upgrade and injects it into the appropriate NetworkManager connection profile to avoid the above problem.  If the system isn't configured for DHCP or if the wicked lease file can't be found, or if there's somehow more than one wicked lease file, no action will be taken.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9260

#### Test plan:
Perform the following steps either on hosts in our lab, or manually using VMs created with virt-manager/libvirt. Don't use ipxe-examples, because in that case the DHCP server allocates fixed IP addresses based on MAC address and you won't hit the problem.

1. Demonstrate the problem:
    - Install Harvester v1.6.x, using DHCP for IP address. A single node is fine. Make a note of the host's IP address.
    - Upgrade to Harvester v1.7.0. Confirm the IP address of the host has changed and Harvester doesn't start correctly.
2. Verify the fix:
    - Install Harvester v1.6.x again. Make at note of the IP address. It may be different again from what it was before, because this is a new install.
    - Update to Harvester v1.7.1 with this fix. Confirm that the IP address does _not_ change this time, and that Harvester starts correctly.

If you're doing this in VMs you can take a snapshot after the initial v1.6.x install, then revert to that rather than re-installing to save a bit of time.

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -561,6 +561,31 @@ generate_networkmanager_config() {
   # better to get stuck here in that way?  Or would it be better to barrel
   # on regardless and continue the OS upgrade with the knowledge that
   # when the node comes back up after reboot, networking may be broken?
+
+  # In case the system is configured for DHCP, we need to try to keep the
+  # previous DHCP client ID to ensure that IP addresses don't change on
+  # upgrade.  If we're using DHCP, there will be exactly one NM connection
+  # profile which includes the string "method=auto" (it will be either
+  # bridge-mgmt or vlan-mgmt), and there will also be a wicked DHCP lease
+  # file.  To be careful, we assign to arrays here to be sure that there's
+  # only one matching NM connection profile, and one wicked lease file.
+  # If either of these things are not true, we do nothing.
+  local dhcp_connection_profiles=($(grep -l '^method=auto$' ${HOST_DIR}/usr/local/.state/etc-NetworkManager.bind/system-connections/*nmconnection 2>/dev/null))
+  local wicked_leases=($(ls ${HOST_DIR}/var/lib/wicked/lease*xml 2>/dev/null))
+  if [ ${#dhcp_connection_profiles[@]} -eq 1 ] && [ ${#wicked_leases[@]} -eq 1 ]; then
+    echo "Found wicked lease and DHCP connection profile, attempting to update DHCP client ID"
+    local dhcp_client_id=$(yq '.lease."ipv4:dhcp".client-id // ""' ${wicked_leases[0]})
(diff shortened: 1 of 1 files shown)


tserong:
Rebased and fixed codefactor complaint.

package/upgrade/upgrade_node.sh
```diff
@@ -561,6 +561,31 @@ generate_networkmanager_config() {
...
+  # profile which includes the string "method=auto" (it will be either
+  # bridge-mgmt or vlan-mgmt), and there will also be a wicked DHCP lease
+  # file.  To be careful, we assign to arrays here to be sure that there's
+  # only one matching NM connection profile, and one wicked lease file.
+  # If either of these things are not true, we do nothing.
+  local dhcp_connection_profiles=($(grep -l '^method=auto$' ${HOST_DIR}/usr/local/.state/etc-NetworkManager.bind/system-connections/*nmconnection 2>/dev/null))
+  local wicked_leases=($(ls ${HOST_DIR}/var/lib/wicked/lease*xml 2>/dev/null))
+  if [ ${#dhcp_connection_profiles[@]} -eq 1 ] && [ ${#wicked_leases[@]} -eq 1 ]; then
+    echo "Found wicked lease and DHCP connection profile, attempting to update DHCP client ID"
+    local dhcp_client_id=$(yq '.lease."ipv4:dhcp".client-id' ${wicked_leases[0]})
```
w13915984028: is here
```
if [[ "$dhcp_client_id" != "null" ]]; then
```

//# local var escaps `Error: no matches found`, and return value is `null` if not found the key on yaml

w13915984028 (dismissed):
LGTM, thanks.

package/upgrade/upgrade_node.sh
tserong: Good catch, I'll update to deal with that possibility.

tserong: I never really liked how `yq` would return the string `null` in this case, so I've changed it to `yq '.lease."ipv4:dhcp".client-id // ""'`. This makes it return an empty string in case the value isn't found.

tserong:
Rebased

Vicente-Cheng (approved):
lgtm, thanks for the fix!

PR #9769, opened 2025-12-19
GHA: add workflow_dispatch

Vicente-Cheng:
- we can build any version branch just as we want

#### Problem:
We cannot build version-head branch easily

#### Solution:
add workflow_dispatch then we can build any version-branch just as we want


starbops (approved):
LGTM, thanks.

ibrokethecloud (approved):

PR #9772, opened 2025-12-20
GHA: use the simple way for trigger workflow

Vicente-Cheng:
Since we have many steps will reference the ref/head,
    we should directly let workflow_dispatch on each target branch.

    Then we can ensure we can get the correct commit.

    Revert the follows:
    Revert "GHA: add workflow_dispatch"

    This reverts commit c12f43a5cce36655231c036b050972852064c162.

    Revert "GHA: correct the workflow case for fetch the sha and branch"

    This reverts commit 56bdc80410c65a65766d1683dfc3bd9d78ac4f84.

#### Problem:
Use the wrong branch when building ISO through workflow_dispatch

#### Solution:
Simplify the mechanism, we can backport the target branch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,54 +9,22 @@ on:
   schedule:
     - cron: "10 0,12 * * *"
   workflow_dispatch:
-    inputs:
-      target_branch:
-        description: 'Target branch (must be v1.x format or master)'
-        required: true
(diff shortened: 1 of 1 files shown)


wheatdog (approved):

Vicente-Cheng:
@mergifyio backport v1.7

Issue #9435, opened 2025-11-06
[BUG] When VM is `Pending migration`, the `abortMigration` action is not allowed


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

In some cases like the resourcequota does not allow VM to be migrated, the `abortMigration` does not work

Refer: https://github.com/w13915984028/harvester-develop-summary/blob/main/ts-ref/v170-vm-migration-with-resource-quota-and-overhead-ratio.md#tests-2-single-vm-with-overhead-ratio-15-but-changed-when-vm-is-running

<img width="2438" height="518" alt="Image" src="https://github.com/user-attachments/assets/7d78d9e4-e829-4124-93c0-e46385102688" />

The error comes from below code:

https://github.com/harvester/harvester/blob/363164e7ed3665333808272fc12bfb4fbba816a4/pkg/api/vm/handler.go#L524-L526

You can only run kubectl to delete the `pending` vmim instance.

```
NAME        PHASE       VMI
vm1-f2sxv   Succeeded   vm1
vm1-sm587   Pending     vm1
```

### Describe the solution you'd like

Allow `abortMigration` in this case.

### Describe alternatives you've considered

Run kubectl to delete the `pending` vmim instance.

innobead:
I believe this is a bug rather than an enhancement.

TachunLin:
Verified **fixed** on v1.8.0 (master-8639cfb1-head). Close this issue.

### Result

<details><summary>$\color{green}{\textsf{PASS}}$ Can abort migration action when vm in Pending migration state (using overloaded migration list) $~~$</summary>

  https://github.com/user-attachments/assets/c49866f6-df11-4e50-9d67-f031b1a83262
    

 </details>


<details><summary>$\color{green}{\textsf{PASS}}$ Can abort migration action when vm in Pending migration state (using overhead ratio) $~~$</summary>

  https://github.com/user-attachments/assets/954ae536-43d1-424b-8b5a-7c5e9e183179
  

 </details>



### Test Information
* Test Environment: Two nodes harvester on remote kvm machine
* Harvester version: `master-8639cfb1-head`

### Verify Steps

<details><summary> Abort vm in pending migration using overloaded migration list </summary>

  Refer to: https://github.com/harvester/harvester/pull/9784#issue-3756739228

</details>


<details><summary> Abort vm in pending migration using overhead ratio </summary>

[...]

PR #9784, opened 2025-12-23
Allow aborting of pending migrations

martindekov:
Allowing aborting migrations in pending state. As well as kubevirt's definition of running migration. Covered the addition with test case.

#### Problem:
Migration stuck in `Pending` state cannot be aborted.

#### Solution:
Allow aborting migration in that state by adding it as addition to the check.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9435

#### Test plan:

<details>
  <summary>1. On 2 node harvester cluster each node built with this change:</summary>

image
</details>

2. Created 3 VMs on `node-1`

<details>
  <summary>3. Migrated all 3 on `node-2` where we allow only 2 migrations at once so 1 was stuck in Pending state</summary>

image
</details>

<details>
  <summary>4. Was able to abort migration of the third VM which was pending migration of the other 2</summary>

image
</details>

N/A

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -552,16 +552,18 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
 }
 
 func isVmimAbortable(vmim *kubevirtv1.VirtualMachineInstanceMigration) bool {
-	if vmim.Status.Phase == kubevirtv1.MigrationPhaseUnset ||
-		vmim.Status.Phase == kubevirtv1.MigrationPending ||
-		vmim.Status.Phase == kubevirtv1.MigrationScheduling ||
-		vmim.Status.Phase == kubevirtv1.MigrationScheduled ||
-		vmim.Status.Phase == kubevirtv1.MigrationPreparingTarget ||
-		vmim.Status.Phase == kubevirtv1.MigrationTargetReady ||
-		vmim.Status.Phase == kubevirtv1.MigrationRunning {
+	switch vmim.Status.Phase {
+	case kubevirtv1.MigrationPhaseUnset,
+		kubevirtv1.MigrationPending,
+		kubevirtv1.MigrationScheduling,
+		kubevirtv1.MigrationScheduled,
+		kubevirtv1.MigrationPreparingTarget,
+		kubevirtv1.MigrationTargetReady,
+		kubevirtv1.MigrationRunning:
 		return true
+	default:
+		return false
 	}
-	return false
 }
 
 func (h *vmActionHandler) findMigratableNodes(rw http.ResponseWriter, namespace, name string) error {
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -538,7 +538,7 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
 	migrationUID := getMigrationUID(vmi)
 	for _, vmim := range vmims {
 		if migrationUID == string(vmim.UID) {
-			if !vmim.IsRunning() && vmim.Status.Phase != kubevirtv1.MigrationPending {
+			if !isVmimAbortable(vmim) {
 				return fmt.Errorf("cannot abort the migration as it is in %q phase", vmim.Status.Phase)
 			}
 			// Migration is aborted by deleting the VMIM object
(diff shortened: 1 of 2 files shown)


pkg/api/vm/handler.go
```diff
@@ -538,7 +538,7 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
 	migrationUID := getMigrationUID(vmi)
 	for _, vmim := range vmims {
 		if migrationUID == string(vmim.UID) {
-			if !vmim.IsRunning() {
+			if !vmim.IsRunning() && vmim.Status.Phase != kubevirtv1.MigrationPending {
```
w13915984028: It is a bit complex:

kubevirt definition:

```
func (m *VirtualMachineInstanceMigration) IsRunning() bool {
	switch m.Status.Phase {
	case MigrationFailed, MigrationPending, MigrationPhaseUnset, MigrationSucceeded, MigrationWaitingForSync, MigrationSynchronizing:
		return false
	}
	return true
}
```

Harvester summarized condition

https://github.com/harvester/harvester/blob/4ac9c4153de37fc01c99146e9b9a87169db4b6eb/pkg/controller/master/migration/vmim_controller.go#L50-L63

Namely, Harvester allow to abort when the vmim in following phase, we could encap a function e.g. IsVmimPendingOrRunning(vmim)

```
kubevirtv1.MigrationPhaseUnset 
kubevirtv1.MigrationPending
kubevirtv1.MigrationScheduling
kubevirtv1.MigrationScheduled
kubevirtv1.MigrationPreparingTarget
kubevirtv1.MigrationTargetReady
kubevirtv1.MigrationRunning
```
And here, check ` !IsVmimPendingOrRunning(vmim)`

pkg/api/vm/handler_test.go
```diff
@@ -422,6 +422,47 @@ func TestAbortMigrateAction(t *testing.T) {
...
+					Status: kubevirtv1.VirtualMachineInstanceMigrationStatus{
+						Phase: kubevirtv1.MigrationPending,
+					},
+				},
+			},
+			expected: output{
+				vmInstanceMigrations: []*kubevirtv1.VirtualMachineInstanceMigration{},
+				err:                  nil,
+			},
+		},
```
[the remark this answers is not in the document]
martindekov: Test added

pkg/api/vm/handler.go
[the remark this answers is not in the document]
martindekov: All migration states are in the body of a new function in harvester explicitly stated

martindekov: Thanks for the suggestion Jian, good one, now all the Migration states are in the body of a new function `isVmimAbortable` which is the only difference to what you said the name should be - `IsVmimPendingOrRunning` as I think it would be easier for a reader to go into the body to understand what phase we allow to be aborted. I will leave the discussion open it if makes sense or for further comments on it

pkg/api/vm/handler.go
```diff
@@ -551,6 +551,19 @@ func (h *vmActionHandler) abortMigration(namespace, name string) error {
 	return nil
 }
 
+func isVmimAbortable(vmim *kubevirtv1.VirtualMachineInstanceMigration) bool {
+	if vmim.Status.Phase == kubevirtv1.MigrationPhaseUnset ||
+		vmim.Status.Phase == kubevirtv1.MigrationPending ||
```
w13915984028: The `switch...case...` might be a bit better than `if` here.

pkg/api/volume/handler.go
```diff
@@ -100,7 +100,7 @@ func (h *ActionHandler) assertPVCNotInUse(pvcNamespace, pvcName string) error {
 	// find any pod use this PVC (same validation on CDI)
 	index := fmt.Sprintf("%s-%s", pvcNamespace, pvcName)
 	if pods, err := h.pods.GetByIndex(util.IndexPodByPVC, index); err == nil && len(pods) > 0 {
-		podList := []string{}
+		podList := make([]string, 0, len(pods))
```
w13915984028, line 103: please check the `master-head` code to see if rebase is required, as https://github.com/harvester/harvester/pull/9851 has been merged.

w13915984028 (commented):
Please check those comments, thanks.

pkg/api/volume/handler.go
martindekov: Did the rebase and accepted the current changes - e.g. the changes from this: https://github.com/harvester/harvester/pull/9851

pkg/api/vm/handler.go
martindekov: Now switch case is used as opposed to if and ||

w13915984028 (approved):
LGTM, thanks.

Issue #9720, opened 2025-12-11
[BUG] Can't completely delete airgapped upgrade image when clicking delete button, prompt error message


TachunLin:
### Describe the Bug

On `v1.7.0-rc5`, after we upload a Harvester image to create the airgapped upgrade image.

* When we click the delete button and try to delete the upgrade image. 
  <img width="1674" height="677" alt="Image" src="https://github.com/user-attachments/assets/d3bc5574-57dd-4227-8419-fe262bc41a38" />
  
* It would prompt error message 
  ```
  admission webhook "validator.harvesterhci.io" denied the request: storage class harvester-v1.7.0-rc5-amd64.iso is used by virtual machine images: [image-q9sfz]
  ```
  
  <img width="1665" height="727" alt="Image" src="https://github.com/user-attachments/assets/0edbf983-9df4-49a5-8265-9ae5a24e68d0" />
  
* Click the delete image again will prompt another message
  ```
  virtualmachineimages.harvesterhci.io "image-q9sfz" not found
  ```
  <img width="599" height="226" alt="Image" src="https://github.com/user-attachments/assets/956e94e0-f4c0-4a8b-b632-5d6e4c2bdf98" />
  
* But actually the upgrade image have been deleted, but the storage class remains

* full video clip

  https://github.com/user-attachments/assets/35dc2395-6bd6-47b0-82f2-e5e6b4e18fb7

### To Reproduce

1. Open server-version setting -> Upgrade 
2. Upload a Harvester ISO to create the airgapped upgrade image
3. Wait for the image upload complete
4. Switch to `Delete Existing Image` 
5. Find the create image, system will prompt message to confirm deletion
6. Click the Delete button

### Expected Behavior

When click the delete button should be able to delete the airgapped upgrade image from list without error message like image is used or not found. 

It would be better if we can also delete the corresponding storage class. 

### Support Bundle for Troubleshooting

[supportbundle_e5a32eda-2b49-424c-95c4-57b2ff51bb81_2025-12-11T08-17-54Z.zip](https://github.com/user-attachments/files/24098177/supportbundle_e5a32eda-2b49-424c-95c4-57b2ff51bb81_2025-12-11T08-17-54Z.zip)

### Environment

- Harvester version: v1.7.0-rc5
- Underlying Infrastructure: Three nodes bare-metal machines

### Additional context

When we tried to delete the storage class of the upgrade image first then delete the image.

* It also prompt error message to prevent to proceed. 
  ```
[...]

PR #9785, opened 2025-12-24
Ignore deleting vmimage in storage class validator

brandboat:
#### Problem:

StorageClass validator blocks the storage class deletion even if its related vm image is marked as deleting.

This is a problem when we perform the following operation sequentially:
1. delete vmimage
2. delete the storage class that relates to the vmimage

In step 1 k8s marks vmimage as deleting, but the vm image has not been fully removed yet, then step 2 fails because the storage class validate treat it as still in-use.

#### Solution:

ignore deleting vm images in storage class validator

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9720

#### Test plan:
well-described in issue description, see https://github.com/harvester/harvester/issues/9720

https://github.com/user-attachments/assets/26833d86-7d30-48de-aa70-339edc0c9186

### DIFF
--- a/pkg/webhook/resources/storageclass/validator.go
+++ b/pkg/webhook/resources/storageclass/validator.go
@@ -322,7 +322,9 @@ func (v *storageClassValidator) validateVMImageUsage(sc *storagev1.StorageClass)
 
 	usedVMImages := make([]string, 0, len(vmimages))
 	for _, vmimage := range vmimages {
-		usedVMImages = append(usedVMImages, vmimage.Name)
+		if vmimage.DeletionTimestamp == nil {
+			usedVMImages = append(usedVMImages, vmimage.Name)
+		}
 	}
 
 	if len(usedVMImages) > 0 {
--- a/pkg/webhook/resources/storageclass/validator_test.go
+++ b/pkg/webhook/resources/storageclass/validator_test.go
@@ -408,6 +408,123 @@ func Test_storageClassValidator_validateEncryption(t *testing.T) {
 	}
 }
 
+func Test_storageClassValidator_validateVMImageUsage(t *testing.T) {
+	now := metav1.Now()
+
+	tests := []struct {
+		name                 string
+		storageClass         *storagev1.StorageClass
+		vmImages             []*v1beta1.VirtualMachineImage
+		expectError          bool
+		expectedErrorMessage string
+	}{
+		{
+			name: "no VM images using the storage class",
+			storageClass: &storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "unused-storage-class",
+				},
+			},
+			vmImages:    []*v1beta1.VirtualMachineImage{},
+			expectError: false,
+		},
+		{
+			name: "storage class in use by active VM image",
+			storageClass: &storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-storage-class",
+				},
+			},
+			vmImages: []*v1beta1.VirtualMachineImage{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-vmimage",
+						Namespace: "default",
+						Annotations: map[string]string{
+							util.AnnotationStorageClassName: "test-storage-class",
+						},
+					},
+				},
+			},
+			expectError: true,
+		},
+		{
+			name: "storage class with VM image being deleted should not block deletion",
+			storageClass: &storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-storage-class",
+				},
+			},
+			vmImages: []*v1beta1.VirtualMachineImage{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:              "deleting-vmimage",
+						Namespace:         "default",
+						DeletionTimestamp: &now,
+						Annotations: map[string]string{
+							util.AnnotationStorageClassName: "test-storage-class",
+						},
+					},
+				},
+			},
+			expectError: false,
+		},
+		{
+			name: "storage class with mix of active and deleting VM images",
+			storageClass: &storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-storage-class",
+				},
+			},
+			vmImages: []*v1beta1.VirtualMachineImage{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "active-vmimage",
+						Namespace: "default",
+						Annotations: map[string]string{
+							util.AnnotationStorageClassName: "test-storage-class",
+						},
+					},
+				},
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:              "deleting-vmimage",
+						Namespace:         "default",
+						DeletionTimestamp: &now,
+						Annotations: map[string]string{
+							util.AnnotationStorageClassName: "test-storage-class",
+						},
+					},
+				},
+			},
+			expectError: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			var vmImageObjects []runtime.Object
+			for _, img := range tc.vmImages {
+				vmImageObjects = append(vmImageObjects, img)
+			}
+
+			harvesterClientSet := harvesterFake.NewSimpleClientset(vmImageObjects...)
+			fakeVMImageCache := fakeclients.VirtualMachineImageCache(harvesterClientSet.HarvesterhciV1beta1().VirtualMachineImages)
+			fakeSecretCache := fakeclients.SecretCache(corefake.NewSimpleClientset().CoreV1().Secrets)
+			fakeStorageClassCache := fakeclients.StorageClassCache(corefake.NewSimpleClientset().StorageV1().StorageClasses)
+			fakeVolumeSnapshotClassCache := fakeclients.VolumeSnapshotClassCache(harvesterFake.NewSimpleClientset().SnapshotV1().VolumeSnapshotClasses)
+			validator := NewValidator(fakeStorageClassCache, fakeSecretCache, fakeVMImageCache, fakeVolumeSnapshotClassCache, newFakeClient()).(*storageClassValidator)
+
+			err := validator.validateVMImageUsage(tc.storageClass)
+			if tc.expectError {
+				assert.NotNil(t, err, tc.name)
+			} else {
+				assert.Nil(t, err, tc.name)
+			}
+		})
+	}
+}
+
 func Test_storageClassValidator_Delete(t *testing.T) {
 	tests := []struct {
 		name         string


Yu-Jack (approved):
Thanks!

WebberHuang1118:
Hi @brandboat,

In my opinion, we should not change the fundamental behavior of the StorageClass validator. Since there is still a VMI image referencing the target StorageClass, blocking the deletion is reasonable.

If I understand correctly, in this case the StorageClass deletion is triggered by the VMI controller. We could handle this webhook error as a special case and add a retry mechanism.

Thanks.

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -322,7 +322,9 @@ func (v *storageClassValidator) validateVMImageUsage(sc *storagev1.StorageClass)
 
 	usedVMImages := make([]string, 0, len(vmimages))
 	for _, vmimage := range vmimages {
-		usedVMImages = append(usedVMImages, vmimage.Name)
+		if vmimage.DeletionTimestamp == nil {
+			usedVMImages = append(usedVMImages, vmimage.Name)
+		}
```
WebberHuang1118, line 327: In my opinion, we should not change the fundamental behavior of the StorageClass validator. Since there is still a VMI image referencing the target StorageClass, blocking the deletion is reasonable.

If I understand correctly, in this case the StorageClass deletion is triggered by the VMI controller. We could handle this webhook error as a special case and add a retry mechanism.

WebberHuang1118 (commented):
Hi @brandboat,

Please take a look at the comments. 

I’m wondering whether [this observation](https://github.com/harvester/harvester/issues/9720#issuecomment-3641009663) is actually the root cause. If it were, the StorageClass related to the VMI should never be deleted. However, based on my testing in my environment, that doesn’t seem to be the case.

I also noticed that the VMI-related StorageClass shows up in the UI, as seen [here](https://github.com/harvester/harvester/issues/9720#issue-3718405227), which usually doesn’t happen for this type of StorageClass.

WebberHuang1118:
Hi @brandboat,

This is a bit confusing, so I’d like to confirm my understanding first.

Initially, it was mentioned that this was caused by the VMI controller’s handling of backing images, as described [here](https://github.com/harvester/harvester/issues/9720#issuecomment-3641009663). 

However, it turns out that the upgrade VMI actually uses CDI, which suggests the issue is related to the frontend VMI and StorageClass removal order instead, as discussed [here](https://github.com/harvester/harvester/issues/9720#issuecomment-3641140951).

Do I understand this correctly?

If so, does the CDI image really need a dedicated StorageClass? And even if such a StorageClass exists, it doesn’t necessarily need to be deleted along with the VMI deletion, right? What do you think?

pkg/webhook/resources/storageclass/validator.go
brandboat: Actually the storage class deletion is triggered by UI, see https://github.com/harvester/harvester-ui-extension/blob/main/pkg/harvester/pages/c/_cluster/airgapupgrade/index.vue#L364-L380

brandboat:
> Do I understand this correctly?

@WebberHuang1118, yes your understanding is right, the root cause is mentioned in the pr desc, the issue is triggered when we first delete vm image then delete its storage class, however the storage class validator don't allow us to do so because storage class still have vmimage in-use though vmimage is in deleting state.

> If so, does the CDI image really need a dedicated StorageClass? And even if such a StorageClass exists, it doesn’t necessarily need to be deleted along with the VMI deletion, right? What do you think?

For the upgrade vm image, we create a dedicated storage class named after the vm image name. The storage class has the same config as `longhorn-static`. But we already block user to use this storage class, we still need to create a new one for the upgrade image, and delete it once its no longer in-use - either after upgrade completes or if the image is manually deleted from the UI.
To me, it feels a bit odd to leave behind a storage class when it’s no longer in use. It becomes a leftover, and over time these unused storage classes could accumulate without providing any real value. I think we should clean them up instead of keeping them around.

If this PR isn’t acceptable, what about this: 
We can relax the restriction on using the longhorn-static storage class - allow upgrade images to use `longhorn-static` storage class directly. This way, we wouldn’t need to create a dedicated storage class for each upgrade image.
https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/persistentvolumeclaim/validator.go#L163-L166

WebberHuang1118:
> the root cause is mentioned in the pr desc, the issue is triggered when we first delete vm image then delete its storage class, however the storage class validator don't allow us to do so because storage class still have vmimage in-use though vmimage is in deleting state.

One thing is not clear to me is such behavior applies to all vm images, the vmi controller deletes the sc and bi in order if the vmi is backed by LH, if the backend is cdi, there is no operations related to sc,and these behavior appear fine. 

Why only the upgrade image has such situation? The sc is deleted by front-end for the upgrade case?

brandboat:
> Why only the upgrade image has such situation?

This is closely related to deprecating the upgrade-repo VM (see https://github.com/harvester/harvester/issues/7101
). To make the upgrade repo more robust, we switched from a VM-based approach to a deployment-based one. As part of this change, the underlying PVC of the VM image must use filesystem mode with RWX access, that's why we have this special handling for upgrade images.

> The sc is deleted by front-end for the upgrade case?

There are two different upgrade flows in Harvester:

1. Standard upgrade flow (via the Version CR and Upgrade CR):
In this case, the temporary storage class is created by the backend and is eventually cleaned up by the backend as well.
2. Air-gap upgrade flow (the scenario this PR aims to fix):
This flow is more complex. The VM image is uploaded directly by the user rather than being created by the backend. As a result, the temporary storage class is created by the UI and is also deleted by the UI. This is because the VMImage’s `.spec.targetStorageClassName` must be populated upfront, which means the storage class has to be created before the VM image is created.

WebberHuang1118:
> There are two different upgrade flows in Harvester:

> Standard upgrade flow (via the Version CR and Upgrade CR):
In this case, the temporary storage class is created by the backend and is eventually cleaned up by the backend as well.
Air-gap upgrade flow (the scenario this PR aims to fix):
This flow is more complex. The VM image is uploaded directly by the user rather than being created by the backend. As a result, the temporary storage class is created by the UI and is also deleted by the UI. This is because the VMImage’s .spec.targetStorageClassName must be populated upfront, which means the storage class has to be created before the VM image is created.

If I understand correctly, both the standard upgrade flow and the air-gap upgrade delete the temporary StorageClass. The difference is that in the standard upgrade the backend is responsible for deleting the StorageClass, while in the air-gap upgrade the frontend handles it.

According to the use description, this issue only occurs in the air-gap case? 

Before the GUI deletes the StorageClass, does it confirm that the upgrade VMI has been fully cleaned up?

brandboat:
> If I understand correctly, both the standard upgrade flow and the air-gap upgrade delete the temporary StorageClass. The difference is that in the standard upgrade the backend is responsible for deleting the StorageClass, while in the air-gap upgrade the frontend handles it.
>
> According to the use description, this issue only occurs in the air-gap case?

Correct

> Before the GUI deletes the StorageClass, does it confirm that the upgrade VMI has been fully cleaned up?

Nope, the GUI trigger VMI deletion without checking it already cleaned up or not and then trigger SC deletion.
In the beginning I thought we have to wait until VMI deletion then do SC deletion in the UI side - https://github.com/harvester/harvester/issues/9720#issuecomment-3641016092, and later realized that this should be fixed via backend, if the VMImage is in deleting stage, the SC should be considered unused and can be cleaned up, that's what I tried to fix in this PR.

brandboat:
> In my opinion, we should not change the fundamental behavior of the StorageClass validator. Since there is still a VMI image referencing the target StorageClass, blocking the deletion is reasonable.

Regarding https://github.com/harvester/harvester/pull/9785#discussion_r2650145902- One potential risk of this PR is that if deleting VMI images actually not removed, e.g. UI trigger VMI deletion & SC deletion, while VMI deletion operation stuck, user accidentally remove deletion timestamp, then VMI won't be deleted eventually.
Though I think this rarely happen, but is something we need to be aware of.

Back to the alternative solution discussion, there are 2 alternative ways to do this

1. proposed in comment https://github.com/harvester/harvester/pull/9785#issuecomment-3698163637
> If this PR isn’t acceptable, what about this:
We can relax the restriction on using the longhorn-static storage class - allow upgrade images to use longhorn-static storage class directly. This way, we wouldn’t need to create a dedicated storage class for each upgrade image.
https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/persistentvolumeclaim/validator.go#L163-L166

2. wait until VMI deletion completes then do SC deletion in the UI side.

@WebberHuang1118, @w13915984028, thoughts?

WebberHuang1118:
Thanks for the explanation, @brandboat.

> Regarding https://github.com/harvester/harvester/pull/9785#discussion_r2650145902 One potential risk of this PR is that if deleting VMI images actually not removed, e.g. UI trigger VMI deletion & SC deletion, while VMI deletion operation stuck, user accidentally remove deletion timestamp, then VMI won't be deleted eventually.

Based on the discussions so far, I think the current StorageClass validator design is reasonable. The StorageClass should not be deleted until the related VM is fully removed, so we may not need to change this behavior in this PR.

> Back to the alternative solution discussion, there are 2 alternative ways to do this
> 1. proposed in comment https://github.com/harvester/harvester/pull/9785#issuecomment-3698163637
> 2. wait until VMI deletion completes then do SC deletion in the UI side.
> 

I prefer option 2, as it’s more rational from an overall design perspective. If this isn’t feasible on the front end, option 1 would also be acceptable to me.

w13915984028:
@brandboat 

There looks to be a few challenges here:

(1) If UI code has retry mechanism?  

The `async await` will rely on the running sequence that the webhook can have it's cached data synced with the vmimage is gone/is updated with the deletionTimeStamp.

If UI runs too fast, then the backend change (point 3) might not cover all cases.

```
UI
...
          extraActionAfterRemove: async() => {
            await this.deleteImageStorageClass(imageDisplayName);
          }
```

(2) If UI code panics/breaks (e.g. network issue) inbetween SC and vmimage deletion, as UI is not in controller fashion, the vmimage might still be leftover, but very rarely. Even (1) is fully fixed.

(3) The backend change, per your PR, are there any labels/annotations to identify this air-gapped upgrade related vmimage & SC ? if we can identify it, then relax the SC webhook check to ignore the vmimage with deletiontimestap might be a good balance.

brandboat:
> (3) The backend change, per your PR, are there any labels/annotations to identify this air-gapped upgrade related vmimage & SC ? if we can identify it, then relax the SC webhook check to ignore the vmimage with deletiontimestap might be a good balance.

@w13915984028, yes, the annotation is here:`harvesterhci.io/os-upgrade-image`, I'll go this way - relax the sc deletion  webhook

brandboat:
@w13915984028, for (1) & (2), indeed that is a problem, thanks for pointing that out.

> (3) The backend change, per your PR, are there any labels/annotations to identify this air-gapped upgrade related vmimage & SC ? if we can identify it, then relax the SC webhook check to ignore the vmimage with deletiontimestap might be a good balance.

We already have the annotation :`harvesterhci.io/os-upgrade-image`. But I feel it's a bit odd that the SC webhook's deletion check would ignore the deletionTimestamp when the annotation is present in the VM image. The special-case logic is hard to reason about, and even with a comment it's easy to miss why the behavior differs.

I'm considering an alternative approach: (as I proposed in https://github.com/harvester/harvester/pull/9785#issuecomment-3701569187, the option1)

Relax the restriction on using longhorn-static for upgrade images, i.e. allow upgrade image PVCs to use `longhorn-static` directly. This is the current restriction I’m referring to:
https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/persistentvolumeclaim/validator.go#L163-L166

Since the temporary SC we create is basically the same as `longhorn-static`, this avoids per-upgrade-image SC creation and we don't need to worry about the whole “temporary SC deletion” headache.

We originally block longhorn-static because users mistakenly use it for VMs and run into trouble —so it should remain “internal only.” see https://github.com/harvester/harvester/issues/8561. Upgrade images are an internal use case too, so allowing longhorn-static only when harvesterhci.io/os-upgrade-image is set seems reasonable.

WDYT?

WebberHuang1118:
> I'm considering an alternative approach: (as I proposed in https://github.com/harvester/harvester/pull/9785#issuecomment-3701569187, the option1)
> 
> Relax the restriction on using longhorn-static for upgrade images, i.e. allow upgrade image PVCs to use longhorn-static directly. This is the current restriction I’m referring to:
> https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/persistentvolumeclaim/validator.go#L163-L166
> 
> Since the temporary SC we create is basically the same as longhorn-static, this avoids per-upgrade-image SC creation and we don't need to worry about the whole “temporary SC deletion” headache.
> 
> We originally block longhorn-static because users mistakenly use it for VMs and run into trouble —so it should remain “internal only.” see https://github.com/harvester/harvester/issues/8561. Upgrade images are an internal use case too, so allowing longhorn-static only when harvesterhci.io/os-upgrade-image is set seems reasonable.

This proposal looks good to me. It’s rational and doesn’t conflict with the StorageClass webhook design. Since the StorageClass is tied to the VMI, in my opinion the existing protection in the StorageClass validator is necessary and should be preserved. Unless there’s a strong justification, we should avoid introducing special handling based on deletion timestamps or annotations at the webhook level. Thanks.

w13915984028:
@brandboat It looks the `longhorn-static` SC is a bit different with the default SC `harvester-longhorn`, both the labels/annotations and those normal parameters, are we sure it can be used for this case?

UI upgrade seems to have no option to set param like replica-count
image

but a normal image has such param:
image



chart doc also says:

>deploy/charts/harvester/charts/longhorn/questions.yaml:          Default name of Longhorn Static StorageClass. The "storageClassName" is assigned to PVs and PVCs that are created for an existing Longhorn volume. The "storageClassName" can also be used as a label, so it is possible to use a Longhorn StorageClass to bind a workload to an existing PV without creating a Kubernetes StorageClass object. The "storageClassName" needs to be an existing StorageClass. The default value is "longhorn-static".



```
$kubectl get storageclass longhorn-static -oyaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2025-11-12T09:20:29Z"
  name: longhorn-static
  resourceVersion: "5647"
  uid: caa3e394-1afc-47de-95aa-61341168eb81
parameters:
  staleReplicaTimeout: "30"
reclaimPolicy: Delete
volumeBindingMode: Immediate


$kubectl get storageclass harvester-longhorn -oyaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    harvesterhci.io/is-reserved-storageclass: "true"
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: "2025-11-12T09:19:54Z"
  labels:
  name: harvester-longhorn
  resourceVersion: "3743"
  uid: f92f9629-700e-4a36-88f4-bd3aadea833e
parameters:
  baseImage: ""
  fromBackup: ""
  migratable: "true"
  numberOfReplicas: "3"
  staleReplicaTimeout: "30"
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

BTW, I did not find a section from Harvester document about the feature https://github.com/harvester/harvester/issues/7690, maybe I missed some key words.

connect: https://github.com/harvester/harvester/issues/9860

brandboat:
> @brandboat It looks the longhorn-static SC is a bit different with the default SC harvester-longhorn, both the labels/annotations and those normal parameters, are we sure it can be used for this case?

Currently the content of temporary sc for VMImage usage is the same as longhorn-static, so it's ok to use longhorn-static SC.

> UI upgrade seems to have no option to set param like replica-count

Currently, we don’t expose this parameter on the VMImage creation page (either backing image or cdi). However, I agree with your point — if we plan to let users customize this field in the future, we would still need either a temporary storage class or a dedicated storage class for upgrade VMImages.

That said, since we don’t have this requirement today, and it’s more of a potential future need, we can address it when we actually introduce that customization.

For now, I personally lean toward reusing longhorn-static. c.c. @Vicente-Cheng, @WebberHuang1118

WebberHuang1118:
> > @brandboat It looks the longhorn-static SC is a bit different with the default SC harvester-longhorn, both the labels/annotations and those normal parameters, are we sure it can be used for this case?
> 
> Currently the content of temporary sc for VMImage usage is the same as longhorn-static, so it's ok to use longhorn-static SC.
> 
> > UI upgrade seems to have no option to set param like replica-count
> 
> Currently, we don’t expose this parameter on the VMImage creation page (either backing image or cdi). However, I agree with your point — if we plan to let users customize this field in the future, we would still need either a temporary storage class or a dedicated storage class for upgrade VMImages.
> 
> That said, since we don’t have this requirement today, and it’s more of a potential future need, we can address it when we actually introduce that customization.
> 
> For now, I personally lean toward reusing longhorn-static. c.c. @Vicente-Cheng, @WebberHuang1118

Agree with @brandboat’s idea. Instead of introducing a special rule to tweak the StorageClass webhook (as mentioned in [here](https://github.com/harvester/harvester/pull/9785#issuecomment-3708865315)), it may be better to keep the webhook behavior consistent.

We can continue using the existing `longhorn-static` StorageClass. Harvester doesn’t need to delete it, which also avoids the webhook rejection (while the VMImage behavior isn’t fully clear yet).

Meanwhile, if we truly need a custom StorageClass in the future, the current design should be flexible enough to introduce one without compatibility concerns.

WDYT? @Vicente-Cheng

w13915984028:
@brandboat Another case arise:

The `longhorn-static` SC is an OOB SC from the view of Harvester (not installed & managed via helm chart, no helm labels/annotations), on UI, it has a hint that it is internal and can't be deleted.

image


But, if I delete `longhorn-static` SC from `kubectl`, it just runs smoothly. Webhook does not deny it, as `longhorn-static` does not have the `harvesterhci.io/is-reserved-storageclass: "true"` annotation like `harvester-longhorn`.

https://github.com/harvester/harvester/blob/dedd291efd8c659a9f05e13b62b615477e708993/pkg/webhook/resources/storageclass/validator.go#L335-L357

```
harv41:/home/rancher # kubectl get storageclass -A
NAME                           PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
harvester-longhorn (default)   driver.longhorn.io   Delete          Immediate           true                   62d
longhorn                       driver.longhorn.io   Delete          Immediate           true                   62d
longhorn-image-666qx           driver.longhorn.io   Delete          Immediate           true                   57d
longhorn-image-f2vpr           driver.longhorn.io   Delete          Immediate           true                   56d
longhorn-image-fzk7j           driver.longhorn.io   Delete          Immediate           true                   61d
longhorn-image-p8vl4           driver.longhorn.io   Delete          Immediate           true                   56d
longhorn-static                driver.longhorn.io   Delete          Immediate           true                   62d
vmstate-persistence            driver.longhorn.io   Delete          Immediate           true                   62d
harv41:/home/rancher # kubectl delete storageclass longhorn-static
storageclass.storage.k8s.io "longhorn-static" deleted
harv41:/home/rancher # kubectl get storageclass -A
NAME                           PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
harvester-longhorn (default)   driver.longhorn.io   Delete          Immediate           true                   62d
longhorn                       driver.longhorn.io   Delete          Immediate           true                   62d
longhorn-image-666qx           driver.longhorn.io   Delete          Immediate           true                   57d
longhorn-image-f2vpr           driver.longhorn.io   Delete          Immediate           true                   56d
longhorn-image-fzk7j           driver.longhorn.io   Delete          Immediate           true                   61d
longhorn-image-p8vl4           driver.longhorn.io   Delete          Immediate           true                   56d
vmstate-persistence            driver.longhorn.io   Delete          Immediate           true                   62d
harv41:/home/rancher # 
```

By nature, we can assume that some customers remove it. If UI tries to refer it, another issue could happen: the `longhorn-static` does not exist at all. Note: Harvester does not manage `longhorn-static` directly to ensure it's lifecycle.

Another point, to decouple with `longhorn`, Harvester adds `harvester-longhorn` as the default SC from installation;  from source code, we do not allow to refer `longhorn-static` but has no protection upon it.  Even if we want to refer kind of pre-existing SC, the `harvester-longhorn` looks better than `longhorn-static`.

https://github.com/harvester/harvester/blob/dedd291efd8c659a9f05e13b62b615477e708993/pkg/webhook/resources/persistentvolumeclaim/validator.go#L156-L169


Again, what's the consideration for UI to implement a new OS-image rather than reuse the original image creation process? It seems to import some normal/corner error cases.

image

brandboat:
@w13915984028, thanks for the feedback

> But, if I delete longhorn-static SC from kubectl, it just runs smoothly. Webhook does not deny it, as longhorn-static does not have the harvesterhci.io/is-reserved-storageclass: "true" annotation like harvester-longhorn.

Good point. In that case, we should create a dedicated StorageClass specifically for upgrade VMImage usage.

This StorageClass should:
- be restricted to upgrade VMImages only, and
- include the protection annotation harvesterhci.io/is-reserved-storageclass: "true" to prevent accidental deletion.

c.c. @WebberHuang1118, @Vicente-Cheng let me know if you have any concern.

>  Even if we want to refer kind of pre-existing SC, the harvester-longhorn looks better than longhorn-static.

Sadly we cannot do that, `harvester-longhorn` have this param
```
migratable: "true"
```
which blocks us from creating RWX filesystem PVC.

> Again, what's the consideration for UI to implement a new OS-image rather than reuse the original image creation process? It seems to import some normal/corner error cases.

I’m not entirely sure about this. The air-gapped OS upgrade UI has been around for a while, but the VM Image page is conceptually meant for uploading/downloading images used by VMs. Now that we’re using the deployment-based upgrade repo instead of the VM-based one, reusing the VM Image page for this purpose feels a bit mismatched in terms of semantics. If we really want to go this direction, we would need to clearly differentiate between these two use cases, which adds extra complexity. I’m not convinced that’s a good idea.

w13915984028:
@brandboat 

The current vm image creation, has a cooperation between front-end and backend, the mode has been used and tested from long time. The os-upgrade seems not fully adopting the same pattern, instead, it creates & removes things from UI, and in turn, imports those new issues.

Could OS-upgrade adopt similar pattern? The customized SC parameters, especially the replica count, is important. in case like 2-mgmt + 1 witness, the default 3 replicas will cause LH to have tons of warnings. When multi versions are released, use different SC to carry differnt ISO is also very possible.

If we can't fix it in an easy way, I would prfer to mark this as a known issue and write the workaround on document atm. And then design an enhancement solution first. thanks.

brandboat:
@w13915984028, I just tried delete the `longhorn-static` and found out it is recreated after few seconds. This is done by 
https://github.com/longhorn/longhorn-manager/blob/v1.10.x/controller/setting_controller.go#L875-L890

So I think we can still keep using `longhorn-static` here, regarding the replica count, user can always change it by themselves.

w13915984028:
Per above code, I set this scenario:

1. create new SC longhorn-static1
2. edit longhorn setting default-longhorn-static-storage-class  to above sc
3. delete `longhorn-static`
4. it won't be recreated at least after 14 minutes

I guess the `migratable: "true"` can even be set to `longhorn-static1`

using `longhorn-static` from current implementation, have a few flaws

the uncontrolled SC, can be none-existing, or with unwanted params

```
cat > lh1.yaml << 'EOF'
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: longhorn-static1
parameters:
  staleReplicaTimeout: "30"
reclaimPolicy: Delete
volumeBindingMode: Immediate
EOF


kubectl get storageclass longhorn-static1  -oyaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2026-01-14T09:42:39Z"
  name: longhorn-static1
  resourceVersion: "4002766"
  uid: f8056255-a086-4e48-8127-3dfd63aebbe3
parameters:
  staleReplicaTimeout: "30"
reclaimPolicy: Delete
volumeBindingMode: Immediate

kubectl get settings.longhorn -n longhorn-system default-longhorn-static-storage-class -oyaml
apiVersion: longhorn.io/v1beta2
kind: Setting
metadata:
  annotations:
    longhorn.io/configmap-resource-version: "3740"
  creationTimestamp: "2025-11-12T09:20:27Z"
  generation: 2
  name: default-longhorn-static-storage-class
  namespace: longhorn-system
  resourceVersion: "4002924"
  uid: 9a1ebac7-e767-46eb-b7b6-5d980db64983
status:
  applied: true
value: longhorn-static1

kubectl get storageclass -A
NAME                           PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
harvester-longhorn (default)   driver.longhorn.io   Delete          Immediate           true                   63d
longhorn                       driver.longhorn.io   Delete          Immediate           true                   63d
longhorn-image-666qx           driver.longhorn.io   Delete          Immediate           true                   57d
longhorn-image-f2vpr           driver.longhorn.io   Delete          Immediate           true                   56d
longhorn-image-fzk7j           driver.longhorn.io   Delete          Immediate           true                   61d
longhorn-image-p8vl4           driver.longhorn.io   Delete          Immediate           true                   56d
longhorn-static1               driver.longhorn.io   Delete          Immediate           true                   14m
vmstate-persistence            driver.longhorn.io   Delete          Immediate           true                   63d


```

WebberHuang1118:
> Again, what's the consideration for UI to implement a new OS-image rather than reuse the original image creation process? It seems to import some normal/corner error cases.

If I understand correctly, @brandboat mentioned this in [this comment](https://github.com/harvester/harvester/pull/9785#issuecomment-3698453018): Harvester requires a StorageClass with **filesystem mode and RWX access** for upgrade deployment. This differs from OS images created by Harvester for VM creation, which use **block and migratable** mode.

In addition, the original design creates a dedicated StorageClass for upgrade deployment, and the GUI deletes the related VM image and StorageClass directly without checking the resource status. This led to the original situation described in [https://github.com/harvester/harvester/pull/9785#issuecomment-3698163637](https://github.com/harvester/harvester/pull/9785#issuecomment-3698163637).

> when we first delete vm image then delete its storage class, however the storage class validator don't allow us to do so because storage class still have vmimage in-use though vmimage is in deleting state.

We may need better protection for the `longhorn-static` StorageClass; however, IMO, this should be a separate issue. We can focus on the StorageClass used for upgrade deployment in this case.

w13915984028:
If we transfer to a dedicated SC for OS-upgrade, then the UI/controller should be responsible to create & check the related params, to ensure purpose usage.

The `longhorn-static` is a dummy SC, which can be totally missing or with other customized params on those running cluster, Harvester should avoid using it as more as possible.

Vicente-Cheng:
> If we transfer to a dedicated SC for OS-upgrade, then the UI/controller should be responsible to create & check the related params, to ensure purpose usage.
> 
> The `longhorn-static` is a dummy SC, which can be totally missing or with other customized params on those running cluster, Harvester should avoid using it as more as possible.

I discussed with @brandboat, looks like `longhorn-static` will persist with their controller (REF: https://github.com/longhorn/longhorn-manager/blob/v1.10.x/controller/setting_controller.go#L855-L893).

So, reusing this storageclass was workable. And we can avoid handling another SC ourselves.

Vicente-Cheng:
>1. create new SC longhorn-static1
>2. edit longhorn setting default-longhorn-static-storage-class  to above sc
>3. delete `longhorn-static`
>4. it won't be recreated at least after 14 minutes

I know that would be the issue, but I don't think every customer will use that way to "break" the system.
Because this StorageClass is for internal usage, as we announced, we can add a document to tell people not do anything with longhorn-static. 

For the extra validator, we can do that later.

brandboat:
superseded by https://github.com/harvester/harvester/pull/9895

Issue #9531, opened 2025-11-18
[ENHANCEMENT] Should deny disabling nvidia-driver-toolkit when there is Enabled vGPU Devices or MIG Configurations


albinsun:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

https://github.com/user-attachments/assets/f7e843cb-654f-4d08-8a43-c525207aa790

### Describe the solution you'd like

Deny disabling nvidia-driver-toolkit when there is Enabled vGPU Devices or MIG Configurations

innobead:
I believe this should be a backend effort. Removed `require/ui` but added `area/ui` for the relationship. 

cc @harvester/ui

albinsun:
Let's wait `rc1` for testing due to the [addon tagging lacking restriction](https://github.com/harvester/harvester/issues/8652#issuecomment-3354387555).

albinsun:
Test FAIL, can not enable a PCI GPU device.
Please help to check.

## Logs
* Support bundle: [supportbundle_994fcc60-7f41-4486-b7f2-9c877bef648f_2026-03-17T02-50-38Z.zip](https://github.com/user-attachments/files/26041552/supportbundle_994fcc60-7f41-4486-b7f2-9c877bef648f_2026-03-17T02-50-38Z.zip)



## Environment
* `harvester-v1.8.0-rc1`
  * Profile: Single nodes AMD64 Bare-metal HPE ProLiant DL360 
  * **ui-source**: `External` (https://releases.rancher.com/harvester-ui/dashboard/release-harvester-v1.8/index.html)


## Test Scenario
### :red_circle: pcidevices-controller
1. <details>
     <summary> ✔️ Enable <b>pcidevices-controller</b>
     </summary>
     
     <img width="1561" height="769" alt="Image" src="https://github.com/user-attachments/assets/53a6a8fa-ae65-4e90-9cfd-1c38a030b04b" />
   </details>

1. <details open>
     <summary> ❌ Enable a PCI GPU device
     </summary>
     
     * No Enable action on GPU device
       <img width="1561" height="388" alt="Image" src="https://github.com/user-attachments/assets/1c4a26c4-3b3d-4ec0-b194-97b0130a7147" />
       <img width="1561" height="388" alt="Image" src="https://github.com/user-attachments/assets/9fd66136-d863-4e7b-8f6e-64822f68353b" />
     * But do on other devices
[...]

Issue #9172, opened 2025-09-17
[BUG] Direct Creation of VirtualMachineInstance not possible in Harvester 1.6.0


SvenKirschbaum:
### Describe the Bug

Due to the changes done in https://github.com/harvester/harvester/pull/8339 first included with Harvester 1.6.0 it is no longer possible to directly create the Kubevirt v1.VirtualMachineInstance CRD without creating a v1.VirtualMachine. When trying to do so the operation fails with `admission webhook "mutator.harvesterhci.io" denied the request: virtualmachines.kubevirt.io "<vmi-name>" not found`.

Relevant Code Segment: https://github.com/harvester/harvester/blob/v1.6.0/pkg/webhook/resources/virtualmachineinstance/mutator.go#L48-L51

### To Reproduce

Try creating a minimal VMI without a "wrapping" VM, e.g. via kubectl.

### Expected Behavior

I would expect the webhook to allow direct VMI creation, as it was possible before Harvester 1.6.0.

### Support Bundle for Troubleshooting

Can be provided on request, but i don't think this is necessary.

### Environment

- Harvester version: 1.6.0
- Impacted VM: N/A
- Impacted volume (PV): N/A
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Baremetall Dell Servers
- Rancher version: N/A

### Additional context

Background: Our usecase is  deploying  ephemeral VMs for Gitlab Runner Jobs, which due to their nature do not benefit from wrapping them in a VM first.

### Workaround and Mitigation

No workaround known yet.

SvenKirschbaum:
Pinging @FrankYang0529 and @w13915984028 as Author/Reviewer of the change that caused this.

Vicente-Cheng:
Hi @SvenKirschbaum, 
Just curious, what scenario would we like to create VMI directly instead of a VM instance?

SvenKirschbaum:
> Hi [@SvenKirschbaum](https://github.com/SvenKirschbaum), Just curious, what scenario would we like to create VMI directly instead of a VM instance?

We don't have a hard requirement on directly creating VMIs. Our Usecase is provisioning Virtual Machines to execute Gitlab CI/CD Jobs, for which we implemented a [fleeting plugin](https://docs.gitlab.com/runner/fleet_scaling/fleeting/). The Virtual Machines have no persistent Storage (containerDisk as boot disk) and are ephemeral, with a typical lifetime of well under five minutes, after which they are deleted.

[...]

Issue #8943, opened 2025-08-21
[ENHANCEMENT]  UI double-confirm on `Abort Migration`


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

The `migration` can be triggered by 4 ways from Harvester v1.6.0, UI needs to add a double-confirm when user clicks `Abort Migration`.

For automatically triggered migration, user should not abort it unless the owner operation (upgrade, node-maintenance, hot-plug) was cancelled/stopped/stucking.

Backend will try to add annotation like `migration source` via https://github.com/harvester/harvester/issues/8924, UI fetchs it to show

Suppose when user clicks `Abort Migration` a pop-up window says
> `This migration is triggered by Harvester upgrade, are you sure to stop it?`

It at least gives user a very clean information about the action.

### Additional context

https://github.com/harvester/docs/pull/849#issuecomment-3209309907

Vicente-Cheng:
It should be implemented with #8924.

Vicente-Cheng:
Hi @w13915984028, 
Could you use one ticket for this thing?

It looks like we can use the #8924 for both things.
WDYT?

Vicente-Cheng:
minor UI enhancement, move to planning, and remove the assignee

Issue #8685, opened 2025-07-17
[ENHANCEMENT] Some further enhancements on ovn addon


w13915984028:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

With https://github.com/harvester/network-controller-harvester/pull/194 to work around issue https://github.com/harvester/harvester/issues/8677

Some further enhancments come:

### Describe the solution you'd like

1. Prevent the ovn from being disabled when vm/vmis are still referring related nads

  https://github.com/harvester/harvester/blob/f4d5848a84decd5a3321508b1169f853fcdbb594/pkg/webhook/resources/addon/validator.go#L74

e.g. When an VM refers to ovn related nad, and vm is running, then the ovn addon is disabled, it should be avoided.

2. Double confirm when `disable` addon from UI, one click is not enough. @a110605 

3. When a VM has dangling ovn nad, have better hint when `start` a VM

e.g. One VM uses an ovn nad, and vm is stopped, ovn is disabled,  after that, user click `start` on the VM

Harvester mutator adds network related selector automatically, but as ovn is only hosted by mgmt network, need to check if another kind of selector is also needed,

The mutator/validator is better to check and refuse such vm to be `start` with user-friendly message.

rrajendran17:
@w13915984028 Like to confirm that is this a common problem with all our existing add-ons ?

I remember I was seeing a case with managed-dhcp add-on and related  ippool and vmnetcfg resources existed even after deleting the add-on.

w13915984028:
@rrajendran17 No, it is not a general problem of existing add-ons. But some new extensions triggered by the new ovn addon.

There are two main considerations:

**1. How crd and deployment work**

  1.1 For instance, the `rancher-monitoring` and` rancher-logging` separate `crd` and `deployment`. crd is pre-installed, add-on enable/disable just install/remove deployments, all existing configurations like alertmanager rules and pvcs like prometheus tsdb data are kept.  vm-dhcp-controller is similar.

  1.2 Ideally, the `kube-ovn` can also adopt the solution, but due to the fact that it is mainly maintained by third-party, we need to follow the upstream.

**2. The layer/influence of add-on**

  2.1 `rancher-monitoring`, `rancher-logging` are some additional features to assit the cluster, disable them won't affect the general workload like vm, guest cluster.

[...]

PR #9801, opened 2025-12-31
feat: check whether vms use hostdevice or gpu before disbling pcidevices-controller and nvidia-driver-toolkit

Yu-Jack:
#### Problem:
When disabling pcidevices-controller and  nvidia-driver-toolkit, we don't check whether VMs use it or not. 

#### Solution:
We need to check whether VMs use it or not. For the Intuitive way, we can directly check `vm.spec`.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9531

#### case1 - pcidevices-controller

1. Enable pcidevices-controller
2. Start a VM with passthrough pcidevices / usbdevices
3. Disable pcidevices-controller

#### case2 - nvidia-driver-toolkit

1. Enable pcidevices-controller and nvidia-driver-toolkit
2. Enable SR-IOV GPU or MIG GPU
3. Start a VM with passthrough vGPU
4. Disable pcidevices-controller and nvidia-driver-toolkit

Both cases should pop up a rejected response like this:

image

### DIFF
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -3,6 +3,8 @@ package addon
 import (
 	"fmt"
 
+	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -10,8 +12,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	validationutil "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
-
-	"github.com/sirupsen/logrus"
 
 	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
@@ -21,7 +21,6 @@ import (
 	"github.com/harvester/harvester/pkg/util/logging"
 	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
-	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 )
 
 const (
@@ -195,11 +194,19 @@ func (v *addonValidator) validatePCIDevicesControllerAddonUpdate(newAddon *v1bet
 }
 
 func (v *addonValidator) validateNvidiaDriverToolkitAddonUpdate(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) error {
-	// check when nvidia-driver-toolkit addon is being disabled
-	if oldAddon.Spec.Enabled && !newAddon.Spec.Enabled {
-		return v.validateNvidiaDriverToolkitAddon()
+	// not being disabled, no validation needed
+	if !oldAddon.Spec.Enabled || newAddon.Spec.Enabled {
+		return nil
 	}
-	return nil
+
+	// check for skip annotation
+	if newAddon.Annotations != nil && newAddon.Annotations[util.AnnotationSkipNvidiaDriverToolkitAddonWebhookCheck] == "true" {
+		logrus.Warnf("%v addon is being disabled but webhook check is skipped", util.NvidiaDriverToolkitName)
+		return nil
+	}
+
+	// perform validation when nvidia-driver-toolkit addon is being disabled
+	return v.validateNvidiaDriverToolkitAddon()
 }
 
 func validateVClusterAddon(newAddon *v1beta1.Addon) error {
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -103,6 +103,7 @@ func (v *addonValidator) validateNewAddon(newAddon *v1beta1.Addon) error {
 }
 
 func (v *addonValidator) validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) error {
+	// Validate common fields
 	if newAddon.Spec.Chart != oldAddon.Spec.Chart {
 		return werror.NewBadRequest("chart field cannot be changed.")
 	}
@@ -111,57 +112,81 @@ func (v *addonValidator) validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon 
 		return werror.NewBadRequest(fmt.Sprintf("cannot perform operation, as an existing operation is in progress on addon %s", oldAddon.Name))
 	}
 
-	if newAddon.Name == vClusterAddonName && newAddon.Namespace == vClusterAddonNamespace && newAddon.Spec.Enabled {
-		return validateVClusterAddon(newAddon)
+	switch newAddon.Name {
+	case vClusterAddonName:
+		return v.validateVClusterAddonUpdate(newAddon, oldAddon)
+	case util.RancherLoggingName:
+		return v.validateRancherLoggingAddonUpdate(newAddon, oldAddon)
+	case util.DeschedulerName:
+		return v.validateDeschedulerAddonUpdate(newAddon, oldAddon)
+	case util.PCIDevicesControllerName:
+		return v.validatePCIDevicesControllerAddonUpdate(newAddon, oldAddon)
+	case util.NvidiaDriverToolkitName:
+		return v.validateNvidiaDriverToolkitAddonUpdate(newAddon, oldAddon)
 	}
 
-	if newAddon.Name == util.RancherLoggingName {
-		if oldAddon.Spec.Enabled == newAddon.Spec.Enabled {
-			// spec `enabled` is not changed
-			return nil
-		}
-		skip := newAddon.Annotations[util.AnnotationSkipRancherLoggingAddonWebhookCheck] == "true"
-		// check when addon is `enabled`
-		//   block if upgradeLog has deployed a managedchart as logging-operator
-		if newAddon.Spec.Enabled {
-			if skip {
-				logrus.Warnf("%v addon is enabled but webhook check is skipped", util.RancherLoggingName)
-				return nil
-			}
-			return v.validateEnableRancherLoggingAddon(newAddon)
-		}
+	return nil
+}
 
-		// check when addon is `disabled`
-		//   block if upgradeLog sits on top of rancher-logging addon
+func (v *addonValidator) validateVClusterAddonUpdate(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) error {
+	if newAddon.Namespace != vClusterAddonNamespace || !newAddon.Spec.Enabled {
+		return nil
+	}
+	return validateVClusterAddon(newAddon)
+}
+
+func (v *addonValidator) validateRancherLoggingAddonUpdate(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) error {
+	if oldAddon.Spec.Enabled == newAddon.Spec.Enabled {
+		// spec `enabled` is not changed
+		return nil
+	}
+
+	skip := newAddon.Annotations[util.AnnotationSkipRancherLoggingAddonWebhookCheck] == "true"
+
+	// check when addon is `enabled`
+	//   block if upgradeLog has deployed a managedchart as logging-operator
+	if newAddon.Spec.Enabled {
 		if skip {
-			logrus.Warnf("%v addon is disabled but webhook check is skipped", util.RancherLoggingName)
+			logrus.Warnf("%v addon is enabled but webhook check is skipped", util.RancherLoggingName)
 			return nil
 		}
-		return v.validateDisableRancherLoggingAddon(newAddon)
+		return v.validateEnableRancherLoggingAddon(newAddon)
 	}
 
-	if newAddon.Name == util.DeschedulerName && newAddon.Spec.Enabled {
-		if newAddon.Annotations != nil && newAddon.Annotations[util.AnnotationSkipDeschedulerAddonWebhookCheck] == "true" {
-			return nil
-		}
+	// check when addon is `disabled`
+	//   block if upgradeLog sits on top of rancher-logging addon
+	if skip {
+		logrus.Warnf("%v addon is disabled but webhook check is skipped", util.RancherLoggingName)
+		return nil
+	}
+	return v.validateDisableRancherLoggingAddon(newAddon)
+}
 
-		return v.validateDeschedulerAddon(newAddon)
+func (v *addonValidator) validateDeschedulerAddonUpdate(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) error {
+	if !newAddon.Spec.Enabled {
+		return nil
 	}
 
+	if newAddon.Annotations != nil && newAddon.Annotations[util.AnnotationSkipDeschedulerAddonWebhookCheck] == "true" {
+		return nil
+	}
+
+	return v.validateDeschedulerAddon(newAddon)
+}
+
+func (v *addonValidator) validatePCIDevicesControllerAddonUpdate(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) error {
 	// check when pcidevices-controller addon is being disabled
-	if newAddon.Name == util.PCIDevicesControllerName {
-		if oldAddon.Spec.Enabled && !newAddon.Spec.Enabled {
-			return v.validatePCIDevicesControllerAddon(newAddon)
-		}
+	if oldAddon.Spec.Enabled && !newAddon.Spec.Enabled {
+		return v.validatePCIDevicesControllerAddon(newAddon)
 	}
+	return nil
+}
 
+func (v *addonValidator) validateNvidiaDriverToolkitAddonUpdate(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) error {
 	// check when nvidia-driver-toolkit addon is being disabled
-	if newAddon.Name == util.NvidiaDriverToolkitName {
-		if oldAddon.Spec.Enabled && !newAddon.Spec.Enabled {
-			return v.validateNvidiaDriverToolkitAddon(newAddon)
-		}
+	if oldAddon.Spec.Enabled && !newAddon.Spec.Enabled {
+		return v.validateNvidiaDriverToolkitAddon(newAddon)
 	}
-
 	return nil
 }
 
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -175,11 +175,23 @@ func (v *addonValidator) validateDeschedulerAddonUpdate(newAddon *v1beta1.Addon,
 }
 
 func (v *addonValidator) validatePCIDevicesControllerAddonUpdate(newAddon *v1beta1.Addon, oldAddon *v1beta1.Addon) error {
-	// check when pcidevices-controller addon is being disabled
-	if oldAddon.Spec.Enabled && !newAddon.Spec.Enabled {
-		return v.validatePCIDevicesControllerAddon()
+	// not being disabled, no validation needed
+	if !oldAddon.Spec.Enabled || newAddon.Spec.Enabled {
+		return nil
 	}
-	return nil
+
+	// check for skip annotation
+	// 1. some users may accidentally disable the addon and start VMs without detaching devices
+	// 2. some users may want to forcibly disable the addon to initialize their devices state when vm is in stopped state
+	// in general, we hope users can detach devices from VMs before disabling the addon for safety and understanding what they are doing
+	// We check VM for case 1 and provide skip annotation for case 2
+	if newAddon.Annotations != nil && newAddon.Annotations[util.AnnotationSkipPCIDevicesControllerAddonWebhookCheck] == "true" {
+		logrus.Warnf("%v addon is being disabled but webhook check is skipped", util.PCIDevicesControllerName)
+		return nil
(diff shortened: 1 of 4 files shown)


pkg/webhook/resources/addon/validator.go
```diff
@@ -262,3 +279,56 @@ func (v *addonValidator) validateDeschedulerAddon(newAddon *v1beta1.Addon) error
...
+	}
+
+	if len(vmsWithDevices) > 0 {
+		return werror.NewBadRequest(fmt.Sprintf("pcidevices-controller addon cannot be disabled as the following VMs are using passthrough devices: %v", vmsWithDevices))
+	}
+
+	return nil
+}
+
+func (v *addonValidator) validateNvidiaDriverToolkitAddon(newAddon *v1beta1.Addon) error {
```
ibrokethecloud: would it make sense to just check if there any SRIOVGPUDevice enabled and ignore the VMs. Attempt to disable a SRIOVGPUDevice should trigger additional validation to inform users of VM's where the vGPU's may be in use?

Yu-Jack: > would it make sense to just check if there any SRIOVGPUDevice enabled and ignore the VMs.

I originally planed to check that, but SRIOVGPUDevice CR doesn't exist in Harvester side. So, I chose to validate the VM instead. If we want to check SRIOVGPUDevice, we need to generate the CR or use generic k8s API to fetch the resources, which I think both might not be a good practice for our case.

> Attempt to disable a SRIOVGPUDevice should trigger additional validation to inform users of VM's where the vGPU's may be in use?

It should trigger. But, disabling addon shouldn't trigger.

ibrokethecloud: there is the remote possibility a user may have directly passed through a GPU to be managed via kubevirt directly, in which case disabling the nvidia-driver-toolkit will be blocked even though there may be no vGPUs in use etc

https://kubevirt.io/user-guide/compute/host-devices/#preparation-of-mediated-devices-such-as-vgpu

we should probably just document the new change in behaviour in case someone hits the edge case

other option could be to use the dynamic client and fetch unstructured objects and parse them

Yu-Jack: I think if some users ignore our pcidevice-controller and nvidia-driver-toolkit addons, and pass through a GPU to the VM directly, they might not be our target in this case because they won't try to disable the addon, even enable the addon.

The reason which I wouldn't like to use dynamic client and fetch the objects is that it creates a dependency between harvester and pcidevice controller.

But, I just check other codes. It seems we allow using dynamic client to fetch unstructured objects. If we decided to use dynamic client, do you think we should also change the way of validating pcidevice-controller addon?

WDYT?

Yu-Jack: But, to be honest, I think it's more straightforward that we only check VM usage.

ibrokethecloud (dismissed):

pkg/webhook/resources/addon/validator.go
```diff
@@ -262,3 +304,56 @@ func (v *addonValidator) validateDeschedulerAddon(newAddon *v1beta1.Addon) error
 	}
 	return nil
 }
+
+func (v *addonValidator) validatePCIDevicesControllerAddon(newAddon *v1beta1.Addon) error {
```
WebberHuang1118: newAddon is not used

pkg/webhook/resources/addon/validator.go
```diff
@@ -262,3 +304,56 @@ func (v *addonValidator) validateDeschedulerAddon(newAddon *v1beta1.Addon) error
...
+		if vm.Spec.Template != nil {
+			// Check for HostDevices (PCI devices and USB devices)
+			if len(vm.Spec.Template.Spec.Domain.Devices.HostDevices) > 0 {
+				vmsWithDevices = append(vmsWithDevices, fmt.Sprintf("%s/%s", vm.Namespace, vm.Name))
+				continue
+			}
+			// Check for GPUs (vGPU devices)
+			if len(vm.Spec.Template.Spec.Domain.Devices.GPUs) > 0 {
+				vmsWithDevices = append(vmsWithDevices, fmt.Sprintf("%s/%s", vm.Namespace, vm.Name))
+			}
```
WebberHuang1118: Suggest using early return to reduce nesting

pkg/webhook/resources/addon/validator.go
```diff
@@ -262,3 +304,56 @@ func (v *addonValidator) validateDeschedulerAddon(newAddon *v1beta1.Addon) error
...
+	}
+
+	if len(vmsWithDevices) > 0 {
+		return werror.NewBadRequest(fmt.Sprintf("pcidevices-controller addon cannot be disabled as the following VMs are using passthrough devices: %v", vmsWithDevices))
+	}
+
+	return nil
+}
+
+func (v *addonValidator) validateNvidiaDriverToolkitAddon(newAddon *v1beta1.Addon) error {
```
WebberHuang1118: newAddon is not used

pkg/webhook/resources/addon/validator.go
```diff
@@ -262,3 +304,56 @@ func (v *addonValidator) validateDeschedulerAddon(newAddon *v1beta1.Addon) error
...
+	}
+
+	var vmsWithGPUs []string
+	for _, vm := range vms {
+		if vm.Spec.Template != nil {
+			// Check for GPUs (vGPU devices and SR-IOV GPU devices)
+			if len(vm.Spec.Template.Spec.Domain.Devices.GPUs) > 0 {
+				vmsWithGPUs = append(vmsWithGPUs, fmt.Sprintf("%s/%s", vm.Namespace, vm.Name))
+			}
+		}
```
WebberHuang1118: Suggest using early return to reduce nesting

WebberHuang1118 (commented):
Generally LGTM. Just a few suggestions. Thanks.

pkg/webhook/resources/addon/validator.go
Yu-Jack: it's for collecting the completed list.

pkg/webhook/resources/addon/validator.go
Yu-Jack: it's for collecting completed this for users.

pkg/webhook/resources/addon/validator.go
WebberHuang1118: I mean, this for example:
```
	var vmsWithDevices []string
	for _, vm := range vms {
		if vm.Spec.Template == nil {
			continue
		}

		// Check for HostDevices (PCI devices and USB devices)
		if len(vm.Spec.Template.Spec.Domain.Devices.HostDevices) > 0 {
			vmsWithDevices = append(vmsWithDevices, fmt.Sprintf("%s/%s", vm.Namespace, vm.Name))
			continue
		}
		// Check for GPUs (vGPU devices)
		if len(vm.Spec.Template.Spec.Domain.Devices.GPUs) > 0 {
			vmsWithDevices = append(vmsWithDevices, fmt.Sprintf("%s/%s", vm.Namespace, vm.Name))
		}
	}
```

pkg/webhook/resources/addon/validator.go
WebberHuang1118: I mean, this for example:
```
	var vmsWithGPUs []string
	for _, vm := range vms {
		if vm.Spec.Template == nil {
			continue
		}

		// Check for GPUs (vGPU devices and SR-IOV GPU devices)
		if len(vm.Spec.Template.Spec.Domain.Devices.GPUs) > 0 {
			vmsWithGPUs = append(vmsWithGPUs, fmt.Sprintf("%s/%s", vm.Namespace, vm.Name))
		}
	}
```

pkg/webhook/resources/addon/validator.go
Yu-Jack: okay

Yu-Jack: done

pkg/webhook/resources/addon/validator.go
Yu-Jack: okay, done

WebberHuang1118 (dismissed):

Yu-Jack:
@WebberHuang1118 i forgot to fix the test file, it should be fine right now

WebberHuang1118 (dismissed):

pkg/webhook/resources/addon/validator.go
```diff
@@ -262,3 +304,58 @@ func (v *addonValidator) validateDeschedulerAddon(newAddon *v1beta1.Addon) error
...
+	var vmsWithDevices []string
+	for _, vm := range vms {
+		if vm.Spec.Template == nil {
+			continue
+		}
+		// Check for HostDevices (PCI devices and USB devices)
+		if len(vm.Spec.Template.Spec.Domain.Devices.HostDevices) > 0 {
+			vmsWithDevices = append(vmsWithDevices, fmt.Sprintf("%s/%s", vm.Namespace, vm.Name))
+			continue
+		}
```
w13915984028, line 343: A question. thanks.

Did you consider to check vmi instead of vm? 

In general, if a VM is stopped, the pci/driver addon might be allowed to disable, sometimes user do this operation to fix or debug bugs, or simply force the device return to initial status.

If we check on VM, it means user needs to edit all those VMs to remove the referrence to the devices.

Suppose a VM is using such devices and addon is anyway disabled or failed in the upgrade path (it is possbile, addon runs into error status when it is upgraded, or redeployed due to param changes), then when click the `start` button of the VM, the VM will runs into error status, it is expected and can be relatively easier to toubleshooted.

The VM defintion is more like a place-holder, if VM can really run, is determined when VM is started.

Another example is the network related definition, currently we request the VM to be stopped to do network related operations, but do not request to change VM defintion.

https://github.com/harvester/network-controller-harvester/blob/67b26a944114641d2b43ef4c37524dc0a38f0b5e/pkg/webhook/vlanconfig/validator.go#L163-L166

Yu-Jack: > If we check on VM, it means user needs to edit all those VMs to remove the referrence to the devices.

From my understanding, if you need to detach the pci device, you need to stop the vm. Then, the VMI won't exist. So, checking the VMI might not help for this case.

> In general, if a VM is stopped, the pci/driver addon might be allowed to disable, sometimes user do this operation to fix or debug bugs, or simply force the device return to initial status.

That's novelty operation for me. If we allow this, then this PR shouldn't be merged. So, the question is "Should we allow this or should we provide other ways to let users force their device to return to initial status?"

> Suppose a VM is using such devices and addon is anyway disabled or failed in the upgrade path (it is possbile, addon runs into error status when it is upgraded, or redeployed due to param changes), then when click the start button of the VM, the VM will runs into error status, it is expected and can be relatively easier to toubleshooted.

So, you mean when upgrading the Harvester, addon will be disabled and enabled if some parameters in that addon are changed? It seems like a critical issue from day-1.

w13915984028: A VM definition, can have oversized cpu/memory/storage, non-existing network, non-existing selector target. They are not strictly denied when the vm object is created. Instead, the `start` action won't success.

Forcely request user to change VM's definition on a stopped VM,  should be avoided as more as possible. If we do, very likely, some existing pipelines/tools from the customer, will be broken.

also cpu manager:

https://github.com/harvester/harvester/blob/f07446138227e6c48ee3fcd4bac1b3f0e1c5cad6/pkg/webhook/resources/node/validator.go#L235C6-L235C25

Yu-Jack: Got it. It makes sense for me. Let me switch to check VMI.

w13915984028: Let's also add some notes to the document, reminder user to carefully do those operations.

Yu-Jack: (this is for https://github.com/harvester/harvester/pull/9801#discussion_r2715985916)
Because I just got confused myself, I thought I needed to check the stopped VM. But, our target is not that. So, it makes sense for me to check VMI.

Yu-Jack: > Let's also add some notes to the document, reminder user to carefully do those operations.

Okay, I'll draft first version and request you for a review.

pkg/webhook/resources/addon/validator.go
```diff
@@ -262,3 +304,52 @@ func (v *addonValidator) validateDeschedulerAddon(newAddon *v1beta1.Addon) error
...
+		// Check for GPUs (vGPU devices)
+		if len(vmi.Spec.Domain.Devices.GPUs) > 0 {
+			vmisWithDevices = append(vmisWithDevices, fmt.Sprintf("%s/%s", vmi.Namespace, vmi.Name))
+		}
+	}
+
+	if len(vmisWithDevices) > 0 {
+		return werror.NewBadRequest(fmt.Sprintf("pcidevices-controller addon cannot be disabled as the following VMIs are using passthrough devices: %v", vmisWithDevices))
```
w13915984028: refer the const PCIDevicesControllerName

pkg/webhook/resources/addon/validator.go
```diff
@@ -262,3 +304,52 @@ func (v *addonValidator) validateDeschedulerAddon(newAddon *v1beta1.Addon) error
...
+	var vmisWithGPUs []string
+	for _, vmi := range vmis {
+		// Check for GPUs (vGPU devices and SR-IOV GPU devices)
+		if len(vmi.Spec.Domain.Devices.GPUs) > 0 {
+			vmisWithGPUs = append(vmisWithGPUs, fmt.Sprintf("%s/%s", vmi.Namespace, vmi.Name))
+		}
+	}
+
+	if len(vmisWithGPUs) > 0 {
+		return werror.NewBadRequest(fmt.Sprintf("nvidia-driver-toolkit addon cannot be disabled as the following VMIs are using GPU devices: %v", vmisWithGPUs))
```
w13915984028: const as well

w13915984028 (dismissed):
LGTM, thanks.

pkg/webhook/resources/addon/validator.go
WebberHuang1118: Hi @Yu-Jack,

I’m a bit confused — why did we switch to checking the VMI? If the VM isn’t running, wouldn’t we miss checking that VM when disabling the addon?

WebberHuang1118:
Hi @Yu-Jack,

In my opinion, if we check the VMI for the use case discussed [here](https://github.com/harvester/harvester/pull/9801#discussion_r2719326896), it feels somewhat non-intuitive. So I think checking VM manifest should be safer as [mentioned](https://github.com/harvester/harvester/pull/9801#discussion_r2719327161)

I understand @w13915984028’s concern raised [here](https://github.com/harvester/harvester/pull/9801#discussion_r2715772267); 

however, whether we choose to check the VM or the VMI manifest will directly affect the user experience, so this trade-off needs to be considered.

cc @ibrokethecloud @w13915984028

pkg/webhook/resources/addon/validator.go
Yu-Jack: > I’m a bit confused — why did we switch to checking the VMI? If the VM isn’t running, wouldn’t we miss checking that VM when disabling the addon?

You can check the video from the original issue. When we have some running VM with host devices, we shouldn't allow users to disable the addon. 

I think checking VMI is correct for sure. But, the question is that "do we allow users to disable addon when the VM still have some addon setting in a stopped status"?

Yu-Jack: cc @w13915984028 @ibrokethecloud

WebberHuang1118: Yes, I've watched the video. However, does checking the VM manifest miss running VMs? If I understand correctly, checking the VM manifest is safer, right?

WebberHuang1118: So if we checked the VMI instead of VM, assume a VM with VGPU is stopped, Harvester will still alllow disable the addon, and as the user wants to start the VM, the user will find the VM can's boot up, even the addon is disabled, however, the VM still stick with the device, do I understand correctly?

Yu-Jack: Yeah, for me, I think checking VM is more safer, but @w13915984028 mentioned some cases https://github.com/harvester/harvester/pull/9801#discussion_r2715904421, which I think we should consider that.

Yu-Jack: It seems like there are two options

- Checking VMI
- Checking VM

Checking VM is more safer, but it requires users to detach the host device, then they can disable the addon.
Checking VMI is for running VM. Users can stop the VMs and restart the addon to reset the host device status.

Another question is that,  I think when we start a VM from the `virtualmachine` manifest, sometimes it fails to start, which it very common use case. So checking VMI is more like a loose way, checking VM is more like a strict way.

I'm not sure which one we want, so I'll revert my latest change first to block the approval.

ibrokethecloud: VM is a better object to verify as our entire UI flow / webhooks are based around VM objects.

For example a user may be able to disable the addons if VMI's are not running, but VM's will not schedule until the addon is enabled and all involved devices are configured again.

This may result in a bad UX where users may need to figure out why their previously running VM's are not running.

With current approach, users will be forced to remove the devices from the VM before addon can be disabled.

w13915984028:
It is a good chance to discuss a bit more:

We have two main vm-infra related addons: gpu, ovn;  for now, if vm is running and then disable them, webhook has no check. mentioned on ticket https://github.com/harvester/harvester/issues/8685

From users' perspective, there are two kinds of disabling:
(1) Longterm/forever, no longer to use this feature
(2) Temparary, disable and enable in a row,
(2.1) system/node maintainence, hardware replacement
(2.2) simply reset the driver/redeploy everything to initial state to test/recover bugs
(2.3) fail to upgrade/change the addon, cause the addon in abnormal status, then disable&enable to recover

The scenairo (2) is reasonable and common, and it is allowed on existing clusters. With this PR to check VM, the (2) is more complex as you need to remove usage from VM first. For some customers, this would be a breaking change.

We could assume that:
(1)When there is running VMIs, the `disable` needs an error, should be blocked
(2)When there is stopped VM which will be affeced, the `disable` needs a warning (note: stopped vm can have stale/false information)

If we could add an UI double confirm pop up window when user clicks `disable` addon, which warns like `those vms will not be able to start if addon is disabled`, but user can choose to continue;  and the webhook denis the running vmi case. 
If user disables addon from API or kubectl, then only webhook applies.

(I have a ticket for the double-confirma enhancement on `abort migration` https://github.com/harvester/harvester/issues/8943, it is also fit for addon)

This gives user more flexibility.

BTW, 
(1) there is scenairo that only VMI is created, no VM
https://github.com/harvester/harvester/issues/9172

(2) the current Harvester webhook, seems not check if the VM's gpu/pci device claim is really existing, do we have other webhook which checks? (sure, kubevirt/k8s won't allow vm to boot, as no device could be found)  if no, the vm with dangling resources is already possible

and this line https://github.com/harvester/harvester/pull/9801/changes#diff-04d8741be1b0b02cbddd196c96717c3da5ad2d0c8b9ee26047393d142482ff12R321 will be a bit tricky, a zombie VM with gpu clam might block addon disabling

https://github.com/harvester/harvester/blob/f990e7a9dda5b3a9db3812a4417e8ef4d3faa776/pkg/webhook/resources/virtualmachine/validator.go#L223-L245

Yu-Jack:
> The scenairo (2) is reasonable and common, and it is allowed on existing clusters. With this PR to check VM, the (2) is more complex as you need to remove usage from VM first. For some customers, this would be a breaking change.

To be honest, from the user experience side, I'll go with validing VMI instead VM. I don't think it's a proper decision to change the user experience. 

> (2) the current Harvester webhook, seems not check if the VM's gpu/pci device claim is really existing, do we have other webhook which checks? (sure, kubevirt/k8s won't allow vm to boot, as no device could be found) if no, the vm with dangling resources is already possible

We have a webhook to validate that if users'd like to disable the PCI/USB/vGPUdevice, we will check whether the device is assigned to a VM or not (https://github.com/harvester/pcidevices/blob/4d454a6747c2b72dbb4b81bb0b09f81bc2a1702a/pkg/webhook/pcideviceclaim.go#L90-L116).

However, I don't think validating whether the claim exists or not will be a good option here. Like you mentioned, some users might stop the VM and restart the addon to initialize the device status. In that case, if we check whether the claim exists, users won't be able to restart the addon because webhook will reject it due to a existed claim. Then, they will still need to disable the device firstly and restart the addon.

> (1) there is scenairo that only VMI is created, no VM https://github.com/harvester/harvester/issues/9172

Besides, it's a bit weird for me in that case. Form the Harvester product perspective, it seems they don't follow what we want users to do, which I can understand, but just feel a bit weird for that use case. The only VMI workload seems like a use case for pure KubeVirt environment. However, I do agree with removing the VMI webhook in that case.

It seems like validating VMI will be only option. cc @WebberHuang1118  @ibrokethecloud  WDYT?

WebberHuang1118 (dismissed):
LGTM, thanks.

Yu-Jack:
Gentle ping @w13915984028

w13915984028 (commented):
It looks the `running vmi check` is removed, how do we handle such cases:

(1) The `running vmi` must be stopped before disabling gpu/pci when it uses the devices, this will cause workload panic;  this check can not be bypassed.  Also the vm-less vmi.

(2) The `vm` definition can be out-of-sync with `vmi` on the runtime, UI allows to edit `vm` config and select `not restart`. The current PR can still hit (1).  This vm check can be bypassed with annotation for the easy system maintenance.

In short, check vm only does not cover (1) fully.

Please let me know if I missed something, thanks.

w13915984028 (dismissed):
LGTM, thanks.

Per offline discussion, let's document more details and the concered cases:

The check focused on VM definition when addon is disabled; if user is maintaining the system and try to disable&enable in a row, the check could be bypassed via an annotation.

Meanwhile, if the vm, vmi is out-of-sync (e.g. remove devices from VM definition but not reboot the vm at once), the disabling check might skip those running instances.

Aslo better to add comments on vm/pci related chapters to remind user to reboot vm without a delay after editing devices.

w13915984028 (approved):
LGTM, thanks.

Issue #9257, opened 2025-10-03
[ENHANCEMENT] Add cluster name to supportbundle archive


ravarga:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

The supportbundle archive name currently doesn't contain the cluster name. When performing upgrades of multiple clusters and collecting issue evidence it is confusing and cumbersome to differentiate between clusters then.

### Describe the solution you'd like

It should be possible to add the cluster name to the supportconfig filename, eventually replacing the id descriptor.

### Additional context

#SURE-10689

Yu-Jack:
Example:

Original
- `supportbundle_c2eb6186-525f-411e-91ee-e88ed090d66d_2025-09-17T03-39-17Z`

Expected
- `supportbundle_{cluster_name}_c2eb6186-525f-411e-91ee-e88ed090d66d_2025-09-17T03-39-17Z`
- or `supportbundle_{cluster_name}_2025-09-17T03-39-17Z`

The goal is to make it clearer for identification.

Vicente-Cheng:
We still need to confirm whether we need that.

For me, giving the field to type the cluster name for the support bundle, most likely, we rename the support bundle after it is generated.

lanfon72:
Verified this enhancement has been applied.

Test Information
----
* Environment: **MegaRAC SP-X 3 nodes**
* Harvester Version: **v1.8.0-rc1**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Login to dashboard navigate to _Advanced/Settings_
3. The default `support-bundle-file-name` should be empty
4. Edit `support-bundle-file-name`, it should NOT allow upper case characters and only accept RFC1123 naming rule.
5. Configure `support-bundle-file-name` to `sb` then generate support bundle
6. The generated support bundle should fit the naming rule `supportbundle_{support-bundle-file-name}_{yyyy-mm-ddThh-mm-ssZ}`
7. Reset the `support-bundle-file-name` and generate support bundle again
8. The generated support bundle should fit the namine rule `supportbundle_{random-hex-with-dash}_{yyyy-mm-ddThh-mm-ssZ}`

Issue #4738, opened 2023-11-17
[ENHANCEMENT] Add a setting like `cluster-display-name` to store customized Harvester cluster name


w13915984028:
**Is your enhancement related to a problem? Please describe.**

There is no formal way to set/get the name of a Harvester Cluster.

**Describe the solution you'd like**

The Harvester cluster ID/name/display-name/owner/remark/..., are something useful, and it should not be depedent on the Rancher management.

Add a setting like `cluster-display-name` to set/get the Harvester Cluster Name.

**Describe alternatives you've considered**

Define a dedicated CRD to store.

Or added as annotations to certain k8s cluster level object.

**Additional context**

The `SNMP MIB-2 System` can be an reference for cluster level object `System`.

https://www.alvestrand.no/objectid/1.3.6.1.2.1.1.html

```
Subsidiary references (single level)
[1.3.6.1.2.1.1.1](https://www.alvestrand.no/objectid/1.3.6.1.2.1.1.1.html) - sysDescr
[1.3.6.1.2.1.1.2](https://www.alvestrand.no/objectid/1.3.6.1.2.1.1.2.html) - sysObjectID
[1.3.6.1.2.1.1.3](https://www.alvestrand.no/objectid/1.3.6.1.2.1.1.3.html) - sysUpTime
[1.3.6.1.2.1.1.4](https://www.alvestrand.no/objectid/1.3.6.1.2.1.1.4.html) - sysContact
[1.3.6.1.2.1.1.5](https://www.alvestrand.no/objectid/1.3.6.1.2.1.1.5.html) - sysName
[1.3.6.1.2.1.1.6](https://www.alvestrand.no/objectid/1.3.6.1.2.1.1.6.html) - sysLocation
[1.3.6.1.2.1.1.7](https://www.alvestrand.no/objectid/1.3.6.1.2.1.1.7.html) - sysServices
```

Yu-Jack:
@w13915984028 Do we still need this issue? 

When I was working on https://github.com/harvester/harvester/pull/9804, I found this issue. Although I think https://github.com/harvester/harvester/pull/9804 only needs a display name instead of cluster name, but I just want to ensure whether we need this issue.

PR #9804, opened 2026-01-02
hep: customize support bundle file name [skip ci]

Yu-Jack:
#### Problem:
In original issue, it mentions we can use cluster name to do that. But, we don't have it and actually we have an adding cluster name issue before (https://github.com/harvester/harvester/issues/4738).

For this case, I think we just need a display name instead of cluster name.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9257


Yu-Jack:
Actually, I don't think we need this HEP ... Just rename the tarball. Otherwise, we use cluster name to rename it.

Issue #9261, opened 2025-10-03
[FEATURE] Support CD-ROM hotpluggable inject/eject


wheatdog:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

KubeVirt supports CD-ROM inject/eject at version 1.6. We should bring this feature to Harvester.

### Describe the solution you'd like

According to the announcement [^1], a new feature gate "DeclarativeHotplugVolumes" is introduced at KubeVirt 1.6 and the existing "HotplugVolumes" is going to be deprecated.

With the new feature gate, these new features are supported:
- volume can be hotplugged by directing editing the VirtualMachine spec
- CD-ROM can be injected/ejected without reboot

We should migrate to the new feature gate "DeclarativeHotplugVolumes".

### Additional context

> IMPORTANT - Breaking Change:
> Declarative hotplug is incompatible with the existing "HotplugVolumes" feature gate. Key differences:
> 
> 1. Ephemeral volumes are not supported - The declarative API cannot handle volumes that are hotplugged to a VMI but not persisted in the VirtualMachine definition
> 2. Automatic cleanup - If you switch from "HotplugVolumes" to "DeclarativeHotplugVolumes", all ephemeral hotplug volumes will be unplugged the next time the VM reconciles
> 
> Migration Path:
> - If you only used 'virtctl addvolume' with the '--persist' flag, you should be unaffected
> - If you have ephemeral hotplug volumes, you'll need to either:
>   - Convert them to persistent volumes before switching, or
>   - Accept that they will be unplugged during the transition

According to the announcement [^1], it's a breaking change to switch feature gate from "HotplugVolumes" to "DeclarativeHotplugVolumes". We should plan the migration carefully.

reference: https://kubevirt.io/user-guide/storage/hotplug_volumes/

[^1]: https://groups.google.com/g/kubevirt-dev/c/-tRpvU6CvVY

w13915984028:
Original planned issue https://github.com/harvester/harvester/issues/3880.

@wheatdog Can 8283 cover the below scenario from  #3880? Thanks.

When a new VM (mainly Windows VM) is created, user adds a CD-ROM for booting, after the VM is up and Windows OS is installed, the CD-ROM can be removed without the restaring of VM (power off, power on) .

wheatdog:
[...]

PR #9840, opened 2026-01-07
docs(hep): CD-ROM Hotplug Volumes [skip ci]

wheatdog:
#### Problem:

Please refer to the HEP.

#### Solution:

Please refer to the HEP.

#### Related Issue(s):

#9261 

#### Test plan:

Please refer to the HEP.

For UI/UX, please refer to #9837

### DIFF
--- a/enhancements/20260106-cdrom-hotplug-volumes.md
+++ b/enhancements/20260106-cdrom-hotplug-volumes.md
@@ -1,0 +1,324 @@
+# CD-ROM Hotplug Volumes
+
+## Summary
+
+This enhancement proposal addresses the current limitation in Harvester where changing CD-ROM media for a Virtual Machine (VM) necessitates a VM restart. By leveraging KubeVirt's support for CD-ROM volume insertion and ejection (available since version 1.6), this HEP aims to introduce the capability to dynamically manage CD-ROM *volumes* on running VMs without interruption. This significantly improves user experience by allowing seamless swapping of ISO images or other CD-ROM content.
+
+It is important to distinguish this feature from attaching or detaching the CD-ROM *device* itself, which still typically requires a VM restart. This proposal focuses solely on the live management of the content (volume) within an already attached CD-ROM device.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/9261
+
+## Motivation
+
+### Goals
+
+- Support inserting a CD-ROM volume to a running VM with an empty CD-ROM device without restarting
+- Support ejecting a CD-ROM volume from a CD-ROM device of a running VM without restarting
+
+### Non-goals
+
+- Attach a CD-ROM device to a running VM without restarting
+- Detach a CD-ROM device from a running VM without restarting
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Insert an image into an empty CD-ROM device
+
+Users shall be able to insert an ISO image into an empty CD-ROM device of a running Virtual Machine. This enables dynamic loading of installation media or utilities without requiring a VM restart.
+
+#### Story 2: Eject an image from an occupied CD-ROM device
+
+Users shall be able to eject an ISO image from an occupied CD-ROM device of a running Virtual Machine. This allows for safe removal of media and prepares the drive for new content without interrupting the VM's operation.
+
+#### Story 3: Replace an image in an occupied CD-ROM device
+
+Users shall be able to replace an existing ISO image in an occupied CD-ROM device of a running Virtual Machine by first ejecting the current volume and then inserting a new one. This provides flexibility for multi-stage software installations or troubleshooting.
+
+#### Story 4: Create a Virtual Machine with an empty CD-ROM device
+
+Users shall be able to create a Virtual Machine with an empty CD-ROM device. This allows for insertion of additional media later.
+
+### User Experience In Detail
+
+#### Via UI
+
+The following steps outline the user workflow for managing CD-ROM volumes on a running VM:
+
+1. Navigate to the details page of a running VM.
+2. On the **Volumes** tab, the user can find the list of attached CD-ROM devices.
+3. Empty CD-ROM devices connected via the SATA bus will display an "Insert" action button. Clicking this button will open a dialog for the user to select an image to insert.
+4. Occupied CD-ROM devices connected via the SATA bus will display an "Eject" action button. Clicking this button will eject and delete the corresponding volume.
+
+#### Via Terraform
+
+In the nested block `disk` of `harvester_virtualmachine` resource, user can already specified whether the disk is hotpluggable by `hot_plug`. `hot_plug` would automatically set to `true` if unspecified when the `type` is `cd-rom` and `bus` is `sata`.
+
+to define a VM with empty CD-ROM devices, they can define a nested `disk` block without `image` field like
+```
+  disk {
+    name       = "cdrom-disk"
+    type       = "cd-rom"
+    bus        = "sata"
+    boot_order = 2
+  }
+```
+
+to insert the image, they can simply add the `image`, `size`, and `hot_plug` fields (optional: `auto_delete`)
+```diff
+  disk {
+    name       = "cdrom-disk"
+    type       = "cd-rom"
+    bus        = "sata"
+    boot_order = 2
+
++   size        = "10Gi"
++   hot_plug    = true
++   image       = harvester_image.k3os.id
++   auto_delete = true
+  }
+```
+
+to eject the image, they can remove the `image`, `size`, and `hot_plug` fields (optional: `auto_delete`)
+```diff
+  disk {
+    name       = "cdrom-disk"
+    type       = "cd-rom"
+    bus        = "sata"
+    boot_order = 2
+
+-   size        = "10Gi"
+-   hot_plug    = true
+-   image       = harvester_image.k3os.id
+-   auto_delete = true
+  }
+```
+
+to replace the image, they can change the `image` and `size` fields
+
+```diff
+  disk {
+    name       = "cdrom-disk"
+    type       = "cd-rom"
+    bus        = "sata"
+    boot_order = 2
+
+-   size       = "10Gi"
++   size       = "15Gi"
+    hot_plug   = true
+-   image       = harvester_image.k3os.id
++   image       = harvester_image.opensuse.id
+    auto_delete = true
+  }
+```
+
+### API changes
+
+Two new actions will be added to the `VirtualMachine` resource to handle CD-ROM volume operations.
+
+**Insert a CD-ROM volume**
+
+This action inserts an image into a specified empty CD-ROM device.
+
+```
+POST /v1/harvester/vm/<vm>?action=insertCdRomVolume
+Content-Type: application/json
+
+{
+  "deviceName": "cdrom-vol-abc",
+  "imageName": "default/image-pvvwf"
+}
+
+# Response
+- status code: 204
+```
+
+**Eject a CD-ROM volume**
+
+This action ejects the volume from a specified CD-ROM device.
+
+```
+POST /v1/harvester/vm/<vm>?action=ejectCdRomVolume
+Content-Type: application/json
+
+{
+  "deviceName": "cdrom-vol-abc"
+}
+
+# Response
+- status code: 204
+```
+
+**Deprecation of `ejectCdRom` Action**
+
+The existing `ejectCdRom` action ("Eject CD-ROM" on UI), which detaches the entire CD-ROM *device* and requires a VM restart, will be deprecated. The new `ejectCdRomVolume` action is preferred as it only ejects the volume without causing VM downtime. The old action will be removed.
+
+## Design
+
+### Implementation Overview
+
+#### KubeVirt Feature Gate Migration
+
+As noted in the KubeVirt documentation[^1], this feature relies on the `DeclarativeHotplugVolumes` feature gate. This will replace the older `HotplugVolumes` gate. According to the announcement[^2], this is a breaking change. Fortunately, ephemeral hotplug volumes isn't used in Harvester, making this a safe and direct replacement.
+
+#### CD-ROM Related VM Actions
+
+The Harvester API server will expose two new actions on `VirtualMachine` resources. These actions will modify the `VirtualMachine` resource by adding or removing the volume definition from `spec.template.spec.volumes` and updating the corresponding PVC information in the `harvesterhci.io/volumeClaimTemplates` annotation.
+
+- **`insertCdRomVolume`**: This action will patch the target `VirtualMachine` by adding a new `persistentVolumeClaim` to the `spec.template.spec.volumes` list and a corresponding entry to the `harvesterhci.io/volumeClaimTemplates` annotation. The PVC will be marked with `hotpluggable: true`.
+
+```diff
+ apiVersion: kubevirt.io/v1
+ kind: VirtualMachine
+ metadata:
+   annotations:
+     harvesterhci.io/volumeClaimTemplates: >-
+-      [{"metadata":{"name":"vm-ws-1-rootdisk-giua3"},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"32Gi"}},"volumeMode":"Block","storageClassName":"harvester-longhorn"}}]
++      [{"metadata":{"name":"vm-ws-1-cdrom-disk-9sfws","annotations":{"harvesterhci.io/imageId":"default/image-pvvwf"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"5Gi"}},"volumeMode":"Block","storageClassName":"longhorn-image-pvvwf"}},{"metadata":{"name":"vm-ws-1-rootdisk-giua3"},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"32Gi"}},"volumeMode":"Block","storageClassName":"harvester-longhorn"}}]
+ ...
+ spec:
+   template:
+     spec:
+       domain:
+         devices:
+           disks:
+             - bootOrder: 1
+               cdrom:
+                 bus: sata
+               name: cdrom-disk
+             - bootOrder: 2
+               disk:
+                 bus: virtio
+               name: rootdisk
+             - bootOrder: 3
+               cdrom:
+                 bus: sata
+               name: virtio-container-disk
+             - disk:
+                 bus: virtio
+               name: cloudinitdisk
+ ...
+       volumes:
++        - name: cdrom-disk
++          persistentVolumeClaim:
++            claimName: vm-ws-1-cdrom-disk-9sfws
++            hotpluggable: true
+         - name: rootdisk
+           persistentVolumeClaim:
+             claimName: vm-ws-1-rootdisk-giua3
+         - containerDisk:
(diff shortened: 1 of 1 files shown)


ibrokethecloud:
proposal lgtm.

with the new feature gate allowing hotplug of cdrom's should we by default mutate new vm's to attach an empty cdrom object?

wheatdog:
> with the new feature gate allowing hotplug of cdrom's should we by default mutate new vm's to attach an empty cdrom object?

@ibrokethecloud, thanks for the suggestion! I can kinda see the benefit of always having a spare cdrom device in a VM. But, on the other hand, I'm not sure whether it's a good idea having mutator *always* attach an empty cdrom device to the VM. I prefer having users explicitly demand it and the flexibility of creating VM without cdrom devices.

I have a related (but not exactly the same) discussion with @a110605 and @houhoucoop before. In an early version of this HEP, I proposed to enable users to add an empty cdrom device to VM from UI (first point in #9837). It's removed since we don't find the real usage of it and prefer having users always specify an image while attaching a new cdrom device.

enhancements/20260106-cdrom-hotplug-volumes.md
```diff
@@ -0,0 +1,329 @@
...
+
+https://github.com/harvester/harvester/issues/9779
+
+Currently, Harvester blocks the live migration of any VM with a CD-ROM attached, assuming the underlying volume access mode is `ReadWriteOnce` (RWO), which is not migratable. This check is overly restrictive, as CD-ROMs backed by `ReadWriteMany` (RWX) images are safe to migrate.
+
+This enhancement will relax the feasibility check. Instead of blocking migration based on the disk type (`cd-rom`), the validation logic will inspect the status of `LiveMigratable` condition from the VM. This will allow live migration for VMs with CD-ROMs.
+
+#### Blocking DataVolume Usage
+
+In Harvester, we have a different way to automate importing virtual machine disks onto PVCs instead of using [DataVolume](https://kubevirt.io/user-guide/storage/disks_and_volumes/#datavolume). To avoid uncontrolled scenarios, we'll block the usage of DataVolume under `spec.template.spec.volumes` for `VirtualMachine`.
```
wheatdog: It was discussed in the HEP review once and with @Vicente-Cheng offline. We would like to collect more inputs on whether we want to block the usage of DataVolume in the webhook.

cc @harvester/dev

enhancements/20260106-cdrom-hotplug-volumes.md
```diff
@@ -0,0 +1,335 @@
...
+### Non-goals
+
+- Attach a CD-ROM device to a running VM without restarting
+- Detach a CD-ROM device from a running VM without restarting
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Insert an image into an empty CD-ROM drive
```
w13915984028: A question, thanks:

should we add a new story about the potential change on the new VM creation?

currently, if a `cd-rom` is selected, the `image` field is mandatory;  after the new implementation, this field can be left blank, or totally gone?

Image

does it also means: we need to document or even add a cluster level setting, or have a `tick` box like `support cd-rom hot-plug`:

then Harvester appends a dummy `cd-rom` device to newly created VM automatically (besides the docker-machie-dirver case)

only when a VM has a blank `cd-rom` device, it can have the hot-plug ability ? thanks.

enhancements/20260106-cdrom-hotplug-volumes.md
starbops: I'm not fond of adding a validating webhook to prevent knowledgeable/advanced users who interact with the API directly. We can restrict usage by not providing options for resources other than VirtualMachineImage in the UI, but adding a webhook to allow only VirtualMachineImage seems a bit paranoid.
But I feel it's okay to add it at the beginning, as we may have loose ends, such as not displaying the DataVolume-type CD-ROM properly on the UI. We can always adjust based on user feedback. It's understandable for a feature only to support a subset of targets in an early stage.

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: > should we add a new story about the potential change on the new VM creation?
> 
> currently, if a cd-rom is selected, the image field is mandatory; after the new implementation, this field can be left blank, or totally gone?

In an early version of this HEP, I did propose to support VM creation with empty cdrom devices, namely, making the image field optional (first point in #9837). After discussing with @a110605 and @houhoucoop for UI/UX, we couldn't found valid use cases and prefer having users always specify an image while attaching a new cdrom device. Therefore, the image field would remain mandatory.

> does it also means: we need to document or even add a cluster level setting, or have a tick box like support cd-rom hot-plug:
> 
> then Harvester appends a dummy cd-rom device to newly created VM automatically (besides the docker-machie-dirver case)
> 
> only when a VM has a blank cd-rom device, it can have the hot-plug ability ? thanks.

I think we don't need a setting to toggle cdrom volume hotplug. This feature only works for VMs with cdrom devices, and those VMs would eventually gain the ability to hotplug cdrom volumes for insertion/ejection. Users can still create VMs w/o cdrom devices. We wouldn't append empty cdrom device for newly created VM automatically.

Here is my thought: after rolling out this feature,
- Existing Stopped VMs with occupied cdrom devices: cdrom volumes gain hot-plug capabilities when starting it
- Existing Running VMs with occupied cdrom devices: cdrom volumes gain hot-plug capabilities when restarting it
- Existing Stopped VMs w/o cdrom devices: No affected
- Existing Running VMs w/o cdrom devices: No affected
- Newly created VMs with occupied cdrom devices: cdrom volumes would automatically have hot-plug capabilities
- Newly created VMs w/o cdrom devices: No affected

As for attaching/detaching cdrom **devices** instead of insertion/ejection cdrom **volumes**, that would always necessitate a restart.

starbops: > I have a related (but not exactly the same) discussion with @a110605 and @houhoucoop before. In an early version of this HEP, I proposed to enable users to add an empty cdrom device to VM from UI (first point in https://github.com/harvester/harvester/issues/9837). It's removed since we don't find the real usage of it and prefer having users always specify an image while attaching a new cdrom device.

I suppose this is a required change. In the end, if the user cannot have an empty CD-ROM device VM created in the first place, how come they start using the CD-ROM hot-plug function? Unless I'm missing something here.

enhancements/20260106-cdrom-hotplug-volumes.md
```diff
@@ -0,0 +1,335 @@
...
+**Insert a CD-ROM volume**
+
+This action inserts an image into a specified empty CD-ROM device.
+
+```
+POST /v1/harvester/vm/<vm>?action=insertCdRomVolume
+Content-Type: application/json
+
+{
+  "volumeName": "cdrom-vol-abc",
```
starbops: Is there any reason that this would be called `volumeName`? I feel calling it `deviceName` is better if the API is meant to "insert an image to a CD-ROM device." The volume's name is more like intermediate information that only Harvester/KubeVirt cares about. And it becomes much clearer that the API allows the user to choose which CD-ROM device to insert the image from.

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: > In the end, if the user cannot have an empty CD-ROM device VM created in the first place, how come they start using the CD-ROM hot-plug function?

Our assumption for the journey is like, 
1) While creating VMs, users know which os they want to install. They could (and must) directly specify in the image field of the cdrom volume then proceed the os installation process.
2) After finish the os installation, they could just eject the media out of the cdrom device without restarting. That's when the hot-plug function come into play.
3) If users would like to acquire certain files (eg. drivers, etc) from the media, they can insert it to the cdrom device without restarting. This is another use case for the hot-plug function.

I think whether we would like to support attaching an empty cdrom device from UI (either during VM creation or editing) is discussible. We currently prefer not supporting it since we don't find a valid use case and would like to reduce the scope for UI changes.

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: Good suggestion. I agree that `deviceName` is better. Let me revise it. Thanks!

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: https://github.com/harvester/harvester/blob/fe439fa33bc4609402f4d882e7cd464e7d9e8e7a/enhancements/20260106-cdrom-hotplug-volumes.md?plain=1#L293-L295

Here is another related UI behavior change to review. Thanks!

starbops: Thanks for elaborating. I guess I want to know whether we eventually take off the restrictions, because it looks like a tight-schedule compromise to me. Maybe we can add a note to articulate the limitation/restriction in different phases, and keep the design as versatile as possible if we're going to support it in the future.

wheatdog: Understood. Let me add back the related changes to the HEP. We'll evaluate it while implementing UI part of this feature.

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: fixed by https://github.com/harvester/harvester/pull/9840/commits/aed95b9c61e7a40832b895b7c5449d108a23a3c5

enhancements/20260106-cdrom-hotplug-volumes.md
```diff
@@ -0,0 +1,335 @@
...
+- **`insertCdRomVolume`**: This action will patch the target `VirtualMachine` by adding a new `persistentVolumeClaim` to the `spec.template.spec.volumes` list and a corresponding entry to the `harvesterhci.io/volumeClaimTemplates` annotation. The PVC will be marked with `hotpluggable: true`.
+
+```diff
+ apiVersion: kubevirt.io/v1
+ kind: VirtualMachine
+ metadata:
+   annotations:
+     harvesterhci.io/volumeClaimTemplates: >-
+-      [{"metadata":{"name":"vm-ws-1-rootdisk-giua3"},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"32Gi"}},"volumeMode":"Block","storageClassName":"harvester-longhorn"}}]
++      [{"metadata":{"name":"vm-ws-1-cdrom-disk-9sfws","annotations":{"harvesterhci.io/imageId":"default/image-pvvwf"}},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"5Gi"}},"volumeMode":"Block","storageClassName":"longhorn-image-pvvwf"}},{"metadata":{"name":"vm-ws-1-rootdisk-giua3"},"spec":{"accessModes":["ReadWriteMany"],"resources":{"requests":{"storage":"32Gi"}},"volumeMode":"Block","storageClassName":"harvester-longhorn"}}]
```
ihcsim, line 180: wouldn't changing the annotation trigger a restart required condition?

update: AIUI, this needs to be updated as the content of the cd-rom changes, right?

enhancements/20260106-cdrom-hotplug-volumes.md
ihcsim: No, we should not block the usage of data volume. It's required for future features like virtual machine pools.

enhancements/20260106-cdrom-hotplug-volumes.md
```diff
@@ -0,0 +1,335 @@
...
+               name: vm-ws-1-rmeb7
+           name: cloudinitdisk
+ ...
+```
+
+- **Deprecate `ejectCdRom` VM action**: The old `ejectCdRom` action, which detaches the entire device, will be removed.
+
+#### Handling Existing VMs with CD-ROMs
+
+For existing VMs with attached CD-ROMs, the ability to eject the media without a restart will be enabled upon the next VM shutdown. When the VM's VMI is deleted (during a shutdown), the controller will patch the `spec.template.spec.volumes` of any `cd-rom` devices connected via SATA bus to include the `hotpluggable: true` flag.
```
ihcsim: > the controller will patch the `spec.template.spec.volumes`

as in the Harvester controller?

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: This wouldn't trigger a restart required condition, and yes, this is needed for the cd-rom content changes.

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: Yes, my original plan is like this: https://github.com/harvester/harvester/pull/9937/changes/f78a05f5dd4704c1b61500f7e4acba2e5de027d0 - set hotpluggable to true in the corresponding VM spec while deleting VMI.

This ensures the experience that all VMs would have CD-ROM hotplug volumes after at least one restart:
1. New VMs would automatically have CD-ROM hotplug volumes since the hotpluggable fields are set from FE
2. Existing VMs at first would not have hotpluggable fields but it would be set to true after a restart.

I feel like another way to implement this could be using vm mutator webhook: 
1. In `Create`, we always patch SATA CD-ROM volume to have `true` hotpluggable
2. In `Update`, to prevent accidentally trigger the restart required condition for existing running VM, we'll patch `true` hotpluggable only when `runStrategy` updated from `Halted` to `RerunOnFailure`.

WDYT?

enhancements/20260106-cdrom-hotplug-volumes.md
ihcsim: yeah, you are right - i thought while testing annotating a vm, i saw it triggered a restart required condition; must be my imagination.

enhancements/20260106-cdrom-hotplug-volumes.md
ihcsim: I don't have an opinion on whether to do it within the controller or the webhook. For `Create`, can the UI send us the correct YAML payload with cdrom hotplug enabled, instead of us patching it in the backend? 

And AIUI, we are proposing to enable cdrom hotplug by default. Will there be cases where user wants to disable it say due to kubevirt bugs, exploits etc.?

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: Good point, I think we should keep it as is for now. I'll remove this part later.

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: That's a good point. I think having backend patching it to enforce cdrom hotplug volume might not be a good design in terms of flexibility. I like your suggestion! Here is my revised design:
1. For `Create`, UI send the correct YAML payload with cdrom hotplug volume enabled.
2. For adding a new cdrom device to a VM, UI make sure that the payload for this new device should have cdrom hotplug volume enabled.
3. For existing VMs, cdrom hotplug volume would keep disabled. To enable it, they have two options
   1. delete and re-add the cd-rom volume in the Edit Page
   2. directly update the VM YAML to add `hotpluggable: true`

With this new design, we encourage cdrom hotplug volume to be enabled by default from UI but also have the flexibility to disable it via editing YAML directly.

WDYT?

ihcsim: Yeah, unless we have a strong requirement for it, I think it's better to leave existing VMs as-is. Like you said, we don't want to accidentally trigger the restart required condition for existing workloads.

wheatdog: Got it. I'll revise this part. Thanks for bringing this up!

ihcsim: Thanks - Back to my original question about disabling cdrom hotplug, the only way to do it is via editing the YAML, correct? Personally, I don't mind having a UI checkbox that is on by default, but I'm also ok with just using YAML.

wheatdog: Yes, I think we can leverage YAML editor to disable cdrom hotplug volume.

enhancements/20260106-cdrom-hotplug-volumes.md
```diff
@@ -0,0 +1,335 @@
...
+               name: virtio-container-disk
+             - disk:
+                 bus: virtio
+               name: cloudinitdisk
+ ...
+       volumes:
++        - name: cdrom-disk
++          persistentVolumeClaim:
++            claimName: vm-ws-1-cdrom-disk-9sfws
++            hotpluggable: true
```
WebberHuang1118, line 208: So for a KubeVirt VM, is hot-plugging a CD-ROM volume essentially the same as hot-plugging a regular volume we already introduced, or are there differences between the two resources?

enhancements/20260106-cdrom-hotplug-volumes.md
```diff
@@ -0,0 +1,335 @@
...
+
+https://github.com/harvester/harvester/issues/9779
+
+Currently, Harvester blocks the live migration of any VM with a CD-ROM attached, assuming the underlying volume access mode is `ReadWriteOnce` (RWO), which is not migratable. This check is overly restrictive, as CD-ROMs backed by `ReadWriteMany` (RWX) images are safe to migrate.
+
+This enhancement will relax the feasibility check. Instead of blocking migration based on the disk type (`cd-rom`), the validation logic will inspect the status of `LiveMigratable` condition from the VM. This will allow live migration for VMs with CD-ROMs.
+
+#### Blocking DataVolume Usage
+
+In Harvester, we have a different way to automate importing virtual machine disks onto PVCs instead of using [DataVolume](https://kubevirt.io/user-guide/storage/disks_and_volumes/#datavolume). To avoid uncontrolled scenarios, we'll block the usage of DataVolume under `spec.template.spec.volumes` for `VirtualMachine`.
```
WebberHuang1118: IIUC, `use DataVolume vs PVC` shouldn’t be specific to CD-ROM hot-plug. Could you explain more?

wheatdog: Sorry that I haven't revised the HEP yet.

This one was raised and discussed during HEP review meeting. The conclusion is to not block DataVolume. For details, we can refer to [this thread](https://github.com/harvester/harvester/pull/9840#discussion_r2693336078).

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: The difference is that a CD-ROM disk device needs to be present while creating VM such that the volume can be hotplugged to. Disk Hotplug doesn't have this prerequisite.

> https://kubevirt.io/user-guide/storage/hotplug_volumes/#inject-cd-rom
> To inject a CD-ROM into a running VirtualMachine, a cdrom type disk must be declared on the VM when it is started.

enhancements/20260106-cdrom-hotplug-volumes.md
wheatdog: Empty cdrom device made it into the implementation! I'll revise HEP accordingly.

wheatdog:
@ihcsim @starbops @WebberHuang1118 @w13915984028 I've revised the HEP and the corresponding implementation PRs are all merged. Could you help me review and approve this HEP PR again? Thanks!

w13915984028 (approved):
LGTM, thanks.

Issue #9607, opened 2025-11-25
[TASK] Use CODEOWNERS to efficient the review progress


Vicente-Cheng:
### Task request

GitHub provides CODEOWNERS to improve review progress.
REF: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection
e.g. we can define the specific component for the dedicated team/person.

We should use this approach to improve our code review process.

### Additional context

Corresponding Repos:
- [x] harvester/harvester, PR: https://github.com/harvester/harvester/pull/9853
- [ ] harvester/harvester-installer, PR: TBD
- [ ] harvester/harvester-ui-extension, PR: TBD
- [x] harvester/node-disk-manager, PR: https://github.com/harvester/node-disk-manager/pull/259
- [ ] harvester/node-manager, PR: TBD
- [ ] harvester/go-common, PR: TBD
- [ ] harvester/vm-import-controller, PR: TBD
- [ ] harvester/terraform-provider-harvester, PR: TBD
... TBD

_No response_

Vicente-Cheng:
moved to v1.9.0 first
I am still working on it

Vicente-Cheng:
All related PRs are merged, closed

PR #9853, opened 2026-01-11
common: add CODEOWNERS [skip ci]

Vicente-Cheng:
#### Problem:
Introduce CODEOWNERS

#### Solution:
Introduce CODEOWNERS

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9607

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,118 @@
+# To enhance the review process, we have designated specific teams as code owners for different parts of the repository.
+
+# We setup the codeowners with the following rules:
+# Common folders across repositories are assigned to the dev team.
+# Top 3 contributors of each directory are assigned as code owners, if don't have enough contributors, dev team will be assigned.
+# If the number of commits of a contributor is less than 2, that contributor will be ignored.
+
+# The corresponding git commit is used to determine the code owners.
+# git log --since="24 months ago" --pretty=format:'%an' -- <path> | sort | uniq -c | sort -nr | head -n 3
+
+** @harvester/maintainer
+
+/.github/ @harvester/dev
+/api/ @harvester/dev
+/cmd/ @harvester/dev
+/deploy/ @harvester/dev
+/docs/ @harvester/dev
+/enhancements/ @harvester/dev
+/hack/ @harvester/dev
+/misc/ @harvester/dev
+/package/ @harvester/dev
+/pkg/** @harvester/dev  # first level assignment, will be overridden by sub-folder assignments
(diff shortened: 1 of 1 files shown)


brandboat (dismissed):

starbops (dismissed):

votdev (dismissed):

CODEOWNERS
```diff
@@ -0,0 +1,121 @@
...
+
+# We setup the codeowners with the following rules:
+# Common folders across repositories are assigned to the dev team.
+# Top 3 contributors of each directory are assigned as code owners, if don't have enough contributors, dev team will be assigned.
+# If the number of commits of a contributor is less than 2, that contributor will be ignored.
+
+# The corresponding git commit is used to determine the code owners.
+# git log --since="24 months ago" --pretty=format:'%an' -- <path> | sort | uniq -c | sort -nr | head -n 3
+
+* @harvester/maintainer
```
innobead: What is the purpose of requesting maintainers to review the files in the top folder?

Or, do you want to ask maintainers to review uncovered files? Might it be `** @harvester/maintainer` instead?

CODEOWNERS
```diff
@@ -0,0 +1,121 @@
...
+/.github/ @harvester/dev
+/api/ @harvester/dev
+/cmd/ @harvester/dev
+/deploy/ @harvester/dev
+/docs/ @harvester/dev
+/enhancements/ @harvester/dev
+/hack/ @harvester/dev
+/misc/ @harvester/dev
+/package/ @harvester/dev
+/pkg/api/backuptarget @Yu-Jack @harvester/dev
```
innobead, line 23: Should we add `/pkg/**` for `@harvester/dev` by default? I assume new folders might be added in the future, so a default review for Go code is necessary.

innobead (changes requested):
This is a good start. A few comments to be resolved. 

This should be regularly reviewed over time (or eventually via an automatic update via workflow, as per the git command you provided). 

We can also adjust the applied ruleset to request reviewers based on recent code changes.

CODEOWNERS
Vicente-Cheng: `*` is equal to `**` in this description, I align with `**`

CODEOWNERS
Vicente-Cheng: nice suggestion, updated!

innobead (approved):

innobead:
Let's refine this gradually over time.

Issue #9383, opened 2025-10-29
[FEATURE] Expose Disk Cache Mode


m-ildefons:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

The disk cache mode is an important setting for fine-tuning the I/O performance and consistency.
The default cache behavior depends on the host filesystem and whether it supports direct I/O or not.

### Describe the solution you'd like

The disk cache mode should be exposed both in the Harvester UI and in the integrations (e.g. Terraform provider) in such a way that users can easily choose from the available behaviors.
The default setting should match the current behavior.

### Additional context

https://kubevirt.io/user-guide/storage/disks_and_volumes/#disk-device-cache

Vicente-Cheng:
Hi @m-ildefons, 
Since the above PR https://github.com/harvester/harvester/pull/9858 was merged, could we move this one to Ready for Testing?

m-ildefons:
This is not yet ready.
This issue needs two PRs:
1) a change to the `VMBuilder` interface (https://github.com/harvester/harvester/pull/9858)
2) a change in the Terraform provider (https://github.com/harvester/terraform-provider-harvester/pull/197)

The second of those changes can only happen once there has been a Harvester release, which contains the first change. As far as I'm aware, v1.8.0 will be the first Harvester release that contains https://github.com/harvester/harvester/pull/9858
So the second change can not yet happen.

innobead:
Moved to 1.9.0. There is no sprint specified as well. Let's make sure when working on issues, need to move them to the current sprint.

asc5543:
## Test result
Test PASS

## Environment
* Harvester ISO version: `v1.9.0-dev-20260607`
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): `2 node via ipxe-example`


## Test Scenario
### Preparation
1. <details> <summary> Install Harvester with v1.9.0-dev-20260607. </summary>

    <img width="1917" height="980" alt="Image" src="https://github.com/user-attachments/assets/e3a8e3df-3731-492e-af97-a063a568d178" />

   </details>

1. <details> <summary> Build terraform </summary>

    ```
    go mod vendor
    rm -rf ~/.terraform.d
    CGO_ENABLED=0 go build -mod=vendor -o bin/terraform-provider-harvester .
    mkdir -p ~/.terraform.d/plugins/registry.terraform.io/harvester/harvester/0.0.0-dev/linux_amd64
[...]

PR #9858, opened 2026-01-12
Builder: Add Disk Cache Mode, Errors and Tests

m-ildefons:
1) Add errors to the VMBuilder

Add error facility to the VMBuilder. This allows the builder to collect errors e.g. when volume information is inconsistent with disks or when parameters reference devices that don't exist in the domain spec.

2) Add disk cache mode

Add a method for setting the cache mode of a VM disk device.

3) Add tests for VMBuilder disk cache mode setting method

related-to: harvester/harvester#9383

#### Problem:

To facilitate setting the disk cache mode, e.g. for the Terraform provider, it must be exposed in the VM Builder. So far nothing there allows setting the cache mode for disks, meaning that the default values will always be used.

#### Solution:

Add a new method to the VMBuilder, which allows setting the caching mode of a disk device.

Implementing this function as a new method, rather than modifying the existing `Disk(...)` method, allows keeping backwards compatibility and leaves the interface simpler in case default settings suffice.

The downside is that there is the possibility that the caller can try to set the disk cache mode for a disk that doesn't (yet) exist in the domain spec. To handle this situation, the VMBuilder has been expanded to allow collecting errors during the definition of the VM.
These errors are then exposed when applying the build in the function call to `VM()` - much like other existing error cases during the builder application.

#### Related Issue(s):

- harvester/harvester#9383

#### Test plan:

Unit tests included.

### DIFF
--- a/pkg/builder/disk.go
+++ b/pkg/builder/disk.go
@@ -91,6 +91,23 @@ func (v *VMBuilder) Disk(diskName, diskBus string, isCDRom bool, bootOrder uint)
 	return v
 }
 
+func (v *VMBuilder) DiskCacheMode(name string, mode kubevirtv1.DriverCache) *VMBuilder {
+	diskFound := false
+
+	for i, disk := range v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Disks {
+		if disk.Name == name {
+			v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Disks[i].Cache = mode
+			diskFound = true
+		}
+	}
+
+	if !diskFound {
+		v.Error = fmt.Errorf("disk %s does not exist in VM %s", name, v.VirtualMachine.Name)
+	}
+
+	return v
+}
+
 func (v *VMBuilder) Volume(diskName string, volume kubevirtv1.Volume) *VMBuilder {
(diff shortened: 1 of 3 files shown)


pkg/builder/disk.go
```diff
@@ -91,6 +91,23 @@ func (v *VMBuilder) Disk(diskName, diskBus string, isCDRom bool, bootOrder uint)
 	return v
 }
 
+func (v *VMBuilder) DiskCacheMode(name string, mode kubevirtv1.DriverCache) *VMBuilder {
+	diskFound := false
+
+	for i, disk := range v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Disks {
+		if disk.Name == name {
+			v.VirtualMachine.Spec.Template.Spec.Domain.Devices.Disks[i].Cache = mode
+			diskFound = true
```
ihcsim: nit: can `break` or `return` early here.

ihcsim (dismissed):
LGTM.

pkg/builder/disk.go
m-ildefons: Thanks, fixed.

ihcsim (approved):

PR #9862, opened 2026-01-14
Add FOSSA scanning workflow

macedogm:
This PR adds FOSSA license scanning workflow.

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -1,0 +1,34 @@
+name: FOSSA Scanning
+
+on:
+  push:
+    branches: ["main", "master", "release/**"]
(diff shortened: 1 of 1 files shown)


.github/workflows/fossa.yml
```diff
@@ -0,0 +1,34 @@
+name: FOSSA Scanning
+
+on:
+  push:
+    branches: ["main", "master", "release/**"]
```
ihcsim: The release branches' naming convention in `h/h` is `release-**`, not `release/**`.

Also, we don't use `main` branch, but I think it's ok to keep it for in case we replace `master` with `main` eventually.

ihcsim (approved):

Issue #6903, opened 2024-10-28
[BUG] `context deadline exceeded` is observed when creating schedulevmbackup if the NFS server is down


w13915984028:
# case 2 is moved to https://github.com/harvester/harvester/issues/9865

**Describe the bug**

Two cases need to be solved:

case 1: webhook context deadline exceeded

 `context deadline exceeded` is observed when creating schedulevmbackup if the NFS server is down

>Internal error occurred: failed calling webhook "validator.harvesterhci.io": failed to call webhook: Post "https://harvester-webhook.harvester-system.svc:443/v1/webhook/validation?timeout=10s": context deadline exceeded

case 2: left `Error` backup can't be deleted

**To Reproduce**

---

case 1:

1. configure a valid NFS as backup target
2. stop the NFS server/NFS service
3. create a new `schedulevmbackup`, it reports below error

![image](https://github.com/user-attachments/assets/0c1c614b-7222-4988-b5fa-04e91077de0e)

Harvester-webhook does not log obvious information

---

case 2:

1. start the NFS server
2. create a new `schedulevmbackup`
![image](https://github.com/user-attachments/assets/2474a42f-9390-4b71-8a8f-cfa8a23efca1)
3. stop the NFS server/NFS service
4. the schedule is put to suspended after the timer arrives and the bakcup failed
![image](https://github.com/user-attachments/assets/00932056-e485-43c9-b60c-324716726866)

5. delete the source VM, it is ok
![image](https://github.com/user-attachments/assets/bd65571f-61dc-4f0d-8290-2516106b6c98)

6. delete the schedule, it is also ok
![image](https://github.com/user-attachments/assets/375f892e-0146-4d51-b7dd-e43936e8a833)

there is no schedule
![image](https://github.com/user-attachments/assets/ac7535e2-6498-4b33-8280-bcb23d9e1e68)

7. delete the `Error` backup, it does not hint any error
![image](https://github.com/user-attachments/assets/7090295f-f9ff-47c6-bac2-a561af52bdf5)

8. refresh, the backup is still
![image](https://github.com/user-attachments/assets/b00c479c-70f5-4003-9dac-179de1b90795)

UI action returns OK.
![image](https://github.com/user-attachments/assets/27134317-e173-4572-9266-4c7fd6bd6027)

the crd object has `deletionTimestamp: "2024-10-28T16:22:34Z"`, but looks to be blocked

<details>
<summary>related virtualmachinebackups object has deletion timestamp</summary>
<br>

[...]

PR #9863, opened 2026-01-14
feat: add timeout to connect backup target

Yu-Jack:
#### Problem:
The webhook timeout is 10 second, but longhorn timeout is 300 second. If endpoint is not reachable, we return the context canceled.

#### Solution:
We should have our own timeout when calling this function.

image

image

#### Related Issue(s):
https://github.com/harvester/harvester/issues/6903

### DIFF
--- a/pkg/util/backup/backup.go
+++ b/pkg/util/backup/backup.go
@@ -1,10 +1,12 @@
 package backup
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/longhorn/backupstore"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
@@ -43,9 +45,16 @@ func GetBackupStoreDriver(secretCache ctlcorev1.SecretCache, target *settings.Ba
 	}
 
 	endpoint := ConstructEndpoint(target)
-	bsDriver, err := backupstore.GetBackupStoreDriver(endpoint)
+
+	// Add 5 second timeout for backup store driver initialization.
+	// There might be a goroutine leak if the driver doesn't end properly,
+	// Although we can pass ctx, but the underlying driver implementation doesn't support it.
+	// So we should be careful when using GetBackupStoreDriver function.
+	bsDriver, err := util.RunWithTimeoutAndResult(5*time.Second, func(_ context.Context) (backupstore.BackupStoreDriver, error) {
+		return backupstore.GetBackupStoreDriver(endpoint)
+	})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to connect to backup target, reason: %w", err)
 	}
 	return bsDriver, nil
 }
--- a/pkg/util/backup/backup.go
+++ b/pkg/util/backup/backup.go
@@ -18,6 +18,8 @@ import (
 
 const (
 	VMImageMetadataFolderPath = "harvester/vmimages/"
+	// The webhook timeout is 10 seconds, so we can't set too long timeout here.
+	ConnectBackupStoreTimeout = 8 * time.Second
 )
 
 func ConstructEndpoint(target *settings.BackupTarget) string {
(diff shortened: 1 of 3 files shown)


pkg/util/backup/backup.go
```diff
@@ -43,9 +45,16 @@ func GetBackupStoreDriver(secretCache ctlcorev1.SecretCache, target *settings.Ba
 	}
 
 	endpoint := ConstructEndpoint(target)
-	bsDriver, err := backupstore.GetBackupStoreDriver(endpoint)
+
+	// Add 5 second timeout for backup store driver initialization.
+	// There might be a goroutine leak if the driver doesn't end properly,
+	// Although we can pass ctx, but the underlying driver implementation doesn't support it.
+	// So we should be careful when using GetBackupStoreDriver function.
+	bsDriver, err := util.RunWithTimeoutAndResult(5*time.Second, func(_ context.Context) (backupstore.BackupStoreDriver, error) {
```
Yu-Jack: I think 5 second is enough for this case. If not, we can have our own timeout setting. But, I don't think it's necessary at the moment. Just focus on context canceled issue. Otherwise, it might be out of the scope.

pkg/util/backup/backup.go
```diff
@@ -43,9 +45,16 @@ func GetBackupStoreDriver(secretCache ctlcorev1.SecretCache, target *settings.Ba
...
+		return backupstore.GetBackupStoreDriver(endpoint)
```
Yu-Jack, line 55: FYI, this is a potential goroutine leak. Usually every function should take ctx to support cancelation. Unluckily, `backupstore.GetBackupStoreDriver` doesn't have this. But, in this function, we should still pass ctx to the function for the future usage.

pkg/util/backup/backup.go
w13915984028: 5s might be a bit tight on some slow network env, could it be like 8? and, it is used in two places, could it be deined as a const? thanks.

w13915984028 (commented):
thanks.

pkg/util/backup/backup.go
Yu-Jack: sure, i'm fine with 8s.

pkg/util/timeout.go
```diff
@@ -0,0 +1,34 @@
...
+	}()
+
+	select {
+	case res := <-done:
+		return res.value, res.err
+	case <-ctx.Done():
+		var zero T
+		return zero, fmt.Errorf("timeout after %v", timeout)
+	}
+}
```
WebberHuang1118: Based on current usage, this function always returns `BackupStoreDriver`, so the generic type seems unnecessary here. Do you plan to use this differently in the future?

pkg/webhook/resources/setting/validator.go
```diff
@@ -575,9 +577,17 @@ func (v *settingValidator) validateBackupTarget(setting *v1beta1.Setting) error
 	// S3: https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/s3/s3.go#L38-L87
 	// NFS: https://github.com/longhorn/backupstore/blob/56ddc538b85950b02c37432e4854e74f2647ca61/nfs/nfs.go#L46-L81
 	endpoint := backuputil.ConstructEndpoint(target)
-	if _, err := backupstore.GetBackupStoreDriver(endpoint); err != nil {
-		return werror.NewInvalidError(err.Error(), settings.KeywordValue)
+
+	// There might be a goroutine leak if the driver doesn't end properly,
+	// Although we can pass ctx, but the underlying driver implementation doesn't support it.
+	// So we should be careful when using GetBackupStoreDriver function.
+	_, err = util.RunWithTimeoutAndResult(backuputil.ConnectBackupStoreTimeout, func(_ context.Context) (backupstore.BackupStoreDriver, error) {
```
WebberHuang1118, line 584: Ditto for passing the context.

WebberHuang1118 (commented):
Thanks for the enhancement. Please review the comments.

pkg/util/backup/backup.go
WebberHuang1118: Yes, I think we should pass the context to control the separate goroutine.

pkg/util/timeout.go
Yu-Jack: >  this function always returns BackupStoreDriver

This function is under util. I want to make it more usable, that's why I used generic. I hope this one can be used in other places in the future.

pkg/webhook/resources/setting/validator.go
Yu-Jack: same reason, passing context is good way to let underlying function know we support cancelation.

pkg/util/timeout.go
WebberHuang1118: Cool, thanks.

pkg/util/backup/backup.go
Yu-Jack: We can't because backupstore.GetBackupStoreDriver doesn't support it.

WebberHuang1118 (approved):
LGTM, thanks.

w13915984028 (approved):
LGTM, thanks.

Issue #9843, opened 2026-01-07
[BUG] Raw image upload breaks when using external storage


yulken:
### Describe the Bug

A follow up to https://github.com/harvester/harvester/issues/8636.
<img width="1078" height="534" alt="Image" src="https://github.com/user-attachments/assets/22139a5e-b7d4-4421-89d9-885ec0d64e3b" />
For some reason, uploading ext4.raw fails

### To Reproduce

- Configure Longhorn V2, so there's some external storage for CDI to work with.
- Verify everything works as normal by uploading an ISO and a qcow2 image, ensuring you use an LHv2 storage class
- Create and upload raw image: `qemu-img create ext4.raw 100M ; /usr/sbin/mkfs.ext4 ext4.raw`

### Expected Behavior

Uploads should always either succeed, or, if they're going to fail, they should fail cleanly with some sort of suitable error message.

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.6.0 and v1.7.0
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Virtualized single node 
- Rancher version:

yulken:
There's some discussion regarding the solution for this issue in the following PR: https://github.com/harvester/harvester/pull/9363

noahgildersleeve:
- [x] Test with native Harvester UI PASS

Validated with 1 node KVM/libvirt VM on Harvester `v1.8.0-dev-20260201`. 

<img width="2305" height="380" alt="Image" src="https://github.com/user-attachments/assets/15c6cb99-2c80-40ca-bf4e-385ba7e75a55" />
<img width="2310" height="431" alt="Image" src="https://github.com/user-attachments/assets/971e88d0-5d07-4e96-be80-e9dbe00a1143" />

- [x] Test with upstream Rancher with imported Harvester and `ui-extensions` FAIL

Tested with 1 node KVM/libvirt VM on Harvester `v1.8.0-dev-20260201`, and Rancher  v2.14.0-alpah1 and UI extensions v1.8.0-dev.

<img width="452" height="345" alt="Image" src="https://github.com/user-attachments/assets/43aafc01-0235-431b-a095-fc7fa91e88ab" />


Currently blocked by waiting for new alpha of Rancher v2.14.0

irishgordo:
@noahgildersleeve  I'll take this one since you're currently working on automation

irishgordo:
[...]

PR #9868, opened 2026-01-15
Fix cdi vm image upload failed

brandboat:
#### Problem:

We are using the wrong upload api for `multipart/form-data`, instead of using `/v1alpha1/upload` , we need to switch to `/v1beta1/upload-form`.
However, CDI can only handle upload-form api correctly if form data field name is file, see
https://github.com/kubevirt/containerized-data-importer/blob/release-v1.62/pkg/uploadserver/uploadserver.go#L117
So we have to do some changes in the harvester-ui-extension as currently we use chunk as field name in the form data, see
[https://github.com/harvester/harvester-ui-extension/blob/c169853e5a09ad7545075d8c2[…]ec2/pkg/harvester/models/harvesterhci.io.virtualmachineimage.js](https://github.com/harvester/harvester-ui-extension/blob/c169853e5a09ad7545075d8c2ffb4cb3fdc64ec2/pkg/harvester/models/harvesterhci.io.virtualmachineimage.js#L322)

#### Solution:

- if upload http request header content-type is `multipart/form-data`, use endpoint `/v1alpha1/upload-form` instead of `/v1alpha1/upload`.
- fix an issue that uploaded VMImage never entered initialized state which ends up create VM failed.
- move retry mechaism of UpdateVirtualSizeAndSize from cdi.go to upload.go

#### Related Issue(s):

#9843

#### Test plan:

**_This needs to be tested with patch https://github.com/harvester/harvester-ui-extension/pull/684_**

test all VM image upload scenario

##### VM image creation

1. setup a harvester cluster with lhv2 or lvm ... any third-party storage is fine, what we want is testing vm image with cdi backend
1. [Leap-16.0-online-installer-x86_64.install.iso](https://download.opensuse.org/distribution/leap/16.0/offline/Leap-16.0-online-installer-x86_64.install.iso) upload this iso file in the VM Image page with the 3-rd party storage class.
    - before this patch, this iso file cannot be successfully uploaded to harvester
1. create VM with the uploaded vm image
1. upload the same iso but this time use harvseter-longhorn storage class and create VM

##### airgapped upgrade iso upload

1. goto Settings > server-version > 3-dots button > Upgrade
3. upload harvester iso file, please use a regular one, e.g.harvester-v1.7.0-amd64.iso

Note: if you encountered the `context cancel` issue which cause upload failed, please run below command to increase proxy-read|send-timeout.

```
kubectl annotate ingress rancher-expose \
  -n cattle-system \
  nginx.ingress.kubernetes.io/proxy-read-timeout="1800" \
  nginx.ingress.kubernetes.io/proxy-send-timeout="1800" \
  --overwrite
```

https://github.com/harvester/harvester/issues/9313#issuecomment-3409815722

### DIFF
--- a/pkg/image/cdi/upload.go
+++ b/pkg/image/cdi/upload.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -258,7 +257,7 @@ func (cu *Uploader) DoUpload(vmImg *harvesterv1.VirtualMachineImage, req *http.R
 	}
 	defer uploadResp.Body.Close()
 
-	body, err := ioutil.ReadAll(uploadResp.Body)
+	body, err := io.ReadAll(uploadResp.Body)
 	if err != nil {
 		return fmt.Errorf("failed to read response body: %w", err)
 	}
--- a/pkg/image/cdi/upload.go
+++ b/pkg/image/cdi/upload.go
@@ -17,8 +17,10 @@ import (
 
 	ctlstoragev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/storage/v1"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	cdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
 	uploadcdiv1 "kubevirt.io/containerized-data-importer-api/pkg/apis/upload/v1beta1"
 
@@ -118,12 +120,8 @@ func (cu *Uploader) DoUpload(vmImg *harvesterv1.VirtualMachineImage, req *http.R
 	}
 	logrus.Debugf("Qcow header index: %v, virtual size: %v", qcowHeaderIndex, virtualSize)
 
-	dataContent := rawContent
-	if multipartFormat {
-		dataContent = rawContent[headerEndIndex+4:]
-	}
-	if _, err := cu.vmio.UpdateVirtualSizeAndSize(vmImg, virtualSize, fileSize); err != nil {
-		return fmt.Errorf("failed to update VM Image size and virtual size: %v", err)
+	if err = cu.updateVirtualSizeAndSize(vmImg, virtualSize, fileSize); err != nil {
+		return err
 	}
 
 	// check VMImage status again (for size/virtual size)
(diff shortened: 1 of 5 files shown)


pkg/image/cdi/upload.go
```diff
@@ -32,7 +32,8 @@ import (
 
 const (
 	CDIUploadURLRaw = "cdi-uploadproxy.harvester-system"
-	UploadProxyURI  = "/v1alpha1/upload"
+	UploadURI       = "/v1alpha1/upload"
+	UploadFormURI   = "/v1alpha1/upload-form"
```
ihcsim: any reasons why we don't want to use [`/v1beta1`](https://github.com/kubevirt/containerized-data-importer/blob/cdf8d9085bd6c2fc2f22aa53489e47eb08ba3240/pkg/common/common.go#L285-L287)?

pkg/image/cdi/upload.go
```diff
@@ -125,11 +126,10 @@ func (cu *Uploader) DoUpload(vmImg *harvesterv1.VirtualMachineImage, req *http.R
 	}
 	logrus.Debugf("Qcow header index: %v, virtual size: %v", qcowHeaderIndex, virtualSize)
 
-	dataContent := rawContent
-	if multipartFormat {
-		dataContent = rawContent[headerEndIndex+4:]
```
ihcsim, line 130: looks like `headerEndIndex` can be removed from [line 107](https://github.com/brandboat/harvester/blob/cb6dc9c4b06b3c0c5ffb15177454531d52b4a330/pkg/image/cdi/upload.go#L107).

pkg/image/cdi/upload.go
```diff
@@ -369,3 +374,38 @@ func (cu *Uploader) updateVMImageProgress(vmImg *harvesterv1.VirtualMachineImage
...
+
+func (cu *Uploader) updateVMIWithRetryOnConflict(
+	vmImg *harvesterv1.VirtualMachineImage,
+	updateFunc func(*harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error),
+) (*harvesterv1.VirtualMachineImage, error) {
+	namespace := vmImg.Namespace
+	name := vmImg.Name
+	var updatedVMI *harvesterv1.VirtualMachineImage
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
ihcsim: might want to retain the [original retry implementation ](https://github.com/harvester/harvester/blob/bcb6f66683234e92ac7eb3dced5b329863fc6a28/pkg/image/common/operator.go#L395-L396) which covers other errors like server-side timeout, server unavailable etc.

pkg/image/cdi/upload.go
brandboat: Good question, I don't know either, I just followed the pattern as we use v1alpha1 in upload api.
I think is ok to use v1beta1, will test later, thanks for pointing that out.

pkg/image/cdi/upload.go
```diff
@@ -125,11 +126,10 @@ func (cu *Uploader) DoUpload(vmImg *harvesterv1.VirtualMachineImage, req *http.R
...
 
-	dataContent := rawContent
-	if multipartFormat {
-		dataContent = rawContent[headerEndIndex+4:]
-	}
-	if _, err := cu.vmio.UpdateVirtualSizeAndSize(vmImg, virtualSize, fileSize); err != nil {
+	// retry is needed here as the vm image controller could update the VMImage at the same time
+	if _, err := cu.updateVMIWithRetryOnConflict(vmImg, func(img *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
+		return cu.vmio.UpdateVirtualSizeAndSize(img, virtualSize, fileSize)
+	}); err != nil {
```
WebberHuang1118: The `DoUpload()` function is becoming quite long. Suggest factoring out this part into a separate function.

pkg/image/cdi/upload.go
```diff
@@ -228,7 +228,12 @@ func (cu *Uploader) DoUpload(vmImg *harvesterv1.VirtualMachineImage, req *http.R
 	hookedReader := io.TeeReader(newBody, progress)
 
 	token := retUploadTokenRequest.Status.Token
-	cdiUploadURL := fmt.Sprintf("https://%s%s", CDIUploadURLRaw, UploadProxyURI)
+	var cdiUploadURL string
+	if multipartFormat {
+		cdiUploadURL = fmt.Sprintf("https://%s%s", CDIUploadURLRaw, UploadFormURI)
+	} else {
+		cdiUploadURL = fmt.Sprintf("https://%s%s", CDIUploadURLRaw, UploadURI)
+	}
```
WebberHuang1118: ditto

WebberHuang1118 (commented):
Generally LGTM; just a few suggestions for refactoring, thanks.

WebberHuang1118 (dismissed):

pkg/image/common/operator.go
```diff
@@ -234,12 +231,10 @@ func (vmio *vmiOperator) UpdateSize(old *harvesterv1.VirtualMachineImage, size i
 }
 
 func (vmio *vmiOperator) UpdateVirtualSizeAndSize(old *harvesterv1.VirtualMachineImage, virtualSize, size int64) (*harvesterv1.VirtualMachineImage, error) {
-	return vmio.updateWithRetry(old, func(vmi *harvesterv1.VirtualMachineImage) (*harvesterv1.VirtualMachineImage, error) {
-		logrus.Debugf("Updating VMImage %s/%s: virtualSize=%d, size=%d (resourceVersion: %s)", vmi.Namespace, vmi.Name, virtualSize, size, vmi.ResourceVersion)
-		vmi.Status.VirtualSize = virtualSize
-		vmi.Status.Size = size
-		return vmi, nil
-	})
```
Vicente-Cheng, line 242: Could we keep the debug log here?
~It would be used when we try to figure out the broken upload.~
Not the same thing, but keep this log would be fine

Vicente-Cheng (commented):
just a nit

Vicente-Cheng (approved):
thanks

ihcsim (approved):

Issue #9782, opened 2025-12-23
[BUG] fine tune UI logic to not inject maxSockets on arm clusters


ibrokethecloud:
### Describe the Bug

https://github.com/harvester/harvester-ui-extension/pull/539 added logic in the UI extension where if the VM was not using cpu/mem hot plug then `maxSockets: 1` was added to the VM definition.

This helped avoid setting default hotpluggable CPUs in the VM domain and helped fix https://github.com/harvester/harvester/issues/9059

On ARM nodes hot plugging of CPU/MEM is not supported, however when CPU pinning is used, a specific code path in virt-launchers domain generation logic is triggered which causes a nil pointer exception and causes the launcher pod to crash.

### To Reproduce

Install Harvester v1.6.1 on ARM hosts
Enable CPU pinning
Create a VM with CPU pinning enabled
VM fails to start due to `maxSockets` and virt-wrapper bug being triggered by the presence of `maxSockets: 1`

### Expected Behavior

VM should boot successfully

### Support Bundle for Troubleshooting

NA

### Environment

- Harvester version: v1.6.1
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Ampere Ultra ARM host
- Rancher version:

### Workaround and Mitigation

- Edit VM as yaml
- Remove the `maxSockets: 1` line from the CPU section

ibrokethecloud:
Needs to be backported to v1.7.1 as well

brandboat:
related to upstream issue: https://github.com/kubevirt/kubevirt/issues/16395

a110605:
After discussed with @brandboat , we will ~~hide the cpu/mem hotplug checkbox~~ and won't add maxSockets for ARM machine.

brandboat:
Not only UI, it seems the backend also forces maxSockets to 1 when `enableCPUAndMemoryHotplug` is not enabled.
See: https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/virtualmachine/mutator.go#L344

We have to add additional check here to detect if the VM is an ARM machine, otherwise the maxSocket could still be added to the VM manifest if cpumemory hotplug not enabled. c.c. @Vicente-Cheng, @ibrokethecloud, @FrankYang0529

irishgordo:
@lanfon72  I'll take this one 😄

albinsun:
Test PASS, closed as fixed.

> [!CAUTION]
[...]

PR #9876, opened 2026-01-16
fix: do not patch maxSockets to ARM VM

brandboat:
#### Problem:

On ARM nodes hot plugging of CPU/MEM is not supported, however when CPU pinning is used, a specific code path in virt-launchers domain generation logic is triggered which causes a nil pointer exception and causes the launcher pod to crash.

#### Solution:

do not patch maxSockets to ARM VM.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9782

#### Test plan:

This needs to be tested with UI patch https://github.com/harvester/harvester-ui-extension/pull/674

1. prepare ARM harvester cluster
2. enable cpu pinning to all nodes
3. create a VM, make sure the VM is working


ibrokethecloud:
PR looks good, please rebase.

ibrokethecloud (approved):

Vicente-Cheng (approved):

Issue #9815, opened 2026-01-05
[BUG] Failed upgrade from 1.6.1 to 1.7.0


Kev-Johnson:
### Describe the Bug

upgrading a 4 node cluster from 1.6.1 to 1.7.0. Cluster has a defined storage network, but this was created through the UI, and the known issue sounded like it only applied if the /oem files had been manually edited, so I continued.

2 nodes upgraded with no issue (harvester2, harvester3), 1 node got stuck at post-draining (harvester4), but this often happens with my upgrades, and a manual reboot cleared this issue. Once we were down to the final node (harvester1) things all went wrong - I eventually shut down all VMs, and continued poking at things, then noticed that the NIC names had changed (likely expected due to the change from wicked to NetworkManager), so I created a new cluster network and associated vm network configs. Unfortunately, the separate bond I had in place for storage and VM traffic was down. I was unable to move the storage network across to the new config as there were obviously pods running that were dependent on it.

Ultimately I ended up pulling harvester1 from the cluster and standing up a new 1 node cluster so I could restore the VMs and get them running. Attaching the support bundle in case helpful for others.

### To Reproduce

Create a multinode cluster at 1.6.1 with a separate bond for VM and storage traffic.
Configure storage network on this bond interface.
Start upgrade to 1.7.0.
...
...
...
Once down to the last node, all VMs are running on this node and cannot be migrated. Longhorn can see all nodes in the cluster, but only has storage network connectivity to the node that has not been upgraded.

### Expected Behavior

Network migration would handle the renaming of the interfaces due to the change from wicked to NetworkManager

### Support Bundle for Troubleshooting

Will send by email - unable to upload

### Environment

- Harvester version:
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

innobead:
> Network migration would handle the renaming of the interfaces due to the change from wicked to NetworkManager

@tserong is there any known issue here?

cc @harvester/system

rrajendran17:
@Kev-Johnson  It is not expected interface names to be changed during an upgrade.That looks to be the root cause of the problem.
The network configs under a cluster network in harvester by default expects all interface name under the vlanconfigs to be same and tries to find and configure those interfaces into the bond.During the issue I see the following "Link Not found" error due to which the vlanstatus never became ready and the storage network bond was down without any interfaces.

```
[...]

Issue #9802, opened 2025-12-31
[BUG] Network Configuration Migration Failure During v1.6.0 to v1.7.0 Upgrade


vishrm:
### Describe the Bug

Harvester v1.7.0 upgrade fails to update NetworkManager connection profiles when interface names change from traditional naming (ens1f0/ens1f1) to predictable naming (ens1f0np0/ens1f1np1), causing nodes to lose network connectivity after reboot and preventing them from rejoining the cluster.

During upgrade from Harvester v1.6.0 to v1.7.0, nodes fail to rejoin the cluster after reboot due to network configuration migration failure. After a node reboots with the new v1.7.0 OS, the management network bond fails to initialize because NetworkManager connection profiles still reference old interface names that no longer exist.
This requires console/iDRAC access to manually fix each node before the upgrade can proceed to the next node, effectively making unattended upgrades impossible.

Network Stack Migration:

v1.6.0 used wicked for network management with traditional interface naming
v1.7.0 uses NetworkManager with predictable interface naming

Interface Naming Change:

v1.6.0 interfaces: ens1f0, ens1f1
During v1.7.0 boot, kernel renames: eth0 → ens1f0np0, eth1 → ens1f1np1
NetworkManager connection profiles still reference old names: ens1f0, ens1f1

Result:

Bond slave connections cannot find interfaces ens1f0 and ens1f1
Bond mgmt-bo starts with no slaves attached
Bond cannot generate MAC address without slaves
Management network fails completely
Node cannot rejoin cluster

### To Reproduce

1. Deploy Harvester v1.6.0 cluster with bonded network interfaces for management network
2. Verify network uses traditional interface naming (ens1f0, ens1f1, etc.)
3. Initiate upgrade to v1.7.0 via Web UI
4. Wait for first node to complete image preloading and begin upgrade
5. Observe node reboots after OS upgrade
6. Node fails to rejoin cluster - shows NotReady status
7. Console access shows network connectivity failure
8. Bond has no slaves, cannot generate MAC address

### Expected Behavior

The upgrade process should handle network configuration migration automatically:

1. **Pre-upgrade detection**: Identify current interface names and network manager in use
2. **Configuration migration**: Update NetworkManager profiles with new interface names before first reboot
3. **Validation**: Verify network connectivity after upgrade before marking node as succeeded
4. **Automatic recovery**: If network fails, attempt to fix configuration automatically
5. **Clear error messages**: If manual intervention is required, provide specific instructions

[...]

PR #9877, opened 2026-01-16
fix: add ifname=name:mac to kernel args on upgrade

tserong:
#### Problem:
It's possible for updated NIC drivers to result in persistent network interface names changing after upgrade, e.g. to add port numbers, so `enp6s0f0` becomes `enp6s0f0np0`. This in turn breaks network configuration which is of course based on the original (pre-upgrade) interface names.

#### Solution:
On upgrade, if we add kernel command line arguments for each existing NIC with their current names and MAC addresses, it means that when we boot into a new OS, we are guaranteed to keep the existing interface names.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9815
https://github.com/harvester/harvester/issues/9802

#### Test plan:
To verify the implementation is correct:
- Install Harvester.
- Run `ip link` to get a list of network interfaces. We're interested here in actual ethernet adapaters, so anything that starts with `en`, for example:
  ```
  # ip link | grep -A1 ': en'
  2: enp1s0: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast master mgmt-bo state UP mode DEFAULT group default qlen 1000
      link/ether 52:54:00:a6:2d:da brd ff:ff:ff:ff:ff:ff
  3: enp7s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
      link/ether 52:54:00:e3:19:34 brd ff:ff:ff:ff:ff:ff
  ```
- Upgrade to a version with this fix. After upgrade, you should see `ifname=` lines matching the above interface names and MAC addresses in the grub config, and thus in the kernel command line:
  ```
  # grub2-editenv /oem/grubenv list | grep kernel
  third_party_kernel_args= multipath=off ifname=enp1s0:52:54:00:a6:2d:da enp7s0:52:54:00:e3:19:34
  
  # cat /proc/cmdline
  BOOT_IMAGE=(loop0)/boot/vmlinuz [...] ifname=enp1s0:52:54:00:a6:2d:da enp7s0:52:54:00:e3:19:34 [...] 
  ```

To simulate the fix solving the problem in a libvirt environment:

1. Install Harvester without this fix in a VM.
2. Upgrade to a newer version of Harvester that does _not_ include this fix. When the VM reboots after upgrading the OS, shut down the VM, and edit the NIC config with virt-manager or virsh to put the NIC on a different PCI bus. For example, in my testing, I originally had `bus="0x01"`, which gave me a NIC named `enp1s0`. Then I changed it to `bus="0x08"`, which gives `enp8s0`.
3. Start the VM again (being sure to select the correct grub boot option - it may have landed on "fallback" by default depending on how quick you were with the power off)
4. Verify networking is broken because the NIC name has changed. This simulates the problem.
5. Begin again at step 1, with your virtual NIC back on `bus="0x01"`, then upgrade to a version of Harvester that _does_ include this fix. Again, switch the NIC to a different bus after the OS upgrade.
6. Verify the NIC now has the same pre-upgrade name, even though it's on a different bus, and that networking works correctly. `ip link` will show something like the following, if `enp8s0` was successfully renamed back to `enp1so`:
  ```
     3: enp1s0: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast master mgmt-bo state UP mode DEFAULT group default qlen 1000
      link/ether 52:54:00:a6:2d:da brd ff:ff:ff:ff:ff:ff
      altname enp8s0
  ```

To verify this actually fixes the problem with Intel X710 NICs:

- Install Harvester v1.6.x on a host with those NICs. Verify the NIC names don't have port numbers (e.g. `eno1`)
- Upgrade to v1.7.1 with this fix applied. Verify those NIC names _still_ don't have port numbers, but check the `ip link` output. You'll see something like the following, with `altname` appended, which includes the port number. That's the name the interface _would_ have come up with if we hadn't renamed it:
  ```
  # ip link | grep -A2 ': en'
  2: eno1: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master mgmt-bo state UP mode DEFAULT group default qlen 1000
      link/ether d4:c9:ef:ce:30:68 brd ff:ff:ff:ff:ff:ff
      altname eno1np0
  --
  3: eno2: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc mq state DOWN mode DEFAULT group default qlen 1000
      link/ether d4:c9:ef:ce:30:69 brd ff:ff:ff:ff:ff:ff
      altname eno2np1
  ```

This potentially adds an extra step necessary if a user ever has to replace a NIC in future. Consider:
- Before upgrade NIC names are whatever they are based on their device properties per https://www.freedesktop.org/software/systemd/man/latest/systemd.net-naming-scheme.html
- After upgrade, NIC names are explicitly set by MAC address, to the names used on the host before upgrade.
- Assume the management NIC fails. User replaces it in the same slot, with the exact same model NIC. One of two things will happen now:
  1. The NIC is _not_ one of the Intel X710s that has the problem with port names being appended due to updated drivers. Because it's in the same slot, it should come up with the same name as the old NIC. This works regardless of the `ifname=....` kernel command line argument for this interface name, because the new NIC has a different _MAC_ address, so that argument has no effect.
  2. The NIC _is_ one of the Intel X710s that has the problem with port names being appended. It will come up with `np0` (or whatever) on the end of the name, and networking won't work properly. In this case, the user has to tweak the kernel command line arguments to map the MAC address of the new NIC to the desired name. This will need to be documented.

### DIFF
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -633,26 +633,20 @@ EOF
 
 set_nic_names_by_mac_address() {
   # get current third_party_kernel_args
-  local current_args=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
-
-  # strip any existing ifname= args (in case we're upgrading again, we don't want to end up with duplicates)
-  local new_args=""
-  for a in $current_args ; do
-    echo $a
-    [[ "$a" =~ ^ifname=.* ]] && continue
-    new_args="$new_args $a"
-  done
+  local args=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
 
   # append ifname=name:mac for all existing en* interfaces
   for i in $HOST_DIR/sys/class/net/en* ; do
     [ -e "$i" ] || continue
     local name=$(basename $i)
     local mac=$(cat "$i/address")
-    new_args="$new_args ifname=$name:$mac"
+    # don't add duplicates if there's already an ifname= for this mac address
+    [[ "$args" =~ ifname=[^[:space:]]+:$mac ]] && continue
+    args="$args ifname=$name:$mac"
   done
 
   # save updated third_party_kernel_args
-  chroot $HOST_DIR grub2-editenv /oem/grubenv set third_party_kernel_args="${new_args}"
+  chroot $HOST_DIR grub2-editenv /oem/grubenv set third_party_kernel_args="${args}"
 }
 
 upgrade_os() {
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -631,6 +631,30 @@ stages:
 EOF
 }
 
+set_nic_names_by_mac_address() {
+  # get current third_party_kernel_args
+  local current_args=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
+
+  # strip any existing ifname= args (in case we're upgrading again, we don't want to end up with duplicates)
+  local new_args=""
+  for a in $current_args ; do
+    echo $a
+    [[ "$a" =~ ^ifname=.* ]] && continue
+    new_args="$new_args $a"
+  done
+
+  # append ifname=name:mac for all existing en* interfaces
+  for i in $HOST_DIR/sys/class/net/en* ; do
+    [ -e "$i" ] || continue
+    local name=$(basename $i)
+    local mac=$(cat "$i/address")
+    new_args="$new_args ifname=$name:$mac"
+  done
+
+  # save updated third_party_kernel_args
+  chroot $HOST_DIR grub2-editenv /oem/grubenv set third_party_kernel_args="${new_args}"
+}
+
 upgrade_os() {
   # The trap will be only effective from this point to the end of the execution
   trap clean_up_tmp_files EXIT
@@ -761,6 +785,11 @@ EOF
   if [ -e ${HOST_DIR}/etc/ssh/sshd_config.d/sftp.conf ]; then
     sed -i 's%/usr/lib/ssh/sftp-server%/usr/libexec/ssh/sftp-server%' ${HOST_DIR}/etc/ssh/sshd_config.d/sftp.conf
   fi
+
+  # Work around for
+  #   https://github.com/harvester/harvester/issues/9802
+  #   https://github.com/harvester/harvester/issues/9815
+  set_nic_names_by_mac_address
 
   umount $tmp_rootfs_mount
   rm -rf $tmp_rootfs_squashfs
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -635,6 +635,7 @@ set_nic_names_by_mac_address() {
   # get current third_party_kernel_args
   local args=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
 
+  local update_args=0
   # append ifname=name:mac for all existing en* interfaces
   for i in $HOST_DIR/sys/class/net/en* ; do
     [ -e "$i" ] || continue
@@ -644,10 +645,13 @@ set_nic_names_by_mac_address() {
     # don't add duplicates if there's already an ifname= for this mac address
     [[ "$args" =~ ifname=[^[:space:]]+:$mac ]] && continue
     args="$args ifname=$name:$mac"
+    update_args=1
   done
 
-  # save updated third_party_kernel_args
-  chroot $HOST_DIR grub2-editenv /oem/grubenv set third_party_kernel_args="${args}"
+  if [ $update_args -eq 1 ]; then
+    # save updated third_party_kernel_args
+    chroot $HOST_DIR grub2-editenv /oem/grubenv set third_party_kernel_args="${args}"
+  fi
 }
 
 upgrade_os() {
--- a/package/upgrade/upgrade_node.sh
+++ b/package/upgrade/upgrade_node.sh
@@ -638,6 +638,7 @@ set_nic_names_by_mac_address() {
   # append ifname=name:mac for all existing en* interfaces
   for i in $HOST_DIR/sys/class/net/en* ; do
     [ -e "$i" ] || continue
+    [ -e "$i/address" ] || continue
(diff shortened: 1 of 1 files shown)


tserong:
I've finally managed to get through a test of this on real hardware:

- NICs are Intel X722 with the i40e driver on a HPE host.
- These come up as `eno1` and `eno2` on Harvester v1.6.1, and as `eno1np0` and `eno2np1` on Harvester v1.7.1.

Tested as follows:

- Installed Harvester v1.6.1 with `eno1` as the management interface.
- Upgraded to Harvester v1.7.0, confirmed network configuration was broken as in related issues because the interface was now `eno1np0`.
- Re-installed Harvester v1.6.1.
- Upgraded to Harvester v1.7-head + this patch. Confirmed that networking functioned correctly.
  - Confirmed `ip link` output:
    ```
    # ip link | grep -A2 ': en'
    2: eno1: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master mgmt-bo state UP mode DEFAULT group default qlen 1000
        link/ether d4:c9:ef:ce:30:68 brd ff:ff:ff:ff:ff:ff
        altname eno1np0
    --
    3: eno2: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc mq state DOWN mode DEFAULT group default qlen 1000
        link/ether d4:c9:ef:ce:30:69 brd ff:ff:ff:ff:ff:ff
        altname eno2np1
    ```
  - Confirmed `ifname` params in `/proc/cmdline`:
    ```
    # cat /proc/cmdline 
    BOOT_IMAGE=(loop0)/boot/vmlinuz console=tty1 root=LABEL=COS_STATE cos-img/filename=/cOS/active.img panic=0 net.ifnames=1 rd.cos.oemlabel=COS_OEM rd.cos.mount=LABEL=COS_OEM:/oem rd.cos.mount=LABEL=COS_PERSISTENT:/usr/local rd.cos.oemtimeout=120 audit=1 audit_backlog_limit=8192 intel_iommu=on amd_iommu=on iommu=pt multipath=off ifname=eno1:d4:c9:ef:ce:30:68 ifname=eno2:d4:c9:ef:ce:30:69 rd.emergency=reboot rd.shell=0 panic=5 systemd.crash_reboot systemd.crash_shell=0
    ```

tserong:
Reworked to avoid adding duplicate ifname= entries, but without dropping any existing ones so we don't clobber things like `ifname=netboot:xxxx` or any possible manual changes. Tested by hand from the shell, but need to run this through another upgrade test to double check.

tserong:
Re-tested reworked version through v1.6.1->v1.7.0+fix upgrade in VM environment, and on real hardware. Went through fine.

package/upgrade/upgrade_node.sh
```diff
@@ -631,6 +631,24 @@ stages:
 EOF
 }
 
+set_nic_names_by_mac_address() {
+  # get current third_party_kernel_args
+  local args=$(chroot $HOST_DIR grub2-editenv /oem/grubenv list |grep third_party_kernel_args | awk -F"third_party_kernel_args=" '{print $2}')
+
+  # append ifname=name:mac for all existing en* interfaces
```
w13915984028, line 639: should we only add the mgmt network used en* interfaces? the production deployment can have many nic cards and some might be used/managed by other purpose

package/upgrade/upgrade_node.sh
```diff
@@ -631,6 +631,24 @@ stages:
...
+  for i in $HOST_DIR/sys/class/net/en* ; do
+    [ -e "$i" ] || continue
+    local name=$(basename $i)
+    local mac=$(cat "$i/address")
```
w13915984028, line 644: check if `"$i/address"` is existing to be even secure

package/upgrade/upgrade_node.sh
```diff
@@ -631,6 +631,24 @@ stages:
...
+    # don't add duplicates if there's already an ifname= for this mac address
+    [[ "$args" =~ ifname=[^[:space:]]+:$mac ]] && continue
+    args="$args ifname=$name:$mac"
+  done
+
+  # save updated third_party_kernel_args
```
w13915984028: above loop can get none result or does not change at all, could we add a flag from the loop, to run L649 only when needed

w13915984028 (commented):
Thanks, some concerns about the potential change scope.

package/upgrade/upgrade_node.sh
tserong: It's not only the management interfaces we care about. We need to ensure that all ethernet device names remain the same after upgrade. If not, an upgrade could break e.g. a separately configured storage or cluster network due to interface name changes.

ibrokethecloud (approved):
lgtm. I was able to upgrade a local and seeder provisioned environment where seeder already names nics with a custom name

pre upgrade kernel arguments:
```
dell-190-tink-system:~ # grub2-editenv /oem/grubenv list
extra_cmdline=ifname=netboot:f8:f2:1e:5c:ce:88
third_party_kernel_args=multipath=off
dell-190-tink-system:~ #
```

after upgrade kernel arguments:
```
dell-190-tink-system:~ # grub2-editenv /oem/grubenv list
extra_cmdline=ifname=netboot:f8:f2:1e:5c:ce:88
third_party_kernel_args=multipath=off ifname=eno1:b8:ca:3a:6f:45:b0 ifname=eno2:b8:ca:3a:6f:45:b2 ifname=eno3:b8:ca:3a:6f:45:b4 ifname=eno4:b8:ca:3a:6f:45:b5 ifname=enp130s0f1:f8:f2:1e:5c:ce:89
```

Vicente-Cheng (approved):

package/upgrade/upgrade_node.sh
```diff
@@ -631,6 +631,29 @@ stages:
...
+
+  local update_args=0
+  # append ifname=name:mac for all existing en* interfaces
+  for i in $HOST_DIR/sys/class/net/en* ; do
+    [ -e "$i" ] || continue
+    [ -e "$i/address" ] || continue
+    local name=$(basename $i)
+    local mac=$(cat "$i/address")
+    # don't add duplicates if there's already an ifname= for this mac address
+    [[ "$args" =~ ifname=[^[:space:]]+:$mac ]] && continue
```
megabreit, line 646: This duplicate detection is inadequate. With the used method, the second interfaces of a bond will be dropped (it has the same MAC address as the first one). This results in an incomplete grubenv and broken bonds after reboot.

tserong: Indeed. This has been raised in https://github.com/harvester/harvester/issues/10397

Issue #9708, opened 2025-12-10
[FEATURE] Terraform harvester_virtualmachine Allow enabling CPU and memory hotplug


brinnjoyce:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

We would like to manage the creation/deletion of VMs with Terraform. As part of this we would like to be able to increase the CPU/RAM of a VM without restarting it.

This can be done in Harvester but not with the Terraform provider.

If it's enabled manually in Harvester the resources end up out of sync too, e.g.:
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.test-ubuntu-vm.harvester_virtualmachine.wfxvm will be updated in-place
  ~ resource "harvester_virtualmachine" "wfxvm" {
      ~ cpu                     = 1 -> 4
        id                      = "inf-dev/inf-dv-test10"
      ~ memory                  = "64Gi" -> "16Gi"
        name                    = "inf-dv-test10"
        tags                    = {}
        # (14 unchanged attributes hidden)

        # (3 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```
It looks like the CPU goes of sync because Harvester changed it from 4 cores 1 socket to 1 core 4 sockets when hotplug was enabled. So allowing terraform to apply this change results in the VM having 16 CPUs - but with a resource limit of 4.

These both look like a limitation of the `github.com/harvester/harvester/pkg/builder` library you're using - e.g. this only sets `spec.domain.resources.limits.memory` for the memory config and depends on Harvester setting `spec.domain.memory.guest` to match this, which doesn't work with hotplug enabled as the limit needs to be higher than that.

### Describe the solution you'd like

An added option to enable CPU and Memory hotplug for the harvester_virtualmachine resource.
It should likely allow configuring the maximum CPU and Memory too.

### Describe alternatives you've considered

Tried enabling this outside of Terraform but this causes state drift which is incorrectly reconciled.

w13915984028:
cc @FrankYang0529 @m-ildefons

Vicente-Cheng:
Hi @m-ildefons, 
could we move this one to Ready for Testing?
I saw the above PR was merged

m-ildefons:
This is not yet ready.
This issue needs two PRs:

1) a change to the VMBuilder interface (https://github.com/harvester/harvester/pull/9879)
[...]

PR #9879, opened 2026-01-16
VMBuilder: Add methods for CPU/memory requests

m-ildefons:
Add differentiated resources for CPU and memory resource requests to the VMBuilder. These functions allow the caller to make specific arrangements for the CPU topologies and for the memory of the guest VM. This is necessary to facilitate CPU and memory hotplugging with the VMBuilder

related-to: harvester/harvester#9708

#### Problem:

The VMBuilder does not expose sufficiently flexible methods for setting CPU topology and memory attributes to facilitate CPU and memory hotplugging.

#### Solution:

Add more flexible methods for expressing CPU topology and memory attributes to the VMBuilder

#### Related Issue(s):

harvester/harvester#9708

### DIFF
--- a/pkg/builder/vm_test.go
+++ b/pkg/builder/vm_test.go
@@ -48,19 +48,22 @@ func TestCPUCores(t *testing.T) {
 			continue
 		}
 
-		if !tc.expectError {
-			if testVM == nil || testVM.Spec.Template.Spec.Domain.CPU == nil {
-				t.Errorf("test %s: unexpected nil value for domain cpu", tc.description)
-				continue
-			}
-			if testVM.Spec.Template.Spec.Domain.CPU.Cores < 1 {
-				t.Errorf("test %s: unexpected value for cpu cores", tc.description)
-				continue
-			}
-			if tc.expectation != testVM.Spec.Template.Spec.Domain.CPU.Cores {
-				t.Errorf("test %s: unexpected cpu cores value", tc.description)
-				continue
-			}
+		if tc.expectError {
+			continue
+		}
+
+		if testVM == nil || testVM.Spec.Template.Spec.Domain.CPU == nil {
+			t.Errorf("test %s: unexpected nil value for domain cpu", tc.description)
+			continue
+		}
+
+		if tc.expectation != testVM.Spec.Template.Spec.Domain.CPU.Cores {
+			t.Errorf("test %s: unexpected cpu coresvalue: %d, expected: %d",
+				tc.description,
+				testVM.Spec.Template.Spec.Domain.CPU.Cores,
+				tc.expectation,
+			)
+			continue
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


w13915984028 (dismissed):
LGTM, thanks.

BTW, could we add some tests to vm_test.go to cover those new methods?

pkg/builder/vm_test.go
```diff
@@ -2,15 +2,266 @@ package builder
...
+			continue
+		}
+
+		if !tc.expectError {
+			if testVM == nil || testVM.Spec.Template.Spec.Domain.CPU == nil {
+				t.Errorf("test %s: unexpected nil value for domain cpu", tc.description)
+				continue
+			}
+			if testVM.Spec.Template.Spec.Domain.CPU.Threads < 1 {
+				t.Errorf("test %s: unexpected value for cpu cores", tc.description)
```
w13915984028: L174~L177 could be merged to L178, which prints the expected value and real value

pkg/builder/vm_test.go
```diff
@@ -2,15 +2,266 @@ package builder
...
+			builder:     NewVMBuilder("test").GuestMemory("foobar"),
+			expectation: nil,
+			expectError: true,
+		},
+	}
+
+	for _, tc := range testcases {
+		testVM, err := tc.builder.VM()
+
+		if tc.expectError && err == nil {
```
w13915984028, line 232: merge L223~L231 to:
```
		if tc.expectError != (err != nil) {
			t.Errorf("test %s: expectError %v, but got %v", tc.description, tc.expectError, err != nil )
			continue
		}
```

pkg/builder/vm_test.go
```diff
@@ -2,15 +2,266 @@ package builder
...
+			t.Errorf("test %s: unexpected error", tc.description)
+			continue
+		}
+
+		if !tc.expectError && err != nil {
+			t.Errorf("test %s: expected error", tc.description)
+			continue
+		}
+
+		if !tc.expectError {
```
w13915984028:
```
		if tc.expectError {
			continue
		}
```

and reduce the indent of following code

w13915984028 (commented):
A few code placement suggestions, thanks.

pkg/builder/vm_test.go
m-ildefons: This conditional is wrong.
Consider this:
```golang
err := nil
expectError := false
```

Then `err != nil` will evaluate to `false`. Meaning the conditional will simplify to `false != false`, which itself is `false`. However it should overall be true.

Did you mean?
```golang
if tc.expectError == (err != nil) {
...
}
```

pkg/builder/vm_test.go
m-ildefons: Ok, but that makes it even harder to read because you're relying on fall-through logic, which isn't obvious.

pkg/builder/vm_test.go
m-ildefons: ok

pkg/builder/vm_test.go
w13915984028: yeah, it seems to be 

```
if ! (tc.expectError == (err != nil)) {
  expect .., but got ...
}
```

w13915984028 (approved):
LGTM, thanks.

pkg/builder/vm_test.go
m-ildefons: No that's still wrong

ihcsim (approved):

PR #9895, opened 2026-01-21
fix: use longhorn-static sc for upgrade vm image

brandboat:
#### Problem:

StorageClass validator blocks the storage class deletion even if its related vm image is marked as deleting.

This is a problem when we perform the following operation sequentially:
1. delete vmimage
2. delete the storage class that relates to the vmimage

In step 1 k8s marks vmimage as deleting, but the vm image has not been fully removed yet, then step 2 fails because the storage class validate treat it as still in-use.

#### Solution:

see discussion in https://github.com/harvester/harvester/pull/9785, in the end we decided to use `longhorn-static` for upgrade vm image,

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9720

#### Test plan:

well-described in issue description, see https://github.com/harvester/harvester/issues/9720

### DIFF
--- a/pkg/webhook/resources/persistentvolumeclaim/validator.go
+++ b/pkg/webhook/resources/persistentvolumeclaim/validator.go
@@ -155,17 +155,7 @@ func (v *pvcValidator) Update(_ *types.Request, oldObj runtime.Object, newObj ru
 
 func (v *pvcValidator) Create(request *types.Request, newObj runtime.Object) error {
 	newPVC := newObj.(*corev1.PersistentVolumeClaim)
-
-	if newPVC.Spec.StorageClassName == nil {
-		return nil
-	}
-
-	if *newPVC.Spec.StorageClassName == util.StorageClassLonghornStatic || *newPVC.Spec.StorageClassName == util.StorageClassVmstatePersistence {
-		message := fmt.Sprintf("can not create volume with the reserved storage class %s", *newPVC.Spec.StorageClassName)
-		return werror.NewInvalidError(message, "spec.storageClassName")
-	}
-
-	return nil
+	return v.validateInternalUsage(newPVC)
 }
 
 func (v *pvcValidator) checkGoldenImageAnno(pvc *corev1.PersistentVolumeClaim) error {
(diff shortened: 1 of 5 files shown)


pkg/webhook/resources/persistentvolumeclaim/validator_test.go
```diff
@@ -0,0 +1,236 @@
...
+					Name:      "test-pvc",
+					Namespace: "default",
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					StorageClassName: ptr.To(util.StorageClassVmstatePersistence),
+				},
+			},
+			expectError:   true,
+			errorContains: "reserved storage class",
+		},
```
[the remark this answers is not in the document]
brandboat: no need, covered by TestIsBelongToUpgradeImage

pkg/webhook/resources/persistentvolumeclaim/validator_test.go
```diff
@@ -0,0 +1,236 @@
...
+				assert.NotNil(t, err, tc.name)
+				if tc.errorContains != "" {
+					assert.Contains(t, err.Error(), tc.errorContains, tc.name)
+				}
+			} else {
+				assert.Nil(t, err, tc.name)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
brandboat: covered by TestIsBelongToUpgradeImage

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -161,8 +161,14 @@ func (v *pvcValidator) Create(request *types.Request, newObj runtime.Object) err
 	}
 
 	if *newPVC.Spec.StorageClassName == util.StorageClassLonghornStatic || *newPVC.Spec.StorageClassName == util.StorageClassVmstatePersistence {
```
WebberHuang1118: I realize this is unrelated to the changes in this PR, but is it possible to extract this logic into a utility function? This would allow for a more compact code snippet, e.g.:
```
	if !internelSC(newPVC) {
		return nil
	}

	isInternalUsage, err := v.isInternalUsage(newPVC)
	if err != nil {
		return werror.NewInternalError(fmt.Sprintf("failed to check if pvc %s/%s is for internal usage: %v", newPVC.Namespace, newPVC.Name, err))
	}
	if !isInternalUsage {
		message := fmt.Sprintf("can not create volume with the reserved storage class %s", *newPVC.Spec.StorageClassName)
		return werror.NewInvalidError(message, "spec.storageClassName")
	}

	return nil
```

WebberHuang1118 (dismissed):
LGTM, just a nit, thanks for the fix.

pkg/webhook/resources/persistentvolumeclaim/validator.go
brandboat: fixed in the latest commit, thanks

WebberHuang1118 (dismissed):

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -198,3 +188,54 @@ func (v *pvcValidator) checkGoldenImageAnno(pvc *corev1.PersistentVolumeClaim) e
 	}
 	return nil
 }
+
+func (v *pvcValidator) validateInternalUsage(pvc *corev1.PersistentVolumeClaim) error {
+	if !hasReservedSC(pvc) {
+		return nil
+	}
+
+	isBelongToUpgradeImage, err := v.isBelongToUpgradeImage(pvc)
```
w13915984028, line 197: L193 checks two SC,
```
StorageClassVmstatePersistence = "vmstate-persistence" and `longhorn-static`
```

From L197, is it dedicated for `longhorn-static` ? I did not find the  usage of StorageClassVmstatePersistence on `harvester/harvester`, if L197 is only upon `longhorn-static`, or both?

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -198,3 +188,54 @@ func (v *pvcValidator) checkGoldenImageAnno(pvc *corev1.PersistentVolumeClaim) e
...
+}
+
+func (v *pvcValidator) isBelongToUpgradeImage(pvc *corev1.PersistentVolumeClaim) (bool, error) {
+	for _, owner := range pvc.OwnerReferences {
+		// upgrade vm image will create 3 pvcs
+		// 1. image-xxxxx pvc (owner kind: DataVolume) with longhorn-static storage class, the owner here is the underlying data volume of upgrade vm image
+		// 2. prime-{uuid} pvc (owner kind: PersistentVolumeClaim) with longhorn-static storage class, the owner here is the image-xxxxx pvc
+		// 3. prime-{uuid}-scratch pvc (owner kind: Pod) with default storage class, generally harvester-longhorn, the owner here is the cdi importer/upload pod
+		// note that vm image and data volume and image-xxxxx pvc share the same name
+		if owner.Kind != "DataVolume" && owner.Kind != "PersistentVolumeClaim" {
```
w13915984028: The `DataVolume` seems not exposed as a const by kubevirt, could Harvester define one for internal usage?

pkg/webhook/resources/persistentvolumeclaim/validator.go
```diff
@@ -198,3 +188,54 @@ func (v *pvcValidator) checkGoldenImageAnno(pvc *corev1.PersistentVolumeClaim) e
...
+
+		if val, ok := upgradeImage.Annotations[util.AnnotationUpgradeImage]; ok && val == "True" {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func hasReservedSC(pvc *corev1.PersistentVolumeClaim) bool {
```
w13915984028: `hasReservedSC` seems like `isUsingReservedSC`

w13915984028 (dismissed):
LGTM, thanks.

pkg/webhook/resources/persistentvolumeclaim/validator.go
brandboat: Are you suggesting a better name? I couldn’t find any method called isUsingReservedSC.

pkg/webhook/resources/persistentvolumeclaim/validator.go
brandboat: isBelongToUpgradeImage is only for `longhorn-static`, will add a check in this method.

brandboat:
gentle ping @w13915984028, all comments are addressed. PTAL, thanks

w13915984028 (dismissed):
LGTM, thanks for the fast fix.

w13915984028 (approved):
LGTM, thanks.

WebberHuang1118 (approved):

Issue #9759, opened 2025-12-17
[BUG] Node status shows Cordoned when the node is powered off


asc5543:
### Describe the Bug

In e2e automation test, if a node is powered off, we expected the status on Harvester would be `in-progress` or `unavailable`, and the status is as expected in harvester `v1.6.1`.
And the test fail in harvester `v1.7.0-rc6`, the status is `draining` or `Cordoned`

<img width="1906" height="1050" alt="Image" src="https://github.com/user-attachments/assets/4213738a-d9a2-4919-899d-3fa90f1ab2d5" />
<img width="1906" height="1050" alt="Image" src="https://github.com/user-attachments/assets/6f90e6e9-b602-4dd9-b7e0-d2f2e3cbc78f" />

Refer to https://github.com/harvester/harvester/pull/9480, this is new changes in `v1.7.0`, we may need to discuss which status is better.

### To Reproduce

1. Create a 2 node cluster
2. Power off one node
3. Check the status on Harvester UI

### Expected Behavior

The node status should be `in-progress` or `unavailable`

### Support Bundle for Troubleshooting

[supportbundle_d82cfac5-fb42-461f-be18-969ab1c51085_2025-12-16T03-06-15Z.zip](https://github.com/user-attachments/files/24206109/supportbundle_d82cfac5-fb42-461f-be18-969ab1c51085_2025-12-16T03-06-15Z.zip)

### Environment

- Harvester version: `v1.7.0-rc6`
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

Vicente-Cheng:
We need to review the host state machine for v1.7.1.

cc @a110605, @ibrokethecloud

innobead:
Added `regression`. 

@harvester/qa, please add `regression` label when opening a new bug, and it's a regression.

> In e2e automation test, if a node is powered off, we expected the status on Harvester would be `in-progress` or `unavailable`, and the status is as expected in harvester `v1.6.1`. And the test fail in harvester `v1.7.0-rc6`, the status is `draining` or

Vicente-Cheng:
We did not display the node condition directly on the UI. It's converted. So maybe need @a110605 help with that.

a110605:
[...]

PR #9897, opened 2026-01-21
improve node failure detection logic

ibrokethecloud:
#### Problem:

We introduced changes as part of PR https://github.com/harvester/harvester/pull/9480 to add a custom condition when a node drain is triggered due to node failure. The condition is picked up by wrangler status aggregation, and used by steve to render a more informative message in Harvester UI.

However as observed even after the out-of-service taint has been setup, the drain condition is not updated, which results in the UI still rendering the `draining` message.

#### Solution:

PR introduces two commits
* first commit fixes the condition reconcile logic to ensure Drain condition is marked true to indicate drain is complete, which subsequently ensures UI renders a node unavailable message.
* second commit adds a node annotation controller, which reconciles conditions in a predefined order to eventually update a pair of annotations. These annotations can be used by UI, which will allow us in the future to control node messages via harvester controller changes
* 
#### Related Issue(s):

https://github.com/harvester/harvester/issues/9759
https://github.com/harvester/harvester/issues/8945


pkg/controller/master/node/node_down_controller.go
```diff
@@ -126,34 +126,28 @@ func (h *nodeDownHandler) checkNodeReady(node *corev1.Node) error {
...
+		// try and check if condition exists
+		drainCondition := getNodeCondition(node.Status.Conditions, HarvesterNodeCondDrained)
+		// both taints exist so lets mark condition as true
 		if getNodeTaint(node.Spec.Taints, virtconfig.NodeDrainTaintDefaultKey) != nil && getNodeTaint(node.Spec.Taints, corev1.TaintNodeOutOfService) != nil {
-			// already added both taints, no need to process again
-			return nil
+			if drainCondition != nil && drainCondition.Status == corev1.ConditionFalse {
+				_, err := h.createOrUpdateNodeCondition(node, HarvesterNodeCondDrained, corev1.ConditionTrue, HarvesterNodeDrainedCondReasonDrained, HarvesterNodeDrainedCondMsg)
+				return err
+			}
```
WebberHuang1118, line 139: This part is a bit verbose. Can we refactor it into a utility function?

pkg/controller/master/node/node_down_controller.go
```diff
@@ -180,48 +174,40 @@ func (h *nodeDownHandler) checkNodeReady(node *corev1.Node) error {
...
-		return true, nil
-	}
-	waitEnough, err := h.tryWaitForCondTimeout(cond, vmMigrationTimeout)
-	if err != nil {
-		// error case, dont' care about waitEnough
-		return false, err
-	}
-	if !waitEnough {
-		return false, nil
+func (h *nodeDownHandler) addOutOfServiceTaintAfterMigrationTimeout(node *corev1.Node, cond *corev1.NodeCondition, vmMigrationTimeout int64) error {
```
WebberHuang1118: This function name is a bit long. Suggest shortening it, e.g., `waitAndAddOutOfServiceTaint`.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -287,46 +273,41 @@ func (h *nodeDownHandler) fetchVMForceResetPolicy() (*settings.VMForceResetPolic
...
 	return nil
 }
 
 func (h *nodeDownHandler) cleanupResources(node *corev1.Node) error {
-	if err := h.addTaints(node, nonGracefulTaint); err != nil {
+	if _, err := h.addTaints(node, nonGracefulTaint); err != nil {
 		return fmt.Errorf("failed to add non-graceful taint to node %s: %w", node.Name, err)
 	}
 	return nil
 }
```
WebberHuang1118: The function name doesn't intuitively reflect its behavior. Suggest renaming it to `forceCleanupWithOutOfServiceTaint`, for example.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -126,34 +126,28 @@ func (h *nodeDownHandler) checkNodeReady(node *corev1.Node) error {
 			return nil
 		}
 
+		// try and check if condition exists
+		drainCondition := getNodeCondition(node.Status.Conditions, HarvesterNodeCondDrained)
+		// both taints exist so lets mark condition as true
 		if getNodeTaint(node.Spec.Taints, virtconfig.NodeDrainTaintDefaultKey) != nil && getNodeTaint(node.Spec.Taints, corev1.TaintNodeOutOfService) != nil {
-			// already added both taints, no need to process again
-			return nil
+			if drainCondition != nil && drainCondition.Status == corev1.ConditionFalse {
```
WebberHuang1118: Should we initialize `drainCondition` if it doesn't exist?

WebberHuang1118 (commented):
Few suggestions. Please take a look. Thanks.

Vicente-Cheng (dismissed):
Thanks!

pkg/controller/master/node/node_down_controller.go
ibrokethecloud: not needed, it will only be added after the first kubevirt taint is applied, so the logic pointed out by copilot is not a reachable condition

WebberHuang1118: I realize this might not be strictly necessary, as the condition should already be initialized here: [https://github.com/ibrokethecloud/harvester/blob/6bab8aadea8fc11db39c32c50892aff5ec16fd11/pkg/controller/master/node/node_down_controller.go#L284-L295](https://github.com/ibrokethecloud/harvester/blob/6bab8aadea8fc11db39c32c50892aff5ec16fd11/pkg/controller/master/node/node_down_controller.go#L284-L295).

However, since we use `createOrUpdateNodeCondition` at [https://github.com/ibrokethecloud/harvester/blob/6bab8aadea8fc11db39c32c50892aff5ec16fd11/pkg/controller/master/node/node_down_controller.go#L134-L136](https://github.com/ibrokethecloud/harvester/blob/6bab8aadea8fc11db39c32c50892aff5ec16fd11/pkg/controller/master/node/node_down_controller.go#L134-L136), checking `drainCondition != nil` seems redundant. WDYT?

ibrokethecloud: yeah makes sense.. i will drop the check since it will never get to that condition

pkg/controller/master/node/node_down_controller.go
ibrokethecloud: may be hard to have a common utility function sine this is used in different places and is already a utility function

pkg/controller/master/node/node_down_controller.go
ibrokethecloud: it is a helper function and is used in multiple locations https://github.com/harvester/harvester/pull/9897/files#diff-57cd35f115224f491e56aae03e79e635b2afeb17d64c1d79b314c98eaafb0702L309

pkg/controller/master/node/node_down_controller.go
WebberHuang1118: How about this? And we can reuse `hasKubevirtDrainTaint` for `addKubevirtTaintAfterExtraTimeout` and `hasOutOfServiceTaint` for `addOutOfServiceTaintAfterMigrationTimeout`.

```
		// check if both taints exist
		hasKubevirtDrainTaint := getNodeTaint(node.Spec.Taints, virtconfig.NodeDrainTaintDefaultKey) != nil
		hasOutOfServiceTaint := getNodeTaint(node.Spec.Taints, corev1.TaintNodeOutOfService) != nil

		// both taints exist so lets mark condition as true
		if hasKubevirtDrainTaint && hasOutOfServiceTaint {
			if drainCondition.Status == corev1.ConditionFalse {
				_, err := h.createOrUpdateNodeCondition(node, HarvesterNodeCondDrained, corev1.ConditionTrue, HarvesterNodeDrainedCondReasonDrained, HarvesterNodeDrainedCondMsg)
				return err
			}
		}
```

pkg/controller/master/node/node_down_controller.go
WebberHuang1118: Seems not relevant? I mean the function name cleanupResources appears not intuitively reflect its behavior

ibrokethecloud: does `forceNodeCleanup` work better?

WebberHuang1118: > does `forceNodeCleanup` work better?

LGTM, thanks.

Vicente-Cheng (dismissed):

WebberHuang1118 (approved):

Vicente-Cheng (approved):

Vicente-Cheng:
@mergifyio backport v1.7

ibrokethecloud:
@mergify backport v1.7

w13915984028:
Thanks for the PR, please double check if there is a small race window:

https://github.com/harvester/harvester/blob/b2d4b1e62cfc9ee8cb8e51aff3b12611c0a4f36d/pkg/webhook/resources/upgrade/validator.go#L391-L397

https://github.com/harvester/harvester/blob/b2d4b1e62cfc9ee8cb8e51aff3b12611c0a4f36d/pkg/controller/master/node/node_down_controller.go#L109-L113

When a node is put to ready, Harvester upgrade object is created, and the `removeHarvesterNodeDrainedCond` is not called.

Issue #5859, opened 2024-05-23
[TASK] Integrate harvester-installer to harvester/harvester repo


bk201:
### Description
This issue proposes the migration and integration of the `harvester-installer` repository into the main `harvester/harvester` codebase. Consolidating these repositories will streamline the development lifecycle, unify our CI/CD pipelines, and eliminate synchronization overhead.

---

#### 1. Streamlined Feature Development & Local Testing
Currently, features spanning both the core controller and the installer require cross-repo PRs, which complicates dependency management and local validation.
* **Unified Workflow:** Developers can implement and test full-stack changes (from installation logic to controller behavior) within a single PR.
* **Atomic Commits:** Ensure that breaking changes in the installer are always in sync with the required controller versions.

#### 2. Resolution of Build Artifact Collisions
Our current decoupled build process creates friction in the release pipeline:
* **Race Conditions:** ISOs are currently triggered by events in both repositories (merges, tags, and cron jobs), leading to potential file collisions and version confusion.
* **Simplified Maintenance:** Critical images, such as the `harvester-cluster-repo`, are currently generated on the installer side. Integration allows for a centralized build logic where all artifacts are derived from a single source of truth.

#### 3. Enhanced Test Coverage (Vagrant & Integration)
Integration allows us to leverage installer-specific testing suites for core changes:
* **Cross-Pollination:** Changes made to the Harvester controller can be immediately validated by the existing Vagrant-based installation tests.
* **Shift Left:** Catch installation regressions caused by controller changes earlier in the development cycle.

---

### Proposed Task List
* [ ] Merge `harvester-installer` code into the main repository.
* [ ] Merge `harvester-installer` CI workflows into the main repository.
* [ ] Update Dockerfiles and ISO building scripts to reflect the new directory structure.
* [ ] Consolidate GitHub Actions and release automation to prevent artifact overwrites.
* [ ] Update documentation and developer contribution guides
* [ ] Archive the harvester-installer repo once all branches are EOM.

w13915984028:
After the ISO built is unified and moved to Harvester, we can further return failure when encounter any version mismatching of addons. That was discussed https://github.com/harvester/harvester-installer/pull/750#discussion_r1627069473

tserong:
As part of this work we should consider ways to make it easier to include modified dashboard tarballs in the ISO images (see discussion at https://forums.rancher.com/t/ui-dashboard-customisation/43639/1)

innobead:
[...]

PR #9908, opened 2026-01-22
fix: build: always pull harvester-installer image

tserong:
#### Problem:
When building Harvester locally, it's possible for your docker cache to have an older version of the harvester-installer image, which means the upgrade container of your local build will use a potentially stale harvester-installer binary. 

#### Solution:
We can solve this by pulling the container on every build.  This should have no real effect on our CI or release builds which run in a clean environment each time.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9909

#### Test plan:
N/A

### DIFF
--- a/scripts/package-upgrade
+++ b/scripts/package-upgrade
@@ -27,7 +27,10 @@ cp ../../bin/upgrade-helper .
 
 BUILD_ARGS="--build-arg ARCH=${ARCH}"
 if [ -n "${HARVESTER_INSTALLER_REF}" ]; then
+    docker pull rancher/harvester-installer:${HARVESTER_INSTALLER_REF}
     BUILD_ARGS="${BUILD_ARGS} --build-arg HARVESTER_INSTALLER_REF=${HARVESTER_INSTALLER_REF}"
+else
+    docker pull rancher/harvester-installer:master-head
 fi
 docker build -f ${DOCKERFILE} ${BUILD_ARGS} -t ${IMAGE} .
 echo Built ${IMAGE}


ibrokethecloud (approved):

wheatdog:
https://github.com/harvester/harvester/blob/ec8b006212df145c217fc0753ee33b02e0b36997/package/upgrade/Dockerfile#L1-L3

I would like to confirm that the reason that `package-upgrade` could use the stale local cache of `rancher/harvester-installer` is due to that we introduce harvester-installer as a dependency of harvester-upgrade in https://github.com/harvester/harvester/commit/1184a8f6f367696bc12b2d4bc80bd9d5c51e7320?

I think we can have this merged first. However, somehow it feels like there could be a bigger problem behind this issue about our repository topology: there is circular dependency between harvester and harvester-installer.

wheatdog (approved):

brandboat:
> there is circular dependency between harvester and harvester-installer.

I remember we discussed merging the harvester and harvester-installer repositories before, but I’m not sure why we decided not to proceed in the end.

tserong:
> I would like to confirm that the reason that `package-upgrade` could use the stale local cache of `rancher/harvester-installer` is due to that we introduce harvester-installer as a dependency of harvester-upgrade in [1184a8f](https://github.com/harvester/harvester/commit/1184a8f6f367696bc12b2d4bc80bd9d5c51e7320)?

Yes, but ordinarily this shouldn't be a problem for our regular daily builds and CI, because those are built in a clean environment each time, so there shouldn't be a cache anywhere to _be_ stale. @ibrokethecloud and I have only hit this problem when doing local builds.

If we were making changes in future to the harvester-installer binary, which were also needed by the upgrade container, then provided the change to harvester-installer is merged and published before any matching change is made to harverster, there's no problem.

> there is circular dependency between harvester and harvester-installer.

Yeah, this does kinda suck. It will (or should) go away eventually as part of https://github.com/harvester/harvester/issues/5859

Issue #9919, opened 2026-01-22
[BUG] Not able to create VirtualMachines with persistent vTPM and UEFI on Harvester 1.7.0 version


JavierLagosChanclon:
### Describe the Bug

When I try to create a VirtualMachine on a new Harvester cluster v1.7.0 version by selecting vTPM and UEFI persistent feature enabled it fails by showing the following event on the VirtualMachine.

```console
  - lastTransitionTime: "2026-01-21T16:11:00Z"
    message: 'admission webhook "validator.harvesterhci.io" denied the request: can
      not create volume with the reserved storage class vmstate-persistence'
    reason: FailedBackendStorageCreate
```

I would say that the change implemented here https://github.com/harvester/harvester/issues/8561 has broken this action on Harvester. 

### To Reproduce

Create a Harvester on v1.7.0 version and create a VirtualMachine with UEFI and vTPM persistency enabled

### Expected Behavior

I would expect the VirtualMachine to be created as it is on previous Harvester versions.

### Support Bundle for Troubleshooting

No support bundle yet.

### Environment

- Harvester version: v1.7.0 version

w13915984028:
I reproduced it, when create new VM and tick below options

<img width="2606" height="230" alt="Image" src="https://github.com/user-attachments/assets/a597b6a1-1f2d-4b16-b21b-6ea67da8966a" />

```
    - lastProbeTime: null
      lastTransitionTime: "2026-01-22T08:29:32Z"
      message: 'admission webhook "validator.harvesterhci.io" denied the request:
        can not create volume with the reserved storage class vmstate-persistence'
      reason: FailedBackendStorageCreate
      status: "False"
      type: Synchronized
    currentCPUTopology:
```

It might be a side-effect of  https://github.com/harvester/harvester/issues/8561

also check the discussion on PR https://github.com/harvester/harvester/pull/9895

brandboat:
Workaround
1. copy the storage class `vmstate-persistence` and name it to `vmstate-persistence1` (or something else)
2. edit the KubeVirt manifest `.spec.configuration.vmStateStorageClass` and set it to `vmstate-persistence1`

WebberHuang1118:
[...]

PR #9921, opened 2026-01-22
fix: do not block vmstate-persistence usage

brandboat:
#### Problem:

https://kubevirt.io/user-guide/compute/persistent_tpm_and_uefi_state/#tpm-with-persistent-state

if user set TPM with persistent state to the VM, kubevirt will create a pvc with sc `vmstate-persistence`. While we block any PVC with sc `vmstate-persistence` in webhook validtor, see https://github.com/harvester/harvester/pull/9366.

#### Solution:

do not block vmstate-persistence usage, this PR works as a band-aid, if we want to block vmstate-persistence usage, we have to figure out other solution.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9919

#### Test plan:
see issue desc

https://github.com/user-attachments/assets/4400777f-3f94-4b39-9180-1a146924f5e8

### DIFF
--- a/pkg/webhook/resources/persistentvolumeclaim/validator.go
+++ b/pkg/webhook/resources/persistentvolumeclaim/validator.go
@@ -239,8 +239,5 @@ func (v *pvcValidator) isBelongToUpgradeImage(pvc *corev1.PersistentVolumeClaim)
 }
 
 func isUsingReservedSC(pvc *corev1.PersistentVolumeClaim) bool {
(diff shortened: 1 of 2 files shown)


WebberHuang1118 (approved):

brandboat:
an alternative solution: kubevirt add label- `persistent-state-for` to that extra persistent state pvc, perhaps we can leverage on this label.

Vicente-Cheng (approved):

Issue #2661, opened 2022-08-12
[FEATURE] Support backup and restore through harvester-csi-driver in guest cluster


FrankYang0529:
**Is your feature request related to a problem? Please describe.**
Currently, when users create PVC through harvester-csi-driver in the guest cluster, they don't have a convenient way to backup/restore PVCs in LH. I try following the steps, but I can't restore PVC to the guest cluster.

1. In LH page, click `Create Backup` for a PVC.
2. Restore latest Backup to a Volume.
3. Click `Create PV/PVC` on the Volume.
4. In Harvester page, click `Add Volume` to the guest cluster.
5. Try to apply following PV/PVC in the guest cluster, but it doesn't work.
```
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/provisioned-by: driver.harvesterhci.io
  name: pvc-37f9bfdf-3a57-46d3-b45f-ffab7068fa29
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 1Gi
  csi:
    driver: driver.harvesterhci.io
    fsType: ext4
    volumeAttributes:
      storage.kubernetes.io/csiProvisionerIdentity: 1660213625498-8081-driver.harvesterhci.io
    volumeHandle: pvc-37f9bfdf-3a57-46d3-b45f-ffab7068fa29
  persistentVolumeReclaimPolicy: Delete
  storageClassName: harvester
  volumeMode: Filesystem
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    volume.beta.kubernetes.io/storage-provisioner: driver.harvesterhci.io
    volume.kubernetes.io/storage-provisioner: driver.harvesterhci.io
  name: pvc-from-harvester
  namespace: default
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: harvester
  volumeMode: Filesystem
  volumeName: pvc-37f9bfdf-3a57-46d3-b45f-ffab7068fa29
```

**Describe the solution you'd like**
Add support in harvester-csi-driver.

FrankYang0529:
cc @yaocw2020 Do you have any idea for users to manually restore PVC to the guest cluster now? Thank you.

bk201:
@Vicente-Cheng Please help check if backup/restore can be supported via CSI.

anultravioletaurora:
[...]

PR #9924, opened 2026-01-22
docs(hep): Guest Cluster Volume Remote Backup Restore [skip ci]

WebberHuang1118:
#### Problem:
Please refer to this HEP

#### Solution:
Please refer to this HEP

#### Related Issue(s):
#2661

### DIFF
--- a/enhancements/20260123-harvester-csi-remote-backup.md
+++ b/enhancements/20260123-harvester-csi-remote-backup.md
@@ -1,0 +1,657 @@
+# Harvester CSI Remote Volume Backup and Restore
+
+## Summary
+
+This design document outlines the implementation of remote volume backup and restore functionality in the Harvester CSI driver. The feature enables users to create remote backups of persistent volumes and restore them, providing disaster recovery and cross-cluster data migration capabilities for workloads running in guest Kubernetes clusters.
+
+### Related Issues
+
+- [Issue #2661](https://github.com/harvester/harvester/issues/2661)
+
+## Motivation
+
+### Goals
+
+- **Enable Remote Backup Operations**: Implement CSI snapshot capabilities using remote backup targets for persistent volumes
+- **Support Volume Restoration from Remote Backups**: Allow creation of new volumes from existing remote backups stored outside the cluster
+- **Provide Disaster Recovery**: Enable backup and restore across different guest clusters for disaster recovery scenarios as long as the guest cluster has the same Harvester as provider
+- **Maintain Compatibility**: Work alongside existing local snapshot functionality without breaking changes
+
+## Proposal
+
+### User Stories
+
+**Story 1: As a cluster administrator**
+I want to create remote backups of my persistent volumes to remote storage so that I can recover data even if the entire cluster is lost.
+
+**Story 2: As a DevOps engineer**
+I want to restore volumes from remote backups in a different guest cluster so that I can migrate workloads across environments or recover from disaster scenarios.
+
+**Story 3: As a backup administrator**
+I want to list all remote backups so that I can manage retention policies and select appropriate backups for restoration.
+
+**Story 4: As a system operator**
+I want backup operations to be transparent to applications using standard Kubernetes VolumeSnapshot APIs with appropriate parameters to distinguish between local and remote backups.
+
+## Design
+
+### Architecture Overview
+
+The remote backup and restore implementation follows a distributed architecture with clear separation of concerns:
+
+- **Host Cluster (Harvester)**: Manages backup/restore controllers and executes storage operations
+- **Guest Cluster (CSI Driver)**: Provides CSI interface and orchestrates backup/restore workflows
+
+The solution leverages custom Harvester resources (`VolumeRemoteBackup` and `VolumeRemoteRestore`) that utilize Longhorn's backup functionality to remote targets. An interface-based design allows flexibility to support third-party storage backends in the future if needed.
+
+#### Why VolumeRemoteBackup and VolumeRemoteRestore Abstractions?
+
+A critical design consideration is how namespace boundaries are handled during restore operations between guest and host clusters. The `VolumeRemoteBackup` and `VolumeRemoteRestore` resources provide a **namespace-agnostic abstraction** that solves this complexity.
+
+**The Namespace Mapping Challenge:**
+
+When restoring a volume in a guest cluster (e.g., from `gvol1` to `gvol2`), the underlying host Harvester cluster has its own namespace topology (e.g., `hvol1` to `hvol2`). The standard Kubernetes VolumeSnapshot mechanism requires different workflows depending on whether the source and target are in the same namespace:
+
+- **Same namespace**: `PVC` → `VolumeSnapshot` → `New PVC` (simple path)
+- **Cross-namespace**: `PVC` → `VolumeSnapshot` → `VolumeSnapshotContent` → `New VolumeSnapshot` → `New PVC` (requires cross-namespace reference handling)
+
+**Guest-to-Host Namespace Mapping Scenarios:**
+
+The complexity arises from how guest cluster namespaces map to host cluster namespaces during restore operations:
+
+| Scenario | Guest Cluster Topology | Host Cluster Topology | Workflow Complexity |
+|----------|------------------------|----------------------|---------------------|
+| **Case 1**: Same namespace, same cluster | `gvol1` and `gvol2` both in `ns-a` | `hvol1` and `hvol2` both in `hns-a` | Simple: Same-namespace restore on both sides |
+| **Case 2**: Different namespaces, same cluster | `gvol1` in `ns-a`, `gvol2` in `ns-b` | `hvol1` and `hvol2` both in `hns-a` | Complex: Guest sees cross-namespace restore, host sees same-namespace restore |
+| **Case 3**: Different clusters | `gvol1` in `ns-a`, `gvol2` in `ns-b` | `hvol1` and `hvol2` can be in same or different namespaces | Complex: Guest sees cross-namespace restore, host can see same-namespace restore or cross-namespace restore |
+
+**Design Solution:**
+
+The `VolumeRemoteBackup` and `VolumeRemoteRestore` abstractions eliminate this complexity by:
+
+1. **Decoupling from VolumeSnapshot constraints**: Backup metadata (handle, source PVC spec) is stored in VolumeRemoteBackup.status, removing the need to manage VolumeSnapshot/VolumeSnapshotContent across namespace boundaries
+
+2. **Unified workflow**: VolumeRemoteRestore references any VolumeRemoteBackup via `spec.from: "namespace/name"`, enabling **cross-namespace access** where the restore controller can access VolumeRemoteBackup resources in any namespace
+
+3. **Simplified CSI driver**: The driver doesn't need namespace-specific logic; it simply creates VolumeRemoteBackup/VolumeRemoteRestore resources with appropriate cross-namespace references
+
+**Example - Cross-namespace restore on Harvester:**
+
+```
+Guest Cluster:
+  namespace-a/gvol1 (PVC) --[CSI CreateSnapshot]--> VolumeRemoteBackup in Harvester (hvol-ns1)
+  namespace-b/gvol2 (PVC) <--[CSI CreateVolume]---- VolumeRemoteRestore in Harvester (hvol-ns2)
+
+Host Harvester:
+  hvol-ns1/hvol1 (PVC) → hvol-ns1/VolumeRemoteBackup (stores backup handle)
+  hvol-ns2/VolumeRemoteRestore (spec.from: "hvol-ns1/volumeRemoteBackup-name") → hvol-ns2/hvol2 (PVC)
+```
+
+In this scenario, even though guest PVCs are in different namespaces, the host Harvester cluster also maps them to different namespaces (`hvol-ns1` and `hvol-ns2`). The VolumeRemoteRestore in `hvol-ns2` can reference the VolumeRemoteBackup in `hvol-ns1` using the cross-namespace reference format `"hvol-ns1/volumeRemoteBackup-name"`. The VolumeRemoteRestore controller handles this cross-namespace access transparently, retrieving the backup handle from the remote storage target regardless of namespace boundaries.
+
+This abstraction layer ensures consistent behavior across all namespace topology scenarios, whether same-namespace or cross-namespace on either the guest or host side.
+
+### Component Architecture
+
+#### 1. API Resources (Host Cluster)
+
+Two new Custom Resource Definitions (CRDs) are introduced in the Harvester host cluster:
+
+- **`VolumeRemoteBackup`**: Represents a remote backup operation and stores backup metadata
+- **`VolumeRemoteRestore`**: Represents a restore operation from an existing backup
+
+#### 2. Controllers (Host Cluster)
+
+Controllers run in the Harvester host cluster and handle the actual backup/restore operations:
+
+- **VolumeRemoteBackup Controller**: Orchestrates backup operations to remote storage
+- **VolumeRemoteRestore Controller**: Orchestrates restore operations from remote storage
+
+#### 3. CSI Driver (Guest Cluster)
+
+The CSI driver in the guest cluster acts as a client that:
+
+- Creates and monitors `VolumeRemoteBackup` and `VolumeRemoteRestore` custom resources
+- Translates CSI operations to backup/restore workflows
+- Communicates with the host cluster via Kubernetes API
+
+### Snapshot Type Routing (Guest Cluster)
+
+The implementation extends the existing CSI snapshot interface to support different snapshot types:
+
+- **Default (empty `type`)**: Local snapshot using standard `VolumeSnapshot` resources
+- **`type=backup`**: Remote backup using `VolumeRemoteBackup` resources
+
+### Snapshot ID Format (Guest Cluster)
+
+Different ID formats distinguish between snapshot types:
+
+- **Local snapshots**: Simple name format (e.g., `snapshot-name`)
+- **Remote backups**: Namespaced format (e.g., `backup:namespace:name`)
+
+This format enables:
+- Unambiguous identification of backup vs. local snapshots
+- Cross-namespace backup operations
+- Simplified parsing and routing logic
+
+---
+
+## Implementation Details
+
+### Part 1: Host Cluster Implementation (Harvester)
+
+This section describes the changes required in the main Harvester repository.
+
+#### 1.1 New Custom Resource Definitions
+
+**VolumeRemoteBackup CRD**
+
+```go
+type VolumeRemoteBackup struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+    
+    Spec   VolumeRemoteBackupSpec   `json:"spec"`
+    Status VolumeRemoteBackupStatus `json:"status,omitempty"`
+}
+
+type VolumeRemoteBackupSpec struct {
+    // Type of backup, currently only "lh" (Longhorn) is supported
+    Type VolumeRemoteBackupType `json:"type"`
+    
+    // Source PVC name to backup
+    Source string `json:"source"`
+}
+
+type VolumeRemoteBackupStatus struct {
+    // Handle is the remote backup identifier in the backup target
+    Handle string `json:"handle"`
+    
+    // CSIProvider indicates which CSI driver created the backup
+    CSIProvider string `json:"csiProvider"`
+    
+    // SourceSpec contains the original PVC specification for restoration
+    SourceSpec corev1.PersistentVolumeClaimSpec `json:"sourceSpec,omitempty"`
+    
+    // Success indicates if the backup completed successfully
+    Success bool `json:"success,omitempty"`
+    
+    // Conditions provide detailed status information
+    Conditions []Condition `json:"conditions,omitempty"`
+}
+```
+
+**Resource Immutability:**
+
+The `VolumeRemoteBackup` resource is **immutable** once created. The `spec` fields cannot be modified after creation to ensure backup integrity and prevent inconsistencies between the backup metadata and the actual backup data stored in the remote target. This immutability is enforced by a Harvester webhook that validates and rejects any update attempts to the `spec` fields. Only the `status` fields can be updated by the controller during the backup lifecycle.
+
+**VolumeRemoteRestore CRD**
+
+```go
+type VolumeRemoteRestore struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+    
+    Spec   VolumeRemoteRestoreSpec   `json:"spec"`
+    Status VolumeRemoteRestoreStatus `json:"status,omitempty"`
+}
+
+type VolumeRemoteRestoreSpec struct {
+    // Type of restore, currently only "lh" (Longhorn) is supported
+    Type VolumeRemoteRestoreType `json:"type"`
+    
+    // From specifies the source VolumeRemoteBackup in format "namespace/name"
+    From string `json:"from"`
+}
+
+type VolumeRemoteRestoreStatus struct {
+    // Success indicates if the restore completed successfully
+    Success bool `json:"success,omitempty"`
+    
+    // Conditions provide detailed status information
+    Conditions []Condition `json:"conditions,omitempty"`
+}
+```
+
+#### 1.2 VolumeRemoteBackup Controller
+
+The VolumeRemoteBackup controller runs in the Harvester host cluster and manages the backup lifecycle.
+
+**Controller Responsibilities:**
+- Watch for new `VolumeRemoteBackup` resources created by CSI drivers
+- Validate backup specifications and source PVC existence
+- Orchestrate backup operations through the storage backend interface
+- Update `VolumeRemoteBackup` status with progress and completion information
+- Handle backup lifecycle and cleanup operations
+- Enforce immutability of `VolumeRemoteBackup` resources by rejecting any updates to the `spec` fields
+
+**Backup Workflow:**
+
+1. **Detection**: Controller watches for newly created `VolumeRemoteBackup` resources
+2. **Validation**: Validates the backup specification and verifies source PVC exists
+3. **Snapshot Creation**: Creates a volume snapshot in the storage backend (e.g., Longhorn)
+4. **Remote Upload**: Initiates upload of snapshot data to the configured remote target (S3, NFS, etc.)
+5. **Metadata Capture**: Stores PVC specification in `status.sourceSpec` for future restoration
+6. **Status Update**: Updates `status.handle` with the remote backup identifier/URL
+7. **Completion**: Sets `status.success=true` and updates conditions
+
+**Key Operations:**
+- Creates volume snapshots in the storage backend
+- Uploads snapshot data to configured remote targets (e.g., S3, NFS)
+- Captures PVC metadata for future restoration
+- Manages backup lifecycle and cleanup
+- Provides detailed progress through status conditions
+
+```mermaid
+sequenceDiagram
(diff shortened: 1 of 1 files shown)


enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,569 @@
...
+#### Why PVCBackup and PVCRestore Abstractions?
+
+A critical design consideration is how namespace boundaries are handled during restore operations between guest and host clusters. The `PVCBackup` and `PVCRestore` resources provide a **namespace-agnostic abstraction** that solves this complexity.
+
+**The Namespace Mapping Challenge:**
+
+When restoring a volume in a guest cluster (e.g., from `gvol1` to `gvol2`), the underlying host Harvester cluster has its own namespace topology (e.g., `hvol1` to `hvol2`). The standard Kubernetes VolumeSnapshot mechanism requires different workflows depending on whether the source and target are in the same namespace:
+
+- **Same namespace**: `PVC` → `VolumeSnapshot` → `New PVC` (simple path)
+- **Cross-namespace**: `PVC` → `VolumeSnapshot` → `VolumeSnapshotContent` → `New VolumeSnapshot` → `New PVC` (requires cross-namespace reference handling)
```
ihcsim, line 56: Upstream has a [cross-namespaced snapshot API](https://kubernetes-csi.github.io/docs/cross-namespace-data-sources.html), but it's still in alpha and has some odd dependencies. We should keep an eye on it.

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,569 @@
...
+## Design
+
+### Architecture Overview
+
+The remote backup and restore implementation follows a distributed architecture with clear separation of concerns:
+
+- **Host Cluster (Harvester)**: Manages backup/restore controllers and executes storage operations
+- **Guest Cluster (CSI Driver)**: Provides CSI interface and orchestrates backup/restore workflows
+
+The solution leverages custom Harvester resources (`PVCBackup` and `PVCRestore`) that utilize Longhorn's backup functionality to remote targets. An interface-based design allows flexibility to support third-party storage backends in the future if needed.
```
ihcsim: My loosely held opinion: can we call the CRD something like `VolumeBackup` and `VolumeRestore`? To me, `PVCBackup` and `PVCRestore` imply the backup/restore of the PVC manifest, not the volume that the claim references. In K8s concept, "claim" is generally a namespace-scoped abstraction that references a cluster-scoped lower level resources, so that a namespace-scoped workload can consume it.

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,569 @@
...
+- Creates and monitors `PVCBackup` and `PVCRestore` custom resources
+- Translates CSI operations to backup/restore workflows
+- Communicates with the host cluster via Kubernetes API
+
+### Snapshot Type Routing
+
+The implementation extends the existing CSI snapshot interface to support different snapshot types:
+
+- **Default (empty `type`)**: Local snapshot using standard `VolumeSnapshot` resources
+- **`type=backup`**: Remote backup using `PVCBackup` resources
```
ihcsim: To confirm, the idea is to define `type` as part of the volume snapshot class `parameters` property, right? It might potentially create a conflict with other 3rd party CSI drivers if they also have `parameters.type`. One alternative is to use label on the `VolumeSnapshot` directly.

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,569 @@
...
+
+### Snapshot ID Format
+
+Different ID formats distinguish between snapshot types:
+
+- **Local snapshots**: Simple name format (e.g., `snapshot-name`)
+- **Remote backups**: Namespaced format (e.g., `backup:namespace:name`)
```
ihcsim, line 130: How about just use a label on the snapshot? If anything, it will make discovering and listing easier. Something like `csi.harvesterhci.io/snapshot-mode: [backup|snapshot]`.

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,569 @@
...
+
+type PVCBackupStatus struct {
+    // Handle is the remote backup identifier in the backup target
+    Handle string `json:"handle"`
+    
+    // CSIProvider indicates which CSI driver created the backup
+    CSIProvider string `json:"csiProvider"`
+    
+    // SourceSpec contains the original PVC specification for restoration
+    SourceSpec corev1.PersistentVolumeClaimSpec `json:"sourceSpec,omitempty"`
```
ihcsim, line 174: Is there anything in the annotations or labels list that we will like to persist?

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,569 @@
...
+- Watch for new `PVCRestore` resources created by CSI drivers
+- Validate restore specifications and source backup existence
+- Orchestrate restore operations through the storage backend interface
+- Create new PVCs bound to restored volumes
+- Update `PVCRestore` status with progress and completion information
+
+**Restore Workflow:**
+
+1. **Detection**: Controller watches for newly created `PVCRestore` resources
+2. **Backup Resolution**: Locates the source `PVCBackup` resource specified in `spec.from`
```
ihcsim: Per previous discussion, can you add some info on how to ensure the `PVCBackup` resource is protected from deletion and mutation? I think we should also include the original `PVCBackup` in the backup artifact and restore only references the `PVCBackup` that is included with the backup artifact, not the in-cluster one.

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,569 @@
...
+
+**Backup Workflow:**
+
+1. **Detection**: Controller watches for newly created `PVCBackup` resources
+2. **Validation**: Validates the backup specification and verifies source PVC exists
+3. **Snapshot Creation**: Creates a volume snapshot in the storage backend (e.g., Longhorn)
+4. **Remote Upload**: Initiates upload of snapshot data to the configured remote target (S3, NFS, etc.)
+5. **Metadata Capture**: Stores PVC specification in `status.sourceSpec` for future restoration
+6. **Status Update**: Updates `status.handle` with the remote backup identifier/URL
+7. **Completion**: Sets `status.success=true` and updates conditions
```
ihcsim, line 236: Once completed, I think `PVCBackup` should be immutable.

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,569 @@
...
+- CSI driver communicates with the host cluster via Kubernetes API
+- Creates `PVCBackup` and `PVCRestore` resources in the host cluster
+- Polls resource status until operations complete
+- Translates CSI operations to backup/restore workflows
+
+#### 2.2 Core CSI Operations
+
+The CSI driver extends existing operations to support remote backup and restore workflows.
+
+**CreateSnapshot - Extended for Backup Mode**
```
ihcsim, line 376: So the way a guest cluster user interacts with this is by directly creating the `VolumeSnapshot` resource, right?

Conversely, the way to perform a restore is to manually create the PVC with its data source pointing to the volume snapshot resource, if it is in the same namespace. If it's a restore to a different namespace, guest cluster user is responsible for creating the additional volume snapshot based on the volume snapshot content. Then reference the volume snapshot from the PVC, correct?

Anything stopping us from bundling the Harvester CSI driver with backup and restore CRDs to automate this manual workflow? It will increase to scope, but maybe something that can do post 1.8?

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Thanks for the information—we should follow it. However, it won’t cover all of our use cases, since our scope also includes **inter-cluster** backup/restore (i.e., backing up from one guest cluster and restoring to another cluster).

However, this can help other scenarios in the future.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Actually, I had the same idea as you.

However, `VolumeBackup` ([VolumeBackup API](https://github.com/harvester/harvester/blob/5c832461f4e6d3c92e9f82a90a4ca0278b27a5a7/pkg/apis/harvesterhci.io/v1beta1/backup.go#L123-L155)) and `VolumeRestore` ([VolumeRestore API](https://github.com/harvester/harvester/blob/5c832461f4e6d3c92e9f82a90a4ca0278b27a5a7/pkg/apis/harvesterhci.io/v1beta1/backup.go#L240-L255)) are already in use.

To keep a clear separation, I chose `PVCBackup` and `PVCRestore`. Do you have any suggestions?

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: For third-party storage, what matters is how we handle data backup and restore on the Harvester side. That’s why I designed a separate host-layer abstraction (i.e., separation and control via `PVCBackup`/`PVCRestore`), which is isolated from the guest cluster.

This parameter only affects the **guest** Harvester CSI driver and how it communicates with Harvester. 

For example, if we support backup/restore for guest cluster volumes that map to third-party storage on Harvester, it would only change how the `PVCBackup` and `PVCRestore` controllers perform storage operations on the host side. 

From the guest cluster perspective, it would still use the same `VolumeSnapshotClass` objects—one for local snapshots and one for remote backups—with identical parameters.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: For a CSI driver, in snapshot-related CSI calls, it only sees the CSI requests (e.g., `CreateSnapshot`, `DeleteSnapshot`, `ListSnapshots`, etc.). It doesn’t receive the `VolumeSnapshot` / `VolumeSnapshotContent` manifests themselves. That’s why the driver needs to rely on the **Snapshot ID** to persist any required information.

Longhorn CSI follows a similar approach as well (see the snapshot handling logic in [Longhorn CSI snapshot controller code](https://github.com/longhorn/longhorn-manager/blob/0469081aff47f8975315d2a35f74b9e0a4402ce7/csi/controller_server.go#L154-L202)).

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Since backup isn’t part of the CSI standard, and Longhorn doesn’t support volume restore and expansion at the same time, we need to know the source PVC size so we can validate it during volume restoration.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Thanks—that’s a good idea. IMO, `PVCBackup` should be immutable after creation. I’ll update it accordingly.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: As the remote volume backup/restore targets:

* intra–guest-cluster backup/restore, and
* inter–guest-cluster backup/restore (as long as the guest clusters use the same Harvester as the provider),

From the host perspective, regardless of the guest cluster topology, everything maps to `PVCBackup` and `PVCRestore` CRs on the same Harvester.

IMO, we appear not need to include `PVCBackup` in the backup artifact for deletion—we can enforce safeguards via the Harvester webhook. (For mutation, I’ll prevent it as discussed in [immutability discussion](https://github.com/harvester/harvester/pull/9924/changes/BASE..ffa2820a6d20a7541ea20440a847a1d554936fbc#r2776171743)).

There are two scenarios for the relationship between `PVCBackup` and `PVCRestore`:

* `PVCRestore` is created (from guest cluster volume restore) with a `PVCBackup` reference and completes:

  * After the restore completes, the target volume is decoupled from the source volume.
  * Deleting `PVCBackup` won’t affect the target volume or the completed `PVCRestore` CR.

* `PVCRestore` is created after the referenced `PVCBackup` has already been deleted:

  * The `PVCRestore` creation should be rejected by the Harvester webhook because the referenced `PVCBackup` no longer exists.
  * As a corner case, if the `PVCBackup` is deleted during an ongoing restore, the Harvester webhook should also reject that deletion.

I’ll update the HEP accordingly—thanks.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: > If it's a restore to a different namespace, guest cluster user is responsible for creating the additional volume snapshot based on the volume snapshot content. Then reference the volume snapshot from the PVC, correct?

To restore into a different namespace or a different guest cluster, the `VolumeSnapshotContent` should be created first, then bound to a `VolumeSnapshot`, and finally a new volume (PVC) should be created from that `VolumeSnapshot`.

 
> Anything stopping us from bundling the Harvester CSI driver with backup and restore CRDs to automate this manual workflow? It will increase to scope, but maybe something that can do post 1.8?

Yes—we can consider adding an automation mechanism in the future. For now, we probably shouldn’t make the workflow too complex in the initial version, while still keeping the architectural flexibility in case this requirement becomes necessary later.

Aside from the above context, one concern I have with automation is that it could create too many snapshot-related resources across all guest clusters. 

For example, if guest cluster **g1** creates 100 volume backups, and the same Harvester also serves other guest clusters (**g2–g6**), an automated snapshot recovery mechanism might cause **g2–g6** to each create a `VolumeSnapshot` and a `VolumeSnapshotContent` for all 100 backup volumes. That could mean up to **500 × 2 = 1000** snapshot resources that may never be used. 

In contrast, a manual/on-demand workflow would only create snapshot resources when they’re actually needed.

We can discuss this in more detail in later versions if it becomes necessary.

enhancements/20260123-harvester-csi-remote-backup.md
votdev: What about `RemoteVolumeBackup` and `RemoteVolumeRestore`?

WebberHuang1118: @votdev, thanks for your suggestion. However, even if the volume is backed up to a remote target, it can still be restored to the same cluster—similar to a local snapshot restore. I’m wondering whether this could be confusing.

votdev: Perhaps the first suggestion is a little confusing because "Remote" is at the beginning and it is not clear to which context it belongs. Therefore, here is another suggestion: `VolumeRemoteBackup` and `VolumeRemoteRestore`. Here it may become clearer that this is a backup to a remote target.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Updated accordingly—thanks.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Updated accordingly—thanks.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: That’s a good idea—thanks, @votdev.

@ihcsim, if you also agree with the naming suggestion from @votdev, I can update everything accordingly, including the HEP, [Harvester PR](https://github.com/harvester/harvester/pull/9925), and [Harvester CSI PR](https://github.com/harvester/harvester-csi-driver/pull/68).

WebberHuang1118:
Hi @ihcsim @votdev,

Thanks for your feedback. I’ve replied and updated the HEP accordingly—please take a look, thanks.

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: sounds good

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: Yeah, I was thinking about the guest clusters which we have less control over. Oh, this is specifically for the `harvester-snapshot` volume snapshot class, not other 3rd party CSI driver, right? Then we are good.

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: > The PVCRestore creation should be rejected by the Harvester webhook because the referenced PVCBackup no longer exists.

I feel like this is one of core topics to discuss. Technically, we need to be able to perform a restore on the guest cluster, as long as the remote backup artifacts still exist, regardless of what happens to Harvester. If the `PVCBackup` object is gone with no ways to reconstruct it, then the remote backup artifacts become useless. E.g., in Velero, the remote backup is always restorable regardless of what happens to Velero itself and the rest of the K8s cluster.

Is it difficult to include the `PVCBackup` manifest in the backup artifact?

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: Sorry, I don't mean automating the guest snapshot creation or recovery. What I mean is we introduce `VolumeBackup` and `VolumeRestore` CRDs (completely different from the Harvester's ones) and make these CRDs part of the Harvester CSI driver deployment. Then user can work directly with the guest-side `VolumeBackup` and `VolumeRestore`  CRDs and let the CSI driver handle the creation (and garbage collection) of the `VolumeSnapshot` and `VolumeSnapshotContent` resources.

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: What I meant is `SourceSpec` only captures the spec of the PVC. Do we need to also include the labels and annotations of the PVC?

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Yes, you're correct.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: So far, I don’t see any labels or annotations that we need to preserve during restore. The current data structure is flexible enough to extend later if needed.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: It’s doable, but it increases the complexity. Harvester would need a mechanism to periodically scan the artifact and reassemble the `VolumeRemoteBackup` (originally `PVCBackup`) and its underlying resources. That logic is also backup/restore engine–specific. For example, if the engine is Longhorn, the underlying resources are the CSI `VolumeSnapshot` and `VolumeSnapshotContent` objects on Harvester, and this would vary across different engines.

Considering Harvester may not always stay aligned with the artifact. For example, if the S3 target is disconnected and the `VolumeRemoteBackup` CR is deleted in Harvester, the user could still reassemble the required resources once the artifact is reconnected, as long as the backup data still exists. However, an automatic discovery mechanism feels too complex at the current stage.

I’d prefer documenting the manual steps first, and then we can plan an automated mechanism in the next phase.

WDYT?

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Do you mean introducing new guest-cluster CRDs like `VolumeBackup` and `VolumeRestore`? We don’t have that kind of deployment so far, and it would imply the Harvester CSI driver needs controllers to manage those new CRDs.

From my perspective, new CRDs may not be necessary. With the current design, guest cluster users can continue using the standard CSI Snapshot APIs, while the Harvester CSI driver creates and tracks `VolumeRemoteBackup` and `VolumeRemoteRestore` on Harvester, and Harvester handles the storage operations in its controllers.

If we introduce new CRDs in the guest cluster, users may no longer be able to rely on the standard CSI snapshot APIs. Instead, they’d need to interact with the new guest-cluster CRDs, which might not be expected (or supported) by some applications.

> Anything stopping us from bundling the Harvester CSI driver with backup and restore CRDs to automate this manual workflow? It will increase to scope, but maybe something that can do post 1.8?

I understand this could make the workflow more straightforward. However, it adds a new layer in the guest cluster. Given that Harvester already has an abstraction layer with `VolumeRemoteBackup` and `VolumeRemoteRestore`, this would make the architecture more complex. 

If it turns out to be necessary, I think we can plan it after 1.8.

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: To confirm, I wasn't thinking of automatically restoring deleted in-cluster `VolumeRemoteBackup` objects.

I would like to know if it is possible to take a backup of the `VolumeRemoteBackup` YAML spec and push it to the remote backup target? I thought we are already doing it for `VirtualMachineBackup` objects. The key is that during restore, Harvester re-constructs the `VolumeRemoteBackup` object from the remote YAML spec, instead of depending on the in-cluster `VolumeRemoteBackup` object. That way, restore still works even if the in-cluster `VolumeRemoteBackup` objects are deleted.

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: Yeah, I recognize that we don't have CRDs associated with our CSI driver at the moment. Hence, I thought it might be something for future release considerations.

From a guest cluster backup/restore UX perspective, having a separate `VolumeBackup` and `VolumeRestore` CRDs mean users can interact with those resources specifically for backup/restore operations, while preserving the CSI snapshot CRDs for snapshot operations. The CRDs can encapsulate backup/restore specific attributes. Users don't need to keep a separate `VolumeSnapshotClass` for backup/restore, and (more importantly, imo) also save users having to create and garbage-collect `VolumeSnapshot` and `VolumeSnapshotContent` resources for cross-namespace restores, because the CSI driver controller should be able to manage all that in the background.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: So you mean when having restoration with `VolumeRemoteBackup`, even the `VolumeRemoteBackup` CR is gone, if the backup data still exists in artifact, the `VolumeRemoteRestore` controller still can get the backup content from artifact?

If so, it will make the `VolumeRemoteRestore` webhook can not have safegurad protect based on `VolumeRemoteBackup`'s existence

If we put the `VolumeRemoteBackup` manifest to artifact, I'd prefer to scan and pull back the manifest from artifact, and `VolumeRemoteRestore` creation webhook based on `VolumeRemoteBackup`'s existence, like what we currenly works for `VMBackup` and `VMRestore`

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Yes, I agree with your point. 

How about this approach: in v1.8, users can use the standard CSI Snapshot APIs to perform volume backup and restore in the guest cluster. Then in a future release (e.g., v1.9), in addition to the standard CSI APIs, we could offer an optional custom CR-based workflow for guest cluster volume backup and restore.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Updated accordingly, thanks.

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: Sounds good. Re: the guest cluster `VolumeSnapshot` resources, if they somehow disappear, then there is no way to perform a restore, right?

enhancements/20260123-harvester-csi-remote-backup.md
ihcsim: So it looks like although we push the `.cfg` file of the `VMBackup` resource to the remote backup target, we don't use them during restore. I don't think that's correct. For this HEP, as a start, we can follow the same approach, but the in-cluster `VolumeRemoteBackup` resources must be immutable.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: No—as long as the `VolumeRemoteBackup` CR exists in Harvester, the guest cluster can have data restoration through the flow: `VolumeSnapshotContent` → `VolumeSnapshot` → restore PVC.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Yes—your understanding is correct. The VMRestore controller doesn’t refer to the artifact; it relies on the VMBackup resource.

I agree with your point that we should have some kind of protection mechanism for the artifact, especially since we plan to support [issue-1850](https://github.com/harvester/harvester/issues/1850). We currently have a draft PR ([PR-9833](https://github.com/harvester/harvester/pull/9833)), but there’s still no synchronization across clusters via the artifact.

Back to this topic:

> For this HEP, as a start, we can follow the same approach, but the in-cluster VolumeRemoteBackup resources must be immutable.

Let’s stick with the current approach for now. I can create an enhancement ticket to track this after the HEP decision is made.

WebberHuang1118:
Hi all,

I’ve updated the HEP with a test plan and addressed all the feedback. Please take a look—thanks.

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,657 @@
...
+
+#### 3. CSI Driver (Guest Cluster)
+
+The CSI driver in the guest cluster acts as a client that:
+
+- Creates and monitors `VolumeRemoteBackup` and `VolumeRemoteRestore` custom resources
+- Translates CSI operations to backup/restore workflows
+- Communicates with the host cluster via Kubernetes API
+
+### Snapshot Type Routing
```
Vicente-Cheng: Is this for guest cluster?
If yes, could we add `(Guest cluster)` like above?

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,657 @@
...
+
+The implementation extends the existing CSI snapshot interface to support different snapshot types:
+
+- **Default (empty `type`)**: Local snapshot using standard `VolumeSnapshot` resources
+- **`type=backup`**: Remote backup using `VolumeRemoteBackup` resources
+
+### Snapshot ID Format
```
Vicente-Cheng: ditto

enhancements/20260123-harvester-csi-remote-backup.md
```diff
@@ -0,0 +1,657 @@
...
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+    
+    Spec   VolumeRemoteBackupSpec   `json:"spec"`
+    Status VolumeRemoteBackupStatus `json:"status,omitempty"`
+}
+
+type VolumeRemoteBackupSpec struct {
+    // Type of backup, currently only "lh" (Longhorn) is supported
+    Type VolumeRemoteBackupType `json:"type,omitempty"`
```
Vicente-Cheng: If this is mandatory, we should remove omitempty

Vicente-Cheng (dismissed):
justa some nits, others lgtm

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Updated per your suggestion—thanks.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Updated per your suggestion—thanks.

enhancements/20260123-harvester-csi-remote-backup.md
WebberHuang1118: Updated per your suggestion—thanks.

brandboat (approved):

PR #9925, opened 2026-01-22
feat: introduce CR for guest volume backup/restore

WebberHuang1118:
#### Problem:
Introduce CR volumeremotebackup and volumeremoterestore for guest volume backup/restore

#### Solution:
Introduce CR volumeremotecbackup and volumeremoterestore for guest volume backup/restore

#### Related Issue(s):
#2661 

#### Test plan:
* Update **Harvester** and the **Harvester webhook** deployments with this PR.
* Create an RWX block PVC in a namespace (e.g., `backup/vol1`) from a VM image.
* Attach the volume to a VM and write some data.
* Create a `VolumeRemoteBackup` CR that points to the PVC above, for example:
  ```yaml
  kind: VolumeRemoteBackup
  metadata:
    namespace: backup
    name: pvcb-vol1
  spec:
    type: lh
    source: vol1
  ```
* Wait until `VolumeRemoteBackup.status.ready` becomes `true`.
* In another namespace, create a `VolumeRemoteRestore` resource that points to the `VolumeRemoteBackup` CR you just created, for example:
  ```yaml
  kind: VolumeRemoteRestore
  metadata:
    namespace: default
    name: pvcr-vol1
  spec:
    type: lh
    from: backup/pvcb-vol1
  ```
* Wait until `VolumeRemoteRestore.status.ready` becomes `true`. A PVC with the same name should be restored.
* Attach the restored PVC to a VM and verify data integrity.


WebberHuang1118:
Hi all,

I’ve updated this PR based on the HEP feedback ([comment](https://github.com/harvester/harvester/pull/9924#discussion_r2775979769)). 

This PR is also required before we can proceed further with the Harvester CSI PR [[WIP] support remote volume backup and restore](https://github.com/harvester/harvester-csi-driver/pull/68).

Could you please take a look? Thanks.

deploy/charts/harvester-crd/templates/harvesterhci.io_pvcbackups.yaml
```diff
@@ -0,0 +1,311 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
ihcsim: i think we can remove this and `pvcrestores` since they are replaced by `volumeremotebackups` and `volumeremoterestores`.

pkg/controller/master/volumeremotebackup/register.go
```diff
@@ -0,0 +1,303 @@
...
+	"github.com/harvester/harvester/pkg/volumeremotebackup/common"
+	"github.com/harvester/harvester/pkg/volumeremotebackup/driver"
+	"github.com/harvester/harvester/pkg/volumeremotebackup/longhorn"
+)
+
+const (
+	pvcbackupControllerName      = "pvc-backup-controller"
+	volumeSnapshotControllerName = "volume-snapshot-controller"
+	pvcrestoreControllerName     = "pvc-restore-controller"
+	pvControllerName             = "pvc-controller"
```
ihcsim:
```suggestion
	pvcControllerName             = "pvc-controller"
```
or just make it consistent with the CRD with the `volume-remote-` prefix?

pkg/controller/master/volumeremotebackup/register.go
```diff
@@ -0,0 +1,303 @@
...
+func (r *remoteBackupResolver) getResourceType() string {
+	return remoteBackupKindName
+}
+
+func (r *remoteBackupResolver) enqueue(namespace, name string) {
+	r.handler.vrbController.Enqueue(namespace, name)
+}
+
+// pvcRestoreCacheAdapter adapts PVCRestore cache to resourceCache interface
+type pvcRestoreCacheAdapter struct {
```
ihcsim:
```suggestion
type remoteRestoreCacheAdapter struct {
```

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
```diff
@@ -0,0 +1,307 @@
...
+		rbh.vrbController.EnqueueAfter(rbh.bo.GetNamespace(vrb), rbh.bo.GetName(vrb), 5*time.Second)
+		return vrb, nil
+	}
+
+	if apierrors.IsNotFound(err) {
+		logrus.Infof("Resource not found for PVCBackup %s/%s, creating it", rbh.bo.GetNamespace(vrb), rbh.bo.GetName(vrb))
+		return rbh.handleResourceNotFound(vrb, op)
+	}
+
+	vrb, SetErr := rbh.bo.SetError(vrb, err)
```
ihcsim:
```suggestion
	vrb, setErr := rbh.bo.SetError(vrb, err)
```
i assume it's meant to be lower case?

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
```diff
@@ -0,0 +1,307 @@
...
+	if common.IsRetryLater(err) {
+		logrus.Infof("PVCBackup %s/%s not ready, retrying in 5 seconds", rbh.bo.GetNamespace(vrb), rbh.bo.GetName(vrb))
+		rbh.vrbController.EnqueueAfter(rbh.bo.GetNamespace(vrb), rbh.bo.GetName(vrb), 5*time.Second)
+		return vrb, nil
+	}
+
+	if apierrors.IsNotFound(err) {
+		logrus.Infof("Resource not found for PVCBackup %s/%s, creating it", rbh.bo.GetNamespace(vrb), rbh.bo.GetName(vrb))
+		return rbh.handleResourceNotFound(vrb, op)
+	}
```
ihcsim, line 63: I don't think we can assume that the `IsNotFound` error is always due to a missing `VolumeRemoteBackup` object. E.g., `lbo.getVolumeSnapshotForRemoteBackup()` can return an `IsNotFound` error too.

Same with `rrh.handleResourceNotFound()` where `lbr.getVSCForRestore()` may return an `IsNotFound` for volume snapshot content, instead of `VolumeRemoteRestore`.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
```diff
@@ -0,0 +1,307 @@
...
+
+	updatedPb, err = rbh.bo.UpdateCSIProvider(updatedPb)
+	if err != nil {
+		return vrb, err
+	}
+
+	updatedPb, err = rbh.bo.UpdateSourceSpec(updatedPb)
+	if err != nil {
+		return vrb, err
+	}
```
ihcsim: What is the reason behind calling one setter at a time, instead of just modify a clone of the backup object and then call update once? It's also unclear to me why some of the setters made a clone and then call update (like `SetSuccess()` and `UpdateCSIProvider()`), while others simply modify the in-memory object (like `SetReady()`).

pkg/volumeremotebackup/longhorn/restore.go
```diff
@@ -0,0 +1,572 @@
...
+
+	errorMsg := "VolumeSnapshotContent is in error state"
+	if vsc.Status.Error.Message != nil {
+		errorMsg = *vsc.Status.Error.Message
+	}
+	return fmt.Errorf("%s", errorMsg)
+}
+
+// checkVolumeSnapshotError checks if a VolumeSnapshot is in an error state.
+func (lbr *LHRestoreOperation) checkVolumeSnapshotError(vs *snapshotv1.VolumeSnapshot) error {
```
ihcsim: can we make the `lbo.checkVolumeSnapshotError()` a common function shared between LH backup and restore?

pkg/volumeremotebackup/common/operator.go
```diff
@@ -0,0 +1,476 @@
...
+	GetName      func(T) string
+	GetKind      func(T) string
+	GetUID       func(T) types.UID
+	DeepCopy     func(T) T
+	GetStatus    func(T) S
+	SetStatus    func(T, S) T
+}
+
+// ResourceOperator provides common operations for PVCBackup/PVCRestore
+type ResourceOperator[T ResourceFields, S StatusFields] struct {
```
ihcsim: I'd think something like the `ResourceOperator` is needed if we intend to have other kind of backup and restore APIs, besides `VolumeRemoteBackup` and `VolumeRemoteRestore`? These two should be the main and only backup/restore APIs, supported by different kinds of drivers with their own `BackupOperations` and `RestoreOperations` implementation.

At the moment, `ResourceOperator`, `BackupOperator` and `RestoreOperator` look a bit complicated for a collection of getters and setters.

I also think `ResourceOperator` unnecessary couples the implementation of the `BackupOperator` and `RestoreOperator`. To me, backup and restore operators should be kept separate in case their caches, accessors etc. divert. The `ResourceOperator` forces one operator to inherit whatever the other needs.

pkg/volumeremotebackup/longhorn/backup.go
```diff
@@ -0,0 +1,246 @@
...
+
+// BuildOwnerReference creates an owner reference for a PVC backup.
+func (lbo *LHBackupOperation) BuildOwnerReference(vrb *harvesterv1.VolumeRemoteBackup) metav1.OwnerReference {
+	return metav1.OwnerReference{
+		APIVersion: harvesterv1.SchemeGroupVersion.String(),
+		Kind:       lbo.bo.GetKind(vrb),
+		Name:       lbo.bo.GetName(vrb),
+		UID:        lbo.bo.GetUID(vrb),
+		Controller: ptr.To(true),
+	}
```
ihcsim, line 119: What's the purpose of making the `VolumeRemoteBackup` object the owner? The upstream snapshot controller and some CSI drivers tend to append their owner references to the snapshot object too, so things like cascading deletion (if that's what we are trying to achieve) won't necessary work. What's the trade-off with just using label for discovery purposes?

ihcsim (commented):
Thanks for the PR. First round of review with some questions.

deploy/charts/harvester-crd/templates/harvesterhci.io_pvcbackups.yaml
WebberHuang1118: Updated per your suggestion—thanks.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
WebberHuang1118: Updated per your suggestion—thanks.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
WebberHuang1118: Yes—the message is misleading. `IsNotFound` is returned when the `VolumeSnapshot` doesn’t exist.

I’ve updated the message to reflect the correct context. Thanks.

Also, for `IsNotFound` handling, `rbh.handleResourceNotFound()` will first check whether the required resource exists, and create it if needed.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
WebberHuang1118: > It's also unclear to me why some of the setters made a clone and then call update (like SetSuccess() and UpdateCSIProvider()), while others simply modify the in-memory object (like SetReady())

All the mentioned functions clone the object first, then call `Update` to commit the changes.

> What is the reason behind calling one setter at a time, instead of just modify a clone of the backup object and then call update once?

Thanks for pointing this out. I added a `Finalize()` method to `BackupOperator`. `Finalize()` applies all changes to the in-memory object first, then commits them to the API server with a single `Update()` call.

pkg/volumeremotebackup/longhorn/restore.go
WebberHuang1118: Updated per your suggestion—thanks.

pkg/volumeremotebackup/common/operator.go
WebberHuang1118: Initially, I designed `BackupOperator` and `RestoreOperator` separately. However, since the two resources have similar data structures, I found a lot of duplicated logic in their implementations, so I extracted the shared logic into `ResourceOperator`.

I know this makes the implementation a bit more complex and less straightforward, but it also keeps the code more compact and reduces duplication.

Do you have a strong opinion on this? I’m fine reverting back to the separate pattern if needed. However, given our schedule, I don’t think the current design is overly complicated, and switching back to the separate pattern wouldn’t be a minor change. 

If possible, I’d like to keep the current implementation. If this design becomes too complex in the future, we can refactor back to a separate pattern then—there shouldn’t be any compatibility concerns.

pkg/volumeremotebackup/longhorn/backup.go
WebberHuang1118: It’s not for discovery purposes—it ensures that deleting a `VolumeRemoteBackup` also deletes the corresponding `VolumeSnapshot`.

From the user’s perspective: if a volume’s remote backup is removed, the related CSI snapshot should be cleaned up as well.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
ihcsim: how about just create `VolumeSnapshot` in `lbo.readiness()` if `lbo.getVolumeSnapshotForRemoteBackup()` returns `IsNotFound`? Then `rbh.handleReadinessError()` only needs to handle `common.IsRetryLater(err)`. Currently, we are calling `lbo.readiness()` twice; once inside `lbo.Readiness()` and then again in `lbo.Create()`.

pkg/volumeremotebackup/longhorn/backup.go
ihcsim: If the volume snapshot ends up with more than one owner references, say from upstream snapshot controller or other CSI drivers, the cascading deletion won't work until all the owner references are deleted. I think it's safer to let our controller delete the `VolumeSnapshot`.

WebberHuang1118: This kind of `VolumeSnapshot` should have only one owner reference, i.e., the `VolumeRemoteBackup` CR, since `VolumeRemoteBackup` is the source of the `VolumeSnapshot`.

If the two resources are in the same namespace, I think it’s straightforward to use an owner reference to get cascading deletion automatically.

For example, we already have our own cascading-deletion logic for VM images, backing images, and StorageClasses, and we’ve hit some race conditions there. It’s doable, but it can also get more complicated.

WDYT?

ihcsim: Basically, we can't assume that our controller has sole ownership over an object from a 3rd party CRD. Unlike vmimage and bi which we own/control, `VolumeSnapshot` is a very commonly used upstream CRD and this really depends on the provisioner behind it. E.g., if we start using volume group snapshot, the upstream snapshot controller (and Longhorn's fork of it - `csi-snapshotter`) will [append additional owner references to it](https://github.com/longhorn/csi-snapshotter/blob/f21cb02763e7cd6a7fc84846f106b83119b5371d/pkg/common-controller/snapshot_controller.go#L452-L455).

Just keep this in mind if the cascading deletion assumption we made here today becomes invalid in the future.

pkg/volumeremotebackup/common/operator.go
ihcsim: I am ok with if other reviewers agree to it too. To me, it's a matter of whether the rest of the team can easily understand and help maintain the code. This code pattern is somewhat different from the other controllers. I don't know if we are necessary saving much because the duplication it reduces is taken up by the unfamiliar additional layers of abstraction.

To confirm, I think the design on the ~device~ driver and backup/restore operations are great; just not sure about the operator abstraction.

WebberHuang1118: Yes, I agree with your point, @ihcsim.

@Vicente-Cheng @brandboat TL;DR: @ihcsim feels the `ResourceOperator` abstraction for `BackupOperator` and `RestoreOperator` adds some complexity and follows a different pattern from our other controllers.

From my perspective, it does make the implementation a bit less straightforward, but it also keeps the code more compact and reduces duplication, and it has worked fine so far.

I’m OK to switch back to a fully separated design for `BackupOperator` and `RestoreOperator`, but given our schedule, and that wouldn’t be a minor change.

WDYT?

pkg/volumeremotebackup/longhorn/backup.go
WebberHuang1118: We already use owner references for `VolumeSnapshot` in many places in the codebase 😄

Do you have a strong opinion on this? Using an owner reference is simpler for managing related resources. 

Since it works with the latest Longhorn version, I’d like to keep it if possible—maybe we can add a source comment based on the information you shared.

WDYT?

WebberHuang1118:
Hi @Vicente-Cheng @brandboat,

Please check my discussion with @ihcsim ([PR-9925 comment](https://github.com/harvester/harvester/pull/9925#discussion_r2928882960)), thanks.

WebberHuang1118:
> Hi @Vicente-Cheng @brandboat,
> 
> Please check my discussion with @ihcsim ([PR-9925 comment](https://github.com/harvester/harvester/pull/9925#discussion_r2928882960)), thanks.

Hi @ihcsim @Vicente-Cheng @brandboat,

Re: [PR-9925 discussion](https://github.com/harvester/harvester/pull/9925#discussion_r2928882960)

I’ve updated the implementation to use a separate pattern. Please take a look—thanks.

pkg/volumeremotebackup/common/operator.go
WebberHuang1118: Please refer to [this comment](https://github.com/harvester/harvester/pull/9925#issuecomment-4066143829). Thanks.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
```diff
@@ -0,0 +1,287 @@
...
+
+	ready, err := op.Readiness(vrb)
+	if err != nil {
+		return rbh.handleReadinessError(vrb, op, err)
+	}
+
+	if ready {
+		return rbh.handleReadyBackup(vrb)
+	}
```
ihcsim, line 43: Do we need enqueue `vrb` if `!ready && err == nil`? Same with the restore side.

pkg/volumeremotebackup/longhorn/backup.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// Skip if handle is already set
+	if vrb.Status.Handle != "" {
+		return true, nil
+	}
+
+	// Check if VolumeSnapshotContent is bound
+	if vs.Status.BoundVolumeSnapshotContentName == nil {
+		return false, nil
```
ihcsim, line 137: I assume we'd want to do a retry here with something like `return false, common.ErrRetryLater`? Also, at line 140 and 148.

pkg/volumeremotebackup/longhorn/restore.go
```diff
@@ -0,0 +1,559 @@
...
+// readiness checks if all resources for the PVCRestore are ready.
+// Returns true if ready, false if not yet ready, or false with error if failed.
+func (lbr *LHRestoreOperation) readiness(vrr *harvesterv1.VolumeRemoteRestore) (bool, error) {
+	// Step 1: Check VolumeSnapshotContent
+	vsc, err := lbr.checkVSCReadiness(vrr)
+	if err != nil {
+		return false, err
+	}
+	if vsc == nil {
+		return false, nil
```
ihcsim: should this return `common.ErrRetryLater`? same for line 480 and 489.

pkg/volumeremotebackup/longhorn/restore.go
```diff
@@ -0,0 +1,559 @@
...
+func (lbr *LHRestoreOperation) createVolumeSnapshot(
+	vrr *harvesterv1.VolumeRemoteRestore,
+	vsc *snapshotv1.VolumeSnapshotContent,
+	vsClass *snapshotv1.VolumeSnapshotClass,
+) (*snapshotv1.VolumeSnapshot, error) {
+	vsName := lbr.ro.GetName(vrr)
+	vsNamespace := lbr.ro.GetNamespace(vrr)
+	vscName := vsc.Name
+
+	vs := &snapshotv1.VolumeSnapshot{
```
ihcsim, line 283: does `vs` need an ownerref to `vrr`?

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
```diff
@@ -0,0 +1,287 @@
...
+
+	updatedPr, err := rrh.ro.SetReady(vrr)
+	if err != nil {
+		return vrr, err
+	}
+
+	updatedPr, err = rrh.ro.SetSuccess(updatedPr, true)
+	if err != nil {
+		return vrr, err
+	}
```
ihcsim: With the recent change, I think these are no longer persisted. Looks to me it's missing the`Finalize()` for restore.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
```diff
@@ -0,0 +1,287 @@
...
+
+func (vsh *volumeSnapshotHandler) tryResolveAndEnqueueByRef(vsNamespace, vsName string, controllerRef *metav1.OwnerReference) {
+	tryResolveAndEnqueueByRef(vsh.resolvers, "VolumeSnapshot", vsNamespace, vsName, controllerRef)
+}
+
+func (vsh *volumeSnapshotHandler) tryResolveAndEnqueueByAnnotation(vsNamespace, vsName string, annotations map[string]string) {
+	tryResolveAndEnqueueByAnnotation(vsh.resolvers, "VolumeSnapshot", vsNamespace, vsName, annotations)
+}
+
+func (vsh *volumeSnapshotHandler) onChanged(_ string, vs *snapshotv1.VolumeSnapshot) (*snapshotv1.VolumeSnapshot, error) {
```
ihcsim, line 247: If the only purpose of the `volumeSnapshotHandler` and `pvcHandler` handlers is to trigger a reconciliation of the owner `volumeRemoteBackup`, will the `relatedresource.Watch()` function do what you want? It can be used to resolve and enqueue the owner `volumeRemoteBackup` following changes in `volumeSnapshot` and `persistentVolumeClaim`.

See example at https://github.com/harvester/harvester/blob/a5211a4008ad26d61049da9f9da279ce75b6e458/pkg/controller/master/vmimagedownloader/register.go#L37

ihcsim (commented):
To confirm my understanding on Harvester side, we are not writing to nor reading from the remote backup target, right? Everything is managed as volume snapshots. I thought in the HEP we mentioned using the remote backup target.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
WebberHuang1118: Based on the Longhorn implementation in `pkg/volumeremotebackup/longhorn/backup.go`, the reconcile is already triggered by `VolumeSnapshot` changes. 

However, adding an explicit enqueue at this layer would improve readability and make the logic more general.

Thanks for pointing this out.

pkg/volumeremotebackup/longhorn/backup.go
WebberHuang1118: Since CSI volume changes will trigger `VolumeRemoteBackup` reconciliation, we don’t need to explicitly set a retry here.

pkg/volumeremotebackup/longhorn/restore.go
WebberHuang1118: Will update per your suggestion, thanks.

pkg/volumeremotebackup/longhorn/restore.go
WebberHuang1118: No—if we set the `VolumeSnapshot` owner reference to `VolumeRemoteRestore`, then deleting `VolumeRemoteRestore` would also delete the backup on the Longhorn side. That’s why I added special logic in `func (lbr *LHRestoreOperation) Delete()`.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
WebberHuang1118: `relatedresource.Watch` is not a direct replacement here because these handlers resolve related resources through both owner references and annotations. 

Also, `volumeSnapshotHandler` and `pvcHandler` may enqueue different owner types, including `VolumeRemoteBackup` and `VolumeRemoteRestore`, while `relatedresource.Watch` is better suited to a single resolver/enqueuer relationship.

ihcsim (dismissed):

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
ihcsim: i thought the `volumeSnapshotHandler` and `pvcHandler` would only have either one backup resolver or one restore resolver, but never more than one resolvers? also, the `Watch` resolver is just a function which we should be able to implement to handle both resolve by ownerref or by annotations?

ihcsim:
@WebberHuang1118 just one last question at https://github.com/harvester/harvester/pull/9925#pullrequestreview-3955504085 regarding not using remote backup target. The rest LGTM.

WebberHuang1118:
> To confirm my understanding on Harvester side, we are not writing to nor reading from the remote backup target, right? Everything is managed as volume snapshots. I thought in the HEP we mentioned using the remote backup target.

Yes—Harvester and the Harvester CSI driver themselves don’t write to (or even read from) the remote backup target. The underlying implementation is in Longhorn: Longhorn interacts with the remote backup target, and it happens purely within the Longhorn layer.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
```diff
@@ -0,0 +1,289 @@
...
+	op driver.RestoreOperation,
+) (*harvesterv1.VolumeRemoteRestore, error) {
+	if err := op.Create(vrr); err != nil {
+		return vrr, err
+	}
+	logrus.Infof("Successfully created resource for PVCRestore %s/%s", rrh.ro.GetNamespace(vrr), rrh.ro.GetName(vrr))
+	return rrh.ro.SetProcessing(vrr)
+}
+
+func (rrh *remoteRestoreHandler) handleReadyRestore(vrr *harvesterv1.VolumeRemoteRestore) (*harvesterv1.VolumeRemoteRestore, error) {
```
Vicente-Cheng, line 167: just curious, compared to handleReadyBackup()
Don't we have the corresponding Finalize on the restore operator?

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
Vicente-Cheng: Not sure if I am understanding correctly, but it seems we have both backup/restore resolver for the vsh. resolvers.

pkg/volumeremotebackup/longhorn/backup.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	if srcProvisioner != "" {
+		vs.Annotations[util.AnnotationStorageProvisioner] = srcProvisioner
+	}
+
+	return lbo.vsClient.Create(vs)
+}
+
+// BuildOwnerReference creates an owner reference for a PVC backup.
+func (lbo *LHBackupOperation) BuildOwnerReference(vrb *harvesterv1.VolumeRemoteBackup) metav1.OwnerReference {
```
Vicente-Cheng, line 112: I saw the discussion you had before. Could we have a clearer comment here for the purpose of this owner reference and the potential risk?

pkg/volumeremotebackup/longhorn/restore.go
```diff
@@ -0,0 +1,559 @@
...
+		vscCache:     vscCache,
+		vscClient:    vscClient,
+		pvcCache:     pvcCache,
+		pvcClient:    pvcClient,
+		scCache:      scCache,
+	}
+}
+
+// BuildOwnerReference creates an owner reference for a PVCRestore.
+func (lbr *LHRestoreOperation) BuildOwnerReference(vrr *harvesterv1.VolumeRemoteRestore) metav1.OwnerReference {
```
Vicente-Cheng, line 69: ditto, I believe that would be the same reason, but we would like to have a clear comment for other developers.

pkg/volumeremotebackup/longhorn/restore.go
```diff
@@ -0,0 +1,559 @@
...
+		return err
+	}
+
+	// Create VolumeSnapshotContent, VolumeSnapshot, and PVC
+	return lbr.ensureRestoreResourcesExist(vrr, vrb, vsClass)
+}
+
+// createOrGet is a helper function that attempts to create a resource, and if it already exists,
+// retrieves it from the cache instead. This reduces repetitive error handling code.
+func createOrGet[T any](
```
Vicente-Cheng, line 122: I think we created some generic common function on this PR. 
Please consider moving that to the https://github.com/harvester/go-common in the future.

Vicente-Cheng (dismissed):
lgtm, just a few nits, thanks for the effort!

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
ihcsim: Currently, the snapshot and PVC handlers hold their own references to both the backup and restore resolvers. From what I can see, the backup resolver is only used with backup objects while the restore resolver is only used with restore objects.

Furthermore, the backup resolver only performs ownerref resolution for snapshots, while the restore resolver reacts to changes to both snapshots and PVCs, by first performing ownerref resolution, and if that didn't work, do annotation resolution.

To me, that corresponds to a `Watch` for snapshots in the backup handler and a `Watch` for snapshots and PVCs in the restore handler.

Here's the diff of what I think is the equivalence of using `Watch`:

```diff
diff --git a/pkg/controller/master/volumeremotebackup/register.go b/pkg/controller/master/volumeremotebackup/register.go
index 67036b715..3e6f3ac03 100644
--- a/pkg/controller/master/volumeremotebackup/register.go
+++ b/pkg/controller/master/volumeremotebackup/register.go
@@ -2,10 +2,13 @@ package pvcbackup
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	ctlstoragev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/config"
@@ -14,6 +17,7 @@ import (
 	"github.com/harvester/harvester/pkg/volumeremotebackup/common"
 	"github.com/harvester/harvester/pkg/volumeremotebackup/driver"
 	"github.com/harvester/harvester/pkg/volumeremotebackup/longhorn"
+	"github.com/rancher/wrangler/v3/pkg/relatedresource"
 )
 
 const (
@@ -253,6 +257,64 @@ func registerBackupHandlers(
 
 	vrbs.OnChange(ctx, pvcbackupControllerName, pvcbackupHandler.OnChanged)
 	vrbs.OnRemove(ctx, pvcbackupControllerName, pvcbackupHandler.OnRemove)
+	relatedresource.Watch(ctx, "backup-snapshot-watcher", pvcbackupHandler.resolve, vrbs, vss)
+}
+
+func (b *remoteBackupHandler) resolve(_, _ string, obj runtime.Object) ([]relatedresource.Key, error) {
+	if meta, ok := obj.(metav1.Object); ok {
+		return resolveObjectOwnerRef(meta)
+	}
+	// backup handler doesn't use annotation-based references
+	return nil, nil
+}
+
+func resolveObjectOwnerRef(obj metav1.Object) ([]relatedresource.Key, error) {
+	for _, ownerRef := range obj.GetOwnerReferences() {
+		if ownerRef.Kind == remoteBackupKind.Kind // make kind an argument {
+			return []relatedresource.Key{
+				{
+					Namespace: obj.GetNamespace(),
+					Name:      ownerRef.Name,
+				},
+			}, nil
+		}
+	}
+
+	return nil, nil
+}
+
+func resolveObjectOwnerRefWithAnnotation(obj metav1.Object) ([]relatedresource.Key, error) {
+	ref, exists := obj.GetAnnotations()[driver.AnnotationPVCRestoreRef]
+	if !exists || ref == "" {
+		return nil, fmt.Errorf("")
+	}
+
+	namespacedName := strings.Split(ref, "/")
+	if len(namespacedName) == 2 {
+		return []relatedresource.Key{
+			{
+				Namespace: namespacedName[0],
+				Name:      namespacedName[1],
+			},
+		}, nil
+	}
+
+	return nil, fmt.Errorf("")
+}
+
+func (b *remoteRestoreHandler) resolve(ns, name string, obj runtime.Object) ([]relatedresource.Key, error) {
+	meta, ok := obj.(metav1.Object)
+	if !ok {
+		return nil, nil
+	}
+
+	keys, err := resolveObjectOwnerRef(meta)
+	if len(keys) > 0 && err == nil {
+		return keys, nil
+	}
+
+	// If no owner reference found, try to resolve using annotation
+	return resolveObjectOwnerRefWithAnnotation(meta)
 }
 
 func registerRestoreHandlers(
@@ -300,4 +362,6 @@ func registerRestoreHandlers(
 
 	vrrs.OnChange(ctx, pvcrestoreControllerName, remoteRestoreHandler.OnChanged)
 	vrrs.OnRemove(ctx, pvcrestoreControllerName, remoteRestoreHandler.OnRemove)
+	relatedresource.Watch(ctx, "restore-snapshot-watcher", remoteRestoreHandler.resolve, vrrs, vss)
+	relatedresource.Watch(ctx, "restore-pvc-watcher", remoteRestoreHandler.resolve, vrrs, pvcs)
 }
```

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
WebberHuang1118: For `VolumeRemoteBackup`, in the original implementation I had multiple setters, and each setter issued an API request. As @ihcsim suggested ([discussion](https://github.com/harvester/harvester/pull/9925#discussion_r2914533044)), I squashed those into a single API request.

For `VolumeRemoteRestore`, since there are only two setters, I didn’t specifically squash them.

pkg/volumeremotebackup/longhorn/restore.go
WebberHuang1118: Sure, I've created [issue #10248](https://github.com/harvester/harvester/issues/10248) to track this.

pkg/volumeremotebackup/longhorn/backup.go
WebberHuang1118: Updated per your suggestion, thanks.

pkg/volumeremotebackup/longhorn/restore.go
WebberHuang1118: Updated per your suggestion, thanks.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
WebberHuang1118: The `vsh` resolvers include `remotebackupHandler` and `remoteRestoreHandler` because the CSI `VolumeSnapshot` is responsible for triggering the reconciliation of `VolumeRemoteBackup` and `VolumeRemoteRestore`.

The CSI `VolumeSnapshot` connects to `VolumeRemoteBackup` via an owner reference, and to `VolumeRemoteRestore` via an annotation. The reason `VolumeRemoteRestore` cannot use an owner reference to connect to the CSI `VolumeSnapshot` is detailed in [this PR comment](https://github.com/harvester/harvester/pull/9925#discussion_r2944226339).

Additionally, the `pvch` resolvers include `remoteRestoreHandler` because the PVC is also responsible for triggering the reconciliation of `VolumeRemoteRestore`.

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
Vicente-Cheng: got it!

Vicente-Cheng (approved):

WebberHuang1118:
@Mergifyio backport v1.8

pkg/controller/master/volumeremotebackup/volumeremotebackup.go
WebberHuang1118: @ihcsim Considering the schedule and as discussed with @Vicente-Cheng, I merged the PR first. We can make further improvements later if needed. Thanks for your feedback!

PR #9935, opened 2026-01-23
additional changes to pod mutator to allow support for forklift

ibrokethecloud:
#### Problem:

When using forklift with harvester, the disk conversion by the virt-v2v fails with the cryptic error

```
No interfaces with usable IPv6 routes
UNIX domain socket bound at /tmp/libguestfs4z8aH8/passt.sock
Couldn't create user namespace: Operation not permitted
libguestfs: trace: v2v: launch = -1 (error)
virt-v2v: error: libguestfs error: passt exited with status 1
rm -rf -- '/tmp/v2vnbdkit.HVbzom'
```

This seems to stem from the securityContext and secCompProfile applied to the virt-v2v pod when running on non openshift clusters.

#### Solution:

The PR adds a minor change to harvester webhook to mutate virt-v2v pods to have privileged permissions to allow successful conversion of disks.

In addition the converter pod container use the ImagePullPolicy of `Always`. The mutator also mutates this to `IfNotPresent`

#### Related Issue(s):

https://github.com/harvester/suse-virtualization-mgmt/issues/18

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -129,6 +129,7 @@ require (
 	kubevirt.io/kubevirt v1.7.0
 	sigs.k8s.io/cluster-api v1.9.5
 	sigs.k8s.io/controller-runtime v0.21.0
+	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8
 	sigs.k8s.io/kind v0.30.0
 	sigs.k8s.io/kustomize/kyaml v0.19.0
 	sigs.k8s.io/yaml v1.6.0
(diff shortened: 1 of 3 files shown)


wheatdog (approved):

tserong (approved):
LGTM - the only question I have is, do we care about what Copilot mentions about bypassing HTTP proxy and additional CA certificate patches for virt-v2v pods?

ibrokethecloud:
those changes are not needed as forklift has its own mechanism to define cert fingerprint via secrets based on source

ihcsim (approved):
LGTM - Only thought I have is whether we want to add a `test` patch op before the `replace` op to ensure the replacement only happens if the patch path exists.

ihcsim:
Another alternative is to represent this using a CEL mutating admission policy and bundle it with the forklift-operator Helm chart. Then we can keep forklift stuff out of the core code base.

ibrokethecloud:
i like the idea of the Mutating admission policy https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy

creating this task to track the maturity of mutating admission policy and we can convert this webhook and package it with the forklift chart itself

https://github.com/harvester/harvester/issues/10128

Issue #9779, opened 2025-12-22
[ENHANCEMENT] Relax the live migration feasibility check for VMs while using CD-ROM


wheatdog:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

In https://github.com/harvester/harvester/pull/3463, we introduce additional constrain for live migration as mentioned in [Harvester doc](https://docs.harvesterhci.io/v1.6/vm/live-migration#non-migratable-virtual-machines). This limitation is main reason for feature like https://github.com/harvester/harvester/issues/3880.

> Volume with the following properties:
> - Type: CD-ROM or Container Disk
> - Access mode: ReadWriteOnce
> - StorageClass replica count: 1 (This is not detected in all cases.)

This limitation is related to this one in [KubeVirt doc](https://kubevirt.io/user-guide/compute/live_migration/#limitations)

> Virtual machines using a PersistentVolumeClaim (PVC) must have a shared ReadWriteMany (RWX) access mode to be live migrated.

### Describe the solution you'd like

We should consider revise the logic to relax the constrain introduce in https://github.com/harvester/harvester/pull/3463. Instead of checking CD-ROM / container disk, we should check pvc access mode as the one mentioned in KubeVirt doc. 

Or, we could consider directly refer to the conditions `LiveMigratable` and `StorageLiveMigratable` in VM if suitable.

With this revision, the need for https://github.com/harvester/harvester/issues/3880 could be mitigated.

wheatdog:
related to #9261

innobead:
> related to [#9261](https://github.com/harvester/harvester/issues/9261)

JFYI. We can leverage the relationship setting to set the relationship. 

Moved this back to 1.8.0.

wheatdog:
I see. Thanks!

TachunLin:
Additional context: 
1. We can consider testing the VM migration with container disk

TachunLin:
Verified **fixed** on `v1.8.0-dev-260301`. Close this issue.
 
### Result

<details><summary>$\color{green}{\textsf{PASS}}$ Can migrate a running VM <strong>without</strong> a CD-ROM device to other nodes $~~$</summary>

1. Can migrate a VM with No Media cdrom device to another node

https://github.com/user-attachments/assets/da96f9f1-76ea-49ef-80a5-e079f8c16bfd

 </details>


[...]

PR #9937, opened 2026-01-23
feat: add CD-ROM hotplug volumes

wheatdog:
#### Problem:

To support inserting and ejecting CD-ROM volumes without a restart.

#### Solution:

According to the [announcement](https://github.com/harvester/harvester/issues/9261#user-content-fn-1-29143c8d52e5c4dff7572d6891417a0b), a new feature gate "DeclarativeHotplugVolumes" is introduced at KubeVirt 1.6 and the existing "HotplugVolumes" is going to be deprecated.

1. Migrate the feature gate from "HotplugVolumes" to "DeclarativeHotplugVolumes"
2. Add 2 new VM actions:
    1. `injectCdRomVolume`: add a new entry to volumes from VM spec and volumeClaimTemplates from VM annotation 
    2. `ejectCdRomVolume`: remove the corresponding element from volumes and volumeClaimTemplates from VM. Then, delete the corresponding volume (pvc).
3. Remove the existing VM action `ejectCdRom` and related checks -> prefer the new action `ejectCdRomVolume` that doesn't require a restart.
4. ~~VM mutator webhook would patch hotpluggable to true for volumes of SATA CD-ROM devices to enable hotplug volume automatically.~~ -> For flexibility, this is handled from FE.
5. ~~For existing VMs with CD-ROM, the hotpluggable field would be patched when the VM is stopped / restarted. This is handled by `vmControllerSetSataCdRomHotpluggable`.~~ -> Old VMs would act as is: see https://github.com/harvester/harvester/pull/9840#discussion_r2729889106
6. Relax the live migration constrain for VM with CD-ROM devices and container disks.
7. Use `kubernetes.Interface` instead of `kubernetes.Clientset` for type annotations since it's an interface, which is easier to be mocked in unit tests.

#### Related Issue(s):

#9261 
#9779

#### Test plan:

1. **Feature Gate Verification**: After enabling the `DeclarativeHotplugVolumes` feature gate, perform basic checks to ensure existing volume hot-plug functionalities are not regressed.
2. **Volume Insertion**: For a running VM with an empty CD-ROM device, verify that a new image can be successfully inserted and that the content is accessible from within the guest OS.
3. **Volume Ejection**: For a running VM with an occupied CD-ROM device, verify that the volume can be successfully ejected and is no longer accessible from within the guest OS. Additionally, the corresponding volume is deleted.
4. **Live Migration with CD-ROM**: Verify that a running VM with a CD-ROM can be successfully live-migrated.
5. **Existing VM Compatibility**: For a VM created before this enhancement, verify that after a restart, the CD-ROM volume becomes ejectable.

Additionally, please refer to the test plan in HEP: https://github.com/harvester/harvester/pull/9840

https://kubevirt.io/user-guide/storage/hotplug_volumes/#inject-cd-rom


pkg/api/vm/handler.go
```diff
@@ -300,22 +302,120 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
...
+	if err != nil {
 		return err
 	}
 
-	if !reflect.DeepEqual(vm, vmCopy) {
-		if _, err := h.vms.Update(vmCopy); err != nil {
+	volumeMode := corev1.PersistentVolumeBlock
+	newPvc := corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-%s-%s", name, input.DeviceName, rand.String(5)),
```
w13915984028: how do we ensure the generated name does not over limits (253 chars)

pkg/api/vm/formatter.go
```diff
@@ -90,10 +92,8 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	resource.AddAction(request, addVolume)
 	resource.AddAction(request, removeVolume)
 	resource.AddAction(request, cloneVM)
-
-	if canEjectCdRom(vm) {
-		resource.AddAction(request, ejectCdRom)
-	}
+	resource.AddAction(request, insertCdRomVolume)
```
w13915984028: per following code, the original logic to check if a place-holder cdrom device seems still make sense, only if a cd-rom device is there, the insert action could be triggered

pkg/api/vm/handler.go
```diff
@@ -300,22 +302,120 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
...
+			PersistentVolumeClaim: &kubevirtv1.PersistentVolumeClaimVolumeSource{
+				PersistentVolumeClaimVolumeSource: corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: newPvc.Name,
+				},
+				Hotpluggable: true,
+			},
+		},
+	}
+
+	pos, err := getSataCdRomDiskPos(vm, input.DeviceName)
```
w13915984028: move L368 before L337? to check if a pos can be found first

pkg/api/vm/handler.go
```diff
@@ -300,22 +302,120 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
...
+		return err
+	}
+
+	vmCopy.Spec.Template.Spec.Volumes = volumes
+	if _, err = h.vms.Update(vmCopy); err != nil {
+		return err
+	}
+
+	for _, name := range toRemoveClaimNames {
+		if err := h.clientSet.CoreV1().PersistentVolumeClaims(vm.Namespace).Delete(ctx, name, metav1.DeleteOptions{}); err != nil {
```
w13915984028: skip isNotFound error ?

pkg/api/vm/handler.go
```diff
@@ -371,35 +471,22 @@ func (h *vmActionHandler) stopVM(namespace, name string) error {
...
 
-	if err := removeVolumeClaimTemplatesFromVMAnnotation(vm, toRemoveClaimNames); err != nil {
+	volumeClaimTemplates = slices.Insert(volumeClaimTemplates, pos, pvc)
+	updateVolumeClaimTemplateBytes, err := json.Marshal(volumeClaimTemplates)
+	if err != nil {
 		return err
 	}
-	vm.Spec.Template.Spec.Volumes = volumes
-	vm.Spec.Template.Spec.Domain.Devices.Disks = disks
+	vm.Annotations[util.AnnotationVolumeClaimTemplates] = string(updateVolumeClaimTemplateBytes)
```
w13915984028, line 486: when write to annotation, check if it is nil, it is safe to read a nil map but panic to write

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -524,3 +525,46 @@ func (h *VMController) getPVCStorageClass(pvc *corev1.PersistentVolumeClaim) (*s
...
+		return nil, nil
+	}
+
+	cdRomDiskNames := make(map[string]struct{})
+	for _, disk := range vm.Spec.Template.Spec.Domain.Devices.Disks {
+		if disk.CDRom != nil && disk.CDRom.Bus == kubevirtv1.DiskBusSATA {
+			cdRomDiskNames[disk.Name] = struct{}{}
+		}
+	}
+
```
w13915984028: could check len(cdRomDiskNames) == 0 to return earlier

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -524,3 +525,46 @@ func (h *VMController) getPVCStorageClass(pvc *corev1.PersistentVolumeClaim) (*s
...
+		if _, exists := cdRomDiskNames[volume.Name]; exists {
+			if volume.PersistentVolumeClaim != nil && !volume.PersistentVolumeClaim.Hotpluggable {
+				volume.PersistentVolumeClaim.Hotpluggable = true
+			}
+		}
+	}
+
+	if _, err := h.vmClient.Update(vmCopy); err != nil {
+		return vmi, err
+	}
```
w13915984028: two questions on this controller

(1) it is a bit over-kill to always do such check on vmi change, could we add kind of annotations to vm to avoid later check?

(2) If VM is still running, the patch might cause kubevirt complains about the restart action is required as vm spec is changed, did you check if this happen?  

the alternative is to do the patch when vm is put to off

w13915984028 (commented):
Thanks for the PR, some questions:

pkg/api/vm/handler.go
```diff
@@ -300,22 +302,120 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
...
 	vm, err := h.vmCache.Get(namespace, name)
 	if err != nil {
 		return err
 	}
 
 	vmCopy := vm.DeepCopy()
-	if err := ejectCdRomFromVM(vmCopy, diskNames); err != nil {
+
+	parts := strings.SplitN(input.ImageName, "/", 2)
+	if len(parts) != 2 {
```
w13915984028, line 320: it also deserves a double check, sometime if an object does not have/include namespace, it fallback to the `default` ns, should we tolerant this case?

pkg/controller/master/virtualmachine/vm_controller.go
wheatdog: Thanks for bringing this up. I've removed the related code. We decide to not have this behavior and old VMs would act as is: see https://github.com/harvester/harvester/pull/9840#discussion_r2729889106 for related discussion of the design.

pkg/api/vm/handler.go
wheatdog: good point, let me try to cap the prefix.

pkg/controller/master/virtualmachine/vm_controller.go
wheatdog: thanks, changes to vm_controller are reverted - https://github.com/harvester/harvester/pull/9937#discussion_r2752745269

pkg/api/vm/handler.go
wheatdog: At the beginning of this function, we already have this that check the existence of the map key.
```
volumeClaimTemplatesStr, ok := vm.Annotations[util.AnnotationVolumeClaimTemplates]
if !ok {
	return nil
}
```

I think it's safe to directly write here.

pkg/api/vm/formatter.go
wheatdog: sounds good. Let me add some checks here before exposing those new actions.

pkg/api/vm/handler.go
wheatdog: Thanks, I prefer that the object is explicitly namespaced `<namespace>/<name>` if possible. This is also the id used in `/v1/harvester/harvesterhci.io.virtualmachineimages`.

pkg/api/vm/handler.go
wheatdog: addressed by https://github.com/harvester/harvester/pull/9937/commits/04a51f2ea61002b8585dcc9ff932ba38e08b7186

pkg/api/vm/handler.go
wheatdog: addressed by https://github.com/harvester/harvester/pull/9937/commits/a71b8a1544cc2f1452374a2e9c606a8d4d6238d3

pkg/api/vm/handler.go
wheatdog: addressed by https://github.com/harvester/harvester/pull/9937/commits/7deb1cc8736769f7b736f76a2d55073e68937845

pkg/api/vm/formatter.go
wheatdog: addressed by https://github.com/harvester/harvester/pull/9937/commits/b0cbd5c3015aa173fe72ed678b158ad885ace5ed

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -174,3 +174,41 @@ func volumeSupportRWXForVM(accessModes []corev1.PersistentVolumeAccessMode, prov
 
 	return slices.Contains(accessModes, corev1.ReadWriteMany)
 }
+
+func IsDiskSataCdRom(disk *kubevirtv1.Disk) bool {
```
ibrokethecloud: this may not be a valid check

https://kubevirt.io/user-guide/cluster_admin/virtual_machines_on_Arm64/#disks-and-volumes

ARM only supports SCSI or Virtio based CDROMs so we need to factor this in.

wheatdog: This is a limitation of KubeVirt's support for cdrom hotplug volume to use SATA only: https://kubevirt.io/user-guide/storage/hotplug_volumes/#supported-disk-busses

related issue: https://github.com/kubevirt/kubevirt/issues/16310

So, cdrom hotplug volume is not supported in ARM clusters. I think this deserves a note in the Harvester doc.

ibrokethecloud: we need to gate the actions to check OS ARCH, and not add cdrom hotplug actions on ARM environments

wheatdog: With the current implementation, 
- injectCdRomVolume action is gated by `SupportInjectCdRomVolume`
- ejectCdRomVolume action is gated by `SupportEjectCdRomVolume`

They follow the requirements from [KubeVirt's doc](https://kubevirt.io/user-guide/storage/hotplug_volumes/) and already satisfy the suggestion to not add cdrom hotplug actions on ARM environments. Do you think we still need to explicitly check OS ARCH?

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -174,3 +174,57 @@ func volumeSupportRWXForVM(accessModes []corev1.PersistentVolumeAccessMode, prov
...
+	return disk.CDRom != nil && disk.CDRom.Bus == kubevirtv1.DiskBusSATA
+}
+
+func SupportInjectCdRomVolume(vm *kubevirtv1.VirtualMachine) (bool, error) {
+	volumeMaps := map[string]struct{}{}
```
ibrokethecloud, line 192: We should add this to ensure the check returns false on arm hosts and eventually the UI actions for hot pluggable CDROM is not available
```suggestion
    if runtime.GOARCH == "arm64" {
        return false, nil
    }
	volumeMaps := map[string]struct{}{}
```

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -174,3 +174,57 @@ func volumeSupportRWXForVM(accessModes []corev1.PersistentVolumeAccessMode, prov
...
+		if !useSata {
+			return false, fmt.Errorf("empty cd-rom device should connect via SATA bus")
+		}
+		hasEmptyCdRom = true
+	}
+	return hasEmptyCdRom, nil
+}
+
+func SupportEjectCdRomVolume(vm *kubevirtv1.VirtualMachine) (bool, error) {
+	volumeMaps := map[string]struct{}{}
```
ibrokethecloud, line 216: same as above

```suggestion
    if runtime.GOARCH == "arm64" {
        return false, nil
    }
	volumeMaps := map[string]struct{}{}
```

pkg/api/vm/handler.go
```diff
@@ -301,22 +304,131 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
 	return nil, nil
 }
 
-func (h *vmActionHandler) ejectCdRom(ctx context.Context, name, namespace string, diskNames []string) error {
+func getSataCdRomDiskPos(vm *kubevirtv1.VirtualMachine, deviceName string) (int, error) {
```
ibrokethecloud: I am a bit confused by this check and why we need to return device position

based on my reading https://kubevirt.io/user-guide/storage/hotplug_volumes/ as long as there is a cdrom available in the `vm.spec.template.spec.devices[*].disk` of type cdrom with a sata bus, we just need to add a hot pluggable volume in `vm.spec.template.volumes` with the same name as `disk` found from `vm.spec.template.spec.devices[*].disk`.

If there is no volume already attached to the cdrom disk, then we can hot plug a disk or if there is one present we can unplug the disk.

Looking and using the position to insert the volume is redundant.

pkg/util/virtualmachine/virtualmachine.go
wheatdog: sure, I'll revise this to explicitly handle for arm64. Thanks.

wheatdog: addressed by https://github.com/harvester/harvester/pull/9937/commits/13d382ee4e8928bd5d56a1a9042058b53c511861

pkg/util/virtualmachine/virtualmachine.go
wheatdog: I have revised this since https://github.com/harvester/harvester/pull/9937/commits/1244bf7e3c53d9b9a4fbc25d3a715af62d26384e

pkg/api/vm/handler.go
```diff
@@ -301,22 +304,114 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
...
+		return err
+	}
+
+	vmCopy.Spec.Template.Spec.Volumes = volumes
+	if _, err = h.vms.Update(vmCopy); err != nil {
+		return err
+	}
+
+	for _, name := range toRemoveClaimNames {
+		if err := h.clientSet.CoreV1().PersistentVolumeClaims(vm.Namespace).Delete(ctx, name, metav1.DeleteOptions{}); err != nil {
```
wheatdog: I have noticed that this line sometimes could cause error like the one below that prevent the volume to be deleted.

Image

It seems related to the following snippet and could be related to the freshness of VMByPVCIndex. 
https://github.com/harvester/harvester/blob/c50975695f06b36932879bf6b3e7c37caf220b07/pkg/webhook/resources/persistentvolumeclaim/validator.go#L100-L110

@ibrokethecloud @w13915984028 I'd like to consult with you on how to handle case like this to "eventually" delete the volume. Thanks!

Should I add an annotation to the PVC first and add a logic in controller to garbage collect it later?

wheatdog: I'm trying retry for deletion: https://github.com/harvester/harvester/pull/9937/commits/3581e6f9a2c5293d25e7c75fe00a10cb737286f5

w13915984028: seems retry is the feasible solution atm.

the webhook check is expected to keep unchanged, e.g. another closed PR https://github.com/harvester/harvester/pull/9686;  it could block the deletion due to cache out-of-sync, and the controller has the context to try a bit more, and will success in most cases

ibrokethecloud (dismissed):
lgtm. thanks.

pkg/api/vm/handler.go
wheatdog: thanks, sorry I missed this comment. And yes, this turns out not just redundant but also problematic for cases like error out when inserting image to the last one of multiple empty cdrom. I've removed this and simplified the logic.

wheatdog: already addressed in https://github.com/harvester/harvester/pull/9937/commits/3528e332aa5cb898e1de5134d79081e1c068cab5

pkg/api/vm/handler.go
wheatdog, line 1199: I found a regression to "Add Volume".  The API return errors but the volume is actually attached successfully. 

> failed to update vm default/hp, error: Operation cannot be fulfilled on virtualmachines.kubevirt.io "hp": the object has been modified; please apply your changes to the latest version and try again

https://github.com/user-attachments/assets/88273c1e-dfe2-460e-9297-1e581cb66964

It seems that "Detach Volume" is fine.

I'll take a look.

wheatdog: This error comes from 

https://github.com/harvester/harvester/blob/5c832461f4e6d3c92e9f82a90a4ca0278b27a5a7/pkg/api/vm/handler.go#L957-L965

wheatdog: @w13915984028 @ibrokethecloud in https://github.com/harvester/harvester/pull/9937/commits/1d28d54e6039593074153fc0790691fc8cfc7965, I've migrated the existing addVolume/removeVolume functions to [the new declarative way](https://kubevirt.io/user-guide/storage/hotplug_volumes/#hotplug-disklun) such that update is called once to prevent the error.


https://github.com/user-attachments/assets/a914dc31-937e-4b5f-929a-38671315ca37

pkg/api/vm/formatter.go
```diff
@@ -485,3 +479,21 @@ func canHotUnplugNic(vm *kubevirtv1.VirtualMachine) bool {
 	ok, _ := virtualmachine.SupportHotUnplugNic(vm)
 	return ok
 }
+
+func canInsertCdRomVolume(vm *kubevirtv1.VirtualMachine) bool {
+	if runtime.GOARCH == "arm64" {
```
w13915984028: it is better if the `arm64` is defined as a const for public usage

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -174,3 +174,66 @@ func volumeSupportRWXForVM(accessModes []corev1.PersistentVolumeAccessMode, prov
...
+	for _, volume := range vm.Spec.Template.Spec.Volumes {
+		if volume.PersistentVolumeClaim != nil && volume.PersistentVolumeClaim.Hotpluggable {
+			volumeMaps[volume.Name] = struct{}{}
+		}
+	}
+
```
w13915984028, line 222: L222 check `len(volumeMaps) == 0` to return early

pkg/util/virtualmachine/virtualmachine.go
```diff
@@ -174,3 +174,66 @@ func volumeSupportRWXForVM(accessModes []corev1.PersistentVolumeAccessMode, prov
...
+	for _, disk := range vm.Spec.Template.Spec.Domain.Devices.Disks {
+		if disk.CDRom == nil {
+			continue
+		}
+		_, hotpluggable := volumeMaps[disk.Name]
+		if !hotpluggable {
+			continue
+		}
+		useSata := isDiskSataCdRom(&disk)
+		if !useSata {
```
w13915984028, line 237: this function is used for both UI menu render and webhook check, there are some differences:

UI render does not care error, as long as the first target is found, it could return directly, e.g. L236 could trigger return directly; and L233 could also cause UI render failure

webhook/controller check all devices to ensure no error

could we split it into two functions for each purpose?

pkg/api/vm/handler.go
```diff
@@ -301,25 +304,118 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
...
+
+	vmImage, err := h.vmImageCache.Get(parts[0], parts[1])
+	if err != nil {
 		return err
 	}
 
-	if !reflect.DeepEqual(vm, vmCopy) {
-		if _, err := h.vms.Update(vmCopy); err != nil {
-			return err
+	imgSize := max(vmImage.Status.VirtualSize, vmImage.Status.Size)
```
w13915984028, line 329: could we put L329~L366 to a new function and make insertCdRomVolume() shorter

w13915984028 (commented):
A few last questions, thanks.

pkg/api/vm/handler.go
wheatdog: I prefer keeping it as is. The most of the space it took is related to initialization of pvc and kubevirt volume, which is the main point of this function.

pkg/api/vm/formatter.go
wheatdog: sure.

pkg/api/vm/handler.go
w13915984028: I have no strong options here, as long as CI does not complain.

pkg/api/vm/formatter.go
brandboat: What about this? We have done something similar for arm arch checking before:

https://github.com/harvester/harvester/blob/8639cfb12122194e8fa52acf53426d719e8cdd9f/pkg/webhook/resources/virtualmachine/mutator.go#L369-L384

wheatdog: This looks good! I'll use this one instead. Thanks!

wheatdog: Oops, `isARMArchitecture` require a `KubeVirtCache` to perform `kv.Status.DefaultArchitecture == "arm64"`. I would like to stick with the original `runtime.GOARCH` for now. Thanks for the suggestion still!

brandboat: It's fine, this is more like a nit, the original is good enough as I don't think we have mixed arch use case for now.

pkg/util/virtualmachine/virtualmachine.go
wheatdog: The idea is to separate 2 intentions:
1. Capability of insertion/ejection
2. Correctness of VM spec


Though, as offline discussed, I would like to stick with the current implementation for now. Thanks for the suggestion still!

> I still kinda prefer having one function to handle both intentions since
> 
> 1. They are quite related. To determine the capability of insertion/ejection, we also need to validate the correctness of the corresponding part of VM spec.
> 2. Their implementations are quite similar. They iterate volumes and disks in VM spec.
> 
> As for UI, it has its own logic canDoVolumeHotplugAction in [UI PR](https://github.com/harvester/harvester-ui-extension/pull/703/changes) instead of relying on VM action APIs.

pkg/api/vm/formatter.go
wheatdog: fixed by https://github.com/harvester/harvester/pull/9937/commits/2fdd817b2ac346603f79be31564956e574bf0d78

pkg/util/virtualmachine/virtualmachine.go
wheatdog: fixed by https://github.com/harvester/harvester/pull/9937/commits/5762271fe39766bc1355c8657d95d22d4be64383

w13915984028 (dismissed):
LGTM, thanks.

BTW, please squash the commits manually or utlize the gihtub merge &  cut off some commit messages to make the final commit message shorter.

ibrokethecloud (dismissed):

ibrokethecloud (approved):

w13915984028 (approved):
LGTM, thanks.

Vicente-Cheng (approved):

PR #9941, opened 2026-01-26
fix: skip patching VM-less VMI in mutator

wheatdog:
#### Problem:

Couldn't create a VMI directly since the VMI mutator currently expect the existence of the wrapping VM CR.

#### Solution:

Skip patching VM-less VMI in mutator.

#### Related Issue(s):

#9172 

#### Test plan:

- Create a VMI directly via kubectl

N/A

### DIFF
--- a/pkg/webhook/resources/virtualmachineinstance/mutator.go
+++ b/pkg/webhook/resources/virtualmachineinstance/mutator.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	kubevirtv1 "kubevirt.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pohanhuang (approved):
LGTM

pohanhuang:
Hi @wheatdog ,

After some thought about this PR, I have a question.

Since #9172 is about ephemeral VMs for GitLab runners, have we considered just helping users create a VM + VMI pair together instead of allowing bare VMI?

The way I see it:
- Current fix: Skip the mutator for bare VMI (works, but leaves bare VMI as a supported path)
- Alternative: Provide a pattern or API for "create VM + VMI atomically" (keeps architecture clean)

What do you think?

wheatdog:
Hi @pohanhuang, thanks for bringing this up. I prefer this solution since
- creating VM-less VMIs directly is supported in vanilla KubeVirt
- this limitation was a side effect of https://github.com/harvester/harvester/pull/8339, which is part of the solution to preserve MAC addresses of VMs and shouldn't affect VMI creation

The goal is to unblock power users, like the reporter of this issue, to use most of the feature in vanilla KubeVirt.

> have we considered just helping users create a VM + VMI pair together instead of allowing bare VMI

> Provide a pattern or API for "create VM + VMI atomically"

As mentioned in [KubeVirt's doc](https://kubevirt.io/user-guide/architecture/#application-layout), it's suggested to not create VMI directly but by a high level resource, and guess what, like VirtualMachine (VM)! This is the typical usage in Harvester. So I think we don't need to build anything new to co-create VM + VMI together. They could use VM if they want.

pohanhuang:
Ah, makes sense! Thanks for explaining! The fix looks good to me.

wheatdog:
@harvester/dev could you help me review this short PR when you are available? I would like to have a few more reviews before merging. Thanks!

brandboat (approved):
LGTM

Issue #9753, opened 2025-12-16
[ENHANCEMENT] Enhance the resource-quota processing when overhead-ratio is adjusted


w13915984028:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

We noticed this combined scenrio:
(0) Install a cluster with the default overhead-ratio
(1) Set a name-space resource-quota e.g. 5Gi memory
(2) Run a VM with 4Gi memory
(3) Migrate the VM, it works
(4) Adjust the overhead-ratio from default "1.5" to a value like "8"

(5) Migrate the VM multi times, and it will be blocked claiming the RQ is not sufficient.  The pre-condition is: before the overhead-ratio is increased, the namespace RQ has been used upto like 90%.

For more details:

https://github.com/w13915984028/harvester-develop-summary/blob/main/ts-ref/v170-vm-migration-with-resource-quota-and-overhead-ratio.md

### Describe the solution you'd like

When `overhead-ratio` is adjusted, the already running VMs under a RQ controlled namespace, can still migrate. To let the existing workloads to run as smooth as possible.

### Describe alternatives you've considered

When overhead-ratio is adjusted, amend the namespace RQ as well, to get them matched.

albinsun:
Test FAIL when set resource quota.
With the set `rancher-v2.15.0-rc1`(`ui-extension-1.9.0-rc3`) + `harvester-v1.9.0-rc3`, **Resource Quota** can not be set on **Project** and **Namespace**.
Note that it works with `rancher-v2.14.3`(`ui-extension-v1.7.2`) + `harvester-v1.7.3-rc1`, see the **Reproduce** section.


Similar to harvester/harvester/issues/10556
Thanks @a110605's info.



### Logs
* Support bundle: [supportbundle_f7feba3d-aa50-479a-ab1b-c8cffd83522f_2026-07-30T08-50-09Z.zip](https://github.com/user-attachments/files/30538992/supportbundle_f7feba3d-aa50-479a-ab1b-c8cffd83522f_2026-07-30T08-50-09Z.zip)


### Environment
* `harvester-v1.9.0-rc3`
  * Profile: 4 nodes AMD64 Bare-metal HPE ProLiant DL360 
  * **ui-source**: `Auto`
* `rancher-v2.15.0-rc1`
  * `harvester-ui-extension-1.9.0-rc3`


### Test Scenario
1. <details>
   <summary> Install a cluster with the default overhead-ratio <code>1.5</code>
[...]

PR #9942, opened 2026-01-26
Enhance the RQ processing when overhead-ratio is involving

w13915984028:
#### Problem:

An already running VM might fail to migration if the system level setting [additional-guest-memory-overhead-ratio](https://docs.harvesterhci.io/v1.7/advanced/index#additional-guest-memory-overhead-ratio) is changed after the VM has started to run.

#### Solution:

For already running VMs, try best to let them continue to migrate, when they are stopped and then the quota will strictly restrict them to start again only if the quota has enough free space.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9753

document PR to explain the details: 

https://github.com/harvester/docs/pull/981

#### Test plan:

Per https://github.com/harvester/harvester/pull/9942#issuecomment-3915516457

Future enhancement: https://github.com/harvester/harvester/issues/10093


Felipalds (dismissed):

w13915984028:
Test report:

(1) Setup a 2+ node cluster
(2) Per doc https://github.com/w13915984028/harvester-develop-summary/blob/main/ts-ref/v170-vm-migration-with-resource-quota-and-overhead-ratio.md#environment-setup, create namespace and resourcequota `3 core, 5Gi memory`
(3) Create a VM with `2core, 4GI memory`
(4) Migrate VM mutli-times, all ok
(5) Set `additional-guest-memory-overhead-ratio` to a value like `10`
(6) Migrate the VM, with this PR, it is successful

// auto-scale per migrating vm (existing feature)
>time="2026-02-17T15:34:14Z" level=debug msg="resourcequota quota-test/default-cq7qs is updated from {Hard:map[limits.cpu:{i:{value:3 scale:0} d:{Dec:<nil>} s:3 Format:DecimalSI} limits.memory:{i:{value:5242880000 scale:0} d:{Dec:<nil>} s: Format:BinarySI}] Scopes:[] ScopeSelector:nil} to {Hard:map[limits.cpu:{i:{value:5000 scale:-3} d:{Dec:<nil>} s: Format:DecimalSI} limits.memory:{i:{value:12487836313 scale:0} d:{Dec:<nil>} s: Format:BinarySI}] Scopes:[] ScopeSelector:nil}"

// compensate (add more) the gap due to the de-facto that the VMI has used more memory  (new feature)
>time="2026-02-17T15:34:14Z" level=debug msg="resourcequota quota-test/default-cq7qs is updated from {Hard:map[limits.cpu:{i:{value:5 scale:0} d:{Dec:<nil>} s:5 Format:DecimalSI} limits.memory:{i:{value:12487836313 scale:0} d:{Dec:<nil>} s:12487836313 Format:DecimalSI}] Scopes:[] ScopeSelector:nil} to {Hard:map[limits.cpu:{i:{value:5 scale:0} d:{Dec:<nil>} s:5 Format:DecimalSI} limits.memory:{i:{value:14684130354 scale:0} d:{Dec:<nil>} s: Format:BinarySI}] Scopes:[] ScopeSelector:nil}"


Harvester compensate additional quota to let the VM finish the migration.

(7) Start migration, `abortMigrate` in the middle, or delete the `vmim` object in the middle, the UI shows the VM with correct status

(8) Stop this VM and start again, the webhook denies it, as a stopped VM can't escape the limitation of quota
>start virtual machine quota-test/vm4 failed, Internal error occurred: admission webhook "validator.harvesterhci.io" denied the request: memory insufficient resources due to resource quota

w13915984028:
@ibrokethecloud @ihcsim @Vicente-Cheng  @Yu-Jack @martindekov 

This PR is ready, please help review, thanks.

pkg/util/resourcequota/calculator.go
```diff
@@ -63,6 +63,27 @@ func (c *Calculator) VMPodsExist(namespace, vmName string) (bool, error) {
 	return true, nil
 }
 
+func (c *Calculator) getNamespaceResourceQuotaAndRQ(namespace string) (*v3.NamespaceResourceQuota, *corev1.ResourceQuota, error) {
```
martindekov: nit: can be changed to `getNamespaceRQAndDefaultRQ` to reflect what is actually being taken from where

pkg/util/resourcequota/calculator.go
```diff
@@ -63,6 +63,27 @@ func (c *Calculator) VMPodsExist(namespace, vmName string) (bool, error) {
...
+	nrq, err := c.getNamespaceResourceQuota(namespace)
```
martindekov: What is the difference between this `getNamespaceResourceQuota` and `getNamespaceDefaultResourceQuota`? I expected one quota per namespace where if you have multiple quotas, they will be named differently each, but the parameter is namespace only here so how we can have 2 resource quotas in same namespace?

pkg/util/resourcequota/calculator.go
```diff
@@ -131,13 +147,17 @@ func (c *Calculator) getNamespaceResourceQuota(vm *kubevirtv1.VirtualMachine) (*
 	return resourceQuota, nil
 }
 
-func (c *Calculator) getResourceQuota(vm *kubevirtv1.VirtualMachine) ([]*corev1.ResourceQuota, error) {
+// get the first default resourcequota from a given namespace
+func (c *Calculator) getNamespaceDefaultResourceQuota(namespace string) (*corev1.ResourceQuota, error) {
```
martindekov, line 152: If the selector differentiates the getNamespaceDefaultResourceQuota from getNamespaceResourceQuota, can we have two quotas with the selector:
```
util.LabelManagementDefaultResourceQuota: "true"
```

Otherwise if we expect only 1 per namespace, should we have a check `len(rqs) > 1` as well?

pkg/util/resourcequota/calculator.go
```diff
@@ -88,30 +109,25 @@ func (c *Calculator) CheckIfVMCanStartByResourceQuota(vm *kubevirtv1.VirtualMach
...
 		logrus.Debugf("CheckIfVMCanStartByResourceQuota: default resource quota is not found in the namespace %s, skip check", vm.Namespace)
 		return nil
 	}
 
-	return c.containsEnoughResourceQuotaToStartVM(vm, nrq, rqs[0])
+	return c.containsEnoughResourceQuotaToStartVM(vm, nrq, rq)
 }
 
-func (c *Calculator) getNamespaceResourceQuota(vm *kubevirtv1.VirtualMachine) (*v3.NamespaceResourceQuota, error) {
+func (c *Calculator) getNamespaceResourceQuota(namespace string) (*v3.NamespaceResourceQuota, error) {
```
martindekov: This method before returned single namespace quota if definitions below were fitting. Wile the other returned list of resource quotas in a virtual machine namespace, now seems like the only difference is that one is using selector and the other conditions, should we merge the methods in 1? 

Also what is the goal of this? To return valid default namespace resource quota? Can we even have multiple quotas per namespace?

pkg/util/resourcequota/calculator.go
```diff
@@ -480,3 +543,34 @@ func calculateVMStorageQuantity(vm *kubevirtv1.VirtualMachine) (resource.Quantit
...
+	// hard limit:
+	limitMem := rq.Spec.Hard.Name(corev1.ResourceLimitsMemory, resource.BinarySI)
+	// used
+	usedMem := rq.Status.Used.Name(corev1.ResourceLimitsMemory, resource.BinarySI)
+
+	usedMem.Add(mem)
+	// used + migration target pod exceeds limit, compensate the delta
+	if usedMem.Cmp(*limitMem) == 1 {
+		usedMem.Sub(*limitMem)
+		usedMem.Add(*resource.NewQuantity(additionalCompensationMemeory, resource.BinarySI))
```
martindekov: Fundamental question for the change if we always have that buffer - `additionalCompensationMemeory` to allow vm to run smoothly, can we just enable/disable the quotas instead of going all the way to recalculating and allowing overheads?

martindekov (commented):
I will most likely need couple of passes over the change as not familiar with the feature itself. Added some questions regarding the "default" and "namespaced" resource quotas after this change. Before we used the different methods where one listed RQ in a namespace and the other picked based on condition, this PR changes this fundamental difference.

Also my understanding is that we allow overhead of the quota to let migration happen, but if we always allow a buffer, wouldn't it be easier to make the feature toggle on/off as opposed to changing the feature itself to loose the restriction?

Note I am lacking overall context so my questions might be meaningless if I approach the problem from the wrong angle so apologies in advance for that

pkg/util/resourcequota/calculator.go
w13915984028: Add some backgrounds about the RQ management:

The RQ has such scenario:
e.g. a namespace has 5Gi memory limits, and a VM is booted with 4Gi user requests memory, it runs when overhead-ration is the default value 1.5,

1. when a new (additional) VM is booted, it is denied as no free quota.
2. when this VM is running, and overhead-ratio is changed from 1.5 to 10, and then migrate the vm, in existing code, it will be blocked after a few tries; new PR allows it to migrate anyway
3.  when this VM is running,  overhead-ratio is changed from 1.5 to 10 and this VM is stopped, then it can't boot again, in both existing code and new PR, as RQ is not enough for a cold booted VM

if we disable quota when vm is migrating, then it can't effectively deal case (1), and this is the main target to use RQ, control the resource usage all the time,  the VM migration, is a special case, it occupies a very small percentage of time on the whole lifecycle of a namespace/project, manuall toggle the switch to allow migration happen is not practical, hence we have the RQ auto-scaling feature to let migration be successful

customer might adjust the system setting overhead-ratio on the fly, which is used to eliminate VM OOM effectively, in this case, the strongly wanted feature is to let those already running VMs can still migrate

we did not find a better solution, compensate the RQ to meet above requirement looks to be the only solution we can offer now.

w13915984028:
@github-copilot review

pkg/controller/master/migration/vmi_controller.go
```diff
@@ -1,82 +1,43 @@
...
-			return vmi, err
-		}
-		if len(pods) > 0 {
-			if err := h.pods.Delete(vmi.Namespace, pods[0].Name, &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
-				return vmi, err
-			}
+			logrus.Infof("vmi %s/%s finished migration but fail to reset Harvester related state %s", vmi.Namespace, vmi.Name, err.Error())
+			return nil, err
 		}
 	}
```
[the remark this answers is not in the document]
w13915984028: This block of code was dead and not used anymore, hence remove it.

pkg/util/resourcequota/calculator.go
```diff
@@ -146,7 +166,8 @@ func (c *Calculator) containsEnoughResourceQuotaToStartVM(
 	namespaceResourceQuota *v3.NamespaceResourceQuota,
 	rq *corev1.ResourceQuota) error {
 	// get running migrations' used resource
-	vmimsCPU, vmimsMem, _, err := c.getRunningVMIMResources(rq)
+	// vmimsCPU, vmimsMem, _, err := c.getVMIMResourcesFromRQAnnotation(rq)
+	vmimsCPU, vmimsMem, _, err := GetVMIMResourcesAndCompensationFromRQAnnotation(rq)
```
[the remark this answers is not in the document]
w13915984028: I am double checking & testing this comments, will reply later.

pkg/controller/master/migration/vmi_controller.go
```diff
@@ -1,82 +1,43 @@
...
 func (h *Handler) OnVmiChanged(_ string, vmi *kubevirtv1.VirtualMachineInstance) (*kubevirtv1.VirtualMachineInstance, error) {
-	if vmi == nil || vmi.DeletionTimestamp != nil ||
-		vmi.Annotations == nil || vmi.Status.MigrationState == nil {
+	if vmi == nil || vmi.DeletionTimestamp != nil {
 		return vmi, nil
 	}
 
-	if vmi.Annotations[util.AnnotationMigrationUID] == string(vmi.Status.MigrationState.MigrationUID) &&
-		vmi.Status.MigrationState.Completed {
+	if IsVmiResetHarvesterMigrationAnnotationRequired(vmi) {
```
Vicente-Cheng: just want to confirm, this part is not related to the RQ enhancement, right? Just a small refactor for the current check mechanism?

pkg/controller/master/migration/vmi_controller.go
```diff
@@ -1,82 +1,43 @@
...
+		logrus.Debugf("vmi %s/%s finished migration, reset Harvester related state", vmi.Namespace, vmi.Name)
+		// note: this is a bit redundant with vmim controller, which runs below function when vmim is finished
 		if err := h.resetHarvesterMigrationStateInVmiAndSyncVM(vmi); err != nil {
-			logrus.Infof("vmi %s/%s finished migration but fail to reset state %s", vmi.Namespace, vmi.Name, err.Error())
-			return vmi, err
-		}
-	}
-
-	if vmi.Status.MigrationState.Completed && vmi.Status.MigrationState.AbortStatus == kubevirtv1.MigrationAbortSucceeded {
-		// clean up leftover pod on abortion success
```
Vicente-Cheng, line 49: I don't see that this cleanup was moved to another controller. Don't we need that anymore?

pkg/controller/master/migration/vmim_controller.go
```diff
@@ -110,137 +151,167 @@ func (h *Handler) syncVM(vmi *kubevirtv1.VirtualMachineInstance) error {
...
+	}
+	if rq == nil {
 		return nil
 	}
-
-	rqCpy := rqs[0].DeepCopy()
-	if ok := rqutils.ContainsMigratingVM(rqCpy, vmi.Name, string(vmi.UID)); ok {
-		logrus.Debugf("scaleResourceQuotaWithVMI: the resource quota in the namespace %s and vm %s is already scaled, skip updating", vmi.Namespace, vmi.Name)
+	// already included
+	if rqutils.ContainsMigratingVM(rq, string(vmi.UID)) {
```
Vicente-Cheng, line 164: The debug log above
```
logrus.Debugf("scaleResourceQuotaWithVMI: the resource quota in the namespace %s and vm %s is already scaled, skip updating", vmi.Namespace, vmi.Name)
```
Looks useful if we return early here. Could we add this back?

pkg/controller/master/migration/vmim_controller.go
```diff
@@ -110,137 +151,167 @@ func (h *Handler) syncVM(vmi *kubevirtv1.VirtualMachineInstance) error {
...
 		return nil
 	}
 
+	rqCpy := rq.DeepCopy()
 	// only update to ResourceQuota annotation, do not change ResourceQuota spec directly in this step
 	needUpdate, rqToUpdate, rl := rqutils.CalculateScaleResourceQuotaWithVMI(rqCpy, vmi, util.GetAdditionalGuestMemoryOverheadRatioWithoutError(h.settingCache))
 	if !needUpdate {
-		logrus.Debugf("scaleResourceQuotaWithVMI: no need to update resource quota, skip updating namespace %s and vm %s", vmi.Namespace, vmi.Name)
```
Vicente-Cheng, line 168: ditto

pkg/controller/master/migration/vmim_controller.go
```diff
@@ -110,137 +151,167 @@ func (h *Handler) syncVM(vmi *kubevirtv1.VirtualMachineInstance) error {
...
+
+	rqToUpdate := rq.DeepCopy()
+	// only update to ResourceQuota annotation, do not change ResourceQuota spec directly in this step
+	needUpdate, rqToUpdate, rl := rqutils.CalculateCompensationResourceQuotaWithVMI(rqToUpdate, vmi, util.GetAdditionalGuestMemoryOverheadRatioWithoutError(h.settingCache))
+	if !needUpdate {
+		logrus.Debugf("compensateResourceQuotaBase: no need to update resource quota, skip updating namespace %s and vm %s", vmi.Namespace, vmi.Name)
+		return nil
+	}
+
+	logrus.Debugf("compensateResourceQuotaBase: compensate resource quota %s in namespace %s for vm %s : %v", rq.Name, vmi.Namespace, vmi.Name, rl)
```
Vicente-Cheng: Could we move this to Infof? This log is worth showing to the end user.

pkg/util/resourcequota/calculator.go
```diff
@@ -63,6 +63,27 @@ func (c *Calculator) VMPodsExist(namespace, vmName string) (bool, error) {
...
+	nrq, err := c.getNamespaceResourceQuota(namespace)
+	if err != nil {
+		return nil, nil, err
+	}
+	if nrq == nil {
+		return nil, nil, nil
+	}
+
+	// get resource quota limits from ResourceQuota
+	rq, err := c.getNamespaceDefaultResourceQuota(namespace)
```
Vicente-Cheng, line 77: how about `rq`/`rqDefault`?

Do we have non-NS RQ here?
If no, maybe rq could be present nrq.

Just a style discussion, no strong opinion

pkg/util/resourcequota/calculator.go
```diff
@@ -209,28 +230,61 @@ func (c *Calculator) calculateVMActualOverhead(vm *kubevirtv1.VirtualMachine) *r
...
 	for _, rl := range vms {
 		cpu.Add(*rl.Name(corev1.ResourceLimitsCPU, resource.DecimalSI))
 		mem.Add(*rl.Name(corev1.ResourceLimitsMemory, resource.BinarySI))
 		storage.Add(*rl.Name(corev1.ResourceRequestsStorage, resource.BinarySI))
 	}
 
+	// from the compensation
+	cpu.Add(*compRl.Name(corev1.ResourceLimitsCPU, resource.DecimalSI))
+	mem.Add(*compRl.Name(corev1.ResourceLimitsMemory, resource.BinarySI))
+	storage.Add(*compRl.Name(corev1.ResourceRequestsStorage, resource.BinarySI))
```
Vicente-Cheng, line 256: I am not sure we need to add a storage resourceQuota. Looks like the migration only needs to take care of the CPU/Memory; we don't need extra storage during live migration.

Or did I miss any case?

pkg/util/resourcequota/calculator.go
```diff
@@ -258,25 +312,21 @@ func (c *Calculator) getVMPods(namespace, vmName string) ([]*corev1.Pod, error)
...
-		logrus.Debugf("CheckStorageResourceQuota: not found any default resource quota in the namespace %s", vm.Namespace)
+	if rq == nil {
+		logrus.Debugf("CheckStorageResourceQuota: not found any default resource quota in the namespace %s, skip check", vm.Namespace)
 		return nil
 	}
 
-	rq := rqs[0]
-
-	_, _, vmimsStorage, err := c.getRunningVMIMResources(rq)
+	// _, _, vmimsStorage, err := c.getVMIMResourcesFromRQAnnotation(rq)
```
Vicente-Cheng: redundant?

pkg/util/resourcequota/utils.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/harvester/harvester/pkg/util"
 )
 
+const (
+	// even kubevirt can't 100% precisely calculate the exact memory a vmi POD will consume
+	// we add additional 128Mi when compensate RQ, to ensure the vmi migration target pod can be created
+	additionalCompensationMemory = 128 << 20
```
Vicente-Cheng, line 16: > // even kubevirt can't 100% precisely calculate the exact memory a vmi POD will consume

as you said, how do we know the 128Mi is the reasonable magic number?

Vicente-Cheng (commented):
overall lgtm, just some questions

BTW, I suggest we can separate the refactor changes into another commits
That we can focus on the core mechanism improvement.
Thanks!

pkg/util/resourcequota/calculator.go
w13915984028: The function is renamed to `getNamespaceResourceQuota`, which calls `getRancherNamespaceResourceQuota` and `getNamespaceDefaultResourceQuota` to return those two objects for usage, as only when two objects are both existing, the following operaitons can be taken

It mainly aims to reuse the code, before the changes, the purposed code block occurs on several places.

pkg/util/resourcequota/calculator.go
w13915984028: The `getNamespaceResourceQuota` is now `getRancherNamespaceResourceQuota`, it means the `Rancher Manager` settled value on Harvester namespace, it is the source of true.

The `getNamespaceDefaultResourceQuota` returns the first `k8s ResourceQuota` object which has the label `LabelManagementDefaultResourceQuota`.

Per K8s definition, a namespace can have multi ResourceQuota objects, which is not checked/denied before, and this PR does not plan to change this point, just return the first one it case it exists.

The function has updated comments:
```
// Get the Rancher NamespaceResourceQuota (via namespace annotation) and the first K8s ResourceQuota of a given namespace
// They work together to influence the following actions
func (c *Calculator) getNamespaceResourceQuota(namespace string) (*v3.NamespaceResourceQuota, *corev1.ResourceQuota, error) {
```

pkg/util/resourcequota/calculator.go
w13915984028: Good catch. The existing code has this flaw in theory, it can't deal the case `len(rqs) > 1` well. The PR will not touch this check, instead will work on it via ticket https://github.com/harvester/harvester/issues/10185.

Also be careful of this note:
https://github.com/harvester/harvester/issues/10185#issuecomment-4034626943

pkg/util/resourcequota/utils.go
w13915984028: no magic, it is based on test and estimation, it might be updated on future versions, like kubevirt updates the vm overhead values version by version if necessary

https://github.com/kubevirt/kubevirt/blob/9d0345041704fe0d12a7cbb8eeecb2823d1cb703/pkg/virt-controller/services/template.go#L114-L123

pkg/controller/master/migration/vmi_controller.go
w13915984028: This block of code is dead, kubevirt does not have leftover pods any more.

The change is reverted to reduce change scope.

pkg/util/resourcequota/calculator.go
w13915984028: atm, `controller` only injects compensation on memory, but could add more in the future

for the `calculator`, it adds any potential compensation, does't assume if controller injects or not, not strictly tied with `controller`.

just keep L254 `mem.Add(*compRl.Name(corev1.ResourceLimitsMemory, resource.BinarySI))` is also ok

w13915984028:
@Vicente-Cheng Thanks for your reivew, I will check if & how to split the PR to reduce the change scope, thanks.

pkg/controller/master/migration/vmim_controller.go
```diff
@@ -25,49 +26,85 @@ const (
...
-	abortRequested := isAbortRequest(vmim)
+	// when a migration is aborted, the phase will transfer to kubevirtv1.MigrationFailed finally
+	abortRequested := isAbortRequested(vmim)
 	phase := vmim.Status.Phase
 	// debug log shows, after vmim was deleted, the OnChange here may be called two times
 	logrus.Debugf("syncing vmim %s/%s/%s phase %v abortRequested %v deleted %t", vmim.Namespace, vmim.Name, vmi.Name, phase, abortRequested, vmim.DeletionTimestamp != nil)
 
-	if phase == kubevirtv1.MigrationPhaseUnset && !abortRequested {
-		// set StatePending for UI status showing and menu rendering
+	switch phase {
```
ibrokethecloud, line 50: irrespective of the `phase` we always run `return vmim, h.setVmiMigrationUIDAnnotation(vmi, string(vmim.UID), StateAbortingMigration)` when `abortRequested` is true.

would it make sense to move it out of the `switch` block and deal with it earlier?

pkg/controller/master/migration/utils.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/harvester/harvester/pkg/util"
+)
+
+// when migration is done or aborted, vmi.Status.MigrationState has Completed or Failed
+func IsVmiResetHarvesterMigrationAnnotationRequired(vmi *kubevirtv1.VirtualMachineInstance) bool {
+	if vmi == nil || vmi.Status.MigrationState == nil {
+		return false
+	}
+	return (vmi.Status.MigrationState.Completed || vmi.Status.MigrationState.Failed) &&
+		vmi.Annotations[util.AnnotationMigrationUID] == string(vmi.Status.MigrationState.MigrationUID)
```
Yu-Jack: Could we make it more clearer or shorter? Like:

```go
done := vmi.Status.MigrationState.Completed || vmi.Status.MigrationState.Failed
foundMigration = vmi.Annotations[util.AnnotationMigrationUID] == string(vmi.Status.MigrationState.MigrationUID)

return done && foundMigration
```

or like:

```go
if !vmi.Status.MigrationState.Completed && !vmi.Status.MigrationState.Failed {
    return false
}

return vmi.Annotations[util.AnnotationMigrationUID] == string(vmi.Status.MigrationState.MigrationUID)
```

pkg/controller/master/migration/vmim_controller.go
```diff
@@ -110,137 +151,167 @@ func (h *Handler) syncVM(vmi *kubevirtv1.VirtualMachineInstance) error {
...
-
-	// restore ResourceQuota through vmi resource specifications
-	if err := h.restoreResourceQuotaWithVMI(vmi); err != nil {
-		return fmt.Errorf("failed to restore resource quota with vmi: %v", err)
+	if len(rqs) == 0 {
+		return nil, nil
 	}
-
-	return nil
+	return rqs[0], nil
```
Yu-Jack, line 194: if we always return the first one, should we return the error when length is greater than 1?

pkg/controller/master/migration/vmim_controller.go
```diff
@@ -110,137 +151,167 @@ func (h *Handler) syncVM(vmi *kubevirtv1.VirtualMachineInstance) error {
...
+		return nil
+	}
+	for _, cond := range vmim.Status.Conditions {
+		if cond.Type == kubevirtv1.VirtualMachineInstanceMigrationRejectedByResourceQuota {
+			// vmim is blocked due to quota, compensate in special cases
+			if cond.Status == corev1.ConditionTrue {
+				return h.compensateResourceQuotaBase(vmim, vmi)
+			}
+			return nil
+		}
```
Yu-Jack: A Nit:

```suggestion
		if cond.Type != kubevirtv1.VirtualMachineInstanceMigrationRejectedByResourceQuota {
			continue
		}
		// vmim is blocked due to quota, compensate in special cases
		if cond.Status != corev1.ConditionTrue {
			return nil
		}
		return h.compensateResourceQuotaBase(vmim, vmi)
```

pkg/controller/master/migration/vmim_controller.go
w13915984028: will double-check this via ticket https://github.com/harvester/harvester/issues/10185

for now, we keep as before, to avoid unexpected side-effects

pkg/controller/master/migration/vmim_controller.go
w13915984028: in the `switch case`, there are two excepions: phase `kubevirtv1.MigrationFailed, kubevirtv1.MigrationSucceeded` do not check `abortRequested` https://github.com/harvester/harvester/pull/9942/changes#diff-6e586a1e51666890526dd9f12edb8768cbb1ba98fff577fdce3d72d5f714f439R70  ;      

the `abortRequested` is judged via `vmim.Status.Conditions` not the phase

if we move `vmim, h.setVmiMigrationUIDAnnotation(vmi, string(vmim.UID), StateAbortingMigration)` ahead, we still need to check the phase and specially process the two exceptions

 the `switch case` ensures we cover all phases

pkg/controller/master/migration/utils.go
w13915984028: the function is moved to `migration/vmi_controller.go` and adopts the suggestion, thanks

 https://github.com/harvester/harvester/pull/9942/changes#diff-0d2a80ee7025e4fa38ec2083878ca6c8fe410fc3a2c235d709c3120fcdc813e1R50

pkg/util/resourcequota/calculator.go
w13915984028: removed.

pkg/controller/master/migration/vmim_controller.go
w13915984028: Updated to info.

pkg/controller/master/migration/vmim_controller.go
w13915984028: The vmim controller reconciller is normally triggered several times for each vmim, the below log

```
logrus.Debugf("scaleResourceQuotaWithVMI: update resource quota with vmi %s/%s", vmi.Namespace, vmi.Name)
```

is the pre-condition to cause `rqutils.ContainsMigratingVM(rq, string(vmi.UID))  to be true`, 

the logs  `...update resource quota with vmi` has implicitly include the removed log `is already scaled, skip updating`, from my test, `is already scaled, skip updating` is shown several times and is kind of noise.


```
	// already included
	if rqutils.ContainsMigratingVM(rq, string(vmi.UID)) {
		return nil
	}

	rqCpy := rq.DeepCopy()
	// only update to ResourceQuota annotation, do not change ResourceQuota spec directly in this step
	needUpdate, rqToUpdate, rl := rqutils.CalculateScaleResourceQuotaWithVMI(rqCpy, vmi, util.GetAdditionalGuestMemoryOverheadRatioWithoutError(h.settingCache))
	if !needUpdate {
		return nil
	}

	// add migrating vmi information to ResourceQuota
	if err := rqutils.AddMigratingVM(rqToUpdate, string(vmi.UID), rl); err != nil {
		return err
	}
	logrus.Debugf("scaleResourceQuotaWithVMI: update resource quota with vmi %s/%s", vmi.Namespace, vmi.Name)
	_, err = h.rqs.Update(rqToUpdate)
```

pkg/controller/master/migration/vmim_controller.go
w13915984028: same with this https://github.com/harvester/harvester/pull/9942#discussion_r2920731624

pkg/controller/master/migration/vmi_controller.go
w13915984028: I reverted those changes upon vmi_controller.go to reduce the change scope.

w13915984028:
FYI: The core changes are moved to PR https://github.com/harvester/harvester/pull/10197, which removes all refactors.  PR 10197 is not fully ready yet, when it is done, I will invite you to do a final review. thanks.

w13915984028:
Close this PR, it was replaced by https://github.com/harvester/harvester/pull/10197, thanks.

Issue #9954, opened 2026-01-27
[TASK] decrease the storage overcommit-config to 100%, cpu overcommit-config to 1000%


Vicente-Cheng:
### Task request

To align with the Longhorn default value and avoid the potential out-of-space issue, we should decrease the storage overcommit-config to 100% (original 200%) , cpu overcommit-config to 1000% (original 1600%)

w13915984028:
BTW: Document,release-notes are both needed to highlight this change, and also suggest user to adjust the upgraded cluster accordingly. thanks.

Vicente-Cheng:
Hi @starbops, 
please help verify this one

starbops:
Validated along with #10087. Please see the result at https://github.com/harvester/harvester/issues/10087#issuecomment-4116777586

Close as completed.

PR #9956, opened 2026-01-27
settings: decrease some overcommit ratios

Vicente-Cheng:
- decrease CPU to 1000%, storage to 100%
        - To align with default configuration of Longhorn, Kubevirt
        - add the upgrade part, we will keep the current running
          configuration (no matter default/new applied)

#### Problem:
We should align the storage overcommit with Longhorn's default settings

#### Solution:
Align the storage overcommit

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9954
https://github.com/harvester/harvester/issues/10087

#### Test plan:
1. Create Harvester cluster
2. Ensure the default storage overcommit was 100%
3. Ensure the default CPU overcommit was 1000%
4. Ensure the old settings `{"cpu":1600,"memory":150,"storage":200}` was applied as value

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -13,6 +13,44 @@ pre_upgrade_manifest() {
   fi
 }
 
+# Preserve the current overcommit-config value before upgrade.
+# The new version lowers the default storage overcommit from 200% to 100%.
+# For users who never customized the setting (value is empty), we pin the
+# current default so their effective configuration is not silently changed.
+preserve_overcommit_config() {
+  if [[ ! "$UPGRADE_PREVIOUS_VERSION" =~ ^v1\.7\.[0-9]$ ]]; then
+    echo "Skip migrate overcommit config if you are not upgrade from v1.7.x, current version: $UPGRADE_PREVIOUS_VERSION"
+    return
+  fi
+  local setting_name="overcommit-config"
+  local setting_json
+  setting_json=$(kubectl get settings.harvesterhci.io "$setting_name" -o json 2>/dev/null) || {
+    echo "Warning: failed to get $setting_name setting, skipping preservation"
+    return 0
+  }
+
+  local current_value
+  current_value=$(echo "$setting_json" | jq -r '.value // empty')
+
+  if [ -n "$current_value" ]; then
+    echo "overcommit-config already has a customized value, skipping preservation"
+    return 0
+  fi
+
+  local default_value
+  default_value=$(echo "$setting_json" | jq -r '.default // empty')
+
+  if [ -z "$default_value" ]; then
+    echo "Warning: overcommit-config has no default value, skipping preservation"
+    return 0
+  fi
+
+  echo "Preserving current overcommit-config default as explicit value: $default_value"
+  local patch_json
+  patch_json=$(jq -n --arg v "$default_value" '{"value": $v}')
+  kubectl patch settings.harvesterhci.io "$setting_name" --type merge -p "$patch_json"
+}
+
 wait_managed_chart() {
   namespace=$1
   name=$2
@@ -1376,6 +1414,7 @@ wait_repo
 detect_repo
 detect_upgrade
 pre_upgrade_manifest
+preserve_overcommit_config
 pause_all_charts
 skip_restart_rancher_system_agent
 upgrade_rancher
--- a/pkg/settings/settings.go
+++ b/pkg/settings/settings.go
@@ -43,7 +43,7 @@ var (
 	DefaultStorageClass                    = NewSetting(DefaultStorageClassSettingName, "longhorn")
 	HTTPProxy                              = NewSetting(HTTPProxySettingName, "{}")
 	VMForceResetPolicySet                  = NewSetting(VMForceResetPolicySettingName, InitVMForceResetPolicy())
-	OvercommitConfig                       = NewSetting(OvercommitConfigSettingName, `{"cpu":1600,"memory":150,"storage":200}`)
+	OvercommitConfig                       = NewSetting(OvercommitConfigSettingName, `{"cpu":1000,"memory":150,"storage":100}`)
 	VipPools                               = NewSetting(VipPoolsConfigSettingName, "")
 	AutoDiskProvisionPaths                 = NewSetting(AutoDiskProvisionPathsSettingName, "")
 	CSIDriverConfig                        = NewSetting(CSIDriverConfigSettingName, `{"driver.longhorn.io":{"volumeSnapshotClassName":"longhorn-snapshot","backupVolumeSnapshotClassName":"longhorn"}}`)
--- a/pkg/controller/master/setting/handler.go
+++ b/pkg/controller/master/setting/handler.go
@@ -36,8 +36,8 @@ var (
 	bootstrapSettings = []string{
 		settings.SSLCertificatesSettingName,
 		settings.KubeconfigDefaultTokenTTLMinutesSettingName,
-		// The Longhorn storage over-provisioning percentage is set to 100, whereas Harvester uses 200.
-		// This needs to be synchronized when Harvester starts.
+		// We need to sync overcommit config on bootstrap
+		// Now, storage overcommit was aligned with Longhorn setting,
 		settings.OvercommitConfigSettingName,
 		// always run this when Harvester POD starts
 		settings.AdditionalGuestMemoryOverheadRatioName,


pkg/settings/settings.go
```diff
@@ -42,7 +42,7 @@ var (
 	DefaultStorageClass                    = NewSetting(DefaultStorageClassSettingName, "longhorn")
 	HTTPProxy                              = NewSetting(HTTPProxySettingName, "{}")
 	VMForceResetPolicySet                  = NewSetting(VMForceResetPolicySettingName, InitVMForceResetPolicy())
-	OvercommitConfig                       = NewSetting(OvercommitConfigSettingName, `{"cpu":1600,"memory":150,"storage":200}`)
+	OvercommitConfig                       = NewSetting(OvercommitConfigSettingName, `{"cpu":1600,"memory":150,"storage":100}`)
```
w13915984028: what's the consideration on upgrade path, if original value is 200, then replace it to 100 ? and keep other values. thanks.

Vicente-Cheng: We don't decrease the storage overcommit automatically during the upgrade. It might introduce risks (some replicas might not be scheduled/rebuilt when we decrease the overcommit ratio), so it is only applied to the fresh installation user. Old users need to update themselves.

w13915984028 (dismissed):
LGTM, thanks.

Let's also add this change to release note and document, reminder user to review and update this config after upgrading to latest version if necessary, Harvester does not revert existing values automatically,.

starbops (commented):
As discussed with @Vicente-Cheng offline, we might need to handle the upgrade path carefully, as it will lower the default value once the new Harvester version goes live. Thanks @w13915984028 for raising the concern.

Vicente-Cheng:
Hi @w13915984028, @starbops 
updated for the upgrade part, please recheck
Thanks!

package/upgrade/upgrade_manifests.sh
```diff
@@ -13,6 +13,40 @@ pre_upgrade_manifest() {
...
+  local setting_json
+  setting_json=$(kubectl get settings.harvesterhci.io "$setting_name" -o json 2>/dev/null) || {
+    echo "Warning: failed to get $setting_name setting, skipping preservation"
+    return 0
+  }
+
+  local current_value
+  current_value=$(echo "$setting_json" | jq -r '.value // empty')
+
+  if [ -n "$current_value" ]; then
```
w13915984028, line 35: A question:

The Copilot seems to be correct.

```
by default, the `value` field is empty

kubectl get settings.harvesterhci overcommit-config -ojson > occ.yaml

harv41:/home/rancher # cat occ.yaml  | jq -r '.value // empty'
harv41:/home/rancher # cat occ.yaml  | jq -r '.value'
null
```

consider: 

we always update the `default` to let it align with v180 expected values, no matter `default` is not existing or with another value, as `default` belongs to Harvester

case 1: user has set it, then the `value` field is not empty, the `(customized)value` is used first; it is still respected on v180

case 2: user has set a customized value, and after upgrade to v180, user resets the config to `use default`, then the updated `default` is used, seemlessly like a newly installed cluster

case 3: user never set this, after upgrade, the updated new `default` is used

WDYT? thanks.

w13915984028: BTW:

I had an old PR https://github.com/harvester/harvester/pull/6358

it solved one thing:

If Harvester changes the `default`  of any existing setting, then Harvester controller/upgrade logic needs to take care of this change, to ensure `default` is synced on/after the upgrade

starbops: I believe the recent change is to address case 3.

> case 3: user never set this, after upgrade, the updated new default is used

With the preservation, the (old) default value will be written into the `value` field, and the `default` field will later be updated with the new default value when Harvester is upgraded. This is to eliminate the potential impact to clusters, as mentioned in https://github.com/harvester/harvester/pull/9956#discussion_r2732502403.

Vicente-Cheng: Thanks @starbops!

Hi @w13915984028,

Our current plans are below:
We don't want to decrease the overcommit ratio with the running cluster, so we will keep the 200% (previous configuration) after the upgrade if the user did not set it up.

If the user sets any value on this, we will respect their changes.

For the new overcommit ratio, it only applies to the newly installed.
Users need to change the overcommit ratio after they upgrade.

So, for the priority with running a cluster (which means the upgrade path)
The current value > old default value (200%)
We don't apply the 100% directly to prevent the case you also mentioned in this comment (https://github.com/harvester/harvester/pull/9956#discussion_r2732502403)

Thanks!

w13915984028:
@Vicente-Cheng @starbops 

I checked a local v171->v180 upgraded cluster, the `overcommit-config` was never touched on v171, after upgrade, the `default` value was still same as v171;  Harvester does not change the `default` by default.

If we just copy `default` to `value`, then it seems not necessary.

Or we copy `default` to `value` and then set the `default` as the new ones per v180.

https://github.com/harvester/harvester/pull/9956/changes#diff-8c5fdd3c29c4031d6332031bafd4f1c97820a492be51384b2b22689c6322d31fR44-R47


Please define the expected stories on upgrade path https://github.com/harvester/harvester/pull/9956#issue-3861027375, thanks.


```
...
$ kubectl get settings.harvesterhci overcommit-config -oyaml
default: '{"cpu":1600,"memory":150,"storage":200}'

$ kubectl get upgrade.harvesterhci.io -A -oyaml
apiVersion: v1
...
    imageID: harvester-system/hvst-upgrade-n7c9x
    nodeStatuses:
      harv41:
        state: Succeeded
    previousVersion: v1.7.1
    repoInfo: |
      release:
        harvester: d2ae63a3
        harvesterChart: 0.0.0-upg10094-d2ae63a3
        os: Harvester master
        kubernetes: v1.35.1+rke2r1
        rancher: v2.14.0-alpha5
        monitoringChart: 108.0.2+up77.9.1-rancher.11
    singleNode: harv41
```

Vicente-Cheng:
Hi @w13915984028, 

I do the upgrade test with this PR.
before upgrade:
```
# kubectl get settings.harvesterhci overcommit-config -oyaml
default: '{"cpu":1600,"memory":150,"storage":200}'
kind: Setting
metadata:
  annotations:
    harvesterhci.io/hash: d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f
  creationTimestamp: "2026-03-02T19:05:16Z"
  generation: 2
  name: overcommit-config
  resourceVersion: "6347"
  uid: ebafce64-8da0-401b-8204-4663a1f6252a
status:
  conditions:
  - lastUpdateTime: "2026-03-02T19:05:16Z"
    status: "False"
    type: configured
```

after upgrade:
```
default: '{"cpu":1600,"memory":150,"storage":100}'
kind: Setting
metadata:
  annotations:
    harvesterhci.io/hash: a8cce3883564fbca9a06cbe52bfdc857e0626ee07f6492350040c66c
  creationTimestamp: "2026-03-02T19:05:16Z"
  generation: 5
  name: overcommit-config
  resourceVersion: "331837"
  uid: ebafce64-8da0-401b-8204-4663a1f6252a
status:
  conditions:
  - lastUpdateTime: "2026-03-03T02:47:35Z"
    status: "True"
    type: configured
value: '{"cpu":1600,"memory":150,"storage":200}'
```

We can see we keep the previous default value `'{"cpu":1600,"memory":150,"storage":200}'` and the new default `'{"cpu":1600,"memory":150,"storage":100}'`.

I think that's what we want, right?
cc @starbops

package/upgrade/upgrade_manifests.sh
```diff
@@ -13,6 +13,40 @@ pre_upgrade_manifest() {
   fi
 }
 
+# Preserve the current overcommit-config value before upgrade.
+# The new version lowers the default storage overcommit from 200% to 100%.
+# For users who never customized the setting (value is empty), we pin the
+# current default so their effective configuration is not silently changed.
+preserve_overcommit_config() {
+  local setting_name="overcommit-config"
```
w13915984028, line 25: we could add a check before L21

```
  if [[ ! "$UPGRADE_PREVIOUS_VERSION" =~ ^v1\.7\.[0-9]$ ]]; then
    return
  fi
```

w13915984028 (dismissed):
LGTM, thanks.

BTW, please update the detailed test cases discussion on the review to the `Test plan`.

```
Test plan:
Create Harvester cluster
Ensure the default storage overcommit was 100% (before was 200%)
```

starbops (dismissed):
LGTM, thank you.

Vicente-Cheng:
Please don't merge this PR at this moment, we need to discuss if we want to adjust the CPU ratio also.

Vicente-Cheng:
Hi @w13915984028, @starbops, 

I updated this PR. The main change after the last review was
- also decrease the CPU overcommit ratio to 1000% (see https://github.com/harvester/harvester/issues/10087)
- add version checking before applying the old overcommit settings (only execute when upgrading from v1.7.x)

package/upgrade/upgrade_manifests.sh
```diff
@@ -13,6 +13,40 @@ pre_upgrade_manifest() {
...
+
+  if [ -z "$default_value" ]; then
+    echo "Warning: overcommit-config has no default value, skipping preservation"
+    return 0
+  fi
+
+  echo "Preserving current overcommit-config default as explicit value: $default_value"
+  local patch_json
+  patch_json=$(jq -n --arg v "$default_value" '{"value": $v}')
+  kubectl patch settings.harvesterhci.io "$setting_name" --type merge -p "$patch_json"
```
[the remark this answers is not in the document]
w13915984028: it makes sense to apply this suggestions, add ` || echo ...` to escape the error, even though many existing code did not do

w13915984028 (approved):
LGTM, thanks for the fix.

starbops (approved):
LGTM, thank you.

Vicente-Cheng:
@mergifyio backport v1.8

PR #9958, opened 2026-01-28
HEP for pod security standards setting [skip ci]

ibrokethecloud:
#### Problem:

HEP for introduction of new `cluster-pod-security-standard` setting harvester

#### Related Issue(s):

https://github.com/harvester/harvester/issues/8196

### DIFF
--- a/enhancements/20260128-pod-security-standards.md
+++ b/enhancements/20260128-pod-security-standards.md
@@ -1,0 +1,105 @@
+# Harvester PSS Support
+
+## Summary
+
+This enhancement allows Harvester cluster admins to control pod security standards at a cluster level, while allowing fine grained control to allow different pod security standards at per namespace level. By default this setting would be disabled. When enabled by the cluster administrator the `baseline` pod security standard will be applied to all namespaces, except a predefined list of harvester system component namespaces. In addition cluster admins can also apply `privileged` and `restricted` pod security standards on user namespaces or even `whitelist` namespaces which will allow namespace to be excluded from being managed by harvester.
+
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/8196
+
+## Motivation
+
+### Goals
+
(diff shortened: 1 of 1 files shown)


ihcsim (approved):
LGTM. We should add a documentation item to describe what users who have adopted the steps described in the [KB](https://harvesterhci.io/kb/2025/05/08/using-pod-security-standard) should/can do next.

I think for users who are satisfied with the KB, they can keep that configuration, or they can switch over to this new approach, or as discussed, use both to avoid having the webhook being part of the access vector to the user workloads.

WebberHuang1118:
As Harvester **v1.8** will support mounting **Secret/ConfigMap/ServiceAccount** into a VM via **virtio-fs**, and this may require certain node resources (e.g., FUSE), I discussed with @ibrokethecloud and validated the **PSS baseline** policy behavior for virtio-fs scenarios.

**Premise**

* The target namespace is enforced with the **baseline** policy:

```bash
kubectl label --overwrite ns <namespace> \
  pod-security.kubernetes.io/enforce=baseline \
  pod-security.kubernetes.io/enforce-version=latest
```

* A **privileged pod** cannot be created under PSS baseline. It is rejected with an error like:
  `Error from server (Forbidden): ... violates PodSecurity "baseline:latest": privileged ...`

Example manifest:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: privileged-pod-sample
spec:
  containers:
  - name: privileged-container
    image: busybox:stable
    command: ["sh", "-c", "while true; do sleep 3600; done"]
    securityContext:
      privileged: true
```

**Virtio-fs in VM**

* Following the [virtio-fs test plan](https://github.com/harvester/harvester/pull/9981#issue-3874400985), under **baseline** policy, a VM can still mount **Secret/ConfigMap/ServiceAccount** via virtio-fs successfully.

cc @ibrokethecloud @Vicente-Cheng

WebberHuang1118 (approved):
LGTM, thanks.

votdev (approved):
LGTM

pohanhuang (approved):
LGTM

Issue #9813, opened 2026-01-05
[BUG] Download image with wrong checksum will continue retry over the limited 3 times


albinsun:
### Describe the Bug

Retry attempted **6/3** failed due to error: failed to process sync file: the expected checksum
<img width="1567" height="772" alt="Image" src="https://github.com/user-attachments/assets/5932e156-0b6b-4c4f-b867-bb5491abad79" />

Retry attempted **52/3** failed due to error: failed to process sync file: the expected checksum
<img width="1567" height="772" alt="Image" src="https://github.com/user-attachments/assets/0cee49b9-e7c1-45ee-82ad-e35308c56b81" />

### To Reproduce

Create an image using URL with wrong checksum

### Expected Behavior

Failed to create with 3/3 fail attempts.

### Support Bundle for Troubleshooting

1. v1.7.1-dev-20260104: [supportbundle_6-3_3nodesWitness_2026-01-05T08-36-00Z.zip](https://github.com/user-attachments/files/24429784/supportbundle_6-3_3nodesWitness_2026-01-05T08-36-00Z.zip)

2. v1.7-25edb7f3-head: [supportbundle_52-3_4nodes_2026-01-05T08-41-05Z.zip](https://github.com/user-attachments/files/24429789/supportbundle_52-3_4nodes_2026-01-05T08-41-05Z.zip)

### Environment

- Harvester version:
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure: QEMU-KVM and Baremetal
- Rancher version:

m-ildefons:
## Validation Test

<details><summary> 🟢 Install Harvester v1.8.0-dev-20260201</summary></details>

<details><summary> 🟢 Create new VM Image with mismatched checksum</summary><img width="3836" height="2051" alt="Image" src="https://github.com/user-attachments/assets/2e03cbac-b3f2-4ecf-8d93-74c8c9ef4071" /></details>

<details><summary> 🟢 Wait for three download attempts to fail</summary><img width="3827" height="2046" alt="Image" src="https://github.com/user-attachments/assets/e3a07dae-17d8-4c08-b22c-45fd2b18b881" /></details>

<details><summary> 🟢 After three failed attempts, VM Image transitions to `failed` state and no further download attempts are made</summary><img width="3840" height="2132" alt="Image" src="https://github.com/user-attachments/assets/d69c12df-2764-47cf-af56-b7f04ac69f2a" /></details>

## Conclusion

[...]

PR #9961, opened 2026-01-28
fix: stop image download retries after max attempts

pohanhuang:
#### Problem:
- ISO Image download will exceed the retry limit, root cause is the backingImage controller does not check retry when reconciling.

#### Solution:
- Early return in backing operator when read retry limit

#### Related Issue(s):
#9813

#### Test plan:
1. create a iso image with fake checksum
  - url : https://cloud-images.ubuntu.com/jammy/current/jammy-server-cloudimg-amd64.img
  - checksum: a0f50d914d522368827119cdcf7225f9b31f32f8774ec2b255bfc167b410ba531dde73a609faa263c1b536d0c0ad510cd3e682c707a504d5c5972de1ece63f16
2. ensure it retry failed with no exceed retry max

[Expected result]
Screenshot 2026-01-28 at 4 58 24 PM

### DIFF
--- a/pkg/image/backingimage/backing_image_controller.go
+++ b/pkg/image/backingimage/backing_image_controller.go
@@ -53,7 +53,7 @@ func (h *backingImageHandler) OnChanged(_ string, bi *lhv1beta2.BackingImage) (*
 		return nil, err
 	}
 
-	if harvesterv1.ImageRetryLimitExceeded.IsTrue(vmi) {
+	if h.vmio.IsRetryLimitExceeded(vmi) {
 		return bi, nil
 	}
 
--- a/pkg/image/backingimage/backing_image_controller.go
+++ b/pkg/image/backingimage/backing_image_controller.go
@@ -52,6 +52,11 @@ func (h *backingImageHandler) OnChanged(_ string, bi *lhv1beta2.BackingImage) (*
 	} else if err != nil {
 		return nil, err
 	}
+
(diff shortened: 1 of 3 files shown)


pkg/image/backingimage/backing_image_controller.go
```diff
@@ -52,6 +52,11 @@ func (h *backingImageHandler) OnChanged(_ string, bi *lhv1beta2.BackingImage) (*
 	} else if err != nil {
 		return nil, err
 	}
+
+	if harvesterv1.ImageRetryLimitExceeded.IsTrue(vmi) {
+		return bi, nil
+	}
+
```
WebberHuang1118, line 59: Suggest wrapping this into a utility function under `pkg/image/common/operator.go`. There’s already a similar pattern there, for example:

```go
func (vmio *vmiOperator) IsImported(vmi *harvesterv1.VirtualMachineImage) bool {
	return harvesterv1.ImageImported.IsTrue(vmi)
}
```

WebberHuang1118 (commented):
Thanks for the enhancement—please take a look at the comment.

pkg/image/backingimage/backing_image_controller.go
pohanhuang: Thanks for the advice, just add it.
Please help review when you are available, thanks.

WebberHuang1118 (approved):
LGTM, thanks.

Issue #9328, opened 2025-10-17
[FEATURE] Remotely get Status of Harvester and Nodes Deployment and Readiness


martindekov:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

In: https://github.com/harvester/harvester/issues/5303 customer reported inability to check harvester and node status.

### Describe the solution you'd like

This issue is feature request for monitoring endpoint which will report the status similar to how the console reports the status explained in comment:
https://github.com/harvester/harvester/issues/5303#issuecomment-2177263705

### Describe alternatives you've considered

Documentation change explaining how to get the state through client calls is the current alternative

martindekov:
After some discussions we decided that we are going to add endpoint - `/v1/harvester/readyz` with logic which returns ready or not through status determined by what we documented in the attached document PR. The endpoint will be read only and authenticated through the init/join token.

m-ildefons:
https://github.com/harvester/harvester/issues/9327 needs to be revisited to document the new solution with the readyz path, especially how authentication is done and how the user can fetch the token before the installation has completed.

martindekov:
Good suggestion Moritz, I was thinking of using the master issue: https://github.com/harvester/harvester/issues/5303 as final documentation change and with it close the whole story.

lanfon72:
Verified this feature has been applied.

Test Information
----
* Environment: **Baremetal MegaRAC SP-X 3 nodes** and KVM local env
* Harvester Version: **v1.8.0-dev-20260301**
* **ui-source** Option: **Auto**

### Verify Steps
1. Check the test plan in https://github.com/harvester/harvester/pull/9963#issue-3865426882
    a. Test with first node created
    b. Test with new node joining
    c. Test with token's authentication
    d. Response form: `{ready: bool, timestamp: str, (optional) msg: str}`
2. To notice that the **TOKEN** mentioned in test the PR is configured during cluster installation, not generated by API server.

Issue #5303, opened 2024-03-06
[ENHANCEMENT] Remotely get Status of Harvester and Nodes Deployment and Readiness


Martin-Weiss:
**Is your enhancement related to a problem? Please describe.**
When deploying Harvester or additional nodes - the automation i.e. terraform needs to check for the "deployment complete" / "node ready", "harvester ready" before additional automation happens.

For this we need some sort of readiness check / deployment status / deployment complete that we can validate remotely.

**Describe the solution you'd like**
curl https://<harvester-vip>/status <- here we should get something like "deployed and available" or "not ready"
curl https://<node-ip>/status <- here we should get something like "deployed and available" or "not ready"

Maybe more status feedback would make sense as well.

**Describe alternatives you've considered**
ssh, RKE2 status, kubectl - but all of that need authentication / authorization and that might not be available, yet.

starbops:
We can check the readiness of the cluster by the following command:

```
curl -X GET "$APISERVER/readyz?verbose" --header "Authorization: Bearer $TOKEN" --insecure
```

But that's for the Kubernetes cluster itself, and it does not mean all of the Harvester components are ready (plus it also requires authentication).

The current checking criteria and logic implemented for the console dashboard: https://github.com/harvester/harvester-installer/blob/4adc9fea30a6e6f612b3e75fdd8f6eb37cb50c3d/pkg/console/dashboard_panels.go#L518-L538 (thanks @tserong for pointed out)

We can leverage that and expose a dedicated API endpoint for the readiness probe of the entire Harvester cluster. For now, users can run the following commands against the cluster to know its state:

```
# Check if the Harvester pods are ready
kubectl -n harvester-system get pods -l app.kubernetes.io/name=harvester -l app.kubernetes.io/component=apiserver --no-headers -o custom-columns="STATUS:status.conditions[?(@.type=='Ready')].status"

# Check if the Harvester webhook pods are ready
kubectl -n harvester-system get pods -l app.kubernetes.io/name=harvester -l app.kubernetes.io/component=webhook-server --no-headers -o custom-columns="STATUS:status.conditions[?(@.type=='Ready')].status"

# Check if the Rancher pods are ready
kubectl -n cattle-system get pods -l app=rancher --no-headers -o custom-columns="STATUS:status.conditions[?(@.type=='Ready')].status"

# Check if the API is ready
curl -fk https://<VIP>/version
```

Martin-Weiss:
[...]

PR #9963, opened 2026-01-28
feat: introduce readyz path

martindekov:
Introducing readyz endpoint in harvester to report status of the cluster. The path is protected through the Bearer token we use to join/init harvester. The definition of readiness is what we use in the documentation in the bash script in Troubleshooting/Installation section: Check the status of Harvester components.

The change introduces the following:
* readyz path for cluster readiness check
* authentication middleware function of the endpoint through init token
* bump go-common and use namespace constants in it

#### Problem:
We only have documentation on how to check harvester cluster readiness through script and kubernetes client:
https://docs.harvesterhci.io/v1.7/troubleshooting/index#check-the-status-of-harvester-components

#### Solution:
Provide user with readiness path which is secure through authentication using the init/join token

#### Related Issue(s):
https://github.com/harvester/harvester/issues/5303 - general issue
https://github.com/harvester/harvester/issues/9328 - concrete subtask

#### Test plan:

Unit tests covering the readiness logic and E2E Tests covering the auth/z described blow.

##### E2E Tests:

Using a script which calls the endpoint during operations in harvester:

<details>
<summary>Test 1 - Calling endpoint with init/join token and VIP configured during installation eventually reports ready true</summary>
<br>

1. Built the change and created first harvester node with it
2. During installation configured VIP and init/join token
3. Created a script to call the endpoint in cycle once installation starts in a loop:
```bash
#!/bin/bash

URL="https://192.168.122.164/v1/harvester/readyz"
TOKEN="uniquetoken123"
OUTFILE="X.txt"

while true; do
    TIMESTAMP=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
    echo "[$TIMESTAMP] REQUEST: curl -k -H \"Authorization: Bearer $TOKEN\" $URL" >> "$OUTFILE"
    RESPONSE=$(curl -k -H "Authorization: Bearer $TOKEN" "$URL" 2>&1)
    echo "[$TIMESTAMP] RESPONSE:" >> "$OUTFILE"
    echo "$RESPONSE" &>> "$OUTFILE"
    sleep 1
done
```
4. The output text file called `create_first_node.txt` eventually reported: `{"ready":true,...`
</details>

<details>
<summary>Test 2 - When operations in harvester being done - image upload/vm creation etc. readiness report is not interrupted</summary>
<br>

1. While running the script mentioned in `Test 1` above but with: `OUTFILE="add_image_create_vm.txt"`
2. Added vm image to the cluster
3. Created VM in the cluster
4. The output of calling the endpoint every second showed the cluster was: `{"ready":true...` all the time from logs in the `add_image_create_vm.txt`
5. This proves that some harvester operations involving pods like vm creation or image upload won't make the cluster report false readiness
</details>

<details>
<summary>Test 3 - Adding a new harvester node will report rke control plane not ready when the new node is added to the rke cluster</summary>
<br>

Simply adding a harvester node throughout the whole process involves short report of readiness to false for around half minute as shown in the logs during process of adding a node: `add_node_to_cluster.txt` again using the script in Test 1 but with ``OUTFILE="add_node_to_cluster.txt"``
</details>

<details>
<summary>Test 4 - Auth/z when token is either missing or wrong</summary>
<br>

```
mdekov@localhost:~/go/src/github.com/harvester/harvester/.idea/reallifetests> curl -k -H "Authorization: Bearer " https://192.168.122.226/v1/harvester/readyz
Bearer token requiredmdekov@localhost:~/go/src/github.com/harvester/harvester/.idea/reallifetests> curl -k -H "Authorization: Bearer wrongtoken" https://192.168.122.226/v1/harvester/readyz
Unauthorizedmdekov@localhost:~/go/src/github.com/harvester/harvester/.idea/reallifetests> curl -k  https://192.168.122.226/v1/harvester/readyz
Bearer token requiredmdekov@localhost:~/go/src/github.com/harvester/harvester/.idea/reallifetests> 
```
The above is the literal response in terminal, ordering it to make it more readable by hand:
```
mdekov@localhost:~/go/src/github.com/harvester/harvester/.idea/reallifetests> curl -k -H "Authorization: Bearer " https://192.168.122.226/v1/harvester/readyz
Bearer token required
mdekov@localhost:~/go/src/github.com/harvester/harvester/.idea/reallifetests> curl -k -H "Authorization: Bearer wrongtoken" https://192.168.122.226/v1/harvester/readyz
Unauthorized
mdekov@localhost:~/go/src/github.com/harvester/harvester/.idea/reallifetests> curl -k  https://192.168.122.226/v1/harvester/readyz
Bearer token required
mdekov@localhost:~/go/src/github.com/harvester/harvester/.idea/reallifetests> 
```
</details>

Files from the tests:
[create_first_node.txt](https://github.com/user-attachments/files/25259876/create_first_node.txt)
[add_image_create_vm.txt](https://github.com/user-attachments/files/25259875/add_image_create_vm.txt)
[add_node_to_cluster.txt](https://github.com/user-attachments/files/25259872/add_node_to_cluster.txt)

N/A


pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,183 @@
...
+		ready, msg := h.arePodsReady(pods)
+		if !ready {
+			return false, msg
+		}
+	}
+
+	return true, ""
+}
+
+func (h *ReadyzHandler) arePodsReady(pods []*corev1.Pod) (bool, string) {
```
w13915984028: only specific types of pod are needed, and when a deployment have multi-pods, as long as the leader pod is active, it is active in most cases

seems the related `service` is an alternative

the `https` endpoint ready is a bit different with console:

When user can get an http response from this endpoint, the alone-path `nginx pod, rancher pod, harvester pod  (leaders), k8s-api-servers, managedchart(implicitly)` are all ready.

martindekov: I see I was thinking that checking all pods might be too much, but as per the agreement I just translated this check to golang:
https://docs.harvesterhci.io/v1.7/troubleshooting/index/#check-the-status-of-harvester-components

In order to reduce the complexity and discussion around which components (pods) we should actually check. 

So instead of pods you say we can check services? If so what services you think would be critical in determining readiness ?

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,183 @@
...
+
+type RancherdConfig struct {
+	Token string `yaml:"token"`
+}
+
+func NewReadyzHandler(scaled *config.Scaled) *ReadyzHandler {
+	return &ReadyzHandler{
+		scaled: scaled,
+	}
+}
```
Yu-Jack: Nit:

We only use pod cache so we can just PodCache as one of the struct for better understanding like:

```go
type ReadyzHandler struct {
    podCache PodCache
}
```

pkg/api/readyz/readyz_handler.go
Yu-Jack: The readiness means Harvester is ready to go so we just need to focus on the important **entries**. For example, you can access the following service with API or checking POD status, it should be enough for the first version.

1. longhorn manager
2. kubevirt controller
3. system upgrade controller
4. harvester cluster repo
5. harvester load balancer
6. harvester network controller manager
7. harvester node disk manager
8. etc...

The namespace you used should be fine.

However, we don't forbid users from creating resources in those namespaces, including pod. So, there would be a chance to check user's pods. 

A possible way is to check specific pods instead of all pods under certain namespaces.

pkg/api/readyz/readyz_handler.go
martindekov: Original change used pod cache only, but now we use rkecluster object as well as pods and we might need more objects in the future if we change the definition of readiness so I decided to leave as, let me know if you insist on changing to specific caches and I can revisit

pkg/api/readyz/readyz_handler.go
martindekov: As per offline discussions we agreed that:
1. valid response means harvester is implicitly validated to be running
2. longhorn manager as core harvester component
3. kubevirt controller as core harvester component
4. rkecluster object condition is ready true and status ready true as well - k8s is running

All the above should still report ready true when new node is being added so content of the code is changed to that.

pkg/api/readyz/readyz_handler.go
martindekov: With the testing I replaced usage of all objects in scaled to podCache and representation of rke control plane cache: `	rkeCache generic.CacheInterface[*rkev1.RKEControlPlane]` and now the handler looks like:
```
type ReadyzHandler struct {
	podCache ctlcorev1.PodCache
	rkeCache generic.CacheInterface[*rkev1.RKEControlPlane]
}
```

This addresses this feedback so I will resolve the discussion

pkg/api/readyz/readyz_handler.go
martindekov: Latest change addresses official definition of readiness so resolving the discussion

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to read OEM config: %w", err)
+	}
+
+	var config OEMConfig
+	if err := yaml.Unmarshal(data, &config); err != nil {
+		return "", fmt.Errorf("failed to parse OEM config: %w", err)
+	}
+
+	content := findFileContent(config, "/etc/rancher/rancherd/config.yaml")
```
w13915984028: The `cluster-init` OEM file has this join token; 

The `rancherd` run only once.
```
cat /etc/rancher/rancherd/config.yaml

role: cluster-init
nodeName: harv31
token: "..."
```

for other mgmt/control node, I did not have an env to check if they also have this value,  as the harvester leader pod could be scheduled to them after a cluster has multi-nodes

please double check, thanks.

I am still checking the source code, to see which CRD object is used to carrry this token.

it is a bit tricky, this token is passed to rke2, and it could only be rotated from this old doc:
https://harvesterhci.io/kb/harvester_security_best_practices#cluster-token-rke2-token-rotation

each node has `/oem/harvester.config`, the `token` looks useful here.
```
cat /oem/harvester.config 
schemeversion: 1
serverurl: ""
token: <>
...
```

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,219 @@
...
+			return
+		}
+		hasher := hashers.GetHasher()
+		tokenHash, tokenInitErr = hasher.CreateHash(expectedToken)
+	})
+
+	if tokenInitErr != nil {
+		return false
+	}
+	hasher, err := hashers.GetHasherForHash(tokenHash)
```
w13915984028: need to update the test case about which hash algorithm is used by default; and how to generate the hash from the user side

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return ""
+}
+
+func (h *ReadyzHandler) clusterReady() (bool, string) {
+	rkeControlPlane, err := h.rkeCache.Get(
+		util.FleetLocalNamespaceName,
+		util.LocalClusterName)
+	if err != nil {
+		return false, "rkeControlPlane not found"
```
w13915984028, line 74: use `fmt.Sprintf("rkeControlPlane not found, error %s", err.Error())` to keep original error

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+	}
+	if !ready {
+		return false, "rkeControlPlane is not ready"
+	}
+
+	longhornManagerSelector := labels.SelectorFromSet(labels.Set(longhornTypes.GetManagerLabels()))
+	longhornPods, err := h.podCache.List(common.LonghornSystemNamespaceName, longhornManagerSelector)
+	if err != nil {
+		return false, "failed to check longhorn-manager pods"
```
w13915984028, line 90: ditto

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	if !hasAtLeastOneReadyPod(longhornPods) {
+		return false, "longhorn-manager pods not ready"
+	}
+
+	virtControllerSelector := labels.SelectorFromSet(labels.Set{kubevirtv1.AppLabel: "virt-controller"})
+	virtPods, err := h.podCache.List(common.HarvesterSystemNamespaceName, virtControllerSelector)
+	if err != nil {
+		return false, "failed to check virt-controller pods"
```
w13915984028, line 105: ditto

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,219 @@
...
+	if !h.validateToken(providedToken) {
+		return nil, apierror.NewAPIError(validation.PermissionDenied, "Invalid token")
+	}
+
+	ready, msg := h.clusterReady()
+	timestamp := time.Now().UTC().Format(time.RFC3339)
+
+	if !ready {
+		// Log detailed reason for debugging
+		logrus.Infof("Cluster not ready: %s", msg)
```
w13915984028: Does the final code tend to use `logrus.Debugf` here

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,219 @@
...
+		if err != nil {
+			tokenInitErr = err
+			logrus.Errorf("Failed to read token from OEM config: %v", err)
+			return
+		}
+		hasher := hashers.GetHasher()
+		tokenHash, tokenInitErr = hasher.CreateHash(expectedToken)
+	})
+
+	if tokenInitErr != nil {
```
w13915984028: Do we want to differentiate this init error with other errors? now only `log` can tell which step fails.

w13915984028 (commented):
Thanks for your big efforts. @martindekov 

Some questions:

pkg/api/readyz/readyz_handler.go
martindekov: So yeah I was hoping there is some in cluster resource which I can use opposed to mounting the config file on each pod in each node and also it seems like its present on each node as per the testing when scaling up the cluster auth/z worked as expected.

I can spin up env as well and look in the cluster where this token might be located/mounted and read what you suggested, thanks for the reference. Let me know if you find the CRD first.

pkg/api/readyz/readyz_handler.go
martindekov: During my testing I saw that sometimes the error contains internal info like pod name and we agreed that:
```
# happy path where everything is ready
$ curl -k https://<vip>/v1/harvester/readyz
{"ready":true, timestamp":"<utc-datetime-in-rfc3339>"}

# endpoint reachable but not all services are ready - we can't include any sensitive info or expose any service names in msg
$ curl -k https://<vip>/v1/harvester/readyz
{"ready":false, "msg":"foo bar", "timestamp":"<utc-datetime-in-rfc3339>"}

# hang until client-side timeout
$ curl -k https://<vip>/v1/harvester/readyz
```
This part:
```
# endpoint reachable but not all services are ready - we can't include any sensitive info or expose any service names in msg
```

when I wrote the original code with what you suggested I saw the sensitive info. Also the key in the JSON body response is `msg` not `error` for this reason. To avoid that alternative to returning error is logging the error internally and whoever has access to see logs it can look why exactly is that. 

Let me know if you think logging the full error is viable and returning brief message what is generally not ready.

pkg/api/readyz/readyz_handler.go
martindekov: This is the alternative I mentioned:
https://github.com/harvester/harvester/pull/9963#discussion_r2767690820

originally I logged every error in detail in addition to returning brief message. So we can use this to logrus.Error here to log more details.

pkg/api/readyz/readyz_handler.go
martindekov: I think this is a good idea, we can extract:
```
	tokenOnce.Do(func() {
		expectedToken, err := h.getTokenFromOEM()
		if err != nil {
			tokenInitErr = err
			logrus.Errorf("Failed to read token from OEM config: %v", err)
			return
		}
		hasher := hashers.GetHasher()
		tokenHash, tokenInitErr = hasher.CreateHash(expectedToken)
	})

	if tokenInitErr != nil {
		return false
	}
```

in a method like `loadToken` and do the validation in this method, that way we can differentiate between server and client side error

pkg/api/readyz/readyz_handler.go
w13915984028: The return response to external is a general message without details;

meanwhile, use a `logrus.Debugf` to print the error before HTTP return to external; thus when set Harvester log level to debug, we can know more details.

```
httpHanlder:
    processing()
    print log
    return general response
    
processing()
   return detailed error
```

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,219 @@
...
+	authHeader := req.Header.Get("Authorization")
+	if !strings.HasPrefix(authHeader, "Bearer ") {
+		return nil, apierror.NewAPIError(validation.Unauthorized, "Bearer token required")
+	}
+
+	providedToken := strings.TrimPrefix(authHeader, "Bearer ")
+	if !h.validateToken(providedToken) {
+		return nil, apierror.NewAPIError(validation.PermissionDenied, "Invalid token")
+	}
+
```
m-ildefons: If I read this correctly, you're doing an authentication check here. And my understanding of the PR is that this is necessary to allow authentication with the cluster registration token instead of requiring account setup first.

This is fine in concept, but I don't agree with the implementation here.

The implementation of the authentication mechanism should be separated from the implementation of the readyz handler.
The router supports "middlewares", which are essentially just composable handlers. See `recoveryMiddleware` at `pkg/server/router.go` line 118 for an example.

I think the authentication mechanism here should be implemented as middleware and applied at the route level to the `/v1/harvester/readyz` route.
The implementation of the readyz page handler should be independent from the authentication.
This provides a better separation of concerns, improves testability and makes the authentication reusable for other handlers.

Having such a reusable authentication middleware would make it possible to improve several aspects of cluster bootstrapping like resetting the admin password and retrieving the cluster kubeconfig.

w13915984028: The process `from 0 to 1` involves manythings, this PR does not get too much help from existing code, but from scratch;  the authentication could be optimized in next step to avoid this PR to be way complext. WDYT? @m-ildefons @martindekov , thanks.

martindekov: Hey Moritz, thanks for the review! I initially opened this as WIP, just for comments like that - how we handle auth/z/token read it from where/integrate the flow as part of what/etc. I don't insist on specific implementation/approach.

So I was not aware that the handler supports middleware I see we use gorilla mux and harvester server, but last time I wrote such code was in java where we did exactly what you suggested, but libraries/tools/approaches/expectations are different between languages/teams/tasks so if we take the middleware approach in this codebase as a whole, we can add it for auth/z. I have seen feedback "too complex for the purpose" so that's why I like previews/WIPs where such discussions are resolved before you commit to writing the tests and executing e2e with specific idea in mind.

I can switch gears but rewriting with this split along with unit and e2e testing will take some time. I will leave the discussion open for other folks to chime in and first address the previous feedback and then we can think of splitting it in middleware/handler. If you think this split is a must for the purposes, and insist - let me know I can spend time splitting it. For me the only cost is the added time for refactoring it.

martindekov: As addition to the comment above I can open a separate ticket to refactor towards the middleware approach and leave this discussion open post merge and use it as reference to the ticket.

I think your suggestion is a good one!

m-ildefons: Yeah, sorry I'm a bit late with this kind of feedback, since it is changing the fundamental structure of the PR's code. I've specifically not made this a blocking change request, because I don't think this it needs to be handled inside this PR.

If you think changing the authentication approach now would take too long, then opening a separate issue to fix this later is a reasonable approach.

martindekov: ok then will opt to open an issue with the suggestion and won't resolve the discussion for context when we pick it up :beers:

pkg/api/readyz/readyz_handler.go
martindekov: Added unit tests to show the default algorithm - its sha3. Will keep the discussion open in case you'd like to instantiate one of the others. The available structs are:
```
	case ScryptVersion:
		return ScryptHasher{}, nil
	case SHA256Version:
		return Sha256Hasher{}, nil
	case SHA3Version:
		return Sha3Hasher{}, nil
```

I can either continue using the default sha3 if you think it does the job, or move to any of the above

pkg/api/readyz/readyz_handler.go
martindekov: So I tried to mostly do that^. The exception is in clusterReady, where I found it convenient for the error to be logged in place and method be reused to create adequate brief message version of it. So it's mostly like the above with this exception, let me know if it makes sense.

pkg/api/readyz/readyz_handler.go
martindekov: debug added in place

pkg/api/readyz/readyz_handler.go
martindekov: debug added in place

pkg/api/readyz/readyz_handler.go
martindekov: now the thing I mentioned above is extracted in loadToken and errors are two failed to initialize and failed to validate

pkg/api/readyz/readyz_handler.go
martindekov: Did not add test case about which hash algorithm is used by default as I instantiated `Sha256Hasher` by default.

pkg/api/readyz/readyz_handler.go
martindekov: Changed to debugf

pkg/api/readyz/readyz_handler.go
martindekov: Latest commit addresses those concerns. Now I read the token from secret, not from file. I found the right secret to be: `local-rke-state` in the `fleet-local` namespace. Unit tests added and upgraded with the approach and e2e testing sections were upgraded with the e2e testing as well

martindekov: And a note through kubectl the secret is base64 encoded, but called through API it comes as string so no additional decoding needed. I did decode and through debugging found the format was not encoded unlike when querying through kubectl which was unexpected, but that's the reason the change is only 
Get -> save
and not 
Get -> base64 decode -> save

pkg/api/readyz/readyz_handler.go
martindekov: @m-ildefons @w13915984028 the issue I opened captures 2 points we discussed offline/here:
https://github.com/harvester/harvester/issues/10019

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,194 @@
...
+	ctx.SetStatusOK()
+	return map[string]any{
+		"ready":     true,
+		"timestamp": timestamp,
+	}, nil
+}
+
+func (h *ReadyzHandler) loadToken() error {
+	h.tokenOnce.Do(func() {
```
w13915984028: when cluster is first installed & initialized or some other tricky cases, the token might come later than the first time of readyz query, and if `tokenInitErr` is there, we don't retry, and there is no secret controller to update this `tokenInitErr`

did you consider to retry each time until once it is picked?

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,194 @@
...
+
+	if len(tokenBytes) == 0 {
+		return "", fmt.Errorf("token is empty in secret")
+	}
+
+	return string(tokenBytes), nil
+}
+
+func (h *ReadyzHandler) clusterReady() (bool, string) {
+	rkeControlPlane, err := h.rkeCache.Get(
```
w13915984028, line 69: could consider to split rkeControlPlane, LH, kubevirt checks into 3 small functions,  it will make `clusterReady` easier to extend and read

w13915984028 (dismissed):
LGTM, thanks.

pkg/api/readyz/readyz_handler.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (h *ReadyzHandler) loadToken() error {
+	h.tokenOnce.Do(func() {
+		expectedToken, err := h.getTokenFromSecret()
+		if err != nil {
+			h.tokenInitErr = err
+			return
+		}
+		hasher := hashers.GetHasher()
+		h.tokenHash, h.tokenInitErr = hasher.CreateHash(expectedToken)
```
ihcsim: I am not comfortable with the idea of storing secret in memory, in a long running process. It might leak into crash dump. I feel this is redundant because the handler is already fetching the secret from its in-memory cache.

pkg/api/readyz/readyz_handler.go
ihcsim: Separating out the auth logic makes sense and it will help with some upcoming changes. The middleware approach requires us to hard code the list of authenticated vs non-authenticated paths. We can consider avoiding that by wrapping the authentication logic into a wrapper function that takes and returns an `http.Handler`. Then we only need to call the wrapper function on protected paths:

```diff
diff --git a/pkg/server/router.go b/pkg/server/router.go
index de577dc05..72ac6347b 100644
--- a/pkg/server/router.go
+++ b/pkg/server/router.go
@@ -67,7 +67,7 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
 		r.scaled.CoreFactory.Core().V1().Pod().Cache(),
 		r.scaled.CoreFactory.Core().V1().Secret().Cache(),
 		r.scaled.Management.RKEFactory.Rke().V1().RKEControlPlane().Cache()))
-	m.Path("/v1/harvester/readyz").Methods("GET").Handler(readyzHandlerv1)
+	m.Path("/v1/harvester/readyz").Methods("GET").Handler(auth(readyzHandlerv1))
 
 	// --- END of preposition routes ---
 
@@ -123,6 +123,11 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
 	return m
 }
 
+func auth(next http.Handler) http.Handler {
+	// do authn here then return next handler
+	return next
+}
+
 func recoveryMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
```

pkg/api/readyz/readyz_handler.go
martindekov: >did you consider to retry each time until once it is picked?

So I didn't consider retry, I thought naturally if part of automation this secret existing is part of the definition of ready. I was thinking that automation cares of waiting until ready:true. When secret does not exist probably other parts of the cluster won't work as well so I didn't want to block the http call on it by retrying

martindekov: Also that's why I performed the testing above where in a loop I query the endpoint - to see how the response evolves with time and progress of readiness of harvester

pkg/api/readyz/readyz_handler.go
martindekov: Good point can refactor

pkg/api/readyz/readyz_handler.go
martindekov: removed the whole loadToken - we don't need singleton also we don't read secret from file so hashing and singleton are redundant given the complexity is handled by the secretCache provided by k8s. Also the cache is taking care of watching updates over the CRD. It also has the secret encoded and internal decodes it for us and provides it as string. In validation however I compared with constant time to still provide security when comparing strings, not using basic string comparioson.

pkg/api/readyz/readyz_handler.go
martindekov: This now calls 3 separate methods which in turn define longhorn/rke/kubevirt readiness. Very good suggestion, things look simpler and more intuitive now.

pkg/api/readyz/readyz_handler.go
martindekov: Since we take the secret through cache, remove the whole hasing. k8s takes care of storing the secret encoded and decodes it through the API call so we received as as plain text. The complexity of handling the lifecycle/security of the secret is already handled internally we just use the API. As explained above I still compared bytes with constant time, as opposed to basic string comparison.

martindekov: Will resolve the discussion as I addressed this

pkg/api/readyz/readyz_handler.go
martindekov: Due to all the simplifications from the reviews - getting rid of file, getting rid of managing the file life-cycle in memory through singleton, getting rid of hashing all due to k8s handling the life-cycle and integrity of the secret internally moving to the suggestion above was easy. Also moved the testing. Still I think the issue I opened should be kept. Having auth function as middleware in the router is still robust and can be improved. Also using the init/join token for authentication provides only authentication but no possibility for authorization. I considered the above suggestion by @ihcsim and implemented it, but this approach is still robust IMHO.

m-ildefons (dismissed):

w13915984028 (dismissed):
LGTM, thanks.

ihcsim (dismissed):

w13915984028 (approved):
LGTM, thanks.

Issue #9964, opened 2026-01-28
[BUG] upgrade stuck due to the mismatch instance-manager checksum


Vicente-Cheng:
### Describe the Bug

We will calculate the instance-manager checksum to match the label `longhorn.io/instance-manager-image`.
The value format would be `imi-sha512{0:8}`.

The issue was that we use the wrong image path to calculate.
The current image path in instance-manager after v1.10.2 would include the `docker.io` prefix.
But in the upgrade script, the image path is still w/o `docker.io`

That means we will get the wrong image checksum, so we are still stuck waiting the instance-manager to check like below:
```
hvst-upgrade-q8vxz-apply-manifests-pwhf6 instance-manager (aio)(v1) (image=longhornio/longhorn-instance-manager:v1.10.2-rc1) count is not 1 on node hp-129-tink-system, will retry...
hvst-upgrade-q8vxz-apply-manifests-pwhf6 instance-manager (aio)(v1) (image=longhornio/longhorn-instance-manager:v1.10.2-rc1) count is not 1 on node hp-129-tink-system, will retry...
hvst-upgrade-q8vxz-apply-manifests-pwhf6 instance-manager (aio)(v1) (image=longhornio/longhorn-instance-manager:v1.10.2-rc1) count is not 1 on node hp-129-tink-system, will retry...
hvst-upgrade-q8vxz-apply-manifests-pwhf6 instance-manager (aio)(v1) (image=longhornio/longhorn-instance-manager:v1.10.2-rc1) count is not 1 on node hp-129-tink-system, will retry...
hvst-upgrade-q8vxz-apply-manifests-pwhf6 instance-manager (aio)(v1) (image=longhornio/longhorn-instance-manager:v1.10.2-rc1) count is not 1 on node hp-129-tink-system, will retry...
hvst-upgrade-q8vxz-apply-manifests-pwhf6 instance-manager (aio)(v1) (image=longhornio/longhorn-instance-manager:v1.10.2-rc1) count is not 1 on node hp-129-tink-system, will retry...
hvst-upgrade-q8vxz-apply-manifests-pwhf6 instance-manager (aio)(v1) (image=longhornio/longhorn-instance-manager:v1.10.2-rc1) count is not 1 on node hp-129-tink-system, will retry...
```

### To Reproduce

1. create v1.6.1 cluster with at least 2 nodes
2. upgrade to v1.7.1-rc2
3. upgrade will be stuck with the above logs

### Expected Behavior

[...]

PR #9966, opened 2026-01-28
upgrade: enhance the registry value fetching

Vicente-Cheng:
- Now we did not hard-coded this value

#### Problem:
We use the wrong image path to calculate the checksum

#### Solution:
enhance the registry value fetching

#### Related Issue(s):
#9964 

#### Test plan:
1. create v1.6.1 cluster
2. upgrade with ISO including this PR
3. upgrade w/o issues


starbops (approved):
LGTM, thanks for the fix!

Vicente-Cheng:
tested with the upgrade path, see below:
1. create 2 node cluster with one running VM
2. upgrade to this commit
3. everything looks well

Vicente-Cheng:
@mergifyio backport v1.7

PR #9980, opened 2026-01-30
chore: fix deprecated api

pohanhuang:
#### Problem:

- The current `.golangci.yaml` excludes Staticcheck [SA1019](https://staticcheck.dev/docs/checks#SA1019) (deprecated API usage), so CI won’t flag deprecated API calls, and we may keep using APIs after they are deprecated.

#### Solution:

- Remove the exclusion for SA1019 in .golangci.yaml.

#### Expected result:
- CI will fail when code uses deprecated APIs, ensuring we catch and remove deprecated API calls during development.

#### Related Issue(s):

#9982

#### Test plan:
- make validation, ensure no deprecate api comes up.
- manually run the vm restore / replace, ensure it works as it is. 

- Found our current golangci-linter is not correct for v2, so I make some changes to adapt the current v2

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,7 +1,7 @@
 FROM quay.io/costoolkit/releases-teal:grub2-live-0.0.4-2 AS grub2-mbr
 FROM quay.io/costoolkit/releases-teal:grub2-efi-image-live-0.0.4-2 AS grub2-efi
 
-FROM golang:1.25-bookworm
+FROM golang:1.25.6-bookworm
 
 ARG DAPPER_HOST_ARCH
 ENV ARCH=$DAPPER_HOST_ARCH
@@ -45,7 +45,7 @@ RUN apt-get update -qq && apt-get install -y --no-install-recommends \
     docker-ce=5:20.10.* \
     && rm -rf /var/lib/apt/lists/*
 ## install golangci
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s latest
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.8.0
 
 ## install controller-gen
 RUN GO111MODULE=on go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.17.1
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/harvester/harvester
 
-go 1.25.6
+go 1.25
 
 replace (
 	// k8s.io/code-generator still remains v0.31.5 due to kubevirt v1.6.0.
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/harvester/harvester
 
-go 1.25
+go 1.25.6
 
 replace (
 	// k8s.io/code-generator still remains v0.31.5 due to kubevirt v1.6.0.
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -47,8 +47,7 @@ linters:
         text: "QF1012:" # Use fmt.Fprint(...) instead of Write([]byte(fmt.Sprint(...)))
 formatters:
   enable:
-    - goimports # checks if the code and import statements are formatted according to the 'goimports' command
-    - golines # checks if code is formatted, and fixes long lines
+    - gofmt
 run:
   timeout: 5m
   tests: true
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -47,9 +47,6 @@ linters:
         text: "QF1008:" # could remove embedded field from selector (staticcheck)
       - linters:
           - staticcheck
-        text: "SA1019:" # function is deprecated (staticcheck)
-      - linters:
-          - staticcheck
         text: "QF1012:" # Use fmt.Fprint(...) instead of Write([]byte(fmt.Sprint(...)))
 run:
   timeout: 5m
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,6 +1,6 @@
-version: 2
+version: "2"
 linters:
-  disable-all: true
+  default: none
   enable:
     - errcheck
     - govet
@@ -10,9 +10,6 @@ linters:
     - unused
     - misspell
     - prealloc
-  formatters:
-    enable:
-      - gofmt
   settings:
     gosec:
       # https://github.com/securego/gosec#available-rules
@@ -23,13 +20,13 @@ linters:
         - G402 # TLS InsecureSkipVerify set true
         - G505 # Blocklisted import crypto/md5: weak cryptographic primitive
       config:
-        G306: "0644" # Poor file permissions used when writing to a new file 
-  exclude-files:
-    - /zz_generated_
-    - _generated
-  exclude-dirs:
-    - generated
+        G306: "0644" # Poor file permissions used when writing to a new file
   exclusions:
+    paths:
+      - vendor
+      - pkg/generated
+      - .*zz_generated.*
+      - .*_generated.*
     presets:
       - comments
       - common-false-positives
@@ -48,6 +45,10 @@ linters:
       - linters:
           - staticcheck
         text: "QF1012:" # Use fmt.Fprint(...) instead of Write([]byte(fmt.Sprint(...)))
+formatters:
+  enable:
+    - goimports # checks if the code and import statements are formatted according to the 'goimports' command
+    - golines # checks if code is formatted, and fixes long lines
 run:
   timeout: 5m
   tests: true
--- a/pkg/api/supportbundle/download_handler.go
+++ b/pkg/api/supportbundle/download_handler.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"strconv"
 	"time"
@@ -87,7 +86,7 @@ func (h *DownloadHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
 
 	if resp.StatusCode != http.StatusOK {
 		msg := fmt.Sprintf("Unexpected status code %d from manager.", resp.StatusCode)
-		body, err := ioutil.ReadAll(resp.Body)
+		body, err := io.ReadAll(resp.Body)
 		if err == nil {
 			var errResp harvesterv1.ErrorResponse
 			if e := json.Unmarshal(body, &errResp); e == nil {
--- a/pkg/controller/admission/wait.go
+++ b/pkg/controller/admission/wait.go
@@ -20,7 +20,7 @@ const (
 
 // Wait waits for the admission webhook server to register ValidatingWebhookConfiguration and MutatingWebhookConfiguration resources.
 func Wait(ctx context.Context, clientSet *kubernetes.Clientset) error {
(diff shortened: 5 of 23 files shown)


pohanhuang:
@Vicente-Cheng This PR enables CI to detect deprecated API usage automatically, so we can address it proactively.

pohanhuang:
@WebberHuang1118 I've test with restore and replace with backup with https://docs.harvesterhci.io/v1.8/vm/backup-restore/ using private build, and it works fine.

Yu-Jack (dismissed):
LGTM.

wheatdog:
Thanks for looking into this!

To make sure that there is no surprise from the separate CI runs of the same commit (or commit with almost identical content), I think we should pin these:
1. golang version -> we should also pin the patch version of golang in [Dockerfile.dapper](https://github.com/harvester/harvester/blob/4fd67dddf44c37fa0297025000aa47781351b838/Dockerfile.dapper#L4C6-L4C26), like `FROM golang:1.25.6-bookworm`
2. all the dependencies -> it should already be satisfied thanks to go.mod/go.sum
3. golangci-lint itself -> we should pin the version in [Dockerfile.dapper](https://github.com/harvester/harvester/blob/4fd67dddf44c37fa0297025000aa47781351b838/Dockerfile.dapper#L48), it always installs the latest version now.

This should make the linter result more deterministic. I would like to prevent the frustration of "CI somehow suddenly failed on something I don't touch".

Instead, we could review those linter errors while bumping golang/dependencies/golangci-lint.

WDYT?

pohanhuang:
Hi @wheatdog,
In the latest commit, I pinned the Go and golangci-lint versions in the Docker build file and go.mod to ensure consistency.

Please help review when you are available, thanks.

pohanhuang:
@wheatdog @Yu-Jack Can you help review when you are available?
Thanks

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/harvester/harvester
 
-go 1.25
+go 1.25.6
```
wheatdog: I'm not sure whether this would cause issue to other repos like [node-disk-manager](https://github.com/harvester/node-disk-manager/blob/master/go.mod#L41). [vm-import-controller](https://github.com/harvester/vm-import-controller/blob/main/go.mod#L10), etc..

for their go.mod files, would they need to also include patch version in their go directive?
https://github.com/harvester/node-disk-manager/blob/master/go.mod#L3

https://github.com/search?q=org%3Aharvester+%22github.com%2Fharvester%2Fharvester%22+path%3Ago.mod&type=code

wheatdog: I'm thinking maybe we could leave this as is. Change the go image in Docker.dapper could be enough.

pohanhuang: got it, will revert it.

pohanhuang:
Hi @wheatdog , I just revert the changes on go.mod, please take a look when you are available, thanks.

wheatdog (dismissed):
Thanks! Let's wait for a few more reviews before merging.

pohanhuang:
[@harvester/dev](https://github.com/orgs/harvester/teams/dev) could you help me review this short PR when you are available? I would like to have a few more reviews before merging. Thanks!

w13915984028 (dismissed):
LGTM, thanks.

pohanhuang:
Hi @WebberHuang1118 @harvester/maintainer 
could you help me review this short PR when you are available? I would like to have a few more reviews before merging. Thanks!

pohanhuang:
@Vicente-Cheng can you help review this PR when you are available? thanks

ihcsim:
Will the linter failed our CI if we have dependencies that violate the check? KubeVirt (one of our main dependency) [excludes it for some cases](https://github.com/kubevirt/kubevirt/blob/69ce3d068509f719a3c4c450bc63c92346ff0a56/hack/linter/.golangci.yml#L100-L103). Is it possible to make this a warning, instead of failing CI? From what I can see in other projects, this particular check isn't always mandatory; some projects enforce it, others exclude it. 

FWIW, I think checks like this is best enforced in the IDE level. My IDE, with the default `gopls` settings, screams at me if I tried to use e.g., the `io/ioutil` package.

pohanhuang:
Hi @ihcsim ,
Yes — with the current setup, this check will fail CI if dependencies (or our code) violate it.

That said, the intent of the current implementation is simply to ensure the static check runs and is enforced, so we don’t accidentally introduce deprecated API usage (e.g., io/ioutil).

If we want to treat it as warning-only, we could adjust the CI configuration (or exclude specific cases like KubeVirt does), but we’d need to decide whether we still want to keep it as a required gate.

Based on the current changes, I believe the changes are needed, so I think we can keep it in this way.

Dockerfile.dapper
```diff
@@ -45,7 +45,7 @@ RUN apt-get update -qq && apt-get install -y --no-install-recommends \
     docker-ce=5:20.10.* \
     && rm -rf /var/lib/apt/lists/*
 ## install golangci
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s latest
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.8.0
```
Vicente-Cheng, line 48: actually we changed the version to `latest` recently (maybe last year)
because sometimes we would not remember to follow the newer linter version.

Do we have any concern to use the latest?

cc @innobead

Dockerfile.dapper
```diff
@@ -1,7 +1,7 @@
 FROM quay.io/costoolkit/releases-teal:grub2-live-0.0.4-2 AS grub2-mbr
 FROM quay.io/costoolkit/releases-teal:grub2-efi-image-live-0.0.4-2 AS grub2-efi
 
-FROM golang:1.25-bookworm
+FROM golang:1.25.6-bookworm
```
Vicente-Cheng, line 4: iiuc, the 1.25-bookworm should use golang 1.25.6 right?
If we pin to a specific version now, we might not use the latest version if we forget to bump.

Could we have the more specific reason to pin to a specific Go version?
cc @wheatdog

ihcsim:
> Yes — with the current setup, this check will fail CI if dependencies (or our code) violate it.

I don't  think we should block CI because of dependencies' check violations. We might not even be able to fix them. Can we just exclude the `vendor` folder like what Rancher did [here](https://github.com/rancher/rancher/blob/5460540d149d5ef14138f93bce66b093985d94da/.golangci.json#L77)?

pohanhuang:
Hi @harvester/dev,
Please review this PR when you have time.
Summary of changes:

- Fixed golangci-lint config - Our config has been broken for a while. I've updated it to follow the [v2 configuration format](https://golangci-lint.run/docs/configuration/file/#formatters-configuration).
- Excluded vendor and generated folders - CI shouldn't fail due to lint violations in dependencies or generated code that we can't fix. This follows the same approach [Rancher uses](https://github.com/rancher/rancher/blob/5460540d149d5ef14138f93bce66b093985d94da/.golangci.json#L77).
- Suppressed deprecated API warnings - Excluded SA1019 (deprecated API usage) for vendor and generated files, since we don't control those.

Let me know if you have any questions or concerns.

pohanhuang:
[CodeFactor](https://github.com/harvester/harvester/pull/9980/checks?check_run_id=62571581894) alert can be ignored, because the linter changes trigger the check, and those //TODO should keep it as it was.

Dockerfile.dapper
wheatdog: My 2 cents, as mentioned [here](https://github.com/harvester/harvester/pull/9980#issuecomment-3822980249), the concern is the frustration of "CI somehow suddenly failed on something we are not currently working on".

While the `latest` is used, a new version of golangci-lint could be adopted silently and potentially break the CI. It affects all the WIP branches / PRs and force us to context switch and fix the linter errors immediately. Instead, I feel like it would be better to handle those linter errors when explicitly bumping golangci-lint, which could be periodically initiated by Renovate (like [this post did](https://www.jvt.me/posts/2022/12/15/renovate-golangci-lint/)).

Dockerfile.dapper
wheatdog: The core idea here is similar, to prevent surprises from the silent golang bump, that might cause unexpected CI failure. I think Renovate could also help us bump image automatically like [this](https://docs.renovatebot.com/docker/).

w13915984028 (dismissed):
LGTM, thanks for your big efforts.

Dockerfile.dapper
pohanhuang: Hi @Vicente-Cheng @innobead,

What do you think about pinning the version here?

I think using a specific version would be better (like sbomscanner does: https://github.com/kubewarden/sbomscanner/blob/main/Dockerfile.controller#L2). For example, if we're using Go 1.25.5 now, but specify `latest`, the toolchain version may not match — it might pick up a local version like 1.25.7, which could break the build.

The tricky part is that the error message isn't obvious — it just shows a toolchain failure without indicating the root cause.

I think we should either pin the 
- toolchain version 
- pin the entire Go version.

What do you think?
cc: @wheatdog

WebberHuang1118 (dismissed):

pkg/builder/vm.go
```diff
@@ -183,7 +183,9 @@ func (v *VMBuilder) Memory(memory string) *VMBuilder {
 	if len(v.VirtualMachine.Spec.Template.Spec.Domain.Resources.Limits) == 0 {
 		v.VirtualMachine.Spec.Template.Spec.Domain.Resources.Limits = corev1.ResourceList{}
 	}
-	v.VirtualMachine.Spec.Template.Spec.Domain.Resources.Limits[corev1.ResourceMemory] = resource.MustParse(memory)
```
ibrokethecloud, line 186: this actually makes the code harder to read.

a majority of changes are wrapping up long lines only, if there is an issue with `lll` linter we should either tweak it disable it?

https://golangci-lint.run/docs/linters/configuration/#lll

pkg/containerd/registry.go
```diff
@@ -33,9 +33,9 @@ type AuthConfig struct {
 
 // TLSConfig contains the CA/Cert/Key used for a registry
 type TLSConfig struct {
-	CAFile             string `toml:"ca_file" yaml:"ca_file"`
```
ibrokethecloud, line 36: is this reported as a formatting error?

pohanhuang: I found that when I ran the golangci-lint migration command, it accidentally regenerated the config with some differences.

pkg/builder/vm.go
pohanhuang: I'm fine with both options. @Vicente-Cheng Do you think we should create another issue to decide on the lll linter policy?

My suggestion:
- For this PR: Remove golines formatter to keep the diff focused on deprecated API fixes only
- Separately: Open an issue to discuss whether we want to enable golines/lll as a team-wide standard

This way we can review the deprecated API changes without the formatting noise. WDYT?

Vicente-Cheng: yeah, we can split the formatter issue.
let's focus on the deprecated API changes.

pohanhuang:
Hi @harvester/dev ,
Could you help review this linter issue when you're available?

Also created a follow-up [issue](https://github.com/harvester/harvester/issues/10027) to discuss adding new linters. Please kindly share your thoughts when you get a chance!

Thanks! 🙏

wheatdog (approved):

Vicente-Cheng (approved):

ibrokethecloud (approved):

WebberHuang1118 (approved):

PR #9994, opened 2026-02-02
UI Server: Fix API Paths

m-ildefons:
#### Problem:

The "View in API" function relies on a index.html page generated by Rancher APIServer. This page refers to a Javascript and a CSS document hosted under the `/api-ui` path, however that path isn't routed correctly in Harvester. The actual path where these two documents are available is underneath `/dashboard/api-ui`.

#### Solution:

To fix this, inject the correct path into the index page used for the "View in API" function.
Remove the disfunct implementation stub for the `/api-ui` route as it's no longer needed, since all other web UI paths are served from underneath the `/dashboard` path as well.

#### Related Issue(s):

related-to: harvester/harvester#9614

#### Test plan:

1) Install Harvester
2) Go to Settings --> UI
3) Set `ui-source` to `Bundled`
4) Go to Preferences
5) Enable "View in API"
6) Navigate to any resource in the cluster, click on the menu --> "View in API"

Expected outcome:

There should be a HTML page similar to this screenshot, showing the JSON representation of the selected resource:
Screenshot at 2026-02-02 16-59-55

### DIFF
--- a/pkg/server/router.go
+++ b/pkg/server/router.go
@@ -94,7 +94,6 @@ func (r *Router) Routes(h router.Handlers) http.Handler {
 	vueUI := ui.Vue
 	m.Handle("/dashboard/", vueUI.IndexFile())
 	m.PathPrefix("/dashboard/").Handler(vueUI.IndexFileOnNotFound())
-	m.PathPrefix("/api-ui").Handler(vueUI.ServeAsset("api-ui"))
 
 	if r.options.RancherURL != "" {
 		host, scheme, err := parseRancherServerURL(r.options.RancherURL)
--- a/pkg/server/ui/api.go
+++ b/pkg/server/ui/api.go
@@ -1,10 +1,16 @@
 package ui
 
 import (
+	"fmt"
+
 	apiserver "github.com/rancher/apiserver/pkg/server"
 	"github.com/rancher/apiserver/pkg/writer"
 
 	"github.com/harvester/harvester/pkg/settings"
+)
+
+const (
+	UIAssetPathPrefix = "/dashboard/api-ui"
 )
 
 func ConfigureAPIUI(server *apiserver.Server) {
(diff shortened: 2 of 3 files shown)


Yu-Jack (approved):
LGTM.

I took a look what's going on here. Our `/api-ui/ui.min.css` actually matches the embedded rancher route, but it's still not found. However, when you try `/api-ui/api-ui/ui.min.css`, this one exists. Files location in rancher are `/usr/share/rancher/ui`.

https://github.com/rancher/rancher/blob/3d66af043d1a52a83aeb5c4aa67ada04f0f23ce4/pkg/ui/routes.go#L30
https://github.com/rancher/steve/blob/c80924b808c7bf52ac77781b12ccc1ff3b6d1c57/pkg/ui/handler.go#L148

The reason why this bug came out is that rancher did some changes in route.

v2.12.3: https://github.com/rancher/rancher/blob/d6c02c653d6b6bba1a137c33ca4fc841d62b0172/pkg/ui/routes.go#L30
v2.13.0: https://github.com/rancher/rancher/blob/f93c99323f67e60fc50d17aa4b0ec6f5219575aa/pkg/ui/routes.go#L30

You can see they modified it from `ServeAsset` to `ServeAPIUI`. So, using `/dashboard` might be our only option for this case.

Vicente-Cheng:
Hi @m-ildefons, 
Thanks for your fix, but please keep the PR clean. We don't need any label/milestone/project on the PR side. We tracked the PR with the corresponding issues.

Thanks!

Vicente-Cheng (approved):

m-ildefons:
> I took a look what's going on here. Our /api-ui/ui.min.css actually matches the embedded rancher route, but it's still not found. However, when you try /api-ui/api-ui/ui.min.css, this one exists. Files location in rancher are /usr/share/rancher/ui.

That would explain a lot. I had also tried to make it work under the original path, using the `/api-ui` route, but couldn't. But of course I hadn't checked to see if that route was already occupied by the embedded Rancher :man_facepalming: 

Note, `/api-ui/api-ui/ui.min.css` exists and it seemingly also works ok, but it has a different content than `/dashboard/api-ui/ui.min.css`. So to avoid confusion and surprises in the future (when Rancher inevitably introduces breaking changes again), I prefer the `/dashboard/api-ui/...` solution.

> We don't need any label/milestone/project on the PR side.

Noted. I'm sorry, I'll avoid it in the future.

Issue #9999, opened 2026-02-03
[BUG] Restrict enableDHCP update on a subnet when overlay VM instances are using it


rrajendran17:
### Describe the Bug

This https://github.com/harvester/harvester/issues/9957 takes care of patching the managedtap binding by default to overlay vms during create and update if dhcp is enabled on the subnet.

There can be cases where enabledhcp is true in the subnet while vm creation and later it is updated to false and vm's interface binding stays in managedtap binding.This requires the vm to be saved and restarted.
The vice-versa case should also work.

This task should take of restricting such actions (enableDHCP value changes on a subnet) when vmis are using it.

### To Reproduce

case 1:

1.create an overlay network
2.create a subnet with enableDHCP: true
3.create a vm
4.Check the interface binding of the vm and it shows as managedtap binding
5.Now update the enableDHCP: false in the subnet
6.webhook validation error

case 2:

1.create an overlay network
2.create a subnet with enableDHCP: false
3.create a vm
4.Check the interface binding of the vm and it shows as bridge
5.Now update the enableDHCP: true in the subnet
6.webhook validation error

case 3:
1.create an overlay network
2.create a subnet with enableDHCP: false
3.create a vm
4.Check the interface binding of the vm and it shows as bridge
5.Now stop the vm
6.Now update the enableDHCP: true in the subnet
7.step 6 should be successful
8.Start the VM and it should be running with managedtap binding.

### Expected Behavior

subnet validation webhook must restrict the changes when vmis are present.

### Support Bundle for Troubleshooting

.

### Environment

- Harvester version:v1.7.0

rrajendran17:
Add changes to subnet validation webhook to restrict enableDHCP changes when vmis are using those subnets.
VMs must be stopped before updating enableDHCP value on subnets.

asc5543:
According to https://github.com/harvester/harvester/issues/9957#issuecomment-4053668606 , the interface is still `bridge`, not `managedtap`

rrajendran17:
https://github.com/harvester/harvester/pull/10215 is needed for this testing.Moving to review, once the fix is merged will move it to "Ready to Testing"

Vicente-Cheng:
Please wait for the rc2 to test

asc5543:
## Test result
Test PASS

## Environment
* Harvester ISO version: `v1.8.0-rc2`
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): `2 node via ipxe-example`

[...]

Issue #9957, opened 2026-01-27
[ENHANCEMENT] Patch managedtap binding to overlay VMs by default


rrajendran17:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

kube-ovn DHCP did not work with the bridge binding because the Linux bridge sits between the VM TAP device and kubeOVN's OVS datapath,interfering with DHCP broadcast traffic that KubeOVN's DHCP implementation depends on. As a result, VMs were not assigned with IPs and default route was not installed on VMs using overlay Networks. Changing the interface binding to managedtap fixes this by attaching the VM’s TAP device directly to OVS, giving kube-ovn full visibility and control over the VM’s MAC and DHCP traffic. 

https://github.com/harvester/harvester/issues/8847 was a proposal to solve this by editing the VM template interfaces with managedtap binding every time before VM creation. This works but this requires user effort to edit this field every time and it will be a problem when creating downstream clusters backed by overlay networks, where users has to edit the VM once created and then restart to apply this.

### Describe the solution you'd like

This enhancement proposes to patch the managedtap binding to interfaces of overlay VMs by VM mutator by default during VM creation.

### Describe alternatives you've considered

Other alternative was to add a field in UI for users to select managedtap binding when creating VMs using overlay Networks. 

For overlay interfaces, managedtap should be the default binding, as it provides correct L2 visibility and reliable DHCP, unlike the legacy bridge mode.So better option is to patch this at the backend by default instead of making it user aware through UI.

### Additional context
https://github.com/harvester/harvester/issues/9999

_No response_

rrajendran17:
@oilbeater Currently if the interface has managedtap binding, then kubeovn uses DHCP to allocate ip addresses to VMs and this requires subnet to have enableDHCP as true.
Without managedtap binding, with bridge mode, vms get ip addresses from kubeovn IPAM but there is a kubevirt issue associated with this.
1.default route not installed on the vm guest os.

To solve above I created this GH issue where idea is to patch managedtap binding by default on vms, but this also requires enableDHCP to be true on subnets used by these vms. I would like to know if its correct to enableDHCP as true on all subnets by default from Harvester along with managedtap binding ? Are there any other alternatives from kubeovn ?

mrostamii:
@rrajendran17 What if we add a new boolean input when creating a new VM in UI? Or, it could be added to the "Networks" tab.

The steps could be:
- User chooses a Network
[...]

PR #10000, opened 2026-02-03
patch managedtap binding for overlay vms

rrajendran17:
#### Problem:
overlay vms does not work with ovn dhcp on bridge mode

#### Solution:
patch managedtap binding by default for overlay vms during create/update if dhcp is enabled in the subnet.
This needs to be taken care https://github.com/harvester/harvester/issues/9999 if managedtap binding is patched by default based on enableDHCP option in subnet

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9957

#### Test plan:
case 1:

1.create an overlay network
2.create a subnet with enableDHCP: true
3.create a vm attach it to overlay network
4.Check the interface binding of the vm and it shows as managedtap binding
5.update the config to a overlay network on subnet with enableDHCP: false
6.check if the interface mode changes to bridge 

case 2:

1.create an overlay network
2.create a subnet with enableDHCP: false
3.create a vm attach it to overlay network
4.Check the interface if it shows as bridge
5.update the config to a overlay network on subnet with enableDHCP: true
6.check if the interface mode changes to managedtap binding

case 3:

1.create an overlay network
2.create a subnet with enableDHCP: true
3.create a vm attach it to overlay network
4.Check the interface binding of the vm and it shows as managedtap binding
5.update the config to a l2network
6.check if the interface mode changes to bridge 

case 4:

1.create an overlay network
2.create a subnet with enableDHCP: true
3.create a vm and attach it to l2network 
4.Check the interface if it shows as bridge
5.update the config to a overlay network on subnet with enableDHCP: true created in step 1 and 2
6.check if the interface mode changes to managedtap binding

case 5:
1.create an overlay network
2.create a subnet with enableDHCP: false
3.create a vm attach it to l2network
4.Check the interface if it shows as bridge
5.update the config to a overlay network on subnet with enableDHCP: false created in step 1 and 2
6.check if the interface mode stays in bridge mode

case 5:
verify if default route is installed on vm guest os with managedtap binding
1.create an overlay network
2.create a subnet with enableDHCP: true
3.create a vm attach it to overlay network
4.Check the interface binding of the vm and it shows as managedtap binding
5.login to the vm and check "ip route show" (it should show a default route via gw in the subnet cidr range)


pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -36,19 +40,22 @@ func NewMutator(
...
 
 type vmMutator struct {
 	types.DefaultMutator
-	setting ctlharvesterv1.SettingCache
-	nad     ctlcniv1.NetworkAttachmentDefinitionCache
-	kubvirt ctlkubevirtv1.KubeVirtCache
+	setting       ctlharvesterv1.SettingCache
+	nad           ctlcniv1.NetworkAttachmentDefinitionCache
+	kubvirt       ctlkubevirtv1.KubeVirtCache
+	kubeovnSubnet ctlkubeovnv1.SubnetCache
```
w13915984028, line 58: one question: the OVN related CRDs are installed until OVN addon is enabled, before that, the mutator might report error, did you test this? we did some tricks to restart the webhook when the ovn addon is enabled, but I did not find the PR yet.

rrajendran17: As we discussed, this check https://github.com/rrajendran17/harvester-harvester/blob/77f4aa88a61d68caf3fe9786c515d6e5dbf9c438/pkg/webhook/resources/virtualmachine/mutator.go#L759 makes sure that the subnet crd missing error is applicable only for overlay NADs.
The overlay VMs will fail to start anyways if relevant ovn resources/crds are not found. 

This https://github.com/harvester/harvester/issues/8685 also needs to be fixed which makes sure that there are no overlay vms before disabling the ovn add-on, so there are no dangling vms.I will add these changes as well and raise a separate PR.

w13915984028: BTW: please also check if below info occurs on the webhook, even though the mutator is triggered when a VM is created/updated, the webhook POD itself might watch OVN related CRDs from POD is up, the potential continuous error log will be a bit annoying

```
Failed to list *v1.MyCustomResource: the server could not find the requested resource
```


btw: the current hack of checking CRD existing

https://github.com/harvester/network-controller-harvester/blob/67b26a944114641d2b43ef4c37524dc0a38f0b5e/cmd/webhook/main.go#L124-L131

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -683,3 +700,119 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
...
+	if err != nil {
+		return err, enabled
+	}
+
+	if nad.Labels == nil {
+		return fmt.Errorf("NAD labels are nil for %s/%s", namespace, name), enabled
+	}
+
+	//overlay network and list subnet error
+	if subneterr, enabled = m.checkDHCPEnabledOnSubnet(namespace, name); subneterr != nil && nad.Labels[keyNetworkType] == string(OverlayNetwork) {
```
w13915984028: Consider to use below to replace L754

The mutator only cares the correct overlay NAD type and patches related info, and avoids to return un-necessary error like L754 does. Strict check (if required) is mainly done on validator.

It returns early when NAD is not overlay.

```
	if nad.Labels[keyNetworkType] != string(OverlayNetwork) {
		return nil, false
	}
```

refer:
https://github.com/harvester/harvester/blob/8639cfb12122194e8fa52acf53426d719e8cdd9f/pkg/webhook/resources/virtualmachine/mutator.go#L475-L478

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -141,6 +153,11 @@ func (m *vmMutator) Update(_ *types.Request, oldObj runtime.Object, newObj runti
 		return nil, err
 	}
 
+	patchOps, err = m.patchManagedTapBinding(newVM, patchOps)
```
w13915984028, line 156: patch newly created VM is necessary;  for existing VMs, not sure we should patch them.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -683,3 +700,119 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
...
+	}
+
+	interfaceList := getInterfaceList(vm)
+	for idx, iface := range vm.Spec.Template.Spec.Domain.Devices.Interfaces {
+		networkName, exists := interfaceList[iface.Name]
+		if !exists {
+			continue
+		}
+
+		getNadAndDHCPOptionErr, enabled := m.getNadAndDHCPOption(networkName, vm)
```
w13915984028: From below case description, it seems the `getNadAndDHCPOption` needs to return: `nad type, dhcp enabled, error`

there are cases:

```
non-overlay to overlay; plus dhcp enabled or not
overlay to overlay, dhcp changed
overlay to non-overlay, skip dhcp enabled or not

no change at all
```

For `update` case, more considerations are needed:

The `vm` update might be triggered by other case like adding new label/annoation, update status;  if `mutator` is triggered anyway and updates the interfaces, it might cause UI shows VM rebooting is required.

I am not fully sure: should we mutator VM only when it is created or mutator VM whenever there is updates on VM, Or when VM's spec network/interface has changes ?

The NAD option dhcp could be toggled independently, which might cause VM's spec is changed implicitly.

w13915984028 (commented):
Some further questions, thanks.

pkg/webhook/resources/virtualmachine/mutator.go
rrajendran17: we allow update of network interfaces type on a running vm, so IMO patching should also handle update cases for change in network types from overlay to non overlay and vice-versa.Otherwise the vm interfaces will be left with older binding and will cause error.

pkg/webhook/resources/virtualmachine/mutator.go
rrajendran17: Right, may be a stricter check like only if the vm spec interfaces change, then call patchManagedTapBinding should avoid unnecessary updates.
Also this https://github.com/harvester/harvester/issues/9999 takes care of restricting the toggling of dhcp flag on a subnet for a running VM.The VM has to be stopped before enableDHCP flag update on ovn subnet.

pkg/webhook/resources/virtualmachine/mutator.go
rrajendran17: you are right, I see those errors if crd is not present.I have reused the same from network controller to fix this.

pkg/webhook/resources/virtualmachine/mutator.go
rrajendran17: The logic patches contents only when there are related network interfaces changes, but the only instance where this differs is when an interface is overlay network(managedtap binding already set) and user does some other update operation (not related to network interfaces), the patchmanagedtapbinding function has to remove the masquerade interface binding set on interface.This is because kubevirt by default tries to add masquerade interfacebindingmethod on an interface if its not bridge mode though managedtap binding is present and two interface binding methods are not allowed at the same time.(I could raise an upstream issue for this)

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -683,3 +703,140 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
 
 	return patchOps, nil
 }
+
+func (m *vmMutator) checkDHCPEnabledOnSubnet(namespace, name string) (error, bool) {
+	if !m.subnetCRDExists {
```
ibrokethecloud: we should move this as a condition to calling `patchManagedTapBinding` in Create/Update handlers to avoid having to run all the remaining checks when CRD does not really exist

pkg/webhook/server/mutation.go
```diff
@@ -31,12 +32,16 @@ func Mutation(clients *clients.Clients, options *config.Options) (http.Handler,
 	vmImgCache := clients.HarvesterFactory.Harvesterhci().V1beta1().VirtualMachineImage().Cache()
 	vmCache := clients.KubevirtFactory.Kubevirt().V1().VirtualMachine().Cache()
 	kubevirtCache := clients.KubevirtFactory.Kubevirt().V1().KubeVirt().Cache()
+	var kubeovnSubnetCache ctlkubeovnv1.SubnetCache
+	if crdExists && clients.KubeovnFactory != nil {
```
w13915984028: looks check `crdExists` is enough, 

when `crdExists`, the input `clients.KubeovnFactory` can't be empty

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -36,19 +40,25 @@ func NewMutator(
...
 
 type vmMutator struct {
 	types.DefaultMutator
-	setting ctlharvesterv1.SettingCache
-	nad     ctlcniv1.NetworkAttachmentDefinitionCache
-	kubvirt ctlkubevirtv1.KubeVirtCache
+	setting         ctlharvesterv1.SettingCache
+	nad             ctlcniv1.NetworkAttachmentDefinitionCache
+	kubvirt         ctlkubevirtv1.KubeVirtCache
+	kubeovnSubnet   ctlkubeovnv1.SubnetCache
```
w13915984028: `subnetCRDExists` could be removed, use `kubeovnSubnet == nil` to check if the cache is valid or not

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -141,6 +160,15 @@ func (m *vmMutator) Update(_ *types.Request, oldObj runtime.Object, newObj runti
 		return nil, err
 	}
 
+	if !m.subnetCRDExists {
```
w13915984028: remove this block, check `kubeovnSubnet == nil` at the beginning of `patchManagedTapBinding` to return quickly

`Create()`, `Update()` do not involve the trick

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -683,3 +711,136 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
...
+	}
+
+	return enabled, nil
+}
+
+func (m *vmMutator) patchManagedTapBinding(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) (types.PatchOps, error) {
+	if vm == nil || vm.Spec.Template == nil {
+		return patchOps, nil
+	}
+
```
w13915984028, line 774: check `m.kubeovnSubnet == nil` and return directly

pkg/webhook/resources/virtualmachine/mutator_test.go
```diff
@@ -1363,3 +1368,274 @@ func TestPatchInterfaceMacAddress(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			clientset := fake.NewSimpleClientset()
+			createDefaultKubeVirt(clientset)
+			mutator := setupTestMutator(clientset)
```
w13915984028, line 1715: extend the `setupTestMutator` to support the subnetcache is/is-not nil cases

the `patchManagedTapBinding` works really only when subnetcache is not nil

pkg/webhook/server/server.go
```diff
@@ -56,8 +60,60 @@ func New(ctx context.Context, restConfig *rest.Config, options *config.Options)
...
+			checkStatus, err := isSubnetsCRDPresent(ctx, cfg)
+			if err != nil {
+				continue
+			}
+
+			if initialBootStatus == checkStatus {
+				continue
+			}
+
+			logrus.Info("change in CRD state detected, triggering shutdown")
```
w13915984028, line 103: Harvester webhook is called more frequently than network webhook.

In a multi-cluster env, when there are 3 harvester-webhook PODs, are they killed simultaneously when ovn subnet CRD is created/removed? what's your observation?

We might need to consider an alternative solution:

Let the addon controller or add a new CRD controller to roll-update harvester-webhook, network-webhook deployment when subnet CRD is created or removed.  Instead of let each webhook to kill self.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -683,3 +711,136 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
...
+	//Overlay network and subnets are not present
+	if subnetErr, enabled = m.checkDHCPEnabledOnSubnet(namespace, name); subnetErr != nil {
+		return enabled, subnetErr
+	}
+
+	return enabled, nil
+}
+
+func (m *vmMutator) patchManagedTapBinding(vm *kubevirtv1.VirtualMachine, patchOps types.PatchOps) (types.PatchOps, error) {
```
w13915984028, line 770: what the consideration of this case:

1. A vm has overlay interface
2. The VM is stopped
3. OVN is disabled (is this allowed, or we force user to remove overlay interface first?)

If dangling overlay interface is possiblly existing, the mutator patch is skipped due to above crd/subnetcache check. What's your consideration of this check? 

Could you list all current webhook (including network) rules about the interface&dhcp&ovn enable/disable  changes, need to figure out if there are some gaps, thanks.

Harvester VM mutator webhook is very error-prone, some side-effects might happen unexpectedly.

currently:
(1) Addon webhook: seems no protection upon ovn disabling
https://github.com/harvester/harvester/blob/cb0e860cec312242337befad2f05bf48c1b506c9/pkg/webhook/resources/addon/validator.go#L104

(2) Network webhook: doesn't allow ovn nad dhcp option change when vmi is running
https://github.com/harvester/network-controller-harvester/pull/236

Might need to refactor `patchManagedTapBinding`:
(a)
when `subnetcache==nil`, remove any existing binding;  this does not rely on the nad option dhcp enabled
```
`{"op":"remove","path":"/spec/template/spec/domain/devices/interfaces/%d/binding"}`,
```

(b) when `subnetcache != nil`, get option and patch accordingly

https://github.com/harvester/harvester/pull/10000/changes#diff-1bde4c748f58e474d67fd449ce94739fe11076c2ba5c63f79cfda2f94c4cb0a3R800-R810

w13915984028 (commented):
Some questions about the code placement, thanks.

pkg/webhook/server/server.go
w13915984028: cc @ibrokethecloud for this candidate solution

pkg/webhook/resources/virtualmachine/mutator.go
rrajendran17: https://github.com/harvester/harvester/issues/8685 should take care of restricting users from disabling the ovn addon when overlay VM are using it including the stopped one. This will make sure we remove the overlay vm before disabling the ovn addon, so it is good to call patchedmanagedtapbinding only when subnet crds are enabled.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -683,3 +700,141 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
 
 	return patchOps, nil
 }
+
+func (m *vmMutator) checkDHCPEnabledOnSubnet(namespace, name string) (error, bool) {
```
w13915984028: the usual return format is `(bool, error)`

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -683,3 +700,141 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
...
+	if err != nil {
+		return enabled, err
+	}
+
+	if nad.Labels[keyNetworkType] != string(OverlayNetwork) {
+		return false, nil
+	}
+
+	//Overlay network and subnets are not present
+	if subnetErr, enabled = m.checkDHCPEnabledOnSubnet(namespace, name); subnetErr != nil {
```
w13915984028: replace L764~L768 with `m.checkDHCPEnabledOnSubnet(namespace, name)`

 after change `checkDHCPEnabledOnSubnet` return param order

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -683,3 +700,141 @@ func (m *vmMutator) patchInterfaceMacAddress(vm *kubevirtv1.VirtualMachine, patc
...
+		}
+		interfaceList[network.Name] = network.Multus.NetworkName
+	}
+	return interfaceList
+}
+
+func (m *vmMutator) getNadAndDHCPOption(networkName string, vm *kubevirtv1.VirtualMachine) (bool, error) {
+	words := strings.Split(networkName, "/")
+	var namespace, name string
+	var enabled bool
```
w13915984028: remove `enabled, subnetErr` var,

L751 `return false, fmt.Errorf`
L756 `return false, err`
to keep similar formating like L760

pkg/webhook/resources/virtualmachine/mutator_test.go
```diff
@@ -1363,3 +1376,338 @@ func TestPatchInterfaceMacAddress(t *testing.T) {
...
+		name       string
+		interfaces []kubevirtv1.Interface
+		networks   []kubevirtv1.Network
+		expected   []patch
+		nad1       *cniv1.NetworkAttachmentDefinition
+		nad2       *cniv1.NetworkAttachmentDefinition
+		ovnsubnet  *kubeovnapiv1.Subnet
+	}{
+		{
+			name: "overlay network removes bridge and adds managedtap binding",
```
w13915984028, line 1397: please add another case: when vm has already been patched, the mutator should not patch again

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -15,17 +15,21 @@ import (
...
 	"github.com/harvester/harvester/pkg/util/network"
 	"github.com/harvester/harvester/pkg/util/virtualmachine"
 	"github.com/harvester/harvester/pkg/webhook/types"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
 )
 
 const (
 	networkGroup      = "network.harvesterhci.io"
 	keyClusterNetwork = networkGroup + "/clusternetwork"
+	OverlayNetwork    = "OverlayNetwork"
```
w13915984028: if `OverlayNetwork` is a public const, the better place is utils/contst, or use `overlayNetwork` as const for now

w13915984028 (commented):
Some additional code cosmetic suggestions, thanks.

rrajendran17:
@ibrokethecloud @w13915984028 The PR is now ready with all updated changes.Please check.Thanks

w13915984028 (dismissed):
LGTM, thanks.

BTW, the `CRD watching and self-killing` could be discussed and optimized on other ticket.

w13915984028 (approved):
LGTM, thanks.

pkg/webhook/server/server.go
ibrokethecloud: ideally the kubeovn addon is to be treated more like a cni, and I'd expect less frequent changes to its state. We already need to reboot the nodes as part of the uninstall process. Another option during the upgrade phase may be to add a random delay of between say 5 to 30 seconds, so on HA setups all pods do not restart at the same time.

ibrokethecloud (approved):

Issue #10001, opened 2026-02-03
[ENHANCEMENT] Improve integration test reliability: bump kind (K8s) and ginkgo versions


pohanhuang:
### Task request

Harvester CI uses `kind` to spin up a Kubernetes cluster for integration/e2e tests.

- Current `kindest/node` (Kubernetes v1.31) version used in CI is not aligned with the Kubernetes version we intend to test against v1.34.
- The `ginkgo` version in go.mod differs from the version pinned in the dapper Dockerfile; we should sync them as well to avoid inconsistencies between CI and local runs.

wheatdog:
The installation of `kind` is defined in https://github.com/harvester/harvester/blob/10d64609d1bd5f9e51a800515a76ac2f9d5ad48d/Dockerfile.dapper#L58-L59

The expected k8s version to be used is defined in https://github.com/harvester/harvester-installer/blob/650752c7c95af2126813044c80be4fc2cccc0571/scripts/version-rke2#L1

PR #10003, opened 2026-02-03
ci: bump kind, ginkgo version in dapper

pohanhuang:
#### Problem:
- Current kindest/node (Kubernetes v1.31) version used in CI is not aligned with the Kubernetes version we intend to test against v1.34.
- The ginkgo version in go.mod differs from the version pinned in the dapper Dockerfile; we should sync them as well to avoid inconsistencies between CI and local runs.

#### Solution:

- bump kind to [v0.30.0 ](https://github.com/kubernetes-sigs/kind/releases/tag/v0.30.0)
  - in go.mod
  - in Dockerfile.dapper
- ginkgo version v2.23.4 align with [go.mod](https://github.com/harvester/harvester/blob/master/go.mod)
#### Related Issue(s):
#10001


ihcsim:
We should also update the [`kind` module](https://github.com/harvester/harvester/blob/c50975695f06b36932879bf6b3e7c37caf220b07/go.mod#L144) which is used to run the [integration test](https://github.com/harvester/harvester/tree/master/tests).

pohanhuang:
Hi @ihcsim ,
Thanks for reminding! You're absolutely correct. In this PR I also bump the kind version as well.

ihcsim (approved):

pohanhuang:
@wheatdog Can you help review this bump PR when you are available?
Thanks.

pohanhuang:
Hi @harvester/dev,
Can you help review when you are available? I would like to get more review before it merge.
Thanks.

wheatdog (approved):

Issue #10067, opened 2026-02-18
[ENHANCEMENT] Switch from nginx ingress to traefik


ibrokethecloud:
### Task request

With the nginx ingress [deprecation](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/) we need to plan work needed to move to traefik

RKE2 already has support for Traefik, and we need to look at changes needed for exposing the rancher ingress via traefik.

The task is a place holder to investigate changes needed in Harvester once we have more clarity on changes being done on RKE2 end.

### Additional context

- https://www.suse.com/c/trade-the-ingress-nginx-retirement-for-up-to-2-years-of-rke2-support-stability
- Traefik will be the default ingress controller in rke2 1.3.6. https://github.com/rancher/rke2/issues/9786

innobead:
Moved this to 1.8.0, since @ibrokethecloud is working on it.

cc @harvester/release

ibrokethecloud:
Though the install / upgrade path PR's are ready, moving this to v1.9.0 to give us more time to QA and find edge cases for ingress behaviour.

ihcsim:
Please remove `rancher/mirrored-library-nginx:1.29.1-alpine` from the Harvester bundle once we switch over to Traefik, per requirement from internal audit - Thanks.

lanfon72:
Verified this enhancement has been applied.

Test Information
----
* Environment: **MegaRAC SP-X 3 nodes**
* Harvester Version: **v1.8.1** -> **v1.9.0-RC3**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Perform upgrade
3. Upgrade should completed successfully
4. Check embedded Rancher
    1. `ingress-controller` be replaced by traefik
    <img width="1883" height="489" alt="Image" src="https://github.com/user-attachments/assets/5dd174cf-6060-423c-9e3a-0a22f8dc87e5" />
    2. No more nginx/ingress daemon-set
      <img width="1575" height="391" alt="Image" src="https://github.com/user-attachments/assets/e0eb9c30-636f-4383-94ff-239040318708" />
    3. There should have `TLSStore` and `TLSOption` CRD in traefik scope
      <img width="1014" height="884" alt="Image" src="https://github.com/user-attachments/assets/c8cd0c52-10c6-478a-81fd-743f947cd6ec" />
5. Check there have a new settings `traefik-default-tls-options` available
[...]

PR #10074, opened 2026-02-20
switch to traefik ingress controller

ibrokethecloud:
#### Problem:

Nginx ingress is being deprecated. RKE2 has already supported traefik as an ingress controller.

The PR introduces changes to ensure existing clusters are switched to traefik during upgrade.

#### Solution:

The PR introduces the following changes
* Check in the upgrade script to add definition to switch to traefik as ingress controller. 
* Changes to rancher controller to annotate traefik service with kube-vip annotations. Controller will continue to sync nginx svc until traefik svc shows up, to handle the upgrade scenario where post upgrade nginx will be uninstalled and swapped with traefik.
* Changes to setting controller to sync default TLS certificate from nginx helmchartconfig to the traefik native `TLSStore` store.
* Changes to setting controller to sync SSL ciphers to traefik native `TLSOptions` CRD

Traefik does not have a simple construct to limit TLS versions. As a result the SSL setting for protocols has not yet been synced and we need to decide on a better solution.
#### Related Issue(s):

https://github.com/harvester/harvester/issues/10067


pkg/controller/master/rancher/register.go
```diff
@@ -291,3 +294,75 @@ func (h *Handler) RancherTokenOnChange(_ string, token *v3.Token) (*v3.Token, er
...
+	// set vip loadBalancer type and ip
+	enabled, err := strconv.ParseBool(vip.Enabled)
+	if err != nil {
+		return err
+	}
+
+	svcCopy := svc.DeepCopy()
+
+	if enabled && vip.ServiceType == corev1.ServiceTypeLoadBalancer {
+		svc.Spec.Type = vip.ServiceType
```
w13915984028, line 355: it looks unusual to update the original object instead of copied object

pkg/controller/master/rancher/register.go
w13915984028, line 153: it is not related to the scope of this PR, but seems safe to remove the real `legacy` function `cleanUpLegacyExposeService`

package/upgrade/upgrade_node.sh
```diff
@@ -832,6 +839,7 @@ command_post_drain() {
 
   generate_hostname_persistance
 
+  switch_ingress
```
w13915984028, line 842: not sure the `command_pre_drain` is a better place to put this and let rke2 switch to traefik when new rke2 server is loaded

pkg/controller/master/setting/ssl_parameters.go
```diff
@@ -28,37 +34,49 @@ func (h *Handler) syncSSLParameters(setting *harvesterv1.Setting) error {
...
-		return err
-	}
-	toUpdateHelmChartConfig := helmChartConfig.DeepCopy()
-
-	var values = make(map[string]interface{})
-	if err := yaml.Unmarshal([]byte(helmChartConfig.Spec.ValuesContent), &values); err != nil {
-		return err
-	}
-
+	tlsOptionRes := schema.GroupVersionResource{Group: "traefik.io", Version: "v1alpha1", Resource: "tlsoptions"}
```
w13915984028, line 37: old code handles helmchartconfig object to ensure the change is synced with rke2 from the source;  if we update downstream object directly, could it be amended by rke2?

w13915984028 (commented):
Some questions, thanks.

pkg/controller/master/rancher/register.go
ibrokethecloud: i am not sure how it matters.. the entire aim of the copy is to allow me to identify if there were any changes and accordingly trigger an update of the object

pkg/controller/master/setting/ssl_parameters.go
ibrokethecloud: we do not need to inject traefik values via helmchart config, since the secret reference changes based on setting being used. I have made some more changes in the latest commit.

w13915984028: rke2 usually runs helm-install/helm-update on booting up/leader switch over each time, if our change is not on helmchart/helmchartconfig, it could be amended by rek2. i am not sure directly change secret/tlsoption could persist

https://github.com/rancher/rke2-charts/blob/8f1aa9cbd3103b27809268cc890cd1333f954848/charts/rke2-traefik-1.30-1.31/rke2-traefik/27.0.204/templates/tlsoption.yaml#L2

rke2-guest cluster has switched nginx to traefik by default, it has such an object created

For Harvester, when ingress controller becomes traefik, things tend to be similar.

```
root@gc3-pool1-pxz6f-qszp2:~# kubectl get helmchartconfig -n kube-system rke2-traefik -oyaml
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  annotations:
    objectset.rio.cattle.io/applied: ...
    objectset.rio.cattle.io/id: ""
    objectset.rio.cattle.io/owner-gvk: k3s.cattle.io/v1, Kind=Addon
    objectset.rio.cattle.io/owner-name: managed-chart-config
    objectset.rio.cattle.io/owner-namespace: kube-system
  creationTimestamp: "2026-05-05T09:15:09Z"
  generation: 1
  labels:
    objectset.rio.cattle.io/hash: df7606cd22a7f791463faa741840385c45b26d8a
  name: rke2-traefik
  namespace: kube-system
  resourceVersion: "392"
  uid: 390560d4-1d6b-47b4-bc80-2359bf7e45c5
spec:
  failurePolicy: reinstall
  valuesContent: '{"global":{"cattle":{"clusterId":"c-m-6df52dsp"}}}'

```

ibrokethecloud:
To build the PR needs associated installer branch changes from the installer branch referenced by PR https://github.com/harvester/harvester-installer/pull/1232

pkg/controller/master/rancher/register.go
```diff
@@ -291,3 +308,75 @@ func (h *Handler) RancherTokenOnChange(_ string, token *v3.Token) (*v3.Token, er
...
+		svc.Spec.LoadBalancerIP = vip.IP
```
w13915984028: In current coce, https://github.com/harvester/harvester/blob/f29abc8df8e4b61ce4bb160af9c5623168a0bf8e/pkg/controller/master/rancher/register.go#L192, the `svc.Spec.LoadBalancerIP` is not set

pkg/controller/master/setting/ssl_certificate.go
```diff
@@ -34,7 +33,7 @@ func (h *Handler) syncSSLCertificate(setting *harvesterv1.Setting) error {
 }
 
 func (h *Handler) resetCertificates() error {
-	if err := h.updateIngressDefaultCertificate(util.CattleSystemNamespaceName, util.InternalTLSSecretName); err != nil {
+	if err := h.updateTraefikTLSStore(util.InternalTLSSecretName); err != nil {
```
w13915984028, line 36: per traefik doc, when TLSStore is updated, it watches and acts automatically, the `redeploySSLCertificateWorkload` seems not necessary to redeploy traefik daemonset

pkg/controller/master/rancher/daemonset.go
```diff
@@ -0,0 +1,138 @@
...
+	if ds.Name == util.Rke2TraefikAppName && ds.Namespace == util.KubeSystemNamespace {
+		logrus.Info("found traefik ds, checking if ingress object needs an update")
+		// traefik exists.. lets verify ingress and update settings
+		ingressObj, err := h.ingresses.Get(util.CattleSystemNamespaceName, util.RancherExposeIngressName, metav1.GetOptions{})
+		if err != nil {
+			return ds, fmt.Errorf("error looking up ingress object %s: %w", util.RancherExposeIngressName, err)
+		}
+
+		var rerunExpose bool
+		// update ingressClassName to traefik if needed
```
w13915984028, line 42: a peer PR update is needed on https://github.com/harvester/harvester-installer/pull/1232

pkg/controller/master/rancher/daemonset.go
```diff
@@ -0,0 +1,138 @@
...
+		if ingressObj.Spec.IngressClassName == nil || *ingressObj.Spec.IngressClassName != util.TraefikIngressClassName {
+			ingressObj.Spec.IngressClassName = ptr.To(util.TraefikIngressClassName)
+			if _, err := h.ingresses.Update(ingressObj); err != nil {
+				return ds, fmt.Errorf("error updating ingress class on rancher-expose ingress: %w", err)
+			}
+			rerunExpose = true
+		}
+
+		// re-run registerExposeService may be needed since in HA nodes during upgrade traefik / nginx will be flapping until all controlplane nodes
+		// get the update as part of the OS update
```
w13915984028, line 52: should move L54  before L44? to enusre the ingress obejct is updated only when traefik service is also existing

pkg/controller/master/rancher/register.go
```diff
@@ -132,19 +154,33 @@ func Register(ctx context.Context, management *config.Management, options config
...
+			return h.createIngressExposeService()
+		}
+		return nil
 	}
-	// Update annotations of the ingress-expose service created before Harvester v1.2.0
-	if svc.Annotations == nil || svc.Annotations[keyKubevipIgnoreServiceSecurity] != trueStr {
-		return h.updateIngressExposeService(svc)
+
+	// clean up old nginx service as it is no longer needed
+	if err := h.cleanupIngressExpose(); err != nil {
```
w13915984028, line 184: this cleanup is always called

we could utilize https://github.com/harvester/harvester-installer/pull/1232 to add an annotaion to Ingress or some object to help the controller know if the cluster is newly installed or upgraded, to stop unnecessary try to delete the legacy object

w13915984028 (commented):
Some quetions per latest code, thanks.

pkg/controller/master/setting/ssl_parameters.go
ibrokethecloud: the rke2 upgrade/restart should not impact this, as traefik only allows one default tlsstore across all namespaces. My initial plan was to copy the tls secret from cattle-system into kube-system namespace, but since tlsstore can be in any namespace, it was a lot easier to just watch ssl settings and accordingly update the tlsstore to point to the correct secret

pkg/controller/master/rancher/register.go
ihcsim: a local copy is fine; only potentially problematic, if it's a reference to a cache object

w13915984028 (dismissed):
We could merge this PR to have more times to test the feature, and fix some edge cases if necessary.

LGTM, thanks.

pkg/controller/master/rancher/register.go
```diff
@@ -291,3 +294,75 @@ func (h *Handler) RancherTokenOnChange(_ string, token *v3.Token) (*v3.Token, er
...
+			svc.Annotations[keyKubevipRequestIP] = vip.IP
+			svc.Annotations[keyKubevipHwaddr] = vip.HwAddress
+			svc.Annotations[keyKubevipLoadBalancerIPs] = vipDHCPLoadBalancerIP
+		} else {
+			svc.Annotations[keyKubevipLoadBalancerIPs] = vip.IP
+		}
+	}
+	if !reflect.DeepEqual(svc, svcCopy) {
+		if _, err = h.Services.Update(svc); err != nil {
+			return fmt.Errorf("error updating %s svc: %w", svc.Name, err)
```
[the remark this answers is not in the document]
ihcsim: @copilot Re-evaluate your feedback. The `svc` is not a reference to an informer cache object. See the signature of the `Service` field in the `Handler` struct on line 66.

pkg/controller/master/rancher/register.go
```diff
@@ -139,19 +161,33 @@ func Register(ctx context.Context, management *config.Management, options config
...
+			return h.createIngressExposeService()
+		}
+		return nil
 	}
-	// Update annotations of the ingress-expose service created before Harvester v1.2.0
-	if svc.Annotations == nil || svc.Annotations[keyKubevipIgnoreServiceSecurity] != trueStr {
-		return h.updateIngressExposeService(svc)
+
+	// clean up old nginx service as it is no longer needed
+	if err := h.cleanupIngressExpose(); err != nil {
```
ihcsim, line 184: shall we patch traefik first and make sure it works, before deleting nginx?

pkg/controller/master/setting/ssl_parameters.go
```diff
@@ -28,37 +34,49 @@ func (h *Handler) syncSSLParameters(setting *harvesterv1.Setting) error {
...
-	}
-	if sslParameter.Protocols == "" {
-		data.RemoveValue(values, "controller", "config", "ssl-protocols")
-	} else {
-		data.PutValue(values, sslParameter.Protocols, "controller", "config", "ssl-protocols")
+		return h.dynamicClient.Resource(tlsOptionRes).Namespace(util.KubeSystemNamespace).Delete(h.ctx, defaultTLSOptionName, metav1.DeleteOptions{})
 	}
+	ciphers := strings.Split(sslParameter.Ciphers, ":")
+	obj := generateUnstructuredTLSOption(ciphers)
+	return h.apply.WithDynamicLookup().WithSetID(util.HarvesterChartReleaseName).ApplyObjects(obj)
```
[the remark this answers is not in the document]
ihcsim: I vaguely we remember we touched on this topic before, so yes, this will be a problem because there isn't a direct mapping between the format of the OpenSSL cipher and the IANA cipher.

We have to document this and/or provide a tool that, for example, calls the [`tls.CipherSuiteName()`](https://pkg.go.dev/crypto/tls#CipherSuiteName) which uses user-provided cipher hex to determine the IANA cipher string. The full mapping can be found at https://testssl.sh/openssl-iana.mapping.html. For example, if user uses the `ECDHE-ECDSA-AES128-GCM-SHA256` cipher, they can get its hex of `0xc02b` from the list, which `tls.CipherSuiteName()` will map to Go's `TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384`.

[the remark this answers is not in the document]
ihcsim: The other (less preferred) alternative is to maintain a mapping of the OpenSSL cipher to the Go's cipher list at https://pkg.go.dev/crypto/tls?utm_source=godoc#pkg-constants inside our code. It's not a long list, but I'll prefer not to have to do this.

pkg/controller/master/rancher/register.go
ibrokethecloud: cleanup is needed first since the vip needs to be freed from nginx service and mapped to traefik service

pkg/controller/master/rancher/register.go
ibrokethecloud: ideally if cluster is new.. there will be nothing to cleanup as nginx will never be installed

ihcsim (dismissed):

Vicente-Cheng (approved):
verified with the fresh installation and some basic operations.
lgtm, thanks!

Vicente-Cheng:
@mergifyio backport v1.9

Issue #9336, opened 2025-10-20
[BUG] downstream cluster provisioning broken for non cluster admins when pcidevices is enabled


ibrokethecloud:
### Describe the Bug

Cluster provisioning breaks when `pcidevices` addon is enabled on downstream cluster and user is a project member only.

The error seems to arise from the fact that during cluster provisioning the UI checks for the following
* pcidevices addon is enabled
* if enabled checks for vgpudevice list, which is used to populate the advanced section

vgpudevices are a cluster level resource. For users without cluster level access, the `vgpudevice` list fails and returns a 404.

This breaks the UI

<img width="1342" height="532" alt="Image" src="https://github.com/user-attachments/assets/a68b0160-3bd8-4bdb-924a-09e35131d20f" />

failing call

<img width="475" height="204" alt="Image" src="https://github.com/user-attachments/assets/ef96f1e8-5b0c-47c7-921f-71aa9b325abf" />

### To Reproduce

To reproduce

* Install Rancher v2.11.0 / v2.12.1
* Import a Harvester v1.6.0 cluster
* Enable PCIdevices addon on Harvester
* Create a local user named `demo` to Rancher
* Add `demo` user to `Default` Project in Harvester cluster
* Login as Demo and try and create a downstream cluster using the Harvester provider on the Harvester cluster from previous step

### Expected Behavior

the 404 error should be handled gracefully

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.6.0
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version: v2.12.0

### Workaround and Mitigation

Create a custom cluster role `list-vgpu-devices` in Rancher

<img width="2235" height="651" alt="Image" src="https://github.com/user-attachments/assets/0c37aae3-9d4a-45cb-90bf-dbf3ff4f9ce3" />

Assign users this role at Harvester cluster level

After re-login the error should not appear in the UI

ibrokethecloud:
Able to reproduce issue on both v2.11.0 and v2.12.1

a110605:
HI @ibrokethecloud, 
[...]

PR #10078, opened 2026-02-22
add extra permissions to harvester-authenticated role to allow users …

ibrokethecloud:
#### Problem:

standard rancher users while trying to provision downstream clusters on harvester see the following error on the cluster provisioning form

image

Based on feedback from @a110605  the issue is being caused because standard users do not have permissions to list `system-namespaces` management.cattle.io setting from the underlying harvester cluster

#### Solution:

A simple fix introduced in the PR is to update `harvester-authenticated` cluster role to include the extra permissions to `read` and `watch` the `system-namespaces` setting.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/9336

### DIFF
--- a/pkg/data/public.go
+++ b/pkg/data/public.go
@@ -7,7 +7,8 @@ import (
 )
 
 const (
-	publicNamespace = "harvester-public"
+	publicNamespace  = "harvester-public"
+	systemNamespaces = "system-namespaces"
 )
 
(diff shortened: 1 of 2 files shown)


ihcsim (approved):

brandboat (approved):
LGTM, thanks!

Issue #9412, opened 2025-11-03
[FEATURE] Apply CDI Settings To Non-Root Volumes


ihcsim:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

The CDI settings exposed via the Harvester CDI annotations only work with virtual machine image and root disk volumes. They don't work with non-root volumes (with `Source` set to `New`) created directly via the UI, because these volumes have no associated underlying CDI data volumes. (By default, the CDI storage profiles only apply to CDI data volumes, not PVCs.) This limitation prevents the creation of non-root disk volumes based on 3rd-party CSI solutions that rely on the Harvester CDI annotations to configure the CDI storage profiles.

### Describe the solution you'd like

Extend the Harvester CDI settings to non-root disk volumes using CDI [PVC rendering](https://github.com/kubevirt/containerized-data-importer/blob/4c3dc65a772e0cd66fdbabf75c26b688f4b4afb9/doc/pvc-mutating-webhook-rendering.md). This involves:

* Update the `cdi` custom resource with the `WebhookPvcRendering` feature gate
* Label opt-in PVCs with the `cdi.kubevirt.io/applyStorageProfile: "true"` label
* Update 'Edit as YAML' option in the 'Create Volume' page to not assume `ReadWriteMany` is the default access mode - it should just be blank

### Describe alternatives you've considered

Don't support non-CDI non-root volumes or let user configure the CDI PVC rendering manually, both of which seem unreasonable.

While certain CDI attributes like the PVC access modes can be configured via the volume 'Edit as YAML' option, others like clone strategy, snapshot class etc. are not.

innobead:
@harvester/release ~moved this to 1.8.0.~ Need to clarify the detailed use cases first. cc @Vicente-Cheng

Vicente-Cheng:
related to #8285

innobead:
Moved to 1.9.0 first due to limited time for the upcoming 1.8.0. However, everything is on track, so we will see if we can move this back to 1.8.0 if bandwidth is available. 

cc @harvester/release

Vicente-Cheng:
I will update the HEP before next week

lanfon72:
move back to review because:
1. no HEP refereed.
2. `Pre Ready-For-Testing Checklist` not update

Vicente-Cheng:
> move back to review because:
> 
> 1. no HEP refereed.
> 2. `Pre Ready-For-Testing Checklist` not update

[...]

Issue #8285, opened 2025-05-14
[FEATURE] In-Place Volume live migration to a different storage provider (Storage Migration)


forbesguthrie:
Request description:
As a customer, I want to be able to move a VM from one volume to another, or just a subset of a VM's disks. The source and destination could be Longhorn Volumes or an externally supported Volume, i.e. iSCSI/FC backed volume via CSI. In the future, when we add support for File-based storage (e.g. NFSv3) this should also work.  The VM will remain on the same host during this operation.

Actual behavior:
Currently, customers cannot migrate a VM's disks while the VM is running

Expected behavior:
The VM should remain turned on and the guest OS should be unaware of the underlying disk movement. 

The impact on the workload would be minimal, much like a Live Migration.

The performance of the Guest OS can be reduced during the operation, but not so much that disk errors are seen by the Guest OS.

Customers should be able to initiate this via the UI and API call.

There should be approximate feedback on the progress of the operation (% complete, and/or time taken/remaining).

We should block other VM operations during this time as appropriate, i.e. live migration from host to host, until complete.

We should check disk capacity is available on Destination (and keep checking during transfer)

Stretch goal should be to make it easy in the UI to move multiple VMs between volumes with one workflow. For example a customer needs to remove all VMs on a Volume to another volume.

Workaround:
Is a workaround available and implemented? yes/no
What is the workaround:

Additional notes:
Possible approach: https://kubevirt.io/user-guide/storage/volume_migration/

innobead:
This might be covered in harvester/suse-virtualization-mgmt#18 

cc @harvester/storage @ibrokethecloud

lanfon72:
Verified there the feature not working on restored VM, move back to implement.
Both snapshot and backup will create new volume for the VM, but no annotation added.

<img width="832" height="575" alt="Image" src="https://github.com/user-attachments/assets/6f00acb2-e989-4564-9ce1-86366b19edbc" />

But user can still Edit the VM then Save (without any modification), the annotation will be added back.

Test Information
----
* Environment: **Baremetal DL360 4 nodes**
* Harvester Version: **v1.8.0-rc2**
* **ui-source** Option: **Auto**

### Verify Steps
**Prerequisites**
1. Install Harvester with at least 2 nodes and having additional disks
2. Enable 3rd party Storage (Longhorn V2) in _Advanced/Settings_
3. Navigate to hosts page and edit nodes to add storage provisioned by Longhorn V2
[...]

PR #10086, opened 2026-02-23
HEP: Support Online Storage Migration [skip ci]

Vicente-Cheng:
#### Problem:
HEP for Storage Migration

#### Solution:
HEP for Storage Migration

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8285


enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### Story 2: Migrate to third-party storage
+
+A user has integrated a third-party storage solution (e.g., Ceph, NetApp) into Harvester. They want to move existing VM volumes to leverage the new storage backend without downtime.
+
+### User Experience In Detail
+
+1. The user create the target volume for migration, which is a new PVC with the desired StorageClass.
+2. The user navigates to the VM detail page and selects a volume to migrate.
+3. The system validates that the migration is feasible (e.g., sufficient capacity in the target storage).
```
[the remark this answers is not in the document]
Vicente-Cheng: The dst volume should be bigger (at least equal to) than the source volume, so we could skip the runtime monitor.

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,155 @@
...
+        "storageClassName": "longhorn",
+        "volumeMode": "Block"
+      }
+    }
+  }
+]
+```
+
+- `pvc`: The original PVC spec for the volume, same as the existing format.
+- `replacement` (optional): The name of the new PVC that the volume should be migrated to. When this field is set, the system will trigger a storage migration for this volume during VM migration. If not set, the volume remains unchanged.
```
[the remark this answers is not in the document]
Vicente-Cheng: left for manual cleanup

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,155 @@
...
+## Design
+
+### High-level Architecture
+
+![storage migration Architecture](./20260224-storage-migration/architecture.png)
+
+### Implementation Overview
+
+1. **New API Field**: Extend the `harvesterhci.io/volumeClaimTemplates` annotation to include an optional `replacement` field that specifies the target PVC for migration.
+2. **VM Controller**: Enhance the VM controller to detect when a volume has a `replacement` specified. When such a volume is detected, the controller will initiate a storage migration process during VM migration.
```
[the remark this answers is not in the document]
Vicente-Cheng: because we leverage the VM migration for the storage migration, at this moment, most VM operations would not be accepted.

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,155 @@
...
+
+The VM Controller will update the following fields in the VM spec:
+- Update the `volumes` to reference the new PVC specified in the `replacement` field. (Prefer the `dataVolume` field if the new PVC is created as a DataVolume, otherwise fallback to `persistentVolumeClaim`).
+- Update the `updateVolumesStrategy` to `Migration` to trigger the migration process.
+
+3. **DataVolume Integration**: Leverage CDI's DataVolume as the top layer for volume management. The migration would use a new DataVolume that references the target PVC. The DataVolume introduces an abstraction layer that allows for seamless data copying (cloning) between different storage backends.
```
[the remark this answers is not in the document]
Vicente-Cheng: I will handle this one later, keep here

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,157 @@
+# Storage Migration
```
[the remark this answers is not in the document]
Vicente-Cheng: we will also provide the document for the offline storage migration, so keep the title here

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,157 @@
...
+- Provide documentation describing the recommended offline (VM powered-off) storage migration procedure, without introducing new offline migration features.
+- Add DataVolume layer on top of PersistentVolumeClaim to support various storage vendors and enable storage migration.
+
+### Non-goals [optional]
+
+- Cross-cluster storage migration is not in scope.
+
+## Proposal
+
+Introduce a storage migration mechanism that allows users to migrate VM volumes between different storage classes. The migration process will leverage Kubevirt's volume migration mechanism. The data movement will be handled during the VM migration, so the ETA is basically the same as the VM migration but with the additional data copy inside.
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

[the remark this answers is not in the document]
Vicente-Cheng: updated

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,157 @@
...
+
+A user has a VM running on LVM local storage. They now need to enable live migration for the VM, which requires replicated storage. The user wants to migrate the volume to Longhorn replicated storage so the VM can be live-migrated across nodes.
+
+#### Story 2: Migrate to third-party storage
+
+A user has integrated a third-party storage solution (e.g., Ceph, NetApp) into Harvester. They want to move existing VM volumes to leverage the new storage backend without downtime.
+
+### User Experience In Detail
+
+1. The user creates the target volume for migration, which is a new PVC with the desired StorageClass.
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,157 @@
...
+        "storageClassName": "longhorn",
+        "volumeMode": "Block"
+      }
+    }
+  }
+]
+```
+
+- `pvc`: The original PVC spec for the volume, same as the existing format.
+- `replacement` (optional): The name of the new PVC that the volume should be migrated to. When this field is set, the system will trigger a storage migration for this volume during VM migration. If not set, the volume remains unchanged.
```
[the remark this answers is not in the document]
Vicente-Cheng: >What happens if multiple volumes have replacement fields set simultaneously?

We will update just once, so for multiple volumes, once VM live migration can handle all storage migration.

>Can migrations be queued, or must they be done one at a time?

cannot queued

>What prevents concurrent updates to the annotation from multiple sources (UI, API, automation)?

If the VM is under migration (which mean take effect), other update will be block

>How is the migration state tracked if the annotation is the sole source of truth?

We will check the VM migration condition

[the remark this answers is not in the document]
Vicente-Cheng: we should consider moving the whole annotation to CRD in a later version

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,157 @@
...
+
+The VM Controller will update the following fields in the VM spec:
+- Update `spec.template.spec.volumes` to reference the new storage backend. The controller MUST:
+  - Prefer the `dataVolume` volume source when the migration flow has created or associated a DataVolume for the `replacement` PVC (for example, when a DataVolume CR exists that owns or references the target PVC), and set `volume.dataVolume.name` to that DataVolume name.
+  - Fallback to the `persistentVolumeClaim` volume source (setting `volume.persistentVolumeClaim.claimName` to the `replacement` PVC name) when no such DataVolume is present for the target volume.
+- Update the `updateVolumesStrategy` to `Migration` to trigger the migration process.
+
+3. **DataVolume Integration**: Leverage CDI's DataVolume as the top layer for volume management. The migration would use a new DataVolume that references the target PVC. The DataVolume introduces an abstraction layer that allows for seamless data copying (cloning) between different storage backends.
+
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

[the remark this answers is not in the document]
Vicente-Cheng: snapshot would not consider this case, we need to update the limitation of snapshot
I will update later (backup also)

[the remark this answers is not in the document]
Vicente-Cheng: We would not have progress at this moment, just rely on the VM live migration related CRD/conditions

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,157 @@
...
+- (Option 2) Use UI to trigger the migration by selecting the target PVC for the volume.
+- Verify that the migration process starts and completes successfully.
+
+#### OffLine Migration Tests
+
+- Create VM with a volume on the source storage class.
+- Stop the VM to simulate offline migration.
+- Clone source PVC to a new PVC with the target storage class using DataVolume.
+- Update VM template to reference the new PVC.
+- Start the VM and verify it uses the new volume successfully.
```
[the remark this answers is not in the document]
Vicente-Cheng: yeah, we should create an separate issue for offline document

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Story 1: Migrate from local storage to replicated storage for HA
+
+A user has a VM running on LVM local storage. They now need to enable live migration for the VM, which requires replicated storage. The user wants to migrate the volume to Longhorn replicated storage so the VM can be live-migrated across nodes.
+
+#### Story 2: Migrate to third-party storage
+
+A user has integrated a third-party storage solution (e.g., Ceph, NetApp) into Harvester. They want to move existing VM volumes to leverage the new storage backend without downtime.
+
+### User Experience In Detail
```
ihcsim, line 39: I think the UI should:

* indicate that storage migration is progressing
* disable certain functions like stopping the VM
* report the `ManualRecoveryCondition` condition

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,174 @@
...
+
+1. The user creates the target volume for migration, which is a new PVC with the desired StorageClass.
+2. The user navigates to the VM detail page and selects a volume to migrate.
+3. The system validates that the migration is feasible (e.g., sufficient capacity in the target storage).
+4. The user confirms the migration, and the system begins the migration process by initiating a VM live migration while copying the volume data in the background.
+5. Once the migration is complete, the VM automatically uses the new volume.
```
ihcsim, line 45: Does KubeVirt automatically remove the old volume? We should include in the workflow the chance for user to take a snapshot or backup of the old volume, before the migration starts.

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,174 @@
...
+        },
+        "storageClassName": "longhorn",
+        "volumeMode": "Block"
+      }
+    }
+  }
+]
+```
+
+- `pvc`: The original PVC spec for the volume, same as the existing format.
```
ihcsim: In this example, does `pvc` remain `my-vol` after the migration? Or should it be updated to `new-pvc-name`? What happens if `new-pvc-name` is then migrated to `new-pvc-name2`?

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,174 @@
...
+- `targetVolume` (optional): The name of the new PVC that the volume should be migrated to. When this field is set, the system will trigger a storage migration for this volume during VM live migration. If not set, the volume remains unchanged.
```
ihcsim, line 89: Not certain but I feel like this should be a separate annotation. The `harvesterhci.io/volumeClaimTemplates` annotation should only reflect the VM's _current_ volume. When the `targetVolume` annotation is set, it triggers a storage migration and the existing `harvesterhci.io/volumeClaimTemplates` annotation should be completely replaced to reflect the new volume configuration.

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Implementation Overview
+
+1. **New API Field**: Extend the `harvesterhci.io/volumeClaimTemplates` annotation to include an optional `targetVolume` field that specifies the target PVC for migration.
+2. **VM Controller**: Enhance the VM controller to detect when a volume has a `targetVolume` specified. When such a volume is detected, the controller will initiate a storage migration process during VM live migration.
+
+The VM Controller will update the following fields in the VM spec:
+- Update `spec.template.spec.volumes` to reference the new storage backend. The controller MUST:
+  - Prefer the `dataVolume` volume source when the migration flow has created or associated a DataVolume for the `targetVolume` PVC (for example, when a DataVolume CR exists that owns or references the target PVC), and set `volume.dataVolume.name` to that DataVolume name.
+  - Fallback to the `persistentVolumeClaim` volume source (setting `volume.persistentVolumeClaim.claimName` to the `targetVolume` PVC name) when no such DataVolume is present for the target volume.
```
ihcsim, line 105: Is the fallback included simply to support existing pre-1.8 volumes? Personally, I think online volume migration should just be a 1.8 and newer features, requiring data volumes. The fallback can be added later if there are enough requests for it. My thinking is to simplify the implementation without it being bogged down by legacy decision.

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,174 @@
...
+- Update the `updateVolumesStrategy` to `Migration` to trigger the migration process.
```
ihcsim, line 106: I think this was asked during the meeting, but to confirm, will this trigger the "restart required" condition on the VM?

enhancements/20260224-storage-migration.md
```diff
@@ -0,0 +1,174 @@
...
+## Motivation
+
+### Goals
+
+- Allow users to migrate VM volumes from one StorageClass to another.
+- Implement support for online storage migration only; offline storage migration behavior remains unchanged.
+- Provide documentation describing the recommended offline (VM powered-off) storage migration procedure, without introducing new offline migration features.
+- Add DataVolume layer on top of PersistentVolumeClaim to support various storage vendors and enable storage migration.
+
+### Non-goals [optional]
```
ihcsim, line 21: We should also mention that live migration of hotplug volumes (i.e. guest cluster workloads volumes) are not supported per https://kubevirt.io/user-guide/storage/volume_migration/#limitations.

enhancements/20260224-storage-migration.md
Vicente-Cheng: >Does KubeVirt automatically remove the old volume?

No, User need to manually remove the old volume

>We should include in the workflow the chance for user to take a snapshot or backup of the old volume, before the migration starts.

We will have the document to tell the user they should take a snapshot or a backup first.

enhancements/20260224-storage-migration.md
Vicente-Cheng: We will update the annotation after the migration is complete and clean up the targetVolume.
Also, we will block the annotation update during the migration

enhancements/20260224-storage-migration.md
ibrokethecloud: not entirely related to the HEP itself, but if we make a UI change to ensure volumes page creates volumes via the datavolume interface, then we can inherit all CDI settings on 3rd party CSI.

WebberHuang1118:
Hi @Vicente-Cheng,

I have a few questions:

* After a storage migration succeeds—either the offline method in the documentation or the online method via VM migration—how do we plan to handle the original volume? Will the controller delete it, or should we let the user decide whether to keep it?
* If the storage migration is performed on the guest-cluster node VM, IIUC this could break the Harvester CSI driver. Should we block this scenario, or document it as a known limitation?

Please take a look—thanks.

Vicente-Cheng:
> Hi @Vicente-Cheng,
> 
> I have a few questions:
> 
> * After a storage migration succeeds—either the offline method in the documentation or the online method via VM migration—how do we plan to handle the original volume? Will the controller delete it, or should we let the user decide whether to keep it?

Users should handle it themselves.

> * If the storage migration is performed on the guest-cluster node VM, IIUC this could break the Harvester CSI driver. Should we block this scenario, or document it as a known limitation?
> 

It's hard to block a user from using it if the hotplug volume comes from the guest cluster.
~We can handle the better UX on the guest cluster on v1.9.0.
For v1.8.0, we could provide the document if the user wants to migrate their guest cluster volume.~
I will test some scenario and update the limitation with guest cluster

> Please take a look—thanks.

enhancements/20260224-storage-migration.md
Vicente-Cheng: yes, we will update this annotation again once the storage migration completed

enhancements/20260224-storage-migration.md
Vicente-Cheng: I am ok with that, we can only support the volume which is under the DataVolume layer.

enhancements/20260224-storage-migration.md
Vicente-Cheng: No, it would not trigger the restart required condition

enhancements/20260224-storage-migration.md
Vicente-Cheng: sure, it's a good suggestion!

WebberHuang1118:
> > Hi @Vicente-Cheng,
> > I have a few questions:
> > 
> > * After a storage migration succeeds—either the offline method in the documentation or the online method via VM migration—how do we plan to handle the original volume? Will the controller delete it, or should we let the user decide whether to keep it?
> 
> Users should handle it themselves.
> 
> > * If the storage migration is performed on the guest-cluster node VM, IIUC this could break the Harvester CSI driver. Should we block this scenario, or document it as a known limitation?
> 
> It's hard to block a user from using it if the hotplug volume comes from the guest cluster. ~We can handle the better UX on the guest cluster on v1.9.0. For v1.8.0, we could provide the document if the user wants to migrate their guest cluster volume.~ I will test some scenario and update the limitation with guest cluster
> 
> > Please take a look—thanks.

Having documentation covering these two points should be sufficient.

WebberHuang1118 (dismissed):

enhancements/20260224-storage-migration.md
Vicente-Cheng: >indicate that storage migration is progressing

Now storage migration leverages the normal VM migration and low-level device mirror, and it would not have any progress for it.
So at this moment, we were having a hard time making real progress for it.

>disable certain functions like stopping the VM

I think we still need to give user the way to reset (cancel) this.
Stopping and resetting the VM would be another way, so I suggest we should keep the operation as we have now.
WDYT?

> report the ManualRecoveryCondition condition
Yeah, we should. Updated!

ihcsim (approved):

brandboat (approved):

ibrokethecloud (approved):

PR #10088, opened 2026-02-23
Support Online Storage Migration

Vicente-Cheng:
#### Problem:
Support Online Storage Migration

#### Solution:
Support Online Storage Migration

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8285

### DIFF
--- a/pkg/api/vm/formatter.go
+++ b/pkg/api/vm/formatter.go
@@ -53,6 +53,8 @@ const (
 	updateResourceQuotaAction        = "updateResourceQuota"
 	deleteResourceQuotaAction        = "deleteResourceQuota"
 	cpuAndMemoryHotplug              = "cpuAndMemoryHotplug"
+	storageMigration                 = "storageMigration"
+	cancelStorageMigration           = "cancelStorageMigration"
 )
 
 type vmformatter struct {
@@ -130,6 +132,7 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	if canMigrate(vf.nodeCache, vmi) {
 		resource.AddAction(request, migrate)
 		resource.AddAction(request, findMigratableNodes)
+		resource.AddAction(request, storageMigration)
 
 		if canCPUAndMemoryHotplug(vm) {
 			resource.AddAction(request, cpuAndMemoryHotplug)
@@ -172,6 +175,23 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	if canDismissInsufficientResourceQuota(vm) {
 		resource.AddAction(request, dismissInsufficientResourceQuota)
 	}
+
+	if hasActiveStorageMigration(vm) {
+		resource.AddAction(request, cancelStorageMigration)
+	}
+}
+
+func hasActiveStorageMigration(vm *kubevirtv1.VirtualMachine) bool {
+	entries, err := util.UnmarshalVolumeClaimTemplates(vm.Annotations[util.AnnotationVolumeClaimTemplates])
+	if err != nil {
+		return false
+	}
+	for _, entry := range entries {
+		if entry.TargetVolume != "" {
+			return true
+		}
+	}
+	return false
 }
 
 func (vf *vmformatter) canPause(vmi *kubevirtv1.VirtualMachineInstance) bool {
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -476,17 +476,17 @@ func appendVolumeClaimTemplatesFromVMAnnotation(vm *kubevirtv1.VirtualMachine, p
 	if !ok {
 		return nil
 	}
-	volumeClaimTemplates := make([]corev1.PersistentVolumeClaim, 0, 1)
-	if err := json.Unmarshal([]byte(volumeClaimTemplatesStr), &volumeClaimTemplates); err != nil {
-		return err
-	}
-
-	volumeClaimTemplates = append(volumeClaimTemplates, pvc)
-	updateVolumeClaimTemplateBytes, err := json.Marshal(volumeClaimTemplates)
+	entries, err := util.UnmarshalVolumeClaimTemplates(volumeClaimTemplatesStr)
 	if err != nil {
 		return err
 	}
-	vm.Annotations[util.AnnotationVolumeClaimTemplates] = string(updateVolumeClaimTemplateBytes)
+
+	entries = append(entries, util.VolumeClaimTemplateEntry{PersistentVolumeClaim: pvc})
+	data, err := util.MarshalVolumeClaimTemplates(entries)
+	if err != nil {
+		return err
+	}
+	vm.Annotations[util.AnnotationVolumeClaimTemplates] = data
 	return nil
 }
 
@@ -495,20 +495,21 @@ func removeVolumeClaimTemplatesFromVMAnnotation(vm *kubevirtv1.VirtualMachine, t
 	if !ok {
 		return nil
 	}
-	var volumeClaimTemplates, toUpdateVolumeClaimTemplates []corev1.PersistentVolumeClaim
-	if err := json.Unmarshal([]byte(volumeClaimTemplatesStr), &volumeClaimTemplates); err != nil {
-		return err
-	}
-	for _, volumeClaimTemplate := range volumeClaimTemplates {
-		if !slice.ContainsString(toRemoveDiskNames, volumeClaimTemplate.Name) {
-			toUpdateVolumeClaimTemplates = append(toUpdateVolumeClaimTemplates, volumeClaimTemplate)
-		}
-	}
-	toUpdateVolumeClaimTemplateBytes, err := json.Marshal(toUpdateVolumeClaimTemplates)
+	entries, err := util.UnmarshalVolumeClaimTemplates(volumeClaimTemplatesStr)
 	if err != nil {
 		return err
 	}
-	vm.Annotations[util.AnnotationVolumeClaimTemplates] = string(toUpdateVolumeClaimTemplateBytes)
+	var toUpdate []util.VolumeClaimTemplateEntry
+	for _, entry := range entries {
+		if !slice.ContainsString(toRemoveDiskNames, entry.Name) {
+			toUpdate = append(toUpdate, entry)
+		}
+	}
+	data, err := util.MarshalVolumeClaimTemplates(toUpdate)
+	if err != nil {
+		return err
+	}
+	vm.Annotations[util.AnnotationVolumeClaimTemplates] = data
 	return nil
 }
 
@@ -1227,12 +1228,16 @@ func (h *vmActionHandler) cloneVM(name string, namespace string, input CloneInpu
 	if err != nil {
 		return fmt.Errorf("clone volumes error for new vm %s/%s, err %w", newVM.Namespace, newVM.Name, err)
 	}
-	newPVCsString, err := json.Marshal(newPVCs)
+	newEntries := make([]util.VolumeClaimTemplateEntry, len(newPVCs))
+	for i, pvc := range newPVCs {
+		newEntries[i] = util.VolumeClaimTemplateEntry{PersistentVolumeClaim: pvc}
+	}
+	newPVCsString, err := util.MarshalVolumeClaimTemplates(newEntries)
 	if err != nil {
-		return fmt.Errorf("cannot marshal value %+v, err: %w", newPVCs, err)
+		return fmt.Errorf("cannot marshal value %+v, err: %w", newEntries, err)
 	}
 
-	newVM.ObjectMeta.Annotations[util.AnnotationVolumeClaimTemplates] = string(newPVCsString)
+	newVM.ObjectMeta.Annotations[util.AnnotationVolumeClaimTemplates] = newPVCsString
 	if newVM, err = h.vmClient.Create(newVM); err != nil {
 		return fmt.Errorf("cannot create new VM %s/%s, err: %w", newVM.Namespace, newVM.Name, err)
 	}
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -44,7 +44,6 @@ import (
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/builder"
 	nodecontroller "github.com/harvester/harvester/pkg/controller/master/node"
-	ctlcdiv1 "github.com/harvester/harvester/pkg/generated/controllers/cdi.kubevirt.io/v1beta1"
 	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctlcniv1 "github.com/harvester/harvester/pkg/generated/controllers/k8s.cni.cncf.io/v1"
 	ctlkubevirtv1 "github.com/harvester/harvester/pkg/generated/controllers/kubevirt.io/v1"
@@ -78,7 +77,6 @@ type vmActionHandler struct {
 	vmio                      vmicommon.VMIOperator
 
 	backupClient            ctlharvesterv1.VirtualMachineBackupClient
-	datavolumeClient        ctlcdiv1.DataVolumeClient
 	pvcClient               ctlcorev1.PersistentVolumeClaimClient
 	resourceQuotaClient     ctlharvesterv1.ResourceQuotaClient
 	restoreClient           ctlharvesterv1.VirtualMachineRestoreClient
@@ -300,6 +298,20 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
 			return nil, apierror.NewAPIError(validation.InvalidBodyContent, "Failed to decode request body: %v "+err.Error())
 		}
 		return nil, h.cpuAndMemoryHotplug(namespace, name, input)
+	case storageMigration:
+		var input StorageMigrationInput
+		if err := json.NewDecoder(r.Body).Decode(&input); err != nil {
+			return nil, apierror.NewAPIError(validation.InvalidBodyContent, "Failed to decode request body: %v "+err.Error())
+		}
+		if input.SourceVolume == "" || input.TargetVolume == "" {
+			return nil, apierror.NewAPIError(validation.InvalidBodyContent, "Parameter `sourceVolume` and `targetVolume` are required")
+		}
+		if input.SourceVolume == input.TargetVolume {
+			return nil, apierror.NewAPIError(validation.InvalidBodyContent, "sourceVolume and targetVolume must be different")
+		}
+		return nil, h.storageMigration(namespace, name, input)
+	case cancelStorageMigration:
+		return nil, h.cancelStorageMigration(namespace, name)
 	default:
 		return nil, apierror.NewAPIError(validation.InvalidAction, "Unsupported action")
 	}
@@ -1539,4 +1551,100 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
 	patchData := fmt.Sprintf("[%s]", strings.Join(patchOps, ","))
 	_, err := h.vmClient.Patch(namespace, name, k8stypes.JSONPatchType, []byte(patchData))
 	return err
+}
+
+func (h *vmActionHandler) storageMigration(namespace, name string, input StorageMigrationInput) error {
+	vm, err := h.vmCache.Get(namespace, name)
+	if err != nil {
+		return err
+	}
+
+	annStr := vm.Annotations[util.AnnotationVolumeClaimTemplates]
+	entries, err := util.UnmarshalVolumeClaimTemplates(annStr)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal volumeClaimTemplates annotation: %w", err)
+	}
+
+	found := false
+	for i, entry := range entries {
+		if entry.Name == input.SourceVolume {
+			entries[i].TargetVolume = input.TargetVolume
+			found = true
+			break
+		}
+	}
+	if !found {
+		return fmt.Errorf("source volume %s not found in volumeClaimTemplates annotation", input.SourceVolume)
+	}
+
+	newAnn, err := util.MarshalVolumeClaimTemplates(entries)
+	if err != nil {
+		return fmt.Errorf("failed to marshal volumeClaimTemplates annotation: %w", err)
+	}
+
+	vmCopy := vm.DeepCopy()
+	vmCopy.Annotations[util.AnnotationVolumeClaimTemplates] = newAnn
+	_, err = h.vmClient.Update(vmCopy)
+	return err
+}
+
+func (h *vmActionHandler) cancelStorageMigration(namespace, name string) error {
+	vm, err := h.vmCache.Get(namespace, name)
+	if err != nil {
+		return err
+	}
+
+	annStr := vm.Annotations[util.AnnotationVolumeClaimTemplates]
+	entries, err := util.UnmarshalVolumeClaimTemplates(annStr)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal volumeClaimTemplates annotation: %w", err)
+	}
+
+	// Build a map of targetVolume -> sourceVolume from entries that have active migrations
+	targetToSource := map[string]string{}
+	for i, entry := range entries {
+		if entry.TargetVolume != "" {
+			targetToSource[entry.TargetVolume] = entry.Name
+			entries[i].TargetVolume = ""
+		}
+	}
+	if len(targetToSource) == 0 {
+		return fmt.Errorf("no active storage migration found")
+	}
+
+	newAnn, err := util.MarshalVolumeClaimTemplates(entries)
+	if err != nil {
+		return fmt.Errorf("failed to marshal volumeClaimTemplates annotation: %w", err)
+	}
+
+	vmCopy := vm.DeepCopy()
+	vmCopy.Annotations[util.AnnotationVolumeClaimTemplates] = newAnn
+
+	// Restore VM spec volumes that were swapped to target back to source
+	for i, vol := range vmCopy.Spec.Template.Spec.Volumes {
+		if vol.PersistentVolumeClaim == nil {
+			continue
+		}
+
+		sourceName, ok := targetToSource[vol.PersistentVolumeClaim.ClaimName]
+		if !ok {
+			continue
+		}
+
+		vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = kubevirtv1.VolumeSource{
+			PersistentVolumeClaim: &kubevirtv1.PersistentVolumeClaimVolumeSource{
+				PersistentVolumeClaimVolumeSource: corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: sourceName,
+				},
+				Hotpluggable: vol.PersistentVolumeClaim.Hotpluggable,
+			},
+		}
+	}
+
+	delete(vmCopy.Annotations, util.AnnotationWaitingStorageMigration)
+	vmCopy.Spec.UpdateVolumesStrategy = nil
+	if _, err = h.vmClient.Update(vmCopy); err != nil {
+		return err
+	}
+	return nil
 }
--- a/pkg/api/vm/handler_test.go
+++ b/pkg/api/vm/handler_test.go
@@ -2,11 +2,11 @@ package vm
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"testing"
 	"time"
 
+	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
@@ -22,7 +22,6 @@ import (
 	"github.com/harvester/harvester/pkg/generated/clientset/versioned/fake"
 	"github.com/harvester/harvester/pkg/util"
 	"github.com/harvester/harvester/pkg/util/fakeclients"
-	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 )
 
 func TestMigrateAction(t *testing.T) {
@@ -1492,11 +1491,10 @@ func TestInsertCdRomVolumeAction(t *testing.T) {
 
 	assert.Len(t, vmUpdated.Spec.Template.Spec.Volumes, 1, "Should have hot-plug new volume")
 
-	var pvcs []corev1.PersistentVolumeClaim
 	volumeClaimTemplates := vmUpdated.ObjectMeta.Annotations[util.AnnotationVolumeClaimTemplates]
-	err = json.Unmarshal([]byte(volumeClaimTemplates), &pvcs)
+	entries, err := util.UnmarshalVolumeClaimTemplates(volumeClaimTemplates)
 	assert.Nil(t, err, "Should unmarshal volumeClaimTemplates annotation correctly")
-	assert.Len(t, pvcs, 1, "Should add to volumeClaimTemplates annotation")
+	assert.Len(t, entries, 1, "Should add to volumeClaimTemplates annotation")
 }
 
 func TestEjectCdRomVolumeAction(t *testing.T) {
@@ -1581,11 +1579,10 @@ func TestEjectCdRomVolumeAction(t *testing.T) {
 
 	assert.Len(t, vmUpdated.Spec.Template.Spec.Volumes, 0, "Should have hot-unplug volume")
 
-	var pvcs []corev1.PersistentVolumeClaim
 	volumeClaimTemplates := vmUpdated.ObjectMeta.Annotations[util.AnnotationVolumeClaimTemplates]
-	err = json.Unmarshal([]byte(volumeClaimTemplates), &pvcs)
+	entries, err := util.UnmarshalVolumeClaimTemplates(volumeClaimTemplates)
 	assert.Nil(t, err, "Should unmarshal volumeClaimTemplates annotation correctly")
-	assert.Len(t, pvcs, 0, "Should remove from volumeClaimTemplates annotation")
+	assert.Len(t, entries, 0, "Should remove from volumeClaimTemplates annotation")
 
 	_, err = pvcCache.Get(pvcNamespace, pvcName)
 	assert.True(t, apierrors.IsNotFound(err), "Should delete pvc")
--- a/pkg/api/vm/handler_test.go
+++ b/pkg/api/vm/handler_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 	"time"
 
-	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
@@ -22,6 +21,7 @@ import (
 	"github.com/harvester/harvester/pkg/generated/clientset/versioned/fake"
 	"github.com/harvester/harvester/pkg/util"
 	"github.com/harvester/harvester/pkg/util/fakeclients"
+	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 )
 
 func TestMigrateAction(t *testing.T) {
--- a/pkg/controller/master/virtualmachine/vm_controller.go
+++ b/pkg/controller/master/virtualmachine/vm_controller.go
@@ -1,7 +1,6 @@
 package virtualmachine
 
 import (
-	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
@@ -63,14 +62,15 @@ func (h *VMController) createPVCsFromAnnotation(_ string, vm *kubevirtv1.Virtual
 	if vm == nil || vm.DeletionTimestamp != nil {
 		return nil, nil
 	}
-	volumeClaimTemplates, ok := vm.Annotations[util.AnnotationVolumeClaimTemplates]
-	if !ok || volumeClaimTemplates == "" {
+	volumeClaimTemplatesStr, ok := vm.Annotations[util.AnnotationVolumeClaimTemplates]
+	if !ok || volumeClaimTemplatesStr == "" {
 		return nil, nil
 	}
-	var pvcs []*corev1.PersistentVolumeClaim
-	if err := json.Unmarshal([]byte(volumeClaimTemplates), &pvcs); err != nil {
+	entries, err := util.UnmarshalVolumeClaimTemplates(volumeClaimTemplatesStr)
+	if err != nil {
 		return nil, err
 	}
+	pvcs := util.GetPVCsFromVolumeClaimTemplates(entries)
 
 	for _, pvcAnno := range pvcs {
 		imageName := ""
--- a/pkg/controller/master/virtualmachine/vm_controller.go
+++ b/pkg/controller/master/virtualmachine/vm_controller.go
@@ -501,6 +501,216 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
 	return nil
 }
 
+// SetTargetVolumeStrategy detects targetVolume entries in volumeClaimTemplates
+// and updates the VM spec to trigger KubeVirt volume migration.
+func (h *VMController) SetTargetVolumeStrategy(_ string, vm *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
+	if vm == nil || vm.DeletionTimestamp != nil || vm.Spec.Template == nil {
+		return vm, nil
+	}
+
+	volumeClaimTemplatesStr, ok := vm.Annotations[util.AnnotationVolumeClaimTemplates]
+	if !ok || volumeClaimTemplatesStr == "" {
+		return vm, nil
+	}
+
+	entries, err := util.UnmarshalVolumeClaimTemplates(volumeClaimTemplatesStr)
+	if err != nil {
+		return nil, err
+	}
+
+	// Collect target volumes: old PVC name -> new name
+	targetVolumes := map[string]string{}
+	for _, entry := range entries {
+		if entry.TargetVolume != "" {
+			targetVolumes[entry.Name] = entry.TargetVolume
+		}
+	}
+	if len(targetVolumes) == 0 {
+		return vm, nil
+	}
+
+	vmCopy := vm.DeepCopy()
+	for i, vol := range vmCopy.Spec.Template.Spec.Volumes {
+		oldName, hotpluggable := getVolumeBackingRef(vol)
+
+		// oldName is empty means the volume is not a PVC, skip it
+		if oldName == "" {
+			continue
+		}
+
+		newName, ok := targetVolumes[oldName]
+		if !ok {
+			continue
+		}
+
+		vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = buildPVCVolumeSource(newName, hotpluggable)
+	}
+
+	strategy := kubevirtv1.UpdateVolumesStrategyMigration
+	vmCopy.Spec.UpdateVolumesStrategy = &strategy
+
+	if reflect.DeepEqual(vm, vmCopy) {
+		return h.clearMigrationWait(vm)
+	}
+
+	// First round: set spec + waiting annotation
+	vmCopy.Annotations[util.AnnotationWaitingStorageMigration] = "true"
+	return h.vmClient.Update(vmCopy)
+}
+
+// getVolumeBackingRef returns the PVC name and hotpluggable flag if the volume is a PVC, otherwise returns ok=false.
+func getVolumeBackingRef(vol kubevirtv1.Volume) (name string, hotpluggable bool) {
+	if vol.PersistentVolumeClaim != nil {
+		return vol.PersistentVolumeClaim.ClaimName, vol.PersistentVolumeClaim.Hotpluggable
+	}
+	return "", false
+}
+
+func buildPVCVolumeSource(name string, hotpluggable bool) kubevirtv1.VolumeSource {
+	return kubevirtv1.VolumeSource{
+		PersistentVolumeClaim: &kubevirtv1.PersistentVolumeClaimVolumeSource{
+			PersistentVolumeClaimVolumeSource: corev1.PersistentVolumeClaimVolumeSource{
+				ClaimName: name,
+			},
+			Hotpluggable: hotpluggable,
+		},
+	}
+}
+
+func (h *VMController) clearMigrationWait(vm *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
+	if vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusMigrating {
+		if _, ok := vm.Annotations[util.AnnotationWaitingStorageMigration]; ok {
+			vmCopy := vm.DeepCopy()
+			delete(vmCopy.Annotations, util.AnnotationWaitingStorageMigration)
+			return h.vmClient.Update(vmCopy)
+		}
+		return vm, nil
+	}
+	// Not yet migrating, keep waiting
+	h.vmController.EnqueueAfter(vm.Namespace, vm.Name, 5*time.Second)
+	return vm, nil
+}
+
+// CleanupTargetVolumeAnnotation updates the volumeClaimTemplates annotation after volume migration completes.
+// Migration is considered complete when:
+// 1. The annotation entry has a non-empty targetVolume field
+// 2. The VM spec volume already points to the targetVolume name
+// 3. The VM does NOT have a VolumesChange condition (migration finished)
+// When all conditions are met, the entry's PVC fields are updated from the actual new PVC in the cluster.
+func (h *VMController) CleanupTargetVolumeAnnotation(_ string, vm *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
+	if vm == nil || vm.DeletionTimestamp != nil || vm.Spec.Template == nil {
+		return vm, nil
+	}
+
+	volumeClaimTemplatesStr, ok := vm.Annotations[util.AnnotationVolumeClaimTemplates]
+	if !ok || volumeClaimTemplatesStr == "" {
+		return vm, nil
+	}
+
+	// Skip cleanup if storage migration is waiting to start
+	if _, ok := vm.Annotations[util.AnnotationWaitingStorageMigration]; ok {
+		return vm, nil
+	}
+
+	// Wait until VM is back to Running and VolumesChange condition is gone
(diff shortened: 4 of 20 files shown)


pkg/util/volume.go
```diff
@@ -24,6 +24,58 @@ var (
...
+func UnmarshalVolumeClaimTemplates(data string) ([]VolumeClaimTemplateEntry, error) {
+	if data == "" {
+		return nil, nil
+	}
+
+	// Try new format first
+	var entries []VolumeClaimTemplateEntry
+	if err := json.Unmarshal([]byte(data), &entries); err == nil && len(entries) > 0 && entries[0].PVC.Name != "" {
+		return entries, nil
+	}
```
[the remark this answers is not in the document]
Vicente-Cheng: If the legacy format, `entries[0].PVC` would be empty, so `entries[0].PVC.Name` would be `""`
This is not the case.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -691,3 +696,66 @@ func (v *vmValidator) checkCdRomVolumeIsValid(vm *kubevirtv1.VirtualMachine) err
...
+				return werror.NewInvalidError(
+					fmt.Sprintf("targetVolume PVC %s/%s not found", vm.Namespace, entry.TargetVolume),
+					fieldPath,
+				)
+			}
+			return werror.NewInternalError(fmt.Sprintf("failed to get targetVolume PVC %s/%s: %v", vm.Namespace, entry.TargetVolume, err))
+		}
+
+		targetSize := targetPVC.Spec.Resources.Requests.Storage()
+		if targetSize.Cmp(*sourceSize) < 0 {
```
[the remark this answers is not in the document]
Vicente-Cheng: The sourceSize would not be nil. If no keys, it would return zero.

We have similar cases on func `checkPVCStorageRequestChange()`

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -691,3 +696,66 @@ func (v *vmValidator) checkCdRomVolumeIsValid(vm *kubevirtv1.VirtualMachine) err
...
+			return werror.NewInvalidError(
+				fmt.Sprintf("targetVolume %s size (%s) must not be less than source volume %s size (%s)",
+					entry.TargetVolume, targetSize.String(), entry.PVC.Name, sourceSize.String()),
+				fieldPath,
+			)
+		}
+	}
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

pkg/api/vm/handler_test.go
```diff
@@ -2,11 +2,11 @@ package vm
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"testing"
 	"time"
 
+	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

pkg/util/volume.go
```diff
@@ -24,6 +24,58 @@ var (
 	PersistentVolumeClaimsKind = schema.GroupVersionKind{Group: "", Version: "v1", Kind: "PersistentVolumeClaim"}
 )
 
+// VolumeClaimTemplateEntry represents a single entry in the volumeClaimTemplates annotation.
+// New format wraps PVC with an optional targetVolume field for storage migration.
+type VolumeClaimTemplateEntry struct {
+	PVC          corev1.PersistentVolumeClaim `json:"pvc"`
+	TargetVolume string                       `json:"targetVolume,omitempty"`
+}
```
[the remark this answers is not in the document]
Vicente-Cheng: already have on another PR

Vicente-Cheng:
we are lacking the UI part, but for the backend controller. It's ready for review.

w13915984028:
Thanks for the PR.

The first question:

As this feature will be on top of vm migration, the current migration has UI render to hide the menu and api action handler to double check.

https://github.com/harvester/harvester/blob/014fbeae150e0c8b7148bd65cf73f40bc8a0af22/pkg/api/vm/formatter.go#L268

https://github.com/harvester/harvester/blob/014fbeae150e0c8b7148bd65cf73f40bc8a0af22/pkg/api/vm/handler.go#L515

How will storage migration collaborate with existing migrate/hot-plug?

e.g.  a current migration is ongoing, a single-node cluster, a non-migratable vm ...

pkg/util/volume.go
```diff
@@ -24,6 +24,58 @@ var (
...
+
+// UnmarshalVolumeClaimTemplates parses the volumeClaimTemplates annotation value.
+// It auto-detects the format: new format ([]VolumeClaimTemplateEntry) or
+// legacy format ([]corev1.PersistentVolumeClaim), converting legacy to new format.
+func UnmarshalVolumeClaimTemplates(data string) ([]VolumeClaimTemplateEntry, error) {
```
w13915984028, line 38: Harvester controller can handle this change, what's the considertaions for 

`docker-machine-driver, terrfrom-provider and third-party`

they can write an old formatted data, Harvester controller converts it to new format, and those tools can't decode again;  it is very difficult to force all those tools to upgrade alone with Harvester v180

do we consider the sotlution to add a new annotation to represent the `migration target`? after migration is done, remove this new annotation, update original template annotation with new pvc

the vm backup & restore could be another challenge, v180 new vm annotation format can't be explained by a v170 cluster

this breaking change might trigger additional un-noticed cases

 the `TargetVolume string   ` is more like a `intermediate variable`;  a new object array (then as a new annotation) like below coud get the task done
 
 ```
type volume-migration-request struct {
	SourcePvcNamespace string
	SourcePvcName string
	TargetVolume string  
}
```

File: pkg/webhook/resources/virtualmachine/validator.go [hunk 7]
```diff
@@ -691,3 +696,87 @@ func (v *vmValidator) checkCdRomVolumeIsValid(vm *kubevirtv1.VirtualMachine) err
 
 	return nil
 }
+
+func (v *vmValidator) checkTargetVolumes(vm *kubevirtv1.VirtualMachine) error {
```
w13915984028: could you add unit test code to cover the different failure cases

w13915984028 (commented):
A few questions, thanks.

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,187 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
 	return nil
 }
 
+// SetTargetVolumeStrategy detects targetVolume entries in volumeClaimTemplates
+// and updates the VM spec to trigger KubeVirt volume migration.
+func (h *VMController) SetTargetVolumeStrategy(_ string, vm *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
```
ihcsim, line 506: From what I can see, `targetVolume` is only used in the vm controller and validator. As brought up in the [HEP](https://github.com/harvester/harvester/pull/10086/changes#r2849647930), if we use another annotation to track the oldName/newName mapping, and leave the existing `volumeClaimTemplates` alone, I think all the changes in the handler.go, disk.go, template_version_controller.go, indexer.go, calculator.go etc. won't be needed.

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,187 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+		_, dvErr := h.dataVolumeClient.Get(vm.Namespace, newName, metav1.GetOptions{})
+		if dvErr == nil {
+			vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = kubevirtv1.VolumeSource{
+				DataVolume: &kubevirtv1.DataVolumeSource{
+					Name:         newName,
+					Hotpluggable: vol.PersistentVolumeClaim.Hotpluggable,
+				},
+			}
+		} else {
+			vmCopy.Spec.Template.Spec.Volumes[i].PersistentVolumeClaim.ClaimName = newName
```
ihcsim: I don't think we can assume that just because the previous GET returned an error, this is automatically going to be a PVC. Is it really necessary to support this fallback to PVC? Can't make this a 1.8 feature that works only with data volumes?

pkg/util/resourcequota/calculator.go
```diff
@@ -463,19 +463,18 @@ func IsEmptyResourceQuota(rq *corev1.ResourceQuota) bool {
...
-	err := json.Unmarshal([]byte(volumeClaimTemplates), &pvcs)
+	entries, err := util.UnmarshalVolumeClaimTemplates(volumeClaimTemplatesStr)
 	if err != nil {
 		return storage, fmt.Errorf("failed to unmarshal the volumeClaimTemplates annotation: %w", err)
 	}
 
-	for _, pvc := range pvcs {
-		storage.Add(*pvc.Spec.Resources.Requests.Storage())
+	for _, entry := range entries {
+		storage.Add(*entry.PVC.Spec.Resources.Requests.Storage())
```
ibrokethecloud: should we also be adding the target PVC capacity, since that also contributes to the underlying namespace quota

pkg/util/volume.go
Vicente-Cheng: We have the corresponding support matrix for node-driver and terraform-provider.
So, don't support the old version of node-driver, and terraform-provider would be ok.


I discussed with @ibrokethecloud, and now the annotation format would look like
```
type VolumeClaimTemplateEntry struct {
	corev1.PersistentVolumeClaim `json:",inline"`
	TargetVolume                 string `json:"targetVolume,omitempty"`
}
```

It would be compatible with the old format.

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: We use the new format, as I mentioned on https://github.com/harvester/harvester/pull/10088#discussion_r2870283903

Now we keep most operations the same as before.

pkg/webhook/resources/virtualmachine/validator.go

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -691,3 +696,87 @@ func (v *vmValidator) checkCdRomVolumeIsValid(vm *kubevirtv1.VirtualMachine) err
...
+
+	entries, err := util.UnmarshalVolumeClaimTemplates(annStr)
+	if err != nil {
+		return werror.NewInvalidError(
+			fmt.Sprintf("failed to unmarshal volumeClaimTemplates annotation: %v", err),
+			fmt.Sprintf("metadata.annotations[%s]", util.AnnotationVolumeClaimTemplates),
+		)
+	}
+
+	fieldPath := fmt.Sprintf("metadata.annotations[%s]", util.AnnotationVolumeClaimTemplates)
```
ibrokethecloud: if we move this to earlier than L707, then `fieldPath` can be used in return on L709

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,178 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+
+	vmCopy := vm.DeepCopy()
+	for i, vol := range vmCopy.Spec.Template.Spec.Volumes {
+		var oldName string
+		if vol.PersistentVolumeClaim != nil {
+			oldName = vol.PersistentVolumeClaim.ClaimName
+		}
+		if oldName == "" {
+			continue
+		}
```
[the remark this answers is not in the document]
ibrokethecloud: agree, if a vm has had a migration to a datavolume, subsequent attempts for storage migration will be missed as we are only checking for source being pvc

pkg/webhook/resources/virtualmachine/validator.go

pkg/util/resourcequota/calculator.go
Vicente-Cheng: discussed offline, the target pvc was calculated before

pkg/controller/master/virtualmachine/vm_controller.go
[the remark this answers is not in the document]
Vicente-Cheng: updated, we can handle datavolume now

ibrokethecloud (dismissed):
lgtm. thanks for the enhancement.

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -388,68 +393,62 @@ func (v *vmValidator) checkVMStoppingStatus(oldVM *kubevirtv1.VirtualMachine, ne
...
 		}
 	}
 	return nil
 }
 
-func (v *vmValidator) checkVolumeAnnotations(oldVM, newVM *kubevirtv1.VirtualMachine) error {
+func (v *vmValidator) checkVolumeAnnotations(oldVM, newVM *kubevirtv1.VirtualMachine) ([]util.VolumeClaimTemplateEntry, error) {
 	oldAnn := oldVM.Annotations[util.AnnotationVolumeClaimTemplates]
 	newAnn := newVM.Annotations[util.AnnotationVolumeClaimTemplates]
 	if oldAnn == "" || newAnn == "" || oldAnn == newAnn {
```
w13915984028, line 415: original code might have an issue:

if `oldAnn == "" but newAnn != ""` then the check is jumped, are there any other check upon `newAnn` ?

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -691,3 +690,86 @@ func (v *vmValidator) checkCdRomVolumeIsValid(vm *kubevirtv1.VirtualMachine) err
...
+		if len(entries) == 0 {
+			return nil
+		}
+	}
+
+	// Build a set of volume names from VM spec for validation
+	vmVolumeNames := map[string]bool{}
+	if vm.Spec.Template != nil {
+		for _, vol := range vm.Spec.Template.Spec.Volumes {
+			if vol.PersistentVolumeClaim != nil {
```
w13915984028, line 716: is this case possible:

for different volume from PVC and datavolume, they happend to have same name? the `vmVolumeNames` cannot identify this

w13915984028 (commented):
Not finished, post those two questions first. thanks.

pkg/api/vm/formatter.go
```diff
@@ -172,6 +174,27 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	if canDismissInsufficientResourceQuota(vm) {
 		resource.AddAction(request, dismissInsufficientResourceQuota)
 	}
+
+	if vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusRunning {
```
w13915984028: this block might needs to be added under L132, the `canMigrate` has more checks than the `statusRunning' (e.g single-node cluster, vm itself cannot migrate due to node selector or pci devices)

when storageMigration is on top of traditional vm live migration

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,225 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+	}
+	// Not yet migrating, keep waiting
+	h.vmController.EnqueueAfter(vm.Namespace, vm.Name, 5*time.Second)
+	return vm, nil
+}
+
+// CleanupTargetVolumeAnnotation updates the volumeClaimTemplates annotation after volume migration completes.
+// Migration is considered complete when:
+// 1. The annotation entry has a non-empty targetVolume field
+// 2. The VM spec volume already points to the targetVolume name
```
w13915984028, line 597: what happens in case the migration failed ?

pkg/api/vm/handler.go
```diff
@@ -1870,3 +1888,116 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
...
+	}
+
+	vmCopy := vm.DeepCopy()
+	vmCopy.Annotations[util.AnnotationVolumeClaimTemplates] = newAnn
+
+	// Restore VM spec volumes that were swapped to target back to source
+	for i, vol := range vmCopy.Spec.Template.Spec.Volumes {
+		var claimName string
+		var hotpluggable bool
+		if vol.PersistentVolumeClaim != nil {
```
w13915984028: given the `sourcename` and `targetname` as the input params, but from this block, not sure if datavolume and pvc can happen to have same name, then the input params might not be enough to unify the object

w13915984028 (commented):
some additional questions, thanks

w13915984028:
BTW, please add detailed test plans to the PR or issue, &/ with local test report, that will help reviewers to double check all expected behaviours / exceptions, and QA also works per it. Thanks.

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,225 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+		newName, ok := targetVolumes[oldName]
+		if !ok {
+			continue
+		}
+
+		vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = h.buildVolumeSource(vm.Namespace, newName, hotpluggable)
+	}
+
+	strategy := kubevirtv1.UpdateVolumesStrategyMigration
+	vmCopy.Spec.UpdateVolumesStrategy = &strategy
```
ihcsim, line 550: do we actually need to set/reset this strategy every time? i thought it's more like a static configuration that ui can just set and then we leave it alone, no?

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -388,68 +393,62 @@ func (v *vmValidator) checkVMStoppingStatus(oldVM *kubevirtv1.VirtualMachine, ne
...
-	if err := json.Unmarshal([]byte(data), &pvcs); err != nil {
-		return nil, err
-	}
-	for _, pvc := range pvcs {
-		if pvc.Namespace == "" {
-			pvc.Namespace = ns
-		}
-	}
-	return pvcs, nil
+	return newEntries, v.checkPVCsStorageRequestsAndClass(oldPvcMap, newPvcMap)
```
ihcsim: i think `checkPVCsStorageRequestsAndClass()` prevents changes on storage class, which i thought is one of the use cases for storage migration

ihcsim (dismissed):
Overall LGTM, barring Jian's comment at https://github.com/harvester/harvester/pull/10088#discussion_r2889557346.

Part of me also wonder if the UI can just submit a new VM YAML to the backend, with the source/dest volume names change, let KubeVirt do its thing, and then we handle only the volume size validation, volume template annotation rebuild, errors and revert (upstream [has API for this](https://github.com/kubevirt/kubevirt/blob/50b9b4d0cf1caaf56147411a1e6765134e0e82b9/staging/src/kubevirt.io/api/core/v1/types.go#L2098-L2100)), and if any, clean up.

Feels like the backend code is doing a lot more than it needs to be because of the new annotation struct.

UPDATE: By "new VM YAML", I meant "edited YAML" of the same VM, not a completely new VM.

w13915984028:
Recalled another scenario:

Like nic-hot-plug, the volume migration also needs to skip guest cluster VM atm, before this feature is integrated with Rancher Manager (if it is required & supported). thanks.

https://github.com/harvester/harvester/blob/729a302eefbd9d464554e14a38662f6c8effde8c/pkg/util/virtualmachine/virtualmachine.go#L79-L82

pkg/webhook/resources/virtualmachine/validator.go
Vicente-Cheng: I think the `oldAnn == ""` situation might not exist?
Since we rely on this annotation to create the VM spec, if we missed it, the VM spec might be entirely controlled by the user.

Also, GUI relies on this annotation. If the user missed this annotation, we should not treat this as our control.
If we want to align or prevent, maybe we should create a separate issue to handle it.

pkg/webhook/resources/virtualmachine/validator.go
Vicente-Cheng: yeah, the DataVolume will create the same name PVC.
That's our assumption.

pkg/webhook/resources/virtualmachine/validator.go
w13915984028: I filed https://github.com/harvester/harvester/issues/10165 to track it, let's remove it from this PR's scope and discussion, thanks.

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: The controller of the Kubevirt side will try another migration until we cancel it.

pkg/api/vm/handler.go
Vicente-Cheng: Like this [comment](https://github.com/harvester/harvester/pull/10088#discussion_r2902881048), this is our assumption.

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: yes, we can leave here, but I would like to remove it to keep every rollback like we do the storage migration

pkg/api/vm/formatter.go
Vicente-Cheng: nice catch, updated!

Vicente-Cheng:
> Part of me also wonder if the UI can just submit a new VM YAML to the backend, with the source/dest volume names change, let KubeVirt do its thing, and then we handle only the volume size validation, volume template annotation rebuild, errors and revert (upstream [has API for this](https://github.com/kubevirt/kubevirt/blob/50b9b4d0cf1caaf56147411a1e6765134e0e82b9/staging/src/kubevirt.io/api/core/v1/types.go#L2098-L2100)), and if any, clean up.

Hi @ihcsim, 
Sorry, I am not really sure about your question.
>Part of me also wonder if the UI can just submit a new VM YAML to the backend, with the source/dest volume names change, let KubeVirt do its thing, and then we handle only the volume size validation

Is that what we currently do?
We only add the destination volume name, and the controller will help us to change it.
Or you mean we should do it w/o any backend controller, just rely on the API call?

Vicente-Cheng:
> Recalled another scenario:
> 
> Like nic-hot-plug, the volume migration also needs to skip guest cluster VM atm, before this feature is integrated with Rancher Manager (if it is required & supported). thanks.
> 
> https://github.com/harvester/harvester/blob/729a302eefbd9d464554e14a38662f6c8effde8c/pkg/util/virtualmachine/virtualmachine.go#L79-L82

I will do some checks with the guest cluster scenario. Maybe we can track it on another issue.
WDYT?

pkg/webhook/resources/virtualmachine/validator.go
w13915984028: From this block of code

https://github.com/harvester/harvester/pull/10088/changes#diff-1cc9e10e35794556fa7be7339c4a95391cf18852a1cabf389138d7e9a7827752R533-R545

We can further assume that: after some volume migratings, volume attache/detach (vm stops) operations, a VM could have same named `pvc` and `datavolume` but different `volumes.name` under vm spec: (please correct me if I am wrong), thanks.

if this is the case, the `sourcename` and `targetname` could mis-match

```
      volumes:
        - name: disk-0
          persistentVolumeClaim:
            claimName: vm4-disk-0-nhj15
        - cloudInitNoCloud:
            networkDataSecretRef:
              name: vm4-mlwuy
            secretRef:
              name: vm4-mlwuy
          name: cloudinitdisk
        - name: disk2
          persistentVolumeClaim:   // migrate this one
            claimName: vol-2
            hotpluggable: true
        - name: disk3              // the VM has already with a datavolume
          dataVolume:
            name: vol-2
```

together with question: https://github.com/harvester/harvester/pull/10088/changes#r2904431646

we did not expose `datavolume` in the past;  from the description from https://kubevirt.io/user-guide/storage/volume_migration/, it seems `datavolume` and `pvc` names are independent

pkg/controller/master/virtualmachine/vm_controller.go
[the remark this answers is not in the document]
w13915984028: in this case, another question arise:

the entries's definition sticks to `pvc, targetvolume`, if works upon `datavolume` as well, could the `volumeClaimTemplates` and other parts miss some related processing? thanks.

```
entries, err := util.UnmarshalVolumeClaimTemplates(volumeClaimTemplatesStr)
```

e.g. `finalizeCompletedMigrations` only cleans the targetVolume which is encoded into the `PVC + targetVolume` entries

w13915984028:
@Vicente-Cheng It is good to track guest cluster case on another issue/PR to limit this PR's scope, thanks.

ihcsim:
> Or you mean we should do it w/o any backend controller, just rely on the API call?

Yeah, that's my question. Instead, of sending an input of src/target volumes, does the UI have the mechanism to update the VM's spec directly? (I thought the UI does that it some places.) AIUI, the only thing that really needs to change for online storage migration to work is just the data volume or pvc names within the VM spec. Then KubeVirt will update the VM status with the volume migration status and details. I feel like most of the changes in this PR are needed only because we are trying to manage the migration state ourselves in the annotations.

I'm still ok with my previous approval, as long as we address the cases Jian pointed out.

w13915984028:
FYI: Per offline discussion, I extract those open topics to other tickets https://github.com/harvester/harvester/issues/10183 and https://github.com/harvester/harvester/issues/10184 thanks.

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,225 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+	}
+
+	// First round: set spec + waiting annotation
+	vmCopy.Annotations[util.AnnotationWaitingStorageMigration] = "true"
+	return h.vmClient.Update(vmCopy)
+}
+
+func (h *VMController) buildVolumeSource(namespace, name string, hotpluggable bool) kubevirtv1.VolumeSource {
+	_, dvErr := h.dataVolumeClient.Get(namespace, name, metav1.GetOptions{})
+	if dvErr == nil {
```
brandboat: Perhaps we should directly throw error if the dvErr isn't NotFound error?

pkg/api/vm/handler.go
```diff
@@ -1870,3 +1888,116 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
...
+
+		sourceName, ok := targetToSource[claimName]
+		if !ok {
+			continue
+		}
+
+		// Determine if source is a DataVolume or PVC
+		_, dvErr := h.datavolumeClient.Get(namespace, sourceName, metav1.GetOptions{})
+		if dvErr == nil {
+			vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = kubevirtv1.VolumeSource{
```
brandboat: Perhaps we should directly throw error if the dvErr isn't NotFound error?

pkg/api/vm/formatter.go
```diff
@@ -53,6 +53,8 @@ const (
 	updateResourceQuotaAction        = "updateResourceQuota"
```
brandboat, line 53: Do we need to add a validation to prevent user from deleting the source during storage migration? The current implementation only prevents the target volume from deletion.

https://github.com/harvester/harvester/blob/f38813f33a19d7138ef42529f238225c77fa3176/pkg/webhook/resources/persistentvolumeclaim/validator.go#L100-L110

pkg/api/vm/handler.go
```diff
@@ -1870,3 +1888,116 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
...
+			}
+		}
+	}
+
+	delete(vmCopy.Annotations, util.AnnotationWaitingStorageMigration)
+	vmCopy.Spec.UpdateVolumesStrategy = nil
+	if _, err = h.vmClient.Update(vmCopy); err != nil {
+		return err
+	}
+	return h.abortMigration(namespace, name)
```
brandboat: Is this necessary? From the [manual](https://kubevirt.io/user-guide/storage/volume_migration/#volume-migration-cancellation), I didn't see KubeVirt mention we need to delete the VMIM.

brandboat (commented):
Thanks for the great feature! Just a few minor comments. I haven't manually tested it yet, but I'll definitely do so once I have a moment.

pkg/api/vm/formatter.go
Vicente-Cheng: nice catch. updated

pkg/api/vm/handler.go
Vicente-Cheng: Yeah, we need that. Only restoring the source PVC cannot complete the whole restore flow.

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: we now drop the dataVolume support, please recheck

pkg/api/vm/handler.go
Vicente-Cheng: we now drop the dataVolume support, please recheck

Vicente-Cheng:
Hi folks, here is the latest change. 
Now we only use the PVC as the migration target.

pkg/api/vm/handler.go
```diff
@@ -300,6 +298,17 @@ func (h *vmActionHandler) Do(ctx *harvesterServer.Ctx) (harvesterServer.Response
 			return nil, apierror.NewAPIError(validation.InvalidBodyContent, "Failed to decode request body: %v "+err.Error())
 		}
 		return nil, h.cpuAndMemoryHotplug(namespace, name, input)
+	case storageMigration:
+		var input StorageMigrationInput
+		if err := json.NewDecoder(r.Body).Decode(&input); err != nil {
+			return nil, apierror.NewAPIError(validation.InvalidBodyContent, "Failed to decode request body: %v "+err.Error())
+		}
+		if input.SourceVolume == "" || input.TargetVolume == "" {
```
w13915984028, line 306: should we add another check to ensure:

`input.SourceVolume != input.TargetVolume`, to avoid migrating to self

pkg/api/vm/handler.go
```diff
@@ -1535,3 +1549,99 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
...
+			},
+		}
+	}
+
+	delete(vmCopy.Annotations, util.AnnotationWaitingStorageMigration)
+	vmCopy.Spec.UpdateVolumesStrategy = nil
+	if _, err = h.vmClient.Update(vmCopy); err != nil {
+		return err
+	}
+	return h.abortMigration(namespace, name)
```
w13915984028: if `h.vmClient.Update` is successful, but `h.abortMigration(namespace, name)` fails, are there chances to retry?

```
	if _, err = h.vmClient.Update(vmCopy); err != nil {
		return err
	}
	return h.abortMigration(namespace, name)
```

w13915984028 (commented):
Some last questions, thanks.

pkg/webhook/resources/virtualmachine/validator.go
w13915984028: It is not related to this PR. The corner case will be handled on 

https://github.com/harvester/harvester/issues/10165#issuecomment-4080685357

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,216 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+
+	vmCopy := vm.DeepCopy()
+	for i, vol := range vmCopy.Spec.Template.Spec.Volumes {
+		var oldName string
+		var hotpluggable bool
+		if vol.PersistentVolumeClaim != nil {
+			oldName = vol.PersistentVolumeClaim.ClaimName
+			hotpluggable = vol.PersistentVolumeClaim.Hotpluggable
+		} else if vol.DataVolume != nil {
+			oldName = vol.DataVolume.Name
```
w13915984028: per latest code, should `oldName = vol.DataVolume.Name` be excluded from the map?

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,216 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+		}
+
+		newName, ok := targetVolumes[oldName]
+		if !ok {
+			continue
+		}
+
+		vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = buildPVCVolumeSource(newName, hotpluggable)
+	}
+
```
w13915984028, line 548: it looks essential to add a flag between L533~L553 to indicate a real volume swap happens,

L556 `vmCopy.Spec.UpdateVolumesStrategy = &strategy` will cause `reflect.DeepEqual` is true in most cases, even the loop does not hit any target

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,216 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+		return vm, nil
+	}
+
+	// Skip cleanup if storage migration is waiting to start
+	if _, ok := vm.Annotations[util.AnnotationWaitingStorageMigration]; ok {
+		return vm, nil
+	}
+
+	// Wait until VM is back to Running and VolumesChange condition is gone
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusRunning {
```
w13915984028, line 616: L623, L626 check seems should be moved before L603; most of the VM changes are not related to volume migration, that check will return early

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,216 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+	return false
+}
+
+func getVMSpecVolumeNames(vm *kubevirtv1.VirtualMachine) map[string]bool {
+	names := map[string]bool{}
+	for _, vol := range vm.Spec.Template.Spec.Volumes {
+		if vol.PersistentVolumeClaim != nil {
+			names[vol.PersistentVolumeClaim.ClaimName] = true
+		}
+		if vol.DataVolume != nil {
```
w13915984028: is `DataVolume` still kept?

w13915984028 (commented):
questions:

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,216 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+			continue
+		}
+
+		newName, ok := targetVolumes[oldName]
+		if !ok {
+			continue
+		}
+
+		vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = buildPVCVolumeSource(newName, hotpluggable)
+	}
```
WebberHuang1118, line 547: Suggestion: a small refactor, like this:
```suggestion
	for i, vol := range vmCopy.Spec.Template.Spec.Volumes {
		oldName, hotpluggable, ok := getVolumeBackingRef(vol)
		if !ok {
			continue
		}

		newName, ok := targetVolumes[oldName]
		if !ok {
			continue
		}

		vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = buildPVCVolumeSource(newName, hotpluggable)
	}
```

```
func getVolumeBackingRef(vol kubevirtv1.Volume) (name string, hotpluggable bool, ok bool) {
	if vol.PersistentVolumeClaim != nil {
		return vol.PersistentVolumeClaim.ClaimName, vol.PersistentVolumeClaim.Hotpluggable, true
	}
	if vol.DataVolume != nil {
		return vol.DataVolume.Name, vol.DataVolume.Hotpluggable, true
	}
	return "", false, false
}
```

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,216 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+	for _, vol := range vm.Spec.Template.Spec.Volumes {
+		if vol.PersistentVolumeClaim != nil {
+			names[vol.PersistentVolumeClaim.ClaimName] = true
+		}
+		if vol.DataVolume != nil {
+			names[vol.DataVolume.Name] = true
+		}
+	}
+	return names
+}
```
WebberHuang1118, line 670: Suggestion reuses the function mentioned in previous comment:
```suggestion
func getVMSpecVolumeNames(vm *kubevirtv1.VirtualMachine) map[string]bool {
	names := map[string]bool{}
	for _, vol := range vm.Spec.Template.Spec.Volumes {
		name, _, ok := getVolumeBackingRef(vol)
		if ok {
			names[name] = true
		}
	}
	return names
}
```

pkg/util/indexeres/virtualmachine.go
```diff
@@ -26,7 +28,22 @@ func VMByPVC(obj *kubevirtv1.VirtualMachine) ([]string, error) {
...
+		if err == nil {
+			for _, entry := range entries {
+				if entry.Name != "" {
+					results = append(results, ref.Construct(obj.Namespace, entry.Name))
+				}
+			}
+		}
+	}
+
+	return ds.SliceDedupe(results), nil
```
WebberHuang1118: Suggestion: a small refactor, like this:
```suggestion
	results = append(results, vmSourcePVCsFromAnnotation(obj)...)

	return ds.SliceDedupe(results), nil
}

func vmSourcePVCsFromAnnotation(obj *kubevirtv1.VirtualMachine) []string {
	if obj.Annotations == nil {
		return nil
	}

	entries, err := util.UnmarshalVolumeClaimTemplates(obj.Annotations[util.AnnotationVolumeClaimTemplates])
	if err != nil {
		return nil
	}

	var results []string
	for _, entry := range entries {
		if entry.Name != "" {
			results = append(results, ref.Construct(obj.Namespace, entry.Name))
		}
	}

	return results
```

pkg/api/vm/handler.go
```diff
@@ -1535,3 +1549,99 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
...
+		}
+
+		vmCopy.Spec.Template.Spec.Volumes[i].VolumeSource = kubevirtv1.VolumeSource{
+			PersistentVolumeClaim: &kubevirtv1.PersistentVolumeClaimVolumeSource{
+				PersistentVolumeClaimVolumeSource: corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: sourceName,
+				},
+				Hotpluggable: vol.PersistentVolumeClaim.Hotpluggable,
+			},
+		}
```
WebberHuang1118, line 1641: If the migration completes during the handler process, and Harvester reverts to the original PVC here, would that trigger another migration and cause a rollback?

pkg/api/vm/handler.go
Vicente-Cheng: updated

pkg/api/vm/handler.go
Vicente-Cheng: No, consider that it is the UI action, UI will see the error, and the user could retry.

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: nice catch. updated

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: removed, updated

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: Yes, you are right, we also need to check the annotation `AnnotationWaitingStorageMigration` first. 
Moved these three conditions for the early return.

updated

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: DeepEqual compares the value of the point if the point of the old/new var was non-nil.
So, the above case you mentioned would not happen.

pkg/controller/master/virtualmachine/vm_controller.go
```diff
@@ -501,6 +501,210 @@ func (h *VMController) createDataVolume(pvcName, pvcNS, imageID, imageNS string,
...
+		PersistentVolumeClaim: &kubevirtv1.PersistentVolumeClaimVolumeSource{
+			PersistentVolumeClaimVolumeSource: corev1.PersistentVolumeClaimVolumeSource{
+				ClaimName: name,
+			},
+			Hotpluggable: hotpluggable,
+		},
+	}
+}
+
+func (h *VMController) handleMigrationWait(vm *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
```
WebberHuang1118: IIUC, this function is meant to clear the migration-wait annotation—maybe we should rename it?
```suggestion
func (h *VMController) clearMigrationWait(vm *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
```

WebberHuang1118 (commented):
LGTM overall—just a couple of questions and a few small refactoring suggestions. Thanks.

pkg/api/vm/handler.go
Vicente-Cheng: yeah, if we can catch the timing, it will trigger another migration.
but I think that would be the edge case...

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: updated

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: updated

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: updated

pkg/util/indexeres/virtualmachine.go
Vicente-Cheng: updated

pkg/api/vm/handler.go
WebberHuang1118: Is it possible to have a safeguard in place to prevent this from happening? It appears there is no atomic way for us to check this, right?

pkg/api/vm/formatter.go
```diff
@@ -172,6 +175,23 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	if canDismissInsufficientResourceQuota(vm) {
 		resource.AddAction(request, dismissInsufficientResourceQuota)
 	}
+
+	if vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusMigrating && hasActiveStorageMigration(vm) {
+		resource.AddAction(request, cancelStorageMigration)
```
w13915984028, line 180: If a VM's volume migration is stucking at `waiting to start` e.g. the system has reached the burst mgiration count;  user won't see this `cancelStorageMigration` menu

pkg/api/vm/formatter.go
```diff
@@ -130,6 +132,7 @@ func (vf *vmformatter) formatter(request *types.APIRequest, resource *types.RawR
 	if canMigrate(vf.nodeCache, vmi) {
 		resource.AddAction(request, migrate)
 		resource.AddAction(request, findMigratableNodes)
+		resource.AddAction(request, storageMigration)
```
w13915984028, line 135: as storageMigration has a waiting period, if a migration has been triggered, then new `storageMigration` should be blocked,

namely, check `hasActiveStorageMigration` is essential, unless we allow to `migrate multi volumes`

pkg/api/vm/handler.go
```diff
@@ -1535,3 +1552,99 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
...
+	annStr := vm.Annotations[util.AnnotationVolumeClaimTemplates]
+	entries, err := util.UnmarshalVolumeClaimTemplates(annStr)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal volumeClaimTemplates annotation: %w", err)
+	}
+
+	found := false
+	for i, entry := range entries {
+		if entry.Name == input.SourceVolume {
+			entries[i].TargetVolume = input.TargetVolume
```
w13915984028, line 1571: if only 1 volume migration is allowed per time, then a check is needed to ensure there is only one migration request

w13915984028 (commented):
See below challenge on volume migration state machine:

A.  only allow 1 volume migration request on 1 vm

(1) migrate menu rendering needs to check if there is already request (e.g. pending with `util.AnnotationWaitingStorageMigration`)
(2) allow to cancel
(3) if a VM is put to stop, the leftover migration request needs to be cleaned

B. allow multi volume migration request on 1 vm

To be discussed if this is confirmed.

thanks.

WebberHuang1118 (dismissed):
LGTM! Thanks for the new feature.

pkg/util/volume.go
votdev: Just an idea, maybe we should introduce a versioned envelope for things like this in the future. That way, we can make changes without any major issues, and the parser can automatically adjust and convert things based on the version.
```
{
  "version": "1.0",
  "content": [
    {
      "spec": {
        "accessModes": [
          "ReadWriteMany"
        ],
        "resources": {
          "requests": {
            "storage": "10Gi"
          }
        }
      },
      "targetVolume": "target-pvc"
    }
  ]
}
```

pkg/util/volume.go
```diff
@@ -24,6 +24,47 @@ var (
...
+	var entries []VolumeClaimTemplateEntry
+	if err := json.Unmarshal([]byte(data), &entries); err != nil {
+		return nil, err
+	}
+	return entries, nil
+}
+
+// MarshalVolumeClaimTemplates serializes entries to JSON for the annotation.
+func MarshalVolumeClaimTemplates(entries []VolumeClaimTemplateEntry) (string, error) {
+	data, err := json.Marshal(entries)
```
votdev, line 55: I think it is necessary to handle `nil` explicitly here, otherwise `json.Marshal(nil)` returns `"null"` which is not correct compared to the code in `UnmarshalVolumeClaimTemplates`.

```suggestion
    if entries == nil {
        return "", nil
    }
	data, err := json.Marshal(entries)
```
or

```suggestion
    if len(entries) == 0 {
        return "", nil
    }
	data, err := json.Marshal(entries)
```
if we want to return an empty string for an empty list as input arg.

pkg/api/vm/handler.go
brandboat: this is guaranteed by the input `StorageMigrationInput`, right? as the StorageMigrationInput only accept single pair of source + target.

pkg/api/vm/handler.go
```diff
@@ -1535,3 +1552,99 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
 	_, err := h.vmClient.Patch(namespace, name, k8stypes.JSONPatchType, []byte(patchData))
 	return err
 }
+
+func (h *vmActionHandler) storageMigration(namespace, name string, input StorageMigrationInput) error {
+	vm, err := h.vmCache.Get(namespace, name)
```
votdev, line 1557: This code is called from within an API action handler, so if we use the cache here, the object might be stale and might produce an error that bubbles up to the UI. Does it make more sense to request the resource from the API server instead (regardless of the fact that the cache is, of course, faster)?

```suggestion
	vm, err := h.vmClient.Get(namespace, name)
```

pkg/api/vm/handler.go
```diff
@@ -1535,3 +1552,99 @@ func (h *vmActionHandler) cpuAndMemoryHotplug(namespace, name string, input CPUA
...
+	}
+
+	vmCopy := vm.DeepCopy()
+	vmCopy.Annotations[util.AnnotationVolumeClaimTemplates] = newAnn
+	_, err = h.vmClient.Update(vmCopy)
+	return err
+}
+
+func (h *vmActionHandler) cancelStorageMigration(namespace, name string) error {
+	vm, err := h.vmCache.Get(namespace, name)
```
votdev, line 1592: See my comment in the `storageMigration` function.
```suggestion
	vm, err := h.vmClient.Get(namespace, name)
```

pkg/api/vm/handler.go
w13915984028: I mean: when one migration action is triggered (e.g. vol A), but not starting, could user trigger a new migration (vol B) ?

atm, `storageMigrate` is added per blow, before the vmim object is created, it can be called several times, in theory, multi `source + target` can be added, I need to confirm: is this allowed.

https://github.com/Vicente-Cheng/harvester/blob/0c59c52c01ebd458c1778efa3b744228f7a8650f/pkg/api/vm/formatter.go#L135

pkg/api/vm/formatter.go
Vicente-Cheng: I will add a webhook to allow only one target volume. I think that we can ensure only one volume migration

pkg/api/vm/handler.go
Vicente-Cheng: see above, I will add a webhook to allow only one target volume. I think that we can ensure only one volume migration

w13915984028: allow only one volume migration is a better solution for now, thanks.

pkg/util/volume.go
Vicente-Cheng: nice catch, now I check the entry length before Marshal.
updated

pkg/api/vm/handler.go
Vicente-Cheng: Consider the wrangler mechanism, the vmCache should be almost the same as the vmClient caller.
If we want to improve that, we should create another issue because most API callers use the cache instead of the client.

WDYT?

pkg/api/vm/handler.go
Vicente-Cheng: ditto

pkg/api/vm/handler.go
Vicente-Cheng: updated

pkg/api/vm/formatter.go
Vicente-Cheng: updated

pkg/api/vm/formatter.go
Vicente-Cheng: Thanks for the hint, I relaxed the condition. 
Once we have the targetVolume, we can call the cancel storage migration.

pkg/api/vm/handler.go
Vicente-Cheng: yeah, I think we can leave here to see if this situation would be easy to hit or not.

w13915984028 (approved):
LGTM, thanks for your great efforts.

ihcsim (approved):

pkg/util/volume.go
ihcsim: I think overall we should just avoid using annotations to build finite state machine, to store states. The proper place for storing the actual states is the `status`. In this case, KubeVirt already provides lots of state information about all volume migrations of a vm as https://kubevirt.io/api-reference/main/definitions.html#_v1_volumeupdatestate. Versioning the annotation template means the controller has to manage the version schemes while maintaining "backward compatibility".

Vicente-Cheng:
@mergifyio backport v1.8

pkg/api/vm/handler.go
votdev: I'm fine with the current implementation; I always use the cache for cost and speed reasons. I just thought I'd ask the group if it makes sense to deviate from this in this particular case.

PR #10107, opened 2026-02-26
restrict disabling of kubeovnoperator addon when VMs are using it

rrajendran17:
#### Problem:
kubeovn operator addon can be disabled now even when the VMs are using the overlay networks.

#### Solution:
restrict users from disabling the kubeovn operator addon when VMs are using it.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/8685

#### Test plan:
1.Enable kubeovn operator addon
2.Create overlay network
3.Create subnet and attach to a provider (overlay network) created in step 2.
4.Create a VM and attach it to the overlay network created in step 2.
5.Wait for VM to move to running.
6.Disable the kubeovn operator addon (webhook should throw error)
7.Stop the VM.
8.Disable the kubeovn operator addon (webhook should throw error)
9.Delete the VM
10.Disable the kubeovn operator addon (should be succesfully disabled) - Follow documentation to disable kubeovn operator


pkg/webhook/resources/addon/validator.go
```diff
@@ -378,3 +384,25 @@ func (v *addonValidator) validateNvidiaDriverToolkitAddon() error {
...
+	}
+
+	subnets, err := v.kubeovnSubnet.List("", labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to retrieve subnets err=%v", err)
+	}
+
+	//V4UsingIPs will be non-zero even when the VMs are not in running state.
+	for _, subnet := range subnets {
+		if subnet.Status.V4UsingIPs != 0 {
```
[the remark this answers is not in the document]
rrajendran17: vms can be identified from the subnet name.The existing code should be fine.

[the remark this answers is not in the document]
rrajendran17: Harvester does not use ipv6 atm.

pkg/webhook/resources/addon/validator.go
```diff
@@ -378,3 +384,30 @@ func (v *addonValidator) validateNvidiaDriverToolkitAddon() error {
...
+	if v.kubeovnSubnet == nil {
+		return nil
+	}
+
+	subnets, err := v.kubeovnSubnet.List("", labels.Everything())
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("failed to retrieve subnets err=%v", err))
+	}
+
+	//V4UsingIPs will be non-zero even when the VMs are not in running state.
```
w13915984028: a question about this comment:

if a VM is created with ovn network ref but never started, does subnet also have `subnet.Status.V4UsingIPs != 0` ?

and the `default` subnet, could it have some exceptions?

check vm, vmi are the `traditional` solution to identify if a nad is being used

pkg/webhook/resources/addon/validator.go
```diff
@@ -122,6 +126,8 @@ func (v *addonValidator) validateUpdatedAddon(newAddon *v1beta1.Addon, oldAddon
 		return v.validatePCIDevicesControllerAddonUpdate(newAddon, oldAddon)
 	case util.NvidiaDriverToolkitName:
 		return v.validateNvidiaDriverToolkitAddonUpdate(newAddon, oldAddon)
+	case util.KubeOVNOperatorName:
```
w13915984028, line 129: another question: do we allow ovn addon to be deleted if some vms are still using it?

w13915984028 (commented):
Thanks for your PR, two questions.

pkg/webhook/resources/addon/validator.go
rrajendran17: when an overlay vm is created with stopped state, subnet.status.V4UsingIPs will be 0 (since no IP allocated yet). 
IMO, This flow (creating VMs created in stopped state, disabling the operator, and then starting the vms again) is technically possible but not representative of a standard production use case.Also currently disabling the kubeovn operator requires a standard procedure which is documented (cleanup) before disabling the operator.

pkg/webhook/resources/addon/validator.go
rrajendran17: Yes looks like we allow it currently but I feel it has to be restricted for kubeovn addon as it requires a specific procedure before actual disable from UI.I have also included kubeovnoperator name to the list of addons restricted from delete.

pkg/webhook/resources/addon/validator.go
rrajendran17: Using traditional solution to check for vm from nad is also an another option.We will need to copy the same logic from network controller in Harvester.Or we need to see if we can create something in common util in future for network controller and harvester to use for similar logics.

Since we have subnet cache already available in Harvester repo, the code becomes simpler by checking subnet's status.

w13915984028: about the comment:

>//V4UsingIPs will be non-zero even when the VMs are not in running state.

should it be below ?

>// V4UsingIPs will be non-zero when the VMs are in running state.

pkg/webhook/resources/addon/validator.go
```diff
@@ -305,8 +311,8 @@ func (v *addonValidator) Delete(_ *types.Request, oldObj runtime.Object) error {
 		return nil
 	}
 	// don't allow delete non-experimental addons
-	//  strictly protect rancher-monitoring and rancher-logging
-	if oldAddon.Name == util.RancherLoggingName || oldAddon.Name == util.RancherMonitoringName || oldAddon.Labels[util.AddonExperimentalLabel] != "true" {
+	//  strictly protect rancher-monitoring and rancher-logging and kubeovn-operator
```
w13915984028, line 314: should `KubeOVNOperatorName` be protected from being deleted only when it is `enabled` ?

when it has been disabled, or never enabled, deleting it seems reasonable

w13915984028 (commented):
last questions, thanks.

pkg/webhook/resources/addon/validator.go
rrajendran17: From UI, we have option only to disable/enable any add-on.The delete operation could only be performed from cli.I am just wondering do we have to really allow delete of any add-ons and is this really a needed operation ?

pkg/webhook/resources/addon/validator.go
rrajendran17: I meant even the VMs in stopped state.I have rephrased and updated it.Thanks

pkg/webhook/resources/addon/validator.go
w13915984028: From document like https://docs.harvesterhci.io/v1.7/advanced/addons/lvm-local-storage#installing-and-enabling-the-add-on

the `experimental-addons` is not pre-packed and needs to be installed via kubectl manually, it will not be surprised that user use `kubectl` to delete them directly

We don't allow delete the `enabled` kube-ovn, will be the wanted case.

w13915984028 (commented):
LGTM, thanks.

pkg/webhook/resources/addon/validator.go
```diff
@@ -378,3 +384,30 @@ func (v *addonValidator) validateNvidiaDriverToolkitAddon() error {
...
+	if err != nil {
+		return werror.NewInternalError(fmt.Sprintf("failed to retrieve subnets err=%v", err))
+	}
+
+	//V4UsingIPs will be non-zero both when the VM is in running and moved to stopped state.
+	for _, subnet := range subnets {
+		if subnet.Status.V4UsingIPs != 0 {
+			return werror.NewBadRequest(fmt.Sprintf("kubeovn-operator addon cannot be disabled as VMs attached to overlay network %s in subnet %s are still in use, delete the VMs before disabling the addon", subnet.Spec.Provider, subnet.Name))
+		}
+	}
```
starbops, line 410: I have a question: will the gateway IP be listed in this field `V4UsingIPs`?

rrajendran17: gateways IPs by default are added to the excluded list and not assigned to any VMs.So V4UsingIPs will not include gatewayIP of that subnet.

starbops (approved):
LGTM, thank you.

ibrokethecloud (approved):
lgtm. thanks for the enhancement.

rrajendran17:
@mergify backport v1.8

PR #10113, opened 2026-02-27
Fix: Failed image cannot be deleted due to finalizer deadlock caused by duplicate DisplayName

### LINKED ISSUE #10109: [BUG] Failed image cannot be deleted due to finalizer lock when a successful upload exists
### Describe the Bug

When an image upload fails (e.g., due to network interruption or page refresh), the system fails to properly intercept subsequent uploads of an image with the exact same name in the same namespace. If a duplicate image is subsequently uploaded successfully, the original failed image resource gets stuck in the `Terminating` state and cannot be deleted. The deletion operation is actually blocked by the Validating Webhook during the finalizer removal process, resulting in a deadlock. The failed image is only automatically cleaned up by the system after the successful image is deleted.

### To Reproduce

1. Upload an `ubuntu.iso` image to a specific namespace (e.g., `wzy`).
2. Refresh the web page continuously during the upload process to cause the upload to fail. (Note: if you delete this failed image immediately, the deletion succeeds).
3. Keep the failed image resource without deleting it, and re-upload the `ubuntu.iso` image to the same namespace. Wait normally until this second upload succeeds.
4. Attempt to delete the failed image from step 2. Notice that the image cannot be deleted and remains stuck in the `Terminating` state.
5. Delete the successful image from step 3. After a short while, observe that the failed image is automatically cleaned up.

### Expected Behavior

1. The system should strictly enforce the uniqueness of `DisplayName` within the same namespace. Even if the first upload fails and the resource is not deleted, the system should intercept subsequent creation requests for images with the same name at the frontend or Validating Webhook level.
2. Even in the presence of stale data, the deletion logic of a failed image should not be blocked by duplicate name validation. The finalizer should be removed normally to prevent resource deadlocks.

### Support Bundle for Troubleshooting

Since the root cause has been clearly identified at the code level (Validating Webhook & Finalizer deadlock) and a PR with the fix is being submitted, I haven't attached a support bundle here. 

However, if the maintainers still need a bundle to verify the environment state, please let me know and I will gladly generate and provide it.

### Environment

- Harvester version: Harvester 1.7.1
- Impacted VM:  N/A
- Impacted volume (PV):  N/A
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): Baremetal
- Rancher version: N/A

### Additional context

After tracing the source code, this issue is caused by a closed-loop deadlock resulting from two interrelated defects:

1. **Missing label bypasses duplicate validation:** The `CheckDisplayName` validator relies on the `harvesterhci.io/imageDisplayName` label to detect duplicates. However, in the current logic, this label is only synced by the `IsImported` branch of the controller's `OnChanged` function *after* a successful image import. A failed image never reaches this branch and therefore lacks the label. When a second image with the same name is uploaded, the validator cannot find the label on the failed image, incorrectly allowing the creation request.
2. **Finalizer removal blocked by duplicate validation (Deadlock):**
When a user deletes the failed image, the K8s API successfully sets the `DeletionTimestamp`, and the controller's `OnRemove` successfully cleans up the underlying resources. Subsequently, the Wrangler framework initiates an `UPDATE` request to strip the Finalizer. However, this `UPDATE` request passes through the Validating Webhook and triggers `CheckDisplayName`. At this point, because the *successful* duplicate image already exists and has the label, the validator determines that the current `UPDATE` operation has a "duplicate name conflict" and rejects the request. The Finalizer cannot be removed, leaving the image permanently stuck.

### Workaround and Mitigation

If a user is currently experiencing this deadlock (a failed image stuck in `Terminating`), there are two workarounds:

**(Via UI - Safe):**
Delete the *successfully* uploaded duplicate image first. Once the successful image is removed, the `imageDisplayName` label conflict is resolved. The system will then automatically bypass the Validating Webhook block and finish terminating the stuck failed image. After that, you can re-upload the image.

Hero2633:
**Problem:**
When a failed image upload is followed by a successful upload of the same image name in the same namespace, the failed image becomes undeletable (stuck in `Terminating` state). 
This happens because the failed image lacks the `imageDisplayName` label (which was previously only added after a successful import), thus bypassing the duplicate name validation. When a user attempts to delete the failed image, the Validating Webhook blocks the finalizer removal because it detects the newly uploaded successful image as a "duplicate conflict", causing a deadlock.

**Solution:**
This PR implements a two-pronged approach to fix the root cause and ensure backward compatibility:
1. Unconditionally injecting the `imageDisplayName` label in the Mutating Webhook (`mutator.go`) during the `Create` phase. This ensures all images have the label from the start, allowing the Validating Webhook to intercept duplicate uploads immediately.
2. Adjusting the label sync logic in the Controller (`vm_image_controller.go`) to retrofit labels for existing legacy images before the `IsImported` check.

**Related Issue(s):**
Fixes #10109

**Test plan:**
1. Upload an image and interrupt it (e.g., refresh the page) to simulate an upload failure.
2. Attempt to upload a second image with the exact same name to the same namespace. (Expected: Should be blocked by validation).
3. Delete the initial failed image. (Expected: Should delete successfully without the finalizer deadlock).

**Additional documentation or context:**
N/A


Hero2633:
"Ping to keep this PR alive. @Vicente-Cheng @WebberHuang1118 @Yu-Jack @brandboat Gentle reminder on this PR!

I also just updated the branch with the latest base, so the workflows are currently awaiting a maintainer's approval to run. Could you please approve the workflows and take a look at the code when you have a moment? Thank you!"

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -29,17 +29,21 @@ func (h *vmImageHandler) OnChanged(_ string, vmi *harvesterv1.VirtualMachineImag
...
+	// New images get this label from the Mutating Webhook at creation time, but
+	// images created before this change may be missing it. On controller startup,
+	// the informer resyncs all existing resources, triggering OnChanged for each,
+	// so any old images missing the label will have it added here.
+	if vmi.Spec.DisplayName != vmi.Labels[util.LabelImageDisplayName] {
+		toUpdate := vmi.DeepCopy()
+		if toUpdate.Labels == nil {
+			toUpdate.Labels = map[string]string{}
 		}
+		toUpdate.Labels[util.LabelImageDisplayName] = vmi.Spec.DisplayName
```
brandboat, line 42: I'm thinking of deprecating LabelImageDisplayName and replacing it with a dedicated indexer. We mainly use the label to avoid full scans when checking for duplicates https://github.com/harvester/harvester/commit/69a6b1b2517f30cd0dd81852cb575a29f0b1e683, but an indexer feels like a cleaner solution. @WebberHuang1118, do you have any thoughts on this?

brandboat (commented):
Thanks for the fix @Hero2633, and sorry for the late review. We’ve been busy with the 1.8 release recently, which has caused some delays in our PR reviews.

pkg/controller/master/image/vm_image_controller.go
Hero2633: Hi @brandboat,

No worries about the delay, and congrats to the team on the 1.8 release!

Using a dedicated indexer sounds like a great long-term solution. However, since the current system still uses LabelImageDisplayName, this PR actually fixes an immediate bug: if an image upload fails, it might bypass the current IsImported check, miss the label, and cause duplicate image creations later.

I am happy to wait for @WebberHuang1118 's thoughts. If the team wants to merge this as a quick bug fix first, or prefers me to try implementing the indexer in this PR, please just let me know!

As this is my very first open-source PR, I am just excited to contribute and learn from you all. Thanks!

brandboat (dismissed):

pkg/controller/master/image/vm_image_controller.go
WebberHuang1118: Yes, agree with @brandboat. If we plan to apply the label via a mutator, this controller behavior should not be needed.

brandboat (commented):
Thanks for the update, @Hero2633!

Could you please also file a PR to the https://github.com/harvester/docs repo? We need to update the [documentation](https://docs.harvesterhci.io/v1.8/image/upload-image) to notify users that displayImageName must be unique within the same namespace to avoid conflicts.

Also, it looks like the current CI failure is related to the changes in this PR, could you take a look and see if it needs a quick fix or a rebase?

WebberHuang1118 (commented):
Thanks for the PR, @Hero2633. LGTM overall.

One minor concern: the Harvester documentation does not explicitly state that a VM image’s display name cannot be duplicated in the [VM Images](https://docs.harvesterhci.io/v1.7/image/upload-image/) section.

As discussed with @brandboat,  @Hero2633 would you mind also adding a [Harvester docs](https://github.com/harvester/docs) update along with the change introduced in this PR?

Hero2633:
@brandboat @WebberHuang1118 ,OK

pkg/controller/master/image/vm_image_controller.go
WebberHuang1118: @Hero2633, do you have any thoughts on the comment above?

Hero2633: Hi @WebberHuang1118 @brandboat,

Thank you both for the great suggestion! Moving the logic to a mutator is indeed a brilliant idea—it makes the architecture much cleaner and more robust. 

The currently submitted PR now includes the modifications for the mutator. Here is a quick summary of the current changes:

**1. Mutator Injection**
I've implemented the mutating webhook to inject the `imageDisplayName` label right at creation time. This ensures the validating webhook can easily detect duplicates via label selectors. 
*(See code below)*:

https://github.com/Hero2633/harvester/blob/a71fddf5acaaa2522049a797327b363a25b8dde1/pkg/webhook/resources/virtualmachineimage/mutator.go#L53-L78

**2. Controller Patching (Kept for Backward Compatibility)**
I deliberately kept the patching logic in the controller's `OnChanged` hook in the following commit:

https://github.com/Hero2633/harvester/blob/a71fddf5acaaa2522049a797327b363a25b8dde1/pkg/controller/master/image/vm_image_controller.go#L27-L77

This is strictly for **backward compatibility**. If a user upgrades from an older version (e.g., `v1.7.1`), any existing failed images missing the label will be silently patched during the informer's startup resync. Without this, the deadlock bug could resurface for legacy images.

**3. Documentation Update**
I have also updated the corresponding documentation to clarify the image display name uniqueness requirements. You can find the doc PR here: harvester/docs#1028.

All CI checks have passed. Could you please take a look at both the code and doc changes when you have a chance? Thanks!

brandboat (approved):

Hero2633:
@Mergifyio queue

pkg/controller/master/image/vm_image_controller.go
WebberHuang1118: IMO, it would make the logic cleaner to have the mutator handle updating the display name label consistently. For failed VMIs with a missing label, I don’t think there is a need to keep separate duplicated logic for that case, as it makes the mechanism a bit confusing.

pkg/webhook/resources/virtualmachineimage/mutator.go
```diff
@@ -58,6 +59,15 @@ func (m *virtualMachineImageMutator) Create(_ *types.Request, newObj runtime.Obj
...
 	}
 
+	// Inject the imageDisplayName label at creation time so that the validating
+	// webhook's CheckDisplayName can always detect duplicate display names via
+	// label selector, regardless of whether the image upload succeeds or fails.
+	if len(vmi.Labels) == 0 {
+		patchOps = append(patchOps, `{"op": "add", "path": "/metadata/labels", "value": {}}`)
+	}
+	escapedLabel := patch.EscapeJSONPointer(util.LabelImageDisplayName)
+	patchOps = append(patchOps, fmt.Sprintf(`{"op": "add", "path": "/metadata/labels/%s", "value": "%s"}`, escapedLabel, vmi.Spec.DisplayName))
```
WebberHuang1118, line 69: Also need a validator to ensure the spec’s display name is immutable; otherwise, it may cause inconsistency between the spec and the label.

pkg/controller/master/image/vm_image_controller.go
Hero2633: Hi @WebberHuang1118,

Thank you for the feedback! Here are my thoughts on the two points:

1. Regarding the Validator Call Chain:
Actually, this restriction is already implemented and active in the current PR (leveraging the inherent mechanism from v1.7.1). For update requests, the VirtualMachineImage validating webhook enters the Update path at:
[pkg/webhook/resources/virtualmachineimage/validator.go#L68](https://github.com/Hero2633/harvester/blob/a71fddf5acaaa2522049a797327b363a25b8dde1/pkg/webhook/resources/virtualmachineimage/validator.go#L65-L69)

It then dispatches to the backend-specific update logic, for example:
[pkg/image/backingimage/validate.go#L36](https://github.com/Hero2633/harvester/blob/a71fddf5acaaa2522049a797327b363a25b8dde1/pkg/image/backingimage/validate.go#L69-L71)

This path calls CheckUpdateDisplayName(...), which compares the old vs. new spec.displayName and rejects the request if a change is detected (see [pkg/image/common/validator.go#L109](https://github.com/Hero2633/harvester/blob/a71fddf5acaaa2522049a797327b363a25b8dde1/pkg/image/common/validator.go#L109-L114)).

I have also manually verified this using kubectl edit; the system correctly rejects any attempt to modify the displayName. Therefore, this PR already ensures displayName immutability.

2. Regarding the Controller Logic:
Respectfully, I still hold a different view on this point. I believe a bug fix should be comprehensive and "closed-loop." If we remove this controller logic, users upgrading from older versions (like v1.7.1) will still encounter deadlocks for their existing, failed images that lack the necessary labels.

I find it difficult to compromise system robustness for the sake of absolute code cleanliness when it means leaving a known bug in the user's legacy data. To address the concern about potential confusion, I will add detailed comments to the controller code to clearly explain that this logic is required for backward compatibility and to handle specific upgrade scenarios.

Thanks again for your patient review!

brandboat: > I believe a bug fix should be comprehensive and "closed-loop." If we remove this controller logic, users upgrading from older versions (like v1.7.1) will still encounter deadlocks for their existing, failed images that lack the necessary labels.

What about we skip the validation when deletionTimestamp is not zero? c.c. @WebberHuang1118 

```go
func (v *virtualMachineImageValidator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) error {
	newVMI := newObj.(*v1beta1.VirtualMachineImage)
	oldVMI := oldObj.(*v1beta1.VirtualMachineImage)
	if !newVMI.DeletionTimestamp.IsZero() {
		return nil
	}
	return v.validators[util.GetVMIBackend(oldVMI)].Update(oldVMI, newVMI)
}
```

Yu-Jack: As I know and tested, any successful virtual images should have `harvesterhci.io/imageDisplayName` label. 

That means we don't need to update the label of any existed virtual images in `OnChanged` because those images should be normal images. So, mutator should be enough.

For the deletion, I think @brandboat 's thought would be a solution. From my perspective, the issue seems like when deleting a virtual machine image, and it also triggers the update validator because patching some deletion metadata. Then, the update validator blocks the request.

Unless we need to check something before deleting a virtual machine image. Otherwise, I think just returning nil would be better.

Image

Hero2633: Hi @WebberHuang1118,@Yu-Jack

Thank you for the guidance. However, I’ve identified two significant risks with the current top-level short-circuiting approach in the `Update` validator:

**Risk 1: Over-broad Impact (Global Bypass)**
Implementing the `DeletionTimestamp` check at the top level of the `Update` function is a "one-size-fits-all" solution. It creates a validation gap where **all** critical Spec checks are bypassed once an object is in a `Terminating` state. This could lead to inconsistent data if the object is modified during deletion. 

**Risk 2: The "Upgrade Blind Spot" (Critical for v1.7.1 Users)**
This is the most concerning part. The current fix relies entirely on the presence of labels, but it doesn't account for legacy data:
1. A user on **v1.7.1** has a failed `ubuntu.iso` (which lacks the `imageDisplayName` label).
2. The user upgrades to the version with this new fix.
3. The user uploads a new `ubuntu.iso` with the same name.
4. The **Validator fails to intercept** the new upload because the label selector cannot find the legacy `ubuntu.iso` (since it has no label).
5. **Result**: Two duplicate images exist in the same namespace, which breaks the uniqueness constraint.

**Conclusion:**
This is why I strongly advocate for keeping the controller's patching logic. The controller is the only mechanism that can "backfill" missing labels for legacy data during a startup resync. Without it, the fix is incomplete for existing users.

I would suggest moving the `DeletionTimestamp` check into specific validation functions (like `CheckDisplayName`) instead of the top-level `Update`, and keeping the controller logic to ensure a truly robust, closed-loop fix.

What do you think?

Yu-Jack: > A user on v1.7.1 has a failed ubuntu.iso (which lacks the imageDisplayName label).

I have one question:  `Users should delete those failed images, right? Do we really have that case?`.  I'm asking this because  it's not a upgrade case in your case. So, I was wondering if we really have this case. For most users, I believe they will delete the failed image after they find it failed. If we do have this case, we can keep it.

> Yu-Jack: Unless we need to check something before deleting a virtual machine image. Otherwise, I think just returning nil would be better.
> Hero2633: I would suggest moving the DeletionTimestamp check into specific validation functions (like CheckDisplayName) instead of the top-level Update

Actually, if we really need to validate whether the resource being deleted requires more validation, I think we can move DeletionTimestamp check into specific function. But, if not, we can probably put it on the top level.

Hero2633: Hi @Yu-Jack,

You are absolutely right—a single failed image can indeed be deleted in v1.7.1.

However, my primary concern shifted from "deletion capability" to "data consistency." If we rely solely on the proposed short-circuiting at the top level, we leave a loophole for the uniqueness constraint:

The Scenario: A user on v1.7.1 has a failed ubuntu.iso (which lacks the imageDisplayName label). The user might forget to delete it and proceeds with the upgrade.

The Logic Gap: After upgrading, the user uploads a new ubuntu.iso. Since the legacy image lacks the label, our Webhook's Label Selector will find zero matches and allow the creation.

The Result: We end up with duplicate displayNames within the same namespace, which violates a core constraint of Harvester.

While users should clean up failed images, a robust system shouldn't rely on "best-effort" human behavior. Our goal should be to ensure that the "one name per namespace" rule is enforced across all data, including legacy resources.

The current proposed fix addresses the "deletion deadlock" (the symptom), but it doesn't fully resolve the "uniqueness violation" (the root cause) for upgraded clusters. This is why I advocate for the controller's patching logic—it is the only way to ensure legacy data is correctly indexed and visible to our new validation logic.

brandboat: The goal is to allow users to delete failed VM images without friction. In my opinion, simply loosening the validation check should be enough. Regarding consistency, I don't see the value in labeling failed images, furthermore, we can already have multiple failed images sharing the same displayName without that label, right? Given that, do you think adding the label is actually necessary here?

Hero2633: "I see your point. Since duplicate displayNames are already allowed for unlabeled failed images, I have no objections to this approach."

WebberHuang1118: If we use the mutator to update the display name label, then regardless of whether the VM Image succeeds or fails afterward, deletion will not be blocked, starting from the version that includes this mutator mechanism.

For older versions before this fix, if there is any failed VM Image, the user can simply apply the workaround to remove the resource finalizer. The failed VM Image should then be deletable, since it does not contain valid data. From my perspective, it makes sense to update the documentation to describe this workaround for this kind of legacy failed VM Image.

This also allows us to keep the logic clean by having only the mutator update the display name label.

WebberHuang1118: BTW, if we agree with the approach above, IMO, we don’t need to specifically handle the case where the deletion timestamp is not zero in this [discussion](https://github.com/harvester/harvester/pull/10113#discussion_r3159081470).
```
	if !newVMI.DeletionTimestamp.IsZero() {
		return nil
	}
```

brandboat:
Hi @Hero2633, just following up on this. Let me know if you’ve had a chance to look into it, or if you're running into any roadblocks. Thanks!

Hero2633:
@brandboat Thanks for the approval! I don't have the permission to trigger mergify. Could you please help me add this PR to the merge queue?

pkg/controller/master/image/vm_image_controller.go
```diff
@@ -29,17 +29,21 @@ func (h *vmImageHandler) OnChanged(_ string, vmi *harvesterv1.VirtualMachineImag
...
+		return h.vmiClient.Update(toUpdate)
+	}
```
brandboat: Please remove the controller logic for label syncing, and then this looks good to go. As discussed in [this comment](https://github.com/harvester/harvester/pull/10113#discussion_r3160073114), we want to avoid handling this in the controller.

brandboat (commented):
@Hero2633 Glad to have you back, and thanks for taking care of this! I just have one last comment, and then we're good to go.

pkg/controller/master/image/vm_image_controller.go
Hero2633: OK! I've pushed the latest commit!

brandboat (approved):

pkg/webhook/resources/virtualmachineimage/mutator.go
brandboat: @Hero2633, could you please help address this comment?

Hero2633: @brandboat  Thanks for the reminder. This has been addressed by CheckUpdateDisplayName, which rejects updates when spec.displayName changes.

Vicente-Cheng (approved):
thanks

WebberHuang1118 (approved):
LGTM, thanks.

brandboat:
@Mergify backport v1.7 v1.8

brandboat:
@Mergify backport v1.9

Issue #8295, opened 2025-05-15
[ENHANCEMENT] prevent the multiple blockdevice CR for the same disk


Vicente-Cheng:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

We encounter the edge case that the wwn is changed after the system upgrade.
For example:
```
lrwxrwxrwx 1 root root 13 2024-12-12 15:01 nvme-eui.01000000010000005cd2e44370345351 -> ../../nvme0n1
convert to -->
lrwxrwxrwx 1 root root 13 2025-05-12 16:48 nvme-nvme.8086-50484c4e313038303031333836503443474e-44656c6c204578707265737320466c617368204e564d6520503436313020362e34544220534646-00000001 -> ../../nvme0n1
```

We now use the wwn to avoid the duplicated disk, but it seems the wwn might not be very reliable in the edge case.
We should improve that with more references like serial number + vendor ...etc
To prevent the NDM from creating another blockdevice CR.

### Describe the solution you'd like

Consider more references like wwn/serial number to skip the redundant/duplicated BD (for the same underlying disk)

tserong:
https://github.com/harvester/node-disk-manager/pull/266 should take care of this (I'll probably end up closing this issue as a duplicate of https://github.com/harvester/harvester/issues/6261)

Issue #6261, opened 2024-08-05
[ENHANCEMENT] Improve the non-wwn device operations


Vicente-Cheng:
**Is your enhancement related to a problem? Please describe.**
The NDM (node-disk-manager) depends on the wwn or filesystem uuid. The traditional SCSI devices all have wwn but the NVME device did not have the wwn.
That means we need to create the filesystem on the NVME device so that the NDM can identify it.
The NVME device would be widely used on harvester clusters to improve the deployment steps. We should give a better way to handle the name device.

**Describe the solution you'd like**
We should create a new mechanism to identify each device on the node. And it could be applied to large-scale deployment.

**Describe alternatives you've considered**
None

**Additional context**
Here is the udev info on the NVME device
```
# udevadm info /dev/sdb
P: /devices/pci0000:49/0000:49:01.0/0000:4a:00.0/host0/target0:0:276/0:0:276:0/block/sdb
N: sdb
L: 0
S: disk/by-id/scsi-SNVMe_Dell_NVMe_ISE_PS_BICCN0374I050173Z
S: disk/by-id/scsi-2ace42e0036104985
S: disk/by-path/pci-0000:4a:00.0-scsi-0:0:276:0
E: DEVPATH=/devices/pci0000:49/0000:49:01.0/0000:4a:00.0/host0/target0:0:276/0:0:276:0/block/sdb
E: DEVNAME=/dev/sdb
E: DEVTYPE=disk
E: MAJOR=8
E: MINOR=16
E: SUBSYSTEM=block
E: USEC_INITIALIZED=28250607
E: DONT_DEL_PART_NODES=1
E: ID_PATH=pci-0000:4a:00.0-scsi-0:0:276:0
E: ID_PATH_TAG=pci-0000_4a_00_0-scsi-0_0_276_0
E: SCSI_TPGS=0
E: SCSI_TYPE=disk
E: SCSI_VENDOR=NVMe
E: SCSI_VENDOR_ENC=NVMe\x20\x20\x20\x20
E: SCSI_MODEL=Dell_NVMe_ISE_PS
E: SCSI_MODEL_ENC=Dell\x20NVMe\x20ISE\x20PS
E: SCSI_REVISION=.0.0
E: ID_SCSI=1
E: ID_SCSI_INQUIRY=1
E: SCSI_IDENT_SERIAL=BICCN0374I050173Z
E: SCSI_IDENT_LUN_EUI64=ace42e0036104985
[...]

Issue #9900, opened 2026-01-21
[BUG] Disk disappears and re-enumerates (/dev/sdd → /dev/sdo) after SAS disconnect; NDM remounts without cleaning old mount leading to duplicate mounts on same extra-disks directory and Longhorn/Backing Image /data becomes unusable


liuwenyu2:
### Describe the Bug

On a Harvester node, a data disk temporarily disappears at the kernel/HBA layer (DID_NO_CONNECT), is removed as /dev/sdd, and then reappears as /dev/sdo (same WWN). Harvester Node Disk Manager (NDM) detects the devPath change and mounts the disk back to the same /var/lib/harvester/extra-disks/<id> directory, but the old/stale mount for /dev/sdd is not cleaned up. This results in the same mountpoint showing two sources (old /dev/sdd and new /dev/sdo). After that, Longhorn Backing Image (and/or related Longhorn components) that use the host disk path mapped to /data can hit I/O errors; running ls under /data fails.

<img width="397" height="97" alt="Image" src="https://github.com/user-attachments/assets/714684a1-b955-4b62-ac63-38e5e4921319" />

This looks like NDM may be over-mounting the new device on top of an old mount without ensuring the old mount is fully removed, leaving stale mount state and causing subsequent I/O errors in Longhorn/backing-image-manager.

### Impact

“This can lead to stale mounts and stale file handles, and once /data becomes unusable Longhorn may require replica rebuild or backing image re-download, which is disruptive.”

“In HCI production environments, transient disk path flaps are realistic; the system should fail safe and recover deterministically.”

### Reproduce

What happened (timeline)
At 2026-01-04 11:28:43:
         Kernel reports I/O timeout / path disconnect on /dev/sdd:
         hostbyte=DID_NO_CONNECT, command age ~33s
         mpt3sas_transport_port_remove removes the device/port and handle
        The same physical disk is immediately re-discovered and re-enumerated as /dev/sdo with the same SAS address/handle and WWN.

NDM logs show:
       udev remove for /dev/sdd
       udev add for /dev/sdo
      NDM recognizes the blockdevice is “coming back”, notes devPath changed /dev/sdd → /dev/sdo, and mounts /dev/sdo to the same extra-disks directory.
Afterwards:
     findmnt shows two sources for the same TARGET (old /dev/sdd and new /dev/sdo)
     Longhorn/backing-image-manager (mapped path /data) hits I/O errors; ls under /data fails.

Evidence / Logs
[...]

PR #10115, opened 2026-03-02
HEP: Rework BlockDevice naming to avoid requiring WWNs [skip ci]

tserong:
#### Problem:
We can't really cope properly with disks that don't have WWNs, or whose WWNs change (the latter should never happen, but it did at least once in the past when a kernel update resulted in IDs for a certain make and model of NVMe changing).

#### Solution:
This HEP :-)

#### Related Issue(s):
https://github.com/harvester/harvester/issues/6261
https://github.com/harvester/harvester/issues/8295

#### Test plan:
See details in HEP

#### Note:
See also https://github.com/harvester/node-disk-manager/pull/266 which has lots of comments in the code

### DIFF
--- a/enhancements/20260302-ndm-dont-require-wwn.md
+++ b/enhancements/20260302-ndm-dont-require-wwn.md
@@ -1,0 +1,117 @@
+# Rework BlockDevice naming to avoid requiring WWNs
+
+## Summary
+
+Harvester's node-disk-manager (NDM) component creates BlockDevice CRs which map to disks each host in the cluster. This allows Harvester to provision disks for use by Longhorn or LVM. The current naming for BD CRs is deterministic, based primarily on the WWN of the disk, or if a WWN is not present, falling back to a possible filesystem UUID. This means that Harvester can't really work with disks that don't have WWNs, as happens with some RAID controllers. It also results in problems if a disk's WWN changes - this shouldn't ever happen, but we've seen at least one case where WWNs _did_ change for a certain make and model of NVMe due to a kernel update. We need to rework how NDM identifies and tracks disks and BD CRs so that it doesn't require that devices have WWNs in order to operate, and so that it will gracefully handle the case where WWNs might change.
+
+### Related Issues
+
+- https://github.com/harvester/harvester/issues/6261
+- https://github.com/harvester/harvester/issues/8295
+
+## Motivation
+
+### Goals
+
+- Enable NDM to work with disks that don't have WWNs
+- Ensure NDM doesn't get confused when an existing disk's WWN is changed
+- Existing deployments (which will still be using the original naming scheme) must continue to work correctly after upgrading to the new version of NDM
+
+
+## Proposal
+
+- Decouple BlockDevice CR name generation from any actual property of the underying disk.
+- When scanning disks, identify and tie the disks back to BD CRs based on, in order of preference:
+  - What's actually on the disk, i.e. UUID (of filesystem or LVM volume) if present
+  - WWN if present
+  - Vendor+Model+Serial+BusPath if neither of the above are present
+
+### User Stories
+
+This proposal is internal to NDM so there aren't really any user stories aside from the fact that once implemented the user won't have to apply [workarounds](https://harvesterhci.io/kb/handle_disks_without_wwns) in order to use disks that don't have WWNs.
+
+## Design
+
+### Implementation Overview
+
+Update NDM's scanner so that it operates as follows:
+
+- Scan all disks in the system.
+  - For each disk:
+    - If it has a UUID:
+        - Check the list of existing BDs for that UUID, if there's a match we've found an existing device.
+    - If it has a WWN and hasn't already been found by UUID:
+        - Check the list of existing BDs for that WWN, if there's a match we've found an existing device.
+    - If the device doesn't match the above, check for an existing device by Vendor+Model+Serial+BusPath.  If this matches, we've found an existing device.
+  - If we've found an existing device by one of the above three methods:
+    - Check its status against the status of the disk we found, and update it to match if necessary.  This will pick up changed WWNs and re-activate provisioned devices that have gone away and come back.
+  - If we haven't found any existing BD, create a new BD CR for this disk.  The name of the new CR is just a randomly generated UUID.
+- Once the scan of all disks is complete, for each existing BD that _wasn't_ found:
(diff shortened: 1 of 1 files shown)


enhancements/20260302-ndm-dont-require-wwn.md
```diff
@@ -0,0 +1,104 @@
...
+- Once the scan of all disks is complete, for each existing BD that _wasn't_ found:
+  - If the BD was provisioned, mark it inactive (it might have been removed and be coming back later, or it's a multipath device and multipathd isn't running for whatever reason - it will be marked active automatically when it comes back online)
+  - If the BD wasn't provisioned, delete it
+
+Note some implications of the above, which are different to how NDM operated before:
+
+- BD naming is random and has no connection to any property of the disk.
+- BDs that haven't been provisioned will be removed from the list of BDs if the device is removed from the system.  If you add the same disk back later it will get a new random name.
+- BDs that _have_ been provisioned will remain in the list of BDs - they'll just become inactive if the device is removed (this is the same as the previous behaviour for provisioned BDs).
+
```
WebberHuang1118, line 58: Could you describe how NDM would handle an edge case like this scenario ([harvester#9900 comment](https://github.com/harvester/harvester/issues/9900#issuecomment-3777668873))?

Just my two cents: since it’s not practical to automatically recover from corner cases like this, providing a clear error message, stopping further operations, and having related documentation should be sufficient.

WebberHuang1118 (dismissed):
LGTM, just a question, thanks.

enhancements/20260302-ndm-dont-require-wwn.md
```diff
@@ -0,0 +1,104 @@
...
+- Existing deployments (which will still be using the original naming scheme) must continue to work correctly after upgrading to the new version of NDM
+
+
+## Proposal
+
+- Decouple BlockDevice CR name generation from any actual property of the underying disk.
+- When scanning disks, identify and tie the disks back to BD CRs based on, in order of preference:
+  - What's actually on the disk, i.e. UUID (of filesystem or LVM volume) if present
+  - WWN if present
+  - Vendor+Model+Serial+BusPath if neither of the above are present
```
bk201, line 27: If a user intentionally swaps a disk without wwn and the disk comes back with a different buspath, I guess the behavior is no harm because:
- If the disk is non-provisioned, then it's OK to have a different CR.
- If the disk is provisioned, then it is certainly already found at the first step.
Just want to check if my theory is correct.

tserong: Yep, correct.

bk201 (dismissed):

enhancements/20260302-ndm-dont-require-wwn.md
tserong: I've done some investigation of that issue (see analysis at https://github.com/harvester/harvester/issues/9900#issuecomment-3995162113) and this new implementation won't change how that edge case currently behaves.

I absolutely agree that weird cases like that need to be handled better and logged properly, but that specific issue is occuring inside NDM's LHv1 provisioner code (which handles the mounting/unmounting), and isn't directly touched by the implementation of this HEP, so I think we should try to handle that as a separate work item. I am happy to take that issue and work on it if nobody else beats me to it BTW :-)

enhancements/20260302-ndm-dont-require-wwn.md
```diff
@@ -0,0 +1,104 @@
...
+
+Note some implications of the above, which are different to how NDM operated before:
+
+- BD naming is random and has no connection to any property of the disk.
+- BDs that haven't been provisioned will be removed from the list of BDs if the device is removed from the system.  If you add the same disk back later it will get a new random name.
+- BDs that _have_ been provisioned will remain in the list of BDs - they'll just become inactive if the device is removed (this is the same as the previous behaviour for provisioned BDs).
+
+### Test plan
+
+TBD
```
innobead: @tserong still needs to come out with the best coverage plan for testing for @harvester/qa reference.

enhancements/20260302-ndm-dont-require-wwn.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Note
+
+Recall that the scanner gives preference to "what's actually on the disk", which for a provisioned LHv1 volume or LVM volume is a UUID.  I had hoped to be able to detect LHv2 provisioned volumes in a similar way, but I've hit a wall in that LHv2 volumes don't seem to have any sort of ID that's easy to extract without starting up SPDK somehow and interrogating the device.
+
+There _is_ data we could interrogate, notably the signature `SPDKBLOB` at the start of the device, and, later a name which actually _does_ match a BD CR (in the example below it's `2e303caa-89ab-404a-902b-85fccdf2a189`), but we'd need to write some tool that knows how to walk SPDK blobs to extract that data.  That's still not entirely relabile though because (for NVMe devices at least), once LH has activated them and SPDK has taken over, the device node (/(dev/nvme0xxx) disappears and we can't read that disk anymore.
+
+TL;DR: right now, LHv2 volumes will be checked by WWN, and if there is no WWN it will fall back to Vendor+Model+Serial+BusPath.  It's not exactly what I wanted but it'll still work.
```
innobead, line 86: @derekbit, do we store disk or FS info in Longhorn node/disk CR for Longhorn disks in V2? If yes, @tserong might just use the info stored in Longhorn CR instead of scanning at runtime, since Longhorn already manages them.

Vicente-Cheng (dismissed):
lgtm, thanks for the great idea!

enhancements/20260302-ndm-dont-require-wwn.md
derekbit: We store the block-type disk UUID from spdk in the `node.status.diskStatus`.
For tne node.spec.disks, we only store disk name and path.

tserong: The disk name and path in node.spec.disks will be the same information we already have in harvester's BD CRs.

Is there a way of reading the block-type disk UUID from spdk directly from the disk somehow? If so, I could use that in the same way I use FS UUID for LHv1 disks and LVM volumes to identify the disk once it's provisioned.

votdev (dismissed):
LGTM

tserong:
Rebased and added test plan - is this sufficiently detailed?

bk201 (approved):

PR #10117, opened 2026-03-02
Refactor createOrUpdateNodeCondition to do not modify the original slice

votdev:
- Additionally add unit tests to increase code coverage.
- Introduce public API to find and update node conditions.

### DIFF
--- a/pkg/controller/master/node/node_down_controller.go
+++ b/pkg/controller/master/node/node_down_controller.go
@@ -85,22 +85,11 @@ func (h *nodeDownHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev1.No
 	return node, nil
 }
 
-func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.NodeConditionType) *corev1.NodeCondition {
-	var cond *corev1.NodeCondition
-	for i := range conditions {
-		if conditions[i].Type == conditionType {
-			cond = &conditions[i]
-			break
-		}
-	}
-	return cond
-}
-
 // We treat the node not ready (False/Unknown) as node down
 // In this situation, we will wait for the extra timeout and then add taints to trigger migration and cleanup
 func (h *nodeDownHandler) checkNodeReady(node *corev1.Node) error {
 	// get Ready condition
-	cond := getNodeCondition(node.Status.Conditions, corev1.NodeReady)
+	cond := util.FindNodeStatusCondition(node.Status.Conditions, corev1.NodeReady)
 	if cond == nil {
 		return fmt.Errorf("can't find %s condition in node %s", corev1.NodeReady, node.Name)
 	}
@@ -126,7 +115,7 @@ func (h *nodeDownHandler) checkNodeReady(node *corev1.Node) error {
 		}
 
 		// try and check if condition exists
-		drainCondition := getNodeCondition(node.Status.Conditions, HarvesterNodeCondDrained)
+		drainCondition := util.FindNodeStatusCondition(node.Status.Conditions, HarvesterNodeCondDrained)
 		// both taints exist so lets mark condition as true
 		// check if both taints exist
 		hasKubevirtDrainTaint := getNodeTaint(node.Spec.Taints, virtconfig.NodeDrainTaintDefaultKey) != nil
@@ -160,7 +149,7 @@ func (h *nodeDownHandler) checkNodeReady(node *corev1.Node) error {
 		}
 
 		// reset HarvesterNodeFailure condition if needed
-		cond := getNodeCondition(node.Status.Conditions, HarvesterNodeCondDrained)
+		cond := util.FindNodeStatusCondition(node.Status.Conditions, HarvesterNodeCondDrained)
 		if cond != nil {
 			// remove the condition if exists
 			logrus.Infof("Removing HarvesterNodeDrained condition from node %s", node.Name)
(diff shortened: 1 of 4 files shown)


Vicente-Cheng (dismissed):
good catch!

pkg/controller/master/node/node_down_controller.go
```diff
@@ -88,9 +88,8 @@ func (h *nodeDownHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev1.No
 func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.NodeConditionType) *corev1.NodeCondition {
 	var cond *corev1.NodeCondition
 	for i := range conditions {
-		c := conditions[i]
-		if c.Type == conditionType {
-			cond = &c
+		if conditions[i].Type == conditionType {
+			cond = &conditions[i]
 			break
```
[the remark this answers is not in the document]
votdev: Refactored the function based on this tip.

pkg/util/node.go
```diff
@@ -96,3 +97,53 @@ func CountNonWitnessNodes(nodes []*corev1.Node) int {
...
+			if c.Type == condType {
+				node.Status.Conditions[id] = newCond
+			}
+		}
+	}
+}
+
+// GetNodeCondition returns the condition of the node based on the condition type.
+// If the condition doesn't exist, it will return nil.
+func GetNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.NodeConditionType) *corev1.NodeCondition {
```
w13915984028: To double check: use a public function to get a pointer of a slice internal elements, is a bit error-prone, data might be changed unexpectedly

instead, return a full copy of a target element with a bool param to indicate if it finds/not;  and use an input condition to update/append to a slice is also safe

```
GetNodeCondition() (bool, corev1.NodeCondition)
SetNodeCondition(newCondition corev1.NodeCondition)
```

it is similar to `map` operations of get/set

The robustness weigh over the small performance loss

votdev: I fully agree with your arguments, but I don't want to touch and change the code that has been calling the (previously) local function. I would also like to keep the option open with this function that it could be used in the future in another place where the current behavior is set, namely that I get a pointer to the original condition in order to modify it.

The current function signature is equal to the functions in the meta package:
- https://github.com/kubernetes/apimachinery/blob/v0.35.2/pkg/api/meta/conditions.go#L91
- https://github.com/kubernetes/apimachinery/blob/v0.35.2/pkg/api/meta/conditions.go#L31

I've modified the function description to "inform/warn" the developer which is using it.

w13915984028 (commented):
This PR mixes bug fix and code refactor, adds some public APIs with certain concerns on exposing internal pointers.

From the perspective of v180, I would suggest to strictly limit the scope on bug fix.

In case others think this PR is also fit for v180, I am also OK. thanks.

pkg/util/node.go
w13915984028: Meta v1 has more functions on this file, it could also due to the back-compatibility to keep some unsafe functions, and metav1 also does not take the responsibility to ensure the caller is always correct.

I am a bit reluctant to add them to public APIs.

votdev:
> This PR mixes bug fix and code refactor, adds some public APIs with certain concerns on exposing internal pointers.
> 
> From the perspective of v180, I would suggest to strictly limit the scope on bug fix.
> 
> In case others think this PR is also fit for v180, I am also OK. thanks.

Fixing the error of the local slice copy implies the need to correct the `createOrUpdateNodeCondition` function.

The newly added API functions behave no differently than similar functions in the Kubernetes Go API or helper functions in Harvester or [VM Importer](https://github.com/harvester/vm-import-controller/blob/4bb4e1e96e742e2f94d6468e3d26c69a20015852/pkg/util/conditions.go#L10), for example.

pkg/controller/master/node/node_down_controller.go
```diff
@@ -85,23 +85,11 @@ func (h *nodeDownHandler) OnNodeChanged(_ string, node *corev1.Node) (*corev1.No
...
 }
 
-func getNodeCondition(conditions []corev1.NodeCondition, conditionType corev1.NodeConditionType) *corev1.NodeCondition {
-	var cond *corev1.NodeCondition
-	for i := range conditions {
-		c := conditions[i]
-		if c.Type == conditionType {
-			cond = &c
-			break
-		}
```
ihcsim, line 95: I suspect the original intention is to _not_ modify the original `condition` object, which makes sense since the `node` came from the indexer cache. AFAICT, the only place where the `condition` returned by `getNodeCondition()` is modified is in the `createOrUpdateNodeCondition()` method. That should already be working. We shouldn't modify the `condition` of a cached `node` object.

Re: the upstream apimachinery `FindNodeStatusCondition()`, from what I can see in the K8s src, the returned object is mainly used either for read-only execution, or write execution on a cloned `node`, but not directly to a cached `node`.

pkg/controller/master/node/node_down_controller.go
w13915984028, line 124: BTW:

the current code, does not check `drainCondition == nil` but access `drainCondition.Status` directly;  maybe there is an implicit assumption: when `hasKubevirtDrainTaint && hasOutOfServiceTaint`, the `drainCondition` is not nil

it is a bit headache to get a pointer then play on it, from time to time we might be trapped

https://github.com/harvester/harvester/blob/e1e0eeba11f8f6fdb3979656aa9f03b45448524d/pkg/controller/master/node/node_down_controller.go#L129-L140

```
		drainCondition := getNodeCondition...
...

			if drainCondition.Status == corev1.ConditionFalse {
```

When we define them as public functions, even there is warning on comment, it still does not block any of such usage

pkg/controller/master/node/node_down_controller.go
votdev: > I suspect the original intention is to not modify the original condition object, which makes sense since the node came from the indexer cache.

It is somewhat difficult to discern this intention from the code. It could simply be an oversight or an unintentional error. If it were intentional, I would expect a comment in the code or documentation of the function.

votdev:
My approach to programming is to be forward-thinking and deliver general-purpose solutions. For the functions in question, I do not return copies of the conditions because if the developer wants to, they can create a copy of the node resource. However, if they want the pointer to the condition, then a function that returns a copy is of no use and ultimately results in duplicate code. Therefore, the approach of returning the pointer to the condition is the more universal solution that covers both requirements.

If the majority here cannot agree with this approach, then I can of course rewrite the functions.

pkg/controller/master/node/node_down_controller.go
ihcsim: That's the only reason I can think of why there is a `c := conditions[i]` which results in a "hard" copy of the struct. Yeah, I wouldn't have known without looking closely at the code that `getNodeCondition()` returns a clone.

My only thought right now is that user of the `SetNodeStatusCondition()` method needs to be aware that they should pass in the clone, not the cache, of the node, as you did in `createOrUpdateNodeCondition()`. My question on it is do we actually need this new method since it isn't used anywhere else? If we want a general `util` method for this, I think it's safer and cleaner to just make `createOrUpdateNodeCondition()` public (or has an `util` equivalence) so that we don't end up passing around a modified, but not persisted node object. IOW, what do I do with that node returned by `SetNodeStatusCondition()` - do I need to persist it or not?

Otherwise, the rest LGTM.

votdev:
The function `FindNodeStatusCondition` returns now a copy of the condition. Therefore, the caller can now do anything with it without running the risk of modifying the cache object.
The PR does not change anything fundamental in the original code, but it does make the code a little clearer and offers a public API that can be used by others. In addition, the `SetNodeStatusCondition` help function is a little tidier and more logical. Finally, there are now more unit tests available to test and ensure functionality.

ihcsim (approved):

votdev:
@Mergifyio backport v1.8

Issue #10110, opened 2026-02-27
[TASK] Bump golang/k8s/BCI Image for v1.8.0


Vicente-Cheng:
### Task request

Bump the following components:

- golang -> v1.25.6/v1.25.7
- k8s -> v1.34 or v1.33 (depends on the kubevirt/client-go)
- BCI image -> 16.0

| Done | Repository | Owner |
| --- | --- | --- |
| ✅ | https://github.com/harvester/cloud-provider-harvester note: the chart&image is OOB from Harvester v180 release | @w13915984028   |
| ✅ | https://github.com/harvester/docker-machine-driver-harvester | @wheatdog   |
| ✅ | https://github.com/harvester/eventrouter | @martindekov  |
| ✅ | https://github.com/harvester/go-common | @Vicente-Cheng   |
| ✅ | https://github.com/harvester/harvester | @Yu-Jack  |
| ✅ | https://github.com/harvester/harvester-csi-driver | @Vicente-Cheng |
| ✅ | https://github.com/harvester/networkfs-manager | @Vicente-Cheng |
| ✅ | https://github.com/harvester/csi-driver-lvm | @Vicente-Cheng |
| ✅ | https://github.com/harvester/harvester-installer | @tserong |
| ✅ | https://github.com/harvester/load-balancer-harvester | @martindekov   |
| ✅ | https://github.com/harvester/network-controller-harvester | @starbops  |
| ✅ | https://github.com/harvester/node-disk-manager | @Vicente-Cheng |
| ✅  | https://github.com/harvester/node-manager | @brandboat  |
| ✅ | https://github.com/harvester/pcidevices | @Yu-Jack  |
| ✅ | https://github.com/harvester/seeder | @ibrokethecloud |
| ✅ | https://github.com/harvester/rancherd | @wheatdog  |
| ✅| https://github.com/rancher/support-bundle-kit | @Yu-Jack  |
| ✅ | https://github.com/harvester/vm-dhcp-controller | @starbops   |
| ✅ | https://github.com/harvester/vm-import-controller | @votdev   |
| ✅ | https://github.com/harvester/terraform-provider-harvester | @wheatdog  |
| ✅ | https://github.com/harvester/storage-validator | @ibrokethecloud  |
| ✅ | https://github.com/harvester/kubeovn-operator | @ibrokethecloud  |

innobead:
[...]

PR #10122, opened 2026-03-03
chore: bump golang version and bci image

Yu-Jack:
https://github.com/harvester/harvester/issues/10110

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -1103,7 +1103,7 @@ spec:
   tolerations:
   - operator: "Exists"
   upgrade:
-    image: registry.suse.com/bci/bci-base:15.7
+    image: registry.suse.com/bci/bci-base:16.0
     command:
     - chroot
     - /host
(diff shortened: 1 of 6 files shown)


wheatdog (dismissed):
LGTM, thanks!

bk201 (dismissed):
Thank you!

starbops (dismissed):
LGTM, thank you.

starbops:
Installing the `util-linux` package can resolve the issue. However, I am not sure whether it brings too many unnecessary dependencies to the image.

package/upgrade/upgrade_manifests.sh
```diff
@@ -1103,7 +1103,7 @@ spec:
   tolerations:
   - operator: "Exists"
   upgrade:
-    image: registry.suse.com/bci/bci-base:15.7
+    image: registry.suse.com/bci/bci-base:16.0
```
Yu-Jack, line 1106: @starbops Will it have the issue if we don't have `mount` command here?

starbops: No, that SUC plan job does not require `mount`.

bk201 (approved):
nice catch!

Issue #7867, opened 2025-03-17
[FEATURE] Reduce Longhorn resource reservations for customers using 3rd party storage


asettle:
**Is your feature request related to a problem? Please describe.**

Customers using third-party storage solutions with SUSE Virt do not require Longhorn’s full resource reservations, leading to unnecessary CPU and memory consumption. This feature request proposes reducing Longhorn’s resource reservations when an external storage provider is in use, optimizing cluster resource utilization.

**Describe the solution you'd like**

Longhorn should detect when third-party storage is enabled and dynamically reduce its resource reservations accordingly. Users should have a configurable option to adjust Longhorn’s resource allocations based on their needs. Resource optimizations should not impact core Longhorn functionalities for users who still rely on it.

**Describe alternatives you've considered**

Users can manually adjust Longhorn’s CPU/memory requests and limits through custom configurations, but this is not officially supported.

**Additional context**

Longhorn currently reserves resources based on its default configurations, even when third-party storage is utilized. This leads to unnecessary resource consumption, impacting cluster efficiency and limiting available resources for other workloads. Other third-party storage providers may be needed to refine the optimal resource allocation settings.

innobead:
cc @Vicente-Cheng

innobead:
If users don't use Longhorn for VM volume disks, do we have any features relying on Longhorn? IIRC, the new upgrade repo VM is relying on RWX volumes provided by Longhorn. What else? @harvester/storage

WebberHuang1118:
> If users don't use Longhorn for VM volume disks, do we have any features relying on Longhorn? IIRC, the new upgrade repo VM is relying on RWX volumes provided by Longhorn. What else? [@harvester/storage](https://github.com/orgs/harvester/teams/storage)

As discussed with @Vicente-Cheng,

Built-in Rancher Monitoring, Rancher Logging, the upgrade logger, and the upgrade repository all rely on Longhorn.

Another aspect is the storage network. I remember someone asking whether NetApp storage can use a storage network; however, this is a Longhorn-specific feature designed to segregate storage data traffic.

For stability, in my opinion, we should keep the upgrade repository on Longhorn volumes.

innobead:
> Another aspect is the storage network. I remember someone asking whether NetApp storage can use a storage network; however, this is a Longhorn-specific feature designed to segregate storage data traffic.

[...]

PR #10125, opened 2026-03-03
feat: introduce a setting to control Longhorn Instance Manager (IM) resources

WebberHuang1118:
#### Problem:
Let Harvester users configure Longhorn Instance Manager (IM) CPU usage from Harvester (via either the UI or CLI).

#### Solution:
Introduce a new Harvester setting, `instance-manager-resources`, e.g.,
```
default: '{"cpu":{"v1":"12","v2":"12"}}'
kind: Setting
metadata:
  annotations:
    harvesterhci.io/hash: f70e0094431c520294b322a918d9dc0cc11a005d12a43dd45b758293
  creationTimestamp: "2026-03-03T07:52:31Z"
  generation: 7
  name: instance-manager-resources
  resourceVersion: "23115941"
  uid: b32cc7f2-a0fd-4042-bd11-5f34221b4e52
status:
  conditions:
  - lastUpdateTime: "2026-03-03T08:26:02Z"
    status: "True"
    type: configured
value: '{"cpu":{"v1":"20","v2":"20"}}'
```

This will sync to the Longhorn `guaranteed-instance-manager-cpu` setting for v1 and v2 IM CPU usage. 

Since Longhorn may extend IM resource controls in the future [longhorn#6351](https://github.com/longhorn/longhorn/issues/6351), this Harvester setting gives us flexibility for future needs—for example, if memory becomes configurable, we could extend this setting to include that as well.
```
default: '{"cpu":{"v1":"12","v2":"12"}, "memory":{"v1":"1G","v2":"2G"}}'
kind: Setting
metadata:
  annotations:
    harvesterhci.io/hash: f70e0094431c520294b322a918d9dc0cc11a005d12a43dd45b758293
  creationTimestamp: "2026-03-03T07:52:31Z"
  generation: 7
  name: instance-manager-resources
  resourceVersion: "23115941"
  uid: b32cc7f2-a0fd-4042-bd11-5f34221b4e52
status:
  conditions:
  - lastUpdateTime: "2026-03-03T08:26:02Z"
    status: "True"
    type: configured
value: '{"cpu":{"v1":"20","v2":"20"}, "memory":{"v1":"5G","v2":"5G"}}'
```

#### Related Issue(s):
#7867 

#### Test plan:
* Build Harvester v1.7.1 and update the Longhorn setting **Guaranteed Instance Manager CPU** ([docs](https://longhorn.io/docs/1.11.1/references/settings/#guaranteed-instance-manager-cpu)) to a specific value, e.g., `{"v1":"10","v2":"10"}`.
- Update the Harvester with ISO from this PR
- Longhorn setting **Guaranteed Instance Manager CPU** is not changed
- Create a running VM and try to update the setting—it should be rejected, e.g.,
  ```
  $ kubectl patch setting.harvesterhci.io instance-manager-resources   --type=merge   -p '{"value":"{\"cpu\":{\"v1\":\"20\",\"v2\":\"20\"}}"}'` 
  The request is invalid: value: please detach all Longhorn volumes before configuring instance-manager-resources
  ```
- Make sure there are no running VMs or attached Longhorn volumes.
- Update the setting again—it should succeed this time
  ```
  $ kubectl patch setting.harvesterhci.io instance-manager-resources   --type=merge   -p '{"value":"{\"cpu\":{\"v1\":\"20\",\"v2\":\"20\"}}"}'
  setting.harvesterhci.io/instance-manager-resources patched
  ```
- * Reset `setting.harvesterhci.io/instance-manager-resources` to the default value. The Longhorn **Guaranteed Instance Manager CPU** setting ([docs](https://longhorn.io/docs/1.11.1/references/settings/#guaranteed-instance-manager-cpu)) will then be set back to the default: `{"v1":"12","v2":"12"}`.

### DIFF
--- a/pkg/controller/master/setting/instance_manager_resources.go
+++ b/pkg/controller/master/setting/instance_manager_resources.go
@@ -1,0 +1,125 @@
+package setting
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhtypes "github.com/longhorn/longhorn-manager/types"
+	"k8s.io/apimachinery/pkg/labels"
+
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester/pkg/settings"
+	"github.com/harvester/harvester/pkg/util"
+)
+
+const msgWaitForAttachedVolumes = "waiting for all volumes detached: %s"
+
+func (h *Handler) syncLHIMResources(setting *harvesterv1.Setting) error {
+	resources, err := settings.DecodeLHIMResources(setting.Value)
+	if err != nil {
+		return fmt.Errorf("invalid JSON %q for %s: %w", setting.Value, setting.Name, err)
+	}
+	if resources.IsEmpty() {
+		// Ignore the initial default value so Harvester does not overwrite an
+		// existing Longhorn setting until the user has configured this setting
+		// with a meaningful value at least once.
+		if setting.Annotations[util.AnnotationHash] == "" {
+			return nil
+		}
+		return h.syncLHIMResourcesToLonghornDefault()
+	}
+	v1CPU := resources.CPU.V1
+	v2CPU := resources.CPU.V2
+
+	if err := lhtypes.ValidateCPUReservationValues(lhtypes.SettingNameGuaranteedInstanceManagerCPU, v1CPU); err != nil {
+		return fmt.Errorf("invalid CPU value for cpu.v1 %q in %s: %w", v1CPU, setting.Name, err)
+	}
+	if err := lhtypes.ValidateCPUReservationValues(lhtypes.SettingNameGuaranteedInstanceManagerCPU, v2CPU); err != nil {
+		return fmt.Errorf("invalid CPU value for cpu.v2 %q in %s: %w", v2CPU, setting.Name, err)
+	}
+
+	if err := h.checkLonghornVolumeDetached(); err != nil {
+		return err
+	}
+
+	lhSetting, err := h.longhornSettingCache.Get(util.LonghornSystemNamespaceName, string(lhtypes.SettingNameGuaranteedInstanceManagerCPU))
+	if err != nil {
+		return err
+	}
+
+	lhValueBytes, err := json.Marshal(map[string]string{
+		string(lhv1beta2.DataEngineTypeV1): v1CPU,
+		string(lhv1beta2.DataEngineTypeV2): v2CPU,
+	})
+	if err != nil {
+		return err
+	}
+
+	lhValue := string(lhValueBytes)
+	if lhSetting.Value == lhValue {
+		return nil
+	}
+
+	lhSettingCpy := lhSetting.DeepCopy()
+	lhSettingCpy.Value = lhValue
+	_, err = h.longhornSettings.Update(lhSettingCpy)
+	return err
+}
+
+func (h *Handler) syncLHIMResourcesToLonghornDefault() error {
+	lhSetting, err := h.longhornSettingCache.Get(util.LonghornSystemNamespaceName, string(lhtypes.SettingNameGuaranteedInstanceManagerCPU))
+	if err != nil {
+		return err
+	}
+
+	defaultValue, err := defaultLonghornLHIMResourcesValue()
+	if err != nil {
+		return err
+	}
+	if lhSetting.Value == defaultValue {
+		return nil
+	}
+
+	if err := h.checkLonghornVolumeDetached(); err != nil {
+		return err
+	}
+
+	lhSettingCpy := lhSetting.DeepCopy()
+	lhSettingCpy.Value = defaultValue
+	_, err = h.longhornSettings.Update(lhSettingCpy)
+	return err
+}
+
+func defaultLonghornLHIMResourcesValue() (string, error) {
+	valueBytes, err := json.Marshal(map[string]string{
+		string(lhv1beta2.DataEngineTypeV1): lhtypes.SettingDefinitionGuaranteedInstanceManagerCPU.Default,
+		string(lhv1beta2.DataEngineTypeV2): lhtypes.SettingDefinitionGuaranteedInstanceManagerCPU.Default,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return string(valueBytes), nil
+}
+
+func (h *Handler) checkLonghornVolumeDetached() error {
+	volumes, err := h.longhornVolumeCache.List(util.LonghornSystemNamespaceName, labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list longhorn volumes: %w", err)
+	}
+
+	attachedVolumes := make([]string, 0)
+	for _, volume := range volumes {
+		if volume.Status.State != lhv1beta2.VolumeStateDetached {
+			attachedVolumes = append(attachedVolumes, volume.Name)
+		}
+	}
+
+	if len(attachedVolumes) > 0 {
+		return fmt.Errorf(msgWaitForAttachedVolumes, strings.Join(attachedVolumes, ","))
+	}
+
+	return nil
+}
--- a/pkg/settings/settings_helper.go
+++ b/pkg/settings/settings_helper.go
@@ -6,6 +6,7 @@ import (
 	"reflect"
 	"strconv"
 
+	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	"github.com/sirupsen/logrus"
 
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
@@ -105,6 +106,78 @@ type Overcommit struct {
 	CPU     int `json:"cpu"`
 	Memory  int `json:"memory"`
 	Storage int `json:"storage"`
+}
+
+type LHIMCPUConfig struct {
+	V1 string `json:"v1"`
+	V2 string `json:"v2"`
+}
+
+type LHIMResourcesConfig struct {
+	CPU LHIMCPUConfig `json:"cpu"`
+}
+
+func (c *LHIMResourcesConfig) IsEmpty() bool {
+	return c == nil || (c.CPU.V1 == "" && c.CPU.V2 == "")
+}
+
+func DecodeLHIMResources(value string) (*LHIMResourcesConfig, error) {
+	if value == "" {
+		return &LHIMResourcesConfig{}, nil
+	}
+
+	raw := map[string]json.RawMessage{}
+	if err := json.Unmarshal([]byte(value), &raw); err != nil {
+		return nil, fmt.Errorf("unmarshal failed, error: %w, value: %s", err, value)
+	}
+	cpuRaw, ok := raw["cpu"]
+	if !ok {
+		return nil, fmt.Errorf("cpu is required")
+	}
+
+	cpuMap := map[string]json.RawMessage{}
+	if err := json.Unmarshal(cpuRaw, &cpuMap); err != nil {
+		return nil, fmt.Errorf("cpu must be an object with v1/v2, error: %w", err)
+	}
+	if len(cpuMap) == 0 {
+		return &LHIMResourcesConfig{}, nil
+	}
+
+	v1Raw, ok := cpuMap[string(lhv1beta2.DataEngineTypeV1)]
+	if !ok {
+		return nil, fmt.Errorf("cpu.v1 is required")
+	}
+	v2Raw, ok := cpuMap[string(lhv1beta2.DataEngineTypeV2)]
+	if !ok {
+		return nil, fmt.Errorf("cpu.v2 is required")
+	}
+
+	v1, err := decodeLHIMCPUValue(v1Raw)
+	if err != nil {
+		return nil, fmt.Errorf("invalid cpu.v1: %w", err)
+	}
+	v2, err := decodeLHIMCPUValue(v2Raw)
+	if err != nil {
+		return nil, fmt.Errorf("invalid cpu.v2: %w", err)
+	}
+
+	return &LHIMResourcesConfig{
+		CPU: LHIMCPUConfig{
+			V1: v1,
+			V2: v2,
+		},
+	}, nil
+}
+
+func decodeLHIMCPUValue(raw json.RawMessage) (string, error) {
+	var stringValue string
+	if err := json.Unmarshal(raw, &stringValue); err != nil {
+		return "", err
+	}
+	if _, err := strconv.Atoi(stringValue); err != nil {
+		return "", err
+	}
+	return stringValue, nil
 }
 
 type SSLCertificate struct {
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -22,10 +22,13 @@ import (
 	// S3 Ref: https://github.com/longhorn/backupstore/blob/3912081eb7c5708f0027ebbb0da4934537eb9d72/s3/s3.go#L33-L37
 	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/harvester/go-common/ds"
+	networkv1 "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io/v1beta1"
+	"github.com/harvester/harvester-network-controller/pkg/utils"
 	"github.com/longhorn/backupstore"
 	_ "github.com/longhorn/backupstore/nfs" //nolint
 	_ "github.com/longhorn/backupstore/s3"  //nolint
 	lhv1beta2 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhtypes "github.com/longhorn/longhorn-manager/types"
 	"github.com/rancher/lasso/pkg/log"
 	mgmtv3 "github.com/rancher/rancher/pkg/generated/controllers/management.cattle.io/v3"
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
@@ -42,9 +45,6 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/clientcmd"
 	kubevirtv1 "kubevirt.io/api/core/v1"
-
-	networkv1 "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io/v1beta1"
-	"github.com/harvester/harvester-network-controller/pkg/utils"
 
 	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/containerd"
@@ -110,6 +110,7 @@ var validateSettingFuncs = map[string]validateSettingFunc{
 	settings.KubeconfigDefaultTokenTTLMinutesSettingName:       validateKubeConfigTTLSetting,
 	settings.AdditionalGuestMemoryOverheadRatioName:            validateAdditionalGuestMemoryOverheadRatio,
 	settings.MaxHotplugRatioSettingName:                        validateMaxHotplugRatio,
+	settings.LHIMResourcesSettingName:                          validateLHIMResources,
 }
 
 type validateSettingUpdateFunc func(oldSetting *v1beta1.Setting, newSetting *v1beta1.Setting) error
@@ -203,6 +204,9 @@ func NewValidator(
 	validateSettingUpdateFuncs[settings.KubeVirtMigrationSettingName] = validator.validateUpdateKubeVirtMigration
 
 	validateSettingDeleteFuncs[settings.ClusterPodSecurityStandardSettingName] = validator.validateDeleteClusterPodSecurityStandard
+
+	validateSettingUpdateFuncs[settings.LHIMResourcesSettingName] = validator.validateUpdateLHIMResources
+	validateSettingDeleteFuncs[settings.LHIMResourcesSettingName] = validator.validateDeleteLHIMResources
 
 	return validator
 }
@@ -2014,6 +2018,40 @@ func validateUpdateMaxHotplugRatio(_ *v1beta1.Setting, newSetting *v1beta1.Setti
 	return validateMaxHotplugRatio(newSetting)
 }
 
+func validateLHIMResources(setting *v1beta1.Setting) error {
+	// If value is empty, this setting is using default and should be ignored.
+	if setting.Value == "" {
+		return nil
+	}
+
+	return validateLHIMResourcesHelper(settings.KeywordValue, setting.Value)
+}
+
+func validateLHIMResourcesHelper(field, value string) error {
+	if value == "" {
+		return nil
+	}
+
+	resources, err := settings.DecodeLHIMResources(value)
+	if err != nil {
+		return fmt.Errorf("failed to parse %v: %w", field, err)
+	}
+	if resources.IsEmpty() {
+		return nil
+	}
+
+	v1CPU := resources.CPU.V1
+	v2CPU := resources.CPU.V2
+	if err := lhtypes.ValidateCPUReservationValues(lhtypes.SettingNameGuaranteedInstanceManagerCPU, v1CPU); err != nil {
+		return fmt.Errorf("failed to validate %v cpu.v1: %w", field, err)
+	}
+	if err := lhtypes.ValidateCPUReservationValues(lhtypes.SettingNameGuaranteedInstanceManagerCPU, v2CPU); err != nil {
+		return fmt.Errorf("failed to validate %v cpu.v2: %w", field, err)
+	}
+
+	return nil
+}
+
 func validateMaxHotplugRatioHelper(field, value string) error {
 	if value == "" {
 		return nil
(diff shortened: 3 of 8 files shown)


pkg/settings/settings_helper.go
```diff
@@ -107,6 +107,78 @@ type Overcommit struct {
...
+	}
+
+	var stringValue string
+	if err := json.Unmarshal(raw, &stringValue); err == nil {
+		intValue, convErr := strconv.Atoi(stringValue)
+		if convErr != nil {
+			return 0, convErr
+		}
+		return intValue, nil
+	}
```
brandboat: I'm a bit curious why do we accept both string and int formats in this new setting?

WebberHuang1118: Good catch, remove the redundant logic—thanks.

pkg/controller/master/setting/instance_manager_resources.go
```diff
@@ -0,0 +1,89 @@
...
+		"v1": v1CPU,
+		"v2": v2CPU,
+	})
+	if err != nil {
+		return err
+	}
+	lhValue := string(lhValueBytes)
+
+	lhSettingCpy := lhSetting.DeepCopy()
+	if lhSettingCpy.Value == lhValue {
```
w13915984028: put below to L60 ?

```
	if lhSetting.Value == lhValue {
		return nil
	}
```

pkg/webhook/resources/setting/validator_test.go
```diff
@@ -1350,3 +1352,116 @@ func Test_validateMaxHotplugRatio(t *testing.T) {
...
+				Default:    `{"cpu":{"v1":12,"v2":12}}`,
+				Value:      `{"cpu":{"v1":20,"v2":30}}`,
+			},
+			errMsg: "",
+		},
+		{
+			name: "fail to create instance-manager-resources with invalid value",
+			args: &v1beta1.Setting{
+				ObjectMeta: metav1.ObjectMeta{Name: settings.LHIMResourcesSettingName},
+				Value:      `{"cpu":{"v1":41,"v2":12}}`,
```
w13915984028: please also cover values like below 

`{"cpu":{"v1":41.8,"v2":12}}`
`{"cpu":{"v1":"abc","v2":12}}`
`{"cpu":{"v1":"","v2":12}}`

pkg/controller/master/setting/instance_manager_resources.go
```diff
@@ -0,0 +1,89 @@
...
+	v2CPU := strconv.Itoa(resources.CPU.V2)
+
+	if err := lhtypes.ValidateCPUReservationValues(lhtypes.SettingNameGuaranteedInstanceManagerCPU, v1CPU); err != nil {
+		return fmt.Errorf("invalid CPU value for cpu.v1 %q in %s: %w", v1CPU, setting.Name, err)
+	}
+	if err := lhtypes.ValidateCPUReservationValues(lhtypes.SettingNameGuaranteedInstanceManagerCPU, v2CPU); err != nil {
+		return fmt.Errorf("invalid CPU value for cpu.v2 %q in %s: %w", v2CPU, setting.Name, err)
+	}
+
+	if err := h.checkLonghornVolumeDetached(); err != nil {
```
w13915984028, line 43: Did you consider to:

 move L34 after L60, when the setting is different with LH and a real sync is needed, then check volume.

given the case: user deletes this setting (never changed, same with LH), and later Harvester/user recreates this setting

w13915984028 (commented):
A few questions, thanks.

pkg/controller/master/setting/instance_manager_resources.go
WebberHuang1118: There’s a default value defined in `pkg/settings/settings.go`:
```
LHIMResources                          = NewSetting(LHIMResourcesSettingName, `{"cpu":{"v1":"12","v2":"12"}}`)
```

So I don’t think this check is strictly necessary, thanks.

pkg/webhook/resources/setting/validator_test.go
WebberHuang1118: Updated per your suggestion—thanks.

pkg/controller/master/setting/instance_manager_resources.go
WebberHuang1118: I’d like to keep the current design. The Harvester setting shouldn’t be deleted by the user. Even if the setting is removed—or if the sync is one-way and the Harvester's setting changes—the reconcile will be triggered and will eventually make the Harvester and Longhorn settings consistent.

Also, checking volume detachment on every reconcile is safer, and the volume status is retrieved from the cache, so it should be a lightweight operation.

pkg/settings/settings_helper.go
```diff
@@ -107,6 +107,73 @@ type Overcommit struct {
...
+
+	return 0, fmt.Errorf("must be an integer")
```
brandboat: just a habbit of taste

```suggestion
	if err := json.Unmarshal(raw, &stringValue); err != nil {
		return 0, err
	}

	return strconv.Atoi(stringValue)
```

pkg/settings/settings_helper.go
```diff
@@ -107,6 +107,73 @@ type Overcommit struct {
 	Storage int `json:"storage"`
 }
 
+type LHIMCPUConfig struct {
+	V1 int `json:"v1"`
+	V2 int `json:"v2"`
```
brandboat: We might not need to convert the CPU config from a string to an int. Since we aren't using the v1 and v2 values anywhere else yet—they're essentially just placeholders for Longhorn setting—we can keep them as strings. This avoids unnecessary conversions on the Harvester side, another benefit is that we can simply use json marshal/unmarshal which makes life a lot easier.

Just my 2 cents :)

```
	V1 string `json:"v1"`
	V2 string `json:"v2"`
```

pkg/webhook/resources/setting/validator.go
```diff
@@ -191,6 +192,7 @@ func NewValidator(
 	validateSettingUpdateFuncs[settings.UpgradeConfigSettingName] = validator.validateUpdateUpgradeConfig
 
 	validateSettingUpdateFuncs[settings.LonghornV2DataEngineSettingName] = validator.validateUpdateLonghornV2DataEngine
+	validateSettingUpdateFuncs[settings.LHIMResourcesSettingName] = validator.validateUpdateLHIMResources
```
brandboat: what about adding another validator to prevent user from deleting the setting?
like this one, wdyt?

```
func (v *settingValidator) validateDeleteVMMigrationNetwork(_ *v1beta1.Setting) error {
	return werror.NewMethodNotAllowed(fmt.Sprintf("Disallow delete setting name %s", settings.VMMigrationNetworkSettingName))
}
```

pkg/controller/master/setting/instance_manager_resources.go
brandboat: Perhaps Jian's goal is to avoid an unnecessary deep copy if the value is already the same

brandboat (dismissed):
Overall it looks good. There are a few minor tweaks, but they're totally optional.

pkg/settings/settings_helper.go
WebberHuang1118: This is better, thanks for the suggestion.

pkg/controller/master/setting/instance_manager_resources.go
WebberHuang1118: Oh, that makes sense—thanks @brandboat and @w13915984028.

WebberHuang1118: Updated per your suggestion—thanks.

pkg/settings/settings_helper.go
WebberHuang1118: Updated per your suggestion—thanks.

pkg/webhook/resources/setting/validator.go
WebberHuang1118: Updated per your suggestion—thanks.

pkg/controller/master/setting/instance_manager_resources.go
w13915984028: current approach works on new instllation, LH is passively synced by Harvester with default values when system in installed and  at that time there is on workload

what's the consideration on upgrade path? e.g. v171, LH has been set with some custom values, the controller might stuck on reporting failure as it assumes some workloads are online but we wanto to change LH settings


we might need to generate the new setting on upgrade path, which fetches the current LH related values into `value` part, and the controller skips the sync when setting is same with LH

thanks.

pkg/settings/settings_helper.go
```diff
@@ -107,6 +107,71 @@ type Overcommit struct {
...
+type LHIMCPUConfig struct {
+	V1 string `json:"v1"`
+	V2 string `json:"v2"`
+}
+
+type LHIMResourcesConfig struct {
+	CPU LHIMCPUConfig `json:"cpu"`
+}
+
+func DecodeLHIMResources(value string) (*LHIMResourcesConfig, error) {
```
brandboat, line 124: this is not needed anymore, right? Could we use json.unmarshal to replace this method?
```
var config LHIMResourcesConfig
json.Unmarshal(value, &config)
```

brandboat (dismissed):
one nit

pkg/settings/settings_helper.go
WebberHuang1118: Still need `DecodeLHIMResources()` to perform some fundamental validation—for example, using `decodeLHIMCPUValue()` to verify the input is numeric.

pkg/controller/master/setting/instance_manager_resources.go
WebberHuang1118: This seems to be a different topic from the original one ([discussion](https://github.com/harvester/harvester/pull/10125#discussion_r2886605946)).

For the “what if the user deletes this Harvester setting?” discussion, I’ve added a webhook to reject deletion, per @brandboat’s suggestion ([discussion](https://github.com/harvester/harvester/pull/10125#discussion_r2887466529)).

For the upgrade consideration ([discussion](https://github.com/harvester/harvester/pull/10125#discussion_r2888283738)),I’d like to keep the current design and avoid making the logic overly complex. My thinking is:
* Before v1.8.0, Harvester doesn’t expose this Longhorn setting. If a user configures it directly in Longhorn, they should be aware it’s outside Harvester’s scope.
  * We can add documentation for this scenario.
* After upgrading to v1.8.0, once this Harvester setting is touched, the controller will sync the setting between Harvester and Longhorn.
* For existing Harvester settings that bridge to Longhorn (e.g., storage network, tiering, etc.), Harvester doesn’t introduce this kind of behavior.

brandboat (dismissed):

pkg/controller/master/setting/instance_manager_resources.go
w13915984028: take a look of this feature https://docs.harvesterhci.io/v1.8/install/harvester-configuration#installharvesterlonghorndefault_settingsguaranteedinstancemanagercpu, it starts from Harvester v1.4.0, 

aims to help user to customize the setting when cluster is installed via a Harvester configuration file,

some users tend to run infrastructure from code, and sets things on day 0, avoid on day 1 via UI or kubectl

there seems to be two challenges:
(1) the new installation, when a `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` is set via the configure file, will it be amended by the new setting?
(2) the upgrade path

WDYT? thanks.

w13915984028: BTW:

LH's settings are not all exposed to Harvester UI. Sometimes, user can only adjust things via LH UI, even Harvester document mentions how to access embedded LH UI https://docs.harvesterhci.io/v1.8/troubleshooting/harvester/#access-embedded-rancher-and-longhorn-dashboards.


Some existing clusters are continuously upgraded from far old 1.1 to 1.5, 1.6, 1.7

when we add new features, we can't assume that LH side are always just the default values, or we can simply single-way overwrite exising values via new Harvester settings.

WebberHuang1118: If Harvester doesn’t expose this setting, I don’t think so. 

As I mentioned, configuring it directly in Longhorn is outside Harvester’s scope. 

Also, for many existing Harvester settings that integrate with Longhorn (e.g., storage network, tiering, etc.), the Harvester setting doesn’t sync back from Longhorn, as far as I understand.

Also, once the new setting is touched, the controller will sync it between Harvester and Longhorn.

w13915984028: ok, we back to the above topic, the Harvester exposed configuration `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU`, how to handle it?

WebberHuang1118: I can add logic to avoid having the Harvester setting’s default value overwrite `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` if it’s already configured.

If the new setting is explicitly configured, the controller will sync the values between Harvester and Longhorn.

WebberHuang1118: Updated with the scenario above: treat `{"cpu":{}}` (default value) as an explicit no-op value. In this case, we skip Longhorn setting updates.

brandboat (dismissed):

pkg/controller/master/setting/instance_manager_resources.go
w13915984028: seems a trouble on this series of oepration, please double check:

(1) Via UI set an explicit value like "15"
(2) Via UI `clear setting/use the default value`, the setting value becomes empty
(3) The new setting, seems not synced to LH, and LH does not change to the default value `12` or empty.

w13915984028: Some thoughts:

For the upgrade path:

 if `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` is set, then `helm get values` could fetch it and convert to the new setting's `value` part, create the setting on v17*->v180 upgrade path
 
 For new installation, if `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` is set as well, TBD.
 
 For the controller sync with LH, the `equal case` needs to be identified to avoid warning `workloads need to be stopped` to cover above two.

WebberHuang1118: > seems a trouble on this series of oepration, please double check:
> 
> (1) Via UI set an explicit value like "15"
> (2) Via UI `clear setting/use the default value`, the setting value becomes empty
> (3) The new setting, seems not synced to LH, and LH does not change to the default value `12` or empty.

As described in the source code comments:
```
	// Treat {"cpu":{}} as an explicit no-op value. In this case we skip
	// Longhorn setting updates.
	if resources.IsEmpty() {
		return nil
	}
```

`{"cpu":{}}` means “skip Longhorn setting updates and keep the current value as-is.” The rationale is:
* This is the default value for the new Harvester setting. If `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` is already configured, this avoids overwriting the existing Longhorn setting.
* This new setting aims to provide a centralized place for Harvester users to manage separate Longhorn IM resource settings. Currently, Longhorn supports configuring IM CPU allocation, but it may introduce more IM controls in the future. For example, if Longhorn later adds a separate setting to configure IM memory usage, we can extend this Harvester setting to include memory. In that case, if a user only wants to update memory, they could set something like `{"cpu":{}, "memory":"10G"}` and wouldn’t need to specify CPU if they don’t intend to change it.

As mentioned in the [discussion](https://github.com/harvester/harvester/pull/10125#discussion_r2894400602) , to avoid over-complicating the implementation, the new setting is not designed to be retroactive.

The main purpose is to provide an interface for Harvester users to reduce Longhorn IM resource usage—especially for day-2 scenarios where users primarily rely on third-party storage. So I think the must-have points are:
* If the new setting is configured with a valid value, sync it to Longhorn.
* If `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` is already configured, do not overwrite it.

TBH, since the related Longhorn setting can only be changed when all volumes are detached, in our experience not many users adjust the IM resource allocation setting.

Given the current design already provides flexibility, if more complex logic is needed in the controller or webhook later, we can iterate. 

For now, I’d like to keep the implementation concise and focused on the mandatory parts, and then see if we get any feedback. 

cc @Vicente-Cheng

w13915984028: As a commercial system, a new setting at least should:

(1) keep backword compatiable with legacy versions, especially given `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` has been existing for several versions,

(2) the back and force adjusting of this setting is idempotency, some customers tend to tune the system frequently before finally deploy workloads, or test the system robustness on the PoC stage

when the feature can only work on happy path, and can't handle above mentioned cluster upgrade, re-set scenario smoothly.  If others say "we can go this way".  OK, I keep my views aside.

WebberHuang1118: > (1) keep backword compatiable with legacy versions, especially given install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU has been existing for several versions

`install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` is a configuration from Longhorn setting. This new Harvester setting isn’t meant to be tied only to that Longhorn setting—IMO, it’s a different concept. It’s designed as a centralized place to control Instance Manager (IM) resource usage.

> (2) the back and force adjusting of this setting is idempotency, some customers tend to tune the system frequently before finally deploy workloads, or test the system robustness on the PoC stage

I don’t see any obstacles that would prevent users from tuning the system under the current design.

> when the feature can only work on happy path, and can't handle above mentioned cluster upgrade, re-set scenario smoothly. If others say "we can go this way". OK, I keep my views aside

I don’t think this only works on the happy path—unless you have a specific real-world case in mind? I thought we all agreed that retroactive behavior isn’t needed ([https://github.com/harvester/harvester/pull/10125#discussion_r2894582060](https://github.com/harvester/harvester/pull/10125#discussion_r2894582060)).

I think we’ve already gone over a similar discussion several times. For the upgrade path, the setting won’t overwrite `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU`. Once this setting is touched, the controller will sync the value between Harvester and Longhorn.

WebberHuang1118: Hi  @Vicente-Cheng 

I’d like to hear your thoughts on the above topic, I'm open to discussion, thanks,

WebberHuang1118: Actually, `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` is no longer functional, since Longhorn changed the setting format for controlling IM CPU allocation. 

In Harvester Installer, `GuaranteedInstanceManagerCPU` is declared as an `int` ([config.go L136](https://github.com/harvester/harvester-installer/blob/b17d47c250660b0ba66b296081a7f97e97077ef9/pkg/config/config.go#L136)), but the corresponding Longhorn setting is now represented as two key-value pairs ([Guaranteed Instance Manager CPU](https://longhorn.io/docs/1.10.1/references/settings/#guaranteed-instance-manager-cpu)). 

This setting has gone through multiple changes over time:
* Initially, it only controlled v1 CPU allocation.
* Then it was split into separate settings for v1 and v2 CPU allocation.
* Later, v1 and v2 CPU allocation were combined into a single setting again.

It doesn’t seem practical to track all historical Longhorn setting changes, and we probably shouldn’t get hung up on an invalid installer setting. Since this new Harvester setting is introduced in `v1.8.0`, the mechanism is:
* Whether it’s a fresh install or an upgrade from v1.7.x, the default behavior is to keep the Longhorn setting as-is.
* If the Harvester setting is explicitly configured, the Harvester controller will sync the IM CPU allocation to Longhorn.

pkg/webhook/resources/setting/validator.go
```diff
@@ -1994,6 +2032,51 @@ func validateMaxHotplugRatioHelper(field, value string) error {
...
+	volumes, err := v.lhVolumeCache.List(util.LonghornSystemNamespaceName, labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	for _, volume := range volumes {
+		if volume.Status.State != lhv1beta2.VolumeStateDetached {
+			return werror.NewInvalidError("please detach all Longhorn volumes before configuring instance-manager-resources", settings.KeywordValue)
+		}
+	}
```
Vicente-Cheng, line 2116: I am curious about this part of the above controller mechanism.
Even if we let the user configure a new CPU limitation, the controller will wait for all volumes to be detached.
Why do we need to block here?

pkg/controller/master/setting/instance_manager_resources.go
Vicente-Cheng: I think if
>install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU is no longer functional, since Longhorn changed the setting format for controlling IM CPU allocation.

The things gonna be simple, the current design, with an empty default value, would not affect the upgrade user.

The upgraded user will keep their current value.

WDYT?

Vicente-Cheng (dismissed):
just a nit, thanks

pkg/controller/master/setting/instance_manager_resources.go
w13915984028: Per latest code,
```
	if setting.Value == "" {
		return nil
	}
	
	
	if resources.IsEmpty() {
		return nil
	}


func (c *LHIMResourcesConfig) IsEmpty() bool {
	return c == nil || (c.CPU.V1 == "" && c.CPU.V2 == "")
}	
```

It has two `no-op` cases:
 `value == ""` or `c.CPU.V1 == "" && c.CPU.V2 == ""`

(1) upgrade case, the empty intial value ensures things are kept as-it. PASS

(2) User set an explicit value like `15`, it is synced to LH. PASS

(3) User clears the setting via set `value = ""`. Now, controller no-op, LH is left with `15`. ???

(4) User set another value `18`, it is synced to LH. PASS

(5)  User clears the setting via set `value = ""`. Now, controller no-op, LH is left with `18`. ???

For case (3) (5):
(a) The result is not predictable when `reset to default`, this differs from most of the Harvester settings.

(b) User can only login into LH UI or via kubectl to know the LH current real setting values.

(c) If we want to say (3) (5) are invalid cases, then I have no words.

The root cause is the ambiguity of `value == ""`: it represents the intial/none-set value, and also the `reset-value`.


If `install.harvester.longhorn.default_settings.guaranteedInstanceManagerCPU` has been deprecated by LH, use an upgrade script to convert v170 existing values to the new settings' `value` part, and set a default value for this setting, will make things clear.

I did not get the point why we are so reluctant to smoothly convert existing values on upgrade path.

pkg/webhook/resources/setting/validator.go
w13915984028: Before workloads are detached, if we sync to LH directly, LH pods will be replaced, it might cause data loss, hence controller waits until all workloads are stopped.

From the time of webhook check pass (no volume attached) to the point controller works, there is a small time window, inbetween, new workloads might be loaded like new VM is booted or new guest cluster PVC is attached. The controller does a workload detached recheck, makes sense.

On the other hand, I still prefer to have a `equalize check` before volume check. If the new setting values are fully same with LH, warn workload to be detached can be skipped.

It also helps the upgrade path, when upgrade converts v170 LH current values to a new `setting.values` for v180, v180 controller just happily returns as no real sync is needed.

WebberHuang1118: > Why do we need to block here?

Because applying this setting will restart the Instance Manager (IM) pods. This is also described in the Longhorn docs under **Danger Zone**: [Danger Zone settings](https://longhorn.io/docs/1.11.0/references/settings/#danger-zone).

WebberHuang1118:
Discussed with @Vicente-Cheng, and we reached a conclusion on the settings behavior:
* Since the new Harvester setting is introduced in `v1.8.0`, whether upgrading from v1.7.x or doing a fresh install, the Longhorn **Guaranteed Instance Manager CPU** setting will be kept as-is.
* Once the new Harvester setting is configured with an explicit value, the Harvester controller will sync it with Longhorn’s **Guaranteed Instance Manager CPU** setting.
* If the new Harvester setting is reset back to default, the Harvester controller will sync Longhorn’s **Guaranteed Instance Manager CPU** setting back to the Longhorn default value `{"v1":"12","v2":"12"}`.

pkg/webhook/resources/setting/validator.go
Vicente-Cheng: thanks for the explaination, let's keep it

Vicente-Cheng (approved):
lgtm, thanks!

pkg/controller/master/setting/instance_manager_resources.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	v1CPU := resources.CPU.V1
+	v2CPU := resources.CPU.V2
+
+	if err := lhtypes.ValidateCPUReservationValues(lhtypes.SettingNameGuaranteedInstanceManagerCPU, v1CPU); err != nil {
+		return fmt.Errorf("invalid CPU value for cpu.v1 %q in %s: %w", v1CPU, setting.Name, err)
+	}
+	if err := lhtypes.ValidateCPUReservationValues(lhtypes.SettingNameGuaranteedInstanceManagerCPU, v2CPU); err != nil {
+		return fmt.Errorf("invalid CPU value for cpu.v2 %q in %s: %w", v2CPU, setting.Name, err)
+	}
```
brandboat, line 41: nit: We could move ValidateCPUReservationValues into DecodeLHIMResources to simplify the flow. Currently we’re decoding and validating LHIMResourcesConfig separately. No big deal if you want to keep it this way.

brandboat (approved):
LGTM, thanks!

pkg/controller/master/setting/instance_manager_resources.go
WebberHuang1118: @brandboat Thanks for the feedback.

I’d like to keep the setting value decoding and validation separate—they target different concerns from my perspective.

Vicente-Cheng:
> Discussed with @Vicente-Cheng, and we reached a conclusion on the settings behavior:
> 
> * Since the new Harvester setting is introduced in `v1.8.0`, whether upgrading from v1.7.x or doing a fresh install, the Longhorn **Guaranteed Instance Manager CPU** setting will be kept as-is.
> * Once the new Harvester setting is configured with an explicit value, the Harvester controller will sync it with Longhorn’s **Guaranteed Instance Manager CPU** setting.
> * If the new Harvester setting is reset back to default, the Harvester controller will sync Longhorn’s **Guaranteed Instance Manager CPU** setting back to the Longhorn default value `{"v1":"12","v2":"12"}`.

Hi @w13915984028, 
Currently, we have applied the above mechanism to the previous discussion. Please create a new ticket if there are any improvements we should make. Thanks!

WebberHuang1118:
@mergify backport v1.8

PR #10132, opened 2026-03-04
task: add option to registration url setting

ihcsim:
#### Problem:

Provide option to configure the client of the `cluster-registration-url` setting handler.

#### Related Issue(s):

https://github.com/harvester/suse-virtualization-mgmt/issues/5

#### Test plan:

see https://github.com/harvester/suse-virtualization-mgmt/issues/5

### DIFF
--- a/pkg/settings/settings.go
+++ b/pkg/settings/settings.go
@@ -19,7 +19,7 @@ var (
 
 	AdditionalCA                           = NewSetting(AdditionalCASettingName, "")
 	APIUIVersion                           = NewSetting(APIUIVersionSettingName, "1.1.9") // Please update the HARVESTER_API_UI_VERSION in package/Dockerfile when updating the version here.
-	ClusterRegistrationURL                 = NewSetting(ClusterRegistrationURLSettingName, "")
+	ClusterRegistrationURL                 = NewSetting(ClusterRegistrationURLSettingName, `{"url":"","insecureSkipTLSVerify":false}`)
 	ServerVersion                          = NewSetting(ServerVersionSettingName, "dev")
 	UIIndex                                = NewSetting(UIIndexSettingName, DefaultDashboardUIURL)
 	UIPath                                 = NewSetting(UIPathSettingName, "/usr/share/harvester/harvester")
--- a/pkg/settings/settings_helper.go
+++ b/pkg/settings/settings_helper.go
@@ -303,3 +303,23 @@ func GetPodSecuritySetting(setting *harvesterv1.Setting) (*PodSecuritySetting, e
 	}
 	return pssSetting, nil
 }
+
+type ClusterRegistrationURLSetting struct {
+	URL                   string `json:"url"`
+	InsecureSkipTLSVerify bool   `json:"insecureSkipTLSVerify"`
+}
+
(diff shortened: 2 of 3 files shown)


pkg/settings/settings.go
```diff
@@ -19,7 +19,7 @@ var (
 
 	AdditionalCA                           = NewSetting(AdditionalCASettingName, "")
 	APIUIVersion                           = NewSetting(APIUIVersionSettingName, "1.1.9") // Please update the HARVESTER_API_UI_VERSION in package/Dockerfile when updating the version here.
-	ClusterRegistrationURL                 = NewSetting(ClusterRegistrationURLSettingName, "")
```
w13915984028, line 22: a quick question: do we need the help of upgrade scripts to convert v17* existing non-empty settings `ClusterRegistrationURL` to new format? thanks.

ihcsim: It's not needed because the ui and the backend handler will convert the old format into the json format, when the value is next used. Unlike other settings, this value is not regularly used. The cattle-agent has its own secret to store the upstream URL.

ibrokethecloud (approved):

pkg/settings/settings.go
w13915984028: Based on points from Copilot https://github.com/harvester/harvester/pull/10132#discussion_r2890122366, I would suggest we convert it on upgrade path, and controller side just treat any unmarshall error as error, don't fallback

BTW, do we have an UI issue/pr to expose the additional field and compse the new format?  thanks.

pkg/settings/settings_helper.go
```diff
@@ -303,3 +303,23 @@ func GetPodSecuritySetting(setting *harvesterv1.Setting) (*PodSecuritySetting, e
...
+	value := setting.Default
+	if setting.Value != "" {
+		value = setting.Value
+	}
+
+	if err := json.Unmarshal([]byte(value), reg); err != nil {
+		logrus.Warnf("%s, value: %s. treating the registration URL value directly for backward compatibility", err.Error(), value)
+		reg.URL = value
+		reg.InsecureSkipTLSVerify = true
+	}
```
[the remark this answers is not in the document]
w13915984028: The suggestion from Copilot is reasonable

if we do a converting on upgrade path, then the controller can directly return any JSON unmarshall error, it need not to fallback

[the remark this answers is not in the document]
ihcsim: I did some testing on the current behaviour of bad inputs handling. If the input happens to be a malformed URL (JSON or not), the subsequent call to `httpClient.Get()` will fail. We thought of adding a `url.ParseRequestURI()` here but there is no difference in the outcome, when compared to failing at `httpClient.Get()`.

pkg/settings/settings.go
ihcsim: Per offline discussion, in our initial implementation, we did include the upgrade path. However, from our testing, changing the setting during upgrade would cause Harvester to get disconnected from Rancher temporarily while the cattle-agent re-spawns. While it comes back fairly quickly in my environment, we aren't able to guarantee that this will be the case in users' environment.

w13915984028 (approved):
LGTM, thanks.

Vicente-Cheng (approved):
thanks!

ihcsim:
@Mergifyio backport v1.7

PR #10135, opened 2026-03-05
feat: support adding support_bundle_file_name to customize bundle name

Yu-Jack:
#### Problem:

Since the customer wants a global name for support bundle, we can provide a support bundle file name setting. It's not directly related to the cluster name. 

NOTE: Although the customer can rename the file after downloading, the custom still wants a pre-defined name for that.

#### Solution:
Provide a support bundle file name setting.

- If it's empty, format will be `supportbundle_c2eb6186-525f-411e-91ee-e88ed090d66d_2025-09-17T03-39-17Z`.
- If not, format will be `supportbundle_{support-bundle-file-name}_2025-09-17T03-39-17Z`.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9257

#### Test plan:

After configuring the setting, remember to generate support bundle.

case 1: empty support bundle file name
case 2: non-empty support bundle file name
case 3: the format only supports [RFC 1123 Label Names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names), you can try with other invalid strings

### DIFF
--- a/pkg/controller/master/supportbundle/manager.go
+++ b/pkg/controller/master/supportbundle/manager.go
@@ -102,6 +102,10 @@ func (m *Manager) Create(sb *harvesterv1.SupportBundle, image string, pullPolicy
 									Value: sb.Name,
 								},
 								{
+									Name:  "SUPPORT_BUNDLE_FILE_NAME",
+									Value: settings.SupportBundleFileName.Get(),
+								},
+								{
 									Name:  "SUPPORT_BUNDLE_DESCRIPTION",
 									Value: sb.Spec.Description,
 								},
--- a/pkg/webhook/resources/setting/validator.go
+++ b/pkg/webhook/resources/setting/validator.go
@@ -38,6 +38,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/clientcmd"
 	kubevirtv1 "kubevirt.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


pkg/webhook/resources/setting/validator.go
```diff
@@ -834,6 +837,36 @@ func validateUpdateSupportBundle(_ *v1beta1.Setting, newSetting *v1beta1.Setting
...
+// validateSupportBundleFileNameHelper validates that the file name follows RFC 1123 Label Names
+// using Kubernetes built-in validation.
+func validateSupportBundleFileNameHelper(value string) error {
+	if value == "" {
+		return nil
+	}
+
+	errs := validation.IsDNS1123Label(value)
+	if len(errs) > 0 {
+		return fmt.Errorf("invalid file name: %s", strings.Join(errs, "; "))
```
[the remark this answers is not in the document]
w13915984028: this suggestion makes sense, mention `DNS-1123 label` is helpfu

pkg/controller/master/supportbundle/manager.go
```diff
@@ -101,6 +101,10 @@ func (m *Manager) Create(sb *harvesterv1.SupportBundle, image string, pullPolicy
 									Name:  "SUPPORT_BUNDLE_NAME",
 									Value: sb.Name,
 								},
+								{
+									Name:  "SUPPORT_BUNDLE_FILE_NAME",
+									Value: settings.SupportBundleFileName.Get(),
```
w13915984028, line 106: I see below code has a trap, when target settings is not existing (e.g. user deletes the setting), it does not check before call `s.Default`

it is not fully related to this PR, but an exisint issue on current code, and could be triggered here:


we might need to enhance them to ensure `settings.SupportBundleFileName.Get()` won't panic

```
func (s Setting) Get() string {
	if provider == nil {
		s := settings[s.Name]
		return s.Default
	}
	return provider.Get(s.Name)
}
```

https://github.com/harvester/harvester/blob/2271ea36714e600e9fddd10b337690c0a802b433/pkg/controller/global/settings/register.go#L37-L38

w13915984028 (commented):
One question, thanks.

pkg/controller/master/supportbundle/manager.go
Yu-Jack: I'm not sure what you meant. Have you tried it? I tried, but It won't throw panic after deleting that setting.

```
time="2026-03-06T02:18:54Z" level=warning msg="Failed to fetch setting support-bundle-file- │
│ name from cache, attempting direct API call: settings.harvesterhci.io \"support-bundle-file │
│ -name\" not found"                                                                          │
│ time="2026-03-06T02:18:54Z" level=error msg="Failed to fetch setting support-bundle-file-na │
│ me from API, falling back to default: settings.harvesterhci.io \"support-bundle-file-name\" │
│  not found"
```

I even deleted support-bundle-node-collection-time to test this case, it ended up using the default value from the code base. So, I think it should be fine.

w13915984028: I did not get a chance to test. It is good to hear the code works robustly.

w13915984028 (approved):
LGTM, thanks.

pohanhuang (approved):
LGTM

PR #10145, opened 2026-03-06
chore: restrict action with minimal permission

Yu-Jack:


pohanhuang (changes requested):
The pull-requests: write permission is currently set at the top level, which means all jobs will inherit this write permission even if they don't need it. 
We should move it to the job level to follow the least-privilege principle.

Yu-Jack:
> The pull-requests: write permission is currently set at the top level, which means all jobs will inherit this write permission even if they don't need it. We should move it to the job level to follow the least-privilege principle.

But, we only have one job. I think we can move it to job level when we have the second job in this action? WDYT?

pohanhuang:
@Yu-Jack that would be great, but if the current setting still falls in global level, not job level

.github/workflows/issue-management-link-backport-pr.yaml
```diff
@@ -10,6 +10,9 @@ on:
 env:
   REPO_NAME: harvester/harvester
 
+permissions:
+  pull-requests: write
+
 jobs:
   check-backport:
     runs-on: ubuntu-latest
```
pohanhuang, line 15:
```suggestion
jobs:
  check-backport:
    runs-on: ubuntu-latest
    permissions:
      pull-requests: write
```

pohanhuang (approved):
LGTM, thanks for the help

PR #10202, opened 2026-03-12
[v190] fix: reject cpumanager changes from user

martindekov:
Rejecting cpu manager changes coming from users as harvester controller is already using the value to configure it and we silently drop it.

Now we determine if request is coming from controller or user and reject it in case of user request as opposed to silen drop.

#### Problem:
We silently [clear node selector terms](https://github.com/harvester/harvester/blob/d8985bcddfd0f3ba016470197a4170f5ebe8b740/pkg/webhook/resources/virtualmachine/mutator.go#L439-L453) related with `cpumanager` and starting with prefix `network.harvesterhci.io`.

#### Solution:
Reject user requests on a webhook level when they try to configure the `cpumanager`

> Note: original issue did not discuss the prefix `network.harvesterhci.io` so this is the reason the change only rejects on cpumanger config. We can extend the scope and reject the network as well

#### Related Issue(s):
https://github.com/harvester/harvester/issues/7583

#### Test plan:
Unit tests added for Create/Update in the mutator

E2E Test plan covers both user and command line interfaces:

* Basic Scenarios:
  * Create VM with cpumanager affinity - rejection  (`vm-1`)
  * Create VM with other affinity key - accepted (`vm-1`)
  * Create VM with no affinity - accepted (`vm-2`)
  * Add cpumanager term to `vm-2` - rejected
  * Add other affinity to `vm-2` - accepted

* CPU Pinning tests (`vm-3`):
  * Create VM with cpu pinning - accepted (controller changes it) (`vm-3`)
  * Update the cpumanager field on rejected (users can't change it)
  * Change requests.memory of vm-3 - accepted (other updates are allowed) (user interface + command line interface edit + view)
  * Remove cpumanager annotation - rejected
  * Turn off cpu pinning on vm-3 - accepted (controller manages it) (edit through command line interface view result in user interface)

* CPU Pinning odd values (`vm-1`):
  * Update with empty values slice
  * Update with null values slice
  * Update with multiple values slice 
  * Update with multiple cpumanager duplicate key slice

Basic Scenarios Results:
<details>
<summary> :green_circle: Built and ran environment with the commit</summary>
image
</details>

<details>
<summary> :green_circle: Create VM with cpumanager affinity - rejection</summary>
image
</details>

<details>
<summary> :green_circle: Create VM with other affinity key - accepted</summary>
image
image
</details>

<details>
<summary> :green_circle: Create VM with no affinity - accepted</summary>
image
</details>

<details>
<summary> :green_circle: Add cpumanager term to vm-2 - rejected</summary>
image
</details>

<details>
<summary> :green_circle: Add other affinity to `vm-2` - accepted</summary>
image
</details>

CPU Pinning tests Results:
<details>
<summary> :green_circle: Create VM with cpu pinning - accepted</summary>
image
image
</details>

<details>
<summary> :green_circle: Update the cpumanager field on - rejected </summary>
image
image
</details>

<details>
<summary> :green_circle: Change requests.memory of vm-3 - accepted (user interface + command line interface)</summary>
in cli result:
image

```
node-1:/home/rancher # kubectl get vm vm-3 -o yaml | grep memory
        memory:
            memory: 2Gi
            memory: 2Gi
node-1:/home/rancher # kubectl edit vm vm-3 
virtualmachine.kubevirt.io/vm-3 edited
node-1:/home/rancher # kubectl get vm vm-3 -o yaml | grep memory
        memory:
            memory: 3Gi
            memory: 3Gi
```

</details>

<details>
<summary> :green_circle:  Remove cpumanager annotation - rejected </summary>
image
image

</details>

<details>
<summary> :green_circle: Turn off cpu pinning on vm-3 - accepted (command line interface edit user interface view)</summary>
image
image
image
</details>

CPU Pinning odd values all rejected 

<details>
<summary> :green_circle: All rejected (`vm-1`) </summary>

```
node-1:/home/rancher # kubectl patch vm vm-1 -n default --type=merge -p='{"spec":{"template":{"spec":{"affinity":{"nodeAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":{"nodeSelectorTerms":[{"matchExpressions":[{"key":"cpumanager","operator":"In","values":[]}]}]}}}}}}}'
The request is invalid: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution: key "cpumanager" is automatically operated by Harvester controller and can't be changed manually
node-1:/home/rancher # kubectl patch vm vm-1 -n default --type=merge -p='{"spec":{"template":{"spec":{"affinity":{"nodeAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":{"nodeSelectorTerms":[{"matchExpressions":[{"key":"cpumanager","operator":"In","values":null}]}]}}}}}}}'
The request is invalid: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution: key "cpumanager" is automatically operated by Harvester controller and can't be changed manually
node-1:/home/rancher # kubectl patch vm vm-1 -n default --type=merge   -p='{"spec":{"template":{"spec":{"affinity":{"nodeAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":{"nodeSelectorTerms":[{"matchExpressions":[{"key":"cpumanager","operator":"In","values":["true","false"]}]}]}}}}}}}'
The request is invalid: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution: key "cpumanager" is automatically operated by Harvester controller and can't be changed manually
node-1:/home/rancher # kubectl patch vm vm-1 -n default --type=merge -p='{"spec":{"template":{"spec":{"affinity":{"nodeAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":{"nodeSelectorTerms":[{"matchExpressions":[{"key":"cpumanager","operator":"In","values":["true"]}]},{"matchExpressions":[{"key":"cpumanager","operator":"In","values":["false"]}]}]}}}}}}}'
The request is invalid: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution: key "cpumanager" is automatically operated by Harvester controller and can't be changed manually
node-1:/home/rancher # 
```
</details>

### DIFF
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -92,10 +92,8 @@ func (m *vmMutator) Create(request *types.Request, newObj runtime.Object) (types
 
 	patchOps = patchDefaultCPU(vm, patchOps)
 
-	if request == nil || !request.IsFromController() {
-		if err := m.checkReservedAffinityKeys(nil, vm); err != nil {
-			return nil, err
-		}
+	if err := checkCpuManagerAffinityTermChanged(request, nil, vm); err != nil {
+		return nil, err
 	}
 
 	patchOps, err = m.patchAffinity(vm, patchOps)
@@ -153,10 +151,8 @@ func (m *vmMutator) Update(request *types.Request, oldObj runtime.Object, newObj
 		patchOps = patchRunStrategy(newVM, patchOps)
 	}
 
-	if request == nil || !request.IsFromController() {
-		if err := m.checkReservedAffinityKeys(oldVM, newVM); err != nil {
-			return nil, err
-		}
+	if err := checkCpuManagerAffinityTermChanged(request, oldVM, newVM); err != nil {
+		return nil, err
 	}
 
 	patchOps, err = m.patchAffinity(newVM, patchOps)
@@ -913,43 +909,43 @@ func isCreatedByForklift(vm *kubevirtv1.VirtualMachine) bool {
 	return migrationExists && planExists && vmIDExists
 }
 
-func (m *vmMutator) checkReservedAffinityKeys(oldVM, newVM *kubevirtv1.VirtualMachine) error {
-	if reflect.DeepEqual(vmAffinityNotNil(oldVM), vmAffinityNotNil(newVM)) {
+func checkCpuManagerAffinityTermChanged(request *types.Request, oldVM, newVM *kubevirtv1.VirtualMachine) error {
+	if request != nil && request.IsFromController() {
 		return nil
 	}
-	if reflect.DeepEqual(cpuManagerAffinityTerms(oldVM), cpuManagerAffinityTerms(newVM)) {
+	getCPUManagerExprs := func(vm *kubevirtv1.VirtualMachine) []v1.NodeSelectorRequirement {
+		if vm == nil || vm.Spec.Template == nil {
+			return nil
+		}
+		affinity := vm.Spec.Template.Spec.Affinity
+		if affinity == nil || affinity.NodeAffinity == nil ||
+			affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+			return nil
+		}
+		var exprs []v1.NodeSelectorRequirement
+		for _, term := range affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
+			for _, expr := range term.MatchExpressions {
+				if expr.Key == kubevirtv1.CPUManager {
+					exprs = append(exprs, expr)
+				}
+			}
+		}
+		if len(exprs) > 1 {
+			slices.SortFunc(exprs, func(a, b v1.NodeSelectorRequirement) int {
+				if cmp := strings.Compare(string(a.Operator), string(b.Operator)); cmp != 0 {
+					return cmp
+				}
+				return strings.Compare(fmt.Sprintf("%v", a.Values), fmt.Sprintf("%v", b.Values))
+			})
+		}
+		return exprs
+	}
+
+	if reflect.DeepEqual(getCPUManagerExprs(oldVM), getCPUManagerExprs(newVM)) {
 		return nil
 	}
 	return werror.NewInvalidError(
 		fmt.Sprintf("key %q is automatically operated by Harvester controller and can't be changed manually", kubevirtv1.CPUManager),
 		"spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution",
 	)
-}
-
-func vmAffinityNotNil(vm *kubevirtv1.VirtualMachine) *v1.Affinity {
-	if vm == nil || vm.Spec.Template == nil {
-		return nil
-	}
-	return vm.Spec.Template.Spec.Affinity
-}
-
-func cpuManagerAffinityTerms(vm *kubevirtv1.VirtualMachine) []v1.NodeSelectorTerm {
-	affinity := vmAffinityNotNil(vm)
-	if affinity == nil || affinity.NodeAffinity == nil ||
-		affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
-		return nil
-	}
-	var terms []v1.NodeSelectorTerm
-	for _, term := range affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
-		for _, expr := range term.MatchExpressions {
-			if expr.Key == kubevirtv1.CPUManager {
-				terms = append(terms, term)
-				break
-			}
-		}
-	}
-	slices.SortFunc(terms, func(a, b v1.NodeSelectorTerm) int {
-		return strings.Compare(fmt.Sprintf("%v", a), fmt.Sprintf("%v", b))
-	})
-	return terms
 }
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -916,11 +917,11 @@ func (m *vmMutator) checkReservedAffinityKeys(oldVM, newVM *kubevirtv1.VirtualMa
 	if reflect.DeepEqual(vmAffinityNotNil(oldVM), vmAffinityNotNil(newVM)) {
 		return nil
 	}
-	if reflect.DeepEqual(reservedAffinityTerms(oldVM), reservedAffinityTerms(newVM)) {
+	if reflect.DeepEqual(cpuManagerAffinityTerms(oldVM), cpuManagerAffinityTerms(newVM)) {
 		return nil
 	}
 	return werror.NewInvalidError(
-		fmt.Sprintf("key %q is reserved and managed by Harvester", kubevirtv1.CPUManager),
+		fmt.Sprintf("key %q is automatically operated by Harvester controller and can't be changed manually", kubevirtv1.CPUManager),
 		"spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution",
 	)
 }
@@ -932,7 +933,7 @@ func vmAffinityNotNil(vm *kubevirtv1.VirtualMachine) *v1.Affinity {
 	return vm.Spec.Template.Spec.Affinity
 }
 
-func reservedAffinityTerms(vm *kubevirtv1.VirtualMachine) []v1.NodeSelectorTerm {
+func cpuManagerAffinityTerms(vm *kubevirtv1.VirtualMachine) []v1.NodeSelectorTerm {
 	affinity := vmAffinityNotNil(vm)
 	if affinity == nil || affinity.NodeAffinity == nil ||
 		affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
@@ -947,5 +948,8 @@ func reservedAffinityTerms(vm *kubevirtv1.VirtualMachine) []v1.NodeSelectorTerm 
 			}
 		}
 	}
+	slices.SortFunc(terms, func(a, b v1.NodeSelectorTerm) int {
+		return strings.Compare(fmt.Sprintf("%v", a), fmt.Sprintf("%v", b))
+	})
 	return terms
 }
--- a/pkg/webhook/resources/virtualmachine/mutator.go
+++ b/pkg/webhook/resources/virtualmachine/mutator.go
@@ -3,6 +3,7 @@ package virtualmachine
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 
@@ -23,6 +24,7 @@ import (
 	"github.com/harvester/harvester/pkg/util"
 	"github.com/harvester/harvester/pkg/util/network"
 	"github.com/harvester/harvester/pkg/util/virtualmachine"
+	werror "github.com/harvester/harvester/pkg/webhook/error"
 	"github.com/harvester/harvester/pkg/webhook/types"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
 )
@@ -77,7 +79,7 @@ func (m *vmMutator) Resource() types.Resource {
 	}
 }
 
-func (m *vmMutator) Create(_ *types.Request, newObj runtime.Object) (types.PatchOps, error) {
+func (m *vmMutator) Create(request *types.Request, newObj runtime.Object) (types.PatchOps, error) {
 	vm := newObj.(*kubevirtv1.VirtualMachine)
 
 	logrus.Debugf("create VM %s/%s", vm.Namespace, vm.Name)
@@ -88,6 +90,12 @@ func (m *vmMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patch
 	}
 
 	patchOps = patchDefaultCPU(vm, patchOps)
+
+	if request == nil || !request.IsFromController() {
+		if err := m.checkReservedAffinityKeys(nil, vm); err != nil {
+			return nil, err
+		}
+	}
 
 	patchOps, err = m.patchAffinity(vm, patchOps)
 	if err != nil {
@@ -113,7 +121,7 @@ func (m *vmMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patch
 	return patchOps, err
 }
 
-func (m *vmMutator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) (types.PatchOps, error) {
+func (m *vmMutator) Update(request *types.Request, oldObj runtime.Object, newObj runtime.Object) (types.PatchOps, error) {
 	newVM := newObj.(*kubevirtv1.VirtualMachine)
 	oldVM := oldObj.(*kubevirtv1.VirtualMachine)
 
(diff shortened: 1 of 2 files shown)


pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -249,12 +255,20 @@ func (v *vmValidator) Create(_ *types.Request, newObj runtime.Object) error {
 	return nil
 }
 
-func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj runtime.Object) error {
+//revive:disable
```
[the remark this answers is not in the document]
martindekov: complexity of the method is subjective, its validation on Update can't be shorter, but is series of simple calls related with validations

pkg/webhook/resources/virtualmachine/validator.go
```diff
@@ -220,12 +220,18 @@ func comparemacs(macList1 map[string]string, macList2 map[string]string) bool {
...
 	vm := newObj.(*kubevirtv1.VirtualMachine)
 	if vm == nil {
 		return nil
 	}
 
+	if request == nil || !request.IsFromController() {
+		if err := v.checkReservedAffinityKeys(nil, vm); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
martindekov: I will just remove the logic from the mutator to stop dropping the cpumanager. Leaving discussion open for others to check the approach

pkg/webhook/resources/virtualmachine/validator_test.go
```diff
@@ -1057,6 +1065,169 @@ func TestVmValidator_Update(t *testing.T) {
 			newSpec:       nil,
 			expectedError: false,
 		},
+		{
+			name:    "user sets reserved cpumanager affinity key – rejected",
+			request: newUserRequest("harvester"),
+			oldVM:   templateVM.DeepCopy(),
+			newVM:   templateVM.DeepCopy(),
+			newSpec: &kubevirtv1.VirtualMachineSpec{
```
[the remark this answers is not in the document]
martindekov: Related with below, will fix

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -837,3 +851,38 @@ func (m *vmMutator) patchManagedTapBinding(vm *kubevirtv1.VirtualMachine, patchO
 
 	return patchOps, nil
 }
+
+func (m *vmMutator) checkReservedAffinityKeys(oldVM, newVM *kubevirtv1.VirtualMachine) error {
+	newTerms := reservedAffinityTerms(newVM)
```
w13915984028: in most cases, `vm.Spec.Template.Spec.Affinity` does not change at all, a deepequal upon `vm.Spec.Template.Spec.Affinity` of old and new can return early

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -837,3 +851,38 @@ func (m *vmMutator) patchManagedTapBinding(vm *kubevirtv1.VirtualMachine, patchO
...
+	if len(newTerms) == 0 {
```
w13915984028: if user removes the affinity, the check is skipped, is this intended?

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -84,6 +86,12 @@ func (m *vmMutator) Create(_ *types.Request, newObj runtime.Object) (types.Patch
 
 	patchOps = patchDefaultCPU(vm, patchOps)
 
+	if request == nil || !request.IsFromController() {
```
w13915984028: the `checkReservedAffinityKeys` is in general belonging to `validator`, not `mutator`

what's the consideration to add it here

w13915984028 (commented):
A few questions, thanks.

pkg/webhook/resources/virtualmachine/mutator.go
martindekov: This was the reason I moved it from the validator to mutator:
https://github.com/harvester/harvester/pull/10202#discussion_r2952119039

the mutator removes the annotation so validator always receives empty cpuamanger requirement even if the user set it so we never actually trigger the webhook

pkg/webhook/resources/virtualmachine/mutator.go
martindekov: It's a miss on my part, removing it still is effectively changing it, I will revisit and won't allow it. Still even if you remove it it will be silently dropped, good catch

pkg/webhook/resources/virtualmachine/mutator.go
martindekov: Good catch in case no changes to affinity will return early without the more complex comparison logic for the reserved terms

martindekov: Early return :heavy_check_mark:

pkg/webhook/resources/virtualmachine/mutator.go
martindekov: if user removes the affinity it's rejected and test changed :heavy_check_mark:

pkg/webhook/resources/virtualmachine/mutator.go
martindekov: will leave this open no changes here

pohanhuang:
`reservedAffinityTerms()` preserves the current slice order, but reflect.DeepEqual() is order-sensitive. 

While array order is typically preserved through JSON serialization, I'm not sure if Kubernetes guarantee a canonical ordering for NodeSelectorTerms, MatchExpressions, or Values. So equivalent affinity rules with reordered entries could still be rejected.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -837,3 +851,41 @@ func (m *vmMutator) patchManagedTapBinding(vm *kubevirtv1.VirtualMachine, patchO
...
+	var terms []v1.NodeSelectorTerm
+	for _, term := range affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
+		for _, expr := range term.MatchExpressions {
+			if expr.Key == kubevirtv1.CPUManager {
+				terms = append(terms, term)
+				break
+			}
+		}
+	}
+	return terms
```
pohanhuang: Using sort to preserve the ordering of terms to ensure the order maintain.
```suggestion
	return sort(terms)
```

What do you think?

martindekov: Thanks for the feedback PoHan, you are right order is not guaranteed, but in our case from the above we have only 1 term so this slice as of now can only be of length 1 so ordering it is redundant. Further down we could add ordering when we decide to add more reserved terms to the list now it would be wasted function call.

Regarding DeepEqual direct usage we use it in a lot of places to compare objects in the code and I just followed:
https://github.com/search?q=repo%3Aharvester%2Fharvester%20.DeepEqual&type=code

Even we compare whole vm objects or annotations which should follow the unordered feedback you gave:
https://github.com/search?q=repo%3Aharvester%2Fharvester+.DeepEqual%28vm&type=code

I am curious on the internals of how that work and why we use it so often. I will leave the discussion open and in case you find the response meaningful you can resolve it or raise concern on what I wrote

pohanhuang: I think the code is in the right place, thanks for explaining and bring up the context

votdev: I'm also in favor of sorting, even though we only have one term right now. As luck would have it, the list won't be sorted if we add more terms down the line. That's why I think we should implement the function properly now. I am in favor of proactive implementation.

martindekov: ok, will sort in incoming commit will resolve the discussion once I add it

pohanhuang:
Overall LGTM, just one minor part about the `TestMutatorUpdate`, should we also test the controller request here? I think most of the cases comes from user, not controller request.

martindekov:
So from all the feedback I gathered in here and through discussions:
* Rename the function to `cpuManagerAffinityTerms` to narrow down the semantics of the function - currently cpumanager is the only term which we reserve
* Add controller test case for the TestMutatorUpdate
* Sort the terms from `cpuManagerAffinityTerms`  just in case
* Change message to: `key %q is automatically operated by Harvester controller and can't be changed manually`

Once I finish off the other work will add commit addressing the above and re-request reviews:

cc @votdev @w13915984028 @pohanhuang

pkg/webhook/resources/virtualmachine/mutator.go
martindekov: Added slices sort before returning :heavy_check_mark:

pohanhuang (dismissed):
LGTM, thanks for the fix.

pkg/webhook/resources/virtualmachine/mutator.go
martindekov: I revisited this, no longer return slice, but rather the cpu manager expression directly if it's in the slice, so as we no longer return slices we no loner need to sort them.

pohanhuang:
Hi @martindekov , 
I'm a bit confused about the latest commits, can you share the reason behind it? I was wondering why did you remove the sorting.

martindekov:
Sorry @pohanhuang we were discussing offline with @w13915984028 that original change assumed multiple terms will be used as we drop not only cpumanager term but network related ones as well. But since the original problem was cpumanager related only we discussed that we should not leak the scope so collapsed all the functions into 1 - specific for cpumanager check and that's it I think 1 function for 1 term check is simpler.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -846,3 +856,35 @@ func (m *vmMutator) patchManagedTapBinding(vm *kubevirtv1.VirtualMachine, patchO
 
 	return patchOps, nil
 }
+
+func checkCpuManagerAffinityTermChanged(request *types.Request, oldVM, newVM *kubevirtv1.VirtualMachine) error {
```
martindekov, line 861: This basically encapsulates all the prvious logic under 1 method so the call to it is also clearer as well:
```
	if err := checkCpuManagerAffinityTermChanged(request, nil, vm); err != nil {
		return nil, err
	}
```

The original logic was started with assumption we might get the network term as well as it's dropped as well:
https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/virtualmachine/mutator.go#L531
but we agreed to keep the scope to the original issue, but I left the logic broken down for eventual addition, but really that might never happen so abstract code for potential future extension never worked out as expected.

So revisited the change everything is encapsulated in single function call for this term only.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -846,3 +856,35 @@ func (m *vmMutator) patchManagedTapBinding(vm *kubevirtv1.VirtualMachine, patchO
...
+			for _, expr := range term.MatchExpressions {
+				if expr.Key == kubevirtv1.CPUManager {
+					return &expr
+				}
+			}
+		}
+		return nil
+	}
+
+	if reflect.DeepEqual(getCPUManagerExpr(oldVM), getCPUManagerExpr(newVM)) {
```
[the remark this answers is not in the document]
martindekov: Rejection standpoint it's valid, more than 2 terms are not rejected by the webhook, functionally however user setting this cannot be bypassed. But will sort the slice for the more than 1 value case.

pkg/webhook/resources/virtualmachine/mutator_test.go
```diff
@@ -1790,6 +1797,296 @@ func TestPatchManagedTapBinding(t *testing.T) {
...
+	}
+
+	setup := func() types.Mutator {
+		clientset := fake.NewSimpleClientset()
+		createDefaultKubeVirt(clientset)
+		_ = clientset.Tracker().Add(&harvesterv1.Setting{
+			ObjectMeta: metav1.ObjectMeta{Name: "overcommit-config"},
+			Default:    `{"cpu":100,"memory":100,"storage":100}`,
+		})
+		return setupTestMutator(clientset)
```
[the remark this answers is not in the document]
martindekov: others used the string, will add

pohanhuang:
@martindekov Thanks for the implementation, it makes more sense now! 

However, could you add a test case specifically for non-cpumanager keys (e.g. network-related affinity) to explicitly verify that only cpumanager changes are rejected? 

This would make the intent clearer and prevent regressions.

pkg/webhook/resources/virtualmachine/mutator.go
```diff
@@ -846,3 +857,44 @@ func (m *vmMutator) patchManagedTapBinding(vm *kubevirtv1.VirtualMachine, patchO
...
+				}
+				return strings.Compare(fmt.Sprintf("%v", a.Values), fmt.Sprintf("%v", b.Values))
+			})
+		}
+		return exprs
+	}
+
+	if reflect.DeepEqual(getCPUManagerExprs(oldVM), getCPUManagerExprs(newVM)) {
+		return nil
+	}
```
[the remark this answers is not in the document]
martindekov: So tested all suggestions `unordered set, differences in value ordering (or nil vs empty slices) ` results are all rejected and actually key having multiple values - unordered set are rejected as well:

```
node-1:/home/rancher # kubectl patch vm vm-1 -n default --type=json \
>   -p='[{"op":"replace","path":"/spec/template/spec/affinity/nodeAffinity/requiredDuringSchedulingIgnoredDuringExecution/nodeSelectorTerms/0/matchExpressions/0/values","value":null}]'
The request is invalid: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution: key "cpumanager" is automatically operated by Harvester controller and can't be changed manually
node-1:/home/rancher # kubectl patch vm vm-1 -n default --type=json   -p='[{"op":"replace","path":"/spec/template/spec/affinity/nodeAffinity/requiredDuringSchedulingIgnoredDuringExecution/nodeSelectorTerms/0/matchExpressions/0/values","value":["true","false"]}]'
The request is invalid: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution: key "cpumanager" is automatically operated by Harvester controller and can't be changed manually
node-1:/home/rancher # kubectl patch vm vm-1 -n default --type=json   -p='[{"op":"replace","path":"/spec/template/spec/affinity/nodeAffinity/requiredDuringSchedulingIgnoredDuringExecution/nodeSelectorTerms/0/matchExpressions/0/values","value":[]}]'
The request is invalid: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution: key "cpumanager" is automatically operated by Harvester controller and can't be changed manually
```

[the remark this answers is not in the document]
martindekov: Resolving as practically impossible as per above testing ^

martindekov:
Thanks for the suggestion @pohanhuang good one. Rebased, addressed through testing last review from copilot as iirelevant and added your test suggestion for network related key - not rejected. Otherwise generally rejecting cpumanager among other affinity is present already.

I think this is the final look of the change. Will build re-test everything.

pohanhuang (dismissed):
LGTM.
nit: add test for controller modify the non-cpumanager term, ensure it pass.

martindekov:
Done @pohanhuang won't merge as we still haven't released 1.8 and master is mirror of it, discussed with @w13915984028.

pohanhuang (dismissed):
LGTM

w13915984028 (dismissed):
LGTM, thanks.

pohanhuang (approved):
LGTM

w13915984028 (approved):
LGTM, thanks.

martindekov:
Updated testing section, everything works like a fine wine. Thanks @pohanhuang @w13915984028 for the reviews

PR #10228, opened 2026-03-17
chore: bump general job image

bk201:
This ensures the bci-base image is packaged in the ISO as well.

#### Problem:

- The general job image is not bumped and packaged. The controller promotion jobs still use an outdated version.
- outdated version bci image is packaged in the ISO. This cause image not found failure when ugprading a cluster in an air-gapped environment. ([Related change](https://github.com/harvester/harvester/pull/10122/changes#diff-8c5fdd3c29c4031d6332031bafd4f1c97820a492be51384b2b22689c6322d31fR1106))

Screenshot 2026-03-17 at 8 20 45 AM

#### Solution:

- Bump the bci-base image to 16.0

#### Related Issue(s):

- https://github.com/harvester/harvester/issues/10110

#### Test plan:

- Create a new installation and make sure registry.suse.com/bci/bci-base:16.0 is pre-loaded.
- Upgrade v1.7.1 to the build ISO in an air-gapped environment and make sure the upgrade succeeds.

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -599,7 +599,7 @@ generalJob:
     # In the Golang limitation, we couldn't parse like "tag": 12.20.
     # So, we must use "tag": "12.20" instead.
     # More details in PR-4896 (https://github.com/harvester/harvester/pull/4896)
-    tag: 15.7
+    tag: "16.0"
 
 whereabouts:
   enabled: true


deploy/charts/harvester/values.yaml
```diff
@@ -599,7 +599,7 @@ generalJob:
     # In the Golang limitation, we couldn't parse like "tag": 12.20.
     # So, we must use "tag": "12.20" instead.
     # More details in PR-4896 (https://github.com/harvester/harvester/pull/4896)
-    tag: 15.7
+    tag: "16.0"
```
bk201, line 602: @Yu-Jack Can you help me double-check? I need to quote the 16.0 string here. Otherwise, the code will interpret the value as a float and pull the image `..../bci-base:16`

brandboat: This is a coincidence I think, the value 15.7/16.0 are all numbers, I guess YARML parser treat it as float not string, adding double quote here make sense.

brandboat (approved):
LGTM, thanks!

starbops (approved):
LGTM, thank you.

deploy/charts/harvester/values.yaml
Yu-Jack: I remember we dealt with this issue before. Just add double quote. But, I forgot which issue ..

Yu-Jack (approved):
Thanks!

bk201:
@mergify backport v1.8

PR #10262, opened 2026-03-20
chore: remove pull_request_target

Yu-Jack:
This's for RST requirement to remove all `pull_request_target`. Use two steps workflows to achieve the goal with workflow event trigger.

- First workflow saves all data we need in next workflow.
- Second workflow can have the write permission to add the comment or change the assignee in the PR.

I've tested this in my private repo. It seems good. We can try this.

### DIFF
--- a/.github/auto-assign-config.yaml
+++ b/.github/auto-assign-config.yaml
@@ -1,4 +1,3 @@
 addAssignees: author
 addReviewers: true
-numberOfReviewers: 0
 reviewers: []
--- a/.github/workflows/issue-management-link-backport-pr-action.yaml
+++ b/.github/workflows/issue-management-link-backport-pr-action.yaml
@@ -1,0 +1,85 @@
+name: "Issue Management Link Backport PR Issue"
+
+on:
+  workflow_run:
+    workflows: ["Issue Management Backport PR Collect Data"]
+    types: [completed]
+
+env:
+  REPO_NAME: harvester/harvester
+
+jobs:
+  link-backport-issue:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    permissions:
+      actions: read
+      pull-requests: write
+    steps:
+    - name: Download PR data artifact
+      uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
+      with:
+        name: pr-data
+        run-id: ${{ github.event.workflow_run.id }}
+        github-token: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+
+    - name: Load PR data
+      run: |
+        cat pr-data.env >> $GITHUB_ENV
+
+    - name: Link the PR with the backport issue
+      if: ${{ env.BACKPORT == 'true' && env.ORIGINAL_ISSUE_NUMBER != '' }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+        GH_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
+      run: |
+        for issue_number in ${{ env.ORIGINAL_ISSUE_NUMBER }}; do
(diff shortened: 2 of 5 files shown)


martindekov (commented):
Hey Jack, is this ready for review? Also could we put Description and wire issue to it to get some context for review, thanks!

Yu-Jack:
> Hey Jack, is this ready for review? Also could we put Description and wire issue to it to get some context for review, thanks!

No. It's not ready yet, but it'll be ready for review soon. The original issue is a private repo, so I won't put the link here. I'll have related members to review this. Thanks!

.github/auto-assign-config.yaml
```diff
@@ -1,4 +1,3 @@
 addAssignees: author
 addReviewers: true
-numberOfReviewers: 0
```
Yu-Jack, line 3: We don't use `rancher/gh-issue-mgr/auto-assign-action@main` anymore, but we can still use this file.

.github/workflows/issue-management-link-backport-pr-action.yaml
```diff
@@ -0,0 +1,85 @@
+name: "Issue Management Link Backport PR Issue"
+
+on:
+  workflow_run:
+    workflows: ["Issue Management Backport PR Collect Data"]
```
Yu-Jack, line 5: Original name is `"[Issue Management] Backport PR Collect Data"`, but the GHA parsing failed. So, just remove bracket.

pohanhuang:
Thanks for the work!

I have a question about the `workflow_run` trigger in this workflow.

Since `workflow_run` runs with the base repo's permissions (including `pull-requests: write`), have you tested the scenario where a PR is opened from a fork?

e.g. `jack/harvester` ← `test/harvester`

Specifically, I'm wondering - Can a fork PR trigger this workflow?

ihcsim (dismissed):

ihcsim:
> Specifically, I'm wondering - Can a fork PR trigger this workflow?

I think forks shouldn't be allowed to trigger this workflow. It should only be run for trusted PRs created by mergify in this repo.

For other reviewers, what I am looking for in the review is:

* the removal of the `pull_request_target` event handler from the workflows so that they no longer have broad permissions to be exploited for sensitive information like the r/w gh token
* the removal of any actions that check out fork pull requests and then executes in-tree scripts while the workflow is running. E.g., if a workflow has `run: ./scripts/version`, a fork pull request could have modified `./scripts/version` to do something else we didn't expect

Yu-Jack:
> Thanks for the work!
> 
> I have a question about the `workflow_run` trigger in this workflow.
> 
> Since `workflow_run` runs with the base repo's permissions (including `pull-requests: write`), have you tested the scenario where a PR is opened from a fork?
> 
> e.g. `jack/harvester` ← `test/harvester`
> 
> Specifically, I'm wondering - Can a fork PR trigger this workflow?

It can work after maintainers approve it. This is a fork PR from jack-qa/docs to Yu-Jack/docs (https://github.com/Yu-Jack/docs/pull/15). So, it should be fine. Also, if you're in the org, the workflow should be triggered automatically. cc @ihcsim  @pohanhuang 

image

---

image

---

Also, you can see the change in the preview website.

https://69c1d271dbfb77cd18232005--test-harvester-docs.netlify.app/v1.8/advanced/addons

image

ihcsim (dismissed):

ihcsim:
@Yu-Jack I see what you mean. I initially thought the question was specifically about the backport `workflow_run` workflow. Ideally, the backport workflow should only run on pull requests created by mergify, not just based on the PR title. I think it's ok for now because we are not checking out any code.

So yeah, only the "Collect Data" workflows are triggered by pull request events. These are where we dealt with any untrusted content. Then when successfully completed, the `workflow_run` workflows are run with trusted download content.

Yu-Jack:
Atter merging docs PR, this is a quick demo https://github.com/harvester/docs/pull/996. Also, i used sha to replace the version.

pohanhuang (approved):
LGTM, Thanks for the work!

ihcsim (approved):

Issue #6424, opened 2024-08-26
[FEATURE] Terraform provider: vGPU support


m-ildefons:
**Is your feature request related to a problem? Please describe.**

Add vGPU support to the Terraform provider

**Additional context**
https://docs.harvesterhci.io/v1.3/advanced/vgpusupport
https://kubevirt.io/user-guide/compute/mediated_devices_configuration/

Issue #6423, opened 2024-08-26
[FEATURE] Terraform provider: PCI passthrough


m-ildefons:
**Is your feature request related to a problem? Please describe.**

Add PCI/PCIe passthrough support to the Terraform provider.

**Additional context**
https://docs.harvesterhci.io/v1.3/advanced/addons/pcidevices
https://kubevirt.io/user-guide/compute/host-devices/

borismand:
Hi @m-ildefons, do you know if this is going to be added? 
Is there any programmatic way to attach a PCI passthrough device to a VM?
It would be great to have the option to find out what devices are not in use and to attach them to VMs programmatically.

m-ildefons:
Hi @borismand ,

thanks for showing interest.
This feature is getting implemented at some point, but I can't tell you when exactly that's going to be.

For the time being you'll have to make do with just the plain Kubernetes API.

asc5543:
## Test result
Test PASS

## Environment
* Harvester ISO version: `v1.9.0-dev-20260621`
* Underlying Infrastructure (e.g. Baremetal with Dell PowerEdge R630): `1 node via Baremetal`


## Test Scenario
### Preparation
1. <details> <summary> Install Harvester</summary>

    <img width="1915" height="1000" alt="Image" src="https://github.com/user-attachments/assets/46aba7f4-fac5-481a-a605-b37368813589" />

   </details>

1. <details> <summary> Setup terraform environment with the following command</summary>

    ```
    go mod vendor
    rm -rf ~/.terraform.d
    CGO_ENABLED=0 go build -mod=vendor -o bin/terraform-provider-harvester .
    mkdir -p ~/.terraform.d/plugins/registry.terraform.io/harvester/harvester/0.0.0-dev/linux_amd64
    cp bin/terraform-provider-harvester ~/.terraform.d/plugins/registry.terraform.io/harvester/harvester/0.0.0-dev/linux_amd64/terraform-provider-harvester_v0.0.0-dev
    ```

   </details>

1. <details> <summary> Download kubeconfig from harvester </summary>

    <img width="1915" height="1000" alt="Image" src="https://github.com/user-attachments/assets/755f1da1-1075-4535-9a24-3f1e142a5bc2" />

   </details>

### Test Steps
1. <details>
   <summary> ✔ Case 1 - Set up terraform config with Multiple VFs, multiple VMs with one VF each
   </summary>

    ```
    # Configure root PCI device, enable several VFs.
[...]

Issue #6422, opened 2024-08-26
[FEATURE] Terraform provider: USB passthrough


m-ildefons:
**Is your feature request related to a problem? Please describe.**

Add USB passthrough support to the Terraform provider.

**Additional context**
https://kubevirt.io/user-guide/compute/host-devices/#usb-host-passthrough
https://github.com/harvester/harvester/issues/1710
https://github.com/harvester/harvester/pull/5430

Issue #10026, opened 2026-02-08
[BUG] vGPU functionality is broken in v1.8 builds


ibrokethecloud:
### Describe the Bug

Harvester v1.8.x builds switch to SL Micro 6.2 based builds and bumped linux kernel to v6.14.x

NVIDIA KVM/AIE drivers for both time sliced and MIG backed vGPU's have changed the behaviour of devices to no longer generate mdev devices.

https://docs.nvidia.com/vgpu/latest/grid-vgpu-user-guide/index.html#creating-sriov-vgpu-device-vendor-specific-vfio-framework

Kubevirt relies on mdev devices for passing through the vGPU to a VM workload.

https://kubevirt.io/user-guide/compute/mediated_devices_configuration/

As a result of driver changes, vGPU's never get created by the pcidevices controller, as the controller is scanning for mdev devices for configuration of vGPU profiles. 

NVIDIA documentation suggests a different way of configuring the vGPU profiles, which needs to be factored into the pcidevices controller.

In addition since the devices are now passed as host devices, there will be a change in the UI workflow in how vGPU's are identified and passed through to the VM.

When using host devices, the libvirt domain attempts to bind the device to vfio driver, which now needs to be skipped, as the vGPU device is needs to stay bound to the nvidia driver to allow passthrough vGPU to work successfully.

### To Reproduce

- Install Harvester v1.8.x build a machine with A2
- Enable pcidevices controller
- Enable nvidia driver toolkit using one of the newer kvm drivers. like NVIDIA-Linux-x86_64-580.126.08-vgpu-kvm.run
- Enable sriovgpudevice backed by the A2 GPU

No vGPU devices are created as the /sys fs device structure has changed for nvidia vGPU devices

### Expected Behavior

vGPU's should be create successfully

### Support Bundle for Troubleshooting

N/A

### Environment

- Harvester version: v1.8-head
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): NVIDIA A2 GPU
- Rancher version:

ibrokethecloud:
related PR: https://github.com/kubevirt/kubevirt/pull/16710

TachunLin:
Just checked the latest `v1.8.0-dev-260215` since we revert the base OS back to SL Micro 6.1.



[...]

PR #10263, opened 2026-03-20
VMBuilder: Add host devices conditionally

m-ildefons:
Add methods to the VM builder for adding host devices conditionally. These methods contain checks to avoid adding the same host device multiple times.
By doing these checks in the VMBuilder interface, integrations like the Terraform provider do not need to implement them in multiple locations separately.

#### Problem:

The VMBuilder only offers basic support for adding host devices to VMs and no support for adding GPUs.
This leaves a lot of the logic, like checking for duplicated devices, to the integrations.

#### Solution:

Implement the missing functionality in the VMBuilder.

#### Related Issue(s):

related-to: harvester/harvester#6422
related-to: harvester/harvester#6423
related-to: harvester/harvester#6424

#### Test plan:

Unit test included.

### DIFF
--- a/pkg/builder/device.go
+++ b/pkg/builder/device.go
@@ -24,6 +24,7 @@ func (v *VMBuilder) Input(inputName string, inputType kubevirtv1.InputType, inpu
 	return v
 }
 
+// Unconditionally adds the specified host device
 func (v *VMBuilder) HostDevice(name, hostDeviceName, tag string) *VMBuilder {
 	hostDevices := v.VirtualMachine.Spec.Template.Spec.Domain.Devices.HostDevices
 	hostDevice := kubevirtv1.HostDevice{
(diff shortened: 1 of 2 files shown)


pkg/builder/device.go
```diff
@@ -38,8 +39,57 @@ func (v *VMBuilder) HostDevice(name, hostDeviceName, tag string) *VMBuilder {
...
+	}
+
+	if !found {
+		v.HostDevice(name, hostDeviceName, tag)
+	}
+	return v
+}
+
+// Unconditionally adds the speficied GPU device
+func (v *VMBuilder) GPU(name, hostDeviceName, tag string, opts *kubevirtv1.VGPUOptions) *VMBuilder {
```
ibrokethecloud, line 60: we probably do need GPU / AddGPU anymore since the vGPU devices are now passed as HostDevices

https://github.com/harvester/harvester/issues/10026

m-ildefons: I'd leave it for now.

1) It's still necessary for legacy NVidia GPUs, which don't support SR-IOV. Those expose virtual functions only as `mdev`.
2) Newer NVidia GPUs, which support SR-IOV, still support `mdev` as VFIO framework: https://docs.nvidia.com/vgpu/latest/grid-vgpu-user-guide/index.html#creating-vgpu-device-red-hat-el-kvm Only the vendor-specific VFIO framework does not expose `mdev` devices.
3) It may be necessary for non-NVidia GPUs (haven't checked)
4) The VMBuilder interface just mirrors what KubeVirt exposes and KubeVirt has not yet deprecated this in KubeVirt v1.7.0
5) This is going to be used in the Terraform provider. If possible, the Terraform provider should also work with slightly older Versions of Harvester (e.g. v1.7.1) and those still use the `mdev` VFIO framework, even for newer NVidia GPUs.

ibrokethecloud (approved):

bk201 (approved):

PR #10272, opened 2026-03-23
Fix validation order in VM webhook

m-ildefons:
#### Problem:

Order of validations in the VirtualMachine validation webhook could lead to changes to VM metadata being wrongfully accepted if the VM spec does not change.

#### Solution:

Change order of validations in the VirtualMachine validation webhook such that the short-circuit logic comes after the validation of metadata.

#### Related Issue(s):

related-to: harvester/harvester#6835

#### Test plan:

- Create test VM
- Update test VM with invalid label `harvesterhci.io/maintain-mode-strategy: fail`
- Update to test VM should be rejected

https://github.com/harvester/harvester/issues/6835#issuecomment-4110302850 test case "Valid strategy creation UI through Advanced Options"

### DIFF
--- a/pkg/webhook/resources/virtualmachine/validator.go
+++ b/pkg/webhook/resources/virtualmachine/validator.go
@@ -289,6 +289,10 @@ func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj run
 		return err
 	}
 
+	if err := v.checkMaintenanceModeStrategyIsValid(newVM, oldVM); err != nil {
+		return err
+	}
+
 	// This logic will return in case interfaces are existing and not changed
 	// make sure new validation logic is added above this comment
 	if oldVM.Spec.Template != nil && newVM.Spec.Template != nil && reflect.DeepEqual(oldVM.Spec.Template.Spec.Domain.Devices.Interfaces, newVM.Spec.Template.Spec.Domain.Devices.Interfaces) {
@@ -296,10 +300,6 @@ func (v *vmValidator) Update(_ *types.Request, oldObj runtime.Object, newObj run
 	}
 
 	if err := v.checkForDuplicateMacAddrs(newVM); err != nil {
-		return err
-	}
-
-	if err := v.checkMaintenanceModeStrategyIsValid(newVM, oldVM); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


martindekov (dismissed):
LGTM Moritz, do we have unit tests where we can add 1 to the slice so we can validate the behaviour before after? Approving beforehand

martindekov (dismissed):
Thanks for adding the test Moritz, LGTM!

m-ildefons:
Do you mean something like [this](https://github.com/harvester/harvester/pull/10272/changes/e74580e8b7b311292ff57617827e0831d59a5dcd)? Where the execution order of the individual checks is tested to ensure that nothing is omitted due to short-circuit returns?
There are already existing unit tests for the checker functions. For the maintenance mode strategy label those exercise all sorts of creation/update scenarios, so I know that the check is good. However as far as I can tell there was nothing testing the order of the checks within the `Update(...)` function, which is why I missed that short-circuit logic.

martindekov:
So it would've been good to integrate the test of something which already exists - like change annotation, no spec changes between old and new:
1. Old code (unit test) fails
2. New code (unit test) passes

That way it's easy to distinguish that in theory (unit tested) the change would work in practice (QA e2e test). I approve as not familiar with how current test suite for Update is structured regarding this exact validation logic, if it can be integrated - good, if not - separate test is good as well.

Overall I think that if you have 2 same VMs but with different annotations - no spec changes on:
func TestVmValidator_Update(t *testing.T)

should trigger the bug, no need to add separate test for this

votdev (approved):

martindekov (approved):
Thanks for addressing the final feedback Moritz, nothing to add LGTM!

martindekov:
Should we ping mergify to backport? Seems like label didn't open backported PR once this was merged cc @m-ildefons

m-ildefons:
Yeah, I noticed the label didn't trigger the automation as I expected.
Fortunately, this is a pretty easy backport, so I could do it manually fairly fast: https://github.com/harvester/harvester/pull/10283

votdev:
> Yeah, I noticed the label didn't trigger the automation as I expected. Fortunately, this is a pretty easy backport, so I could do it manually fairly fast: #10283

The label feature for doing this has been removed some time ago. You simply need to use the `(at)Mergifyio backport vX.Y` phrase.

Issue #10101, opened 2026-02-25
[BUG] stale nodeCache could lead to silent failures


apoorvajagtap:
### Describe the Bug

Currently, `planHandler.Onchanged` fetches the nodeList from cache [here](https://github.com/harvester/harvester/blob/master/pkg/controller/master/upgrade/plan_controller.go#L51), proceeds with cleanup only when all nodes are updated with `latestHash`.

In cases where the cache returns stale data showing some nodes still on the old hash (even though they’ve already been updated successfully), the handler returns nil [here](https://github.com/harvester/harvester/blob/master/pkg/controller/master/upgrade/plan_controller.go#L56), causing the event to be silently dropped.

This can result in:
- cleanup logic never running
- upgrade flows appearing stuck
- controller state not converging despite nodes being updated

### Expected Behavior

The controller should reliably trigger cleanup once all nodes are updated.

### Support Bundle for Troubleshooting

Available in the respective JIRA card.

### Environment

- Harvester version: 1.6.1

### Additional context

Related ticket: SURE-11060

### Workaround and Mitigation

Workaround:
If the upgrade reports `Successful`, adding `harvesterhci.io/image-cleanup-plan-completed: "true"` manually helps to trigger the cleanup.

w13915984028:
test plan:

refer: https://github.com/harvester/harvester/issues/10345#issuecomment-4258541548

albinsun:
Test PASS, close together with harvester/harvester/issues/10345.

PR #10274, opened 2026-03-23
Enqueue plans when all nodes are not updated to latestHash

apoorvajagtap:
#### Problem:
During plan completion checks for nodes, stale nodeCache could lead to silent failures.

#### Solution:
This commit ensures enqueuing of following Handlers:
- upgradeHandler when `UpgradeCompleted.IsUnknown`
- nodeHandler when `node.Status.NodeInfo.OSImage` is not yet at  expectedVersion 
- planHandler in case not all the nodes are not at latestPlanHash.

#### Related Issue(s):
#10101 

#### Test plan:
- Unit tests to cover all discussed scenarios were added

### DIFF
--- a/pkg/controller/master/upgrade/plan_controller.go
+++ b/pkg/controller/master/upgrade/plan_controller.go
@@ -2,29 +2,32 @@ package upgrade
 
 import (
 	"strconv"
+	"time"
 
+	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
+	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	upgradectlv1 "github.com/harvester/harvester/pkg/generated/controllers/upgrade.cattle.io/v1"
 	"github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io"
 	upgradev1 "github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io/v1"
 	v1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
-
-	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
-	ctlharvesterv1 "github.com/harvester/harvester/pkg/generated/controllers/harvesterhci.io/v1beta1"
-	upgradectlv1 "github.com/harvester/harvester/pkg/generated/controllers/upgrade.cattle.io/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
 )
 
 // planHandler syncs on plan completions
 // When a plan completes, it set the NodesPrepared condition of upgrade CRD to be true.
 type planHandler struct {
-	namespace     string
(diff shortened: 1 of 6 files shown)


w13915984028:
An alternative solution is to add `EnqueueAfter` before return nil, it can avoid the direct list from api-server each time. 

https://github.com/harvester/harvester/blob/7f502747a4cc7d7ed03e51db86f8591991421f77/pkg/controller/master/upgrade/plan_controller.go#L55-L57

BTW: The upgrade-controller does not have a `timer` atm, it could book the latest waiting object and uses the timer to trigger the latest object reconcilling if time is due.  This will ensure the whole upgrade process has a heart-beat and avoid such scenario. thanks.

apoorvajagtap:
The latest commit pushed incorporates the EnqueueAfter logic. I am currently working towards the global timer approach for upgrade-controller. Will mark the PR ready for review once those changes are tested & pushed here.

pkg/controller/master/upgrade/plan_controller.go
```diff
@@ -36,26 +38,19 @@ func (h *planHandler) OnChanged(_ string, plan *upgradev1.Plan) (*upgradev1.Plan
...
-	selector = selector.Add(*requirementPlanNotLatest)
-	nodes, err := h.nodeCache.List(selector)
+	nodesAtLatestHash, err := allAtLatestHash(h.nodeCache, plan)
 	if err != nil {
 		return plan, err
 	}
-	if len(nodes) != 0 {
+	if !nodesAtLatestHash {
+		jitter := time.Duration(rand.Intn(5)) * time.Second
+		h.planEnqueueAfter(plan, time.Second*20+jitter)
```
starbops: Could we make the interval as a const?

starbops (commented):
Could we make these as constants?

apoorvajagtap:
Hi @starbops @w13915984028 I have pushed the latest commit that ensures enqueuing of following Handlers:

- upgradeHandler when UpgradeCompleted.IsUnknown
- nodeHandler when node.Status.NodeInfo.OSImage is not yet at expectedVersion
- planHandler in case not all the nodes are not at latestPlanHash.

As of now I’ve kept the enqueue intervals for the handlers separate to avoid introducing complex dependencies (as each handler owns its failure domain independently with isolated failures) . If we find a global timer (similar to a watchdog) for the upgrade flow to be more appropriate, I’m happy to explore and implement that. Let me know your thoughts, thanks!

pkg/controller/master/upgrade/plan_controller.go
```diff
@@ -36,26 +38,20 @@ func (h *planHandler) OnChanged(_ string, plan *upgradev1.Plan) (*upgradev1.Plan
...
-	}
-	selector = selector.Add(*requirementPlanNotLatest)
-	nodes, err := h.nodeCache.List(selector)
+	nodesAtLatestHash, err := allAtLatestHash(h.nodeCache, plan)
 	if err != nil {
 		return plan, err
 	}
-	if len(nodes) != 0 {
+	if !nodesAtLatestHash {
+		// Add jitter to avoid multiple plans re-enqueuing simultaneously
```
w13915984028, line 52: please add a log here with namespace/name, e.g.
```
    name: hvst-upgrade-455j7-cleanup
    namespace: cattle-system
```
currently we don't have any log about the waiting for plan

w13915984028 (dismissed):
LGTM, thanks for the fix.

pkg/controller/master/upgrade/plan_controller.go
apoorvajagtap: Thanks Jian, added the logs.

w13915984028 (dismissed):
LGTM, thanks.

w13915984028:
@mergify backport v1.8

starbops (approved):
LGTM, thank you for the fix!

bk201:
@mergifyio backport v1.7

Issue #5820, opened 2024-05-16
[FEATURE] Baremetal container workload support GA


ibrokethecloud:
**Is your feature request related to a problem? Please describe.**

v1.3.0 introduced experimental support for baremetal workloads on harvesters.

This issue is to track the work need to mark this as a GA feature.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

ibrokethecloud:
we should ensure users cannot upgrade rke2 versions for harvester clusters from the cluster management UI

![Image](https://github.com/user-attachments/assets/30fc0611-efa6-4be1-86f4-a5c179aabefc)

belgaied2:
> we should ensure users cannot upgrade rke2 versions for harvester clusters from the cluster management UI
> 

@ibrokethecloud I would even suggest putting a webhook in place for `cluster.provisioning.cattle.io` that blocks the version change.

ibrokethecloud:
Based on some quick validation with Rancher v2.12.x and Harvester v1.6.x I can see that cluster-management page does not allow users to edit the Harvester clusters any more

<img width="1652" height="482" alt="Image" src="https://github.com/user-attachments/assets/9c11dd06-1c0b-4006-8cb3-eccd9f5cb96c" />

This is likely driven from label `provider.cattle.io: harvester` when a Harvester cluster is added to Rancher.

Some additional changes are needed to Rancher
* Remove `Experimental` flag from `harvester-baremetal-container-workload` feature flag

<img width="917" height="60" alt="Image" src="https://github.com/user-attachments/assets/31a4d6d6-6e2b-4572-a7a7-8115ccdcc97d" />

This needs a minor change in feature definitions: https://github.com/rancher/rancher/blob/main/pkg/features/feature.go#L113

* Need to add `Harvester` system namespaces to rancher setting `system-namespaces`. This setting ensures that rancher hides harvester system specific namespaces from the UI. The extra namespaces are `harvester-system`, `cattle-monitoring-system`, `cattle-logging-system` and `cattle-dashboards`
  The rancher defaults are applied from https://github.com/rancher/rancher/blob/main/pkg/settings/setting.go#L35C2-L35C18
 Without this change harvester and addon charts are visible in Rancher Apps, and this allows users ability to edit/upgrade/remove Harvester components from outside of Harvester

[...]

PR #10295, opened 2026-03-25
sync harvester specific namespaces into System project

ibrokethecloud:
#### Problem:

For baremetal workload support, we need to ensure the Harvester specific system namespaces are moved to System project when Harvester is imported into a Rancher cluster. This will make it easier for rbac via rancher as admins could be assigned specific roles on the System project to get access to harvester components.

#### Solution:

PR adds a minor namespace handler to update a harvester system namespace list into the System project by looking up project details from `kube-system` namespace, which is auto-populated into the System namespace by rancher when a cluster is imported. 

#### Related Issue(s):

https://github.com/harvester/harvester/issues/5820

### DIFF
--- a/pkg/controller/master/rancher/namespace.go
+++ b/pkg/controller/master/rancher/namespace.go
@@ -3,6 +3,8 @@ package rancher
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
@@ -14,6 +16,11 @@ import (
 
 	"github.com/harvester/harvester/pkg/settings"
 	"github.com/harvester/harvester/pkg/util"
+)
+
+const (
+	projectAnnotationKey   = "field.cattle.io/projectId"
+	defaultRecheckInterval = 300 * time.Second
 )
 
 func (h *Handler) onNamespaceRemoved(_ string, namespace *corev1.Namespace) (*corev1.Namespace, error) {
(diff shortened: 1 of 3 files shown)


pkg/controller/master/rancher/namespace.go
```diff
@@ -106,3 +113,85 @@ func (h *Handler) onNamespaceRemoved(_ string, namespace *corev1.Namespace) (*co
...
+		if err := h.addProjectIDToHarvesterSystemNamespaces(projectID); err != nil {
+			return namespace, err
+		}
+	} else {
+		if err := h.removeProjectIDFromHarvesterSystemNamespaces(); err != nil {
+			return namespace, err
+		}
+	}
+
+	h.NamespaceController.EnqueueAfter(key, defaultRecheckInterval)
```
ihcsim, line 134: why do we need to enqueue and recheck this namespace?

pkg/controller/master/rancher/namespaces_test.go
```diff
@@ -0,0 +1,69 @@
...
+				projectAnnotationKey: invalidProjectAnnotationValue,
+			},
+		},
+	}
+
+	demoNamespace = &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "harvester-demo",
+		},
+	}
```
ihcsim, line 36: minor: asset that `demoNamespace` has no annotation added

pkg/controller/master/rancher/namespace.go
ibrokethecloud: the idea was that if a user moves any of the system namespaces to a different project we can still recheck all managed namespaces and bring them back into system project

pkg/controller/master/rancher/namespaces_test.go
ibrokethecloud: i have updated the test cases to leverage this better

ihcsim (approved):

Yu-Jack (approved):
Thanks for the PR!

tserong (approved):

PR #10299, opened 2026-03-26
Remove the redundant filesystem volume mechanism and add data migration volume api

Vicente-Cheng:
#### Problem:
- Since we have a new interface on the UI, we don't need the old mechanism for filesystem volume
- add the data migration action for offline storage migration

#### Solution:
- drop the redundant mechanism for filesystem volume
- add the data migration action for offline storage migration

#### Related Issue(s):

#### Test plan:
- create harvester cluster
- ensure the volume creation could be completed with both interface PVC/DataVolume (refer video on https://github.com/harvester/harvester-ui-extension/pull/776)
- ensure the volume data migration could be completed w/o any issue (ref video on https://github.com/harvester/harvester-ui-extension/pull/776)


pkg/controller/master/pvc/pvc_controller.go
```diff
@@ -6,54 +6,16 @@ import (
...
-			}
-			logrus.Infof("Created filesystem blank source %s/%s for PVC %s/%s", pvc.Namespace, util.ImportSourceFSBlank, pvc.Namespace, pvc.Name)
-		} else {
-			return pvc, err
-		}
-	}
-	return pvc, nil
+	dataVolumeClient ctlcdiv1.DataVolumeClient
+	pvcClient        ctlcorev1.PersistentVolumeClaimClient
+	pvcController    ctlcorev1.PersistentVolumeClaimController
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

pkg/controller/master/pvc/register.go
```diff
@@ -12,17 +12,14 @@ const (
...
 	ctlpvc := management.CoreFactory.Core().V1().PersistentVolumeClaim()
 
 	pvcHandler := &pvcHandler{
-		volImportSourceClient: volImportSource,
-		dataVolumeClient:      dataVolume,
-		pvcClient:             ctlpvc,
-		pvcController:         ctlpvc,
+		dataVolumeClient: dataVolume,
+		pvcClient:        ctlpvc,
+		pvcController:    ctlpvc,
```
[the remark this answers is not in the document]
Vicente-Cheng: updated

pkg/api/volume/handler.go
```diff
@@ -461,6 +476,72 @@ func (h *ActionHandler) snapshot(_ context.Context, pvcNamespace, pvcName, snaps
...
+	}
+
+	vms, err := h.vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(pvcNamespace, pvcName))
+	if err != nil {
+		return fmt.Errorf("failed to check VMs using PVC %s/%s: %v", pvcNamespace, pvcName, err)
+	}
+	if len(vms) > 0 {
+		return fmt.Errorf("PVC %s/%s is currently attached to VM %s/%s", pvcNamespace, pvcName, vms[0].Namespace, vms[0].Name)
+	}
+
```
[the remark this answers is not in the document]
Vicente-Cheng: nice catch

pkg/api/volume/handler.go
```diff
@@ -461,6 +476,72 @@ func (h *ActionHandler) snapshot(_ context.Context, pvcNamespace, pvcName, snaps
...
+				},
+			},
+			Storage: &cdiv1.StorageSpec{
+				StorageClassName: &input.TargetStorageClassName,
+				Resources: corev1.VolumeResourceRequirements{
+					Requests: pvc.Spec.Resources.Requests,
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
Vicente-Cheng: storageprofile would help us

pkg/api/volume/handler.go
```diff
@@ -461,6 +476,76 @@ func (h *ActionHandler) snapshot(_ context.Context, pvcNamespace, pvcName, snaps
...
+	if err != nil {
+		return fmt.Errorf("failed to check VMs using PVC %s/%s: %v", pvcNamespace, pvcName, err)
+	}
+	if len(vms) > 0 {
+		return fmt.Errorf("PVC %s/%s is currently attached to VM %s/%s", pvcNamespace, pvcName, vms[0].Namespace, vms[0].Name)
+	}
+
+	if err := h.assertPVCNotInUse(pvcNamespace, pvcName); err != nil {
+		return fmt.Errorf("PVC %s/%s is currently in use: %v", pvcNamespace, pvcName, err)
+	}
```
WebberHuang1118: Just a nit:
```suggestion
	pvc, err := h.validateDataMigration(pvcNamespace, pvcName, input)
	if err != nil {
		return err
	}
```
```
func (h *ActionHandler) validateDataMigration(pvcNamespace, pvcName string, input DataMigrationInput) (*corev1.PersistentVolumeClaim, error) {
	pvc, err := h.pvcCache.Get(pvcNamespace, pvcName)
	if err != nil {
		return nil, fmt.Errorf("failed to get source PVC %s/%s: %v", pvcNamespace, pvcName, err)
	}

	if _, err := h.scCache.Get(input.TargetStorageClassName); err != nil {
		return nil, fmt.Errorf("failed to get target StorageClass %s: %v", input.TargetStorageClassName, err)
	}

	if _, err := h.pvcCache.Get(pvcNamespace, input.TargetVolumeName); err == nil {
		return nil, fmt.Errorf("PVC %s/%s already exists", pvcNamespace, input.TargetVolumeName)
	} else if !apierrors.IsNotFound(err) {
		return nil, fmt.Errorf("failed to check existing PVC %s/%s: %v", pvcNamespace, input.TargetVolumeName, err)
	}

	if _, err := h.dataVolumes.Get(pvcNamespace, input.TargetVolumeName, metav1.GetOptions{}); err == nil {
		return nil, fmt.Errorf("DataVolume %s/%s already exists", pvcNamespace, input.TargetVolumeName)
	} else if !apierrors.IsNotFound(err) {
		return nil, fmt.Errorf("failed to check existing DataVolume %s/%s: %v", pvcNamespace, input.TargetVolumeName, err)
	}

	vms, err := h.vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(pvcNamespace, pvcName))
	if err != nil {
		return nil, fmt.Errorf("failed to check VMs using PVC %s/%s: %v", pvcNamespace, pvcName, err)
	}
	if len(vms) > 0 {
		return nil, fmt.Errorf("PVC %s/%s is currently attached to VM %s/%s", pvcNamespace, pvcName, vms[0].Namespace, vms[0].Name)
	}

	if err := h.assertPVCNotInUse(pvcNamespace, pvcName); err != nil {
		return nil, fmt.Errorf("PVC %s/%s is currently in use: %v", pvcNamespace, pvcName, err)
	}

	return pvc, nil
}
```

pkg/api/volume/handler.go
```diff
@@ -461,6 +476,76 @@ func (h *ActionHandler) snapshot(_ context.Context, pvcNamespace, pvcName, snaps
...
+
+	vms, err := h.vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(pvcNamespace, pvcName))
+	if err != nil {
+		return fmt.Errorf("failed to check VMs using PVC %s/%s: %v", pvcNamespace, pvcName, err)
+	}
+	if len(vms) > 0 {
+		return fmt.Errorf("PVC %s/%s is currently attached to VM %s/%s", pvcNamespace, pvcName, vms[0].Namespace, vms[0].Name)
+	}
+
+	if err := h.assertPVCNotInUse(pvcNamespace, pvcName); err != nil {
```
brandboat: Might out of scope here, I'm wondering do we need similar check in webhook to prevent user from provisioning DV to perform volume migration through cli?

pkg/api/volume/handler.go
```diff
@@ -461,6 +476,76 @@ func (h *ActionHandler) snapshot(_ context.Context, pvcNamespace, pvcName, snaps
...
+
+	if _, err := h.dataVolumes.Create(dv); err != nil {
+		logrus.WithFields(logrus.Fields{
+			"namespace":  pvcNamespace,
+			"name":       pvcName,
+			"targetName": input.TargetVolumeName,
+			"apiVersion": "cdi.kubevirt.io/v1beta1",
+			"kind":       "DataVolume",
+			"err":        err,
+		}).Error("failed to create DataVolume for data migration")
```
brandboat: nit:
```
			"namespace":  pvcNamespace,
			"name":       pvcName,
			"apiVersion": "v1",
			"kind":       "PersistentVolumeClaim",
		}).WithError(err).Error("failed to create image from PVC")
```

pkg/api/volume/schema.go
```diff
@@ -56,12 +59,16 @@ func RegisterSchema(scaled *config.Scaled, server *server.Server, _ config.Optio
 				actionSnapshot: {
 					Input: "snapshotVolumeInput",
 				},
+				actionDataMigration: {
+					Input: "dataMigrationInput",
```
brandboat, line 63: ~~Or we can hide the action once the volume is attached to a pod, but I'm totally fine with the current approach.~~ It's already done in the current implementation.

brandboat (dismissed):
LGTM, nothing material, just some minor comments

pkg/controller/master/virtualmachine/vm_controller.go
WebberHuang1118, line 125: Suggest reducing the nesting:
```suggestion
		pvc, err := h.pvcCache.Get(vm.Namespace, pvcAnno.Name)
		if err != nil && !apierrors.IsNotFound(err) {
			return nil, err
		}
		if apierrors.IsNotFound(err) {
			if !createPVCWithDataVolume {
				pvcAnno.Namespace = vm.Namespace
				if _, err = h.pvcClient.Create(pvcAnno); err != nil {
					return nil, err
				}
			}
			continue
		}
```

WebberHuang1118 (dismissed):
LGTM. Just a few refactoring suggestions—thanks for the enhancement.

pkg/api/volume/handler.go
```diff
@@ -461,6 +476,76 @@ func (h *ActionHandler) snapshot(_ context.Context, pvcNamespace, pvcName, snaps
...
+		return fmt.Errorf("failed to check existing PVC %s/%s: %v", pvcNamespace, input.TargetVolumeName, err)
+	}
+
+	if _, err := h.dataVolumes.Get(pvcNamespace, input.TargetVolumeName, metav1.GetOptions{}); err == nil {
+		return fmt.Errorf("DataVolume %s/%s already exists", pvcNamespace, input.TargetVolumeName)
+	} else if !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to check existing DataVolume %s/%s: %v", pvcNamespace, input.TargetVolumeName, err)
+	}
+
+	vms, err := h.vmCache.GetByIndex(indexeresutil.VMByPVCIndex, ref.Construct(pvcNamespace, pvcName))
```
ibrokethecloud: we already check this before allow action to be added to the volume, do we need to check again?

ibrokethecloud (dismissed):

pkg/api/volume/handler.go
Vicente-Cheng: discussed offline, keep that to see any feedback.

pkg/api/volume/handler.go
Vicente-Cheng: nice suggestion, udpated

pkg/api/volume/handler.go
Vicente-Cheng: updated!

pkg/controller/master/virtualmachine/vm_controller.go
Vicente-Cheng: updated

pkg/api/volume/handler.go
Vicente-Cheng: good point, let me remove the redundant

brandboat (dismissed):

w13915984028:
@Vicente-Cheng Do we have detailed test-plan/report to refer the full story? thanks.

Vicente-Cheng:
> @Vicente-Cheng Do we have detailed test-plan/report to refer the full story? thanks.

Thanks for the reminder, I updated the test case. 
For the whole story, you can refer to the non-root volume issue [[FEATURE] Apply CDI Settings To Non-Root Volumes #9412](https://github.com/harvester/harvester/issues/9412).

The main goal is to expose the data volume interface to the user.
And data migration would be use of the offline migration story

WebberHuang1118 (dismissed):

pkg/api/volume/handler.go
```diff
@@ -461,6 +476,77 @@ func (h *ActionHandler) snapshot(_ context.Context, pvcNamespace, pvcName, snaps
 	return nil
 }
 
+func (h *ActionHandler) dataMigration(_ context.Context, pvcNamespace, pvcName string, input DataMigrationInput) error {
+	pvc, err := h.validateDataMigration(pvcNamespace, pvcName, input)
+	if err != nil {
+		return err
+	}
+
+	dv := &cdiv1.DataVolume{
```
w13915984028, line 485: After a DataVolume is handle off for migration, if it is more than creating a CRD object (from second video on https://github.com/harvester/harvester-ui-extension/pull/776, it took a while even for the simple test volume), how to ensure the following process:

- Know the status of migration
- When it is migrating, allow/don't allow another migration
- Prevent source PVC from being deleted
- Prevent source PVC from being attached to VM
- If necessary, allow `cancel data migration`

w13915984028 (commented):
A main question, thanks.

pkg/api/volume/handler.go
Vicente-Cheng: >Know the process of migration

DataVolume has progress, but volume page did not have, we can enhance that in later version

>When it is migrating, allow/don't allow another migration

we block the storage migration once it was used by any pod, so it would block another migration

>Prevent source PVC from being deleted

Delete source PVC would not be deleted since it was attached

>Prevent source PVC from being attached to VM

address by this commit https://github.com/harvester/harvester/pull/10299/changes/39890a4e21fdafa6588da890fe8b15cfd9daec9f
BTW, originally we don't have any corresponding check for attach VM, so this is not the only case would encounter the issue

>If necessary, allow cancel data migration

We can do it once we get any feedback.

Vicente-Cheng: Also, we can trace those things in a new issue. 
If we don't have major concerns about the current one, we can create some enhancement tickets for follow-up.

w13915984028: Offload the open topics to https://github.com/harvester/harvester/issues/10305, thanks.

WebberHuang1118 (dismissed):

brandboat (dismissed):

w13915984028 (dismissed):
LGTM, thanks.

BTW, on the document PR, the limitation needs to be mentioned: to migrate a PVC, it should be detached from any workloads first, just stop it's parent VM is not enough.

Vicente-Cheng:
We need to handle
- Prevent source PVC from being attached to VM

by issue https://github.com/harvester/harvester/issues/10305

So I keep the PR clean, and we can do further enhancement.

Vicente-Cheng:
@mergifyio backport v1.8

w13915984028 (approved):
LGTM, thanks.

Vicente-Cheng:
Since @WebberHuang1118 already approved with this version, bypass the rule

PR #10300, opened 2026-03-26
fix: check if host device has parent sriovgpu label

Yu-Jack:
#### Problem:
Because vm.domain.devices.gpus is no longer used ([reason](https://github.com/harvester/pcidevices/pull/159)), we don't need to check gpus anymore.

#### Solution:
Use dynamic client to fetch details.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9531
https://github.com/harvester/pcidevices/pull/159

#### Test plan:

disabling case 01 : VM with a vgpu device in host device filed.

https://github.com/user-attachments/assets/e9f34a2c-1be1-4c55-a093-9ce95819dae0

disabling case 02: VM with a nvidia host device. Allow disabling nvidia-toolkit addon because it's a host device.

https://github.com/user-attachments/assets/487a4a7b-b143-468f-b0d5-9b8a72c90f8f

### DIFF
--- a/pkg/webhook/resources/addon/validator.go
+++ b/pkg/webhook/resources/addon/validator.go
@@ -1,17 +1,22 @@
 package addon
 
 import (
+	"context"
 	"fmt"
+	"time"
 
 	ctlcorev1 "github.com/rancher/wrangler/v3/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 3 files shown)


ibrokethecloud (approved):

WebberHuang1118 (approved):
LGTM. thanks for the enhancement.

Vicente-Cheng:
@mergifyio backport v1.8

Issue #10243, opened 2026-03-18
[ENHANCEMENT] Ensure the Kubevirt volume operation call works with v1.8.0


Vicente-Cheng:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Since we moved from `HotplugVolumes` to `DeclarativeHotplugVolumes`, some behavior would be blocked.
For example, we cannot operate the VMI directly with a subresource call (related PR: https://github.com/harvester/harvester-csi-driver/pull/69)
REF: https://kubevirt.io/user-guide/storage/hotplug_volumes/

With `DeclarativeHotplugVolumes`, Kubevirt has better consistency handling for the VM/VMI.
but we need to change the caller. 
So, we need to move to `patch` the VM instead of the subresource call.

### Describe the solution you'd like

Check all the current subresource call now, and moved them to `patch` if needed

Vicente-Cheng:
The Harvester CSI Driver was the only one affected directly (the function was broken), others would keep working.
So, since we already fixed that on the Harvester CSI driver, moved to v1.9.
I can handle the others.

hoo29:
Has there been stressed testing with the new `DeclarativeHotplugVolumes` support? It looks like there are open issues on kubevirt https://github.com/kubevirt/kubevirt/issues/16937 that might cause something similar to https://github.com/harvester/harvester/issues/9761 to happen again.

Vicente-Cheng:
> Has there been stressed testing with the new `DeclarativeHotplugVolumes` support? It looks like there are open issues on kubevirt [kubevirt/kubevirt#16937](https://github.com/kubevirt/kubevirt/issues/16937) that might cause something similar to [#9761](https://github.com/harvester/harvester/issues/9761) to happen again.

Hi @hoo29, 
I will retry the stress testing with the new feature gate to see how Harvester-CSI-Driver works.

TachunLin:
Verified **fixed** on `v1.9.0-rc4`. Close this issue. 

### Result
<details><summary>$\color{green}{\textsf{PASS}}$ Trigger upgrade when vm in volume hotplug should be blocked by webhook $~~$</summary>

1. When the VM have a pending `volumeRequest` like this

  ```
  harvester-node-0:~ # kubectl get vm vm1 -n default -o json | jq '.status.volumeRequests'
  [
    {
      "addVolumeOptions": {
        "disk": {
          "disk": {
            "bus": "scsi"
          },
          "name": "stuckvol999",
[...]

PR #10306, opened 2026-03-26
Add vm patch access controler to csi-driver cluster role

brandboat:
#### Problem:

this is a follow up of https://github.com/harvester/harvester-csi-driver/pull/75, now we need the vm patch permission in  https://github.com/harvester/harvester-csi-driver/blob/master/pkg/csi/controller_server.go#L868

#### Solution:

Add vm patch permission to `harvesterhci.io:csi-driver`

#### Related Issue(s):

https://github.com/harvester/harvester/issues/10243

#### Test plan:

1. Prepare a harvester cluster that include this patch.
2. Provision a rke2 guest cluster
3. Update harvester-cloud-provider image to `v0.2.17`, you can run the cmd below
```
kubectl set image daemonset/harvester-csi-driver -n kube-system harvester-csi-driver=rancher/harvester-csi-driver:v0.2.7
```
4. create a pvc
```
kubectl apply -f - <<EOF
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: fs-vol
spec:
  accessModes:
    - ReadWriteOnce
  volumeMode: Filesystem
  resources:
    requests:
      storage: 2Gi
EOF
```
5. make sure the pvc is in Bound Status
```
root@rke2-pool1-znpgh-jfwrv:~# kubectl get pvc
NAME     STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
fs-vol   Bound    pvc-ef88c4d3-ac53-46c2-a6ca-ba4f2461cae7   2Gi        RWO            harvester      <unset>                 15m
```

### DIFF
--- a/deploy/charts/harvester/templates/rbac.yaml
+++ b/deploy/charts/harvester/templates/rbac.yaml
@@ -215,3 +215,9 @@ rules:
   - volumesnapshotcontents/status
   verbs:
   - '*'
+- apiGroups:
+  - kubevirt.io
+  resources:
+  - virtualmachines
+  verbs:
+  - patch
--- a/deploy/charts/harvester/templates/rbac.yaml
+++ b/deploy/charts/harvester/templates/rbac.yaml
@@ -169,55 +169,55 @@ metadata:
     app.kubernetes.io/component: apiserver
   name: harvesterhci.io:csi-driver
 rules:
-- apiGroups:
-  - storage.k8s.io
-  resources:
-  - storageclasses
-  - volumeattachments
-  verbs:
-  - get
-  - list
-  - watch
-- apiGroups:
-  - harvesterhci.io
-  resources:
-  - networkfilesystems
-  - networkfilesystems/status
-  - volumeremotebackups
-  - volumeremoterestores
-  verbs:
-  - '*'
-- apiGroups:
-  - harvesterhci.io
-  resources:
-  - settings
-  verbs:
-  - get
-  - list
-  resourceNames:
-  - csi-online-expand-validation
-  - csi-driver-config
-- apiGroups:
-  - longhorn.io
-  resources:
-  - volumes
-  - volumes/status
-  verbs:
-  - get
-  - list
-- apiGroups:
-  - snapshot.storage.k8s.io
-  resources:
-  - volumesnapshots
-  - volumesnapshotcontents
-  - volumesnapshotclasses
-  - volumesnapshots/status
-  - volumesnapshotcontents/status
-  verbs:
-  - '*'
-- apiGroups:
-  - kubevirt.io
-  resources:
-  - virtualmachines
-  verbs:
-  - patch
+  - apiGroups:
+      - storage.k8s.io
+    resources:
+      - storageclasses
+      - volumeattachments
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - harvesterhci.io
+    resources:
+      - networkfilesystems
+      - networkfilesystems/status
+      - volumeremotebackups
+      - volumeremoterestores
+    verbs:
+      - '*'
+  - apiGroups:
+      - harvesterhci.io
+    resources:
+      - settings
+    verbs:
+      - get
+      - list
+    resourceNames:
+      - csi-online-expand-validation
+      - csi-driver-config
+  - apiGroups:
+      - longhorn.io
+    resources:
+      - volumes
+      - volumes/status
+    verbs:
+      - get
+      - list
+  - apiGroups:
+      - snapshot.storage.k8s.io
+    resources:
+      - volumesnapshots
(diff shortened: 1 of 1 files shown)


deploy/charts/harvester/templates/rbac.yaml
w13915984028, line 154: <del>it it ok to add `patch` after L154 ? thanks. </del> 

skip ../

brandboat: I’m a bit hesitant about this. This change grants broad patch permissions for VMIs, which also extends to cloud-provider-harvester. I'm concerned we might be giving away too much permission, but I’d love to hear what others think

deploy/charts/harvester/templates/rbac.yaml
```diff
@@ -214,4 +214,10 @@ rules:
...
   - volumesnapshotcontents/status
   verbs:
-  - '*'
\ No newline at end of file
+  - '*'
+- apiGroups:
+  - kubevirt.io
+  resources:
+  - virtualmachines
+  verbs:
```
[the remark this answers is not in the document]
brandboat: fixed in [8dd7751](https://github.com/harvester/harvester/pull/10306/commits/8dd7751838b76f0acbe7af057969a0423599c5fe)

w13915984028:
Does this requirement mean the csi-driver v1.27 can only work upon Harvester v1.8.0 by default? For legacy Harvester versions, manually update Harvester RBAC is required. thanks.

Vicente-Cheng:
> Does this requirement mean the csi-driver v1.27 can only work upon Harvester v1.8.0 by default? For legacy Harvester versions, manually update Harvester RBAC is required. thanks.

csi driver v0.1.27 will check the host cluster version, 
Before v1.8, we would not use `patch`
see here: https://github.com/harvester/harvester-csi-driver/blob/master/pkg/csi/controller_server.go#L531-L536

how useDeclarativeHotplug configure:
see here: https://github.com/harvester/harvester-csi-driver/blob/master/pkg/csi/manager.go#L142-L151

Vicente-Cheng (approved):
thanks for adding this

w13915984028:
Another question: thanks.

`harvesterhci.io:cloudprovider` has following rbac, but `harvesterhci.io:csi-driver` has no, 

is below line not requiring such rbac? 

https://github.com/harvester/harvester-csi-driver/blob/846c761e76b817c23503af87643e0e6c4dbbc809/pkg/csi/controller_server.go#L792
```
	vm, err := cs.virtClient.VirtualMachine(cs.namespace).Get(context.TODO(), nodeID, metav1.GetOptions{})
```

```
  - apiGroups:
      - kubevirt.io
    resources:
      - virtualmachines
      - virtualmachineinstances
    verbs:
      - get
      - list
      - watch
```

Vicente-Cheng:
I remember I checked in the live environment, the harvester-csi-driver has that capability. 
But that's a good question, let me check again.

brandboat:
The harvester-csi-driver kubeconfig is generated by this api `POST /v1/harvester/kubeconfig`, and the API creates a ServiceAccount and binds it to two ClusterRoles: `harvesterhci.io:cloudprovider`, `harvesterhci.io:csi-driver`.

That's why harvester-csi-driver have both `harvesterhci.io:cloudprovider` & `harvesterhci.io:csi-driver` permissions.

https://github.com/rancher/dashboard/blob/master/shell/edit/provisioning.cattle.io.cluster/rke2.vue#L1780-L1801
https://github.com/harvester/harvester/blob/master/pkg/api/kubeconfig/generate_handler.go#L93-L130

w13915984028:
> The harvester-csi-driver kubeconfig is generated by this api `POST /v1/harvester/kubeconfig`, and the API creates a ServiceAccount and binds it to two ClusterRoles: `harvesterhci.io:cloudprovider`, `harvesterhci.io:csi-driver`.
> 
> That's why harvester-csi-driver have both `harvesterhci.io:cloudprovider` & `harvesterhci.io:csi-driver` permissions.
> 
> https://github.com/harvester/harvester/blob/master/pkg/api/kubeconfig/generate_handler.go#L93-L130

Good output.

Then is it better to put the `patch` under `harvesterhci.io:cloudprovider`? to keep consistency. thanks.

Vicente-Cheng:
I am fine with keeping that on harvester-csi-driver, but we need a comment explaining the reason.

Vicente-Cheng:
@brandboat could you create the document on the Harvester-CSI-Driver repo?

Consider the RC3 timeline. I would like to merge this first

Vicente-Cheng:
@mergifyio backport v1.8

w13915984028:
@mergify backport v1.7

PR #10326, opened 2026-03-30
HEP: support CDI way for non-root volume [skip ci]

Vicente-Cheng:
#### Problem:
HEP: support CDI way for non-root volume

#### Solution:
HEP: support CDI way for non-root volume

#### Related Issue(s):

We discussed the feature offline. This HEP is the record for the discussion summary.


WebberHuang1118 (approved):
LGTM, thanks.

ihcsim (approved):

Issue #10359, opened 2026-04-09
[BUG] VM Network disappeared after a v1.7.1 to v1.8.0-rc5 upgrade


albinsun:
### Describe the Bug

In recent upgrade regression from `v1.7.1` to `v1.8.0-rc5` we occasionally found that the VLAN NAD may disappear after upgrade even though the VMs uses the NAD still there.

<img width="1562" height="819" alt="Image" src="https://github.com/user-attachments/assets/2c4b8175-8260-4b28-9ce8-2b9e84c9440d" />

### To Reproduce

It's a regression test suite, roughly the steps are:

1. Setup a 4 nodes harvester-v1.7.1
1. Create a simple VM (**ug-vm**), a VM with custom storage class (**sc-ug-vm**)
1. Backup the simple VM
1. Restore backup to a new VM (**r-ug-vm**)
1. Upgrade to `v1.8.0-rc5`
1. ❌ Check VM and network status
   > VMs still `Running` but the NAD is gone.
   <img width="1562" height="473" alt="Image" src="https://github.com/user-attachments/assets/7c149c89-5af0-426e-a31a-c6a93f313dca" />
   <img width="1562" height="819" alt="Image" src="https://github.com/user-attachments/assets/2c4b8175-8260-4b28-9ce8-2b9e84c9440d" />

### Expected Behavior

VM Network should still exist and work after upgrade.

### Support Bundle for Troubleshooting

* [hvst-upgrade-bd8zn-upgradelog-archive-2026-04-09T15-21-58Z.zip](https://github.com/user-attachments/files/26608334/hvst-upgrade-bd8zn-upgradelog-archive-2026-04-09T15-21-58Z.zip)
* [supportbundle_50367575-d418-43e3-a06e-17e7a10e3b2b_2026-04-09T15-22-46Z.zip](https://github.com/user-attachments/files/26608341/supportbundle_50367575-d418-43e3-a06e-17e7a10e3b2b_2026-04-09T15-22-46Z.zip)

### Environment

- Harvester version: `v1.7.1` -> `v1.8.0-rc5`
- Impacted VM:
  * `ug-vm-08h11m12s343249-04-09`   
  * `sc-ug-vm-08h11m12s343249-04-09`
  * `r-ug-vm-08h11m12s343249-04-09`
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): 4 nodes baremetal
- Rancher version:

[...]

PR #10366, opened 2026-04-13
Patch rke2-multus chart failurepolicy to abort

w13915984028:
#### Problem:

`rke2-multus` could be reinstalled and cause the existing NADs are lost

code: 
https://github.com/k3s-io/klipper-helm/blob/293bbdb989cf138f680d1eb6ea04672312bce1b4/entry#L69

#### Solution:

Patch rke2-multus chart failurepolicy to abort, to avoid `reinstall` to remove all existing NADs

#### Related Issue(s):

https://github.com/harvester/harvester/issues/10359

#### Test plan:

As the issue is hard to reproduce, we run the script locally to ensure it's function:

```
fresh run:
...
check and patch rke-multus failurepolicy to abort, to avoid it potentially reinstall chart and caused NAD CRD recreated
Resource 'rke2-multus' not found. Creating...
helmchartconfig.helm.cattle.io/rke2-multus created
Waiting 10s for RKE2 controller to sync new config...


kubectl get job -n kube-system helm-install-rke2-multus -oyaml
        - name: FAILURE_POLICY
          value: abort

re-run:

check and patch rke-multus failurepolicy to abort, to avoid it potentially reinstall chart and caused NAD CRD recreated
Verified: failurePolicy is already 'abort'.

```

We have another PR https://github.com/harvester/harvester/pull/10402 to drop&remove a .skip file under rke2 manifest path, to guide rke2 only upgrade rke2-multus chart when then upgrade is done, to reduce the racing window.

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -5,12 +5,69 @@ UPGRADE_TMP_DIR="/tmp/upgrade"
 
 source $SCRIPT_DIR/lib.sh
 
+# The default RKE2 failurePolicy is `reinstall`.
+# Recreating the NAD CRD can cause all VM networking to be lost.
+patch_rke2_multus_config() {
+  echo "Check and patch rke2-multus failurePolicy to abort, to avoid a reinstall of the chart and recreation of the NAD CRD."
+
+  local name="rke2-multus"
+  local namespace="kube-system"
+  local manifest="$UPGRADE_TMP_DIR/rke2-multus-helmchartconfig.yaml"
+
+  mkdir -p "$UPGRADE_TMP_DIR"
+
+  # 1. Prepare Manifest
+  cat > "$manifest" <<EOF
+apiVersion: helm.cattle.io/v1
+kind: HelmChartConfig
+metadata:
+  name: $name
+  namespace: $namespace
+spec:
+  failurePolicy: abort
+EOF
+
+  # 2. Check if the resource exists
+  local get_output
+  get_output=$(kubectl get helmchartconfig "$name" -n "$namespace" 2>&1) || true
+  if [[ "$get_output" == *"NotFound"* || "$get_output" == *"not found"* ]]; then
+      echo "Resource '$name' not found. Creating..."
+      kubectl apply -f "$manifest"
+      echo "Waiting 10s for RKE2 controller to sync new config..."
+      sleep 10
+  elif [[ "$get_output" == *"Error from server"* ]]; then
+      # This catches RBAC or API errors that aren't "NotFound"
+      echo "CRITICAL ERROR: kubectl check failed with: $get_output"
+      return 1
+  else
+
+    # 3. Resource exists, check policy
+    local current_policy
+    current_policy=$(kubectl get helmchartconfig "$name" -n "$namespace" -o jsonpath='{.spec.failurePolicy}')
+
+    if [ "$current_policy" != "abort" ]; then
+      echo "Current policy is '${current_policy:-unset}'. Patching to 'abort'..."
+      kubectl patch helmchartconfig "$name" -n "$namespace" \
+        --type merge -p '{"spec":{"failurePolicy":"abort"}}'
+      echo "Waiting 10s for RKE2 controller to sync patch..."
+      sleep 10
+    else
+      echo "Verified: failurePolicy is already 'abort'. No action required."
+    fi
+  fi
+}
+
 pre_upgrade_manifest() {
   if [ -e "/usr/local/share/migrations/upgrade_manifests/${UPGRADE_PREVIOUS_VERSION}/pre-hook.sh" ]; then
     echo "Executing ${UPGRADE_PREVIOUS_VERSION} pre-hook..."
     # Use source to pass current shell's variables to target script
     source "/usr/local/share/migrations/upgrade_manifests/${UPGRADE_PREVIOUS_VERSION}/pre-hook.sh"
   fi
+
+  # Safety Gate: Ensure rke2-multus helmchart failurePolicy is 'abort' before proceeding.
+  # Since 'set -e' is active, any failure here will safely halt the upgrade
+  # to prevent potential NAD CRD deletion and VM networking loss.
+  patch_rke2_multus_config
 }
 
 # Preserve the current overcommit-config value before upgrade.
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -28,32 +28,40 @@ spec:
 EOF
 
   # 2. Check if the resource exists
-  local get_output
-  get_output=$(kubectl get helmchartconfig "$name" -n "$namespace" 2>&1) || true
-  if [[ "$get_output" == *"NotFound"* || "$get_output" == *"not found"* ]]; then
+  local EXIT_CODE=0
+  # Using a global variable to ensure EXIT_CODE is captured correctly under 'set -e'
+  rke2_multus_chart_config_output=$(kubectl get helmchartconfig "$name" -n "$namespace" 2>&1) || EXIT_CODE=$?
+
+  if [[ $EXIT_CODE -ne 0 ]]; then
+    if [[ "$rke2_multus_chart_config_output" == *"NotFound"* || "$rke2_multus_chart_config_output" == *"not found"* ]]; then
       echo "Resource '$name' not found. Creating..."
       kubectl apply -f "$manifest"
       echo "Waiting 10s for RKE2 controller to sync new config..."
       sleep 10
-  elif [[ "$get_output" == *"Error from server"* ]]; then
-      # This catches RBAC or API errors that aren't "NotFound"
-      echo "CRITICAL ERROR: kubectl check failed with: $get_output"
-      return 1
-  else
-
-    # 3. Resource exists, check policy
-    local current_policy
-    current_policy=$(kubectl get helmchartconfig "$name" -n "$namespace" -o jsonpath='{.spec.failurePolicy}')
-
-    if [ "$current_policy" != "abort" ]; then
-      echo "Current policy is '${current_policy:-unset}'. Patching to 'abort'..."
-      kubectl patch helmchartconfig "$name" -n "$namespace" \
-        --type merge -p '{"spec":{"failurePolicy":"abort"}}'
-      echo "Waiting 10s for RKE2 controller to sync patch..."
-      sleep 10
+      # When 'kubectl apply' is successful, we don't check helm-install job;
+      # RKE2 controller ensures the HelmChartConfig is used.
+      return 0
     else
-      echo "Verified: failurePolicy is already 'abort'. No action required."
+      # Catch critical errors (RBAC, API timeouts, etc.)
+      echo "CRITICAL ERROR: kubectl check failed with: $rke2_multus_chart_config_output EXIT_CODE:$EXIT_CODE"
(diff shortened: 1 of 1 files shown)


w13915984028:
@mergify backport v1.8

package/upgrade/upgrade_manifests.sh
```diff
@@ -5,12 +5,77 @@ UPGRADE_TMP_DIR="/tmp/upgrade"
...
+
+  if [[ $EXIT_CODE -ne 0 ]]; then
+    if [[ "$rke2_multus_chart_config_output" == *"NotFound"* || "$rke2_multus_chart_config_output" == *"not found"* ]]; then
+      echo "Resource '$name' not found. Creating..."
+      kubectl apply -f "$manifest"
+      echo "Waiting 10s for RKE2 controller to sync new config..."
+      sleep 10
+      # When 'kubectl apply' is successful, we don't check helm-install job;
+      # RKE2 controller ensures the HelmChartConfig is used.
+      return 0
```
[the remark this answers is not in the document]
w13915984028: note: the real rke2-multus upgrade happens far after the upgrade process has started, until the first node is rebooted, then rke2 does the upgrade.  check `helm-install-rke2-multus` job at this stage, makes little sense

we just ensure the helmchartconfig object is created/patched

package/upgrade/upgrade_manifests.sh
```diff
@@ -5,12 +5,77 @@ UPGRADE_TMP_DIR="/tmp/upgrade"
 
 source $SCRIPT_DIR/lib.sh
 
+# The default RKE2 failurePolicy is `reinstall`.
+# Recreating the NAD CRD can cause all VM networking to be lost.
+patch_rke2_multus_config() {
+  echo "Check and patch rke2-multus failurePolicy to abort, to avoid a reinstall of the chart and recreation of the NAD CRD."
```
wheatdog, line 11: Would like to confirm: Although `failurePolicy` of `rke2-multus` is updated to `abort`, we can still rely on CrashLoopBackOff of pod running `upgrade_manifests.sh` to retry for resilience? Is this correct?

w13915984028: yes, if any of the `kubectl get/create/patch helmchartconfig` in this function failed except not-found, the `apply-manifest` will exit and be retried, until it ensured the target helmchart config is there

starbops: Is that so? I suppose a `helm rollback` is required to save us from such a situation (a `failed` release status) and thus continue with the `helm upgrade`.

starbops (approved):
LGTM, thank you. We need a clear document for users to know how to handle rke2-multus chart upgrade failures.

wheatdog (approved):

package/upgrade/upgrade_manifests.sh
```diff
@@ -5,12 +5,77 @@ UPGRADE_TMP_DIR="/tmp/upgrade"
...
+  # 1. Prepare Manifest
+  cat > "$manifest" <<EOF
+apiVersion: helm.cattle.io/v1
+kind: HelmChartConfig
+metadata:
+  name: $name
+  namespace: $namespace
+spec:
+  failurePolicy: abort
+EOF
```
ihcsim, line 28: Can we just `kubectl apply` this manifest regardless of whether the `HelmChartConfig` resource already exists or not, and regardless of what the existing value (if any) of the `failurePolicy` is? Essentially, just let `kubectl` do a 3-way merge to ensure the desired state.

package/upgrade/upgrade_manifests.sh
```diff
@@ -5,12 +5,77 @@ UPGRADE_TMP_DIR="/tmp/upgrade"
...
+  local name="rke2-multus"
+  local namespace="kube-system"
+  local manifest="$UPGRADE_TMP_DIR/rke2-multus-helmchartconfig.yaml"
+
+  mkdir -p "$UPGRADE_TMP_DIR"
+
+  # 1. Prepare Manifest
+  cat > "$manifest" <<EOF
+apiVersion: helm.cattle.io/v1
+kind: HelmChartConfig
```
ihcsim, line 22: Do we know what's the difference between specifying the `failurePolicy` via a `HelmChartConfig` vs. updating the existing in-cluster `HelmChart`, which also has the `failurePolicy` attribute?

```sh
$ k -n kube-system get helmchart rke2-multus -ojsonpath='{.spec.failurePolicy}'
reinstall%                                                                
```

From what I can see, certain custom changes made to the in-cluster `HelmChart` (like `failurePolicy`) will survive upgrade.

package/upgrade/upgrade_manifests.sh
```diff
@@ -5,12 +5,77 @@ UPGRADE_TMP_DIR="/tmp/upgrade"
...
+metadata:
+  name: $name
+  namespace: $namespace
+spec:
```
ihcsim, line 26: I wonder if Helm will leave the CRD alone if we use this `HelmChartConfig` to set `.Values.manifests.customResourceDefinition: false`? It means the CRD won't be upgraded, and possibly, also not deleted on chart upgrade failures?

package/upgrade/upgrade_manifests.sh
starbops: IIRC, the source of the in-cluster HelmChart resources is static manifests lying in each node's `/var/lib/rancher/rke2/server/manifests/` directory. Those manifest files will be overwritten whenever the RKE2 server restarts.
Customizing the HelmChart resources with their respective HelmChartConfig is the way to go.

package/upgrade/upgrade_manifests.sh
starbops: We can't reliably assume there will always be a `last-applied-configuration` annotation on the HelmChartConfig resource. I suppose a 3-way merge requires that annotation to do the computation?

package/upgrade/upgrade_manifests.sh
starbops: Doesn't that indicate the CRD would be removed upon upgrades?

package/upgrade/upgrade_manifests.sh
w13915984028: A short clarification:

(1) the apply_manifest tried to create a rke2-multus `helmchartconfig` object to guide rke2;  apply_manifest only cares about helmchartconfig crd object is created or patched successfully; any kubectl get/create/patch failure just triggers the apply_manifest pod to be replaced

(2) the rke2-server does rke2-charts upgrade per it's design; for any chart, if the `helmchartconfig` sets `failurePolicy` to abort, rke2 will keep the `failed` chart as is, doesn't forcely reinstall it;  the failed chart needs user manual operation like `helm rollback` to recover;  it gave us a chance to troubleshooting the root cause, hopefully, we get the RCA quickly and have a real fix to replace this workaround.

Harvester upgrade: only checks rke2-server version, does not check rke2-charts status, a failed rke2-chart won't block Harvester upgrade. Those two are loosely coupled.

package/upgrade/upgrade_manifests.sh
w13915984028: same concerns:

https://github.com/rancher/rke2-charts/blob/4efb3781aefbbb90bee5b154fdb6dd63e2049f4f/charts/rke2-multus/rke2-multus/v4.2.314/templates/customResourceDefinition.yaml#L14

any rke2-server could run `helm install/upgrade` upon it, helm might just remove the CRD directly

Another solution is to hack rke2-charts to add below annotation, also not easy:

(1) patch `/var/lib/rancher/rke2/server/manifests/rke2-multus.yaml` on each node, to set this annotation, it is in a base64+gunzip  `.spec.chartContent` field
(2) ensure helmchart object is updated as well (rke2 does it automatically)
(3) on the node upgrade stage, after rke2 files are copied, patch it, but not sure it could survive after node reboots

```
metadata:
  annotations:
    "helm.sh/resource-policy": keep
```

if this is ensured, even when rke2-server uninstall the rke2-multus chart, the NAD CRD is still kept

package/upgrade/upgrade_manifests.sh
ihcsim: It should work; `kubectl` might show a `resource is missing the kubectl.kubernetes.io/last-applied-configuration annotation` benign warning.

package/upgrade/upgrade_manifests.sh
ihcsim: It's possible to separate the NAD CRD from Helm's ownership using both the `manifests.customResourceDefinition: false` value and `"helm.sh/resource-policy": keep` annotation, so that if the chart upgrade failed due to _any_ reasons, then CRD and CRs don't get deleted. But it means that any future changes to the CRD will need to be handled separately. Setting the failure policy to `abort` seems to be the safest option.

If we don't want Helm to manage the CRD, this works for me when upgrading from `rke2-multus:v4.2.407` to `rke2-multus:v4.2.408`:

```sh
# install 4.2.407 k8s
$ helm upgrade --install rke2-multus ./v4.2.407

$ helm list
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                   APP VERSION
rke2-multus     default         1               2026-04-15 11:26:05.684210159 -0700 PDT deployed        rke2-multus-v4.2.407    4.2.4  

$ k get crd network-attachment-definitions.k8s.cni.cncf.io
NAME                                             CREATED AT
network-attachment-definitions.k8s.cni.cncf.io   2026-04-15T18:21:34Z

# annotate nad crd
$ k annotate crd network-attachment-definitions.k8s.cni.cncf.io helm.sh/resource-policy=keep
customresourcedefinition.apiextensions.k8s.io/network-attachment-definitions.k8s.cni.cncf.io annotated

# crd is still managed by helm
$ helm get manifest rke2-multus | grep -i "NetworkAttachmentDefinition"    

# upgrade to 4.2.408 with manifests.customResourceDefinition=false
$ helm upgrade --install rke2-multus ./v4.2.408 --set manifests.customResourceDefinition=false

# crd still exists
$ k get crd network-attachment-definitions.k8s.cni.cncf.io
NAME                                             CREATED AT
network-attachment-definitions.k8s.cni.cncf.io   2026-04-15T18:21:34Z

# crd no longer managed by helm
$ helm get manifest rke2-multus | grep -i "NetworkAttachmentDefinition"
# no output

# uninstall rke2-multus
$ helm uninstall rke2-multus

$  helm list
NAME    NAMESPACE       REVISION        UPDATED STATUS  CHART   APP VERSION

# crd still exists
$ k get crd network-attachment-definitions.k8s.cni.cncf.io
NAME                                             CREATED AT
network-attachment-definitions.k8s.cni.cncf.io   2026-04-15T18:21:34Z

# annotation still intact
$ k get crd network-attachment-definitions.k8s.cni.cncf.io -ojsonpath="{.metadata.annotations['helm\.sh/resource-policy']}"
keep%
```

starbops: This is interesting. So I guess the code path for handling the annotation between `helm uninstall` and `helm upgrade` is quite different. From my experience, the annotation (added to the live resource by the user) is not respected when uninstalling chart releases.

w13915984028:
@wheatdog @starbops @ihcsim 

As we are preparing for a rc6 to get this workaround for more tests, hopefully the root cause could be catched in the following days.

For those alternatives/optimizations upon the code, I will not update but keep current code as is, there is no functional errors for now and they are tested to work per expectation. thanks.

w13915984028:
@mergify backport v1.7

Issue #5681, opened 2024-04-25
[FEATURE] Allow particular ports access as default security policy


mingshuoqiu:
**Is your enhancement related to a problem? Please describe.**
There should be a way to blacklist external network or whitelist the Harvester node network only to protection from malicious scan and access.

**Describe the solution you'd like**
1. Set the default policy for incoming or forwarding traffic to DROP instead of ACCEPT.
2. Allow the particular ports required for communication between nodes, and HTTP(s), SSH for basic access.
3. Master node and Worker/Witness node should have different white list.

**Additional context**
Optional: allow user's application for security. Ex. check-mk

mingshuoqiu:
For Master Node:
```
iptables -P INPUT DROP
iptables -P FORWARD DROP

iptables -A INPUT -p tcp --dport 80 -m conntrack --ctstate NEW,ESTABLISHED -j ACCEPT
iptables -A INPUT -p tcp --dport 443 -m conntrack --ctstate NEW,ESTABLISHED -j ACCEPT
iptables -A INPUT -p tcp --dport 22 -m conntrack --ctstate NEW,ESTABLISHED -j ACCEPT

iptables -A INPUT -p udp --dport 8472 -j ACCEPT
iptables -A INPUT -p tcp -m multiport --dports 6443:6444 -j ACCEPT
iptables -A INPUT -p tcp --dport 9345 -j ACCEPT
iptables -A INPUT -p tcp -m multiport --dports 10248:10250 -j ACCEPT 
iptables -A INPUT -p tcp -m multiport --dports 10256:10260 -j ACCEPT
iptables -A INPUT -p tcp -m multiport --dports 2379:2382 -j ACCEPT

iptables -A INPUT -p tcp --dport 10010 -j ACCEPT
iptables -A INPUT -p tcp --dport 9091 -j ACCEPT
iptables -A INPUT -p tcp --dport 9099 -j ACCEPT
iptables -A INPUT -p tcp --dport 2112 -j ACCEPT

iptables -A INPUT -m conntrack --ctstate ESTABLISHED -j ACCEPT
```

For other Nodes:
```
iptables -P INPUT DROP
iptables -P FORWARD DROP

iptables -A INPUT -p tcp --dport 80 -m conntrack --ctstate NEW,ESTABLISHED -j ACCEPT
iptables -A INPUT -p tcp --dport 443 -m conntrack --ctstate NEW,ESTABLISHED -j ACCEPT
[...]

PR #10372, opened 2026-04-13
hep: add iptables for nodes

Yu-Jack:
#### Problem:
The customer observed that the default policy of iptables on the Harvester hosts don't drop by default. The default policy must be set to "drop" and only known and allowed traffic can pass.

#### Solution:
Although this's a HEP PR, but there won't by any code changes in harvester/harvester repo. Instead, we'll provide a comprehensive documentation to customers.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/5681

docs: https://github.com/harvester/docs/pull/1034

### DIFF
--- a/enhancements/20260318-node-iptables-policy.md
+++ b/enhancements/20260318-node-iptables-policy.md
@@ -1,0 +1,673 @@
+# Node-level iptables Policy Guide
+
+## Summary
+
+Harvester nodes are exposed to potential malicious scans and unauthorized access because there is currently no default inbound/forward traffic filtering policy. This HEP provides a documentation-only guide that describes the services and ports used on each Harvester node role, along with ready-to-use iptables rules that users can manually apply to harden their clusters.
+
+**No Harvester code changes are required.** Users apply the iptables rules directly on each node via SSH or CloudInit CRD.
+
+For background, harvester-installer [PR 1213](https://github.com/harvester/harvester-installer/pull/1213) applied iptables rules via a script.
+
+### Related Issues
+
+- https://github.com/harvester/harvester/issues/5681
+
+## Motivation
+
+### Goals
+
+- Document all services and their listening ports on each Harvester node role (control-plane, worker, witness).
+- Provide ready-to-use iptables rule sets that users can review, customize, and apply manually.
+- Ensure the documentation clearly explains which ports are required and why, so users can make informed decisions about their security posture.
+
+### Non-goals
+
+- Egress filtering (out of scope).
+- Target pod/vm-level traffic.
+- Modifying any Harvester source code, controllers, or CRDs.
+- Automated rule management or reconciliation via a controller.
+
+## Proposal
+
+### Proposed Solution
+
+#### 1. Documentation of Services and Ports
+
+Provide a comprehensive reference document that lists every service running on each Harvester node role, its listening port(s), protocol, scope (external vs. internal), and purpose. This enables users to understand the network surface of their cluster before applying any filtering rules.
+
+#### 2. Ready-to-Use iptables Rules
+
+Provide sample iptables commands for each node role (control-plane, worker, witness) that users can:
+
+1. **Review** — understand which ports are opened and why.
+2. **Customize** — add or remove rules to match their specific network topology and requirements.
+3. **Apply manually** — run the commands via SSH on each node.
+
+The rules use a custom chain (`HARVESTER_INPUT`) inserted into the `INPUT` hook, making them easy to manage without interfering with other iptables rules on the system.
+
+#### 3. No Default Rules Shipped
+
+No iptables rules are applied automatically on installation or upgrade. This ensures existing cluster traffic is never disrupted.
+
+#### 4. Rule Persistence via CloudInit CRD
+
+iptables rules are not persistent across reboots by default. The recommended approach for Harvester is to use the [`CloudInit` CRD](https://docs.harvesterhci.io/v1.7/advanced/cloudinitcrd/) to persist rules.
+
+Example `CloudInit` resource for control-plane nodes:
+
+```yaml
+apiVersion: node.harvesterhci.io/v1beta1
+kind: CloudInit
+metadata:
+  name: iptables-control-plane
+spec:
+  matchSelector:
+    node-role.kubernetes.io/control-plane: "true"
+    node-role.kubernetes.io/etcd: "true"
+  filename: 99-iptables-control-plane
+  contents: |
+    stages:
+      boot:
+        - name: "Apply iptables rules for control-plane node"
+          commands:
+            - iptables -D INPUT -i mgmt-br -j HARVESTER_INPUT 2>/dev/null || true
+            - iptables -F HARVESTER_INPUT 2>/dev/null || true
+            - iptables -X HARVESTER_INPUT 2>/dev/null || true
+            - iptables -N HARVESTER_INPUT
+            - iptables -I INPUT -i mgmt-br -j HARVESTER_INPUT
+            - iptables -A HARVESTER_INPUT -m conntrack --ctstate ESTABLISHED,RELATED -m comment --comment "Accept return traffic from established connections" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p icmp -m comment --comment "ICMP (ping, MTU discovery, etc.)" -j ACCEPT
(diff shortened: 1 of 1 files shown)


enhancements/20260318-node-iptables-policy.md
```diff
@@ -0,0 +1,489 @@
+# Node-level iptables Policy Guide
+
+## Summary
+
+Harvester nodes are exposed to potential malicious scans and unauthorized access because there is currently no default inbound/forward traffic filtering policy. This HEP provides a documentation-only guide that describes the services and ports used on each Harvester node role, along with ready-to-use iptables rules that users can manually apply to harden their clusters.
+
+**No Harvester code changes are required.** Users apply the iptables rules directly on each node via SSH or CloudInit CRD.
```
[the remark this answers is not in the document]
Yu-Jack: We won't consider `FORWARD`. It's irrelevant.

enhancements/20260318-node-iptables-policy.md
```diff
@@ -0,0 +1,673 @@
...
+            - iptables -I INPUT -i mgmt-br -j HARVESTER_INPUT
+            - iptables -A HARVESTER_INPUT -m conntrack --ctstate ESTABLISHED,RELATED -m comment --comment "Accept return traffic from established connections" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p icmp -m comment --comment "ICMP (ping, MTU discovery, etc.)" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p udp --sport 67 --dport 68 -m comment --comment "DHCP server reply" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p udp --sport 68 --dport 67 -m comment --comment "DHCP client request" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 22 -m comment --comment "SSH" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 80 -m comment --comment "Harvester UI HTTP" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 443 -m comment --comment "Harvester UI HTTPS" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 2112 -m comment --comment "RKE2 kube-vip Prometheus metrics" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 2379:2380 -m comment --comment "etcd client/peer" -j ACCEPT
```
m-ildefons, line 86: This will allow connections to etcd from anywhere.
Would it not be better to restrict connections to etcd to IP addresses, which originate within the cluster's subnet?

enhancements/20260318-node-iptables-policy.md
```diff
@@ -0,0 +1,673 @@
...
+            - iptables -A HARVESTER_INPUT -p tcp --dport 6443 -m comment --comment "Kubernetes API server" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 6641:6644 -m comment --comment "OVN NB/SB DB and JSON-RPC" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 8080 -m comment --comment "kube-ovn-webhook HTTP" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 8443 -m comment --comment "kube-ovn-webhook HTTPS" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 9091 -m comment --comment "calico-node metrics endpoint (Prometheus format)" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 9345 -m comment --comment "RKE2 supervisor API" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 9796 -m comment --comment "Prometheus node-exporter metrics" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 10250 -m comment --comment "kubelet API" -j ACCEPT
```
m-ildefons, line 94: What about the other Kubernetes ports, e.g. 10259 for the kube-scheduler?
https://kubernetes.io/docs/reference/networking/ports-and-protocols/

enhancements/20260318-node-iptables-policy.md
```diff
@@ -0,0 +1,673 @@
...
+            - iptables -A HARVESTER_INPUT -p tcp --dport 10661 -m comment --comment "kube-ovn-monitor metrics" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 10665 -m comment --comment "kube-ovn-daemon" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p tcp --dport 30000:32767 -m comment --comment "NodePort TCP range" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p udp --dport 30000:32767 -m comment --comment "NodePort UDP range" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p udp --dport 8472 -m comment --comment "VXLAN Flannel/Canal" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -p udp --dport 4789 -m comment --comment "VXLAN Kube-OVN" -j ACCEPT
+            - iptables -A HARVESTER_INPUT -m limit --limit 10/min -m comment --comment "Rate-limit logging of dropped packets" -j LOG --log-prefix HARVESTER_DROP_ --log-level 4
+            - iptables -A HARVESTER_INPUT -m comment --comment "Drop all other traffic" -j DROP
```
m-ildefons, line 102: This is a weird way to configure dropping packages.

The packages are processed on the `INPUT` chain. From there, we call the `HARVESTER_INPUT` chain. Instead of adding a rule to drop all packages at the end of the `HARVESTER_INPUT` chain, the solution should be to switch the default policy of the `INPUT` chain to `DROP`.
This would allow users to call additional custom chains from the `INPUT` chain and only drop packages if none of their custom chains accept them.

By having a rule to drop all traffic on the `HARVESTER_INPUT` chain we make this difficult, because:
1) the order of chains called from the `INPUT` chain matters
2) the user can't easily see what the behavior of the `INPUT` chain is without inspecting all custom chains called from the `INPUT` chain

Setting the default policy on the `INPUT` chain to `DROP` is better, because it shows clearly what the behavior of the `INPUT` chain is if no rules match.
It also makes the behavior independent of the order of any custom chains.

m-ildefons (commented):
In general, this HEP makes sense to me. There are a few points though:

1) Not restricting the source address ranges where possible makes the accepting rules less effective. E.g. the ports for etcd only need to allow connections from inside the cluster's own IP range.

2) The constructions of the chains itself is a bit off IMO. I'd prefer to work with a default policy of `DROP` on the `INPUT` chain instead of a drop-rule at the end of the `HARVESTER_INPUT` chain.

3) IMO it would be a good idea to break the `HARVESTER_INPUT` chain up into several smaller chains, each dealing with a specific purpose.
For example, we could have a `HARVESTER_SYSTEM` chain for all systemic services like SSH and a different `HARVESTER_MONITORING` chain with all rules for monitoring services.

enhancements/20260318-node-iptables-policy.md
Yu-Jack: It's bound at 127.0.0.1, so it's internal. We can skip it because we're handling with host network instead of others. You can check appendix in the HEP.

enhancements/20260318-node-iptables-policy.md
Yu-Jack: In general, this iptables is kind of a template, so users can add or change some rules based on their requirements. But, I'll mention about restricting source IP address on specific cases.

enhancements/20260318-node-iptables-policy.md
Yu-Jack: Good point. Let me try what we need to do when we use `default: DROP` policy.

ibrokethecloud (approved):

enhancements/20260318-node-iptables-policy.md
Yu-Jack: I'll go with `Default: ACCEPT` and `DROP` as the last rule. Users can just easily delete the `HARVESTER_INPUT` chain to restore when encountering any issues without any potential risks. Also, splitting it into different chains might be a bit complicated in this case. However, I think `Default: DROP` might be our ultimate way.

rrajendran17 (approved):
LGTM,Thanks

Issue #934, opened 2021-06-08
[EPIC] IPv6 support


ApolloDS:
**Is your feature request related to a problem? Please describe.**
When installing the ISO only IPv4 can be set, IPv6 is missing.

**Describe the solution you'd like**
IPv6 support

**Describe alternatives you've considered**
no alternatives.

**Additional context**
![grafik](https://user-images.githubusercontent.com/3502089/121159914-eaa95380-c84b-11eb-9807-1eafec34ea7f.png)

telmich:
I'm also interested in giving harvester a try, but I only operate IPv6 only networks (i.e. not IPv4 at all)

weihanglo:
Maybe related 

https://github.com/flannel-io/flannel/pull/1398
https://github.com/k3s-io/k3s/issues/284

ronnyaa:
> I'm also interested in giving harvester a try, but I only operate IPv6 only networks (i.e. not IPv4 at all)

nothing greenfield have ipv4 for sure. shocked to see that harvester that  is fairly new is not ipv6 primarily.

jhoelzel:
are there any updates here? Both flannel and k3s now support ipv6 dual-stack which should be perfect for most cases.

jhoelzel:
are there any updates here? Both flannel and k3s now support ipv6 dual-stack which should be perfect for most cases.

jhoelzel:
are there any updates here? Both flannel and k3s now support ipv6 dual-stack which should be perfect for most cases.

bk201:
IPv6 is disabled at OS level in v1.1.1 due to k3s reports different node IP after upgrading: https://github.com/harvester/harvester/issues/2962

itworksdoesntwork:
and how to enable it ?
we have plenty of ipv6 but as you guys can assume ipv4 is a little difficult to get and very expensive

irony:
Any news here?

joshuarestivo:
Bump.

innobead:
cc @mingshuoqiu @starbops

innobead:
@mingshuoqiu Please investigate this early, even though it's not planned for 1.7. We can implement it in phases, starting with IPv6 exclusive and then dual stack.

cc @harvester/network

derhornspieler:
[...]

PR #10379, opened 2026-04-14
hep: ipv6 support

martindekov:
Adding harvester enhancement proposal document for ipv6 support.

The general design is driven by parallel slice to ipv4 through the
harvester stack checking what's supported and trying to draw a line
and support same baseline functionalities for ipv6 with the mechanisms
of the protocol.

#### Problem:
Currently harvester supports ipv4 only

#### Solution:
Design at least baseline ipv6 support for further enhancement

#### Related Issue(s):
https://github.com/harvester/harvester/issues/934

#### Test plan:
N/A

Added in references section to the doc


enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
+# IPv6 Support
+
+This enhancement introduces first-class IPv6 support in Harvester host and VM networking for IPv6-only and dual-stack environments.
```
rrajendran17, line 3: IPv6 is disabled by default on Harvester host.We need to enable it during Harvester installation before mgmt interface configuration.
net.ipv6.conf.all.disable_ipv6 = 1(set it to 0)

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+- Configure RKE2 control-plane settings (node-ip, cluster-cidr, service-cidr, tls-san, stack-preference) for the selected IP family so the Kubernetes cluster operates correctly in IPv6 and dual-stack environments.
+- Enable IPv6 in the kube-ovn overlay that carries VM network traffic, so VM networks support IPv6 CIDR and gateway settings end-to-end.
+- Ensure the Harvester web UI, API, and VM VNC/Serial console proxy are reachable via the IPv6 management VIP. This requires the Ingress controller to bind on IPv6 and the console proxy to accept IPv6 connections, so that operators and users in IPv6-only environments can reach all management surfaces.
+- Support IPv6 addressing on the VM live migration network so VMs can be live-migrated between nodes in IPv6-only and dual-stack clusters. The migration network is a `HostNetworkConfig`-managed VLAN; IPv6 support there follows directly from the host network config changes in this HEP. KubeVirt selects the migration network via a `NetworkAttachmentDefinition` and binds the migration server on the interface address — both Multus and KubeVirt v1.7.0 support IPv6 for this path.
```
rrajendran17, line 73: We do already have a migration network supported.Why not extend the existing VM Migration network spec to include v6 range as well?
Also we have HostNetworkConfig CRD already for configuring additional vlan sub interfaces on Harvester hosts.https://docs.harvesterhci.io/v1.8/networking/hostnetworkconfig

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+
+### Non-goals
+
+- Replacing current CNI architecture in this HEP.
+- Implementing NAT64, DNS64, or other protocol translation services.
+- Automatic conversion of all existing IPv4-only user network definitions to IPv6.
+- Guest OS IPv6 configuration: injecting IPv6 addresses via Cloud-Init after NIC attachment, or displaying IPv6 addresses reported by the qemu-guest-agent in the UI. The VM NIC is attached with IPv6 capability at the network layer; address configuration inside the guest is guest-OS-specific and out of scope for this HEP.
+- DHCPv6 support. DHCPv6 requires server infrastructure that is not present in all IPv6 environments. IPv4 parity for dynamic addressing is satisfied by SLAAC for this HEP, since SLAAC is the infrastructure-free native mechanism. DHCPv6 can be revisited in a future HEP once SLAAC support is stable.
```
rrajendran17, line 82: Do we miss DNS or any other config if we use only SLAAC and not DHCPv6 in IPv6 only case?

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+- Storage network (SN) IPv6, dedicated VLAN in IPv6-only clusters. The Harvester `storage-network` setting requires an IPv4 CIDR as a documented prerequisite; the dedicated storage network feature is unavailable in IPv6-only environments. Longhorn replication traffic falls back to the pod network, which is covered by this HEP. This is tracked as follow-on work for a subsequent release.
```
rrajendran17: Why not extend the storage network settings to include ipv6 range, Whereabouts should support ipv6 range and could be specified as range in the storage network NAD

rrajendran17: I believe this could only be a testing only effort.

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+- VM management network (`mgmt`) IPv6 egress via NAT/Masquerade. VMs using the default `mgmt` network rely on NAT/Masquerade for external connectivity. Enabling IPv6 egress through this path requires changes to the `mgmt` bridge and masquerade rules that are independent from host and cluster networking scope. This is tracked as follow-on work for a subsequent release.
+- Backup target IPv6: configuring NFS or S3 backup targets reachable only via IPv6 addresses. Longhorn backup and restore traffic to external NFS/S3 endpoints involves storage and data protection paths that are outside the networking scope of this HEP. This is tracked as follow-on work for a subsequent release.
+
+## IP Family Modes
+
+Harvester management and host networking can be configured in three IP family modes. The mode is chosen at install time and determines how nodes, the management VIP, the Kubernetes control plane, and host interfaces are addressed.
+
+| | IPv4-only | Dual-stack | IPv6-only |
```
rrajendran17: Will this become a hard limit if user configures ipv4 or ipv6 only and later wants to shift to a new mode? Also other networks like VM,Migration and storage must be configurable for different modes based on underlying host network mode.

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+
+## Proposal
+
+### Scope
+
+This enhancement covers:
+
+- Installer and host network configuration accept and validate IPv6 management settings. Management interfaces use static addressing; stable addresses are required for cluster control-plane stability.
+- VM network route settings allow IPv6 CIDR and gateway where applicable.
+- Host network config APIs and controllers support IPv6 addresses for L3 interface configuration, covering both static and SLAAC modes.
```
rrajendran17, line 125: I believe you want to extend ipv6 support for this https://docs.harvesterhci.io/v1.8/networking/hostnetworkconfig

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+
+#### Story 8: Access web UI and VM console in IPv6-only environment
+
+TODO: detail
+
+### User Experience In Detail
+
+#### Installation
+
+- Existing install flow is extended with IPv6 fields where management settings are provided. Management interfaces always use static addressing, since stable predictable addresses are required for cluster control-plane communication. This is installer and host-networking work, not only Kubernetes API work; see [install.management_interface](https://docs.harvesterhci.io/v1.8/install/harvester-configuration#installmanagement_interface) for the current management interface configuration schema.
```
rrajendran17: You mean static address for mgmt node ip in case of ipv6?

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+- Validation includes IPv6 address/prefix format, gateway format, and family consistency.
+- For dual-stack, at least one reachable route is required for cluster bootstrap.
+
+#### Cluster Network and Host Network Configuration
+
+- HostNetworkConfig is extended to represent IPv6 L3 addresses in both static and SLAAC modes. For the practical meaning of cluster network, bridge, bond, VLAN, and host-side networking, see [Cluster Network](https://docs.harvesterhci.io/v1.8/networking/index) and the [Harvester Network Deep Dive](https://docs.harvesterhci.io/v1.8/networking/deep-dive).
+- Users can choose static or SLAAC for host VLAN sub-interfaces.
```
rrajendran17: We use only L2 for the VM Network. IPv6 is required only if user configures additional L3 interfaces on harvester host.You can make this statement more clearer.

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+- For SLAAC, the default gateway is not a config field — it is populated by the upstream router via Router Advertisement. The controller does not need to set a gateway explicitly.
+- Status conditions show per-node apply state and failure reason.
+
+#### VM Network Configuration
+
+- VM network route settings support IPv6 CIDR and gateway. This is largely extending current IPv4-oriented route and validation behavior rather than inventing a new network model; see [VM Network](https://docs.harvesterhci.io/v1.8/networking/harvester-network).
+- Connectivity checks include IPv6 path validation from all applicable nodes.
+
```
rrajendran17, line 221: May be we will require a field for user to select the addr family,so network controller can check connectivity from harvester host to gw ip on that vlan and also populate the CIDR and gateway settings for ipv6 at the backend.

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+Example VM Network route (dual-stack):
+
+```yaml
+route:
+  cidrV4: 192.168.100.0/24
+  gatewayV4: 192.168.100.1
+  cidrV6: 2001:db8:4000::/64
+  gatewayV6: 2001:db8:4000::1
+```
+
```
rrajendran17, line 349: I do not exactly get the purpose of new APIs and CRD.

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+- Extend API validation/webhooks for IPv6 fields in relevant resources.
+- Extend controllers and status reporting to handle IPv6 conditions.
+- Ensure upgrade workflows preserve and validate IPv6 settings.
+- Configure the Ingress controller (nginx) to listen on IPv6 addresses so the web UI, API, and VNC/Serial console proxy are reachable in IPv6-only and dual-stack deployments. Ensure the console proxy service binds on IPv6. Front-end URL construction for bracket notation is a UI concern and is covered in component 4.
+
+3. network-controller-harvester
+- Extend host network config reconciliation for static and SLAAC modes.
+- For SLAAC, configure NetworkManager profiles with `ipv6.method=auto` and do not set a gateway field; the gateway is derived from Router Advertisements.
+- Add IPv6 route and reachability checks for VM network route configuration.
+- Surface per-node condition details for IPv6 failures.
```
rrajendran17, line 408: We do not require a new CRD.We need to extend additional ipv6 support on HostNetworkConfig CRD(but this can also be done in later phases)

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+- Fix console and VNC URL construction to emit bracket notation (`wss://[<vip-v6>]/...`) when the management VIP is an IPv6 address. Without this, the browser rejects the URL before a connection is attempted.
+
+5. docs
+- Add user and operator guides for IPv6-only and dual-stack deployment.
+- Document that SLAAC requires a router sending Router Advertisements on the segment and that the gateway is not a user-configured field.
+- Add troubleshooting guidance for common IPv6 issues, including RA not received.
+- Document how to access the web UI and VM console in IPv6-only and dual-stack deployments, including the required browser URL format (`https://[<vip-v6>]`), TLS certificate SAN requirements, and expected console WebSocket URL form.
+
+6. kube-ovn
+- Enable IPv6 address families in the kube-ovn overlay network used for VM traffic.
```
rrajendran17: overlay VM network is independent of address family.We just need on the subnet configuration.

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+
+#### VM Network Tests
+
+1. Create VM network route with IPv6 CIDR/gateway.
+2. Verify connectivity checks and status updates.
+3. Validate dual-stack route settings.
+4. Validate webhook rejection of mismatched family input.
+5. L2 VM-to-VM IPv6 ping on the same node over a VLAN network: attach two VMs to the same VLAN and verify ICMPv6 echo reply.
+6. L2 VM-to-VM IPv6 ping across nodes over a VLAN network: schedule two VMs on different nodes and verify ICMPv6 echo reply, confirming the overlay underlay carries IPv6 frames correctly.
+
```
rrajendran17, line 496: Please Include tests for overlay VMs as well.

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
...
+1. Create HostNetworkConfig with static IPv6 per node.
+2. Create HostNetworkConfig with SLAAC mode; verify RA-derived address appears in status.
+3. Verify SLAAC gateway is populated from RA and not from a user config field.
+4. Reboot nodes and verify persistence of static and SLAAC configurations.
+5. Add and remove nodes; verify reconciliation and status reporting.
+6. Negative test: SLAAC mode with no RA on segment surfaces clear condition/error.
+
+#### VM Network Tests
+
```
rrajendran17, line 489: UI displays only v4 addresses, UI needs to display both V4/V6 from VM

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,483 @@
+# IPv6 Support
```
rrajendran17, line 1: You can also summarise the HEP into types of networks supported by Harvester currently with ipv4 which will be extended to v6 and the harvester component responsible.For example 
1.MGMT network physical connectivity(node IP/VIP)
2.Cluster/Pod Networking
3. VM Networking
4. KubeOVN Overlay Networking
5. VM Migration Network
6. Storage Network
7. Guest clusters 
8. HostNetworkConfig

enhancements/20260408-ipv6-support.md
martindekov: I had no idea, very good catch, thanks for the suggestion this was a true gap, will add soon to the goals to set it on nodes which are ipv6 or dual stack

enhancements/20260408-ipv6-support.md
martindekov: Good catch, SLAAC indeed does not provide DNS I didn't know, RA only sends out ip and gateway when SLAAC is used, but ipv6 has RFC 6106 to remove the dhcpv6 for that need and network manger that we use is 1.52.0 which supports configuring it we actually configure - RDNSS as workaround in the SLAAC case so basically this injects dns server as well next to the gateway and ip, we need to configure during install or through controller, we can dig deeper how, but the big picture is there no need for dhcpv6:

1. ipv6 already figured how to work around dns when SLAAC is used
2. our network manager supports configuring this
3. we need to check how to propagate this config and in what situations (installation, enablement etc)

Regarding any other config part there is `RFC 8981` as well, which forces rotation of ipv6 ips, which is not present for ipv4 so we need to make sure IPs in the ipv6 format stay consistent and I might add goal to have consistent IP for ivp6 format as well as ipv4 and through that goal set up network manager config to follow through. 

I think the two above are the most critical gas in the context of harvester but the good thing is they dont depend on dhcpv6 infrastructure and adoption for ipv6 MVP support

enhancements/20260408-ipv6-support.md
martindekov: Good catch as well, will add, thanks for the suggestion

enhancements/20260408-ipv6-support.md
martindekov: The hep covers this across goals in cope and design testing etc so resolving the concern, very good catch and suggestion!

enhancements/20260408-ipv6-support.md
martindekov: Tests added in the section, resolving

enhancements/20260408-ipv6-support.md
martindekov: Same as the slaac - covered across the relevant HEP sections, thanks for catching that!

enhancements/20260408-ipv6-support.md
martindekov: Updated the API section to start with explanation how changes are additive and intent is to not have any breaking changes. The design section generally was left off for after our meeting since design should be driven by the goals non/goals so that's why this was not changed from the first commit

enhancements/20260408-ipv6-support.md
martindekov: True, actually that was what I was working on, I found issues:
- https://github.com/longhorn/longhorn/issues/11869
- https://github.com/longhorn/longhorn/issues/12513

Which suggested longhorn on ipv6 is problematic so added in a hurry - not supported, but reading through the issues found out the problems were resolved on the version we are using so now this is as a goal and fixed.

enhancements/20260408-ipv6-support.md
martindekov: Stated in the goals it's immutable, good catch

enhancements/20260408-ipv6-support.md
martindekov: The API section explains that the changes are additive and nothing breaking is being added. Can add links if you request as well I used the structs directly to show examples

enhancements/20260408-ipv6-support.md
martindekov: Yes, added more details to clear the confusion of where we need static address:
```
the installer must record the management address before first boot to write node-ip into
the RKE2  config,derive etcd peer URLs, populate tls-san for the API server certificate, and
configure the kube-vip VIP — all of which require a known, stable address at install time. 
```

enhancements/20260408-ipv6-support.md
martindekov: Added the separation section and statement is clearer

enhancements/20260408-ipv6-support.md
martindekov: There are not new APIs and CRD, API section now states changes are additive and has inline comments, reusing structs, nothing new/nothing breaking

enhancements/20260408-ipv6-support.md
martindekov: Good suggesion added `Network Types Extended by This HEP` as part of the proposal

enhancements/20260408-ipv6-support.md
martindekov: Added optional `Family` field, however we can infer the family from the fact that we have ipv6 cidr so it can be redundant and the optional field revisited once we reach discussing the design

enhancements/20260408-ipv6-support.md
martindekov: No new CRD, the API section now explicitly states changes are additive

enhancements/20260408-ipv6-support.md
martindekov: I put this in non goals from the get go to reduce scope, we can revisit the decision if we think this is needed for MVP, there is already a lot of work and I want the HEP to deliver the base which we then can improve with other smaller heps and this suggestion is a good one

enhancements/20260408-ipv6-support.md
martindekov: Added in the kube-ovn section the detail for this:
```
The kube-ovn overlay transport (VXLAN/Geneve) is address-family agnostic and requires
no code changes for IPv6. The required work is configuration only: update kube-ovn
subnet spec.cidrBlock to accept an IPv6 or dual-stack CIDR value, and spec.gateway to
accept an IPv6 or comma-separated dual-stack gateway. These fields already support
IPv6 in kube-ovn v1.15.4 (see upstream evidence table).
```

enhancements/20260408-ipv6-support.md
martindekov: This is the non goal I refer originally was there:
```
Guest OS IPv6 configuration: injecting IPv6 addresses via Cloud-Init after NIC
attachment, or displaying IPv6 addresses reported by the qemu-guest-agent
in the UI. The VM NIC is attached with IPv6 capability at the network layer;
address configuration inside the guest is guest-OS-specific and out of scope
for this HEP.
```

enhancements/20260408-ipv6-support.md
rrajendran17: you can categorize the sub tasks of this EPIC based on above networking types involved.

enhancements/20260408-ipv6-support.md
rrajendran17: Reminder/place holder to add these to sub tasks and you can also reduce the scope of this HEP by organizing multiple HEPs for different sub tasks.I am sure we will need separate HEPs.

Once we have definitive subtasks(features under this EPIC), you can change the description,goals, non-goals of this HEP catering to this specific functionality.

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,572 @@
...
+
+#### Operational Workflows
+
+- Reboot, maintenance mode, and node replacement preserve and reconcile IPv6 configuration.
+- Documentation provides equivalent commands and examples for IPv4-only, IPv6-only, and dual-stack setups.
+
+### API changes
+
+All changes are additive at `network.harvesterhci.io/v1beta1`. No fields are removed. Existing `HostNetworkConfig` resources that use only `mode` and `ips` continue to work unchanged — the new fields are all optional. `slaac` is added as a new valid value for the existing `mode` field. The new `family` and `assignments` fields are optional and coexist with the existing `ips` map; during the transition period both are accepted, with `assignments` taking precedence when present. `VMNetworkRouteSpec` gains two optional fields (`cidrV6`, `gatewayV6`) alongside the existing IPv4 fields. No schema version bump is required.
+
```
rrajendran17, line 235: HostNetworkconfig is relatively new feature introduced in v1.8.0, so extending it for ipv6 will not be a hard requirement for now.We can add this to later phases of ipv6 support and work on other major ipv6 support on Harvester.

rrajendran17: IMO, the order of priority for different phases of ipv6 support,
1.MGMT network physical connectivity(node IP/VIP)
2.Cluster/Pod Networking
3. VM Networking
4. Guest clusters
5. VM Migration Network,Storage Network
6. KubeOVN Overlay Networking
7. HostNetworkConfig

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,572 @@
...
+// VMNetworkRouteSpec — additive delta only.
+// Existing cidrV4 and gatewayV4 fields are unchanged.
+// New optional fields: cidrV6, gatewayV6.
+type VMNetworkRouteSpec struct {
+    CIDRv4   string `json:"cidrV4,omitempty"`   // existing
+    Gateway4 string `json:"gatewayV4,omitempty"` // existing
+    CIDRv6   string `json:"cidrV6,omitempty"`   // new
+    Gateway6 string `json:"gatewayV6,omitempty"` // new
+}
+```
```
rrajendran17, line 292: The existing one for VM Network connectivity check is as below,we will need to extend this for v6.
```
type Layer3NetworkConf struct {
	Mode         Mode         `json:"mode,omitempty"`
	CIDR         string       `json:"cidr,omitempty"`
	Gateway      string       `json:"gateway,omitempty"`
	ServerIPAddr string       `json:"serverIPAddr,omitempty"`
	Connectivity Connectivity `json:"connectivity,omitempty"`
	Outdated     bool         `json:"outdated,omitempty"`
}
```

enhancements/20260408-ipv6-support.md
```diff
@@ -0,0 +1,572 @@
...
+    // --- existing fields (unchanged) ---
+    ClusterNetwork string                `json:"clusterNetwork"`
+    VlanID         uint16                `json:"vlanID"`
+    NodeSelector   *metav1.LabelSelector `json:"nodeSelector,omitempty"`
+    Mode           IPAssignmentMode      `json:"mode,omitempty"`         // existing
+    IPs            map[string]string     `json:"ips,omitempty"`          // existing IPv4 map
+    // --- new optional fields ---
+    Family         IPFamily              `json:"family,omitempty"`
+    Assignments    []HostIPAssignment    `json:"assignments,omitempty"`
+}
```
rrajendran17, line 281: The existing spec for HostNetworkConfig is as below,we can extend this for v6 by just including Hostv6IPs map[string]IPAddr, it will be empty if just v4 is supported.
```
type HostNetworkConfigSpec struct {
        // +optional

        // Optional: select target nodes by labels
        // If empty, applies to all nodes
        // +optional
        NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
        // Required, non-empty
        // +kubebuilder:validation:MinLength=1
        ClusterNetwork string `json:"clusterNetwork"`

        // +optional
        Underlay bool `json:"underlay"`

        // Required, must be 1-4094
        // +kubebuilder:validation:Minimum=1
        // +kubebuilder:validation:Maximum=4094
        VlanID uint16 `json:"vlanID"`

        // Required, non-empty, only 'static' or 'dhcp'
        // +kubebuilder:validation:Enum=static;dhcp
        // +kubebuilder:validation:MinLength=1
        Mode string `json:"mode"`

        // Optional map, validated if mode=static
        // +optional
        // +kubebuilder:validation:MaxProperties=500
        HostIPs map[string]IPAddr `json:"ips,omitempty"`
}
```

martindekov:
Thanks for the review @rrajendran17 we discussed some critical things like dual stack support for longhorn which would affect some of the implementation/component/design decisions so that's why I originally did not touch the sections much until we agree on the scope.

So longhorn dual stack support problematic - https://github.com/longhorn/longhorn/issues/10928 (need to figure if in dual stack scenarios longhorn can work as if it thinks its in ipv4 in theory it should)

But if this is a blocker longhorn is fundamental for harvester, so ipv6 only scenario can be the only viable approach but on customer side, most true infrastructures are dual stack due to the flexibility. So do we want working MVP or do we want something meaningful for the customer. Once we have that we can proceed.

martindekov:
I will keep this as is as it's good exploration of the whole IPV6 story and will close the PR. Will open separate PR to reflect the subtask of dual stack with ipv4 family first support for this epic:
https://github.com/harvester/harvester/issues/10637

Issue #10349, opened 2026-04-06
[BUG] Failed to upgrade from v1.7.1 to v1.8.0-rc5, stuck in third node `Pre-draining` state to wait vm migration start


TachunLin:
### Describe the Bug

With three nodes Harvester with two management role and one witness role.
And integrated with Rancher v2.14.0, have RKE2 guest cluster.

* When upgrading from v1.7.1 to v1.8.0-rc5. 

* The upgrade stuck on in third node `Pre-draining` state

<img width="1600" height="769" alt="Image" src="https://github.com/user-attachments/assets/bd464bde-0f0c-4ca7-8212-e90fdb0f90d0" />

*  Check the pre-drain pod log, it loops in `Waiting for VM live-migration or shutdown`

  ```
  hvst-upgrade-crtst-pre-drain-n3-hp50-kdszs Waiting for VM live-migration or shutdown...(count: 1)                                                          │
  │ hvst-upgrade-crtst-pre-drain-n3-hp50-kdszs + echo 'Waiting for VM live-migration or shutdown...(count: 1)' 
  ```
  
  https://github.com/user-attachments/assets/3eb67a00-cf65-4533-8381-d2690aa04e6e

* Check the VMIM events, it have failure
  ```
  Failed to set MigrationStat in VMI status. :the server rejected our request due to an error`
  ```
  
  ```
  Events:                                                                                                                                                    │
  │   Type     Reason              Age   From                       Message                                                                                    │
  │   ----     ------              ----  ----                       -------                                                                                    │
  │   Normal   SuccessfulCreate    30m   virtualmachine-controller  Created migration target pod virt-launcher-rke2-v1345-pool1-jclpl-65mld-2p968              │
  │   Normal   SuccessfulCreate    30m   virtualmachine-controller  Created attachment pod hp-volume-j4czl                                                     │
  │   Normal   SuccessfulHandOver  29m   virtualmachine-controller  Migration target pod is ready for preparation by virt-handler.                             │
  │   Warning  FailedHandOver      29m   virtualmachine-controller  Failed to set MigrationStat in VMI status. :the server rejected our request due to an erro │
  │ r in our request
  ```
  

  <img width="1585" height="793" alt="Image" src="https://github.com/user-attachments/assets/be43f265-1836-4c51-87f6-7df18359c1dc" />

### To Reproduce

Encounter this issue during the v1.8.0-rc5 release testing.

Please refer the following for the detailed test steps
[...]

PR #10386, opened 2026-04-15
feat: disable LiveMigrate strategy during upgrades

bk201:
#### Problem:

In v1.7, CPU and memory hot-plugging were enabled via LiveMigrate. However, this causes the KubeVirt controller to migrate all VMs simultaneously to update virt-launcher pods immediately after an operator upgrade.

#### Solution:

- Temporarily remove LiveMigrate from workloadUpdateMethods. VMs will instead migrate naturally during node upgrades, updating the virt-launcher image at that time.
- Re-enable the method post-upgrade to restore hot-plugging.

Note: CPU/Memory hot-plugging is unavailable during the upgrade period.

#### Related Issue(s):

- https://github.com/harvester/harvester/issues/10349

#### Test plan:

- Build the ISO and upgrade from the previous version.
- Create some VMs
- After the apply-manifest job, you can see the managed chart has this diff compare patches
  ``` 
      - apiVersion: kubevirt.io/v1
        jsonPointers:
        - /spec/workloadUpdateStrategy/workloadUpdateMethods
        kind: KubeVirt
        name: kubevirt
  ```
  and the kubevirt CR should have workloadUpdateMethod set to an empty list:

  ```
     workloadUpdateStrategy:                                                                                                                                                                 
       workloadUpdateMethods: []
  ```

- You should not see any `-workload-update-*` `virtualmachineinstancemigrations` after the upgrade.
- After the upgrade, the Harvester managed chart should not have the diff compare patches, and kubevirt CR should be set to
  ```
     workloadUpdateStrategy:                                                                                                                                                                 
       workloadUpdateMethods:
         - LiveMigrate
  ```

### DIFF
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -924,6 +924,39 @@ patch_longhorn_settings() {
   yq -e '.spec.values.longhorn' $target || echo "fail to get info .spec.values.longhorn"
 }
 
+patch_kubevirt_compare_patches() {
+  local target=$1
+  # patch diff compare patches to avoid complaining kubevirt resource is modified
+
+  # Check if the kubevirt comparePatches entry already exists
+  local EXIT_CODE=0
+  yq -e '.spec.diff.comparePatches[] | select(.apiVersion == "kubevirt.io/v1" and .kind == "KubeVirt" and .name == "kubevirt")' $target > /dev/null 2>&1 || EXIT_CODE=$?
+
+  if [ $EXIT_CODE != 0 ]; then
+    echo "Adding kubevirt comparePatches entry to $target"
+    # Ensure spec.diff.comparePatches exists as an array
+    yq -i '.spec.diff.comparePatches = .spec.diff.comparePatches // []' $target
+    # Add the kubevirt entry
+    yq -i '.spec.diff.comparePatches += [{"apiVersion": "kubevirt.io/v1", "kind": "KubeVirt", "name": "kubevirt", "jsonPointers": ["/spec/workloadUpdateStrategy/workloadUpdateMethods"]}]' $target
+  else
+    echo "kubevirt comparePatches entry already exists in $target, skip adding"
+  fi
+}
+
+disable_kubevirt_live_migrate() {
+  echo "Setting kubevirt workloadUpdateMethods to empty array"
+
+  local kubevirt_patch_file="kubevirt-workload-update-patch.yaml"
+  cat > ${kubevirt_patch_file} <<EOF
+spec:
+  workloadUpdateStrategy:
+    workloadUpdateMethods: []
+EOF
+
+  kubectl patch kubevirts.kubevirt.io kubevirt -n harvester-system --patch-file ${kubevirt_patch_file} --type merge
+  rm -f ${kubevirt_patch_file}
+}
+
 upgrade_managedchart_harvester_crd() {
   echo "Upgrading Harvester CRD managedchart fleet-local/harvester-crd"
 
@@ -973,6 +1006,8 @@ EOF
   fi
 
   patch_longhorn_settings ${hpatch}
+  patch_kubevirt_compare_patches ${hpatch}
+  disable_kubevirt_live_migrate
 
   update_managedchart_patch_file_annotations ${hpatch} $REPO_HARVESTER_CHART_VERSION
   update_managedchart_patch_file_unpause ${hpatch}
--- a/package/upgrade/upgrade_manifests.sh
+++ b/package/upgrade/upgrade_manifests.sh
@@ -926,6 +926,7 @@ patch_longhorn_settings() {
 
 patch_kubevirt_compare_patches() {
   local target=$1
+  local json_pointer="/spec/workloadUpdateStrategy/workloadUpdateMethods"
   # patch diff compare patches to avoid complaining kubevirt resource is modified
 
   # Check if the kubevirt comparePatches entry already exists
@@ -937,13 +938,23 @@ patch_kubevirt_compare_patches() {
     # Ensure spec.diff.comparePatches exists as an array
     yq -i '.spec.diff.comparePatches = .spec.diff.comparePatches // []' $target
     # Add the kubevirt entry
-    yq -i '.spec.diff.comparePatches += [{"apiVersion": "kubevirt.io/v1", "kind": "KubeVirt", "name": "kubevirt", "jsonPointers": ["/spec/workloadUpdateStrategy/workloadUpdateMethods"]}]' $target
+    JSON_POINTER=$json_pointer yq -i '.spec.diff.comparePatches += [{"apiVersion": "kubevirt.io/v1", "kind": "KubeVirt", "name": "kubevirt", "jsonPointers": [strenv(JSON_POINTER)]}]' $target
   else
-    echo "kubevirt comparePatches entry already exists in $target, skip adding"
+    # Entry exists, check if the specific jsonPointer is already in the array
+    local POINTER_EXISTS=0
+    JSON_POINTER=$json_pointer yq -e '.spec.diff.comparePatches[] | select(.apiVersion == "kubevirt.io/v1" and .kind == "KubeVirt" and .name == "kubevirt") | .jsonPointers[] | select(. == strenv(JSON_POINTER))' $target > /dev/null 2>&1 || POINTER_EXISTS=$?
+
+    if [ $POINTER_EXISTS != 0 ]; then
+      echo "kubevirt comparePatches entry exists but jsonPointer $json_pointer not found, adding it"
+      # Find the index of the kubevirt entry and append the jsonPointer
+      JSON_POINTER=$json_pointer yq -i '(.spec.diff.comparePatches[] | select(.apiVersion == "kubevirt.io/v1" and .kind == "KubeVirt" and .name == "kubevirt") | .jsonPointers) += [strenv(JSON_POINTER)]' $target
+    else
+      echo "kubevirt comparePatches entry with jsonPointer $json_pointer already exists in $target, skip adding"
+    fi
   fi
 }
 
-disable_kubevirt_live_migrate() {
+disable_kubevirt_live_migrate_workload_update() {
   echo "Setting kubevirt workloadUpdateMethods to empty array"
 
   local kubevirt_patch_file="kubevirt-workload-update-patch.yaml"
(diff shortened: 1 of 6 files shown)


package/upgrade/upgrade_manifests.sh
```diff
@@ -973,6 +1006,8 @@ EOF
   fi
 
   patch_longhorn_settings ${hpatch}
+  patch_kubevirt_compare_patches ${hpatch}
```
wheatdog, line 1022: `patch_kubevirt_compare_patches` patches `${hpatch}`, that is the new managedchart going to be applied. The next line `disable_kubevirt_live_migrate` updates the **current** KubeVirt CR. I think it would cause the fleet bundle to report drift right away due to the fact that the updated `comparePatches` is not in the current managedchart.

I'm wondering whether we should move L1009-L1010 before L997 and `patch_kubevirt_compare_patches` should patch the current managedchart instead. With L997, the new managedchart can also get the updated `comparePatches`.

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -942,6 +945,83 @@ func (h *upgradeHandler) addUpgradeLabelToDeschedulerAddons(upgrade *harvesterv1
...
+		logrus.Info("No comparePatches found in harvester managedchart, skip removing")
+		return nil
+	}
+
+	// Filter out the kubevirt comparePatch entry
+	var updatedComparePatches []fleet.ComparePatch
+	for _, patch := range managedChart.Spec.Diff.ComparePatches {
+		if patch.APIVersion == "kubevirt.io/v1" && patch.Kind == "KubeVirt" && patch.Name == "kubevirt" {
+			// Check if JsonPointers contains the specific path we're looking for
+			containsWorkloadUpdateMethods := slices.Contains(patch.JsonPointers, "/spec/workloadUpdateStrategy/workloadUpdateMethods")
```
wheatdog: nit: if `patch.JsonPointers` contains items more than `"/spec/workloadUpdateStrategy/workloadUpdateMethods"`, it would be dropped together.

package/upgrade/upgrade_manifests.sh
```diff
@@ -924,6 +924,39 @@ patch_longhorn_settings() {
   yq -e '.spec.values.longhorn' $target || echo "fail to get info .spec.values.longhorn"
 }
 
+patch_kubevirt_compare_patches() {
+  local target=$1
+  # patch diff compare patches to avoid complaining kubevirt resource is modified
+
+  # Check if the kubevirt comparePatches entry already exists
+  local EXIT_CODE=0
+  yq -e '.spec.diff.comparePatches[] | select(.apiVersion == "kubevirt.io/v1" and .kind == "KubeVirt" and .name == "kubevirt")' $target > /dev/null 2>&1 || EXIT_CODE=$?
```
wheatdog, line 934: should we also check if `"/spec/workloadUpdateStrategy/workloadUpdateMethods"` exists in `jsonPointers`?

wheatdog (commented):
This PR looks good to me in general, just got a few questions. Thanks!

package/upgrade/upgrade_manifests.sh
bk201: both are addressed

package/upgrade/upgrade_manifests.sh
bk201: Move the kubevirt patch function earlier.
In fact, because the chart is already paused, the drift situation will not happen.

package/upgrade/upgrade_manifests.sh
```diff
@@ -924,6 +924,50 @@ patch_longhorn_settings() {
   yq -e '.spec.values.longhorn' $target || echo "fail to get info .spec.values.longhorn"
 }
 
+patch_kubevirt_compare_patches() {
```
w13915984028, line 927: a minor optimization: skip single-node, but keep `as is` also fine

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -942,6 +945,98 @@ func (h *upgradeHandler) addUpgradeLabelToDeschedulerAddons(upgrade *harvesterv1
 	return nil
 }
 
+// enableKubevirtWorkloadLiveMigrate enables KubeVirt workload live migration
+// This brings back vCPU/memory hotplug features.
+// We disable the feature to avoid messy live migration during upgrade because virt-launcher pod images are outdated
+func (h *upgradeHandler) enableKubevirtWorkloadLiveMigrate(upgrade *harvesterv1.Upgrade) (*harvesterv1.Upgrade, error) {
```
w13915984028, line 950: a side-effect:

if user has manually changed things on managedchart, we will replace them unconditionally

pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -942,6 +945,98 @@ func (h *upgradeHandler) addUpgradeLabelToDeschedulerAddons(upgrade *harvesterv1
...
+		return nil
+	}
+
+	// Filter out the kubevirt comparePatch entry or update its jsonPointers
+	var updatedComparePatches []fleet.ComparePatch
+	removed := false
+	for _, patch := range managedChart.Spec.Diff.ComparePatches {
+		if patch.APIVersion == "kubevirt.io/v1" && patch.Kind == "KubeVirt" && patch.Name == "kubevirt" {
+			// Check if JsonPointers contains the specific path we're looking for
+			containsWorkloadUpdateMethods := slices.Contains(patch.JsonPointers, "/spec/workloadUpdateStrategy/workloadUpdateMethods")
```
w13915984028: the `slices.Contains` does an exact matching upon all emelents, no matter exists or not;  could remove slices.Contains, just run frrom 1004 to iterate anyway;  avoid doule looping when elements are in

w13915984028 (commented):
Thanks for the PR, some questions.

wheatdog (dismissed):
Thanks! I'm trying to test this locally but it seems my local build got some problem. I think it's not related to the change in this PR. I'll continue trying.

package/upgrade/upgrade_manifests.sh
```diff
@@ -924,6 +924,50 @@ patch_longhorn_settings() {
...
+  echo "Setting kubevirt workloadUpdateMethods to empty array"
+
+  local kubevirt_patch_file="kubevirt-workload-update-patch.yaml"
+  cat > ${kubevirt_patch_file} <<EOF
+spec:
+  workloadUpdateStrategy:
+    workloadUpdateMethods: []
+EOF
+
+  kubectl patch kubevirts.kubevirt.io kubevirt -n harvester-system --patch-file ${kubevirt_patch_file} --type merge
```
w13915984028, line 967: another point: `sleep 30s` might be required, as kubevirt takes time to apply the config change, let it finish the sync before moving to next step

pkg/controller/master/upgrade/upgrade_controller.go
bk201: Would you mind elaborating more and giving me an example? The function modifies an element in the `.Spec.Diff.ComparePatches` list, and the granularity is down to a single JSON pointer inside the element.

package/upgrade/upgrade_manifests.sh
bk201: The `WorkloadUpdateMethods` field is read on each reconciliation.

From the [source code](https://github.com/kubevirt/kubevirt/blob/50b9b4d0cf1caaf56147411a1e6765134e0e82b9/pkg/virt-controller/watch/workload-updater/workload-updater.go#L450-L462) we can  see:
- The workload-updater doesn't do anything if kubevirt is not in deployed (we wait until kubevirt is deployed in the upgrade system service phase)
- And it calls c.getUpdateData, the functions [gets the field every time](https://github.com/kubevirt/kubevirt/blob/50b9b4d0cf1caaf56147411a1e6765134e0e82b9/pkg/virt-controller/watch/workload-updater/workload-updater.go#L381)

Once we patch the value and the call returns, we can assure the fields are already updated. So I'd like to avoid more waiting in the upgrade job.

pkg/controller/master/upgrade/upgrade_controller.go
bk201: Good suggestion! Refactor in https://github.com/harvester/harvester/pull/10386/changes#diff-ad1688890111ebb67c1e9830de55577e7b83f1828d79f6f1f9609122cac236d3R1020-R1041 and add test cases.

pkg/controller/master/upgrade/upgrade_controller.go
w13915984028: The concern is:

a user could modify Harvester managedchart to customize  the `WorkloadLiveMigrate` (kubevirt.Spec.WorkloadUpdateStrategy.WorkloadUpdateMethod) or any other kubevirt related configs before upgrade, that's allowed and persisted.

when user sets this to a non-harvester-expected value, the controller overwrites it here.

this would not be a regular case, we could write it down on the document.

package/upgrade/upgrade_manifests.sh
w13915984028: My previous test upon other kubevirt related config (e.g. the overhead-ratio) was:

when kubevirt CRD object is changed, virt-operator (or other kubevirt component) will sync upon it, and if a new VM is created, it will be rejected by kubevirt webhook with errror like `kubevirt is not ready`, this gap is less than 1 min. Sure, I did not test all kubevirt object fields.

pkg/controller/master/upgrade/upgrade_controller.go
bk201: I got your point now. The WorkloadUpdateMethods should only be reverted to `LiveMigrate` to support CPU/memory hot plugging. The other possible value is `Evict`, which will shut down VMs when the user hotplugs CPU or memory to the VM. So I guess we only have this choice.

wheatdog:
I tested migrating v1.7.1 to https://github.com/harvester/harvester/pull/10386/changes/6ca25b5a3d0b5ba81061c2d9a6ac370bde2e2349 in a 2-node cluster, the result looks good.

- The KubeVirt comparePatches is removed from ManagedChart harvester after upgrading
- The fleet bundle is stable
- Only `kubevirt-evacuation-` VMIMs is observed during upgrading
- LiveMigrate is in workloadUpdateMethods of KubeVirt CR after upgrading

[supportbundle_84e0c53e-774a-4855-b1be-ffebf376f4c6_2026-04-16T08-28-20Z.zip](https://github.com/user-attachments/files/26777498/supportbundle_84e0c53e-774a-4855-b1be-ffebf376f4c6_2026-04-16T08-28-20Z.zip)

wheatdog (approved):

w13915984028 (approved):
LGTM, thanks.

starbops (approved):
LGTM, thank you.

bk201:
@mergifyio backport v1.8

bk201:
@mergifyio backport v1.7

Issue #10047, opened 2026-02-12
[BUG] Guest Cluster to Harvester Longhorn Storage via RWX / NFS connectivty loss during upgrade from 1.7.0 to 1.7.1


Martin-Weiss:
### Describe the Bug

We had harvester 1.7.0 running with a downstream RKE2 v1.34.2+rke2r1 guest cluster on top.
In the guest cluster we are using harvester-rwx class mapped to longhorn-rwx as documented and with Harvester 1.7.0 this was working well.

Now we upgaded harvester 1.7.0 -> harvester 1.7.1 and we have left the guest cluster with its workload running - so VMs were automatically life migrated as part of this upgrade.

After the upgrade of Harvester we had to realize that the NFS volumes used and mounted in the downstream clusters did NOT work anymore / got hung / stuck with these errors in dmesg:

```
[1831368.199155] [T29376] nfs: server 10.5.3.164 not responding, timed out
[1831369.351251] [T29566] nfs: server 10.5.3.164 not responding, timed out
[1831373.575187] [T20723] nfs: server 10.5.3.164 not responding, timed out
[1831378.219192] [T25433] nfs: server 10.5.3.164 not responding, timed out
[1831385.095145] [T25433] nfs: server 10.5.3.164 not responding, timed out
```

After some research we have seen that the sharemanager PODs seem to have got different IP addresses (sharemanager POD has this IP: 10.5.3.158) .

We can see this as well:
```
time="2026-02-12T08:09:12Z" level=info msg="GRPC call: /csi.v1.Node/NodeStageVolume request: {\"staging_target_path\":\"/var/lib/kubelet/plugins/kubernetes.io/csi/driver.harvesterhci.io/305962a306c70337e309257711159075b1703d9365be920dc9ed39bb351b86df/globalmount\",\"volume_capability\":{\"access_mode\":{\"mode\":\"MULTI_NODE_MULTI_WRITER\"},\"mount\":{\"fs_type\":\"ext4\"}},\"volume_context\":{\"hostStorageClass\":\"longhorn-rwx\",\"storage.kubernetes.io/csiProvisionerIdentity\":\"1768995775672-8081-driver.harvesterhci.io\"},\"volume_id\":\"pvc-45d0edde-990b-4717-91c8-4909f7192a3c\"}"
[...]

PR #10404, opened 2026-04-16
hep: hep for stable rwx mount point [skip ci]

WebberHuang1118:
#### Problem:
https://github.com/harvester/harvester/issues/10047

#### Solution:
HEP for https://github.com/harvester/harvester/issues/10047

#### Related Issue(s):
HEP for https://github.com/harvester/harvester/issues/10047

### DIFF
--- a/enhancements/20260427-rwx-static-share-manager-network.md
+++ b/enhancements/20260427-rwx-static-share-manager-network.md
@@ -1,0 +1,310 @@
+# RWX Static Share Manager Network Identity
+
+## Summary
+
+This design document describes the Harvester backend changes that provide a stable secondary network identity for Longhorn RWX Share Manager pods. The feature builds on the existing Harvester RWX network selection, derives a static NetworkAttachmentDefinition from the already selected source network, learns the first working IP and MAC from the running Share Manager pod, and then reuses that identity on later pod recreations.
+
+The implementation supports both RWX networking modes:
+
+- **Dedicated RWX network**: Share Manager traffic uses the NetworkAttachmentDefinition created for the RWX network setting.
+- **Shared storage network**: Share Manager traffic uses the existing storage-network NetworkAttachmentDefinition when the RWX setting is configured to share the storage network.
+
+Once a Share Manager identity is learned, Harvester records the source NAD, derived static NAD, IP address, MAC address, and interface name on the Longhorn `ShareManager`. The Harvester pod mutator then updates the pod's Multus network attachment request so the same interface, IP, and MAC are requested on subsequent pod starts.
+
+### Related Issues
+
+- [Issue #10047](https://github.com/harvester/harvester/issues/10047)
+
+## Motivation
+
+### Goals
+
+- **Provide stable L3 identity**: Preserve the Share Manager pod IP across pod recreation after the initial working identity is learned.
+- **Provide stable L2 identity**: Preserve the Share Manager pod MAC address together with the static IP.
+- **Reuse the existing RWX network source**: Apply static identity management to the NAD that Harvester already selects for RWX traffic, whether that is the dedicated RWX network or the shared storage network.
+- **Avoid manual per-volume network configuration**: Automatically opt in Longhorn Share Managers when a Harvester RWX network is configured.
+- **Protect the learned address from reuse**: Add the learned Share Manager IP to the source NAD `ipam.exclude` list so Whereabouts does not allocate it to another pod.
+- **Clean up derived state**: Remove stale derived NADs and source NAD excludes when the source network changes or the Share Manager is removed.
+
+## Proposal
+
+### User Stories
+
+**Story 1: As a Harvester administrator**
+I want RWX Share Manager pods to keep a stable network identity on the configured RWX network so that NFS traffic has a predictable endpoint.
+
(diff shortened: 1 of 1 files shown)


enhancements/20260427-rwx-static-share-manager-network.md
```diff
@@ -0,0 +1,290 @@
...
+
+- If `share-storage-network` is `true`, Harvester reads the storage-network setting and uses its `storage-network.settings.harvesterhci.io/net-attach-def` annotation.
+- If `share-storage-network` is `false`, Harvester uses the RWX setting's `rwx-network.settings.harvesterhci.io/net-attach-def` annotation.
+
+The selected source NAD must be marked as either a Harvester storage network or Harvester RWX network through the existing annotations:
+
+- `storage-network.settings.harvesterhci.io`
+- `rwx-network.settings.harvesterhci.io`
+
+If the RWX setting is empty, the source NAD cannot be found, or the source NAD is not one of the supported Harvester-managed network types, the Share Manager controller clears the static network state and records the reason in `rwx-harvester-csi.harvesterhci.io/static-ip-status`.
```
ihcsim, line 73: i thought if the rwx setting (i.e. `{"share-storage-network": "false", "network": {}}`) is empty, it means the k8s cluster network is used as the rwx network?

enhancements/20260427-rwx-static-share-manager-network.md
```diff
@@ -0,0 +1,290 @@
...
+- The MAC address from the interface status.
+- The prefix length from the source NAD IPAM range.
+
+The controller writes the learned identity back to the Longhorn `ShareManager`:
+
+```yaml
+rwx-harvester-csi.harvesterhci.io/nad: "harvester-system/rwx-network-762g9"
+rwx-harvester-csi.harvesterhci.io/static-nad: "harvester-system/rwx-network-762g9-static"
+rwx-harvester-csi.harvesterhci.io/ip: "172.16.0.22/24"
+rwx-harvester-csi.harvesterhci.io/mac: "02:00:00:00:00:01"
```
ihcsim, line 128: can we group this into one annotation where the value is a json object because these properties work together to serve one purpose?

enhancements/20260427-rwx-static-share-manager-network.md
WebberHuang1118: Yes, if the RWX network uses the pod network—which is not recommended for guest RWX volumes in the [RWX volumes support documentation](https://docs.harvesterhci.io/v1.8/rancher/csi-driver/#rwx-volumes-support)—then there is no way to apply this proposal’s mechanism to make the Share Manager Pod maintain a static IP.

enhancements/20260427-rwx-static-share-manager-network.md
WebberHuang1118: Thanks for the suggestion! I’d lean toward keeping them separate, though.

The shared `rwx-harvester-csi.harvesterhci.io/` prefix already indicates that they belong together. These annotations represent different meanings and logic in the controller flow, and some of them may be updated frequently within a short period.

Keeping them as separate annotations makes them easier to inspect via `kubectl -o jsonpath`, allows them to work with field selectors, and makes it possible to patch a single field without a read-modify-write operation on the whole blob.

enhancements/20260427-rwx-static-share-manager-network.md
```diff
@@ -0,0 +1,310 @@
...
+harvester-system/rwx-network-762g9-static
+```
+
+The derived NAD keeps the source CNI configuration but replaces `ipam` with:
+
+```json
+{"type":"static"}
+```
+
+Hash labels used by the source storage or RWX network controller are removed from the derived NAD so the static NAD is managed by the Share Manager network flow instead of being treated as a source network object.
```
ihcsim, line 108:
```suggestion
Hash labels used by the source network or RWX network controller are removed from the derived NAD so the static NAD is managed by the Share Manager network flow instead of being treated as a source network object.
```
i assume this is supposed to be "source network"

enhancements/20260427-rwx-static-share-manager-network.md
ihcsim: i think it needs to be documented that this fix won't work on pod network

enhancements/20260427-rwx-static-share-manager-network.md
```diff
@@ -0,0 +1,310 @@
...
+- `name` and `namespace` to the derived static NAD.
+- `ips` to the learned IP in CIDR form.
+- `mac` to the learned MAC.
+- `interface` to the requested interface.
+
+The mutator only updates an existing interface selection. If the requested interface is not already present in the pod's network annotation, the mutator leaves the pod unchanged.
+
+#### 6. Pod Removal Guard
+
+The controller watches Share Manager pod removals. If the RWX network is configured but the owning Share Manager has not yet completed the static identity annotations, Harvester requeues the Share Manager and returns an error from the remove handler. This gives the controller a chance to learn and persist the pod's current network identity before the pod disappears.
```
ihcsim, line 163: instead of relying on the "returns an error" behaviour, can we add a finalizer to the share manager to prevent its pod from disappearing?

PR #10405, opened 2026-04-16
fix the mac for share manger rwx interface

WebberHuang1118:
#### Problem:
https://github.com/harvester/harvester/issues/10047

#### Solution:
HEP https://github.com/harvester/harvester/pull/10404

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10047


WebberHuang1118:
Replaced by [harvester/harvester#10487](https://github.com/harvester/harvester/pull/10487) for better resilience.

Issue #10031, opened 2026-02-09
[TASK] Automate artifact dependency version management aligned with release maintenance 


pohanhuang:
### Task request

During PR #9980 review, the team discussed the need to pin specific versions of Go and golangci-lint for reproducible builds. However, pinning versions introduces maintenance overhead.

### Goal
Implement updatecli to automatically create PRs for version updates
- [base image](https://github.com/harvester/harvester/blob/master/Dockerfile.dapper#L4) go build version for dapper
- [go.mod ](https://github.com/harvester/harvester/blob/master/go.mod#L3)
- https://github.com/harvester/harvester/issues/10031#issuecomment-3870361134

### Example
similar to how SBOMscanner manages **Example:** https://github.com/kubewarden/sbomscanner/pull/815

### Progress Dashboard

- ✅ means done
- ❌ means not support
- empty means in progress 

| Repo | Support master | Support 1.7 | Support 1.8 |
|------|---------------|-------------|-------------|
| [harvester/harvester](https://github.com/harvester/harvester/pulls) | ✅ | ✅ | ✅ |
| [harvester/network-controller-harvester](https://github.com/harvester/network-controller-harvester/pulls) |✅ | ✅ | ✅ |
| [harvester/node-manager](https://github.com/harvester/node-manager/pulls) |✅ | ✅ | ✅ |
| [harvester/node-disk-manager](https://github.com/harvester/node-disk-manager/pulls) |✅ | ✅ | ✅ |
| [harvester/networkfs-manager](https://github.com/harvester/networkfs-manager/pulls) |✅ | ✅ | ✅ |
| [harvester/harvester-load-balancer](https://github.com/harvester/harvester-load-balancer/pulls) |✅ | ✅ | ✅ |
| [harvester/pcidevices](https://github.com/harvester/pcidevices/pulls) |✅ | ✅ | ✅ |
| [harvester/eventrouter](https://github.com/harvester/eventrouter/pulls) | | | |
| [harvester/vm-import-controller](https://github.com/harvester/vm-import-controller/pulls) |✅ | ✅ | ✅ |
| [harvester/vm-dhcp-controller](https://github.com/harvester/vm-dhcp-controller/pulls) |✅ | ✅ | ✅ |
| [harvester/seeder](https://github.com/harvester/seeder/pulls) |✅ | ✅ | ✅ |
| [harvester/csi-driver-lvm](https://github.com/harvester/csi-driver-lvm/pulls) |✅ | ❌ | ❌ |
[...]

PR #10436, opened 2026-04-21
hep: artifact dependency verion management [skip ci]

pohanhuang:
#### Problem:
#10031

#### Solution:
HEP for #10031

#### Related Issue(s):
HEP for #10031

### DIFF
--- a/enhancements/20260420-artifact-dependency-version-mgmt.md
+++ b/enhancements/20260420-artifact-dependency-version-mgmt.md
@@ -1,0 +1,229 @@
+# Automated Dependency Updates with Renovate + Updatecli
+
+## Summary
+
+Current Renovate setup doesn't handle complex dependency updates well. Developers still do a lot of manual work for:
+
+- K8s version bumps (need to coordinate multiple `k8s.io/*` modules), <https://github.com/harvester/harvester/issues/10110>.
+- Longhorn updates (spans Helm charts and Go modules), <https://github.com/harvester/harvester/pull/10193>.
+- Go version updates (need to touch `Dockerfile.dapper`, CI workflows, and `go.mod` together).
+
+This proposes using Renovate with better grouping rules + Updatecli for complex multi-file updates. Result: one PR per logical change instead of scattered manual updates.
+
+### Related Issues
+
+<https://github.com/harvester/harvester/issues/10031>.
+
+## Motivation
+
+### Goals
+
+- Group related dependencies into single PRs (all `k8s.io/*` modules together, not 15 separate PRs).
+- Handle multi-file updates that Renovate can't do (Go version in `Dockerfile.dapper` + CI + `go.mod`), e.g. <https://github.com/pohanhuang/harvester-installer/pull/28>.
+- Centralize update policies - write once, reuse across repos, <https://github.com/pohanhuang/harvester-updatecli-poc>.
+
+### Non-goals
+
+- Disable auto-merge (still need human review).
+
+## Introduction
+
+**Updatecli** automates multi-file updates across different formats (Go modules, Dockerfiles, YAML) in a single atomic change.
+
+**Policies** are reusable update configurations packaged as OCI artifacts (e.g., `ghcr.io/harvester/updatecli-policies:latest`). Write update logic once, reuse across repos.
+
+**Workflow**:
+
+1. Policy maintainers publish update logic to `ghcr.io/harvester/updatecli-policies/{policy}:v1.0.0`
+2. Each repo adds `updatecli/updatecli-compose.yaml` that references the policy and provides repo-specific values.
+3. GitHub Actions runs `updatecli compose apply` on schedule (e.g., weekly)
+4. Updatecli pulls the policy, detects version changes, and creates a PR updating all affected files together (e.g., `Dockerfile.dapper` + `go.mod` + CI workflows for Go bumps), e.g. <https://github.com/pohanhuang/harvester/pull/226>.
+
+## Proposal
+
+Use two tools to achieve artifact automation bumping:
+
+1. **Renovate** - for Go modules with better grouping.
+2. **Updatecli** - for multi-file or complex updates Renovate can't handle.
+
+### User Stories
+
+#### K8s Dependency Updates
+
+**Before**: K8s patch release → 15+ separate Renovate PRs for each `k8s.io/*` module → devs figure out how to manually group them (e.g., [issue #10110](https://github.com/harvester/harvester/issues/10110)) → manually verify compatibility → create grouped PR by hand.
+
+**After**: One PR with all `k8s.io/*` modules grouped, e.g. <https://github.com/pohanhuang/harvester/pull/157>.
+
+#### Go Version Updates
+
+**Before**: Update Go 1.25.7 → 1.25.8 requires touching:
+
+- `Dockerfile.dapper`
+- .github/workflows/*.yml
+- `go.mod`
+- Easy to miss a file or have inconsistent versions.
+
+**After**: Updatecli does it in one PR. All files updated together, example PR e.g. <https://github.com/pohanhuang/harvester/pull/226>.
+
+#### Longhorn Go Module Updates
+
+**Before**: Longhorn release → manually update multiple Go modules (e.g., [fork PR #209](https://github.com/pohanhuang/harvester/pull/209) touching `longhorn-manager`, `longhorn-instance-manager`, etc.) → manual version verification across modules.
+
+**After**: One PR with all Longhorn Go modules grouped and updated together.
+
+### User Experience In Detail
+
+**Setup** (one-time per repo):
+
+- Add `updatecli/` config files.
+- Update `renovate.json` to extend the centralized config.
+- Add GitHub Actions workflow for Updatecli.
+
+**Daily usage**:
+
+- Review PRs as they come in
+- Titles like "chore: update k8s deps to v1.33.8"
+- Approve/merge after CI passes
+
+No changes for contributors.
+
+### API changes
+
+None.
+
+## Design
+
+### Implementation Overview
+
+**Phase 1: Fix Renovate grouping**
+
+Rename `harvester/renovate` → `harvester/dependency-automation` to reflect broader scope (Renovate + Updatecli).
+
+Adopt grouping rules from [PR #5](https://github.com/harvester/renovate/pull/5), which groups related dependencies to reduce PR noise:
+
+- **Group 1**: k8s + kubevirt + rancher (these share compatibility constraints and update together)
+- **Group 2**: longhorn (independent update cycle, needs separate k8s compatibility check)
+
+This reduces 15+ separate PRs per k8s release → 2 PRs total.
+
+```json
+{
+  "packageRules": [
+    {
+      "description": "Group k8s, kubevirt and rancher packages",
+      "matchManagers": ["gomod"],
+      "matchPackagePatterns": [
+        "k8s\\.io/", 
+        "^kubevirt\\.io/",
+        "^github\\.com/rancher/rancher$"
+      ],
+      "groupName": "k8s + kubevirt + rancher dependencies",
+      "prBodyNotes": [
+        "⚠️ **Check longhorn compatibility before merging.**"
+      ]
+    },
+    {
+      "description": "Group longhorn packages",
+      "matchManagers": ["gomod"],
+      "matchPackagePatterns": ["^github\\.com/longhorn/"],
+      "groupName": "longhorn dependencies",
+      "prBodyNotes": [
+        "⚠️ **Check k8s compatibility before merging.**"
+      ]
+    }
+  ]
+}
+```
+
+**Phase 2: Create Updatecli policies**
+
+Develop policies in `harvester/dependency-automation/policies/` for multi-file updates Renovate can't handle:
+
+- **Go version**: Updates `Dockerfile.dapper` + `.github/workflows/*.yml` + `go.mod` together (see [example PR](https://github.com/pohanhuang/harvester/pull/226))
+
+**Phase 3: Package and distribute policies**
+
+Publish policies as OCI artifacts to `ghcr.io/harvester/dependency-automation/{policy}:{version}` using GitHub Actions in the `dependency-automation` repo. Repos pull the latest policies at runtime, so updates to policy logic propagate automatically.
+
+**Phase 4: Enable in target repos**
+
+Each repo adds minimal configuration:
+
+```yaml
+# .github/workflows/updatecli.yml
+name: Updatecli
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 8 * * *"
+
+jobs:
+  updatecli:
+    name: Bump golang version
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install Updatecli
+        uses: updatecli/updatecli-action@2cc8e6d8e356d76b0280cdd03766c36596a0614e # v3.0.0
+
+      - name: Diff (dry-run on PR)
+        if: github.ref != 'refs/heads/master'
+        env:
+          GITHUB_TOKEN: ${{ secrets.HARVESTER_TOKEN }}
+        run: |
+          updatecli compose diff --file updatecli/updatecli-compose.yaml
+
+      - name: Apply
+        if: github.ref == 'refs/heads/master'
+        env:
+          GITHUB_TOKEN: ${{ secrets.HARVESTER_TOKEN }} # Will replace this with Github Apps to autoroates the token.
+        run: |
+          updatecli compose apply --file updatecli/updatecli-compose.yaml
+```
+
+### Test plan
+
+**Unit testing**:
+
+- Run `updatecli diff` on each policy before publishing to verify it detects updates correctly.
+- Test against sample repos to ensure file patterns match, see current result, e.g. <https://github.com/pohanhuang/harvester/pull/226>.
+
+**Integration testing**:
+
+Pilot repos: `harvester/harvester-installer`.
+
+Success criteria:
+
+- PRs created with correct file updates (no missed files, no extra files)
+- No merge conflicts with concurrent development
+- CI passes on generated PRs
+- Manual review confirms version consistency across updated files
+
+**Rollout schedule**:
+
+- **Week 1-2**: Enable on pilot repos, monitor PR quality
+- **Week 3-4**: Fix issues, tune grouping rules and policies based on feedback
+- **Week 5+**: Gradual rollout to remaining repos (1-2 repos per week)
+
+### Upgrade strategy
+
+Dev infrastructure only—no end-user impact.
+
+**Migration per repo**:
+
+1. Add `updatecli/` directory with compose configuration
+2. Update `renovate.json` to extend `harvester/dependency-automation:default.json`
+3. Add `.github/workflows/updatecli.yml`
+4. Existing Renovate setup continues until new config is merged (no disruption)
+
+## Note
+
+**Future enhancements**:
+
+- Automated backports to release branches (e.g., `v1.8`, `v1.7.1`)
--- a/enhancements/20260420-artifact-dependency-version-mgmt.md
+++ b/enhancements/20260420-artifact-dependency-version-mgmt.md
@@ -1,14 +1,14 @@
-# Automated Dependency Updates with Renovate + Updatecli
+# Automated Dependency Updates with Renovate
 
 ## Summary
 
-Current Renovate setup doesn't handle complex dependency updates well. Developers still do a lot of manual work for:
+Current Renovate setup doesn't handle dependency updates well. Developers still do a lot of manual work for:
 
 - K8s version bumps (need to coordinate multiple `k8s.io/*` modules), <https://github.com/harvester/harvester/issues/10110>.
 - Longhorn updates (spans Helm charts and Go modules), <https://github.com/harvester/harvester/pull/10193>.
-- Go version updates (need to touch `Dockerfile.dapper`, CI workflows, and `go.mod` together).
+- Go version updates (need to touch Dockerfile and `go.mod` together).
 
-This proposes using Renovate with better grouping rules + Updatecli for complex multi-file updates. Result: one PR per logical change instead of scattered manual updates.
+This proposes using Renovate with proper grouping rules and automated merge policies. Result: one PR per logical change with auto-merge for patches/security fixes.
 
 ### Related Issues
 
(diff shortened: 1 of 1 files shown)


pohanhuang:
Hi @harvester/dev ,
This is the HEP for artifact dependency update, please help review it when you are available.

Open question:
@starbops for the renovate bumping, do you think we should group the kube-ovn related into bumping k8s group?

pohanhuang:
cc: @Vicente-Cheng @innobead @olblak

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,229 @@
...
+1. Policy maintainers publish update logic to `ghcr.io/harvester/updatecli-policies/{policy}:v1.0.0`
+2. Each repo adds `updatecli/updatecli-compose.yaml` that references the policy and provides repo-specific values.
+3. GitHub Actions runs `updatecli compose apply` on schedule (e.g., weekly)
+4. Updatecli pulls the policy, detects version changes, and creates a PR updating all affected files together (e.g., `Dockerfile.dapper` + `go.mod` + CI workflows for Go bumps), e.g. <https://github.com/pohanhuang/harvester/pull/226>.
+
+## Proposal
+
+Use two tools to achieve artifact automation bumping:
+
+1. **Renovate** - for Go modules with better grouping.
```
wheatdog: Wondering whether we have explored the capability of Updatecli for bumping Go modules, like https://www.updatecli.io/docs/plugins/autodiscovery/golang/. It would be great if we can depend on only one of these tools.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,229 @@
...
+
+      - name: Install Updatecli
+        uses: updatecli/updatecli-action@2cc8e6d8e356d76b0280cdd03766c36596a0614e # v3.0.0
+
+      - name: Diff (dry-run on PR)
+        if: github.ref != 'refs/heads/master'
+        env:
+          GITHUB_TOKEN: ${{ secrets.HARVESTER_TOKEN }}
+        run: |
+          updatecli compose diff --file updatecli/updatecli-compose.yaml
```
wheatdog: What's the typical content inside `updatecli/updatecli-compose.yaml`? Does it contain reference to the policy `ghcr.io/harvester/dependency-automation/{policy}:{version}`?

enhancements/20260420-artifact-dependency-version-mgmt.md
pohanhuang: Hi @wheatdog,

Thanks for the input! Using only Updatecli was actually my initial plan for this project, but there are a few considerations.

Updatecli does not handle general Go dependency updates in the same way Renovate does. For example, Harvester Go updates may include:

- Normal Go module bumps, such as `golang.org/x/text v0.33.0`
- Kubernetes-related dependencies, such as `k8s.io/*`
- Pseudo-version updates, such as `github.com/rancher/system-upgrade-controller/pkg/apis v0.0.0-20250306000150-b1a9781accab`

With Updatecli, we would need to explicitly define what to update, how to detect the new version, and how to apply the change for each case. This could become complex to maintain.

Renovate already handles these cases well, and we only need to configure which updates should be grouped into the same PR. Since Harvester already has an existing Renovate setup, reusing and extending it seems more practical.

I also discussed this with @olblak, the author of Updatecli. He mentioned that Updatecli is more suitable for specific and controlled update workflows, while Renovate is likely a better fit for general dependency updates like this.

enhancements/20260420-artifact-dependency-version-mgmt.md
pohanhuang: @wheatdog Great question, and yes!
updatecli-compose will contain multiple policy at once, for example https://github.com/pohanhuang/harvester-installer/blob/master/updatecli/updatecli-compose.yaml.

This file contains 2 policy
- `ghcr.io/pohanhuang/harvester/golang/version:0.1.3`, policy to update go version across go.mod, dockerfile, actions
- `ghcr.io/updatecli/policies/updatecli/autodiscovery:0.8.1`, policy will check if other policy is up-to-date, ensuring all of the repo sharing the same latest policy

pohanhuang: https://github.com/pohanhuang/harvester-installer/pull/29 shows the `updatecli/autodiscovery:0.8.1` will bump the `ghcr.io/pohanhuang/harvester/golang/version:0.1.3` when there is a newer one comes out!

enhancements/20260420-artifact-dependency-version-mgmt.md
wheatdog: Sounds good! Thanks for the information!

enhancements/20260420-artifact-dependency-version-mgmt.md
wheatdog: Cool, that's neat!

wheatdog (dismissed):

wheatdog:
LGTM, let's wait for other inputs. Thanks!

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,229 @@
...
+
+**Integration testing**:
+
+Pilot repos: `harvester/harvester-installer`.
+
+Success criteria:
+
+- PRs created with correct file updates (no missed files, no extra files)
+- No merge conflicts with concurrent development
+- CI passes on generated PRs
```
Yu-Jack: Does this include go build and go test? In my experience, I often run into failing to run go build an go test after bumping kubevirt and k8s packages. So, we need to ensure we'll run that.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,229 @@
...
+
+**Updatecli** automates multi-file updates across different formats (Go modules, Dockerfiles, YAML) in a single atomic change.
+
+**Policies** are reusable update configurations packaged as OCI artifacts (e.g., `ghcr.io/harvester/updatecli-policies:latest`). Write update logic once, reuse across repos.
+
+**Workflow**:
+
+1. Policy maintainers publish update logic to `ghcr.io/harvester/updatecli-policies/{policy}:v1.0.0`
+2. Each repo adds `updatecli/updatecli-compose.yaml` that references the policy and provides repo-specific values.
+3. GitHub Actions runs `updatecli compose apply` on schedule (e.g., weekly)
```
Yu-Jack: I remember we have a github action to test running Harvester every day https://github.com/harvester/harvester-installer/blob/master/.github/workflows/vagrant-install.yaml. We could set up the schedule before running that, then we could find out the problem early.

Also, do we need a conspicuous notification to let us know the bumping PR is ready, like assigning to whom?

enhancements/20260420-artifact-dependency-version-mgmt.md
pohanhuang: I think we should include that as well, However, I remember the CI will run `go test` is that correct?

enhancements/20260420-artifact-dependency-version-mgmt.md
pohanhuang: About the assignee, in every automate PR, I will try to assign harvester/dev, which will alert everyone to check about this bump.

What do you think?

enhancements/20260420-artifact-dependency-version-mgmt.md
Yu-Jack: Basically running `make ci` will cover the case.

enhancements/20260420-artifact-dependency-version-mgmt.md
Yu-Jack: Sounds good. After all we don't have a specific person handling bumping.

wheatdog:
Do we have plan to create PRs that bumping actions automatically?
https://github.com/search?q=org%3Aharvester+%22uses%3A%22+path%3A.github&type=code

pohanhuang:
@wheatdog thanks for the review, and yes we do support.
See the example PR here https://github.com/pohanhuang/harvester-installer/pull/30/changes

Any other good suggestion?

wheatdog:
Thanks @pohanhuang, the example PR to bump GH action looks good to me!

enhancements/20260420-artifact-dependency-version-mgmt.md
pohanhuang: I found the https://github.com/pohanhuang/harvester-installer/actions/runs/24827448201/workflow?pr=30 will trigger build iso which will end up trigger `make ci`.
Which I believe is for most of the harvester developer repo.

pohanhuang:
@harvester/dev can you help review this issue?
In the setup (renovate + updatecli) can take care of
- go modules update, gourping with k8s / longhorn
- action updates
- go version update in one PR.

appreciate for any good suggestions, thanks.

enhancements/20260420-artifact-dependency-version-mgmt.md
pohanhuang: @Yu-Jack , does this sounds good to you ?

m-ildefons:
To me this HEP seems to be a technical approach to an organizational issue.
IMO, at the core of the problem is that we have several components in different repositories, and the separation between the repositories does not reflect the architecture of the software anymore. As a matter of fact, in some cases the separation even creates circular dependencies.

For example, [the addons repository](https://github.com/harvester/addons) contains things that are used by the installer. As far as I can tell, they are _exclusively_ used by the installer. This repository should be merged into the [harvester installer repository](https://github.com/harvester/harvester-installer). The advantages of merging the repositories would be:
- version bumps become atomic single PRs without the need for fragile automation
- the installer repository would become more self-contained (i.e. building the installer would not longer require a clone of the addons repository)

There are several other similar examples within our organization.
I'm not advocating for a true mono-repo approach, but for a consolidation of repositories, in places where the dependencies are so tight, that the multi-repo approach is a burden more than a benefit.

Further examples are:
- network-controller-harvester
- node-manager
- node-disk-manager
- upgrade-toolkit
- go-common

All of these repositories contain components that are included in Harvester by default. This means Harvester depends on them. But these components also depend on Harvester (e.g. for CRD type definitions). This creates a circular dependency, which is always annoying.
Additionally, this makes dependency/version bumping unnecessarily hard, because each bump is split over several PRs.
While automation can help with the most tedious parts of the problem, it can not help with the general difficulty of vetting changes across repos.

My point is that we can simplify version and dependency management by a lot, if we reduce the number of separately maintained components, which each go into a Harvester build. This would best be done by merging the repositories into the main Harvester repo, where dependencies are then centrally managed and where the different components can depend upon each other with no additional maintenance cost.
Prime candidates for getting merged are components which have a circular or at least a very tight relationship with our main Harvester repo.

pohanhuang:
Hi @m-ildefons 
Thanks for the thoughtful feedback! The concern about repo consolidation is valid and worth discussing — however, I think it's a separate topic from what this HEP is trying to address.

This HEP focuses on dependency version management automation (keeping versions in sync across repos), not on the repo structure itself. The consolidation you're proposing would be a more fundamental architectural change that deserves its own dedicated discussion and issue.

I'd suggest opening a separate issue to track the repo consolidation idea, so it gets the attention it deserves without conflating it with the scope of this HEP.
Would that work for you?
cc: @Vicente-Cheng @innobead

m-ildefons:
Yes, I know this HEP shouldn't be conflated with repository structures. However both this HEP and a consolidation of our repositories tackle the same problem: version and dependency management.

I've brought up the topic of repo structure in this HEP discussion, because I think this HEP can not be discussed without discussion repo structure.

The amount of repos we deal with has a heavy influence on the amount of work going into dependency and version management. This means that the motivation for this HEP is also influenced by our repo structure.
In the case that consolidating our repos doesn't let us completely avoid the need for automatic dependency and version management, then the implementation of this such automation would still be heavily influenced by any change in our repo structure.
Therefore the implementation would best be done by first re-organizing the repos and then introducing the automation tools, because this lets us avoid re-doing the automation whenever we make changes to the repository structure.

Bottom line is that I'm proposing a repository consolidation _in addition_ to this HEP because it helps address the underlying problem.
I'm proposing it here and now, because I think this HEP would become much simpler and much easier to implement, once we no longer have a forest of inter-dependent repositories.

pohanhuang:
Hi @m-ildefons,

Thanks for elaborating — the point about repo structure is valid. That said, since the current automation design is largely repo-agnostic and invalid PRs can be handled manually, I'd prefer to ship the bumping policy first and treat consolidation as a follow-up track.

Happy to discuss if you feel strongly about the sequencing.
cc: @Vicente-Cheng @innobead

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+## Introduction
+
+**Updatecli** automates multi-file updates across different formats (Go modules, Dockerfiles, YAML) in a single atomic change.
+
+**Policies** are reusable update configurations packaged as OCI artifacts (e.g., `ghcr.io/harvester/updatecli-policies:latest`). Write update logic once, reuse across repos.
+
+**Workflow**:
+
+1. Policy maintainers publish update logic to `ghcr.io/harvester/updatecli-policies/{policy}:v1.0.0`
+2. Each repo adds `updatecli/updatecli-compose.yaml` that references the policy and provides repo-specific values.
```
ihcsim: Since all the repos are Go projects, won't the policy composition be similar, if not the same? Can the proposed `updatecli` GHA workflow pulls whatever policies it needs from GHCR? Then we also don't need to worry about the policies autodiscovery stuff. There will be less maintenance overhead for each repo, if the GHA workflow is the only thing needs to be added.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+### Related Issues
+
+<https://github.com/harvester/harvester/issues/10031>.
+
+## Motivation
+
+### Goals
+
+- Group related dependencies into single PRs (all `k8s.io/*` modules together, not 15 separate PRs).
+- Handle multi-file updates that Renovate can't do (Go version in `Dockerfile.dapper` + CI + `go.mod`), e.g. <https://github.com/pohanhuang/harvester-installer/pull/28>.
```
innobead: IIRC, this should be handled in renovate's go-version datasource, since it's a general capability for runtime version updates?

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+- Centralize update policies - write once, reuse across repos, <https://github.com/pohanhuang/harvester-updatecli-poc>.
```
innobead: Poan created a repo named harvester/renovate, but we can make it more general, independent from the tool name, like harvester/dep-update.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Non-goals
+
+- Disable auto-merge (still need human review).
```
innobead: The goals for dependency updates:
- For patch or security updates, auto-merge is required. (for any patch, should wait at least a period of time before merging, like 2 weeks))
- For minor updates, human review is required.
- Minor/patch/security updates apply to all active branches, including master/main trunk.
- Active release branches allow only patch/security updates.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Introduction
+
+**Updatecli** automates multi-file updates across different formats (Go modules, Dockerfiles, YAML) in a single atomic change.
```
innobead: If renovate can handle this, what else can we benefit from using updatecli? I would like to know how we can benefit from it if things can't be done in one tool. Just KISS. 

I know updatecli is used in security projects because it's very good at general file updates and supports a declarative pipeline design.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+
+**Policies** are reusable update configurations packaged as OCI artifacts (e.g., `ghcr.io/harvester/updatecli-policies:latest`). Write update logic once, reuse across repos.
```
innobead: Is this something like a renovate reference config that can be shared in other repos as well?

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+2. Each repo adds `updatecli/updatecli-compose.yaml` that references the policy and provides repo-specific values.
+3. GitHub Actions runs `updatecli compose apply` on schedule (e.g., weekly)
+4. Updatecli pulls the policy, detects version changes, and creates a PR updating all affected files together (e.g., `Dockerfile.dapper` + `go.mod` + CI workflows for Go bumps), e.g. <https://github.com/pohanhuang/harvester/pull/226>.
+
+## Proposal
+
+Use two tools to achieve artifact automation bumping:
+
+1. **Renovate** - for Go modules with better grouping.
+2. **Updatecli** - for multi-file or complex updates Renovate can't handle.
```
innobead: Suggest clarifying what complex updates we need.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+
+### User Stories
+
+#### K8s Dependency Updates
+
+**Before**: K8s patch release → 15+ separate Renovate PRs for each `k8s.io/*` module → devs figure out how to manually group them (e.g., [issue #10110](https://github.com/harvester/harvester/issues/10110)) → manually verify compatibility → create grouped PR by hand.
```
innobead: This was caused by an incomplete renovate setting. I believe it never worked before. 

Basically, what we should do:
- group them, automatically approve, and merge for some specific updates such as general patches and security fixes.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+
+**After**: One PR with all `k8s.io/*` modules grouped, e.g. <https://github.com/pohanhuang/harvester/pull/157>.
+
+#### Go Version Updates
+
+**Before**: Update Go 1.25.7 → 1.25.8 requires touching:
+
+- `Dockerfile.dapper`
```
innobead: dapper should be out here, since it should be removed. please check with Kiefer

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+- .github/workflows/*.yml
+- `go.mod`
+- Easy to miss a file or have inconsistent versions.
+
+**After**: Updatecli does it in one PR. All files updated together, example PR e.g. <https://github.com/pohanhuang/harvester/pull/226>.
```
innobead: As mentioned, all should be able to be handled in renovate from my understanding. Please correct my understanding.

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,230 @@
...
+
+#### Longhorn Go Module Updates
+
+**Before**: Longhorn release → manually update multiple Go modules (e.g., [fork PR #209](https://github.com/pohanhuang/harvester/pull/209) touching `longhorn-manager`, `longhorn-instance-manager`, etc.) → manual version verification across modules.
+
+**After**: One PR with all Longhorn Go modules grouped and updated together.
+
+### User Experience In Detail
+
+**Setup** (one-time per repo):
```
innobead, line 94: Should have a unified config (e.g., renovate). Per-repo config can base on it and customize its overridden settings. I believe updatecli might have similar way.

innobead (dismissed):
Overall this looks straightforward, but I have a few questions and suggestions that should be addressed first.

pohanhuang:
@ihcsim @innobead I've addressed all the feedback, could you please re-review?

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `osvVulnerabilityAlerts` enabled.
+- Group related dependencies such as Golang toolchain, SUSE BCI base images, and K8s plus Rancher dependencies.
+- Exclude risky or manually managed updates such as Longhorn, KubeVirt, Helm, and selected pinned dependencies.
+
+**Branch policy**:
+
+- `main` and `master`: normal shared policy applies.
+- Release branches matching `v*`: not scanned by default.
+- A repository can opt release branches in with `baseBranchPatterns`; when it does, only patch and security updates are allowed there.
```
ihcsim, line 63: or just manually backport security fixes to the active release branches as we see fit?

enhancements/20260420-artifact-dependency-version-mgmt.md
```diff
@@ -0,0 +1,174 @@
...
+
+**Update policies**:
+
+- **Security updates**: enabled by default and require manual review.
+- **Non-security patch/minor/major updates**: disabled by default.
+- **Major Go updates**: explicitly disabled.
+
+**Shared config defaults**:
+
+- `osvVulnerabilityAlerts` enabled.
```
ihcsim, line 55: I am not familiar with OSV, but ideally, the pull requests that Renovate open should directly address the security issues raised by dependabot at e.g., https://github.com/harvester/harvester/security/dependabot. IOW, as these pull requests are reviewed, merged, we should see the list of security issues go down.

ihcsim: What's the difference between `osvVulnerabilityAlerts` and `VulnerabilityAlerts`?

enhancements/20260420-artifact-dependency-version-mgmt.md
pohanhuang: Good point. For now, since there are fewer than 5 open security PRs on the release branches, I think we can handle them manually as we see fit — no need to automate backports at this stage.

We can revisit if the volume grows.

enhancements/20260420-artifact-dependency-version-mgmt.md
pohanhuang: `vulnerabilityAlerts` pulls from GitHub's Advisory Database — the same source as Dependabot. So when Renovate opens a PR via this channel and it gets merged, the corresponding Dependabot alert will automatically close. This is the primary mechanism to drive the Dependabot alert count down.

`osvVulnerabilityAlerts` is a separate, newer feature (currently experimental) that pulls from osv.dev. It may surface vulnerabilities not yet in GitHub's Advisory Database, so it acts as additional coverage. 

In the updated config, both are enabled — `vulnerabilityAlerts` to directly address the Dependabot alert list, and `osvVulnerabilityAlerts` as a supplementary signal for broader coverage.

pohanhuang: ### Before we merge the OpenTelemetry-Go
otel-dependency-not-merge

### After merge OpenTelemetry-Go
Screenshot 2026-06-25 at 12 03 27 PM

The current config does resolve the problem like this.

ihcsim (dismissed):

bk201 (approved):

ihcsim (approved):

Vicente-Cheng (approved):
thanks!

PR #10440, opened 2026-04-21
enhancement: Support Storage Migration for Restored VMs

WebberHuang1118:
#### Problem:
https://github.com/harvester/harvester/issues/10353

#### Solution:
Support Storage Migration for Restored VMs

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10353

#### Test plan:
* Update Harvester and Harvester Webhook with the images from this PR.
* Create a VM `vm` with a Longhorn v1 PVC as the root disk.
* Create a VM snapshot `vm-s` from the running VM `vm`.
* Create a VM backup `vm-b` from the running VM `vm`.
* Restore `vm-s` as a new VM `vm-s-r`, then trigger storage migration for `vm-s-r` using a Longhorn v2 volume as the target. The source and target volumes should be the same size.
* Restore `vm-b` as a new VM `vm-b-r`, then trigger storage migration for `vm-b-r` using a Longhorn v2 volume as the target. The source and target volumes should be the same size.
* Stop VM `vm`, then restore snapshot `vm-s` with the `replace existing` option.
* Trigger storage migration for VM `vm` using a Longhorn v2 volume as the target. The source and target volumes should be the same size.

### DIFF
--- a/pkg/controller/master/backup/restore.go
+++ b/pkg/controller/master/backup/restore.go
@@ -633,13 +633,81 @@ func (h *RestoreHandler) reconcileVM(
 	}
 	vmCpy.Annotations[lastRestoreAnnotation] = restoreID
 	vmCpy.Annotations[restoreNameAnnotation] = vmRestore.Name
-	delete(vmCpy.Annotations, util.AnnotationVolumeClaimTemplates)
+
+	// Build volumeClaimTemplates annotation from restored PVCs to enable storage migration
+	volumeClaimTemplatesStr, err := h.buildVolumeClaimTemplates(vmRestore)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build volumeClaimTemplates annotation: %w", err)
+	}
+	vmCpy.Annotations[util.AnnotationVolumeClaimTemplates] = volumeClaimTemplatesStr
 
 	if vm, err = h.vms.Update(vmCpy); err != nil {
 		return nil, err
 	}
 
 	return vm, nil
+}
+
+func sanitizeVolumeClaimTemplateSpec(pvc *corev1.PersistentVolumeClaim) corev1.PersistentVolumeClaimSpec {
+	spec := corev1.PersistentVolumeClaimSpec{
+		AccessModes:      append([]corev1.PersistentVolumeAccessMode(nil), pvc.Spec.AccessModes...),
+		StorageClassName: pvc.Spec.StorageClassName,
+		VolumeMode:       pvc.Spec.VolumeMode,
+	}
+
+	if len(pvc.Spec.Resources.Requests) > 0 || len(pvc.Spec.Resources.Limits) > 0 {
+		spec.Resources = corev1.VolumeResourceRequirements{}
+
+		if len(pvc.Spec.Resources.Requests) > 0 {
+			spec.Resources.Requests = pvc.Spec.Resources.Requests.DeepCopy()
+		}
+
+		if len(pvc.Spec.Resources.Limits) > 0 {
+			spec.Resources.Limits = pvc.Spec.Resources.Limits.DeepCopy()
+		}
+	}
+
+	return spec
+}
+
+func buildVolumeClaimTemplateAnnotations(pvc *corev1.PersistentVolumeClaim) map[string]string {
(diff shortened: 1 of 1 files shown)


pkg/controller/master/backup/restore.go
```diff
@@ -642,6 +648,81 @@ func (h *RestoreHandler) reconcileVM(
...
+
+func buildVolumeClaimTemplateAnnotations(pvc *corev1.PersistentVolumeClaim) map[string]string {
+	if len(pvc.Annotations) == 0 {
+		return nil
+	}
+
+	annotations := map[string]string{}
+	for key, value := range pvc.Annotations {
+		lowerKey := strings.ToLower(key)
+		if strings.Contains(lowerKey, "imageid") || strings.HasSuffix(lowerKey, "/imageid") {
```
brandboat: I'm wondering if the lowercase check is necessary here, if not, perhaps we can simplify the method to below
```
func buildVolumeClaimTemplateAnnotations(pvc *corev1.PersistentVolumeClaim) map[string]string {
    if val, ok := pvc.Annotations[util.AnnotationImageID]; ok {
        return map[string]string{util.AnnotationImageID: val}
    }
    return nil
}
```

pkg/controller/master/backup/restore.go
```diff
@@ -642,6 +648,81 @@ func (h *RestoreHandler) reconcileVM(
...
+}
+
+// buildVolumeClaimTemplates builds the volumeClaimTemplates annotation from restored PVCs
+// This is needed to enable storage migration and other PVC management features for restored VMs
+func (h *RestoreHandler) buildVolumeClaimTemplates(vmRestore *harvesterv1.VirtualMachineRestore) (string, error) {
+	entries := make([]util.VolumeClaimTemplateEntry, 0, len(vmRestore.Status.VolumeRestores))
+
+	for _, vr := range vmRestore.Status.VolumeRestores {
+		// Get the actual PVC to include latest annotations like imageID
+		pvc, err := h.pvcCache.Get(vmRestore.Namespace, vr.PersistentVolumeClaim.ObjectMeta.Name)
```
brandboat, line 687: Do we actually need to fetch the PVC from the cache again here? Since `reconcileVM` triggers after `reconcileVolumeRestores`, the PVCs should already be created. Furthermore, it seems all the information required to compose the VolumeClaimTemplateEntry is already available within VolumeRestores.

My main concern is the race condition: there’s no guarantee the cache will populate immediately after we create the PVC in reconcileVolumeRestores. 
```go
// in reconcileVolumeRestores
if err = h.createRestoredPVC(vmRestore, volumeBackup, volumeRestore); err != nil {
```
Because we return an error on line 702 if it's not found, updateStatusError will trigger a false alarm by emitting an error event and the not found error status.
While this isn't breaking, since the controller will eventually reconcile and succeed. But it might create unnecessary noise in the logs and status events.

brandboat (dismissed):
LGTM, left a few non-blocking comments.
Thanks for the patch!

pkg/controller/master/backup/restore.go
WebberHuang1118: Yes, good idea, thanks for the feedback

pkg/controller/master/backup/restore.go
WebberHuang1118: Yes. Since storage migration depends on the correctness of this annotation, IMO, it would be safer to check the PVC existence more carefully. Otherwise, it may cause issues during storage migration, and the root cause would not be obvious to users.

As you mentioned, even if the cache cannot find the PVC, it will return an error here and trigger another reconcile round, so it should eventually reach the expected state.

BTW, as discussed with @Vicente-Cheng before, if I understand correctly, the cache should be able to find the recreated resource immediately.

brandboat (approved):

pkg/controller/master/backup/restore.go
Vicente-Cheng: Yeah, the cache would be empty if the controller just started. In this case, it should be safe.

Vicente-Cheng (approved):

Vicente-Cheng:
@mergifyio backport v1.8

Issue #10448, opened 2026-04-22
[TASK] Remove non-orgs users when creating backport issue


Yu-Jack:
### Task request

Currently, we add non-orgs users to assignees when creating the corresponding backport issue, which causes the fail.

### Additional context

`Error: -22 00:50:54 [ERROR] 422 {"message": "Validation Failed", "errors": [{"message": "assignees xxxxx cannot be assigned to this issue", "value": ["xxxxx"], "resource": "Issue", "field": "assignees", "code": "invalid"}], "documentation_url": "https://docs.github.com/rest/issues/issues#create-an-issue", "status": "422"}`

Yu-Jack:
Fixed.

PR #10449, opened 2026-04-22
ci: remove non-org users when creating backport issue

Yu-Jack:
#### Solution:
Use `has_in_collaborators` to detect if the user belongs to org or not.

#### Related Issue(s):

https://github.com/harvester/harvester/issues/10448


brandboat (approved):
LGTM, thanks!

PR #10477, opened 2026-04-27
build: replace Dapper with Docker BuildKit multi-stage build system

bk201:
#### Problem:

We need to remove Dapper.

#### Solution:

Replace the Dapper-based build system with a single multi-stage Dockerfile driven by the Makefile via `docker build --target <stage>`. Each build target gets a dedicated stage with BuildKit cache mounts for Go module and build caches.

- Rename Dockerfile.dapper and move to Dockerfile.builder; add multi-stage Dockerfile with stages for build, validate, validate-ci, test, test-integration, generate-manifest, generate-openapi, prepare-addons, build-iso, and bundle
- Rewrite Makefile with inline docker build calls, a DOCKER_BUILD shorthand variable, and CI-aware color banners via BANNER
- Add scripts/mk-build-iso for DinD ISO builds with named-container artifact extraction via docker cp
- Add scripts/prepare-addons to clone and cache the addons repo, busting the build cache via the remote branch SHA
- Remove Dapper orchestration scripts (arm, ci, default, entry, generate, help, release) — replaced by Makefile targets

#### Related Issue(s):

https://github.com/harvester/harvester/issues/10478

#### Test plan:

The goal of this PR is to migrate away from Dapper while ensuring a seamless transition for both developers and CI pipelines.

* **No Interface Changes**: All existing `make <target>` commands function exactly as they did under Dapper.
* **Workflow Compatibility**: Environment variables (e.g., `REPO`, `PUSH`) remain unchanged to ensure zero disruption to CI/CD workflows and local development environments.
* **Minimal Script Alteration**: Script logic has been preserved as much as possible to minimize regression risk. While there are several areas identified for further cleanup, those will be addressed in future refactoring PRs to keep this migration focused and safe.

* **Note** Because we don't pass the `.git` dir into the build container anymore, we need a corresponding change in the installer: https://github.com/harvester/harvester-installer/pull/1274 for the `build-iso` target. You can test the codes:

```
# build images and push
# repo must be somewhere you can push, in the form: `172.17.0.1:5000/bk201z`
PUSH=true REPO=<your repo> make package-all


# build iso
HARVESTER_INSTALLER_REPO=http://192.168.2.22:3000/kiefer/harvester-installer.git HARVESTER_INSTALLER_REF=wip-buildx-worktree make build-iso
```


ibrokethecloud (dismissed):
lgtm.

Makefile
```diff
@@ -1,29 +1,177 @@
...
+# ---- Validate CI (dirty check after go generate + go mod tidy) ----
+validate-ci: builder-image
+	$(BANNER)
+	$(DOCKER_BUILD) --target validate-ci
+
+
+# ---- Test ----
+test: builder-image
+	$(BANNER)
+	$(DOCKER_BUILD) $(if $(CODECOV_TOKEN),--secret id=codecov_token$(comma)env=CODECOV_TOKEN --no-cache-filter=test) --target test
```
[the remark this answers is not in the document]
bk201: Fixed.

scripts/prepare-addons
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# pull addon repo to a docker image and generate addons manifests
+# addon source and manifests are copied to dist/prepare-addons
+set -e
+
+TOP_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
+
+REMOTE_URL="https://github.com/harvester/addons.git"
+ADDONS_BRANCH="main"
```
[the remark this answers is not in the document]
bk201: Fixed in the commit: [Pass add-ons repo default branch from Makefile](https://github.com/harvester/harvester/pull/10477/changes/edb7d5494f5934ca3861ab8a2c5f126dae46cea3)

Dockerfile
```diff
@@ -0,0 +1,97 @@
...
+# ---- prepare-addons ----
+FROM builder AS prepare-addons
+ARG ADDONS_BRANCH=main
+
+# re-pull when remote sha changed
+ARG REMOTE_SHA=unknown
+
+RUN mkdir -p /dist/prepare-addons
+# clone addons repo
+RUN git clone --branch ${ADDONS_BRANCH} --single-branch --depth 1 \
```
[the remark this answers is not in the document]
bk201: fixed

wheatdog (dismissed):
I followed the test plan to build iso in a git worktree folder and it looks good. Thanks!

Makefile
```diff
@@ -1,29 +1,186 @@
...
+	    -f $(ROOT)/Dockerfile.builder \
+	    -t $(MK_BUILDER_IMAGE) \
+	    $(ROOT)
+
+
+# ---- Pre-generate version env for container builds (no .git needed inside Docker) ----
+# Also handles git worktree checkouts where .git is a pointer file to an external directory.
+gen-version-env:
+	$(BANNER)
+	@bash $(ROOT)/scripts/version > /dev/null
```
ihcsim, line 72: the redirect will send all the echo outputs to the void

Makefile
```diff
@@ -1,29 +1,186 @@
...
+# ---- Test ----
+test: builder-image gen-version-env
+	$(BANNER)
+	$(DOCKER_BUILD) $(if $(CODECOV_TOKEN),--secret id=codecov_token_$(MK_REPO_ID)$(comma)env=CODECOV_TOKEN --no-cache-filter=test) --target test
+
+
+# ---- Test integration ----
+test-integration: builder-image gen-version-env
+	$(BANNER)
+	$(DOCKER_BUILD) --target test-integration -t harvester-test-integration:$(MK_REPO_ID)
```
ihcsim: Can we omit the build step and run `docker run` with `$(MK_BUILDER_IMAGE)` and pass in the workdir with `-w`?

Makefile
```diff
@@ -1,29 +1,186 @@
...
+	docker run --rm --privileged --network host \
+	    -v /var/run/docker.sock:/var/run/docker.sock \
+		-v harvester-test-integration-go-cache-${MK_REPO_ID}:/go/src/github.com/harvester/harvester/.cache/go-build \
```
ihcsim: how is `harvester-test-integration-go-cache-${MK_REPO_ID}` created?

bk201: It's created by Docker automatically when running the command for the first time.

Makefile
bk201: I can't find how to do this in a one-shot "docker run" command; the building step is inevitable. One possible way is via Docker Compose, but I guess we don't want to introduce another layer here.

ihcsim: Hmm.. I thought in this case `harvester-test-integration:$(MK_REPO_ID)` is just `$(MK_BUILDER_IMAGE)` with the workdir specified. Will this work?
```sh
$	docker run --rm --privileged --network host \
	    -v /var/run/docker.sock:/var/run/docker.sock \
		-v harvester-test-integration-go-cache-${MK_REPO_ID}:/go/src/github.com/harvester/harvester/.cache/go-build \
        -w /go/src/github.com/harvester/harvester
	    $(MK_BUILDER_IMAGE) \
	    ./scripts/test-integration
```

bk201: In the Dockerfile, we create a base stage to copy sources. If we use the builder image here, we'll miss the `COPY . .` part.
```
ARG MK_BUILDER_IMAGE
FROM ${MK_BUILDER_IMAGE} AS builder

FROM builder AS base
WORKDIR /go/src/github.com/harvester/harvester

# to exclude some files, add them in .dockerignore
COPY . .
```

My idea is we keep the `test-integration` target in the Dockerfile, and we can fine-tune the sources that need to be in the target when needed.

Makefile
bk201: Sourcing the `version` file in various scripts prints out some outputs. So I reduce noise here. Do you want me to print the output? There is no harm at all.

ihcsim: right, initially, i was thinking of the error messages, but then just notice that they are redirected to stderr, so we are good here - thanks.

Makefile
ihcsim: Right, I missed the `COPY . .`, which can also be done with ~`-v $(pwd):$(pwd)`~ `-v $(pwd):$(workdir)`. It's also ok to keep it as-is; just thought we might want to keep the Dockerfile smaller and less complex as brought up in slack.

docs/build-system.md
```diff
@@ -0,0 +1,284 @@
+# Build System
```
ihcsim: Feels like an obscure location for this kind of content. How about moving the build system architecture content into the ["Build & Development" section](https://github.com/harvester/harvester/wiki#build--development) of the wiki, and the "Quick Start" to the DEVELOPER_GUIDE.md?

bk201: Certainly, I'll move it to a wiki page.

bk201: Move the doc to https://github.com/harvester/harvester/wiki/Build-System

Vicente-Cheng (dismissed):
lgtm, thanks for it!

Makefile
```diff
@@ -1,29 +1,175 @@
...
+generate-openapi: gen-version-env
+	$(BANNER)
+	$(DOCKER_BUILD) --target generate-openapi-output --output type=local,dest=$(ROOT)
+
+
+# ---- Cache addons repo and generate addons manifests ---
+prepare-addons:
+	$(BANNER)
+	$(ROOT)/scripts/prepare-addons
+	# $(DOCKER_BUILD) --target prepare-addons-output --output type=local,dest=.
```
wheatdog: We could remove this stale comment-out command

Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+    ./scripts/generate-openapi
+
+FROM scratch AS generate-openapi-output
+COPY --from=generate-openapi /go/src/github.com/harvester/harvester/api/openapi-spec/swagger.json /api/openapi-spec/swagger.json
+COPY --from=generate-openapi /go/src/github.com/harvester/harvester/scripts/known-api-rule-violations.txt /scripts/known-api-rule-violations.txt
+
+
+# ---- prepare-addons ----
+FROM builder AS prepare-addons
+ARG ADDONS_BRANCH=main
```
wheatdog: We could omit the default ARG value to prevent confusion since there is already the default fallback in `scripts/prepare-addons`

Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+
+# re-pull when remote sha changed
+ARG REMOTE_SHA=unknown
```
wheatdog: ditto, we could omit the default ARG value

wheatdog: Nice mechanism for caching!

Makefile
```diff
@@ -1,29 +1,175 @@
...
+	@rm -f $(ROOT)/package/harvester $(ROOT)/package/harvester-webhook
+	@rm -f $(ROOT)/package/upgrade/upgrade-helper
+	@rm -rf $(ROOT)/package/upgrade/addons
+	@rm -rf $(ROOT)/dist/prepare-addons
+	@rm -rf $(ROOT)/dist/artifacts $(ROOT)/dist/harvester-cluster-repo
+
+clean-all: clean
+	$(BANNER)
+	@docker rmi -f $(MK_ADDONS_IMAGE) || true
+	@docker rmi -f harvester-iso-builder:$(MK_REPO_ID) harvester-test-integration:$(MK_REPO_ID) || true
```
wheatdog: How about using variable `MK_ISO_BUILDER_IMAGE` to replace `harvester-iso-builder:$(MK_REPO_ID)`

We could also have a dedicate variable for `harvester-test-integration:$(MK_REPO_ID)` given its multiple referencing.

Dockerfile
```diff
@@ -0,0 +1,199 @@
+FROM quay.io/costoolkit/releases-teal:grub2-live-0.0.4-2 AS grub2-mbr
+FROM quay.io/costoolkit/releases-teal:grub2-efi-image-live-0.0.4-2 AS grub2-efi
+
+FROM golang:1.25.7-bookworm AS builder
+
+ARG CONTAINER_WORKDIR=/go/src/github.com/harvester/harvester
+
+ARG MK_HOST_ARCH
+ENV ARCH=$MK_HOST_ARCH
```
wheatdog, line 9: Future enhancement idea: for cross-compiling (e.g. setting `DOCKER_DEFAULT_PLATFORM=linux/amd64` from an Apple Silicon laptop), we should refer to `TARGETARCH` ([ref](https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope)) instead of `MK_HOST_ARCH`, such that `DOCKER_DEFAULT_PLATFORM` is properly honored. 

Likewise, we also need to change this line in `scripts/version`
```
    ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
```

I think we can handle this in the future PRs.

wheatdog (dismissed):
I think we could merge this PR. Just leave a few comments for nits and future enhancement ideas. Thanks for this great improvement on the DX!

Dockerfile
bk201: Fixed. Thanks!

Dockerfile
bk201: Fixed.

Makefile
bk201: Fixed.

Vicente-Cheng (approved):

wheatdog (approved):

Issue #10422, opened 2026-04-17
[ENHANCEMENT] Prevent started VM after cloning or snapshotting


forbesguthrie:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

Prevent a VM from automatically starting up immediately after being created from Snapshots or Backups.
and 
Add a selection item called "Automatic startup setting immediately after creation" as an execution option for cloning.

innobead:
@WebberHuang1118 @Vicente-Cheng is this behavior already available? 

cc @harvester/storage

WebberHuang1118:
> [@WebberHuang1118](https://github.com/WebberHuang1118) [@Vicente-Cheng](https://github.com/Vicente-Cheng) is this behavior already available?
> 
> cc [@harvester/storage](https://github.com/orgs/harvester/teams/storage)

Harvester will automatically start the VM after the snapshot or backup is restored.

noahgildersleeve:
Testing with Test plan from PR

<details>
<summary>Test plan</summary>

> #### Test plan:
> _**Prereqs**_
> 
> * Create a VM named `test01`.
> * Setup a NFS or S3 backup target. You can run a S3 bakup target easily by running
> 
> ```
> $ docker run -p 9000:9000 -p 9001:9001 quay.io/minio/minio server /data --console-address ":9001"
> ```
> 
[...]

Issue #4098, opened 2023-06-16
[ENHANCEMENT] Allow user to customize snapshot/backup deadline


connorkuehl:
**Is your enhancement related to a problem? Please describe.**

Since bcacb1b0145dd1ee15db1f438a52ae091c70c78c: during a snapshot or backup, Harvester will check if the VM is running the QEMU Guest Agent, and if it is, it will instruct the guest agent to freeze the guest file systems and flush pending I/O. That PR uses a very conservative timeout value of 1 second, which is probably fine for CPU bound VMs but most likely not sufficient for a VM running a ton of I/O.

As a result, Harvester can support perfectly consistent snapshots for some VMs but not all VMs. Ideally, Harvester can be positioned to provide consistent snapshots for all workloads.

**Describe the solution you'd like**

There is no 1 perfect timeout that Harvester can apply to every VM snapshot/backup. The end user is best-equipped to choose how long Harvester should freeze their VM's filesystems.

Harvester could follow KubeVirt's lead and allow a user to specify a deadline after which the guest file systems are unfrozen and the snapshot is marked as fail. 

When the user goes to take a VM Snapshot or Backup, they could:

* Specify a deadline, overriding the default
* Use Harvester's default deadline (maybe 5 minutes or so)
* Choose not to have a deadline (snapshot takes as long as it takes and the FS is frozen the entire time)

> When snapshotting a running vm the controller will check for qemu guest agent in the vm. If the agent exists it will freeze the vm filesystems before taking the snapshot and unfreeze after the snapshot. It is recommended to take online snapshots with the guest agent for a better snapshot, if not present a best effort snapshot will be taken.
> [ ... ]
> The vmSnapshot has a default deadline of 5 minutes. If the vmSnapshot has not succeessfully completed before the deadline, it will be marked as Failed. The VM will be unfrozen and the created snapshot content will be cleaned up if necessary. The vmSnapshot object will remain in Failed state until deleted by the user.

https://kubevirt.io/user-guide/operations/snapshot_restore_api/#snapshot-a-virtualmachine

**Describe alternatives you've considered**

**Additional context**

Vicente-Cheng:
Since @votdev is working on another backup/restore related configuration #10422, @votdev could you handle this one together?

WebberHuang1118:
[...]

PR #10480, opened 2026-04-27
Prevent VM start after restore from backup or snapshot

votdev:
#### Problem:
VMs restored from backup or snapshot are always automatically started.

#### Solution:
This PR introduces a new `haltAfterRestore` field to the `VirtualMachineRestoreSpec` and the `RestoreInput`, which allows users to prevent a VM from automatically starting after it's been restored from a backup or created from a snapshot/clone.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10422

#### Test plan:
***Prereqs***
- Create a VM named `test01`.
- Setup a NFS or S3 backup target. You can run a S3 bakup target easily by running
```
$ docker run -p 9000:9000 -p 9001:9001 quay.io/minio/minio server /data --console-address ":9001"
```
You need to access the UI by `http://127.0.0.1:9001/login` and create the "data" bucket that is used in these tests.
grafik
- Create a backup of `test01` and name it `back01`.
grafik
- Create a snapshot of `test01` and name it `snap01`.
grafik

***Case 1***
- Apply the following manifest `restore-backup.yaml`:
```
$ cat <<EOF | k apply -f -
kind: VirtualMachineRestore
metadata:
  name: restore-back01-test02
  namespace: default
spec:
  newVM: true
  target:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: test02
  virtualMachineBackupName: back01
  haltAfterRestore: true
EOF
```
- There should be a VM named `test02` which is in halted mode.
grafik

***Case 2***
- Apply the following manifest `restore-snapshot.yaml`:
```
$ cat <<EOF | k apply -f -
kind: VirtualMachineRestore
metadata:
  name: restore-snap01-test03
  namespace: default
spec:
  newVM: true
  target:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: test03
  virtualMachineBackupName: snap01
  haltAfterRestore: true
EOF
```
- There should be a VM named `test03` which is in halted mode.
grafik

***Case 3***
- Delete VM `test02`.
- Apply the following manifest `restore-backup.yaml`:
```
$ cat <<EOF | k apply -f -
kind: VirtualMachineRestore
metadata:
  name: restore-back01-test02
  namespace: default
spec:
  newVM: true
  target:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: test02
  virtualMachineBackupName: back01
  haltAfterRestore: false
EOF
```
- There should be a VM named `test02` which is running.

***Case 4***
- Delete VM `test03`.
- Apply the following manifest `restore-snapshot.yaml`:
```
$ cat <<EOF | k apply -f -
kind: VirtualMachineRestore
metadata:
  name: restore-snap01-test03
  namespace: default
spec:
  newVM: true
  target:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: test03
  virtualMachineBackupName: snap01
  haltAfterRestore: false
EOF
```
- There should be a VM named `test03` which is running.

***Case 5***
- Delete VM `test02`.
- Apply the following manifest `restore-backup.yaml`:
```
$ cat <<EOF | k apply -f -
kind: VirtualMachineRestore
metadata:
  name: restore-back01-test02
  namespace: default
spec:
  newVM: true
  target:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: test02
  virtualMachineBackupName: back01
EOF
```
- There should be a VM named `test02` which is running.

***Case 6***
- Delete VM `test03`.
- Apply the following manifest `restore-snapshot.yaml`:
```
$ cat <<EOF | k apply -f -
kind: VirtualMachineRestore
metadata:
  name: restore-snap01-test03
  namespace: default
spec:
  newVM: true
  target:
    apiGroup: kubevirt.io
    kind: VirtualMachine
    name: test03
  virtualMachineBackupName: snap01
EOF
```
- There should be a VM named `test03` which is running.

### DIFF
--- a/pkg/api/vm/handler.go
+++ b/pkg/api/vm/handler.go
@@ -876,7 +876,18 @@ func (h *vmActionHandler) restoreBackup(vmName, vmNamespace string, input Restor
 	if _, err := h.backupCache.Get(vmNamespace, input.BackupName); err != nil {
 		return err
 	}
-	apiGroup := kubevirtv1.SchemeGroupVersion.Group
+
+	restore := buildVirtualMachineRestore(vmName, vmNamespace, input)
+
+	_, err := h.restoreClient.Create(restore)
+	if err != nil {
+		return fmt.Errorf("failed to create restore, error: %v", err)
+	}
+
+	return nil
+}
+
+func buildVirtualMachineRestore(vmName, vmNamespace string, input RestoreInput) *harvesterv1.VirtualMachineRestore {
 	restore := &harvesterv1.VirtualMachineRestore{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      input.Name,
@@ -884,7 +895,7 @@ func (h *vmActionHandler) restoreBackup(vmName, vmNamespace string, input Restor
 		},
 		Spec: harvesterv1.VirtualMachineRestoreSpec{
 			Target: corev1.TypedLocalObjectReference{
-				APIGroup: &apiGroup,
+				APIGroup: ptr.To(kubevirtv1.SchemeGroupVersion.Group),
 				Kind:     kubevirtv1.VirtualMachineGroupVersionKind.Kind,
 				Name:     vmName,
 			},
@@ -893,12 +904,16 @@ func (h *vmActionHandler) restoreBackup(vmName, vmNamespace string, input Restor
 			NewVM:                         false,
 		},
 	}
-	_, err := h.restoreClient.Create(restore)
-	if err != nil {
-		return fmt.Errorf("failed to create restore, error: %s", err.Error())
+
+	if input.KeepMacAddress {
+		restore.Spec.KeepMacAddress = true
 	}
 
-	return nil
+	if input.HaltAfterRestore {
+		restore.Spec.HaltAfterRestore = true
+	}
+
+	return restore
 }
 
 func (h *vmActionHandler) checkBackupTargetConfigured() error {
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -9231,6 +9231,9 @@
           "deletionPolicy": {
             "type": "string"
           },
+          "haltAfterRestore": {
+            "type": "boolean"
+          },
           "keepMacAddress": {
             "type": "boolean"
           },
--- a/deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachinerestores.yaml
+++ b/deploy/charts/harvester-crd/templates/harvesterhci.io_virtualmachinerestores.yaml
@@ -62,6 +62,11 @@ spec:
                 description: DeletionPolicy defines that to do with resources when
                   VirtualMachineRestore is deleted
                 type: string
+              haltAfterRestore:
+                description: |-
+                  HaltAfterRestore defines whether the VM should remain halted after the restore is complete.
+                  If false (default), the VM will be started after a successful restore.
+                type: boolean
               keepMacAddress:
                 description: |-
                   KeepMacAddress only works when NewVM is true.
--- a/pkg/api/vm/handler_test.go
+++ b/pkg/api/vm/handler_test.go
@@ -16,12 +16,13 @@ import (
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	kubevirtutil "kubevirt.io/kubevirt/pkg/virt-operator/util"
 
+	cniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
 	harvesterv1 "github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/harvester/pkg/controller/master/migration"
 	"github.com/harvester/harvester/pkg/generated/clientset/versioned/fake"
 	"github.com/harvester/harvester/pkg/util"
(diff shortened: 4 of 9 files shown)


WebberHuang1118 (commented):
@votdev 

Suggest aggregating the customized parts for this PR’s source issue, [#10422](https://github.com/harvester/harvester/issues/10422), and a similar issue, [#4098](https://github.com/harvester/harvester/issues/4098), into a well-considered overall design. Maybe one PR can address both issues.

IMO, there may be no need to change the manifest. Suggest using a single annotation and marshaling the options into the annotation value, which could make the logic simpler.

cc @Vicente-Cheng @ihcsim @brandboat

votdev:
> I don't think annotations are a good idea because they aren't documented, and no one can remember the name of an annotation, let alone the available values. Changes to the CR, on the other hand, are documented and can be retrieved at any time using Kubernetes tools. That's why I would prefer changes to the CR.

I don't think annotations are a good idea because they aren't documented, and no one can remember the name of an annotation, let alone the available values. Changes to the CR, on the other hand, are documented and can be retrieved at any time using Kubernetes tools. That's why I would prefer changes to the CR.

WebberHuang1118:
> I don't think annotations are a good idea because they aren't documented, and no one can remember the name of an annotation, let alone the available values.

I’m okay with changing the manifest instead of using annotations, if you think that is more suitable.

votdev:
Maybe I should explain a bit more about why I modified the CR and added the `haltAfterRestore` field. In the first version of the PR with commit https://github.com/harvester/harvester/pull/10480/changes/569052c38d31b56e184d4b3d6b1d421d66969273, I introduced the `runStrategy` field. This is very powerful and would give the user countless more options to influence the VM’s behavior during restoration. BUT, this makes things far too complicated; and I’m not sure if such extensive control is even useful. Therefore, I revised the PR and introduced only the boolean value `haltAfterRestore`. This is easier for the user to understand and also much easier to handle from an implementation perspective. I preferred a change to the CR over an annotation because the field is documented thanks to OpenAPI, and Kubernetes automatically validates the values for me. It’s all much more elegant than an annotation.

WebberHuang1118:
> Maybe I should explain a bit more about why I modified the CR and added the `haltAfterRestore` field. In the first version of the PR with commit [569052c](https://github.com/harvester/harvester/commit/569052c38d31b56e184d4b3d6b1d421d66969273), I introduced the `runStrategy` field. This is very powerful and would give the user countless more options to influence the VM’s behavior during restoration. BUT, this makes things far too complicated; and I’m not sure if such extensive control is even useful. Therefore, I revised the PR and introduced only the boolean value `haltAfterRestore`. This is easier for the user to understand and also much easier to handle from an implementation perspective. I preferred a change to the CR over an annotation because the field is documented thanks to OpenAPI, and Kubernetes automatically validates the values for me. It’s all much more elegant than an annotation.

That makes sense, thanks for the detailed explanation.

brandboat:
> In the first version of the PR with commit https://github.com/harvester/harvester/commit/569052c38d31b56e184d4b3d6b1d421d66969273, I introduced the runStrategy field.

@votdev Great work on this. Just a small housekeeping reminder: could we squash the commits into one before merge? I’m looking at 569052c and noticed it's mostly undone by the following commit. To keep the git log clean and easy for the rest of the team to follow (and to avoid any 'Rebase and Merge' mishaps), squashing them into a single commit would be ideal!

pkg/api/vm/handler.go
```diff
@@ -876,15 +876,26 @@ func (h *vmActionHandler) restoreBackup(vmName, vmNamespace string, input Restor
 	if _, err := h.backupCache.Get(vmNamespace, input.BackupName); err != nil {
 		return err
 	}
-	apiGroup := kubevirtv1.SchemeGroupVersion.Group
+
+	restore := buildVirtualMachineRestore(vmName, vmNamespace, input)
+
+	_, err := h.restoreClient.Create(restore)
+	if err != nil {
+		return fmt.Errorf("failed to create restore, error: %s", err.Error())
```
brandboat: nit:
```suggestion
		return fmt.Errorf("failed to create restore, error: %v", err)
```

brandboat (dismissed):

pkg/api/vm/handler.go
votdev: Fixed

votdev:
> > In the first version of the PR with commit [569052c](https://github.com/harvester/harvester/commit/569052c38d31b56e184d4b3d6b1d421d66969273), I introduced the runStrategy field.
> 
> @votdev Great work on this. Just a small housekeeping reminder: could we squash the commits into one before merge? I’m looking at [569052c](https://github.com/harvester/harvester/commit/569052c38d31b56e184d4b3d6b1d421d66969273) and noticed it's mostly undone by the following commit. To keep the git log clean and easy for the rest of the team to follow (and to avoid any 'Rebase and Merge' mishaps), squashing them into a single commit would be ideal!

Done.

brandboat (approved):

Yu-Jack (approved):
LGTM, just share some thoughts.

I think using CR is more suitable way for this case because this is not like using other CRs which are not controlled by us. For example, we added a lot of annotations in `VirtualMachine` CR. That's because it belongs to third party, not us. 

Besides, I think introducing RunStrategy concept would be useful for future case. That said, as I know, we haven't had enough use cases for this part. We can keep this in mind.

Issue #9515, opened 2025-11-14
[BUG] Create an image which name length over the limitaion via YAML, the image still can be created


asc5543:
### Describe the Bug

Create an image which name length over the limitaion via YAML, the image still can be created.

https://github.com/user-attachments/assets/8b0809f5-d4f0-4e7d-980a-12613cc8b960

<img width="1903" height="1043" alt="Image" src="https://github.com/user-attachments/assets/f26b63e5-0af9-4b9f-ba26-008360ef26fd" />

### To Reproduce

1. Click `Create` on images page
1. Fill in `very-long-image-name-for-testing-kubernetes-name-length-limits-that-exceeds-maximum` in Name 
    <img width="1903" height="1043" alt="Image" src="https://github.com/user-attachments/assets/1b5d2f7a-b960-4719-aa92-b404ff46fe06" />
1. Fill in the URL and click `Edit as YAML`, then click `Create`
    <img width="1903" height="1043" alt="Image" src="https://github.com/user-attachments/assets/fd5d416b-7a21-49b0-b397-375a5ce566f9" />

### Expected Behavior

The UI should shows error message and block the creation.

<img width="1903" height="1043" alt="Image" src="https://github.com/user-attachments/assets/b447a44e-af5f-4393-85b3-c3d0f3587d51" />

### Support Bundle for Troubleshooting

[longnameimage.zip](https://github.com/user-attachments/files/23541476/longnameimage.zip)

### Environment

- Harvester version: `v1.7.0-rc1`
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Additional context

This issue is found on `v1.7.0-rc1` automation e2e test.
http://10.115.1.7/job/harvester-install-and-test/20/TestReport/

lanfon72:
Verified this bug has been fixed.

Test Information
----
* Environment: **MegaRAC SP-X  3 nodes**
* Harvester Version: **v1.9.0-dev-20260510**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. Follow *To Reproduce* steps in https://github.com/harvester/harvester/issues/9515#issue-3624240266

PR #10490, opened 2026-04-28
Make sure `spec.displaName` is a valid K8s label value

votdev:
#### Problem:
Create an image which name length over the limitaion via YAML, the image still can be created.

#### Solution:
Add an additional check in the `VirtualMachineImage` validator that validates the `spec.displaName` field is a valid K8s label value; e.g. the length is limited to 63 chars because the field is later used as a label.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9515

#### Test plan:
***Case 1***
- Go to the `Images` page and press the `Create` button.
- Insert `xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx`
- Insert `https://download.cirros-cloud.net/0.6.2/cirros-0.6.2-x86_64-disk.img` into the `URL` field.
- Press the `Edit as YAML` button.
- Press the `Create` button.
- The following error message must be displayed.
grafik

***Case 2***
Run the following command:
```
$ cat <<EOF | kubectl apply -f -
kind: VirtualMachineImage
metadata:
  name: image-1abcd
  namespace: default
spec:
  backend: backingimage
  displayName: >-
    _xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
  sourceType: download
  targetStorageClassName: harvester-longhorn
  url: https://download.cirros-cloud.net/0.6.2/cirros-0.6.2-x86_64-disk.img
EOF
```
The output must look like:
```
harvester-node-0:/home/rancher # cat <<EOF | kubectl apply -f -
> apiVersion: harvesterhci.io/v1beta1
> kind: VirtualMachineImage
> metadata:
...
>   sourceType: download
>   targetStorageClassName: harvester-longhorn
>   url: https://download.cirros-cloud.net/0.6.2/cirros-0.6.2-x86_64-disk.img
> EOF
The request is invalid: spec.displayName: displayName is not a valid Kubernetes label value: must be no more than 63 characters; a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

### DIFF
--- a/pkg/image/common/validator.go
+++ b/pkg/image/common/validator.go
@@ -13,6 +13,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/validation"
 	authorizationv1client "k8s.io/client-go/kubernetes/typed/authorization/v1"
 
 	"github.com/harvester/harvester/pkg/apis/harvesterhci.io/v1beta1"
@@ -85,7 +86,8 @@ func (v *vmiValidator) GetStatusSC(vmi *v1beta1.VirtualMachineImage) string {
 	return vmi.Status.StorageClassName
 }
 
-func (v *vmiValidator) CheckDisplayName(vmi *v1beta1.VirtualMachineImage) error {
+// commonCheckDisplayName performs several validations on the `DisplayName` field.
+func (v *vmiValidator) commonCheckDisplayName(vmi *v1beta1.VirtualMachineImage) error {
 	if vmi.Spec.DisplayName == "" {
 		return werror.NewInvalidError("displayName is required", fieldDisplayName)
 	}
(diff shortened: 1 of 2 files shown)


WebberHuang1118 (dismissed):

brandboat (dismissed):

pkg/image/common/validator.go
```diff
@@ -90,6 +91,11 @@ func (v *vmiValidator) CheckDisplayName(vmi *v1beta1.VirtualMachineImage) error
 		return werror.NewInvalidError("displayName is required", fieldDisplayName)
 	}
 
+	// The `displayName` field is later used as a Kubernetes label value.
+	if errs := validation.IsValidLabelValue(vmi.Spec.DisplayName); len(errs) > 0 {
+		return werror.NewInvalidError(fmt.Sprintf("displayName is not a valid Kubernetes label value: %s", strings.Join(errs, "; ")), fieldDisplayName)
+	}
```
ihcsim: Just a note, I think this will block updates to any existing images with nonconforming display name, as `CheckDisplayName()` is called by `CheckUpdateDisplayName()`. Since display name is immutable, there won't be a way for user to fix the issue. Might be better to limit this validation to new images only, and let K8s backend handle any illegal label use of existing images' display name.

brandboat: Good point!

votdev: Fixed

brandboat (approved):

ihcsim (approved):

Vicente-Cheng (approved):
thanks!

Issue #10388, opened 2026-04-15
[BUG] vGPU VMs is not regarded as Non-Migratable when triggering upgrade.


albinsun:
### Describe the Bug

vGPU VMs is not (but should be) regarded as Non-Migratable when triggering upgrade.

### To Reproduce

1. Setup `harvester-v1.7.1`
2. Create 2 VMs, **pcigpu** with PCI GPU and **vgpu** one with vGPU
   <img width="1566" height="768" alt="Image" src="https://github.com/user-attachments/assets/8b494f91-3e59-4a22-97ef-c9610847b66f" />
   <img width="1566" height="768" alt="Image" src="https://github.com/user-attachments/assets/3d2059ab-0649-4b8d-8373-230aab84a0d9" />
   <img width="1566" height="768" alt="Image" src="https://github.com/user-attachments/assets/b1dfb3ad-ff69-44de-820d-5023e6ea10dd" />

2. Trigger upgrade to `v1.8.0-rc5`, only **pcigpu** is considered as **non-live migratable** but not **vgpu**.
   <img width="1566" height="768" alt="Image" src="https://github.com/user-attachments/assets/0b4e3a54-b576-4406-a914-bbb6fb22e62c" />

### Expected Behavior

vGPU VMs should be regard as Non-Migratable.
See https://docs.harvesterhci.io/v1.7/vm/live-migration#non-migratable-virtual-machines

### Support Bundle for Troubleshooting

[supportbundle_bb01a69c-818c-450f-8558-0866923364eb_2026-04-15T05-17-19Z.zip](https://github.com/user-attachments/files/26736801/supportbundle_bb01a69c-818c-450f-8558-0866923364eb_2026-04-15T05-17-19Z.zip)

### Environment

- Harvester version: `v1.7.1`
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Workaround and Mitigation

Manually stop the VM before or during the upgrade.

Yu-Jack:
Um .. this one is interesting. Starting from v1.8, vGPU device has been considered as a PCI device which is in `vmi.Spec.Domain.Devices.HostDevices`. 

```go
// vGPU device before v1.8: 
vmi.Spec.Domain.Devices.GPUs

// vGPU device from v1.8:
vmi.Spec.Domain.Devices.HostDevices
```

So, the code actually works when users try to upgrade from **v1.8**.

[...]

PR #10508, opened 2026-04-30
feat: add more explanation when checking hostDevice during upgrade

Yu-Jack:
#### Problem:
Actually, there are no problems here. 

Starting from v1.8.0, vgpu is moved to hostDevices. So, we just need to check hostDevices.

#### Solution:
Add more words in log.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/10388


pkg/util/virtualmachineinstance/virtualmachineinstance_test.go
```diff
@@ -497,7 +497,12 @@ func Test_GetNonLiveMigratableVMIs(t *testing.T) {
 					Spec: kubevirtv1.VirtualMachineInstanceSpec{
 						Domain: kubevirtv1.DomainSpec{
 							Devices: kubevirtv1.Devices{
-								HostDevices: []kubevirtv1.HostDevice{},
```
Yu-Jack, line 500: The original test case seems weird. Empty host devices should be also considered as a live-migratable VM.

starbops (approved):
LGTM, thank you

Issue #9777, opened 2025-12-22
[TASK] Build benchmark policy


innobead:
### Task request

Harvester had a benchmark report for VM disk performance, but lacks a consistent benchmarking mechanism and framework. This task focuses on creating receipts to establish a regular benchmark and report along with each feature/minor release.

### Additional context

- https://github.com/harvester/harvester/wiki/Harvester-Performance-Result
- https://github.com/kubevirt/kubevirt/blob/main/docs/perf-scale-benchmarks.md
- https://www.redhat.com/en/blog/kubevirt-scale-test-creating-400-vmis-per-node

cc @harvester/pm @harvester/tpm

innobead:
@krumware is interested in helping with benchmarking. @ibrokethecloud @Vicente-Cheng @ihcsim, any ideas on the test plan or tools for his reference? It might be good to start with a HEP.

ihcsim:
For general network, storage and system performance benchmarking with KubeVirt vm workloads, https://github.com/cloud-bulldozer/benchmark-operator might be a good place to start. It uses system tools like uperf, fio and stressng to run e2e tests on vm and vmi. We don't have to use the operator if we don't like it, but at least it gives some ideas of the nature of the performance tests.

In terms of test cases, I am thinking about them from the perspective of Harvester use cases which may involve measuring multiple software performance indicators, instead of just lower level measurements like raw IOPS, MiB/s. The goal is to give user a rough view of _"how much timedoes it take to complete task X, given N amount of CPU,  memory and network bandwith"_.

A quick brain dump on the test cases: I think one high-level use case where performance matters much to users is vm migration, which involves both low level network and storage operations. Measurements covering network latency and throughput among vm workloads on in-cluster vm network and storage network will also be interesting. On the guest cluster side, there was a recent question on the throughput of the Harvester CSI driver. The context is about whether the CSI driver can keep up during rapid shutting down and recreation of guest nodes and their associated virtual machines, before LH on the Harvester side timed out.

How do others feel about this?

krumware:
I've done a ton of work on this in the past few weeks, and have automation for running ramp tests and saturation tests on dual stack, and across multiple test clusters. It's orchestrating the cloud-bulldozer operator but adding multiple test orchestration to it. Also using Garage to store logs/results/hardware/OS-info from the workloads to reduce overhead requirements that might be introduced by running an elasticsearch in-cluster (also to reduce complexity). Will definitely be sharing.

[...]

Issue #8493, opened 2025-06-16
[TASK] VM performance benchmarking


rrajendran17:
### Task request

Created this task as placeholder to experiment VM performance on secondary networks using multus CNI (without any virtual functions like SR-IOV)

The outcome of this task should benchmark VM Performance/throughput achievable on physical NICs using iperf tests
- Different Linux modes like active/backup,LACP on mgmt and user defined cluster (downtime during switching/link down)
- tests with GRO/GSO disabled
-  tests with Jumbo frames 
- Identify the bottlenecks
- Identify ways to improve performance
- work on KB article for performance issues with certain NIC drivers with flags GRO/GSO enabled
   and workaround

### Additional context
Related to the following issues:
https://github.com/harvester/harvester/issues/7224
https://github.com/harvester/harvester/issues/8154

PR #10526, opened 2026-05-05
hep: benchmark policy

pohanhuang:
#### Problem:
#9777

#### Solution:
HEP for #9777

#### Related Issue(s):
HEP for #9777

#### Open Question
1. **Cluster topology**: Does Harvester support the same 4-node setup used by KubeVirt's SIG-scale presubmit (1 control-plane + 3 workers, see
   [kubevirt-presubmits-1.8.yaml](https://github.com/kubevirt/project-infra/blob/main/github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits-1.8.yaml)?

   I want to keep the reproduction close to the https://github.com/kubevirt/kubevirt/blob/main/docs/perf-scale-benchmarks.md.
   If not, should we re-run the benchmark ourselves with a Harvester-supported topology to keep the comparison fair (1 control-plane + 2 worker node) ?

2. **KubeVirt version baseline**: Which KubeVirt version should we compare
   against — the one bundled with the Harvester release under test (e.g. Harvester 1.8.0 vs its embedded KubeVirt version), or a fixed upstream version (e.g. Harvester 1.8.0 vs vanilla KubeVirt 1.8.0)?

### DIFF
--- a/enhancements/20260424-harvester-benchmark-policy.md
+++ b/enhancements/20260424-harvester-benchmark-policy.md
@@ -59,32 +59,43 @@ N/A
 
 ### Implementation Overview
 
-The benchmark policy is organized into four categories. Each defines scope, tooling, topology, methodology, and key metrics.
+The benchmark policy is organized into three categories. Each defines scope, tooling, topology, methodology, and key metrics.
 
 #### Category 1: Resource Footprint
 
-- **Scope**: CPU and memory consumption of each Harvester component (pods) at idle and under VMs load.
+- **Scope**: CPU and memory consumption of each Harvester component at idle and under maximum VM density.
 - **Tooling**:
-  - `kubectl top pods`, know the pod cpu, memory usages
-  - `grep -E 'MemTotal|MemFree|MemAvailable|SReclaimable' /proc/meminfo`, host memory allocation.
-  - `kubectl describe node | grep -A5 'Allocated resources'`, know the scheduler cpu/memory resource allocation.
+  - `kubectl top pods` — per-pod CPU/memory usage.
+  - `ps -C containerd,kubelet,rke2 -o rss=` — host-level process RSS.
+  - `grep -E 'MemTotal|MemFree|MemAvailable|SReclaimable' /proc/meminfo` — host memory allocation.
+  - `kubectl describe node | grep -A5 'Allocated resources'` — scheduler CPU/memory resource allocation.
+  - `etcdctl endpoint status` — etcd db size.
 - **Topology**:
   - Single-node: 1 control-plane node, no worker nodes.
   - Multi-node: 3-node cluster (1 control-plane, 2 workers).
 - **Methodology**:
   - [Harvester Resource Footprint gist](https://gist.github.com/Vicente-Cheng/ae8d5ce743f94c0c9a333574813bc777).
-  - Capture component-level CPU/memory at rest, then with increasing VM counts (e.g., 10 VMs, 20 VMs) to understand how resource consumption scales.
+  - Phase 1: Measure at idle (0 VMs running) to establish the platform baseline.
+  - Phase 2: Create VMs until the scheduler rejects — record the maximum VM count and which resource (CPU, memory, or other) is the bottleneck.
+  - Capture per-namespace breakdown (kube-system, harvester-system, longhorn-system, cattle-system) at each phase.
+  - Compare single-node vs multi-node to understand how platform overhead scales with node count.
 - **Key metrics**:
-  - CPU (millicores), memory RSS (MiB), per pod, per VM count.
-  - VMs can be created in a given environment before a component becomes the bottleneck.
+  - Platform idle cost: total CPU (millicores) and memory (MiB) with 0 VMs.
+  - Maximum VM count before scheduler rejects, and the bottleneck resource.
+  - Scheduler-reserved vs actual physical consumption ratio.
+  - etcd db size (MiB).
 
-#### Category 2: Performance Benchmark (I/O)
+#### Category 2: Performance Benchmark
+
+##### 2a: Storage and Network I/O
 
 - **Scope**: Storage throughput/latency and network throughput.
 - **Tooling**:
   - Storage: `fio` (sequential and random read/write, 4K and 128K block sizes)
   - Network: `iperf3` (TCP throughput, UDP jitter)
-- **Topology**: Multi-node cluster.
+- **Topology**:
+  - Single-node: 1 control-plane node, no worker nodes.
+  - Multi-node: 3-node cluster (1 control-plane, 2 workers).
 - **Methodology**: [Harvester Performance wiki](https://github.com/harvester/harvester/wiki/Harvester-Performance-Result).
   - Raw Device measure - fio / iperf3
   - VMs on Longhorn volumes for storage tests.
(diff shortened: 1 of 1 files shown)


enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,165 @@
...
+  - Network: `iperf3` (TCP throughput, UDP jitter)
+- **Topology**:
+  - Single-node: 1 control-plane node, no worker nodes.
+  - Multi-node: 3-node cluster (1 control-plane, 2 workers).
+- **Methodology**: [Harvester Performance wiki](https://github.com/harvester/harvester/wiki/Harvester-Performance-Result).
+  - Raw Device measure - fio / iperf3
+  - VMs on Longhorn volumes for storage tests.
+    - Linux VM - fio
+    - Windows VM - fio
+- **Key metrics**: IOPS, throughput (MB/s), latency (ms), bandwidth (Gbps).
```
WebberHuang1118: Suggest adding more details about storage network connectivity, such as network bandwidth between nodes and whether a dedicated storage network is used, for example, when LH is the provider.

It may also be helpful to include any storage performance impact from the hypervisor, if applicable, such as comparing direct storage access on the node with storage access inside the guest VM.

WebberHuang1118 (commented):
Some suggestions from the storage perspective. Please take a look, thanks.

enhancements/20260424-harvester-benchmark-policy.md
pohanhuang: Hi @WebberHuang1118 
Thanks for sharing the insight, I just update the HEP, please help review, thanks.

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,169 @@
...
+  - Compare single-node vs multi-node to understand how platform overhead scales with node count.
+- **Key metrics**:
+  - Platform idle cost: total CPU (millicores) and memory (MiB) with 0 VMs.
+  - Maximum VM count before scheduler rejects, and the bottleneck resource.
+  - Scheduler-reserved vs actual physical consumption ratio.
+  - etcd db size (MiB).
+
+#### Category 2: Performance Benchmark
+
+##### 2a: Storage and Network I/O
```
rrajendran17, line 130: I have created a placeholder tasks in the past for bench marking on VM Networking.
https://github.com/harvester/harvester/issues/8493. you can add those cases to the HEP. Please let me know in case you need more details regarding each case.

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### Category 1: Resource Footprint
+
+- **Scope**: CPU and memory consumption of each Harvester component at idle and under maximum VM density.
+- **Tooling**:
+  - `kubectl top pods` — per-pod CPU/memory usage.
+  - `ps -C containerd,kubelet,rke2 -o rss=` — host-level process RSS.
+  - `grep -E 'MemTotal|MemFree|MemAvailable|SReclaimable' /proc/meminfo` — host memory allocation.
+  - `kubectl describe node | grep -A5 'Allocated resources'` — scheduler CPU/memory resource allocation.
+  - `etcdctl endpoint status` — etcd db size.
```
ihcsim: One thing to note is that these only offer us a point-in-time view of the ([gauge](https://prometheus.io/docs/concepts/metric_types/#gauge)) usages, which may not be accurate. E.g., the spikes and abnormalies may happen _during_ the performance test, which can be easy to miss, unless you have a way of recording them over time. The _final_ values these tools report are what we will see _after_ your performance test is done.

I think we will need to enable monitoring to let Prometheus capture the relevant metrics, to determine the max/min/average usage trend over time.

ibrokethecloud:
based on our quick discussion we should
* install monitoring stack during benchmarking to allow us to improve metric collection across various components during the upgrade. we can document the additional overhead due to the monitoring itself as part of the benchmarking results
* we probably do not need compare benchmarks against vanilla kubevirt install in all scenarios
* we should document the steps used for running the tests. If needed this allows users to run these steps to benchmark their own systems.

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,169 @@
...
+  - **Storage network**: 10GbE dedicated NIC for Longhorn replication traffic, isolated from management and VM networks.
+  - **Inter-node bandwidth**: Measured via iperf3 before benchmark runs to confirm line rate (record actual observed bandwidth).
+- **Methodology**: [Harvester Performance wiki](https://github.com/harvester/harvester/wiki/Harvester-Performance-Result).
+  - **Storage overhead**:
+    - Layer 0 (Raw Device): fio directly on host block device
+    - Layer 1 (Longhorn volume, host): fio on Longhorn PVC mounted to a pod on the host
+    - Layer 2 (guest VM): fio inside Linux/Windows VM backed by Longhorn volume
+  - **Network overhead**:
+    - Layer 0 (Raw Device): confirm inter-node line rate via iperf3
+    - VM network: iperf3 between two VMs to measure guest network throughput.
```
WebberHuang1118: We may also need a test matrix for the case where LH has only one local replica. In this case, there is no network traffic for LH I/O.

enhancements/20260424-harvester-benchmark-policy.md
ihcsim: We should remove line 87 and 91 as we won't be using them to measure cpu/memory usage. Try these PromQL queries from the Prometheus console (we can fiddle around with the query labels to e.g. filter by namespace, pod labels etc.):

```
# actual cpu usage (in cores) per pod in the past 5 mins:
sum(rate(container_cpu_usage_seconds_total{container!="", container!="POD"}[5m])) by (namespace, pod)

# ratio of cpu usage to cpu limit (if limit is defined)
sum(rate(container_cpu_usage_seconds_total{container!="", container!="POD"}[5m])) by (namespace, pod) 
/ 
sum(kube_pod_container_resource_limits{resource="cpu"}) by (namespace, pod) * 100

# memory usage usage (working set) per pod
sum(container_memory_working_set_bytes{container!="", container!="POD"}) by (namespace, pod)

# ratio of memory usage to memory limit (if limit is defined)
sum(container_memory_working_set_bytes{container!="", container!="POD"}) by (namespace, pod) 
/ 
sum(kube_pod_container_resource_limits{resource="memory"}) by (namespace, pod) * 100
```

References:
* some pod-level metrics can be found at https://kubernetes.io/docs/reference/instrumentation/metrics/
* container-level metrics can be found at https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,225 @@
...
+      - **Replicated**: Set Longhorn volume replicas to 2 to measure replication overhead on storage throughput.
+  - **Inter-node bandwidth**: Measured via iperf3 before benchmark runs to confirm line rate (record actual observed bandwidth).
+- **Methodology**: [Harvester Performance wiki](https://github.com/harvester/harvester/wiki/Harvester-Performance-Result).
+  - **Storage overhead (layered testing)**:
+    - **Layer 0 (Raw Device)**: fio directly on host block device (multi-node host controller node).
+    - **Layer 1 (Longhorn volume, host)**: fio on Longhorn PVC mounted to a pod on the host (multi-node, with replica count ≥ 2).
+    - **Layer 2 (guest VM)**: fio inside Linux/Windows VM backed by Longhorn volume (multi-node, with Kube-OVN enabled/disabled comparison).
+  - **Network overhead (layered testing)**:
+    - **Layer 0 (Raw Device)**: Confirm inter-node line rate via iperf3 between bare metal nodes or host pods.
+    - **Layer 1 (VM network)**: iperf3 between two VMs on different nodes to measure guest network throughput.
```
rrajendran17, line 158: Also please include tests to verify Jumbo frames (as fragmentation and reassembly can cause overhead) and also tests with different bond modes (eg, active/backup, 802.3ad/lacp). Based on the availability of HW in our lab we can try to benchmark these on BCM, Dell etc(some of the NIC drivers had throughput issues in the past).

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,225 @@
...
+- **Storage/network throughput**: What IOPS, bandwidth, and latency can VMs expect under realistic workloads?
+
+**Story 2: As a Harvester software engineer improving performance**
+I want profiling data so that I can identify bottlenecks, catch potential OOMs before they hit production, and prove my optimizations work. Specifically:
+
+- **Per-component resource profiling**: CPU/memory consumption of each pod as VM count grows, to understand which component becomes the bottleneck first (e.g., instance-manager at ~20 GiB per pod with 149 volumes).
+- **Memory pressure analysis**: Identify components at risk of OOM — comparing scheduler-reserved memory vs actual RSS, and flagging components where actual consumption grows disproportionately with VM count.
+- **Upgrade per-phase timing**: Break down upgrade duration into pre-drain, node drain, post-drain, and total, so we can pinpoint which phase is slow and whether optimizations reduce it (see [#10421](https://github.com/harvester/harvester/issues/10421)).
+- **VM migration profiling**: Duration and resource cost of live migration under varying VM sizes and storage backends, since migration involves both network and storage I/O.
```
rrajendran17: Also add VMs using overlay Network when testing VM migration.

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,225 @@
...
+  - **Monitoring**:
+    - Prometheus (enabled via monitoring stack) — track node-level disk I/O, network bandwidth, and CPU iowait during tests.
+- **Topology**:
+  - **Multi-node (3-node cluster)**: For cluster-level storage/network tests.
+    - All three nodes: 1 control-plane, 2 workers.
+    - **Storage network**: 10GbE dedicated NIC for Longhorn replication traffic, isolated from management and VM networks.
+    - **Kube-OVN enabled**: Test with Kube-OVN networking to measure its impact on VM network performance (compare against default Canal/Multus).
+    - **Longhorn storage profiles**:
+      - **Local-single-replica**: Set Longhorn volume replicas to 1 and place the workload on the node hosting that replica to measure stack overhead without Longhorn replication traffic.
+      - **Replicated**: Set Longhorn volume replicas to 2 to measure replication overhead on storage throughput.
```
WebberHuang1118: In most cases, if users choose multiple replicas to improve data resilience, they usually choose 3 replicas.

WebberHuang1118: Should we focus only on LH?

We may want to do deeper profiling for LH since it is the most recommended storage option. However, with the increasing demand for third-party storage, we may also need profiling at the CSI layer.

This means both Layer 1 and Layer 2 can still be checked.

pohanhuang:
@rrajendran17 @WebberHuang1118 @ihcsim can you help take a look about the overall benchmark policy HEP when you are available?
Thanks.

pohanhuang:
cc: @starbops

WebberHuang1118 (approved):
LGTM, thanks.

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,231 @@
...
+### Goals
+
+- **Define Benchmark Categories**: Establish canonical categories covering Harvester's performance surface — Resource Footprint (including VM creation latency) and I/O Performance.
+- **Standardize Tooling and Methodology**: Specify tools and procedures for each category so results are reproducible by any team member and users can run benchmarks in their own environments.
+- **Maintain a Shared Baseline**: Publish benchmark results to the Harvester wiki each release and compare against the previous baseline.
+
+### Non-goals
+
+- Full automation of benchmark execution.
+- Profiling and root-cause analysis (a follow-up action triggered by benchmark results, not part of the benchmark itself).
```
ihcsim, line 25: later on, it will be also interesting to see what we can get out of golang pprof

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,231 @@
...
+    - Create VMs incrementally with 100ms interval (minimal spec: Cirros or Alpine, 100m CPU, 128Mi memory) until the scheduler rejects new VMs.
+    - **During VM creation**, capture VM creation latency and component resource usage via Prometheus:
+      - `kubevirt_vmi_phase_transition_time_from_creation_seconds` (histogram) — P50, P95, P99.
+      - Per-pod CPU and memory time-series to identify transient spikes and bottlenecks during the run.
+      - Total time from first VM creation to last VM reaching Running (batch throughput).
+    - Record the maximum VM count and identify the bottleneck resource (CPU, memory).
+    - Capture per-component resource consumption at max VM density using the same Prometheus metrics as Phase 1, and report peak as well as average usage.
+    - Ensure workloads are evenly distributed among workers to avoid skewed results.
+  - Compare single-node vs multi-node to understand how platform overhead scales with node count.
+  - **User-reproducible script**: Provide a standalone script with VM manifest templates and PromQL queries so users can run in their own environment to validate expected resource footprint and VM creation performance.
```
ihcsim, line 115: if possible, we should look into morphing the storage validation tool to include these benchmark tooling, so that we maintain just one user-facing tool, instead of a scattered collection of tools and scripts

enhancements/20260424-harvester-benchmark-policy.md
```diff
@@ -0,0 +1,231 @@
...
+### User-Reproducible Benchmark Scripts
+
+All benchmark categories will include standalone scripts that users can run in their own Harvester environments to:
+- Validate that their hardware meets expected performance baselines.
+- Debug performance issues by comparing local results against published benchmarks.
+- Reproduce benchmark results for support cases.
+
+### Future Work
+
+- Benchmark automation (scheduled runs, result diffing, automatic regression issue filing).
```
ihcsim, line 229: one option is to run the benchmark scripts/tools on gha

ihcsim (approved):

enhancements/20260424-harvester-benchmark-policy.md
pohanhuang: Let me add those in the non goals

enhancements/20260424-harvester-benchmark-policy.md
pohanhuang: @ihcsim does harvester has such tools?
Can you point out the right repo? https://github.com/harvester/storage-validator is this the right one?

enhancements/20260424-harvester-benchmark-policy.md
pohanhuang: My intention is to do like https://github.com/kubevirt/ci-performance-benchmarks, which we generate the result in the CI, then we can access it through a github page.

ihcsim (approved):

ihcsim:
I think we have made good progress on this HEP to collect and analyze the initial sets of requirements. I'm merging this HEP, and we will continue to iterate on it.

Issue #10529, opened 2026-05-05
[BUG] Upgrade could fail without retrying when entering the node upgrade phase


starbops:
### Describe the Bug

The upgrade from v1.7.1 to v1.7-head failed abruptly upon entering the node upgrade phase.

The Upgrade CR indicated that the upgrade attempt had ended:

```yaml
...
status:
  conditions:
  - lastUpdateTime: "2026-05-05T09:51:40Z"
    reason: 'Internal error occurred: failed calling webhook "rancher.cattle.io.clusters.provisioning.cattle.io":
      failed to call webhook: Post "https://rancher-webhook.cattle-system.svc:443/v1/webhook/mutation/clusters.provisioning.cattle.io?timeout=10s":
      no endpoints available for service "rancher-webhook"'
    status: "False"
    type: Completed
...
```

From the Harvester pod's log:

```
...
time="2026-05-05T09:51:30Z" level=info msg="Start upgrading Kubernetes runtime to v1.34.7+rke2r1"
time="2026-05-05T09:51:36Z" level=info msg="VM default/vm001 is migratable, removing skipping descheduling annotation"
time="2026-05-05T09:51:40Z" level=info msg="handle upgrade harvester-system/hvst-upgrade-b4xss with labels map[harvesterhci.io/latestUpgrade:true harvesterhci.io/upgradeStat
e:Failed]"
time="2026-05-05T09:51:40Z" level=info msg="upgrade failed... starting post-upgrade cleanup"
...
```

By observing the `rancher-webhook` pod, it did restart nearly the time point.

### To Reproduce

1. Prepare a 2-node cluster in v1.7.1
2. Trigger upgrade to v1.7-head
3. The upgrade could fail after the system service upgrade phase but before the node upgrade phase

### Expected Behavior

The upgrade controller should have kept retrying updating the provisioning Cluster CR to overcome transient errors.

### Support Bundle for Troubleshooting

[supportbundle_6e618aee-336a-4c44-a26c-82f91f56ed32_2026-05-05T11-15-37Z.zip](https://github.com/user-attachments/files/27397051/supportbundle_6e618aee-336a-4c44-a26c-82f91f56ed32_2026-05-05T11-15-37Z.zip)

### Environment

- Harvester version: custom-built v1.7-head (along with several upgrade-related fixes)
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630): ipxe-examples (KVM)
[...]

PR #10536, opened 2026-05-06
fix(upgrade): retry transient cluster update via reconcile loop

starbops:
#### Problem:

When the rancher-webhook pod is momentarily unavailable, the apiserver rejects the provisioning Cluster CR update with an "Internal error occurred: failed calling webhook ..." error. The upgrade controller previously treated this as a permanent failure and required users to restart the upgrade from scratch.

#### Solution:

Return the error from upgradeKubernetes so the wrangler controller framework re-queues and retries the reconcile until the webhook recovers. Before the cluster update, persist the in-memory replica-replenishment-wait-interval and reenable-descheduler-addon annotations on the Upgrade CR; otherwise a return-without-persist would discard them and the next reconcile would skip the save block (the live values are already mutated), leaving cleanup unable to restore the original Longhorn setting and re-enable the descheduler.

#### Related Issue(s):

#10529

#### Test plan:

1. Prepare a multi-node v1.8.0 Harvester cluster
2. Upgrade the cluster using a master-head build ISO image
3. Keep an eye on the upgrade progress; When the apply-manifest job is about to complete, scale the `rancher-webhook` deployment down to 0 to mimic the webhook unavailablility
4. The upgarde should stall and never enters the node upgrade phase
5. Observe the log from the harvester pod, the `Start upgrading Kubernetes runtime...` messages should appear multiple times
6. Scale the `rancher-webhook` deployment up to the original replica count
7. The upgrade should be unblocked and eventually completes

### DIFF
--- a/pkg/controller/master/upgrade/upgrade_controller.go
+++ b/pkg/controller/master/upgrade/upgrade_controller.go
@@ -727,7 +727,7 @@ func (h *upgradeHandler) resetLatestUpgradeLabel(latestUpgradeName string) error
 func (h *upgradeHandler) upgradeKubernetes(kubernetesVersion string) error {
 	cluster, err := h.clusterCache.Get("fleet-local", "local")
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get fleet-local/local cluster from cache: %w", err)
 	}
 
 	toUpdate := cluster.DeepCopy()
@@ -754,8 +754,10 @@ func (h *upgradeHandler) upgradeKubernetes(kubernetesVersion string) error {
 	updateDrainHooks(&toUpdate.Spec.RKEConfig.UpgradeStrategy.WorkerDrainOptions.PreDrainHooks, preDrainAnnotation)
 	updateDrainHooks(&toUpdate.Spec.RKEConfig.UpgradeStrategy.WorkerDrainOptions.PostDrainHooks, postDrainAnnotation)
 
-	_, err = h.clusterClient.Update(toUpdate)
-	return err
+	if _, err := h.clusterClient.Update(toUpdate); err != nil {
+		return fmt.Errorf("failed to update fleet-local/local cluster to kubernetes version %s: %w", kubernetesVersion, err)
+	}
+	return nil
 }
 
 func updateDrainHooks(hooks *[]rkev1.DrainHook, annotation string) {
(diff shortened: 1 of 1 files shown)


pkg/controller/master/upgrade/upgrade_controller.go
```diff
@@ -449,11 +449,21 @@ func (h *upgradeHandler) OnChanged(_ string, upgrade *harvesterv1.Upgrade) (*har
...
+				return nil, err
+			}
+			toUpdate = persisted.DeepCopy()
+
 			// go with RKE2 pre-drain/post-drain hooks
 			logrus.Infof("Start upgrading Kubernetes runtime to %s", info.Release.Kubernetes)
 			if err := h.upgradeKubernetes(info.Release.Kubernetes); err != nil {
-				setUpgradeCompletedCondition(toUpdate, StateFailed, corev1.ConditionFalse, err.Error(), "")
-				return h.upgradeClient.Update(toUpdate)
+				return upgrade, err
```
w13915984028, line 464: the `upgradeKubernetes` has two error return, it is better to wrap the errors with `fmt.Errorf(...)` to have a detailed information

scrictly, L720 error should be failed out right now, L748 deserves retry

https://github.com/harvester/harvester/blob/f29abc8df8e4b61ce4bb160af9c5623168a0bf8e/pkg/controller/master/upgrade/upgrade_controller.go#L720

https://github.com/harvester/harvester/blob/f29abc8df8e4b61ce4bb160af9c5623168a0bf8e/pkg/controller/master/upgrade/upgrade_controller.go#L748

w13915984028 (commented):
Following two fail points could also have retry.

https://github.com/harvester/harvester/blob/f29abc8df8e4b61ce4bb160af9c5623168a0bf8e/pkg/controller/master/upgrade/upgrade_controller.go#L389-L390

https://github.com/harvester/harvester/blob/f29abc8df8e4b61ce4bb160af9c5623168a0bf8e/pkg/controller/master/upgrade/upgrade_controller.go#L423-L425

pkg/controller/master/upgrade/upgrade_controller.go
starbops: I've updated `fmt.Errorf(...)` in the two places in 0441f7ce2862b5143b861f9f8fbd2d89f987b5ea. The failure modes should be distinguishable in the log.

w13915984028 (approved):
LGTM, thanks.

bk201 (approved):

Vicente-Cheng (approved):
thanks for the quick fix!

Vicente-Cheng:
@mergifyio backport v1.7

Vicente-Cheng:
@mergifyio backport v1.8

Issue #10532, opened 2026-05-05
[BUG] Upgrade from v1.7.x to v1.8.0 stuck in Post-draining when storage-network is configured with RWX volumes


brandboat:
### Describe the Bug

When upgrading from Harvester v1.7.1 to v1.8.0 with [storage-network-for-rwx-volume-enabled](https://longhorn.io/docs/1.10.2/references/settings/#storage-network-for-rwx-volume-enabled) set to true, the upgrade gets stuck permanently with one node in "Post-draining" state. The upgrade-repo deployment (replicas=2) is degraded (1/2 ready) because the second pod fails to mount its RWX PVC:

```
MountVolume.MountDevice failed for volume "pvc-ea2fbba7-7e5b-4446-a008-b2277892e8c8":
  mount -t nfs -o vers=4.1,noresvport,timeo=600,retrans=5,softerr
    None:/pvc-ea2fbba7-7e5b-4446-a008-b2277892e8c8 /var/lib/kubelet/plugins/...
  Output: mount.nfs: Failed to resolve server None: Name or service not known
```

### Root Cause:
Before the upgrade, with `storage-network-for-rwx-volume-enabled=true`, Longhorn creates a headless service (ClusterIP=None) for each share-manager and uses the FQDN endpoint `nfs://pvc-xxx.longhorn-system.svc.cluster.local/pvc-xxx`. This works because CoreDNS resolves the FQDN via the Endpoints object, which is populated with the pod's storage network IP.

During upgrade, Harvester's `OnRWXNetworkChange` → initRWXNetwork → [`isLHRWXShareStorageNetwork`](https://github.com/harvester/harvester/blob/6b392d8075c06c8c10938cd15cff062ed75b957e/pkg/controller/master/storagenetwork/storage_network.go#L1128-L1140)
[...]

PR #10543, opened 2026-05-07
Fix race in isLHRWXShareStorageNetwork during upgrade

brandboat:
#### Problem:

see #10532 desc

#### Solution:

Previously NotFound on the storage-network setting was swallowed, causing initRWXNetwork to mark rwx-network as initialized=true with share-storage-network=false before Longhorn's one-time migration had a chance to create the endpoint-network-for-rwx-volume CR. This caused harvester to overwrite endpoint-network-for-rwx-volume with an empty value, poisoning existing RWX mounts.

Return the error on NotFound to force a requeue until both settings exist.

#### Related Issue(s):

#10532

#### Test plan:

Note that this issue doesn't happen all the time.
1. prepare harvester v1.7.x cluster
2. following https://docs.harvesterhci.io/v1.7/advanced/storagenetwork and [https://docs.harvesterhci.io/v1.7/rancher/csi-driver/#rwx-volumes-support for](https://docs.harvesterhci.io/v1.7/rancher/csi-driver/#rwx-volumes-supportor) setting up the storage network.
3. perform upgrade to latest v1.8 branch
4. upgrade successfully

### DIFF
--- a/pkg/controller/master/storagenetwork/storage_network.go
+++ b/pkg/controller/master/storagenetwork/storage_network.go
@@ -1108,6 +1108,7 @@ func (h *Handler) initRWXNetwork(setting *harvesterv1.Setting) (*harvesterv1.Set
 	}
 	setting.Annotations[util.RWXNetworkInitializedAnno] = "true"
 	if !isShare {
+		logrus.Info("Longhorn RWX endpoint and storage network are not shared, initializing rwx-network with share-storage-network=false")
 		return h.settings.Update(setting)
 	}
 
@@ -1117,6 +1118,7 @@ func (h *Handler) initRWXNetwork(setting *harvesterv1.Setting) (*harvesterv1.Set
 		return setting, fmt.Errorf("failed to marshal rwx-network config: %v", err)
 	}
 	setting.Value = string(shareConfigJSON)
+	logrus.Info("Longhorn RWX endpoint and storage network are shared, initializing rwx-network with share-storage-network=true")
 	newSetting, err := h.settings.Update(setting)
 	if err != nil {
 		logrus.Errorf("failed to update rwx-network setting during init: %v", err)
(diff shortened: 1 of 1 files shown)


pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -1128,15 +1130,20 @@ func (h *Handler) initRWXNetwork(setting *harvesterv1.Setting) (*harvesterv1.Set
...
 	lhStorageNetwork, err := h.longhornSettings.Get(util.LonghornSystemNamespaceName, longhornStorageNetworkName, metav1.GetOptions{})
-	if err != nil && !apierrors.IsNotFound(err) {
+	// do NOT swallow NotFound here. During upgrade, Harvester's initRWXNetwork may run before Longhorn's
+	// one-time migration creates the endpoint-network-for-rwx-volume CR. Silently returning false would
+	// mark rwx-network as initialized=true with share-storage-network=false, causing harvester to
+	// overwrite endpoint-network-for-rwx-volume with an empty value and poison existing RWX mounts.
+	// Returning an error forces a requeue until both settings exist. See https://github.com/harvester/harvester/issues/10532
+	if err != nil || lhStorageNetwork == nil {
 		return false, fmt.Errorf("failed to get longhorn %s setting: %v", longhornStorageNetworkName, err)
 	}
```
WebberHuang1118, line 1140: IIUC, if `lhStorageNetwork` is `nil`, it will result in a not-found error.
```suggestion
	if err != nil {
return false, fmt.Errorf("failed to get longhorn %s setting: %v", longhornStorageNetworkName, err)
}
```

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -1128,15 +1130,20 @@ func (h *Handler) initRWXNetwork(setting *harvesterv1.Setting) (*harvesterv1.Set
...
 	lhRWXEndpoint, err := h.longhornSettings.Get(util.LonghornSystemNamespaceName, longhornEndpointNetworkForRWXVolume, metav1.GetOptions{})
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil || lhRWXEndpoint == nil {
 		return false, fmt.Errorf("failed to get longhorn %s setting: %v", longhornEndpointNetworkForRWXVolume, err)
 	}
```
WebberHuang1118, line 1144: IIUC, if lhRWXEndpoint is nil, it will result in a not-found error.
```suggestion
	if err != nil {
return false, fmt.Errorf("failed to get longhorn %s setting: %v", longhornEndpointNetworkForRWXVolume, err)
}
```

WebberHuang1118 (commented):
Thanks for the fix. Few suggestions.

pkg/controller/master/storagenetwork/storage_network.go
brandboat: Yeah, but if I don't add the nil check, go-lint raise error, see https://github.com/harvester/harvester/actions/runs/25475090211/job/74746798094
```
Running: golangci-lint run with --timeout=10m
pkg/controller/master/storagenetwork/storage_network.go:1145:91: SA5011: possible nil pointer dereference (staticcheck)
	logrus.Debugf("Longhorn storage network: %v, RWX endpoint network: %v", lhStorageNetwork.Value, lhRWXEndpoint.Value)
	                                                                                         ^
pkg/controller/master/storagenetwork/storage_network.go:1146:9: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
	return lhStorageNetwork != nil && lhRWXEndpoint != nil &&
```

Vicente-Cheng (approved):
lgtm, thanks for the fix

ihcsim (approved):

WebberHuang1118 (approved):
LGTM, thanks.

Issue #10540, opened 2026-05-06
[BUG] f29abc8d airgapped Harvester master build fails to render harvester dashboard


irishgordo:
### Describe the Bug

UI Fails to render in Airgapped Harvester Master Build:

```
curl -k https://10.115.16.193
<a href="/dashboard/">Found</a>.
```

### To Reproduce

1. airgap master build harvester
2. terminal ui shows things as live, but dashboard never materializes 

### Expected Behavior

Harvester dashboard to load

### Support Bundle for Troubleshooting

n/a

### Environment

- Harvester version: master

### Additional context

<img width="3825" height="493" alt="Image" src="https://github.com/user-attachments/assets/df1bf3b6-80fa-4921-be03-2a5513878214" />

### Workaround and Mitigation

n/a

bk201:
The issue is isolated to non-formal releases due to how we handle the `ui-source`'s `auto` setting. In these builds, the [backend attempts to pull the UI from an external index](https://github.com/harvester/harvester/blob/df34e6d612570b74b9742787a4bd58fa663149cd/pkg/server/ui/ui.go#L81-L88) before falling back to local assets.
- In my air-gapped environment, this request doesn't "fail" immediately; instead, it hangs for 9 minutes because the DNS server (configured in /etc/resolv.conf) is reachable, but the external route is not. This also answers why I can reach the dashboard after a while.
- I confirmed this by changing the nameserver to an unreachable IP (e.g., 8.8.8.8), which triggered an immediate failure and fallback.


Formal releases are unaffected because we prioritize bundled assets.

bk201:
@irishgordo FYI. The fix is merged to master.

irishgordo:
Thanks for the fix @bk201 
this looks good on the latest v1.9.0-dev build

I'll go ahead and close this out

<img width="3830" height="1896" alt="Image" src="https://github.com/user-attachments/assets/d9f16128-e987-4bfd-ad8e-f1ed23e93923" />

PR #10544, opened 2026-05-07
fix(backend): reduce dashboard load delay in air-gapped environments

bk201:
#### Problem:

Deploy harvester non-formal releases in an air-gapped environment, we can't load the dashboard immidately after the deploy is finished.

#### Solution:

Add a timeout
#### Related Issue(s):

https://github.com/harvester/harvester/issues/10540

#### Test plan:

* Deploy the harvester in an air-gapped environment; one node is OK.
  * You can configure nodes with a DNS server that can still be reached. But ensure the nodes can't reach the outside world with IP.
* After the cluster is ready, open the dashboard, and it should work.

In non-formal releases (e.g., master or local builds), the backend defaults to fetching the external UI index before falling back to local assets.
In certain air-gapped environments—specifically those with functioning DNS but no external routing—this request can hang for up to 10 minutes before timing out.
This PR introduces a 30-second timeout to the external index request to ensure a faster fallback to bundled assets when the network is restricted.

### DIFF
--- a/pkg/server/ui/ui.go
+++ b/pkg/server/ui/ui.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	responsewriter "github.com/rancher/apiserver/pkg/middleware"
 	"github.com/sirupsen/logrus"
@@ -17,6 +18,7 @@ import (
 
 var (
 	insecureClient = &http.Client{
+		Timeout: 30 * time.Second,
 		Transport: &http.Transport{
 			Proxy: http.ProxyFromEnvironment,
 			TLSClientConfig: &tls.Config{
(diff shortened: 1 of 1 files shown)


pohanhuang (dismissed):

Yu-Jack (dismissed):

pkg/server/ui/ui.go
```diff
@@ -17,6 +18,7 @@ import (
 
 var (
 	insecureClient = &http.Client{
```
w13915984028, line 20: https://github.com/harvester/harvester/blob/c3cffc1f48f4a1fa0ca7375db200dae392fcceed/pkg/server/ui/ui.go#L70-L73

https://github.com/harvester/harvester/blob/c3cffc1f48f4a1fa0ca7375db200dae392fcceed/pkg/server/ui/ui.go#L116

both of the caller do not log/loose the detailed failure, could you also help update them with related logrus info/debug, thanks.

bk201: Log messages added!

w13915984028 (approved):
LGTM, thanks for the kind fix.

Yu-Jack (approved):

Vicente-Cheng:
@mergifyio backport v1.8

Vicente-Cheng (approved):

PR #8547, opened 2025-06-25
charts: bump various v1.6 components

Vicente-Cheng:
#### Problem:
Bump various components charts for v1.6

#### Solution:
Bump various components charts for v1.6

PR #10550, opened 2026-05-08
bump: use v1.8 head and support-bundle-kit v0.0.85

wheatdog:
#### Problem:

For v1.8.1 sprint release, we need to make sure the image tag is v1.8-head

#### Solution:

Bump image tags

Similar to https://github.com/harvester/harvester/pull/8547
There will be a separate PR to bump dependency charts: https://github.com/harvester/harvester/pull/10551

### DIFF
--- a/deploy/charts/harvester/values.yaml
+++ b/deploy/charts/harvester/values.yaml
@@ -242,7 +242,7 @@ containers:
 
       ## Specify the tag of image.
       ##
-      tag: master-head
+      tag: v1.8-head
 
     ## Specify whether to run in HCI mode. This option enables additional controllers.
     ## defaults to "false", this will be enabled in ISO mode installation.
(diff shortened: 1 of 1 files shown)


deploy/charts/harvester/values.yaml
```diff
@@ -590,7 +590,7 @@ support-bundle-kit:
   image:
     imagePullPolicy: IfNotPresent
     repository: rancher/support-bundle-kit
-    tag: master-head
+    tag: v0.0.85
```
wheatdog, line 593: https://github.com/rancher/support-bundle-kit/releases/tag/v0.0.85

deploy/charts/harvester/values.yaml
```diff
@@ -242,7 +242,7 @@ containers:
 
       ## Specify the tag of image.
       ##
-      tag: master-head
+      tag: v1.8-head
```
w13915984028, line 245: not sure this trick could help:

```
# Define the anchor at the top
common:
    tag: &imageTag v1.8-head

frontend:
    image:
        tag: *imageTag

backend:
    image:
        tag: *imageTag

worker:
    image:
        tag: *imageTag
```

wheatdog: Yes, we can refine this along with the automatic release script.

wheatdog:
Will create a new PR later

wheatdog:
https://github.com/harvester/harvester/pull/10552

Issue #9135, opened 2025-09-15
[FEATURE] Implement a unified backup and restore interface framework


WebberHuang1118:
### Is your feature request related to a problem? Please describe (👍 if you like this request)

Implement a general interface to support multiple data backup and restore engines. This serves as preparation for [#8367](https://github.com/harvester/harvester/issues/8367), enabling VMs with third-party storage to perform remote backup and restore.

### Describe the solution you'd like

Implement a general interface to accommodate multiple data backup/restore engines

cheese:
KubeVirt is planning a CBT interface https://github.com/kubevirt/enhancements/issues/25

lanfon72:
Verified this feature has been applied and no regression founded.

Test Information
----
* Environment: **MegaRAC SP-X 3 nodes**
* Harvester Version: **v1.9.0-rc2**, **v1.9.0-rc3**
* **ui-source** Option: **Auto**

### Verify Steps
1. Install Harvester with any nodes
2. run full regression testing for backup and snapshot

PR #10557, opened 2026-05-12
feature: general framework for VM backup/restore engine

WebberHuang1118:
#### Problem:
#9135 

#### Solution:
General framework for VM backp/restore engine

#### Related Issue(s):
#9135 

#### Test plan:
**Prerequisite:**
* Build a Harvester `v1.8.0` cluster.
* Create a VM named `vm1`, start it, and write some data to it.
* Configure the backup target.
* Create a VM snapshot `vm1-s` from `vm1`.
* Create a VM backup `vm1-b` from `vm1`.
* Update Harvester and Harvester Webhook deployment to this PR

**Upgrade Case:**
* Restore `vm1-s` to a new VM named `vm1-s-r`. Verify that `vm1-s-r` can start and that the data integrity is correct.
* Restore `vm1-b` to a new VM named `vm1-b-r`. Verify that `vm1-b-r` can start and that the data integrity is correct.

**Recovery Case:**
* Set the backup target back to the default.
* Remove `vm1-b`.
* Reconnect the backup target with `Refresh Interval` set to 30 seconds.
* Verify that Harvester automatically recovers `vm1-b`.
* Restore `vm1-b` to a new VM named `vm1-b-r2` in a different namespace from `vm1`. Verify that `vm1-b-r2` can start and that the data integrity is correct.

**From Scratch:**
* Create a VM named `vm2`, start it, and write some data to it.
* Create a VM snapshot `vm2-s` from `vm2`.
* Create a VM backup `vm2-b` from `vm2`.
* Restore `vm2-s` to a new VM named `vm2-s-r`. Verify that `vm2-s-r` can start and that the data integrity is correct.
* Restore `vm2-b` to the existing VM `vm2` with the deletion policy. Verify that `vm2` can start and that the data integrity is correct.


pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1203 @@
...
+func (vmbo *vmbackupOperator) GetVolBackupPVCSpec(vb *harvesterv1.VolumeBackup) corev1.PersistentVolumeClaimSpec {
+	if vb == nil {
+		return corev1.PersistentVolumeClaimSpec{}
+	}
+	return vb.PersistentVolumeClaim.Spec
+}
+
+func (vmbo *vmbackupOperator) SetVolBackupName(vb *harvesterv1.VolumeBackup, name string) error {
+	if vb == nil {
+		return fmt.Errorf("volume backup is nil")
```
votdev: Make `volume backup is nil` a const. The error text is used several times across this file.

Maybe create a new error and make use of it here.
```
errVolBackupIsNil = errors.New("volume backup is nil")
```

pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1203 @@
...
+	})
+
+	_, err := vmbo.Update(old, newVMb)
+	return err
+}
+
+func (vmbo *vmbackupOperator) initStatus(vmb *harvesterv1.VirtualMachineBackup, vm *kubevirtv1.VirtualMachine) *harvesterv1.VirtualMachineBackup {
+	vmb.Status = harvesterv1.VirtualMachineBackupStatus{
+		ReadyToUse: ptr.To(false),
+		SourceUID:  &vm.UID,
```
votdev: To keep consistency, use `ptr.To` here as well.
```suggestion
		SourceUID:  ptr.To(vm.UID),
```

pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1203 @@
...
+		SourceSpec: &harvesterv1.VirtualMachineSourceSpec{
```
votdev:
```suggestion
		SourceSpec: ptr.To(harvesterv1.VirtualMachineSourceSpec{
```

pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1203 @@
...
+	}
+	return storageCapacity.Value(), nil
+}
+
+// buildVolumeBackup constructs a VolumeBackup object from PVC and PV information
+func buildVolumeBackup(vmbName, volName, pvcName string, pvc *corev1.PersistentVolumeClaim, pv *corev1.PersistentVolume, volSize int64) *harvesterv1.VolumeBackup {
+	volumeBackupName := fmt.Sprintf(volumeBackupNameFormat, vmbName, pvcName)
+
+	return &harvesterv1.VolumeBackup{
+		Name:          &volumeBackupName,
```
votdev:
```suggestion
		Name:          ptr.To(volumeBackupName),
```

pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1203 @@
...
+		return 0, fmt.Errorf("PV %s has no storage capacity defined", pv.Name)
+	}
+	return storageCapacity.Value(), nil
+}
+
+// buildVolumeBackup constructs a VolumeBackup object from PVC and PV information
+func buildVolumeBackup(vmbName, volName, pvcName string, pvc *corev1.PersistentVolumeClaim, pv *corev1.PersistentVolume, volSize int64) *harvesterv1.VolumeBackup {
+	volumeBackupName := fmt.Sprintf(volumeBackupNameFormat, vmbName, pvcName)
+
+	return &harvesterv1.VolumeBackup{
```
votdev: You are already using `ptr.To` in this helper function, so IMO you should use it everywhere to keep the code consistent.


```suggestion
	return ptr.To(harvesterv1.VolumeBackup{
```

pkg/backup/common/vshelper.go
```diff
@@ -0,0 +1,281 @@
...
+
+	// Extract PVC metadata
+	srcSCName := ptr.Deref(pvc.Spec.StorageClassName, "")
+	srcImageID := pvc.Annotations[util.AnnotationImageID]
+	srcProvisioner := util.GetProvisionedPVCProvisioner(pvc, h.scCache)
+
+	// Build VolumeSnapshot
+	vs := &snapshotv1.VolumeSnapshot{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            *h.vmbo.GetVolBackupName(vb),
```
votdev: `GetVolBackupName` can return `nil`. If not evaluated previously, this can cause a runtime error.
If the function parameter `vb` is considered valid and not `nil` when the `CreateVolumeSnapshotFromPVC` function is called, then we can, of course, ignore it.

pkg/backup/common/vshelper.go
```diff
@@ -0,0 +1,281 @@
...
+			Namespace:       h.vmbo.GetNamespace(vmb),
+			OwnerReferences: []metav1.OwnerReference{ownerRef},
+			Annotations:     make(map[string]string),
+		},
+		Spec: snapshotv1.VolumeSnapshotSpec{
+			Source: snapshotv1.VolumeSnapshotSource{
+				PersistentVolumeClaimName: &pvcName,
```
votdev:
```suggestion
				PersistentVolumeClaimName: ptr.To(pvcName),
```

pkg/controller/master/backup/backup.go
```diff
@@ -24,786 +22,320 @@ import (
...
+	vmBackupKindName = "VirtualMachineBackup"
 
-	backupTargetAnnotation       = "backup.harvesterhci.io/backup-target"
-	backupBucketNameAnnotation   = "backup.harvesterhci.io/bucket-name"
-	backupBucketRegionAnnotation = "backup.harvesterhci.io/bucket-region"
-
-	defaultFreezeDuration = 1 * time.Second
-	snapRevise            = "-snap-revise"
+	// engineRetryDelay is the delay before retrying volume backup creation when engine returns ErrRetryLater
+	engineRetryDelay = 5 * time.Second
```
votdev, line 53: Maybe choose a better variable name so that it is more clear that this is used for the `EnqueueAfter` call.

pkg/backup/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,460 @@
+package longhorn
```
votdev, line 1: Would it make sense to move the entire `longhorn` directory under `engine`? Since the code in `longhorn` is one of the possible engines, I think it makes sense from a thematic standpoint.

pkg/backup/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,89 @@
+package snapshot
```
votdev, line 1: It might also make sense to move this directory under `engine` if we want to create a subdirectory structure.

pkg/controller/master/backup/backup_metadata.go
```diff
@@ -645,97 +649,205 @@ func (h *MetadataHandler) checkExistingVMBackup(target *settings.BackupTarget) e
...
+			"cannot find volume in the backup target for a ready VMBackup, change the VMBackup to not ready",
+			fmt.Sprintf("cannot find volume %s in the backup target", pvName),
+			logrus.Fields{"volume": pvName},
+		)
+	}
+
+	if volumes[pvName].Backups[*lhBackupName] == nil {
+		return h.markVolumeNotReady(
+			vmb, vb,
+			"cannot find longhorn backup in the backup target for a ready VMBackup, change the VMBackup to not ready",
```
votdev: Proper nouns should be in uppercase.
```suggestion
			"cannot find Longhorn backup in the backup target for a ready VMBackup, change the VMBackup to not ready",
```

pkg/controller/master/backup/backup_status.go
```diff
@@ -127,61 +92,100 @@ func (h *Handler) resolveVolSnapshotRef(namespace string, controllerRef *metav1.
...
+	vmb, err := h.getVMBackupFromLHBackup(lhBackup)
+	if err != nil || vmb == nil {
+		return nil, err
+	}
+
+	if h.vmbo.GetBackupTarget(vmb) == nil {
+		return nil, nil
+	}
+
+	vsc, err := h.vscCache.Get(strings.Replace(lhBackup.Status.SnapshotName, "snapshot", "snapcontent", 1))
```
votdev: Maybe introduce a helper function for `strings.Replace(xxx, "snapshot", "snapcontent", 1)` because this code exists twice. As a nice bonus, you can also document what the string replacement is all about.

pkg/controller/master/backup/restore.go
```diff
@@ -327,461 +411,319 @@ func (h *RestoreHandler) PersistentVolumeClaimOnChange(_ string, pvc *corev1.Per
...
+	vmAnnotations := h.buildVMAnnotations(vmr, sourceSpec)
 
-	newVMSpecAnnotations, err := sanitizeVirtualMachineAnnotationsForRestore(restore, vmCpy.Spec.Template.ObjectMeta.Annotations)
+	specAnnotations, err := h.buildVMSpecAnnotations(vmr, sourceSpec)
 	if err != nil {
 		return nil, err
 	}
 
-	defaultRunStrategy := resolveRunStrategyFromRestoreAndBackup(restore, backup)
+	runStrategy := h.getDefaultRunStrategy(vmr, sourceSpec)
```
votdev: IMO we should keep the var name to indicate what it is used for.
```suggestion
	defaultRunStrategy := h.getDefaultRunStrategy(vmr, sourceSpec)
```

pkg/restore/pvchelper/pvchelper.go
```diff
@@ -0,0 +1,84 @@
+package pvchelper
```
votdev, line 1: Move the directory under `engine`?

pkg/restore/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,200 @@
+package snapshot
```
votdev, line 1: Move the directory under `engine`?

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,630 @@
...
+	return nil
+}
+
+// updateVolumeBackup updates the volume backup name with snap-revise suffix
+func (lre *LonghornRestoreEngine) updateVolumeBackup(vmb *harvesterv1.VirtualMachineBackup, vb *harvesterv1.VolumeBackup) error {
+	volName := lre.vmbo.GetVolBackupVolumeName(vb)
+	volBackupName := lre.vmbo.GetVolBackupName(vb)
+
+	if volBackupName == nil {
+		return fmt.Errorf("volume backup name is nil")
```
votdev: Introduce a dedicated error for this as this type of error is used across this PR. See https://github.com/harvester/harvester/pull/10557/changes#r3257002590

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,630 @@
...
+		return nil, fmt.Errorf("longhorn backup name is nil for volume backup")
+	}
+
+	lhBackup, err := lre.lhbackupCache.Get(util.LonghornSystemNamespaceName, *lhBackupName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Longhorn backup %s: %w", *lhBackupName, err)
+	}
+
+	if lhBackup.Status.VolumeName == "" {
+		return nil, fmt.Errorf("longhorn backup %s has empty volume name", lhBackup.Name)
```
votdev:
```suggestion
		return nil, fmt.Errorf("Longhorn backup %s has empty volume name", lhBackup.Name)
```
or
```suggestion
		return nil, fmt.Errorf("the Longhorn backup %s has empty volume name", lhBackup.Name)
```

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,630 @@
...
+	vmbo                common.VMBackupOperator
+	vmro                restorecommon.VMRestoreOperator
+	pvcCache            ctlcorev1.PersistentVolumeClaimCache
+	pvcClient           ctlcorev1.PersistentVolumeClaimClient
+	scCache             ctlstoragev1.StorageClassCache
+	vsCache             ctlsnapshotv1.VolumeSnapshotCache
+	vsClient            ctlsnapshotv1.VolumeSnapshotClient
+	vscCache            ctlsnapshotv1.VolumeSnapshotContentCache
+	vscClient           ctlsnapshotv1.VolumeSnapshotContentClient
+	lhbackupCache       ctllonghornv2.BackupCache
```
votdev:
```suggestion
	lhBackupCache       ctllonghornv2.BackupCache
```

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,630 @@
...
+	lhbackupVolumeCache ctllonghornv2.BackupVolumeCache
```
votdev:
```suggestion
	lhBackupVolumeCache ctllonghornv2.BackupVolumeCache
```

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,630 @@
...
+	lhengineCache       ctllonghornv2.EngineCache
```
votdev:
```suggestion
	lhEngineCache       ctllonghornv2.EngineCache
```

pkg/backup/common/operator.go
WebberHuang1118: Updated per suggestion, thanks

pkg/backup/common/operator.go
WebberHuang1118: Updated per suggestion, thanks

pkg/backup/common/operator.go
WebberHuang1118: Updated per suggestion, thanks

pkg/backup/common/operator.go
WebberHuang1118: Updated per suggestion, thanks

pkg/backup/common/operator.go
WebberHuang1118: Updated per suggestion, thanks

pkg/controller/master/backup/backup.go
WebberHuang1118: I'd prefer to keep engineRetryDelay as-is. The name conveys the why (engine asked for retry) and the what (a delay), which I think reads more clearly at the call site than naming it after the EnqueueAfter mechanism.

pkg/backup/common/vshelper.go
WebberHuang1118: Updated per suggestion, thanks

pkg/backup/common/vshelper.go
WebberHuang1118: Updated per suggestion, thanks

pkg/backup/engine/longhorn/longhorn.go
WebberHuang1118: Updated per suggestion, thanks

pkg/backup/engine/snapshot/snapshot.go
WebberHuang1118: Updated per suggestion, thanks

pkg/restore/pvchelper/pvchelper.go
WebberHuang1118: `pvchelper` is a helper used by those implementations, not an engine itself. I’d like to keep it as-is.

pkg/controller/master/backup/backup_metadata.go
WebberHuang1118: Updated per suggestion, thanks

pkg/controller/master/backup/backup_status.go
WebberHuang1118: Updated per suggestion, thanks

pkg/controller/master/backup/restore.go
WebberHuang1118: Updated per suggestion, thanks

pkg/restore/engine/snapshot/snapshot.go
WebberHuang1118: Updated per suggestion, thanks

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Updated per suggestion, thanks

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Updated per suggestion, thanks

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Updated per suggestion, thanks

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Updated per suggestion, thanks

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Updated per suggestion, thanks

votdev (dismissed):
LGTM, thx

votdev (dismissed):

pkg/apis/harvesterhci.io/v1beta1/backup.go
brandboat, line 97: Is backupType immutable? I'm a bit worry if a VMBackup changes from backup to snapshot, existing status and artifacts may be interpreted by the wrong engine.

pkg/restore/engine/engine.go
```diff
@@ -0,0 +1,24 @@
...
+
+var (
+	ErrRetryLater = errors.New("retry later error")
+)
+
+// RestoreEngine defines the interface for restoring VM volumes
+type RestoreEngine interface {
+	// Reconcile ensures the per-volume restore at volIndex matches desired state:
+	// it creates the restored PVC if missing, otherwise checks status. Idempotent.
+	Reconcile(vmr *harvesterv1.VirtualMachineRestore, vmb *harvesterv1.VirtualMachineBackup, volIndex int) error
```
brandboat, line 18: What about

```
Reconcile(vmr *harvesterv1.VirtualMachineRestore, vmb *harvesterv1.VirtualMachineBackup, vb *VolumeBackup) error
```

pkg/restore/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	pvcName := sre.vmro.GetVolRestorePVCName(vr)
+	namespace := sre.vmro.GetNamespace(vmr)
+
+	// Check if PVC already exists
+	pvc, err := sre.pvcCache.Get(namespace, pvcName)
+	if apierrors.IsNotFound(err) {
+		// Create PVC from VolumeSnapshot
+		vb := sre.vmbo.GetVolBackup(vmb, volIndex)
```
brandboat, line 61: We are assuming that vmr.Status.VolumeRestores[i] always maps directly to vmb.Status.VolumeBackups[i]. Is that guaranteed to be true? Relying on matching indices feels a bit risky. Why not match them by VolumeBackupName instead?

pkg/restore/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,200 @@
...
+		pvcCache:  pvcCache,
+		pvcClient: pvcClient,
+		vsCache:   snapshotCache,
+	}
+}
+
+func (sre *SnapshotRestoreEngine) Reconcile(
+	vmr *harvesterv1.VirtualMachineRestore,
+	vmb *harvesterv1.VirtualMachineBackup,
+	volIndex int,
```
brandboat, line 47: Sorry I'm not a big fan of this, what about passing VolumeRestore and VolumeBackup?
```
type RestoreEngine interface {
    Reconcile(
        vmr *VirtualMachineRestore,
        vmb *VirtualMachineBackup,
        vr *VolumeRestore,
        vb *VolumeBackup,
    ) error
}
```

pkg/backup/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,89 @@
...
+		vmbo:     vmbo,
+		vsHelper: common.NewVolumeSnapshotHelper(vsCache, vsClient, nil, nil, vmbo, pvcCache, scCache),
+		pvcCache: pvcCache,
+		scCache:  scCache,
+	}
+}
+
+func (se *SnapshotEngine) Reconcile(
+	vmb *harvesterv1.VirtualMachineBackup,
+	volIndex int,
```
brandboat, line 42: What about VolumeBackup ?

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,630 @@
...
+		volumeCache:         volumeCache,
+		vmBackupController:  vmBackupController,
+		vmBackupCache:       vmBackupCache,
+	}
+}
+
+func (lre *LonghornRestoreEngine) Reconcile(
+	vmr *harvesterv1.VirtualMachineRestore,
+	vmb *harvesterv1.VirtualMachineBackup,
+	volIndex int,
```
brandboat, line 112: ditto.

pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1220 @@
...
+	cache ctlharvesterv1.VirtualMachineBackupCache,
+	vsClassCache ctlsnapshotv1.VolumeSnapshotClassCache,
+	vmCache ctlkubevirtv1.VirtualMachineCache,
+	vmiCache ctlkubevirtv1.VirtualMachineInstanceCache,
+	pvcCache ctlcorev1.PersistentVolumeClaimCache,
+	pvCache ctlcorev1.PersistentVolumeCache,
+	secretCache ctlcorev1.SecretCache,
+	virtSubresourceRestClient rest.Interface,
+) VMBackupOperator {
+	return &vmbackupOperator{
```
Yu-Jack: The more parameters you have, the longer caller should append.

I suggest using builder pattern to initilize the operator.

Try `common.GetVMBackupOperatorBuilder().SetVMCache(vmCache).SetPvcCache(pvcCache)`. Then, you can just ignore other parameters you don't need.


For example, there are two different used operators in your codes. With builder pattern, it would be easy to use this function.

```go
vmbo := common.GetVMBackupOperator(vmBackups,
		vmBackups.Cache(),
		nil,
		vms.Cache(),
		nil,
		nil,
		nil,
		secrets.Cache(),
		nil)

vmbo := backupcommon.GetVMBackupOperator(
		nil, // client not needed for restore operations
		controllers.vmbs.Cache(),
		controllers.vsClasses.Cache(),
		controllers.vms.Cache(),
		controllers.vmis.Cache(),
		controllers.pvcs.Cache(),
		controllers.pvs.Cache(),
		controllers.secrets.Cache(),
		nil, // virtSubresourceRestClient not needed
	)
```

pkg/webhook/indexeres/indexer.go
```diff
@@ -94,7 +94,8 @@ func vmBackupSnapshotByPVCNamespaceAndName(obj *harvesterv1.VirtualMachineBackup
 }
 
 func vmBackupByIsProgressing(obj *harvesterv1.VirtualMachineBackup) ([]string, error) {
-	isProgressingStr := strconv.FormatBool(backup.IsBackupProgressing(obj))
+	vmbo := backupcommon.GetVMBackupOperator(nil, nil, nil, nil, nil, nil, nil, nil, nil)
+	isProgressingStr := strconv.FormatBool(vmbo.IsProcessing(obj))
```
Yu-Jack: Another point related to `VMBackupOperator` interface.

It includes too many responsibilities. For example, here you just use `IsProcessing` to check, but you pass all nul parameters to the operator and don't use other functions.

I think current `VMBackupOperator` is more like a **god class**.

However, you can reverse the whole logic by:

1. Reduce the responsibilities of `VMBackupOperator` interface.
2. Create another suitable interface for this behavior
3. Inject this new interface to original operator 


Just an example.

```go
type VMBackupStatusChecker interface {
    IsProcessing(vmb *harvesterv1.VirtualMachineBackup) bool
}
```

When you need it in your original VMBackupOperator, you can set it with builder pattern or other ways like

```go
backupcommon.GetVMBackupOperator().SetStatusChecker(newVMBackupStatusChecker())
```

The whole picture will look like:

```go
// VMBackupStatusChecker handles pure in-memory status checks.
// Implementations require no Kubernetes client dependencies.
type VMBackupStatusChecker interface {
    IsReady(vmb *harvesterv1.VirtualMachineBackup) bool
    IsMissingStatus(vmb *harvesterv1.VirtualMachineBackup) bool
    IsProcessing(vmb *harvesterv1.VirtualMachineBackup) bool
    IsErrMsgSynced(vmb *harvesterv1.VirtualMachineBackup, errMsg string) bool
    IsTargetConsistent(vmb *harvesterv1.VirtualMachineBackup, target *settings.BackupTarget) bool
    IsTransitToNonReady(vmb *harvesterv1.VirtualMachineBackup) bool
}

type vmBackupStatusChecker struct{}

func NewVMBackupStatusChecker() VMBackupStatusChecker {
    return &vmBackupStatusChecker{}
}

func (c *vmBackupStatusChecker) IsProcessing(vmb *harvesterv1.VirtualMachineBackup) bool {
    return vmb.Status.Error == nil &&
        (vmb.Status.ReadyToUse == nil || !*vmb.Status.ReadyToUse)
}


type vmbackupOperator struct {
    VMBackupStatusChecker                          // embedded — delegates Is* calls
    client       ctlharvesterv1.VirtualMachineBackupClient
    cache        ctlharvesterv1.VirtualMachineBackupCache
    // ...
}

func GetVMBackupOperator(
    client ctlharvesterv1.VirtualMachineBackupClient,
    // ... other deps
) VMBackupOperator {
    return &vmbackupOperator{
        VMBackupStatusChecker: NewVMBackupStatusChecker(),
        client:                client,
        // ...
    }
}

// Or 

func GetVMBackupOperator(...).SetStatusChecker(c VMBackupStatusChecker) VMBackupOperator

// In indexr
func vmBackupByIsProgressing(obj *harvesterv1.VirtualMachineBackup) ([]string, error) {
    checker := backupcommon.NewVMBackupStatusChecker()
    return []string{strconv.FormatBool(checker.IsProcessing(obj))}, nil
}
```

Yu-Jack (commented):
Thank for managing this refactor effort. I have few suggestions.

pkg/controller/master/backup/backup.go
```diff
@@ -878,234 +410,75 @@ func (h *Handler) uploadVMBackupMetadata(vmBackup *harvesterv1.VirtualMachineBac
...
 		if err != nil {
-			if !apierrors.IsNotFound(err) {
-				return err
-			}
+			logrus.WithError(err).WithFields(logrus.Fields{
+				"vmBackup":     h.vmbo.GetName(vmb),
+				"volumeBackup": *h.vmbo.GetVolBackupName(&vb),
+				"namespace":    h.vmbo.GetNamespace(vmb),
+				"backupType":   h.vmbo.GetType(vmb),
+			}).Warn("failed to force delete volume backup")
```
brandboat, line 471: The original implementation keep retrying delete volume backups if there is any error raise up, but in this patch we remove this behavior and show warning log instead, is this intentional?

pkg/restore/engine/engine.go
WebberHuang1118: Because in most cases, the main control flow calls `Reconcile`, such as in [backup.go#L323-L344](https://github.com/WebberHuang1118/harvester/blob/58635fa675c0f8d26aa9f42483ef9859ff8b61ed/pkg/controller/master/backup/backup.go#L323-L344) and [restore.go#L497-L508](https://github.com/WebberHuang1118/harvester/blob/4c96b931ecd860ba4f59e405f13b168b9dd80ba1/pkg/controller/master/backup/restore.go#L497-L508).

There is a loop that traverses `VolumeBackups`, which is a slice in `VirtualMachineBackupStatus`. During the loop, if we pass `vb` directly, it is only a copy of the item in the `vmb.Status.VolumeBackups` slice. Any operation on it will not be reflected in the actual item in the slice.

That is why we pass the index instead, so the implementation can operate on the actual item in the slice while keeping the preparation before passing it to `Reconcile` less complex.

pkg/restore/engine/snapshot/snapshot.go
WebberHuang1118: This is legacy behavior from day one.

I agree with your point, but this part has been working fine so far, and the purpose of this PR is to build a backup and restore framework as preparation for third-party storage support. Therefore, I’d like to keep the changes minimal if possible.

pkg/restore/engine/snapshot/snapshot.go
WebberHuang1118: Ditto: [harvester/harvester#10557 comment](https://github.com/harvester/harvester/pull/10557#discussion_r3370646387).

pkg/backup/engine/snapshot/snapshot.go
WebberHuang1118: Ditto: [harvester/harvester#10557 comment](https://github.com/harvester/harvester/pull/10557#discussion_r3370646387).

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Ditto: [harvester/harvester#10557 comment](https://github.com/harvester/harvester/pull/10557#discussion_r3370646387).

pkg/backup/common/operator.go
votdev: The Functional Options Pattern would, of course, be even better, but we don't use it in Harvester.

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,744 @@
...
+			continue
+		}
+
+		currentVolName := lre.vmbo.GetVolBackupVolumeName(currentVb)
+		if currentVolName != volName {
+			continue
+		}
+
+		// Found the matching volume backup, update its name using the interface
+		newName := *volBackupName + snapRevise
```
brandboat: nit: we can put this line before the entering the for-loop `#L318` to avoid redundant variable allocations.

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,744 @@
...
+	return fmt.Errorf("vmbackup %s/%s volumebackup %s has no matched entry %v",
+		lre.vmbo.GetNamespace(vmb), lre.vmbo.GetName(vmb), *volBackupName, volName)
+}
+
+func (lre *LonghornRestoreEngine) createVolumeSnapshotForRestore(
+	vmr *harvesterv1.VirtualMachineRestore,
+	vb *harvesterv1.VolumeBackup,
+) (string, error) {
+	vsName := lre.constructVolumeSnapshotName(vmr, vb)
+	vscName := lre.constructVolumeSnapshotContentName(vmr, vb)
```
brandboat: nit: We can put this line above the VolumeSnapshotContent check (before L365) since we might really need it if VolumeSnapshot already exists

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,744 @@
...
+	vs, err := lre.vsCache.Get(lre.vmro.GetNamespace(vmr), vsName)
+	if err == nil {
+		return vs.Name, nil
+	}
+	if !apierrors.IsNotFound(err) {
+		return "", err
+	}
+
+	// Create VolumeSnapshotContent first
+	vsc, err := lre.createVolumeSnapshotContent(vmr, vb, vscName, vsName)
```
brandboat, line 374: Do we need to check if VolumeSnapshotContent exists before creating a new one?

pkg/controller/master/backup/restore.go
```diff
@@ -394,469 +450,216 @@ func (h *RestoreHandler) VMOnChange(_ string, vm *kubevirtv1.VirtualMachine) (*k
...
-}
-
-func buildVolumeClaimTemplateAnnotations(pvc *corev1.PersistentVolumeClaim) map[string]string {
-	if val, ok := pvc.Annotations[util.AnnotationImageID]; ok {
-		return map[string]string{util.AnnotationImageID: val}
-	}
-	return nil
+	vm.Annotations[lastRestoreAnnotation] = h.vmro.GetRestoreID(vmr)
+	vm.Annotations[restorecommon.RestoreNameAnnotation] = h.vmro.GetName(vmr)
+	delete(vm.Annotations, util.AnnotationVolumeClaimTemplates)
```
brandboat: Correct me if I'm wrong, looks like this fix https://github.com/harvester/harvester/commit/462ac0ce0a93dc57dcce1e65b3b2cd9b8259d24b
 is not included in this patch?

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,744 @@
...
+		if !rs.IsRestoring {
+			continue
+		}
+
+		replicaRestoreProgressSum += rs.Progress
+	}
+
+	if numReplica > 0 {
+		return int64(replicaRestoreProgressSum / numReplica), nil
+	}
```
brandboat, line 554: If one replica finishes and Longhorn reports IsRestoring=false, its progress contributes 0 to the average. That means progress can go backward, for example:

```
replica A: IsRestoring=false, Progress=100
replica B: IsRestoring=true,  Progress=80
average returned: 80 / 2 = 40
```

pkg/controller/master/backup/restore.go
```diff
@@ -866,559 +669,156 @@ func getNewVMAnnotations(restore *harvesterv1.VirtualMachineRestore, sourceAnnot
...
-				Kind:     volumeSnapshotKindName,
-				Name:     dataSourceName,
-			},
-			Resources:        volumeBackup.PersistentVolumeClaim.Spec.Resources,
-			StorageClassName: volumeBackup.PersistentVolumeClaim.Spec.StorageClassName,
-			VolumeMode:       volumeBackup.PersistentVolumeClaim.Spec.VolumeMode,
-		},
-	})
-	return err
+	// 4. Cleanup and complete
```
brandboat: nit:
```suggestion
	// Cleanup and complete
```

pkg/controller/master/backup/restore.go
```diff
@@ -866,559 +669,156 @@ func getNewVMAnnotations(restore *harvesterv1.VirtualMachineRestore, sourceAnnot
...
-				return nil
-			}
-		}
-		return h.restClient.Put().Namespace(vm.Namespace).Resource("virtualmachines").SubResource("start").Name(vm.Name).Do(h.context).Error()
-	case kubevirtv1.RunStrategyHalted:
-		return h.restClient.Put().Namespace(vm.Namespace).Resource("virtualmachines").SubResource("start").Name(vm.Name).Do(h.context).Error()
-	default:
-		// skip
-	}
+	h.vmrController.EnqueueAfter(vmr.Namespace, vmr.Name, restoreProgressPoll)
```
brandboat, line 850: Just a heads-up: the Longhorn engine implementation already watches for LonghornEngine object changes and enqueues the VMRestore to update progress here:
```go
	lre.lhEngineController.OnChange(ctx, lhEngineWatcherName, func(_ string, lhEngine *lhv1beta2.Engine) (*lhv1beta2.Engine, error) {
		return lre.onLHEngineChanged(lhEngine, enqueueVMRestore)
	})
```
We might want to remove this polling interval or find a clean way to refactor out the duplicate enqueue behavior from lhEngineController.

pkg/controller/master/backup/restore.go
```diff
@@ -866,559 +669,156 @@ func getNewVMAnnotations(restore *harvesterv1.VirtualMachineRestore, sourceAnnot
...
+	// Start VM before checking status
+	if err := h.vmro.StartVM(h.context, vm); err != nil {
+		return h.vmro.UpdateError(vmr, fmt.Errorf("failed to start vm, err:%s", err.Error()))
 	}
 
-	if err := h.deleteOldPVC(restoreCpy, vm); err != nil {
-		return h.updateStatusError(vmRestore, fmt.Errorf("error cleaning up, err:%s", err.Error()), false)
+	if vm.Status.Ready {
+		return nil
 	}
```
brandboat: Maybe we can put this before StartVM, then we don’t need to perform a duplicate VM start if VM is already in Ready state.

pkg/restore/engine/engine.go
brandboat: I would lean toward passing the exact struct instead of relying on the index, but that's just my preference!

pkg/apis/harvesterhci.io/v1beta1/backup.go
```diff
@@ -63,8 +95,9 @@ type VirtualMachineBackupSpec struct {
 	Source corev1.TypedLocalObjectReference `json:"source"`
 
 	// +kubebuilder:default:="backup"
-	// +kubebuilder:validation:Enum=backup;snapshot
+	// +kubebuilder:validation:Enum=backup;snapshot;restic;kopia
```
brandboat: `;restic;kopia`

This seems to be a future implementation, are we going to allow these 2 in this patch?

pkg/backup/common/operator.go
WebberHuang1118: Updated with the builder pattern, thanks.

pkg/webhook/indexeres/indexer.go
WebberHuang1118: Updated with the builder pattern, thanks.

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/controller/master/backup/restore.go
WebberHuang1118: Good catch, thanks.

WebberHuang1118: Updated, thanks.

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: Good catch, updated, thanks.

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: It is not needed, as this creates the VSC on demand for cross-namespace VM restore. The VSC resource is dedicated to the specific `VMRestore`.

pkg/controller/master/backup/backup.go
WebberHuang1118:
```
func (h *Handler) forceDeleteVolBackups(vmb *harvesterv1.VirtualMachineBackup) error {
	backupEngine := h.getBackupEngine(vmb)

	for i, vb := range h.vmbo.GetVolBackups(vmb) {
		if h.vmbo.GetVolBackupName(&vb) == nil {
			continue
		}

		err := backupEngine.ForceDelete(vmb, i)
		if err != nil {
				"vmBackup":     h.vmbo.GetName(vmb),
				"volumeBackup": *h.vmbo.GetVolBackupName(&vb),
				"namespace":    h.vmbo.GetNamespace(vmb),
				"backupType":   h.vmbo.GetType(vmb),
			}).Warn("failed to force delete volume backup")
			return err
		}
	}
	return nil
}
```

The logic here will log and return the error immediately if any error occurs, so the next reconcile can retry.

pkg/controller/master/backup/restore.go
WebberHuang1118: This is a preparation for third-party storage restore progress reporting, as this scenario cannot rely on the LH engine.

pkg/controller/master/backup/restore.go
WebberHuang1118: updated per suggestion, thanks.

pkg/apis/harvesterhci.io/v1beta1/backup.go
WebberHuang1118: Good catch, updated, thanks.

pkg/restore/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,205 @@
...
+
+	pvcName := sre.vmro.GetVolRestorePVCName(vr)
+	namespace := sre.vmro.GetNamespace(vmr)
+
+	// Check if PVC already exists
+	pvc, err := sre.pvcCache.Get(namespace, pvcName)
+	if apierrors.IsNotFound(err) {
+		// Create PVC from VolumeSnapshot
+		vb := sre.vmbo.GetVolBackup(vmb, volIndex)
+		return sre.createPVCFromSnapshot(vmr, vr, vb)
```
votdev: When a PVC is not found, [Reconcile](https://github.com/harvester/harvester/blob/9b3a486bf2321d23a82c54b583c93ad63e886cc7/pkg/controller/master/backup/restore.go#L499) returns `nil` (instead of `engine.ErrRetryLater`) after calling `createPVCFromSnapshot`, which falsely signals `isVolumesReady = true` and triggers VM start or restore finalization before the PVC is actually `ClaimBound`.

```suggestion
        if err := sre.createPVCFromSnapshot(vmr, vr, vb); err != nil {
			return err
		}
		return engine.ErrRetryLater
```

pkg/restore/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	pvcName := sre.vmro.GetVolRestorePVCName(vr)
+	namespace := sre.vmro.GetNamespace(vmr)
+
+	// Check if PVC already exists
+	pvc, err := sre.pvcCache.Get(namespace, pvcName)
+	if apierrors.IsNotFound(err) {
+		// Create PVC from VolumeSnapshot
+		vb := sre.vmbo.GetVolBackup(vmb, volIndex)
```
votdev, line 61: `vb` can be nil. Is this handled in the later `createPVCFromSnapshot` call?

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,747 @@
...
+
+	pvcName := lre.vmro.GetVolRestorePVCName(vr)
+	namespace := lre.vmro.GetNamespace(vmr)
+
+	// Check if PVC already exists
+	pvc, err := lre.pvcCache.Get(namespace, pvcName)
+	if apierrors.IsNotFound(err) {
+		vb := lre.vmbo.GetVolBackup(vmb, volIndex)
+		return lre.createPVCFromSnapshot(vmr, vmb, vr, vb)
```
votdev: When a PVC is not found, `Reconcile` returns `nil` (instead of `engine.ErrRetryLater`) after calling `createPVCFromSnapshot`, which falsely signals `isVolumesReady = true` and triggers VM start or restore finalization before the PVC is actually `ClaimBound`.

```suggestion
		if err := lre.createPVCFromSnapshot(vmr, vmb, vr, vb); err != nil {
			return err
		}
		return engine.ErrRetryLater
```

pkg/restore/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,747 @@
...
+		return fmt.Errorf("volume restore at index %d not found", volIndex)
+	}
+
+	pvcName := lre.vmro.GetVolRestorePVCName(vr)
+	namespace := lre.vmro.GetNamespace(vmr)
+
+	// Check if PVC already exists
+	pvc, err := lre.pvcCache.Get(namespace, pvcName)
+	if apierrors.IsNotFound(err) {
+		vb := lre.vmbo.GetVolBackup(vmb, volIndex)
```
votdev, line 125: `vb` can be `nil`.

pkg/backup/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+	vsClassMap map[string]snapshotv1.VolumeSnapshotClass,
+) error {
+	logrus.Infof("SnapshotEngine Reconcile called for VMBackup %s/%s volume index %d",
+		se.vmbo.GetNamespace(vmb), se.vmbo.GetName(vmb), volIndex)
+
+	vb := se.vmbo.GetVolBackup(vmb, volIndex)
```
votdev, line 48: Can be `nil`. If `GetVolBackupName` is handling `nil` results correctly, we can ignore the `nil` result of `GetVolBackup`.

pkg/backup/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+	vsName := *se.vmbo.GetVolBackupName(vb)
```
votdev, line 53: `getVolBackupName` returns `nil` if `vb` is `nil`. So we NEED to check the returned value before dereferencing.

Maybe use this code or something similar, e.g. throw an error immediately.
```suggestion
	vsName := ""
	if name := se.vmbo.GetVolBackupName(vb); name != nil {
		vsName = *name
	}
```

pkg/restore/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (sre *SnapshotRestoreEngine) createPVC(
+	vmr *harvesterv1.VirtualMachineRestore,
+	vr *harvesterv1.VolumeRestore,
+	vb *harvesterv1.VolumeBackup,
+	vsName string,
+) error {
+	pvc := sre.buildPVC(vmr, vr, vb, vsName)
+	_, err := sre.pvcClient.Create(pvc)
+	return err
```
votdev, line 164: Does the error give enough details in case `Create` fails? Otherwise we should use something like this:

```suggestion
	return fmt.Errorf("failed to create PVC %s/%s: %w", namespace, pvcName, err)
```

pkg/restore/engine/snapshot/snapshot.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func (sre *SnapshotRestoreEngine) checkPVCStatus(pvc *corev1.PersistentVolumeClaim) error {
+	return pvchelper.CheckPVCStatus(pvc)
+}
+
+func (sre *SnapshotRestoreEngine) UpdateProgress(vr *harvesterv1.VolumeRestore) (int64, error) {
+	// For local snapshots, we don't have detailed progress tracking
+	// Return 100 if it's complete (handled by PVC status check in Create)
+	return 100, nil
```
votdev, line 203: What about having a simple progress update based on the `status.phase`?

```suggestion
	if vr == nil {
		return 0, fmt.Errorf("volume restore is nil")
	}

	pvcName := sre.vmro.GetVolRestorePVCName(vr)
	if pvcName == "" {
		return 0, nil
	}

	namespace := sre.vmro.GetVolRestorePVCNamespace(vr)
	if namespace == "" {
		return 0, nil
	}

	pvc, err := sre.pvcCache.Get(namespace, pvcName)
	if apierrors.IsNotFound(err) {
		// PVC creation may still be in-flight for this volume.
		return 0, nil
	}
	if err != nil {
		return 0, fmt.Errorf("failed to get PVC %s/%s while updating progress: %w", namespace, pvcName, err)
	}

	switch pvc.Status.Phase {
	case corev1.ClaimBound:
		return 100, nil
	case corev1.ClaimPending:
		return 50, nil
	default:
		return 0, nil
	}
```

pkg/backup/engine/longhorn/longhorn.go
```diff
@@ -0,0 +1,595 @@
...
+			return "", fmt.Errorf("failed to set volume backup ready to use: %w", err)
+		}
+		return msg, nil
+	}
+
+	if vols[pvName] == nil {
+		return handleMissing(fmt.Sprintf("cannot find volume %s in the backup target", pvName))
+	}
+
+	lhBackupName := *le.vmbo.GetVolBackupLHBackupName(vb)
```
votdev, line 291: `GetVolBackupLHBackupName` can return `nil`. This will cause a crash when `nil` is dereferenced.

pkg/backup/engine/snapshot/snapshot.go
votdev, line 53: `GetVolBackupName` can return `nil`. Dereferencing `nil` without a check will cause a crash.

pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1292 @@
...
+
+	newVMb.Annotations[util.AnnotationSnapshotFreezeFS] = strconv.FormatBool(value)
+	return vmbo.Update(old, newVMb)
+}
+
+func (vmbo *vmbackupOperator) withFreezeFSAnnotation(vmb *harvesterv1.VirtualMachineBackup) (*harvesterv1.VirtualMachineBackup, *kubevirtv1.VirtualMachineInstance, error) {
+	sourceVMI, err := vmbo.getBackupSourceInstance(vmb)
+	if apierrors.IsNotFound(err) {
+		updatedVMb, updateErr := vmbo.saveFreezeFSAnnotation(vmb, false)
+		return updatedVMb, nil, updateErr
```
votdev, line 1231: The original error from `getBackupSourceInstance` is hidden and not forwarded.

pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1292 @@
...
+
+func (a *vmbackupReader) GetVolBackup(vmb *harvesterv1.VirtualMachineBackup, index int) *harvesterv1.VolumeBackup {
+	if index < 0 || index >= len(vmb.Status.VolumeBackups) {
+		return nil
+	}
+	return &vmb.Status.VolumeBackups[index]
+}
+
+func (a *vmbackupReader) GetSanitizeVolBackups(vmb *harvesterv1.VirtualMachineBackup) []harvesterv1.VolumeBackup {
+	sanitizeVolBackups := vmb.Status.VolumeBackups
```
votdev, line 226: Add a note that this mutation is wanted, otherwise every AI will report that issue as a problem.

pkg/restore/engine/snapshot/snapshot.go
WebberHuang1118: Good catch, thanks.

pkg/restore/engine/snapshot/snapshot.go
WebberHuang1118: updated, thanks.

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: updated, thanks.

pkg/restore/engine/longhorn/longhorn.go
WebberHuang1118: updated, thanks.

pkg/backup/common/operator.go
```diff
@@ -0,0 +1,1292 @@
...
+
+func (a *vmbackupReader) GetVolBackups(vmb *harvesterv1.VirtualMachineBackup) []harvesterv1.VolumeBackup {
+	return vmb.Status.VolumeBackups
+}
+
+func (a *vmbackupReader) GetVolBackup(vmb *harvesterv1.VirtualMachineBackup, index int) *harvesterv1.VolumeBackup {
+	if index < 0 || index >= len(vmb.Status.VolumeBackups) {
+		return nil
+	}
+	return &vmb.Status.VolumeBackups[index]
```
votdev, line 220: Why do we get a reference to the cached object here, whereas `GetSourceSpec` returns a deep copy?

pkg/backup/engine/snapshot/snapshot.go
WebberHuang1118: good catch, thanks.

pkg/backup/engine/snapshot/snapshot.go
WebberHuang1118: updated, thanks.

pkg/restore/engine/snapshot/snapshot.go
WebberHuang1118: The GUI does not actually surface VM snapshot restore progress, so I think we can keep it as-is.

pkg/restore/engine/snapshot/snapshot.go
WebberHuang1118: It surfaces the API error, which should provide enough detail here. For now, I think we may not need to wrap the source error with additional context.

pkg/backup/engine/longhorn/longhorn.go
WebberHuang1118: If the source `vb` is not `nil`, `GetVolBackupLHBackupName` will not return `nil`. Therefore, I added the `nil` handling to the source `vb` retrieval logic: [longhorn.go#L352-L355](https://github.com/WebberHuang1118/harvester/blob/f811a7cf12edd62d2c934af6f708dc456cd2aef6/pkg/backup/engine/longhorn/longhorn.go#L352-L355).

pkg/backup/common/operator.go
WebberHuang1118: If `apierrors.IsNotFound(err)` is true, it means the VM is not running. Therefore, we do not need to handle the not-found error and can focus only on the result of saving the annotation.

pkg/backup/common/operator.go
WebberHuang1118: updated, thanks.

pkg/backup/engine/snapshot/snapshot.go
WebberHuang1118: Ditto as [harvester/harvester#10557 comment](https://github.com/harvester/harvester/pull/10557#discussion_r3395007706). I added handling for the case where `vb` is `nil`.

pkg/backup/common/operator.go
WebberHuang1118: Because `GetSourceSpec` returns `VirtualMachineSourceSpec`, which contains multiple nested pointers, directly referencing the source object in [`MapVolumesToRestoredPVCs`](https://github.com/WebberHuang1118/harvester/blob/f811a7cf12edd62d2c934af6f708dc456cd2aef6/pkg/restore/common/operator.go) would mutate the source object.

This change would then affect all controllers that reference the same object.

votdev (dismissed):

pkg/webhook/resources/virtualmachinebackup/validator.go
```diff
@@ -197,18 +203,23 @@ func (v *virtualMachineBackupValidator) Update(_ *types.Request, oldObj runtime.
...
+		if v.vmrr.IsDeleting(vmRestore) || !v.vmrr.HasStatus(vmRestore) {
 			continue
 		}
-		if v1beta1.BackupConditionProgressing.IsTrue(vmRestore) {
-			return werror.NewBadRequest(fmt.Sprintf("vmrestore %s/%s is in progress", vmRestore.Namespace, vmRestore.Name))
+
+		if v.vmrr.IsProgressing(vmRestore) {
+			restoreName := v.vmrr.GetName(vmRestore)
+			restoreNamespace := v.vmrr.GetNamespace(vmRestore)
+			return werror.NewBadRequest(fmt.Sprintf("vmrestore %s/%s is in progress", restoreNamespace, restoreName))
```
brandboat, line 225: The IsProgressing only check if vmr complete status is false/nil
```go
func (c *vmrestoreReader) IsProgressing(vmr *harvesterv1.VirtualMachineRestore) bool {
	return vmr.Status.Complete == nil || !*vmr.Status.Complete
}
```
While in SetErrorCondition only flips the conditions to error and does not set Complete=true
```go
func (vmro *vmrestoreOperator) SetErrorCondition(vmr *harvesterv1.VirtualMachineRestore, err error) *harvesterv1.VirtualMachineRestore {
	vmr = vmro.applyCondition(vmr, newProgressingCondition(corev1.ConditionFalse, "Error", err.Error()))
	vmr = vmro.applyCondition(vmr, newReadyCondition(corev1.ConditionFalse, "Error", err.Error()))
	return vmr
}
```
That means a failed restore remains “progressing” forever. The validator here blocks VMBackup deletion.

Similar thing could also happen in `HasActiveRestore`.
In virtualmachinerestore validator.go, the failed vmrestore makes the validator blocks future restores on the same VM 
```go
func (v *restoreValidator) validateNoInProgressRestore(vm *kubevirtv1.VirtualMachine) error {
	hasInProgress, err := webhookutil.HasActiveRestore(v.vmRestore, v.vmrr, vm.Namespace, vm.Name)
	if err != nil {
		return werror.NewInternalError(fmt.Sprintf("Failed to get the VM-related restores, err: %+v", err))
	}

	if hasInProgress {
		return werror.NewInvalidError(fmt.Sprintf("Please wait for the previous VM restore on the %s/%s to be complete first.", vm.Namespace, vm.Name), fieldTargetName)
	}

	return nil
}
```

pkg/controller/master/backup/restore.go
```diff
@@ -866,559 +754,156 @@ func getNewVMAnnotations(restore *harvesterv1.VirtualMachineRestore, sourceAnnot
...
-// VMBackup controller to create new VolumeSnapshot/VolumeSnapshotContent
-// and link VMBackup to these new resources
-func (h *RestoreHandler) reviseVolumeSnapshot(vb harvesterv1.VolumeBackup, vr *harvesterv1.VirtualMachineRestore) error {
-	vmBackup, err := h.backupCache.Get(vr.Spec.VirtualMachineBackupNamespace, vr.Spec.VirtualMachineBackupName)
-	if err != nil {
-		return err
+// getDefaultRunStrategy returns the run strategy for the restored VM.
+// HaltAfterRestore wins over the source VM's run strategy; if neither is set
+// we fall back to RerunOnFailure.
+func (h *RestoreHandler) getDefaultRunStrategy(
```
brandboat: getDefaultRunStrategy now is only used in the test, shouldn't this be invoked in ensureVMStartedAndReady? In this patch the restored vm runstrategy is always Halted.

pkg/webhook/util/snapshot.go
```diff
@@ -120,107 +120,56 @@ func CheckTotalSnapshotSizeOnNamespace(
...
-	// Get wrapper provisioner for checking snapshot/backup ability
-	provisioner, err := provisionerWrapper(pvc, engineCache, scCache)
-	if err != nil {
-		return err
-	}
-
-	ability := getAbilityForProvisioner(provisioner)
-	if !ability.IsSupported(bt) {
-		return fmt.Errorf("provisioner %s is not supported for type %s", provisioner, bt)
-	}
```
brandboat, line 188: Before then we reject remote backup with non-longhorn pvcs, but in this patch we remove this limitation. Perhaps we should keep it. I'm worried that user create vmbackup with type=backup with other non-longhorn pvcs.

This can be fixed in the separate pr from my perspective.

pkg/controller/master/backup/restore.go
```diff
@@ -866,559 +754,156 @@ func getNewVMAnnotations(restore *harvesterv1.VirtualMachineRestore, sourceAnnot
...
 		if err != nil {
-			if apierrors.IsNotFound(err) {
-				continue
-			}
 			return err
 		}
-
-		err = h.pvcClient.Delete(vol.Namespace, vol.Name, &metav1.DeleteOptions{})
-		if err != nil {
+		if err := h.vmro.SetVolRestoreProgress(vr, int(progress)); err != nil {
```
brandboat, line 822: I think the VMRestore aggregate progress calculation was accidentally dropped. Now we only updates each volume restore progress and returns without setting the top-level `vmr.Status.Progress` until final completion jumps it to 100.

brandboat (commented):
final round, overall lgtm

pkg/webhook/util/snapshot.go
WebberHuang1118: This part was mostly for LH v2, as LH v2 did not support backup in previous versions.

Since LH v2 now supports backup, this part has become somewhat duplicated with the setting documented here: [CSI Driver Config](https://docs.harvesterhci.io/v1.9/advanced/index#csi-driver-config). Therefore, I removed this part, and the webhook now relies on that setting.

Since the GUI does not allow third-party CSI drivers to set a `VolumeSnapshotClass`, the safeguard mechanism should be sufficient here.

pkg/webhook/resources/virtualmachinebackup/validator.go
WebberHuang1118: updated, thanks.

pkg/controller/master/backup/restore.go
WebberHuang1118: If `VMRestore` is not specified to be halted after restore, as shown [here](https://github.com/WebberHuang1118/harvester/blob/3fef37446bd8a8d0b24c9b0ae53057b75a07feec/pkg/controller/master/backup/restore.go#L873-L875), the `VMRestore` controller will start the VM, as shown [here](https://github.com/WebberHuang1118/harvester/blob/3fef37446bd8a8d0b24c9b0ae53057b75a07feec/pkg/controller/master/backup/restore.go#L881-L884). 

This sets the `runStrategy` to `RerunOnFailure`.

pkg/controller/master/backup/restore.go
WebberHuang1118: Good catch, updated, thanks.

pkg/controller/master/backup/restore.go
brandboat: Pardon, in StartVM, if vm runstrategy is halted, we only run startvm sub resource command, I'm not quite sure where do we change the vm runstrategy to rerunonfailure?

````go
	case kubevirtv1.RunStrategyHalted:
		return vmro.startVMSubresource(ctx, vm)
````

WebberHuang1118: Yes, this leads to [`pkg/restore/common/operator.go`](https://github.com/WebberHuang1118/harvester/blob/25fcb2c0abac642601211a51888381d88ae0c6b2/pkg/restore/common/operator.go), and on the KubeVirt side, this API will change the `runStrategy` to `RerunOnFailure`.

brandboat: Now I get it, we update runStrategy to RerunOnFailure here:
https://github.com/harvester/harvester/blob/master/pkg/webhook/resources/virtualmachine/mutator.go#L199-L206

On the kubevirt side, if we run start a vm with halted runStrategy, the runStrategy is changed from halted to always: https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-api/rest/lifecycle.go#L111
https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-api/rest/lifecycle.go#L636

Thanks!

brandboat (dismissed):
LGTM, thanks for take care all comments

pkg/controller/master/backup/restore.go
brandboat: BTW, getDefaultRunStrategy and its related tests should be removed in my opinion

WebberHuang1118: > BTW, getDefaultRunStrategy and its related tests should be removed in my opinion

Updated, thanks.

brandboat (dismissed):

ihcsim (dismissed):
overall lgtm - thanks!

brandboat (dismissed):

brandboat (approved):

ihcsim (approved):

votdev (approved):

Issue #9375, opened 2025-10-27
[BUG] Storage network optimizations


rrajendran17:
### Describe the Bug

enhance error handling as per https://github.com/harvester/harvester/pull/9364 discussions

### To Reproduce

https://github.com/harvester/harvester/pull/7126#issue-2728257072

More details and test cases will be added later after the enhancement.

### Expected Behavior

.

### Support Bundle for Troubleshooting

.

### Environment

- Harvester version: >= v1.5.0
- Rancher version:N/A

w13915984028:
cc @martindekov also PR https://github.com/harvester/harvester/pull/9364

martindekov:
Is this issue and https://github.com/harvester/harvester/issues/9141 duplicates ?

w13915984028:
A short summary of those to-be-optimized parts:

(1) The use of node object
In K8s, most CRD object has `spec` and `status` field, the controller works per `spec` and update `status` accordingly.

When a node is down/offline, K8s controller updates it's status and say like `kubelet stops to update node...`.

The code on PR https://github.com/harvester/harvester/pull/9364, use len(node) to make further decision, the assumption is not correct all the time. As a consumer of the  K8s `node` object, check `status` will make the code even robust.


(2) Remove the out-dated code 
`rancher-monitoring` has been an addon for some versions, it is safe to remove related code like

https://github.com/harvester/harvester/blob/a9efdf23d3a32dc2728003d914b9ac18ccda1fb5/pkg/controller/master/storagenetwork/storage_network.go#L549

https://github.com/harvester/harvester/blob/a9efdf23d3a32dc2728003d914b9ac18ccda1fb5/pkg/controller/master/storagenetwork/storage_network.go#L630

(3) Double check the `onChange`

SN controller uses annotation and status to control the `enable` of the SN applying and stops when it is done, and later, when onChange happens, it should check `ReasonCompleted` first, when it is set, then don't try those following operations like `checkPodStatusAndStart`, which might race with other user operations.

In short, the `enable SN` only triggers the workflow in a limited range of time, and stops via its `ReasonCompleted` status.

note: When Harvester (leader) contrller restarts, the setting `onChange` is triggered.

https://github.com/harvester/harvester/blob/a9efdf23d3a32dc2728003d914b9ac18ccda1fb5/pkg/controller/master/storagenetwork/storage_network.go#L176-L180

w13915984028:
@martindekov #9141 has certain overlap with #9375

[...]

PR #10563, opened 2026-05-12
Storage Network Optimizations

rrajendran17:
#### Problem:
Storage Network Optimizations

#### Solution:
1.Remove unused rancher monitoring start/stop code
2.Check for node status during validation of whereabouts IP allocation
3.Skip storage network configurations if configuration is complete
4.Restart monitoring pods only after validation of whereabouts IP allocation.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9375

#### Test plan:
case 1: Node status
1.Have a 2 node cluster
2.Make sure not to have any images created or vms running.(this is to prevent bim pod from running).
   Check if instance manager pods are running on each node (Node count 2)
3.Move one of the nodes to maintenance mode
4.Check if instance manager pod is running only on one node.
5.Create storage network and verify if storage network is configured successfully without error "IP Allocation failure for longhorn pods"
6.Whereabouts allocated 1 IP and len(nodes) will be 2 and validation would have failed without the PR changes.
With the PR changes, length of whereabouts IP allocation will be >= to 1 node.

### DIFF
--- a/pkg/controller/master/storagenetwork/storage_network.go
+++ b/pkg/controller/master/storagenetwork/storage_network.go
@@ -476,7 +476,7 @@ func (h *Handler) validateIPAddressesAllocations(setting *harvesterv1.Setting) e
 	//Formula - https://docs.harvesterhci.io/v1.4/advanced/storagenetwork/
 	//Dynamic parameters like number of images download/upload, backing-image-manager and backing-image-ds are skipped
 	//and only the number of nodes each running an instance-manager pod is used
-	MinAllocatableIPAddrs := util.CountNonWitnessNodes(nodes)
+	MinAllocatableIPAddrs := util.CountNonWitnessHealthyNodes(nodes)
 
 	var config network.Config
 
@@ -659,34 +659,6 @@ func (h *Handler) checkGrafanaStatusAndStart() error {
 	return nil
 }
 
-func (h *Handler) checkRancherMonitoringStatusAndStart() error {
-	// check managedchart fleet-local/rancher-monitoring paused
-	monitoring, err := h.managedChartCache.Get(util.FleetLocalNamespaceName, util.RancherMonitoring)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			logrus.Infof("rancher monitoring not found. skip")
-			return nil
-		}
-		return fmt.Errorf("rancher monitoring get error %v", err)
-	}
-
-	// check pause or not
-	if _, ok := monitoring.Annotations[util.PausedStorageNetworkAnnotation]; ok {
-		logrus.Infof("current Rancher Monitoring paused: %v", monitoring.Spec.Paused)
-		logrus.Infof("start rancher monitoring")
-		monitoringCopy := monitoring.DeepCopy()
-		monitoringCopy.Spec.Paused = false
-		delete(monitoringCopy.Annotations, util.PausedStorageNetworkAnnotation)
-
-		if _, err := h.managedCharts.Update(monitoringCopy); err != nil {
-			return fmt.Errorf("rancher monitoring error %v", err)
-		}
-		return nil
-	}
-
-	return nil
-}
-
 func (h *Handler) checkVMImportControllerStatusAndStart() error {
 	// check deployment harvester-system/harvester-harvester-vm-import-controller replica
 	vmImportControllerDeploy, err := h.deploymentCache.Get(util.HarvesterSystemNamespaceName, util.HarvesterVMImportController)
@@ -733,39 +705,7 @@ func (h *Handler) checkPodStatusAndStart() error {
 		return err
 	}
 
-	if err := h.checkRancherMonitoringStatusAndStart(); err != nil {
-		return err
-	}
-
 	return h.checkVMImportControllerStatusAndStart()
-}
-
-func (h *Handler) checkRancherMonitoringStatusAndStop() error {
-	// check managedchart fleet-local/rancher-monitoring paused
-	monitoring, err := h.managedChartCache.Get(util.FleetLocalNamespaceName, util.RancherMonitoring)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			logrus.Infof("rancher monitoring not found. skip")
-			return nil
-		}
-		return fmt.Errorf("rancher monitoring get error %v", err)
-	}
-
-	// check pause or not
-	if !monitoring.Spec.Paused {
-		logrus.Infof("current Rancher Monitoring paused: %v", monitoring.Spec.Paused)
-		logrus.Infof("stop rancher monitoring")
-		monitoringCopy := monitoring.DeepCopy()
-		monitoringCopy.Annotations[util.PausedStorageNetworkAnnotation] = "false"
-		monitoringCopy.Spec.Paused = true
-
-		if _, err := h.managedCharts.Update(monitoringCopy); err != nil {
-			return fmt.Errorf("rancher monitoring error %v", err)
-		}
-		return nil
-	}
-
-	return err
 }
 
 func (h *Handler) checkPrometheusStatusAndStop() error {
@@ -881,10 +821,6 @@ func (h *Handler) checkVMImportControllerStatusAndStop() error {
 
 // check Pod status, if all pods are stopped, return true
 func (h *Handler) checkPodStatusAndStop() error {
-	if err := h.checkRancherMonitoringStatusAndStop(); err != nil {
-		return err
-	}
-
 	if err := h.checkPrometheusStatusAndStop(); err != nil {
 		return err
 	}
--- a/pkg/controller/master/node/promote_controller.go
+++ b/pkg/controller/master/node/promote_controller.go
@@ -310,7 +310,7 @@ func selectPromoteNode(nodeList []*corev1.Node) *corev1.Node {
 				managementZones[zone] = true
 				managementOrHealthyHarvesterWorkerZones[zone] = true
 			}
-		} else if isHealthyNode(node) && isHarvesterNode(node) &&
+		} else if util.IsHealthyNode(node) && isHarvesterNode(node) &&
 			!isWorkerPreferredNode(node) && !isExtraWitnessNode(node, len(witnessPreferred), witnessPromoted) {
 			if zone != "" {
 				managementOrHealthyHarvesterWorkerZones[zone] = true
@@ -385,23 +385,6 @@ func isExtraWitnessNode(node *corev1.Node, numOfWitnessNode int, promotedWitness
 func isWorkerPreferredNode(node *corev1.Node) bool {
 	_, found := node.Labels[HarvesterWorkerNodeLabelKey]
 	return found
-}
-
-// isHealthyNode determine whether it's an healthy node
-func isHealthyNode(node *corev1.Node) bool {
-	for _, c := range node.Status.Conditions {
-		if c.Type == corev1.NodeReady && c.Status != corev1.ConditionTrue {
-			// skip unready nodes
-			return false
-		}
-
-		if c.Type != corev1.NodeReady && c.Status == corev1.ConditionTrue {
-			// skip node with conditions like nodeMemoryPressure, nodeDiskPressure, nodePIDPressure
-			// and nodeNetworkUnavailable equal to true
-			return false
-		}
-	}
-	return true
 }
 
 // isHarvesterNode determine whether it's an Harvester node based on the node's label
--- a/pkg/util/node.go
+++ b/pkg/util/node.go
@@ -42,6 +42,23 @@ func IsPromoteStatusIn(node *corev1.Node, statuses ...string) bool {
 	return false
 }
 
+// IsHealthyNode determine whether it's an healthy node
+func IsHealthyNode(node *corev1.Node) bool {
+	for _, c := range node.Status.Conditions {
+		if c.Type == corev1.NodeReady && c.Status != corev1.ConditionTrue {
+			// skip unready nodes
+			return false
+		}
+
+		if c.Type != corev1.NodeReady && c.Status == corev1.ConditionTrue {
+			// skip node with conditions like nodeMemoryPressure, nodeDiskPressure, nodePIDPressure
+			// and nodeNetworkUnavailable equal to true
+			return false
+		}
+	}
+	return true
+}
+
 func IsWitnessNodeWithoutPromotionStatus(node *corev1.Node) bool {
 	val, found := node.Labels[HarvesterWitnessNodeLabelKey]
 	if found && val == "true" {
@@ -86,11 +103,12 @@ func IsManagementRole(node *corev1.Node) bool {
 }
 
 // count the number of nodes running instance manager pod
-func CountNonWitnessNodes(nodes []*corev1.Node) int {
+func CountNonWitnessHealthyNodes(nodes []*corev1.Node) int {
 	count := 0
 
 	for _, node := range nodes {
-		if !IsWitnessNodeWithoutPromotionStatus(node) {
+		// consider only non-wintess and healthy nodes, as witness nodes won't run LH pods and unhealthy nodes won't be scheduled with LH pods
+		if !IsWitnessNodeWithoutPromotionStatus(node) && IsHealthyNode(node) {
 			count++
 		}
 	}


rrajendran17:
@WebberHuang1118 @w13915984028 I also want to start a discussion regarding stopping/starting of monitoring Pods during Storage Network Configuration.
https://github.com/harvester/harvester/blob/b9c1da09b5bd8a72353953a49f9a6275898c609f/pkg/controller/master/storagenetwork/storage_network.go#L228

From my understanding,stopping and starting of monitoring pods makes sure not to record network down notifications during storage network configurations as there is change in network during this time and traffic loss is expected.
But IMO, these pod must record these activities and also it is possible to miss other alerts during these pods downtime.
Also the code checks if all volumes are detached which may fail if these monitoring pod are running, that is an another reason why these pods might have been stopped and restarted, but its possible to skip these pods while checking for detached volumes. 

To summarize, I feel stop/start of monitoring pods must be removed from code and they should continue to monitor the resources during the storage network configuration. Users will be aware of the timeline when storage network is reconfigured and these alarms should be handled as expected. Please let me know if I miss any history for this design decision and this could be changed now.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -501,6 +506,11 @@ func (h *Handler) validateIPAddressesAllocations(setting *harvesterv1.Setting) e
 }
 
 func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting) (*harvesterv1.Setting, error) {
+	//skip if sn configuration is already completed
```
w13915984028: It looks safe to put this block of code before L524 as following code are state based;  but still run L510~L523, for RWX to have chance to be reconcilled

cc @brandboat to double check

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -534,10 +536,19 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
 		return setting, err
```
w13915984028: not related to this PR, the original

return `setting, err` discards the effect of `Enqueue`

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -534,10 +536,19 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
...
 	}
 
+	// check if we need to restart monitoring pods
+	if err := h.checkPodStatusAndStart(); err != nil {
+		if _, updateConditionErr := h.setConfiguredCondition(setting, false, ReasonInProgress, MsgRestartPod); updateConditionErr != nil {
+			return setting, fmt.Errorf("update status error %v", updateConditionErr)
+		}
+		return setting, err
```
w13915984028: similar to above validateIPAddressesAllocations, use `h.settingsController.Enqueue(setting.Name)` or `return setting, err` ?

pkg/controller/master/storagenetwork/storage_network.go
w13915984028: could we add a log of this error in case of Enqueue for better troubleshooting?

when Enqueue and return nil, framework won't log any error
when return error directly, framework log error and reconcille the object in backoff mode

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -534,10 +536,19 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
 		return setting, err
 	}
 
+	// check if we need to restart monitoring pods
+	if err := h.checkPodStatusAndStart(); err != nil {
+		if _, updateConditionErr := h.setConfiguredCondition(setting, false, ReasonInProgress, MsgRestartPod); updateConditionErr != nil {
```
w13915984028: similar, if use Enqueue, please add a log

w13915984028 (dismissed):
A few questions, thanks.

w13915984028:
> @WebberHuang1118 @w13915984028 I also want to start a discussion regarding stopping/starting of monitoring Pods during Storage Network Configuration.
> 
> https://github.com/harvester/harvester/blob/b9c1da09b5bd8a72353953a49f9a6275898c609f/pkg/controller/master/storagenetwork/storage_network.go#L228
> 
> From my understanding,stopping and starting of monitoring pods makes sure not to record network down notifications during storage network configurations as there is change in network during this time and traffic loss is expected. But IMO, these pod must record these activities and also it is possible to miss other alerts during these pods downtime. Also the code checks if all volumes are detached which may fail if these monitoring pod are running, that is an another reason why these pods might have been stopped and restarted, but its possible to skip these pods while checking for detached volumes.
> 
> To summarize, I feel stop/start of monitoring pods must be removed from code and they should continue to monitor the resources during the storage network configuration. Users will be aware of the timeline when storage network is reconfigured and these alarms should be handled as expected. Please let me know if I miss any history for this design decision and this could be changed now.

@rrajendran17 

The limitation comes from LH or other CSI provisioner, it needs to re-attache those volumes when SN is changing (SH<->others/pod network). Without stopping all volume related workloads (vm, monitoring or others), workload could run into panic/hang, or LH will report errors.  Storage is the very low level resources, monitoring works relying on them.

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: yes right.Looks like both making it redundant. I have removed the Enqueue, functionality looks good with just the returning of error for retry and reconciliation and its in sync with rest of the code.

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: rwx reconciliation is  needed only when storage network parameters change (shared mode), and it will not require reconciliation when storage network config is complete.Anyways @brandboat can reconfirm.

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: Will stick to existing return err

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: Will stick to existing return err

pkg/controller/master/storagenetwork/storage_network.go
brandboat: If the storage-network value hasn't changed, there's no need to update the rwx-network. An early return makes sense here because the condition remains true if storage network hasn't changed.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -511,33 +521,33 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
...
 	}
 
 	err = h.validateIPAddressesAllocations(setting)
 	if err != nil {
-		setting, err := h.setConfiguredCondition(setting, false, ReasonInProgress, MsgIPAssignmentFailure)
-		if err != nil {
-			return setting, fmt.Errorf("update status error %v", err)
+		_, updCondErr := h.setConfiguredCondition(setting, false, ReasonInProgress, MsgIPAssignmentFailure)
+		if updCondErr != nil {
+			return setting, fmt.Errorf("update status error %v", updCondErr)
```
w13915984028: two points:
(1) use `%w` in error wrap
(2) it looks in this extreme unlucky case, the original `err` was lost, it is better to concatenate the errors

https://github.com/harvester/harvester/pull/10563/changes#diff-87504a1be112ef39a8cdaf8814e6bc2da202ca091364097a26d589d62f8c9248R541 is same

pkg/controller/master/storagenetwork/storage_network.go
w13915984028: the last concern:

from my understanding, the original code is not strict state machine based; if it is, the state should be checked on line like https://github.com/harvester/harvester/blob/60555dde23abde8ab762a217212029be92edb81b/pkg/controller/master/storagenetwork/storage_network.go#L192

on original code, each time, when Harvester controller leader runs, it goes to this function `handleLonghornSettingPostConfig` and calls enqueue RWX. it does no harm if configures are in sync; if not, they are synced.

with the new code, this process is shortcuted,

but it feels not a whole solution to utilize the state machine control

w13915984028 (commented):
two questions, thanks.

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: https://github.com/harvester/harvester/blob/b239e1778adc309ca596f25ee23b61ca6285bf68/pkg/controller/master/storagenetwork/storage_network.go#L1195 rwxnetwork is not updated when there is no change in storage network nad during its reconcilation. So I am fine moving the check after the rwxnetwork shared with storage validation.

WebberHuang1118:
> @WebberHuang1118 @w13915984028 I also want to start a discussion regarding stopping/starting of monitoring Pods during Storage Network Configuration.
> 
> https://github.com/harvester/harvester/blob/b9c1da09b5bd8a72353953a49f9a6275898c609f/pkg/controller/master/storagenetwork/storage_network.go#L228
> 
> From my understanding,stopping and starting of monitoring pods makes sure not to record network down notifications during storage network configurations as there is change in network during this time and traffic loss is expected. But IMO, these pod must record these activities and also it is possible to miss other alerts during these pods downtime. Also the code checks if all volumes are detached which may fail if these monitoring pod are running, that is an another reason why these pods might have been stopped and restarted, but its possible to skip these pods while checking for detached volumes.
> 
> To summarize, I feel stop/start of monitoring pods must be removed from code and they should continue to monitor the resources during the storage network configuration. Users will be aware of the timeline when storage network is reconfigured and these alarms should be handled as expected. Please let me know if I miss any history for this design decision and this could be changed now.

My understanding is similar to @w13915984028’s comment in [harvester/harvester#10563](https://github.com/harvester/harvester/pull/10563#issuecomment-4439892169): the monitoring Pods will attach LH volumes, and the storage network cannot be updated while any LH volume is running.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -733,41 +722,9 @@ func (h *Handler) checkPodStatusAndStart() error {
...
-		monitoringCopy.Spec.Paused = true
-
-		if _, err := h.managedCharts.Update(monitoringCopy); err != nil {
-			return fmt.Errorf("rancher monitoring error %v", err)
-		}
-		return nil
-	}
-
-	return err
-}
```
WebberHuang1118, line 769: One question:

@rrajendran17, as discussed in [harvester/harvester#10563](https://github.com/harvester/harvester/pull/10563#issuecomment-4533907520), if the monitoring Pod is running, it will use a running LH volume and prevent the storage network change.

In this scenario, IIUC, we still need to automatically stop and start the monitoring Pods during storage network reconfiguration.

So, can we skip the automatic handling for monitoring Pods?

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -162,6 +162,11 @@ func Register(ctx context.Context, management *config.Management, _ config.Optio
 	return nil
 }
 
+func getConfiguredCondition(setting *harvesterv1.Setting) bool {
+	completed := harvesterv1.SettingConfigured.IsTrue(setting)
+	return completed
+}
```
WebberHuang1118:
```suggestion
func completed(setting *harvesterv1.Setting) bool {
	return harvesterv1.SettingConfigured.IsTrue(setting)
}
```

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -511,12 +516,10 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
...
-	if err := h.checkPodStatusAndStart(); err != nil {
-		if _, updateConditionErr := h.setConfiguredCondition(setting, false, ReasonInProgress, MsgRestartPod); updateConditionErr != nil {
-			return setting, fmt.Errorf("update status error %v", updateConditionErr)
-		}
-		return setting, err
+	//skip if sn configuration is already completed
+	completed := getConfiguredCondition(setting)
+	if completed {
+		return setting, nil
 	}
```
WebberHuang1118:
```suggestion
	//skip if sn configuration is already completed
	if completed(setting) {
		return setting, nil
	}
```

WebberHuang1118 (commented):
Thanks for the improvement. One question and a few suggestions. Please take a look.

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -526,18 +529,32 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
...
+	// check if we need to restart monitoring pods
+	if err := h.checkPodStatusAndStart(); err != nil {
+		if _, updateConditionErr := h.setConfiguredCondition(setting, false, ReasonInProgress, MsgRestartPod); updateConditionErr != nil {
+			return setting, errors.Join(
+				fmt.Errorf("failed to restart monitoring pods: %w", err),
+				fmt.Errorf("failed to update configured condition: %w", updateConditionErr),
+			)
+		}
+		return setting, fmt.Errorf("failed to restart monitoring pods: %w", err)
 	}
```
WebberHuang1118, line 535: Suggest making this part more compact:
```
func (h *Handler) failCondition(
	setting *harvesterv1.Setting,
	originalErr error,
	reason string,
	msg string,
	context string,
) (*harvesterv1.Setting, error) {
	if _, err := h.setConfiguredCondition(setting, false, reason, msg); err != nil {
		return setting, errors.Join(
			fmt.Errorf("%s: %w", context, originalErr),
			fmt.Errorf("failed to update configured condition: %w", err),
		)
	}

	return setting, fmt.Errorf("%s: %w", context, originalErr)
}

if err := h.validateIPAddressesAllocations(setting); err != nil {
	return h.failCondition(setting, err, ReasonInProgress, MsgIPAssignmentFailure, "validate IP allocation failed")
}

if err := h.checkPodStatusAndStart(); err != nil {
	return h.failCondition(setting, err, ReasonInProgress, MsgRestartPod, "failed to restart monitoring pods")
}
```

pkg/controller/master/storagenetwork/storage_network.go
```diff
@@ -511,12 +516,10 @@ func (h *Handler) handleLonghornSettingPostConfig(setting *harvesterv1.Setting)
...
 
-	// check if we need to restart monitoring pods
-	if err := h.checkPodStatusAndStart(); err != nil {
-		if _, updateConditionErr := h.setConfiguredCondition(setting, false, ReasonInProgress, MsgRestartPod); updateConditionErr != nil {
-			return setting, fmt.Errorf("update status error %v", updateConditionErr)
-		}
-		return setting, err
+	//skip if sn configuration is already completed
+	completed := getConfiguredCondition(setting)
```
w13915984028: another side effect of this change:

in a corner/rare case: when a new node is added to the system, LH related new PODs can't get the related IPs.

in this past, when Harvester POD is rebooted, the SN will show error status;  but with this change, could such error still be triggered?

w13915984028 (dismissed):
LGTM, thanks.

With a last open question:

pkg/controller/master/storagenetwork/storage_network.go
w13915984028: The `managedchart` based `rancher-monitoring` has been removed for days;

the current `checkPrometheusStatusAndStart` & `checkPrometheusStatusAndStop` pairs (also grafana, alertmanager) have covered the the monitoring related volume based workloads. remove the dead function `checkRancherMonitoringStatusAndStop` should be fine.

WebberHuang1118: thanks @w13915984028

pkg/controller/master/storagenetwork/storage_network.go
rrajendran17: I believe new node addition can be similar to a reboot, the storage network change will be triggered, L225-254 https://github.com/harvester/harvester/blob/bfc85b48cbd6283fc50060244e7713a1c05fdcbf/pkg/controller/master/storagenetwork/storage_network.go#L223 will set status to In Progress, handleLonghornSettingPostConfig should handle other operations only if status in In Progress and can skip if its already completed.

Can you explain in detail about the error msg you saw in the past ?
"in this past, when Harvester POD is rebooted, the SN will show error status; but with this change, could such error still be triggered?"

rrajendran17: Tried the following sequence,
1.Have 2 nodes in the cluster and cordon node 2
2.Configure storage network with only 1 IP available (1 instance manager pod alone running on node 1)
3.Uncordon the node-2
4.Instance manager pod on node-2 will be container creating state (whereabouts failed to allocate IP)
5. There will not be any reconcilation during new node addition/re-join
6. Now reboot the master node (this will retrigger reconcilation of storage network)
7. Since whereabouts not able to allocate ip address for the IM pod on node-2 ideally validateIPAddressesAllocations should be called to show error msg, but if the code is check for status already completed, then it will skip the validateIPAddressesAllocations validation.

As we discussed, Though the above scenario (node join + reboot of master node) mentioned is very rare, but still possible where error in the status could be missed, so for now I have removed the completed check.

w13915984028 (dismissed):
LGTM, thanks.

WebberHuang1118 (approved):

Issue #9775, opened 2025-12-22
[BUG] Online Expansion fails for Encrypted Longhorn Volumes


WebberHuang1118:
### Describe the Bug

Encrypted Longhorn volumes cannot be expanded online

### To Reproduce

1. On Harvester v1.7, create a VM using an [encrypted image](https://docs.harvesterhci.io/v1.7/image/image-security#encrypt-a-virtual-machine-image) as the root disk.
2. After the VM is running, expand the PVC size.
3. The VM fails to recognize the expanded root disk size.
4. On the cluster node, the Longhorn iSCSI device (`/dev/sdb`) grows, but the dm-crypt device (`pvc-0cbf830f-7a97-4360-8147-7c316e615f1d`) does not.
  ```
  hp-155-tink-system:/home/rancher # dmsetup ls --tree
  pvc-0cbf830f-7a97-4360-8147-7c316e615f1d (254:0)
   └─ (8:16)
  hp-155-tink-system:/home/rancher # ls -al /dev/sdb
  brw-rw----. 1 root disk 8, 16 Dec 19 10:49 /dev/sdb
  hp-155-tink-system:/home/rancher # blockdev --getsize64 /dev/sdb
  32212254720
  hp-155-tink-system:/home/rancher # blockdev --getsize64 /dev/mapper/pvc-0cbf830f-7a97-4360-8147-7c316e615f1d
  21458059264
  ```
5. Since the CSI-published device (the dm-crypt device in this case) retains its original size, KubeVirt cannot propagate the new size to the running VM.

### Expected Behavior

The dm-crypt device should also be expanded so that the volume resize is properly propagated to the VM.

### Support Bundle for Troubleshooting

[supportbundle_8345f4be-3acb-4d6e-99bd-3306ffcc3421_2025-12-19T11-02-28Z.zip](https://github.com/user-attachments/files/24286769/supportbundle_8345f4be-3acb-4d6e-99bd-3306ffcc3421_2025-12-19T11-02-28Z.zip)

### Environment

- Harvester version: v1.7.0-rc6
- Impacted VM: 
- Impacted volume (PV): 
- Underlying Infrastructure (e.g., Baremetal with Dell PowerEdge R630):
- Rancher version:

### Additional context

This appears to stem from multiple Longhorn behaviors:

The fundamental issue is that the Longhorn-published dm-crypt device does not grow alongside the underlying iSCSI device, as described in [issue #12366](https://github.com/longhorn/longhorn/issues/12366).

[...]

PR #10568, opened 2026-05-13
enhancement: Add node expansion secret as optional and reduce duplicated code

WebberHuang1118:
#### Problem:
https://github.com/harvester/harvester/issues/9775

#### Solution:
Add node expansion secret support as optional to allow encrypted volume expansion.

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9775

#### Test plan:
* Prepare a Harvester `v1.8.0` cluster with LH `v1.11.1`.
* Boot a VM with an encrypted LH volume as the root disk.
* Check the related storage size on the cluster node. The related block device is 16 MB smaller than expected.
  ```
  hp-155-tink-system:/home/rancher # lsblk /dev/sdb
  NAME                                       MAJ:MIN RM SIZE RO TYPE  MOUNTPOINTS
  sdb                                          8:16   0   6G  0 disk  
  └─pvc-0a0d064c-1803-4c3e-8dc9-84a1b1743670 254:0    0   6G  0 crypt 
  hp-155-tink-system:/home/rancher # blockdev --getsize64 /dev/sdb
  6442450944
  hp-155-tink-system:/home/rancher # blockdev --getsize64 /dev/mapper/pvc-0a0d064c-1803-4c3e-8dc9-84a1b1743670 
  6425673728
  ```
* Update the LH deployments with the following manifests:
  * [crds.yaml](https://github.com/WebberHuang1118/longhorn/blob/v1.12.0-rc1-direct-apply/chart/templates/crds.yaml)
  * [daemonset-sa.yaml](https://github.com/WebberHuang1118/longhorn/blob/v1.12.0-rc1-direct-apply/chart/templates/daemonset-sa.yaml)
  * [clusterrole.yaml](https://github.com/WebberHuang1118/longhorn/blob/v1.12.0-rc1-direct-apply/chart/templates/clusterrole.yaml)
  * [deployment-driver.yaml](https://github.com/WebberHuang1118/longhorn/blob/v1.12.0-rc1-direct-apply/chart/templates/deployment-driver.yaml)
* After the LH engine live-upgrades to v1.12, check the related storage size on the cluster node. The related block device have the correct size.
  ```
  p-155-tink-system:/home/rancher # lsblk /dev/sdb
  NAME                                       MAJ:MIN RM SIZE RO TYPE  MOUNTPOINTS
  sdb                                          8:16   0   6G  0 disk  
  └─pvc-0a0d064c-1803-4c3e-8dc9-84a1b1743670 254:0    0   6G  0 crypt 
  hp-155-tink-system:/home/rancher # blockdev --getsize64 /dev/sdb
  6459228160
  hp-155-tink-system:/home/rancher # blockdev --getsize64 /dev/mapper/pvc-0a0d064c-1803-4c3e-8dc9-84a1b1743670 
  6442450944
  ```
* Trigger VM live migration. The related block device should also migrate to another node. Then check the related storage size on the target cluster node.
  ```
  hp-107-tink-system:/home/rancher # lsblk /dev/sdd
  NAME                                       MAJ:MIN RM SIZE RO TYPE  MOUNTPOINTS
  sdd                                          8:48   0   6G  0 disk  
  └─pvc-0a0d064c-1803-4c3e-8dc9-84a1b1743670 254:0    0   6G  0 crypt 
  hp-107-tink-system:/home/rancher # blockdev --getsize64 /dev/sdd
  6459228160
  hp-107-tink-system:/home/rancher # blockdev --getsize64 /dev/mapper/pvc-0a0d064c-1803-4c3e-8dc9-84a1b1743670 
  6442450944
  ```
* Trigger volume online expansion. Then check the related storage size on the target cluster node and inside the guest VM.

### DIFF
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -149,6 +149,8 @@ const (
 	CSINodePublishSecretNamespaceKey = "csi.storage.k8s.io/node-publish-secret-namespace"
 	CSINodeStageSecretNameKey        = "csi.storage.k8s.io/node-stage-secret-name"
 	CSINodeStageSecretNamespaceKey   = "csi.storage.k8s.io/node-stage-secret-namespace"
+	CSINodeExpandSecretNameKey       = "csi.storage.k8s.io/node-expand-secret-name"
+	CSINodeExpandSecretNamespaceKey  = "csi.storage.k8s.io/node-expand-secret-namespace"
 
 	LabelUpgradeReadMessage          = prefix + "/read-message"
 	LabelUpgradeState                = prefix + "/upgradeState"
@@ -309,3 +311,18 @@ const (
 
 	GoArchArm64 = "arm64"
 )
+
+// CSISecretKeyPair groups a CSI secret-name parameter key with its matching namespace key.
+type CSISecretKeyPair struct {
+	NameKey      string
+	NamespaceKey string
+}
+
+// CSIEncryptionSecretKeyPairs lists the CSI parameter key pairs used to reference
+// the encryption secret for a StorageClass, in the order CSI expects them.
+var CSIEncryptionSecretKeyPairs = []CSISecretKeyPair{
+	{CSIProvisionerSecretNameKey, CSIProvisionerSecretNamespaceKey},
+	{CSINodeStageSecretNameKey, CSINodeStageSecretNamespaceKey},
+	{CSINodePublishSecretNameKey, CSINodePublishSecretNamespaceKey},
+	{CSINodeExpandSecretNameKey, CSINodeExpandSecretNamespaceKey},
+}
--- a/pkg/webhook/resources/storageclass/validator.go
+++ b/pkg/webhook/resources/storageclass/validator.go
@@ -33,20 +33,38 @@ const (
 	errorMessageReservedStorageClass = "storage class %s is reserved by Harvester and can't be deleted"
 )
 
+type secretPair struct {
+	nameKey, namespaceKey string
+	required              bool
+}
+
+// optionalEncryptionSecretKeys are CSI key pairs whose absence is tolerated by the validator.
+var optionalEncryptionSecretKeys = map[string]bool{
+	util.CSINodeExpandSecretNameKey: true,
+}
+
 var (
 	availableCiphers  = []string{"aes-xts-plain", "aes-xts-plain64", "aes-cbc-plain", "aes-cbc-plain64", "aes-cbc-essiv:sha256"}
 	availablePBKDFs   = []string{"argon2i", "argon2id", "pbkdf2"}
 	availableHash     = []string{"sha256", "sha384", "sha512"}
 	availableKeySizes = []string{"256", "384", "512"}
 
-	pairs = [][2]string{
-		{util.CSIProvisionerSecretNameKey, util.CSIProvisionerSecretNamespaceKey},
-		{util.CSINodeStageSecretNameKey, util.CSINodeStageSecretNamespaceKey},
-		{util.CSINodePublishSecretNameKey, util.CSINodePublishSecretNamespaceKey},
-	}
+	encryptionPairs = buildEncryptionPairs()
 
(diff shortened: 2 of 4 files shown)


pkg/webhook/resources/storageclass/validator_test.go
```diff
@@ -375,6 +381,58 @@ func Test_storageClassValidator_validateEncryption(t *testing.T) {
 			},
 			expectError: true,
 		},
+		{
+			name: "optional node-expand secret omitted is allowed",
```
w13915984028, line 385: it seems the `different secret` test case could be added

pkg/util/constants.go
```diff
@@ -149,6 +149,8 @@ const (
 	CSINodePublishSecretNamespaceKey = "csi.storage.k8s.io/node-publish-secret-namespace"
 	CSINodeStageSecretNameKey        = "csi.storage.k8s.io/node-stage-secret-name"
 	CSINodeStageSecretNamespaceKey   = "csi.storage.k8s.io/node-stage-secret-namespace"
+	CSINodeExpandSecretNameKey       = "csi.storage.k8s.io/node-expand-secret-name"
```
w13915984028, line 152: could consider to add a helper to return the refactored `secretPair` for common usage on validator.go and imiage mutator.go

w13915984028 (dismissed):
LGTM, thanks.

Yu-Jack (dismissed):
Thanks for the PR!

pkg/webhook/resources/storageclass/validator.go
```diff
@@ -33,16 +33,22 @@ const (
...
 
-	pairs = [][2]string{
-		{util.CSIProvisionerSecretNameKey, util.CSIProvisionerSecretNamespaceKey},
-		{util.CSINodeStageSecretNameKey, util.CSINodeStageSecretNamespaceKey},
-		{util.CSINodePublishSecretNameKey, util.CSINodePublishSecretNamespaceKey},
+	encryptionPairs = []secretPair{
+		{util.CSIProvisionerSecretNameKey, util.CSIProvisionerSecretNamespaceKey, true},
+		{util.CSINodeStageSecretNameKey, util.CSINodeStageSecretNamespaceKey, true},
+		{util.CSINodePublishSecretNameKey, util.CSINodePublishSecretNamespaceKey, true},
+		{util.CSINodeExpandSecretNameKey, util.CSINodeExpandSecretNamespaceKey, false},
```
brandboat: Just a note, not a review:
This change is required because Longhorn silently ignores encrypted volume expansion if the StorageClass lacks the `csi.storage.k8s.io/node-expand-secret-name` parameter. Since Harvester restricts users from modifying StorageClass parameters post-deployment (Day 2), we must proactively add csi.storage.k8s.io/node-expand-secret-name to the whitelist.

If the parameter is missing, in longhorn-csi-plugin daemonset you shall see log below:
```
Skip encrypted device resizing for volume pvc-e4b9c95d-c2be-4e2a-8f73-7f73b9a4dc1b node expansion since the secret is empty, maybe the related feature gate is not enabled" func="csi.(*NodeServer).expandEncryptedVolume" file="node_server.go:893" component=csi-node-server function=expandEncryptedVolume
```

brandboat (dismissed):
LGTM, thanks for the fix

pkg/webhook/resources/storageclass/validator.go
WebberHuang1118: Yes, your description is correct.

Since the published [image security documentation](https://docs.harvesterhci.io/v1.8/image/image-security) missed the node expansion secret, I designed the node expansion secret as optional for compatibility.

I think we need to update the documentation accordingly.

WebberHuang1118: The source issue, [harvester/harvester#10568](https://github.com/harvester/harvester/pull/10568), is labeled with `require/doc`.

pkg/webhook/resources/storageclass/validator_test.go
WebberHuang1118: Updated per suggestion, thanks.

pkg/util/constants.go
WebberHuang1118: Good idea, having certain refactoring per suggestion, thanks.

brandboat (approved):

Issue #9995, opened 2026-02-02
[ENHANCEMENT] Refactor System CSI Driver Snapshot Settings for Multiple SnapClasses.


ParadoxGuitarist:
### Is your enhancement request of an existing feature related to a problem? Please describe (👍 if you like this request)

tldr; The current settings in the csi-driver permit you to specify **One** `VolumeSnapshotClass` for all `storageClasses` under the driver, but that `VolumeSnapshotClass` may not work for all `storageClass` PVs resulting in `VMSnapshot` failures. 

This issue is specific to the ceph rbd driver (the one that is upstream and incorporated with the rook-ceph rbd driver that is fully supported) when you use the driver with multiple (external) clusters. However any third party driver that might use multiple `VolumeSnapshotClass` objects *might* also be affected, depending on their own workflows. 

When connecting the ceph-rbd driver to more than 1 cluster, you provide a clusterID as both part of the `StorageClass` and the `VolumeSnapshotClass`. Both reference different secrets and clusters as part of their required fields, but what it really boils down to is that the `VolumeSnapshotClass` will only work with the matching clusterIDs. 

Here's an example of what connecting to two different clusters might look like for the yaml of the StorageClass and VolumeSnapshotClass objects: 

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
   name: slow-ceph
provisioner: rbd.csi.ceph.com
parameters:
   clusterID: ceph-slow-1234
   pool: outerrim
   imageFeatures: layering
   csi.storage.k8s.io/provisioner-secret-name: csi-rbd-rust
   csi.storage.k8s.io/provisioner-secret-namespace: default
   csi.storage.k8s.io/controller-expand-secret-name: csi-rbd-rust
   csi.storage.k8s.io/controller-expand-secret-namespace: default
   csi.storage.k8s.io/node-stage-secret-name: csi-rbd-rust
   csi.storage.k8s.io/node-stage-secret-namespace: default
reclaimPolicy: Retain
allowVolumeExpansion: true
mountOptions:
   - discard
---
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: slow-ceph-snapclass
driver: rbd.csi.ceph.com
parameters:
  clusterID: ceph-slow-1234
  csi.storage.k8s.io/snapshotter-secret-name: csi-rbd-rust
  csi.storage.k8s.io/snapshotter-secret-namespace: default
deletionPolicy: Delete
[...]

PR #10576, opened 2026-05-14
Use storage class snapshot class for VM snapshots

officialasishkumar:
#### Problem:
VM snapshots currently resolve one VolumeSnapshotClass per CSI driver from the csi-driver-config setting. CSI drivers such as Ceph RBD can back multiple StorageClasses that require different VolumeSnapshotClass objects, so a VM snapshot can fail when the selected driver-level snapshot class does not match the PVC StorageClass.

#### Solution:
Resolve the VolumeSnapshotClass per volume for snapshot-type VM backups. The StorageClass CDI annotation cdi.harvesterhci.io/storageProfileVolumeSnapshotClass now takes precedence for VM snapshots, while the existing csi-driver-config behavior remains the fallback and remains unchanged for backup-type VM backups.

Also allow snapshot admission and snapshot actions when the PVC StorageClass has that snapshot class annotation, and apply the same StorageClass override to the standalone volume snapshot action.

#### Related Issue(s):
Issue #9995

#### Test plan:
- go test ./pkg/controller/master/backup ./pkg/webhook/util ./pkg/api/vm ./pkg/api/volume
- go test ./pkg/util

N/A


pkg/api/volume/handler.go
```diff
@@ -426,12 +426,14 @@ func (h *ActionHandler) snapshot(_ context.Context, pvcNamespace, pvcName, snaps
 	}
 
 	provisioner := util.GetProvisionedPVCProvisioner(pvc, h.scCache)
-	csiDriverInfo, err := settings.GetCSIDriverInfo(provisioner)
-	if err != nil {
-		return err
+	volumeSnapshotClassName := util.GetPVCStorageClassSnapshotClassName(pvc, h.scCache)
+	if volumeSnapshotClassName == "" {
+		csiDriverInfo, err := settings.GetCSIDriverInfo(provisioner)
+		if err != nil {
```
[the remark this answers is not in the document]
officialasishkumar: Fixed by resolving the effective StorageClass name before creating snapshot annotations and omitting that annotation when no effective class is available.

pkg/controller/master/backup/backup.go
```diff
@@ -403,6 +401,19 @@ func (h *Handler) getCSIDriverMap(backup *harvesterv1.VirtualMachineBackup) (map
...
+// getCSIDriverMap retrieves VolumeSnapshotClassName for each csi driver.
+// It is kept for backward-compatible status fields. Snapshot creation resolves
+// the VolumeSnapshotClass per volume because a single CSI driver can serve
+// multiple storage classes with different snapshot classes.
+func (h *Handler) getCSIDriverMap(backup *harvesterv1.VirtualMachineBackup) (map[string]string, map[string]snapshotv1.VolumeSnapshotClass, error) {
+	csiDriverConfig, err := loadCSIDriverConfig()
+	if err != nil {
+		return nil, nil, err
+	}
+
```
WebberHuang1118: Hi @officialasishkumar,

Thanks for the idea. However, I’d like to keep the [CSI driver config](https://docs.harvesterhci.io/v1.8/advanced/index#csi-driver-config) as the only source for referencing the `VolumeSnapshotClass` during VM backup snapshot creation, rather than spreading the reference across multiple places.

Yes, the current [CSI driver config](https://docs.harvesterhci.io/v1.8/advanced/index#csi-driver-config) design is not suitable for cases where one storage provider requires different `VolumeSnapshotClass` settings for different scenarios.

I’ll discuss with the team how to extend the [CSI driver config](https://docs.harvesterhci.io/v1.8/advanced/index#csi-driver-config) to support this use case. My current idea is to sync CDI’s per-StorageClass `VolumeSnapshotClass` configuration back to this setting, while keeping the Harvester controller referencing the [CSI driver config](https://docs.harvesterhci.io/v1.8/advanced/index#csi-driver-config) as the single source of truth.

Since I’m working on the larger refactoring scope for [third-party backup/restore support](https://github.com/harvester/harvester/issues/9135), I’d suggest holding this PR temporarily. I’ll apply “change requested” to avoid an unexpected merge.

After [harvester/harvester#9135](https://github.com/harvester/harvester/issues/9135) is decided and merged, I’ll continue this work based on your current implementation and invite you for review.

Thanks.

WebberHuang1118 (changes requested):
Thanks for the idea. Please check [this comment](https://github.com/harvester/harvester/pull/10576#discussion_r3245670092).

pkg/controller/master/backup/backup.go
officialasishkumar: Updated in 62ea239383f1e0253a529948d5070d33166cfb9f to keep VM backup snapshot creation referencing `VolumeSnapshotClass` only through `csi-driver-config`. The StorageClass annotation handling is now limited to the standalone volume snapshot API path.

Issue #5486, opened 2024-03-28
[FEATURE] Support load balancers on non-management interfaces for guest clusters


starbops:
**Is your feature request related to a problem? Please describe.**

If a guest Kubernetes cluster consists of multi-network-attached VMs, all the load balancer IP addresses allocated for the LB type of Services can still only be assigned to the VMs' management interface.

One intention of an additional network is to separate the workload traffic from management traffic in terms of cluster stability and security. Sometimes, we want to leverage the load balancer feature to expose such a workload. However, with the current implementation, the load balancer IP addresses can only attached to the management network. Hence, the traffic still tangles together.

The reason behind is that the `kube-vip` chart deployed is configured with no `vip_interface` specified, so the Pods end up working on the management interfaces.

**Describe the solution you'd like**

We might need the operator logic for `kube-vip` being implemented in `harvester-cloud-provider`, to dynamically decide what network interface `kube-vip` should work on.

**Describe alternatives you've considered**

Manually update the `vip_interface` environment variable for the `kube-vip` DaemonSet can also do the trick.

**Additional context**

With the underlying network infra configured appropriately, the workload endpoint can still be reached despite the LB IP address allocated from the `IP pool-B` associated to `network-B` is attached to the management interface (attached to `network-A`).

For example, creating a RKE2 guest cluster with three nodes using the Rancher Integration feature. Each node will be attached to two discrete networks:

- `net-100`: VLAN100, 192.168.100.0/24
- `net-101`: VLAN101, 192.168.101.0/24

<img width="1532" alt="image" src="https://github.com/harvester/harvester/assets/1827717/db1134a0-3b77-4b55-9a57-c894a0cefd85">

Create an IP pool for `net-101`, called `pool-101`. It should be look like the following:

```shell
$ kubectl get ippools.loadbalancer pool-101 -o yaml
apiVersion: loadbalancer.harvesterhci.io/v1beta1
kind: IPPool
metadata:
  creationTimestamp: "2024-03-27T06:01:44Z"
  finalizers:
  - wrangler.cattle.io/harvester-ipam-controller
  generation: 43
  labels:
    loadbalancer.harvesterhci.io/global-ip-pool: "true"
    loadbalancer.harvesterhci.io/vid: "101"
  name: pool-101
  resourceVersion: "366450180"
  uid: 6804dce7-7ea2-4ce3-9840-0b44c500a8f2
spec:
  ranges:
  - gateway: 192.168.101.1
[...]

PR #10585, opened 2026-05-15
hep: LB IP on non management network interface for guest cluster[skip ci]

Yu-Jack:
#### Problem:
https://github.com/harvester/harvester/issues/5486

#### Solution:

Write common nad-to-interface mapping config map data:

```yaml
apiVersion: v1
data:
  interface-nad-mapping: '{"default/mgmt-vlan1":"enp1s0","default/net123":"enp2s0"}'
kind: ConfigMap
metadata:
  name: harvester-nad-mapping
  namespace: kube-system
```

For ippool & DHCP mode:

1). Frontend reads the config map and display NADs to users
2). Frontend passes `"cloudprovider.harvesterhci.io/network":"{VM Network selected from users}"`
3). Backend converts the VM network to linux interface and put it into `kube-vip.io/serviceInterface: {interface}`. Then kube-vip can bind ip to the correct interface.

Then, for DHCP mode:

4). Remove the `cloudprovider.harvesterhci.io/network` annotation

#### Related Issue(s):
https://github.com/harvester/harvester/issues/5486

cloud-provider: https://github.com/harvester/cloud-provider-harvester/pull/75
frontend: https://github.com/rancher/dashboard/pull/17885
load-balancer: https://github.com/harvester/load-balancer-harvester/pull/110

### DIFF
--- a/enhancements/20260515-lb-on-non-management-interface.md
+++ b/enhancements/20260515-lb-on-non-management-interface.md
@@ -1,0 +1,182 @@
+# Support Load Balancers on Non-Management Interfaces for Guest Clusters
+
+## Summary
+
+When a guest cluster has multiple network interfaces, all `LoadBalancer` Service VIPs are bound to the management interface regardless of which network the IP pool or DHCP server belongs to. This enhancement fixes the binding by allowing users to select a NAD name from the UI; the cloud provider resolves it to the correct Linux interface via a shared ConfigMap and sets `kube-vip.io/serviceInterface` on the Service.
+
+### Related Issues
+
+https://github.com/harvester/harvester/issues/5486
+
+## Motivation
+
+kube-vip defaults to the management interface when no interface hint is provided, defeating network isolation for workload-network IP pools or DHCP servers.
+
+### Goals
+
+- **IP pool mode**: VIPs are automatically bound to the interface matching the IP pool's subnet.
+- **DHCP mode**: Users can select the correct network interface from the UI.
+
+### Non-goals
+
+- Deal with asymmetric network topology where nodes have different interface configurations, e.g.:
+
+```
+Management Node:  enp1s0 → net-mgmt (192.168.100.0/24)
+Worker Node:      enp1s0 → net-mgmt (192.168.100.0/24)
+                  enp2s0 → net-101  (192.168.101.0/24)
+```
+
+## Introduction
+
+`harvester-cloud-provider` runs in the guest cluster and connects to both the guest cluster (to manage Services) and the Harvester cluster (to manage `LoadBalancer` CRs and query VMI resources).
+
+When a `LoadBalancer` Service is created:
+
+1. The CCM service controller calls `EnsureLoadBalancer()`.
+2. The cloud provider creates a `LoadBalancer` CR on Harvester.
+3. Harvester allocates an IP from the matching pool.
+4. The cloud provider writes the IP to `kube-vip.io/loadbalancerIPs`.
+5. kube-vip binds the VIP to an interface — defaulting to management.
+
+**IP pool mode — broken state:**
+
+```
+enp1s0 → net-mgmt (192.168.100.0/24)
+enp2s0 → net-101  (192.168.101.0/24)
+```
+
+```shell
+$ ip addr show enp1s0
+    inet 192.168.101.57/32 scope global enp1s0   ← VIP on wrong interface
+```
+
+**DHCP mode — broken state:**
+
+In DHCP mode `kube-vip.io/loadbalancerIPs` is set to `0.0.0.0`. kube-vip creates a macvlan sub-interface and sends a DHCP request using the macvlan's unique MAC. `serviceInterface: auto` cannot work here — it finds interfaces by subnet match, and `0.0.0.0` matches nothing — so kube-vip falls back to the management interface and is not able to use non-management network.
+
+## Proposal
+
+### IP Pool Mode
+
+**Network selection**: The frontend reads the NAD keys from the `harvester-nad-mapping` ConfigMap (in `kube-system`) and shows them as a dropdown — e.g. `default/net123`. The user's selection sets `cloudprovider.harvesterhci.io/network: default/net123` on the Service. The cloud provider resolves the NAD name to a Linux interface via the ConfigMap and writes `kube-vip.io/serviceInterface` accordingly; the load-balancer controller picks the matching IPPool.
+
+### DHCP Mode
+
+Both DHCP and IP pool modes use the same unified approach: the frontend selects a NAD name and the cloud provider resolves it to a Linux interface via the ConfigMap.
+
+**Interface mapping ConfigMap**
+
+Using `VirtualMachineInstance.spec.networks` (nic name → NAD name) joined with `VirtualMachineInstance.status.interfaces` (nic name → Linux interface name, from the KubeVirt guest agent), the cloud provider syncs a ConfigMap in `kube-system`:
+
+```yaml
+apiVersion: v1
+data:
+  interface-nad-mapping: '{"default/mgmt-vlan1":"enp1s0","default/net123":"enp2s0"}'
+kind: ConfigMap
+metadata:
+  name: harvester-nad-mapping
+  namespace: kube-system
+```
+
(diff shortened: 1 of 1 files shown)


enhancements/20260515-lb-on-non-management-interface.md
```diff
@@ -0,0 +1,189 @@
...
+
+In DHCP mode `kube-vip.io/loadbalancerIPs` is set to `0.0.0.0`. kube-vip creates a macvlan sub-interface and sends a DHCP request using the macvlan's unique MAC. `serviceInterface: auto` cannot work here — it finds interfaces by subnet match, and `0.0.0.0` matches nothing — so kube-vip falls back to the management interface and is not able to use non-management network.
+
+## Proposal
+
+### IP Pool Mode
+
+**Interface binding**: Set `kube-vip.io/serviceInterface: auto` alongside `kube-vip.io/loadbalancerIPs`. kube-vip's `autoFindInterface()` locates the interface whose subnet contains the VIP.
+
+**Network selection**: The cloud provider annotates each Node with `cloudprovider.harvesterhci.io/ippool-network-mapping`, a JSON map of IPPool name → VM network (`spec.selector.network`) for pools scoped to this guest cluster — e.g. `{"n123":"default/net123"}`. The frontend shows a dropdown like `n123 (default/net123)`; the user's selection sets `cloudprovider.harvesterhci.io/network: default/net123` on the Service. The cloud provider copies this to the LoadBalancer CR, and the load-balancer controller picks the matching IPPool.
```
w13915984028: There are some limitations on this solution:
(1) One network can map to multi IPPools, IPPool has selector & priority to match the request dynamically
(2) Pools might be changed dynamically on the Harvester side

Recall the guest-cluster provisioning process:
User creates node-pool, select network/nic on the vm/node interface. For a guest-cluster, it sees a couple of Harvester networks(nads). When create a service, UI supply a dropdown to list all the networks for this guest cluster.

Suppose the cloud-provider annotates all networks information to all nodes, UI utilize this information on service creation: `when creating service, user selects it's target network.`

There is another limitation need to consider:

Rancher encapsulates the guest cluster with multi node pools, in a guest cluster, it could be:

```
node-pool-mgmt: nad-mgmt
node-pool-worker1: nad-mgmt, nad-workload1
node-pool-worker2: nad-mgmt, nad-workload2
```

At the view of guest cluster, it has three nads: nad-mgmt, nad-workload1, nad-workload2; but vms/nodes are in asymetric mode.

For a serivce like service-worker1, IPPool can offer IP based on network `nad-workload1`,  kube-vip needs to set the VIP on node-pool-worker1, and service backend workloads are sheduled to node-pool-worker1.  If/not kue-vip could support this, is to be verified.

Or we limit the symmetric-network-topology for the first version.

enhancements/20260515-lb-on-non-management-interface.md
```diff
@@ -0,0 +1,189 @@
...
+
+Interfaces without a multus `name` (calico, kube-vip macvlan, etc.) are excluded automatically.
+
+**Frontend interface selection**
+
+The frontend reads the annotation from any Node and shows a dropdown when DHCP mode is selected:
+
+```
+enp1s0  (default/mgmt-vlan1)
+enp2s0  (default/net123)       ← user selects this
```
w13915984028: Does this solution has the precondition:

(1) Guest cluster must be in symmetric-network-topology mode 

(2) Guest cluster must have a pre-configured second-network IP in advance, to utilize the `service:externalTrafficPolicy:cluster` (k8s does the internal SNAT/DNAT)?

enhancements/20260515-lb-on-non-management-interface.md
```diff
@@ -0,0 +1,189 @@
...
+```shell
+4: vip-3fa2c1d8@enp2s0: ...   ← macvlan on correct interface
+    inet 192.168.101.82/32 ...
+```
+
+### User Stories
+
+#### Story 1
+
+I want an app exposed on `net-101`. The UI shows the available pool networks scoped to my guest cluster. I select `default/net-101`; the frontend sets `cloudprovider.harvesterhci.io/network: default/net-101` on the Service. The cloud provider allocates an IP from the matching pool; kube-vip automatically places the VIP on `enp2s0`.
```
w13915984028: It's essential to carry `cloudprovider.harvesterhci.io/network` on multi-network scenario. Backend needs to adjust the findNetwork https://github.com/harvester/load-balancer-harvester/blob/47802cf0b7fdf0f25c02cdf06e02c9a7446964ac/pkg/webhook/loadbalancer/mutator.go#L150

w13915984028 (commented):
@Yu-Jack

Thanks for the HEP, some open points.

enhancements/20260515-lb-on-non-management-interface.md
Yu-Jack: Yeah, that's why I opened this PR https://github.com/harvester/load-balancer-harvester/pull/110. I don't want to change the existing flow. So, if LB has this annotation, I'll just use it.

enhancements/20260515-lb-on-non-management-interface.md
Yu-Jack: I don't quite understand this question. Could you explain more? Are you talking to this issue https://github.com/harvester/harvester/issues/10586?

Yu-Jack: I understand the first one and will reply on above discussion.

enhancements/20260515-lb-on-non-management-interface.md
Yu-Jack: Good question! I didn't notice asymmetric-network-topology case. I hope we can focus on symmetric-network-topology for the first version to ensure we won't be out of the scope too much. But, I'll still do some investigation about this. Also, we'll need a document to mention this.

w13915984028: Yes, from Rancher UI, there is no way to restrict guest cluster node pool network selection for now, we just document this limitation for the first stage.

enhancements/20260515-lb-on-non-management-interface.md
```diff
@@ -0,0 +1,195 @@
...
+KeyKubevipServiceInterface = "kube-vip.io/serviceInterface"
+KeyInterfaceNADMapping     = "cloudprovider.harvesterhci.io/interface-nad-mapping"
+KeyIPPoolNetworkMapping    = "cloudprovider.harvesterhci.io/ippool-network-mapping"
+```
+
+No new CRDs, no changes to the `LoadBalancer` CR schema.
+
+**RBAC**: The `harvesterhci.io:cloudprovider` ClusterRole must grant `get/list/watch` on `ippools`. Updated in `deploy/charts/harvester/templates/rbac.yaml`.
+
+The cloud provider's service account (e.g. `default:test`) is currently bound via a namespace-scoped **RoleBinding**, which cannot grant access to cluster-scoped resources like IPPool. The binding in `pkg/api/kubeconfig/generate_handler.go` must be changed from RoleBinding to **ClusterRoleBinding** so the cloud provider can list IPPools at cluster scope.
```
w13915984028: this could be a breaking change.

things might be optimized like:

when create service, user specify the target network (NAD), which is already exposed to user when creating guest cluster, the machine networks(NAD) are explicitly added

this selection is converted to `guestcluster.harvesterhci.io/network` on guest service object annotation

then HCP further translates it to `loadbalancer.harvesterhci.io/network` annotation
https://github.com/harvester/cloud-provider-harvester/blob/fd1321fb509f630405fe4457bd7d8e2df1ef1b42/pkg/cloud-controller-manager/loadbalancer.go#L284

backend loadbalancer uses `loadbalancer.harvesterhci.io/network` as key to match a pool
https://github.com/harvester/load-balancer-harvester/blob/7e4233c7d5b6cc2241e67d26e681c56d6b4e64d9/pkg/webhook/loadbalancer/mutator.go#L234


all of those are existing logic (improved & optimized on latest HCP and loadblancer-harvester)

guest-side, the user interface sticks to Harvester network(NAD) for service creation, and the controller translates it to interface or other information under the hood

use clusterrole to list ippool is not necessary; and list ip pool will also import tight couple with loadbalancer as LB uses a complex logic to determin which pool is fit

Yu-Jack: I've updated the HEP and PR to only show NAD and remove the clusterrolebinding.

w13915984028 (approved):
LGTM, thanks.

